PR #1072, opened 2025-01-23
docs: add SIP transfer via LLM tool call example

jamsea:


examples/dialin-chatbot/bot_daily.py
```diff
@@ -87,6 +141,44 @@ async def on_first_participant_joined(transport, participant):
...
+
+        async def transfer_call(
+            function_name, tool_call_id, args, llm: LLMService, context, result_callback
+        ):
+            logger.debug(f"transfer_call: {function_name} {tool_call_id} {args}")
+
+            # sip_url = "sip:your_user_name@sip.linphone.org"
+
+            sip_url = (
+                f"sip:your_username@dailyco.sip.twilio.com?x-daily_id={room_url.split('/')[-1]}"
```
vr000m, line 161: this sip uri needs to come from the developer. It will be different.

vipyne: What @vr000m said.  `x-daily_id` is not currently used.
sip_uri comes from the response to adding sip related configuration to the daily room.

vr000m: This is a custom parameter that we are sending to Twilio, so that the receiving SIP agent can pull the transcript for that daily room, based on the x-daily-id.

aconchillo:
Closing. We have an example that does that here: https://github.com/pipecat-ai/pipecat-examples/tree/main/phone-chatbot/daily-pstn-call-transfer. If that's not the case, please create a new example here https://github.com/pipecat-ai/pipecat-examples

PR #1079, opened 2025-01-23
Update cartesia.py to use the new model pointer `sonic`

ecdeng:
…onic-english`

We are now using `sonic` as a pointer to the latest stable release (https://docs.cartesia.ai/build-with-sonic/models#continuous-updates). sonic-english will forever point to `sonic-2024-10-19`, which is already out of date.

Updating the model name to the updated naming conventions we rolled out.


src/pipecat/services/cartesia.py
```diff
@@ -88,7 +88,7 @@ def __init__(
         voice_id: str,
         cartesia_version: str = "2024-06-10",
```
xingweitian, line 89: Is cartesia_version still needed in this case?

ecdeng: Yeah that's the API version we should keep that.

xingweitian: Got it!

markbackman (approved):
LGTM! Thanks for letting us know.

PR #1080, opened 2025-01-24
Add voicemail detection example

DominicStewart:
Created an example that shows how to do voicemail detection.


examples/foundational/32-voicemail-detection.py
```diff
@@ -0,0 +1,165 @@
...
+from pipecat.services.openai import OpenAILLMService
+from pipecat.transports.services.daily import DailyParams, DailyTransport
+
+load_dotenv(override=True)
+
+logger.remove(0)
+logger.add(sys.stderr, level="DEBUG")
+
+
+async def start_terminate_call(
```
markbackman, line 35: Is there a reason why you're including this? The state callback is most useful if you want the LLM to say something before the function call takes a while to complete. Perhaps we can remove it?

DominicStewart: Sounds good. I'll remove it and update this PR (But understood that it may end up not actually being an example)

examples/phone-chatbot/bot_voicemail_detection.py
```diff
@@ -0,0 +1,177 @@
+#
```
markbackman, line 1: Can or should this be integrated into the bot_daily.py file?

examples/phone-chatbot/bot_voicemail_detection.py
```diff
@@ -0,0 +1,177 @@
...
+            await start_dialout(transport, dialoutSettings)
+        if state == "left":
+            await task.queue_frame(EndFrame())
+
+    @transport.event_handler("on_first_participant_joined")
+    async def on_first_participant_joined(transport, participant):
+        if not useDialout:
+            logger.info("First participant joined")
+            await transport.capture_participant_transcription(participant["id"])
+            messages.append(
```
markbackman, line 130: You're setting the context in three places? I think it can just be set on line 82. Then, the code on line 136 is used to queue the frames, which will result in a completion.

So, I think you should remove the context from lines 133 and 146.

DominicStewart: Yeah, apologies. I hadn't finished this yet. Was trying a few things but decided to submit what I had yesterday.

examples/phone-chatbot/README.md
```diff
@@ -1,3 +1,5 @@
+<!-- @format -->
```
markbackman, line 1: Was this intentional? Is it needed?

markbackman (approved):
YES!! Thank you for working to integrate this into the example. The dial-out example is more complete, thanks to this.

Just one tiny question, otherwise ✅

examples/phone-chatbot/README.md
DominicStewart: Not intentional, I think it was added by vs code. Will fix and merge

PR #1081, opened 2025-01-24
Added retry functionality and a new callback to the UserIdleProcessor

markbackman:

@aconchillo I've decided to add the retry logic back into the UserIdleProcessor. I've seen some code without it and it's unnecessarily complicated. Instead, we can help developers keep things simple by handling some of this on their behalf. This would be a common enough pattern that we should enable it.

### DIFF
--- a/src/pipecat/processors/user_idle_processor.py
+++ b/src/pipecat/processors/user_idle_processor.py
@@ -5,7 +5,7 @@
 #
 
 import asyncio
-from typing import Awaitable, Callable
+from typing import Awaitable, Callable, Union
 
 from pipecat.frames.frames import (
     BotSpeakingFrame,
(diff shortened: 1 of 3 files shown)


src/pipecat/processors/user_idle_processor.py
```diff
@@ -42,24 +54,60 @@ async def handle_idle(processor: "UserIdleProcessor") -> None:
...
+
+        Args:
+            callback: The callback function to wrap.
+
+        Returns:
+            A wrapped callback that returns bool to indicate whether to continue monitoring.
+        """
+
+        async def wrapper(processor: "UserIdleProcessor", retry_count: int) -> bool:
+            if len(callback.__code__.co_varnames) == 1:
```
filipi87, line 90: If I am not mistaken, `co_varnames` can include all local variables inside the callBack function, that could lead to mistakes here. 
Maybe refactor to use `co_argcount` that should ignore the local variables. 

Or even better, try to use `inspect.signature`: https://docs.python.org/3/library/inspect.html#introspecting-callables-with-the-signature-object

filipi87 (approved):
Just a small comment about how we might improve the way we are checking the correct function to invoke, but other than that, everything looks good. 🚀

PR #1082, opened 2025-01-24
Update function calling examples to push a TextFrame in the start_cal…

markbackman:
…lback

Removing the note in these examples because it's no longer relevant. Also, adding the llm.push_frame(TextFrame(message)) to this example since that's come up in Discord.

Note: this only applies to OpenAILLMService LLMs.


filipi87 (approved):
Makes sense.

PR #1086, opened 2025-01-24
Small fixes to the Daily REST API helper

markbackman:

Two small things related to the Daily REST API helper:
- Fix type validation error by converting timestamp to integer before Pydantic validation
- Improve token parameter construction by using proper Pydantic model constructors instead of dict unpacking

This is clean up from a PR that was merged earlier today.

### DIFF
--- a/src/pipecat/transports/services/helpers/daily_rest.py
+++ b/src/pipecat/transports/services/helpers/daily_rest.py
@@ -296,7 +296,9 @@ class DailyRESTHelper:
             room_url: Daily room URL
             expiry_time: Token validity duration in seconds (default: 1 hour)
             owner: Whether token has owner privileges
-            params: Parameters for creating a Daily meeting token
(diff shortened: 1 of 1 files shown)


filipi87 (approved):
Looks good. 🚀

src/pipecat/transports/services/helpers/daily_rest.py
```diff
@@ -309,25 +311,21 @@ async def get_token(
...
         if params is None:
             params = DailyMeetingTokenParams(
-                **{
-                    "properties": {
-                        "room_name": room_name,
-                        "is_owner": owner,
-                        "exp": int(expiration),
-                    }
-                }
+                properties=DailyMeetingTokenProperties(
```
markbackman, line 322: @filipi87 perhaps we can just use the defined `DailyMeetingTokenProperties`?

PR #1089, opened 2025-01-25
Add text-to-text example for using frame processors

cch41:
It's helpful when developing to be able to interact with the agent text-text in quiet environments. This adds a basic UI and example of how to use frame processors to handle text i/o in a pipeline. Also feel free to suggest better ways of doing this.

Would also be nice if anyone wants to add bypassing TTS completely fast faster responses. It would also be cool to be able to text-only chat with the agent directly in the terminal so we don't need to open a browser window.

Screenshot 2025-01-24 at 4 44 24 PM

### DIFF
--- a/examples/text-exchange-websocket-server/Dockerfile
+++ b/examples/text-exchange-websocket-server/Dockerfile
@@ -1,0 +1,15 @@
+FROM python:3.10-bullseye
+
+RUN mkdir /app
+
+COPY *.py /app/
+COPY requirements.txt /app/
+COPY .env /app/
+
+WORKDIR /app
+
+RUN pip3 install -r requirements.txt
+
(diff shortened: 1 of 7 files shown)


filipi87:
Hi @cch41, we have added two new example on Pipecat:
- A text to text example, as requested in this pr: 
  - https://github.com/pipecat-ai/pipecat/blob/main/examples/foundational/41a-text-only-webrtc.py
- Another example in which a user can speak and/or type text to communicate with the bot
  - https://github.com/pipecat-ai/pipecat/blob/main/examples/foundational/41b-text-and-audio-webrtc.py

So closing this PR in favor of using the examples provided above. 

Let me know in case you need any help.

PR #1098, opened 2025-01-28
fixing a race condition with bot join

Vaibhav159:
There has been a race condition in daily webrtc, where the participant joins the room before the bot, making the `on_participant_joined` event getting missed, is this intended or there is a bug here? Seems like a bug to me, since we are using transport with bot and missing events hampers the bot.


Vaibhav159:
@aconchillo @markbackman open to discussion here, the patch might not be the best, but seems like a bug which should be fixed.

aconchillo:
@Vaibhav159 Thank you for racing the issue. I've submitted a different approach in #1099 . Closing this one.

PR #1099, opened 2025-01-28
transports(daily): queue events until join completes

aconchillo:

Currently, we can get events as soon as we join the signalling server but before `on_joined` is triggered. We now delay events after `on_joined` gets triggered to be more predictable.

Supersedes #1098 .


markbackman (approved):
LGTM. This change makes sense. Nice!

PR #1100, opened 2025-01-28
use `task.cancel()` when participant leaves/disconnects

aconchillo:

If you push an `EndFrame` this will cause the bot to run through everything that is internally queued (which could take seconds). Instead, if a participant disconnects there is nothing else to be sent and therefore we should stop immediately.

### DIFF
--- a/examples/twilio-chatbot/bot.py
+++ b/examples/twilio-chatbot/bot.py
@@ -11,7 +11,6 @@ from dotenv import load_dotenv
 from loguru import logger
 
 from pipecat.audio.vad.silero import SileroVADAnalyzer
-from pipecat.frames.frames import EndFrame
 from pipecat.pipeline.pipeline import Pipeline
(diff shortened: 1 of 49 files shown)


examples/twilio-chatbot/bot.py
```diff
@@ -11,7 +11,7 @@
 from loguru import logger
 
 from pipecat.audio.vad.silero import SileroVADAnalyzer
-from pipecat.frames.frames import EndFrame
+from pipecat.frames.frames import CancelFrame
```
markbackman, line 14: I think you can remove this.

markbackman (approved):
LGTM! Thanks for updating these 🙌

examples/twilio-chatbot/bot.py
aconchillo: Yep, removed!

PR #1102, opened 2025-01-29
Fix breaking change for elevenlabs models

danthegoodman1:
For models like `eleven_turbo_v2` which are better for english, they don't support a language param. This change in 0.0.54 broke them by forcing setting a language param.

This change allows None to actually be set

### DIFF
--- a/src/pipecat/services/elevenlabs.py
+++ b/src/pipecat/services/elevenlabs.py
@@ -197,7 +197,7 @@ class ElevenLabsTTSService(WordTTSService, WebsocketService):
             "sample_rate": sample_rate_from_output_format(output_format),
             "language": self.language_to_service_language(params.language)
             if params.language
-            else "en",
+            else None,
             "output_format": output_format,
             "optimize_streaming_latency": params.optimize_streaming_latency,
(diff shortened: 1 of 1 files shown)


danthegoodman1:
With the websocket client, which is what we used prior to 0.0.54 because it was broken in 0.0.53, we just got a warning log, so it never impacted us:
```
2025-01-28 18:16:56.413 | WARNING  | pipecat.services.elevenlabs:_connect_websocket:336 - Language code [en] not applied. Language codes can only be used with multilingual models: eleven_flash_v2_5, eleven_multilingual_v2, eleven_turbo_v2_5
```

Now it's actually breaking.

danthegoodman1:
Prior to using the HTTP version of this processor (because it did not work in 0.0.53, so we used the normal one), we just saw a warning:
```
2025-01-28 18:16:56.413 | WARNING  | pipecat.services.elevenlabs:_connect_websocket:336 - Language code [en] not applied. Language codes can only be used with multilingual models: eleven_flash_v2_5, eleven_multilingual_v2, eleven_turbo_v2_5
```
So it never impacted us previously

PR #1103, opened 2025-01-29
services(tts): allow pushing silence audio before TTSStoppedFrame

aconchillo:

Allow pushing silence audio frames before `TTSStoppedFrame`. This might be useful for testing purposes, for example, passing bot audio to an STT service which usually needs additional audio data to detect the utterance stopped.


markbackman (approved):
LGTM

PR #1104, opened 2025-01-29
serializers(twilio): handle transport message frames

aconchillo:

`TwilioSerializer` now supports transport message frames. With this we can create Twilio emulators.


markbackman (approved):
LGTM!

PR #1105, opened 2025-01-29
added new websocket client transport

aconchillo:

Just a new websocket transport client.


markbackman (approved):
This looks great!

Can you add an example showing how to use this? That can be a separate task, but may make sense to include here.

aconchillo:
> This looks great!
> 
> Can you add an example showing how to use this? That can be a separate task, but may make sense to include here.

Yes, I have an example for `twilio-chatbot` that I will put in another PR. I will point it from here when it's ready.

PR #1106, opened 2025-01-29
moving test utils inside of package

Vaibhav159:
In 0.0.54 we added test utils but it was outside of pipecat package, making it unusable to anyone outside of pipecat, would like to move it within package so we can test it on custom pipeline and processors.


Vaibhav159:
@aconchillo @markbackman ^

aconchillo:
Yes, makes sense! Thank you @Vaibhav159 !

aconchillo (approved):

PR #1107, opened 2025-01-29
fix: elevenlabs tts increase websocket max message size limit to 16MB

AngeloGiacco:
A potential cause of issues between pipecat and elevenlabs is having the default websocket max message size of 1mb, bumping up to match the maximum used on our end of 16mb

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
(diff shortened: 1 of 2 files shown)


markbackman:
@AngeloGiacco thanks so much. Can you push a changelog entry for this one? I think this would be sufficient.
```
### Fixed

- Fixed an issue where `ElevenLabsTTSService` messages would return a 1009 websocket error by increasing the max message size limit to 16MB.
```

markbackman (approved):
LGTM! Thanks @AngeloGiacco for the fix.

PR #1110, opened 2025-01-29
frames: added metadata field to Frame class

aconchillo:

Added a `metadata` field to `Frame` which makes it possible to pass custom data to all frames.


filipi87 (approved):
Looks good! 
I assume the best way to pass the metadata to the next frame is still being determined, so we’ll handle that in a future PR, right?

aconchillo:
> Looks good! I assume the best way to pass the metadata to the next frame is still being determined, so we’ll handle that in a future PR, right?

Yes, that's right. We need to give that more thought.

PR #1114, opened 2025-01-30
coming back to 0.0.53 to try to isolate bug + modified resampy to soxr

Raphyo:

### DIFF
--- a/src/pipecat/audio/utils.py
+++ b/src/pipecat/audio/utils.py
@@ -8,14 +8,14 @@ import audioop
 
 import numpy as np
 import pyloudnorm as pyln
-import resampy
+import soxr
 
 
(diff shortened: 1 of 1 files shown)


aconchillo:
@Raphyo Hi! What issue did you find?

Raphyo:
Hey Aleix,

it was related to this thread
<https://discord.com/channels/1239284677165056021/1334151685240651817> (but
with websocket transport in my case).  I reverted to the previous pipecat
version (and on the road added soxr to avoid choppy voice) to see if the
openai connection issue I faced was due to 0.0.54, but it's not since the
issue was also there for 0.0.53.

After some investigation, I understood that the issue was related to
geolocation of the servers. Somehow in France, there is a connection issue
with openAI. Also Kevin faced the same issue and is based in France. Not
sure what is the root cause though.

Thanks,
Raph

PR #1115, opened 2025-01-30
Improve ElevenLabs language checking logic

markbackman:

I learned that the "v2" series of models are English only and don't accept language parameter. This change is updating to align to that fact. Also, I'm changing the default language to `None` so that ElevenLabs' defaults are used.


filipi87 (approved):
Looks good. I guess if no language is defined, the default will be English ?

markbackman:
> Looks good. I guess if no language is defined, the default will be English ?

Correct. That's test empirically. I don't think it mentions it in the docs.

PR #1116, opened 2025-01-30
More Storybot updates

chadbailey59:
This changes the strategy for image generation to get more detailed pictures by separately prompting the LLM with the entire story and past image prompts.


src/pipecat/services/google/google.py
```diff
@@ -1029,3 +1031,70 @@ async def run_tts(self, text: str) -> AsyncGenerator[Frame, None]:
...
+                ),
+            )
+            await self.stop_ttfb_metrics()
+
+            if not response or not response.generated_images:
+                logger.error(f"{self} error: image generation failed")
+                yield ErrorFrame("Image generation failed")
+                return
+
+            for img_response in response.generated_images:
```
aconchillo, line 1085: I assume that the `generate_images` above is asynchronous and will generate the images we want in there. That is `response.generated_images` is already populated with the images instead of being a synchronous iterator that blocks at every iteration. I guess it needs to be like that otherwise the `async` functoin would not make much sense. Is that right?

aconchillo (approved):

src/pipecat/services/google/google.py
chadbailey59: That's correct. It's not obvious from the code, but self._client.models.generate_images is synchronous, and the async version is self._client.aio.models.generate_images

PR #1117, opened 2025-01-30
adding conversion of tools from openai to gemini format

Vaibhav159:
Adding logic to automatically convert openai tools format to gemini for easy switch without worrying about tool conversion.


Vaibhav159:
@aconchillo @markbackman this should make things easy while working with multiple llms.

PR #1118, opened 2025-01-30
introduce TaskManager

aconchillo:

This replaces the current global task manager by a `TaskManager` per `PipelineTask`. This avoids having warnings about dangling tasks that have been created by a different `PipelineTask` that is still running.

Also, the `TaskManager` is passed through the `StartFrame`, so two things need to happen:

- Frame processors need to call `super().process_frame()`, otherwise things won't be initialized properly.
- If you want to use the `TaskManager` to create tasks you need to create tasks after `StartFrame` is processed.

It is now possible to specify the event loop a `PipelineTask` should run in. This allows complex applications where `PipelineTask`s can run in multiple threads.

Because we now support multiple threads, the global `obj_id()` and `obj_count()` functions now require a lock.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,19 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- It is now possible to specify the asyncio event loop that a `PipelineTask` and
+  all the processors should run on by passing it as a new argument to the
+  `PipelineRunner`. This could allow running pipelines in multiple threads each
+  one with its own event loop.
+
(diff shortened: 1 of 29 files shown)


chadbailey59 (approved):
LGTM... silly question: do we have an example that shows multiple PipelineTasks?

aconchillo:
> LGTM... silly question: do we have an example that shows multiple PipelineTasks?

We will in a twilio-chatbot update that I'm working on.

markbackman (approved):
LGTM! Walked through the code, which looks good. Ran a set of examples and only found the one issue in RTVI, which you've fixed. I think this one is ready to go.

PR #1120, opened 2025-01-31
First draft at integrating operator dialout example

DominicStewart:
This adds to the phone chatbot example. The next blog post is focusing on transferring calls. There are many ways to 'transfer' calls. One of the ways many customers want to do it is to add an operator to the call by dialing out. So in this case:

1. Customer rings bot
2. Customer has conversation with bot
3. Customer escalates call and asks to speak to manager
4. Bot starts dialout to operator (manager / supervisor)
5. Operator joins the call
6. Bot summarizes call to manager / supervisor (while customer is on the call)
7. Bot waits in the call
8. Customer and operator have a conversation

This example will also allow us to use the voicemail detection + transfer example together. For example:
1. Bot dials out
2. If talking to voicemail, it will detect it and leave a message and put down the call
3. If talking to a human, it will have a conversation
4. Then bot can escalate to operator and dial out to operators number

Still a work in progress, but I wanted to get @markbackman’s thoughts on the approach. I initially considered breaking this into a separate file since we’ll be adding more transfer examples later (e.g., SIP transfer, blind transfer without context, warm transfer where a second bot joins a different room and retrieves call context via RAG). That said, this approach seems extendable without too many issues.

### DIFF
--- a/examples/phone-chatbot/bot_daily.py
+++ b/examples/phone-chatbot/bot_daily.py
@@ -34,6 +34,7 @@ async def terminate_call(
     function_name, tool_call_id, args, llm: LLMService, context, result_callback
 ):
     """Function the bot can call to terminate the call upon completion of a voicemail message."""
+    logger.debug("Terminating call")
     await llm.queue_frame(EndTaskFrame(), FrameDirection.UPSTREAM)
     await result_callback("Goodbye")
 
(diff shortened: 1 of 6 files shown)


DominicStewart:
Still not finished (or tested), but almost there.

PR #1121, opened 2025-01-31
introduce audio resamplers

aconchillo:


src/pipecat/transports/network/fastapi_websocket.py
```diff
@@ -163,7 +163,7 @@ async def write_raw_audio_frames(self, frames: bytes):
 
     async def _write_frame(self, frame: Frame):
         try:
-            payload = self._params.serializer.serialize(frame)
+            payload = await self._params.serializer.serialize(frame)
```
vengad-arrowhead, line 166: can this (as well as deserialzie) be scheduled on different thread as resampling is CPU intensive and using separate thread prevent pipeline stalling ?

aconchillo: We will take care of this in a separate PR.

src/pipecat/audio/utils.py
```diff
@@ -75,39 +91,41 @@ def exp_smoothing(value: float, prev_value: float, factor: float) -> float:
     return prev_value + factor * (value - prev_value)
 
 
-def ulaw_to_pcm(ulaw_bytes: bytes, in_sample_rate: int, out_sample_rate: int):
+def ulaw_to_pcm(ulaw_bytes: bytes, in_rate: int, out_rate: int, resampler: BaseAudioResampler):
     # Convert μ-law to PCM
     in_pcm_bytes = audioop.ulaw2lin(ulaw_bytes, 2)
```
markbackman, line 98: Should this variable be `in_ulaw_bytes`?

src/pipecat/audio/utils.py
```diff
@@ -75,39 +91,41 @@ def exp_smoothing(value: float, prev_value: float, factor: float) -> float:
...
 
 
-def pcm_to_ulaw(pcm_bytes: bytes, in_sample_rate: int, out_sample_rate: int):
+def pcm_to_ulaw(pcm_bytes: bytes, in_rate: int, out_rate: int, resampler: BaseAudioResampler):
     # Resample
-    in_pcm_bytes = resample_audio(pcm_bytes, in_sample_rate, out_sample_rate)
+    in_pcm_bytes = resampler.resample(pcm_bytes, in_rate, out_rate)
 
     # Convert PCM to μ-law
     ulaw_bytes = audioop.lin2ulaw(in_pcm_bytes, 2)
```
markbackman, line 111: Should this be named `out_ulaw_bytes` for consistency with the `ulaw_to_pcm` function?

src/pipecat/audio/utils.py
markbackman: Same comments for the `alaw_to_pcm` and `pcm_to_alaw` functions below.

aconchillo: No, because they are converted to PCM. So `in_pcm_bytes` has raw PCM, not ulaw.

src/pipecat/audio/utils.py
aconchillo: Yeah, we could call it `out_...`. Updated. Also updated the `alaw` one.

src/pipecat/audio/utils.py
aconchillo: Oh! Sorry, you meant the argument!

markbackman (approved):
This looks really good! Thanks for the organized commit history; made reviewing easy.

Also, I ran a number of examples that were impacted by these changes and things work as expected. 🚀

PR #1127, opened 2025-02-03
Remove created LLMFullResponseEndFrames from async TTS providers

chadbailey59:
Storybot uses ElevenLabs' websocket TTS generation. It generates audio one "story page" at a time. There are three story pages and one 'story prompt' in each LLM response.

If the delay between story pages is longer than `stop_frame_timeout_s` (because of image generation, for example), the ElevenLabs `WordTTSService` will emit a `TTSStoppedFrame`, and then when TTS starts again, it emits a `TTSStartedFrame`.

But that `TTSStoppedFrame` also causes the creation of a `LLMFullResponseEndFrame`. This breaks the assistant context aggregator running downstream.

I suppose I could increase my `stop_frame_timeout_s` value to accommodate for the image generation pauses, but that WordTTSService property isn't exposed in the subclasses. Regardless, I think it's incorrect to generate a `LLMFullResponseEndFrame` here, but I admittedly don't know enough about the surrounding context.

EDIT:

My LLM's full output looks like this literal string: `"This is sentence one. [break] This is sentence two. [break] This is sentence three. [break] This is the last sentence."` A processor right after the LLM processor breaks it into multiple text frames, resulting in this behavior:

```
LLMFullResponseStartFrame()
TextFrame("This is sentence one.")
 <4-8 second delay to generate an image based on that sentence>
TextFrame("This is sentence two.")
 <4-8 second delay to generate an image based on that sentence>
TextFrame("This is sentence three.")
 <4-8 second delay to generate an image based on that sentence>
TextFrame("This is the last sentence.")
LLMFullResponseEndFrame()
```

But those extra LLMFullResponseEndFrames make it look like this:

```
LLMFullResponseStartFrame()
TextFrame("This is sentence one.")
 <4-8 second delay to generate an image based on that sentence>
LLMFullResponseEndFrame(inserted by TTS 2 seconds into image gen)
TextFrame("This is sentence two.")
 <4-8 second delay to generate an image based on that sentence>
LLMFullResponseEndFrame(inserted by TTS 2 seconds into image gen)
TextFrame("This is sentence three.")
 <4-8 second delay to generate an image based on that sentence>
LLMFullResponseEndFrame(inserted by TTS 2 seconds into image gen)
TextFrame("This is the last sentence.")
LLMFullResponseEndFrame()
```

The fact that we have a mismatched number of LLMFullResponseStartFrames and LLMFullResponseEndFrames is what makes me call it a bug.

### DIFF
--- a/examples/foundational/14e-function-calling-gemini.py
+++ b/examples/foundational/14e-function-calling-gemini.py
@@ -63,7 +63,7 @@ async def main():
         )
 
         llm = GoogleLLMService(
-            model="gemini-1.5-flash-latest",
+            model="gemini-2.0-flash-exp",
             # model="gemini-exp-1114",
             api_key=os.getenv("GOOGLE_API_KEY"),
         )
--- a/examples/storytelling-chatbot/src/bot.py
+++ b/examples/storytelling-chatbot/src/bot.py
@@ -12,19 +12,22 @@ import sys
 import aiohttp
 from dotenv import load_dotenv
 from loguru import logger
-from processors import StoryImageProcessor, StoryProcessor
+from processors import StoryBreakReinsertProcessor, StoryImageProcessor, StoryProcessor
 from prompts import CUE_USER_TURN, LLM_BASE_PROMPT
 from utils.helpers import load_images, load_sounds
 
 from pipecat.audio.vad.silero import SileroVADAnalyzer
 from pipecat.frames.frames import EndFrame
+from pipecat.pipeline.parallel_pipeline import ParallelPipeline
 from pipecat.pipeline.pipeline import Pipeline
 from pipecat.pipeline.runner import PipelineRunner
+from pipecat.pipeline.sync_parallel_pipeline import SyncParallelPipeline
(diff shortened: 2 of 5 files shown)


aconchillo:
I believe the problem with this approach is that the assistant context aggregator is expecting `LLMFullResponseEndFrame`. I need to think about this.

chadbailey59:
The assistant context aggregator is definitely expecting `LLMFullResponseEndFrame`... and it gets one, eventually, when the LLM response is actually finished. :)

To state the problem a different way: If there's a custom FrameProcessor between the LLM output and the TTS input, and that FrameProcessor has a good reason to queue/buffer parts of the LLM response for more than a few seconds, the TTS processor will create a `LLMFullResponseEndFrame` after the first delay. But because the TTS processor doesn't then emit a `LLMFullResponseStartFrame` _after_ the pause, the context aggregator misses everything after that first pause because of the fake `LLMFullResponseEndFrame`.

chadbailey59:
I'm going to close this PR for now because I'm using a different approach for Storybot, but I'll open an issue to reference this.

PR #459, opened 2024-09-14
add OpenAISTTService

joachimchauvet:
This also works with OpenAI-compatible API providers such as Groq (requested in #254) by passing the `base_url`.

```python
stt = OpenAISTTService(
    api_key=os.getenv("GROQ_API_KEY"),
    base_url="https://api.groq.com/openai/v1",
    model="distil-whisper-large-v3-en"
)
```

PR #1128, opened 2025-02-04
Add Whisper STT service using OpenAI API

golbin:
This patch originates from #459 but has been stuck for a while, so I created this version with a small fix.


src/pipecat/services/openai.py
```diff
@@ -390,6 +397,61 @@ async def run_image_gen(self, prompt: str) -> AsyncGenerator[Frame, None]:
...
+    def can_generate_metrics(self) -> bool:
+        return True
+
+    async def run_stt(self, audio: bytes) -> AsyncGenerator[Frame, None]:
+        try:
+            await self.start_processing_metrics()
+            await self.start_ttfb_metrics()
+
+            response: Transcription = await self._client.audio.transcriptions.create(
+                file=("audio.wav", audio, "audio/wav"), model=self.model_name
```
aconchillo, line 436: Does this save a file to disk? We should not add that by default, probably an argument to the constructor, like `save_audio_file: Optional[str]: None`.

aconchillo: Ooops, duh! Sorry, I misread the code!

aconchillo (approved):

aconchillo:
LGTM! @golbin Do you mind fixing the conflicts? :pray:

golbin:
@aconchillo I resolved the conflict

golbin:
@aconchillo everything OK?

markbackman:
Thanks @golbin!

PR #1129, opened 2025-02-04
Pipecat improvements for the instant voice demo

filipi87:
Preventing from trying to join twice when using ParallalePipeline.

### DIFF
--- a/src/pipecat/transports/network/websocket_server.py
+++ b/src/pipecat/transports/network/websocket_server.py
@@ -6,6 +6,7 @@
 
 import asyncio
 import io
+import json
 import time
 import wave
 from typing import Awaitable, Callable, Optional
@@ -21,6 +22,9 @@ from pipecat.frames.frames import (
     OutputAudioRawFrame,
     StartFrame,
     StartInterruptionFrame,
+    TextFrame,
+    TransportMessageFrame,
+    TransportMessageUrgentFrame,
 )
 from pipecat.processors.frame_processor import FrameDirection
 from pipecat.serializers.base_serializer import FrameSerializer
@@ -46,6 +50,7 @@ class WebsocketServerCallbacks(BaseModel):
     on_client_connected: Callable[[websockets.WebSocketServerProtocol], Awaitable[None]]
     on_client_disconnected: Callable[[websockets.WebSocketServerProtocol], Awaitable[None]]
     on_session_timeout: Callable[[websockets.WebSocketServerProtocol], Awaitable[None]]
+    on_websocket_ready: Callable[[], Awaitable[None]]
 
 
 class WebsocketServerInputTransport(BaseInputTransport):
@@ -96,6 +101,7 @@ class WebsocketServerInputTransport(BaseInputTransport):
     async def _server_task_handler(self):
         logger.info(f"Starting websocket server on {self._host}:{self._port}")
         async with websockets.serve(self._client_handler, self._host, self._port) as server:
+            await self._callbacks.on_websocket_ready()
             await self._stop_server_event.wait()
 
     async def _client_handler(self, websocket: websockets.WebSocketServerProtocol, path):
@@ -186,6 +192,12 @@ class WebsocketServerOutputTransport(BaseOutputTransport):
             await self._write_frame(frame)
             self._next_send_time = 0
 
+    async def send_message(self, frame: TransportMessageFrame | TransportMessageUrgentFrame):
+        message_frame = TextFrame(
+            text=json.dumps(frame.message),
+        )
+        await self._write_frame(message_frame)
+
     async def write_raw_audio_frames(self, frames: bytes):
         if not self._websocket:
             # Simulate audio playback with a sleep.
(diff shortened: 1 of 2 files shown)


markbackman:
@filipi87 The daily changes look good! Given the direction of the demo, should we keep the websocket changes?

I'll wait to approve for now. Also, you'll want to squash before merging.

markbackman (approved):
LGTM!

filipi87:
> @filipi87 The daily changes look good! Given the direction of the demo, should we keep the websocket changes?

I think it's okay to keep the WebSocket changes since they only send the RTVI messages as text

src/pipecat/transports/services/daily.py
```diff
@@ -291,6 +289,7 @@ def __init__(
         self._transcription_ids = []
         self._transcription_status = None
 
+        self._joining = False
```
aconchillo, line 292: Why do we need this? I had this before, but since we now guarantee that `join()` block until completed. There shouldn't be a case where you try to join more than once concurrently.

src/pipecat/transports/services/daily.py
```diff
@@ -1198,7 +1200,22 @@ async def _on_error(self, error):
...
+                for base64_audio in filter(None, audio_list):  # Filter out None values
+                    pcm_bytes = base64.b64decode(base64_audio)
+                    frame = InputAudioRawFrame(
+                        audio=pcm_bytes,
+                        sample_rate=data["sampleRate"],
+                        num_channels=data["numChannels"],
+                    )
+                    await self._input.push_audio_frame(frame)
+            else:
+                await self._input.push_app_message(message, sender)
```
aconchillo, line 1218: I don't think this should be included. If I understand correctly, this is to make a demo and it's not a standard app message.

I think this should go in the demo itself as:

```
@transport.event_handler("on_app_message"):
async def on_app_message(transport, message, sender):
            if message["type"] in {"raw-audio", "raw-audio-batch"}:
                data = message["data"]
                audio_list = data.get(
                    "base64AudioBatch", [data.get("base64Audio")]
                )  # Ensure a list

                for base64_audio in filter(None, audio_list):  # Filter out None values
                    pcm_bytes = base64.b64decode(base64_audio)
                    frame = InputAudioRawFrame(
                        audio=pcm_bytes,
                        sample_rate=data["sampleRate"],
                        num_channels=data["numChannels"],
                    )
                    await transport.input().push_audio_frame(frame)
```

src/pipecat/transports/services/daily.py
```diff
@@ -1198,7 +1200,22 @@ async def _on_error(self, error):
 
     async def _on_app_message(self, message: Any, sender: str):
         if self._input:
-            await self._input.push_app_message(message, sender)
+            if message["type"] in {"raw-audio", "raw-audio-batch"}:
```
aconchillo, line 1203: I just ran into an exception using Prebuilt:

```
  File "/home/aleix/src/pipecat-ai/pipecat/src/pipecat/transports/services/daily.py", line 1203, in _on_app_message
    if message["type"] in {"raw-audio", "raw-audio-batch"}:
       └ {'event': 'request-chat-history'}

KeyError: 'type'
```

src/pipecat/transports/network/websocket_server.py
```diff
@@ -186,6 +192,12 @@ async def process_frame(self, frame: Frame, direction: FrameDirection):
             await self._write_frame(frame)
             self._next_send_time = 0
 
+    async def send_message(self, frame: TransportMessageFrame | TransportMessageUrgentFrame):
+        message_frame = TextFrame(
+            text=json.dumps(frame.message),
+        )
+        await self._write_frame(message_frame)
```
aconchillo, line 199: I don't think this should go in here. This should be the responsibility of the serializer that knows how to transform a `TransportMessageFrame` into whatever is needed.

So, in the demo you would implement a serializer that performs the conversion.

aconchillo:
@filipi87 @markbackman I think we should revert this PR and only include the `on_websocket_ready` change in a new PR. I think everything else can be done in the specific demo.

markbackman:
> @filipi87 @markbackman I think we should revert this PR and only include the `on_websocket_ready` change in a new PR. I think everything else can be done in the specific demo.

@aconchillo I think we want to make this available outside of the demo too, for anyone that wants a fast client connection. Maybe we should have a quick chat?

aconchillo:
For the record, we discussed the comments above and we will do the following changes in a subsequent PR:

- Remove `TextFrame` from `send_message()` that should be handled by  serializer.
- Implement `send_message()` in `FastAPIWebsocketTransport`.
- Maybe add a `audio_in_stream_on_start` to `TransportParams`, or actually maybe add a `enable_audio_in_on_start(enable: bool)` to BaseInputTransport.
- Add a new `start_audio_in_streaming()` that will need to be used if we disabled audio in streaming.
- Move RTVI specific logic to `RTVIProcessor`. We will need to pass the transport to `RTVIProcessor` constructor as optional and initialized to None. From here we will be able to call `start_audio_in_streaming()`.

filipi87:
All the changes mentioned above are now implemented in this PR:
https://github.com/pipecat-ai/pipecat/pull/1215/files

PR #1131, opened 2025-02-04
introduce PipelineParams audio input/output sample rates

aconchillo:

This PR adds new fields to `PipelineParams` to control audio input and output sample rates for the whole pipeline. This allows controlling sample rates from a single place instead of having to specify sample rates in each service. Setting a sample rate to a service is still possible and will override the value from `PipelineParams`.

It also removes the ability to change sample rates at runtime since this is not supported by many services or transports and it doesn't make much sense to do so.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- Added new fields to `PipelineParams` to control audio input and output sample
+  rates for the whole pipeline. This allows controlling sample rates from a
+  single place instead of having to specify sample rates in each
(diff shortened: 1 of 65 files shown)


markbackman (approved):
LGTM! I tested all of the changed services (except for XTTS, which I don't have a setup for). I found a few small issues and pushed fixes as individual commits. You might want to look at them to see if you have any changes you want to make.

This is going to be such a nice developer experience fix! 🙌

aconchillo:
> LGTM! I tested all of the changed services (except for XTTS, which I don't have a setup for). I found a few small issues and pushed fixes as individual commits. You might want to look at them to see if you have any changes you want to make.
> 
> This is going to be such a nice developer experience fix! 🙌

Thank you! Np, I tested XTTS.

PR #1133, opened 2025-02-04
Fixing the issue in Krisp when trying to create more than one

filipi87:
Fixing the issue in Krisp when trying to create more than one krispAudioProcessor.

### DIFF
--- a/src/pipecat/audio/filters/krisp_filter.py
+++ b/src/pipecat/audio/filters/krisp_filter.py
@@ -20,6 +20,22 @@ except ModuleNotFoundError as e:
     raise Exception(f"Missing module: {e}")
 
 
+class KrispProcessorManager:
+    """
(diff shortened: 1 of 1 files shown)


src/pipecat/audio/filters/krisp_filter.py
```diff
@@ -20,6 +20,24 @@
...
 
 
+class KrispProcessorManager:
+    """
+    Ensures that only one KrispAudioProcessor instance exists for the entire program.
+    """
+    _krisp_instance = None
+
+    @classmethod
+    def get_processor(cls, sample_rate: int, sample_type: str, channels: int, model_path: str):
```
filipi87, line 30: This is a workaround for cases where we know Krisp is always used with the same sample_rate, sample_type, and model_path. Further investigation is needed to fix the issue and release a new version of pipecat-krisp.

markbackman (approved):
LGTM

PR #1136, opened 2025-02-05
Create perplexity.py

porcelaincode:

Adds `PerplexityLLMService` for using sonar service


markbackman:
Thanks so much for the proposal. If you don't mind, I'd like to go with this PR: https://github.com/pipecat-ai/pipecat/pull/1177. It follows the pattern we've set in place where we extend the `OpenAILLMService` base class only as functions need to overridden. This keeps things well aligned with OpenAI and easier to maintain.

PR #1148, opened 2025-02-06
Fix sample rate handling in Twilio and Telnyx serializers

markbackman:

The serializers were incorrectly using the output sample rate when creating input audio frames. This caused audio detection to fail when input and output sample rates differed (e.g., 8kHz input for Deepgram, 16kHz output for ElevenLabs).

Added comments to make this code easier to understand.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,9 +9,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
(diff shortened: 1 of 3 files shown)


aconchillo:
LGTM! So weird! I would have swore I fixed this locally 100%! I might have reverted back by mistake :thinking:

aconchillo:
In these serializers it's a bit confusing. Since Twilio/Telnyx is both input an output. With this change the input sample rate is assumed to the output as well.

aconchillo (approved):

PR #1153, opened 2025-02-06
BaseInputTransport: show VAD results when interruptions not allowed

aconchillo:

When interruptions was disabled we were not logging user started/stopped speaking.


markbackman (approved):
LGTM

PR #1154, opened 2025-02-06
serializers: don't update twilio/telnyx sample rates

aconchillo:

We don't want to propagate the pipeline input/output sample rates into the serializers, instead we need to do that explicitly from the serializer constructor.

For example, `TwilioFrameSerializer` has `twilio_sample_rate` and `sample_rate`:

- The `twilio_sample_rate` inidicate the sample rate that Twilio will send and receive.
- The `sample_rate` indicates the sample rate of the input transport.

So if `twilio_sample_rate` is 8000 and `sample_rate` is 16000, one we receive audio from Twilio we will resample it to 16000. Then, through the pipeline we will work with 16000. At the end, the serializer will convert 16000 to 8000 which is what Twilio is expecting.


markbackman (approved):
LGTM! This further simplifies things.

PR #1157, opened 2025-02-06
fixing test case error

Vaibhav159:
This pull request includes changes to fix an issue with test imports and improve logging in the `pipecat` module. The most important changes are listed below:

Bug Fixes:
* Fixed an issue causing test import to fail due to `logger.remove` in `CHANGELOG.md`.

Logging Improvements:
* Added a check for existing handlers before removing the logger in `src/pipecat/tests/utils.py`.

### DIFF
--- a/src/pipecat/tests/utils.py
+++ b/src/pipecat/tests/utils.py
@@ -22,7 +22,8 @@ from pipecat.pipeline.runner import PipelineRunner
 from pipecat.pipeline.task import PipelineParams, PipelineTask
 from pipecat.processors.frame_processor import FrameDirection, FrameProcessor
 
-logger.remove(0)
+if logger._core.handlers:
+    logger.remove(0)
 logger.add(sys.stderr, level="TRACE")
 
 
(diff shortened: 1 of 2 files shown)


Vaibhav159:
@aconchillo @markbackman seems like this got introduced in 0.0.55 release.

src/pipecat/tests/utils.py
```diff
@@ -22,7 +22,8 @@
 from pipecat.pipeline.task import PipelineParams, PipelineTask
 from pipecat.processors.frame_processor import FrameDirection, FrameProcessor
 
-logger.remove(0)
+if logger._core.handlers:
+    logger.remove(0)
```
aconchillo, line 26: What if we just do `logger.remove()`? Accessing private members is usually not a good idea.

aconchillo: Another option is to simply remove these lines. I added them so I didn't have to do it in every test, but maybe it's better for people to control logging themselves.

aconchillo: Yes, I think we want this last option. I'll create a PR for that.

aconchillo:
Thank you @Vaibhav159! I really appreciate all your PRs!!! :heart: . I just created https://github.com/pipecat-ai/pipecat/pull/1160 so we can include in 0.0.56 which we will release in a few minutes.

PR #1159, opened 2025-02-06
Gemini 2.0 Flash Lite RAG example

adriancowham:
Added an example to the foundational dir that demonstrates how to use the Gemini 2.0 Flash Lite model for RAG instead of using traditional RAG approaches.

### DIFF
--- a/examples/foundational/33-gemini-rag.py
+++ b/examples/foundational/33-gemini-rag.py
@@ -3,6 +3,49 @@
 #
 # SPDX-License-Identifier: BSD 2-Clause License
 #
+"""CrossFit Games 2025 Rulebook RAG Demo.
+
+This example demonstrates a Model-Assisted Generation (MAG) chatbot using Google's Gemini model.
+This example uses 2 Gemini models:
+- Gemini 2.0 Flash: This is the voice model that is used to generate the response.
+- Gemini 2.0 Flash Lite: This is the model that is used to answer questions about the CrossFit Games 2025 rulebook - information that isn't yet publicly
+indexed by Gemini (or any other LLM).
+
+How it works:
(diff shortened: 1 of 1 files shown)


adriancowham:
I included a file that is 38K tokens to use as the RAG context. Not sure if you want that checked into the repo or not. Happy to remove it and have devs add their own file.

examples/foundational/32-gemini-rag.py
```diff
@@ -0,0 +1,209 @@
...
+
+logger.remove(0)
+logger.add(sys.stderr, level="INFO")
+
+video_participant_id = None
+
+
+def get_rag_content():
+    """Get the cache content from the file."""
+    with open("assets/rag-content.txt", "r") as f:
```
markbackman, line 38: This looks up the rag-content.txt locally. Can we make this so the script can be run from any location? This should do it:
```
def get_rag_content():
    """Get the cache content from the file."""
    script_dir = os.path.dirname(os.path.abspath(__file__))
    rag_content_path = os.path.join(script_dir, "assets", "rag-content.txt")
    with open(rag_content_path, "r") as f:
        return f.read()
```

markbackman:
Nice example! Thanks for creating it. Would you mind doing two things:
1. Rename the file to `33-gemini-rag.py`. (32 is already taken)
2. Can you add a docstring at the top of the `33-gemini-rag.py` explaining what it is and how to use it. For example, mention that this is data from the CrossFit Games 2025 rulebook, which is not yet indexed by Gemini. Ask questions like: a, b, c to use the knowledge base.

Also, one comment about ensuring the rag-content.py file is usable no matter where the script is run from.

adriancowham:
@markbackman Thanks for the reviewing the PR and for the feedback! I addressed your comments, let me know if there's anything else I can do to improve the PR.

PR #1163, opened 2025-02-07
Add RimeTTSService

markbackman:

In progress. Waiting to resolve a few things related to word/timestamp alignment.


src/pipecat/services/rime.py
```diff
@@ -4,20 +4,338 @@
...
+        # Store service configuration
+        self._api_key = api_key
+        self._url = url
+        self._voice_id = voice_id
+        self._model = model
+        self._settings = {
+            "speaker": voice_id,
+            "modelId": model,
+            "audioFormat": "pcm",
+            "samplingRate": sample_rate,
```
aconchillo, line 118: Maybe initialize this to 0, I think it easier to tell that a 0 still means uninitialized (even if the constructor is None)

src/pipecat/services/rime.py
```diff
@@ -4,20 +4,338 @@
...
+        """Close websocket connection and clean up tasks."""
+        await self._disconnect_websocket()
+        if self._receive_task:
+            await self.cancel_task(self._receive_task)
+            self._receive_task = None
+
+    async def _connect_websocket(self):
+        """Connect to Rime websocket API with configured settings."""
+        try:
+            settings = {k: str(v) for k, v in self._settings.items() if v is not None}
```
aconchillo, line 187: we might not need this since `samplingRate` is now != None (if we do the other change)

src/pipecat/services/rime.py
```diff
@@ -4,20 +4,338 @@
...
+                        else:
+                            # Add space between words (not before punctuation)
+                            needs_space = word_pairs and not words[i - 1].strip(",.!?") == ""
+                            if needs_space:
+                                word = " " + word
+                            word_pairs.append((word, adjusted_start))
+
+                    if word_pairs:
+                        await self.add_word_timestamps(word_pairs)
+                        self._cumulative_time = ends[-1] + self._cumulative_time
```
aconchillo, line 278: it would be nice to put all this in a function like we do in elevenlabs.

markbackman:
@aconchillo good feedback! I updated those points.

I'm seeing an issue with the word timings. I think the timings are faster than the audio stream which results in: 1) incorrectly timed onTtsText callbacks (for RTVI), 2) not getting the complete text into the LLM context. I reached out to Rime to see if this is an issue that they're aware of. For now, I'll leave this as a draft.

src/pipecat/services/rime.py
```diff
@@ -4,20 +4,344 @@
...
+    async def _handle_interruption(self, frame: StartInterruptionFrame, direction: FrameDirection):
+        """Handle interruption by clearing current context."""
+        await super()._handle_interruption(frame, direction)
+        await self.stop_all_metrics()
+        if self._context_id:
+            await self._get_websocket().send(json.dumps(self._build_clear_msg()))
+            self._started = False
+            self._context_id = None
+
+    def _calculate_word_times(self, words: list, starts: list, ends: list) -> list:
```
filipi87, line 223: It feels like the `ends` list is included as a function argument, but it is never actually used in the function logic. Do we need it here ?

filipi87 (approved):
I left a comment about one possible improvement, but everything looks good.

src/pipecat/services/rime.py
markbackman: We do need to pass `ends` for the zip iteration, but don't need it in the loop. So, I'm updating to make that clear.

PR #1166, opened 2025-02-07
rtvi: separate specific google RTVI into a GoogleRTVIObserver

### LINKED ISSUE #1165: RTVI should not couple with Google service
In this [pr](https://github.com/pipecat-ai/pipecat/pull/1051), in rtvi.py:

```
from pipecat.services.google.frames import LLMSearchOrigin, LLMSearchResponseFrame
```

Now users have to install google dependencies even they do not use google.

aconchillo:

This is untested. But the idea is that you would use `GoogleRTVIObserver` instead of the `RTVIObserver`.

Fixes #1165

### DIFF
--- a/examples/news-chatbot/server/news_bot.py
+++ b/examples/news-chatbot/server/news_bot.py
@@ -23,7 +23,7 @@ from pipecat.processors.frame_processor import FrameDirection, FrameProcessor
 from pipecat.processors.frameworks.rtvi import RTVIConfig, RTVIProcessor
 from pipecat.services.cartesia import CartesiaTTSService
 from pipecat.services.deepgram import DeepgramSTTService
-from pipecat.services.google import GoogleLLMService, LLMSearchResponseFrame
+from pipecat.services.google import GoogleLLMService, GoogleRTVIObserver, LLMSearchResponseFrame
 from pipecat.transports.services.daily import DailyParams, DailyTransport
 from pipecat.utils.text.markdown_text_filter import MarkdownTextFilter
 
@@ -102,6 +102,7 @@ async def main():
 
         llm = GoogleLLMService(
             api_key=os.getenv("GOOGLE_API_KEY"),
+            model="gemini-1.5-flash-002",
             system_instruction=system_instruction,
             tools=tools,
         )
@@ -141,7 +142,7 @@ async def main():
             pipeline,
             PipelineParams(
                 allow_interruptions=True,
-                observers=[rtvi.observer()],
+                observers=[GoogleRTVIObserver(rtvi)],
             ),
         )
 
--- a/src/pipecat/services/google/rtvi.py
+++ b/src/pipecat/services/google/rtvi.py
@@ -27,6 +27,9 @@ class RTVIBotLLMSearchResponseMessage(BaseModel):
 
 
 class GoogleRTVIObserver(RTVIObserver):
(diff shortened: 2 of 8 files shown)


aconchillo:
Another option is that we make the frames generic.

aconchillo:
> Another option is that we make the frames generic.

A processor in the `processors` package should not depend on things in the `services` package.

markbackman:
@aconchillo looks good. I made a few updates to get this one working.

@filipi87 with the rollout of gemini 2.0, I realized that Google changed the google search tool format a bit. I'm not updating that here, but we should do that at some point. For now, I'm pinning the examples to Gemini 1.5 Flash.

filipi87:
> Another option is that we make the frames generic.

I think this option could make sense. After all, I believe we expect the other LLM to provide similar features in the near future. What do you think?

That is why I had created the frame as `LLMSearchResponseFrame`, instead of something specific to Google.

filipi87:
> @filipi87 with the rollout of gemini 2.0, I realized that Google changed the google search tool format a bit. I'm not updating that here, but we should do that at some point. For now, I'm pinning the examples to Gemini 1.5 Flash.

Yes, it was only working with gemini 1.5. We would need to add support for gemini 2.0, I guess this will require even some changes inside our GoogleLLMService.

src/pipecat/services/google/rtvi.py
```diff
@@ -0,0 +1,54 @@
...
+
+class RTVIBotLLMSearchResponseMessage(BaseModel):
+    label: Literal["rtvi-ai"] = "rtvi-ai"
+    type: Literal["bot-llm-search-response"] = "bot-llm-search-response"
+    data: RTVISearchResponseMessageData
+
+
+class GoogleRTVIObserver(RTVIObserver):
+    def __init__(self, rtvi: FrameProcessor):
+        super().__init__(rtvi)
```
aconchillo, line 31: I just added this to make it more clear.

examples/news-chatbot/server/news_bot.py
```diff
@@ -141,7 +142,7 @@ async def main():
             pipeline,
             PipelineParams(
                 allow_interruptions=True,
-                observers=[rtvi.observer()],
+                observers=[GoogleRTVIObserver(rtvi)],
```
aconchillo, line 145: Seeing this change I'm thinking we should deprecate `rtvi.observer()` and force people to instantiate `RTVIObserver` directly.

markbackman: Not a bad idea! That would avoid having to add this:
```
rtvi = RTVIProcessor(config=RTVIConfig(config=[]))

PipelineParams(
    observers=[rtvi.observer()],
),
```

That becomes:
```
PipelineParams(
    observers=[RTVIObserver()],
),
```

Much cleaner!

aconchillo: Well, you would do `RTVIObserver(rtvi)`, you still need access to the `rtvi` object.

aconchillo: I just updated the PR with these changes and updated the CHANGELOG.

markbackman: Oh right :/

Still, probably cleaner.

markbackman (approved):
LGTM!

filipi87 (approved):
Makes sense. 🚀

PR #1176, opened 2025-02-08
Deprecate stt_service parameter in STTMuteFilter

markbackman:

The `STTMuteFilter` currently takes an `STTService` parameter but only uses it to check mute state. This is unnecessary coupling since the filter can manage this state internally.

This PR:
- Makes the `stt_service` parameter optional
- Adds a deprecation warning when the parameter is provided
- Implements internal mute state tracking
- Maintains full backwards compatibility

I've tested using the 24 example that this doesn't introduce a regression.

### DIFF
--- a/src/pipecat/processors/filters/stt_mute_filter.py
+++ b/src/pipecat/processors/filters/stt_mute_filter.py
@@ -11,7 +11,6 @@ such as during function calls, bot speech, or custom conditions. It helps manage
 the STT service should be active or inactive during a conversation.
 """
 
-import warnings
 from dataclasses import dataclass
 from enum import Enum
 from typing import Awaitable, Callable, Optional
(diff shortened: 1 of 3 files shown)


src/pipecat/processors/filters/stt_mute_filter.py
```diff
@@ -71,28 +72,36 @@ class STTMuteFilter(FrameProcessor):
...
 
     Args:
-        stt_service: Service handling speech-to-text functionality
         config: Configuration specifying muting strategies
+        stt_service: STT service instance (deprecated, will be removed in future version)
         **kwargs: Additional arguments passed to parent class
     """
 
-    def __init__(self, stt_service: STTService, config: STTMuteConfig, **kwargs):
+    def __init__(self, config: STTMuteConfig, stt_service: Optional[STTService] = None, **kwargs):
```
aconchillo, line 80: This looks like a breaking change, which is fine. We should foce keyword arguments which save us from breaking changes like this:

```
 def __init__(self, *, config: STTMuteConfig, stt_service: Optional[STTService] = None, **kwargs):
 ```

File: src/pipecat/processors/filters/stt_mute_filter.py [hunk 2]
```diff
@@ -71,28 +72,36 @@ class STTMuteFilter(FrameProcessor):
...
         super().__init__(**kwargs)
-        self._stt_service = stt_service
         self._config = config
+        if stt_service is not None:
+            warnings.warn(
```
aconchillo, line 84: Move `import warnings` here so we can remove the whole block in the future and we don't forget removing the `import`.

aconchillo (approved):

aconchillo:
LGTM! Minor comments.

src/pipecat/processors/filters/stt_mute_filter.py
markbackman: This change, in itself, it's breaking, but I do want to remove the `stt_service` arg which will be breaking. That's why I'm including the deprecation warning.

PR #1177, opened 2025-02-08
Add PerplexityLLMService

markbackman:

There was a community request for this. Interestingly, they don't support function calling, but they do searches natively. Pretty slick. (It would be great if they did both!)


aconchillo (approved):

aconchillo:
LGTM!

PR #1182, opened 2025-02-08
Update 28c option to output to log line only by default

markbackman:

Making the DB option one of the TranscriptHandlers you need to opt-in to for this example. This prevents `example.db` from being creating in your directory.


aconchillo (approved):

aconchillo:
LGTM!

Issue #254, opened 2024-06-25
[Feature request] Groq transcriptions support


gaceladri:
[Here](https://console.groq.com/docs/speech-text) are the docs for the new Whisper models available in Groq Cloud. I'd like to replace my Deepgram STT provider with the Groq transcription service. It would be great if you could add this integration. 🙏

aconchillo:
> [Here](https://console.groq.com/docs/speech-text) are the docs for the new Whisper models available in Groq Cloud. I'd like to replace my Deepgram STT provider with the Groq transcription service. It would be great if you could add this integration. 🙏

Is this real-time?

gaceladri:
Great question! Let's compare Deepgram and Groq in terms of speed and cost:

Speed:
According to recent [benchmarks](https://twitter.com/realsanketp/status/1807158858168795450), Deepgram still appears to be the fastest option for speech-to-text processing, despite Groq's impressive performance in other areas.

Cost:
- Groq: $0.03 per hour
- Deepgram: $0.258 per hour

This makes Groq approximately 8.6 times cheaper than Deepgram for the same amount of processing time.

While Groq offers a significant cost advantage, Deepgram maintains its edge in processing speed. The choice between the two may depend on your specific use case:

- If speed is your top priority, Deepgram might be the better option.
- If you're looking to optimize costs and can tolerate slightly slower processing, Groq could be more suitable.

Considering these benchmarks, I'd recommend maintaining Deepgram as our primary speech-to-text (STT) provider. Its superior speed continues to outweigh the cost savings offered by Groq. At least for my app.

gaceladri:
Feel free to close the issue if you don't want to put this on your priority list. I'd prioritize the function calling with Gemini more, as that would bring much greater benefits by enabling us to use Gemini 1.5 Flash instead of GPT-4o.

skelleex:
Would also like to have Groq LLMs and Whisper STT from Groq implemented.

joachimchauvet:
@skelleex It is already possible to use LLMs running on Groq with the OpenAILLMService
```
llm = OpenAILLMService(
    name="LLM", api_key=GROQ_API_KEY, model=model_name, base_url="https://api.groq.com/openai/v1"
)
```

This feature request is mostly related to OpenAI translation API for cloud-hosted whisper STT.

markbackman:
Added here: https://github.com/pipecat-ai/pipecat/pull/1183

PR #1183, opened 2025-02-08
Add GroqSTTService, BaseWhisperSTTService, and refactor OpenAISTTService

markbackman:

- Add `BaseWhisperSTTService`, which is the base class for `OpenAISTTService` and `GroqSTTService`.
- Refactored `OpenAISTTService` to use this new base class.
- Added `GroqSTTService` which now uses the new base class too.

Updated 14f example to use `GroqSTTService`.

While I was at it, updated `GroqLLMService` to use `llama-3.3-70b-versatile` as the default model.

Also addresses issue: https://github.com/pipecat-ai/pipecat/issues/254.


src/pipecat/services/whisper_base.py
```diff
@@ -0,0 +1,94 @@
...
+    from openai.types.audio import Transcription
+except ModuleNotFoundError as e:
+    logger.error(f"Exception: {e}")
+    logger.error(
+        "In order to use OpenAI, you need to `pip install pipecat-ai[openai]`. Also, set `OPENAI_API_KEY` environment variable."
+    )
+    raise Exception(f"Missing module: {e}")
+
+
+class BaseWhisperSTTService(SegmentedSTTService):
```
aconchillo, line 26: maybe rename file to `base_whisper.py`? we usually do that in all the other base classes.

aconchillo (approved):

aconchillo:
Awesome! LGTM! One minor comment.

PR #1184, opened 2025-02-08
fix: incorrectly changed the base type in my last pull request for  L…

TheCodingLand:
Fix base class of LocalAudioTransport

### DIFF
--- a/src/pipecat/transports/local/audio.py
+++ b/src/pipecat/transports/local/audio.py
@@ -14,7 +14,8 @@ from pipecat.frames.frames import InputAudioRawFrame, StartFrame
 from pipecat.processors.frame_processor import FrameProcessor
 from pipecat.transports.base_input import BaseInputTransport
 from pipecat.transports.base_output import BaseOutputTransport
-from pipecat.transports.base_transport import BaseTransport, TransportParams
+from pipecat.transports.base_transport import TransportParams, BaseTransport
+
 
 try:
     import pyaudio
@@ -116,7 +117,7 @@ class LocalAudioOutputTransport(BaseOutputTransport):
             )
 
 
-class LocalAudioTransport(LocalTransportParams):
+class LocalAudioTransport(BaseTransport):
     def __init__(self, params: LocalTransportParams):
         super().__init__()
         self._params = params


markbackman (approved):
Approved, just need to fix linting. The import order needs to be alphabetical.

aconchillo:
Ooops,my bad. 🤦‍♂️

examples/foundational/01a-local-audio.py
```diff
@@ -26,7 +25,7 @@
 
 
 async def main():
-    transport = LocalAudioTransport(TransportParams(audio_out_enabled=True))
+    transport = LocalAudioTransport(LocalTransportParams(audio_out_enabled=True))
```
markbackman, line 28: I don't think this is necessary. LocalTransportParams is a subclass of TransportParams. The actual `audio_out_enabled` is in `TransportParams`, so it seems fine to keep it.

examples/foundational/01a-local-audio.py
```diff
@@ -41,7 +40,7 @@ async def say_something():
         await asyncio.sleep(1)
         await task.queue_frames([TTSSpeakFrame("Hello there, how is it going!"), EndFrame()])
 
-    runner = PipelineRunner()
+    runner = PipelineRunner(handle_sigint=False if sys.platform == "win32" else True)
```
markbackman, line 43: We can probably drop this too. In general, Pipecat doesn't support Windows. We probably need a higher level message than just something in this demo.

TheCodingLand:
```python
transport = LocalAudioTransport(TransportParams(audio_out_enabled=True))
```

Since the type of param LocalAudioTransport has changed, it's requried to use the new class, since LocalAudioTransport now reads input_device_index and output_device_index to setup pyaudio.

TransportParams doesn't have the default value.
I could manage the case but seems inconsistent with the rest of the code where params are subclassed by service taht requires specific parameters.

Any reason why pipecat doesnt support windows ?

examples/foundational/01a-local-audio.py
aconchillo: I think this change is correct.

examples/foundational/01a-local-audio.py
aconchillo: Currently, only `daily-python` doesn't really support Windows. I'd say everything else supports Windows, even though I haven't personally tried it.

examples/local-input-select-stt/requirements.txt
```diff
@@ -0,0 +1,8 @@
+--extra-index-url https://download.pytorch.org/whl/cu124
+torch==2.5.0+cu124
+torchvision 
+torchaudio
+pipecat[whisper, openai]
+textual==1.0.0
+pydantic-settings==2.7.1
+pyaudio==0.2.14
```
aconchillo, line 8: you can remove `pyaudio` and specify `local` option in `pipecat`

examples/local-input-select-stt/requirements.txt
```diff
@@ -0,0 +1,8 @@
+--extra-index-url https://download.pytorch.org/whl/cu124
+torch==2.5.0+cu124
+torchvision 
+torchaudio
```
aconchillo, line 4: why do we need all these torch imports?

aconchillo:
@TheCodingLand This is awesome!

PR #1186, opened 2025-02-09
Add language support to WhisperSTTService

markbackman:

### DIFF
--- a/src/pipecat/services/whisper.py
+++ b/src/pipecat/services/whisper.py
@@ -15,6 +15,7 @@ from loguru import logger
 
 from pipecat.frames.frames import ErrorFrame, Frame, TranscriptionFrame
 from pipecat.services.ai_services import SegmentedSTTService
+from pipecat.transcriptions.language import Language
 from pipecat.utils.time import time_now_iso8601
 
 try:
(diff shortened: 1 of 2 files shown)


src/pipecat/services/whisper.py
```diff
@@ -26,18 +27,219 @@
...
+    Args:
+        language: A Language enum value representing the input language.
+
+    Returns:
+        str or None: The corresponding Whisper language code, or None if not supported.
+
+    Note:
+        Only includes languages officially supported by Whisper.
+    """
+    language_map = {
```
filipi87, line 67: It seems that all the languages we need to pass to Whisper only support the first two-letter abbreviation and do not support the country.
Couldn't we just create a generic function to extract the first two characters instead of using this map ?

filipi87 (approved):
I left a comment regarding if we could improve how we are converting the language.
But either way it all looks good.

src/pipecat/services/whisper.py
filipi87: Makes sense. I like the idea of being explicitly about which languages  Whisper actually supports. 👍

filipi87: Makes sense. I like the idea of being explicitly about which languages  Whisper actually supports. 👍

Issue #1192, opened 2025-02-10
ElevenLabs disconnect websocket connection seems to be buggy.


Vaibhav-Lodha:
Continosuly getting `ElevenLabsFlashTTSService error closing websocket: no close frame received or sent` error even though pipeline has ended with `EndFrame`

<img width="531" alt="Image" src="https://github.com/user-attachments/assets/c15093ce-330d-47b4-bbd1-ab57e51eee78" />

Vaibhav-Lodha:
More logs around the same - 

<img width="717" alt="Image" src="https://github.com/user-attachments/assets/20b04284-1521-4a8f-9be5-101fd6f400c5" />

aconchillo:
We did some changes in this PR https://github.com/pipecat-ai/pipecat/pull/1222. I have tested `EndFrame` with ElevenLabs and Cartesia and I don't see any issues. @Vaibhav-Lodha If this was reproducible, any chance you could try `main`?

PR #1194, opened 2025-02-10
fixing disconnect issue

Vaibhav159:
Fixed an issue[#1192] in 11labs where we are trying to reconnect/disconnect the websocket connection even when the connection is already closed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -42,6 +42,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   `RTVIBotLLMProcessor`) are now deprecated, instantiate an `RTVIObserver`
   instead.
 
+### Fixed
+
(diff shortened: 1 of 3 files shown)


Vaibhav159:
@aconchillo @markbackman have we review this pr, i believe this will solve the disconnect issue.

Vaibhav-Lodha:
Do refer the second SS of the issue, i also believe once we have seen endframe, we should not attempt to reconnect

Vaibhav159:
Do refer the second SS of the issue, i also believe once we have seen endframe, we should not attempt to reconnect

aconchillo (approved):

aconchillo:
@Vaibhav159 Thank you, LGTM. Needs to fix conflicts.

Vaibhav159:
@aconchillo resolved the conflicts.

PR #1195, opened 2025-02-10
chore(ai_services): raise exception if api_key None or empty

Vaibhav159:


Vaibhav159:
@aconchillo @markbackman this is up for review, but want to get your thoughts on one thing, i feel setting self._api_key in each service is a bit overkill, since we can put it in core base classes like `LLMService`/ `TTSService` and once `init` is done, can safely refer the same at all places.

Vaibhav159:
@aconchillo @markbackman did we get a chance to look at this pr?

markbackman:
Apologies for not looking at this. I'm closing since it has so many conflicts. I'm torn as to whether this is needed or not. Perhaps it would make for a better developer experience, but the warnings from the services are pretty clear when a key isn't provided.

PR #1196, opened 2025-02-10
AudioBufferProcessor: handle continuous and intermittent user audio

### LINKED ISSUE #1172: Audio alignment issues with AudioBufferProcessor
I'm using the [AudioBufferProcessor](https://github.com/pipecat-ai/pipecat/blob/main/src/pipecat/processors/audio/audio_buffer_processor.py) (slightly modified) to record both user and bot audio, but having issues with audio synchronization. It seems like the culprit is the `quiet_time < 0.1` threshold and the timing logic for adding silence. My user audio frames are streamed constantly and much faster than my bot audio (but with less bytes per frame) and it seems like after a little while the bot audio and user audio no longer lines up correctly. 

Has anyone faced this issue before or have any workarounds/other ideas?

Thanks!

aconchillo:

We need two different strategies to record audio properly depending if user audio is continuous or intermittent.

Fixes #1172

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -44,6 +44,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- Fixed an issue that was causing `AudioBufferProcessor` to not record
+  synchronized audio.
+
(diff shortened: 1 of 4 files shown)


markbackman (approved):
LGTM!

Issue #1001, opened 2025-01-16
Add GoogleSTTService


markbackman:
There have been a number of requests for this in Discord. I'm creating an issue to track it.

The Pipecat team hasn't started work on this and it would be behind a number of other tasks. If anyone in the community wants to take this one, we're happy to support you!

CodeNameR:
Is there any update on this? I can take a stab, if anyone else hasn't started working on it

markbackman:
I started working on this today. I should have a PR soon.

markbackman:
Merged!

PR #1197, opened 2025-02-10
Add GoogleSTTService

markbackman:

Finally getting around to adding this: https://github.com/pipecat-ai/pipecat/issues/1001.


src/pipecat/services/google/google.py
```diff
@@ -8,6 +8,11 @@
 import base64
 import io
 import json
+import os
+
+# Suppress gRPC fork warnings
+os.environ["GRPC_ENABLE_FORK_SUPPORT"] = "false"
```
markbackman, line 14: Suppressing `Other threads are currently calling into gRPC, skipping fork() handlers` warnings, which are unavoidable and benign, as far as I have read.

src/pipecat/services/google/google.py
```diff
@@ -1097,3 +1408,380 @@ async def run_image_gen(self, prompt: str) -> AsyncGenerator[Frame, None]:
...
+
+        # Create recognition config with explicit audio format
+        self._config = cloud_speech.StreamingRecognitionConfig(
+            config=cloud_speech.RecognitionConfig(
+                explicit_decoding_config=cloud_speech.ExplicitDecodingConfig(
+                    encoding=cloud_speech.ExplicitDecodingConfig.AudioEncoding.LINEAR16,
+                    sample_rate_hertz=self.sample_rate,
+                    audio_channel_count=1,
+                ),
+                language_codes=[self._settings["language_code"]],
```
KaliCharan-V, line 1669: @markbackman Can we have the support to pass multiple languages? GoogleSTT supports it, same with Azure too(Azure has a limit of 3).

src/pipecat/services/google/google.py
```diff
@@ -1097,3 +1408,380 @@ async def run_image_gen(self, prompt: str) -> AsyncGenerator[Frame, None]:
...
+                json_account_info = json.load(f)
+                self._project_id = json_account_info.get("project_id")
+            creds = service_account.Credentials.from_service_account_file(credentials_path)
+        else:
+            raise ValueError("Either credentials or credentials_path must be provided")
+
+        if not self._project_id:
+            raise ValueError("Project ID not found in credentials")
+
+        self._client = speech_v2.SpeechAsyncClient(credentials=creds)
```
KaliCharan-V, line 1500: https://cloud.google.com/speech-to-text/docs/samples/speech-multi-region-client
Don't we need to pass client_options like this?
            client_options=ClientOptions(api_endpoint="us-speech.googleapis.com"),

markbackman: @KaliCharan-V, yup! I just stumbled on this too as I was testing. Just added it.

src/pipecat/services/google/google.py
markbackman: Good idea. Just added it. It will now take a str or List of Languages.

examples/foundational/07n-interruptible-google.py
```diff
@@ -46,14 +44,16 @@ async def main():
             ),
         )
 
-        stt = DeepgramSTTService(api_key=os.getenv("DEEPGRAM_API_KEY"))
+        stt = GoogleSTTService(
```
filipi87, line 47: 🎉

src/pipecat/services/google/google.py
```diff
@@ -51,10 +62,13 @@
     import google.ai.generativelanguage as glm
     import google.generativeai as gai
     from google import genai
-    from google.cloud import texttospeech_v1
+    from google.api_core.client_options import ClientOptions
+    from google.cloud import speech_v2, texttospeech_v1
```
filipi87, line 66: I was wondering, we already have so many services integrated with Pipecat. Would it be worth adding a link to the original documentation for each new service we integrate ? 

I'm not sure where the best place to include it would be, but it would make it much easier to see which documentation we used as a reference when developing the integration.

src/pipecat/services/google/google.py
filipi87, line 1: I know we're already doing this for most services, but I'm adding it here for future discussion.
Should we split each service into its own file ? For example, GoogleLLMService, GoogleSTTService, and GoogleTTSService ?
Maybe we could create a directory for each service provider and organize each service into its own file within that directory.

markbackman: Yes! I mentioned the same thing to Aleix in a DM. I didn't want to do it in this PR, but I think we need to break up these services into smaller files for better maintainability. I can work on that soon.

src/pipecat/services/google/google.py
markbackman: Good point. I've added some links to the service docs when they're useful, but we should probably do that more frequently. Unfortunately, Google's docs (and some others) are fairly difficult to follow.

src/pipecat/services/google/google.py
```diff
@@ -1097,3 +1410,459 @@ async def run_image_gen(self, prompt: str) -> AsyncGenerator[Frame, None]:
...
+            languages: List of languages for recognition. First language is primary.
+        """
+        logger.info(f"Switching STT languages to: {languages}")
+        self._settings["language_codes"] = [
+            self.language_to_service_language(lang) for lang in languages
+        ]
+        # Recreate stream with new languages
+        if self._streaming_task:
+            await self._disconnect()
+            await self._connect()
```
filipi87, line 1578: Nit: We perform this check multiple times. Maybe centralize it in a new method, like `_reconnect_if_needed` or something similar.

src/pipecat/services/google/google.py
```diff
@@ -1097,3 +1410,459 @@ async def run_image_gen(self, prompt: str) -> AsyncGenerator[Frame, None]:
...
+            enable_word_confidence: Enable/disable word confidence scores.
+            enable_interim_results: Enable/disable interim results.
+            enable_voice_activity_events: Enable/disable voice activity detection.
+            location: New Google Cloud location.
+
+        Note:
+            Changes that affect the streaming configuration will cause
+            the stream to be reconnected.
+        """
+        needs_reconnect = False
```
filipi87, line 1635: Do we need this ? It feels like if the user has invoked this method and changed any of the options, we would need to check whether a reconnection is necessary at the end anyway.

src/pipecat/services/google/google.py
```diff
@@ -1097,3 +1410,459 @@ async def run_image_gen(self, prompt: str) -> AsyncGenerator[Frame, None]:
...
+            )
+
+            while True:
+                try:
+                    audio_data = await self._request_queue.get()
+                    if audio_data is None:  # Sentinel value to stop
+                        break
+
+                    # Check streaming limit
+                    if (int(time.time() * 1000) - self._stream_start_time) > self.STREAMING_LIMIT:
```
filipi87, line 1770: Cool, very clean approach about how to handle this limit. =)

src/pipecat/services/google/google.py
filipi87: Yeah, that's why I was wondering about creating a dedicated place, maybe a specific markdown file, with all the links to the documentations we've used as a reference for integrating the services. This could be very helpful in the future for fixing issues or simply as a reference to understand why we developed things that way.

src/pipecat/services/google/google.py
filipi87: Yep, that would be great.

filipi87 (approved):
All looks good. Pretty nice addition. 🚀

src/pipecat/services/google/google.py
markbackman: Google had a nice example of how to do this!
https://cloud.google.com/speech-to-text/docs/transcribe-streaming-audio#endless-streaming

I'm going to mention that in the comments, so it's not lost.

src/pipecat/services/google/google.py
markbackman: Good idea!

PR #1198, opened 2025-02-11
Add prompt and temperature args to OpenAI and Groq hosted Whisper STT…

markbackman:
… services

Community request.

It's unclear how the `prompt` input will be used. I can imagine wanting to feed in the previous context to it. We could expose a method for that, but for now, I think this is a good start. Let's see how people want to use it first.

### DIFF
--- a/src/pipecat/services/groq.py
+++ b/src/pipecat/services/groq.py
@@ -55,7 +55,7 @@ class GroqSTTService(BaseWhisperSTTService):
         base_url: API base URL. Defaults to "https://api.groq.com/openai/v1".
         language: Language of the audio input. Defaults to English.
         prompt: Optional text to guide the model's style or continue a previous segment.
-        temperature: Sampling temperature between 0 and 1. Defaults to 0.0.
+        temperature: Optional sampling temperature between 0 and 1. Defaults to 0.0.
         **kwargs: Additional arguments passed to BaseWhisperSTTService.
     """
 
--- a/src/pipecat/services/groq.py
+++ b/src/pipecat/services/groq.py
@@ -67,7 +67,7 @@ class GroqSTTService(BaseWhisperSTTService):
         base_url: str = "https://api.groq.com/openai/v1",
         language: Optional[Language] = Language.EN,
         prompt: Optional[str] = None,
-        temperature: float = 0.0,
+        temperature: Optional[float] = None,
         **kwargs,
     ):
         super().__init__(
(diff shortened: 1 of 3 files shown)


golbin:
Great! ✨

For reference, Whisper Prompt can be used to:

- Describe the situation (e.g., a meeting, a casual call, a call center conversation)
- Include proper nouns
- Reference previous messages to provide context

If it can automatically refer to previous messages, that would be great.

In fact, Groq or Fal Whisper is just as fast as Deepgram, even without WebSocket streaming. However, its capabilities—such as multilingual support and automatic typo correction—are more advanced than other STT solutions. So, it’s a significant leap forward.

src/pipecat/services/groq.py
```diff
@@ -64,17 +66,35 @@ def __init__(
...
+            "file": ("audio.wav", audio, "audio/wav"),
+            "model": self.model_name,
+            "response_format": "json",
+            "language": self._language,
+        }
+
+        if self._prompt is not None:
+            kwargs["prompt"] = self._prompt
+
+        if self._temperature is not None:
```
aconchillo, line 97: `_temperature` is initialized to 0.0. Should it be optional in the constructor? If it's 0.0 do we want to actually use it?

src/pipecat/services/openai.py
```diff
@@ -418,17 +420,37 @@ def __init__(
...
+        kwargs = {
+            "file": ("audio.wav", audio, "audio/wav"),
+            "model": self.model_name,
+            "language": self._language,
+        }
+
+        if self._prompt is not None:
+            kwargs["prompt"] = self._prompt
+
+        if self._temperature is not None:
```
aconchillo, line 450: Same here.

markbackman:
@aconchillo makes sense. Just pushed an update.

src/pipecat/services/groq.py
```diff
@@ -54,6 +54,8 @@ class GroqSTTService(BaseWhisperSTTService):
         api_key: Groq API key. Defaults to None.
         base_url: API base URL. Defaults to "https://api.groq.com/openai/v1".
         language: Language of the audio input. Defaults to English.
+        prompt: Optional text to guide the model's style or continue a previous segment.
+        temperature: Sampling temperature between 0 and 1. Defaults to 0.0.
```
filipi87, line 58: Maybe fix the description to mention that it now is optional ?

src/pipecat/services/openai.py
```diff
@@ -408,6 +408,8 @@ class OpenAISTTService(BaseWhisperSTTService):
         api_key: OpenAI API key. Defaults to None.
         base_url: API base URL. Defaults to None.
         language: Language of the audio input. Defaults to English.
+        prompt: Optional text to guide the model's style or continue a previous segment.
+        temperature: Sampling temperature between 0 and 1. Defaults to 0.0.
```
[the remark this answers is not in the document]
markbackman: Updated.

src/pipecat/services/groq.py
markbackman: Updated.

PR #1208, opened 2025-02-11
Add new STTMuteStrategy: MUTE_UNTIL_FIRST_BOT_COMPLETE

markbackman:

This came up in Discord. Using the `FIRST_SPEECH` strategy, it's possible to interrupt the bot between the TTS generation and the `BotStartedSpeakingFrame`.  This is a race condition, though it's more possible to happen with slow TTS services. This new strategy is for apps where the bot will speak first and you want no user interruption between join and the bot finishing its output.

`FIRST_SPEECH` remains and will work more flexibly, if the user needs to speak first, but subsequently, you don't want the bot's first line interrupted.


src/pipecat/processors/filters/stt_mute_filter.py
```diff
@@ -37,16 +37,18 @@ class STTMuteStrategy(Enum):
...
         ALWAYS: Mute during all bot speech
         CUSTOM: Allow custom logic via callback
     """
 
-    FIRST_SPEECH = "first_speech"  # Mute only during first bot speech
-    FUNCTION_CALL = "function_call"  # Mute during function calls
-    ALWAYS = "always"  # Mute during all bot speech
-    CUSTOM = "custom"  # Allow custom logic via callback
+    FIRST_SPEECH = "first_speech"
+    MUTE_UNTIL_FIRST_BOT_COMPLETE = "mute_until_first_bot_complete"
```
markbackman, line 48: I went verbose on the name so it is less confusing given the existence of `FIRST_SPEECH`.

src/pipecat/processors/filters/stt_mute_filter.py
```diff
@@ -124,6 +139,10 @@ async def _should_mute(self) -> bool:
                         self._first_speech_handled = True
```
filipi87, line 139: Should it still remain here, now that we are changing this value inside BotStoppedSpeakingFrame ?

filipi87: Ok, I think I see the differences between them. 👍

filipi87 (approved):
LGTM!

PR #1209, opened 2025-02-12
reimplement LLM response aggregators

### LINKED ISSUE #984: Short utterances don't trigger the VAD
# Description
We've received a number of reports that short utterances like, "OK", "Yes, "No" aren't "heard" by the bot. The root cause for this issue is that the VAD is not triggered. The default value to detect speech is a `start_secs` duration of 0.2 seconds or longer.

As a quick fix, the VAD's `start_secs` value can be lowered to 0.15 or even 0.1 seconds. But, this may result in unintended consequences like interruptions triggering unexpectedly.

As a resolution to this issue, we want to experiment with different solutions to ensure that interruptions happen only with speech and that short utterances are transcribed and result in input to the LLM.

aconchillo:

This PR fixes multiple issue where user transcriptions where not being handled properly. It was possible for short utterances to not trigger VAD which would cause user transcriptions to be ignored. It was also possible for one or more transcriptions to be generated after VAD in which case they would also be ignored.

The implementation is much simpler than before and also simpler than other proposed solutions. Instead of trying to add logic for every single case, we differentiate 2 main cases:

1. Transcription in between `UserStartedSpeakingFrame` and `UserStoppedSpeakingFrame`.
2. Transcriptions than don't match (1).

For (2), we simply start a timer and push the aggregation if we need to. This solves all of the weird cases we were not handling before.

Fixes #984


aconchillo:
This is currently being tested.

src/pipecat/processors/aggregators/llm_response.py
```diff
@@ -28,121 +35,105 @@
...
 
-class LLMResponseAggregator(FrameProcessor):
+class BaseLLMResponseAggregator(FrameProcessor):
+    """This is the base class for all LLM response aggregators. These
+    aggregators process incoming frames and aggregate content until they are
+    ready to push the aggregation. In the case of a user, an aggregation might
+    be a full transcription received from the STT service.
+
+    The LLM response aggregators also keep a store (e.g. a message list or an
+    LLM context) of the current conversation, that is, it store the messages
```
markbackman, line 45:
```suggestion
    LLM context) of the current conversation, that is, it stores the messages
```

src/pipecat/processors/aggregators/llm_response.py
```diff
@@ -290,31 +198,219 @@ async def _push_aggregation(self):
...
-            accumulator_frame=TranscriptionFrame,
-            interim_accumulator_frame=InterimTranscriptionFrame,
-        )
+            self.reset()
+
+
+class LLMUserContextAggregator(LLMContextResponseAggregator):
+    """This is a user LLM aggregator that uses an LLM context to store the
+    conversation. It aggregates transcriptions from the STT service and it has
+    logic to handle multiple escenarios where transcriptions are received
```
markbackman, line 207:
```suggestion
    logic to handle multiple scenarios where transcriptions are received
```

src/pipecat/processors/aggregators/llm_response.py
```diff
@@ -290,31 +198,219 @@ async def _push_aggregation(self):
...
+        """
+        if not self._user_speaking:
+            diff_time = time.time() - self._last_user_speaking_time
+            if diff_time > self._bot_interruption_timeout:
+                await self.push_frame(BotInterruptionFrame(), FrameDirection.UPSTREAM)
+
+
+class LLMAssistantContextAggregator(LLMContextResponseAggregator):
+    """This is an assistant LLM aggregator that uses an LLM context to store the
+    conversation. It aggregates text frames that go between
```
markbackman, line 330:
```suggestion
    conversation. It aggregates text frames received between
```

aconchillo: Updated.

src/pipecat/processors/aggregators/llm_response.py
aconchillo: Updated.

src/pipecat/processors/aggregators/llm_response.py
aconchillo: Updated.

src/pipecat/processors/aggregators/llm_response.py
```diff
@@ -153,109 +144,27 @@ async def _push_aggregation(self):
...
-            messages=messages,
-            role="user",
-            start_frame=UserStartedSpeakingFrame,
-            end_frame=UserStoppedSpeakingFrame,
-            accumulator_frame=TranscriptionFrame,
-            interim_accumulator_frame=InterimTranscriptionFrame,
-        )
-
-
-class LLMFullResponseAggregator(FrameProcessor):
```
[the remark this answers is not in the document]
aconchillo: I don't see it being used anywhere, that's why I got rid of it. :thinking:

markbackman (approved):
LGTM! 🚀

PR #1210, opened 2025-02-12
CartesiaTTSService: reset context id when flushing audio

aconchillo:

We need to clear the context id when we are flushing audio.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -52,6 +52,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- Fixed a `CartesiaTTSService` issue that could cause audio overlapping in some
+  cases.
+
 - Fixed an issue that was causing `AudioBufferProcessor` to not record
   synchronized audio.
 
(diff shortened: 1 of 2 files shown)


golbin:
I’m not exactly sure, but would it be better to maintain the context ID? As far as I know, keeping the previous context makes the conversation sound more natural.

markbackman:
> I’m not exactly sure, but would it be better to maintain the context ID? As far as I know, keeping the previous context makes the conversation sound more natural.

@golbin the `context_id` is a specific feature related to Cartesia: https://docs.cartesia.ai/api-reference/tts/working-with-web-sockets/contexts. Pipecat makes use of it to both improve Cartesia output, but also track a bot's turn. This is how interruption handling works in the context of a websocket connection. There is nothing lost in terms of functionality with what @aconchillo is proposing. This is a bug fix for overlapping audio.

markbackman:
@aconchillo this doesn't do what we want it to. It causes the audio to get prematurely truncated.

aconchillo:
Yes, this was late night nonsense. Closing, sorry for the noise.

PR #1211, opened 2025-02-12
Unified format for function calling

filipi87:
Unified format for function calling.

### DIFF
--- a/examples/unified-format-function-calling/base_function_calling.py
+++ b/examples/unified-format-function-calling/base_function_calling.py
@@ -1,0 +1,134 @@
+#
+# Copyright (c) 2024–2025, Daily
+#
+# SPDX-License-Identifier: BSD 2-Clause License
+#
+
+import os
+import sys
+
+import aiohttp
+from dotenv import load_dotenv
+from loguru import logger
+from runner import configure
+
+from pipecat.adapters.schemas.function_schema import FunctionSchema
+from pipecat.adapters.schemas.tools_schema import ToolsSchema
+from pipecat.audio.vad.silero import SileroVADAnalyzer
+from pipecat.frames.frames import TTSSpeakFrame
+from pipecat.pipeline.pipeline import Pipeline
+from pipecat.pipeline.runner import PipelineRunner
+from pipecat.pipeline.task import PipelineParams, PipelineTask
+from pipecat.processors.aggregators.openai_llm_context import OpenAILLMContext
+from pipecat.services.ai_services import LLMService
+from pipecat.services.cartesia import CartesiaTTSService
+from pipecat.transports.services.daily import DailyParams, DailyTransport
+
+logger.remove(0)
+logger.add(sys.stderr, level="DEBUG")
+
+load_dotenv(override=True)
+
+
+async def start_fetch_weather(function_name, llm, context):
+    """Push a frame to the LLM; this is handy when the LLM response might take a while."""
+    await llm.push_frame(TTSSpeakFrame("Let me check on that."))
+    logger.debug(f"Starting fetch_weather_from_api with function_name: {function_name}")
+
+
+async def fetch_weather_from_api(function_name, tool_call_id, args, llm, context, result_callback):
+    await result_callback({"conditions": "nice", "temperature": "75"})
+
+
+class WeatherBot:
+    """Generic base class for setting up and running an LLM-powered bot."""
+
+    def __init__(self, llm: LLMService):
+        """Initialize the base handler with a specific LLM."""
+        self.llm = llm
+
+    async def run(self):
+        """Set up and start the processing pipeline."""
+        async with aiohttp.ClientSession() as session:
+            (room_url, token) = await configure(session)
+
+            transport = DailyTransport(
+                room_url,
+                token,
+                "Respond bot",
+                DailyParams(
+                    audio_out_enabled=True,
+                    transcription_enabled=True,
+                    vad_enabled=True,
+                    vad_analyzer=SileroVADAnalyzer(),
+                ),
+            )
+
+            tts = CartesiaTTSService(
+                api_key=os.getenv("CARTESIA_API_KEY"),
+                voice_id="79a125e8-cd45-4c13-8a67-188112f4dd22",  # British Lady
+            )
+
+            # Register a function_name of None to get all functions
+            # sent to the same callback with an additional function_name parameter.
+            self.llm.register_function(
+                None, fetch_weather_from_api, start_callback=start_fetch_weather
+            )
+
+            weather_function = FunctionSchema(
+                name="get_current_weather",
+                description="Get the current weather",
+                properties={
+                    "location": {
+                        "type": "string",
+                        "description": "The city and state, e.g. San Francisco, CA",
+                    },
+                    "format": {
+                        "type": "string",
+                        "enum": ["celsius", "fahrenheit"],
+                        "description": "The temperature unit to use. Infer this from the user's location.",
+                    },
+                },
+                required=["location"],
+            )
+            tools = ToolsSchema(standard_tools=[weather_function])
+
+            messages = [
+                {
+                    "role": "system",
+                    "content": "You are a helpful assistant who can report the weather in any location in the universe. Respond concisely. Your response will be turned into speech so use only simple words and punctuation.",
+                },
+                {"role": "user", "content": " Start the conversation by introducing yourself."},
+            ]
+
+            context = OpenAILLMContext(messages, tools)
+            context_aggregator = self.llm.create_context_aggregator(context)
+
+            pipeline = Pipeline(
+                [
+                    transport.input(),
+                    context_aggregator.user(),
+                    self.llm,
+                    tts,
+                    transport.output(),
+                    context_aggregator.assistant(),
+                ]
+            )
+
+            task = PipelineTask(
+                pipeline,
+                params=PipelineParams(
+                    allow_interruptions=True,
+                    enable_metrics=True,
+                    enable_usage_metrics=True,
+                    report_only_initial_ttfb=True,
+                ),
+            )
+
+            @transport.event_handler("on_first_participant_joined")
+            async def on_first_participant_joined(transport, participant):
+                await transport.capture_participant_transcription(participant["id"])
+                await task.queue_frames([context_aggregator.user().get_context_frame()])
+
+            runner = PipelineRunner()
+            await runner.run(task)
--- a/examples/unified-format-function-calling/multimodal_base_function_calling.py
+++ b/examples/unified-format-function-calling/multimodal_base_function_calling.py
@@ -1,0 +1,126 @@
+#
+# Copyright (c) 2024–2025, Daily
+#
+# SPDX-License-Identifier: BSD 2-Clause License
+#
+
+import sys
+from typing import List
+
+import aiohttp
+from dotenv import load_dotenv
+from loguru import logger
+from runner import configure
+
+from pipecat.adapters.schemas.function_schema import FunctionSchema
+from pipecat.adapters.schemas.tools_schema import ToolsSchema
+from pipecat.audio.vad.silero import SileroVADAnalyzer
+from pipecat.frames.frames import TTSSpeakFrame
+from pipecat.pipeline.pipeline import Pipeline
+from pipecat.pipeline.runner import PipelineRunner
+from pipecat.pipeline.task import PipelineParams, PipelineTask
+from pipecat.processors.aggregators.openai_llm_context import OpenAILLMContext
+from pipecat.services.ai_services import LLMService
+from pipecat.transports.services.daily import DailyParams, DailyTransport
+
+logger.remove(0)
+logger.add(sys.stderr, level="DEBUG")
+
+load_dotenv(override=True)
+
+
+async def start_fetch_weather(function_name, llm, context):
+    """Push a frame to the LLM; this is handy when the LLM response might take a while."""
+    await llm.push_frame(TTSSpeakFrame("Let me check on that."))
+    logger.debug(f"Starting fetch_weather_from_api with function_name: {function_name}")
+
+
+async def fetch_weather_from_api(function_name, tool_call_id, args, llm, context, result_callback):
+    await result_callback({"conditions": "nice", "temperature": "75"})
+
+
+class MultimodalWeatherBot:
+    """Generic base class for setting up and running an LLM-powered bot."""
+
+    def __init__(self, llm: LLMService):
+        """Initialize the base handler with a specific LLM."""
+        self.llm = llm
+
+    @staticmethod
+    def tools() -> ToolsSchema:
+        weather_function = FunctionSchema(
+            name="get_current_weather",
+            description="Get the current weather",
+            properties={
+                "location": {
+                    "type": "string",
+                    "description": "The city and state, e.g. San Francisco, CA",
+                },
+                "format": {
+                    "type": "string",
+                    "enum": ["celsius", "fahrenheit"],
+                    "description": "The temperature unit to use. Infer this from the user's location.",
+                },
+            },
+            required=["location"],
+        )
+        return ToolsSchema(standard_tools=[weather_function])
+
+    async def run(self):
+        """Set up and start the processing pipeline."""
+        async with aiohttp.ClientSession() as session:
+            (room_url, token) = await configure(session)
+
+            transport = DailyTransport(
+                room_url,
+                token,
+                "Respond bot",
+                DailyParams(
+                    audio_out_enabled=True,
+                    vad_enabled=True,
+                    vad_analyzer=SileroVADAnalyzer(),
+                    vad_audio_passthrough=True,
+                ),
+            )
+
+            # Register a function_name of None to get all functions
+            # sent to the same callback with an additional function_name parameter.
+            self.llm.register_function(
+                None, fetch_weather_from_api, start_callback=start_fetch_weather
+            )
+
+            messages = [
+                {
+                    "role": "system",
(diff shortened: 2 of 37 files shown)


tests/integration/integration_unified_function_calling.py
filipi87, line 1: Should we make this test to run on CI as well ? What about the API_KEYs ?

tests/integration/test_integration_unified_function_calling.py
filipi87, line 1: Should we make this test to run on CI as well ? What about the API_KEYs, which one should we use ?

src/pipecat/processors/aggregators/openai_llm_context.py
```diff
@@ -75,7 +84,9 @@ def messages(self) -> List[ChatCompletionMessageParam]:
         return self._messages
 
     @property
-    def tools(self) -> List[ChatCompletionToolParam] | NotGiven:
+    def tools(self) -> List[ChatCompletionToolParam] | NotGiven | List[Any]:
```
filipi87, line 87: Almost all the services invoke this method to retrieve the tools they are going to use. So, I have changed it here to make this change more transparent. 

However, another possible approach is to leave this method returning exactly the same thing as before and create a new one to return the tools already formatted for the specific LLM. 

Maybe something like `adapted_tools` or `llm_formatted_tools` ?

What do you think ?

src/pipecat/services/ai_services.py
```diff
@@ -135,10 +137,24 @@ async def process_generator(self, generator: AsyncGenerator[Frame | None, None])
...
     def __init__(self, **kwargs):
         super().__init__(**kwargs)
         self._callbacks = {}
         self._start_callbacks = {}
+        self._adapter = self.adapter_class()
+
+    def get_llm_adapter(self) -> BaseLLMAdapter:
+        return self._adapter
+
+    # TODO: should we create a specific type to return that they all should respect ?
```
filipi87, line 153: Any thoughts on this one ?

filipi87: The reason we created this method here is that all LLM services already have an implementation of `create_context_aggregator`. Essentially, wherever we have the `Context`, we invoke `create_context_aggregator`.

So, inside `create_context_aggregator`, we are now injecting the `adapter` into the context so that we can return the tools formatted for each specific LLM.

tests/integration/test_integration_unified_function_calling.py
```diff
@@ -0,0 +1,95 @@
...
+    # This will fail if an exception is raised
+    await llm.process_frame(frame, FrameDirection.DOWNSTREAM)
+
+    # Assert that the mock function was called
+    mock_fetch_weather.assert_called_once()
+
+
+@pytest.mark.skipif(os.getenv("OPENAI_API_KEY") is None, reason="OPENAI_API_KEY is not set")
+@pytest.mark.asyncio
+async def test_unified_function_calling_openai():
```
filipi87, line 74: So far, I have created only three integration tests, one for each different adapter. However, it would be easier to expand this to cover all the services, just as I did with the examples.

tests/test_function_calling_adapters.py
filipi87, line 1: Unit tests to make sure that we are parsing the functions as expected.

filipi87:
Ok, I believe this PR is now ready for review. 

With this PR, we now support a unified format for specifying function calling across all LLM services. For example:
```
weather_function = FunctionSchema(
    name="get_current_weather",
    description="Get the current weather",
    properties={
        "location": {
            "type": "string",
            "description": "The city and state, e.g. San Francisco, CA",
        },
        "format": {
            "type": "string",
            "enum": ["celsius", "fahrenheit"],
            "description": "The temperature unit to use. Infer this from the user's location.",
        },
    },
    required=["location"],
)
tools = ToolsSchema(standard_tools=[weather_function])
```

It also maintains compatibility with the previous method, ensuring that existing implementations aren't broken.

I have tested it with the following LLM services: Anthropic, Azure, Deepseek, Gemini (GoogleLLMService and GeminiMultimodalLiveLLMService), Grok, Groq, OpenAI (OpenAILLMService and OpenAIRealtimeBetaLLMService), OpenRouter, Together and Nim.

This PR also includes:
- Unit tests to validate that the adapters are performing the conversions as expected.
- Integration tests to verify that everything works with some of the LLMs.
- Examples for all the services mentioned above.

src/pipecat/services/ai_services.py
markbackman: Same thoughts as the previous comment. This seems like the right thing. Again, we should just expect that tools are automatically translated via the adapter, so injecting the adapter makes sense.

src/pipecat/services/anthropic.py
```diff
@@ -84,6 +85,9 @@ class AnthropicLLMService(LLMService):
     use `AsyncAnthropicBedrock` and `AsyncAnthropicVertex` clients
     """
 
+    # Overriding the default adapter to use the Gemini one.
```
markbackman, line 88: typo

```suggestion
    # Overriding the default adapter to use the Anthropic one.
```

markbackman (approved):
This is really great work!! 🙌 

Just one question, otherwise, this looks ready to go. I also tested out the examples and things work as expected (including some LLM providers like Cerebras not handling function calls well...)

One thought: should we update the foundational examples to also use the `FunctionSchema` for function calling? Since this is the standard way, we should align all examples to work in the same way.

examples/unified-format-function-calling/base_function_calling.py
```diff
@@ -0,0 +1,134 @@
...
+                    },
+                    "format": {
+                        "type": "string",
+                        "enum": ["celsius", "fahrenheit"],
+                        "description": "The temperature unit to use. Infer this from the user's location.",
+                    },
+                },
+                required=["location"],
+            )
+            tools = ToolsSchema(standard_tools=[weather_function])
```
aconchillo, line 94: Why in this example this doesn't go into a `static_method`?

examples/unified-format-function-calling/base_function_calling.py
```diff
@@ -0,0 +1,134 @@
...
+    """Push a frame to the LLM; this is handy when the LLM response might take a while."""
+    await llm.push_frame(TTSSpeakFrame("Let me check on that."))
+    logger.debug(f"Starting fetch_weather_from_api with function_name: {function_name}")
+
+
+async def fetch_weather_from_api(function_name, tool_call_id, args, llm, context, result_callback):
+    await result_callback({"conditions": "nice", "temperature": "75"})
+
+
+class BaseFunctionCallingHandler:
```
aconchillo, line 43: Should we call this `WeatherBot`? See below.

examples/unified-format-function-calling/multimodal_base_function_calling.py
```diff
@@ -0,0 +1,126 @@
...
+    """Push a frame to the LLM; this is handy when the LLM response might take a while."""
+    await llm.push_frame(TTSSpeakFrame("Let me check on that."))
+    logger.debug(f"Starting fetch_weather_from_api with function_name: {function_name}")
+
+
+async def fetch_weather_from_api(function_name, tool_call_id, args, llm, context, result_callback):
+    await result_callback({"conditions": "nice", "temperature": "75"})
+
+
+class MultimodalBaseFunctionCallingHandler:
```
aconchillo, line 42: And this one `MultimodalWeatherBot`. See my other comment.

examples/unified-format-function-calling/standard-function-calling-anthropic.py
```diff
@@ -0,0 +1,29 @@
...
+
+    def __init__(self):
+        llm = AnthropicLLMService(
+            api_key=os.getenv("ANTHROPIC_API_KEY"), model="claude-3-5-sonnet-20240620"
+        )
+        super().__init__(llm)
+
+
+if __name__ == "__main__":
+    asyncio.run(WeatherBot().run())
```
aconchillo, line 29: If we have the `WeatherBot` base classes, then we can simply do:

```
if __name__ == "__main__":
Why not simply:

```
if __name__ == "__main__":
        llm = AzureLLMService(
            api_key=os.getenv("AZURE_CHATGPT_API_KEY"),
            endpoint=os.getenv("AZURE_CHATGPT_ENDPOINT"),
            model=os.getenv("AZURE_CHATGPT_MODEL"),
        )
       asyncio.run(WeatherBot().run(llm))
```

 No need to define a `WeatherBot` in all the other files.

File: examples/unified-format-function-calling/base_function_calling.py (hunk 6 above)
Comment (aconchillo) [reply examples/unified-format-function-calling/base_function_calling.py:43]: Because this base class is only for a weather bot.

File: examples/unified-format-function-calling/standard-function-calling-gemini-multimodal.py [hunk 9]
```diff
@@ -0,0 +1,38 @@
...
+            voice_id="Puck",
+            transcribe_user_audio=True,
+            transcribe_model_audio=True,
+            tools=tools_def,
+        )
+        super().__init__(llm)
+
+
+if __name__ == "__main__":
+    asyncio.run(WeatherBot().run())
```
aconchillo, line 38: And here:

```
if __name__ == "__main__":
        search_tool = {"google_search": {}}
        tools_def = MultimodalBaseFunctionCallingHandler.tools()
        tools_def.custom_tools = {AdapterType.GEMINI: [search_tool]}

        llm = GeminiMultimodalLiveLLMService(
            voice_id="Puck",
            transcribe_user_audio=True,
            transcribe_model_audio=True,
            tools=tools_def,
        )
        asyncio.run(MultiModalWeatherBot().run(llm))
```

src/pipecat/adapters/schemas/function_schema.py
```diff
@@ -0,0 +1,39 @@
...
+
+        :param name: Name of the function.
+        :param description: Description of the function.
+        :param properties: Dictionary defining properties types and descriptions.
+        :param required: List of required parameters.
+        """
+        self.name = name
+        self.description = description
+        self.properties = properties
+        self.required = required
```
aconchillo, line 24: It seems these fields should be private (need an `_` suffix). If we want some of them to be accessible from outside we should create a `@property` so they are read-only.

src/pipecat/adapters/schemas/tools_schema.py
```diff
@@ -0,0 +1,31 @@
...
+    ) -> None:
+        """
+        A schema for tools that includes both standardized function schemas
+        and custom tools that do not follow the FunctionSchema format.
+
+        :param standard_tools: List of tools following FunctionSchema.
+        :param custom_tools: List of tools in a custom format (e.g., search_tool).
+        """
+        self.standard_tools = standard_tools
+        self.custom_tools = custom_tools
```
aconchillo, line 31: Same. If these are read-only, make them private and create properties.

src/pipecat/services/ai_services.py
```diff
@@ -135,10 +137,24 @@ async def process_generator(self, generator: AsyncGenerator[Frame | None, None])
...
 
+    # OpenAILLMAdapter is used as the default adapter since it aligns with most LLM implementations.
+    # However, subclasses should override this with a more specific adapter when necessary.
+    adapter_class: Type[BaseLLMAdapter] = OpenAILLMAdapter
+
     def __init__(self, **kwargs):
         super().__init__(**kwargs)
         self._callbacks = {}
         self._start_callbacks = {}
+        self._adapter = self.adapter_class()
```
aconchillo, line 148: This is assuming the adapter doesn't have any arguments. But what if it has in the future? Why not passing it to the constructor instead?

```
    def __init__(self, *, adapter: BaseLLMAdapter, **kwargs):
        super().__init__(**kwargs)
        self._callbacks = {}
        self._start_callbacks = {}
        self._adapter = adapter
```

aconchillo: In which case this won't be needed:

```
    adapter_class: Type[BaseLLMAdapter] = OpenAILLMAdapter
```

src/pipecat/services/anthropic.py
```diff
@@ -84,6 +85,9 @@ class AnthropicLLMService(LLMService):
...
+    adapter_class = AnthropicLLMAdapter
```
aconchillo, line 89: Then instead of this we would do:

```
super().__init__(adapter = AnthropicLLMAdapter(), **kwargs)
```

src/pipecat/services/ai_services.py
```diff
@@ -135,10 +137,24 @@ async def process_generator(self, generator: AsyncGenerator[Frame | None, None])
...
+
+    def get_llm_adapter(self) -> BaseLLMAdapter:
+        return self._adapter
```
aconchillo, line 151: Maybe we could us a property instead

```
@property
def llm_adapter(self) - > BaseLLMAdapter:
   return self._adapter
```

src/pipecat/services/google/google.py
```diff
@@ -940,6 +942,9 @@ class GoogleLLMService(LLMService):
     franca for all LLM services, so that it is easy to switch between different LLMs.
     """
 
+    # Overriding the default adapter to use the Gemini one.
+    adapter_class = GeminiLLMAdapter
```
aconchillo, line 946: And here:

```
super().__init__(adapter=GeminiLLMAdapter(), **kwargs)
```

src/pipecat/services/ai_services.py
filipi87: Great. Thank you!

src/pipecat/services/gemini_multimodal_live/gemini.py
```diff
@@ -435,7 +440,7 @@ async def _connect(self):
                 )
             if self._tools:
                 logger.debug(f"Gemini is configuring to use tools{self._tools}")
-                config.setup.tools = self._tools
+                config.setup.tools = self.get_llm_adapter().from_standard_tools(self._tools)
```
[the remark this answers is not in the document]
filipi87: We may receive the `tools` inside the `GeminiMultimodalLiveLLMService` constructor instead of through the context.

So, we are basically implementing it here to ensure that our tools are in the right format, even if we have received them through the constructor.

examples/unified-format-function-calling/base_function_calling.py
filipi87: Hi @aconchillo, I'm not sure I understood the question. Do you mean transforming everything into a static method instead of having the class `BaseFunctionCallingHandler` ?

examples/unified-format-function-calling/multimodal_base_function_calling.py
filipi87: Done!

examples/unified-format-function-calling/standard-function-calling-anthropic.py
filipi87: I renamed all the examples to include the name of the service.

I thought this would make it more didactic, but I can refactor it to match what you mentioned above if you believe that’s the best way to proceed.

```
class AnthropicWeatherBot(WeatherBot):
    """Main class defining the LLM and passing it to the base handler."""

    def __init__(self):
        llm = AnthropicLLMService(
            api_key=os.getenv("ANTHROPIC_API_KEY"), model="claude-3-5-sonnet-20240620"
        )
        super().__init__(llm)


if __name__ == "__main__":
    asyncio.run(AnthropicWeatherBot().run())
```

examples/unified-format-function-calling/standard-function-calling-gemini-multimodal.py
filipi87: Same as above. Let me know what do you think:

```
class GeminiMultimodalWeatherBot(MultimodalWeatherBot):
    """Main class defining the LLM and passing it to the base handler."""

    def __init__(self):
        search_tool = {"google_search": {}}
        tools_def = MultimodalWeatherBot.tools()
        tools_def.custom_tools = {AdapterType.GEMINI: [search_tool]}

        llm = GeminiMultimodalLiveLLMService(
            voice_id="Puck",
            transcribe_user_audio=True,
            transcribe_model_audio=True,
            tools=tools_def,
        )
        super().__init__(llm)


if __name__ == "__main__":
    asyncio.run(GeminiMultimodalWeatherBot().run())

```

src/pipecat/adapters/schemas/function_schema.py
filipi87: Done.

src/pipecat/adapters/schemas/tools_schema.py
filipi87: Done.

src/pipecat/services/ai_services.py
filipi87: The goal of the adapter is to adapt the function calling (and possibly messages in the future) for a specific service. So, it should be as simple as possible and have a single, consistent behavior. That’s why I didn’t think we would need any arguments.

I also didn’t consider the possibility of needing to dynamically change the adapter class at runtime for each instance.

That’s why I chose to use class attributes, as this approach provides a default behavior that can be overridden via subclassing, while keeping the constructor signature clean.

Let me know what you think, if it makes sense. Otherwise I can refactor to implement as you suggested above. 🙂

filipi87:
> One thought: should we update the foundational examples to also use the FunctionSchema for function calling? Since this is the standard way, we should align all examples to work in the same way.

@markbackman , does it make sense to update, or just remove all the examples where the single purpose was to show the function calling working ? Anyway, I think we can do one way or another in a follow up PR.

filipi87:
Hi @aconchillo, @markbackman, I’m going to move forward and merge this PR to prevent new conflicts. 

But let me know your thoughts on the explanations I provided above. I'm happy to push some refactors if needed to change the adapter for the constructor if you think that’s the best option.

Thank you all for the great reviews.

PR #1213, opened 2025-02-12
BaseOutputTransport: use specific VAD stop secs for the bot

aconchillo:

This PR fixes an issue that was causing `BotStoppedSpeakingFrame` to be generated too late. This could then cause issues unblocking `STTMuteFilter` later than desired.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -66,6 +66,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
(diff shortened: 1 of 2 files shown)


markbackman (approved):
LGTM! I tested a bunch of services with a frame observer. The only notable thing was OpenAITTSService. They appeared to be having degraded service during my testing, which caused some unexpected BotStoppedSpeakingFrames to occur. But, I believe that's unexpected. All others passed with flying colors. We can tune the timing if needed. Thanks for the quick fix!

PR #1214, opened 2025-02-13
Improve STTMuteFilter, add tests

markbackman:

This PR:
- Changes `STTMuteFrame` from `ControlFrame` to `SystemFrame`. We want the STTMuteFrame to not be queued.
- With changing `STTMuteFrame` to `ControlFrame`, the frame processing ordering needs to change. This PR ensures that the frames are correctly ordered.
- Confirmed ordering via new unit tests: There is a missing test for the `FUNCTION_CALL` strategy due to more test utils that need to be added. I listed that as a TODO.

I don't think any of this warrants a changelog entry. Happy to add one if it helps.

### DIFF
--- a/src/pipecat/processors/filters/stt_mute_filter.py
+++ b/src/pipecat/processors/filters/stt_mute_filter.py
@@ -155,24 +155,25 @@ class STTMuteFilter(FrameProcessor):
         """Processes incoming frames and manages muting state."""
         await super().process_frame(frame, direction)
 
-        # Handle function call state changes
+        # First determine if we need to change mute state
+        should_mute = None
+
         if isinstance(frame, FunctionCallInProgressFrame):
             self._function_call_in_progress = True
-            await self._handle_mute_state(await self._should_mute())
+            should_mute = await self._should_mute()
         elif isinstance(frame, FunctionCallResultFrame):
             self._function_call_in_progress = False
-            await self._handle_mute_state(await self._should_mute())
-        # Handle bot speaking state changes
+            should_mute = await self._should_mute()
         elif isinstance(frame, BotStartedSpeakingFrame):
             self._bot_is_speaking = True
-            await self._handle_mute_state(await self._should_mute())
+            should_mute = await self._should_mute()
         elif isinstance(frame, BotStoppedSpeakingFrame):
             self._bot_is_speaking = False
             if not self._first_speech_handled:
                 self._first_speech_handled = True
-            await self._handle_mute_state(await self._should_mute())
+            should_mute = await self._should_mute()
 
-        # Handle frame propagation
+        # Then push the original frame
         if isinstance(
             frame,
             (
@@ -190,3 +191,7 @@ class STTMuteFilter(FrameProcessor):
         else:
             # Pass all other frames through
             await self.push_frame(frame, direction)
+
+        # Finally handle mute state change if needed
+        if should_mute is not None and should_mute != self.is_muted:
+            await self._handle_mute_state(should_mute)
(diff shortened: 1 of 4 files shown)


tests/test_stt_mute_filter.py
```diff
@@ -0,0 +1,217 @@
...
+            STTMuteFrame,  # mute=False
+        ]
+
+        await run_test(
+            filter,
+            frames_to_send=frames_to_send,
+            expected_down_frames=expected_returned_frames,
+        )
+
+    # TODO: Revisit once we figure out how to test SystemFrames and DataFrames
```
aconchillo, line 89: I guess the `SleepFrame` will fix this or is there something else?

aconchillo (approved):

aconchillo:
LGTM!

src/pipecat/processors/filters/stt_mute_filter.py
```diff
@@ -155,24 +156,32 @@ async def process_frame(self, frame: Frame, direction: FrameDirection):
...
+            should_mute = await self._should_mute()
         elif isinstance(frame, BotStoppedSpeakingFrame):
             self._bot_is_speaking = False
             if not self._first_speech_handled:
                 self._first_speech_handled = True
-            await self._handle_mute_state(await self._should_mute())
+            should_mute = await self._should_mute()
 
-        # Handle frame propagation
+        # Then push the original frame
```
aconchillo, line 184: Should we include `InputAudioRawFrame` below as well?

tests/test_stt_mute_filter.py
markbackman: I'm not sure. I'm going to wait until those utils are merged, then I'll take another look.

src/pipecat/processors/filters/stt_mute_filter.py
markbackman: Technically, yes, but it's not really required. When the STTMuteFrame is true, we skip `self.run_stt` here: https://github.com/pipecat-ai/pipecat/blob/main/src/pipecat/services/ai_services.py#L528-L530. It doesn't look like the `InputRawAudioFrame` is pushed past the STTService, is that right?

I'm happy to follow your lead on what you think is best. I'll hold off on merging until you weigh in.

aconchillo: You can see `audio_passthrough=True` to an `STTService`. I think we should move these to lines https://github.com/pipecat-ai/pipecat/blob/main/src/pipecat/services/ai_services.py#L541-L542 inside `process_audio_frame()` and run them only if not muted.

```
    async def process_audio_frame(self, frame: AudioRawFrame):
        if self._muted:
            return

        await self.process_generator(self.run_stt(frame.audio))
        if self._audio_passthrough:
```

markbackman: Ok, that makes sense. I just made that change and tested. Things seem good. Just requested re-review just in case though. We'll merge this before today's release.

src/pipecat/processors/filters/stt_mute_filter.py
```diff
@@ -155,24 +156,32 @@ async def process_frame(self, frame: Frame, direction: FrameDirection):
...
+        # Handle initial state on StartFrame
+        if isinstance(frame, StartFrame):
+            # Check if we should start muted
+            should_mute = await self._should_mute()
+            if should_mute:
+                await self._handle_mute_state(True)
+
+        # First determine if we need to change mute state
+        should_mute = None
+
```
filipi87, line 168: Couldn't this just be like the other ifs below or what am I missing 😅? Like this:
```suggestion
        # First determine if we need to change mute state
        should_mute = None
        
        # Handle initial state on StartFrame
        if isinstance(frame, StartFrame):
            # Check if we should start muted
            should_mute = await self._should_mute()
```

filipi87 (approved):
LGTM!

PR #1215, opened 2025-02-13
Instant voice demo improvements - part 02

filipi87:

### DIFF
--- a/src/pipecat/processors/frameworks/rtvi.py
+++ b/src/pipecat/processors/frameworks/rtvi.py
@@ -5,6 +5,7 @@
 #
 
 import asyncio
+import base64
 from dataclasses import dataclass
 from typing import (
     Any,
@@ -31,6 +32,7 @@ from pipecat.frames.frames import (
     ErrorFrame,
     Frame,
     FunctionCallResultFrame,
+    InputAudioRawFrame,
     InterimTranscriptionFrame,
     LLMFullResponseEndFrame,
     LLMFullResponseStartFrame,
@@ -58,7 +60,9 @@ from pipecat.processors.aggregators.openai_llm_context import (
     OpenAILLMContextFrame,
 )
 from pipecat.processors.frame_processor import FrameDirection, FrameProcessor
+from pipecat.transports.base_input import BaseInputTransport
 from pipecat.transports.base_output import BaseOutputTransport
+from pipecat.transports.base_transport import BaseTransport
 from pipecat.utils.string import match_endofsentence
 
 RTVI_PROTOCOL_VERSION = "0.3.0"
@@ -819,6 +823,7 @@ class RTVIProcessor(FrameProcessor):
         self,
         *,
         config: RTVIConfig = RTVIConfig(config=[]),
+        transport: Optional[BaseTransport] = None,
         **kwargs,
     ):
         super().__init__(**kwargs)
@@ -843,6 +848,14 @@ class RTVIProcessor(FrameProcessor):
 
         self._register_event_handler("on_bot_started")
         self._register_event_handler("on_client_ready")
+
+        self._input_transport = None
+        self._transport = transport
+        if self._transport:
+            input_transport = self._transport.input()
+            if isinstance(input_transport, BaseInputTransport):
+                self._input_transport = input_transport
+                self._input_transport.enable_audio_in_stream_on_start(False)
 
     def observer(self) -> RTVIObserver:
         import warnings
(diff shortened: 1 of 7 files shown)


src/pipecat/processors/frameworks/rtvi.py
```diff
@@ -844,6 +849,12 @@ def __init__(
         self._register_event_handler("on_bot_started")
         self._register_event_handler("on_client_ready")
 
+        self._transport = transport
+        if self._transport:
+            input_transport = self._transport.input()
+            if isinstance(input_transport, BaseInputTransport):
+                input_transport.enable_audio_in_stream_on_start(False)
```
aconchillo, line 856: I think we can just do:

```
        if self._transport:
            self._transport.input().enable_audio_in_stream_on_start(False)
```

`input()` should always be a `BaseInputTransport`.

src/pipecat/processors/frameworks/rtvi.py
```diff
@@ -1025,9 +1038,40 @@ async def _handle_message(self, message: RTVIMessage):
             logger.warning(f"Exception processing message: {e}")
 
     async def _handle_client_ready(self, request_id: str):
+        logger.info("_handle_client_ready")
```
aconchillo, line 1041: This should probably be `debug` and something like `Received client-ready` or something like that.

src/pipecat/processors/frameworks/rtvi.py
aconchillo: or probably better, to make sure we have an input transport:

```
        if self._transport and self._transport.input():
            self._transport.input().enable_audio_in_stream_on_start(False)
```

aconchillo: One more though... Since we are using the input transport everywhere, maybe:

```
self._input_transport = transport.input() if transport else None

if self._input_transport:
            self._input_transport.enable_audio_in_stream_on_start(False)
```

src/pipecat/processors/frameworks/rtvi.py
```diff
@@ -1025,9 +1038,40 @@ async def _handle_message(self, message: RTVIMessage):
...
         self._client_ready_id = request_id
         await self.set_client_ready()
 
+    async def _handle_audio_buffer(self, data):
+        if not self._transport:
+            return
+
+        input_transport = self._transport.input()
+        if not isinstance(input_transport, BaseInputTransport):
+            return
```
aconchillo, line 1056: Since we store `_input_transport`:

```
        if not self._input_transport:
            return
```

src/pipecat/transports/services/daily.py
```diff
@@ -839,6 +838,13 @@ def __init__(self, client: DailyTransportClient, params: DailyParams, **kwargs):
     def vad_analyzer(self) -> Optional[VADAnalyzer]:
         return self._vad_analyzer
 
+    def start_audio_in_streaming(self):
+        logger.debug(f"Start receiving audio")
```
aconchillo, line 842: Should we move the log line inside the `if` below in case audio in is disabled?

aconchillo:
Cool! I think it makes more sense like this. Looks good! Just added minor comments.

src/pipecat/processors/frameworks/rtvi.py
filipi87: I am going to implement your suggestion of creating a variable for `_input_transport`, so we can use it anywhere. That will already help a lot.

filipi87: Done!

src/pipecat/processors/frameworks/rtvi.py
filipi87: Done!

src/pipecat/processors/frameworks/rtvi.py
filipi87: Done!

src/pipecat/transports/services/daily.py
filipi87: Done!

filipi87:
> Cool! I think it makes more sense like this. Looks good! Just added minor comments.
Yep, I agree. It makes more sense handling the messages inside RTVI.

I believe I have addressed all the comments. It should be ready for another round of review.

aconchillo (approved):

aconchillo:
LGTM! It would be nice to rebase and cleanup the commits to just a few commits that make sense.

PR #1218, opened 2025-02-13
Add UserIdleProcessor tests

markbackman:

This adds the new tests/util.py from: https://github.com/pipecat-ai/pipecat/pull/1209. And uses the SleepFrame() for timing related operations to test the UserIdleProcessor.

### DIFF
--- a/tests/test_user_idle_processor.py
+++ b/tests/test_user_idle_processor.py
@@ -1,0 +1,216 @@
+#
+# Copyright (c) 2024-2025 Daily
+#
+# SPDX-License-Identifier: BSD 2-Clause License
+#
+
+import asyncio
+import unittest
+
+from pipecat.frames.frames import (
+    BotSpeakingFrame,
+    BotStartedSpeakingFrame,
+    BotStoppedSpeakingFrame,
+    UserStartedSpeakingFrame,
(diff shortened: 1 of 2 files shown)


tests/test_user_idle_processor.py
```diff
@@ -0,0 +1,220 @@
...
+        frames_to_send = [
+            # Start conversation
+            UserStartedSpeakingFrame(),
+            UserStoppedSpeakingFrame(),
+            # Wait almost long enough for idle timeout
+            SleepFrame(sleep=0.08),
+            # Bot speaking frame should reset idle timer
+            BotSpeakingFrame(),
+            # Wait almost long enough for idle timeout again
+            SleepFrame(sleep=0.08),
```
aconchillo, line 71: We might want to increase the timeout to 0.2 and use 0.1 here. Very short times might be tricky in some machines or CPU loads.

aconchillo (approved):

tests/test_user_idle_processor.py
markbackman: Good idea. I'll relax some of the timings. That'll make the test slightly longer, but I think it's fine to avoid flakiness.

PR #1220, opened 2025-02-13
Instant voice example.

filipi87:
Instant voice demo example.


markbackman (approved):
This is really great! The connection does feel fast! ⚡ 

Suggestion: we should change the demo's name. How about `instant-voice-connection`?

PR #1222, opened 2025-02-14
WebsocketService: handle clean server disconnection

aconchillo:

The websocket async iterator doesn't raise an exception when the server disconnects cleanly. We should handle that and raise an exception so we can reconnect.


markbackman (approved):
LGTM!

PR #1223, opened 2025-02-14
audio context tts service and cartesia fixes

aconchillo:

We add a new `AudioContextWordTTSService`. This is a TTS base class for TTS services that handling multiple separate audio requests.

We use the new  `AudioContextWordTTSService` in `CartesiaTTSService` which fixes audio overlapping since it's now possible to send multiple requests at once.

### DIFF
--- a/src/pipecat/services/ai_services.py
+++ b/src/pipecat/services/ai_services.py
@@ -419,7 +419,7 @@ class WordTTSService(TTSService):
 
     async def start(self, frame: StartFrame):
         await super().start(frame)
-        await self._create_words_task()
+        self._create_words_task()
 
     async def stop(self, frame: EndFrame):
         await super().stop(frame)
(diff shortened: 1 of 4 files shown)


src/pipecat/services/ai_services.py
```diff
@@ -469,6 +469,115 @@ async def _words_task_handler(self):
...
+    The audio received from the TTS will be played in context order. That is, if
+    we requested audio for a context "A" and then audio for context "B", the
+    audio from context ID "A" will be played first.
+
+    """
+
+    def __init__(self, **kwargs):
+        super().__init__(**kwargs)
+        self._contexts_queue = asyncio.Queue()
+        self._contexts: Dict[str, asyncio.Queue] = {}
```
markbackman, line 487:
```
self._contexts_queue = asyncio.Queue()
self._contexts: Dict[str, asyncio.Queue] = {}
```
Do you need to initialize these here? They're also initialized in `_create_audio_context_task`

aconchillo: This is just so we have completions plus we should always list all the variables the class will use in the constructor. Otherwise we might create variables in other function and you wonder how that happened.

markbackman (approved):
LGTM!

PR #1231, opened 2025-02-16
Fix Type Hint in OpenAIContext::call_function

M1ngXU:
`dict[str, Any]` might be the wrong type, but `str` is almost always not the correct one

Maybe `Any` is a better type hint, but this should _usually_ be `dict[str, Any]`.

One usage is [here](https://github.com/pipecat-ai/pipecat/blob/b45f7fee6f66f8cf632d10674e0ee2b2185caf31/src/pipecat/services/anthropic.py#L213):

> arguments=json.loads(json_accumulator) if json_accumulator else dict()

The OpenAI service also passes a parsed JSON.

### DIFF
--- a/src/pipecat/processors/aggregators/openai_llm_context.py
+++ b/src/pipecat/processors/aggregators/openai_llm_context.py
@@ -193,7 +193,7 @@ class OpenAILLMContext:
         *,
         function_name: str,
         tool_call_id: str,
-        arguments: str,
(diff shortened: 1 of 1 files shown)


markbackman:
You were right. `str` was not correct. We ended up going with `arguments: Mapping[str, Any],`. Thanks for the contribution 🙇

PR #1233, opened 2025-02-17
Updated example to use Gemini

DominicStewart:
Updated example to use Gemini


DominicStewart:
Sounds good. Kwin did mention to just update the current example. But this is an easy change.

DominicStewart:
@aconchillo can I get a review please?

aconchillo (approved):

aconchillo:
LGTM! Please, use the option to squash so we keep a single commit.

PR #1240, opened 2025-02-18
deepgram: handle error event and reconnect

aconchillo:

We were not handling the `Error` event, so in the case of a sudden network disconnection we would just not do anything. we now try to reconnect. Note that the `start()` method sits there until connection is established so no need for retries (which I tried but they simply didn't solve anything).


markbackman (approved):
LGTM

PR #1241, opened 2025-02-18
deepgram: use the new nova-3 model as default

aconchillo:

### DIFF
--- a/src/pipecat/services/deepgram.py
+++ b/src/pipecat/services/deepgram.py
@@ -129,7 +129,7 @@ class DeepgramSTTService(STTService):
         default_options = LiveOptions(
             encoding="linear16",
             language=Language.EN,
-            model="nova-2-general",
+            model="nova-3-general",
             channels=1,
             interim_results=True,
             smart_format=True,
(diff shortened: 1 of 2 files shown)


src/pipecat/services/deepgram.py
```diff
@@ -127,7 +127,7 @@ def __init__(
         default_options = LiveOptions(
             encoding="linear16",
             language=Language.EN,
-            model="nova-2-general",
+            model="nova-3-general",
```
Vaibhav-Lodha, line 130: Nova 3 is only available in EN as of now, so we can do something like this - 

model="nova-2-general" if not (live_options or live_options.language == Language.EN) else "nova-3-general"

aconchillo: I need to check, but I'd rather let Deepgram return the error instead of us checking. The reason is that if they suddenly support more languages we would need to quickly release an update.

markbackman (approved):
LGTM

src/pipecat/services/deepgram.py
markbackman: Agreed that we should let Deepgram return the error. They will support more languages for Nova 3; Pipecat shouldn't be the gatekeeper for that.

aconchillo: Just made it clear in the CHANGELOG. Added an example on how to use the previous mode.

PR #1245, opened 2025-02-19
STTMuteFilter: change suppressed logging to trace

aconchillo:

Now that we also suppress `InputAudioRawFrame` it's better to have this at trace level, otherwise it's too verbose. We already have the other mute/unmute messages.


markbackman (approved):
Good catch!

PR #1250, opened 2025-02-20
AssistantContextAggregator: append aggregation and tools in the same turn

aconchillo:

You can have an LLM tool with the following prompt:

```
                function={
                    "name": "get_current_weather",
                    "description": "Always tell the user you will check right away before getting the current weather",
```

This might generate text tokens and function call in the same turn. Currently, we are discarding the text and only handling the function call.

### DIFF
--- a/src/pipecat/services/openai.py
+++ b/src/pipecat/services/openai.py
@@ -631,10 +631,13 @@ class OpenAIAssistantContextAggregator(LLMAssistantContextAggregator):
         run_llm = False
         properties: Optional[FunctionCallResultProperties] = None
 
-        aggregation = self._aggregation
+        aggregation = self._aggregation.strip()
         self.reset()
 
         try:
+            if aggregation:
+                self._context.add_message({"role": "assistant", "content": aggregation})
+
             if self._function_call_result:
                 frame = self._function_call_result
                 properties = frame.properties
@@ -668,9 +671,6 @@ class OpenAIAssistantContextAggregator(LLMAssistantContextAggregator):
                     else:
                         # Default behavior is to run the LLM if there are no function calls in progress
                         run_llm = not bool(self._function_calls_in_progress)
-
-            else:
-                self._context.add_message({"role": "assistant", "content": aggregation})
 
             if self._pending_image_frame_message:
                 frame = self._pending_image_frame_message
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,9 @@ stt = DeepgramSTTService(..., live_options=LiveOptions(model="nova-2-general"))
 
 ### Fixed
 
+- Fixed a context aggregator issue that would not append the LLM text response
+  to the context if a function call happened in the same LLM turn.
+
 - Fixed an issue that was causing HTTP TTS services to push `TTSStoppedFrame`
   more than once.
 
--- a/src/pipecat/processors/aggregators/llm_response.py
+++ b/src/pipecat/processors/aggregators/llm_response.py
@@ -145,6 +145,9 @@ class LLMResponseAggregator(BaseLLMResponseAggregator):
             frame = LLMMessagesFrame(self._messages)
             await self.push_frame(frame)
 
+            # Reset our accumulator state.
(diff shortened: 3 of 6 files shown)


aconchillo:
Interestingly, with this change if we send `TTSSpeakFrame` in the function start call, that text will also be added to the context since it's sent before the `LLMFullResponseEndFrame`. 

People have been asking about this and I guess this solves it. But I'm now wondering if this is a fix for what we couldn't do before or a bug. Thoughts?

src/pipecat/services/openai.py
```diff
@@ -635,6 +635,9 @@ async def push_aggregation(self):
         self.reset()
 
         try:
+            if aggregation:
```
filipi87, line 638: Should we implement this same fix in the cases: 
- `GrokAssistantContextAggregator` 
- `GoogleAssistantContextAggregator`

Looking at the code it feels they would face the same issue.

filipi87:
> People have been asking about this and I guess this solves it. But I'm now wondering if this is a fix for what we couldn't do before or a bug. Thoughts?

From the perspective of someone who is still learning Pipecat, I think it does feel like a fix. I guess I would expect everything I say to be part of the context, including the `TTSSpeakFrame`.

src/pipecat/services/openai.py
aconchillo: Ah, yes! Nice catch! Updated.

aconchillo:
> > People have been asking about this and I guess this solves it. But I'm now wondering if this is a fix for what we couldn't do before or a bug. Thoughts?
> 
> From the perspective of someone who is still learning Pipecat, I think it does feel like a fix. I guess I would expect everything I say to be part of the context, including the `TTSSpeakFrame`.

I'm not sure anymore. Before we were only added things to the context really said by the LLM, this is said by the developer.

Also, if we push `TTSSpeakFrame` from outside of a function call, that will not be added to the context (because it doesn't go between `LLMFullResponseStartFrame` and `LLMFullResponseEndFrame`).

filipi87:
> Also, if we push TTSSpeakFrame from outside of a function call, that will not be added to the context (because it doesn't go between LLMFullResponseStartFrame and LLMFullResponseEndFrame).

I see. Yeah, I think what we push through TTSSpeakFrame should always be part of the context or never be part of it. The current behavior does feel like a bug.

filipi87 (approved):
I know we have an ongoing discussion about whether TTSSpeakFrame should be included in the context, but I believe that can be handled in a follow-up PR to avoid blocking this one.

aconchillo:
> I know we have an ongoing discussion about whether TTSSpeakFrame should be included in the context, but I believe that can be handled in a follow-up PR to avoid blocking this one.

Actually, I'm not sure we can do that now. Because the text frames we append to the context are the ones output from the TTS. And at that point we don't know if they came from a `TTSSpeakFrame` or from an LLM response. So, right now they will be included in the context.

danthegoodman1:
> I see. Yeah, I think what we push through TTSSpeakFrame should always be part of the context or never be part of it. The current behavior does feel like a bug.

Based on the fact that you can access `run_tts` outside and get audio frames back (we do this to cache them for states), this would cause a few breaking changes I would have to write a ton of code for.

aconchillo:
> > I see. Yeah, I think what we push through TTSSpeakFrame should always be part of the context or never be part of it. The current behavior does feel like a bug.
> 
> Based on the fact that you can access `run_tts` outside and get audio frames back (we do this to cache them for states), this would cause a few breaking changes I would have to write a ton of code for.

Do you run `run_tts` on the same TTS instance that you add to the Pipeline? Or is it a separate one?

> 
> That also just seems like a separate issue, this is fixing a repeatable bug. The above would be changing behavior that's already understood by users. I say that as someone who desperately needs this bug solved.

Yes, separate but related. We are almost there though.

danthegoodman1:
> Do you run run_tts on the same TTS instance that you add to the Pipeline? Or is it a separate one?

Same currently, in theory we could make another 🤔

Issue #1247, opened 2025-02-19
TTSStoppedFrame getting triggered twice in several TTS Services


Vaibhav-Lodha:
I was going through this `elevenlabshttpservice` and other services as well and was wonder around why we need to push `TTSStoppedFrame` twice, also if my try fails before actual `TTSStartedFrame` then why should we even emit `TTSStoppedFrame` frame.

```
   try:
            await self.start_ttfb_metrics()

            async with self._session.post(
                url, json=payload, headers=headers, params=params
            ) as response:
                if response.status != 200:
                    error_text = await response.text()
                    logger.error(f"{self} error: {error_text}")
                    yield ErrorFrame(error=f"ElevenLabs API error: {error_text}")
                    return

                await self.start_tts_usage_metrics(text)
                yield TTSStartedFrame()

                async for chunk in response.content:
                    if chunk:
                        await self.stop_ttfb_metrics()
                        yield TTSAudioRawFrame(chunk, self._settings["sample_rate"], 1)

                yield TTSStoppedFrame()

        except Exception as e:
            logger.error(f"Error in run_tts: {e}")
            yield ErrorFrame(error=str(e))

        finally:
            yield TTSStoppedFrame()
```

Vaibhav-Lodha:
FYI: @jamsea ^

PR #1253, opened 2025-02-20
HTTP TTS services stopped frame

aconchillo:

- `RimeHttpTTSService` needs an `aiohttp.ClientSession` to be passed to the constructor as all the other HTTP-based services.

- `RimeHttpTTSService` doesn't use a default voice ID anymore.

Fixes https://github.com/pipecat-ai/pipecat/issues/1247

### DIFF
--- a/src/pipecat/services/cartesia.py
+++ b/src/pipecat/services/cartesia.py
@@ -364,9 +364,6 @@ class CartesiaHttpTTSService(TTSService):
     async def run_tts(self, text: str) -> AsyncGenerator[Frame, None]:
         logger.debug(f"Generating TTS: [{text}]")
 
-        await self.start_ttfb_metrics()
-        yield TTSStartedFrame()
-
         try:
             voice_controls = None
             if self._settings["speed"] or self._settings["emotion"]:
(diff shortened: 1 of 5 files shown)


src/pipecat/services/rime.py
```diff
@@ -391,36 +393,31 @@ async def run_tts(self, text: str) -> AsyncGenerator[Frame, None]:
 
         try:
             await self.start_ttfb_metrics()
-            await self.start_tts_usage_metrics(text)
-
-            yield TTSStartedFrame()
-
-            async with aiohttp.ClientSession() as session:
```
filipi87, line 398: I see, we were creating a new session for each request.

src/pipecat/services/playht.py
```diff
@@ -416,6 +417,8 @@ async def run_tts(self, text: str) -> AsyncGenerator[Frame, None]:
                         await self.stop_ttfb_metrics()
```
filipi87, line 417: Do we still need this here ?

src/pipecat/services/cartesia.py
```diff
@@ -386,14 +385,17 @@ async def run_tts(self, text: str) -> AsyncGenerator[Frame, None]:
                 _experimental_voice_controls=voice_controls,
             )
 
+            await self.start_tts_usage_metrics(text)
```
filipi87, line 388: Just to be sure, I see that we are invoking `start_tts_usage_metrics`, but I don't see us invoking anything to stop these metrics. Does this happen inside `stop_ttfb_metrics`, or is it something we are missing or not needed?

aconchillo: Yes, `start_tts_usage_metrics` doesn't have a stop counterpart.

filipi87 (approved):
LGTM!

PR #1260, opened 2025-02-20
Dom/gemini system prompt switching

DominicStewart:
Updated gemini example to use Gemini 2.0 Flash Lite


examples/phone-chatbot/README.md
```diff
@@ -106,20 +106,40 @@ curl -X POST "http://localhost:7860/daily_start_bot" \
...
 ```shell
 curl -X POST "http://localhost:7860/daily_gemini_start_bot" \                                                                                                        py pipecat
      -H "Content-Type: application/json" \
      -d '{"detectVoicemail": true}'
 ```
 
-Any request body properties supported by `/daily_start_bot` (such as "detectVoicemail", "dialoutnumber", etc) can also be passed to `/daily_gemini_start_bot`. The only difference is that calling the Gemini endpoint will start a Gemini bot session.
+All request body parameters supported by /daily_start_bot (such as detectVoicemail, dialoutNumber, etc.) are also compatible with /daily_gemini_start_bot.
+
+This example uses context switching to help steer the bot in the right direction. As Flash Lite is a smaller model, getting it to consistently call functions was difficult for these longer prompts. Breaking the prompt
```
markbackman, line 125: It may make more sense to just state what the best practice is.

```suggestion
This example uses context switching to help steer the bot in the right direction. As Flash Lite is a smaller model, breaking the prompt down into smaller piece helps to improve the bot's accuracy.
```

Also, what do you mean by smaller prompts? Maybe a short example would help to explain this line.

examples/phone-chatbot/README.md
```diff
@@ -106,20 +106,40 @@ curl -X POST "http://localhost:7860/daily_start_bot" \
...
+down into smaller pieces helped improve the accuracy of the bot.
+
+**Implementation Details**
+The implementation is available in bot_daily_gemini.py and features:
+
+Staged prompting approach: Breaking down complex tasks into smaller, more focused prompts to improve the lightweight model's performance
```
markbackman, line 131: Maybe make this a list?

```suggestion
- Staged prompting approach: Breaking down complex tasks into smaller, more focused prompts to improve the lightweight model's performance
```

examples/phone-chatbot/README.md
```diff
@@ -106,20 +106,40 @@ curl -X POST "http://localhost:7860/daily_start_bot" \
...
+Dynamic context switching: The bot can change its behavior in real-time based on what it detects (voicemail vs. human caller)
+Function-based architecture: Uses function calling to trigger context switches and call termination
+
+**Optimizations for Lightweight Models**
+Working with Gemini 2.0 Flash Lite required some specific optimizations:
+
+Simplified prompts: Each prompt focuses on a single task with clear instructions
```
markbackman, line 138: Same here, let's make it a list.
```suggestion
- Simplified prompts: Each prompt focuses on a single task with clear instructions
```

examples/phone-chatbot/README.md
```diff
@@ -106,20 +106,40 @@ curl -X POST "http://localhost:7860/daily_start_bot" \
...
+Function-driven state changes: The model calls specific functions to switch between different conversation modes
+Reduced context requirements: Each stage maintains only the context needed for its specific purpose
+
+This approach significantly improves the consistency of function calling in this lightweight model, which was challenging with longer, more complex prompts.
```
markbackman, line 142: Delete this too?

examples/phone-chatbot/bot_daily.py
```diff
@@ -49,7 +49,11 @@ async def main(
     # If you are handling this via Twilio, Telnyx, set this to None
     # and handle call-forwarding when on_dialin_ready fires.
 
-    dialin_settings = DailyDialinSettings(call_id=callId, call_domain=callDomain)
+    # We don't want to specify dialin settings if we're not dialing in
```
markbackman, line 52:
```suggestion
    # We don't want to specify dial-in settings if we're not dialing in
```

examples/phone-chatbot/bot_daily_gemini.py
```diff
@@ -33,11 +33,75 @@
...
+        # Create messages with updated system instruction
+        messages = [
+            {
+                "role": "system",
+                "content": system_instruction,
+            }
+        ]
+
+        # Update context with new messages
+        self._context_aggregator.set_messages(messages)
```
markbackman, line 52: Maybe a question for @aconchillo but is this the recommended way to set the context? I would expect apps to push an `LLMMessagesUpdateFrame` to set_messages.

To do that, either this class becomes a FrameProcessor, which probably isn't right. Or, it receives the task, so it can just queue_frames from the PipelineTask directly.

I'm not sure what's best, but we should agree since this will be an example that others will reference.

examples/phone-chatbot/bot_daily_gemini.py
```diff
@@ -33,11 +33,75 @@
...
+    def __init__(self, llm, context_aggregator):
+        self._llm = llm
+        self._context_aggregator = context_aggregator
+
+    async def switch_context(self, system_instruction):
+        """Switch the context to a new system instruction based on what the bot hears."""
+        # Create messages with updated system instruction
+        messages = [
+            {
+                "role": "system",
```
markbackman, line 46: Does Gemini support a `system` message?

examples/phone-chatbot/bot_daily_gemini.py
DominicStewart: Actually Aleix and I got on a call last week and this is what we landed on. I originally did do LLMMessagesUpdateFrame but there was a timing issue that caused problems. The issue is the context isn't updated immediately and that was causing function calls to get called twice. 

Aleix suggested using set_messages because that would set the new context instantly

examples/phone-chatbot/bot_daily_gemini.py
DominicStewart: It doesnt and all of the other Gemini examples used system too.

chadbailey59:
Merging this because I need to provide a link to it tonight.

chadbailey59 (approved):

PR #1263, opened 2025-02-21
fixing deepgram mismatch

Vaibhav159:
Fixes an issue in `DeepgramSTTService` where `sample_rate` passed to the `LiveOptions` was not being used, causing the service to use the default sample rate of pipeline.

This happens when `audio_in_sample_rate` is different from the sample rate we want in `DeepgramSTTService` and haven't passed it directly in the constructor via `sample_rate` params


aconchillo:
@Vaibhav159 Great catch! LGTM!

aconchillo (approved):

PR #1270, opened 2025-02-21
Added support to `ProtobufFrameSerializer` to send the transport messages

filipi87:
Added support to `ProtobufFrameSerializer` to send the messages from `TransportMessageFrame` and `TransportMessageUrgentFrame`.


src/pipecat/frames/frames.proto
```diff
@@ -35,10 +35,15 @@ message TranscriptionFrame {
   string timestamp = 5;
 }
 
+message JsonFrame {
```
aconchillo, line 38: I'd rename this to `MessageFrame` instead.

src/pipecat/frames/frames.proto
```diff
@@ -35,10 +35,15 @@ message TranscriptionFrame {
...
+    string data = 1;
+}
+
 message Frame {
   oneof frame {
     TextFrame text = 1;
     AudioRawFrame audio = 2;
     TranscriptionFrame transcription = 3;
+    JsonFrame json = 4;
```
aconchillo, line 47:
```
MessageFrame message = 4;
```

aconchillo:
Good idea! Just added a comment about the naming.

src/pipecat/frames/frames.proto
filipi87: Done.

src/pipecat/frames/frames.proto
filipi87: Done.

src/pipecat/serializers/protobuf.py
```diff
@@ -15,15 +16,23 @@
...
     TextFrame,
     TranscriptionFrame,
+    TransportMessageFrame,
+    TransportMessageUrgentFrame,
 )
 from pipecat.serializers.base_serializer import FrameSerializer, FrameSerializerType
 
 
+@dataclasses.dataclass
+class JSONFrame:
```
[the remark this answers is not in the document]
filipi87: Done.

filipi87:
Thanks for the review, @aconchillo! I’ve addressed all the comments.

src/pipecat/frames/frames.proto
```diff
@@ -35,10 +35,15 @@ message TranscriptionFrame {
...
+message MessageFrame {
+    string data = 1;
+}
+
 message Frame {
   oneof frame {
     TextFrame text = 1;
     AudioRawFrame audio = 2;
     TranscriptionFrame transcription = 3;
+    MessageFrame json = 4;
```
aconchillo, line 47: should this be

```
MessageFrame message = 4;
```

src/pipecat/serializers/protobuf.py
```diff
@@ -15,15 +16,24 @@
...
+class MessageFrame:
+    data: str
+
+
 class ProtobufFrameSerializer(FrameSerializer):
     SERIALIZABLE_TYPES = {
         TextFrame: "text",
         OutputAudioRawFrame: "audio",
         TranscriptionFrame: "transcription",
+        MessageFrame: "json",
```
aconchillo, line 36: `"message"`?

src/pipecat/frames/frames.proto
filipi87: I left it as JSON because we are always dumping the message as JSON. 

But maybe we should call it `data` to match this:
```
# Data class for converting transport messages into Protobuf format.
@dataclasses.dataclass
class MessageFrame:
    data: str
```

So, what do you think ? json, message or data ?

aconchillo (approved):

aconchillo:
LGTM! Just added one more comment, ignore if you think `json` is good, but since we renamed the type...

src/pipecat/frames/frames.proto
filipi87: I have renamed to `message`. 👍

Issue #1258, opened 2025-02-20
Agent should send interrupt on push_aggregation or `EmulateUserStartedSpeakingFrame` ?


seanmuirhead:
Have some cases where the agent is saying the same thing twice in a row without the use getting the chance to speak. I've been able to replicate, and I think it is because we are not pushing `StartInterruptionFrames` when we push `EmulateUserStartedSpeakingFrame` and when we are pushing an aggregation from the `LLMUserContextAggregator`?

aconchillo:
> Have some cases where the agent is saying the same thing twice in a row without the use getting the chance to speak. I've been able to replicate, and I think it is because we are not pushing `StartInterruptionFrames` when we push `EmulateUserStartedSpeakingFrame` and when we are pushing an aggregation from the `LLMUserContextAggregator`?

In theory, if a `EmulateUserStartedSpeakingFrame` is pushed upstream this will immediately cause an interruption: `StartInterruptionFrame` + `UserStartedSpeakingFrame`.

Can you describe Pipecat version and what processors do you have in your pipeline? For example, do you use a `STTMuteFilter`?

seanmuirhead:
> In theory, if a EmulateUserStartedSpeakingFrame is pushed upstream this will immediately cause an interruption: StartInterruptionFrame + UserStartedSpeakingFrame.

Ah makes sense


> Can you describe Pipecat version and what processors do you have in your pipeline? For example, do you use a STTMuteFilter?


Version: `0.0.57`

Yes we do use the `STTMuteFilter`, here is our Pipeline:
```
pipeline = Pipeline(
        [
            transport.input(),
            stt_mute_filter,  # Should always be before STT
            stt,
            context_aggregator.user(),
            llm,
            tts,
            custom_frame_processor,
            transport.output(),
            context_aggregator.assistant(),
        ],
    )

    task = PipelineTask(pipeline, params=PipelineParams(allow_interruptions=True))
```

The `custom_frame_processor` controls the `stt_mute_filter` sometimes, depending on a variety of TTS events

PR #1272, opened 2025-02-21
services: fix some TTS websocket service interruption handling

aconchillo:

@markbackman We already talked about this before. People are really hitting this issue, so even if not nice we need to support all TTS services as good as we can.

I did an optimization though. If the bot is not speaking and the user talks, we don't reconnect. We only reconnect if it's a real interruption.

Fixes https://github.com/pipecat-ai/pipecat/issues/1258

### DIFF
--- a/src/pipecat/services/ai_services.py
+++ b/src/pipecat/services/ai_services.py
@@ -15,6 +15,7 @@ from loguru import logger
 from pipecat.audio.utils import calculate_audio_volume, exp_smoothing
 from pipecat.frames.frames import (
     AudioRawFrame,
+    BotStartedSpeakingFrame,
     BotStoppedSpeakingFrame,
     CancelFrame,
     EndFrame,
@@ -40,6 +41,7 @@ from pipecat.frames.frames import (
 from pipecat.metrics.metrics import MetricsData
 from pipecat.processors.aggregators.openai_llm_context import OpenAILLMContext
 from pipecat.processors.frame_processor import FrameDirection, FrameProcessor
+from pipecat.services.websocket_service import WebsocketService
 from pipecat.transcriptions.language import Language
 from pipecat.utils.string import match_endofsentence
 from pipecat.utils.text.base_text_filter import BaseTextFilter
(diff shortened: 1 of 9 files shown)


src/pipecat/services/ai_services.py
```diff
@@ -434,6 +436,12 @@ async def _stop_frame_handler(self):
 
 
 class WordTTSService(TTSService):
+    """This a base class for TTS services that support word timestamps. Word
```
markbackman, line 439:
```suggestion
    """This is a base class for TTS services that support word timestamps. Word
```

src/pipecat/services/ai_services.py
```diff
@@ -503,11 +511,93 @@ async def _words_task_handler(self):
             self._words_queue.task_done()
 
 
-class AudioContextWordTTSService(WordTTSService):
-    """This services allow us to send multiple TTS request to the services. Each
-    request could be multiple sentences long which are grouped by context. For
-    this to work, the TTS service needs to support handling multiple requests at
-    once (i.e. multiple simultaneous contexts).
+class WebsocketTTSService(TTSService, WebsocketService):
+    """This a base class for websocket-based TTS services."""
```
markbackman, line 515:
```suggestion
    """This is a base class for websocket-based TTS services."""
```

markbackman: Same typo in other docstrings.

aconchillo: Done

src/pipecat/services/ai_services.py
aconchillo: Done

aconchillo:
> The code change makes sense and looks good. In testing the services, I see that Cartesia, ElevenLabs, and Rime all work as expected. In testing LMNT, Fish, and PlayHT, I see that the disconnecting causes a websocket error code 1000:
> ...
>  ...
> In all cases, the audio kept working, but these websocket errors may be problematic.

Good catch. These services disconnected from the websocket before canceling the receiving task.

markbackman:
This is working more reliably now.

For LMNT, I saw one websocket 1000 error, but not for the other `InterruptibleTTSService`s. I'm not sure what would have caused that error given the code change.

aconchillo:
> This is working more reliably now.
> 
> For LMNT, I saw one websocket 1000 error, but not for the other `InterruptibleTTSService`s. I'm not sure what would have caused that error given the code change.

OK, we now stop sending `eof` to LMNT. This seems to fix previous issues. Let me know if it works for you.

markbackman (approved):
Looks great now!

PR #1273, opened 2025-02-22
openai: add retry logic to get_chat_completions()

aconchillo:

Just a draft in case we want to discuss about it. I don't think we might need this after all.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- Added retry logic to `get_chat_completions()` fo all OpenAI-based LLM
+  services.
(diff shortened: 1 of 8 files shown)


aconchillo:
If something fails (network issue, access issue, timeout issue if timeout is set). This will trigger an exception and the LLM will simply say nothing. The user will talk again and if everything is restored it should just work.

Adding retry logic might delay the response but if things recover at least there will be a response. It's probably better than the LLM not saying anything. At least we give it one more chance.

aconchillo:
> @aconchillo do we have evidence that this is needed?

We do have some users that are seeing the LLM get stuck. As we discussed we will no add retry logic because that might be unexpected by the user. If there's a timeout they might want to handle it differently. Instead, I will create another PR so the user can register a new event handler.

PR #1278, opened 2025-02-24
Update AnthropicLLMService to use claude-3-7-sonnet-20250219 by default

markbackman:

This is the new state of the art model from Anthropic. It supports normal completions by default (and thinking as an option, which I'm not including due to the token use and latency).


filipi87 (approved):
LGTM!

PR #1280, opened 2025-02-24
services(llm): add on_completion_timeout event

aconchillo:


markbackman (approved):
Nice! 🙌 

This approach looks safe and effective to me. Emitting an event seems like the right call (vs retrying). This leaves the decision to the developer, which feels like the right call.

PR #1283, opened 2025-02-25
PipelineTask: add check_dangling_tasks parameter

aconchillo:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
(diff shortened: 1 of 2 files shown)


markbackman (approved):
LGTM.

This makes sense to add. I'm not sure why someone would want to disable it, but it's a handy option to have around.

aconchillo:
> LGTM.
> 
> This makes sense to add. I'm not sure why someone would want to disable it, but it's a handy option to have around.

The use case is when you first run a Pipeline and want to keep the processors around and you "stop" the pipeline but don't end it. In this case there will still be dangling tasks, but we don't want to report it.

PR #1285, opened 2025-02-25
handle stop task gracefully

aconchillo:

A couple of changes:

- We clarify the use of `StopFrame` and `StopTaskFrame`. Similar to `EndFrame` and `EndTaskFrame`, the `StopFrame` is pushed from the task and the `StopTaskFrame` is pushed from inside the pipeline. The stop frame is useful to stop the task but leaves the processors initialized.
- We avoid creating tasks which have already been created.

It might be that not all processors support this (i.e. being able to receive more than one StartFrame), but it's now working on a few tests I've ran.


filipi87:
@aconchillo, @markbackman, just thinking out loud.

Previously, we had `StartFrame` and `EndFrame`, which were meant to be invoked only once:
- `StartFrame`: The first frame that should be pushed down a pipeline.
- `EndFrame`: Indicates that a pipeline has ended, and frame processors and pipelines should be shut down.

Now, we are introducing `StopFrame`:
- `StopFrame`: Stops the task but keeps the processors initialized.

This makes total sense. However, instead of invoking `StartFrame` again, should we create a new frame to resume the pipeline, `ResumeFrame`?

That way, `StartFrame` would still be invoked only once, as the first frame to initialize the pipeline.

Does that make sense?

aconchillo:
> @aconchillo, @markbackman, just thinking out loud.
> 
> Previously, we had `StartFrame` and `EndFrame`, which were meant to be invoked only once:
> 
> * `StartFrame`: The first frame that should be pushed down a pipeline.
> * `EndFrame`: Indicates that a pipeline has ended, and frame processors and pipelines should be shut down.
> 
> Now, we are introducing `StopFrame`:
> 
> * `StopFrame`: Stops the task but keeps the processors initialized.
> 
> This makes total sense. However, instead of invoking `StartFrame` again, should we create a new frame to resume the pipeline, `ResumeFrame`?
> 
> That way, `StartFrame` would still be invoked only once, as the first frame to initialize the pipeline.
> 
> Does that make sense?

The `StopTaskFrame` already existed but we broke it in a previous release. What didn't exist is `StopFrame`. They both achieve the same exact thing but pushing from different places.

Stopping a Pipeline means the `runner.run(task)` will exit but it will not stop the processors. You can then use those same processors in a new Pipeline (and new task). This new pipeline might have old processosr but might also have new ones and the new ones need the `StartFrame`. So, you will do `runner.run(new_task)` and that will push a new `StartFrame` which will only initialize the new processors. The old processors will  already be initialized by the previous task.

So, I don't think we need a `ResumeFrame` for now. In this PR we (or try to) make `StartFrame` to be idempotent for processors which are already initialized.

PR #1288, opened 2025-02-25
TTSService: Remove newlines before sending text to TTS service to gen…

markbackman:
…erate

This came up in chatting in Discord. It sounds like the newline can cause problems for services like Azure TTS. Most are not impacted by this, but this is a safe way to ensure that text going into the TTS service is cleaner.

For example, Anthropic will, at times, generate text like this:
```

Because they make up everything!
```

Instead, we should strip that leading newline.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- The base `TTSService` class now strips leading newlines before sending text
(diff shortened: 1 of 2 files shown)


markbackman:
Does this warrant a changelog entry?

filipi87 (approved):
Looks good. Regarding your question, I think we could add a mention of that in the Changelog, since this could cause issues with some TTS providers.

markbackman:
I added the changelog entry.

aconchillo (approved):

PR #1289, opened 2025-02-25
small TTS http improvements

aconchillo:


markbackman (approved):
This change seems safe to make. In running the examples, I can't reproduce the static sound and there are no regressions. Approving!

PR #1290, opened 2025-02-25
P2P WebRTC transport option to Pipecat

filipi87:
P2P WebRTC transport option to Pipecat

We have two examples:
- [video-transform](https://github.com/pipecat-ai/pipecat/tree/aiortc_example/examples/aiortc/video-transform): demonstrating how to send and receive audio and video using SmallWebRTCTransport, and typescript. It also performs some image processing on the video frames using OpenCV.
- [voice-agent](https://github.com/pipecat-ai/pipecat/tree/aiortc_example/examples/aiortc/voice-agent): the simplest way to create a voice agent with SmallWebRTCTransport.


src/pipecat/services/gemini_multimodal_live/gemini.py
```diff
@@ -543,7 +545,13 @@ async def _send_user_audio(self, frame):
...
         if self._video_input_paused:
             return
-        # logger.debug(f"Sending video frame to Gemini: {frame}")
+
+        now = time.time()
+        if now - self._last_sent_time < 1:
+            return  # Ignore if less than 1 second has passed
+
+        self._last_sent_time = now  # Update last sent time
+        logger.debug(f"Sending video frame to Gemini: {frame}")
```
filipi87, line 554: If we send too many images in a second to Gemini, for example, at 30fps, Gemini will just start ignoring them.

examples/aiortc/video-transform/client/typescript/src/smallWebRTCTransport.ts
```diff
@@ -0,0 +1,472 @@
...
+    private isReconnecting = false;
+    private keepAliveInterval: number | null = null;
+    private audioDevice: string | undefined;
+    private videoDevice: string | undefined;
+
+    constructor(callbacks: SmallWebRTCTransportCallbacks) {
+        this._callbacks = callbacks
+        // for testing reconnections
+        // @ts-ignore
+        window.attemptReconnection = this.attemptReconnection.bind(this)
```
filipi87, line 45: Need to remember to remove this, only for testing.

kwindla (approved):
@filipi87 this is really nice. lgtm to merge!

examples/aiortc/video-transform/client/typescript/README.md
```diff
@@ -0,0 +1,27 @@
+# JavaScript Implementation
```
markbackman, line 1: Question on having this in client/typescript: Do we plan on having more client examples? If so, this is great. If not, then we can just put this in the client directory.

examples/aiortc/video-transform/README.rst
```diff
@@ -0,0 +1,43 @@
+# Video transform
```
markbackman, line 1: Should we change the name of the example directory to something other than `aiortc`? We want a name that will make it easy to understand what this is demonstrating. The key points are that this is a WebRTC example and also we want to differentiate it from Daily/LiveKit. So, maybe `p2p-webrtc` or something like that. WDYT?

examples/aiortc/voice-agent/README.rst
```diff
@@ -0,0 +1,34 @@
...
+3. Copy env.example to .env and configure:
+   - Add your API keys
+4. Start the server:
+   ```bash
+   python server.py
+   ```
+
+### Next, connect using the client app:
+
+Visit http://localhost:7860 in your browser.
```
markbackman, line 27: In just connecting via localhost, I get this error:
```
Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'getUserMedia')
```

I think this is because WebRTC wants an https connection. I was able to work around this by using ngrok to connect. Do you see the same thing?

Oddly, I don't see that when running the video example. Perhaps that's because the Vite server is doing something to satisfy browser requirements.. 🤷

examples/aiortc/voice-agent/README.rst
```diff
@@ -0,0 +1,34 @@
...
+1. Create and activate a virtual environment:
+   ```bash
+   python3 -m venv venv
+   source venv/bin/activate  # On Windows: venv\Scripts\activate
+   ```
+2. Install requirements:
+   ```bash
+   pip install -r requirements.txt
+   ```
+3. Copy env.example to .env and configure:
```
markbackman, line 18: I skipped right over this step in both of my read throughs. Maybe we just make this explicit so that it's easier to see. (I was looking for code blocks, I think).

3. Copy env.example to .env and add your Google API key:
```bash
cp env.example .env
```

markbackman (approved):
This is really, really good!

I don't have any comments on the implementation, but left a few on the examples. Mostly small nits, but things that I think could make it easier to use.

Anyway, approving and excited to get this added to Pipecat! Nice work 👏

examples/aiortc/voice-agent/README.rst
kwindla: Mark, local getUserMedia works for me when I use http://localhost:7860 but not http://0.0.0.0:7860 on both Chrome and Safari.

examples/aiortc/video-transform/README.rst
kwindla: +1 to p2p-webrtc

examples/aiortc/video-transform/client/typescript/README.md
filipi87: Yep, I am thinking on adding an iOS client for that. I would just need to create the transport first. 
But it would be pretty easy to add. Maybe we can plan this for the next couple of weeks.

examples/aiortc/video-transform/README.rst
filipi87: I think that makes sense. 👍

examples/aiortc/voice-agent/README.rst
filipi87: - For the voice agent, you can just run the server and connect through http://localhost:7860/ which is going to work, because as Kwin mentioned getUserMedia works for localhost.

- For the video-transform example, you need to execute both:
  - server: server.py
  - client: yarn dev.
  - And then your are going to access through http://localhost:5173/, inside the vite.config.js, we are redirecting the requests to the server.

markbackman: @filipi87 I understand that. What I'm saying is that my console outputs the following when running:
```
INFO:     Started server process [95133]
INFO:     Waiting for application startup.
INFO:     Application startup complete.
INFO:     Uvicorn running on http://0.0.0.0:7860 (Press CTRL+C to quit)
```

Someone is bound to click on `http://0.0.0.0:7860` (or copy it) and then not get access to media. I'm recommending either a README or a server change to reduce the likelihood that someone ends up going to `0.0.0.0`.

examples/aiortc/video-transform/client/typescript/src/smallWebRTCTransport.ts
```diff
@@ -0,0 +1,372 @@
+// TODO: we should refactor everything that is inside this class,
```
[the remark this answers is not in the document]
filipi87: done.

examples/aiortc/video-transform/client/typescript/src/smallWebRTCTransport.ts
filipi87: done.

examples/aiortc/voice-agent/README.rst
filipi87: done.

examples/aiortc/voice-agent/README.rst
filipi87: Ok, I have changed the server to start by default as localhost:
http://localhost:7860/

filipi87:
Thank you for the reviews @markbackman and @kwindla . 🙌

PR #1291, opened 2025-02-25
PlayHTHttpTTSService now takes a separate protocol input

markbackman:

This change is backwards compatible by parsing the protocol from the voice_engine, if provided.

### DIFF
--- a/src/pipecat/services/playht.py
+++ b/src/pipecat/services/playht.py
@@ -323,7 +323,8 @@ class PlayHTHttpTTSService(TTSService):
         api_key: str,
         user_id: str,
         voice_url: str,
-        voice_engine: str = "Play3.0-mini-http",  # Options: Play3.0-mini-http, Play3.0-mini-ws
+        voice_engine: str = "Play3.0-mini",
+        protocol: str = "http",  # Options: http, ws
         sample_rate: Optional[int] = None,
         params: InputParams = InputParams(),
         **kwargs,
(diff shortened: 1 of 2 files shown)


src/pipecat/services/playht.py
```diff
@@ -337,12 +338,24 @@ def __init__(
...
+
+        # Check if voice_engine contains protocol information (backward compatibility)
+        if "-http" in voice_engine:
+            # Extract the base engine name
+            voice_engine = voice_engine.replace("-http", "")
+            protocol = "http"
+        elif "-ws" in voice_engine:
+            # Extract the base engine name
+            voice_engine = voice_engine.replace("-ws", "")
+            protocol = "ws"
```
aconchillo, line 350: How would `ws` work in this HTTP service?

markbackman: This is an available PlayHT option. From their AsyncClient code:
```
    def tts(
        self,
        text: Union[str, list[str]],
        options: TTSOptions,
        voice_engine: Optional[str] = None,
        protocol: Optional[str] = None,
        streaming: bool = True
    ) -> AsyncIterable[bytes]:
        metrics = self._telemetry.start("tts-request")
        try:
            voice_engine, protocol = get_voice_engine_and_protocol(voice_engine, protocol)

            if protocol == "http":
                return self._tts_http(text, options, voice_engine, metrics, streaming)
            elif protocol == "ws":
                if not streaming:
                    raise ValueError("Non-streaming is not supported for WebSocket API")
                return self._tts_ws(text, options, voice_engine, metrics)
            elif protocol == "grpc":
                if not streaming:
                    raise ValueError("Non-streaming is not supported for gRPC API")
                return self._tts_grpc(text, options, voice_engine, metrics)
            else:
                raise ValueError(f"Unknown protocol {protocol}")
        except Exception as e:
            metrics.finish_error(str(e))
            raise e
```

In talking with them directly, they mentioned their HTTP streaming supports options using protocols of `http` and `ws`. Their docs are not the best, so it's unclear what effects the protocol has. (They mentioned `grpc` is only available on-prem. Anyway, hopefully that clears things up.

filipi87: Yeah, your changes make sense. But the API is kind of weird. I would expect to define the protocol when creating the AsyncClient.

filipi87 (approved):
LGTM!

src/pipecat/services/playht.py
markbackman: Yup, it is weird, but that's how the API works...

PR #1293, opened 2025-02-26
log pipecat version on application startup

aconchillo:

We needed this.

### DIFF
--- a/src/pipecat/__init__.py
+++ b/src/pipecat/__init__.py
@@ -1,0 +1,13 @@
+#
+# Copyright (c) 2024–2025, Daily
+#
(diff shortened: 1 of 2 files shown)


src/pipecat/__init__.py
```diff
@@ -0,0 +1,13 @@
...
+# SPDX-License-Identifier: BSD 2-Clause License
+#
+
+from importlib.metadata import version
+
+from loguru import logger
+
+__version__ = version("pipecat-ai")
+
+logger.info(f"ᓚᘏᗢ Pipecat {__version__} ᓚᘏᗢ")
```
markbackman, line 13: ❤️

markbackman (approved):
LGTM!

PR #1294, opened 2025-02-26
improve test-requirements.txt

aconchillo:


src/pipecat/processors/filters/stt_mute_filter.py
```diff
@@ -94,7 +93,7 @@ class STTMuteFilter(FrameProcessor):
     """
 
     def __init__(
-        self, *, config: STTMuteConfig, stt_service: Optional[STTService] = None, **kwargs
+        self, *, config: STTMuteConfig, stt_service: Optional[FrameProcessor] = None, **kwargs
```
aconchillo, line 96: This is to remove the `services` dependency from `processors`. I know this is deprecated, but just for the record and someone else reads this, the `processors` should never have a `services` dependency.

markbackman (approved):
LGTM

PR #1296, opened 2025-02-26
GoogleLLMService: always send text with audio

aconchillo:

It seems that sometimes, if we send inline audio data and no text part, everything will be fine. However, sometimes Google libraries just complain.

According to https://cloud.google.com/vertex-ai/generative-ai/docs/model-reference/inference#request text/inline_data/file_dat is a union. Maybe it's only true in the HTTP API.


markbackman (approved):
LGTM

PR #1302, opened 2025-02-26
services: minor LLM and TTS logging improvements

aconchillo:

Minor logging improvements so we can tell which processor is doing what (e.g. in the case you have 3 LLMs like in the 22 examples series).


markbackman (approved):
Good idea!

filipi87 (approved):
LGTM!

PR #1303, opened 2025-02-27
Add a new generic server to client message and frame type

markbackman:

Here's the client-side of this work: https://github.com/pipecat-ai/pipecat-client-web/pull/95.

The goal is to make server to client messaging easier.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,16 @@ All notable changes to **Pipecat** will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Added
+
+- Added a new frame, `ServerMessageFrame`, and RTVI message `RTVIServerMessage`
+  which provides a generic mechanism for sending custom messages from server to
+  client. The `ServerMessageFrame` can be processed by either a `RTVIProcessor`
+  or `RTVIObserver` and will be delivered to the client's `onServerMessage`
+  callback.
+
 ## [0.0.58] - 2025-02-26
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,9 +11,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Added a new frame, `ServerMessageFrame`, and RTVI message `RTVIServerMessage`
   which provides a generic mechanism for sending custom messages from server to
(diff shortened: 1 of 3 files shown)


src/pipecat/processors/frameworks/rtvi.py
```diff
@@ -934,6 +949,9 @@ async def process_frame(self, frame: Frame, direction: FrameDirection):
             await self.push_frame(frame, direction)
         elif isinstance(frame, TransportMessageUrgentFrame):
             await self._handle_transport_message(frame)
+        elif isinstance(frame, ServerMessageFrame):
+            await self._handle_server_message(frame)
+            await self.push_frame(frame, direction)
```
filipi87, line 954: Just to confirm, do we need this here? If we have both the Observer and the Frame Processor, wouldn't we end up sending the message twice to the client? I thought only the change inside the Observer would be enough.

CHANGELOG.md
```diff
@@ -5,6 +5,16 @@ All notable changes to **Pipecat** will be documented in this file.
...
 
+## [Unreleased]
+
+### Added
+
+- Added a new frame, `ServerMessageFrame`, and RTVI message `RTVIServerMessage`
+  which provides a generic mechanism for sending custom messages from server to
+  client. The `ServerMessageFrame` can be processed by either a `RTVIProcessor`
+  or `RTVIObserver` and will be delivered to the client's `onServerMessage`
+  callback.
```
filipi87, line 16: In general, if I am not mistaken, we only send messages to the client from the Observer. Our RTVIProcessor doesn’t send messages to the client unless it is a result of some action. So, probably, just the changes from the Observer should be enough here?

src/pipecat/processors/frameworks/rtvi.py
markbackman: Makes sense!

markbackman:
@filipi87 I removed the logic from the RTVIProcessor based on our discussion.

filipi87 (approved):
LGTM!

src/pipecat/frames/frames.py
```diff
@@ -704,6 +706,16 @@ def __str__(self):
...
 
 
+@dataclass
+class ServerMessageFrame(SystemFrame):
+    """A frame for sending server messages to the client."""
+
+    data: Any
+
+    def __str__(self):
+        return f"{self.name}(data: {self.data})"
```
aconchillo, line 716: I believe I would have put this in the `rtvi.py` file because it doesn't have any other use. And probably rename it to `RTVIServerMessageFrame` to make it clear this is RTVI specific.

markbackman: Makes sense. PR for that: https://github.com/pipecat-ai/pipecat/pull/1312

PR #1308, opened 2025-02-27
adding GoogleLLMOpenAIBetaService

Vaibhav159:
GoogleLLMOpenAIBetaService supports google gemini models with openai format

### DIFF
--- a/src/pipecat/services/google/google.py
+++ b/src/pipecat/services/google/google.py
@@ -12,6 +12,8 @@ import os
 import time
 
 from google.api_core.exceptions import DeadlineExceeded
+from openai import AsyncStream
+from openai.types.chat import ChatCompletionChunk
 
 # Suppress gRPC fork warnings
 os.environ["GRPC_ENABLE_FORK_SUPPORT"] = "false"
@@ -56,6 +58,8 @@ from pipecat.services.google.frames import LLMSearchResponseFrame
 from pipecat.services.openai import (
     BaseOpenAILLMService,
     OpenAIAssistantContextAggregator,
+    OpenAILLMService,
+    OpenAIUnhandledFunctionException,
     OpenAIUserContextAggregator,
 )
 from pipecat.transcriptions.language import Language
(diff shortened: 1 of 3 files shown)


aconchillo:
This is great!

src/pipecat/services/google/google.py
```diff
@@ -1188,6 +1193,118 @@ def create_context_aggregator(
...
+            tool_id_list.append(tool_call_id)
+
+            logger.debug(
+                f"Function list: {functions_list}, Arguments list: {arguments_list}, Tool ID list: {tool_id_list}"
+            )
+            for index, (function_name, arguments, tool_id) in enumerate(
+                zip(functions_list, arguments_list, tool_id_list), start=1
+            ):
+                if function_name == "":
+                    # TODO(remove: once Google fixes the returns index = None bug instead of returning the actual index):, resulting in "" function name
```
aconchillo, line 1290: Dumb question. Since this is compatible with OpenAI, why do we need `_process_context`? I assume it's because of this issue you mention here. If so, I'd make it more clear that we can remove the whole function when that's resolved.

aconchillo (approved):

aconchillo:
LGTM! Just a minor question.

src/pipecat/services/google/google.py
Vaibhav-Lodha: Correct, updated the TODO with the same.

Vaibhav-Lodha:
@aconchillo this should be gtg now.

aconchillo:
Just one more request. Since this is now reading (except for the minor nit). Do you mind rebasing and creating just a sinlge commit (or 2 if you want to keep the example in a different commit). This is just to avoid a bunch of commits that solve formatting, etc.

Vaibhav-Lodha:
@aconchillo updated the pr, and instead of making a new pr or merging commits, how about we squash them while merging the pr?
image

aconchillo:
> @aconchillo updated the pr, and instead of making a new pr or merging commits, how about we squash them while merging the pr? image

Sometimes I do that, but squashing doesn't add the merge commit that we would like to keep.

Vaibhav159:
@aconchillo understood, have updated the PR, this should be gtg, also check #1318 for the vertex ai usage with gemini models.

PR #1311, opened 2025-02-27
add new LLMFullResponseAggregator

aconchillo:

We used to have this processor. It is now simpler and had an event handler. I've seen people trying to get the LLM completion before the TTS.

### DIFF
--- a/src/pipecat/processors/aggregators/llm_response.py
+++ b/src/pipecat/processors/aggregators/llm_response.py
@@ -22,6 +22,7 @@ from pipecat.frames.frames import (
     LLMMessagesFrame,
     LLMMessagesUpdateFrame,
     LLMSetToolsFrame,
+    LLMTextFrame,
     StartFrame,
     StartInterruptionFrame,
     TextFrame,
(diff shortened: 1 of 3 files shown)


aconchillo:
This can be useful if you want to stop the whole completion even if it's not all spoken.

src/pipecat/processors/aggregators/llm_response.py
```diff
@@ -36,6 +37,51 @@
...
+
+        self._aggregation = ""
+        self._started = False
+
+        self._register_event_handler("on_completion")
+
+    async def process_frame(self, frame: Frame, direction: FrameDirection):
+        await super().process_frame(frame, direction)
+
+        if isinstance(frame, StartInterruptionFrame):
```
filipi87, line 59: Could it be worth sending what we have aggregated so far in this case, before we clean it? Maybe, along with the aggregation, we could also include information on whether it was completed or interrupted.

I have no strong feelings about this, just wondering if you see any use case where this could be useful.

filipi87 (approved):
LGTM!

markbackman (approved):
LGTM

src/pipecat/processors/aggregators/llm_response.py
aconchillo: Good idea! I can add an argument that is `complete: bool` or `final: bool`.

aconchillo: @filipi87 Please, take a look again.

filipi87: The only thing I was thinking about is the variable name. I'm wondering if it's clear enough or if it could cause some confusion. 

Maybe it would help to improve the documentation to specify when `final` will be `true` or `false`.

Or even be more verbose, like: `aggregation_status`: `completed` or `interrupted`.

What do you think ?

aconchillo: OK, as discussed. Changed from `final` to `completed`. Thank you!

PR #1317, opened 2025-02-28
Update gladia.py

Allenmylath:
# Gladia Speech Detection Parameters

According to the Gladia documentation (https://docs.gladia.io/api-reference/v2/live/init), the speech threshold parameter accepts values between 0 and 1, where:

- Values closer to 1 enable Gladia to better isolate speech from background noise
- This higher threshold helps filter out non-speech audio elements

The speech threshold parameter is notably used in Gladia's demo applications to provide cleaner speech recognition results in noisy environments.

### DIFF
--- a/src/pipecat/services/gladia.py
+++ b/src/pipecat/services/gladia.py
@@ -136,6 +136,8 @@ class GladiaSTTService(STTService):
         maximum_duration_without_endpointing: Optional[int] = 10
         audio_enhancer: Optional[bool] = None
         words_accurate_timestamps: Optional[bool] = None
+        speech_threshold: Optional[float] = .99
+        
 
     def __init__(
         self,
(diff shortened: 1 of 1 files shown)


src/pipecat/services/gladia.py
```diff
@@ -166,6 +166,7 @@ def __init__(
             "maximum_duration_without_endpointing": params.maximum_duration_without_endpointing,
             "pre_processing": {
                 "audio_enhancer": params.audio_enhancer,
+                "speech_threshold": 0.99,
```
aconchillo, line 169: we should probably add this to `InputParams` and then use  `params.speech_threshold` here.

Allenmylath: changes made

PR #1318, opened 2025-02-28
adding vertex google llm

Vaibhav159:
This adds vertex ai gemini llms which are designed for more production level loads


Vaibhav159:
@aconchillo Can we also review this one, it will help with better and stable gemini models via vertex.

Vaibhav159:
@markbackman @filipi87 Can either of us review this pr?

src/pipecat/services/google/google.py
```diff
@@ -1307,6 +1307,79 @@ async def _process_context(self, context: OpenAILLMContext):
                     )
 
 
+class GoogleVertexAIService(OpenAILLMService):
+    """Implements inference with Google's AI models via Vertex AI while maintaining OpenAI API compatibility.
+    Reference:
+    https://cloud.google.com/vertex-ai/generative-ai/docs/multimodal/call-vertex-using-openai-library
```
filipi87, line 1313: cool. 🙂

src/pipecat/services/google/google.py
```diff
@@ -1307,6 +1307,79 @@ async def _process_context(self, context: OpenAILLMContext):
...
+        # https://cloud.google.com/vertex-ai/generative-ai/docs/learn/locations
+        location: str = "us-east4"
+        project_id: str
+
+    def __init__(
+        self,
+        *,
+        credentials: Optional[str] = None,
+        credentials_path: Optional[str] = None,
+        model: str = "google/gemini-1.5-flash",
```
filipi87, line 1328: Should we default to `google/gemini-2.0-flash-001` ?

Vaibhav159: we can, thing is `google/gemini-2.0-flash-001` is not available in all regions as of now, so was thinking to keep it a model which is available widely

examples/foundational/14p-function-calling-gemini-vertex-ai.py
```diff
@@ -0,0 +1,142 @@
...
+                                "type": "string",
+                                "enum": ["celsius", "fahrenheit"],
+                                "description": "The temperature unit to use. Infer this from the users location.",
+                            },
+                        },
+                        "required": ["location", "format"],
+                    },
+                },
+            )
+        ]
```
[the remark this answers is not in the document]
Vaibhav159: updated

src/pipecat/services/google/google.py
```diff
@@ -1333,6 +1334,79 @@ async def _process_context(self, context: OpenAILLMContext):
                     )
 
 
+class GoogleVertexAIService(OpenAILLMService):
```
markbackman, line 1337: @Vaibhav159 why is this `GoogleVertexAIService` and not `GoogleVertexLLMService`?

markbackman: Also, I'm not familiar with Vertex AI vs Google AI studio. I know you added `GoogleLLMOpenAIBetaService` recently. Are the two services different enough that they require different class implementations?

In adding this, we'll have _3_ different Google LLM services, which makes it more confusing for a user to decide which one to pick.

Vaibhav-Lodha: Ideally yes, google ai studio and vertex ai is quite different, as vertex ai is a part of cloud offering and for teams working on gemini should switch to vertex ai for its security and other features related to GCP cloud like fine tuning and others.

Ideally, i feel once google moves out of beta for openai format we can remove the old format and use only new one, or we can also move vertex ai code to other file as its auth is different from google ai studio auth.

image

markbackman: > Ideally, i feel once google moves out of beta for openai format we can remove the old format and use only new one, or we can also move vertex ai code to other file as its auth is different from google ai studio auth.

The OpenAI format for Google is buggy, as far as I've read and hear, so I don't know if it makes sense to move in that direction yet (or maybe at all).

> Ideally yes, google ai studio and vertex ai is quite different, as vertex ai is a part of cloud offering and for teams working on gemini should switch to vertex ai for its security and other features related to GCP cloud like fine tuning and others.

Understood. As for the naming, it seems like `GoogleVertexLLMService` is more correct. Do you agree?

markbackman (approved):
LGTM. Thank you 🙏

src/pipecat/services/google/google.py
Vaibhav159: agreed, thought to add it as a easy way to add vertex ai if someone is looking for quick testing and i am kind of positive google will move forward with this sooner or later but in the end its your call if we want to add it or not, also have updated the name here.

PR #1320, opened 2025-02-28
Updated example to switch pipelines per the original request

DominicStewart:
Updates example to:

- Switch pipelines
- Use different LLMs for voicemail detection / conversation
- Use audio in for VM detection and STT for conversation

### DIFF
--- a/examples/phone-chatbot/bot_daily_gemini.py
+++ b/examples/phone-chatbot/bot_daily_gemini.py
@@ -328,17 +328,22 @@ async def main(
     else:
         logger.debug("+++++ No dialout number; assuming dialin")
 
-        @transport.event_handler("on_dialin_connected")
-        async def on_dialin_connected(transport, data):
-            # This event is not firing for some reason
-            logger.debug("+++++ Dial-in connected")
-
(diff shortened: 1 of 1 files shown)


examples/phone-chatbot/bot_daily_gemini.py
```diff
@@ -164,20 +175,10 @@ async def main(
...
     dialout_number: Optional[str],
 ):
-    # dialin_settings are only needed if Daily's SIP URI is used
-    # If you are handling this via Twilio, Telnyx, set this to None
-    # and handle call-forwarding when on_dialin_ready fires.
-
-    # We don't want to specify dial-in settings if we're not dialing in
     dialin_settings = None
-    if callId and callDomain:
+    if callId != "None" and callDomain != "None":
```
aconchillo, line 179: I think what you probably want is to make the arguments optional. That is:

```
async def main(....., callId: Optional[str], callDomain: Optional[str], ...)
```

and then you can keep doing:

```
    if callId and callDomain:
```

aconchillo (approved):

PR #1321, opened 2025-03-01
LLMService: add user/assistant args to create_context_aggregator()

aconchillo:

Allow passing user (`user_kwargs`) and assistant (`assistant_kwargs`) context aggregator parameters when using `create_context_aggregator()`. The values are passed as a mapping that will then be converted to arguments.

Currently we only have a function `create_context_aggregator()` to create both objects and we need a way to be able to pass custom arguments to both aggregators. One option is to add each argument but we will have to remember to add them every time to each of the different implementations. Also, each parameter needs to be renamed with a prefix `user_` or `assistant_`. Instead, this PR uses a couple of dictionaries `user_kwargs` and `assistant_kwargs` that are then used as keyword arguments. This is also future proof since there's no need to explicitly add every argument, of course it can be more error prone and will only be detected at runtime.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- Allow passing user (`user_kwargs`) and assistant (`assistant_kwargs`) context
+  aggregator parameters when using `create_context_aggregator()`. The values are
+  passed as a mapping that will then be converted to arguments.
+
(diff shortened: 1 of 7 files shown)


markbackman (approved):
LGTM!

Where's the best place to document how to use this? The best place might be to get this info into the reference docs with good docstrings in this code. Do you mind adding those?

aconchillo:
> LGTM!
> 
> Where's the best place to document how to use this? The best place might be to get this info into the reference docs with good docstrings in this code. Do you mind adding those?

Yep! Added!

PR #1325, opened 2025-03-02
Fix overlapping audio issue in AudioBufferProcessor

golbin:
This PR addresses the problem where audio segments were duplicated or overlapped specifically when the `buffer_size` option was enabled in the `AudioBufferProcessor`.

**Cause of the Issue:**
- When `buffer_size` was set, audio buffers from user and bot audio streams weren't properly aligned, causing overlapping segments during merging.

**Changes:**
- Modified `merge_audio_buffers()` to align user and bot audio buffer sizes before merging.
- Preserved leftover audio data in buffers after each merge to prevent data loss.
- Updated `_call_on_audio_data_handler()` to properly handle separated audio chunks.

**Benefits:**
- Eliminates audio overlapping and duplication when using the `buffer_size` option.
- Ensures continuity of audio processing without losing audio data.
- Simplifies event handling by clearly distinguishing merged and individual audio streams.

### DIFF
--- a/src/pipecat/processors/audio/audio_buffer_processor.py
+++ b/src/pipecat/processors/audio/audio_buffer_processor.py
@@ -5,7 +5,7 @@
 #
 
 import time
-from typing import Optional
+from typing import Optional, Tuple
 
 from pipecat.audio.utils import create_default_resampler, interleave_stereo_audio, mix_audio
 from pipecat.frames.frames import (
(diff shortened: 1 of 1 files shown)


golbin:
Hi @aconchillo @markbackman any thoughts on this?

golbin:
@aconchillo Any thoughts on this?

I applied this patch to my server during past a few weeks, and it basically works. However, if there's an audio delay due to network issues or CPU bottlenecks, the bot audio gets pushed back and becomes out of sync.

golbin:
Not sure if this is the right way to implement it for Pipecat, even though it seems to work fine.

golbin:
It may be resolved the latest version. (0.0.63~0.0.65)

But it's still with `user_continuous_stream=True` option.

markbackman:
@golbin what is the issue that's being resolved? I'm not aware of any overlapping audio in the update to date version.

golbin:
The issue happens when the user’s mic is off and then turned on shortly after. Silence should be inserted for the off period, but it’s missing — so the user’s audio starts too early and overlaps with the bot.

The issue occurred with both user_continuous_stream=True and False before. It seems to have been resolved for the False option between versions 0.0.63 and 0.0.65. However, the issue still remains when user_continuous_stream=True.

golbin:
When `buffer_size` was set.

PR #1326, opened 2025-03-02
Add speed as InputParam to ElevenLabs TTS services

markbackman:

In addition, did some cleanup by creating a common function to build the voice settings dictionary.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- Added `speed` as an `InputParam` for both `ElevenLabsTTSService` and
+  `ElevenLabsHttpTTSService`.
+
 - Added new `LLMFullResponseAggregator` to aggregate full LLM completions. At
   every completion the `on_completion` event handler is triggered.
 
--- a/src/pipecat/services/elevenlabs.py
+++ b/src/pipecat/services/elevenlabs.py
@@ -116,6 +116,44 @@ def output_format_from_sample_rate(sample_rate: int) -> str:
     return "pcm_16000"
 
 
+def build_elevenlabs_voice_settings(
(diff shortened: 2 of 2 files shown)


kwindla (approved):
lgtm

nicougou:
@markbackman may i ask why you made it so the `speed` is not applied without `stability` and `similarity_boost` ?

When using the ElevenLabs playground, they don't seem to enforce this rule, and since the `stability` and `similarity_boost` are poorly documented in their docs, it feels to experimental to try to set them up to work only on speed without side effects 🤔 

```
    voice_settings = {}
    if settings["stability"] is not None and settings["similarity_boost"] is not None:
        voice_settings["stability"] = settings["stability"]
        voice_settings["similarity_boost"] = settings["similarity_boost"]
        if settings["style"] is not None:
            voice_settings["style"] = settings["style"]
        if settings["use_speaker_boost"] is not None:
            voice_settings["use_speaker_boost"] = settings["use_speaker_boost"]
        if settings["speed"] is not None:
            voice_settings["speed"] = settings["speed"]
    else:
        if settings["style"] is not None:
            logger.warning(
                "'style' is set but will not be applied because 'stability' and 'similarity_boost' are not both set."
            )
        if settings["use_speaker_boost"] is not None:
            logger.warning(
                "'use_speaker_boost' is set but will not be applied because 'stability' and 'similarity_boost' are not both set."
            )
        if settings["speed"] is not None:
            logger.warning(
                "'speed' is set but will not be applied because 'stability' and 'similarity_boost' are not both set."
            )

    return voice_settings or None
```

markbackman:
There used to be a requirement for it. If you specified other values without stability and similarity_boost, it would return and error. It seems that's no longer the case, which is great, so we'll update the logic.

markbackman:
FYI: https://github.com/pipecat-ai/pipecat/pull/1641

PR #1331, opened 2025-03-04
Added ultravox service

milo157:
#### Creating a Ultravox Service compatible with Pipecat

This PR creates a Ultravox service that runs a [Ultravox model](https://huggingface.co/fixie-ai/ultravox-v0_5-llama-3_1-8b) of your choice. This replaces the STT and LLM components with the Ultravox model giving great latency results! 

In order to run Ultravox you need to use a GPU (We recommend [Cerebrium](https://www.cerebrium.ai)) 

What was tested:
- End-to-end pipeline
- Metrics implementation to track TTFB

What was not tested:
- I am unsure of how the service scales based on concurrency limitations of the Ultravox model you select

### DIFF
--- a/examples/foundational/07u-interruptible-ultravox.py
+++ b/examples/foundational/07u-interruptible-ultravox.py
@@ -1,0 +1,93 @@
+#
+# Copyright (c) 2024–2025, Daily
+#
+# SPDX-License-Identifier: BSD 2-Clause License
+#
+
+import asyncio
+import os
+import sys
+
+import aiohttp
+from dotenv import load_dotenv
+from loguru import logger
+from runner import configure
+
+from pipecat.audio.vad.silero import SileroVADAnalyzer
+from pipecat.pipeline.pipeline import Pipeline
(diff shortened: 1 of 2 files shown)


kwindla:
It's fantastic to have Ultravox support in Pipecat!

Question: does Ultravox also output audio? Can this service be configurable to output only text frames, only audio frames, or both?

Is it possible to add a single-file example in the `examples/foundational/` directory that shows how to use this service?

milo157:
@kwindla it doesn't support output audio yet, apparently its in the works so you have to use Cartesia/ElevenLabs/Rime. etc in the mean time. 

Added foundational example but needs to be run on a GPU

aconchillo (approved):

arpanpreneur:
Does tool calling / function calling work with it? Will we have support for this in pipecat-flows anytime soon?

moadel321:
> Does tool calling / function calling work with it? Will we have support for this in pipecat-flows anytime soon?

Did you find an answer to this ?

PR #1332, opened 2025-03-05
introduce BaseObject class

aconchillo:

This PR introduces a new `BaseObject` class (subclass of ABC) which is now the base class for the most important Pipecat classes. This makes it so we can now use `@abstractmethod` properly (it only makes sense if you inherit from ABC) and also it adds event handler support. Before, event handling was reimplemented in a few classes separately. This unifies it and also allows classes that didn't have event handler support have it.


markbackman:
@aconchillo can you add base_object.py?

aconchillo:
> @aconchillo can you add base_object.py?

Done! I was aware I forgot it, but then I forgot it again 😅

src/pipecat/utils/base_object.py
filipi87, line 1: Wouldn't it make sense to extract everything related to the event handler into yet another class—for example, EventEmitter or something similar—and then make BaseObject inherit from this class as well?

I guess this could be helpful in cases where we don't need to inherit from BaseObject but still want to use EventEmitter.

I have created something similar in my draft PR for WebRTC [here](https://github.com/pipecat-ai/pipecat/pull/1290/files#diff-da5be7496987e14819bf65537dcbe870c174b8c1a380ece4278fc5436007d849).

filipi87 (approved):
Nice improvement!
I only added a single comment about a possible refactor, but everything looks great anyway. 🚀

src/pipecat/utils/base_object.py
aconchillo: My thought was that everything (or almost everything) should be based on `BaseObject`. The almost everything is for example things like audio mixers and audio filters or observers. I think we can try without the `EventEmitter` for now. We can add it if we see we need to use events on things that are not `BaseObject`s.

filipi87: Yep, that is easy enough to refactor if needed. 👍

PR #1336, opened 2025-03-06
Pipecat small fixes and refactored function calling examples

filipi87:
Improvements:
- Fixed the base_whisper services to implement `set_language`.
- Added missing `flush_audio` method to `AzureTTSService`.
- Refactored the 14 series examples to use the unified format for function calling.
- Refactored gemini multimodal example to use the unified format for function calling.
- Removing the extra examples inside the unified-format-function-calling folder.


src/pipecat/services/azure.py
```diff
@@ -577,6 +577,10 @@ def _handle_canceled(self, evt):
         logger.error(f"Speech synthesis canceled: {evt.result.cancellation_details.reason}")
         self._audio_queue.put_nowait(None)
 
+    async def flush_audio(self):
+        logger.trace(f"{self}: flushing audio")
+        # TODO: check what we need to implement here ?
```
filipi87, line 582: @markbackman, @aconchillo , do we need to do something here ?

markbackman: In other services, we do things like reset the context ID, but the AzureTTSService doesn't have one to reset. We also use it to push any requirements to end the stream, but I'm not sure Azure has any.

Perhaps just the trace is sufficient for now?

markbackman (commented):
Changes look good. I wrote a comment about the flush audio, but I, too, am not sure what we should use it for. I think it's OK to just includes the trace logging for now since flush is required by the TTSService base class.

Other changes look good. Should we also remove the examples/unified-format-function-calling folder since the 14 examples are sufficient now?

src/pipecat/services/azure.py
markbackman: Maybe just logging is sufficient then.

filipi87:
> Other changes look good. Should we also remove the examples/unified-format-function-calling folder since the 14 examples are sufficient now?

I was thinking about this. The reason why I have not removed yet it is because I think those examples have the advantage of showing that simply changing the LLM is enough and that everything else can remain the same, which is a cool thing to demonstrate.

However, at the same time, we are somehow duplicating code. 

So, I don’t have strong feelings about this. I'm fine with either removing them or leaving them there. 😅

markbackman:
I think we should remove them to avoid example bloat. We already have so many examples that probably makes sense to keep them essential and curated.

filipi87:
> I think we should remove them to avoid example bloat. We already have so many examples that probably makes sense to keep them essential and curated.

Done. I have also refactored the gemini multimodal example, to show how to use the `custom_tools`.

markbackman (approved):
LGTM @filipi87! Thanks.

Let's see if Aleix has any requests for the AzureTTSService.

filipi87:
I am merging it, at least so we don't keep some examples blocked due to the errors above. However, I can include a fix in a follow-up PR in case Aleix believes something else is needed for the AzureTTSService.

PR #1341, opened 2025-03-07
Add a set_language convenience method for GoogleSTTService

markbackman:

Also, adds a flush_audio() method to GoogleTTSService, which is a no-op.

### DIFF
--- a/src/pipecat/services/google/google.py
+++ b/src/pipecat/services/google/google.py
@@ -58,7 +58,6 @@ from pipecat.processors.frame_processor import FrameDirection
 from pipecat.services.ai_services import ImageGenService, LLMService, STTService, TTSService
 from pipecat.services.google.frames import LLMSearchResponseFrame
 from pipecat.services.openai import (
-    BaseOpenAILLMService,
     OpenAIAssistantContextAggregator,
     OpenAILLMService,
     OpenAIUnhandledFunctionException,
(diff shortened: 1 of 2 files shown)


filipi87 (approved):
I know that Aleix will revisit this and check if all the abstract methods make sense, but I'm leaving my approval here in case this is blocking you in any way.

markbackman:
@filipi87 now that the abstract method is not longer a constraint, should this be `set_language` or `set_languages`? Google STT accepts either a single language or a list of languages. Any preference?

filipi87:
Maybe we should go with:
- `set_language(self, language: Language):`: for a single language.
- `set_languages(self, languages: List[Language])`: for the list.

And the `set_language` would always invoke `set_languages`.

What do you think ? This way we keep compatible in case we wish to enforce again `set_language` as abstract method.

markbackman:
@filipi87 that makes sense! Re-requesting your review since this PR is now different than the original.

src/pipecat/services/google/google.py
```diff
@@ -58,7 +58,6 @@
 from pipecat.services.ai_services import ImageGenService, LLMService, STTService, TTSService
 from pipecat.services.google.frames import LLMSearchResponseFrame
 from pipecat.services.openai import (
-    BaseOpenAILLMService,
```
markbackman, line 61: Removing unused import.

filipi87 (approved):
Looks good. 🚀

PR #1359, opened 2025-03-11
Tavus support for custom output rate

lucasrothman:

Tavus supports audio at 24kHz; it makes sense to have the output sample rate be customizable instead of hardcoded to 16kHz.

### DIFF
--- a/src/pipecat/services/tavus.py
+++ b/src/pipecat/services/tavus.py
@@ -37,7 +37,7 @@ class TavusVideoService(AIService):
         replica_id: str,
         persona_id: str = "pipecat0",  # Use `pipecat0` so that your TTS voice is used in place of the Tavus persona
         session: aiohttp.ClientSession,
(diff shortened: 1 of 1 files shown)


src/pipecat/services/tavus.py
```diff
@@ -37,13 +37,15 @@ def __init__(
         replica_id: str,
         persona_id: str = "pipecat0",  # Use `pipecat0` so that your TTS voice is used in place of the Tavus persona
         session: aiohttp.ClientSession,
+        output_sample_rate: int = 16000,
```
aconchillo, line 40: I'm wondering if we should just call this `sample_rate` as the rest of the services.

aconchillo:
Thank you! LGTM! Just one minor comment.

markbackman (approved):
LGTM! Thank you.

aconchillo (approved):

PR #1362, opened 2025-03-12
adding DTMFAggregator for telephony services.

Vaibhav159:
DTMFAggregator processor: Aggregates DTMF digits, outputs TranscriptionFrame on timeout or "#".


Vaibhav-Lodha:
@markbackman DTMF frames are generated from the Telephony serializers like twilio and we would need these as inputs in llm, hence using the transcript frame to keep the functionality as same.

markbackman:
I'm sorry we neglected this for too long. Here's an approach to solve the same problem, updated with some different considerations for configuration.
https://github.com/pipecat-ai/pipecat/pull/1889

Closing this one out, but thank you. We should have done this sooner!

PR #1363, opened 2025-03-12
adding llm mute filter

Vaibhav159:
Added `LLMMutefilter` as a general-purpose filter to mute transcript frames to reach context aggregators and eventually the LLM service. This can be used to prevent the LLM from processing unwanted audio or cases where we want to mute the user's audio basis some condition like end of call or back to back user utterances like `Hello, Hi`, etc.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -69,6 +69,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Added `AzureRealtimeBetaLLMService` to support Azure's OpeanAI Realtime API. Added
   foundational example `19a-azure-realtime-beta.py`.
 
+- Added `LLMMutefilter` as a general-purpose filter to mute transcript frames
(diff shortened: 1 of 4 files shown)


Vaibhav-Lodha:
@markbackman so my initial idea was to mute the llm basis on what user is saying, which can be received from stt, lets say user is continuously saying `hi` 4 times and making bot unable to respond, we can add custom function which makes sure on `5th` attempt user `hi` is not interrupting the pipeline and similar several cases where we would want to user to block the interruption basis on pipeline, and yes initially i coded it as a replacement of stt mute filter so added the similar strats.

markbackman:
@Vaibhav-Lodha I'm closing this PR. We're planning to add an interruption plan for number of word counting. This will likely be implemented in the aggregator, but that's work to still be completed. I believe this will cover the essence of what you're trying to accomplish, but done in a different way. Thanks for helping to push us in this direction 🙏

PR #1365, opened 2025-03-12
Fix Image.frombytes() mode param usage

akhilam512:
currently we're passing `frame.format` which typically contains file format strings like 'PNG' or 'JPEG' as the mode parameter to[ PIL.Image.frombytes()](https://pillow.readthedocs.io/en/stable/reference/Image.html#PIL.Image.frombytes)

This function actually expects image [modes](https://pillow.readthedocs.io/en/stable/handbook/concepts.html#concept-modes) like 'RGB', 'RGBA', etc

I standardized all Image.frombytes() calls to use 'RGB' mode instead, fixed this in six files:

- `src/pipecat/transports/base_output.py`
- `src/pipecat/services/gemini_multimodal_live/events.py`
- `src/pipecat/services/moondream.py`
- `src/pipecat/services/google/google.py`
- `src/pipecat/services/anthropic.py`
- `src/pipecat/processors/aggregators/openai_llm_context.py`

For most of the places where it's used, 'RGB' as the image mode works perfectly since it is immediately convert to JPEG anyway. However in `base_output.py`, I went with 'RGB' for consistency, though I wonder if preserving other modes is useful? Not sure if a png with an alpha channel can be passed here for example, and we might prefer RGBA in that case (to account for that, we might need to add a dedicated class member for storing image mode in `ImageRawFrame` perhaps)

### DIFF
--- a/src/pipecat/processors/aggregators/openai_llm_context.py
+++ b/src/pipecat/processors/aggregators/openai_llm_context.py
@@ -172,7 +172,7 @@ class OpenAILLMContext:
         self, *, format: str, size: tuple[int, int], image: bytes, text: str = None
     ):
         buffer = io.BytesIO()
-        Image.frombytes(format, size, image).save(buffer, format="JPEG")
+        Image.frombytes("RGB", size, image).save(buffer, format="JPEG")
         encoded_image = base64.b64encode(buffer.getvalue()).decode("utf-8")
 
         content = []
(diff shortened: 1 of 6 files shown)


akhilam512:
Hmm sorry this isn't a complete solution still, I was a bit too eager to send the PR. Testing this now I get:

```
unexpected exception: not enough image data
```

I think this is because as its name suggests `frombytes()` expects raw bytes but we're still sending a compressed bytes via a png/jpg file. So frombytes() is probably not the right usage here at all, instead using something like `Image.open(io.BytesIO(image))` and then saving the resultant image seems to work

markbackman:
@akhilam512 mind if I turn this into a draft until it's ready for review?

markbackman:
@akhilam512 is this something you're still working on or can I close it?

PR #1366, opened 2025-03-12
Add integration for Neuphonic TTS

adnansiddiquei:
This PR implements a service to use Neuphonic as a TTS provider. I am an engineer at Neuphonic.
- https://neuphonic.com/
- https://docs.neuphonic.com/quickstart


adnansiddiquei:
Awesome, I'll add in this example now!

adnansiddiquei:
I have added two examples and switched this PR into ready for review!

src/pipecat/services/neuphonic.py
```diff
@@ -0,0 +1,365 @@
...
+        # Convert enum value to string and get the base language part (e.g. es-ES -> es)
+        lang_str = str(language.value)
+        base_code = lang_str.split("-")[0].lower()
+        # Look up the base code in our supported languages
+        result = base_code if base_code in BASE_LANGUAGES.values() else None
+
+    return result
+
+
+class NeuphonicTTSService(WordTTSService, WebsocketService):
```
aconchillo, line 74: Please use `WebsocketWordTTSService` instead.

aconchillo: Actually, it seemsNeuphonic doesn't support word timestamps yet. In which case, we probably need `InterruptibleTTSService`. This will make sure interruptions work correctly.

src/pipecat/services/neuphonic.py
```diff
@@ -0,0 +1,365 @@
...
+        WordTTSService.__init__(
+            self,
+            aggregate_sentences=True,
+            push_text_frames=False,
+            push_stop_frames=True,
+            stop_frame_timeout_s=2.0,
+            sample_rate=sample_rate,
+            **kwargs,
+        )
+        WebsocketService.__init__(self)
```
aconchillo, line 100: remove.

src/pipecat/services/neuphonic.py
```diff
@@ -0,0 +1,365 @@
...
+        api_key: str,
+        voice_id: Optional[str] = None,
+        model: str = "neu_hq",
+        url: str = "wss://api.neuphonic.com",
+        sample_rate: Optional[int] = 22050,
+        encoding: str = "pcm_linear",
+        params: InputParams = InputParams(),
+        **kwargs,
+    ):
+        WordTTSService.__init__(
```
aconchillo, line 91: Change to `super().__init__(...` once we change the base class.

adnansiddiquei:
Thanks @aconchillo I have corrected all of your raised review comments.

I have also removed the notion of "models" in the Neuphonic implementation as Neuphonic now routes requests to the respective model based on the provided `voice_id` to make things simpler. So the end user does not need to worry about which model they need to select. Therefore I have removed all `self.set_model_name(model)` calls, I hope this doesn't mess with the metrics or anything on the `pipecat` side?

aconchillo (approved):

aconchillo:
> Thanks @aconchillo I have corrected all of your raised review comments.
> 
> I have also removed the notion of "models" in the Neuphonic implementation as Neuphonic now routes requests to the respective model based on the provided `voice_id` to make things simpler. So the end user does not need to worry about which model they need to select. Therefore I have removed all `self.set_model_name(model)` calls, I hope this doesn't mess with the metrics or anything on the `pipecat` side?

Thank you! I think it should be OK.

aconchillo:
LGTM!

PR #1371, opened 2025-03-13
Add TranscriptProcessor support for OpenAIRealtimeBetaLLMService

markbackman:

Since the speech-to-speech LLMs output both text and audio tokens, they should emit both `LLMTextFrame` and `TTSTextFrame`. I've added the latter in this PR.

In adding it, I found that `OpenAIRealtimeBetaLLMService` outputs text tokens differently than the other TTSServices. I updated the TranscriptProcessor to support this.


CHANGELOG.md
```diff
@@ -82,6 +82,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- Updated `TranscriptProcessor` to support text output from
+  `OpenAIRealtimeBetaLLMService`.
+
+- `OpenAIRealtimeBetaLLMService` now pushes a `TTSTextFrame`.
```
filipi87, line 88: Should we do the same for the GeminiMultimodalLiveLLMService, or is it already working there?

filipi87: We can do in a follow up PR, no need to include it here.

src/pipecat/processors/transcript_processor.py
```diff
@@ -90,9 +90,50 @@ def __init__(self, **kwargs):
...
+        Result: "Hello there! How's it going?"
+        """
         if self._current_text_parts and self._aggregation_start_time:
-            content = " ".join(self._current_text_parts).strip()
+            # Build content with intelligent spacing
+            content = ""
+            for i, part in enumerate(self._current_text_parts):
+                # Add a space only when the current part doesn't start with
+                # whitespace or punctuation/special characters
+                if i > 0 and not part.startswith((" ", ".", ",", "!", "?", ";", ":", "'", '"')):
```
filipi87, line 131: Would it make sense to include some other special characters, like:
`"(", ")", "[", "]", "{", "}", "-", "_", "/", "\\", "|", "~")`

What do you think ?

filipi87 (approved):
I have suggested a couple of other special characters to consider, but it looks good. Nice improvement.

CHANGELOG.md
markbackman: It's a one liner, so I added here. The downside though, is that the `GeminiMultimodalLiveLLMService` works differently than other services. The Gemini API doesn't currently output text, so we generate it instead.

This generation comes _after_ the bot started speaking and stopped speaking events. So, the TranscriptProcessor event handler doesn't fire until the next time the user speaks.

This is suboptimal, but I'm inclined to not cater for the case because we anticipate that the `GeminiMultimodalLiveLLMService` will add text output support, making this work as expected in the near future.

markbackman:
@filipi87 I finally came back to this and after a lot of experimentation I landed on a new algorithm for this. Check out the docstring in the code for the two types of transcripts. I use an algorithm to determine the type of TTS processor output, so I can decide to apply a space or not. This seems like a better approach than handling character special cases, which becomes very complex, very fast.

Anyway, let me know what you think.

src/pipecat/processors/transcript_processor.py
```diff
@@ -90,19 +90,70 @@ def __init__(self, **kwargs):
...
-            content = " ".join(self._current_text_parts).strip()
+            # Heuristic to detect pre-spaced fragments
+            uses_prespacing = False
+            if len(self._current_text_parts) > 1:
+                # Check if any fragment after the first one starts with whitespace
+                has_spaced_parts = any(
+                    part and part[0].isspace() for part in self._current_text_parts[1:]
+                )
+                if has_spaced_parts:
+                    uses_prespacing = True
```
filipi87, line 134: The only difference between this one and the previous one was that, I guess, if we receive the punctuation as an independent TextFrame, it would probably not be formatted the expected way, right? Like in this case:
```
TTSTextFrame: ["Hello"]
TTSTextFrame: ["there!"]
TTSTextFrame: ["!"]
TTSTextFrame: ["How"]
TTSTextFrame: ["is"]
TTSTextFrame: ["it"]
TTSTextFrame: ["going"]
TTSTextFrame: ["?"]
```

I guess this would result to this: "Hello there ! How is it going ?"

But I'm not sure if this is a scenario we are actually going to encounter. And I agree with you, this algorithm now feels a lot cleaner and easier to maintain. So, if we think we won’t receive the above example, this works perfectly.

filipi87 (approved):
Just a comment to double-check whether we want to handle this scenario or not. Either way, this is already a big improvement, and we can address that in the future if needed.

src/pipecat/processors/transcript_processor.py
markbackman: That scenario doesn't currently exist. The four word / timestamp pair providers are Cartesia, Rime, ElevenLabs, and OpenAI Realtime. The first three append the punctuations like "Hello!" or "Hello,". OpenAI Realtime chunks everything out with spaces.

I can't think of a different way to deliver these, so this seems like a sensible choice to keep the logic simple.

PR #1372, opened 2025-03-13
fix for 26-gemini-multimodal-live.py

kwindla:
We were mistakenly sending an initial messages frame as "assistant" (which translates to "model" in our context management code) to the Gemini Multimodal Live API. The back end seems to have gotten more strict and now expects that last message to be a "user" message.


aconchillo (approved):

aconchillo:
LGTM

PR #1373, opened 2025-03-13
Canonical updated audio processing

adriancowham:
Updated the Canonical service to use the `AudioBufferProcess`'s `on_audio_data` event:
- `AudioBufferProcess` is no longer a required param to the Canonical constructor.
- The Canonical service now supports chunked audio buffers from the `AudioBufferProcess`


aconchillo:
LGTM! I believe this will go well together with this new PR https://github.com/pipecat-ai/pipecat/pull/1376. When that's merged event handlers will be handled in separate tasks so even if it takes time to process/send audio to Canonical it will not affect the pipeline.

examples/canonical-metrics/bot.py
```diff
@@ -113,14 +113,20 @@ async def main():
                 llm,
                 tts,
                 transport.output(),
-                canonical,  # uploads audio buffer to Canonical AI for metrics
                 audio_buffer_processor,  # captures audio into a buffer
+                canonical,  # uploads audio buffer to Canonical AI for metrics
```
aconchillo, line 117: I'm now wondering if it makes sense to add `canonical` to the pipeline anymore. I believe the answer is no.

And continuing with that thought. I'm wondering is `CanonicalMetricsService` should be it's own python package in `pypi`. This way it can be used from any platform.

adriancowham: @aconchillo I was asking my self the same thing, but I currently depend on on the `EndFrame` and `CancelFrame` to detect the end of the conversation and trigger the upload to our backend. I guess the alternative would be to use
```
@transport.event_handler("on_participant_left")
```
...to trigger the end of call processing.

Also, I'm working on another PR where we collect data from the `MetricsFrame`, which requires `CanonicalMetricsService` to still be part of the pipeline.

aconchillo: Got it!

markbackman:
Sorry for the delay @adriancowham. Can you rebase? Then, we'll approve and merge.

adriancowham:
@markbackman I'm going to close this PR, unfortunately we're winding down Canonical. I'll submit another PR to remove our code and the example. Thanks for everything...more to come!

markbackman:
@adriancowham I'm sorry to hear 😢 

Hopefully this means something else good will come next!

PR #1376, opened 2025-03-14
event handlers are now executed in separate tasks

aconchillo:

All event handlers are now executed in separate tasks in order to prevent blocking the pipeline. It is possible that event handlers take some time to execute in which case the pipeline would be blocked waiting for the event handler to complete.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -95,6 +95,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- All event handlers are now executed in separate tasks in order to prevent
+  blocking the pipeline. It is possible that event handlers take some time to
+  execute in which case the pipeline would be blocked waiting for the event
+  handler to complete.
+
 - Updated `TranscriptProcessor` to support text output from
   `OpenAIRealtimeBetaLLMService`.
 
--- a/src/pipecat/utils/base_object.py
+++ b/src/pipecat/utils/base_object.py
@@ -4,6 +4,7 @@
 # SPDX-License-Identifier: BSD 2-Clause License
 #
 
+import asyncio
 import inspect
 from abc import ABC
 from typing import Optional
@@ -17,7 +18,14 @@ class BaseObject(ABC):
     def __init__(self, *, name: Optional[str] = None):
         self._id: int = obj_id()
         self._name = name or f"{self.__class__.__name__}#{obj_count(self)}"
+
+        # Registered event handlers.
         self._event_handlers: dict = {}
+
+        # Set of tasks being executed. When a task finishes running it gets
+        # automatically removed from the set. When we cleanup we wait for all
+        # event tasks still being executed.
+        self._event_tasks = set()
 
     @property
     def id(self) -> int:
@@ -26,6 +34,12 @@ class BaseObject(ABC):
     @property
     def name(self) -> str:
         return self._name
+
+    async def cleanup(self):
+        if self._event_tasks:
+            event_names, tasks = zip(*self._event_tasks)
+            logger.debug(f"{self} wating on event handlers to finish {list(event_names)}...")
+            await asyncio.wait(tasks)
 
     def event_handler(self, event_name: str):
         def decorator(handler):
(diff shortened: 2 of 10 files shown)


CHANGELOG.md
```diff
@@ -82,6 +82,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- All event handlers are now executed in separate tasks in order to prevent
```
kompfner, line 85: Just thinking about how the change described here would affect the new pipecat-flows "function" action feature that we were just talking about (the feature introduced in [this PR](https://github.com/pipecat-ai/pipecat-flows/pull/107))...

The idea is for the user to be able to specify various actions that occur sequentially after a bot's turn has ended:

```python
NodeConfig(
  task_messages=[
    {
      "role": "system",
      "content": "Greet the user and tell them a fun fact."
    },
    post_actions=[
      ActionConfig(
        type="function",
        handler=after_the_bot_turn
      ),
      ActionConfig(
        type="function",
        handler=after_the_first_action
      ),
      ActionConfig(
        type="end_conversation"
      )
    ]
  ]
)
```

With the changes in this PR, does that mean that we won't be able to guarantee the below expected order of operations?
* `after_the_bot_turn` runs
* `after_the_first_action` runs
*  the conversation ends

I'm especially concerned about the last one. Seems like the fact that there's a queue for all event-based tasks should take care of `after_the_bot_turn` and `after_the_first_action` happening in sequence. But ending the conversation just involves processing an `EndFrame`, which I believe doesn't involve any events. So would that potentially happen _first_?

kompfner: I guess more generally...this PR makes it so that you can't rely on events to do things "inline" in the pipeline (i.e. after the pipeline is done handling all previous frames but before subsequent frames).

Maybe this is would _only_ be an issue today for pipecat-flows function actions because they're the only thing that _needs_ events to fire "inline" in the pipeline? Does nothing else make that same assumption about events?

kompfner: If it only affects pipecat-flows function actions, then we would have to re-architect how those work...maybe we wouldn't use the `on_frame_reached_upstream` event, and instead have a way to register some callback directly on pipecat when a frame arrives all the way downstream (so...essentially a "synchronous" event callback...blech). Either that, or we go back to recommending that customers build their own frame processor that handles their own custom control frames...also blech. Also, wasn't `on_frame_reached_upstream` specifically built with pipecat-flows function actions in mind? Or was there another use-case? If there's no other use-case: maybe we don't need it, then? And if there is another use-case: would it suffer from the same sequencing issue as pipecat-flows function actions, illustrated above?

aconchillo: We just had a great discussion with @kompfner. Making event handlers a task will not be an issue.

Pipecat Flows new function actions need to happen in order. To achieve this, we currently queue a new custom function frame (but we don't wait for it to be executed). However, this custom frame doesn't guarantee when the custom function will be executed. That is, queuing just happens right away but you don't really know when the function frame will reach Pipecat Flows again so we can execute it. For example, you might queue the following:

```
  FunctionFrame1, TSSpeakFrame1
```

Note that when FunctionFrame1 goes through the TTS service, the next frame will be processed and the next frame is TTSSpeakFrame1. This means the TTS service will likely speak TTSSpeakFrame1 before even FunctionFrame1 has been executed in Pipecat Flows (because it hasn't reach Pipecat Flows yet).

The solution is for Pipecat Flows to wait until FunctionFrame1 is executed before executing the next action. That is, after queuing the custom frame we need to explicitly wait for that custom frame to reach Pipecat Flows again.

Guaranteeing action ordering in Pipecat Flows (by waiting for custom frame to be really executed) doesn't affect this PR, that is, it is indifferent if event handlers are executed in separate tasks or not.

src/pipecat/utils/base_object.py
```diff
@@ -19,6 +20,11 @@ def __init__(self, *, name: Optional[str] = None):
         self._name = name or f"{self.__class__.__name__}#{obj_count(self)}"
         self._event_handlers: dict = {}
 
+        self._scheduled_tasks = set()
```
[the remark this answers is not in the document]
aconchillo: Renamed to `_event_tasks`. 

I also got rid of the queue and the additional task that was processing it. I believe that launching a task is fast enough that we don't need to worry about having a specific tasks to launch tasks, we just do it right away.

markbackman (approved):
Approving for Paul!

PR #1377, opened 2025-03-14
PipelineTask: only call event handlers if a filter is matched

aconchillo:

Because of https://github.com/pipecat-ai/pipecat/pull/1376 event handlers will now be tasks. This means that we would create a task for every frame receive downstream or upstream. Usually, audio frames make it all the way down and we don't really want to create a task for each audio frame.

Even before, without tasks, I don't think we really wanted to call event handlers because most of the time those frames would just be ignored.

So, we now force the user to specify which frames they are interested in.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Added `on_frame_reached_upstream` and `on_frame_reached_downstream` event
   handlers to `PipelineTask`. Those events will be called when a frame reaches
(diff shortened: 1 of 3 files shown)


filipi87 (approved):
Makes sense.

kompfner (approved):
Testing right now on the warm transfer demo...but makes sense, you don't necessarily want to be firing away for every single type of frame.

kompfner:
> Testing right now on the warm transfer demo...

Works as intended!

kompfner:
Updated https://github.com/pipecat-ai/pipecat-flows/pull/107 to use `set_reached_downstream_filter()`.

PR #1378, opened 2025-03-14
removed most deprecations

aconchillo:

It's been a while for all these deprecations.


CHANGELOG.md
```diff
@@ -85,6 +85,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Updated the default mode for `CartesiaTTSService` and
   `CartesiaHttpTTSService` to `sonic-2`.
 
+### Removed
+
+- Removed deprecated `audio.resample_audio()`.
```
filipi87, line 90: Would it be worth to add here what they should use instead ? Like: `use 'create_default_resampler()' instead`

filipi87: The same applies to all the others that we have removed, in case we think this is a worthwhile approach.

filipi87 (approved):
Cool! Always nice to remove deprecated stuff to keep the code clean. 🚀

Not sure if you’ve already done that, but is there any chance that any of our demos were still using some of this deprecated code ? Maybe the RTVI stuff ?

markbackman (approved):
LGTM!

CHANGELOG.md
aconchillo: Done!

aconchillo:
> Cool! Always nice to remove deprecated stuff to keep the code clean. 🚀
> 
> Not sure if you’ve already done that, but is there any chance that any of our demos were still using some of this deprecated code ? Maybe the RTVI stuff ?

All the examples were already updated when we deprecated each of these.

mwood23:
There's two demo repos that are used fairly often by the community because they're more batteries included:
- https://github.com/pipecat-ai/gemini-multimodal-live-demo
- https://github.com/pipecat-ai/open-sesame

They're referenced in the Discord pretty often and the code shows patterns that other examples don't have (like HTTP mode) so it might not hurt to give those a look too!

PR #1379, opened 2025-03-14
introduce text aggregators

aconchillo:

Added new `BaseTextAggregator`. Text aggregators are used by the TTS service to aggregate LLM tokens and decide when the aggregated text should be pushed to the TTS service. It also allows for the text to be manipulated while it's being aggregated.

### DIFF
--- a/src/pipecat/utils/text/base_text_aggregator.py
+++ b/src/pipecat/utils/text/base_text_aggregator.py
@@ -1,0 +1,57 @@
+#
+# Copyright (c) 2024–2025, Daily
+#
+# SPDX-License-Identifier: BSD 2-Clause License
+#
+
+from abc import ABC, abstractmethod
+from typing import Optional
+
+
+class BaseTextAggregator(ABC):
+    """This is the base class for text aggregators. Text aggregators are usually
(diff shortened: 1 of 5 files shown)


aconchillo:
I'll add docs and tests in the morning.

src/pipecat/utils/text/base_text_aggregator.py
```diff
@@ -0,0 +1,22 @@
...
+    @abstractmethod
+    def aggregate(self, text: str) -> Optional[str]:
+        pass
+
+    @abstractmethod
+    def handle_interruption(self):
+        pass
+
+    @abstractmethod
+    def reset(self) -> str:
```
filipi87, line 21: Just wondering about the name, should it be something like `get_and_reset` or `retrieve_and_reset` to make it clear that this is how we retrieve the aggregated text ?

aconchillo: I ended up creating a property `text` and `reset()` doesn't return anything.

aconchillo:
> I'll add docs and tests in the morning.

This is ready now.

filipi87 (approved):
LGTM!

PR #1381, opened 2025-03-16
Update the 34-audio-recording.py example to include an STT processor

markbackman:

I'm including an STT processor because the need for audio_passthrough=True isn't a clear requirement. Hoping that this example helps.


filipi87 (approved):
LGTM!

PR #1387, opened 2025-03-18
Add PatternPairAggregator

markbackman:

Extends the BaseTextAggregator with one that's aimed at:
- Removing text from the LLM output before it is provided to the TTS service
- Includes a handler to provide the content between the pattern pair, so that the content can be used to do things, like change voices, which I show in a demo

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- Added new `PatternPairAggregator` that extends `BaseTextAggregator` to
+  identify content between matching pattern pairs in streamed text. This allows
+  for detection and processing of structured content like XML-style tags that
+  may span across multiple text chunks or sentence boundaries.
+
 - Added new `BaseTextAggregator`. Text aggregators are used by the TTS service
   to aggregate LLM tokens and decide when the aggregated text should be pushed
   to the TTS service. It also allows for the text to be manipulated while it's
@@ -123,6 +128,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Other
 
 - Update the `34-audio-recording.py` example to include an STT processor.
+
+- Added foundational example `35-voice-switching.py` showing how to use the new
+  `PatternPairAggregator`.
 
 - Added a Pipecat Cloud deployment example to the `examples` directory.
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -130,7 +130,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Update the `34-audio-recording.py` example to include an STT processor.
 
 - Added foundational example `35-voice-switching.py` showing how to use the new
-  `PatternPairAggregator`.
+  `PatternPairAggregator`. This example shows how to encode information for the
+  LLM to instruct TTS voice changes, but this can be used to encode any
+  information into the LLM response, which you want to parse and use in other
+  parts of your application.
 
 - Added a Pipecat Cloud deployment example to the `examples` directory.
 
--- a/examples/foundational/35-pattern-pair-voice-switching.py
+++ b/examples/foundational/35-pattern-pair-voice-switching.py
@@ -4,6 +4,46 @@
 # SPDX-License-Identifier: BSD 2-Clause License
 #
 
+"""Pattern Pair Voice Switching Example with Pipecat.
+
+This example demonstrates how to use the PatternPairAggregator to dynamically switch
+between different voices in a storytelling application. It showcases how pattern matching
+can be used to control TTS behavior in streaming text from an LLM.
+
+The example:
+    1. Sets up a storytelling bot with three distinct voices (narrator, male, female)
+    2. Uses pattern pairs (<voice>name</voice>) to trigger voice switching
+    3. Processes the patterns in real-time as text streams from the LLM
+    4. Removes the pattern tags before sending text to TTS
+
+The PatternPairAggregator:
+    - Buffers text until complete patterns are detected
+    - Identifies content between start/end pattern pairs
+    - Triggers callbacks when patterns are matched
+    - Processes patterns that may span across multiple text chunks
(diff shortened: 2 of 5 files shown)


examples/foundational/35-voice-switching.py
```diff
@@ -0,0 +1,192 @@
...
+
+
+async def main():
+    async with aiohttp.ClientSession() as session:
+        (room_url, token) = await configure(session)
+
+        transport = DailyTransport(
+            room_url,
+            token,
+            "Storytelling Bot",
```
filipi87, line 46: Nit: should we change to "Multiple voices Bot" or anything like this.

examples/foundational/35-voice-switching.py
filipi87, line 1: When I first looked at the example, I thought,: Why aren’t we using function calling for this ? 😅

But after reading the description, it made total sense. 

That makes me wonder if we could improve the example name, maybe something like `35-multiple-voices-bot` or another name that clearly indicates the bot will automatically play multiple characters with different voices.

CHANGELOG.md
```diff
@@ -116,6 +121,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Other
 
+- Added foundational example `35-voice-switching.py` showing how to use the new
+  `PatternPairAggregator`.
```
filipi87, line 125: Maybe leave the description more complete:

Added foundational example `35-voice-switching.py` showing how to use the new
  `PatternPairAggregator` to make the bot automatically play multiple characters with different voices.

src/pipecat/utils/text/pattern_pair_aggregator.py
```diff
@@ -0,0 +1,262 @@
...
+
+        Registers a new pattern pair with a unique identifier. The aggregator
+        will look for text that starts with the start pattern and ends with
+        the end pattern, and treat the content between them as a match.
+
+        Args:
+            pattern_id: Unique identifier for this pattern pair.
+            start_pattern: Pattern that marks the beginning of content.
+            end_pattern: Pattern that marks the end of content.
+            remove_match: Whether to remove the matched content from the text.
```
filipi87, line 96: This is a nice one.

examples/foundational/35-voice-switching.py
```diff
@@ -0,0 +1,192 @@
...
+                voice_id = VOICE_IDS[voice_name]
+                tts.set_voice(voice_id)
+                logger.info(f"Switched to {voice_name} voice")
+            else:
+                logger.warning(f"Unknown voice: {voice_name}")
+
+        pattern_aggregator.on_pattern_match("voice_tag", on_voice_tag)
+
+        # Set the pattern aggregator on the TTS service
+        tts._text_aggregator = pattern_aggregator
```
filipi87, line 85: I guess we should pass the `text_aggregator` when creating the `CartesiaTTSService`, because otherwise, it feels like we are modifying a private variable.

examples/foundational/35-voice-switching.py
markbackman: The benefit of using this approach is that function calls are slow, adding noticeable latency to the response. With this approach, the LLM can output many encoded instructions in a single turn. Also, the applications extend beyond just voice switching. You can now encode any information into the LLM response and just parse it out. Two other common cases I've heard are DTMF codes and thinking tokens.

examples/foundational/35-voice-switching.py
markbackman: Totally! Clearly the late night was getting to me.

CHANGELOG.md
markbackman: I'm taking a chance to educate a bit:
```
- Added foundational example `35-voice-switching.py` showing how to use the new
  `PatternPairAggregator`. This example shows how to encode information for the
  LLM to instruct TTS voice changes, but this can be used to encode any
  information into the LLM response, which you want to parse and use in other
  parts of your application.
```

I'll make this clear in docs too.

markbackman:
Thanks for the quick review @filipi87! This should be ready again.

examples/foundational/35-voice-switching.py
filipi87: Yep, I think it makes a lot of sense. 👍

filipi87 (approved):
Pretty cool. 🚀

examples/foundational/35-pattern-pair-voice-switching.py
```diff
@@ -0,0 +1,230 @@
...
+            end_pattern="</voice>",
+            remove_match=True,
+        )
+
+        # Register handler for voice switching
+        def on_voice_tag(match: PatternMatch):
+            voice_name = match.content.strip().lower()
+            if voice_name in VOICE_IDS:
+                voice_id = VOICE_IDS[voice_name]
+                tts.set_voice(voice_id)
```
aconchillo, line 111: I think this is fine since the processor executing this code is actually the TTS. In general, we would want to use frames. Maybe it's worth adding a comment about this.

src/pipecat/utils/text/pattern_pair_aggregator.py
```diff
@@ -0,0 +1,262 @@
...
+                        self._handlers[pattern_id](pattern_match)
+                    except Exception as e:
+                        logger.error(f"Error in pattern handler for {pattern_id}: {e}")
+
+                # Remove the pattern from the text if configured
+                if remove_match:
+                    processed_text = processed_text.replace(full_match, "", 1)
+                    modified = True
+
+        return processed_text, modified
```
aconchillo, line 179: Maybe this could have gone to `utils.string`. The function signature would allow passing the set of pairs `(start_tag, end_tag, remove_match)`.

examples/foundational/35-pattern-pair-voice-switching.py
markbackman: This needs to execute very quickly, so the method is the best way to go, I think. Even with the method, sometimes it's not fast enough.

src/pipecat/utils/text/pattern_pair_aggregator.py
markbackman: Happy to move it if you feel strongly. I don't really have a preference.

PR #1389, opened 2025-03-18
Set tool choice from context aggregator

balalofernandez:
Created `LLMSetToolChoiceFrame` to modify the llm's `tool_choice` during runtime.

### DIFF
--- a/src/pipecat/frames/frames.py
+++ b/src/pipecat/frames/frames.py
@@ -362,6 +362,12 @@ class LLMSetToolsFrame(DataFrame):
 
     tools: List[dict]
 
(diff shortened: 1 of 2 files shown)


markbackman:
@balalofernandez I'm not sure how this is different from `LLMSetToolsFrame`. `LLMSetToolsFrame` sets the tools for the LLM to use. As far as I recall, each time you set tools, it overwrites all tools. That is, you provide the complete set of tools for the context in `LLMSetToolsFrame`. It's not clear how `LLMSetToolChoiceFrame` is different. Can you explain?

PR #1392, opened 2025-03-18
Fix an issue where GoogleSTTService would timeout due to stream inact…

markbackman:
…ivity

### DIFF
--- a/src/pipecat/services/google/google.py
+++ b/src/pipecat/services/google/google.py
@@ -1972,7 +1972,8 @@ class GoogleSTTService(STTService):
                         break
 
                 except Exception as e:
-                    logger.error(f"Stream error, attempting to reconnect: {e}")
+                    logger.warning(f"{self} Reconnecting: {e}")
+
                     await asyncio.sleep(1)  # Brief delay before reconnecting
                     self._stream_start_time = int(time.time() * 1000)
                     continue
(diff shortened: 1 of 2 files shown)


src/pipecat/services/google/google.py
```diff
@@ -1972,7 +1972,11 @@ async def _stream_audio(self):
                         break
 
                 except Exception as e:
-                    logger.error(f"Stream error, attempting to reconnect: {e}")
+                    if "Stream timed out" in str(e):
+                        logger.debug("Reconnecting after stream timeout")
+                    else:
+                        logger.error(f"Stream error, attempting to reconnect: {e}")
```
aconchillo, line 1978: I would just put one `logger.warning` instead. Parsing the error message might not be ideal. They might change the message. Something like:

```
logger.warning(f"{self} Reconnecting: {e}")
```

src/pipecat/services/google/google.py
```diff
@@ -2024,4 +2028,13 @@ async def _process_responses(self, streaming_recognize):
...
 
         except Exception as e:
-            logger.error(f"Error processing Google STT responses: {e}")
+            # Check if the stream has timed out
+            # This can occur due to no inbound stream, which occurs when using an STTMuteFilter
+            if "Stream timed out" in str(e):
+                logger.debug("Google STT stream timed out due to inactivity - will reconnect")
+            else:
+                # Log unexpected errors as errors
+                logger.error(f"Error processing Google STT responses: {e}")
```
aconchillo, line 2037: Same here.

aconchillo (approved):

aconchillo:
LGTM. Minor comment.

src/pipecat/services/google/google.py
markbackman: Sounds good. That'll leave errors and warnings in the logs, but I guess that's OK. We could try to string match on the error code (409), but that's also a string that could change.

PR #1394, opened 2025-03-18
function calls as tasks

aconchillo:

The main idea of the PR is that now function calls run in separate tasks. This prevents the pipeline from being blocked and even allows the user to keep interacting with the LLM while the function call is being executed.

We also deprecate `start_callback`. This was just a function that was being executed before the actual function call, so there's really no need for it.

It also fixes an issue that was some times causing the assistant aggregator to not aggregate texts during function calls.


CHANGELOG.md
```diff
@@ -113,6 +124,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Updated the default mode for `CartesiaTTSService` and
   `CartesiaHttpTTSService` to `sonic-2`.
 
+### Deprecated
+
+- Passing a `start_callback` to `LLMService.register_function()` is now
```
markbackman, line 129: This is kind of a bummer. `start_callback` provided a really easy way for developers to hook into the function call and emit text. What's our recommendation for doing that now.

The use case is: the function is going to take a while to execute, so you add a function to push a TTSSpeakFrame to say something like "hold on while I look that up..."

CHANGELOG.md
```diff
@@ -9,6 +9,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- When registering a function call it is now possible to indicate if you want
+  the function call to be cancelled if there's a user interruption via
+  `cancel_on_interruption` (defaults to True). This is now possible because
```
markbackman, line 14: Very cool! Should we set `cancel_on_interruption=False` in our examples? I would imagine this will be the preferred way to use function calls.

CHANGELOG.md
markbackman, line 129: The recommendation is just to add `start_callback` functionality to the handler that's called by the function call, right?

CHANGELOG.md
aconchillo: As we discussed, this is now set to `False` by default.

src/pipecat/frames/frames.py
```diff
@@ -706,6 +715,18 @@ def __str__(self):
         return f"{self.name}(pts: {pts}, text: [{self.text}], size: {self.size}, format: {self.format})"
 
 
+@dataclass
+class UserImageMessageFrame(SystemFrame):
+    """An image associated to a user."""
```
kompfner: This is the same description as given to `UserImageRawFrame`. What makes the `UserImageMessageFrame` different? Describe that.

src/pipecat/frames/frames.py
```diff
@@ -634,6 +634,15 @@ class FunctionCallInProgressFrame(SystemFrame):
     function_name: str
     tool_call_id: str
     arguments: str
+    cancel_on_interruption: bool
+
+
+@dataclass
+class FunctionCallCancelFrame(SystemFrame):
+    """A frame to signal a function cal has been cancelled."""
```
kompfner, line 642: Nit: `call`

src/pipecat/services/ai_services.py
```diff
@@ -137,6 +141,13 @@ async def process_generator(self, generator: AsyncGenerator[Frame | None, None])
                     await self.push_frame(f)
 
 
+@dataclass
+class FunctionCallEntry:
+    function_name: Optional[str]
+    callback: Any  # TODO(aleix): add proper typing.
```
kompfner, line 147: just leaving a note here since this is marked as todo

src/pipecat/services/ai_services.py
```diff
@@ -137,6 +141,13 @@ async def process_generator(self, generator: AsyncGenerator[Frame | None, None])
                     await self.push_frame(f)
 
 
+@dataclass
+class FunctionCallEntry:
```
kompfner, line 145: nit: `FunctionEntry` might be better because it pertains to a registered function, not an invocation of a function

src/pipecat/services/ai_services.py
```diff
@@ -146,38 +157,74 @@ class LLMService(AIService):
...
         # for all functions
-        self._callbacks[function_name] = callback
-        # QUESTION FOR CB: maybe this isn't needed anymore?
+        self._functions[function_name] = FunctionCallEntry(
+            function_name=function_name,
+            callback=callback,
+            cancel_on_interruption=cancel_on_interruption,
+        )
+
+        # Start callbacks are now deprecated.
```
kompfner, line 206: 👍 . Could never figure out quite why these were needed.

src/pipecat/services/ai_services.py
kompfner, line 147: nit: `callback` is the function itself, right? any reason it's called `callback`? (to me, that makes it sound like something that happens on completion). But maybe this is standard naming in Python or this project that I'm just ignorant of, in which case ignore me!

src/pipecat/services/ai_services.py
```diff
@@ -217,6 +257,106 @@ async def request_image_frame(self, user_id: str, *, text_content: Optional[str]
...
+        self,
+        context: OpenAILLMContext,
+        tool_call_id: str,
+        function_name: str,
+        arguments: str,
+        run_llm: bool = True,
+    ):
+        if function_name in self._functions.keys():
+            entry = self._functions[function_name]
+        elif None in self._functions.keys():
```
kompfner, line 283: Just for my understanding: if a function is registered with no name (`None`), then it will be called anytime a function with an "unrecognized" name is called?

kompfner (approved):
Mostly just focused on the function-calls-as-tasks change itself and not all the fallout from that change. This'll be a nice change!

src/pipecat/services/ai_services.py
aconchillo: You are right.

src/pipecat/services/ai_services.py
aconchillo: It could be `function`.

src/pipecat/frames/frames.py
aconchillo: This is outdated. `UserImageMessageFrame` doesn't exist anymore.

aconchillo: Well, I added it in this PR but then simplified a few things and just used `UserImageRequestFrame` and `UserImageRawFrame`.

src/pipecat/services/ai_services.py
aconchillo: Currently the callbacks signatures are too long. I'm planning to create a data class and pass it as a single argument.

aconchillo: Well, callbacks are just functions called by other functions, so I think it's a good term.

aconchillo: I guess it's an old school term in this sense.

PR #1395, opened 2025-03-19
TTSService: allow passing multiple text filters and aggregators

aconchillo:

This PR allows multiple text filters and aggregators, instead of a single one. This way we can have smaller classes that each do one thing and this makes them composable.


markbackman (approved):
Nice! LGTM

PR #1396, opened 2025-03-19
SegmentedSTTService: allow audio to pass-through downstream

aconchillo:

STT services based on `SegmentedSTTService` were not passing through audio even if `audio_passthrough` was True. This was preventing audio recording from working.


markbackman (approved):
LGTM!

PR #1400, opened 2025-03-19
Allow OpenAIRealtimeBetaLLMService to push silence when done

nikp06:
- Implemented the possibility to push silence when audio is done, similar to how `TTSService` class is handling this.
- This change is necessary for systems with client-side vad (e.g. bot-bot conversation), in order for the client to know, when the server is done with audio generation.


markbackman:
OpenAIRealtimeLLMService (and the Beta variant) push all frame types to make it look similar to a cascade architecture bot. Pushing silence shouldn't be required anymore. Closing out this PR.

PR #1401, opened 2025-03-19
Add support for previous text in elevenlabs http processor

### LINKED ISSUE #1399: Elevenlabs use `previous_text` to improve generation
Pipecat chops up LLM generations send to TTS by punctuation to reduce latency. This creates problems when it feeds TTS a short utterance like `Ok.`, which often causes it to SCREAM randomly, and at best, have a completely different tonality. Short TTS messages give highly variable results, and destroy immersion.

If the TTS was optionally able to access the context, then it could take the most recent message(s) it had generated and populate that in the `previous_text` parameter to enable far smoother TTS generations.

danthegoodman1:

Allows the user to optionally provide the context array for the previous `n` messages that the assistant has said to pass to the `previous_text` parameter to enable more natural sounding speech with elevenlabs.

Only for the HTTP processor atm.

Closes #1399

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - It is now possible to disable `SoundfileMixer` when created. You can then use
   `MixerEnableFrame` to dynamically enable it when necessary.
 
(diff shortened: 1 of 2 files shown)


danthegoodman1:
This works _really well_ for short utterances. Less random screaming.

PR #1404, opened 2025-03-19
GeminiMultimodalLiveLLMService: fix duplicated messages in context

### LINKED ISSUE #1384: Duplication of text messages in gemini.py
This part in ```gemini.py``` leads to duplication of text messages in my webrtc pipeline sent with ```LLMMessagesAppendFrame```:

```python
class GeminiMultimodalLiveUserContextAggregator(OpenAIUserContextAggregator):
        # kind of a hack just to pass the LLMMessagesAppendFrame through, but it's fine for now
        if isinstance(frame, LLMMessagesAppendFrame):
```

aconchillo:

Fixes #1384

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -142,6 +142,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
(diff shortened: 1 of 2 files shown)


markbackman (approved):
Good catch!

PR #1405, opened 2025-03-19
WebsocketTTSService: add `on_connection_error` and `reconnect_on_error`

aconchillo:

Instead of reconnecting to the TTS services, this allows to fallback to a different provider (e.g. using a `ParallelPipeline`) if something goes wrong.

### DIFF
--- a/src/pipecat/services/ai_services.py
+++ b/src/pipecat/services/ai_services.py
@@ -549,11 +549,25 @@ class WordTTSService(TTSService):
 
 
 class WebsocketTTSService(TTSService, WebsocketService):
-    """This is a base class for websocket-based TTS services."""
+    """This is a base class for websocket-based TTS services.
 
-    def __init__(self, **kwargs):
+    If an error occurs with the websocket, an "on_connection_error" event will
+    be triggered:
+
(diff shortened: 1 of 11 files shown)


src/pipecat/services/ai_services.py
```diff
@@ -549,11 +549,25 @@ async def _words_task_handler(self):
 
 
 class WebsocketTTSService(TTSService, WebsocketService):
-    """This is a base class for websocket-based TTS services."""
+    """This is a base class for websocket-based TTS services.
 
-    def __init__(self, **kwargs):
+    If an error occurs with the websocket a "on_connection_error" event will be
```
markbackman, line 554: Nits
```suggestion
    If an error occurs with the websocket, an "on_connection_error" event will be
```

markbackman: Same in other locations.

markbackman (approved):
This looks good and works as expected. Nice!

src/pipecat/services/ai_services.py
aconchillo: Fixed!

PR #1406, opened 2025-03-20
PipelineTask: automatically cancel tasks if pipeline is idle

aconchillo:

This PR adds support for detecting idle pipelines. By default, if no activity has been detected during 5 minutes, the `PipelineTask` will be automatically cancelled. It is possible to override this behavior by passing `cancel_on_idle_timeout=False`. It is also possible to change the default timeout with `idle_timeout_secs`. Finally, a `on_idle_timeout` event handler will be triggered if the idle timeout is reached (whether the pipeline task is cancelled or not).

### DIFF
--- a/src/pipecat/pipeline/task.py
+++ b/src/pipecat/pipeline/task.py
@@ -5,6 +5,7 @@
 #
 
 import asyncio
+import time
 from typing import Any, AsyncIterable, Dict, Iterable, List, Optional, Tuple, Type
 
 from loguru import logger
@@ -13,6 +14,7 @@ from pydantic import BaseModel, ConfigDict
 from pipecat.clocks.base_clock import BaseClock
 from pipecat.clocks.system_clock import SystemClock
 from pipecat.frames.frames import (
+    BotSpeakingFrame,
     CancelFrame,
     CancelTaskFrame,
     EndFrame,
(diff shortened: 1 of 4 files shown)


src/pipecat/pipeline/task.py
```diff
@@ -133,12 +136,27 @@ async def on_frame_reached_upstream(task, frame):
...
         params: Configuration parameters for the pipeline.
         observers: List of observers for monitoring pipeline execution.
         clock: Clock implementation for timing operations.
         check_dangling_tasks: Whether to check for processors' tasks finishing properly.
+        idle_timeout_secs: Timeout (in seconds) to consider pipeline idle or
+            None. If a pipeline is idle the pipeline task will be cancelled
+            automatically.
+        idle_timeout_frames: A tuple with the frames that should trigger an idle
+            timeout if not received withing `idle_timeout_seconds`.
+        cancel_on_idle_timeout: Whther the pipeline task should be cancelled if
```
markbackman, line 158:
```suggestion
        cancel_on_idle_timeout: Whether the pipeline task should be cancelled if
```

markbackman (approved):
LGTM!

I tested to make sure that the timer is getting reset with activity and that the pipeline correctly shuts down on timeout. I see it working as expected. Code looks good too. Nice! 🙌 

Tiny typo to fix, that's it.

src/pipecat/pipeline/task.py
aconchillo: Fixed!

filipi87 (approved):
Makes sense.

PR #1409, opened 2025-03-20
SegmentedSTTService: use VAD events to detect valid audio

aconchillo:

This PR fixes  a `SegmentedSTTService` issue that was causing audio to be sent prematurely to the STT service. Instead of analyzing the volume in this service we rely on VAD events which use both VAD and volume.


kwindla (approved):
lgtm

src/pipecat/services/ai_services.py
```diff
@@ -859,79 +860,64 @@ async def process_frame(self, frame: Frame, direction: FrameDirection):
...
+        wav.close()
+        content.seek(0)
+
+        await self.process_generator(self.run_stt(content.read()))
+
+        # Start clean.
+        self._audio_buffer.clear()
+
+    async def process_audio_frame(self, frame: AudioRawFrame, direction: FrameDirection):
+        # If the user is speaking the audio buffer will keep growin.
```
markbackman, line 914:
```suggestion
        # If the user is speaking the audio buffer will keep growing.
```

markbackman (approved):
LGTM! Nice improvement 🔥

PR #1413, opened 2025-03-20
LLMUserContextAggregator: fix emulated user started/stopped speaking issues

aconchillo:

This PR ensures that emulated user started/stopped frames are only used if there really was no VAD. Before this was based on timeouts which was error prone. For example, if we had real VAD events but the STT service took time generating the transcription we would push the emulated frames causing an unnecessary interruption.

Now, we don't use timeouts any more, instead we know that if we receive VAD events most likely we need a transcription so we just wait for it. And, in the cases we got a transcription and there were no VAD events, that's the only time when we trigger the emulated frames.


CHANGELOG.md
```diff
@@ -216,6 +219,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Other
 
+- Added a new Ultravox example
+  `examples/foundational/07u-interruptible-ultravox.py`.
```
markbackman, line 223: To save you rebasing pain, I caught this too and numbered them as:

- 07u for neuophonic
- 07v for ultravox

markbackman (approved):
This is a nice improvement. I think the logic makes sense. It's nice to eliminate the timeout value since that was a bit arbitrary.

PR #1420, opened 2025-03-20
Fixed ultravox service bugs

milo157:
Fixed the following issues:
- Not passing frames to Cartesia
- Not processing audio correctly


src/pipecat/services/ultravox.py
```diff
@@ -362,18 +368,20 @@ async def _process_audio_buffer(self) -> AsyncGenerator[Frame, None]:
...
-                        yield ErrorFrame("Empty transcription result")
+                    sentences = []
+                    current = ""
+                    for char in full_response.strip():
+                        current += char
+                        if char in '.!?':
+                            sentences.append(current.strip())
+                            current = ""
+                    if current.strip():  # Add any remaining text
+                        sentences.append(current.strip())
```
aconchillo, line 380: Probably better to use `match_endofsentence()` otherwise you my cut things like `Dr.` or `Ms.` or `3.14`. It would be something like (untested):

```
current_pos = 0
full_response_end = len(full_response)
while len(ull_response) > 0 :
    index = match_endofsentence(full_response)
    if index == 0:
        break
    sentences.append(full_response[:index])
    full_response = full_response[index:]
```

src/pipecat/services/ultravox.py
```diff
@@ -362,18 +368,20 @@ async def _process_audio_buffer(self) -> AsyncGenerator[Frame, None]:
...
+
+                    for sentence in sentences:
+                        text_frame = TextFrame(text=sentence)
```
aconchillo, line 383: I believe this should be a `TranscriptionFrame`

aconchillo: Oh, sorry, this needs to be `LLMTextFrame`. Also, you need to add this:

```
if sentences:
    yield LLMFullResponseStartFrame()

for sentence in sentences:
     text_frame = LLMTextFrame(text=sentence)
    yield text_frame

if sentences:
    yield LLMFullResponseEndFrame()
```

aconchillo: Thinking more about it, I believe we probably don't need all the new code and we just need the following:

```
full_response = full_response.strip()
if full_response:
    yield LLMFullResponseStartFrame()
    yield LLMTextFrame(text=full_response)
    yield LLMFullResponseEndFrame()
```

aconchillo (approved):

PR #1423, opened 2025-03-21
ElevenLabs: add support for a sample rate of 8000

aconchillo:

ElevenLabs has added PCM 8000, so no need to add ulaw support.


markbackman (approved):
🎉

Issue #1428, opened 2025-03-21
[OpenAI] LLM Context - `set_tools` has a typing issue?


chongzluong:
I'm looking at the `set_tools` function within the `OpenAILLMContext` class [here](https://github.com/pipecat-ai/pipecat/blob/main/src/pipecat/processors/aggregators/openai_llm_context.py#L159). It seems to suggest that `tools` can be `ToolsSchema`, but this errors out for me because the `ToolsSchema` class does not support `len(tools)`.

filipi87:
Hi @chongzluong, thank you for reporting this.

I was able to reproduce the issue, and we will be pushing a fix soon.

In the meantime, you can use `ToolsSchema` by passing the `tools` in the constructor when creating the `OpenAILLMContext`.

PR #1429, opened 2025-03-21
Only checking the length if tools is a list.

filipi87:
Only checking the length if tools is a list.

This PR fixes https://github.com/pipecat-ai/pipecat/issues/1428


markbackman (approved):
LGTM! Thanks for the quick fix.

PR #1430, opened 2025-03-21
[WIP] Add asyncio lock to OpenAI TTS

richtermb:
When testing out OpenAI's new TTS model family, I ran into an issue where each consecutive sentence would terminate the previous one-- there was no synchronization between utterances.

Whether `aggregate_sentences` was enabled or not, the issue persisted. By using a lock along with a invocation of `asyncio.sleep`, we can make sure that there is no overlap of audio and temporal ordering is preserved.

### DIFF
--- a/src/pipecat/services/openai.py
+++ b/src/pipecat/services/openai.py
@@ -4,9 +4,11 @@
 # SPDX-License-Identifier: BSD 2-Clause License
 #
 
+import asyncio
 import base64
 import io
 import json
+import time
 from dataclasses import dataclass
 from typing import Any, AsyncGenerator, Dict, List, Literal, Mapping, Optional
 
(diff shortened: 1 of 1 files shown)


markbackman:
@richtermb I'm not seeing this in testing with `OpenAITTSService` using 0.0.60. Would you mind trying the [07g](https://github.com/daily-co/pipecat-march-openai/pull/11) example and let me know if you see it there too?

If you can't hit it in 07g, do you mind sharing a repro case for this?

markbackman:
@richtermb I'm just checking in one more time before closing this out.

markbackman:
Closing out this PR due to inactivity. Feel free to reopen if I closed prematurely.

PR #1434, opened 2025-03-23
examples: add foundational 07x-interruptible-local.py

aconchillo:

Note that this might suffer from not having echo cancellation.


markbackman (approved):
LGTM

PR #1442, opened 2025-03-24
LLMAssistantContextAggregator: create a task to run on_context_updated

aconchillo:

This is an issue that surfaced with Flows. The problem is that `on_context_updated` is executed in the internal Pipecat input queue task and if it blocks for whatever reason, the whole pipeline will block.

This is the PR that made it block https://github.com/pipecat-ai/pipecat-flows/pull/107

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- Fixed a `GoogleAssistantContextAggregator` issue where function calls
+  placeholders where not being updated when then function call result was
+  different from a string.
+
 - Fixed an issue that would cause `LLMAssistantContextAggregator` to block
   processing more frames while processing a function call result.
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- Fixed an issue that would cause `LLMAssistantContextAggregator` to block
+  processing more frames while processing a function call result.
+
 - Fixed an issue where the `RTVIObserver` would report two bot started and
   stopped speaking events for each bot turn.
 
--- a/src/pipecat/frames/frames.py
+++ b/src/pipecat/frames/frames.py
@@ -384,7 +384,7 @@ class FunctionCallResultFrame(DataFrame):
 
     function_name: str
     tool_call_id: str
-    arguments: str
+    arguments: Any
     result: Any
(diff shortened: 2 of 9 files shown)


aconchillo:
I'm not sure about this one yet. Moving to draft.

kompfner:
In general, it makes sense to me that whenever the pipeline makes an asynchronous call out of "pipeline processing land" into "user land", it should be done in another "thread" to not block processing, as we have no control over what the user might want to do in that asynchronous call.

The only question is whether pipecat flows logic expects that no additional frames will be processed until `on_context_updated()` (which actually triggers the work of setting the next node via `setNode()`) finishes...but it seems like if that expectation exists, it should be enforced on the pipecat flows side of the wall.

kompfner:
Btw, I tested this change w/the the new "function" action type and it _still_ hangs indefinitely. So this doesn't solve the problem (at least not fully).

kompfner (approved):

kompfner:
> Btw, I tested this change w/the the new "function" action type and it _still_ hangs indefinitely. So this doesn't solve the problem (at least not fully).

It now seems to work for me 👍

kompfner (changes requested):
Retracting my approval. After more testing, I'm seeing conversation state with the bot get entirely screwy...

kompfner:
> Retracting my approval. After more testing, I'm seeing conversation state with the bot get entirely screwy...

Seems like the LLM continues behaving as if it were in the previous state, as if the function call result never came back and influenced it...

kompfner:
> > Retracting my approval. After more testing, I'm seeing conversation state with the bot get entirely screwy...
> 
> Seems like the LLM continues behaving as if it were in the previous state, as if the function call result never came back and influenced it...

Hmm, even backing out this change (and the corresponding change in https://github.com/pipecat-ai/pipecat-flows/pull/102 that depended on it) things are sounding really screwy/broken for me.

I have a theory of what's going on—it has to do with the change to "task-ify" LLM function calls. I previously had a function call that immediately responded with an error status, but now this is what's being added to the LLM context:

Screenshot 2025-03-25 at 1 16 43 PM

kompfner:
> it has to do with the change to "task-ify" LLM function calls

Turns out this is an issue specific to the Gemini pipecat service—or at least it does not happen with OpenAI.

Anyway, this issue is not related to this PR, so re-adding my approval.

kompfner (approved):

aconchillo:
> > it has to do with the change to "task-ify" LLM function calls
> 
> Turns out this is an issue specific to the Gemini pipecat service—or at least it does not happen with OpenAI.
> 
> Anyway, this issue is not related to this PR, so re-adding my approval.

The `IN_PROGRESS` issues is also fixed in this PR.

PR #1443, opened 2025-03-24
Handle missing token counts in AsyncAnthropicBedrock client properly

adithyaxx:
When a custom `AsyncAnthropicBedrock` client is passed to `AnthropicLLMService`, the attributes `cache_creation_input_tokens` and `cache_read_input_tokens` in usage information are `None`. This was not handled properly, resulting in: `anthropic cache creation unsupported operand type(s) for +=: 'int' and 'NoneType'`. 

This PR ensures that None values are handled safely.


markbackman (approved):
LGTM. Thank you!

PR #1447, opened 2025-03-25
feature/support instructions in OpenAITTSService

nicougou:
This PR brings two changes: 

### base_url support in OpenAITTSService and OpenAIImageGenService
While the env var `OPENAI_BASE_URL` is supported, that allows more granular configuration.

### Support `instructions` in OpenAITTSService
There is support for `instructions` in the realtime api, but not in the TTS service.
I used the client `extra_body` to achieve that, maybe we could also accept a whole `extra_body` parameter in OpenAITTSService, but this first draft is the minimum change necessary to get the instructions.

### DIFF
--- a/src/pipecat/services/openai.py
+++ b/src/pipecat/services/openai.py
@@ -520,7 +520,9 @@ class OpenAITTSService(TTSService):
         self.set_model_name(model)
         self.set_voice(voice)
         self._instructions = instructions
-        self._client = AsyncOpenAI(api_key=api_key, base_url=base_url, websocket_base_url=websocket_base_url)
+        self._client = AsyncOpenAI(
+            api_key=api_key, base_url=base_url, websocket_base_url=websocket_base_url
+        )
 
     def can_generate_metrics(self) -> bool:
         return True
--- a/src/pipecat/services/openai.py
+++ b/src/pipecat/services/openai.py
@@ -503,7 +503,6 @@ class OpenAITTSService(TTSService):
         *,
         api_key: Optional[str] = None,
         base_url: Optional[str] = None,
-        websocket_base_url: Optional[str] = None,
         voice: str = "alloy",
         model: str = "gpt-4o-mini-tts",
         sample_rate: Optional[int] = None,
@@ -520,9 +519,7 @@ class OpenAITTSService(TTSService):
         self.set_model_name(model)
         self.set_voice(voice)
         self._instructions = instructions
-        self._client = AsyncOpenAI(
-            api_key=api_key, base_url=base_url, websocket_base_url=websocket_base_url
-        )
+        self._client = AsyncOpenAI(api_key=api_key, base_url=base_url)
 
     def can_generate_metrics(self) -> bool:
         return True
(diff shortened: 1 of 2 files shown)


src/pipecat/services/openai.py
```diff
@@ -501,9 +502,12 @@ def __init__(
         self,
         *,
         api_key: Optional[str] = None,
+        base_url: Optional[str] = None,
+        websocket_base_url: Optional[str] = None,
```
markbackman, line 506: This service is HTTP based. I think we should remove the `websocket_base_url`.

src/pipecat/services/openai.py
```diff
@@ -538,11 +544,17 @@ async def run_tts(self, text: str) -> AsyncGenerator[Frame, None]:
         try:
             await self.start_ttfb_metrics()
 
+            # Setup extra body parameters
+            extra_body = {}
+            if self._instructions:
```
markbackman, line 549: `instructions` are only supported for the new `gpt-4o-mini-tts` model. We should not apply the instructions when the model is `tts-1` or `tts-1-hd` and issue a warning.

markbackman:
Out of curiosity, why add the `base_url` to both of these services?

src/pipecat/services/openai.py
nicougou: @markbackman good point, removed

src/pipecat/services/openai.py
nicougou: @markbackman you are good hardcoding such logic ?
I was working under the idea of letting the user manage the models compatibilities and keep this implementation agnostic, but i can add this 👍

nicougou:
> Thanks for the submission!
> 
> Out of curiosity, why add the `base_url` to both of these services?

@markbackman main reason is to be able to proxy our requests to [Cloudflare AI Gateway](https://developers.cloudflare.com/ai-gateway/) for monitoring/security purposes.

I only focused on OpenAI driven by my selfish needs so far, but to make it better al global pass on all providers supporting a base_url would be nice (on-prem deployments, etc)

markbackman:
> > Thanks for the submission!
> > Out of curiosity, why add the `base_url` to both of these services?
> 
> @markbackman main reason is to be able to proxy our requests to [Cloudflare AI Gateway](https://developers.cloudflare.com/ai-gateway/) for monitoring/security purposes.
> 
> I only focused on OpenAI driven by my selfish needs so far, but to make it better al global pass on all providers supporting a base_url would be nice (on-prem deployments, etc)

That makes sense. I'm happy to include it but wanted to make sure there was a good use case.

src/pipecat/services/openai.py
markbackman: We've generally defaulted to the API for this type of thing, but it's a judgment call. In this case, it may be confusing enough to add the logic and warnings.

markbackman: If we do add it, perhaps a constant would make sense.

markbackman: Actually, for now, let's just keep it simple and see if people get stuck on it. There will be a clear error returned from the endpoint when using the property.

markbackman:
Oh, one more thing @nicougou. Can you submit a changelog entry?

nicougou:
> LGTM! Thanks for the submission 🙌

@markbackman My pleasure !

> Oh, one more thing @nicougou. Can you submit a changelog entry?

Just did, let me know if the format/wording needs fixing 🙏

PR #1451, opened 2025-03-26
Set tool choice from context aggregator

filipi87:
Created `LLMSetToolChoiceFrame` to modify the llm's `tool_choice` during runtime.


markbackman (approved):
LGTM!

PR #1453, opened 2025-03-26
New GroqTTSService

kwindla:
New GroqTTSService.

Groq docs: https://console.groq.com/docs/text-to-speech

### DIFF
--- a/src/pipecat/services/groq.py
+++ b/src/pipecat/services/groq.py
@@ -118,6 +118,8 @@ class GroqTTSService(TTSService):
         speed: Optional[float] = 1.0
         seed: Optional[int] = None
 
+    GROQ_SAMPLE_RATE = 48000  # Groq TTS only supports 48kHz sample rate
+
     def __init__(
         self,
         *,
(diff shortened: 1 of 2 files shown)


src/pipecat/services/groq.py
```diff
@@ -5,10 +5,14 @@
 #
 
 
-from typing import Optional
+from typing import AsyncGenerator, Optional
 
+from groq import AsyncGroq
```
markbackman, line 10: Should we load this in a try/except?
```
try:
    from groq import AsyncGroq
    logger.error(
        "In order to use Groq, you need to `pip install pipecat-ai[groq]`. Also, set a `GROQ_API_KEY` environment variable."
    )
```

src/pipecat/services/groq.py
```diff
@@ -98,3 +102,62 @@ async def _transcribe(self, audio: bytes) -> Transcription:
...
+        async for data in response.iter_bytes():
+            if measuring_ttfb:
+                await self.stop_ttfb_metrics()
+                measuring_ttfb = False
+            # remove wav header if present
+            if data.startswith(b"RIFF"):
+                data = data[44:]
+                if len(data) == 0:
+                    continue
+            yield TTSAudioRawFrame(data, 48000, 1)
```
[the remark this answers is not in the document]
markbackman: The default for Pipecat is 24khz. Is 48khz the only rate accepted?

src/pipecat/services/groq.py
kwindla: Yes! Fixed.

markbackman (approved):
LGTM!

aconchillo (approved):

Issue #1454, opened 2025-03-26
GoogleTTS is yeilding 2 TTSStoppedFrames per chunk generation


SubhrajitPrusty:
https://github.com/pipecat-ai/pipecat/blob/59fdfe697dff0b9eaf6befe10d30b6b6174bbbdd/src/pipecat/services/google/google.py#L1554-L1561

GoogleTTS is yeilding 2 TTSStoppedFrames per chunk generation. I guess this was missed in fix #1247

markbackman:
You're right! Fixed here: https://github.com/pipecat-ai/pipecat/pull/1459. Thanks for letting us know.

PR #1459, opened 2025-03-26
Fix: GoogleTTSService was emitting two TTSStoppedFrames

markbackman:

Fixes https://github.com/pipecat-ai/pipecat/issues/1454

We mistakenly did yielded a `TTSStoppedFrame` in both the `run_tts` `try` block and the `finally` block. Now, it's just emitted in the `finally` block.

### DIFF
--- a/src/pipecat/services/google/google.py
+++ b/src/pipecat/services/google/google.py
@@ -1343,6 +1343,7 @@ class GoogleVertexLLMService(OpenAILLMService):
         **kwargs,
     ):
         """Initializes the VertexLLMService.
+
         Args:
             credentials (Optional[str]): JSON string of service account credentials.
             credentials_path (Optional[str]): Path to the service account JSON file.
@@ -1366,9 +1367,11 @@ class GoogleVertexLLMService(OpenAILLMService):
     @staticmethod
     def _get_api_token(credentials: Optional[str], credentials_path: Optional[str]) -> str:
         """Retrieves an authentication token using Google service account credentials.
+
         Args:
             credentials (Optional[str]): JSON string of service account credentials.
             credentials_path (Optional[str]): Path to the service account JSON file.
+
         Returns:
             str: OAuth token for API authentication.
         """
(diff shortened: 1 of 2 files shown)


src/pipecat/services/google/google.py
```diff
@@ -1551,8 +1554,6 @@ async def run_tts(self, text: str) -> AsyncGenerator[Frame, None]:
                 yield frame
                 await asyncio.sleep(0)  # Allow other tasks to run
 
-            yield TTSStoppedFrame()
```
kompfner, line 1554: Is it a problem if we send a `TTSStoppedFrame` in `finally` after an error that occurred before we ever sent a `TTSStartedFrame`? It would mean a mismatch—a stop without a corresponding start.

kompfner: (Not sure if anyone is basing any bookkeeping off of those start/stop events, but just wanted to raise the issue in case it matters)

kompfner (approved):
Pre-approving, in the case the start/stop mismatch problem isn't an issue. 

Otherwise, we need to keep track of whether we've gotten as far as sending a start frame before deciding whether we need to send a stop frame in the `finally`.

src/pipecat/services/google/google.py
markbackman: Perhaps a question for @aconchillo. This aligns with a change he made about a month ago. Here's the same code in Cartesia: https://github.com/pipecat-ai/pipecat/blob/main/src/pipecat/services/cartesia.py#L400-L402.

aconchillo: I think @kompfner is right. We should probably put `yield TTSStartedFrame()` at the very top before any exception can happen. Worst case, we will have `TTSStartedFrame` and `TTSStoppedFrame` without any audio frame in between which I think it should be safe.

markbackman: Ok, picking the other TTSStopedFrame.

I'll change this in other services to match then...

kompfner: > Ok, picking the other TTSStopedFrame.

I don't think this is quite what @aconchillo was suggesting. Just picking the other `TTSStopedFrame` creates the possibility of the start occurring without the corresponding stop. My understanding of the suggestion is that you make the start frame occur at the _very top_ of the `try` block, before anything else happens.

PR #1460, opened 2025-03-26
segmented tts ignore emulated frames

aconchillo:

This PR fixes an issue with SegmentedTTSServices. These services use VAD events to send audio to the services. The problem is that we can also generate emulated VAD events and those should be ignored.

### DIFF
--- a/src/pipecat/processors/aggregators/llm_response.py
+++ b/src/pipecat/processors/aggregators/llm_response.py
@@ -249,7 +249,7 @@ class LLMUserContextAggregator(LLMContextResponseAggregator):
         self._waiting_for_aggregation = False
 
     async def handle_aggregation(self, aggregation: str):
-        self._context.add_message({"role": self.role, "content": self._aggregation})
+        self._context.add_message({"role": self.role, "content": aggregation})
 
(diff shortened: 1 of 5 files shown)


src/pipecat/processors/aggregators/llm_response.py
```diff
@@ -308,10 +310,16 @@ async def _stop(self, frame: EndFrame):
...
 
-    async def _handle_user_started_speaking(self, _: UserStartedSpeakingFrame):
+    async def _handle_user_started_speaking(self, frame: UserStartedSpeakingFrame):
         self._user_speaking = True
         self._waiting_for_aggregation = True
 
+        # If we get a non-emulated UserStartedSpeakingFrame but we are in the
+        # middle of emulating VAD, let's stop emulating VAD (i.e. don't send the
+        # EmulateUserStoppedSpeakingFrame).
+        if not frame.emulated and self._emulating_vad:
```
markbackman, line 320: This is just to avoid two EmulatedUserStartedSpeakingFrames being emitted, one after another?

markbackman (approved):
LGTM.

PR #1465, opened 2025-03-28
Tavus / Deepgram TTS compatibility

roey-priel:
Previously, DeepgramTTS service was sending a `TTSStoppedFrame` frame right after a `TTSAudioRawFrame`, which caused choppiness in the audio sent to the Tavus room. This (along with a reduction in the batch size) resolves the issue

### DIFF
--- a/src/pipecat/services/deepgram.py
+++ b/src/pipecat/services/deepgram.py
@@ -102,10 +102,10 @@ class DeepgramTTSService(TTSService):
                 await self.stop_ttfb_metrics()
                 chunk = audio_buffer.read(chunk_size)
                 if not chunk:
-                    yield TTSStoppedFrame()
                     break
                 frame = TTSAudioRawFrame(audio=chunk, sample_rate=self.sample_rate, num_channels=1)
                 yield frame
+            yield TTSStoppedFrame()
 
         except Exception as e:
             logger.exception(f"{self} exception: {e}")


src/pipecat/services/deepgram.py
```diff
@@ -97,17 +97,16 @@ async def run_tts(self, text: str) -> AsyncGenerator[Frame, None]:
...
             while True:
                 await self.stop_ttfb_metrics()
                 chunk = audio_buffer.read(chunk_size)
                 if not chunk:
+                    yield TTSStoppedFrame()
                     break
                 frame = TTSAudioRawFrame(audio=chunk, sample_rate=self.sample_rate, num_channels=1)
                 yield frame
 
-                yield TTSStoppedFrame()
```
aconchillo, line 109: Instead of adding it to the break. Just move this out of the loop.

aconchillo (approved):

lucasrothman:
@roey-priel Thanks for the fix, we had disabled Deepgram TTS previously with Tavus. 

We're on a slightly outdated pipecat version but I think there might be similar issue with OpenAI TTS. We wrote our own quick fix (using an asyncio lock that honestly doesn't work that great), but might be something to test/look into and make sure the official pipecat service works. Not a high priority for us as it's a fallback service but just wanted to flag.

PR #1469, opened 2025-03-28
Refactor GladiaSTTService to support addition params

markbackman:

Also, reorganize this service into multiple files. I think this is a direction we should generally be going.

Two deprecations in this change for which I have warnings for both:
1. With the new config params, `language` is superseded by a `language_config`
2. Rather than use a nested InputParams, I'm suggesting that we align (or at least consider) moving the InputParams outside of the main class. This makes using it more expected.

e.g.

Before:
```
from pipecat.services.gladia import GladiaSTTService, LanguageConfig

stt = GladiaSTTService(
    api_key=os.getenv("GLADIA_API_KEY"),
    params=GladiaSTTService.InputParams(
        language_config=LanguageConfig(
            languages=[Language.EN],
        )
    ),
)
```

After:
```
from pipecat.services.gladia import GladiaInputParams, GladiaSTTService, LanguageConfig

stt = GladiaSTTService(
    api_key=os.getenv("GLADIA_API_KEY"),
    params=GladiaInputParams(
        language_config=LanguageConfig(
            languages=[Language.EN],
        )
    ),
)
```

### DIFF
--- a/src/pipecat/services/gladia.py
+++ b/src/pipecat/services/gladia.py
@@ -6,7 +6,8 @@
 
 import base64
 import json
-from typing import AsyncGenerator, Optional
+import warnings
+from typing import Any, AsyncGenerator, Dict, List, Optional, Union
 
 import aiohttp
 from loguru import logger
@@ -35,6 +36,14 @@ except ModuleNotFoundError as e:
 
 
 def language_to_gladia_language(language: Language) -> Optional[str]:
+    """Convert a Language enum to Gladia's language code format.
+
+    Args:
+        language: The Language enum value to convert
+
+    Returns:
+        The Gladia language code string or None if not supported
+    """
     BASE_LANGUAGES = {
         Language.AF: "af",
         Language.AM: "am",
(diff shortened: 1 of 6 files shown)


src/pipecat/services/gladia/config.py
```diff
@@ -0,0 +1,165 @@
...
+        realtime_processing: Real-time processing features
+        messages_config: WebSocket message filtering options
+    """
+
+    encoding: Optional[str] = "wav/pcm"
+    bit_depth: Optional[int] = 16
+    channels: Optional[int] = 1
+    custom_metadata: Optional[Dict[str, Any]] = None
+    endpointing: Optional[float] = None
+    maximum_duration_without_endpointing: Optional[int] = 10
```
markbackman, line 160: I'm leaving this set to `10`, which prevents Gladia from prematurely splitting the content. the range for this is 5-60 seconds. We'll have to experiment with this value to see what works best.

aconchillo:
This looks great! Only one thing. We should keep the old file and expose `GladiaSTTService` from the new location.

aconchillo:
> This looks great! Only one thing. We should keep the old file and expose `GladiaSTTService` from the new location.

With a deprecation warning. Same thing we had for VAD.

https://github.com/pipecat-ai/pipecat/blob/v0.0.58/src/pipecat/vad/silero.py

src/pipecat/services/gladia/__init__.py
```diff
@@ -0,0 +1,31 @@
...
+    "GladiaInputParams",
+    "LanguageConfig",
+    "PreProcessingConfig",
+    "CustomVocabularyItem",
+    "CustomVocabularyConfig",
+    "CustomSpellingConfig",
+    "TranslationConfig",
+    "RealtimeProcessingConfig",
+    "MessagesConfig",
+]
```
aconchillo, line 31: Note sure we want to import everything in here. This will increase load time. I think it might be better to load from the module itself.

aconchillo: Actually, we need, to not break backwards compatibility. I'm wondering if we should force everyone to use the specific module instead and add a warning here that these are deprecated.

src/pipecat/services/gladia.py
```diff
@@ -4,250 +4,12 @@
...
-
-from pipecat.frames.frames import (
-    CancelFrame,
-    EndFrame,
-    Frame,
-    InterimTranscriptionFrame,
-    StartFrame,
-    TranscriptionFrame,
+warnings.warn(
+    "Module `pipecat.services.gladia` is deprecated, use `pipecat.services.gladia.stt` instead",
```
markbackman, line 10: I'm pointing to the service file for now because the module export is `pipecat.service.gladia` and the old file location was `pipecat.service.gladia`. This, at least, differentiates the two.

src/pipecat/services/gladia/__init__.py
```diff
@@ -0,0 +1,10 @@
+#
+# Copyright (c) 2024–2025, Daily
+#
+# SPDX-License-Identifier: BSD 2-Clause License
+#
+
+from pipecat.services.gladia.stt import GladiaSTTService
+
+# Export only the main service class
+__all__ = ["GladiaSTTService"]
```
markbackman, line 10: Exporting only the service in __init__.py.

src/pipecat/services/gladia/config.py
```diff
@@ -0,0 +1,177 @@
...
+# SPDX-License-Identifier: BSD 2-Clause License
+#
+
+from typing import Any, Dict, List, Optional, Union
+
+from pydantic import BaseModel
+
+from pipecat.transcriptions.language import Language
+
+__all__ = [
```
markbackman, line 13: Exporting the configs so you can `from pipecat.services.gladia.config import *` if you wanted.

markbackman:
@aconchillo I made those updates and renamed the service.py to stt.py to follow the pattern in your other PR.

src/pipecat/services/gladia.py
markbackman: Thinking about this more, it might not be needed because the export from services/gladia/__init__.py is for the GladiaSTTService. So, you'd access it with pipecat.services.gladia, just like you currently do. @aconchillo leaving it like this for now, but I think the deprecation isn't needed.

src/pipecat/services/gladia/config.py
aconchillo: I think that's the default. The `__all__` is used if you only want to export certain things, but in this case I believe we are exporting everything, so `__all__` is not really needed.

src/pipecat/services/gladia/__init__.py
```diff
@@ -0,0 +1,16 @@
...
+import warnings
+
+with warnings.catch_warnings():
+    warnings.simplefilter("always")
+    warnings.warn(
+        "Module `pipecat.services.gladia` is deprecated, use `pipecat.services.gladia.stt` instead",
+        DeprecationWarning,
+    )
+
+from .stt import *
```
aconchillo, line 16: Unfortunately, after testing my PR I realized this doesn't really work, because you will also get deprecation event if we import `gladia.stt`. I've updated my PR with a new approach that works well, that is, you can import `from pipecat.services.gladia.stt import ...` (without warning) or `from pipecat.service.gladia import ...` (with warnings).

markbackman: I've updated to match the pattern in #1471. Once you merge that, I'll rebase this one.

Also, I'm following your pattern and moving the language mapping back into the service file since it's not commonly used.

I'm keeping the config.py separate. It's a ton of classes and this may be a/the direction we go when moving InputParams outside of the service class.

markbackman:
@aconchillo this has been rebased. I updated __init__.py to use the new DeprecatedModuleProxy location. Also, I resolved all comments. I think this one is ready for review.

filipi87 (approved):
LGTM!

PR #1471, opened 2025-03-28
services: restructure services into folders

aconchillo:


aconchillo:
Unit Tests are passing. I will now run all examples to make sure they work with the deprecated warnings. Once they all work I will create a new commit to fix the warnings.

markbackman:
This looks really good. I think we also need a pass on the examples to use the correct modules. Really liking this change! I can help to test.

aconchillo:
This is ready. I ran all test twice without changes (to see deprecated messages and make sure they were still working) and with changes to make sure the new packages work.

We should merge this ASAP before we keep making changes to the old files.

src/pipecat/services/base_openai.py
```diff
@@ -0,0 +1,296 @@
+#
```
markbackman, line 1: I see this now lives outside of a service directory. Is there a reason to not keep it with openai? I see that openai.llm and ollama.llm use it. (Why does ollama.llm use BaseOpenAILLMService and not OpenAILLMService like the other subclasses?)

src/pipecat/services/deprecated.py
```diff
@@ -0,0 +1,32 @@
+#
```
markbackman, line 1: When running a service with a deprecated import path, I see three instances of the deprecation warning:
```
REDACTED/pipecat/src/pipecat/services/deprecated.py:15: DeprecationWarning: Module `pipecat.services.cartesia` is deprecated, use `pipecat.services.cartesia.tts` instead
  warnings.warn(
REDACTED/pipecat/src/pipecat/services/deprecated.py:15: DeprecationWarning: Module `pipecat.services.cartesia` is deprecated, use `pipecat.services.cartesia.tts` instead
  warnings.warn(
REDACTED/pipecat/src/pipecat/services/deprecated.py:15: DeprecationWarning: Module `pipecat.services.cartesia` is deprecated, use `pipecat.services.cartesia.tts` instead
  warnings.warn(
```

Instead, we should only show one instance and per module import and point out which file and import is causing the issue. This change to `_warn_deprecated_access` works for me locally:
```
# Track which modules we've already warned about
_warned_modules = set()


def _warn_deprecated_access(globals: Dict[str, Any], attr, old: str, new: str):
    import warnings

    # Only warn once per old->new module pair
    module_key = (old, new)
    if module_key not in _warned_modules:
        with warnings.catch_warnings():
            warnings.warn(
                f"Module `pipecat.services.{old}` is deprecated, use `pipecat.services.{new}` instead.",
                DeprecationWarning,
                stacklevel=3,
            )
        _warned_modules.add(module_key)

    return globals[attr]
```

It returns:
```
REDACTED/pipecat/examples/foundational/07-interruptible.py:21: DeprecationWarning: Module `pipecat.services.cartesia` is deprecated, use `pipecat.services.cartesia.tts` instead
  from pipecat.services.cartesia import CartesiaTTSService
```

markbackman (commented):
This looks really, really good! This organization makes sense.

I just had two comments:
- one question about BaseOpenAILLMService
- one suggestion for improving the deprecation warning

This must have been tedious! Thanks for working through it 🙏

src/pipecat/services/deprecated.py
markbackman: Note: I had to change the `stacklevel` to 3 to catch the warning in the user's code. At 2, it would just return info about the `DeprecatedModuleProxy`.

aconchillo: Ah, nice! I've added this in a new commit. Btw, I ended up moving deprecations to __init__.py instead of deprecated.py, so it's more hidden.

aconchillo:
> This looks really, really good! This organization makes sense.
> 
> I just had two comments:
> 
> * one question about BaseOpenAILLMService
> * one suggestion for improving the deprecation warning
> 
> This must have been tedious! Thanks for working through it 🙏

All done! Yes, a few hours :smile: .

markbackman (approved):
LGTM!

Also, it seems there are a few tests that need to be updated as well, but nothing blocking my approval. This is great!

aconchillo:
> LGTM!
> 
> Can you add a changelog entry?
> 
> Also, it seems there are a few tests that need to be updated as well, but nothing blocking my approval. This is great!

Changelog entry added. Also updated these missing tests. I have left 04, 08 (they are very old and don't even work...one day) and gladia because you have the other PR.

examples/foundational/07-interruptible-vad.py
examples/foundational/07v-interruptible-neuphonic-http.py
examples/foundational/37-mem0.py
examples/twilio-chatbot/client.py
tests/integration/test_integration_unified_function_calling.py
tests/test_context_aggregators.py
tests/test_piper_tts.py

PR #1474, opened 2025-03-29
Changelog entry for mem0 service

kwindla:
Changelog line.


markbackman (approved):
LGTM! I pushed a commit with minor updates, like adding to the README and formatting/organization.

I'll add docs for this shortly.

PR #1478, opened 2025-03-29
Add QwenLLMService

markbackman:

For the Daily team, I shared an API key to test in 1Password.

I tested both `qwen-plus` and `qwen2.5-72b-instruct`. Both are really fast and had accurate function calling in my limited testing.


kwindla (approved):
lgtm

PR #1487, opened 2025-03-31
SmallWebRTCTransport: Improvements to work with mobile

filipi87:
SmallWebRTCTransport: Improvements to work with mobile
- Adding support to `yuvj420p` which is the format that we receive from mobile iOS.
- Improvements on the video transform example to work on mobile.


markbackman (approved):
LGTM

PR #1491, opened 2025-04-01
introducing urgent frames

aconchillo:

### DIFF
--- a/src/pipecat/frames/frames.py
+++ b/src/pipecat/frames/frames.py
@@ -73,8 +73,24 @@ class Frame:
 
 @dataclass
 class SystemFrame(Frame):
-    """System frames are frames that are not internally queued by any of the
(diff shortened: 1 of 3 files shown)


markbackman (commented):
This appears to simplify things by a fair amount, which is great. Before approving, I'd like to do some testing. But, the code change makes sense.

aconchillo:
I've made another change which is that urgent frames will not be interrupted. So, system frames and urgent frames are not interruptible, data and control frames are.

aconchillo:
> It makes a lot of sense. Really nice job. 🔥
> 
> Would it be possible to develop some automated tests so we could see the different kinds of frames working together in a predictable way ?

Yes yes. This is just on going. I'll add unit tests soon.

aconchillo:
Closing in favor of #1776

PR #1499, opened 2025-04-01
BaseOutputTransport: allow setting 10ms output audio chunks

aconchillo:

This allows modifying the default 20ms audio chunks. I've seen people having issues with websockets and 20ms. I believe we should probably also change the default for FastAPI to 40ms. For now, this gives the option to the user to change it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- Added `TransportParams.audio_out_10ms_chunks` parameter to allow controlling
+  the amount of audio being sent by the output transport. It defaults to 2, so
+  20ms audio chunks are sent.
+
 - Added `QwenLLMService` for Qwen integration with an OpenAI-compatible
   interface. Added foundational example `14q-function-calling-qwen.py`.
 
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -9,6 +9,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- Added `TransportParams.audio_out_10ms_chunks` parameter to allow controlling
+  the amount of audio being sent byt the output transport. It defaults to 2, so
```
markbackman, line 13:
```suggestion
  the amount of audio being sent by the output transport. It defaults to 2, so
```

aconchillo: Updated.

markbackman (approved):
I think this is OK, but it's a pretty low level control to expose. Is it not possible to set a universal default for it?

Approving not to block.

aconchillo:
> I think this is OK, but it's a pretty low level control to expose. Is it not possible to set a universal default for it?
> 
> Approving not to block.

Well, 20ms was supposed to be good but it seems to not work well in some cases. :shrug: 

Since this is quite a low-level framework, exposing parameters like this might not be that bad. Hardcoded values might not work well for all environments and/or transports.

PR #1500, opened 2025-04-01
BaseOutputTransport: optimize BotSpeakingFrames

aconchillo:

Sending frames every 20ms has a performance hit. With this change we push BotSpeakingFrame every 200ms, or at every chunk size as before if the chunk size is greater than 200ms (now that it will be configurable).

### DIFF
--- a/src/pipecat/transports/base_output.py
+++ b/src/pipecat/transports/base_output.py
@@ -336,13 +336,22 @@ class BaseOutputTransport(FrameProcessor):
             return without_mixer(BOT_VAD_STOP_SECS)
 
     async def _sink_task_handler(self):
+        # Push a BotSpeakingFrame every 200ms, we don't really need to push it
+        # at every audio chunk. If the audio chunk is bigger than 200ms, push at
+        # every audio chunk.
+        TOTAL_CHUNK_MS = self._params.audio_out_10ms_chunks * 10
+        BOT_SPEAKING_CHUNK_PERIOD = max(int(200 / TOTAL_CHUNK_MS), 1)
+        bot_speaking_counter = 0
         async for frame in self._next_frame():
             # Notify the bot started speaking upstream if necessary and that
             # it's actually speaking.
             if isinstance(frame, TTSAudioRawFrame):
                 await self._bot_started_speaking()
-                await self.push_frame(BotSpeakingFrame())
-                await self.push_frame(BotSpeakingFrame(), FrameDirection.UPSTREAM)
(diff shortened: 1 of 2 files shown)


markbackman:
This change makes sense, but running it causes:
```
2025-04-01 13:33:29.439 | ERROR    | pipecat.utils.asyncio:run_coroutine:113 - DailyOutputTransport#0::_sink_task_handler: unexpected exception: 'DailyParams' object has no attribute 'audio_out_10ms_chunks'
```

src/pipecat/transports/base_output.py
```diff
@@ -335,13 +335,22 @@ async def with_mixer(vad_stop_secs: float) -> AsyncGenerator[Frame, None]:
...
+        BOT_SPEAKING_CHUNK_PERIOD = max(int(200 / TOTAL_CHUNK_MS), 1)
+        bot_speaking_counter = 0
         async for frame in self._next_frame():
             # Notify the bot started speaking upstream if necessary and that
             # it's actually speaking.
             if isinstance(frame, TTSAudioRawFrame):
                 await self._bot_started_speaking()
-                await self.push_frame(BotSpeakingFrame())
-                await self.push_frame(BotSpeakingFrame(), FrameDirection.UPSTREAM)
+                if bot_speaking_counter % BOT_SPEAKING_CHUNK_PERIOD == 0:
```
markbackman, line 349: Won't this check always be true? If so, what's the purpose of it?

markbackman, line 349: Won't this always evaluate to True? If so, what is the purpose of this check?

aconchillo: It will only evaluate to true when the modulo is 0. `BOT_SPEAKING_CHUNK_PERIOD` is initialized as:

```
        TOTAL_CHUNK_MS = self._params.audio_out_10ms_chunks * 10
        BOT_SPEAKING_CHUNK_PERIOD = max(int(200 / TOTAL_CHUNK_MS), 1)
```

So, if  `self._params.audio_out_10ms_chunks` is 2 (the default), `TOTAL_CHUNK_MS` will be 20. Then `BOT_SPEAKING_CHUNK_PERIOD` will be 200/20 which is 5.

So `bot_speaking_counter % BOT_SPEAKING_CHUNK_PERIOD == 0` will be True when `bot_speaking_counter` is 0 or 5.

markbackman (approved):
LGTM

PR #1501, opened 2025-04-01
TranscriptProcessor: send TranscriptionUpdateFrame after interruption

aconchillo:

This makes sure `TranscriptionUpdateFrame` is never discarded when there's an interruption.


markbackman (approved):
LGTM

PR #1508, opened 2025-04-02
LLMAssistantContextAggregator should push BotStoppedSpeakingFrames

markbackman:

In testing the STTMuteFilter with GeminiMultimodalLiveLLMService, I found that the STTMuteFilter would not unmute after the bot stopped speaking. In tracking it down, it appears that the `LLMAssistantContextAggregator` handles the BotStoppedSpeakingFrame but does not push the frame.

This resolves the issue, but I'm not sure if the frame wasn't pushed for a reason.

### DIFF
--- a/src/pipecat/processors/aggregators/llm_response.py
+++ b/src/pipecat/processors/aggregators/llm_response.py
@@ -149,7 +149,8 @@ class BaseLLMResponseAggregator(FrameProcessor):
     @abstractmethod
     def reset(self):
         """Reset the internals of this aggregator. This should not modify the
-        internal messages."""
+        internal messages.
+        """
         pass
 
     @abstractmethod
@@ -446,6 +447,7 @@ class LLMAssistantContextAggregator(LLMContextResponseAggregator):
             await self._handle_user_image_frame(frame)
         elif isinstance(frame, BotStoppedSpeakingFrame):
             await self.push_aggregation()
+            await self.push_frame(frame, direction)
         else:
             await self.push_frame(frame, direction)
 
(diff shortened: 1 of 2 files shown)


src/pipecat/processors/aggregators/llm_response.py
```diff
@@ -446,6 +447,7 @@ async def process_frame(self, frame: Frame, direction: FrameDirection):
             await self._handle_user_image_frame(frame)
         elif isinstance(frame, BotStoppedSpeakingFrame):
             await self.push_aggregation()
+            await self.push_frame(frame, direction)
```
aconchillo, line 450: It's weird this fixes the problem because the assistant aggregator goes after the output transport and the STT mute filter is way up in the pipeline. So, I'm not sure how could that affect the STT mute filter. How does the pipeline look like?

markbackman:
The context aggregator was in the wrong place! Closing this and fixing in our examples.

aconchillo:
> The context aggregator was in the wrong place! Closing this and fixing in our examples.

I was thinking we might actually want this PR because processors down the pipeline might need this frame.

xingweitian:
> > The context aggregator was in the wrong place! Closing this and fixing in our examples.
> 
> I was thinking we might actually want this PR because processors down the pipeline might need this frame.

Agreed, for example a processor that decides whether we should hang up the phone call. I used to rely on BotStoppedSpeakingFrame, after the recent version upgrade, it stopped working. I had to change the processor to check OpenAILLMContextFrame; luckily, we still have it even at the end of the pipeline.

aconchillo (approved):

aconchillo:
LGTM.

PR #1509, opened 2025-04-02
Add new voices to OpenAITTSService

markbackman:

These voices are not publicly supported. Also, the `-latest` model keeps reporting a 503, so I'm switching the example to just use the default.

### DIFF
--- a/examples/foundational/07g-interruptible-openai.py
+++ b/examples/foundational/07g-interruptible-openai.py
@@ -59,7 +59,7 @@ async def main():
             prompt="Expect words related to dogs, such as breed names.",
         )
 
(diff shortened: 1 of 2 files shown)


filipi87 (approved):
LGTM!

duane1024:
Thanks @markbackman! Weirdly, OpenAI documentation refers to 11 voices, but on their docs page, they only list 10 and forget to mention `verse` as a valid voice.  I saw it on openai.fm, but it's also in the python API here: https://github.com/openai/openai-python/blob/main/src/openai/types/audio/speech_create_params.py and works on my installation. Happened to be the one I was testing with, so if you are able to add `verse` as well, I would greatly appreciate!

markbackman:
Added!
https://github.com/pipecat-ai/pipecat/pull/1518

PR #1514, opened 2025-04-03
processors: add ProducerProcessor and ConsumerProcessor

aconchillo:

These processors can be used for advanced uses cases. One example is to send frames from one branch of a `ParallelPipeline` to another branch.


markbackman (approved):
Works great!

PR #1521, opened 2025-04-03
Increase BOT_VAD_STOP_SECS for services with slower speech patterns

markbackman:

In testing with GeminiMultimodalLiveLLMService, I see cases where the speech can take upwards of 0.31 to 0.32 seconds between speech. We want to avoid BotStoppedSpeakingFrame getting emitted when the bot is going to speak again right away. We also want the stopped speaking frame to come relatively soon after the bot stops speaking. For this reason, I'm nudging the value up ever so slightly.

### DIFF
--- a/src/pipecat/transports/base_output.py
+++ b/src/pipecat/transports/base_output.py
@@ -37,7 +37,7 @@ from pipecat.processors.frame_processor import FrameDirection, FrameProcessor
 from pipecat.transports.base_transport import TransportParams
 from pipecat.utils.time import nanoseconds_to_seconds
 
-BOT_VAD_STOP_SECS = 0.3
+BOT_VAD_STOP_SECS = 0.35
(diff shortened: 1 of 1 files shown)


src/pipecat/transports/base_output.py
```diff
@@ -37,7 +37,7 @@
 from pipecat.transports.base_transport import TransportParams
 from pipecat.utils.time import nanoseconds_to_seconds
 
-BOT_VAD_STOP_SECS = 0.3
+BOT_VAD_STOP_SECS = 0.325
```
aconchillo, line 40: Should we do 0.35?

markbackman: Sure. I think that's fine.

filipi87 (approved):
LGTM!

PR #1522, opened 2025-04-03
Pipecat Cloud dynamic transport

chadbailey59:
This PR adds a new transport called `PipecatCloudTransport`. Its goal is to make it easy to use three different transports without having to change your code:

* `SmallWebRTCTransport`, for easy local development
* `DailyTransport`, for local or production use with Pipecat Cloud
* `FastAPIWebsocketTransport`, for connecting a Pipecat Cloud bot to a Twilio phone number

There's an example use case in `examples/pcc-transport-chatbot/server/bot.py`. It's based on the simple-chatbot example. Notable differences include:

* The `main()` function now accepts a single argument: `session_args: PipecatCloudSessionArguments`
* `main()` uses a `PipecatCloudTransport` (obviously)
* The `bot()` function, used by Pipecat Cloud, still takes the same `args: DailySessionArguments` argument, but converts it to a `PipecatCloudSessionArguments` instance before calling main()
* There's a `local_daily()` function, which runs the bot locally but uses the DailyTransport. It's called when you run `python bot.py`
* There's also a `local_webrtc()` function, which is called if you want to run the SmallWebRTC example. To use that example, you need a custom frontend anyway. I've copied the frontend and server from `examples/p2p-webrtc/voice-agent`. You can use it by running `python server.py`, then opening `http://localhost:7860` in your browser.

TODO:

* We'll need to make the SmallWebRTCTransport runner more portable. We can probably fold the server and a basic frontend into the client library.
* This example is missing a few lines of boilerplate websockets code that Pipecat Cloud needs to run the bot with Twilio. I'll clean that up soon.


aconchillo:
I believe this should go into `pipecatcloud` package. It's creating a circular dependency, pipecacloud bot depends on pipecat and pipecat depends on pipecatcloud.

src/pipecat/transports/services/pipecat_cloud.py
```diff
@@ -0,0 +1,313 @@
...
+        await self._transport.stop(frame)
+
+    async def cancel(self, frame):
+        """Cancel the transport."""
+        await self._transport.cancel(frame)
+
+    @property
+    def input(self):
+        """Get the input transport."""
+        return self._transport.input
```
aconchillo, line 308: This should be:

```
    def input(self) -> FrameProcessor:
        return self._transport.input()
```

same for the output below. In the base class input and output are not properties.

src/pipecat/transports/services/pipecat_cloud.py
```diff
@@ -0,0 +1,313 @@
...
+        """Start the transport."""
+        await self._transport.start(frame)
+
+    async def stop(self, frame):
+        """Stop the transport."""
+        await self._transport.stop(frame)
+
+    async def cancel(self, frame):
+        """Cancel the transport."""
+        await self._transport.cancel(frame)
```
aconchillo, line 303: when are start/stop/cancel being used? this is usually added in frame processors, it shouldn't be needed here.

src/pipecat/transports/services/pipecat_cloud.py
```diff
@@ -0,0 +1,313 @@
...
+            **base_params,
+            api_url=self.api_url,
+            api_key=self.api_key,
+            dialin_settings=self.dialin_settings,
+            transcription_enabled=self.transcription_enabled,
+            transcription_settings=self.transcription_settings,
+        )
+
+
+class PipecatCloudSessionArguments:
```
aconchillo, line 109: I don't think we really need this class. In the example we already have a way to differentiate between `local_daily`, `bot`, `local_webrtc`. We can just use the right SessionArguments there. For example:

```
async def local_webrtc(webrtc_connection):
    await main(WebRTCSessionArguments(webrtc_connection=webrtc_connection))
```

And we will still be able to do `isinstance(args,...)`.

src/pipecat/transports/services/pipecat_cloud.py
```diff
@@ -0,0 +1,313 @@
...
+    _EVENT_MAPPINGS = {
+        DailyTransport: {
+            "on_client_connected": "on_first_participant_joined",
+            "on_client_disconnected": "on_participant_left",
+        },
+    }
+
+    def __init__(
+        self,
+        session_args: PipecatCloudSessionArguments,
```
aconchillo, line 193: This would be `SessionArguments`

src/pipecat/transports/services/pipecat_cloud.py
aconchillo: If we were to add more fields to each SessionArguments the `__init__` parameters would not be really nice.

src/pipecat/transports/services/pipecat_cloud.py
```diff
@@ -0,0 +1,313 @@
...
+
+        Args:
+            event_name: The name of the event to handle. Common events:
+                - "on_client_connected": Called when a client connects
+                - "on_client_disconnected": Called when a client disconnects
+                Other event names are passed through to the underlying transport.
+
+        Returns:
+            A decorator that registers the handler function.
+        """
```
aconchillo, line 279: Couldn't we just call:

```
return self._transport.event_handler(event_name)
```

?

src/pipecat/transports/services/pipecat_cloud.py
```diff
@@ -0,0 +1,313 @@
...
+    """
+
+    # Event name mappings for each transport type
+    # Only include events that need to be mapped differently
+    _EVENT_MAPPINGS = {
+        DailyTransport: {
+            "on_client_connected": "on_first_participant_joined",
+            "on_client_disconnected": "on_participant_left",
+        },
+    }
```
aconchillo, line 189: I'm not sure this would work well. Each event has different number of arguments. For example, `on_participant_left` has 2, the participant and the reason. And websocket `on_client_disconnected` only has one. So you would end up with python errors.

aconchillo: One option would be to do this per event. So, this class would have the common events we want to expose and it would register an event handler with the internal transport. Mainly it would acats as a proxy. Something like:

```
self._register_handler("on_client_connected")

@self._transport.event_handler("on_participant_left"):
async def on_participant_left(self, participant, reason):
    self._call_event_handler("on_client_disconnected", participant)
```

That said, the user then will not no what is the first argument. It might change between transports...

src/pipecat/transports/services/pipecat_cloud.py
```diff
@@ -0,0 +1,313 @@
...
+                input_name=input_name,
+                output_name=output_name,
+            )
+        else:
+            raise ValueError(f"Unsupported session arguments type: {type(args)}")
+
+        # Register any handlers that were added before transport creation
+        for event_name, handlers in self._pending_handlers.items():
+            for handler in handlers:
+                self._register_handler(event_name, handler)
```
aconchillo, line 250: I don't think this is needed. The transport is created in the constructor, so it's impossible to have pending handlers because the constructor hasn't finished yet.

src/pipecat/transports/services/pipecat_cloud.py
```diff
@@ -0,0 +1,313 @@
...
+    add_wav_header: bool = False
+    serializer: Optional[FrameSerializer] = None
+    session_timeout: Optional[int] = None
+
+    # DailyParams fields
+    api_url: str = "https://api.daily.co/v1"
+    api_key: str = ""
+    dialin_settings: Optional[DailyDialinSettings] = None
+    transcription_enabled: bool = False
+    transcription_settings: DailyTranscriptionSettings = DailyTranscriptionSettings()
```
aconchillo, line 64: I believe this will lead to issues if we add new fields. we will have to maintain it and add daily fields and fastapi fields or new webrtc fields in this class.

Instead, why don't we create the specific params in each bot function (we already have that), for example, from the bot.py:

```
def update_transport_params(params: TransportParams) -> TransportParams:
            params.audio_out_enabled=True,  # Enable output audio for the bot
            params.camera_out_enabled=True,  # Enable the camera output for the bot
            params.camera_out_width=1024,  # Set the camera output width
            params.camera_out_height=576,  # Set the camera output height
            params.vad_enabled=True,  # Enable VAD to handle user speech
            params.vad_analyzer=SileroVADAnalyzer(),  # Use the Silero VAD analyzer
            params.vad_audio_passthrough=True,  # Pass audio through VAD for user speech to the rest of the pipeline
        return params

async def bot(args: DailySessionArguments):
   ...
   params = update_transport_params(DailyTransportParams(transcription_enabled=True))
   ....
   
async def local_webrtc(webrtc_connection):
    args = WebRTCSessionArguments(webrtc_connection=webrtc_connection)
    params = update_transport_params(TransportParams())
    
    await main(args, params)
```

Well... I guess the last question is, why don't we just create the transport in this function and pass the transport to main?

aconchillo:
See for example this PR: https://github.com/daily-co/pcc-groq-llama/pull/1. Instead of DailyTransport and FastAPIWebsocketTransport we could use the new PipecatCloudTransport but passing the session arguments we need.

To improve the event handlers, I believe we could create new event handlers for only `PipecatCloudTransport`. That is, we register `on_client_connected` and `on_client_disconnected`. And we get those because we register the event handlers on the internal transport. For example:

```
self._register_handler("on_client_connected")

if isinstance(args, DailySessionArguments):
    transport = DailyTransport(....)

    @transport.event_handler("on_participant_joined")
    async def on_participant_joined(transport, participant):
        await self._call_event_handler("on_client_connected")

elif isinstance(args, SessionArguments):
    transport = FastAPIWebsocketTransport(....)

        await self._call_event_handler("on_client_connected")
```

aconchillo:
Summary of my suggestion:

- No need for `PipecatCloudParams` and `PipecatCloudSessionArguments`.
- `PipecatCloudTransport` constructor receives `args: SessionArguments` and `params: TransportParams` and based on args type it chooses the transport as it's doing now. Of course, the transport params need to match the transport.
- Event handlers: 
  - Register new handlers `on_pcc_client_connected` and `on_pcc_client_disconnected` (just a suggestion). Those will be called because we will proxy `on_participant_joined/on_client_connected` and `on_participant_left/on_client_disconnected` from the created transports (see https://github.com/pipecat-ai/pipecat/pull/1522#issuecomment-2781272662).
  - Optionally, we can just proxy the rest of event handlers, but not sure if it makes a lot of sense (see https://github.com/pipecat-ai/pipecat/pull/1522#discussion_r2029906571).
  
With this:
- You make the code simpler in terms of creating the transport which this PR was already doing.
- You still need to use the right SessionArguments and TransportParams subclasses, I think that's fine since we have the starting functions to do that and it's clear. We don't need to create complicated abstractions that need to be maintained.
- We have a couple of common event handlers directly proxied from the transport handlers (fixes current argument issues). All the new event handler code in the PR is not necessary.

So basically this:

```
class PipecatCloudTransport(BaseTransport):

  def __init__(
        self,
        *,
        args: SessionArguments,
        params: TransportParams,
        input_name: Optional[str] = None,
        output_name: Optional[str] = None,
    ):
        super().__init__(input_name=input_name, output_name=output_name)
        
        self._register_handler("on_pcc_client_connected")
        self._register_handler("on_pcc_client_disconnected")

        # Create the appropriate transport based on session arguments type
        if isinstance(args, WebSocketSessionArguments):
            logger.info("Using FastAPIWebsocketTransport")
            self._transport = FastAPIWebsocketTransport(
                args.websocket,
                params=params,
                input_name=input_name,
                output_name=output_name,
            )
            
            @self._transport.event_handler("on_client_connected"):
                await self._call_event_handler("on_pcc_client_connected")
            ....
        elif isinstance(args, DailySessionArguments):
            logger.info("Using DailyTransport")
            self._transport = DailyTransport(
                args.room_url,
                args.token,
                # TODO-CB: Bot name is missing from DailySessionArguments
                "Bot",
                params=params,
                input_name=input_name,
                output_name=output_name,
            )
            
            @self._transport.event_handler("on_participant_joined"):
            async def on_participant_joined(transport, participant):
                await self._call_event_handler("on_pcc_client_connected")

            @self._transport.event_handler("on_participant_left"):
            async def on_participant_left(transport, participant, reason):
                await self._call_event_handler("on_pcc_client_disconnected")
                
        elif isinstance(args, WebRTCSessionArguments):
            logger.info("Using SmallWebRTCTransport")
            self._transport = SmallWebRTCTransport(
                args.webrtc_connection,
                params=params,
                input_name=input_name,
                output_name=output_name,
            )
            
            @self._transport.event_handler("on_client_connected"):
                await self._call_event_handler("on_pcc_client_connected")
            ....
        else:
            raise ValueError(f"Unsupported session arguments type: {type(args)}")
```

examples/p2p-webrtc/voice-agent/bot.py
```diff
@@ -36,9 +40,11 @@
...
 
 async def run_bot(webrtc_connection):
-    pipecat_transport = SmallWebRTCTransport(
-        webrtc_connection=webrtc_connection,
-        params=TransportParams(
+    pipecat_transport = PipecatCloudTransport(
+        session_args=SessionArguments(
+            webrtc_connection=webrtc_connection,
+        ),
+        params=PipecatCloudParams(
```
filipi87, line 47: I think these examples inside `p2p-webrtc` should continue using `SmallWebRTCTransport`, since the intention here was to be didactic and provide an easy way to understand how to use this transport.

Maybe we could create a new folder inside the examples, like `pcc`, or even better, as @aconchillo  suggested a public repo for `pipecat-cloud`, where we would add both the new transport `PipecatCloudTransport`  and all the examples using it ?

examples/pcc-transport-chatbot/client-react/package.json
```diff
@@ -0,0 +1,28 @@
...
+  "scripts": {
+    "dev": "next dev",
+    "build": "next build",
+    "start": "next start",
+    "lint": "next lint"
+  },
+  "dependencies": {
+    "@pipecat-ai/client-js": "^0.3.5",
+    "@pipecat-ai/client-react": "^0.3.5",
+    "@pipecat-ai/daily-transport": "^0.3.7",
```
filipi87, line 14: We should probably have something similar for the client as well, a wrapper for all the available transports, like `pcc-transport`.

examples/pcc-transport-chatbot/client-react/src/providers/RTVIProvider.tsx
```diff
@@ -0,0 +1,38 @@
...
+import { PropsWithChildren, useEffect, useState } from 'react';
+
+const MY_CUSTOM_DATA = { foo: 'bar' };
+
+export function RTVIProvider({ children }: PropsWithChildren) {
+  const [client, setClient] = useState<RTVIClient | null>(null);
+
+  useEffect(() => {
+    console.log('Setting up Transport and Client');
+    const transport = new DailyTransport();
```
filipi87, line 15: Create `PipecatCloudTransport` would be nice, so we could use the same example to test all the available transports.

examples/pcc-transport-chatbot/server/bot.py
```diff
@@ -0,0 +1,330 @@
...
+# SPDX-License-Identifier: BSD 2-Clause License
+#
+
+"""OpenAI Bot Implementation.
+
+This module implements a chatbot using OpenAI's GPT-4 model for natural language
+processing. It includes:
+- Real-time audio/video interaction through Daily
+- Animated robot avatar
+- Text-to-speech using ElevenLabs
```
filipi87, line 13: nit: Cartesia.

examples/pcc-transport-chatbot/server/bot.py
```diff
@@ -0,0 +1,330 @@
...
+        await main(pcc_args)
+        logger.info("Bot process completed")
+    except Exception as e:
+        logger.exception(f"Error in bot process: {str(e)}")
+        raise
+
+
+# Local development
+async def local_daily():
+    # TODO-CB: This becomes SmallWebRTCTransport
```
filipi87, line 302: Yeah, having something like Prebuilt, like you suggested in a Slack thread, would be pretty cool.
But for sure, that's something for a future PR. 🙂

examples/pcc-transport-chatbot/server/index.html
filipi87, line 1: The downside of using this interface is that you wouldn't have access to render the video.
Maybe we should already start focusing on creating the `PipecatCloudTransport` for the client as well.

Or at least, make our web client support both `DailyTransport` and `SmallWebRTCTransport`? That way, we'd have a single interface and wouldn't need this workaround.

chadbailey59:
Closing this in favor of the approach in #1544 .

PR #1524, opened 2025-04-03
TTS: Skip generation when there is no text

markbackman:

Most TTS services don't like messages with empty text. This change centralizes the logic in TTSService to only generate audio from text when there is text.


filipi87 (approved):
Makes sense!

PR #1525, opened 2025-04-03
Enable usage of Application Default Credentials in Google services

shaiyon:
This PR adds support for Application Default Credentials (ADC) in Google service implementations (Vertex LLM, Speech-to-Text, and Text-to-Speech). Previously, these services required explicitly providing service account credentials either as a JSON string or file path.

While this approach works, it's not ideal in situations where IAM service accounts are configured at the environment level. This change enables more secure and flexible authentication patterns, especially in production environments:
- GCP environments: Automatically uses the attached service account
- Local development: Uses credentials from `gcloud auth login`

I welcome all feedback, and look forward to contributing more to pipecat!

### DIFF
--- a/src/pipecat/services/google/llm_vertex.py
+++ b/src/pipecat/services/google/llm_vertex.py
@@ -17,6 +17,8 @@ from loguru import logger
 from pipecat.services.openai.llm import OpenAILLMService
 
 try:
+    from google.auth import default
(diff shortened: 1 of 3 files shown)


markbackman:
This looks great. Can you add a changelog entry for this?

markbackman (approved):
LGTM

PR #1528, opened 2025-04-04
Enable Conditional Prosody Tags in Amazon Polly & Default Credentials

adithyaxx:
**Summary**
This PR enhances Amazon Polly integration by enabling support for prosody SSML tags based on the engine type and enabling default credentials.

**Key Changes**

- Enables prosody tags for controlling speech rate, pitch, and volume.
  - Pitch is enabled for standard engine only.
- Adds support for loading credentials from environment variables when not explicitly provided.


aconchillo:
Closing. Already included in #1753

Issue #1530, opened 2025-04-04
SmallWebRTCTransport has choppy audio


markbackman:
Starting in 0.0.62, the SmallWebRTCTransport produces choppy audio, as reported by `arvis8` on Discord. See the video for evidence.

https://github.com/user-attachments/assets/15888d5e-0ba8-430d-9eda-56a536c5b1ac

This can be reproduced on the current `main` (as of April 4, 2025).

filipi87:
Fixed in this PR: https://github.com/pipecat-ai/pipecat/pull/1531

filipi87:
Hi @yoiang, this issue has been fixed in [Pipecat 0.63.0](https://github.com/pipecat-ai/pipecat/releases/tag/v0.0.63). So, if you're using this version or above, you should be good.

You no longer need to specify the `audio_out_10ms_chunks` value. Pipecat's default settings should work just fine. That said, the value you're using, `audio_out_10ms_chunks=2`, is still good. We’re now using `04` by default.

Are you able to reproduce the issue using one of our examples?
- https://github.com/pipecat-ai/pipecat/tree/main/examples/p2p-webrtc

yoiang:
@filipi87 gotcha, maybe it is something else about my environment that is causing this choppiness, thank you for the insight and examples, I'll check them out 🚀

PR #1531, opened 2025-04-04
Fixed SmallWebRTCTransport to support dynamic chunk values.

filipi87:
Fixed SmallWebRTCTransport to support dynamic chunk values.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- Fixed `SmallWebRTCTransport` to support dynamic values for
+  `TransportParams.audio_out_10ms_chunks`. Previously, it only worked with 20ms 
+  chunks.
+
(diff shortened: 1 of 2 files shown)


src/pipecat/transports/base_transport.py
```diff
@@ -31,7 +31,7 @@ class TransportParams(BaseModel):
     audio_out_sample_rate: Optional[int] = None
     audio_out_channels: int = 1
     audio_out_bitrate: int = 96000
-    audio_out_10ms_chunks: int = 4
+    audio_out_10ms_chunks: int = 2
```
markbackman, line 34: We intentionally changed this value to 4 for 40ms chunks as this improves CPU usage. Rather than reverting back, is there a parameter in the WebRTC code that we can change to align with this chunk size?

filipi87: Makes sense. I have refactored the `SmallWebRTCTransport` to support dynamic values.
Thank you Mark.

filipi87:
Fixes https://github.com/pipecat-ai/pipecat/issues/1530

markbackman (approved):
Very nice! Thank you 👏

PR #1533, opened 2025-04-04
Example interoping between SmallWebRTC and Daily

filipi87:
Example interoping between SmallWebRTC and Daily

### DIFF
--- a/examples/p2p-webrtc/daily-interop-bridge/client/typescript/.gitignore
+++ b/examples/p2p-webrtc/daily-interop-bridge/client/typescript/.gitignore
@@ -1,0 +1,24 @@
+# Logs
+logs
+*.log
+npm-debug.log*
+yarn-debug.log*
+yarn-error.log*
+pnpm-debug.log*
+lerna-debug.log*
+
+node_modules
+dist
+dist-ssr
+*.local
+
+# Editor directories and files
+.vscode/*
+!.vscode/extensions.json
+.idea
+.DS_Store
+*.suo
+*.ntvs*
+*.njsproj
+*.sln
+*.sw?
--- a/examples/p2p-webrtc/daily-interop-bridge/client/typescript/README.md
+++ b/examples/p2p-webrtc/daily-interop-bridge/client/typescript/README.md
@@ -1,0 +1,27 @@
+# JavaScript Implementation
+
+Basic implementation using the [Pipecat JavaScript SDK](https://docs.pipecat.ai/client/js/introduction).
+
+## Setup
+
+1. Run the bot server. See the [server README](../../README).
+
+2. Navigate to the `client/typescript` directory:
+
+```bash
+cd client/typescript
+```
+
+3. Install dependencies:
+
+```bash
+npm install
+```
+
+4. Run the client app:
+
+```
+npm run dev
+```
+
+5. Visit http://localhost:5173 in your browser.
--- a/examples/p2p-webrtc/daily-interop-bridge/client/typescript/index.html
+++ b/examples/p2p-webrtc/daily-interop-bridge/client/typescript/index.html
@@ -1,0 +1,68 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+    <meta charset="UTF-8"/>
+    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+    <title>WebRTC demo</title>
+</head>
+<body>
+
+<div class="container">
+    <!-- Settings Bar -->
+    <div class="status-bar">
+        <div class="option">
+            <label>Audio</label>
+            <select id="audio-input">
+                <option value="" selected>Default device</option>
+            </select>
+            <select id="audio-codec">
+                <option value="default" selected>Default codecs</option>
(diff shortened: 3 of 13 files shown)


src/pipecat/transports/base_output.py
```diff
@@ -386,10 +386,13 @@ async def _cancel_camera_task(self):
     async def _draw_image(self, frame: OutputImageRawFrame):
         desired_size = (self._params.camera_out_width, self._params.camera_out_height)
 
+        # TODO: we should refactor in the future to support dynamic resolutions
+        # which is kind of what happens in P2P connections.
+        # We need to add support for that inside the DailyTransport
```
filipi87, line 391: What do you think, is it something we wish to support ?

vr000m: I think there are two back to back webrtc, so the quality or resolution would be determined by the link that is constrained.

filipi87: What I mean is that currently, inside Pipecat, the DailyTransport always works with a fixed resolution and resizes the input if it doesn’t arrive in the expected resolution.

examples/p2p-webrtc/daily-interop-bridge/README.md
```diff
@@ -0,0 +1,63 @@
...
+
+A Pipecat example demonstrating how to interoperate audio and video between `SmallWebRTCTransport` and `DailyTransport`.
+
+## 🚀 Quick Start
+
+### 1️⃣ Start the Bot Server
+
+#### 📂 Navigate to the Server Directory
+```bash
+cd server
```
markbackman, line 11: I think you can remove this one. The directory is flat.

examples/p2p-webrtc/daily-interop-bridge/env.example
```diff
@@ -0,0 +1 @@
+GOOGLE_API_KEY=
```
markbackman, line 1: This needs to be updated with the env vars for this example. Looks like DAILY_API_KEY at a minimum plus `DAILY_SAMPLE_ROOM_URL` as optional.

examples/p2p-webrtc/daily-interop-bridge/README.md
```diff
@@ -0,0 +1,63 @@
...
+   cp env.example .env
+   ```
+   - Add your API keys
+
+#### ▶️ Run the Server
+```bash
+python server.py
+```
+
+### 2️⃣ Connect Using SmallWebRTC Prebuilt UI
```
markbackman, line 38: This needs to be updated to reflect how to connect. Something like:
1. Connect with SmallWebRTCPrebuiltUI
2. Connect with Daily Prebuilt

src/pipecat/transports/base_output.py
```diff
@@ -386,10 +386,13 @@ async def _cancel_camera_task(self):
...
         if frame.size != desired_size:
             image = Image.frombytes(frame.format, frame.size, frame.image)
             resized_image = image.resize(desired_size)
-            logger.warning(f"{frame} does not have the expected size {desired_size}, resizing")
+            # logger.warning(f"{frame} does not have the expected size {desired_size}, resizing")
```
markbackman, line 395: I see this warning all the time. It's really chatty. What causes it and rather than commenting it out, can the issue be resolved?

markbackman (commented):
Looks good, just a few comments about updating the README and one question about the video resolution warning in base_output.py

src/pipecat/transports/base_output.py
filipi87: To resolve this, we need to add support for DailyTransport to handle dynamic resolutions.

Currently, when we create the `camera_device`, we define a fixed `width` and `height`, so all frames written to it must match that resolution.

https://github.com/pipecat-ai/pipecat/blob/235cd9cecc534e757d02bc3e61947de7f3238967/src/pipecat/transports/services/daily.py#L389-L390

We'd need to modify `daily-python` to support this.

So, I think it's out of scope for this PR, and we should probably discuss it with Aleix to get his thoughts.

examples/p2p-webrtc/daily-interop-bridge/README.md
filipi87: Fixed.

examples/p2p-webrtc/daily-interop-bridge/env.example
filipi87: Fixed.

examples/p2p-webrtc/daily-interop-bridge/README.md
filipi87: Yep, I have added both at the Readme. Let me know what do you think.

filipi87:
@markbackman , I believe I have addressed all your comments.
Thank you for reviewing it.

src/pipecat/transports/base_output.py
markbackman: Yup, that makes sense. Thanks for reminding me. I know you mentioned this before.

markbackman (approved):
LGTM

PR #1539, opened 2025-04-08
Small wbertc mute state

filipi87:
In progresss.


src/pipecat/transports/network/webrtc_connection.py
```diff
@@ -305,5 +338,14 @@ def ask_to_renegotiate(self):
 
         self._renegotiation_in_progress = True
         self.send_app_message(
-            {"type": SIGNALLING_TYPE, "message": SignallingMessage.RENEGOTIATE.value}
+            {"type": SIGNALLING_TYPE, "message": RenegotiateMessage().model_dump()}
```
filipi87, line 341: This is going to be a breaking change from our current code, but I think it makes sense to add in case we need more complex messages in the future.

filipi87: Or maybe we should send both, to keep compatibility with the old version ?

markbackman: SmallWebRTCTransport is so new that it should be OK to break. I don't think there are many users yet.

markbackman (approved):
LGTM

PR #1552, opened 2025-04-10
Add image capture to SmallWebRTCTransport

markbackman:

This is needed to support equivalent functionality in the DailyTransport. I'm opting to not include the `capture_participant_video` convenience method, as you can just set `camera_in_enabled=True,` in the transport, but we could definitely add it for parity.

(Also, some docstring linting inflates the number of changes.)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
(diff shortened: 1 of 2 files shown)


filipi87 (approved):
> I'm opting to not include the capture_participant_video convenience method, as you can just set camera_in_enabled=True, in the transport, but we could definitely add it for parity.

I think we can leave this for the near future, where I believe we'll need to add support for screen audio and video tracks, maybe even custom tracks.
For now, this looks great. Thank you for doing this, Mark.

PR #1561, opened 2025-04-11
Fixing voice agent example

filipi87:
Fixing voice agent example to work with Pipecat 0.0.62

### DIFF
--- a/examples/p2p-webrtc/voice-agent/bot.py
+++ b/examples/p2p-webrtc/voice-agent/bot.py
@@ -44,6 +44,8 @@ async def run_bot(webrtc_connection):
             vad_enabled=True,
             vad_analyzer=SileroVADAnalyzer(),
             vad_audio_passthrough=True,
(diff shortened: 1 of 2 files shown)


examples/p2p-webrtc/voice-agent/bot.py
```diff
@@ -44,6 +44,7 @@ async def run_bot(webrtc_connection):
             vad_enabled=True,
             vad_analyzer=SileroVADAnalyzer(),
             vad_audio_passthrough=True,
+            audio_out_10ms_chunks=2,
```
filipi87, line 47: We don't this parameter with a are using the version from main. 
SmallWebRTC now works with any audio chunk size:
 https://github.com/pipecat-ai/pipecat/blob/main/CHANGELOG.md#fixed

markbackman (approved):
LGTM. We can undo this once 0.0.63 is released.

PR #1563, opened 2025-04-11
Add marker file for static type checkers

Bnowako:
I am pretty new to python so I hope this is correct.

Static type checkers are complaining about missing type information for pipecat:
Screenshot 2025-04-11 at 11 01 37

Pipecat supoorts inline types and accordingly to https://typing.python.org/en/latest/spec/distributing.html#packaging-type-information there is missing marker file, which is added in this PR.


markbackman:
Hmm, I don't see this locally. Configuring your local env can be a bit tricky. I don't think this is the fix though. Make sure you have the right python version linked to your interpreter. That might vary based on IDE.

Thanks for the submission, but I'm going to close this. Reach out on Discord if you're hitting issues configuring your local env.

Bnowako:
@markbackman 
Do you use pyright or mypy? pyright needs it. Take a look [here](https://github.com/microsoft/pyright/blob/main/docs/typed-libraries.md#library-interface):
Screenshot 2025-04-17 at 21 22 16

As a reference I found fastapi PR which works well with pyright -> https://github.com/fastapi/fastapi/pull/209

Also did you see [this python spec?](https://typing.python.org/en/latest/spec/distributing.html#packaging-type-information)
Screenshot 2025-04-17 at 21 19 29

It seems like some lagacy thing, but some type checkers needs it, right now It requires workaround.

I know it's a minor thing, but also it's just one file, so I leave it up to you

markbackman:
No, I don't use either.

Getting an opinion from two others on the team: @aconchillo @filipi87.

aconchillo (approved):

aconchillo:
I think this is fine. I don't have any issues with pyright but this seems standard enough that we can include it. Thank you!

aconchillo:
Related PR https://github.com/pipecat-ai/pipecat/pull/1609

Issue #913, opened 2024-12-26
Please reduce the required version of numpy to <2


phamtrung0633:
Numpy~=2.1.3 requirement of pipecat 0.0.52 make it impossible to be used with langchain, is this really necessary ?

ms1design:
It's still to early to adopt `numpy 2+` as on other platforms like `arm` the adoption move slower that on `x86`. This makes `pipecat-ai` less friendly to devlopers.

markbackman:
@aconchillo I'm assigning to you since you made the bump.

aconchillo:
I think it should be OK to move back. I'll make the changes.

ms1design:
@aconchillo I overriden with `numpy<2` and everything works as usual.

markbackman:
We merged this to revert numpy to <2: https://github.com/pipecat-ai/pipecat/pull/997.

ms1design:
Thanks!

PR #1576, opened 2025-04-14
chore(deps): Update dependencies

reneleonhardt:
### Updates
- Update Python dependencies
- Let Dependabot update Python dependencies, Dockerfiles and GitHub Actions

### Documentation
- Installing dev to run `pytest` seems counterintuitive, at least coverage and pytest* should be moved

### Recommendations
- CI should be tested against all supported Python versions (so 3.11-3.13 are missing)
- If the official Dockerfile is needed, it should be documented and modernized (slim, latest versions like Python 3.13 and [OpenSSL 3](https://packages.debian.org/bookworm/openssl), [multi-stage](https://pythonspeed.com/articles/multi-stage-docker-python/#solution-2-virtualenv))


pyproject.toml
```diff
@@ -24,14 +24,14 @@ dependencies = [
     "audioop-lts~=0.2.1; python_version>='3.13'",
     "loguru~=0.7.3",
     "Markdown~=3.7",
-    "numpy~=1.26.4",
-    "Pillow~=11.1.0",
+    "numpy>=1.26.4, <2.5",
```
markbackman, line 27: We tried to make this update a few months ago and it broke a number of folks. I don't think we're ready for this yet.

markbackman: This was the issue: https://github.com/pipecat-ai/pipecat/issues/913

reneleonhardt: Thank you, I didn't search for closed issues, it seems to be an open problem 😅
This has been fixed and released 3 days ago, my PR pulled those versions already:
https://github.com/langchain-ai/langchain/issues/26026

You could copy their new numpy constraints for instance:
https://github.com/langchain-ai/langchain/pull/30714
```
    "numpy>=1.26.2; python_version<'3.13'",
    "numpy>=2.1.0; python_version>='3.13'",
```

markbackman: The issue was about rolling back to numpy < 2, so it was closed 🙂

What is the need for numpy 2? Rather than jump the gun, I'd prefer to be measured about updating without breaking anyone again.

reneleonhardt: No need, I reverted to numpy 1 already, now you know that at least on Python 3.13 langchain is making you no trouble anymore for numpy 2 😄

aconchillo:
Closing. We have updated the packaging system for uv and now it is possible to install newer versions of numpy.

PR #1577, opened 2025-04-14
docs: Fix TrackStopped typo in SimpleChatbot

jamsea:
This was spelled wrong so the event wouldn't come through


markbackman (approved):
Good catch! 🕵️

PR #1579, opened 2025-04-14
Fixed an issue in `SmallWebRTCTransport`

filipi87:
Fixed an issue in `SmallWebRTCTransport` where an error was thrown if the client did not create a video transceiver.


markbackman (approved):
LGTM

PR #1580, opened 2025-04-14
Fixing the voice agent example to always create the video transceiver.

filipi87:
Fixing the voice agent example to always create the video transceiver.


markbackman (approved):
LGTM

PR #1581, opened 2025-04-14
Configuring the voice-agent example to wait for all the ice candidates.

filipi87:
Configuring the voice-agent example to use ice-servers and wait for all the ice candidates.

### DIFF
--- a/examples/p2p-webrtc/voice-agent/README.md
+++ b/examples/p2p-webrtc/voice-agent/README.md
@@ -46,6 +46,46 @@ http://localhost:7860
 
 ---
 
+## WebRTC ICE Servers Configuration
+
+When implementing WebRTC in your project, **STUN** (Session Traversal Utilities for NAT) and **TURN** (Traversal Using Relays around NAT) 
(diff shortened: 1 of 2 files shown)


markbackman:
@filipi87 when I run this example, my connection doesn't start. Is there something I need to change or configure? When I switch back to `main`, it works as expected.

filipi87:
> @filipi87 when I run this example, my connection doesn't start. Is there something I need to change or configure? When I switch back to main, it works as expected.

No, it should just work. I just tested it here, and it's working fine on my end. 

Do you see any errors in the console or on the Python server?

examples/p2p-webrtc/voice-agent/README.md
```diff
@@ -46,6 +46,46 @@ http://localhost:7860
 
 ---
 
+## WebRTC ICE Servers Configuration
```
filipi87, line 49: @markbackman , let me know what you think.

markbackman (approved):
LGTM! Great description. I'm going to add that to the docs too because it's so helpful. Thank you

PR #1588, opened 2025-04-14
LLM aggregator params

aconchillo:

We replace kwargs with actual classes which will make it easier to set parameters and allow users to configure the aggregators.


src/pipecat/services/anthropic/llm.py
```diff
@@ -120,21 +119,19 @@ def create_context_aggregator(
         self,
         context: OpenAILLMContext,
         *,
-        user_kwargs: Mapping[str, Any] = {},
-        assistant_kwargs: Mapping[str, Any] = {},
```
filipi87, line 124: In the changelog, we mention that these parameters are deprecated, but here we are removing them. 

Is that intentional ? I think we should fix either the changelog or here.

src/pipecat/services/gemini_multimodal_live/gemini.py
```diff
@@ -871,21 +881,19 @@ def create_context_aggregator(
         self,
         context: OpenAILLMContext,
         *,
-        user_kwargs: Mapping[str, Any] = {},
-        assistant_kwargs: Mapping[str, Any] = {},
```
filipi87, line 875: Same here.

src/pipecat/services/anthropic/llm.py
aconchillo: Ooops, they are not deprecated. They are just removed. No one has ever used those anyways.

aconchillo: Changelog updated.

filipi87 (approved):
LGTM!

PR #1589, opened 2025-04-14
Introduce new helper type for acting as a liaison between rtvi message…

mattieruth:
…s and the llm

This somewhat parallels the helper system the RTVI clients have.

The best way to describe this new class is to see how it's implemented in the simple chatbot example, but basically, if you want to handle rtvi messages generated from client llm helpers, the server introduces its own llm helper that handles those messages.

Marking this as a draft pull request because i need to test with a service other than openai, but otherwise, I think it's ready and could use feedback for this approach.

### DIFF
--- a/src/pipecat/processors/frameworks/rtvi_helpers/llm.py
+++ b/src/pipecat/processors/frameworks/rtvi_helpers/llm.py
@@ -1,0 +1,210 @@
+import sys
+from enum import Enum
+from typing import Any, Dict, List
+
+from openai._types import NotGiven
+from pydantic import BaseModel
+
+from pipecat.frames.frames import (
+    LLMMessagesAppendFrame,
+    LLMMessagesUpdateFrame,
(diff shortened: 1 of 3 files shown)


src/pipecat/processors/frameworks/rtvi_helpers/llm.py
```diff
@@ -0,0 +1,210 @@
...
+
+class RTVILLMActionType(StrEnum):
+    """Enum for RTVI LLM action types."""
+    APPEND_TO_MESSAGES = "append_to_messages",
+    GET_CONTEXT = "get_context",
+    SET_CONTEXT = "set_context",
+    RUN = "run",
+
+
+class RTVIHelper(BaseModel):
```
mattieruth, line 44: if/when we introduce other types of helpers (tts, stt), I can move this to a separate file

mattieruth:
going a different direction

PR #1592, opened 2025-04-15
Smart turn

filipi87:
Smart turn


src/pipecat/frames/frames.py
```diff
@@ -571,6 +571,15 @@ class EmulateUserStoppedSpeakingFrame(SystemFrame):
     pass
 
 
+@dataclass
+class UserEndOfTurnFrame(SystemFrame):
```
filipi87, line 575: We need to decide this:
- In the same user turn, should we emit multiple `UserStartedSpeaking` and `UserStoppedSpeaking` events, and also include a new event to mark the end of the turn, in this case, `UserEndOfTurnFrame`?
  - If we take this approach, we will need change the LLM services to react to the `UserEndOfTurnFrame` instead of to `UserStoppedSpeaking`.
- Or should we only emit a single `UserStartedSpeaking` and `UserStoppedSpeaking` event ? In that case, do we still need a `UserEndOfTurnFrame` ?

src/pipecat/transports/base_input.py
```diff
@@ -189,7 +201,9 @@ async def _handle_vad(self, audio_frame: InputAudioRawFrame, vad_state: VADState
             frame = None
             if new_vad_state == VADState.SPEAKING:
                 frame = UserStartedSpeakingFrame()
-            elif new_vad_state == VADState.QUIET:
+            # TODO: need to double check if this is the expected behavior
+            # Not triggering the UserStoppedSpeakingFrame if the turn analyser is enabled
+            elif new_vad_state == VADState.QUIET and not self.end_of_turn_analyzer:
```
filipi87, line 206: Same here.
And if we choose this approach, I will need to refactor to only trigger the `UserStartedSpeakingFrame` once.

src/pipecat/frames/frames.py
markbackman: I'm interested in @aconchillo's POV.

To date `UserStoppedSpeakingFrame` signifies that the VAD detects end of speech and that the user's turn had ended. Since the two were one and the same, a single event was sufficient.

With smart-turn, end of speech and end of turn can now occur at different times. Given this, we should consider adding a new event for `UserEndOfTurnFrame`.

If we do this, we'll probably want to add a `UserEndOfTurnFrame` to the non-smart-turn case as well, so our logic doesn't vary too much between smart-turn and non-smart-turn cases.

@aconchillo do you have any insights?

markbackman:
@filipi87 this is very cool!

I tried testing by setting the `SmartTurnParams` to `stop_secs` of 5.0 seconds. In doing this, I noticed two important things:
1. The frame counting mechanism results in the wrong time. I'm seeing the waiting time amount to half of the `stop_secs` time (e.g. 2.5 seconds of waiting instead of 5.0 seconds). I think this is due to differences in the chunk size. Can you take a look to ensure that the timing matches the setting? You might be able to calcuate the actual duration of the chunk instead of just a fixed frame (e.g. `actual_chunk_duration_ms = len(audio_int16) / (self._sample_rate / 1000)`).
2. Using the slower stop_secs time, it's easier to test how well the end of turn logic works. I've noticed that if I break a sentence across two phrases, I see an unexpected result. Here's what I see:
  a) Say "What's the uh...", receive INCOMPLETE
  b) Within the next 5 seconds, say "the capital of Spain?"
  
  I would expect to get a COMPLETE after b), but, I get an INCOMPLETE. I think what's happening (haven't checked) is that we're sending only the most audio chunk. That is, the model sees only "the capital of Spain" and deems that to be incomplete. Instead, after a) is returned as incomplete, I think we want to keep the chunk from a) and continue buffering audio until the next time the user stops speaking. This way, the smart-turn model gets the full input (e.g. "What's the uh... <pause> the capital of Spain?"). I have a hunch that this will result in a COMPLETE response after b) as the full audio input is closer to the training data and is an actual complete thought.

src/pipecat/frames/frames.py
filipi87: > With smart-turn, end of speech and end of turn can now occur at different times. Given this, we should consider adding a new event for UserEndOfTurnFrame.

Exactly, if we take this approach, we will need change the LLM services to react to the `UserEndOfTurnFrame` instead of to `UserStoppedSpeaking`. And probably in a couple of other places in our code.

> If we do this, we'll probably want to add a UserEndOfTurnFrame to the non-smart-turn case as well, so our logic doesn't vary too much between smart-turn and non-smart-turn cases.

Yep. That would help to smooth things.

src/pipecat/audio/turn/base_smart_turn.py
```diff
@@ -0,0 +1,172 @@
...
+
+        # Extract the audio segment
+        segment_audio_chunks = [chunk for _, chunk in audio_buffer[start_index : end_index + 1]]
+        segment_audio = np.concatenate(segment_audio_chunks)
+
+        logger.debug(f"Segment audio chunks after start index: {len(segment_audio)}")
+
+        # Limit maximum duration
+        if len(segment_audio) / self.sample_rate > self._params.max_duration_secs:
+            segment_audio = segment_audio[: int(self._params.max_duration_secs * self.sample_rate)]
```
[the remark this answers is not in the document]
filipi87: Yes, you were right. I have pushed a fix for that. 
Nice catch.

src/pipecat/audio/turn/smart_turn.py
```diff
@@ -0,0 +1,75 @@
...
+                try:
+                    logger.debug("Response JSON:")
+                    logger.debug(response.json())
+                    return response.json()
+                except requests.exceptions.JSONDecodeError:
+                    logger.debug("Response Content (non-JSON):")
+                    logger.debug(response.text)
+            else:
+                logger.debug("Response Content (Error):")
+                logger.debug(response.text)
```
aconchillo, line 66: We probably want all this to be `logger.trace`

src/pipecat/audio/turn/smart_turn.py
```diff
@@ -0,0 +1,75 @@
...
+        np.save(buffer, audio_array)
+        serialized_bytes = buffer.getvalue()
+        logger.debug(f"Serialized size: {len(serialized_bytes)} bytes")
+        return serialized_bytes
+
+    def _send_raw_request(self, data_bytes: bytes):
+        headers = {"Content-Type": "application/octet-stream"}
+        logger.debug(
+            f"Sending {len(data_bytes)} bytes as raw body to {self.remote_smart_turn_url}..."
+        )
```
aconchillo, line 44: Logger trace

src/pipecat/audio/turn/smart_turn.py
```diff
@@ -0,0 +1,75 @@
...
+        # Use a session to reuse connections (keep-alive)
+        self.session = requests.Session()
+        self.session.headers.update({"Connection": "keep-alive"})
+
+    def _serialize_array(self, audio_array: np.ndarray) -> bytes:
+        logger.debug("Serializing NumPy array to bytes...")
+        buffer = io.BytesIO()
+        np.save(buffer, audio_array)
+        serialized_bytes = buffer.getvalue()
+        logger.debug(f"Serialized size: {len(serialized_bytes)} bytes")
```
aconchillo, line 37: Logger trace.

src/pipecat/audio/turn/smart_turn.py
```diff
@@ -0,0 +1,75 @@
...
+        if not self.remote_smart_turn_url:
+            logger.error("REMOTE_SMART_TURN_URL is not set.")
+            raise Exception("REMOTE_SMART_TURN_URL environment variable must be provided.")
+
+        # Use a session to reuse connections (keep-alive)
+        self.session = requests.Session()
+        self.session.headers.update({"Connection": "keep-alive"})
+
+    def _serialize_array(self, audio_array: np.ndarray) -> bytes:
+        logger.debug("Serializing NumPy array to bytes...")
```
aconchillo, line 33: Logger trace

src/pipecat/audio/turn/local_smart_turn.py
```diff
@@ -0,0 +1,81 @@
...
+        #   brew install git-lfs
+        #   # Hugging Face uses LFS to store large model files, including .mlpackage
+        #   git lfs install
+        #   # Clone the repo with the smart_turn_classifier.mlpackage
+        #   git clone https://huggingface.co/pipecat-ai/smart-turn
+        #
+        # Then set the env variable:
+        #   export LOCAL_SMART_TURN_MODEL_PATH=./smart-turn
+        # or add it to your .env file
+        smart_turn_model_path = os.getenv("LOCAL_SMART_TURN_MODEL_PATH")
```
aconchillo, line 46: I'd pass a `model_path` argument instead of defining an environment variable. You can use the environment variable on the test files.

src/pipecat/audio/turn/smart_turn.py
```diff
@@ -0,0 +1,75 @@
...
+import requests
+from loguru import logger
+
+from pipecat.audio.turn.base_smart_turn import BaseSmartTurn
+
+
+class SmartTurnAnalyzer(BaseSmartTurn):
+    def __init__(self, **kwargs):
+        super().__init__(**kwargs)
+        self.remote_smart_turn_url = os.getenv("REMOTE_SMART_TURN_URL")
```
aconchillo, line 22: We should pass this as a `base_url` instead of an environment variable. Use the variable in the test file.

aconchillo: or simply `url`

src/pipecat/transports/base_input.py
```diff
@@ -211,10 +234,23 @@ async def _audio_task_handler(self):
...
+                # If silence exceeds threshold, we are going to receive EndOfTurnState.COMPLETE
+                end_of_turn_state = self._params.end_of_turn_analyzer.append_audio(
+                    frame.audio, is_speech
+                )
+                if end_of_turn_state == EndOfTurnState.COMPLETE:
+                    await self._handle_end_of_turn_complete(end_of_turn_state)
+                # Otherwise we are going to trigger to check if the turn is completed based on the VAD
+                elif vad_state == VADState.QUIET and vad_state != previous_vad_state:
+                    await self._handle_end_of_turn()
+
```
aconchillo, line 253: Can we move this to a function to keep this function smaller:

```
if self._params.end_of_turn_analyzer:
    await self._run_turn_analyzer(frame, vad_state, previous_vad_state)
```

src/pipecat/audio/turn/base_smart_turn.py
```diff
@@ -0,0 +1,176 @@
...
+
+
+# Enum for end-of-turn detection states
+class EndOfTurnState(Enum):
+    COMPLETE = 1
+    INCOMPLETE = 2
+
+
+# Default timing parameters
+STOP_SECS = 1
```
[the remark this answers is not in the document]
filipi87: Done!

src/pipecat/audio/turn/base_smart_turn.py
```diff
@@ -0,0 +1,176 @@
...
+PRE_SPEECH_MS = 0
+MAX_DURATION_SECONDS = 8  # Max allowed segment duration
+USE_ONLY_LAST_VAD_SEGMENT = False
+
+
+class SmartTurnParams(BaseModel):
+    stop_secs: float = STOP_SECS
+    pre_speech_ms: float = PRE_SPEECH_MS
+    max_duration_secs: float = MAX_DURATION_SECONDS
+    use_only_last_vad_segment: bool = USE_ONLY_LAST_VAD_SEGMENT
```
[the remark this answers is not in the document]
filipi87: Done.

src/pipecat/audio/turn/base_smart_turn.py
```diff
@@ -0,0 +1,176 @@
...
+class EndOfTurnState(Enum):
+    COMPLETE = 1
+    INCOMPLETE = 2
+
+
+# Default timing parameters
+STOP_SECS = 1
+PRE_SPEECH_MS = 0
+MAX_DURATION_SECONDS = 8  # Max allowed segment duration
+USE_ONLY_LAST_VAD_SEGMENT = False
```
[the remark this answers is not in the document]
filipi87: Done.

src/pipecat/audio/turn/smart_turn.py
filipi87: Done.

src/pipecat/audio/turn/smart_turn.py
filipi87: Done.

src/pipecat/audio/turn/smart_turn.py
filipi87: Done.

src/pipecat/audio/turn/local_smart_turn.py
filipi87: done.

src/pipecat/audio/turn/smart_turn.py
filipi87: done.

src/pipecat/transports/base_input.py
filipi87: Done.

src/pipecat/transports/base_input.py
```diff
@@ -187,17 +196,49 @@ async def _handle_vad(self, audio_frame: InputAudioRawFrame, vad_state: VADState
...
+            # - Creating the UserStartedSpeakingFrame multiple times
+            can_create_user_frames = (
+                self._params.end_of_turn_analyzer is None
+                or not self._params.end_of_turn_analyzer.speech_triggered
+            )
+            if can_create_user_frames:
+                if new_vad_state == VADState.SPEAKING:
+                    frame = UserStartedSpeakingFrame()
+                elif new_vad_state == VADState.QUIET:
+                    frame = UserStoppedSpeakingFrame()
```
filipi87, line 210: @markbackman , @aconchillo , this is the logic preventing emitting these events if the turn analyser is enabled.

examples/foundational/38-smart-turn.py
```diff
@@ -0,0 +1,111 @@
...
+    )
+
+    stt = DeepgramSTTService(api_key=os.getenv("DEEPGRAM_API_KEY"))
+
+    tts = CartesiaTTSService(
+        api_key=os.getenv("CARTESIA_API_KEY"),
+        voice_id="71a7ad14-091c-4e8e-a314-022ece01c121",  # British Reading Lady
+    )
+
+    llm = OpenAILLMService(api_key=os.getenv("OPENAI_API_KEY"), model="gpt-4o")
```
markbackman, line 53: Removing the model so we don't have to update it later.

```suggestion
    llm = OpenAILLMService(api_key=os.getenv("OPENAI_API_KEY"))
```

examples/foundational/38a-local-smart-turn.py
```diff
@@ -0,0 +1,129 @@
...
+    )
+
+    stt = DeepgramSTTService(api_key=os.getenv("DEEPGRAM_API_KEY"))
+
+    tts = CartesiaTTSService(
+        api_key=os.getenv("CARTESIA_API_KEY"),
+        voice_id="71a7ad14-091c-4e8e-a314-022ece01c121",  # British Reading Lady
+    )
+
+    llm = OpenAILLMService(api_key=os.getenv("OPENAI_API_KEY"), model="gpt-4o")
```
markbackman, line 71: Same here.

```suggestion
    llm = OpenAILLMService(api_key=os.getenv("OPENAI_API_KEY"))
```

examples/foundational/38a-local-smart-turn.py
```diff
@@ -0,0 +1,129 @@
...
+    transport = SmallWebRTCTransport(
+        webrtc_connection=webrtc_connection,
+        params=TransportParams(
+            audio_in_enabled=True,
+            audio_out_enabled=True,
+            vad_enabled=True,
+            vad_analyzer=SileroVADAnalyzer(params=VADParams(stop_secs=0.2)),
+            vad_audio_passthrough=True,
+            end_of_turn_analyzer=LocalCoreMLSmartTurnAnalyzer(
+                smart_turn_model_path=smart_turn_model_path, params=SmartTurnParams(stop_secs=5)
```
markbackman, line 59: Should we just leave the default?

```suggestion
                smart_turn_model_path=smart_turn_model_path, params=SmartTurnParams()
```

filipi87: Yep, I have changed for testing, but we can restore to the default values.

examples/foundational/38-smart-turn.py
```diff
@@ -0,0 +1,111 @@
...
+
+    remote_smart_turn_url = os.getenv("REMOTE_SMART_TURN_URL")
+
+    transport = SmallWebRTCTransport(
+        webrtc_connection=webrtc_connection,
+        params=TransportParams(
+            audio_in_enabled=True,
+            audio_out_enabled=True,
+            vad_enabled=True,
+            vad_analyzer=SileroVADAnalyzer(params=VADParams(stop_secs=0.2)),
```
markbackman, line 40: One tricky part about using this will be that the VADParams value will need to be adjusted to a shorter time, like 0.2 second. Do we want to leave that to developers or can/should we set a different default when the SmartTurnAnalyzer is used? (I don't think the latter idea is that easy to implement and might be surprising to developers if there are conditionally different values. Maybe we should just write a guide about this instead!)

markbackman (approved):
So good! Approving only with a few minor suggestions.

examples/foundational/38-smart-turn.py
filipi87: I think for this first version, we should leave it as is and recommend changing the stop_secs inside VADParams to 0.2.
Once we're really comfortable with it, we can reconsider changing its default value based on whether the SmartTurnAnalyzer is enabled.
If we go this route, maybe the user shouldn't even need to define the vad_analyzer. It could be automatically enabled when SmartTurnAnalyzer is enabled.
In any case, I think that should be implemented in a follow up PR.

src/pipecat/transports/base_input.py
markbackman: I added an Observer for my testing and can confirm that the `UserStartedSpeakingFrame` and `UserStoppedSpeakingFrame` work as expected.

src/pipecat/frames/frames.py
markbackman, line 575: Oh! Just catching this. We should delete this frame as it's unused.

src/pipecat/transports/base_input.py
```diff
@@ -187,17 +196,49 @@ async def _handle_vad(self, audio_frame: InputAudioRawFrame, vad_state: VADState
...
+    async def _handle_end_of_turn(self):
+        if self.end_of_turn_analyzer:
+            state = await self.get_event_loop().run_in_executor(
+                self._executor, self.end_of_turn_analyzer.analyze_end_of_turn
+            )
+            await self._handle_end_of_turn_complete(state)
+
+    async def _handle_end_of_turn_complete(self, state: EndOfTurnState):
+        if state == EndOfTurnState.COMPLETE:
+            await self.push_frame(UserEndOfTurnFrame())
```
markbackman: I guess we should just remove this, too.

examples/foundational/38-smart-turn.py
markbackman: Yup! I agree with all of this. No change right now. I'll make sure to make this clear in the docs, which I'll write right now.

pyproject.toml
```diff
@@ -79,6 +79,8 @@ qwen = []
 rime = [ "websockets~=13.1" ]
 riva = [ "nvidia-riva-client~=2.19.0" ]
 sentry = [ "sentry-sdk~=2.23.1" ]
+local-smart-turn = [ "coremltools>=8.0", "transformers", "torch==2.5.0", "torchaudio==2.5.0" ]
+remote-smart-turn = []
```
markbackman, line 83: @filipi87 maybe this should just be `smart-turn` to match the package name?

PR #1595, opened 2025-04-15
Update client/server demos to kick off conversation in on_client_read…

markbackman:
…y handler

This is the best practice, which ensures that the client is ready to start the session. We want to reinforce that in our demos.


filipi87 (approved):
LGTM!

PR #1596, opened 2025-04-15
Update services and examples to use gpt-4.1 by default

markbackman:

### DIFF
--- a/examples/canonical-metrics/bot.py
+++ b/examples/canonical-metrics/bot.py
@@ -72,7 +72,7 @@ async def main():
             # voice_id="gD1IexrzCvsXPHUuT0s3",
         )
 
-        llm = OpenAILLMService(api_key=os.getenv("OPENAI_API_KEY"), model="gpt-4o")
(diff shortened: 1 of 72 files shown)


examples/canonical-metrics/bot.py
```diff
@@ -72,7 +72,7 @@ async def main():
             # voice_id="gD1IexrzCvsXPHUuT0s3",
         )
 
-        llm = OpenAILLMService(api_key=os.getenv("OPENAI_API_KEY"), model="gpt-4o")
+        llm = OpenAILLMService(api_key=os.getenv("OPENAI_API_KEY"), model="gpt-4.1")
```
filipi87, line 75: Unless the example needs to use a specific model, shouldn't we just remove the model from here ? 
This would make everything a lot easier the next time we needed to update the default model.

markbackman: Good point. That thought crossed my mind yesterday. I'll do that now.

filipi87 (approved):
I think we should just remove the model from all the examples. That would make everything a lot easier the next time we need to update the default model. But this looks good. 🎉

PR #1598, opened 2025-04-16
Added word/timestamp pairs to ElevenLabsHttpTTSService

markbackman:

This adds word/timestamp pairs to the ElevenLabsHttpTTSService. As part of the change, we're using a new API. Per docs:
```
Converts text into speech using a voice of your choice and returns a stream of JSONs containing audio as a base64 encoded string together with information on when which character was spoken.
```

This requires different audio processing as well.

With this change, ElevenLabsHttpTTSService becomes a WordTTSService, as it's still HTTP based, but it now includes word / timestamp data.

I've tested the word timing using simple-chatbot and this implementation works well!

### DIFF
--- a/src/pipecat/services/elevenlabs/tts.py
+++ b/src/pipecat/services/elevenlabs/tts.py
@@ -25,7 +25,7 @@ from pipecat.frames.frames import (
     TTSStoppedFrame,
 )
 from pipecat.processors.frame_processor import FrameDirection
-from pipecat.services.tts_service import InterruptibleWordTTSService, TTSService
+from pipecat.services.tts_service import InterruptibleWordTTSService, WordTTSService
 from pipecat.transcriptions.language import Language
 
 # See .env.example for ElevenLabs configuration needed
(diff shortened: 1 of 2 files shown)


src/pipecat/services/elevenlabs/tts.py
```diff
@@ -566,17 +652,59 @@ async def run_tts(self, text: str) -> AsyncGenerator[Frame, None]:
...
-                async for chunk in response.content.iter_chunked(CHUNK_SIZE):
-                    if len(chunk) > 0:
-                        await self.stop_ttfb_metrics()
-                        yield TTSAudioRawFrame(chunk, self.sample_rate, 1)
         except Exception as e:
             logger.error(f"Error in run_tts: {e}")
             yield ErrorFrame(error=str(e))
         finally:
             await self.stop_ttfb_metrics()
-            yield TTSStoppedFrame()
```
filipi87, line 582: Why don't we need this anymore?

I looked at the `WordTTSService` code, but it seems like this is only emitted when one of the words is `TTSStoppedFrame`.

Still, I feel like I'm missing something.

```
            elif word == "TTSStoppedFrame" and timestamp == 0:
                frame = TTSStoppedFrame()
```

markbackman: The explicit `yield TTSStoppedFrame()` is no longer needed because we now use `push_stop_frames=True` in the constructor. This activates the parent class's automatic detection system, which monitors audio flow and sends a `TTSStoppedFrame` after 2 seconds without audio generation.

This is a consistent pattern for other services that use word-timestamps.

PR #1600, opened 2025-04-16
Add previous_text context to ElevenLabsHttpTTSService

markbackman:

From 11Labs docs on previous_text:
```
The text that came before the text of the current request. Can be used to improve the speech’s continuity when concatenating together multiple generations or to influence the speech’s continuity in the current generation.
```

For HTTP requests, this parameter provides continuity between HTTP responses. We're passing in previous sentences from the bot's current turn.

### DIFF
--- a/src/pipecat/services/elevenlabs/tts.py
+++ b/src/pipecat/services/elevenlabs/tts.py
@@ -12,6 +12,7 @@ from typing import Any, AsyncGenerator, Dict, List, Literal, Mapping, Optional, 
 import aiohttp
 from loguru import logger
 from pydantic import BaseModel, model_validator
+from sentry_sdk import push_scope
 
 from pipecat.frames.frames import (
     CancelFrame,
--- a/src/pipecat/services/elevenlabs/tts.py
+++ b/src/pipecat/services/elevenlabs/tts.py
@@ -12,13 +12,13 @@ from typing import Any, AsyncGenerator, Dict, List, Literal, Mapping, Optional, 
 import aiohttp
 from loguru import logger
 from pydantic import BaseModel, model_validator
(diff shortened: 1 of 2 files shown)


src/pipecat/services/elevenlabs/tts.py
```diff
@@ -526,16 +530,23 @@ async def start(self, frame: StartFrame):
...
         self._started = False
+        self._previous_text = ""
 
     async def push_frame(self, frame: Frame, direction: FrameDirection = FrameDirection.DOWNSTREAM):
         await super().push_frame(frame, direction)
         if isinstance(frame, (StartInterruptionFrame, TTSStoppedFrame)):
             # Reset timing on interruption or stop
             self._started = False
             self._cumulative_time = 0
+            self._previous_text = ""
```
filipi87, line 541: Like we are resetting the exactly same variables here.

src/pipecat/services/elevenlabs/tts.py
```diff
@@ -614,6 +629,11 @@ async def run_tts(self, text: str) -> AsyncGenerator[Frame, None]:
             "model_id": self._model_name,
         }
 
+        # Include previous text as context if available
+        if self._previous_text:
+            payload["previous_text"] = self._previous_text
+            print(f"Previous text: {self._previous_text}")
```
filipi87, line 635: nit: should we keep it, maybe replace by `logger` ?

filipi87 (approved):
Just some nits, but all looks good, nice improvement.

src/pipecat/services/elevenlabs/tts.py
markbackman: Yes! My mistake.

PR #1601, opened 2025-04-16
Add eject_at_token_exp to Daily REST helpers, modify default values

markbackman:

It's common to expire user access based on tokens, so we're not providing a convenience to eject at token expiration.

Also, I'm adjusting the default values for two DailyRoomProperties to better align with Pipecat:
- `enable_prejoin_ui` defaults to False. This property is most useful when it comes to video conferencing calls. It serves no purpose if you're not using Daily Prebuilt.
- `eject_at_room_exp` now defaults to False. This property ejects all participants from the room when the room expires. Without a default room exp, it doesn't make sense to have this property set to True by default.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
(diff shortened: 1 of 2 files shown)


filipi87 (approved):
LGTM!

PR #1603, opened 2025-04-16
deepgram/tavus fixes

aconchillo:

Use Deepgram's async methods instead of `asyncio.to_thread()`.

TavusVideoService uses app-messages to send audio. These has size limitations and rate limiting. This PR also fixes that by sending smaller chunks and waiting a little bit before sending the next chunk.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- Fixed a `TavusVideoService` issue that was causing audio choppiness.
+
 - Fixed an issue in `SmallWebRTCTransport` where an error was thrown if the
   client did not create a video transceiver.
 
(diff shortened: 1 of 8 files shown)


CHANGELOG.md
```diff
@@ -27,6 +27,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- Fixed a `TavusVideoService` issue that was causing audio choppiness.
+
+- `DeepgramTTSService` now uses HTTP requests directly instead of using
+  Deepgram's SDK. Deepgram's SDK was somehow blocking the event loop causing the
+  audio to be choppy.
```
filipi87, line 34: Interesting. Is it worth to open an issue for them ?

filipi87 (approved):
Makes sense. LGTM!

CHANGELOG.md
aconchillo: I removed this change. Just used their async version instead, it seems to work in normal case (no Tavus involved).

PR #1620, opened 2025-04-19
Add a SmartTurnResultFrame

markbackman:

### DIFF
--- a/src/pipecat/audio/turn/base_turn_analyzer.py
+++ b/src/pipecat/audio/turn/base_turn_analyzer.py
@@ -6,7 +6,7 @@
 
 from abc import ABC, abstractmethod
 from enum import Enum
-from typing import Optional
+from typing import Any, Callable, Optional
 
 
 class EndOfTurnState(Enum):
@@ -15,18 +15,21 @@ class EndOfTurnState(Enum):
 
 
 class BaseTurnAnalyzer(ABC):
-    """
-    Abstract base class for analyzing user end of turn.
-    """
+    """Abstract base class for analyzing user end of turn."""
 
-    def __init__(self, *, sample_rate: Optional[int] = None):
+    def __init__(
+        self,
+        *,
+        sample_rate: Optional[int] = None,
+        on_result: Optional[Callable[[dict], Any]] = None,
+    ):
         self._init_sample_rate = sample_rate
         self._sample_rate = 0
+        self._on_result = on_result
 
     @property
     def sample_rate(self) -> int:
-        """
-        Returns the current sample rate.
+        """Returns the current sample rate.
 
         Returns:
             int: The effective sample rate for audio processing.
@@ -34,8 +37,7 @@ class BaseTurnAnalyzer(ABC):
         return self._sample_rate
 
     def set_sample_rate(self, sample_rate: int):
-        """
-        Sets the sample rate for audio processing.
+        """Sets the sample rate for audio processing.
 
         If the initial sample rate was provided, it will use that; otherwise, it sets to
         the provided sample rate.
@@ -48,8 +50,7 @@ class BaseTurnAnalyzer(ABC):
     @property
     @abstractmethod
     def speech_triggered(self) -> bool:
-        """
-        Determines if speech has been detected.
+        """Determines if speech has been detected.
 
         Returns:
             bool: True if speech is triggered, otherwise False.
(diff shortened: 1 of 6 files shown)


src/pipecat/audio/turn/base_turn_analyzer.py
```diff
@@ -15,27 +15,29 @@ class EndOfTurnState(Enum):
...
-    """
-
-    def __init__(self, *, sample_rate: Optional[int] = None):
+    """Abstract base class for analyzing user end of turn."""
+
+    def __init__(
+        self,
+        *,
+        sample_rate: Optional[int] = None,
+        on_result: Optional[Callable[[dict], Any]] = None,
```
aconchillo, line 24: We might want to inherit from `BaseObject` instead of `ABC` this will give us event handlers by default. We can then register `on_result` in here and then call `_call_event_handler("on_result"...)`.

Maybe we should call it `on_turn_result` or `on_turn_detection` or something more descriptive.

aconchillo: Ah, wait. Why do we need to be able to set a callback if we push a frame?

src/pipecat/frames/frames.py
```diff
@@ -615,6 +615,26 @@ class MetricsFrame(SystemFrame):
...
+    e2e_processing_time_ms: float  # Complete end-to-end time (includes network round trip)
+
+    def __str__(self):
+        return (
+            f"{self.name}(is_complete: {self.is_complete}, "
+            f"probability: {self.probability:.4f}, "
+            f"model: {self.inference_time_ms:.2f}ms, "
+            f"server: {self.server_total_time_ms:.2f}ms, "
+            f"e2e: {self.e2e_processing_time_ms:.2f}ms)"
+        )
```
aconchillo, line 635: I'm wondering if this should be a `MetricsFrame`.

src/pipecat/audio/turn/base_turn_analyzer.py
markbackman: The analyzer isn't a frame processor. Given that, can I push a frame?

aconchillo: Ah yes. You can do it as an event handler as well. So, the same you did with the callback you do with an event handler.

src/pipecat/transports/base_input.py
```diff
@@ -263,3 +265,35 @@ async def _audio_task_handler(self):
...
+
+        This is called potentially from a background thread, so we need to schedule it on the event loop.
+
+        Args:
+            result: The prediction result dict from the smart turn analyzer.
+        """
+        # Get current event loop and schedule the async handler
+        asyncio.run_coroutine_threadsafe(
+            self._async_handle_smart_turn_result(result), self.get_event_loop()
+        )
```
aconchillo, line 280: If we do it with an event handler we don't need to do this. The event handler will already do the right thing.

src/pipecat/audio/turn/base_turn_analyzer.py
markbackman: Ok, this now uses an event handler instead. Also, this makes the switch to the BaseObject, which is nice. Thanks for the tip.

src/pipecat/frames/frames.py
markbackman: I think this makes sense.

src/pipecat/audio/turn/base_turn_analyzer.py
```diff
@@ -45,11 +51,23 @@ def set_sample_rate(self, sample_rate: int):
...
+        """Get and clear the last prediction result.
+
+        This method should be implemented by subclasses to provide access
+        to the most recent prediction result. The default implementation
+        returns None.
+
+        Returns:
+            Optional[Dict[str, Any]]: The last prediction result, if any.
+        """
+        return None
```
aconchillo, line 65: Do we nerd this function? Who is supposef to use it? Also, maybe it shouldn’t clear the value so you can always access to it.

src/pipecat/audio/turn/base_turn_analyzer.py
```diff
@@ -15,27 +17,28 @@ class EndOfTurnState(Enum):
...
-    """
-    Abstract base class for analyzing user end of turn.
+    """Abstract base class for analyzing user end of turn.
+
+    This class inherits from BaseObject to leverage its event handling system
+    while still defining an abstract interface through abstract methods.
     """
 
     def __init__(self, *, sample_rate: Optional[int] = None):
+        super().__init__()
```
aconchillo, line 27: We probably don't need this.

aconchillo (approved):

aconchillo:
LGTM! Just one minor thing

PR #1621, opened 2025-04-19
adding codec options to SIP dial-in

vr000m:
https://docs.daily.co/guides/products/dial-in-dial-out/sip#sip-dial-in-audio-and-video allows setting codecs, this shows how you can override the default codec


markbackman:
@vr000m given what we learned is this required or can we close it out?

vr000m:
While the above PR does give you the option for setting a codec. We now have a domain property that can override the default and provides slightly better control because you dont have to redeploy your bot to change the codec.

```bash
curl --request POST \
  --url https://api.daily.co/v1/ \
  --header 'Authorization: API_KEY' \
  --header 'Content-Type: application/json' \
  --data '{
"properties": {
	"default_sip_codecs": {
		"audio": ["PCMU"],
		"video": ["VP8"]
	}
  }
}'
```

PR #1623, opened 2025-04-21
Azure TTS fixed by clearing the audio queue before synthesizing the next text

phamtrung0633:
---

#### 📝 Description of Changes

This PR fixes an issue where audio would get stuck in the queue when an interruption occurs during Azure TTS synthesis.

Previously, if Azure hadn’t finished outputting audio for the current text when an interruption occurred, the remaining audio frames would remain in the queue and never get played. The TTS processing logic breaks upon encountering a `None` in the audio queue — a signal that typically denotes the end of synthesis for a particular utterance.

However, if a `None` from a previously interrupted synthesis remains in the queue, it causes the next synthesized text to be skipped or delayed. For example, consider the following sequence in the audio queue after an interruption of "Text A" and the start of synthesis for "Text B":

```
["Text A audio 1", "Text A audio 2", None, "Text B audio 1", "Text B audio 2"]
```

In this case, when the processor hits the leftover `None`, it prematurely breaks and doesn't proceed to play the "Text B" audio, effectively causing audio to lag one step behind.

This PR ensures the queue is properly flushed or handled to avoid this desynchronization issue.

---

### DIFF
--- a/src/pipecat/services/azure/tts.py
+++ b/src/pipecat/services/azure/tts.py
@@ -17,6 +17,7 @@ from pipecat.frames.frames import (
     TTSAudioRawFrame,
     TTSStartedFrame,
     TTSStoppedFrame,
+    TTSTextFrame,
 )
 from pipecat.services.azure.common import language_to_azure_language
 from pipecat.services.tts_service import TTSService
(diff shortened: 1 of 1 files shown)


src/pipecat/services/azure/tts.py
```diff
@@ -195,6 +196,44 @@ def _handle_canceled(self, evt):
...
+
+        if text:
+            await self.process_generator(self.run_tts(text))
+
+        await self.stop_processing_metrics()
+
+        if self._push_text_frames:
+            # We send the original text after the audio. This way, if we are
+            # interrupted, the text is not added to the assistant context.
+            await self.push_frame(TTSTextFrame(text))
```
filipi87: Would it work if instead of doing this we just override the `_handle_interruption` method ? 

```
    async def _handle_interruption(self, frame: StartInterruptionFrame, direction: FrameDirection):
        # Clear the audio queue in case there's still audio in it, causing the next audio response
        # to be cut off by the 'None' element returned at the end of the previous audio synthesis.
        # Empty the audio queue before processing the new text
        while not self._audio_queue.empty():
            self._audio_queue.get_nowait()
            self._audio_queue.task_done()
```

filipi87: cc: @markbackman for thoughts on this.

phamtrung0633: Hi @filipi87 , yes it would.  However due to Azure tts's run_tts function is designed to process tts synchronously meaning when a tts request is made all audio must be received before next one, it makes me do this way / logically if the way the system works is to use audio_queue with every tts request must be fulfilled completely before the next one, it is intuitive to clean the queue before every single tts request. 
Because who knows if somehow interruption signal got ignored by some modifications or something if we rely on it

filipi87: @phamtrung0633, if that's the case, couldn't you insert this code to clear the queue inside `run_tts` in the `AzureTTSService`, right at the beginning ? It seems like it would have the same effect.

phamtrung0633: You can yeah that's true

filipi87 (approved):
Thank you @phamtrung0633 , I have pushed the changes and I am going to merge the PR, so we can include it in the next Pipecat release.

src/pipecat/services/azure/tts.py
markbackman: LGTM @filipi87! Thanks for looking into this.

phamtrung0633:
Awesome, thank you @filipi87

PR #1631, opened 2025-04-22
Returning the turn as complete if the request don’t return a result within SmartTurnParams stop_secs

filipi87:
Returning the turn as complete if the request don’t return a result within SmartTurnParams stop_secs


markbackman (approved):
LGTM! My only question is whether we should use httpx or aiohttp. We use aiohttp in most places. Though, it comes with more overhead, so I see why httpx is a good choice here.

src/pipecat/audio/turn/smart_turn.py
```diff
@@ -25,9 +24,10 @@ def __init__(self, url: str, **kwargs):
             logger.error("remote_smart_turn_url is not set.")
             raise Exception("remote_smart_turn_url must be provided.")
 
-        # Use a session to reuse connections (keep-alive)
-        self.session = requests.Session()
-        self.session.headers.update({"Connection": "keep-alive"})
+        self.client = httpx.AsyncClient(
+            headers={"Connection": "keep-alive"},
+            timeout=httpx.Timeout(self._params.stop_secs),
+        )
```
aconchillo, line 30: we should use `aiohttp` which is what we use everywhere.

aconchillo: and we should pass an `aiohttp_session` argument as well, as we do with all the services.

aconchillo:
> LGTM! My only question is whether we should use httpx or aiohttp. We use aiohttp in most places. Though, it comes with more overhead, so I see why httpx is a good choice here.

I also made that comment. `aoihttp` is in our base dependencies, `httpx` is not. I think we should stick to `aiohttp`, it also faster and more efficient than `httpx` from what I've read.

src/pipecat/audio/turn/smart_turn/fal_smart_turn.py
```diff
@@ -0,0 +1,26 @@
...
+
+from pipecat.audio.turn.smart_turn.http_smart_turn import HttpSmartTurnAnalyzer
+
+
+class FalSmartTurnAnalyzer(HttpSmartTurnAnalyzer):
+    def __init__(
+        self,
+        *,
+        aiohttp_session: aiohttp.ClientSession,
+        url: str = "https://fal.run/fal-ai/smart-turn/raw",
```
filipi87, line 19: Cool. 🎉

filipi87:
Thank you @aconchillo and @markbackman for pushing this improvements. 🎉

PR #1634, opened 2025-04-22
Automatically hangup Twilio calls

markbackman:

@aconchillo I also added a return for None, which it looks like was missing. Happy to remove that if it's not needed.


src/pipecat/transports/network/websocket_server.py
```diff
@@ -205,6 +206,8 @@ async def process_frame(self, frame: Frame, direction: FrameDirection):
         if isinstance(frame, StartInterruptionFrame):
             await self._write_frame(frame)
             self._next_send_time = 0
+        elif isinstance(frame, EndFrame):
+            await self._write_frame(frame)
```
aconchillo, line 210: You can add this instead

```
        await super().stop(frame)
        await self._write_frame(frame)
```

src/pipecat/transports/network/fastapi_websocket.py
```diff
@@ -207,6 +207,7 @@ async def start(self, frame: StartFrame):
 
     async def stop(self, frame: EndFrame):
         await super().stop(frame)
+        await self._write_frame(frame)
```
aconchillo, line 210: We should also do it on `CancelFrame` as well.

src/pipecat/serializers/twilio.py
```diff
@@ -68,7 +128,44 @@ async def serialize(self, frame: Frame) -> str | bytes | None:
...
+            from twilio.rest import Client
+
+            account_sid = os.getenv("TWILIO_ACCOUNT_SID")
+            auth_token = os.getenv("TWILIO_AUTH_TOKEN")
+
+            if not account_sid or not auth_token:
+                logger.warning(
+                    "Missing Twilio credentials (TWILIO_ACCOUNT_SID, TWILIO_AUTH_TOKEN), cannot hang up call"
+                )
+                return
```
aconchillo, line 146: I think it's better if we pass them to the serializer as we do in the services and set them to None by default.

aconchillo: That is, avoid using environment variables.

src/pipecat/serializers/twilio.py
```diff
@@ -41,13 +77,37 @@ def __init__(self, stream_sid: str, params: InputParams = InputParams()):
...
+        For EndFrames, initiates call termination if auto_hang_up is enabled.
+
+        Args:
+            frame: The Pipecat frame to serialize.
+
+        Returns:
+            Serialized data as string or bytes, or None if the frame isn't handled.
+        """
+        if self._params.auto_hang_up and isinstance(frame, (EndFrame)):
+            asyncio.create_task(self._hang_up_call())
```
aconchillo, line 108: I don't think we should create a task. EndFrame closes things in order which means we might be closing other stuff while the task is being executed. Since the cleanup happens in order I think it should be fine without a task. Also, this is a task we don't await which is not nice.

aconchillo: So, just

```
await self._hang_up_call()
```

src/pipecat/serializers/twilio.py
```diff
@@ -68,7 +128,44 @@ async def serialize(self, frame: Frame) -> str | bytes | None:
...
+
+            if not account_sid or not auth_token:
+                logger.warning(
+                    "Missing Twilio credentials (TWILIO_ACCOUNT_SID, TWILIO_AUTH_TOKEN), cannot hang up call"
+                )
+                return
+
+            # Create Twilio client and hang up the call using the correct CallSid
+            client = Client(account_sid, auth_token)
+            client.calls(self._call_sid).update(status="completed")
```
aconchillo, line 150: I'm wondering if we should use the REST API instead. This way we can make it async and we don't really need to include the whole Twilio package.

markbackman: Updated.

src/pipecat/serializers/twilio.py
markbackman: Updated.

src/pipecat/transports/network/fastapi_websocket.py
markbackman: Added.

src/pipecat/transports/network/websocket_server.py
markbackman: Updated.

CHANGELOG.md
```diff
@@ -9,6 +9,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- Added automatic hangup logic to the Twilio serializer. This feature hangs up
+  the Twilio call when an `EndFrame` is received. It is enabled by default and
+  is configurable via the `auto_hang_up` `InputParam`. To use this feature, set
+  up the following env vars: `TWILIO_ACCOUNT_SID` and `TWILIO_AUTH_TOKEN`.
```
aconchillo, line 15: Mention `CancelFrame` and remove the environment variables.

src/pipecat/transports/network/websocket_server.py
```diff
@@ -195,6 +196,10 @@ async def start(self, frame: StartFrame):
         await self._params.serializer.setup(frame)
         self._send_interval = (self._audio_chunk_size / self.sample_rate) / 2
 
+    async def stop(self, frame: EndFrame):
+        await super().stop(frame)
+        await self._write_frame(frame)
```
aconchillo, line 201: We need the same for `cancel()`

aconchillo:
LGTM! Added a couple of final comments.

src/pipecat/transports/network/websocket_server.py
markbackman: Added a `cancel` method which calls `_write_frame`.

PR #1642, opened 2025-04-23
Add deepgram and cartesia to foundational example requirements to mak…

markbackman:
…e quickstart smoother

Most examples use these two services, so we should include them by default. OpenAI would be the other but it's included with pipecat-ai.


filipi87 (approved):
LGTM!

PR #1644, opened 2025-04-23
Add transports examples to foundational examples

markbackman:

In switching to SmallWebRTCTransport for all examples, we don't have simple transport examples for audio chat. I'm removing the previous 04 example: speech and utterance, as it's unclear what it was accomplishing. This adds an example for SmallWebRTCTransport, DailyTransport, and LivekitTransport (renaming 29 example).

This should make it easier to find a basic example of how to do things with these different transports.


filipi87 (approved):
Makes sense. 🚀

PR #1648, opened 2025-04-23
input transports now always push audio frames

aconchillo:

In this PR we change a couple of things:

- Input transport always push audio frames down the pipeline unless `audio_in_passthrough` is set to False.
- We deprecate `vad_audio_passthrough` in favor of `audio_in_passthrough`.
- We deprecate `vad_enabled` in favor of `audio_in_enabled` and `vad_analyzer`.

### DIFF
--- a/examples/foundational/run.py
+++ b/examples/foundational/run.py
@@ -35,6 +35,9 @@ ice_servers = ["stun:stun.l.google.com:19302"]
 # Mount the frontend at /
 app.mount("/client", SmallWebRTCPrebuiltUI)
 
+# Store program arguments
+args: argparse.Namespace = argparse.Namespace()
+
 # Store the bot module and function info
 bot_module: Any = None
 run_bot_func: Optional[Callable] = None
@@ -116,7 +119,7 @@ async def offer(request: dict, background_tasks: BackgroundTasks):
 
         # We've already checked that run_bot_func exists
         assert run_bot_func is not None
-        background_tasks.add_task(run_bot_func, pipecat_connection)
+        background_tasks.add_task(run_bot_func, pipecat_connection, args)
 
     answer = pipecat_connection.get_answer()
     # Updating the peer connection inside the map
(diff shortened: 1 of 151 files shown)


aconchillo:
Also changed `camera` to `video` which I believe is more appropiate.

examples/foundational/run.py
```diff
@@ -116,7 +119,7 @@ async def handle_disconnected(webrtc_connection: SmallWebRTCConnection):
 
         # We've already checked that run_bot_func exists
         assert run_bot_func is not None
-        background_tasks.add_task(run_bot_func, pipecat_connection)
+        background_tasks.add_task(run_bot_func, pipecat_connection, args)
```
markbackman, line 122: It's a bummer that we have to add boilerplate like this (`_: argparse.Namespace`) to every function signature. Perhaps we could offer two function signatures to make this more flexible:

```
background_tasks.add_task(run_bot_func, pipecat_connection, args)
```

and

```
background_tasks.add_task(run_bot_func, pipecat_connection)
```

In `import_bot_file`, we could check to see if args are provided. Something like this:
```
    # Check for run_bot function first
    if hasattr(module, "run_bot"):
        run_func = module.run_bot
        # Check if the function accepts a WebRTC connection
        sig = signature(run_func)
        is_webrtc = len(sig.parameters) > 0
        # Check if it needs the args parameter
        needs_args_param = len(sig.parameters) > 1
        return module, run_func, is_webrtc, needs_args_param
```

WDYT?

markbackman (commented):
The change looks really good. I like how this simplifies things. The TransportParams were a bit confusing. This definitely streamlines things and makes them more self evident.

The only question I have is about the foundational run.py file and how we can simplify the function signature in more places.

examples/foundational/run.py
aconchillo: I thought about it but this adds even more magic. This change makes it obvious that you can pass program arguments. It's true that it's more verbose, but without it you would have no clue unless you find an example that does it.

Python doesn't have function overloading and this would be adding it.

markbackman: Ok, given that, I have no objection to this change. It really improves things quite nicely.

markbackman (approved):
LGTM!

PR #1651, opened 2025-04-24
Block transcription frames in STTMuteFilter

lucasrothman:

When using Daily hosted transcription the STTMuteFilter doesn't work as desired because `TranscriptionFrames` are still being pushed down the pipeline.

This will cause generations even when the VAD events are blocked because of the timeout mechanism in the context manager to handle receiving transcripts outside of `UserStartedSpeaking` and `UserStoppedSpeaking` events.

Blocking transcription frames properly mutes the user.


markbackman:
Thanks again for this change! Implemented the same thing here, so closing this one:
https://github.com/pipecat-ai/pipecat/pull/1702

PR #1664, opened 2025-04-25
Improved: Ultravox performance

kylegani:
This pull request improves the performance of the Ultravox STT service, by yielding text frames as they are created. It also pre-warms the model at service startup so that the first real request is faster.

### DIFF
--- a/src/pipecat/services/ultravox/stt.py
+++ b/src/pipecat/services/ultravox/stt.py
@@ -71,7 +71,7 @@ class UltravoxModel:
         stop_token_ids: Optional token IDs to stop generation
     """
 
-    def __init__(self, model_name: str = "fixie-ai/ultravox-v0_4_1-llama-3_1-8b"):
+    def __init__(self, model_name: str = "fixie-ai/ultravox-v0_5-llama-3_1-8b"):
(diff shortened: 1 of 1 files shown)


src/pipecat/services/ultravox/stt.py
```diff
@@ -218,9 +218,60 @@ def __init__(
         self._temperature = temperature
         self._max_tokens = max_tokens
         self._connection_active = False
+        self._warm_up_duration_sec = 1
 
         logger.info(f"Initialized UltravoxSTTService with model: {model_name}")
 
+    async def warm_up_model(self):
```
aconchillo, line 225: add `_` prefix since this is an internal function.

src/pipecat/services/ultravox/stt.py
```diff
@@ -237,6 +288,9 @@ async def start(self, frame: StartFrame):
         """
         await super().start(frame)
         self._connection_active = True
+
+        await self.warm_up_model()
```
aconchillo, line 292: `await self._warm_up_model()`

aconchillo (approved):

PR #1666, opened 2025-04-25
Add VADUserStartedSpeakingFrame and VADUserStoppedSpeakingFrame

markbackman:

As the CHANGELOG points out, this helps to differentiate from user started/stopped speaking frames in the case where smart turn detection is used.

One thing I didn't add but considered was the speaking duration to the stopped frame. It would be easy to handle in the application code, but we could consider making things easier by adding it here. I'm indifferent.

---

Implementation-wise, I followed the same pattern we used for the smart turn analyzer frames.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- Added `VADUserStartedSpeakingFrame` and `VADUserStoppedSpeakingFrame`,
+  indicating when the VAD detected the user to start and stop speaking. These
+  events are helpful when using smart turn detection, as the user's stop time
+  can differ from when their turn ends (signified by UserStoppedSpeakingFrame).
+
 - Added `TranslationFrame`, a new frame type that contains a translated
   transcription.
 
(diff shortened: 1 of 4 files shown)


filipi87:
@markbackman, what if instead changing the `vad_analyser` to return a `tuple`, and need to handle a new even inside the `base_input`, we just changed this code inside the `base_input`, from this:

```
if can_create_user_frames:
    if new_vad_state == VADState.SPEAKING:
        frame = UserStartedSpeakingFrame()
    elif new_vad_state == VADState.QUIET:
        frame = UserStoppedSpeakingFrame()
```

to this:

```
if new_vad_state == VADState.SPEAKING:
    await self.push_frame(VADUserStartedSpeakingFrame())
    if can_create_user_frames:
        frame = UserStartedSpeakingFrame()
elif new_vad_state == VADState.QUIET:
    await self.push_frame(VADUserStoppedSpeakingFrame())
    if can_create_user_frames:
        frame = UserStoppedSpeakingFrame()
```

What do you think ?

markbackman:
> What do you think ?

I'm in favor. This further simplifies things. I initially went the route I'm proposing because I was going to return the speaking duration, but the more I think about that, the less I like it. Mainly because the duration can be determined in multiple ways, so it's better to leave it to the developer.

I'll make this change.

filipi87 (approved):
Looks good. 🚀

lucasrothman:
🚀

PR #1667, opened 2025-04-25
function call single parameter

aconchillo:

This PR simplifies function call parameters by simply using a single argument `FunctionCallParams`. This will also allow us to introduce new fields in the future without breaking any previous code.

### DIFF
--- a/src/pipecat/processors/frameworks/rtvi.py
+++ b/src/pipecat/processors/frameworks/rtvi.py
@@ -251,7 +251,7 @@ class RTVIBotReady(BaseModel):
 class RTVILLMFunctionCallMessageData(BaseModel):
     function_name: str
     tool_call_id: str
-    args: dict
+    arguments: Mapping[str, Any]
 
 
 class RTVILLMFunctionCallMessage(BaseModel):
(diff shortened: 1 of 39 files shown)


markbackman (approved):
Nice clean up! 🧹

src/pipecat/processors/frameworks/rtvi.py
```diff
@@ -251,7 +251,7 @@ class RTVIBotReady(BaseModel):
 class RTVILLMFunctionCallMessageData(BaseModel):
     function_name: str
     tool_call_id: str
-    args: dict
+    arguments: Mapping[str, Any]
```
mwood23, line 254: Sorry to helicopter review, but I think this is a breaking change. The associated types for this live in this repo: https://github.com/pipecat-ai/pipecat-client-web/blob/1bc412c35241002c44182e5c495d674e304821a0/client-js/src/helpers/llm.ts#L18 and I didn't see a PR go in to update it there.

Just hit this myself and wanted to flag before the official release. 🙏

markbackman: FYI: should be fixed here: https://github.com/pipecat-ai/pipecat/pull/1690. Thanks for the heads up 🙌

PR #1668, opened 2025-04-25
Update examples with new transport param names

vipyne:

A few more examples to update to new transport param names. as per #1648


markbackman (approved):
Thank you!

PR #1671, opened 2025-04-26
support for rime arcana model

kwindla:
Add "Accept: audio/wav" hard-coding for Rime's new Arcana model, and strip wav header if necessary.


markbackman (approved):
LGTM. Rebased and added a changelog entry.

PR #1672, opened 2025-04-27
Use "use_original_timestamps" only for sonic-2 model

golbin:

The `use_original_timestamps` option only works correctly with the Cartesia `sonic-2` model.

For other models, setting this option to True causes the Assistant Context to fail to aggregate properly.

Since version 0.0.63, using a model other than Sonic-2 cause the Assistant Context to not continue correctly.

Fixed: #1803


markbackman:
@golbin I just tested 07-interruptible.py with the model set to `sonic` and I see context aggregation working correctly. Before we add this change, can you help me understand under which cases things are failing?

src/pipecat/services/cartesia/tts.py
```diff
@@ -166,7 +166,7 @@ def _build_msg(
             "output_format": self._settings["output_format"],
             "language": self._settings["language"],
             "add_timestamps": add_timestamps,
-            "use_original_timestamps": True,
+            "use_original_timestamps": True if self.model_name == "sonic-2" else False,
```
markbackman, line 169: `use_original_timestamps: True` only fails for `sonic` (and works for `sonic-2` and `sonic-turbo`. Let's assume future models will continue to support this and only opt out original sonic models.

```suggestion
            "use_original_timestamps": False if self.model_name == "sonic" else True,
```

markbackman:
Can you also add a CHANGELOG entry?

```
### Fixed

- Fixed an issue with `CartesiaTTSService` where `TTSTextFrame` messages weren't being emitted when the model was set to `sonic`. This resulted in the assistant context not being updated with assistant messages.
```

We'll make sure this is in the next release.

src/pipecat/services/cartesia/tts.py
golbin: I’ll change it.

golbin:
@markbackman I applied your suggestion and updated the CHANGELOG.

markbackman (approved):
LGTM. Thanks!

PR #1684, opened 2025-04-28
Add `load_dotenv` to patient intake example

carschandler:

`bot.py` has it but `server.py` doesn't, so by default the instructions in the README don't work.


markbackman (approved):
Good catch! 🕵️

Thank you.

markbackman:
Please install the pre-commit hook to check for linting errors. You can also set up Ruff.

PR #1687, opened 2025-04-29
Fix imports for modal deployment example

flixoflax:
Fix imports for modal deployment example from 

```python
from pipecat.services.cartesia.tts import CartesiaTTSService
```

to

```python
from pipecat.services.cartesia import CartesiaTTSService
from pipecat.services.openai import OpenAILLMService
```


aconchillo:
These should be the correct ones:

```
from pipecat.services.cartesia.tts import CartesiaTTSService
```

maybe you are running an older Pipecat version? This was changed in Pipecat 0.0.62 https://github.com/pipecat-ai/pipecat/releases/tag/v0.0.62

PR #1690, opened 2025-04-29
RTVIProcessor: use single FunctionCallParams

aconchillo:


src/pipecat/processors/frameworks/rtvi.py
```diff
@@ -61,6 +61,7 @@
     OpenAILLMContextFrame,
 )
 from pipecat.processors.frame_processor import FrameDirection, FrameProcessor
+from pipecat.services.llm_service import FunctionCallParams
```
aconchillo, line 64: We shouldn't import `services` in `processors` package. We should probably move `FunctionCallParams` somewhere else.

filipi87 (approved):
LGTM!

Issue #1686, opened 2025-04-29
Context not correctly set


NicolaiMinter:
### pipecat version

0.0.65

### Python version

3.13.3

### Operating System

macOS Sequoia 15.4

### Issue description

My transcription messages are not correctly saved in the current pipecat version for Gemini Live:

### Reproduction steps

Minimal reproduction code:

```python
#
# Copyright (c) 2024–2025, Daily
#
# SPDX-License-Identifier: BSD 2-Clause License
#

import sys

from dotenv import load_dotenv
from fastapi import WebSocket
from loguru import logger
from pipecat.audio.vad.silero import SileroVADAnalyzer
from pipecat.audio.vad.vad_analyzer import VADParams
from pipecat.pipeline.pipeline import Pipeline
from pipecat.pipeline.runner import PipelineRunner
from pipecat.pipeline.task import PipelineParams, PipelineTask
from pipecat.processors.aggregators.openai_llm_context import OpenAILLMContext
from pipecat.serializers.protobuf import ProtobufFrameSerializer
from pipecat.services.gemini_multimodal_live.gemini import (
    GeminiMultimodalLiveLLMService,
)
from pipecat.transports.network.fastapi_websocket import (
    FastAPIWebsocketParams,
    FastAPIWebsocketTransport,
)

load_dotenv(override=True)

logger.remove(0)
logger.add(sys.stderr, level="DEBUG")


async def run_bot(websocket_client: WebSocket, testing: bool = False):
    transport = FastAPIWebsocketTransport(
        websocket=websocket_client,
        params=FastAPIWebsocketParams(
            audio_out_enabled=True,
            add_wav_header=True,
            vad_enabled=True,
            vad_analyzer=SileroVADAnalyzer(params=VADParams(stop_secs=1)),
            vad_audio_passthrough=True,
            serializer=ProtobufFrameSerializer(),
            session_timeout=60,  # 3 minutes
        ),
    )

    voice_id = "Puck"

    llm = GeminiMultimodalLiveLLMService(
        api_key=GOOGLE_API_KEY,
        model="models/gemini-2.0-flash-live-001",
        system_instruction="You are a helpful LLM in a WebRTC call. Your goal is to demonstrate your capabilities in a succinct way. Your output will be converted to audio so don't include special characters in your answers. Respond to what the user said in a creative and helpful way.",
        transcribe_model_audio=True,
        transcribe_user_audio=True,
        voice_id=voice_id,
    )

[...]

PR #1696, opened 2025-04-29
Fix: GeminiMultimodalLiveLLMService was appending tokens to the context

markbackman:

Issue: https://github.com/pipecat-ai/pipecat/issues/1686

### DIFF
--- a/src/pipecat/processors/aggregators/llm_response.py
+++ b/src/pipecat/processors/aggregators/llm_response.py
@@ -36,6 +36,7 @@ from pipecat.frames.frames import (
     StartInterruptionFrame,
     TextFrame,
     TranscriptionFrame,
+    TTSTextFrame,
     UserImageRawFrame,
     UserStartedSpeakingFrame,
     UserStoppedSpeakingFrame,
@@ -493,7 +494,7 @@ class LLMAssistantContextAggregator(LLMContextResponseAggregator):
             await self._handle_llm_start(frame)
         elif isinstance(frame, LLMFullResponseEndFrame):
             await self._handle_llm_end(frame)
-        elif isinstance(frame, TextFrame):
+        elif isinstance(frame, TTSTextFrame):
             await self._handle_text(frame)
         elif isinstance(frame, LLMMessagesAppendFrame):
             self.add_messages(frame.messages)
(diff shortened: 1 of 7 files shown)


src/pipecat/processors/aggregators/llm_response.py
```diff
@@ -493,7 +494,7 @@ async def process_frame(self, frame: Frame, direction: FrameDirection):
             await self._handle_llm_start(frame)
         elif isinstance(frame, LLMFullResponseEndFrame):
             await self._handle_llm_end(frame)
-        elif isinstance(frame, TextFrame):
+        elif isinstance(frame, TTSTextFrame):
```
markbackman, line 497: This might not be a safe change to make, but I'm proposing it. TextFrame is better, but in the case of S2S models, they output both TTSTextFrame and LLMTextFrame. Using just `TextFrame` results in duplicates.

We could instead do some filtering on: "is not LLMTextFrame and is TextFrame" but that feels a bit strange.

markbackman: @aconchillo any thoughts on the right thing to do here?

src/pipecat/services/gemini_multimodal_live/gemini.py
```diff
@@ -858,24 +852,20 @@ async def _handle_evt_tool_call(self, evt):
...
-            await self.push_frame(LLMFullResponseStartFrame())
-            await self.push_frame(LLMTextFrame(text=text))
-            await self.push_frame(TTSTextFrame(text=text))
-            await self.push_frame(LLMFullResponseEndFrame())
+
+        if not text:
+            return
+
+        await self.push_frame(LLMTextFrame(text=text))
+        await self.push_frame(TTSTextFrame(text=text))
```
aconchillo, line 868: I think we only need

```
        await self.push_frame(TTSTextFrame(text=text))
```

see other comment.

aconchillo:
LGTM

src/pipecat/services/gemini_multimodal_live/gemini.py
aconchillo: Nevermind. AS we discussed removing `LLMTextFrame` would result in other processors and observers not to work.

PR #1701, opened 2025-04-30
Add context_window_compression support to GeminiMultimodalLiveLLMService

markbackman:

Without this, the Gemini API will disconnect the session once the context limit is hit. From Google's docs:
> Session duration can be extended to unlimited by enabling session [compression](https://ai.google.dev/gemini-api/docs/live#context-window-compression). Without compression, audio-only sessions are limited to 15 minutes, and audio plus video sessions are limited to 2 minutes. Exceeding these limits without compression will terminate the connection.

Source: https://ai.google.dev/gemini-api/docs/live#maximum-session-duration

### DIFF
--- a/src/pipecat/services/gemini_multimodal_live/gemini.py
+++ b/src/pipecat/services/gemini_multimodal_live/gemini.py
@@ -265,6 +265,15 @@ class GeminiVADParams(BaseModel):
     silence_duration_ms: Optional[int] = Field(default=None)
 
 
+class ContextWindowCompressionParams(BaseModel):
+    """Parameters for context window compression."""
+
+    enabled: bool = Field(default=False)
+    trigger_tokens: Optional[int] = Field(
+        default=None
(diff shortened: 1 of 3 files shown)


filipi87 (approved):
Nice catch.

src/pipecat/services/gemini_multimodal_live/gemini.py
```diff
@@ -355,6 +365,9 @@ def __init__(
             "language": self._language_code,
             "media_resolution": params.media_resolution,
             "vad": params.vad,
+            "context_window_compression": params.context_window_compression.model_dump()
+            if params.context_window_compression
+            else None,
```
aconchillo, line 370: This should be `{}`

markbackman: Fixing here: https://github.com/pipecat-ai/pipecat/pull/1696

PR #1704, opened 2025-04-30
Amazon Nova Sonic LLM service

kompfner:


src/pipecat/services/aws_nova_sonic/aws.py
```diff
@@ -0,0 +1,979 @@
...
+    InvokeModelWithBidirectionalStreamOperationInput,
+)
+from aws_sdk_bedrock_runtime.config import Config, HTTPAuthSchemeResolver, SigV4AuthScheme
+from aws_sdk_bedrock_runtime.models import (
+    BidirectionalInputPayloadPart,
+    InvokeModelWithBidirectionalStreamInput,
+    InvokeModelWithBidirectionalStreamInputChunk,
+    InvokeModelWithBidirectionalStreamOperationOutput,
+    InvokeModelWithBidirectionalStreamOutput,
+)
```
filipi87, line 29: We should probably encapsulate the imports in something like this:

```
try:
    from aws_sdk_bedrock_runtime.client import (
       BedrockRuntimeClient,
       InvokeModelWithBidirectionalStreamOperationInput,
    )
    from aws_sdk_bedrock_runtime.config import Config, HTTPAuthSchemeResolver, SigV4AuthScheme
    from aws_sdk_bedrock_runtime.models import (
        BidirectionalInputPayloadPart,
        InvokeModelWithBidirectionalStreamInput,
        InvokeModelWithBidirectionalStreamInputChunk,
        InvokeModelWithBidirectionalStreamOperationOutput,
        InvokeModelWithBidirectionalStreamOutput,
    )
    logger.error(
        "In order to use AWS Nova Sonic, you need to `pip install pipecat-ai[aws]`. Also, set `AWS_SECRET_ACCESS_KEY`, `AWS_ACCESS_KEY_ID`, and `AWS_REGION` environment variable."
    )
```

examples/foundational/20e-persistent-context-aws-nova-sonic.py
```diff
@@ -0,0 +1,256 @@
...
+from pipecat.transports.base_transport import TransportParams
+from pipecat.transports.network.small_webrtc import SmallWebRTCTransport
+from pipecat.transports.network.webrtc_connection import SmallWebRTCConnection
+
+load_dotenv(override=True)
+
+BASE_FILENAME = "/tmp/pipecat_conversation_"
+
+
+async def fetch_weather_from_api(function_name, tool_call_id, args, llm, context, result_callback):
```
filipi87, line 34: Function calls now receive a single parameter `FunctionCallParams` instead of `(function_name, tool_call_id, args, llm, context, result_callback)` which is now deprecated.

So it would be something like this:
`async def fetch_weather_from_api(params: FunctionCallParams):`

examples/foundational/20e-persistent-context-aws-nova-sonic.py
```diff
@@ -0,0 +1,256 @@
...
+        {
+            "conditions": "nice",
+            "temperature": temperature,
+            "format": args["format"],
+            "timestamp": datetime.now().strftime("%Y%m%d_%H%M%S"),
+        }
+    )
+
+
+async def get_saved_conversation_filenames(
```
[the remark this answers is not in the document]
filipi87: I guess this change have been included in the last week or so. Pipecat is changing too fast. 😅

examples/foundational/20e-persistent-context-aws-nova-sonic.py
```diff
@@ -0,0 +1,256 @@
...
+from pipecat.pipeline.task import PipelineParams, PipelineTask
+from pipecat.processors.aggregators.openai_llm_context import OpenAILLMContext
+from pipecat.services.aws_nova_sonic.aws import AWSNovaSonicLLMService
+from pipecat.transports.base_transport import TransportParams
+from pipecat.transports.network.small_webrtc import SmallWebRTCTransport
+from pipecat.transports.network.webrtc_connection import SmallWebRTCConnection
+
+load_dotenv(override=True)
+
+BASE_FILENAME = "/tmp/pipecat_conversation_"
```
filipi87, line 31: nit: Should we try to retrieve this from an environemnt variable or maybe an argument ?

examples/foundational/39-aws-nova-sonic.py
```diff
@@ -0,0 +1,181 @@
...
+from pipecat.transports.network.small_webrtc import SmallWebRTCTransport
+from pipecat.transports.network.webrtc_connection import SmallWebRTCConnection
+
+# Load environment variables
+load_dotenv(override=True)
+
+# logging.basicConfig(
+#     level=logging.DEBUG,
+#     format='%(asctime)s - %(levelname)s - %(message)s'
+# )
```
filipi87, line 33: nit: I guess we can remove it, and also the `import logging`.

examples/foundational/39-aws-nova-sonic.py
```diff
@@ -0,0 +1,181 @@
...
+
+
+async def fetch_weather_from_api(function_name, tool_call_id, args, llm, context, result_callback):
```
filipi87, line 36: Same comments about using the single parameter `FunctionCallParams`.

examples/foundational/39-aws-nova-sonic.py
```diff
@@ -0,0 +1,181 @@
...
+
+    # Register function for function calls
+    # you can either register a single function for all function calls, or specific functions
+    # llm.register_function(None, fetch_weather_from_api)
+    llm.register_function("get_current_weather", fetch_weather_from_api)
+
+    # Set up context and context management.
+    # AWSNovaSonicService will adapt OpenAI LLM context objects with standard message format to
+    # what's expected by Nova Sonic.
+    # TODO: since we can't trigger a response upon joining, this isn't particularly useful
```
filipi87, line 117: nit: Is this TODO still true ? I think it's working now since your hack, at least I’m always receiving the fun fact.

src/pipecat/adapters/services/aws_nova_sonic_adapter.py
```diff
@@ -0,0 +1,40 @@
...
+                    )
+                },
+            }
+        }
+
+    def to_provider_tools_format(self, tools_schema: ToolsSchema) -> List[Dict[str, Any]]:
+        """Converts function schemas to Openai Realtime function-calling format.
+
+        :return: Openai Realtime formatted function call definition.
+        """
```
filipi87, line 37: nit:
```suggestion
        """Converts function schemas to AWS Nova Sonic function-calling format.

        :return: AWS Nova Sonic formatted function call definition.
        """
```

src/pipecat/services/aws_nova_sonic/aws.py
```diff
@@ -0,0 +1,973 @@
...
+        tools: Optional[ToolsSchema] = None,
+        send_transcription_frames: bool = True,
+        **kwargs,
+    ):
+        super().__init__(**kwargs)
+        self._secret_access_key = secret_access_key
+        self._access_key_id = access_key_id
+        self._region = region
+        self._model = model
+        self._client: BedrockRuntimeClient = None
```
filipi87, line 133: Should it be ?
```suggestion
        self._client: Optional[BedrockRuntimeClient] = None
```

src/pipecat/services/aws_nova_sonic/aws.py
```diff
@@ -0,0 +1,973 @@
...
+        self._voice_id = voice_id
+        self._system_instruction = system_instruction
+        self._tools = tools
+        self._send_transcription_frames = send_transcription_frames
+        self._context: AWSNovaSonicLLMContext = None
```
filipi87, line 138: Should it be ?
```suggestion
        self._context: Optional[AWSNovaSonicLLMContext] = None
```

src/pipecat/services/aws_nova_sonic/aws.py
```diff
@@ -0,0 +1,973 @@
...
+            InvokeModelWithBidirectionalStreamInput,
+            InvokeModelWithBidirectionalStreamOutput,
+            InvokeModelWithBidirectionalStreamOperationOutput,
+        ] = None
+        self._receive_task = None
+        self._prompt_name = None
+        self._input_audio_content_name = None
+        self._content_being_received = None
+        self._assistant_is_responding = False
+        self._context_available = False
```
filipi87, line 149: If `self._context` is `Optional`, do we still need this ?

src/pipecat/services/aws_nova_sonic/aws.py
```diff
@@ -0,0 +1,973 @@
...
+
+    async def start(self, frame: StartFrame):
+        await super().start(frame)
+        # TODO: maybe connect but don't send history until we get all of our settings?
+        # how do we know how long to wait?
+        # ah, i think we'll *always* get at least one OpenAILLMContextFrame which kicks things off
+        # so we need to send the initial history when:
+        # - we're connected
+        # - we've gotten the first context
+        # i *think* this is what's controlled by _api_session_ready/_run_llm_when_api_session_ready
```
filipi87, line 169: It looks like in some of the other LLMs, we're even creating the client and connecting inside the `__init__`. But it feels correct to create it here, as you're doing.

src/pipecat/services/aws_nova_sonic/aws.py
```diff
@@ -0,0 +1,973 @@
...
+            pass
+        elif isinstance(frame, UserStartedSpeakingFrame):
+            # print("[pk] UserStartedSpeakingFrame")
+            pass
+        elif isinstance(frame, StopInterruptionFrame):
+            # print("[pk] StopInterruptionFrame")
+            pass
+        elif isinstance(frame, UserStoppedSpeakingFrame):
+            # print("[pk] UserStoppedSpeakingFrame")
+            pass
```
filipi87, line 219: Inside the GeminiMultimodalLiveLLMService, it looks like we use the UserStartedSpeakingFrame and UserStoppedSpeakingFrame to buffer the user audio and later transcribe it. But I believe we're already receiving the user transcription here, right?

As for interruptions, my guess is that they're automatically handled by Nova Sonic as well.

So, I don't think we need to handle those frames here.

src/pipecat/services/aws_nova_sonic/aws.py
```diff
@@ -0,0 +1,973 @@
...
+
+        await self.push_frame(frame, direction)
+
+    async def _handle_context(self, context: OpenAILLMContext):
+        # TODO: reset connection if needed (if entirely new context object provided, for instance)
```
filipi87, line 224: I guess this will be needed when integrating with Pipecat Flows.

src/pipecat/services/aws_nova_sonic/aws.py
```diff
@@ -0,0 +1,973 @@
...
+            # reliably marked END_TURN (sometimes the *first* one is, but not the last...bug?)
+            #
+            # Then I considered schemes where we tally or match up SPECULATIVE text blocks with
+            # FINAL text blocks to know how many or which FINAL blocks to expect, but user
+            # interruptions throw a wrench in these schemes: depending on the exact timing of the
+            # interruption, we should or shouldn't expect some FINAL blocks.
+            await asyncio.sleep(0.25)
+            self._assistant_is_responding = False
+            await self._report_assistant_response_ended()
+            self._handling_bot_stopped_speaking = False
```
filipi87, line 267: nit: We can remove this one, since we are already setting it outside the if.

src/pipecat/services/aws_nova_sonic/aws.py
```diff
@@ -0,0 +1,973 @@
...
+        # Wait until we're done sending the assistant response trigger audio before sending audio
+        # from the user's mic
+        if self._triggering_assistant_response:
+            return
+
+        # TODO: check if _audio_input_paused? what causes that?
+        await self._send_user_audio_event(frame.audio)
+
+    async def _handle_bot_stopped_speaking(self):
+        # Protect against back-to-back BotStoppedSpeaking calls, which I've observed
```
filipi87, line 244: Interesting, I wouldn't expect to receive the `BotStoppedSpeakingFrame` multiple times. Is it easy to reproduce ?

src/pipecat/services/aws_nova_sonic/aws.py
```diff
@@ -0,0 +1,1002 @@
...
+        if not self._content_being_received:
+            return
+
+        # Get audio
+        audio_content = event_json["audioOutput"]["content"]
+        # print(f"[pk] audio output. content: {len(audio_content)}")
+
+        # Push audio frame
+        audio = base64.b64decode(audio_content)
+        # TODO: make sample rate + channels (used in multiple places) consts
```
filipi87, line 785: 👍

src/pipecat/services/aws_nova_sonic/aws.py
```diff
@@ -0,0 +1,1002 @@
...
+                    function_name=function_name,
+                    arguments=arguments,
+                )
+            elif None in self._functions.keys():
+                await self.call_function(
+                    context=self._context,
+                    tool_call_id=tool_call_id,
+                    function_name=function_name,
+                    arguments=arguments,
+                )
```
filipi87, line 823: Can we remove this if ? It looks like we are invoking the same code in both:
```
await self.call_function(
    context=self._context,
    tool_call_id=tool_call_id,
    function_name=function_name,
    arguments=arguments,
)
```

src/pipecat/services/aws_nova_sonic/aws.py
```diff
@@ -0,0 +1,1002 @@
...
+
+        self._content_being_received = False
+
+    async def _handle_completion_end_event(self, event_json):
+        # print("[pk] completion end")
+        pass
+
+    async def _report_assistant_response_started(self):
+        # Report that the assistant has started their response.
+        print("[pk] LLM full response started")
```
filipi87, line 865: Refactor the code to instead of using `print`s  use the `logger`. We are using prints in a couple of places.

src/pipecat/services/aws_nova_sonic/aws.py
```diff
@@ -0,0 +1,1002 @@
...
+        # This should never happen
+        if not self._content_being_received:
+            return
+        content = self._content_being_received
+
+        content_end = event_json["contentEnd"]
+        stop_reason = content_end["stopReason"]
+        # print(f"[pk] content end: {content}.\n  stop_reason: {stop_reason}")
+        # if content.role == Role.ASSISTANT:
+        # print(f"[pk] assistant content end: {content}.\n  stop_reason: {stop_reason}")
```
filipi87, line 839: There are a couple of places in the code where we have commented-out sections. 

Could we remove them, or, if they are needed, uncomment and use `logger.trace` instead for printing ?

src/pipecat/services/aws_nova_sonic/aws.py
```diff
@@ -0,0 +1,1002 @@
...
+    async def _report_assistant_response_ended(self):
+        # Report that the assistant has finished their response.
+        print("[pk] LLM full response ended")
+        await self.push_frame(LLMFullResponseEndFrame())
+
+        # Report that equivalent of TTS (this is a speech-to-speech model) stopped.
+        print("[pk] TTS stopped")
+        await self.push_frame(TTSStoppedFrame())
+
+        # For an explanation of this hack, see _report_assistant_response_text_added.
```
filipi87, line 902: Maybe just leave it as a `TODO:` here as well.

src/pipecat/services/aws_nova_sonic/context.py
```diff
@@ -0,0 +1,223 @@
...
+
+class AWSNovaSonicUserContextAggregator(OpenAIUserContextAggregator):
+    async def process_frame(
+        self, frame: Frame, direction: FrameDirection = FrameDirection.DOWNSTREAM
+    ):
+        await super().process_frame(frame, direction)
+
+        # Parent does not push LLMMessagesUpdateFrame
+        if isinstance(frame, LLMMessagesUpdateFrame):
+            await self.push_frame(AWSNovaSonicMessagesUpdateFrame(context=self._context))
```
filipi87, line 170: Why do we need to push the `AWSNovaSonicMessagesUpdateFrame` here ? It looks like we're not using this new frame anywhere. Just trying to understand how we intend to use it in the long term.

examples/foundational/20e-persistent-context-aws-nova-sonic.py
kompfner: Fixed! Caught it after the rebase.

examples/foundational/20e-persistent-context-aws-nova-sonic.py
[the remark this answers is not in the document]
kompfner: Fixed

examples/foundational/20e-persistent-context-aws-nova-sonic.py
kompfner: This file is pretty much a copy-paste from all the other `20*` examples. So will punt on deviating from them, if that's cool with you.

examples/foundational/39-aws-nova-sonic.py
kompfner: Fixed

examples/foundational/39-aws-nova-sonic.py
kompfner: Fixed

examples/foundational/39-aws-nova-sonic.py
kompfner: Good catch. This TODO is out of date. Removing.

filipi87 (approved):
I'm leaving my pre-approval here, since there's nothing critical in the comments and everything seems to be working fine. Feel free to push the other fixes in a follow-up PR, so we don't block this merge in case it's needed for today's release.

src/pipecat/adapters/services/aws_nova_sonic_adapter.py
kompfner: Fixed

src/pipecat/services/aws_nova_sonic/aws.py
kompfner: Wonder if we could use the `UserStoppedSpeaking` frame to better aggregate the user transcriptions. As it is today, they can end up split up into separate messages in the context. Anyway, jotting this down as a later TODO.

src/pipecat/services/aws_nova_sonic/aws.py
kompfner: Exactly

src/pipecat/services/aws_nova_sonic/aws.py
```diff
@@ -0,0 +1,973 @@
...
+            self._context_available = True
+            await self._finish_connecting_if_context_available()
+
+    async def _handle_input_audio_frame(self, frame: InputAudioRawFrame):
+        # Wait until we're done sending the assistant response trigger audio before sending audio
+        # from the user's mic
+        if self._triggering_assistant_response:
+            return
+
+        # TODO: check if _audio_input_paused? what causes that?
```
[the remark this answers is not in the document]
kompfner: Yeah, I saw that. It looks like the wait-a-bit-before-unmuting hack might not be necessary for AWS Nova Sonic? Tempted to punt on this for now.

src/pipecat/services/aws_nova_sonic/aws.py
kompfner: good catch!

kompfner: fixed

src/pipecat/services/aws_nova_sonic/aws.py
kompfner: ended up adding them as params: https://github.com/pipecat-ai/pipecat/pull/1704/commits/650afeed32501908108021b577e6129b8b049dc1. wdyt?

filipi87: I think it makes sense. But what if the user configures a different sample rate? Will it be supported? If not, are we going to resample it or throw an exception?

src/pipecat/services/aws_nova_sonic/aws.py
kompfner: Fixed (ini both places)

src/pipecat/services/aws_nova_sonic/aws.py
kompfner: Fixed

src/pipecat/services/aws_nova_sonic/aws.py
kompfner: Fixed

src/pipecat/services/aws_nova_sonic/aws.py
filipi87: Oh, I see, Yeah, I think you are right, the single if should be enough.

src/pipecat/services/aws_nova_sonic/aws.py
kompfner: Done

src/pipecat/services/aws_nova_sonic/aws.py
kompfner: Fixed. A lot cleaner, thanks for the suggestion!

PR #1707, opened 2025-04-30
[Rime] Add new params to RimeTTSService

MaCaki:

Adding the following params to the RimeTTSService params, which are supported by the rime websockets api: 

* [pauseBetweenBrackets](https://docs.rime.ai/api-reference/endpoint/websockets-json#param-pause-between-brackets)
* [phonemizeBetweenBrackets](https://docs.rime.ai/api-reference/endpoint/websockets-json#param-phonemize-between-brackets)

We json serialize the bool values as the rime api currently expects `true/false`, and is case-sensitive (patch to rime ws backend incoming)

### DIFF
--- a/src/pipecat/services/rime/tts.py
+++ b/src/pipecat/services/rime/tts.py
@@ -68,6 +68,8 @@ class RimeTTSService(AudioContextWordTTSService):
         language: Optional[Language] = Language.EN
         speed_alpha: Optional[float] = 1.0
         reduce_latency: Optional[bool] = False
+        pause_between_brackets: Optional[bool] = False
+        phonemize_between_brackets: Optional[bool] = False
 
     def __init__(
         self,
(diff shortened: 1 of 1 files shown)


src/pipecat/services/rime/tts.py
```diff
@@ -117,6 +119,8 @@ def __init__(
             else "eng",
             "speedAlpha": params.speed_alpha,
             "reduceLatency": params.reduce_latency,
+            "pauseBetweenBrackets": json.dumps(params.pause_between_brackets),
```
markbackman, line 122: I don't think json.dumps is required here as `phonemize_between_brackets` is just a bool. Same for `pause_between_brackets`.

```suggestion
            "pauseBetweenBrackets": params.pause_between_brackets,
```

markbackman (commented):
Thanks for the submission! Just one comment; otherwise, looks good.

src/pipecat/services/rime/tts.py
MaCaki: when these are stringified into url params with `"&".join(f"{k}={v}" for k, v in self._settings.items())`, python `True` becomes `True`, while the rime ws api is currently looking for `true`, which is the json bool. 

We're rolling out a change next week to be case insensitve about bool strings, but just wanted to make this patch in the mean time. 

lmk what you think. 

Given that we'll be making that change pretty soon, I'm also ok just stringifying the python bool without worrying about upper/lower case

markbackman: Got it. Makes sense. Thank you.

markbackman (approved):
LGTM.

@MaCaki Can you just add a brief changelog entry for this one?

PR #1708, opened 2025-05-01
Push GeminiMultimodalLiveLLMService TranscriptionFrame Upstream, remo…

markbackman:
…ve direct context addition

I found that GeminiMultimodalLiveLLMService was previously just adding content directly to the context. Instead, we should rely on the aggregator. So, I'm pushing the TranscriptionFrame upstream now.

A nice benefit is that we can remove the hack in the `GeminiMultimodalLiveAssistantContextAggregator` where we were ignoring TranscriptionFrames.

In addition, I want to show how to use this frame to get transcripts, so I'm updating 26a to do that. This is helpful as the placement is unusual. Normally, this would be before the user context aggregator, but because the user context aggregator doesn't push TranscriptionFrames through the pipeline, it has to come after in this case, as the LLM is pushing the TranscriptionFrame.

### DIFF
--- a/examples/foundational/26a-gemini-multimodal-live-transcription.py
+++ b/examples/foundational/26a-gemini-multimodal-live-transcription.py
@@ -12,10 +12,12 @@ from loguru import logger
 
 from pipecat.audio.vad.silero import SileroVADAnalyzer
 from pipecat.audio.vad.vad_analyzer import VADParams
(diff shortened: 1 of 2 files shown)


aconchillo (approved):

aconchillo:
LGTM

PR #1711, opened 2025-05-02
small updates to groq

kwindla:
1. GroqSTTService events come in after UserStoppedSpeaking but the transcription result arrives all at once. So the default aggregation timeout in the user aggregator of 500ms was noticeably slowing things down for no reason. Added code to change the timeout in both groq examples.

2. Updated the examples to use llama 4 maverick.

3. Updated the groq library version in `pyproject.toml`


filipi87 (approved):
LGTM!

markbackman (approved):
LGTM

Issue #1712, opened 2025-05-02
GeminiMultimodalModalities.TEXT Stopped working


FaBeyyy:
### pipecat version

latest main branch

### Python version

3.12

### Operating System

macos 15

### Issue description

Using  ```modalities=GeminiMultimodalModalities.TEXT```with ```GeminiMultimodalLiveLLMService``` completely stopped producing any output in the current main branch compared to the latest release 0.6.4

### Reproduction steps

Reproduction steps

Clone the repo and check out the latest main branch.

Set up GeminiMultimodalLiveLLMService with modalities=GeminiMultimodalModalities.TEXT.

Run the service and send user audio input

Observe that no output is returned.

### Expected behavior

Text input should be processed and an appropriate response should be returned, as it does in version 0.6.4.

### Actual behavior

No response is returned from the service

### Logs

```shell

```

markbackman:
You're right. Fixing it here: https://github.com/pipecat-ai/pipecat/pull/1714. Apologies!

FaBeyyy:
> You're right. Fixing it here: [#1714](https://github.com/pipecat-ai/pipecat/pull/1714). Apologies!

Thanks! Appreciate the quick response :)

PR #1714, opened 2025-05-02
Restore TEXT modalities support to GeminiMultimodalLiveLLMService

markbackman:

Mistakenly removed support for this when fixing the context.

Fixes: #1712

### DIFF
--- a/src/pipecat/services/gemini_multimodal_live/gemini.py
+++ b/src/pipecat/services/gemini_multimodal_live/gemini.py
@@ -354,6 +354,7 @@ class GeminiMultimodalLiveLLMService(LLMService):
         self._bot_is_speaking = False
         self._user_audio_buffer = bytearray()
         self._bot_audio_buffer = bytearray()
+        self._bot_text_buffer = ""
 
         self._sample_rate = 24000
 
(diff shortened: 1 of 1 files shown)


src/pipecat/services/gemini_multimodal_live/gemini.py
```diff
@@ -892,13 +902,23 @@ async def _handle_evt_tool_call(self, evt):
...
         self._bot_is_speaking = False
+        text = self._bot_text_buffer
+        self._bot_text_buffer = ""
+
+        # Pertains to modalities set to TEXT only
+        if text:
+            await self.push_frame(LLMFullResponseEndFrame())
+
         await self.push_frame(TTSStoppedFrame())
         await self.push_frame(LLMFullResponseEndFrame())
```
filipi87, line 913: Are we not sending `LLMFullResponseEndFrame` twice in this case ? Also, should we be triggering `TTSStoppedFrame` ? 

Maybe this should be something like this:
```
if text:
else:
    await self.push_frame(TTSStoppedFrame())
```

filipi87: Actually I guess even if the modality is `text`, we still wish to trigger the `TTSStoppedFrame`, right ?

markbackman: Good catch. Maybe we want:
```
        # Only push the TTSStoppedFrame the bot is outputting audio
        # when text is found, modalities is set to TEXT and no audio
        # is produced.
        if not text:
            await self.push_frame(TTSStoppedFrame())

```

filipi87 (approved):
LGTM!

PR #1719, opened 2025-05-02
Add direct muting to the STTMuteFilter

markbackman:

This came up in a conversation with someone from the community. It's a bit of a different way for the STTMuteFilter to be used, but it can help in cases where you want to proactively mute based on activity in the pipeline. For example, the user cited wanting to ignore user input for the goodbye message. You could imagine pushing a RequestSTTMuteFrame when triggering a goodbye message, which makes that flow easier to build.

You could alternatively build for the end message use case with a CUSTOM strategy. I'm torn about offering this functionality, as you can mistakenly leave the user muted, if you're not managing frames correctly.

One oddity. In writing tests, the STTMuteFrame comes before the RequestSTTMuteFrame. I don't really understand why this is the case. @aconchillo maybe you know?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- Added two new frames `RequestSTTMuteFrame`, `RequestSTTUnmuteFrame`. These
+  frames tell the `STTMuteFilter` to directly mute or unmute the user and
+  take precedent over the mute strategies when an `RequestSTTMuteFrame` is
+  processed.
+
 - `BaseOutputTransport` now allows multiple destinations if the transport
   implementation supports it (e.g. Daily's custom tracks). With multiple
   destinations it is possible to send different audio or video tracks with a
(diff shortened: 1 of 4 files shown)


markbackman:
Thinking about this just for a bit longer, I don't like it. I'd much rather someone use the CUSTOM strategy plus a callback to accomplish the same thing. Ex:
```
   # Custom mute callback that checks session state
    async def session_timeout_mute_callback(stt_filter: STTMuteFilter) -> bool:
        return session_manager.is_ending
    
    # Configure the mute processor with the custom callback
    stt_mute_processor = STTMuteFilter(
        config=STTMuteConfig(
            strategies={
                STTMuteStrategy.MUTE_UNTIL_FIRST_BOT_COMPLETE,
                STTMuteStrategy.CUSTOM,
            },
            should_mute_callback=session_timeout_mute_callback,
        ),
    )
```

where the session_manager is some class in the code that can be updated.

The CUSTOM callback is consulted everytime the bot speaks, so it scopes muting the user to preventing interruptions. This will avoid state mismanagement, which could cause bigger problems.

Closing out this PR for that reason, but glad I documented this approach in case it's useful again in the future.

rahultyl:
Thanks @markbackman for taking this up. I agree frame based approach can be prone to state mismanagement. [Documentation](I added more detail to the docs on how to use CUSTOM, including citing your case: https://docs.pipecat.ai/server/utilities/filters/stt-mute#custom-muting-logic) is more clear now on how to use multiple strategies and custom strategy.

PR #1723, opened 2025-05-02
Only display the destination in the bot started/stopped speaking log …

markbackman:
…when there is a desintation

### DIFF
--- a/src/pipecat/transports/base_output.py
+++ b/src/pipecat/transports/base_output.py
@@ -380,7 +380,9 @@ class BaseOutputTransport(FrameProcessor):
 
         async def _bot_started_speaking(self):
             if not self._bot_speaking:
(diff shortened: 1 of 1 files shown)


src/pipecat/transports/base_output.py
```diff
@@ -380,7 +380,9 @@ async def _cancel_audio_task(self):
 
         async def _bot_started_speaking(self):
             if not self._bot_speaking:
-                logger.debug(f"Bot [{self._destination}] started speaking")
+                logger.debug(
+                    f"Bot{f' [{self._destination}]' if self._destination is not None else ''} started speaking"
```
aconchillo, line 384: small simplification:

```
[{self._destination}]' if self._destination else ''}
```

aconchillo (approved):

PR #1729, opened 2025-05-03
Add support for OpenTelemetry tracing

markbackman:


src/pipecat/processors/metrics/frame_processor_metrics.py
```diff
@@ -23,8 +24,25 @@ class FrameProcessorMetrics:
     def __init__(self):
         self._start_ttfb_time = 0
         self._start_processing_time = 0
+        self._last_ttfb_time = 0
         self._should_report_ttfb = True
 
+    @property
+    def ttfb_ms(self) -> Optional[float]:
+        """Get the current TTFB value in milliseconds.
```
aconchillo, line 32: I believe these are seconds. `time.time()` would return 0.5 for 500ms.

markbackman: Yup! Thank you. I had the time originally in ms, but switched to sec to match our other metrics.

src/pipecat/services/aws/tts.py
```diff
@@ -247,4 +249,28 @@ def read_audio_data(**args):
...
+                    span=current_span,
+                    service_name=service_name,
+                    model=self._settings.get("engine", "standard"),
+                    voice_id=self._voice_id,
+                    text=text,
+                    settings=self._settings,
+                    character_count=len(text),
+                    operation_name="tts",
+                    ttfb_ms=ttfb_ms,
+                )
```
aconchillo, line 274: I'm wondering if we could create annotations for all this so we don't have to write these things every time.

```
    @tracing(name="polly_tts")
    async def run_tts(self, text: str) -> AsyncGenerator[Frame, None]:
```
@tracing would call:

```
    @traced(attachment_strategy=AttachmentStrategy.CHILD, name="polly_tts")
```

One concern I have is that this is OpenTelemetry specific. I think we should make it so you can replace the service you want to use for telemetry in an easy way.

src/pipecat/utils/tracing/__init__.py
```diff
@@ -0,0 +1,7 @@
+#
```
aconchillo, line 1: Maybe this should all go to a new `tracing` package. And then have `tracing/opentelemetry` .

aconchillo:
Two main comments:

- I think we should allow more telemetry frameworks.
- There should be a function that returns the telemetry data. This function would be called with our decorators. The data would be generic so we can use it/adapat it to any telemetry framework.

filipi87 (approved):
Looks good. Both examples appear to be working fine.

src/pipecat/services/aws/tts.py
filipi87: As far as I could understand, OpenTelemetry is an open-source framework whose goal is to unify tracing, metrics, and logs into a single standard.

And it looks like most vendors support the OpenTelemetry Protocol: https://opentelemetry.io/ecosystem/vendors/

And even in cases where that’s not possible, we should be able to write a custom exporter to create anything custom and send it to any destination.

So, I don't think it would be bad to rely on OpenTelemetry for that. But maybe I am still missing something.

Is there any particular reason why are you concerned about being OpenTelemetry specific for structuring and collecting our traces ?

filipi87:
> I think we should allow more telemetry frameworks.
There should be a function that returns the telemetry data. This function would be called with our decorators. The data would be generic so we can use it/adapat it to any telemetry framework.

I think this can still be the case using OpenTelemetry, right?

Like, I believe OpenTelemetry provides a standard that works with any framework. We would use it for collecting traces, metrics, and logs, but we can export that data to any other framework of our choice.

And if a framework doesn’t natively support OpenTelemetry Protocol, we can write a custom exporter to send the data to any destination.

Does it make sense ?

kompfner:
> I think we should allow more telemetry frameworks.

@aconchillo I initially had a similar hunch that it might benefit us to architect our solution so that there's an "intermediate" layer where a customer could choose to either use the OTel SDK (the default) or different ones—like Datadog's `ddtrace`—to send data (allowing them to not even import OTel SDK). A customer might already be wedded to their telemetry framework of choice and might not want to have to learn how to wire up OTel data to Datadog, for example.

Though...I suppose this might depend on how easy it is to build a custom OTel SpanExporter, because maybe OTel _can_ be the easy-enough intermediate layer, as @filipi87 points out, and we don't have to reinvent the wheel and provide our own. And maybe from within the custom OTel SpanExporter, the customer can make the appropriate `ddtrace` calls _without_ having to have the OTel SDK actually send the data at all. The OTel SDK would just be a "thin" abstraction layer, into which you can plug whatever telemetry-sending tools you want.

Maybe we should just try and see what it takes to get OTel spans and send them via `ddtrace`, to sanity-check that that'd work.

filipi87:
Just to complement what @kompfner said, have you seen how extensive the list of vendors who native support OpenTelemetry ?
- https://opentelemetry.io/ecosystem/vendors/

For example, Datadog is one of them: https://docs.datadoghq.com/opentelemetry/

kompfner:
> Just to complement what @kompfner said, have you seen how extensive the list of vendors who native support OpenTelemetry ?
> 
> * https://opentelemetry.io/ecosystem/vendors/
> 
> For example, Datadog is one of them: https://docs.datadoghq.com/opentelemetry/

I understand that people can point OTel output directly to an ingestion endpoint in their vendor of choice. 

But I'm fairly confident they will not always want to do that. Datadog docs, for example, are very clear that OTel output provides a limited subset of the stuff that their own libraries gather.

We need to make sure our escape hatch works so that customers can do _anything they want_ with what the OTel library is gathering.

src/pipecat/observers/turn_tracking_observer.py
```diff
@@ -0,0 +1,173 @@
...
+        if isinstance(data.frame, StartFrame):
+            # Start the first turn immediately when the pipeline starts
+            if self._turn_count == 0:
+                await self._start_turn(data.timestamp)
+        elif isinstance(data.frame, UserStartedSpeakingFrame):
+            await self._handle_user_started_speaking(data.timestamp)
+        elif isinstance(data.frame, BotStartedSpeakingFrame):
+            self._is_bot_speaking = True
+            # Cancel any pending turn end timer when bot starts speaking again
+            self._cancel_turn_end_timer()
```
aconchillo, line 75: Add `await self._handle_bot_started_speaking(data)`

src/pipecat/observers/turn_tracking_observer.py
```diff
@@ -0,0 +1,173 @@
...
+        if len(self._processed_frames) > len(self._frame_history):
+            # Rebuild the set from the current deque contents
+            self._processed_frames = set(self._frame_history)
+
+        if isinstance(data.frame, StartFrame):
+            # Start the first turn immediately when the pipeline starts
+            if self._turn_count == 0:
+                await self._start_turn(data.timestamp)
+        elif isinstance(data.frame, UserStartedSpeakingFrame):
+            await self._handle_user_started_speaking(data.timestamp)
```
aconchillo, line 71: Usually better to pass `data` in case you need other stuff.

src/pipecat/observers/turn_tracking_observer.py
```diff
@@ -0,0 +1,173 @@
...
+        elif isinstance(data.frame, BotStartedSpeakingFrame):
+            self._is_bot_speaking = True
+            # Cancel any pending turn end timer when bot starts speaking again
+            self._cancel_turn_end_timer()
+        elif isinstance(data.frame, BotStoppedSpeakingFrame) and self._is_bot_speaking:
+            self._is_bot_speaking = False
+            # Schedule turn end with timeout
+            # This is needed to handle cases where the bot's speech ends and then resumes
+            # This can happen with HTTP TTS services or function calls
+            self._schedule_turn_end(data.timestamp)
```
aconchillo, line 81: Also, add a handler.

src/pipecat/observers/turn_tracking_observer.py
```diff
@@ -0,0 +1,173 @@
...
+from pipecat.frames.frames import (
+    BotStartedSpeakingFrame,
+    BotStoppedSpeakingFrame,
+    StartFrame,
+    UserStartedSpeakingFrame,
+)
+from pipecat.observers.base_observer import BaseObserver, FramePushed
+
+
+class TurnTrackingObserver(BaseObserver):
```
aconchillo, line 21: It would be nice to add a test for this.

src/pipecat/services/anthropic/llm.py
```diff
@@ -147,6 +148,7 @@ def create_context_aggregator(
         assistant = AnthropicAssistantContextAggregator(context, params=assistant_params)
         return AnthropicContextAggregatorPair(_user=user, _assistant=assistant)
 
+    @traced_llm(name="anthropic_process_context")
```
aconchillo, line 151: Could we have a default `name` for all these decorators?

aconchillo: So you just have to do

```
@traced_llm
async def whatever
```

and it can use the class type for example. Not sure how these names are used.

src/pipecat/services/stt_service.py
```diff
@@ -104,6 +104,39 @@ async def process_frame(self, frame: Frame, direction: FrameDirection):
...
+
+        # Build basic attributes
+        attributes = {
+            "service_name": service_name,
+            "model": getattr(self, "model_name", "unknown"),
+        }
+
+        # Add language if available
+        if hasattr(self, "_settings") and "language" in self._settings:
+            attributes["language"] = self._settings["language"]
```
aconchillo, line 129: Maybe we can add a `TODO`. It feels we should have a `language` property.

src/pipecat/services/stt_service.py
```diff
@@ -104,6 +104,39 @@ async def process_frame(self, frame: Frame, direction: FrameDirection):
...
+        Returns a dictionary of attributes that can be used for tracing.
+
+        Args:
+            **kwargs: Additional attributes to include
+
+        Returns:
+            dict: Attributes for this service
+        """
+        # Extract service name from class name
+        service_name = self.__class__.__name__.replace("STTService", "").lower()
```
aconchillo, line 119: This assume all classes will have the `STTService` word, why remove it?

src/pipecat/utils/tracing/service_decorators.py
```diff
@@ -0,0 +1,455 @@
...
+                        # Update TTFB metric at the end
+                        ttfb_ms = getattr(getattr(self, "_metrics", None), "ttfb_ms", None)
+                        if ttfb_ms is not None:
+                            current_span.set_attribute("metrics.ttfb_ms", ttfb_ms)
+
+            return wrapper
+
+    if func is not None:
+        return decorator(func)
+    return decorator
```
aconchillo, line 210: Maybe we can remove duplicated code:

```
def traced_tts(func: Optional[Callable] = None, *, name: Optional[str] = None) -> Callable:
    if not OPENTELEMETRY_AVAILABLE:
        return _noop_decorator if func is None else _noop_decorator(func)

    def decorator(f):
        # Check if we're dealing with a coroutine or an async generator
        is_async_generator = inspect.isasyncgenfunction(f)

        @functools.wraps(f)
        async def wrapper(self, text, *args, **kwargs):
            if not is_tracing_available():
                if is_async_generator:
                    async for item in f(self, text, *args, **kwargs):
                        yield item
                    return
                else:
                    return await f(self, text, *args, **kwargs)


            # Get the turn context first, then fall back to service context
            turn_context = get_current_turn_context()
            parent_context = turn_context or _get_parent_service_context(self)

            # Create a new span as child of the turn span or service span
            tracer = trace.get_tracer("pipecat")
            with tracer.start_as_current_span(
                    name or f.__name__, context=parent_context
            ) as current_span:
                try:
                    # Immediately add attributes to the span
                    service_name = self.__class__.__name__.replace("TTSService", "").lower()

                    # Add TTS attributes right away
                    add_tts_span_attributes(
                        span=current_span,
                        service_name=service_name,
                        model=getattr(self, "model_name", "unknown"),
                        voice_id=getattr(self, "_voice_id", "unknown"),
                        text=text,
                        settings=getattr(self, "_settings", {}),
                        character_count=len(text),
                        operation_name="tts",
                        cartesia_version=getattr(self, "_cartesia_version", None),
                        context_id=getattr(self, "_context_id", None),
                    )

                    if is_async_generator:
                        # For async generators, we need to yield from it
                        async for item in f(self, text, *args, **kwargs):
                            yield item
                    else:
                        return await f(self, text, *args, **kwargs)

                except Exception as e:
                    # Log any exception but don't disrupt the main flow
                    logging.warning(f"Error in TTS tracing: {e}")
                    raise
                finally:
                    # Update TTFB metric at the end
                    ttfb_ms = getattr(getattr(self, "_metrics", None), "ttfb_ms", None)
                    if ttfb_ms is not None:
                        current_span.set_attribute("metrics.ttfb_ms", ttfb_ms)

            return wrapper

    if func is not None:
        return decorator(func)
    return decorator
```

aconchillo: Mainly doing this:

```
                if is_async_generator:
                    async for item in f(self, text, *args, **kwargs):
                        yield item
                    return
                else:
                    return await f(self, text, *args, **kwargs)
```

src/pipecat/utils/tracing/service_decorators.py
```diff
@@ -0,0 +1,455 @@
...
+                            elif hasattr(context, "messages"):
+                                messages = context.messages
+
+                            # For OpenAI-style messages, use standard JSON serialization
+                            if messages:
+                                try:
+                                    # Standard message format should serialize fine
+                                    serialized_messages = json.dumps(messages)
+                                except Exception as e:
+                                    serialized_messages = f"Error serializing messages: {str(e)}"
```
aconchillo, line 372: we could move this block out of the paren if/else, so:

```
                      if is_google_service:
                            # Handle Google service specifically
                            if hasattr(context, "get_messages_for_logging"):
                                messages = context.get_messages_for_logging()
                        else:
                            # Handle other services like OpenAI
                            if hasattr(context, "get_messages"):
                                messages = context.get_messages()
                            elif hasattr(context, "messages"):
                                messages = context.messages

                        if messages:
                            try:
                                # Standard message format should serialize fine
                                serialized_messages = json.dumps(messages)
                            except Exception as e:
                                serialized_messages = f"Error serializing messages: {str(e)}"

```

src/pipecat/utils/tracing/service_decorators.py
```diff
@@ -0,0 +1,455 @@
...
+
+                    try:
+                        # Detect if we're using Google's service
+                        is_google_service = "google" in service_name.lower()
+
+                        # Try to get messages based on service type
+                        messages = None
+                        serialized_messages = None
+
+                        if is_google_service:
```
aconchillo, line 348: we really should unify these so we don't do specific service things. Add a TODO.

examples/open-telemetry-tracing/bot.py
```diff
@@ -0,0 +1,156 @@
...
+from pipecat.services.openai.llm import OpenAILLMService
+from pipecat.transports.base_transport import TransportParams
+from pipecat.transports.network.small_webrtc import SmallWebRTCTransport
+from pipecat.transports.network.webrtc_connection import SmallWebRTCConnection
+from pipecat.utils.tracing.setup import setup_tracing
+
+load_dotenv(override=True)
+
+# Initialize tracing if enabled
+if os.getenv("ENABLE_TRACING", "false").lower() == "true":
```
aconchillo, line 34: Maybe we can just simplify and do:

```
if os.getenv("ENABLE_TRACING"):
```

just requiring that it's set.

src/pipecat/services/aws/tts.py
aconchillo: We discussed this with @markbackman . I thought it was cleaner because you can then have more control in your `@tracing` decorator which is what Mark ended up implementing. For example, you can have turn-based spans, etc. Also, as I will comment in the other thread, it's possible that we might to do something different than OpenTelemetry, I'll add more details there.

aconchillo:
> > I think we should allow more telemetry frameworks.
> > There should be a function that returns the telemetry data. This function would be called with our decorators. The data would be generic so we can use it/adapat it to any telemetry framework.
> 
> I think this can still be the case using OpenTelemetry, right?
> 
> Like, I believe OpenTelemetry provides a standard that works with any framework. We would use it for collecting traces, metrics, and logs, but we can export that data to any other framework of our choice.
> 
> And if a framework doesn’t natively support OpenTelemetry Protocol, we can write a custom exporter to send the data to any destination.
> 
> Does it make sense ?

Yes, absolutely. What's currently implemented works great for my comment. In the first implementation we had boilerplate code in every TTS service each with it's own specific attributes, etc. We now have a unified function in each service base class and each of the services just has one decorator. This will be much easier to maintain in case we need to add something non-OpenTelemetry. I understand we can add our exporters to OpenTelemetry, but this is a bit different. Having these functions allow you to add other things.

aconchillo:
> > I think we should allow more telemetry frameworks.
> 
> @aconchillo I initially had a similar hunch that it might benefit us to architect our solution so that there's an "intermediate" layer where a customer could choose to either use the OTel SDK (the default) or different ones—like Datadog's `ddtrace`—to send data (allowing them to not even import OTel SDK). A customer might already be wedded to their telemetry framework of choice and might not want to have to learn how to wire up OTel data to Datadog, for example.
> 
> Though...I suppose this might depend on how easy it is to build a custom OTel SpanExporter, because maybe OTel _can_ be the easy-enough intermediate layer, as @filipi87 points out, and we don't have to reinvent the wheel and provide our own. And maybe from within the custom OTel SpanExporter, the customer can make the appropriate `ddtrace` calls _without_ having to have the OTel SDK actually send the data at all. The OTel SDK would just be a "thin" abstraction layer, into which you can plug whatever telemetry-sending tools you want.
> 
> Maybe we should just try and see what it takes to get OTel spans and send them via `ddtrace`, to sanity-check that that'd work. That would be a pretty valuable public example out the gate, too.

Yes, absolutely. But there might be more things we want to add a part than what OpenTelemtry gives us. For example, there's Weights&Biases which has their own decorators and can for example send audio from the session to their server. What I was suggesting was abstracting things a little bit and make them less OpenTelemetry oriented by just having our own decorator, which allows us to have more control, such as the turn-based spans which @markbackman has implemented. I believe it will make it easier to integrate it with other tracing facilities.

And with tracing I don't mean tracing as we understand it but something a bit wider.

I'm very happy for how the PR ended. It's still OpenTelemtry oriented but we have the potential to easily adapt it without changing the services too much.

src/pipecat/observers/turn_tracking_observer.py
markbackman: I added tests. We don't have an observer pattern for `run_test`. I made the observer a FrameProcessor, so it could function within the unit test frame work. We can change this later, but I think it does the trick. I was able to test the main conditions I was interested in.

src/pipecat/services/anthropic/llm.py
markbackman: Great idea! Just added this.

src/pipecat/observers/turn_tracking_observer.py
aconchillo: Just wrote this https://github.com/pipecat-ai/pipecat/pull/1806.

src/pipecat/utils/tracing/service_decorators.py
```diff
@@ -81,6 +81,31 @@ def _add_token_usage_to_span(span, token_usage):
...
+
+    # Remove common service type suffixes
+    if service_type in suffixes and class_name.endswith(suffixes[service_type]):
+        service_name = class_name[: -len(suffixes[service_type])]
+    else:
+        service_name = class_name
+
+    # Return a name in the format "servicename_type"
+    # Examples: "openai_llm", "deepgram_stt", "cartesia_tts"
+    return f"{service_name}_{service_type}"
```
aconchillo, line 106: I don't think we should require all classes to be named `XXXSTTService`. We could, but maybe people would do `XXXSTTSomethingService` which would break this.

Maybe we could just use the full class name. So, the span name would be `llm_openaillmservice`. This would allow you to match the specific class name, with `openai_llm` you don't really know which service you were running. If you could search for spans that are `llm_*` that would also be useful.

aconchillo: And it's fine if it's more verbose.

markbackman: Good point! This is cleaner.

src/pipecat/services/stt_service.py
markbackman: Maybe it's better to just leave the name as the service name, like CartesiaTTSService. WDYT?

markbackman: That's what I went for. I think this is helpful to have logged.

src/pipecat/utils/tracing/service_decorators.py
markbackman: Agreed. I'm proposing a different approach which simplifies things too.

src/pipecat/observers/turn_tracking_observer.py
markbackman: Updated! Thank you. We should write tests for the other observers too. A future todo!

src/pipecat/utils/tracing/service_decorators.py
```diff
@@ -0,0 +1,455 @@
...
+                        tool_count = 0
+
+                        if tools:
+                            try:
+                                if is_google_service:
+                                    # Special handling for Google tools
+                                    serialized_tools = json.dumps(tools)
+                                else:
+                                    # Standard serialization for other services
+                                    serialized_tools = json.dumps(tools)
```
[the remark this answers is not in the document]
markbackman: Nothing. I just left over from iterating on formats. Fixing.

src/pipecat/utils/tracing/__init__.py
markbackman: I think we opted to keep it as is, yes?

src/pipecat/utils/tracing/service_decorators.py
markbackman: added

src/pipecat/utils/tracing/service_decorators.py
```diff
@@ -0,0 +1,408 @@
...
+
+        @contextlib.asynccontextmanager
+        async def tracing_context(self, text):
+            """Async context manager for TTS tracing."""
+            if not is_tracing_available():
+                yield None
+                return
+
+            service_class_name = self.__class__.__name__
+            span_name = f"tts_{service_class_name.lower()}"
```
aconchillo, line 117: the function you had was good, so you can use it for all services:

```
_get_service_name(self, "tts")
```

which is

```
           service_class_name = self.__class__.__name__.lower()
           span_name = f"{sevice_prefix}_{service_class_name}"
```

markbackman: Good point. I've added this back.

src/pipecat/utils/tracing/__init__.py
aconchillo: yes. this was old.

aconchillo:
LGTM!

aconchillo (approved):

PR #1731, opened 2025-05-04
Remove Color conversion by using YUV420 directly; Stop sending silenc…

Antonyesk601:
Fix bad performance on Pipecat Cloud and Arm64 due to color space conversion. Fix issue making OpenAI Realtime mode not work with Simli Service


markbackman:
I don't think this is still an issue. If so, please reopen.

PR #1733, opened 2025-05-05
Adding real-time throttling to the audio mixer.

filipi87:
Adding real-time throttling to the audio mixer.


src/pipecat/transports/base_output.py
```diff
@@ -454,6 +457,11 @@ async def with_mixer(vad_stop_secs: float) -> AsyncGenerator[Frame, None]:
                             num_channels=self._params.audio_out_channels,
                         )
                         yield frame
+                    # Throttle frame generation to maintain real-time pacing and avoid flooding with silence
+                    wait = next_frame_time - time.time()
+                    if wait > 0:
+                        await asyncio.sleep(wait)
+                    next_frame_time += chunk_duration
```
aconchillo, line 464: This should not be necessary as long as the consumer reads the frames properly. That is, it's the responsibility of the transport (write frames) to control the timing. 

```
                    await self._transport.write_raw_audio_frames(frame.audio, self._destination)
```

This blocks for the write amount of time.

aconchillo: That is, `yield` waits for the next `async for` loop iteration.

src/pipecat/transports/network/small_webrtc.py
```diff
@@ -484,9 +484,10 @@ def __init__(
         self._params = params
 
     async def start(self, frame: StartFrame):
-        await super().start(frame)
         await self._client.setup(self._params, frame)
         await self._client.connect()
+        # Parent start.
+        await super().start(frame)
```
filipi87, line 490: @aconchillo, this is what I mentioned that also fixes the issue.

src/pipecat/transports/base_output.py
aconchillo: I think I like the second approach better. I had to do the same in the Daily transport actually. There's probably a better but I didn't have time to think about it.

aconchillo: I think we shouldn't add code than in theory we don't really need.

filipi87: I see your point, but I'm not sure that it's really unnecessary code.

The way I see it, this code is preventing `with_mixer` from consuming all the CPU, and enforcing it to generate audio (silence) only for that specific chunk duration.

filipi87:
Closing this PR in favor of this one, where we are only fixing how we are initilizing the SmallWebRTCTransport.:
https://github.com/pipecat-ai/pipecat/pull/1744

src/pipecat/transports/base_output.py
aconchillo: Yes, the thing is that we shouldn't really need it if things were properly initialized because `write_frames` already takes care of it.

PR #1744, opened 2025-05-05
Fixed an audio mixer issue when used alongside `SmallWebRTCTransport`.

filipi87:
Fixed an audio mixer issue when used alongside `SmallWebRTCTransport`.


aconchillo (approved):

filipi87:
We don't need this anymore, @aconchillo fixed in this PR using a different approach: https://github.com/pipecat-ai/pipecat/pull/1745#pullrequestreview-2816137682

PR #1746, opened 2025-05-05
Simple chatbot: React Native client

filipi87:
Simple chatbot: React Native client


examples/simple-chatbot/client/react-native/README.md
```diff
@@ -0,0 +1,89 @@
...
+
+#### Running on Android
+
+After plugging in an Android device [configured for debugging](https://developer.android.com/studio/debug/dev-options), run the following command:
+
+```
+npm run android
+```
+
+#### Running on iOS
```
kompfner, line 47: Hilarious how much more explanation "Running on iOS" requires than "Running on Android" 😆 😭

examples/simple-chatbot/client/react-native/src/hooks/useVoiceClientNavigation.ts
```diff
@@ -0,0 +1,21 @@
...
+import { useNavigation } from '@react-navigation/native';
+import { useVoiceClient } from '../context/VoiceClientContext';
+
+export const useVoiceClientNavigation = () => {
+  const navigation = useNavigation();
+  const { inCall } = useVoiceClient();
+
+  useEffect(() => {
+    if (inCall) {
+      // TODO, refactor this
```
kompfner, line 11: Is there still something you wanted TODO here before merging?

examples/simple-chatbot/client/react-native/src/context/VoiceClientContext.tsx
```diff
@@ -0,0 +1,229 @@
...
+      .on("remoteAudioLevel", (level: number) => {
+        if (botSpeakingRef.current) {
+          setRemoteAudioLevel(level)
+        }
+      })
+      .on("userStartedSpeaking", () => {
+        // nothing to do here
+      })
+      .on("userStoppedSpeaking", () => {
+        setLocalAudioLevel(0)
```
kompfner, line 106: Just to check my understanding: the reason we explicitly set audio level to 0 here is because the Pipecat RN client might report audio levels slightly above 0 even when the user is not determined to be speaking by VAD, but we don't want those above-0 values showing up in our visualization until VAD thinks the user _is_ speaking?

Hm, wait, that can't be the reason...because "userStoppedSpeaking" is a one-time event. If we wanted the visualization to be zero-ed out until "userStartedSpeaking" we would need to track the state `isUserSpeaking` and make sure that `localAudioLevel` is 0 until `isUserSpeaking` is `true`...

So I'm not sure what the reason is.

examples/simple-chatbot/client/react-native/src/hooks/useVoiceClientNavigation.ts
filipi87: Yep, I have just fixed the typescript warning. Thank you.

examples/simple-chatbot/client/react-native/src/context/VoiceClientContext.tsx
filipi87: > Just to check my understanding: the reason we explicitly set audio level to 0 here is because the Pipecat RN client might report audio levels slightly above 0 even when the user is not determined to be speaking by VAD, but we don't want those above-0 values showing up in our visualization until VAD thinks the user is speaking?

Actually, I implemented this for the bot. So, when the bot is not speaking, the interface should behave as expected. Otherwise, you're right, we keep receiving audio levels slightly above 0.

But we don't need to do that for the local audio level. I'm going to remove it.

filipi87: done.

filipi87:
Thank you @kompfner . 🙌

PR #1748, opened 2025-05-06
task manager dictionary and cleanup PipelineTask

aconchillo:

A couple of changes:

- If for some reason Pipeline task is cancelled from outside Pipecat we were not cleaning up things properly.
- Also, we use a dictionary instead of a set of tasks. For some reason we were having issues and dangling tasks were being reported after being removed from the set. Anyways, using a dictionary fixes the issue.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- Fixed a `PipelineTask` issue that would cause tasks to not be cancelled if
+  task was cancelled from outside of Pipecat.
+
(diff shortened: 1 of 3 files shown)


markbackman (approved):
LGTM

mattieruth (approved):
this looks good. i don't fully understand the cleanup_pipeline boolean and how to know it will be True in this case, but I trust you :)

aconchillo:
> this looks good. i don't fully understand the cleanup_pipeline boolean and how to know it will be True in this case, but I trust you :)
> 

The only case it will be true is if `asyncio.CancelledError` is triggered, otherwise the nominal case will run in which case we already cleanup and no need to cleanup again.

PR #1754, opened 2025-05-06
Removing the Canonical integration

adriancowham:
I think this about does it. Will check the docs to see if that needs to be scrubbed.


aconchillo:
@adriancowham Hi there! Can you describe why are we removing the integration?

markbackman:
There's a bit more to remove. I'll supersede with this one: https://github.com/pipecat-ai/pipecat/pull/1757.

Thanks and good luck!

adriancowham:
Thanks @markbackman, apologies for missing the some of the files.

PR #1766, opened 2025-05-07
fix multiple audio video sources

aconchillo:

This fixes `26c-gemini-multimodal-live-video.py`. You can ask what Gemini sees in your camera or in your screen and it should work (as long as you are sharing both your camera and your screen).


markbackman (approved):
Very cool! 👏

PR #1768, opened 2025-05-07
Add DebugLogObserver

markbackman:

The goal here is to add an observer class that makes it easy to filter on specific frames and optionally their source and destination. Basically, I got sick of writing a custom observer for debugging. This gives me more immediate access to seeing logs that I want by just specifying it directly in the observers arg in PipelineParams.

### DIFF
--- a/src/pipecat/observers/loggers/debug_log_observer.py
+++ b/src/pipecat/observers/loggers/debug_log_observer.py
@@ -1,0 +1,218 @@
+#
+# Copyright (c) 2024–2025, Daily
+#
+# SPDX-License-Identifier: BSD 2-Clause License
+#
+
+from dataclasses import fields, is_dataclass
+from enum import Enum, auto
+from typing import Dict, List, Optional, Set, Tuple, Type, Union
+
+from loguru import logger
+
+from pipecat.frames.frames import Frame
+from pipecat.observers.base_observer import BaseObserver, FramePushed
+from pipecat.processors.frame_processor import FrameDirection
+
+
+class FrameEndpoint(Enum):
+    """Specifies which endpoint (source or destination) to filter on."""
+
+    SOURCE = auto()
+    DESTINATION = auto()
+
+
+class DebugLogObserver(BaseObserver):
+    """Observer that logs frame activity with detailed content to the console.
+
+    Automatically extracts and formats data from any frame type, making it useful
+    for debugging pipeline behavior without needing frame-specific observers.
+
+    Args:
+        frame_types: Optional list of frame types to log, or a dict with frame type
+            filters. If None, logs all frame types.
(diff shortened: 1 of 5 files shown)


filipi87 (approved):
It will make things a lot easier to debug. 🚀

src/pipecat/observers/loggers/debug_log_observer.py
```diff
@@ -0,0 +1,218 @@
...
+    Examples:
+        Log all frames from all services:
+        ```python
+        observer = DebugLogObserver()
+        ```
+
+        Log specific frame types from any source/destination:
+        ```python
+        from pipecat.frames.frames import TranscriptionFrame, InterimTranscriptionFrame
+        observer = DebugLogObserver(frame_types=[TranscriptionFrame, InterimTranscriptionFrame])
```
aconchillo, line 45: Instead of a List, I've been using a Tuple instead, emulating the `isinstance()` approach:

```
        observer = DebugLogObserver(frame_types=(TranscriptionFrame, InterimTranscriptionFrame))
```

src/pipecat/observers/loggers/debug_log_observer.py
```diff
@@ -0,0 +1,218 @@
...
+
+        Log frames with specific source/destination filters:
+        ```python
+        from pipecat.frames.frames import StartInterruptionFrame, UserStartedSpeakingFrame, LLMTextFrame
+        from pipecat.transports.base_output_transport import BaseOutputTransport
+        from pipecat.services.stt_service import STTService
+
+        observer = DebugLogObserver(frame_types={
+            # Only log StartInterruptionFrame when source is BaseOutputTransport
+            StartInterruptionFrame: (BaseOutputTransport, FrameEndpoint.SOURCE),
```
aconchillo, line 56: Can you do:

```
            StartInterruptionFrame: BaseOutputTransport,
```

or would you need two entries one for SOURCE and one for DESTINATION?

aconchillo: Pretty sure we need two entries. I think that's fine.

aconchillo (approved):

aconchillo:
LGTM! Just added a comment to change a List for a Tuple.

src/pipecat/observers/loggers/debug_log_observer.py
markbackman: I intentionally kept it to one entry per item to keep it simple. I think that makes sense. You really need to know what you're doing with the source and destination. This is pretty much a tool for us to use.

src/pipecat/observers/loggers/debug_log_observer.py
markbackman: Good idea!

PR #1770, opened 2025-05-07
AWS Nova Sonic service - make interruption handling more reliable, in…

kompfner:
… terms of:

- not getting the conversation into a "stuck" state
- not losing assistant text that should've made it into the context


src/pipecat/services/aws_nova_sonic/aws.py
```diff
@@ -248,25 +249,45 @@ async def _handle_input_audio_frame(self, frame: InputAudioRawFrame):
...
+                # FINAL text blocks to know how many or which FINAL blocks to expect, but user
+                # interruptions throw a wrench in these schemes: depending on the exact timing of
+                # the interruption, we should or shouldn't expect some FINAL blocks.
+                if delay_to_catch_trailing_assistant_text:
+                    # This delay length is a balancing act between "catching" trailing assistant
+                    # text that is quite delayed but not waiting so long that user text comes in
+                    # first and results in a bit of context message order scrambling.
+                    await asyncio.sleep(1.25)
+                self._assistant_is_responding = False
+                await self._report_assistant_response_ended()
```
filipi87, line 282: Would it make sense if we receive a user text, we check if the `self. _assistant_is_responding` is `True` and if it is, invoke `await self._report_assistant_response_ended()` before doing anything else ?

kompfner: Yeah, and then after recording the user text immediately go back to reporting another assistant response started?

I think we have some more upcoming work to do around aggregating user text as well in the context...maybe we can tackle it then.

Hopefully AWS adds streamed assistant text before too long and all of this complexity can go away. But not sure where that is on the roadmap, or if it's on there.

PR #1780, opened 2025-05-08
GoogleLLMService: deprecate google-generativeai and use gemini-2.0-flash

aconchillo:


examples/foundational/32-gemini-grounding-metadata.py
```diff
@@ -33,7 +32,7 @@
 
 
 # Function handlers for the LLM
-search_tool = {"google_search_retrieval": {}}
+search_tool = {"google_search": {}}
```
filipi87, line 35: You will probably need to make the same changes inside the `news-chatbot` example. We are also using `google_search_retrieval` there.

aconchillo: Thanks! Updated!

examples/news-chatbot/server/news_bot.py
```diff
@@ -46,7 +46,7 @@
 # The threshold is a floating-point value in the range [0,1] and defaults to 0.3.
 # If the threshold value is 0, the response is always grounded with Google Search; if it's 1, it never is.
 search_tool = {
-    "google_search_retrieval": {
```
filipi87, line 49: It looks like the parameters that we were using here are not valid anymore. 
We are receiving a couple of errors if we try to execute:

```
2025-05-08 18:48:43.419 | ERROR    | pipecat.services.google.llm:_process_context:620 - GoogleLLMService#0 exception: 2 validation errors for GenerateContentConfig
tools.0.Tool.google_search.dynamic_retrieval_config
  Extra inputs are not permitted [type=extra_forbidden, input_value={'mode': 'MODE_DYNAMIC', 'dynamic_threshold': 0}, input_type=dict]
    For further information visit https://errors.pydantic.dev/2.10/v/extra_forbidden
tools.0.callable
  Input should be callable [type=callable_type, input_value={'google_search': {'dynam...dynamic_threshold': 0}}}, input_type=dict]
    For further information visit https://errors.pydantic.dev/2.10/v/callable_type
Traceback (most recent call last):

```

filipi87: https://ai.google.dev/gemini-api/docs/grounding?lang=python#google-search-retrieval

_Note: Google Search retrieval is only compatible with Gemini 1.5 models. For Gemini 2.0 models, you should use Search as a tool._

aconchillo: I'll update it. Those only work for gemini 1.5. I need to check for gemini 2.0

aconchillo: Actually the example forces geminie 1.5. So maybe we just leave news-bot as it was. It should keep working.

filipi87: Yeah, becaue looking at the docs, I don't think it works with any other model.

filipi87: I just tested it, and if you switch back to the previous model, you should continue using google_search_retrieval instead. So basically, you should roll back any changes in this example, and everything should work fine.

filipi87 (approved):
LGTM. We just need to roll back the changes from the `news_bot`.

PR #1784, opened 2025-05-09
Handle audio transcript from gpt-4o-audio and clean up logs

thsunkid:
- Handle audio transcript from gpt-4o-audio / gpt-4o-mini-audio if they are used for LLM or STT+LLM in the pipeline.
- ~~Clean up the `context.get_messages_for_logging()` message log. Right now, the log shows up with double square brackets, which is confusing when debugging.~~

### DIFF
--- a/src/pipecat/services/anthropic/llm.py
+++ b/src/pipecat/services/anthropic/llm.py
@@ -166,7 +166,7 @@ class AnthropicLLMService(LLMService):
             await self.start_processing_metrics()
 
             logger.debug(
-                f"{self}: Generating chat [{context.system}] | [{context.get_messages_for_logging()}]"
+                f"{self}: Generating chat [{context.system}] | {context.get_messages_for_logging()}"
             )
 
             messages = context.messages
--- a/src/pipecat/services/anthropic/llm.py
+++ b/src/pipecat/services/anthropic/llm.py
@@ -166,7 +166,7 @@ class AnthropicLLMService(LLMService):
             await self.start_processing_metrics()
 
             logger.debug(
-                f"{self}: Generating chat [{context.system}] | {context.get_messages_for_logging()}"
+                f"{self}: Generating chat [{context.system}] | [{context.get_messages_for_logging()}]"
             )
 
             messages = context.messages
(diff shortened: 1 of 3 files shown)


markbackman:
@thsunkid I don't see this in the streaming response docs:
https://platform.openai.com/docs/api-reference/chat-streaming

Are you thinking of the Realtime API? That's a separate LLM service in Pipecat:
https://github.com/pipecat-ai/pipecat/blob/main/src/pipecat/services/openai_realtime_beta/openai.py

thsunkid:
@markbackman Thanks for the quick response. 
It's under this docs https://platform.openai.com/docs/guides/audio?api-mode=chat#add-audio-to-your-existing-application
https://platform.openai.com/docs/models/gpt-4o-audio-preview

This model is the pre- Realtime API that uses Chat Completions API.

markbackman:
> @markbackman Thanks for the quick response. It's under this docs https://platform.openai.com/docs/guides/audio?api-mode=chat#add-audio-to-your-existing-application https://platform.openai.com/docs/models/gpt-4o-audio-preview
> 
> This model is the pre- Realtime API that uses Chat Completions API.

Thanks for sharing. I wasn't aware of this. Though, looking through the Pipecat implementation, the openai.base_llm doesn't support multimodal input yet. Are you using the `extra` field to specify the modality and audio params? If we're going to add support for this, we should treat audio input as a first class modality.

Also, can you please revert the logging changes that you've proposed? The square brackets help to detect spaces in the output, so they actually serve a purpose.

thsunkid:
> Are you using the extra field to specify the modality and audio params?

Yes, that's what I did. I made a `MultiModalOpenAIUserContextAggregator` to aggregate the audio context, and created a `MultiModalOpenAILLMService` class which is almost the same as `OpenAILLMService` except it uses the `MultiModalOpenAIUserContextAggregator`. Modality is passed to `MultiModalOpenAILLMService` via the `args` parameter. 

I'm not sure if this is too hacky and we want to overload with the `openai.base_llm` with these logic or we actually want to keep them separate as this.

Thanks for letting me know, and sorry for the piggyback change. I’ll revert (though sometimes this displays the messages in ways that make me doubt there is a bug how in the message is aggregated...).

markbackman (approved):
Thanks for the explanation. I think this change makes sense. I'm not sure if we would add support for audio, but this at least allows you to do that if you use `extra`, which would otherwise be difficult.

Thanks for the submission 🙇

PR #1793, opened 2025-05-11
Fix Deepgram TTS streaming

kwindla:
Change `DeepgramTTSService` to use the `stream_raw()` function in the Deepgram Python library, so that we can stream the audio bytes.


aconchillo (approved):

aconchillo:
LGTM!

PR #1799, opened 2025-05-12
Using audio source for capturing Daily's participant audio

filipi87:
Using audio source for capturing Daily's participant audio


markbackman (approved):
LGTM

PR #1800, opened 2025-05-12
introduce frame processors setup

aconchillo:

This PR introduces `FrameProcessor.setup()`. This allows setting up processors before `StartFrame` is being pushed. One of the problems we had is that we relied on `StartFrame` to initialize things, however you might want to start some tasks even before receiving `StartFrame` (e.g. to even handle `StartFrame`). So, `FrameProcessor.setup()` allows you to do that. There are things that I have moved from `StartFrame` to `FrameProcessorSetup`, for example, the clock, the system manager and the observer. Having the task manager allows you to create Pipecat controlled task even before receiving the `StartFrame`.

Also, we have all seen the annoying message:

```
XXXX not properly initialized, missing super().process_frame(frame, direction)
```

It doesn't give much information and we don't really know what could cause it. In this PR, we change this for:

```
Trying to process {frame} but StartFrame not received yet
```

This will give you a hint that your application is receiving frames before the `StartFrame`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- A new function `FrameProcessor.setup()` has been added to allow setting up
+  frame processors before receiving a `StartFrame`. This is what's happening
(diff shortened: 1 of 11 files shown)


filipi87 (approved):
Makes sense. Nice improvement.

markbackman (approved):
LGTM

PR #1608, opened 2025-04-17
simple text chatbot example

vipyne:
A simplified version of Simple Chatbot with text-only input and output, leveraging RTVI `action_llm_append_to_messages`.  Still uses Daily Transport (for now).

textbot

PR #1804, opened 2025-05-12
add text chatbot example using small webrtc transport

vipyne:

Text only chat bot example using small webrtc transport (_not_ Daily) and RTVI.  

~TODO: add `text_enabled` to transport params to guide UI on whether or not to display chat input.~
Used a `RTVIServerMessageFrame` to update the UI.

Requires: https://github.com/pipecat-ai/small-webrtc-prebuilt/pull/9

Test with pipecat example [here](https://github.com/pipecat-ai/pipecat/pull/1804).

### DIFF
--- a/examples/foundational/41-text-only-webrtc.py
+++ b/examples/foundational/41-text-only-webrtc.py
@@ -1,0 +1,167 @@
+#
+# Copyright (c) 2024–2025, Daily
+#
+# SPDX-License-Identifier: BSD 2-Clause License
+#
+
+import argparse
+import asyncio
+import io
+import os
+import re
+import shutil
+import sys
+
+import aiohttp
+from dotenv import load_dotenv
+from loguru import logger
+from PIL import Image
+
+from pipecat.audio.vad.silero import SileroVADAnalyzer
+from pipecat.frames.frames import (
+    Frame,
+    FunctionCallResultFrame,
+    LLMMessagesAppendFrame,
+    URLImageRawFrame,
+)
+from pipecat.pipeline.pipeline import Pipeline
+from pipecat.pipeline.runner import PipelineRunner
+from pipecat.pipeline.task import PipelineParams, PipelineTask
+from pipecat.processors.aggregators.openai_llm_context import OpenAILLMContext
+from pipecat.processors.frame_processor import FrameDirection, FrameProcessor
+from pipecat.processors.frameworks.rtvi import (
+    ActionResult,
+    RTVIAction,
+    RTVIActionArgument,
+    RTVIConfig,
+    RTVIObserver,
+    RTVIProcessor,
+    RTVIServerMessageFrame,
+)
+from pipecat.services.openai.llm import OpenAILLMService
+from pipecat.transports.base_transport import TransportParams
+from pipecat.transports.network.small_webrtc import SmallWebRTCTransport
+from pipecat.transports.network.webrtc_connection import SmallWebRTCConnection
+
+load_dotenv(override=True)
+
+
+async def run_bot(webrtc_connection: SmallWebRTCConnection, _: argparse.Namespace):
+    logger.info(f"Starting bot")
+
+    transport = SmallWebRTCTransport(
+        webrtc_connection=webrtc_connection,
+        params=TransportParams(),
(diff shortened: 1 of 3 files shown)


examples/foundational/41-text-only-webrtc.py
```diff
@@ -0,0 +1,165 @@
...
+                enable_metrics=True,
+            ),
+            observers=[RTVIObserver(rtvi)],
+        )
+
+        @rtvi.event_handler("on_client_ready")
+        async def on_client_ready(rtvi):
+            logger.info("Pipecat client ready.")
+            await rtvi.set_bot_ready()
+            # Update small webrtc UI to just handle text
```
mattieruth, line 133: what is small webrtc UI? Is this a Pipecat standard? Or are these custom messages we have introduced as part of this example to pair with an example client? Where is that client?

mattieruth (approved):
This looks good. A couple comments:

1. Can we add an overview and more inline documentation? I wish our examples did a better job of explaining beyond just the code and the name of the example. What are we trying to show how to do here?

2. TBH WebRTC seems overkill for a text-only communication. It seems like if I were building a text only pipeline, I would be happy with a simple websocket connection. I know that for the ease of building this example, that didn't make sense since we don't have an rtvi client-side websocket transport, but i guess it gets back to the purpose of this example. If it's to show text-only, then we DO have python websocket transports we could use without RTVI. If RTVI is important, then we can stick with the SmallWebRTC approach but should probably consider an RTVI-capable client-side websocket transport.

examples/foundational/41-text-only-webrtc.py
mattieruth: thanks. while i did want the answers to that question for myself, my bigger point was that we need documentation here to answer those questions for anyone coming to this example. the existing comment is not enough.

vipyne:
> 2. TBH WebRTC seems overkill for a text-only communication. It seems like if I were building a text only pipeline, I would be happy with a simple websocket connection. I know that for the ease of building this example, that didn't make sense since we don't have an rtvi client-side websocket transport, but i guess it gets back to the purpose of this example. If it's to show text-only, then we DO have python websocket transports we could use without RTVI. If RTVI is important, then we can stick with the SmallWebRTC approach but should probably consider an RTVI-capable client-side websocket transport.

I think an RTVI-capable client-side websocket transport sounds like a worthy pursuit. I'm hesitant to take that on right now with impending RTVI changes, but it makes sense to me that eventually there should be a plain jane websocket transport in [this list](https://github.com/pipecat-ai/pipecat-client-web-transports/tree/main/transports).

I think the aim here is to show text usage with the lowest lift from existing client-side libraries.  My assumption is folks using pipecat are using audio input in addition to text, which makes webrtc make sense as a transport.  This example aims to be simple by only showing text.  I will add another example that shows text and audio (and maybe even video).  

re:
> python websocket transports we could use without RTVI.
There is [this websocket example](https://github.com/pipecat-ai/pipecat/tree/main/examples/websocket-server) I considered as a base for a text chat bot example.  Maybe that should still be on the TODO list.  But we don't have many other examples using protobufs and I hesitate to build on that (at least not without putting more thought into it).


I agree this could use much more commentary. I'll add that as well.

vipyne:
There is also [this text chatbot example](https://github.com/pipecat-ai/pipecat/pull/1608) that uses daily transport.  it's pretty heavy.  linked here for posterity, but I imagine it'll be moot soon.

filipi87 (commented):
Looks great. I have added one note about the text-voice example, that I guess it would be nice to add.

And before we merge this, we need to first release the new version of `pipecat-ai-small-webrtc-prebuilt` with support for text. 🙂

vipyne:
> @vipyne, In the text-and-voice example: if I speak the user's part, the spoken input is not shown in the conversation. Only typed input appears.
> I think it would be helpful to display spoken user input as well, so we can clearly distinguish between the user's and the bot's turns.

@filipi87 https://github.com/pipecat-ai/small-webrtc-prebuilt/pull/13  🎸

filipi87 (approved):
LGTM! 🚀

PR #1811, opened 2025-05-14
Fix: Resolve an issue where tracing imports were required

markbackman:

Fixing an issue I introduced by merging the OTel tracing feature.

Clarifying about the change since it came up in comments:
```
The fix to the runtime error is using string literals for the OTel types.

Adding TYPE_CHECKING helps IDEs better understand what the string literal types correspond to.
```

### DIFF
--- a/src/pipecat/utils/tracing/conversation_context_provider.py
+++ b/src/pipecat/utils/tracing/conversation_context_provider.py
@@ -5,7 +5,12 @@
 #
 
 import uuid
-from typing import Optional
+from typing import TYPE_CHECKING, Optional
+
+# Import types for type checking only
+if TYPE_CHECKING:
+    from opentelemetry.context import Context
+    from opentelemetry.trace import SpanContext
 
 from pipecat.utils.tracing.setup import is_tracing_available
 
@@ -22,7 +27,7 @@ class ConversationContextProvider:
     """
 
     _instance = None
-    _current_conversation_context: Optional[Context] = None
+    _current_conversation_context: Optional["Context"] = None
     _conversation_id: Optional[str] = None
 
     @classmethod
@@ -33,7 +38,7 @@ class ConversationContextProvider:
         return cls._instance
 
     def set_current_conversation_context(
-        self, span_context: Optional[SpanContext], conversation_id: Optional[str] = None
+        self, span_context: Optional["SpanContext"], conversation_id: Optional[str] = None
     ):
         """Set the current conversation context.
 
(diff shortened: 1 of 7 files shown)


test-requirements.txt
```diff
@@ -1 +1 @@
--e ".[anthropic,aws,google,langchain,tracing]"
+-e ".[anthropic,aws,google,langchain]"
```
markbackman, line 1: The clue was right here! I clearly lost this in the fog of building the feature.

src/pipecat/utils/tracing/conversation_context_provider.py
```diff
@@ -5,7 +5,12 @@
 #
 
 import uuid
-from typing import Optional
+from typing import TYPE_CHECKING, Optional
+
+# Import types for type checking only
```
kompfner, line 10: 👍

src/pipecat/utils/tracing/service_decorators.py
```diff
@@ -16,19 +16,26 @@
...
+    from opentelemetry import context as context_api
+    from opentelemetry import trace
 
+# Conditional imports for runtime
 from pipecat.utils.tracing.service_attributes import (
     add_llm_span_attributes,
     add_stt_span_attributes,
     add_tts_span_attributes,
 )
-from pipecat.utils.tracing.setup import OPENTELEMETRY_AVAILABLE, is_tracing_available
```
kompfner, line 29: looks like we also import `OPENTELEMETRY_AVAILABLE` in `class_decorators.py`—should we ake the same change there?

test-requirements.txt
kompfner: 🕵

src/pipecat/utils/tracing/service_decorators.py
markbackman: Yes, good catch.

kompfner (approved):

src/pipecat/utils/tracing/conversation_context_provider.py
markbackman: The fix to the runtime error is using string literals for the types.

Adding TYPE_CHECKING helps IDEs better understand what the string literal types correspond to.

kompfner: Gotcha! So like the VS Code Python extension is doing this static type checking. Got it! It must somehow gracefully handle opentelemetry not being available, then...

PR #1819, opened 2025-05-15
LmntTTSService: add model param and additional languages

kaikato:

Adds a model parameter to LMNT TTS constructor to allow switching models. LMNT offers two models, the stable `aurora` model and experimental `blizzard` model, and prior to this change there is no way to use `blizzard`.

Also adds additional languages now supported by LMNT on the `blizzard` model.


src/pipecat/services/lmnt/tts.py
```diff
@@ -135,6 +149,7 @@ async def _connect_websocket(self):
                 "format": self._settings["format"],
                 "sample_rate": self.sample_rate,
                 "language": self._settings["language"],
+                "model": self._model_name,
```
markbackman, line 152: This should be `model_name`, which is set in the AIServices base class.
```suggestion
                "model": self.model_name,
```

kaikato: done, thank you!

markbackman (approved):
LGTM!

markbackman:
Oh, one final thing. Can you also add a CHANGELOG entry? Maybe something like:
```
### Added

- Added additional languages to `LmntTTSService`. Languages include: <insert languages>
```

Once that's added, I can merge this down.

kaikato:
> Oh, one final thing. Can you also add a CHANGELOG entry? Maybe something like:
> 
> ```
> ### Added
> 
> - Added additional languages to `LmntTTSService`. Languages include: <insert languages>
> ```
> 
> Once that's added, I can merge this down.

Done! Hopefully using separate entries is fine, although I could combine if that's preferred.

PR #1826, opened 2025-05-16
Enhance truncate_content function to handle missing model encodings and update model usage to utilize OPENAI_MODEL environment variable. Revise task instructions for clarity on user engagement during explanations.

sanjeed5:

### DIFF
--- a/examples/studypal/env.example
+++ b/examples/studypal/env.example
@@ -3,3 +3,4 @@ DAILY_API_KEY= # Create here: https://dashboard.daily.co/developers
 OPENAI_API_KEY= # Create here: https://platform.openai.com/docs/overview
 CARTESIA_API_KEY= # Create here: https://play.cartesia.ai/console
 CARTESIA_VOICE_ID= # Find here: https://play.cartesia.ai/  
+OPENAI_MODEL=gpt-4.1-mini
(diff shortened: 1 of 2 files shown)


examples/studypal/env.example
```diff
@@ -2,4 +2,5 @@ DAILY_SAMPLE_ROOM_URL= # Follow instructions here and put your https://YOURDOMAI
 DAILY_API_KEY= # Create here: https://dashboard.daily.co/developers 
 OPENAI_API_KEY= # Create here: https://platform.openai.com/docs/overview
 CARTESIA_API_KEY= # Create here: https://play.cartesia.ai/console
-CARTESIA_VOICE_ID= # Find here: https://play.cartesia.ai/  
\ No newline at end of file
+CARTESIA_VOICE_ID= # Find here: https://play.cartesia.ai/  
+OPENAI_MODEL=gpt-4.1-mini
```
markbackman, line 6: Instead of adding a model env var, I think we should just remove it all together and use the latest default model from the OpenAILLMService. This will make maintenance easier.

markbackman:
Thanks for the submission! This generally looks good. If you can remove the LLM model env var and arg all together, I think this is good to go.

markbackman:
@sanjeed5 pinging about my previous comment so we can get this merged.

aconchillo:
Closing. The `studypal` example has been moved here: https://github.com/pipecat-ai/pipecat-examples/tree/main/studypal. Please, submit a PR there if this is still relevant. :pray:

PR #1828, opened 2025-05-17
DailyTransport: replace virtual speaker and microphones

aconchillo:

This requires a new version of daily-python which hasn't been released yet.


CHANGELOG.md
```diff
@@ -59,6 +59,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- `DailyTransport` now captures audio from individual participants instead of
+  the whole room. This allows identifying audio frames per participant.
```
filipi87, line 63: 🎉

src/pipecat/transports/services/daily.py
```diff
@@ -705,14 +668,14 @@ async def capture_participant_audio(
...
-            f"Starting to capture audio from participant {participant_id} to {audio_source}"
-        )
+        logger.info(f"Starting to capture [{audio_source}] audio from participant {participant_id}")
 
         self._client.set_audio_renderer(
             participant_id,
             self._audio_data_received,
             audio_source=audio_source,
+            sample_rate=sample_rate,
+            callback_interval_ms=callback_interval_ms,
```
filipi87, line 678: Cool. Just waiting for the new version of `daily-python` so I can test it. 🙂

src/pipecat/transports/services/daily.py
```diff
@@ -1034,32 +1005,26 @@ async def capture_participant_audio(
...
     ):
-        await self._client.capture_participant_audio(
-            participant_id, self._on_participant_audio_data, audio_source
-        )
+        if self._streaming_started:
+            await self._client.capture_participant_audio(
+                participant_id, self._on_participant_audio_data, audio_source, sample_rate
+            )
+        else:
+            self._capture_participant_audio.append((participant_id, audio_source, sample_rate))
```
filipi87, line 1015: Shouldn't we also call `await self._client.capture_participant_audio` here? Or is there a reason it's not needed?

I'm still missing something about how this would work if a third participant joins.

aconchillo: If "stream on start" is not enabled, we don't want to capture the audio, so we just stop the list of participants that keep joining until `start_streaming()` is called.

filipi87: My question is more like this: if `audio_in_stream_on_start` is `enabled` and we invoke `start_audio_in_streaming`, we begin capturing audio from the participants currently in the meeting.

But if another participant joins later and we invoke `await self._client.capture_participant_audio`, will their audio also be captured ?

aconchillo: That's what `if self._streaming_started` does. If you have already called `start_audio_in_streaming()` it will set `_streaming_started` to True, so you capture right away.

filipi87 (approved):
Looks good. To be honest, I'm not sure what I was confused about yesterday. It all makes perfect sense now. 😂

aconchillo:
> Looks great!
> 
> Once daily-python 0.19.0 is published, I'll test.
> 
> Future question: will we also move camera to the custom video track?

Yes, but cameras don't have the process restriction since they don't depend on `libwebrtc`'s ADM (AudioDeviceModule). I will add custom video tracks in the future and specifying a track instead of a "device" would also be possible.

PR #1829, opened 2025-05-18
PipelineTask: add add_observer()

aconchillo:

This allows adding observers after task creation. This can be useful in cases where the task is passed around after being created and you still need would like to add more observers.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- Added `PipelineTask.add_observer()` and `PipelineTask.remove_observer()` to
+  allow mangaging observers at runtime. This is useful for cases where the task
+  is passed around to other code components that might want to observe the
+  pipeline dynamically.
+
 - Added `user_id` field to `TranscriptionMessage`. This allows identifying the
   user in a multi-user scenario. Note that this requires that
   `TranscriptionFrame` has the `user_id` properly set.
--- a/tests/test_pipeline.py
+++ b/tests/test_pipeline.py
@@ -8,13 +8,8 @@ import asyncio
 import time
 import unittest
 
-from pipecat.frames.frames import (
-    EndFrame,
-    HeartbeatFrame,
-    StartFrame,
-    StopFrame,
-    TextFrame,
-)
+from pipecat.frames.frames import EndFrame, HeartbeatFrame, StartFrame, StopFrame, TextFrame
+from pipecat.observers.base_observer import BaseObserver, FramePushed
 from pipecat.pipeline.parallel_pipeline import ParallelPipeline
 from pipecat.pipeline.pipeline import Pipeline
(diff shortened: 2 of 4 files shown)


filipi87 (commented):
The code looks good, but it seems like the tests are failing.
Are the test failures related to the changes in this PR ?

CHANGELOG.md
```diff
@@ -9,6 +9,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- Added `PipelineTask.add_observer()` to allow adding observers after task
```
kompfner, line 12: Maybe add a sentence like `Useful for cases where the task is passed around to other code components that need to add their own observers`.

tests/test_pipeline.py
```diff
@@ -95,6 +96,54 @@ async def test_task_single(self):
...
+                    frame_add_received = True
+
+        identity = IdentityFilter()
+        pipeline = Pipeline([identity])
+        task = PipelineTask(pipeline, observers=[CustomObserver()])
+        task.set_event_loop(asyncio.get_event_loop())
+        task.add_observer(CustomAddObserver())
+
+        await task.queue_frames([TextFrame(text="Hello Downstream!"), EndFrame()])
+        await task.run()
```
kompfner, line 143: Is it worth adding a unit test that adds an observer _after_ the pipeline is up and running? I can imagine needing to dynamically add an observer some time after the conversation starts.

kompfner: In fact...thinking about how I could've implemented the "deferred" `post_actions` execution in Pipecat Flows (where the `post_actions` only occur after the bot stops speaking, which in turn might be waiting on the user to speak first) using `add_observer()`:

* As soon as we set a node that requires deferred `post_actions`, we use `add_observer()` to detect when the bot stops speaking
* And then as soon as we handle the bot stopping speaking, we remove the observer...

...so should we also have a corresponding `remove_observer()`?

aconchillo:
> The code looks good, but it seems like the tests are failing. Are the test failures related to the changes in this PR ?

I don't think so. It's a weird issue. I need to track it down.

tests/test_pipeline.py
aconchillo: > Is it worth adding a unit test that adds an observer _after_ the pipeline is up and running? I can imagine needing to dynamically add an observer some time after the conversation starts.

Yep, I can do that.

aconchillo: I just added `remove_observer()` and also added a test for delayed `add_observer()` and also `remove_observer()`.

aconchillo:
> > The code looks good, but it seems like the tests are failing. Are the test failures related to the changes in this PR ?
> 
> I don't think so. It's a weird issue. I need to track it down.

Fixed it! `__init__` arguments can't have default values when you use lists or dictionaries, because those will be shared with all instances of the class. You always have to do:

```
argument: Optional[SomeType] = None
```

and then

```
self._argument = argument or SomeType()
```

kompfner (approved):

PR #1830, opened 2025-05-18
PipelineTask: add new started/stopped/ended/cancelled events

aconchillo:

New pipeline task events to know when a pipeline is started, stopped, ended or cancelled. They will all be triggered when the corresponding frames reach the end of the pipeline.

### DIFF
--- a/src/pipecat/pipeline/task.py
+++ b/src/pipecat/pipeline/task.py
@@ -144,7 +144,26 @@ class PipelineTask(BaseTask):
     `LLMFullResponseEndFrame` are received within `idle_timeout_secs`.
 
        @task.event_handler("on_idle_timeout")
-       async def on_idle_timeout(task):
+       async def on_pipeline_idle_timeout(task):
+           ...
+
+    There are also events to know if a pipeline has been started, stopped, ended
+    or cancelled.
+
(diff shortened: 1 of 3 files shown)


src/pipecat/pipeline/task.py
```diff
@@ -552,8 +575,16 @@ async def _process_down_queue(self):
             if isinstance(frame, self._reached_downstream_types):
                 await self._call_event_handler("on_frame_reached_downstream", frame)
 
-            if isinstance(frame, (EndFrame, StopFrame)):
+            if isinstance(frame, StartFrame):
+                await self._call_event_handler("on_pipeline_started", frame)
```
kwindla, line 579: If I'm reading the code correctly, this event is called after the StartFrame has passed all the way through the pipeline. So when you get this event, all the processors in the pipeline have already seen the StartFrame. Is that right?

aconchillo: Exactly. We have two hidden processors at the beginning and end of a pipeçine that allow us to do this.

aconchillo:
> Cool. Would it also be worth to add an example using it ?

I don't think that's really necessary. These are just events. We have plenty of examples on how to register for events and we have documentation for it.

PR #1831, opened 2025-05-18
Cartesia speed option for sonic-2

golbin:
Added support for speed control in `CartesiaTTSService`.  The latest sonic-2 model now supports "slow", "normal", and "fast" speed options.

Reference: https://docs.cartesia.ai/2024-11-13/api-reference/tts/sse#request.body.speed

Note that the existing experimental_controls speed setting is maintained for backward compatibility but may be deprecated in future versions.


golbin:
This PR includes some overlapping changes with #1672. Please merge that one first if possible, or merge only this if that's more convenient.

golbin:
@markbackman I checked with the Cartesia team. The speed option was removed at one point and later reintroduced, but it's not the same as the original experimental control. Please refer to the documentation for clarification.

golbin:
The new speed option was added just 5 days ago.

golbin:
We may also need to change the API version.

`cartesia_version: str = "2024-06-10",`

markbackman:
I got clarification from the Cartesia team that there's a new top-level speed parameter that is also experimental, but is replacing the old _experimental_controls param.

This required updating to the cartesia 2.x package, which was different enough. Given that, I have a PR for the change here: https://github.com/pipecat-ai/pipecat/pull/1839.

golbin:
Thank you, @markbackman!

Just one suggestion: it might be helpful to mention in the CHANGELOG.md that the available speed options are now only "slow", "normal", and "fast".
This is because the previous version allowed numeric values and other string-based speed inputs.

markbackman:
> Thank you, @markbackman!
> 
> Just one suggestion: it might be helpful to mention in the CHANGELOG.md that the available speed options are now only "slow", "normal", and "fast". This is because the previous version allowed numeric values and other string-based speed inputs.

Good idea!

PR #1835, opened 2025-05-18
Add examples/open-telemetry-tracing-langfuse

marcklingen:
Thanks to the opentelemetry tracing support added by @markbackman in #1729, Langfuse (repo) can now be used to trace and evaluate applications created with pipecat.

While adding documentation on Langfuse regarding the use of Langfuse tracing for Pipecat via OTel, I realized it is trickier than expected. Langfuse only accepts OTel traces through HTTP/protobuf, not via gRPC. Therefore, I have created this example as a fork of the original one by Mark to help others get started quickly.

### DIFF
--- a/examples/open-telemetry-tracing-langfuse/README.md
+++ b/examples/open-telemetry-tracing-langfuse/README.md
@@ -1,0 +1,138 @@
+# Langfuse Tracing for Pipecat via OpenTelemetry
+
+This demo showcases [Langfuse](https://langfuse.com) tracing integration for Pipecat services via OpenTelemetry, allowing you to visualize service calls, performance metrics, and dependencies.
+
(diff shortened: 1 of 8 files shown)


markbackman (approved):
LGTM!

I may end up organizing things a little differently. I'm thinking about a top-level directory called `open-telemetry-tracing` that would contain multiple directories, so `langfuse`, `jaeger`, etc with a README, requirements.txt, and bot.py file.

But, I don't think that needs to stop anything here. this is great!

marcklingen:
> LGTM!
> 
> I may end up organizing things a little differently. I'm thinking about a top-level directory called `open-telemetry-tracing` that would contain multiple directories, so `langfuse`, `jaeger`, etc with a README, requirements.txt, and bot.py file.
> 
> But, I don't think that needs to stop anything here. this is great!

sounds like a great plan, thanks for the speedy review and getting this shipped!

PR #1841, opened 2025-05-19
make BaseTextAggregator and BaseTextFilter functions async

aconchillo:

This allows users to be able to use async functions they might need (e.g. callbacks). For example, see how the switching voices examples gets simplified.


markbackman (approved):
LGTM! I tested a few examples too.

Issue #1812, opened 2025-05-14
Streaming implementation for Google TTS (Chirp3 HD)


manishkjs1:
### Problem Statement

I am part of GCP sales engineering team and wanted to request integration of streaming module in TTS (Chirp3HD) code.

Right now latency in batch mode is way high for any reasonable voice to voice conversation.

### Proposed Solution

The exact module is `StreamingSynthesizeRequest` and API details are [here](https://cloud.google.com/text-to-speech/docs/reference/rpc/google.cloud.texttospeech.v1#streamingsynthesizerequest).

The TTS latency in streaming mode (~200ms) is way better than in batch mode (~800ms).

### Additional Context

I have had got done similar update in Livekit and it works there. I would love to create a PR for this and contribute.

### Would you be willing to help implement this feature?

- [x] Yes, I'd like to contribute
- [ ] No, I'm just suggesting

manishkjs1:
@markbackman @aconchillo - FYR please

aristid:
Hello, I wanted to try this out and implemented the change.
- I noticed slight quality degradation (everything else unchanged). Do you have an idea what could cause this?
- I think it works for all voices, at least Journey and Neural2 seem to work. Shouldn't it be the case?
Thank you!

aristid:
Ach, I see, StreamingSynthesisInput doesn't support SSML, that's why.

markbackman:
@aristid one option is to keep a streaming version and non-streaming version. We have other examples of this. For example `CartesiaTTSService` is the websocket API service and `CartesiaHttpTTSService` is the HTTP API service. We could follow a similar pattern for `GoogleTTSService`. That would be `GoogleTTSService` for streaming and `GoogleHttpTTSService` for non-streaming (or some other concise but descriptive name).

aristid:
Should I make a pull request? GoogleStreamingTTSService, to keep the http version working? It really is much faster, making using standard Gemini feasible instead of Gemini Live. The quality is a tick worse though, hope @manishkjs1 can say something about that.

markbackman:
> should I make a pull request?

Yes, please! 🙌 

> GoogleStreamingTTSService, to keep the http version working? It really is much faster, making using standard Gemini feasible instead of Gemini Live. The quality is a bit worse though, hope [@manishkjs1](https://github.com/manishkjs1) can say something about that.

Our convention has been:
- `f"{service_name}TTSService"` for the primary implementation, e.g. `GoogleTTSService`
[...]

PR #1843, opened 2025-05-19
Streaming implementation for Google TTS

aristid:
I went on and implemented [Streaming implementation for Google TTS ](https://github.com/pipecat-ai/pipecat/issues/1812)
It really makes a huge speed difference. However, the quality seems slightly degraded, as if the output is compressed. I couldn't find the reason for this. Sample rate is unchanged in my pipeline, 24000.
Additionally and unrelated, I had to change rate param to sample_rate as per https://cloud.google.com/text-to-speech/docs/reference/rpc/google.cloud.texttospeech.v1#google.cloud.texttospeech.v1.StreamingAudioConfig

Sorry if I missed something.

### DIFF
--- a/src/pipecat/services/google/tts.py
+++ b/src/pipecat/services/google/tts.py
@@ -8,8 +8,6 @@ import asyncio
 import json
 import os
 
-from pipecat.utils.tracing.service_decorators import traced_tts
-
 # Suppress gRPC fork warnings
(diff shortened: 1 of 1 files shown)


aristid:
Sorry, something went wrong uploading the changed file. Will fix and reopen.

PR #1844, opened 2025-05-19
add docs link at top of readme

vipyne:


markbackman:
Rather than just link to the docs, we should provide more targeted information. Maybe something like:

**Pipecat** is an open-source Python framework for building real-time voice and multimodal conversational agents. Orchestrate audio and video, AI services, different transports, and conversation pipelines effortlessly—so you can focus on what makes your agent unique.

> Want to dive right in? [Install Pipecat](https://docs.pipecat.ai/getting-started/installation) then try the [quickstart](https://docs.pipecat.ai/getting-started/quickstart).

PR #1847, opened 2025-05-19
Keep span identifiers in attributes only

markbackman:

With this change, the span name will be either `conversation` for the top-level span and `turn` for the child span. Then, each of those spans will have an attribute:
- `conversation.id` for the conversation ID
- `turn.number` for the turn number


sunahsuh (approved):
LGTM!

PR #1850, opened 2025-05-19
TranscriptionMessage: add user_id field

aconchillo:

This allows us to identify which user said what in a multi-user scenario.


markbackman (approved):
LGTM

PR #1854, opened 2025-05-20
fix(elevenlabs tts): message parameter naming

sklinglernv:
Changes:

- Fix an an issue with parameter naming when parsing Elevenlab responses
- Fix authorization issue when flushing audio
- Use contexts more actively to send out TTSStopedFrames more punctually

### DIFF
--- a/src/pipecat/services/elevenlabs/tts.py
+++ b/src/pipecat/services/elevenlabs/tts.py
@@ -276,8 +276,16 @@ class ElevenLabsTTSService(AudioContextWordTTSService):
 
     async def flush_audio(self):
         if self._websocket and self._context_id:
-            msg = {"context_id": self._context_id, "flush": True}
+            self._context_id_to_close = self._context_id
+            msg = {"context_id": self._context_id, "flush": True, "xi_api_key": self._api_key}
             await self._websocket.send(json.dumps(msg))
+            msg = {
+                "context_id": self._context_id,
+                "close_context": True,
+                "xi_api_key": self._api_key,
+            }
+            await self._websocket.send(json.dumps(msg))
+            self._context_id = None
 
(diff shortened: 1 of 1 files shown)


src/pipecat/services/elevenlabs/tts.py
```diff
@@ -406,12 +414,17 @@ async def _receive_messages(self):
...
                 # Context has finished
-                if self._context_id == received_ctx_id:
+                if (
+                    self._context_id == received_ctx_id
+                    or self._context_id_to_close == received_ctx_id
+                ):
                     self._context_id = None
+                    self._context_id_to_close = None
                     self._started = False
+                    await self.push_frame(TTSStoppedFrame())
```
markbackman, line 427: We're relying on the parent class to push TTSStoppedFrames for Websocket based TTS services. If you're looking to respond based on when a bot if finished speaking, you should use the BotStoppedSpeakingFrame and not the TTSStoppedFrame.

Can you please remove this from the PR?

src/pipecat/services/elevenlabs/tts.py
```diff
@@ -386,7 +394,7 @@ async def _receive_messages(self):
             msg = json.loads(message)
             # Check if this message belongs to the current context
             # The default context may return null/None for context_id
-            received_ctx_id = msg.get("context_id")
+            received_ctx_id = msg.get("contextId")
```
markbackman, line 397: Nice catch on this. The docs for the v1 multi endpoint I received had errors, indicating that these were snake_case. Thanks for catching this and isFinal.

sklinglernv:
PR reduced to just correct the variable names.

AngeloGiacco:
Apologies for this one guys! have updated the preview docs, for now due to some intricacies with dataclasses it is context_id on input and contextId on output. i will fix so that we standardise the docs in a backwards compatible manner

PR #1857, opened 2025-05-20
avoid mutable default constructor values

aconchillo:

This is an important bug fix that could result in unexpected behavior if multiple instance of the same class were used. In Python default arguments are evaluated only once when the function is defined. This means that if you have a class with a list argument that defaults to `[]` and then you append items to it, when you create a new instance of the same class and you don't provide a list, that same original list will be used.

See this example:

```
>>> class A:
...     def __init__(self, values = []):
...             self.values = values
... 
>>> a = A()
>>> a.values.append(1)
>>> a.values
[1]
>>> new_a = A()
>>> new_a.values
[1]
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -93,6 +93,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- Fixed an issue that would cause multiple instances of the same class to behave
+  incorrectly if any of the given constructor arguments defaulted to a mutable
+  value (e.g. lists, dictionaries, objects).
+
 - Fixed an issue with `CartesiaTTSService` where `TTSTextFrame` messages weren't
   being emitted when the model was set to `sonic`. This resulted in the
   assistant context not being updated with assistant messages.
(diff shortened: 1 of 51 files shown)


markbackman (approved):
Seems like a safe change. In addition, it looks like you're moving many inputs to be optional probably where they were or should have been before and/or setting default values to None where they're optional. That's definitely helpful.

Out of curiosity, how did you find this issue?

aconchillo:
> Seems like a safe change. In addition, it looks like you're moving many inputs to be optional probably where they were or should have been before and/or setting default values to None where they're optional. That's definitely helpful.
> 

I don't think I changed any behavior. All of them had default values, which means we just set them to `None`. And now, we check if it's `None` and if it is we initialize with a new instance (not shared).

> Out of curiosity, how did you find this issue?

With observers. They were initialized as an empty list `[]`. The context aggregator test were failing so I printed the observers (because that's a PR that was failing) and I saw 30 observers registered (the turn tracking ones). So, I started looking and realized we were doing `List[BaseObserver] = []` in the constructor, which ends up being shared among all `PipelineTask`s that the test was running.

PR #1859, opened 2025-05-20
Update OTel attribute names

markbackman:

This change does a few things:
- Keeps the service span to just common names like `stt`, `llm`, or `tts`. This is a change from including the service name in the service type. This is to align with the change where span names don't include identifiers
- Related, change `service.name` to just `service` as there is nothing else nested
- Remove service type identifiers from from attribute names like `metrics.tts.character_count`
- Follow the OTel naming recommendations from Langfuse: https://langfuse.com/docs/opentelemetry/get-started#property-mapping. The only change I'm making is for llm tokens. This will allow them to revert these changes: https://github.com/langfuse/langfuse/pull/6903/files.

### DIFF
--- a/src/pipecat/utils/tracing/service_attributes.py
+++ b/src/pipecat/utils/tracing/service_attributes.py
@@ -18,6 +18,39 @@ if is_tracing_available():
     from opentelemetry.trace import Span
 
 
+def _get_gen_ai_system_from_service_name(service_name: str) -> str:
+    """Extract the standardized gen_ai.system value from a service class name.
+
+    Source:
+    https://opentelemetry.io/docs/specs/semconv/attributes-registry/gen-ai/#gen-ai-system
+
+    Uses standard OTel names where possible, with special case mappings for
+    service names that don't follow the pattern.
+    """
+    SPECIAL_CASE_MAPPINGS = {
+        # AWS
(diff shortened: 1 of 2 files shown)


markbackman:
FYI @marcklingen. Sorry for the extra work, but hopefully aligning attribute names will allow things to align better in the long run.

sunahsuh (commented):
Looking at the langfuse documentation, they link to the opentelemetry GenAI conventions here: https://opentelemetry.io/docs/specs/semconv/attributes-registry/gen-ai/ and they map those conventions to langfuse properties
I feel like we should stick to those so we can keep this implementation as generic as possible – wdyt?

marcklingen:
> FYI @marcklingen. Sorry for the extra work, but hopefully aligning attribute names will allow things to align better in the long run.

no worries, thanks for tagging me here, super helpful

made the updates on our side: https://github.com/langfuse/langfuse/pull/6944

markbackman:
@sunahsuh I've made changes to align with the spec where available. Only LLM services are called out; I tried to reuse fields for STT and TTS where it makes sense, but that was limited.

Let me know what you think about this approach.

sunahsuh (approved):
This is great, excited to see this ship!

PR #1866, opened 2025-05-21
BaseObserver: inherit from BaseObject so we can have events

aconchillo:


markbackman (approved):
Yes! I was wondering about this in working with the observer last week. This is nice 👏

PR #1868, opened 2025-05-22
Add Google streaming TTS as a base TTS service

aristid:
Add Google streaming TTS as of #1812. Greatly improves response times.

Following suggestion of @markbackman, this PR renames the non-streaming service to GoogleHttpTTSService, but see below, maybe it shouldn't be the standard service yet.
- Google streaming only supports Chirp3 and Journey voices.
- Google streaming doesn't support SSML.
- StreamingAudioConfig should support speaking_rate, but does not, currently commented out.
https://cloud.google.com/text-to-speech/docs/reference/rpc/google.cloud.texttospeech.v1#google.cloud.texttospeech.v1.StreamingAudioConfig

I added a second of silence (line 515), because otherwise the sentences are uncomfortably hurried after each other without any pause after the full stop.

### DIFF
--- a/src/pipecat/services/google/tts.py
+++ b/src/pipecat/services/google/tts.py
@@ -202,7 +202,7 @@ def language_to_google_tts_language(language: Language) -> Optional[str]:
     return language_map.get(language)
 
 
-class GoogleTTSService(TTSService):
+class GoogleHttpTTSService(TTSService):
     class InputParams(BaseModel):
         pitch: Optional[str] = None
         rate: Optional[str] = None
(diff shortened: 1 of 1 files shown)


src/pipecat/services/google/tts.py
```diff
@@ -202,7 +202,7 @@ def language_to_google_tts_language(language: Language) -> Optional[str]:
     return language_map.get(language)
 
 
-class GoogleTTSService(TTSService):
+class GoogleHttpTTSService(TTSService):
```
markbackman, line 205: In this class on line 227, please update it to:
```
        params = params or GoogleHttpTTSService.InputParams()
```

src/pipecat/services/google/tts.py
```diff
@@ -379,3 +379,148 @@ async def run_tts(self, text: str) -> AsyncGenerator[Frame, None]:
...
+        *,
+        credentials: Optional[str] = None,
+        credentials_path: Optional[str] = None,
+        voice_id: str = "en-US-Neural2-A",
+        sample_rate: Optional[int] = None,
+        params: InputParams = InputParams(),
+        **kwargs,
+    ):
+        super().__init__(sample_rate=sample_rate, **kwargs)
+
```
markbackman, line 405: After line 404, add:
```
        params = params or GoogleTTSService.InputParams()
```

src/pipecat/services/google/tts.py
```diff
@@ -379,3 +379,148 @@ async def run_tts(self, text: str) -> AsyncGenerator[Frame, None]:
...
+        language: Optional[Language] = Language.EN
+        gender: Optional[Literal["male", "female", "neutral"]] = None
+        google_style: Optional[Literal["apologetic", "calm", "empathetic", "firm", "lively"]] = None
+
+    def __init__(
+        self,
+        *,
+        credentials: Optional[str] = None,
+        credentials_path: Optional[str] = None,
+        voice_id: str = "en-US-Neural2-A",
```
markbackman, line 399: The Chirp 3 voices are not fully supported. Let's make it the default here in and in the HttpService too.

```suggestion
        voice_id: str = "en-US-Chirp3-HD-Charon",
```

src/pipecat/services/google/tts.py
```diff
@@ -379,3 +379,148 @@ async def run_tts(self, text: str) -> AsyncGenerator[Frame, None]:
...
+
+    def can_generate_metrics(self) -> bool:
+        return True
+
+    def language_to_service_language(self, language: Language) -> Optional[str]:
+        return language_to_google_tts_language(language)
+
+    @traced_tts
+    async def run_tts(self, text: str) -> AsyncGenerator[Frame, None]:
+        # Chirp3 and Journey voices only. Does not support SSML.
```
markbackman, line 458: Can this comment be removed? There is no SSML in this function.

src/pipecat/services/google/tts.py
```diff
@@ -379,3 +379,148 @@ async def run_tts(self, text: str) -> AsyncGenerator[Frame, None]:
...
+            voice = texttospeech_v1.VoiceSelectionParams(
+                language_code=self._settings["language"], name=self._voice_id
+            )
+
+            streaming_config = texttospeech_v1.StreamingSynthesizeConfig(
+                voice=voice,
+                streaming_audio_config=texttospeech_v1.StreamingAudioConfig(
+                    audio_encoding=texttospeech_v1.AudioEncoding.PCM,
+                    sample_rate_hertz=self.sample_rate,
+                    #speaking_rate=self._settings["rate"],
```
markbackman, line 473: Remove?

src/pipecat/services/google/tts.py
```diff
@@ -379,3 +379,148 @@ async def run_tts(self, text: str) -> AsyncGenerator[Frame, None]:
...
+                audio_buffer += chunk
+                while len(audio_buffer) >= CHUNK_SIZE:
+                    piece = audio_buffer[:CHUNK_SIZE]
+                    audio_buffer = audio_buffer[CHUNK_SIZE:]
+                    yield TTSAudioRawFrame(piece, self.sample_rate, 1)
+                    await asyncio.sleep(0)
+
+            if audio_buffer:
+                yield TTSAudioRawFrame(audio_buffer, self.sample_rate, 1)
+                await asyncio.sleep(0)
```
markbackman, line 512: Remove this line. Can you also remove the other lines in the HttpTTSService too. They're not needed.

src/pipecat/services/google/tts.py
markbackman: Ahh, looks like only Chirp 3 voices are supported. In trying a non-Chirp 3, I get:
```
error: TTS generation error: 400 Currently, only Chirp 3: HD voices are supported for streaming synthesis., fatal: False)
```

src/pipecat/services/google/tts.py
```diff
@@ -379,3 +379,148 @@ async def run_tts(self, text: str) -> AsyncGenerator[Frame, None]:
             logger.exception(f"{self} error generating TTS: {e}")
             error_message = f"TTS generation error: {str(e)}"
             yield ErrorFrame(error=error_message)
+
+
+class GoogleTTSService(TTSService):
```
markbackman, line 384: Can you add a brief docstring for this class mentioning the supported voices? Maybe something like:

```
"""Text-to-Speech service using Google Cloud Text-to-Speech API.

    Converts text to speech using Google's TTS models with streaming synthesis
    for low latency. Supports extensive language coverage and voice customization
    including pitch, rate, volume, and speaking style.

    Args:
        credentials: JSON string containing Google Cloud service account credentials.
        credentials_path: Path to Google Cloud service account JSON file.
        voice_id: Google TTS voice identifier (e.g., "en-US-Neural2-A").
        sample_rate: Audio sample rate in Hz.
        params: Voice parameters including pitch, rate, volume, and style settings.

    Notes:
        Requires Google Cloud credentials via service account JSON, file path, or
            default application credentials (GOOGLE_APPLICATION_CREDENTIALS env var).
        Only Chirp 3 voices are supported.

    Example:
        ```python
        tts = GoogleTTSService(
            credentials_path="/path/to/service-account.json",
            voice_id="en-US-Chirp3-HD-Charon",
            params=GoogleTTSService.InputParams(
                language=Language.EN_US,
                pitch="+2st",
                rate="1.2",
                google_style="empathetic"
            )
        )
        ```
    """

src/pipecat/services/google/tts.py
markbackman: Yes, let's remove. We can cover this in the docstring.

markbackman:
This looks great! Just a few minor comments. I tested and it works really well. Thanks for adding this 🙌

markbackman:
You can either install Ruff to lint the file or run the install the pre-commit hook:
https://github.com/pipecat-ai/pipecat?tab=readme-ov-file#%EF%B8%8F-hacking-on-the-framework-itself

aristid:
Streaming does not support pitch, rate, volume, and speaking style. These are set by Pipecat through SSML usually which isn't supported for streaming, it's text only. Google documentation mentions speaking_rate, but it doesn't work.
It's interesting asyncio.sleep(0) isn't needed, will try without it, maybe it's the source of quality degradation.
It seems to me that without additional pause the pause between sentences is shorter than the pause between words in the same sentence, which sounds strange.
May take me a few days to get back to this. If you'd like to publish soon, please feel free to do whatever changes are needed!

src/pipecat/services/google/tts.py
markbackman: @aristid can you update the default voice here to `en-US-Chirp3-HD-Charon`?

markbackman:
Looks good. Sorry for all the last things. This just needs to be rebased on `main` to fix the conflict in the CHANGELOG and then I think we can merge it!

PR #1871, opened 2025-05-22
Update default model for Anthropic to Claude Sonnet 4

markbackman:

Tested function calling and latency. Both seem inline with Sonnet 3.7.


filipi87 (approved):
LGTM!

PR #1876, opened 2025-05-23
Update AssemblyAI Streaming STT

m-ods:
Migrates AssemblyAI Speech-to-Text to the new v3 Streaming API for improved performance and reliability.

- Structured message types (Begin, Turn, Termination) with Pydantic models
- Replaced `assemblyai` dependency with `websockets`


src/pipecat/services/assemblyai/stt.py
```diff
@@ -73,97 +90,176 @@ async def cancel(self, frame: CancelFrame):
...
+            return f"{self._api_endpoint_base_url}?{query_string}"
+        return self._api_endpoint_base_url
 
     async def _connect(self):
-        """Establish a connection to the AssemblyAI real-time transcription service.
+        try:
+            ws_url = self._build_ws_url()
+            headers = {
+                "Authorization": self._api_key,
+                "User-Agent": f"AssemblyAI/1.0 (integration=Pipecat/{pipecat_version})",
```
markbackman, line 128: Can you explain what this does? Is this just for attribution?

src/pipecat/services/assemblyai/models.py
```diff
@@ -0,0 +1,61 @@
...
+
+
+# Union type for all possible message types
+AnyMessage = BeginMessage | TurnMessage | TerminationMessage
+
+
+class AssemblyAIConnectionParams(BaseModel):
+    sample_rate: int = 16000
+    encoding: Literal["pcm_s16le", "pcm_mulaw"] = "pcm_s16le"
+    formatted_finals: bool = False
```
markbackman, line 57: The TranscriptionFrame produced by the STTService is used in places that users will eventually see. It's common to either create a transcript of the conversation or add subtitles.

For that reason, it's helpful to capitalize and punctuate the transcripts. Can `formatted_finals` be set to `True`?

src/pipecat/services/assemblyai/stt.py
m-ods: This is just for attribution, and it doesn't change the way the API functions.

 For reference [see our SDK code here](http://github.com/AssemblyAI/assemblyai-python-sdk/blob/e14d31e0c9b6b92953b66d1816f8ad0a1b5891ef/assemblyai/streaming/v3/client.py#L77)

src/pipecat/services/assemblyai/models.py
m-ods: Formatted finals add about 50ms of latency - is there a way where we can send the formatted final for later use but use the unformatted final for the voice agent since that will be faster?

src/pipecat/services/assemblyai/stt.py
```diff
@@ -73,97 +90,176 @@ async def cancel(self, frame: CancelFrame):
...
-        self, transcript: str, is_final: bool, language: Optional[Language] = None
-    ):
-        """Handle a transcription result with tracing."""
-        await self.stop_ttfb_metrics()
-        await self.stop_processing_metrics()
+    def _build_ws_url(self) -> str:
+        """Build WebSocket URL with query parameters using urllib.parse.urlencode."""
+        params = {
+            k: str(v).lower() if isinstance(v, bool) else v
+            for k, v in self._connection_params.dict().items()
```
markbackman, line 115: `dict()` yields:
```
The method "dict" in class "BaseModel" is deprecated
  The `dict` method is deprecated; use `model_dump` instead.
```

src/pipecat/services/assemblyai/stt.py
```diff
@@ -73,97 +90,176 @@ async def cancel(self, frame: CancelFrame):
...
-                frame = InterimTranscriptionFrame(transcript.text, "", timestamp, language)
+            except Exception as e:
+                logger.warning(f"Error during termination handshake: {e}")
 
-            asyncio.run_coroutine_threadsafe(
-                self._handle_transcription(transcript.text, is_final, language),
-                self.get_event_loop(),
-            )
+            if self._receive_task:
+                self._receive_task.cancel()
```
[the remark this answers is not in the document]
markbackman: The try/except also happens due to the Pipecat utils, so we should be able to remove that.

markbackman:
We also need to set up tracing for the STT service. You can see an example in Deepgram: https://github.com/pipecat-ai/pipecat/blob/main/src/pipecat/services/deepgram/stt.py#L194-L199.

There is a service_decorators file, which shows how the decorators work. You need a traced function with `@traced_stt` that receives `self, transcript, is_final, language` as args. Here's that service decorator utility: https://github.com/pipecat-ai/pipecat/blob/main/src/pipecat/utils/tracing/service_decorators.py#L222.

src/pipecat/services/assemblyai/stt.py
```diff
@@ -37,31 +48,37 @@ def __init__(
         self,
         *,
         api_key: str,
-        sample_rate: Optional[int] = None,
-        encoding: Optional[AudioEncoding] = None,
-        language=Language.EN,  # Only English is supported for Realtime
+        language: Language = Language.EN,  # AssemblyAI only supports English
+        api_endpoint_base_url: str = "wss://streaming.assemblyai.com/v3/ws",
+        connection_params: AssemblyAIConnectionParams = AssemblyAIConnectionParams(),
+        vad_force_turn_endpoint: bool = False,
```
markbackman, line 54: What's the recommended use of this param? It seems like forcing the end of turn can reduce the latency. What's the downside to enabling it, if any?

src/pipecat/services/assemblyai/stt.py
```diff
@@ -73,97 +90,176 @@ async def cancel(self, frame: CancelFrame):
...
+            headers = {
+                "Authorization": self._api_key,
+                "User-Agent": f"AssemblyAI/1.0 (integration=Pipecat/{pipecat_version})",
+            }
+            self._websocket = await websockets.connect(
+                ws_url,
+                extra_headers=headers,
+            )
+            self._connected = True
+            self._receive_task = asyncio.create_task(self._receive_task_handler())
```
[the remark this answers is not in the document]
aconchillo: The reason is so we keep track of what tasks Pipecat creates and at the end we print if any of the tasks hasn't been cancelled. It helps with debugging.

src/pipecat/services/assemblyai/stt.py
```diff
@@ -73,97 +90,176 @@ async def cancel(self, frame: CancelFrame):
...
+        if not message.transcript:
+            return
+        await self.stop_ttfb_metrics()
+        if message.end_of_turn:
+            await self.push_frame(
+                TranscriptionFrame(
+                    message.transcript,
+                    "",  # participant
+                    time_now_iso8601(),
+                    self._language,
```
aconchillo, line 253: I just merged a PR that adds an optoinal `result` field to `TranscritionFrame` and `InterimTranscriptionFrame`. The idea is to set this field to whatever raw response the STT service returned. In this case, I think we would do `result=message`. Then, the user can inspect what the STT returned.

src/pipecat/services/assemblyai/stt.py
markbackman: This might be a safe feature to enable. The VAD detects end of speech and emits a UserStoppedSpeakingFrame. This happens only after speech has stopped and the stop_secs timeout has occurred. So, usually that's ~0.5 to 0.8 seconds of silence. It could be optimal and not risky to use this feature to force a final transcription. We do something similar with Deepgram.

markbackman (approved):
LGTM!

Just one remaining question about how to set the default value of `vad_force_turn_endpoint`. I think we could set it to `True`, which would match Deepgram, but you're more aware of risks related to AssemblyAI than I am.

If the concern is just Pipecat cutting off audio, then I think we're OK. the UserStoppedSpeakingFrame is a clear indicator that the transcript can be finalized.

The only other TODO is to add a CHANGELOG entry.

PR #1877, opened 2025-05-23
Aggregate TextFrames from LLM Completion for Tracing

a6kme:
I would like to send my traces to an exporter and then annotate those traces to create eval datasets. In order for me to do that, I need to have responses in the traces. 

This PR aggregates incoming TextFrames from LLMService to achieve that.


markbackman:
Hey, thanks for the submission. This is a great idea. I followed a similar path but made a few additional changes: https://github.com/pipecat-ai/pipecat/pull/1886. I'm going to close this out. This will be in the next release, which should be early next week.

PR #1878, opened 2025-05-23
Add PlivoFrameSerializer

markbackman:


markbackman:
> Look good. This is basically the same approach that we are already doing with the TwilioFrameSerializer, right ? 
> 
> 
> 
> Cool. 🎉🚀

Yup! Follows the same pattern.

PR #1889, opened 2025-05-24
Add DTMFAggregator

markbackman:

This adds a DTMFAggregator, inspired by this older PR: https://github.com/pipecat-ai/pipecat/pull/1362.

### DIFF
--- a/src/pipecat/processors/aggregators/dtmf_aggregator.py
+++ b/src/pipecat/processors/aggregators/dtmf_aggregator.py
@@ -14,6 +14,7 @@ from pipecat.frames.frames import (
     Frame,
     InputDTMFFrame,
     KeypadEntry,
+    StartFrame,
     TranscriptionFrame,
 )
 from pipecat.processors.frame_processor import FrameDirection, FrameProcessor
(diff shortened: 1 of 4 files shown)


markbackman:
I've struggled to get tests to be 100% happy. They're running and passing now but with this warning:
```
RuntimeWarning: coroutine 'DTMFAggregator._aggregation_task_handler' was never awaited
```

@aconchillo maybe you have a tip?

src/pipecat/processors/aggregators/dtmf_aggregator.py
```diff
@@ -0,0 +1,162 @@
...
+                    if self._aggregation:
+                        await self._flush_aggregation()
+                finally:
+                    self._digit_event.clear()
+        except asyncio.CancelledError:
+            # Task was cancelled - exit cleanly
+            pass
+        except Exception as e:
+            # Log unexpected errors but don't crash
+            print(f"Unexpected error in DTMF aggregation task: {e}")
```
aconchillo, line 142: you just need the while loop. we internally capture `CancelledError`. also, when you capture `CancelledError` you need to re-raised it. But anyways, I think just the while loop would be fine and shorter.

src/pipecat/processors/aggregators/dtmf_aggregator.py
```diff
@@ -0,0 +1,162 @@
...
+                await self._flush_aggregation()
+            await self._stop()
+            await self.push_frame(frame, direction)
+            return
+
+        # Push all other frames downstream immediately
+        await self.push_frame(frame, direction)
+
+        if isinstance(frame, InputDTMFFrame):
+            await self._handle_dtmf_frame(frame)
```
aconchillo, line 87: maybe this would be a bit cleaner:

```
        if isinstance(frame, (EndFrame, CancelFrame)):
            # Flush any pending aggregation before stopping
            if self._aggregation:
                await self._flush_aggregation()
            await self._stop()
        elif isinstance(frame, InputDTMFFrame):
            await self._handle_dtmf_frame(frame)
            
        # Push all other frames downstream immediately
```

src/pipecat/processors/aggregators/dtmf_aggregator.py
```diff
@@ -0,0 +1,162 @@
...
+
+    async def _stop(self) -> None:
+        """Stops and cleans up the aggregation task."""
+        if self._aggregation_task:
+            await self.cancel_task(self._aggregation_task)
+            self._aggregation_task = None
+
+    async def cleanup(self) -> None:
+        await super().cleanup()
+        if self._aggregation_task:
```
aconchillo, line 69: no need to check, already checked inside.

src/pipecat/processors/aggregators/dtmf_aggregator.py
```diff
@@ -0,0 +1,162 @@
...
+            self._create_aggregation_task()
+
+        digit_value = frame.button.value
+        self._aggregation += digit_value
+
+        # If this is the first digit, send BotInterruptionFrame upstream
+        # But use a separate task to avoid interfering with current frame processing
+        if len(self._aggregation) == 1:
+            # Use create_task to avoid queue issues
+            self.create_task(self._send_interruption_frame())
```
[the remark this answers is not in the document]
markbackman: For some reason, it seems that pushing the BotInterruptionFrame is interfering with the asyncio queue. Isolating it in a task seems to resolve it.

src/pipecat/processors/aggregators/dtmf_aggregator.py
markbackman: FYI: I need to push the InputDTMFFrame before the _handle_dtmf_frame in order to get the InputDTMFFrames in order before the TranscriptionFrame.

I cleaned this up a bit to ensure unit tests are passing.

markbackman:
@aconchillo this is cleaned up. All errors are resolved, tests are passing, and it works well with the Twilio example, which I'm not including here. We might need a simple DTMF example in foundational now that we support Twilio there...

src/pipecat/processors/aggregators/dtmf_aggregator.py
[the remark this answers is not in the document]
aconchillo: Ah, yes! This error will be solved when we queue system frames!

src/pipecat/processors/aggregators/dtmf_aggregator.py
aconchillo: Ah, OK!

aconchillo: The current code is good and aligns with other parts of pipecat (i.e. push_frame in every if block)

aconchillo (approved):

PR #1898, opened 2025-05-26
Fixing TavusTransport with some TTS services.

filipi87:
Fixing TavusTransport with some TTS services.

### DIFF
--- a/src/pipecat/transports/services/tavus.py
+++ b/src/pipecat/transports/services/tavus.py
@@ -11,6 +11,8 @@ from pydantic import BaseModel
 
 from pipecat.audio.utils import create_default_resampler
 from pipecat.frames.frames import (
+    BotStartedSpeakingFrame,
+    BotStoppedSpeakingFrame,
     CancelFrame,
     EndFrame,
     Frame,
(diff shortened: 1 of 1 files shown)


markbackman (approved):
LGTM

Tested with DeepgramTTSService and all looks well on 21-tavus-layer-tavus-transport.py.

aconchillo:
@filipi87 Can you describe the issue we are fixing?

aconchillo (approved):

aconchillo:
> Hi @aconchillo ,
> 
> The issue was that we were using the `TTSStartedFrame` to create the inference ID that we sent to Tavus, which we are calling `_current_idx_str` in both `TavusTransport` and `TavusVideoService`.
> 
> So, the problem was that the audio frames, TTSStartedFrame, and TTSStoppedFrames were handled in different queues. Consequently, there were instances where `_current_idx_str` was updated before all the audio was processed by Tavus. This resulted in only parts of the audio being spoken, typically the beginning of each sentence.
> 
> Another issue involved how we calculated the wait time, which sometimes caused the replica to speak the first utterance but then remain muted for an extended period.
> 
> Both issues are easily reproducible when using `DeepgramTTS` or `OpenAITTS`.

OK! Thank you!

PR #1902, opened 2025-05-27
Updated SmallWebRTCTransport to align with how other transports handle on_client_disconnected

filipi87:
Updated SmallWebRTCTransport to align with how other transports handle on_client_disconnected


aconchillo:
LGTM! Thank you!

PR #1910, opened 2025-05-28
Add Soniox STT service integration

matejmarinko-soniox:
This PR introduces a new STT provider `SonioxSTTService` that integrates with [Soniox’s](https://soniox.com) real-time WebSocket API to provide robust multilingual speech-to-text capabilities within the Pipecat framework.

## STTService

Usage:

```py
from pipecat.services.soniox.stt import SonioxSTTService
from pipecat.services.soniox.config import SonioxInputParams

stt = SonioxSTTService(
    api_key=os.getenv("SONIOX_API_KEY"),
    params=SonioxInputParams(
        language_hints=[Language.EN, Language.ES, Language.JA, Language.ZH],
    ),
)
```

## Examples

Pull request also adds two relevant examples for `07za-interruptible-soniox.py` and `13f-soniox-transcription.py`.

Regarding the interruptible example - all similar examples start with a letter prefix, but we ran out of letters. Let me know what would be a suitable file name.

## Already in use

We are already using Pipecat internally for our own production deployments, including the voice-to-text chatbot live on [soniox.com/docs](https://soniox.com/docs), which is powered by the integration in this PR.

Documentation will be added in a separate pull request to pipecat-ai/docs repository.

Resolves: #2141

### DIFF
--- a/src/pipecat/services/soniox/stt.py
+++ b/src/pipecat/services/soniox/stt.py
@@ -4,6 +4,8 @@
 # SPDX-License-Identifier: BSD 2-Clause License
 #
 
+"""Soniox speech-to-text service implementation."""
+
 import asyncio
 import json
 import time


matejmarinko-soniox:
Hi @aconchillo, @markbackman just checking in on this PR—happy to answer any questions or make changes if needed!

src/pipecat/services/soniox/__init__.py
```diff
@@ -0,0 +1,13 @@
+#
+# Copyright (c) 2024–2025, Daily
+#
+# SPDX-License-Identifier: BSD 2-Clause License
+#
+
+import sys
```
markbackman: You can leave this file blank. We're moving away from this pattern in an upcoming release. The recommended way to access a service is via the full path, e.g.
`pipecat.services.soniox.stt`.

src/pipecat/services/soniox/config.py
```diff
@@ -0,0 +1,36 @@
...
+
+from pydantic import BaseModel
+
+from pipecat.transcriptions.language import Language
+
+
+class SonioxInputParams(BaseModel):
+    """Real-time transcription settings.
+
+    Attributes:
```
markbackman: For correct docstring auto-generation:

```suggestion
    Parameters:
```

src/pipecat/services/soniox/config.py
```diff
@@ -0,0 +1,36 @@
...
+        languages: List of language codes to use for transcription
+        code_switching: Whether to auto-detect language changes during transcription
+    """
+
+    model: str = "stt-rt-preview"
```
markbackman: Should model be an arg in `SonioxSTTService`?

src/pipecat/services/soniox/config.py
```diff
@@ -0,0 +1,36 @@
...
+
+    audio_format: Optional[str] = "pcm_s16le"
```
markbackman: Same for `audio_format`, `num_channels`, `sample_rate`. Might they belong in the `SonioxSTTService` args?

`sample_rate` is already included there, so this is a duplicate.

src/pipecat/services/soniox/config.py
markbackman: Can you update the Parameters to match what's in the class?

src/pipecat/services/soniox/stt.py
```diff
@@ -0,0 +1,369 @@
...
+    For complete API documentation, see: https://soniox.com/docs/speech-to-text/api-reference/websocket-api
+    """
+
+    def __init__(
+        self,
+        *,
+        api_key: str,
+        url: str = "wss://stt-rt.soniox.com/transcribe-websocket",
+        sample_rate: Optional[int] = None,
+        params: SonioxInputParams = SonioxInputParams(),
```
markbackman: Aligning with other services:
```suggestion
        params: Optional[SonioxInputParams] = SonioxInputParams(),
```

src/pipecat/services/soniox/stt.py
```diff
@@ -0,0 +1,369 @@
...
+            enable_vad: Listen to `UserStoppedSpeakingFrame` to send finalize message to Soniox.
+            auto_finalize_delay_ms: If no new tokens are received for a while and there is active
+                transcription (only InterimTranscriptionFrame), finalize the transcription by
+                sending the finalize message so user can receive the final transcript. If set
+                to `None`, the auto finalize feature is disabled.
+            **kwargs: Additional arguments passed to the STTService.
+        """
+        sample_rate = sample_rate or (params.sample_rate if params.sample_rate else None)
+        super().__init__(sample_rate=sample_rate, **kwargs)
+
```
markbackman, line 141: Related to the params comment above, after super().__init__, add:
```
 params = params or SonioxInputParams()
```

src/pipecat/services/soniox/stt.py
```diff
@@ -0,0 +1,369 @@
...
+        # Transcription frame will be only sent after we get the "endpoint" event.
+        self._final_transcription_buffer = ""
+
+        async def send_endpoint_transcript():
+            if self._final_transcription_buffer:
+                await self.push_frame(
+                    TranscriptionFrame(
+                        self._final_transcription_buffer,
+                        "",
```
markbackman: This change may have occurred after you created the PR, but the base STTService class now has a user ID to apply, based on the transport in use.

```suggestion
                        self._user_id,,
```

src/pipecat/services/soniox/stt.py
```diff
@@ -0,0 +1,369 @@
...
+                    else:
+                        non_final_transcription += token["text"]
+
+                if self._final_transcription_buffer or non_final_transcription:
+                    await self.push_frame(
+                        InterimTranscriptionFrame(
+                            # Even final tokens are sent as interim tokens as we want to send
+                            # nicely formatted messages - therefore waiting for the endpoint.
+                            self._final_transcription_buffer + non_final_transcription,
+                            "",
```
markbackman: Same as the TranscriptionFrame:
```suggestion
                            self._user_id,,
```

src/pipecat/services/soniox/stt.py
```diff
@@ -0,0 +1,369 @@
...
+        # Transcription frame will be only sent after we get the "endpoint" event.
+        self._final_transcription_buffer = ""
+
+        async def send_endpoint_transcript():
+            if self._final_transcription_buffer:
+                await self.push_frame(
+                    TranscriptionFrame(
+                        self._final_transcription_buffer,
+                        "",
+                        time_now_iso8601(),
```
markbackman: Is there a language value that can be added to this frame as well? Same for the InterimTranscriptionFrame.

markbackman: Also, we've started to include the raw result, as developers have been making use of it. The latest TranscriptionFrame definitions is:
```
@dataclass
class TranscriptionFrame(TextFrame):
    """Text frame containing speech transcription data.

    A text frame with transcription-specific data. The `result` field
    contains the result from the STT service if available.

    Parameters:
        user_id: Identifier for the user who spoke.
        timestamp: When the transcription occurred.
        language: Detected or specified language of the speech.
        result: Raw result from the STT service.
    """

    user_id: str
    timestamp: str
    language: Optional[Language] = None
    result: Optional[Any] = None
```

Where TextFrame, as a base class provides `text: str`. The same is true of the InterimTranscriptionFrame.

src/pipecat/services/soniox/stt.py
```diff
@@ -0,0 +1,369 @@
...
+        if not self._websocket:
+            return
+
+        # Transcription frame will be only sent after we get the "endpoint" event.
+        self._final_transcription_buffer = ""
+
+        async def send_endpoint_transcript():
+            if self._final_transcription_buffer:
+                await self.push_frame(
+                    TranscriptionFrame(
```
markbackman, line 311: When pushing the transcription frame, we also want to capture this event for tracing. There's an `@traced_stt` decorator that you can apply to a tracing method, like this:
https://github.com/pipecat-ai/pipecat/blob/main/src/pipecat/services/deepgram/stt.py#L262-L267

src/pipecat/services/soniox/config.py
matejmarinko-soniox: `SonioxInputParams` object is passed as an argument when constructing the `SonioxSTTService`. 

These parameters were designed to match our Websocket API input as close as possible - https://soniox.com/docs/speech-to-text/api-reference/websocket-api#configuration-parameters. Other top-level parameters are more Pipecat specific.

I followed the implementation of other services as a base. Parameters in [Gladia](https://github.com/pipecat-ai/pipecat/blob/main/src/pipecat/services/gladia/stt.py#L208) are passed in a similar way. We can also move all parameters to a single level, if you think that makes more sense.

Regarding the `sample_rate`:
- Soniox supports multiple audio formats. For compressed audio, `sample_rate` parameter in `SonioxInputParams` is irrelevant. I wanted to keep the support for streaming of other audio formats.
- If I understand correctly, Pipecat's [STTService](https://github.com/pipecat-ai/pipecat/blob/main/src/pipecat/services/stt_service.py#L30) needs internal sample rate, which is the parameter passed directly in `SonioxSTTService`. If that parameter is not needed, we will remove that.

src/pipecat/services/soniox/stt.py
```diff
@@ -0,0 +1,369 @@
...
+            await self.push_error(ErrorFrame(f"{self} error: {e}"))
+
+    async def _finalize_if_no_tokens_task_handler(self):
+        """Call finalize if no new tokens are received for a configured duration."""
+        if not self._websocket or self._websocket.closed or self._auto_finalize_delay_ms is None:
+            return
+
+        try:
+            while True:
+                await asyncio.sleep(0.5)
```
markbackman: Can you explain why the 0.5 sec sleep is required?

src/pipecat/services/soniox/config.py
```diff
@@ -0,0 +1,36 @@
...
+#
+
+from typing import List, Optional
+
+from pydantic import BaseModel
+
+from pipecat.transcriptions.language import Language
+
+
+class SonioxInputParams(BaseModel):
```
markbackman: If this is the only config class, you could move `SonioxInputParams` into the .stt file, so that you only have to import once. I think either way is fine. Just an option.

examples/foundational/07za-interruptible-soniox.py
```diff
@@ -0,0 +1,115 @@
+#
```
markbackman: I apologize for being slow in reviewing this one. I think since this was submitted the example format changed. You can grab anyone of the 07* examples and mimic it, just replacing the STT service.

The same goes for the 13f example.

markbackman:
This looks really good. I tested the service and it works nicely 👏 

In addition to the comment, one other small request. Our docstrings format changed in the most recent release. You can see the new style guide here:
https://github.com/pipecat-ai/pipecat/blob/main/CONTRIBUTING.md#code-style-and-documentation

This will ensure that reference docs are auto-generated.

src/pipecat/services/soniox/stt.py
```diff
@@ -0,0 +1,369 @@
...
+    """
+
+    def __init__(
+        self,
+        *,
+        api_key: str,
+        url: str = "wss://stt-rt.soniox.com/transcribe-websocket",
+        sample_rate: Optional[int] = None,
+        params: SonioxInputParams = SonioxInputParams(),
+        enable_vad: bool = True,
```
markbackman: Can you explain how VAD event work? Most Pipecat bots run with a local VAD to detect when the user has started and stopped speaking. Based on this implementation, it seems like you're relying on these events to determine when to finalize. Is that right?

src/pipecat/services/soniox/config.py
matejmarinko-soniox: Deepgram also has the exact same support for sample rates - https://github.com/pipecat-ai/pipecat/blob/main/src/pipecat/services/deepgram/stt.py#L79. If you think this should be done differently, please let me know.

src/pipecat/services/soniox/stt.py
matejmarinko-soniox: There are two different options users can choose.
1. Use local VAD (this is toggle by `enable_vad`) parameter. When users stops speaking, we [send finalize event](https://soniox.com/docs/speech-to-text/core-concepts/manual-finalization). All tokens are immediately finalized and `<fin>` token is returned to reflect that.
2. [Endpoint detection](https://soniox.com/docs/speech-to-text/core-concepts/endpoint-detection) provided by Soniox (similar to VAD). This returns `<end>` tokens when users stops speaking.

When either `<fin>` or `<end>` is received, transcription is sent further.

src/pipecat/services/soniox/config.py
markbackman: I'm happy to align closely with your API.

Re: `sample_rate`, it might be the only outlier.
> Soniox supports multiple audio formats. For compressed audio, sample_rate parameter in SonioxInputParams is irrelevant. I wanted to keep the support for streaming of other audio formats.

Pipecat receives audio in PCM format, which needs to align with the `sample_rate` used by the STT service. Pipecat has a convenience where the sample rate is passed to all processors in the pipeline via the StartFrame to ensure that the processors are all using the same audio input and output sample rates, respectively.

So, the sample rate in SonioxInputParams is the sample rate that should be set by the StartFrame. The StartFrame is handled in the base STTService class. This is why the `sample_rate` is passed to `super().__init__`.

I believe there should be only a single `sample_rate` value, which should reside in the SonioxSTTService constructor. (If the transport input is set to a different value than the STTService, then the audio won't be handled correctly, so it's best to ensure that they're aligned). Does that make sense?

src/pipecat/services/soniox/stt.py
matejmarinko-soniox: This was just a fallback, which is no longer needed as our API now always returns end tokens. Will remove the finalization task.

src/pipecat/services/soniox/stt.py
markbackman: Ok, that's helpful. Is it possible to use the local VAD and the endpoint detection at the same time? It sounds like an either/or case. Is that right? Though, I see that you have both `enable_vad` set to True and `enable_endpoint_detection` set to True by default. Can you explain how they work together?

I have other thoughts, but first, I want to make sure I understand how these two interact.

src/pipecat/services/soniox/stt.py
matejmarinko-soniox: Soniox returns language at token level. Single transcription frame could contain multiple languages. 
I think returning raw result is the ideal solution here as it returns the correct result to developers.

examples/foundational/07za-interruptible-soniox.py
matejmarinko-soniox: What would you suggest for the example naming? In case of `07-` all letters from `07a` to `07z` are already taken. Is naming `07za` okay?

src/pipecat/services/soniox/config.py
matejmarinko-soniox: Will leave it here, same as other API params.
https://github.com/pipecat-ai/pipecat/pull/1910#discussion_r2194200583

src/pipecat/services/soniox/stt.py
matejmarinko-soniox: I changed the implementation so we are buffering tokens instead of only string results. Raw result is now returned on both transcription frame types.

src/pipecat/services/soniox/stt.py
matejmarinko-soniox: Nice catch, thanks! Yes, generally it is advised to use either one or the other. I changed the default `enable_vad` to `False`.

Developers can also use both if they want to, but might cause unexpected issues.
For example, getting `<fin>` event after finalization is called and then `<end>` event slightly later with only punctuation for the end of sentence added.

This might still add value in some cases, where the transcription delay is crucial (so tokens are forwarded when either one of the methods detects an end).

I am curious what your other thoughts are.

src/pipecat/services/soniox/config.py
matejmarinko-soniox: I have no strong opinions here, moved it to the `stt.py`.

src/pipecat/services/soniox/config.py
matejmarinko-soniox: So Pipecat always receives single channel audio in PCM format? In that case, this totally makes sense.

For all examples this seems to be true, but I wasn't 100% sure if different audio format can be used in custom transformers.

matejmarinko-soniox: Code was updated so there is a single `sample_rate` parameter in the `SonioxSTTService` constructor.

examples/foundational/07za-interruptible-soniox.py
matejmarinko-soniox: @markbackman, just circling back on this. I'd appreciate your thoughts so we can move this forward.

markbackman: Let's go with `07aa`. This will put you higher on the list too :)

src/pipecat/services/soniox/stt.py
markbackman: Ok, that's what I thought. For Pipecat, the optimal approach is to use the local VAD. Why? It's faster as there is no network roundtrip and it yields `UserStartedSpeakingFrame`/`UserStoppedSpeakingFrame`.

My recommendation would be to not include an `enable_vad` arg and instead make that the default behavior. Then, when the `enable_endpoint_detection` arg is set to True, it would assume the local VAD is not in use and would instead use events from Soniox which could be used by the application to push frames. You can see this in the deepgram/stt source and [07c-vad](https://github.com/pipecat-ai/pipecat/blob/main/examples/foundational/07c-interruptible-deepgram-vad.py) example, where using the Deepgram VAD emits events, then those VAD events need to be handled by the developer.

Happy to have a quick chat today if you'd like to talk this through. Just let me know.

examples/foundational/07za-interruptible-soniox.py
matejmarinko-soniox: Great! Renamed.

src/pipecat/services/soniox/stt.py
matejmarinko-soniox: I see your point, but we're intentionally sticking with our own endpoint detection. We believe it provides a significantly better user experience because it's a core part of our model, not a standard VAD that just waits for periods of silence. It’s more advanced and allows for a much more natural flow. If you have time, I suggest you to try endpoint detection in [Soniox Console](https://console.soniox.com) and see the difference in behaviour.

Regarding events, you're right that we don't have `UserStartedSpeakingFrame` or `UserStoppedSpeakingFrame` since our definition of an endpoint is different.
- Our model doesn't return any "user started speaking" events, so we cannot accurately return those events.
- In case of "user stopped speaking" events, we're open to adding event-emitting if you think that makes sense. But again, that is not exactly same as events emmited by VADs.

Hope this clarifies our perspective. Let me know if that makes sense.

markbackman: Pipecat relies on `UserStartedSpeakingFrame` and `UserStoppedSpeakingFrame`, so not including them will create a poor experience. That is because the STT is not a standalone program. It needs to communicate and interoperate with all other programs. Part of that is indicating to the rest of the pipeline when the user is speaking. This is important for aggregating user input.

At a minimum, you need to provide events so that a developer can use them.

Also, that's great that Soniox has excellent endpointing! But, similar to the previous point, Pipecat operates as an orchestrator of all services in the Pipeline. It's best to align with the architecture of the pipeline. If you don't, then SonioxSTTService will not perform well.

Getting an opinion from a few other maintainers on this one: cc @filipi87 @kompfner @kwindla.

kwindla: I strongly agree that we should try to align all turn taking implementations to emit the same frame types.

Being able to use pipeline elements interchangeably is a core design goal for Pipecat.

- The context aggregators use both `UserStartedSpeaking` and `UserStoppedSpeaking` frames in the aggregation logic.
- User interfaces use both those frames (via RTVI messages) to update UI state to show whether a user is speaking and to display interim transcript messages with the right styles.
- Some developers will have written custom pipeline elements that use both those frames.

Because of this, I think it's worth trying to figure out a standard way to support the `UserStartedSpeaking` and `UserStoppedSpeaking` frames in the Soniox implementation.

Maybe that means always recommending that users configure pipelines with local VAD, but the Soniox implementation blocks and re-sends the speaking event frames from VAD based on the Soniox inference/state. Or maybe there's another way to do it that I haven't thought of!

matejmarinko-soniox: Let me know if I understand correctly: We can still recommend using VAD and then possibly block sending the `UserStartedSpeaking` or `UserStoppedSpeaking` according to state of Soniox transcription. After Soniox transcription returns an endpoint, we should return the transcription along with the `UserStoppedSpeaking` frame. 

How can we block sending of the events? Ideally, one could simply "not forward" the frame in `process_frame()` method.

For example:
- User starts speaking and we receive `UserStartedSpeaking` frame -> forward it normally.
- When user stops speaking, but Soniox API still hasn't returned the `<end>` token, we don't send `UserStoppedSpeaking` further. After the Soniox API returns the `<end>` token, we create a new `UserStoppedSpeaking` frame and send it along with the transcription.

There are multiple possible cases, I made a [simple Excalidraw](https://excalidraw.com/#json=FKNAYGiW7p1aIoRoW-4aQ,n5_GBSXN7Ce4Ay9eGeSt4Q), which should cover all cases.

Am I thinking in the right direction with this approach?

kompfner: I might be totally missing something, but it seems like maybe we’re talking past each other a bit in this comment thread.

It doesn’t seem to me like your implementation is swallowing local-VAD-driven events like `UserStartedSpeakingFrame` or `UserStoppedSpeakingFrame`—the rest of the pipeline can still make use of those as always. And in the example you provided, `vad_analyzer` is indeed still enabled. So if I’m understanding things right, the `enable_vad` setting’s purpose is just for telling Soniox whether or not to use the local VAD event to force a transcription endpoint—it’s not meant as a signal to the developer that they should be ditching `vad_analyzer` and instead replacing it wholesale with Soniox’s endpoint detection.

Is my understanding correct? If so…

Then I don’t see much of an issue here. `enable_vad` seems pretty much analogous to, say, `AssemblyAISTTService`’s `vad_force_turn_endpoint`. 

The only suggestions here might be:
* Naming: something more like the aforementioned `vad_force_turn_endpoint` might make it clearer what it does and prevent confusion about whether it should be used alongside a `vad_analyzer` at all
* Default value: as far as I can tell, all the other Pipecat STT services that have a “finalize” action rely on `UserStoppedSpeakingFrame` by default, so having a final transcript available as closely as possible to that event might align better with Pipecat users’ expectations. I believe only `AssemblyAISTTService` allows for configuring whether to use their own endpointing v the local-VAD-driven endpointing (and of course `DeepgramTTSService` gives you a way of opting out of using `vad_analyzer` entirely, but, again, I don’t *think* that’s what we’re talking about here)

kompfner: One more suggestion, alluded to earlier in this thread: providing _both_ `enable_vad` and `enable_endpoint_detection` might be a bit confusing to the developer, if the recommendation is to use one and only one of them. (In particular, it'd be great to avoid the possible misconfiguration of setting them both to `False`, which would...what? Make it so that final transcriptions are never emitted?)

matejmarinko-soniox: @kompfner your understanding is correct. I just checked the AssemblyAI's implementation and this is pretty much exactly the same as ours.

- Naming: Agree that the name is not perfect, will update it to reflect that.
- Default value: We would still prefer having this turned off by default. As stated above, we believe this will lead to better results (transcription chunks that make "more sense"). If you really think going with `vad_force_turn_endpoint` enabled would align much better with other Pipecat's pipeline parts, we can still go with that.

Going with a single parameters sounds like a better idea. Thanks for the suggestion!

kompfner: > We would still prefer having this turned off by default

Linking to [this very similar discussion](https://github.com/pipecat-ai/pipecat/pull/1876#discussion_r2114697688) where we ended up with `vad_force_turn_endpoint` defaulting to `True` (I think the takeaway is that the latency benefit of the local-VAD-driven finalization may have been worth it). But ultimately you know the service best and have done the most testing.

matejmarinko-soniox: We have removed the `enable_endpoint_detection` parameter and renamed the existing one to `vad_force_turn_endpoint`, as suggested. This change ensures that either Soniox endpointing or a VAD-forced endpoint is active at all times.

Regarding the default value, we will keep `vad_force_turn_endpoint` set to `False`. We believe our endpoint detection provides a much better user experience than traditional VAD.

For example, during natural pauses in speech (e.g., when a user is thinking), our system does not prematurely end the turn because it is aware of the conversational context and tone. This is the seamless experience we want as the default for Soniox.

I will update the documentation to reflect this behaviour.

kompfner:
Gave the service a test run and it worked very nicely for me! 👏 

Given @markbackman's prior thorough code review and our satisfactory resolution to the big open discussion, this looks good to me! 👍 

The only remaining thing is fixing the minor docstring issue that `ruff check` picked up on.

PR #1914, opened 2025-05-29
output transport can now handle DTMF keypress

aconchillo:

It is now possible to handle DTMF keypress in the output transport. For example, the DailyTransport has support for DTMF and it is now possible to send DTMF by just pushing DTMF frames.


aconchillo:
> LGTM. just double checking if we did mean to include the changes to `write_raw_video_frame`.

I have removed those changes. I'll put it in another PR.

PR #1915, opened 2025-05-29
use uvloop as the new event loop on Linux and macOS

aconchillo:

See https://uvloop.readthedocs.io/. This makes Pipecat run a bit smoother! We should test it thoroughly.

### DIFF
--- a/src/pipecat/__init__.py
+++ b/src/pipecat/__init__.py
@@ -4,10 +4,26 @@
 # SPDX-License-Identifier: BSD 2-Clause License
 #
 
+import asyncio
+import sys
 from importlib.metadata import version
 
 from loguru import logger
 
 __version__ = version("pipecat-ai")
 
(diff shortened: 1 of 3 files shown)


src/pipecat/__init__.py
```diff
@@ -4,10 +4,26 @@
...
+
+        import uvloop
+
+        event_loop = f"uvloop {uvloop.__version__}"
+        asyncio.set_event_loop_policy(uvloop.EventLoopPolicy())
+    except ImportError:
+        logger.debug(f"Couldn't find `uvloop`")
+        pass
+
+logger.info(f"ᓚᘏᗢ Pipecat {__version__} ({event_loop}; Python {sys.version}) ᓚᘏᗢ")
```
markbackman, line 29: Good idea about the python version!

markbackman (approved):
LGTM

PR #1918, opened 2025-05-29
Fix OTel examples to use new runner

markbackman:

Should we create one common location for the run.py and daily_runner.py? That will help us avoid duplicating the same file everywhere.

Also, the relative import was not working for me. I had to fallback to the previous approach.

### DIFF
--- a/examples/open-telemetry/daily_runner.py
+++ b/examples/open-telemetry/daily_runner.py
@@ -1,0 +1,64 @@
+#
+# Copyright (c) 2024–2025, Daily
+#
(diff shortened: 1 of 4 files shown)


aconchillo:
> Should we create one common location for the run.py and daily_runner.py? That will help us avoid duplicating the same file everywhere.
> 

Yes, we should! I just did it: https://github.com/pipecat-ai/pipecat/pull/1924

> Also, the relative import was not working for me. I had to fallback to the previous approach.

Yeah, I figured. I forgot to test it.

markbackman:
Closed in favor of #1924

PR #1921, opened 2025-05-29
add STT result field to TranscriptionFrame/InterimTranscriptionFrame

aconchillo:

Some people have requested to access STT service results in case they use custom settings. This PR adds a new `result` field with whatever comes from the STT service.


markbackman:
Is the idea to provide the raw response back or is "result" a specific field for each of these providers?

PR #1923, opened 2025-05-29
Fix implementation of ttfb_ms metric

philipp-eisen:
Fixes: `ttfb_ms` was providing seconds instead of milliseconds

### DIFF
--- a/src/pipecat/processors/metrics/frame_processor_metrics.py
+++ b/src/pipecat/processors/metrics/frame_processor_metrics.py
@@ -28,18 +28,18 @@ class FrameProcessorMetrics:
         self._should_report_ttfb = True
 
     @property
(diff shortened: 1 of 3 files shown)


aconchillo:
LGTM! But I'll leave the final word to @markbackman on this one.

markbackman:
We've been trying to align with OpenTelemetry standards. OTel doesn't have a time standard but generally recommends seconds. Perhaps instead, the change could be to update the attribute name to `ttfb_sec`?

markbackman:
@philipp-eisen can you make sure you either install Ruff or install the pre-commit hooks?
https://github.com/pipecat-ai/pipecat?tab=readme-ov-file#%EF%B8%8F-hacking-on-the-framework-itself

PR #1925, opened 2025-05-29
Websocket client web app to test Twilio.

filipi87:
Websocket client web app to test Twilio.


examples/twilio-chatbot/ws_test_client/src/app.ts
```diff
@@ -0,0 +1,255 @@
...
+  /**
+   * Initialize and connect to the bot
+   * This sets up the RTVI client, initializes devices, and establishes the connection
+   */
+  public async connect(): Promise<void> {
+    try {
+      const startTime = Date.now();
+
+      //const transport = new DailyTransport();
+      const transport = new PipecatWebsocketTransport();
```
filipi87, line 157: In order to make this work, we would need to change the `PipecatWebsocketTransport` to accept custom `serializers`, similar to how the `FastAPIWebsocketTransport` does on the server.

This way, we would be able to create a `TwilioSerializer`, which would send messages in the same format as Twilio does.

filipi87: Done.

examples/twilio-chatbot/ws_test_client/package.json
```diff
@@ -0,0 +1,25 @@
...
+  "devDependencies": {
+    "@types/node": "^22.13.1",
+    "@types/protobufjs": "^6.0.0",
+    "@vitejs/plugin-react-swc": "^3.7.2",
+    "typescript": "^5.7.3",
+    "vite": "^6.0.2"
+  },
+  "dependencies": {
+    "@pipecat-ai/client-js": "^0.4.0",
+    "@pipecat-ai/websocket-transport": "^0.4.1"
```
filipi87: We need to update this version once we merge this PR:
https://github.com/pipecat-ai/pipecat-client-web-transports/pull/46

filipi87:
This is now ready for review and working. Need to use the `WebSocketTransport` provided by this PR: 
https://github.com/pipecat-ai/pipecat-client-web-transports/pull/46

examples/twilio-chatbot/ws_test_client/package.json
filipi87: Done.

markbackman:
@filipi87 should this new client replace the existing client? I think it would be cleaner than a full client, but maybe it's worth getting @aconchillo's opinion on that since he wrote the example.

I approve, but before I give my official ✅ let's have a chat about which of the clients we want this to use.

examples/twilio-chatbot/ws_test_client/README.md
```diff
@@ -0,0 +1,27 @@
+# JavaScript Implementation
+
+Basic implementation using the [Pipecat JavaScript SDK](https://docs.pipecat.ai/client/js/introduction).
+
+## Setup
+
+1. Run the bot server. See the [server README](../README).
+
+2. Navigate to the `client/javascript` directory:
```
markbackman: This README is already in the ws_test_client directory, so no navigation is required.

If this ends up replacing the existing client, then maybe these readme steps should go in the main twilio-chatbot readme.

filipi87: Fixed

filipi87:
@markbackman , I have restructured the examples, let me know what do you think. 🙂

markbackman (approved):
Looks great!

PR #1926, opened 2025-05-30
handle StopFrame in base input transport and stop pushing frames

aconchillo:

We now handle `StopFrame` in the input transport. We stop pushing frames downstream until we get a new `StartFrame`. This will avoid some warning messages about `StartFrame` not being received if we plug the transport into a new pipeline.


DominicStewart (approved):
Tested this, worked well! LGTM!

filipi87 (approved):
Makes sense. Nice improvement. 🚀

aconchillo:
I've update this PR to allow all other transport to do the same.

PR #1931, opened 2025-05-30
Add support for interruption configuration

markbackman:


src/pipecat/transports/base_input.py
```diff
@@ -230,20 +246,45 @@ async def _handle_user_interruption(self, frame: Frame):
...
+    # Handle bot speaking state
+    #
+
+    async def _handle_bot_started_speaking(self, frame: BotStartedSpeakingFrame):
+        self._bot_speaking = True
+        await self.push_frame(frame)
+
+    async def _handle_bot_stopped_speaking(self, frame: BotStoppedSpeakingFrame):
+        self._bot_speaking = False
+        await self.push_frame(frame)
```
aconchillo, line 286: I usually keep the `push_frame` in `process_frame`. This way they are all in one place. So the handler only do specific things to the handler, but they don't forward the original frame.

src/pipecat/processors/aggregators/llm_response.py
```diff
@@ -266,6 +268,8 @@ def __init__(
         self._seen_interim_results = False
         self._waiting_for_aggregation = False
 
+        self._interruption_config: Optional[InterruptionConfig] = None
```
aconchillo, line 271: This would go in `FrameProcessor`

src/pipecat/frames/frames.py
```diff
@@ -438,6 +438,20 @@ class OutputDTMFFrame(DTMFFrame, DataFrame):
...
+    """Configuration for interruption behavior.
+
+    When specified, the bot will not be interrupted immediately when the user speaks.
+    Instead, interruption will only occur when the configured conditions are met.
+
+    Args:
+        min_words: If set, user must speak at least this many words to interrupt
+    """
+
+    min_words: Optional[int] = None
```
aconchillo, line 452: Maybe we could have an empty base class `InterruptionStrategy` and then subclasses, for example:

```
@dataclass
class InterruptionStrategy:
    pass

@dataclass
class NumWordsInterruptionStrategy(InterruptionStrategy):
    """Strategy for interruption behavior.
    When specified, the bot will not be interrupted immediately when the user speaks.
    Instead, interruption will only occur when the configured conditions are met.
    Args:
        min_words: If set, user must speak at least this many words to interrupt
    """

    min_words: Optional[int] = None
```

we would then use `isinstance(config, NumWordsInterruptionStrategy`. The Strategy indicates there's going to be a different behavior depending on the class you choose.

src/pipecat/transports/base_input.py
```diff
@@ -128,6 +137,9 @@ async def start(self, frame: StartFrame):
 
         self._sample_rate = self._params.audio_in_sample_rate or frame.audio_in_sample_rate
 
+        # Store interruption configuration
+        self._interruption_config = frame.interruption_config
```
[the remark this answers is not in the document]
markbackman: Ok, I added a `@property` to the FrameProcessor to access interruption_config and then use that in the BaseInputTransport and LLMUserContextAggregator. Let me know if this is what you're thinking. It's in the "Code review feedback" commit.

src/pipecat/transports/base_input.py
```diff
@@ -230,20 +246,45 @@ async def _handle_user_interruption(self, frame: Frame):
...
             # Make sure we notify about interruptions quickly out-of-band.
-            if self.interruptions_allowed:
+            if should_push_immediate_interruption and self.interruptions_allowed:
                 await self._start_interruption()
                 # Push an out-of-band frame (i.e. not using the ordered push
                 # frame task) to stop everything, specially at the output
                 # transport.
                 await self.push_frame(StartInterruptionFrame())
+            elif self._interruption_config and self._bot_speaking:
+                logger.trace(
```
aconchillo, line 261: Maybe `debug` since we might want to know what was going on when they send us logs.

src/pipecat/transports/base_input.py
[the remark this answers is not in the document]
aconchillo: Yes, that was it.

CHANGELOG.md
```diff
@@ -9,6 +9,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- Added `interruption_strategies` to `PipelineParams` which uses an
+  `InterruptionStrategy` to specify criteria required to interrupt the bot when
+  it's speaking. You can specify `min_words` to require the user to say at
```
aconchillo, line 14: Mention there's `MinWordsInterruptionStrategy`.

PR #1934, opened 2025-05-30
Fix type checker error with missing function call in the small WebRTC transport example

counterleft:
My local pyright wasn't happy with a call to `close()` in the small WebRTC transport example.

SmallWebRTCConnection doesn't have a `close()`. There's a `_close()` but I assume that's private due to its naming. The closest function that uses `_close()` is `disconnect()`. I assume then, that the intended resource freeing function call should be to `disconnect()`.


counterleft:
👋 Ah, yeah you're right, Aleix. A search for "pc.close()" shows a few more instances. I'll fix those real quick.

PR #1936, opened 2025-06-01
Add daily to the foundational examples requirements.txt

markbackman:

With the changed import, the quickstart is broken. As a quick fix, we can just suggest to download the `run.py` from the new location and run the local import.

Also, in testing the quickstart, I noticed that we now require the `daily` optional dependencies to be installed.

### DIFF
--- a/examples/foundational/requirements.txt
+++ b/examples/foundational/requirements.txt
@@ -1,5 +1,5 @@
 fastapi
 uvicorn
 python-dotenv
-pipecat-ai[webrtc,deepgram,cartesia]
+pipecat-ai[webrtc,daily,deepgram,cartesia]
 pipecat-ai-small-webrtc-prebuilt


markbackman:
@aconchillo we're also getting feedback like this: https://discord.com/channels/1239284677165056021/1378643467189751898/1378643467189751898. Maybe we need to make this easier in general?

examples/foundational/01-say-one-thing.py
```diff
@@ -53,6 +53,11 @@ async def on_client_connected(transport, client):
 
 
 if __name__ == "__main__":
-    from pipecat.examples.run import main
+    try:
+        # Try the installed package import first
+        from pipecat.examples.run import main
+    except ImportError:
+        # Fall back to local import
+        from run import main
```
aconchillo: I don't think we need this. This will be solved when we release tomorrow. If you try examples from main you need to use main. If you try to run examples with an older version of pipecat then you need to run that examples version.

examples/foundational/requirements.txt
```diff
@@ -1,5 +1,5 @@
 fastapi
 uvicorn
 python-dotenv
-pipecat-ai[webrtc,deepgram,cartesia]
+pipecat-ai[webrtc,daily,deepgram,cartesia]
```
aconchillo, line 4: This one is fine.

markbackman:
The issue is the quickstart: https://docs.pipecat.ai/getting-started/quickstart. It helps you set up a project without downloading main. Maybe we should change that approach to downloading the repo first? WDYT?

aconchillo:
> The issue is the quickstart: https://docs.pipecat.ai/getting-started/quickstart. It helps you set up a project without downloading main. Maybe we should change that approach to downloading the repo first? WDYT?

That or just remove downloading `run.py`. When people run `pip install -r requirements.txt` that will install pipecat and therefore the new examples packages. So, starting tomorrow we should be good (removing that one curl command)

markbackman:
Sounds good. Updating this PR to just change the requirements.txt.

aconchillo:
LGTM

vintageplayer:
Can you also check the daily depedency? It was not found on mac os.
pip 25.0.1
Python 3.13.3

markbackman:
@vintageplayer it's defined and installs for me without an issue: https://github.com/pipecat-ai/pipecat/blob/main/pyproject.toml#L51

PR #1938, opened 2025-06-01
allow custom interruption strategies

aconchillo:

This PR allows users to write their own custom interruptions strategies.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,12 +26,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Added OpenTelemetry tracing for `GeminiMultimodalLiveLLMService` and
   `OpenAIRealtimeBetaLLMService`.
 
-- Added `interruption_strategies` to `PipelineParams` using
-  `MinWordsInterruptionStrategy` to specify minimum words required to interrupt
-  the bot when it's speaking. Use
-  `interruption_strategies=[MinWordsInterruptionStrategy(min_words=N)]` to
-  require users to speak at least N words before interrupting. If not
-  specified, the normal interruption behavior applies.
+- Added initial support for interruption strategies, which determine if the user
+  should interrupt the bot while the bot is speaking. Interruption strategies
+  can be based on factors such as audio volume or the number of words spoken by
+  the user. These can be specified via the new `interruption_strategies` field
+  in `PipelineParams`. A new `MinWordsInterruptionStrategy` strategy has been
+  introduced which triggers an interruption if the user has spoken a minimum
+  number of words. If no interruption strategies are specified, the normal
+  interruption behavior applies. If multiple strategies are provided, the first
+  one that evaluates to true will trigger the interruption.
 
 - `BaseInputTransport` now handles `StopFrame`. When a `StopFrame` is received
   the transport will pause sending frames downstream until a new `StartFrame` is
--- a/src/pipecat/audio/interruptions/base_interruption_strategy.py
+++ b/src/pipecat/audio/interruptions/base_interruption_strategy.py
@@ -1,0 +1,38 @@
+#
+# Copyright (c) 2024–2025, Daily
+#
+# SPDX-License-Identifier: BSD 2-Clause License
+#
+
+from abc import ABC, abstractmethod
+
+
+class BaseInterruptionStrategy(ABC):
+    """This is a base class for interruption strategies. Interruption strategies
+    decide when the user can interrupt the bot while the bot is speaking. For
+    example, there could be strategies based on audio volume or strategies based
+    on the number of words the user spoke.
+
+    """
+
+    async def append_audio(self, audio: bytes, sample_rate: int):
+        """Appends audio to the strategy. Not all strategies handle audio."""
+        pass
+
+    async def append_text(self, text: str):
+        """Appends text to the strategy. Not all strategies handle text."""
+        pass
+
+    @abstractmethod
+    async def should_interrupt(self) -> bool:
+        """This is called when the user stops speaking and it's time to decide
+        whether the user should interrupt the bot. The decision will be based on
+        the aggregated audio and/or text.
+
+        """
+        pass
+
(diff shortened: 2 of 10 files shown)


CHANGELOG.md
```diff
@@ -26,12 +26,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
-  specified, the normal interruption behavior applies.
+- Added initial support for interruption strategies, which determine if the user
+  should interrupt the bot while the bot is speaking. Interruption strategies
+  can be based on factors such as audio volume or the number of words spoken by
+  the user. These can be specified via the new `interruption_strategies` field
+  in `PipelineParams`. A new `MinWordsInterruptionStrategy` strategy has been
+  introduced which triggers an interruption if the user has spoken a minimum
+  number of words. If no interruption strategies are specified, the normal
+  interruption behavior applies. If multiple strategies are provided, all of
+  them must evaluate to true for the interruption to occur.
```
filipi87: Maybe as a future improvement: if the user provides multiple strategies, allow them to choose between "all" or `any`.

src/pipecat/processors/aggregators/llm_response.py
```diff
@@ -351,24 +355,18 @@ async def push_aggregation(self):
...
-        # Check strategies one by one until first match
         for strategy in self.interruption_strategies:
-            if isinstance(strategy, MinWordsInterruptionStrategy):
-                if self._should_interrupt_min_words(strategy):
-                    return True
+            await strategy.append_text(self._aggregation)
+            result = await strategy.should_interrupt()
+            await strategy.reset()
+            if result:
+                return True
```
filipi87: In the changelog you mentioned:
- "If multiple strategies are provided, all of them must evaluate to true for the interruption to occur."

If I'm not mistaken, in this case, we're interrupting if any of them are true , right ? I don’t think that’s wrong. But if we choose to keep it like this maybe just fix the changelog.

filipi87: Also, just double checking, is it correct always invoke `await strategy.reset()` outside the `if result` ? Even if we are handling with a strategy that it is manipulating audio instead ?

aconchillo: > In the changelog you mentioned:
> 
> * "If multiple strategies are provided, all of them must evaluate to true for the interruption to occur."
> 
> If I'm not mistaken, in this case, we're interrupting if any of them are true , right ? I don’t think that’s wrong. But if we choose to keep it like this maybe just fix the changelog.

Oops. That's wrong. I changed it to the old behavior (any) and forgot to update the CHANGELOG.

aconchillo: > Also, just double checking, is it correct always invoke `await strategy.reset()` outside the `if result` ? Even if we are handling with a strategy that it is manipulating audio instead ?

I just removed `reset()`. Things get reset when you call `should_interrupt()`.

CHANGELOG.md
aconchillo: I think people could do that by wrapping multiple strategies into one. We could maybe add a `AllInterruptionStrategy([strategy1, strategy2...])` in the future.

src/pipecat/audio/interruptions/base_interruption_strategy.py
```diff
@@ -0,0 +1,33 @@
...
+class BaseInterruptionStrategy(ABC):
+    """This is a base class for interruption strategies. Interruption strategies
+    decide when the user can interrupt the bot while the bot is speaking. For
+    example, there could be strategies based on audio volume or strategies based
+    on the number of words the user spoke.
+
+    """
+
+    async def append_audio(self, audio: bytes, sample_rate: int):
+        """Appends audio to the strategry. Not all strategies handle audio."""
```
markbackman:
```suggestion
        """Appends audio to the strategy. Not all strategies handle audio."""
```

src/pipecat/audio/interruptions/base_interruption_strategy.py
```diff
@@ -0,0 +1,33 @@
...
+
+class BaseInterruptionStrategy(ABC):
+    """This is a base class for interruption strategies. Interruption strategies
+    decide when the user can interrupt the bot while the bot is speaking. For
+    example, there could be strategies based on audio volume or strategies based
+    on the number of words the user spoke.
+
+    """
+
+    async def append_audio(self, audio: bytes, sample_rate: int):
```
markbackman, line 18: Are you adding this in the event users want to write their own interruption strategy with audio?

CHANGELOG.md
markbackman: I had thought about the same thing @aconchillo but opted to not add right now since we only have one strategy.

src/pipecat/audio/interruptions/base_interruption_strategy.py
aconchillo: Yep!

filipi87 (approved):
I was wondering if we also need a method to handle `cleanup`. For instance, if we have multiple strategies and one of them matches with `interrupt == true`, how do we notify the other strategies to clean up their data, for example, removing audio or text from the cache?

Other than that, everything looks good to me.

aconchillo:
> I was wondering if we also need a method to handle `cleanup`. For instance, if we have multiple strategies and one of them matches with `interrupt == true`, how do we notify the other strategies to clean up their data, for example, removing audio or text from the cache?
> 

They all cleanup on `should_interrupt()`.

> Other than that, everything looks good to me.

aconchillo:
> > I was wondering if we also need a method to handle `cleanup`. For instance, if we have multiple strategies and one of them matches with `interrupt == true`, how do we notify the other strategies to clean up their data, for example, removing audio or text from the cache?
> 
> They all cleanup on `should_interrupt()`.
> 

Ah, you mean the following ones because we short circuit. OK, I have added `reset()` back which we call for all strategies now.

filipi87:
All looks good. 🚀

PR #1944, opened 2025-06-02
Adding the direction when pushing the frame.

filipi87:
Adding the direction when pushing the BotStartedSpeaking and BotStoppedSpeaking frames.


aconchillo:
LGTM!

PR #1949, opened 2025-06-03
Fixed transport destination issue

filipi87:
Fixed an issue where OutputAudioRawFrame.transport_destination was being reset to None instead of retaining its intended value before sending the audio frame to the transports.

### DIFF
--- a/src/pipecat/services/google/llm.py
+++ b/src/pipecat/services/google/llm.py
@@ -83,7 +83,7 @@ class GoogleUserContextAggregator(OpenAIUserContextAggregator):
             await self.push_frame(frame)
 
             # Reset our accumulator state.
-            self.reset()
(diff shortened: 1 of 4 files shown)


aconchillo (approved):

aconchillo:
LGTM!

src/pipecat/services/google/llm.py
```diff
@@ -83,7 +83,7 @@ async def push_aggregation(self):
             await self.push_frame(frame)
 
             # Reset our accumulator state.
-            self.reset()
+            await self.reset()
```
aconchillo, line 86: There's another `reset()` in user_response.py that needs fixing as well.

filipi87: done.

PR #1952, opened 2025-06-03
feat: Enhance GladiaSTTService with reconnection and audio buffer management features

jqueguiner:
For Gladia STT provider:
- Added parameters for maximum reconnection attempts, reconnection delay, and maximum audio buffer size.
- Implemented automatic reconnection logic with exponential backoff.
- Introduced audio buffer management to handle audio data efficiently, including trimming excess data.
- Updated connection handling to ensure proper cleanup and management of WebSocket connections.
- Enhanced audio sending logic to support buffered audio transmission after reconnections.

### DIFF
--- a/src/pipecat/services/gladia/stt.py
+++ b/src/pipecat/services/gladia/stt.py
@@ -195,6 +195,9 @@ class GladiaSTTService(STTService):
         sample_rate: Optional[int] = None,
         model: str = "solaria-1",
         params: Optional[GladiaInputParams] = None,
+        max_reconnection_attempts: int = 5,
+        reconnection_delay: float = 1.0,
+        max_buffer_size: int = 1024 * 1024 * 5,  # 5MB default buffer
         **kwargs,
     ):
         """Initialize the Gladia STT service.
@@ -207,6 +210,9 @@ class GladiaSTTService(STTService):
             model: Model to use ("solaria-1", "solaria-mini-1", "fast",
                 or "accurate")
             params: Additional configuration parameters
+            max_reconnection_attempts: Maximum number of reconnection attempts
+            reconnection_delay: Initial delay between reconnection attempts (exponential backoff)
+            max_buffer_size: Maximum size of audio buffer in bytes
             **kwargs: Additional arguments passed to the STTService
         """
         super().__init__(sample_rate=sample_rate, **kwargs)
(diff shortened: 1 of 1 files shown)


src/pipecat/services/gladia/stt.py
```diff
@@ -293,36 +315,149 @@ def _prepare_settings(self) -> Dict[str, Any]:
...
+            try:
+                # Initialize session if needed
+                if not self._session_url:
+                    settings = self._prepare_settings()
+                    response = await self._setup_gladia(settings)
+                    self._session_url = response["url"]
+                    self._reconnection_attempts = 0
+
+                # Connect with automatic reconnection
+                async for websocket in websockets.connect(self._session_url):
```
filipi87: I believe we should do it like this to correctly manage the WebSocket connection's lifecycle.

```suggestion
                async with websockets.connect(self._session_url) as websocket:
```

src/pipecat/services/gladia/stt.py
```diff
@@ -293,36 +315,149 @@ def _prepare_settings(self) -> Dict[str, Any]:
...
+                                f"Max reconnection attempts ({self._max_reconnection_attempts}) reached"
+                            )
+                            self._should_reconnect = False
+                            break
+
+                        delay = self._reconnection_delay * (2 ** (self._reconnection_attempts - 1))
+                        logger.info(
+                            f"Reconnecting in {delay} seconds (attempt {self._reconnection_attempts}/{self._max_reconnection_attempts})"
+                        )
+                        await asyncio.sleep(delay)
```
filipi87: We are repeating the reconnection logic inside the exception handler. It might be better to extract it into a helper function.

Or, we could make GladiaSTTService extend WebsocketService and rely on that for the reconnection logic. =)

src/pipecat/services/gladia/stt.py
```diff
@@ -293,36 +315,149 @@ def _prepare_settings(self) -> Dict[str, Any]:
...
+            # Trim buffer if it exceeds max size
+            if len(self._audio_buffer) > self._max_buffer_size:
+                trim_size = len(self._audio_buffer) - self._max_buffer_size
+                self._audio_buffer = self._audio_buffer[trim_size:]
+                self._bytes_sent = max(0, self._bytes_sent - trim_size)
+                logger.warning(f"Audio buffer exceeded max size, trimmed {trim_size} bytes")
+
+        # Send audio if connected
+        if self._connection_active and self._websocket and not self._websocket.closed:
+            await self._send_audio(audio)
```
filipi87: The socket could close between the check if it closed and when we are actually sending it. Maybe it would be better wrap self._send_audio(audio) in a try-except block

src/pipecat/services/gladia/stt.py
```diff
@@ -293,36 +315,149 @@ def _prepare_settings(self) -> Dict[str, Any]:
...
+                        self._websocket = websocket
+                        self._connection_active = True
+                        logger.info("Connected to Gladia WebSocket")
+
+                        # Send buffered audio if any
+                        await self._send_buffered_audio()
+
+                        # Start tasks
+                        receive_task = asyncio.create_task(self._receive_task_handler())
+                        keepalive_task = asyncio.create_task(self._keepalive_task_handler())
```
filipi87: I think that where we're creating and using `receive_task` and `keepalive_task` inside `_connection_handler`, we should actually assign them to `self._receive_task` and `self._keepalive_task` ?

Otherwise, `self._receive_task` and `self._keepalive_task` are defined but never assigned.

PR #1956, opened 2025-06-03
Make `PipelineTask.add_observer()` synchronous. This allows callers t…

kompfner:
…o call it before `run()`ning the `PipelineTask` first. Without this change, if they tried to do that, they would get an error because the `TaskManager`'s event loop hadn't been set yet.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- Make `PipelineTask.add_observer()` synchronous. This allows callers to call it before doing the
+  work of running the `PipelineTask` (i.e. without invoking `PipelineTask.set_event_loop()` first).
+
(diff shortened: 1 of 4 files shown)


kompfner:
I had originally needed the change in this PR for https://github.com/pipecat-ai/pipecat-flows/pull/142, but then I changed my approach in that PR. I still think the change in this PR is a good idea, though, so I'm leaving the PR open.

kompfner:
This PR might need more testing. I only validated that it worked with an earlier iteration https://github.com/pipecat-ai/pipecat-flows/pull/142.

aconchillo (approved):

aconchillo:
LGTM! Can you please add a CHANGELOG entry (Changed section I guess)?

kompfner:
Before I merge this...are we concerned that this is a breaking change to the API? (That fact just dawned on me)

aconchillo:
> Before I merge this...are we concerned that this is a breaking change to the API? (That fact just dawned on me)

This was a pretty new feature. Ideally we don't want to add breaking changes, but in this cases I believe it's fine.

PR #1960, opened 2025-06-05
disable uvloop by default and just let the user set it

aconchillo:

I've got a report that uvloop is causing issues in a system running Django, so I think it's safer to just not set it by default and let the user decide what to do.


filipi87 (approved):
LGTM!

PR #1967, opened 2025-06-05
Remove instantiation of unused http client session from certain examples

counterleft:
These examples don't make any HTTP requests with `session` so there
doesn't seem be a need to create one in the first place. Probably a
copy-paste from a previous example.


counterleft:
I was playing with the `text-webrtc` example and noticed the unused `session`. Found a few more instances of it (`livekit`, `text-and-audio-webrtc`) and removed those as well.

aconchillo (approved):

aconchillo:
LGTM! Thank you!

PR #1968, opened 2025-06-06
Add FR and DE to Rime TTS language map

JasonNero:
This PR adds `fr/fra` and `de/ger` to the supported languages for Rime TTS.

For reference, all available language-voice combinations can be found here: https://users.rime.ai/data/voices/all-v2.json


markbackman:
Thanks! Someone filed an issue for this. I realized the HTTP service was missing lang support so I add the additional languages and the HTTP lang support here: https://github.com/pipecat-ai/pipecat/pull/1978.

PR #1979, opened 2025-06-09
buffer audio from TTS service before pushing frames

aconchillo:

There's an issue with TTS services that do chunking. The issue is that we start playing (pushing audio frames) too early. What happens is that we start playing audio with the first chunk but the next bytes from the network might come a bit later (because maybe network issues) and we already finished playing those initial bytes causing audio glitches.

The solution is to download a few seconds of audio which is fast and then start playing. This will ensure that we have enough buffer (those few seconds) to download more data while the audio keeps playing.

### DIFF
--- a/src/pipecat/services/aws/tts.py
+++ b/src/pipecat/services/aws/tts.py
@@ -253,7 +253,8 @@ class AWSPollyTTSService(TTSService):
 
             yield TTSStartedFrame()
 
-            CHUNK_SIZE = 1024
+            CHUNK_SIZE = self.chunk_size
+
             for i in range(0, len(audio_data), CHUNK_SIZE):
                 chunk = audio_data[i : i + CHUNK_SIZE]
                 if len(chunk) > 0:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,9 @@ asyncio.set_event_loop_policy(uvloop.EventLoopPolicy())
 
 ### Fixed
 
+- Fixed an issue with various TTS services that would cause audio glitches at
+  the start of every bot turn.
+
 - Fixed an `AssemblyAISTTService` issue that could cause unexpected behavior
   when yielding empty `Frame()`s.
 
--- a/src/pipecat/services/google/tts.py
+++ b/src/pipecat/services/google/tts.py
@@ -362,8 +362,8 @@ class GoogleHttpTTSService(TTSService):
             # Skip the first 44 bytes to remove the WAV header
             audio_content = response.audio_content[44:]
 
-            # Read and yield audio data in chunks
-            CHUNK_SIZE = 1024
+            CHUNK_SIZE = self.chunk_size
+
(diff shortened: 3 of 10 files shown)


markbackman:
I played around with this and it certainly fixes the audio choppiness, but it adds noticeable latency. I think we need to find the balance where enough audio is received to avoid choppy audio, but keep the latency low. In just experimenting, 8kb isn't enough, but somewhere between 12-16kb seems to be sufficient.

Is there a reason why you chose a large number? Was it to just be safe, ensuring the audio is ready to playback?

kwindla (approved):
lgtm

src/pipecat/services/aws/tts.py
```diff
@@ -253,7 +253,14 @@ def read_audio_data(**args):
...
             yield TTSStartedFrame()
 
-            CHUNK_SIZE = 1024
+            # We download a few seconds of audio before we start pushing the
+            # first audio frame. This will make sure we download the rest of the
+            # audio while audio is being played without causing audio glitches
+            # (specially at the beginning). Of course, this will also depend on
+            # how fast the TTS service generates bytes.
+            CHUNK_SECONDS = 1
+            CHUNK_SIZE = self.sample_rate * CHUNK_SECONDS * 2  # 2 bytes/sample
```
filipi87: nit: Inside the `BaseOutputTransport`, we created a property called `audio_chunk_size`. Wouldn't it be worth doing the same here for the `TTSService`, just to avoid duplicating this code across all tts services ?

aconchillo: Yes, I'll do that!

aconchillo:
> I played around with this and it certainly fixes the audio choppiness, but it adds noticeable latency. I think we need to find the balance where enough audio is received to avoid choppy audio, but keep the latency low. In just experimenting, 8kb isn't enough, but somewhere between 12-16kb seems to be sufficient.
> 

That would be half second for 16000 sample rate. I'll try that.

> Is there a reason why you chose a large number? Was it to just be safe, ensuring the audio is ready to playback?

Yes, basically to be safe we have enough buffer.

src/pipecat/services/aws/tts.py
aconchillo: Updated. Please, take another look.

aconchillo:
> > I played around with this and it certainly fixes the audio choppiness, but it adds noticeable latency. I think we need to find the balance where enough audio is received to avoid choppy audio, but keep the latency low. In just experimenting, 8kb isn't enough, but somewhere between 12-16kb seems to be sufficient.
> 
> That would be half second for 16000 sample rate. I'll try that.
> 
> > Is there a reason why you chose a large number? Was it to just be safe, ensuring the audio is ready to playback?
> 
> Yes, basically to be safe we have enough buffer.

Half a second also seems to work fine. I'll leave that.

filipi87 (approved):
Nice improvement. 🚀🔥

PR #1982, opened 2025-06-09
[Add] Support for Cartesia AI STT

shreygupta2809:
Adds a new CartesiaSTTService, which is a websocket based implementation to transcribe audio.

Adds a foundational example in `13f-cartesia-transcription.py`


markbackman:
I'm not aware that Cartesia offers STT services. Can you link to the docs?

src/pipecat/services/cartesia/stt.py
```diff
@@ -0,0 +1,232 @@
...
+    EndFrame,
+    Frame,
+    InterimTranscriptionFrame,
+    StartFrame,
+    TranscriptionFrame,
+    UserStartedSpeakingFrame,
+    UserStoppedSpeakingFrame,
+)
+from pipecat.processors.frame_processor import FrameDirection
+from pipecat.services.ai_services import STTService
```
aconchillo: This should be 

```
from pipecat.services.stt_service import STTService
```

to avoid the deprecation warning.

src/pipecat/services/cartesia/stt.py
```diff
@@ -0,0 +1,232 @@
...
+            else:
+                # For interim transcriptions, just push the frame without tracing
+                await self.push_frame(
+                    InterimTranscriptionFrame(transcript, "", time_now_iso8601(), language)
+                )
+
+    async def _disconnect(self):
+        if self._receiver_task:
+            self._receiver_task.cancel()
+            self._receiver_task = None
```
filipi87: Should we await for the task to cancel ? Something like this.
```
if self._receiver_task:
    self._receiver_task.cancel()
    try:
        await self._receiver_task
        pass
    except Exception as e:
        logger.exception(f"Unexpected exception while cancelling task: {e}")
    self._receiver_task = None
```

shreygupta2809: Added

PR #1984, opened 2025-06-09
CartesiaSTTService cleanup

markbackman:

One other observation: this service does not abide by the sample rate being set by the PipelineParams. In other services, we initialize to an Optional[int] of None and then let the StartFrame define the sample_rate. Should we not do the same here? The one caveat is that CartesiaSTTService supports 16khz only.

cc @aconchillo @filipi87

### DIFF
--- a/src/pipecat/services/cartesia/stt.py
+++ b/src/pipecat/services/cartesia/stt.py
@@ -73,7 +73,7 @@ class CartesiaSTTService(STTService):
         self,
         *,
         api_key: str,
-        base_url: str = None,
+        base_url: str = "",
         sample_rate: int = 16000,
         live_options: Optional[CartesiaLiveOptions] = None,
         **kwargs,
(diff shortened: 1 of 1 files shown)


aconchillo:
> One other observation: this service does not abide by the sample rate being set by the PipelineParams. In other services, we initialize to an Optional[int] of None and then let the StartFrame define the sample_rate. Should we not do the same here? The one caveat is that CartesiaSTTService supports 16khz only.
> 
> cc @aconchillo @filipi87

Yes, we should probably follow the same pattern. I think we need to consider adding resampling (and a warning) if incoming audio is not supported.

markbackman:
@aconchillo or @filipi87 want to push changes to this PR?

markbackman:
> Yes, we should probably follow the same pattern. I think we need to consider adding resampling (and a warning) if incoming audio is not supported.

Agreed. Since this will ship today, let's just add these two small changes. We can fix later in a separate PR.

PR #1992, opened 2025-06-10
Add `additional_span_attributes` param to PipelineTask for extra otel…

sunahsuh:
… attributes

This adds a parameter to PipelineTask that lets you add any additional attributes you'd like to the conversation span (for Pipecat Cloud, we need the ability to tag by session ID, but other folks might want to add a user ID, etc.)


markbackman (approved):
LGTM

PR #1996, opened 2025-06-11
fix: Add silence to the user recording buffer when using the mute filter

nickmarq:
When using the AudioBufferProcessor with the mute filter there is an overlapping user and bot audio.
This is cause by the mute filter blocking the `InputAudioRawFrame`.

To fix it, we need to add some silence to the user buffer when receiving a `OutputAudioRawFrame` and the user buffer is smaller.

Before:

Screenshot 2025-06-11 at 10 33 33 AM

After:

Screenshot 2025-06-11 at 10 33 13 AM

(Not the same recording, btw)


nickmarq:
That should also fix this [PR](https://github.com/pipecat-ai/pipecat/pull/1325)

aconchillo:
I believe this was fixed in this PR https://github.com/pipecat-ai/pipecat/pull/2024. Can you reproduce your issue with Pipecat >= 0.0.72?

PR #1998, opened 2025-06-11
Fix: 38-smart-turn-fal by lazy loading the aiohttp session

markbackman:

@aconchillo I'm open to other approaches. I was looking for something that works within the current structure.

### DIFF
--- a/examples/foundational/38-smart-turn-fal.py
+++ b/examples/foundational/38-smart-turn-fal.py
@@ -27,7 +27,23 @@ from pipecat.transports.services.daily import DailyParams
 
 load_dotenv(override=True)
 
-aiohttp_session = aiohttp.ClientSession()
+
+# Lazy session that creates the actual session when first used
+class LazySession:
+    def __init__(self):
+        self._session = None
+
+    def __getattr__(self, name):
+        if self._session is None:
+            self._session = aiohttp.ClientSession()
+        return getattr(self._session, name)
+
+    async def close(self):
+        if self._session:
+            await self._session.close()
+
+
+aiohttp_session = LazySession()
 
 # We store functions so objects (e.g. SileroVADAnalyzer) don't get
 # instantiated. The function will be called when the desired transport gets


examples/foundational/38-smart-turn-fal.py
```diff
@@ -27,7 +27,23 @@
...
+        if self._session is None:
+            self._session = aiohttp.ClientSession()
+        return getattr(self._session, name)
+
+    async def close(self):
+        if self._session:
+            await self._session.close()
+
+
+aiohttp_session = LazySession()
```
aconchillo: I think we should just call `aiohttp.ClientSession()` inside `transport_params` and keep things simple.

aconchillo: Ah wait, that would be the same we have today... One sec.

markbackman: I tried this first, but the `transport_params` lambdas are called before any event loop is running. Since 
`aiohttp.ClientSession()` requires a running event loop, calling it in the `transport_params` lambda throws:

```
RuntimeError: no running event loop
```

For reference, in run.py:
```
    params = transport_params[args.transport]()
    match args.transport:
        case "daily":
            run_example_daily(run_example, args, params)
        case "webrtc":
            run_example_webrtc(run_example, args, params)
        case "twilio":
            run_example_twilio(run_example, args, params)
```

Any alternatives?

aconchillo: OK. Here's a PR so we can create the session in `transport_params`: https://github.com/pipecat-ai/pipecat/pull/2016

The problem was that we were creating the params outside of an async function. We now delay the creation of the params inside async functions.

aconchillo: We will need to remove the `.close()` at the end, but it's fine.

markbackman: Yup, that's totally fine. I approved that change.

markbackman: I made the update and will rebase after #2016 is merged.

markbackman: Ok, rebased and ready for review.

filipi87 (approved):
LGTM!

PR #1999, opened 2025-06-11
RTVI 1.0.0 Updates 🎉 

mattieruth:
This PR introduces the server-side rtvi updates to match the new RTVI 1.0.0 protocol. The primary protocol changes are:

1. client-ready/bot-ready messages now both include a `version` and `about` section
2. action-related messages have been removed (deprecated) in lieu of client-server messages and some built-in types
3. service configuration message have been removed (deprecated) (security concerns. should be replaced with custom client-server messages)
4. new `client-message` and `server-response` messages for custom messaging
5. new `append-to-context` message

In support of the new messages, this PR introduces two new frames: `RTVIClientMessageFrame` and `RTVIServerResponseFrame`. FrameProcessors can listen for client messages via the `RTVIClientMessageFrame` and respond to those messages by pushing the `RTVIServerResponseFrame`. Alternately, you listen/respond using the new rtvi event handler: `@rtvi.event_handler("on_client_message")` and responding directly using `rtvi.send_server_response` or `rtvi.send_error_response`.

For the new `append-to-context` message, the rtvi frame processor now optionally takes a context_aggregator as part of its constructor to automatically handle and push the context updates.

With the new versioning information from the client, we log the client details and minimally use the client version to modify our message handling. This needs to be expanded such that deprecated messages are ignored for newer client versions.

In addition to the rtvi updates, this PR updates all web client examples to use the new 1.0 Pipecat Client.

THIS PR SHOULD NOT BE MERGED UNTIL (AND SHORTLY AFTER) THE Client PRs have been merged and released.

Corresponding PRS:
- [pipecat-client-web](https://github.com/pipecat-ai/pipecat-client-web/pull/117)
- [pipecat-client-web-transports](https://github.com/pipecat-ai/pipecat-client-web-transports/pull/47)
- [docs](https://github.com/pipecat-ai/docs/pull/265)

### DIFF
--- a/examples/simple-chatbot/server/bot-openai.py
+++ b/examples/simple-chatbot/server/bot-openai.py
@@ -273,7 +273,7 @@ async def main():
         context = OpenAILLMContext(messages, tools)
         context_aggregator = llm.create_context_aggregator(context)
 
-        rtvi = RTVIProcessor(config=RTVIConfig(config=[]), context_aggregator=context_aggregator)
+        rtvi = RTVIProcessor(config=RTVIConfig(config=[]))
 
         ta = TalkingAnimation()
 
--- a/src/pipecat/processors/frameworks/rtvi.py
+++ b/src/pipecat/processors/frameworks/rtvi.py
@@ -1122,7 +1122,6 @@ class RTVIProcessor(FrameProcessor):
         *,
         config: Optional[RTVIConfig] = None,
         transport: Optional[BaseTransport] = None,
-        context_aggregator: Optional[OpenAIContextAggregatorPair] = None,
         **kwargs,
     ):
         """Initialize the RTVI processor.
@@ -1130,8 +1129,6 @@ class RTVIProcessor(FrameProcessor):
         Args:
             config: Initial RTVI configuration.
             transport: Transport layer for communication.
-            context_aggregator: Context aggregator for OpenAI LLMs.
-                Needed to support the append-to-context message.
             **kwargs: Additional arguments passed to parent class.
         """
         super().__init__(**kwargs)
@@ -1163,7 +1160,6 @@ class RTVIProcessor(FrameProcessor):
             if isinstance(input_transport, BaseInputTransport):
                 self._input_transport = input_transport
                 self._input_transport.enable_audio_in_stream_on_start(False)
-        self._context_aggregator = context_aggregator
 
     def register_action(self, action: RTVIAction):
         """Register an action that can be executed via RTVI.
@@ -1584,22 +1580,19 @@ class RTVIProcessor(FrameProcessor):
         await self._handle_get_config(request_id)
 
     async def _handle_update_context(self, data: RTVIAppendToContextData, request_id: str):
+        if data.role not in ["user", "assistant"]:
+            logger.warning(f"Unknown role {data.role} in RTVIAppendToContext")
+            await self._send_error_response(
+                request_id, f"Invalid role {data.role} for RTVIAppendToContext"
+            )
+
         if data.run_immediately:
             await self.interrupt_bot()
-        frame = LLMMessagesAppendFrame(messages=[{"role": data.role, "content": data.content}])
+        frame = LLMMessagesAppendFrame(
+            messages=[{"role": data.role, "content": data.content}],
+            run_llm=data.run_immediately,
+        )
         await self.push_frame(frame)
-        if data.run_immediately and self._context_aggregator:
-            # If specified, immediately push the full context frame to trigger an LLM run.
-            if data.role not in ["user", "assistant"]:
-                logger.warning(f"Unknown role {data.role} in RTVIAppendToContext")
-                await self._send_error_response(
-                    request_id, f"Invalid role {data.role} for RTVIAppendToContext"
-                )
-
-            # note that _context_aggregator.assistant().get_context_frame() returns
-            # the same frame, so we can just use user() here no matter the role.
-            frame = self._context_aggregator.user().get_context_frame()
-            await self.push_frame(frame)
 
     async def _handle_client_message(self, msg_id: str, data: RTVIRawClientMessageData):
         """Handle a client message frame."""
--- a/examples/deployment/modal-example/client/javascript/package-lock.json
+++ b/examples/deployment/modal-example/client/javascript/package-lock.json
@@ -9,49 +9,62 @@
       "version": "1.0.0",
       "license": "ISC",
       "dependencies": {
-        "@pipecat-ai/client-js": "^0.3.5",
-        "@pipecat-ai/daily-transport": "^0.3.10"
+        "@pipecat-ai/client-js": "file:../../../../../../pipecat-client-web/client-js",
+        "@pipecat-ai/daily-transport": "file:../../../../../../pipecat-client-web-transports/transports/daily"
       },
       "devDependencies": {
         "vite": "^6.3.5"
       }
     },
-    "node_modules/@babel/runtime": {
-      "version": "7.27.0",
-      "resolved": "https://registry.npmjs.org/@babel/runtime/-/runtime-7.27.0.tgz",
-      "integrity": "sha512-VtPOkrdPHZsKc/clNqyi9WUA8TINkZ4cGk63UUE3u4pmB2k+ZMQRDuIOagv8UVd6j7k0T3+RRIb7beKTebNbcw==",
-      "license": "MIT",
-      "dependencies": {
-        "regenerator-runtime": "^0.14.0"
-      },
-      "engines": {
-        "node": ">=6.9.0"
-      }
-    },
-    "node_modules/@daily-co/daily-js": {
-      "version": "0.77.0",
-      "resolved": "https://registry.npmjs.org/@daily-co/daily-js/-/daily-js-0.77.0.tgz",
-      "integrity": "sha512-icNXKieKAkRR/C5dcPjrCkL1jQGFp5C5WtLHy5uHAdTztm+mo9wlPJuehbWaGOM3TV24mgWHZ/+8jOys1G0I4w==",
+    "../../../../../../pipecat-client-web-transports/transports/daily": {
+      "name": "@pipecat-ai/daily-transport",
+      "version": "0.4.0",
       "license": "BSD-2-Clause",
       "dependencies": {
-        "@babel/runtime": "^7.12.5",
-        "@sentry/browser": "^8.33.1",
-        "bowser": "^2.8.1",
-        "dequal": "^2.0.3",
-        "events": "^3.1.0"
+        "@daily-co/daily-js": "^0.77.0"
       },
-      "engines": {
-        "node": ">=10.0.0"
+      "devDependencies": {
+        "@pipecat-ai/client-js": "file:../../../pipecat-client-web/client-js",
+        "eslint": "9.11.1",
+        "eslint-config-prettier": "^9.1.0",
+        "eslint-plugin-simple-import-sort": "^12.1.1"
+      },
+      "peerDependencies": {
+        "@pipecat-ai/client-js": "file:../../../pipecat-client-web/client-js"
+      }
+    },
+    "../../../../../../pipecat-client-web/client-js": {
+      "name": "@pipecat-ai/client-js",
+      "version": "0.4.0",
+      "license": "BSD-2-Clause",
+      "dependencies": {
+        "@types/events": "^3.0.3",
+        "clone-deep": "^4.0.1",
+        "events": "^3.3.0",
+        "typed-emitter": "^2.1.0",
+        "uuid": "^10.0.0"
+      },
+      "devDependencies": {
+        "@jest/globals": "^29.7.0",
+        "@types/clone-deep": "^4.0.4",
+        "@types/jest": "^29.5.12",
+        "@types/uuid": "^10.0.0",
+        "eslint": "^9.11.1",
+        "eslint-config-prettier": "^9.1.0",
+        "eslint-plugin-simple-import-sort": "^12.1.1",
+        "jest": "^29.7.0",
+        "ts-jest": "^29.2.5"
       }
     },
     "node_modules/@esbuild/aix-ppc64": {
-      "version": "0.25.4",
-      "resolved": "https://registry.npmjs.org/@esbuild/aix-ppc64/-/aix-ppc64-0.25.4.tgz",
-      "integrity": "sha512-1VCICWypeQKhVbE9oW/sJaAmjLxhVqacdkvPLEjwlttjfwENRSClS8EjBz0KzRyFSCPDIkuXW34Je/vk7zdB7Q==",
+      "version": "0.25.5",
+      "resolved": "https://registry.npmjs.org/@esbuild/aix-ppc64/-/aix-ppc64-0.25.5.tgz",
+      "integrity": "sha512-9o3TMmpmftaCMepOdA5k/yDw8SfInyzWWTjYTFCX3kPSDJMROQTb8jg+h9Cnwnmm1vOzvxN7gIfB5V2ewpjtGA==",
       "cpu": [
         "ppc64"
       ],
       "dev": true,
+      "license": "MIT",
       "optional": true,
       "os": [
         "aix"
@@ -61,13 +74,14 @@
       }
     },
     "node_modules/@esbuild/android-arm": {
-      "version": "0.25.4",
-      "resolved": "https://registry.npmjs.org/@esbuild/android-arm/-/android-arm-0.25.4.tgz",
-      "integrity": "sha512-QNdQEps7DfFwE3hXiU4BZeOV68HHzYwGd0Nthhd3uCkkEKK7/R6MTgM0P7H7FAs5pU/DIWsviMmEGxEoxIZ+ZQ==",
+      "version": "0.25.5",
+      "resolved": "https://registry.npmjs.org/@esbuild/android-arm/-/android-arm-0.25.5.tgz",
+      "integrity": "sha512-AdJKSPeEHgi7/ZhuIPtcQKr5RQdo6OO2IL87JkianiMYMPbCtot9fxPbrMiBADOWWm3T2si9stAiVsGbTQFkbA==",
       "cpu": [
         "arm"
       ],
       "dev": true,
+      "license": "MIT",
       "optional": true,
       "os": [
         "android"
@@ -77,13 +91,14 @@
       }
     },
     "node_modules/@esbuild/android-arm64": {
-      "version": "0.25.4",
-      "resolved": "https://registry.npmjs.org/@esbuild/android-arm64/-/android-arm64-0.25.4.tgz",
-      "integrity": "sha512-bBy69pgfhMGtCnwpC/x5QhfxAz/cBgQ9enbtwjf6V9lnPI/hMyT9iWpR1arm0l3kttTr4L0KSLpKmLp/ilKS9A==",
+      "version": "0.25.5",
+      "resolved": "https://registry.npmjs.org/@esbuild/android-arm64/-/android-arm64-0.25.5.tgz",
+      "integrity": "sha512-VGzGhj4lJO+TVGV1v8ntCZWJktV7SGCs3Pn1GRWI1SBFtRALoomm8k5E9Pmwg3HOAal2VDc2F9+PM/rEY6oIDg==",
       "cpu": [
         "arm64"
       ],
       "dev": true,
+      "license": "MIT",
       "optional": true,
       "os": [
         "android"
@@ -93,13 +108,14 @@
       }
     },
     "node_modules/@esbuild/android-x64": {
-      "version": "0.25.4",
-      "resolved": "https://registry.npmjs.org/@esbuild/android-x64/-/android-x64-0.25.4.tgz",
-      "integrity": "sha512-TVhdVtQIFuVpIIR282btcGC2oGQoSfZfmBdTip2anCaVYcqWlZXGcdcKIUklfX2wj0JklNYgz39OBqh2cqXvcQ==",
+      "version": "0.25.5",
+      "resolved": "https://registry.npmjs.org/@esbuild/android-x64/-/android-x64-0.25.5.tgz",
+      "integrity": "sha512-D2GyJT1kjvO//drbRT3Hib9XPwQeWd9vZoBJn+bu/lVsOZ13cqNdDeqIF/xQ5/VmWvMduP6AmXvylO/PIc2isw==",
       "cpu": [
         "x64"
       ],
       "dev": true,
+      "license": "MIT",
       "optional": true,
       "os": [
         "android"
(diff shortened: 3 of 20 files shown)


examples/deployment/modal-example/client/javascript/src/app.js
```diff
@@ -93,23 +93,12 @@ class ChatbotClient {
...
-        endpoints: {
-          connect: '/connect',
-        },
-        requestData: {
-          bot_name: 'openai',
-        },
-      },
+    // Initialize the Pipecat client with a DailyTransport and our configuration
+    this.pcClient = new PipecatClient({
+      transport: DailyTransport.create(),
```
mattieruth: for most clients, this is the primary change:
- transport now takes a factory function
- `params` is not a thing. instead, see `connect()` where you can still pass in the endpoint for getting credentials.

examples/deployment/modal-example/client/javascript/src/app.js
```diff
@@ -311,21 +300,27 @@ class ChatbotClient {
...
 
       // Initialize audio/video devices
       this.log('Initializing devices...');
-      await this.rtviClient.initDevices();
+      await this.pcClient.initDevices();
 
       // Connect to the bot
       this.log(`Connecting to bot: ${selectedBot}`);
-      await this.rtviClient.connect();
+      await this.pcClient.connect({
```
mattieruth, line 316: connect takes one of two things:
1. a set of options matching what the transport expects (like DailyCallOptions `{url: <>, token: <>}`)
2. an object defining a REST endpoint that should return the set of options matching what the transport expects:
    ```
    { endpoint: 'https://yourendpoint/connect` } => returns { url: <>, token: <> }
    ```

examples/instant-voice/client/javascript/src/app.ts
```diff
@@ -113,30 +110,23 @@ class InstantVoiceClient {
         onBotTranscript: (data) => this.log(`Bot: ${data.text}`),
         onMessageError: (error) => console.error('Message error:', error),
         onError: (error) => console.error('Error:', error),
-      },
+        onAudioBufferingStarted: () => {
```
mattieruth, line 113: a change:  `DailyTransport` now extends the RTVIMessage callbacks defined in the client so that you simply add the `onAudioBufferingStarted/Stopped` callbacks same as you would your other callbacks. But you do have to do the type dance below to tell it that we're passing in `DailyEventCallbacks`. This makes it far more intuitive and built-in than having to define and create your own custom helper.

examples/news-chatbot/client/javascript/src/app.js
```diff
@@ -16,86 +16,17 @@
...
-import {
-  LogLevel,
-  RTVIClient,
-  RTVIClientHelper,
-  RTVIEvent,
-} from '@pipecat-ai/client-js';
+import { LogLevel, PipecatClient, RTVIEvent } from '@pipecat-ai/client-js';
 import { DailyTransport } from '@pipecat-ai/daily-transport';
 
-class SearchResponseHelper extends RTVIClientHelper {
```
mattieruth, line 27: this completely goes away :)

mattieruth, line 27: this class completely goes away and we just need the handler function below :)

examples/news-chatbot/client/javascript/src/app.js
```diff
@@ -251,6 +175,8 @@ class ChatbotClient {
               this.setupMediaTracks();
             }
           },
+          // Handle search response events
+          onBotLlmSearchResponse: this.handleSearchResponse.bind(this),
```
mattieruth, line 179: no more helper rigamarole, just register your handler

examples/simple-chatbot/client/javascript/src/app.js
```diff
@@ -16,19 +16,34 @@
  * - Browser with WebRTC support
  */
 
-import { RTVIClient, RTVIEvent } from '@pipecat-ai/client-js';
+import { PipecatClient, RTVIEvent } from '@pipecat-ai/client-js';
 import { DailyTransport } from '@pipecat-ai/daily-transport';
 
+function _generateRandomWeather() {
```
mattieruth, line 22: i've expanded this sample to implement function calling that requires a back/forth with the server. In this case, I am bulding the common "give me the weather" function call, but implementing the logic for getting the weather on the client. This is not how you would normally implement this. Typically, you'd probably do that on the server too, but you might need to ask the client for its location so there could still be a back/forth. I left it this way to be very clear about requiring the client to be in the mix.

examples/simple-chatbot/client/javascript/src/app.js
```diff
@@ -42,6 +57,7 @@ class ChatbotClient {
     this.statusSpan = document.getElementById('connection-status');
     this.debugLog = document.getElementById('debug-log');
     this.botVideoContainer = document.getElementById('bot-video-container');
+    this.deviceSelector = document.getElementById('device-selector');
```
mattieruth, line 60: i've also added a device selector to the sample because i was tired of not having it 😂

examples/simple-chatbot/client/javascript/src/app.js
```diff
@@ -129,6 +179,49 @@ class ChatbotClient {
...
     // Set up listeners for media track events
     this.setupTrackListeners();
+
+    await this.pcClient.initDevices();
+    this.setupEventListeners();
+    this.registerFunctionCallHandlers();
+  }
+
+  registerFunctionCallHandlers() {
+    this.pcClient.registerFunctionCallHandler(
```
mattieruth, line 190: here's the interesting v2 bit, demonstrating how to register function call handlers on the client.

examples/word-wrangler-gemini-live/server/bot.py
```diff
@@ -228,7 +228,24 @@ async def local_daily():
 
 # Local development entry point
 if LOCAL_RUN and __name__ == "__main__":
+    parser = argparse.ArgumentParser(
```
mattieruth, line 230: for local testing, i updated the server/bot so that the client personality selection actually worked.

src/pipecat/processors/frameworks/rtvi.py
```diff
@@ -64,12 +66,13 @@
...
     FunctionCallParams,  # TODO(aleix): we shouldn't import `services` from `processors`
 )
+from pipecat.services.openai.llm import OpenAIContextAggregatorPair
 from pipecat.transports.base_input import BaseInputTransport
 from pipecat.transports.base_output import BaseOutputTransport
 from pipecat.transports.base_transport import BaseTransport
 from pipecat.utils.string import match_endofsentence
 
-RTVI_PROTOCOL_VERSION = "0.3.0"
+RTVI_PROTOCOL_VERSION = "1.0.0"
```
kompfner, line 82: 🥳

src/pipecat/processors/frameworks/rtvi.py
```diff
@@ -143,29 +194,116 @@ class RTVIConfig(BaseModel):
...
+    """
+
+    msg_id: str
+    type: str
+    data: Optional[Any] = None
+
+
+@dataclass
+class RTVIServerResponseFrame(SystemFrame):
+    """A frame for sending messages from the client to the RTVI server.
```
kompfner: I think this docstring is wrong. Looks like it's copy-pasted from `RTVIClientMessageFrame`

src/pipecat/processors/frameworks/rtvi.py
```diff
@@ -143,29 +194,116 @@ class RTVIConfig(BaseModel):
...
+
+class RTVIRawServerResponseData(BaseModel):
+    """Data structure for server responses to client messages."""
+
+    t: str
+    d: Optional[Any] = None
+
+
+class RTVIServerResponse(BaseModel):
+    """A response message from the client to the RTVI server.
```
kompfner: I think this docstring is backwards (both sentences). Implies that the server sends the initial message and the client responds.

src/pipecat/processors/frameworks/rtvi.py
```diff
@@ -178,70 +316,134 @@ class RTVIMessage(BaseModel):
...
+
+class RTVIClientReadyData(BaseModel):
+    version: str
+    about: AboutClientData
+
+
 class RTVIBotReadyData(BaseModel):
     version: str
-    config: List[RTVIServiceConfig]
+    config: Optional[List[RTVIServiceConfig]] = None
```
kompfner, line 572: `config` should be considered deprecated, right? If so, wonder if it's worth a comment or anything else calling it out as such.

src/pipecat/processors/frameworks/rtvi.py
```diff
@@ -629,13 +848,30 @@ async def _handle_metrics(self, frame: MetricsFrame):
         message = RTVIMetricsMessage(data=metrics)
         await self.push_transport_message_urgent(message)
 
+    async def _send_server_response(self, frame: RTVIServerResponseFrame):
+        """Send a response to the client for a specific request."""
+        message = RTVIServerResponse(
+            id=str(frame.client_msg.msg_id),
+            data=RTVIRawServerResponseData(t=frame.client_msg.type, d=frame.data),
+        )
+        await self.push_transport_message_urgent(message)
```
kompfner, line 1101: It legit took me *forever* to wrap my head around the wildly circuitous song and dance of:
- catching a user-inserted `RTVIServerResponseFrame` in the `RTVIObserver`
- unwrapping the data from it
- re-wrapping it into a `TransportMessageUrgentFrame`
- pushing that frame *from* the `RTVIProcessor` that we keep a reference to (like, essentially faking that the `RTVIProcessor` sent that frame)
- letting that frame make its way down to the `BaseOutputTransport` part of the pipeline
- where it's eventually sent to the client

What's maybe most confusing is that `TransportMessageUrgentFrame` is *both* something that the `RTVIProcessor` handles from the input transport *and* sends downstream to the output transport.

src/pipecat/processors/frameworks/rtvi.py
```diff
@@ -690,6 +947,21 @@ def set_errors_enabled(self, enabled: bool):
     async def interrupt_bot(self):
         await self.push_frame(BotInterruptionFrame(), FrameDirection.UPSTREAM)
 
+    async def send_server_message(self, data: Any):
```
kompfner, line 1222: I see an example of using `send_server_response()` but no examples of using `send_server_message()`

src/pipecat/processors/frameworks/rtvi.py
kompfner: Ah, I see where you "deprecate" `config` by not sending it to newer clients. Cool 👍 

(Still maybe worth a comment here like "Deprecated; not sent to clients speaking RTVI version >= 1.0")

examples/news-chatbot/client/javascript/src/app.js
kompfner: is this because the concept of a bot LLM search response has been standardized and should now be treated as more universal than a google thing?

If so—should we also move the server-side definitions (e.g. `RTVIBotLLMSearchResponseMessage`) out of the google-specific `rtvi.py` and into the main, shared `rtvi.py`?

examples/simple-chatbot/client/javascript/src/app.js
kompfner: Maybe worth a code comment to this effect?

examples/simple-chatbot/server/bot-openai.py
```diff
@@ -69,6 +81,68 @@
...
+        """
+        await super().process_frame(frame, direction)
+
+        if isinstance(frame, FunctionCallResultFrame):
+            print("Function call result:", frame.tool_call_id, frame.result)
+            if "weather" in frame.result and "condition" in frame.result["weather"]:
+                frame.result["weather"]["condition"] = "hazy"
+            if frame.tool_call_id in self.waiting_calls:
+                del self.waiting_calls[frame.tool_call_id]
+        # elif isinstance(frame, RTVIClientMessageFrame):
```
kompfner, line 124: Maybe a comment noting why you might choose this pattern over the `@rtvi.event_handler("on_client_message")` pattern (if there's a reason to do it this way).

I think I understand why this code is here and commented-out, though: you want to be able to quickly uncomment it and test it with this alternate pattern.

examples/simple-chatbot/server/bot-openai.py
```diff
@@ -69,6 +81,68 @@
...
+    manages the response.
+    """
+
+    # currently does nothing but tracks waiting calls but could be used
+    waiting_calls: Dict[str, FunctionCallParams] = {}
+
+    def __init__(self):
+        super().__init__()
+
+    async def fetch_weather(self, params: FunctionCallParams):
```
kompfner, line 103: If it weren't for trying to illustrate how you could track `waiting_calls`, you wouldn't need all the `CustomProcessor` rigamarole at all, would you?

Wonder if—since this is a public demo—it'd be better to just have `fetch_weather` be a standalone function and cut out a lot of boilerplate.

src/pipecat/processors/frameworks/rtvi.py
```diff
@@ -915,6 +1258,41 @@ async def _handle_update_config(self, request_id: str, data: RTVIUpdateConfig):
...
         await self._handle_get_config(request_id)
 
+    async def _handle_update_context(self, data: RTVIAppendToContextData):
+        if data.run_immediately:
+            await self.interrupt_bot()
+        frame = LLMMessagesAppendFrame(messages=[{"role": data.role, "content": data.content}])
+        await self.push_frame(frame)
+        if data.run_immediately and self._context_aggregator:
+            # If specified, immediately push the full context frame to trigger an LLM run.
+            match data.role:
```
kompfner: I don't think it matters whether you pull the context using `self._context_aggregator.user().get_context_frame()` or  `self._context_aggregator.assistant().get_context_frame()`. They point to the exact same context, IIRC (it's a slightly confusing Pipecat quirk that folks have discussed in the past). So this bit of code might mislead the reader into thinking otherwise.

src/pipecat/processors/frameworks/rtvi.py
```diff
@@ -915,6 +1258,41 @@ async def _handle_update_config(self, request_id: str, data: RTVIUpdateConfig):
         await self._update_config(RTVIConfig(config=data.config), data.interrupt)
         await self._handle_get_config(request_id)
 
+    async def _handle_update_context(self, data: RTVIAppendToContextData):
+        if data.run_immediately:
+            await self.interrupt_bot()
+        frame = LLMMessagesAppendFrame(messages=[{"role": data.role, "content": data.content}])
+        await self.push_frame(frame)
+        if data.run_immediately and self._context_aggregator:
```
kompfner: Hmm...isn't it theoretically possible that this `run_immediately` case is run before the `LLMMessagesAppendFrame` has reached a context aggregator and had the desired effect on the context?

Ah wait...I suppose it doesn't matter because when you then get a context frame using `get_context_frame()`, that frame has a reference to the "live" context rather than a snapshot of it at that point in time. So...we're probably good.

src/pipecat/processors/frameworks/rtvi.py
mattieruth: yeah, i legit took forever to understand that logic too. i'm not sure i ever got as far as you or if i finally just trusted the existing code and the fact that the new stuff worked. but i'm with you, it's confusing. probably worth a round of internal cleanup or better inline comments at some point

examples/simple-chatbot/server/bot-openai.py
mattieruth: yeah, you're reminding me that i probably didn't mean to leave quite a few of these changes in simple-chatbot. it and word wrangler were definitely was my primary testing harness

mattieruth: i'll chat with @markbackman and see if he has a good idea for better showing this stuff off and if this is the right example to do so in.

examples/news-chatbot/client/javascript/src/app.js
mattieruth: when we had our initial discussion, kwin believed that most llms would end up supporting this behavior, but for now i think it's still just google. But TBH, I didn't give much thought to the existing server-side handling/setup for this feature. Not sure if it should move

examples/simple-chatbot/server/bot-openai.py
mattieruth: not sure i follow, but maybe it will be moot anyway after i chat with mark and figure out the best way and where to demo this stuff.

src/pipecat/processors/frameworks/rtvi.py
mattieruth: good point. i added it for ease-of-use and since it seemed to make sense, given the other two methods but I guess i never demo'd it. i'll try to work it in to whatever example we land on.

src/pipecat/processors/frameworks/rtvi.py
mattieruth: i think that was my understanding but cc @aconchillo to confirm.

src/pipecat/processors/frameworks/rtvi.py
mattieruth: oh really? i didn't know that. I based this on what i saw in code and just assumed. if so, that's much simpler :)

src/pipecat/processors/frameworks/rtvi.py
```diff
@@ -942,12 +1320,19 @@ async def _handle_action(self, request_id: Optional[str], data: RTVIActionRun):
             await self._push_transport_message(message)
 
     async def _send_bot_ready(self):
+        config = None
+        if self._client_version[0] < 1:
+            config = self._config.config
```
mattieruth, line 1647: i should probably add this logic in other places too (refuse context updates and actions), but wasn't sure how aggressive to be on this first release.

examples/news-chatbot/client/javascript/src/app.js
kompfner: ok to not do it right away! but i do think it'd be nice if server-side organization matched client-side: if something is a universal (as opposed to provider-specific) concept on one side of the fence, it should also be universal on the other side

mattieruth: i looked at this real quick-like... looks like that would be simple, except that it would need to know what a `LLMSearchResponseFrame` looks like, which is defined in the google service and i'm guessing we don't want rtvi to depend on goog. so maybe a question for @aleix as to if it makes any sense to move that Frame also up to a more generic place or keep it all as-is.

src/pipecat/processors/frameworks/rtvi.py
```diff
@@ -915,6 +1258,41 @@ async def _handle_update_config(self, request_id: str, data: RTVIUpdateConfig):
...
+            # If specified, immediately push the full context frame to trigger an LLM run.
+            match data.role:
+                case "user":
+                    frame = self._context_aggregator.user().get_context_frame()
+                    await self.push_frame(frame)
+                case "assistant":
+                    frame = self._context_aggregator.assistant().get_context_frame()
+                    await self.push_frame(frame)
+                case _:
+                    logger.warning(f"Unknown role {data.role} in RTVIAppendToContext")
```
aconchillo: I don't think we need to distinguish between the role here, I believe both should return the same frame.

src/pipecat/processors/frameworks/rtvi.py
aconchillo: Oh, I just made that same comment.

Also, why do we need to push a context frame? I don't think we need a copy of `self._context_aggregator`. Actually, it would be great if we didn't need it. In theory, everything should be done through frames, we should always try to avoid passing instances of other classes and calling their methods.

aconchillo: I'm updating the LLMMessagesAppendFrame so we don't need to send the context from here.

aconchillo: https://github.com/pipecat-ai/pipecat/pull/2119

src/pipecat/processors/frameworks/rtvi.py
aconchillo: Mmmmm. I think we should just push a transport message directly. The `RTVIObserver` is supposed to be used for frames originating from other processors. But if we can push a frame directly, we should do that.

aconchillo: So, basically, moving this exact same code to the RTVIProcessor.

src/pipecat/processors/frameworks/rtvi.py
aconchillo: Actually, the `LLMMessagesAppendFrame` might get processed after we do `self._context_aggregator.user().get_context_frame()`, so this might work sometimes and other it might not. I send a PR so we don't need to send a context frame, and therefore we don't need to pass the context aggregator to the constructor.

src/pipecat/processors/frameworks/rtvi.py
aconchillo: Ah, wait.... Is it possible that some processor can send `RTVIServerResponseFrame`? If so, I guess this is fine. But agreed, it might be confusing.

src/pipecat/processors/frameworks/rtvi.py
mattieruth: cool. i'll leave as-is for now and i guess update once that goes in? for now, i've updated to check the role and send an error if it's not 'user' or 'assistant' and then just use the user to get the frame.

src/pipecat/processors/frameworks/rtvi.py
```diff
@@ -1244,6 +1579,39 @@ async def _handle_update_config(self, request_id: str, data: RTVIUpdateConfig):
         await self._update_config(RTVIConfig(config=data.config), data.interrupt)
         await self._handle_get_config(request_id)
 
+    async def _handle_update_context(self, data: RTVIAppendToContextData, request_id: str):
+        if data.role not in ["user", "assistant"]:
+            logger.warning(f"Unknown role {data.role} in RTVIAppendToContext")
+            await self._send_error_response(
+                request_id, f"Invalid role {data.role} for RTVIAppendToContext"
+            )
```
aconchillo, line 1587: should we return here?

aconchillo: Actually, maybe we should remove this check. In theory we have this protection:

```
    role: Literal["user", "assistant"] | str
```

And `str` could be used to set any role. I think it's more generic if we remove this block.

src/pipecat/processors/frameworks/rtvi.py
mattieruth: sorry, i'm lost. what's the question? what should i do? 😂

aconchillo: Nothing, sorry. I was just going back and forth. Trying to understand why we had to do this.

examples/simple-chatbot/client/javascript/src/app.js
mattieruth: in order to get this merged and released, i'm punting the work of making an example that makes sense for this stuff to another PR and removing the function handling and client messaging bits from this file for now.

mattieruth:
closing in favor of the cleaned up, more professional looking #2136 :)

Issue #1919, opened 2025-05-29
Pipeline get Frozen


anotine10:
### pipecat version

0.0.68

### Python version

3.13.2

### Operating System

macos

### Issue description

I run pipecat pipeline and connect to twilio outbound call, through websocke (implemented on top of django channels).

Conversation starts well, i can speak with Assistant. Then, when i start to interact more strangely (interrupting, or making long sentence with some silence), there is a point in time where Assistant get frozen.
From that point, assistant will never speak again. 
If i put background music, i can still hear the background, but Assistant will never speak again. I can reproduce with and without background music.

### Reproduction steps

It is hard to find a strict guidelines to reproduce, but basically, i made interruptions of assistant.

While debugging i have figured the following:
- it all starts with a StartInterruptionFrame being pushed into pipeline.
- The task that pushes that frames seems to be frozen when trying to run the _start_interruption() of the OpenAiLLMService. It can not go beyond the cancel_input_task() function. 
- It seems that task that is getting stuck, he also the task that was handling frames at the begin from the pipeline. so no more input frame is getting pushed after in the pipeline.

StartInterruptionFrame being pushed here at the beginning:
```
    async def _handle_vad(self, audio_frame: InputAudioRawFrame, vad_state: VADState):
        new_vad_state = await self._vad_analyze(audio_frame)
        if (
            new_vad_state != vad_state
            and new_vad_state != VADState.STARTING
            and new_vad_state != VADState.STOPPING
        ):
            frame = None
            # If the turn analyser is enabled, this will prevent:
            # - Creating the UserStoppedSpeakingFrame
            # - Creating the UserStartedSpeakingFrame multiple times
            can_create_user_frames = (
                self._params.turn_analyzer is None
                or not self._params.turn_analyzer.speech_triggered
            )
            if new_vad_state == VADState.SPEAKING:
                await self.push_frame(VADUserStartedSpeakingFrame())
                if can_create_user_frames:
                    frame = UserStartedSpeakingFrame()
            elif new_vad_state == VADState.QUIET:
                await self.push_frame(VADUserStoppedSpeakingFrame())
                if can_create_user_frames:
                    frame = UserStoppedSpeakingFrame()

            if frame:
                <<<<<<< HERE
                await self._handle_user_interruption(frame)

            vad_state = new_vad_state
        return vad_state
```

task get stuck here (FrameProcessor/OpenaiLLMService
```
    async def _start_interruption(self):
        try:
[...]

PR #2004, opened 2025-06-13
Pipeline freeze experiments

filipi87:
In progress....

### DIFF
--- a/src/pipecat/processors/frame_processor.py
+++ b/src/pipecat/processors/frame_processor.py
@@ -299,9 +299,8 @@ class FrameProcessor(BaseObject):
             # Cancel the input task. This will stop processing queued frames.
             await self.__cancel_input_task()
         except Exception as e:
-            logger.exception(f"Uncaught exception in {self}: {e}")
+            logger.exception(f"Uncaught exception in {self} when handling _start_interruption: {e}")
             await self.push_error(ErrorFrame(str(e)))
-            raise
 
         # Create a new input queue and task.
         self.__create_input_task()
--- a/src/pipecat/utils/asyncio.py
+++ b/src/pipecat/utils/asyncio.py
@@ -174,10 +174,19 @@ class TaskManager(BaseTaskManager):
         except asyncio.CancelledError:
             # Here are sure the task is cancelled properly.
             pass
+        except GeneratorExit as e:
+            logger.exception(f"{name}: generator exit error while cancelling task: {e}")
+            raise
         except Exception as e:
             logger.exception(f"{name}: unexpected exception while cancelling task: {e}")
+        except BaseException as e:
+            logger.critical(f"{name}: fatal base exception while cancelling task: {e}")
+            raise
         finally:
-            self._remove_task(task)
+            try:
+                self._remove_task(task)
+            except Exception as e:
+                logger.warning(f"{name}: failed to remove task: {e}")
 
     def current_tasks(self) -> Sequence[asyncio.Task]:
         """Returns the list of currently created/registered tasks."""


src/pipecat/transports/base_input.py
```diff
@@ -355,26 +363,38 @@ async def _run_turn_analyzer(
...
-            previous_vad_state = vad_state
-            if self._params.vad_analyzer:
-                vad_state = await self._handle_vad(frame, vad_state)
-
-            if self._params.turn_analyzer:
-                await self._run_turn_analyzer(frame, vad_state, previous_vad_state)
+            try:
+                frame: InputAudioRawFrame = await asyncio.wait_for(
+                    self._audio_in_queue.get(), timeout=AUDIO_INPUT_TIMEOUT_SECS
+                )
```
filipi87, line 369: In WebRTC, we are always receiving silence when the user is not speaking, but this is not always true for WebSocket, it depends on the client's implementation. 
So here is a safeguard in case we stop receiving audio frames.

src/pipecat/processors/frame_processor.py
```diff
@@ -96,6 +96,10 @@ def __init__(
         self.__input_event = asyncio.Event()
         self.__input_frame_task: Optional[asyncio.Task] = None
 
+        # Handle to cancel the tasks gracefully
+        self.__input_frame_task_cancelling = False
+        self.__push_frame_task_cancelling = False
```
filipi87: In theory, invoking `task.cancel()` schedules a `CancelledError` to be thrown at the next `await` inside the coroutine, allowing it to exit gracefully.

But we've observed some rare cases where this does not happen as expected, and the task may hang during cancellation. Like if it's stuck in synchronous code, not hitting an await, or suppressing the `CancelledError`.

So I am adding this as a safeguard mechanism to ensure the task is properly cancelled even when the standard `task.cancel()` approach fails.
This approach has been suggested in this issue: https://github.com/pipecat-ai/pipecat/issues/1919#issuecomment-2921498074
And it seems to be working.

src/pipecat/processors/aggregators/llm_response.py
```diff
@@ -391,10 +391,21 @@ async def _handle_user_started_speaking(self, frame: UserStartedSpeakingFrame):
...
+                await self.push_aggregation()
+        else:
+            # Handles the case where both the user and the bot are not speakin.
+            # Normally, when the user stops speaking, new text is expected,
+            # which triggers the bot to respond. However, if no new text
+            # is received (e.g., due to a glitch), this safeguard ensures
+            # the bot doesn't hang indefinitely while waiting to speak again.
+            if not self._bot_speaking:
+                logger.debug("User stopped speaking but no new aggregation received. Forcing aggregation processing to resume bot response.")
+                await self._process_aggregation()
```
filipi87: Handling the case where user stopped speaking but no new aggregation received, and the bot had stopped speaking.

src/pipecat/utils/asyncio.py
```diff
@@ -174,10 +174,19 @@ async def cancel_task(self, task: asyncio.Task, timeout: Optional[float] = None)
         except asyncio.CancelledError:
             # Here are sure the task is cancelled properly.
             pass
+        except GeneratorExit as e:
+            logger.exception(f"{name}: generator exit error while cancelling task: {e}")
+            raise
```
aconchillo: `BaseException` should cover `GeneratorExit`. Maybe we can just use that since we print the exception.

src/pipecat/utils/asyncio.py
```diff
@@ -174,10 +174,19 @@ async def cancel_task(self, task: asyncio.Task, timeout: Optional[float] = None)
...
             logger.exception(f"{name}: unexpected exception while cancelling task: {e}")
+        except BaseException as e:
+            logger.critical(f"{name}: fatal base exception while cancelling task: {e}")
+            raise
         finally:
-            self._remove_task(task)
+            try:
+                self._remove_task(task)
+            except Exception as e:
+                logger.warning(f"{name}: failed to remove task: {e}")
```
aconchillo: we are already handling exceptions inside `_remove_task`, probably not needed here.

src/pipecat/transports/network/fastapi_websocket.py
```diff
@@ -70,11 +70,22 @@ def receive(self) -> typing.AsyncIterator[bytes | str]:
...
+                else:
+                    await self._websocket.send_text(data)
+        except Exception as e:
+            logger.error(
+                f"{self} exception sending data: {e.__class__.__name__} ({e}), application_state: {self._websocket.application_state}"
+            )
+            # For some reason the websocket is disconnected, and we are not able to send data
+            # So let's properly handle it and disconnect the transport
+            if self._websocket.application_state == WebSocketState.DISCONNECTED:
+                logger.info("Closing already disconnected websocket!")
```
aconchillo: Maybe this could be a warning.

src/pipecat/transports/base_input.py
```diff
@@ -355,26 +363,38 @@ async def _run_turn_analyzer(
...
+                    await self._run_turn_analyzer(frame, vad_state, previous_vad_state)
+
+                # Push audio downstream if passthrough is set.
+                if self._params.audio_in_passthrough:
+                    await self.push_frame(frame)
+
+                self._audio_in_queue.task_done()
+            except asyncio.TimeoutError:
+                if self._user_speaking:
+                    logger.info(
```
aconchillo: make it a `warning`

aconchillo: nit: to receive -> receiving

src/pipecat/processors/frame_processor.py
```diff
@@ -388,29 +392,42 @@ async def __input_frame_task_handler(self):
                 self.__should_block_frames = False
                 logger.trace(f"{self}: frame processing resumed")
 
-            (frame, direction, callback) = await self.__input_queue.get()
+            if self.__input_frame_task_cancelling and self.__input_queue.qsize() == 0:
+                logger.warning(f"{self}: input frame task exiting due to cancellation")
+                self.__input_frame_task_cancelling = False
+                break
```
aconchillo: I'm not sure I understand the changes in this file. It seems we will now cancel after processing all the input queue instead of right away.

src/pipecat/transports/base_input.py
```diff
@@ -355,26 +363,37 @@ async def _run_turn_analyzer(
...
+                if self._params.audio_in_passthrough:
+                    await self.push_frame(frame)
 
-            # Push audio downstream if passthrough is set.
-            if self._params.audio_in_passthrough:
-                await self.push_frame(frame)
+                self._audio_in_queue.task_done()
+            except asyncio.TimeoutError:
+                if self._user_speaking:
+                    logger.info("Forcing user stopped speaking due to timeout to receive audio frame!")
```
markbackman: I'm assuming that this is an unexpected case that would happen only if there's a communication issue on the transport. For that reason, maybe this should be a `logger.warning`?

src/pipecat/processors/frame_processor.py
aconchillo: @filipi87 Do you have a way to reproduce this?

src/pipecat/transports/base_input.py
markbackman: This makes sense! Just to make extra sure, are there any cases where the transport would expect to not receive continuous audio?

Just thinking through scenarios where the track is manipulated on the client level (e.g. client's track is set to canSend: False). I think that's probably a rare case and one that wouldn't make sense for voice AI use cases, but just asking to understand if this is _always_ true.

In the websocket case, I don't think there's a way to have this level of track control, so media should be constant.

src/pipecat/processors/frame_processor.py
aconchillo: If the issue is sync code which prevents from hitting an `await` I believe the solution would be to warn the user that their code is wrong. I feel this solution is a bit brittle (even if it seems to work).

src/pipecat/processors/aggregators/llm_response.py
```diff
@@ -391,10 +391,23 @@ async def _handle_user_started_speaking(self, frame: UserStartedSpeakingFrame):
...
-        if not self._seen_interim_results:
-            await self.push_aggregation()
+        # push.
+        if len(self._aggregation) > 0:
+            # If the last thing we saw is an interim transcription, let's wait
+            # pushing the aggregation as we will probably get a final transcription.
+            if not self._seen_interim_results:
+                await self.push_aggregation()
+        else:
+            # Handles the case where both the user and the bot are not speakin.
```
markbackman:
```suggestion
            # Handles the case where both the user and the bot are not speaking.
```

aconchillo: nit: speaking

src/pipecat/processors/aggregators/llm_response.py
```diff
@@ -391,10 +391,23 @@ async def _handle_user_started_speaking(self, frame: UserStartedSpeakingFrame):
...
+            # Normally, when the user stops speaking, new text is expected,
+            # which triggers the bot to respond. However, if no new text
+            # is received (e.g., due to a glitch), this safeguard ensures
+            # the bot doesn't hang indefinitely while waiting to speak again.
+            if not self._bot_speaking:
+                logger.debug(
+                    "User stopped speaking but no new aggregation received. Forcing aggregation processing to resume bot response."
+                )
+                await self._process_aggregation()
```
markbackman: What does it mean to _process_aggregation when there is no new text to aggregate? This scenario can happen when:
- User speech is detected by the VAD
- STT didn't return a final transcript

This can happen and is a valid case. Normally, we recommend using something like the UserIdleProcessor to detect and handle these "dead end" parts to the conversation. This is better than just forcing the bot to respond again without any new user input.

aconchillo: This would add a message with an empty content I believe. How do LLMs respond to that? I'm wondering if we could hit an issue depending on the LLM. Also, what do we mean by a "glitch"? If there's really a glitch it seems we should try to understand what these glitches are. Is there a way to reproduce this?

src/pipecat/pipeline/task.py
```diff
@@ -662,18 +669,23 @@ async def _idle_monitor_handler(self):
...
+                        running = await self._idle_timeout_detected(frame_buffer)
 
                 self._idle_queue.task_done()
+
             except asyncio.TimeoutError:
-                running = await self._idle_timeout_detected()
+                running = await self._idle_timeout_detected(frame_buffer)
 
-    async def _idle_timeout_detected(self) -> bool:
+    async def _idle_timeout_detected(self, last_frames) -> bool:
```
aconchillo: `last_frames: Deque[Frame]`

examples/freeze_test/README.md
```diff
@@ -0,0 +1,59 @@
+# Freeze Test Client
```
aconchillo: nit: all our examples uses a dash, so `freeze-test`

src/pipecat/utils/asyncio.py
filipi87: True. I initially had this one with the intention of maybe not raising the exception in this case. But then I changed my mind. Yeah, for logging only, BaseException should be enough. 👍

src/pipecat/processors/frame_processor.py
filipi87: Hi @aconchillo, we still try to cancel right away, since we keep invoking `self.cancel_task`.

In theory, when we invoke a task cancellation, it schedules a `CancelledError` to be thrown at the next `await` inside the coroutine.

But if something happens inside the code, for example, it is capturing and suppressing the `CancelledError`, or in any other unexpected scenario, the task may keep running, and the `await task` may never return.

So this `self.__input_frame_task_cancelling` is acting as a safeguard mechanism to ensure the task is properly cancelled even on these cases when the standard `task.cancel()` approach fails.

filipi87: I have been able to reproduce it with the demo I created, but in all my tests, I’ve only been able to reproduce it twice so far. So it’s not easily reproducible. It usually happens with long sentences and a lot of interruptions. 🤕

src/pipecat/processors/aggregators/llm_response.py
filipi87: Hi @markbackman, this is handling the case where, like you said:
- User speech is detected by the VAD.
- This interrupts the bot, which stops speaking.
- STT doesn't return any transcription, no text.

So, when VAD detects that the user has stopped speaking, it triggers the bot to resume using the previous context.
Otherwise, the bot would hang, waiting for the user to speak again.

But you are right, one thing we should do here is trigger the resume only if the bot was speaking previously, otherwise it doesn't make sense.

filipi87: That "glitch" word maybe was not the best choice. I mean, if we receive some audio, which is detected by VAD, but doesn't produces any transcription. 

I tried to explain it better [here](https://github.com/pipecat-ai/pipecat/pull/2004#discussion_r2157017127).

src/pipecat/utils/asyncio.py
filipi87: Done.

src/pipecat/utils/asyncio.py
filipi87: Done.

src/pipecat/transports/network/fastapi_websocket.py
filipi87: Done.

src/pipecat/transports/base_input.py
filipi87: Done.

src/pipecat/transports/base_input.py
filipi87: Done.

src/pipecat/processors/aggregators/llm_response.py
filipi87: Done.

filipi87: Done.

src/pipecat/pipeline/task.py
filipi87: Done.

examples/freeze_test/README.md
filipi87: Done.

src/pipecat/transports/base_input.py
filipi87: For WebRTC, in theory, we should always receive at least silence.

For WebSocket, we depend on the client side to send the messages. So, if it’s a custom client, you can choose not to send anything, like in the `freeze-test` where that's possible.

For Twilio, I believe this issue should only occur in cases of temporary failure, which could result in audio not being sent maybe ? I'm not sure what happens if the user puts the call on hold, whether we should still receive audio in that case.

src/pipecat/processors/aggregators/llm_response.py
filipi87: > Is there a way to reproduce this?

You can reproduce this using the freeze-test demo.
Click "Connect" to start recording audio. If I’m not mistaken, it records 5 seconds of audio.
Once the recording is complete, it repeatedly sends the same audio to the bot at random intervals.
There are two buttons you can use to start and stop sending the recording.
Sometimes, if you click "Start" and then quickly click "Stop," you might trigger this edge case.

filipi87: I have implemented to only trigger this if the bot was previously speaking.

src/pipecat/transports/base_input.py
markbackman: Thanks for the explanation.

src/pipecat/processors/aggregators/llm_response.py
```diff
@@ -266,6 +266,7 @@ def __init__(
 
         self._user_speaking = False
         self._bot_speaking = False
+        self._was_bot_speaking = False
```
markbackman, line 269: Do we also need to reset this state when the bot stops speaking in the `_handle_bot_stopped_speaking` handler?

markbackman: I think the cases would be:
- Bot currently speaking, then user speaks but no transcript => Should prompt the bot to keep talking
- Bot stops speaking, then user speaks but no transcript => Do nothing (up to developers to handle, maybe through UserIdleProcessor?)

src/pipecat/processors/aggregators/llm_response.py
filipi87: I have changed to push only the frame, like this:
`await self.push_frame(OpenAILLMContextFrame(self._context))`

Instead of invoking `_process_aggregation `, which would add a message with an empty content as @aconchillo mentioned in another comment.

filipi87: For testing, I am using the freeze-test demo, saying a sentence like: `Hello. Can you tell me a brief history of Brazil?` followed by some mix of word and noise, which will pass our VAD, but which we are not going to receive any transcription from deepgram.

src/pipecat/processors/aggregators/llm_response.py
filipi87: @markbackman , I had to remove this. I have explained it [in this comment](https://github.com/pipecat-ai/pipecat/pull/2004#discussion_r2157613611).

src/pipecat/processors/aggregators/llm_response.py
filipi87: Ok, I think I found an easier way to reproduce that issue where the doesn’t respond.
In the same sentence I start speaking in english, than I do a small pause, and then I say another word in any other language, which will make us don’t receive a transcription from Deepgram. 

If you test this on `main`, the bot will not respond. Here it is responding fine due to this change.

filipi87: I have explained [it here](https://github.com/pipecat-ai/pipecat/pull/2004#discussion_r2157640738) an easier way to reproduce.

src/pipecat/processors/aggregators/llm_response.py
filipi87: Yes, this is exaclty what is happening, because we are reseting the state of `self._was_bot_speaking` inside `_handle_user_started_speaking`.

src/pipecat/processors/frame_processor.py
filipi87: @aconchillo I have removed these changes as we agreed during the meeting. 👍

src/pipecat/processors/aggregators/llm_response.py
filipi87: I have updated the `freeze-test` example, now if you execute it, you should be able to reproduce the issue where the bot was not speaking because the STT doesn’t return any transcription.

src/pipecat/processors/aggregators/llm_response.py
```diff
@@ -391,10 +409,12 @@ async def _handle_user_started_speaking(self, frame: UserStartedSpeakingFrame):
     async def _handle_user_stopped_speaking(self, _: UserStoppedSpeakingFrame):
         self._user_speaking = False
         # We just stopped speaking. Let's see if there's some aggregation to
-        # push. If the last thing we saw is an interim transcription, let's wait
-        # pushing the aggregation as we will probably get a final transcription.
-        if not self._seen_interim_results:
-            await self.push_aggregation()
+        # push.
+        if len(self._aggregation) > 0:
```
filipi87: I kept this if here so that `push_aggregation` is not invoked right away in this case, and instead waits until the timeout.

src/pipecat/processors/aggregators/llm_response.py
```diff
@@ -356,6 +357,21 @@ async def push_aggregation(self):
             else:
                 # No interruption config - normal behavior (always push aggregation)
                 await self._process_aggregation()
+        elif not self._user_speaking and self._was_bot_speaking:
```
markbackman: Do we need to check `not self._user_speaking`?

`push_aggregation` can be called from two places:
- `_handle_user_stopped_speaking`, which sets `self._user_speaking = False`
- `_aggregation_task_handler`, which has a check for `if not self._user_speaking:`

src/pipecat/processors/aggregators/llm_response.py
```diff
@@ -356,6 +357,21 @@ async def push_aggregation(self):
...
+            # Handles the case where both the user and the bot are not speaking,
+            # and the bot was previously speaking before the user interruption.
+            # Normally, when the user stops speaking, new text is expected,
+            # which triggers the bot to respond. However, if no new text
+            # is received, this safeguard ensures
+            # the bot doesn't hang indefinitely while waiting to speak again.
+            if not self._bot_speaking:
```
markbackman: Is this if condition required? Maybe just move `not self._bot_speaking` up to the elif?

src/pipecat/processors/aggregators/llm_response.py
markbackman: This makes sense.

src/pipecat/processors/aggregators/llm_response.py
```diff
@@ -355,6 +357,21 @@ async def push_aggregation(self):
             else:
                 # No interruption config - normal behavior (always push aggregation)
                 await self._process_aggregation()
+        elif not self._user_speaking and not self._seen_interim_results and self._was_bot_speaking:
```
markbackman: Do we need to check `not self._user_speaking`?

`push_aggregation` can only be called from `_aggregation_task_handler`, which has a check for `if not self._user_speaking:`

markbackman: Maybe this should be:
```
elif not self._seen_interim_results and self._was_bot_speaking and not self._bot_speaking:
```

Your comment makes it clear in which case this can occur.

filipi87: Yep, we can do it like this. I am going to to this change.

src/pipecat/processors/aggregators/llm_response.py
filipi87: done.

src/pipecat/processors/aggregators/llm_response.py
filipi87: done.

src/pipecat/processors/frame_processor.py
```diff
@@ -389,15 +387,17 @@ async def __input_frame_task_handler(self):
...
-            self.__input_queue.task_done()
+            try:
+                # Process the frame.
+                await self.process_frame(frame, direction)
+                # If this frame has an associated callback, call it now.
+                if callback:
+                    await callback(self, frame, direction)
+            except Exception as e:
+                logger.exception(f"{self}: error processing frame: {e}")
+                await self.push_error(ErrorFrame(str(e)))
```
aconchillo, line 421: I think this is good. Before we would just log the exception (inside `create_task`) but we would not be able to continue. Maybe this will help.

src/pipecat/transports/base_input.py
```diff
@@ -355,30 +363,42 @@ async def _run_turn_analyzer(
...
+
+                if self._params.turn_analyzer:
+                    await self._run_turn_analyzer(frame, vad_state, previous_vad_state)
+
+                # Push audio downstream if passthrough is set.
+                if self._params.audio_in_passthrough:
+                    await self.push_frame(frame)
+
+                self._audio_in_queue.task_done()
+            except asyncio.TimeoutError:
```
aconchillo, line 391: Oh, I was just thinking. What happens with, for example, websocket clients that don't send audio all the time? I think this might generate undesired `UserStoppedSpeakingFrame`?

filipi87: We are just triggering this if the user is speaking, so if we have receive an `UserStartedSpeakingFrame`

aconchillo: That is, this change assumes the transport is always generating audio frames, but that might not be always true.

src/pipecat/processors/frame_processor.py
filipi87: Exactly.

src/pipecat/transports/base_input.py
aconchillo: Actually, this would only be if the user was already speaking. So, that's probably fine...

PR #2005, opened 2025-06-13
add hume tts integration

ingridxx:
TTS request is valid but missing output file when running `test_hume_tts.py`

### DIFF
--- a/src/pipecat/services/hume/__init__.py
+++ b/src/pipecat/services/hume/__init__.py
@@ -1,0 +1,10 @@
+#
+# Copyright (c) 2024–2025, Daily
+#
+# SPDX-License-Identifier: BSD 2-Clause License
+#
+
+from .tts import HumeTTSService
+
+
+__all__ = ["HumeTTSService"]
--- a/src/pipecat/services/hume/tts.py
+++ b/src/pipecat/services/hume/tts.py
@@ -1,0 +1,131 @@
+#
+# Copyright (c) 2024–2025, Daily
+#
+# SPDX-License-Identifier: BSD 2-Clause License
+#
+
+import base64
+import json
+from typing import Any, AsyncGenerator, Dict, Optional
+
+import aiohttp
+from loguru import logger
+
+from pipecat.frames.frames import (
+    ErrorFrame,
+    Frame,
+    TTSAudioRawFrame,
+    TTSStartedFrame,
+    TTSStoppedFrame,
+)
+from pipecat.services.tts_service import TTSService
+from pipecat.utils.tracing.service_decorators import traced_tts
+
+
+
+class HumeTTSService(TTSService):
+    class UtteranceParams:
+        def __init__(
+            self,
+            description: Optional[str] = None,
+            speed: Optional[float] = None,
+            trailing_silence: Optional[float] = None,
+            voice_id: Optional[str] = None,
+            voice_provider: Optional[str] = None,
+        ):
+            self.description = description
+            self.speed = speed
+            self.trailing_silence = trailing_silence
+            self.voice_id = voice_id
+            self.voice_provider = voice_provider
+
+    def __init__(
+        self,
+        api_key: str,
+        aiohttp_session: aiohttp.ClientSession,
+        utterance_params: Optional[UtteranceParams] = None,
+        format_type: str = "mp3",
+        sample_rate: int = 48000,
+        **kwargs,
+    ):
+        super().__init__(
+            aggregate_sentences=True,
+            push_text_frames=False,
+            push_stop_frames=True,
+            sample_rate=sample_rate,
+            **kwargs,
+        )
+        self._api_key = api_key
+        self._session = aiohttp_session
+        self._utterance_params = utterance_params or self.UtteranceParams()
+        self._format_type = format_type
+        self._base_url = "https://api.hume.ai/v0/tts/stream/json"
+        self._cumulative_time = 0
+        self._started = False
+
+    def _build_utterance_dict(self, text: str) -> Dict[str, Any]:
+        utt = {"text": text}
+        if self._utterance_params.description:
+            utt["description"] = self._utterance_params.description
+        if self._utterance_params.speed:
+            utt["speed"] = self._utterance_params.speed
+        if self._utterance_params.trailing_silence:
+            utt["trailing_silence"] = self._utterance_params.trailing_silence
+        if self._utterance_params.voice_id:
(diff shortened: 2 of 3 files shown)


src/pipecat/services/hume/__init__.py
```diff
@@ -4,5 +4,7 @@
 # SPDX-License-Identifier: BSD 2-Clause License
 #
```
markbackman, line 5: You can leave this file blank. The pattern is to import from:

```
from pipecat.services.hume.tts import HumeTTSService
```

in application code.

src/pipecat/services/hume/tts.py
```diff
@@ -0,0 +1,131 @@
...
+    TTSStartedFrame,
+    TTSStoppedFrame,
+)
+from pipecat.services.tts_service import TTSService
+from pipecat.utils.tracing.service_decorators import traced_tts
+
+
+
+class HumeTTSService(TTSService):
+    class UtteranceParams:
```
markbackman, line 27: Nit: Maybe define this outside of the `HumeTTSService` class for ease of use?

src/pipecat/services/hume/tts.py
```diff
@@ -0,0 +1,131 @@
...
+            self.trailing_silence = trailing_silence
+            self.voice_id = voice_id
+            self.voice_provider = voice_provider
+
+    def __init__(
+        self,
+        api_key: str,
+        aiohttp_session: aiohttp.ClientSession,
+        utterance_params: Optional[UtteranceParams] = None,
+        format_type: str = "mp3",
```
markbackman, line 47: Do you support wav?

src/pipecat/services/hume/tts.py
```diff
@@ -0,0 +1,131 @@
...
+        sample_rate: int = 48000,
```
markbackman, line 48: You don't want to se the sample rate here. Instead, you want it to be set to None.

I wrote the following guidance recently on how this works:
```
### Sample Rate Handling

Sample rates are set via PipelineParams and passed to each frame processor at initialization. The pattern is to _not_ set the sample rate value in the constructor of a given service. Instead, use the `start()` method to initialize sample rates from the frame:

```python
    """Start the service."""
    await super().start(frame)
    self._settings["output_format"]["sample_rate"] = self.sample_rate
    await self._connect()
```

Note that `self.sample_rate` is a `@property` set in the TTSService base class, which provides access to the private sample rate value obtained from the StartFrame.
```

Check out other TTS services to see how the pattern is applied.

File: src/pipecat/services/hume/tts.py [hunk 5]
```diff
@@ -0,0 +1,131 @@
...
+        **kwargs,
+    ):
+        super().__init__(
```
markbackman, line 51: For an HTTP based TTS service, you want the following:
```
```

- aggregate_sentences: defaults to True, so no need to set
- push_text_frames: the TTS service should push text frames downstream. This is required for context aggregation. Default is True.
- push_stop_frames: You want the HumeTTSService to yield the frame instead. False is the default and desired value.

src/pipecat/services/hume/tts.py
```diff
@@ -0,0 +1,131 @@
...
+            push_text_frames=False,
+            push_stop_frames=True,
+            sample_rate=sample_rate,
+            **kwargs,
+        )
+        self._api_key = api_key
+        self._session = aiohttp_session
+        self._utterance_params = utterance_params or self.UtteranceParams()
+        self._format_type = format_type
+        self._base_url = "https://api.hume.ai/v0/tts/stream/json"
```
markbackman, line 62: In other services, the base_url is an arg. It could make sense here, too, for flexibility.

src/pipecat/services/hume/tts.py
```diff
@@ -0,0 +1,131 @@
...
+        self._cumulative_time = 0
```
markbackman, line 63: Remove since this is unused.
```suggestion
```

src/pipecat/services/hume/tts.py
```diff
@@ -0,0 +1,131 @@
...
+        self._started = False
```
markbackman, line 64: Remove since this is unused.

```suggestion
```

src/pipecat/services/hume/tts.py
```diff
@@ -0,0 +1,131 @@
...
+            async with self._session.post(self._base_url, headers=headers, json=payload) as response:
+                # print(json.dumps(payload, indent=2))
+                if response.status != 200:
+                    error_text = await response.text()
+                    yield ErrorFrame(error=f"Hume TTS API error: {error_text}")
+                    return
+
+                yield TTSStartedFrame()
+
+                async for line in response.content.iter_chunked(1024 * 1024):  # 1MB chunks
```
markbackman, line 109: From the TTSService class, there's a `chunk_size` `@property`.
```
    @property
    def chunk_size(self) -> int:
        """Get the recommended chunk size for audio streaming.

        This property indicates how much audio we download (from TTS services
        that require chunking) before we start pushing the first audio
        frame. This will make sure we download the rest of the audio while audio
        is being played without causing audio glitches (specially at the
        beginning). Of course, this will also depend on how fast the TTS service
        generates bytes.

        Returns:
            The recommended chunk size in bytes.
        """
        CHUNK_SECONDS = 0.5
        return int(self.sample_rate * CHUNK_SECONDS * 2)  # 2 bytes/sample
```

We've found this work well to avoid audio glitches in playback. It's helpful to use the property so we can uniformly adjust all HTTP based services.

test_hume_tts.py
```diff
@@ -0,0 +1,38 @@
+import asyncio
```
markbackman, line 1: Instead of a test, can you write a foundational example. Follow the [07-interruptible pattern](https://github.com/pipecat-ai/pipecat/blob/main/examples/foundational/07-interruptible.py).

markbackman:
Nice start! Sorry for the delayed feedback. We've been busy.

Four other things to do:
1. Update the [README](https://github.com/pipecat-ai/pipecat/blob/main/README.md) with your service.
2. Add env var to the [env.example](https://github.com/pipecat-ai/pipecat/blob/main/env.example).
3. Add docs: https://github.com/pipecat-ai/docs
4. Make sure to lint your code. You can install the pre-commit hook using `uv run pre-commit install` from the base of the repo.

markbackman:
Oops. Closing in favor of:
https://github.com/pipecat-ai/pipecat/pull/2518

PR #2008, opened 2025-06-14
fix groq wav file header parsing

kwindla:
Fix popping sound at the beginning of Groq TTS segments. The headers aren't simple 44-byte long basic structs, so we need to parse them dynamically rather than just chop off 44 bytes.


markbackman (approved):
LGTM!

PR #2009, opened 2025-06-15
Fix bug voice settings are never being sent to the Elevenlabs API

otaqwawi:
The voice settings block will never be executed because the condition if not self._context_id is never true at this point.
This appears to be a bug since voice settings are never being sent to the API. 

I'll propose a fix to ensure the voice settings are properly sent. The best place to include voice settings is in 
run_tts when we create the new context.

### DIFF
--- a/src/pipecat/services/elevenlabs/tts.py
+++ b/src/pipecat/services/elevenlabs/tts.py
@@ -428,26 +428,9 @@ class ElevenLabsTTSService(AudioContextWordTTSService):
                 break
 
     async def _send_text(self, text: str):
-        if self._websocket:
-            if not self._context_id:
-                # First message for a new context - need a space to initialize
-                msg = {"text": " ", "context_id": str(uuid.uuid4())}
-
-                # Add voice settings only in first message for a context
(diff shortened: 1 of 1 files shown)


markbackman (approved):
LGTM! Thanks for the fix.

Can you fix the limiting issues? You can use the pre-commit hooks from the [README](https://github.com/pipecat-ai/pipecat?tab=readme-ov-file#%EF%B8%8F-hacking-on-the-framework-itself).

Also, mind adding a changelog entry?

```
### Fixed

- Fixed an issue where voice settings weren't applied to ElevenLabsTTSService.
```

otaqwawi:
Yes sure, please @markbackman

markbackman:
@otaqwawi we'll likely release a new Pipecat version by EOW. Let me know if you have any questions. I'd love to get this change included.

otaqwawi:
Yes sir @markbackman, please go on

markbackman:
Thanks @otaqwawi. I'm implementing this in #2017 which has the linting fixes and changelog.

PR #2012, opened 2025-06-16
FrameProcessor: handle new FrameProcessorPauseFrame/FrameProcessorResumeFrame

aconchillo:

Sometimes it might be necessary to control if a frame processor is processing frame or not. Instead of calling methods directly (which is still possible) we just use frame as usual.


aconchillo:
Also added `FrameProcessorPauseUrgentFrame` and `FrameProcessorResumeUrgentFrame`. There was an issue in the previous PR. We should match the name of the processor.

PR #2020, opened 2025-06-18
Add Sambanova LLM and STT integration

snova-jorgep:
[SambaNova](https://cloud.sambanova.ai/) lntegration
allowing users to use [LLMs and STT](https://docs.sambanova.ai/cloud/docs/get-started/supported-models) from SambaNova Cloud 

docs PR: https://github.com/pipecat-ai/docs/pull/264


src/pipecat/services/sambanova/llm.py
```diff
@@ -0,0 +1,180 @@
...
+            functions_list.append(function_name)
+            arguments_list.append(arguments)
+            tool_id_list.append(tool_call_id)
+
+            function_calls = []
+
+            for function_name, arguments, tool_id in zip(
+                functions_list, arguments_list, tool_id_list
+            ):
+                # This allows compatibility until SambaNova API introduces indexing in tool calls.
```
markbackman, line 166: Can you explain what this limitation is?

src/pipecat/services/sambanova/llm.py
```diff
@@ -0,0 +1,180 @@
...
+        api_key: Optional[str] = None,
+        base_url: Optional[str] = None,
+        **kwargs: Dict[Any, Any],
+    ) -> Any:
+        """Create OpenAI-compatible client for SambaNova API endpoint."""
+
+        logger.debug(f"Creating SambaNova client with API {base_url}")
+        return super().create_client(api_key, base_url, **kwargs)
+
+    async def get_chat_completions(
```
markbackman, line 56: Are you overriding here to remove param options?

src/pipecat/services/sambanova/__init__.py
```diff
@@ -0,0 +1,14 @@
...
+#
+
+import sys
+
+from pipecat.services import DeprecatedModuleProxy
+
+from .llm import *
+from .stt import *
+
+sys.modules[__name__] = DeprecatedModuleProxy(globals(), "sambanova", "sambanova.[llm,stt,tts]")
```
markbackman: You can remove this since there is no prior version to deprecate.
```suggestion
```

markbackman: Instead, I think this can just be:
```
from .llm import *
from .stt import *
```

markbackman:
Awesome!! Just a few questions above. Also, a few additional requests:
1. Can you add a 14 series example, along the lines of https://github.com/pipecat-ai/pipecat/blob/main/examples/foundational/14-function-calling.py? This will make it easy to add to our eval suite for testing (and for discoverability for developers).
2. Can you add a 13 series example, along the lines of https://github.com/pipecat-ai/pipecat/blob/main/examples/foundational/13b-deepgram-transcription.py.
3. Can you add a changelog entry?

src/pipecat/services/sambanova/llm.py
snova-jorgep: Currently sambanova api for tool calling and streaming is enabled. It returns the tool call in one single chunk, and doesn't include a tool call index for this reason. When this is added on the api side, we will send a new PR to not override the _process_context method

src/pipecat/services/sambanova/llm.py
snova-jorgep: right to not pass parameters like  "frequency_penalty", "presence_penalty", and "seed" that are not available in the sambanova api

snova-jorgep:
@markbackman thanks! i have addressed your comments, including changelog addition, also added and tested the 13 and 14 foundational examples

markbackman (approved):
@snova-jorgep this looks really great! Nice work on this one 🙌 

Can you rebase on `main` to resolve the conflict? With that, this is ready to merge.

snova-jorgep:
@markbackman solved, thanks!

PR #2023, opened 2025-06-19
allow_interruptions=True

markbackman:

I think this is what we want. `allow_interruptions=False` is rarely used and arguably needs to be updated since the behavior is not what anyone expects.

### DIFF
--- a/examples/translation-chatbot/bot.py
+++ b/examples/translation-chatbot/bot.py
@@ -26,6 +26,7 @@ from pipecat.pipeline.pipeline import Pipeline
 from pipecat.pipeline.runner import PipelineRunner
 from pipecat.pipeline.task import PipelineParams, PipelineTask
 from pipecat.processors.aggregators.openai_llm_context import OpenAILLMContext
+from pipecat.processors.filters.stt_mute_filter import STTMuteConfig, STTMuteFilter, STTMuteStrategy
 from pipecat.processors.frame_processor import FrameDirection, FrameProcessor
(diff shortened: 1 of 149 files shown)


aconchillo (approved):

aconchillo:
Should we also update all of our examples?

filipi87 (approved):
Makes a lot of sense. 🚀

markbackman:
> Should we also update all of our examples?

Yup! I'll do this.

examples/translation-chatbot/bot.py
```diff
@@ -183,7 +193,6 @@ async def on_transcript_update(processor, frame):
         task = PipelineTask(
             pipeline,
             params=PipelineParams(
-                allow_interruptions=False,  # We don't want to interrupt the translator bot
```
markbackman, line 186: `allow_interruptions=False` doesn't work as expected, so replacing with the STTMuteFilter for now. We should really make `allow_interruptions=False` work like STTMuteFilter ALWAYS.

markbackman:
Ok, examples are all updated. I did the following:
- Removed `allow_interruptions=True`, where set as this is the default now
- Made sure examples include `enable_metrics=True` and `enable_usage_metrics=True` as this is useful for debugging latency
- Removed `report_only_initial_ttfb=True` since metrics are always useful for examples
- Where `allow_interruptions=False`, removed and replaced with STTMuteFilter ALWAYS where appropriate

PR #2030, opened 2025-06-19
MCPClient streamable_http transport support

yousifa:
- added streamablehttp support
- bumped mcp version to 1.9.4 (to support streamablehttp)
- added additional_headers and streamable_http params to MCPClient

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,29 +5,31 @@ All notable changes to **Pipecat** will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-## [Unreleased]
+## [0.0.72] - 2025-06-26
 
 ### Added
 
 - Added logging and improved error handling to help diagnose and prevent potential
   Pipeline freezes.
 
+- Added `WatchdogQueue`, `WatchdogPriorityQueue`, `WatchdogEvent` and
+  `WatchdogAsyncIterator`. These helper utilities reset watchdog timers
+  appropriately before they expire. When watchdog timers are disabled, the
+  utilities behave as standard counterparts without side effects.
+
 - Introduce task watchdog timers. Watchdog timers are used to detect if a
-  Pipecat task is taking longer than expected (by default 5 seconds). It is
+  Pipecat task is taking longer than expected (by default 5 seconds). Watchdog
+  timers are disabled by default and can be enabled globally by passing
+  `enable_watchdog_timers` argument to `PipelineTask` constructor. It is
   possible to change the default watchdog timer timeout by using the
-  `watchdog_timeout` constructor argument when creating a `PipelineTask`. With
-  watchdog timers it is also possible to log how long each processing step is
-  taking (e.g. processing an element from a queue inside a task). This is done
-  with the `enable_watchdog_logging` constructor argument when creating a
-  `PipelineTask.` It is also possible to control these two values per each frame
-  processor. That is, you can set set `enable_watchdog_logging` and
+  `watchdog_timeout` argument. You can also log how long it takes to reset the
+  watchdog timers which is done with the `enable_watchdog_logging`. You can
+  control all these settings per each frame processor or even per task. That is,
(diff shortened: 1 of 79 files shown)


vipyne:
thank you @yousifa !
do you have a favorite (accessible) mcp server that can be used to illustrate a `streamablehttp` connection ? 
perhaps something we can add to `examples/foundational/39-*` ?

vipyne (changes requested):
[Here is a branch](https://github.com/pipecat-ai/pipecat/compare/rebased-2030?expand=1) with some proposed changes. Most deal with recent SSE changes on `main`. Can you rebase on `main` and add these changes to this PR @yousifa ?
Feel free to tag @vipyne with any questions.

src/pipecat/services/mcp_service.py
```diff
@@ -148,14 +215,74 @@ async def mcp_tool_wrapper(
...
+            logger.debug(f"Executing tool '{function_name}' with call ID: {tool_call_id}")
+            logger.trace(f"Tool arguments: {json.dumps(arguments, indent=2)}")
+            try:
+                async with self._client(
+                    url=self._server_params.url,
+                    headers=self._server_params.headers,
+                    timeout=self._server_params.timeout,
+                    sse_read_timeout=self._server_params.sse_read_timeout,
+                    terminate_on_close=self._server_params.terminate_on_close,
+                    auth=self._server_params.auth,
```
vipyne: It seems `auth` is not added to `streamable ` yet.  I made [a PR here](https://github.com/modelcontextprotocol/python-sdk/pull/1028), but we'll see.  
For now, I say we remove it.  We could add it as an additional param in MCPClient class, but I'd rather wait until the spec settles a little.  Folks can change that themselves if they absolutely need that functionality right now.  Open to other opinions on this front though.

examples/foundational/39c-mcp-run-http.py
```diff
@@ -0,0 +1,131 @@
...
+    )
+
+    llm = OpenAILLMService(
+        api_key=os.getenv("OPENAI_API_KEY"), model="gpt-4o-mini"
+    )
+
+    try:
+        mcp = MCPClient(
+            server_params=StreamableHttpParameters(
+                url="https://api.githubcopilot.com/mcp/",
```
vipyne, line 72: I love this example! Thank you.
Can you point to docs for how to generate the `GITHUB_PERSONAL_ACCESS_TOKEN` and what properties the PAT needs to have?
This is what finally worked for me: https://stackoverflow.com/a/79219196 but you probably know best @yousifa

vipyne (changes requested):
- remove `auth=self._server_params.auth,` lines
- add comment about `GITHUB_PERSONAL_ACCESS_TOKEN`

after that, lgtm!

edit: oh, I still want to take a closer look at the google stuff... could we make the llm in the example google to exercise those code paths? @yousifa

examples/foundational/39c-mcp-run-http.py
yousifa: I couldn't actually find docs on this so it was trial and error, but i'll write it in comments

src/pipecat/services/mcp_service.py
yousifa: or we can just pass everything in _server_params as **kwargs instead of explicitly passing?

yousifa:
@vipyne can we get this in before v0.0.72 👀🙏

vipyne:
> @vipyne can we get this in before v0.0.72 👀🙏

@yousifa can you rebase on latest main by any chance? just makes things a little cleaner.
now worries if not. reviewing now.  will see re: release.

vipyne:
so as I am looking at this, I think the google llm stuff is out of scope here.  can you open a separate PR with just those changes?  that way this PR & code changes are only about adding the streamable http client.

yousifa:
sure will do now. I will change the example to be with openai then so it doesn't break

vipyne:
hi @yousifa, I think something went sideways. can you try pulling latest main, then rebasing?

vipyne:
feel free to ping @vipyne on discord and I can help.

yousifa:
yep, i rebased on old main i think, working on it

yousifa:
@vipyne should be good to go

PR #2031, opened 2025-06-19
FishAudioTTSService: adds normalize to InputParams and allows selection of control model

byron-the-bulb:
Added 'normalize' to InputParams which is necessary to use the paralanguage and phoneme controls (needs to be set to 'false', according to official docs)
Added the ability to select the voice control TTS model (speech-1.5, speech-1.6,s1, etc.). Unfortunately Fish Audio made the brilliant decision to call their voices and their AI models both "models"... This parameter allows you to select the latter. If not specified it will use whatever the default TTS model the API is set for (unclear from Fish Audio documentation)

### DIFF
--- a/src/pipecat/services/fish/tts.py
+++ b/src/pipecat/services/fish/tts.py
@@ -42,6 +42,7 @@ class FishAudioTTSService(InterruptibleTTSService):
     class InputParams(BaseModel):
         language: Optional[Language] = Language.EN
         latency: Optional[str] = "normal"  # "normal" or "balanced"
+        normalize: False
         prosody_speed: Optional[float] = 1.0  # Speech speed (0.5-2.0)
         prosody_volume: Optional[int] = 0  # Volume adjustment in dB
 
@@ -50,6 +51,7 @@ class FishAudioTTSService(InterruptibleTTSService):
         *,
         api_key: str,
         model: str,  # This is the reference_id
+        control_model : str = None,
         output_format: FishAudioOutputFormat = "pcm",
         sample_rate: Optional[int] = None,
         params: Optional[InputParams] = None,
@@ -65,6 +67,7 @@ class FishAudioTTSService(InterruptibleTTSService):
         params = params or FishAudioTTSService.InputParams()
 
         self._api_key = api_key
+        self._control_model = control_model
         self._base_url = "wss://api.fish.audio/v1/tts/live"
         self._websocket = None
         self._receive_task = None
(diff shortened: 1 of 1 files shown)


src/pipecat/services/fish/tts.py
```diff
@@ -42,6 +42,7 @@ class FishAudioTTSService(InterruptibleTTSService):
     class InputParams(BaseModel):
         language: Optional[Language] = Language.EN
         latency: Optional[str] = "normal"  # "normal" or "balanced"
+        normalize: False
```
markbackman: Should be:

```suggestion
        normalize: Optional[bool] = False
```

Though, I don't see any reference to `normalize` in the websocket API docs. I only see it referenced in the HTTP API docs. I tested with it but I can't really see a difference. Can you point to the docs that confirm this parameter is supported?

src/pipecat/services/fish/tts.py
```diff
@@ -50,6 +51,7 @@ def __init__(
         *,
         api_key: str,
         model: str,  # This is the reference_id
+        control_model : str = None,
```
markbackman, line 54: Setting this to None would pass in the None value to the header which doesn't seem right. Instead, this should either be programmatically included when set or set to a default value, like `speech-1.5`.

```suggestion
        control_model : str = None,
```

src/pipecat/services/fish/tts.py
byron-the-bulb: https://docs.fish.audio/text-to-speech/fine-grained-control 
The docs are a bit unclear in regards to WebSockets, when I used AskAI it recommended to set it to false.
It definitely made a difference for me, before it would read the paralanguage cues out loud, after setting it to false it applied them.

Missed the type, sorry, I'll resubmit

markbackman: I've reached out to the Fish team to confirm whether or not this is supported. In the past they've been quick to reply. Let's see what they say.

markbackman: Ok! The Fish team confirmed that normalize is supported by the websocket API.

markbackman:
@byron-the-bulb if you can make the suggested changes, I think this PR will be ready to merge.

src/pipecat/services/fish/tts.py
byron-the-bulb: Great. I have also been experimenting a bit more. In my tests their latest 's1' control model with normalize=false provided the best emotional and paralanguage support.

byron-the-bulb:
> @byron-the-bulb if you can make the suggested changes, I think this PR will be ready to merge.

Done

markbackman:
I'm closing this in favor of #2137 as there's some additional clean up to do to align naming.

PR #2034, opened 2025-06-20
small fix for processor pause/resume frames

kwindla:
frame.processor.name rather than frame.name

### DIFF
--- a/src/pipecat/processors/frame_processor.py
+++ b/src/pipecat/processors/frame_processor.py
@@ -296,11 +296,11 @@ class FrameProcessor(BaseObject):
         await self.__cancel_push_task()
 
     async def __pause(self, frame: FrameProcessorPauseFrame | FrameProcessorPauseUrgentFrame):
-        if frame.name == self.name:
+        if frame.processor.name == self.name:
             await self.pause_processing_frames()
 
     async def __resume(self, frame: FrameProcessorResumeFrame | FrameProcessorResumeUrgentFrame):
-        if frame.name == self.name:
+        if frame.processor.name == self.name:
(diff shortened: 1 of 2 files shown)


src/pipecat/processors/frame_processor.py
```diff
@@ -296,11 +296,11 @@ async def __cancel(self, frame: CancelFrame):
         await self.__cancel_push_task()
 
     async def __pause(self, frame: FrameProcessorPauseFrame | FrameProcessorPauseUrgentFrame):
-        if frame.name == self.name:
+        if frame.processor.name == self.name:
```
aconchillo, line 299: it would actually be `frame.processor`. Maybe the field should be `process_name` to avoid confusion.

aconchillo: I just updated the PR. Let me know if it makes sense.

kwindla: Hmm. It feels a little cleaner to me to construct the frame just by passing it the processor. Like this:

```
FrameProcessorPauseFrame(processor=tts)
```

Rather than

```
FrameProcessorPauseFrame(processor_name=tts.name)

```

But maybe I'm misunderstanding. And either way, I defer to your judgment.

filipi87 (approved):

src/pipecat/processors/frame_processor.py
aconchillo: I agree. I just avoided using `FrameProcessor` for the circular imports. But you can actually avoid that with `TYPE_CHECKING`.

PR #2036, opened 2025-06-20
Speechmatics STT + multi-speaker conversations

sam-s10s:
This is an initial release of the Speechmatics STT plugin for pipecat. It not only takes advantage of our fast STT engine, but also introduces the speaker diarization - this means the agent knows who is speaking in multi-speaker conversations.

User: `<S1>Hello, it's John here!</S1> <S2>And Emma!</S2>`
Assistant: `Hello both of you!`
User: `<S2>Who is speaking now?</S2>`
Assistant: `That's Emma!`

By providing instructions in the system context, it is possible for larger LLMs to be able to deduce from the speaker tags who is who. With Speechmatics, you can also pass in known speaker voice-prints and this will mean the STT and LLM will know who is speaking by name and not just an index.


examples/foundational/07a-interruptible-speechmatics.py
```diff
@@ -0,0 +1,128 @@
+#
```
markbackman, line 1: 07a is helpful for testing. Do you mind also adding a 13* series for just the raw transcription?

src/pipecat/services/speechmatics/README.md
```diff
@@ -0,0 +1,30 @@
+# Speechmatics STT Service
+
+NOTES TO GO HERE!
```
markbackman: I think we can remove this.

You can submit docs for the service here:
https://github.com/pipecat-ai/docs

I'd follow the Deepgram example, as that's most up to date: https://docs.pipecat.ai/server/services/stt/deepgram

src/pipecat/services/speechmatics/__init__.py
```diff
@@ -0,0 +1,13 @@
+#
```
markbackman, line 1: This can be simplified to:
```
#
# Copyright (c) 2024–2025, Daily
#
#


from .stt import *
```

The deprecation notice was only needed for older services before we changed the package structure.

src/pipecat/services/speechmatics/README.md
markbackman: Please make sure the docs path matches `/server/services/stt/speechmatics` to align with the README entry.

src/pipecat/services/speechmatics/stt.py
```diff
@@ -0,0 +1,608 @@
...
+    attaches_to: str = ""
+    content: str = ""
+    speaker: Optional[str] = None
+    confidence: float = 1.0
+    result: Optional[Any] = None
+
+
+@dataclass
+class SpeechFrame(TranscriptionFrame):
+    """TranscriptionFrame with speaker_id.
```
markbackman: We auto-generate reference docs. Please make sure to read this guide on docstrings:
https://github.com/pipecat-ai/pipecat/blob/main/CONTRIBUTING.md#docstring-conventions

src/pipecat/services/speechmatics/stt.py
```diff
@@ -0,0 +1,608 @@
+#
+# Copyright (c) 2024–2025, Daily
+#
+# SPDX-License-Identifier: BSD 2-Clause License
+#
+
+import asyncio
+import datetime
+import re
+from dataclasses import asdict, dataclass
```
markbackman: There are a few unused imports. Please remove them.

Also, please use a try/except for importing packages. The imports section should look like:
```
import asyncio
import datetime
import re
from typing import Any, AsyncGenerator, Optional
from urllib.parse import urlencode

from loguru import logger

    CancelFrame,
    EndFrame,
    Frame,
    InterimTranscriptionFrame,
    StartFrame,
    TranscriptionFrame,
)
from pipecat.services.stt_service import STTService

try:
    from speechmatics.rt import (
        AsyncClient,
        AudioEncoding,
        AudioFormat,
        OperatingPoint,
        ServerMessageType,
        TranscriptionConfig,
        __version__,
    )
    from speechmatics.rt._models import ConversationConfig, SpeakerDiarizationConfig
    logger.error(
        "In order to use Speechmatics, you need to `pip install pipecat-ai[speechmatics]`."
    )
```

src/pipecat/services/speechmatics/stt.py
```diff
@@ -0,0 +1,608 @@
...
+            Defaults to `None`.
+        confidence: Confidence of the fragment (0.0 to 1.0).
+            Defaults to `1.0`.
+        result: Raw result of the fragment.
+            Defaults to `None`.
+    """
+
+    start_time: float
+    end_time: float
+    language: str = "en"
```
markbackman: Let's use Pipecat's Language module in place of strings.

```suggestion
    language: str = Language.EN
```

src/pipecat/services/speechmatics/stt.py
```diff
@@ -0,0 +1,608 @@
...
+    add custom markup for the speaker within the transcript for the LLM to use
+    to identify different speakers in a multi-speaker conversation.
+
+    Attributes:
+        text: The text of the frame.
+        user_id: The ID string.
+        timestamp: The timestamp of the frame.
+        language: The language of the frame.
+        result: The result of the frame.
+        speaker_id: The ID of the speaker.
```
markbackman: How is speaker_id different from user_id? `user_id` is intended for use by the user being transcribed. Maybe we can eliminate the need for this subclass and just use the TranscriptionFrame?

Same question about `text` vs `result`.

src/pipecat/services/speechmatics/stt.py
```diff
@@ -0,0 +1,608 @@
...
+            Defaults to `None`.
+        **kwargs: Additional arguments passed to STTService.
+    """
+
+    def __init__(
+        self,
+        *,
+        api_key: str,
+        language: Language = Language.EN,
+        base_url: str = "eu2.rt.speechmatics.com",
```
markbackman: In docs, I see `wss://eu2.rt.speechmatics.com/v2` referenced as the websocket API (in a few places):
- https://docs.speechmatics.com/rt-api-ref
- https://docs.speechmatics.com/introduction/rt-guide#real-time-transcription-examples

Should it be v2?

src/pipecat/services/speechmatics/stt.py
sam-s10s: I am working on this and realised that the docs were missing.

src/pipecat/services/speechmatics/stt.py
sam-s10s: Of course and I will sort this, as suggested.

src/pipecat/services/speechmatics/stt.py
sam-s10s: That's my mistake.

src/pipecat/services/speechmatics/stt.py
sam-s10s: Speaking with the Pipecat team earlier, we discussed the use of `user_id` for this purpose. This usually refers to the channel participant, but in our case there can be multiple speakers in the same audio channel. The `speaker_id` was used for the wrapping of the text with the speaker reference for the LLM.

sam-s10s: is `text` not the text for the frame and `result` the raw payload(s) for the words?

src/pipecat/services/speechmatics/stt.py
sam-s10s: There is a utility function to prepare the wss URL. Let me review this for consistency.

src/pipecat/services/speechmatics/stt.py
```diff
@@ -0,0 +1,608 @@
...
+    InterimTranscriptionFrame,
+    StartFrame,
+    TranscriptionFrame,
+)
+from pipecat.services.stt_service import STTService
+from pipecat.transcriptions.language import Language
+from pipecat.utils.time import time_now_iso8601
+
+
+class AudioBuffer:
```
sam-s10s, line 50: Our Real-Time Python SDK requests a fixed number of frames. The audio buffer is used to consume the frames from Pipecat and then deliver these as frames become available. If no audio is available, then the transcription may come to a premature end, so this ensures that the STT is fed with audio as it becomes available.

examples/foundational/07a-interruptible-speechmatics.py
```diff
@@ -0,0 +1,128 @@
...
+}
+
+
+async def run_example(transport: BaseTransport, _: argparse.Namespace, handle_sigint: bool):
+    """Run example using Speechmatics STT."""
+    logger.info(f"Starting bot")
+
+    stt = SpeechmaticsSTTService(
+        api_key=os.getenv("SPEECHMATICS_API_KEY"),
+        base_url=os.getenv("SPEECHMATICS_ENDPOINT", None),
```
markbackman: Maybe either omit or add `eu2.rt.speechmatics.com` as the default option.

```suggestion
        base_url=os.getenv("SPEECHMATICS_ENDPOINT", "eu2.rt.speechmatics.com"),
```

markbackman:
This works great! I left some feedback.

One additional thing. I notice that `TranscriptionFrame`s are pushed as:
```
text: '<S1>Hey, can you hear me?</S1>', user_id: '', timestamp: '2025-06-27T21:02:01.245+00:00', language: 'en'
```

resulting in context messages like this:
```
{"role": "user", "content": "<S1>Can you tell me a joke?</S1>"}
```

We don't have an established pattern for multi-user input in the context just yet. So, for now, I would suggest leaving out the <S1></S1> tags. Instead, I would encourage adding any user identifiers to the TranscriptionFrame's `user_id` field. Also, I'd recommend sticking to the `TranscriptionFrame` instead of creating a subclass. Pipecat uses the `TranscriptionFrame` in its context aggregation, which is critical to providing input to the LLM.

sam-s10s:
@markbackman - thank you for the helpful comments.

As you may notice in the example `examples/foundational/07a-interruptible-speechmatics.py`, we tell the LLM that individual speakers are identified / wrapped in the `<Sn/>` tags, as you show in your example. If there are multiple speakers within the same turn, then this is correctly showing them as `<S1/><S2/><S1/>` for the turn.

The team did mention that could implement a hook for the point at which the frames are transferred to the LLM, so this may be a better place to do the wrapping so we don't sub-class?

The key is that the speaker id is handed to the LLM, otherwise the content of who is speaking may be lost.

markbackman:
> The team did mention that could implement a hook for the point at which the frames are transferred to the LLM, so this may be a better place to do the wrapping so we don't sub-class?

Pipecat uses the `TranscriptionFrame` as input to the `LLMUserContextAggregator`. We should stick to this pattern. If you populate the `user_id` param in the `TranscriptionFrame`, then we can add logic to the `LLMUserContextAggregator` to handle the multi-speaker case. The benefits are:
-  `TranscriptionFrame` remains the data object handling the user input
- We can uniformly handle multiple users without special prompting in the LLM

sam-s10s:
> * We can uniformly handle multiple users without special prompting in the LLM

LLMs need to be aware of _who said _what_ so that they can use this to give context to the conversation. For example, if one speaker says their name is `Sam` and another `Mark`, then the LLM knows who is speaking. This dramatically enhances the interactions.

After extensive research into this, we have found that LLMs are very good at understanding XML – way better than JSON – and by providing them with the guidance (such as `Use '<Sn/>' tags to identify different speakers - do not use tags in your replies.`), then they nail the multi-speaker situation.

I could go on about this, but I do firmly believe that we must retain the speaker id in the messages sent to the LLM.

src/pipecat/services/speechmatics/stt.py
```diff
@@ -0,0 +1,608 @@
...
+                content += " " + frag.content
+            results.append(frag.result)
+
+        # Timestamp
+        ts = (self._start_time + datetime.timedelta(seconds=start_time)).isoformat(
+            timespec="milliseconds"
+        )
+
+        # Return the SpeechData object
+        return SpeechFrame(
```
markbackman: For a frame to be used, it needs to be pushed down the Pipeline. Right now, the SpeechFrame is used only as an internal object.

I think this point is moot as we should continue using TranscriptionFrame, but just mentioning in case we change course.

markbackman: Oh, I see...

Actually, SpeechFrame is just the speech fragment object. I think using `SpeechFrame` as a name is confusing. Maybe `SpeechFragment` is more appropriate.

A `Frame` is a special class in Pipecat. Frames are pushed by FrameProcessors through a Pipeline. So, we should only use `Frame` in a class name when it matches the base `Frame` class definition.

Given that, I don't think this should subclass `TranscriptionFrame`. That will help to clarify the intent too.

markbackman:
> LLMs need to be aware of _who said what so that they can use this to give context to the conversation. For example, if one speaker says their name is Sam and another Mark, then the LLM knows who is speaking. This dramatically enhances the interactions.

Indeed! By providing the speaker information in the TranscriptionFrame via `user_id`, the context_aggregator can do this work on behalf of the developer.

> After extensive research into this, we have found that LLMs are very good at understanding XML – way better than JSON – and by providing them with the guidance (such as Use '<Sn/>' tags to identify different speakers - do not use tags in your replies.), then they nail the multi-speaker situation.

That makes sense, and I've seen the same with XML in responses. The one challenge is that >95% (probably higher) of voice AI use is one user to one bot, so we don't want to add a burden to every developer using SpeechmaticsSTTService with adding a mention of <Sn> tag in their prompt, even if it doesn't applies. Conversely, if the context_aggregator (or maybe based STTService class) does this on their behalf, then we can add an arg, like enable_multiuser_input (name TBD), which would then add tags to the messages by extracting the user_id from the TranscriptionFrame. This is an API we need to sort out, but this would provide Pipecat developers more control. WDYT?

sam-s10s:
@markbackman - I have gone through your comments and updated the draft PR. I will continue to test this and also add in other example(s) and check the documentation.

sam-s10s:
@markbackman - following on from Friday, I have made some updates to the PR. I have also added the ability to create an STT session either using a Language or a language string. The reason for this approach is that we have some language codes that combine languages (such as `cmn_en`), so this caters for those use cases. There are also two new languages, `Esperanto` and `Uyghur`, so I have added these to the main Language class (if such a thing is permitted?).

sam-s10s:
@markbackman - I have updated the PR. I have reduced the `aggregation_timeout` for the LLM to be `0.005s`, as the default 0.5s was causing a ... 500ms delay before going to the LLM.

src/pipecat/services/speechmatics/stt.py
```diff
@@ -0,0 +1,772 @@
...
+    from speechmatics.rt import (
+        AsyncClient,
+        AudioEncoding,
+        AudioFormat,
+        OperatingPoint,
+        ServerMessageType,
+        TranscriptionConfig,
+        __version__,
+    )
+    from speechmatics.rt._models import ConversationConfig, SpeakerDiarizationConfig
```
dumitrugutu: In latest rt sdk version (0.3.0) you can import `ConversationConfig, SpeakerDiarizationConfig` from `speechmatics.rt`

examples/foundational/07a-interruptible-speechmatics.py
```diff
@@ -0,0 +1,139 @@
...
+async def run_example(transport: BaseTransport, _: argparse.Namespace, handle_sigint: bool):
+    """Run example using Speechmatics STT."""
+    logger.info(f"Starting bot")
+
+    stt = SpeechmaticsSTTService(
+        api_key=os.getenv("SPEECHMATICS_API_KEY"),
+        language=Language.EN,
+        output_locale=Language.EN_GB,
+        end_of_utterance_silence_trigger=0.5,
+        enable_speaker_diarization=True,
```
dumitrugutu, line 77: Use enhanced operating_point -- let users try our best models :)

src/pipecat/services/speechmatics/stt.py
```diff
@@ -0,0 +1,772 @@
...
+        super().__init__(sample_rate=sample_rate, **kwargs)
+
+        # Client configuration
+        self._api_key: str = api_key
+        self._language: Optional[Language] = language
+        self._language_code: Optional[str] = language_code
+        self._base_url: str = base_url
+        self._domain: Optional[str] = domain
+        self._output_locale: Optional[Language] = output_locale
+        self._output_locale_code: Optional[str] = None
```
filipi87: I guess we need to assign the output_locale_code here ?
```suggestion
        self._output_locale_code: Optional[str] = output_locale_code
```

src/pipecat/services/speechmatics/stt.py
```diff
@@ -0,0 +1,772 @@
...
+        # Final transcript event
+        @self._client.on(ServerMessageType.ADD_TRANSCRIPT)
+        def _evt_on_final_transcript(message: dict[str, Any]):
+            self._handle_transcript(message, is_final=True)
+
+        # End of Utterance
+        @self._client.on(ServerMessageType.END_OF_UTTERANCE)
+        def _evt_on_end_of_utterance(message: dict[str, Any]):
+            logger.debug("End of utterance received from STT")
+            asyncio.create_task(self._send_frames(finalized=True))
```
filipi87: Instead of creating a new thread, I believe we should do like this:

```suggestion
            asyncio.run_coroutine_threadsafe(self._send_frames(finalized=True), self.get_event_loop())
```

src/pipecat/services/speechmatics/stt.py
```diff
@@ -0,0 +1,772 @@
...
+            message=message,
+            is_final=is_final,
+        )
+
+        # Skip if unchanged
+        if not has_changed:
+            return
+
+        # Send frames
+        asyncio.create_task(self._send_frames())
```
filipi87: Same here:
```suggestion
        asyncio.run_coroutine_threadsafe(
            self._send_frames(), self.get_event_loop()
        )
```

src/pipecat/services/speechmatics/stt.py
```diff
@@ -0,0 +1,772 @@
...
+            self._handle_transcript(message, is_final=True)
+
+        # End of Utterance
+        @self._client.on(ServerMessageType.END_OF_UTTERANCE)
+        def _evt_on_end_of_utterance(message: dict[str, Any]):
+            logger.debug("End of utterance received from STT")
+            asyncio.create_task(self._send_frames(finalized=True))
+
+        # Start the client in a thread
+        asyncio.create_task(self._run_client())
```
filipi87: Here it looks like we actually need to create a task, but probably best use the taskManager provided by Pipecat. So you would need to create like this:
```suggestion
        self._client_task = self.create_task(self._run_client())
```

src/pipecat/services/speechmatics/stt.py
```diff
@@ -0,0 +1,772 @@
...
+
+    async def _disconnect(self) -> None:
+        """Disconnect from the STT service."""
+        # Stop the audio buffer
+        self._audio_buffer.stop()
+
+        # Disconnect the client
+        if self._client:
+            await self._client.close()
```
filipi87: And after disconnecting, making sure that the task is destroyed:

```suggestion
        # Disconnect the client
        if self._client:
            await self._client.close()
            
        if self._client_task:
            await self.cancel_task(self._client_task)
            self._client_task = None     
```

markbackman:
@sam-s10s I'm doing some final testing and I've found that when running the 07a and 13h examples that I can't cancel the python process running those files. I think this may be due to a task that isn't closing properly. I haven't looked yet, but I wanted to flag it.

Repro is:
- Run the example
- While running, Ctrl+C to cancel it

examples/foundational/07a-interruptible-speechmatics.py
```diff
@@ -0,0 +1,140 @@
...
+    """Run example using Speechmatics STT."""
+    logger.info(f"Starting bot")
+
+    stt = SpeechmaticsSTTService(
+        api_key=os.getenv("SPEECHMATICS_API_KEY"),
+        language=Language.EN,
+        output_locale=Language.EN_GB,
+        operating_point=OperatingPoint.ENHANCED,
+        end_of_utterance_silence_trigger=0.5,
+        enable_speaker_diarization=True,
```
markbackman, line 77: It might be useful to add a comment stating something along the lines of: diarization is useful for multi-party calls.

examples/foundational/07a-interruptible-speechmatics.py
```diff
@@ -0,0 +1,140 @@
...
+async def run_example(transport: BaseTransport, _: argparse.Namespace, handle_sigint: bool):
+    """Run example using Speechmatics STT."""
+    logger.info(f"Starting bot")
+
+    stt = SpeechmaticsSTTService(
+        api_key=os.getenv("SPEECHMATICS_API_KEY"),
+        language=Language.EN,
+        output_locale=Language.EN_GB,
+        operating_point=OperatingPoint.ENHANCED,
+        end_of_utterance_silence_trigger=0.5,
```
markbackman: I see that you set `end_of_utterance_silence_trigger` to 0.5. Is this just to show how the API can be used? The default value is set to None. Does that use a Speechmatics default? If so, what is that value? Also, perhaps you provide or link to Speechmatics docs on how to set that value (maybe in the stt.py args).

examples/foundational/07a-interruptible-speechmatics.py
```diff
@@ -0,0 +1,140 @@
...
+        vad_analyzer=SileroVADAnalyzer(),
+    ),
+}
+
+
+async def run_example(transport: BaseTransport, _: argparse.Namespace, handle_sigint: bool):
+    """Run example using Speechmatics STT."""
+    logger.info(f"Starting bot")
+
+    stt = SpeechmaticsSTTService(
```
markbackman, line 74: We often link to these examples to recommend the preferred way to configure a service. Rather than show how to exercise the API, it could be useful to set the example with recommended values. If that's what this is already, then great!

src/pipecat/services/speechmatics/stt.py
```diff
@@ -0,0 +1,787 @@
...
+
+class AudioBuffer:
+    """Audio buffer for STT clients.
+
+    The Python SDK expects audio in a pre-defined number of frames. This
+    buffer will accumulate the data from the pipeline and provide it to the
+    STT client in the correct lengths, waiting for the number of frames to
+    be available.
+
+    Args:
```
markbackman: Sorry for changing the rules on you, but we've made a bunch of docstring formatting changes, which are now in the [CONTRIBUTING](https://github.com/pipecat-ai/pipecat/blob/main/CONTRIBUTING.md#code-style-and-documentation) guide. This work was happening concurrently with your development, so apologies that we switched things up on you.

I think the main thing is to move the `Args` docstrings to `def __init__`. With that, linting should pass.

src/pipecat/services/speechmatics/stt.py
```diff
@@ -0,0 +1,787 @@
...
+    def stop(self) -> None:
+        """Close the audio buffer."""
+        self._closed = True
+
+
+@dataclass
+class SpeechFragment:
+    """Fragment of an utterance.
+
+    Attributes:
```
markbackman: dataclasses should list members as `Parameters`, which is mainly to do with how sphinx and autodoc generate HTML. It's a major rough edge of working with the docs autogeneration.

```suggestion
    Parameters:
```

src/pipecat/services/speechmatics/stt.py
```diff
@@ -0,0 +1,787 @@
...
+            return
+
+        # If final, then re=parse into TranscriptionFrame
+        if finalized:
+            # Reset the speech fragments
+            self._speech_fragments.clear()
+
+            # Transform frames
+            frames = [
+                TranscriptionFrame(**frame._as_frame_attributes(self._text_format))
```
markbackman, line 527: For OpenTelemetry tracing, the SpeechmaticsSTTService class needs a separate method that can be traced. This can be something along these lines:
https://github.com/pipecat-ai/pipecat/blob/main/src/pipecat/services/deepgram/stt.py#L262-L267

Then you would just call that method where you're pushing the `TranscriptionFrame`.

src/pipecat/services/speechmatics/__init__.py
aconchillo: We should actually keep the file empty and remove all the lines, because otherwise we are allowing to import `pipecat.services.speechmatics`. We want to force people to use `pipecat.services.speechmatics.stt` for all new services.

sam-s10s:
@markbackman - are you able to give me any sort of stack trace on the issue you mentioned about the python process locking / not releasing? I don't see the same issue on my machine, which is typical! :)

PR #2037, opened 2025-06-20
Fix: Correctly close the context for ElevenLabsTTSService

markbackman:

This simplifies the ElevenLabsTTSService context management. The goal of this PR is to close the context in one place only, along with the corresponding turn state (`_context_id` and `_started`).

The previous issue was that the `_context_id` was being cleared when flushing (race condition!) which caused the logic in `_handle_interruption` to not run. Due to a limitation in how 11Labs expects contexts to be managed (<5 per session), the context count was growing and would exceed the limit.

Now, context management is simpler to track, as interruptions do the work. This feels like a safe approach since there's an interruption (i.e. StartInterruptionFrame) that occurs every time a user speaks. This will always precede the TTS response.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -41,6 +41,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- Fixed an issue with `ElevenLabsTTSService` where the context was not being
+  closed.
+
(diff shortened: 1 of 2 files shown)


filipi87 (approved):
Based on how you explained it during the meeting, it makes a lot of sense. Nice cleanup!

aconchillo (approved):

PR #2048, opened 2025-06-23
Fix missing video_in_enabled in vision bot.py for Moondream template

thibaudbrg:
The parameter video_in_enabled=True was missing in DailyParams, which prevented image capture  from working. Without this parameter, UserImageRequestFrame would be sent but no actual image data would be captured from participants.

This fix enables the "Let me take a look" functionality to work as  intended by allowing the transport to capture video frames for vision processing with Moondream.


markbackman (approved):
Nice catch! 🕵️

PR #2049, opened 2025-06-23
introduce watchdog timers

aconchillo:

From the CHANGELOG:

Introduce task watchdog timers. Watchdog timers are used to detect if a Pipecat task is taking longer than expected (by default 5 seconds). It is possible to change the default watchdog timer timeout by using the `watchdog_timeout` constructor argument when creating a `PipelineTask`. With watchdog timers it is also possible to log how long each processing step is taking (e.g. processing an element from a queue inside a task). This is done with the `enable_watchdog_logging` constructor argument when creating a `PipelineTask.` It is also possible to control these two values per each frame processor. That is, you can set set `enable_watchdog_logging` and `watchdog_timeout` when creating any frame processor through their constructor arguments. Finally, you can also set these values per task. So, if you are writing a frame processor that creates multiple tasks and you only want to enable logging for one of them, you can do so by passing the same argument names to the `FrameProcessor.create_task()` function. Note that watchdog timers only work with Pipecat tasks but not if you use `asycio.create_task()` or similar.

### DIFF
--- a/src/pipecat/pipeline/task.py
+++ b/src/pipecat/pipeline/task.py
@@ -6,7 +6,7 @@
 
 import asyncio
 import time
-from typing import Any, AsyncIterable, Dict, Iterable, List, Optional, Sequence, Tuple, Type
+from typing import Any, AsyncIterable, Dict, Iterable, List, Optional, Tuple, Type
 
 from loguru import logger
 from pydantic import BaseModel, ConfigDict, Field
@@ -33,10 +33,10 @@ from pipecat.metrics.metrics import ProcessingMetricsData, TTFBMetricsData
 from pipecat.observers.base_observer import BaseObserver
 from pipecat.observers.turn_tracking_observer import TurnTrackingObserver
 from pipecat.pipeline.base_pipeline import BasePipeline
-from pipecat.pipeline.base_task import BaseTask
+from pipecat.pipeline.base_task import BasePipelineTask, PipelineTaskParams
 from pipecat.pipeline.task_observer import TaskObserver
 from pipecat.processors.frame_processor import FrameDirection, FrameProcessor, FrameProcessorSetup
-from pipecat.utils.asyncio import BaseTaskManager, TaskManager
+from pipecat.utils.asyncio import WATCHDOG_TIMEOUT, BaseTaskManager, TaskManager, TaskManagerParams
 from pipecat.utils.tracing.setup import is_tracing_available
 from pipecat.utils.tracing.turn_trace_observer import TurnTraceObserver
 
@@ -45,7 +45,10 @@ HEARTBEAT_MONITOR_SECONDS = HEARTBEAT_SECONDS * 10
 
 
 class PipelineParams(BaseModel):
-    """Configuration parameters for pipeline execution.
+    """Configuration parameters for pipeline execution. These parameters are
+    usually passed to all frame processors using through `StartFrame`. For other
+    generic pipeline task parameters use `PipelineTask` constructor arguments
+    instead.
 
     Attributes:
         allow_interruptions: Whether to allow pipeline interruptions.
@@ -60,6 +63,7 @@ class PipelineParams(BaseModel):
         send_initial_empty_metrics: Whether to send initial empty metrics.
         start_metadata: Additional metadata for pipeline start.
         interruption_strategies: Strategies for bot interruption behavior.
+
     """
 
     model_config = ConfigDict(arbitrary_types_allowed=True)
@@ -71,11 +75,11 @@ class PipelineParams(BaseModel):
     enable_metrics: bool = False
     enable_usage_metrics: bool = False
     heartbeats_period_secs: float = HEARTBEAT_SECONDS
+    interruption_strategies: List[BaseInterruptionStrategy] = Field(default_factory=list)
     observers: List[BaseObserver] = Field(default_factory=list)
     report_only_initial_ttfb: bool = False
     send_initial_empty_metrics: bool = True
     start_metadata: Dict[str, Any] = Field(default_factory=dict)
-    interruption_strategies: List[BaseInterruptionStrategy] = Field(default_factory=list)
 
 
 class PipelineTaskSource(FrameProcessor):
@@ -125,7 +129,7 @@ class PipelineTaskSink(FrameProcessor):
         await self._down_queue.put(frame)
 
 
-class PipelineTask(BaseTask):
+class PipelineTask(BasePipelineTask):
     """Manages the execution of a pipeline, handling frame processing and task lifecycle.
 
     It has a couple of event handlers `on_frame_reached_upstream` and
@@ -172,21 +176,24 @@ class PipelineTask(BaseTask):
     Args:
         pipeline: The pipeline to execute.
         params: Configuration parameters for the pipeline.
-        observers: List of observers for monitoring pipeline execution.
-        clock: Clock implementation for timing operations.
+        additional_span_attributes: Optional dictionary of attributes to propagate as
+            OpenTelemetry conversation span attributes.
+        cancel_on_idle_timeout: Whether the pipeline task should be cancelled if
+            the idle timeout is reached.
         check_dangling_tasks: Whether to check for processors' tasks finishing properly.
+        clock: Clock implementation for timing operations.
+        conversation_id: Optional custom ID for the conversation.
+        enable_tracing: Whether to enable tracing.
+        enable_turn_tracking: Whether to enable turn tracking.
+        enable_watchdog_logging: Whether to print task processing times.
+        idle_timeout_frames: A tuple with the frames that should trigger an idle
+            timeout if not received withing `idle_timeout_seconds`.
         idle_timeout_secs: Timeout (in seconds) to consider pipeline idle or
             None. If a pipeline is idle the pipeline task will be cancelled
             automatically.
-        idle_timeout_frames: A tuple with the frames that should trigger an idle
-            timeout if not received withing `idle_timeout_seconds`.
-        cancel_on_idle_timeout: Whether the pipeline task should be cancelled if
-            the idle timeout is reached.
-        enable_turn_tracking: Whether to enable turn tracking.
-        enable_turn_tracing: Whether to enable turn tracing.
-        conversation_id: Optional custom ID for the conversation.
-        additional_span_attributes: Optional dictionary of attributes to propagate as
-            OpenTelemetry conversation span attributes.
+        observers: List of observers for monitoring pipeline execution.
+        watchdog_timeout_secs: Watchdog timer timeout (in seconds). A warning
+            will be logged if the watchdog timer is not reset before this timeout.
     """
 
     def __init__(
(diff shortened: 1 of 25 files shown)


src/pipecat/pipeline/task.py
```diff
@@ -172,55 +176,59 @@ async def on_pipeline_cancelled(task, frame: CancelFrame):
     Args:
```
filipi87, line 176: I guess we are missing the docstring for `watchdog_timeout`.

filipi87: And for `enable_watchdog_logging`

src/pipecat/processors/frame_processor.py
```diff
@@ -171,14 +179,30 @@ async def stop_all_metrics(self):
...
-        return self._task_manager.create_task(coroutine, name)
+        return self._task_manager.create_task(
+            coroutine,
+            name,
+            enable_watchdog_logging=(
+                enable_watchdog_logging
+                if enable_watchdog_logging
+                else self._enable_watchdog_logging
+            ),
+            watchdog_timeout=watchdog_timeout if watchdog_timeout else self._watchdog_timeout,
```
filipi87, line 202: nit: could we just simplify to something like this:
```suggestion
            enable_watchdog_logging = enable_watchdog_logging or self._enable_watchdog_logging,
            watchdog_timeout=watchdog_timeout or self._watchdog_timeout,
```

src/pipecat/processors/frame_processor.py
```diff
@@ -190,6 +214,16 @@ async def wait_for_task(self, task: asyncio.Task, timeout: Optional[float] = Non
             raise Exception(f"{self} TaskManager is still not initialized.")
         await self._task_manager.wait_for_task(task, timeout)
 
+    def start_watchdog(self):
+        if not self._task_manager:
+            raise Exception(f"{self} TaskManager is still not initialized.")
+        self._task_manager.start_watchdog(asyncio.current_task())
```
filipi87: Since this validation is already handled inside `get_task_manager()`, could we simplify those by just calling:
```suggestion
        self.get_task_manager().start_watchdog(asyncio.current_task())
```

filipi87: The same for the other methods.

src/pipecat/processors/frame_processor.py
```diff
@@ -171,14 +179,30 @@ async def stop_all_metrics(self):
...
+            name,
+            enable_watchdog_logging=(
+                enable_watchdog_logging
+                if enable_watchdog_logging
+                else self._enable_watchdog_logging
+            ),
+            watchdog_timeout=watchdog_timeout if watchdog_timeout else self._watchdog_timeout,
+        )
 
     async def cancel_task(self, task: asyncio.Task, timeout: Optional[float] = None):
```
filipi87, line 205: Would it be possible to add a watchdog here, when we are cancelling the task ?

filipi87:
Hi, this is great. 🙂

If I understood it correctly, in most cases we’re basically always starting the watchdog after `get()` and resetting it on `task_done()`. 

I’m wondering if we couldn't to extend `asyncio.Queue` and create a `WatchdogQueue`, where this behavior is handled automatically whenever `get()` or `task_done()` is called.

Something like this:

```
class WatchdogQueue(asyncio.Queue):
    def __init__(self, *args, **kwargs):
        super().__init__(*args, **kwargs)

    async def get(self):
        item = await super().get()
        task = asyncio.current_task()
        if task:
            # TODO: add our watchdog logic to start
        return item

    def task_done(self):
        task = asyncio.current_task()
        if task:
            # TODO: add our watchdog logic to reset
        super().task_done()
```

So, basically this would happen transparent for anyone using it.

What do you think @aconchillo ?

src/pipecat/services/assemblyai/stt.py
```diff
@@ -190,13 +190,17 @@ async def _receive_task_handler(self):
             while self._connected:
                 try:
                     message = await self._websocket.recv()
+                    self.start_watchdog()
                     data = json.loads(message)
                     await self._handle_message(data)
+                    self.reset_watchdog()
```
filipi87: I guess since we also have it inside the `finally`, we can remove this one.

src/pipecat/services/aws/stt.py
```diff
@@ -334,6 +337,8 @@ async def _receive_loop(self):
                 else:
                     logger.debug(f"{self} Other message type received: {headers}")
                     logger.debug(f"{self} Payload: {payload}")
+
+                self.reset_watchdog()
```
filipi87: I guess since we also have it inside the finally, we can remove this one.

src/pipecat/services/aws_nova_sonic/aws.py
```diff
@@ -726,11 +728,13 @@ async def _receive_task_handler(self):
                         elif "completionEnd" in event_json:
                             # Handle the LLM completion ending
                             await self._handle_completion_end_event(event_json)
-
+                self.reset_watchdog()
```
filipi87: Same here. I guess we can remove this one.

src/pipecat/utils/asyncio.py
```diff
@@ -181,15 +264,72 @@ async def cancel_task(self, task: asyncio.Task, timeout: Optional[float] = None)
 
     def current_tasks(self) -> Sequence[asyncio.Task]:
         """Returns the list of currently created/registered tasks."""
-        return list(self._tasks.values())
+        return [data.task for data in self._tasks.values()]
 
-    def _add_task(self, task: asyncio.Task):
+    def start_watchdog(self, task: asyncio.Task):
+        """Starts the given task watchdog timer. If not reset, a warning will be
+        logged indicating the task is stalling.
```
filipi87: We are just showing a warning in case the task doesn't exist. Should we add some validation, or just update this comment ?

filipi87: Because I guess it could make sense to check if we have already started the watchdog and print a warning in this case.

src/pipecat/utils/asyncio.py
```diff
@@ -181,15 +264,72 @@ async def cancel_task(self, task: asyncio.Task, timeout: Optional[float] = None)
...
+
+        """
         name = task.get_name()
-        self._tasks[name] = task
+        if name in self._tasks:
+            self._tasks[name].watchdog_start.set()
```
filipi87: Should we also clear the `watchdog_timer` in this case ?

filipi87:
@aconchillo, it looks like if I just interrupt the bot a couple of times, actually more like 5 or 6 times, I already start to see these warnings:
```
2025-06-24 11:06:12.406 | WARNING  | pipecat.utils.asyncio:wait_for_reset:324 - OpenAILLMService#1::__input_frame_task_handler: task is taking too long 5.0 second(s) (forgot to reset watchdog?)
18
2025-06-24 11:06:17.411 | WARNING  | pipecat.utils.asyncio:wait_for_reset:324 - OpenAILLMService#1::__input_frame_task_handler: task is taking too long 5.0 second(s) (forgot to reset watchdog?)
2025-06-24 11:06:18.906 | DEBUG    | pipecat.transports.base_output:_bot_stopped_speaking:410 - Bot stopped speaking
18
2025-06-24 11:06:22.413 | WARNING  | pipecat.utils.asyncio:wait_for_reset:324 - OpenAILLMService#1::__input_frame_task_handler: task is taking too long 5.0 second(s) (forgot to reset watchdog?)
```

src/pipecat/processors/frame_processor.py
aconchillo: Unfortunately no, because here we are checking for None value first.

src/pipecat/processors/frame_processor.py
aconchillo: Good idea! Done!

src/pipecat/processors/frame_processor.py
aconchillo: I added it to the more internal `cancel_task`.

src/pipecat/pipeline/task.py
aconchillo: Done!

src/pipecat/services/assemblyai/stt.py
aconchillo: Yep!

src/pipecat/services/aws/stt.py
aconchillo: Yes!

src/pipecat/services/aws_nova_sonic/aws.py
aconchillo: Yes :-D

src/pipecat/utils/asyncio.py
aconchillo: Doesn't hurt. Done!

src/pipecat/utils/asyncio.py
aconchillo: Good idea. Done!

aconchillo:
> Hi, this is great. 🙂
> 
> If I understood it correctly, in most cases we’re basically always starting the watchdog after `get()` and resetting it on `task_done()`.
> 
> I’m wondering if we couldn't to extend `asyncio.Queue` and create a `WatchdogQueue`, where this behavior is handled automatically whenever `get()` or `task_done()` is called.
> 
> Something like this:
> 
> ```
> class WatchdogQueue(asyncio.Queue):
>     def __init__(self, *args, **kwargs):
>         super().__init__(*args, **kwargs)
> 
>     async def get(self):
>         item = await super().get()
>         task = asyncio.current_task()
>         if task:
>             # TODO: add our watchdog logic to start
>         return item
> 
>     def task_done(self):
>         task = asyncio.current_task()
>         if task:
>             # TODO: add our watchdog logic to reset
>         super().task_done()
> ```
> 
> So, basically this would happen transparent for anyone using it.
> 
> What do you think @aconchillo ?

I thinks it might be good idea  in the future, but I think for now we shouldn't add more types until we understand a bit more if watchdog timers are useful and how they are used. We can always add more stuff in the future.

aconchillo:
> @aconchillo, it looks like if I just interrupt the bot a couple of times, actually more like 5 or 6 times, I already start to see these warnings:
> 
> ```
> 2025-06-24 11:06:12.406 | WARNING  | pipecat.utils.asyncio:wait_for_reset:324 - OpenAILLMService#1::__input_frame_task_handler: task is taking too long 5.0 second(s) (forgot to reset watchdog?)
> 18
> 2025-06-24 11:06:17.411 | WARNING  | pipecat.utils.asyncio:wait_for_reset:324 - OpenAILLMService#1::__input_frame_task_handler: task is taking too long 5.0 second(s) (forgot to reset watchdog?)
> 2025-06-24 11:06:18.906 | DEBUG    | pipecat.transports.base_output:_bot_stopped_speaking:410 - Bot stopped speaking
> 18
> 2025-06-24 11:06:22.413 | WARNING  | pipecat.utils.asyncio:wait_for_reset:324 - OpenAILLMService#1::__input_frame_task_handler: task is taking too long 5.0 second(s) (forgot to reset watchdog?)
> ```

Ooops, yes. Should be fixed now.

src/pipecat/utils/asyncio.py
```diff
@@ -165,6 +248,8 @@ async def cancel_task(self, task: asyncio.Task, timeout: Optional[float] = None)
         name = task.get_name()
         task.cancel()
         try:
+            # Make sure to reset watchdog if a task is cancelled.
+            self.reset_watchdog(task)
```
filipi87, line 252: 🎉

filipi87 (approved):
Looks good, nice improvement. 🚀

markbackman:
@aconchillo can you write up a short guide for the docs on how to use the watchdog timer?

aconchillo:
> @aconchillo can you write up a short guide for the docs on how to use the watchdog timer?

Sure!

PR #2051, opened 2025-06-24
Implement "direct functions", which allow you to bypass specifying a …

kompfner:
…function configuration (as a `FunctionSchema` or in a provider-specific format) and use the Python function directly. Metadata is gathered automatically from the function signature and docstring.


kompfner:
Note: stay tuned for a corresponding docs update PR.

src/pipecat/adapters/schemas/direct_function.py
```diff
@@ -0,0 +1,227 @@
...
+
+    "Direct" functions' metadata is automatically extracted from their function signature and
+    docstrings, allowing them to be used without accompanying function configurations (as
+    `FunctionSchema`s or in provider-specific formats).
+    """
+
+    async def __call__(self, params: "FunctionCallParams", **kwargs: Any) -> None: ...
+
+
+class BaseDirectFunctionWrapper:
```
kompfner, line 34: Why is there a `BaseDirectFunctionWrapper` in addition to a `DirectFunctionWrapper`?

Glad you asked. I'm planning to update the Pipecat Flows direct functions to share the same logic. So stay tuned for a Pipecat Flows PR.

kompfner:
Note: need changelog update here. Will add that.

src/pipecat/adapters/schemas/direct_function.py
```diff
@@ -0,0 +1,228 @@
...
+
+        # Get function description
+        self.description = (docstring.description or "").strip()
+
+        # Get function parameters as JSON schemas, and the list of required parameters
+        self.properties, self.required = self._get_parameters_as_jsonschema(
+            self.function, docstring.params
+        )
+
+    # TODO: maybe to better support things like enums, check if each type is a pydantic type and use its convert-to-jsonschema function
```
filipi87, line 92: Is this something we hope to include in this PR ?

src/pipecat/services/llm_service.py
```diff
@@ -295,6 +329,11 @@ def unregister_function(self, function_name: Optional[str]):
         if self._start_callbacks[function_name]:
             del self._start_callbacks[function_name]
 
+    def unregister_direct_function(self, handler: Any):
+        wrapper = DirectFunctionWrapper(handler)
+        del self._functions[wrapper.name]
+        # Note: no need to remove start callback here, as direct functions don't support start callbacks.
```
filipi87, line 335: 👍

src/pipecat/adapters/schemas/direct_function.py
filipi87: Actually even if we need this let's include in a follow up PR. Or maybe just remove the TODO if not needed.

kompfner: This is a note for a future PR (this TODO was ported from Pipecat Flows, where this logic originally lived)

Issue #2042, opened 2025-06-22
ElevenLabs multi keepalive


ldenoue:
Currently we only pass an empty JSON to keep alive https://github.com/pipecat-ai/pipecat/blob/8c6a441064f7cd57d60c6b9fd8038f2f9ea4ffa5/src/pipecat/services/elevenlabs/tts.py#L425

but ElevenLabs doc says we should also pass a `context_id` and empty `text`?

https://elevenlabs.io/docs/api-reference/text-to-speech/v-1-text-to-speech-voice-id-multi-stream-input#send.keepContextAlive

markbackman:
You're right. The beta docs we received didn't address this point, but in talking with the 11Labs team, this is the recommended pattern. Note: keepalives work without the context_id, but we want to align to the docs, so I've added https://github.com/pipecat-ai/pipecat/pull/2052.

PR #2052, opened 2025-06-24
Send context_id when available in ElevenLabsTTSService keepalive message

markbackman:

In most cases, there will be a context_id available, so we can refresh the connection for the context_id. There's an edge case to handle, too, where an interruption occurs but no transcript is received. In this case, the context_id is None, so we just push a keepalive message without the context.

In talking with a member of the 11Labs team, they'll add a general purpose keepalive which will keep all contexts alive. That will help to simplify this case in the future.

Resolves: #2042


filipi87 (approved):
LGTM!

PR #2053, opened 2025-06-24
Creating an environment variable for sentry dsn.

filipi87:
Creating an environment variable for sentry dsn.


markbackman (approved):
LGTM

PR #2056, opened 2025-06-24
Update google libraries used in google audio-in examples

kwindla:
Looks like we updated the Google genai library we use in the `GoogleLLMService` but not in some of the examples. There are some types/method differences that were breaking the example context management.


markbackman (approved):
Nice catch!

PR #2062, opened 2025-06-24
allow LLM services to manage watchdog timers

aconchillo:

LLMs might take more than 5 seconds (the watchdog timer default) to generate the full response. In Pipecat we start pushing tokens as soon as we receive them but that doesn't mean the LLM is done. With this PR the LLMs have control of the watchdog timers and use them in the token generation instead.


aconchillo:
Closing in favor of https://github.com/pipecat-ai/pipecat/pull/2063

PR #2063, opened 2025-06-25
no need to call start_watchdog() only reset_watchdog()

aconchillo:

This PR makes it easy to work with watchdog timers. I followed @filipi87 suggestion and extended it to async iterators and others. So now, we can just use these watchdog utilities when we read from a websocket, from a queue or a generic async iterator. The watchdog timer will always reset if we are waiting for data.

So, the main watchdog timers purpose is to find blocking code when processing data (not when waiting for data).


src/pipecat/utils/watchdog_event.py
```diff
@@ -0,0 +1,24 @@
...
+        self._reseter = reseter
+        self._timeout = timeout
+
+    async def wait(self):
+        while True:
+            try:
+                await asyncio.wait_for(super().wait(), timeout=self._timeout)
+                self._reseter.reset_watchdog()
+            except asyncio.TimeoutError:
+                self._reseter.reset_watchdog()
```
filipi87, line 44: Don't we need to do anything else in this case ?

src/pipecat/utils/watchdog_priority_queue.py
```diff
@@ -0,0 +1,25 @@
...
+#
+# SPDX-License-Identifier: BSD 2-Clause License
+#
+
+import asyncio
+
+from pipecat.utils.watchdog_reseter import WatchdogReseter
+
+
+class WatchdogPriorityQueue(asyncio.PriorityQueue):
```
filipi87, line 12: I thought we would also need to override the `task_done` method, and reset the watchdog once the task is completed.

src/pipecat/utils/watchdog_queue.py
```diff
@@ -0,0 +1,25 @@
...
+#
+# SPDX-License-Identifier: BSD 2-Clause License
+#
+
+import asyncio
+
+from pipecat.utils.watchdog_reseter import WatchdogReseter
+
+
+class WatchdogQueue(asyncio.Queue):
```
filipi87, line 12: Same here, I thought we would also need to override the `task_done` method, and reset the watchdog once the task is completed.

src/pipecat/utils/watchdog_priority_queue.py
```diff
@@ -0,0 +1,25 @@
...
+    def __init__(self, reseter: WatchdogReseter, maxsize: int = 0, timeout: float = 1.0) -> None:
+        super().__init__(maxsize)
+        self._reseter = reseter
+        self._timeout = timeout
+
+    async def get(self):
+        while True:
+            try:
+                item = await asyncio.wait_for(super().get(), timeout=self._timeout)
+                self._reseter.reset_watchdog()
```
filipi87, line 47: It makes sense, so we reset the watchdog as soon as we get the element. 

And I thought in most cases we would reset the watchdog again at the `task_done` method.

src/pipecat/utils/watchdog_priority_queue.py
```diff
@@ -0,0 +1,25 @@
...
+                return item
+            except asyncio.TimeoutError:
+                self._reseter.reset_watchdog()
```
filipi87, line 50: Would it also make sense to create some kind of internal monitor for the queue while waiting for elements?
For example, let's say it hits five timeouts in a row without retrieving any new elements. In that case, we could print a warning.
I think this could be helpful. And each time we retrieve a new element from the queue, we reset this counter.

src/pipecat/utils/wathcdog_async_iterator.py
```diff
@@ -0,0 +1,42 @@
...
+
+        try:
+            item = await asyncio.wait_for(self._iter.__anext__(), timeout=self._timeout)
+            self._processor.reset_watchdog()
+            return item
+        except asyncio.TimeoutError:
+            self._processor.reset_watchdog()
+            return await self.__anext__()
+        except StopAsyncIteration:
+            raise
```
filipi87: Wouldn't be better to use a loop instead of recursion in this case ?

```
while True:
    try:
        item = await asyncio.wait_for(self._iter.__anext__(), timeout=self._timeout)
        self._processor.reset_watchdog()
        return item
    except asyncio.TimeoutError:
        self._processor.reset_watchdog()
        continue
    except StopAsyncIteration:
        raise
```

src/pipecat/pipeline/parallel_pipeline.py
```diff
@@ -202,18 +203,14 @@ async def _parallel_push_frame(self, frame: Frame, direction: FrameDirection):
     async def _process_up_queue(self):
         while True:
             frame = await self._up_queue.get()
-            self.start_watchdog()
             await self._parallel_push_frame(frame, FrameDirection.UPSTREAM)
             self._up_queue.task_done()
```
filipi87, line 210: I added a comment inside the `WatchdogQueue` about the `task_done`, I thought we would also reset the watchdog there.

src/pipecat/pipeline/task.py
```diff
@@ -673,6 +676,8 @@ async def _idle_monitor_handler(self):
                     self._idle_queue.get(), timeout=self._idle_timeout_secs
                 )
 
+                self._task_manager.reset_watchdog(self._idle_monitor_task)
```
filipi87: Considering that we are reseting the watchdog inside the queue, as soon as we get the element, do we also need it here ?

src/pipecat/services/anthropic/llm.py
```diff
@@ -203,7 +204,7 @@ async def _process_context(self, context: OpenAILLMContext):
             json_accumulator = ""
 
             function_calls = []
-            async for event in response:
+            async for event in WatchdogAsyncIterator(response, processor=self):
```
filipi87: I can see why you created the WatchdogAsyncIterator now =)

src/pipecat/services/assemblyai/stt.py
```diff
@@ -189,8 +189,7 @@ async def _receive_task_handler(self):
         try:
             while self._connected:
                 try:
-                    message = await self._websocket.recv()
-                    self.start_watchdog()
+                    message = await asyncio.wait_for(self._websocket.recv(), timeout=1.0)
```
filipi87, line 192: I guess we need to handle the `TimeoutError` in this case now, right ?

src/pipecat/services/aws/stt.py
```diff
@@ -284,9 +284,7 @@ async def _receive_loop(self):
                 break
 
             try:
-                response = await self._ws_client.recv()
-
-                self.start_watchdog()
+                response = await asyncio.wait_for(self._ws_client.recv(), timeout=1.0)
```
filipi87, line 287: Same here, I guess we need to handle the `TimeoutError` in this case now, right ?

filipi87: Also this default timeout which we are using, shouldn't we calculate it based on the `watchdog_timeout` ? Because the user can define the `watchdog_timeout` and in theory this timeout should always be lower than that.

src/pipecat/services/aws_nova_sonic/aws.py
```diff
@@ -697,9 +697,9 @@ async def _receive_task_handler(self):
         try:
             while self._stream and not self._disconnecting:
                 output = await self._stream.await_output()
-                result = await output[1].receive()
+                result = await asyncio.wait_for(output[1].receive(), timeout=1.0)
```
filipi87, line 700: Why do we need the `asyncio.wait_for` in this case ?

filipi87: And if we do need it, I guess we would need to handle the `TimeoutError`.
But it feels like we shouldn't need the `asyncio.wait_for` here.

src/pipecat/services/gladia/stt.py
```diff
@@ -561,15 +560,11 @@ async def _receive_task_handler(self):
                                 translation, "", time_now_iso8601(), translated_language
                             )
                         )
-
-                self.reset_watchdog()
```
filipi87: It does feels we need this one right ? Like, the task is completed, so we should reset the watchdog. Otherwise we are only going to do this internally in the `WatchdogAsyncIterator`, when we receive the next element.

src/pipecat/services/riva/stt.py
```diff
@@ -285,9 +285,7 @@ async def _handle_response(self, response):
     async def _response_task_handler(self):
         while True:
             response = await self._response_queue.get()
-            self.start_watchdog()
             await self._handle_response(response)
-            self.reset_watchdog()
```
filipi87, line 290: Shouldn't I need to invoke `self._response_queue.task_done()` here ?

src/pipecat/services/tavus/video.py
```diff
@@ -217,7 +218,5 @@ async def _handle_audio_frame(self, frame: OutputAudioRawFrame):
     async def _send_task_handler(self):
         while True:
             frame = await self._queue.get()
-            self.start_watchdog()
             if isinstance(frame, OutputAudioRawFrame) and self._client:
                 await self._client.write_audio_frame(frame)
-            self.reset_watchdog()
```
filipi87, line 223: Same here, should I invoke `self._queue.task_done()` ?

src/pipecat/transports/network/small_webrtc.py
```diff
@@ -422,19 +423,18 @@ async def cancel(self, frame: CancelFrame):
...
 
         except Exception as e:
             logger.error(f"{self} exception receiving data: {e.__class__.__name__} ({e})")
 
     async def _receive_video(self):
         try:
-            async for video_frame in self._client.read_video_frame():
-                self.start_watchdog()
+            video_iterator = self._client.read_video_frame()
+            async for video_frame in video_iterator:
```
filipi87: Do I need to wrap into `WatchdogAsyncIterator` ?

src/pipecat/utils/watchdog_event.py
filipi87: I mean some kind of internal monitor. For example, let's say it hits five timeouts in a row, in that case, we could print a warning.

src/pipecat/utils/watchdog_priority_queue.py
aconchillo: Not sure that's a good idea. Because it really depends on what the queue is doing. We don't want to be logging warnings for no reason.

src/pipecat/utils/wathcdog_async_iterator.py
aconchillo: Ah, yes. I forgot Python doesn't have tail recursion!

src/pipecat/pipeline/parallel_pipeline.py
aconchillo: Yes, I think that makes sense.

src/pipecat/pipeline/task.py
aconchillo: Oh, this queue should not be a WatchdogQueue actually (it already has a timeout). That's why there's the reset.

src/pipecat/services/assemblyai/stt.py
aconchillo: Ooops, yes!

src/pipecat/services/aws/stt.py
aconchillo: Yes!

src/pipecat/services/aws_nova_sonic/aws.py
aconchillo: Because `receive()` will wait until receiving the first item. So we need to reset the watchdog while we don't. Fixed.

src/pipecat/services/gladia/stt.py
aconchillo: Ah, yes!

src/pipecat/services/riva/stt.py
aconchillo: Yes!

src/pipecat/services/tavus/video.py
aconchillo: Yes!

src/pipecat/transports/network/small_webrtc.py
aconchillo: Yes!

src/pipecat/utils/wathcdog_async_iterator.py
aconchillo: Fixed!

src/pipecat/utils/watchdog_priority_queue.py
aconchillo: Yes!

src/pipecat/utils/watchdog_queue.py
aconchillo: Yes!

src/pipecat/utils/watchdog_priority_queue.py
aconchillo: Yes!

src/pipecat/utils/watchdog_event.py
aconchillo: Same with the queue. I think it's fine. We don't want to print warnings when we shouldn't.

src/pipecat/utils/watchdog_async_iterator.py
```diff
@@ -0,0 +1,60 @@
...
+
+                # The task has finish, so we will create a new one for th next item.
+                self._current_anext_task = None
+
+                return item
+            except asyncio.TimeoutError:
+                self._reseter.reset_watchdog()
+            except StopAsyncIteration:
+                self._current_anext_task = None
+                raise
```
filipi87, line 68: If we change it like this, we are going to prevent the exception that we are receiving when ending the pipeline:

```suggestion
            except StopAsyncIteration:
                # Await or retrieve the task result to suppress asyncio warning
                if self._current_anext_task and self._current_anext_task.done():
                    try:
                        self._current_anext_task.result()
                    except StopAsyncIteration:
                        pass
                self._current_anext_task = None
                raise
```

src/pipecat/utils/asyncio.py
```diff
@@ -316,30 +276,30 @@ def _remove_task(self, task: asyncio.Task):
...
+            finally:
+                timer.clear()
+
+    def _task_done_handler(self, task: asyncio.Task):
+        name = task.get_name()
+        try:
+            task_data = self._tasks[name]
+            if task_data.watchdog_task:
+                task_data.watchdog_task.cancel()
+                task_data.watchdog_task = None
```
filipi87, line 307: Should we also invoke `_remove_task` ? Probably inside the finally.

src/pipecat/pipeline/parallel_pipeline.py
```diff
@@ -83,8 +84,8 @@ def __init__(self, *args):
 
         self._up_task = None
         self._down_task = None
-        self._up_queue = asyncio.Queue()
-        self._down_queue = asyncio.Queue()
```
filipi87, line 87: If we remove this from here, we will need to also refactor the code below, because we are trying to use both queues here in the `__init__` method.

filipi87: Line 98 and 99

src/pipecat/pipeline/parallel_pipeline.py
```diff
@@ -154,7 +155,8 @@ async def process_frame(self, frame: Frame, direction: FrameDirection):
         elif isinstance(frame, EndFrame):
             await self._stop()
 
-    async def _start(self):
+    async def _start(self, frame: StartFrame):
+        self._watchdog_timers_enabled = frame.watchdog_timers_enabled
```
filipi87: It looks like don't have this field `watchdog_timers_enabled` inside the `StartFrame `

filipi87 (approved):
Looks great. Nice refactor. 🚀

PR #2064, opened 2025-06-25
Made new example of voicemail detection with parallel pipelines instead

DominicStewart:
#### Creates a new example of how to do Voicemail detection with parallel pipelines


vr000m:
So this is not strictly a three way pipeline, mainly two. The first one produces the tts if the gate is not switched to the other parallel pipeline.

aconchillo:
Closing. We now have a `VoicemailDetector` processor https://github.com/pipecat-ai/pipecat/blob/main/src/pipecat/extensions/voicemail/voicemail_detector.py

PR #2067, opened 2025-06-25
Update base service class docstrings for better docs auto-generation

markbackman:

I'm working on cleaning up the docs. I'm planning to rely more on the reference API docs to avoid the docs getting out of sync. This is the first of a few PRs that will focus on docs changes. I'll be careful to segment by commit. Any code changes will be in separate commits.

In this PR, there is only one code change to llm_service. Otherwise, all changes are docstrings.


src/pipecat/services/llm_service.py
```diff
@@ -355,7 +355,7 @@ async def run_function_calls(self, function_calls: Sequence[FunctionCallFromLLM]
             else:
                 await self._sequential_runner_queue.put(runner_item)
 
-    async def call_start_function(self, context: OpenAILLMContext, function_name: str):
+    async def _call_start_function(self, context: OpenAILLMContext, function_name: str):
```
markbackman, line 356: FYI: this appears to be used only in the LLMService class, so making it private.

filipi87 (approved):
I have skimmed through the docs, and they look good. This will definitely make things easier to maintain. 🎉

filipi87 (approved):

PR #2072, opened 2025-06-26
utils(asyncio): simplify watchdog helpers

aconchillo:

We simplify watchdog utilities. Since we know the current task we are currently running we can check if watchdog timer is enabled for that task. So, no need to pass if watchdog timers are enabled or not all around.


filipi87 (approved):
Makes sense. Nice cleanup.

PR #2073, opened 2025-06-26
Start HeartBeat when all processors have processed StartFrame

a6kme:
Some of the processors like `STTService` and `TTSService` don't push `StartFrame` ahead in the pipeline, unless they have connected with their service providers. This delays `StartFrame` in downstream processors. 

If we receive `HeartBeatFrame` frame before `StartFrame`, we will get `AttributeError`: `'Processor' object has no attribute '_FrameProcessor__input_queue'`. 

Idea is to start HeartBeats after StartFrame has been processed by all the Processors in the pipeline.


aconchillo (approved):

aconchillo:
Good catch! LGTM! Thank you!

Issue #2076, opened 2025-06-26
streamable http MCP client service doesn't play nice with google LLMs


vipyne:
### pipecat version

0.0.72

### Python version

3.13

### Operating System

macos 15.5

### Issue description

functions fail to register when using streamable http MCP client and google LLM.

### Reproduction steps

1. use GoogleLLMService as LLM in `39c-mcp-run-http.py` example.
2. run bot; bot fails to register functions

### Expected behavior

should be no difference than any other LLM.  functions should register

### Actual behavior

functions fail to register

### Logs

```shell
2025-06-26 15:03:16.863 | DEBUG    | pipecat.adapters.base_llm_adapter:from_standard_tools:17 - Retrieving the tools using the adapter: <class 'pipecat.adapters.services.gemini_adapter.GeminiLLMAdapter'>
2025-06-26 15:03:16.863 | DEBUG    | pipecat.adapters.base_llm_adapter:from_standard_tools:17 - Retrieving the tools using the adapter: <class 'pipecat.adapters.services.gemini_adapter.GeminiLLMAdapter'>
2025-06-26 15:03:16.865 | ERROR    | pipecat.services.google.llm:_process_context:768 - GoogleLLMService#0 exception: 2 validation errors for GenerateContentConfig
tools.0.Tool.function_declarations.41.parameters.properties.files.items.additionalProperties
  Extra inputs are not permitted [type=extra_forbidden, input_value=False, input_type=bool]
    For further information visit https://errors.pydantic.dev/2.10/v/extra_forbidden
tools.0.callable
  Input should be callable [type=callable_type, input_value={'function_declarations':...repo', 'pullNumber']}}]}, input_type=dict]
    For further information visit https://errors.pydantic.dev/2.10/v/callable_type
```

vipyne:
cc @yousifa

yousifa:
addressed in: https://github.com/pipecat-ai/pipecat/pull/2077

PR #2077, opened 2025-06-26
Mcp http gemini support

yousifa:
Google Gemini does not support schemas where `additionalProperties` is set to `false`. This PR cleans up schemas where that property is `false` for Google Gemini models only and updates the streamable http mcp example to gemini.

Resolves: https://github.com/pipecat-ai/pipecat/issues/2076


yousifa:
@vipyne tagging you for review

src/pipecat/services/mcp_service.py
```diff
@@ -81,14 +81,69 @@ async def register_tools(self, llm) -> ToolsSchema:
...
+        """
+        # Check the class name or module to determine if it's a Google LLM service
+        class_name = llm.__class__.__name__
+        module_name = llm.__class__.__module__
+
+        return (
+            "google" in class_name.lower()
+            or "gemini" in class_name.lower()
+            or "google" in module_name.lower()
+        )
```
aconchillo: You could maybe do:

```
def _is_google_llm(self, llm: LLMService) -> bool:
    return isinstance(llm, (GoogleLLMService, GeminiMultimodalLiveLLMService))
```

aconchillo: And probably better. Add a method to LLMService:

```
def needs_mcp_clean_schema(self):
    return False
```

And override them in the LLM service that needs them. Then, in this code just use:

```
        if llm and llm.needs_mcp_clean_schema():
```

vipyne: and maybe we only have to check once before `register_tools` and set flag ie:
```
...
self._is_google_llm = self.needs_mcp_clean_schema()
tools_schema = await self._register_tools(llm)
...
```

then maybe
```
        # Only clean properties for Google/Gemini LLM services
        properties = (
            self._clean_schema_for_google(properties) if self._is_google_llm else properties
        )
```

vipyne: exact naming can be changed, but I like the not calling out google by name - other llms could have similar validation issues/needs in the future.

yousifa: I like these suggestions. Implemented and waiting for review

vipyne (commented):
I'm a little torn on the word "clean"; like maybe it could be: `use_alternate_function_schema_for_mcp`
which is a little wordy-er but is more flexible incase things change in the future and we need to do things like add keys or something.
wdyt @aconchillo ?

yousifa:
@vipyne @aconchillo any strong opinions here? Happy to make the change, lmk

yousifa:
@vipyne I renamed the all instances of 'clean' to 'get_alternate' or similar. I think that makes sense. I'm running into a bunch of linter errors now though, not sure if something was changed on the strictness since my last commit.

vipyne:
@yousifa just one more lint!

yousifa:
Finally ready to merge!

vipyne (approved):

PR #2085, opened 2025-06-27
Fixing pipeline freeze when using Python 3.10

filipi87:
Fixing pipeline freeze when using Python 3.10


src/pipecat/utils/asyncio/task_manager.py
```diff
@@ -302,6 +302,10 @@ async def _watchdog_task_handler(self, task_data: TaskData):
 
         while True:
             try:
+                if task_data.task.done():
+                    logger.warning(f"{name}: task is already done, cancelling watchdog task.")
```
filipi87: I guess we can change it to debug ?

src/pipecat/utils/asyncio/watchdog_queue.py
```diff
@@ -26,23 +26,26 @@ def __init__(
         super().__init__(maxsize)
         self._manager = manager
         self._timeout = timeout
+        self._cancel = False
 
     async def get(self):
-        if self._manager.task_watchdog_enabled:
-            return await self._watchdog_get()
-        else:
-            return await super().get()
```
aconchillo, line 54: I'm not sure this is a good idea, since `wait_for` creates a task for each `get()` coroutine.

filipi87: I see, I thought this was safe enough to use. But I guess if we can prevent from using it, better.

I have refactored to use a cancel sentinel mechanism instead of `asyncio.wait_for`. What do you think ?

src/pipecat/processors/user_idle_processor.py
```diff
@@ -180,6 +180,7 @@ async def _idle_task_handler(self) -> None:
         while True:
             try:
                 await asyncio.wait_for(self._idle_event.wait(), timeout=self._timeout)
+                self.reset_watchdog()
```
aconchillo: Should we add this to `finally`?

aconchillo: I think this should be a `WatchdogEvent`. I believe it will work as expected.

filipi87: True, this should be a `WatchdogEvent` 👍

src/pipecat/utils/asyncio/watchdog_queue.py
aconchillo: I think this is better. We won't have the overhead of creating a task for each frame! Maybe we can add a comment that this is necessary because of Python 3.10.

aconchillo:
Do we need to cancel all the Watchdog queues and priority queues we have?

src/pipecat/utils/asyncio/watchdog_queue.py
filipi87: Done.

src/pipecat/processors/user_idle_processor.py
filipi87: Done.

filipi87:
> Do we need to cancel all the Watchdog queues and priority queues we have?

I believe we should cancel all the queues that are used inside tasks we can create and destroy multiple times. That’s what I’ve tried to do in this PR.
Do you know of any queues that might still be missing?

aconchillo:
> > Do we need to cancel all the Watchdog queues and priority queues we have?
> 
> I believe we should cancel all the queues that are used inside tasks we can create and destroy multiple times. That’s what I’ve tried to do in this PR. Do you know of any queues that might still be missing?

I think `ParallelPipeline`, `ConsumerProcessor` and `RTVIProcessor`.

aconchillo:
LGTM! :clap:

PR #2088, opened 2025-06-28
Turn off thinking for Gemini models by default

kwindla:
The gemini-2.5-flash model has thinking mode enabled by default. This adds significant latency. I think we should turn thinking mode off, unless a thinking config has been passed into `GoogleLLMService`.

Two other changes in this PR
1. I changed the default model to gemini-2.5-flash in the two Google 07 foundational examples.
2. I added a new `reasoning_tokens` field to the LLM usage metrics frame, and wired this up for the `GoogleLLMService`.
3. I also wired up cache tokens for the `GoogleLLMService` but haven't been able to get any cached token counts from the API. Possibly implicit caching doesn't actually tell you when it's happening. We can ask the AI Studio team.

### DIFF
--- a/examples/foundational/07n-interruptible-google.py
+++ b/examples/foundational/07n-interruptible-google.py
@@ -61,7 +61,12 @@ async def run_example(transport: BaseTransport, _: argparse.Namespace, handle_si
         credentials=os.getenv("GOOGLE_TEST_CREDENTIALS"),
     )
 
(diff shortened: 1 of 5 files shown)


filipi87 (approved):
LGTM!

venkatesan-boopalan-zoomrx:
@kwindla Thanks a lot.

PR #2089, opened 2025-06-28
The rest of the docstrings!

markbackman:

Adds/updates docstrings for the rest of src/pipecat.

I've opted to exclude magic methods (e.g. __str__) from checks, as they're usually self-evident.

I've also updated the pre-commit.sh and workflow to run the checks. Everything is green ✅.

With this, we're ready to enforce docstrings. This will help to keep our reference docs up to date. With this, I'll link out to docs.pipecat.ai to rely more on the reference docs since they're complete and as accurate as they can be.


aconchillo (approved):

filipi87 (approved):
LGTM! 🚀

Issue #1653, opened 2025-04-24
User audio contains ticking noise when recording with AudioBufferProcessor


golbin:
### pipecat version

0.0.58~0.0.65

### Python version

3.12

### Operating System

Linux, MacOS

### Issue description

I'm using the `AudioBufferProcessor` to handle audio recording.

While the bot voice and background audio sound fine, there's a consistent ticking noise whenever the user's voice is captured.

I've attached a sample recording that highlights this issue.

Is this expected behavior? If not, could this be a configuration issue? I'd appreciate any guidance on how to resolve it.

[recording-noise.m4a.zip](https://github.com/user-attachments/files/19888884/recording-noise.m4a.zip)

Thanks!

### Reproduction steps

I'm using DailyTransport and AudioBufferProcessor pretty much the same way as in the samples. For the client, I'm using Daily React Native.

```
        DailyParams(
            audio_in_enabled=True,
            audio_out_enabled=True,
            camera_out_enabled=False,
            vad_enabled=True,
            vad_analyzer=SileroVADAnalyzer(),
            vad_audio_passthrough=True,
        ),
```

### Expected behavior

No ticking noise occurs.

### Actual behavior

While the bot voice and background audio sound fine, there's a consistent ticking noise whenever the user's voice is captured.

### Logs

```shell

```

aconchillo:
The clicking issue is because of upsampling. The `AudioBufferProcessor` is probably set to a higher sample rate than the user's input sample rate. Getting to a higher sample rate from a lower sample rate is tricky and not all libraries do it well. It might be we are using the libraries in a wrong way. I'll take a look.

aconchillo:
> The clicking issue is because of upsampling. The `AudioBufferProcessor` is probably set to a higher sample rate than the user's input sample rate. Getting to a higher sample rate from a lower sample rate is tricky and not all libraries do it well. It might be we are using the libraries in a wrong way. I'll take a look.

To temporarily fix this issue try to match the sample rate so there's now upscaling needed.

golbin:
This is causing the problem, right?

```
audio_in_sample_rate: 16000
audio_out_sample_rate: 24000
```

It's tricky because we don't need a higher audio_in_sample_rate, but having a lower audio_out_sample_rate leads to a poor user experience...

aristid:
[...]

PR #2091, opened 2025-06-30
Creating a new stream resampler which avoids clicks.

filipi87:
Creating a new stream resampler which avoids clicks.

Fixes: #2091, #1653.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- Added a new `SOXRStreamAudioResampler` for processing audio in chunks or streams.
+
 - Added new `DailyParams.audio_in_user_tracks` to allow receiving one track per
   user (default) or a single track from the room (all participants mixed).
 
@@ -55,6 +57,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Added `session_token` parameter to `AWSNovaSonicLLMService`.
 
 ### Changed
+
+- Updated all the services to use the new `SOXRStreamAudioResampler`, ensuring smooth
+  transitions and eliminating clicks.
 
 - Upgraded `daily-python` to 0.19.4.
 
--- a/src/pipecat/audio/utils.py
+++ b/src/pipecat/audio/utils.py
@@ -15,14 +15,40 @@ import audioop
 
 import numpy as np
 import pyloudnorm as pyln
-import soxr
 
 from pipecat.audio.resamplers.base_audio_resampler import BaseAudioResampler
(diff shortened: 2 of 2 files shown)


src/pipecat/audio/resamplers/soxr_stream_resampler.py
```diff
@@ -0,0 +1,32 @@
...
+    Audio resampler implementation using the SoX resampler library.
+    It keeps an internal history which avoids clicks at chunk boundaries
+    """
+
+    def __init__(self, in_rate: float, out_rate: float):
+        self.in_rate = in_rate
+        self.out_rate = out_rate
+        self.soxr_stream = soxr.ResampleStream(
+            in_rate=in_rate, out_rate=out_rate, num_channels=1, quality="VHQ", dtype="int16"
+        )
```
filipi87: Clicks occur when using `soxr.resample` on audio chunks because it resets the internal filter state for each chunk

`soxr.ResampleStream` preserves internal filter state across chunks, ensuring smooth transitions and eliminating clicks.

markbackman:
In testing, the audio works as expected for me!

With two base resampler classes when should each be reused? I see we use the `create_default_resampler` for the serializers, but the `create_stream_resampler` for the AudioBufferProcessor and a few transport scenarios.

Also, is the `ResampyResampler` no longer needed? If so, maybe we deprecate it.

filipi87:
@markbackman, I have refactored, now all the services are using the `create_stream_resampler`.

aconchillo:
> In testing, the audio works as expected for me!
> 
> With two base resampler classes when should each be reused? I see we use the `create_default_resampler` for the serializers, but the `create_stream_resampler` for the AudioBufferProcessor and a few transport scenarios.
> 
> Also, is the `ResampyResampler` no longer needed? If so, maybe we deprecate it.

I suggested to remove the `create_default_resampler` and to use the new resampler everywhere. I think the PR is not updated yet.

filipi87:
> I suggested to remove the create_default_resampler and to use the new resampler everywhere. I think the PR is not updated yet.

Hi @aconchillo, yep, it was already updated. We were already using the new resampler everywhere. The only thing I hadn't done yet was deprecate `create_default_resampler`, which I'm going to do now.

> When would the SOXRAudioResampler be recommended for use compared to SOXRStreamAudioResampler. Maybe we could provide some guidance in the docstrings?

I will do that. 👍

aconchillo (approved):

aconchillo:
LGTM! Yay!

aconchillo:
Oh! We are missing a CHANGELOG entry explaining this change.

filipi87:
> Oh! We are missing a CHANGELOG entry explaining this change.

Yep, I always leave to add it just before merge, to avoid the conflicts. 😅

filipi87:
Thank you @markbackman and @aconchillo 🙌

PR #2095, opened 2025-06-30
examples: add --esp32 for SDP munging if host name specified

aconchillo:

If the user specifies the host name `e.g. --host 192.168.1.20` we perform SDP munging where we:

- Only use SHA-256 fingerprint.
- Remove all ICE candidates except the specified host.

We could add a new argument `--sdp-host`, but for now this might be enough.

### DIFF
--- a/src/pipecat/examples/run.py
+++ b/src/pipecat/examples/run.py
@@ -8,6 +8,7 @@ import argparse
 import asyncio
 import json
 import os
+import re
 import sys
 from contextlib import asynccontextmanager
 from typing import Any, Callable, Dict, Mapping, Optional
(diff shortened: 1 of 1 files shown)


src/pipecat/examples/run.py
```diff
@@ -96,12 +124,6 @@ def run_example_webrtc(
     # Store connections by pc_id
     pcs_map: Dict[str, SmallWebRTCConnection] = {}
 
-    ice_servers = [
-        IceServer(
-            urls="stun:stun.l.google.com:19302",
-        )
-    ]
```
aconchillo, line 103: We don't really need this to run examples.

PR #2099, opened 2025-07-01
Add user_id to gladia stt

DominicStewart:
#### Add user_id to transcription frames coming from stt services


DominicStewart:
Before I work on all the services, is this generally how I should be thinking about doing this?

markbackman:
This came up in another thread, so I just spent time to add it:
https://github.com/pipecat-ai/pipecat/pull/2103

You were definitely on the right track. I used the STTService base class to simplify things.

PR #2103, opened 2025-07-01
Add user_id to transcription frames

markbackman:

Also adds user_id to OTel tracing.


CHANGELOG.md
```diff
@@ -41,6 +41,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   tools = ToolsSchema(standard_tools=[do_something])
   ```
 
+  - `user_id` is now populated in the `TranscriptionFrame` and
+    `InterimTranscriptionFrame` when using a service that provides a `user_id`,
```
```suggestion
    `InterimTranscriptionFrame` when using a transport service that provides a `user_id`,
```

src/pipecat/services/stt_service.py
```diff
@@ -241,10 +254,19 @@ async def process_audio_frame(self, frame: AudioRawFrame, direction: FrameDirect
         Continuously buffers audio, growing the buffer while user is speaking and
         maintaining a small buffer when not speaking to account for VAD delay.
 
+        If the frame has a user_id, it is stored for later use in transcription.
```
kompfner, line 257: This kind of assumes an architecture where each user's stream of input audio frames gets fed into an independent (parallel) STT service, right?

What will we do in the case we discussed in person earlier, of a single stream of input audio frames (representing multiple users speaking) being diarized and converted into transcript frames associated with those multiple users? Trying to envision what the next step of changes would be to this logic if we wanted to support that case.

CHANGELOG.md
markbackman: Good catch 👍

src/pipecat/services/stt_service.py
markbackman: Correct. I had considered three scenarios:

1. A simple 1:1 call
2. A future where there's a wrapper processor that takes multiple STT processors—each STT processor gets a single audio stream
3. A single STT processor is fed multiple audio streams

This will work well for cases 1 and 2 without modification.

Case 3 is tricky—this would mean the input streams are mixed and we're reliant on the STT to diarize. We have no way to ensure which user says what and also no way to ensure user_1 through user_n are diarized consistently. I haven't thought of a way of how to handle this. Maybe using the raw response from the STT provider to update the TranscriptionFrame's user_id with the STT's diarization.

WDYT?

kompfner: > This will work well for cases 1 and 2 without modification.

Agree! I'm assuming we're going to pursue 2 before 3.

> no way to ensure user_1 through user_n are diarized consistently

I'm guessing that for STT services that have persistent connections, as long as the connection remains alive it'll keep calling user_1 user_1 throughout the length of the connection? I honestly don't know enough about this off the top of my head though.

But yeah, in this "multiplexing"/diarizing case, the STT service itself could override the (meaningless) `self._user_id` with the ids the STT call sends back. 👍

markbackman: > it'll keep calling user_1 user_1 throughout the length of the connection

In building Ambient Scribe last year, we found that, at least for Deepgram, the speaker IDs can switch during the conversation. There are a bunch of factors but it's not reliable enough for what we want to do here. I think we'll ultimately need multiple tracks.

For now, I think we should move ahead with this. We can revisit as we look more closely at multi-user inputs.

PR #2105, opened 2025-07-01
Clean up docstrings after DirectFunction merge

markbackman:

### DIFF
--- a/src/pipecat/adapters/schemas/direct_function.py
+++ b/src/pipecat/adapters/schemas/direct_function.py
@@ -1,3 +1,17 @@
+#
+# Copyright (c) 2024–2025, Daily
+#
+# SPDX-License-Identifier: BSD 2-Clause License
+#
+
+"""Direct function wrapper utilities for LLM function calling.
+
+This module provides utilities for wrapping "direct" functions that handle LLM
+function calls. Direct functions have their metadata automatically extracted
+from function signatures and docstrings, allowing them to be used without
+manual configuration.
+"""
+
 import inspect
 import types
 from typing import (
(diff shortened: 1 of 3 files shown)


src/pipecat/adapters/schemas/direct_function.py
```diff
@@ -18,6 +32,7 @@
 import docstring_parser
 
 from pipecat.adapters.schemas.function_schema import FunctionSchema
+from pipecat.services.llm_service import FunctionCallParams
```
markbackman: Added this missing import, too.

kompfner: Omitting this import was intentional, to avoid circular dependencies. Note that we refer to the `"FunctionCallParams"` type (in quotes) throughout this file. If we import this, things break.

markbackman: Good to know! Removing.

src/pipecat/adapters/schemas/direct_function.py
```diff
@@ -1,3 +1,17 @@
...
+# Copyright (c) 2024–2025, Daily
+#
+# SPDX-License-Identifier: BSD 2-Clause License
+#
+
+"""Direct function wrapper utilities for LLM function calling.
+
+This module provides utilities for wrapping "direct" functions that handle LLM
+function calls. Direct functions have their metadata automatically extracted
+from function signatures and docstrings, allowing them to be used without
```
kompfner, line 11: Suggestion: "allowing them to be used without accompanying configurations (as
`FunctionSchema`s or in provider-specific formats)."

This matches the wording elsewhere is and is more grokkable at a glance.

src/pipecat/services/llm_service.py
```diff
@@ -312,6 +303,16 @@ def register_direct_function(
         *,
         cancel_on_interruption: bool = True,
     ):
+        """Register a direct function handler for LLM function calls.
+
+        Direct functions have their metadata automatically extracted from their
+        signature and docstring, eliminating the need for manual configuration.
```
kompfner: Same suggestion as before, providing a bit more specificity on what manual configuration is

PR #2110, opened 2025-07-01
DailyTransport: allow receiving audio in a single track

aconchillo:

This brings back reading a single audio track instead of individual tracks per participant.


filipi87 (approved):
LGTM!

PR #2111, opened 2025-07-02
Fix imports and requirements in `examples/`

akaihola:
# Problems
Some of the import statements in `examples/` are invalid due to refactorings in `pipecat` or its dependencies.

Many of the `requirements.txt` files in `examples/` are missing dependencies which are needed for running the examples. Some examples are missing the `requirements.txt` file altogether.

# Solution
This branch contains the following tooling:

`test_all_examples.py` and `test_imports.py` to
- create a virtualenv with a compatible Python version for each example directory
- install dependencies from `requirements.txt`
- import each Python module in the example using that virtualenv
- report import errors

`extract_imports.py` to
- grep all code and documentation for `import pipecat.*` and `from pipecat.* import`
- extract those statements from the AST of all modules
- output all unique identifiers imported from `pipecat`

Using those tools, several missing requirements and invalid imports were identified and corrected.

# Use of an AI assistant
Creation of the tooling was accelerated with help from the Roo Code assistant. Commits with changes to `pipecat` code and documentation were only done manually, after carefully reviewing suggestions from the assistant.

In `.roo/rules/`, the assistant wrote down rules and guidelines it was asked to follow while using the tooling to suggest corrections.


markbackman:
I'm sorry, but we can't accept this change. Thanks for pointing out that it was AI generated. We're not opposed to AI generated code, but this change modifies core source that would break things. I'm closing it.

I've gone through the change and found a few useful things, which I'm including here: https://github.com/pipecat-ai/pipecat/pull/2115 (plus other unrelated cleanup fixes).

akaihola:
>found a few useful things, which I'm including here: #2115

Thanks @markbackman, that's exactly what I was hoping to achieve with this PR!

I suppose any remaining corrections should be proposed via separate PRs?

PR #2114, opened 2025-07-02
Upgrade google-genai version to 1.24.0

markbackman:

Request from a community member. As we discussed, we need to make the package versions more permissive.


aconchillo:
I'm now adding notes to `Changed` section so we can keep track on what we update.

markbackman:
> I'm now adding notes to `Changed` section so we can keep track on what we update.

I've added a "Changed" changelog entry.

PR #2124, opened 2025-07-02
FrameProcessor: remove unnecessary push task

aconchillo:
When we call `FrameProcessor.push_frame()` we end up calling `FrameProcessor.queue_frame()` on the next or previous processor. This queue already guarantees the order.

That is, there's no need to have a two queues next to each other.


filipi87 (approved):
Yep, it does makes sense. Nice improvement. 

As you mentioned, inside the `queue_frame`, we are already distinguishing between the `SystemFrames` and the other frames to insert into the queue. 

Maybe it would even be worth adding it to the release, if we’re able to do enough tests. Otherwise, I agree, let's merge it only after this release.

PR #2129, opened 2025-07-03
Add persona validation (check that microphone is enabled)

carolin-tavus:

- With the recent update to the transport mechanism of audio to Tavus (custom audio stream instead of app messages) update the personas need to have the microphone enabled in their transport input settings.
- This PR adds a validation for this


filipi87 (approved):
Looks good. Thank you for doing this @carolin-tavus. 🚀

filipi87:
@carolin-tavus, before merging, just run this script inside Pipecat:

`./scripts/fix-ruff.sh`

It will format the code. That’s the only thing. Everything else looks good!

PR #2137, opened 2025-07-03
FishAudioTTSService: arg cleanup, add new InputParam and arg

markbackman:

Fish Audio has changed their naming:
- `model` is now `reference_id`, so we're deprecating `model` and using `reference_id` instead
- Also, we're adding the arg called `model_id` which is the actual model and a new param to `InputParams` called `normalize`.

Replacing #2031 which started this work.


filipi87 (approved):
LGTM!

Issue #2143, opened 2025-07-04
Duplicate code


aristid:
https://github.com/pipecat-ai/pipecat/blob/dbc76389d8a818fda27a45443c7d31c25d9c1a91/src/pipecat/services/gemini_multimodal_live/gemini.py#L572C1-L576C83

These lines are duplicated:
```
# Initialize the File API client
self.file_api = GeminiFileAPI(api_key=api_key, base_url=file_api_base_url)
```

markbackman:
Thanks for the heads up. Resolved in: https://github.com/pipecat-ai/pipecat/pull/2146.

PR #2146, opened 2025-07-04
Removing duplicated code inside Gemini.

filipi87:
Removing duplicated code inside Gemini.


filipi87:
Fixes the issue: https://github.com/pipecat-ai/pipecat/issues/2143

PR #2147, opened 2025-07-04
UserIdleProcessor: Account for function calls in progress

markbackman:

This came up from a developer in Discord. Function call handlers can take a while to return a result, so this handles that case.

Since frames are queued, the sequence should always be:
- user started speaking
- user stopped speaking
- llm inference
- function call in progress
- function call result

Given this, it seems safe to use the same `_interrupted` flag for user speaking and function call tracking.

### DIFF
--- a/src/pipecat/processors/user_idle_processor.py
+++ b/src/pipecat/processors/user_idle_processor.py
@@ -15,6 +15,8 @@ from pipecat.frames.frames import (
     CancelFrame,
     EndFrame,
     Frame,
+    FunctionCallInProgressFrame,
+    FunctionCallResultFrame,
     StartFrame,
     UserStartedSpeakingFrame,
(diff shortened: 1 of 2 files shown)


src/pipecat/processors/user_idle_processor.py
```diff
@@ -168,6 +170,13 @@ async def process_frame(self, frame: Frame, direction: FrameDirection) -> None:
                 self._idle_event.set()
             elif isinstance(frame, BotSpeakingFrame):
                 self._idle_event.set()
+            elif isinstance(frame, FunctionCallInProgressFrame):
+                # Function calls can take longer than the timeout, so we want to prevent idle callbacks
```
filipi87, line 174: True, makes sense.

filipi87 (approved):
Yepp, makes sense. 👍

PR #2148, opened 2025-07-04
Refactoring to use aioboto3 instead of boto3

filipi87:

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -42,7 +42,7 @@ Website = "https://pipecat.ai"
 [project.optional-dependencies]
 anthropic = [ "anthropic~=0.49.0" ]
 assemblyai = [ "websockets~=13.1" ]
-aws = [ "boto3~=1.37.16", "websockets~=13.1" ]
+aws = [ "aioboto3~=15.0.0", "websockets~=13.1" ]
 aws-nova-sonic = [ "aws_sdk_bedrock_runtime~=0.0.2" ]
 azure = [ "azure-cognitiveservices-speech~=1.42.0"]
 cartesia = [ "cartesia~=2.0.3", "websockets~=13.1" ]
--- a/src/pipecat/services/aws/llm.py
+++ b/src/pipecat/services/aws/llm.py
@@ -55,7 +55,7 @@ from pipecat.services.llm_service import LLMService
 from pipecat.utils.tracing.service_decorators import traced_llm
 
 try:
-    import boto3
+    import aioboto3
     import httpx
     from botocore.config import Config
 except ModuleNotFoundError as e:
@@ -749,13 +749,17 @@ class AWSBedrockLLMService(LLMService):
                 read_timeout=300,  # 5 minutes
                 retries={"max_attempts": 3},
             )
-        session = boto3.Session(
-            aws_access_key_id=aws_access_key,
-            aws_secret_access_key=aws_secret_key,
-            aws_session_token=aws_session_token,
-            region_name=aws_region,
-        )
-        self._client = session.client(service_name="bedrock-runtime", config=client_config)
+
+        self._aws_session = aioboto3.Session()
+
+        # Store AWS session parameters for creating client in async context
+        self._aws_params = {
+            "aws_access_key_id": aws_access_key,
+            "aws_secret_access_key": aws_secret_key,
+            "aws_session_token": aws_session_token,
+            "region_name": aws_region,
+            "config": client_config,
+        }
 
         self.set_model_name(model)
         self._settings = {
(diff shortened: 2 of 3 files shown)


src/pipecat/frames/frames.py
```diff
@@ -1177,7 +1177,7 @@ class StopFrame(ControlFrame):
 
 
 @dataclass
-class HeartbeatFrame(ControlFrame):
+class HeartbeatFrame(SystemFrame):
```
filipi87: We need to make `HeartbeatFrame` a `SystemFrame` if we wish to receive it at regular intervals.
I will add more details in a comment in the PR.

filipi87:
Even though fixing the `AWSBedrockLLMService` and  `AWSPollyTTSService` to working async, the `HeartbeatFrame` will take several seconds to be processed. And the reason is:
- As a result of the LLM, several `LLMTextFrame` will be generated and queued at once. Sometimes in my tests, more than 30 at once.
- So, this means if the `HeartbeatFrame` is not a `SystemFrame`, it is only going to be processed once all these frames are processed.
- That means now inside the `AWSPollyTTSService`, or actually any other TTS service, we will create multiple `TTSAudioRawFrame` for each `LLMTextFrame`.
- So, only after processing all `LLMTextFrame` and pushing several `TTSAudioRawFrame` is that we would process the `HeartbeatFrame`.

For example:
```
0, id: 623, name: 'TTSAudioRawFrame#177', metadata: {} at 9.87s
2025-07-04 15:43:13.522 | DEBUG    | pipecat.observers.loggers.debug_log_observer:on_push_frame:217 - AWSPollyTTSService#0 → SmallWebRTCOutputTransport#0: TTSAudioRawFrame sample_rate: 24000, num_channels: 1, num_frames: 12000, id: 1096, name: 'TTSAudioRawFrame#441', metadata: {} at 9.87s
2025-07-04 15:43:13.522 | DEBUG    | pipecat.observers.loggers.debug_log_observer:on_push_frame:217 - AWSPollyTTSService#0 → SmallWebRTCOutputTransport#0: TTSAudioRawFrame sample_rate: 24000, num_channels: 1, num_frames: 12000, id: 1097, name: 'TTSAudioRawFrame#442', metadata: {} at 9.87s
2025-07-04 15:43:13.522 | DEBUG    | pipecat.observers.loggers.debug_log_observer:on_push_frame:217 - AWSPollyTTSService#0 → SmallWebRTCOutputTransport#0: TTSAudioRawFrame sample_rate: 24000, num_channels: 1, num_frames: 12000, id: 1098, name: 'TTSAudioRawFrame#443', metadata: {} at 9.87s
2025-07-04 15:43:13.522 | DEBUG    | pipecat.observers.loggers.debug_log_observer:on_push_frame:217 - AWSPollyTTSService#0 → SmallWebRTCOutputTransport#0: TTSAudioRawFrame sample_rate: 24000, num_channels: 1, num_frames: 10200, id: 1099, name: 'TTSAudioRawFrame#444', metadata: {} at 9.87s
2025-07-04 15:43:13.522 | DEBUG    | pipecat.processors.metrics.frame_processor_metrics:stop_processing_metrics:152 - AWSPollyTTSService#0 processing time: 1.3070080280303955
2025-07-04 15:43:13.523 | DEBUG    | pipecat.observers.loggers.debug_log_observer:on_push_frame:217 - AWSPollyTTSService#0 → SmallWebRTCOutputTransport#0: HeartbeatFrame id: 117, name: 'HeartbeatFrame#1', metadata: {}, timestamp: 1712013209 at 9.87s
2025-07-04 15:43:13.523 | DEBUG    | pipecat.observers.loggers.debug_log_observer:on_push_frame:217 - AWSPollyTTSService#0 → SmallWebRTCOutputTransport#0: HeartbeatFrame id: 182, name: 'HeartbeatFrame#2', metadata: {}, timestamp: 2712686209 at 9.87s
2025-07-04 15:43:13.523 | DEBUG    | pipecat.observers.loggers.debug_log_observer:on_push_frame:217 - AWSPollyTTSService#0 → SmallWebRTCOutputTransport#0: HeartbeatFrame id: 285, name: 'HeartbeatFrame#3', metadata: {}, timestamp: 3712335709 at 9.87s
2025-07-04 15:43:13.523 | DEBUG    | pipecat.observers.loggers.debug_log_observer:on_push_frame:217 - AWSPollyTTSService#0 → SmallWebRTCOutputTransport#0: HeartbeatFrame id: 341, name: 'HeartbeatFrame#4', metadata: {}, timestamp: 4711999959 at 9.87s
2025-07-04 15:43:13.523 | DEBUG    | pipecat.observers.loggers.debug_log_observer:on_push_frame:217 - AWSPollyTTSService#0 → SmallWebRTCOutputTransport#0: HeartbeatFrame id: 523, name: 'HeartbeatFrame#5', metadata: {}, timestamp: 5713038709 at 9.87s
2025-07-04 15:43:13.523 | DEBUG    | pipecat.observers.loggers.debug_log_observer:on_push_frame:217 - AWSPollyTTSService#0 → SmallWebRTCOutputTransport#0: HeartbeatFrame id: 736, name: 'HeartbeatFrame#6', metadata: {}, timestamp: 6713698542 at 9.87s
2025-07-04 15:43:13.523 | DEBUG    | pipecat.observers.loggers.debug_log_observer:on_push_frame:217 - AWSPollyTTSService#0 → SmallWebRTCOutputTransport#0: HeartbeatFrame id: 797, name: 'HeartbeatFrame#7', metadata: {}, timestamp: 7713733375 at 9.87s
2025-07-04 15:43:13.523 | DEBUG    | pipecat.observers.loggers.debug_log_observer:on_push_frame:217 - AWSPollyTTSService#0 → SmallWebRTCOutputTransport#0: HeartbeatFrame id: 1022, name: 'HeartbeatFrame#8', metadata: {}, timestamp: 8713295459 at 9.87s
2025-07-04 15:43:13.523 | DEBUG    | pipecat.observers.loggers.debug_log_observer:on_push_frame:217 - AWSPollyTTSService#0 → SmallWebRTCOutputTransport#0: HeartbeatFrame id: 1083, name: 'HeartbeatFrame#9', metadata: {}, timestamp: 9713129292 at 9.87s
```

And it would be the same from the output transport perspective, in this case, it would process all the `TTSAudioRawFrame` first (since they were received first), and only after processing all of them would it process all the `HeartbeatFrame`, which will basically be processed all at once after the bot stops speaking.

examples/foundational/14r-function-calling-aws.py
```diff
@@ -13,6 +13,10 @@
 from pipecat.adapters.schemas.function_schema import FunctionSchema
 from pipecat.adapters.schemas.tools_schema import ToolsSchema
 from pipecat.audio.vad.silero import SileroVADAnalyzer
+from pipecat.frames.frames import (
```
markbackman: This is helpful for testing, but let's remove all changes from this file before merging.

src/pipecat/frames/frames.py
markbackman: Aleix just made the change to `ControlFrame`. In the 0.0.72 changelog:
```
- `HeartbeatFrame`s are now control frames. This will make it easier to detect
  pipeline freezes. Previously, heartbeat frames were system frames which meant
  they were not get queued with other frames, making it difficult to detect
  pipeline stalls.
```

Rather than flip it back here, we should discuss and figure out what we want this value to be.

markbackman (commented):
LGTM. In testing the services, both work well. This one is almost ready to approve.

@filipi87 I think the remaining to dos on this one are:
- Revert 14r example changes
- Add changelog entry mentioning the aioboto3 change
- Consider delaying the HeartbeatFrame change; we should figure out what we want this to be

src/pipecat/frames/frames.py
markbackman: > As a result of the LLM, several LLMTextFrame will be generated and queued at once. Sometimes in my tests, more than 30 at once.

What you're saying about queueing LLMTextFrames is true though. I see this for other LLM services too. I guess this comes down to the purpose of the HeartbeatFrame. Are we looking for services that block the pipeline temporarily or completely?

filipi87: I guess if we’re looking for complete pipeline freezes, we now have the watchdog in place to handle that.

As for the `HeartbeatFrame`, when I see the name, it suggests that it should be to send at regular intervals. However, if it gets queued, it will almost always be processed after several `LLMTextFrame` and many `TTSAudioRawFrames`. Because the output transport processes all the `TTSAudioRawFrames` sequentially, and we always introduce delays between them to send audio at the correct pace. So the `HeartbeatFrame` will likely be delayed by several seconds most of the time.

Maybe a better approach for the future, once we design the protocol and know exactly which frames the processor will handle and which it won’t, would be to have different queues for each. This way, frames that the processor isn’t going to handle can be passed directly to the next element without delay.

filipi87: But for now, I agree. Let’s keep the `HeartbeatFrame` as a `ControlFrame`, since changing that isn’t the goal of this PR. And we can discuss it again once Aleix is back. 🙂

filipi87:
Thank you for the review, @markbackman. I believe I have addressed all your comments.

PR #2156, opened 2025-07-07
WhisperSTTService: Add additional whisper model variants

shahrukhx01:
Add additional whisper model variants including large turbo v3 and whisper small.


src/pipecat/services/whisper/stt.py
```diff
@@ -51,16 +51,20 @@ class Model(Enum):
...
+        LARGE_V3: Latest best quality multilingual model, slower inference.
+        LARGE_V3_TURBO: Fast multilingual model, slightly lower quality than LARGE_V3.
         DISTIL_LARGE_V2: Fast multilingual distilled model.
         DISTIL_MEDIUM_EN: Fast English-only distilled model.
     """
 
     # Multilingual models
     TINY = "tiny"
     BASE = "base"
+    SMALL = "small"
```
markbackman, line 63: Can you include a docstring for this?

src/pipecat/services/whisper/stt.py
```diff
@@ -51,16 +51,20 @@ class Model(Enum):
         BASE: Basic multilingual model, good speed/quality balance.
         MEDIUM: Medium-sized multilingual model, better quality.
         LARGE: Best quality multilingual model, slower inference.
+        LARGE_V3: Latest best quality multilingual model, slower inference.
```
markbackman: Does this have a corresponding model?

markbackman: Also, please keep the order in the docstring and code matching.

markbackman:
@shahrukhx01 thanks for this! Just a few small things to clean up.

The docs are now fully auto-generated, so just ensure that docstrings are accurate:
https://pipecat-docs.readthedocs.io/en/latest/api/pipecat.services.whisper.stt.html

We'll be updating docs.pipecat.ai gradually to remove reference information and point to the auto-generated docs.

src/pipecat/services/whisper/stt.py
shahrukhx01: added.

src/pipecat/services/whisper/stt.py
shahrukhx01: I mis-read the Large model's value however, forgot to remove the docstring.
Now the docstring and values are in matching order

PR #2159, opened 2025-07-07
TranscriptProcessor: Add tracking for tool call results

shahrukhx01:
Background: I need to track successful track tool calls alongside user and assistant messages. Thereby, adding this small PR to extend the `TranscriptProcessor`

- Added a ToolResultProcessor that captures tool call results in conversation transcripts, this dumps tool call metadata into stringified JSON fields include (function name, call ID, arguments, and results) 
- Updated the tests in `tests/test_transcript_processor.py` incl. `test_transcript_processor_factory`
- Added two new tests `test_tool_call_result`, `test_no_tool_call`
- Also, added the `tool` role in `TranscriptionMessage` dataclass
- Added a new example `examples/foundational/28a-transcription-processor-tool-calls.py`

### DIFF
--- a/examples/foundational/28a-transcription-processor-tool-calls.py
+++ b/examples/foundational/28a-transcription-processor-tool-calls.py
@@ -1,0 +1,234 @@
+#
+# Copyright (c) 2024–2025, Daily
+#
+# SPDX-License-Identifier: BSD 2-Clause License
+#
+
+import argparse
+import os
+from typing import List, Optional
+
+from dotenv import load_dotenv
+from loguru import logger
+
+from pipecat.adapters.schemas.function_schema import FunctionSchema
+from pipecat.adapters.schemas.tools_schema import ToolsSchema
+from pipecat.audio.vad.silero import SileroVADAnalyzer
+from pipecat.frames.frames import TranscriptionMessage, TranscriptionUpdateFrame, TTSSpeakFrame
+from pipecat.pipeline.pipeline import Pipeline
+from pipecat.pipeline.runner import PipelineRunner
+from pipecat.pipeline.task import PipelineParams, PipelineTask
+from pipecat.processors.aggregators.openai_llm_context import OpenAILLMContext
+from pipecat.processors.transcript_processor import TranscriptProcessor
+from pipecat.services.cartesia.tts import CartesiaTTSService
+from pipecat.services.deepgram.stt import DeepgramSTTService
+from pipecat.services.llm_service import FunctionCallParams
+from pipecat.services.openai.llm import OpenAILLMService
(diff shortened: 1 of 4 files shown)


shahrukhx01:
Example `transcripts.txt` file for a sample conversation:
```
[2025-07-07T15:56:12.114+00:00] assistant: Hello! I am your helpful assistant. I can answer questions, recommend places to eat, check the weather, and explain complicated topics in everyday language. Just let me know what you need help with!
[2025-07-07T15:56:23.821+00:00] user: Alright. I need to know the weather today.
[2025-07-07T15:56:25.191+00:00] assistant: Sure thing! Please tell me your city and state so I can get the current weather for you.
[2025-07-07T15:56:30.712+00:00] user: I live in New York.
[2025-07-07T15:56:31.858+00:00] tool: {"function_name": "get_current_weather", "tool_call_id": "call_azwK7eKfKFxsRuJ01eXIcOEY", "arguments": {"location": "New York", "format": "fahrenheit"}, "result": {"conditions": "nice", "temperature": "75"}}
[2025-07-07T15:56:31.986+00:00] assistant: Let me check on that.
[2025-07-07T15:56:34.617+00:00] assistant: The weather in New York today is nice, with a temperature around 75 degrees Fahrenheit.
```

@aconchillo @markbackman Would this feature be a useful addition? 🤔

markbackman:
The TranscriptProcessor was built to reflect the "spoken" words of the conversation, so I'm inclined to reject this change as it adds additional scope beyond the goal of the processor.

Rather than add to the TranscriptProcess, you could just log the entire context, which contains all of the raw information that's send to the LLM. That's already included in the OTel tracing implementation.

shahrukhx01:
> The TranscriptProcessor was built to reflect the "spoken" words of the conversation, so I'm inclined to reject this change as it adds additional scope beyond the goal of the processor.
> 
> Rather than add to the TranscriptProcess, you could just log the entire context, which contains all of the raw information that's send to the LLM. That's already included in the OTel tracing implementation.

Sure! Sounds good.  I will close this PR, in case even having a separate subclass of TranscriptProcess won't also add any value?

markbackman:
> Sure! Sounds good. I will close this PR, in case even having a separate subclass of TranscriptProcess won't also add any value?

Yes, please.

I saw you ask a separate question about how to pass tool results to the front end. In your function call handler, you are responsible for providing the result to the LLM. In that location, you can also push an RTVIServerMessageFrame which you could use to share tool call information with your client application. (The client app would receive an onServerMessage callback or ServerMessage event with the data.)

shahrukhx01:
Thanks @markbackman!

PR #2167, opened 2025-07-08
RivaSTTService: reset the watchdog in an async function

markbackman:

RivaSTTService had an issue in the 0.0.74 release—the `reset_watchdog()` was being called in a sync function where no event loop was present. This was causing an error upon initialization. This change moves `reset_watchdog()` to the `_response_task_handler()` which appears to be a more suitable place.

@filipi87 you worked on the watchdog feature with Aleix, so please let me know if this location makes sense. It resolves the previous error.


src/pipecat/services/riva/stt.py
```diff
@@ -339,6 +338,7 @@ async def _handle_response(self, response):
     async def _response_task_handler(self):
         while True:
             response = await self._response_queue.get()
+            self.reset_watchdog()
```
filipi87: @markbackman, this shouldn't even be needed, since we're using the WatchDogQueue, which already handles it internally:
- https://github.com/pipecat-ai/pipecat/blob/3c6a208101ce6e6bba7d3decd2ec6832bc134f70/src/pipecat/utils/asyncio/watchdog_queue.py#L100

So, you can remove it from here

src/pipecat/services/riva/stt.py
```diff
@@ -279,7 +279,6 @@ def _response_handler(self):
             streaming_config=self._config,
         )
         for response in responses:
-            self.reset_watchdog()
```
filipi87, line 282: Yep, it should not be here. 👍

src/pipecat/services/riva/stt.py
markbackman: Ahh, ok, so just remove the `reset_watchdog()` entirely?

filipi87: Yep.

PR #2171, opened 2025-07-08
Minimal AWS Strands demo

kompfner:
I've also put some time into a demo that takes advantage of Strands' ability to "show its work", leveraging its `stream_async` method to output its reasoning steps into `TTSSpeakFrame`s as they're available.

I ran into a couple of roadblocks, though:
1. I couldn't get the Strands agent not to starve the thread that would process the `TTSSpeakFrame`s, so all the reasoning steps were spoken too late, after the agent was already done with all its work.
2. I found that the reasoning messages were way too unreliable—they were too often padded with a bunch of hallucinatory junk before finally landing on the task at hand. (Surprisingly, the hallucinatory junk doesn't seem to affect the agent's ability to get to the right answer...🤔)

So, for now, just going ahead and adding just the "minimal" demo, where the Strands agent is invoked as a "black box" tool.


kompfner:
Hm, I might want to put this into a folder with its own README rather than having it be a "foundational" example, not least because it requires the developer to `pip install` a separate tool (Strands)...

PR #2173, opened 2025-07-08
fix nextjs webhook example num_endpoints null check

### LINKED ISSUE #2172: next.js webhook reports wrong num_endpoints
### pipecat version

master

### Python version

N/A

### Operating System

WSL

### Issue description

The next.js webhook handler always reports num_endpoints=2 even though call_transfer is not sent to it.  It's undefined, but the code is checking for null.  Hat tip to @vr000m and @markbackman in discord for making it clear that matters 😄 

### Reproduction steps

1. Set up the nextjs webhook and point PCC to it
2. Call the bot

### Expected behavior

- num_endpoints reported should be 1...
- on_dialin_stopped should not get called until the person hangs up

### Actual behavior

- num_endpoints = 2
- on_dialin_stopped gets called within ms or seconds after the call starts

### Logs

```shell

```

captaincaius:
Fixes #2172


markbackman (approved):
LGTM!

Issue #2174, opened 2025-07-09
Updates in parallel_pipeline.py from 0.0.73 to 0.0.74 may cause BUG when user speaks


EtFourier:
### pipecat version

0.0.74

### Python version

3.12.7

### Operating System

Ubuntu 22.04

### Issue description

I switched my project from pipecat 0.0.73 to 0.0.74, and then I found that my ParallelPipeline would throw the BUGs below when the user spoke, and the bot got stuck for a long while for some reason.
I'm pretty sure this is about parallel_pipeline.py, because if I replace v0.0.74 parallel_pipeline.py with the v0.0.73 parallel_pipeline.py, and keep every other script v0.0.74, it works well.

2025-07-08 19:35:10.327 | DEBUG    | pipecat.transports.base_output:_bot_stopped_speaking:580 - Bot stopped speaking
2025-07-08 19:35:17.914 | DEBUG    | pipecat.transports.base_input:_handle_user_interruption:333 - User started speaking
2025-07-08 19:35:17.922 | DEBUG    | pipecat.utils.asyncio.watchdog_queue:get:65 - Received WatchdogQueueCancelFrame, throwing CancelledError to force cancelling
2025-07-08 19:35:17.923 | DEBUG    | pipecat.utils.asyncio.watchdog_queue:get:65 - Received WatchdogQueueCancelFrame, throwing CancelledError to force cancelling
2025-07-08 19:35:19.634 | DEBUG    | pipecat.transports.base_input:_handle_user_interruption:357 - User stopped speaking

### Reproduction steps

Your project has a ParallelPipeline, and you just update pipecat from v0.0.73 to v0.0.74, when the user speaks, it occurs.

### Expected behavior

User can speak to the bot normally.

### Actual behavior

Got DEBUG messages below, and the bot gets stuck:
2025-07-08 19:35:17.922 | DEBUG    | pipecat.utils.asyncio.watchdog_queue:get:65 - Received WatchdogQueueCancelFrame, throwing CancelledError to force cancelling
2025-07-08 19:35:17.923 | DEBUG    | pipecat.utils.asyncio.watchdog_queue:get:65 - Received WatchdogQueueCancelFrame, throwing CancelledError to force cancelling

### Logs

```shell

```

markbackman:
You're right. I can repro this by running the 15 foundational example:
https://github.com/pipecat-ai/pipecat/blob/main/examples/foundational/15-switch-voices.py

We'll look into this.

[...]

PR #2183, opened 2025-07-10
Fixed an issue in `ParallelPipeline` that caused errors when handling interruptions.

filipi87:
Fixed an issue in `ParallelPipeline` that caused errors when handling interruptions.


filipi87:
This PR fixes this issue: https://github.com/pipecat-ai/pipecat/issues/2174

CHANGELOG.md
```diff
@@ -5,6 +5,13 @@ All notable changes to **Pipecat** will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Fixed
+
+- Fixed an issue in `ParallelPipeline` that caused errors when handling
```
markbackman: I see the issue when I'm not interrupting as well. It might be worth reflecting that in the changelog.

markbackman (approved):
LGTM and fixes the issue in my testing.

What was the root cause. By switching to `get_nowait()`, I assume something was blocking?

filipi87:
> What was the root cause. By switching to get_nowait(), I assume something was blocking?

The root cause was that we were using the same queue across multiple tasks, and we were also trying to get elements from the queue outside a task. So, in some cases, when trying to invoke get from the watchdogqueue during draining, it would throw a cancelled error.

By using get_nowait, it doesn't go through that logic and will simply drain the queue, which is what we want in this case.

markbackman:
> > What was the root cause. By switching to get_nowait(), I assume something was blocking?
> 
> The root cause was that we were using the same queue across multiple tasks, and we were also trying to get elements from the queue outside a task. So, in some cases, when trying to invoke get from the watchdogqueue during draining, it would throw a cancelled error.
> 
> By using get_nowait, it doesn't go through that logic and will simply drain the queue, which is what we want in this case.

Thank you 🙇

Issue #1054, opened 2025-01-21
User's audio is not received in Pipecat [finalize failed]


Atharv24:
Sometimes, the user keeps speaking, and the bot behaves as if user is on mute. Did some digging around with trace logs, and can see the finalize call fails just after the UserStoppedSpeakingFrame. Therefore, LLM does not receive the Final transcription frame, and no response is generated from the LLM.

![Image](https://github.com/user-attachments/assets/d7f1a1b5-5cb3-4e53-af41-2a7ec9560c01)

Atharv24:
I believe the error arises from here, since the deepgram docs specify that the return value is not guaranteed [https://developers.deepgram.com/docs/finalize]
![Image](https://github.com/user-attachments/assets/66898351-05a0-4fe4-9eb8-09a965053cc1)

markbackman:
Was subsequent audio transcribed by Deepgram after the `finalize failed` message? That is, was only that one audio sample not transcribed and subsequent chunks were transcribed?

Atharv24:
@markbackman The issue once arises, remains for a few seconds (sometimes minutes too). Although it fixes itself after that.

chadbailey59:
From @Atharv24 in the Pipecat Discord:

> However, there were InterimTranscriptionFrames being received before the final flush failed. Maybe we can add some sort of error boundary and stitch the interim frames together even if the final flush fails? 

This is a good idea, but it would be difficult to test, since we can't reliably reproduce a 'finalize failed' from Deepgram. (Maybe we can hack the client to test it?)

pesterhazy:
We also see InterimTranscriptFrames that aren't followed by a TranscriptFrame.

Is this unexpected behavior from Deepgram, or is the issure more related to how Pipecat is integrating with Deepgram?

markbackman:
> We also see InterimTranscriptFrames that aren't followed by a TranscriptFrame.

This is a possible scenario with Deepgram. If a final transcript wasn't provided, my understanding is that the confidence was low enough that a final couldn't be generated.

abhijitpal1247:
I also experienced a similar error, and it remained persistent until we restarted the pipecat process.
@markbackman , can we create a new deepgram connection when we encounter such an error?

agilebean:
I have the same question as @pesterhazy:
Can the pipecat framework 'repair' this issue on-the-fly?

I found it consistently shows this error, transcribing the HumanMessage only after the second half of the sentence:
[...]

PR #2184, opened 2025-07-10
Fixing an issue where Pipecat was not receiving the user's audio

filipi87:
Unlike `create_file_resampler`, which always returns audio data, `create_stream_resampler` may sometimes return empty data.

This happens because `create_stream_resampler`, which uses `SOXRStreamAudioResampler`, maintains internal state between calls to preserve audio quality at chunk boundaries. It buffers audio internally to handle resampling more accurately, which can result in either empty output or a larger than expected output size due to this buffering behavior.

To address this, we implemented three fixes:
- Avoid creating an InputAudioRawFrame when there are no audio bytes.
- Skip serializing a JSON message when no audio is present.
- Fixed the VAD analyzer to process the full audio buffer as long as it contains more than the minimum required bytes per iteration, instead of only analyzing the first chunk.

Fixes: #1054

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- Fixed an issue where using audio input with a sample rate requiring resampling
+  could result in empty audio being passed to STT services, causing errors.
+
+- Fixed the VAD analyzer to process the full audio buffer as long as it contains
+  more than the minimum required bytes per iteration, instead of only analyzing
+  the first chunk.
+
 - Fixed an issue in ParallelPipeline that caused errors when attempting to drain
   the queues.
 


examples/twilio-chatbot/client/typescript/src/app.ts
filipi87, line 1: This just fixes the test app following recent changes introduced by the new Pipecat Client SDK.

src/pipecat/serializers/twilio.py
```diff
@@ -81,7 +86,7 @@ def __init__(
         self._twilio_sample_rate = self._params.twilio_sample_rate
         self._sample_rate = 0  # Pipeline input rate
 
-        self._input_resampler = create_stream_resampler()
+        self._input_resampler = create_file_resampler()
```
filipi87: It doesn’t seem like there’s an issue inside the resampler itself. Instead, the output it produces appears to be causing some issues with Deepgram.

For example, if we replace the `DeepgramSTTService` with the `OpenAISTTService`, everything works as expected even using the `create_stream_resampler`.

For now, this is an workaround to avoid the issue.

filipi87: Ok, I see what the issue is now. Unlike create_file_resampler, which always returns audio data, `create_stream_resampler` sometimes returns empty. In such cases, we were still creating an `InputAudioRawFrame`, but the audio bytes were empty, which caused side effects in the STT services.

markbackman: 💡 That's surprising! Good find.

markbackman:
This looks good and works well in my testing. Can you also update telnyx, plivo, and exotel with a similar change? I tested telnyx and it suffers the same issue, so I would imagine the others will have the same problem.

Are there any other areas you can think of that would be impacted by this?

src/pipecat/services/stt_service.py
```diff
@@ -152,6 +152,13 @@ async def process_audio_frame(self, frame: AudioRawFrame, direction: FrameDirect
         else:
             self._user_id = ""
 
+        if not frame.audio:
+            # Ignoring in case we don't have audio to transcribe.
+            logger.warning(
+                f"Empty audio frame received for STT service: {self.name} {frame.num_frames}"
+            )
+            return
```
filipi87, line 160: This is a safeguard: if we still create empty audio somewhere, we will show a warning and ignore it.

src/pipecat/audio/vad/vad_analyzer.py
```diff
@@ -183,36 +183,37 @@ def analyze_audio(self, buffer) -> VADState:
...
-            match self._vad_state:
-                case VADState.STARTING:
-                    self._vad_state = VADState.QUIET
-                    self._vad_starting_count = 0
-                case VADState.SPEAKING:
-                    self._vad_state = VADState.STOPPING
-                    self._vad_stopping_count = 1
-                case VADState.STOPPING:
-                    self._vad_stopping_count += 1
+        while len(self._vad_buffer) >= num_required_bytes:
```
filipi87, line 186: @markbackman, this was another needed change due to the new resampler. 

Since we can now receive audio chunks that are bigger, the `analyze_audio` method needs to process all available data. So I had to add this loop here to properly analyze the entire audio buffer in the VAD analyzer.

filipi87:
> Are there any other areas you can think of that would be impacted by this?

I’ve pushed a couple of other changes. I think that should be all for now. If you could give it another round of reviewing, that would be great.

filipi87:
This PR fixes this issue: https://github.com/pipecat-ai/pipecat/issues/1054

src/pipecat/services/stt_service.py
markbackman: Makes sense, especially how some services like Deepgram are sensitive to empty audio as they emit a `finalize failed` error. This will also let us track if there are further issues.

Issue #2160, opened 2025-07-07
Emulated user speaking frames don't respect the VAD stop_secs setting


markbackman:
### pipecat version

0.0.74

### Python version

3.12.8

### Operating System

macOS 15.5

### Issue description

When the VAD doesn't detect speech but a final transcript is returned, Pipecat creates an `EmulateUserStartedSpeakingFrame` and a corresponding `UserStartedSpeakingFrame` along with the stop frame equivalents. The resulting `EmulateUserStoppedSpeakingFrame` and corresponding `UserStoppedSpeakingFrame` that are created don't wait for the stop_secs, which means that it's possible for the LLM to provide a completion _while_ the user is still speaking.

### Reproduction steps

You can artificially create this scenario by setting the VAD params to:
```
vad_analyzer=SileroVADAnalyzer(params=VADParams(confidence=1.0, stop_secs=2.0)),
```

Note: The high confidence nearly ensures that the VAD events won't fire. This would not be the case, but it helps to make the repro case more reliable.

Then, just speak a few sentences, taking a short pause less than the stop_secs time of 2.0 seconds.

### Expected behavior

- Wait until the stop_secs time has elapsed to emit the `EmulateUserStoppedSpeakingFrame` and corresponding `UserStoppedSpeakingFrame`.

### Actual behavior

- The LLM completion happens `aggregation_timeout` seconds after the first transcript is received, where `aggregation_timeout` is 0.5 seconds by default. Since this can happen while the user is speaking, the interruption frame is also ignored, it seems. This results in the bot speaking over the user.

### Logs

```shell

```

PR #2185, opened 2025-07-10
fix: use a different aggregation timeout for emulated user speech

markbackman:

This PR fixes: #2160.

The root cause of this issue is that emulated user speech (e.g. no VAD speech detection but the arrival of a TranscriptionFrame) results in pushing an aggregation after the `aggregation_timeout`. The `aggregation_timeout` is 0.5s and the default VAD `stop_secs` is 0.8s, so this often results in a completion occurring before desired. There are two new cases to handle, both of which occur when there is VAD emulation:

1. No turn analyzer present: In this case, the `_aggregation_task_handler` has a separate timeout for the emulated speech case where the timeout is set according to the `VADParams` `stop_secs`. This ensures a common UX for cases where there's both VAD detected speech and VAD emulation.
2. Turn analyzer present: In this case, we can't simply use the `VADParams` `stop_secs` because the `stop_secs` value is usually set very low (e.g. 0.2sec). Instead, we set a separate `LLMUserAggregatorParams` called `turn_emulated_vad_timeout` that defaults to 0.8sec (matching the `VADParams` `stop_secs` default). This default is likely sufficient, but if not, then it's configurable through the user aggregator. It's not ideal to have another value to set, but I don't see a better solution.

This requires two updates to frames:
1. The `StartFrame` gets a new field called `has_turn_analyzer` which defaults to False and is updated by the `BaseInputTransport` before it's pushed downstream.
2. A new `VADParamsNotificationFrame` is added which contains the `VADParams`. This frame is what informs the `LLMUserContextAggregator` of the `VADParams` `stop_secs`. It's emitted by the `BaseInputTransport` both when handling the `StartFrame` and when a `VADParamsUpdateFrame` is handled. This ensures that the `LLMUserContextAggregator` has the correct values.

Also, there's one miscellaneous fix for `.pre-commit-config.yaml`. The update is so that the pre-commit hook uses the pyproject.toml linting rules.

### DIFF
--- a/src/pipecat/frames/frames.py
+++ b/src/pipecat/frames/frames.py
@@ -616,6 +616,7 @@ class StartFrame(SystemFrame):
         enable_usage_metrics: Whether to enable usage metrics collection.
         interruption_strategies: List of interruption handling strategies.
         report_only_initial_ttfb: Whether to report only initial time-to-first-byte.
+        has_turn_analyzer: Whether a turn analyzer is configured in the pipeline.
     """
 
     audio_in_sample_rate: int = 16000
@@ -625,6 +626,7 @@ class StartFrame(SystemFrame):
     enable_usage_metrics: bool = False
     interruption_strategies: List[BaseInterruptionStrategy] = field(default_factory=list)
     report_only_initial_ttfb: bool = False
+    has_turn_analyzer: bool = False
 
 
 @dataclass
--- a/src/pipecat/frames/frames.py
+++ b/src/pipecat/frames/frames.py
@@ -1145,6 +1145,20 @@ class OutputDTMFUrgentFrame(DTMFFrame, SystemFrame):
     pass
 
 
+@dataclass
+class VADParamsNotificationFrame(SystemFrame):
+    """Frame for notifying processors of VAD parameter changes.
+
+    This frame is pushed downstream when VAD parameters are set or updated,
+    allowing processors to adjust their behavior accordingly.
+
+    Parameters:
+        params: Current VAD parameters.
+    """
+
+    params: VADParams
+
+
 #
 # Control frames
 #
--- a/src/pipecat/frames/frames.py
+++ b/src/pipecat/frames/frames.py
@@ -28,6 +28,7 @@ from typing import (
 )
 
 from pipecat.audio.interruptions.base_interruption_strategy import BaseInterruptionStrategy
+from pipecat.audio.turn.smart_turn.base_smart_turn import SmartTurnParams
 from pipecat.audio.vad.vad_analyzer import VADParams
 from pipecat.metrics.metrics import MetricsData
 from pipecat.transcriptions.language import Language
@@ -616,7 +617,6 @@ class StartFrame(SystemFrame):
         enable_usage_metrics: Whether to enable usage metrics collection.
         interruption_strategies: List of interruption handling strategies.
         report_only_initial_ttfb: Whether to report only initial time-to-first-byte.
-        has_turn_analyzer: Whether a turn analyzer is configured in the pipeline.
     """
 
     audio_in_sample_rate: int = 16000
@@ -626,7 +626,6 @@ class StartFrame(SystemFrame):
     enable_usage_metrics: bool = False
     interruption_strategies: List[BaseInterruptionStrategy] = field(default_factory=list)
     report_only_initial_ttfb: bool = False
-    has_turn_analyzer: bool = False
 
 
 @dataclass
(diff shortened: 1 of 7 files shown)


kompfner:
High-level question, before even looking at the code:

Definitely makes sense that we don't want `aggregation_timeout` firing too early in the VAD-didn't-trigger case, as it risks interrupting the user.

But does exactly matching the VAD's `stop_secs` necessary make sense? Seems like we're measuring two conceptually different time gaps to trigger pushing the user aggregation:
1. In the VAD-triggered case: the amount of time after the last voice-like utterance when the VAD system determines it's safe to say the user is done
2. In the VAD-not-triggered case: the amount of time after we received our last bit of transcription when we determine it's safe to say the user is done

1 has to do with analyzing audio. 2 has to do with receiving blobs of text from a network.

Hmm...but if you've configured your VAD to be longer than the default, the behavior you want is for the system to "wait a bit longer" before deciding the user is done speaking...so we should also do that in scenario 2 above...

OK, I'm on board with ensuring that `aggregation_timeout` is _at least_ as long as `stop_secs`, to honor the developer's desire to "wait a bit longer", if they've configured that.

src/pipecat/processors/aggregators/llm_response.py
```diff
@@ -618,9 +631,35 @@ async def _cancel_aggregation_task(self):
...
+                #   for emulated VAD scenarios where we need to allow users time to
+                #   finish speaking (e.g. 0.8s).
+                #
+                # - WITHOUT turn analyzer: Use VAD's stop_secs directly to maintain
+                #   consistent user experience between real VAD detection and
+                #   emulated VAD scenarios.
+                if not self._emulating_vad:
+                    timeout = self._params.aggregation_timeout
+                elif self._has_turn_analyzer:
+                    timeout = self._params.turn_emulated_vad_timeout
```
filipi87, line 660: When the turn analyzer is enabled, why don't we use the `stop_secs` from `SmartTurnParams` in this case ? 

It seems like this would better match the behavior we're using with VAD. What do you think ?

src/pipecat/frames/frames.py
```diff
@@ -1145,6 +1147,20 @@ class OutputDTMFUrgentFrame(DTMFFrame, SystemFrame):
...
+    """Frame for notifying processors of VAD parameter changes.
+
+    This frame is pushed downstream when VAD parameters are set or updated,
+    allowing processors to adjust their behavior accordingly.
+
+    Parameters:
+        params: Current VAD parameters.
+    """
+
+    params: VADParams
```
filipi87: @markbackman, what if we create something more generic, like, `SpeechControlParamsFrame`.

Something like this:

```
@dataclass
class SpeechControlParamsFrame(SystemFrame):
    """
    Frame for notifying processors of speech control parameter changes.
    This includes parameters for both VAD (Voice Activity Detection) and
    turn-taking analysis. It allows downstream processors to adjust their
    behavior based on updated interaction control settings.

    Parameters:
        vad_params: Current VAD parameters.
        turn_params: Current turn-taking analysis parameters.
    """

    vad_params: Optional[VADParams] = None
    turn_params: Optional[SmartTurnParams] = None
```

This way, you wouldn’t need to create a new field, turn_emulated_vad_timeout, inside the llm_response, and it would respect the stop_secs value from either the VAD or the TURN.

Also, you wouldn't need to add `has_turn_analyzer` in the `StartFrame`. 

What do you think ?

src/pipecat/processors/aggregators/llm_response.py
markbackman: We could do that and I thought about this. The one concern is that `SmartTurnParams` `stop_secs` defaults to 3 seconds, which is a bit long to wait if the VAD didn't detect speech for whatever reason. Decoupling the two makes sense (I think) because:
- `SmartTurnParams` `stop_secs` governs how long to wait when the smart-turn model has found that the response is incomplete. In this case, you want to give the user extra time
- When the VAD doesn't fire, the smart-turn model doesn't provide a result. So, we should treat this as a normal, non-smart-turn enabled input. So, a shorter time would be more appropriate.

src/pipecat/frames/frames.py
markbackman: I like this idea, in general, even if we don't go with your suggestion above. This would make this information available to other processors.

In general, making this type of config info available across processors is valuable because there's no easy way to get access from processor to processor.

src/pipecat/processors/aggregators/llm_response.py
filipi87: Ok, yeah, I think I see what you mean, the behavior is truly slightly different in this case. Thank you for explaining it.

src/pipecat/frames/frames.py
filipi87: Agreed. And maybe this would at least allow us to remove the `has_turn_analyzer` from the `StartFrame` in this case.

markbackman: Ok, I'll make this change. I think it's helpful.

filipi87:
@markbackman, I was testing it here, and it seems to be working nicely. 🎉

filipi87 (approved):
LGTM! Nice work.

PR #2190, opened 2025-07-11
Mrkb/add textbox simple chatbot

mattieruth:
This PR fixes an issue with the server-side handling of `append-to-context` and adds a text entry box to the simple-chatbot example to exemplify this behavior.


markbackman (approved):
LGTM! Can you add a changelog entry?

In the near future, let's update simple chatbot to use the voice-ui-kit.

PR #2201, opened 2025-07-14
Only create the EmulateUserStartedSpeakingFrame if we have received a transcription

filipi87:
- Fixed an issue where, in some edge cases, the `EmulateUserStartedSpeakingFrame` could be created even if we didn't have a transcription.

### DIFF
--- a/src/pipecat/processors/aggregators/llm_response.py
+++ b/src/pipecat/processors/aggregators/llm_response.py
@@ -693,7 +693,11 @@ class LLMUserContextAggregator(LLMContextResponseAggregator):
         # to emulate VAD (i.e. user start/stopped speaking), but we do it only
         # if the bot is not speaking. If the bot is speaking and we really have
         # a short utterance we don't really want to interrupt the bot.
-        if not self._user_speaking and not self._waiting_for_aggregation:
+        if (
+            not self._user_speaking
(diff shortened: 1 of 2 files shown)


src/pipecat/processors/aggregators/llm_response.py
```diff
@@ -693,7 +693,11 @@ async def _maybe_emulate_user_speaking(self):
         # to emulate VAD (i.e. user start/stopped speaking), but we do it only
         # if the bot is not speaking. If the bot is speaking and we really have
         # a short utterance we don't really want to interrupt the bot.
-        if not self._user_speaking and not self._waiting_for_aggregation:
+        if (
+            not self._user_speaking
+            and not self._waiting_for_aggregation
+            and len(self._aggregation) > 0
```
kompfner, line 699: this makes a lot of sense to me!

i'm actually not totally clear why it wasn't misfiring all the time before...? based on the existing logic in this file, i'd expect it to fire repeatedly whenever there was a period of silence, since the timeout in `_aggregation_task_handler` would keep going off...but i'm probably missing something

kompfner: oh wait...the timeout actually doesn't trigger this. my bad 🤦 . i understand now

PR #2204, opened 2025-07-15
mcp_tool_wrapper using FunctionCallParams

yousifa:
This PR resolves a warning to use FunctionCallParams in MCPClient when registering tools:
```
 DeprecationWarning: Function calls with parameters `(function_name, tool_call_id, arguments, llm, context, result_callback)` are deprecated, use a single `FunctionCallParams` parameter instead.
 ```


markbackman:
@yousifa nice improvement! If this is ready to go (which I think it is), I can merge it down.

yousifa:
Should be good to go!

PR #2208, opened 2025-07-16
(work-in-progress....not ready yet) Inworld HTTP TTS Service

padillamt:
(work-in-progress....not ready yet)

### DIFF
--- a/examples/foundational/07aa-interruptible-inworld-http.py
+++ b/examples/foundational/07aa-interruptible-inworld-http.py
@@ -1,0 +1,117 @@
+#
+# Copyright (c) 2024–2025, Daily
+#
+# SPDX-License-Identifier: BSD 2-Clause License
+#
+
+import argparse
+import os
+
+import aiohttp
+from dotenv import load_dotenv
+from loguru import logger
+
+from pipecat.audio.vad.silero import SileroVADAnalyzer
+from pipecat.pipeline.pipeline import Pipeline
+from pipecat.pipeline.runner import PipelineRunner
+from pipecat.pipeline.task import PipelineParams, PipelineTask
+from pipecat.processors.aggregators.openai_llm_context import OpenAILLMContext
+from pipecat.services.deepgram.stt import DeepgramSTTService
+from pipecat.services.openai.llm import OpenAILLMService
+from pipecat.services.inworld.tts import InworldHttpTTSService
+from pipecat.transports.base_transport import BaseTransport, TransportParams
+from pipecat.transports.network.fastapi_websocket import FastAPIWebsocketParams
+from pipecat.transports.services.daily import DailyParams
+
+load_dotenv(override=True)
+
+
+# We store functions so objects (e.g. SileroVADAnalyzer) don't get
+# instantiated. The function will be called when the desired transport gets
+# selected.
+transport_params = {
+    "daily": lambda: DailyParams(
+        audio_in_enabled=True,
+        audio_out_enabled=True,
+        vad_analyzer=SileroVADAnalyzer(),
+    ),
+    "twilio": lambda: FastAPIWebsocketParams(
+        audio_in_enabled=True,
+        audio_out_enabled=True,
+        vad_analyzer=SileroVADAnalyzer(),
+    ),
+    "webrtc": lambda: TransportParams(
+        audio_in_enabled=True,
+        audio_out_enabled=True,
+        vad_analyzer=SileroVADAnalyzer(),
+    ),
+}
+
+
+async def run_example(transport: BaseTransport, _: argparse.Namespace, handle_sigint: bool):
+    logger.info(f"Starting bot")
+
+    # Create an HTTP session
+    async with aiohttp.ClientSession() as session:
+        stt = DeepgramSTTService(api_key=os.getenv("DEEPGRAM_API_KEY"))
+
+        tts = InworldHttpTTSService(
+            api_key=os.getenv("INWORLD_API_KEY", ""),
+            voice_id="Ashley",
+            model="inworld-tts-1",
+            aiohttp_session=session,
+        )
+
+        llm = OpenAILLMService(api_key=os.getenv("OPENAI_API_KEY"))
+
+        messages = [
+            {
+                "role": "system",
+                "content": "You are a helpful LLM in a WebRTC call. Your goal is to demonstrate your capabilities in a succinct way. Your output will be converted to audio so don't include special characters in your answers. Respond to what the user said in a creative and helpful way.",
+            },
+        ]
+
+        context = OpenAILLMContext(messages)
+        context_aggregator = llm.create_context_aggregator(context)
+
+        pipeline = Pipeline(
+            [
+                transport.input(),  # Transport user input
+                stt,
+                context_aggregator.user(),  # User responses
+                llm,  # LLM
+                tts,  # TTS
+                transport.output(),  # Transport bot output
+                context_aggregator.assistant(),  # Assistant spoken responses
+            ]
+        )
+
+        task = PipelineTask(
+            pipeline,
+            params=PipelineParams(
+                enable_metrics=True,
+                enable_usage_metrics=True,
+            ),
+        )
+
+        @transport.event_handler("on_client_connected")
+        async def on_client_connected(transport, client):
+            logger.info(f"Client connected")
+            # Kick off the conversation.
+            messages.append({"role": "system", "content": "Please introduce yourself to the user."})
+            await task.queue_frames([context_aggregator.user().get_context_frame()])
+
+        @transport.event_handler("on_client_disconnected")
+        async def on_client_disconnected(transport, client):
+            logger.info(f"Client disconnected")
+            await task.cancel()
+
+        runner = PipelineRunner(handle_sigint=handle_sigint)
+
+        await runner.run(task)
+
+
+if __name__ == "__main__":
+    from pipecat.examples.run import main
+
+    main(run_example, transport_params=transport_params)
--- a/src/pipecat/services/inworld/__init__.py
+++ b/src/pipecat/services/inworld/__init__.py
@@ -1,0 +1,13 @@
+#
+# Copyright (c) 2024–2025, Daily
+#
+# SPDX-License-Identifier: BSD 2-Clause License
+#
+
+import sys
+
+from pipecat.services import DeprecatedModuleProxy
+
+from .tts import *
+
+sys.modules[__name__] = DeprecatedModuleProxy(globals(), "inworld", "inworld.tts")
--- a/src/pipecat/services/inworld/tts.py
+++ b/src/pipecat/services/inworld/tts.py
@@ -68,7 +68,7 @@ def language_to_inworld_language(language: Language) -> Optional[str]:
     return result
 
 
-class InworldTTSService(TTSService):
+class InworldHttpTTSService(TTSService):
     """Inworld HTTP-based TTS service.
 
     Provides text-to-speech using Inworld's HTTP API for simpler, non-streaming
--- a/src/pipecat/services/inworld/tts.py
+++ b/src/pipecat/services/inworld/tts.py
@@ -1,0 +1,265 @@
+#
+# Copyright (c) 2024–2025, Daily
+#
+# SPDX-License-Identifier: BSD 2-Clause License
+#
+
+"""Inworld's text-to-speech service implementations."""
+
+import base64
+import json
+import uuid
+import warnings
+from typing import AsyncGenerator, List, Optional, Union
+
+import aiohttp
+from loguru import logger
+from pydantic import BaseModel, Field
+import io, json, base64
+
+from pipecat.frames.frames import (
+    CancelFrame,
+    EndFrame,
+    ErrorFrame,
+    Frame,
+    StartFrame,
+    StartInterruptionFrame,
+    TTSAudioRawFrame,
+    TTSStartedFrame,
+    TTSStoppedFrame,
+)
+from pipecat.processors.frame_processor import FrameDirection
+from pipecat.services.tts_service import AudioContextWordTTSService, TTSService
+from pipecat.transcriptions.language import Language
+from pipecat.utils.asyncio.watchdog_async_iterator import WatchdogAsyncIterator
+from pipecat.utils.text.base_text_aggregator import BaseTextAggregator
+from pipecat.utils.text.skip_tags_aggregator import SkipTagsAggregator
+from pipecat.utils.tracing.service_decorators import traced_tts
+
+
+def language_to_inworld_language(language: Language) -> Optional[str]:
+    """Convert Pipecat's Language enum to Inworld's language code.
+
+    Args:
+        language: The Language enum value to convert.
+
+    Returns:
+        The corresponding Inworld language code, or None if not supported.
+    """
+    BASE_LANGUAGES = {
+        Language.EN: "en",
+        Language.ES: "es",
+        Language.FR: "fr",
+        Language.KO: "ko",
+        Language.NL: "nl",
+        Language.ZH: "zh",
+    }
+
+    result = BASE_LANGUAGES.get(language)
+
+    # If not found in base languages, try to find the base language from a variant
+    if not result:
+        # Convert enum value to string and get the base language part (e.g. es-ES -> es)
+        lang_str = str(language.value)
+        base_code = lang_str.split("-")[0].lower()
+        # Look up the base code in our supported languages
+        result = base_code if base_code in BASE_LANGUAGES.values() else None
+
+    return result
+
+
+class InworldTTSService(TTSService):
+    """Inworld HTTP-based TTS service.
+
+    Provides text-to-speech using Inworld's HTTP API for simpler, non-streaming
+    synthesis. Suitable for use cases where streaming is not required and simpler
+    integration is preferred.
+    """
+
+    class InputParams(BaseModel):
+        """Input parameters for Inworld HTTP TTS configuration.
+
+        Parameters:
+            language: Language to use for synthesis.
+            speed: Voice speed control (string or float).
+            emotion: List of emotion controls.
+
+                .. deprecated:: 0.0.68
+                        The `emotion` parameter is deprecated and will be removed in a future version.
+        """
+
+        language: Optional[Language] = Language.EN
+        voice_id: str = "Ashley"
+
+    def __init__(
+        self,
+        *,
+        api_key: str,
+        aiohttp_session: aiohttp.ClientSession,
+        model: str = "inworld-tts-1",
+        base_url: str = "https://api.inworld.ai/tts/v1/voice:stream",
+        sample_rate: Optional[int] = 48000,
+        encoding: str = "LINEAR16",
+        params: Optional[InputParams] = None,
+        **kwargs,
+    ):
+        """Initialize the Inworld HTTP TTS service.
+
+        Args:
+            api_key: Inworld API key for authentication.
+            aiohttp_session: Shared aiohttp session for HTTP requests.
+            voice_id: ID of the voice to use for synthesis.
+            model: TTS model to use (e.g., "sonic-2").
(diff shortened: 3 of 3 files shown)


stemper-dev:
If it's a work in progress, delete this from the title and set it as a "draft" not as "open"

src/pipecat/services/inworld/__init__.py
```diff
@@ -0,0 +1,13 @@
...
+# Copyright (c) 2024–2025, Daily
+#
+# SPDX-License-Identifier: BSD 2-Clause License
+#
+
+import sys
+
+from pipecat.services import DeprecatedModuleProxy
+
+from .tts import *
```
markbackman: @padillamt you can leave this file blank. We're moving away from this import pattern.

src/pipecat/services/inworld/tts.py
```diff
@@ -0,0 +1,500 @@
...
+- Comprehensive error handling and metrics tracking
+
+Technical Implementation:
+- Uses aiohttp for HTTP streaming connections
+- Implements JSON line-by-line parsing for streaming responses
+- Handles base64-encoded audio data with proper decoding
+- Manages audio continuity to prevent clicks and artifacts
+- Integrates with Pipecat's frame-based pipeline architecture
+
+Usage:
```
markbackman: For docstring purposes, can you update to the following?

```
Usage::

    tts = InworldHttpTTSService(
        api_key=os.getenv("INWORLD_API_KEY"),
        voice_id="Ashley",
        model="inworld-tts-1",
        aiohttp_session=session
    )
```

src/pipecat/services/inworld/tts.py
```diff
@@ -0,0 +1,500 @@
...
+#
+# SPDX-License-Identifier: BSD 2-Clause License
+#
+
+"""Inworld AI Text-to-Speech Service Implementation.
+
+This module provides integration with Inworld AI's HTTP-based TTS API, enabling
+real-time text-to-speech synthesis with high-quality, natural-sounding voices.
+
+Key Features:
```
markbackman, line 13: For docstring purposes, can you add a new line before a list both here in and after `Technical Implementation`?

src/pipecat/services/inworld/tts.py
markbackman: The same should apply for any other examples, which will apply code syntax styling in the reference docs.

examples/foundational/07aa-interruptible-inworld-http_copy.py
```diff
@@ -0,0 +1,117 @@
+#
```
markbackman: Looks like this file is a duplicate and should be deleted.

src/pipecat/services/inworld/tts.py
```diff
@@ -0,0 +1,500 @@
...
+                model="inworld-tts-1",               # TTS model
+                aiohttp_session=session,             # Required HTTP session
+                sample_rate=48000,                   # Audio quality
+            )
+    """
+
+    class InputParams(BaseModel):
+        """Input parameters for Inworld HTTP TTS configuration.
+
+        Parameters:
```
markbackman, line 146: These should be parameters that your API supports. It looks like this is a copy from Cartesia. Not all services will support the same features, which is totally fine.

src/pipecat/services/inworld/__init__.py
padillamt: Ok, thank you!

src/pipecat/services/inworld/tts.py
padillamt: Thank you!

examples/foundational/07aa-interruptible-inworld-http_copy.py
padillamt: Yes, that is no longer needed and I intended to nuke it. Nuked.

src/pipecat/services/inworld/tts.py
padillamt: Done. Thank you!

src/pipecat/services/inworld/tts.py
padillamt: Yep, that needed to be removed and hadn't been done yet. Thank you for the reminder.

src/pipecat/services/inworld/tts.py
```diff
@@ -0,0 +1,500 @@
...
+
+    def __init__(
+        self,
+        *,
+        api_key: str,
+        aiohttp_session: aiohttp.ClientSession,
+        voice_id: str = "Ashley",
+        model: str = "inworld-tts-1",
+        base_url: str = "https://api.inworld.ai/tts/v1/voice:stream",
+        sample_rate: Optional[int] = 48000,
```
markbackman: This should be `None` so that the sample rate is set in the TTSService by the StartFrame. This is something you can see in the CartesiaHttpTTSService source.

```suggestion
        sample_rate: Optional[int] = None,
```

padillamt:
Draft

src/pipecat/services/inworld/tts.py
padillamt: Done! Thank you.

src/pipecat/services/inworld/tts.py
```diff
@@ -0,0 +1,442 @@
...
+                                audio_data = audio_chunk
+
+                                # ========================================================
+                                # STEP 10: WAV HEADER REMOVAL (CRITICAL FOR AUDIO QUALITY)
+                                # ========================================================
+                                # Each audio chunk may have its own WAV header (44 bytes)
+                                # These headers contain metadata and will sound like clicks if played
+                                # We must strip them from EVERY chunk, not just the first one
+                                if (
+                                    len(audio_chunk) > 44  # Ensure chunk is large enough
```
feifan-ai: By default, inworld tts api returns mp3 instead of wav format. Do you prefer to use wav format here? If so, we need to update the `audio_config`. If it's wav format, why do we need to remove wave header here? Wave file is playable with wav header.

src/pipecat/services/inworld/tts.py
```diff
@@ -0,0 +1,442 @@
...
+            # Start measuring Time To First Byte (TTFB) for performance tracking
+            await self.start_ttfb_metrics()
+
+            # Signal to the pipeline that TTS generation has started
+            # This allows downstream processors to prepare for incoming audio
+            yield TTSStartedFrame()
+
+            # Flag to track if we're processing the first audio chunk
+            # Used for WAV header handling and debugging
+            is_first_chunk = True
```
feifan-ai: Is this param still used? I didn't see any logs related to `is_first_chunk` for debugging.

padillamt: Yes, you're right. Thank you. That was used for fixing a problem but I since handled it differently with the 

` audio_chunk.startswith(b"RIFF") `

check. Will remove.

src/pipecat/services/inworld/tts.py
padillamt: The Pipecat pipeline expects raw PCM, and hence I used a Wav return format and remove the header (first 44 bytes) from the first batch. Header bytes will be incorrectly interpreted as audio.

padillamt:
Ready for Pipecat review. Thank you!

examples/foundational/07aa-interruptible-inworld-http.py
```diff
@@ -0,0 +1,129 @@
+#
```
markbackman: 07aa is now taken. You'll be 07ab.

src/pipecat/services/inworld/__init__.py
```diff
@@ -0,0 +1,13 @@
+#
+# Copyright (c) 2024–2025, Daily
+#
+# SPDX-License-Identifier: BSD 2-Clause License
+#
+
+# import sys
+
+# from pipecat.services import DeprecatedModuleProxy
```
markbackman: You can just leave this file blank.

src/pipecat/services/inworld/tts.py
```diff
@@ -0,0 +1,600 @@
...
+                    temperature=0.8,
+                ),
+            )
+    """
+
+    class InputParams(BaseModel):
+        """Input parameters for Inworld TTS configuration.
+
+        Parameters:
+            voice_id: Voice selection for speech synthesis (e.g., "Ashley", "Hades").
```
markbackman: I know the conventions aren't clearly communicated, but things that are common are best left in the args. So I would move `voice_id` and maybe `model` to the args. Leaving only `temperature` here. But, that provides room for growth as you have more features.

src/pipecat/services/inworld/tts.py
```diff
@@ -0,0 +1,600 @@
...
+        voice_id: Optional[str] = "Ashley"  # defaults to the Ashley voice
+        model: Optional[str] = "inworld-tts-1"  # defaults to the inworld-tts-1 model
+        temperature: Optional[float] = None  # optional temperature control (range: [0, 2])
+
+    def __init__(
+        self,
+        *,
+        api_key: str,
+        aiohttp_session: aiohttp.ClientSession,
+        streaming: bool = True,
```
markbackman, line 164: Nice! This is a better solution than two classes. This will be easier to maintain going forward.

src/pipecat/services/inworld/tts.py
```diff
@@ -0,0 +1,600 @@
...
+        base_url: Optional[str] = None,
```
markbackman: It looks like `streaming` controls the `base_url`. Do you also want to expose the URL or should the `streaming` arg be used to set it. I would suggest one or the other, but not both as that could be confusing.

src/pipecat/services/inworld/tts.py
```diff
@@ -0,0 +1,600 @@
...
+                temperature=0.8,
+            ),
+        )
+"""
+
+import base64
+import io
+import json
+import uuid
+import warnings
```
markbackman: You can remove this. It's unused.
```suggestion
```

src/pipecat/services/inworld/tts.py
```diff
@@ -0,0 +1,600 @@
...
+                base_url = "https://api.inworld.ai/tts/v1/voice:stream"  # Streaming endpoint
+            else:
+                base_url = "https://api.inworld.ai/tts/v1/voice"  # Non-streaming endpoint
+        self._base_url = base_url  # API endpoint URL
+
+        # Build settings dictionary that matches Inworld's API expectations
+        # This will be sent as JSON payload in each TTS request
+        # Note: Language is automatically inferred from text by Inworld's models
+        self._settings = {
+            "voiceId": params.voice_id or "Ashley",  # Voice selection from params
```
markbackman: Since you're already setting defaults above, I think this can be simplified:
```suggestion
            "voiceId": params.voice_id,  # Voice selection from params
```

src/pipecat/services/inworld/tts.py
```diff
@@ -0,0 +1,600 @@
...
+            "modelId": params.model or "inworld-tts-1",  # TTS model selection from params
```
markbackman: Same here:
```suggestion
            "modelId": params.model,  # TTS model selection from params
```

src/pipecat/services/inworld/tts.py
```diff
@@ -0,0 +1,600 @@
...
+                "sample_rate_hertz": 0,  # Will be set in start() from parent service
+            },
+        }
+
+        # Add optional temperature parameter if provided (valid range: [0, 2])
+        if params.temperature is not None:
+            self._settings["temperature"] = params.temperature
+
+        # Register voice and model with parent service for metrics and tracking
+        self.set_voice(params.voice_id or "Ashley")  # Used for logging and metrics
```
markbackman: Set this from self._settings["voice_id"] now since you've defined it above.

```suggestion
        self.set_voice(self._settings["voice_id"])  # Used for logging and metrics
```

src/pipecat/services/inworld/tts.py
```diff
@@ -0,0 +1,600 @@
...
+        self.set_model_name(params.model or "inworld-tts-1")  # Used for performance tracking
```
markbackman: Same here.

src/pipecat/services/inworld/tts.py
```diff
@@ -0,0 +1,600 @@
...
+            # ================================================================================
+            # Log any unexpected errors and notify the pipeline
+            logger.error(f"{self} exception: {e}")
+            await self.push_error(ErrorFrame(f"Error generating TTS: {e}"))
+        finally:
+            # ================================================================================
+            # STEP 8: CLEANUP AND COMPLETION
+            # ================================================================================
+            # Always stop metrics tracking, even if errors occurred
+            await self.stop_ttfb_metrics()
```
markbackman: This is where you want to stop all metrics, which calls:
```
    async def stop_all_metrics(self):
        """Stop all active metrics collection."""
```

stop_ttfb_metrics was likely called earlier, so it's redundant, but this will also stop the processing metrics.

```suggestion
            await self.stop_all_metrics()
```

markbackman: Instead, `stop_ttfb_metrics()` should be moved to the place where you receive the first bytes (e.g. to measure "time to first byte".)

src/pipecat/services/inworld/tts.py
markbackman: Actually, you have a bunch of unused imports. The imports should be:
```
import base64
import json

import aiohttp
from loguru import logger
from pydantic import BaseModel

    CancelFrame,
    EndFrame,
    ErrorFrame,
    Frame,
    StartFrame,
    TTSAudioRawFrame,
    TTSStartedFrame,
    TTSStoppedFrame,
)
from pipecat.services.tts_service import TTSService
from pipecat.utils.tracing.service_decorators import traced_tts
```

examples/foundational/07aa-interruptible-inworld-http.py
```diff
@@ -0,0 +1,129 @@
...
+    ),
+}
+
+
+async def run_example(transport: BaseTransport, _: argparse.Namespace, handle_sigint: bool):
+    logger.info(f"Starting bot")
+
+    # Create an HTTP session
+    async with aiohttp.ClientSession() as session:
+        stt = OpenAISTTService(
```
markbackman: I know it's one more API to get, but OpenAISTTService is very slow. Deepgram is used in most examples and is a safe choice. Gladia, Speechmatics, Assembly are all great options, too. Perhaps Deepgram is the best pick since it's used elsewhere:
```
```

examples/foundational/07aa-interruptible-inworld-http.py
```diff
@@ -0,0 +1,129 @@
...
+                tts,  # TTS
+                transport.output(),  # Transport bot output
+                context_aggregator.assistant(),  # Assistant spoken responses
+            ]
+        )
+
+        task = PipelineTask(
+            pipeline,
+            params=PipelineParams(
+                audio_out_sample_rate=24000,
```
markbackman: This is the default value, so you can omit.
```suggestion
```

markbackman:
Looks really good! Just a handful of comments, but nothing major.

Please add a concise changelog entry in addition to working through the feedback. Great work. 👏

examples/foundational/07aa-interruptible-inworld-http.py
padillamt: Done!

src/pipecat/services/inworld/__init__.py
padillamt: Done!

src/pipecat/services/inworld/tts.py
padillamt: Thank you for your help!

src/pipecat/services/inworld/tts.py
padillamt: Done!

src/pipecat/services/inworld/tts.py
padillamt: Done!

examples/foundational/07aa-interruptible-inworld-http.py
padillamt: Done to renamed 07ab file

examples/foundational/07aa-interruptible-inworld-http.py
padillamt: Thank you for the tip. Tried it (can get a free Deepgram API key to try it out....nice) and it's indeed noticeably faster. Much appreciated!
This is in the 7ab file

src/pipecat/services/inworld/tts.py
```diff
@@ -0,0 +1,592 @@
...
+                try:
+                    # ================================================================
+                    # PARSE JSON AND EXTRACT AUDIO
+                    # ================================================================
+                    # Parse the JSON line - should contain audio data
+                    chunk_data = json.loads(line_str)
+
+                    # Check if this line contains audio content
+                    # Inworld's response format: {"result": {"audioContent": "base64data"}}
+                    if "result" in chunk_data and "audioContent" in chunk_data["result"]:
```
markbackman, line 470: Add `await self.stop_ttfb_metrics()` here too.

src/pipecat/services/inworld/tts.py
padillamt: Yes, good point. Removed exposure of the URL and have it now set internally by the `streaming` arg since we need that anyway. One less parameter for the user to worry about. Thank you!

src/pipecat/services/inworld/tts.py
padillamt: Done!

src/pipecat/services/inworld/tts.py
```diff
@@ -0,0 +1,589 @@
...
+
+        for i in range(0, len(audio_data), CHUNK_SIZE):
+            chunk = audio_data[i : i + CHUNK_SIZE]
+            if len(chunk) > 0:
+                await self.stop_ttfb_metrics()
+                yield TTSAudioRawFrame(
+                    audio=chunk,
+                    sample_rate=self.sample_rate,
+                    num_channels=1,
+                )
```
filipi87, line 540: I guess, in theory, all of this could be handled inside `_process_audio_chunk`, so both `_process_non_streaming_response` and `_process_streaming_response` could call it. Maybe we could even rename it to `_process_audio`. Inside it, we would:
- Decode the base64 audio data to binary
- Strip the WAV header
- Chunk the audio data if needed
- Create the TTSAudioRawFrame

What do you think ?

src/pipecat/services/inworld/tts.py
padillamt: Done.

examples/foundational/07ab-interruptible-inworld-http.py
```diff
@@ -0,0 +1,124 @@
...
+        # Inworld TTS Service - Unified streaming and non-streaming
+        # Set streaming=True for real-time audio, streaming=False for complete audio generation
+        streaming = True  # Toggle this to switch between modes
+
+        tts = InworldTTSService(
+            api_key=os.getenv("INWORLD_API_KEY", ""),
+            aiohttp_session=session,
+            streaming=streaming,  # True: real-time chunks, False: complete audio then playback
+            params=InworldTTSService.InputParams(
+                voice_id="Ashley",
```
markbackman: With the move of `voice_id` and `model` out of `InputParams`, this example needs to change to match.

src/pipecat/services/inworld/tts.py
padillamt: As per our Slack conversation, I'll look into this once I'm back from OOO. This change will be addressed through a different PR in the future if we do it. Thank you for the suggestion.

examples/foundational/07ab-interruptible-inworld-http.py
padillamt: Ah, thank you.  Forgot to push that change. Updated.

markbackman (approved):
LGTM! Nice work 🙌

padillamt:
Thank you very much, Mark and Filipi!

Issue #904, opened 2024-12-22
Optimize Session Startup by Parallelizing Component Initialization


golbin:
# Description
Is this reporting a bug or feature request?  
This is a feature request to optimize the session startup process.

### Issue description
The current session startup process initializes components (TTS, LLM, STT, and Daily Join) sequentially, which results in higher latency before the session becomes operational. By refactoring the process to initialize these components in parallel, the startup time can be significantly reduced.

### Expected behavior
The initialization of TTS, LLM, STT, and Daily Join should occur in parallel, resulting in a noticeable reduction in session startup time.

### Actual behavior
Components are initialized sequentially, leading to a slower startup time.

aconchillo:
Thank you for reporting this @golbin . The reason behind the current behavior is to make all the components are successfully initialzed in order so we don't push frames prematurely. But yes, maybe we could possibly initialize everything in parallel and just wait for the last service being initialized before starting pushing frames.

golbin:
Yes, handling parallel processing might be a bit challenging, but it could significantly reduce the startup time.

If we can reuse certain components, the startup could become extremely fast. Additionally, if we can maintain connections to TTS and STT even after the session ends, we could potentially onboard new users with almost no waiting time.

For a simpler approach, we could start by parallelizing only the Daily transport. This alone could substantially reduce the waiting time, as the time spent on the Daily join process is roughly equivalent to the combined time required to connect to TTS and STT.

| Time                   | Event                                     | Time Difference |
|------------------------|-------------------------------------------|-----------------|
| 2024-12-21 10:36:27.580 | Modules: Session started                 | 0.001 seconds   |
| 2024-12-21 10:36:27.580 | VAD: Set VAD parameters                  | 0.000 seconds   |
| 2024-12-21 10:36:27.649 | Audio: Silero VAD loaded                 | 0.069 seconds   |
| 2024-12-21 10:36:27.684 | Processors: Link frame processors        | 0.035 seconds   |
| 2024-12-21 10:36:27.909 | Services: Connect to Cartesia            | 0.225 seconds   |
| 2024-12-21 10:36:28.178 | Daily: Join session                      | 0.269 seconds   |
| 2024-12-21 10:36:28.616 | Daily: Joined session                    | 0.438 seconds   |
| 2024-12-21 10:36:29.042 | Modules: Sync session state              | 0.426 seconds   |
| 2024-12-21 10:36:29.043 | Modules: Update session state            | 0.001 seconds   |
[...]

PR #2215, opened 2025-07-17
Add Parallel Startup Option for Pipeline Services

golbin:
#### Summary
- Introduced the `parallel_startup` parameter in `PipelineParams` and `PipelineTask` to enable experimental parallel initialization of pipeline services (such as `AIService`, input/output transports, etc.).
- When `parallel_startup` is set to `True`, all relevant services are started concurrently using `asyncio.gather`, instead of sequentially.
- This change aims to improve pipeline startup performance, especially for pipelines with multiple services.

#### Performance Improvement
- With parallel startup enabled, service initialization time was reduced significantly:
    - **Serial:** Services started in 2.23s, 2.57s, 3.11s
    - **Parallel:** Services started in 1.12s, 1.46s, 1.64s
- This demonstrates nearly a 2x improvement in startup time.
- I measured it three times on my MacBook.

#### Additional Notes
- The default value for `parallel_startup` remains `False` to ensure backward compatibility.
- This is an experimental feature and may require further testing for stability.

### DIFF
--- a/src/pipecat/pipeline/task.py
+++ b/src/pipecat/pipeline/task.py
@@ -663,15 +663,10 @@ class PipelineTask(BasePipelineTask):
         )
         start_frame.metadata = self._params.start_metadata
 
-        pre_startup_services = self._collect_pre_startup_services(self._pipeline)
-        if pre_startup_services:
-            start_time = time.time()
-            if self._params.parallel_startup:
+        if self._params.parallel_startup:
+            pre_startup_services = self._collect_pre_startup_services(self._pipeline)
+            if pre_startup_services:
                 await asyncio.gather(*(s.start(start_frame) for s in pre_startup_services))
-            else:
-                for s in pre_startup_services:
-                    await s.start(start_frame)
-            logger.info(f"AI and transport services started in {time.time() - start_time:.2f}s")
 
         await self._source.queue_frame(start_frame, FrameDirection.DOWNSTREAM)
 
--- a/src/pipecat/pipeline/task.py
+++ b/src/pipecat/pipeline/task.py
@@ -45,6 +45,9 @@ from pipecat.pipeline.base_pipeline import BasePipeline
 from pipecat.pipeline.base_task import BasePipelineTask, PipelineTaskParams
 from pipecat.pipeline.task_observer import TaskObserver
 from pipecat.processors.frame_processor import FrameDirection, FrameProcessor, FrameProcessorSetup
+from pipecat.services.ai_service import AIService
+from pipecat.transports.base_input import BaseInputTransport
+from pipecat.transports.base_output import BaseOutputTransport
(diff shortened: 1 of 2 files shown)


golbin:
How about this proposal? @aconchillo

I think it might work better with the following patch: https://github.com/pipecat-ai/pipecat/pull/2526

aconchillo:
Thank you @golbin ! This is a neat idea, I'll take a look. The reason for being serial is to avoid processors pushing things down the pipeline when other processors are not ready to receive frames yet. It makes things more deterministic. If we do it in parallel it would be harder to know when the pipeline is fully ready.

aconchillo:
> Thank you @golbin ! This is a neat idea, I'll take a look. The reason for being serial is to avoid processors pushing things down the pipeline when other processors are not ready to receive frames yet. It makes things more deterministic. If we do it in parallel it would be harder to know when the pipeline is fully ready.

I'll think about it though. It might be possible to start in parallel and then each of the processors (not only services) report if they have started. So, the speed would be `max(processor1, processor2....)` instead of `processor1 + processor2 ....` as today.

golbin:
Yes @aconchillo parallelism is tricky. But if we can apply it to services only, it actually becomes more predictable.

Or we could try reflecting it only in the transport layer, since the connection is usually the slowest part and we can control the waiting events there.

golbin:
As we use Pipecat more seriously, the number of processors and services in the pipeline will only keep growing, so parallel execution will become increasingly important.

Just like Pipecat reducing voice chat latency directly improves user experience, the latency between a user pressing a button and actually starting the conversation is also a critical part of the user experience. 🙂

aconchillo:
> As we use Pipecat more seriously, the number of processors and services in the pipeline will only keep growing, so parallel execution will become increasingly important.
> 
> Just like Pipecat reducing voice chat latency directly improves user experience, the latency between a user pressing a button and actually starting the conversation is also a critical part of the user experience. 🙂

Agreed. Everything after the pipeline is started is optimized for real-time and happens concurrently. You also have parallel pipeline for more advanced things. But yeah, the startup can definitely be improved! Thank you so much for spending the time to think about this and even provide POCs! :bow:

venkatesan-boopalan-zoomrx:
@golbin What is the plan for releasing the parallel support option? We’re currently exploring ways to reduce initial latency, and your changes would be helpful for us.

golbin:
> @golbin What is the plan for releasing the parallel support option? We’re currently exploring ways to reduce initial latency, and your changes would be helpful for us.

It depends on @aconchillo’s team... ☺️

aconchillo:
I believe this might not be necessary anymore. See https://github.com/pipecat-ai/pipecat/issues/904#issuecomment-3730931142

markbackman:
Closing per @aconchillo's comments.

PR #2217, opened 2025-07-17
feat: Add Simli Trinity models support to pipecat

Antonyesk601:
A few modifications needed to support the new Simli models in pipecat!
Adds client side buffering to make sure audio is being fed to Simli properly

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,7 +26,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   pushed by the `BaseInputTransport` at Start and any time a
   `VADParamsUpdateFrame` is received.
 
-- Add support for Simli Trinity Avatars, a new `is_trinity_avatar` parameter is added to indicate whether the Simli `faceId` is a trinity faceId or not which is essential for getting a consistent experience with trinity faces.
+- Added support for Simli Trinity Avatars. A new `is_trinity_avatar` parameter
+  has been introduced to specify whether the provided `faceId` corresponds to a
+  Trinity avatar, which is required for optimal Trinity avatar performance.
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   pushed by the `BaseInputTransport` at Start and any time a
   `VADParamsUpdateFrame` is received.
 
+- Add support for Simli Trinity Avatars, a new `is_trinity_avatar` parameter is added to indicate whether the Simli `faceId` is a trinity faceId or not which is essential for getting a consistent experience with trinity faces.
+
 ### Changed
 
 - Two package dependencies have been updated:
--- a/src/pipecat/services/simli/video.py
+++ b/src/pipecat/services/simli/video.py
@@ -77,9 +77,8 @@ class SimliVideoService(FrameProcessor):
 
         self._audio_task: asyncio.Task = None
         self._video_task: asyncio.Task = None
-        self._started = False
         self._is_trinity_avatar = is_trinity_avatar
-        self._previously_interrupted = True
+        self._previously_interrupted = is_trinity_avatar
         self._audio_buffer = bytearray()
 
     async def _start_connection(self):
@@ -153,7 +152,7 @@ class SimliVideoService(FrameProcessor):
                 resampled_frames = self._simli_resampler.resample(old_frame)
                 for resampled_frame in resampled_frames:
                     audioBytes = resampled_frame.to_ndarray().astype(np.int16).tobytes()
-                    if self._previously_interrupted and self._is_trinity_avatar:
+                    if self._previously_interrupted:
                         self._audio_buffer.extend(audioBytes)
                         if len(self._audio_buffer) >= 128000:
                             try:
@@ -176,12 +175,9 @@ class SimliVideoService(FrameProcessor):
                     await self._simli_client.playImmediate(self._audio_buffer)
                     self._previously_interrupted = False
                     self._audio_buffer = bytearray()
-
             except Exception as e:
                 logger.exception(f"{self} exception: {e}")
             return
-            self._started = True
-            await self._simli_client.send((0).to_bytes(1, "big") * 6000)
         elif isinstance(frame, (EndFrame, CancelFrame)):
             await self._stop()
         elif isinstance(frame, (StartInterruptionFrame, UserStartedSpeakingFrame)):
--- a/src/pipecat/services/simli/video.py
+++ b/src/pipecat/services/simli/video.py
@@ -48,7 +48,7 @@ class SimliVideoService(FrameProcessor):
         use_turn_server: bool = False,
         latency_interval: int = 0,
         simli_url: str = "https://api.simli.ai",
-        isTrinityAvatar: bool = False,
+        is_trinity_avatar: bool = False,
     ):
         """Initialize the Simli video service.
 
@@ -57,6 +57,7 @@ class SimliVideoService(FrameProcessor):
             use_turn_server: Whether to use TURN server for connection. Defaults to False.
             latency_interval: Latency interval setting for sending health checks to check the latency to Simli Servers. Defaults to 0.
             simli_url: URL of the simli servers. Can be changed for custom deployments of enterprise users.
+            is_trinity_avatar: boolean to tell simli client that this is a Trinity avatar which reduces latency when using Trinity.
 
         """
         super().__init__()
@@ -77,7 +78,7 @@ class SimliVideoService(FrameProcessor):
         self._audio_task: asyncio.Task = None
         self._video_task: asyncio.Task = None
         self._started = False
-        self._is_gs_avatar = isTrinityAvatar
+        self._is_trinity_avatar = is_trinity_avatar
         self._previously_interrupted = True
         self._audio_buffer = bytearray()
 
@@ -134,7 +135,6 @@ class SimliVideoService(FrameProcessor):
         await super().process_frame(frame, direction)
         if isinstance(frame, StartFrame):
             await self._start_connection()
-            await self.push_frame(frame, direction)
         elif isinstance(frame, TTSAudioRawFrame):
             # Send audio frame to Simli
             try:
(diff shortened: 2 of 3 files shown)


src/pipecat/services/simli/video.py
```diff
@@ -45,24 +47,39 @@ def __init__(
...
     ):
         """Initialize the Simli video service.
 
         Args:
             simli_config: Configuration object for Simli client settings.
             use_turn_server: Whether to use TURN server for connection. Defaults to False.
-            latency_interval: Latency interval setting for video processing. Defaults to 0.
+            latency_interval: Latency interval setting for sending health checks to check the latency to Simli Servers. Defaults to 0.
+            simli_url: URL of the simli servers. Can be changed for custom deployments of enterprise users.
+
```
kompfner, line 61: Nit: we probably want an explanation of `isTrinityAvatar` here

src/pipecat/services/simli/video.py
```diff
@@ -118,8 +133,8 @@ async def process_frame(self, frame: Frame, direction: FrameDirection):
         """
         await super().process_frame(frame, direction)
         if isinstance(frame, StartFrame):
-            await self.push_frame(frame, direction)
             await self._start_connection()
+            await self.push_frame(frame, direction)
```
kompfner: Do we need this here now that we `push_frame` at the end of `process_frame`? Wondering if we're now pushing `StartFrame` twice.

src/pipecat/services/simli/video.py
```diff
@@ -137,19 +152,47 @@ async def process_frame(self, frame: Frame, direction: FrameDirection):
...
+            try:
+                if self._previously_interrupted and len(self._audio_buffer) > 0:
+                    await self._simli_client.playImmediate(self._audio_buffer)
+                    self._previously_interrupted = False
+                    self._audio_buffer = bytearray()
+
+            except Exception as e:
+                logger.exception(f"{self} exception: {e}")
+            return
+        if isinstance(frame, StartFrame):
```
kompfner: Is it intentional that you're handling `StartFrame` twice in this function? (It's handled at the top as well)

src/pipecat/services/simli/video.py
```diff
@@ -137,19 +152,47 @@ async def process_frame(self, frame: Frame, direction: FrameDirection):
...
+            if not self._preinitialize:
```
kompfner: I may totally be missing something, but I'm not seeing `_preinitialize` or `_started` used anywhere

src/pipecat/services/simli/video.py
```diff
@@ -45,24 +47,39 @@ def __init__(
...
             use_turn_server: Whether to use TURN server for connection. Defaults to False.
-            latency_interval: Latency interval setting for video processing. Defaults to 0.
+            latency_interval: Latency interval setting for sending health checks to check the latency to Simli Servers. Defaults to 0.
+            simli_url: URL of the simli servers. Can be changed for custom deployments of enterprise users.
+
         """
         super().__init__()
-        self._simli_client = SimliClient(simli_config, use_turn_server, latency_interval)
+        self._initialized = False
+        simli_config.maxIdleTime += 5
```
kompfner, line 65: What's the reasoning behind boosting these values after the developer has set them?

src/pipecat/services/simli/video.py
```diff
@@ -137,19 +152,47 @@ async def process_frame(self, frame: Frame, direction: FrameDirection):
...
+            await self._simli_client.send((0).to_bytes(1, "big") * 6000)
         elif isinstance(frame, (EndFrame, CancelFrame)):
             await self._stop()
-            await self.push_frame(frame, direction)
-        elif isinstance(frame, StartInterruptionFrame):
-            await self._simli_client.clearBuffer()
-            await self.push_frame(frame, direction)
-        else:
-            await self.push_frame(frame, direction)
+        elif isinstance(frame, (StartInterruptionFrame, UserStartedSpeakingFrame)):
```
kompfner, line 183: Why do we need to handle both kinds of frames here? Wouldn't one do the trick?

src/pipecat/services/simli/video.py
```diff
@@ -137,19 +152,47 @@ async def process_frame(self, frame: Frame, direction: FrameDirection):
...
+            if not self._previously_interrupted:
+                await self._simli_client.clearBuffer()
+            self._previously_interrupted = True and self._is_gs_avatar
```
kompfner: Isn't this equivalent to `self._previously_interrupted = self._is_gs_avatar`?

src/pipecat/services/simli/video.py
```diff
@@ -45,24 +47,39 @@ def __init__(
         simli_config: SimliConfig,
         use_turn_server: bool = False,
         latency_interval: int = 0,
+        simli_url: str = "https://api.simli.ai",
+        isTrinityAvatar: bool = False,
```
kompfner: We usually use snake case—`is_trinity_avatar`

src/pipecat/services/simli/video.py
```diff
@@ -45,24 +47,39 @@ def __init__(
...
-        self._pipecat_resampler_event = asyncio.Event()
         self._pipecat_resampler: AudioResampler = None
+        self._pipecat_resampler_event = asyncio.Event()
         self._simli_resampler = AudioResampler("s16", "mono", 16000)
 
-        self._initialized = False
         self._audio_task: asyncio.Task = None
         self._video_task: asyncio.Task = None
+        self._started = False
+        self._is_gs_avatar = isTrinityAvatar
```
kompfner: Is there a reason to give this variable a different name than the parameter that the dev passes in?

src/pipecat/services/simli/video.py
```diff
@@ -137,19 +152,47 @@ async def process_frame(self, frame: Frame, direction: FrameDirection):
 
                 resampled_frames = self._simli_resampler.resample(old_frame)
                 for resampled_frame in resampled_frames:
-                    await self._simli_client.send(
-                        resampled_frame.to_ndarray().astype(np.int16).tobytes()
-                    )
+                    audioBytes = resampled_frame.to_ndarray().astype(np.int16).tobytes()
+                    if self._previously_interrupted and self._is_gs_avatar:
```
kompfner: I believe you don't need to check `self._is_gs_avatar` here, since `self._previously_interrupted` can only be `True` if `self._is_gs_avatar` is, too...right?

src/pipecat/services/simli/video.py
```diff
@@ -137,19 +152,47 @@ async def process_frame(self, frame: Frame, direction: FrameDirection):
...
+                        self._audio_buffer.extend(audioBytes)
+                        if len(self._audio_buffer) >= 128000:
```
kompfner, line 157: Trying to wrap my head around how this buffering is meant to work.

Can you let me know if the following understanding is correct:
- If using the new avatar type, you want to buffer up to the first 8 seconds (is that correct?) of audio of each bot response before sending it to the server
- If the bot response goes for more than 8 seconds, you start sending any incoming audio bytes as they come in, without buffering
- You only want to do this if this is a response to a user utterance (so, not the very first time the bot speaks)

If so: wouldn't that possibly create quite significant delays in the conversation?

(Pardon me if I'm misreading, which I may be)

src/pipecat/services/simli/video.py
Antonyesk601: Accounts for some delay in actual between maxIdleTime on simli side vs preceived time on pipecat. Doesn't affect the end price since we round down to closest minute. Just makes for a nicer user experience

src/pipecat/services/simli/video.py
kompfner: Looks like `_started` is still around but isn't really used for anything?

src/pipecat/services/simli/video.py
kompfner: I believe this hasn't been resolved yet

kompfner (changes requested):
Thanks for the changes! Still have a few unresolved comments

Antonyesk601:
> Thanks for the changes! Still have a few unresolved comments

Pushed the requested changes

kompfner:
It looks like there are still unresolved comments—the most important being [this one](https://github.com/pipecat-ai/pipecat/pull/2217#discussion_r2214166054). I may be misreading/misunderstanding the code (and pardon me if that's case—feel free to reply and tell me I'm wrong 😛 ), but given my understanding of it my concern would be that an 8-second delay would be too much for a real-time conversation.

Antonyesk601:
I did clarify here [here](https://github.com/pipecat-ai/pipecat/pull/2217#discussion_r2215370461) that there's indeed a 4 second delay (128kb is 4 seconds on pcm16 16khz mono) but it's required for this iteration of trinity models and you can actually see it in action on https://simli.com/gs-demo which is powered by pipecat. Additionally 4 seconds of audio usually gets generated in less than 4 seconds so the effective delay is less than 4 seconds

kompfner:
> I did clarify here [here](https://github.com/pipecat-ai/pipecat/pull/2217#discussion_r2215370461) that there's indeed a 4 second delay (128kb is 4 seconds on pcm16 16khz mono) but it's required for this iteration of trinity models and you can actually see it in action on https://simli.com/gs-demo which is powered by pipecat. Additionally 4 seconds of audio usually gets generated in less than 4 seconds so the effective delay is less than 4 seconds

Ah apologies, I didn't see your clarification (and somehow still don't? that link isn't working for me). But thanks for explaining again! Sounds good 👍

kompfner (approved):

markbackman:
@Antonyesk601 can you also add a changelog entry?

CHANGELOG.md
```diff
@@ -26,6 +26,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   pushed by the `BaseInputTransport` at Start and any time a
   `VADParamsUpdateFrame` is received.
 
+- Add support for Simli Trinity Avatars, a new `is_trinity_avatar` parameter is added to indicate whether the Simli `faceId` is a trinity faceId or not which is essential for getting a consistent experience with trinity faces.
```
markbackman: Nit: wording clarification (and line break at 80 chars).

```suggestion
- Added support for Simli Trinity Avatars. A new `is_trinity_avatar` parameter
  has been introduced to specify whether the provided `faceId` corresponds to a
  Trinity avatar, which is required for optimal Trinity avatar performance.
```

markbackman (approved):
LGTM!

markbackman (approved):
LGTM!

PR #2223, opened 2025-07-20
Add text input handling to unify context for realtimeInput stream of GeminiMultimodalLiveService

getchannel:
## Summary
This pull request fixes a contextual disconnect in the Gemini Multimodal Live API integration.  
Previously, text inputs went through the **clientContent** stream, while audio and video used **realtimeInput**.  
Because the Gemini API processes these streams concurrently, their contexts are not guaranteed to interleave, which can cause text to drift out of sync with simultaneous audio/video.

The PR adds a pathway that sends text through **realtimeInput**, so text, audio, and video now share the same context stream for coherent, synchronized interactions.

---

### `src/pipecat/services/gemini_multimodal_live/events.py`
- **RealtimeInput** model: added optional `text` field (parity with `BidiGenerateContentRealtimeInput`).
- Added **TextInputMessage** class to carry text into the `realtimeInput` stream.

### `src/pipecat/services/gemini_multimodal_live/gemini.py`
- Added new branch in `process_frame` to treat `LLMTextFrame` as a primary input.
- Implemented `_send_user_text` to forward text via `realtimeInput`.
- After sending text, emit a `clientContent` message with `turnComplete: true` to trigger a model response for text‑only turns.

---

## Backward Compatibility
Fully preserved.  
Existing workflows that send text through `LLMMessagesAppendFrame` (via **clientContent**) continue to function unchanged.  
This PR simply introduces an **additional** real‑time text pathway.

### DIFF
--- a/src/pipecat/services/gemini_multimodal_live/gemini.py
+++ b/src/pipecat/services/gemini_multimodal_live/gemini.py
@@ -971,9 +971,7 @@ class GeminiMultimodalLiveLLMService(LLMService):
         evt = events.TextInputMessage.from_text(text)
         await self.send_client_event(evt)
         # After sending text, we need to signal that the turn is complete.
-        evt = events.ClientContentMessage.model_validate(
-            {"clientContent": {"turnComplete": True}}
-        )
+        evt = events.ClientContentMessage.model_validate({"clientContent": {"turnComplete": True}})
         await self.send_client_event(evt)
 
     async def _send_user_video(self, frame):
--- a/src/pipecat/services/gemini_multimodal_live/gemini.py
+++ b/src/pipecat/services/gemini_multimodal_live/gemini.py
@@ -733,6 +733,8 @@ class GeminiMultimodalLiveLLMService(LLMService):
                 # Support just one tool call per context frame for now
                 tool_result_message = context.messages[-1]
                 await self._tool_result(tool_result_message)
+        elif isinstance(frame, LLMTextFrame):
+            await self._send_user_text(frame.text)
         elif isinstance(frame, InputAudioRawFrame):
             await self._send_user_audio(frame)
             await self.push_frame(frame, direction)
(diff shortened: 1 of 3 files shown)


getchannel:
Sidenote from the API docs mention that `mediaChunks[]`  DEPRECATED: Use one of audio, video, or text instead. 

We should consider updating the `mediaChunks` in another PR.

getchannel:
@markbackman @vipyne thoughts on this? thanks

src/pipecat/services/gemini_multimodal_live/gemini.py
```diff
@@ -733,6 +733,8 @@ async def process_frame(self, frame: Frame, direction: FrameDirection):
                 # Support just one tool call per context frame for now
                 tool_result_message = context.messages[-1]
                 await self._tool_result(tool_result_message)
+        elif isinstance(frame, LLMTextFrame):
```
vipyne: I'm confused by this. why would we send LLM response text back to the LLM as if it was user input?

Maybe I am missing something. Do you have a example of this feature in action (I assume it would require a custom frontend.  Could be a great usecase for [voice ui kit](https://github.com/pipecat-ai/voice-ui-kit).

getchannel: Oh shoot, good catch. Looking at the pre-existing frames, should i create a new frame `UserTextFrame` for this purpose?

`TextFrame` - TTS would synthesize user input 
`TranscriptionFrame` - Already used for speech, can't distinguish programmatic text

getchannel: or `InputTextRawFrame` to keep with the existing pattern

getchannel: updated with `InputTextRawFrame` to create new frame for this purpose in line with existing pattern.

getchannel:
> hi @getchannel thanks for taking a look at this. It would be great to have an example (in foundational) in lieu of a test for now.
> 
> although on that note- re: "...mediaChunks[] DEPRECATED...", I've been slowly trying to migrate us to use gemini's genai module (kinda messy WIP branch [here](https://github.com/pipecat-ai/pipecat/tree/vp-genai-client-tuesday)). Let me know if you'd like to pair or help out on this.

Would be happy to help, in the past the ai studio endpoint has been better for me, but as features get released on vertex and the rate limits increase, that might be the future going forward. I won't switch my stack over yet personally, until files api, and a few other features make it over there. I will however evolve to whichever one seems like the path moving forward, I wish the google team provided more guidance on this.

aconchillo (approved):

aconchillo:
LGTM!

aconchillo:
Just missing CHANGELOG entry

vipyne:
added changelog here: https://github.com/pipecat-ai/pipecat/pull/2305

PR #2229, opened 2025-07-21
Add Mistral LLM Support to Pipecat

obigroup:
This PR introduces integration support for Mistral's Large Language Models within the Pipecat framework.

## 🚀 Features

- **Mistral LLM Integration**: Complete implementation of Mistral models support in Pipecat
- **Core Functionality**: Basic LLM operations are fully functional and tested
- **Function Calling Ready**: Architecture designed to support function calling (should work but needs additional testing)

## 🧪 Testing Status

- ✅ **Core LLM functionality**: Tested and working
- ⚠️ **Function calling**: Not yet tested but implementation should be compatible
- ⚠️ **Context management**: Implementation complete but open to feedback and improvements

## 🤔 Areas for Review

I'm particularly interested in feedback on:

- **Context Management**: The current implementation handles contexts but I'm not 100% confident it's optimal. Any suggestions for improvements would be greatly appreciated.
- **Function Calling**: While the architecture should support it, additional testing and validation would be valuable.

## 📝 Implementation Notes

The integration follows Pipecat's established patterns for LLM providers and should be compatible with existing workflows and configurations.

## 🤝 Collaboration Welcome

@GaspardBT and the Mistral team - if you'd like to take a look and potentially contribute to this integration, this PR is open for collaboration! Your insights would be invaluable for ensuring optimal integration with Mistral's capabilities.

## 📋 Checklist

- [x] Basic LLM functionality implemented
- [x] Core integration tested
- [ ] Function calling thoroughly tested
- [ ] Context management reviewed and optimized
- [ ] Documentation updated

---

*This is a community contribution aimed at expanding Pipecat's LLM ecosystem. Feedback and contributions are welcome!*

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -96,6 +96,7 @@ ultravox = [ "transformers>=4.48.0", "vllm~=0.7.3" ]
 webrtc = [ "aiortc~=1.11.0", "opencv-python~=4.11.0.86" ]
 websocket = [ "websockets~=13.1", "fastapi~=0.115.6" ]
 whisper = [ "faster-whisper~=1.1.1" ]
+mistralai = [ "mistralai (>=1.9.2,<2.0.0)" ]
 
 [tool.setuptools.packages.find]
 where = ["src"]
--- a/src/pipecat/services/mistral/__init__.py
+++ b/src/pipecat/services/mistral/__init__.py
@@ -1,0 +1,5 @@
+#
+# Copyright (c) 2024–2025, Daily
+#
+# SPDX-License-Identifier: BSD 2-Clause License
+#
--- a/src/pipecat/services/mistral/llm.py
+++ b/src/pipecat/services/mistral/llm.py
@@ -1,0 +1,582 @@
+import asyncio
+import json
+from dataclasses import dataclass
+from typing import TYPE_CHECKING, Any, Dict, List, Mapping, Optional
+
+import httpx
+from loguru import logger
+from mistralai import (
+    ConversationUsageInfo,
+    FunctionCallEvent,
+    MessageOutputEvent,
+    ResponseDoneEvent,
+    ResponseStartedEvent,
+)
+from openai import NOT_GIVEN, NotGiven
+from pydantic import BaseModel, Field
+
+from pipecat.adapters.schemas.tools_schema import ToolsSchema
+from pipecat.frames.frames import (
+    Frame,
+    FunctionCallCancelFrame,
+    FunctionCallFromLLM,
+    FunctionCallInProgressFrame,
+    FunctionCallResultFrame,
+    LLMFullResponseEndFrame,
+    LLMFullResponseStartFrame,
+    LLMMessagesFrame,
+    LLMTextFrame,
+    LLMUpdateSettingsFrame,
+)
+from pipecat.metrics.metrics import LLMTokenUsage
+from pipecat.processors.aggregators.llm_response import (
+    LLMAssistantAggregatorParams,
+    LLMAssistantContextAggregator,
+    LLMUserAggregatorParams,
+    LLMUserContextAggregator,
+)
+from pipecat.processors.aggregators.openai_llm_context import (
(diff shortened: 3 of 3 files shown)


markbackman:
I believe Mistral supports the OpenAI client and completions spec. You can make the code much simpler by using the OpenAILLMService as a base class and just overriding where necessary. Here's an example:
https://github.com/pipecat-ai/pipecat/blob/main/src/pipecat/services/groq/llm.py

You could also just use the OpenAILLMService and provide a Mistral base_url instead if you want to avoid writing a new class.

obigroup:
It doesn't seem possible to simplify it the same way as your example. I'm using:

```python
self._mistral_client.beta.conversations.start_stream_async
# or
await self._mistral_client.beta.conversations.start_stream_async
```

as opposed to `await self._client.chat.completions.create(**params)` used in `BaseOpenAILLMService`.

This forces me to rewrite the `async def process_context` function.

## Why I'm Using Mistral's Conversation API

I believe it's more appropriate to use Mistral's conversation function, which allows me to send only the current message instead of always sending the complete context. This provides several benefits:

- **Reduced network data usage**
- **Less processing overhead on Mistral's side** 
- **Faster response times from the LLM**

## Code Quality Considerations

I also think my code has stronger typing to ensure better readability and maintainability.

## Request for Code Review

Would you mind doing a review of my code to point out more specifically which elements I could simplify?

I'd appreciate your feedback on whether the conversation API approach makes sense given the performance benefits, or if there are other ways to achieve similar optimizations while maintaining compatibility with the existing base classes.

Thanks!

obigroup:
@markbackman 

Thanks for the suggestion! I've just simplified the Mistral LLM code to be closer to the OpenAI model following your recommendation.

I've also added a new example `14u-function-calling-mistral.py` that allows testing tool calling functionality.

Everything seems to be working well. Feel free to do a review when you get a chance!

markbackman:
Ok, I had some time to look at this one this morning. I was curious to find the differences between OpenAI and Mistral. They are:
1. Mistral expects `prefix=True` to be appended to `assistant` messages when they are the final message in the context.
2. Mistral processes functions differently, as far as I can tell. OpenAI gets tools from the streamed response, but Mistral appears to read them from the context. When using the OpenAI `run_function_calls` function, this results in tools being called twice.

I've taken a crack at a targeted solution for both of these problems here:
https://github.com/pipecat-ai/pipecat/pull/2408

obigroup:
I've noted the changes you've made regarding the differences between OpenAI and Mistral implementations:

1. The `prefix=True` requirement for assistant messages when they're the final message in context
2. The different function processing behavior that was causing tools to be called twice

However, I'd like to understand why you're not using `self._client.beta.conversations.start_stream_async`? This approach seems more performant than sending the entire context to the API each time.

Thank you.

markbackman:
> I've noted the changes you've made regarding the differences between OpenAI and Mistral implementations:
> 
> 1. The `prefix=True` requirement for assistant messages when they're the final message in context
> 2. The different function processing behavior that was causing tools to be called twice
> 
> However, I'd like to understand why you're not using `self._client.beta.conversations.start_stream_async`? This approach seems more performant than sending the entire context to the API each time.
> 
> Thank you.

My implementation uses the OpenAI client, not the Mistral client. The OpenAI implementation uses the `AsyncOpenAI` client which creates an `AsyncStream` that's returned. So, I believe this is roughly equivalent without adding a new dependency and keeping the code more streamlined and maintainable. (We maintain a lot of integrations, so this is important!)

In testing, the response times are very fast!

obigroup:
I understand the choice for code maintainability. I'll test this with a long conversation that has significant context, as can happen with pipecat-flow.

I'm currently experiencing 1.5 second delays with Anthropic, so I'm hoping this will be faster with Mistral.

Thank you.

obigroup:
I am closing this PR; the implementation of Mistral is planned in this PR: https://github.com/pipecat-ai/pipecat/pull/2408

Issue #2040, opened 2025-06-21
Improve the reliability of end of sentence detect


markbackman:
### Problem Statement

The `match_endofsentence` utility fails to detect specific cases, like:
- Sentences ending in an acronym
- Sentences ending in a number

### Proposed Solution

Instead of using custom logic for end of sentence detection, Pipecat should use a library like NLTK.

### Would you be willing to help implement this feature?

- [ ] Yes, I'd like to contribute
- [ ] No, I'm just suggesting

joshwhiton:
Finding that the sentence boundary detection doesn't account for unclosed parentheses. Wondering if you ever encountered?

match_endofsentence() in pipecat/utils/string.py only checks:
  1. NLTK's sent_tokenize() finds a boundary
  2. Last character is in SENTENCE_ENDING_PUNCTUATION

It doesn't verify that brackets, parentheses, or quotes are balanced before considering a sentence complete. Trailing character after a period gets cut off and pushed out of the pipeline later. 

Affects any streaming LLM output containing parenthetical remarks or quoted text with internal punctuation. Common patterns that break:
  - "(Note: this is important.)"
  - "She said, 'Stop right there.' before leaving"
  - "See the docs [link]."

I encountered on 0.0.86 but can't seen any reason why wouldn't be the same in 0.0.97

markbackman:
@joshwhiton the case you mentioned is now handled as of 0.0.97. We just refactored the aggregation logic here:
https://github.com/pipecat-ai/pipecat/pull/3132

I tested this exact case and it works as expected.

Also, the cases I listed in the comment are handled. Closing this issue.

Issue #1994, opened 2025-06-11
Mixed up transcriptions


tpcgold:
### pipecat version

0.0.71

### Python version

3.13

### Operating System

MacOS, Linux

### Issue description

The call transcriptions get mixed up with 0.0.71 and GeminiMultimodalLiveLLMService
This is a new issue with 0.0.71 it was not there with 0.0.67

### Reproduction steps

use 
```python
llm = GeminiMultimodalLiveLLMService(
            api_key=GOOGLE_API_KEY,
            model="models/gemini-2.5-flash-preview-native-audio-dialog",  # Consider making configurable
            voice_id="Aoede",  # Consider making configurable
            system_instruction=system_instruction_content,
            transcribe_user_audio=True,
            transcribe_model_audio=True,
            params=llm_params,
            tools=tools,
        )
context = OpenAILLMContext(initial_messages)
context.caller_number = (
            caller_number  # Store caller number in context for tools to access
        )
context_aggregator = llm.create_context_aggregator(context)
transcript = TranscriptProcessor()  # Processes transcription frames
pipeline = Pipeline(
            [
                transport.input(),  # Receives frames from Twilio
                #stt_mute_filter,  # Mutes STT during bot speech
                context_aggregator.user(),  # Aggregates user messages into context
                transcript.user(),  # Processes user transcription frames
                llm,  # The core LLM service
                transport.output(),  # Sends frames (TTS audio) to Twilio
                #audiobuffer,  # Buffers incoming audio
                transcript.assistant(),  # Processes assistant transcription frames
                context_aggregator.assistant(),  # Aggregates assistant messages into context
            ]
        )

# Handler for transcription updates
@transcript.event_handler("on_transcript_update")
async def on_transcript_update(processor, frame):
    for msg in frame.messages:
        if isinstance(msg, TranscriptionMessage):
            timestamp = f"[{msg.timestamp}] " if msg.timestamp else ""
            role_prefix = msg.role.capitalize() if msg.role else "Unknown"
            line = f"{timestamp}{role_prefix}: {msg.content}"
            logger.info(f"Transcript: {line}")
            # Append to the list for final save
            transcript_data.append({"role": msg.role, "content": msg.content})
```

### Expected behavior

correct transcription timeline-wise.

this bug can also be observed with the `LLMLogObserver()`
```python
        task = PipelineTask(
            pipeline,
            params=PipelineParams(
                allow_interruptions=True,  # Allow user to interrupt bot speech
[...]

Issue #1692, opened 2025-04-29
TextFrame() issue with All Capital Letters in a word in sentence


sphatate:
### pipecat version

0.0.60

### Python version

3.12

### Operating System

Ubuntu 22.04

### Issue description

When we try to send all capital words in TextFrame(), the tts is not called and bot never speaks
For example 

Working -> TextFrame("Hello, I am Raj. Am I Speaking with Ramesh")
Issue -> TextFrame("Hello, I am Raj. Am I Speaking with RAMESH")

We have tried with google, elevenlabs tts and same issue with both

### Reproduction steps

1. Do this
tts.queue_frame(TextFrame("Hello, I am Raj. Am I Speaking with RAMESH"))

### Expected behavior

TextFrame("Hello, I am Raj. Am I Speaking with RAMESH") Ideally this should also work.

Why having words all capitalize are causing issue

### Actual behavior

TextFrame("Hello, I am Raj. Am I Speaking with RAMESH") 

### Logs

```shell

```

markbackman:
You're right, this is an issue. We are planning to migrate away from the internal `match_endofsentence` util to something more robust. We haven't put the time in yet, but this is something we plan to improve.

markbackman:
There are a few issues like this one. I'm going to close this and instead track in an issue where we'll ultimately implement an improvement to end of sentence detection: #2040.

PR #2235, opened 2025-07-22
Update match_endofsentence to use NLTK sentence tokenizer

markbackman:

The regex based end of sentence detection is too fragile and has too many edge cases. Instead, I'm proposing that we move to using NLTK's punkt_tab sentence tokenizer along with a secondary check for punctuation.

The secondary check is required due to the LLM streaming token use case—NLTK will see an individual word, like `Hello`, and deem it as a complete sentence. Since all LLM output is streamed as tokens, this would result in TTS aggregation breaking. Instead, we're interested in NLTK's sentence detection (and edge case handling for things like acronyms, numbers, emails, etc) along with an added check that the sentence ends in some form of punctuation. Since Pipecat supports a wide range of languages, I asked Claude for some help determining end of sentence punctuation for a broader set of languages. We can expand this as needed.

I also expanded tests to include some edge cases that were failing previously.

Note: `nltk` is now a required dependency with Pipecat.

Should resolve or improve: #2040, #1692, #1994.

### DIFF
--- a/tests/test_utils_string.py
+++ b/tests/test_utils_string.py
@@ -16,10 +16,13 @@ class TestUtilsString(unittest.IsolatedAsyncioTestCase):
         assert match_endofsentence("This is a sentence?") == 19
         assert match_endofsentence("This is a sentence;") == 19
         assert match_endofsentence("This is a sentence...") == 21
-        assert match_endofsentence("This is a sentence . . .") == 24
-        assert match_endofsentence("This is a sentence. ..") == 22
+        assert match_endofsentence("This is a sentence. This is another one") == 19
         assert match_endofsentence("This is for Mr. and Mrs. Jones.") == 31
-        assert match_endofsentence("U.S.A and U.S.A..") == 17
+        assert match_endofsentence("Meet the new Mr. and Mrs.") == 25
+        assert match_endofsentence("U.S.A. and N.A.S.A.") == 19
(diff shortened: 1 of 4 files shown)


tests/test_utils_string.py
```diff
@@ -16,52 +16,177 @@ async def test_endofsentence(self):
         assert match_endofsentence("This is a sentence?") == 19
         assert match_endofsentence("This is a sentence;") == 19
         assert match_endofsentence("This is a sentence...") == 21
-        assert match_endofsentence("This is a sentence . . .") == 24
-        assert match_endofsentence("This is a sentence. ..") == 22
+        assert match_endofsentence("This is a sentence . . .") == 20
```
filipi87: If I am not mistaken, before we were considering this as one sentence:
"This is a sentence . . ."

But now, we are considering that the end of sentence finished at the first "."

Is it expected?

markbackman: This is a weird punctuation situation, which really is more of a stretch to test the logic than it is a case we need to handle. Maybe we should just remove it since it's confusing. WDYT?

filipi87: Yeah, I think it's not something we need to handle, so it's better to remove it. 👍

markbackman: I'll replace with a better, more realistic test.

PR #2239, opened 2025-07-22
Added `set_log_level` to `DailyTransport`

filipi87:
Added `set_log_level` to `DailyTransport`

### DIFF
--- a/src/pipecat/transports/services/daily.py
+++ b/src/pipecat/transports/services/daily.py
@@ -62,6 +62,9 @@ try:
         VirtualCameraDevice,
         VirtualSpeakerDevice,
     )
+    from daily import (
+        LogLevel as DailyLogLevel,
+    )
 except ModuleNotFoundError as e:
     logger.error(f"Exception: {e}")
     logger.error(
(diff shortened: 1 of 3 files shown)


src/pipecat/transports/services/daily.py
```diff
@@ -62,6 +62,9 @@
         VirtualCameraDevice,
         VirtualSpeakerDevice,
     )
+    from daily import LogLevel as DailyLogLevel
```
kompfner: Can't this be part of the previous `from daily import` statement?

src/pipecat/transports/services/daily.py
```diff
@@ -62,6 +62,9 @@
...
+
+    LogLevel = DailyLogLevel
```
kompfner: Indentation?

Also...seems like almost everything else exported from this file is prefixed with `Daily`, so why bother mapping it to an un-prefixed `LogLevel`?

src/pipecat/transports/services/daily.py
filipi87: When I execute `./scripts/fix-ruff.sh` to apply our format, it refactor to its own import.

src/pipecat/transports/services/daily.py
filipi87: I am leaving it as DailyLogLevel as you suggested. Makes sense. 👍

PR #2240, opened 2025-07-22
Adding support for handle_sigterm

filipi87:
Adding support for handle_sigterm


markbackman (approved):
LGTM! Nice solution 🎉

PR #2242, opened 2025-07-22
Upgrade websockets to support asyncio implementation

markbackman:

Our current websockets implementation uses the legacy websockets API. This constrains the versions we support to 13.x and makes it hard for developers to support newer packages that require newer websockets versions. We've had reports of this on and off. This upgrades the websockets API to the asyncio implementation and expands the support to >=13.1.0 (current pinned version) and <15.0.0, as the 15 major version has additional changes that aren't required just yet.

Also, moving to the asyncio implementation, allows us to support versions 13 and 14 without requiring adapters between the two API types / versions.

The main changes with the asyncio implementation are:
- **Import changes**: `websockets.client` has been replaced in favor of `websockets.legacy.client` and `websockets.asyncio.client`. We're moving from legacy to asyncio, so Pipecat now uses the `websockets.asyncio.client` import.
- **Connection state**: The `open` and `closed` properties are removed, requiring use of the `state` property with `State.OPEN`/`State.CLOSED` ([link](https://websockets.readthedocs.io/en/stable/howto/upgrade.html#open-and-closed))
- **Connect method**: `websockets.connect` changes to a new `connect()` method. Since we use connect elsewhere, I'm aliasing as `websocket_connect` to make it clearer what's connecting.
- **Header parameter**: `connect()` replaces `extra_headers` with `additional_headers`, which is just a 1:1 variable swap ([link](https://websockets.readthedocs.io/en/stable/howto/upgrade.html#arguments-of-connect))
- **Server handlers**: The `path` argument of connection handlers is deprecated and removed in the asyncio server implementation

Source:
https://websockets.readthedocs.io/en/stable/howto/upgrade.html

I've tested using 13.1 and 14.2 across the changed examples.


src/pipecat/services/aws/stt.py
```diff
@@ -174,7 +176,7 @@ async def run_stt(self, audio: bytes) -> AsyncGenerator[Frame, None]:
         """
         try:
             # Ensure WebSocket is connected
-            if not self._ws_client or not self._ws_client.open:
+            if not self._ws_client or self._ws_client.state is State.CLOSED:
```
filipi87, line 179: I was looking at all the possible states:
- https://websockets.readthedocs.io/en/stable/reference/sansio/common.html#websockets.protocol.State
CONNECTING = 0
OPEN = 1
CLOSING = 2
CLOSED = 3

So, we are slightly changing the logic here. I don't think this is an issue though.

markbackman: Agreed. We are changing the logic, but the way this was originally written was inconsistent with other checks. Normally, we're checking for:
- Has the websocket client been created OR
- is the socket connection closed

I think this brings the check more inline with the other checks in other services.

filipi87 (approved):
LGTM!

PR #2248, opened 2025-07-23
feat(tts): integrate Async TTS engine into pipecat

ashotbagh:
Added Async TTS engine into pipecat core.


src/pipecat/services/asyncai/__init__.py
```diff
@@ -0,0 +1,13 @@
+#
+# Copyright (c) 2024–2025, Daily
+#
+# SPDX-License-Identifier: BSD 2-Clause License
+#
```
markbackman: You can leave this file blank. We've moved away from this pattern and only need the deprecation module for older services.

pyproject.toml
```diff
@@ -42,6 +42,7 @@ Website = "https://pipecat.ai"
 [project.optional-dependencies]
 anthropic = [ "anthropic~=0.49.0" ]
 assemblyai = [ "websockets~=13.1" ]
+asyncai = [ "websockets~=13.1" ]
```
markbackman: We've recently bumped to the websocket version support and now use the asyncio implementation.

```suggestion
asyncai = [ "websockets>=13.1,<15.0" ]
```

src/pipecat/services/asyncai/tts.py
```diff
@@ -0,0 +1,513 @@
...
+        Args:
+            text: The text to synthesize into speech.
+
+        Yields:
+            Frame: Audio frames containing the synthesized speech.
+        """
+        logger.debug(f"{self}: Generating TTS [{text}]")
+
+        try:
+            if not self._websocket or self._websocket.closed:
```
markbackman: Update for the asyncio implementation:

```suggestion
            if not self._websocket or self._websocket.state is State.CLOSED:
```

src/pipecat/services/asyncai/tts.py
```diff
@@ -0,0 +1,513 @@
...
+                logger.error(f"{self} error, unknown message type: {msg}")
+
+    async def _keepalive_task_handler(self):
+        """Send periodic keepalive messages to maintain WebSocket connection."""
+        KEEPALIVE_SLEEP = 10 if self.task_manager.task_watchdog_enabled else 3
+        while True:
+            self.reset_watchdog()
+            await asyncio.sleep(KEEPALIVE_SLEEP)
+            try:
+                if self._websocket and self._websocket.open:
```
markbackman: Update for the asyncio implementation:

```suggestion
                if self._websocket and self._websocket.state is State.OPEN:
```

src/pipecat/services/asyncai/tts.py
```diff
@@ -0,0 +1,513 @@
...
+        
+        if self._keepalive_task:
+            await self.cancel_task(self._keepalive_task)
+            self._keepalive_task = None
+
+        await self._disconnect_websocket()
+
+    async def _connect_websocket(self):
+        try:
+            if self._websocket and self._websocket.open:
```
markbackman: Update for the asyncio implementation:

```suggestion
            if self._websocket and self._websocket.state is State.OPEN:
```

src/pipecat/services/asyncai/tts.py
```diff
@@ -0,0 +1,513 @@
...
+    TTSStoppedFrame,
+)
+from pipecat.processors.frame_processor import FrameDirection
+from pipecat.services.tts_service import AudioContextWordTTSService, TTSService
+from pipecat.transcriptions.language import Language
+from pipecat.utils.asyncio.watchdog_async_iterator import WatchdogAsyncIterator
+from pipecat.utils.tracing.service_decorators import traced_tts
+
+try:
+    import websockets
```
markbackman, line 36: Add these imports for the websockets changes I'm suggesting. We're importing `connect` as `websocket_connect` to help distinguish from other uses of "connect".
```suggestion
    import websockets
    from websockets.asyncio.client import connect as websocket_connect
    from websockets.protocol import State
```

src/pipecat/services/asyncai/tts.py
```diff
@@ -0,0 +1,513 @@
...
+            self._keepalive_task = None
+
+        await self._disconnect_websocket()
+
+    async def _connect_websocket(self):
+        try:
+            if self._websocket and self._websocket.open:
+                return
+            logger.debug("Connecting to Async")
+            self._websocket = await websockets.connect(
```
markbackman: Update for asyncio implementation:
```suggestion
            self._websocket = await websocket_connect(
```

src/pipecat/services/asyncai/tts.py
```diff
@@ -0,0 +1,513 @@
...
+        # Convert enum value to string and get the base language part (e.g. en-En -> en)
+        lang_str = str(language.value)
+        base_code = lang_str.split("-")[0].lower()
+        # Look up the base code in our supported languages
+        result = base_code if base_code in BASE_LANGUAGES.values() else None
+
+    return result
+
+
+class AsyncAITTSService(AudioContextWordTTSService):
```
markbackman: The `AudioContextWordTTSService ` class should only be used when your service supports the following:
- Websocket API
- Word/timestamp pairs
- Interruption messages (or a context_id for tracking generations)

I don't think this is the case based on the reference docs. This means we need to select a different option. From what I understand in the docs, this is a websocket API with no timestamps and no way to track a context across multiple generations. If so, then you'll need to use the `InterruptibleTTSService` base class. This base class disconnects and reconnects the websocket when interruptions occur. This is less than ideal, but is required as there's no way to correlate the text provide to the audio chunk being returned.

As you add those features, we can change the base class to support more features.

markbackman: Note: the `InterruptibleTTSService` handles the disconnect/connect for you, so no code is required in this class.

src/pipecat/services/asyncai/tts.py
```diff
@@ -0,0 +1,513 @@
...
+
+    def __init__(
+        self,
+        *,
+        api_key: str,
+        voice_id: str,
+        version: str = "v1",
+        url: str = "wss://api.async.ai/text_to_speech/websocket/ws",
+        model: str = "asyncflow_v2.0",
+        sample_rate: int = 32000,
```
markbackman: This should be set to None, so that Pipecat's can pass the value to the class. This works by the user defining the audio_out_sample_rate in the PipelineParams. That sample_rate value is passed to all processors in the pipeline via a StartFrame. You can access the sample rate value via `self.sample_rate`.

```suggestion
        sample_rate: Optional[int] = None,
```

src/pipecat/services/asyncai/tts.py
```diff
@@ -0,0 +1,513 @@
...
+            url: WebSocket URL for Async TTS API.
+            model: TTS model to use (e.g., "asyncflow_v2.0").
+            sample_rate: Audio sample rate.
+            encoding: Audio encoding format.
+            container: Audio container format.
+            params: Additional input parameters for voice customization.
+            aggregate_sentences: Whether to aggregate sentences within the TTSService.
+            **kwargs: Additional arguments passed to the parent service.
+        """
+        # Aggregating sentences still gives cleaner-sounding results and fewer
```
markbackman: You can remove this comment.

src/pipecat/services/asyncai/tts.py
```diff
@@ -0,0 +1,513 @@
...
+        # full sentence should only "cost" us 15ms or so with GPT-4o or a Llama
+        # 3 model, and it's worth it for the better audio quality.
+        #
+        # We also don't want to automatically push LLM response text frames,
+        # because the context aggregators will add them to the LLM context even
+        # if we're interrupted.
+        super().__init__(
+            aggregate_sentences=aggregate_sentences,
+            push_text_frames=False,
+            pause_frame_processing=True,
```
markbackman, line 120: Add:
```suggestion
            pause_frame_processing=True,
            push_stop_frames=True,
```

src/pipecat/services/asyncai/tts.py
```diff
@@ -0,0 +1,513 @@
...
+                "sample_rate": sample_rate,
+            },
+            "language": self.language_to_service_language(params.language)
+            if params.language
+            else "en",
+        },
+        
+        self.set_model_name(model)
+        self.set_voice(voice_id)
+        self._global_context_id = str(uuid.uuid4())
```
markbackman: How is this used? For example, in Cartesia, you can provide a context ID along with text in order to track a "turn". A turn is defined as all of the audio generation requests for a bot's turn. The provided context ID is sent back to Pipecat in messages along with the generated audio. This lets Pipecat track the turn and knows when to stop the turn either when the bot is finished speaking or is interrupted by the user. If this isn't supported, then let's remove this.

src/pipecat/services/asyncai/tts.py
```diff
@@ -0,0 +1,513 @@
...
+
+        self._context_id = None
```
markbackman: Same comment here, as well as all instances of `self._context_id`.

src/pipecat/services/asyncai/tts.py
```diff
@@ -0,0 +1,513 @@
...
+        finally:
+            self._context_id = None
+            self._websocket = None
+
+    def _get_websocket(self):
+        if self._websocket:
+            return self._websocket
+        raise Exception("Websocket not connected")
+
+    async def _handle_interruption(self, frame: StartInterruptionFrame, direction: FrameDirection):
```
markbackman: You won't need this function if the base class changes to `InterruptibleTTSService`.

src/pipecat/services/asyncai/tts.py
```diff
@@ -0,0 +1,513 @@
...
+        async for message in WatchdogAsyncIterator(
+            self._get_websocket(), manager=self.task_manager
+        ):
+            msg = json.loads(message)
+            context_id = self._global_context_id
+            if not msg:
+                continue
+
+            if "final" in msg and msg["final"] is True:
+                await self.stop_ttfb_metrics()
```
markbackman: Remove this. The TTFB (time to first byte) metric is only stopped after receiving the audio message.
```suggestion
```

src/pipecat/services/asyncai/tts.py
```diff
@@ -0,0 +1,513 @@
...
+                await self.remove_audio_context(context_id)
```
markbackman: Remove if you end up not managing an audio context. Only needed if context_id is supported.

src/pipecat/services/asyncai/tts.py
```diff
@@ -0,0 +1,513 @@
...
+            elif msg.get("audio"):
+                await self.stop_ttfb_metrics()
+                frame = TTSAudioRawFrame(
+                    audio=base64.b64decode(msg["audio"]),
+                    sample_rate=self.sample_rate,
+                    num_channels=1,
+                )
+                await self.append_to_audio_context(context_id, frame)
```
markbackman: Remove if you end up not managing an audio context. Only needed if context_id is supported.

src/pipecat/services/asyncai/tts.py
```diff
@@ -0,0 +1,513 @@
...
+
+        try:
+            if not self._websocket or self._websocket.closed:
+                await self._connect()
+
+            if not self._context_id:
+                await self.start_ttfb_metrics()
+                yield TTSStartedFrame()
+                self._context_id = self._global_context_id
+                await self.create_audio_context(self._context_id)
```
markbackman: Remove if you end up not managing an audio context. Only needed if context_id is supported.

src/pipecat/services/asyncai/tts.py
```diff
@@ -0,0 +1,513 @@
...
+                await self._connect()
+                return
+            yield None
+        except Exception as e:
+            logger.error(f"{self} exception: {e}")
+
+class AsyncAIHttpTTSService(TTSService):
+    """HTTP-based Async TTS service.
+
+    Provides text-to-speech using Async's HTTP streaming API for simpler,
```
markbackman, line 357: I see you have a version of the HTTP endpoint that provides word timestamps. That's greatly advantageous as interruptions can be tracked down to the word level, as opposed to assuming the entire bot output was spoken.

How much higher is the latency of the `https://api.async.ai/text_to_speech/with_timestamps` endpoint vs `https://api.async.ai/text_to_speech/streaming`?

src/pipecat/services/asyncai/tts.py
```diff
@@ -0,0 +1,513 @@
...
+        params = params or AsyncAITTSService.InputParams()
+
+        self._api_key = api_key
+        self._api_version = version
+        self._url = url
+        self._settings = {
+            "output_format": {
+                "container": container,
+                "encoding": encoding,
+                "sample_rate": sample_rate,
```
markbackman: We'll initialize elsewhere:
```suggestion
                "sample_rate": 0,
```

src/pipecat/services/asyncai/tts.py
```diff
@@ -0,0 +1,513 @@
...
+        return json.dumps(msg)
+    
+    async def start(self, frame: StartFrame):
+        """Start the Async TTS service.
+
+        Args:
+            frame: The start frame containing initialization parameters.
+        """
+        await super().start(frame)
+        await self._connect()
```
markbackman, line 180: Initialize sample rate. This allows it to get the StartFrame value.
```suggestion
        self._settings["output_format"]["sample_rate"] = self.sample_rate
        await self._connect()
```

src/pipecat/services/asyncai/tts.py
```diff
@@ -0,0 +1,513 @@
...
+    def __init__(
+        self,
+        *,
+        api_key: str,
+        voice_id: str,
+        aiohttp_session: aiohttp.ClientSession,
+        model: str = "asyncflow_v2.0",
+        url: str = "https://api.async.ai",
+        version: str = "v1",
+        sample_rate: int = 32000,
```
markbackman, line 390: Same pattern as above for sample rate.

Also, add a `start` function to set sample rate, like this:
```
        """Start the AsyncAI HTTP TTS service.

        Args:
        """
        await super().start(frame)
        self._settings["output_format"]["sample_rate"] = self.sample_rate
```

markbackman: Same pattern as above for sample rate
```suggestion
        sample_rate: Optional[int] = None,
```

src/pipecat/services/asyncai/tts.py
```diff
@@ -0,0 +1,513 @@
...
+        params = params or AsyncAIHttpTTSService.InputParams()
+
+        self._api_key = api_key
+        self._base_url = url
+        self._api_version = version
+        self._settings = {
+            "output_format": {
+                "container": container,
+                "encoding": encoding,
+                "sample_rate": sample_rate,
```
markbackman: Same as above.

```suggestion
                "sample_rate": 0,
```

src/pipecat/services/asyncai/tts.py
```diff
@@ -0,0 +1,513 @@
...
+        """
+        return language_to_async_language(language)
+
+    async def start(self, frame: StartFrame):
+        """Start the Async HTTP TTS service.
+
+        Args:
+            frame: The start frame containing initialization parameters.
+        """
+        await super().start(frame)
```
markbackman, line 448: Same as above.

```suggestion
        await super().start(frame)
        self._settings["output_format"]["sample_rate"] = self.sample_rate
```

src/pipecat/services/asyncai/tts.py
```diff
@@ -0,0 +1,513 @@
...
+            async with self._session.post(url, json=payload, headers=headers) as response:
+                if response.status != 200:
+                    error_text = await response.text()
+                    logger.error(f"Async API error: {error_text}")
+                    await self.push_error(ErrorFrame(f"Async API error: {error_text}"))
+                    raise Exception(f"Async API returned status {response.status}: {error_text}")
+
+                audio_data = await response.read()
+
+            await self.start_tts_usage_metrics(text)
```
markbackman, line 498: Move this up.
```suggestion
```

src/pipecat/services/asyncai/tts.py
```diff
@@ -0,0 +1,513 @@
...
+            }
+            url = f"{self._base_url}/text_to_speech/streaming"
+
+            async with self._session.post(url, json=payload, headers=headers) as response:
+                if response.status != 200:
+                    error_text = await response.text()
+                    logger.error(f"Async API error: {error_text}")
+                    await self.push_error(ErrorFrame(f"Async API error: {error_text}"))
+                    raise Exception(f"Async API returned status {response.status}: {error_text}")
+
```
markbackman, line 495:
```suggestion
            await self.start_tts_usage_metrics(text)

```

examples/foundational/07aa-interruptible-asyncai-http.py
```diff
@@ -0,0 +1,116 @@
+#
```
markbackman: Rename examples to 07ac.

markbackman:
@ashotbagh Nice! Thanks for contributing. I provided a bunch of feedback, but hopefully it's not too much to work through. I think the biggest thing to sort out is what the base class should be for the websocket API. We need better docs on this for external developers (and we're going to refactor the base classes so the inheritance isn't so complex).

Two additional files needed:
1. Update this page with your optional dep (`pipecat-ai[asyncai]`) for reference docs auto-generation: https://github.com/pipecat-ai/pipecat/blob/main/docs/api/requirements.txt
2. Add a short changelog entry: https://github.com/pipecat-ai/pipecat/blob/main/CHANGELOG.md

You'll need to rebase as well to pick up recent changes to pyproject.toml.

ashotbagh:
@markbackman 
All set! 
- Rebased onto current main so the latest pyproject.toml is included.
- Switched AsyncAITTSService to inherit from InterruptibleStreamingTTSService (matches the pattern you highlighted).
- Added pipecat-ai[asyncai] to docs/api/requirements.txt for auto‑gen docs.
- Dropped a concise entry in CHANGELOG.md under “Added”.

Let me know if anything else jumps out—happy to iterate. Thanks for the thorough review!

src/pipecat/services/asyncai/tts.py
ashotbagh: Great question!

- /text_to_speech/with_timestamps – not streaming. The service synthesizes the entire utterance first and then returns the audio plus word‑level timestamps in one shot. 
- /text_to_speech/streaming -  true streaming.  First audio chunk arrives in ~150–300 ms, then ~50 ms per chunk afterward. No timestamps yet.

For now we’ve added the streaming endpoint because it meets the low‑latency requirement.  
Word‑level timestamps will be added to the streaming API; once that completed we’ll update `AsyncAITTSService` so we can interrupt at word boundaries.

src/pipecat/services/asyncai/__init__.py
```diff
@@ -1,9 +1,3 @@
-#
-# Copyright (c) 2024–2025, Daily
-#
-# SPDX-License-Identifier: BSD 2-Clause License
-#
-
 import sys
```
markbackman, line 1: The entire `__init__.py` file should be blank.

src/pipecat/services/asyncai/__init__.py
```diff
@@ -0,0 +1,7 @@
+import sys
```
markbackman: The entire __init__.py file should be blank.

src/pipecat/services/asyncai/tts.py
```diff
@@ -0,0 +1,507 @@
...
+# Copyright (c) 2024–2025, Daily
+#
+# SPDX-License-Identifier: BSD 2-Clause License
+#
+
+"""Async text-to-speech service implementations."""
+
+import base64
+import json
+import uuid
```
markbackman: Remove unused import:
```suggestion
```

src/pipecat/services/asyncai/tts.py
```diff
@@ -0,0 +1,507 @@
...
+        container: str = "raw",
+        params: Optional[InputParams] = None,
+        aggregate_sentences: Optional[bool] = True,
+        **kwargs,
+    ):
+        """Initialize the Async TTS service.
+
+        Args:
+            api_key: Async API key.
+            voice_id: ID of the voice to use for synthesis.
```
markbackman: nit: be more specific and link to docs.
```suggestion
            voice_id: UUID of the voice to use for synthesis. See docs for a full list:
                https://docs.async.ai/list-voices-16699698e0
```

examples/foundational/07ac-interruptible-asyncai.py
```diff
@@ -0,0 +1,111 @@
...
+    ),
+}
+
+
+async def run_example(transport: BaseTransport, _: argparse.Namespace, handle_sigint: bool):
+    logger.info(f"Starting bot")
+
+    stt = OpenAISTTService(api_key=os.getenv("OPENAI_API_KEY"))
+
+    tts = AsyncAITTSService(
```
markbackman, line 55: Maybe provide a fallback voice_ID in case the env variable isn't listed?

src/pipecat/services/asyncai/tts.py
```diff
@@ -0,0 +1,507 @@
...
+        container: str = "raw",
+        params: Optional[InputParams] = None,
+        **kwargs,
+    ):
+        """Initialize the Async TTS service.
+
+        Args:
+            api_key: Async API key.
+            voice_id: ID of the voice to use for synthesis.
+            model: TTS model to use (e.g., "asyncflow_v2.0").
```
markbackman, line 392: Missing a docstring for aiohttp_session:

```suggestion
            aiohttp_session: An aiohttp session for making HTTP requests.
            model: TTS model to use (e.g., "asyncflow_v2.0").
```

examples/foundational/07ac-interruptible-asyncai-http.py
```diff
@@ -0,0 +1,116 @@
...
+    ),
+}
+
+
+async def run_example(transport: BaseTransport, _: argparse.Namespace, handle_sigint: bool):
+    logger.info(f"Starting bot")
+
+    # Create an HTTP session
+    async with aiohttp.ClientSession() as session:
+        stt = OpenAISTTService(api_key=os.getenv("OPENAI_API_KEY"))
```
markbackman: I highly recommend switching to Deepgram, AssemblyAI, Speechmatics, or Gladia. OpenAI STT is very slow.
```
```

Deepgram is used widely, so developers likely have a key.

examples/foundational/07ac-interruptible-asyncai.py
```diff
@@ -0,0 +1,111 @@
...
+        audio_out_enabled=True,
+        vad_analyzer=SileroVADAnalyzer(),
+    ),
+}
+
+
+async def run_example(transport: BaseTransport, _: argparse.Namespace, handle_sigint: bool):
+    logger.info(f"Starting bot")
+
+    stt = OpenAISTTService(api_key=os.getenv("OPENAI_API_KEY"))
```
markbackman: Same comment here:

```suggestion
    stt = DeepgramSTTService(api_key=os.getenv("DEEPGRAM_API_KEY"))
```

CHANGELOG.md
```diff
@@ -9,6 +9,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- **async.ai TTS integration** (https://async.ai/)
```
markbackman: Nit on formatting:
```
- Added async.ai TTS integration (https://async.ai/)
  - `AsyncAITTSService` – WebSocket-based streaming TTS with interruption support
  - `AsyncAIHttpTTSService` – HTTP-based streaming TTS service
  - Example scripts:
    - `examples/foundational/07ac-interruptible-asyncai.py` (WebSocket demo)
    - `examples/foundational/07ac-interruptible-asyncai-http.py` (HTTP demo)
```

markbackman:
Nice! Just a few additional, minor comments.

Please install the pre-commit hook:
```
pre-commit install
```

or run the [linting scripts](https://github.com/pipecat-ai/pipecat/tree/main/scripts) to fix formatting issues.

In testing the websocket service I noticed times with very high latency, 10-20 seconds to start streaming audio. Usually the first output was reliable, but sometimes even the first output came with a lot of latency.

Also, usually the first voice generation was "stable". Subsequent generations had significant variability in the output. I was testing with `August`, uuid: `000fc99f-437d-47c5-8d1e-0ba5b13ce94d`.

ashotbagh:
Pre‑commit hooks are installed and formatting is fixed. 
The long‑latency cases look like rare cold‑starts—could you retry and let me know if they persist?

markbackman:
> The long‑latency cases look like rare cold‑starts—could you retry and let me know if they persist?
@ashotbagh I can test again once this is merged.

Sorry, you'll have to rebase again due to a conflict in the Changelog. Apologies!

UPDATE: I was able to resolve this directly via Github.

markbackman (approved):
LGTM. Thanks for working through this with me!

PR #2254, opened 2025-07-23
ElevenLabsTTSService: Only reset the context_id when interruptions ar…

markbackman:
…e enabled

In the ElevenLabsTTSService code, the prior assumption was that interruptions were enabled. This means that the `_handle_interruption` function would reset the context ID and `run_tts` would create a new ID.

Now, ElevenLabsTTSService only creates a new context ID in `run_tts` if no current ID exists. The avoids the concurrency limit imposed by the ElevenLabs websocket API by reusing the same context ID throughout the conversation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -45,6 +45,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   `pyneuphonic` package. This removes a package requirement, allowing Neuphonic
   to work with more services.
 
+- Updated `ElevenLabsTTSService` to handle the case where
+  `allow_interruptions=False`. Now, when interruptions are disabled, the same
+  context ID will be used throughout the conversation.
+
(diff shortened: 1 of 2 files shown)


markbackman:
@filipi87 the team we're working with on this one got back finding another edge case related to `allow_interruptions=False`. I pushed one more commit where we check to see if there's an audio context being tracked already. If not, then create one. I think this should be the last change. Sorry for the re-review.

filipi87 (approved):
Makes sense. 🚀
Do we need to implement the same approach in the other TTS services ?

markbackman:
> Makes sense. 🚀 Do we need to implement the same approach in the other TTS services ?

I'm not sure. This is needed only for `AudioContextWordTTSService` TTS classes, which includes CartesiaTTSService and RimeTTSService. I'm going to merge this for ElevenLabs now because it's blocking someone. We can revisit Cartesia and Rime soon.

PR #2255, opened 2025-07-24
pyproject.toml dependency updates to support better cross compatibility

markbackman:

This PR resolves dependency conflicts that were preventing `uv sync` from working and makes pipecat-ai more compatible with the broader Python ecosystem. One additional change is needed to achieve this goal: https://github.com/pipecat-ai/pipecat/pull/2256.

This will also make it easier to run the release evals, as all packages can be installed simultaneously.

## Changes:

- **numba**: Reducing the minimum numba dependency to 0.60.0. This still satisfies the requirement for python 3.10 and above, while also working compatibly with the `llvm` version required by the `ultravox` optional dependency.
- **aws_sdk_bedrock_runtime**: The `aws_sdk_bedrock_runtime` only supports python 3.12 and newer, so formalizing that constraint.
- **torch** and **torchaudio**: Relaxing from exact pinning (`==2.5.0`) to compatible range (`~=2.5.0`) to allow patch updates. This improves compatibility with other ML packages while maintaining API stability.
- **pyht**: Relaxing version constraint from `~=0.1.12` to `>=0.1.6` to resolve grpcio conflicts with nvidia-riva-client. Users can install both services and pin specific versions in their applications if needed.
- **fastapi**: Adding support for fastapi>=0.115.6,<0.117.0 to allow for use of newer FastAPI features.

For other core dependencies (`aiohttp`, `Markdown`, `nltk`, `numpy`, `Pillow`, `pydantic`, `openai`):
- Adding support up from the pinned version up to the next major version. These are common libraries that are stable. Overly constraining them makes Pipecat difficult to work with.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,20 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- Dependency compatibility improvements: Relaxed version constraints for core
+  dependencies to support broader version ranges while maintaining stability:
+
+  - `aiohttp`, `Markdown`, `nltk`, `numpy`, `Pillow`, `pydantic`, `openai`,
+    `numba`: Now support up to the next major version (e.g. `numpy>=1.26.4,<3`)
+  - `pyht`: Relaxed to `>=0.1.6` to resolve `grpcio` conflicts with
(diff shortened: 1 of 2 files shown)


filipi87 (approved):
Nice job. 🔥🚀

PR #2259, opened 2025-07-24
Play delayed messages from `ElevenLabsTTSService` if they still belong to the current context.

filipi87:
Play delayed messages from `ElevenLabsTTSService` if they still belong to the current context.

### DIFF
--- a/src/pipecat/services/elevenlabs/tts.py
+++ b/src/pipecat/services/elevenlabs/tts.py
@@ -548,8 +548,12 @@ class ElevenLabsTTSService(AudioContextWordTTSService):
             # Check if this message belongs to the current context.
             # This should never happen, so warn about it.
             if not self.audio_context_available(received_ctx_id):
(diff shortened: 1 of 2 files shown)


src/pipecat/services/elevenlabs/tts.py
```diff
@@ -548,8 +548,14 @@ async def _receive_messages(self):
             # Check if this message belongs to the current context.
             # This should never happen, so warn about it.
             if not self.audio_context_available(received_ctx_id):
-                logger.warning(f"Ignoring message from unavailable context: {received_ctx_id}")
-                continue
+                if self._context_id == received_ctx_id:
+                    logger.debug(
+                        f"Received a delayed message,  recreating the context: {self._context_id}"
```
markbackman: Remove extra space.

```suggestion
                        f"Received a delayed message, recreating the context: {self._context_id}"
```

PR #2260, opened 2025-07-24
Fixed an issue in `AudioBufferProcessor` that caused garbled audio

filipi87:
Fixed an issue in `AudioBufferProcessor` that caused garbled audio when `enable_turn_audio` was enabled and audio resampling was required.


filipi87:
The issue was that the SOXRStreamAudioResampler maintains an internal history, so it should not be reused across different contexts. However, this is exactly what was happening inside the AudioBufferProcessor when enable_turn_audio was enabled, we were injecting the same chunk of audio twice into the same resampler.

PR #2266, opened 2025-07-25
HeyGen implementation for Pipecat - HeyGenTransport

filipi87:
HeyGen implementation for Pipecat.
- `HeyGenTransport`: where Pipecat, the user, and HeyGen would all join the same room.


filipi87:
Regarding `43-heygen-transport.py`, as soon as the user joins, Pipecat gets disconnected from LiveKit.

We are waiting for HeyGen, they acknowledge that they may need to issue two tokens, one for the browser client and another for the voice agent, or see if they can allow more than one participant to use the same token.

We will need to wait until they have a solution for this. But once they have, It should be fairly straightforward to implement the changes.

src/pipecat/transports/heygen/transport.py
```diff
@@ -0,0 +1,385 @@
+#
```
aconchillo, line 1: We need to move this file to it's own directory now: `pipecat.transports.heygen.transport`

aconchillo:
@filipi87 Is this still valid?

filipi87:
> @filipi87 Is this still valid?

It is, but we are waiting for HeyGen to provide a way for us to retrieve more than one access token. They mentioned a couple of weeks ago that they are working on this.

filipi87:
This PR is now ready for review. 🎉🙌

src/pipecat/transports/heygen/transport.py
```diff
@@ -0,0 +1,381 @@
...
+            direction: The direction of frame flow in the pipeline
+
+        Note:
+            Special handling is implemented for:
+            - StartInterruptionFrame: Triggers interruption of current speech
+            - UserStartedSpeakingFrame: Initiates agent listening mode
+            - UserStoppedSpeakingFrame: Stops agent listening mode
+        """
+        await super().process_frame(frame, direction)
+        if isinstance(frame, StartInterruptionFrame):
```
kompfner: Update to `InterruptionFrame`

filipi87: Nice catch. Going to do that.

PR #2269, opened 2025-07-26
New runner module

markbackman:
## UPDATE

In brainstorming with Kwin, we're opting to consolidate to a single runner—pipecat.runner.run. This will be focused on the PCC style bot() method in the file. This will both simplify the runner code and make all code easily deployable.

With that, I've:
- removed: `pipecat.runner.local`
- renamed: `pipecat.runner.cloud` to `pipecat.runner.run`
- simplified examples since there's no only one runner version

Also, I'll remove the pipecat.examples.run module and update all examples in a separate PR.

Details updated below, but forgive reference to two runners (cloud, local):

---

The PR is trying to accomplish two things:

1. Improve upon the pipecat.examples.runner such that any transport can be run. Additionally, lazy load the imports, so you don't need to specify all of the imports to run only one transport.
2. Create a runner for Pipecat Cloud that allows you to use the same file both locally and when cloud deployed. More on this in a moment.

In addition, this PR includes what was going to be a quickstart, but is more of an example for the runners. Before this is merged, This includes example files that will be cleaned up before merging.

### Runner details

**`pipecat.runner.run`**
This is used for files that will be deployed to Pipecat Cloud. The key benefit is that **the same bot code works both locally and when deployed** - no changes required. When deployed, Pipecat Cloud provides the runner functionality. When developing locally, this module provides equivalent functionality.

Command-line arguments:
```
--host: Server host address (default: localhost)
--port: Server port (default: 7860)
-t/--transport: Transport type (daily, webrtc, twilio, telnyx, plivo)
-x/--proxy: Public proxy hostname for telephony webhooks
--esp32: Enable SDP munging for ESP32 compatibility (requires --host with IP address)
-d/--direct: Connect directly to Daily room (automatically sets transport to daily)
-v/--verbose: Increase logging verbosity
```

Usage examples:
- SmallWebRTCTransport: `python bot.py` (default)
- Daily: `python bot.py --transport daily`
- Daily (direct connection): `python bot.py --direct
- Twilio, Telnyx, Plivo: `python bot.py --transport twilio --proxy username.ngrok.io`

Code example:
```
async def bot(session_args: DailySessionArguments | SmallWebRTCSessionArguments):

    # session_args types are used to determine which transport to run
    if isinstance(session_args, DailySessionArguments):
        from pipecat.transports.services.daily import DailyParams, DailyTransport

        transport = DailyTransport(
            session_args.room_url,
            session_args.token,
            "Pipecat Bot",
            params=DailyParams(
                audio_in_enabled=True,
                audio_out_enabled=True,
                vad_analyzer=SileroVADAnalyzer(),
            ),
        )

    elif isinstance(session_args, SmallWebRTCSessionArguments):
        from pipecat.transports.network.small_webrtc import SmallWebRTCTransport

        transport = SmallWebRTCTransport(
            params=TransportParams(
                audio_in_enabled=True,
                audio_out_enabled=True,
                vad_analyzer=SileroVADAnalyzer(),
            ),
            webrtc_connection=session_args.webrtc_connection,
        )

    # Pass the transport and run the bot
    await run_bot(transport)
```

**No code changes are required to run locally or deployed.** This works by the `pipecat.runner.cloud` module injecting `session_args` into the `bot()` method. This mimics the same behavior that the Pipecat Cloud bot starter uses.

> Note: In case you need to do anything conditional in the bot.py file, the `pipecat.runner.cloud` sets an env var called `LOCAL_RUN` to `1`. You can use that to conditionally do things in code. For example, if your DailyTransport, you could opt to include Krisp only when deployed to PCC.

### Other details

- Both runners utilize FastAPI endpoints, where the clients connect to a given endpoint based on the transport type. This is the runner logic, which is encapsulated in the module. This approach hides how the runner works, but it is open source so anyone can read it if they need to learn how it works.

- These changes move away from the subprocessor pattern, which is something we're generally endorsing.

### Examples:

Lastly, of the examples provided, I _think_ the `local-simple-bot.py` is a candidate quickstart example.

The other examples were included to help me test the different permutations of the runner scripts. I need to refactor or remove them. Other than `local-simple-bot` (which will be renamed), the other examples won't reside in Quickstart.

Related, many Pipecat examples use the prior local runner from `pipecat.examples.runner`. The `pipecat.runner.local` is essentially the same module, just with extended options. I'm removing the old runner and using the new runner everywhere. I've broken out that PR here: https://github.com/pipecat-ai/pipecat/pull/2275.


markbackman:
Adding a new dataclass to the `pipecatcloud` client in support of this effort. With that, we'll be able to use imports from one module to check types. This will make the cloud runner API easier to use.
https://github.com/daily-co/pipecat-cloud/pull/62

Until then, the type is locally defined in cloud.py.

examples/runner-examples/01-all-transport-bot.py
```diff
@@ -0,0 +1,211 @@
...
+    async def on_client_disconnected(transport, client):
+        logger.info("Client disconnected")
+        await task.cancel()
+
+    runner = PipelineRunner(handle_sigint=False)
+    await runner.run(task)
+
+
+async def bot(
+    session_args: DailySessionArguments | SmallWebRTCSessionArguments | WebSocketSessionArguments,
```
aconchillo: Does this mean we are tying all the examples to Pipecat Cloud?

pyproject.toml
```diff
@@ -81,10 +81,12 @@ openai = [ "websockets>=13.1,<15.0" ]
 openpipe = [ "openpipe~=4.50.0" ]
 openrouter = []
 perplexity = []
+pipecatcloud = [ "pipecatcloud>=0.2.0" ]
```
aconchillo: why would we need this?

src/pipecat/runner/run.py
```diff
@@ -0,0 +1,495 @@
...
+    from fastapi.responses import HTMLResponse, RedirectResponse
+except ImportError as e:
+    logger.error(f"Runner dependencies not available: {e}")
+    logger.error("To use Pipecat runners, install with: pip install pipecat-ai[runner]")
+    raise ImportError(
+        "Runner dependencies required. Install with: pip install pipecat-ai[runner]"
+    ) from e
+
+try:
+    from pipecatcloud.agent import DailySessionArguments, WebSocketSessionArguments
```
aconchillo: I'm not sure this is a good idea. We are now tying `pipecat-ai` to `pipecatcloud` which is a Daily project. I think it would be better to create `DailyRunnerArguments`, `WebSocketRunnerArguments`, `SmallWebRTCRunnerArguments`, etc. in `pipecat-ai`. Then, in pipecat-cloud we could inherit from that:

```
class DailySessionArgument(DailyRunnerArguments):
    session_id: str
```

or a breaking change:


```
class DailySessionArguments:
    runner: DailyRunnerArguments
    session_id: str
```

something like that.

aconchillo: Actually, it needs to be the former so it's compatible with the current deployment. So, no breaking change.

examples/runner-examples/01-all-transport-bot.py
```diff
@@ -0,0 +1,207 @@
...
+    async def on_client_disconnected(transport, client):
+        logger.info("Client disconnected")
+        await task.cancel()
+
+    runner = PipelineRunner(handle_sigint=False)
+    await runner.run(task)
+
+
+async def bot(
+    session_args: DailyRunnerArguments | SmallWebRTCRunnerArguments | WebSocketRunnerArguments,
```
aconchillo: maybe call then `runner_args`

examples/runner-examples/01-all-transport-factory-bot.py
```diff
@@ -0,0 +1,150 @@
...
+    async def on_client_disconnected(transport, client):
+        logger.info("Client disconnected")
+        await task.cancel()
+
+    runner = PipelineRunner(handle_sigint=False)
+    await runner.run(task)
+
+
+async def bot(
+    session_args: DailyRunnerArguments | SmallWebRTCRunnerArguments | WebSocketRunnerArguments,
```
aconchillo: same here

examples/runner-examples/02-two-transport-bot.py
```diff
@@ -0,0 +1,141 @@
...
+    @transport.event_handler("on_client_disconnected")
+    async def on_client_disconnected(transport, client):
+        logger.info("Client disconnected")
+        await task.cancel()
+
+    runner = PipelineRunner(handle_sigint=False)
+    await runner.run(task)
+
+
+async def bot(session_args: DailyRunnerArguments | SmallWebRTCRunnerArguments):
```
aconchillo: and here.

examples/runner-examples/03-single-transport-bot.py
```diff
@@ -0,0 +1,112 @@
...
+    @transport.event_handler("on_client_disconnected")
+    async def on_client_disconnected(transport, client):
+        logger.info("Client disconnected")
+        await task.cancel()
+
+    runner = PipelineRunner(handle_sigint=False)
+    await runner.run(task)
+
+
+async def bot(session_args: SmallWebRTCRunnerArguments):
```
aconchillo: and here.

examples/runner-examples/01-all-transport-bot.py
aconchillo: never mind.

examples/runner-examples/01-all-transport-factory-bot.py
aconchillo: I think this could just be `RunnerArguments` to simplify. We already have the `isinstance` checks.

examples/runner-examples/02-two-transport-bot.py
aconchillo: use `RunnerArguments`

examples/runner-examples/02-two-transport-bot.py
```diff
@@ -0,0 +1,141 @@
...
+        from pipecat.transports.network.small_webrtc import SmallWebRTCTransport
+
+        transport = SmallWebRTCTransport(
+            params=TransportParams(
+                audio_in_enabled=True,
+                audio_out_enabled=True,
+                vad_analyzer=SileroVADAnalyzer(),
+            ),
+            webrtc_connection=session_args.webrtc_connection,
+        )
```
aconchillo, line 142: Maybe a warning here.

examples/runner-examples/02-two-transport-bot.py
```diff
@@ -0,0 +1,141 @@
...
+
+    await run_bot(transport)
```
aconchillo, line 151: initialize `transport = None` at the beginning and check `if transport:...`

examples/runner-examples/03-single-transport-bot.py
aconchillo: use `RunnerArguments`

src/pipecat/runner/run.py
```diff
@@ -0,0 +1,462 @@
...
+        # Your bot logic here
+        await run_pipeline(transport)
+
+    if __name__ == "__main__":
+        from pipecat.runner.run import main
+        main()
+
+Multiple transport example::
+
+    async def bot(session_args):
```
aconchillo: you could use

```
    async def bot(session_args: RunnerArguments):
```

src/pipecat/runner/run.py
```diff
@@ -0,0 +1,462 @@
...
+        # Type-safe transport detection
+        if isinstance(session_args, DailyRunnerArguments):
+            transport = setup_daily_transport(session_args)  # Your application code
+        elif isinstance(session_args, SmallWebRTCRunnerArguments):
+            transport = setup_webrtc_transport(session_args)  # Your application code
+        elif isinstance(session_args, WebSocketRunnerArguments):
+            transport = setup_telephony_transport(session_args)  # Your application code
+
+        # Your bot implementation
+        await run_pipeline(transport)
```
aconchillo, line 50: maybe add check `if transport: ...` (which requires a transport = None.

examples/runner-examples/01-all-transport-bot.py
```diff
@@ -0,0 +1,207 @@
...
+from pipecat.services.deepgram.stt import DeepgramSTTService
+from pipecat.services.openai.llm import OpenAILLMService
+
+load_dotenv(override=True)
+
+# Check if we're running locally
+IS_LOCAL_RUN = os.environ.get("LOCAL_RUN", "0") == "1"
+
+
+async def run_bot(transport):
```
aconchillo: `transport: BaseTransport`

examples/runner-examples/01-all-transport-bot.py
```diff
@@ -0,0 +1,207 @@
...
+    @transport.event_handler("on_client_disconnected")
+    async def on_client_disconnected(transport, client):
+        logger.info("Client disconnected")
+        await task.cancel()
+
+    runner = PipelineRunner(handle_sigint=False)
+    await runner.run(task)
+
+
+async def bot(
```
aconchillo: `RunnerArguments` to simplify

examples/runner-examples/01-all-transport-bot.py
```diff
@@ -0,0 +1,207 @@
...
+            params=FastAPIWebsocketParams(
+                audio_in_enabled=True,
+                audio_out_enabled=True,
+                add_wav_header=False,
+                vad_analyzer=SileroVADAnalyzer(),
+                serializer=serializer,
+            ),
+        )
+
+    await run_bot(transport)
```
aconchillo, line 212: check `if transport`

examples/runner-examples/01-all-transport-factory-bot.py
```diff
@@ -0,0 +1,150 @@
...
+    "plivo": lambda: FastAPIWebsocketParams(
+        audio_in_enabled=True,
+        audio_out_enabled=True,
+        vad_analyzer=SileroVADAnalyzer(),
+        # add_wav_header and serializer will be set automatically
+    ),
+}
+
+
+async def run_bot(transport):
```
aconchillo: `transport: BaseTransport`

examples/runner-examples/02-two-transport-bot.py
```diff
@@ -0,0 +1,141 @@
...
+from pipecat.services.openai.llm import OpenAILLMService
+
+load_dotenv(override=True)
+
+
+# Check if we're running locally
+IS_LOCAL_RUN = os.environ.get("LOCAL_RUN", "0") == "1"
+
+
+async def run_bot(transport):
```
aconchillo: `transport: BaseTransport`

examples/runner-examples/03-single-transport-bot.py
```diff
@@ -0,0 +1,112 @@
...
+from pipecat.services.cartesia.tts import CartesiaTTSService
+from pipecat.services.deepgram.stt import DeepgramSTTService
+from pipecat.services.openai.llm import OpenAILLMService
+from pipecat.transports.base_transport import TransportParams
+from pipecat.transports.network.small_webrtc import SmallWebRTCTransport
+
+load_dotenv(override=True)
+
+
+async def run_bot(transport):
```
aconchillo: `transport: BaseTransport`

aconchillo (approved):
LGTM! Nice work! Just added some minor typing comments.

examples/runner-examples/01-all-transport-bot.py
markbackman: Moved types into the runner. No longer requiring a PCC dependency.

examples/runner-examples/02-two-transport-bot.py
markbackman: 👍

PR #2283, opened 2025-07-28
Add quickstart example

markbackman:


markbackman:
> Hey @markbackman, I think this looks great.
> 
> Do you also think we should do something about RN, iOS, and Android SDKs ?
> 
> Perhaps create specific quickstarts for each platform ? Or maybe transform the `client-server-web` example into just `client-server` and include all possible clients there ?
> 
> Anyway, these would be to a follow up PR.

My idea is to refer to the simple-chatbot as the next step. So, the learning path for someone building with a client SDK would be:
1. quickstart
2. client-server-web
3. simple-chatbot

There will be guides for each and as next steps, it will suggest the next thing to do. Do you think that works? My goal is to keep the quickstart examples simple and focused. Then, rely on other demos to assist with the learning journey.

filipi87:
> There will be guides for each and as next steps, it will suggest the next thing to do. Do you think that works? My goal is to keep the quickstart examples simple and focused. Then, rely on other demos to assist with the learning journey.

Yep, I believe this should work fine as well. 🙂

markbackman:
Ugh, forgot to merge this into the runner branch before it was merged. Merge conflicts are nasty, so close and will reopen a new branch.

PR #2286, opened 2025-07-28
fix: Improve TranscriptProcessor detection for transcript type

markbackman:

The `isspace()` function checks for any whitespace character, not just spaces. When an LLM outputs text that contains newline characters (`\n`), this is incorrectly detected as a space, resulting in an incorrect transcript aggregation strategy.

Instead, the check is now explicitly for spaces.

Note: if newlines were output by the LLM, the TranscriptProcessor should not clean them up. The job of the TranscriptProcessor is to output the text received without modification. The `text_filters` feature of the TTSService can be used to modify the text. For example, setting `text_filters=[MarkdownTextFilter()]` will strip newline characters.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -74,6 +74,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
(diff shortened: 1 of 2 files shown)


aconchillo:
LGTM

aconchillo (approved):

PR #2288, opened 2025-07-29
Update placement of STTMuteFilter in examples to reflect the new reco…

markbackman:
…mmendation

We found a race condition where a Transcript can make it down the pipeline. Moving the STTMuteFilter to later in the Pipeline resolves this issue.


examples/foundational/24-stt-mute-filter.py
```diff
@@ -111,8 +111,8 @@ async def run_example(transport: BaseTransport, _: argparse.Namespace, handle_si
     pipeline = Pipeline(
         [
             transport.input(),  # Transport user input
-            stt_mute_processor,  # Add the mute processor before STT
             stt,  # STT
+            stt_mute_processor,  # Add the mute processor before STT
```
kompfner: Comment no longer right

kompfner:
Wait...how does muting work, with `STTMuteFilter` downstream from the STT service? Looking at the code it seems like the `STTMuteFilter` works by pushing `STTMuteFrame`s downstream, which tell any STT services whether or not to mute...

markbackman:
> Wait...how does muting work, with `STTMuteFilter` downstream from the STT service? Looking at the code it seems like the `STTMuteFilter` works by pushing `STTMuteFrame`s downstream, which tell any STT services whether or not to mute...

You're right! We need to push the `STTMuteFrame` upstream and downstream. Adding that here.

Though, the STTMuteFilter already works—the mechanism for that is blocking the TranscriptionFrame, which the processor does natively. The `STTMuteFrame` is what stops the STT service from producing transcripts while it is "muted". Nonetheless, the change you're suggesting is absolutely right. I've pushed a commit with that change.

examples/foundational/24-stt-mute-filter.py
markbackman: Good catch. I could have sworn I fixed this. I suppose it was in the docs.. Anyway, fixed here.

aconchillo (approved):

aconchillo:
LGTM

PR #2293, opened 2025-07-30
Fix Piper TTS Service

azain47:
Update Piper TTS Service to work with the newer [Piper GPL](https://github.com/OHF-Voice/piper1-gpl) Version, that uses JSON as its payload.

# Issue
Pipecat returns an error when running the PiperTTSService:
```html
pipecat.services.piper.tts:run_tts:90 - PiperTTSService#0 error getting audio (status: 500, error: <!doctype html>
<html lang=en>
<title>500 Internal Server Error</title>
<h1>Internal Server Error</h1>
<p>The server encountered an internal error and was unable to complete your request. Either the server is overloaded or there is an error in the application.</p>
)
```

When looking at Piper TTS HTTP Server Debug messages:
```
json.decoder.JSONDecodeError: Expecting value: line 1 column 1 (char 0)
INFO:werkzeug:127.0.0.1 - - POST / HTTP/1.1" 500 -
```

# Resolution
Send JSON data by changing the Content-Type and sending `json=text` in `session.post()`

# Additional Note
In the [Usage Example Docs](https://docs.pipecat.ai/server/services/tts/piper#usage-example) , the example uses `http://localhost:5000/api/tts` as its base URL, which has been changed to just `/` in the newer piper version. 
[Source](https://github.com/OHF-Voice/piper1-gpl/blob/0e29e407ade1dd39ef6d25745bbcbf186f497a9d/src/piper/http_server.py#L173)
Please update the documentation as well with the same.

### DIFF
--- a/src/pipecat/services/piper/tts.py
+++ b/src/pipecat/services/piper/tts.py
@@ -22,7 +22,7 @@ from pipecat.services.tts_service import TTSService
 from pipecat.utils.tracing.service_decorators import traced_tts
 
 
-# This assumes a running TTS service running: https://github.com/rhasspy/piper/blob/master/src/python_run/README_http.md
+# This assumes a running TTS service running: https://github.com/OHF-Voice/piper1-gpl/blob/main/docs/API_HTTP.md
 class PiperTTSService(TTSService):
     """Piper TTS service implementation.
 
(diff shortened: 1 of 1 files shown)


aconchillo (approved):

aconchillo:
> I'm not familiar with Piper, but have a question:
> 
> Does this break compatibility with older versions of Piper?

I believe it would but I'd say everyone would just update to the new repo.

ezhilkumaran:
There is one minor change needed for the PiperTTSService to work. Update the JSON data object format by changing the below line in piper -> tts.py,
  Remove -          async with self._session.post(self._base_url, json=text, headers=headers) as response:   
  Add -        async with self._session.post(self._base_url, json={"text": text}, headers=headers) as response:

{ "text": "This is a test." } is the correct format as per Piper docs. 

Would be happy to create a PR.

markbackman:
This was merged today:
https://github.com/pipecat-ai/pipecat/pull/2332

ezhilkumaran:
That's great! Thanks!

PR #2299, opened 2025-07-30
Fix CJK spacing issue with add_timestamps in Cartesia TTS

golbin:
When using the `add_timestamps` option in Cartesia TTS, there is a problem specific to CJK (Chinese, Japanese, Korean) languages.

For example, when generating audio for the phrase `"안녕 반가워"`, enabling `add_timestamps` results in output being split by character, not by word like `"안"`, `"녕"`, `"반"`, `"가"`, `"워"`. In this case, the space between `"녕"` and `"반"` is not preserved, leading to incorrect segmentation.

To work around this, we’ve added support for disabling the `add_timestamps` option. When `add_timestamps=False`, the raw input text will be passed directly to the next stage in the pipeline.

This issue is critical for handling CJK conversations properly, and we hope it can be merged quickly.

Additionally, it appears that the `expect_stripped_words=True` default value in `LLMAssistantAggregatorParams` was set to accommodate this behavior (though we’re not completely certain). However, when TTS fails to handle spacing accurately, this setting may introduce incorrect spacing—this could affect English as well.

Therefore, when disabling `add_timestamps`, we recommend also setting `expect_stripped_words=False` in `LLMAssistantAggregatorParams`:

```python
assistant_params=LLMAssistantAggregatorParams(
    expect_stripped_words=False,
)
```

### DIFF
--- a/src/pipecat/services/cartesia/tts.py
+++ b/src/pipecat/services/cartesia/tts.py
@@ -25,6 +25,7 @@ from pipecat.frames.frames import (
     TTSAudioRawFrame,
     TTSStartedFrame,
     TTSStoppedFrame,
+    TTSTextFrame,
 )
 from pipecat.processors.frame_processor import FrameDirection
 from pipecat.services.tts_service import AudioContextWordTTSService, TTSService
@@ -107,6 +108,7 @@ class CartesiaTTSService(AudioContextWordTTSService):
         language: Optional[Language] = Language.EN
         speed: Optional[Union[str, float]] = ""
         emotion: Optional[List[str]] = []
+        add_timestamps: Optional[bool] = True
 
     def __init__(
         self,
@@ -176,6 +178,7 @@ class CartesiaTTSService(AudioContextWordTTSService):
             "speed": params.speed,
             "emotion": params.emotion,
         }
+        self._add_timestamps = params.add_timestamps
         self.set_model_name(model)
         self.set_voice(voice_id)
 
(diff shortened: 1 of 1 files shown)


golbin:
This issue may also affect other TTS services.

markbackman:
@golbin I'm sorry, but not being a Chinese, Japanese, or Korean speaker, I'm not sure I understand the root cause. Is the issue that Cartesia is delivering word/timestamp alignment data differently for CJK languages (compared to non-CJK languages) or is this a bug in Pipecat.

Related, in early June, I reported an issue to Cartesia about their Japanese word timestamps. The issue was that it stopped outputting data after a period of time. They acknowledged and mentioned that they would fix. I haven't followed up. Anyway, this makes me wonder where the root cause lies. Any info you can share would be helpful in figuring out the right solution.

golbin:
I just checked ElevenLabs as well, and it has the same issue.

markbackman:
> I just checked ElevenLabs as well, and it has the same issue.

Good to know I'd like to understand the fundamental issue. Is there a way you can lay out the problem in detail, including debug data?

For example, in 11Labs, words are split on spaces. Is that not working reliably or is there something inherently different about the other languages where we can't split on spaces?

golbin:
Simply refer this example.

When I say "안녕. 오늘 날씨 좋다."

When `expect_stripped_words=True`

LLMAssistantAggregator receive like "안 녕 . 오 늘 날 씨 좋 다 ."

When `expect_stripped_words=False`

LLMAssistantAggregator receive like "안녕.오늘날씨좋다."

And it's solved this patch is applied Cartesia but ElevenLabs still has this problem.

golbin:
I think Cartesia and ElevenLabs split sentences into characters, not words if sentences have CJK characters

markbackman:
Ok! I understand the issue and have a proposed fix:
https://github.com/pipecat-ai/pipecat/pull/2304

I think we can keep the change in the CartesiaTTSService class. Check it out. Closing in favor of #2304. Really appreciate you explaining this issue 🙌

golbin:
Thank you!

PR #2300, opened 2025-07-30
Fixed a race condition in FastAPIWebsocketClient

filipi87:
Fixed a race condition in FastAPIWebsocketClient that occurred when attempting to send a message while the client was disconnecting.

Fixes this [issue](https://github.com/pipecat-ai/pipecat/issues/2209).


aconchillo (approved):

aconchillo:
LGTM

markbackman (approved):
LGTM! I was looking at this code too and wondering if this is where the race condition resides. Thanks for tracking it down 🙌

PR #2303, opened 2025-07-31
Add quickstart demos

markbackman:

Three new quickstart examples that will live in the Pipecat repo:
1. examples/quickstart: This is _the_ quickstart which has a [docs](https://github.com/pipecat-ai/docs/pull/292) write up and will be the first example developers will use. It's simple and uses the new runner.

Next, developers will be looking for either a client/server example or phone. So we give them two options:
3. examples/phone-bot-twilio (name could be improved): This is just like the quickstart but with a different transport and a README informing how to do extra things to make a phone call
4. examples/client-server-web (again, better name?): This uses the quickstart bot file and provides a React/Vite app that uses the console from the voice-ui-kit. It gives the same outcome as 1, but teaches about the clients

2 and 3 will get write ups in the docs as well, as an entry point into their respective areas.


markbackman:
@filipi87 these are the same changes you approved here: https://github.com/pipecat-ai/pipecat/pull/2283.

I merged in the wrong order and created a mess of merge conflicts, so I just opened a new branch. Apologies!

PR #2321, opened 2025-08-01
Add dev to optional-dependencies

markbackman:

Adding `dev` to pyproject.toml so we can replace `pip install -r dev-requirements.txt` with `pip install -e .[dev]`

or, in uv land: `uv sync --extra dev`

This is a pre-req to updating our CI with uv, which will speed up times by 3-5x.


filipi87 (approved):
LGTM!

Issue #2322, opened 2025-08-01
AWS Transcribe Unsupported Language for Tagalog


kennethtegrado:
### pipecat version

0.0.77

### Python version

3.13.3

### Operating System

macOS 15.5

### Issue description

Unsupported error for tagalog even though AWS has support tagalog

### Reproduction steps

Just run AWSTranscribe with tl-PH

### Expected behavior

There should be no `AWSTranscribeSTTService#0 Failed to connect to AWS Transcribe: Unsupported language: tl-PH` since there's support for tl-PH in the official docs.

### Actual behavior

No error for tl-PH

### Logs

```shell
2025-08-01 12:14:21.263 | DEBUG    | pipecat.services.aws.stt:run_stt:180 - WebSocket not connected, attempting to reconnect...
2025-08-01 12:14:21.263 | DEBUG    | pipecat.services.aws.stt:_connect:224 - AWSTranscribeSTTService#0 Starting connection process...
2025-08-01 12:14:21.263 | ERROR    | pipecat.services.aws.stt:_connect:290 - AWSTranscribeSTTService#0 Failed to connect to AWS Transcribe: Unsupported language: tl-PH
2025-08-01 12:14:21.263 | WARNING  | pipecat.services.aws.stt:_disconnect:310 - AWSTranscribeSTTService#0 Error closing WebSocket connection: 'NoneType' object has no attribute 'close'
2025-08-01 12:14:21.263 | ERROR    | pipecat.services.aws.stt:run_stt:184 - Failed to reconnect: Unsupported language: tl-PH
2025-08-01 12:14:21.263 | ERROR    | pipecat.processors.frame_processor:_check_started:621 - UserTranscriptProcessor#0 Trying to process InputAudioRawFrame#646(pts: None, source: None, size: 640, frames: 320, sample_rate: 16000, channels: 1) but StartFrame not received yet
```

PR #2323, opened 2025-08-01
AWSTranscribeSTTService: Add remaining language mapping in language_to_service_language

kennethtegrado:
I added the remaining languages in AWS from this [link](https://docs.aws.amazon.com/transcribe/latest/dg/supported-languages.html) to `language_to_service_language` in `stt.py` under service for aws

Fixes: #2322

### DIFF
--- a/src/pipecat/services/aws/stt.py
+++ b/src/pipecat/services/aws/stt.py
@@ -321,16 +321,233 @@ class AWSTranscribeSTTService(STTService):
             AWS Transcribe compatible language code, or None if unsupported.
         """
         language_map = {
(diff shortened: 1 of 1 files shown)


markbackman:
Thanks for the contribution, but I think this needs some changes before it's ready.

1. The change includes dialects that aren't supported. (e.g. Spanish only supports `es-ES`, `es-US`, but a long list is included)
2. Only languages that support a `Data Input` of `streaming` will work, so the list needs to be narrowed down quite a bit

With these changes, I think this PR will be ready to go.

markbackman:
Closing in favor of #2362

PR #2327, opened 2025-08-01
Add source parameter to ErrorFrame and set it in FrameProcessor. Upda…

richtermb:
…ted error handling in AnthropicLLMService and DeepgramSTTService to include ErrorFrame with source information.


src/pipecat/frames/frames.py
```diff
@@ -652,10 +652,12 @@ class ErrorFrame(SystemFrame):
     Parameters:
         error: Description of the error that occurred.
         fatal: Whether the error is fatal and requires bot shutdown.
+        source: The frame processor that generated the error.
     """
 
     error: str
     fatal: bool = False
+    source: Optional[FrameProcessor] = None
```
aconchillo: You need to use `"FrameProcessor"` instead of `FrameProcessor`

src/pipecat/frames/frames.py
```diff
@@ -652,10 +652,12 @@ class ErrorFrame(SystemFrame):
     Parameters:
         error: Description of the error that occurred.
         fatal: Whether the error is fatal and requires bot shutdown.
+        source: The frame processor that generated the error.
     """
 
     error: str
     fatal: bool = False
+    source: Optional["FrameProcessor"] = None
```
richtermb: @aconchillo should I take inspiration from `MetricsFrame`  / `MetricsData` and use a `processor: str` field instead?

aconchillo: No, I think this is fine.

CHANGELOG.md
```diff
@@ -65,6 +65,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Added `on_transcription_stopped` and `on_transcription_error` to Daily
   callbacks.
 
+- Added `source` field to `ErrorFrame` to indicate `FrameProcessor` that generated the error.
```
aconchillo: I think you need to rebase and add this line to a new `### Added` section under `[Unreleased]`.

aconchillo: It's currently added to the release from yesterday.

aconchillo (approved):

src/pipecat/frames/frames.py
aconchillo: So sorry for the late comment. I think it would be better to rename `source` to `processor`. It is more clear and explicit and it matches other existing frame types.

markbackman: Before we merge, we should change the naming if that's what we want.

markbackman (approved):
LGTM!

PR #2328, opened 2025-08-01
Align 04b livekit example with other foundational examples

markbackman:

Structurally matching the other examples.

Removing unused Deepgram VAD params (to use these, you have to handle Deepgram events which this file doesn't do).


filipi87 (approved):
LGTM!

PR #2329, opened 2025-08-01
fix: LiveKitTransport, don't push empty AudioRawFrames

markbackman:

Reports in Discord of users seeing:
```
WARNING  | pipecat.services.stt_service:process_audio_frame:157 - Empty audio frame received for STT service: DeepgramSTTService#0 0
```

This was due to LiveKit pushing empty audio frames. The new warning in the STTService was flagging empty audio frames. This fix updates LiveKitInputTransport to only push non-zero AudioRawFrames.

### DIFF
--- a/src/pipecat/transports/services/livekit.py
+++ b/src/pipecat/transports/services/livekit.py
@@ -605,6 +605,11 @@ class LiveKitInputTransport(BaseInputTransport):
                 pipecat_audio_frame = await self._convert_livekit_audio_to_pipecat(
                     audio_frame_event
                 )
+
+                # Skip frames with no audio data
+                if len(pipecat_audio_frame.audio) == 0:
+                    continue
+
                 input_audio_frame = UserAudioRawFrame(
                     user_id=participant_id,
                     audio=pipecat_audio_frame.audio,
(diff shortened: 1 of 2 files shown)


src/pipecat/transports/services/livekit.py
```diff
@@ -605,6 +605,11 @@ async def _audio_in_task_handler(self):
                 pipecat_audio_frame = await self._convert_livekit_audio_to_pipecat(
                     audio_frame_event
                 )
+
+                # Skip frames with no audio data
+                if len(pipecat_audio_frame.audio) == 0:
+                    continue
```
aconchillo, line 611: Mmmm, this is weird. Does this mean the resampler is returning 0?

aconchillo: When I tried adding the streaming resampler the first time I believe I was seeing the resampler not always returning data, that's why I didn't use it. But I assumed I was doing something wrong. @filipi87 since you implemented the new resampler, you might now.

filipi87: Yep, that is exactly it @aconchillo.

Unlike the previous resampler, which always returns audio data, `create_stream_resampler` may sometimes return empty data.

This happens because `create_stream_resampler`, which uses `SOXRStreamAudioResampler`, maintains internal state between calls to preserve audio quality at chunk boundaries. It buffers audio internally to handle resampling more accurately, which can result in either empty output or a larger than expected output size due to this buffering behavior.

We have handled most of this issues in this previous [PR](https://github.com/pipecat-ai/pipecat/pull/2184).

But it looks like we have missed this fix for Livekit.

PR #2332, opened 2025-08-02
Fix PiperTTSService to send TTS input as JSON object

hankehly:
### Problem

The `PiperTTSService.run_tts` method currently sends a raw string as the POST payload:

```python
async with self._session.post(self._base_url, json=text, headers=headers) as response:
```

However, according to [Piper's HTTP API documentation](https://github.com/OHF-Voice/piper1-gpl/blob/main/docs/API_HTTP.md), the expected format is a JSON object:

```json
{ "text": "..." }
```

This discrepancy causes the test `tests/test_piper_tts.py::test_run_piper_tts_success` to fail, as the service responds with an error when the payload format is incorrect.

### Fix
This PR updates the request to wrap text in the required dictionary format.

### Related Issues
* PR #2293 (merged recently) also modified this part of the code but did not fix the incorrect payload structure.
* This PR addresses the root cause of the failing test and restores compatibility with the Piper HTTP API.

### Outcome
* Fixes the failing unit test test_run_piper_tts_success.
* Ensures compliance with Piper’s HTTP API contract.
* No changes to external interfaces or breaking changes introduced.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Fixed an issue in `LiveKitTransport` where empty `AudioRawFrame`s were pushed
   down the pipeline. This resulted in warnings by the STT processor.
 - Fixed `PiperTTSService` to send text as a JSON object in the request body,
-  resolving compatibility with Piper's HTTP API.  
+  resolving compatibility with Piper's HTTP API.
 
 ## [0.0.77] - 2025-07-31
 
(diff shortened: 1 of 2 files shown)


hankehly:
Done!

markbackman:
Sorry, one more thing—you'll need to fix the linting errors.

The easiest way will be to run:
```
./scripts/fix-ruff.sh
```

For future commits, you can install the pre-commit hook to alert you of formatting changes:
```
pre-commit install
```

hankehly:
Sorry. We should be good now.

Issue #2277, opened 2025-07-27
Azure TTS - Speech synthesis canceled: CancellationReason.Error


yohan-altrium:
### pipecat version

0.0.76

### Python version

3.11

### Operating System

macOS 15.5

### Issue description

I have a Pipecat project that uses Azure TTS. Somtimes, it fails with the following error. 

```
pipecat.processors.metrics.frame_processor_metrics:stop_ttfb_metrics:131 - AssemblyAISTTService#0 TTFB: 0.0025179386138916016
2025-07-27 19:40:04.720 | DEBUG    | pipecat.processors.metrics.frame_processor_metrics:stop_ttfb_metrics:131 - AzureTTSService#0 TTFB: 2.8991360664367676
2025-07-27 19:40:04.720 | DEBUG    | pipecat.processors.metrics.frame_processor_metrics:stop_processing_metrics:152 - AzureTTSService#0 processing time: 2.8993260860443115
2025-07-27 19:40:05.985 | ERROR    | pipecat.services.azure.tts:_handle_canceled:266 - Speech synthesis canceled: CancellationReason.Error
2025-07-27 19:40:05.985 | ERROR    | pipecat.services.azure.tts:_handle_canceled:266 - Speech synthesis canceled: CancellationReason.Error
```

When I looked at the source, I see the following:
```
    def _handle_canceled(self, evt):
        """Handle synthesis cancellation."""
        logger.error(f"Speech synthesis canceled: {evt.result.cancellation_details.reason}")
        self._audio_queue.put_nowait(None)
```

Is there a way to get more context about these CancellationReason.Error so that we know what caused it (network, rate limit etc). Azure's Console also doesn't give much information so it's difficult to understand what went wrong here.

### Reproduction steps

Sparodic.

### Expected behavior

Ideally it should not fail, and if it does, having more details about what went wrong would be helpful.

### Actual behavior

Fails with the given error message. Not enough context to understand why.

### Logs

```shell
pipecat.processors.metrics.frame_processor_metrics:stop_ttfb_metrics:131 - AssemblyAISTTService#0 TTFB: 0.0025179386138916016
2025-07-27 19:40:04.720 | DEBUG    | pipecat.processors.metrics.frame_processor_metrics:stop_ttfb_metrics:131 - AzureTTSService#0 TTFB: 2.8991360664367676
[...]

PR #2333, opened 2025-08-02
Fixes 2277 - SSML reserved characters causes Azure TTS to fail

yohan-altrium:
This is related to issue #2277. 

The current implementation of AzureTTS generates SSML markup for Azure by directly embedding the LLM generated text to it. However, SSML has reserved. characters and according to Microsoft documentation (https://github.com/pipecat-ai/pipecat/compare/main...yohan-altrium:pipecat:fix/2277-azure-tts-ssml-reserved-characters?expand=1), Azure expects the & sign, and < and > signs to be escaped and replaced with entity format. (Note: Microsoft does not seem to require us to escape quotes).

This pull request fixes this issue by maintaining a list of characters to be escaped in `AzureBaseTTSService`, and escaping the text to replace those sensitive characters with the correct entity format.

### DIFF
--- a/src/pipecat/services/azure/tts.py
+++ b/src/pipecat/services/azure/tts.py
@@ -128,6 +128,14 @@ class AzureBaseTTSService(TTSService):
             "volume": params.volume,
         }
 
+        # Define SSML escape mappings based on SSML reserved characters
+        # See - https://learn.microsoft.com/en-us/azure/ai-services/speech-service/speech-synthesis-markup-structure
+        self.ssml_escape_chars = {
+            '&': '&amp;',
+            '<': '&lt;',
+            '>': '&gt;'
+        }
+
         self._api_key = api_key
         self._region = region
         self._voice_id = voice
@@ -154,6 +162,10 @@ class AzureBaseTTSService(TTSService):
 
     def _construct_ssml(self, text: str) -> str:
         language = self._settings["language"]
+
+        # Escape special characters
+        escaped_text = self._escape_text(text)
+
         ssml = (
             f"<speak version='1.0' xml:lang='{language}' "
             "xmlns='http://www.w3.org/2001/10/synthesis' "
@@ -183,7 +195,7 @@ class AzureBaseTTSService(TTSService):
         if self._settings["emphasis"]:
             ssml += f"<emphasis level='{self._settings['emphasis']}'>"
 
-        ssml += text
+        ssml += escaped_text
 
         if self._settings["emphasis"]:
             ssml += "</emphasis>"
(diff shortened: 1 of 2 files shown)


src/pipecat/services/azure/tts.py
```diff
@@ -128,6 +128,14 @@ def __init__(
             "volume": params.volume,
         }
 
+        # Define SSML escape mappings based on SSML reserved characters
+        # See - https://learn.microsoft.com/en-us/azure/ai-services/speech-service/speech-synthesis-markup-structure
+        self.ssml_escape_chars = {
```
markbackman: Let's make this a class constant and move to the top of the class:
```
SSML_ESCAPE_CHARS = {
    '&': '&amp;',
    '<': '&lt;',
    '>': '&gt;',
}
```

Also, from the docs:
> Special characters such as quotation marks, apostrophes, and brackets, must be escaped. For more information, see [Extensible Markup Language (XML) 1.0: Appendix D](https://www.w3.org/TR/xml/#sec-entexpand).

Shall we handle this too?

src/pipecat/services/azure/tts.py
```diff
@@ -197,6 +209,20 @@ def _construct_ssml(self, text: str) -> str:
 
         return ssml
 
+    def _escape_text(self, text: str) -> str:
+        """Escape special characters in text for SSML.
```
markbackman:
```suggestion
        """Escape special characters in text for SSML.
        
        Escapes XML/SSML reserved characters according to Microsoft documentation:
        - & becomes &amp;
        - < becomes &lt;
        - > becomes &gt;
```

src/pipecat/services/azure/tts.py
```diff
@@ -197,6 +209,20 @@ def _construct_ssml(self, text: str) -> str:
...
+
+        Args:
+            text: The text to escape.
+
+        Returns:
+            The escaped text.
+        """
+        escaped_text = text
+        for char, escape_code in self.ssml_escape_chars.items():
```
markbackman: Use the class constant from above.
```suggestion
        for char, escape_code in self.SSML_ESCAPE_CHARS.items():
```

markbackman (commented):
Looks good! Thanks for catching this. Just a few comments before this is ready to merge.

Can you also add a changelog entry?

Reminder: Don't forget to install pre-commit hooks:
```
uv run pre-commit install
```

yohan-altrium:
Hi @markbackman,

Sure, I'll do the suggested changes and update this PR. 

Regarding the quotes and apostrophes, although it is not well formed XML, it seems that Azure supports those. It worked fine when I tested so I left those as is. Let me try to escape those and test to confirm that it works after escaping (which it should, but just want to make sure). If it works fine, I'll escape those too.

PR #2336, opened 2025-08-02
Suppress pydub SyntaxWarning from the cartesia module

markbackman:

`cartesia` imports `pydub`. `pydub` currently has an invalid escape sequence, resulting in these errors spilling over to the console:
```
/Users/mbackman/src/pipecat/.venv/lib/python3.12/site-packages/pydub/utils.py:300: SyntaxWarning: invalid escape sequence '\('
  m = re.match('([su]([0-9]{1,2})p?) \(([0-9]{1,2}) bit\)$', token)
/Users/mbackman/src/pipecat/.venv/lib/python3.12/site-packages/pydub/utils.py:301: SyntaxWarning: invalid escape sequence '\('
  m2 = re.match('([su]([0-9]{1,2})p?)( \(default\))?$', token)
/Users/mbackman/src/pipecat/.venv/lib/python3.12/site-packages/pydub/utils.py:310: SyntaxWarning: invalid escape sequence '\('
  elif re.match('(flt)p?( \(default\))?$', token):
/Users/mbackman/src/pipecat/.venv/lib/python3.12/site-packages/pydub/utils.py:314: SyntaxWarning: invalid escape sequence '\('
  elif re.match('(dbl)p?( \(default\))?$', token):
```

This issue is tracked here:
https://github.com/jiaaro/pydub/issues/801

It's over a year old, so it's unclear when it will be resolved.

This clutter can be confusing as developers may think this comes from Pipecat. Since this is out of our control, we should suppress the warning. This PR adds targeted filterwarnings for this message.

### DIFF
--- a/src/pipecat/services/cartesia/tts.py
+++ b/src/pipecat/services/cartesia/tts.py
@@ -34,6 +34,10 @@ from pipecat.utils.text.base_text_aggregator import BaseTextAggregator
 from pipecat.utils.text.skip_tags_aggregator import SkipTagsAggregator
 from pipecat.utils.tracing.service_decorators import traced_tts
 
+# Suppress regex warnings from pydub (used by cartesia)
+warnings.filterwarnings("ignore", message="invalid escape sequence", category=SyntaxWarning)
+
+
 # See .env.example for Cartesia configuration needed
 try:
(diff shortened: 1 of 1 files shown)


vipyne (approved):
👍

PR #2340, opened 2025-08-02
Add body args support to development runner

markbackman:

Parity with Pipecat Cloud. You can pass in `body` via CLI and then the `bot()` method makes it available to the application code.

Just like PCC, this is compatible only with Daily at this point in time.

For now, SmallWebRTCTransport doesn't support this. But, SmallWebRTCTransport is not yet support for PCC. Once that's added, we'll add types and match support here too.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Added the ability to pass `body` properties to the development runner using
+  the `--body`/`-b` command. The `body` accepts an object provided using valid
+  JSON format. Currently, the `body` is only accepted when using
+  `DailyTransport`.
+
 ### Fixed
 
 - Fixed an issue in `LiveKitTransport` where empty `AudioRawFrame`s were pushed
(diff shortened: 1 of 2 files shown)


chadbailey59:
Looks good to me... but to be a _bit_ pedantic, it's weird that `DailyRunnerArguments` takes the `body` parameter, but the `DailyTransport` doesn't actually do anything with it. In our example code create_transport, it just kind of gets eaten. My current approach (including a type workaround for PCC, and getting the PCC session ID) is this:

```python
    logger.debug(f"runner args: {runner_args}")
    if isinstance(runner_args, PCCDailySessionArguments):
        session_id = runner_args.session_id
        runner_args = DailyRunnerArguments(
            room_url=runner_args.room_url,
            token=runner_args.token,
            body=runner_args.body,
        )
    else:
        session_id = None

    if isinstance(runner_args, DailyRunnerArguments):
        body = runner_args.body
    else:
        body = None
    transport = await create_transport(runner_args, transport_params)
    await run_bot(transport, session_id, body)
```

markbackman:
> it's weird that DailyRunnerArguments takes the body parameter, but the DailyTransport doesn't actually do anything with it.

That's not quite how it works. `DailyRunnerArguments` is a type, so it doesn't take any inputs. The runner mimics the PCC bot starter, which injects information into the bot.py file via the bot() method. The inserted information needs to abide by the `DailyRunnerArguments` type.

Also, in all cases, the DailyTransport doesn't use the `body` information ever—not in the developer runner or in the DailyTransport. You can parse the body, which can contain information to be used to set things in the application code. One application could be values for DailyTransport args, but it could also be something like prompt information for the LLM or a voice_id for the TTS service.

Does that make sense?

markbackman:
I need to rework this anyway. I'm going to close this in favor of a different approach. The data needs to be handled in the runner via a call to the /connect endpoint. That change is hopefully coming at some point this weekend.

PR #2342, opened 2025-08-02
Development runner handles body information in the RTVI connect request

markbackman:

Check out the changelog for the changes.

The motivation is for the development runner to more closely mimic the Pipecat Cloud `/start` endpoint.

All of the changes are intended to reduce the amount of custom code needs to be written between a local and production deployment.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,11 +9,29 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- The development runner how handles custom `body` data for `DailyTransport`.
+  The `body` data is passed to the Pipecat client. You can POST to the `/start`
+  endpoint with a request body of:
+
+  ```
+  {
+      "createDailyRoom": true,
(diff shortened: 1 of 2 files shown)


chadbailey59 (approved):

CHANGELOG.md
```diff
@@ -7,11 +7,36 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- The development runner how handles custom `body` data, which is passed to the
+  Pipecat client. You can POST to the `/start` endpoint with a request body of:
```
filipi87: I know that, by looking at the body below, that should be clear, but should we mention here that we're handling this for the DailyTransport case ?

markbackman: Good point. Yes. Once we add Small WebRTC to Pipecat Cloud, we'll add body support in the runner for it, too. This will avoid a potential breaking change, since the runner follows the patterns of Pipecat Cloud.

filipi87 (approved):
Nice improvement.

PR #2345, opened 2025-08-03
TaskObserver: don't inspect on_push_frame signature for every frame

aconchillo:

This PR fixes a performance issue where for each received frame in each observer we were inspecting the `on_push_frame` to decide if this was the old or new signature. This is only necessary once.


markbackman (approved):
LGTM

aconchillo:
I added a similar thing for the LLMService. Not as critical as the observer one, it's just a very small optimization in that case.

PR #2346, opened 2025-08-03
me trying to get heygen working

adenta:

### DIFF
--- a/examples/foundational/43a-heygen-video-service.py
+++ b/examples/foundational/43a-heygen-video-service.py
@@ -4,6 +4,18 @@
 # SPDX-License-Identifier: BSD 2-Clause License
 #
 
+# /// script
+# requires-python = ">=3.11"
+# dependencies = [
+#   "pipecat-ai>=0.0.77",
+#   "deepgram-sdk",
+#   "cartesia",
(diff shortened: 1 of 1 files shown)


examples/foundational/43a-heygen-video-service.py
```diff
@@ -4,6 +4,18 @@
 # SPDX-License-Identifier: BSD 2-Clause License
 #
 
+# /// script
```
markbackman, line 7: I love this idea!! I'm going to add to others.

This is not quite right, though. It should be:
```
# /// script
# dependencies = [
#   "pipecat-ai[daily,webrtc,silero,deepgram,cartesia,google,heygen,runner]>=0.0.77",
# ]
# ///
```

You don't need to specify `requires-python` as that's part of `pipecat-ai` already.

Also, you should install service packages via the optional dependencies from pipecat-ai. You can see those options here:
https://github.com/pipecat-ai/pipecat/blob/main/pyproject.toml#L45

markbackman (approved):
LGTM! Thanks. I've already completed this for the remaining examples: https://github.com/pipecat-ai/pipecat/pull/2347

PR #2347, opened 2025-08-03
Update foundational examples with script dependencies

markbackman:

Add [inline script dependencies](https://packaging.python.org/en/latest/specifications/inline-script-metadata/#inline-script-metadata) to all of the foundational examples. With this, you can do:

```
uv sync
uv run example.py
```

No requirements.txt to deal with.

This is supported by `uv` and a bunch of other python tools.

---

UPDATE:
This makes like much easier for developers, which is great, but makes testing more difficult, as these files won't run from `main` :/

But, we do have a way to support this, if we want to:
```
import os

# Only include script dependencies if not in development mode
if not os.getenv("PIPECAT_DEV_MODE"):
    # /// script
    # dependencies = [
    #   "pipecat-ai[daily,webrtc,websocket,runner,silero,deepgram,google,cartesia,heygen]>=0.0.77",
    # ]
    # ///
    pass
```

It's a bit messy to have at the top of each file, but it would allow:
- development to use main or whatever is installed locally
- developers using the file to not have to fumble with requirements installation

I'm on the fence about what to do...


kwindla:
I feel like we use these scripts a lot in development, so it's valuable to be able to run them against the current local pipecat project files.

It seems like your PIPECAT_DEV_MODE environment variable that makes that possible is worth having in place in all these scripts, even though it's extra complexity.

markbackman:
We've opted to _not_ move in this direction. Instead, we'll recommend that developers:
```
uv sync --all-extras --no-extra krisp
```

which will install all deps. We'll make this guidance clear for each file. Closing this out for this reason.

PR #2352, opened 2025-08-04
Implementing if the bot it is speaking or not based on the SpeechOutputAudioRawFrame

filipi87:
Fixed an issue where `BotStartedSpeakingFrame` and `BotStoppedSpeakingFrame` were not emitted when using `TavusVideoService` or `HeyGenVideoService`.

### DIFF
--- a/src/pipecat/frames/frames.py
+++ b/src/pipecat/frames/frames.py
@@ -239,6 +239,18 @@ class TTSAudioRawFrame(OutputAudioRawFrame):
 
 
 @dataclass
+class SpeechOutputAudioRawFrame(OutputAudioRawFrame):
+    """An audio frame part of a speech audio stream.
+
+    This frame is part of a continuous stream of audio frames containing speech.
+    The audio stream might also contain silence frames, so a process to distinguish
+    between speech and silence might be needed.
+    """
+
+    pass
+
+
+@dataclass
 class URLImageRawFrame(OutputImageRawFrame):
     """Image frame with an associated URL.
 
--- a/src/pipecat/transports/base_output.py
+++ b/src/pipecat/transports/base_output.py
@@ -21,7 +21,7 @@ from loguru import logger
 from PIL import Image
 
 from pipecat.audio.mixers.base_audio_mixer import BaseAudioMixer
-from pipecat.audio.utils import create_stream_resampler
+from pipecat.audio.utils import create_stream_resampler, is_silence
 from pipecat.frames.frames import (
     BotSpeakingFrame,
     BotStartedSpeakingFrame,
@@ -35,6 +35,7 @@ from pipecat.frames.frames import (
     OutputDTMFUrgentFrame,
     OutputImageRawFrame,
     OutputTransportReadyFrame,
+    SpeechOutputAudioRawFrame,
     SpriteFrame,
     StartFrame,
     StartInterruptionFrame,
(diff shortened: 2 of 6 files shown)


src/pipecat/frames/frames.py
```diff
@@ -238,6 +238,18 @@ class TTSAudioRawFrame(OutputAudioRawFrame):
     pass
 
 
+@dataclass
+class WebRTCOutputAudioRawFrame(OutputAudioRawFrame):
```
markbackman: Should this be:
- `OutputWebRTCAudioRawFrame`
- `WebRTCOutputAudioRawFrame`

Any strong feelings?

filipi87: Yeah, no strong feelings, the other options that I had thought:
- `StreamOutputAudioRawFrame`
- `OutputStreamAudioRawFrame`

filipi87: Another option:
- `VoiceOutputAudioRawFrame`.

filipi87: Probably the winner: `StreamOutputAudioRawFrame`.
@aconchillo it is going to give some extra thought about the name.

filipi87: - `SpeechOutputAudioRawFrame`

filipi87:
@aconchillo, I’ve refactored the code as we discussed in the meeting. Let me know what you think.

src/pipecat/frames/frames.py
```diff
@@ -238,6 +238,18 @@ class TTSAudioRawFrame(OutputAudioRawFrame):
...
 
 
+@dataclass
+class SpeechOutputAudioRawFrame(OutputAudioRawFrame):
+    """Raw audio frame received from integrated service.
+
+    This frame contains either pure speech or complete silence,
+    without any background noise. It represents audio output
+    generated by external TTS services, typically clean and filtered.
+    """
```
aconchillo, line 248: I would change this comment to something like below, to indicate this is a stream of audio containing speech. I would not mention TTS, because then it's confusing with TTSAudioRawFrame.

"""An audio frame part of a speech audio stream.

This frame is part of a continuous stream of audio frames containing speech. The audio stream might also contain silence frames, so a process to distinguish between speech and silence might be needed.
"""

src/pipecat/transports/base_output.py
```diff
@@ -671,10 +672,24 @@ async def _audio_task_handler(self):
             TOTAL_CHUNK_MS = self._params.audio_out_10ms_chunks * 10
             BOT_SPEAKING_CHUNK_PERIOD = max(int(200 / TOTAL_CHUNK_MS), 1)
             bot_speaking_counter = 0
+            webrtc_audio_last_speaking_time = 0
```
aconchillo: rename to `speech_last_speaking_time`, don't mention `webrtc`

filipi87: Oops, nice catch.

src/pipecat/frames/frames.py
filipi87: Done.

src/pipecat/transports/base_output.py
filipi87: Done.

aconchillo:
LGTM

PR #2354, opened 2025-08-04
Remove inline script from foundational 43a

markbackman:

Removing the inline script dependencies. Because the scripts are required to use .toml syntax, there's no way to install packages from main. This disrupts the development and eval workflows. We'll follow up with added guidance about how to run these files.


aconchillo (approved):

aconchillo:
LGTM

PR #2356, opened 2025-08-04
Added Sarvam TTS Websocket Implementation

pratham-sarvam:
Added TTS websocket implementation for Sarvam.


vinayak-sarvam:
Hi @markbackman, Can you please review and provide your feedback on this PR.

src/pipecat/services/sarvam/tts.py
```diff
@@ -240,3 +248,327 @@ async def run_tts(self, text: str) -> AsyncGenerator[Frame, None]:
         finally:
             await self.stop_ttfb_metrics()
             yield TTSStoppedFrame()
+
+
+class SarvamWebsocketTTSService(WebsocketTTSService):
```
markbackman: The convention is for the websocket service to be called `ServiceTTSService` and the HTTP service to be called `ServiceHttpTTSService`. So, let's update the HTTP implementation to `SarvamHttpTTSService`.

Also, this will need to be a different base class—`InterruptibleTTSService`. (I know, it's confusing! We're refactoring the base classes.)

```suggestion
class SarvamTTSService(InterruptibleTTSService):
```

examples/foundational/07z-interruptible-sarvam.py
```diff
@@ -18,7 +18,7 @@
 from pipecat.processors.aggregators.openai_llm_context import OpenAILLMContext
 from pipecat.services.deepgram.stt import DeepgramSTTService
 from pipecat.services.openai.llm import OpenAILLMService
-from pipecat.services.sarvam.tts import SarvamTTSService
+from pipecat.services.sarvam.tts import SarvamWebsocketTTSService
```
markbackman: Per my other suggestion, just leave this as `SarvamTTSService`. This will use the ws implementation.

examples/foundational/07z-interruptible-sarvam.py
```diff
@@ -56,10 +56,11 @@ async def run_example(transport: BaseTransport, _: argparse.Namespace, handle_si
     async with aiohttp.ClientSession() as session:
```
markbackman: Since the websocket implementation doesn't need a session, you can just remove this.

src/pipecat/services/sarvam/tts.py
```diff
@@ -240,3 +248,327 @@ async def run_tts(self, text: str) -> AsyncGenerator[Frame, None]:
...
+    """Minimalist TTS service for Sarvam AI WebSocket endpoint using Pipecat base class.
```
markbackman: These docstrings are used in auto-generating docs, so you might want something a little more complete.

Check out the docstring from the current service as an example.

src/pipecat/services/sarvam/tts.py
```diff
@@ -240,3 +248,327 @@ async def run_tts(self, text: str) -> AsyncGenerator[Frame, None]:
...
+    Supports streaming text-to-speech with buffering and real-time audio generation.
+    """
+
+    def __init__(
+        self,
+        *,
+        api_key: str,
+        model: str = "bulbul:v2",
+        target_language_code: str = "hi-IN",
+        pitch: float = 0.0,
```
markbackman: Move args like `pitch`, `pace`, `loudness`, `enable_preprocessing`, `min_buffer_size`, `max_chunk_length` to an `InputParams` class. This separates out the core functionality from configurability.

src/pipecat/services/sarvam/tts.py
```diff
@@ -240,3 +248,327 @@ async def run_tts(self, text: str) -> AsyncGenerator[Frame, None]:
...
+        pace: float = 1.0,
+        speaker: str = "anushka",
```
markbackman: `voice_id` is the convention across TTS services in Pipecat.

```suggestion
        voice_id: str = "anushka",
```

src/pipecat/services/sarvam/tts.py
```diff
@@ -240,3 +248,327 @@ async def run_tts(self, text: str) -> AsyncGenerator[Frame, None]:
...
+            output_audio_bitrate: Bitrate for output audio format (default “128k”).
+            **kwargs: Optional keyword arguments forwarded to WebsocketTTSService (such as
+                `push_stop_frames`, `sample_rate`, task manager parameters, event hooks, etc.)
+                to customize transport behavior or enable metrics support.
+
+        This method sets up the internal TTS configuration mapping, constructs the WebSocket
+        URL based on the chosen model, and initializes state flags before connecting.
+        """
+        # Initialize parent class first
+        super().__init__(
```
markbackman, line 350: Should be:
```
        super().__init__(
            push_text_frames=False,
            pause_frame_processing=True,
            push_stop_frames=True,
            sample_rate=sample_rate,
            **kwargs,
        )
```

src/pipecat/services/sarvam/tts.py
```diff
@@ -240,3 +248,327 @@ async def run_tts(self, text: str) -> AsyncGenerator[Frame, None]:
...
+            push_stop_frames=True,
+            sample_rate=speech_sample_rate,
+            **kwargs,
+        )
+
+        # WebSocket endpoint URL
+        self._websocket_url = f"wss://api.sarvam.ai/text-to-speech/ws?model={model}"
```
markbackman: I think it could be helpful to add `base_url` to the class args:
```
url: str = "wss://api.sarvam.ai/text-to-speech/ws",
```

then set: `self._websocket_url = f"{url}?model={model}"`.

src/pipecat/services/sarvam/tts.py
```diff
@@ -6,23 +6,31 @@
 
 """Sarvam AI text-to-speech service implementation."""
 
+import asyncio
 import base64
+import json
 from typing import AsyncGenerator, Optional
 
 import aiohttp
+import websockets
```
markbackman: This needs to be in a try/except. Also, we support the asyncio implementation now, so you need to implement websockets slightly differently. I'll comment throughout.

```
try:
    logger.error("In order to use Sarvam, you need to `pip install pipecat-ai[sarvam]`.")
```

src/pipecat/services/sarvam/tts.py
```diff
@@ -240,3 +248,327 @@ async def run_tts(self, text: str) -> AsyncGenerator[Frame, None]:
...
+        if self._keepalive_task:
+            await self.cancel_task(self._keepalive_task)
+            self._keepalive_task = None
+
+        await self._disconnect_websocket()
+
+    async def _connect_websocket(self):
+        """Establish WebSocket connection to Sarvam API."""
+        try:
+            if self._websocket and not self._websocket.closed:
```
markbackman: websocket asyncio change

Also, is it OK to make this check for the websocket being OPEN?

Docs:
https://websockets.readthedocs.io/en/stable/howto/upgrade.html#open-and-closed

```suggestion
            if self._websocket and self._websocket.state is State.OPEN:
```

src/pipecat/services/sarvam/tts.py
```diff
@@ -240,3 +248,327 @@ async def run_tts(self, text: str) -> AsyncGenerator[Frame, None]:
...
+
+        Args:
+            frame: The frame to push.
+            direction: The direction to push the frame.
+        """
+        await super().push_frame(frame, direction)
+        if isinstance(frame, (TTSStoppedFrame, StartInterruptionFrame)):
+            self._started = False
+
+    async def _flush_and_reconnect(self):
```
markbackman: This is unused. Can we remove?

src/pipecat/services/sarvam/tts.py
```diff
@@ -240,3 +248,327 @@ async def run_tts(self, text: str) -> AsyncGenerator[Frame, None]:
...
+    async def _connect_websocket(self):
+        """Establish WebSocket connection to Sarvam API."""
+        try:
+            if self._websocket and not self._websocket.closed:
+                logger.debug("Webscoket already connected")
+                return
+
+            logger.debug("Connecting to Sarvam TTS Websocket")
+            subprotocols = [f"api-subscription-key.{self._api_key}"]
+            self._websocket = await websockets.connect(
```
markbackman: websocket asyncio change

```suggestion
            self._websocket = await websocket_connect(
```

src/pipecat/services/sarvam/tts.py
```diff
@@ -240,3 +248,327 @@ async def run_tts(self, text: str) -> AsyncGenerator[Frame, None]:
...
+                self._websocket_url,
+                subprotocols=subprotocols,
```
markbackman: Should we just use `additional_headers` (ws asyncio replacement for extra_headers) to set the X-API-Key?

src/pipecat/services/sarvam/tts.py
```diff
@@ -240,3 +248,327 @@ async def run_tts(self, text: str) -> AsyncGenerator[Frame, None]:
...
+                ping_interval=None,  # We'll handle pings manually
```
markbackman: Why are these pings and timeouts needed?

src/pipecat/services/sarvam/tts.py
```diff
@@ -240,3 +248,327 @@ async def run_tts(self, text: str) -> AsyncGenerator[Frame, None]:
...
+        self,
+        *,
+        api_key: str,
+        model: str = "bulbul:v2",
+        target_language_code: str = "hi-IN",
+        pitch: float = 0.0,
+        pace: float = 1.0,
+        speaker: str = "anushka",
+        loudness: float = 1.0,
+        speech_sample_rate: int = 24000,
```
markbackman: This needs to be initialized to None. The sample rate for the service is set by the PipelineTask, which sends a StartFrame with the sample_rate value. This ensures that all output components set the value accordingly.

Also, nit: let's call this `sample_rate` to match other TTS classes.

```suggestion
        sample_rate: Optional[int] = None,
```

src/pipecat/services/sarvam/tts.py
```diff
@@ -240,3 +248,327 @@ async def run_tts(self, text: str) -> AsyncGenerator[Frame, None]:
...
+        self.set_model_name(model)
+        self.set_voice(speaker)
+        # Configuration parameters
+        self._config = {
+            "target_language_code": target_language_code,
+            "pitch": pitch,
+            "pace": pace,
+            "speaker": speaker,
+            "loudness": loudness,
+            "speech_sample_rate": speech_sample_rate,
```
markbackman: Set to `0` as we'll initialize the value in `start()`.

```suggestion
            "sample_rate": 0,
```

src/pipecat/services/sarvam/tts.py
```diff
@@ -240,3 +248,327 @@ async def run_tts(self, text: str) -> AsyncGenerator[Frame, None]:
...
+            **kwargs,
+        )
+
+        # WebSocket endpoint URL
+        self._websocket_url = f"wss://api.sarvam.ai/text-to-speech/ws?model={model}"
+        self._api_key = api_key
+        self.set_model_name(model)
+        self.set_voice(speaker)
+        # Configuration parameters
+        self._config = {
```
markbackman: nit: we've called this `_settings` in other TTS services.

vinayak-sarvam:
Hi @markbackman, Appreciate your prompt response, thanks for the detailed feedback. Our team will look into this.

src/pipecat/services/sarvam/tts.py
```diff
@@ -240,3 +248,327 @@ async def run_tts(self, text: str) -> AsyncGenerator[Frame, None]:
...
+
+        Args:
+            api_key: Sarvam API key for authenticating TTS requests.
+            model: Identifier of the Sarvam speech model (default “bulbul:v2”).
+            target_language_code: BCP-47 code of the target language (default “hi-IN”).
+            pitch: Relative pitch adjustment (default 0.0).
+            pace: Playback speed multiplier (default 1.0).
+            speaker: Speaker voice name (default “anushka”).
+            loudness: Relative audio volume (default 1.0).
+            speech_sample_rate: Sampling frequency for output audio (default 24000 Hz).
```
```suggestion
            sample_rate: Sampling frequency for output audio (default 24000 Hz).
```

src/pipecat/services/sarvam/tts.py
```diff
@@ -240,3 +248,327 @@ async def run_tts(self, text: str) -> AsyncGenerator[Frame, None]:
...
+                `push_stop_frames`, `sample_rate`, task manager parameters, event hooks, etc.)
+                to customize transport behavior or enable metrics support.
+
+        This method sets up the internal TTS configuration mapping, constructs the WebSocket
+        URL based on the chosen model, and initializes state flags before connecting.
+        """
+        # Initialize parent class first
+        super().__init__(
+            push_stop_frames=True,
+            sample_rate=speech_sample_rate,
```
```suggestion
            sample_rate=sample_rate,
```

src/pipecat/services/sarvam/tts.py
```diff
@@ -240,3 +248,327 @@ async def run_tts(self, text: str) -> AsyncGenerator[Frame, None]:
...
+        """
+        return language_to_sarvam_language(language)
+
+    async def start(self, frame: StartFrame):
+        """Start the Sarvam TTS service.
+
+        Args:
+            frame: The start frame containing initialization parameters.
+        """
+        await super().start(frame)
```
markbackman, line 417: Assuming the naming changes I'm recommending, we initialize the sample_rate here.

```suggestion
        await super().start(frame)
        self._settings["sample_rate"] = self.sample_rate
```

This sets the value from the TTSService base class, which is handling the StartFrame and exposing a `sample_rate` property.

src/pipecat/services/sarvam/tts.py
```diff
@@ -240,3 +248,327 @@ async def run_tts(self, text: str) -> AsyncGenerator[Frame, None]:
...
+                await self._websocket.close()
+        except Exception as e:
+            logger.error(f"{self} error closing websocket: {e}")
+        finally:
+            self._started = False
+            self._websocket = None
+
+    async def _receive_messages(self):
+        """Receive and process messages from Sarvam WebSocket."""
+        async for message in WatchdogAsyncIterator(self._websocket, manager=self.task_manager):
```
markbackman: Add a new function called `_get_websocket()`:
```
    def _get_websocket(self):
        if self._websocket:
            return self._websocket
        raise Exception("Websocket not connected")
```

Then, here, use it:
```
async for message in WatchdogAsyncIterator(
    self._get_websocket(), manager=self.task_manager
):
```

src/pipecat/services/sarvam/tts.py
```diff
@@ -240,3 +248,327 @@ async def run_tts(self, text: str) -> AsyncGenerator[Frame, None]:
...
+                    error_msg = msg["data"]["message"]
+                    logger.error(f"TTS Error: {error_msg}")
+
+                    # If it's a timeout error, the connection might need to be reset
+                    if "too long" in error_msg.lower() or "timeout" in error_msg.lower():
+                        logger.warning("Connection timeout detected, service may need restart")
+
+                    await self.push_frame(ErrorFrame(f"TTS Error: {error_msg}"))
+
+                elif msg.get("type") == "pong":
```
markbackman: Is this needed?

src/pipecat/services/sarvam/tts.py
```diff
@@ -240,3 +248,327 @@ async def run_tts(self, text: str) -> AsyncGenerator[Frame, None]:
...
+        Args:
+            text: The text input to synthesize.
+
+        Yields:
+            Frame objects including TTSStartedFrame, TTSAudioRawFrame(s), or TTSStoppedFrame.
+        """
+        logger.debug(f"Generating TTS: [{text}]")
+
+        try:
+            if not self._websocket or (getattr(self._websocket, "close_code", None) is not None):
```
markbackman: websocket asyncio:

```suggestion
            if not self._websocket or self._websocket.state is State.CLOSED:
```

src/pipecat/services/sarvam/tts.py
```diff
@@ -240,3 +248,327 @@ async def run_tts(self, text: str) -> AsyncGenerator[Frame, None]:
...
+                await self._connect()
+
+            try:
+                if not self._started:
+                    await self.start_ttfb_metrics()
+                    yield TTSStartedFrame()
+                    self._started = True
+                    self._cumulative_time = 0
```
markbackman: Remove: unused. This would be used for word/timestamp pairs.

src/pipecat/services/sarvam/tts.py
```diff
@@ -240,3 +248,327 @@ async def run_tts(self, text: str) -> AsyncGenerator[Frame, None]:
...
+            "speech_sample_rate": speech_sample_rate,
+            "enable_preprocessing": enable_preprocessing,
+            "min_buffer_size": min_buffer_size,
+            "max_chunk_length": max_chunk_length,
+            "output_audio_codec": output_audio_codec,
+            "output_audio_bitrate": output_audio_bitrate,
+        }
+        # self._websocket = None
+        self._started = False
+        self._cumulative_time = 0
```
markbackman: Remove: unused. This would be used for word/timestamp pairs.

src/pipecat/services/sarvam/tts.py
```diff
@@ -240,3 +248,327 @@ async def run_tts(self, text: str) -> AsyncGenerator[Frame, None]:
...
+                    await self.start_ttfb_metrics()
+                    yield TTSStartedFrame()
+                    self._started = True
+                    self._cumulative_time = 0
+                await self._send_text(text)
+                await self.start_tts_usage_metrics(text)
+            except Exception as e:
+                logger.error(f"{self} error sending message: {e}")
+                yield TTSStoppedFrame()
+                logger.debug("error wala disconnect")
```
markbackman: What is `wala`?

src/pipecat/services/sarvam/tts.py
```diff
@@ -240,3 +248,327 @@ async def run_tts(self, text: str) -> AsyncGenerator[Frame, None]:
...
+
+    def can_generate_metrics(self) -> bool:
+        """Check if this service can generate processing metrics.
+
+        Returns:
+            True, as Sarvam service supports metrics generation.
+        """
+        return True
+
+    def language_to_service_language(self, language: Language) -> Optional[str]:
```
markbackman, line 400: You might want to add a method to change voices. I'm assuming we'd have to disconnect the websocket to accomplish that, right? (Assuming suggested names)

```
    async def _update_settings(self, settings: Mapping[str, Any]):
        """Update service settings and reconnect if voice changed."""
        prev_voice = self._voice_id
        await super()._update_settings(settings)
        if not prev_voice == self._voice_id:
            logger.info(f"Switching TTS voice to: [{self._voice_id}]")
            await self._disconnect()
            await self._connect()
```

markbackman: This uses the base class to handle the frame and update the value. You can test this by pushing a `TTSUpdateSettingsFrame`:

In your example:
```
        # Kick off the conversation.
        messages.append({"role": "system", "content": "Please introduce yourself to the user."})
        await task.queue_frames([context_aggregator.user().get_context_frame()])
        # Wait 15 seconds and then change the voice.
        await asyncio.sleep(15)
        await task.queue_frame(TTSUpdateSettingsFrame(settings={"voice_id": "new-voice-id"}))
```

markbackman:
Looks pretty good!

In addition to the comments, can you also update `pyproject.toml` to have an optional dependency:
```
sarvam = [ "websockets>=13.1,<15.0" ]
```

Lastly, please install the pre-commit hook:
```
uv run pre-commit install
```

This will help to ensure linting is correct.

src/pipecat/services/sarvam/tts.py
pratham-sarvam: Addressed in a PR

examples/foundational/07z-interruptible-sarvam.py
pratham-sarvam: Addressed in a PR

examples/foundational/07z-interruptible-sarvam.py
pratham-sarvam: Addressed in a PR

src/pipecat/services/sarvam/tts.py
pratham-sarvam: @vinayak-sarvam can you please pick up these docstring as per our internal policy.

src/pipecat/services/sarvam/tts.py
pratham-sarvam: Addressed in a PR

src/pipecat/services/sarvam/tts.py
pratham-sarvam: Addressed in a PR

src/pipecat/services/sarvam/tts.py
pratham-sarvam: Addressed in a PR

src/pipecat/services/sarvam/tts.py
pratham-sarvam: Addressed in a PR

src/pipecat/services/sarvam/tts.py
pratham-sarvam: Addressed in a PR

src/pipecat/services/sarvam/tts.py
pratham-sarvam: Addressed in a PR

src/pipecat/services/sarvam/tts.py
pratham-sarvam: This is not removed as of now, might need it in change voice function, will remove after implementing all the changes.

src/pipecat/services/sarvam/tts.py
pratham-sarvam: Addressed in a PR

src/pipecat/services/sarvam/tts.py
pratham-sarvam: Addressed in a PR

src/pipecat/services/sarvam/tts.py
pratham-sarvam: Our websocket automatically closes in case there is no incoming message for a timeout period (currently 1 min) so in case we want persistent connection we need to send ping message.

src/pipecat/services/sarvam/tts.py
pratham-sarvam: Addressed in a PR

src/pipecat/services/sarvam/tts.py
pratham-sarvam: Addressed in a PR

src/pipecat/services/sarvam/tts.py
pratham-sarvam: Addressed in a PR

src/pipecat/services/sarvam/tts.py
pratham-sarvam: @vinayak-sarvam please help here as well

src/pipecat/services/sarvam/tts.py
pratham-sarvam: Addressed in a PR

src/pipecat/services/sarvam/tts.py
pratham-sarvam: Addressed in a PR

src/pipecat/services/sarvam/tts.py
pratham-sarvam: Addressed in a PR

src/pipecat/services/sarvam/tts.py
pratham-sarvam: This might not be needed. Will remove this after the final changes

src/pipecat/services/sarvam/tts.py
pratham-sarvam: Addressed in a PR

src/pipecat/services/sarvam/tts.py
pratham-sarvam: Which variable are we talking about here exactly? What is the usecase behind each of them?

src/pipecat/services/sarvam/tts.py
pratham-sarvam: Addressed in a PR

src/pipecat/services/sarvam/tts.py
pratham-sarvam: Addressed in a PR
I have added this logic here but somehow the disconnect isn't happening, can you help guide what could be going wrong.

pratham-sarvam:
@markbackman, We have addressed the comments you mentioned, thanks for detailed feedback. Can you review the changes once?
TIA

examples/foundational/07z-interruptible-sarvam.py
```diff
@@ -5,13 +5,15 @@
 #
 
 
+import asyncio
```
markbackman, line 8: Let's keep the current example—you can rename it to 07z-interruptible-sarvam-http.py. Then, add a new one for the websocket service. That would be 07z-interruptible-sarvam.py.

src/pipecat/services/sarvam/tts.py
```diff
@@ -240,3 +261,406 @@ async def run_tts(self, text: str) -> AsyncGenerator[Frame, None]:
...
+
+        Args:
+            api_key: Sarvam API key for authenticating TTS requests.
+            model: Identifier of the Sarvam speech model (default “bulbul:v2”).
+            voice_id: Voice identifier for synthesis (default “anushka”).
+            url: WebSocket URL for connecting to the TTS backend (default production URL).
+            aggregate_sentences: Whether to merge multiple sentences into one audio chunk (default True).
+            sample_rate: Desired sample rate for the output audio in Hz (overrides default if set).
+            params: Optional input parameters to override global configuration.
+            target_language_code: BCP-47 code of the target language (default “hi-IN”).
```
markbackman: Remove the rest of the args from `target_language_code` through `output_audio_bitrate` (keeping **kwargs). These are now defined as `Parameters:` in `InputParams`.

src/pipecat/services/sarvam/tts.py
markbackman: Still unused. Remove.

src/pipecat/services/sarvam/tts.py
```diff
@@ -240,3 +261,406 @@ async def run_tts(self, text: str) -> AsyncGenerator[Frame, None]:
...
+            self._get_websocket(), manager=self.task_manager
+        ):
+            if isinstance(message, str):
+                msg = json.loads(message)
+                if msg.get("type") == "audio":
+                    # Check for interruption before processing audio
+                    await self.stop_ttfb_metrics()
+                    audio = base64.b64decode(msg["data"]["audio"])
+                    frame = TTSAudioRawFrame(audio, self.sample_rate, 1)
+                    logger.debug("pushing audio frames from tts to output")
```
markbackman, line 589: Remove this log line.

src/pipecat/services/sarvam/tts.py
markbackman: Does the keepalive messages accomplish that? You're sending messages every 10 seconds.

src/pipecat/services/sarvam/tts.py
markbackman: `self._cumulative_time = 0`. This is unused here and where it's initialized. You can remove both.

src/pipecat/services/sarvam/tts.py
```diff
@@ -6,23 +6,40 @@
...
 from loguru import logger
 from pydantic import BaseModel, Field
 
 from pipecat.frames.frames import (
+    CancelFrame,
+    EndFrame,
     ErrorFrame,
     Frame,
     StartFrame,
+    StartInterruptionFrame,
```
markbackman, line 26: Unused import. Remove.
```suggestion
```

src/pipecat/services/sarvam/tts.py
```diff
@@ -240,3 +261,406 @@ async def run_tts(self, text: str) -> AsyncGenerator[Frame, None]:
...
+        try:
+            if not self._websocket or self._websocket.state is State.CLOSED:
+                await self._connect()
+
+            try:
+                if not self._started:
+                    await self.start_ttfb_metrics()
+                    yield TTSStartedFrame()
+                    self._started = True
+                    self._cumulative_time = 0
```
```suggestion
```

src/pipecat/services/sarvam/tts.py
markbackman: Remove this log line.
```suggestion
```

src/pipecat/services/sarvam/tts.py
```diff
@@ -240,3 +261,406 @@ async def run_tts(self, text: str) -> AsyncGenerator[Frame, None]:
...
+        self._settings["speech_sample_rate"] = self.sample_rate
+        await self._connect()
+
+    async def stop(self, frame: EndFrame):
+        """Stop the Sarvam TTS service.
+
+        Args:
+            frame: The end frame.
+        """
+        logger.debug("getting stop frame")
```
markbackman: Remove logline.

```suggestion
```

src/pipecat/services/sarvam/tts.py
```diff
@@ -240,3 +261,406 @@ async def run_tts(self, text: str) -> AsyncGenerator[Frame, None]:
...
+        await super().stop(frame)
+        await self._disconnect()
+
+    async def cancel(self, frame: CancelFrame):
+        """Cancel the Sarvam TTS service.
+
+        Args:
+            frame: The cancel frame.
+        """
+        logger.debug("getting cancel frame")
```
markbackman: Remove logline.

```suggestion
```

src/pipecat/services/sarvam/tts.py
```diff
@@ -6,23 +6,40 @@
...
 
+import asyncio
 import base64
-from typing import AsyncGenerator, Optional
+import json
+from typing import Any, AsyncGenerator, Mapping, Optional
 
 import aiohttp
+
+try:
```
markbackman: Move this try/except block to the bottom of the imports.

src/pipecat/services/sarvam/tts.py
```diff
@@ -240,3 +261,406 @@ async def run_tts(self, text: str) -> AsyncGenerator[Frame, None]:
...
+            await self.cancel_task(self._keepalive_task)
+            self._keepalive_task = None
+
+        await self._disconnect_websocket()
+
+    async def _connect_websocket(self):
+        """Establish WebSocket connection to Sarvam API."""
+        try:
+            if self._websocket and self._websocket.state is State.OPEN:
+                logger.debug("Webscoket already connected")
```
markbackman: Remove logline.

```suggestion
```

src/pipecat/services/sarvam/tts.py
```diff
@@ -240,3 +261,406 @@ async def run_tts(self, text: str) -> AsyncGenerator[Frame, None]:
...
+                return
+
+            logger.debug("Connecting to Sarvam TTS Websocket")
```
markbackman: This is a duplicate to the same logline on line 534.

```suggestion
```

src/pipecat/services/sarvam/tts.py
```diff
@@ -240,3 +261,406 @@ async def run_tts(self, text: str) -> AsyncGenerator[Frame, None]:
...
+        if self._websocket:
+            # Send empty text for keepalive
+            msg = {"type": "ping"}
+            await self._websocket.send(json.dumps(msg))
+
+    async def _send_text(self, text: str):
+        """Send text to Neuphonic WebSocket for synthesis."""
+        if self._websocket:
+            msg = {"type": "text", "data": {"text": text}}
+            logger.debug(f"Sending text to websocket: {msg}")
```
markbackman: Remove logline.

```suggestion
```

markbackman:
Looks really close! Just a little more clean up. Also, please add a changelog entry.

Also, don't for get to lint the code! (`uv run pre-commit install` will point you in the right direction; `uv run scripts/fix-ruff.sh` might do some cleanup for you)

src/pipecat/services/sarvam/tts.py
```diff
@@ -97,7 +112,7 @@ def __init__(
         api_key: str,
         voice_id: str = "anushka",
         model: str = "bulbul:v2",
-        aiohttp_session: aiohttp.ClientSession,
+        aiohttp_session: Optional[aiohttp.ClientSession] = None,
```
markbackman: I think Optional is OK, but I think requiring it might be better. If you do make it optional, please move it further down in the args list such that it's with other Optionals.

src/pipecat/services/sarvam/tts.py
```diff
@@ -324,6 +340,7 @@ def __init__(
         model: str = "bulbul:v2",
         voice_id: str = "anushka",
         url: str = "wss://api.sarvam.ai/text-to-speech/ws",
+        aiohttp_session: Optional[aiohttp.ClientSession] = None,
```
markbackman, line 323: The Websocket connection shouldn't have an aiohttp_session arg. It connects via a websocket connection.

src/pipecat/services/sarvam/tts.py
```diff
@@ -240,3 +277,371 @@ async def run_tts(self, text: str) -> AsyncGenerator[Frame, None]:
...
+        await super().cancel(frame)
+        await self._disconnect()
+
+    async def flush_audio(self):
+        """Flush any pending audio synthesis by sending stop command."""
+        if self._websocket:
+            msg = {"type": "flush"}
+            await self._websocket.send(json.dumps(msg))
+
+    async def process_frame(self, frame: Frame, direction: FrameDirection):
```
markbackman, line 457: This should not be needed. It should be handled by the base classes. May I ask why you had to add it?

markbackman: Other services use:
```
        """Push a frame downstream with special handling for stop conditions.

        Args:
            frame: The frame to push.
            direction: The direction to push the frame.
        """
        if isinstance(frame, (TTSStoppedFrame, StartInterruptionFrame)):
            self._started = False
```

src/pipecat/services/sarvam/tts.py
pratham-sarvam: We are inclined to keep it optional as well, since not a lot of other implementations are following this structure of passing session into a Frame processor, so we feel it's better to keep it uniform. Moved down

src/pipecat/services/sarvam/tts.py
pratham-sarvam: This is specifically put in to ensure backward compatibility, to ensure that if people are using our old code which refers to SarvamTTSService as the http implementation as opposed to the new implementation which now refers to websocket implementation. So, we want the old code to still work.

src/pipecat/services/sarvam/tts.py
pratham-sarvam: This is needed for our websocket class to send any pending response. We send this flush signal at the end of LLM response as seen in wordTTSService implementation :
```
        elif isinstance(frame, (LLMFullResponseEndFrame, EndFrame)):
            await self.flush_audio()
```
Do let us know if we should call flush_audio someway else since in TTS service I only see flush audio being called at TTSSpeak Frame received.

src/pipecat/services/sarvam/tts.py
```diff
@@ -97,9 +112,9 @@ def __init__(
         api_key: str,
         voice_id: str = "anushka",
         model: str = "bulbul:v2",
-        aiohttp_session: aiohttp.ClientSession,
         base_url: str = "https://api.sarvam.ai",
         sample_rate: Optional[int] = None,
+        aiohttp_session: Optional[aiohttp.ClientSession] = None,
```
markbackman: Sorry, I should have been more clear. This _must_ be required and passed in. It's up to the application code to pass in a session that has scope for the entire session. Developers can share the session with others that require a session and they can use an async context (e.g. `async with aiohttp.ClientSession()`) to automatically close out the session. If the session isn't properly shared and closed, it will cause the code to hang. Please make this required.

Doing so will match every other service. No services uses aiohttp for REST API interaction with an Optional.

markbackman: Also, after the change, resort the inputs—keep requires and optionals together.

src/pipecat/services/sarvam/tts.py
```diff
@@ -109,24 +124,26 @@ def __init__(
             api_key: Sarvam AI API subscription key.
             voice_id: Speaker voice ID (e.g., "anushka", "meera"). Defaults to "anushka".
             model: TTS model to use ("bulbul:v1" or "bulbul:v2"). Defaults to "bulbul:v2".
-            aiohttp_session: Shared aiohttp session for making requests.
             base_url: Sarvam AI API base URL. Defaults to "https://api.sarvam.ai".
             sample_rate: Audio sample rate in Hz (8000, 16000, 22050, 24000). If None, uses default.
+            aiohttp_session: Optional shared aiohttp session for making requests.
```
markbackman: Update comment to match the change (and positioning).

src/pipecat/services/sarvam/tts.py
```diff
@@ -109,24 +124,26 @@ def __init__(
...
         """
         super().__init__(sample_rate=sample_rate, **kwargs)
 
-        params = params or SarvamTTSService.InputParams()
+        params = params or SarvamHttpTTSService.InputParams()
 
         self._api_key = api_key
         self._base_url = base_url
-        self._session = aiohttp_session
+        self._session = aiohttp_session or aiohttp.ClientSession()
```
markbackman: Required after the change I suggested.

```suggestion
        self._session = aiohttp_session
```

src/pipecat/services/sarvam/tts.py
```diff
@@ -109,24 +124,26 @@ def __init__(
...
+        self._session_owner = aiohttp_session is None
```
markbackman: Remove
```suggestion
```

src/pipecat/services/sarvam/tts.py
```diff
@@ -164,6 +181,26 @@ async def start(self, frame: StartFrame):
         await super().start(frame)
         self._settings["sample_rate"] = self.sample_rate
 
+    async def stop(self, frame: EndFrame):
```
markbackman: Remove `stop` and `cancel`.

src/pipecat/services/sarvam/tts.py
markbackman: Got it. Then, do this—mark it as deprecated and emit a warning that it will be removed in a future version.

src/pipecat/services/sarvam/tts.py
```diff
@@ -240,3 +277,371 @@ async def run_tts(self, text: str) -> AsyncGenerator[Frame, None]:
...
+        **kwargs,
+    ):
+        """Initialize the Sarvam TTS service with voice and transport configuration.
+
+        Args:
+            api_key: Sarvam API key for authenticating TTS requests.
+            model: Identifier of the Sarvam speech model (default "bulbul:v2").
+            voice_id: Voice identifier for synthesis (default "anushka").
+            url: WebSocket URL for connecting to the TTS backend (default production URL).
+            aiohttp_session: Optional shared aiohttp session. To maintain backward compatibility.
```
markbackman, line 336:
```suggestion
            aiohttp_session: Optional shared aiohttp session. To maintain backward compatibility.
            
                .. deprecated:: 0.0.81
                    aiohttp_session is no longer used. This parameter will be removed in a future version.
```

src/pipecat/services/sarvam/tts.py
```diff
@@ -240,3 +277,371 @@ async def run_tts(self, text: str) -> AsyncGenerator[Frame, None]:
...
+        super().__init__(
+            aggregate_sentences=aggregate_sentences,
+            push_text_frames=True,
+            pause_frame_processing=True,
+            push_stop_frames=True,
+            sample_rate=sample_rate,
+            **kwargs,
+        )
+        params = params or SarvamTTSService.InputParams()
+
```
markbackman: Add:

```
        if aiohttp_session is not None:
            warnings.warn(
                "The 'aiohttp_session' parameter is deprecated and will be removed in a future version. "
                DeprecationWarning,
                stacklevel=2,
            )
```

src/pipecat/services/sarvam/tts.py
markbackman: I'll have to try this out before merging to see what's needed.

src/pipecat/services/sarvam/tts.py
markbackman: In most other cases, you need to call:
```
            await self._disconnect()
            await self._connect()
```

to manually disconnect/connect the websocket.

For example, in the ElevenLabsTTSService, we do:
```
    async def _update_settings(self, settings: Mapping[str, Any]):
        """Update service settings and reconnect if voice changed."""
        prev_voice = self._voice_id
        await super()._update_settings(settings)
        if not prev_voice == self._voice_id:
            logger.info(f"Switching TTS voice to: [{self._voice_id}]")
            await self._disconnect()
            await self._connect()
```

pratham-sarvam:
@markbackman Thanks for the prompt review and comments, I have made the suggested changes. Please have a look at them and let me know if we are good to go.
Once again thanks for the feedback and help :)

src/pipecat/services/sarvam/tts.py
pratham-sarvam: Okay thanks. Do let me know what else is required here.

src/pipecat/services/sarvam/tts.py
pratham-sarvam: Sorry, this is an old comment, we addressed that issue, our websocket doesn't need to disconnect to change voices, we can send config message and it works as is. There was some small issue we fixed it (in the example file)

pratham-sarvam:
> Ok! Very close.
> 
> I looked into this:
> 
> ```
>     async def process_frame(self, frame: Frame, direction: FrameDirection):
>         """Process a frame and flush audio if it's the end of a full response."""
>         if isinstance(frame, LLMFullResponseEndFrame):
>             await self.flush_audio()
>         return await super().process_frame(frame, direction)
> ```
> 
> I tried just commenting it out and I don't see any difference in behavior. In what case is this needed? I think it can be removed, but maybe I'm missing a case where it's required.
> 
> Aside from that, everything else looks good. Once we get this final item resolved, we can merge this down.


I tested with commenting it out and the audio from the last segment in TTS gets missed out and it gets spoken in the next turn. This happens when you don't send explicit flush so the websocket doesn't start sending the response for the last segment.

pratham-sarvam:
You can look at this log line : `Generating TTS: [text]` and see that the last segment is sent but is not played back in audio, which is the issue of not flushing at LLM end response.

markbackman:
> You can look at this log line : `Generating TTS: [text]` and see that the last segment is sent but is not played back in audio, which is the issue of not flushing at LLM end response.

Ok, that makes sense. I see what you mean.

That looks good.

I found one last missing item. The `SarvamTTSService` class is missing a push_frame method. You should include:
```
        """Push a frame downstream with special handling for stop conditions.

        Args:
            frame: The frame to push.
            direction: The direction to push the frame.
        """
        if isinstance(frame, (TTSStoppedFrame, StartInterruptionFrame)):
            self._started = False
```

With that, this one is good to go!

pratham-sarvam:
@markbackman We cannot merge this yet as when I was testing interruptions, I saw cases where the TTS just stops giving output and doesn't respond post that. Can we have a look at why that is happening? Is it due to something which is happening in the base classes and we are not handling in our implementation or is it some other bug we are missing? Happy to schedule a short call as well in case that is required.

markbackman:
> @markbackman We cannot merge this yet as when I was testing interruptions, I saw cases where the TTS just stops giving output and doesn't respond post that. Can we have a look at why that is happening? Is it due to something which is happening in the base classes and we are not handling in our implementation or is it some other bug we are missing? Happy to schedule a short call as well in case that is required.

I have not seen this. A few things:
1. Can you rebase on `main`?
2. What python version are you running?
3. What's the repro case? I tested interruptions and did not see this happening.

pratham-sarvam:
> > @markbackman We cannot merge this yet as when I was testing interruptions, I saw cases where the TTS just stops giving output and doesn't respond post that. Can we have a look at why that is happening? Is it due to something which is happening in the base classes and we are not handling in our implementation or is it some other bug we are missing? Happy to schedule a short call as well in case that is required.
> 
> I have not seen this. A few things:
> 
> 1. Can you rebase on `main`?
> 2. What python version are you running?
> 3. What's the repro case? I tested interruptions and did not see this happening.

1. I did now.
2. Python 3.11.5
3. It is the case where I keep interrupting the bot and this is the error I get after some time, I have added a fix can you please have a look. This error doesn't repro after this fix :

 ```
Received an unexpected media stream error while reading the audio.
2025-08-20 09:29:45.093 | DEBUG    | pipecat.transports.network.webrtc_connection:on_iceconnectionstatechange:289 - ICE connection state is closed, connection is closed
2025-08-20 09:29:45.093 | DEBUG    | pipecat.transports.network.webrtc_connection:_handle_new_connection_state:454 - Connection state changed to: closed
2025-08-20 09:29:45.093 | INFO     | pipecat.runner.run:handle_disconnected:221 - Discarding peer connection for pc_id: SmallWebRTCConnection#0
2025-08-20 09:29:45.093 | DEBUG    | pipecat.transports.network.small_webrtc:on_closed:247 - Client connection closed.
2025-08-20 09:29:45.094 | INFO     | __main__:on_client_disconnected:110 - Client disconnected
2025-08-20 09:29:45.094 | DEBUG    | pipecat.pipeline.task:_cancel:492 - Cancelling pipeline task PipelineTask#0
2025-08-20 09:29:45.095 | DEBUG    | pipecat.services.deepgram.stt:_disconnect:239 - Disconnecting from Deepgram
2025-08-20 09:29:46.449 | DEBUG    | pipecat.services.sarvam.tts:_disconnect_websocket:537 - Disconnecting from Sarvam
^CINFO:     Shutting down
INFO:     Waiting for background tasks to complete. (CTRL+C to force quit)
^CINFO:     Finished server process [88216]
ERROR:    Traceback (most recent call last):
  File "/Users/prathamgupta/.pyenv/versions/3.11.5/lib/python3.11/asyncio/runners.py", line 190, in run
    return runner.run(main)
           ^^^^^^^^^^^^^^^^
  File "/Users/prathamgupta/.pyenv/versions/3.11.5/lib/python3.11/asyncio/runners.py", line 118, in run
    return self._loop.run_until_complete(task)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "uvloop/loop.pyx", line 1512, in uvloop.loop.Loop.run_until_complete
  File "uvloop/loop.pyx", line 1505, in uvloop.loop.Loop.run_until_complete
  File "uvloop/loop.pyx", line 1379, in uvloop.loop.Loop.run_forever
  File "uvloop/loop.pyx", line 557, in uvloop.loop.Loop._run
  File "uvloop/loop.pyx", line 476, in uvloop.loop.Loop._on_idle
  File "uvloop/cbhandles.pyx", line 83, in uvloop.loop.Handle._run
  File "uvloop/cbhandles.pyx", line 63, in uvloop.loop.Handle._run
  File "/Users/prathamgupta/Documents/Sarvam/pipecat/venv/lib/python3.11/site-packages/uvicorn/server.py", line 70, in serve
    with self.capture_signals():
  File "/Users/prathamgupta/.pyenv/versions/3.11.5/lib/python3.11/contextlib.py", line 144, in __exit__
    next(self.gen)
  File "/Users/prathamgupta/Documents/Sarvam/pipecat/venv/lib/python3.11/site-packages/uvicorn/server.py", line 331, in capture_signals
    signal.raise_signal(captured_signal)
  File "/Users/prathamgupta/.pyenv/versions/3.11.5/lib/python3.11/asyncio/runners.py", line 157, in _on_sigint
    raise KeyboardInterrupt()
KeyboardInterrupt

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/Users/prathamgupta/Documents/Sarvam/pipecat/venv/lib/python3.11/site-packages/starlette/routing.py", line 701, in lifespan
    await receive()
  File "/Users/prathamgupta/Documents/Sarvam/pipecat/venv/lib/python3.11/site-packages/uvicorn/lifespan/on.py", line 137, in receive
    return await self.receive_queue.get()
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/prathamgupta/.pyenv/versions/3.11.5/lib/python3.11/asyncio/queues.py", line 158, in get
    await getter
asyncio.exceptions.CancelledError

2025-08-20 09:29:51.579 | WARNING  | pipecat.pipeline.task:_print_dangling_tasks:834 - Dangling tasks detected: ['SmallWebRTCOutputTransport#0::_clock_task_handler', 'SmallWebRTCOutputTransport#0::_audio_task_handler']
2025-08-20 09:29:51.579 | DEBUG    | pipecat.pipeline.runner:run:88 - Runner PipelineRunner#0 finished running PipelineTask#0
ERROR:asyncio:an error occurred during closing of asynchronous generator <async_generator object BaseOutputTransport.MediaSender._next_frame.<locals>.without_mixer at 0x163117140>
asyncgen: <async_generator object BaseOutputTransport.MediaSender._next_frame.<locals>.without_mixer at 0x163117140>
Traceback (most recent call last):
  File "/Users/prathamgupta/.pyenv/versions/3.11.5/lib/python3.11/asyncio/runners.py", line 190, in run
    return runner.run(main)
           ^^^^^^^^^^^^^^^^
  File "/Users/prathamgupta/.pyenv/versions/3.11.5/lib/python3.11/asyncio/runners.py", line 118, in run
    return self._loop.run_until_complete(task)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "uvloop/loop.pyx", line 1512, in uvloop.loop.Loop.run_until_complete
  File "uvloop/loop.pyx", line 1505, in uvloop.loop.Loop.run_until_complete
  File "uvloop/loop.pyx", line 1379, in uvloop.loop.Loop.run_forever
  File "uvloop/loop.pyx", line 557, in uvloop.loop.Loop._run
  File "uvloop/loop.pyx", line 476, in uvloop.loop.Loop._on_idle
  File "uvloop/cbhandles.pyx", line 83, in uvloop.loop.Handle._run
  File "uvloop/cbhandles.pyx", line 63, in uvloop.loop.Handle._run
  File "/Users/prathamgupta/Documents/Sarvam/pipecat/venv/lib/python3.11/site-packages/uvicorn/server.py", line 70, in serve
    with self.capture_signals():
  File "/Users/prathamgupta/.pyenv/versions/3.11.5/lib/python3.11/contextlib.py", line 144, in __exit__
    next(self.gen)
  File "/Users/prathamgupta/Documents/Sarvam/pipecat/venv/lib/python3.11/site-packages/uvicorn/server.py", line 331, in capture_signals
    signal.raise_signal(captured_signal)
  File "/Users/prathamgupta/.pyenv/versions/3.11.5/lib/python3.11/asyncio/runners.py", line 157, in _on_sigint
    raise KeyboardInterrupt()
KeyboardInterrupt

During handling of the above exception, another exception occurred:

RuntimeError: aclose(): asynchronous generator is already running
Exception ignored in: <coroutine object BaseOutputTransport.MediaSender._clock_task_handler at 0x133291c40>
Traceback (most recent call last):
  File "/Users/prathamgupta/Documents/Sarvam/pipecat/src/pipecat/transports/base_output.py", line 846, in _clock_task_handler
    timestamp, _, frame = await self._clock_queue.get()
                          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/prathamgupta/Documents/Sarvam/pipecat/src/pipecat/utils/asyncio/watchdog_priority_queue.py", line 84, in get
    get_result = await super().get()
                 ^^^^^^^^^^^^^^^^^^^
  File "/Users/prathamgupta/.pyenv/versions/3.11.5/lib/python3.11/asyncio/queues.py", line 160, in get
    getter.cancel()  # Just in case getter is not done yet.
    ^^^^^^^^^^^^^^^
  File "uvloop/loop.pyx", line 1281, in uvloop.loop.Loop.call_soon
  File "uvloop/loop.pyx", line 669, in uvloop.loop.Loop._call_soon
  File "uvloop/loop.pyx", line 678, in uvloop.loop.Loop._call_soon_handle
  File "uvloop/loop.pyx", line 673, in uvloop.loop.Loop._append_ready_handle
  File "uvloop/loop.pyx", line 705, in uvloop.loop.Loop._check_closed
RuntimeError: Event loop is closed
ERROR:asyncio:Task was destroyed but it is pending!
task: <Task pending name='SmallWebRTCOutputTransport#0::_clock_task_handler' coro=<TaskManager.create_task.<locals>.run_coroutine() done, defined at /Users/prathamgupta/Documents/Sarvam/pipecat/src/pipecat/utils/asyncio/task_manager.py:234> wait_for=<Future cancelled> cb=[TaskManager._task_done_handler()]>
ERROR:asyncio:Task was destroyed but it is pending!
task: <Task pending name='SmallWebRTCOutputTransport#0::_audio_task_handler' coro=<TaskManager.create_task.<locals>.run_coroutine() done, defined at /Users/prathamgupta/Documents/Sarvam/pipecat/src/pipecat/utils/asyncio/task_manager.py:234> wait_for=<Future pending cb=[Task.task_wakeup()]> cb=[TaskManager._task_done_handler()]>
ERROR:asyncio:Task was destroyed but it is pending!
task: <Task pending name='Task-32140' coro=<Queue.get() running at /Users/prathamgupta/.pyenv/versions/3.11.5/lib/python3.11/asyncio/queues.py:160> wait_for=<Future cancelled> cb=[_release_waiter(<Future pendi...ask_wakeup()]>)() at /Users/prathamgupta/.pyenv/versions/3.11.5/lib/python3.11/asyncio/tasks.py:421]>
Exception ignored in: <coroutine object Queue.get at 0x163117e60>
Traceback (most recent call last):
  File "/Users/prathamgupta/.pyenv/versions/3.11.5/lib/python3.11/asyncio/queues.py", line 160, in get
    getter.cancel()  # Just in case getter is not done yet.
    ^^^^^^^^^^^^^^^
  File "uvloop/loop.pyx", line 1281, in uvloop.loop.Loop.call_soon
  File "uvloop/loop.pyx", line 669, in uvloop.loop.Loop._call_soon
  File "uvloop/loop.pyx", line 678, in uvloop.loop.Loop._call_soon_handle
  File "uvloop/loop.pyx", line 673, in uvloop.loop.Loop._append_ready_handle
  File "uvloop/loop.pyx", line 705, in uvloop.loop.Loop._check_closed
RuntimeError: Event loop is closed
```

There are dangling tasks due to race conditions and I have addressed those, please check

pratham-sarvam:
@markbackman @filipi87 Please feel free to add comments to the latest changes we added.

filipi87 (approved):
Looking at the code, it looks great. I’ve also tested both examples, and they are working nicely.

I’ll approve and merge this. I’ll add the changelog in a follow-up PR, so this one doesn’t stay blocked.

Thank you for doing this!

vinayak-sarvam:
Thank you @markbackman and @filipi87 for your reviews, suggestions and helping on this on. We really appreciate it.

pratham-sarvam:
Thanks @markbackman and @filipi87 for prompt feedbacks and testing. It was really helpful as well.

PR #2357, opened 2025-08-04
Added detailed latency logging to UserBotLatencyLogObserver.

filipi87:
Added detailed latency logging to UserBotLatencyLogObserver, capturing average response time between user stop and bot start, as well as minimum and maximum response latency.


aconchillo:
Nice!

PR #2358, opened 2025-08-05
system frames are now queued

aconchillo:

A couple of important changes:

- System frames:  System frames are now queued. Before, system frames could be generated from any task and would not guarantee any order which was causing undesired behavior. Also, it was possible to get into some rare recursion issues because of the way system frames were executed (they were executed in-place, meaning calling `push_frame()` would finish after the system frame traversed all the pipeline). This makes system frames more deterministic.

- PipelineTask: Before CancelFrames didn't need to be waited for because system frames were processed in-place and therefore calling push_frame() would finalize after it traversed all the pipeline. Now, system frames are queued so we need to wait until CancelFrame reaches the end of the pipeline.

### DIFF
--- a/src/pipecat/pipeline/task.py
+++ b/src/pipecat/pipeline/task.py
@@ -500,6 +500,8 @@ class PipelineTask(BasePipelineTask):
             # out-of-band from the main streaming task which is what we want since
             # we want to cancel right away.
             await self._source.push_frame(CancelFrame())
+            # Wait for CancelFrame to make it throught the pipeline.
+            await self._wait_for_pipeline_end()
             # Only cancel the push task. Everything else will be cancelled in run().
             if self._process_push_task:
                 await self._task_manager.cancel_task(self._process_push_task)
@@ -654,7 +656,7 @@ class PipelineTask(BasePipelineTask):
         while running:
             frame = await self._push_queue.get()
             await self._source.queue_frame(frame, FrameDirection.DOWNSTREAM)
-            if isinstance(frame, (EndFrame, StopFrame)):
+            if isinstance(frame, (CancelFrame, EndFrame, StopFrame)):
                 await self._wait_for_pipeline_end()
             running = not isinstance(frame, (CancelFrame, EndFrame, StopFrame))
             cleanup_pipeline = not isinstance(frame, StopFrame)
@@ -727,6 +729,7 @@ class PipelineTask(BasePipelineTask):
                 self._pipeline_end_event.set()
             elif isinstance(frame, CancelFrame):
                 await self._call_event_handler("on_pipeline_cancelled", frame)
+                self._pipeline_end_event.set()
             elif isinstance(frame, HeartbeatFrame):
                 await self._heartbeat_queue.put(frame)
             self._down_queue.task_done()
--- a/src/pipecat/pipeline/task.py
+++ b/src/pipecat/pipeline/task.py
@@ -459,15 +459,20 @@ class PipelineTask(BasePipelineTask):
             # awaiting a task.
             pass
         finally:
-            # It's possibe that we get an asyncio.CancelledError from the
-            # outside, if so we need to make sure everything gets cancelled
-            # properly.
-            if cleanup_pipeline:
-                await self._cancel()
-            await self._cancel_tasks()
-            await self._cleanup(cleanup_pipeline)
-            if self._check_dangling_tasks:
-                self._print_dangling_tasks()
+            # We only cancel things cleanly if we know we are the ones
+            # cancelling. It's possibe that we get an asyncio.CancelledError
+            # from the outside, in which case it is very likely other tasks have
+            # been already cancelled (e.g. when python is shutting down) so we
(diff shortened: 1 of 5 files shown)


aconchillo:
The reason for not having two completely separate tasks at the input level is because if you get a SystemFrame and then, let's say, a TextFrame after that. It's possible that since these are separate tasks and we don't control which one executes first the task with TextFrame gets executed before the SystemFrame.

So, instead of that, we have an input task with a priority queue. The input task will read frames from the priority queue:

- If the frame is a system frame then it will be processed in this same task. So frame ordering is guaranteed for both system frames and other frames. In the previous implementation a SystemFrame would block anything since it was just a function call, this matches this behavior.
- If the frame is not a system frame it will be put into a queue that is processed by another task. This way system frames can be processed quickly and in order.

src/pipecat/pipeline/task.py
```diff
@@ -459,15 +459,16 @@ async def run(self, params: PipelineTaskParams):
...
-            await self._cancel_tasks()
-            await self._cleanup(cleanup_pipeline)
-            if self._check_dangling_tasks:
-                self._print_dangling_tasks()
+            # We only cancel things cleanly if we know we are the ones
+            # cancelling. It's possibe that we get an asyncio.CancelledError
+            # from the outside, in which case it is very likely other tasks have
+            # been already cancelled (e.g. when python is shutting down) so we
+            # can't assume things are being cancelled nicely.
+            if self._cancelled:
```
filipi87, line 467: Were we facing some issue or side effect that led us to add this check here ?

Would it be worth adding a warning in case it hasn't been cancelled by us, so we can track whether this change introduces any unexpected behavior ?

src/pipecat/processors/frame_processor.py
```diff
@@ -622,41 +726,68 @@ def _check_started(self, frame: Frame):
...
-            self.__input_event.clear()
-            self.__input_queue = WatchdogQueue(self.task_manager)
+            self.__input_queue = FrameProcessorQueue(self.task_manager)
             self.__input_frame_task = self.create_task(self.__input_frame_task_handler())
 
     async def __cancel_input_task(self):
-        """Cancel the input processing task."""
+        """Cancel the non-system frame input processing task."""
         if self.__input_frame_task:
             self.__input_queue.cancel()
```
filipi87: @aconchillo , the `self.__input_queue` it is now a `FrameProcessorQueue`, which contains both, the SystemFrames and non-system frame, so I guess in this case we are cancelling everything, right ?

aconchillo: Yes, we cancel both. See:

```

    def cancel(self):
        """Cancel both internal queues.
        This method is used to stop processing and release any pending tasks
        in both the system and main queues. Typically used during shutdown
        or cleanup to prevent further processing of frames.
        """
        self.__main_queue.cancel()
        self.__system_queue.cancel()
```

src/pipecat/pipeline/task.py
aconchillo: Yes, it's mainly an issue on application exit. Let's say you have the following code:

```
asyncio.wait([runner.run(task), other_task], return_when=FIRST_COMPLETED)

# Application exits here
```

If `other_task` finishes first, the application will exit and all Python tasks will start being cancelled. We will be able to capture that in the `asyncio.CancelledError` handler but at that point we can't make any guarantees on what tasks are running because Python cancels the rest of tasks as well.

Before, we were calling `self._cancel()`. `_cancel()` pushes a `CancelFrame`. Before, a `queue_frame(CancelFrame())` would block until it's done, however now, since `CancelFrame` is queued, we need to wait for it. So, inside `_cancel()` you can now see a `_wait_for_pipeline_end()`. However, with the application exit issue mentioned above, `_wait_for_pipeline_end()` might get stuck forever because the tasks that would trigger the needed event have been cancelled by Python.

You might need to read this a few times :-), it might be confusing. But agreed, a warning would be nice. The warning can recommend to call `task.cancel()`.

aconchillo: Added a warning.

src/pipecat/processors/frame_processor.py
filipi87: What I mean is that in the comments, we are mentioning as we were cancelling only the `non-system frame`. Shouldn't we leave it as it was before ? 

`Cancel the input processing task.`

src/pipecat/pipeline/task.py
filipi87: I see, yeah, that makes sense. My only concern is that this change could leave some tasks without being properly canceled, but I think we can use the warning to monitor that and see if we’ll need to add any extra safeguards in the future.

aconchillo: Yes, but that would be an error on the developer. With this warning they should know.

src/pipecat/processors/frame_processor.py
aconchillo: Oh, I see. Sorry, I missed the comment. Yes, let me change that.

PR #2366, opened 2025-08-05
add sigint/sigterm to RunnerArguments

aconchillo:

Currently, the example application don't have a way to know if they should set `handle_sigint` or not. In this PR we add `handle_sigint` and `handle_sigterm` to `RunnerArguments` and then we pass that to `run_bot()`. Note that passing `RunnerArguments` might come in handy if the examples need some values from it.

### DIFF
--- a/src/pipecat/runner/run.py
+++ b/src/pipecat/runner/run.py
@@ -145,6 +145,7 @@ async def _run_telephony_bot(websocket: WebSocket):
 
     # Just pass the WebSocket - let the bot handle parsing
     runner_args = WebSocketRunnerArguments(websocket=websocket)
+    runner_args.handle_sigint = False
 
     await bot_module.bot(runner_args)
 
@@ -222,6 +223,7 @@ def _setup_webrtc_routes(app: FastAPI, esp32_mode: bool = False, host: str = "lo
 
             bot_module = _get_bot_module()
             runner_args = SmallWebRTCRunnerArguments(webrtc_connection=pipecat_connection)
+            runner_args.handle_sigint = False
             background_tasks.add_task(bot_module.bot, runner_args)
 
         answer = pipecat_connection.get_answer()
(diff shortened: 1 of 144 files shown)


markbackman:
Good idea! Mostly LGTM.

One issue. When running Daily via a fastapi server, you need sigint to be False. Update here:
```
def _setup_daily_routes(app: FastAPI):
    """Set up Daily-specific routes."""

    @app.get("/")
    async def start_agent():
        """Launch a Daily bot and redirect to room."""
        print("Starting bot with Daily transport")

        import aiohttp

        from pipecat.runner.daily import configure

            room_url, token = await configure(session)

            # Start the bot in the background with empty body for GET requests
            bot_module = _get_bot_module()
            runner_args = DailyRunnerArguments(room_url=room_url, token=token, body={})
            runner_args.handle_sigint = False # <===  ADD HERE
            asyncio.create_task(bot_module.bot(runner_args))
            return RedirectResponse(room_url)
```

aconchillo:
> Good idea! Mostly LGTM.
> 
> One issue. When running Daily via a fastapi server, you need sigint to be False. Update here:
> 

Ah, yes. Good catch!

src/pipecat/runner/run.py
```diff
@@ -308,6 +310,7 @@ async def _handle_rtvi_request(request: Request):
             # Start the bot in the background with extracted body data
             bot_module = _get_bot_module()
             runner_args = DailyRunnerArguments(room_url=room_url, token=token, body=bot_body)
+            runner_args.handle_sigint = False
```
markbackman, line 314: This isn't the right place. You should add it to below line 268.

aconchillo: Argh. Well, that line as well right?

aconchillo: Done.

markbackman (approved):
LGTM!

PR #2369, opened 2025-08-05
PipelineTask: always try to cancel things

aconchillo:

In a previous commit we only cleanup things if the user run `task.cancel()`. However, if the task finishes cleanly we were not cancelling anything.


aconchillo:
@filipi87 I realized the previous approach is wrong. The issue was that before we had something like this:

```
        finally:
            # It's possibe that we get an asyncio.CancelledError from the
            # outside, if so we need to make sure everything gets cancelled
            # properly.
            if cleanup_pipeline:
                await self._cancel()
            await self._cancel_tasks()
            await self._cleanup(cleanup_pipeline)
            if self._check_dangling_tasks:
                self._print_dangling_tasks()
            self._finished = True
```

The issue was leaving:

```
            if cleanup_pipeline:
                await self._cancel()
```

This is what was causing the freeze because of what I explained in this comment: https://github.com/pipecat-ai/pipecat/pull/2358#discussion_r2255370821

Since we don't do that anymore, it's OK to try to cleanup everything if possible.

PR #2371, opened 2025-08-06
Update docs auto-generation to use uv

markbackman:

Following this guidance: https://github.com/astral-sh/uv/issues/10074

This moves to uv and removes a significant number of mocks. I've selectively mocked packages by download size to make sure docs still build in around 2 minutes. The net of this is an improvement. I wish it was faster and caching was supported on RTD, but 🤷


docs/api/conf.py
```diff
@@ -28,6 +29,7 @@
 
 suppress_warnings = [
     "autodoc.mocked_object",
+    "toc.not_included",
```
markbackman, line 32: Was causing a warning that cause the build to failed. Suppressing works to resolve the error.

docs/api/conf.py
```diff
@@ -43,147 +45,24 @@
...
-    "google",
-    "krisp",
-    "langchain",
-    "lmnt",
-    "noisereduce",
-    "openpipe",
-    "simli",
-    "soundfile",
-    "soniox",
+    # Krisp - has build issues on some platforms
```
markbackman, line 50: We can now use `uv` to install all deps significantly reducing the burden of mocking while still having fast build times.

docs/api/index.rst
```diff
@@ -14,7 +14,7 @@ Quick Links
 * `Join our Community <https://discord.gg/pipecat>`_
 
 .. toctree::
-   :maxdepth: 3
+   :maxdepth: 2
```
markbackman, line 17: A depth of 3 caused too many entries on the top-level pages. 2 is nice ✨

pyproject.toml
```diff
@@ -121,6 +121,14 @@ dev = [
     "python-dotenv~=1.1.1",
 ]
 
+docs = [
```
markbackman, line 124: Required as we're removing docs/api/requirements.txt

src/pipecat/adapters/schemas/direct_function.py
```diff
@@ -44,7 +44,7 @@ class DirectFunction(Protocol):
 
     "Direct" functions' metadata is automatically extracted from their function signature and
     docstrings, allowing them to be used without accompanying function configurations (as
-    `FunctionSchema`s or in provider-specific formats).
+    FunctionSchemas or in provider-specific formats).
```
markbackman, line 47: Backticks don't seem to be well supported.

src/pipecat/services/speechmatics/stt.py
```diff
@@ -74,7 +74,7 @@ class DiarizationFocusMode(str, Enum):
 class AdditionalVocabEntry:
     """Additional vocabulary entry.
 
-    Attributes:
+    Parameters:
```
markbackman, line 77: Fixing errors.

pyproject.toml
filipi87: Yep, makes sense. 👍

PR #2372, opened 2025-08-06
Wider package support for python-dotenv dev dep

markbackman:

This should make it easier to work with. python-dotenv is very stable, so low risk in supporting through to the next major version.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -118,7 +118,7 @@ dev = [
     "ruff~=0.12.1",
     "setuptools~=78.1.1",
     "setuptools_scm~=8.3.1",
-    "python-dotenv~=1.1.1",
+    "python-dotenv>=1.0.1,<2.0.0",
 ]
 
 [tool.setuptools.packages.find]


examples/foundational/28-transcription-processor.py
```diff
@@ -117,9 +118,10 @@ async def run_bot(transport: BaseTransport, runner_args: RunnerArguments):
 
     stt = DeepgramSTTService(api_key=os.getenv("DEEPGRAM_API_KEY"))
 
-    tts = CartesiaTTSService(
-        api_key=os.getenv("CARTESIA_API_KEY"),
-        voice_id="71a7ad14-091c-4e8e-a314-022ece01c121",  # British Reading Lady
+    tts = ElevenLabsTTSService(
```
filipi87: Was including this change intentional? If so, we can probably remove the import from CartesiaTTSService.

markbackman: Oops! No. Fixing.

PR #2375, opened 2025-08-06
Gladia region selection

fabrice404:
## Summary

This PR introduces a new `region` parameter to the Gladia Speech-to-Text (STT) module.

---

### 🌍 Region Selection

- Added a `region` parameter to configure the STT endpoint region.
- **Accepted values**: `eu-west` (default), `us-west`.
- If not explicitly set, the default behavior remains `eu-west` to preserve backward compatibility.


pyproject.toml
```diff
@@ -9,7 +9,7 @@ description = "An open source framework for voice (and multimodal) assistants"
 license = "BSD-2-Clause"
 license-files = ["LICENSE"]
 readme = "README.md"
-requires-python = ">=3.10"
+requires-python = ">=3.10,<=3.13.4"
```
markbackman: Let's not constrain the entire project. Instead, we can add a python requirement only for the specific optional dependency. Can you update to:
```
local-smart-turn = [ "coremltools>=8.0", "transformers", "torch~=2.5.0; python_version<='3.13.4'", "torchaudio~=2.5.0; python_version<='3.13.4'" ]
```

markbackman (commented):
Looks good. Just one small suggestion.

Also, can you add a changelog entry for this?

markbackman (approved):
LGTM! Thank you 🙌

PR #2377, opened 2025-08-06
Fixed an issue in BaseOutputTransport where the loop could consume 10…

filipi87:
Fixed an issue in the `BaseOutputTransport`, mainly reproducible with `FastAPIWebsocketOutputTransport` when the audio mixer was enabled, where the loop could consume 100% CPU by continuously returning without delay, preventing other asyncio tasks (such as cancellation or shutdown signals) from being processed.


src/pipecat/transports/base_output.py
```diff
@@ -658,6 +658,11 @@ async def with_mixer(vad_stop_secs: float) -> AsyncGenerator[Frame, None]:
                             num_channels=self._params.audio_out_channels,
                         )
                         yield frame
+                        # Allow other asyncio tasks to execute by adding a small sleep
+                        # Without this sleep, in task cancellation scenarios, this loop would
+                        # continuously return without any delay, leading to 100% CPU utilization
+                        # and preventing cancel/stop signals from being processed properly
+                        await asyncio.sleep(0)
```
aconchillo, line 665: This is because the transport would not write the frame and therefore not sleeping. For example, in this case:

```
        if self._client.is_closing:
            return

        if not self._client.is_connected:
            # Simulate audio playback with a sleep.
            await self._write_audio_sleep()
            return
```

I think this would fix the issue as well.

```
        if self._client.is_closing or not self._client.is_connected:
            # Simulate audio playback with a sleep.
            await self._write_audio_sleep()
            return
````

Actually, I'm wondering if with this fix we should just do:

```
        if self._client.is_closing or not self._client.is_connected:
            return
```

so people implementing transports don't need to think about adding the sleep

aconchillo: Same for the websocket_server.py

filipi87: I think it's a little different because the `_write_audio_sleep` function sleeps for a specific duration based on the `audio_chunk_size`. For example, if we send 40ms of audio, it sleeps for 40ms.

And I guess all the transports are handling that today.

The additional sleep I added is to prevent CPU spikes across all transports. This is a precaution in case other transports are implemented similarly to the FastAPI transport, where it didn't sleep when closing.

filipi87: >        if self._client.is_closing or not self._client.is_connected:

It would, but I think adding it directly to the BaseOutputTransport works better as a safeguard.

aconchillo: That's right. The `await self._write_audio_sleep()` is always done when we are sending but when we are not connected we don't really need to do it (after your fix). The reason I added it is because we were getting 100% CPU spikes. So, it's exactly the same reason. Note that if you add `await self._write_audio_sleep()` to the `if self._client.is_closing` it would fix the issue as well, so both `self._client.is_closing` and `self._client.is_connected` are the same case.

filipi87: Yep, fixed, I have removed the sleep from both when they are not connected yet.

aconchillo (approved):

aconchillo:
LGTM!

PR #2378, opened 2025-08-06
Add new python-compatiblity workflow to check for dependency compatib…

markbackman:
…ility across supported python versions

This workflow only runs when changing pyproject.toml.

It tests to make sure pyproject.toml is compatible across supported versions.


pyproject.toml
```diff
@@ -34,8 +34,6 @@ dependencies = [
     "resampy~=0.4.3",
     "soxr~=0.5.0",
     "openai>=1.74.0,<2",
-    # Explicit dependency pins for Python 3.11+ compatibility
```
markbackman, line 37: No longer needed due to other changes.

aconchillo (approved):

PR #2380, opened 2025-08-06
Deprecate `LLMMessagesFrame`, `LLMUserResponseAggregator`, and `LLMAssistantResponseAggregator`

kompfner:
`LLMMessagesFrame` doesn't any value over `LLMMessagesUpdateFrame` with `run_llm=True`, when there is a context aggregator in play (which is most cases). When there isn't a context aggregator in the pipeilne, an `OpenAILLMContextFrame` does the trick. Having yet another frame type just adds confusion, and will only make it harder to do some upcoming planned refactoring work.

`LLMUserResponseAggregator` and `LLMAssistantResponseAggregator` are deprecated because they depend on the now-deprecated `LLMMessagesFrame`. Fortunately, they're hardly used.


src/pipecat/frames/frames.py
```diff
@@ -478,6 +478,12 @@ def __str__(self):
 class LLMMessagesFrame(DataFrame):
     """Frame containing LLM messages for chat completion.
 
+    .. deprecated:: 0.0.78
```
kompfner: Note to self: update this (and all the other instances of this in this PR) if for some reason this PR doesn't make it in before 0.0.78.

kompfner:
Note that the mem0 service _does_ create a `LLMMessagesFrame` but hasn't been updated accordingly in this PR. That's because it only creates one in response to receiving one, essentially "passing it on". For that reason, we only need to remove it once we actually _remove_ `LLMMessagesFrame`.

kompfner:
Stay tuned for a corresponding PR to https://github.com/pipecat-ai/pipecat-examples

env.example
```diff
@@ -29,6 +29,9 @@ CARTESIA_API_KEY=...
 DAILY_API_KEY=...
 DAILY_SAMPLE_ROOM_URL=https://...
 
+# Deepgram
```
markbackman, line 32: 🕵️

examples/foundational/07b-interruptible-langchain.py
```diff
@@ -97,8 +100,9 @@ async def run_bot(transport: BaseTransport, runner_args: RunnerArguments):
     )
     lc = LangchainProcessor(history_chain)
 
-    tma_in = LLMUserResponseAggregator()
-    tma_out = LLMAssistantResponseAggregator()
+    context = OpenAILLMContext()
+    tma_in = LLMUserContextAggregator(context=context)
```
markbackman, line 104: Do you know what `tma` stands for?

src/pipecat/processors/aggregators/llm_response.py
```diff
@@ -1006,11 +1011,21 @@ def __init__(
     ):
         """Initialize the user response aggregator.
 
+        .. deprecated:: 0.0.78
```
markbackman: I think you only need to include once in the class level docstring.

examples/foundational/22b-natural-conversation-proposal.py
```diff
@@ -60,24 +59,15 @@
...
     async def process_frame(self, frame: Frame, direction: FrameDirection):
         await super().process_frame(frame, direction)
         # We must not block system frames.
         if isinstance(frame, SystemFrame):
             await self.push_frame(frame, direction)
             return
 
-        # Just treat an LLMMessagesFrame as complete, no matter what.
-        if isinstance(frame, LLMMessagesFrame):
-            await self._notifier.notify()
```
markbackman, line 76: Was the notifier not used?

markbackman (commented):
Looks really good! Thanks for working through this. I'll wait for Aleix to take a peek before approving.

examples/foundational/07b-interruptible-langchain.py
kompfner: Nope

kompfner: But decided not to dig too deep here. Just doing the minimal translation away from deprecated stuff in this PR.

src/pipecat/processors/aggregators/llm_response.py
kompfner: Sounds good. Updating.

examples/foundational/22b-natural-conversation-proposal.py
kompfner: The notifier is used to open the "output gate" that lets LLM output through to the user. 

Within `StatementJudgeContextFilter` the notifier was only used here just in case an `LLMMessagesFrame` was programmatically sent down the pipeline (as opposed to real "organic" human utterances), in which case we were going ahead and automatically treating it as a full statement rather than a partial statement and simply bypassing the "is this a full statement" check.

Distinguishing between "organic" and "synthetic"/programmatic utterances didn't seem to me to be a particularly important edge case to handle, for two reasons:
- programmatic utterances will presumably be complete statements and pass the `StatementJudgeContextFilter` with flying colors, and
- though it is a bit of wasted computation (the deleted lines were an optimization), synthetic utterances will presumably be very rare compared to human ones

kompfner: Also: these lines were only an incomplete/leaky optimization (if you take this example as an illustration of how to implement this kind of filter generically), as it was also still possible to send programmatic _context frames_ (as opposed to `LLMMessagesFrame`s) down the pipeline. (Worth noting: `LLMMessagesAppend/UpdateFrame`s would also result in context frames, not `LLMMessagesFrame`s, so they would also not pass this test)

examples/foundational/07b-interruptible-langchain.py
markbackman: I was just curious. Not blocking anything.

examples/foundational/22b-natural-conversation-proposal.py
markbackman: Makes sense! Thank you 🙇

markbackman (approved):
Ok, approving! I took another look at this one and think the changes make sense.

This is a nice way to not break anyone using Langchain while giving them a path to supporting the new format.

If merged, this will be in today's release, which is 0.0.78, so your deprecation version will be correct.

kompfner:
> Stay tuned for a corresponding PR to https://github.com/pipecat-ai/pipecat-examples

Here it is: https://github.com/pipecat-ai/pipecat-examples/pull/12

kompfner:
> If merged, this will be in today's release, which is 0.0.78, so your deprecation version will be correct.

Shoot, missed the release by accident. Bumping to 0.0.79.

PR #2382, opened 2025-08-07
log: warning -> trace for elevenlabs tts unavailable context

vipyne:

see inline comments.

### DIFF
--- a/src/pipecat/services/elevenlabs/tts.py
+++ b/src/pipecat/services/elevenlabs/tts.py
@@ -546,7 +546,6 @@ class ElevenLabsTTSService(AudioContextWordTTSService):
                 continue
 
             # Check if this message belongs to the current context.
-            # This should never happen, so warn about it.
             if not self.audio_context_available(received_ctx_id):
                 if self._context_id == received_ctx_id:
                     logger.debug(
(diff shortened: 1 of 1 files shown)


src/pipecat/services/elevenlabs/tts.py
```diff
@@ -546,15 +546,17 @@ async def _receive_messages(self):
...
                     logger.debug(
                         f"Received a delayed message, recreating the context: {self._context_id}"
                     )
                     await self.create_audio_context(self._context_id)
                 else:
-                    logger.warning(f"Ignoring message from unavailable context: {received_ctx_id}")
+                    # This can happen if a message is received _after_ we have closed a context
+                    # due to user interruption but _before_ the `isFinal` message for the context
+                    # is received.
+                    logger.trace(f"Ignoring message from unavailable context: {received_ctx_id}")
```
markbackman: This change seems fine. The log being a warning is probably more alarming than it needs to be. For the case you mentioned, it would have no effect. Perhaps a debug log as this could be useful for debugging future issues. For example, the warning helped us find an issue for a team building a non-conversational AI use case with this service.

```suggestion
                    logger.debug(f"Ignoring message from unavailable context: {received_ctx_id}")
```

PR #2383, opened 2025-08-07
properly handle concurrent.futures.CancelledError

aconchillo:

When using futures we need to make sure we handle `concurrent.futures.CancelledError`. This would get triggered if the asyncio task running the future is cancelled with, but note that this is different than `asyncio.CancelledError` (but equivalent).


markbackman (approved):
LGTM

Issue #2385, opened 2025-08-07
ParallelPipelineSink error: '__process_queue' attribute missing after recent FrameProcessor changes


jamsea:
## Bug Description

After the recent changes to how FrameProcessor handles system frames (commit a5ea6e16), running examples that use `ParallelPipeline` results in an `AttributeError`.

## Error Message

```
ParallelPipelineSink#0::__input_frame_task_handler: unexpected exception: 'ParallelPipelineSink' object has no attribute '_FrameProcessor__process_queue'
```

## Root Cause

The issue stems from a race condition introduced in commit a5ea6e16 "FrameProcessor: system frames are now queued". 

In the new implementation:
1. `__input_frame_task_handler` tries to access `self.__process_queue` (line 786 in frame_processor.py)
2. However, `__process_queue` is only created when `__create_process_task()` is called
3. `__create_process_task()` is only called when a `StartFrame` is received (in the `__start` method)
4. But the input task is created immediately in `setup()` and can receive frames before `StartFrame` is processed

This causes the `__input_frame_task_handler` to try to access a non-existent `__process_queue` attribute.

## Steps to Reproduce

1. Set up environment with Python 3.12 and recent pipecat version
2. Run any example that uses `ParallelPipeline`, such as:
   ```bash
   uv run examples/foundational/15-switch-voices.py
   ```
3. The error occurs when `ParallelPipelineSink` processes frames before receiving a `StartFrame`

## Environment

- **Python Version**: 3.12.9
- **Pipecat Commit**: ad1841b739152da00dfa502f34799eaa09df6c96
- **Operating System**: macOS
- **Affected File**: `src/pipecat/processors/frame_processor.py`
- **Problem Introduced**: Commit a5ea6e16 (Aug 4, 2025)

## Affected Code

In `src/pipecat/processors/frame_processor.py`, line 786:
```python
async def __input_frame_task_handler(self):
    while True:
        (frame, direction, callback) = await self.__input_queue.get()

        if isinstance(frame, SystemFrame):
            await self.__process_frame(frame, direction, callback)
        else:
            await self.__process_queue.put((frame, direction, callback))  # ← This line fails
```

The `__process_queue` is only created in `__create_process_task()` which is called in `__start()` when a `StartFrame` is received.

## Proposed Fix

The `__process_queue` should be created when the input task is created, not just when `StartFrame` is received. This ensures that non-system frames can be queued even before the processor is formally "started".

Possible fix in `__create_input_task()`:
[...]

PR #2386, opened 2025-08-07
Only push the StartFrame when all parallel pipelines have processed it

filipi87:
Only push the StartFrame when all parallel pipelines have processed it.

Fixes this [issue](https://github.com/pipecat-ai/pipecat/issues/2385).

### DIFF
--- a/src/pipecat/processors/frame_processor.py
+++ b/src/pipecat/processors/frame_processor.py
@@ -256,6 +256,7 @@ class FrameProcessor(BaseObject):
         self.__should_block_frames = False
         self.__process_event = None
         self.__process_frame_task: Optional[asyncio.Task] = None
+        self.__process_queue = None
 
     @property
     def id(self) -> int:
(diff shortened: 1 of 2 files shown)


src/pipecat/processors/frame_processor.py
```diff
@@ -783,6 +784,12 @@ async def __input_frame_task_handler(self):
             if isinstance(frame, SystemFrame):
                 await self.__process_frame(frame, direction, callback)
             else:
+                if self.__process_queue is None:
+                    logger.warning(
+                        f"{self}: __process_queue is None when processing frame {frame.name}"
+                    )
+                    logger.warning(f"{self}: has frame processor started {self.__started}")
+                    continue
```
filipi87: This is just a safeguard, we should not see this warning.

aconchillo: I don't think it is ever possible to hit this. If we receive a `StartInterruptionFrame`, `self.__process_frame()` will be executed which will cancel the process task and will recreate the queue.

I would remove this code. I'm not a fan of adding safeguards if we know for sure they can never happen. Plus, it pollutes the code :sweat_smile: .

aconchillo (approved):

src/pipecat/processors/frame_processor.py
filipi87: We were hitting that code before the change to the ParallelPipeline. So I thought about leave it there, in case there is some other edge case we haven’t fixed yet.

aconchillo: As we discussed privately, raising a `RuntimeError` is probably the way to go. Thank you! :bow:

PR #2387, opened 2025-08-07
Retry chat completions for OpenAILLMService and its subclasses

markbackman:

### DIFF
--- a/src/pipecat/services/openai/base_llm.py
+++ b/src/pipecat/services/openai/base_llm.py
@@ -93,7 +93,7 @@ class BaseOpenAILLMService(LLMService):
         project=None,
         default_headers: Optional[Mapping[str, str]] = None,
         params: Optional[InputParams] = None,
-        timeout: Optional[float] = 5.0,
+        retry_timeout_secs: Optional[float] = 5.0,
         retry_on_timeout: Optional[bool] = False,
         **kwargs,
     ):
@@ -107,7 +107,7 @@ class BaseOpenAILLMService(LLMService):
             project: OpenAI project ID.
             default_headers: Additional HTTP headers to include in requests.
             params: Input parameters for model configuration and behavior.
-            timeout: Request timeout in seconds. Defaults to 5.0 seconds.
+            retry_timeout_secs: Request timeout in seconds. Defaults to 5.0 seconds.
             retry_on_timeout: Whether to retry the request once if it times out.
             **kwargs: Additional arguments passed to the parent LLMService.
         """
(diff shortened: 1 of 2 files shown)


src/pipecat/services/cerebras/llm.py
```diff
@@ -55,20 +54,13 @@ def create_client(self, api_key=None, base_url=None, **kwargs):
         logger.debug(f"Creating Cerebras client with api {base_url}")
         return super().create_client(api_key, base_url, **kwargs)
 
-    async def get_chat_completions(
+    def build_chat_completion_params(
```
filipi87, line 57: This makes things a lot better. Nice improvement.

src/pipecat/services/openai/base_llm.py
```diff
@@ -91,6 +93,8 @@ def __init__(
         project=None,
         default_headers: Optional[Mapping[str, str]] = None,
         params: Optional[InputParams] = None,
+        timeout: Optional[float] = 5.0,
```
filipi87: nit: In some services and processors in our code, we describe the timeout in a more verbose way, but in others, it’s just timeout. 
I think I’d still rather use the verbose way, so something like `retry_timeout_secs`. What do you think ?

src/pipecat/services/openai/base_llm.py
```diff
@@ -184,6 +192,36 @@ async def get_chat_completions(
...
+        params = self.build_chat_completion_params(context, messages)
+
+        if self._retry_on_timeout:
+            try:
+                chunks = await asyncio.wait_for(
+                    self._client.chat.completions.create(**params), timeout=self._timeout
+                )
+                return chunks
+            except (APITimeoutError, asyncio.TimeoutError):
+                # Retry, this time without a timeout so we get a response
```
filipi87, line 204: Yep, I think it makes sense. 👍

filipi87: Shouldn't we add a debug log to understand how often this happens ?

filipi87 (approved):
Just a couple of nit suggestions, but it looks good. 🚀

src/pipecat/services/openai/base_llm.py
markbackman: Good idea. Updating the name.

src/pipecat/services/openai/base_llm.py
markbackman: Yes, makes sense. I'll make it DEBUG for now, so we can observe in logs.

PR #2388, opened 2025-08-07
14v adding OpenAI stt tts llm functioncalling

vr000m:

### DIFF
--- a/examples/foundational/14v-function-calling-openai.py
+++ b/examples/foundational/14v-function-calling-openai.py
@@ -19,10 +19,10 @@ from pipecat.pipeline.task import PipelineParams, PipelineTask
 from pipecat.processors.aggregators.openai_llm_context import OpenAILLMContext
 from pipecat.runner.types import RunnerArguments
 from pipecat.runner.utils import create_transport
-from pipecat.services.openai.stt import OpenAISTTService
-from pipecat.services.openai.tts import OpenAITTSService
 from pipecat.services.llm_service import FunctionCallParams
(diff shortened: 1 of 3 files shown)


examples/foundational/14v-function-calling-openai.py
```diff
@@ -0,0 +1,175 @@
...
+
+    # voice choices: ash, ballad, or any other voice available in the OpenAI TTS API
+    # see https://www.openai.fm/
+    tts = OpenAITTSService(api_key=os.getenv("OPENAI_API_KEY"), 
+        voice="ballad",
+        instructions="Please speak clearly and at a moderate pace."
+    )
+
+    # model choices: gpt-4o, gpt-4.1
+    llm = OpenAILLMService(api_key=os.getenv("OPENAI_API_KEY"), model="gpt-4.1")
```
markbackman: For future maintenance reasons, we've opted to remove the model from these examples and just use the default.

```suggestion
    llm = OpenAILLMService(api_key=os.getenv("OPENAI_API_KEY"))
```

markbackman (approved):
LGTM. Just a tiny comment.

Also, install the pre-commit linter:
```
uv run pre-commit install
```

or, if you're not running uv:
```
python pre-commit install
```

You can also just fix using the fix-ruff.sh script in `/scripts`

examples/foundational/14v-function-calling-openai.py
vr000m: I had 4o in the original commit and then changed to 4.1. but makes sense to remove it and use defaults!

PR #2392, opened 2025-08-07
deprecate TTSService.say() method

aconchillo:

With this PR we deprecate `TTSService.say()` and we force to push a `TTSSpeakFrame` instead. Calling functions directly is a discouraged pattern in Pipecat because, for example, it might cause issues with frame ordering.

### DIFF
--- a/src/pipecat/services/tts_service.py
+++ b/src/pipecat/services/tts_service.py
@@ -269,9 +269,20 @@ class TTSService(AIService):
     async def say(self, text: str):
         """Immediately speak the provided text.
 
+        .. deprecated:: 0.0.79
+            Push a `TTSSpeakFrame` instead to ensure frame ordering is maintained.
+
         Args:
             text: The text to speak.
         """
+        import warnings
+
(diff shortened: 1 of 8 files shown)


src/pipecat/services/tts_service.py
```diff
@@ -272,6 +272,14 @@ async def say(self, text: str):
         Args:
             text: The text to speak.
         """
+        import warnings
```
markbackman, line 278: Can you add a deprecation directive to the docstring?
```
"""Immediately speak the provided text.

.. deprecated:: 0.0.79
    Push a `TTSSpeakFrame` instead to ensure frame ordering is maintained.

Args:
    text: The text to speak.
"""
```

This will get added to docs.

markbackman (approved):
LGTM. Just a note about updating the say() method's docstring.

src/pipecat/services/tts_service.py
aconchillo: done!

PR #2395, opened 2025-08-07
examples(foundational): move 15/15a logic into its own processor

aconchillo:

Update 15/15a examples to show how to enclose complex logic into a single reusable processor.


examples/foundational/15-switch-voices.py
```diff
@@ -10,6 +10,7 @@
 from dotenv import load_dotenv
 from loguru import logger
 from openai.types.chat import ChatCompletionToolParam
+from websockets.frames import Frame
```
markbackman: Looks like a typo. Should be:

```suggestion
from pipecat.frames.frames import Frame
```

aconchillo: I'd have swore I fixed that. Done!

examples/foundational/15-switch-voices.py
```diff
@@ -165,7 +171,7 @@ async def on_client_connected(transport, client):
         messages.append(
             {
                 "role": "system",
-                "content": f"Please introduce yourself to the user and let them know the voices you can do. Your initial responses should be as if you were a {current_voice}.",
+                "content": f"Please introduce yourself to the user and let them know the voices you can do. Your initial responses should be as if you were a {switch_voices.current_voice}.",
```
markbackman: Should be `tts`

```suggestion
                "content": f"Please introduce yourself to the user and let them know the voices you can do. Your initial responses should be as if you were a {tts.current_voice}.",
```

markbackman (approved):
Very nice! Just a small few things, but very nice.

The API for:
```
        super().__init__(
            # News Lady voice
            [FunctionFilter(self.news_lady_filter), news_lady],
            # British Reading Lady voice
            [FunctionFilter(self.british_lady_filter), british_lady],
            # Barbershop Man voice
            [FunctionFilter(self.barbershop_man_filter), barbershop_man],
        )
```

Isn't pretty but it makes sense. We can just document it better.

examples/foundational/15-switch-voices.py
aconchillo: Fixed! I didn't save!

PR #2397, opened 2025-08-08
feat: add minor tweaks to GladiaSTTService

fabrice404:

### DIFF
--- a/src/pipecat/services/gladia/stt.py
+++ b/src/pipecat/services/gladia/stt.py
@@ -488,11 +488,14 @@ class GladiaSTTService(STTService):
 
     async def _setup_gladia(self, settings: Dict[str, Any]):
         async with aiohttp.ClientSession() as session:
+            params = {}
+            if self._region:
+                params["region"] = self._region
             async with session.post(
                 self._url,
                 headers={"X-Gladia-Key": self._api_key},
                 json=settings,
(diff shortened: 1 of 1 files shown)


src/pipecat/services/gladia/stt.py
```diff
@@ -204,7 +204,7 @@ def __init__(
         self,
         *,
         api_key: str,
-        region: Optional[Literal["us-west", "eu-west"]] = "eu-west",
+        region: Literal["us-west", "eu-west"] | None,
```
markbackman: What are the desired states for this arg? Do you want a default value? Or, if the region is None, does Gladia already have a built-in default? If so, maybe you want:
```
region: Optional[Literal["us-west", "eu-west"]] = None,
```

markbackman: Sorry, revisiting this, I think we want:
```
region: Literal["us-west", "eu-west"] | None = None
```

It provides clear intent about the default value.

markbackman (commented):
Change looks good. I think we just want to set the default value for `region` to `None`.

src/pipecat/services/gladia/stt.py
fabrice404: Yes, that's the intent, no region by default. I pushed the fix

markbackman (approved):
LGTM!

PR #2402, opened 2025-08-08
Add voicemail detection

markbackman:

This PR adds a new voicemail_detection module to Pipecat, making it easier to add support for detecting voicemail for outbound calling use cases.

The module provides two pipeline components:

- `VoicemailDetector.detector()`: A ParallelPipeline that classifies calls as CONVERSATION or VOICEMAIL based on user speech input
- `VoicemailDetector.buffer()`: Buffers TTS output until classification decision is made

Usage:
```python
# A callback provided to provide the response to the voicemail detector
async def handle_voicemail(processor):
    await processor.push_frame(TTSSpeakFrame("Hi, please leave a message after the beep."))

# Create an instance of the voice detector
# Optional `system_prompt` arg to overwrite the prompt
# Optional `voicemail_response_delay` to modify the wait time for the callback firing
detector = VoicemailDetector(
    llm=classification_llm,
    on_voicemail_detected=handle_voicemail
)

# Add instances to your pipeline
pipeline = Pipeline([
    transport.input(),
    stt,
    detector.detector(),      # Classification
    context_aggregator.user(),
    llm,
    tts, 
    detector.buffer(),        # TTS buffering
    transport.output(),
])
```

The logic is a state machine, which controls inputs to ensure only either the conversation or voicemail logic are executed.
voicemail_detector_logic

---

### Considerations:
**Fast Classification for Real-time Conversations**
The system makes classification decisions on the first available input (single word, phrase, or multiple sentences) to ensure timely responses for live conversations. Once classified, gates close the appropriate processing branches.

**Selective Frame Flow Through Closed Gates**
Even when gates are closed, certain critical frames are allowed through:
- `UserStartedSpeakingFrame`/`UserStoppedSpeakingFrame` → Enable voicemail callback timing
- `BotInterruptionFrame` → Allow classification processor to trigger interruptions  
- `StartInterruptionFrame`/`StopInterruptionFrame` → Enable pipeline clearing
- End/Cancel frames → Ensure proper pipeline shutdown

**TTS Buffering Strategy**
- **Conversation detected**: Buffer is flushed, allowing immediate bot response
- **Voicemail detected**: Buffer is cleared since conversation responses aren't appropriate for voicemail
- This ensures conversation responses are ready immediately when needed

**Speech-based Timing Control**
Voicemail callback timing uses `UserStoppedSpeakingFrame` events rather than fixed delays, allowing the system to naturally wait for voicemail greetings to complete before responding.


CHANGELOG.md
```diff
@@ -9,6 +9,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- Added `pipecat.utils.voicemail`, a module for detecting voicemail vs. live
```
markbackman: We need to move this module to a different location. I'm not sure where the best is though.

scripts/evals/run-release-evals.py
```diff
@@ -189,8 +203,15 @@ async def main(args: argparse.Namespace):
...
-        await runner.run_eval(test, prompt, eval)
+    # Parse test config: (test, prompt, eval) or (test, prompt, eval, user_speaks_first)
+    for test_config in TESTS:
+        if len(test_config) == 3:
+            test, prompt, eval = test_config
+            user_speaks_first = False
+        else:
+            test, prompt, eval, user_speaks_first = test_config
+
+        await runner.run_eval(test, prompt, eval, user_speaks_first)
```
aconchillo, line 221: I think it's better to add `USER_SPEAKS_FIRST`, `BOT_SPEAKS_FIRST` in every test instead of checking the length of the tuple. It's more verbose, but clearer.

src/pipecat/utils/voicemail/voicemail_detector.py
```diff
@@ -0,0 +1,695 @@
...
+                BotInterruptionFrame,
+                EndTaskFrame,
+                EndFrame,
+                CancelTaskFrame,
+                CancelFrame,
+                UserStartedSpeakingFrame,
+                UserStoppedSpeakingFrame,
+                StartInterruptionFrame,
+                StopInterruptionFrame,
+            ),
```
aconchillo: Not for this PR. But I'm realizing we should make this easier. New people coming in, might think this is hard because you need to know all the existing frames. We also have `StopFrame` and `StopTaskFrame`.

src/pipecat/utils/voicemail/voicemail_detector.py
```diff
@@ -0,0 +1,695 @@
...
+                self._gate_opened = False
+        except asyncio.CancelledError:
+            logger.debug(f"{self}: Gate task was cancelled")
+            raise
+        except Exception as e:
+            logger.exception(f"{self}: Error in gate task: {e}")
+            raise
+
+
+class ConversationGate(FrameProcessor):
```
aconchillo: `ClassifierGate` and `ConversationGate` seem the same. Probably we just need one implementation?

src/pipecat/utils/voicemail/voicemail_detector.py
```diff
@@ -0,0 +1,695 @@
...
+        self._context_aggregator = llm.create_context_aggregator(self._context)
+
+        # Create notification system for coordinating between components
+        self._gate_notifier = EventNotifier()  # Signals classification completion
+        self._conversation_notifier = EventNotifier()  # Signals conversation detected
+        self._voicemail_notifier = EventNotifier()  # Signals voicemail detected
+
+        # Create the processor components
+        self._classifier_gate = ClassifierGate(self._gate_notifier)
+        self._conversation_gate = ConversationGate(self._voicemail_notifier)
```
aconchillo: Following my other comment. It feels this should be:

```
        self._classifier_gate = VoicemailGate(self._gate_notifier)
        self._conversation_gate = VoicemailGate(self._voicemail_notifier)
```

for example

vipyne:
I think we could call it something more generic, just `detector` instead of `voicemailDetector`.  
then `detector.classifier()` after STT and `detector.buffer()` after TTS.*

I could see this classifying conversations into more than just human conversation / voicemail... but also voicemail/human/IVR/adult/minor/etc.

*I sort of wish there was a little more symmetry between `detector.{classifier,buffer}` ala `transport.{intput,output}` and `context_aggregator.{assistent,user}`.  even `voicemail.{detector,buffer}` doesn't feel quite right.

scripts/evals/run-release-evals.py
markbackman: Good point.

src/pipecat/utils/voicemail/voicemail_detector.py
markbackman: Yes, I can create a base class and then just have the `ClassifierGate` and `ConversationGate` pass the notifiers to the base class.

markbackman: This keeps the state management a bit simpler, I think.

src/pipecat/utils/voicemail/voicemail_detector.py
markbackman: I ended up keeping the subclasses, so this isn't changed.

markbackman:
voicemail is just the instance name in the example. It could be anything, so I'm not too worried about that naming.

> I could see this classifying conversations into more than just human conversation / voicemail... but also voicemail/human/IVR/adult/minor/etc.

For now, I'm focused on voicemail. I'm not sure what additional built-in types we'll support. The state machine logic in VoicemailDetector is very specific to this use case. So, I could imagine adding more built-in modules like this, which would have other specialized state machines for the use case. It's unclear right now whether this will be reusable. If it is, we can modify the naming.

src/pipecat/utils/voicemail/voicemail_detector.py
aconchillo: Yes, it's fine.

kompfner:
Reading through the diagram, makes a lot of sense!

High level question: would it make sense to buffer _before_ sending to the LLM while deciding if the incoming audio is a voicemail message? (Or at least have the option to do so?) Not sure how price sensitive some customers are but it seems like by avoiding running LLM + TTS while we're classifying, it could net some savings.

I assume the decision to buffer afterwards is to minimize latency if the classifier comes back with "it's a real person—go!".

markbackman:
> Reading through the diagram, makes a lot of sense!
> 
> High level question: would it make sense to buffer _before_ sending to the LLM while deciding if the incoming audio is a voicemail message? (Or at least have the option to do so?) Not sure how price sensitive some customers are but it seems like by avoiding running LLM + TTS while we're classifying, it could net some savings.
> 
> I assume the decision to buffer afterwards is to minimize latency if the classifier comes back with "it's a real person—go!".

Buffering TTS generation minimizes latency, which can be a real win. For outbound calling, if a human answers, this will speed up the response time by 200-500ms, which is helpful.

Also, going with this pattern, makes the developer experience better. You only have to pass the classifier LLM to the VoicemailDetector. If we opted to gate after the conversation LLM, we'd have to move more things into the VoicemailDetector which makes using it slightly more difficult to use. But worse, the shape of the pipeline would change as well. We want the Pipeline to look "normal", which this achieves.

Pipeline for this design:
```
        [
            transport.input(),
            stt,
            voicemail.detector(),  # Voicemail detection — between STT and User context aggregator
            context_aggregator.user(),
            llm,
            tts,
            voicemail.gate(),  # TTS gating — Immediately after the TTS service
            transport.output(),
        ]
```

Pipeline if we gated after LLM:
```
        [
            transport.input(),
            stt,
            voicemail,
            tts,
            transport.output(),
        ]
```

For all those reasons, we went with this approach.

src/pipecat/extensions/voicemail/voicemail_detector.py
```diff
@@ -0,0 +1,660 @@
...
+    async def process_frame(self, frame: Frame, direction: FrameDirection):
+        """Process frames and control gate state based on notifier signals.
+
+        Args:
+            frame: The frame to process.
+            direction: The direction of frame flow in the pipeline.
+        """
+        await super().process_frame(frame, direction)
+
+        # Gate logic: open gate allows all frames, closed gate filters frames
```
kompfner, line 97: Do we want this logic to be essentially match `FunctionFilter`'s, of allowing through only `SystemFrame`s when the gate is closed?

One thought: if we don't want to duplicate `FunctionFilter`'s logic, we could have `NotifierGate` just subclass `FunctionFilter`, initializing it with a function that just reads a bool set by the notifier.

src/pipecat/extensions/voicemail/voicemail_detector.py
```diff
@@ -0,0 +1,660 @@
...
+)
+from pipecat.pipeline.parallel_pipeline import ParallelPipeline
+from pipecat.processors.aggregators.openai_llm_context import OpenAILLMContext
+from pipecat.processors.frame_processor import FrameDirection, FrameProcessor, FrameProcessorSetup
+from pipecat.services.llm_service import LLMService
+from pipecat.sync.base_notifier import BaseNotifier
+from pipecat.sync.event_notifier import EventNotifier
+
+
+class NotifierGate(FrameProcessor):
```
kompfner, line 47: This might be generic enough to be more broadly useful—maybe it can go under the `filters` folder, as `NotifierFilter`?

kompfner:
> Buffering TTS generation minimizes latency, which can be a real win. For outbound calling, if a human answers, this will speed up the response time by 200-500ms, which is helpful.

This makes a lot of sense. Probably worth the extra little spending to have responses queued up and ready to go.

> Pipeline if we gated after LLM

To be clear, what I was asking about was gating _before_ the LLM, with the following pipeline shape:

```python
        [
            transport.input(),
            stt,
            voicemail.detector(),  # Voicemail detection *and* gating, all in one
            context_aggregator.user(),
            llm,
            tts,
            transport.output(),
        ]
```

But regardless: my hunch is you're right that response latency probably trumps a little extra cost of getting those buffered responses ready (and a little extra code complexity, both on our end as well as the customers')

src/pipecat/extensions/voicemail/voicemail_detector.py
```diff
@@ -0,0 +1,660 @@
...
+        """Initialize the voicemail detector with classification and buffering components.
+
+        Args:
+            llm: LLM service used for voicemail vs conversation classification.
+                Should be fast and reliable for real-time classification.
+            voicemail_response_delay: Delay in seconds after user stops speaking
+                before triggering the voicemail event handler. This allows voicemail
+                responses to be played back after a short delay to ensure the response
+                occurs during the voicemail recording. Default is 2.0 seconds.
+            system_prompt: Optional custom system prompt for classification. If None,
```
kompfner: Should we call this `custom_system_prompt` to make extra clear that the customer really doesn't need to provide this (and that they'd in fact be taking on an extra onus of ensuring that the prompt contents match expectations)

src/pipecat/extensions/voicemail/voicemail_detector.py
```diff
@@ -0,0 +1,660 @@
...
+                self._classification_processor,
+                self._context_aggregator.assistant(),
+            ],
+        )
+
+        # Register the voicemail detected event after super().__init__()
+        self._register_event_handler("on_voicemail_detected")
+
+    def _validate_prompt(self, prompt: str) -> None:
+        """Validate custom prompt contains required response format instructions.
```
kompfner, line 655: When working on AWS Nova Sonic, I also ran into a scenario where I needed the developer to provide a specific snippet in their prompt. I ended up vending a constant with that snippet.

Wonder if it'd be useful to vend something like:
```python
CLASSIFIER_RESPONSE_INSTRUCTION="Respond with ONLY \"CONVERSATION\" if a person answered, or \"VOICEMAIL\" if it's voicemail/recording."
```

And then in this function you could add to the `logger.warning`, pointing the developer to append `VoicemailDetector.CLASSIFIER_RESPONSE_INSTRUCTION` to their system prompt.

src/pipecat/extensions/voicemail/voicemail_detector.py
markbackman: Good idea!

src/pipecat/extensions/voicemail/voicemail_detector.py
markbackman: Nice suggestion. I'll add this.

aconchillo:
LGTM. :clap:

PR #2403, opened 2025-08-08
FrameProcessor: fix system frame higher priorty and use a PriortyQueue

aconchillo:

This PR fixes an issue that's causing system frames to not be considered higher priority. The problem is that to insert an item to the queue we were doing:

```
await self.__process_queue.put((frame, direction, callback))
```

But then, when we check to see if the inserted frame has higher priority we do:

```
async def put(self, item: Any):
    if isinstance(item, SystemFrame):
        await self.__system_queue.put(item)
     else:
        await self.__main_queue.put(item)
```

Note that `item` is a tuple, so `isinstance_item, SystemFrame)` is never satisfied.

To fix this we would just need to check the proper item in the tuple but I went a bit further and simplified the whole thing to use a PriorityQueue instead.

### DIFF
--- a/README.md
+++ b/README.md
@@ -114,7 +114,8 @@ You can get started with Pipecat running on your local machine, then move your a
 
 ### Prerequisites
 
-**Python Version:** 3.10+
+**Minimum Python Version:** 3.10
+**Recommended Python Version:** 3.11-3.12
 
 ### Setup Steps
 
--- a/src/pipecat/processors/frame_processor.py
+++ b/src/pipecat/processors/frame_processor.py
@@ -14,7 +14,7 @@ management, and frame flow control mechanisms.
 import asyncio
 from dataclasses import dataclass
 from enum import Enum
-from typing import Any, Awaitable, Callable, Coroutine, List, Optional, Sequence
+from typing import Any, Awaitable, Callable, Coroutine, List, Optional, Sequence, Tuple
 
 from loguru import logger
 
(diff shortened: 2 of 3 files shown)


README.md
```diff
@@ -114,7 +114,8 @@ You can get started with Pipecat running on your local machine, then move your a
 
 ### Prerequisites
 
-**Python Version:** 3.10+
+**Minimum Python Version:** 3.10
+**Recommended Python Version:** 3.11+
```
markbackman: 3.13 is missing support for things like PyTorch and has more package compatibility challenges. For that reason, maybe we should recommend 3.11 and 3.12.

```suggestion
**Recommended Python Versions:** 3.11-3.12
```

aconchillo: Done!

src/pipecat/processors/frame_processor.py
```diff
@@ -581,7 +558,7 @@ async def process_frame(self, frame: Frame, direction: FrameDirection):
         if isinstance(frame, StartFrame):
             await self.__start(frame)
         elif isinstance(frame, StartInterruptionFrame):
-            await self._start_interruption()
+            # await self._start_interruption()
```
filipi87: Why is it not needed anymore ?

aconchillo: Ooops! Sorry. I was just testing something. Thanks!

aconchillo:
> Yep, using a priority queue instead of two internal queues makes a lot of sense. That’s probably more optimized than our previous solution. Nice catch.
> 
> I have only one remaining question: why are we no longer invoking `self._start_interruption` when receiving the `StartInterruptionFrame` ?

Nothing. It was just me testing something.

src/pipecat/processors/frame_processor.py
aconchillo: Updated.

filipi87 (approved):
LGTM!

PR #2405, opened 2025-08-09
FrameProcessor: introduce direct mode

aconchillo:

This PR improves some frame processors performance by introducing a direct mode. In direct mode a frame processor will process frames right away avoiding the need for internal queues and tasks. This is useful for some simple processos. For example, in processors that wrap other processors (e.g. `Pipeline`, `ParallelPipeline`), we add one processor before and one after the wrapped processors (internally, you will see them as sources and sinks). These sources and sinks don't do any special processing and they basically forward frames. So, for these simple processors we now enable the new direct mode which avoids creating any internal tasks (and queues) and therefore improves performance.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- Added `enable_direct_mode` argument to `FrameProcessor`. The direct mode is
+  for processors which require very little I/O or compute resources, that is
+  processors that can perform their task almost immediately. These type of
+  processors don't need any of the internal tasks and queues usually created by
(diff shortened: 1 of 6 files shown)


filipi87 (approved):
Cool. Have you been able to measure the impact on performance with these changes ?

Nice improvement. 🚀

aconchillo:
> Cool. Have you been able to measure the impact on performance with these changes ?
> 
> Nice improvement. 🚀

Just visually with htop. It can reduce 1-2% of CPU. Depending on how complicated is the pipeline (i.e. if it include 1 or 2 parallel pipelines or other internal pipelines). It seems we are going to start providing more utilities (like voice mail detection) that might have complicated pipelines, this will be very useful as parallel pipelines won't have any impact now.

PR #2408, opened 2025-08-10
Add MistralLLMService

markbackman:

Mistral uses the OpenAI chat completion format but has two key differences:
1. Mistral expects `prefix=True` to be appended to `assistant` messages when they are the final message in the context.
2. Mistral processes functions differently, as far as I can tell. OpenAI gets tools from the streamed response, but Mistral appears to read them from the context. When using the OpenAI `run_function_calls` function, this results in tools being called twice.

In this PR, `MistralLLMService` inherits from `OpenAILLMService`, but works around the two issues cited:
1. Mutates the context by adding `prefix=True` to `assistant` messages at the end of the context.
2. Overrides the `run_function_calls` function to ensure that the tool is only called once.

The remainder of the base class works as expected, AFAICT.


obigroup:
Hi! I see there's currently work being done on Mistral LLM integration, and I was wondering if there are any plans to also integrate Mistral support into pipecat-flow?

It would be great to have consistency across both projects and allow users to leverage Mistral's capabilities in the flow-based approach as well.

Is this something that's already being considered as part of the current Mistral integration work, or would it be handled separately?

Let me know if there's anything I could help with to make this happen!

src/pipecat/services/mistral/llm.py
```diff
@@ -0,0 +1,188 @@
...
+            else:
+                logger.trace(
+                    f"Skipping already-executed function call: {call.function_name}:{call.tool_call_id}"
+                )
+
+        # Call parent method with filtered list
+        if calls_to_execute:
+            await super().run_function_calls(calls_to_execute)
+
+    async def get_chat_completions(
```
filipi87: Probably worth holding this PR until the [Retry chat completions PR](https://github.com/pipecat-ai/pipecat/pull/2387/files) is merged, since it’s going to change this signature.

examples/foundational/14w-function-calling-mistral.py
```diff
@@ -0,0 +1,165 @@
...
+        api_key=os.getenv("CARTESIA_API_KEY"),
+        voice_id="71a7ad14-091c-4e8e-a314-022ece01c121",  # British Reading Lady
+    )
+
+    llm = MistralLLMService(api_key=os.getenv("MISTRAL_API_KEY"))
+
+    # You can also register a function_name of None to get all functions
+    # sent to the same callback with an additional function_name parameter.
+    llm.register_function("get_current_weather", fetch_weather_from_api)
+    llm.register_function("get_restaurant_recommendation", fetch_restaurant_recommendation)
```
filipi87, line 78: I wonder if we should start using the `register_direct_function` in our examples instead. It does feel so much simpler to use.

src/pipecat/services/mistral/llm.py
```diff
@@ -0,0 +1,188 @@
...
+        OpenAI (Stream-based detection):
+        - Detects function calls only from streaming chunks as the LLM generates them
+        - Second LLM completion doesn't re-detect existing tool_calls in message history
+        - Function calls execute exactly once
+
+        Mistral (Message-based detection):
+        - Detects function calls from the complete message history on each completion
+        - Second LLM completion with the response re-detects the same tool_calls from
+          previous messages
+        - Without filtering, function calls would execute twice
```
filipi87, line 106: Is this true for all models ? For example, if we use `mistral-medium-latest`, would it have the same behavior ?

filipi87 (approved):
LGTM. 🚀

Just a couple more nit questions for my own understanding.

And it’s probably worth holding off on this until we merge the other PR, where we’re adding support for the OpenAI LLM retry.

src/pipecat/services/mistral/llm.py
markbackman: Updated for the new inheritance pattern.

examples/foundational/14w-function-calling-mistral.py
markbackman: Yes, we might want to do this. I'm not going to do update this example, but we can do a sweep on the demos once we want to make the change.

src/pipecat/services/mistral/llm.py
markbackman: In testing, it works the same with the `mistral-medium-latest` model. I would expect that the API is consistent regardless of the model.

pyproject.toml
```diff
@@ -73,6 +73,7 @@ lmnt = [ "websockets>=13.1,<15.0" ]
 local = [ "pyaudio~=0.2.14" ]
 mcp = [ "mcp[cli]~=1.9.4" ]
 mem0 = [ "mem0ai~=0.1.94" ]
+mistral = []
```
obigroup, line 76: @markbackman Why didn't you specify a version? For example: [ "mistralai (>=1.9.2,<2.0.0)" ]

markbackman: Mistral uses the openai client, so the mistral client isn't used at all.

PR #2409, opened 2025-08-10
Refactor PlayHTHttpTTSService to use aiohttp

markbackman:

This makes PlayHTHttpService easier to use with other services.


src/pipecat/services/playht/tts.py
```diff
@@ -441,40 +440,46 @@ def __init__(
...
+
+            output_format: Audio output format. Defaults to "wav".
             sample_rate: Audio sample rate. If None, uses default.
             params: Additional input parameters for voice customization.
             **kwargs: Additional arguments passed to parent TTSService.
         """
         super().__init__(sample_rate=sample_rate, **kwargs)
 
+        # Warn about deprecated protocol parameter if explicitly provided
+        if protocol is not None:
```
aconchillo: `if protocol`

markbackman: 👍

src/pipecat/services/playht/tts.py
```diff
@@ -542,41 +527,76 @@ async def run_tts(self, text: str) -> AsyncGenerator[Frame, None]:
...
+                    json=payload,
+                ) as response:
+                    if response.status not in (200, 201):
+                        error_text = await response.text()
+                        raise Exception(f"PlayHT API error {response.status}: {error_text}")
+
+                    in_header = True
+                    buffer = b""
+
+                    async for chunk in response.content.iter_chunked(8192):
```
aconchillo: we should use

```
                CHUNK_SIZE = self.chunk_size

                async for chunk in response.content.iter_chunked(CHUNK_SIZE):
```

aconchillo: that something i will fix with the refactoring. but for now. we need this to avoid initial audio glitches.

aconchillo (approved):

aconchillo:
LGTM. Just one comment about the chunk size.

src/pipecat/services/playht/tts.py
markbackman: 👍

PR #2420, opened 2025-08-12
Development runner: handle_sigint defaults to False

markbackman:

@aconchillo this is what we talked about earlier today. Now this matches the `pipecatcloud` fallback types.

All examples that used runner_args.handle_sigint were set to False, so the new default should let them proceed without any changes.


aconchillo (approved):

aconchillo:
Perfect. LGTM.

Issue #2421, opened 2025-08-12
400 "unsupported language" on Whisper backends after push frame `STTUpdateSettingsFrame`


thsunkid:
### pipecat version

0.0.77

### Python version

3.13.x

### Operating System

macOS (Darwin 24.4.0)

### Issue description

Setting STT language at runtime via `STTUpdateSettingsFrame` causes Whisper-based services to send an enum (e.g., `Language.EN`) instead of a language code (e.g., "en") to the API, leads to a 400 error from Whisper-compatible backends. 

### Reproduction steps

Reproduction steps
1. Use `GroqSTTService` with `pipecat==0.0.77`.
2. Start the service and later send an `STTUpdateSettingsFrame` with settings={"language": `Language.EN`}.
3. Trigger a transcription so `_transcribe` uses `self._language`.
4. Observe the API error.

### Expected behavior

Should be find to use the STT after sending the `STTUpdateSettingsFrame` with language update down the stream.

### Actual behavior

Got the following errror:

### Logs

```shell
Exception during transcription: Error code: 400 - {'error': {'message': 'unsupported language: Language.EN
Language must be one of: [eu ro te sr sl hy af haw gu uz uk el da mn nn ln su yue es fi ur bg cy mt fr tr lb mg ha et km tg bo jv en ru it th hr sk gl lo zh ml is pa ps my bs yo oc am fo ba de ka as ko pl ar vi hu lt la so ja hi cs ta no mi lv bn fa az kn mk sq sw be sd sv he ne kk si sn yi ht pt id br mr tk sa tl tt ca nl ms]}}
```

markbackman:
Resolved in #2422

PR #2422, opened 2025-08-12
Fix: assigns string code instead of Language enum to BaseWhisperSTTService._language

thsunkid:
To fix the issue described here https://github.com/pipecat-ai/pipecat/issues/2421

Root cause: `BaseWhisperSTTService.__init__` maps `Language` → `str` via `language_to_service_language`, but `BaseWhisperSTTService.set_language` assigns the `Language` enum directly to `self._language`, which caused the error when sending that value to the API. (Here I tested Groq only - but think it's an inconsistent pattern need to be fixed for all).


markbackman (approved):
Nice catch 🙇 

Mind adding a short changelog entry?

PR #2424, opened 2025-08-12
FrameProcessor: fix race condition on FrameProcessorQueue

aconchillo:

We need to increment the counters before the await otherwise we could go to a different task that could add an item with the same counter.

Also, we need to handle non-frame items as well.

### DIFF
--- a/src/pipecat/processors/frame_processor.py
+++ b/src/pipecat/processors/frame_processor.py
@@ -38,7 +38,10 @@ from pipecat.observers.base_observer import BaseObserver, FramePushed
 from pipecat.processors.metrics.frame_processor_metrics import FrameProcessorMetrics
 from pipecat.utils.asyncio.task_manager import BaseTaskManager
 from pipecat.utils.asyncio.watchdog_event import WatchdogEvent
-from pipecat.utils.asyncio.watchdog_priority_queue import WatchdogPriorityQueue
+from pipecat.utils.asyncio.watchdog_priority_queue import (
+    WatchdogPriorityCancelSentinel,
+    WatchdogPriorityQueue,
+)
 from pipecat.utils.asyncio.watchdog_queue import WatchdogQueue
 from pipecat.utils.base_object import BaseObject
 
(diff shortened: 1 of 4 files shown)


src/pipecat/processors/frame_processor.py
```diff
@@ -92,6 +92,7 @@ class FrameProcessorQueue(WatchdogPriorityQueue):
 
     """
 
+    BASE_PRIORITY = 0
```
filipi87: Is this the expected behavior, that items which are not frames should have higher priority than `SystemFrames`?

Do you have an example where this would be the case?

aconchillo: Yes, the watchdog sentinel.

src/pipecat/processors/frame_processor.py
```diff
@@ -103,8 +104,9 @@ def __init__(self, manager: BaseTaskManager):
 
         """
         super().__init__(manager)
-        self.__counter = 0
-        self.__system_counter = 0
+        self.__base_counter = 0
+        self.__high_counter = 0
+        self.__low_counter = 0
 
     async def put(self, item: Tuple[Frame, FrameDirection, FrameCallback]):
```
filipi87: I was looking at the docstring for this method, and we should probably fix it since it still refers to having more than one queue. 
We probably missed this one in the previous PR.

src/pipecat/processors/frame_processor.py
aconchillo: It is the only thing other than frames that we push and it needs to happen fast because we are cancelling.

src/pipecat/processors/frame_processor.py
```diff
@@ -118,11 +120,14 @@ async def put(self, item: Tuple[Frame, FrameDirection, FrameCallback]):
         """
         frame, _, _ = item
         if isinstance(frame, SystemFrame):
-            await super().put((self.HIGH_PRIORITY, self.__system_counter, item))
-            self.__system_counter += 1
+            self.__high_counter += 1
+            await super().put((self.HIGH_PRIORITY, self.__high_counter, item))
```
filipi87, line 124: Yeah, make sense. Nice catch!

src/pipecat/processors/frame_processor.py
aconchillo: Ah, yes. Let me update it.

src/pipecat/processors/frame_processor.py
filipi87: Yeah, that makes sense. I just wonder if we shouldn’t specifically check for this case to use the highest priority, so there’s no chance of accidentally including something else in the future.

src/pipecat/processors/frame_processor.py
aconchillo: Updated.

src/pipecat/processors/frame_processor.py
aconchillo: OK, I have updated the code so it's more clear now.

src/pipecat/utils/asyncio/watchdog_priority_queue.py
```diff
@@ -20,7 +20,7 @@
 
 
 @dataclass
-class _WatchdogPriorityCancelSentinel:
+class WatchdogPriorityCancelSentinel:
```
filipi87, line 23: One final thing, now that it is public, we need to add docstring here.

PR #2427, opened 2025-08-12
WatchdogPriorityQueue: this is now a base class

aconchillo:

Sub-classes need to implement `put_cancel_item()` and `get_value()` methods. The reason is because we are inserting the watchdog cancel sentinel and there needs to be a way to compare it with other items. The `__lt__` method was not enough because it depends on the order of comparisson. To make it more robust sub-classes now need to indicate how to insert a watchdogcancel sentinel, following the priority scheme.

### DIFF
--- a/src/pipecat/processors/frame_processor.py
+++ b/src/pipecat/processors/frame_processor.py
@@ -97,13 +97,11 @@ class FrameProcessorQueue(WatchdogPriorityQueue):
             manager (BaseTaskManager): The task manager used by the internal watchdog queues.
 
         """
-        super().__init__(manager)
+        super().__init__(manager, tuple_size=3)
         self.__high_counter = 0
         self.__low_counter = 0
 
-    async def put(
-        self, item: WatchdogPriorityCancelSentinel | Tuple[Frame, FrameDirection, FrameCallback]
-    ):
+    async def put(self, item: Tuple[Frame, FrameDirection, FrameCallback]):
         """Put an item into the priority queue.
 
         System frames (`SystemFrame`) have higher priority than any other
@@ -114,10 +112,6 @@ class FrameProcessorQueue(WatchdogPriorityQueue):
             item (Any): The item to enqueue.
 
         """
-        if isinstance(item, WatchdogPriorityCancelSentinel):
-            await super().put(item)
-            return
-
         frame, _, _ = item
         if isinstance(frame, SystemFrame):
             self.__high_counter += 1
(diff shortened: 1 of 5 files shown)


src/pipecat/processors/frame_processor.py
```diff
@@ -126,6 +120,9 @@ async def put(
             self.__low_counter += 1
             await super().put((self.LOW_PRIORITY, self.__low_counter, item))
 
+    def put_cancel_sentinel(self):
+        super().put_nowait((float("-inf"), 0, WatchdogPriorityCancelSentinel()))
+
     async def get(self) -> Any:
```
filipi87: I guess we don’t need to override this method anymore, right?

filipi87: Yeah, but if we don't who is invoking it would need to do something like this:
`_, _, item`

So, yeah, never mind, we still need this. 👍

filipi87 (commented):
The main issue was that we were previously unable to know the tuple size needed to put and retrieve elements. That’s basically why we tried the approach using `__lt__` with `WatchdogPriorityCancelSentinel`.

In this approach, we always need to extend `WatchdogPriorityQueue`, and the classes must override `get_value` and `put_cancel_sentinel`.

What if, instead, we could instantiate `WatchdogPriorityQueue` by defining this behavior upfront, the number of items in the tuple ? So it would still be possible to know how to insert and retrieve elements without needing to be extended.

What do you think ?

aconchillo:
> The main issue was that we were previously unable to know the tuple size needed to put and retrieve elements. That’s basically why we tried the approach using `__lt__` with `WatchdogPriorityCancelSentinel`.
> 
> In this approach, we always need to extend `WatchdogPriorityQueue`, and the classes must override `get_value` and `put_cancel_sentinel`.
> 
> What if, instead, we could instantiate `WatchdogPriorityQueue` by defining this behavior upfront, the number of items in the tuple ? So it would still be possible to know how to insert and retrieve elements without needing to be extended.
> 
> What do you think ?

Yes, simple, so I just implemented this suggestion. We are forcing the queue to behave in a certain way, but I believe that's fine.

The issue I saw was that we still don't know what values the items prior to the sentinel are. That is, if it's a tuple with 3 values and the sentinel goes in 3rd place, what are the values for 1st and 2nd? We can't assume they are numerical values if we want a generic queue. But it's a trade-off of simplicity vs flexibility. Simplicity is probably right in this case.

src/pipecat/processors/frame_processor.py
```diff
@@ -114,10 +112,6 @@ async def put(
             item (Any): The item to enqueue.
 
         """
-        if isinstance(item, WatchdogPriorityCancelSentinel):
-            await super().put(item)
-            return
```
aconchillo, line 119: We don't need this because the sentinel uses `put_nowait`.

PR #2431, opened 2025-08-13
Bugfix/aws nova sonic interruption propagation

SureshkumarKV:
Nova Sonic can detect and handle interruptions natively, known as barge-in support and is much more realiable that VADs.  However, the current pipecat implementation for Nova Sonic ignores the interrupt event emitted by nova sonic. 

This PR is to correctly process the interrupt event from the Nova sonic by creating and propagating a StartInterruptionFrame. It has been well tested and works well.

Here is a reference for the Nova sonic's barge-in support: 
https://github.com/aws-samples/amazon-nova-samples/blob/main/speech-to-speech/sample-codes/console-python/nova_sonic.py

Speech to speech models like Nova Sonic are seeing a lot of traction, and I myself have helped multiple customers with this patch. So kindly review and merge this PR at the earliest. 

Thank you !

### DIFF
--- a/src/pipecat/services/aws_nova_sonic/aws.py
+++ b/src/pipecat/services/aws_nova_sonic/aws.py
@@ -931,7 +931,7 @@ class AWSNovaSonicLLMService(LLMService):
 
         if content.role == Role.ASSISTANT:
             if content.type == ContentType.TEXT:
-                # Ignore non-final text, and the "interrupted" message (which isn't meaningful text)
+                # Ignore non-final text, but do process the "interrupted" message
                 if content.text_stage == TextStage.FINAL:
                     if stop_reason != "INTERRUPTED":
                         if self._assistant_is_responding:


kompfner:
Thanks for the PR!

Just going to jot down some top-of-mind questions here.

- What was the main impact on users of not having this? Pipecat pipelines are typically set up with VAD running on the input at the top of the pipeline. Was it that some interruptions were not being detected by VAD but were being detected by Nova Sonic, or vice versa?
- With the user interruption detected by AWS Nova Sonic, would the recommendation be disabling VAD on the input? If so...I think we'd need to also be generating `UserStartedSpeaking` frames here...and maybe also `UserStoppedSpeaking` frames and `StopInterruption` frames. (The way to do this would probably be with upstream-sent `EmulateUserStartedSpeakingFrame` and `EmulateUserStoppedSpeakingFrame`)
- Should we honor the pipeline-level `allow_interruptions` property here? (If we sent an upstream `BotInterruptionFrame`, which instructs the input to emit a `StartInterruptionFrame`, we might get that for free)

IIRC part of the decision to punt on handling the Nova Sonic-driven interruption detection was due to the complexity of reasoning about different, possibly divergent sources of truth about whether a user interruption was ongoing (VAD and Nova Sonic). So we went with the "standard" Pipecat pattern of letting VAD be the one source of truth.

But that doesn't mean this isn't worth pursuing! It might end up just being a bit more involved.

SureshkumarKV:
Thank you for your kind feedback.

1. In my testing the interruption from Nova sonic felt quicker and probably more realiable in the face of noisy environments.
2. Yes, I was having VAD disabled. It may be a challenge to create UserStartedSpeaking, UserStoppedSpeaking etc reliably based on the interruot event from sonic, becasue sonic will trigger it only if there was a literal interruption (Not of the user started talking after the bot finished talking)
3. pipeline-level allow_interruptions should be honoured for consistency.

I now understand why you didnt include the sonic based interruption, particularly the point 2 above can be a challenge.
Honestly I am not able to propose a modification where we can address the point 2 properly.

src/pipecat/services/aws_nova_sonic/aws.py
```diff
@@ -930,11 +931,15 @@ async def _handle_content_end_event(self, event_json):
...
-                        # Text added to the ongoing assistant response
-                        await self._report_assistant_response_text_added(content.text_content)
+                # Ignore non-final text, but do process the "interrupted" message
+                if content.text_stage == TextStage.FINAL:
+                    if stop_reason != "INTERRUPTED":
+                        if self._assistant_is_responding:
+                            # Text added to the ongoing assistant response
+                            await self._report_assistant_response_text_added(content.text_content)
+                    else:
+                        # We got an interruption(barge-in) detection from Nova Sonic.
```
kompfner, line 941: Hey! Pardon the delay in getting back to you. 🙏 

I have a suggestion which hopefully gives us a path forward here:


```suggestion
                        # We got an interruption(barge-in) detection from Nova Sonic.
                        # We probably *also* detected the interruption in the pipeline via VAD, but
                        # this is a good "backup" to make sure we catch it. It doesn't seem to hurt
                        # to have both.
                        await self.push_frame(BotInterruptionFrame(), FrameDirection.UPSTREAM)
```

The idea is that we let both VAD and AWS Nova Sonic catch interruptions. In my (albeit limited) testing, it doesn't seem to hurt to have both enabled (they usually fire _very_ close together anyway).

Above, I've changed `StartInterruptionFrame` to `BotInterruptionFrame`, which is a frame that essentially means "please fire a `StartInterruptionFrame`". This just guarantees that the `StartInterruptionFrame` originates from the same place in the pipeline it would usually start at if detect by VAD—the input transport.

I also thought a bit more about the question of honoring `allow_interruptions`, and...I'm not sure it really applies here. When you're using a speech-to-speech model like Nova Sonic, you're yielding control over interruptions to the model—the bot will be interruptible regardless of what `allow_interruptions` setting the user configured the Pipecat pipeline with. So I think it's OK to ignore it here.

WDYT?

kompfner:
Would you mind also adding an entry to the CHANGELOG? Thanks!

markbackman:
@kompfner is this PR still relevant?

kompfner:
> @kompfner is this PR still relevant?

I actually think it might be—speech-to-speech (aka realtime) services do VAD/turn detection on their own, so I think it'd be great if we found a uniform pattern they can all adopt to ensure they can be used without local VAD (the motivation behind #3429) or alongside local VAD.

The pattern for speech-to-speech services could look like that of `LLMUserAggregator` or `UserTurnProcessor`:
- As soon as it detects that the user has started speaking (whether by a barge-in or otherwise):
  - `await self.push_interruption_task_frame_and_wait()`
  - `await self.broadcast_frame(UserStartedSpeakingFrame)`
- As soon as it detects the user stopped speaking (or the bot started responding, whichever signal is actually available)
  - `await self.broadcast_frame(UserStoppedSpeakingFrame)`

My assumption (and correct me if I'm wrong, @aconchillo, who has thought more about this) is that it _probably_ doesn't hurt to have the above signals coming from the speech-to-speech services even if local VAD + default turn strategies are enabled, meaning they'd _also_ emit those signals (meaning we'd have duplicate signals). If the user wanted to ensure there wouldn't be duplicate interruptions + user started/stopped frames, they could configure their `LLMUserAggregator` with `ExternalUserTurnStrategies()` (at the possible expense of the speed of local-VAD-driven turn detection).

After all, the speech-to-speech service really always _is_ doing VAD/turn detection in the audible conversation, whether or not the pipeline is _also_ configured to do it independently.

It looks like OpenAI Realtime already always emits interruptions + user started/stopped frames: https://github.com/pipecat-ai/pipecat/blob/main/src/pipecat/services/openai/realtime/llm.py#L787-L795

kompfner:
OK, looks like AWS Nova Sonic API has the same issue as Gemini Live API (see https://github.com/pipecat-ai/pipecat/pull/3429), which is the lack of a reliable signal to base user started/stopped speaking frames off of. My recommendation here will be to do the same as in that PR—a one-line code change plus a comment:

```
# NOTE: while the service triggers interruptions in
# the specific case of barge-ins, it does *not*
# emit UserStarted/StoppedSpeakingFrames, as the
# Nova Sonic API does not give us broadly reliable
# signals to base those off of. Pipelines that
# require turn tracking (like those using context
# aggregators) still need an independent way to
# track turns, such as local Silero VAD in
# combination with the context aggregator default
# turn strategies.
```

And a corresponding CHANGELOG entry:
```
- Added handling for `server_content.interrupted` signal in the AWS Nova Sonic service for faster interruption response in the case where there isn't already turn tracking in the pipeline, e.g. local VAD + context aggregators. When there is already turn tracking in the pipeline, the additional interruption does no harm.
```

This should only apply to very rare cases, I would think/hope, since you can't really use `AWSNovaSonicLLMService` without context aggregators (it relies on a received `LLMContext` to kick off the conversation), so disabling local VAD would only make it so that those context aggregators can't accurately track turns (which I suppose is fine if you don't really need it to...). Also, in my testing, local VAD is almost always a bit faster.

UPDATE: will go ahead and just make this change in another PR (`main` has drifted quite a bit from this PR).

kompfner:
> UPDATE: will go ahead and just make this change in another PR (main has drifted quite a bit from this PR).

Actually, for some reason (didn't dig into it too much) simply firing an additional interruption frame here does _not_ play nicely at all with interruption frames fired by the default user turn start strategies because of the strange timing of the Nova Sonic user transcriptions. 

So, pretty much, local VAD today is also absolutely needed for AWS Nova Sonic in a Pipecat pipeline...

Closing this PR for now.

PR #2440, opened 2025-08-13
Support for runtime LLM switching

kompfner:
## UPDATE (2025-08-25)

Two follow-up PRs have been rolled into this one:

* https://github.com/pipecat-ai/pipecat/pull/2477
* https://github.com/pipecat-ai/pipecat/pull/2485

Collectively, they round out the infrastructure we need for supporting runtime LLM switching and making it possible for Pipecat Flows to add that feature as well.

## Background

We want developers to be able to swap LLMs at runtime. This is useful for failover, among other scenarios.

Currently, one reason developers are inhibited from implementing LLM-swappable pipelines is because different LLMs use different LLM-specific context machinery (contexts and context aggregators) under the hood. Keeping the various contexts in sync would be very onerous—if not impossible—with today’s architecture.

The simplest solution from the developer’s point of view is for there to be a “universal” context and associated set of context aggregators that work across all LLMs, so there aren't multiple contexts to keep in sync.

## Changes in this PR

This PR introduces "universal" (LLM-agnostic) machinery:
- A new universal context
    - `LLMContext`
- A new universal context aggregator pair
    - `LLMContextAggregatorPair`, comprised of:
        - `LLMUserAggregator` (”LLMUserContextAggregator” was already taken ☹️)
        - `LLMAssistantAggregator` (”LLMAssistantContextAggregator” was already taken ☹️)
- A new frame for setting a new universal context on an LLM
    - `LLMContextFrame` (eventually replacing `OpenAILLMContextFrame`)
- Just-in-time translation from universal context formats to LLM-specific formats as needed by the LLM, with the aid of LLM-specific adapters

So far, support for the new universal context machinery has only been added to Google and OpenAI. That means that only those services will be usable in applications that require switching between LLMs at runtime. Support can be rolled out incrementally to all the other LLM services.

## Usage

From the developer's point of view, switching to using the new universal context machinery is a matter of going from this:

```python
context = OpenAILLMContext(messages, tools)
context_aggregator = llm.create_context_aggregator(context)
```

To this:

```python
context = LLMContext(messages, tools)
```

The new pattern is optional for now, of course. Both are supported. But developers will have to use the new pattern if they want to do any runtime LLM switching.

## Examples

* The `14x` and `14y` foundational examples provide basic validation that the universal context machinery works in plain OpenAI and Google scenarios.
* The `prototype-llm-failover.py` example found in [this pipecat-flows draft PR](https://github.com/pipecat-ai/pipecat-flows/pull/174) lets you play around with switching back and forth between LLMs at runtime.
  * UPDATE: the Pipecat Flows changes in that draft PR are now dependent on https://github.com/pipecat-ai/pipecat/pull/2477 and https://github.com/pipecat-ai/pipecat/pull/2485, which are ahead of, and open against, this PR

## Next steps

The problem of keeping context synced between multiple LLMs isn't the only deterrent today to building scenarios that involve runtime LLM switching. We also don't have primitives that make it simple to insert multiple LLMs into a pipeline. As you can see in [`prototype-llm-failover.py`](https://github.com/pipecat-ai/pipecat-flows/pull/174), building an "LLM switcher" involves some boilerplate. We could provide something like an `LLMSwitcher` that can be used directly in your pipeline. It could be configurable, with various switching strategy—`manual`, `failover`, `custom`, etc.

Once an `LLMSwitcher` exists, Pipecat Flows ought to take _that_, rather than array of LLM services, as in [the prototype PR](https://github.com/pipecat-ai/pipecat-flows/pull/174).

We also need to roll out support for the universal LLM context to the other LLM services.


examples/foundational/14w-function-calling-universal-context.py
```diff
@@ -0,0 +1,170 @@
...
+
+    messages = [
+        {
+            "role": "system",
+            "content": "You are a helpful LLM in a WebRTC call. Your goal is to demonstrate your capabilities in a succinct way. Your output will be converted to audio so don't include special characters in your answers. Respond to what the user said in a creative and helpful way.",
+        },
+    ]
+
+    context = LLMContext(messages, tools)
+    context_aggregator = LLMContextAggregatorPair.create(context)
```
kompfner: The above two lines are an illustration of the main developer-facing change

src/pipecat/adapters/services/gemini_adapter.py
```diff
@@ -6,21 +6,66 @@
...
+    """Gemini-specific adapter for Pipecat.
+
+    Handles:
+    - Extracting parameters for Gemini's API from a universal LLM context
+    - Converting Pipecat's standardized tools schema to Gemini's function-calling format.
+    - Extracting and sanitizing messages from the LLM context for logging with Gemini.
     """
 
-    def to_provider_tools_format(self, tools_schema: ToolsSchema) -> List[Dict[str, Any]]:
+    def get_llm_invocation_params(self, context: LLMContext) -> GeminiLLMInvocationParams:
```
kompfner, line 57: Here's the main new method in the adapters. It's just a one-shot just-in-time translation from universal `LLMContext` into whatever is needed to invoke the LLM.

src/pipecat/adapters/services/gemini_adapter.py
```diff
@@ -39,3 +84,214 @@ def to_provider_tools_format(self, tools_schema: ToolsSchema) -> List[Dict[str,
...
+                elif c["type"] == "image_url":
+                    parts.append(
+                        Part(
+                            inline_data=Blob(
+                                mime_type="image/jpeg",
+                                data=base64.b64decode(c["image_url"]["url"].split(",")[1]),
+                            )
+                        )
+                    )
+                elif c["type"] == "input_audio":
```
kompfner, line 306: Most of this translation work in this file is a straight port from `GoogleLLMContext._restructure_from_openai_messages()` and `GoogleLLMContext.from_standard_message()`.

This part handling translation from the universal "input_audio" message to the Google-specific one is new: we previously only handled adding input audio messages directly into `GoogleLLMContext`s, via `GoogleLLMContext.add_audio_frames_message()`.

I was on the fence about whether to just say "`add_audio_frames_message()` is unsupported for now when using the universal `LLMContext`—please use `GoogleLLMContext`, as before". But we would need support for `LLMContext.add_audio_frames_message()` sooner or later, since: a) we do eventually want everyone to replace `GoogleLLMContext` with `LLMContext` so we can deprecate the former and b) we do eventually want support for audio messages in OpenAI and other services.

src/pipecat/processors/aggregators/llm_context.py
```diff
@@ -0,0 +1,272 @@
...
+
+from pipecat.adapters.schemas.tools_schema import ToolsSchema
+from pipecat.frames.frames import AudioRawFrame, Frame
+
+# "Re-export" types from OpenAI that we're using as universal context types.
+# NOTE: if universal message types need to someday diverge from OpenAI's, we
+# should consider managing our own definitions. But we should do so carefully,
+# as the OpenAI messages are somewhat of a standard and we want to continue
+# supporting them.
+LLMStandardMessage = ChatCompletionMessageParam
```
kompfner, line 39: I went back and forth between creating a new "real" (as opposed to `TypedDict`-based) `LLMStandardMessage` class with fields like `role` and `content`, or doing what I ultimately did here, which was to alias the OpenAI type (which is a `TypedDict`).

My reasoning:
- We want to keep supporting the OpenAI message format anyway, which has become something of a standard
- We don't want to be in the business of maintaining a mirror of OpenAI's message type definition, especially since the type definition actually isn't that trivial (it'd be a fair amount of code)
- Wrapping all messages in a `LLMStandardMessage` constructor would make context declaration rather more verbose
- There's nothing stopping us from still defining a "real" `LLMSpecificMessage` type for messages that are specific to a particular LLM service (which we do below)

src/pipecat/processors/aggregators/llm_context.py
```diff
@@ -0,0 +1,272 @@
...
+            messages: Initial list of conversation messages.
+            tools: Available tools for the LLM to use.
+            tool_choice: Tool selection strategy for the LLM.
+        """
+        self._messages: List[LLMContextMessage] = messages if messages else []
+        self._tools: ToolsSchema | NotGiven = tools
+        self._tool_choice: LLMContextToolChoice | NotGiven = tool_choice
+        self._validate_tools()
+
+    def get_messages(self, llm_specific_filter: Optional[str] = None) -> List[LLMContextMessage]:
```
kompfner, line 85: This method is how LLM-specific adapters read messages from the context. When they read messages, they get all universal messages plus only those that are targeted specifically at a their LLM.

src/pipecat/processors/aggregators/llm_response_universal.py
```diff
@@ -0,0 +1,845 @@
...
+            tool_choice: Tool choice configuration for the context.
+        """
+        self._context.set_tool_choice(tool_choice)
+
+    async def reset(self):
+        """Reset the aggregation state."""
+        self._aggregation = ""
+
+
+class LLMUserAggregator(LLMContextAggregator):
```
kompfner, line 172: As you can see, there's quite a bit of code duplicated between the universal `LLMUserAggregator`/`LLMAssistantAggregator` and `LLMUserContextAggregator`/`LLMAssistantContextAggregator`.

I didn't want to get too clever and DRY things up at the expense of adding complexity, since _ideally_ we'd get to the point relatively quickly of being able to declare these new universal aggregators the go-to ones and deprecating the old ones.

But it does mean, at least in the short term, that we may have to apply some bug fixes to two places.

src/pipecat/processors/aggregators/llm_response_universal.py
kompfner, line 1: Some day we should be able to rename this to just `llm_response.py`, once the context aggregators in that file are deprecated and then removed.

src/pipecat/processors/aggregators/llm_response_universal.py
kompfner, line 172: These new aggregators are named `LLMUserAggregator`/`LLMAssistantAggregator` because `LLMUserContextAggregator`/`LLMAssistantContextAggregator` were already taken.

As a consolation bonus, though, these names now better match the name of their params (`LLMUserAggregatorParams`, `LLMAssistantAggregatorParams`), which have _not_ been duplicated.

src/pipecat/services/google/llm.py
```diff
@@ -437,11 +440,20 @@ def from_standard_message(self, message):
                 )
         elif role == "tool":
             role = "model"
+            try:
```
kompfner, line 449: Discovered a bug while working on this PR. `FunctionResponse()` requires a dict. So here we're wrapping the tool response in a dict if it’s not already one. That’s what already we were already doing in `GoogleAssistantContextAggregator._update_function_call_result()`, where we're not handling translating from an OpenAI context but rather working with a Google context directly.

I first made the fix in the new `GeminiLLMAdapter._from_universal_context_message()` and back-ported it here.

src/pipecat/services/google/llm.py
```diff
@@ -916,9 +964,18 @@ async def process_frame(self, frame: Frame, direction: FrameDirection):
 
         if isinstance(frame, OpenAILLMContextFrame):
             context = GoogleLLMContext.upgrade_to_google(frame.context)
+        elif isinstance(frame, LLMContextFrame):
```
kompfner, line 1017: Here you can see that we still support `OpenAILLMContextFrame`. 

Universal context machinery is additive, at the moment—the developer can choose whether to use `OpenAILLMContextFrame`s and `llm.create_context_aggregator()` or follow the new "universal" pattern.

If they go with the new pattern, `LLMContextFrame`s will go through the pipeline instead of `OpenAILLMContextFrame`.

src/pipecat/services/google/llm.py
```diff
@@ -419,7 +422,14 @@ def from_standard_message(self, message):
         role = message["role"]
         content = message.get("content", [])
         if role == "system":
-            self.system_message = content
+            # System instructions are returned as plain text
+            if isinstance(content, str):
+                self.system_message = content
+            elif isinstance(content, list):
+                # If content is a list, we assume it's a list of text parts, per the standard
```
kompfner, line 428: Noticed this bug as I was working. Here I back-ported the fix to the non-universal-context code path.

src/pipecat/processors/aggregators/llm_context.py
```diff
@@ -0,0 +1,272 @@
...
+
+        Args:
+            format: Image format (e.g., 'RGB', 'RGBA').
+            size: Image dimensions as (width, height) tuple.
+            image: Raw image bytes.
+            text: Optional text to include with the image.
+        """
+        buffer = io.BytesIO()
+        Image.frombytes(format, size, image).save(buffer, format="JPEG")
+        # TODO: we might not want the universal format to be base64 encoded, since encoding is not needed by all LLM services; today, te Gemini adapter has to decode from base64, which is less than ideal.
```
kompfner: The alternative would be to store the image messages not encoded and just-in-time encode them for OpenAI inference...

...but I'm not sure it's a significant cost, and there's something very very nice about being able to assume that every message in the `LLMContext` truly is in the standard OpenAI format.

So I'm leaning towards closing this TODO. Thoughts?

src/pipecat/services/openai/base_llm.py
```diff
@@ -342,10 +368,21 @@ async def process_frame(self, frame: Frame, direction: FrameDirection):
...
         elif isinstance(frame, LLMMessagesFrame):
+            # NOTE: LLMMessagesFrame is deprecated, so we don't support the newer universal
+            # LLMContext with it
             context = OpenAILLMContext.from_messages(frame.messages)
         elif isinstance(frame, VisionImageRawFrame):
+            # This is only useful in very simple pipelines because it creates
+            # a new context. Generally we want a context manager to catch
+            # UserImageRawFrames coming through the pipeline and add them
+            # to the context.
+            # TODO: support the newer universal LLMContext with a VisionImageRawFrame equivalent?
```
kompfner, line 468: I'm tempted to say "we'll deal with this in a future PR"

src/pipecat/processors/aggregators/llm_context.py
```diff
@@ -0,0 +1,272 @@
...
+        """
+        self._messages[:] = messages
+
+    def set_tools(self, tools: ToolsSchema | NotGiven = NOT_GIVEN):
+        """Set the available tools for the LLM.
+
+        Args:
+            tools: A ToolsSchema or NOT_GIVEN to disable tools.
+        """
+        # TODO: convert empty ToolsSchema to NOT_GIVEN if needed?
```
kompfner: I think this might be an edge case that's not handled in today's `OpenAILLMContext` either

kompfner: Now handling the edge case here but didn't back-port the fix to `OpenAILLMContext`.

src/pipecat/services/openai/base_llm.py
```diff
@@ -328,11 +354,24 @@ async def _process_context(self, context: OpenAILLMContext):
...
+    @property
+    def supports_universal_context(self) -> bool:
+        """Check if this service supports universal LLMContext.
+
+        Returns:
+            Whether service supports universal LLMContext.
+        """
+        # Return True in subclasses that support universal LLMContext
+        # This property lets us gradually roll out support for universal
+        # LLMContext to OpenAI-like services in a controlled manner.
```
kompfner, line 431: For now we'll just mark most "OpenAI-like" services as not supporting universal context. I foresee us being able to add these in one by one fairly quickly (most should just be a matter of testing)

kompfner: (This approach is mostly in service of being able to land this PR without having to do _all_ the work up-front)

src/pipecat/processors/aggregators/llm_context.py
```diff
@@ -0,0 +1,278 @@
...
+
+        Args:
+            format: Image format (e.g., 'RGB', 'RGBA').
+            size: Image dimensions as (width, height) tuple.
+            image: Raw image bytes.
+            text: Optional text to include with the image.
+        """
+        buffer = io.BytesIO()
+        Image.frombytes(format, size, image).save(buffer, format="JPEG")
+        # TODO: we might not want the universal format to be base64 encoded, since encoding is not needed by all LLM services; today, te Gemini adapter has to decode from base64, which is less than ideal.
```
markbackman:
```suggestion
        # TODO: we might not want the universal format to be base64 encoded, since encoding is not needed by all LLM services; today, the Gemini adapter has to decode from base64, which is less than ideal.
```

src/pipecat/services/openai/base_llm.py
markbackman: I think that's fine. Is the idea that this is just a "beta" implementation that handles text for OpenAI and Google? If so, I think that makes sense and keeps the scope small enough to get the core changes in place. We can add from there.

src/pipecat/adapters/base_llm_adapter.py
```diff
@@ -11,21 +11,44 @@
...
+
+class BaseLLMAdapter(ABC, Generic[TLLMInvocationParams]):
     """Abstract base class for LLM provider adapters.
 
-    Provides a standard interface for converting between Pipecat's standardized
-    tool schemas and provider-specific tool formats. Subclasses must implement
-    provider-specific conversion logic.
+    Provides a standard interface for converting to provider-specific formats.
+
+    Handles:
```
markbackman, line 30: Add a new line before lists to comply with the docs auto-gen rules:

```suggestion
    Handles:

```

src/pipecat/adapters/services/gemini_adapter.py
```diff
@@ -6,21 +6,72 @@
...
+        """
+        messages = self._from_universal_context_messages(self._get_messages(context))
+        return {
+            "system_instruction": messages.system_instruction,
+            "messages": messages.messages,
+            # NOTE; LLMContext's tools are guaranteed to be a ToolsSchema (or NOT_GIVEN)
+            "tools": self.from_standard_tools(context.tools),
+        }
+
+    def to_provider_tools_format(self, tools_schema: ToolsSchema) -> List[Any]:
```
markbackman: I see the return type is now `Any`. Is there a reason to make it more permissive?

src/pipecat/adapters/services/gemini_adapter.py
```diff
@@ -39,3 +90,213 @@ def to_provider_tools_format(self, tools_schema: ToolsSchema) -> List[Dict[str,
...
+                            part["inline_data"]["data"] = "..."
+            except Exception as e:
+                logger.debug(f"Error: {e}")
+            messages_for_logging.append(obj)
+        return messages_for_logging
+
+    def _get_messages(self, context: LLMContext) -> List[LLMContextMessage]:
+        return context.get_messages("google")
+
+    @dataclass
```
markbackman, line 125: Do we have a rule of thumb of when to use a dataclass and when to use a TypedDict?

I've defaulted to `@dataclass` which comes with extensibility, but I see you used a `TypedDict` above, which has me question other uses in the codebase. 🤔

src/pipecat/adapters/services/gemini_adapter.py
```diff
@@ -39,3 +90,213 @@ def to_provider_tools_format(self, tools_schema: ToolsSchema) -> List[Dict[str,
...
+    def _from_universal_context_messages(
+        self, universal_context_messages: List[LLMContextMessage]
+    ) -> ConvertedMessages:
+        """Restructures messages to ensure proper Google format and message ordering.
+
+        This method handles conversion of OpenAI-formatted messages to Google format,
+        with special handling for function calls, function responses, and system messages.
+        System messages are added back to the context as user messages when needed.
+
+        The final message order is preserved as:
```
markbackman, line 141: newline for docs

```suggestion
        The final message order is preserved as:

```

src/pipecat/adapters/services/gemini_adapter.py
```diff
@@ -39,3 +90,213 @@ def to_provider_tools_format(self, tools_schema: ToolsSchema) -> List[Dict[str,
...
+        1. Function calls (from model)
+        2. Function responses (from user)
+        3. Text messages (converted from system messages)
+
+        Note:
```
markbackman: This will provide code fencing styling.
```suggestion
        Note::

```

src/pipecat/processors/aggregators/llm_context.py
```diff
@@ -0,0 +1,278 @@
...
+# as the OpenAI messages are somewhat of a standard and we want to continue
+# supporting them.
+LLMStandardMessage = ChatCompletionMessageParam
+LLMContextToolChoice = ChatCompletionToolChoiceOptionParam
+NOT_GIVEN = OPEN_AI_NOT_GIVEN
+NotGiven = OpenAINotGiven
+
+
+@dataclass
+class LLMSpecificMessage:
```
markbackman, line 46: What's an example usage of this? Is this just an internal class?

src/pipecat/adapters/services/open_ai_adapter.py
```diff
@@ -6,22 +6,62 @@
...
-    function calling capabilities.
+    messages: List[ChatCompletionMessageParam]
+    tools: List[ChatCompletionToolParam] | OpenAINotGiven
+    tool_choice: ChatCompletionToolChoiceOptionParam | OpenAINotGiven
+
+
+class OpenAILLMAdapter(BaseLLMAdapter[OpenAILLMInvocationParams]):
+    """OpenAI-specific adapter for Pipecat.
+
+    Handles:
```
markbackman, line 42: newline for docs

```suggestion
    Handles:

```

src/pipecat/processors/aggregators/llm_context.py
markbackman: This seems like a sensible approach. OpenAI messages are the standard, so adopting their types make sense.

src/pipecat/processors/aggregators/llm_context.py
markbackman: Oh! Is this targeting a _specific_ LLM, the case where there are multiple LLMs?

src/pipecat/processors/aggregators/llm_context.py
markbackman: No strong feelings. OpenAI being the standard format is a nice decision to stick to. It creates a uniform definition/expectation across the board.

src/pipecat/processors/aggregators/llm_response_universal.py
markbackman: Nice, clean port here. Very easy to follow. I think two separate files makes a lot of sense for future deprecation. We just need to remember to update logic in both places, as we make changes.

I think the only planned changes are around interruption strategies which would be in both user context aggregators.

Is the idea that we'll keep these names (e.g. `LLMUserAggregator`) for the long-haul? It's a bummer to lose the word `Context` from the name, but I think it's OK; these are mostly internal classes. Also, the file renaming should be non-breaking as that's an internal reference only, too.

src/pipecat/services/google/llm_openai.py
```diff
@@ -62,6 +62,15 @@ def __init__(
         """
         super().__init__(api_key=api_key, base_url=base_url, model=model, **kwargs)
 
+    @property
+    def supports_universal_context(self) -> bool:
+        """Check if this service supports universal LLMContext.
+
+        Returns:
+            False, as GoogleLLMOpenAIBetaService does not yet support universal LLMContext.
```
markbackman, line 82: Unrelated, I think we ultimately want to deprecate this service. Google does not recommend this pattern for production use.

Also, with a universal context, having an "OpenAI-way" of interfacing with Google doesn't make any sense.

Given that, maybe mark it as deprecated now?

src/pipecat/services/google/llm_vertex.py
```diff
@@ -139,3 +139,12 @@ def _get_api_token(credentials: Optional[str], credentials_path: Optional[str])
         creds.refresh(Request())  # Ensure token is up-to-date, lifetime is 1 hour.
 
         return creds.token
+
+    @property
```
markbackman, line 143: This also uses OpenAI, but we really want it to use Google. Maybe we can rewrite this class separately from this PR. So, don't mark as deprecated.

src/pipecat/services/openai/base_llm.py
markbackman: `supports_universal_context` makes sense. Should it be included in this class and set to `return False`? Or, is this an `@abstractmethod` that inheriting classes must implement? (Or, just implemented here as `pass`?)

src/pipecat/processors/aggregators/llm_response_universal.py
kompfner: > Is the idea that we'll keep these names (e.g. LLMUserAggregator) for the long-haul?

Once we deprecate and eventually remove the "old" aggregator classes (e.g. `LLMUserContextAggregator`), we could reclaim those names. I agree that they're better.

src/pipecat/processors/aggregators/llm_context.py
kompfner: Closing the TODO, then

src/pipecat/services/openai/base_llm.py
kompfner: Universal `LLMContext` handles audio and image messages, too :)

It just doesn't yet handle this rather niche `VisionImageRawFrame`, which is a special frame that overwrites all prior context. I'd need to put more time into understanding what exactly this is for and whether it even matters what flavor of context (universal or not) it creates under the hood. It seems like this frame type is created only by `VisionImageFrameAggregator`, which is used in the foundational `12-*` examples.

So I'm just punting this one into a future (but near-term) PR.

src/pipecat/services/openai/base_llm.py
kompfner: `return False` made some sense to me because if some class were to subclass `BaseOpenAIService`, behaving as though they don't support universal LLM context is the "safe" choice.

To be honest, though, this is all kind of academic, so whatever choice we make here doesn't really matter: only `OpenAILLMService` subclasses `BaseOpenAILLMService` today, and a lot of services subclass _that_ one. And honestly, supporting the new context across all those services should be fairly quick work in a couple of follow-up PRs, so I foresee this being very short-lived.

src/pipecat/processors/aggregators/llm_context.py
kompfner: Removed this code comment

src/pipecat/adapters/services/gemini_adapter.py
kompfner: There was, but I can't quite remember what it was.

Maybe it was to align with `BaseLLMAdapter.to_provider_tools_format()`'s return type? If so, that might be a misguided choice—most `BaseLLMAdapter` subclasses are more specific in what they return from `to_provider_tools_format()`...

Going ahead and reverting back to `List[Dict[str, Any]]`

src/pipecat/adapters/services/gemini_adapter.py
kompfner: Can't remember if there was a good reason to go with `TypedDict`. Looking back, you're right that `@dataclass` does seem like a better choice. Let me make that change.

src/pipecat/processors/aggregators/llm_context.py
kompfner: Actually, there's not really a concrete use-case yet, but in one of our team meetings @aconchillo and @filipi87 pointed out (wisely) that it'd be good to have an "escape hatch" mechanism for specifying messages in the conversation history that aren't (or aren't yet) supported by the universal `LLMContext` format, in the same way that we can provide non-standard, LLM-specific tools in `ToolsSchema`.

I can see this being useful when experimental or cutting-edge functionality is added to certain LLMs and it doesn't yet make sense to standardize the functionality to the universal `LLMContext`. It just ensures that development on those LLM services isn't slowed down by the need to add to submit a change to a universal protocol. 

This, of course, is only useful when using that LLM by itself. As soon as you (the developer) want to add LLM switching or failover, you'll need to ensure that no `LLMSpecificMessage`s are in your shared context—at that point, you need to consider whether you can excise them or try to formally "promote" them into the universal format.

Does that make sense?

src/pipecat/adapters/services/gemini_adapter.py
kompfner: In attempting to convert all invocation params types (e.g. `GeminiLLMInvocationParams`) from `TypedDict`s to `@dataclass`, I now recall why I went with `TypedDict` in the first place: it makes it trivial to use the `*InvocationParams` types as a bag of parameters to pass directly to an LLM. For example, in `build_chat_completion_params()`:

```python
params.update(params_from_context) # params_from_context is a TypedDict
```

src/pipecat/services/google/llm_openai.py
```diff
@@ -40,6 +40,10 @@ class GoogleLLMOpenAIBetaService(OpenAILLMService):
     Note: This service includes a workaround for a Google API bug where function
     call indices may be incorrectly set to None, resulting in empty function names.
 
+    .. deprecated:: 0.0.81
```
kompfner: Note to self: I'm assuming this PR will make it in for 0.0.81. Otherwise, update this!

src/pipecat/adapters/services/gemini_adapter.py
```diff
@@ -39,3 +90,215 @@ def to_provider_tools_format(self, tools_schema: ToolsSchema) -> List[Dict[str,
...
+        messages = []
+
+        # Process each message, preserving Google-formatted messages and converting others
+        for message in universal_context_messages:
+            if isinstance(message, LLMSpecificMessage):
+                # Assume that LLMSpecificMessage wraps a message in Google format
+                messages.append(message.message)
+                continue
+
+            # Convert standard format to Google format
```
kompfner, line 163: Note: in [a PR that builds on top of this one](https://github.com/pipecat-ai/pipecat/pull/2485), I've made a small (but significant) [change here, which you should read about](https://github.com/pipecat-ai/pipecat/pull/2485/commits/657a72670f4797b7b277f658ad11999c62364116)

markbackman (approved):
LGTM!

src/pipecat/services/google/llm_openai.py
kompfner: Update to 0.0.82. This'll make the next release.

src/pipecat/pipeline/service_switcher.py
```diff
@@ -6,13 +6,22 @@
...
 from typing import Any, Generic, List, Optional, Type, TypeVar
 
+from pipecat.frames.frames import ControlFrame, Frame
 from pipecat.pipeline.parallel_pipeline import ParallelPipeline
 from pipecat.processors.filters.function_filter import FunctionFilter
 from pipecat.processors.frame_processor import FrameDirection, FrameProcessor
 
 
+@dataclass
+class ServiceSwitcherFrame(ControlFrame):
```
markbackman: Should this be defined here or in pipecat.frames.frames.py? Maybe the latter?

src/pipecat/pipeline/service_switcher.py
```diff
@@ -34,6 +43,28 @@ def is_active(self, service: FrameProcessor) -> bool:
...
+
+        Args:
+            frame: The frame to handle.
+            direction: The direction of the frame (upstream or downstream).
+        """
+        raise NotImplementedError("Subclasses must implement this method.")
+
+
+@dataclass
+class ManuallySwitchServiceFrame(ServiceSwitcherFrame):
```
markbackman: Same question about where this frame resides.

src/pipecat/pipeline/service_switcher.py
kompfner: I'm never totally sure where to put frame definitions. For the most part they're in `frames.py` as you point out. But there are also cases where really really specific frames are defined near the thing that uses them.

PR #2442, opened 2025-08-13
Mb/retry completion

markbackman:

I previously added this retry logic to OpenAILLMService and its subclasses ([PR](https://github.com/pipecat-ai/pipecat/pull/2387)). This adds similar support to `AnthropicLLMService` and `AWSBedrockLLMService`.

The remaining service is `GoogleLLMService`. In looking at its client, the response from the request returns immediately with a stream. Given that, the same approach will not work. If we want to add this feature to `GoogleLLMService`, then we'll have to do something more complex.


filipi87 (approved):
LGTM!

PR #2446, opened 2025-08-14
Add support to smallWebRTCTransport for receiving screenshare videos

mattieruth:
This PR goes hand-in-hand with https://github.com/pipecat-ai/pipecat-client-web-transports/pull/56

### DIFF
--- a/src/pipecat/transports/network/small_webrtc.py
+++ b/src/pipecat/transports/network/small_webrtc.py
@@ -679,25 +679,28 @@ class SmallWebRTCInputTransport(BaseInputTransport):
             # Start screen video reception if it's not already running
             self._receive_screen_video_task = self.create_task(self._receive_video("screenVideo"))
 
-    async def capture_participant_video(
+    async def capture_participant_media(
         self,
-        video_source: str = "camera",
+        source: str = "camera",
     ):
-        """Capture video from a specific participant.
+        """Capture media from a specific participant.
 
         Args:
-            video_source: Video source to capture from.
+            source: Media source to capture from. ("camera", "microphone", or "screenVideo")
         """
         # If we're not already receiving video, try to get a frame now
         if (
-            video_source == "camera"
-            and not self._receive_video_task
-            and self._params.video_in_enabled
+            source == "microphone"
+            and not self._receive_audio_task
+            and self._params.audio_in_enabled
         ):
+            # Start audio reception if it's not already running
+            self._receive_audio_task = self.create_task(self._receive_audio())
+        elif source == "camera" and not self._receive_video_task and self._params.video_in_enabled:
             # Start video reception if it's not already running
             self._receive_video_task = self.create_task(self._receive_video("camera"))
         elif (
-            video_source == "screenVideo"
+            source == "screenVideo"
             and not self._receive_screen_video_task
             and self._params.video_in_enabled
         ):
@@ -892,15 +895,34 @@ class SmallWebRTCTransport(BaseTransport):
 
     async def capture_participant_video(
         self,
+        participant_id: str = None,
+        framerate: int = 30,
         video_source: str = "camera",
+        color_format: str = "RGB",
     ):
         """Capture video from a specific participant.
 
         Args:
-            participant_id: ID of the participant to capture video from.
-            framerate: Desired framerate for video capture.
+            participant_id: Unused parameter, kept for compatibility.
+            framerate: Unused parameter, kept for compatibility.
             video_source: Video source to capture from.
-            color_format: Color format for video frames.
+            color_format: Unused parameter, kept for compatibility.
         """
         if self._input:
-            await self._input.capture_participant_video(video_source)
+            await self._input.capture_participant_media(source=video_source)
+
+    async def capture_participant_audio(
+        self,
+        participant_id: str = None,
+        audio_source: str = "microphone",
+        sample_rate: int = 16000,
+    ):
+        """Capture audio from a specific participant.
+
+        Args:
+            participant_id: Unused parameter, kept for compatibility.
+            audio_source: Audio source to capture from. (currently, "microphone" is the only supported option)
+            sample_rate: Unused parameter, kept for compatibility.
+        """
+        if self._input:
+            await self._input.capture_participant_media(source=audio_source)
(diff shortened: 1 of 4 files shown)


src/pipecat/transports/network/small_webrtc.py
```diff
@@ -547,11 +559,11 @@ async def start(self, frame: StartFrame):
 
         await self._client.setup(self._params, frame)
         await self._client.connect()
+        await self.set_transport_ready(frame)
```
filipi87, line 567: In the other transports, we’ve typically placed it last, once everything inside the transport is set up. Is there a reason why we need it here in this case ?

src/pipecat/transports/network/small_webrtc.py
```diff
@@ -547,11 +559,11 @@ async def start(self, frame: StartFrame):
...
         if not self._receive_audio_task and self._params.audio_in_enabled:
             self._receive_audio_task = self.create_task(self._receive_audio())
         if not self._receive_video_task and self._params.video_in_enabled:
-            self._receive_video_task = self.create_task(self._receive_video())
-        await self.set_transport_ready(frame)
+            self._receive_video_task = self.create_task(self._receive_video("camera"))
```
filipi87: It would be nice to create some constants for the `camera` and `screenVideo` sources.

src/pipecat/transports/network/small_webrtc.py
```diff
@@ -840,3 +897,37 @@ async def _on_client_connected(self, webrtc_connection):
     async def _on_client_disconnected(self, webrtc_connection):
         """Handle client disconnection events."""
         await self._call_event_handler("on_client_disconnected", webrtc_connection)
+
+    async def capture_participant_video(
+        self,
+        participant_id: str = None,
+        framerate: int = 30,
+        video_source: str = "camera",
+        color_format: str = "RGB",
```
filipi87: Inside the runner, we already check the transport before invoking it, so does it make sense to keep these parameters here, considering that they are never actually used ?
And it does feels weird to have a `participant_id` here. 😅
What do you think ? Or is there anything else blocking us from remove these fields ?
```
elif isinstance(transport, SmallWebRTCTransport):
            await transport.capture_participant_video(video_source="camera")
```

filipi87: I guess the `framerate` would probably be nice to support in the future.

src/pipecat/transports/network/small_webrtc.py
```diff
@@ -840,3 +897,37 @@ async def _on_client_connected(self, webrtc_connection):
...
+            color_format: Unused parameter, kept for compatibility.
+        """
+        if self._input:
+            await self._input.capture_participant_media(source=video_source)
+
+    async def capture_participant_audio(
+        self,
+        participant_id: str = None,
+        audio_source: str = "microphone",
+        sample_rate: int = 16000,
```
filipi87: Same here. It would be nice to remove the unused parameters.

src/pipecat/transports/network/webrtc_connection.py
```diff
@@ -39,6 +39,7 @@
 SIGNALLING_TYPE = "signalling"
 AUDIO_TRANSCEIVER_INDEX = 0
 VIDEO_TRANSCEIVER_INDEX = 1
+SCREEN_VIDEO_TRANSCEIVER_INDEX = 2
```
filipi87, line 42: I was looking at the Pipecat docs:
- https://docs.pipecat.ai/server/services/transport/small-webrtc

I thought we explained the transceiver order there, but it seems we don’t. 🤕 
Probably something I should improve in the future.

src/pipecat/transports/network/webrtc_connection.py
```diff
@@ -343,6 +347,9 @@ async def connect(self):
             video_input_track = self.video_input_track()
             if video_input_track:
                 await self.video_input_track().discard_old_frames()
+            screen_video_input_track = self.screen_video_input_track()
+            if screen_video_input_track:
+                await self.screen_video_input_track().discard_old_frames()
             self.ask_to_renegotiate()
```
filipi87: Since you are already making some changes here, I included this `self.ask_to_renegotiate()` because sometimes the WebRTC connection was established before the bot was ready for receiving video. In those cases, we could lose some video frames, and it would take a couple of seconds before we received a key frame to finally start displaying video. So, this prevents this issue.

All this explanation just to say: I believe we should only invoke this if we have a `video_input_track` or `screen_video_input_track`. For audio-only, we shouldn’t need to invoke it.

src/pipecat/transports/network/webrtc_connection.py
```diff
@@ -94,14 +95,16 @@ class SmallWebRTCTrack:
...
+    def __init__(self, track: MediaStreamTrack, index: int):
         """Initialize the WebRTC track wrapper.
 
         Args:
             track: The underlying MediaStreamTrack to wrap.
+            index: The index of the track in the transceiver (0 for mic, 1 for cam, 2 for screen)
         """
         self._track = track
         self._enabled = True
+        self.source_index = index
```
filipi87: Are we using `self.source_index` anywhere ?

src/pipecat/transports/network/small_webrtc.py
mattieruth: For posterity sake: initially, i was trying to match the api of daily so that I could update the `util` usage to just:

```
if isinstance(transport, DailyTransport) or isinstance(transport, SmallWebRTC):
```

and treat them the same, but then i think i kept hitting issues and backed off that, realizing they simply weren't the same.

And so I agree, I think i can just remove all the unused vars here. We can add back `framerate` another time.

src/pipecat/transports/network/webrtc_connection.py
mattieruth: i bet i was using it for debug logging and forgot to remove it when i removed the logs. It _is_ sort of handy for that, but i'll remove it.

src/pipecat/transports/network/webrtc_connection.py
```diff
@@ -378,7 +388,11 @@ async def delayed_task():
     def force_transceivers_to_send_recv(self):
         """Force all transceivers to bidirectional send/receive mode."""
         for transceiver in self._pc.getTransceivers():
-            transceiver.direction = "sendrecv"
+            # For now, we only support sendrecv for camera audio and video (the first two transceivers)
```
filipi87, line 391: Makes sense. 👍

filipi87 (approved):
Looks good. 🚀

src/pipecat/transports/network/small_webrtc.py
mattieruth: i honestly can't remember now why i changed this order but i'm pretty sure it mattered 🤔

PR #2451, opened 2025-08-14
AudioBufferProcessor: fix overlap when buffer size is set

aconchillo:

This PR makes sure we reset recording (buffers and timers) when needed.


src/pipecat/processors/audio/audio_buffer_processor.py
```diff
@@ -288,8 +290,6 @@ async def _call_on_audio_data_handler(self):
             self._num_channels,
         )
 
-        self._reset_audio_buffers()
```
aconchillo, line 291: The reason for not adding `reset_recording` here is for separation of concerns. we call it manually wherever is needed.

PR #2452, opened 2025-08-14
Runner: set handle_sigint to True for Daily direct

aconchillo:

We set default `handle_sigint` to False but Daily direct still needs to handle SIGINT properly.


markbackman (approved):
LGTM

PR #2456, opened 2025-08-16
Only set last_frame_time when handling OutputAudioRawFrame

a6kme:
We don't want to set `last_frame_time` on other frames like `HeartBeatFrame`, `LLMGeneratedTextFrame`, `InterruptionFrames` so that we can calculate `diff_time` and compare it against `vad_stop_secs` properly


aconchillo:
Good catch! LGTM!

PR #2457, opened 2025-08-16
Remove VAD check when using mixer

a6kme:
When we are using a mixer, we should send `BotStoppedSpeaking` frame as soon as we are done playing all the audio generated by TTS in the current turn.

### DIFF
--- a/src/pipecat/transports/base_output.py
+++ b/src/pipecat/transports/base_output.py
@@ -634,7 +634,6 @@ class BaseOutputTransport(FrameProcessor):
                         await self._bot_stopped_speaking()
 
             async def with_mixer(vad_stop_secs: float) -> AsyncGenerator[Frame, None]:
-                last_frame_time = 0
                 silence = b"\x00" * self._audio_chunk_size
                 while True:
                     try:
@@ -642,14 +641,13 @@ class BaseOutputTransport(FrameProcessor):
                         self._transport.reset_watchdog()
                         if isinstance(frame, OutputAudioRawFrame):
                             frame.audio = await self._mixer.mix(frame.audio)
-                        last_frame_time = time.time()
                         yield frame
                     except asyncio.QueueEmpty:
                         self._transport.reset_watchdog()
-                        # Notify the bot stopped speaking upstream if necessary.
-                        diff_time = time.time() - last_frame_time
-                        if diff_time > vad_stop_secs:
-                            await self._bot_stopped_speaking()
+
+                        # TTS Generated audio is finished. Lets notify pipeline
+                        # with BotStoppedSpeaking
+                        await self._bot_stopped_speaking()
                         # Generate an audio frame with only the mixer's part.
                         frame = OutputAudioRawFrame(
                             audio=await self._mixer.mix(silence),


a6kme:
@markbackman  / @aconchillo - I am not sure if this makes super sense. Would love to hear your thoughts on this. Thanks!

markbackman:
@a6kme can you explain what issue you're trying to solve?

The BotStoppedSpeakingFrame is emitted once all of the bot's speech has finished. Since there's no correlation between TTS generation requests and the returned audio chunks, there's no way to definitively know once the bot speaking will stop. To solve for this, we use a VAD parameter, similar to input, to determine when enough silence has occurred to state—with enough confidence—that the bot is finished speaking.

I believe we want to keep this algorithm in place. Removing this would emit more BotStoppedSpeakingFrame than desired. I'm inclined to close this, but want to hear your input first.

a6kme:
@markbackman - I can't say about other TTS Services, but when I use ElevenLabs, in a single turn, the delay between two TTS Segments are always more than `BOT_VAD_STOP_SECS`, which is 0.35 seconds by default. So, for my case, we are not emitting more number of `BotStoppedSpeakingFrame` than we were emitting earlier.

We also add around 0.5 seconds worth of [silence in the end when we are using ElevenLabs](https://github.com/pipecat-ai/pipecat/blob/8fa44863fbcfaaf9a3cdec141acf5cafe6a36d00/src/pipecat/services/tts_service.py#L844C25-L844C26). 

If I let both be there, then my `STTMuteFilter` is unmuted with a delay of around 0.85 seconds, which can cause issues when the user is trying to speak something immediately after the (actually) bot stops speaking. The system would know only after 0.85 seconds that we need to unmute the `STT`.

markbackman:
> @markbackman - I can't say about other TTS Services, but when I use ElevenLabs, in a single turn, the delay between two TTS Segments are always more than `BOT_VAD_STOP_SECS`, which is 0.35 seconds by default. So, for my case, we are not emitting more number of `BotStoppedSpeakingFrame` than we were emitting earlier.
> 
> We also add around 0.5 seconds worth of [silence in the end when we are using ElevenLabs](https://github.com/pipecat-ai/pipecat/blob/8fa44863fbcfaaf9a3cdec141acf5cafe6a36d00/src/pipecat/services/tts_service.py#L844C25-L844C26).
> 
> If I let both be there, then my `STTMuteFilter` is unmuted with a delay of around 0.85 seconds, which can cause issues when the user is trying to speak something immediately after the (actually) bot stops speaking. The system would know only after 0.85 seconds that we need to unmute the `STT`.

What is the latency between responses? You should be able to tell from the Bot stopped speaking to bot started speaking times.

Also, is this `ElevenLabsTTSService` or `ElevenLabsHttpTTSService`?

a6kme:
After making the change that I have proposed in the PR for us, in a single turn, I get a bot started speaking frame after bot stopped speaking frame ranging from 100 ms - 500 ms. I am not sure if this can be termed as "latency between responses". Its just the way TTSService are generating response. Example: We deliberately [add pause in a single TTS generation by ](https://help.elevenlabs.io/hc/en-us/articles/13416374683665-How-can-I-add-pauses)putting hyphens (`-`) in the text, so that the synthesised speech can sound more realistic. 

Current design choice that I have made is a tradeoff for us, since we are getting multiple bot started speaking / bot stopped speaking in the same turn. But this solves the problem, that if a user speaks something really small sentence immediately after bot stops speaking, like "yeah" or "no", I am able to unmute my STTMuteFilter right in time to capture those responses.

This is `ElevenLabsTTSService`.

markbackman:
Ok, so it sounds like you're deliberately adding pauses which causes the bot stopped/started speaking frames. Got it. Given that, I don't think we should remove any logic to accommodate. We could make the Bot stop_secs value controllable. I would much rather prefer that change over this one. I'm going to close this PR because we don't want to fundamentally change the way this works.

PR #2458, opened 2025-08-16
feat: add semantic caching support with cachelm integration

devanmolsharma:
This PR introduces semantic caching support by integrating with CacheLM, enabling smarter and faster retrieval of cached responses. Instead of relying solely on exact-match lookups, CacheLM leverages embeddings and semantic similarity to improve cache hit rates.


markbackman:
This is very cool, but out of scope for the main Pipecat repo. We now have community integrations, where this would be a better fit:

Feel free to submit there if you're interested.

PR #2462, opened 2025-08-18
Smart Turn V3 support

marcus-daily:

### DIFF
--- a/src/pipecat/audio/turn/smart_turn/local_smart_turn_v3.py
+++ b/src/pipecat/audio/turn/smart_turn/local_smart_turn_v3.py
@@ -18,22 +18,14 @@ from loguru import logger
 from pipecat.audio.turn.smart_turn.base_smart_turn import BaseSmartTurn
 
 try:
-    from transformers import WhisperFeatureExtractor
     import onnxruntime as ort
+    from transformers import WhisperFeatureExtractor
 except ModuleNotFoundError as e:
     logger.error(f"Exception: {e}")
     logger.error(
         "In order to use LocalSmartTurnAnalyzerV3, you need to `pip install pipecat-ai[local-smart-turn-v3]`."
     )
     raise Exception(f"Missing module: {e}")
-
-
-def build_session(onnx_path):
-    so = ort.SessionOptions()
-    so.execution_mode = ort.ExecutionMode.ORT_SEQUENTIAL
-    so.inter_op_num_threads = 1
-    so.graph_optimization_level = ort.GraphOptimizationLevel.ORT_ENABLE_ALL
-    return ort.InferenceSession(onnx_path, sess_options=so)
 
 
 class LocalSmartTurnAnalyzerV3(BaseSmartTurn):
@@ -57,8 +49,13 @@ class LocalSmartTurnAnalyzerV3(BaseSmartTurn):
 
         logger.debug("Loading Local Smart Turn v3 model...")
 
+        so = ort.SessionOptions()
+        so.execution_mode = ort.ExecutionMode.ORT_SEQUENTIAL
+        so.inter_op_num_threads = 1
+        so.graph_optimization_level = ort.GraphOptimizationLevel.ORT_ENABLE_ALL
+
         self._feature_extractor = WhisperFeatureExtractor(chunk_length=8)
-        self._session = build_session(smart_turn_model_path)
+        self._session = ort.InferenceSession(smart_turn_model_path, sess_options=so)
 
         logger.debug("Loaded Local Smart Turn v3")
 


pyproject.toml
```diff
@@ -91,6 +91,7 @@ runner = [ "python-dotenv>=1.0.0,<2.0.0", "uvicorn>=0.32.0,<1.0.0", "fastapi>=0.
 sambanova = []
 sentry = [ "sentry-sdk~=2.23.1" ]
 local-smart-turn = [ "coremltools>=8.0", "transformers", "torch>=2.5.0,<3", "torchaudio>=2.5.0,<3" ]
+local-smart-turn-v3 = [ "transformers", "torch>=2.5.0,<3", "torchaudio>=2.5.0,<3", "onnxruntime" ]
```
markbackman: The `silero` optional dependency also uses `onnxruntime`. It's currently pinned to `"onnxruntime~=1.20.1"`, but we should probably:
1. Relax that version to `"onnxruntime>=1.20.1, <2"`
2. Match the same version for `local-smart-turn-v3`

The other question is, should we keep `local-smart-turn` as a separate optional dependency? I'm assuming one is GPU accelerated the other CPU. Is that right?

marcus-daily: Thanks for the review! That makes sense regarding the `onnxruntime` version, and 1.20.1 or above will work for the opset used in the current model build.

> The other question is, should we keep local-smart-turn as a separate optional dependency? I'm assuming one is GPU accelerated the other CPU. Is that right?

Yes, the `local-smart-turn-v3` dependency uses CPU inference by default using `onnxruntime`, which isn't necessary for v2 or v1. I'm hoping that by release time, `torch`, `transformers`, and all the gigabytes of NVIDIA dependencies won't be necessary for v3, in which case the dependencies will look like this:

```
local-smart-turn = [ "coremltools>=8.0", "transformers", "torch>=2.5.0,<3", "torchaudio>=2.5.0,<3" ]
local-smart-turn-v3 = [ "onnxruntime>=1.20.1, <2" ]
```

In which case I think it makes sense to keep them separate.

markbackman: Makes sense! Thanks for the details.

src/pipecat/audio/turn/smart_turn/local_smart_turn_v3.py
```diff
@@ -0,0 +1,106 @@
...
+
+
+class LocalSmartTurnAnalyzerV3(BaseSmartTurn):
+    """Local turn analyzer using the smart-turn-v3 ONNX model.
+
+    Provides end-of-turn detection using locally-stored ONNX model,
+    enabling offline operation without network dependencies.
+    """
+
+    def __init__(self, *, smart_turn_model_path: str, **kwargs):
```
markbackman, line 38: Do we want to bundled the onnx model with Pipecat?

We do that for Silero, storing in silero_vad.onnx.

We could bundle the onnx model and then allow the ability to override with a different model. WDYT?

LackesLab:
Awesome work! I love the new turn detection—it's a significant improvement over the previous version.

@markbackman , I recently noticed that managing dependencies via uv leads to the installation of the `pytorch-gpu` package, which significantly increases the size of the `.venv` folder.

Is there a way to control dependency installation so we can choose to install only the CPU version of PyTorch for a slimmer setup? 

Or do we need to manually replace the torch library ourselves after the environment has been set up?

markbackman:
> Awesome work! I love the new turn detection—it's a significant improvement over the previous version.
> 
> @markbackman , I recently noticed that managing dependencies via uv leads to the installation of the `pytorch-gpu` package, which significantly increases the size of the `.venv` folder.
> 
> Is there a way to control dependency installation so we can choose to install only the CPU version of PyTorch for a slimmer setup?
> 
> Or do we need to manually replace the torch library ourselves after the environment has been set up?

cc @marcus-daily

marcus-daily:
@LackesLab @markbackman I've been able to remove the torch/torchaudio dependencies entirely. `transformers` is still necessary but isn't that big.

https://github.com/pipecat-ai/smart-turn/commit/c05b0aab5dee2fa0c168040523824207b437a673

I'll open a PR add add the same changes to Pipecat shortly.

marcus-daily:
PR here: https://github.com/pipecat-ai/pipecat/pull/2664

marcus-daily:
For future reference, the above PR saves about 6.5GB worth of dependencies in `venv`, bringing the overall size down to 880MB. (compared to 740MB without Smart Turn).

The remaining 140MB is from the `transformers` dependency, which we could remove (by vendoring the audio preprocessing code) but that's probably not a huge priority.

PR #2464, opened 2025-08-18
various frame processor updates

aconchillo:

This PR provides a few improvements:

- Use direct mode in both `Pipeline` and `ParallelPipeline`. Improves performance.
- Avoid creating tasks for every frame and sub-pipeline in `ParallelPipeline`. Improves performance.
- Added new `FrameProcessor` methods: `processors()` and `entry_processors()`
- `Pipeline` class now allows passing source and sink elements. This reduces code complexity in, for example, `ParallelPipeline` and allows returning the single sub-pipeline processors when calling `processors()`.
- Removes `set_parent()` and `get_parent()` they were just broken an unused.
- Adds new observer method `on_process_frame`.

More improvements to `ParallelPipeline`s coming soon.

### DIFF
--- a/src/pipecat/pipeline/parallel_pipeline.py
+++ b/src/pipecat/pipeline/parallel_pipeline.py
@@ -11,106 +11,15 @@ sub-pipelines concurrently, with coordination for system frames and proper
 handling of pipeline lifecycle events.
 """
 
-import asyncio
 from itertools import chain
-from typing import Awaitable, Callable, Dict, List
+from typing import Dict, List
 
 from loguru import logger
 
-from pipecat.frames.frames import (
-    CancelFrame,
-    EndFrame,
-    Frame,
-    StartFrame,
-    StartInterruptionFrame,
-    SystemFrame,
-)
+from pipecat.frames.frames import EndFrame, Frame, StartFrame
 from pipecat.pipeline.base_pipeline import BasePipeline
-from pipecat.pipeline.pipeline import Pipeline
+from pipecat.pipeline.pipeline import Pipeline, PipelineSink, PipelineSource
 from pipecat.processors.frame_processor import FrameDirection, FrameProcessor, FrameProcessorSetup
-from pipecat.utils.asyncio.watchdog_queue import WatchdogQueue
-
-
-class ParallelPipelineSource(FrameProcessor):
-    """Source processor for parallel pipeline branches.
-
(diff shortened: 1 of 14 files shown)


src/pipecat/pipeline/parallel_pipeline.py
```diff
@@ -132,28 +41,69 @@ def __init__(self, *args):
...
+
+        Entry processors are the first non-compound processors in a pipeline.
+        For parallel pipelines, this includes the first non-compound processor
+        from each parallel branch. Non-compound processors will return an empty
+        list.
+
+        Returns:
+            The list of the first non-compound processors.
+        """
+        return self._pipelines
```
filipi87, line 105: Shouldn't this be `self._sources` ? If so, we also need to create [self._sources](url) and store the sources.

src/pipecat/pipeline/parallel_pipeline.py
```diff
@@ -219,79 +140,15 @@ async def process_frame(self, frame: Frame, direction: FrameDirection):
...
-    async def _handle_interruption(self):
-        """Handle interruption by cancelling tasks, draining queues, and restarting."""
-        await self._cancel()
-        await self._drain_queues()
-        await self._create_tasks()
+        # Parallel pipeline synchronized frames.
+        if isinstance(frame, (StartFrame, EndFrame)):
+            self._frame_counter[frame.id] = len(self._pipelines)
+            await self.pause_processing_system_frames()
+            await self.pause_processing_frames()
```
filipi87, line 147: Cool, that’s a much cleaner approach.

src/pipecat/pipeline/parallel_pipeline.py
aconchillo: Actually, this is fine. I believe what's wrong is the documentation. It's the list of first processors. Now with the simplification in ParallelPipeline we actually have Pipelines which are the first processors. I'll update the documentation.

filipi87: I see. What about `SyncParallelPipeline` ? Should we return `self._sources` there, or should we also return `self._pipelines` ?

Also, I don’t see where `entry_processors` are used in the code, so I’m wondering why they were included. Just trying to understand why we need them. 🙂

aconchillo: Documentation updated

PR #2465, opened 2025-08-18
Changing heygen log level to trace.

filipi87:
Changing heygen log level to trace.

### DIFF
--- a/src/pipecat/services/heygen/client.py
+++ b/src/pipecat/services/heygen/client.py
@@ -248,7 +248,7 @@ class HeyGenClient:
         if event_type == "agent.state":
             logger.debug(f"HeyGenClient ws received agent status: {event}")
         else:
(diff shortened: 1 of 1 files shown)


src/pipecat/services/heygen/client.py
```diff
@@ -248,7 +248,7 @@ async def _handle_ws_server_event(self, event: dict) -> None:
         if event_type == "agent.state":
             logger.debug(f"HeyGenClient ws received agent status: {event}")
         else:
-            logger.error(f"HeyGenClient ws received unknown event: {event_type}")
+            logger.trace(f"HeyGenClient ws received unknown event: {event_type}")
```
filipi87, line 251: This shouldn’t have been an error,maybe debug. But considering the amount of unexpected messages we’re getting from HeyGen, I’m going to change it to trace for now.

PR #2467, opened 2025-08-19
Runner support for SmallWebRTC Prebuilt configuration object 

jptaylor:


src/pipecat/runner/run.py
```diff
@@ -177,7 +182,12 @@ def _create_server_app(
     return app
 
 
-def _setup_webrtc_routes(app: FastAPI, esp32_mode: bool = False, host: str = "localhost"):
+def _setup_webrtc_routes(
+    app: FastAPI,
+    esp32_mode: bool = False,
+    host: str = "localhost",
+    prebuilt_ui_config: SmallWebRTCPrebuiltArguments = None,
```
filipi87, line 189: I guess in this case it should be Optional.
```suggestion
    prebuilt_ui_config: Optional[SmallWebRTCPrebuiltArguments] = None,
```

src/pipecat/runner/run.py
```diff
@@ -151,7 +152,11 @@ async def _run_telephony_bot(websocket: WebSocket):
 
 
 def _create_server_app(
-    transport_type: str, host: str = "localhost", proxy: str = None, esp32_mode: bool = False
+    transport_type: str,
+    host: str = "localhost",
+    proxy: str = None,
+    esp32_mode: bool = False,
+    prebuilt_ui_config: SmallWebRTCPrebuiltArguments = None,
```
filipi87, line 159: I guess in this case it should be Optional.
```suggestion
    prebuilt_ui_config: Optional[SmallWebRTCPrebuiltArguments] = None,
```

src/pipecat/runner/run.py
```diff
@@ -435,7 +445,7 @@ def _validate_and_clean_proxy(proxy: str) -> str:
     return proxy
 
 
-def main():
+def main(prebuilt_ui_config: SmallWebRTCPrebuiltArguments = None):
```
filipi87, line 448: Same here.
```suggestion
def main(prebuilt_ui_config: Optional[SmallWebRTCPrebuiltArguments] = None):
```

src/pipecat/runner/types.py
```diff
@@ -65,3 +65,67 @@ class SmallWebRTCRunnerArguments(RunnerArguments):
...
+    Parameters:
+        no_rtvi: Disables RTVI related functionality. Default: False
+        server_rtvi_version: Specifies the RTVI version in use by the server. Default: None
+        no_user_audio: Disables user audio input entirely. Default: False
+        no_user_video: Disables user video input entirely. Default: False
+        user_video_enabled: Enables user video input. Default: False
+        user_audio_enabled: Enables user audio input. Default: True
+        no_audio_output: Disables audio output for the bot. Default: False
+        no_bot_audio: Disables audio visualization for the bot. Default: False
+        no_bot_video: Disables video visualization for the bot. Default: True
```
filipi87, line 83: @jptaylor, the options feels a little bit confusing sometimes. Some options are phrased as negatives (`no_user_audio`) while others are phrased as positives (`user_audio_enabled`). 

And it would be nice to make it clear regarding what is only for "visualization".

So, I would propose something like this for example:

```
user_audio_enabled: true     # User can speak
user_video_enabled: false    # User camera off by default
bot_audio_enabled: true      # Bot can output audio
bot_video_enabled: false     # Bot video output disabled
bot_audio_visuals_enabled: false  # No waveform or visualization
```

filipi87: And the same for the other options, instead of using no_something. Like `no_rtvi`, use it as `rtvi_enabled`.

filipi87: It feels like it would be easier for me to understand, but let me know what you think.

filipi87 (commented):
@jptaylor I think this approach makes a lot of sense. I left a couple of comments, but it gives us the flexibility to provide different settings for each example.

I have also tagged Mark, who has much more knowledge about the runners.

markbackman:
Generally looks good. +1 to Filipi's comments about making the config `Optional`.

Also, two other things:
1. Add a changelog entry
2. Fix formatting issues. You can install the pre-commit hook (`uv run pre-commit install`) or use the fix-ruff util (`./scripts/fix-ruff.sh`).

markbackman:
@jptaylor is this still relevant?

PR #2470, opened 2025-08-19
Update ElevenLabsTTSService: reconnect on model and language changes

markbackman:


src/pipecat/services/tts_service.py
```diff
@@ -258,7 +258,7 @@ async def _update_settings(self, settings: Mapping[str, Any]):
                     self._settings[key] = self.language_to_service_language(value)
             elif key == "model":
                 self.set_model_name(value)
-            elif key == "voice":
+            elif key == "voice" or key == "voice_id":
```
markbackman, line 261: Also, making the voice id key more flexible, as most services use `voice_id` but TTSUpdateSettingsFrame expects `voice`.

PR #2471, opened 2025-08-19
Add foundational 13j for Azure STT

markbackman:


uv.lock
filipi87, line 1: Just double checking that these changes are also intentional.

filipi87 (approved):
LGTM!

uv.lock
markbackman: Yup. The lock file gets pulled in automatically.

PR #2473, opened 2025-08-19
adds support to Azure STT for creating InterimTranscriptFrames 

TheNotary:

I was having trouble getting speech to come through my pipeline in realtime, it would only come through after the recording was finished.  After a lot of debugging, I realized there are two hooks for getting transcripts from Azure, one called `recognized` which fires at the completion of the stt process, and another called `recognizing` which fires after a chunk of the recording has been recognized.  

This PR just adds a hook, and follows the coding patterns presented in the `_on_handle_recognized` handler.


aconchillo:
Thank you! LGTM!

PR #2474, opened 2025-08-20
Add UserStarted/StoppedSpeakingFrames to idle_timeout_frames

markbackman:

@aconchillo we talked about this the other day. I think `UserStartedSpeakingFrame`, `UserStoppedSpeakingFrame` are good signals for user activity. Did you have anything else in mind?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- Added new frames to the `idle_timeout_frames` arg: `TranscriptionFrame`,
+  `InterimTranscriptionFrame`, `UserStartedSpeakingFrame`, and
+  `UserStoppedSpeakingFrame`. These additions serve as indicators of user
+  activity in the pipeline idle detection logic.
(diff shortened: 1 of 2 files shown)


aconchillo (approved):

aconchillo:
What happens if the user talks for longer than the idle timeout? I guess we would timeout...

markbackman:
> What happens if the user talks for longer than the idle timeout? I guess we would timeout...

Someone asked that in Discord. We could opt to look at TranscriptionFrames instead, which would be a more direct signal.

Maybe user speaking frames and TranscriptionFrames. That would certainly indicate speech is happening. It's a parallel to the bot activity monitoring.

WDYT?

aconchillo:
> > What happens if the user talks for longer than the idle timeout? I guess we would timeout...
> 
> Someone asked that in Discord. We could opt to look at TranscriptionFrames instead, which would be a more direct signal.
> 
> Maybe user speaking frames and TranscriptionFrames. That would certainly indicate speech is happening. It's a parallel to the bot activity monitoring.
> 
> WDYT?

Yes, probably interim transcriptions and transcriptions.

PR #2476, opened 2025-08-20
add asyncio timeout() context manager

aconchillo:
This is an implementation of Python's `asyncio.timeout()` which was introduced in 3.11. We strongly encourage the use of the following approach instead of `asyncio.wait_for()`.

    async with timeout(delay):
      # Your async code

In Python 3.12, `asyncio.wait_for()` is implemented in terms of `asyncio.timeout()` which fixed a bunch of issues. However, this was never backported (because of the lack of `async.timeout()`) and there are still many remainig issues, specially in Python 3.10, in `async.wait_for()`.

See https://github.com/python/cpython/pull/98518


src/pipecat/utils/asyncio/task_manager.py
```diff
@@ -20,6 +20,8 @@
 
 from loguru import logger
 
+import pipecat.utils.asyncio.timeout
```
filipi87: Why in this case are you not only using from `pipecat.utils.asyncio.timeout import wait_for`, like in the other places ?

src/pipecat/utils/asyncio/task_manager.py
```diff
@@ -285,7 +287,8 @@ async def wait_for_task(self, task: asyncio.Task, timeout: Optional[float] = Non
         name = task.get_name()
         try:
             if timeout:
-                await asyncio.wait_for(task, timeout=timeout)
+                async with pipecat.utils.asyncio.timeout.timeout(timeout):
+                    await task
```
filipi87: Same here.

filipi87: And in the other places of this class.

src/pipecat/utils/asyncio/timeout.py
```diff
@@ -0,0 +1,28 @@
...
+
+"""Implementation of Python's `asyncio.wait_for()`.
+
+This module uses `wait_for2` package to implement `asyncio.wait_for()` for
+Python < 3.12.
+
+In Python 3.12, `asyncio.wait_for()` is implemented in terms of
+`asyncio.timeout()` which fixed a bunch of issues. However, this was never
+backported (because of the lack of `async.timeout()`) and there are still many
+remainig issues, specially in Python 3.10, in `async.wait_for()`.
```
filipi87, line 15: Amazing finding 🔥

src/pipecat/utils/asyncio/task_manager.py
aconchillo: Ooops! This was from a previous commit. Updated!

src/pipecat/utils/asyncio/task_manager.py
aconchillo: Yep, previous commit that I forgot to update.

PR #2477, opened 2025-08-20
Introduce an affordance to `LLMService` for generating a summary of a…

kompfner:
… conversation directly (i.e. without going through the pipeline).

This abstraction will allow us to update Pipecat Flows to avoid reaching into LLM service internals to generate summaries.

In addition to being a helpful refactor to remove a fragile part of Pipecat Flows (which, it turns out, [had already broken without us noticing!](https://github.com/pipecat-ai/pipecat/pull/2477#discussion_r2289114976)), this change helps set the stage for supporting the upcoming `LLMSwitcher`, where the “active” LLM will only be determined at runtime—today, Pipecat Flows needs to know ahead of time what type of LLM it’s working with, to load an LLM-specific “adapter” that does the work of generating summaries, among other things.

**NOTE: this PR isn't open against `main`. It's dependent on https://github.com/pipecat-ai/pipecat/pull/2440**

**The Pipecat Flows branch I'm testing this against is https://github.com/pipecat-ai/pipecat-flows/pull/174. I'll leave that PR in Draft until all the Pipecat changes have landed that will allow it to properly support runtime LLM switching**

### DIFF
--- a/src/pipecat/services/anthropic/llm.py
+++ b/src/pipecat/services/anthropic/llm.py
@@ -42,6 +42,7 @@ from pipecat.frames.frames import (
     VisionImageRawFrame,
 )
 from pipecat.metrics.metrics import LLMTokenUsage
+from pipecat.processors.aggregators.llm_context import LLMContext
 from pipecat.processors.aggregators.llm_response import (
     LLMAssistantAggregatorParams,
     LLMAssistantContextAggregator,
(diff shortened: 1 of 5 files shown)


src/pipecat/services/aws/llm.py
```diff
@@ -790,6 +791,78 @@ def can_generate_metrics(self) -> bool:
...
+                    "maxTokens": 8192,
+                    "temperature": 0.7,
+                    "topP": 0.9,
+                },
+            }
+
+            request_params["system"] = [{"text": summary_prompt}]
+
+            # Call Bedrock without streaming
+            response = self._client.converse(**request_params)
```
kompfner: Ah, glad I'm doing this refactor! Discovered that Pipecat Flows is currently broken, since it reaches into `AWSBedrockLLMService` to grab this `_client` property, which actually went away with this change: https://github.com/pipecat-ai/pipecat/pull/2148

kompfner: Perfect illustration of why Pipecat Flows shouldn't be reaching into LLM service internals...

kompfner: Fixed!

kompfner:
Note to self: I've validated that summarization is properly working when not using the new `LLMContext`. Still need to validate the `LLMContext` code paths.

kompfner:
> Note to self: I've validated that summarization is properly working when not using the new `LLMContext`. Still need to validate the `LLMContext` code paths.

Validated! Here were the validations I ran, using the Pipecat Flows `patient_intake_*` examples from [this PR branch](https://github.com/pipecat-ai/pipecat-flows/pull/174):

- [x] OpenAI
  - [x] OpenAI-specific context
  - [x] Universal context
- [x] Gemini
  - [x] Gemini-specific context
  - [x] Universal context
- [x] Anthropic
  - [x] Anthropic-specific context
  - [x] Universal context (should print not supported yet message)
- [x] AWS Bedrock
  - [x] AWS Bedrock-specific context
  - [x] Universal context (should print not supported yet message)

src/pipecat/services/anthropic/llm.py
```diff
@@ -198,6 +199,56 @@ async def _create_message_stream(self, api_call, params):
             response = await api_call(**params)
             return response
 
+    async def generate_summary(
```
kompfner, line 202: Make this more generic? Just a one-off, out-of-band LLM call? Thanks @markbackman

kompfner:
Going ahead and merging into the base branch, as it's getting too difficult to manage this as a separate branch. Will continue making the suggested improvements there.

src/pipecat/services/anthropic/llm.py
kompfner: Done! Changes landed in https://github.com/pipecat-ai/pipecat/pull/2440/

kompfner: @markbackman @aconchillo

PR #2480, opened 2025-08-20
replace asyncio.wait_for for wait_for2.wait_for

aconchillo:

Instead of having a separate version, we just replace the main one. This way we don't require to change any code.


markbackman (approved):
LGTM

filipi87 (approved):
Even better. 👏

PR #2487, opened 2025-08-21
remove watchdog timers and specific asyncio implementations

aconchillo:

Watchdog timers have been removed. They were introduced in 0.0.72 to help diagnose pipeline freezes. Unfortunately, they proved ineffective since they required developers to use Pipecat-specific queues, iterators, and events to correctly reset the timer, which limited their usefulness and added friction.


filipi87 (approved):
Pre-approving.

Just waiting for the final tests to decide whether we should keep the queues, but the code is much better without the watchdog. 🙂

PR #2493, opened 2025-08-22
PipelineRunner/PipelineTask: fix asyncio task cancellation

aconchillo:

This PR fixed an issue where PipelineRunner and PipelineTask don't handle asyncio task cancellation. We need to raise the exception from `PipelineTask` and handle it in `PipelineRunner`. This way, we can properly cancel the pipeline task.

This fixes an issue we had in release evals where if the eval takes a long time we cancel the asyncio task. This was causing all sort of problems and not allowing the following evals to execute correctly.

### DIFF
--- a/src/pipecat/pipeline/task.py
+++ b/src/pipecat/pipeline/task.py
@@ -370,11 +370,9 @@ class PipelineTask(BasePipelineTask):
             # We have already cleaned up the pipeline inside the task.
             cleanup_pipeline = False
         except asyncio.CancelledError:
-            # We are awaiting on the push task and it might be cancelled
-            # (e.g. Ctrl-C). This means we will get a CancelledError here as
-            # well, because you get a CancelledError in every place you are
-            # awaiting a task.
-            pass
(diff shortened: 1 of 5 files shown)


markbackman (approved):
LGTM. Also, tested this with the release evals and it works as expected.

src/pipecat/pipeline/task.py
```diff
@@ -370,11 +370,9 @@ async def run(self, params: PipelineTaskParams):
...
             cleanup_pipeline = False
         except asyncio.CancelledError:
-            # We are awaiting on the push task and it might be cancelled
-            # (e.g. Ctrl-C). This means we will get a CancelledError here as
-            # well, because you get a CancelledError in every place you are
-            # awaiting a task.
-            pass
+            # Raise exception back to the pipeline runner so it can cancel this
+            # task properly.
+            raise
```
filipi87, line 375: Was it what was causing the freeze in the unity tests you mentioned yesterday?

aconchillo: Nope. Unrelated

PR #2504, opened 2025-08-25
CartesiaTTSService: reconnect on Cartesia's timeout

aconchillo:

Cartesia doesn't support keep alive messages so it will just disconnect after 5 minutes. We cancel this use case by simply reconnecting. Note that this is not made generic because we don't know how the other `_receive_messages()` implementations might behave. So, this needs to be done per service.


markbackman (approved):
LGTM

PR #2512, opened 2025-08-26
TextFrame: add skip_tts field

aconchillo:

This PR adds `skip_tts` field to `TextFrame`. This lets a text frame bypass TTS while still being included in the LLM context. Useful for cases like structured text that isn’t meant to be spoken but should still contribute to context.

It also adds support for switching between text-only and audio-only modes within the same pipeline. This is done by pushing `LLMConfigureOutputFrame(skip_tts=True)`. The LLM will still generate tokens and add them to the context, but they will not be sent to TTS.

### DIFF
--- a/src/pipecat/frames/frames.py
+++ b/src/pipecat/frames/frames.py
@@ -1351,14 +1351,22 @@ class LLMFullResponseStartFrame(ControlFrame):
     more TextFrames and a final LLMFullResponseEndFrame.
     """
 
-    pass
+    skip_tts: bool = field(init=False)
+
+    def __post_init__(self):
+        super().__post_init__()
+        self.skip_tts = False
 
 
 @dataclass
 class LLMFullResponseEndFrame(ControlFrame):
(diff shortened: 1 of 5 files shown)


src/pipecat/frames/frames.py
```diff
@@ -601,6 +606,20 @@ class LLMEnablePromptCachingFrame(DataFrame):
...
 
+@dataclass
+class LLMSkipTTSFrame(DataFrame):
+    """Frame to indicate whether LLM generated tokens should skip the TTS
+    service if one is in the pipeline. This could be useful to still append
+    tokens to the LLM context without being spoken.
+
+    Parameters:
+        skip_tts: Whether LLM tokens should skip the TTS service (if any).
+
```
markbackman: Remove this blank line to solve a linting issue.

markbackman (approved):
Nice!!

Just small linting clean up for this one.

PR #2513, opened 2025-08-26
Support for the new WhatsApp Cloud API

filipi87:
Support for the new WhatsApp Cloud API:
- Fixing to not use mid to decide if the transceiver should be sendrecv or not.
- Supporting User-initiated Calls: https://developers.facebook.com/docs/whatsapp/cloud-api/calling/user-initiated-calls

### DIFF
--- a/src/pipecat/transports/network/webrtc_connection.py
+++ b/src/pipecat/transports/network/webrtc_connection.py
@@ -387,9 +387,10 @@ class SmallWebRTCConnection(BaseObject):
 
     def force_transceivers_to_send_recv(self):
         """Force all transceivers to bidirectional send/receive mode."""
-        for transceiver in self._pc.getTransceivers():
-            # For now, we only support sendrecv for camera audio and video (the first two transceivers)
-            if transceiver.mid == "0" or transceiver.mid == "1":
+        transceivers = self._pc.getTransceivers()
+        # For now, we only support sendrecv for camera audio and video (the first two transceivers)
+        for i, transceiver in enumerate(transceivers):
+            if i < 2:  # First two transceivers (camera audio and video)
                 transceiver.direction = "sendrecv"
             else:
                 transceiver.direction = "recvonly"
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- Added support for WhatsApp User-initiated Calls.
+
 - Added new audio filter `AICFilter`, speech enhancement for improving VAD/STT
   performance, no ONNX dependency.
   See https://ai-coustics.com/sdk/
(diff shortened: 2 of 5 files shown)


src/pipecat/transports/network/webrtc_connection.py
```diff
@@ -387,9 +387,10 @@ async def delayed_task():
 
     def force_transceivers_to_send_recv(self):
         """Force all transceivers to bidirectional send/receive mode."""
-        for transceiver in self._pc.getTransceivers():
-            # For now, we only support sendrecv for camera audio and video (the first two transceivers)
-            if transceiver.mid == "0" or transceiver.mid == "1":
```
filipi87, line 392: The mid is assigned by the browser and can vary between browsers. It is even possible to modify it in the SDP. 

So, instead, we should use the transceiver index to identify tracks, as done elsewhere in the code.

mattieruth: makes sense. thanks.

mattieruth (approved):

src/pipecat/transports/network/webrtc_connection.py
mattieruth: oops. i just approved, but now just realized that this is all of the WhatsApp Cloud API support, not just this patch 🤦‍♀️

filipi87: Yeah. I would just do this patch, but it would make things harder to test with WhatsApp while it was not merged, so I changed the PR back to draft to include everything.

aconchillo:
I think these files should go to `services/whatsapp` instead of `utils/whatsapp`, similiar to what we do with `services/daily`

src/pipecat/utils/whatsapp/api.py
```diff
@@ -0,0 +1,345 @@
...
+
+        Args:
+            whatsapp_token: WhatsApp access token for authentication
+            phone_number_id: Your business phone number ID from WhatsApp Business API
+            session: aiohttp ClientSession for making HTTP requests
+        """
+        self.phone_number_id = phone_number_id
+        self.session = session
+        self.whatsapp_url = f"{self.BASE_URL}{phone_number_id}/calls"
+        self.whatsapp_token = whatsapp_token
```
aconchillo: if these are internal we should prefix them with underscore.

aconchillo:
> I think these files should go to `services/whatsapp` instead of `utils/whatsapp`, similiar to what we do with `services/daily`

Sorry, I meant `transport/services`.

filipi87:
> > I think these files should go to `services/whatsapp` instead of `utils/whatsapp`, similiar to what we do with `services/daily`
> 
> Sorry, I meant `transport/services`.

I had thought about that, but if you look closely, it’s not actually a new transport. We are still going to use the `SmallWebRTCTransport`. It’s just some utilities to make the WhatsApp integration easier.

Do you still think we should move it to the transport folder? Maybe transport/utils instead of transport/services?

src/pipecat/utils/whatsapp/api.py
filipi87: Done.

aconchillo (approved):

PR #2515, opened 2025-08-26
Add `LLMRunFrame` to trigger an LLM response, replacing `context_aggr…

kompfner:
…egator.user().get_context_frame()`

See the CHANGELOG entry for a description of the change and new recommended usage.

### DIFF
--- a/src/pipecat/frames/frames.py
+++ b/src/pipecat/frames/frames.py
@@ -538,6 +538,17 @@ class LLMMessagesFrame(DataFrame):
 
 
 @dataclass
+class LLMRunFrame(DataFrame):
+    """Frame to trigger LLM processing with current context.
+
+    A frame that instructs the LLM service to process the current context and
+    generate a response.
+    """
+
+    pass
+
+
+@dataclass
 class LLMMessagesAppendFrame(DataFrame):
     """Frame containing LLM messages to append to current context.
 
@@ -557,9 +568,8 @@ class LLMMessagesAppendFrame(DataFrame):
 class LLMMessagesUpdateFrame(DataFrame):
     """Frame containing LLM messages to replace current context.
 
-    A frame containing a list of new LLM messages. These messages will
-    replace the current context LLM messages and should generate a new
-    LLMMessagesFrame.
+    A frame containing a list of new LLM messages to replace the current
+    context LLM messages.
 
     Parameters:
         messages: List of message dictionaries to replace current context.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,47 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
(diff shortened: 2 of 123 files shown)


kompfner:
I'm debating whether to update every example that uses `LLMRunFrame()` immediately after a `messages.append(...)` to use `LLMMessagesAppendFrame(messages=..., run_llm=True)` instead...is it worth it?

kompfner:
> I'm debating whether to update every example that uses `LLMRunFrame()` immediately after a `messages.append(...)` to use `LLMMessagesAppendFrame(messages=..., run_llm=True)` instead...is it worth it?

I'm deciding it's not worth the time, for the purposes of this PR.

src/pipecat/frames/frames.py
```diff
@@ -552,9 +563,8 @@ class LLMMessagesAppendFrame(DataFrame):
 class LLMMessagesUpdateFrame(DataFrame):
     """Frame containing LLM messages to replace current context.
 
-    A frame containing a list of new LLM messages. These messages will
-    replace the current context LLM messages and should generate a new
-    LLMMessagesFrame.
+    A frame containing a list of new LLM messages to replace the current
+    context LLM messages.
```
kompfner, line 572: Updated this comment because it was out of date (`LLMMessagesUpdateFrame` doesn't trigger an `LLMMessagesFrame`)

kompfner:
Just noticed we also need to update examples 22b-d and 25, which all instantiate `OpenAILLMContextFrame` directly...stay tuned

kompfner:
> Just noticed we also need to update examples 22b-d and 25, which all instantiate `OpenAILLMContextFrame` directly...stay tuned

Hmm, actually, looking at those closer, those appear to be "advanced" examples that actually warrant creating context frames with customized context.

markbackman (approved):
Beautiful! 🥹

markbackman:
> I'm debating whether to update every example that uses `LLMRunFrame()` immediately after a `messages.append(...)` to use `LLMMessagesAppendFrame(messages=..., run_llm=True)` instead...is it worth it?

I'm not sure why the messages are broken up like they are. Maybe it's to show that you could dynamically update the messages at connection time. But, as the examples stand, the messages would be better off just created initially. Not something that needs to change here, but just pointing it out.

kompfner:
> > I'm debating whether to update every example that uses `LLMRunFrame()` immediately after a `messages.append(...)` to use `LLMMessagesAppendFrame(messages=..., run_llm=True)` instead...is it worth it?
> 
> I'm not sure why the messages are broken up like they are.

I kinda understand the inclination to break the messages up between initialization-time and connection-time messages. It more clearly distinguishes between system instructions (i.e. the bot's "role") and specific commands (i.e. the bot's "task").

That said, I don't feel strongly that one pattern or another is more "right". So let's just err on the side of doing nothing, for now #laziness

PR #2516, opened 2025-08-26
RTVIProcessor: make check sure client version is set

aconchillo:

Found this while using an example with `RTVIProcessor` and Daily Prebuilt. I called `rtvi.set_client_ready()` manually but then the following error showed up:

```
  File "/home/aleix/src/pipecat-ai/pipecat/src/pipecat/processors/frameworks/rtvi.py", line 1626, in _send_bot_ready         if self._client_version[0] < 1:                                                                                             │    └ []                                                                                                                └ <pipecat.processors.frameworks.rtvi.RTVIProcessor object at 0x7a1959ddbad0>                                                                                                                                                              IndexError: list index out of range                                                                                      
```

### DIFF
--- a/src/pipecat/processors/frameworks/rtvi.py
+++ b/src/pipecat/processors/frameworks/rtvi.py
@@ -1124,7 +1124,9 @@ class RTVIProcessor(FrameProcessor):
         self._bot_ready = False
         self._client_ready = False
         self._client_ready_id = ""
-        self._client_version = []
+        # Default to 0.3.0 which is the last version before actually having a
+        # "client-version".
+        self._client_version = [0, 3, 0]
         self._errors_enabled = True
 
         self._registered_actions: Dict[str, RTVIAction] = {}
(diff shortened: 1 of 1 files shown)


markbackman (approved):

src/pipecat/processors/frameworks/rtvi.py
```diff
@@ -1623,7 +1623,7 @@ async def _handle_action(self, request_id: Optional[str], data: RTVIActionRun):
     async def _send_bot_ready(self):
         """Send the bot-ready message to the client."""
         config = None
-        if self._client_version[0] < 1:
```
mattieruth, line 1626: yes and... maybe we also initialize `_client_version` to the default i'm using in `handle_client_ready`? ([0,3,0])

aconchillo: why is 0.3.0 the default?

mattieruth: i chose the version prior to releasing support for having a client_version so that if we see 0.3.0 then we know it's the old-school version

mattieruth: essentially, a way to say "version is < 1.0.0"

aconchillo: OK, updated.

PR #2517, opened 2025-08-26
unify get_messages_for_logging()

aconchillo:

Some implementations were returning a list and some were returning a JSON string. They should all return a list and the user would decide if it wants to transform that into JSON.

I saw this while fixing an issue in Whisker and trying out Gemini and OpenAI.


markbackman (approved):
LGTM. Nice catch.

PR #2523, opened 2025-08-27
Speechmatics TTS connection issue

sam-s10s:
Connection to the STT service drops prematurely (rogue `finally`).


markbackman (approved):
@sam-s10s that's definitely the issue. It works correctly for me with that change.

I'm approving this change since we'll release Pipecat today and I'd like to include this change.

Can you also add a quick changelog entry?

```
### Fixed

- Fixed an issue preventing `SpeechmaticsSTTService` from transcribing audio.
```

(or something along those lines.)

Issue #2498, opened 2025-08-23
RTVIProcessor Trying to process TransportMessageUrgentFrame but StartFrame not received


remisharrock:
### pipecat version

0.0.80

### Python version

3.12.3

### Operating System

ubuntu 24.04.2 LTS

### Issue description

when I start my server, as soon as a client connects I get this error. 

```
python server.py --host 0.0.0.0                                                                                                                                                         ─╯
2025-08-23 23:29:18.995 | INFO     | pipecat:<module>:14 - ᓚᘏᗢ Pipecat 0.0.80 (Python 3.12.3 (main, Jun 18 2025, 17:59:45) [GCC 13.3.0]) ᓚᘏᗢ
INFO:     Started server process [3816410]
INFO:     Waiting for application startup.
INFO:     Application startup complete.
INFO:     Uvicorn running on http://0.0.0.0:7860 (Press CTRL+C to quit)
INFO:     127.0.0.1:35422 - "OPTIONS /connect HTTP/1.1" 200 OK
INFO:     127.0.0.1:35428 - "POST /connect HTTP/1.1" 200 OK
INFO:     127.0.0.1:35428 - "OPTIONS /api/offer HTTP/1.1" 200 OK
2025-08-23 23:29:37.436 | INFO     | __main__:offer:72 - Client #1 - Nouvelle connexion établie (pc_id: SmallWebRTCConnection#0)
INFO:     127.0.0.1:35422 - "POST /api/offer HTTP/1.1" 200 OK
2025-08-23 23:29:37.533 | ERROR    | pipecat.processors.frame_processor:_check_started:709 - RTVIProcessor#0 Trying to process TransportMessageUrgentFrame#0(message: {'label': 'rtvi-ai', 'type': 'metrics', 'data': {'ttfb': [{'processor': 'AzureSTTService#0', 'value': 0.0}, {'processor': 'DifyLLMService#0', 'value': 0.0}, {'processor': 'AzureTTSService#0', 'value': 0.0}], 'processing': [{'processor': 'AzureSTTService#0', 'value': 0.0}, {'processor': 'DifyLLMService#0', 'value': 0.0}, {'processor': 'AzureTTSService#0', 'value': 0.0}]}}) but StartFrame not received yet
2025-08-23 23:29:37.534 | INFO     | pipecat.transports.network.small_webrtc:connect:407 - Connecting to Small WebRTC
[...]

PR #2526, opened 2025-08-27
PipelineTask: wait for StartFrame to reach end of pipeline

aconchillo:

When using RTVI we were getting this error on startup:

```
RTVIProcessor#0 Trying to process TransportMessageUrgentFrame#0(message: {'label': 'rtvi-ai', 'type': 'metrics', 'data': {'ttfb': [{'processor': 'DeepgramSTTService#0', 'value': 0.0}, {'processor': 'OpenAILLMService#0', 'value': 0.0}, {'processor': 'CartesiaTTSService#0', 'value': 0.0}], 'processing': [{'processor': 'DeepgramSTTService#0', 'value': 0.0}, {'processor': 'OpenAILLMService#0', 'value': 0.0}, {'processor': 'CartesiaTTSService#0', 'value': 0.0}]}}) but StartFrame not received yet   
```

The problem is that observers listen for `on_push_frame` and this might get called before `StartFrame` reaches the RTVI processor.

The reason is that in PipelineTask we were doing:

```
queue_frame(StartFrame())
queue_frame(MetricsFrame())
```

`queue_frame(MetricsFrame())` will generate an `on_push_frame` before `StartFrame` actually gets to the `RTVIProcessor` causing the error mentioned above.

So, we now wait for `StartFrame` to reach the end of the pipeline before doing anything else. This means that you are not able to speed up things anymore. That is, before it was possible to send something along `StartFrame` so processors that have already started are already processing things (even if the processors down below are not started yet). But I believe it makes things more reliable and predictable. Plus, I don't think anyone was using this trick.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -87,6 +87,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- `PipelineTask` now waits for `StartFrame` to reach the end of the pipeline
+  before pushing any other frames.
+
 - Updated `CartesiaTTSService` and `CartesiaHttpTTSService` to align with
   Cartesia's changes for the `speed` parameter. It now takes only an enum of
   `slow`, `normal`, or `fast`.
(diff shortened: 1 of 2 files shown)


aconchillo:
Fixes #2498

markbackman (approved):
LGTM

PR #2528, opened 2025-08-28
PipelineTask: handle BotInterruptionFrame in PipelineTask

aconchillo:

This allows having processors before the input transport or even not having an input transport.

A couple of questions:
1. what if a pipeline doesn’t have an input transport?
2. what if a pipeline has something before the input transport? do we want that to be cancelled as well?

Before:
1. pushing BotInterruptionFrame has no effect
2. it would not be cancelled

Now:
1. pushing BotInterruptionFrame keeps working
2. it would be cancelled

It feels that “Now” is more correct? Also, if you place processors before the input you still have the option to ignore `StartInterruptionFrame`. With the current approach, `StartInterruptionFrame` was being swallowed by the input transport.

### DIFF
--- a/src/pipecat/transports/base_input.py
+++ b/src/pipecat/transports/base_input.py
@@ -22,7 +22,6 @@ from pipecat.audio.turn.base_turn_analyzer import (
 )
 from pipecat.audio.vad.vad_analyzer import VADAnalyzer, VADState
 from pipecat.frames.frames import (
-    BotInterruptionFrame,
     BotStartedSpeakingFrame,
     BotStoppedSpeakingFrame,
     CancelFrame,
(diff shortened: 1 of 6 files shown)


src/pipecat/transports/base_input.py
```diff
@@ -335,13 +332,6 @@ async def process_frame(self, frame: Frame, direction: FrameDirection):
     # Handle interruptions
     #
 
-    async def _handle_bot_interruption(self, frame: BotInterruptionFrame):
-        """Handle bot interruption frames."""
-        logger.debug("Bot interruption")
-        if self.interruptions_allowed:
-            await self._start_interruption()
```
aconchillo, line 342: This was actually unnecessary before, because the base class `FrameProcessor` already calls `await self._start_interruption()` when it gets a `StartInterruptionFrame`.

tests/test_dtmf_aggregator.py
```diff
@@ -69,6 +70,7 @@ async def test_timeout_aggregation(self):
             aggregator,
             frames_to_send=frames_to_send,
             expected_down_frames=expected_down_frames,
+            pipeline_params=PipelineParams(allow_interruptions=False),
```
aconchillo, line 74: This is necessary because `DTMFAggregator` pushes a `BotInterruptionFrame`, so we just need to ignore the `StartInterruptionFrame` now that it's pushed from the task.

aconchillo:
Closing in favor of #2621

PR #2529, opened 2025-08-28
Add IVR navigation

markbackman:

This PR adds a new ivr_navigator module to Pipecat, to allow the bot to navigate an IVR call tree.

Usage:
```
# Initialize IVRNavigator
ivr_navigator = IVRNavigator(
    llm=llm,
    ivr_prompt="You are an IVR navigator calling CVS Pharmacy on behalf of John Smith, date of birth 01/01/1990, phone number 123-456-7890, zip code 12345, Rx number 1234567. Your goal is to check the status of his prescription. Once you've received the prescription status, call the end_call function with the reason 'Call completed' to end the call.",
    conversation_prompt="You are talking to a customer service representative at CVS Pharmacy. Check on the prescription status for John Smith, date of birth 01/01/1990, phone number 123-456-7890, zip code 12345, Rx number 1234567.",
    )

# Use in a pipeline
pipeline = Pipeline(
    [
        transport.input(),
        stt,
        context_aggregator.user(),
        ivr_navigator, # LLM + IVR navigation
        tts,
        transport.output(),
    ]
)

# Handle events
@ivr_navigator.event_handler("on_ivr_stuck")
async def on_ivr_stuck(ivr_navigator):
    logger.warning("IVR stuck, ending call")
    await task.queue_frame(EndFrame())

@ivr_navigator.event_handler("on_ivr_completed")
async def on_ivr_completed(processor):
    # Option 1: Continue with bot conversation
    await processor.start_conversation()
    
    # Option 2: Transfer to human (default - no action needed)
    # await transfer_to_human_agent()
```

The logic for the module is:
IVR-flow

### Considerations

**Goal oriented IVR navigation**
The IVR navigation relies on the LLM to figure out how to navigate the call tree based on a provided goal. In testing, this works well.

**Navigation mechanics**
The IVR will need to navigate with a combination of keypresses (DTMF tones) and voice feedback.

**Flexibly handle conversations or IVR navigation**
In outbound calling there are (at least) two cases to handle:
1. Calling a number where the call flow is not previously known (e.g. conversing with a human or navigating an IVR)
2. Calling a number knowing that the bot will navigate an IVR system

**Navigation end states**
Navigating an IVR tree may result in 3 potential states:
1. IVR goal met, no further conversation. In the on_ivr_completed handler, do things like, push CancelTaskFrame or log to third party system.
2. IVR goal met, transition to conversation. In the on_ivr_completed, call start_conversation to configure bot for further conversation.
3. LLMs will not be perfect in navigating the IVR tree. If they get stuck, emit an event to allow taking action.


CHANGELOG.md
```diff
@@ -9,6 +9,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- Added `pipecat.extensions.ivr` for automated IVR system navigation with
+  configurable goals and conversation handling. Supports DTMF input, verbal
+  responses, and intelligent menu traversal.
+
```
aconchillo, line 15: this should be added to the new release.

src/pipecat/extensions/ivr/ivr_navigator.py
```diff
@@ -0,0 +1,456 @@
...
+)
+from pipecat.pipeline.pipeline import Pipeline
+from pipecat.processors.aggregators.openai_llm_context import OpenAILLMContextFrame
+from pipecat.processors.frame_processor import FrameDirection, FrameProcessor
+from pipecat.services.llm_service import LLMService
+from pipecat.utils.text.pattern_pair_aggregator import PatternMatch, PatternPairAggregator
+
+
+class IVRProcessor(FrameProcessor):
+    """IVR processor for Pipecat."""
```
aconchillo: maybe we should explain more here about how it works in general.

markbackman: Updated

src/pipecat/extensions/ivr/ivr_navigator.py
```diff
@@ -0,0 +1,486 @@
...
+
+        # XML pattern aggregation
+        self._aggregator = PatternPairAggregator()
+        self._setup_xml_patterns()
+
+        # Register IVR events
+        self._register_event_handler("on_ivr_stuck")
+        self._register_event_handler("on_ivr_completed")
+
+    async def start_conversation(self):
```
markbackman: Right now, the way to start a conversation at the end of an IVR tree is to call `ivr_processor.start_conversation()`. This might not be preferred, but the reason for this is that the event emitted includes `ivr_processor` as `self`, so it feels more appropriate to call the method than queue the frame to itself.

Perhaps we need to do this a different way, but that's how it is now. Happy to pair and work through an improvement on this, @aconchillo.

src/pipecat/extensions/ivr/ivr_navigator.py
```diff
@@ -0,0 +1,486 @@
...
+        if isinstance(frame, StartFrame):
+            # Push the StartFrame right away
+            await self.push_frame(frame, direction)
+
+            # Update the context with the appropriate prompt based on the initial mode
+            if self._initial_mode == "conversation":
+                if self._conversation_prompt is None:
+                    raise ValueError(
+                        "conversation_prompt is required when initial_mode is 'conversation'"
+                    )
```
aconchillo: maybe we could do this in the constructor as a precondition, this way when we process StartFrame we know we are good.

src/pipecat/extensions/ivr/ivr_navigator.py
```diff
@@ -0,0 +1,486 @@
...
+
+                # Set the conversation prompt and push it upstream
+                messages = [{"role": "system", "content": self._conversation_prompt}]
+                llm_update_frame = LLMMessagesUpdateFrame(messages=messages)
+                await self.push_frame(llm_update_frame, FrameDirection.UPSTREAM)
+
+                # Set the VAD parameters delay and push it upstream
+                params = VADParams(stop_secs=self._conversation_response_delay)
+                vad_update_frame = VADParamsUpdateFrame(params=params)
+                await self.push_frame(vad_update_frame, FrameDirection.UPSTREAM)
```
aconchillo: One concern about this is that it might override any other user settings. It's probably be better to pass `conversation_vad_params` and `ivr_vad_params` to the constructor. This way you would be able to use the same VADParams you use in the input transport.

src/pipecat/extensions/ivr/ivr_navigator.py
```diff
@@ -0,0 +1,486 @@
...
+        Args:
+            value: The DTMF value to send (0-9, *, #).
+        """
+        logger.debug(f"DTMF detected: {value}")
+
+        try:
+            # Convert the value to a KeypadEntry
+            keypad_entry = KeypadEntry(value)
+            dtmf_frame = OutputDTMFUrgentFrame(button=keypad_entry)
+            await self.push_frame(dtmf_frame, FrameDirection.DOWNSTREAM)
```
aconchillo: No need to specify DOWNSTREAM, we are usually not explicit.

src/pipecat/extensions/ivr/ivr_navigator.py
```diff
@@ -0,0 +1,486 @@
...
+        """
+        logger.debug(f"IVR status detected: {status}")
+
+        if status == "detected":
+            await self._handle_ivr_detected()
+        elif status == "completed":
+            await self._handle_ivr_completed()
+        elif status == "stuck":
+            await self._handle_ivr_stuck()
+        elif status == "wait":
```
aconchillo: Should we have an enum for this? `IVRStatus`?

src/pipecat/extensions/ivr/ivr_navigator.py
```diff
@@ -0,0 +1,486 @@
...
+            else None
+        )
+        self._ivr_response_delay = ivr_response_delay
+        self._conversation_response_delay = conversation_response_delay
+        self._initial_mode = initial_mode
+
+        # Track conversation turn count to optimize pure conversation scenarios
+        self._conversation_turn_count = 0
+
+        print(self._ivr_prompt)
```
aconchillo: remove

src/pipecat/extensions/ivr/ivr_navigator.py
```diff
@@ -0,0 +1,486 @@
...
+Remember: Respond with `<dtmf>NUMBER</dtmf>` (single or multiple for sequences), `<ivr>completed</ivr>`, `<ivr>stuck</ivr>`, `<ivr>wait</ivr>`, OR natural language text when verbal responses are requested. No other response types."""
+
+    def __init__(
+        self,
+        *,
+        llm: LLMService,
+        ivr_prompt: str,
+        conversation_prompt: Optional[str] = None,
+        ivr_response_delay: Optional[float] = 2.0,
+        conversation_response_delay: Optional[float] = 0.8,
```
aconchillo: we would pass `VADParams` here. This would be None by default, and then you would do:

```
        self._ivr_vad_params = ivr_vad_params or VADParams(....)
        self._conversation_vad_params = conversation_vad_params or VADParams(....)
```

Maybe be the `conversation_vad_params` should be mandatory so you are forced to pass the same one as the input transport.

src/pipecat/extensions/ivr/ivr_navigator.py
```diff
@@ -0,0 +1,486 @@
...
+
+                # Set the VAD parameters delay and push it upstream
+                params = VADParams(stop_secs=self._conversation_response_delay)
+                vad_update_frame = VADParamsUpdateFrame(params=params)
+                await self.push_frame(vad_update_frame, FrameDirection.UPSTREAM)
+            else:
+                # Set the IVR prompt and push it upstream
+                messages = [{"role": "system", "content": self._ivr_prompt}]
+                llm_update_frame = LLMMessagesUpdateFrame(messages=messages)
+                await self.push_frame(llm_update_frame, FrameDirection.UPSTREAM)
```
aconchillo: Just add a quick comment that we push UPSTREAM because the IVR processor is placed after the LLM by the IVRNavigator.

CHANGELOG.md
```diff
@@ -5,6 +5,14 @@ All notable changes to **Pipecat** will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Added
+
+- Added `pipecat.extensions.ivr` for automated IVR system navigation with
```
kompfner, line 12: Should we add the usage example here that you have in your PR notes? I personally like to have an example committed alongside the code change.

kompfner:
These questions are only based on looking at the usage example in the PR overview:

* Does the user need to also specify "call the end_call function with the reason 'Call completed' to end the call.") in the `conversation_prompt`?
* Is the `end_call` function something that the IVR navigator brings? If so:
  * Is there a way to help the user not have to write the same "call the end_call function with..." boilerplate in all their prompts?
  * How are we recommending that the user get the desired information (e.g. prescription status) out of the navigator and stored somewhere? By including in the prompt an instruction to call a function (e.g. "call save_prescription_status")? Or does `end_call` take an argument with the "payload" of info that the bot is trying to get out of the call?
  * Can we do something to help with the presumably large amount of overlap between the IVR and conversation prompts (since the overall goal of the bot is likely to be the same regardless of whether they're talking to an IVR system or human)
* Does "on_ivr_completed" only indicate that the bot is now talking to a person? Or can it also mean that the call is done? (i.e. that `end_call` has been called?)
  * If the former, then a different name might clarify. If the latter, how can we check what happened, in the body of the handler?
  * If it's also called after an `end_call`, does that mean it's also called when the bot's conversation with a human is done, too? Does that mean it can potentially be called twice?

src/pipecat/extensions/ivr/ivr_navigator.py
markbackman: Good idea.

src/pipecat/extensions/ivr/ivr_navigator.py
chadbailey59: @markbackman as a user of this, I think my expected behavior would be that I could provide an ivr_prompt and a conversation_prompt, and when IVR navigation is finished, the bot would Just(tm) transition to conversation mode. Maybe we add a 'convenience' that if you haven't specified an on_ivr_completed handler, we call start_conversation?

src/pipecat/extensions/ivr/ivr_navigator.py
markbackman: I think we want `conversation_vad_params` to be Optional` since it's optional to transition to a conversation.

src/pipecat/extensions/ivr/__init__.py
```diff
@@ -0,0 +1,18 @@
...
+This module provides automated navigation of IVR phone systems using LLM-based
+decision making and DTMF tone generation.
+
+Public Interface:
+    IVRNavigator: Main pipeline for automated IVR system navigation
+"""
+
+from .ivr_navigator import IVRNavigator
+
+__all__ = ["IVRNavigator"]
```
aconchillo: Maybe all this is not required. By default we will make everything public. For example, we now need `IVRStatus` to be public.

src/pipecat/extensions/ivr/ivr_navigator.py
```diff
@@ -0,0 +1,443 @@
...
+    - DTMF command processing (`<dtmf>1</dtmf>`)
+    - IVR state management (see IVRStatus enum: `<ivr>detected</ivr>`, `<ivr>completed</ivr>`, `<ivr>stuck</ivr>`, `<ivr>wait</ivr>`)
+    - Automatic prompt and VAD parameter switching
+    - Event emission via on_ivr_status_changed for detected, completed, and stuck states
+    """
+
+    def __init__(
+        self,
+        *,
+        mode_detection_prompt: str,
```
aconchillo: Maybe `classifier_prompt`?

src/pipecat/extensions/ivr/ivr_navigator.py
```diff
@@ -0,0 +1,443 @@
...
+            return
+
+        if ivr_status == IVRStatus.DETECTED:
+            await self._handle_ivr_detected()
+        elif ivr_status == IVRStatus.COMPLETED:
+            await self._handle_ivr_completed()
+        elif ivr_status == IVRStatus.STUCK:
+            await self._handle_ivr_stuck()
+        elif ivr_status == IVRStatus.WAIT:
+            await self._handle_ivr_wait()
```
aconchillo: using `match` would probably look nicer

src/pipecat/extensions/ivr/__init__.py
markbackman: You're right about `IVRStatus`. I want to make it clear that the IVRProcessor is not meant to be used directly. Maybe it's not worth specifying that though. WDYT?

src/pipecat/extensions/ivr/ivr_navigator.py
```diff
@@ -0,0 +1,443 @@
...
+        Args:
+            frame: The frame to process.
+            direction: The direction of frame flow in the pipeline.
+        """
+        if isinstance(frame, (OpenAILLMContextFrame, LLMContextFrame)):
+            # Extract messages and pass to IVR processor
+            if isinstance(frame, OpenAILLMContextFrame):
+                all_messages = frame.context.messages
+            else:
+                all_messages = frame.context.get_messages()
```
aconchillo: `OpenAILLMContextFrame` has a `get_messages()` so you just need:

```
all_messages = frame.context.get_messages()
```

src/pipecat/extensions/ivr/ivr_navigator.py
```diff
@@ -0,0 +1,443 @@
...
+
+            # Store messages in the IVR processor for mode switching
+            self._ivr_processor._saved_messages = all_messages
```
aconchillo: always better to expose a public function in `IVRProcessor` instead of assigning a private variable directly

aconchillo: it allows you to do other stuff if you need to.

src/pipecat/extensions/ivr/__init__.py
markbackman: I'll just make it all public. No harm in that.

src/pipecat/extensions/ivr/ivr_navigator.py
```diff
@@ -64,22 +64,22 @@ class IVRProcessor(FrameProcessor):
...
-        mode_detection_prompt: str,
+        classifier_prompt: str,
         ivr_prompt: str,
         ivr_vad_params: Optional[VADParams] = None,
     ):
         """Initialize the IVR processor.
 
         Args:
-            mode_detection_prompt: System prompt for mode detection.
+            classifier_prompt: System prompt for classifying.
```
aconchillo: `for classifying IVR or conversation`

aconchillo (approved):

aconchillo:
LGTM!

PR #2532, opened 2025-08-28
Universal DTMF support

aconchillo:

This PR adds support for universal DTMF for all output transports.

- Script to generate WAV file for each DTMF tone (0,1,2,3,4,5,6,7,8,9,*,#)
- WAV files added to the `pipecat.audio.dtmf` module.
- Added helper function to load DTMF file `load_dtmf_audio()`. This caches the DTMF audio so you only load it once.
- Implement `BaseOuputTransport.write_dtmf()` by sending DTMF audio directly. Sub-classes don't need to implement `write_dtmf()` anymore.


CHANGELOG.md
```diff
@@ -5,6 +5,31 @@ All notable changes to **Pipecat** will be documented in this file.
...
+
+- `pipecat.frames.frames.KeypadEntry` is deprecated and has been moved to
+  `pipecat.audio.dtmf.types.KeypadEntry`.
+
+## Removed
+
+- `DailyTransport.write_dtmf()` has been removed in favor of the generic
+  `BaseOutputTransport.write_dtmf()`.
+
+- Remove deprecated `DailyTransport.send_dtmf()`.
```
markbackman, line 26: I think this is OK. No one should be calling send_dtmf() directly.

src/pipecat/audio/dtmf/types.py
```diff
@@ -0,0 +1,45 @@
+# Copyright (c) 2024–2025, Daily
+#
+# SPDX-License-Identifier: BSD 2-Clause License
+#
+
+"""Core frame definitions for the Pipecat AI framework.
```
markbackman: Update this docstring to reference just dtmf audio frame types.

aconchillo: Done!

CHANGELOG.md
aconchillo: Nope.

src/pipecat/frames/frames.py
```diff
@@ -41,9 +41,13 @@
     from pipecat.processors.frame_processor import FrameProcessor
 
 
-class KeypadEntry(str, Enum):
+class DeprecatedKeypadEntry:
```
markbackman, line 44: This change is breaking. In running my ivr code with this, I see:
```
2025-08-28 19:28:53.234 | ERROR    | pipecat.utils.text.pattern_pair_aggregator:_process_complete_patterns:176 - Error in pattern handler for dtmf: 'DeprecatedKeypadEntry' object is not callable
```

markbackman (commented):
This works really well! One additional benefit is that the keypresses are shorter than the ones that Daily uses via SignalWire. This speeds up the call time by a noticeable amount (which is great!)

Just a few comments. We should avoid breaking with this change.

src/pipecat/frames/frames.py
aconchillo: weird... do you have the latest?

markbackman: Yes, I pulled the latest. I just checked again.

I also double checked that the same code works on main.

aconchillo:
> This works really well! One additional benefit is that the keypresses are shorter than the ones that Daily uses via SignalWire. This speeds up the call time by a noticeable amount (which is great!)
> 
> Just a few comments. We should avoid breaking with this change.

Looking at it.

aconchillo:
> > This works really well! One additional benefit is that the keypresses are shorter than the ones that Daily uses via SignalWire. This speeds up the call time by a noticeable amount (which is great!)
> > Just a few comments. We should avoid breaking with this change.
> 
> Looking at it.

This should be fixed now.

markbackman (approved):
Works great now. I tested the old and new ways. Ship it!

PR #2533, opened 2025-08-29
Added STT Translate Websocket Implementation

pratham-sarvam:
Implemented a new Sarvam speech-to-text service which is a speech to text translate streaming websocket.


pyproject.toml
```diff
@@ -91,7 +91,7 @@ rime = [ "websockets>=13.1,<15.0" ]
 riva = [ "nvidia-riva-client~=2.21.1" ]
 runner = [ "python-dotenv>=1.0.0,<2.0.0", "uvicorn>=0.32.0,<1.0.0", "fastapi>=0.115.6,<0.117.0", "pipecat-ai-small-webrtc-prebuilt>=1.0.0"]
 sambanova = []
-sarvam = [ "websockets>=13.1,<15.0" ]
+sarvam = [ "sarvamai", "websockets>=13.1,<15.0" ]
```
markbackman, line 94: Can we be more specific about the `sarvamai` package version? Maybe:
```
sarvam = [ "sarvamai>=0.1.19,<1", "websockets>=13.1,<15.0" ]
```
(or whatever upper bound makes sense)

src/pipecat/services/sarvam/stt.py
```diff
@@ -0,0 +1,392 @@
...
+    VADUserStartedSpeakingFrame,
+    VADUserStoppedSpeakingFrame,
+)
+from pipecat.services.stt_service import STTService
+from pipecat.transcriptions.language import Language
+from pipecat.utils.time import time_now_iso8601
+from pipecat.utils.tracing.service_decorators import traced_stt
+
+try:
+    import httpx
```
markbackman, line 36: Let's remove `httpx`, which would be a new dependency. I see this just just used for query param parsing. Instead, let's use:
```
from urllib.parse import urlencode
```

This import would be above, outside of the try/except block.

src/pipecat/services/sarvam/stt.py
```diff
@@ -0,0 +1,392 @@
...
+
+        try:
+            # Build WebSocket URL and headers manually
+            ws_url = (
+                self._client._client_wrapper.get_environment().production
+                + "/speech-to-text-translate/ws"
+            )
+
+            # Add query parameters
+            query_params = httpx.QueryParams()
```
markbackman, line 225: Since we're removing httpx, this should be:
```
query_params = {"model": self._model, "vad_signals": "true"}
query_string = urlencode(query_params)
ws_url = ws_url + f"?{query_string}"
```

src/pipecat/services/sarvam/stt.py
```diff
@@ -0,0 +1,392 @@
...
+try:
+    import httpx
+    import websockets
+    from sarvamai import AsyncSarvamAI
+    from sarvamai.speech_to_text_translate_streaming.socket_client import (
+        AsyncSpeechToTextTranslateStreamingSocketClient,
+    )
+except ModuleNotFoundError as e:
+    logger.error(f"Exception: {e}")
+    logger.error("In order to use Sarvam, you need to `pip install sarvamai websockets httpx`.")
```
markbackman, line 44: Update to:
```suggestion
    logger.error("In order to use Sarvam, you need to `pip install pipecat-ai[sarvam]`.")
```

markbackman: This will install the optional dependencies from the `sarvam` pyproject.toml entry.

src/pipecat/services/sarvam/stt.py
```diff
@@ -0,0 +1,392 @@
...
+
+
+class EventResponse(BaseModel):
+    """Response structure for VAD events."""
+
+    type: Literal["events"]
+    data: EventData
+
+
+class SarvamSpeechToTextTranslateService(STTService):
```
markbackman, line 93: Can we shorten the name to `SarvamSTTService`? This would be more consistent with other, similar services.

```suggestion
class SarvamSTTService(STTService):
```

src/pipecat/services/sarvam/stt.py
```diff
@@ -0,0 +1,392 @@
...
+
+    Provides real-time speech recognition using Sarvam's WebSocket API.
+    """
+
+    def __init__(
+        self,
+        *,
+        api_key: str,
+        model: str = "saaras:v2.5",
+        language_code: str = "hi-IN",
```
markbackman, line 104: Pipecat provides Language enums:
```
```

Instead, you should use:
```suggestion
        language_code: str = Langage.HI_IN,
```

Then, include a function that takes in a Language and returns a string. The supported languages/dialects can be included. You can find examples in many of the service files. Gladia for example: https://github.com/pipecat-ai/pipecat/blob/main/src/pipecat/services/gladia/stt.py#L47.

src/pipecat/services/sarvam/stt.py
```diff
@@ -0,0 +1,392 @@
...
+        """Initialize the Sarvam STT service.
+
+        Args:
+            api_key: Sarvam API key for authentication.
+            model: Sarvam model to use for transcription.
+            language_code: Language code for transcription (e.g., "hi-IN", "kn-IN").
+            sample_rate: Audio sample rate. Defaults to 16000 if not specified.
+            **kwargs: Additional arguments passed to the parent STTService.
+        """
+        super().__init__(sample_rate=sample_rate or 16000, **kwargs)
```
markbackman, line 117:
```suggestion
        super().__init__(sample_rate=sample_rate, **kwargs)
```

src/pipecat/services/sarvam/stt.py
```diff
@@ -0,0 +1,392 @@
...
+
+        try:
+            # Convert audio bytes to base64 for Sarvam API
+            audio_base64 = base64.b64encode(audio).decode("utf-8")
+
+            message = {
+                "audio": {
+                    "data": audio_base64,
+                    "encoding": "audio/wav",
+                    "sample_rate": self.sample_rate,
```
markbackman, line 199: This is where the base classes sample rate value is provided. No need to pass to the constructor. This happens automatically by using the `self.sample_rate` property.

src/pipecat/services/sarvam/stt.py
```diff
@@ -0,0 +1,392 @@
...
+        self.set_model_name(model)
+        self._api_key = api_key
+        self._model = model
+        self._language_code = language_code
+
+        self._client = AsyncSarvamAI(api_subscription_key=api_key)
+        self._websocket = None
+        self._websocket_connection = None
+        self._listening_task = None
+        self._is_connected = False
```
markbackman, line 128: Why do you need to track this state? Can't you just rely on the websocket's connection state? This would be much preferred.

src/pipecat/services/sarvam/stt.py
```diff
@@ -0,0 +1,392 @@
...
+            **kwargs: Additional arguments passed to the parent STTService.
+        """
+        super().__init__(sample_rate=sample_rate or 16000, **kwargs)
+
+        self.set_model_name(model)
+        self._api_key = api_key
+        self._model = model
+        self._language_code = language_code
+
+        self._client = AsyncSarvamAI(api_subscription_key=api_key)
```
markbackman, line 124: Why do you need the client? It seems like you're rely on websockets mainly. Can this class just use one or the other--websocket or AsyncSarvamAI?

src/pipecat/services/sarvam/stt.py
```diff
@@ -0,0 +1,392 @@
...
+        logger.debug(f"Received response: {response}")
+
+        try:
+            if response["type"] == "events":
+                parsed = EventResponse(**response)
+                signal = parsed.data.signal_type
+                timestamp = parsed.data.occured_at
+                logger.debug(f"VAD Signal: {signal}, Occurred at: {timestamp}")
+
+                if signal == VADSignal.START:
```
markbackman, line 325: The STT service shouldn't push these frames. If you want to make VAD events available, I'd recommend emitting events similar how it's done in Deepgram:
https://github.com/pipecat-ai/pipecat/blob/main/src/pipecat/services/deepgram/stt.py#L267-L272

Then, the application code can handle this. Again, using the Deepgram events:
https://github.com/pipecat-ai/pipecat/blob/main/examples/foundational/07c-interruptible-deepgram-vad.py#L98-L104

This issue is, you can't just push these frames from the STT. They need to be pushed throughout the entire pipeline. This is something you can do in the application code from the PipelineTask. This ensures that all processes receive the frames.

src/pipecat/services/sarvam/stt.py
```diff
@@ -0,0 +1,392 @@
...
+                    )
+
+                # await self._handle_transcription(transcript, is_final, language)
+                await self.stop_processing_metrics()
+
+        except Exception as e:
+            logger.error(f"Error handling Sarvam response: {e}")
+            await self.push_error(ErrorFrame(f"Failed to handle response: {e}"))
+
+    def _map_language_code_to_enum(self, language_code: str) -> Language:
```
markbackman, line 362: Oh! You have a language mapping function. Let's use it, as I mentioned above.

markbackman:
Hi guys, I'd love to close this out soon. Any update on this one?

shreyas-sarvam:
> Hi guys, I'd love to close this out soon. Any update on this one?

Hey @markbackman, unfortunately @pratham-sarvam will not be able to work on this PR.
I have implemented the changes as mentioned by you in this PR: #2821.
Looking forward your comments to closing it out quickly.

markbackman:
Sounds good. Closing out this PR.

I'll look at the other one hopefully by next week.

PR #2534, opened 2025-08-29
pyproject: update pyright and ruff

aconchillo:


pyproject.toml
```diff
@@ -115,11 +115,11 @@ dev = [
     "grpcio-tools~=1.67.1",
     "pip-tools~=7.4.1",
     "pre-commit~=4.2.0",
-    "pyright~=1.1.402",
+    "pyright~=1.1.404",
```
markbackman: In looking at release history, they've made patch releases for the last 3 years, so `~=` would pick up those changes and use the latest already. We could be more explicit and do:
```
"pyright>=1.1.404,<1.2",
```

which is the same thing.

pyproject.toml
```diff
@@ -115,11 +115,11 @@ dev = [
...
     "pytest~=8.4.1",
     "pytest-asyncio~=1.1.0",
     "pytest-aiohttp==1.1.0",
-    "ruff~=0.12.1",
+    "ruff~=0.12.11",
```
markbackman: Could we do:
```
"ruff>=0.12.11,<1",
```

so we don't have to worry about updating this one?

markbackman (commented):
Just a few comments.

pyproject.toml
aconchillo: Done.

pyproject.toml
aconchillo: Done.

PR #2536, opened 2025-08-29
Gladia - add missing config parameters

fabrice404:
Adding some missing parameters, and pass pipecat version to Gladia so we can trace pipecat transcriptions more easily.


markbackman (approved):
LGTM.

Can you please add a changelog entry?

Issue #2530, opened 2025-08-28
Frozen bug when closing again!!!!


anotine10:
### pipecat version

0.0.81

### Python version

3.13

### Operating System

ubuntu

### Issue description

With 0.0.81, the error that makes the closing pipeline fails again there! I think it was solved in the 0.0.78. 

When the following log appears, i can not close correctly the pipeline
```

2025-08-28 17:13:57.931 | ERROR    | pipecat.transports.network.fastapi_websocket:send:130 - <pipecat.transports.network.fastapi_websocket.FastAPIWebsocketClient object at 0x315a360d0> exception sending data: WebSocketDisconnect (), application_state: WebSocketState.DISCONNECTED
```
Also something weird happen, before that, it seems that the bot could not start talking again: I was talking, and the bot did not answer.

### Reproduction steps

Not sure how to reproduce it

### Expected behavior

Pipeline should stop

### Actual behavior

pipeline is stuck somewhere

### Logs

```shell

```

anotine10:
hi @filipi87 , in the 0.0.81, do you reproduce again the error that was supposed to be fixed in https://github.com/pipecat-ai/pipecat/issues/2209 ?

filipi87:
Hi @anotine10, thank you for reporting this. Let me try to reproduce it again with 0.0.81 and see if there was an unexpected regression.

filipi87:
Hi @anotine10,

It was not the same error we had fixed before, but the result was somewhat similar: the Pipeline was not ending properly.

In this [PR](https://github.com/pipecat-ai/pipecat/pull/2552), I have pushed two fixes that I believe will resolve the problem:
- Fixed an issue where Deepgram swallowed asyncio.CancelledError during disconnect, preventing tasks from being cancelled.
- Added a timeout around canceling frame processor tasks to prevent indefinite hangs when cancellation is swallowed by third-party code.

It would be great if you could give it a try.

anotine10:
hi @filipi87 , thansk for taking it.! 

I have added one comment in the PR.
I have tested with the PR, and i get the following message:

```
[...]

PR #2552, opened 2025-09-01
Fixed an issue where the pipeline could freeze.

filipi87:
Fixed an issue where the pipeline could freeze.
- Fixed an issue where Deepgram swallowed `asyncio.CancelledError` during 
  disconnect, preventing tasks from being cancelled.
- Added a timeout around cancel frame processor tasks to prevent indefinite 
  hangs when cancellation is swallowed by third-party code.


filipi87:
Fixes this [issue](https://github.com/pipecat-ai/pipecat/issues/2530).

src/pipecat/processors/frame_processor.py
```diff
@@ -742,7 +742,10 @@ def __create_input_task(self):
     async def __cancel_input_task(self):
         """Cancel the frame input processing task."""
         if self.__input_frame_task:
-            await self.cancel_task(self.__input_frame_task)
+            # Apply a timeout as a safeguard: if a library swallows asyncio.CancelledError,
+            # the task would otherwise never be cancelled. With a timeout, we can detect this
+            # situation and surface it in the logs instead of hanging indefinitely.
+            await self.cancel_task(self.__input_frame_task, 3)
```
anotine10: What is the possible impact of having some task not properly cancelled? Will they hangup forever and be a possible problem in the scenario of handling thousands of calls ? will this prevent to properly clean resource and make memory grow up infinitly ?

aconchillo: The input task is only cancelled at the very end when application is closing. This timeout prevents hanging during closing the application.

@filipi87 let's create a constant for this, `INPUT_TASK_CANCEL_TIMEOUT_SECS` or something like that.

src/pipecat/services/deepgram/stt.py
```diff
@@ -237,7 +238,15 @@ async def _connect(self):
     async def _disconnect(self):
         if self._connection.is_connected:
             logger.debug("Disconnecting from Deepgram")
-            await self._connection.finish()
+            result = await self._connection.finish()
+            # Deepgram swallows asyncio.CancelledError internally and returns False instead.
+            # This prevents proper cancellation propagation: tasks awaiting on queue.get()
+            # would remain stuck if we didn’t normalize this back into a CancelledError.
```
aconchillo, line 244: @filipi87 once you file a GH isssue to Deepgram, add the link to this comment.

filipi87: Done: https://github.com/deepgram/deepgram-python-sdk/issues/570

src/pipecat/processors/frame_processor.py
filipi87: Done!

aconchillo:
LGTM!

anotine10:
hi @filipi87 , are you sure that raising the asyncio.CancelledError is the good solution? 
the function _disconnect() is called in different scenarios where we may not expect an asyncio.CancelledError when something fails.
_disconnect is called in set_model, set_language and normal stop()

filipi87:
> hi @filipi87 , are you sure that raising the asyncio.CancelledError is the good solution?

Hi @anotine10, we have decided for a different solution in a follow up PR: https://github.com/pipecat-ai/pipecat/pull/2572/files

We are not raising the CancelledError anymore here.

anotine10:
But is it still valid for normal pipeline? 
And also what about this Issue (https://github.com/pipecat-ai/pipecat/issues/2530) which was supposed to be fixed by this PR.

filipi87:
It is still fixed. We just found out that we were only able to replicate it when using ParallelPipeline. That was the only case where Deepgram was swallowing the asyncio.CancelledError, because the cleanUp method was invoked while Deepgram was still finishing.

We also kept the timeout around cancel input tasks to prevent indefinite hangs when cancellation is swallowed by third-party code, in case it happens again with any other third party in the future.

anotine10:
But if you removed the part of raising the CancelledError, what is the code that is suppoed to fix the error? is it just the timeout around the cancel input tasks?

PR #2560, opened 2025-09-02
Added Daily SIP room creation utility

markbackman:

Also, updated configure() to take option kwargs for room and token durations.


src/pipecat/runner/daily.py
```diff
@@ -19,28 +23,54 @@
...
 from pipecat.transports.services.helpers.daily_rest import (
     DailyRESTHelper,
     DailyRoomParams,
     DailyRoomProperties,
+    DailyRoomSipParams,
 )
 
 
-async def configure(aiohttp_session: aiohttp.ClientSession) -> Tuple[str, str]:
+class SipRoomConfig(BaseModel):
```
filipi87: I wonder if we shouldn't have a `DailyRoomConfig`, which would include `room_url` and `token`, and a `DailySipRoomConfig`, which would extend `DailyRoomConfig` and add the extra `sip_endpoint`.

filipi87: We could use just `DailyRoomConfig` and make `sip_endpoint` optional.

A single `configure` method could then handle all parameters and create a SIP room only if the SIP option is specified.

markbackman: Good idea. I need to preserve unpacking the tuple values (room_url, token) for backwards compatibility. But, this is a better approach going forward.

I've consolidated into a single function and now just return `DailyRoomConfig`.

src/pipecat/runner/daily.py
```diff
@@ -64,45 +123,83 @@ async def configure(aiohttp_session: aiohttp.ClientSession) -> Tuple[str, str]:
...
+        exp=expiration_time,
+        eject_at_room_exp=True,
+    )
+
+    # Add SIP configuration if enabled
+    if sip_enabled:
+        sip_params = DailyRoomSipParams(
+            display_name=sip_caller_phone,
+            video=sip_enable_video,
+            sip_mode=sip_mode,
```
vr000m: I think we only support dial-in

markbackman: Good to know! I was looking into this earlier today and it was unclear. I'll make that the case here. Thank you.

src/pipecat/runner/daily.py
```diff
@@ -64,45 +121,83 @@ async def configure(aiohttp_session: aiohttp.ClientSession) -> Tuple[str, str]:
...
+        eject_at_room_exp=True,
+    )
+
+    # Add SIP configuration if enabled
+    if sip_enabled:
+        sip_params = DailyRoomSipParams(
+            display_name=sip_caller_phone,
+            video=sip_enable_video,
+            sip_mode="dial-in",
+            num_endpoints=sip_num_endpoints,
```
vr000m, line 168: Maybe we need codecs?

The default can be null

```
"codecs" : {
        "audio": ["OPUS"],
        "video": ["H264"]
      }
```

vr000m (approved):

src/pipecat/runner/daily.py
markbackman: Good idea. Adding this now.

PR #2563, opened 2025-09-02
Expand universal `LLMContext` support to more LLMs

kompfner:
Went through and painstakingly tested each one of these, just in case the new `LLMContext` introduced some new incompatibility.


CHANGELOG.md
```diff
@@ -9,6 +9,28 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+  Here are the newly-supported services:
+
+  - Azure
+  - Cerebras
+  - Deepseek
+  - Fireworks AI
+  - Google Vertex AI
+  - Grok
+  - Groq
+  - Mistral
```
kompfner, line 24: Technically Mistral was already (unintentionally) supported before this change, but no matter—it’s worth mentioning now, as Mistral wasn’t on the list of supported services in the last CHANGELOG entry

examples/foundational/14i-function-calling-fireworks.py
```diff
@@ -75,9 +75,13 @@ async def run_bot(transport: BaseTransport, runner_args: RunnerArguments):
     # sent to the same callback with an additional function_name parameter.
     llm.register_function("get_current_weather", fetch_weather_from_api)
 
-    @llm.event_handler("on_function_calls_started")
-    async def on_function_calls_started(service, function_calls):
-        await tts.queue_frame(TTSSpeakFrame("Let me check on that."))
+    # Disabling for now, as it ends up tripping up the model in this example
```
kompfner, line 78: Didn't want to dig too deeply in this PR, but just calling this out.

examples/foundational/14k-function-calling-cerebras.py
```diff
@@ -72,9 +72,13 @@ async def run_bot(transport: BaseTransport, runner_args: RunnerArguments):
     # sent to the same callback with an additional function_name parameter.
     llm.register_function("get_current_weather", fetch_weather_from_api)
 
-    @llm.event_handler("on_function_calls_started")
-    async def on_function_calls_started(service, function_calls):
-        await tts.queue_frame(TTSSpeakFrame("Let me check on that."))
+    # Disabling for now, as we end up in an infinite inference loop with the
```
kompfner, line 75: Worth noting

kompfner: As discussed with @aconchillo, perhaps we should explore (in a separate PR) how to make it so that "let me check on that" ends up _before_ the tool call in the context's conversation history.

The problem is that any frames you enqueue within an "on_function_calls_started" will end up *after* the tool call because `FunctionCallInProgressFrame`—which triggers adding the tool call to context—is a `SystemFrame`, which gets priority in the pipeline.

kompfner:
While working on this, I discovered that the Together.ai service is in somewhat of a broken state. Try out the 14c example.

`LLMContext` works as well as `OpenAILLMContext` in Together.ai, which is to say, not that well. I can plan to look into it in a bit.

examples/foundational/14i-function-calling-fireworks.py
markbackman: This is fine. I hit another case where the `on_function_cals_started` event causes issues. IIRC, it was with Mistral.

examples/foundational/14k-function-calling-cerebras.py
markbackman: This is a known issue with Cerebras function calling. I just reached out to a PM at Cerebras asking if function calling is supported when using the OpenAI client.

markbackman (approved):
LGTM! Release evals will be a good test for this to make sure all is well.

kompfner:
> LGTM! Release evals will be a good test for this to make sure all is well.

Evals won't cover `LLMContext` usage with all of these LLMs yet. We'd need to either:
- Migrate all the examples to use `LLMContext`, or
- Duplicate all the examples so there's an `LLMContext` "flavor" of each of them

markbackman:
> > LGTM! Release evals will be a good test for this to make sure all is well.
> 
> Evals won't cover `LLMContext` usage with all of these LLMs yet. We'd need to either:
> 
> * Migrate all the examples to use `LLMContext`, or
> * Duplicate all the examples so there's an `LLMContext` "flavor" of each of them

Are we ready to make that switch to the foundational examples yet? (Duplicating the examples is something we should avoid.)

examples/foundational/14k-function-calling-cerebras.py
kompfner: Cerebras (at least as configured in this example) does seem to get this wrong where most get it right.

But IMO it really is an understandable mistake, given the wonky order messages end up in the context convo history.

I still think it might be worth digging into the underlying Pipecat behavior, described above, that results in the order being scrambled, as it may be affecting multiple LLMs.

kompfner: > function calling is supported when using the OpenAI client

It definitely is. It works fine when we don't stick a message after the function call before running inference again.

markbackman: What I mean is that function calls, themselves, don't work using CerebrasLLMService. They never have. At one point, the Cerebras team flagged it as a "known issue" on their end. I'm waiting to see if this is still the case.

The fact that `on_function_calls_started` fires over and over again is because the function call itself fires over and over again.

kompfner:
> > > LGTM! Release evals will be a good test for this to make sure all is well.
> > 
> > 
> > Evals won't cover `LLMContext` usage with all of these LLMs yet. We'd need to either:
> > 
> > * Migrate all the examples to use `LLMContext`, or
> > * Duplicate all the examples so there's an `LLMContext` "flavor" of each of them
> 
> Are we ready to make that switch to the foundational examples yet? (Duplicating the examples is something we should avoid.)

Not quite yet, I think.

I've duplicated OpenAI + Google since those involve quite a bit of bespoke code to support. I'm planning to do the same for Anthropic. These OpenAI-derivative services are low-risk, so I won't duplicate all of their examples.

In a bit, we'll update everything to universal context and then remove the duplicated examples in one fell swoop.

PR #2565, opened 2025-09-02
transports: reorganize module

aconchillo:

This PR reorganizes transports. It creates one folder per transport (like we do for services) instead of `network` and `services`. I believe it makes things much cleaner and easier to extend.


markbackman (approved):
LGTM

Issue #2558, opened 2025-09-02
AWSBedrockLLMService crashes when passing retry_timeout_secs, retry_on_timeout params


ezisezis:
### pipecat version

0.0.82

### Python version

3.9.6

### Operating System

MacOS 15.6.1

### Issue description

I wanted to introduce some redundancy and retry on timeout, we are using `AWSBedrockLLMService` but adding the params  breaks the pipeline and bot does not work. 

Potential fix from Cursor:
<img width="1064" height="615" alt="Image" src="https://github.com/user-attachments/assets/913db956-af78-4914-8584-e83dcdf5a829" />

### Reproduction steps

1. Build a bot using `AWSBedrockLLMService`
2. Add params to it `retry_timeout_secs=4.5`, `retry_on_timeout=True`
3. Watch it fail to start up (see logs)

### Expected behavior

App works

### Actual behavior

App crashes (see logs attached)

### Logs

```shell
2025-09-02 12:24:26.366 | ERROR    | pipecat.services.aws.llm:_process_context:1097 - CustomAWSBedrockLLMService#0 exception: object dict can't be used in 'await' expression
Traceback (most recent call last):

  File "/Users/*/Development/*/services/pipecat-ai-moderator/src/run.py", line 228, in <module>
    main()
    └ <function main at 0x1345e8ea0>

  File "/Users/*/Development/*/services/pipecat-ai-moderator/src/run.py", line 224, in main
    uvicorn.run(app, host=args.host, port=args.port)
    │       │   │         │    │          │    └ 7860
    │       │   │         │    │          └ Namespace(host='0.0.0.0', port=7860)
    │       │   │         │    └ '0.0.0.0'
    │       │   │         └ Namespace(host='0.0.0.0', port=7860)
    │       │   └ <fastapi.applications.FastAPI object at 0x1345b1400>
    │       └ <function run at 0x104ad9d00>
    └ <module 'uvicorn' from '/Users/*/Development/*/services/pipecat-ai-moderator/.venv/lib/python3.13/si...

  File "/Users/*/Development/*/services/pipecat-ai-moderator/.venv/lib/python3.13/site-packages/uvicorn/main.py", line 579, in run
    server.run()
    │      └ <function Server.run at 0x104b18e00>
    └ <uvicorn.server.Server object at 0x1345b2e40>
[...]

PR #2568, opened 2025-09-03
fix timeout handling in AWSBedrockLLMService

ezisezis:
# Fix double await in AWSBedrockLLMService._create_converse_stream

## Problem

Fixes issue https://github.com/pipecat-ai/pipecat/issues/2558.

When using `retry_timeout_secs` and `retry_on_timeout` parameters with `AWSBedrockLLMService`, the following error occurs:
````
TypeError: object dict can't be used in 'await' expression
````

The error is caused by a double `await` on line 883 in `pipecat/services/aws/llm.py`:

```python
response = await asyncio.wait_for(
    await client.converse_stream(**request_params), timeout=self._retry_timeout_secs
)
```

## Root Cause

The issue occurs because:

1. `client.converse_stream(**request_params)` returns a coroutine
2. The inner `await client.converse_stream(**request_params)` resolves this coroutine to a dictionary response
3. `asyncio.wait_for()` then tries to await this dictionary, which is not awaitable
4. This results in the TypeError: "object dict can't be used in 'await' expression"

## Solution

Remove the inner `await` and pass the coroutine directly to `asyncio.wait_for()`:

```python
response = await asyncio.wait_for(
    client.converse_stream(**request_params), timeout=self._retry_timeout_secs
)
```

This allows `asyncio.wait_for()` to properly handle the timeout on the coroutine execution.

## Testing

- ✅ Verified the fix resolves the TypeError when using retry parameters
- ✅ Confirmed timeout functionality works as expected
- ✅ Retry logic executes properly on timeout
- ✅ Normal operation (without timeout) remains unaffected

## Impact

- **Breaking Changes**: None
- **Backward Compatibility**: Fully maintained
- **Performance**: No impact on performance, only fixes broken functionality

This fix enables users to properly utilize the timeout and retry functionality in AWS Bedrock LLM service without encountering runtime errors.

### DIFF
--- a/src/pipecat/services/aws/llm.py
+++ b/src/pipecat/services/aws/llm.py
@@ -880,7 +880,7 @@ class AWSBedrockLLMService(LLMService):
         if self._retry_on_timeout:
             try:
                 response = await asyncio.wait_for(
-                    await client.converse_stream(**request_params), timeout=self._retry_timeout_secs
+                    client.converse_stream(**request_params), timeout=self._retry_timeout_secs
                 )
                 return response
             except (ReadTimeoutError, asyncio.TimeoutError) as e:


aconchillo:
Thank you! LGTM!

PR #2574, opened 2025-09-03
Pk/expand universal llm context support to anthropic

kompfner:


src/pipecat/services/anthropic/llm.py
```diff
@@ -294,24 +320,22 @@ async def _process_context(self, context: OpenAILLMContext):
...
+                f"{self}: Generating chat from {context_type_for_logging} context [{params_from_context['system']}] | {messages_for_logging}"
             )
 
-            messages = context.messages
-            if self._settings["enable_prompt_caching_beta"]:
-                messages = context.get_messages_with_cache_control_markers()
-
-            api_call = self._client.messages.create
-            if self._settings["enable_prompt_caching_beta"]:
-                api_call = self._client.beta.prompt_caching.messages.create
```
kompfner, line 307: Note that there was no need to use the `beta` client when using cache control—it’s been GA for a while, and is GA in the Anthropic library version we depend on (0.49)

src/pipecat/adapters/services/anthropic_adapter.py
```diff
@@ -56,7 +82,228 @@ def get_messages_for_logging(self, context) -> List[Dict[str, Any]]:
...
+        if isinstance(content, str):
+            # fix empty text
+            if content == "":
+                content = "(empty)"
+        elif isinstance(content, list):
+            for item in content:
+                # fix empty text
+                if item["type"] == "text" and item["text"] == "":
+                    item["text"] = "(empty)"
+                # handle image_url -> image conversion
```
kompfner, line 251: Note that image message handling (seen here) had never been implemented in the translation-from-OpenAI codepath in `AnthropicLLMContext` (there was a TODO), so this is new. I didn't bother to back-port it.

examples/foundational/14b-function-calling-anthropic-video.py
```diff
@@ -97,7 +97,7 @@ async def run_bot(transport: BaseTransport, runner_args: RunnerArguments):
     llm = AnthropicLLMService(
         api_key=os.getenv("ANTHROPIC_API_KEY"),
         model="claude-3-7-sonnet-latest",
-        enable_prompt_caching_beta=True,
+        params=AnthropicLLMService.InputParams(enable_prompt_caching_beta=True),
```
kompfner: Found this oopsie while validating Anthropic cache control

src/pipecat/adapters/services/anthropic_adapter.py
```diff
@@ -56,7 +82,228 @@ def get_messages_for_logging(self, context) -> List[Dict[str, Any]]:
...
+                img_idx = image_indices[0]
+                first_txt_idx = text_indices[0]
+                if img_idx > first_txt_idx:
+                    # Move image before the first text
+                    image_item = content.pop(img_idx)
+                    content.insert(first_txt_idx, image_item)
+
+        return message
+
+    def _with_cache_control_markers(self, messages: List[MessageParam]) -> List[MessageParam]:
```
kompfner, line 276: This is substantially different from what we did before in `AnthropicLLMContext`. 

There, the solution was a stateful thing where a run of `_process_context()` was being treated as a "turn" and a run of `context.set_messages()` was being treated as a "reset" of the context.

This bookkeeping was needed in order to maintain a `turns_above_cache_threshold` state. Anthropic only caches prompts that are bigger than 1024 input tokens, so the logic was there to avoid asking it to cache prompts smaller than that.

In order to use universal `LLMContext`, we can't be tracking state in our context—the LLM invocation has to just be a function of the context.

Fortunately, it turns out that we could entirely ditch our prior bookkeeping! We can simply **always** apply the cache control markers (when caching is on) rather than tracking input token length, since Anthropic simply ignores those markers when the prompt is too short ([see docs](https://docs.anthropic.com/en/docs/build-with-claude/prompt-caching#cache-limitations)). Maybe this is an improvement they made since this was in beta, when the `turns_above_cache_threshold` code was written.

src/pipecat/adapters/services/anthropic_adapter.py
```diff
@@ -56,7 +82,241 @@ def get_messages_for_logging(self, context) -> List[Dict[str, Any]]:
...
+                # If we have only have a system message in the list, all we can really do
+                # without introducing too much magic is change the role to "user".
+                messages[0]["role"] = "user"
+            else:
+                # If we have more than one message, we'll pull the system message out of the
+                # list.
+                system = messages[0]["content"]
+                messages.pop(0)
+
+        # Convert any subsequent "system"-role messages to "user"-role
```
kompfner, line 136: This is new. It allows us to use "system"-role messages in universal context without having to think too hard about which LLM we're going to be using. This matches the Gemini adapter behavior now.

examples/foundational/14b-function-calling-anthropic-video.py
kompfner: Side note: the plan is to rename `enable_prompt_caching_beta` to just `enable_prompt_caching` (deprecating but still allowing for now the old name).

This might come in this PR, or in a subsequent one.

src/pipecat/services/anthropic/llm.py
```diff
@@ -112,15 +115,21 @@ class InputParams(BaseModel):
         """Input parameters for Anthropic model inference.
 
         Parameters:
-            enable_prompt_caching_beta: Whether to enable beta prompt caching feature.
+            enable_prompt_caching: Whether to enable the prompt caching feature.
+            enable_prompt_caching_beta (deprecated): Whether to enable the beta prompt caching feature.
+
+                .. deprecated:: 0.0.83
```
kompfner: Ah looks like I'll have to bump to 0.0.84

examples/foundational/14b-function-calling-anthropic-video.py
kompfner: Went ahead and included the rename to `enable_prompt_caching` in this PR.

src/pipecat/services/anthropic/llm.py
kompfner: Done.

src/pipecat/adapters/services/anthropic_adapter.py
markbackman: This makes sense.

src/pipecat/services/anthropic/llm.py
```diff
@@ -112,15 +115,21 @@ class InputParams(BaseModel):
...
             max_tokens: Maximum tokens to generate. Must be at least 1.
             temperature: Sampling temperature between 0.0 and 1.0.
             top_k: Top-k sampling parameter.
             top_p: Top-p sampling parameter between 0.0 and 1.0.
             extra: Additional parameters to pass to the API.
         """
 
-        enable_prompt_caching_beta: Optional[bool] = False
+        enable_prompt_caching: Optional[bool] = None
+        enable_prompt_caching_beta: Optional[bool] = None
```
markbackman, line 132: We should add a deprecation warning if `enable_prompt_caching_beta` is used.

src/pipecat/services/anthropic/llm.py
```diff
@@ -215,18 +232,18 @@ async def run_inference(
...
+            params = adapter.get_llm_invocation_params(
+                context, enable_prompt_caching=self._settings["enable_prompt_caching"]
+            )
+            messages = params["messages"]
+            system = params["system"]
         else:
             context = AnthropicLLMContext.upgrade_to_anthropic(context)
             messages = context.messages
-            system = getattr(context, "system", None) or system_instruction
+            system = getattr(context, "system", None) or system_instruction or NOT_GIVEN
```
markbackman: I think I missed this in the last review. Providing a `system_instruction` (which is an arg for the `run_inference` method from the LLMService base class) is an explicit way to set the system instructions. Or, for Anthropic, you can provide a list of messages where the first message is `system` and that will be counted as a system message.

Is that right? If so, this makes sense. Just want to make sure I'm following this correctly.

markbackman (approved):
This looks really good! It makes sense and follows the other patterns well.

I tested both examples and all is well. One notable item: when asking Anthropic to use the camera tool, I see this warning: `FunctionCallResultFrame tool_call_id [toolu_01XeXnVPT4CCTmm5zxDMZdoJ] is not running`. I don't see the same for a tool that doesn't involve the camera. The same warning happens on `main`. This isn't a blocker, but would be good to resolve / understand if possible.

kompfner:
> This looks really good! It makes sense and follows the other patterns well.
> 
> I tested both examples and all is well. One notable item: when asking Anthropic to use the camera tool, I see this warning: `FunctionCallResultFrame tool_call_id [toolu_01XeXnVPT4CCTmm5zxDMZdoJ] is not running`. I don't see the same for a tool that doesn't involve the camera. The same warning happens on `main`. This isn't a blocker, but would be good to resolve / understand if possible.

Yeah I've seen that warning with other services, too, IIRC. Doesn't seem to be related to universal v LLM-specific context. But yes, we should come back and resolve it.

src/pipecat/services/anthropic/llm.py
kompfner: Ah good call

kompfner: Done

src/pipecat/services/anthropic/llm.py
kompfner: That's right! Context-provided one takes precedence.

kompfner: Hmm...looking at this code again it looks like I missed honoring the `system_instruction` arg in the `LLMContext` branch of the logic.

kompfner: Looking at it even more closely, I'm trying to remember why I initially built this with the option to specify `system_instruction` at all...

I'm not sure there's a good enough reason? I suppose it would let people not have to inject an initial "system" message in their context, which might be more idiomatic when using things like `GoogleLLMContext` or `AnthropicLLMContext`? But then again, if they were directly creating one of those contexts in code, they could always assign directly to the `system` property of those contexts...

So...I think we should remove this param, actually. It's not worth the complexity of properly handling (which might involve in some cases—like OpenAI—actually injecting a "system" message into the context).

Since this method was _just_ shipped last week, what do you think about making this a breaking change?

kompfner: I actually now think the `system_instruction` arg was a holdover from when the method was refactored from a more usa-case-specific `generate_summary()`, where it made sense to separate the context from the summary prompt.

markbackman: I agree. I think removing `system_instruction` makes sense. This was not documented and isn't obvious that it can be used without understanding the inheritance from LLMService. I think sticking to the messages is the right way to go.

PR #2576, opened 2025-09-03
fix: update default Cerebras model to GPT-OSS-120B

joyceerhl:

`llama-3.3-70b` is sometimes prone to looping on tool calls. This PR updates the default for Cerebras inference to `gpt-oss-120b`, which is supported in production in Cerebras' inference API and exhibits stronger performance on tool calling.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -90,6 +90,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Updated `RimeTTSService`'s flush_audio message to conform with Rime's official API.
 
+- - Updated the default model for `CerebrasLLMService` to GPT-OSS-120B.
+
 ### Removed
 
 - Remove `StopInterruptionFrame`. This was a legacy frame that was not being
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -90,7 +90,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Updated `RimeTTSService`'s flush_audio message to conform with Rime's official API.
 
-- - Updated the default model for `CerebrasLLMService` to GPT-OSS-120B.
+- Updated the default model for `CerebrasLLMService` to GPT-OSS-120B.
 
 ### Removed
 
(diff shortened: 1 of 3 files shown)


markbackman:
When trying `qwen-3-32b`, I see this error:
```
CerebrasLLMService#0: error processing frame: Error code: 400 - {'message': '"stream=true" with structured outputs is currently not supported for reasoning models.', 'type': 'invalid_request_error', 'param': 'tools_mode', 'code': 'wrong_api_format'}
```

Though, when trying both `llama-4-scout-17b-16e-instruct` and `gpt-oss-120b`, I get successful responses and tool calls work. Maybe one of those models is a better option?

markbackman:
Also, along with this change, can you update the corresponding foundational example to remove setting the llama-3.3 model? That change is right here:
https://github.com/pipecat-ai/pipecat/blob/main/examples/foundational/14k-function-calling-cerebras.py#L70

Also, please add a [Changelog](https://github.com/pipecat-ai/pipecat/blob/main/CHANGELOG.md) entry:
```
### Changed

- Updated the default model for `CerebrasLLMService` to MODEL.
```

CHANGELOG.md
```diff
@@ -90,6 +90,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Updated `RimeTTSService`'s flush_audio message to conform with Rime's official API.
 
+- - Updated the default model for `CerebrasLLMService` to GPT-OSS-120B.
```
markbackman: nit: remove extra `-`

```suggestion
- Updated the default model for `CerebrasLLMService` to GPT-OSS-120B.
```

examples/foundational/14k-function-calling-cerebras.py
```diff
@@ -67,7 +67,7 @@ async def run_bot(transport: BaseTransport, runner_args: RunnerArguments):
         voice_id="71a7ad14-091c-4e8e-a314-022ece01c121",  # British Reading Lady
     )
 
-    llm = CerebrasLLMService(api_key=os.getenv("CEREBRAS_API_KEY"), model="llama-3.3-70b")
+    llm = CerebrasLLMService(api_key=os.getenv("CEREBRAS_API_KEY"), model="gpt-oss-120b")
```
markbackman: nit: let's just remove the model all together. This will use the default mode from CerebrasLLMService.
```suggestion
    llm = CerebrasLLMService(api_key=os.getenv("CEREBRAS_API_KEY"))
```

PR #2578, opened 2025-09-03
add UserSpeakingFrame and UserStartedSpeakingFrame/UserStopeedSpeakingFrame updates

aconchillo:

In this PR we make a couple of changes:
- Push UserStartedSpeakingFrame and UserStoppedSpeaking frame upstream.
- Add a new UserSpeakingFrame, similar to BotSpeakingFrame that is pushed upstream and downstream while the user is speaking.
-


markbackman (approved):
Nice improvement! 👏

PR #2583, opened 2025-09-04
fix: openai image generator now initiates URLImageRawFrame with correct order of arguments

sassanh:

The OpenAI image generator was initiating `URLImageRawFrame` with the wrong order of arguments, `image_url` should be last argument if it is initiating it with positional arguments (inherited dataclasses, list their properties after the inherited class in the `__init__` function.)

I made it use keyword arguments to be safer.

If we have a typechecking task using pyright for example in the CI/CD, it should help to prevent similar issues.


aconchillo:
Thank you!

PR #2591, opened 2025-09-04
wait until endframe fully through pipeline before audiobuffer and turn_tracking

vipyne:

in `turn_tracking_observer` and `audio_buffer_processor`, as soon as an `EndFrame` is seen, the turn is ended / the recording stops.  
`EndFrame` usually indicates a graceful shutdown, so there could still be other frames in the pipeline.  this could result in a turn not fully being captured or a recording missing the final turn segment of audio.
I think we can solve this by checking the source of the `EndFrame` and if it is the Pipeline, we can be sure all other frames have been processed.  (Source because we push EndFrames upstream).
wdyt @aconchillo @filipi87 ?

### DIFF
--- a/src/pipecat/observers/turn_tracking_observer.py
+++ b/src/pipecat/observers/turn_tracking_observer.py
@@ -23,6 +23,7 @@ from pipecat.frames.frames import (
     StartFrame,
     UserStartedSpeakingFrame,
 )
+from pipecat.pipeline.pipeline import Pipeline
 from pipecat.observers.base_observer import BaseObserver, FramePushed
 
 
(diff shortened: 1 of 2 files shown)


markbackman:
@vipyne I'm not sure this is the right thing to do. The EndFrame is a ControlFrame and is queued. This allows all frames before it to be processed before the Pipeline ends. For example, you could push a TTSSpeakFrame and EndFrame back to back and be guaranteed that the TTSSpeakFrame will complete before the EndFrame hits all of the processors. This would include the AudioBufferProcessor—all audio created before it would be captured.

Delaying the EndFrame for frames after it to be processed seems like it breaks how the EndFrame functions. I'm happy to defer to @aconchillo, but I think making this change would change how the EndFrame operates.

If this isn't working as described, then there's a problem elsewhere that we should resolve.

aconchillo:
> @vipyne I'm not sure this is the right thing to do. The EndFrame is a ControlFrame and is queued. This allows all frames before it to be processed before the Pipeline ends. For example, you could push a TTSSpeakFrame and EndFrame back to back and be guaranteed that the TTSSpeakFrame will complete before the EndFrame hits all of the processors. This would include the AudioBufferProcessor—all audio created before it would be captured.
> 
> Delaying the EndFrame for frames after it to be processed seems like it breaks how the EndFrame functions. I'm happy to defer to @aconchillo, but I think making this change would change how the EndFrame operates.
> 
> If this isn't working as described, then there's a problem elsewhere that we should resolve.

Yes, `EndFrame` guarantees that there's nothing after it and that everything before it is properly processed by all frame processors. So, it is correct for `AudioBufferProcessor` to shutdown everything when it sees an `EndFrame` because everything else has already been processed.

If that's not the case then it might indicate an issue. Did we see something not working?

vipyne:
discussing internally, closing for now

Issue #2581, opened 2025-09-03
Add support for gpt-realtime


dtran320:
### Service Name

OpenAI gpt-realtime

### Service Website

https://platform.openai.com/docs/guides/realtime-conversations

### Service Description

`gpt-realtime` generally available as of August 28th, beats `gpt-4o-realtime-preview` in [both pricing and performance](https://platform.openai.com/docs/models/compare?model=gpt-4o-realtime-preview). Can we drop-in replace the model `gpt-4o-realtime-preview-2025-06-03` in `OpenAIRealtimeBetaLLMService` or should we create a new `OpenAIRealtimeLLMService`?

<img width="908" height="868" alt="Image" src="https://github.com/user-attachments/assets/6cc1cc99-b29f-4b91-a309-568af0a78dde" />

### Would you be willing to help implement this service?

- [x] Yes, I'd like to contribute
- [ ] No, I'm just suggesting

CodeNameR:
We need this too - 
There are some guidelines here for Beta to GA migration
https://platform.openai.com/docs/guides/realtime

markbackman:
PR with this update: https://github.com/pipecat-ai/pipecat/pull/2596

PR #2596, opened 2025-09-05
Add OpenAIRealtimeLLMService, AzureRealtimeLLMService

markbackman:

For reviewing, see the [first commit](https://github.com/pipecat-ai/pipecat/pull/2596/commits/4b5967c82241e8fe016e8cd232ec9ca4a6d6439d) for the changes in comparison to the OpenAIRealtimeBetaLLMService. Relevant materials used:
- Migration guidance: https://platform.openai.com/docs/guides/realtime#beta-to-ga-migration
- Reference docs: https://platform.openai.com/docs/api-reference/realtime

Largely, the change is to keys and API shape. This results in a number of new classes to support the nested `audio` object.

It looks like OpenAI is moving towards using `.added` and `.done` for their response and conversation APIs. This replaces `.created`, which was removed as the event is no longer in the reference docs for the GA API. At the moment, `.done` is a no-op as this port was to maintain existing functionality.

One area that changed in a surprising way was function calling. I'm now relying on the function call arguments `.done` event to indicate when the function call completes and is ready to use. The prior approached used `_handle_assistant_output` which could occur in multiple places. This did not work with the API changes, so it was removed, as it only handled function calls. Now, functions are called as a result of the `response.function_call_arguments.done` which is a bit cleaner. Like other LLMs, we do track pending calls and remove them as they're being called. Perhaps this could be done differently, but it works reliably in my testing.

Given the large divergence of the Beta and GA APIs, I'm opting to keep them in separate modules. This will allow OpenAIRealtimeBetaLLMService to keep working, now with a deprecation warning. We can remove the module with the 1.0.0 release, which is when I'm marking this as EOL.

I'm also making similar changes to AzureOpenAIRealtimeBetaLLMService. Though, it appears we don't have access to the model yet, so I can't test. We'll need to contact Microsoft to get access.

Resolves https://github.com/pipecat-ai/pipecat/issues/2581

### DIFF
--- a/src/pipecat/services/openai_realtime/openai.py
+++ b/src/pipecat/services/openai_realtime/openai.py
@@ -4,7 +4,7 @@
 # SPDX-License-Identifier: BSD 2-Clause License
 #
 
-"""OpenAI Realtime Beta LLM service implementation with WebSocket support."""
+"""OpenAI Realtime LLM service implementation with WebSocket support."""
 
 import base64
 import json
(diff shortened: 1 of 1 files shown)


kwindla (approved):
lgtm

PR #2597, opened 2025-09-05
Fix Anthropic tool-less usage

kompfner:


kompfner:
I accidentally introduced this bug in https://github.com/pipecat-ai/pipecat/pull/2574. Not updating the CHANGELOG since this bug was never published.

PR #2598, opened 2025-09-05
Remove `VisionImageRawFrame`, which was previously being handled dire…

kompfner:
…ctly by the LLM services, and deprecate the associated `VisionImageFrameAggregator`.

Removing `VisionImageRawFrame` lets us simplify LLM services' logic, getting us closer to the idealized architecture where all they care about is handling context frames.

This change is in service of getting us closer to ready to deprecate usage of `OpenAILLMContext` and subclasses in favor of the universal `LLMContext`, at least for the traditional text-to-text LLMs.

Why remove `VisionImageRawFrame` rather than deprecate? It's "internal"—only created by `VisionImageFrameAggregator`—and never intended to be used directly by users (it would be difficult to use directly anyway).

Move the logic that was once in `VisionImageFrameAggregator` directly into the examples. Reasoning:
- If `UserImageRequester` is defined in the examples, it makes sense for `UserImageProcessor` to be too, as it’s the flip side of the same coin, so to speak
- The logic is now pretty trivial
- This kind of one-shot, history-less image-describing pipeline shouldn't be common at all; it's ok for it to live in examples rather than as a dedicated class
- In the short term, this enables us to create `LLMContext`s for services that support it and `OpenAILLMContext`s for services that don't yet (AWS)

This commit also adds missing translation from OpenAI-format image context messages to AWS format. Note that this isn't a wasted effort in the face of the upcoming migration to universal `LLMContext`—this work will be reused as it has to be implemented there too.


src/pipecat/services/aws/llm.py
```diff
@@ -399,9 +382,34 @@ def from_standard_message(self, message):
         elif isinstance(content, list):
             new_content = []
             for item in content:
+                # fix empty text
                 if item.get("type", "") == "text":
                     text_content = item["text"] if item["text"] != "" else "(empty)"
                     new_content.append({"text": text_content})
+                # handle image_url -> image conversion
```
kompfner, line 389: As described in the PR notes, this was missing before. This will be ported over to the universal `LLMContext` code paths when they're implemented for AWS.

kompfner: Side note: I believe `AWSBedrockLLMContext.add_image_frame_message()` is currently broken. Not bothering to fix it since we hope to move away from direct use of that LLM-specific context soon

src/pipecat/services/aws/llm.py
```diff
@@ -967,7 +975,9 @@ async def _process_context(self, context: AWSBedrockLLMContext):
             }
 
             # Add system message
-            request_params["system"] = context.system
+            system = getattr(context, "system", None)
```
kompfner, line 977: This was a bug I discovered along the way. When we upgrade from `OpenAILLMContext` to `AWSBedrockLLMContext` we never properly initialize the `system` field. This felt like the lowest-risk way to address the problem.

This will be moot once we migrate to the universal `LLMContext`.

src/pipecat/services/anthropic/llm.py
```diff
@@ -495,12 +494,6 @@ async def process_frame(self, frame: Frame, direction: FrameDirection):
             context = frame.context
         elif isinstance(frame, LLMMessagesFrame):
             context = AnthropicLLMContext.from_messages(frame.messages)
-        elif isinstance(frame, VisionImageRawFrame):
```
kompfner, line 498: Removing this block of code across all the LLM services was the initial impetus for this PR.

CHANGELOG.md
```diff
@@ -20,6 +20,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+### Removed
+
+- Remove `VisionImageRawFrame` in favor of context frames (`LLMContextFrame` or
+  `OpenAILLMContextFrame`).
+
+### Deprecated
+
+- Deprecate `VisionImageFrameAggregator` because `VisionImageRawFrame` has been
+  removed. See the `12*` examples for the new recommended replacement pattern.
+
```
aconchillo, line 26: Needs to be rebased to put it in the right place.

aconchillo:
Nice! LGTM!

CHANGELOG.md
kompfner: Good catch!

src/pipecat/processors/aggregators/vision_image_frame.py
```diff
@@ -10,13 +10,22 @@
...
+    OpenAILLMContext,
+    OpenAILLMContextFrame,
+)
 from pipecat.processors.frame_processor import FrameDirection, FrameProcessor
 
 
 class VisionImageFrameAggregator(FrameProcessor):
     """Aggregates consecutive text and image frames into vision frames.
 
+    .. deprecated:: 0.84.0
```
kompfner: Ah, I need to update this version now.

kompfner: (Not to mention I got the version wrong here 😬 )

PR #2599, opened 2025-09-05
DTMF: Add support for native DTMF implementation where available

markbackman:

Since Daily supports RTP events, we should use them over tones. This change restores Daily's use of the send_dtmf functionality. It preserves the audio tones for websocket services, which can only send MediaStreams.

This change allows for new transports to declare native support and override, applying their own DTMF functionality. For example, we would use Livekit's native DTMF, not tones. But, until we implement that native capability, Livekit will use Pipecat's audio tones, which serves as a fallback.

### DIFF
--- a/src/pipecat/transports/base_output.py
+++ b/src/pipecat/transports/base_output.py
@@ -219,7 +219,34 @@ class BaseOutputTransport(FrameProcessor):
         pass
 
     async def write_dtmf(self, frame: OutputDTMFFrame | OutputDTMFUrgentFrame):
-        """Write a DTMF tone to the transport.
+        """Write a DTMF tone using the transport's preferred method.
+
+        Args:
+            frame: The DTMF frame to write.
+        """
+        if self._supports_native_dtmf():
+            await self._write_dtmf_native(frame)
+        else:
+            await self._write_dtmf_audio(frame)
+
+    def _supports_native_dtmf(self) -> bool:
+        """Override in transport implementations that support native DTMF.
+
+        Returns:
+            True if the transport supports native DTMF, False otherwise.
+        """
+        return False
+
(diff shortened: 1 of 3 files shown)


chadbailey59:
I would have thought we could Just™ implement the audio version directly in the `write_dtmf` function of the base class and override it in the Daily transport, but I didn't look at the inheritance hierarchy enough to know if that's actually possible or not. This works too 👍

chadbailey59 (approved):

chadbailey59:
The only possibly-unexpected behavior is that with Daily `send_dtmf`, other non-phone call participants (like a third party in the call in a browser) won't actually hear the tones. They do hear them with the audio version, though.

src/pipecat/transports/base_output.py
```diff
@@ -219,7 +219,34 @@ async def write_audio_frame(self, frame: OutputAudioRawFrame):
...
+
+        Args:
+            frame: The DTMF frame to write.
+        """
+        if self._supports_native_dtmf():
+            await self._write_dtmf_native(frame)
+        else:
+            await self._write_dtmf_audio(frame)
+
+    def _supports_native_dtmf(self) -> bool:
```
aconchillo, line 232: Let's make this public

aconchillo: Oh... I see. nvm

aconchillo (approved):

aconchillo:
LGTM

markbackman:
> The only possibly-unexpected behavior is that with Daily `send_dtmf`, other non-phone call participants (like a third party in the call in a browser) won't actually hear the tones. They do hear them with the audio version, though.

@chadbailey59 this was the case before the change we made last week.

PR #2603, opened 2025-09-06
Add SelfClosingTagAggregator, example 45, and unit tests

markbackman:

But, this enables you to use Cartesia's <break /> tags to insert pauses. Currently, the LLM tokens are streamed and a floating point number for time will be seen as an end of sentence marker (e.g. `<break time="0.1s"/>`). This results in the pause not being inserted and the break tag being spoken. With this new aggregator, it allows the TTS to skip over self closing tags, enabling features like Cartesia's break feature to work properly.

I'm not sure about the name... This is a take on the SkipTagAggregator, but for self closing tags. SelfClosingSkipTagAggregator is too long. Any thoughts on names?

### DIFF
--- a/src/pipecat/utils/text/self_closing_tag_aggregator.py
+++ b/src/pipecat/utils/text/self_closing_tag_aggregator.py
@@ -1,0 +1,162 @@
+#
+# Copyright (c) 2024–2025, Daily
+#
+# SPDX-License-Identifier: BSD 2-Clause License
+#
+
+"""Self-closing tag aggregator for handling XML-style self-closing tags.
+
+This module provides a generic text aggregator that can handle any self-closing
+XML-style tags (e.g., <break time="0.1s"/>, <pause duration="500ms"/>, etc.)
+that should prevent sentence boundary detection when incomplete during streaming.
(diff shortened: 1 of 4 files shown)


src/pipecat/utils/text/self_closing_tag_aggregator.py
```diff
@@ -0,0 +1,162 @@
...
+        # For custom patterns
+        aggregator = SelfClosingTagAggregator(
+            patterns=[r'<break\\s+time="[^"]*"\\s*/?>', r'<pause\\s+duration="[^"]*"\\s*/>']
+        )
+    """
+
+    def __init__(
+        self,
+        tags: Optional[List[str]] = None,
+        patterns: Optional[List[str]] = None,
```
filipi87, line 52: Should it be possible to provide both?

tests/test_self_closing_tag_aggregator.py
filipi87, line 1: 🎉🚀

markbackman:
This is no longer needed now that we have more capable aggregation and text transformation.

Issue #2522, opened 2025-08-27
Duplicate Participant livekit transport layer


anand-therattil:
### pipecat version

0.0.80

### Python version

3.10

### Question

During the Self hosted livekit server facing an issue with the duplicate participant 
```
2025-08-27 12:24:22,543 - faster_whisper - INFO - Processing audio with duration 00:01.483
2025-08-27 12:24:26,650 - livekit - WARNING - livekit::rtc_engine:388:livekit::rtc_engine - failed to connect: Connection("wait_pc_connection timed out"), retrying... (1/3)
2025-08-27 12:24:26,651 - livekit - INFO - livekit_api::signal_client::signal_stream:96:livekit_api::signal_client::signal_stream - connecting to ws://localhost:7880/rtc?sdk=python&protocol=15&auto_subscribe=1&adaptive_stream=0&version=0.22.0&access_token=...
2025-08-27 12:24:26,656 - livekit - WARNING - livekit::rtc_engine:446:livekit::rtc_engine - received session close: "signal client closed: \"stream closed\"" UnknownReason Resume
2025-08-27 12:24:26,661 - livekit - INFO - livekit::room:1191:livekit::room - disconnected from room with reason: DuplicateIdentity
```

### What I've tried

tested using different token but livekit only accepts the token only once 
the livekit is hosted on the local system 
able to see the models everything getting loaded but also able to see the agent is connected to the room 
but then the duplicate error comes up

markbackman:
This may be a good question for the Livekit team.

anand-therattil:
ok thank @markbackman will raise it with the livekit team

murugancmi:
@markbackman it was working fine 0.0.76 but when we upgrade 0.0.81 and 0.0.80 its not working

markbackman:
I've tested using 0.0.82 and main using the 04b example and Livekit Meet. This works as expected for me.
- 04b: https://github.com/pipecat-ai/pipecat/blob/main/examples/foundational/04b-transports-livekit.py
- Livekit Meet: https://meet.livekit.io/

Are you able to repro the issue using that example?

(I'm using Livekit Cloud.)

sender001:
We hit the same issue. It happens when a previous room session is still alive and you try to join another room with the same participant identity.

What happened on our side:
  
  We started a LiveKit room with participant identity “ABC”.
  
  The participant left and joined a different room (still “ABC”).
  
[...]

PR #2604, opened 2025-09-06
Feature: Add support for pipecat video stream and minor bug fixes

alexyzhou:
[Feature]
- Add support for livekit video input-stream. If the param video_in_enabled is set to True, the input transport will generate UserImageRawFrame.
[Fix]
- Fix the issue of duplicate participant when connecting to livekit [issue](https://github.com/pipecat-ai/pipecat/issues/2522)
The root cause is LiveKitInputTransport and LiveKitOutputTransport both tries to call _client.connect() method when receiving StartFrame. And in some cases the function will be called simultaneously, resulting in this issue. I used asyncio.Lock to make sure only one call can pass to this function at a time.
- Fix the issue of message encoding error, especially for rtvi messages (normally represented as python dict), which will cause message encode error in send_message function in LiveKitOutputTransport

### DIFF
--- a/src/pipecat/transports/livekit/transport.py
+++ b/src/pipecat/transports/livekit/transport.py
@@ -12,8 +12,8 @@ event handling for conversational AI applications.
 """
 
 import asyncio
-from dataclasses import dataclass
 import json
+from dataclasses import dataclass
 from typing import Any, Awaitable, Callable, List, Optional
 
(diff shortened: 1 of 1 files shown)


aconchillo:
LGTM! Thank you!

alexyzhou:
Thanks for reviewing!

PR #2608, opened 2025-09-07
audio(filters): deprecate NoisereduceFilter

aconchillo:

Noisereduce doesn't work well for real-time.


aconchillo:
> LGTM. Can you add a changelog entry too?

Added

PR #2612, opened 2025-09-08
Integrating the new Deepgram model (Flux) with Pipecat

filipi87:
Integrating the new Deepgram model (Flux) with Pipecat: 
  - Creating `WebsocketSTTService` base class.
  - Deepgram Flux speech-to-text service implementation.
  - New example using DeepgramFluxSTTService

References:
- https://developers.deepgram.com/flux-early-access

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- Added `WebsocketSTTService` base class for websocket-based STT services.
+  Combines STT functionality with websocket connectivity, providing automatic
+  error handling and reconnection capabilities with exponential backoff.
+
+- Added `DeepgramFluxSTTService` for real-time speech recognition using
+  Deepgram's Flux WebSocket API. Flux understands conversational flow and
+  automatically handles turn-taking.
+
 - Added RTVI messages for user/bot audio levels and system logs.
 
 - Include OpenAI-based LLM services cached tokens to `MetricsFrame`.
--- a/examples/foundational/07c-interruptible-deepgram-flux.py
+++ b/examples/foundational/07c-interruptible-deepgram-flux.py
@@ -1,0 +1,118 @@
+#
+# Copyright (c) 2024–2025, Daily
+#
+# SPDX-License-Identifier: BSD 2-Clause License
+#
+
+
+import os
+
+from dotenv import load_dotenv
+from loguru import logger
+
+from pipecat.frames.frames import LLMRunFrame
+from pipecat.pipeline.pipeline import Pipeline
+from pipecat.pipeline.runner import PipelineRunner
+from pipecat.pipeline.task import PipelineParams, PipelineTask
+from pipecat.processors.aggregators.llm_response_universal import (
+    LLMContext,
+    LLMContextAggregatorPair,
+)
+from pipecat.runner.types import RunnerArguments
+from pipecat.runner.utils import create_transport
+from pipecat.services.deepgram.flux.stt import DeepgramFluxSTTService
+from pipecat.services.deepgram.tts import DeepgramTTSService
+from pipecat.services.openai.llm import OpenAILLMService
+from pipecat.transports.base_transport import BaseTransport, TransportParams
+from pipecat.transports.daily.transport import DailyParams
+from pipecat.transports.websocket.fastapi import FastAPIWebsocketParams
+
+load_dotenv(override=True)
+
+
+# We store functions so objects (e.g. SileroVADAnalyzer) don't get
+# instantiated. The function will be called when the desired transport gets
+# selected.
+transport_params = {
+    "daily": lambda: DailyParams(
+        audio_in_enabled=True,
+        audio_out_enabled=True,
+    ),
+    "twilio": lambda: FastAPIWebsocketParams(
+        audio_in_enabled=True,
+        audio_out_enabled=True,
+    ),
+    "webrtc": lambda: TransportParams(
+        audio_in_enabled=True,
+        audio_out_enabled=True,
+    ),
+}
+
+
+async def run_bot(transport: BaseTransport, runner_args: RunnerArguments):
+    logger.info(f"Starting bot")
+
+    stt = DeepgramFluxSTTService(api_key=os.getenv("DEEPGRAM_API_KEY"))
+
+    tts = DeepgramTTSService(api_key=os.getenv("DEEPGRAM_API_KEY"), voice="aura-2-andromeda-en")
+
+    llm = OpenAILLMService(api_key=os.getenv("OPENAI_API_KEY"))
+
+    messages = [
+        {
+            "role": "system",
+            "content": "You are a helpful LLM in a WebRTC call. Your goal is to demonstrate your capabilities in a succinct way. Your output will be converted to audio so don't include special characters in your answers. Respond to what the user said in a creative and helpful way.",
+        },
+    ]
+
+    context = LLMContext(messages)
+    context_aggregator = LLMContextAggregatorPair(context)
+
+    pipeline = Pipeline(
+        [
+            transport.input(),  # Transport user input
+            stt,  # STT
+            context_aggregator.user(),  # User responses
+            llm,  # LLM
+            tts,  # TTS
+            transport.output(),  # Transport bot output
+            context_aggregator.assistant(),  # Assistant spoken responses
+        ]
+    )
+
+    task = PipelineTask(
+        pipeline,
+        params=PipelineParams(
+            enable_metrics=True,
+            enable_usage_metrics=True,
+        ),
+        idle_timeout_secs=runner_args.pipeline_idle_timeout_secs,
+    )
+
+    @transport.event_handler("on_client_connected")
+    async def on_client_connected(transport, client):
+        logger.info(f"Client connected")
+        # Kick off the conversation.
+        messages.append({"role": "system", "content": "Please introduce yourself to the user."})
+        await task.queue_frames([LLMRunFrame()])
+
+    @transport.event_handler("on_client_disconnected")
+    async def on_client_disconnected(transport, client):
+        logger.info(f"Client disconnected")
+        await task.cancel()
+
+    runner = PipelineRunner(handle_sigint=runner_args.handle_sigint)
+
+    await runner.run(task)
+
+
+async def bot(runner_args: RunnerArguments):
+    """Main bot entry point compatible with Pipecat Cloud."""
+    transport = await create_transport(runner_args, transport_params)
+    await run_bot(transport, runner_args)
+
+
+if __name__ == "__main__":
+    from pipecat.runner.run import main
+
+    main()
--- a/src/pipecat/services/deepgram/flux/stt.py
+++ b/src/pipecat/services/deepgram/flux/stt.py
@@ -1,0 +1,636 @@
+#
+# Copyright (c) 2024–2025, Daily
+#
+# SPDX-License-Identifier: BSD 2-Clause License
+#
+
+"""Deepgram Flux speech-to-text service implementation."""
+
+import json
+from enum import Enum
+from typing import Any, AsyncGenerator, Dict, Optional
+
+from loguru import logger
+from pydantic import BaseModel
+
+from pipecat.frames.frames import (
+    CancelFrame,
+    EndFrame,
+    ErrorFrame,
+    Frame,
+    InterimTranscriptionFrame,
+    StartFrame,
+    TranscriptionFrame,
+    UserStartedSpeakingFrame,
+    UserStoppedSpeakingFrame,
+)
+from pipecat.processors.frame_processor import FrameDirection
+from pipecat.services.stt_service import WebsocketSTTService
+from pipecat.transcriptions.language import Language
+from pipecat.utils.time import time_now_iso8601
+from pipecat.utils.tracing.service_decorators import traced_stt
+
+try:
+    from websockets.asyncio.client import connect as websocket_connect
+    from websockets.protocol import State
+except ModuleNotFoundError as e:
+    logger.error(f"Exception: {e}")
+    logger.error("In order to use Deepgram Flux, you need to `pip install pipecat-ai[deepgram]`.")
+    raise Exception(f"Missing module: {e}")
+
+
+class FluxMessageType(str, Enum):
+    """Deepgram Flux WebSocket message types.
+
+    These are the top-level message types that can be received from the
+    Deepgram Flux WebSocket connection.
+    """
+
+    RECEIVE_CONNECTED = "Connected"
+    RECEIVE_FATAL_ERROR = "Error"
+    TURN_INFO = "TurnInfo"
+
+
+class FluxEventType(str, Enum):
+    """Deepgram Flux TurnInfo event types.
+
+    These events are contained within TurnInfo messages and indicate
+    different stages of speech processing and turn detection.
+    """
+
+    START_OF_TURN = "StartOfTurn"
+    TURN_RESUMED = "TurnResumed"
+    END_OF_TURN = "EndOfTurn"
+    EAGER_END_OF_TURN = "EagerEndOfTurn"
+    UPDATE = "Update"
+
+
+class DeepgramFluxSTTService(WebsocketSTTService):
+    """Deepgram Flux speech-to-text service.
+
+    Provides real-time speech recognition using Deepgram's WebSocket API with Flux capabilities.
+    Supports configurable models, VAD events, and various audio processing options
+    including advanced turn detection and EagerEndOfTurn events for improved conversational AI performance.
+    """
+
+    class InputParams(BaseModel):
+        """Configuration parameters for Deepgram Flux API.
+
+        This class defines all available connection parameters for the Deepgram Flux API
+        based on the official documentation.
+
+        Attributes:
+            eager_eot_threshold: Optional. EagerEndOfTurn/TurnResumed are off by default.
+                You can turn them on by setting eager_eot_threshold to a valid value.
+                Lower values = more aggressive EagerEndOfTurning (faster response, more LLM calls).
+                Higher values = more conservative EagerEndOfTurning (slower response, fewer LLM calls).
+            eot_threshold: Optional. End-of-turn confidence required to finish a turn (default 0.7).
+                Lower values = turns end sooner (more interruptions, faster responses).
+                Higher values = turns end later (fewer interruptions, more complete utterances).
+            eot_timeout_ms: Optional. Time in milliseconds after speech to finish a turn
+                regardless of EOT confidence (default 5000).
+            keyterm: List of keyterms to boost recognition accuracy for specialized terminology.
+            mip_opt_out: Optional. Opts out requests from the Deepgram Model Improvement Program
+                (default False).
+            tag: List of tags to label requests for identification during usage reporting.
+        """
+
+        eager_eot_threshold: Optional[float] = None
+        eot_threshold: Optional[float] = None
+        eot_timeout_ms: Optional[int] = None
+        keyterm: list = []
+        mip_opt_out: Optional[bool] = None
+        tag: list = []
+
+    def __init__(
+        self,
+        *,
+        api_key: str,
+        url: str = "wss://api.deepgram.com/v2/listen",
+        sample_rate: Optional[int] = None,
+        model: str = "flux-general-en",
+        flux_encoding: str = "linear16",
+        params: Optional[InputParams] = None,
+        **kwargs,
+    ):
+        """Initialize the Deepgram Flux STT service.
+
+        Args:
+            api_key: Deepgram API key for authentication. Required for API access.
+            url: WebSocket URL for the Deepgram Flux API. Defaults to the preview endpoint.
+            sample_rate: Audio sample rate in Hz. If None, uses the rate from params or 16000.
+            model: Deepgram Flux model to use for transcription. Currently only supports "flux-general-en".
+            flux_encoding: Audio encoding format required by Flux API. Must be "linear16".
+                Raw signed little-endian 16-bit PCM encoding.
+            params: InputParams instance containing detailed API configuration options.
+                If None, default parameters will be used.
+            **kwargs: Additional arguments passed to the parent WebsocketSTTService class.
+
+        Example:
+            ```python
+            # Basic usage with default parameters
+            stt = DeepgramFluxSTTService(api_key="your-api-key")
+
+            # Advanced usage with custom parameters
+            params = DeepgramFluxSTTService.InputParams(
+                eager_eot_threshold=0.5,
+                eot_threshold=0.8,
+                keyterm=["AI", "machine learning", "neural network"],
+                tag=["production", "voice-agent"]
+            )
+            stt = DeepgramFluxSTTService(
+                api_key="your-api-key",
+                model="flux-general-en",
+                params=params
+            )
+            ```
+        """
+        super().__init__(sample_rate=sample_rate, **kwargs)
+
+        self._api_key = api_key
+        self._url = url
+        self._model = model
+        self._params = params or DeepgramFluxSTTService.InputParams()
+        self._flux_encoding = flux_encoding
+        # This is the currently only supported language
+        self._language = Language.EN
+        self._websocket_url = None
+        self._receive_task = None
+
+    async def _connect(self):
+        """Connect to WebSocket and start background tasks.
+
+        Establishes the WebSocket connection to the Deepgram Flux API and starts
+        the background task for receiving transcription results.
+        """
+        await self._connect_websocket()
+
+        if self._websocket and not self._receive_task:
+            self._receive_task = self.create_task(self._receive_task_handler(self._report_error))
+
+    async def _disconnect(self):
+        """Disconnect from WebSocket and clean up tasks.
+
+        Gracefully disconnects from the Deepgram Flux API, cancels background tasks,
+        and cleans up resources to prevent memory leaks.
+        """
+        try:
+            # Cancel background tasks BEFORE closing websocket
+            if self._receive_task:
+                await self.cancel_task(self._receive_task, timeout=2.0)
+                self._receive_task = None
+
+            # Now close the websocket
+            await self._disconnect_websocket()
+
+        except Exception as e:
+            logger.error(f"Error during disconnect: {e}")
+        finally:
+            # Reset state only after everything is cleaned up
+            self._websocket = None
+
+    async def _connect_websocket(self):
+        """Establish WebSocket connection to API.
+
+        Creates a WebSocket connection to the Deepgram Flux API using the configured
+        URL and authentication headers. Handles connection errors and reports them
+        through the event handler system.
+        """
+        try:
+            if self._websocket and self._websocket.state is State.OPEN:
+                return
+
+            self._websocket = await websocket_connect(
+                self._websocket_url,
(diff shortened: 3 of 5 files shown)


src/pipecat/services/deepgram/flux/stt.py
```diff
@@ -0,0 +1,642 @@
...
+    """
+
+    class InputParams(BaseModel):
+        """Configuration parameters for Deepgram Flux API.
+
+        This class defines all available connection parameters for the Deepgram Flux API
+        based on the official documentation.
+
+        Attributes:
+            language: Language of the audio input. Defaults to English.
```
nkaimakis: NOTE: Flux currently is English-only. We've also centralized our language/model specification into the `model` param I.e. `model=flux-general-en` rather than Nova-3 style `model=nova-3&language=en`

src/pipecat/services/deepgram/flux/stt.py
```diff
@@ -0,0 +1,642 @@
...
+            flux_encoding: Audio encoding format required by Flux API. Must be "linear16".
+                Raw signed little-endian 16-bit PCM encoding.
+            preflight_threshold: End-of-turn confidence required to fire a preflight event (default 0.3).
```
nkaimakis: as I mentioned in slack, we've decided to turn Preflight/SpeechResumed off by default. to enable, set preflight_threshold to a valid value.

src/pipecat/services/deepgram/flux/stt.py
```diff
@@ -0,0 +1,642 @@
...
+                Lower values = more aggressive preflighting (faster response, more LLM calls).
+                Higher values = more conservative preflighting (slower response, fewer LLM calls).
+            eot_threshold: End-of-turn confidence required to finish a turn (default 0.7).
+                Lower values = turns end sooner (more interruptions, faster responses).
+                Higher values = turns end later (fewer interruptions, more complete utterances).
+            eot_timeout_ms: Time in milliseconds after speech to finish a turn regardless of EOT confidence (default 3000).
```
nkaimakis: we are updating default to 5000 here

src/pipecat/services/deepgram/flux/stt.py
```diff
@@ -0,0 +1,642 @@
...
+        eot_timeout_ms: int = 3000
+        keyterm: list = []
+        mip_opt_out: bool = False
+        tag: list = []
+
+    def __init__(
+        self,
+        *,
+        api_key: str,
+        url: str = "wss://api.preview.deepgram.com/v2/listen",
```
nkaimakis: NOTE: WS url will change to `wss://api.deepgram.com/v2/listen` for GA

src/pipecat/services/deepgram/flux/stt.py
filipi87: Oo I see. I was aware that it was English only, but since I hadn’t found it in your API, I thought this would be the way to set the language in the future, instead of directly in the model.

Cool, I’ll fix it. Thanks for catching it.

src/pipecat/services/deepgram/flux/stt.py
filipi87: I will change it to an environment variable to make it easier to update.

src/pipecat/services/deepgram/flux/stt.py
```diff
@@ -0,0 +1,642 @@
...
+
+        The service stops TTFB (Time To First Byte) metrics when the first
+        substantial update is received, indicating successful processing start.
+
+        Args:
+            transcript: The current partial transcript text for the ongoing turn.
+        """
+        if transcript:
+            logger.trace(f"Update event: {transcript}")
+            await self.stop_ttfb_metrics()
```
bd-g: The first update in a stream is not guaranteed to be an `Update` message. The start of a stream will look as follows:

- Websocket is opened to the server
- Client receives a `Connected` message from the server.
- Client receives a `TurnInfo` message from the server, which will be either an `Update` *or* `StartOfTurn` message.

src/pipecat/services/deepgram/flux/stt.py
```diff
@@ -0,0 +1,642 @@
...
+
+        The service will:
+        - Send a BotInterruptionFrame upstream to stop bot speech
+        - Send a UserStartedSpeakingFrame downstream to notify other components
+        - Start metrics collection for measuring response times
+        """
+        logger.debug("User started speaking")
+        await self.push_frame(BotInterruptionFrame(), FrameDirection.UPSTREAM)
+        await self.push_frame(UserStartedSpeakingFrame(), FrameDirection.DOWNSTREAM)
+        await self.start_metrics()
```
bd-g: A `StartOfTurn` message can include the first few words of user speech. Do we want to pass the `transcript` into this function for `trace` logging, similar to what we do for the `Update` message?

src/pipecat/services/deepgram/flux/stt.py
```diff
@@ -0,0 +1,642 @@
...
+class FluxMessageType(str, Enum):
+    """Deepgram Flux WebSocket message types.
+
+    These are the top-level message types that can be received from the
+    Deepgram Flux WebSocket connection.
+    """
+
+    RECEIVE_CONNECTED = "receiveConnected"
+    RECEIVE_FATAL_ERROR = "receiveFatalError"
+    TURN_INFO = "TurnInfo"
```
bd-g:
```suggestion
    RECEIVE_CONNECTED = "Connected"
    RECEIVE_FATAL_ERROR = "Error"
    TURN_INFO = "TurnInfo"
```

If I'm reading this right, these values should match the `type` field on these message types from the API spec :smile:

filipi87: Yeah true. I think I got confused by the way the types are named here: 
- https://deepgram-preview-ade746fd-c854-4268-8be6-81f95f55ded3.docs.buildwithfern.com/reference/flux-early-access/flux-early-access#receive
 
Inside each object the types are correct, like you mentioned, but the section headers don’t match them, which is probably what led me to this mistake. 🤕

src/pipecat/services/deepgram/flux/stt.py
filipi87: But since we rely on Deepgram to tell us when the user has started or stopped speaking, we start the metrics at `StartOfTurn`. 

Then, inside the `Update`, it’s as if we’ve just received the first transcription for that specific turn. That’s why we stop the TTFB at this point. Like it marks the moment we get the first transcription for that turn.

Does it make sense ?

src/pipecat/services/deepgram/flux/stt.py
bd-g: Great point! That part of the docs can be misleading. 

cc @nkaimakis good feedback here for how we present event types in our docs

src/pipecat/services/deepgram/flux/stt.py
filipi87: The metrics are for each turn.

bd-g: Got it, so the TTFB metric is not for the first meaningful byte for the entire conversation, but the first meaningful byte for the current turn?

If that is the case, that is related to my comment on `StartOfTurn` - the first words of a response are included on that message type. So the TTFB metrics should be recorded on that event instead of `Update` events.

src/pipecat/services/deepgram/flux/stt.py
filipi87: Yeah, we can do that. 👍

What I’m wondering in this case is whether the TTFB metric even makes sense. I thought we’d receive it as soon as you had some VAD, meaning you had identified that the user was speaking, but before any actual transcription. 

So the TTFB would really just be the difference between `StartOfTurn` and the first transcript received inside `Update`.

src/pipecat/services/deepgram/flux/stt.py
filipi87: > Got it, so the TTFB metric is not for the first meaningful byte for the entire conversation, but the first meaningful byte for the current turn?

Exactly.

filipi87: > If that is the case, that is related to my comment on StartOfTurn - the first words of a response are included on that message type. So the TTFB metrics should be recorded on that event instead of Update events.

I have replied in the other comment. Because if this is the case, the TTFB wouldn't make sense I guess.

src/pipecat/services/deepgram/flux/stt.py
bd-g: It's not just VAD based, but yes, this event triggers as soon as we think the User has started speaking. I'll note that "a StartOfTurn message *can* include the first few words of user speech" - it doesn't have to. I think your TTFB metric still makes a lot of sense, to measure the time between when we think the user started speaking to the first word of speech.

I'd suggest something like the following logic:
```
   async def _handle_start_of_turn(self, transcript: str):
        logger.debug(f"User started speaking{': ' + transcript if transcript else ''}")
        await self.push_frame(UserStartedSpeakingFrame(), FrameDirection.DOWNSTREAM)
        await self.start_metrics()
        if transcript:
...

And keep the `stop_ttfb_metrics` call in the `_handle_update` function, since there are many cases where the `StartOfTurn` will be empty and the first word will be passed in the `Update` event.

File: src/pipecat/services/deepgram/flux/stt.py (hunk 7 above)
Comment (filipi87) [reply src/pipecat/services/deepgram/flux/stt.py]: Also, regarding the docs:
- https://deepgram-preview-ade746fd-c854-4268-8be6-81f95f55ded3.docs.buildwithfern.com/reference/flux-early-access/flux-early-access#send
  - It would be great to have examples here showing exactly what we need to send. I guess that would make things a lot easier for the developers.

And one more question regarding the `CloseStream`. In the Pipecat case, we’re always sending audio and only stop when we disconnect. So it’s not totally clear why this message is necessary, unless it’s some kind of “flush”, like saying, “I’m disconnecting, so send any remaining transcription/events you have.”

But maybe it is needed in other use cases. But wondering if we should keep sending it ?

File: src/pipecat/services/deepgram/flux/stt.py (hunk 6 above)
Comment (filipi87) [reply src/pipecat/services/deepgram/flux/stt.py]: Yes, that’s how I was thinking about it too. It does feel a bit odd for the TTFB to be basically “0” in a case where we’re already receiving a transcript, but I guess we can move forward like this. 👍

File: src/pipecat/services/deepgram/flux/stt.py [hunk 8]
```diff
@@ -0,0 +1,642 @@
...
+            f"mip_opt_out={str(self._params.mip_opt_out).lower()}",
+        ]
+
+        # Add keyterm parameters (can have multiple)
+        for keyterm in self._params.keyterm:
+            url_params.append(f"keyterm={keyterm}")
+
+        # Add tag parameters (can have multiple)
+        for tag_value in self._params.tag:
+            url_params.append(f"tag={tag_value}")
```
filipi87, line 282: @bd-g, @nkaimakis, one more question about the docs.

Here: https://developers.deepgram.com/flux-early-access#connection-parameters
, the parameter is described as "tags".

But in the API reference: https://deepgram-preview-ade746fd-c854-4268-8be6-81f95f55ded3.docs.buildwithfern.com/reference/flux-early-access/flux-early-access#request.query.tag
, it’s described as "tag".

I’ve been following the API Reference, but it seems like something worth clarifying or fixing. Also, it’s not entirely clear from the docs whether multiple tags can be passed. 

For "keyterm" you explicitly mention: “Pass multiple keyterm query parameters to boost multiple keyterms,” which makes it clearer.

bd-g: Good callout! We should clarify that.

Both `keyterm` and `tag` should be spelled in their singular form, and both can be passed multiple times (so pass `keyterm=something&keyterm=other` to pass multiple keyterms, and `tag=billing-identifier-a&tag=identifier-b` to pass multiple tags)

Love the docs feedback! :heart:

src/pipecat/services/deepgram/flux/stt.py
filipi87: Done.

src/pipecat/services/deepgram/flux/stt.py
filipi87: Fixed the description.

src/pipecat/services/deepgram/flux/stt.py
filipi87: Done.

src/pipecat/services/deepgram/flux/stt.py
filipi87: Ok, I have created a new environment variable DEEPGRAM_FLUX_URL, and changing the default URL to prod.

src/pipecat/services/deepgram/flux/stt.py
nkaimakis: > I'll note that "a StartOfTurn message can include the first few words of user speech" - it doesn't have to.


Small clarification here -- we actually do guarantee that StartOfTurn will contain a non-empty transcript.

src/pipecat/services/deepgram/flux/stt.py
nkaimakis: > unless it’s some kind of “flush”, like saying, “I’m disconnecting, so send any remaining transcription/events you have.”

This is exactly the intended usage of CloseStream. Upon receiving a CloseStream, the server will process all remaining audio and then gracefully close the session.

src/pipecat/services/deepgram/flux/stt.py
filipi87: If that is the case, not sure if we should keep the TTFB. 

What do you think @markbackman ?

src/pipecat/services/deepgram/flux/stt.py
filipi87: Fixed the events.

src/pipecat/services/deepgram/flux/stt.py
filipi87: I have disabled the TTFB metrics for now and added an explanation in the code.

filipi87:
Hi @nkaimakis, @bd-g, thank you for the great feedback. I believe I have addressed all your comments. Let me know if there is anything else we should improve. 🙂

src/pipecat/services/deepgram/flux/stt.py
```diff
@@ -0,0 +1,630 @@
...
+                A turn will be finished when this much time has passed after speech, regardless of EOT confidence.
+            keyterm: List of keyterms to boost recognition accuracy for specialized terminology.
+                Pass multiple keyterms to improve recognition of domain-specific words.
+            mip_opt_out: Opts out requests from the Deepgram Model Improvement Program (default False).
+                Refer to Deepgram docs for pricing impacts before setting this to True.
+            tag: List of tags to label requests for identification during usage reporting.
+        """
+
+        flux_encoding: str = "linear16"
+        eager_eot_threshold: float = 0.3
```
nkaimakis: are you intentionally enabling the EagerEndOfTurn workflow by default with this on your side?

env.example
```diff
@@ -35,6 +35,9 @@ DAILY_SAMPLE_ROOM_URL=https://...
 # Deepgram
 DEEPGRAM_API_KEY=...
 
+# Deepgram Flux
+DEEPGRAM_FLUX_URL=...
```
markbackman: Can we remove this since the url is now GA?

CHANGELOG.md
```diff
@@ -9,6 +9,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- Added `WebsocketSTTService` base class for websocket-based STT services.
```
markbackman, line 12: Nice!

examples/foundational/07c-interruptible-deepgram-flux.py
```diff
@@ -0,0 +1,120 @@
...
+        audio_out_enabled=True,
+    ),
+}
+
+
+async def run_bot(transport: BaseTransport, runner_args: RunnerArguments):
+    logger.info(f"Starting bot")
+
+    stt = DeepgramFluxSTTService(
+        api_key=os.getenv("DEEPGRAM_API_KEY"), url=os.getenv("DEEPGRAM_FLUX_URL")
```
markbackman: Remove the URL since it's GA now?

```suggestion
        api_key=os.getenv("DEEPGRAM_API_KEY")
```

src/pipecat/services/deepgram/flux/stt.py
```diff
@@ -0,0 +1,630 @@
...
+            eot_timeout_ms: Time in milliseconds after speech to finish a turn regardless of EOT confidence (default 5000).
+                A turn will be finished when this much time has passed after speech, regardless of EOT confidence.
+            keyterm: List of keyterms to boost recognition accuracy for specialized terminology.
+                Pass multiple keyterms to improve recognition of domain-specific words.
+            mip_opt_out: Opts out requests from the Deepgram Model Improvement Program (default False).
+                Refer to Deepgram docs for pricing impacts before setting this to True.
+            tag: List of tags to label requests for identification during usage reporting.
+        """
+
+        flux_encoding: str = "linear16"
```
markbackman: We don't have hard and fast rules for what goes in InputParams (we should!).

But, things like the encoding has historically been in the constructor.

markbackman: Again, regarding conventions for InputParams, we do have an established pattern of inheriting the default value of the vendor's API and using None, unless a specific default makes sense. Should we specify None for all InputParams and use Deepgram's defaults instead?

src/pipecat/services/deepgram/flux/stt.py
```diff
@@ -0,0 +1,630 @@
...
+        api_key: str,
+        url: str = "wss://api.deepgram.com/v2/listen",
+        sample_rate: Optional[int] = None,
+        model: str = "flux-general-en",
+        params: Optional[InputParams] = None,
+        **kwargs,
+    ):
+        """Initialize the Deepgram Flux STT service.
+
+        Creates a new instance of the Deepgram Flux speech-to-text service with the specified
```
markbackman: Just a nit comment, so feel free to ignore.

You can remove lines 120-122 to keep the __init__ docstring simple:
```
"""Initialize the Deepgram Flux STT service.

Args:
            api_key: Deepgram API key for authentication. Required for API access.
            url: WebSocket URL for the Deepgram Flux API. Defaults to the preview endpoint.
            sample_rate: Audio sample rate in Hz. If None, uses the rate from params or 16000.
            model: Deepgram Flux model to use for transcription. Currently only supports "flux-general-en".
            params: InputParams instance containing detailed API configuration options.
                If None, default parameters will be used.
            **kwargs: Additional arguments passed to the parent WebsocketSTTService class.
"""

The other detail, including the example, might be best in the class level docstring.

File: src/pipecat/services/deepgram/flux/stt.py [hunk 15]
```diff
@@ -0,0 +1,630 @@
...
+            stt = DeepgramFluxSTTService(
+                api_key="your-api-key",
+                model="flux-general-en",
+                params=params
+            )
+            ```
+        """
+        super().__init__(sample_rate=sample_rate, **kwargs)
+
+        if params is None:
```
markbackman: Pattern matching to other services, we use:
```
self._params = params or DeepgramFluxSTTService.InputParams()
```

src/pipecat/services/deepgram/flux/stt.py
```diff
@@ -0,0 +1,630 @@
...
+
+    async def _disconnect(self):
+        """Disconnect from WebSocket and clean up tasks.
+
+        Gracefully disconnects from the Deepgram Flux API, cancels background tasks,
+        and cleans up resources to prevent memory leaks.
+        """
+        try:
+            # First, set a flag to prevent new operations
+            self._disconnecting = True
```
markbackman: I think this variable is not used.

src/pipecat/services/deepgram/flux/stt.py
```diff
@@ -0,0 +1,630 @@
...
+            # Now close the websocket
+            await self._disconnect_websocket()
+
+        except Exception as e:
+            logger.error(f"Error during disconnect: {e}")
+        finally:
+            # Reset state only after everything is cleaned up
+            self._started = False
+            self._websocket = None
+            self._disconnecting = False
```
markbackman: Same here.

src/pipecat/services/deepgram/flux/stt.py
```diff
@@ -0,0 +1,630 @@
...
+                self._receive_task = None
+
+            # Now close the websocket
+            await self._disconnect_websocket()
+
+        except Exception as e:
+            logger.error(f"Error during disconnect: {e}")
+        finally:
+            # Reset state only after everything is cleaned up
+            self._started = False
```
markbackman: Unused?

src/pipecat/services/deepgram/flux/stt.py
```diff
@@ -0,0 +1,630 @@
...
+        Args:
+            frame: The start frame containing initialization parameters and metadata.
+        """
+        await super().start(frame)
+
+        # Build URL with all parameters from InputParams
+        url_params = [
+            f"model={self._model}",
+            f"sample_rate={self.sample_rate}",
+            f"encoding={self._params.flux_encoding}",
```
markbackman: Revisit the url_params building based on the InputParams None values.

src/pipecat/services/deepgram/flux/stt.py
```diff
@@ -0,0 +1,630 @@
...
+        Raises:
+            Exception: If the WebSocket connection is not established or if there
+                are issues sending the audio data.
+        """
+        if not self._websocket:
+            logger.error("Not connected to Deepgram Flux.")
+            yield ErrorFrame("Not connected to Deepgram Flux.", fatal=True)
+            return
+
+        await self._websocket.send(audio)
```
markbackman: Should we wrap this in a try/except in the event of Deepgram errors?

src/pipecat/services/deepgram/flux/stt.py
```diff
@@ -0,0 +1,630 @@
...
+        - Send a BotInterruptionFrame upstream to stop bot speech
+        - Send a UserStartedSpeakingFrame downstream to notify other components
+        - Start metrics collection for measuring response times
+
+        Args:
+            transcript: maybe the first few words of the turn.
+        """
+        logger.debug("User started speaking")
+        await self.push_interruption_task_frame_and_wait()
+        await self.push_frame(UserStartedSpeakingFrame(), FrameDirection.DOWNSTREAM)
```
markbackman, line 519: Should we also push upstream too? Since this is coming from the STT, that would ensure all processors experience the event.

src/pipecat/services/deepgram/flux/stt.py
```diff
@@ -0,0 +1,630 @@
...
+        """
+        logger.debug("User stopped speaking")
+
+        await self.push_frame(
+            TranscriptionFrame(
+                transcript,
+                self._user_id,
+                time_now_iso8601(),
+                self._language,
+                result=transcript,
```
markbackman: Does Flux have an equivalent "raw" result that we can provide here? That would mimic other services that make the raw data available.

src/pipecat/services/deepgram/flux/stt.py
```diff
@@ -0,0 +1,630 @@
...
+            )
+        )
+        await self._handle_transcription(transcript, True, self._language)
+        await self.stop_processing_metrics()
+        await self.push_frame(UserStoppedSpeakingFrame(), FrameDirection.DOWNSTREAM)
```
markbackman, line 568: Same comment about pushing upstream, too.

src/pipecat/services/deepgram/flux/stt.py
```diff
@@ -0,0 +1,630 @@
...
+        # begin early.
+        #
+        # However, if speech resumes or the transcripts differ from the final EndOfTurn, the
+        # EagerEndOfTurn response should be cancelled to avoid incorrect or partial responses.
+        #
+        # Pipecat doesn't yet provide built-in Gate/control mechanisms to:
+        # 1. Start LLM/TTS processing early on EagerEndOfTurn events
+        # 2. Cancel in-flight processing when TurnResumed occurs
+        #
+        # By pushing EagerEndOfTurn transcripts as InterimTranscriptionFrame, we enable
```
markbackman, line 599: I think this makes sense now until we figure out how to handle this type of early transcript estimate.

examples/foundational/07c-interruptible-deepgram-flux.py
markbackman: I was able to run this example with the default URL, so I think we can remove it.

env.example
filipi87: I guess we can. Done!

examples/foundational/07c-interruptible-deepgram-flux.py
filipi87: Done.

src/pipecat/services/deepgram/flux/stt.py
filipi87: I have refactored and moved the `flux_encoding` in the constructor, and made the `InputParams` optional.

src/pipecat/services/deepgram/flux/stt.py
filipi87: Nice catch. I’ve made it optional now, so it’s not enabled by default anymore.

src/pipecat/services/deepgram/flux/stt.py
filipi87: Done.

src/pipecat/services/deepgram/flux/stt.py
filipi87: Done.

src/pipecat/services/deepgram/flux/stt.py
filipi87: It probably became unnecessary after some refactoring. Nice catch. I’ve removed it.

src/pipecat/services/deepgram/flux/stt.py
filipi87: It probably became unnecessary after some refactoring. Nice catch. I’ve removed it.

src/pipecat/services/deepgram/flux/stt.py
filipi87: It probably became unnecessary after some refactoring. Nice catch. I’ve removed it.

src/pipecat/services/deepgram/flux/stt.py
filipi87: Yep, fixed it.

src/pipecat/services/deepgram/flux/stt.py
filipi87: Makes sense. I am not yelding an ErrorFrame in case of some Exception.

src/pipecat/services/deepgram/flux/stt.py
filipi87: I don’t think it’s generally needed, but I guess it makes sense to send. Done!

src/pipecat/services/deepgram/flux/stt.py
filipi87: Yep. Done.

src/pipecat/services/deepgram/flux/stt.py
filipi87: Ok, I am now adding the `data` instead. Which is the message that we receiving containing event type, transcript and some extra metadata.

markbackman (approved):
Looks great! Really nice addition 🙌

aconchillo (approved):

aconchillo:
LGTM!

PR #2613, opened 2025-09-08
Apply additional fixups to context messages to meet Mistral-specific …

kompfner:
…requirements

### DIFF
--- a/src/pipecat/services/mistral/llm.py
+++ b/src/pipecat/services/mistral/llm.py
@@ -57,16 +57,18 @@ class MistralLLMService(OpenAILLMService):
         logger.debug(f"Creating Mistral client with api {base_url}")
         return super().create_client(api_key, base_url, **kwargs)
 
-    def _apply_mistral_assistant_prefix(
+    def _apply_mistral_fixups(
         self, messages: List[ChatCompletionMessageParam]
     ) -> List[ChatCompletionMessageParam]:
-        """Apply Mistral's assistant message prefix requirement.
+        """Apply fixups to messages to meet Mistral-specific requirements.
 
-        Mistral requires assistant messages to have prefix=True when they
-        are the final message in a conversation. According to Mistral's API:
-        - Assistant messages with prefix=True MUST be the last message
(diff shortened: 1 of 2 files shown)


src/pipecat/services/mistral/llm.py
```diff
@@ -57,16 +57,18 @@ def create_client(self, api_key=None, base_url=None, **kwargs):
         logger.debug(f"Creating Mistral client with api {base_url}")
         return super().create_client(api_key, base_url, **kwargs)
 
-    def _apply_mistral_assistant_prefix(
+    def _apply_mistral_fixups(
```
kompfner, line 60: Note: there are enough Mistral-specific requirements on messages that Mistral might warrant its own Adapter.

But that's a slightly bigger refactor, and I wanted to get a fix in relatively quickly in the short term without changing the structure of our code.

kompfner:
Note: once this PR's changes have made it into a published version of Pipecat, we should bump the Pipecat dependency in Pipecat Flows, as Pipecat Flows doesn't really work with Mistral without them.

src/pipecat/services/mistral/llm.py
markbackman: Thanks for looking into this. So Mistral requires an assistant message immediately after a function call? Given that, an adapter might be a better way to go. I agree that this quick fix is fine for now.

src/pipecat/services/mistral/llm.py
```diff
@@ -88,6 +107,8 @@ def _apply_mistral_assistant_prefix(
         if last_message.get("role") == "assistant" and "prefix" not in last_message:
             last_message["prefix"] = True
 
+        print(f"Fixed messages for Mistral: {fixed_messages}")
```
markbackman: Remove

markbackman (approved):
LGTM

src/pipecat/services/mistral/llm.py
kompfner: whoops. good catch!

PR #2620, opened 2025-09-09
Dince/keyterms assemblyai

dan-ince-aai:
- Added keyterms support for AssemblyAI service, docs can be found here https://www.assemblyai.com/docs/speech-to-text/universal-streaming/keyterms-prompting

- Renamed formatted_finals to format_turns

- Removed word_finalization_max_wait_time

### DIFF
--- a/src/pipecat/services/assemblyai/models.py
+++ b/src/pipecat/services/assemblyai/models.py
@@ -103,7 +103,7 @@ class AssemblyAIConnectionParams(BaseModel):
     Parameters:
         sample_rate: Audio sample rate in Hz. Defaults to 16000.
         encoding: Audio encoding format. Defaults to "pcm_s16le".
-        format_turns: Whether to enable transcript formatting. Defaults to True.
+        formatted_finals: Whether to enable transcript formatting. Defaults to True.
         end_of_turn_confidence_threshold: Confidence threshold for end-of-turn detection.
         min_end_of_turn_silence_when_confident: Minimum silence duration when confident about end-of-turn.
         max_turn_silence: Maximum silence duration before forcing end-of-turn.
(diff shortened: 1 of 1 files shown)


src/pipecat/services/assemblyai/models.py
```diff
@@ -103,17 +103,17 @@ class AssemblyAIConnectionParams(BaseModel):
...
+        format_turns: Whether to enable transcript formatting. Defaults to True.
         end_of_turn_confidence_threshold: Confidence threshold for end-of-turn detection.
         min_end_of_turn_silence_when_confident: Minimum silence duration when confident about end-of-turn.
         max_turn_silence: Maximum silence duration before forcing end-of-turn.
+        keyterms_prompt: List of key terms to improve the recognition accuracy of. Defaults to None.
     """
 
     sample_rate: int = 16000
     encoding: Literal["pcm_s16le", "pcm_mulaw"] = "pcm_s16le"
-    formatted_finals: bool = True
```
markbackman, line 115: Can you deprecate these and warn about using this? Just removing them is a breaking change which we're looking to avoid.

dan-ince-aai: @markbackman I have reverted this change as we still support formatted_finals, we can look to deprecate this in a separate PR at a later date.

src/pipecat/services/assemblyai/models.py
```diff
@@ -104,16 +104,16 @@ class AssemblyAIConnectionParams(BaseModel):
         sample_rate: Audio sample rate in Hz. Defaults to 16000.
         encoding: Audio encoding format. Defaults to "pcm_s16le".
         formatted_finals: Whether to enable transcript formatting. Defaults to True.
-        word_finalization_max_wait_time: Maximum time to wait for word finalization in milliseconds.
```
markbackman, line 107: Removing this is a breaking change. Instead, let's deprecate it and we'll remove it in a future major version bump.

```
word_finalization_max_wait_time: Maximum time to wait for word finalization in milliseconds.

    .. deprecated:: 0.0.86
        This parameter is deprecated and will be removed in version 1.0.0.

```

(Note the indentation and newlines. This is needed for docs auto-gen.)

We can also warn when it's used.

Search around for `DeprecationWarning` in the code base and you'll find examples.

PR #2621, opened 2025-09-09
interruption task frame

aconchillo:

This PR introduces a few changes:

- `BotInterruptionFrame` is now `InterruptionTaskFrame`
- `StartInterruptionFrame` is now `InterruptionFrame`
- Adds a base class `TaskFrame(SystemFrame)` that all pipeline task frames inherit from
- Adds a new `FrameProcessor` method `push_interruption_task_frame_and_wait()`. This causes a pipeline interruptino and continues normal processing after it.

### DIFF
--- a/src/pipecat/pipeline/task.py
+++ b/src/pipecat/pipeline/task.py
@@ -640,9 +640,12 @@ class PipelineTask(BasePipelineTask):
             logger.debug(f"{self}: received stop task frame {frame}")
             await self.queue_frame(StopFrame())
         elif isinstance(frame, InterruptionTaskFrame):
-            # Tell the task we should interrupt the pipeline.
+            # Tell the task we should interrupt the pipeline. Note that we are
+            # bypassing the push queue and directly queue into the
+            # pipeline. This is in case the push task is blocked waiting for a
+            # pipeline-ending frame to finish traversing the pipeline.
             logger.debug(f"{self}: received interruption task frame {frame}")
-            await self.queue_frame(InterruptionFrame())
+            await self._pipeline.queue_frame(InterruptionFrame())
         elif isinstance(frame, ErrorFrame):
             if frame.fatal:
                 logger.error(f"A fatal error occurred: {frame}")
--- a/src/pipecat/processors/frame_processor.py
+++ b/src/pipecat/processors/frame_processor.py
@@ -29,6 +29,7 @@ from pipecat.frames.frames import (
     FrameProcessorResumeFrame,
     FrameProcessorResumeUrgentFrame,
     InterruptionFrame,
+    InterruptionTaskFrame,
     StartFrame,
     SystemFrame,
 )
@@ -218,6 +219,9 @@ class FrameProcessor(BaseObject):
         self.__should_block_frames = False
         self.__process_event: Optional[asyncio.Event] = None
         self.__process_frame_task: Optional[asyncio.Task] = None
+
+        self._wait_for_interruption = False
+        self._wait_interruption_event = asyncio.Event()
 
     @property
     def id(self) -> int:
@@ -542,6 +546,14 @@ class FrameProcessor(BaseObject):
         if self._cancelling:
             return
 
+        # If we are waiting for an interruption we will bypass all queued system
+        # frames and we will process the frame right away. This is because a
+        # previous system frame might be waiting for the interruption frame and
+        # it's blocking the input task.
+        if self._wait_for_interruption and isinstance(frame, InterruptionFrame):
+            await self.__process_frame(frame, direction, callback)
+            return
+
         if self._enable_direct_mode:
             await self.__process_frame(frame, direction, callback)
         else:
@@ -620,6 +632,32 @@ class FrameProcessor(BaseObject):
 
         await self.__internal_push_frame(frame, direction)
 
+        if isinstance(frame, InterruptionFrame):
+            self._wait_interruption_event.set()
+
+    async def push_interruption_task_frame_and_wait(self):
+        """Push an interruption task frame upstream and wait for the interruption.
+
+        This function sends an `InterruptionTaskFrame` upstream to the pipeline
+        task and waits to receive the corresponding `InterruptionFrame`. When
+        the function finishes it is guaranteed that the `InterruptionFrame` has
+        been pushed downstream.
+        """
+        self._wait_for_interruption = True
+
+        await self.push_frame(InterruptionTaskFrame(), FrameDirection.UPSTREAM)
+
+        # Wait for an `InterruptionFrame` to come to this processor and be
+        # pushed. Take a look at `push_frame()` to see how we first push the
+        # `InterruptionFrame` and then we set the event in order to maintain
+        # frame ordering.
+        await self._wait_interruption_event.wait()
+
+        # Clean the event.
+        self._wait_interruption_event.clear()
+
+        self._wait_for_interruption = False
+
     async def __start(self, frame: StartFrame):
         """Handle the start frame to initialize processor state.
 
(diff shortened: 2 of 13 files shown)


src/pipecat/processors/frame_processor.py
```diff
@@ -620,6 +632,41 @@ async def push_frame(self, frame: Frame, direction: FrameDirection = FrameDirect
 
         await self.__internal_push_frame(frame, direction)
 
+        if isinstance(frame, InterruptionFrame):
+            self._wait_interruption_event.set()
+
+    async def push_interruption_task_frame_and_wait(self, timeout: Optional[float] = None):
+        """Push an interruption task frame upstream and wait for the interruption.
+
+        This function sends an `InterruptionTaskFrame` upstream to the pipeline
```
kompfner, line 641: Nice clear description 👍

src/pipecat/processors/frame_processor.py
```diff
@@ -620,6 +632,41 @@ async def push_frame(self, frame: Frame, direction: FrameDirection = FrameDirect
 
         await self.__internal_push_frame(frame, direction)
 
+        if isinstance(frame, InterruptionFrame):
+            self._wait_interruption_event.set()
+
+    async def push_interruption_task_frame_and_wait(self, timeout: Optional[float] = None):
```
kompfner: When would someone need to/want to set a timeout here? We're pretty much guaranteed to receive the `InterruptionFrame`, right, and in a fairly timely manner?

Wondering if we can start simpler and then only introduce the timeout if the need arises.

src/pipecat/processors/frame_processor.py
```diff
@@ -669,19 +716,24 @@ async def __resume(self, frame: FrameProcessorResumeFrame | FrameProcessorResume
...
-        self.__create_process_task()
-
-    async def _stop_interruption(self):
-        """Stop handling an interruption."""
-        # Nothing to do right now.
-        pass
+        if not self._wait_for_interruption:
+            # Create a new process queue and task.
+            self.__create_process_task()
+        else:
```
kompfner: I feel like this `else` is conceptually more related to the `if` on line 723:

"if waiting for interruption, cancel. otherwise, reset"

Then you wouldn't need to have the slightly awkward "see comment above".

kompfner: Actually could the logic all be grouped together like this for readability?

```python
if not self._wait_for_interruption:
    # "Hard" reset, which cancels the ongoing task
    await self.__cancel_process_task()
    self.__create_process_task()
else:
    # "Soft" reset, which lets the ongoing task finish
    self.__reset_process_task()
```

kompfner: Then you could maybe even factor the "hard" reset logic into `__reset_process_task()`?

```python
await self.__reset_process_task(hard=not self._wait_for_interruption)
```

src/pipecat/frames/frames.py
```diff
@@ -1289,6 +1266,100 @@ class SpeechControlParamsFrame(SystemFrame):
...
+    """
+
+    pass
+
+
+@dataclass
+class BotInterruptionFrame(InterruptionTaskFrame):
+    """Frame indicating the bot should be interrupted.
+
+    .. deprecated:: 0.0.85
```
kompfner, line 1342: Good riddance. This name _always_ confused me.

Love the name changes in this PR!

src/pipecat/frames/frames.py
```diff
@@ -1289,6 +1266,100 @@ class SpeechControlParamsFrame(SystemFrame):
     turn_params: Optional[SmartTurnParams] = None
 
 
+#
+# Task frames
+#
+
+
+@dataclass
+class TaskFrame(SystemFrame):
```
kompfner, line 1275: 👍

src/pipecat/frames/frames.py
```diff
@@ -1289,6 +1266,100 @@ class SpeechControlParamsFrame(SystemFrame):
...
+    """Base frame for task frames.
+
+    This is a base class for frames that are meant to be sent and handled
```
kompfner, line 1278: If we _do_ eventually want to have some generic mechanisms built around `TaskFrame`s (like `await self.push_task_frame_and_wait(InterruptionTaskFrame())`), is this too loose a definition for task frames?

Would it be useful to define it more narrowly, like:

```python
This is a base class for frames that are meant to be sent upstream as requests
to the pipeline task to trigger equivalent downstream frames 
(e.g. `InterruptionTaskFrame` == "please send an `InterruptionFrame` downstream`)
```

aconchillo: Well, this might not be actually true for all task frames. As we discussed, it's possible that some task frames don't have a corresponding downstream partner.

However, I can mention that in some cases this involves sending a downstream frame like the case you mention.

src/pipecat/processors/frame_processor.py
aconchillo: I'm getting obsessed with freezing pipelines.  That's why I added it, but I think it's OK to remove. I thought: what if a custom processor blocks system frames? We would hang there... but if that's the case, the issue is on the custom processor blocking, the timeout would be a work around, so yes, let's get rid of it. It also made the code ugly :smiley:

src/pipecat/processors/frame_processor.py
aconchillo: Yes. Let me clean it up.

aconchillo: This is what I did:

```
            if self._wait_for_interruption:
                # If we get here we know the process task was just waiting for
                # an interruption (push_interruption_task_frame_and_wait()), so
                # we can't cancel the task because it might still need to do
                # more things (e.g. pushing a frame after the
                # interruption). Instead we just drain the queue because this is
                # an interruption.
                self.__reset_process_task()
            else:
                # Cancel and re-create the process task including the queue.
                await self.__cancel_process_task()
                self.__create_process_task()
```

I left the `reset_process_task()` as is without arguments.

src/pipecat/transports/base_output.py
```diff
@@ -299,7 +298,7 @@ async def process_frame(self, frame: Frame, direction: FrameDirection):
         elif isinstance(frame, CancelFrame):
             await self.cancel(frame)
             await self.push_frame(frame, direction)
-        elif isinstance(frame, StartInterruptionFrame):
```
kompfner, line 302: If we're deprecating _but not yet removing_ `StartInterruptionFrame`, should we be checking for

```python
isinstance(frame, (StartInterruptionFrame, InterruptionFrame))
```

throughout this PR?

src/pipecat/pipeline/task.py
```diff
@@ -627,13 +629,23 @@ async def _source_push_frame(self, frame: Frame, direction: FrameDirection):
...
+            logger.debug(f"{self}: received cancel task frame {frame}")
             await self.queue_frame(CancelFrame())
         elif isinstance(frame, StopTaskFrame):
             # Tell the task we should stop nicely.
+            logger.debug(f"{self}: received stop task frame {frame}")
             await self.queue_frame(StopFrame())
+        elif isinstance(frame, InterruptionTaskFrame):
+            # Tell the task we should interrupt the pipeline. Note that we are
+            # bypassing the push queue and directly queue into the
+            # pipeline. This is in case the push task is waiting for a
```
kompfner: Not sure I quite follow this reasoning...

Is it because you want the interruption to be propagated even while we're waiting for `_wait_for_pipeline_end()`?

kompfner: if so, suggested wording fix to make that clearer:

"This is in case the push task is blocked waiting for a pipeline-ending frame to finish traversing the pipeline"

src/pipecat/transports/base_output.py
aconchillo: No. I made `StartInterruptionFrame` a subclass of `InterruptionFrame` to avoid issues like this.

src/pipecat/pipeline/task.py
aconchillo: Exactly. OK, updated!

src/pipecat/transports/base_output.py
kompfner: Ah, I missed that detail! Clever solution 👍

PR #2622, opened 2025-09-10
Add to and from phone information to the development runner

markbackman:

More testing required.


Mrunmay8558:
if not from_number or not to_number:
                # Check for extra_headers in both root level and start event
                extra_headers = call_data_raw.get("extra_headers", "") or start_data.get(
                    "extra_headers", ""
                )
                if extra_headers:
                    # Parse format: "{X-PH-from: 14129162450, X-PH-to: 17242775935}"
                    import re

                    from_match = re.search(r"X-PH-from:\s([^,\s}]+)", extra_headers)
                    to_match = re.search(r"X-PH-to:\s([^,\s}]+)", extra_headers)

                    if from_match and not from_number:
                        from_number = from_match.group(1).strip()
                    if to_match and not to_number:
                        to_number = to_match.group(1).strip()

Instead of to and from variable, why can't we directly get an object or string where we can pass base64-encoded data as a dynamic variable?

markbackman:
FYI: regarding query params. I started here, but extraHeaders are the recommended way by Plivo to inject custom data into the websocket messages. This is what we want — data added to the websocket messages that can be parsed in your bot file.

Are you doing inbound or outbound calling?

For inbound calling, you have no other opportunity to ID the user other from the number. So, this is a starting point to make adding programmatic data for inbound calls possible.

For outbound calling, we could add a "custom_data" field. I'm not thrilled with the idea, but it would make sending in data when you call easier. I can do that now.

**Important: you can already do this. The development runner is just a convenience to make it easier to do things. So, if you want to do this now, you can send data via extraHeaders and parse from the start_event.**

src/pipecat/runner/utils.py
```diff
@@ -163,13 +176,47 @@ async def parse_telephony_websocket(websocket: WebSocket):
...
+
+            # Extract from/to from extra_headers if available
+            from_number = ""
+            to_number = ""
+
+            # First try customParameters (for query parameter approach)
+            if custom_params.get("from"):
+                from_number = custom_params.get("from", "")
+            if custom_params.get("to"):
+                to_number = custom_params.get("to", "")
```
filipi87: We are already providing a default value if the key doesn’t exist. So, do we need the if ?

markbackman:
@filipi87 sorry for the false start. Based on the feedback from @Mrunmay8558 I'm making one final round of updates. This will be ready for review this afternoon.

PR #2627, opened 2025-09-10
Support TwilioFrameSerializer region/edge settings. Close #2625.

### LINKED ISSUE #2625: TwilioFrameSerializer supports other region than the default us1 region
### Problem Statement

[TwilioFrameSerializer](https://github.com/pipecat-ai/pipecat/blob/main/src/pipecat/serializers/twilio.py) statically construct the endpoint url to us1 endpoint url, without consider the region/edge. If auth_token from other region is provided, 401 Error when terminating the code.

### Proposed Solution

Add optional parameter in the constructor to allow specifying region and edge
Use the region and edge prefix when construct the endpoint url if provided

### Alternative Solutions

_No response_

### Additional Context

_No response_

- [ ] No, I'm just suggesting

jessieweiyi:

#2625 

Support TwilioFrameSerializer region/edge settings so that calls are terminated successfully for regions other than default us1 region. See https://www.twilio.com/docs/global-infrastructure/using-the-twilio-rest-api-in-a-non-us-region


src/pipecat/serializers/twilio.py
```diff
@@ -172,10 +176,13 @@ async def _hang_up_call(self):
                     f"Cannot hang up Twilio call: missing required parameters: {', '.join(missing)}"
                 )
                 return
+            
+            region_prefix = f"{self._region}." if self._region else ""
+            edge_prefix = f"{self._edge}." if self._edge else ""
```
aconchillo: Is it possible to have only `edge` or only `region`? If not, we should check and raise an exception mentioning that both `edge` and `region` are required if one is set.

src/pipecat/serializers/twilio.py
```diff
@@ -173,9 +179,18 @@ async def _hang_up_call(self):
                 )
                 return
 
+            if (region and not edge) or (edge and not region):
```
markbackman: We should validate the input at initialization time and raise an error.

Maybe something like this:
```
if (region and not edge) or (edge and not region):
    raise ValueError(
        "Both edge and region parameters are required if one is set. "
        f"Twilio's FQDN format requires both: api.{{edge}}.{{region}}.twilio.com. "
        f"Got: region='{region}', edge='{edge}'"
)
```

src/pipecat/serializers/twilio.py
```diff
@@ -61,6 +61,8 @@ def __init__(
         call_sid: Optional[str] = None,
         account_sid: Optional[str] = None,
         auth_token: Optional[str] = None,
+        region: Optional[str] = None,
+        edge: Optional[str] = None,
```
markbackman, line 65: Can you update the Args docstrings for this?

```
region: Twilio region (e.g., "au1", "ie1"). Must be specified with edge.
edge: Twilio edge location (e.g., "sydney", "dublin"). Must be specified with region.
```

markbackman:
Also, please fix the formatting. You can install the pre-commit hook. From the root of the repo:
```
uv run pre-commit install
```

Or, you can run the fix-ruff.sh script:
```
./scripts/fix-ruff.sh
```

src/pipecat/serializers/twilio.py
jessieweiyi: Thanks. I was following existing code pattern. I have moved the existing code to match the same pattern as well.

nasimuddin01:
Waiting for this to get merged into production. I am having same issue.

markbackman (approved):
Nice clean up! Thanks for the contribution 🙏

PR #2634, opened 2025-09-11
feat: add bey video service

niqodea:
Add `BeyVideoService` class to generate avatars with [Beyond Presence](https://www.beyondpresence.ai/).

Setting this as draft as it is not fully ready yet but would like to receive some feedback by the maintainers. Thanks!

### DIFF
--- a/examples/foundational/45-bey-video-service.py
+++ b/examples/foundational/45-bey-video-service.py
@@ -1,0 +1,165 @@
+#
+# Copyright (c) 2024–2025, Daily
+#
+# SPDX-License-Identifier: BSD 2-Clause License
+#
+
+import os
+from typing import Any
+
+import aiohttp
+from dotenv import load_dotenv
+from loguru import logger
+
+from pipecat.audio.vad.silero import SileroVADAnalyzer
+from pipecat.frames.frames import LLMRunFrame
+from pipecat.pipeline.pipeline import Pipeline
+from pipecat.pipeline.runner import PipelineRunner
+from pipecat.pipeline.task import PipelineParams, PipelineTask
+from pipecat.processors.aggregators.openai_llm_context import OpenAILLMContext
+from pipecat.runner.types import RunnerArguments
+from pipecat.runner.utils import create_transport
+from pipecat.services.bey.video import BeyVideoService
+from pipecat.services.openai.llm import OpenAILLMService
+from pipecat.services.openai.stt import OpenAISTTService
+from pipecat.services.openai.tts import OpenAITTSService
+from pipecat.transports.base_transport import BaseTransport, TransportParams
+from pipecat.transports.daily.transport import DailyParams, DailyTransport
+from pipecat.transports.daily.utils import (
+    DailyRESTHelper,
+    DailyMeetingTokenParams,
+    DailyMeetingTokenProperties,
+)
+
+
+load_dotenv(override=True)
+
+# Ege stock avatar
+# Ref: https://docs.bey.dev/get-started/avatars/default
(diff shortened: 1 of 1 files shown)


src/pipecat/services/bey/video.py
```diff
@@ -0,0 +1,182 @@
...
+    to a Daily room from an external worker managed by Beyond Presence.
+    """
+
+    def __init__(
+        self,
+        api_key: str,
+        avatar_id: str,
+        # TODO: Is it possible to elegantly infer this from the pipeline's transport?
+        # NOTE: Unlike other providers, bey posts video directly to the room,
+        # likely resulting in lower latency
```
niqodea: Is this ok? Or should we adhere to the pattern of streaming the video back to the bot, then have the bot post to the room?

niqodea: If it's ok, what's a more elegant way to retrieve `room_url` that doesn't require the developer to specify it twice in the pipeline (once for transport, once for this component)?

src/pipecat/services/bey/video.py
```diff
@@ -0,0 +1,182 @@
...
+            token,
+            "Bey example Bot",
+            DailyParams(
+                audio_in_enabled=True,
+                video_out_enabled=False,
+                video_out_is_live=False,
+                microphone_out_enabled=False,
+                vad_analyzer=SileroVADAnalyzer(),
+            ),
+        )._client
```
niqodea: I guess this hacky retrieval of the client is also tied to the fact that we want to post directly to the room. Any suggestion here?

matrunchyk: What about SmallWebRTCTransport users?

nicodea: Not sure, I see that other providers like `tavus` define their own `TavusTransportClient`, but it seems to be a simple wrapper around a `DailyTransportClient` anyway. Is it fine to only support this type of client for now?

nicodea: Hey @matrunchyk, our code leverages the `register_audio_destination` from `DailyTransportClient` to send audio from one agent to the other without having it play in the room. Is that also supported in the `SmallWebRTCTransport`? Doesn't seem like it by taking a quick look at the code.

src/pipecat/services/bey/video.py
nicodea: Decided to inject the transport client for now in e8d8543

nicodea:
@matrunchyk refactored stuff a bit and added an example

niqodea:
@matrunchyk marked as ready for review, let me know if there's anything to fix! We would like to merge this soon to properly document it in our docs too. Thank you!

matrunchyk:
@nicodea Sorry I just saw this... There might be some confusion, I am not a member of Pipecat team or Daily team.. I'm just a humble developer 😀

@markbackman This is the person you need I believe 😁


P.S. I was emailing you using contact form a couple of weeks ago, nobody has responded yet, I wonder if the form is working?

niqodea:
Ah I see, sorry for the confusion, and thank you for pinging Mark for me 🙏

nicodea:
Hey @aconchillo, do you have any suggested next step for this? Would be great to get this merged soon :)

markbackman:
Hello! Sorry for the delayed response. We've just rolled out Community Integrations and we'd like to invite you to submit your integration for listing.

While your integration won't be part of Pipecat's core code, it will be discoverable in the [Pipecat docs](https://docs.pipecat.ai/server/services/community-integrations) and fully usable with Pipecat.

Please review the guidelines here and let me know if you have any questions:

markbackman:
Closing this out since you're our first community integration! 🙌

PR #2644, opened 2025-09-12
`run_inference` unit tests

kompfner:
And fix an issue caught by those tests!

### DIFF
--- a/src/pipecat/services/aws/llm.py
+++ b/src/pipecat/services/aws/llm.py
@@ -811,60 +811,55 @@ class AWSBedrockLLMService(LLMService):
         Returns:
             The LLM's response as a string, or None if no response is generated.
         """
-        try:
-            messages = []
-            system = []
-            if isinstance(context, LLMContext):
-                adapter: AWSBedrockLLMAdapter = self.get_llm_adapter()
-                params: AWSBedrockLLMInvocationParams = adapter.get_llm_invocation_params(context)
-                messages = params["messages"]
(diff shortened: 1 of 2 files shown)


src/pipecat/services/aws/llm.py
```diff
@@ -811,60 +811,55 @@ async def run_inference(self, context: LLMContext | OpenAILLMContext) -> Optiona
         Returns:
             The LLM's response as a string, or None if no response is generated.
         """
-        try:
```
kompfner, line 814: The only change here is to remove the `try/except` (everything else is whitespace change). Exception handling is the responsibility of the caller. This change just makes AWS match all the other LLM services in this regard.

tests/test_run_inference.py
```diff
@@ -0,0 +1,261 @@
...
+@pytest.mark.asyncio
+async def test_openai_run_inference_with_llm_context():
+    """Test run_inference with LLMContext returns expected response."""
+    # Create service with mocked client
+    with patch.object(OpenAILLMService, "create_client"):
+        service = OpenAILLMService(model="gpt-4")
+        service._client = AsyncMock()
+
+        # Setup mocks
+        mock_context = MagicMock(spec=LLMContext)
```
kompfner, line 28: Note that here I'm aspirationally not bothering to also check the `OpenAILLMContext` code path. They're similar, anyway.

markbackman (approved):
LGTM

PR #2646, opened 2025-09-12
`ServiceSwitcher` unit tests (ended up being much more than that)

kompfner:
And fix an issue caught by those tests! See [commit message](https://github.com/pipecat-ai/pipecat/pull/2646/commits/a2ad71d02958caf58cfe3ba131349cee6536e8fd) for description. TL;DR: `ServiceSwitcherFrame`s were updated to be properly processed in order.

UPDATE: actually, [_another fix_](https://github.com/pipecat-ai/pipecat/pull/2646/commits/71015bc6a79ec8fc3ac6485e1ff94f69a7da0eeb): we probably generally _don't_ want `ServiceSwitcherFrame`s to be processed in order. They should probably be `SystemFrame`s.

ANOTHER UPDATE: scratch that—we're back to processing the frames in order, as `ControlFrame`s.

### DIFF
--- a/src/pipecat/frames/frames.py
+++ b/src/pipecat/frames/frames.py
@@ -1603,46 +1603,17 @@ class MixerEnableFrame(MixerControlFrame):
 
 
 @dataclass
-class ManuallySwitchServiceBaseFrame:
-    """Base class for frames that request a manual service switch from a ServiceSwitcher.
+class ServiceSwitcherFrame(ControlFrame):
+    """A base class for frames that affect ServiceSwitcher behavior."""
 
-    Must be inherited alongside either ServiceSwitcherFrame or ServiceSwitcherControlFrame.
+    pass
+
+
+@dataclass
+class ManuallySwitchServiceFrame(ServiceSwitcherFrame):
+    """A frame to request a manual switch in the active service in a ServiceSwitcher.
+
+    Handled by ServiceSwitcherStrategyManual to switch the active service.
     """
 
     service: "FrameProcessor"
-
-
-@dataclass
-class ServiceSwitcherFrame(SystemFrame):
-    """A base class for frames that affect ServiceSwitcher behavior immediately."""
-
-    pass
-
-
-@dataclass
-class ManuallySwitchServiceFrame(ServiceSwitcherFrame, ManuallySwitchServiceBaseFrame):
-    """A frame to request an immediate manual switch in the active service in a ServiceSwitcher.
-
-    Handled by ServiceSwitcherStrategyManual to switch the active service.
-    """
-
-    pass
-
-
-@dataclass
-class ServiceSwitcherControlFrame(ControlFrame):
-    """A base class for frames that affect ServiceSwitcher behavior, in order."""
-
-    pass
-
-
-@dataclass
-class ManuallySwitchServiceControlFrame(
-    ServiceSwitcherControlFrame, ManuallySwitchServiceBaseFrame
-):
-    """A frame to request a manual switch in the active service in a ServiceSwitcher, in order.
-
-    Handled by ServiceSwitcherStrategyManual to switch the active service.
-    """
-
-    pass
--- a/src/pipecat/frames/frames.py
+++ b/src/pipecat/frames/frames.py
@@ -1603,17 +1603,46 @@ class MixerEnableFrame(MixerControlFrame):
 
 
 @dataclass
-class ServiceSwitcherFrame(ControlFrame):
-    """A base class for frames that control ServiceSwitcher behavior."""
+class ManuallySwitchServiceBaseFrame:
+    """Base class for frames that request a manual service switch from a ServiceSwitcher.
+
+    Must be inherited alongside either ServiceSwitcherFrame or ServiceSwitcherControlFrame.
+    """
+
+    service: "FrameProcessor"
+
+
+@dataclass
+class ServiceSwitcherFrame(SystemFrame):
+    """A base class for frames that affect ServiceSwitcher behavior immediately."""
 
     pass
 
 
 @dataclass
-class ManuallySwitchServiceFrame(ServiceSwitcherFrame):
-    """A frame to request a manual switch in the active service in a ServiceSwitcher.
+class ManuallySwitchServiceFrame(ServiceSwitcherFrame, ManuallySwitchServiceBaseFrame):
+    """A frame to request an immediate manual switch in the active service in a ServiceSwitcher.
 
     Handled by ServiceSwitcherStrategyManual to switch the active service.
     """
 
-    service: "FrameProcessor"
+    pass
+
+
+@dataclass
+class ServiceSwitcherControlFrame(ControlFrame):
+    """A base class for frames that affect ServiceSwitcher behavior, in order."""
+
+    pass
+
+
+@dataclass
+class ManuallySwitchServiceControlFrame(
+    ServiceSwitcherControlFrame, ManuallySwitchServiceBaseFrame
+):
+    """A frame to request a manual switch in the active service in a ServiceSwitcher, in order.
+
+    Handled by ServiceSwitcherStrategyManual to switch the active service.
+    """
+
+    pass
--- a/src/pipecat/pipeline/service_switcher.py
+++ b/src/pipecat/pipeline/service_switcher.py
@@ -102,16 +102,16 @@ class ServiceSwitcher(ParallelPipeline, Generic[StrategyType]):
             """Initialize the service switcher filter with a strategy and direction."""
 
             async def filter(_: Frame) -> bool:
-                return self.wrapped_service == self.active_service
+                return self._wrapped_service == self._active_service
 
             super().__init__(filter, direction)
-            self.wrapped_service = wrapped_service
-            self.active_service = active_service
+            self._wrapped_service = wrapped_service
+            self._active_service = active_service
 
         async def process_frame(self, frame, direction):
             """Process a frame through the filter, handling special internal filter-updating frames."""
             if isinstance(frame, ServiceSwitcher.ServiceSwitcherFilterFrame):
-                self.active_service = frame.active_service
+                self._active_service = frame.active_service
                 # Two ServiceSwitcherFilters "sandwich" a service. Push the
                 # frame only to update the other side of the sandwich, but
                 # otherwise don't let it leave the sandwich.
--- a/src/pipecat/pipeline/service_switcher.py
+++ b/src/pipecat/pipeline/service_switcher.py
@@ -112,6 +112,11 @@ class ServiceSwitcher(ParallelPipeline, Generic[StrategyType]):
             """Process a frame through the filter, handling special internal filter-updating frames."""
             if isinstance(frame, ServiceSwitcher.ServiceSwitcherFilterFrame):
                 self.active_service = frame.active_service
+                # Two ServiceSwitcherFilters "sandwich" a service. Push the
+                # frame only to update the other side of the sandwich, but
+                # otherwise don't let it leave the sandwich.
+                if direction == self._direction:
+                    await self.push_frame(frame, direction)
                 return
 
             await super().process_frame(frame, direction)
--- a/src/pipecat/pipeline/service_switcher.py
+++ b/src/pipecat/pipeline/service_switcher.py
@@ -12,10 +12,8 @@ from typing import Any, Generic, List, Optional, Type, TypeVar
 from pipecat.frames.frames import (
     ControlFrame,
     Frame,
-    ManuallySwitchServiceBaseFrame,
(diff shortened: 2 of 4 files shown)


src/pipecat/pipeline/service_switcher.py
```diff
@@ -108,6 +90,38 @@ def __init__(self, services: List[FrameProcessor], strategy_type: Type[StrategyT
...
+        async def process_frame(self, frame, direction):
+            """Process a frame through the filter, handling special internal filter-updating frames."""
+            if isinstance(frame, ServiceSwitcher.ServiceSwitcherFilterFrame):
+                self.active_service = frame.active_service
+                return
+
+            await super().process_frame(frame, direction)
+
+    @dataclass
+    class ServiceSwitcherFilterFrame(ControlFrame):
```
kompfner, line 126: I know frames are usually in `frames.py` but this is purely internal—from the `ServiceSwitcher` to the internal `ServiceSwitcherFilter`—so in this case it seemed best to just nest it inside the `ServiceSwitcher` definition (along with `ServiceSwitcherFilter`, which we also don't need users to know or care about).

kompfner:
Hmm...let's hold off on merging this. Fixing the ordering bug is actually causing the Pipecat Flows LLM switching behavior to be not quite right...we must've been inadvertently depending on the buggy behavior!

kompfner:
> Hmm...let's hold off on merging this. Fixing the ordering bug is actually causing the Pipecat Flows LLM switching behavior to be not quite right...we must've been inadvertently depending on the buggy behavior!

Ah, actually that's not it. It's just another bug here, related to upstream filtering. Fixing.

kompfner:
> > Hmm...let's hold off on merging this. Fixing the ordering bug is actually causing the Pipecat Flows LLM switching behavior to be not quite right...we must've been inadvertently depending on the buggy behavior!
> 
> Ah, actually that's not it. It's just another bug here, related to upstream filtering. Fixing.

OK, [fixed that bug](https://github.com/pipecat-ai/pipecat/pull/2646/commits/bba5262bb8b2d9942ffcb7c979d9227443878ce9).

BUT, working through this issue revealed something else: we probably want service switching (in most cases, at least) to be mediated by `SystemFrame`s, not `ControlFrame`s. Context: after putting in [the fix](https://github.com/pipecat-ai/pipecat/pull/2646/commits/a2ad71d02958caf58cfe3ba131349cee6536e8fd) to ensure that `ServiceSwitcherFrame`s were properly processed in order (and the follow-up bugfix), the LLM-switching example in Pipecat Flows still wasn't behaving properly—the _upstream-traveling_ context update triggered by result of the tool call that pushed the `ServiceSwitcherFrame` to switch LLMs would "win" in the race to the LLM service, beating out the downstream-traveling `ServiceSwitcherFrame` that would switch the LLM. That got me thinking: in most cases, won' t the user usually want the service to switch _immediately_ (i.e. be mediated by a `SystemFrame`)?

So what I then did was provide the option to either use a `SystemFrame` or a `ControlFrame` to control the service switcher.

src/pipecat/frames/frames.py
```diff
@@ -1603,17 +1603,46 @@ class MixerEnableFrame(MixerControlFrame):
...
 
     Handled by ServiceSwitcherStrategyManual to switch the active service.
     """
 
-    service: "FrameProcessor"
+    pass
+
+
+@dataclass
+class ServiceSwitcherControlFrame(ControlFrame):
```
markbackman: If we think the ControlFrame is the desired behavior, should we just start with it? Then, introduce the SystemFrame version if we get requests for it?

The ControlFrame seems more natural to use as switching services should ideally be done after the previous operation finishes.

What's the use cases for the `ManuallySwitchServiceFrame` (e.g. the SystemFrame version)?

markbackman (commented):
Tests look good. Just one question about frames.

src/pipecat/frames/frames.py
kompfner: I actually think the other one—the `SystemFrame`—is more likely the better default.

In a failover scenario, for example, you'll probably want to _immediately_ switch away from the problem service rather than letting it drain whatever queue of work it had in front of it first.

Also, in the LLM switching demo in Pipecat Flows, where we're switching LLMs in a tool call, the pure `ControlFrame` approach didn't quite work either:
- We would queue the frame to update the LLM
- We would return a result from the tool call, which would trigger an _upstream_ context frame
- The upstream context frame would usually win the race, causing the bot to respond as the _previously-selected_ LLM

The demo was only working before due to a subtle problem in the implementation, which I fixed in this PR: `ServiceSwitcherFrame`s (which were `ControlFrame`s) were already sort of "jumping the queue" because they would have an effect on both the downstream and upstream filters immediately upon landing at the `ServiceSwitcher`, which is just upstream from the services that it's switching between. Even ignoring the "action at a distance" that it would have on the upstream filter, the effect would still happen too early on the downstream. As described in [this commit message](https://github.com/pipecat-ai/pipecat/pull/2646/commits/a2ad71d02958caf58cfe3ba131349cee6536e8fd), it was failing the test of:
- Send text frame A
- Send "switch service" control frame
- Send text frame B

where we'd expect A to be handled by the old service and B by the new service. In the above scenario, both A and B were being handled by the new service.

src/pipecat/frames/frames.py
```diff
@@ -1603,17 +1603,46 @@ class MixerEnableFrame(MixerControlFrame):
...
 @dataclass
-class ServiceSwitcherFrame(ControlFrame):
-    """A base class for frames that control ServiceSwitcher behavior."""
+class ManuallySwitchServiceBaseFrame:
+    """Base class for frames that request a manual service switch from a ServiceSwitcher.
+
+    Must be inherited alongside either ServiceSwitcherFrame or ServiceSwitcherControlFrame.
+    """
+
+    service: "FrameProcessor"
```
filipi87: I’m wondering if we really need these "**_Manually_**" frames. They don’t seem to add any behavior beyond making it explicit that the frame is supposed to be sent by the user.

But as far as I can tell, we don’t make this distinction in any other frame in Pipecat.

So my guess is that your original proposal with the `ServiceSwitcherFrame` and `ServiceSwitcherControlFrame` should be sufficient here.

What do you think ? It feels like things would be simpler if we avoided this extra distinction.

src/pipecat/frames/frames.py
filipi87: I agree with Paul, in a failover scenario, I imagine you’d want to switch immediately.

What I’m wondering is, since we already have a `ServiceSwitcherControlFrame`, what if we named the other frame `ServiceSwitcherSystemFrame` ? That way, the distinction between the two would be clearer. We also wouldn’t need a default, it would be up to either the user or our internal logic to decide which one to use.

src/pipecat/frames/frames.py
filipi87: In this case, we would probably need a `ServiceSwitcherBaseFrame` to include `service: "FrameProcessor"`. 

Both classes would then need to inherit from it.

src/pipecat/pipeline/service_switcher.py
```diff
@@ -108,6 +96,55 @@ def __init__(self, services: List[FrameProcessor], strategy_type: Type[StrategyT
...
+    class ServiceSwitcherFilterFrame(SystemFrame, ServiceSwitcherFilterBaseFrame):
+        """An internal frame used by ServiceSwitcher to filter frames based on active service."""
+
+        pass
+
+    @dataclass
+    class ServiceSwitcherFilterControlFrame(ControlFrame, ServiceSwitcherFilterBaseFrame):
+        """An internal control frame used by ServiceSwitcher to filter frames based on active service."""
+
+        pass
```
filipi87: If we follow what I mentioned [here](https://github.com/pipecat-ai/pipecat/pull/2646/files#r2352078536), I guess we wouldn't not need these new frames anymore ? 

The `ServiceSwitcherFilter` could process the `ServiceSwitcherBaseFrame` directly ?

src/pipecat/frames/frames.py
kompfner: We already shipped `ServiceSwitcherFrame` in the last version, and I didn't want to introduce a breaking change.

And as a side point, I do think it's reasonable to have a sort of "default" one, if you don't want to have to think too hard.

src/pipecat/frames/frames.py
kompfner: > I’m wondering if we really need these "Manually" frames

They're for use with the `ServiceSwitcherStrategyManual` strategy specifically. We already shipped these with the last version. I don't want to introduce a breaking change.

But I also do think there's some value in having frames meant for use directly by users. For internal frames, we might want to add a bunch of additional fields they'd never care about. Think of it as an API.

src/pipecat/pipeline/service_switcher.py
kompfner: I agree it's a bit more verbose but I personally still like making the distinction between public API and internal messaging here. Per my response to the previous point, though, we already shipped the `Manually` frames, and I don't think I'd want to pass through `Manually` frames here.

src/pipecat/frames/frames.py
kompfner: > In this case, we would probably need a `ServiceSwitcherBaseFrame` to include service: "FrameProcessor".

`ServiceSwitcherBaseFrame` is meant to be much broader than just asking the `ServiceSwitcher` to pick a different service. It's for changing `ServiceSwitcher` settings, which is dependent on which strategy the `ServiceSwitcher` is using.

For example, if we add some kind of failover strategy in the future—say, a response-time based one—we might have some `UpdateServiceSwitcherFailoverSettingsFrame` that has like `metric` and `threshold` fields but no `service` field at all.

kompfner:
After discussing with @markbackman and @filipi87 in person, we decided to:
- Pare down to just the `ControlFrame`s, which are easier to reason about. We can always add `SystemFrame`s if there's customer pull for it.
- Update the Pipecat Flows example to make it work with the `ControlFrame`. This may involve pushing upstream from the assistant aggregator, to ensure that the service switch occurs before the upstream-traveling tool call result triggers another LLM inference.

src/pipecat/pipeline/service_switcher.py
```diff
@@ -108,6 +90,43 @@ def __init__(self, services: List[FrameProcessor], strategy_type: Type[StrategyT
...
+            direction: FrameDirection,
+        ):
+            """Initialize the service switcher filter with a strategy and direction."""
+
+            async def filter(_: Frame) -> bool:
+                return self.wrapped_service == self.active_service
+
+            super().__init__(filter, direction)
+            self.wrapped_service = wrapped_service
+            self.active_service = active_service
```
filipi87: nit: We should probably make these variables private: `_wrapped_service` and `_active_service`.

kompfner: Good call. Updating.

src/pipecat/pipeline/service_switcher.py
```diff
@@ -142,3 +165,9 @@ async def process_frame(self, frame: Frame, direction: FrameDirection):
 
         if isinstance(frame, ServiceSwitcherFrame):
             self.strategy.handle_frame(frame, direction)
+            service_switcher_filter_frame = ServiceSwitcher.ServiceSwitcherFilterFrame(
+                active_service=self.strategy.active_service
+            )
+            # Hack: we need access ParallelPipeline internals here to queue the frame in each of the pipelines
+            for p in self._pipelines:
+                await p.queue_frame(service_switcher_filter_frame, direction)
```
filipi87: I haven't tried it yet, but this would be the equivalent of doing this, right:
`await super().process_frame(service_switcher_filter_frame, direction)`

Would it be better ? What do you think ?

filipi87 (approved):
Looks good. 

I’ve suggested another possible solution for the hack, but I’ll leave it up to you to decide if it’s better or not. 😅

src/pipecat/pipeline/service_switcher.py
kompfner: Just tried it and the unit tests are showing it's not working (at least not as a drop-in replacement). 

It actually reminded me of why we did it this way to begin with: it's not equivalent because within `super().process_frame()` there's a call to `super().process_frame()`, and we actually **don't** want to send the internal-only `ServiceSwitcher.ServiceSwitcherFilterFrame` along the "main" pipeline, only down the branches of the parallel pipeline.

kompfner: I'll add a code comment for posterity.

kompfner: Hmm...running the tests again it's actually working. And you convinced me in Slack that your suggested code _is_ equivalent. Looking at the code again, you're totally right. I must've messed up the test before. Opening a PR with the fix.

PR #2652, opened 2025-09-14
fix: AudioBufferProcessor has_audio returns based on user or bot audi…

markbackman:
…o existing

This was causing no recording to be generated when the STTMuteFilter was enabled using `MUTE_UNTIL_FIRST_BOT_COMPLETE` strategy, the bot speaks first, and the session ends before the user speaks.

Instead, a recording should be produced with either user or bot audio.

For turn audio, it will only produce a recording if there is existing audio.


aconchillo:
LGTM! Makes sense.

PR #2653, opened 2025-09-15
`LLMContext`-adapting unit tests

kompfner:
Plus a fix for a bug discovered thanks to the tests


kompfner:
Btw, more fixes and a public convenience method coming...

kompfner:
> Btw, more fixes and a public convenience method coming...

OK, here's the new convenience method. Instead of doing

```python
LLMSpecificMessage(llm="google", message=my_message)
```

which requires the user to know what the possible values are for the `llm` field, they can now do

```python
llm.create_llm_specific_message(my_message) # where `llm` is the `LLMService`
```

PR #2660, opened 2025-09-16
fix: clean up how UserIdleProcessor handles return False

markbackman:

Surfaced in Discord:
https://discord.com/channels/1239284677165056021/1417456786377543680/1417456786377543680

The comment is correct, but didn't really have an impact because `return False` should only be called when ending the Pipeline and cleanup() was ensuring tasks were cancelled correctly. Nonetheless, this change aligns with how things are done elsewhere.

### DIFF
--- a/src/pipecat/processors/user_idle_processor.py
+++ b/src/pipecat/processors/user_idle_processor.py
@@ -17,7 +17,6 @@ from pipecat.frames.frames import (
     Frame,
     FunctionCallInProgressFrame,
     FunctionCallResultFrame,
-    StartFrame,
     UserStartedSpeakingFrame,
     UserStoppedSpeakingFrame,
 )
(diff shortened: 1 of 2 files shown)


src/pipecat/processors/user_idle_processor.py
```diff
@@ -193,7 +193,6 @@ async def _idle_task_handler(self) -> None:
                     self._retry_count += 1
                     should_continue = await self._callback(self, self._retry_count)
                     if not should_continue:
-                        await self._stop()
                         break
```
aconchillo: usually nicer to do:

```
running = True   (or should_continue)
while running:
   try:
       ....
    except ...:
        ....
        running  = await self._callback(self, self._retry_count)
```

that is, "single point of exit"

aconchillo (approved):

aconchillo:
LGTM

src/pipecat/processors/user_idle_processor.py
markbackman: Good point! Updating.

PR #2663, opened 2025-09-16
Add support for websockets 15.0

### LINKED ISSUE #2658: Increase the websocket version to 15.0.1
### Problem Statement

the current implementation does not support web socket 15.0.1 there should be no issues updating the toml to use the latest web socket version as the 15.0.1 does not have any issues that will break the existing system

### Proposed Solution

updating the toml to support the web socket 15.0.1

### Alternative Solutions

_No response_

### Additional Context

_No response_

- [ ] No, I'm just suggesting

markbackman:

This change updates optional deps that install `websockets` to support the 15.x major version. The goal being to make Pipecat more flexible with other packages.

`websockets` changelog:
https://websockets.readthedocs.io/en/stable/project/changelog.html#

- No breaking changes in 15.x
- One deprecation found that I've removed (use of `ConnectionClosed.code` and `ConnectionClosed.reason` in `AWSTranscribeSTTService`)

Resolves #2658 

UPDATE:
Added a self-referencing optional dependency to pyproject.toml, called `websockets-base`. This makes pyproject DRY by eliminating many optional deps that use websockets, replacing with the new `websockets-base`.


aconchillo:
LGTM

pyproject.toml
```diff
@@ -47,67 +47,68 @@ Website = "https://pipecat.ai"
 [project.optional-dependencies]
 aic = [ "aic-sdk~=1.0.1" ]
 anthropic = [ "anthropic~=0.49.0" ]
-assemblyai = [ "websockets>=13.1,<15.0" ]
-asyncai = [ "websockets>=13.1,<15.0" ]
-aws = [ "aioboto3~=15.0.0", "websockets>=13.1,<15.0" ]
+assemblyai = [ "pipecat-ai[websockets-base]" ]
```
kompfner, line 50: ah cool, didn't know you could do this!

seems like this works when using a local editable pipecat dependency, too, with both `pip install` and `uv sync` (just tested it to convince myself). for some reason i was expecting it to not quite work hehe

markbackman: I had the same thought! I found a few blogs showing that it actually works, then tested it myself.

PR #2665, opened 2025-09-16
Remove uv.lock

markbackman:

The uv.lock only gets in the way. Instead, for a library, omitting the uv.lock seems like the right call. It lets dependencies be installed more flexibly, based on the pyproject.toml, which reflects how developers will be using Pipecat.


aconchillo:
> The uv.lock only gets in the way. Instead, for a library, omitting the uv.lock seems like the right call. It lets dependencies be installed more flexibly, based on the pyproject.toml, which reflects how developers will be using Pipecat.

It's usually a good idea to distribute the lock files. What was the problem?

markbackman:
@aconchillo here's what sent me down this path:
- the uv.lock file is not used by developers when they install from PyPI; it's only used for development
- in development, there will be cases where we want more control over the packages used, to better reflect a given configuration. Including a uv.lock file constrains that.

So, it's a simple question of what we want the development experience to be:
- including uv.lock ensures everyone uses the same dependencies has its upside—also builds are faster
- on the downside, it makes testing different configurations more difficult—you have to be explicit about the configuration

Maybe I'm overthinking it. I'm going to leave it for now as it's not hurting anything at the moment.

PR #2670, opened 2025-09-17
Fix AsyncAI TTS to generate transcription frames like Cartesia TTS

jamsea:
## Problem

AsyncAI TTS service was not generating transcription frames for bot speech, while Cartesia TTS service was. This created inconsistency in bot transcription behavior between different TTS providers.

## Solution

Changed AsyncAI TTS service inheritance from InterruptibleTTSService to WebsocketWordTTSService to enable word timestamp functionality and transcription frame generation.

### Key Changes:

1. **Inheritance Change**: Modified AsyncAITTSService to inherit from WebsocketWordTTSService instead of InterruptibleTTSService
2. **Word Timestamp Generation**: Added word timestamp tracking in _receive_messages method
3. **Text Tracking**: Added _current_text property to track current text for timestamp generation
4. **Transcription Frame Generation**: Implemented proper TTSTextFrame generation for bot speech transcription

### Technical Details:

- When audio is received, the service now calls start_word_timestamps and add_word_timestamps 
- The entire text is added as a single timestamp at time 0.0 for simplicity
- Stop markers are added to properly end the word timestamp sequence
- This generates the same transcription frames that CartesiaTTSService produces

## Testing

This change ensures that AsyncAI TTS behaves consistently with other TTS services that generate transcription frames, enabling proper bot speech transcription in voice applications.

## Impact

- AsyncAI TTS now generates transcription frames like Cartesia TTS
- Consistent bot transcription behavior across TTS providers  
- No breaking changes to existing functionality
- Maintains all existing AsyncAI TTS features

Fixes inconsistency in TTS transcription frame generation across different providers.

### DIFF
--- a/src/pipecat/services/asyncai/tts.py
+++ b/src/pipecat/services/asyncai/tts.py
@@ -292,12 +292,20 @@ class AsyncAITTSService(WebsocketWordTTSService):
                 # Start word timestamps and add the entire text as one "word"
                 # This generates transcription frames for the bot - but only once per text
                 if not self._transcription_generated and self._current_text.strip():
+                    logger.debug(
+                        f"{self}: Generating transcription for text: '{self._current_text}'"
+                    )
                     self.start_word_timestamps()
                     # Add the entire text as a single timestamp at time 0
                     await self.add_word_timestamps([(self._current_text, 0.0)])
                     # Add stop markers to end the word timestamps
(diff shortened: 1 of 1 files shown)


markbackman:
@jamsea thanks for looking into this. The fix is pretty straightforward. The base TTSTextFrame has a `push_text_frame` arg that was set to False for some reason. I'm fixing it here:
https://github.com/pipecat-ai/pipecat/pull/2671

PR #2672, opened 2025-09-17
Monitoring the peer connection while it is in the *connecting* state.

filipi87:
Monitoring the peer connection while it is in the *connecting* state.

### DIFF
--- a/src/pipecat/transports/smallwebrtc/connection.py
+++ b/src/pipecat/transports/smallwebrtc/connection.py
@@ -206,11 +206,16 @@ class SmallWebRTCConnection(BaseObject):
     for real-time audio/video communication.
     """
 
-    def __init__(self, ice_servers: Optional[Union[List[str], List[IceServer]]] = None):
+    def __init__(
+        self,
+        ice_servers: Optional[Union[List[str], List[IceServer]]] = None,
+        connection_timeout_secs: int = 60,
+    ):
         """Initialize the WebRTC connection.
 
         Args:
             ice_servers: List of ICE servers as URLs or IceServer objects.
+            connection_timeout_secs: Timeout in seconds for connecting to the peer.
 
         Raises:
             TypeError: If ice_servers contains mixed types or unsupported types.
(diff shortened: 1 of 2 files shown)


src/pipecat/transports/smallwebrtc/connection.py
```diff
@@ -170,11 +170,16 @@ class SmallWebRTCConnection(BaseObject):
     for real-time audio/video communication.
     """
 
-    def __init__(self, ice_servers: Optional[Union[List[str], List[IceServer]]] = None):
+    def __init__(
+        self,
+        ice_servers: Optional[Union[List[str], List[IceServer]]] = None,
+        connecting_timeout_seconds: int = 60,
```
aconchillo: We use `_secs` in other places. Also, would it make more sense to call this `connection_timeout_secs`? I believe it's quite more common in general.

src/pipecat/transports/smallwebrtc/connection.py
```diff
@@ -476,9 +484,48 @@ def get_answer(self):
...
+        This method schedules a timeout task that will automatically close the
+        connection if it remains in the connecting state for more than the specified
+        timeout, default to 60 seconds.
+        """
+        logger.debug("Monitoring connecting state")
+
+        async def timeout_handler():
+            await asyncio.sleep(
+                self.connecting_timeout_seconds
+            )  # We will close the connection in case we have remained in the connecting state for over 1 minute
```
aconchillo: supernit: put the comment above so it looks better

src/pipecat/transports/smallwebrtc/connection.py
```diff
@@ -476,9 +484,48 @@ def get_answer(self):
...
+            logger.warning(
+                "Failed to establish the connection: remained in the connecting state for too long. Closing peer connection."
+            )
```
aconchillo: Maybe `Timeout establishing the connection to the remote peer. Closing.`.

aconchillo:
LGTM! Tiny comments

PR #2673, opened 2025-09-17
introduce synchronous event handlers

aconchillo:

Currently, all event handlers are executed in a task. In this PR we introduce synchronous event handlers. Synchrnous event handler are registered with:

```python
self._register_event_handler("on_before_disconnected", sync=True)
```

This means the event `on_before_disconnected` will be executed sequential (no task) and therefore it will be executed before actually disconnecting.

We also make these changes:

- DailyTransport: added synchrnous event `on_before_disconnected`.
- LiveKitTransport: added synchrnous event `on_before_disconnected`.
- Multiple event handlers for the same event are executed in separate tasks.

### DIFF
--- a/src/pipecat/utils/base_object.py
+++ b/src/pipecat/utils/base_object.py
@@ -154,35 +154,40 @@ class BaseObject(ABC):
 
         event_handler = self._event_handlers[event_name]
 
-        if event_handler.is_sync:
-            # Just run the handler.
-            await self._run_handler(event_handler, *args, **kwargs)
-        else:
-            # Create the task.
-            task = asyncio.create_task(self._run_handler(event_handler, *args, **kwargs))
+        for handler in event_handler.handlers:
+            if event_handler.is_sync:
+                # Just run the handler.
+                await self._run_handler(event_handler.name, handler, *args, **kwargs)
+            else:
+                # Create the task. Note that this is a task per each function
+                # handler. Users can register to an event handler multiple
+                # times.
+                task = asyncio.create_task(
+                    self._run_handler(event_handler.name, handler, *args, **kwargs)
+                )
 
-            # Add it to our list of event tasks.
-            self._event_tasks.add((event_name, task))
+                # Add it to our list of event tasks.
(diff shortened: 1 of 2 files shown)


src/pipecat/utils/base_object.py
```diff
@@ -14,13 +14,33 @@
...
+@dataclass
+class EventHandler:
+    """Data class to store event handlers information.
+
+    This data class stores the event name, a list of handlers to run for this
+    event, and whether these handlers will be executed in a task.
+
+    Attributes:
+        name (str): The name of the event handler.
+        handlers (List[Any]): A list of functrions to be called when this event is triggered.
```
markbackman:
```suggestion
        handlers (List[Any]): A list of functions to be called when this event is triggered.
```

src/pipecat/transports/daily/transport.py
```diff
@@ -2176,6 +2186,10 @@ async def _on_left(self):
         """Handle room left events."""
         await self._call_event_handler("on_left")
 
+    async def _on_before_leave(self):
```
markbackman, line 2207: Is there a reason why you're mixing names `on_before_leave` with `on_before_disconnect`?

aconchillo: Because for Daily, it's really before leave internally (for the Daily client), but then I use `on_before_disconnect` to match LiveKit's one. We could use `on_before_leave`.

aconchillo:
> LGTM. Just one question to clarify my understanding.
> 
> In general, this concept makes sense, especially for the scenario before leaving. I guess we just need to make it clear that these synchronous events are blocking, right?

Yes, we need to document it.

src/pipecat/transports/daily/transport.py
aconchillo: That is, we keep the `DailyTransportClient` with Daily nomenclature. And for Pipecat we use `on_before_disconnect` for all transports.

aconchillo:
Merging since we need this for other PRs. WE can update the naming if it doesn't make sense before release.

PR #2676, opened 2025-09-18
Fix audio buffer flush and silence handling

golbin:
### Observed behavior

Observed that user-only audio was missing from logged WAVs and bot/user overlap clipped mid-utterance because shorter tracks were being truncated or dropped.

### Sample code

```
    audiobuffer = AudioBufferProcessor(
        buffer_size=240000,
        num_channels=2,
        sample_rate=24000,
    )

    @audiobuffer.event_handler("on_audio_data")
    async def on_audio_data(buffer, audio, sample_rate, num_channels):
        await self.on_audio_data(audio)
```

### Root cause

flush only triggered on user buffer overflow, interleave_stereo_audio trimmed longer tracks, and _reset_recording() wiped buffers/timestamps so silence gaps vanished.

### Fix

- Ensure AudioBufferProcessor flushes when either track hits the threshold while preserving turn buffers.
- Pad the shorter track with silence before dispatching events so user/bot WAV exports stay aligned even if one side is empty.
- Capture flush timestamp before awaiting handlers to keep subsequent silence insertion accurate.

### Suggested checks

run the conversation logger for user-only speech, bot-only speech, and barge-in cases to confirm WAV tracks retain the expected audio and silence.

### DIFF
--- a/tests/test_audio_buffer_processor.py
+++ b/tests/test_audio_buffer_processor.py
@@ -1,0 +1,111 @@
+#
+# Copyright (c) 2024-2025 Daily
+#
+# SPDX-License-Identifier: BSD 2-Clause License
+#
+
+import asyncio
+import struct
+import unittest
+
+from pipecat.frames.frames import InputAudioRawFrame, OutputAudioRawFrame, StartFrame
+from pipecat.processors.audio.audio_buffer_processor import AudioBufferProcessor
+
+
+class _PassthroughResampler:
+    async def resample(self, audio: bytes, in_rate: int, out_rate: int) -> bytes:  # pragma: no cover - trivial
+        return audio
+
+
+class TestAudioBufferProcessor(unittest.IsolatedAsyncioTestCase):
+    async def asyncSetUp(self):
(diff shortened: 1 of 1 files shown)


golbin:
@markbackman @aconchillo @filipi87 Apologies for the direct mention. This is a critical issue related to recording with the buffer, and I’d appreciate it if you could take a look.

markbackman:
@golbin just acknowledging this issue. You're right that there is an issue. It's isolated to `num_channels=2`. If you're blocked, you could try one channel audio for now. We'll look at this proposal before the next release.

golbin:
Thanks @markbackman! Not sure if my patches are the best, but just FYI — I applied this patch to my service and it works fine now.

golbin:
Thanks @markbackman! Not sure if my patches are the best, but just FYI — I applied this patch to my service and it works fine now.

aconchillo:
I'll take a look. Thank you!

aconchillo:
@golbin Thank you! I believe this looks good. I'm going to merge it and look at it more carefully.

PR #2679, opened 2025-09-18
GladiaSTTService: deprecate confidence arg

markbackman:

Removing the use of confidence was advised by the Gladia team. I'm marking it as deprecated and removing the use of the parameter.


src/pipecat/services/gladia/stt.py
```diff
@@ -247,11 +250,20 @@ def __init__(
                     stacklevel=2,
                 )
 
+        if confidence is not None:
```
aconchillo: `if confidence`

aconchillo:
LGTM. Tiny comment.

PR #2680, opened 2025-09-18
DailyTransport sip_call_transfer now automatically receives session_id

markbackman:

Before, you had to handle the session_id and pass it to the sip_call_transfer method. Now, the session_id is automatically provided, allowing developers to just do the following:

```python
transfer_params = {"toEndPoint": operator_number}
await transport.sip_call_transfer(transfer_params)
```

The developer can still pass in the session_id to override.

I think this is an improvement we want to make.

### DIFF
--- a/src/pipecat/transports/daily/transport.py
+++ b/src/pipecat/transports/daily/transport.py
@@ -359,6 +359,7 @@ class DailyTransportClient(EventHandler):
         self._transcription_ids = []
         self._transcription_status = None
         self._dial_out_session_id: str = ""
+        self._dial_in_session_id: str = ""
 
         self._joining = False
         self._joined = False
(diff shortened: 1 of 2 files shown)


src/pipecat/transports/daily/transport.py
```diff
@@ -1157,6 +1169,9 @@ def on_dialin_stopped(self, data: Any):
         Args:
             data: Dial-in stop data.
         """
+        # Cleanup only if our session stopped.
+        if data.get("sessionId") == self._dial_in_session_id:
```
aconchillo, line 1173: Let's use `data.get` in the dialout functions as well.

aconchillo (approved):

aconchillo:
LGTM. Just one small comment.

src/pipecat/transports/daily/transport.py
markbackman: Done.

PR #2682, opened 2025-09-18
Smallwebrtc memory leak

filipi87:
In aiortc, when you receive a `MediaStreamTrack` (audio or video), frames are produced by the track asynchronously.

If the code never consumes these frames (i.e., you don’t read them using track.recv()), they are queued internally in memory, causing a memory leak.

This PR enables the `RTCRtpReceiver` to operate dynamically, so frames will be ignored if they are not being consumed.


markbackman (approved):
In testing, this change maintains current functionality and significantly reduces memory leaking. In my testing (using 07 and 26c), memory use looks stable.

PR #2683, opened 2025-09-18
Deprecate append-to-text handling in favor of new send-text with supp…

mattieruth:
…ort for toggling skip-tts while handling the text

This is the Pipecat portion of this proposal: https://github.com/pipecat-ai/pipecat-client-web/issues/147

It logs a deprecation warning for when it receives `append-to-context` messages and introduces support for handling a new `send-text` message which now defaults to running immediately (i.e. run-llm or run_immediately) and supports a new flag for turning off the bot's audio response (i.e. setting skip-tts to true).

This couples with https://github.com/pipecat-ai/pipecat-client-web/pull/150

### DIFF
--- a/src/pipecat/processors/frameworks/rtvi.py
+++ b/src/pipecat/processors/frameworks/rtvi.py
@@ -41,6 +41,7 @@ from pipecat.frames.frames import (
     FunctionCallResultFrame,
     InputAudioRawFrame,
     InterimTranscriptionFrame,
+    LLMConfigureOutputFrame,
     LLMContextFrame,
     LLMFullResponseEndFrame,
     LLMFullResponseStartFrame,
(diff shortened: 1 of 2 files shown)


src/pipecat/processors/frameworks/rtvi.py
```diff
@@ -1315,6 +1342,9 @@ async def process_frame(self, frame: Frame, direction: FrameDirection):
         # Data frames
         elif isinstance(frame, RTVIActionFrame):
             await self._action_queue.put(frame)
+        elif isinstance(frame, LLMConfigureOutputFrame):
```
aconchillo, line 1345: This will only work if this frame is pushed from upstream (i.e. before the RTVI processor). If for some reaons someone does: `await llm.queue_frame(LLMConfigureOutputFrame(...))` it won't work.

Solutions? None that are too good. Documentation is probably the best one. Another one is for the LLM to send a frame upstream and downstream that it has updated its config. Probably not worth it.

So nothing to do, just mentioning the potential issue.

aconchillo:
LGTM. Missing a CHANGELOG entry(ies)

src/pipecat/processors/frameworks/rtvi.py
```diff
@@ -586,10 +587,35 @@ class RTVILLMFunctionCallMessage(BaseModel):
...
+    options: Optional[RTVISendTextOptions] = None
+
+
 class RTVIAppendToContextData(BaseModel):
     """Data format for appending messages to the context.
 
     Contains the role, content, and whether to run the message immediately.
+
+        .. deprecated:: 0.0.85
+        The RTVI message, append-to-context, has been deprecated. Use send-text
```
markbackman: Indent lines after the `.. deprecated:: 0.0.85` line.

src/pipecat/processors/frameworks/rtvi.py
markbackman: Do we expect RTVI to handle `LLMConfigureOutputFrame`? It's RTVI supposed to just generate `LLMConfigureOutputFrame`?

mattieruth: > Do we expect RTVI to handle LLMConfigureOutputFrame? It's RTVI supposed to just generate LLMConfigureOutputFrame?

we have to handle it so that we can store the llm's current setting at all times in order to if we need to change the state and also be sure we set it back to the state it was in before. see _handle_send_text below.  If we had access to the llm directly, we could just get the current value, but we don't.

markbackman (approved):

PR #2687, opened 2025-09-18
Improving memory cleanup

filipi87:
- Not storing anymore the last frames received to display them in the idle processor.
- Improved memory cleanup inside the SmallWebRTCTransport.
- Calling task_done when reading the audio from the queue.
- Created a new example using the video processor to make it easier to investigate memory leaks.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- Added memory cleanup improvements to reduce memory peaks.
+
 - Added `on_before_process_frame`, `on_after_process_frame`,
   `on_before_push_frame` and `on_after_push_frame`. These are synchronous events
   that get called before and after a frame is processed or pushed. Note that


src/pipecat/services/gemini_multimodal_live/gemini.py
```diff
@@ -1171,6 +1171,9 @@ async def _handle_evt_model_turn(self, evt):
         )
         await self.push_frame(frame)
 
+        # Explicit cleanup
+        del audio
```
aconchillo: I don't understand this one.

aconchillo: We are extending the buffer, which means a copy will be made.

And then we are adding the audio also to a frame. So the reference count should just be 1 (the one in the frame) after this function.

src/pipecat/services/google/llm.py
```diff
@@ -469,14 +469,17 @@ def from_standard_message(self, message):
...
                         Part(
                             inline_data=Blob(
                                 mime_type="image/jpeg",
-                                data=base64.b64decode(c["image_url"]["url"].split(",")[1]),
+                                data=raw_bytes,
                             )
                         )
                     )
+                    # Explicit cleanup
+                    del raw_bytes
```
aconchillo: Same here...

src/pipecat/services/openai/base_llm.py
```diff
@@ -281,17 +281,22 @@ async def _stream_chat_completions_specific_context(
...
                     {
                         "type": "image_url",
                         "image_url": {"url": f"data:image/jpeg;base64,{encoded_image}"},
                     },
                 ]
+                # Explicit cleanup
                 del message["data"]
                 del message["mime_type"]
+                del raw_bytes
+                del encoded_image
```
aconchillo: these variable should be decremented when they are out of scope.

aconchillo: the ones from the dictionary made sense already.

src/pipecat/transports/base_output.py
```diff
@@ -673,6 +674,7 @@ async def with_mixer(vad_stop_secs: float) -> AsyncGenerator[Frame, None]:
                             frame.audio = await self._mixer.mix(frame.audio)
                             last_frame_time = time.time()
                         yield frame
+                        self._audio_queue.task_done()
```
aconchillo, line 677: good catch!

src/pipecat/transports/smallwebrtc/transport.py
```diff
@@ -190,6 +193,10 @@ async def recv(self):
 
         frame = VideoFrame.from_ndarray(frame_data, format="rgb24")
 
+        # Cleanup local references to free memory
+        del frame_data
+        del raw_frame
```
aconchillo: again, shouldn't the reference count be decremented and solve this issue automatically?

src/pipecat/services/gemini_multimodal_live/gemini.py
filipi87: Yeah, I started by trying to reduce peak memory usage in a long-running loop, since it can sometimes take Python a bit of time to release memory. 

But I agree, Python should handle reference counting automatically, and I probably ended up adding this in more places than necessary.

I have remove this change. 👍

src/pipecat/services/google/llm.py
filipi87: Makes sense. 👍

src/pipecat/transports/base_output.py
filipi87: Yeah, this is the one change that made the most difference in preventing the memory from growing in my tests,this and a couple of the changes inside the `SmallWebRTCTransport`.

src/pipecat/transports/smallwebrtc/transport.py
filipi87: In this case, the `recv` method is usually called continuously in a loop, and the video frames are large objects, so I used `del` to explicitly remove the references immediately, rather than waiting for the function to fully exit or for the garbage collector to run. 

In my tests this appears to help prevent memory peaks.

src/pipecat/services/openai/base_llm.py
filipi87: Removed.

aconchillo (approved):

aconchillo:
LGTM

PR #2689, opened 2025-09-19
Update quickstart and foundational examples to use smart-turn v3

markbackman:

Now that smart-turn runs on CPU and the ONNX model is bundled with Pipecat, we can use it directly in our examples. This updates all examples, so we can dogfood this model with our testing 🎉 

Updates:
- Quickstart
- Foundational: Most foundational examples that use SileroVAD were updated. Ones not updated here S2S models and examples that needs to demonstrate specific things (e.g. 13* transcription series).
  - `VADParams` `stop_secs` set to 0.2
  - Default `SmartTurnParams` used

### DIFF
--- a/examples/quickstart/pyproject.toml
+++ b/examples/quickstart/pyproject.toml
@@ -4,7 +4,7 @@ version = "0.1.0"
 description = "Quickstart example for building voice AI bots with Pipecat"
 requires-python = ">=3.10"
 dependencies = [
-    "pipecat-ai[webrtc,daily,silero,deepgram,openai,cartesia,runner]>=0.0.83",
+    "pipecat-ai[webrtc,daily,silero,deepgram,openai,cartesia,local-smart-turn-v3,runner]>=0.0.85",
     "pipecatcloud>=0.2.4"
 ]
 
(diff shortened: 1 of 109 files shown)


examples/quickstart/pyproject.toml
```diff
@@ -4,7 +4,7 @@ version = "0.1.0"
 description = "Quickstart example for building voice AI bots with Pipecat"
 requires-python = ">=3.10"
 dependencies = [
-    "pipecat-ai[webrtc,daily,silero,deepgram,openai,cartesia,runner]>=0.0.83",
+    "pipecat-ai[webrtc,daily,silero,deepgram,openai,cartesia,runner]>=0.0.85",
```
marcus-daily: I think we'll also need the local-smart-turn-v3 optional feature flag here, to make sure `transformers` gets installed.

marcus-daily:
Very nice!! I've added one comment about the dependencies.

examples/quickstart/pyproject.toml
markbackman: Thanks for the reminder 🙇

markbackman: Added the dependency and updated the imports to reflect the longer loading time. (Since first runs are slow, I use extra logging around the imports to help developers, who are running the quickstart for the first time, not get frustrated with the longer loading time of models.)

filipi87 (approved):

PR #2694, opened 2025-09-19
Add remote participant updates to DailyTransport

chadbailey59:
This PR adds a `DailyUpdateRemoteParticipantsFrame`, which enables use of the [`daily.update_remote_participants` function](https://reference-python.daily.co/api_reference.html#daily.CallClient.update_remote_participants) within a sequence of frames.

### DIFF
--- a/src/pipecat/transports/daily/transport.py
+++ b/src/pipecat/transports/daily/transport.py
@@ -25,6 +25,7 @@ from pydantic import BaseModel
 from pipecat.audio.vad.vad_analyzer import VADAnalyzer, VADParams
 from pipecat.frames.frames import (
     CancelFrame,
+    ControlFrame,
     EndFrame,
     ErrorFrame,
     Frame,
@@ -41,7 +42,6 @@ from pipecat.frames.frames import (
     UserAudioRawFrame,
     UserImageRawFrame,
     UserImageRequestFrame,
-    ControlFrame,
 )
 from pipecat.processors.frame_processor import FrameDirection, FrameProcessorSetup
 from pipecat.transcriptions.language import Language
--- a/src/pipecat/transports/daily/transport.py
+++ b/src/pipecat/transports/daily/transport.py
@@ -25,6 +25,7 @@ from pydantic import BaseModel
 from pipecat.audio.vad.vad_analyzer import VADAnalyzer, VADParams
 from pipecat.frames.frames import (
     CancelFrame,
+    ControlFrame,
     EndFrame,
     ErrorFrame,
     Frame,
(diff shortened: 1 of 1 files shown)


src/pipecat/transports/daily/transport.py
```diff
@@ -104,6 +105,15 @@ class DailyInputTransportMessageUrgentFrame(InputTransportMessageUrgentFrame):
 
     participant_id: Optional[str] = None
 
+@dataclass
+class DailyUpdateRemoteParticipantsFrame(DataFrame):
```
filipi87: @chadbailey59, I believe since we are using this frame to update Daily participants settings, it would probably be better to extend `ControlFrame` instead of `DataFrame`.

src/pipecat/transports/daily/transport.py
filipi87, line 1: To fix the format issues, you can execute this command inside the Pipecat root directory:
`./scripts/fix-ruff.sh`

src/pipecat/transports/daily/transport.py
```diff
@@ -104,6 +105,15 @@ class DailyInputTransportMessageUrgentFrame(InputTransportMessageUrgentFrame):
...
+    """Frame to update remote participants in Daily calls.
+
+    Parameters:
+        remote_participants: See https://reference-python.daily.co/api_reference.html#daily.CallClient.update_remote_participants.
+    """
+
+    remote_participants: Optional[Any] = None
```
filipi87: In other parts of the code, we are mapping the remote_participants as: `remote_participants: Mapping[str, Any]`

Shouldn’t we do the same here ? I believe this would match with the docs.

filipi87 (approved):
Hi @chadbailey59, I added a couple of comments to consider, but I’ve already approved so you’re not blocked.

aconchillo:
I'm proposing the following PR instead #2697 . With that PR you would do something like:

```
@transport.event_handler("on_after_push_frame")
async def on_after_pushed_event(transport, frame):
    if isinstance(frame, DailyUpdateRemoteParticipantsFrame):
        await transport.update_remote_participants(frame.remote_participants)
```

aconchillo:
> I'm proposing the following PR instead #2697 . With that PR you would do something like:
> 
> ```
> @transport.event_handler("on_after_push_frame")
> async def on_after_pushed_event(transport, frame):
>     if isinstance(frame, DailyUpdateRemoteParticipantsFrame):
>         await transport.update_remote_participants(frame.remote_participants)
> ```

You would keep the `DailyUpdateRemoteParticipantsFrame` in your application, no need to have that in Pipecat itself.

PR #2696, opened 2025-09-19
Changing InworldTTSService default temperature to 1.1

filipi87:
Changing InworldTTSService default temperature to 1.1

### DIFF
--- a/examples/foundational/07ab-interruptible-inworld-http.py
+++ b/examples/foundational/07ab-interruptible-inworld-http.py
@@ -68,7 +68,7 @@ async def run_bot(transport: BaseTransport, runner_args: RunnerArguments):
             model="inworld-tts-1",
             streaming=streaming,  # True: real-time chunks, False: complete audio then playback
             params=InworldTTSService.InputParams(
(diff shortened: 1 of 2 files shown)


markbackman (approved):
I think this is OK. It seems like a value of `None` might be better. That is, to just let Inworld set the default for the example. But, it's unclear what their default value is. I asked in Slack. Approving as this is fine.

examples/foundational/07ab-interruptible-inworld-http.py
```diff
@@ -68,7 +68,7 @@ async def run_bot(transport: BaseTransport, runner_args: RunnerArguments):
             model="inworld-tts-1",
             streaming=streaming,  # True: real-time chunks, False: complete audio then playback
             params=InworldTTSService.InputParams(
-                temperature=0.8,
+                temperature=1.1,
```
markbackman: I think we should just omit this value so that the default value from Inworld applies.

filipi87: Done.

PR #2697, opened 2025-09-19
FrameProcessor: add before/after events for processed/pushed frames

aconchillo:


markbackman:
I like the concept of this, but won't this generate a tremendous amount of events?

What are the excepted use cases for this?

aconchillo:
> I like the concept of this, but won't this generate a tremendous amount of events?
> 

Not really. If no one has registered an event it's an O(1) operation to just check if there's a handler for that event.

If a handler is there, then the runtime will be that of the handler. But it's exactly the same as having a processor after it. Actually faster since there's no queuing or task involved.

> What are the excepted use cases for this?

This came up when discussing performing transport operations in order (e.g. after a bot has spoken something). The current solution was merged here: https://github.com/pipecat-ai/pipecat/pull/2694

I'm proposing this PR instead (i.e. the above PR should be reverted). I thought there should be a more generic way to achieve the same other than adding one frame per each transport operation we want to execute in order. I believe this PR achieves that. It's a quick way for not having to write a custom processor if you just want to perform one single operation (in order). You still need that custom frame which is ordered which you can then handle in these new events. The new events are synchronous, so they are executed as part of the same task handling the frame.

aconchillo:
> > I like the concept of this, but won't this generate a tremendous amount of events?
> 
> Not really. If no one has registered an event it's an O(1) operation to just check if there's a handler for that event.
> 
> If a handler is there, then the runtime will be that of the handler. But it's exactly the same as having a processor after it.
> 

Also, note that these events are synchronous so no task creation is involved it's just a direct call.

markbackman (approved):
Ok. That resolves my concerns. I think this is really interesting. I would suggest writing a foundational example showing how it's used. That would help with discovery of the feature plus understanding the right way to use it.

examples/foundational/45-before-and-after-events.py
```diff
@@ -0,0 +1,150 @@
...
+    async def transport_on_after_push_frame(transport, frame):
+        if isinstance(frame, CustomAfterPushFrame):
+            logger.info("*** CUSTOM FRAME RECEIVED. Bot has stopped talking, let's do something!")
+
+    @transport.event_handler("on_client_connected")
+    async def on_client_connected(transport, client):
+        logger.info(f"Client connected")
+        # Kick off the conversation.
+        messages.append({"role": "system", "content": "Please introduce yourself to the user."})
+        # CustomSynchronizationFrame is pushed in order so it will reach the
```
kompfner, line 127: I think this comment might be out of date

kompfner: (Pardon the late post-merge review)

examples/foundational/45-before-and-after-events.py
```diff
@@ -0,0 +1,150 @@
...
+            enable_metrics=True,
+            enable_usage_metrics=True,
+        ),
+        idle_timeout_secs=runner_args.pipeline_idle_timeout_secs,
+    )
+
+    @transport.output().event_handler("on_before_process_frame")
+    async def transport_on_before_process_frame(transport, frame):
+        if isinstance(frame, CustomBeforeProcessFrame):
+            logger.info("*** CUSTOM FRAME RECEIVED. Bot is about to talk, let's do something!")
```
kompfner, line 115: Wonder if these logger statements should be more specific that they're only sandwiching the bot's _first_ utterance, not any (which is how I first misread this example). Like:

* `*** CUSTOM FRAME RECEIVED. Bot is about to start introducing themself, let's do something!`
* `*** CUSTOM FRAME RECEIVED. Bot has finishing introducing themself, let's do something!`

PR #2701, opened 2025-09-21
Add a third-party integrations guide

markbackman:

### DIFF
--- a/COMMUNITY_INTEGRATIONS.md
+++ b/COMMUNITY_INTEGRATIONS.md
@@ -1,22 +1,28 @@
-# Third-Party Integrations Guide
+# Community Integrations Guide
 
-Pipecat welcomes community-maintained integrations! As our ecosystem grows, we've established a process for third-party developers to create and maintain their own service integrations while ensuring discoverability for the community.
+Pipecat welcomes community-maintained integrations! As our ecosystem grows, we've established a process for any developer to create and maintain their own service integrations while ensuring discoverability for the Pipecat community.
 
 ## Overview
 
-**What we support:** Third-party maintained integrations that live in separate repositories and are maintained by their authors.
+**What we support:** Community-maintained integrations that live in separate repositories and are maintained by their authors.
 
-**What we don't do:** The Pipecat team does not review, test, or maintain third-party integrations. We provide guidance and list approved integrations for discoverability.
+**What we don't do:** The Pipecat team does not code review, test, or maintain community integrations. We provide guidance and list approved integrations for discoverability.
 
 **Why this approach:** This allows the community to move quickly while keeping the Pipecat core team focused on maintaining the framework itself.
 
-## Requirements for Third-Party Integration Listing
+## Submitting your Integration
 
-To be listed as an official third-party integration, your repository must contain:
+To be listed as an official community integration, follow these steps:
 
-### Required Components
+### Step 1: Build Your Integration
 
-- **Source code** - Complete implementation following Pipecat patterns (See the section below on "Integration Patterns and Examples")
+Create your integration following the patterns and examples shown in the "Integration Patterns and Examples" section below.
+
+### Step 2: Set Up Your Repository
+
+Your repository must contain these components:
+
+- **Source code** - Complete implementation following Pipecat patterns
 - **Foundational example** - Single file example showing basic usage (see [Pipecat examples](https://github.com/pipecat-ai/pipecat/tree/main/examples/foundational))
 - **README.md** - Must include:
 
@@ -31,9 +37,13 @@ To be listed as an official third-party integration, your repository must contai
 - **Code documentation** - Source code with docstrings (we recommend following [Pipecat's docstring conventions](https://github.com/pipecat-ai/pipecat/blob/main/CONTRIBUTING.md#docstring-conventions))
 - **Changelog** - Maintain a changelog for version updates
 
-### Submission Information Required
+### Step 3: Join Discord
 
-When submitting your integration for listing, provide:
+Join our Discord: https://discord.gg/pipecat
+
+### Step 4: Submit for Listing
+
+Post in the `#community-integrations` channel with:
 
 - Service name
 - Service type (STT, LLM, TTS, Image, etc.)
@@ -41,16 +51,6 @@ When submitting your integration for listing, provide:
 - Demo video (approx 30-60 seconds) showing:
   - Core functionality of your integration
   - Handling of an interruption (if applicable to service type)
-
-## Submission Process
-
-1. **Create your integration** following the patterns above
-2. **Set up your repository** with all required components
-3. **Join our Discord:** https://discord.gg/pipecat
-4. **Submit for listing** in the `#third-party-integrations` channel with:
-   - Service name and type
-   - Repository link
-   - Maintainer GitHub usernames
 
 ## Integration Patterns and Examples
 
@@ -306,7 +306,7 @@ except Exception as e:
 
 ## Disclaimer
 
-Third-party integrations are community-maintained and not officially supported by the Pipecat team. Users should evaluate these integrations independently. The Pipecat team reserves the right to remove listings that become unmaintained or problematic.
+Community integrations are community-maintained and not officially supported by the Pipecat team. Users should evaluate these integrations independently. The Pipecat team reserves the right to remove listings that become unmaintained or problematic.
 
 ## Staying Up to Date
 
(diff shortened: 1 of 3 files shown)


golbin:
By the way, it would be great if you could also distribute custom services, processors, and transports via pip as plugins. That way, developers could use Pipecat more flexibly.

markbackman:
> By the way, it would be great if you could also distribute custom services, processors, and transports via pip as plugins. That way, developers could use Pipecat more flexibly.

Just to make sure I understand, you're talking about the third-party integration, right? I've included a Packaging and Distribution section, where I say: `- Consider releasing to PyPI for easier installation`. `pip` installation pulls from PyPI, so you would be able to `pip install` a custom package; that's the idea.

golbin:
Yes. For example, let’s say I’ve built my own LLM called g-llm or a service like openai-custom-service. I’d like to be able to package them so that others can simply run pip install g-llm or pip install openai-custom-service and then use them as plugins within Pipecat.

golbin:
Another benefit of this approach is that if a plugin from another service has a bug or adds new features, we don’t need to wait for Pipecat itself to be modified or updated.

COMMUNITY_INTEGRATIONS.md
```diff
@@ -0,0 +1,262 @@
...
+Pipecat supports telephony provider integration using websocket connections to exchange MediaStreams. These services use a FrameSerializer to serialize and deserialize inputs from the FastAPIWebsocketTransport.
+
+**Examples:**
+
+- [Twilio](https://github.com/pipecat-ai/pipecat/blob/main/src/pipecat/serializers/twilio.py)
+- [Telnyx](https://github.com/pipecat-ai/pipecat/blob/main/src/pipecat/serializers/telnyx.py)
+
+**Considerations:**
+
+- Include hang-up functionality using the provider's native API, ideally using `aiohttp`
```
vr000m, line 166: Consideration for dtmf, if any.

THIRD_PARTY_INTEGRATIONS.md
```diff
@@ -0,0 +1,262 @@
...
+### Submission Information Required
+
+When submitting your integration for listing, provide:
+
+- Service name
+- Service type (STT, LLM, TTS, Image, etc.)
+- Link to your repository
+- GitHub username(s) of maintainers
+
+## Submission Process
```
filipi87: Are we going to require a video showing that the integration works, or keys that we could use to test ourselves ?

COMMUNITY_INTEGRATIONS.md
```diff
@@ -0,0 +1,262 @@
...
+
+Pipecat welcomes community-maintained integrations! As our ecosystem grows, we've established a process for third-party developers to create and maintain their own service integrations while ensuring discoverability for the community.
+
+## Overview
+
+**What we support:** Third-party maintained integrations that live in separate repositories and are maintained by their authors.
+
+**What we don't do:** The Pipecat team does not review, test, or maintain third-party integrations. We provide guidance and list approved integrations for discoverability.
+
+**Why this approach:** This allows the community to move quickly while keeping the Pipecat core team focused on maintaining the framework itself.
```
filipi87, line 11: We are going to keep both the framework as well as the already well-established services used by the community. 

However, I’m not sure how to best mention this, for example, whether we should outline criteria for deciding whether a new service will be maintained by the Pipecat team or considered a third-party service.

COMMUNITY_INTEGRATIONS.md
```diff
@@ -0,0 +1,262 @@
...
+- Use [uv](https://docs.astral.sh/uv/) for packaging (encouraged)
+- Consider releasing to PyPI for easier installation
+- Follow semantic versioning principles
+- Maintain a changelog
+
+### HTTP Communication
+
+For REST-based communication, use aiohttp. Pipecat includes this as a required dependency, so using it prevents adding an additional dependency to your integration.
+
+### Error Handling
```
filipi87, line 282: Should we make any mention of pushing `ErrorFrame` ?

COMMUNITY_INTEGRATIONS.md
aconchillo: Added

THIRD_PARTY_INTEGRATIONS.md
aconchillo: I think that would be too much. I believe if they provide an example that we can run it should be enough.

COMMUNITY_INTEGRATIONS.md
aconchillo: That's a good question. Which ones make it into the official Pipecat repo and which ones don't.

aconchillo: I'm going back and forth about all this now, I think we should discuss. If we have a list of external plugins that don't work (because they are not updated for newer versions of Pipecat or they are not well maintained), this could result in a mess and very disappointing for developers. Specially because we mention "we don't review, test". I understand this is how open source works but still we want to make Pipecat great for everyone... I guess... compromises as usual :smiley:

THIRD_PARTY_INTEGRATIONS.md
```diff
@@ -0,0 +1,316 @@
...
+
+**Why this approach:** This allows the community to move quickly while keeping the Pipecat core team focused on maintaining the framework itself.
+
+## Requirements for Third-Party Integration Listing
+
+To be listed as an official third-party integration, your repository must contain:
+
+### Required Components
+
+- **Source code** - Complete implementation following Pipecat patterns
```
mattieruth, line 25:
```suggestion
- **Source code** - Complete implementation following Pipecat patterns (See the section below on "Integration Patterns and Examples")
```

COMMUNITY_INTEGRATIONS.md
```diff
@@ -0,0 +1,316 @@
...
+
+#### Non-OpenAI Compatible Services
+
+**Requires:** Full implementation
+
+**Examples:**
+
+- [AnthropicLLMService](https://github.com/pipecat-ai/pipecat/blob/main/src/pipecat/services/anthropic/llm.py)
+- [GoogleLLMService](https://github.com/pipecat-ai/pipecat/blob/main/src/pipecat/services/google/llm.py)
+
```
mattieruth, line 101: no "Key requirements"?

THIRD_PARTY_INTEGRATIONS.md
```diff
@@ -0,0 +1,316 @@
...
+
+#### TTSService
+
+**Use for:** HTTP-based services without word/timestamp alignment
+
+**Example:**
+
+- [GoogleHttpTTSService](https://github.com/pipecat-ai/pipecat/blob/main/src/pipecat/services/google/tts.py)
+
+#### TTS Considerations:
```
mattieruth: Maybe decide on a single name for this section? Above it is "key requirements", but here, it seems to have the same type of information and is called "TTS Considerations"

mattieruth: and below you go to just "considerations" or "requirements"

THIRD_PARTY_INTEGRATIONS.md
```diff
@@ -0,0 +1,316 @@
...
+
+- **Source code** - Complete implementation following Pipecat patterns
+- **Foundational example** - Single file example showing basic usage (see [Pipecat examples](https://github.com/pipecat-ai/pipecat/tree/main/examples/foundational))
+- **README.md** - Must include:
+
+  - Introduction and explanation of your integration
+  - Installation instructions
+  - Usage instructions with Pipecat Pipeline
+  - How to run your example
+  - Last Pipecat version tested (e.g., "Tested with Pipecat v0.0.50")
```
vipyne:
```suggestion
  - Pipecat version compatibility (e.g., "Tested with Pipecat v0.0.50")
```

THIRD_PARTY_INTEGRATIONS.md
```diff
@@ -0,0 +1,316 @@
...
+
+## Submission Process
+
+1. **Create your integration** following the patterns above
+2. **Set up your repository** with all required components
+3. **Join our Discord:** https://discord.gg/pipecat
+4. **Submit for listing** in the `#third-party-integrations` channel with:
+   - Service name and type
+   - Repository link
+   - Maintainer GitHub usernames
```
vipyne: why do we need/want this?  wouldn't that be in the repo link?  (at the very least, we can see the commit messages, no?)

do we plan to list the maintainers as well as the integration?

COMMUNITY_INTEGRATIONS.md
markbackman: I skipped over this because there's a lot. I'm going to add the key points.

THIRD_PARTY_INTEGRATIONS.md
markbackman: Good call!

THIRD_PARTY_INTEGRATIONS.md
markbackman: We don't. Since we've added the company attribution suggestion, that replaces this.

PR #2705, opened 2025-09-22
update UserBotLatencyLogObserver to have logging in functions that can be overidden

tzookb:
I recently started using this UserBotLatencyLogObserver, but I wanted to have a diff behavior.
so I needed to copy paste the class.

Instead of copy pasting, we can have the logging in functions, and other engineers can override if desired


tzookb:
appreciate 🙏

PR #2708, opened 2025-09-23
BaseOutputTransport: only push downstream if transport write successful

### LINKED ISSUE #2589: AudioBufferProcessor producing huge audio files with a SoundfileMixer
### pipecat version

0.0.83

### Python version

3.13.5

### Operating System

MacOS

### Issue description

When using a SoundfileMixer with an AudioBufferProcessor, the audio that comes through "on_audio_data" is like 15 minutes long, or in some cases longer. There is a bug happening with the background audio being mistakenly looped and sent to the "on_audio_data" handler.

### Reproduction steps

1. Create a basic pipeline with an AudioBufferProcessor and SoundfileMixer which loops some background noise (I tried with both ogg and wav files)
2. Write audio from on_audio_data to a wavfile

### Expected behavior

With buffer_size at 0, I'd expect the recording to end immediately when the client disconnects.

### Actual behavior

15+min of background audio is being unnecessarily sent to on_audio_data.

### Logs

```shell

```

aconchillo:

This fixes the following issue:

- You use an audio mixer for background sound.
- A transport triggers `on_client_disconnected` because a client disconnects. Therefore the client connection is already closed.
- The `BaseOutputTransport` keeps pushing mixer audio frames non-stop (because transport is disconnected so we return immediately) until the `CancelFrame` is received.

Now we return whether the transport could send or not the frame and we only push downstream if the frame was correctly sent.

Fixes #2589

### DIFF
--- a/src/pipecat/processors/frameworks/strands_agents.py
+++ b/src/pipecat/processors/frameworks/strands_agents.py
@@ -84,11 +84,11 @@ class StrandsAgentsProcessor(FrameProcessor):
             text: The user input text to process through the agent or graph.
         """
         logger.debug(f"Invoking Strands agent with: {text}")
+        ttfb_tracking = True
         try:
             await self.push_frame(LLMFullResponseStartFrame())
             await self.start_processing_metrics()
             await self.start_ttfb_metrics()
-            ttfb_tracking = True
 
             if self.graph:
                 # Graph does not stream; await full result then emit assistant text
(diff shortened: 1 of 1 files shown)


src/pipecat/transports/base_output.py
```diff
@@ -738,12 +744,18 @@ async def _audio_task_handler(self):
...
-                await self._transport.push_frame(frame)
+                # If we are not able to write to the transport we shouldn't
+                # pushb downstream.
+                push_downstream = True
 
-                # Send audio.
+                # Try to send audio to the transport.
                 if isinstance(frame, OutputAudioRawFrame):
-                    await self._transport.write_audio_frame(frame)
+                    push_downstream = await self._transport.write_audio_frame(frame)
```
filipi87: I wonder if we should also handle exceptions here. I know we didn’t do this before, and that in case of an exception the frame wouldn’t be sent downstream, but it would be nice to fail gracefully.
Maybe a future improvement.

aconchillo: Yes. Let's do it, just in case.

PR #2711, opened 2025-09-23
Improve TTFB of Deepgram TTS

adithyaxx:
### Issue
Deepgram's Python SDK produces slow TTS outputs with ~2.1s average TTFB.

### Solution
Switch to direct HTTP POST requests, reducing average TTFB to ~0.8s (more than 2x faster).

### DIFF
--- a/src/pipecat/services/deepgram/tts.py
+++ b/src/pipecat/services/deepgram/tts.py
@@ -10,8 +10,10 @@ This module provides integration with Deepgram's text-to-speech API
 for generating speech from text using various voice models.
 """
 
+import asyncio
 from typing import AsyncGenerator, Optional
 
+import aiohttp
 from loguru import logger
 
 from pipecat.frames.frames import (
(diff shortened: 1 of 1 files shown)


aconchillo:
> ### Issue
> Deepgram's Python SDK produces slow TTS outputs with ~2.1s average TTFB.
> 
> ### Solution
> Switch to direct HTTP POST requests, reducing average TTFB to ~0.8s (more than 2x faster).

Wow! I don't think we'll have time to take a look at this before the upcoming release, but very interesting. I'm wondering if we are doing something wrong in our current code.

adithyaxx:
@aconchillo I feel the current implementation is correct since it follows Deepgram's documentation. However, comparing it to direct API calls shows a difference in TTFB, suggesting the additional delays are coming from Deepgram's Python SDK itself. This finding came from comparing the implementation in [World’s Fastest Talking AI: Deepgram + Groq](https://www.youtube.com/watch?v=J2sbC8X5Pp8) with Pipecat's implementation.

markbackman:
Awesome find! I implemented a version that I think is ready for production here:
https://github.com/pipecat-ai/pipecat/pull/2943

If you want to take the changes from #2943 and make them here, that would be great. Alternatively, we could just close this PR, but I want you to be able to take credit for this one, if you'd like.

markbackman:
@adithyaxx we've taken this change but added a new DeepgramHttpTTSService class. I'm closing this PR since the new PR now have a fair amount of changes. Again, thanks so much for bringing this up!

PR #2716, opened 2025-09-23
Add ElevenLabsSTTService

### LINKED ISSUE #1307: 11labs new STT scribe integration
I saw the new scribe model by 11labs and I think it should be added to pipecat integrations
https://elevenlabs.io/docs/api-reference/speech-to-text/convert
it currently don't have the streaming feature but it should be in the few weeks

markbackman:

Resolves #1307.

Of note: in testing this service, I experienced highly variable latency for transcripts—sometimes fast, sometimes slow. Also, my API key is on an enterprise plan, but I received 429s stating the service was under heavy load. This all makes me less confident in including it, but we have community interest in using it, so I'm adding.


aconchillo:
LGTM

PR #2719, opened 2025-09-23
Update more examples to use universal `LLMContext`. Specifically, upd…

kompfner:
…ate examples we didn't update before because they weren't using `ToolsSchema` for their tool definitions, which is a requirement for using `LLMContext`.

NOTE: oops! Turns out some of these files had *already* been updated to use universal `LLMContext` even though they weren't yet using `ToolsSchema`. This commit should fix those examples.


aconchillo:
LGTM

PR #2721, opened 2025-09-23
Update persistent conversation storage examples to use universal `LLM…

kompfner:
…Context`.

Note that `LLMContext` doesn't have a `get_messages_for_persistent_storage()`; the messages are already in the "standard" format so they can be used directly for storage.


aconchillo:
LGTM!

PR #2722, opened 2025-09-23
examples: update Strands Agents with universal context and add evals

aconchillo:

### DIFF
--- a/src/pipecat/processors/frameworks/strands_agents.py
+++ b/src/pipecat/processors/frameworks/strands_agents.py
@@ -10,12 +10,12 @@ from loguru import logger
 
 from pipecat.frames.frames import (
     Frame,
+    LLMContextFrame,
     LLMFullResponseEndFrame,
     LLMFullResponseStartFrame,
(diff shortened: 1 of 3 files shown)


src/pipecat/processors/frameworks/strands_agents.py
```diff
@@ -71,9 +71,11 @@ async def process_frame(self, frame: Frame, direction: FrameDirection):
             direction: The direction of frame flow in the pipeline.
         """
         await super().process_frame(frame, direction)
-        if isinstance(frame, OpenAILLMContextFrame):
-            text = frame.context.messages[-1]["content"]
-            await self._ainvoke(str(text).strip())
+        if isinstance(frame, LLMContextFrame):
```
kompfner, line 74: I think we still need to support `OpenAILLMContextFrame` for now, though...we haven't deprecated it yet

kompfner: Soon! 🤞

PR #2723, opened 2025-09-23
Fixing the freeze issue inside OpenAILLMService.

filipi87:
Fixed an issue in BaseOpenAILLMService that could cause the pipeline to freeze under certain race conditions.


src/pipecat/services/openai/base_llm.py
```diff
@@ -138,6 +139,7 @@ def __init__(
             default_headers=default_headers,
             **kwargs,
         )
+        self.chunk_stream: Optional[CancellableStream[AsyncStream[AsyncStream]]] = None
```
aconchillo: `self._stream`

aconchillo:
This is similar to what we had before https://github.com/pipecat-ai/pipecat/blob/89f0ff17c0461de2a6dc9ca47ad7490aab4ff32b/src/pipecat/utils/asyncio/watchdog_async_iterator.py

So annoying we have to add it again :-(

src/pipecat/services/openai/base_llm.py
```diff
@@ -416,6 +422,22 @@ async def _process_context(self, context: OpenAILLMContext | LLMContext):
...
+        """Start handling an interruption by cancelling current tasks."""
+        # OpenAI's client swallows asyncio.CancelledError internally, which prevents proper
+        # task cancellation propagation. To ensure proper cancellation behavior:
+        # 1. We check if there's an active chunk stream when receiving an interruption
+        # 2. We explicitly cancel the chunk stream first
+        # 3. This allows the task to be cancelled cleanly afterwards
+        # This approach ensures we don't get stuck in case there was a chunk processing loop
+        # when cancellation is requested.
+        if self.chunk_stream:
+            logger.debug(f"{self}: Cancelling chunk stream due to interruption")
```
aconchillo, line 435: probably `trace`

src/pipecat/utils/asyncio/cancellable_stream.py
```diff
@@ -0,0 +1,87 @@
...
+is required.
+
+The module provides functionality for:
+- Wrapping any async iterator with cancellation capabilities
+- Safe termination of async streams
+- Proper cleanup and synchronization during cancellation
+"""
+
+import asyncio
+from typing import AsyncIterator, Generic, TypeVar
```
aconchillo, line 15: move this file to `utils/asyncio`

aconchillo:
> This is similar to what we had before https://github.com/pipecat-ai/pipecat/blob/89f0ff17c0461de2a6dc9ca47ad7490aab4ff32b/src/pipecat/utils/asyncio/watchdog_async_iterator.py
> 
> So annoying we have to add it again :-(

Actually no! This is different. The previous one was only resetting the watchdog it was not fixing cancellations.

src/pipecat/services/openai/base_llm.py
```diff
@@ -416,6 +422,22 @@ async def _process_context(self, context: OpenAILLMContext | LLMContext):
 
             await self.run_function_calls(function_calls)
 
+    async def _start_interruption(self):
+        """Start handling an interruption by cancelling current tasks."""
+        # OpenAI's client swallows asyncio.CancelledError internally, which prevents proper
```
aconchillo, line 427: Do we know where this happens?

src/pipecat/services/openai/base_llm.py
filipi87: I left it as a debug because it only happens in the race condition scenario, so it would be nice to see when that happens.

src/pipecat/services/openai/base_llm.py
filipi87: It happens inside the iterator: `async for chunk in self.chunk_stream`:

This invokes the class from `openai/_streaming.py`, which invokes to close HTTP connection when a `CancelledError` happens (they are capturing the BaseException). 

I haven’t been able to pinpoint exactly where the exception is being swallowed, but I know it is, because trying to capture the `CancelledError` outside the iterator never works.

That’s why I considered this approach to properly cancel the iterator.

src/pipecat/services/openai/base_llm.py
aconchillo: Ah, OK.

aconchillo:
The reason the freeze is happening when using the freeze_test_bot example is because of Sentry SDK. I just filed this issue: https://github.com/getsentry/sentry-python/issues/4853

If we disable Sentry everything works as expected.

filipi87:
Closing this PR in favor of [this one](https://github.com/pipecat-ai/pipecat/pull/2729). It is a generic implemention, preventing issues that we haven’t been able to identify yet.

This will handle the freeze issue until Sentry fixes and publish a new version of the SDK.

filipi87:
Closing this PR in favor of https://github.com/pipecat-ai/pipecat/pull/2729. It is a generic implemention, preventing issues that we haven’t been able to identify yet.

This will handle the freeze issue until Sentry fixes and publish a new version of the SDK.

PR #2724, opened 2025-09-23
Update natural conversation examples with universal context

kompfner:
Caught a couple of bugs in those examples along the way, too. See commit messages for details.


examples/foundational/22d-natural-conversation-gemini-audio.py
```diff
@@ -765,7 +765,7 @@ async def on_client_connected(transport, client):
         await task.queue_frames([LLMRunFrame()])
 
     @transport.event_handler("on_app_message")
-    async def on_app_message(transport, message):
+    async def on_app_message(transport, message, sender):
```
filipi87, line 767: When we are invoking `on_app_message` inside `SmallWebRTCTransport`, I don’t think the `sender` is being sent. Should we modify it to include the sender ?

kompfner: Oh, hmm...looks like perhaps there's been a long-lived discrepancy between "on_app_message"'s signature between the two transports. So by "fixing" it for Daily here, I've broken it for SmallWebRTC...

No matter what we do (add `sender` to SmallWebRTC or remove `sender` from Daily`) it's a breaking change, huh...

filipi87: Yep, I guess we should fix it inside `SmallWebRTC`, in order to make it work for both.

aconchillo: I think changing it for SmallWebRTC is the right thing. I don't think many people might be using it for SmallWebRTC

src/pipecat/transports/smallwebrtc/transport.py
```diff
@@ -939,7 +939,7 @@ async def _on_app_message(self, message: Any):
         """Handle incoming application messages."""
         if self._input:
             await self._input.push_app_message(message)
-        await self._call_event_handler("on_app_message", message)
+        await self._call_event_handler("on_app_message", message, "")
```
filipi87: Since we send the `participant_id` on the `DailyTransport`, maybe here we should send the `peer_id` instead.

In that case, whoever invokes this callback would need to provide it.

You could do something like:
`self._webrtc_connection.pc_id` inside `_handle_app_message`.

This would also require changing the signature of `SmallWebRTCCallbacks`, which is only used internally.

But we can do in a follow up PR to don't keep blocking this one.

filipi87 (approved):
I added just one comment, but we can address it in a follow-up PR to avoid blocking this one.

src/pipecat/transports/smallwebrtc/transport.py
kompfner: > Since we send the participant_id on the DailyTransport, maybe here we should send the peer_id instead.

To future-proof, in case SmallWebRTC someday supports multiple connections to multiple peers?

kompfner:
> I added just one comment, but we can address it in a follow-up PR to avoid blocking this one.

Happy to take a stab at it now. I also snuck in an update to another example in this PR 😛

src/pipecat/transports/smallwebrtc/transport.py
filipi87: Yeah, I’m not sure if we’re ever going to do that or not. But it is kind of the equivalent from the daily participant id.

And we kind of do the same thing when creating the `UserImageRawFrame`,we use the `pc_id` in this case.

src/pipecat/transports/smallwebrtc/transport.py
```diff
@@ -254,7 +254,7 @@ async def on_closed(connection: SmallWebRTCConnection):
 
         @self._webrtc_connection.event_handler("app-message")
         async def on_app_message(connection: SmallWebRTCConnection, message: Any):
-            await self._handle_app_message(message)
+            await self._handle_app_message(message, connection.pc_id)
```
kompfner, line 257: `connection.pc_id` looks like the string "SmallWebRTCConnection#0" because it's assigned the name of the Python connection object at init time

src/pipecat/transports/smallwebrtc/transport.py
```diff
@@ -935,11 +935,11 @@ async def send_audio(self, frame: OutputAudioRawFrame):
         if self._output:
             await self._output.queue_frame(frame, FrameDirection.DOWNSTREAM)
 
-    async def _on_app_message(self, message: Any):
+    async def _on_app_message(self, message: Any, sender: str):
         """Handle incoming application messages."""
         if self._input:
             await self._input.push_app_message(message)
```
kompfner, line 941: I did not yet go ahead and update this to also include the `sender` because I think we'd need a bigger refactor of transport message frames to accommodate that.

Today, we have a few transport message frames:
- `TransportMessageFrame`
- `TransportMessageUrgentFrame`
- `InputTransportMessageUrgentFrame`

And some "Daily-flavored" ones that subclass those in order to add a `participant_id` field:
- `DailyTransportMessageFrame`
- `DailyTransportMessageUrgentFrame`
- `DailyInputTransportMessageUrgentFrame`

Now if we're bringing the notion of of sender/participant IDs to SmallWebRTC transport, too, I think we need to rethink the `participant_id` field being Daily-only. It might makes sense to make it part of the base transport message frames, and maybe have a new name (if we think "participant" is too Daily-specific).

Anyway, didn't want to dive too deep yet in this PR.

PR #2727, opened 2025-09-24
`StrandsAgentsProcessor` should still support `OpenAILLMContextFrame`…

kompfner:
… until that frame has been deprecated


aconchillo:
LGTM! Good catch!

PR #2729, opened 2025-09-24
FrameProcessor: timeout when cancelling tasks

aconchillo:

This PR fixes an issue where the pipeline could freeze if a task cancellation never completed because a third-party library swallowed asyncio.CancelledError. We now apply a timeout to task cancellations to prevent these freezes. If the timeout is reached, the system logs warnings and leaves dangling tasks behind, which can help diagnose where cancellation is being blocked.

See for example https://github.com/getsentry/sentry-python/issues/4853

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -106,6 +106,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- Fixed an issue where the pipeline could freeze if a task cancellation never
+  completed because a third-party library swallowed asyncio.CancelledError. We
+  now apply a timeout to task cancellations to prevent these freezes. If the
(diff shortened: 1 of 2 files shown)


aconchillo:
> I like this approach for preventing issues that we haven’t been able to identify yet.
> 
> But I always thought we were trying to avoid this case here, since it’s going to create a new async.io task every time an interruption happens. Do you think it is ok ?
> 
> Leaving already my pre-approval in case you think it is fine.

Yes, I believe it's fine. We are creating a new task anyways and that's the one we use. The old task will receive an `asyncio.CancelledError` and when it doesn't handle it properly (i.e. doesn't re-raise) the task will probably stop anyways, but most likely it won't continue normal operation.

This will show up in the logs like:
```
OpenAILLMService#0::__process_frame_task_handler: timed out waiting for task to cancel
```
and potential a dangling task.

PR #2730, opened 2025-09-24
Fix `DailyCallbacks` validation error in `TavusTransportClient` (`on_…

kompfner:
…before_leave` needed to be defined)

### DIFF
--- a/src/pipecat/transports/tavus/transport.py
+++ b/src/pipecat/transports/tavus/transport.py
@@ -221,6 +221,7 @@ class TavusTransportClient:
                 ),
                 on_joined=self._on_joined,
                 on_left=self._on_left,
+                on_before_leave=self._on_before_leave,
                 on_error=partial(self._on_handle_callback, "on_error"),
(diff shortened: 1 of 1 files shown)


src/pipecat/transports/tavus/transport.py
```diff
@@ -274,6 +275,10 @@ async def _on_left(self):
         """Handle left event."""
         logger.debug("TavusTransportClient left!")
 
+    async def _on_before_leave(self):
+        """Handle before leave event."""
+        logger.debug("TavusTransportClient before leave!")
```
kompfner, line 280: This just follows the pattern in this file, of logging when we don't actually need to do anything for a `DailyCallbacks` callback

aconchillo:
I think this one might be better: https://github.com/pipecat-ai/pipecat/pull/2733

aconchillo:
Closing in favor of https://github.com/pipecat-ai/pipecat/pull/2733

PR #2732, opened 2025-09-24
Update `VoicemailDetector` to use universal `LLMContext`

kompfner:


src/pipecat/extensions/voicemail/voicemail_detector.py
```diff
@@ -614,8 +615,8 @@ def __init__(
         ]
 
         # Create the LLM context and aggregators for conversation management
-        self._context = OpenAILLMContext(self._messages)
-        self._context_aggregator = llm.create_context_aggregator(self._context)
+        self._context = LLMContext(self._messages)
```
kompfner, line 618: This change means that we can't yet use realtime LLMs for voicemail detection, but I'm pretty sure we already couldn't

aconchillo: Yes. That was already the case.

PR #2735, opened 2025-09-24
Remove remaining usage of `OpenAILLMContext` throughout the codebase …

kompfner:
…in favor of `LLMContext`, except for:

- Usage in classes that are already deprecated
- Usage related to realtime LLMs, which don't yet support `LLMContext`
- Usage in (soon-to-be-deprecated) code paths related to `OpenAILLMContext` itself and associated machinery

### DIFF
--- a/tests/integration/test_integration_unified_function_calling.py
+++ b/tests/integration/test_integration_unified_function_calling.py
@@ -12,14 +12,12 @@ from dotenv import load_dotenv
 
 from pipecat.adapters.schemas.function_schema import FunctionSchema
 from pipecat.adapters.schemas.tools_schema import ToolsSchema
+from pipecat.frames.frames import LLMContextFrame
 from pipecat.pipeline.pipeline import Pipeline
-from pipecat.processors.aggregators.openai_llm_context import (
-    OpenAILLMContext,
-    OpenAILLMContextFrame,
-)
+from pipecat.processors.aggregators.llm_context import LLMContext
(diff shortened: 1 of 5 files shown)


tests/integration/test_integration_unified_function_calling.py
```diff
@@ -60,21 +63,19 @@ async def _test_llm_function_calling(llm: LLMService):
         },
         {"role": "user", "content": " How is the weather today in San Francisco, California?"},
     ]
-    context = OpenAILLMContext(messages, standard_tools())
-    # This is done by default inside the create_context_aggregator
```
kompfner, line 64: Using universal `LLMContext` let us get rid of this hack 👍

tests/integration/test_integration_unified_function_calling.py
```diff
@@ -48,8 +46,13 @@ def standard_tools() -> ToolsSchema:
 
 
 async def _test_llm_function_calling(llm: LLMService):
-    # Create an AsyncMock for the function
-    mock_fetch_weather = AsyncMock()
+    # Create a mock weather function
```
kompfner, line 49: I changed the way we're mocking the function to avoid this unit test printing a deprecation warning.

The issue was that the `AsyncMock()`'s signature didn't match the single-`FunctionCallParams`-argument signature our code is expecting. Our code would then assume it was a deprecated old-style function.

Issue #2736, opened 2025-09-24
AttributeError in FrameProcessor.__input_frame_task_handler when enable_direct_mode=True


pranav100000:
### pipecat version

0.0.86

### Python version

3.12.9

### Operating System

Debian

### Issue description

There's a bug in FrameProcessor. `__input_frame_task_handler()` that causes an `AttributeError`
when `enable_direct_mode=True`. The method tries to access `self.__process_queue` on line 891, but this attribute is never created when direct mode is enabled.

### Reproduction steps

Steps to Reproduce:
1. Create a DailyTransport (or any transport that uses output processors in direct mode)
2. Register event handlers that trigger frame processing
3. Start the pipeline and trigger participant events
4. The error occurs when frames are processed through the transport

### Expected behavior

  When a transport is configured with `enable_direct_mode=True`, it should:
  - Process frames immediately without errors
  - Work normally for both system and non-system frames
  - Not crash or throw exceptions

### Actual behavior

  When a transport has `enable_direct_mode=True`, it:
  - Crashes with AttributeError when trying to process non-system frames
  - Pipeline becomes unusable - the entire voice/video processing stops working

### Logs

```shell
AttributeError: 'DailyOutputTransport' object has no attribute '_FrameProcessor__process_queue'. Did you mean: '_FrameProcessor__process_event'?
```

markbackman:
This issue has been resolved in the current codebase. The fix ensures that `__process_queue` is always initialized in `__init__()` regardless of the `enable_direct_mode` flag. Additionally, when `enable_direct_mode=True`, the input and process task handlers are no longer created, so the problematic code path is never reached.

Closing this out.

PR #2737, opened 2025-09-24
BugFix: fixed frame processor direct mode bug

pranav100000:

Bug fix for: https://github.com/pipecat-ai/pipecat/issues/2736

process frames directly when in direct mode


markbackman:
This is already resolved on main. See the comment here:
https://github.com/pipecat-ai/pipecat/issues/2736#issuecomment-4139375003

PR #2742, opened 2025-09-25
DailyTransport: deprecated DailyUpdateRemoteParticipantsFrame

aconchillo:

This PR deprecates `DailyUpdateRemoteParticipantsFrame`. Developers can create their own custom frame and add a custom processor after the base output transport, or use on of the new events like, for example, `on_after_push_frame` in the output transport.

The main reason is to avoid having to support one frame per function and the corresponding handling of that frame.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Updated the default model for `AnthropicLLMService` to
   `claude-sonnet-4-5-20250929`.
 
+### Deprecated
+
+- `DailyUpdateRemoteParticipantsFrame` is deprecated and will be removed in a
+  future version. Instead, create your own custom frame and handle it in the
+  `@transport.output().event_handler("on_after_push_frame")` event handler or a
+  custom processor.
+
 ## Fixed
 
 - Fixed a `PipelineTask` issue that could prevent the application to exit if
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -11,6 +11,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Include OpenAI-based LLM services cached tokens to `MetricsFrame`.
 
+### Deprecated
+
+- `DailyUpdateRemoteParticipantsFrame` is deprecated and will be removed in a
+  future version. Create your own custom frame and use a custom processor to
+  handle it or use, for example, `on_after_push_frame` event instead in the
+  output transport.
```
vipyne: wording nit
```suggestion
  future version. Create your own custom frame and handle it in the
  `@transport.output().event_handler("on_after_push_frame")` event handler 
  or a custom processor.
```

src/pipecat/transports/daily/transport.py
```diff
@@ -110,12 +110,31 @@ class DailyInputTransportMessageUrgentFrame(InputTransportMessageUrgentFrame):
...
 
+    def __post_init__(self):
+        super().__post_init__()
+        import warnings
+
+        with warnings.catch_warnings():
+            warnings.simplefilter("always")
+            warnings.warn(
+                "DailyUpdateRemoteParticipantsFrame is deprecated and will be removed in a future version."
+                "Create your own custom frame and use a custom processor to handle it or use, for example, "
```
vipyne: same wording as above nit

CHANGELOG.md
aconchillo: Thank you. Done!

src/pipecat/transports/daily/transport.py
aconchillo: Updated.

PR #2743, opened 2025-09-25
turn analyzer fixes

aconchillo:

In this PR we now run the turn analyzer in a separate thread. We force that by changing `_process_speech_segment` and therefore `_predict_endpoint` from async to sync and then implementing this:

```
        state, result = await loop.run_in_executor(
            self._executor, self._process_speech_segment, self._audio_buffer
        )
```

This way all the CPU-bound implementation don't need to worry about anything and simply perform the computation. For HTTP-based implementation having a sync function implies that they are forced to use `run_coroutine_threadsafe`, so the contract is pretty clear in both cases.

If we kept `_predict_endpoint` async it is not clear that a thread would need to be created.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Include OpenAI-based LLM services cached tokens to `MetricsFrame`.
 
+## Fixed
+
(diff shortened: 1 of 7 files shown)


filipi87 (approved):
Makes sense. 👍

marcus-daily:
Thanks for fixing this @aconchillo!

PR #2747, opened 2025-09-26
Creating the whatsapp routes inside the runner.

filipi87:
Creating the whatsapp routes inside the runner.

This will allow us to have exactly the same behavior when running a Pipecat Cloud example and a local example.

All examples that work with SmallWebRTC should also work with WhatsApp, as long as the required environment variables are provided.


aconchillo (approved):

aconchillo:
LGTM

PR #2750, opened 2025-09-26
Support new `LLMContext` pattern with `AWSNovaSonicLLMService`

kompfner:


src/pipecat/processors/aggregators/llm_context.py
```diff
@@ -107,6 +131,21 @@ def get_messages(self, llm_specific_filter: Optional[str] = None) -> List[LLMCon
             )
         return filtered_messages
 
+    def get_messages_for_persistent_storage(
```
kompfner: This is new functionality on `LLMContext`. There are corresponding unit tests in this PR.

The reason for the `system_instruction` argument is to support usage with LLMs where you might pass the system instruction as a parameter to the `LLMService` rather than via the context.

Pending: I'll update some examples to highlight this functionality.

kompfner: Hmm...as I think about this more I'm not sure we need the complexity of the `system_instruction` argument...

- If you specified your system instruction in the context in the first place, it'll still be there when you read messages for persistent storage
- If you didn't specify your system instruction in the context and instead passed it in as an LLM service parameter, you probably *don't* want it to be in the context when you read messages for persistent storage
- ...and if you really really do need to inject it at the start of the context, it's quite easy to do yourself

kompfner: Which means...we probably don't need this method at all. We can just rely on `get_messages()`.

src/pipecat/services/aws_nova_sonic/aws.py
```diff
@@ -217,6 +217,11 @@ def __init__(
             system_instruction: System-level instruction for the model.
             tools: Available tools/functions for the model to use.
             send_transcription_frames: Whether to emit transcription frames.
+
+                .. deprecated:: 0.0.87
```
kompfner: Note to self: update this if this doesn't make the 0.0.87 release

src/pipecat/services/aws_nova_sonic/aws.py
```diff
@@ -1058,23 +1168,26 @@ def create_context_aggregator(
         *,
         user_params: LLMUserAggregatorParams = LLMUserAggregatorParams(),
         assistant_params: LLMAssistantAggregatorParams = LLMAssistantAggregatorParams(),
-    ) -> AWSNovaSonicContextAggregatorPair:
+    ) -> LLMContextAggregatorPair:
         """Create context aggregator pair for managing conversation context.
 
+        NOTE: this method exists only for backward compatibility. New code
```
kompfner: I'm hoping that changing the return value type of `create_context_aggregator()` like we do here is enough to maintain backward compatibility for almost everyone. i.e. I'm assuming that almost nobody is:
- referring to the `AWSNovaSonic*` aggregator types, or
- relying on those aggregators pushing `OpenAILLMContextFrame`s specifically

I really really didn't want to have to support seamless usage of `AWSNovaSonicLLMService` with _either_ `LLMContext`/`LLMContextAggregatorPair` or `OpenAILLMContext`/`AWSNovaSonicContextAggregatorPair`—given how different the code needs to be to support one or the other, supporting both would be a big lift (far bigger than for any of the text-to-text models).

aconchillo (approved):

aconchillo:
This looks great! Sorry for the delay.

Btw, in the previous release we named the S2S files `llm.py`. I forgot about this one. I believe we should do the same here (maybe another PR).

kompfner:
> This looks great! Sorry for the delay.
> 
> Btw, in the previous release we named the S2S files `llm.py`. I forgot about this one. I believe we should do the same here (maybe another PR).

No you didn't forget about this one! I just didn't rebase yet.

PR #2754, opened 2025-09-29
gemini multimodal live on Vertex AI support

vipyne:

Amend `GeminiMultimodalLiveLLMService` and add `GeminiMultimodalLiveVertexLLMService` wrapper service to allow usage of Gemini Multimodal Live models on Vertex AI platform.

### DIFF
--- a/src/pipecat/services/gemini_multimodal_live/gemini.py
+++ b/src/pipecat/services/gemini_multimodal_live/gemini.py
@@ -662,7 +662,7 @@ class GeminiMultimodalLiveLLMService(LLMService):
             frame: The start frame.
         """
         await super().start(frame)
-        await self._connect(self._model_name)
+        await self._connect()
 
     async def stop(self, frame: EndFrame):
         """Stop the service and close connections.
@@ -798,7 +798,7 @@ class GeminiMultimodalLiveLLMService(LLMService):
         """Return model name."""
         return self._model_name
 
-    async def _connect(self, model_name_or_path: str):
+    async def _connect(self):
         """Establish WebSocket connection to Gemini Live API."""
         if self._websocket:
             # Here we assume that if we have a websocket, we are connected. We
@@ -1158,11 +1158,9 @@ class GeminiMultimodalLiveLLMService(LLMService):
         inline_data = part.inlineData
         if not inline_data:
             return
-        if inline_data.mimeType != f"audio/pcm;rate={self._sample_rate}":
-            # Vertex AI does not always return the sample rate in mimeType
-            if inline_data.mimeType != f"audio/pcm":
-                logger.warning(f"Unrecognized server_content format {inline_data.mimeType}")
-                return
+        if inline_data.mimeType not in ["audio/pcm", f"audio/pcm;rate={self._sample_rate}"]:
+            logger.warning(f"Unrecognized server_content format {inline_data.mimeType}")
+            return
 
         audio = base64.b64decode(inline_data.data)
         if not audio:
--- a/src/pipecat/services/gemini_multimodal_live/gemini.py
+++ b/src/pipecat/services/gemini_multimodal_live/gemini.py
@@ -662,7 +662,7 @@ class GeminiMultimodalLiveLLMService(LLMService):
             frame: The start frame.
         """
         await super().start(frame)
-        await self._connect()
+        await self._connect(self._model_name)
 
     async def stop(self, frame: EndFrame):
         """Stop the service and close connections.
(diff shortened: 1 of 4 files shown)


src/pipecat/services/gemini_multimodal_live/gemini.py
```diff
@@ -788,7 +788,17 @@ async def send_client_event(self, event):
...
+        """Connect to websocket using api key."""
+        logger.info(f"Connecting to wss://{self._base_url}")
+        uri = f"wss://{self._base_url}?key={self._api_key}"
+        self._websocket = await websocket_connect(uri=uri)
+
+    def get_model_name_or_path(self):
+        """Return model name."""
+        return self._model_name
+
+    async def _connect(self, model_name_or_path: str):
```
aconchillo: we are passing this, but we are not using it. we call `self.get_model_name_or_path()` though. Maybe we remove the function (since it's not used outside) and we just use the argument?

src/pipecat/services/gemini_multimodal_live/gemini.py
```diff
@@ -1151,8 +1159,10 @@ async def _handle_evt_model_turn(self, evt):
         if not inline_data:
             return
         if inline_data.mimeType != f"audio/pcm;rate={self._sample_rate}":
-            logger.warning(f"Unrecognized server_content format {inline_data.mimeType}")
-            return
+            # Vertex AI does not always return the sample rate in mimeType
+            if inline_data.mimeType != f"audio/pcm":
```
aconchillo: maybe we can do in a single `if`:

```
if inline_data.mimeType != f"audio/pcm" and inline_data.mimeType != f"audio/pcm;rate={self._sample_rate}":
    return
```

aconchillo: actually:

```
if inline_data.mimeType not in ["audio/pcm", f"audio/pcm;rate={self._sample_rate}"]:
    return
```

src/pipecat/services/gemini_multimodal_live/gemini_vertex.py
```diff
@@ -0,0 +1,222 @@
...
+        """
+
+        # https://cloud.google.com/vertex-ai/generative-ai/docs/learn/locations
+        location: str = "us-central1"
+        project_id: str
+
+    def __init__(
+        self,
+        *,
+        api_key: str = "ignore",
```
aconchillo, line 83: why do we need this?

src/pipecat/services/gemini_multimodal_live/gemini_vertex.py
```diff
@@ -0,0 +1,222 @@
...
+        model="gemini-2.0-flash-exp",  # note: no `models/` prefix
+        params: Optional[VertexInputParams] = None,
+        **kwargs,
+    ):
+        """Initialize the Gemini Multimodal Live LLM service.
+
+        See `GeminiMultimodalLiveLLMService` for more arguments that can be passed via kwargs.
+
+        Args:
+            api_key: Ignored; use credentials to authenticate.
```
aconchillo, line 95: remove?

src/pipecat/services/gemini_multimodal_live/gemini_vertex.py
```diff
@@ -0,0 +1,222 @@
...
+                credentials_path,
+                scopes=["https://www.googleapis.com/auth/cloud-platform"],
+            )
+        else:
+            try:
+                creds, project_id = default(
+                    scopes=["https://www.googleapis.com/auth/cloud-platform"]
+                )
+            except GoogleAuthError:
+                pass
```
aconchillo, line 157: should we raise an exception?

src/pipecat/services/gemini_multimodal_live/gemini.py
aconchillo: I guess it's `self.get_model_name_or_path()` because we subclass. So we don't need the argument.

src/pipecat/services/gemini_multimodal_live/gemini_vertex.py
vipyne: because base class `GeminiMultimodalLiveLLMService` requires an `api_key`.
It's not pretty, but otherwise an api_key needs to be passed in by user, and that feels wrong since it is not used.

src/pipecat/services/gemini_multimodal_live/gemini_vertex.py
vipyne: I was the following the pattern in [llm_vertex.py](https://github.com/pipecat-ai/pipecat/blob/main/src/pipecat/services/google/llm_vertex.py#L144).  But happy to add something if that's the right thing here.

PR #2756, opened 2025-09-29
SDP munging fixes.

filipi87:
- Only remove ESP32 ICE candidates if host is defined.
- Fixing the SDP munging to keep it working on Chrome.


markbackman (approved):
LGTM

PR #2757, opened 2025-09-30
Fix AWS Bedrock timeout exception handling

jamsea:
boto3/botocore uses urllib3 internally, not httpx, so httpx.TimeoutException was never being triggered. This change ensures proper timeout handling for AWS Bedrock operations.

### DIFF
--- a/src/pipecat/services/aws/llm.py
+++ b/src/pipecat/services/aws/llm.py
@@ -61,7 +61,6 @@ from pipecat.utils.tracing.service_decorators import traced_llm
 
 try:
     import aioboto3
-    import httpx
     from botocore.config import Config
     from botocore.exceptions import ReadTimeoutError
 except ModuleNotFoundError as e:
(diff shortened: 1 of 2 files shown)


src/pipecat/services/aws/llm.py
```diff
@@ -1117,7 +1117,7 @@ async def _process_context(self, context: AWSBedrockLLMContext | LLMContext):
             # also get cancelled.
             use_completion_tokens_estimate = True
             raise
-        except httpx.TimeoutException:
+        except (ReadTimeoutError, asyncio.TimeoutError):
```
jamsea, line 1119: @adithyaxx I saw you originally added this line, was `httpx.TimeoutException` intentional here? Or was this accidentally included when copying the OpenAILLMService class?

I ask because I don't see boto3 using httpx

markbackman:
This makes sense.

We need to remove the `httpx` import too. I rebased to remove the commit history and added a changelog entry.

Thanks for finding this.

markbackman (approved):
LGTM

PR #2759, opened 2025-09-30
PipelineTask: avoid cancellation if application is finished

aconchillo:

Don't call internal `_cancel()` if we are already finished. This could cause an issue in some circumstances.


markbackman (approved):
LGTM

PR #2769, opened 2025-10-02
Catch `asyncio.CancelledError` exception on `WebsocketService._receive_task_handler`

gabrielmbmb:
## Description

This PR updates the `WebsocketService._receive_task_handler` method to catch the `asyncio.CancelledError` exception and break the receive messages loop in that case. This is required as some services such as `ElevenLabsTTSService` create a task for this method and then cancel it in its `_disconnect` method. As the code did not specifically catch `asyncio.CancelledError`, it was getting into the reconnect logic, not properly shutting down.

### DIFF
--- a/src/pipecat/services/websocket_service.py
+++ b/src/pipecat/services/websocket_service.py
@@ -83,6 +83,9 @@ class WebsocketService(ABC):
             try:
                 await self._receive_messages()
                 retry_count = 0  # Reset counter on successful message receive
+            except asyncio.CancelledError:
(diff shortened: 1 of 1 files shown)


markbackman:
Thanks for the contribution and sorry for the delay. We've made the WebsocketService more resilient to errors by adding retry logic:
https://github.com/pipecat-ai/pipecat/pull/3038

For now, let's close this PR and see if the new retry logic in the Exception case better handles issues (rather than catching the error).

PR #2771, opened 2025-10-02
Updates to publish workflows

markbackman:

- Use python 3.12
- Fix test URLs and naming

### DIFF
--- a/.github/workflows/publish.yaml
+++ b/.github/workflows/publish.yaml
@@ -5,25 +5,25 @@ on:
     inputs:
       gitref:
         type: string
-        description: "what git tag to build (e.g. v0.0.74)"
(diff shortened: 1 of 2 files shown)


.github/workflows/publish.yaml
```diff
@@ -5,25 +5,25 @@ on:
...
       gitref:
         type: string
-        description: "what git tag to build (e.g. v0.0.74)"
+        description: 'what git tag to build (e.g. v0.0.74)'
         required: true
 
 jobs:
   build:
-    name: "Build and upload wheels"
+    name: 'Build and upload wheels'
```
aconchillo, line 13: Why single quotes?

markbackman: Probably some linting rules somewhere in VSCode.

PR #2773, opened 2025-10-02
Integrating with Krisp Viva audio filter.

filipi87:


src/pipecat/audio/filters/krisp_viva_filter.py
```diff
@@ -0,0 +1,185 @@
...
+from loguru import logger
+
+from pipecat.audio.filters.base_audio_filter import BaseAudioFilter
+from pipecat.frames.frames import FilterControlFrame, FilterEnableFrame
+
+try:
+    import krisp_audio
+except ModuleNotFoundError as e:
+    logger.error(f"Exception: {e}")
+    logger.error("In order to use the Krisp filter, you need to install krisp_audio.")
```
markbackman, line 24: What actual steps should we provide? It seems like they'll need to install the Krisp Python SDK. Is that the right? Maybe we can provide a slightly more precise message.

src/pipecat/audio/filters/krisp_viva_filter.py
```diff
@@ -0,0 +1,185 @@
...
+
+
+def _log_callback(log_message, log_level):
+    logger.info(f"[{log_level}] {log_message}")
+
+
+class KrispVivaFilter(BaseAudioFilter):
+    """Audio filter using Krisp VIVA SDK.
+
+    Provides real-time noise reduction for audio streams using Krisp's
```
markbackman, line 35: Maybe add the support sample rates to the class docstring?

aconchillo:
Nice!

src/pipecat/audio/filters/krisp_viva_filter.py
filipi87: Done.

filipi87:
> This looks good and works well! I see that it parallels the Krisp implementation (https://sdk-docs.krisp.ai/docs/daily-pipecat), but improves on the audio chunk handling. Very nice!

Yes, I had to change that otherwise it was not working nicelly with the SmallWebRTCTransport and with the FastAPIWebsocketTransport.

PR #2776, opened 2025-10-02
Gemini Live service uses the `genai` library rather than WebSockets directly

kompfner:


src/pipecat/services/gemini_multimodal_live/gemini.py
```diff
@@ -492,13 +515,20 @@ def __init__(
...
     ):
         """Initialize the Gemini Multimodal Live LLM service.
 
         Args:
             api_key: Google AI API key for authentication.
             base_url: API endpoint base URL. Defaults to the official Gemini Live endpoint.
+
+                .. deprecated:: 0.0.88
+                    This parameter is deprecated and no longer has any effect.
+                    Please use `http_options` to customize requests made by the
```
kompfner, line 551: See rationale in [this commit message](https://github.com/pipecat-ai/pipecat/pull/2776/commits/f8b6587773d517a3a693ecdd26c355498524d29c). Note that `http_options` is a new argument here.

src/pipecat/services/gemini_multimodal_live/gemini.py
```diff
@@ -492,13 +515,20 @@ def __init__(
...
+        http_options: Optional[HttpOptions] = None,
         **kwargs,
     ):
         """Initialize the Gemini Multimodal Live LLM service.
 
         Args:
             api_key: Google AI API key for authentication.
             base_url: API endpoint base URL. Defaults to the official Gemini Live endpoint.
+
+                .. deprecated:: 0.0.88
```
kompfner: Note to self: bump this version if this PR doesn't make it in before 0.0.88 is released.

src/pipecat/services/gemini_multimodal_live/gemini.py
kompfner, line 500: I noticed that the `genai` API expects an array of modalities (in `LiveConnectConfig`), so I was considering updating this argument to take either a single modality (as it does today) or a list of modalities.

But based on my testing, providing more than one results in the API returning an invalid payload error, so I'm keeping this argument as is for now.

src/pipecat/services/gemini_multimodal_live/gemini.py
```diff
@@ -776,199 +828,171 @@ async def process_frame(self, frame: Frame, direction: FrameDirection):
...
-                        "media_resolution": self._settings["media_resolution"].value,
-                    },
-                    "input_audio_transcription": {},
-                    "output_audio_transcription": {},
-                }
-            }
-
-            # Add context window compression if enabled
+            # Assemble basic configuration
+            config = LiveConnectConfig(
```
patrickloeber, line 863: note that since the latest gen ai SDK v1.41.0 we have a thinking config for the liveconnectconfig, which is important to support for the latest live model (it has thinking on by default). it would be great to support this setting

See
- https://ai.google.dev/gemini-api/docs/live-guide#native-audio-output-thinking
- https://github.com/googleapis/python-genai/blob/main/google/genai/types.py#L14723

src/pipecat/services/gemini_multimodal_live/gemini.py
patrickloeber: this is correct and a current limitation. we might support more in the future: https://ai.google.dev/gemini-api/docs/live-guide#response-modalities

src/pipecat/services/gemini_multimodal_live/gemini.py
patrickloeber: it would also be great to support the native audio features affective dialog and proactive audio, but note that this currently needs the api version set to v1alpha

- https://ai.google.dev/gemini-api/docs/live-guide#affective-dialog
- https://ai.google.dev/gemini-api/docs/live-guide#proactive-audio

src/pipecat/services/gemini_multimodal_live/gemini.py
```diff
@@ -431,6 +459,10 @@ class ContextWindowCompressionParams(BaseModel):
     )  # None = use default (80% of context window)
 
 
+ThinkingConfig = _ThinkingConfig
+HttpOptions = _HttpOptions
```
kompfner: Here's a proposed pattern for user-facing config objects in this file. WDYT?

We had previously been defining our own types in this file, but I believe that was only because we weren't relying on a library like `google-genai` which already defines the necessary types. Now that we're using `google-genai` we can just use theirs. For consistency with the pattern of exporting types directly from this file rather than a dependent library (an implementation detail), I thought it'd be nice to re-export them like this.

If we like this, we can update some other types in this file to follow suit in a future PR.

src/pipecat/services/gemini_multimodal_live/gemini.py
```diff
@@ -446,6 +480,22 @@ class InputParams(BaseModel):
...
+        enable_affective_dialog: Enable affective dialog, which allows Gemini
+            to adapt to expression and tone. Defaults to None.
+            Note that these settings may require specifying a model that
+            supports them, e.g. "gemini-2.5-flash-native-audio-preview-09-2025".
+            Also note that this setting may require specifying an API version that
+            supports it, e.g. HttpOptions(api_version="v1alpha").
+        proactivity: Proactivity settings, which allows Gemini to proactively
+            decide how to behave, such as whether to avoid responding to
+            content that is not relevant. Defaults to None.
+            Note that these settings may require specifying a model that
```
kompfner, line 487: @patrickloeber Is the current recommended (i.e. stable, production-ready, and "featureful" enough for most use cases) default model still "gemini-2.0-flash-live-001"? Or do you think we ought to bump it to "gemini-2.5-flash-native-audio-preview-09-2025"?

src/pipecat/services/gemini_multimodal_live/gemini.py
kompfner: Looks like this PR will probably only make it in time for 0.0.89. Updating.

src/pipecat/services/gemini_multimodal_live/events.py
```diff
@@ -6,525 +6,22 @@
 
 """Event models and utilities for Google Gemini Multimodal Live API."""
 
-import base64
```
markbackman, line 9: Beautiful! ✨ 

Nice to not have to manage these types ☺️

src/pipecat/services/gemini_multimodal_live/events.py
```diff
@@ -6,525 +6,22 @@
...
+    )
+    from google.genai.types import (
+        StartSensitivity as _StartSensitivity,
+    )
+except ModuleNotFoundError as e:
+    logger.error(f"Exception: {e}")
+    logger.error("In order to use Google AI, you need to `pip install pipecat-ai[google]`.")
+    raise Exception(f"Missing module: {e}")
+
+# These aliases are just here for backward compatibility, since we used to
```
markbackman, line 30: 👍 

Perhaps we can deprecate them?

src/pipecat/services/gemini_multimodal_live/gemini.py
markbackman: This proposal is so you can:
```
from pipeact.services.gemini_multimodal_live.gemini import _HttpOptions,  _ProactivityConfig, _ThinkingConfig
```

Instead of from the genai types?
```
from google.genai.types import HttpOptions,  ProactivityConfig, ThinkingConfig 
```

We don't have a precedent for this in Pipecat. Well, at least one that we consciously thought about. For example, Deepgram has a `LiveOptions` type that is required to configure the service. In that service code, we just rely on the SDK to do the lifting, so developers would just use the `LiveOptions` type directly from the Deepgram SDK.

I have a slight preference to use the google-genai types since they're already documented in Google's docs already and it removes a maintenance burden.

src/pipecat/services/gemini_multimodal_live/gemini.py
```diff
@@ -509,14 +569,29 @@ def __init__(
             inference_on_context_initialization: Whether to generate a response when context
                 is first set. Defaults to True.
             file_api_base_url: Base URL for the Gemini File API. Defaults to the official endpoint.
+            http_options: HTTP options for the client.
             **kwargs: Additional arguments passed to parent LLMService.
         """
+        # Check for deprecated parameter usage
+        if (
+            base_url
```
markbackman: Perhaps to make this more clear, you could set the base_url in the args to None. Then just check for base_url being set here.

src/pipecat/services/gemini_multimodal_live/gemini.py
```diff
@@ -509,14 +569,29 @@ def __init__(
...
+                    DeprecationWarning,
+                    stacklevel=2,
+                )
+
         super().__init__(base_url=base_url, **kwargs)
 
         params = params or InputParams()
 
         self._last_sent_time = 0
-        self._api_key = api_key
```
markbackman, line 519: Nit: any reason to remove this? Pipecat generally takes args and assigns them to private variables for use. The same would be true for `http_options`. Perhaps you're removing since they're used only once.

src/pipecat/services/gemini_multimodal_live/gemini.py
```diff
@@ -768,6 +848,11 @@ async def process_frame(self, frame: Frame, direction: FrameDirection):
             # ignore this frame. Use the serverContent.turnComplete API message
             await self.push_frame(frame, direction)
         elif isinstance(frame, LLMMessagesAppendFrame):
+            # NOTE: handling LLMMessagesAppendFrame here in the LLMService is
```
markbackman, line 840: 👍 
This is atypical, but I think this makes sense here since the context is not strictly required to use Gemini Live.

CHANGELOG.md
```diff
@@ -9,6 +9,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- Added some new configuration options to `GeminiMultimodalLiveLLMService`:
```
markbackman, line 12: Nit: It could be worth adding a `### Changed` entry to note that the service now uses the google-genai SDK.

src/pipecat/services/gemini_multimodal_live/gemini.py
markbackman: Ok, I'm playing around with things a bit. I think we should not do this. By re-vending the types, IDEs lose the documentation (at least from what I see in VSCode). It would be better to have developers use those types directly and get docstring support built-in.

markbackman:
@kompfner this is looking really good! I did a bunch of testing with the foundational examples and using langfuse tracing. Everything looks really good. Just a few comments and opinions, but this one is on track. Nice work 👏

src/pipecat/services/gemini_multimodal_live/gemini.py
kompfner: Bumping again, as 0.0.89 was also already released.

src/pipecat/services/gemini_multimodal_live/gemini.py
kompfner: done!

src/pipecat/services/gemini_multimodal_live/gemini.py
kompfner: > IDEs lose the documentation

Not entirely, but partially, yeah. In VS Code, auto-complete works fine but the on-hover documentation is impacted.

> We don't have a precedent for this in Pipecat.

We do this in `LLMContext` for the context message types. They're currently just aliases fo/re-exports of the message types defined in OpenAI's library but we wanted to maintain the freedom to deviate from those types as needed in the future since `LLMContext` is LLM agnostic.

Since that freedom feels less important in this instance and we already have a precedent for directly exposing underlying library types in Deepgram as you pointed out, let's go ahead and directly use the `google-genai` types. Will update.

src/pipecat/services/gemini_multimodal_live/gemini.py
kompfner: > Perhaps you're removing since they're used only once

That's right. I didn't see any need to store any state that we don't need beyond the initializer.

src/pipecat/services/gemini_multimodal_live/gemini.py
kompfner: Yup!

CHANGELOG.md
kompfner: You think that info would be interesting to users? I couldn't decide whether it fell under "interesting enough for a changelog" and "just an implementation detail". But I suppose this changelog does report lots of implementation details already; it's not strictly for stuff that affects the API...OK, I'll add it.

markbackman: Yes and no. I've been opting to over share. In this case, it changes the underlying dependency. As much as we insist that developers use the optional dependencies, I see many developers that specify packages directly. This would be a changelog entry for them.

Issue #2207, opened 2025-07-15
"Malformed function call" causes ValidationError for LLMTokenUsage


whargrove:
### pipecat version

0.0.73

### Python version

3.12.10

### Operating System

N/A

### Issue description

GoogleVertexLLMService emits an error frame: 

```
1 validation error for LLMTokenUsage
completion_tokens
Input should be a valid integer [type=int_type, input_value=None, input_type=NoneType]
```

This happens when the model tries to call an undeclared function. The CompletionUsage data does not meet the expectations of Pipecat's [LLMTokenUsage](https://github.com/pipecat-ai/pipecat/blob/9931ad2ce1632d9e279d5a402656a5a2cdf43eea/src/pipecat/metrics/metrics.py#L51).

An ErrorFrame is sent as expected, and in some cases the pipeline stalls out completely. In other cases, it is able to recover and resume. In extreme cases, I've seen the model (gemini-2.5-flash) go "off the rails" and start regurgitating system instructions when this happens. 🙀  

### Reproduction steps

Declare a function for a Pipecat Flows node, sometimes the model will invoke the wrong function and Vertex API will return a refusal error.

### Expected behavior

Pipecat should handle the ChatCompletionChunk gracefully with None completion_tokens. Emit ErrorFrame.

### Actual behavior

Pydantic ValidationError, Pipecat pipeline becomes stalled.

### Logs

```shell
{
  "timestamp": "2025-07-15T21:07:02.423576Z",
  "logger_name": "pipecat.processors.frame_processor",
  "file_path": "/app/apps/candidate-profile-agent/.venv/lib/python3.12/site-packages/pipecat/processors/frame_processor.py",
  "file_name": "frame_processor.py",
  "function": "__input_frame_task_handler",
  "line_number": 442,
  "module": "frame_processor",
  "process_id": 13,
  "process_name": "MainProcess",
  "thread_id": 281472850612256,
  "thread_name": "MainThread",
  "loguru_level": "ERROR",
  "loguru_level_no": 40,
  "event": "GoogleVertexLLMService#0: error processing frame: 1 validation error for LLMTokenUsage\ncompletion_tokens\n Input should be a valid integer [type=int_type, input_value=None, input_type=NoneType]\n For further information visit https://errors.pydantic.dev/2.10/v/int_type",
  "session_id": "f14dba41-eb5c-4b54-80c2-35cc27c342de",
  "level": "error",
  "filename": "logging.py",
[...]

PR #2779, opened 2025-10-02
Ignore None value for 'completion_tokens' or similar for Gemini

LucasStringPay:
Similar fix as https://github.com/pipecat-ai/pipecat/commit/144ea36c81e0698163493e945a33d5b5b4ebd406 but for regular Gemini instead of Gemini Live ; problem reported by multiple users in https://github.com/pipecat-ai/pipecat/issues/2207


aconchillo:
LGTM! Thank you!

PR #2784, opened 2025-10-03
feat(elevenlabs): add new output formats and deprecate sample_rate-based format selection

disolaterX:
## Description

This PR enhances the ElevenLabs TTS service by adding support for additional output formats and introduces a more explicit API for format selection.

### Changes

1. **Added new output formats** to `ElevenLabsOutputFormat`:
   - **MP3 formats**: `mp3_22050_32`, `mp3_44100_32`, `mp3_44100_64`, `mp3_44100_96`, `mp3_44100_128`, `mp3_44100_192`
   - **PCM formats**: `pcm_8000` (new), `pcm_16000`, `pcm_22050`, `pcm_24000`, `pcm_44100`
   - **Telephony formats**: `ulaw_8000`, `alaw_8000`
   - **Opus formats**: `opus_48000_32`, `opus_48000_64`, `opus_48000_96`, `opus_48000_128`, `opus_48000_192`

2. **Introduced `output_format` parameter** to both `ElevenLabsTTSService` and `ElevenLabsHttpTTSService`:
   - Now at the same level as `voice_id` for better API consistency
   - Optional parameter with fallback to maintain backward compatibility

3. **Deprecated `output_format_from_sample_rate()` function**:
   - Added deprecation warning in docstring
   - Added runtime warning when function is called
   - Function will be removed in a future version

4. **Migration path**:
   - When `output_format` is not provided, automatically falls back to deriving format from `sample_rate`
   - Shows clear warning message encouraging users to migrate to the explicit `output_format` parameter

### Migration Guide

**Old usage:**
```python
service = ElevenLabsTTSService(
    api_key="...",
    voice_id="...",
    sample_rate=24000  # Format derived automatically
)
```

**New usage (recommended):**
```python
service = ElevenLabsTTSService(
    api_key="...",
    voice_id="...",
    output_format="pcm_24000",  # Explicit format
    sample_rate=24000
)
```

### Breaking Changes

⚠️ **Future breaking change**: The `output_format_from_sample_rate()` function and automatic format derivation will be removed in a future version. Users should migrate to explicitly specifying the `output_format` parameter.

### Backward Compatibility

✅ Fully backward compatible - existing code will continue to work with deprecation warnings

### DIFF
--- a/src/pipecat/services/elevenlabs/tts.py
+++ b/src/pipecat/services/elevenlabs/tts.py
@@ -50,7 +50,26 @@ except ModuleNotFoundError as e:
     logger.error("In order to use ElevenLabs, you need to `pip install pipecat-ai[elevenlabs]`.")
     raise Exception(f"Missing module: {e}")
 
-ElevenLabsOutputFormat = Literal["pcm_16000", "pcm_22050", "pcm_24000", "pcm_44100"]
+ElevenLabsOutputFormat = Literal[
+    "mp3_22050_32",
+    "mp3_44100_32",
+    "mp3_44100_64",
+    "mp3_44100_96",
+    "mp3_44100_128",
+    "mp3_44100_192",
+    "pcm_8000",
+    "pcm_16000",
+    "pcm_22050",
+    "pcm_24000",
+    "pcm_44100",
+    "ulaw_8000",
+    "alaw_8000",
+    "opus_48000_32",
+    "opus_48000_64",
+    "opus_48000_96",
+    "opus_48000_128",
+    "opus_48000_192",
+]
 
 # Models that support language codes
 # The following models are excluded as they don't support language codes:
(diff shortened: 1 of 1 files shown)


markbackman:
TTS services need to be configurable via sample rate. The best practice is to specify the sample rate in the PipelineParams for input or output. This ensures that all corresponding service use the same sample rate. So, the `output_format_from_sample_rate()` method is required. Also, I don't see a need for mp3 audio. Pipecat handles pcm audio.

I'd suggest reworking this PR with that in mind.

disolaterX:
> TTS services need to be configurable via sample rate. The best practice is to specify the sample rate in the PipelineParams for input or output. This ensures that all corresponding service use the same sample rate. So, the `output_format_from_sample_rate()` method is required. Also, I don't see a need for mp3 audio. Pipecat handles pcm audio.
> 
> I'd suggest reworking this PR with that in mind.

Hey, the goal was honestly to support telephony codecs. For example, an 8000 sample rate is available for PCM, ALAW, and ULAW. However, in that case, the current function fails to provide the correct information. What do you suggest we do in that case?

And about MP3, I understand your view and will be removing that.

markbackman:
I've thought through this more and I'm not sure there is an actual benefit. If I may ask, what are you looking to accomplish in supporting the additional output formats?

Pipecat already handles conversions, which have a negligible performance cost. I think the added API complexity might be more of a net negative than the performance boost of adding support for the other formats.

disolaterX:
https://github.com/pipecat-ai/pipecat/pull/2784#issuecomment-3388015592

> What are you looking to accomplish in supporting the additional output formats?
Goal is support developer that are using Pipecat to connect to PSTN telephone layer ie Asterisk or any other ePBXs at a scale and don't want to take the cost of transcoding. 


I suggest then we can default to auto finding the codec. If specified is mentioned, then take over; we don't run output_format_from_sample_rate 

what you say ?

markbackman:
What I'm saying is that I don't see the utility to using a different encodings. What use case do you have in mind and what benefit does it provide?

markbackman:
Thanks again for proposing this, but the different encodings add complexity without much benefit. For that reason, we'll close this PR.

PR #2794, opened 2025-10-06
Verifying WhatsApp signature to ensure the webhook request is from WhatsApp.

filipi87:
Verifying WhatsApp signature to ensure the webhook request is from WhatsApp.


src/pipecat/transports/whatsapp/client.py
```diff
@@ -133,10 +138,32 @@ async def handle_verify_webhook_request(
 
         return int(challenge)
 
+    async def _validate_whatsapp_webhook_request(self, raw_body: bytes, sha256_signature: str):
```
filipi87, line 141: Before we were not actually checking that the payload came from whatsapp.

You can check the docs here, "validating payloads":
- https://developers.facebook.com/docs/graph-api/webhooks/getting-started

aconchillo:
LGTM

PR #2795, opened 2025-10-06
Support Gemini Live + Vertex AI

kompfner:
Note that this PR is not open against `main`. Let's get https://github.com/pipecat-ai/pipecat/pull/2776 merged first.

Some final testing is still ongoing.


src/pipecat/services/gemini_live/vertex.py
```diff
@@ -0,0 +1,187 @@
...
+    """
+
+    def __init__(
+        self,
+        *,
+        credentials: Optional[str] = None,
+        credentials_path: Optional[str] = None,
+        location: str = "us-east4",
+        project_id: str,
+        model="google/gemini-2.0-flash-live-preview-04-09",
```
kompfner, line 57: This was the model that most documentation still seemed to point to as a good default

src/pipecat/services/gemini_live/vertex.py
```diff
@@ -0,0 +1,187 @@
...
+    streaming audio responses, and tool usage.
+    """
+
+    def __init__(
+        self,
+        *,
+        credentials: Optional[str] = None,
+        credentials_path: Optional[str] = None,
+        location: str = "us-east4",
+        project_id: str,
```
kompfner, line 56: It seems to me that `project_id` is _not_ optional when using Vertex.

If I'm correct about that, then we have an issue in `GoogleVertexLLMService`—we should either make `InputParams`a non-optional arg in its initializer (thus requiring the user to specify a `project_id`) or do what we did here and pull `location` and `project_id` out of `InputParams` entirely and make them top-level initialization args. I prefer the 2nd approach, since `location` and `project_id` are project-level setup more akin to credentials; everything else in `InputParams` is concerned with model configuration.

src/pipecat/services/gemini_live/vertex.py
```diff
@@ -0,0 +1,187 @@
...
+    supporting both text and audio modalities. It handles voice transcription,
+    streaming audio responses, and tool usage.
+    """
+
+    def __init__(
+        self,
+        *,
+        credentials: Optional[str] = None,
+        credentials_path: Optional[str] = None,
+        location: str = "us-east4",
```
kompfner, line 55: Used the same default region as `GoogleVertexLLMService` (though admittedly I don't have the context on that decision)

src/pipecat/services/gemini_multimodal_live/gemini.py
```diff
@@ -1145,7 +1162,21 @@ async def _handle_msg_model_turn(self, msg: LiveServerMessage):
...
         if not inline_data:
             return
-        if inline_data.mime_type != f"audio/pcm;rate={self._sample_rate}":
+
+        # Check if mime type matches expected format
+        expected_mime_type = f"audio/pcm;rate={self._sample_rate}"
+        if inline_data.mime_type == expected_mime_type:
+            # Perfect match, continue processing
+            pass
+        elif inline_data.mime_type == "audio/pcm":
```
kompfner: When using Vertex the sample rate isn't included in the `mime_type` of received audio messages. This differs from the Gemini API.

src/pipecat/adapters/services/gemini_adapter.py
```diff
@@ -87,9 +87,11 @@ def to_provider_tools_format(self, tools_schema: ToolsSchema) -> List[Dict[str,
             Includes both converted standard tools and any custom Gemini-specific tools.
         """
         functions_schema = tools_schema.standard_tools
-        formatted_standard_tools = [
-            {"function_declarations": [func.to_default_dict() for func in functions_schema]}
-        ]
+        formatted_standard_tools = (
```
kompfner, line 90: When using Vertex, providing an empty list for "function_declarations" would result in an API error.

examples/foundational/26b-gemini-multimodal-live-function-calling.py
```diff
@@ -122,6 +122,9 @@ async def run_bot(transport: BaseTransport, runner_args: RunnerArguments):
         required=["location"],
     )
     search_tool = {"google_search": {}}
+    # KNOWN ISSUE: If using GeminiVertexMultimodalLiveLLMService, it appears
+    # you cannot use the "google_search" tool alongside other tools.
+    # See https://github.com/googleapis/python-genai/issues/941.
```
kompfner, line 127: Ran into this Vertex-specific issue.

examples/foundational/26d-gemini-multimodal-live-text.py
```diff
@@ -80,6 +80,8 @@
 async def run_bot(transport: BaseTransport, runner_args: RunnerArguments):
     logger.info(f"Starting bot")
 
+    # KNOWN ISSUE: If using GeminiVertexMultimodalLiveLLMService, it appears
+    # you cannot specify a modality other than AUDIO.
```
kompfner, line 84: Ran into this Vertex-specific issue.

src/pipecat/services/gemini_live/vertex.py
```diff
@@ -0,0 +1,187 @@
...
+            vertexai=True,
+            credentials=self._credentials,
+            project=self._project_id,
+            location=self._location,
+        )
+
+    @property
+    def file_api(self):
+        """Gemini File API is not supported with Vertex AI."""
+        raise NotImplementedError(
```
kompfner, line 137: This is how we could handle someone attempting to use `llm.file_api` when using Vertex.

src/pipecat/services/gemini_multimodal_live/gemini.py
```diff
@@ -1178,7 +1209,11 @@ async def _handle_msg_tool_call(self, message: LiveServerMessage):
         function_calls_llm = [
             FunctionCallFromLLM(
                 context=self._context,
-                tool_call_id=f.id,
+                tool_call_id=(
+                    # NOTE: when using Vertex AI we don't get server-provided
+                    # tool call IDs here
```
kompfner: This appears to be a Vertex-specific quirk.

src/pipecat/services/gemini_live/vertex.py
markbackman: I think this is fine. Developers will most likely have to specify their own region.

src/pipecat/services/gemini_live/vertex.py
markbackman: I think you're right. We might want to add `location` to the args and deprecate the InputParams version of `location`.

src/pipecat/services/gemini_live/vertex.py
markbackman: For Vertex, I found this page:
https://cloud.google.com/vertex-ai/generative-ai/docs/live-api

It seems like `gemini-live-2.5-flash-preview-native-audio-09-2025` might be the latest option. I've seen stability issues in testing with this model on AI studio though.

src/pipecat/services/gemini_live/vertex.py
```diff
@@ -0,0 +1,187 @@
...
+        location: str = "us-east4",
+        project_id: str,
+        model="google/gemini-2.0-flash-live-preview-04-09",
+        voice_id: str = "Charon",
+        start_audio_paused: bool = False,
+        start_video_paused: bool = False,
+        system_instruction: Optional[str] = None,
+        tools: Optional[Union[List[dict], ToolsSchema]] = None,
+        params: Optional[InputParams] = None,
+        inference_on_context_initialization: bool = True,
```
markbackman, line 64: I noticed this property when testing the non-vertex service. How should this property be used? Is it in a way duplicative of the `LLMRunFrame` or should it be used differently.

src/pipecat/services/gemini_live/vertex.py
markbackman: For now, I think we have to punt on this since it isn't a native feature.

src/pipecat/services/gemini_live/vertex.py
agarwalmehul: For default, you could also use "gemini-live-2.5-flash-preview". Its 1 model behind "gemini-live-2.5-flash-preview-native-audio-09-2025" and seems more stable.

src/pipecat/services/gemini_multimodal_live/gemini.py
```diff
@@ -4,1587 +4,48 @@
...
-        SlidingWindow,
-        SpeechConfig,
-        StartSensitivity,
-        ThinkingConfig,
-        VoiceConfig,
+from pipecat.services.gemini_live.gemini import GeminiMediaResolution as _GeminiMediaResolution
+from pipecat.services.gemini_live.gemini import GeminiVADParams as _GeminiVADParams
+from pipecat.services.gemini_live.gemini import InputParams as _InputParams
+
+with warnings.catch_warnings():
```
kompfner, line 31: I thought it made the most sense to update the import path for all these types. Doing that, we could also capture usage of any of the old types with this single file-level deprecation warning.

I didn't include `language_to_gemini_language` in this file even though it's technically importable since it seemed like clearly a "private" thing.

src/pipecat/services/gemini_multimodal_live/gemini.py
```diff
@@ -4,1587 +4,48 @@
...
+    warnings.simplefilter("always")
+    warnings.warn(
+        "Types in pipecat.services.gemini_multimodal_live.gemini are deprecated. "
```
markbackman, line 34: The deprecation warning should also account for developers that are still using:
```
from pipecat.services.gemini_multimodal_live.gemini import GeminiMultimodalLiveLLMService
```

and `GeminiMultimodalLiveLLMService` in their code. They need to be notified of the service name deprecation, too.

markbackman: So, maybe the message just needs to be made more generic to mention the service and the types.

kompfner:
This PR appears to have gotten into a broken state since the last force push. Closing and re-opening to try to fix it.

kompfner:
Ah, I see. https://www.githubstatus.com/ is showing that Github PRs are having an outage.

kompfner:
Trying re-opening this PR, as the other one has stalled out too

src/pipecat/services/gemini_live/vertex.py
kompfner: https://github.com/pipecat-ai/pipecat/pull/2823

examples/foundational/26d-gemini-multimodal-live-text.py
patrickloeber: response_modalities=["TEXT"] is not supported for native audio models. but this is also the case for the Gemini API, not only Vertex specific

examples/foundational/26b-gemini-multimodal-live-function-calling.py
patrickloeber: Yes, known issue. And note that url context tool & google search tool currently can also cannot combined on the Gemini API with native audio models

examples/foundational/26d-gemini-multimodal-live-text.py
kompfner: Hmm, to me it appears that TEXT _does_ work with the Gemini API. That's actually what this example file is illustrating—it asks Gemini Live to respond with only text, and then uses an external TTS service to do the voice.

patrickloeber: Yes, on Gemini API TEXT should work for `gemini-live-2.5-flash-preview` and `gemini-2.0-flash-live-001`, but not for the `native-audio` models. And it should be the same on Vertex, see also the info box on their docs: https://cloud.google.com/vertex-ai/generative-ai/docs/live-api/tools#native-audio

kompfner: Ah gotcha! Misunderstood. Thanks for the explanation.

thecooltechguy: @patrickloeber on the https://ai.google.dev/gemini-api/docs/models#model-versions page, if you expand on the Gemini 2.5 Flash and then switch to the Gemini 2.5 Flash Live tab, it mentions Audio, Video, text as supported inputs and Audio and text as supported outputs.

dos this mean that native audio models like `gemini-2.5-flash-native-audio-preview-09-2025` support text response outputs?

patrickloeber: Hi, thanks for reporting! No, the docs are not accurate here, gemini-2.5-flash-native-audio-preview-09-2025 only supports AUDIO outputs. I'll get the docs updated

PR #2796, opened 2025-10-06
fix context order when using interruption strategies

aconchillo:

The issue we had is that the `InterruptionFrame` gets to the user aggregator before the assistant and therefore the user aggregation is added first. To ensure task interruptions don't cause any side effects we need we need to wait for the whole pipeline to be interrupted.

This PR adds a new frame `InterruptionCompletedFrame` that when using `InterruptionTaskFrame` will be pushed after the `InterruptionFrame` makes it all the way to the end of the pipeline. So:

- `InterruptionTaskFrame` is pushed upstream.
- The pipeline task pushes `InterruptionFrame`.
- When the `InterruptionFrame` makes it all the way to the end, the pipeline task pushed an `InterruptionCompletedFrame`.

This guarantees that assistant aggregator is updated before the user.


markbackman (approved):
This change makes a lot of sense. Is the timing of any other processor impacted by this change?

I would expect not because the new `InterruptionCompletedFrame` is a SystemFrame and it's directly queued. It should be very fast and not blocked.

Approving.

Are existing unit tests and evals sufficient to test for this? (I think so.)

aconchillo:
> This change makes a lot of sense. Is the timing of any other processor impacted by this change?
> 
> I would expect not because the new `InterruptionCompletedFrame` is a SystemFrame and it's directly queued. It should be very fast and not blocked.
> 

I don't think so.

> Approving.
> 
> Are existing unit tests and evals sufficient to test for this? (I think so.)

Yes, there are existing unit tests. I'm actually updating them.

PR #2804, opened 2025-10-07
Add (relatively spartan) reconnection logic to `GeminiMultimodalLiveLLMService`

kompfner:
...leveraging the Gemini Live session resumption mechanism.

Note that this PR is not open against `main`. Let's get https://github.com/pipecat-ai/pipecat/pull/2776 merged first.

@patrickloeber


src/pipecat/services/gemini_multimodal_live/gemini.py
```diff
@@ -890,6 +906,7 @@ async def _connect(self):
                 ),
                 input_audio_transcription=AudioTranscriptionConfig(),
                 output_audio_transcription=AudioTranscriptionConfig(),
+                session_resumption=SessionResumptionConfig(handle=session_resumption_handle),
```
kompfner, line 901: Note that I didn't make `session_resumption` a user-facing config.

Since session resumption tokens are only good for quite a limited time (2 hours), they seem mostly useful for recovering from socket disconnects. So let's keep it an implementation detail (for now, at least—we can always revisit if the need arises).

markbackman: This seems logical. I would expect the session resumption feature to be deprecated by Google in the future. Keeping it internal will save us from a future deprecation.

src/pipecat/services/gemini_multimodal_live/gemini.py
```diff
@@ -629,6 +631,12 @@ def __init__(
         )
         self._vad_params = params.vad
 
+        # Reconnection tracking
+        self._consecutive_failures = 0
+        self._max_consecutive_failures = 3
+        self._connection_established_threshold = 10.0  # seconds
```
markbackman: Maybe make these reconnection variables constants?

PR #2805, opened 2025-10-07
Allowing to update smallwebrtc and whatsapp properties.

filipi87:
Allowing to update smallwebrtc and whatsapp properties.


markbackman (approved):
LGTM

PR #2816, opened 2025-10-08
Implement ending `GeminiMultimodalLiveLLMService` gracefully (i.e. af…

kompfner:
…ter the bot is finished)


kompfner:
Note: I had originally gone down the path of pausing and resuming frame processing depending on whether the bot was in the middle of its turn, but after brainstorming with @aconchillo we decided that a more surgical approach like this would be preferable.

PR #2821, opened 2025-10-09
Sarvam STT/STTT WS implementation

shreyas-sarvam:
Implemented a new Sarvam speech-to-text service which contains speech-to-text and speech-to-text-translate streaming websocket.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,12 +5,15 @@ All notable changes to **Pipecat** will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-## [Unreleased]
+## [0.0.92] - 2025-10-31 🎃 "The Haunted Edition" 👻
 
 ### Added
 
 - Added supprt for Sarvam Speech-to-Text service (`SarvamSTTService`) with streaming WebSocket
   support for `saarika` (STT) and `saaras` (STT-translate) models.
+
+- Added a new `DeepgramHttpTTSService`, which delivers a meaningful reduction
+  in latency when compared to the `DeepgramTTSService`.
 
 - Add support for `speaking_rate` input parameter in `GoogleHttpTTSService`.
 
@@ -44,13 +47,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
   ```python
   context = LLMContext(messages, tools)
-  context_aggregator = LLMContextAggregatorPair(
-    context,
-    # This part is `OpenAIRealtimeLLMService`-specific.
-    # `expect_stripped_words=False` needed when OpenAI Realtime used with
-    # "audio" modality (the default).
-    assistant_params=LLMAssistantAggregatorParams(expect_stripped_words=False),
-  )
+  context_aggregator = LLMContextAggregatorPair(context)
   ```
 
   (Note that even though `OpenAIRealtimeLLMService` now supports the universal
@@ -127,13 +124,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
   ```python
   context = LLMContext(messages, tools)
-  context_aggregator = LLMContextAggregatorPair(
-    context,
-    # This part is `GeminiLiveLLMService`-specific.
-    # `expect_stripped_words=False` needed when Gemini Live used with AUDIO
-    #  modality (the default).
-    assistant_params=LLMAssistantAggregatorParams(expect_stripped_words=False),
-  )
+  context_aggregator = LLMContextAggregatorPair(context)
   ```
 
   (Note that even though `GeminiLiveLLMService` now supports the universal
@@ -175,12 +166,18 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
-- `UserImageRawFrame` new fields `add_to_context` and `text`. The
-  `add_to_context` field indicates if this image and text should be added to the
-  LLM context (by the LLM assistant aggregator). The `text` field, if set, might
-  also guide the LLM or the vision service on how to analyze the image.
+- The development runner's `/start` endpoint now supports passing
+  `dailyRoomProperties` and `dailyMeetingTokenProperties` in the request body
+  when `createDailyRoom` is true. Properties are validated against the
+  `DailyRoomProperties` and `DailyMeetingTokenProperties` types respectively
+  and passed to Daily's room and token creation APIs.
 
-- `UserImageRequestFrame` new fiels `add_to_context` and `text`. Both fields
+- `UserImageRawFrame` new fields `append_to_context` and `text`. The
+  `append_to_context` field indicates if this image and text should be added to
+  the LLM context (by the LLM assistant aggregator). The `text` field, if set,
+  might also guide the LLM or the vision service on how to analyze the image.
+
+- `UserImageRequestFrame` new fiels `append_to_context` and `text`. Both fields
   will be used to set the same fields on the captured `UserImageRawFrame`.
 
 - `UserImageRequestFrame` don't require function call name and ID anymore.
@@ -213,6 +210,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Deprecated
 
+- The `expect_stripped_words` parameter of `LLMAssistantAggregatorParams` is
+  ignored when used with the newer `LLMAssistantAggregator`, which now handles
+  word spacing automatically.
+
 - `LLMService.request_image_frame()` is deprecated, push a
   `UserImageRequestFrame` instead.
 
@@ -238,6 +239,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   used.
 
 ### Fixed
+
+- Fixed a `PipelineTask` issue that was causing an idle timeout for frames that
+  were being generated but not reaching the end of the pipeline. Since the exact
+  point when frames are discarded is unknown, we now monitor pipeline frames
+  using an observer. If the observer detects frames are being generated, it will
+  prevent the pipeline from being considered idle.
 
 - Fixed an issue in `HumeTTSService` that was only using Octave 2, which does
   not support the `description` field. Now, if a description is provided, it
--- a/examples/foundational/07c-interruptible-deepgram-http.py
+++ b/examples/foundational/07c-interruptible-deepgram-http.py
@@ -1,0 +1,132 @@
+#
+# Copyright (c) 2024–2025, Daily
+#
+# SPDX-License-Identifier: BSD 2-Clause License
+#
+
+
+import os
+
+import aiohttp
+from dotenv import load_dotenv
+from loguru import logger
+
+from pipecat.audio.turn.smart_turn.base_smart_turn import SmartTurnParams
+from pipecat.audio.turn.smart_turn.local_smart_turn_v3 import LocalSmartTurnAnalyzerV3
+from pipecat.audio.vad.silero import SileroVADAnalyzer
+from pipecat.audio.vad.vad_analyzer import VADParams
+from pipecat.frames.frames import LLMRunFrame
+from pipecat.pipeline.pipeline import Pipeline
+from pipecat.pipeline.runner import PipelineRunner
+from pipecat.pipeline.task import PipelineParams, PipelineTask
+from pipecat.processors.aggregators.llm_context import LLMContext
+from pipecat.processors.aggregators.llm_response_universal import LLMContextAggregatorPair
+from pipecat.runner.types import RunnerArguments
+from pipecat.runner.utils import create_transport
+from pipecat.services.deepgram.stt import DeepgramSTTService
+from pipecat.services.deepgram.tts import DeepgramHttpTTSService
+from pipecat.services.openai.llm import OpenAILLMService
+from pipecat.transports.base_transport import BaseTransport, TransportParams
+from pipecat.transports.daily.transport import DailyParams
+from pipecat.transports.websocket.fastapi import FastAPIWebsocketParams
+
+load_dotenv(override=True)
+
+
+# We store functions so objects (e.g. SileroVADAnalyzer) don't get
+# instantiated. The function will be called when the desired transport gets
+# selected.
+transport_params = {
+    "daily": lambda: DailyParams(
+        audio_in_enabled=True,
+        audio_out_enabled=True,
+        vad_analyzer=SileroVADAnalyzer(params=VADParams(stop_secs=0.2)),
+        turn_analyzer=LocalSmartTurnAnalyzerV3(params=SmartTurnParams()),
+    ),
+    "twilio": lambda: FastAPIWebsocketParams(
+        audio_in_enabled=True,
+        audio_out_enabled=True,
+        vad_analyzer=SileroVADAnalyzer(params=VADParams(stop_secs=0.2)),
+        turn_analyzer=LocalSmartTurnAnalyzerV3(params=SmartTurnParams()),
+    ),
+    "webrtc": lambda: TransportParams(
+        audio_in_enabled=True,
+        audio_out_enabled=True,
+        vad_analyzer=SileroVADAnalyzer(params=VADParams(stop_secs=0.2)),
+        turn_analyzer=LocalSmartTurnAnalyzerV3(params=SmartTurnParams()),
+    ),
+}
+
+
+async def run_bot(transport: BaseTransport, runner_args: RunnerArguments):
+    logger.info(f"Starting bot")
+
+    async with aiohttp.ClientSession() as session:
+        stt = DeepgramSTTService(api_key=os.getenv("DEEPGRAM_API_KEY"))
+
+        tts = DeepgramHttpTTSService(
+            api_key=os.getenv("DEEPGRAM_API_KEY"),
+            voice="aura-2-andromeda-en",
+            aiohttp_session=session,
+        )
+
+        llm = OpenAILLMService(api_key=os.getenv("OPENAI_API_KEY"))
+
+        messages = [
+            {
+                "role": "system",
+                "content": "You are a helpful LLM in a WebRTC call. Your goal is to demonstrate your capabilities in a succinct way. Your output will be converted to audio so don't include special characters in your answers. Respond to what the user said in a creative and helpful way.",
+            },
+        ]
+
+        context = LLMContext(messages)
+        context_aggregator = LLMContextAggregatorPair(context)
+
+        pipeline = Pipeline(
+            [
+                transport.input(),  # Transport user input
+                stt,  # STT
+                context_aggregator.user(),  # User responses
+                llm,  # LLM
+                tts,  # TTS
+                transport.output(),  # Transport bot output
+                context_aggregator.assistant(),  # Assistant spoken responses
+            ]
+        )
+
+        task = PipelineTask(
+            pipeline,
+            params=PipelineParams(
+                enable_metrics=True,
+                enable_usage_metrics=True,
+            ),
+            idle_timeout_secs=runner_args.pipeline_idle_timeout_secs,
+        )
+
+        @transport.event_handler("on_client_connected")
+        async def on_client_connected(transport, client):
+            logger.info(f"Client connected")
+            # Kick off the conversation.
+            messages.append({"role": "system", "content": "Please introduce yourself to the user."})
+            await task.queue_frames([LLMRunFrame()])
+
+        @transport.event_handler("on_client_disconnected")
+        async def on_client_disconnected(transport, client):
+            logger.info(f"Client disconnected")
+            await task.cancel()
+
+        runner = PipelineRunner(handle_sigint=runner_args.handle_sigint)
+
+        await runner.run(task)
+
+
+async def bot(runner_args: RunnerArguments):
+    """Main bot entry point compatible with Pipecat Cloud."""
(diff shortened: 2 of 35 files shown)


pyproject.toml
```diff
@@ -38,6 +38,7 @@ dependencies = [
     # Pinning numba to resolve package dependencies
     "numba==0.61.2",
     "wait_for2>=0.4.1; python_version<'3.12'",
+    "sarvamai==0.1.21",
```
markbackman: Can you remove this? This is for required dependencies only. Sarvam should be only in the optional dependency list.

pyproject.toml
```diff
@@ -93,7 +94,7 @@ rime = [ "pipecat-ai[websockets-base]" ]
 riva = [ "nvidia-riva-client~=2.21.1" ]
 runner = [ "python-dotenv>=1.0.0,<2.0.0", "uvicorn>=0.32.0,<1.0.0", "fastapi>=0.115.6,<0.117.0", "pipecat-ai-small-webrtc-prebuilt>=1.0.0"]
 sambanova = []
-sarvam = [ "pipecat-ai[websockets-base]" ]
+sarvam = [ "sarvamai==0.1.21", "websockets>=13.1,<15.0" ]
```
markbackman: Please use the `"pipecat-ai[websockets-base]"` to reference websockets. This helps to keep this section DRY:
```suggestion
sarvam = [ "sarvamai==0.1.21", "pipecat-ai[websockets-base]" ]
```

src/pipecat/services/sarvam/stt.py
```diff
@@ -0,0 +1,411 @@
...
+    from sarvamai import AsyncSarvamAI
+    from sarvamai.core.api_error import ApiError
+    from sarvamai.core.events import EventType
+except ModuleNotFoundError as e:
+    logger.error(f"Exception: {e}")
+    logger.error("In order to use Sarvam, you need to `pip install pipecat-ai[sarvam]`.")
+    raise Exception(f"Missing module: {e}")
+
+
+class TranscriptionMetrics(BaseModel):
```
markbackman: Are these types defined in your python package by any chance?

pyproject.toml
shreyas-sarvam: Okay, made the change.

pyproject.toml
shreyas-sarvam: Sure, made the change.

src/pipecat/services/sarvam/stt.py
shreyas-sarvam: Yes, these types are defined in the `sarvamai` package.
However, the rest of these classes aren't needed as they are handled internally by the package. I have updated the code accordingly.

src/pipecat/services/sarvam/stt.py
```diff
@@ -0,0 +1,341 @@
...
+)
+from pipecat.services.stt_service import STTService
+from pipecat.transcriptions.language import Language
+from pipecat.utils.time import time_now_iso8601
+from pipecat.utils.tracing.service_decorators import traced_stt
+
+try:
+    from sarvamai import AsyncSarvamAI
+    from sarvamai.core.api_error import ApiError
+    from sarvamai.core.events import EventType
```
markbackman, line 29: This looks unused. Remove if it's not needed.

src/pipecat/services/sarvam/stt.py
```diff
@@ -0,0 +1,341 @@
...
+
+    Provides real-time speech recognition using Sarvam's WebSocket API.
+    """
+
+    def __init__(
+        self,
+        *,
+        api_key: str,
+        model: str = "saarika:v2.5",
+        language_code: Language = Language.HI_IN,
```
markbackman: Let's use `language` as the arg name here to align with other STTService classes.

```suggestion
        language: Language = Language.HI_IN,
```

src/pipecat/services/sarvam/stt.py
```diff
@@ -0,0 +1,341 @@
...
+
+    def can_generate_metrics(self) -> bool:
+        """Check if this service can generate processing metrics.
+
+        Returns:
+            True, as Sarvam service supports metrics generation.
+        """
+        return True
+
+    async def set_model(self, model: str):
```
markbackman: You can remove this function. The base STTService class has a `set_model()` method that ties into the metrics system.
https://github.com/pipecat-ai/pipecat/blob/main/src/pipecat/services/stt_service.py#L106-L112

src/pipecat/services/sarvam/stt.py
```diff
@@ -0,0 +1,341 @@
...
+            model: Sarvam model to use for transcription.
+            language_code: Language enum for transcription (e.g., Language.HI_IN, Language.KN_IN).
+            sample_rate: Audio sample rate. Defaults to 16000 if not specified.
+            **kwargs: Additional arguments passed to the parent STTService.
+        """
+        super().__init__(sample_rate=sample_rate, **kwargs)
+
+        self.set_model_name(model)
+        self._api_key = api_key
+        self._model = model
```
markbackman: By calling `set_model_name`, that hooks into the inheritance chain, where the model becomes available as the AIService `@property` of `model_name`, referenced as `self.model_name`. Let's use that instead, in place of `self._model`.

src/pipecat/services/sarvam/stt.py
markbackman: Actually, let's break out `language` to a new class inside of `SarvamSTTService` called `InputParams`. Other service classes use an InputParams class to define how the service is used:

```
    class InputParams(BaseModel):
        """Configuration parameters for Sarvam STT service.

        Parameters:
            language: Target language for transcription. Defaults to HI_IN.
        """

        language: Optional[Language] = Language.HI_IN
```

Then, add an `arg` called `params` to `__init__`:
```
```

Then initialize as:
```
params = params or SarvamSTTService.InputParams()
```

Check out other services to see how this works if you have questions.

src/pipecat/services/sarvam/stt.py
```diff
@@ -0,0 +1,341 @@
...
+        self._language_code = language_code
```
markbackman: This will become:

```suggestion
        self._language_enum = params.language or Language. HI_IN
        self._language_string = self.language_to_service_language(self._language_enum) or "hi-IN"
```

src/pipecat/services/sarvam/stt.py
```diff
@@ -0,0 +1,341 @@
...
+                self._websocket_context = self._sarvam_client.speech_to_text_streaming.connect(
+                    language_code=self._language_string,
+                    model=self._model,
+                    vad_signals=True,
+                    high_vad_sensitivity=True,
+                    sample_rate=str(self.sample_rate),
+                    input_audio_codec="wav",
+                )
+            else:
+                # STT-translate service - auto-detects language
```
markbackman: Does the STT-translate service auto-detect the language and transcribe. Or, does it also translate to other languages? Just making sure I understand the difference. If it's the former, then maybe the comment should be more clear, as including "translation" indicates that the speech will be outputted in a different language.

src/pipecat/services/sarvam/stt.py
```diff
@@ -0,0 +1,341 @@
...
+                    logger.debug("User started speaking")
+                    await self._call_event_handler("on_speech_started")
+
+            elif message.type == "data":
+                await self.stop_ttfb_metrics()
+                transcript = message.data.transcript
+                language_code = message.data.language_code
+                if language_code is None:
+                    language_code = "hi-IN"
+                language = self._map_language_code_to_enum(language_code)
```
markbackman: Does this need to be detected from the message because the language may have been auto-detected?

src/pipecat/services/sarvam/stt.py
```diff
@@ -0,0 +1,341 @@
...
+            "as-IN": Language.AS_IN,
+        }
+        return mapping.get(language_code, Language.HI_IN)
+
+    async def start_metrics(self):
+        """Start TTFB and processing metrics collection."""
+        await self.start_ttfb_metrics()
+        await self.start_processing_metrics()
+
+    @traced_stt
```
markbackman: This isn't being called. You can just call it from the same place that you're creating the TranscriptionFrame.

markbackman:
Looking good so far. A few additional things:

1. Add a changelog entry, adding the new service
2. Update your foundational examples to use your STT service (07z*)

src/pipecat/services/sarvam/stt.py
shreyas-sarvam: Resolved.
It is being used now `self._socket_client.on(EventType.MESSAGE, _message_handler)`

src/pipecat/services/sarvam/stt.py
shreyas-sarvam: Made the change, resolved.

src/pipecat/services/sarvam/stt.py
shreyas-sarvam: Made the change.

src/pipecat/services/sarvam/stt.py
shreyas-sarvam: Made the change.

src/pipecat/services/sarvam/stt.py
shreyas-sarvam: Updated.

src/pipecat/services/sarvam/stt.py
shreyas-sarvam: STT-Translate always returns the output in English language.
Updated the comment.

src/pipecat/services/sarvam/stt.py
shreyas-sarvam: Updated to use this function.

shreyas-sarvam:
> Looking good so far. A few additional things:
> 
> 1. Add a changelog entry, adding the new service
> 2. Update your foundational examples to use your STT service (07z*)

I have made the changes you had requested. Looking forward to the next steps.

examples/foundational/07z-interruptible-sarvam-http.py
```diff
@@ -63,7 +63,11 @@ async def run_bot(transport: BaseTransport, runner_args: RunnerArguments):
 
     # Create an HTTP session
     async with aiohttp.ClientSession() as session:
-        stt = DeepgramSTTService(api_key=os.getenv("DEEPGRAM_API_KEY"))
+        stt = SarvamSTTService(
+            api_key=os.getenv("SARVAM_API_KEY"),
+            model="saarika:v2.5",
+            params=SarvamSTTService.InputParams(language=None),
```
markbackman: Setting `language=None` causes the example to not run. You might want to either omit or include it with a language enum set. You could comment it out, as a clue of how to use it or leave it in.

examples/foundational/07z-interruptible-sarvam.py
```diff
@@ -62,7 +62,12 @@
 async def run_bot(transport: BaseTransport, runner_args: RunnerArguments):
     logger.info(f"Starting bot")
 
-    stt = DeepgramSTTService(api_key=os.getenv("DEEPGRAM_API_KEY"))
+    stt = SarvamSTTService(
+        api_key=os.getenv("SARVAM_API_KEY"),
+        model="saarika:v2.5",
+        # Example: set Hindi; omit or change via set_language at runtime
+        params=SarvamSTTService.InputParams(language=None),
```
markbackman: Same comment here.

src/pipecat/services/sarvam/stt.py
```diff
@@ -0,0 +1,387 @@
...
+        Language.GU_IN: "gu-IN",
+        Language.HI_IN: "hi-IN",
+        Language.KN_IN: "kn-IN",
+        Language.ML_IN: "ml-IN",
+        Language.MR_IN: "mr-IN",
+        Language.TA_IN: "ta-IN",
+        Language.TE_IN: "te-IN",
+        Language.PA_IN: "pa-IN",
+        Language.OR_IN: "od-IN",
+        Language.EN_US: "en-US",
```
markbackman: I tried `EN_US`, but get an API error when testing:
```
Failed to send audio: received 1000 (OK) 'en-US' is not a valid SpeechToTextLanguage; then sent 1000 (OK) 'en-US' is not a valid SpeechToTextLanguage, fatal: False
```

src/pipecat/services/sarvam/stt.py
```diff
@@ -0,0 +1,387 @@
...
+                )
+
+            # Enter the async context manager
+            self._socket_client = await self._websocket_context.__aenter__()
+
+            # Register event handler for incoming messages
+            def _message_handler(message):
+                """Wrapper to handle async response handler."""
+                try:
+                    loop = asyncio.get_running_loop()
```
markbackman: I see you're using your own task management. Pipecat comes with built-in task management, that I'd recommend you should use.

Also, I'd recommend structuring your file more similarly to the other STT services for better maintenance. I wrote up quick version that uses Pipecat task management and does some restructuring to be more similar to other services. You can see it in this gist:
https://gist.github.com/markbackman/af260cce7c5766bd69dc8edc44e06473

I tested minimally and see it working as expected. I'd recommend using those changes and doing more tests to ensure this meets your expectations.

shreyas-sarvam: Sure, I have made a couple of changes to this to make input parameters accessible to users.

src/pipecat/services/sarvam/stt.py
shreyas-sarvam: It was an error. `en-us` is not supported by Sarvam APIs.
Fixed this.

examples/foundational/07z-interruptible-sarvam-http.py
shreyas-sarvam: Resolved.

examples/foundational/07z-interruptible-sarvam.py
shreyas-sarvam: Resolved.

examples/foundational/07z-interruptible-sarvam.py
```diff
@@ -24,9 +24,10 @@
 from pipecat.processors.aggregators.llm_response_universal import LLMContextAggregatorPair
 from pipecat.runner.types import RunnerArguments
 from pipecat.runner.utils import create_transport
-from pipecat.services.deepgram.stt import DeepgramSTTService
 from pipecat.services.openai.llm import OpenAILLMService
+from pipecat.services.sarvam.stt import SarvamSTTService
 from pipecat.services.sarvam.tts import SarvamTTSService
+from pipecat.transcriptions.language import Language
```
markbackman: Remove unused import.

src/pipecat/services/sarvam/stt.py
```diff
@@ -0,0 +1,475 @@
...
+        self.set_model_name(model)
+        self._api_key = api_key
+        self._language_code = params.language
+        # For saarika models, default to "unknown" if language is not provided
+        if params.language:
+            self._language_string = language_to_sarvam_language(params.language)
+        elif "saarika" in model.lower():
+            self._language_string = "unknown"
+        else:
+            self._language_string = None
```
markbackman, line 137: I don't think you want a language of `"unknown"` or `None`, as that causes an API error, at least for the `saarika` model. Don't you want to default to a language so that the API will work if one isn't provided?

src/pipecat/services/sarvam/stt.py
```diff
@@ -0,0 +1,475 @@
...
+        self._prompt = params.prompt
+
+        # Store connection parameters
+        self._sample_rate = params.sample_rate
```
markbackman: No need to store sample rate. It's already handled in the base class.

You can access it as `self.sample_rate` subsequently.

It's important to follow this pattern as the sample rate is set from the StartFrame, which is handled by the STTService base class.

src/pipecat/services/sarvam/stt.py
```diff
@@ -0,0 +1,475 @@
...
+                if self._input_audio_codec.startswith("audio/")
+                else f"audio/{self._input_audio_codec}"
+            )
+
+            # Build method arguments
+            method_kwargs = {
+                "audio": audio_base64,
+                "encoding": encoding,
+            }
+            # Only include sample_rate if provided in params
```
markbackman: You will always have a sample rate, set to `self.sample_rate` that you can apply. Please use that.

src/pipecat/services/sarvam/stt.py
shreyas-sarvam: `saarika` supports "unknown" language code.
It auto-detects language in this case.

src/pipecat/services/sarvam/stt.py
shreyas-sarvam: Understood, resolved.

src/pipecat/services/sarvam/stt.py
shreyas-sarvam: Resolved.

examples/foundational/07z-interruptible-sarvam.py
shreyas-sarvam: Resolved.

src/pipecat/services/sarvam/stt.py
```diff
@@ -0,0 +1,472 @@
...
+    """
+
+    class InputParams(BaseModel):
+        """Configuration parameters for Sarvam STT service.
+
+        Parameters:
+            language: Target language for transcription. Defaults to None (required for saarika models).
+            prompt: Optional prompt to guide translation style/context for STT-Translate models.
+                   Only applicable to saaras (STT-Translate) models. Defaults to None.
+            sample_rate: Audio sample rate in Hz. Overrides the parent sample_rate if provided.
```
markbackman: Remove
```suggestion
```

src/pipecat/services/sarvam/stt.py
```diff
@@ -0,0 +1,472 @@
...
+            vad_signals: Enable VAD signals in response. Defaults to True.
+            high_vad_sensitivity: Enable high VAD (Voice Activity Detection) sensitivity. Defaults to False.
+            input_audio_codec: Audio codec/format of the input file. Defaults to "wav".
+        """
+
+        language: Optional[Language] = None
+        prompt: Optional[str] = None
+        sample_rate: Optional[int] = None
```
markbackman: Remove
```suggestion
```

src/pipecat/services/sarvam/stt.py
```diff
@@ -0,0 +1,472 @@
...
+        vad_signals: bool = True
+        high_vad_sensitivity: bool = False
+        input_audio_codec: str = "wav"
+
+    def __init__(
+        self,
+        *,
+        api_key: str,
+        model: str = "saarika:v2.5",
+        sample_rate: Optional[int] = None,
```
markbackman, line 93: This is the only sample_rate arg you should have.

src/pipecat/services/sarvam/stt.py
```diff
@@ -0,0 +1,472 @@
...
+        # Validate that saarika models don't accept prompt parameter
+        if "saarika" in model.lower():
+            if params.prompt is not None:
+                raise ValueError(
+                    f"Model '{model}' does not accept prompt parameter. "
+                    "Prompts are only supported for STT-Translate models"
+                )
+
+        # Use sample_rate from params if provided, otherwise use the parameter
+        sample_rate = params.sample_rate if params.sample_rate is not None else sample_rate
```
markbackman:
```suggestion
```

Remove. This is not needed.

Sample rate works like this:

The best practice is to set in the PipelineParams. This results in the audio_in_sample_rate being set in the StartFrame. Before starting, each processor handles the StartFrame and parses the corresponding values. STT services parse `audio_in_sample_rate`.

Also, in the service code, you have a `sample_rate` arg, which is passed to super().

In the STTService base class, this is all handled here:
https://github.com/pipecat-ai/pipecat/blob/main/src/pipecat/services/stt_service.py#L137-L145

The STTService base class exposes sample_rate as a property, which you can use as `self.sample_rate`.

So, in your service code, you just need:
1. an arg for sample_rate
2. Pass that sample rate arg to super()
3. Use sample rate as `self.sample_rate` in your code.

src/pipecat/services/sarvam/stt.py
```diff
@@ -0,0 +1,472 @@
...
+            high_vad_sensitivity: Enable high VAD (Voice Activity Detection) sensitivity. Defaults to False.
+            input_audio_codec: Audio codec/format of the input file. Defaults to "wav".
+        """
+
+        language: Optional[Language] = None
+        prompt: Optional[str] = None
+        sample_rate: Optional[int] = None
+        vad_signals: bool = True
+        high_vad_sensitivity: bool = False
+        input_audio_codec: str = "wav"
```
markbackman: This should probably be an arg, since it's fundamental to how this works, yes? Is `wav` the default input_audio_codec?

src/pipecat/services/sarvam/stt.py
shreyas-sarvam: Removed.

src/pipecat/services/sarvam/stt.py
shreyas-sarvam: Removed.

src/pipecat/services/sarvam/stt.py
shreyas-sarvam: Understood.

src/pipecat/services/sarvam/stt.py
shreyas-sarvam: This clears my understanding.
Made the change to reflect this.

src/pipecat/services/sarvam/stt.py
shreyas-sarvam: Yes precisely.
Add this as an arg instead of an `InputParam`.

markbackman (approved):
LGTM! Thanks for working through the comments with me. Congrats 🎉

PR #2823, opened 2025-10-09
Move `location` and `project_id` out of `InputParams` in `GoogleVerte…

kompfner:
…xLLMService`, making them top-level init args instead. We do this for two reasons:

- Conceptually, these args comprise project-level setup, akin to credentials, whereas everything in `InputParams` is concerned with model configuration
- Providing a `project_id` when initializing `GoogleVertexLLMService` should not be optional, but prior to the change in this commit it was (erroneously) treated as optional by dint of `InputParams` being optional

This improvement was discussed [in this comment](https://github.com/pipecat-ai/pipecat/pull/2795#discussion_r2408279142).

### DIFF
--- a/src/pipecat/services/google/llm_vertex.py
+++ b/src/pipecat/services/google/llm_vertex.py
@@ -51,32 +51,45 @@ class GoogleVertexLLMService(OpenAILLMService):
     class InputParams(OpenAILLMService.InputParams):
         """Input parameters specific to Vertex AI.
 
-        .. deprecated:: 0.0.90
-            Use `OpenAILLMService.InputParams` instead and provide `location` and
-            `project_id` as direct arguments to `GoogleVertexLLMService.__init__()`.
-
         Parameters:
             location: GCP region for Vertex AI endpoint (e.g., "us-east4").
+
+                .. deprecated:: 0.0.90
+                    Use `location` as a direct argument to
+                    `GoogleVertexLLMService.__init__()` instead.
+
             project_id: Google Cloud project ID.
+
+                .. deprecated:: 0.0.90
+                    Use `project_id` as a direct argument to
+                    `GoogleVertexLLMService.__init__()` instead.
         """
 
         # https://cloud.google.com/vertex-ai/generative-ai/docs/learn/locations
-        location: str = "us-east4"
-        project_id: str
+        location: Optional[str] = None
+        project_id: Optional[str] = None
 
         def __init__(self, **kwargs):
             """Initializes the InputParams."""
             import warnings
 
             with warnings.catch_warnings():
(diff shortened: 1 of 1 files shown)


src/pipecat/services/google/llm_vertex.py
```diff
@@ -51,6 +51,10 @@ class GoogleVertexLLMService(OpenAILLMService):
     class InputParams(OpenAILLMService.InputParams):
         """Input parameters specific to Vertex AI.
 
+        .. deprecated:: 0.0.90
+            Use `OpenAILLMService.InputParams` instead and provide `location` and
```
markbackman: Ultimately, we want to remove the `InputParams` class and the corresponding `params` arg in the constructor. But due to the inheritance, won't you still be able to use the base class InputParams as `GoogleVertexLLMService.InputParams()`? That's how it works for other services, if I understand correctly.

I'm asking only because of the guidance to use `OpenAILLMService.InputParams`, which shouldn't be needed.

kompfner: Ah yes great catch! Hadn't considered the fact that `GoogleVertexLLMService.InputParams` will just fall through to the base class's `InputParams`. Will make some changes here accordingly.

kompfner: Ok so instead of putting a deprecation warning on `GoogleVertexLLMService.InputParams` as a whole I think I'll just put a deprecation warning on each of the properties that we want to move to the `GoogleVertexLLMService` initializer. Because you're right that we don't want to warn people away from referring to `GoogleVertexLLMService.InputParams` in their code—they shoulda absolutely use it, just not use the properties that aren't in the superclass. And then when `GoogleVertexLLMService.InputParams` is eventually deleted, their code will continue to work, "falling back" to the superclass.

markbackman: This makes sense!

src/pipecat/services/google/llm_vertex.py
```diff
@@ -75,33 +108,56 @@ def __init__(
             credentials: JSON string of service account credentials.
             credentials_path: Path to the service account JSON file.
             model: Model identifier (e.g., "google/gemini-2.0-flash-001").
-            params: Vertex AI input parameters including location and project.
+            location: GCP region for Vertex AI endpoint (e.g., "us-east4").
+            project_id: Google Cloud project ID.
             **kwargs: Additional arguments passed to OpenAILLMService.
         """
-        params = params or OpenAILLMService.InputParams()
```
kompfner, line 81: The last commit also removes this pattern. See https://github.com/pipecat-ai/pipecat/pull/2823/commits/8b62a96878da6863148ed923b2dcf1cb9d246940 for explanation.

PR #2828, opened 2025-10-10
google: rename google.gemini_live.gemini to google.gemini_live.llm

aconchillo:


aconchillo:
> LGTM.
> 
> One small observation: the GoogleVertexLLMService is `llm_vertex`. Not sure if we want to parallel that here. Not a big deal, but it would be more consistent.

I thought about that as well. Done!

PR #2833, opened 2025-10-10
`location` should not be optional when using Google Vertex.

kompfner:
Also, update `GoogleVertexLLMService` initialization pattern in the example file.


aconchillo:
LGTM

PR #2837, opened 2025-10-11
limit openai version up to 2

tzookb:

there are many packages that get blocked to upgrade because of this version lock.
ie: agents-sdk, litellm, and probably more.


markbackman:
We can bump the version, but it takes a little bit more for uv package resolution to complete. Closing this out and replacing with:
https://github.com/pipecat-ai/pipecat/pull/2859

PR #2838, opened 2025-10-11
Added Manta Graph to README

makosst:
Manta Graph can be opened through a button in README, to see the solution in a form of interactive graph. 

image


aconchillo:
LGTM. Thank you!

markbackman:
@makosst is this generated manually or automatically? I see many services are missing.

If this is manually maintained, then I fear it will get out of date very fast.

makosst:
@markbackman generated automatically, but can be tuned as well. We’ll check the services, thank you!

markbackman:
> @markbackman generated automatically, but can be tuned as well. We’ll check the services, thank you!

Thanks! To include this, I think all services should be visible and auto-generated. If that's possible, it would make it more useful to the community 🙏

makosst:
@markbackman updated the services, now grouped by function and underlying services are visible inside the nodes. Thank you!

markbackman:
Thank you 🙏

Issue #981, opened 2025-01-13
Transcript errors when using ElevenLabs Flash 2.5 TTS


chadbailey59:
A few users have reported errors in their bot transcripts that look like this:

> For example.  "Have you ever been hospitt alized" or "I understand. Does anyone in your ff amily have a history..." 

We've tracked this to an issue with ElevenLabs's Flash 2.5 model. We've reported it to them, and they're working on a fix. In the meantime, we recommend using the Turbo 2.5 model, or another provider.

We'll close this issue when it's resolved.

tarungarg546:
@chadbailey59 were we able to get anywhere from here? Did 11labs respond?

markbackman:
The ElevenLabs team acknowledged the issue and are working on it.

tarungarg546:
Is this ElevenLabs or LLM? LLM is the one who usually produces these outputs.

markbackman:
This is ElevenLabs. ElevenLabs outputs word/timestamp pairs, which we use to determine what the bot says, down to the word level. The TTS service outputs these as `TTSTextFrame`s.

markbackman:
> Is there a corresponding issue on the elevenlabs side we can monitor?

Good question. I just asked the 11Labs team. I'll post when I hear back.

larry-cmz:
Did 11Labs ever respond about this one?

markbackman:
> Did 11Labs ever respond about this one?

They acknowledge the issue again, but didn't provide a timeline for a fix...

tarungarg546:
@markbackman any timeline for this fix?

markbackman:
@tarungarg546 no timeline. The 11Labs teams is aware of the issue. I've discussed it a handful of times with them, as frequently as a few weeks ago. This is a model issue that they're working on, AFAIK.

anotine10:
i had that same issue, and i solved it by redefining the function  calculate_word_times and using the following parametrization
```
context_aggregator = llm.create_context_aggregator(
            llm_context,
            assistant_params=LLMAssistantAggregatorParams(expect_stripped_words=False),
)
````
The problem with elevenlabs, is that in the chunk that we receive, some words may be split. So we can not proceed to a strip() post process  and then add space back, because  we would add an artificial space between 2 parts of the same word

So, i monkey patch the calculate_word_times from Elevenlabs with this function
```
def calculate_word_times_patched(
    alignment_info: Mapping[str, Any], cumulative_time: float
[...]

PR #2840, opened 2025-10-12
fix: handle ElevenLabs partial word concatenation across alignment chunks gracefully

Rickaym:
ElevenLabs WebSocket API sends character alignment data in chunks that split words across multiple messages. For example, the word "the" might arrive as "th" in one chunk and "e" in the next chunk, which causes them to be incorrectly treated as two separate words and joined together as "th e".

I first noticed this problem as a voice issue, with my agent pronouncing "the" as "tho ee" and "zero" as "zoo e", I thought it was an ElevenLabs voice problem (also because the transcript I had on was not built from TTSFrames) but after putting in new logs, it turned out that texts were being joined incorrectly. This incorrect concatenation is also embedded into the chat context as visible in `pipecat.services.google.llm:_stream_content_specific_context:853 - GoogleLLMService#2: ...` poisoning the context. I hope to solve this issue and make gemini + elevenlabs pipelines work as good as the rest.

# Reproducing the problem

I was able to spot the problem in my pipeline `google:gemini-2.5-flash` -> `deepgram:nova-3` -> `elevenlabs:eleven_flash_v2:cjVigY5qzO86Huf0OWal`.

Add logging statements to `src\pipecat\services\elevenlabs\tts.py` and keep an eye on the chat context debugging statement from the Google LLM service.
```
# First chunk
'chars': ['W', 'e', "'", 'r', 'e', ' ', 'g', 'o', 'i', 'n', 'g', ' ', 't', 'o', ' ', 'u', 's', 'e', ' ', 't', 'h']

# Second chunk  
'chars': ['e', ' ', '"', 'A', 'C', 'E', ...]
```

The word "the" is split with "th" at the end of chunk 1 and "e" at the start of chunk 2.

# Changes / Fix

Modified calculate_word_times() to track partial words across chunks:
- Added partial_word and partial_word_start_time parameters to maintain state between chunks
- Returns a tuple of (word_times, new_partial_word, new_partial_word_start_time)
- Updated ElevenLabsTTSService to maintain _partial_word and _partial_word_start_time instance variables
- State is properly reset when starting new TTS utterances
Now "th" + "e" correctly produces a single word "the" with the timestamp from when "t" first appeared.

Resolves https://github.com/pipecat-ai/pipecat/issues/981

### DIFF
--- a/src/pipecat/services/elevenlabs/tts.py
+++ b/src/pipecat/services/elevenlabs/tts.py
@@ -827,6 +827,10 @@ class ElevenLabsHttpTTSService(WordTTSService):
         # Store previous text for context within a turn
         self._previous_text = ""
 
+        # Track partial words that span across alignment chunks
+        self._partial_word = ""
+        self._partial_word_start_time = 0.0
+
     def language_to_service_language(self, language: Language) -> Optional[str]:
         """Convert pipecat Language to ElevenLabs language code.
 
@@ -854,6 +858,8 @@ class ElevenLabsHttpTTSService(WordTTSService):
         self._cumulative_time = 0
         self._started = False
         self._previous_text = ""
+        self._partial_word = ""
+        self._partial_word_start_time = 0.0
         logger.debug(f"{self}: Reset internal state")
 
     async def start(self, frame: StartFrame):
@@ -888,11 +894,13 @@ class ElevenLabsHttpTTSService(WordTTSService):
     def calculate_word_times(self, alignment_info: Mapping[str, Any]) -> List[Tuple[str, float]]:
         """Calculate word timing from character alignment data.
 
+        This method handles partial words that may span across multiple alignment chunks.
+
         Args:
             alignment_info: Character timing data from ElevenLabs.
 
         Returns:
-            List of (word, timestamp) pairs.
+            List of (word, timestamp) pairs for complete words in this chunk.
 
         Example input data::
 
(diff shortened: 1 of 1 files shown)


markbackman:
Thanks for digging in to this. There's a long standing issue against ElevenLabs for this:
https://github.com/pipecat-ai/pipecat/issues/981

This is actually an ElevenLabs Flash model _bug_ (or inconsistency). The Turbo model does not have this behavior... I've been hounding the ElevenLabs team to fix this problem, but they have been non-responsive. So, it's probably time to accept that they won't resolve the model-level bug and we'll have to handle it at the application layer.

> I first noticed this problem as a voice issue, with my agent pronouncing "the" as "tho ee" and "zero" as "zoo e", I thought it was an ElevenLabs voice problem (also because the transcript I had on was not built from TTSFrames) but after putting in new logs, it turned out that texts were being joined incorrectly.

This is not actually correct. This disfluency is an issue with the model and may be related to the word splitting that we see. For example, Pipecat can pass in a sentence to ElevenLabs like this:
```
Sure! Why did the scarecrow win an award?  Because he was outstanding in his field.
```

but the alignment data returned by ElevenLabs splits the response mid-word (this is the bug!)
```
Sure! Why did the scarecrow win an award?  Because he was outst anding in his field.
```

So, the issue is completely in the model, where words and corresponding sounds are mispronounced. This is why the Turbo v2.5 model is the default in Pipecat, not Flash v2.5.

Nonetheless, I think a fix for this issue can be helpful; at a minimum, it can reduce the number of times developers flag this issue 🙈.

markbackman:
This change looks really good!

I think there are three changes to make:
1. In ElevenLabsTTSService, if an interruption were to happen at just the wrong time, it's possible that a split word may result in the `partial_word` and `partial_word_start_time` carrying over after the interruption. To avoid that, you'll need to reset the `partial_word` and `partial_word_start_time` in the `_handle_interruption` function.
2. This issue applies to ElevenLabsHttpTTSService, too. Can you make the same update there?
3. Please lint your code. You can set up Ruff as a linter or run the `scripts/fix-ruff.sh` command. You can also install the pre-commit hook to flag linting issues: `uv run pre-commit install`.

Rickaym:
Hey Mark,

I have made the appropriate changes to address your points, also did the pre-commit installation, hope the linting is fine now!
I wasn't aware that flash v2.5 was the culprit... that totally sucks, for some reason I started out with that and has always stuck with it. Time to move to turbo.

Please let me know if there's anything else needed.

Ricky.

PR #2842, opened 2025-10-13
Fix NVIDIA Riva Segmented STT by adding missing is_final parameter to _handle_transcription

nbyers-altira:
The current implementation of the NVIDIA Riva Segmented STT in Pipecat encounters a runtime error due to a mismatch in the _handle_transcription method's signature. Specifically, the method is invoked with three arguments (transcript, is_final, and language), but its definition only accepts two (transcript and language). This leads to a TypeError when processing segmented transcriptions, breaking the STT pipeline for Riva users.

This PR updates the _handle_transcription method signature to include the missing is_final: bool parameter, aligning it with its actual usage in the codebase.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,8 +38,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
-- Fixed an issue in `RivaSTTService` where a runtime error occurred due to a
-  mismatch in the _handle_transcription method's signature.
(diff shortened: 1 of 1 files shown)


markbackman (approved):

Two things:
1. Can you lint the code? You can run `scripts/fix-ruff.sh` and install the pre-commit hook in the future to catch this issue: `uv run pre-commit install`.
2. Can you add a short changelog entry for this?

```
### Fixed

- Fixed an issue in `RivaSTTService` where a runtime error occurred due to a mismatch in the _handle_transcription method's signature.
```

markbackman:
Thank you! 🙌

PR #2846, opened 2025-10-14
feat: add daily_room_properties parameter to configure() function

jamsea:
This PR adds a new optional parameter `daily_room_properties` to the `configure()` function in the Daily runner, allowing users to pass custom DailyRoomProperties instead of relying solely on the default configuration.

## Changes
- Added `daily_room_properties` parameter to `configure()` function
- When provided, custom properties override the default room configuration
- Falls back to default room properties if not provided
- Updated docstring with parameter documentation

## Benefits
- Enables more flexible room configuration for advanced use cases
- Maintains backward compatibility (parameter is optional)
- Allows users to specify custom recording settings, privacy settings, and other room properties

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Added `daily_room_properties` parameter to `configure()` function in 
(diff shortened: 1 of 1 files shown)


markbackman:
Makes sense, but I want to make a few changes:
1. Let's not add a new enable_recording property. I should have just started with room_properties to start.
2. We need to warn about mixed use.
3. We need to handle the different cases.

I've opened this PR which will accomplish that:
https://github.com/pipecat-ai/pipecat/pull/2858

PR #2848, opened 2025-10-14
feat: Add `Anannas` AI LLM service integration

Haleshot:
Hey team! Following up on this [issue](https://github.com/pipecat-ai/docs/issues/390); went with the _in-codebase_ route since Anannas fits the same pattern as OpenRouter and similar OpenAI-compatible services.

This PR adds the service class at `src/pipecat/services/anannas/llm.py`; a function-calling example adapted from the OpenRouter one; added to `pyproject.toml` as an optional dependency; updated `env.example` with the API key variable. 

Links for reference:
- [Anannas](https://anannas.ai/)
- [Anannas Quickstart](https://docs.anannas.ai/quickstart)

Will also submit a docs (`pipecat/docs`)) PR to add the `anannas` page under LLM services.

### DIFF
--- a/src/pipecat/services/anannas/llm.py
+++ b/src/pipecat/services/anannas/llm.py
@@ -1,0 +1,67 @@
+#
+# Copyright (c) 2024–2025, Daily
+#
+# SPDX-License-Identifier: BSD 2-Clause License
+#
+
+"""Anannas AI LLM service implementation.
+
+This module provides an OpenAI-compatible interface for interacting with Anannas AI's
+unified model gateway, extending the base OpenAI LLM service functionality.
+"""
+
+from typing import Optional
+
+from loguru import logger
+
+from pipecat.services.openai.llm import OpenAILLMService
+
+
+class AnannasLLMService(OpenAILLMService):
+    """A service for interacting with Anannas AI's unified model gateway.
+
+    Provides access to 500+ models from OpenAI, Anthropic, Mistral, Gemini, DeepSeek,
+    and other providers through a single OpenAI-compatible interface with built-in
+    observability and routing capabilities.
+    """
+
+    def __init__(
+        self,
+        *,
+        api_key: Optional[str] = None,
+        model: str = "gpt-4o",
+        base_url: str = "https://api.anannas.ai/v1",
+        **kwargs,
+    ):
+        """Initialize the Anannas AI LLM service.
+
+        Args:
+            api_key: The API key for accessing Anannas AI's API. If None, will attempt
+                to read from ANANNAS_API_KEY environment variable.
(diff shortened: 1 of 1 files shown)


examples/foundational/14n-function-calling-anannas.py
Haleshot, line 1: I'm curious re: the naming convention; should this be renamed to `14x-function-calling-anannas`?

markbackman:
Hi @Haleshot, Anannas should be a community integration. Do you mind moving all of this to a repo that you maintain? Are you part of the team building Anannas? If so, that's great! You can add it to your company's github which will help with reputability. If not, no worries, we can still add it.

As a reminder, see this guide for how to set up your repo and what's required to submit:

Haleshot:
> Do you mind moving all of this to a repo that you maintain?

I do not mind moving it to our org's repo. The reason I had gone with the in-codebase approach is because I felt it was similar to the existing integrations present in [`pipecat/services`](https://github.com/pipecat-ai/pipecat/tree/main/src/pipecat/services). I read through the [`Contributing.md`](https://github.com/pipecat-ai/pipecat/blob/main/COMMUNITY_INTEGRATIONS.md) file listed and after seeing the existing [documentation too](https://docs.pipecat.ai/server/services/community-integrations), felt that it (anannas) wouldn't aptly fit there (atleast not from the current categories present).

Any reason for suggesting to go with the new repo route? Just curious (hope this doesn't come across rudely). I suppose maintainability (down the line) + adding to pipecat's repo being a third party provider, etc. are some reasons?

> Are you part of the team building Anannas?

I am, yes.

Unlike the [bey integration](https://github.com/bey-dev/pipecat-bey) which was recently merged; I felt we don't fit in any of the existing categories listed (TTS, STT, LLMs, etc.) and fit better in the `pipecat/services`.

Haleshot:
We also do not have an installable `PyPi` package; we're an Openrouter alternative — a unified API to access any model. Would love to know your final thoughts on where you believe this is more suited.

markbackman:
The reason is that we simply can't keep up with the number of integrations that are launching. Once something is merged into Pipecat, it becomes our responsibility to test, maintain, and support. But, we don't want to limit developers choices when it comes to options, so we want help in maintaining integrations.

Are you not an LLM service? Based on the code you submitted, you're using the OpenAI client to provide completions. I think this is a pretty good categorization for you.

You could create a PyPI package for this integration, which would make it easier to use. This is a pretty low effort thing that would pay off in terms of ease of access.

Hope these answers help explain things.

Haleshot:
Closing since https://github.com/pipecat-ai/docs/pull/401 is a relevant PR.

Issue #2847, opened 2025-10-14
Language Option for AWS Transcribe


kennethtegrado:
### Problem Statement

I need an implementation of language detection so that I can mirror the language of the user. AWS Transcribe streaming option has this capability but right now Pipecat only implements `language-code`.

### Proposed Solution

- Add `identify-language`, `identify-multiple-languages`, and `language_options` in the query when creating the signed url
- Add these params when initiating the `AWSSTTService`.
- Add the language when creating the frame.

### Would you be willing to help implement this feature?

- [x] Yes, I'd like to contribute
- [ ] No, I'm just suggesting

PR #2849, opened 2025-10-14
Add language options in AWS Transcribe STT Service

kennethtegrado:
Changes
- Add `language_options`, `identify_language`, and `identify_multiple_languages`
- Update query params when creating signed_url
- Use `LanguageCode` as argument when creating `TranscriptionFrame` and `InterimTranscriptionFrame`

Ref #2847

### DIFF
--- a/src/pipecat/services/aws/stt.py
+++ b/src/pipecat/services/aws/stt.py
@@ -7,7 +7,14 @@
 """AWS Transcribe Speech-to-Text service implementation.
 
 This module provides a WebSocket-based connection to AWS Transcribe for real-time
-speech-to-text transcription with support for multiple languages and audio formats.
+speech-to-text transcription with support for single and multiple language identification.
+
+Features:
+
+- Single language mode: Transcribe audio in a specific language
+- Multi-language mode: Automatically identify and transcribe from multiple languages
+- Real-time language detection with confidence scores
+- Support for various audio formats and sample rates
 """
 
 import asyncio
@@ -15,7 +22,8 @@ import json
 import os
 import random
 import string
-from typing import AsyncGenerator, Optional
+import warnings
+from typing import AsyncGenerator, List, Optional
 
 from loguru import logger
 
@@ -28,7 +36,12 @@ from pipecat.frames.frames import (
     StartFrame,
     TranscriptionFrame,
 )
-from pipecat.services.aws.utils import build_event_message, decode_event, get_presigned_url
+from pipecat.services.aws.config import AWSInputParams
+from pipecat.services.aws.utils import (
+    build_event_message,
+    decode_event,
+    get_presigned_url,
+)
 from pipecat.services.stt_service import STTService
 from pipecat.transcriptions.language import Language
 from pipecat.utils.time import time_now_iso8601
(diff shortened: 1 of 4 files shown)


src/pipecat/services/aws/stt.py
```diff
@@ -7,15 +7,21 @@
 """AWS Transcribe Speech-to-Text service implementation.
 
 This module provides a WebSocket-based connection to AWS Transcribe for real-time
-speech-to-text transcription with support for multiple languages and audio formats.
+speech-to-text transcription with support for single and multiple language identification.
+
+Features:
```
markbackman, line 12: Add a newline for docstring formatting.
```suggestion
Features:

```

src/pipecat/services/aws/stt.py
```diff
@@ -40,16 +50,26 @@
...
 class AWSTranscribeSTTService(STTService):
     """AWS Transcribe Speech-to-Text service using WebSocket streaming.
 
     Provides real-time speech transcription using AWS Transcribe's streaming API.
-    Supports multiple languages, configurable sample rates, and both interim and
-    final transcription results.
+    Supports single and multi-language identification, configurable sample rates,
+    and both interim and final transcription results.
+
+    Single Language Mode:
```
markbackman: Maybe clean up:

```
    Language Modes:

    - Single: Use the `language` parameter to transcribe audio in a specific language.
    - Multi: Set `identify_multiple_languages=True` and provide `language_options` (list of 2+ languages).
        AWS will automatically detect and transcribe the spoken language. The detected language
        is included in each transcription frame via the `LanguageCode` field.
```

src/pipecat/services/aws/stt.py
```diff
@@ -61,6 +81,12 @@ def __init__(
         region: Optional[str] = "us-east-1",
         sample_rate: int = 16000,
         language: Language = Language.EN,
+        language_options: Optional[List[Language]] = None,
```
markbackman: In other classes, we use an `InputParams` class. Then, in the constructor args, you'd add a `params` arg:
```
```

I would move all new params there + move language there. This would mean deprecating the current arg positioning, which I think is fine.

See other services for how this is done.

src/pipecat/services/aws/stt.py
```diff
@@ -311,160 +424,183 @@ async def _disconnect(self):
...
+        """Build reverse language mapping at class definition time."""
+        reverse = {}
+        for lang_enum, aws_code in AWSTranscribeSTTService._LANGUAGE_MAP.items():
+            if aws_code not in reverse:
+                reverse[aws_code] = lang_enum
+        return reverse
+
+    _REVERSE_LANGUAGE_MAP = None  # Will be initialized after class definition
+
+    def service_language_to_language(self, service_language: str) -> Language | None:
```
markbackman: Move this utility function and the related ones outside of the class definition.

Nit: ideally, you could use the same utility function we have in other services. This helps with maintenance. Ex:
https://github.com/pipecat-ai/pipecat/blob/main/src/pipecat/services/elevenlabs/stt.py#L28-L150

markbackman:
Can you also install the pre-commit hook to check for linting issues?
```
uv run pre-commit install
```

You can also run the fix-ruff.sh or pre-commit.sh scripts independent from the hook.

kennethtegrado:
I'll update @markbackman based on your review thanks

markbackman:
Also, can you please make only essential changes and stick to the current code structure as best as possible. This will help make code reviewing easier and will make maintenance easier as this service will align with other STT services.

kennethtegrado:
Updated @markbackman and I tried to copy the code structure of other services as much as I can

kennethtegrado:
Hi @markbackman any updates on this PR thank you!

PR #2853, opened 2025-10-14
Adding support for trickle ice.

filipi87:
Adding support for trickle ice.

### DIFF
--- a/src/pipecat/transports/smallwebrtc/request_handler.py
+++ b/src/pipecat/transports/smallwebrtc/request_handler.py
@@ -14,6 +14,7 @@ from dataclasses import dataclass
 from enum import Enum
 from typing import Any, Awaitable, Callable, Dict, List, Optional
 
+from aiortc.sdp import candidate_from_sdp
 from fastapi import HTTPException
 from loguru import logger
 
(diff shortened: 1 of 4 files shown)


src/pipecat/transports/smallwebrtc/request_handler.py
```diff
@@ -197,6 +226,19 @@ async def handle_disconnected(webrtc_connection: SmallWebRTCConnection):
...
+        peer_connection = self._pcs_map.get(request.pc_id)
+
+        if not peer_connection:
+            raise HTTPException(status_code=404, detail="Peer connection not found")
+
+        for c in request.candidates:
+            candidate = candidate_from_sdp(c.candidate)
+            candidate.sdpMid = c.sdp_mid
+            candidate.sdpMLineIndex = c.sdp_mline_index
+            await peer_connection.add_ice_candidate(candidate)
```
kompfner, line 240: Just checking: after this there's no need to do any SDP munging/cleanup like we do when handling the initial offer?

filipi87: The SDP munging was only used for the ESP32, which I don't believe supports trickle ICE. So, I think we're good for now. We can add it in the future if needed.

PR #2854, opened 2025-10-14
CartesiaSTTService to inherit from WebsocketSTTService

aconchillo:
#### Please describe the changes in your PR. If it is addressing an issue, please reference that as well.

Just cleaning up things. This allows handling reconnections automatically.

PR #2855, opened 2025-10-14
services: added on_connected/on_disconnected events

aconchillo:

This PR adds `on_connected`/`on_disconnected` events to STT/TTS websocket-based services.

One use case:

- Deepgram is set to "multi".
- Update language from "multi" to "es", for example. Because specific languages are more accurate.
- The update requires a Deepgram disconnect/connect.
- Mute STTFilter while we are doing the update. (push STTMuteFrame)
- In the `on_connected` handler unmute. (push STTMuteFrame)

### DIFF
--- a/src/pipecat/services/cartesia/stt.py
+++ b/src/pipecat/services/cartesia/stt.py
@@ -273,6 +273,7 @@ class CartesiaSTTService(WebsocketSTTService):
             headers = {"Cartesia-Version": "2025-04-16", "X-API-Key": self._api_key}
 
             self._websocket = await websocket_connect(ws_url, additional_headers=headers)
+            await self._call_event_handler("on_connected")
         except Exception as e:
             logger.error(f"{self}: unable to connect to Cartesia: {e}")
 
@@ -285,6 +286,7 @@ class CartesiaSTTService(WebsocketSTTService):
             logger.error(f"{self} error closing websocket: {e}")
         finally:
             self._websocket = None
+            await self._call_event_handler("on_disconnected")
 
     def _get_websocket(self):
         if self._websocket:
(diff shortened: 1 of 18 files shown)


src/pipecat/services/cartesia/stt.py
```diff
@@ -242,6 +242,7 @@ async def _connect(self):
             if self._receiver_task is None or self._receiver_task.done():
                 self._receiver_task = asyncio.create_task(self._receive_messages())
             logger.debug(f"Connected to Cartesia")
+            await self._call_event_handler("on_connected")
```
markbackman, line 276: `on_disconnected` is missing for CartesiaSTTService.

markbackman:
I see a few services are missing:
- Deepgram STT
- Google STT
- Speechmatics STT

Do we want to include the Realtime LLM services? They seem out of scope of what this is shooting to solve.

src/pipecat/services/cartesia/stt.py
aconchillo: Yes, I need the other #2854

aconchillo:
> I see a few services are missing:
> 
> * Deepgram STT
> * Google STT
> * Speechmatics STT
> 

Done! Please take a look again because I have moved the events to the base STTService and TTSService class. It's possible you don't use WebsocketSTTService/WebsocketTTService and use a library instead. In this case you still want the events to be triggered.

> Do we want to include the Realtime LLM services? They seem out of scope of what this is shooting to solve.

Another PR.

src/pipecat/services/tts_service.py
```diff
@@ -626,6 +649,8 @@ def __init__(self, *, reconnect_on_error: bool = True, **kwargs):
         """
         TTSService.__init__(self, **kwargs)
         WebsocketService.__init__(self, reconnect_on_error=reconnect_on_error, **kwargs)
+        self._register_event_handler("on_connected")
```
markbackman: Why do these need to be redefined in the `WebsocketTTSService` class? Doesn't that inherit from the `TTSService` class?

aconchillo: Oh... sorry. I need to remove those.

aconchillo: Done

PR #2856, opened 2025-10-15
Fix an issue in ElevenLabsHttpTTSService where the last word is not e…

markbackman:
…mitted

There's an inconsistency between ElevenLabs' Websocket and HTTP APIs:

Websocket API outputs alignment data ending in a trailing space:
```
{'chars': [..., 't', 'o', 'd', 'a', 'y', '?', ' '], ...}
```

HTTP API lacks the trailing space:
```
{'characters': [..., 't', 'o', 'd', 'a', 'y', '?'], ...}
```

The fix in #2840 handles partial words due to the behavior of the Flash models, where words can be split across alignment data chunks. This algorithm looks for trailing spaces as a sign of complete words, which works great for the Websocket API. The HTTP API needs special handling, so after processing all chunks, we need to emit any remaining partials, which will always be the case due to the missing space.

We're already hacking around the weird behavior in ElevenLabs' Flash models, so this is just another hack to work around another inconsistency in the API.

Note: I reported this inconsistency to ElevenLabs. If they resolve this difference, we can remove this workaround. For now, we need it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- Fixed an issue in `ElevenLabsTTSService` and `ElevenLabsHttpTTSService` where
+  the Flash models would split words, resulting in a space being inserted
+  between words.
+
 - Fixed an issue where audio filters' `stop()` would not be called when using
   `CancelFrame`.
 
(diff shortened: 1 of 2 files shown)


filipi87 (approved):
LGTM. 

Only one question. if they fix the issue, would this workaround cause any side effects ?

markbackman:
> LGTM.
> 
> Only one question. if they fix the issue, would this workaround cause any side effects ?

No impact. If they fix it, no partials would exist at the end of the sequence, so nothing to push.

PR #2863, opened 2025-10-15
add 08-custom-frame-processor.py to foundational examples

vipyne:


examples/foundational/47-custom-frame-processor.py
```diff
@@ -0,0 +1,223 @@
...
+from pipecat.services.cartesia.tts import CartesiaTTSService
+from pipecat.services.deepgram.stt import DeepgramSTTService
+from pipecat.services.openai.llm import OpenAILLMService
+from pipecat.transports.base_transport import BaseTransport, TransportParams
+from pipecat.transports.daily.transport import DailyParams
+
+load_dotenv(override=True)
+
+
+class CustomFrameProcessor(FrameProcessor):
```
markbackman: Maybe we can just do one simple thing with the Frame processor? One idea is to write a MetricsLogger frame processor that console logs the MetricsFrame data.

It could be anything though. I think keeping it simple—to just one thing—will be helpful.

markbackman (commented):
Let's remove example 08 as it's old and we're probably not going to update it. This could take its spot. WDYT?

It would make sense to also update the custom FrameProcessor guide to point to this example. You could update the example in the docs to match this example.
https://docs.pipecat.ai/guides/fundamentals/custom-frame-processor

vipyne:
> It would make sense to also update the custom FrameProcessor guide to point to this example. You could update the example in the docs to match this example.

https://github.com/pipecat-ai/docs/pull/409

PR #2864, opened 2025-10-15
WhatsApp transport debug log -> trace log

vipyne:


filipi87 (approved):
Let’s keep it as a trace for now, until we find a better alternative.
Thank you @vipyne 🚀

PR #2868, opened 2025-10-15
only import whatsapp deps if using whatsapp runner

vipyne:
#### Please describe the changes in your PR. If it is addressing an issue, please reference that as well

I believe these dependencies are already imported [here](https://github.com/pipecat-ai/pipecat/blob/main/src/pipecat/runner/run.py#L201)
```
        from pipecat_ai_small_webrtc_prebuilt.frontend import SmallWebRTCPrebuiltUI

        from pipecat.transports.smallwebrtc.connection import SmallWebRTCConnection
        from pipecat.transports.smallwebrtc.request_handler import (
            SmallWebRTCRequest,
            SmallWebRTCRequestHandler,
        )
```

### DIFF
--- a/src/pipecat/runner/run.py
+++ b/src/pipecat/runner/run.py
@@ -166,6 +166,7 @@ def _create_server_app(
     host: str = "localhost",
     proxy: str,
     esp32_mode: bool = False,
+    whatsapp_enabled: bool = False,
     folder: Optional[str] = None,
 ):
     """Create FastAPI app with transport-specific routes."""
@@ -182,7 +183,8 @@ def _create_server_app(
     # Set up transport-specific routes
     if transport_type == "webrtc":
         _setup_webrtc_routes(app, esp32_mode=esp32_mode, host=host, folder=folder)
-        _setup_whatsapp_routes(app)
+        if whatsapp_enabled:
+            _setup_whatsapp_routes(app)
     elif transport_type == "daily":
         _setup_daily_routes(app)
     elif transport_type in TELEPHONY_TRANSPORTS:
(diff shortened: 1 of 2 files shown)


src/pipecat/runner/run.py
```diff
@@ -313,13 +299,28 @@ def _setup_whatsapp_routes(app: FastAPI):
...
+    WHATSAPP_APP_SECRET
+Keeping it disabled.
+            """
         )
         return
 
+    # only import WhatsApp dependencies if appropriate env vars are present
+    try:
+        from pipecat.transports.whatsapp.api import WhatsAppWebhookRequest
+        from pipecat.transports.whatsapp.client import WhatsAppClient
```
filipi87, line 326: We also need to keep this import:
`from pipecat.transports.smallwebrtc.connection import SmallWebRTCConnection`

Otherwise we are receiving this error:
```
File "/Users/filipifuchter/workspaces/workspaceDaily/rtvi/pipecat/src/pipecat/runner/run.py", line 383, in whatsapp_webhook
    async def connection_callback(connection: SmallWebRTCConnection):
                                              ^^^^^^^^^^^^^^^^^^^^^
NameError: name 'SmallWebRTCConnection' is not defined
```

```suggestion
        from pipecat.transports.smallwebrtc.connection import SmallWebRTCConnection
        from pipecat.transports.whatsapp.api import WhatsAppWebhookRequest
        from pipecat.transports.whatsapp.client import WhatsAppClient```

Comment (filipi87) [reply src/pipecat/runner/run.py:326]: The import was inside the function scope of `_setup_webrtc_routes`, so it’s only available within that function and not accessible to other functions.

File: src/pipecat/runner/run.py [hunk 2]
```diff
@@ -166,6 +166,7 @@ def _create_server_app(
     host: str = "localhost",
     proxy: str,
     esp32_mode: bool = False,
+    whatsapp_mode: bool = False,
```
filipi87: nit: I wonder if something like `whatsapp_enabled` would be better in this case, since we’re basically using it to enable or disable the WhatsApp routes.

vipyne: makes sense. it's more descriptive so I'm ok with changing it to `enabled`.

PR #2872, opened 2025-10-16
PipelineTask: fix task cancellation issues

aconchillo:

This PR fixes a couple of issues:

#### PipelineTask asyncio.CancelledError proper handling

The main issue was that `PipelineTask` was awaiting for a task (`_process_push_queue`). The way `asyncio` task cancellation works is that everything in the `await` chain gets cancelled: if it's a function the function gets cancelled, if it's a task, the task gets cancelled. This was complicating the logic and was error prone.

So, in this PR we simplify the logic and instead of waiting for a task (or a future) we just wait for an event. This way we can keep all the `PipelineTask` running and cancel them when needed.

The release evals which were leaving things in a messy state in case there was a timeout for example, now work very well.

#### PipelineTask.cancel() doesn't block anymore

We go back to Pipecat < 0.0.83 behavior where `PipelineTask.cancel()` was not blocking. Because of the simplified logic we can now do this.

### DIFF
--- a/src/pipecat/pipeline/task.py
+++ b/src/pipecat/pipeline/task.py
@@ -269,6 +269,9 @@ class PipelineTask(BasePipelineTask):
         # StopFrame) has been received at the end of the pipeline.
         self._pipeline_end_event = asyncio.Event()
 
+        # This event is set when the pipeline truly finishes.
+        self._pipeline_finished_event = asyncio.Event()
+
         # This is the final pipeline. It is composed of a source processor,
         # followed by the user pipeline, and ending with a sink processor. The
         # source allows us to receive and react to upstream frames, and the sink
@@ -401,11 +404,7 @@ class PipelineTask(BasePipelineTask):
         await self.queue_frame(EndFrame())
 
     async def cancel(self):
-        """Immediately stop the running pipeline.
-
-        Cancels all running tasks and stops frame processing without
-        waiting for completion.
-        """
+        """Request the running pipeline to cancel."""
         if not self._finished:
             await self._cancel()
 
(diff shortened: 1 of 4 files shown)


tests/test_pipeline.py
```diff
@@ -254,7 +254,7 @@ async def on_frame_reached_downstream(task, frame):
 
         try:
             await asyncio.wait_for(
-                asyncio.shield(task.run(PipelineTaskParams(loop=asyncio.get_event_loop()))),
+                task.run(PipelineTaskParams(loop=asyncio.get_event_loop())),
```
aconchillo, line 257: We also can now cleanup the unit tests and remove weird things we had to do.

src/pipecat/pipeline/task.py
```diff
@@ -417,51 +416,40 @@ async def run(self, params: PipelineTaskParams):
...
-            # task that pushes frames to the very beginning of our pipeline (our
-            # controlled source processor).
-            push_task = await self._create_tasks()
-            await push_task
+        # Setup processors.
+        await self._setup(params)
 
-            # We have already cleaned up the pipeline inside the task.
-            cleanup_pipeline = False
+        # Create all main tasks and wait of the main push task. This is the task
```
markbackman:
```suggestion
        # Create all main tasks and wait for the main push task. This is the task
```

src/pipecat/pipeline/task.py
```diff
@@ -417,51 +416,40 @@ async def run(self, params: PipelineTaskParams):
...
+        # that pushes frames to the very beginning of our pipeline (our
+        # controlled source processor).
+        await self._create_tasks()
 
+        try:
+            # Wait for pipeline to finish.
+            await self._wait_for_pipeline_finished()
             # Pipeline has finished nicely.
             self._finished = True
```
markbackman: Does this do anything now? There's another `self._finished = True` in finally, which supersedes.

markbackman (approved):
Much simpler!

src/pipecat/pipeline/task.py
aconchillo: Ah yes. We don't need this one.

PR #2873, opened 2025-10-16
Update model params for Soniox STT

matejmarinko-soniox:
- Remove deprecated parameters `enable_non_final_tokens` and `max_non_final_tokens_duration_ms`.
- Pass `enable_speaker_diarization` and `enable_language_identification` (values can be accessed on raw tokens).
- Add support for context v3 (upcoming model will have a structured context instead of a string) - https://soniox.com/docs/stt/concepts/context#context-v3

### DIFF
--- a/src/pipecat/services/soniox/stt.py
+++ b/src/pipecat/services/soniox/stt.py
@@ -49,6 +49,33 @@ END_TOKEN = "<end>"
 FINALIZED_TOKEN = "<fin>"
 
 
+class SonioxContextGeneralItem(BaseModel):
+    """Represents a key-value pair for structured general context information."""
+
+    key: str
+    value: str
+
+
+class SonioxContextTranslationTerm(BaseModel):
+    """Represents a custom translation mapping for ambiguous or domain-specific terms."""
+
+    source: str
+    target: str
+
+
+class SonioxContextObject(BaseModel):
+    """Context object for models with context_version 2, for Soniox stt-rt-v3-preview and higher.
+
+    Learn more about context in the documentation:
+    https://soniox.com/docs/stt/concepts/context
+    """
+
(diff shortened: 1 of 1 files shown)


src/pipecat/services/soniox/stt.py
```diff
@@ -60,9 +87,9 @@ class SonioxInputParams(BaseModel):
         audio_format: Audio format to use for transcription.
         num_channels: Number of channels to use for transcription.
         language_hints: List of language hints to use for transcription.
-        context: Customization for transcription.
-        enable_non_final_tokens: Whether to enable non-final tokens. If false, only final tokens will be returned.
```
markbackman, line 64: Removing these two variables will be breaking. It would be better to mark them as deprecated now and emit a warning. Then, we can remove them in the future.

They can even be a no-op, but removing them could cause some developer frustration.

markbackman: You can see examples of deprecations in docstrings or refer to the contributing guide:
https://github.com/pipecat-ai/pipecat/blob/main/CONTRIBUTING.md#code-style-and-documentation

Also, check other examples for using `warnings` to alert developers of the deprecation.

matejmarinko-soniox: Let me clarify the situation: these parameters are actually no longer supported by the official API since August. This is why I initially considered removing them. Currently using those parameters doesn't actually do anything.

Do you think it is still better to mark them as deprecated and show a warning or is removing them a better option?

We expect our newest preview model to be our flagship model (also dropping the -preview suffix) and there shouldn't be any breaking changes in the future.

markbackman: In that case, removing them should be fine.

markbackman (approved):
LGTM! Thanks for the update and sorry for the delay.

PR #2877, opened 2025-10-17
Add ellipsis character to sentence ending punctuation list

nimobeeren:
Our LLM sometimes generates the ellipsis character `…` ([U+2026](https://www.compart.com/en/unicode/U+2026)) in text. This is currently not detected as sentence-ending punctuation, which causes `PatternPairAggregator` to buffer text longer than necessary.

This PR adds `…` to the list of sentence-ending punctuation.


nimobeeren:
This is also causing issues with our voice switching, but that might be due to our custom setup.

PR #2879, opened 2025-10-17
Fix for SpeechmaticsSTTService AdditionVocabEntry entries

sam-s10s:
This fixes an issue where using AdditionalVocabEntry items without any sounds_like items would cause the session to fail.


markbackman (approved):
LGTM! Thanks 🙌

PR #2882, opened 2025-10-18
base output transport cleanup

aconchillo:

Some cleanup preparing for `BotStoppedSpeakingFrame` improvements.

### DIFF
--- a/src/pipecat/transports/base_output.py
+++ b/src/pipecat/transports/base_output.py
@@ -293,15 +293,15 @@ class BaseOutputTransport(FrameProcessor):
         """
         await super().process_frame(frame, direction)
 
-        #
-        # System frames (like InterruptionFrame) are pushed immediately. Other
-        # frames require order so they are put in the sink queue.
-        #
         if isinstance(frame, StartFrame):
             # Push StartFrame before start(), because we want StartFrame to be
             # processed by every processor before any other frame is processed.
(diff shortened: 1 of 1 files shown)


src/pipecat/transports/base_output.py
```diff
@@ -314,21 +314,6 @@ async def process_frame(self, frame: Frame, direction: FrameDirection):
             await self.write_dtmf(frame)
         elif isinstance(frame, SystemFrame):
             await self.push_frame(frame, direction)
-        # Control frames.
-        elif isinstance(frame, EndFrame):
-            await self.stop(frame)
-            # Keep pushing EndFrame down so all the pipeline stops nicely.
-            await self.push_frame(frame, direction)
-        elif isinstance(frame, MixerControlFrame):
```
markbackman, line 322: Why is this routing no longer needed?

markbackman (approved):
This clean up looks good. Just one question for my own understanding.

src/pipecat/transports/base_output.py
aconchillo: It's done in the:

```
        else:
            await self._handle_frame(frame)
```

case

aconchillo: Note how all the removed cases were calling `await self._handle_frame(frame)`

PR #2888, opened 2025-10-20
Fixed an issue where the RTVIProcessor was sending duplicate UserStartedSpeakingFrame and UserStoppedSpeakingFrame messages.

filipi87:
Fixed an issue where the RTVIProcessor was sending duplicate UserStartedSpeakingFrame and serStoppedSpeakingFrame messages.

### DIFF
--- a/src/pipecat/processors/frameworks/rtvi.py
+++ b/src/pipecat/processors/frameworks/rtvi.py
@@ -1018,6 +1018,7 @@ class RTVIObserver(BaseObserver):
 
         if (
             isinstance(frame, (UserStartedSpeakingFrame, UserStoppedSpeakingFrame))
+            and (direction == FrameDirection.DOWNSTREAM)
             and self._params.user_speaking_enabled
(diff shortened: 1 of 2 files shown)


src/pipecat/processors/frameworks/rtvi.py
```diff
@@ -1018,6 +1018,7 @@ async def on_push_frame(self, data: FramePushed):
 
         if (
             isinstance(frame, (UserStartedSpeakingFrame, UserStoppedSpeakingFrame))
+            and (direction == FrameDirection.DOWNSTREAM)
```
filipi87, line 1021: Since these frames are usually created inside the input transport, it feels more natural to expect to receive the downstream frames.

markbackman: Agreed!

markbackman (approved):
LGTM. Thanks for catching and fixing this.

PR #2889, opened 2025-10-20
Support new `LLMContext` pattern with `OpenAIRealtimeLLMService`

kompfner:


src/pipecat/services/openai_realtime/context.py
kompfner, line 1: This file be added back, with a comment mirroring the one in the context.py file at the new path.

src/pipecat/services/openai/realtime/context.py
```diff
@@ -4,7 +4,94 @@
...
+        "# Context frame type\n"
+        "frame: LLMContextFrame\n\n"
+        "# Context type\n"
+        "context: LLMContext\n\n"
+        "# Reading messages from context\n"
+        "messages = context.get_messages()\n"
+        "```\n",
+    )
 
 import copy
```
kompfner, line 87: Note that despite the big deprecation warning at the top of this file, the rest of the file remains.

I realize now that I should've done that with [the equivalent deprecation in AWS Nova Sonic](https://github.com/pipecat-ai/pipecat/pull/2750). Stay tuned for a PR adding the context.py contents back in that service, as well as a clarification to the deprecation message saying that the types are no longer used under the hood.

src/pipecat/services/openai_realtime/context.py
kompfner: Done

src/pipecat/services/openai/realtime/context.py
kompfner: Here it is: https://github.com/pipecat-ai/pipecat/pull/2896

src/pipecat/services/openai/realtime/llm.py
```diff
@@ -121,9 +119,48 @@ def __init__(
...
+            with warnings.catch_warnings():
+                warnings.simplefilter("always")
+                warnings.warn(
+                    "`send_transcription_frames` is deprecated and will be removed in a future version. "
+                    "Transcription frames are always sent.",
+                    DeprecationWarning,
+                    stacklevel=2,
+                )
+
+        # Log warning about transcription frame direction change in 0.0.92
```
kompfner: This is a bit ugly, but we introduced a breaking change that we really want folks to notice. I figure we could keep this here for at least a few versions.

Anyone have any better ideas?

src/pipecat/services/openai/realtime/llm.py
```diff
@@ -121,9 +119,48 @@ def __init__(
...
             **kwargs: Additional arguments passed to parent LLMService.
         """
+        if send_transcription_frames is not None:
+            import warnings
+
+            with warnings.catch_warnings():
+                warnings.simplefilter("always")
+                warnings.warn(
+                    "`send_transcription_frames` is deprecated and will be removed in a future version. "
+                    "Transcription frames are always sent.",
```
kompfner, line 137: This is the same deprecation that we made in AWS Nova Sonic when supporting the `LLMContext` pattern.

src/pipecat/services/openai_realtime/context.py
```diff
@@ -4,18 +4,15 @@
...
-with warnings.catch_warnings():
-    warnings.simplefilter("always")
-    warnings.warn(
-        "Types in pipecat.services.openai_realtime.context are deprecated. "
-        "Please use the equivalent types from "
-        "pipecat.services.openai.realtime.context instead.",
-        DeprecationWarning,
-        stacklevel=2,
-    )
+from pipecat.services.openai.realtime.context import *
```
kompfner, line 18: The deprecation warning will be printed during this import now

src/pipecat/services/openai_realtime/context.py
```diff
@@ -4,18 +4,15 @@
 # SPDX-License-Identifier: BSD 2-Clause License
 #
 
-"""OpenAI Realtime LLM context and aggregator implementations."""
+"""OpenAI Realtime LLM context and aggregator implementations.
 
-import warnings
+.. deprecated:: 0.0.91
```
kompfner, line 9: Even though we're changing this file in this PR, it was technically deprecated last version, hence 0.0.91.

src/pipecat/services/openai/realtime/llm.py
kompfner: Just had a better idea: maybe we could put this warning message inside of `create_context_aggregator()` instead of the initializer. That way, only people who _haven't updated their code to the new patterns_ see the message.

src/pipecat/processors/aggregators/llm_context.py
```diff
@@ -83,9 +83,17 @@ def from_openai_context(openai_context: "OpenAILLMContext") -> "LLMContext":
         Returns:
             New LLMContext instance with converted messages and settings.
         """
+        # Convert tools to ToolsSchema if needed.
```
kompfner, line 85: The advantage of this approach is that dict-format tools are only supported if the user's code is still using `OpenAILLMContext` (i.e. _unwittingly_ using `LLMContext` under the hood). If the user's code is _explicitly_ using the newer `LLMContext`, then if they attempt to use dict-format tools they will get an error telling them to use the modern `ToolsSchema`.

src/pipecat/services/openai/realtime/llm.py
```diff
@@ -653,26 +661,17 @@ async def _handle_evt_response_done(self, evt):
...
+        # We receive text deltas (as opposed to audio transcript deltas) when
+        # the output modality is "text"
         if evt.delta:
             await self.push_frame(LLMTextFrame(evt.delta))
 
     async def _handle_evt_audio_transcript_delta(self, evt):
+        # We receive audio transcript deltas (as opposed to text deltas) when
+        # the output modality is "audio" (the default)
         if evt.delta:
-            await self.push_frame(LLMTextFrame(evt.delta))
```
kompfner, line 675: Note that we no longer push an `LLMTextFrame` when the model is responding with audio, in keeping with the mental model of a speech-to-speech model being the equivalent of `[stt, llm, tts]` in the pipeline. We wouldn't expect to see `LLMTextFrame`s downstream from a TTS service.

This change is necessary in order for `OpenAIRealtimeLLMService` to play nicely with a downstream assistant context aggregator.

kompfner: Hmm...it occurs to me that this _could_ be a breaking change for some, if they were relying on processing `LLMTextFrame` between the LLM service and the assistant context aggregator...

I'll go ahead and add a warning about this.

kompfner: Done: https://github.com/pipecat-ai/pipecat/pull/2889/commits/55fc45c0be6da5ef07245cb605cf9794eed87477

src/pipecat/services/openai/realtime/llm.py
```diff
@@ -376,29 +383,33 @@ async def process_frame(self, frame: Frame, direction: FrameDirection):
...
+    async def _handle_context(self, context: LLMContext):
+        if not self._context:
+            # We got our initial context
+            self._context = context
+            # Initialize our bookkeeping of already-completed tool calls in
+            # the context
+            await self._process_completed_function_calls(send_new_results=False)
+            # Run the LLM at next opportunity
+            await self._create_response()
+        else:
```
kompfner, line 403: Note that now receiving a new context via a context frame no longer serves as a signal to reset the conversation. That’s because we’re now receiving new contexts from the user aggregator every time new messages are added, and from the assistant aggregator when function call results come in. This new code pattern of “diffing” each new context with the previous sets up for doing more sophisticated things in the future, like sending specific messages to OpenAI to modify its internal context representation.

src/pipecat/services/openai/realtime/llm.py
```diff
@@ -819,11 +861,40 @@ def create_context_aggregator(
             the user and one for the assistant, encapsulated in an
             OpenAIContextAggregatorPair.
         """
-        context.set_llm_adapter(self.get_llm_adapter())
-
-        OpenAIRealtimeLLMContext.upgrade_to_realtime(context)
-        user = OpenAIRealtimeUserContextAggregator(context, params=user_params)
+        # Log warning about transcription frame direction change in 0.0.92.
```
kompfner, line 864: Note to self: if the changes in this PR don't land in 0.0.92, make sure to update all references to 0.0.92.

src/pipecat/services/openai/realtime/llm.py
```diff
@@ -376,29 +383,33 @@ async def process_frame(self, frame: Frame, direction: FrameDirection):
             await self._handle_bot_stopped_speaking()
         elif isinstance(frame, LLMMessagesAppendFrame):
             await self._handle_messages_append(frame)
-        elif isinstance(frame, RealtimeMessagesUpdateFrame):
```
kompfner, line 379: I don't believe our previous handling of `LLMMessagesUpdateFrame` (which involved converting it to `RealtimeMessagesUpdateFrame` in a custom context aggregator and sending it here) was quite working: it never actually updated the server with the new conversation history, and it never honored the `run_llm` option.

Just noting it here to point out that this deletion doesn't represent a deprecation.

We should absolutely properly implement this functionality in an future change, though.

src/pipecat/adapters/services/aws_nova_sonic_adapter.py
```diff
@@ -210,4 +210,18 @@ def to_provider_tools_format(self, tools_schema: ToolsSchema) -> List[Dict[str,
             List of dictionaries in AWS Nova Sonic function format.
         """
         functions_schema = tools_schema.standard_tools
-        return [self._to_aws_nova_sonic_function_format(func) for func in functions_schema]
+        standard_tools = [
```
kompfner, line 213: Went ahead and implemented the same tools-shim solution for AWS Nova Sonic, in case someone has their tools defined in a non-`ToolsSchema` way and skipped the last Pipecat version (which would've broken for them).

src/pipecat/services/openai/realtime/llm.py
```diff
@@ -819,11 +861,40 @@ def create_context_aggregator(
...
+            "    llm,\n\n"
+            "    transport.output(),\n"
+            "    transcript.assistant(),\n"
+            "    context_aggregator.assistant(),\n"
+            "  ]\n"
+            ")\n\n"
+            "Also, LLMTextFrames are no longer pushed from "
+            "OpenAIRealtimeLLMService when it's configured with "
+            "output_modalities=['audio']. Listen for TTSTextFrames instead.\n\n"
+            "Once you've made the appropriate changes (if needed), you can"
```
vipyne:
```suggestion
            "Once you've made the appropriate changes (if needed), you can "
```

CHANGELOG.md
```diff
@@ -7,6 +7,85 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+  As a reminder, the context-setup pattern when using `LLMContext` is:
+
+  ```python
+  context = LLMContext(messages, tools)
+  context_aggregator = LLMContextAggregatorPair(context)
+  ```
+
+  (Note that even though `OpenAIRealtimeLLMService` now supports the universal
+  `LLMContext`, it is not meant to be swapped out for another LLM service at
+  runtime.)
```
aconchillo: This might be confusing.... maybe mention "... it is not meant to be swapped out for another LLM service at
  runtime, for example, using the `LLMSwitcher`".

examples/foundational/19-openai-realtime.py
```diff
@@ -198,6 +216,16 @@ async def on_client_connected(transport, client):
         # Kick off the conversation.
         await task.queue_frames([LLMRunFrame()])
 
+        async def set_tools_after_delay():
+            await asyncio.sleep(15)
+            new_tools = ToolsSchema(
+                standard_tools=[weather_function, restaurant_function, get_news_function]
+            )
```
aconchillo: Maybe add a comment to explain what we are doing here. Like: 

```
# Here we are just demonstrating that we can update the tool in real-time with new tools.
```

or something like that.

examples/foundational/19-openai-realtime.py
```diff
@@ -198,6 +216,16 @@ async def on_client_connected(transport, client):
...
+            logger.info("Registering new tool with LLMSetToolsFrame")
+            await task.queue_frames([LLMSetToolsFrame(tools=new_tools)])
+
+        asyncio.create_task(set_tools_after_delay())
```
aconchillo: We don't really need a task, because `on_client_connected` is already a task. So we cna jsut sleep and queue the frames. If we want a new task we should `await` it.

src/pipecat/adapters/services/open_ai_realtime_adapter.py
```diff
@@ -61,7 +75,124 @@ def get_messages_for_logging(self, context) -> List[Dict[str, Any]]:
...
+        # (or supported yet).
+        msgs = []
+        for message in self.get_messages(context):
+            msg = copy.deepcopy(message)
+            if "content" in msg:
+                if isinstance(msg["content"], list):
+                    for item in msg["content"]:
+                        if item["type"] == "image_url":
+                            if item["image_url"]["url"].startswith("data:image/"):
+                                item["image_url"]["url"] = "data:image/..."
```
aconchillo, line 90: Add support for `startswith("http")`. See this #2935

src/pipecat/processors/aggregators/llm_context.py
```diff
@@ -119,6 +126,33 @@ def messages(self) -> List[LLMContextMessage]:
...
+        """Get messages suitable for persistent storage.
+
+        NOTE: the only reason this method exists is because we're "silently"
+        switching from OpenAILLMContext to LLMContext under the hood in some
+        services and don't want to trip up users who may have been relying on
+        this method, which is part of the public API of OpenAILLMContext but
+        doesn't need to be for LLMContext.
+
+        .. deprecated::
+            Use `get_messages()` instead.
```
aconchillo, line 139: specify `0.0.92`

aconchillo:
Did a first quick look and added nits. Looks great! Approving already.

aconchillo:
Btw, when you merge, use "Squash and merge" since all the commits have same description.

src/pipecat/services/openai/realtime/llm.py
```diff
@@ -459,13 +475,20 @@ async def _ws_send(self, realtime_message):
...
+
+            # tools given in the context override the tools in the session properties
+            if llm_invocation_params["tools"]:
+                settings.tools = llm_invocation_params["tools"]
+
+            # instructions in the context come from an initial "system" message in the
+            # messages list, and override instructions in the session properties
+            if llm_invocation_params["system_instruction"]:
+                settings.instructions = llm_invocation_params["system_instruction"]
+
```
aconchillo, line 491: Maybe we should only send the client event if `settings` really changed.

src/pipecat/adapters/services/open_ai_realtime_adapter.py
kompfner: This is just for truncating messages for logging. In the case of an http url, I don't think we need to do anything.

(The above is just copy-pasted from the OpenAI adapter)

src/pipecat/processors/aggregators/llm_context.py
kompfner: I didn't specify a version here just because it's a brand-new method on `LLMContext` to facilitate transition...so it didn't feel quite right to say it was deprecated at a specific version...it's never not been deprecated...you know what I mean?

kompfner:
> Btw, when you merge, use "Squash and merge" since all the commits have same description.

They do not. Only the first line has the same description. There's detailed descriptions below.

src/pipecat/services/openai/realtime/llm.py
kompfner: That'd make sense. Since that would be a new optimization in this service which we should do kind of carefully, I think we should push it to another future PR.

kompfner:
> I know we were hesitant to do this work

Haha I sure was. Updating all these services is a lot of work and the payoff is...they're the same (knock on wood). But yes, hopefully this will make maintenance that much saner. At least it'll let us deprecate some huge code swaths.

kompfner:
OK, I believe the feedback has been addressed. Going ahead and merging.

PR #2890, opened 2025-10-20
Support arbitrary request data when using SmallWebRTC with bot runner

jptaylor:
- **Adds: Support for passing custom request data through when using SmallWebRTC via runner**
- **Fixes: client request data casing mismatches**

The `SmallWebRTCRequest` dataclass expected Python snake_case field names (`request_data`), but the Pipecat JavaScript SDK passes a camelCase property (`requestData`) due to it's natural `APIRequest` typing. This caused custom request data from JS clients to be silently ignored during deserialization. 

- Added a `from_dict()` classmethod to `SmallWebRTCRequest` that maps requestData (camelCase) to request_data (snake_case) when present, enabling proper deserialization from both naming conventions.
- Pass request data through to the bot module in the runner, which we were not currently doing.
- Updated the `RunnerArguments` base class to support `body` which should be support by all current and future transport types.

### DIFF
--- a/src/pipecat/transports/smallwebrtc/request_handler.py
+++ b/src/pipecat/transports/smallwebrtc/request_handler.py
@@ -46,6 +46,7 @@ class SmallWebRTCRequest:
             data["request_data"] = data.pop("requestData")
         return cls(**data)
 
+
 @dataclass
 class IceCandidate:
     """The remote ice candidate object received from the peer connection.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -85,6 +85,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- Runner: `body` property moved to `RunnerArguments` base class as all transports
+  should support arbitrary body data as part of a request.
+
 - `CartesiaSTTService` now inherits from `WebsocketSTTService`.
 
 - Package upgrades:
@@ -105,6 +108,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   to changes to support `LLMContext`. See "Changed" section for details.
 
 ### Fixed
+
+- Fixed an issue where the `SmallWebRTCRequest` dataclass in runner would scrub
+  arbitrary request data from client due to camelCase typing. This fixes data
+  passthrough for JS clients where `APIRequest` is used.
 
 - Fixed an issue in `RivaSegmentedSTTService` where a runtime error occurred due
   to a mismatch in the _handle_transcription method's signature.
--- a/src/pipecat/runner/run.py
+++ b/src/pipecat/runner/run.py
@@ -261,9 +261,8 @@ def _setup_webrtc_routes(
         async def webrtc_connection_callback(connection):
             bot_module = _get_bot_module()
             runner_args = SmallWebRTCRunnerArguments(
-                webrtc_connection=connection,
-                body=request.request_data
-                )
+                webrtc_connection=connection, body=request.request_data
+            )
             background_tasks.add_task(bot_module.bot, runner_args)
 
         # Delegate handling to SmallWebRTCRequestHandler
--- a/src/pipecat/services/aws/nova_sonic/llm.py
+++ b/src/pipecat/services/aws/nova_sonic/llm.py
@@ -627,7 +627,7 @@ class AWSNovaSonicLLMService(LLMService):
             else ""
         )
 
-        prompt_start = f'''
+        prompt_start = f"""
         {{
           "event": {{
             "promptStart": {{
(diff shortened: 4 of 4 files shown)


src/pipecat/runner/types.py
```diff
@@ -24,10 +24,13 @@ class RunnerArguments:
     handle_sigterm: bool = field(init=False)
     pipeline_idle_timeout_secs: int = field(init=False)
 
+    body: Optional[Any] = field(default_factory=dict)
+
     def __post_init__(self):
         self.handle_sigint = False
         self.handle_sigterm = False
         self.pipeline_idle_timeout_secs = 300
+        self.body = self.body or {}
```
filipi87, line 33: I think it makes sense to move the "body" here, since it seems it will be common to all transports and will allow receiving any custom data from the user. 

And we will need to do the same inside the base image.

Tagging @mark for a second opinion on this, since he’s worked a lot more with the runners.

jptaylor: ➕ . Also, inconsistency in transport params for common properties like this can make building multi-transport bot files tricky. For me at least, this makes sense to have as a default.

markbackman: This creates an inheritance issue, resulting in a TypeError.

The new `body` parameter is initialized in the RunnerArguments while the the subclass has a non-default required parameter `room_url`. To solve this, we would need to either:
- defer initializing `body` (e.g. init=False) (this is breaking, I think)
- Make body `kw_only=True` (Potentially breaking if args provided positionally)
- Initialize `body` in the SmallWebRTCRunnerArguments (not breaking)

Maybe this is the best outcome?
```
@dataclass
class RunnerArguments:
    handle_sigint: bool = field(init=False)
    handle_sigterm: bool = field(init=False)
    pipeline_idle_timeout_secs: int = field(init=False)
    body: dict[str, Any] = field(default_factory=dict, kw_only=True)

    def __post_init__(self):
        self.handle_sigint = False
        self.handle_sigterm = False
        self.pipeline_idle_timeout_secs = 300
```

This makes `body` a keyword-arg, but that eliminates ambiguity across transports. (It is semi-breaking though..)

Any changes made here need to be made to `pipecatcloud` types, as well.

markbackman: The least breaking option is to just add to SmallWebRTCRunnerArguments, but you keep the transport inconsistency (for positional args only).

@jptaylor what do you mean by:
> Also, inconsistency in transport params for common properties like this can make building multi-transport bot files tricky.

Is it if you're using positional args?

markbackman: FYI: @aconchillo just proposed this change, coincidentally.

jptaylor: @markbackman I can see how this is breaking. I will wait for @aconchillo to input. 

> Is it if you're using positional args?

Perhaps best to ignore input here as I'm clearly missing important context based on your comment (I didn't consider the positional args case, for example!) My point about multi-transport was that some bot methods can receive args that are either Daily _or_ SmallWebRTC, so having a tight base class definition for request data seems like a good idea. If `body` was on `RunnerArguments`, for example, I would not have had to update the definition for `SmallWebRTCRunnerArguments` as part of this PR.

markbackman: Got it. In the case of multi-transport, you can just use the `RunnerArguments` type. This already works well and is employed in a number of our examples. Two different ways this is done:
- Without create_transport factory: https://github.com/pipecat-ai/pipecat-examples/blob/main/runner-examples/02-two-transport-bot.py#L104-L151
- With create_transport factory: https://github.com/pipecat-ai/pipecat-examples/blob/main/runner-examples/04-all-transport-factory-bot.py#L141-L144

You would only use the specific type if you have a single transport type.

aconchillo:
The `body` argument to `RunnerArguments` is now added here: https://github.com/pipecat-ai/pipecat/pull/2891, it also makes all other fiels in `RunnerArguments` keyword-only.

From this PR, we still have:
- `from_dict()`
- linting
- Pass request data through to the bot module in the runner, which we were not currently doing.

src/pipecat/transports/smallwebrtc/request_handler.py
```diff
@@ -39,6 +39,13 @@ class SmallWebRTCRequest:
     restart_pc: Optional[bool] = None
     request_data: Optional[Any] = None
 
+    @classmethod
+    def from_dict(cls, data: dict):
+        """Accept both snake_case and camelCase for the request_data field."""
+        if "requestData" in data and "request_data" not in data:
+            data["request_data"] = data.pop("requestData")
+        return cls(**data)
```
aconchillo, line 47: What's the reason for allowing both options?

filipi87: I think the main reason is not break our webSDK, which is currently sending it as "requestData".

markbackman:
The majority of this functionality was added in: https://github.com/pipecat-ai/pipecat/pull/2891.

Adding the last part in https://github.com/pipecat-ai/pipecat/pull/2977.

Closing.

PR #2891, opened 2025-10-21
runner: allow starting a bot from Daily's /start endpoint

aconchillo:

This PR adds:

- `PipecatRunnerArguments` that include a `body` field.
- Allows starting a bot from `/start` following Pipecat Cloud's body format.

### DIFF
--- a/src/pipecat/runner/run.py
+++ b/src/pipecat/runner/run.py
@@ -82,6 +82,7 @@ from loguru import logger
 
 from pipecat.runner.types import (
     DailyRunnerArguments,
+    RunnerArguments,
     SmallWebRTCRunnerArguments,
     WebSocketRunnerArguments,
 )
@@ -529,9 +530,9 @@ def _setup_daily_routes(app: FastAPI):
     """Set up Daily-specific routes."""
 
     @app.get("/")
-    async def start_agent():
+    async def create_room_and_start_agent():
         """Launch a Daily bot and redirect to room."""
-        print("Starting bot with Daily transport")
+        print("Starting bot with Daily transport and redirecting to Daily room")
 
         import aiohttp
 
@@ -546,11 +547,11 @@ def _setup_daily_routes(app: FastAPI):
             asyncio.create_task(bot_module.bot(runner_args))
             return RedirectResponse(room_url)
 
-    async def _handle_rtvi_request(request: Request):
-        """Common handler for both /start and /connect endpoints.
+    @app.post("/start")
+    async def start_agent(request: Request):
+        """Handler for /start endpoints.
 
         Expects POST body like::
-
             {
                 "createDailyRoom": true,
                 "dailyRoomProperties": { "start_video_off": true },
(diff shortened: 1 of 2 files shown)


src/pipecat/runner/run.py
```diff
@@ -546,6 +547,31 @@ async def start_agent():
             asyncio.create_task(bot_module.bot(runner_args))
             return RedirectResponse(room_url)
 
+    @app.post("/start")
```
markbackman, line 550: Line 618 already has a `/start` endpoint for `_setup_daily_routes`. Make we could clean up that route to do what you'd like. I think the "_handle_rtvi_request" is too narrow. This is really just a handle the start request type of thing.

src/pipecat/runner/run.py
```diff
@@ -546,66 +547,35 @@ async def start_agent():
...
+            from pipecat.runner.daily import configure
 
-    @app.post("/start")
-    async def rtvi_start(request: Request):
-        """Launch a Daily bot and return connection info for RTVI clients."""
-        return await _handle_rtvi_request(request)
+            async with aiohttp.ClientSession() as session:
+                room_url, token = await configure(session)
+                runner_args = DailyRunnerArguments(room_url=room_url, token=token, body=body)
+                result = {"dailyRoom": room_url, "dailyToken": token}
```
markbackman: We should also return `sessionId` here. This can be any UUID that we create in the runner.

Example object returned by calling `/{agent}/start` in Pipecat Cloud:
```
{
  "dailyRoom": "https://cloud-354b1860310a444ba0f528b6a2f467c4.daily.co/qZbK9nyXCDCmQ0zDDZta",
  "dailyToken": "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9...",
  "sessionId": "639f91d8-d511-4677-a83b-bd7564d5d92f"
}
```

src/pipecat/runner/run.py
```diff
@@ -546,66 +547,35 @@ async def start_agent():
...
         print("Starting bot with Daily transport")
 
-        # Parse the request body
-        try:
-            request_data = await request.json()
-            logger.debug(f"Received request: {request_data}")
-        except Exception as e:
-            logger.error(f"Failed to parse request body: {e}")
-            request_data = {}
+        data = await request.json()
```
markbackman: Any reason to keep a try/except here?

aconchillo: I have added it back.

src/pipecat/runner/run.py
```diff
@@ -546,17 +547,9 @@ async def start_agent():
             asyncio.create_task(bot_module.bot(runner_args))
             return RedirectResponse(room_url)
 
-    async def _handle_rtvi_request(request: Request):
-        """Common handler for both /start and /connect endpoints.
-
-        Expects POST body like::
```
markbackman, line 552: Nit: It may be helpful to keep this docstring.

aconchillo: Added again.

Issue #2878, opened 2025-10-17
Claude Sonnet 4.5 on AWS Bedrock is broken


konstantin-thymia:
### pipecat version

0.0.90

### Python version

3.11

### Operating System

MacOS X

### Issue description

Claude Sonnet 4.5 now requires either `temperature` or `top_p`, but not both as per migration guide 
https://docs.claude.com/en/docs/about-claude/models/migrating-to-claude-4

>Update sampling parameters
>This is a breaking change from the Claude Sonnet 3.7.
>Use only temperature OR top_p, not both

However pipecat supplies default parameters for both arguments if they are undefined and Bedrock throws ValidationException. 

### Reproduction steps

```
import asyncio
import os
from pipecat.services.aws.llm import AWSBedrockLLMService, AWSBedrockLLMContext

async def test_claude_sonnet_4_5():
    llm = AWSBedrockLLMService(
        model="global.anthropic.claude-sonnet-4-5-20250929-v1:0",
        aws_region="eu-west-2",
        aws_access_key=os.getenv("AWS_ACCESS_KEY_ID"),
        aws_secret_key=os.getenv("AWS_SECRET_ACCESS_KEY"),
        aws_session_token=os.getenv("AWS_SESSION_TOKEN"),
        params=AWSBedrockLLMService.InputParams(
            temperature=0.7
        ),
    )
    context = AWSBedrockLLMContext(
        messages=[{"role": "user", "content": "Hello, how are you?"}]
    )
    response = await llm.run_inference(context)
    print(response)

if __name__ == "__main__":
    asyncio.run(test_claude_sonnet_4_5())
```

### Expected behavior

Generates an answer

### Actual behavior

Bedrock throws ValidationException

### Logs

```shell
asyncio.run(test_claude_sonnet_4_5())
  File "/Users/konstantin/miniconda3/envs/thymia/lib/python3.11/asyncio/runners.py", line 190, in run
    return runner.run(main)
           ^^^^^^^^^^^^^^^^
  File "/Users/konstantin/miniconda3/envs/thymia/lib/python3.11/asyncio/runners.py", line 118, in run
    return self._loop.run_until_complete(task)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/konstantin/miniconda3/envs/thymia/lib/python3.11/asyncio/base_events.py", line 653, in run_until_complete
    return future.result()
           ^^^^^^^^^^^^^^^
[...]

PR #2892, opened 2025-10-21
fix: AWSBedrockLLMService compatibility for newer Claude models

markbackman:

Resolves #https://github.com/pipecat-ai/pipecat/issues/2878.

Newer Claude models on Bedrock don't support the latency parameter, so removing from examples. They also don't support `top_p` and `temperature` being specified at the same time in the inference config.

This change removes setting defaults locally and instead relying on the API defaults. This is no breaking as no new required values are needed and we're now defaulting to the same values which were API defaults.

We're building the inference config based on set values.

### DIFF
--- a/src/pipecat/services/aws/llm.py
+++ b/src/pipecat/services/aws/llm.py
@@ -720,11 +720,11 @@ class AWSBedrockLLMService(LLMService):
             additional_model_request_fields: Additional model-specific parameters.
         """
 
-        max_tokens: Optional[int] = Field(default_factory=lambda: 4096, ge=1)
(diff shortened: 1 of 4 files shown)


src/pipecat/services/aws/llm.py
```diff
@@ -826,16 +844,16 @@ async def run_inference(self, context: LLMContext | OpenAILLMContext) -> Optiona
         model_id = self.model_name
 
         # Prepare request parameters
+        inference_config = self._build_inference_config()
+
         request_params = {
             "modelId": model_id,
             "messages": messages,
-            "inferenceConfig": {
```
markbackman, line 832: @kompfner do we need to hard code these values or can we have use the values specified via the inference config above, if provided?

PR #2896, opened 2025-10-21
Add back types that were removed, when they should only have been dep…

kompfner:
…recated

This undoes the removal in https://github.com/pipecat-ai/pipecat/pull/2750.


markbackman (approved):
This change makes sense.

Since these are no longer used, this is about avoiding an error and correctly emitting the deprecation message. Is that right?

PR #2897, opened 2025-10-21
Fixed an issue in the runner's proxy_request

filipi87:
Fixed an issue in the runner's proxy_request where a session that exists but has empty data was being treated as invalid.


markbackman (approved):
LGTM

PR #2899, opened 2025-10-21
Adding support for new bot-output RTVI Message:

mattieruth:
1. TTSTextFrames now include metadata about whether the text was spoken or not along with a type string to describe what the text represents: ex. "sentence", "word", "custom aggregation"
2. Expanded how aggregators work so that the aggregate method returns aggregated text along with the type of aggregation used to create it
3. Deprecated the RTVI bot-transcription event in lieu of...
4. Introduced support for a new bot-output event. This event is meant to be the one stop shop for communicating what the bot actually "says". It is based off TTSTextFrames to communicate both sentence by sentence (or whatever aggregation is used) as well as word by word. In addition, it will include LLMTextFrames, aggregated by sentence when tts is turned off (i.e. skip_tts is true).

Resolves pipecat-ai/pipecat-client-web#158

To help with reviewing this PR, I've written up some of the use cases we were trying to solve with these changes and how you would do so after these changes go in:

Use-Cases:

1. Identify code blocks, keeping them in the context, but not speaking them and treating them as a unique type so that they can be formatted or treated accordingly.
   
   This can be accomplished by providing a custom aggregator to the TTS service that looks for text inside a set of delimeters, along with updating the LLM's system prompt to specify that code should be demarcated with these same delimeters. The TTS service should be configured to
   skip any aggregate found with the given code type.

    ```
    tts_aggregator = PatternPairAggregator()
    tts_aggregator.add_pattern_pair(
        pattern_id="code",
        start_pattern="<code>",
        end_pattern="</code>",
        type="code",
        action=MatchAction.AGGREGATE,
    )
    tts = MyTTSService(
        ...
        text_aggregator=tts_aggregator,
        skip_aggregator_types=["code"],
    )
    ...
    context = LLMContext([
        {
            "role": "system",
            "content": "You are a friendly AI assistant.. yada yada.. If the user makes a request that results in a code snippet, then wrap the entire snippet in <code></code> blocks.",
        },
    ])
    ```

2. Identify speaker changes to change the tts voice. Speaker tags should be removed, but the speaker should be left in the context and not spoken.
   
   This is essentially the same as above, but you would add a callback for when the speaker aggregate is found:
   
   ```
   tts_aggregator.on_pattern_match("speaker", switchSpeaker)
   ```

3. Pipeline can switch between voice and text modes and no matter the mode, the context and the resulting bot-output is updated accordingly, flowing seamlessly and represents the full picture. The aggregation of the output should also be customizable in both modes, so that, for example, if you have customized the tts aggregation to treat a code block separately, code blocks can also be treated separately when the tts is skipped.
   
   This is accomplished the same as above but with the addition of the new ability to set a custom aggregator on the assistant context. The assistant context now looks for LLMTextFrames with `skip_tts` set to True and aggregates those separately. By default it uses a SimpleTextAggregator to segregate sentence-by-sentence. Bot code should look the same as above but providing a copy of the same aggregator to the assistant context:

    ```
    llm_aggregator = PatternPairAggregator()
    llm_aggregator.add_pattern_pair(
        pattern_id="code",
        start_pattern="<code>",
        end_pattern="</code>",
        type="code",
        action=MatchAction.AGGREGATE,
    )
    context = LLMContext(...)
        context=context,
        assistant_params=LLMAssistantAggregatorParams(llm_text_aggregator=llm_aggregator),
    )
    ```

4. The pipeline does not have a TTS. The context and resulting bot-output should be generated from the raw llm output and the aggregation of that output should be customizable.
   
    Maybe hacky, but you can solve this by simply setting the tts_skip flag on the llm service to true. This will cause all LLMTextFrames to be marked as tts_skip=True and thus be aggregated by the assistant context's llm_text_aggregator. See above for how to customize that aggregator.

### DIFF
--- a/tests/test_piper_tts.py
+++ b/tests/test_piper_tts.py
@@ -75,6 +75,7 @@ async def test_run_piper_tts_success(aiohttp_client):
         ]
 
         expected_returned_frames = [
+            AggregatedTextFrame,
             TTSStartedFrame,
             TTSAudioRawFrame,
             TTSAudioRawFrame,
@@ -122,7 +123,7 @@ async def test_run_piper_tts_error(aiohttp_client):
             TTSSpeakFrame(text="Error case."),
         ]
 
-        expected_down_frames = [TTSStoppedFrame, TTSTextFrame]
+        expected_down_frames = [AggregatedTextFrame, TTSStoppedFrame, TTSTextFrame]
 
         expected_up_frames = [ErrorFrame]
 


src/pipecat/services/tts_service.py
```diff
@@ -455,42 +463,53 @@ async def _process_text_frame(self, frame: TextFrame):
...
+            # Process all filter.
+            for filter in self._text_filters:
+                await filter.reset_interruption()
+                text = await filter.filter(text)
 
-        if text:
-            await self.process_generator(self.run_tts(text))
+            if text:
+                await self.push_frame(TTSTextFrame(text, spoken=True, aggregated_by=aggregated_by))
+                await self.process_generator(self.run_tts(text))
```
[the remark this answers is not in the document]
mattieruth: good point. HOWEVER. timing-wise, the goal is to get this out before the tts starts spewing the words as they are spoken. For this reason, I believe we want to go ahead and push the frame first. If the run_tts fails, we have bigger problems than our "spoken=True" flag not being right.

src/pipecat/utils/text/pattern_pair_aggregator.py
```diff
@@ -203,12 +220,16 @@ def _has_incomplete_patterns(self, text: str) -> bool:
...
+            # Again, this is written generically but there only ever should
+            # be one pattern active at a time, so the counts should be 0 or 1.
+            # Which is why we base the return on the first found.
             if start_count > end_count:
-                return True
+                start_index = text.find(start)
+                return [start_index, pattern_info["type"]]
 
-        return False
+        return None, None
```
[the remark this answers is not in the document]
mattieruth: uh. when I returned just `None`, my handling of the return value struggled and changing to None,None fixed it. i'll have to up my python game i guess 🤔

src/pipecat/frames/frames.py
```diff
@@ -351,6 +351,9 @@ class LLMTextFrame(TextFrame):
 class TTSTextFrame(TextFrame):
     """Text frame generated by Text-to-Speech services."""
 
+    aggregated_by: Literal["sentence", "word"] | str
```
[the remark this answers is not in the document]
mattieruth: good point. not sure what the default should be, though. would love any of the reviewers to weigh in here :)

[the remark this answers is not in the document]
vipyne: I vote for `"word"`
`TTSTextFrame`s of the most common TTS services tend to have a word in `.text` value.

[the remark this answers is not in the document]
mattieruth: per discussion in maintainers meeting, i think we decided it was ok NOT to have a default.

src/pipecat/utils/text/pattern_pair_aggregator.py
```diff
@@ -75,16 +92,24 @@ def __init__(self):
...
+        return Aggregation(self._text, "sentence")
 
     def add_pattern_pair(
-        self, pattern_id: str, start_pattern: str, end_pattern: str, remove_match: bool = True
+        self,
+        pattern_id: str,
+        start_pattern: str,
+        end_pattern: str,
+        type: str,
+        action: MatchAction = MatchAction.REMOVE,
```
markbackman, line 121: Removing `remove_match` is a breaking change. Maybe you can keep it, mark it as deprecated, warn when using it, and map the value to MatchAction.REMOVE?

src/pipecat/processors/aggregators/llm_response_universal.py
```diff
@@ -587,6 +591,11 @@ def __init__(
         self._function_calls_in_progress: Dict[str, Optional[FunctionCallInProgressFrame]] = {}
         self._context_updated_tasks: Set[asyncio.Task] = set()
 
+        self._llm_text_aggregator: BaseTextAggregator = (
+            self._params.llm_text_aggregator or SimpleTextAggregator()
+        )
+        self._skip_tts: Optional[bool] = None
```
markbackman: I think you can drop the `Optional` here. `Optional` is more commonly used for `__init__` or function args.
```suggestion
        self._skip_tts: bool = None
```

src/pipecat/processors/aggregators/llm_response_universal.py
```diff
@@ -798,15 +809,49 @@ async def _handle_user_image_frame(self, frame: UserImageRawFrame):
         await self.push_aggregation()
         await self.push_context_frame(FrameDirection.UPSTREAM)
 
-    async def _handle_llm_start(self, _: LLMFullResponseStartFrame):
+    async def _handle_llm_start(self, frame: LLMFullResponseStartFrame):
         self._started += 1
+        if self._skip_tts is None:
+            self._skip_tts = frame.skip_tts
+        await self._maybe_push_llm_aggregation(frame)
```
markbackman: Is this needed? The `LLMFullResponseStartFrame` doesn't contain text to aggregate. Perhaps the `LLMFullResponseStartFrame` is for initializing state for aggregation only?

src/pipecat/processors/aggregators/llm_response_universal.py
```diff
@@ -798,15 +809,49 @@ async def _handle_user_image_frame(self, frame: UserImageRawFrame):
...
         self._started -= 1
         await self.push_aggregation()
+        await self._maybe_push_llm_aggregation(frame)
+
+    async def _maybe_push_llm_aggregation(
+        self, frame: LLMFullResponseStartFrame | LLMTextFrame | LLMFullResponseEndFrame
+    ):
+        aggregate = None
+        should_reset_aggregator = False
+        if self._skip_tts and not frame.skip_tts:
```
markbackman: Of these if/else conditions, you only want one to run at any given time, right? And the ordering would be to execute with the first one encountered? If so, this might require a slightly different logic to ensure that's the case. If not, then, it's fine as is.

src/pipecat/processors/frameworks/rtvi.py
```diff
@@ -1115,14 +1140,33 @@ async def _handle_bot_speaking(self, frame: Frame):
         if message:
             await self.send_rtvi_message(message)
 
+    async def _handle_aggregated_llm_text(self, frame: AggregatedLLMTextFrame):
+        """Handle aggregated LLM text output frames."""
+        isTTS = isinstance(frame, TTSTextFrame)
+        message = RTVIBotOutputMessage(
+            data=RTVIBotOutputMessageData(
```
markbackman: Do we need to add a `bot_output_enabled` flag to turn this RTVI event on and off?

src/pipecat/processors/frameworks/rtvi.py
```diff
@@ -1115,14 +1140,33 @@ async def _handle_bot_speaking(self, frame: Frame):
...
+
     async def _handle_llm_text_frame(self, frame: LLMTextFrame):
         """Handle LLM text output frames."""
         message = RTVIBotLLMTextMessage(data=RTVITextMessageData(text=frame.text))
         await self.send_rtvi_message(message)
 
+        # TODO: Remove all this logic when we fully deprecate bot-transcription messages.
         self._bot_transcription += frame.text
-        if match_endofsentence(self._bot_transcription):
-            await self._push_bot_transcription()
```
markbackman, line 1125: I think you can remove the corresponding `_push_bot_transcription` function now that it's part of `_handle_llm_text_frame`. Not sure if it's used elsewhere, but it doesn't seem to be.

src/pipecat/services/tts_service.py
```diff
@@ -363,7 +372,7 @@ async def process_frame(self, frame: Frame, direction: FrameDirection):
         elif isinstance(frame, TTSSpeakFrame):
             # Store if we were processing text or not so we can set it back.
             processing_text = self._processing_text
-            await self._push_tts_frames(frame.text)
+            await self._push_tts_frames(frame.text, should_speak=True, aggregated_by="word")
```
markbackman: Is `word` the correct aggregated_by field here? TTSSpeakFrame is usually provided as a complete phrase, e.g.:
```
TTSSpeakFrame("Hello there, how are you?")
```

src/pipecat/services/tts_service.py
```diff
@@ -455,42 +464,62 @@ async def _process_text_frame(self, frame: TextFrame):
...
+                await filter.reset_interruption()
+                text = await filter.filter(text)
+
+            if text:
+                if not self._push_text_frames:
+                    # If we are not pushing text frames, we send a TTSTextFrame
+                    # before the audio so downstream processors know what text
+                    # is being spoken. Here, we assume this flag is used when the TTS
+                    # provider supports word timestamps and the TTSTextFrames will be
+                    # generated in the word_task_handler.
```
markbackman: Just to clarify:

```suggestion
                    # generated in the words_task_handler in the WordTTSService subclass.
```

src/pipecat/services/tts_service.py
```diff
@@ -455,42 +464,62 @@ async def _process_text_frame(self, frame: TextFrame):
...
+            await self.start_processing_metrics()
+
+            # Process all filter.
+            for filter in self._text_filters:
+                await filter.reset_interruption()
+                text = await filter.filter(text)
+
+            if text:
+                if not self._push_text_frames:
+                    # If we are not pushing text frames, we send a TTSTextFrame
```
markbackman: This is tricky. In this case, the TTSService is pushing `AggregatedLLMTextFrame`, which means that it's pushing both `AggregatedLLMTextFrame` and `TTSTextFrame`. But, the `AggregatedLLMTextFrame` is ignored in the `LLMAssistantAggregator`, meaning that two `TextFrame` subclasses are pushed, only one results in assistant messages being added to the context.

And, the `AggregatedLLMTextFrame` is pushed by the TTSService so that the RTVIObserver can see the frame in order to emit the messages to the client.

This is pretty complicated; do I have it right? We should capture this somewhere in a comment, so that it's documented.

src/pipecat/processors/aggregators/llm_response.py
```diff
@@ -59,6 +59,7 @@
     OpenAILLMContextFrame,
 )
 from pipecat.processors.frame_processor import FrameDirection, FrameProcessor
+from pipecat.utils.text.base_text_aggregator import BaseTextAggregator
```
markbackman: Does `llm_response.py` need the same changes from `llm_response_universal.py` to be backwards compatible? Or do we have a way to ensure that only the newest aggregator can use these features?

src/pipecat/services/tts_service.py
```diff
@@ -455,42 +464,62 @@ async def _process_text_frame(self, frame: TextFrame):
...
+                    frame.append_to_context = False
+                    await self.push_frame(frame)
+                await self.process_generator(self.run_tts(text))
 
-        await self.stop_processing_metrics()
+            await self.stop_processing_metrics()
 
-        if self._push_text_frames:
-            # We send the original text after the audio. This way, if we are
+        if not should_speak:
```
markbackman: I want to confirm my understanding. You can reach this case if you're using an aggregator that aggregates text that should not be spoken right? This is different than `skip_tts` being set in a TextFrame. A clarifying comment could be helpful for our future selves.

src/pipecat/services/tts_service.py
markbackman: Now I'm questioning my understanding of what this is doing. Can you explain? Why are we pushing the `AggregatedLLMTextFrame` in this case`?

src/pipecat/utils/text/pattern_pair_aggregator.py
```diff
@@ -75,16 +97,24 @@ def __init__(self):
...
+            The text that has been accumulated in the buffer.
         """
-        return self._text
+        pattern_start = self._match_start_of_pattern(self._text)
+        if pattern_start:
+            return Aggregation(self._text, pattern_start[1].get("type", "sentence"))
+        return Aggregation(self._text, "sentence")
 
     def add_pattern_pair(
-        self, pattern_id: str, start_pattern: str, end_pattern: str, remove_match: bool = True
```
markbackman, line 87: This is a breaking change. Instead, you should:
- Leave the `remove_match` arg in the `add_pattern_pair` function, marking it as deprecated
- Add a warning about the deprecation
- Use the `remove_match` to set the `REMOVE` `MatchAction`

Is that doable?

src/pipecat/processors/aggregators/llm_response_universal.py
mattieruth: You know what, you're right. I thought I needed it for the scenario where there is dangling text that might need to be sent from a previous llm output. But sense we always send whatever is left on `LLMFullResponseEndFrame`, there shouldn't ever be dangling text. I'll leave in the initialization of `self._skip_tts` though (and add a comment).

mattieruth: actually, to confirm what i said above. If you get an interrupt, do you still get an `LLMFullResponseEndFrame`?

src/pipecat/processors/aggregators/llm_response_universal.py
mattieruth: i'm not sure i follow, but there is definitely only one path for any given frame here 🤔

src/pipecat/processors/frameworks/rtvi.py
```diff
@@ -704,6 +705,29 @@ class RTVITextMessageData(BaseModel):
...
 
 
+class RTVIBotOutputMessageData(RTVITextMessageData):
+    """Data for bot output RTVI messages.
+
+    Extends RTVITextMessageData to include metadata about the output.
+    """
+
+    spoken: bool = True  # Indicates if the text has been spoken by TTS
+    aggregated_by: Optional[Literal["word", "sentence"] | str] = None
```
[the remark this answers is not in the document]
mattieruth: while it may be redundant, it makes it clear that there's a set of options that can definitely be expected.

src/pipecat/processors/frameworks/rtvi.py
mattieruth: ah, probably

src/pipecat/processors/aggregators/llm_response.py
mattieruth: not sure i follow...

src/pipecat/processors/frameworks/rtvi.py
mattieruth: actually... i don't know. this is more of a replacement for bot-transcription which didn't have a flag 🤔 -- what do you think?

mattieruth: oh actually, it looks like it was behind the bot_tts_enabled flag (weird). and everything else has a flag. i'll add it.

src/pipecat/processors/frameworks/rtvi.py
mattieruth: good catch

src/pipecat/services/tts_service.py
mattieruth: ah! makes sense. not sure why i (or maybe copilot) put "word". Changing to "sentence"

src/pipecat/services/tts_service.py
mattieruth: correct, this essentially maps to text that was aggregated into a a type listed in the new `_skip_aggregator_types` list. but i think i can do better than a comment...

src/pipecat/utils/text/pattern_pair_aggregator.py
mattieruth: yes. and no. the issue is that `type` doesn't have a default, so it has to be listed in the args first. And that in and of itself is breaking. 

That leads to a follow-up question of whether "type" is required or if we can just use the existing "pattern_id", but i sure like the name "type" better for how it's used...

src/pipecat/processors/aggregators/llm_response_universal.py
mattieruth: i guess even then, the next text frame will have the same affect of flipping and sending any dangling text... all that to say: you're right. we don't have to maybe push here :)

src/pipecat/services/tts_service.py
mattieruth: ok. i've updated the comment here. let me know if that's good enough or still confusing.

src/pipecat/utils/text/pattern_pair_aggregator.py
mattieruth: Per an offline chat, there is not great way to retain backwards compatibility without introducing ugly API. So instead, I leaned into the breakage and not only kept `remove_match` as a removed field, but replaced `pattern_id` with the new `type` -- In practice, these essentially are the same thing; a way to identify the pattern type. The key difference of thinking of it less as a unique id vs identifying the conceptual way the pattern will aggregate text (which should still be unique)

src/pipecat/processors/frameworks/rtvi.py
mattieruth: done.

src/pipecat/processors/aggregators/llm_response.py
mattieruth: i think this is the last comment I have not addressed. happy to chat about it so i can understand.

mattieruth: per an offline discussion... no. we are only going to support the feature of being able to generate bot-output when skip_tts is True in the new universal context.

src/pipecat/frames/frames.py
```diff
@@ -348,7 +351,21 @@ class LLMTextFrame(TextFrame):
...
+
+    Parameters:
+        aggregated_by: Method used to aggregate the text frames.
+    """
+
+    aggregated_by: Literal["sentence", "word"] | str
+
+
+@dataclass
+class TTSTextFrame(AggregatedLLMTextFrame):
```
kompfner: Technically you could use a TTS service without an LLM service. In which case it might just be ingesting `TextFrame`s and not `LLMTextFrame`s. In which case...should this base class be called `AggregatedTextFrame`?

src/pipecat/utils/text/pattern_pair_aggregator.py
kompfner: TBH, `type` confuses me a bit, but `pattern_id` makes sense. In the docstring description of it, it's even described as "Identifier for this pattern pair".

Seems like if the rename doesn't make things obviously clearer, it'd be preferable to default to breaking fewer things?

src/pipecat/utils/text/pattern_pair_aggregator.py
kompfner: Is there a reason we can't go with this approach?

src/pipecat/utils/text/pattern_pair_aggregator.py
```diff
@@ -93,41 +120,54 @@ def add_pattern_pair(
...
-            pattern_id: Unique identifier for this pattern pair.
+            type: Identifier for this pattern pair. Should be unique and ideally descriptive.
+                  (e.g., 'code', 'speaker', 'custom'). type can not be 'sentence' as that is
+                  reserved for the default behavior.
             start_pattern: Pattern that marks the beginning of content.
             end_pattern: Pattern that marks the end of content.
-            remove_match: Whether to remove the matched content from the text.
+            action: What to do when a complete pattern is matched:
+                    - MatchAction.REMOVE: Remove the matched pattern from the text.
+                    - MatchAction.KEEP: Keep the matched pattern in the text and treat it as
```
kompfner, line 137: 👍

src/pipecat/utils/text/pattern_pair_aggregator.py
```diff
@@ -227,24 +271,42 @@ async def aggregate(self, text: str) -> Optional[str]:
...
+
+        if len(patterns) > 0:
+            if len(patterns) > 1:
+                logger.warning(
+                    f"Multiple patterns matched: {[p.type for p in patterns]}. Only the first pattern will be returned."
+                )
+            # If the pattern found is set to be aggregated, return it
+            action = self._patterns[patterns[0].type].get("action", MatchAction.REMOVE)
+            if action == MatchAction.AGGREGATE:
+                self._text = ""
```
kompfner, line 333: Wait so if we had so far buffered something like "Here's some code \<code\>foo = 3\</code\>" then we'd only return "\<code\>foo = 3\</code\>" and lose "Here's some code"?

src/pipecat/utils/text/pattern_pair_aggregator.py
```diff
@@ -8,19 +8,34 @@
...
 
 
-class PatternMatch:
+class MatchAction(Enum):
+    """Actions to take when a pattern pair is matched.
+
+    Parameters:
+        REMOVE: Remove the matched pattern from the text.
+        KEEP: Keep the matched pattern in the text as normal text.
+        AGGREGATE: Return the matched pattern as a separate aggregation object.
```
kompfner: This is a really tricky concept. Might make sense to elaborate what "separate aggregation" means?

src/pipecat/utils/text/pattern_pair_aggregator.py
kompfner: Or does it treat the start of a match as the end of a regular sentence, so my example should be impossible?

kompfner: Ah OK I think I see that logic below.

src/pipecat/services/tts_service.py
```diff
@@ -483,14 +502,29 @@ async def _push_tts_frames(self, text: str):
             text = await filter.filter(text)
 
         if text:
+            if not self._push_text_frames:
+                # In a typical pipeline, there is an assistant context aggregator
```
kompfner: Thanks for the thorough explanation here! 👏 

Still hurts my brain, though.

src/pipecat/services/openai/realtime/llm.py
```diff
@@ -672,7 +672,7 @@ async def _handle_evt_audio_transcript_delta(self, evt):
         # We receive audio transcript deltas (as opposed to text deltas) when
         # the output modality is "audio" (the default)
         if evt.delta:
-            await self.push_frame(TTSTextFrame(evt.delta))
+            await self.push_frame(TTSTextFrame(evt.delta, aggregated_by="sentence"))
```
kompfner: Hm, I could've sworn that either OpenAI Realtime or Gemini Live provided output in chunks smaller than sentences...

kompfner: (But maybe I'm thinking of user transcriptions)

CHANGELOG.md
```diff
@@ -39,22 +44,72 @@ reason")`.
...
+            text = match.text # instead of match.content
+         ```
+  - `PatternPairAggregator` now supports `type` and `action` per pattern.
+  - New `MatchAction` enum: `REMOVE`, `KEEP`, `AGGREGATE`, allowing customization for how
+    a match should be handled.
+    - `REMOVE`: The text along with its delimiters will be removed from the streaming text.
+                Sentence aggregation will continue on as if this text did not exist.
+    - `KEEP`: The delimiters will be removed, but the content between them will be kept.
+              Sentence aggregation will continue on with the internal text included.
+    - `AGGREGATE`: The delimiters will be removed and the content between will be treated
```
kompfner: great description!

CHANGELOG.md
```diff
@@ -39,22 +44,72 @@ reason")`.
...
+              as a separate aggregation. Any text before the start of the pattern will be
+              returned early, whether or not a complete sentence was found. Then the pattern
+              will be returned. Then the aggregation will continue on sentence matching after
+              the closing delimeter is found. The content between the delimeters is not
+              aggregated by sentence. It is aggregated as one single block of text.
+    - `PatternMatch` now extends `Aggregation` and provides richer info to handlers.
+
+- Added support for aggregating `LLMTextFrame`s from within the assistant `LLMAssistantAggregator`
+  when `skip_tts` is set to `True`, generating `AggregatedLLMTextFrame`s, therefore supporting
+  `bot-output` even when TTS is turned off. You can customize the aggregator used using the new
```
```suggestion
  the new RTVI `bot-output` event even when TTS is turned off. You can customize the aggregator used using the new
```

(as a way to gently remind the reader what 'bot-output' is here)

CHANGELOG.md
```diff
@@ -39,22 +44,72 @@ reason")`.
...
+  `llm_text_aggregator` field in the `LLMAssistantAggregatorParams`. NOTE: This feature is only
+  supported when using the new universal context.
```
```suggestion
  supported when using the new `LLMContext`.
```

CHANGELOG.md
```diff
@@ -39,22 +44,72 @@ reason")`.
...
+  field `aggregated_by` to represent the conceptual format by which the given text
+  is aggregated. `TTSTextFrame`s now inherit from `AggregatedLLMTextFrame`.
+
+- New `bot-output` RTVI message to represent what the bot actually "says".
+  - The `RTVIObserver` now emits `bot-output` messages based off the new `AggregatedLLMTextFrame`s
+    (`bot-tts-text` and `bot-llm-text` are still supported and generated, but `bot-transcript` is
+    now deprecated in lieu of this new, more thorough, message).
+  - The new `RTVIBotOutputMessage` includes the fields:
+    - `spoken`: A boolean indicating whether the text was spoken by TTS
+    - `aggregated_by`: A string representing how the text was aggregated ("sentence", "word",
```
kompfner, line 42: Actually the literal "custom" or "\<my custom aggregation\>", like in `Aggregation.type`? I know I can read on to get the answer, but just making a note here.

CHANGELOG.md
```diff
@@ -65,29 +120,36 @@ reason")`.
...
-  warning instead of returning None. This allows developers to use newly
-  supported languages before Pipecat's service classes are updated, while still
-  providing guidance on verified languages.
+- `TextFrame` new field `append_to_context` used to indicate if the encompassing
+  text should be added to the LLM context (by the LLM assistant aggregator). It
+  defaults to `True`.
+
+- TTS flow respects aggregation metadata
+  - `TTSService` accepts a new `skip_aggregator_types` to avoid speaking certain aggregation types
+    (asnow determined/returned by the aggregator)
```
```suggestion
    (now determined/returned by the aggregator)
```

kompfner: or

```suggestion
    (as now determined/returned by the aggregator)
```

CHANGELOG.md
kompfner: Ah, got it. It's the latter. So maybe here use "my custom aggregation", like you do elsewhere to indicate that the string is the developer-provided aggregation type string

src/pipecat/processors/aggregators/llm_response_universal.py
```diff
@@ -610,6 +619,8 @@ async def process_frame(self, frame: Frame, direction: FrameDirection):
             await self.push_frame(frame, direction)
         elif isinstance(frame, LLMFullResponseStartFrame):
             await self._handle_llm_start(frame)
+        elif isinstance(frame, LLMTextFrame):
```
kompfner: Maybe worth a comment that this _must_ appear in this `if/elif` chain above `TextFrame` handling.

src/pipecat/processors/aggregators/llm_response_universal.py
```diff
@@ -798,15 +809,50 @@ async def _handle_user_image_frame(self, frame: UserImageRawFrame):
...
+        should_reset_aggregator = False
+        if self._skip_tts and not frame.skip_tts:
+            # When skip_tts transitions to False, we need to push any accumulated text.
+            # This ensures that any remaining text accumulated while TTS was skipped is
+            # sent out when TTS resumes, preventing loss of data and maintaining a smooth
+            # transition.
+            aggregate = self._llm_text_aggregator.text
+            should_reset_aggregator = True
+        self._skip_tts = frame.skip_tts
+        if self._skip_tts:
```
kompfner: For my understanding: if we're in a state where TTS is being bypassed, we're doing the aggregation work here (into sentences, etc) that we would otherwise have done in the TTS service, due to the fact that in our architecture the TTS service is usually responsible for that kind of aggregation.

Is that understanding right?

kompfner: If so: I really really really wish we had landed on some other name for the work of aggregating text into sentences, etc (like TTS does) and aggregating _those_ text chunks into context messages (the primary job of the assistant aggregator).

Calling them both "aggregation" hurts the reader's brain.

kompfner: But maybe this is just the world we're in. Too many things have "aggregator" in their names already and we don't want to unleash an avalanche of renames...

kompfner: Given that, maybe this file could use a little more explanatory commenting to provide an overview of this mechanism and make distinction clear between the two kinds of aggregation at play here.

src/pipecat/processors/frameworks/rtvi.py
```diff
@@ -1437,7 +1477,7 @@ async def process_frame(self, frame: Frame, direction: FrameDirection):
         elif isinstance(frame, RTVIActionFrame):
             await self._action_queue.put(frame)
         elif isinstance(frame, LLMConfigureOutputFrame):
```
kompfner, line 1549: Here we rely on `LLMConfigureOutputFrame` to determine if the pipeline is skipping TTS, but elsewhere (the assistant aggregator, I believe?) we were just inspecting text frames as they were coming in and determining on the fly, based on their `skip_tts` values, whether TTS was being skipped. Why the different approaches?

src/pipecat/utils/text/pattern_pair_aggregator.py
kompfner: Ohh is it because it's not strictly a keyword argument, it's positional?

kompfner:
> Maybe hacky, but you can solve this by simply setting the tts_skip flag on the llm service to true.

This does seem like a tough thing for someone who's building a text-output-only pipeline to have to discover and figure out how to do...`skip_tts` would be an unintuitive thing to set since: a) there's no TTS service in the pipeline anyway, and b) its effect isn't to skip TTS but something unrelated, which is to make `bot-output` work properly...right?

kompfner:
Oh just thought of one more thing: is there an onus on the developer to pass the _same_ text aggregator to both their assistant aggregator and TTS service, or risk behavior changing when toggling between TTS and no-TTS modes?

Is this kind of misconfiguration something we could or should protect against or help the developer get right?

src/pipecat/frames/frames.py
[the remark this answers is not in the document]
aconchillo: Yes, I think that's fine. We want to force the developer specify what type of aggregation is this frame.

src/pipecat/frames/frames.py
aconchillo: Yeah, I think `AggregatedTextFrame` makes sense as it could be used by something that is not an LLM.

src/pipecat/frames/frames.py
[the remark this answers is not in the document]
aconchillo: Instead of `Literal["sentence", "word"]` should we have an enum? 

```
class AggregationType(Enum):
    SENTENCE = "sentence"
    WORD = "word"
```

src/pipecat/frames/frames.py
mattieruth: good idea!

src/pipecat/processors/aggregators/llm_response_universal.py
mattieruth: first of all: your understanding is correct

second: i agree, it's confusing. unfortunately, from various discussions there's really not one single good place to say "here is where in the pipeline we should aggregate". different things want different aggregations at different times for different reasons. it was a conscious decision to have the aggregators not be processors, but rather simple utilities that can be used in any given processor. so i think we're stuck in a world with overlapping aggregators. I'll take a stab at a better name and comments...

mattieruth: oh. semi-related but important to note: i didn't add it for this PR, but i imagined solving filipi's use case of "what if there is no TTS in the pipeline" with the changes in this file. I figured it should be as easy as adding a new flag that's like "no_tts" that would set to turn this aggregator on for all LLMTextFrames.

though, TBH you can also do this today with this PR by simply setting the `skip_tts` flag to true on the llm and leaving it.

CHANGELOG.md
filipi87, line 1: It would be nice to rebase with the latest changes from main so that we can more easily analyze only what you are including here, just to make sure we’re not missing anything.

src/pipecat/frames/frames.py
```diff
@@ -337,11 +338,14 @@ class TextFrame(DataFrame):
     # mandatory fields of theirs to have defaults to preserve
     # non-default-before-default argument order)
     includes_inter_frame_spaces: bool = field(init=False)
+    # Whether this text frame should be appended to the LLM context.
+    append_to_context: bool = field(init=False)
```
filipi87, line 342: Shouldn't this already be `True` here ? It feels that it would be easier to understand, otherwise we need to check inside `__post_init__` to understand that the default behavior it is actually to `True`.

src/pipecat/processors/aggregators/llm_text_processor.py
```diff
@@ -0,0 +1,102 @@
+#
+# Copyright (c) 2024–2025, Daily
+#
+# SPDX-License-Identifier: BSD 2-Clause License
+#
+
+"""LLM text processor module for processing and aggregating raw LLM output text.
```
filipi87, line 7: 🎉

src/pipecat/processors/aggregators/llm_text_processor.py
```diff
@@ -0,0 +1,102 @@
...
+
+    def __init__(self, *, text_aggregator: Optional[BaseTextAggregator] = None, **kwargs):
+        """Initialize the LLM text processor.
+
+        Args:
+            text_aggregator: An optional text aggregator to use for processing LLM text frames. By
+                default, a SimpleTextAggregator aggregating by sentence will be used.
+            **kwargs: Additional arguments passed to parent class.
+
+        TODO: Allow transformations per aggregation type or all (and deprecate the TTS filters).
```
filipi87, line 49: I guess we are going to add this in a follow up PR, right ?

src/pipecat/processors/aggregators/llm_text_processor.py
```diff
@@ -0,0 +1,102 @@
...
+    async def _handle_interruption(self, _):
+        """Handle interruptions by resetting the text aggregator."""
+        await self._text_aggregator.handle_interruption()
+
+    async def reset(self):
+        """Reset the internal state of the text processor and its aggregator."""
+        await self._text_aggregator.reset()
+
+    async def _handle_llm_text(self, in_frame: LLMTextFrame):
+        aggregation = await self._text_aggregator.aggregate(in_frame.text)
```
filipi87, line 86: I see that the `aggregate` method now returns an `Aggregate` object instead of just a `string`. 

So, I believe this will be a breaking change. Is this something we should be concerned about or try to avoid to maintain compatibility ?

src/pipecat/processors/aggregators/llm_text_processor.py
```diff
@@ -0,0 +1,102 @@
...
+        if aggregation:
+            out_frame = AggregatedTextFrame(
+                text=aggregation.text,
+                aggregated_by=aggregation.type,
+            )
+            out_frame.skip_tts = in_frame.skip_tts
```
filipi87, line 92: In this case we are assuming that `skip_tts` has the same value as in the last frame that resulted in the aggregation.

But what if the previous frames, which are also part of this text, have different `skip_tts` values ? Which one should we respect ?

src/pipecat/processors/frameworks/rtvi.py
```diff
@@ -907,9 +934,13 @@ class RTVIObserverParams:
         metrics_enabled: Indicates if metrics messages should be sent.
         system_logs_enabled: Indicates if system logs should be sent.
         errors_enabled: [Deprecated] Indicates if errors messages should be sent.
+        skip_aggregator_types: List of aggregation types to skip sending as tts/output messages.
+          Note: if using this to avoid sending secure information, be sure to also disable
+                bot_llm_enabled to avoid leaking through LLM messages.
```
filipi87, line 939: Interesting. Makes a lot of sense. 👏

src/pipecat/processors/frameworks/rtvi.py
```diff
@@ -704,6 +707,29 @@ class RTVITextMessageData(BaseModel):
     text: str
 
 
+class RTVIBotOutputMessageData(RTVITextMessageData):
+    """Data for bot output RTVI messages.
+
+    Extends RTVITextMessageData to include metadata about the output.
+    """
+
+    spoken: bool = True  # Indicates if the text has been spoken by TTS
```
filipi87: Shouldn't this be `False` by default ?

src/pipecat/processors/frameworks/rtvi.py
```diff
@@ -973,8 +1005,26 @@ def __init__(
...
                 )
 
+        self._aggregation_transforms: List[Tuple[str, Callable[[str, str], Awaitable[str]]]] = []
+
+    def transform_aggregation_type(
+        self, aggregation_type: str, transform_function: Callable[[str, str], Awaitable[str]]
+    ):
+        """Transform text for a specific aggregation type before sending as Bot Output or TTS.
+
+        # TODO: What if someone wanted to remove a registered transform?
```
filipi87: Shouldn't this be two separate methods, one to add and one to remove ? For example:
- `add_aggregation_transformer`
- `remove_aggregation_transformer`

Or something like this ?

src/pipecat/services/cartesia/tts.py
```diff
@@ -182,6 +249,10 @@ def __init__(
             container: Audio container format.
             params: Additional input parameters for voice customization.
             text_aggregator: Custom text aggregator for processing input text.
+
+                .. deprecated:: 0.0.95
+                    Use an LLMTextProcessor before the TTSService for custom text aggregation.
```
filipi87, line 254: 🔥

src/pipecat/services/tts_service.py
```diff
@@ -281,6 +305,22 @@ async def cancel(self, frame: CancelFrame):
             await self.cancel_task(self._stop_frame_task)
             self._stop_frame_task = None
 
+    def transform_aggregation_type(
+        self, aggregation_type: str, transform_function: Callable[[str, str], Awaitable[str]]
+    ):
+        """Transform text for a specific aggregation type.
+
+        # TODO: What if someone wanted to remove a registered transform?
```
filipi87: Same comment from RTVI. Shouldn't this be two separate methods, one to add and one to remove ? For example:
- `add_aggregation_transformer`
- `remove_aggregation_transformer`

src/pipecat/services/cartesia/tts.py
```diff
@@ -257,6 +336,27 @@ def language_to_service_language(self, language: Language) -> Optional[str]:
...
+        """Convenience method to create a pause tag."""
+        return f'<break time="{seconds}s" />'
+
+    def VOLUME_TAG(volume: float) -> str:
+        """Convenience method to create a volume tag."""
+        return f'<volume ratio="{volume}" />'
+
+    def SPEED_TAG(speed: float) -> str:
+        """Convenience method to create a speed tag."""
+        return f'<speed ratio="{speed}" />'
```
filipi87, line 358: Shouldn't we group this inside a class outside Cartesia, to make it easy to understand which are only helper functions ? Something like this:

```
class CartesiaTransformers:
    """Convenience helpers for Cartesia text transformations."""

    @staticmethod
    def SPELL(text: str) -> str:
        return f"<spell>{text}</spell>"

    @staticmethod
    def EMOTION_TAG(emotion: CartesiaEmotion) -> str:
        return f'<emotion value="{emotion}" />'

    @staticmethod
    def PAUSE_TAG(seconds: float) -> str:
        return f'<break time="{seconds}s" />'

    @staticmethod
    def VOLUME_TAG(volume: float) -> str:
        return f'<volume ratio="{volume}" />'

    @staticmethod
    def SPEED_TAG(speed: float) -> str:
        return f'<speed ratio="{speed}" />'
```

src/pipecat/services/rime/tts.py
```diff
@@ -180,6 +192,31 @@ async def set_model(self, model: str):
...
+        self._extra_msg_fields["phonemizeBetweenBrackets"] = True
+        return text.replace(word, f"{phoneme}")
+
+    def INLINE_SPEED(self, text: str, speed: float) -> str:
+        """Convenience method to support inline speeds."""
+        if not self._extra_msg_fields:
+            self._extra_msg_fields = {}
+        speed_vals = self._extra_msg_fields.get("inlineSpeedAlpha", "").split(",")
+        self._extra_msg_fields["inlineSpeedAlpha"] = ",".join(speed_vals + [str(speed)])
+        return f"[{text}]"
```
filipi87, line 219: Same comment as I did to Cartesia, group it inside a class, maybe `RimeTransformers`.

src/pipecat/services/rime/tts.py
```diff
@@ -180,6 +192,31 @@ async def set_model(self, model: str):
...
+    def PAUSE_TAG(seconds: float) -> str:
+        """Convenience method to create a pause tag."""
+        return f"<{seconds * 1000}>"
+
+    def PRONOUNCE(self, text: str, word: str, phoneme: str) -> str:
+        """Convenience method to support Rime's custom pronunciations feature.
+
+        https://docs.rime.ai/api-reference/custom-pronunciation
+        """
+        self._extra_msg_fields["phonemizeBetweenBrackets"] = True
```
filipi87, line 210: It does feel a little bit weird to set a global field from the class here, since we’d need to reset it later. 

This would also prevent us from being able to extract it to a different class, like `RimeTransformers`, which shouldn't be aware of this logic.

But I’m also not sure if the alternatives would be any better, like returning the JSON directly from here, or using some placeholder word that we’d need to parse when building the text later. 🤕

src/pipecat/services/tts_service.py
```diff
@@ -455,13 +497,24 @@ async def _process_text_frame(self, frame: TextFrame):
         text: Optional[str] = None
         if not self._aggregate_sentences:
             text = frame.text
+            aggregated_by = "token"
```
filipi87, line 549: Just to understand, why is the aggregated_by here "token", and not AggregationType.WORD ?

src/pipecat/services/tts_service.py
filipi87: Yep, that was great.

CHANGELOG.md
mattieruth: i rebased yesterday just before pushing and fixing the CHANGELOG. Is it already way out of date?

mattieruth: oh i see. there is stuff mixed in below that's not mine. i must have botched the rebase. the conflicts in this file were a bear.

filipi87: Interesting, it is showing a couple of things that I believe are not related with this PR. For example:

```
### Added

- Added an asyncio event `finished_event` field to `InterruptionFrame`. When
  assigned, the asyncio event will be set when the frame reaches the end of the
  pipeline. You can use this field to know when an interruption made it all the
  way to the end of a pipeline. The field has been also added to
  `InterruptionTaskFrame`.

- Added support for loading external observers. You can now register custom
  pipeline observers by setting the `PIPECAT_OBSERVER_FILES` environment
  variable. This variable should contain a colon-separated list of Python files
  (e.g. `export PIPECAT_OBSERVER_FILES="observer1.py:observer2.py:..."`). Each
  file must define a function with the following signature:
```

But those things I believe are already more than one week old.

mattieruth: yeah. i think the rebase went bad after the release. i just re-rebased and cleaned it up. I've got to get this in because these rebases are rough.

src/pipecat/frames/frames.py
mattieruth: I struggled to understand what `field()` does, but I remember in practice that the only way I could get this to compile and work was to set it as a `field()` with `init=False`. I walked away assuming that `init=False` is not saying that the default value is `False` but that it needed to call `init`??? -- happy to update, if you know how it's supposed to be different, but like I said, this was the only way i got it working.

src/pipecat/processors/aggregators/llm_text_processor.py
mattieruth: yes.

src/pipecat/processors/aggregators/llm_text_processor.py
mattieruth: yes, it's breaking. I don't know a way around this one. I'm hoping not too many people have defined their own Aggregator types??

src/pipecat/processors/aggregators/llm_text_processor.py
mattieruth: hmm. you're right. In previous incarnations of this, where i had an aggregator in rtvi or the assistantcontext, I would check for the flag switching and any time it switched, i would assentially early-terminate the aggregation and send it. perhaps i should re-introduce that.

mattieruth: of course, if we could assume people now use `skip_tts` as a mode shift and all frames between start/end have the same value, that would make this much nicer :)

mattieruth: before i add confusing code to handle intermittent skips/non-skips, i'm inviting @aconchillo and @kompfner to weigh in

src/pipecat/processors/frameworks/rtvi.py
mattieruth: ah, good catch. I made this class early on in the process when True was a better default.

mattieruth: I think `aggregated_by` below is no longer Optional either :)

src/pipecat/processors/frameworks/rtvi.py
mattieruth: yeah, i like that. i think my head was done figuring out names by the time i got to this and just stuffed a TODO in there 😂

src/pipecat/processors/aggregators/llm_response.py
```diff
@@ -95,6 +96,7 @@ class LLMAssistantAggregatorParams:
     """
 
     expect_stripped_words: bool = True
+    llm_text_aggregator: Optional[BaseTextAggregator] = None
```
kompfner: Do we want to print a warning when the application code provides one of these to the (older, soon-to-be-deprecated) `LLMAssistantContextAggregator` in this file?

Like a "hey, this is only supported if you use the new `LLMContextAggregatorPair`?

kompfner: Oh wait...don't we want this argument to go away altogether, now, after your latest changes? We no longer need this "backup" text aggregator for the case where there's no TTS right?

src/pipecat/processors/aggregators/llm_text_processor.py
```diff
@@ -0,0 +1,102 @@
...
+)
+from pipecat.processors.frame_processor import FrameDirection, FrameProcessor
+from pipecat.utils.text.base_text_aggregator import BaseTextAggregator
+from pipecat.utils.text.simple_text_aggregator import SimpleTextAggregator
+
+
+class LLMTextProcessor(FrameProcessor):
+    """A processor for handling or manipulating LLM text frames before they are processed further.
+
+    This processor can be used to pre-aggregate, modify, or filter direct output tokens from the LLM
```
kompfner: Nice description!

kompfner: One thought: might be worth mentioning at this highest level comment that it outputs `AggregatedTextFrame`s?

src/pipecat/processors/frameworks/rtvi.py
mattieruth: i went this way because in the processor, the purpose of skip_tts is to track the mode the llm is in to be sure that we set it back to how it was before we toggle the mode.  So it feels more accurate than trying to detect the mode by handling every frame that has the flag.

src/pipecat/processors/aggregators/llm_text_processor.py
kompfner, line 49: Is this TODO for this PR?

src/pipecat/services/cartesia/tts.py
mattieruth: In Rime, I need access to self, which meant not building these outside the class. so I was trying to be consistent. Wdyt? I struggled with the best way to do this, so open to suggestions.

src/pipecat/processors/aggregators/llm_text_processor.py
```diff
@@ -0,0 +1,102 @@
...
+            direction: The direction of frame flow in the pipeline.
+        """
+        await super().process_frame(frame, direction)
+
+        if isinstance(frame, InterruptionFrame):
+            await self._handle_interruption(frame)
+            await self.push_frame(frame, direction)
+        elif isinstance(frame, LLMTextFrame):
+            await self._handle_llm_text(frame)
+        elif isinstance(frame, LLMFullResponseEndFrame):
```
kompfner, line 68: Do we not also need to handle `LLMFullResponseStartFrame`?

src/pipecat/services/rime/tts.py
mattieruth: this is the reason I didn't go with putting the helpers in a separate class. Rime's API requires these extra fields to be included in the message in order for the text modifications to work. i'll think some more on the best approach.

I wonder if instead of setting the fields message by message, if it would be ok to have a method to just set these fields for the duration if you know you are going to use these features 🤔

src/pipecat/services/tts_service.py
mattieruth: i went this way because if the internal aggregation is turned off and we get here, then we are literally just passing on `LLMTextFrame`s which are token-by-token. Which... is typically word? So I was on the fence with what to call this one and whether `"token"` should be added as one of the built-in aggregated types.

src/pipecat/processors/frameworks/rtvi.py
```diff
@@ -973,8 +1005,26 @@ def __init__(
                     DeprecationWarning,
                 )
 
+        self._aggregation_transforms: List[Tuple[str, Callable[[str, str], Awaitable[str]]]] = []
+
+    def transform_aggregation_type(
+        self, aggregation_type: str, transform_function: Callable[[str, str], Awaitable[str]]
+    ):
+        """Transform text for a specific aggregation type before sending as Bot Output or TTS.
```
kompfner, line 1029: So the use-case here is if you want text to go into the context in one way but output to the client another?

src/pipecat/utils/text/pattern_pair_aggregator.py
mattieruth: correct, any text before an AGGREGATE/pattern should be returned as a sentence (i've toyed with introducing another aggregation type that's like "incomplete", but decided that would make handling harder when you most likely just want to treat it as a "sentence")

src/pipecat/services/cartesia/tts.py
mattieruth: cc @kompfner @aconchillo

src/pipecat/services/tts_service.py
mattieruth: invitation to comment: @kompfner @aconchillo

src/pipecat/processors/aggregators/llm_response.py
mattieruth: oh nice catch! i removed it all from the universal one but forgot about these changes.

src/pipecat/processors/aggregators/llm_text_processor.py
mattieruth: no

src/pipecat/processors/aggregators/llm_text_processor.py
mattieruth: no. Start frames do not provide any text to be aggregated and we do not have any state that needs updating when a new set of frames starts. This is partially because we handle the ResponseEndFrame and ship trailing aggregations then, so we can assume that there's nothing to send along on Start.

src/pipecat/processors/frameworks/rtvi.py
mattieruth: correct. The best example I came up with was credit cards or similar secure information. You may want your TTS to spell out the credit card (and add it to the context), but you don't want to send the number across the wire wholesale, but rather something like "XXXX-XXXX-XXXX-1234"

src/pipecat/utils/text/pattern_pair_aggregator.py
mattieruth: let me know what you think about this approach to backwards compatibility (introducing a new method and calling it just `add_pattern`)

NOTE: There are still breaking changes in this class because the handlers and the `aggregate()` method return a new type.  I don't know any way around those.

CHANGELOG.md
mattieruth: ✅ Done

CHANGELOG.md
mattieruth: ✅ Done

src/pipecat/frames/frames.py
mattieruth: cc @aconchillo -- did i do this wrong?

src/pipecat/processors/frameworks/rtvi.py
```diff
@@ -973,8 +1005,41 @@ def __init__(
...
                 )
 
+        self._aggregation_transforms: List[Tuple[str, Callable[[str, str], Awaitable[str]]]] = []
+
+    def add_bot_output_transformer(
+        self, transform_function: Callable[[str, str], Awaitable[str]], aggregation_type: str = "*"
+    ):
+        """Transform text for a specific aggregation type before sending as Bot Output or TTS.
+
+        # TODO: What if someone wanted to remove a registered transform?
```
kompfner: Is this TODO for this PR or later?

kompfner: Ah wait, looks like you already have a `remove` method. So maybe this TODO can be removed?

src/pipecat/processors/frameworks/rtvi.py
```diff
@@ -1048,12 +1113,15 @@ async def on_push_frame(self, data: FramePushed):
             await self.send_rtvi_message(RTVIBotTTSStartedMessage())
         elif isinstance(frame, TTSStoppedFrame) and self._params.bot_tts_enabled:
             await self.send_rtvi_message(RTVIBotTTSStoppedMessage())
-        elif isinstance(frame, TTSTextFrame) and self._params.bot_tts_enabled:
-            if isinstance(src, BaseOutputTransport):
-                message = RTVIBotTTSTextMessage(data=RTVITextMessageData(text=frame.text))
-                await self.send_rtvi_message(message)
-            else:
+        elif isinstance(frame, AggregatedTextFrame) and (
```
kompfner, line 1132: So if the user wants `bot-output` events and have no TTS service (or are skipping it with `skip_tts=True`), they need to know to use an `LLMTextProcessor` to send `AggregatedTextFrame`s down the pipeline, yeah?

src/pipecat/services/cartesia/tts.py
```diff
@@ -182,6 +249,10 @@ def __init__(
             container: Audio container format.
             params: Additional input parameters for voice customization.
             text_aggregator: Custom text aggregator for processing input text.
+
+                .. deprecated:: 0.0.95
```
kompfner, line 253: 👍

src/pipecat/services/cartesia/tts.py
```diff
@@ -257,6 +336,27 @@ def language_to_service_language(self, language: Language) -> Optional[str]:
         """
         return language_to_cartesia_language(language)
 
+    # A set of Cartesia-specific helpers for text transformations
```
kompfner, line 339: 👍

src/pipecat/services/rime/tts.py
```diff
@@ -123,10 +127,17 @@ def __init__(
...
             **kwargs,
         )
 
+        if not text_aggregator:
+            # Always skip tags added for spelled-out text
+            # Note: This is primarily to support backwards compatibility.
+            #    The preferred way of taking advantage of Rime spelling is
+            #    to use an LLMTextProcessor and/or a text_transformer to identify
+            #    and insert these tags for the purpose of the TTS service alone.
+            self._text_aggregator = SkipTagsAggregator([("spell(", ")")])
```
kompfner, line 139: Ah I see, setting it here instead of passing the default value to super's init method prevents the deprecation from being printed. Nice. 👍

src/pipecat/services/tts_service.py
```diff
@@ -142,6 +153,19 @@ def __init__(
         self._voice_id: str = ""
         self._settings: Dict[str, Any] = {}
         self._text_aggregator: BaseTextAggregator = text_aggregator or SimpleTextAggregator()
+        if text_aggregator:
+            import warnings
+
+            with warnings.catch_warnings():
+                warnings.simplefilter("always")
+                warnings.warn(
+                    "Parameter 'text_aggregator' is deprecated. Use an LLMTextProcessor before the TTSService for custom text aggregation.",
```
kompfner, line 175: 👍

src/pipecat/services/tts_service.py
```diff
@@ -298,6 +322,35 @@ async def cancel(self, frame: CancelFrame):
             await self.cancel_task(self._stop_frame_task)
             self._stop_frame_task = None
 
+    def add_text_transformer(
```
kompfner, line 340: Is there a reason we only want to allow adding transformers dynamically rather than init time?

src/pipecat/processors/frameworks/rtvi.py
```diff
@@ -973,8 +1005,41 @@ def __init__(
                     DeprecationWarning,
                 )
 
+        self._aggregation_transforms: List[Tuple[str, Callable[[str, str], Awaitable[str]]]] = []
+
+    def add_bot_output_transformer(
```
kompfner, line 1024: Is there a reason to only support adding transformers dynamically rather than at init time?

mattieruth: nope. i think i started to add an init for it and then questioned the usability of providing a list of callbacks, but i'll add it. i just convinced myself the other way that it's not as ugly as i thought.

src/pipecat/processors/frameworks/rtvi.py
mattieruth: doh. yup. thanks.

src/pipecat/processors/frameworks/rtvi.py
mattieruth: Correctumundo.

src/pipecat/services/rime/tts.py
mattieruth: yes, but your comment made me see that i accidentally took it out from above. i meant to leave the arg in, but just remove the defaulting to SkipTagsAggregator

mattieruth: oh nm. i have it right. it was just that with the obfuscated lines. i thought i had removed it from the args.

src/pipecat/services/tts_service.py
```diff
@@ -500,14 +566,36 @@ async def _push_tts_frames(self, text: str):
...
+                # is set to False and these are sent word by word as part of the
+                # _words_task_handler in the WordTTSService subclass. However, to
+                # support use cases where an observer may want the full text before
+                # the audio is generated, we send along the AggregatedTextFrame here,
+                # but we set append_to_context to False so it does not cause duplication
+                # in the context. This is primarily used by the RTVIObserver to
+                # generate a complete bot-output.
+                src_frame.append_to_context = False
+                await self.push_frame(src_frame)
+            # Note: Text transformations only affect the text sent to the TTS. This allows
```
kompfner: OK, let me check my understanding: "sentence-level" TTS services can simply pass along its input text, untransformed, as output text. "Word-level" TTS services, send along individual words, which have _already passed through the remote TTS provider_, as output text frames, so they've necessarily already been affected by the transforms.

Is the hope that the individually-outputted words are still suitable for use in a context message, because once they've been handled by the remote TTS provider they shouldn't contain special tags, etc?

src/pipecat/utils/text/pattern_pair_aggregator.py
```diff
@@ -93,41 +127,94 @@ def add_pattern_pair(
         the end pattern, and treat the content between them as a match.
 
         Args:
-            pattern_id: Unique identifier for this pattern pair.
+            type: Identifier for this pattern pair. Should be unique and ideally descriptive.
+                  (e.g., 'code', 'speaker', 'custom'). type can not be 'sentence' as that is
```
kompfner: or 'word', right

kompfner: or is that no longer "reserved" for internal use?

src/pipecat/utils/text/pattern_pair_aggregator.py
```diff
@@ -93,41 +127,94 @@ def add_pattern_pair(
...
+        .. deprecated:: 0.0.95
+            This function is deprecated and will be removed in a future version.
+            Use `add_pattern` with a type and MatchAction instead.
+
+            This method calls `add_pattern` setting type with the provided pattern_id and action
+            to either MatchAction.REMOVE or MatchAction.KEEP based on `remove_match`.
+
+        Args:
+            pattern_id: Identifier for this pattern pair. Should be unique and ideally descriptive.
+                        (e.g., 'code', 'speaker', 'custom'). pattern_id can not be 'sentence' as that is
```
kompfner: see above comment about 'word'

kompfner (approved):
Left a few minor comments. Overall this looks great!

src/pipecat/processors/frameworks/rtvi.py
mattieruth: ✅ added

src/pipecat/services/tts_service.py
mattieruth: nope. ✅ added.

src/pipecat/services/tts_service.py
mattieruth: > OK, let me check my understanding: "sentence-level" TTS services can simply pass along its input text, untransformed, as output text. "Word-level" TTS services, send along individual words, which have already passed through the remote TTS provider, as output text frames, so they've necessarily already been affected by the transforms.

Yes. or maybe better put. (and note that what i'm about to say is how things have always worked): When a TTS service supports word timestamps, we generate TTSTextFrames word-by-word for the context aggregator to collect, ensuring that only the spoken text is added and therefore accurately matching what the client hears. When a TTS service does not support word timestamps, we have to assume the TTS spoke exactly what we sent it and generate a TTSTextFrame with the original full sentence text to be added to the context. Also, to support our best guess, we wait to send the text AFTER the tts has completed so that interrupted sentences aren't added.

So because of these difference in behaviors, when you transform what goes into a word-by-word TTS, the context is going to represent what was actually said and the transform you made. But when you transform a TTS without word-by-word support, we have a choice. We can send the original text or the transformed text.

In my comment above, I say that we have made the choice of sending the original text, but your comment had me looking at the code again and it looks like i'm actually sending along the transformed text.  After typing all this out, i'm torn on what we want here. Originally, I thought "send original". But to your next comment:

> Is the hope that the individually-outputted words are still suitable for use in a context message, because once they've been handled by the remote TTS provider they shouldn't contain special tags, etc?

I'm not sure we ever thought about what we "hoped for" in the word-word output, but I can tell you from my testing yesterday with cartesia, it turns out that it spits out tags with the words! so special tags DO get added to the context.

So now I wonder, maybe the non word-word TTS should push the transformed text to be consistent.

ACK. I don't know.


Relatedly, I have the whole comment and logic above for sending the sentence before it's spoken and only doing so for the word-by-word TTS services. Now that those are marked "spoken: False", maybe I should be doing that no matter what?

going in circles here.

kompfner: > maybe the non word-word TTS should push the transformed text

Yeah it does seem like it might come down to consistency v "doing the right thing" at least some of the time...

If the goal is for the context to be free of TTS-specific cruft/tags, doing the right thing part of the time seems preferable...

Is an alternative to somehow "reverse transform" the word-by-word TTS output to make pristine `TTSTextFrame`s to send downstream?

kompfner: > the goal is for the context to be free of TTS-specific cruft/tags

^ the reason for this, I think, is to not influence how the LLM responds in the future. We may not want it sticking in the tags itself—we want our transformers to do it, so we can do it in a different TTS-specific way for each TTS. This'll leave us from to swap TTS services at runtime.

src/pipecat/processors/aggregators/llm_text_processor.py
filipi87: I can see that you have highlighted that in the CHANGELOG, maybe that is all that we can do for now.

src/pipecat/processors/aggregators/llm_text_processor.py
filipi87: > of course, if we could assume people now use skip_tts and a mode shift and all frames between start/end have the same value, that would make this much nicer :)

Maybe we can assume this for now and leave any refactoring needed for `skip_tts` to a follow up PR, so we don’t start complicating this one.

src/pipecat/frames/frames.py
filipi87: You are right, `init=False` only means it will not be set via the constructor. Sorry, my mistake.

src/pipecat/services/cartesia/tts.py
filipi87: Yeah, I think it makes sense to be consistent. 

I also added some comments inside Rime to check whether it was possible to think of something different there, but I wasn’t happy with any of the alternatives so far. 😅

filipi87 (approved):
I think we still have a couple of small nits, like maybe grouping the transformers from each TTS provider inside a class, but I know we first need to think about a good alternative for Rime in this case.

So I am leaving my approval here, because I know how hard it is to keep a PR like this up to date, and we can implement those improvements in a follow up PR if we decide they are worth it.

Amazing job! This will be a great improvement to Pipecat. 👏🚀

src/pipecat/services/rime/tts.py
mattieruth: nm on my last thought about setting the fields for the duration. that doesn't work because the fields are specific to the message. For example setting speeds on things:

```
{
    ...otherParams,
    "text": "This sentence is [really] [fast]",
    "inlineSpeedAlpha": "0.5, 3"
}
```

src/pipecat/services/tts_service.py
mattieruth: Update: we chatted offline and two things were decided:

1. We should send the `AggregatedTextFrame` along with `append_to_context` `False` no matter the TTS service for consistency. This means that we will always send a bot-output message with the aggregated and `spoken` set to `False`, followed by one or more `bot-output` messages spoken `True` and the same text, just either separated out word-by-word or repeated as a full sentence.

2. The idea of the tts transformation callbacks is specifically meant for modifying the text in a tts specific way that shouldn't affect the context. Now, if a word-by-word TTS delivers back the text with the tags in them. that feels more like a bug and something separate to fix. But given the ideal, the right thing to do in the non-word-by-word TTS case is to send the original text and NOT the transformed text.

mattieruth:
closing in favor of https://github.com/pipecat-ai/pipecat/pull/3107

mattieruth:
closing in favor of cleaned up and rebased https://github.com/pipecat-ai/pipecat/pull/3107

PR #2901, opened 2025-10-21
Add `messages` property to `LLMContext` for usage parity with `OpenAI…

kompfner:
…LLMContext`.

This wasn't really an issue before, when folks were *knowingly* migrating from `OpenAILLMContext` to `LLMContext`. But in the latest AWS Nova Sonic change, we're swapping it out from under folks, so this kind of compatibility is more important.

For context, the reason we *didn't* offer the `messages` property earlier was to aid in the development of `LLMContext`—we wanted to draw attention to all the places where messages were being read from context, so we could find the places where we might need to pass an argument to the read.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -41,9 +41,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   # or
   context: OpenAILLMContext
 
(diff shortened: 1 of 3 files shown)


aconchillo:
LGTM!

PR #2902, opened 2025-10-22
Fix: Properly close OpenAI SSE stream on interruption

jamsea:
## Problem

When an `InterruptionFrame` is received (e.g., when a user interrupts the bot mid-response), it triggers task cancellation via `asyncio.CancelledError`. However, the OpenAI `AsyncStream[ChatCompletionChunk]` was not being properly closed, which could lead to:

- Connection leaks
- Inconsistent stream state  
- Potential resource exhaustion over time

## Root Cause

The `_process_context()` method in `base_llm.py` iterates over the SSE stream with:
```python
async for chunk in chunk_stream:
    # Process chunks...
```

When an `InterruptionFrame` arrives, the frame processor cancels the current process task, but the stream is not explicitly closed, leaving the underlying HTTP/SSE connection in an inconsistent state.

## Solution

This PR wraps the entire stream processing section with proper exception handling:

1. **Try block**: Contains all the stream iteration and processing logic
2. **Except block**: Catches `asyncio.CancelledError`, logs it for debugging, and re-raises it to maintain proper cancellation propagation  
3. **Finally block**: Ensures `chunk_stream.close()` is always called to properly clean up the SSE connection

## Changes

- Added try/except/finally around the `async for chunk in chunk_stream:` loop
- Added explicit `await chunk_stream.close()` in the finally block
- Added debug logging when cancellation occurs
- Preserved all existing functionality while ensuring proper resource cleanup

### DIFF
--- a/src/pipecat/services/openai/base_llm.py
+++ b/src/pipecat/services/openai/base_llm.py
@@ -339,92 +339,100 @@ class BaseOpenAILLMService(LLMService):
             else self._stream_chat_completions_universal_context(context)
         )
 
-        async for chunk in chunk_stream:
-            if chunk.usage:
-                cached_tokens = (
-                    chunk.usage.prompt_tokens_details.cached_tokens
-                    if chunk.usage.prompt_tokens_details
-                    else None
-                )
-                tokens = LLMTokenUsage(
-                    prompt_tokens=chunk.usage.prompt_tokens,
-                    completion_tokens=chunk.usage.completion_tokens,
-                    total_tokens=chunk.usage.total_tokens,
-                    cache_read_input_tokens=cached_tokens,
-                )
-                await self.start_llm_usage_metrics(tokens)
-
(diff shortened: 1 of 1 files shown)


src/pipecat/services/openai/base_llm.py
```diff
@@ -339,92 +339,100 @@ async def _process_context(self, context: OpenAILLMContext | LLMContext):
...
 
-            await self.run_function_calls(function_calls)
+                await self.run_function_calls(function_calls)
+        except asyncio.CancelledError:
+            # Handle cancellation gracefully (e.g., from InterruptionFrame)
+            logger.debug(f"{self}: Stream processing cancelled due to interruption")
+            raise
+        finally:
+            # Ensure the SSE stream is properly closed to avoid connection leaks
+            await chunk_stream.close()
```
jamsea, line 435: @aconchillo are we intentionally not calling this function presently? The main change I'm interested in is line 431, adding a debug log

PR #2905, opened 2025-10-22
update gemini models to 2.5

mhussam-ai:

Updated the Google Gemini Models to 2.5 family

### DIFF
--- a/src/pipecat/services/google/gemini_live/llm.py
+++ b/src/pipecat/services/google/gemini_live/llm.py
@@ -534,7 +534,7 @@ class GeminiLiveLLMService(LLMService):
         *,
         api_key: str,
         base_url: Optional[str] = None,
-        model="models/gemini-2.0-flash-live-001", # e.g. "gemini-2.5-flash-native-audio-preview-09-2025"
+        model="models/gemini-2.0-flash-live-001",
         voice_id: str = "Charon",
         start_audio_paused: bool = False,
         start_video_paused: bool = False,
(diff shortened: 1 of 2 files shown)


src/pipecat/services/google/gemini_live/llm.py
```diff
@@ -534,7 +534,7 @@ def __init__(
         *,
         api_key: str,
         base_url: Optional[str] = None,
-        model="models/gemini-2.0-flash-live-001",
+        model="models/gemini-2.0-flash-live-001", # e.g. "gemini-2.5-flash-native-audio-preview-09-2025"
```
markbackman: No need to comment. I think you can remove:
```suggestion
        model="models/gemini-2.0-flash-live-001",
```

The current default is still the right default, for now.

src/pipecat/services/google/gemini_live/llm_vertex.py
```diff
@@ -51,7 +51,7 @@ def __init__(
         credentials_path: Optional[str] = None,
         location: str,
         project_id: str,
-        model="google/gemini-2.0-flash-live-preview-04-09",
+        model="google/gemini-2.0-flash-live-preview-04-09", # can it be "gemini-2.5-flash-native-audio-preview-09-2025"?
```
markbackman: No, I don't think the same model is supported. Let's remove the comment:
```suggestion
        model="google/gemini-2.0-flash-live-preview-04-09",
```

mhussam-ai: Thanks for clarification , I will remove it.

src/pipecat/services/google/gemini_live/llm.py
mhussam-ai: Alright, Thank you

PR #2906, opened 2025-10-22
Add comprehensive error handling to OpenAI LLM service

jamsea:
This PR adds consistent error propagation to the OpenAI base LLM service, matching the pattern already used in the Anthropic service.

## Changes
- Added ErrorFrame import
- Added catch-all exception handler
- Properly handles CancelledError by re-raising
- Logs exceptions with full traceback
- Pushes ErrorFrame upstream for error detection
- Maintains existing timeout handling

## Problem Solved
Previously, OpenAI LLM service errors were suppressed and not communicated through the frame pipeline. This made it impossible to implement automatic error recovery strategies like LLM hot-switching.

## Benefits
- Enables consistent error detection across all LLM services
- Allows downstream error monitors to detect failures
- Makes LLM hot-switching and fallback strategies possible
- Improves debugging with proper error logging

### DIFF
--- a/src/pipecat/services/openai/base_llm.py
+++ b/src/pipecat/services/openai/base_llm.py
@@ -13,18 +13,13 @@ from typing import Any, Dict, List, Mapping, Optional
 
 import httpx
 from loguru import logger
-from openai import (
-    NOT_GIVEN,
-    APITimeoutError,
-    AsyncOpenAI,
-    AsyncStream,
-    DefaultAsyncHttpxClient,
-)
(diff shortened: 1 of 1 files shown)


src/pipecat/services/openai/base_llm.py
```diff
@@ -460,8 +455,13 @@ async def process_frame(self, frame: Frame, direction: FrameDirection):
                 await self.push_frame(LLMFullResponseStartFrame())
                 await self.start_processing_metrics()
                 await self._process_context(context)
+            except asyncio.CancelledError:
+                raise
```
aconchillo, line 459: If we are not doing anything, no need to handle it. By default it's already raised.

PR #2911, opened 2025-10-23
DailyTransport: update start_dialout/start_recording return values

aconchillo:

A few updates:
- The main one are the return values of `start_dialout` and `start_recording`, which are now `(session_id, error)` or `(recording_id, error)`.
- Also, we move error logging to the transport instead of the client as much as possible.
- We now check for errors and log them in every `daily-python` function called.

### DIFF
--- a/src/pipecat/transports/daily/transport.py
+++ b/src/pipecat/transports/daily/transport.py
@@ -2479,7 +2479,11 @@ class DailyTransport(BaseTransport):
     async def _on_joined(self, data):
         """Handle room joined events."""
         if self._params.transcription_enabled:
-            await self.start_transcription(self._params.transcription_settings)
+            # We report an error because we are starting transcription
+            # internally and if it fails we need to know.
+            error = await self.start_transcription(self._params.transcription_settings)
+            if error:
+                await self._on_error(f"Unable to start transcription: {error}")
         await self._call_event_handler("on_joined", data)
 
     async def _on_left(self):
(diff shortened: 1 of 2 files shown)


src/pipecat/transports/daily/transport.py
```diff
@@ -755,9 +758,6 @@ async def join(self):
 
                 logger.info(f"Joined {self._room_url}")
 
-                if self._params.transcription_enabled:
-                    await self.start_transcription(self._params.transcription_settings)
```
aconchillo: We now do this in `on_joined` callback instead of in the `join` function itself. I think it's much cleaner that burying things.

src/pipecat/transports/daily/transport.py
```diff
@@ -847,9 +847,6 @@ async def leave(self):
         # Call callback before leaving.
         await self._callbacks.on_before_leave()
 
-        if self._params.transcription_enabled:
-            await self.stop_transcription()
```
aconchillo, line 846: Same here. We just stop transcription in `on_before_leave`. Good thing we added support for that!

src/pipecat/transports/daily/transport.py
```diff
@@ -873,181 +873,189 @@ def _cleanup(self):
...
+        """
         future = self._get_event_loop().create_future()
         self._client.start_dialout(settings, completion=completion_callback(future))
-        error = await future
-        if error:
-            logger.error(f"Unable to start dialout: {error}")
+        return await future
 
-    async def stop_dialout(self, participant_id):
+    async def stop_dialout(self, participant_id) -> CallClientError:
```
filipi87: Wouldn't it be more idiomatic if, instead of `CallClientError` being an `Optional[str]`, it were just a `str`, and here we returned an `Optional[CallClientError]`?
The way it is, it feels like it will always return an error. But maybe that's just me. 😅

src/pipecat/transports/daily/transport.py
filipi87: Yep. Agreed. 👍

src/pipecat/transports/daily/transport.py
aconchillo: I think you are right! Done!

PR #2915, opened 2025-10-24
fix: use correct  property names

Rickaym:

Just a small fix to correct the property name of a deprecated signature. stumbled upon this after being recommended by kapa.ai to use the old signature.


markbackman (approved):
Nice catch! 🕵️

PR #2920, opened 2025-10-27
WIP: Add JSON serialization and additional frame testing utilities

chadbailey59:
I'm working on adding some custom FrameProcessor testing capability to the pipecat CLI. This PR adds a few things to help with that.

The goal is to be able to define input frames and optional expected output frames in JSON, like this:

```json
{
  "input_frames": [
    {"type": "TextFrame", "text": "hello world"},
    {"type": "TextFrame", "text": "testing 123"}
  ],
  "expected_output": [
    {"type": "TextFrame", "text": "HELLO WORLD"},
    {"type": "TextFrame", "text": "TESTING 123"}
  ]
}
```

The [existing run_test function](https://github.com/pipecat-ai/pipecat/blob/main/src/pipecat/tests/utils.py#L123) requires you to define expected output frames and has test assertions built in. This PR effectively re-implements run test as `run_test_from_file` to move the assertions part to a separate function. This lets us make expected_output optional, so we can run the FrameProcessor without knowing what output we expect when we're building our FrameProcessor.

The custom validator also adds a bit more utility, such as validating field contents instead of just frame types. It also starts to build out a bit of a testing DSL, like adding `_contains` to a string field to test for contains instead of equals, like: ` {"type": "TextFrame", "text_contains": "TESTING"}` from the example above.

There's some mild duplication in creating a separate `run_test_from_file` function. It's just extremely simple pipeline creation, and I think it's worth it not to disturb the existing `run_test` function. But I could be convinced otherwise.

### DIFF
--- a/src/pipecat/tests/__init__.py
+++ b/src/pipecat/tests/__init__.py
@@ -1,0 +1,12 @@
+#
+# Copyright (c) 2024-2025 Daily
+#
+# SPDX-License-Identifier: BSD 2-Clause License
+#
+
+"""Public testing API for Pipecat frame processors."""
+
+from .serialization import dict_to_frame, frame_to_dict, load_frames_from_json
+from .test_runner import run_test_from_file
+
+__all__ = ["dict_to_frame", "frame_to_dict", "load_frames_from_json", "run_test_from_file"]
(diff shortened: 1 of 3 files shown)


markbackman:
Closing this one out.

PR #2924, opened 2025-10-28
DailyTransport: don't timeout prematurely on join

aconchillo:

Let `daily-python` control the internal retries and timeout instead of randomly timing out.


filipi87 (approved):
LGTM!

PR #2927, opened 2025-10-28
Add RTVIProcessor to foundational example 38b

marcus-daily:


examples/foundational/38b-smart-turn-local.py
```diff
@@ -78,9 +79,12 @@ async def run_bot(transport: BaseTransport, runner_args: RunnerArguments):
     context = LLMContext(messages)
     context_aggregator = LLMContextAggregatorPair(context)
 
+    rtvi = RTVIProcessor(config=RTVIConfig(config=[]))
```
markbackman: This can now be simplified to:

```suggestion
    rtvi = RTVIProcessor()
```

PR #2932, opened 2025-10-28
Update `GeminiLLMService` to work with `LLMContext` and `LLMContextAg…

kompfner:
…gregatorPair`

This PR is mostly following the patterns established in https://github.com/pipecat-ai/pipecat/pull/2889 and https://github.com/pipecat-ai/pipecat/pull/2750, except it had a "head start" because it already had a working context adapter in `GeminiLLMAdapter`.

This PR represents the *last* of the migrations to `LLMContext` (assuming it lands after https://github.com/pipecat-ai/pipecat/pull/2889, which is also still open). After that, we can start the process of deprecating `OpenAILLMContext` and associated code.

NOTE: there's still one occasional bug in context record-keeping that I'd like to investigate: during some user interruptions, the assistant aggregator receives too many `LLMFullResponseEndFrame`, resulting in the loss of some assistant text.

### DIFF
--- a/src/pipecat/services/google/gemini_live/llm.py
+++ b/src/pipecat/services/google/gemini_live/llm.py
@@ -223,7 +223,7 @@ class GeminiLiveContext(OpenAILLMContext):
     Provides Gemini-specific context management including system instruction
     extraction and message format conversion for the Live API.
 
-    .. deprecated:: 0.0.93
+    .. deprecated:: 0.0.92
         Gemini Live no longer uses `GeminiLiveContext` under the hood.
         It now uses `LLMContext`.
     """
@@ -352,7 +352,7 @@ class GeminiLiveUserContextAggregator(OpenAIUserContextAggregator):
     Extends OpenAI user aggregator to handle Gemini-specific message passing
     while maintaining compatibility with the standard aggregation pipeline.
 
-    .. deprecated:: 0.0.93
+    .. deprecated:: 0.0.92
         Gemini Live no longer expects a `GeminiLiveUserContextAggregator`.
         It now expects a `LLMUserAggregator`.
     """
@@ -393,7 +393,7 @@ class GeminiLiveAssistantContextAggregator(OpenAIAssistantContextAggregator):
     to prevent duplicate context entries, as Gemini Live pushes both
     LLMTextFrames and TTSTextFrames.
 
-    .. deprecated:: 0.0.93
+    .. deprecated:: 0.0.92
         Gemini Live no longer uses `GeminiLiveAssistantContextAggregator` under the hood.
         It now uses `LLMAssistantAggregator`.
     """
@@ -443,7 +443,7 @@ class GeminiLiveAssistantContextAggregator(OpenAIAssistantContextAggregator):
 class GeminiLiveContextAggregatorPair:
     """Pair of user and assistant context aggregators for Gemini Live.
 
-    .. deprecated:: 0.0.93
+    .. deprecated:: 0.0.92
         `GeminiLiveContextAggregatorPair` is deprecated.
         Use `LLMContextAggregatorPair` instead.
 


examples/foundational/26a-gemini-live-transcription.py
```diff
@@ -90,7 +92,12 @@ async def run_bot(transport: BaseTransport, runner_args: RunnerArguments):
             #     },
         ],
     )
-    context_aggregator = llm.create_context_aggregator(context)
+    context_aggregator = LLMContextAggregatorPair(
+        context,
+        # `expect_stripped_words=False` needed when Gemini Live used with AUDIO
```
kompfner, line 97: This pattern is a bit ugly.

Prior to this change, though, we actually had a bug—`create_context_aggregator()` would *always* return an assistant context aggregator configured with `expect_stripped_words=False`, even when it ought not to have been (i.e. when the LLM was configured to output text).

src/pipecat/services/google/gemini_live/llm.py
```diff
@@ -219,6 +222,10 @@ class GeminiLiveContext(OpenAILLMContext):
 
     Provides Gemini-specific context management including system instruction
     extraction and message format conversion for the Live API.
+
+    .. deprecated:: 0.0.93
```
kompfner: I'm just guessing that this PR won't land for version 0.0.92.

Note to self: update this if this assumed version turns out to be wrong.

src/pipecat/services/google/gemini_live/llm.py
```diff
@@ -1458,7 +1568,17 @@ async def _handle_msg_output_transcription(self, message: LiveServerMessage):
         # Collect text for tracing
         self._llm_output_buffer += text
 
-        await self.push_frame(LLMTextFrame(text=text))
+        # NOTE: Shoot. When using Vertex AI, output transcription messages
```
kompfner, line 1583: Yuck.

examples/foundational/26a-gemini-live-transcription.py
filipi87: Agreed. I wonder if this is something we could change in the future, so we wouldn't need to pass `expect_stripped_words`, and could instead identify directly inside the aggregator whether the words we receive are stripped or not.

Or maybe this should be provided by the `LLMAdapter`?

It does feel a bit weird that the user has to define that, it seems like something internal to how we handle data from the LLM.

src/pipecat/services/google/gemini_live/llm.py
```diff
@@ -883,6 +957,40 @@ async def process_frame(self, frame: Frame, direction: FrameDirection):
...
 
+    async def _handle_context(self, context: LLMContext):
+        if not self._context:
+            # We got our initial context
+            self._context = context
+            if context.tools:
+                self._tools = context.tools
+            # Initialize our bookkeeping of already-completed tool calls in
+            # the context
+            await self._process_completed_function_calls(send_new_results=False)
```
filipi87, line 971: Is this for the case where we’re loading a previous context ?

src/pipecat/services/google/gemini_live/llm.py
filipi87: Yeah, that is crazy. 🤕

examples/foundational/26a-gemini-live-transcription.py
markbackman: The same is true of OpenAI Realtime, IIRC. How did you work around it there?

Alternatively, is `expect_stripped_words` more of an internal variable to handle the output of an LLM such that it conforms with the context format Pipecat expects? If so, is there a reason it shouldn't be set to False by default so that you don't have to set it when initializing the `LLMContextAggregatorPair`?

This is such a small piece that it will be overlooked and will create problems if we require it.

src/pipecat/services/google/gemini_live/llm.py
kompfner: Yes that's most likely the scenario where there'd be already-completed tool calls in the initial context.

But who knows—there might be other ways users seed context where already-completed tool calls could make their way in there. The point is just that we shouldn't bake in the assumption that the initial context _won't_ contain tool calls.

markbackman: Same question.

src/pipecat/services/google/gemini_live/llm.py
markbackman: We should ask the Google team about this one. This is really weird and makes handling this more difficult.

examples/foundational/26a-gemini-live-transcription.py
markbackman: It looks like `assistant_params.expect_stripped_words = False` is still initialized in the `create_context_aggregator` function. Do we actually need to set this value here?

kompfner:
Snuck in a last couple of bug fixes

examples/foundational/26a-gemini-live-transcription.py
kompfner: > It looks like assistant_params.expect_stripped_words = False is still initialized in the create_context_aggregator function. Do we actually need to set this value here?

Correct. The `create_context_aggregator` function is still around for backward compatibility—it should work exactly as it used to, with users being none the wiser. It's part of the "old pattern" of context setup. So we need to set up `assistant_params` there like it was doing before.

kompfner: > The same is true of OpenAI Realtime, IIRC. How did you work around it there?

I did not hehe. I actually have a personal TODO to go back and fix it.

kompfner: > Alternatively, is expect_stripped_words more of an internal variable to handle the output of an LLM such that it conforms with the context format Pipecat expects? If so, is there a reason it shouldn't be set to False by default so that you don't have to set it when initializing the LLMContextAggregatorPair?

If it's `False` by default it won't work properly for most services.

But I agree with you all that there should ideally be a way for the context aggregator to "auto-detect" the right setting for this.

kompfner: The fundamental problem is that the aggregator and the LLM are two independently-configured services in the pipeline. I'll look into whether there's a way for either:
* The aggregator to not require the `expect_stripped_words` flag at all
* The LLM to somehow communicate to the LLM at pipeline start time what setting to use for that flag
* Something else

kompfner: > Or maybe this should be provided by the LLMAdapter

The assistant aggregator isn't aware of the adapter. It doesn't know which LLM service is upstream of it.

kompfner: Actually...needing to configure the assistant aggregator differently depending on which LLM service is upstream of it is kind of a fundamental problem—we can't have that be the case if we want to be able to dynamically switch LLMs at runtime...

Having the LLM "communicate" downstream at setup time sounds too complicated for the LLM-swapping scenario. I think what @filipi87 is suggesting—of making the aggregator smart enough to not have to be told what to expect—is the preferred solution.

kompfner: I think this work can come in a follow-up PR—when the changes in this PR land, if nobody updates their client code, things will automatically work for them: a properly-configured assistant context aggregator is returned from `create_context_aggregator()`.

kompfner:
OK, gonna go ahead and merge this (with version numbers updated to 0.0.92 throughout) and then work on the `expect_stripped_words` thing in a follow-up PR.

PR #2935, opened 2025-10-28
improve image and vision support

aconchillo:

This PR adds the following:

- New `LLMContext` static methods to generate image (base64 or URL) and audio messages so they can be used with `LLMMessagesAppendFrame` to add images/audio to the context.
- `UserImageRequestFrame` is not tied to function calls anymore.
- `UserImageRawFrame` now has an associated text and if the iamge should be added to the context.
- Vision services, like Moondream, now wait for a `UserImageRawFrame` instead of trying to get an image from the context. With this, you just need to use a parallel pipeline.
- We also deprecated `UserResponseAggregator` which was only used in these 12-series examples. I think the examples now show a more real use case.
- Foundational examples 12 and 14 have been updated with all these changes.

### DIFF
--- a/examples/foundational/14d-function-calling-anthropic-video.py
+++ b/examples/foundational/14d-function-calling-anthropic-video.py
@@ -59,6 +59,10 @@ async def fetch_user_image(params: FunctionCallParams):
 
     await params.result_callback(None)
 
+    # Instead of None, it's possible to also provide a tool call answer to
+    # tell the LLM that we are grabbing the image to analyze.
+    # await params.result_callback({"result": "Image is being captured."})
+
 
 # We store functions so objects (e.g. SileroVADAnalyzer) don't get
 # instantiated. The function will be called when the desired transport gets
--- a/examples/foundational/14d-function-calling-moondream-video.py
+++ b/examples/foundational/14d-function-calling-moondream-video.py
@@ -61,6 +61,10 @@ async def fetch_user_image(params: FunctionCallParams):
 
     await params.result_callback(None)
 
+    # Instead of None, it's possible to also provide a tool call answer to
+    # tell the LLM that we are grabbing the image to analyze.
+    # await params.result_callback({"result": "Image is being captured."})
+
 
 # We store functions so objects (e.g. SileroVADAnalyzer) don't get
 # instantiated. The function will be called when the desired transport gets
--- a/examples/foundational/14d-function-calling-aws-video.py
+++ b/examples/foundational/14d-function-calling-aws-video.py
@@ -59,6 +59,10 @@ async def fetch_user_image(params: FunctionCallParams):
 
     await params.result_callback(None)
 
+    # Instead of None, it's possible to also provide a tool call answer to
+    # tell the LLM that we are grabbing the image to analyze.
+    # await params.result_callback({"result": "Image is being captured."})
+
 
 # We store functions so objects (e.g. SileroVADAnalyzer) don't get
 # instantiated. The function will be called when the desired transport gets
--- a/examples/foundational/14d-function-calling-gemini-flash-video.py
+++ b/examples/foundational/14d-function-calling-gemini-flash-video.py
@@ -59,6 +59,10 @@ async def fetch_user_image(params: FunctionCallParams):
 
     await params.result_callback(None)
 
+    # Instead of None, it's possible to also provide a tool call answer to
+    # tell the LLM that we are grabbing the image to analyze.
+    # await params.result_callback({"result": "Image is being captured."})
+
 
 # We store functions so objects (e.g. SileroVADAnalyzer) don't get
 # instantiated. The function will be called when the desired transport gets
--- a/examples/foundational/14d-function-calling-openai-video.py
+++ b/examples/foundational/14d-function-calling-openai-video.py
@@ -60,6 +60,10 @@ async def fetch_user_image(params: FunctionCallParams):
 
     await params.result_callback(None)
 
+    # Instead of None, it's possible to also provide a tool call answer to
+    # tell the LLM that we are grabbing the image to analyze.
+    # await params.result_callback({"result": "Image is being captured."})
+
 
 # We store functions so objects (e.g. SileroVADAnalyzer) don't get
 # instantiated. The function will be called when the desired transport gets


kompfner:
> We re-add VisionImageRawFrame. This existed, I'm not sure when or why we got rid of it.

We removed it recently. Read about the reasoning here: https://github.com/pipecat-ai/pipecat/pull/2598.

(Looks like you reviewed the above PR hehe).

src/pipecat/adapters/services/bedrock_adapter.py
```diff
@@ -256,15 +256,22 @@ def _from_standard_message(self, message: LLMStandardMessage) -> dict[str, Any]:
                     new_content.append({"text": text_content})
                 # handle image_url -> image conversion
                 if item["type"] == "image_url":
-                    new_item = {
-                        "image": {
-                            "format": "jpeg",
-                            "source": {
-                                "bytes": base64.b64decode(item["image_url"]["url"].split(",")[1])
-                            },
+                    if item["image_url"]["url"].startswith("data:"):
```
markbackman, line 259: Do we need to handle the `http` case here?

src/pipecat/adapters/services/gemini_adapter.py
```diff
@@ -345,7 +345,7 @@ def _from_standard_message(
             for c in content:
                 if c["type"] == "text":
                     parts.append(Part(text=c["text"]))
-                elif c["type"] == "image_url":
+                elif c["type"] == "image_url" and c["image_url"]["url"].startswith("data:"):
```
markbackman, line 346: Same questions here about the `http` case.

kompfner:
> We removed it recently. Read about the reasoning here: #2598.

The *main* reason for removing it was that it was a frame that had this very specific behavior of being _directly_ consumed by LLMs, in lieu of context frames, and that was a pattern I wanted to move away from.

When I was looking for ways to update all LLMs to consume `LLMContext`s rather than `OpenAILLMContext`s, rather than updating the `VisionImageRawFrame` code paths to use `LLMContext`, it seemed like an opportunity to clean up this wonky pattern.

kompfner:
> The _main_ reason for removing it was that it was a frame that had this very specific behavior of being _directly_ consumed by LLMs

I don't disagree that it would be handy for there to be an easy-to-create image frame that folks can add to context, though... (`UserImageRawFrame` in _theory_ should be enough but it has enough additional cruft that isn't always relevant)

...but I'm not sure it should necessarily be revived under the same name as before, as `VisionImageRawFrame` previously had this very specific and quirky context-less behavior associated with it (and code might still be out there expecting that?)

examples/foundational/14d-function-calling-moondream-video.py
```diff
@@ -37,32 +37,24 @@
...
-        await super().process_frame(frame, direction)
-
-        if self._participant_id and isinstance(frame, TextFrame):
-            await self.push_frame(
-                UserImageRequestFrame(self._participant_id, context=frame.text),
-                FrameDirection.UPSTREAM,
-            )
-        else:
-            await self.push_frame(frame, direction)
+async def fetch_user_image(params: FunctionCallParams):
```
markbackman, line 43: Since this example is a bit complex, I think some docstrings can help to explain what happens. We can either do this at the top-level or in the individual functions. I think the latter may be better.

For example, here, we could add:
```
    """Fetch the user image and push it to the LLM.

    When called, this function pushses a UserImageRequestFrame upstream to the
    transport. As a result, the transport will request the user image and push it
    downstream.
    """
```

markbackman: In UserImageProcessor, we could add:
```
    """Converts incoming user images into context frames.

    This processor handles the UserImageRawFrame from the transport, creates an LLMContextMessage
    and pushes an LLMMessagesAppendFrame downstream to be handled by the LLM.
    """
```

markbackman: The same update to 12a-d would be helpful.

Same for example 12 (moondream), but tailored to the different workflow.

examples/foundational/12a-describe-video-gemini-flash.py
```diff
@@ -110,33 +100,57 @@ async def process_frame(self, frame: Frame, direction: FrameDirection):
...
-    # Google Gemini model for vision analysis
-    google = GoogleLLMService(model="gemini-2.0-flash-001", api_key=os.getenv("GOOGLE_API_KEY"))
-
     tts = CartesiaTTSService(
         api_key=os.getenv("CARTESIA_API_KEY"),
         voice_id="71a7ad14-091c-4e8e-a314-022ece01c121",  # British Reading Lady
     )
 
+    # Google Gemini model for vision analysis
+    llm = GoogleLLMService(model="gemini-2.0-flash-001", api_key=os.getenv("GOOGLE_API_KEY"))
```
markbackman: We should have this example use the latest model.

```suggestion
    llm = GoogleLLMService(api_key=os.getenv("GOOGLE_API_KEY"))
```

src/pipecat/adapters/services/bedrock_adapter.py
kompfner: Seems like the `http` case falls under "unsupported". Is that an AWS Bedrock limitation maybe?

src/pipecat/frames/frames.py
```diff
@@ -1305,6 +1305,24 @@ def __str__(self):
         return f"{self.name}(pts: {pts}, user: {self.user_id}, source: {self.transport_source}, size: {self.size}, format: {self.format}, request: {self.request})"
 
 
+@dataclass
+class VisionImageRawFrame(InputImageRawFrame):
```
kompfner: See my PR-level note about reviving this with the previous name.

examples/foundational/12a-describe-video-gemini-flash.py
```diff
@@ -73,15 +65,13 @@ async def process_frame(self, frame: Frame, direction: FrameDirection):
 
         if isinstance(frame, UserImageRawFrame):
             if frame.request and frame.request.context:
-                context = LLMContext()
-                context.add_image_frame_message(
+                message = LLMContext.create_image_message(
```
kompfner: I like this a lot. The less the user has to be aware of context frames, the better.

kompfner: Although...this new pattern of using `LLMMessagesAppendFrame` kind of breaks the original intent (at least my understanding of the original intent) of these `12-*` examples.

The `12-*` examples were demonstrating single-shot, no-persistent-context image analyses. Moondream can do only that, if I'm not mistaken. And the other `12-*` files were showing how you could achieve the equivalent behavior with different services.

But maybe you're claiming that single-shot, no-persistent-context usage is not really realistic usage for services other than single-shot vision services like Moondream. I buy that. 

But if that's the claim, I don't think we should update these examples to overlap so much behaviorally with what's in the `14-*-video` examples. That might introduce confusion—the pattern in `14-*-video` (`request_image_frame`) is what we _actually_ recommend if you just want to grab an image from the user's camera feed, right? So instead, I'd propose that we:

1. Just remove the non-Moondream `12-*` files
2. Update the `14-*-video` examples to demonstrate usage of `LLMContext.create_image_message` in a different way—like simulating capturing an uploaded image, rather than just grabbing a capture of the camera feed
3. Add more `14-*-video` examples for the missing LLM services.

kompfner:
> > The _main_ reason for removing it was that it was a frame that had this very specific behavior of being _directly_ consumed by LLMs
> 
> I don't disagree that it would be handy for there to be an easy-to-create image frame that folks can add to context, though... (`UserImageRawFrame` in _theory_ should be enough but it has enough additional cruft that isn't always relevant)
> 
> ...but I'm not sure it should necessarily be revived under the same name as before, as `VisionImageRawFrame` previously had this very specific and quirky context-less behavior associated with it (and code might still be out there expecting that?)

Actually...maybe `VisionImageRawFrame` as a name makes sense 👍 . It's consumed _only_ by `VisionService`s (context-less one-shot image analyzers), of which there's only 1. `VisionImageRawFrame`'s old behavior of making LLMs also behave like context-less one-shot image analyzers was the thing I was hoping to move away from, and that change has already happened.

kompfner (changes requested):
The main thing for me is avoiding overlap between the `12*` and the `14*video` examples; seems like we're recommending two different patterns for doing roughly the same thing, which can lead to user confusion. I propose a fix [here](https://github.com/pipecat-ai/pipecat/pull/2935/files#r2473526004).

examples/foundational/12a-describe-video-gemini-flash.py
kompfner: Or, as an alternative to step 2, rename the `14-*-video` examples to `12-*`. Then for step 3, show both patterns (`request_image_frame` and `LLMContext.create_image_message`) in _those_ files.

The point is just to avoid recommending two patterns for the same thing. We should have a single point of view.

src/pipecat/adapters/services/bedrock_adapter.py
aconchillo: Yep. AWS Bedrock doesn't support URLs.

src/pipecat/adapters/services/gemini_adapter.py
aconchillo: Gemini doesn't support URLs either.

examples/foundational/12a-describe-video-gemini-flash.py
aconchillo: You are totally right. This is confusing.I think I have an idea. I will change the 12 series to just read an image from the disk and do the one-shot case at the very beginning, this will work for all services.

For 14-video I will change one things which is passing `client_id` as a function call argument just like I did here instead of having a global variable.

examples/foundational/12a-describe-video-gemini-flash.py
aconchillo: Yep!

src/pipecat/adapters/services/bedrock_adapter.py
markbackman: Thanks for confirming!

markbackman:
> The main thing for me is avoiding overlap between the `12*` and the `14*video` examples; seems like we're recommending two different patterns for doing roughly the same thing, which can lead to user confusion. I propose a fix [here](https://github.com/pipecat-ai/pipecat/pull/2935/files#r2473526004).

Yes, this confused me, which is why I was asking for clarification in the 12* examples. The proposed approach will work much better 🙏

aconchillo:
> > > The _main_ reason for removing it was that it was a frame that had this very specific behavior of being _directly_ consumed by LLMs
> > 
> > 
> > I don't disagree that it would be handy for there to be an easy-to-create image frame that folks can add to context, though... (`UserImageRawFrame` in _theory_ should be enough but it has enough additional cruft that isn't always relevant)
> > ...but I'm not sure it should necessarily be revived under the same name as before, as `VisionImageRawFrame` previously had this very specific and quirky context-less behavior associated with it (and code might still be out there expecting that?)
> 
> Actually...maybe `VisionImageRawFrame` as a name makes sense 👍 . It's consumed _only_ by `VisionService`s (context-less one-shot image analyzers), of which there's only 1. `VisionImageRawFrame`'s old behavior of making LLMs also behave like context-less one-shot image analyzers was the thing I was hoping to move away from, and that change has already happened.

Yes. I think `VisionImageRawFrame` are just images that might not belong to a user. You can use a vision service to handle them or even convert them and add them to the LLM context if you want. But I don't think making the vision services manage the LLM context was correct, it was a bit error prone as well IMO.

examples/foundational/12a-describe-video-gemini-flash.py
aconchillo: > You are totally right. This is confusing.I think I have an idea. I will change the 12 series to just read an image from the disk and do the one-shot case at the very beginning, this will work for all services.
> 
> For 14-video I will change one things which is passing `client_id` as a function call argument just like I did here instead of having a global variable.

Actually changing 12-* to read from a file  might not be the right thing, because then we can't really test them with our evals. I think your proposal is better. I will move 12-* to 14-*.

aconchillo:
OK, take a look again. `12-*` is now gone in favor of `14-*-video`. They all now follow the same approach, using `llm.request_image_frame()` except for `Moondream` in which case we generate the `UserImageRequestFrame` manually. In theory we could also use `llm.request_image_frame()` but that should only be used by function call requests and if the LLM is actually handling the image analysis. For Moondream we don't really want to do that because we have a parallel pipeline so the `UserImageRawFrame` would contain info related to the tool call which is what the LLM expects.

Also, an important change. Before, we had to add an `asyncio.sleep(0.5)` in each function call after requesting the user image. This was not good. The reason is because we didn't want to complete the function call right away, otherwise the image would not be processed. This is because we tie user image frame processing to function calls, which is a convenient way for the user because they don't have to do anything.

Before:

1. Function call
2. Request image from function call
3. Sleep for 0.5
4. Assistant aggregator adds image to context and marks function call as COMPLETED.
5. Function call completion. WARNING: We got a warning because (4) was marking function as completed.

Now:

1. Function call
2. Request image from function call. This now waits for an event. We are still inside the function call.
3. Assistant aggregator adds image to context. Event from (2) is set. Does NOT mark as COMPLETED.
4. Function call waiting for event continues.
5. Function call completion. We now mark the function call as COMPLETED.

examples/foundational/14d-function-calling-anthropic-video.py
kompfner, line 1: with this rename, is there no longer a 14b at all?

CHANGELOG.md
```diff
@@ -7,8 +7,28 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
 
+### Added
+
+- Added `VisionImageRawFrame`. This is an input image frame with an associated
+  text. It is usually processed by vision services (e.g. Moondream). The text
+  indicates the vision service what should it describe.
+
+- Added support for including images or audio to LLM context messages using
+  `LLMContext.create_image_message()` or `LLMContext.create_image_url_message()`
+  (not all LLM support URLs) and `LLMContext.create_audio_message()`. For
```
```suggestion
  (not all LLMs support URLs) and `LLMContext.create_audio_message()`. For
```

CHANGELOG.md
```diff
@@ -7,8 +7,28 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Added `VisionImageRawFrame`. This is an input image frame with an associated
+  text. It is usually processed by vision services (e.g. Moondream). The text
+  indicates the vision service what should it describe.
```
```suggestion
guides the vision service on how to analyze the image.
```

CHANGELOG.md
```diff
@@ -7,8 +7,28 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Added `VisionImageRawFrame`. This is an input image frame with an associated
+  text. It is usually processed by vision services (e.g. Moondream). The text
```
kompfner: Should we say what a vision service even is? Like, "a context-less, one-shot..."

kompfner:
> OK, take a look again. 12-* is now gone in favor of 14-*-video

Nice improvement! One question: didn't we want to also somehow demonstrate how the user could append image messages "manually" to context, with one of the new `LLMContext.create_image*message` functions? That's why I was thinking there'd be some tool call to add an image from disk or something like that.

examples/foundational/14d-function-calling-anthropic-video.py
aconchillo: Ah, yep. I forgot to remove that. 14b no longer needed.

CHANGELOG.md
aconchillo: Done!

CHANGELOG.md
aconchillo: Done!

CHANGELOG.md
aconchillo: Probably not needed. People can just look at the vision service documentation.

examples/foundational/14d-function-calling-moondream-video.py
aconchillo: Added!

aconchillo:
> > OK, take a look again. 12-* is now gone in favor of 14-*-video
> 
> Nice improvement! One question: didn't we want to also somehow demonstrate how the user could append image messages "manually" to context, with one of the new `LLMContext.create_image*message` functions? That's why I was thinking there'd be some tool call to add an image from disk or something like that.

Done!

markbackman (approved):
This looks great! I didn't follow the internal chatter, but in revisiting the latest changes, this makes sense. Thanks for breaking out the examples, too. This is much cleaner and shows the fundamentals in 12 + LLM use in 14 👨‍🍳 💋

examples/foundational/assets/cat.jpg
kompfner, line 1: 👍

examples/foundational/12-describe-image-openai.py
```diff
@@ -151,16 +96,28 @@ async def run_bot(transport: BaseTransport, runner_args: RunnerArguments):
...
+        # Kick off the conversation.
+        image = Image.open(image_path)
+        message = LLMContext.create_image_message(
+            image=image.tobytes(),
+            format="RGB",
+            size=image.size,
+            text=question,
+        )
+        messages.append(message)
+        await task.queue_frames([LLMRunFrame()])
```
kompfner, line 120: You could also send a `LLMMessagesAppendFrame()` with `run_llm=True` to do this in one line rather than two.

But I don't have a strong opinion about which pattern to prefer.

In a way, what you have here might be conceptually a bit clearer: you've never run the LLM before in this conversation, so the notion of "appending" to the ongoing conversation doesn't make as much sense as modifying the initial content and then saying "go".

examples/foundational/14b-function-calling-anthropic-video.py
```diff
@@ -100,70 +93,32 @@ async def run_bot(transport: BaseTransport, runner_args: RunnerArguments):
...
-        model="claude-3-7-sonnet-latest",
-        params=AnthropicLLMService.InputParams(enable_prompt_caching=True),
-    )
-    llm.register_function("get_weather", get_weather)
-    llm.register_function("get_image", get_image)
+    # Anthropic for vision analysis
+    llm = AnthropicLLMService(api_key=os.getenv("ANTHROPIC_API_KEY"))
+    llm.register_function("fetch_user_image", fetch_user_image)
 
-    weather_function = FunctionSchema(
```
kompfner, line 111: I agree with keeping these examples more focused on images.

examples/foundational/14d-function-calling-moondream-video.py
```diff
@@ -36,51 +37,55 @@
...
+    user_id = params.arguments["user_id"]
+    question = params.arguments["question"]
+    logger.debug(f"Requesting image with user_id={user_id}, question={question}")
 
-    def set_participant_id(self, participant_id: str):
-        self._participant_id = participant_id
-
-    async def process_frame(self, frame: Frame, direction: FrameDirection):
-        await super().process_frame(frame, direction)
+    # Request the user image frame frame. In this case we don't use
```
kompfner: Essentially, the Moondream mechanism in this file does the same thing as a standard `request_image_frame()` but _without_ the `request.tool_call_id` field of the resulting `UserImageRawFrame` being filled out, thus causing the LLM assistant aggregator to ignore it, right?

A few thoughts:
1. This is a really technical/nuanced thing for users to have to reason about.
2. This seems to rely on what is maybe a quirk in our LLM assistant aggregator implementation? Seems like we might someday change it to be able to accept `UserImageRawFrame`s without them being linked to specific tool calls?
3. How valuable/realistic is the functionality in this example? It might feel a little odd to the user to be talking to a bot like normal, but then for the bot to have _no conversational memory_ only when it's discussing what it's seeing...

I'm wondering if the 12- example is enough for Moondream.

src/pipecat/processors/aggregators/user_response.py
```diff
@@ -27,11 +27,24 @@ class UserResponseAggregator(LLMUserAggregator):
     def __init__(self, **kwargs):
         """Initialize the user response aggregator.
 
+        .. deprecated:: 0.0.92
```
kompfner, line 30: 👍

CHANGELOG.md
```diff
@@ -151,6 +165,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- Updated `MoondreamService` to process `VisionImageRawFrame`.
```
kompfner: Are we worried that this is a breaking change?

examples/foundational/14d-function-calling-moondream-video.py
aconchillo: Yes, totally agree, it's hard. For Moondream I also wanted to show how to do that together with an LLM in the 14 example using a parallel pipeline.

CHANGELOG.md
aconchillo: Not really :sweat_smile:

aconchillo:
> Approving, but left a couple of parting thoughts.
> 
> One is about the breaking change to Moondream. The other is about whether we ought to have the 14d Moondream example.

About Moondream being a breaking change, I think it's OK.

We do have a 14d for Moondream. I don't think a 12 example with Moondream makes sense because you can't really talk to Moondream. With the 12 examples you analyze one image initially but then you can at least talk to the LLM about it.

examples/foundational/12-describe-image-openai.py
```diff
@@ -151,16 +96,28 @@ async def run_bot(transport: BaseTransport, runner_args: RunnerArguments):
...
+
+        # Kick off the conversation.
+        image = Image.open(image_path)
+        message = LLMContext.create_image_message(
+            image=image.tobytes(),
+            format="RGB",
+            size=image.size,
+            text=question,
+        )
+        messages.append(message)
```
filipi87, line 119: Should we update the examples now to push the `UserImageRawFrame` instead, with `add_to_context=True`?

aconchillo: For this particular case, I think it's probably more natural to do it like this, since it's the first message you send to the LLM, otherwise you would do:

```
await self.task.queue_frame(UserImageRawFrame(..., add_to_context=True))
```

to start the conversation, which feels a bit weird.

aconchillo: With the current PR approach we append to messages which is how we always start and then we do:
```
await self.task.queue_frame(LLMRunFrame())
```

src/pipecat/transports/daily/transport.py
```diff
@@ -1839,10 +1839,11 @@ async def _on_participant_video_frame(
         if render_frame:
             frame = UserImageRawFrame(
                 user_id=participant_id,
-                request=request_frame,
                 image=video_frame.buffer,
                 size=(video_frame.width, video_frame.height),
                 format=video_frame.color_format,
+                text=request_frame.text if request_frame else None,
+                add_to_context=request_frame.add_to_context if request_frame else None,
```
filipi87, line 1846: I believe we need to do the same changes for the `SmallWebRTCTransport`.

filipi87 (approved):
Looks great. 🚀

examples/foundational/12-describe-image-openai.py
filipi87: Yeah, I think you are right. Makes sense. 👍

src/pipecat/transports/daily/transport.py
aconchillo: Ah, true. Let me check that.

aconchillo: Done!

PR #2937, opened 2025-10-29
Add Speechmatics TTS

aaronng91:
Add support for Speechmatics TTS and update Speechmatics examples accordingly.

### DIFF
--- a/examples/foundational/07a-interruptible-speechmatics.py
+++ b/examples/foundational/07a-interruptible-speechmatics.py
@@ -6,6 +6,7 @@
 
 import os
 
+import aiohttp
 from dotenv import load_dotenv
 from loguru import logger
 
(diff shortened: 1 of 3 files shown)


src/pipecat/services/speechmatics/tts.py
```diff
@@ -0,0 +1,202 @@
...
+
+        voice: str = "sarah"
+
+    def __init__(
+        self,
+        *,
+        api_key: str | None = None,
+        base_url: str | None = None,
+        aiohttp_session: aiohttp.ClientSession | None = None,
+        sample_rate: Optional[int] = 16000,
```
markbackman: Do you support multiple sample rates? If so, this should be set to `None`, so that Pipecat can set this value from the StartFrame. Note: the StartFrame carries the audio_input_sample_rate and audio_output_sample_rate, so that all processors can be set with the same sample rate. This avoids typos or misconfigurations.

Then, you would:
1. Set `super().__init__(sample_rate=sample_rate, **kwargs)`, which you're doing already
2. Use the `self.sample_rate` property, which comes from the TTSService base class. I see that in use already.

markbackman: I just tested and it seems that maybe only 16khz is accepted. If so, this is a fine approach to start. As the service matures and you add more sample rates, you'll want to follow the pattern I mentioned.

src/pipecat/services/speechmatics/tts.py
```diff
@@ -0,0 +1,202 @@
...
+    """
+
+    class InputParams(BaseModel):
+        """Configuration parameters for Speechmatics TTS service.
+
+        Parameters:
+            voice: Voice model to use for synthesis. Defaults to "sarah".
+        """
+
+        voice: str = "sarah"
```
markbackman: It would make sense to move the voice arg to the constructor and rename it to `voice_id` to follow the pattern from other services. Then, as you support other features, those can appear in the InputParams.

src/pipecat/services/speechmatics/tts.py
```diff
@@ -0,0 +1,202 @@
...
+                `https://preview.tts.speechmatics.com`.
+            aiohttp_session: Shared aiohttp session for HTTP requests.
+            sample_rate: Audio sample rate in Hz. Defaults to 16000.
+            params: Optional[InputParams]: Input parameters for the service.
+            **kwargs: Additional arguments passed to TTSService.
+        """
+        super().__init__(sample_rate=sample_rate, **kwargs)
+
+        # Service parameters
+        self._api_key: str = api_key or os.getenv("SPEECHMATICS_API_KEY")
```
markbackman: The pattern to follow is allowing developers to just specify their own env var name.

```suggestion
        self._api_key: str = api_key
```

Then, for the `api_key` arg, update to:
```
api_key: str,
```

Making it required.

src/pipecat/services/speechmatics/tts.py
```diff
@@ -0,0 +1,202 @@
...
+        self._base_url: str = base_url or "https://preview.tts.speechmatics.com"
```
markbackman: Nit: Update to:

```suggestion
        self._base_url: str = base_url
```

Then, for the `base_url` arg, update it to:
```
base_url: str = "https://preview.tts.speechmatics.com",
```

src/pipecat/services/speechmatics/tts.py
```diff
@@ -0,0 +1,202 @@
...
+        Parameters:
+            voice: Voice model to use for synthesis. Defaults to "sarah".
+        """
+
+        voice: str = "sarah"
+
+    def __init__(
+        self,
+        *,
+        api_key: str | None = None,
```
markbackman: Required. See the comment below.

```suggestion
        api_key: str,
```

src/pipecat/services/speechmatics/tts.py
```diff
@@ -0,0 +1,202 @@
...
+        base_url: str | None = None,
```
markbackman: Required. See the comment below.

```suggestion
        base_url: str = "https://preview.tts.speechmatics.com",
```

src/pipecat/services/speechmatics/tts.py
```diff
@@ -0,0 +1,202 @@
...
+        aiohttp_session: aiohttp.ClientSession | None = None,
```
markbackman: Required.

```suggestion
        aiohttp_session: aiohttp.ClientSession,
```

src/pipecat/services/speechmatics/tts.py
```diff
@@ -0,0 +1,202 @@
...
+        sample_rate: Optional[int] = 16000,
+        params: InputParams | None = None,
```
markbackman: Should be:

```suggestion
        params: Optional[InputParams] = None,
```

src/pipecat/services/speechmatics/tts.py
```diff
@@ -0,0 +1,202 @@
...
+            sample_rate: Audio sample rate in Hz. Defaults to 16000.
+            params: Optional[InputParams]: Input parameters for the service.
+            **kwargs: Additional arguments passed to TTSService.
+        """
+        super().__init__(sample_rate=sample_rate, **kwargs)
+
+        # Service parameters
+        self._api_key: str = api_key or os.getenv("SPEECHMATICS_API_KEY")
+        self._base_url: str = base_url or "https://preview.tts.speechmatics.com"
+        self._session = aiohttp_session or aiohttp.ClientSession()
```
markbackman: Aligning with the change above.

```suggestion
        self._session = aiohttp_session
```

src/pipecat/services/speechmatics/tts.py
markbackman: Maybe the best option for now is setting the arg to:
```
sample_rate: Optional[int] = SPEECHMATICS_SAMPLE_RATE,
```

Defining a const:
```
SPEECHMATICS_SAMPLE_RATE = 16000
```

Warnings when changing the value:
```
if sample_rate and sample_rate != self.SPEECHMATICS_SAMPLE_RATE:
            logger.warning(
                f"Speechmatics TTS only supports {self.SPEECHMATICS_SAMPLE_RATE}Hz sample rate. "
                f"Current rate of {sample_rate}Hz may cause issues."
            )
```

Then, when you support a sample rate, we remove the const, warning and set the arg default to `None`.

src/pipecat/services/speechmatics/tts.py
```diff
@@ -0,0 +1,202 @@
...
+
+        # Check we have required attributes
+        if not self._api_key:
+            raise ValueError("Missing Speechmatics API key")
+        if not self._base_url:
```
markbackman: We can remove this check if you agree that we should provide a default base_url in the `base_url` arg.

src/pipecat/services/speechmatics/tts.py
```diff
@@ -0,0 +1,202 @@
...
+                        first_chunk = False
+
+                    buffer += chunk
+
+                    # Emit a frame for all complete samples currently in buffer
+                    frame = _emit_complete_samples()
+                    if frame:
+                        yield frame
+
+                # Process any remaining bytes in buffer after streaming ends
```
markbackman: Is this needed? I did a bit of testing and see that all chunks appear to be emitted and this logic doesn't ever result in yielding a frame.

If it's not needed, we can simplify the code to something like this:
```
# Process the response in streaming chunks
first_chunk = True
buffer = b""

async for chunk in response.content.iter_any():
    if not chunk:
        continue
    if first_chunk:
        first_chunk = False

    buffer += chunk

    # Emit all complete 2-byte int16 samples from buffer
    if len(buffer) >= 2:
        complete_samples = len(buffer) // 2
        complete_bytes = complete_samples * 2

        audio_data = buffer[:complete_bytes]
        buffer = buffer[complete_bytes:]  # Keep remaining bytes for next iteration

        yield TTSAudioRawFrame(
            audio=audio_data,
            sample_rate=self.sample_rate,
            num_channels=1,
        )
```

I tested this exact thing and it seems to work well.

examples/foundational/07a-interruptible-speechmatics.py
```diff
@@ -87,10 +92,11 @@ async def run_bot(transport: BaseTransport, runner_args: RunnerArguments):
         ),
     )
 
-    tts = ElevenLabsTTSService(
-        api_key=os.getenv("ELEVENLABS_API_KEY"),
-        voice_id=os.getenv("ELEVENLABS_VOICE_ID"),
-        model="eleven_turbo_v2_5",
+    tts = SpeechmaticsTTSService(
```
markbackman: With the change suggested, you'll need to update the example to pass an aiohttp session:

```
        stt = SpeechmaticsSTTService(
            api_key=os.getenv("SPEECHMATICS_API_KEY"),
            params=SpeechmaticsSTTService.InputParams(
                language=Language.EN,
                enable_diarization=True,
                end_of_utterance_silence_trigger=0.5,
                speaker_active_format="<{speaker_id}>{text}</{speaker_id}>",
            ),
        )

        tts = SpeechmaticsTTSService(
            api_key=os.getenv("SPEECHMATICS_API_KEY"),
            params=SpeechmaticsTTSService.InputParams(
                voice="sarah",
            ),
            aiohttp_session=session,
        )

        llm = OpenAILLMService(
            api_key=os.getenv("OPENAI_API_KEY"),
            params=BaseOpenAILLMService.InputParams(temperature=0.75),
        )
```

src/pipecat/services/speechmatics/tts.py
aaronng91: We only support 16khz at the moment, but we do intend to offer other sample rates in the future. I have changed this logic to reflect your suggestion.

src/pipecat/services/speechmatics/tts.py
aaronng91: Good point, I've changed it.

src/pipecat/services/speechmatics/tts.py
aaronng91: Yep agree, I've removed it.

src/pipecat/services/speechmatics/tts.py
aaronng91: Yes, that's a good point. I don't think its needed. I've updated this based on your suggestion.

aaronng91:
Thanks for the detailed review! I have made all the corresponding changes.

PR #2943, opened 2025-10-30
Add DeepgramHttpTTSService

markbackman:

This is a refactored version of https://github.com/pipecat-ai/pipecat/pull/2711.

I do see a significant reduction in latency—I see 200-300ms vs 1-1.75sec. I tripled checked that the TTFB is in the right place and empirically, the TTS is noticeably faster.

This _is_ a breaking change though, as it requires an `aiohttp_session` now. There's no good way around that requirement though. We'd have to introduce a new `DeepgramHttpTTSService` class and deprecate this class. That is possible, but that's even more disruptive than having to add the additional arg.

If we want to move forward with this, ideally, we would make these changes in the original PR. I've notified the author. But, we should decide about the breaking change here.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- Added a new `DeepgramHttpTTSService`, which delivers a meaningful reduction
+  in latency when compared to the `DeepgramTTSService`.
+
 - Add support for `speaking_rate` input parameter in `GoogleHttpTTSService`.
 
 - Added `enable_speaker_diarization` and `enable_language_identification` to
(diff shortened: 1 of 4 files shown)


aconchillo:
> This _is_ a breaking change though, as it requires an `aiohttp_session` now. There's no good way around that requirement though. We'd have to introduce a new `DeepgramHttpTTSService` class and deprecate this class. That is possible, but that's even more disruptive than having to add the additional arg.
> 

I think keeping the current class and adding `DeepgramHttpTTSService` feels better, it's what we have in other cases. If you want the fast version, just use the HTTP version, at least for now. I'm sure this will change in the future and then people can go back and use `DeepgramTTSService`.

markbackman:
> I think keeping the current class and adding DeepgramHttpTTSService feels better, it's what we have in other cases. If you want the fast version, just use the HTTP version, at least for now. I'm sure this will change in the future and then people can go back and use DeepgramTTSService.

@aconchillo that makes sense. I've made this update. This is cleaner.

aconchillo (approved):

aconchillo:
LGTM!

PR #2946, opened 2025-10-30
Deprecate `expect_stripped_words` option from `LLMAssistantAggregatorParams`…

kompfner:
…when used with the newer `LLMAssistantAggregator`, which now handles word spacing automatically.

This commit does not change how it works in the older `LLMAssistantContextAggregator`.

### DIFF
--- a/tests/test_context_aggregators.py
+++ b/tests/test_context_aggregators.py
@@ -6,7 +6,7 @@
 
 import json
 import unittest
-from typing import Any
+from typing import Any, Optional
 
 from pipecat.audio.interruptions.min_words_interruption_strategy import MinWordsInterruptionStrategy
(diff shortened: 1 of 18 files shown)


tests/test_context_aggregators.py
```diff
@@ -948,8 +956,27 @@ class TestOpenAIUserContextAggregator(
 
 
 class TestOpenAIAssistantContextAggregator(
-    BaseTestAssistantContextAggreagator, unittest.IsolatedAsyncioTestCase
+    BaseTestAssistantContextAggregator, unittest.IsolatedAsyncioTestCase
```
aconchillo, line 959: Ooops :sweat_smile:

aconchillo:
Very nice!

PR #2947, opened 2025-10-30
UserImageRawFrame: rename add_to_context to append_to_context

aconchillo:

Just a rename...

### DIFF
--- a/src/pipecat/transports/smallwebrtc/transport.py
+++ b/src/pipecat/transports/smallwebrtc/transport.py
@@ -661,6 +661,8 @@ class SmallWebRTCInputTransport(BaseInputTransport):
                     # UserImageRawFrame. Use a shallow copy so we can remove
                     # elements.
                     for request_frame in self._image_requests[:]:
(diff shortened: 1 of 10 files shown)


src/pipecat/transports/smallwebrtc/transport.py
```diff
@@ -661,17 +661,17 @@ async def _receive_video(self, video_source: str):
                     # UserImageRawFrame. Use a shallow copy so we can remove
                     # elements.
                     for request_frame in self._image_requests[:]:
+                        request_text = request_frame.text if request_frame else None
+                        add_to_context = request_frame.append_to_context if request_frame else None
```
aconchillo, line 665: I just use a variable `add_to_context` here so it fits in one line :sweat_smile:

kompfner (approved):

src/pipecat/transports/smallwebrtc/transport.py
mattieruth: i like your style

PR #2949, opened 2025-10-31
PipelineTask: add IdleFrameObserver to detect idle pipelines

aconchillo:

We the following report:

- `idle_timeout_frames` set to `LLMFullResponseEndFrame` 
- Frame `LLMFullResponseEndFrame`  was being generated
- Pipeline was considered idle.

The reason is because the LLM assistant aggregator swallows `LLMFullResponseEndFrame`.

Since we don't know when frames can be swallowed we now use an observer and an asyncio event that gets set when the expected frames are generated. This also simplifies the timeout logic since we just rely on `asyncio.wait_for()`.

### DIFF
--- a/tests/test_pipeline.py
+++ b/tests/test_pipeline.py
@@ -24,6 +24,7 @@ from pipecat.pipeline.base_task import PipelineTaskParams
 from pipecat.pipeline.parallel_pipeline import ParallelPipeline
 from pipecat.pipeline.pipeline import Pipeline
 from pipecat.pipeline.task import PipelineParams, PipelineTask
+from pipecat.processors.filters.frame_filter import FrameFilter
 from pipecat.processors.filters.identity_filter import IdentityFilter
 from pipecat.processors.frame_processor import FrameDirection, FrameProcessor
 from pipecat.tests.utils import HeartbeatsObserver, run_test
@@ -383,6 +384,7 @@ class TestPipelineTask(unittest.IsolatedAsyncioTestCase):
         idle_timeout_secs = 0.2
         sleep_time_secs = idle_timeout_secs / 2
 
+        # Use the identify filter so the frames just reach the end of the pipeline.
         identity = IdentityFilter()
         pipeline = Pipeline([identity])
         task = PipelineTask(
@@ -392,6 +394,12 @@ class TestPipelineTask(unittest.IsolatedAsyncioTestCase):
         )
 
         async def delayed_frames():
+            """Sending multiple text frames.
+
+            The total amount of elapsed time in this function should be greater
+            than the task idle timeout. If an idle timeout event is triggered it
+            means we haven't detected that the TextFrames have been pushed.
+            """
             await asyncio.sleep(sleep_time_secs)
             await task.queue_frame(TextFrame("Hello Pipecat!"))
             await asyncio.sleep(sleep_time_secs)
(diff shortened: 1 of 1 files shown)


src/pipecat/pipeline/task.py
```diff
@@ -772,33 +811,10 @@ async def _idle_monitor_handler(self):
...
-                        running = await self._idle_timeout_detected()
-                        # Reset `last_frame_time` so we don't trigger another
-                        # immediate idle timeout if we are not cancelling. For
-                        # example, we might want to force the bot to say goodbye
-                        # and then clean nicely with an `EndFrame`.
-                        last_frame_time = time.time()
-
-                self._idle_queue.task_done()
-
+                await asyncio.wait_for(self._idle_event.wait(), timeout=self._idle_timeout_secs)
```
kompfner, line 816: much tidier!

src/pipecat/pipeline/task.py
```diff
@@ -810,7 +826,7 @@ async def _idle_timeout_detected(self) -> bool:
         """
         # If we are cancelling, just exit the task.
         if self._cancelled:
-            return True
+            return False
```
kompfner, line 829: What effect does flipping this value have?

kompfner: It seems like the return value of `_idle_timeout_detected` determines whether to stop the pipeline (based on the assumption that the `_idle_monitor_task` is the only thing keeping the pipeline alive still...right? what if that's not true?).

If we're already mid-cancelling the pipeline when we detect that the pipeline is idle...then now we're saying we keep the `_idle_monitor_task` alive and keep monitoring for future idle detections?

I'm probably misunderstanding this mechanism.

kompfner: Oh, this mechanism is just for detecting a pipeline that's _not being used_, not one that's stuck/blocked for some reason. Got it.

But wait, shouldn't we actually return `True` here, to tell the `_idle_monitor_task` to stop? We don't need it anymore if we're already canceling the pipeline, right?

Side note: the comments in this method state that the return value indicates whether the pipeline task should continue running, which doesn't seem accurate—it indicates whether the idle timeout monitor should continue, right?

kompfner: > We don't need it anymore if we're already canceling the pipeline, right?

Hmm...is the thinking that if cancelling takes a long time and `_cancel_on_idle_timeout` is `False` the user might still want to get an "on_idle_timeout" event mid-cancellation? 🤔

kompfner: I suppose I can see this value going either way...unclear to me what the right behavior is in this edge case.

One piece of feedback, though, is to fix the comments in this method to better describe the return value.

aconchillo: If it's `True` we keep running the task. If it's `False` then `running` becomes `False` so we exit.

aconchillo: > > We don't need it anymore if we're already canceling the pipeline, right?
> 
> Hmm...is the thinking that if cancelling takes a long time and `_cancel_on_idle_timeout` is `False` the user might still want to get an "on_idle_timeout" event mid-cancellation? 🤔

If we cancel we will likely not get the frames we are expecting, but I don't think we want to trigger an idle timeout in that case, we just don't want to keep checking for the idle timeout frames.

kompfner: Ah right `False` means stop. OK I got turned around. This makes sense. No need to keep running the timeout checker. 👍

PR #2959, opened 2025-11-01
CancelFrame: add reason field to indicate why pipeline is being cancelled

aconchillo:

This PR adds a `reason` field to the `CancelFrame`. We might be cancelling the pipeline for different reasons and this allows us to save that reason when, for example, we get a `on_pipeline_finished` event.


aconchillo:
Also added `reason` to `CancelTaskFrame` (which is then assigned to `CancelFrame`) and also to `EndFrame` and `EndTaskFrame`.

PR #2960, opened 2025-11-02
Add Typecast HTTP TTS service and foundational example

sangwonl:
## Summary

- Add an HTTP-based `TypecastTTSService` aligned with the official [Typecast Text-to-Speech REST API](https://typecast.ai/docs/api-reference/endpoint/text-to-speech/text-to-speech)
- Expose the integration via `pipecat.services.typecast`, include a `typecast` optional dependency, and document new env vars for credentials
- Add a foundational interruptible example that pairs Deepgram STT, OpenAI LLM, and Typecast TTS

### Typecast TTS Service

- Implements `src/pipecat/services/typecast/tts.py` using a shared `aiohttp` client; handles ISO 639-3 language mapping plus `prompt`/`output` controls (emotion, intensity, pitch, tempo, format) as documented by Typecast
- Streams WAV audio through the shared `_stream_audio_frames_from_iterator` helper, integrates `start_ttfb_metrics`/`start_tts_usage_metrics`, and enforces the `audio_format="wav"` contract for raw PCM playback
- Re-exports the service from `src/pipecat/services/typecast/__init__.py` so `pipecat.services.typecast` mirrors other provider namespaces

### Foundational Example

- Adds `examples/foundational/07z-interruptible-typecast-http.py`, matching other HTTP samples with configurable transports, Deepgram STT, OpenAI LLM, and optional `TYPECAST_VOICE_ID` overrides
- Manages all HTTP work inside a single `aiohttp.ClientSession`, highlighting how to reuse the service within an interruptible pipeline

### Configuration & Packaging

- Registers a `typecast` optional dependency group inside `pyproject.toml` for targeted installs
- Updates `env.example` with `TYPECAST_API_KEY`/`TYPECAST_VOICE_ID` so users know which secrets are required before running the sample or service

## Usage Example

```python
    tts = TypecastTTSService(
        api_key=os.environ["TYPECAST_API_KEY"],
        aiohttp_session=session,
        params=TypecastTTSService.InputParams(
            prompt_options=PromptOptions(emotion_preset="happy", emotion_intensity=1.2),
        ),
    )
```


markbackman:
We have a community integrations program:

Do you work for the company that builds Typecast? If so, adding the community integration could be a great solution to adding support to Pipecat. Please let me know if you have any questions.

markbackman:
@sangwonl just making sure you saw my note above about adding Typecast to the Community Integrations program. Please check it out and let me know if you have any questions.

sangwonl:
Hello, @markbackman. Sorry for the late reply. When you say to add it as a community integration, do you mean simply getting it listed on [the page](https://docs.pipecat.ai/server/services/community-integrations)? I was wondering — is there any way it could be added as an official Pipecat service instead?

Thanks :)

markbackman:
> Sorry for the late reply. When you say to add it as a community integration, do you mean simply getting it listed on [the page](https://docs.pipecat.ai/server/services/community-integrations)?

Yes, that's the right page.

> is there any way it could be added as an official Pipecat service instead?

Integrations take a lot of our time to maintain, so we're being really picky about new integrations for a bit. Starting as a community integration is a good path to get visibility and can still provide a path for a first-party integration that we help maintain, as the community demands inclusion into Pipecat. Does that make sense?

sangwonl:
Thank you for the detailed explanation. That sounds like a reasonable approach. I’ll go ahead and try it as a community integration.

In that case, should I leave this PR open so it could be used later if there’s community demand for the 1st-party inclusion you mentioned, or would you prefer that I close it?

markbackman:
I'll close this out now since you're listed as a community integration. We can always reference this in the future if we want to port this to first party. Thanks again!

PR #2961, opened 2025-11-03
Fix Gemini Live session resumption. The problem was that we weren't p…

kompfner:
…roperly ignoring send errors during reconnection.

(These are "expected" errors that occur for messages that were being sent when the disconnection occurred)

### DIFF
--- a/src/pipecat/services/google/gemini_live/llm.py
+++ b/src/pipecat/services/google/gemini_live/llm.py
@@ -1717,13 +1717,17 @@ class GeminiLiveLLMService(LLMService):
             self._session_resumption_handle = update.new_handle
 
     async def _handle_send_error(self, error: Exception):
+        # Ignore "expected" errors that may have occurred for messages that
+        # were in-flight when a disconnection occurred.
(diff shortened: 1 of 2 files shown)


src/pipecat/services/google/gemini_live/llm.py
```diff
@@ -1717,13 +1717,15 @@ def _handle_msg_resumption_update(self, message: LiveServerMessage):
             self._session_resumption_handle = update.new_handle
 
     async def _handle_send_error(self, error: Exception):
+        if self._disconnecting or not self._session:
```
kompfner, line 1722: This was really a one-line fix, adding `not self._session` to the condition we're checking for here.

kompfner: This condition—`self._disconnecting or not self._session`—is one we check for all over this file. It essentially means "either we're in the process of disconnecting or currently disconnected".

kompfner:
Ah, forgot the CHANGELOG entry. Will add it before merging.

PR #2964, opened 2025-11-03
Fix 14j foundational example

markbackman:

I'm not sure what the issue was with the meta/llama3.3 model, but it seems to be a bug outside of Pipecat. In switching to the nemotron models, I have no issues running this example.

The nemotron model is great because it's fast. We just need to disable thinking for it, which I show in the example.

### DIFF
--- a/examples/foundational/14j-function-calling-nim.py
+++ b/examples/foundational/14j-function-calling-nim.py
@@ -75,7 +75,12 @@ async def run_bot(transport: BaseTransport, runner_args: RunnerArguments):
         # text_filters=[MarkdownTextFilter()],
     )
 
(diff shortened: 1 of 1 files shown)


kompfner (approved):

kompfner:
One thought: did we want to say in a comment (or somewhere) that there's a known issue with "meta/llama3.3" in NIM?

markbackman:
> One thought: did we want to say in a comment (or somewhere) that there's a known issue with "meta/llama3.3" in NIM?

Maybe, but it's possible that problem will be resolved; we won't be retesting. I left it out because it felt out of scope for what this example is trying to accomplish.

PR #2970, opened 2025-11-04
Assorted tool registration improvements

kompfner:
Some testing still in progress, but wanted to open this up for review.

## Summary

- Update the service switcher example to illustrate registering tools
- Add `LLMSwitcher.register_direct_function()`
- Move schema massaging logic from `MCPClient` to `GeminiLLMAdapter` instead, making `GeminiLLMAdapter` more robust and making `MCPClient` LLM-agnostic (important for the next point)
- Update `MCPClient` to be able to register tools on `LLMSwitcher`s
- Provide a 2-step alternative to `MCPClient.register_tools()`, which lets you pass MCP tools to, say, `GeminiLiveLLMService` in the constructor
  - `MCPClient.get_tools_schema()`
  - `MCPClient.register_tools_schema()`
- Fixed a `GeminiLiveLLMService` bug discovered along the way that caused it to not respond after some tool calls
- Fixed a `GeminiLiveLLMService` where if only *one* of tools or system instruction was provided in the context, the other wouldn't fall back to using the value provided in the constructor (this fix isn't in the CHANGELOG since `GeminiLiveLLMService`'s ability to properly handle context-provided tools and system instruction hasn't been published yet anyway)
- Fixed an issue where `GeminiLiveLLMService` wasn't consistent in what it would do if if it received an `LLMContextFrame` (triggered by an `LLMRunFrame`, say) and there were no user messages in the initial context:
  - If the context contained a system message, that message would be converted to a user message and the LLM would respond
  - If the system message was provided as a constructor argument, though, no user messages would be sent to the LLM, and the LLM would therefore not respond

  (this fix isn't in the CHANGELOG since `GeminiLiveLLMService`'s ability to properly handle context-provided tools and system instruction hasn't been published yet anyway)

### DIFF
--- a/examples/foundational/39d-mcp-run-http-gemini-live.py
+++ b/examples/foundational/39d-mcp-run-http-gemini-live.py
@@ -107,7 +107,7 @@ async def run_bot(transport: BaseTransport, runner_args: RunnerArguments):
 
     await mcp.register_tools_schema(tools, llm)
 
-    context = LLMContext([{"role": "user", "content": "Please introduce yourself."}], NOT_GIVEN)
+    context = LLMContext([{"role": "user", "content": "Please introduce yourself."}])
     context_aggregator = LLMContextAggregatorPair(context)
 
     pipeline = Pipeline(


kompfner:
Fun fact: in testing `GeminiLiveLLMService` with mcp.run SSE, I discovered that Gemini Live does not support hyphens in function names, which mcp.run uses.

CHANGELOG.md
```diff
@@ -28,20 +28,41 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - `CancelFrame` and `CancelTaskFrame` have an optional `reason` field to
   indicate why the pipeline is being canceled. This can be also specified when
-  you cancel a task with `PipelineTask.cancel(reason="cancellation your
-reason")`.
+  you cancel a task with
+  `PipelineTask.cancel(reason="cancellation your reason")`.
+
+- Added support for passing in an `LLMSwicher` to `MCPClient.register_tools()`
+  (as well as the new `MCPClient.register_tools_schema()`).
```
vipyne: nit: maybe add this after line 42 ?

examples/foundational/39d-mcp-run-http-gemini-live.py
```diff
@@ -0,0 +1,165 @@
...
+
+    llm = GeminiLiveLLMService(
+        api_key=os.getenv("GOOGLE_API_KEY"),
+        system_instruction=system,
+        tools=tools,
+    )
+
+    await mcp.register_tools_schema(tools, llm)
+
+    context = LLMContext([{"role": "user", "content": "Please introduce yourself."}], NOT_GIVEN)
```
vipyne: is `NOT_GIVEN` necessary here?  I thought it was default.
it feels a little odd since there _are_ tools from the mcp server.
I get that tools are registered in the line above, but the `NOT_GIVEN` throws me off.
maybe I am missing something here.

kompfner: You're right, it's not necessary, and just adds confusion. Will remove.

src/pipecat/services/mcp_service.py
```diff
@@ -56,75 +57,67 @@ def __init__(
...
-            self._register_tools = self._streamable_http_register_tools
+            self._list_tools = self._streamable_http_list_tools
+            self._tool_wrapper = self._streamable_http_tool_wrapper
         else:
             raise TypeError(
                 f"{self} invalid argument type: `server_params` must be either StdioServerParameters, SseServerParameters, or StreamableHttpParameters."
             )
 
-    async def register_tools(self, llm) -> ToolsSchema:
+    async def register_tools(self, llm: LLMService | LLMSwitcher) -> ToolsSchema:
```
aconchillo, line 78: Since we have removed `needs_mcp_alternate_schema()` is it possible to remove the `llm` argument? We should probably rename `register_tools()` to `get_tools()` or something. This way we decouple the MCP from the LLM.

aconchillo: If I understand correctly, we only need the `llm` to be able to do `register_function()`, if so, we should do that from the outside like we usually do.

src/pipecat/services/mcp_service.py
```diff
@@ -56,75 +57,67 @@ def __init__(
...
+    ) -> None:
+        """Register the MCP tools (previously obtained from get_tools_schema()) with the LLM service.
 
-        return alternate_schema
+        Args:
+            tools_schema: The ToolsSchema to register with the LLM service.
+            llm: The Pipecat LLM service to register tools with.
+        """
+        for function_schema in tools_schema.standard_tools:
+            llm.register_function(function_schema.name, self._tool_wrapper)
```
aconchillo, line 120: Shouldn't we do that outside? We could have a `register_tools_schema` in the base LLMService class that does the same.

```
tools = mcp.get_tools_schema()
llm.register_tools_schema(tools, mcp.tool_handler())
```

or something like that.

src/pipecat/services/mcp_service.py
kompfner: > Since we have removed needs_mcp_alternate_schema() is it possible to remove the llm argument? We should probably rename register_tools() to get_tools() or something

Adding `get_tools()` (actually I named it `get_tools_schema()`) is precisely what we do in this PR. This `register_tools()` remains here as a convenience since it does everything—getting the available tools and registering them—in a single step.

> we only need the llm to be able to do register_function(), if so, we should do that from the outside like we usually do

It's actually not so straightforward. "From the outside" you don't have the tool functions themselves to register with the LLM service. You need the `MCPClient` to provide the function wrapper that you register for each tool. So I've added a convenience for the registration step—`register_tools_schema()`—if you want to do tool getting + registering as 2 distinct steps instead of the older single-step `get_tools()`.

Does that make sense?

src/pipecat/services/mcp_service.py
kompfner: We _could_ do it from the outside, but it feels a bit odd to have a convenience method in `LLMService` to register _the same function_ for every tool in a schema—that feels like an unusual thing to do. It's a very MCP-specific thing to do, though (hence it being in `MCPClient`).

aconchillo (approved):

src/pipecat/services/mcp_service.py
aconchillo: Yes, I think you are right. :ship: it!

PR #2972, opened 2025-11-04
Add customer parameter in Call Data for Exotel

akash-dutta-dev:
Add customer parameter in Call Data for Exotel


akash-dutta-dev:
Refer to document 
https://support.exotel.com/support/solutions/articles/3000108630-working-with-the-stream-and-voicebot-applet
Section - Start message

markbackman:
This makes sense! Thank you.

PR #2974, opened 2025-11-04
Improve language checking in STT and TTS services

markbackman:

Two patterns exist:
1. Services expecting 2-letter base codes (Cartesia, ElevenLabs, LMNT, etc.): Check the map, extract base code from variants (e.g., en-US → en), warn if not verified, and pass through the base code.
2. Services expecting full locale codes or custom formats (AWS, Google, Azure, PlayHT, Rime, etc.): Check the map, warn if not verified, and pass through the original language value as-is.

Adding utility functions to centralize this fallback logic. This change makes services more flexible to work with but still gets the benefits of checking against supported languages and warning when a language isn't found.

I'm adding this because it's easy for service providers to add support for new languages, but Pipecat code be behind in adding new languages. Without this, you can get a failure is currently, `None` is returned. Now, the original value falls through along with a warning.

### DIFF
--- a/src/pipecat/services/speechmatics/stt.py
+++ b/src/pipecat/services/speechmatics/stt.py
@@ -31,7 +31,7 @@ from pipecat.frames.frames import (
 )
 from pipecat.processors.frame_processor import FrameDirection
 from pipecat.services.stt_service import STTService
-from pipecat.transcriptions.language import Language
+from pipecat.transcriptions.language import Language, resolve_language
 from pipecat.utils.tracing.service_decorators import traced_stt
 
 try:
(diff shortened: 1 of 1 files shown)


src/pipecat/transcriptions/language.py
kompfner, line 1: A few thoughts:

- In `fallback_to_base_code()` shouldn't we always emit a warning, like we do in `fallback_to_full_code()`? We're already falling back. Either the user has tried to configure a service with a full language code (e.g. "en-US") when it only supports base codes (e.g. "en") or they've provided an unverified base code (or both—a full code with an unverified base component).
- We might want to broaden the `fallback` methods into `resolve` methods (or maybe just a single `resolve` method), so that they also handle checking the map for the user-provided language before falling back. We have a lot of logic scattered about that looks like:
  ```python
  # Check
  result = LANGUAGE_MAP.get(language)

  # Fall back
  if not result:
    result = fallback_to_base_code(language, LANGUAGE_MAP)
  ```
  That could be simplified to:
  ```python
  result = resolve_language(language, LANGUAGE_MAP, fallback=BaseCode) # BaseCode | FullCode
  ```

markbackman: I had debated a single utility function and at first wanted to keep it simple. But, I think you're right. It's worth centralizing this logic and then just having a single place where the language is checked. I made this update.

src/pipecat/services/speechmatics/stt.py
```diff
@@ -604,11 +604,21 @@ def _process_config(self) -> None:
         Creates a transcription config object based on the service parameters. Aligns
         with the Speechmatics RT API transcription config.
         """
+        # Convert language if it's a Language enum
+        language = self._params.language
+        if isinstance(language, Language):
+            language = _language_to_speechmatics_language(language)
```
markbackman, line 610: The language conversation utility functions weren't previously used. This uses them and aligns with the other STT and TTS services.

src/pipecat/services/whisper/stt.py
```diff
@@ -106,159 +106,70 @@ def language_to_whisper_language(language: Language) -> Optional[str]:
     Note:
         Only includes languages officially supported by Whisper.
     """
-    language_map = {
+    LANGUAGE_MAP = {
```
kompfner, line 109: 👍

kompfner (approved):

PR #2976, opened 2025-11-04
fix interruption task frame context ordering

aconchillo:

When we manually interrupt a pipeline using `push_interruption_task_frame_and_wait()` the context ends up out of order, that is, the user content is added before the assistant content.

This is because of how manual interruption works:

1.  The user said, for example, 2 words and we consider we should interrupt the bot.
2. We push an `InterruptionTaskFrame` upstream from processor A.
3. We wait for the `InterruptionFrame` in processor A.
4. When we get the `InterruptionFrame` in processor A, we append `user` content to the context.
5. When we get the `InterruptionFrame` in the assistant aggregator we appen `assistant` content (leftover) t the context.

Note how (5) should be happening before (4).

In this PR we fix this by makig sure `InterruptionFrame` reached (5) before (4).

### DIFF
--- a/tests/test_context_aggregators.py
+++ b/tests/test_context_aggregators.py
@@ -4,6 +4,7 @@
 # SPDX-License-Identifier: BSD 2-Clause License
 #
 
+import asyncio
 import json
 import unittest
 from typing import Any, Optional
@@ -30,17 +31,23 @@ from pipecat.frames.frames import (
     SpeechControlParamsFrame,
     TextFrame,
     TranscriptionFrame,
+    TTSTextFrame,
     UserStartedSpeakingFrame,
     UserStoppedSpeakingFrame,
 )
 from pipecat.pipeline.pipeline import Pipeline
 from pipecat.pipeline.task import PipelineParams
+from pipecat.processors.aggregators.llm_context import LLMContext
 from pipecat.processors.aggregators.llm_response import (
     LLMAssistantAggregatorParams,
+    LLMAssistantContextAggregator,
     LLMUserAggregatorParams,
     LLMUserContextAggregator,
 )
-from pipecat.processors.aggregators.llm_response_universal import LLMAssistantAggregator
+from pipecat.processors.aggregators.llm_response_universal import (
+    LLMAssistantAggregator,
+    LLMUserAggregator,
+)
 from pipecat.processors.aggregators.openai_llm_context import (
     OpenAILLMContext,
     OpenAILLMContextFrame,
@@ -73,8 +80,11 @@ AGGREGATION_SLEEP = 0.15
 
 class BaseTestUserContextAggregator:
     CONTEXT_CLASS = None  # To be set in subclasses
-    AGGREGATOR_CLASS = None  # To be set in subclasses
-    EXPECTED_CONTEXT_FRAMES = [OpenAILLMContextFrame]
+    CONTEXT_FRAME_CLASS = None  # To be set in subclasses
+    USER_AGGREGATOR_CLASS = None  # To be set in subclasses
+    USER_EXPECTED_CONTEXT_FRAMES = None
+    ASSISTANT_AGGREGATOR_CLASS = None  # To be set in subclasses
+    ASSISTANT_EXPECTED_CONTEXT_FRAMES = None  # To be set in subclasses
 
     def check_message_content(self, context: OpenAILLMContext, index: int, content: str):
         assert context.messages[index]["content"] == content
@@ -86,10 +96,12 @@ class BaseTestUserContextAggregator:
 
     async def test_se(self):
         assert self.CONTEXT_CLASS is not None, "CONTEXT_CLASS must be set in a subclass"
-        assert self.AGGREGATOR_CLASS is not None, "AGGREGATOR_CLASS must be set in a subclass"
+        assert self.USER_AGGREGATOR_CLASS is not None, (
+            "USER_AGGREGATOR_CLASS must be set in a subclass"
+        )
 
         context = self.CONTEXT_CLASS()
-        aggregator = self.AGGREGATOR_CLASS(context)
+        aggregator = self.USER_AGGREGATOR_CLASS(context)
         frames_to_send = [UserStartedSpeakingFrame(), UserStoppedSpeakingFrame()]
         expected_down_frames = [UserStartedSpeakingFrame, UserStoppedSpeakingFrame]
         await run_test(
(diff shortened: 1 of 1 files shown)


src/pipecat/frames/frames.py
```diff
@@ -859,9 +860,13 @@ class InterruptionFrame(SystemFrame):
     speaking (i.e. is interrupting). This is similar to
     UserStartedSpeakingFrame except that it should be pushed concurrently
     with other frames (so the order is not guaranteed).
+
+    Parameters:
+        wait_event: If not None, the event will be set when the frame
```
kompfner: nit: `finished_event`

src/pipecat/processors/frame_processor.py
```diff
@@ -661,31 +655,27 @@ async def push_frame(self, frame: Frame, direction: FrameDirection = FrameDirect
 
         await self._call_event_handler("on_after_push_frame", frame)
 
-        # If we are waiting for an interruption and we get an interruption, then
-        # we can unblock `push_interruption_task_frame_and_wait()`.
-        if self._wait_for_interruption and isinstance(frame, InterruptionFrame):
-            self._wait_interruption_event.set()
-
     async def push_interruption_task_frame_and_wait(self):
```
kompfner, line 658: Do we _always_ want the new behavior with `push_interruption_task_frame_and_wait()`? It seems like the fix in this PR would help with the case in `LLMUserAggregator`, but does it make sense to apply universally? (Maybe it does—just raising the question)

CHANGELOG.md
```diff
@@ -29,10 +35,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `CancelFrame` and `CancelTaskFrame` have an optional `reason` field to
   indicate why the pipeline is being canceled. This can be also specified when
   you cancel a task with `PipelineTask.cancel(reason="cancellation your
-reason")`.
+  reason")`.
```
kompfner: Prettier in VS Code will automatically always undo this change, which is probably why it changed (and will probably keep changing as VS Code users touch this file). It doesn't like the code expression being split over two lines. In another PR I fix this by putting the expression in its own line.

markbackman (approved):
This looks good and works well.

Just to clarify my understanding, should application code always call `push_interruption_task_frame_and_wait` instead of directly pushing an InterruptionTaskFrame upstream?

aconchillo:
> This looks good and works well.
> 
> Just to clarify my understanding, should application code always call `push_interruption_task_frame_and_wait` instead of directly pushing an InterruptionTaskFrame upstream?

You can push an `InterruptionTaskFrame` upstream manually and it will still generate an `InterruptionFrame`, no change on that. That said, if you want to make sure ALL the processors of the pipeline get the interruption before you continue, then you need to wait on this new event.

CHANGELOG.md
aconchillo: I'll put it in its own line.

aconchillo: Actually, it should be `cancellation reason` which might fit :-D.

src/pipecat/processors/frame_processor.py
aconchillo: It depends. As I mentioned, you can still push an `InterruptionTaskFrame` and expect an `InterruptionFrame`. However, this function just waits for the interruption to be processed by every frame processor.

We can find a better name...

src/pipecat/frames/frames.py
aconchillo: thanks. fixed.

src/pipecat/processors/frame_processor.py
kompfner: Sorry, by "apply universally" I meant "changing the behavior of every call to `push_interruption_task_frame_and_wait()`", which this PR does. 

I know that you can always push `InterruptionTaskFrame` without using this utility method. But after this PR, what you can no longer do is easily replicate the previous behavior of `push_interruption_task_frame_and_wait()`, which was to wait for the `InterruptionFrame` to arrive at the current processor.  A while ago we had gone back and forth a lot about the desired behavior of `push_interruption_task_frame_and_wait()` and had ultimately landed on waiting for the `InterruptionFrame` to arrive at the current processor. We had even discussed, and explicitly decided _against_, the behavior that this PR introduces, if I'm not mistaken. I just can't remember exactly why haha.

kompfner: > I just can't remember exactly why haha.

Perhaps one reason was that we were trying to find a suitable replacement for calls like

```python
await self.push_frame(InterruptionFrame())
await self.push_frame(whatever_frame_follows_interruption)
```

that would honor the contract of interrupting all _upstream_ processors (in order), too.

And so we arrived at the _behaviorally most similar solution_ that honored the contract:

```python
# At this point, we know that the InterruptionFrame has been pushed
await self.push_frame(whatever_frame_follows_interruption)
```

After this PR we have the behaviorally distinct:

```python
# At this point, we know InterruptionFrame has made it *all the way downstream*
await self.push_frame(whatever_frame_follows_interruption)
```

aconchillo: A bit of history. Initially, system frames were being pushed in a call chain fashion, so the following code:

```python
await self.push_frame(whatever_frame_follows_interruption)
```

would result in the `InterruptionFrame` generated by `BotInterruptionFrame` to go all the way down. This was ensuring that all processors were cancelled before we proceed to the next line.

We then switched the system frames behavior to be queued, which was necessary but introduced issues, like not being able to use the above pattern anymore.

We then renamed `BotInterruptionFrame` to `InterruptionTaskFrame` (which doesn't really matter for this discussion), but the purpose of this rename was for the `PipelineTask` to generate the `InterruptionFrame` and therefore following all the other frames (`CancelTaskFrame`, `EndTsskFrame`...).

We still wanted to initial behavior, that is, interrupt the whole pipeline and only after that push whatever we want to push. SO we created this function:

```python
await self.push_frame(whatever_frame_follows_interruption)
```

In the first implementation of this function:
- We mistakenly waited for an `InterruptionFrame` to reach the processor that just called `push_interruption_task_frame_and_wait()`. But this is wrong because not all the pipeline is interrupted yet like we used to have with `BotInterruptionFrame`.

We then attempted, in 0.0.88, to fix that by introducing a new frame `InterruptionCompletedFrame`. The idea was that we would wait for that frame to reach the end of the pipeline. But this broke other things, so we quickly release 0.0.89.

This is the third attempt to go back to the original behavior. We don't introduce a new frame and we just add an asyncio event to the `InterruptionFrame`that gets set when it reaches the end of the pipeline.

Users can still push `InterruptionTaskFrame` upstream and an `InterruptionFrame` will still be generated. If you want to make sure the whole pipeline gets interrupted you need to use the asyncio event approach (`push_interruption_task_frame_and_wait()` is jsut a helper for that) , otherwise you don't have to do anything.

PR #2980, opened 2025-11-05
Refactor GoogleVertexLLMService to use GoogleLLMService as a base class

markbackman:

Google recommends using google-genai and not the OpenAI implementation for Vertex. This follows similar patterns to the Gemini Live implementation.

Also, updating the default model to `gemini-2.5-flash`. The `google/` prefix is not needed.

### DIFF
--- a/src/pipecat/services/google/llm_vertex.py
+++ b/src/pipecat/services/google/llm_vertex.py
@@ -99,7 +99,7 @@ class GoogleVertexLLMService(GoogleLLMService):
         *,
         credentials: Optional[str] = None,
         credentials_path: Optional[str] = None,
-        model: str = "google/gemini-2.0-flash-001",
+        model: str = "gemini-2.5-flash",
         location: Optional[str] = None,
         project_id: Optional[str] = None,
         params: Optional[GoogleLLMService.InputParams] = None,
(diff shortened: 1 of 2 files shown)


src/pipecat/services/google/llm_vertex.py
```diff
@@ -100,6 +102,11 @@ def __init__(
         model: str = "google/gemini-2.0-flash-001",
         location: Optional[str] = None,
         project_id: Optional[str] = None,
+        params: Optional[GoogleLLMService.InputParams] = None,
```
markbackman, line 105: Making the args more explicit for better discovery.

src/pipecat/services/google/llm.py
```diff
@@ -726,6 +725,9 @@ def __init__(
         self._tools = tools
         self._tool_config = tool_config
 
+        # Initialize the API client. Subclasses can override this if needed.
```
markbackman, line 728: Following the pattern from Gemini Live the subclass can override without requiring unused args.

src/pipecat/services/google/llm_vertex.py
```diff
@@ -100,6 +102,11 @@ def __init__(
         model: str = "google/gemini-2.0-flash-001",
```
kompfner: Maybe out of scope for this PR, but wanted to make a note: the default model for `GoogleLLMService` has been updated to 'gemini-2.5-flash'. Maybe we're due for an update here, too.

kompfner (approved):

src/pipecat/services/google/llm_vertex.py
markbackman: Good catch! Yes, let's update the model too. I'll do that here.

PR #2962, opened 2025-11-03
feat(minimax): comprehensive updates to TTS service

zhenyujia23-crypto:
- Add support for speech-2.6-hd and speech-2.6-turbo models
- Add 16 new languages (total 40): Afrikaans, Bulgarian, Catalan, Danish, Persian, Filipino, Hebrew, Croatian, Hungarian, Malay, Norwegian, Nynorsk, Slovak, Slovenian, Swedish, Tamil
- Add new emotions: calm and fluent
- Add new parameters: text_normalization (renamed from english_normalization), latex_read, force_cbr, exclude_aggregated_audio, subtitle_enable, subtitle_type
- Extract trace_id from response headers for all requests
- Improve error handling for non-streaming error responses
- Add detailed extra_info logging (audio_length, audio_size, usage_characters, word_count)
- Add validation warnings for language/model compatibility
- Fix silent error issue where HTTP 200 responses with errors were ignored

BREAKING CHANGE: Renamed parameter english_normalization to text_normalization

#### Please describe the changes in your PR. If it is addressing an issue, please reference that as well.

PR #2981, opened 2025-11-05
proposed changes to PR #2962

vipyne:

### DIFF
--- a/src/pipecat/services/minimax/tts.py
+++ b/src/pipecat/services/minimax/tts.py
@@ -81,7 +81,7 @@ def language_to_minimax_language(language: Language) -> Optional[str]:
         Language.YUE: "Chinese,Yue",
         Language.ZH: "Chinese",
     }
-    
+
     # Languages that require speech-2.6-* models
     V26_ONLY_LANGUAGES = {Language.FA, Language.FIL, Language.TA}
 
@@ -110,6 +110,7 @@ class MiniMaxHttpTTSService(TTSService):
             pitch: Pitch adjustment (range: -12 to 12).
             emotion: Emotional tone (options: "happy", "sad", "angry", "fearful",
                 "disgusted", "surprised", "calm", "fluent").
+            english_normalization: Deprecated; use `text_normalization` instead
             text_normalization: Enable text normalization (Chinese/English).
             latex_read: Enable LaTeX formula reading.
             force_cbr: Enable Constant Bitrate (CBR) for audio encoding (MP3 only).
@@ -124,6 +125,7 @@ class MiniMaxHttpTTSService(TTSService):
         volume: Optional[float] = 1.0
         pitch: Optional[int] = 0
         emotion: Optional[str] = None
+        english_normalization: Optional[bool] = None  # Deprecated
         text_normalization: Optional[bool] = None
         latex_read: Optional[bool] = None
         force_cbr: Optional[bool] = None
@@ -135,9 +137,7 @@ class MiniMaxHttpTTSService(TTSService):
         self,
         *,
         api_key: str,
-        base_url: str = "https://api.minimax.io/v1/t2a_v2", 
-        # https://api-uw.minimax.io/v1/t2a_v2
-        # support west of unite state 
+        base_url: str = "https://api.minimax.io/v1/t2a_v2",
         group_id: str,
         model: str = "speech-02-turbo",
         voice_id: str = "Calm_Woman",
@@ -153,6 +153,7 @@ class MiniMaxHttpTTSService(TTSService):
             base_url: API base URL, defaults to MiniMax's T2A endpoint.
                 Global: https://api.minimax.io/v1/t2a_v2
                 Mainland China: https://api.minimaxi.chat/v1/t2a_v2
+                Western United States: https://api-uw.minimax.io/v1/t2a_v2
             group_id: MiniMax Group ID to identify project.
             model: TTS model name. Defaults to "speech-02-turbo". Options include:
                 "speech-2.6-hd", "speech-2.6-turbo" (latest, supports Filipino/Tamil/Persian),
@@ -174,7 +175,6 @@ class MiniMaxHttpTTSService(TTSService):
         self._session = aiohttp_session
         self._model_name = model
         self._voice_id = voice_id
-        self._current_trace_id: Optional[str] = None
 
         # Create voice settings
         self._settings = {
@@ -206,7 +206,7 @@ class MiniMaxHttpTTSService(TTSService):
             service_lang = self.language_to_service_language(params.language)
             if service_lang:
                 self._settings["language_boost"] = service_lang
-                
+
                 # Validate language-model compatibility
                 # Filipino, Tamil, Persian only supported by speech-2.6-* models
                 if params.language in {Language.FA, Language.FIL, Language.TA}:
@@ -238,6 +238,13 @@ class MiniMaxHttpTTSService(TTSService):
                     f"Unsupported emotion: {params.emotion}. Supported emotions: {supported_emotions}"
                 )
 
+        # If `english_normalization`, add `text_normalization` and print warning
+        if params.english_normalization is not None:
+            logger.warning(
+                "Parameter `english_normalization` is deprecated and will be removed in a future version. Use `text_normalization` instead."
+            )
+            self._settings["voice_setting"]["text_normalization"] = params.english_normalization
+
         # Add text_normalization if provided (corrected parameter name)
         if params.text_normalization is not None:
             self._settings["voice_setting"]["text_normalization"] = params.text_normalization
@@ -261,9 +268,9 @@ class MiniMaxHttpTTSService(TTSService):
                     "Subtitles may not be generated. "
                     "For subtitle support, consider implementing a non-streaming TTS service."
                 )
-            
+
             self._settings["subtitle_enable"] = params.subtitle_enable
-            
+
             # Add subtitle_type only when subtitle_enable is True
             if params.subtitle_enable and params.subtitle_type:
                 # Validate subtitle_type
@@ -341,9 +348,6 @@ class MiniMaxHttpTTSService(TTSService):
         """
         logger.debug(f"{self}: Generating TTS [{text}]")
 
-        # Reset trace_id for new request
-        self._current_trace_id = None
-
         headers = {
             "accept": "application/json, text/plain, */*",
             "Content-Type": "application/json",
@@ -361,14 +365,10 @@ class MiniMaxHttpTTSService(TTSService):
             async with self._session.post(
                 self._base_url, headers=headers, json=payload
             ) as response:
-                # Extract trace_id from response header (available in all responses)
-                trace_id = response.headers.get("Trace-Id") or response.headers.get("trace-id") or response.headers.get("X-Trace-Id") or "unknown"
-                self._current_trace_id = trace_id
-                
-                # Log trace_id for all requests
-                logger.info(f"MiniMax TTS request trace_id={trace_id}, status={response.status}")
-                
+                logger.trace(f"MiniMax TTS request status={response.status}")
+
                 if response.status != 200:
+                    error_message = "MiniMax TTS error"
                     # Try to read error response body
                     try:
                         error_body = await response.text()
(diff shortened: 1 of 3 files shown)


src/pipecat/services/minimax/tts.py
```diff
@@ -97,21 +116,35 @@ class InputParams(BaseModel):
...
             emotion: Emotional tone (options: "happy", "sad", "angry", "fearful",
-                "disgusted", "surprised", "neutral").
-            english_normalization: Whether to apply English text normalization.
+                "disgusted", "surprised", "calm", "fluent").
+            english_normalization: Deprecated; use `text_normalization` instead
+            text_normalization: Enable text normalization (Chinese/English).
+            latex_read: Enable LaTeX formula reading.
+            force_cbr: Enable Constant Bitrate (CBR) for audio encoding (MP3 only).
+            exclude_aggregated_audio: Whether to exclude aggregated audio in final chunk.
+            subtitle_enable: Enable subtitle generation (non-streaming only).
```
markbackman: We should consider removing since this parameter and `subtitle_type` as they're for non-streaming only. This MiniMax implementation is streaming only.

The alternative would be to add a `streaming` arg to allow for a non-streaming mode`, but I don't think that's a good idea.

src/pipecat/services/minimax/tts.py
```diff
@@ -97,21 +116,35 @@ class InputParams(BaseModel):
...
             speed: Speech speed (range: 0.5 to 2.0).
             volume: Speech volume (range: 0 to 10).
             pitch: Pitch adjustment (range: -12 to 12).
             emotion: Emotional tone (options: "happy", "sad", "angry", "fearful",
-                "disgusted", "surprised", "neutral").
-            english_normalization: Whether to apply English text normalization.
+                "disgusted", "surprised", "calm", "fluent").
+            english_normalization: Deprecated; use `text_normalization` instead
+            text_normalization: Enable text normalization (Chinese/English).
+            latex_read: Enable LaTeX formula reading.
```
markbackman, line 117: Is this actually applicable? I guess LLMs can output LaTeX format, but I don't know if it makes sense for a TTS to do the same 🤷

src/pipecat/services/minimax/tts.py
```diff
@@ -97,21 +116,35 @@ class InputParams(BaseModel):
...
+            force_cbr: Enable Constant Bitrate (CBR) for audio encoding (MP3 only).
```
markbackman: Is `mp3` and option? Again, this probably doesn't make sense for a streaming use case.

src/pipecat/services/minimax/tts.py
```diff
@@ -178,9 +220,20 @@ def __init__(
             if service_lang:
                 self._settings["language_boost"] = service_lang
 
+                # Validate language-model compatibility
```
markbackman: I'm not sure if we should include this type of checking. It's likely to become stale and need to be maintained. It would be better to rely on the service returning an error about unsupported languages for a given model.

src/pipecat/services/minimax/tts.py
zhenyujia23-crypto: Here is a mismatch. Our model can support streaming subtitle.

src/pipecat/services/minimax/tts.py
```diff
@@ -178,9 +220,20 @@ def __init__(
...
+                            f"Language {params.language.value} ({service_lang}) is only supported by "
+                            f"speech-2.6-hd and speech-2.6-turbo models. "
+                            f"Current model '{model}' may not support this language. "
+                            f"Consider using 'speech-2.6-turbo' or 'speech-2.6-hd'."
+                        )
+
         # Add optional emotion if provided
         if params.emotion:
-            # Validate emotion is in the supported list
+            # Validate emotion is in the supported list (updated per official docs)
```
markbackman: Nit:

```suggestion
            # Validate emotion is in the supported list
```

src/pipecat/services/minimax/tts.py
```diff
@@ -97,21 +116,35 @@ class InputParams(BaseModel):
...
+            language: Language for TTS generation. Supports 40 languages.
+                Note: Filipino, Tamil, and Persian require speech-2.6-* models.
             speed: Speech speed (range: 0.5 to 2.0).
             volume: Speech volume (range: 0 to 10).
             pitch: Pitch adjustment (range: -12 to 12).
             emotion: Emotional tone (options: "happy", "sad", "angry", "fearful",
-                "disgusted", "surprised", "neutral").
-            english_normalization: Whether to apply English text normalization.
+                "disgusted", "surprised", "calm", "fluent").
+            english_normalization: Deprecated; use `text_normalization` instead
```
markbackman, line 110: Use deprecated directive (note indentation and newlines, which are required for proper formatting):

```suggestion
            english_normalization: Deprecated; use `text_normalization` instead

                .. deprecated:: 0.0.93
                    The `english_normalization` parameter is deprecated and will be removed in a future version.
                    Use the `text_normalization` parameter instead.

```

src/pipecat/services/minimax/tts.py
```diff
@@ -189,15 +242,65 @@ def __init__(
...
-                logger.warning(f"Unsupported emotion: {params.emotion}. Using default.")
+                logger.warning(
+                    f"Unsupported emotion: {params.emotion}. Supported emotions: {supported_emotions}"
+                )
 
-        # Add english_normalization if provided
+        # If `english_normalization`, add `text_normalization` and print warning
         if params.english_normalization is not None:
-            self._settings["english_normalization"] = params.english_normalization
+            logger.warning(
```
markbackman: Instead use the `warnings` package for this. You can see examples in other places in the code base.

src/pipecat/services/minimax/tts.py
zhenyujia23-crypto: Yes, we do support. but need transform to the 
Clipboard_Screenshot_1762355106

src/pipecat/services/minimax/tts.py
```diff
@@ -189,15 +242,65 @@ def __init__(
...
+                "Parameter `english_normalization` is deprecated and will be removed in a future version. Use `text_normalization` instead."
+            )
+            self._settings["voice_setting"]["text_normalization"] = params.english_normalization
+
+        # Add text_normalization if provided (corrected parameter name)
+        if params.text_normalization is not None:
+            self._settings["voice_setting"]["text_normalization"] = params.text_normalization
+
+        # Add latex_read if provided
```
markbackman, line 235: Consider removing based if the parameter isn't included.

src/pipecat/services/minimax/tts.py
```diff
@@ -189,15 +242,65 @@ def __init__(
...
+        if params.latex_read is not None:
+            self._settings["voice_setting"]["latex_read"] = params.latex_read
+
+        # Add force_cbr if provided (for MP3 format only)
```
markbackman: Same. Consider removing based if the parameter isn't included.

src/pipecat/services/minimax/tts.py
```diff
@@ -189,15 +242,65 @@ def __init__(
...
+        if params.force_cbr is not None:
+            self._settings["audio_setting"]["force_cbr"] = params.force_cbr
+
+        # Add subtitle settings if provided
```
markbackman: I think we should remove the parameter and this logic since non-streaming isn't supported.

src/pipecat/services/minimax/tts.py
```diff
@@ -244,7 +347,7 @@ async def start(self, frame: StartFrame):
         """
         await super().start(frame)
         self._settings["audio_setting"]["sample_rate"] = self.sample_rate
-        logger.debug(f"MiniMax TTS initialized with sample rate: {self.sample_rate}")
+        logger.debug(f"MiniMax TTS initialized with sample_rate={self.sample_rate}")
```
markbackman: Nit: Not sure this change is needed.

```suggestion
        logger.debug(f"MiniMax TTS initialized with sample_rate: {self.sample_rate}")
```

src/pipecat/services/minimax/tts.py
```diff
@@ -275,26 +378,77 @@ async def run_tts(self, text: str) -> AsyncGenerator[Frame, None]:
...
+                        logger.error(error_message)
+
                     yield ErrorFrame(error=error_message)
                     return
 
                 await self.start_tts_usage_metrics(text)
                 yield TTSStartedFrame()
 
                 # Process the streaming response
+                logger.trace(f"Starting to read streaming response, status={response.status}")
```
markbackman: Remove

src/pipecat/services/minimax/tts.py
```diff
@@ -275,26 +378,77 @@ async def run_tts(self, text: str) -> AsyncGenerator[Frame, None]:
...
                 buffer = bytearray()
 
                 CHUNK_SIZE = self.chunk_size
+                chunk_count = 0
 
                 async for chunk in response.content.iter_chunked(CHUNK_SIZE):
+                    chunk_count += 1
+                    logger.trace(f"Received chunk #{chunk_count}, size={len(chunk)} bytes")
```
markbackman: Remove.

src/pipecat/services/minimax/tts.py
```diff
@@ -275,26 +378,77 @@ async def run_tts(self, text: str) -> AsyncGenerator[Frame, None]:
...
                     if not chunk:
                         continue
 
                     buffer.extend(chunk)
 
+                    # Log raw buffer content for debugging
+                    if chunk_count == 1:
+                        logger.trace(f"Raw buffer content: {buffer[:200]}")  # First 200 bytes
```
markbackman: Remove.

src/pipecat/services/minimax/tts.py
```diff
@@ -275,26 +378,77 @@ async def run_tts(self, text: str) -> AsyncGenerator[Frame, None]:
...
+
+                        # Check if first chunk is a direct JSON error (not streaming format)
```
markbackman: Non-streaming isn't supported, so remove?

src/pipecat/services/minimax/tts.py
zhenyujia23-crypto: This para default is false, doesn't effect streaming. For streaming use case, user can just ignore it.

Clipboard_Screenshot_1762355257

src/pipecat/services/minimax/tts.py
```diff
@@ -311,16 +465,43 @@ async def run_tts(self, text: str) -> AsyncGenerator[Frame, None]:
                         buffer = buffer[next_start:]
 
                         try:
-                            data = json.loads(data_block[5:].decode("utf-8"))
-                            # Skip data blocks containing extra_info
+                            data_str = data_block[5:].decode("utf-8")
+                            logger.trace(
```
markbackman: Remove this and other trace logging.

src/pipecat/services/minimax/tts.py
```diff
@@ -311,16 +465,43 @@ async def run_tts(self, text: str) -> AsyncGenerator[Frame, None]:
...
+                                extra_info = data.get("extra_info", {})
+                                logger.debug(f"Received final chunk with extra info: {extra_info}")
+                                continue  # No audio data in this block
 
+                            # Extract audio data
                             chunk_data = data.get("data", {})
                             if not chunk_data:
                                 continue
 
+                            # Check for subtitle file (if subtitle generation is enabled)
```
markbackman: Remove

markbackman (commented):
Ok, I think the biggest issue to sort out is what the scope of this service is. Pipecat is for real-time use cases, so non-streaming is not really applicable. I think we should remove the non-streaming items from this class to keep it focused and simple.

Also, remove the trace logging, as it's only applicable during development and isn't practical for real-word debugging.

src/pipecat/services/minimax/tts.py
zhenyujia23-crypto: Will remove it

src/pipecat/services/minimax/tts.py
zhenyujia23-crypto: ok

src/pipecat/services/minimax/tts.py
vipyne: perhaps `(non-streaming only).` should be omitted from comment then?

src/pipecat/services/minimax/tts.py
zhenyujia23-crypto: But, I think we still need a place to log the trace_id, coz, it will help us to know the problem from our service.

src/pipecat/services/minimax/tts.py
zhenyujia23-crypto: Got it!

src/pipecat/services/minimax/tts.py
vipyne: @zhenyujia23-crypto is there a way subtitles can be added to the example `07y-interruptible-minimax.py` to exercise this parameter?

src/pipecat/services/minimax/tts.py
vipyne: Is the tracing code necessary for other folks who use this class?
Perhaps a better approach is to create a git patch with all the tracing code that can be applied locally to a user's pipecat code.
I see that it is useful, but it feels like it should be a more "opt-in" feature.

src/pipecat/services/minimax/tts.py
vipyne: will remove.

src/pipecat/services/minimax/tts.py
markbackman: I guess there's no harm in including it, but generally voice agents are best if they skip pronouncing complex math (code or tables).

src/pipecat/services/minimax/tts.py
markbackman: Yes, let's remove if it doesn't apply to streaming then.

vipyne:
closing in favor of #3096

PR #2982, opened 2025-11-05
Improved LocalSmartTurnAnalyzerV3 performance on systems with a low CPU count

marcus-daily:


src/pipecat/audio/turn/smart_turn/local_smart_turn_v3.py
```diff
@@ -35,12 +35,13 @@ class LocalSmartTurnAnalyzerV3(BaseSmartTurn):
     enabling offline operation without network dependencies.
     """
 
-    def __init__(self, *, smart_turn_model_path: Optional[str] = None, **kwargs):
+    def __init__(self, *, smart_turn_model_path: Optional[str] = None, cpu_count=1, **kwargs):
```
aconchillo: This should be `cpu_count: int = 1`

markbackman: Touché! I'll make this update and will merge it.

PR #2986, opened 2025-11-05
Fix docstrings for 0.0.93 release, fix classmethod placement in Reque…

markbackman:
…stHandler

### DIFF
--- a/src/pipecat/transports/smallwebrtc/request_handler.py
+++ b/src/pipecat/transports/smallwebrtc/request_handler.py
@@ -39,13 +39,12 @@ class SmallWebRTCRequest:
     restart_pc: Optional[bool] = None
     request_data: Optional[Any] = None
 
-
-@classmethod
-def from_dict(cls, data: dict):
-    """Accept both snake_case and camelCase for the request_data field."""
-    if "requestData" in data and "request_data" not in data:
-        data["request_data"] = data.pop("requestData")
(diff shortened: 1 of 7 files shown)


src/pipecat/transports/smallwebrtc/request_handler.py
```diff
@@ -39,13 +39,12 @@ class SmallWebRTCRequest:
...
     request_data: Optional[Any] = None
 
-
-@classmethod
-def from_dict(cls, data: dict):
-    """Accept both snake_case and camelCase for the request_data field."""
-    if "requestData" in data and "request_data" not in data:
-        data["request_data"] = data.pop("requestData")
-    return cls(**data)
+    @classmethod
```
markbackman, line 42: This had incorrect indentation, meaning the `@classmethod` wasn't correctly inside of the target class 🙈. Fixed now!

kompfner:
Arggh sorry for all the docstring nonsense I introduced. I have no idea what syntax it expects.

How do you test it? Just like hovering over symbols in VS Code?

markbackman:
No worries. I wish we had good linting rules, but we don't. I just periodically run the docs/api/build-docs.sh and clean up the errors. It only takes maybe 5-10 min and I do it ever few weeks.

If we can add linting rules, that would be amazing. Maybe I'll look hard to find what's out there.

PR #2987, opened 2025-11-05
Fixing the runner start endpoint when enableDefaultIceServers is enabled.

filipi87:
Fixing the runner start endpoint when enableDefaultIceServers is enabled.

### DIFF
--- a/src/pipecat/runner/run.py
+++ b/src/pipecat/runner/run.py
@@ -74,7 +74,7 @@ import uuid
 from contextlib import asynccontextmanager
 from http import HTTPMethod
 from pathlib import Path
-from typing import Any, Dict, List, Optional, TypedDict
+from typing import Any, Dict, List, Optional, TypedDict, Union
 
 import aiohttp
 from fastapi.responses import FileResponse, Response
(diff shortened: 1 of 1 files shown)


src/pipecat/runner/run.py
```diff
@@ -216,6 +216,9 @@ def _setup_webrtc_routes(
         logger.error(f"WebRTC transport dependencies not installed: {e}")
         return
 
+    class IceServer(TypedDict, total=False):
```
markbackman, line 219: I see there's a type alias for RTCIceServer in smallwebrtc.connection. Does that type information not get passed on?

filipi87: The `RTCIceServer` is what `aiortc` requires, but here we only need to return the JSON with the ICE servers to the client. If we use `RTCIceServer` as we did before, our JSON will include some extra fields, like `credentialType`. That works fine for the web, but on mobile (iOS and React Native), it breaks the app, it seems that WebRTC in those cases doesn’t expect to receive that field.

Since it wasn’t needed here, I created a separate class to return the JSON we actually need, similar to how we receive it from PCC.

PR #2990, opened 2025-11-06
fix default riva tts voice_id

vipyne:


vipyne:
Ray was throwing the following error:
```
RivaTTSService#0 exception: <_MultiThreadedRendezvous of RPC that terminated with:
	status = StatusCode.INVALID_ARGUMENT
	details = "Model is not available on server: subvoice requested not found"
	debug_error_string = "UNKNOWN:Error received from peer  {grpc_message:"Model is not available on server: subvoice requested not found", grpc_status:3, created_time:"2025-11-05T18:26:01.290565-06:00"}"
>
```

PR #2992, opened 2025-11-06
Add MetricsLogObserver

markbackman:

I've been writing custom processors to look at MetricsFrames too frequently. This will save me time :)

Also, something for consideration, we could include deduplication logic in the BaseObserver class...

### DIFF
--- a/src/pipecat/observers/loggers/metrics_log_observer.py
+++ b/src/pipecat/observers/loggers/metrics_log_observer.py
@@ -1,0 +1,218 @@
+#
+# Copyright (c) 2024–2025, Daily
+#
+# SPDX-License-Identifier: BSD 2-Clause License
+#
+
+"""Metrics logging observer for Pipecat.
+
+This module provides an observer that logs metrics frames to the console,
+allowing developers to monitor performance metrics, token usage, and other
+statistics in real-time.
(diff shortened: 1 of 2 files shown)


src/pipecat/observers/loggers/metrics_log_observer.py
```diff
@@ -0,0 +1,232 @@
...
+
+        Args:
+            metrics_data: The metrics data to check.
+
+        Returns:
+            True if the metrics should be logged, False otherwise.
+        """
+        metrics_type = type(metrics_data)
+
+        # If include_metrics is specified, only log those types
```
kompfner, line 116: Since `include_metrics` an `exclude_metrics` are mutually exclusive, should we log a warning at init time if someone attempted to set both? The warning could mention that `include_metrics` takes precedence.

kompfner (approved):

src/pipecat/observers/loggers/metrics_log_observer.py
markbackman: Since there are only a handful of metrics data types, I'm just going to drop `exclude_metrics`. It'll be more common to want all or just 1 (or 2) metrics data types.

PR #2994, opened 2025-11-06
Fix incorrect docstring in FrameProcessorQueue.__init__

Toprak2:
Removes reference to non-existent manager parameter in the __init__ docstring and updates the description to accurately reflect the method's actual behavior.


markbackman:
Nice catch! 🕵️ 

I think the docstring can be as simple as:
```
"""Initialize the FrameProcessorQueue."""
```

Mind making that change? (Also, be sure to run scripts/fix-ruff.sh before committing.)

Toprak2:
Done. Had some issues with git but hope it is ok now.

PR #2995, opened 2025-11-06
fix: STTMuteFilter no longer sends STTMuteFrame

markbackman:

Two things:
1. The STTMuteFilter was pushing an STTMuteFrame. This frame mutes input to the STT which can cause bad things to happen (e.g. 409 error with GoogleSTTService). Instead, the STTMuteFilter should _only_ block frames. This is a cleaner implementation, IMO.
2. Remove the is_muted `@property` which is just unnecessary. It was never documented, so it should have no impact.

This is not breaking, but will cause transcripts to appear when they're not desired. If we opt to, we could make it configurable to mute the STT. But, I'm not sure that actually matters. Note: teams have reported asking for the transcript when the STTMuteFilter is enabled, so this would actually provide them with access if they were interested.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -58,6 +58,13 @@ reason")`.
 
 ### Changed
 
+- `STTMuteFilter` no longer sends `STTMuteFrame` to the STT service. The filter
+  now blocks frames locally without instructing the STT service to stop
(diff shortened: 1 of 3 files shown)


aconchillo:
LGTM! Much better!

Issue #2180, opened 2025-07-10
Google STT Stream Timeout Error (409)


obigroup:
### pipecat version

0.0.75

### Service Name

Google STT

### Service or model version

V2

### Issue Description

Google STT service is experiencing frequent stream timeout errors with a 409 status code. The error occurs when the stream times out after receiving no more client requests, causing the STT processing to fail.

### Reproduction Steps

1. Initialize Google STT service in pipecat
2. Start audio streaming to Google STT
3. Wait for periods of silence or reduced audio input
4. The stream times out after a period of inactivity
5. Error 409 is thrown with "Stream timed out after receiving no more client requests"

### Expected Behavior

The Google STT service should handle periods of silence gracefully without timing out, or should implement proper reconnection logic to handle stream timeouts transparently.

### Actual Behavior

The service throws a 409 error when the stream times out, interrupting the speech-to-text processing and requiring manual intervention or service restart.

### Error Logs

```shell
ERROR    | pipecat.services.google.stt:_process_responses:891 - Error processing Google STT responses: 409 Stream timed out after receiving no more client requests. [type_url: "type.googleapis.com/util.StatusProto"
value: "\010\n\022\007generic\0329Stream timed out after receiving no more client requests.*D\013\020\206\326\215'\032;\0229Stream timed out after receiving no more client requests.\014"
]
```

markbackman:
Can you try this on 0.0.76? I'm wondering if the silence processing has improved things for GoogleSTTService?

obigroup:
@markbackman 

It seems that version 0.0.76 introduces a new bug. Now no STT (Speech-to-Text) works with Twilio.

With or without VAD, no audio is being converted to text:

```
2025-07-14 10:17:56.720 | DEBUG    | pipecat.transports.base_input:_handle_user_interruption:348 - User started speaking
2025-07-14 10:17:59.119 | DEBUG    | pipecat.transports.base_input:_handle_user_interruption:372 - User stopped speaking
2025-07-14 10:18:02.599 | DEBUG    | pipecat.transports.base_input:_handle_user_interruption:348 - User started speaking
2025-07-14 10:18:04.997 | DEBUG    | pipecat.transports.base_input:_handle_user_interruption:372 - User stopped speaking
2025-07-14 10:18:19.199 | DEBUG    | pipecat.transports.base_input:_handle_user_interruption:348 - User started speaking
[...]

PR #2997, opened 2025-11-06
GoogleSTTService: Add more robust handling of 409 errors

markbackman:

This issue should not really happen now that the STTMuteFilter isn't blocking input. But, it's still possible to block input, which isn't a fatal error. Instead, it's better to handle it and reconnect.

Also updates 07n to use chirp_3.

Along with https://github.com/pipecat-ai/pipecat/pull/2995, this fixes https://github.com/pipecat-ai/pipecat/issues/2180.


aconchillo:
LGTM!

PR #2998, opened 2025-11-06
BaseOutputTransport: send silence when EndFrame is received

aconchillo:

Give the audio more chance to make it to the client.

### DIFF
--- a/src/pipecat/transports/base_transport.py
+++ b/src/pipecat/transports/base_transport.py
@@ -83,6 +83,7 @@ class TransportParams(BaseModel):
         audio_out_10ms_chunks: Number of 10ms chunks to buffer for output.
         audio_out_mixer: Audio mixer instance or destination mapping.
         audio_out_destinations: List of audio output destination identifiers.
+        audio_out_end_silence_secs: How much silence to send after an EndFrame (0 for no silence).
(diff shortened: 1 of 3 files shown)


src/pipecat/transports/base_transport.py
```diff
@@ -131,6 +132,7 @@ class TransportParams(BaseModel):
     audio_out_10ms_chunks: int = 4
     audio_out_mixer: Optional[BaseAudioMixer | Mapping[Optional[str], BaseAudioMixer]] = None
     audio_out_destinations: List[str] = Field(default_factory=list)
+    audio_out_end_silence_secs: int = 2
```
markbackman, line 135: Is 2 seconds just a starting point or did you do some testing to determine it?

filipi87 (approved):
Nice improvement. 🚀

src/pipecat/transports/base_transport.py
aconchillo: Just a starting point. 2 seconds should be enough. Since it's configurable, people should be able to tweak it depending on the transport.

markbackman: Makes sense.

PR #3000, opened 2025-11-07
Add clarifying documentation about OpenAI Realtime model use

markbackman:

I was caught off guard by the `model` param in SessionProperties being a no-op. I don't want others to be impacted by the same thing.

### DIFF
--- a/src/pipecat/services/openai/realtime/events.py
+++ b/src/pipecat/services/openai/realtime/events.py
@@ -170,6 +170,10 @@ class SessionProperties(BaseModel):
         object: Object type identifier, always "realtime.session".
         id: Unique identifier for the session.
         model: The Realtime model used for this session.
+            Note: The model is set at connection time via the WebSocket URL query
+            parameter and cannot be changed during the session. This field is populated
+            by the server in session.created and session.updated responses, but is effectively
+            a no-op for session.update requests.
         output_modalities: The set of modalities the model can respond with.
         instructions: System instructions for the assistant.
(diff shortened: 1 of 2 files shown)


src/pipecat/services/openai/realtime/events.py
```diff
@@ -170,6 +170,10 @@ class SessionProperties(BaseModel):
         object: Object type identifier, always "realtime.session".
         id: Unique identifier for the session.
         model: The Realtime model used for this session.
+            Note: The model is set at connection time via the WebSocket URL query
```
kompfner, line 173: I wouldn't describe this behavior in terms of the underlying events, which to the developer are implementation details of `OpenAIRealtimeLLMService`.

I would instead say that `model` and `voice` take effect at init time but have no effect when provided to an `LLMUpdateSettingsFrame`.

markbackman: The confusion was that developers were setting `model` in the `SessionProperties` to initialize OpenAIRealtimeLLMService. They should be using the `model` arg. Let me make that more clear in a follow up.

markbackman: Something like this:
```
        model: The Realtime model used for this session.
            Note: The model is set at connection time via model arg in __init__
            and cannot be changed during the session.
```

kompfner: Ohhh I misunderstood. Gotcha.

PR #3001, opened 2025-11-07
Added support for passing in a `ToolsSchema` in lieu of a list of pro…

kompfner:
…vider-specific dicts when initializing `OpenAIRealtimeLLMService`.

I chose to go the somewhat hacky route of adding the `ToolsSchema` support into the `events.SessionProperties` model itself—even though we should never serialize that type when creating events—because the alternative seemed to be to create a new type for `OpenAIRealtimeLLMService` initialization parameters and then we'd have to contend with backward compatibility, which seemed like a bigger headache.

### DIFF
--- a/examples/foundational/19-openai-realtime.py
+++ b/examples/foundational/19-openai-realtime.py
@@ -231,7 +231,12 @@ Remember, your responses should be short. Just one or two sentences, usually. Re
             standard_tools=[weather_function, restaurant_function, get_news_function]
         )
         await task.queue_frames([LLMSetToolsFrame(tools=new_tools)])
-        # Alternative pattern, useful if you're changing other session properties too
+        # Alternative pattern, useful if you're changing other session properties, too.
+        # (Though note that tools in your LLMContext take precedence over those
+        # in session properties, so if you have context-provided tools, prefer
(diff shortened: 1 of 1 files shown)


kompfner:
This PR is in response to stumbling over this quirk while testing. Hopefully we save someone else from stubbing their toe in the same way.

kompfner:
Note that `ToolsSchema` was already supported when using `LLMSetToolsFrame` to update `OpenAIRealtimeLLMService`'s tools.

kompfner:
Hm...but it might not be supported when using `LLMUpdateSettingsFrame`...let me look into that!

kompfner:
> Hm...but it might not be supported when using `LLMUpdateSettingsFrame`...let me look into that!

Fixed.

CHANGELOG.md
```diff
@@ -9,6 +9,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- Added support for passing in a `ToolsSchem` in lieu of a list of provider-
```
markbackman, line 12:
```suggestion
- Added support for passing in a `ToolsSchema` in lieu of a list of provider-
```

filipi87 (approved):
LGTM!

PR #3004, opened 2025-11-07
Improve concatenate_aggregated_text string utility

markbackman:

A user in Discord pointed out how a single word output by OpenAI Realtime is split into multiple segments. This word looks like:
```
["Met", "am", "orph", "osis", "."]
```

Normally OpenAI Realtime has leading or trailing spaces, but because it's a single word, it's mistakenly classified in the "else" condition which is intended for word-by-word fragments.

Gemini Live has a similar issue:
```
["Met", "amo", "rph", "osi", "s.\n"]
```

This change adds edge case handling for single word outputs from OpenAI Realtime and Gemini Live. This works by:
- OpenAI Realtime: Look for a punctuation only token (and no leading/trailing spaces), which is unique to OpenAI
- Gemini Live: Add a condition to classify `\n` as a fragment space.

I added unit tests and tested these services:
- Word level TTS: Cartesia, ElevenLabs, Rime
- Sentence level TTS: Google (all of these services output the same as this is from the TTS aggregator.)
- OpenAI Realtime
- Gemini Live
- AWS Nova Sonic (outputs the full transcript at once, so unimpacted)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -80,6 +80,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- Improved `concatenate_aggregated_text()` to one word outputs from OpenAI
+  Realtime and Gemini Live. Text fragments are now correctly concatenated
+  without spaces when these patterns are detected.
+
 - `STTMuteFilter` no longer sends `STTMuteFrame` to the STT service. The filter
   now blocks frames locally without instructing the STT service to stop
   processing audio. This prevents inactivity-related errors (such as 409 errors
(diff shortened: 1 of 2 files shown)


aconchillo:
LGTM

kompfner:
D'oh. Yeah, in my testing I can absolutely still get OpenAI Realtime to break, even with these heuristics, if I ask it to respond with just one word.

kompfner:
> This simple approach breaks down when there is not meant to be any whitespace in its inputs, like with single-word inputs. That's the root problem, I think.

Ah, sorry for being so pedantic, I just noticed in your PR description you are absolutely aware that this is the root problem!

kompfner:
I did a refactor of this logic mostly to make it clearer for myself while trying to understand it. It uses all the same heuristics you added here, and tries not to change any logic: https://github.com/pipecat-ai/pipecat/pull/3008.

If you agree it's clearer, we could accept that PR as a step in the right direction for now. But as we discussed in Slack, it's not the fundamental change we need to make this bit of logic foolproof. We probably need to add awareness of what the text source is, like we used to do with `create_context_aggregator()`.

A case that still fails: tell OpenAI Realtime "Tell me a long word. Answer with only one word." Not 100% failure, but frequent.

PR #3006, opened 2025-11-07
restore vision/image backwards compatibility

aconchillo:

I didn't test with non-universal contexts. This makes previous examples work.


filipi87 (approved):
I see that we’re restoring it and just marking everything as deprecated. I’m just wondering, if no one has complained, should we simply drop this backward support in this case and mention it as a breaking change in the changelog?

In any case, I’m approving it and leaving the final decision to you.

aconchillo:
> I see that we’re restoring it and just marking everything as deprecated. I’m just wondering, if no one has complained, should we simply drop this backward support in this case and mention it as a breaking change in the changelog?
> 
> In any case, I’m approving it and leaving the final decision to you.

They did complain :sweat_smile: . And people don't upgrade right away so we might break others as well.

PR #3008, opened 2025-11-07
Refactor `concatenate_aggregated_text()` for clarity

kompfner:
This PR is optional. It definitely helped me wrap my head around this logic and feels clearer/more maintainable to _me_, but my brain isn't everyone's :)

### DIFF
--- a/src/pipecat/utils/string.py
+++ b/src/pipecat/utils/string.py
@@ -202,7 +202,8 @@ def concatenate_aggregated_text(text_parts: List[str]) -> str:
     """Concatenate a list of text parts into a single string.
 
     This function joins the provided list of text parts into a single string,
-    taking into account whether or not the parts already contain spacing.
+    taking into account whether or not the parts already likely contain
+    necessary spaces between parts.
 
     This function is useful for aggregating text segments received from LLMs or
     transcription services.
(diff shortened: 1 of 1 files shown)


src/pipecat/utils/string.py
```diff
@@ -213,49 +214,59 @@ def concatenate_aggregated_text(text_parts: List[str]) -> str:
...
-        # - OpenAI Realtime: ['Hey', ' there', '!', ' Great', ' to', ' meet', ' you', '!']
-        # - Gemini Live (spaces): ['Hel', 'lo.', ' Wo', 'u', 'ld ', 'you', ' li', 'ke ', 'to ', 'he', 'ar a joke?\n']
-        # - Gemini Live (newline): ["Met", "amo", "rph", "osi", "s.\n"] → "Metamorphosis."
-        # - Sentence level TTS services: ['Hello!', ' How can I assist you today?']
-        result = "".join(text_parts)
-    else:
-        # Word-by-word fragments without spacing - join with spaces. Examples:
-        # - Word level TTS services: ["Hello", "there.", "How", "are", "you?"] → "Hello there. How are you?"
-        result = " ".join(text_parts)
+    if looks_like_openai_realtime:
```
markbackman, line 258: Nit: You can probably just consolidate these two checks, but that's not essential:

```
if looks_like_gemini_live or looks_like_openai_realtime:
    text_source_spaces_parts = True
```

kompfner:
Closing, as it looks like we need to abandon this heuristic-based approach entirely

PR #3020, opened 2025-11-10
Make the mechanism of adding spaces when concatenating TTS (or speech…

kompfner:
…-to-speech LLM) output text explicit and deterministic, rather than heuristic-based

### DIFF
--- a/src/pipecat/frames/frames.py
+++ b/src/pipecat/frames/frames.py
@@ -330,10 +330,18 @@ class TextFrame(DataFrame):
 
     text: str
     skip_tts: bool = field(init=False)
+    # Whether any necessary inter-frame (leading/trailing) spaces are already
+    # included in the text.
+    # NOTE: Ideally this would be available at init time with a default value,
+    # but that would impact how subclasses can be initialized (it would require
+    # mandatory fields of theirs to have defaults to preserve
+    # non-default-before-default argument order)
+    includes_inter_frame_spaces: bool = field(init=False)
 
     def __post_init__(self):
         super().__post_init__()
         self.skip_tts = False
+        self.includes_inter_frame_spaces = False
 
     def __str__(self):
         pts = format_pts(self.pts)
--- a/src/pipecat/processors/frameworks/langchain.py
+++ b/src/pipecat/processors/frameworks/langchain.py
@@ -107,7 +107,9 @@ class LangchainProcessor(FrameProcessor):
                 {self._transcript_key: text},
                 config={"configurable": {"session_id": self._participant_id}},
             ):
-                await self.push_frame(TextFrame(self.__get_token_value(token)))
+                frame = TextFrame(self.__get_token_value(token))
+                frame.includes_inter_frame_spaces = True
+                await self.push_frame(frame)
         except GeneratorExit:
             logger.warning(f"{self} generator was closed prematurely")
         except Exception as e:
--- a/src/pipecat/services/google/gemini_live/llm.py
+++ b/src/pipecat/services/google/gemini_live/llm.py
@@ -1453,7 +1453,10 @@ class GeminiLiveLLMService(LLMService):
 
             self._bot_text_buffer += text
             self._search_result_buffer += text  # Also accumulate for grounding
-            await self.push_frame(LLMTextFrame(text=text))
+            frame = LLMTextFrame(text=text)
+            # Gemini Live text already includes any necessary inter-chunk spaces
+            frame.includes_inter_frame_spaces = True
+            await self.push_frame(frame)
 
         # Check for grounding metadata in server content
         if msg.server_content and msg.server_content.grounding_metadata:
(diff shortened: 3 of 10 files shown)


src/pipecat/services/google/gemini_live/llm.py
```diff
@@ -1453,7 +1453,9 @@ async def _handle_msg_model_turn(self, msg: LiveServerMessage):
 
             self._bot_text_buffer += text
             self._search_result_buffer += text  # Also accumulate for grounding
-            await self.push_frame(LLMTextFrame(text=text))
+            frame = LLMTextFrame(text=text)
+            frame.includes_inter_frame_spaces = True
```
kompfner, line 1458: Note that this is kind of a no-op, at least for all of our examples, because these `LLMTextFrame`s always go through a TTS and not directly to an assistant context aggregator. But including this should help in case there's a pipeline with an assistant context aggregator immediately after the LLM.

Eventually we could/should do an audit of the other LLM services to make sure they're appropriately labeling the text frames they're outputting.

src/pipecat/services/openai/realtime/llm.py
```diff
@@ -679,13 +679,18 @@ async def _handle_evt_text_delta(self, evt):
         # We receive text deltas (as opposed to audio transcript deltas) when
         # the output modality is "text"
         if evt.delta:
-            await self.push_frame(LLMTextFrame(evt.delta))
+            frame = LLMTextFrame(evt.delta)
```
kompfner, line 682: See previous comment for explanation.

kompfner:
We also should at some point (soon) audit *TTS* services to make sure their outputs are being properly labeled. I'm finding that with several of the TTS services, we're adding extra spaces to the assistant context messages. This, I believe, has been a long-standing issue (at least, until we introduced the problematic heuristic approach with our last release).

src/pipecat/utils/string.py
```diff
@@ -209,25 +209,14 @@ def concatenate_aggregated_text(text_parts: List[str]) -> str:
...
-    # Apply corresponding joining method
-    if contains_spacing_between_fragments:
-        # Fragments already have spacing - just concatenate
-        result = "".join(text_parts)
-    else:
-        # Word-by-word fragments - join with spaces
-        result = " ".join(text_parts)
+    # Concatenate text parts with or without spaces based on the flag
+    separator = " " if add_spaces else ""
+    result = separator.join(text_parts)
```
markbackman, line 219: 😌

src/pipecat/frames/frames.py
```diff
@@ -330,10 +330,16 @@ class TextFrame(DataFrame):
 
     text: str
     skip_tts: bool = field(init=False)
+    # NOTE: Ideally this would be available at init time with a default value,
+    # but that would impact how subclasses can be initialized (it would require
+    # mandatory fields of theirs to have defaults to preserve
+    # non-default-before-default argument order)
+    includes_inter_frame_spaces: bool = field(init=False)
```
markbackman, line 339: I think the naming is not quite right. It's not frame spaces but spaces between the tokens/words/sentences that are contained in the text in the frame. I wonder if there's a slightly more precise wording. Not sure what name though.

markbackman: Actually, maybe this is OK. It's important to make it clear that this is the leading and trailing spaces of the text chunk itself, not spaces between words in the text chunk.

src/pipecat/processors/frameworks/langchain.py
```diff
@@ -107,7 +107,9 @@ async def _ainvoke(self, text: str):
                 {self._transcript_key: text},
                 config={"configurable": {"session_id": self._participant_id}},
             ):
-                await self.push_frame(TextFrame(self.__get_token_value(token)))
+                frame = TextFrame(self.__get_token_value(token))
+                frame.includes_inter_frame_spaces = True
```
kompfner, line 111: Based on the langchain integration test (which actually tests langchain alongside an assistant context aggregator with no TTS service in between), langchain is expected to include spaces. This fixes the integration test.

markbackman, line 111: Nice! Tests FTW.

src/pipecat/frames/frames.py
kompfner: Exactly. The naming of this was very hard. I had the exact same pair of thoughts you had. More accurately, this would be called:

`includes_necessary_inter_frame_spaces`

...but that felt a bit bloated and didn't necessarily make things clearer.

kompfner: I'll add more commenting here for clarity.

kompfner:
Ok I think I've tested enough to feel comfortable that this doesn't introduce concerning regressions. With some TTS services, our assistant messages will have some extra spaces, but this was almost certainly the case before we switched to `concatenate_aggregated_text()` last release.

I can plan to audit all of our TTS + LLM services to make sure they're properly flagging the text frames they output with the appropriate `includes_inter_frame_spaces` values.

PR #3025, opened 2025-11-11
Fix Langfuse tracing for GoogleLLMService with universal LLMContext

jamsea:
- Fixed issue where input appeared as null in Langfuse dashboard for GoogleLLMService
- Added fallback to use adapter's get_messages_for_logging() for universal LLMContext
- Ensures proper message format conversion for Google/Gemini services
- Handles system message conversion to system_instruction format
- Also fixes serialization of empty message lists ([] now serializes correctly)

This fix ensures Langfuse tracing works correctly for Google services using both OpenAILLMContext/GoogleLLMContext and the universal LLMContext.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Fixed an issue with OpenTelemetry where tracing wasn't correctly displaying
   LLM completions and tools when using the universal `LLMContext`.
 
+- Fixed issue where `DeepgramFluxSTTService` failed to connect if passing a
+  `keyterm` or `tag` containing a space.
+
+- Prevented `HeyGenVideoService` from automatically disconnecting after 5 minutes.
+
 ## [0.0.94] - 2025-11-10
 
 ### Deprecated
--- a/src/pipecat/services/deepgram/flux/stt.py
+++ b/src/pipecat/services/deepgram/flux/stt.py
@@ -9,6 +9,7 @@
 import json
 from enum import Enum
 from typing import Any, AsyncGenerator, Dict, Optional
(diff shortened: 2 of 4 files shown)


src/pipecat/utils/tracing/service_decorators.py
```diff
@@ -382,43 +384,57 @@ async def wrapped_start_llm_usage_metrics(tokens):
...
+                            if isinstance(context, OpenAILLMContext):
+                                # OpenAILLMContext and subclasses have their own method
+                                messages = context.get_messages_for_logging()
+                            elif isinstance(context, LLMContext):
+                                # Universal LLMContext - use adapter for provider-native format
+                                if hasattr(self, "get_llm_adapter"):
+                                    adapter = self.get_llm_adapter()
+                                    messages = adapter.get_messages_for_logging(context)
+                            elif hasattr(context, "get_messages"):
+                                # Fallback for unknown context types
```
kompfner, line 403: Not sure how we could have an unknown context type...guess we'd be living in the upside-down

markbackman: I think the only scenario would be where someone uses the GoogleLLMContext, etc. directly. Super unlikely but possible, right? If not, we can remove this fallback.

kompfner: No, because all the "custom" context types subclass `OpenAILLMContext`, which we already handle above

markbackman: Ahh, ok. Then, we can remove the fallbacks.

markbackman: Removing here: https://github.com/pipecat-ai/pipecat/pull/3059

PR #3028, opened 2025-11-11
SpeechmaticsTTS - Support for retry when 503 error to TTS API.

sam-s10s:
If the API returns a 503 error, then a retry will be made.

An `ErrorFrame` will be emitted if the connection times out or the max timeout of 8s is reached.

### DIFF
--- a/src/pipecat/services/speechmatics/tts.py
+++ b/src/pipecat/services/speechmatics/tts.py
@@ -16,7 +16,6 @@ from pydantic import BaseModel
 
 from pipecat.frames.frames import (
     ErrorFrame,
-    FatalErrorFrame,
     Frame,
     TTSAudioRawFrame,
     TTSStartedFrame,
@@ -174,15 +173,16 @@ class SpeechmaticsTTSService(TTSService):
                             continue
 
                         except (ValueError, ArithmeticError):
-                            yield FatalErrorFrame(
-                                error=f"{self} Service unavailable [503] (attempts {attempt})"
+                            yield ErrorFrame(
+                                error=f"{self} Service unavailable [503] (attempts {attempt})",
+                                fatal=True,
                             )
                             return
 
                     # != 200 : Error
                     if response.status != 200:
-                        yield FatalErrorFrame(
-                            error=f"{self} Service unavailable [{response.status}]"
+                        yield ErrorFrame(
+                            error=f"{self} Service unavailable [{response.status}]", fatal=True
                         )
                         return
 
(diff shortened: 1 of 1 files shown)


src/pipecat/services/speechmatics/tts.py
```diff
@@ -109,64 +112,118 @@ async def run_tts(self, text: str) -> AsyncGenerator[Frame, None]:
...
+
+                        except (ValueError, ArithmeticError):
+                            yield FatalErrorFrame(
+                                error=f"{self} Service unavailable [503] (attempts {attempt})"
+                            )
+                            return
+
+                    # != 200 : Error
+                    if response.status != 200:
+                        yield FatalErrorFrame(
```
markbackman: We aren't using the `FatalErrorFrame` anywhere and may deprecate it. Instead, try the `ErrorFrame`, where you set the `fatal` param to `True`.

markbackman (approved):
LGTM. Nice update 🙌

src/pipecat/services/speechmatics/tts.py
```diff
@@ -109,64 +116,118 @@ async def run_tts(self, text: str) -> AsyncGenerator[Frame, None]:
...
 
-                yield TTSStartedFrame()
+                            # Increment attempt
+                            attempt += 1
 
-                # Process the response in streaming chunks
-                first_chunk = True
-                buffer = b""
+                            # Check if we've exceeded the maximum number of attempts
+                            if attempt > self._params.max_retries:
```
markbackman: Even better! This is what I had in mind (along with the `max_retries` InputParam).

src/pipecat/services/speechmatics/tts.py
```diff
@@ -109,64 +116,118 @@ async def run_tts(self, text: str) -> AsyncGenerator[Frame, None]:
...
+
+                    # 503 : Service unavailable
+                    if response.status == 503:
+                        """Calculate the backoff time and retry."""
 
-                await self.start_tts_usage_metrics(text)
+                        try:
+                            # Calculate the backoff time
+                            backoff_time = exponential_backoff_time(
+                                attempt=attempt, min_wait=0.25, max_wait=8.0, multiplier=0.5
```
markbackman, line 154: The only other thought is that 8 seconds is a long time for a max wait. If there's an ephemeral issue, it should clear after a retry or two. If there's a service level issue (e.g. downtime), then retries will all fail. So, the question is how long do you want to wait before deeming the service to be down. That duration is up to you, as you know your service better.

sam-s10s: Done.

PR #3029, opened 2025-11-11
Preventing HeyGenVideoService from disconnecting.

filipi87:
Preventing HeyGenVideoService from disconnecting.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to **Pipecat** will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## Unreleased
+
+### Fixed
+
+- Prevented `HeyGenVideoService` from automatically disconnecting after 5 minutes.
+
 ## [0.0.94] - 2025-11-10
 
 ### Deprecated


filipi87:
This approach feels like a more natural way to prevent `HeyGenVideoService` from disconnecting, since in this case the user never connects directly to HeyGen, as Pipecat acts as a proxy. 

It also avoids the need for an extra Keep Alive mechanism and maintains HeyGen’s default behavior of automatically disconnecting when no user joins, for example, when using `HeyGenTransport` instead of `HeyGenVideoService`.

markbackman:
In testing this, I see disconnects happening sooner than I would expect:

Connect:
```
2025-11-11 12:15:09.917 | INFO     | pipecat.services.heygen.client:_initialize:141 - HeyGen session started
2025-11-11 12:15:09.920 | INFO     | pipecat.transports.daily.transport:join:744 - Joining https://mark.staging.daily.co/waiting_1
```

Disconnected automatically <3 min later:
```
2025-11-11 12:17:57.739 | INFO     | pipecat.services.heygen.video:_on_participant_disconnected:151 - Participant disconnected heygen
livekit::rtc_engine:453:livekit::rtc_engine - received session close: "signal client closed: \"stream closed\"" UnknownReason Resume
```

Is this Livekit being unstable?

filipi87:
> Is this Livekit being unstable?

@markbackman I just reached out to HeyGen for help to understand what’s going on. 

It was working fine for me, I was able to run two sessions that lasted over 10 minutes without any issues. But I just tried again now and ran into the same issue you did.

markbackman:
> > Is this Livekit being unstable?
> 
> @markbackman I just reached out to HeyGen for help to understand what’s going on.
> 
> It was working fine for me, I was able to run two sessions that lasted over 10 minutes without any issues. But I just tried again now and ran into the same issue you did.

Sounds good. Just to mention: I've seen a fair amount of service instability in testing HeyGen recently. Usually it manifests as lag, so I'm not sure if these are LiveKit errors or if it's in HeyGen's service.

PR #3030, opened 2025-11-11
feat: add session token support for AWS Nova Sonic

TamaryOmer:
## Summary
Adds support for AWS session tokens in the Nova Sonic speech-to-speech service integration, enabling 
authentication with temporary AWS credentials.

## Changes

- **Removed explicit credential parameters**: No longer requires secret_access_key, access_key_id, or 
session_token in constructor
- **Added environment credential resolver**: Uses EnvironmentCredentialsResolver to automatically detect 
credentials from: Environment variables (AWS_ACCESS_KEY_ID, AWS_SECRET_ACCESS_KEY, AWS_SESSION_TOKEN), IAM roles , AWS credential files or other standard AWS credential sources
• **Updated dependencies**: Added smithy-aws-core for credential resolution
• **Simplified configuration**: Removed manual credential management code

## Benefits
• **Automatic session token support**: Works seamlessly with temporary credentials
• **Follows AWS best practices**: Uses standard AWS credential chain
• **Reduced complexity**: No need to manually pass credentials
• **Better security**: Supports IAM roles and other secure credential methods

## Breaking Changes
• Constructor no longer accepts secret_access_key, access_key_id, or session_token parameters
• Credentials must now be provided via environment variables or other AWS standard methods

## Migration
Before:
python
```
llm = AWSNovaSonicLLMService(
    secret_access_key="...",
    access_key_id="...",
    session_token="...",  # optional
    region="us-east-1"
)
```

After:
```
python
# Set environment variables or use IAM roles
llm = AWSNovaSonicLLMService(
    region="us-east-1"  # region now defaults to us-east-1
)
```


markbackman:
Thanks for the submission. This unfortunately makes a breaking change that will impact developers using `AWSNovaSonicLLMService`. We're happy to support this pattern, but it needs to work with the current authentication scheme as well and can't be a breaking change. If that's something you're interested in submitting, we can review it.

markbackman:
Closing this as it's gone stale. If you're interested in taking this back up, see my previous comment about this being a breaking change.

PR #3031, opened 2025-11-11
Fix LLM tracing for LLMContext

markbackman:

This fix uses the context type to determine what information to use for tracing. When using the LLMContext, the adapters are used to convert to the native format for tracing.


markbackman:
Building on James's branch instead, here: https://github.com/pipecat-ai/pipecat/pull/3032

PR #3039, opened 2025-11-12
Update GeminiTTSService for streaming, other Google TTS improvements

markbackman:

- GeminiTTSService is now GA and to use the new `gemini-2.5-flash-tts` you need to switch from the google-genai` SDK to the `texttospeech_v1` SDK. This SDK is similar to what the GoogleTTSService uses, but different enough to warrant two separate classes. I've broken out the common functionality into `GoogleBaseTTSService`, then `GoogleTTSService` and `GeminiTTSService` are subclasses.
- This switch from `google-genai` to `texttospeech_v1` requires a breaking change as the two SDKs authenticate differently. But, I think this is unavoidable given that the SDKs are changing. We're just following suit.
- I found that the `GeminiTTSService` was implemented for batch processing (which was maybe all that google-genai supported?) which explains why it was so slow. Now that the `texttospeech_v1` is supported, there's an easy switch to use the streaming methods. This PR makes that change which drastically drops the latency.
- Lastly, I updated the voices for GeminiTTSService and the languages for the Google and Gemini services, as they're different (at the moment, at least).

Note: ideally, in the future, the GeminiTTSService class can go away and merge with the GoogleTTSService, but at the moment, they're different enough that keeping them separate makes sense.

### DIFF
--- a/src/pipecat/services/google/tts.py
+++ b/src/pipecat/services/google/tts.py
@@ -16,6 +16,7 @@ for natural voice control and multi-speaker conversations.
 
 import json
 import os
+import warnings
 
 from pipecat.utils.tracing.service_decorators import traced_tts
 
@@ -51,18 +52,12 @@ except ModuleNotFoundError as e:
     )
     raise Exception(f"Missing module: {e}")
 
-try:
-    from google import genai
-    from google.genai import types
-
-except ModuleNotFoundError as e:
-    logger.error(f"Exception: {e}")
-    logger.error("In order to use Gemini TTS, you need to `pip install pipecat-ai[google]`.")
-    raise Exception(f"Missing module: {e}")
-
 
 def language_to_google_tts_language(language: Language) -> Optional[str]:
     """Convert a Language enum to Google TTS language code.
+
+    Source:
+    https://docs.cloud.google.com/text-to-speech/docs/chirp3-hd
 
     Args:
         language: The Language enum value to convert.
@@ -71,9 +66,6 @@ def language_to_google_tts_language(language: Language) -> Optional[str]:
         The corresponding Google TTS language code, or None if not supported.
     """
     LANGUAGE_MAP = {
-        # Afrikaans
-        Language.AF: "af-ZA",
-        Language.AF_ZA: "af-ZA",
         # Arabic
         Language.AR: "ar-XA",
         # Bengali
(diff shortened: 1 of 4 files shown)


src/pipecat/transcriptions/language.py
```diff
@@ -66,6 +66,7 @@ class Language(StrEnum):
     AR_TN = "ar-TN"
     AR_XA = "ar-XA"
     AR_YE = "ar-YE"
+    AR_001 = "ar-001"
```
markbackman, line 69: New language codes supported by GoogleTTSService or GeminiTTSService.

src/pipecat/services/google/tts.py
```diff
@@ -71,9 +63,6 @@ def language_to_google_tts_language(language: Language) -> Optional[str]:
         The corresponding Google TTS language code, or None if not supported.
     """
     LANGUAGE_MAP = {
-        # Afrikaans
```
markbackman, line 74: These old codes were supported for older models, but in looking at Google's docs, they're moving away from older models and are highlighting (supporting?) only Chirp 3 (GoogleTTSService) and Gemini (GeminiTTSService).

Matching Google's documentation seems like a safe call.

src/pipecat/services/google/tts.py
```diff
@@ -741,56 +1032,60 @@ class GeminiTTSService(TTSService):
...
+        prompt: Optional[str] = None
         multi_speaker: bool = False
         speaker_configs: Optional[List[dict]] = None
 
     def __init__(
         self,
         *,
-        api_key: str,
-        model: str = "gemini-2.5-flash-preview-tts",
+        api_key: Optional[str] = None,
```
filipi87, line 1094: Is it worth keeping the `api_key` here? I believe that to use this service now, they’re required to provide the `credentials` and `credentials_path`, right? 

If that’s the case, it seems they’ll already need to update their implementation, because only providing the `api_key` would not work, so I think we could remove it.

filipi87: I think what I mean is that maybe we should only keep things marked as deprecated if they’re still working.

markbackman: In general, I agree with what you're saying. I was keeping api_key since it's a way to warn developers that the API has changed.

Do you have another suggestion for how to do that?

filipi87: I believe that in most IDEs, once they update the code, they would already see a warning since the signature has changed.

Otherwise, they will likely see an error when trying to execute the code, since the credentials argument would be missing.

But I am not opposed to leaving the `api_key` in place if you think it would help developers understand the required changes. After all, otherwise they would need to check the constructor signature themselves to figure out what’s missing. 👍

filipi87 (approved):
Looks good, nice to see the latency improvement. 👏🚀

src/pipecat/services/google/tts.py
markbackman: Right, that's the idea. Just leave the `api_key` in `__init__` temporarily so that developers can more easily find the change. We should be able to remove this after a few releases.

Issue #3034, opened 2025-11-11
[Feature Request] Support ElevenLabs scribe_v2_realtime STT model


KrishnatejaAnumolu21:
### Problem Statement

ElevenLabs just launched Scribe v2 Realtime STT (https://elevenlabs.io/blog/introducing-scribe-v2-realtime)

### Proposed Solution

It would be great if Pipecat could support it. Thanks in advance!

### Would you be willing to help implement this feature?

- [ ] Yes, I'd like to contribute
- [x] No, I'm just suggesting

markbackman:
We'll add support for this. It's on the to do list.

KrishnatejaAnumolu21:
> We'll add support for this. It's on the to do list.

Thank you!

abdulchotu7:
is this added

markbackman:
> is this added

Yes, it was released a few months ago in 0.0.95, as `ElevenLabsRealtimeSTTService`.

PR #3040, opened 2025-11-12
Add ElevenLabsRealtimeSTTService

markbackman:

Relevant docs used to implement:
https://elevenlabs.io/docs/cookbooks/speech-to-text/streaming

There's a python websocket server example included, which explains how this works.

Resolves https://github.com/pipecat-ai/pipecat/issues/3034

Also, the short VAD limitation has been resolved 🎉

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- Added `ElevenLabsRealtimeSTTService` which implements the Realtime STT
+  service from ElevenLabs.
+
 - Added a `TTSService.includes_inter_frame_spaces` property getter, so that TTS
   services that subclass `TTSService` can indicate whether the text in the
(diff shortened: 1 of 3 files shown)


apolinario:
> With a low VAD stop_secs value, frequently, I'm missing part of the transcript. I've reached out to the 11Labs team and they've confirmed that this is a known limitation to the model. They don't have a planned fix.

Out of curiosity, can we pair it with our own VAD + Smart Turn v3 and benefit from their transcription but use our own VAD+end-of-turn detection for snappiness?

markbackman:
> > With a low VAD stop_secs value, frequently, I'm missing part of the transcript. I've reached out to the 11Labs team and they've confirmed that this is a known limitation to the model. They don't have a planned fix.
> 
> Out of curiosity, can we pair it with our own VAD + Smart Turn v3 and benefit from their transcription but use our own VAD+end-of-turn detection for snappiness?

Indeed! That's using the "manual" commit strategy, which is the default and demonstrated in the 07d example.

louisjoecodes:
Woop! Excited to try this out.

PR #3041, opened 2025-11-12
Apply `includes_inter_frame_spaces = True` in all LLM and TTS service…

kompfner:
…s that need it.

Note that for `LLMTextFrame`s, the right behavior is pretty much always `includes_inter_frame_spaces = True`. I decided *not* to go ahead and make that the default for `LLMTextFrame`s, though, simply to not introduce a subtle behavior change for creative/unexpected use-cases that were relying on text in hand-crafted `LLMTextFrame`s being handled a certain way. Ditto for `TTSTextFrame`s.

Also, fix an issue in `NeuphonicTTSService` where it wasn't pushing `TTSTextFrame`s.

Also, fix the broken `SarvamHttpTTSService` example.

Also, add a couple of missing examples.

## Services validated by hand

* LLM services
  - [x] Anthropic
  - [x] AWS
  - [x] Azure (relies on OpenAI service under the hood for LLMTextFrame creation)
  - [x] Azure Realtime (ah right, still can’t test because we still don’t have model access)
  - [x] Cerebras (relies on OpenAI service under the hood for LLMTextFrame creation)
  - [x] Deepseek (relies on OpenAI service under the hood for LLMTextFrame creation)
  - [x] Fireworks (relies on OpenAI service under the hood for LLMTextFrame creation)
  - [x] Google
  - [x] Google Vertex (relies on Google service under the hood for LLMTextFrame creation)
  - [x] Google OpenAI (deprecated, so won’t do)
  - [x] Gemini Live LLM text
  - [x] Grok (relies on OpenAI service under the hood for LLMTextFrame creation)
  - [x] Groq (relies on OpenAI service under the hood for LLMTextFrame creation)
  - [x] Mistral (relies on OpenAI service under the hood for LLMTextFrame creation)
  - [x] NIM (relies on OpenAI service under the hood for LLMTextFrame creation)
  - [x] OLLama (relies on OpenAI service under the hood for LLMTextFrame creation)
  - [x] OpenAI
    - [x] Didn’t test w/`gpt-4o-audio-preview` model, as it’s pretty challenging to validate (it requires some machinery for feeding audio, rather than text, into the LLM). Based on discussion in https://github.com/pipecat-ai/pipecat/pull/1784, though, not sure this rises to “officially supported” (for most users, OpenAI Realtime may supplant this model).
  - [x] OpenAI Realtime LLM text
  - [x] Openpipe (relies on OpenAI service under the hood for LLMTextFrame creation)
  - [x] Openrouter (relies on OpenAI service under the hood for LLMTextFrame creation)
  - [x] Perplexity (relies on OpenAI service under the hood for LLMTextFrame creation)
  - [x] Qwen (relies on OpenAI service under the hood for LLMTextFrame creation)
  - [x] Sambanova
    - [x] See previous note about `gpt-4o-audio-preview`
  - [x] Together (relies on OpenAI service under the hood for LLMTextFrame creation)
* TTS services
  - [x] AsyncAI
  - [x] AsyncAIHTTP
  - [x] AWS
  - [x] AWS Nova Sonic (note: only outputs TTS text, not LLM text, since no text-only mode)
  - [x] Azure
  - [x] Azure HTTP
  - [x] Cartesia
  - [x] Deepgram
  - [x] DeepgramHTTP
  - [x] Elevenlabs
  - [x] Fish
  - [x] Google
  - ❌ Google HTTP (looks like this one isn’t working properly; it’s only sending one giant text frame, no matter how long repsonse is, with missing punctuation)
  - [x] Gemini
  - [x] Gemini Live TTS text
  - [x] Groq
  - [x] Hume
  - [x] Inworld
  - [x] LMNT
  - [x] Minimax
  - [x] Neuphonic
  - [x] Neuphonic HTTP
  - ❌ OpenAI (looks like this one isn’t working properly; it’s only sending one giant text frame, no matter how long response is, with missing punctuation)
  - [x] OpenAI Realtime TTS text
  - [x] Piper
  - [x] PlayHT (no need to update, as it’s deprecated)
  - [x] PlayHT HTTP (no need to update, as it’s deprecated)
  - [x] Rime
  - [x] Rime HTTP
  - [x] Riva
  - [x] Sarvam
  - [x] Sarvam HTTP
  - [x] Speechmatics
  - [ ] XTTS (could not test; no NVIDIA driver found on system)
* Vision services
  - [x] Moondream

### DIFF
--- a/src/pipecat/services/asyncai/tts.py
+++ b/src/pipecat/services/asyncai/tts.py
@@ -146,6 +146,15 @@ class AsyncAITTSService(InterruptibleTTSService):
         """
         return True
 
+    @property
+    def includes_inter_frame_spaces(self) -> bool:
+        """Indicates that AsyncAI TTSTextFrames include necessary inter-frame spaces.
+
+        Returns:
+            True, indicating that AsyncAI's text frames include necessary inter-frame spaces.
+        """
+        return True
+
     def language_to_service_language(self, language: Language) -> Optional[str]:
         """Convert a Language enum to Async language format.
 
@@ -417,6 +426,15 @@ class AsyncAIHttpTTSService(TTSService):
 
         Returns:
             True, as Async HTTP service supports metrics generation.
+        """
+        return True
+
+    @property
+    def includes_inter_frame_spaces(self) -> bool:
+        """Indicates that AsyncAI TTSTextFrames include necessary inter-frame spaces.
+
+        Returns:
+            True, indicating that AsyncAI's text frames include necessary inter-frame spaces.
         """
         return True
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,19 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Added a `TTSService.includes_inter_frame_spaces` property getter, so that TTS
+  services that subclass `TTSService` can indicate whether the text in the
+  `TTSTextFrame`s they push already contain any necessary inter-frame spaces.
+
 ### Fixed
+
+- Fixed subtle issue of assistant context messages ending up with double spaces
+  between words or sentences.
+
+- Fixed an issue where `NeuphonicTTSService` wasn't pushing `TTSTextFrame`s,
+  meaning assistant messages weren't being written to context.
 
 - Fixed an issue with OpenTelemetry where tracing wasn't correctly displaying
   LLM completions and tools when using the universal `LLMContext`.
--- a/examples/foundational/07f-interruptible-azure-http.py
+++ b/examples/foundational/07f-interruptible-azure-http.py
@@ -1,0 +1,135 @@
+#
+# Copyright (c) 2024–2025, Daily
+#
+# SPDX-License-Identifier: BSD 2-Clause License
+#
+
+
+import os
+
+from dotenv import load_dotenv
+from loguru import logger
+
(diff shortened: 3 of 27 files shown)


kompfner:
I did not yet investigate Google HTTP and OpenAI TTS services. They both seem to be going slightly awry in very much the same way (see notes in PR description). This could be an investigation for a follow-up PR.

src/pipecat/services/asyncai/tts.py
```diff
@@ -146,6 +146,15 @@ def can_generate_metrics(self) -> bool:
         """
         return True
 
+    @property
```
markbackman, line 149: Rather than define this property everywhere, could we define it once in the base TTSService class then have the subclasses set it when they call `super().__init__`?

This would follow the same pattern as the sample_rate property. It would consolidate a lot of the code to just the base class.

markbackman: Ok, I see you have the base class property already. Perhaps we don't want this property to be exposed in `__init__`. In which case, what you have makes sense.

kompfner: > Perhaps we don't want this property to be exposed in __init__

Yeah, my instinct for these kinds of things–properties that should be defined by subclasses and not application code—is that they should be immutable getters, which make it impossible for application code to override.

But then again, we do have precedent, with things like `push_text_frames`, of properties that should only be set by subclasses...

Actually, we have precedent both ways in our codebase. So...I guess I don't feel strongly one way or another.

PR #3042, opened 2025-11-12
Follow-up to #3041

kompfner:
This is best understood commit by commit:

## Commit 1

Override `includes_inter_frame_spaces` in:
- `GoogleHttpTTSService`
- `OpenAITTSService`

The reason I had skipped this work in an [earlier PR](https://github.com/pipecat-ai/pipecat/pull/3041) was because these services seemed to be emitting long, punctuation-free text frames. It turns out that the issue was with the LLM prompt, though, resulting in the LLM nondeterministically excluding all punctuation. An upcoming commit will address that prompt issue.

## Commit 2

Refine our examples' LLM prompts to avoid the issue mentioned in Commit 1.

### DIFF
--- a/examples/foundational/04-transports-small-webrtc.py
+++ b/examples/foundational/04-transports-small-webrtc.py
@@ -77,7 +77,7 @@ async def run_example(webrtc_connection: SmallWebRTCConnection):
     messages = [
         {
             "role": "system",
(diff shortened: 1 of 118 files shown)


kompfner:
Hmm, I'm actually still seeing the missing punctuation issue. Tweaking the prompt again before proceeding.

kompfner:
OK this tweak seems to have done the trick.

PR #3045, opened 2025-11-12
introducing user turn strategies

aconchillo:
## 🧑🤖💬 Introducing user turn strategies

This PR introduces **user turn strategies**, providing a unified and extensible way to determine **when a conversational user turn begins and ends**.

Turn strategies replace implicit and scattered “interrupt” logic with clear, composable policies that work across VAD, transcription-based signals, and turn analysis models.

### User turn start strategies

User turn start strategies indicate **when the user has started speaking**. These are typically used to interrupt the bot or to mark the beginning of a user turn.

Examples include:

* Voice activity detection (VAD)
* Transcription-based detection (including interim results)
* Minimum spoken word thresholds

Available strategies:

* **`VADUserTurnStartStrategy`**
* **`TranscriptionUserTurnStartStrategy`**
* **`MinWordsUserTurnStartStrategy`**

### User turn stop strategies

User turn stop strategies indicate **when the user stops speaking**, which can then be used to determine that the bot turn starts.

These strategies account for real-world conditions where:

* Transcriptions may arrive late or in bursts
* VAD alone may not signal clear turn boundaries
* Turn completion may require model-based reasoning

Available strategies:

* **`TranscriptionUserTurnStopStrategy`**
* **`TurnAnalyzerUserTurnStopStrategy`**

## 🧠 Design goals

* Make turn-taking **explicit and configurable**
* Support multiple strategies
* Decouple turn detection from pipeline internals
* Align terminology with common conversational AI concepts

## 🔧 Configuration

Turn strategies are configured when creating a `PipelineTask`, for example:

```python
task = PipelineTask(
    ...,
    params=PipelineParams(
            stop=[
                    turn_analyzer=LocalSmartTurnAnalyzerV3()
                )
            ],
        ),
    ),
)
```

## 🔄 Migration notes

* To use these strategies, pipelines should migrate to the new **universal `LLMContext`** and **`LLMContextAggregatorPair`**. Not compatible with the previous `OpenAILLMContext`.
* Don't pass `turn_analyzer` to the transport. Use a `TurnAnalyzerUserTurnStopStrategy` instead.
* Don't use deprecated interruption strategies. Use user turn start strategies instead.

UPDATE: User and bot turn strategies are now passed to the `LLMUserAggregator` (see #3296). The reasoning is that a PipelineTask shouldn't have specifics to vocie AI and should be more generic. Moving it to `LLMUserAggregator` makes it cleaner and  more specific to voice pipelines.

```
        context,
                bot=[TurnAnalyzerUserTurnStopStrategy(turn_analyzer=LocalSmartTurnAnalyzerV3())]
            ),
        ),
    )
```

UPDATE: Bot turn strategies are now called user stop strategies. See https://github.com/pipecat-ai/pipecat/pull/3325. This PR description gas been update accordingly.


src/pipecat/turns/turn_analyzer_speaking_strategy.py
```diff
@@ -0,0 +1,175 @@
...
+            try:
+                await asyncio.wait_for(self._event.wait(), timeout=self._timeout)
+                self._event.clear()
+                last_update_ts = time.time()
+            except asyncio.TimeoutError:
+                if not self._text:
+                    continue
+
+                silent_for = time.time() - last_update_ts
+                if match_endofsentence(self._text) or silent_for >= SILENCE_TIMEOUT:
```
aconchillo: It occurred to me that checking for end of sentence also allowed us to use semantic to avoid false positives. This has actually worked really well. Before we were sending requests to the LLM all the time.

changelog/3045.added.md
```diff
@@ -0,0 +1,32 @@
+- Introducing user and bot turn start strategies. Turn start strategies are used
+  to indicate when user and bot turns begin. User turn start strategies indicate
+  when the user starts speaking (e.g. using VAD events or when a user says one
+  or more words). Bot turn start strategies indicate when the bot should start
+  speaking (e.g. using an end-of-turn detection model or by observing incoming
+  transcriptions). In conversational agents, these are often referred to as
+  start/stop speaking or turn-taking plans/policies.
+
+  Available interruption strategies:
```
kompfner: There are others now, too, worth mentioning here. Two of them in conjunction are default, right? (Transcription + VAD)

changelog/3045.added.md
```diff
@@ -0,0 +1,32 @@
...
+    - VADUserTurnStartStrategy (default)
+
+  Avaliable speaking strategies:
```
kompfner: In this changelog entry, seems like we still use the terms "interruption" and "speaking" for "user turn start" and "bot turn start"

src/pipecat/turns/bot/transcription_bot_turn_start_strategy.py
```diff
@@ -0,0 +1,126 @@
...
+
+class TranscriptionBotTurnStartStrategy(BaseBotTurnStartStrategy):
+    """Bot turn start strategy based on transcriptions.
+
+    This strategy assumes the bot should start speaking once a transcription
+    has been received and the user is not actively speaking. It handles
+    multiple or delayed transcription frames gracefully.
+    """
+
+    def __init__(self, *, timeout: float = 0.4):
```
kompfner: Pardon me if you're already working on this, but from my notes based on our synchronous chat, here's what remains to do in this strategy:

* Introduce separate `transcript_delay_timeout` (default 0.4) and `user_resume_speaking_timeout` (default 0.6) values rather than the single `timeout` value
* Honor both timeout values when updating internal state, so that we say the bot turn has started when we've determined that:
  * The user has probably finished speaking, and
  * There are probably no more transcripts to receive

kompfner: Same goes for the `TurnAnalyzerBotTurnStartStrategy`

src/pipecat/turns/bot/transcription_bot_turn_start_strategy.py
kompfner, line 1: File name doesn't match others. Should be `transcription_bot_turn_start_strategy.py`

changelog/3045.added.md
```diff
@@ -0,0 +1,32 @@
+- Introducing user and bot turn start strategies. Turn start strategies are used
+  to indicate when user and bot turns begin. User turn start strategies indicate
+  when the user starts speaking (e.g. using VAD events or when a user says one
+  or more words). Bot turn start strategies indicate when the bot should start
+  speaking (e.g. using an end-of-turn detection model or by observing incoming
```
kompfner: It occurs to me that how we're talking about the distinction between turn-analyzer-based and transcription-based bot-turn-start detection isn't quite right.

The turn-analyzer-based strategy decides it's the bot turn when it determines:
* The user has probably finished speaking (based on the turn analyzer signal)
* We've probably received the last relevant transcription

The transcription-based strategy decides that it's bot's turn when it determines:
* The user has probably finished speaking (based on VAD + a timeout)
* We've probably received the last relevant transcription

So the only difference between the two strategies is in how it determines the user has probably finished speaking:
* VAD + a timeout
* turn analyzer

To avoid misleading the reader, we should make the distinction clear...maybe, since transcription is the common denominator, we actually have:
* `VADBotTurnStartStrategy`
* `TurnAnalyzerBotTurnStartStrategy`

And then all the determining-if-we've-probably-gotten-the-last-transcription logic can be pulled into a shared base class between the two strategies.

(This change wouldn't be just to the class and file names; we'd have to update how we talk about these strategies and the distinction between them, in this changelog entry and the strategies' docstrings.)

WDYT?

src/pipecat/turns/bot/transcription_bot_turn_start_strategy.py
kompfner: Per my last comment in the changelog entry, perhaps `transcript_delay_timeout` would go in the new superclass shared between the two bot turn start strategies.

changelog/3045.added.md
```diff
@@ -0,0 +1,32 @@
...
+  transcriptions). In conversational agents, these are often referred to as
+  start/stop speaking or turn-taking plans/policies.
+
+  Available interruption strategies:
+    - VADUserTurnStartStrategy (default)
+
+  Avaliable speaking strategies:
+    - TranscriptionBotTurnStartStrategy (default)
+    - TurnAnalyzerBotTurnStartStrategy
```
kompfner, line 16: Haven't read the code yet, so pardon me if this comment is just noise, but: would it be possible to detect that the user has configured a turn analyzer on their transport (the old way of specifying a turn analyzer) and if so default to using this strategy?

changelog/3045.deprecated.md
kompfner, line 1: When we make the change to make the VAD `stop_secs` default to 0.2 (which...is the plan to add it to this PR?), we should call it out prominently in the changelog somehow...

kompfner: > is the plan to add it to this PR

I would push for including in this PR the change moving the 0.6 second user-probably-stopped-speaking-timeout from being baked into `stop_secs` to being the default user speaking timeout in the bot turn start strategy. The reason to include it in this PR is because without it things would get temporarily _harder_ for the user: the dependent things (`stop_secs` and whether or not the use the turn analyzer) move further apart.

src/pipecat/turns/bot/transcription_bot_turn_start_strategy.py
aconchillo: Yes. Still need to work on this.

src/pipecat/turns/bot/transcription_bot_turn_start_strategy.py
aconchillo: Thanks! Lots of renaming :sweat_smile:

changelog/3045.deprecated.md
aconchillo: Yes, it will be included in this PR.

changelog/3045.added.md
aconchillo: This is currently not possible. We really don't know what's the configuration of processors, unless we push some frames (which we do in some cases). But, I believe it's better to make things more explicit. Also, having the turn analyzer in the transport wouldn't make sense any more as it would not be used at all,except for this frame to inform the rest of the pipeline but that would feel weird.

changelog/3045.added.md
kompfner: Per our chat: this change will come in a follow-up PR

changelog/3045.added.md
kompfner: Per our chat: if the user had previously been using a turn analyzer on the transport, turn strategies will be disabled (and a warning to that effect will be logged)

changelog/3045.deprecated.md
kompfner: Per our chat: this'll come in a follow-up PR

changelog/3045.deprecated.2.md
```diff
@@ -0,0 +1 @@
+- `TransportParams.turn_analyzer` is deprecated, use the new user and bot turn start strategies.
```
kompfner: In these deprecation messages, it might be helpful to the reader if we referenced the new parameter by name, like:


```suggestion
- `TransportParams.turn_analyzer` is deprecated, use `PipelineTask`'s new `turn_start_strategies` parameter instead.
```

And maybe even follow it up with the specific strategy replacing the deprecated thing:

```suggestion
- `TransportParams.turn_analyzer` is deprecated, use `PipelineTask`'s new `turn_start_strategies` parameter instead; specifically, use the `TurnAnalyzerBotTurnStartStrategy` to use the turn analyzer to trigger bot responses.
```

src/pipecat/audio/interruptions/min_words_interruption_strategy.py
```diff
@@ -17,6 +17,9 @@ class MinWordsInterruptionStrategy(BaseInterruptionStrategy):
     This is an interruption strategy based on a minimum number of words said
     by the user. That is, the strategy will be true if the user has said at
     least that amount of words.
+
+    .. deprecated:: 0.0.98
```
kompfner: Bump deprecation version, as 0.0.98 has already shipped.

src/pipecat/audio/interruptions/min_words_interruption_strategy.py
```diff
@@ -17,6 +17,9 @@ class MinWordsInterruptionStrategy(BaseInterruptionStrategy):
...
+        This class is deprecated, use the new user and bot turn start strategies.
```
kompfner: Per my comment in the changelog: do we want to call out the new field more explicitly by name to help the user find it?

src/pipecat/audio/interruptions/min_words_interruption_strategy.py
kompfner: (This feedback applies throughout PR)

src/pipecat/pipeline/task.py
```diff
@@ -278,6 +292,14 @@ def __init__(
                 additional_span_attributes=self._additional_span_attributes,
             )
             observers.append(self._turn_trace_observer)
+
+        # Initialize default user and bot turn start strategies.
+        if not self._params.turn_start_strategies:
+            self._params.turn_start_strategies = TurnStartStrategies(
+                user=[VADUserTurnStartStrategy(), TranscriptionUserTurnStartStrategy()],
```
kompfner: Defaults should be _both_ VAD and Transcription, I believe

src/pipecat/processors/aggregators/llm_response_universal.py
```diff
@@ -222,36 +241,21 @@ def __init__(
         """
         super().__init__(context=context, role="user", **kwargs)
         self._params = params or LLMUserAggregatorParams()
-        self._vad_params: Optional[VADParams] = None
-        self._turn_params: Optional[SmartTurnParams] = None
 
-        if "aggregation_timeout" in kwargs:
```
kompfner, line 228: So nice to see so many things being simplified away

src/pipecat/processors/aggregators/llm_response_universal.py
```diff
@@ -65,15 +57,42 @@
...
+
+    Parameters:
+        enable_user_speaking_frames: If True, the aggregator will emit frames
+            indicating when the user starts and stops speaking, as well as
+            interruption frames. This is enabled by default, but you may want
+            to disable it if another component (e.g., an STT service) is already
+            generating these frames.
+    """
+
+    enable_user_speaking_frames: bool = True
```
kompfner, line 80: By moving `LLMUserAggregatorParams` into this file and swapping out the previous set of fields (`aggregation_timeout`, `turn_emulated_vad_timeout`, etc) for this new field, it seems like we're introducing a breaking change without a deprecation warning...right? Am I understanding correctly?

src/pipecat/processors/aggregators/llm_response_universal.py
```diff
@@ -443,101 +375,52 @@ async def _handle_transcription(self, frame: TranscriptionFrame):
...
+
+    async def _trigger_user_turn_start(self, strategy: BaseUserTurnStartStrategy):
+        # Reset all user turn start strategies to start fresh.
+        if self.turn_start_strategies:
+            for s in self.turn_start_strategies.user:
+                await s.reset()
+
+        if self._params.enable_user_speaking_frames:
+            logger.debug(f"User started speaking (user turn start strategy: {strategy})")
+            await self.broadcast_frame(UserStartedSpeakingFrame, emulated=strategy is None)
```
kompfner: Do we want to track the TODO here of, instead of broadcasting these frames, telling the pipeline to send down the frame from the source/top of the pipeline, so that things get interrupted in order?

src/pipecat/processors/aggregators/llm_response_universal.py
```diff
@@ -443,101 +375,52 @@ async def _handle_transcription(self, frame: TranscriptionFrame):
...
+            await self.push_frame(InterruptionFrame())
```
kompfner: Do we want to push the interruption from this processor, or do we want to send it the same way as `UserStartedSpeakingFrame`, above?

src/pipecat/processors/aggregators/llm_response_universal.py
kompfner: Not sure I follow the `emulated=strategy is None` logic...it should be considered emulated if it's not triggered by the VAD strategy, right?

src/pipecat/processors/aggregators/llm_response_universal.py
```diff
@@ -443,101 +375,52 @@ async def _handle_transcription(self, frame: TranscriptionFrame):
...
+    async def _on_broadcast_frame(
+        self,
+        strategy: BaseUserTurnStartStrategy | BaseBotTurnStartStrategy,
+        frame_cls: Type[Frame],
+        **kwargs,
+    ):
+        await self.broadcast_frame(frame_cls, **kwargs)
+
+    async def _trigger_user_turn_start(self, strategy: BaseUserTurnStartStrategy):
+        # Reset all user turn start strategies to start fresh.
```
kompfner, line 442: This is what guarantees that only the _first_ of the set of provided strategies to be triggered has an effect, right?

src/pipecat/services/stt_service.py
```diff
@@ -252,20 +252,15 @@ async def process_frame(self, frame: Frame, direction: FrameDirection):
         """Process frames, handling VAD events and audio segmentation."""
         await super().process_frame(frame, direction)
 
-        if isinstance(frame, UserStartedSpeakingFrame):
+        if isinstance(frame, VADUserStartedSpeakingFrame):
             await self._handle_user_started_speaking(frame)
-        elif isinstance(frame, UserStoppedSpeakingFrame):
+        elif isinstance(frame, VADUserStoppedSpeakingFrame):
```
kompfner, line 257: So now segmented STT services will generally work on potentially smaller chunks of audio input, right? 

That's because raw VAD events will be by default more sensitive (especially once we make the change to default them to `stop_secs=2`), since the more "semantic" judgment of "did the user's turn end" (as opposed to "did voice-like audio stop") has been pushed further downstream, to the user aggregator?

Will this negatively impact how well segmented STT services work?

src/pipecat/processors/aggregators/llm_response_universal.py
aconchillo: Yes, this should come from the event instead of here. That is, the user start turn transcription strategy should set this to true.

aconchillo: I'll add a TODO.

src/pipecat/processors/aggregators/llm_response_universal.py
aconchillo: Yeah, this should be broadcast with the TODO mentioned above.

src/pipecat/audio/interruptions/min_words_interruption_strategy.py
aconchillo: Yep!

changelog/3045.deprecated.2.md
aconchillo: Done!

src/pipecat/audio/interruptions/min_words_interruption_strategy.py
aconchillo: Done!

src/pipecat/pipeline/task.py
aconchillo: Not sure I understand. That's what these are already. If VAD fails, we have `TranscriptionUserTurnStartStrategy`

src/pipecat/processors/aggregators/llm_response_universal.py
aconchillo: Done

src/pipecat/processors/aggregators/llm_response_universal.py
aconchillo: Done

src/pipecat/processors/aggregators/llm_response_universal.py
aconchillo: Yes. Also, the guarantee is that we provide frames to the strategies in order and the first one that triggers the turn does it through a synchronous event. Because it's synchronous, the other strategies won't get really executed and when they do they might have been reset.

src/pipecat/services/stt_service.py
aconchillo: Once we change the VAD stops_secs to 0.2, we will need to check.  But I think it should be OK.

aconchillo: This will be in an upcoming PR, so for now we are OK :smile:

src/pipecat/processors/aggregators/llm_response_universal.py
aconchillo: Actually, I think emulated should be completely removed. I don't think there's a use in knowing this anymore.

aconchillo: Done.

src/pipecat/processors/aggregators/llm_response_universal.py
aconchillo: Actually yes. If someone used `LLMUserAggregatorParams` they might run into some issues, since `enable_user_speaking_frames` is not there. One option is to add `enable_user_speaking_frames` so everything keeps working and deprecate the old `LLMUserAggregatorParams`. I'll do that.

aconchillo: Done. Now, if you use the old `LLMUserAggregatorParams` nothing will be broken. Those are now deprecated.

changelog/3045.added.md
aconchillo: Done!

changelog/3045.added.md
aconchillo: Done!

changelog/3045.added.md
aconchillo: I tried it and it doesn't work. If we disable strategies user turn start will work but not the turn start since we are missing the transcription bot turn start strategy.

So, unfortunately, they will just get a yellow warning + deprecation. I don't think we should add more code to handle only turn analyzers.

aconchillo: Things keep working though, except you might get duplicate `UserStartSpeakingFrame` and `UserStoppedSpeakingFrame`.

kompfner: > since we are missing the transcription bot turn start strategy

Ohh you're saying that because we ripped out the user aggregator logic that waits for a transcription (that logic has been moved to the bot turn start strategies), things won't quite work as expected (i.e. as they worked before). I see I see.

kompfner: > Things keep working though

But they will _not_ work as the user wants, right? If the user had previously specified the smart turn analyzer in their input transport, they were telling us that they _didn't_ want the default VAD-plus-timeout-driven behavior, they wanted something "smarter".

With the change in this PR, because our default bot turn start strategy is the VAD-plus-timeout-driven one, they will "regress" to the "less smart" behavior they were trying to avoid by specifying the smart turn analyzer.

Right? Or am I missing something?

kompfner: Ohhhh, or does the presence of the smart turn analyzer in the input transport actually affect when the raw `VADUserStoppedSpeaking` frame is sent? Because then we might be OK...

src/pipecat/pipeline/task.py
kompfner: Weird, I swear I only saw one. Yes, this looks right.

src/pipecat/processors/aggregators/llm_response_universal.py
kompfner: So...this is a (bit) of a breaking change, is it not? Now if someone is using a pipeline in its default configuration and cared about reading the `emulated` value out of `User*SpeakingFrame`s, they'd...what? Only ever see the default `false` value?

kompfner: > I don't think there's a use in knowing this anymore.

Out of curiosity, why was it useful before?

I get that now that there can be arbitrarily many strategies to trigger the user turn start, a boolean is unsuitable for describing which one actually triggered it.

If this information _is_ still useful on the frame for some reason, though, we _could_ consider supplanting the boolean field for one that contains an identifier for the strategy.

src/pipecat/services/stt_service.py
kompfner: > But I think it should be OK.

Out of curiosity, why do you think it'll be OK?

It seems to me that folks using segmented STT will almost certainly be impacted, as there'll suddenly be much less "buffer time" and their STT services will be getting much smaller audio chunks, on every user hesitation.

Should we update how `SegmentedSTTService` works, to not listen to raw VAD input but instead be triggered when the user turn ends according to our strategies? (Which would mean...what? Continuing to use `UserStoppedSpeakingFrame` as the signal instead of `VADUserStoppedSpeakingFrame`?)

changelog/3045.deprecated.2.md
```diff
@@ -0,0 +1 @@
+- `TransportParams.turn_analyzer` is deprecated, use `PipelineTask`'s new `turn_start_strategies` parameter instead.
```
kompfner: Not only is `turn_analyzer` deprecated, it actually won't have an effect anymore (when `LLMContext` is used, which is recommended)

changelog/3045.added.md
kompfner: > Because then we might be OK...

Smart turn analyzer does *not* affect `VADUserStoppedSpeaking` timing. So we need to make this breaking change quite visible (see comment in changelog)

kompfner: From convo: since we'd be breaking smart turn for everyone who's following our best practices (using `LLMContext`), the thing to do is send a frame down from the transport saying "hey, automatically switch to the smart turn bot start strategy"). This would be a temporary, for-backwards-compatibility-only mechanism that would be accompanied by very visible deprecation warnings ("please please use the new turn strategies instead of the smart turn analyzer on the transport")

kompfner: Latest thinking:

Print a (very visible) error in the universal user aggregator saying something to the effect of "hey, your transport-specified smart turn analyzer no longer has an effect. use the new turn-analyzer-based bot turn start strategy instead"

src/pipecat/processors/aggregators/llm_response_universal.py
kompfner: > we could consider supplanting the boolean field for one that contains an identifier for the strategy

...but let's not bother with this right now

src/pipecat/services/stt_service.py
kompfner: In other words: the problem is that `SegmentedSTTService` is now upstream from the "did the user stop speaking" logic, whereas it used to be downstream.

kompfner: One option, if we need it: just hard-code some "lightweight" did-the-user-stop-speaking logic here: just an arbitrary  (or configurable) buffer after VAD detects end of voice activity.

Not super clean and not as configurable as our new turn strategies, but might be good enough for now.

Maybe the default buffer should be 0.6, which is the difference between the previous default VAD `stop_secs` and the forthcoming one.

kompfner (approved):
🥳

changelog/3045.deprecated.2.md
aconchillo: Updated.

changelog/3045.added.md
aconchillo: Done.

src/pipecat/processors/aggregators/llm_response_universal.py
aconchillo: I have added two events `on_user_turn_started` and `on_bot_turn_started` to the user aggregator. This way people will know what strategy triggered the turn start.

src/pipecat/services/stt_service.py
aconchillo: We will do some tests once we update the VAD stop secs to 0.2 and decide then.

divyanshjain-git:
Hi @aconchillo we tested this with 5% of our users for a day. There is just one issue that came up.

MinWordsUserTurnStartStrategy applies the min_words filter unconditionally - even when the bot is NOT speaking. This causes single-word responses like "yes", "no", "ok" to be ignored during normal conversation (not just during interruptions), breaking basic conversational flow.

Root Cause: Unlike TranscriptionUserTurnStartStrategy which tracks _bot_speaking state and only triggers during interruptions, MinWordsUserTurnStartStrategy lacks this awareness.

Proposed Solution: Add a SmartMinWordsUserTurnStartStrategy (or enhance the existing one) that:
When bot is NOT speaking → triggers on any word (1+) for normal conversation
When bot IS speaking → requires min_words to interrupt (stricter interruption filtering)

Why This Matters: Users configuring MinWordsUserTurnStartStrategy(min_words=3) expect it to prevent accidental interruptions, but instead it completely breaks responsiveness to short conversational replies. 

Let me know if i can help with the PR or if you want me to open an issue.

markbackman:
@divyanshjain-git you're right. We'll fix this.

aconchillo:
> @divyanshjain-git you're right. We'll fix this.

Thank you so much @divyanshjain-git for testing this new feature! This should be resolved here: https://github.com/pipecat-ai/pipecat/pull/3284

larry-cmz:
This looks great.
One use case I've been curious about is where we don't allow an emulated (transcription-only) user-start-of-turn to interrupt but we _do_ include that transcribed user speech in the transcript moving forward so it can inform LLM responses.

i.e. the user's speech doesn't interrupt the agent, but the agent still "hears" the user

divyanshjain-git:
@aconchillo Thanks for the speedy resolution. Happy to help.

markbackman:
> This looks great. One use case I've been curious about is where we don't allow an emulated (transcription-only) user-start-of-turn to interrupt but we _do_ include that transcribed user speech in the transcript moving forward so it can inform LLM responses.
> 
> i.e. the user's speech doesn't interrupt the agent, but the agent still "hears" the user

@larry-cmz this level of control will be easier to have with this change. For example, you could use a user start strategy of `VADUserTurnStartStrategy`, which makes the VAD the sole determining factor for a user turn start.

xdevfaheem:
this is pretty cool. I've been eyeing for the v0.0.99 release for the past couple of days... when can we expect?

markbackman:
> this is pretty cool. I've been eyeing for the v0.0.99 release for the past couple of days... when can we expect?

The goal is a release this week. TBD on the exact timing.

PR #3052, opened 2025-11-13
Include pts in video and audio frames in SmallWebRTCClient

martinxsliu:
The incoming video and audio frames didn't have their `pts` property set by the SmallWebRTCClient. This PR sets them.


markbackman:
Tagging @aconchillo and @filipi87 to take a look.

Issue #3023, opened 2025-11-11
Bot message before tool call is not added to context, leading to bot repeating itself.


gitraffe:
### pipecat version

0.0.94

### Python version

3.13.6

### Operating System

macOS 26.0.1 (also reproduced in python:3.12-slim Docker container)

### Issue description

When a model receives an instruction and first produces a text response and then calls a tool, the first text response is not added to the context aggregator, but the text is spoken. Because the text is not in the context, the language model will comply with the original instruction, leading to the bot repeating itself.

I've mostly been able to work around this by aggressively prompting the language model to always make tool calls before speaking any text, but it's not 100% reliable.

Hypothesis: This might be a race condition because the relevant frame doesn't reach the aggregator (e.g., because TTS takes a bit of time) before the harness feeds back the tool call result back to the language model.

### Reproduction steps

1. Run the snippet below using `SERVICE_ACCOUNT_PATH=... GEMINI_API_KEY=... python simple_chat_bot.py --transport webrtc` and connect on http://localhost:7860/client/. `SERVICE_ACCOUNT_PATH` is the path to a Google Service Account json file and `GEMINI_API_KEY` is the standard Gemini key.
2. Instruct the bot with "Say the word banana. Then call get current time." (`get_current_time` is a tool available to the agent) **Edit**: Probably saying "First, say the word banana. Second, call get current time." is a slightly more reliable prompt.
3. The bot will respond "Banana. Banana. The current time is ..."

This doesn't happen the other way round.

1. Instruct the both with "First, call get current time. Second, say the word banana. Third, tell me the time."
2. The bot will respond "Banana. The current time is ...".

```python
#
# Simple chat bot based on Pipecat quickstart example.
# Uses Google services (STT, LLM, TTS) and includes get_current_time() function.
#

"""Simple chat bot for testing function calling behavior.

Run the bot using::

    python playground/simple_chat_bot.py --transport webrtc

Then open http://localhost:7860 in your browser to connect.
"""

import os
from datetime import datetime
from dotenv import load_dotenv
from loguru import logger
from pipecat.audio.turn.smart_turn.local_smart_turn_v3 import LocalSmartTurnAnalyzerV3
from pipecat.audio.vad.silero import SileroVADAnalyzer
from pipecat.adapters.schemas.tools_schema import ToolsSchema
from pipecat.audio.vad.vad_analyzer import VADParams
from pipecat.pipeline.pipeline import Pipeline
[...]

PR #3053, opened 2025-11-13
Ensure that the function call results respect the previous LLM context.

filipi87:
This PR fixes a race condition where the LLM receives instructions to both produce text and invoke a function call at the same time. In those cases, the context would not be updated before the function call result arrived, causing the bot to repeat itself.

For example, the prompt “First, say the word banana. Second, call get_current_time.”, could lead to the bot responding, “Banana. Banana. The current time is …”.

All the details are described in this issue: https://github.com/pipecat-ai/pipecat/issues/3023

The root cause is that all frames related to the function call are `SystemFrames`, while the text frames generated by the LLM are not. This caused the function call result to reach the `LLMAssistantAggregator` before the context was updated with the previous text responses.

This PR fixes that by ensuring that if a function call result is received while an LLM response is still in progress, we wait before adding the function call result to the context and triggering a new inference.

### DIFF
--- a/src/pipecat/processors/aggregators/llm_response_universal.py
+++ b/src/pipecat/processors/aggregators/llm_response_universal.py
@@ -805,7 +805,12 @@ class LLMAssistantAggregator(LLMContextAggregator):
             del self._function_calls_in_progress[frame.tool_call_id]
 
     def _update_function_call_result(self, function_name: str, tool_call_id: str, result: Any):
-        for message in self._function_calls_context_messages:
+        def iter_all():
+            yield from self._function_calls_context_messages
+            # In case on long-running function call, the function may already be added to the context
+            yield from self._context.get_messages()
+
+        for message in iter_all():
             if (
                 not isinstance(message, LLMSpecificMessage)
                 and message["role"] == "tool"


src/pipecat/processors/aggregators/llm_response_universal.py
```diff
@@ -757,8 +767,10 @@ async def _handle_function_call_result(self, frame: FunctionCallResultFrame):
                 run_llm = not bool(self._function_calls_in_progress)
 
         if run_llm:
+            self._flush_function_call_messages_to_context()
             await self.push_context_frame(FrameDirection.UPSTREAM)
 
+        # TODO: need to check this
```
filipi87: TODO: need to handle this.

src/pipecat/processors/aggregators/llm_response_universal.py
```diff
@@ -807,9 +819,11 @@ async def _handle_user_image_frame(self, frame: UserImageRawFrame):
         await self.push_context_frame(FrameDirection.UPSTREAM)
 
     async def _handle_llm_start(self, _: LLMFullResponseStartFrame):
+        logger.info(f"{self} LLMFullResponseStartFrame")
```
filipi87: TODO: remove the logs.

src/pipecat/processors/aggregators/llm_response_universal.py

tonilopezmr:
what was the problem here before?

filipi87:
This PR fixes https://github.com/pipecat-ai/pipecat/issues/3023

src/pipecat/processors/aggregators/llm_response_universal.py
kompfner, line 759: So if we're deferring adding the "IN_PROGRESS" function-call-placeholder message until _after_ any ongoing LLM response is finished...how does this part work if there's an ongoing LLM response? Won't it be a no-op because it can't find the placeholder message to update? 🤔

kompfner: Ah nvm I see that you've updated `_update_function_call_result` to update the "pending" function call context messages.

src/pipecat/processors/aggregators/llm_response_universal.py
```diff
@@ -781,7 +805,7 @@ async def _handle_function_call_cancel(self, frame: FunctionCallCancelFrame):
             del self._function_calls_in_progress[frame.tool_call_id]
 
     def _update_function_call_result(self, function_name: str, tool_call_id: str, result: Any):
-        for message in self._context.get_messages():
+        for message in self._function_calls_context_messages:
```
kompfner: What if the placeholder ("IN_PROGRESS") message has already made it into the context object? Don't we also need to look for it there? I'm thinking about the case of relatively long-running functions, where we might have other user-bot interactions in our context history after the function call placeholder, before it's resolved.

src/pipecat/processors/aggregators/llm_response_universal.py
```diff
@@ -756,17 +784,13 @@ async def _handle_function_call_result(self, frame: FunctionCallResultFrame):
...
-        # sure we don't block the pipeline.
-        if properties and properties.on_context_updated:
-            task_name = f"{frame.function_name}:{frame.tool_call_id}:on_context_updated"
-            task = self.create_task(properties.on_context_updated(), task_name)
-            self._context_updated_tasks.add(task)
-            task.add_done_callback(self._context_updated_task_finished)
+        # Only run if the LLM response has completed (not currently generating),
+        # otherwise defer execution until push_aggregation() is called
+        # (triggered by LLMFullResponseEndFrame or interruption).
+        if not self._started:
```
kompfner, line 790: Seems like relying on `self._started` _still_ leaves a gap (albeit a much smaller one than before) where the function call result could erroneously be inserted in the context in front of any ongoing LLM response, since it can take a little while for the `LLMFullResponseStartFrame` to arrive at the assistant aggregator (if, for example, the TTS service was busy with something).

src/pipecat/processors/aggregators/llm_response_universal.py
filipi87: Yep. 👍

src/pipecat/processors/aggregators/llm_response_universal.py
filipi87: Oo, I see, I believe this should be possible, yeah. I will make sure to handle that. Nice catch.

src/pipecat/processors/aggregators/llm_response_universal.py
filipi87: I don't think the TTS would ever be busy in this case, right ? Since the `LLMFullResponseStartFrame` only happens after some kind of interruption, the entire TTS queue would be discarded anyway I guess.

kompfner: Yeah...probably :)

src/pipecat/processors/aggregators/llm_response_universal.py

src/pipecat/processors/aggregators/llm_response_universal.py
```diff
@@ -650,21 +652,23 @@ async def process_frame(self, frame: Frame, direction: FrameDirection):
...
+            # Push timestamp frame with current time
+            timestamp_frame = LLMContextAssistantTimestampFrame(timestamp=time_now_iso8601())
+            await self.push_frame(timestamp_frame)
 
-        # Push timestamp frame with current time
-        timestamp_frame = LLMContextAssistantTimestampFrame(timestamp=time_now_iso8601())
-        await self.push_frame(timestamp_frame)
+        if self._function_calls_context_messages:
+            self._flush_function_call_messages_to_context()
+            await self.push_context_frame(FrameDirection.UPSTREAM)
```
aconchillo, line 671: Shouldn't this be added before the aggregation? That is, this block should go before the `if self._aggregation`?

filipi87: When they both happened at the same time, the LLM text frames were usually produced even before the function call, and were probably even handled by the TTS. The issue was that, because `LLMTextFrame` isn’t a system frame, it arrives here later. 

So, that is why we are adding the function calls after the aggregation.

But the truth is, unless we also make function calls into `ControlFrames`, we’ll never know the exact position here for sure, whether it should come before or after. 

In my tests, keeping it like this worked fine.

PR #3057, opened 2025-11-14
Improve TTSService handling of long LLM token outputs

markbackman:

The motivation for this change was discovering that Google Gemini outputs long chunks, sometimes containing multiple sentences. The `SimpleTextAggregator` extracts only the first sentence and buffers the remainder of the text. When the `LLMFullResponseEndFrame` is received, all remaining buffered text was being pushed to TTS as one large chunk. This results in potentially many sentences being sent to TTS at once.

The issue is that interruptions capture the last complete sentence; for long outputs with many buffered sentences, it was possible to have many sentences missed during interruption, which could cause the bot to repeat itself or continue speaking after being interrupted.

This PR adds a `flush_next_sentence()` method to the `SimpleTextAggregator` which is used when the `LLMFullResponseEndFrame` is received. Instead of sending all remaining text as one chunk, the buffered sentences are now flushed individually, providing better interruption points throughout the response.

**DON'T MERGE YET. WAIT UNTIL https://github.com/pipecat-ai/pipecat/pull/2899 LANDS.**

### DIFF
--- a/src/pipecat/utils/text/simple_text_aggregator.py
+++ b/src/pipecat/utils/text/simple_text_aggregator.py
@@ -41,30 +41,57 @@ class SimpleTextAggregator(BaseTextAggregator):
         """
         return self._text
 
-    async def aggregate(self, text: str) -> Optional[str]:
-        """Aggregate text and return completed sentences.
+    def _extract_next_sentence(self) -> Optional[str]:
+        """Extract the next complete sentence from the buffer.
 
-        Adds the new text to the buffer and checks for end-of-sentence markers.
-        When a sentence boundary is found, returns the completed sentence and
-        removes it from the buffer.
+        Returns:
+            The first complete sentence if a sentence boundary is found,
+            or None if the buffer is empty or contains only incomplete text.
+        """
+        eos_end_marker = match_endofsentence(self._text)
+        if eos_end_marker:
+            # Extract the first complete sentence
+            sentence = self._text[:eos_end_marker]
+            # Remove it from buffer
+            self._text = self._text[eos_end_marker:]
+            return sentence
+
+        return None
+
+    async def aggregate(self, text: str) -> Optional[str]:
+        """Aggregate text and return the first completed sentence.
+
+        Adds the new text to the buffer and checks for sentence boundaries.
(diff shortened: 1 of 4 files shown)


markbackman:
@mattieruth including you since you're working in this area of code. I don't think this impacts your PR but sharing in case it does.

src/pipecat/utils/text/simple_text_aggregator.py
```diff
@@ -41,30 +41,57 @@ def text(self) -> str:
...
+        if eos_end_marker:
+            # Extract the first complete sentence
+            sentence = self._text[:eos_end_marker]
+            # Remove it from buffer
+            self._text = self._text[eos_end_marker:]
+            return sentence
+
+        return None
+
     async def aggregate(self, text: str) -> Optional[str]:
```
filipi87, line 61: I am wondering if this could be considered a breaking change for anyone using it. We’re keeping the API the same but changing the behavior, and now, to extract the full text, they would need to use it together with `flush_next_sentence`.

So I am not sure if it would be better if we introduce a new method instead. What do you think ?

markbackman: I don't think this changes any functionality. `aggregate()` works consistently between versions, I think.

Previously, it would:
- Return either `None` if no sentences found or the first sentence if an end of sentence is found. The same is true today. It also uses the same logic.

I extracted the logic to avoid duplication. Otherwise, it could have remaining as is without change.

filipi87: Yeah, I think you’re right. Somehow I got confused the first time I looked, but it makes sense now. 

And in both versions we would still need to get any remaining text using `self._text_aggregator.text` at the end. 

Cool, makes sense

markbackman: Exactly.

Because of Mattie's bot output work, I'm going to have to hold off on merging this. We can stitch in this change once she wraps her work up. Thanks for confirming the approach 🙏

markbackman:
Closing due to conflicts, in favor of https://github.com/pipecat-ai/pipecat/pull/3124.

PR #3061, opened 2025-11-15
D'oh! My TTS "inter-frame-spaces" logic was *way* overcomplicated (an…

kompfner:
…d fundamentally mistaken, though it happened to work)

Now:
- For TTS word-by-word output and `TTSSpeakFrames`: `TTSTextFrame`s' have `includes_inter_frame_spaces=False`.
- For all other TTS output: `TTSTextFrame` pass through the received text frames' `includes_inter_frame_spaces` value. So far, this value has always been `True`: LLMs send text chunks already containing all necessary spaces.
- `LLMTextFrame`s set `includes_inter_frame_spaces=False` at init time, per the aforementioned assumption.


kompfner:
This undoes most of https://github.com/pipecat-ai/pipecat/pull/3041 and takes a different, much simpler, approach instead.

kompfner:
This mostly undoes https://github.com/pipecat-ai/pipecat/pull/3041

src/pipecat/frames/frames.py
```diff
@@ -352,7 +352,10 @@ def __str__(self):
 class LLMTextFrame(TextFrame):
     """Text frame generated by LLM services."""
 
-    pass
+    def __post_init__(self):
+        super().__post_init__()
+        # LLM services send text frames with all necessary spaces included
```
markbackman, line 357: This should be universally true. Good idea.

kompfner: Yeah, I was hesitant to do this before _juuuust_ in case anyone was manually creating `LLMTextFrame`s and relying on the old assistant aggregator behavior of inserting spaces. But...that feels like sticking to the wrong pattern for what is hopefully an extreme edge case.

PR #3064, opened 2025-11-16
pass through session request data correctly

jptaylor:

### DIFF
--- a/src/pipecat/runner/run.py
+++ b/src/pipecat/runner/run.py
@@ -322,7 +322,7 @@ def _setup_webrtc_routes(
                         type=request_data["type"],
                         pc_id=request_data.get("pc_id"),
                         restart_pc=request_data.get("restart_pc"),
-                        request_data=request_data,
+                        request_data=active_session,
                     )
                     return await offer(webrtc_request, background_tasks)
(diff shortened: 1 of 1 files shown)


src/pipecat/runner/run.py
```diff
@@ -322,7 +322,7 @@ async def proxy_request(
                         type=request_data["type"],
                         pc_id=request_data.get("pc_id"),
                         restart_pc=request_data.get("restart_pc"),
-                        request_data=request_data,
+                        request_data=active_session,
```
filipi87, line 325: The requests from start and connect are completely independent. That is also true within Pipecat Cloud, so we should not pass the stored session in this case. 

But you’re right, there is still an issue. To actually match the behavior of our client SDK, it should be:
`requestData.requestData`

As you can see [here in the web transport](https://github.com/pipecat-ai/pipecat-client-web-transports/blob/ad419d77b1221528ca1ea8ccaa973c0a8176f81a/transports/small-webrtc-transport/src/smallWebRTCTransport.ts#L721-L723). This is the field where the user can send any custom data when invoking connect.

cc: @mattieruth

mattieruth: ah. truth. `requestData = requestData.requestData` should do it.

PR #3067, opened 2025-11-16
feat: add interruption strategy support to DeepgramFlux STT

eroub:
## Problem

The `DeepgramFluxSTTService` was immediately calling `push_interruption_task_frame_and_wait()` on every `StartOfTurn` event from Deepgram, bypassing the framework's interruption strategy mechanism, `MinWordsInterruptionStrategy`. This prevented users from configuring when bot interruptions should occur based on the amount of user speech.

## Solution

This PR aligns `DeepgramFluxSTTService` with other STT services in pipecat by:

1. **Tracking bot speaking state** - The service now processes `BotStartedSpeakingFrame` and `BotStoppedSpeakingFrame` to maintain awareness of when the bot is speaking
2. **Conditional interruption** - Interruptions are only pushed immediately when:
   - No interruption strategies are configured, OR
   - The bot is not currently speaking
3. **Deferring to aggregators** - When interruption strategies are configured AND the bot is speaking, the service defers the interruption decision to the aggregator (which evaluates strategies like `MinWordsInterruptionStrategy`)

## Benefits

- Users can now configure `MinWordsInterruptionStrategy` or other interruption strategies with Deepgram Flux STT
- Behavior is consistent with other STT services in the framework
- Maintains backward compatibility (immediate interruption when no strategies configured)

## Testing

Tested locally using `MinWordsInterruptionStrategy(min_words=3)` to verify that interruptions only occur after the configured word threshold is met.

### DIFF
--- a/src/pipecat/services/deepgram/flux/stt.py
+++ b/src/pipecat/services/deepgram/flux/stt.py
@@ -15,6 +15,8 @@ from loguru import logger
 from pydantic import BaseModel
 
 from pipecat.frames.frames import (
+    BotStartedSpeakingFrame,
+    BotStoppedSpeakingFrame,
     CancelFrame,
     EndFrame,
     ErrorFrame,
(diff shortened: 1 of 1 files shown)


markbackman:
Thanks for submitting this, but we're working on a refactor of interruptions here:
https://github.com/pipecat-ai/pipecat/pull/3045

I think we should probably close out this PR as Deepgram Flux will be supported after this refactoring.

PR #3073, opened 2025-11-17
BaseTextFilter: only require subclasses to implement filter()

aconchillo:

This is to simplify subclasses. Sometimes they don't need to handle any settings update or interruptions, all they are is a simple one time filter.


markbackman (approved):
Sounds like a positive change.

PR #3074, opened 2025-11-17
Update Moondream example so that Moondream service output makes it in…

kompfner:
…to the context, even if the TTS service is disabled


examples/foundational/14d-function-calling-moondream-video.py
```diff
@@ -130,14 +158,20 @@ async def run_bot(transport: BaseTransport, runner_args: RunnerArguments):
     # If you run into weird description, try with use_cpu=True
     moondream = MoondreamService()
 
+    # Wrap TextFrames with LLM response start/end frames, which makes Moondream
+    # output be treated like LLM responses for the purpose of context
+    # aggregation. Without this, the assistant context aggregator would ignore
+    # Moondream output (if the TTS service is disabled).
```
kompfner, line 164: TBH I'm not entirely sure why it was working with the TTS service enabled. It looks like for some reason the `LLMFullResponseEndFrame` never gets sent (or never makes it to the assistant context aggregator), so the Moondream output is still written to the context, as the aggregator thinks the LLM response is still ongoing...

aconchillo (approved):

PR #3078, opened 2025-11-18
LLMContext: create_image_message/create_audio_message are now async

aconchillo:

Unfortunately, this is a breaking change, but it is necessary to not block the asyncio event loop. Very easy to hit audio choppiness when creating image messages.

### DIFF
--- a/src/pipecat/processors/aggregators/llm_context.py
+++ b/src/pipecat/processors/aggregators/llm_context.py
@@ -14,6 +14,7 @@ translation from this universal context into whatever format it needs, using a
 service-specific adapter.
 """
 
+import asyncio
 import base64
 import io
 import wave
@@ -137,7 +138,7 @@ class LLMContext:
         return {"role": role, "content": content}
 
     @staticmethod
-    def create_image_message(
+    async def create_image_message(
         *,
         role: str = "user",
         format: str,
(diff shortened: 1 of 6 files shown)


src/pipecat/processors/aggregators/llm_context.py
```diff
@@ -154,15 +155,21 @@ def create_image_message(
...
+            buffer = io.BytesIO()
+            Image.frombytes(format, size, image).save(buffer, format="JPEG")
+            encoded_image = base64.b64encode(buffer.getvalue()).decode("utf-8")
+            return encoded_image
+
+        encoded_image = await asyncio.to_thread(encode_image)
+
         url = f"data:image/jpeg;base64,{encoded_image}"
 
         return LLMContext.create_image_url_message(role=role, url=url, text=text)
```
aconchillo, line 169: `create_image_url_message()` is not async because it doesn't do any heavy operation.

aconchillo:
> Makes sense. It's a bummer that it's breaking, but I doubt there's any way around it given the symptom.
> 
> Is it worth mentioning the breaking change in the changelog?

I'll add a warning sign like we've done in the past and mention it's a breaking change explicitly.

aconchillo:
> Makes sense. It's a bummer that it's breaking, but I doubt there's any way around it given the symptom.
> 

People could get around the issue by calling `asyncio.thread()` themselves, but it's very annoying having to do that every time.

PR #3079, opened 2025-11-18
set full model name for base openai models

vipyne:


markbackman (approved):
LGTM.

Was @aconchillo's comment about the use of private variables? If so, that's already a known issue that we need to improve. I think this is fine for now.

PR #3080, opened 2025-11-18
`LLMAssistantAggregator` now properly aggregates text that might be a…

kompfner:
… mix of `includes_inter_frame_spaces=True` and `includes_inter_frame_spaces=False` frames

### DIFF
--- a/src/pipecat/utils/string.py
+++ b/src/pipecat/utils/string.py
@@ -18,6 +18,7 @@ Dependencies:
 """
 
 import re
+from dataclasses import dataclass
 from typing import FrozenSet, List, Optional, Sequence, Tuple
 
 import nltk
@@ -198,7 +199,24 @@ def parse_start_end_tags(
     return (None, current_tag_index)
 
 
-def concatenate_aggregated_text(text_parts: List[str], add_spaces: bool) -> str:
+@dataclass
+class TextPartForConcatenation:
+    """Class representing a part of text for concatenation with concatenate_aggregated_text.
+
+    Attributes:
+        text: The text content.
+        includes_inter_part_spaces: Whether any necessary inter-frame
+            (leading/trailing) spaces are already included in the text.
+    """
+
+    text: str
+    includes_inter_part_spaces: bool
+
+    def __str__(self):
+        return f"{self.name}(text: [{self.text}], includes_inter_part_spaces: {self.includes_inter_part_spaces})"
+
+
+def concatenate_aggregated_text(text_parts: List[TextPartForConcatenation]) -> str:
     """Concatenate a list of text parts into a single string.
 
     This function joins the provided list of text parts into a single string,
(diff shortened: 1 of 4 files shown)


src/pipecat/utils/string.py
```diff
@@ -198,7 +199,24 @@ def parse_start_end_tags(
...
+    """Class representing a part of text for concatenation with concatenate_aggregated_text.
+
+    Attributes:
+        text: The text content.
+        includes_inter_part_spaces: Whether any necessary inter-frame
+            (leading/trailing) spaces are already included in the text.
+    """
+
+    text: str
+    includes_inter_part_spaces: bool
```
mattieruth, line 213: i'll use this same name for my stuff

src/pipecat/utils/string.py
```diff
@@ -208,15 +226,57 @@ def concatenate_aggregated_text(text_parts: List[str], add_spaces: bool) -> str:
...
-        add_spaces: Whether to add spaces between text parts during concatenation.
+        text_parts: A list of text parts to concatenate.
 
     Returns:
         A single concatenated string.
     """
-    # Concatenate text parts with or without spaces based on the flag
-    separator = " " if add_spaces else ""
-    result = separator.join(text_parts)
+    print("Concatenating text parts:", text_parts)
```
mattieruth: don't forget to remove this

src/pipecat/utils/string.py
```diff
@@ -208,15 +226,57 @@ def concatenate_aggregated_text(text_parts: List[str], add_spaces: bool) -> str:
...
+            result += " "
+            append_part(part)
+        else:
+            # This part represents a transition to a new run (spaces -> no spaces, or vice versa).
+            # Add a space if needed, before appending.
+            if not result[-1].isspace() and not part.text[0].isspace():
+                result += " "
+            append_part(part)
+
+    # NOTE: the above logic assumes that runs of text parts with
```
kompfner, line 274: @mattieruth I'm not entirely sure this assumption will always hold...I'm worried about multiple custom aggregations from the `LLMTextProcessor` that each have `includes_inter_part_spaces=True` but each begin and end without a space...

To address that, we might just need to update this method to somehow be aware of text aggregation boundaries.

Though...I'm hard-pressed to think of a custom text aggregator in `LLMTextProcessor` labeling its output as `includes_inter_part_spaces=True` unless it really was just passing through LLM-produced text, in which case I _think_ this assumption holds...maybe?

src/pipecat/utils/string.py
```diff
@@ -208,15 +226,57 @@ def concatenate_aggregated_text(text_parts: List[str], add_spaces: bool) -> str:
...
+            append_part(part)
+        elif not part.includes_inter_part_spaces and not last_includes_inter_part_spaces:
+            # This part is part of an ongoing run that has no spaces included.
+            # Add a space before appending.
+            result += " "
+            append_part(part)
+        else:
+            # This part represents a transition to a new run (spaces -> no spaces, or vice versa).
+            # Add a space if needed, before appending.
+            if not result[-1].isspace() and not part.text[0].isspace():
```
mattieruth, line 270: what about new lines or tabs or other "spacing" i'm forgetting. do we need to consider those?

src/pipecat/utils/string.py
mattieruth: makes sense. i'm good with this. I still stand by that extra spaces is the best assumption so if something came along that did this, I would encourage marking it as `includes_inter_part_spaces=False`

src/pipecat/utils/string.py
kompfner: `isspace()` will return `True` for _any_ whitespace character, which is what we want here: if there's no separation between "runs", add a space (" ")

src/pipecat/utils/string.py
kompfner: thanks! will do

tests/test_context_aggregators.py
```diff
@@ -1005,3 +1005,54 @@ def create_assistant_aggregator_params(
...
+            make_text_frame("helps!", includes_spaces=True),
+            LLMFullResponseEndFrame(),
+        ]
+        expected_down_frames = [*self.EXPECTED_CONTEXT_FRAMES]
+        await run_test(
+            aggregator,
+            frames_to_send=frames_to_send,
+            expected_down_frames=expected_down_frames,
+        )
+        print(context.messages)
```
kompfner: Note to self: remove this

Issue #1781, opened 2025-05-08
OAI Realtime API - All errors are treated as fatal


cvanek706:
### pipecat version

0.0.63

### Python version

3.11.12

### Operating System

Ubuntu 24.04

### Issue description

In pipecat.services.openai_realtime_beta.openai, in the OpenAIRealtimeBetaLLMService object, it treats all error events from the Realtime API as fatal. According to The [API Reference](https://platform.openai.com/docs/api-reference/realtime-server-events/error), "Most errors are recoverable and the session will stay open, we recommend to implementors to monitor and log error messages by default.". This means that any error will kill the bot, even if it is recoverable. Is it possible to change it so that recoverable errors are handled in a way that doesn't unnecessarily kill the bot?

### Reproduction steps

1. Run a Pipecat Bot using the OpenAI Realtime API
2. Attempt to generate a response while the API is already in the middle of one, using this command or something similar: `await llm.send_client_event(events.ResponseCreateEvent(response=events.ResponseProperties("repond to the user")))`

### Expected behavior

Pipecat registers that an error event occured from the realtime API, but the conversation continues anyway.

### Actual behavior

The bot stops running because all errors from the realtime API are treated as fatal errors.

### Logs

```shell
rq-worker      | 2025-05-08 21:44:23.016 | ERROR    | pipecat.pipeline.task:_process_up_queue:473 - A fatal error occurred: ErrorFrame#0(error: Error: event_id='event_BV3KR9aG1ijTmCqWBhkYu' type='error' error=RealtimeError(type='invalid_request_error', code='invalid_value', message='Audio content of 12000ms is already shorter than 15276ms', param=None, event_id=None), fatal: True)
rq-worker      | 2025-05-08 21:44:23.088 | INFO     | pipecat.transports.services.daily:leave:540 - Leaving https://georgebot.daily.co/goody_b60225b5-d73a-4782-808d-b681a309cf5b
rq-worker      | 21:44:23:   0.28s → 6348ce5f-4258-40f2-a493-fd58e3cd165c:store_conversation_message(MessageRecord(client_id='goody', cid=UUID('b60225b5-d73a-4782-808d-b681a309cf5b…)
rq-worker      | 2025-05-08 21:44:23.168 | DEBUG    | pipecat.transports.services.daily:on_transcription_stopped:748 - Transcription stopped
[...]

PR #3084, opened 2025-11-18
Improving error handler.

filipi87:
Improving error handling:
- Added `on_error` to the `FrameProcessor`.
- Renamed `push_error(error: ErrorFrame)` to `push_error_frame(error: ErrorFrame)`.
  - Invokes `on_error`.
  - Pushes the error frame upstream.
- Created a new `push_error(error_msg: Optional[str] = None, exception: Optional[Exception] = None, fatal: bool = False)`:
  - Logs the error message.
  - Creates the `ErrorFrame`.
  - Invokes `push_error_frame`, which triggers `on_error` and pushes the error upstream.
- Replaced all `logger.exception` to `logger.error` 

Fixes https://github.com/pipecat-ai/pipecat/issues/1781

### DIFF
--- a/src/pipecat/frames/frames.py
+++ b/src/pipecat/frames/frames.py
@@ -835,11 +835,13 @@ class ErrorFrame(SystemFrame):
         error: Description of the error that occurred.
         fatal: Whether the error is fatal and requires bot shutdown.
         processor: The frame processor that generated the error.
+        exception: The exception that occurred.
     """
 
     error: str
     fatal: bool = False
     processor: Optional["FrameProcessor"] = None
+    exception: Optional[Exception] = None
 
     def __str__(self):
         return f"{self.name}(error: {self.error}, fatal: {self.fatal})"
--- a/src/pipecat/processors/frame_processor.py
+++ b/src/pipecat/processors/frame_processor.py
@@ -142,6 +142,7 @@ class FrameProcessor(BaseObject):
     - on_after_process_frame: Called after a frame is processed
     - on_before_push_frame: Called before a frame is pushed
     - on_after_push_frame: Called after a frame is pushed
+    - on_error: Called when an error is raised in the frame processing.
     """
 
     def __init__(
@@ -234,6 +235,7 @@ class FrameProcessor(BaseObject):
         self._register_event_handler("on_after_process_frame", sync=True)
         self._register_event_handler("on_before_push_frame", sync=True)
         self._register_event_handler("on_after_push_frame", sync=True)
+        self._register_event_handler("on_error", sync=True)
 
     @property
     def id(self) -> int:
@@ -630,7 +632,43 @@ class FrameProcessor(BaseObject):
         elif isinstance(frame, (FrameProcessorResumeFrame, FrameProcessorResumeUrgentFrame)):
             await self.__resume(frame)
 
-    async def push_error(self, error: ErrorFrame):
+    async def push_error(
+        self,
+        error_msg: str,
+        exception: Optional[Exception] = None,
+        fatal: bool = False,
+    ):
+        """Creates and pushes an ErrorFrame upstream.
+
+        Creates and pushes an ErrorFrame upstream to notify other processors in the
+        pipeline about an error condition. The error frame will include context about
+        which processor generated the error.
+
+        Args:
+            error_msg: Descriptive message explaining the error condition.
+            exception: Optional exception object that caused the error, if available.
+                This provides additional context for debugging and error handling.
+            fatal: Whether this error should be considered fatal to the pipeline.
+                Fatal errors typically cause the entire pipeline to stop processing.
+                Defaults to False for non-fatal errors.
+
+        Example::
+
+            ```python
+            # Non-fatal error
+            await self.push_error("Failed to process audio chunk, skipping")
+
+            # Fatal error with exception context
+            try:
+                result = some_critical_operation()
+            except Exception as e:
+                await self.push_error("Critical operation failed", exception=e, fatal=True)
+            ```
+        """
+        error_frame = ErrorFrame(error=error_msg, fatal=fatal, exception=exception, processor=self)
+        await self.push_error_frame(error=error_frame)
+
+    async def push_error_frame(self, error: ErrorFrame):
         """Push an error frame upstream.
 
         Args:
@@ -638,6 +676,8 @@ class FrameProcessor(BaseObject):
         """
         if not error.processor:
             error.processor = self
+        await self._call_event_handler("on_error", error)
+        logger.error(f"{error.processor} error: {error.error}")
         await self.push_frame(error, FrameDirection.UPSTREAM)
 
     async def push_frame(self, frame: Frame, direction: FrameDirection = FrameDirection.DOWNSTREAM):
@@ -759,8 +799,10 @@ class FrameProcessor(BaseObject):
                 await self.__cancel_process_task()
                 self.__create_process_task()
         except Exception as e:
-            logger.exception(f"Uncaught exception in {self} when handling _start_interruption: {e}")
-            await self.push_error(ErrorFrame(str(e)))
+            await self.push_error(
+                error_msg=f"Uncaught exception handling _start_interruption: {e}",
+                exception=e,
+            )
 
     async def __internal_push_frame(self, frame: Frame, direction: FrameDirection):
         """Internal method to push frames to adjacent processors.
@@ -797,8 +839,7 @@ class FrameProcessor(BaseObject):
                     await self._observer.on_push_frame(data)
                 await self._prev.queue_frame(frame, direction)
         except Exception as e:
-            logger.exception(f"Uncaught exception in {self}: {e}")
-            await self.push_error(ErrorFrame(str(e)))
+            await self.push_error(error_msg=f"Uncaught exception: {e}", exception=e)
 
     def _check_started(self, frame: Frame):
         """Check if the processor has been started.
@@ -874,8 +915,7 @@ class FrameProcessor(BaseObject):
 
             await self._call_event_handler("on_after_process_frame", frame)
         except Exception as e:
-            logger.exception(f"{self}: error processing frame: {e}")
-            await self.push_error(ErrorFrame(str(e)))
+            await self.push_error(error_msg=f"Error processing frame: {e}", exception=e)
 
     async def __input_frame_task_handler(self):
         """Handle frames from the input queue.
--- a/src/pipecat/serializers/plivo.py
+++ b/src/pipecat/serializers/plivo.py
@@ -199,7 +199,7 @@ class PlivoFrameSerializer(FrameSerializer):
                         )
 
         except Exception as e:
-            logger.exception(f"Failed to hang up Plivo call: {e}")
+            logger.error(f"Failed to hang up Plivo call: {e}")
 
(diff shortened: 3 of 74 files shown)


src/pipecat/frames/frames.py
```diff
@@ -812,6 +812,7 @@ class ErrorFrame(SystemFrame):
     error: str
     fatal: bool = False
     processor: Optional["FrameProcessor"] = None
+    exception: Optional[Exception] = None
```
markbackman, line 844: Add this new parameter to the docstring.

src/pipecat/processors/frame_processor.py
```diff
@@ -630,14 +632,54 @@ async def process_frame(self, frame: Frame, direction: FrameDirection):
...
+
+        Args:
+            error_msg: Optional descriptive message explaining the error condition.
+            exception: Optional exception object that caused the error, if available.
+                This provides additional context for debugging and error handling.
+            fatal: Whether this error should be considered fatal to the pipeline.
+                Fatal errors typically cause the entire pipeline to stop processing.
+                Defaults to False for non-fatal errors.
+
+        Example:
```
markbackman: Nit: for docs generation (two colons plus a newline).

```suggestion
        Example::
        
```

src/pipecat/processors/frame_processor.py
```diff
@@ -797,8 +841,7 @@ async def __internal_push_frame(self, frame: Frame, direction: FrameDirection):
                     await self._observer.on_push_frame(data)
                 await self._prev.queue_frame(frame, direction)
         except Exception as e:
-            logger.exception(f"Uncaught exception in {self}: {e}")
-            await self.push_error(ErrorFrame(str(e)))
+            await self.push_error(exception=e)
```
markbackman: More of a curiosity, but why no error_msg here? I'm trying to figure out when we should push_error with an exception vs exception + error_msg.

src/pipecat/serializers/plivo.py
```diff
@@ -199,7 +199,7 @@ async def _hang_up_call(self):
                         )
 
         except Exception as e:
-            logger.exception(f"Failed to hang up Plivo call: {e}")
+            logger.error(f"Failed to hang up Plivo call: {e}")
```
markbackman, line 202: When should we be using push_error vs just calling logger.error? What is the rule we should apply?

src/pipecat/services/google/llm.py
```diff
@@ -793,7 +793,7 @@ def _maybe_unset_thinking_budget(self, generation_params: Dict[str, Any]):
                 return
             generation_params.setdefault("thinking_config", {})["thinking_budget"] = 0
         except Exception as e:
-            logger.exception(f"Failed to unset thinking budget: {e}")
+            logger.error(f"Failed to unset thinking budget: {e}")
```
markbackman, line 796: Why in this situation do we not call push_error?

src/pipecat/services/google/llm.py
```diff
@@ -983,7 +983,7 @@ async def _process_context(self, context: OpenAILLMContext | LLMContext):
         except DeadlineExceeded:
             await self._call_event_handler("on_completion_timeout")
         except Exception as e:
-            logger.exception(f"{self} exception: {e}")
+            logger.error(f"{self} exception: {e}")
```
markbackman: Same here.

src/pipecat/services/stt_service.py
```diff
@@ -329,4 +329,4 @@ def __init__(self, *, reconnect_on_error: bool = True, **kwargs):
 
     async def _report_error(self, error: ErrorFrame):
         await self._call_event_handler("on_connection_error", error.error)
-        await self.push_error(error)
+        await self.push_error_frame(error)
```
markbackman, line 332: It looks like _report_error is used for connection level errors, triggering the on_connection_error event and push_error_frame.

Maybe this should be in in the base AIService class and used consistently across services?

src/pipecat/services/ai_service.py
```diff
@@ -166,6 +166,6 @@ async def process_generator(self, generator: AsyncGenerator[Frame | None, None])
         async for f in generator:
             if f:
                 if isinstance(f, ErrorFrame):
-                    await self.push_error(f)
+                    await self.push_error_frame(f)
```
markbackman, line 169: Why push_error_frame here? If the ErrorFrame is already pushed on the Pipeline, won't this result in other frame processors in the Pipeline pushing the error again?

src/pipecat/processors/frame_processor.py
filipi87: I tried to respect the messages that we already had in the code. So, in places where we didn’t have any specific message, I basically refactored it to:
- `self.push_error(exception=e)`

But where we were already creating an error message that had some specific meaning (rather than something generic, or in cases where we are receiving these errors as event from the service API), it’s still possible to send your custom error message using:
- `await self.push_error(error_msg=your_error_msg)`

src/pipecat/serializers/plivo.py
filipi87: Right now, I am invoking `push_error` inside everything that extends `FrameProcessor`.

In the other classes, for example, this Plivo one, which does not extend `FrameProcessor`, I have kept the current logging, but replaced `logger.exception` with `logger.error`.

Maybe as a future improvement, these classes could throw their errors to be handled by some FrameProcessor, but I didn't want to rock the boat too much on this PR.

src/pipecat/services/google/llm.py
filipi87: I guess here I have probably missed this one.

src/pipecat/services/google/llm.py
filipi87: I guess here I have probably missed this one.

filipi87:
> I could use some clarification as to when to use the push_error_frame vs push_error vs just calling logger.error.

Hi @markbackman , here is the logic I applied:

- `logger.error`: This should be used in classes that do not extend `FrameProcessor`. 
  - As a possible future improvement, these classes could throw their errors to be handled by a FrameProcessor, but I didn't want to introduce major structural changes in this PR.
- `push_error_frame`: I kept these methods because we have a couple of places where we create frames inside an async generator. In those cases, we `yield` the `ErrorFrame` there.
- `push_error`: This is the default. Every class that extends `FrameProcessor` and is not yielding frames inside an async generator should use this one.

src/pipecat/services/ai_service.py
filipi87: This is only used inside the async generators, where we are yielding the frames (including the `ErrorFrame`) instead of already pushing them to the pipeline. 
So we’re essentially keeping the behavior we had before, and it is only pushed to the pipeline here.

src/pipecat/serializers/plivo.py
markbackman: Ahh, that's right! FrameSerializers aren't FrameProcessors! Makes sense. So, call push_error in a FrameProcessor. Otherwise, just logger.error. That seems logical. Thank you.

src/pipecat/services/deepgram/tts.py
```diff
@@ -227,5 +227,5 @@ async def run_tts(self, text: str) -> AsyncGenerator[Frame, None]:
             yield TTSStoppedFrame()
 
         except Exception as e:
-            logger.exception(f"{self} exception: {e}")
+            logger.error(f"{self} exception: {e}")
```
markbackman: Should be push_error?

markbackman:
@filipi87 my review feedback is in the branch:
https://github.com/pipecat-ai/pipecat/pull/3142

Just a few additional things to change.

src/pipecat/processors/frame_processor.py
aconchillo: I think in this case, the error message would be "Uncaught exception"

src/pipecat/processors/frame_processor.py
```diff
@@ -630,14 +632,54 @@ async def process_frame(self, frame: Frame, direction: FrameDirection):
         elif isinstance(frame, (FrameProcessorResumeFrame, FrameProcessorResumeUrgentFrame)):
             await self.__resume(frame)
 
-    async def push_error(self, error: ErrorFrame):
+    async def push_error(
+        self,
+        error_msg: Optional[str] = None,
```
aconchillo: I think this should be mandatory.

src/pipecat/processors/frame_processor.py
```diff
@@ -630,14 +632,54 @@ async def process_frame(self, frame: Frame, direction: FrameDirection):
...
+            await self.push_error("Failed to process audio chunk, skipping")
+
+            # Fatal error with exception context
+            try:
+                result = some_critical_operation()
+            except Exception as e:
+                await self.push_error("Critical operation failed", exception=e, fatal=True)
+            ```
+        """
+        error_message = error_msg or f"{self} exception: {exception}"
```
aconchillo: If we make it mandatory, then we don't need this. Because, `exception` is also optional.

So, right now you could call `self.push_error()` which would produce something weird.

src/pipecat/processors/frame_processor.py
filipi87: Done.

src/pipecat/processors/frame_processor.py
filipi87: done.

src/pipecat/services/google/llm.py
filipi87: Done.

src/pipecat/services/google/llm.py
filipi87: Done.

src/pipecat/processors/frame_processor.py
```diff
@@ -234,6 +235,7 @@ def __init__(
         self._register_event_handler("on_after_process_frame", sync=True)
         self._register_event_handler("on_before_push_frame", sync=True)
         self._register_event_handler("on_after_push_frame", sync=True)
+        self._register_event_handler("on_error", sync=True)
```
aconchillo, line 238: Do we want this to be sync=True? This means we won't create a task for it. Since it's an error this might be fine, I haven't thought much about it.

src/pipecat/processors/frame_processor.py
```diff
@@ -630,14 +632,52 @@ async def process_frame(self, frame: Frame, direction: FrameDirection):
...
+    async def push_error_frame(self, error: ErrorFrame):
         """Push an error frame upstream.
 
         Args:
             error: The error frame to push.
         """
         if not error.processor:
             error.processor = self
+        await self._call_event_handler("on_error", error)
+        logger.error(error.error)
```
aconchillo: I would log the error first, because then we make sure we are calling `on_error` at that point.

aconchillo: Also, we probably want to do:

```
        logger.error(f"{self} error: {error.error}")
```

And remove {self} references form every error log. This way people don't have to thing about it. Adding`{self}` is very useful to know exactly the processor that is failing (you might have multiple of the same type).

filipi87: Done.

src/pipecat/processors/frame_processor.py
filipi87: Yeah, I thought `sync=True` would be the right move for now.

Issue #3063, opened 2025-11-15
Add RimeArcanaTTSService to Support Non-JSON WebSocket Streaming


gokuljs:
### Problem Statement

Pipecat’s current TTS services only support JSON-based WebSocket protocols. The Arcana (Rime) TTS model streams raw binary audio frames instead of JSON messages, so the existing TTS service infrastructure cannot process Arcana output. This prevents users from using Arcana for real-time speech generation within the Pipecat pipeline.

## **Proposed Solution**

Implement a new `RimeArcanaTTSService` that adapts Pipecat’s TTS interface to Arcana’s non-JSON WebSocket protocol.

The service should:

* Establish a WebSocket connection using Arcana’s query-parameter configuration.
* Send **plain text input** (no JSON envelopes).
* Receive **binary audio frames** and forward them directly to the audio pipeline.
* Support Arcana control commands: `<EOS>`, `<CLEAR>`, `<FLUSH>`.
* Implement the required lifecycle and utility methods:
  `init`, `start`, `stop`, `_connect`, `_disconnect`, `_receive_messages`, `_get_websocket`,
  `run_tts`, `_handle_interruption`, `flush_audio`, `language_to_service_language`,
  `can_generate_metrics`, `_update_settings`.
* Reconnect when settings change (since model and voice parameters affect the WebSocket URL).

This enables full Arcana compatibility, consistent service behavior, and real-time speech synthesis support across Pipecat.

### Would you be willing to help implement this feature?

- [x] Yes, I'd like to contribute
- [ ] No, I'm just suggesting

gokuljs:
@markbackman These are the methods I need for this new Rime service: init, start, stop, _connect, _disconnect, _receive_messages, _get_websocket, run_tts, _handle_interruption, flush_audio, language_to_service_language, can_generate_metrics, and _update_settings.

gokuljs:
#3085

PR #3085, opened 2025-11-18
Add RimeNonJsonTTSService for non-JSON WebSocket API support

gokuljs:


gokuljs:
#3063

gokuljs:
@markbackman, any update here?

src/pipecat/services/rime/tts.py
```diff
@@ -574,3 +579,337 @@ async def run_tts(self, text: str) -> AsyncGenerator[Frame, None]:
         finally:
             await self.stop_ttfb_metrics()
             yield TTSStoppedFrame()
+
+
+class RimeNonJsonTTSService(InterruptibleTTSService):
+    """Pipecat TTS service for Rime's non-JSON WebSocket API.
+
+    This service enables Text-to-Speech synthesis over WebSocket endpoints that require plain text (not JSON) messages and return raw audio bytes. It is designed for use with TTS models like Arcana, which currently do not support JSON-based WebSocket protocols (though this may change in the future).
```
markbackman: Nit: can you wrap lines in the docstrings?

src/pipecat/services/rime/tts.py
```diff
@@ -574,3 +579,337 @@ async def run_tts(self, text: str) -> AsyncGenerator[Frame, None]:
...
+
+    def __init__(
+        self,
+        *,
+        api_key: str,
+        voice_id: str,
+        url: str = "wss://users.rime.ai/ws",
+        model: str = "arcana",
+        audio_format: str = "pcm",
+        sample_rate: Optional[int] = 24000,
```
markbackman: This should be set to `None`, which allows the base TTSService to handle picking the sample rate either from this classes __init__ or the sample rate from the StartFrame.

```suggestion
        sample_rate: Optional[int] = None,
```

src/pipecat/services/rime/tts.py
```diff
@@ -574,3 +579,337 @@ async def run_tts(self, text: str) -> AsyncGenerator[Frame, None]:
         finally:
             await self.stop_ttfb_metrics()
             yield TTSStoppedFrame()
+
+
+class RimeNonJsonTTSService(InterruptibleTTSService):
```
markbackman: The `InterruptibleTTSService` base class is meant for websocket TTS services that don't have a message to stop TTS generation. Since you have a CLEAR message, I think you actually want the base class to be `WebsocketTTSService`. Your `_handle_interruptions()` method looks correct given this base class.

src/pipecat/services/rime/tts.py
```diff
@@ -574,3 +579,337 @@ async def run_tts(self, text: str) -> AsyncGenerator[Frame, None]:
...
+            self._websocket = None
+            await self._call_event_handler("on_disconnected")
+
+    def _get_websocket(self):
+        """Get active WebSocket connection or raise exception."""
+        if self._websocket:
+            return self._websocket
+        raise Exception("Websocket not connected")
+
+    async def _handle_interruption(self, frame: InterruptionFrame, direction: FrameDirection):
```
markbackman: This looks good.

src/pipecat/services/rime/tts.py
```diff
@@ -574,3 +579,337 @@ async def run_tts(self, text: str) -> AsyncGenerator[Frame, None]:
...
+                if not self._started:
+                    await self.start_ttfb_metrics()
+                    yield TTSStartedFrame()
+                    self._started = True
+                # Send bare text (not JSON)
+                await self._get_websocket().send(text)
+                await self.start_tts_usage_metrics(text)
+
+            except Exception as e:
+                logger.error(f"{self} exception: {e}")
```
markbackman: We've change error handling patterns. I'll make a few comments on it.

Remove this.
```suggestion
```

src/pipecat/services/rime/tts.py
```diff
@@ -574,3 +579,337 @@ async def run_tts(self, text: str) -> AsyncGenerator[Frame, None]:
...
+                yield ErrorFrame(error=f"{self} error: {e}")
```
markbackman:
```suggestion
                yield ErrorFrame(error=f"Unknown error occurred: {e}")
```

src/pipecat/services/rime/tts.py
```diff
@@ -574,3 +579,337 @@ async def run_tts(self, text: str) -> AsyncGenerator[Frame, None]:
...
+                yield TTSStoppedFrame()
+                await self._disconnect()
+                await self._connect()
+                return
+            yield None
+        except Exception as e:
+            logger.error(f"{self} exception: {e}")
```
markbackman:
```suggestion
```

src/pipecat/services/rime/tts.py
```diff
@@ -574,3 +579,337 @@ async def run_tts(self, text: str) -> AsyncGenerator[Frame, None]:
...
+            yield ErrorFrame(error=f"{self} error: {e}")
```
markbackman:
```suggestion
            yield ErrorFrame(error=f"Unknown error occurred: {e}")
```

src/pipecat/services/rime/tts.py
```diff
@@ -574,3 +579,337 @@ async def run_tts(self, text: str) -> AsyncGenerator[Frame, None]:
...
+            # Build URL with query parameters (only non-None values)
+            params = "&".join(f"{k}={v}" for k, v in self._settings.items() if v is not None)
+            url = f"{self._url}?{params}"
+            headers = {"Authorization": f"Bearer {self._api_key}"}
+            self._websocket = await websocket_connect(
+                url, additional_headers=headers, max_size=1024 * 1024 * 16
+            )
+            await self._call_event_handler("on_connected")
+        except Exception as e:
+            logger.error(f"{self} exception: {e}")
```
markbackman: push_error() calls logger.error() already.
```suggestion
```

src/pipecat/services/rime/tts.py
```diff
@@ -574,3 +579,337 @@ async def run_tts(self, text: str) -> AsyncGenerator[Frame, None]:
...
+            await self.push_error(ErrorFrame(error=f"{self} error: {e}"))
```
markbackman:
```suggestion
            await self.push_error(error_msg=f"Unknown error occurred: {e}", exception=e)
```

src/pipecat/services/rime/tts.py
```diff
@@ -574,3 +579,337 @@ async def run_tts(self, text: str) -> AsyncGenerator[Frame, None]:
...
+        """Close WebSocket connection and clean up state."""
+        try:
+            await self.stop_all_metrics()
+            if self._websocket:
+                # Send EOS command to gracefully close
+                await self._websocket.send("<EOS>")
+                await self._websocket.close()
+                logger.debug("Disconnected from Rime None json websocket")
+        except Exception as e:
+            logger.error(f"{self} exception: {e}")
```
markbackman:
```suggestion
```

src/pipecat/services/rime/tts.py
```diff
@@ -574,3 +579,337 @@ async def run_tts(self, text: str) -> AsyncGenerator[Frame, None]:
...
+            await self.push_error(ErrorFrame(error=f"{self} error: {e}"))
```
markbackman:
```suggestion
            await self.push_error(error_msg=f"Unknown error occurred: {e}", exception=e)
```

src/pipecat/services/rime/tts.py
```diff
@@ -574,3 +579,337 @@ async def run_tts(self, text: str) -> AsyncGenerator[Frame, None]:
...
+                    await self.stop_ttfb_metrics()
+
+                    frame = TTSAudioRawFrame(
+                        audio=message,
+                        sample_rate=self.sample_rate,
+                        num_channels=1,
+                    )
+                    await self.push_frame(frame)
+            except Exception as e:
+                logger.error(f"{self} exception: {e}")
```
markbackman:
```suggestion
```

src/pipecat/services/rime/tts.py
```diff
@@ -574,3 +579,337 @@ async def run_tts(self, text: str) -> AsyncGenerator[Frame, None]:
...
+                await self.push_error(ErrorFrame(error=f"{self} error: {e}"))
```
markbackman:
```suggestion
                await self.push_error(error_msg=f"Error: {msg}")
```

markbackman (commented):
Generally looks good. Just a handful of comments.

Don't forget to add a changelog entry.

changelog/3085.added.md
```diff
@@ -0,0 +1,4 @@
+- Added support for non-JSON streaming mode in `RimeTTSService`. The service now
```
markbackman: The service name should be `RimeNonJsonTTSService`, right?

```suggestion
- Added support for non-JSON streaming mode in `RimeNonJsonTTSService`. The service now
```

markbackman: And really, this is a totally new service class. Maybe it should read:
```
- Added `RimeNonJsonTTSService` which supports non-JSON streaming mode. This new class supports websocket streaming for the Arcana model.
```

Or something along those lines.

src/pipecat/services/rime/tts.py
```diff
@@ -31,7 +31,12 @@
     TTSStoppedFrame,
 )
 from pipecat.processors.frame_processor import FrameDirection
-from pipecat.services.tts_service import AudioContextWordTTSService, TTSService
+from pipecat.services.tts_service import (
+    AudioContextWordTTSService,
+    TTSService,
+    WebsocketTTSService,
+)
+from pipecat.transcriptions import language
```
markbackman: Remove unused import:
```suggestion
```

src/pipecat/services/rime/tts.py
```diff
@@ -608,3 +613,338 @@ async def run_tts(self, text: str) -> AsyncGenerator[Frame, None]:
...
+    that require plain text (not JSON) messages and return raw audio bytes.
+    It is designed for use with TTS models like Arcana, which currently do
+    not support JSON-based WebSocket protocols (though this may change in
+    the future).
+
+    Limitations:
+        - Does not support word-level timestamps or context IDs.
+        - Intended specifically for integrations where the TTS provider only
+          accepts and returns non-JSON messages.
+        - Arcana and similar models may add JSON WebSocket support in the
```
markbackman, line 632: This is more of a note for the future.

```suggestion
      Note:
          - Arcana and similar models may add JSON WebSocket support in the
```

gokuljs:
@markbackman let me test this, and I will get back to you on this.

gokuljs:
@markbackman One of the key observations is that it works correctly in `InterruptibleTTSService` but not in `WebsocketTTSService`. I suspect this points to an issue in our clear-message mechanism. In `InterruptibleTTSService` the client disconnects and reconnects, which is likely why we are not seeing the problem there.

markbackman:
> @markbackman One of the key observations is that it works correctly in `InterruptibleTTSService` but not in `WebsocketTTSService`. I suspect this points to an issue in our clear-message mechanism. In `InterruptibleTTSService` the client disconnects and reconnects, which is likely why we are not seeing the problem there.

The `InterruptibleTTSService` has a `_handle_interruption()` method:
```
    async def _handle_interruption(self, frame: InterruptionFrame, direction: FrameDirection):
        if self._bot_speaking:
            await self._disconnect()
            await self._connect()
```

This disconnects and reconnects the websocket. Ideally, this is avoided because there are scenarios where the websocket teardown and reconnection take time.

Ideally, we would use the CLEAR message to clear the current generation requests as it's more precise. If that's not possible, then we should remove the sending of the CLEAR message and instead rely on the `InterruptibleTTSService` base class.

gokuljs:
@markbackman After investigating with the backend team, it turns out the CLEAR message doesn’t actually interrupt the audio, which explains why audio continues after CLEAR. The non JSON path wasn’t designed to perform a full cancellation. Given that, I think the next step is to rely on InterruptibleTTSService for real interruption and remove our _handle_interruption logic.

PR #3087, opened 2025-11-18
Ensure to push LLMFullResponseEndFrame if the sentence is empty.

filipi87:
Ensure to push LLMFullResponseEndFrame if the sentence is empty.

### DIFF
--- a/src/pipecat/services/tts_service.py
+++ b/src/pipecat/services/tts_service.py
@@ -364,7 +364,10 @@ class TTSService(AIService):
                 sentence, includes_inter_frame_spaces=includes_inter_frame_spaces
             )
             if isinstance(frame, LLMFullResponseEndFrame):
-                if self._push_text_frames:
+                # Ensure the sentence is not empty. In some cases, such as function calling, the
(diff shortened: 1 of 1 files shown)


src/pipecat/services/tts_service.py
```diff
@@ -364,7 +364,10 @@ async def process_frame(self, frame: Frame, direction: FrameDirection):
                 sentence, includes_inter_frame_spaces=includes_inter_frame_spaces
             )
             if isinstance(frame, LLMFullResponseEndFrame):
-                if self._push_text_frames:
+                # Ensure the sentence is not empty. In some cases, such as function calling, the
+                # LLMFullResponseEndFrame may never be pushed into the pipeline, which would
+                # cause issues if we proceed without "or not sentence" this check.
+                if self._push_text_frames or not sentence:
```
filipi87, line 370: @aconchillo , @markbackman , I believe we should investigate if this is the right fix or not for this.

filipi87: Once we are happy with the fix, we can restore the changes from [this PR](https://github.com/pipecat-ai/pipecat/pull/3090).

PR #3101, opened 2025-11-20
feat: Add speaking rate control to Inworld TTS service.

hwuiwon:

- Add speaking rate control to Inworld TTS service.

- Fixed bug: parameters should be in camel case, not snake case according to the api docs

https://docs.inworld.ai/api-reference/ttsAPI/texttospeech/synthesize-speech-stream#body-audio-config-speaking-rate


aconchillo:
LGTM! Do you mind adding a line to the CHANGELOG (following other entries).

hwuiwon:
@aconchillo sounds good! I added to the CHANGELOG

PR #3106, opened 2025-11-20
Fix sample_rate issue in ElevenLabsRealtimeSTTService, add timestamps…

markbackman:
… and logging

Now that reference docs are published, circling back to this service to fix an issue and add more features.

Reference docs:
https://elevenlabs.io/docs/api-reference/speech-to-text/v-1-speech-to-text-realtime

Also:
- Fixes MiniMax TTS changelog entries (they were mistakenly in the last release).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -305,19 +305,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Updated language mappings for the Google and Gemini TTS services to match
   official documentation.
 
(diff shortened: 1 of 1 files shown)


tenpura-shrimp:
well I just spent an hour finding out about this bug, but nice to know it's fixed now.

Do you know when the fix will be released to pypi?

markbackman:
> well I just spent an hour finding out about this bug, but nice to know it's fixed now.
> 
> Do you know when the fix will be released to pypi?

Ideally by tomorrow (Nov 26).

PR #3113, opened 2025-11-21
Initial implementation of `AWSBedrockAgentCoreProcessor`

kompfner:
Here's the associated example: https://github.com/pipecat-ai/pipecat-examples/pull/120

### DIFF
--- a/src/pipecat/services/aws/agent_core.py
+++ b/src/pipecat/services/aws/agent_core.py
@@ -178,11 +178,7 @@ class AWSAgentCoreProcessor(FrameProcessor):
         """Push a text frame, managing output response bookends."""
         # Cancel any pending close task
         if self._close_task and not self._close_task.done():
-            self._close_task.cancel()
-            try:
-                await self._close_task
-            except asyncio.CancelledError:
-                pass
+            await self.cancel_task(self._close_task)
 
         # Open output response if needed
         if not self._output_response_open:
@@ -194,7 +190,7 @@ class AWSAgentCoreProcessor(FrameProcessor):
         self._last_text_frame_time = asyncio.get_event_loop().time()
 
         # Schedule closing the output response after timeout
-        self._close_task = asyncio.create_task(self._close_output_response_after_timeout())
+        self._close_task = self.create_task(self._close_output_response_after_timeout())
 
     async def process_frame(self, frame: Frame, direction: FrameDirection):
         """Process incoming frames and handle LLM message frames.
@@ -255,11 +251,7 @@ class AWSAgentCoreProcessor(FrameProcessor):
                 # Final close if output response is still open after all parts processed
                 if self._output_response_open:
                     if self._close_task and not self._close_task.done():
-                        self._close_task.cancel()
-                        try:
-                            await self._close_task
-                        except asyncio.CancelledError:
-                            pass
+                        await self.cancel_task(self._close_task)
                     self._output_response_open = False
                     await self.push_frame(LLMFullResponseEndFrame())
         else:


src/pipecat/services/aws/agent_core.py
```diff
@@ -0,0 +1,189 @@
...
+                ) as client:
+                    response = await client.invoke_agent_runtime(
+                        agentRuntimeArn=self._agentArn, payload=json.dumps(payload).encode()
+                    )
+                    is_multi_part_response = "text/event-stream" in response.get("contentType", "")
+                    async for part in response.get("response", []):
+                        part_string = part.decode("utf-8")
+                        if is_multi_part_response:
+                            part_string = part_string[6:]  # omit "data: "
+                        print(f"[pk] AWS AgentCore response: {part_string}")
```
kompfner: todo: remove

src/pipecat/services/aws/agent_core.py
```diff
@@ -0,0 +1,189 @@
...
+                        part_json = json.loads(part_string)
+                        text = self._response_to_text_transformer(part_json)
+                        if text:
+                            await self.push_frame(LLMFullResponseStartFrame())
```
kompfner: not quite sure why i had to sandwich every single part of the response in `LLMFullResponseStartFrame`/`LLMFullResponseEndFrame` for TTS to work, but i had to...can investigate

src/pipecat/services/aws/agent_core.py
```diff
@@ -0,0 +1,266 @@
...
+        # Store AWS session parameters for creating client in async context
+        self._aws_params = {
+            "aws_access_key_id": aws_access_key or os.getenv("AWS_ACCESS_KEY_ID"),
+            "aws_secret_access_key": aws_secret_key or os.getenv("AWS_SECRET_ACCESS_KEY"),
+            "aws_session_token": aws_session_token or os.getenv("AWS_SESSION_TOKEN"),
+            "region_name": aws_region or os.getenv("AWS_REGION", "us-east-1"),
+        }
+
+        # Set transformers with defaults
+        self._context_to_payload_transformer = (
```
markbackman, line 157: This seems like a nice, flexible approach. In reading the default_context_to_payload_transformer, I wondered why only the last message. This now makes sense. Developers can provide their own transformer to handle data from their agent. Cool!

src/pipecat/services/aws/agent_core.py
```diff
@@ -0,0 +1,266 @@
...
+        The extracted output text, or None if no text found.
+    """
+    response_json = json.loads(response_line)
+    return response_json.get("response")
+
+
+class AWSAgentCoreProcessor(FrameProcessor):
+    """Processor that runs an Amazon Bedrock AgentCore agent.
+
+    Input::
```
markbackman: Nit: `::` will insert a code block if there's a newline after the `::`. So, you should either add the newline or drop the second colon.

src/pipecat/services/aws/agent_core.py
```diff
@@ -0,0 +1,266 @@
...
+
+        # State for managing output response bookends
+        self._output_response_open = False
+        self._last_text_frame_time: Optional[float] = None
+        self._close_task: Optional[asyncio.Task] = None
+        self._output_response_timeout = 1.0  # seconds
+
+    async def _close_output_response_after_timeout(self):
+        """Close the output response after timeout if no new text frames arrive."""
+        await asyncio.sleep(self._output_response_timeout)
```
markbackman, line 172: In other places, we've used asyncio.wait_for() for similar logic. Here's one place like that:
https://github.com/pipecat-ai/pipecat/blob/main/src/pipecat/extensions/voicemail/voicemail_detector.py#L372-L387

This is a nit. What you have will work. Just suggesting this because we've used this pattern elsewhere.

src/pipecat/services/aws/agent_core.py
```diff
@@ -0,0 +1,266 @@
...
+        if not self._output_response_open:
+            await self.push_frame(LLMFullResponseStartFrame())
+            self._output_response_open = True
+
+        # Push the text frame
+        await self.push_frame(LLMTextFrame(text))
+        self._last_text_frame_time = asyncio.get_event_loop().time()
+
+        # Schedule closing the output response after timeout
+        self._close_task = asyncio.create_task(self._close_output_response_after_timeout())
```
markbackman: Pipecat has its own task manager, so you can just call self.create_task() and self.cancel_task().

src/pipecat/services/aws/agent_core.py
kompfner: Argh each time I do it one way it turns out I was supposed to do it the other way. I need to be able to validate this myself instead of just guessing...

kompfner: Fixed

src/pipecat/services/aws/agent_core.py
kompfner: Good call. Will come back for this refactor...running out of time today and want to get this in.

src/pipecat/services/aws/agent_core.py
kompfner: Good call. Cmon Claude you should know this 😛

kompfner: For some reason when I do the replacement I keep seeing this...

Screenshot 2025-11-24 at 4 46 51 PM

Even though I most definitely am doing `await`s everywhere.

kompfner: Though, AFAICT there is no impact on the functioning of the coe.

Issue #3110, opened 2025-11-21
Add true WebSocket support for Deepgram TTS


agilebean:
### Problem Statement

Pipecat's DeepgramTTSService currently uses REST API (`asyncrest.stream_raw()`) despite the misleading name, resulting in ~3s STT-TTS latency. Deepgram offers a true WebSocket TTS API
  at `wss://api.deepgram.com/v1/speak` that could achieve <1s latency.

  Deepgram added WebSocket TTS support on April 9, 2024 (7 months ago) as documented in their https://deepgram.com/changelog/websocket-text-to-speech-api. The
  https://developers.deepgram.com/docs/tts-websocket-streaming shows how it enables true streaming with immediate audio playback as chunks arrive.

### Proposed Solution

 Implement a new DeepgramWebSocketTTSService class that uses Deepgram's WebSocket API.
```
  class DeepgramWebSocketTTSService(TTSService):
      def __init__(self, api_key: str, voice: str, sample_rate: int = 16000):
          self._ws_url = f"wss://api.deepgram.com/v1/speak?voice={voice}&encoding=linear16&sample_rate={sample_rate}"
          self._api_key = api_key

      async def connect(self):
          self._websocket = await websocket_connect(
              self._ws_url,
              extra_headers={"Authorization": f"Token {self._api_key}"}
          )

      async def run_tts(self, text: str):
          # Send text/tokens as they arrive
          await self._websocket.send(json.dumps({"text": text}))

          # Receive audio chunks
          async for message in self._websocket:
              if isinstance(message, bytes):
                  yield TTSAudioRawFrame(audio=message, ...)
```
### Benefits:
  - True streaming: Audio playback begins with first chunk
  - Interruption handling: Can stop TTS immediately when user interrupts

### Alternative Solutions

  1. Use DeepgramHttpTTSService - However, this has critical issues as documented in https://github.com/pipecat-ai/pipecat/issues/3109:
    - Crashes with buffer alignment errors when chunks aren't multiples of 2 bytes
    - Would require full response buffering for zstd decompression, causing 6-7s latency
  2. Continue using DeepgramTTSService (REST) - Current approach with ~3s latency, which is suboptimal for real-time applications
  3. Custom implementation outside pipecat - Users implement their own WebSocket connection, but this defeats the purpose of using pipecat's unified interface

### Additional Context

According to Deepgram's documentation, their WebSocket API:
  - Supports 40+ concurrent connections
[...]

PR #3115, opened 2025-11-21
Update DeepgramTTSService to use Deepgram's Websocket TTS API

markbackman:

The Deepgram Websocket API for TTS handles interruptions via a Clear message, which makes it a good fit for Pipecat. This services uses that API, changing the `DeepgramTTSService` class to use this new implementation.

This leaves two Deepgram TTS services:
- `DeepgramTTSService`: websocket API
- `DeepgramHttpTTSService`: HTTP API

Resolves https://github.com/pipecat-ai/pipecat/issues/3110


hwuiwon:
@markbackman when will this be released?

markbackman:
> @markbackman when will this be released?

Hopefully this week.

aconchillo:
I think it's better to rename the current `DeepgramTTSService` to `DeepgramHttpTTSService` and don't deprecate it in the future. We do this for other services, Rime, Cartesia, etc. And the websocket service becomes `DeepgramTTSService`. It might be a breaking change but at least we don't keep things the same for all services.

aconchillo:
> I think it's better to rename the current `DeepgramTTSService` to `DeepgramHttpTTSService` and don't deprecate it in the future. We do this for other services, Rime, Cartesia, etc. And the websocket service becomes `DeepgramTTSService`. It might be a breaking change but at least we don't keep things the same for all services.

Sorry, there was a misunderstanding on my side (there are three imlementation, websocket, http and slow http). We should make `DeepgramWebsocketTTSService` the default `DeepgramTTSService` even if it's a breaking change. The slow HTTP version goes away and the good HTTP is renamed to `DeepgramHttpTTSService`

examples/foundational/07c-interruptible-deepgram.py
```diff
@@ -62,7 +62,9 @@ async def run_bot(transport: BaseTransport, runner_args: RunnerArguments):
 
     stt = DeepgramSTTService(api_key=os.getenv("DEEPGRAM_API_KEY"))
 
-    tts = DeepgramTTSService(api_key=os.getenv("DEEPGRAM_API_KEY"), voice="aura-2-andromeda-en")
+    tts = DeepgramWebsocketTTSService(
```
aconchillo: This needs to be updated.

aconchillo:
LGTM! Only missing the example update.

examples/foundational/07c-interruptible-deepgram.py
markbackman: Fixed!

PR #3117, opened 2025-11-22
Fix 11labs realtime dynamic updates

DanteNoguez:

**Problem:** Dynamic settings updates via `STTUpdateSettingsFrame` were not working for the ElevenLabs Realtime STT service.

The `ElevenLabsRealtimeSTTService` was missing the `_settings` attribute that the base `STTService` class expects for dynamic configuration updates. When `STTUpdateSettingsFrame` triggers the `_update_settings` method in the base class, it checks `if key in self._settings` before applying any changes. Without this attribute properly initialized, all dynamic setting updates were silently ignored.

**Solution:**
1. Added `_settings` initialization in `ElevenLabsRealtimeSTTService` to enable dynamic settings updates via `STTUpdateSettingsFrame`
2. Improved language code handling in `set_language()` by using the `language_to_elevenlabs_language()` helper function to ensure proper language code formatting when updating settings

**Testing:**
- Verified that `STTUpdateSettingsFrame` now successfully updates language settings
- Confirmed that language changes trigger proper reconnection with the new settings

### DIFF
--- a/src/pipecat/services/elevenlabs/stt.py
+++ b/src/pipecat/services/elevenlabs/stt.py
@@ -459,9 +459,7 @@ class ElevenLabsRealtimeSTTService(WebsocketSTTService):
         self._audio_format = ""  # initialized in start()
         self._receive_task = None
 
-        self._settings = {
-            "language": params.language_code
-        }
+        self._settings = {"language": params.language_code}
 
     def can_generate_metrics(self) -> bool:
         """Check if the service can generate processing metrics.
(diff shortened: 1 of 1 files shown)


DanteNoguez:
> Nice catch! Mind adding a changelog entry?
> 
> ```
> ### Fixed
> 
> - Fixed an issue in `ElevenLabsRealtimeSTTService` where dynamic language updates were not work.
> ```
> 
> (or something like that)
> 
> Also, please link the code: scripts/fix-ruff.sh (or install pre-commit hook: `uv run pre-commit install`).

Done! @markbackman

PR #3120, opened 2025-11-24
allow waiting for all function calls to complete

aconchillo:

This PR adds a `wait_for_all` flag to the `LLMService` base class. This flag allows waiting for all function calls to complete before returning any result to the LLM. So, when `wait_for_all` is True we won't return any `IN_PROGRESS` result to the LLM.

This is necessary for reasoning models as both the thinking and tool result are part of the same assistant turn.

### DIFF
--- a/uv.lock
+++ b/uv.lock
@@ -36,12 +36,12 @@ wheels = [
 
 [[package]]
 name = "aic-sdk"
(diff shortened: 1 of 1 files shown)


aconchillo:
Did some more tests and fixed function call cancellation issues.

PR #3121, opened 2025-11-24
fix: Handle race condition in ElevenLabs TTS context management

jamsea:
## Problem

Fixes issue where the bot doesn't speak after upgrading to version 0.0.92+, appearing as "ai_not_responsive" errors.

### Root Cause
Race condition between interruption handling and audio delivery in ElevenLabs TTS service:

1. **TTS generation starts** - Audio request sent to ElevenLabs
2. **User interrupts early** (e.g., 43ms later) - Says "Hello?"
3. **InterruptionFrame closes context** - Sets `context_id = None`
4. **Audio arrives from ElevenLabs** (e.g., 227ms after generation)
5. **Audio incorrectly dropped** - Logged as "unavailable context"

The audio was generated but discarded because the context was already invalidated.

### Timeline from Production Logs
```
18:22:21.475 - TTS Generation Started
18:22:21.518 - User Interruption (43ms later)
18:22:21.702 - Audio arrives, ERROR: "Ignoring message from unavailable context"
```

## Solution

Added tracking for recently closed contexts to properly identify and handle delayed audio:

### Changes
- **Added `_last_closed_context_id` field** - Tracks the most recently closed context ID
- **Store context ID before clearing** - During interruption, preserve the ID before setting to None
- **Check for recently closed context** - When audio arrives, check if it's from a recently interrupted context
- **Proper discard with logging** - Delayed audio from interrupted contexts is now properly identified and discarded with clear debug messages
- **Clear on new context** - Reset `_last_closed_context_id` when starting a new context

### Why the Previous Fix Didn't Work
The existing fix attempted to recreate contexts for delayed messages, but the check `if self._context_id == received_ctx_id` would never match because `context_id` was already set to `None` during interruption.

## Testing
- Verified fix addresses the race condition described in issue report
- Ensures delayed audio from interrupted contexts is properly discarded
- Maintains compatibility with existing context management logic

## References
- Related to changes introduced in v0.0.92 (October 31, 2025)
- Commit 662f0487 attempted partial fix (July 24, 2025)


markbackman:
I think we can close this out as the customer that reported this issue found a smoking gun causing this issue. Just checking with @jamsea before we do that.

sameshl:
Hey @markbackman, wondering if this PR being closed implies this bug was fixed on the latest version?

I am facing this exact issue ("ai_not_responsive" due to incorrect handling of context_id for ElevenLabs) in v0.0.96. Happy to open an issue for this but I was wondering if anything is wrong in my setup.

For repro, you can try running https://github.com/pipecat-ai/pipecat/blob/main/examples/foundational/07d-interruptible-elevenlabs.py.

PR #3123, opened 2025-11-24
Use push_error in services

markbackman:

@filipi87 the easiest way to review was to go through services with looking for `logger.error` and replace with `push_error()` where it makes sense. Here's my review. You might want to take a careful look to see where I'm using `push_error()`.

My goal was to eliminate the use of `logger.error()` and replace with `push_error()` as it will give developers more control over when to detect a service failure and switch away from it. I specifically looked in src/pipecat/services.

### DIFF
--- a/src/pipecat/services/aws/stt.py
+++ b/src/pipecat/services/aws/stt.py
@@ -181,7 +181,7 @@ class AWSTranscribeSTTService(STTService):
                 try:
                     await self._connect()
                 except Exception as e:
-                    await self.push_error(exception=e)
+                    yield ErrorFrame(error=f"{self} error: {e}")
                     return
 
             # Format the audio data according to AWS event stream format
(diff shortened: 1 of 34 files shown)


src/pipecat/services/asyncai/tts.py
```diff
@@ -333,16 +332,14 @@ async def run_tts(self, text: str) -> AsyncGenerator[Frame, None]:
                 await self._get_websocket().send(msg)
                 await self.start_tts_usage_metrics(text)
             except Exception as e:
-                logger.error(f"{self} exception: {e}")
-                yield ErrorFrame(error=f"{self} error: {e}")
+                await self.push_error(exception=e)
```
filipi87: @markbackman , regarding this kind of change, for everything inside an `AsyncGenerator`, instead of invoking `push_error` directly, I was always yielding all the frames, including the `ErrorFrame`.

And only inside the [process_generator](https://github.com/pipecat-ai/pipecat/blob/c6fcf58be4a63de1819e30e301f37cc5752c0d3f/src/pipecat/services/ai_service.py#L157), it is where we were actually pushing the `ErrorFrame`.

What do you think the pattern should be here? Because it feels more correct to always yield the frames to the generator instead of pushing them directly.

filipi87: If the goal was simply to remove the `logger`, then perhaps the logging should be moved inside `push_error_frame`.

PR #3124, opened 2025-11-25
Improve TTSService handling of long LLM token outputs

markbackman:

The motivation for this change was discovering that Google Gemini outputs long chunks, sometimes containing multiple sentences. The solution (I think) is to just call `aggregate()` again on the text until all chunks/sentences are processed.


markbackman:
Closing in favor of a different solution that I'll work on in another branch.

PR #3128, opened 2025-11-25
[DO NOT MERGE] Amazon Bedrock AgentCore exploration - WebSocket edition

kompfner:


examples/aws-agentcore/client/src/app.ts
```diff
@@ -0,0 +1,266 @@
...
+      this.log("Initializing devices...");
+      await this.pcClient.initDevices();
+
+      this.log("Connecting to bot...");
+      // await this.pcClient.startBotAndConnect({
+      //   // The baseURL and endpoint of your bot server that the client will connect to
+      //   endpoint: 'http://localhost:7860/connect',
+      // });
+      const aws = new AwsClient({
+        accessKeyId: "<YOUR_ACCESS_KEY>",
```
kompfner, line 197: These need to be filled in

examples/aws-agentcore/bot.py
```diff
@@ -0,0 +1,241 @@
...
+        },
+    ]
+
+    context = LLMContext(messages, tools)
+    context_aggregator = LLMContextAggregatorPair(context)
+
+    # TODO: clean up. Hackily overrides transport created by bot() entrypoint...
+    ws_transport = WebsocketServerTransport(
+        # host=public_ip,
+        port=8080,  # This is the only port we're allowed to bind to...but the problem is it's already taken in order to support the /invoke HTTP entrypoint.
```
kompfner, line 155: Here's the crux of the problem. Port 8080 is already being used to expose the invoke method, which starts the bot (which sets up the pipeline, etc.)

The thing to try next, which is a bit more complicated, is to make it so:
- We _don't_ use the default app entrypoint (`@app.entrypoint`) - we may have to abandon using `BedrockAgentCoreApp` entirely.
- Instead, on the `if __name__ == "__main__"` section, we set up a WebSocket server on port 8080
- Only when we connect we kick off the code to set up the Pipecat pipeline...
- ...and somehow pass that already-established connection to use as the Pipecat WebSocket transport

markbackman:
@kompfner should we keep this open?

kompfner:
> @kompfner should we keep this open?

Not at all. Closing.

PR #3130, opened 2025-11-25
update nvidia services naming

vipyne:


src/pipecat/services/nim/llm.py
```diff
@@ -16,7 +16,7 @@
 from pipecat.services.openai.llm import OpenAILLMService
 
 
-class NimLLMService(OpenAILLMService):
+class NvidiaLLMService(OpenAILLMService):
```
markbackman: This change is breaking, which we want to avoid. Instead, I think we should:
- Add new services in nvidia/* (which you've already done)
- Leave the currently named services in place at their current path with:
  - Deprecation warnings
  - Rather than keep the service code in the old files, just return the new Nvidia*Service code instead.

You can see an example of this from the GeminiMultimodalLiveLLMService deprecation here:
https://github.com/pipecat-ai/pipecat/blob/main/src/pipecat/services/gemini_multimodal_live/gemini.py

pyproject.toml
```diff
@@ -94,6 +94,7 @@ qwen = []
 remote-smart-turn = []
 rime = [ "pipecat-ai[websockets-base]" ]
 riva = [ "nvidia-riva-client~=2.21.1" ]
+nvidia = [ "nvidia-riva-client~=2.21.1" ]
```
markbackman: Can you change the position to keep it alphabetical?

src/pipecat/services/nvidia/__init__.py
```diff
@@ -0,0 +1,16 @@
+#
```
markbackman: I think this file can be left blank. The DeprecatedModuleProxy was for preexisting modules where we were informing developers to use full module paths to access service classes.

CHANGELOG.md
```diff
@@ -300,6 +300,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
   deprecated, use `test_normalization` instead.
 
+- NVIDIA Services name changes (all functionality is unchanged):
+
+  - `NimLLMService` is now deprecated, use `NvidiaLLMService` instead.
+  - `RivaSTTService` is now deprecated, use `NvidiaSTTService` instead.
+  - `RivaTTSService` is now deprecated, use `NvidiaTTSService` instead.
+  - Use `uv pip install pipecat-ai[nvidia]` instead of
+    `uv pip install pipecat-ai[nim]` or `uv pip install pipecat-ai[riva]`
+
```
aconchillo: This should be moved to the new `Unreleased` sections (see CHANGELOG.md.template)

src/pipecat/services/nvidia/__init__.py
aconchillo: Yep

pyproject.toml
aconchillo: Should we remove `riva` and `nim`?

aconchillo:
Since we are renaming examples, we might need to update `scripts/evals/run-release-evals.py`

pyproject.toml
vipyne: I thought we should keep `riva` and `nim` to not break anything.  but I'm ok with removing them if they aren't part of the deprecation plan.

markbackman: We can remove `nim` since it's a no-op.

Removing `riva` will require updates to workflows. Maybe we keep it for now and remove it when we remove the deprecated classes. We can set it up as the following to make managing it easier:
```
riva = [ "pipecat-ai[nvidia]" ]
```

Maybe that's the most developer friendly as it's not breaking.

PR #3131, opened 2025-11-25
fix: add timeout handling for Daily room join/leave operations

eroub:
## Problem

The Daily transport's `_join()` and `_leave()` methods can hang indefinitely when the Daily completion callbacks fail to fire. This occurs because both methods use `await future` without timeout protection, waiting for the `completion_callback` to set the future's result.

When this happens:
- The `leave()` operation never completes
- The `on_left` event never fires
- Pipeline cleanup handlers are blocked

This issue was discovered in production where a call termination hung indefinitely at `transport.leave()`, preventing cleanup.

## Solution

Add timeout protection to both `_join()` and `_leave()` operations:

1. **Added module-level timeout constants** (30 seconds each)
   - `DAILY_JOIN_TIMEOUT_SECS = 30.0`
   - `DAILY_LEAVE_TIMEOUT_SECS = 30.0`

2. **Wrapped async operations with `asyncio.wait_for()`**
   - Both methods now timeout after 30 seconds if the completion callback doesn't fire
   - Timeout errors are logged with room URL context for debugging
   - Error strings are returned following Daily transport's error conventions

3. **Symmetric error handling**
   - `_join()` returns `(None, error_msg)` tuple on timeout (matching its error signature)
   - `_leave()` returns `error_msg` string on timeout (matching its error signature)

### DIFF
--- a/src/pipecat/transports/daily/transport.py
+++ b/src/pipecat/transports/daily/transport.py
@@ -72,6 +72,10 @@ except ModuleNotFoundError as e:
 
 VAD_RESET_PERIOD_MS = 2000
 
+# Timeout constants for Daily operations
+DAILY_JOIN_TIMEOUT_SECS = 30.0
+DAILY_LEAVE_TIMEOUT_SECS = 30.0
+
 
 @dataclass
 class DailyOutputTransportMessageFrame(OutputTransportMessageFrame):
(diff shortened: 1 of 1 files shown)


markbackman:
@aconchillo you might want to take a look.

eroub:
bumping this

aconchillo:
Thank you @eroub and sorry for the delay! This is now fixed in https://github.com/pipecat-ai/pipecat/pull/5316. Closing.

PR #3132, opened 2025-11-25
Refactor end of sentence detection, normalize aggregator input

markbackman:

This PR now does a number of things:

- The `BaseTextAggregator`'s `aggregate()` function returns an `AsyncIterator`. Along with this change, all aggregators now receive text and internally split that text into characters for aggregation. This is important to ensure that aggregation happens consistently in a number of cases (e.g. the text to aggregate is <= the unit of aggregation). This also ensures that long aggregation requests (e.g. multiple sentence inputs are chunked into sentences). **This is a breaking change** which I call out in the changelog.
- The `SimpleTextAggregator` now handles end of sentence detection more robustly. Previously, cases, like `The price is $29.50.` and `Hello, Mr. Smith.` could result in premature sentence splitting (at the first period). Now, `SimpleTextAggregator` uses "lookahead" logic to buffer text up to the next non-whitespace character. This provides the NLTK tokenizer with more information as it assess whether or not text is a complete sentence.
- The `SkipTagAggregator` and `PatternPairAggregator` now inherit from the `SimpleTextAggregator`, allowing them to use the sentence detection from the base class
- The `BaseTextAggregator` adds a `flush()` method which flushes the remaining text upon receiving an `LLMFullResponseEndFrame`. This ensures that an aggregation is always pushed, even if the aggregator's logic to yield a result is not met. I've opted to use this in lieu of a timeout mechanism, which matches the previous implementation (and many other places in Pipecat).
- Fixed a bug in `PatternPairAggregator` where the pattern detection callback could fire multiple times.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,18 +15,21 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- Improved interruption handling to prevent bots from repeating themselves.
+  LLM services that return multiple sentences in a single response (e.g.,
+  `GoogleLLMService`) are now split into individual sentences before being sent
+  to TTS. This ensures interruptions occur at sentence boundaries, preventing
+  the bot from repeating content after being interrupted during long responses.
+
 - Text Aggregation Improvements:
 
-  - Updated `LLMTextProcessor` and `TTSService` to normalize text input by
-    splitting into individual characters before aggregation. This ensures proper
-    sentence boundary detection when LLMs return multiple sentences in a single
-    chunk (e.g., Google Gemini).
-  - All text aggregators now properly support character-by-character streaming
-    input.
+  - **Breaking Change**: `BaseTextAggregator.aggregate()` now returns
+    `AsyncIterator[Aggregation]` instead of `Optional[Aggregation]`. This
+    enables the aggregator to return multiple results based on the provided
+    text.
   - Refactored text aggregators to use inheritance: `SkipTagsAggregator` and
     `PatternPairAggregator` now inherit from `SimpleTextAggregator`, reusing
-    its lookahead-based sentence detection logic via
-    `_check_sentence_with_lookahead()`.
+    the base class's sentence detection logic.
 
 - Updated `AICFilter` to use Quail STT as the default model
   (`AICModelType.QUAIL_STT`). Quail STT is optimized for human-to-machine
@@ -54,6 +57,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Fixed an issue where `LLMTextFrame.skip_tts` was being overwritten by LLM
   services.
 
+- Fixed sentence aggregation to correctly handle ambiguous punctuation in
+  streaming text, such as currency ("$29.95") and abbreviations ("Mr. Smith").
+
+- Fixed bug in `PatternPairAggregator` where pattern handlers could be called
+  multiple times for `KEEP` or `AGGREGATE` patterns.
+
 - Fixed an issue in `SarvamTTSService` where the last sentence was not being
   spoken. Now, audio is flushed when the TTS services receives the
   `LLMFullResponseEndFrame` or `EndFrame`.
@@ -65,10 +74,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   incorrectly pushed after a functional call. This caused an issue with the
   voice-ui-kit's conversational panel rending of the LLM output after a
   function call.
-
-- Fixed a bug in `PatternPairAggregator` where pattern handlers could be called
-  multiple times for patterns with `MatchAction.KEEP` or `MatchAction.AGGREGATE`
-  actions.
 
 ## [0.0.96] - 2025-11-26 🦃 "Happy Thanksgiving!" 🦃
 
--- a/src/pipecat/extensions/ivr/ivr_navigator.py
+++ b/src/pipecat/extensions/ivr/ivr_navigator.py
@@ -18,8 +18,10 @@ from loguru import logger
 from pipecat.audio.dtmf.types import KeypadEntry
 from pipecat.audio.vad.vad_analyzer import VADParams
 from pipecat.frames.frames import (
+    EndFrame,
     Frame,
     LLMContextFrame,
+    LLMFullResponseEndFrame,
     LLMMessagesUpdateFrame,
     LLMTextFrame,
     OutputDTMFUrgentFrame,
@@ -149,10 +151,17 @@ class IVRProcessor(FrameProcessor):
 
         elif isinstance(frame, LLMTextFrame):
             # Process text through the pattern aggregator
-            result = await self._aggregator.aggregate(frame.text)
-            if result:
+            async for result in self._aggregator.aggregate(frame.text):
                 # Push aggregated text that doesn't contain XML patterns
                 await self.push_frame(LLMTextFrame(result.text), direction)
+
+        elif isinstance(frame, (LLMFullResponseEndFrame, EndFrame)):
+            # Flush any remaining text from the aggregator
+            remaining = await self._aggregator.flush()
+            if remaining:
+                await self.push_frame(LLMTextFrame(remaining.text), direction)
+            # Push the end frame
+            await self.push_frame(frame, direction)
 
         else:
             await self.push_frame(frame, direction)
--- a/src/pipecat/processors/aggregators/llm_text_processor.py
+++ b/src/pipecat/processors/aggregators/llm_text_processor.py
@@ -24,7 +24,6 @@ from pipecat.frames.frames import (
     LLMTextFrame,
 )
 from pipecat.processors.frame_processor import FrameDirection, FrameProcessor
-from pipecat.utils.string import split_text_by_characters
 from pipecat.utils.text.base_text_aggregator import BaseTextAggregator
 from pipecat.utils.text.simple_text_aggregator import SimpleTextAggregator
 
@@ -84,19 +83,13 @@ class LLMTextProcessor(FrameProcessor):
         await self._text_aggregator.reset()
 
     async def _handle_llm_text(self, in_frame: LLMTextFrame):
-        # Split text by characters to normalize LLM output into individual characters
-        # This ensures consistent aggregation behavior regardless of LLM chunk size
-        characters = split_text_by_characters(in_frame.text)
-
-        for character in characters:
-            aggregation = await self._text_aggregator.aggregate(character)
-            if aggregation:
-                out_frame = AggregatedTextFrame(
-                    text=aggregation.text,
-                    aggregated_by=aggregation.type,
-                )
-                out_frame.skip_tts = in_frame.skip_tts
-                await self.push_frame(out_frame)
+        async for aggregation in self._text_aggregator.aggregate(in_frame.text):
+            out_frame = AggregatedTextFrame(
+                text=aggregation.text,
+                aggregated_by=aggregation.type,
+            )
+            out_frame.skip_tts = in_frame.skip_tts
+            await self.push_frame(out_frame)
 
     async def _handle_llm_end(self, skip_tts: Optional[bool] = None):
         # Flush any remaining text
--- a/src/pipecat/utils/string.py
+++ b/src/pipecat/utils/string.py
@@ -280,27 +280,3 @@ def concatenate_aggregated_text(text_parts: List[TextPartForConcatenation]) -> s
     result = result.strip()
 
     return result
-
-
-def split_text_by_characters(text: str) -> List[str]:
-    """Split text into individual characters.
-
-    Returns each character as a separate string element, allowing character-by-character
-    processing while maintaining the ability to reconstruct the original text.
-
-    Args:
-        text: The text to split into characters.
-
-    Returns:
-        A list of individual characters.
-
-    Example::
-
-        >>> split_text_by_characters("Hello world!")
-        ["H", "e", "l", "l", "o", " ", "w", "o", "r", "l", "d", "!"]
-        >>> split_text_by_characters("Hi")
-        ["H", "i"]
-        >>> split_text_by_characters("")
-        []
-    """
-    return list(text)
--- a/src/pipecat/utils/text/pattern_pair_aggregator.py
+++ b/src/pipecat/utils/text/pattern_pair_aggregator.py
@@ -13,7 +13,7 @@ support for custom handlers and configurable actions for when a pattern is found
 
 import re
 from enum import Enum
-from typing import Awaitable, Callable, List, Optional, Tuple
+from typing import AsyncIterator, Awaitable, Callable, List, Optional, Tuple
 
 from loguru import logger
 
@@ -256,20 +256,6 @@ class PatternPairAggregator(SimpleTextAggregator):
             matches = list(match_iter)  # Convert to list for safe iteration
 
             for match in matches:
-                # Only process patterns that end at or after last_processed_position
-                # This ensures we only call handlers once when a pattern completes
-                if match.end() <= last_processed_position:
-                    # This pattern was already processed in a previous call
-                    if action != MatchAction.REMOVE:
-                        # For KEEP/AGGREGATE patterns, we still need to track them
-                        content = match.group(1)
-                        full_match = match.group(0)
-                        pattern_match = PatternMatch(
-                            content=content.strip(), type=type, full_match=full_match
-                        )
-                        all_matches.append(pattern_match)
-                    continue
-
                 content = match.group(1)  # Content between patterns
                 full_match = match.group(0)  # Full match including patterns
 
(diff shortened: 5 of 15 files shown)


CHANGELOG.md
```diff
@@ -9,6 +9,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- Added `split_text_by_spaces()` utility function to normalize LLM text output
+  into word-by-word tokens while preserving spacing. This ensures consistent
```
kompfner: Hmm...we quite recently introduced a bunch of complexity (the `includes_inter_frame_spaces` stuff) that (I thought) had already solved this problem...or at least a very similar problem—consistent aggregation of text in the LLM context.

Adding yet another mechanism feels like the wrong way to go—this stuff is getting close to impossible to reason about. The solution shouldn't need to be so complicated or diffuse. We should either fix or rip out the complex mechanism we already have rather than putting another layer of duct tape on. Let's maybe discuss offline.

markbackman: Summarizing our discussion: we talked about moving this logic into the LLMTextProcessor and TTSService so that any LLMService (or any custom processor) can emit text frames without requiring this formatting.

Updating this branch to include that change...

CHANGELOG.md
```diff
@@ -7,8 +7,19 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Added `split_text_by_spaces()` utility function to normalize LLM text output
```
kompfner: This comment is a bit out of date.

Actually...mention of this change should perhaps be in the `Fixed` section. Maybe something along the lines of "Made text aggregation (into sentences, for example, for sending to TTS services) more resilient to varying input text length. This fixes an immediate issue with certain LLM services that output large text chunks, and makes the mechanism more robust/future-proof going forward."

kompfner: Ah nvm, I see that you have something very similar below. Your wording sounds good. My only feedback, then, is to move it to the `Fixed` section. The user doesn't really need to be super aware of the new helper method...it's more of an "under-the-hood" fix, IMO.

src/pipecat/processors/aggregators/llm_text_processor.py
```diff
@@ -83,14 +84,19 @@ async def reset(self):
...
-                aggregated_by=aggregation.type,
-            )
-            out_frame.skip_tts = in_frame.skip_tts
-            await self.push_frame(out_frame)
+        # Split text by spaces to normalize LLM output into word-by-word chunks
+        # This ensures consistent aggregation behavior regardless of LLM chunk size
+        words = split_text_by_spaces(in_frame.text)
+
+        for word in words:
+            aggregation = await self._text_aggregator.aggregate(word)
```
kompfner: We may want to have a comment in `BaseTextAggregator` and/or its `aggregate()` method mentioning that for it to work properly it expects to receive chunks of text small enough (i.e. smaller than the unit of aggregation), maybe even mentioning `split_text_by_spaces()` by name as a useful tool to use before feeding in text, if space-delimited chunks happen to be small enough for the aggregator.

kompfner: Hmm...what if the user has supplied a custom aggregator and space-delimited chunks _aren't_ small enough to be smaller than their unit of aggregation? 🤔

src/pipecat/utils/string.py
```diff
@@ -280,3 +280,55 @@ def append_part(part: TextPartForConcatenation):
...
+    Args:
+        text: The text to split into words.
+
+    Returns:
+        A list of words with trailing spaces preserved (except the last word).
+        Leading spaces are preserved on the first word.
+
+    Example::
+
+        >>> split_text_by_spaces("Hello world!")
```
kompfner: Another thing we may want to handle: something like "Hello\nworld". 

Newline is not a space, but it maybe should be treated like one for the purpose of preparing the text for aggregation.

Also, maybe end-of-sentence characters should be considered delimiters, too, since that effects the `SimpleTextAggregator` aggregation. Like "Hello.World".

And actually, since we might be pattern matching on things like `<`, maybe those might need to be delimiters, too? Like "Hello\<foo\>world".

Seems like there are many many cases we might need to handle for this to really be robust enough to chunk text into sufficiently small pieces for any aggregator to handle it properly...unless...bear with me...what if we just fed text into the aggregator _character by character_ 🤯

src/pipecat/utils/text/skip_tags_aggregator.py
```diff
@@ -70,34 +61,34 @@ async def aggregate(self, text: str) -> Optional[Aggregation]:
...
     async def handle_interruption(self):
-        """Handle interruptions by clearing the buffer.
+        """Handle interruptions by clearing the buffer and tag state.
 
         Called when an interruption occurs in the processing pipeline,
         to reset the state and discard any partially aggregated text.
         """
-        self._text = ""
+        await super().handle_interruption()
+        self._current_tag = None
```
markbackman, line 85: Adding these, which were missing before. I think it makes sense to reset their state.

src/pipecat/utils/text/pattern_pair_aggregator.py
```diff
@@ -218,14 +219,18 @@ def on_pattern_match(
...
-    async def _process_complete_patterns(self, text: str) -> Tuple[List[PatternMatch], str]:
-        """Process all complete pattern pairs in the text.
+    async def _process_complete_patterns(
+        self, text: str, last_processed_position: int = 0
+    ) -> Tuple[List[PatternMatch], str]:
+        """Process newly complete pattern pairs in the text.
 
-        Searches for all complete pattern pairs in the text, calls the
-        appropriate handlers, and optionally removes the matches.
+        Searches for pattern pairs that have been completed since last_processed_position,
```
markbackman, line 227: In moving to character-by-character aggregation, it was possible for a given pattern to be found multiple times. This was possible before, but perhaps less likely. Now, we track the position of the last processed pattern so that we only emit the callback once per pattern.

src/pipecat/processors/aggregators/llm_text_processor.py
```diff
@@ -83,24 +84,27 @@ async def reset(self):
...
-                aggregated_by=aggregation.type,
-            )
-            out_frame.skip_tts = in_frame.skip_tts
-            await self.push_frame(out_frame)
+        # Split text by characters to normalize LLM output into individual characters
+        # This ensures consistent aggregation behavior regardless of LLM chunk size
+        characters = split_text_by_characters(in_frame.text)
+
+        for character in characters:
+            aggregation = await self._text_aggregator.aggregate(character)
```
aconchillo: Are we doing character-by-character to avoid return a list?

aconchillo: why can't we call the `split_text_by_characters` inside the `aggregate`?

aconchillo: a better pattern would be for the `aggregate` function to return an `AsyncIterator`, so you would not need the conditional:

```
async for aggregation in self._text_aggregator.aggregate(in_frame.text):
        out_frame = AggregatedTextFrame(
            text=aggregation.text,
            aggregated_by=aggregation.type,
        )
        out_frame.skip_tts = in_frame.skip_tts
        await self.push_frame(out_frame)
```

tests/test_pattern_pair_aggregator.py
```diff
@@ -136,9 +146,11 @@ async def test_multiple_patterns(self):
         self.aggregator.on_pattern_match("voice", voice_handler)
         self.aggregator.on_pattern_match("emphasis", emphasis_handler)
 
-        # Test with multiple patterns in one text block
+        # Feed text
         text = "Hello <voice>female</voice> I am <em>very</em> excited to meet you!"
-        result = await self.aggregator.aggregate(text)
+        results = []
+        async for result in self.aggregator.aggregate(text):
+            results.append(result)
```
aconchillo: This could be:

```
results = [result async for result in self.aggregator.aggregate(text)]
```

tests/test_pattern_pair_aggregator.py
```diff
@@ -151,16 +163,23 @@ async def test_multiple_patterns(self):
...
         self.assertEqual(self.aggregator.text.text, "")
 
     async def test_handle_interruption(self):
-        # Start with incomplete pattern
-        result = await self.aggregator.aggregate("Hello <test>pattern")
-        self.assertIsNone(result)
+        # Feed text with incomplete pattern
+        results = []
+        async for result in self.aggregator.aggregate("Hello <test>pattern"):
+            results.append(result)
```
aconchillo: same here:

```
results = [result async for result in self.aggregator.aggregate("Hello <test>pattern")]
```

tests/test_pattern_pair_aggregator.py
```diff
@@ -172,20 +191,22 @@ async def test_handle_interruption(self):
...
+        full_text = "Hello <test>This is sentence one. This is sentence two.</test> Final sentence."
+        results = []
 
-        # First sentence contains start of pattern but no end, so no complete pattern yet
-        self.assertIsNone(result)
-
-        # Add second part with pattern end
-        result = await self.aggregator.aggregate(" This is sentence two.</test> Final sentence.")
+        async for result in self.aggregator.aggregate(full_text):
+            results.append(result)
```
aconchillo:
```
results = [result async for result in self.aggregator.aggregate(full_text)]
```

tests/test_simple_text_aggregator.py
```diff
@@ -14,22 +14,128 @@ def setUp(self):
         self.aggregator = SimpleTextAggregator()
 
     async def test_reset_aggregations(self):
-        assert await self.aggregator.aggregate("Hello ") == None
+        # Feed text and collect results
+        results = []
+        async for agg in self.aggregator.aggregate("Hello "):
+            results.append(agg)
```
aconchillo: same here

tests/test_simple_text_aggregator.py
```diff
@@ -14,22 +14,128 @@ def setUp(self):
...
         await self.aggregator.reset()
         assert self.aggregator.text.text == ""
 
     async def test_simple_sentence(self):
-        assert await self.aggregator.aggregate("Hello ") == None
-        aggregate = await self.aggregator.aggregate("Pipecat!")
+        # Feed text: "Hello Pipecat!"
+        results = []
+        async for agg in self.aggregator.aggregate("Hello Pipecat!"):
+            results.append(agg)
```
aconchillo: same here

tests/test_simple_text_aggregator.py
```diff
@@ -14,22 +14,128 @@ def setUp(self):
...
-        aggregate = await self.aggregator.aggregate("Hello Pipecat! How are ")
-        assert aggregate.text == "Hello Pipecat!"
-        # Aggregators should strip leading/trailing spaces when returning text
-        assert self.aggregator.text.text == "How are"
-        aggregate = await self.aggregator.aggregate("you?")
-        assert aggregate.text == "How are you?"
+        # Feed text: "Hello Pipecat! How are you?"
+        results = []
+        async for agg in self.aggregator.aggregate("Hello Pipecat! How are you?"):
+            results.append(agg)
```
aconchillo: same

src/pipecat/processors/aggregators/llm_text_processor.py
markbackman: I've implemented this pattern now.

tests/test_pattern_pair_aggregator.py
markbackman: I've made this change through the three tests.

tests/test_ivr_navigation.py
```diff
@@ -334,10 +335,12 @@ async def test_normal_text_passthrough(self):
 
         frames_to_send = [
             LLMTextFrame(text="Hello, I'm trying to reach billing."),
+            LLMFullResponseEndFrame(),
```
markbackman, line 338: `LLMFullResponseEndFrame` flushes the frame.

src/pipecat/extensions/ivr/ivr_navigator.py
```diff
@@ -149,11 +151,18 @@ async def process_frame(self, frame: Frame, direction: FrameDirection):
...
             # Process text through the pattern aggregator
-            result = await self._aggregator.aggregate(frame.text)
-            if result:
+            async for result in self._aggregator.aggregate(frame.text):
                 # Push aggregated text that doesn't contain XML patterns
                 await self.push_frame(LLMTextFrame(result.text), direction)
 
+        elif isinstance(frame, (LLMFullResponseEndFrame, EndFrame)):
+            # Flush any remaining text from the aggregator
+            remaining = await self._aggregator.flush()
```
markbackman, line 160: We now need to flush the frame when receiving the `LLMFullResponseEndFrame` (and `EndFrame`). This ensures that processing happens as expected, especially for outputs like `<mode>ivr</mode>`.

tests/test_pattern_pair_aggregator.py
```diff
@@ -55,28 +49,38 @@ async def test_pattern_match_and_removal(self):
...
         # Buffer should be empty after returning a complete sentence
         self.assertEqual(self.aggregator.text.text, "")
 
     async def test_pattern_match_and_aggregate(self):
-        # First part doesn't complete the pattern
-        result = await self.aggregator.aggregate("Here is code <code>pattern")
-        self.assertEqual(result.text, "Here is code")
-        self.assertEqual(self.aggregator.text.text, "<code>pattern")
-        self.assertEqual(self.aggregator.text.type, "code_pattern")
+        results = []
```
aconchillo: not needed

aconchillo (approved):

aconchillo:
LGTM!

src/pipecat/processors/aggregators/llm_text_processor.py
```diff
@@ -83,8 +83,7 @@ async def reset(self):
         await self._text_aggregator.reset()
 
     async def _handle_llm_text(self, in_frame: LLMTextFrame):
-        aggregation = await self._text_aggregator.aggregate(in_frame.text)
-        if aggregation:
+        async for aggregation in self._text_aggregator.aggregate(in_frame.text):
```
kompfner, line 86: 👍  Nice that we decided it was OK to update the signature of `aggregate()`.

src/pipecat/utils/text/simple_text_aggregator.py
```diff
@@ -41,30 +42,87 @@ def text(self) -> Aggregation:
...
-    async def aggregate(self, text: str) -> Optional[Aggregation]:
-        """Aggregate text and return completed sentences.
+    async def aggregate(self, text: str) -> AsyncIterator[Aggregation]:
+        """Aggregate text and yield completed sentences.
 
-        Adds the new text to the buffer and checks for end-of-sentence markers.
-        When a sentence boundary is found, returns the completed sentence and
-        removes it from the buffer.
+        Processes the input text character-by-character. When sentence-ending
+        punctuation is detected, it waits for non-whitespace lookahead before
```
kompfner, line 49: why non-whitespace lookahead? if the lookahead character was whitespace wouldn't that tell us the sentence had ended?

kompfner: ah, you address this concern in a comment below. though i'm still not entirely sure in what non-sentence-end case we'd have a whitespace after a period, say...maybe i'm just not thinking of it off the dome. perhaps i'll keep reading and see a test case for this scenario.

src/pipecat/utils/text/skip_tags_aggregator.py
```diff
@@ -11,13 +11,14 @@
...
 
-from pipecat.utils.string import StartEndTags, match_endofsentence, parse_start_end_tags
-from pipecat.utils.text.base_text_aggregator import Aggregation, AggregationType, BaseTextAggregator
+from pipecat.utils.string import StartEndTags, parse_start_end_tags
+from pipecat.utils.text.base_text_aggregator import Aggregation, AggregationType
+from pipecat.utils.text.simple_text_aggregator import SimpleTextAggregator
 
 
-class SkipTagsAggregator(BaseTextAggregator):
+class SkipTagsAggregator(SimpleTextAggregator):
```
kompfner, line 21: 👏 . Very nice that these inherit from a parent rather than re-implementing some of its logic.

src/pipecat/utils/text/simple_text_aggregator.py
markbackman: Consider this case, `Hello, Mr. Smith.`

NLTK sees `Hello, Mr.` and `Hello, Mr. ` both as complete sentences, but it turns out that not enough info is collected. With the first non-whitespace character, NLTK can correctly assert that `Hello, Mr. S" is not a sentence (but instead an acronym).

The logic accounts for this pattern as the remainder from the NLTK sentence position is carried over to the next aggregation.

kompfner: Ah yes, that was the example I wasn't thinking of! Gotcha.

kompfner: I guess we're always going to have some edge cases, huh. Consider these:

"Hello, Mr. Good to meet you."
"Hello, Mr. Goodbar."

Even with the 1-non-whitespace lookahead we really can't distinguish between sentence and non-sentence.

kompfner: I guess it's _very_ rare to end a sentence on, like "Mr." or "Dr." so we probably don't have to worry about that odd case.

markbackman: Yeah, there will always be edge cases. This is better than what we have today. We can increase the buffer, if we find there are too many edge cases that remain. At least we're now supporting common cases like prices and abbreviations and acronyms now.

PR #3134, opened 2025-11-26
Add ResembleAITTSService

markbackman:

cc @krishvadhani19 for review.


src/pipecat/services/resembleai/tts.py
```diff
@@ -0,0 +1,443 @@
...
+        # ResembleAI may send odd-length data even for PCM_16, so buffering helps us
+        # create properly aligned frames while maintaining smooth audio output
+        self._audio_buffers: dict[str, bytearray] = {}
+        self._buffer_threshold_bytes = 2208
+
+        # Jitter buffer: accumulate audio before starting playback to absorb network latency
+        # ResembleAI sends audio in bursts with 300-450ms gaps between them
+        # We need to buffer enough to cover these gaps before starting playback
+        self._jitter_buffer_bytes = 44100  # ~1000ms at 22050Hz to handle 400ms+ network gaps
+        self._playback_started: dict[str, bool] = {}  # Track if we've started playback per request
```
filipi87, line 100: I think that for all other TTS services we send the audio as soon as we get it. 

And I thought we were already handling this buffering before starting playback somewhere else in the pipeline, but I might be mistaken.

So I’m just double checking if this is really needed for this service.

filipi87: I think this is kind of the same thing with `_playback_started`. 

It feels like this service is slightly more complicated than the others we’ve developed, so I just want to confirm that this is really needed here.

markbackman: Yes, we do, but when I do that, the audio is choppy. I think this is an issue with the Resemble TTS service, which would ideally be fixed in the service itself. That is, you can't stream audio from Resemble, you have to stream to the client and then buffer. You can see the version before this without the buffering:
https://github.com/pipecat-ai/pipecat/pull/3134/commits/6c1229e3e0ac51cfc766871a3b3280c71e5539ba

cc @krishvadhani19 who wrote this code.

filipi87: I see,yeah, the downside is that we would basically always add a 400 ms latency due to the buffer. But if this is a current service limitation, I think it makes sense.

filipi87 (approved):
LGTM!

Hopefully we’ll be able to remove this audio buffer in the future, but for now it probably makes sense to keep it to prevent the audio from being choppy.

src/pipecat/services/resembleai/tts.py
markbackman: Agreed. This is a pretty big downside. The best TTS services has a P90+ latency of 200ms.

@krishvadhani19 can we avoid needing to buffer? Without it, I hear very choppy audio.

krishvadhani19:
> cc @krishvadhani19 for review.

Hey @markbackman, the buffer I have maintained is because resemble streams audio in bursts. We have sometimes gaps of more than 500ms. So this initial buffer helps solve it someway. I have communicated the above problem with the team. 
Without the buffer, the audio is extremely choppy.

Once fixed, I will get rid of this buffer from our service. thank you. 

cc: @filipi87

markbackman:
@krishvadhani19 thanks for the reply. Given that, we'll merge as is and once your service no longer requires this, we can remove the buffering.

krishvadhani19:
> @krishvadhani19 thanks for the reply. Given that, we'll merge as is and once your service no longer requires this, we can remove the buffering.

thank you @markbackman.

PR #3135, opened 2025-11-26
Fix race condition in DeepgramFluxSTTService reconnection

jamsea:
## Issue
Resolves the error: `cannot call recv while another coroutine is already running recv or recv_streaming` that occurs after upgrading to 0.0.95 when DeepgramFluxSTTService attempts to reconnect.

## Root Cause
The `_receive_task` and `_watchdog_task` were being created inside `_connect_websocket()`. During reconnection:
1. The existing `_receive_task_handler` continues running from the base `WebsocketService` class
2. It calls `_receive_messages()` which iterates over the websocket
3. When reconnection happens, `_connect_websocket()` is called again
4. Multiple coroutines attempt to receive from the websocket simultaneously → error

## Solution
Moved task creation from `_connect_websocket()` to `_connect()`:
- Tasks are now created only once during initial connection
- During reconnection, only the websocket connection itself is re-established
- The existing `_receive_task` continues running and uses the new websocket connection
- This matches the pattern used by other websocket services (ElevenLabs, Cartesia, Rime, etc.)

## Changes
- Moved `_receive_task` creation to `_connect()` (only created once)
- Moved `_watchdog_task` creation to `_connect()` (only created once)
- Added comments clarifying tasks are created only during initial connection
- `_connect_websocket()` now only handles low-level websocket connection

## Testing
- Verified the pattern matches other websocket services in the codebase
- Ensured proper separation of concerns between connection and task lifecycle

### DIFF
--- a/src/pipecat/services/deepgram/flux/stt.py
+++ b/src/pipecat/services/deepgram/flux/stt.py
@@ -183,6 +183,14 @@ class DeepgramFluxSTTService(WebsocketSTTService):
         """
         await self._connect_websocket()
 
+        # Creating the receiver task (only created once during initial connection)
+        if not self._receive_task:
+            self._receive_task = self.create_task(self._receive_task_handler(self._report_error))
+
+        # Creating the watchdog task (only created once during initial connection)
+        if not self._watchdog_task:
+            self._watchdog_task = self.create_task(self._watchdog_task_handler())
+
     async def _disconnect(self):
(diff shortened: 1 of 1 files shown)


markbackman:
Filipi just spent time fixing issues in Flux here:
https://github.com/pipecat-ai/pipecat/pull/3144

Closing.

PR #3136, opened 2025-11-26
Update Fish Audio default model to s1

kcui5:
Update default model from speech-1.5 to s1 for Fish Audio TTS service


markbackman (approved):
LGTM! Can you please add a changelog entry for this?

aconchillo:
Adding the CHANGELOG entry before next release.

PR #3138, opened 2025-11-26
Add session IDs to Gladia logs

clemos:
I added Gladia session IDs to all logs from GladiaSTTService for easier debug


src/pipecat/services/gladia/stt.py
```diff
@@ -278,6 +279,9 @@ def __init__(
         self._connection_task = None
         self._should_reconnect = True
 
+    def __str__(self):
+        return f"GladiaSTTService [{self._session_id}]"
```
filipi87, line 283: In all our services we log `self.name` to include the specific instance, so it would be better to keep the same pattern and just include the session id, like this:
```suggestion
        return f"{self.name} [{self._session_id}]"
```

filipi87:
@clemos, could you also add a brief description to the changelog and include a note about this log improvement? That would be great.

clemos:
closing in favor of https://github.com/pipecat-ai/pipecat/pull/3223

PR #3144, opened 2025-11-26
Fixed an issue with DeepgramFluxSTTService where it sometimes failed to reconnect.

filipi87:
Fixed an issue with DeepgramFluxSTTService where it sometimes failed to reconnect.

### DIFF
--- a/src/pipecat/services/deepgram/flux/stt.py
+++ b/src/pipecat/services/deepgram/flux/stt.py
@@ -150,7 +150,17 @@ class DeepgramFluxSTTService(WebsocketSTTService):
                     params=params
                 )
         """
-        super().__init__(sample_rate=sample_rate, **kwargs)
+        # Note: For DeepgramFluxSTTService, differently from other processes, we need to create
+        # the _receive_task inside _connect_websocket, because the websocket should only be
+        # considered connected and ready to send audio once we receive from Flux the message
+        # which confirms the connection has been established.
+        # If we try to keep the logic reconnect_on_error, when receiving a message, the
+        # _receive_task_handler would try to reconnect in case of error, invoking the
+        # _connect_websocket again and leading to a case where the first _receive_task_handler
+        # was never destroyed.
+        # So we can keep it here as false, because inside the method send_with_retry, it will
+        # already try to reconnect if needed.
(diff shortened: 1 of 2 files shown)


src/pipecat/services/deepgram/flux/stt.py
```diff
@@ -378,7 +391,6 @@ async def run_stt(self, audio: bytes) -> AsyncGenerator[Frame, None]:
                 are issues sending the audio data.
         """
         if not self._websocket:
-            yield ErrorFrame("Not connected to Deepgram Flux.")
```
filipi87, line 381: We’re already going to receive a few error frames in case of disconnection issues. Keeping this one would just be too much.

src/pipecat/services/deepgram/flux/stt.py
```diff
@@ -150,7 +150,17 @@ def __init__(
...
+        # the _receive_task inside _connect_websocket, because the websocket should only be
+        # considered connected and ready to send audio once we receive from Flux the message
+        # which confirms the connection has been established.
+        # If we try to keep the logic reconnect_on_error, when receiving a message, the
+        # _receive_task_handler would try to reconnect in case of error, invoking the
+        # _connect_websocket again and leading to a case where the first _receive_task_handler
+        # was never destroyed.
+        # So we can keep it here as false, because inside the method send_with_retry, it will
+        # already try to reconnect if needed.
+        super().__init__(sample_rate=sample_rate, reconnect_on_error=False, **kwargs)
```
filipi87, line 163: This is the only changed actually needed for this fix.

src/pipecat/services/websocket_service.py
```diff
@@ -130,24 +130,28 @@ async def _receive_task_handler(self, report_error: Callable[[ErrorFrame], Await
...
+                    if self._reconnect_on_error:
+                        success = await self._try_reconnect(report_error=report_error)
+                        if not success:
+                            break
+                    else:
+                        await report_error(ErrorFrame(message))
+                        break
+        except asyncio.CancelledError:
+            logger.debug(f"{self}: Receive task handler has been cancelled!")
+            raise
```
aconchillo: We usually (unless we have to do something) don't handle `CancelledError` since this is handled by the TaskManager and we have trace logging there.

aconchillo:
LGTM!

PR #3146, opened 2025-11-27
fix: AWSBedrockLLMService was always set to us-east-1

markbackman:

This logic stores the aws client config:
```
        self._aws_params = {
            "aws_access_key_id": aws_access_key or os.getenv("AWS_ACCESS_KEY_ID"),
            "aws_secret_access_key": aws_secret_key or os.getenv("AWS_SECRET_ACCESS_KEY"),
            "aws_session_token": aws_session_token or os.getenv("AWS_SESSION_TOKEN"),
            "region_name": aws_region or os.getenv("AWS_REGION", "us-east-1"),
            "config": client_config,
        }
```

`aws_region` would always be truthy, evaluating to `us-east-1`. Instead, the `aws_region` arg should be `None` and optional, which allows the value to be set by an env var.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -279,6 +279,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- Fixed an issue in `AWSBedrockLLMService` where the `aws_region` arg was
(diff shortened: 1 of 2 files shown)


aconchillo:
Nice catch!

PR #3147, opened 2025-11-27
Fix error handling in DeepramSageMakerSTTService

markbackman:

Aligning with the new error handling pattern.

### DIFF
--- a/src/pipecat/services/deepgram/stt_sagemaker.py
+++ b/src/pipecat/services/deepgram/stt_sagemaker.py
@@ -210,8 +210,7 @@ class DeepgramSageMakerSTTService(STTService):
             try:
                 await self._client.send_audio_chunk(audio)
             except Exception as e:
-                logger.error(f"Error sending audio to SageMaker: {e}")
-                await self.push_error(ErrorFrame(error=f"SageMaker STT error: {e}"))
+                yield ErrorFrame(error=f"Unknown error occurred: {e}")
         yield None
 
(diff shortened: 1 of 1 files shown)


src/pipecat/services/deepgram/stt_sagemaker.py
```diff
@@ -210,8 +210,7 @@ async def run_stt(self, audio: bytes) -> AsyncGenerator[Frame, None]:
             try:
                 await self._client.send_audio_chunk(audio)
             except Exception as e:
-                logger.error(f"Error sending audio to SageMaker: {e}")
-                await self.push_error(ErrorFrame(error=f"SageMaker STT error: {e}"))
+                yield ErrorFrame(error=f"Unknown error occurred: {e}")
```
aconchillo: Unfortunately, in this case we need the `{self}`.

markbackman: Ahh, ok. Adding.

aconchillo: Oh, duh! Nevermind, we already have it here: https://github.com/pipecat-ai/pipecat/blob/main/src/pipecat/processors/frame_processor.py#L680

All good! :ship:

markbackman: (We found that `push_error_frame()` already does this, so `{self}` is not needed.)

PR #3152, opened 2025-11-27
Add filters to MCP Client

RuiDaniel:
This PR enhances the MCPClient by introducing two new optional configuration mechanisms:
- tools_filter – allows registering only a selected subset of tools.
- tools_output_filters – allows post-processing tool outputs through user-provided filter functions.

These features improve security, customization, and output handling when integrating external MCP tools with LLMs.

✨ Key Changes

1. Added tools_filter support
- A new optional argument tools_filter: List[str] | None is added to the constructor.
- During tool registration (_list_tools_helper), tools not included in this list are skipped.
- This enables restricting tool usage to an explicitly whitelisted set.

2. Added tools_output_filters
- A new optional argument tools_output_filters: Dict[str, Callable[[Any], Any]] | None.
- After a tool returns its output, the client now:
  - Checks whether that tool has a configured filter function.
  - Applies the filter to the tool’s raw response.

This is especially useful for:
- Sanitizing output
- Normalizing different tool response formats
- Extracting specific fields from large payloads

✅ Benefits
- Better control over which tools an LLM is allowed to call.
- More robust output handling, reducing the need for downstream parsing logic.
- Backward compatible: if no filters are provided, behavior remains unchanged.

📌 Notes
- No breaking changes were introduced.
- Default behavior remains unchanged when both new arguments are None.

### DIFF
--- a/src/pipecat/services/mcp_service.py
+++ b/src/pipecat/services/mcp_service.py
@@ -47,7 +47,7 @@ class MCPClient(BaseObject):
         self,
         server_params: ServerParameters,
         tools_filter: Optional[List[str]] = None,
-        tools_output_filters: Dict[str, Callable[[Any], Any]] | None = None,
+        tools_output_filters: Optional[Dict[str, Callable[[Any], Any]]] = None,
         **kwargs,
     ):
         """Initialize the MCP client with server parameters.


markbackman:
@vipyne mind taking a look?

src/pipecat/services/mcp_service.py
```diff
@@ -7,7 +7,7 @@
 """MCP (Model Context Protocol) client for integrating external tools with LLMs."""
 
 import json
-from typing import Any, Dict, List, TypeAlias
+from typing import Any, Dict, List, TypeAlias, Callable
```
vipyne: run `./scripts/fix-ruff.sh` please.

src/pipecat/services/mcp_service.py
```diff
@@ -46,17 +46,24 @@ class MCPClient(BaseObject):
     def __init__(
         self,
         server_params: ServerParameters,
+        tools_filter: List[str] | None = None,
```
vipyne: nit: for Optional params like these, I would lean toward explicitly using `Optional` type.
```suggestion
        tools_filter: Optional[List[str]] = None,
```

src/pipecat/services/mcp_service.py
```diff
@@ -46,17 +46,24 @@ class MCPClient(BaseObject):
...
+            tools_filter: Optional list of tool names to register. If None, all tools are registered.
+            tools_output_filters: Optional dict mapping tool names to filter functions that process tool outputs.
+                                  Each filter function receives the raw tool output (any type) and returns the processed output (any type).
             **kwargs: Additional arguments passed to the parent BaseObject.
         """
         super().__init__(**kwargs)
         self._server_params = server_params
         self._session = ClientSession
+        self._tools_filter = tools_filter
+        self._tools_output_filters = tools_output_filters or {}
```
vipyne, line 66: this is awesome.  could you add example usage in one of the `examples/foundational/39*` MCP files?  Or add a new example?

src/pipecat/services/mcp_service.py
```diff
@@ -264,13 +271,26 @@ async def _call_tool(self, session, function_name, arguments, result_callback):
...
-        final_response = response if len(response) else "Sorry, could not call the mcp tool"
-        await result_callback(final_response)
+        # Apply output filter if configured for this tool
+        if function_name in self._tools_output_filters:
+            try:
+                response = self._tools_output_filters[function_name](response)
+                logger.debug(f"Final response (after filter): {response}")
+
+            except Exception:
+                logger.error("Error applying output filter")
```
vipyne: nit
```suggestion
                logger.error(f"Error applying output filter for {function_name}")
```

src/pipecat/services/mcp_service.py
vipyne:
```suggestion
from typing import Any, Callable, Dict, List, Optional, TypeAlias
```

src/pipecat/services/mcp_service.py
```diff
@@ -46,17 +46,24 @@ class MCPClient(BaseObject):
     def __init__(
         self,
         server_params: ServerParameters,
+        tools_filter: List[str] | None = None,
+        tools_output_filters: Dict[str, Callable[[Any], Any]] | None = None,
```
vipyne:
```suggestion
        tools_output_filters: Optional[Dict[str, Callable[[Any], Any]]] = None,
```

src/pipecat/services/mcp_service.py
vipyne: @RuiDaniel 
if you can post an example filter function here, I can add it to an example.
it would be helpful for folks to see a real-world-ish example.  maybe simple regex search and replace or something.

AndreTeixeira1998:
This looks like a really nice feature!

Many times I want to control the output of the MCP Server tool

src/pipecat/services/mcp_service.py
vipyne: https://github.com/pipecat-ai/pipecat/pull/3187

vipyne (approved):

src/pipecat/services/mcp_service.py
RuiDaniel: Thank you for the example!

PR #3156, opened 2025-11-27
fix: DeepgramTTSService, let the base class push TTSStoppedFrame

markbackman:

I found this in testing with an RTVI client. After a function call, the test in the voice-ui-kit was not creating a new line. In pulling the thread, I found that emitting the `TTSStoppedFrame` any time a `Flushed` message was received from Deepgram was a problem because the `LLMFullResponseEndFrame` comes after a function call, resulting in a flush (despite no text), causing an errant TTSStoppedFrame to be pushed.

Now, DeepgramTTSService relies on the base class to push the TTSStoppedFrame.

Also, because it's a websocket TTS service, I'm pausing processing during audio playout to prevent overlapping audio.

### DIFF
--- a/src/pipecat/services/deepgram/tts.py
+++ b/src/pipecat/services/deepgram/tts.py
@@ -71,7 +71,12 @@ class DeepgramTTSService(WebsocketTTSService):
             encoding: Audio encoding format. Defaults to "linear16".
             **kwargs: Additional arguments passed to parent InterruptibleTTSService class.
         """
-        super().__init__(sample_rate=sample_rate, **kwargs)
+        super().__init__(
+            sample_rate=sample_rate,
+            pause_frame_processing=True,
+            push_stop_frames=True,
+            **kwargs,
+        )
 
         self._api_key = api_key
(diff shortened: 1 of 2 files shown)


src/pipecat/services/deepgram/tts.py
```diff
@@ -71,7 +71,12 @@ def __init__(
             encoding: Audio encoding format. Defaults to "linear16".
             **kwargs: Additional arguments passed to parent InterruptibleTTSService class.
         """
-        super().__init__(sample_rate=sample_rate, **kwargs)
+        super().__init__(
+            sample_rate=sample_rate,
+            pause_frame_processing=True,
```
filipi87, line 76: I still need to understand this better, but does it actually start overlapping audio if we change it to false?

markbackman: No, it doesn't overlap, but I think there's a possibility for overlap to occur in some cases. Aleix would recall the history for this one, I'm sure.

PR #3159, opened 2025-11-28
feat(minimax): add streaming word-level subtitles and simplify API

zhenyujia23-crypto:
- Add subtitle_enable parameter for streaming word-level timestamps
- Remove emotion parameter and related processing logic
- Keep exclude_aggregated_audio parameter as requested
- Maintain backward compatibility with english_normalization deprecation
- Add subtitle file processing and logging
- Simplify code and remove complex validation

InputParams now has 9 parameters:
- language, speed, volume, pitch
- english_normalization (deprecated), text_normalization
- latex_read, exclude_aggregated_audio, subtitle_enable

Provides streaming subtitles when subtitle_enable=True with word-level timestamps.

### DIFF
--- a/src/pipecat/services/minimax/tts.py
+++ b/src/pipecat/services/minimax/tts.py
@@ -89,7 +89,7 @@ class MiniMaxHttpTTSService(TTSService):
     """Text-to-speech service using MiniMax's T2A (Text-to-Audio) API.
 
     Provides streaming text-to-speech synthesis using MiniMax's HTTP API
-    with support for various voice settings, emotions, and audio configurations.
+    with support for various voice settings and audio configurations.
     Supports real-time audio streaming with configurable voice parameters.
 
     Platform documentation:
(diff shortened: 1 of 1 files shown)


src/pipecat/services/minimax/tts.py
```diff
@@ -196,26 +195,6 @@ def __init__(
             if service_lang:
                 self._settings["language_boost"] = service_lang
 
-        # Add optional emotion if provided
```
markbackman, line 199: Why are you removing emotion? This feature was just added in the last release? Removing it is a breaking change. You should instead deprecate and warn about it if it's no longer relevant.

src/pipecat/services/minimax/tts.py
```diff
@@ -116,17 +114,18 @@ class InputParams(BaseModel):
             text_normalization: Enable text normalization (Chinese/English).
             latex_read: Enable LaTeX formula reading.
             exclude_aggregated_audio: Whether to exclude aggregated audio in final chunk.
+            subtitle_enable: Enable subtitle generation with word-level timestamps.
```
markbackman, line 117: Word timestamps require a broader change in this class. The subclass needs to change to reflect the change. This would become a `WordTTSService` and would need to implement `add_word_timestamps`, etc. to work with the subclass.

src/pipecat/services/minimax/tts.py
zhenyujia23-crypto: Hey markbackman, if we add emotion into the request body, the latency will be very high

markbackman:
@zhenyujia23-crypto are you still interested in this change? If so, see my latest comments.

PR #3160, opened 2025-11-28
fix: preserve skip_tts flag when LLMTextFrame passes through LLM services

jamsea:
## Problem

When sending an `LLMTextFrame` with `skip_tts=True` through an `LLMSwitcher` (or any LLM service), the frame exits with `skip_tts=False`. This happens because the `LLMService.push_frame()` override unconditionally sets `frame.skip_tts = self._skip_tts`, overwriting any value already set on the frame.

This bug was introduced in commit 16f57be72cec9f718ab1d42df258baac82b39f96.

### Workaround (before this fix)
```python
await self.push_frame(LLMConfigureOutputFrame(skip_tts=True))
```

### 1. Simple Pragmatic Fix (not chosen)
The simplest fix was to only set `skip_tts` when `self._skip_tts=True`:

```python
if isinstance(frame, (LLMTextFrame, LLMFullResponseStartFrame, LLMFullResponseEndFrame)):
    if self._skip_tts:
        frame.skip_tts = True
```

This would preserve frames with `skip_tts=True` passing through and fix the immediate bug. However, it still **modifies frames after creation**, which violates Pipecat's design principle that processors should not mutate frames.

### 2. Architectural Fix (chosen)
The proper solution is to set `skip_tts` **at frame creation time** rather than modifying it in `push_frame()`. This required:

1. Making `skip_tts` an init parameter (with `kw_only=True` to avoid breaking subclass field ordering)
2. Adding `_get_skip_tts()` helper in `LLMService` 
3. Updating all LLM services to pass `skip_tts=self._get_skip_tts()` when creating frames
4. Removing the `push_frame()` override entirely

### Frame Definitions (`frames.py`)
- `TextFrame`: `skip_tts: bool = field(default=False, kw_only=True)`
- `LLMFullResponseStartFrame`: `skip_tts: bool = field(default=False, kw_only=True)`
- `LLMFullResponseEndFrame`: `skip_tts: bool = field(default=False, kw_only=True)`

### LLM Service Base (`llm_service.py`)
- Added `_get_skip_tts()` helper method
- Removed the `push_frame()` override

### LLM Service Implementations
Updated all services to pass `skip_tts=self._get_skip_tts()` when creating frames:
- Anthropic
- AWS (Bedrock, Nova Sonic, AgentCore)
- Google (Gemini, Gemini Live, OpenAI-compatible)
- OpenAI (base, realtime)
- OpenAI Realtime Beta
- SambaNova

## Testing
- Frames with `skip_tts=True` now pass through LLM services unchanged
- `LLMConfigureOutputFrame(skip_tts=True)` continues to work as expected
- No regression in normal LLM text generation flow

### DIFF
--- a/examples/foundational/04a-transports-daily.py
+++ b/examples/foundational/04a-transports-daily.py
@@ -50,10 +50,6 @@ async def main():
             ),
         )
 
-        await transport.update_subscriptions(
-            profile_settings={"base": {"camera": "unsubscribed", "screenVideo": "unsubscribed"}}
-        )
-
         tts = CartesiaTTSService(
             api_key=os.getenv("CARTESIA_API_KEY"),
             voice_id="71a7ad14-091c-4e8e-a314-022ece01c121",  # British Reading Lady


markbackman:
@jamsea can we close this out since you have #3161 too? (Also, Aleix's version)

aconchillo:
Closing, since #3168 is now merged.

PR #3161, opened 2025-11-28
fix: preserve skip_tts flag when frames pass through LLM services (simple fix)

jamsea:
## Problem

When sending an `LLMTextFrame` with `skip_tts=True` through an `LLMSwitcher` (or any LLM service), the frame exits with `skip_tts=False`. This happens because the `LLMService.push_frame()` override unconditionally sets `frame.skip_tts = self._skip_tts`, overwriting any value already set on the frame.

This bug was introduced in commit 16f57be72cec9f718ab1d42df258baac82b39f96.

## Solution

This PR implements a simple, minimal fix: only set `skip_tts=True` when `self._skip_tts` is `True`, rather than unconditionally overwriting the value:

```python
if isinstance(frame, (LLMTextFrame, LLMFullResponseStartFrame, LLMFullResponseEndFrame)):
    if self._skip_tts:
        frame.skip_tts = True
```

### How it works

| Frame's `skip_tts` | Service's `_skip_tts` | Result |
|---------------------|------------------------|--------|
| `False` (default) | `False` (default) | `False` (unchanged) |
| `False` (default) | `True` | `True` (set by service) |
| `True` (explicit) | `False` (default) | `True` (preserved) ✅ |
| `True` (explicit) | `True` | `True` (unchanged) |

## Trade-offs

This fix still modifies frames after creation, which isn't ideal for Pipecat's architecture. However, it's a minimal change that fixes the bug without requiring updates to all LLM service implementations.

## Alternative: Architectural Fix

For a more architecturally pure solution that sets `skip_tts` at frame creation time (no frame mutation), see #3160. That approach:
- Makes `skip_tts` an init parameter on frames
- Updates all LLM services to pass `skip_tts=self._get_skip_tts()` when creating frames
- Removes the `push_frame()` override entirely

Choose this PR for a quick fix, or #3160 for a cleaner long-term solution.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -279,6 +279,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- Fixed an issue where `LLMTextFrame` with `skip_tts=True` would have its value
+  overwritten to `False` when passing through LLM services (e.g., `LLMSwitcher`).
+  The fix preserves frames that already have `skip_tts=True` set.
+
 - Fixed an issue in `AWSBedrockLLMService` where the `aws_region` arg was
   always set to `us-east-1`.
 


aconchillo:
Closing. This was merged in #3168

PR #3162, opened 2025-11-28
Add Quail STT as default model for `AICFilter`

iCorv:
### Summary
Updates `AICFilter` to use Quail STT (`AICModelType.QUAIL_STT`) as the default model. Quail STT is optimized for human-to-machine interaction (voice agents, speech-to-text) and operates at 16 kHz with fixed enhancement parameters.
### Background
Quail STT is the newest speech enhancement model from ai-coustics, optimized for human-to-machine interaction. It:
- Operates at a native sample rate of 16 kHz
- Uses fixed enhancement parameters that cannot be changed during runtime
- Is compatible with the AIC VAD system
### Changes
- Set `AICModelType.QUAIL_STT` as the default `model_type` in `AICFilter.__init__()`
- Deprecated the `noise_gate_enable` parameter (noise gating is handled by the VAD)
- Added deprecation warning when `noise_gate_enable` is used, directing users to `create_vad_analyzer()` for VAD functionality
### Notes
- Existing code continues to work; Quail STT is now the default
- Users explicitly setting a different model type are unaffected
- Users using `noise_gate_enable` will see a deprecation warning and should migrate to `create_vad_analyzer()` for VAD functionality

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,16 @@ All notable changes to **Pipecat** will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Changed
+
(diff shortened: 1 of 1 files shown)


markbackman (approved):
LGTM! Thanks for the update 🙇

PR #3165, opened 2025-11-29
Fix Cartesia STT dynamic updates

DanteNoguez:

**Problem:**
Dynamic settings updates via `STTUpdateSettingsFrame` were failing for the Cartesia STT service.

The service was storing the `CartesiaLiveOptions` object in `_settings`, but the base `STTService` class expects `_settings` to be a dictionary. This caused an error (`argument of type 'CartesiaLiveOptions' is not iterable`) during settings updates.

Additionally, the `set_language()` method was unimplemented. Even if `_settings` had been correct, the service would not have reconnected with the new language settings upon update.

**Solution:**
1. Initialized `_settings` as a dictionary using `.to_dict()` on the options object.
2. Implemented `set_language()` to handle reconnection when the language changes.
    - Note: I intentionally omitted the explicit update of `_settings["language"]` in this method. The base `STTService._update_settings()` already performs this update (at `self._settings[key] = value`) before calling `set_language()`.
3. Updated `_connect_websocket()` to use the dictionary format of `_settings` for compatibility.

**Testing:**
- Verified that `STTUpdateSettingsFrame` now successfully updates language settings.
- Confirmed that language changes trigger a proper reconnection with the new settings.

### DIFF
--- a/src/pipecat/services/cartesia/stt.py
+++ b/src/pipecat/services/cartesia/stt.py
@@ -172,7 +172,7 @@ class CartesiaSTTService(WebsocketSTTService):
                 }
             )
 
-        self._settings = merged_options
+        self._settings = merged_options.to_dict()
         self.set_model_name(merged_options.model)
         self._api_key = api_key
         self._base_url = base_url or "api.cartesia.ai"
@@ -269,7 +269,7 @@ class CartesiaSTTService(WebsocketSTTService):
                 return
             logger.debug("Connecting to Cartesia STT")
 
-            params = self._settings.to_dict()
+            params = self._settings
             ws_url = f"wss://{self._base_url}/stt/websocket?{urllib.parse.urlencode(params)}"
             headers = {"Cartesia-Version": "2025-04-16", "X-API-Key": self._api_key}
 
@@ -363,3 +363,14 @@ class CartesiaSTTService(WebsocketSTTService):
                         language,
                     )
                 )
+
+    async def set_language(self, language: Language):
+        """Set the transcription language.
+
+        Args:
+            language: The language to use for speech-to-text transcription.
+        """
+        logger.debug(f"Switching STT language to: [{language}]")
+        # Reconnect with new settings
+        await self._disconnect()
+        await self._connect()
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,13 @@ All notable changes to **Pipecat** will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Fixed 
+
+- Fixed an issue in `CartesiaSTTService` where dynamic language
+  updates were failing.
+
 ## [0.0.96] - 2025-11-26 🦃 "Happy Thanksgiving!" 🦃
 
 ### Added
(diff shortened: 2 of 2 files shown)


src/pipecat/services/cartesia/stt.py
```diff
@@ -363,3 +363,14 @@ async def _on_transcript(self, data):
...
                 )
+
+    async def set_language(self, language: Language):
+        """Set the transcription language.
+
+        Args:
+            language: The language to use for speech-to-text transcription.
+        """
+        logger.debug(f"Switching STT language to: [{language}]")
+        # Reconnect with new settings
```
markbackman, line 374: Don't you have to set the language in this function?

DanteNoguez: The base STTService._update_settings() already performs this update (at `self._settings[key] = value`) before calling set_language().

markbackman:
I tested this change out, but I see errors when the disconnect/connect happens. Do you see the same?

Also, we need some cleanup in this class, which I'm doing here:
https://github.com/pipecat-ai/pipecat/pull/3197

DanteNoguez:
@markbackman Haven't had any issues testing this... Would it make things easier for you if I add either a foundational example or a unit test to this PR (or the new one you're working on)? Or I can send you a script via Discord if you prefer. Let me know!

markbackman:
> @markbackman Haven't had any issues testing this... Would it make things easier for you if I add either a foundational example or a unit test to this PR (or the new one you're working on)? Or I can send you a script via Discord if you prefer. Let me know!

I'm running your branch by running the [07-interruptible-cartesia-http.py example](https://github.com/pipecat-ai/pipecat/blob/main/examples/foundational/07-interruptible-cartesia-http.py) and changing the `on_client_connected` event handler to:
```
        # Kick off the conversation.
        messages.append({"role": "system", "content": "Please introduce yourself to the user."})
        await asyncio.sleep(10)
        await task.queue_frame(STTUpdateSettingsFrame({"language": "es"}))
```

I see:
```
2025-12-08 16:41:31.984 | INFO     | pipecat.services.stt_service:_update_settings:148 - Updating STT settings: {'model': 'ink-whisper', 'language': 'en', 'encoding': 'pcm_s16le', 'sample_rate': '16000'}
2025-12-08 16:41:31.984 | INFO     | pipecat.services.stt_service:_update_settings:151 - Updating STT setting language to: [es]
2025-12-08 16:41:31.985 | DEBUG    | pipecat.services.cartesia.stt:set_language:372 - Switching STT language to: [es]
2025-12-08 16:41:31.985 | DEBUG    | pipecat.services.cartesia.stt:_disconnect_websocket:278 - Disconnecting from Cartesia STT
2025-12-08 16:41:32.011 | DEBUG    | pipecat.services.cartesia.stt:_connect_websocket:264 - Connecting to Cartesia STT
2025-12-08 16:41:32.016 | ERROR    | pipecat.processors.frame_processor:push_error_frame:695 - CartesiaSTTService#0 exception (/Users/mbackman/src/pipecat/.venv/lib/python3.12/site-packages/websockets/asyncio/connection.py:873): Error processing frame: sent 1000 (OK); then received 1000 (OK)
2025-12-08 16:41:32.016 | WARNING  | pipecat.pipeline.task:_source_push_frame:756 - PipelineTask#0: Something went wrong: ErrorFrame#0(error: Error processing frame: sent 1000 (OK); then received 1000 (OK), fatal: False)
```

Do you not see this?

markbackman:
We now have a complete settings update solution that works across classes, including Cartesia: https://docs.pipecat.ai/guides/fundamentals/service-settings

PR #3168, opened 2025-12-01
LLMTextFrame: don't override skip_tts

aconchillo:

This is a follow-up on #3161. It fixes the issue by making `skip_tts` optional. This allow us to avoid overriding the value if the LLM is not configured. This works well because `False` and `None` are just the same in a conditional statement.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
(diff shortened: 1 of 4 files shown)


markbackman (approved):
This seems fine. @jamsea can you confirm that this resolves the issue you saw?

jamsea:
@aconchillo this is a better way you're right 👍

aconchillo:
This could actually be useful if someone is subclassing an LLM.

PR #3171, opened 2025-12-02
Gradium integration.

LaurentMazare:
This PR integrates the [gradium api](gradium.ai) in pipecat. The api provides a streaming speech-to-text and a streaming text-to-speech models, currently supported languages are English, French, German, Portugese, Spanish.

### DIFF
--- a/src/pipecat/services/gradium/stt.py
+++ b/src/pipecat/services/gradium/stt.py
@@ -22,10 +22,7 @@ from pipecat.frames.frames import (
     Frame,
     StartFrame,
     TranscriptionFrame,
-    UserStartedSpeakingFrame,
-    UserStoppedSpeakingFrame,
 )
-from pipecat.processors.frame_processor import FrameDirection
 from pipecat.services.stt_service import WebsocketSTTService
 from pipecat.transcriptions.language import Language
 from pipecat.utils.time import time_now_iso8601
@@ -229,8 +226,7 @@ class GradiumSTTService(WebsocketSTTService):
 
     async def _handle_end_of_stream(self):
         """Handle termination message."""
-        logger.info("Received end_of_stream message from server")
-        await self.push_frame(EndFrame())
+        logger.debug("Received end_of_stream message from server")
 
     async def _handle_text(self, text: str):
         """Handle transcription results."""


src/pipecat/services/gradium/tts.py
```diff
@@ -0,0 +1,338 @@
...
+
+from loguru import logger
+from pydantic import BaseModel
+
+from pipecat.frames.frames import (
+    CancelFrame,
+    EndFrame,
+    ErrorFrame,
+    Frame,
+    InterruptionFrame,
```
markbackman: Remove the unused import:
```suggestion
```

src/pipecat/services/gradium/tts.py
```diff
@@ -0,0 +1,338 @@
...
+        self._settings = {
+            "voice": voice,
+            "voice_id": voice_id,
+            "model_name": model,
+            "output_format": "pcm",
+        }
+
+        # State tracking
+        self._receive_task = None
+        self._request_id = None
```
markbackman: The request_id isn't used for anything. For now, it should be safe to remove.

src/pipecat/services/gradium/tts.py
```diff
@@ -0,0 +1,338 @@
...
+            temp: Temperature to be used for generation, defaults to 0.6.
+        """
+
+        temp: Optional[float] = 0.6
+
+    def __init__(
+        self,
+        *,
+        api_key: str,
+        voice: Optional[str] = None,
```
markbackman: What's the difference between `voice` and `voice_id`? Are both used/useful?

src/pipecat/services/gradium/tts.py
```diff
@@ -0,0 +1,338 @@
...
+
+        Args:
+            model: The model name to use for synthesis.
+        """
+        self._model = model
+        await super().set_model(model)
+
+    async def _update_settings(self, settings: Mapping[str, Any]):
+        """Update service settings and reconnect if voice changed."""
+        prev_voice = self._voice_id
```
markbackman, line 119: If `voice` and `voice_id` are both ways to specify the voice to use, I think you'll need to add `voice` here as well to change the voice.

markbackman: (Though one way is certainly cleaner!

src/pipecat/services/gradium/tts.py
```diff
@@ -0,0 +1,338 @@
...
+        raise Exception("Websocket not connected")
+
+    async def flush_audio(self):
+        """Flush any pending audio synthesis."""
+        if not self._websocket:
+            return
+        try:
+            msg = {"type": "end_of_stream"}
+            await self._websocket.send(json.dumps(msg))
+            logger.trace(f"{self}: flushing audio")
```
markbackman: It probably makes sense to remove most trace logging.
```suggestion
```

src/pipecat/services/gradium/tts.py
```diff
@@ -0,0 +1,338 @@
...
+        """
+        _state = self._websocket.state if self._websocket is not None else None
+        logger.debug(f"{self}: Generating TTS [{text}] {_state}")
+        try:
+            if not self._websocket or self._websocket.state is State.CLOSED:
+                self._websocket = None
+                await self._connect()
+
+            try:
+                if not self._request_id:
```
markbackman: Related to above, just remove `self._request_id`. The request ID (or context ID in other classes) is used to track turns. But it's only useful in the case where you can send an ID to the TTS service to produce audio consistent across an ID. Here, it looks like it's just set in memory but doesn't affect the service.

markbackman:
In addition to these changes, you'll want to add a foundational example. I'm recommending 07af-interruptible-gradium.py. It would be as follows: 
[07af-interruptible-gradium.py](https://github.com/user-attachments/files/23943882/07af-interruptible-gradium.py)

Also, a few other to dos:
1. Add Gradium to the services list in the README: https://github.com/pipecat-ai/pipecat/blob/main/README.md
2. Add env vars to the env.example file: https://github.com/pipecat-ai/pipecat/blob/main/env.example
3. Add 07af-interruptible-gradium.py to the release evals: https://github.com/pipecat-ai/pipecat/blob/main/scripts/evals/run-release-evals.py#L78
4. Add a changelog entry.
5. Submit a docs PR: https://github.com/pipecat-ai/docs
  - update https://docs.pipecat.ai/server/services/supported-services
  - add a service page, which can be as minimal as others. Only mention custom things as needed.
  - (references are auto-generated)

src/pipecat/services/gradium/tts.py
```diff
@@ -0,0 +1,338 @@
...
+                    await self.start_ttfb_metrics()
+                    yield TTSStartedFrame()
+                    self._request_id = str(uuid.uuid4())
+
+                msg = self._build_msg(text=text)
+                await self._get_websocket().send(json.dumps(msg))
+                await self.start_tts_usage_metrics(text)
+            except Exception as e:
+                logger.error(f"{self} error sending message: {e}")
```
markbackman: We recently updated how error handling works, so you'll need to make a few changes to align. The benefit is that by following this pattern, an ErrorFrame will be emitted which developers can hook into via the `on_error` event.
```suggestion
                yield ErrorFrame(error=f"Unknown error occurred: {e}")
```

src/pipecat/services/gradium/tts.py
```diff
@@ -0,0 +1,338 @@
...
+                yield TTSStoppedFrame()
+                await self._disconnect()
+                await self._connect()
+                return
+            yield None
+        except Exception as e:
+            logger.error(f"{self} exception: {e}")
```
markbackman: Here too:

```suggestion
            yield ErrorFrame(error=f"Unknown error occurred: {e}")
```

src/pipecat/services/gradium/tts.py
```diff
@@ -0,0 +1,338 @@
...
+            ready_msg = await self._websocket.recv()
+            ready_msg = json.loads(ready_msg)
+            if ready_msg["type"] == "error":
+                raise Exception(f"received error {ready_msg['message']}")
+            if ready_msg["type"] != "ready":
+                raise Exception(f"unexpected first message type {ready_msg['type']}")
+
+            await self._call_event_handler("on_connected")
+        except Exception as e:
+            logger.error(f"{self} initialization error: {e}")
```
markbackman:
```suggestion
            await self.push_error(error_msg=f"Unknown error occurred: {e}", exception=e)
```

src/pipecat/services/gradium/tts.py
```diff
@@ -0,0 +1,338 @@
...
+            await self._call_event_handler("on_connection_error", f"{e}")
+
+    async def _disconnect_websocket(self):
+        """Close websocket connection and reset state."""
+        try:
+            await self.stop_all_metrics()
+            if self._websocket:
+                await self._websocket.close()
+        except Exception as e:
+            logger.error(f"{self} error closing websocket: {e}")
```
markbackman:
```suggestion
            await self.push_error(error_msg=f"Unknown error occurred: {e}", exception=e)
```

src/pipecat/services/gradium/tts.py
```diff
@@ -0,0 +1,338 @@
...
+                await self.add_word_timestamps([(msg["text"], msg["start_s"])])
+            elif msg["type"] == "end_of_stream":
+                await self.push_frame(TTSStoppedFrame())
+                await self.stop_all_metrics()
+
+            elif msg["type"] == "error":
+                logger.error(f"{self} error: {msg}")
+                await self.push_frame(TTSStoppedFrame())
+                await self.stop_all_metrics()
+                await self.push_error(ErrorFrame(f"{self} error: {msg['message']}"))
```
markbackman:
```suggestion
                await self.push_error(error_msg=f"Error: {msg['message']}"))
```

src/pipecat/services/gradium/tts.py
```diff
@@ -0,0 +1,338 @@
...
+                await self.push_frame(frame)
+
+            elif msg["type"] == "text":
+                await self.add_word_timestamps([(msg["text"], msg["start_s"])])
+            elif msg["type"] == "end_of_stream":
+                await self.push_frame(TTSStoppedFrame())
+                await self.stop_all_metrics()
+
+            elif msg["type"] == "error":
+                logger.error(f"{self} error: {msg}")
```
markbackman: `push_error()` will also call `logger.error()`, so we can remove here:
```suggestion
```

src/pipecat/services/gradium/stt.py
```diff
@@ -0,0 +1,256 @@
...
+# SPDX-License-Identifier: BSD 2-Clause License
+#
+
+"""Gradium's speech-to-text service implementation.
+
+This module provides integration with Gradium's real-time speech-to-text
+WebSocket API for streaming audio transcription.
+"""
+
+import asyncio
```
markbackman: Remove unused import:
```suggestion
```

src/pipecat/services/gradium/stt.py
```diff
@@ -0,0 +1,256 @@
...
+        api_key: str,
+        api_endpoint_base_url: str = "wss://eu.api.gradium.ai/api/speech/asr",
+        json_config: str | None = None,
+        **kwargs,
+    ):
+        """Initialize the Gradium STT service.
+
+        Args:
+            api_key: Gradium API key for authentication.
+            language: Language code for transcription. Defaults to English (Language.EN).
```
markbackman: `language` is not included in `__init__`, so remove from here:
```suggestion
```

src/pipecat/services/gradium/stt.py
```diff
@@ -0,0 +1,256 @@
...
+        return True
+
+    async def start(self, frame: StartFrame):
+        """Start the speech-to-text service.
+
+        Args:
+            frame: Start frame to begin processing.
+        """
+        await super().start(frame)
+        self._chunk_size_bytes = int(self._chunk_size_ms * self._sample_rate * 2 / 1000)
```
markbackman: Use `self.sample_rate` to reference to property from `super()`.
```suggestion
        self._chunk_size_bytes = int(self._chunk_size_ms * self.sample_rate * 2 / 1000)
```

src/pipecat/services/gradium/stt.py
```diff
@@ -0,0 +1,256 @@
...
+    async def _process_response(self, msg):
+        type_ = msg.get("type", "")
+        if type_ == "text":
+            await self._handle_text(msg["text"])
+        elif type_ == "end_of_stream":
+            await self._handle_end_of_stream()
+        elif type_ == "error":
+            logger.error(f"Gradium error: {msg.get('message', 'Unknown error')}")
+
+    async def _handle_end_of_stream(self):
```
markbackman, line 227: This seems like an unexpected thing to happen from a service. What case are you trying to handle?

Instead, perhaps an ErrorFrame should be pushed. Since Pipecat supports a `ServiceSwitcher` class, it's not a good assumption to assume that a service stopping or failing will mean that the pipeline should not run. Instead, I would push an error here:

```
await self.push_error(error_msg=f"Received end of stream from Gradium server")
```

src/pipecat/services/gradium/stt.py
```diff
@@ -0,0 +1,256 @@
...
+            chunk = bytes(self._audio_buffer[: self._chunk_size_bytes])
+            self._audio_buffer = self._audio_buffer[self._chunk_size_bytes :]
+            chunk = base64.b64encode(chunk).decode("utf-8")
+            msg = {"type": "audio", "audio": chunk}
+            if self._websocket and self._websocket.state is State.OPEN:
+                await self._websocket.send(json.dumps(msg))
+
+        yield None
+
+    async def process_frame(self, frame: Frame, direction: FrameDirection):
```
markbackman: I tihnk you can remove the `process_frame()` method all together.

Instead, just add:
```
```

to the `run_stt()` method. We have work to do to improve the STT metrics anyway. But putting them in run_stt will be more consistent.

src/pipecat/services/gradium/stt.py
```diff
@@ -0,0 +1,256 @@
...
+            await self._websocket.send(json.dumps(setup_msg))
+            ready_msg = await self._websocket.recv()
+            ready_msg = json.loads(ready_msg)
+            if ready_msg["type"] == "error":
+                raise Exception(f"received error {ready_msg['message']}")
+            if ready_msg["type"] != "ready":
+                raise Exception(f"unexpected first message type {ready_msg['type']}")
+
+        except Exception as e:
+            logger.error(f"{self}: unable to connect to Gradium: {e}")
```
markbackman: Error handling updates here too:

```suggestion
            await self.push_error(error_msg=f"Unknown error occurred: {e}", exception=e)
```

src/pipecat/services/gradium/stt.py
```diff
@@ -0,0 +1,256 @@
...
+
+        await self._disconnect_websocket()
+
+    async def _disconnect_websocket(self):
+        try:
+            if self._websocket and self._websocket.state is State.OPEN:
+                logger.debug("Disconnecting from Gradium STT")
+                await self._websocket.close()
+        except Exception as e:
+            logger.error(f"{self} error closing websocket: {e}")
```
markbackman:
```suggestion
            await self.push_error(error_msg=f"Unknown error occurred: {e}", exception=e)
```

src/pipecat/services/gradium/stt.py
```diff
@@ -0,0 +1,256 @@
...
+            await self._connect_websocket()
+
+    async def _process_response(self, msg):
+        type_ = msg.get("type", "")
+        if type_ == "text":
+            await self._handle_text(msg["text"])
+        elif type_ == "end_of_stream":
+            await self._handle_end_of_stream()
+        elif type_ == "error":
+            logger.error(f"Gradium error: {msg.get('message', 'Unknown error')}")
```
markbackman:
```suggestion
            await self.push_error(error_msg=f"Error: {msg}")
```

src/pipecat/services/gradium/tts.py
LaurentMazare: I'll clean it up, `voice` was the old way of doing things, now we only recommend `voice_id` (and won't document `voice` anymore).

src/pipecat/services/gradium/stt.py
LaurentMazare: The idea there is that once the client has finished your asr session, they send an EOS message to the service. The service sends back an EOS message after the audio has been fully processed (as there might be a bit of audio in flight). I think it indeed wouldn't matter to pipecat use cases so happy to remove it all along.

LaurentMazare:
Thanks for the detailed review @markbackman (and for your help before that).
I think all the changes in this repo should be in place now:
- Added Gradium to the README.
- Added `GRADIUM_API_KEY` to the env vars.
- Added a changelog entry.
- Added your example `07af-interruptible-gradium.py`, thanks for putting it up, it worked flawlessly on the first try!

src/pipecat/services/gradium/stt.py
markbackman: Yes, I think we should just remove then.

Instead, maybe just add a debug log when the event is received.

markbackman (approved):
LGTM! Just the one remaining action to related to the end of stream event, then this will be ready to merge.

src/pipecat/services/gradium/stt.py
LaurentMazare: Done, thanks again for the review comments!

PR #3174, opened 2025-12-03
feat(cartesia): make pause_frame_processing configurable in CartesiaTTSService

jamsea:
## Summary

Add `pause_frame_processing` parameter to `CartesiaTTSService` constructor to allow users to disable frame processing pausing during TTS generation.

## Motivation

This is useful for use cases where:
- Text aggregation is disabled (`aggregate_sentences=False`)
- Users want to send text directly to the TTS service without pausing incoming frame processing
- More control over frame flow is needed for custom pipelines

## Changes

- Added `pause_frame_processing` parameter to `CartesiaTTSService.__init__()` with default value `True` (preserving existing behavior)
- Updated docstring to document the new parameter
- Passes the parameter through to the parent `AudioContextWordTTSService` class

## Usage

```python
tts = CartesiaTTSService(
    voice_id="your_voice_id",
    aggregate_sentences=False,
    pause_frame_processing=False,  # New parameter
)
```

### DIFF
--- a/src/pipecat/services/cartesia/tts.py
+++ b/src/pipecat/services/cartesia/tts.py
@@ -234,6 +234,7 @@ class CartesiaTTSService(AudioContextWordTTSService):
         params: Optional[InputParams] = None,
         text_aggregator: Optional[BaseTextAggregator] = None,
         aggregate_sentences: Optional[bool] = True,
+        pause_frame_processing: Optional[bool] = True,
         **kwargs,
     ):
         """Initialize the Cartesia TTS service.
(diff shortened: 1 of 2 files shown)


src/pipecat/services/cartesia/tts.py
```diff
@@ -269,7 +271,7 @@ def __init__(
         super().__init__(
             aggregate_sentences=aggregate_sentences,
             push_text_frames=False,
-            pause_frame_processing=True,
+            pause_frame_processing=pause_frame_processing,
```
aconchillo, line 274: I don't think we want to add a constructor argument, we simply want to set this to `False`.

markbackman:
@aconchillo is going to revisit the concept of pausing frame processing. For now, I think we should close this until that investigation is sorted out.

PR #3175, opened 2025-12-03
Additional functionality related to thinking, for Google and Anthropic LLMs.

kompfner:
## Summary

Adds additional functionality related to "thinking", for Google and Anthropic LLMs.

Thinking, sometimes called "extended thinking" or "reasoning", is an LLM process where the model takes some additional time before giving an answer. It's useful for complex tasks that may require some level of planning and structured, step-by-step reasoning. The model can output its thoughts (or thought summaries, depending on the model) in addition to the answer. The thoughts are usually pretty granular and not really suitable for being spoken out loud in a conversation, but can be useful for logging or prompt debugging.

See [the CHANGELOG entry](https://github.com/pipecat-ai/pipecat/pull/3175/files#diff-418402e8e3d98ee4e11e4f5771e5509376fac9d12629e2e991273b6d787d4be7) for a description of what's in this PR.

## Remaining work

- [x] Add support for Gemini 3 Pro, which can return thought signatures in content parts other than function calls.
- [x] Investigate Anthropic "interleaved" (between-function-call) thought not getting written into context (or getting written into a too-late assistant message in the context)
  - Confirmed that this was due to function call frames being `SystemFrame`s and thus being delivered to the assistant context aggregator _before_ the preceding thought frames. Also confirmed that @aconchillo's work in [this PR](https://github.com/pipecat-ai/pipecat/pull/3189) resolves the issue 👍 
- [x] Fix commit history
- [x] Address any remaining code TODOs
- [x] Remove any lingering debug logging
- [x] Add CHANGELOG entry
- [x] Sanity-check different models
  - My testing has been focused on the default Pipecat Google and Anthropic models, Gemini 2.5 Flash and Claude Sonnet 4.5, respectively
  - Claude 3.7 specifically provides full thoughts instead of thought summaries, so that'll be interesting to test
- [x] Test thinking in a scenario with simultaneous function calls (Gemini will associate the thought signature with the first in a group)
  - Works! But notably this is pretty tricky to test—the model's behavior is nondeterministic, and it seems to prefer making function calls sequentially when it's asked to think and explain its thoughts
- [ ] Support changing LLM thinking configuration "on the fly" during the course of a conversation, which is important for anyone looking to build something like a "think harder" toggle
  - There is a restriction (or maybe a best practice) around when such a change can take place—it should only be at the start of a conceptual turn (i.e. not in the middle of a chain of tool calls, even though the LLM is invoked with each tool call result).
  - In the interest of this PR not getting too involved and take too long, support for this could come in a follow-up PR.

## Notes to reviewers

I'd recommend first running
```bash
uv run python examples/foundational/49-thinking-functions.py -t daily -d --llm google # or anthropic
```
and watching the console output to get a sense of the new functionality.

I'm fairly confident that we're appropriately passing the thought signatures to Gemini (when using Gemini 2.5 series models) because without them thoughts stop after the first function call, but with them thoughts are appropriately interleaved between function calls.

Interleaved thinking/function calling is also working with Claude.

### DIFF
--- a/examples/foundational/49a-thinking-anthropic.py
+++ b/examples/foundational/49a-thinking-anthropic.py
@@ -118,7 +118,7 @@ async def run_bot(transport: BaseTransport, runner_args: RunnerArguments):
                 "content": "Say hello briefly.",
             }
         )
-        # Here are some example example prompts conducive to demonstrating
+        # Here are some example prompts conducive to demonstrating
         # thinking (picked from Google and Anthropic docs).
         # messages.append(
         #     {
--- a/examples/foundational/49b-thinking-google.py
+++ b/examples/foundational/49b-thinking-google.py
@@ -123,7 +123,7 @@ async def run_bot(transport: BaseTransport, runner_args: RunnerArguments):
                 "content": "Say hello briefly.",
             }
         )
-        # Here are some example example prompts conducive to demonstrating
+        # Here are some example prompts conducive to demonstrating
         # thinking (picked from Google and Anthropic docs).
         # messages.append(
         #     {


examples/foundational/49c-thinking-functions-anthropic.py
```diff
@@ -0,0 +1,222 @@
...
+            }
+        )
+        await task.queue_frames([LLMRunFrame()])
+
+    @transport.event_handler("on_client_disconnected")
+    async def on_client_disconnected(transport, client):
+        logger.info(f"Client disconnected")
+        await task.cancel()
+
+    @transcript.event_handler("on_transcript_update")
```
kompfner, line 163: I wasn't sure whether it made more sense to piggyback on the existing "on_transcript_update" event—which can now produce two kinds of messages, `TranscriptionMessage` and `ThoughtTranscriptionMessage`—or to introduce a new event like "on_thought_transcript_update" that treats thoughts conceptually as a side transcript and not the main transcript.

Thoughts? (Joke intended)

src/pipecat/services/google/llm.py
```diff
@@ -918,9 +963,17 @@ async def _process_context(self, context: OpenAILLMContext | LLMContext):
...
                     if candidate.content and candidate.content.parts:
                         for part in candidate.content.parts:
-                            if not part.thought and part.text:
-                                search_result += part.text
-                                await self.push_frame(LLMTextFrame(part.text))
+                            if part.text:
+                                if part.thought:
+                                    # Gemini emits fully-formed thoughts rather
+                                    # than chunks so bracket each thought in
+                                    # start/end
```
kompfner, line 972: Note that there may be multiple fully-formed thoughts before the assistant response, so by doing this (bracketing each thought in start/end) we may actually end up with a transcript (as generated by the `TranscriptProcessor`) looking like:
```
user: ...
thought: ...
thought: ...
thought: ...
assistant: ...
```
In my opinion, that's totally fine, and might even be helpful for debugging. But it does deviate from the pattern of one entry in the transcript per "speaker".

kompfner:
A thought. I'm considering punting on this task

- [x] Add support for Gemini 3 Pro, which can return thought signatures in content parts other than function calls.

to a follow-up PR to keep this one from ballooning too much. In my testing, Gemini 3 Pro doesn't seem to be particularly well-suited for real-time conversations due to high processing times, so it's probably OK if proper support is a fast follow.

**UPDATE**: ended up doing the task.

kompfner:
Marking as ready for review. Mostly just CHANGELOG and some more testing work left, but don't foresee major changes. Some additional stuff can be punted to a follow-up PR, if need be.

kompfner:
From convo with @aconchillo:

- Add `role: Literal["assistant"] = "assistant"` to `ThoughtTranscriptionMessage`
- Get rid of `TranscriptProcessor.thought()`, moving logic into assistant
- Add flag so that we can do `TranscriptProcessor(process_thoughts=True)` (default `False`)

**UPDATE**: done.

kompfner:
Cont'd
- Make it so caller (Google LLM) fully constructs `FunctionInProgressFrame.append_context_message` with something like `{ "type": "fn_call_thought_signature", "signature": ... }` rather than assistant aggregator wrapping its contents in a "tool_call_extra" and constructing its own LLM specific message

**UPDATE**: done

kompfner:
Cont'd:
- Get rid of `LLMThoughtSignatureFrame`
- Use `LLMMessagesAppendFrame` instead, with `LLMSpecificMessageFrame` containing something like `{
                    "type": "non_fn_call_thought_signature", "signature": ...
                }`

**UPDATE**: done

kompfner:
Noticing some unexpected behavior now that we should account for:
- I'm seeing some non-function thought signatures with Gemini 2.5, which the Google docs suggest aren't even a thing...they only seem to be appearing if there are function calls involved in the conversation.
- Non-function thought signatures aren't necessarily appearing at the end of _every_ assistant response...so we need to figure out how to only associate them with the relevant responses. A tad frustratingly, the not-every-response behavior is happening only for non-Gemini-3 (Gemini 2.5 Flash), where I wasn't even expecting these signatures.

**UPDATE**: done

examples/foundational/49a-thinking-anthropic.py
```diff
@@ -0,0 +1,161 @@
...
+    async def on_client_connected(transport, client):
+        logger.info(f"Client connected")
+        # Kick off the conversation.
+        messages.append(
+            {
+                "role": "user",
+                "content": "Say hello briefly.",
+            }
+        )
+        # Here are some example example prompts conducive to demonstrating
```
aconchillo: example example

examples/foundational/49b-thinking-google.py
```diff
@@ -0,0 +1,166 @@
...
+    async def on_client_connected(transport, client):
+        logger.info(f"Client connected")
+        # Kick off the conversation.
+        messages.append(
+            {
+                "role": "user",
+                "content": "Say hello briefly.",
+            }
+        )
+        # Here are some example example prompts conducive to demonstrating
```
aconchillo: example example

aconchillo:
LGTM! The only thing is all that logic needed for Gemini, but that's something that we can always update in the future if we find a better way.

PR #3181, opened 2025-12-04
move pipecat.sync to pipecat.utils.sync

aconchillo:

Something I've been wanting to do for a while before 1.0.


markbackman (approved):
Makes sense!

PR #3183, opened 2025-12-04
Smart Turn analyzer now uses the full context of the turn rather than just the audio since VAD last triggered (fixes #3094)

### LINKED ISSUE #3094: Why does Smart Turn Analyzer Have USE_ONLY_LAST_VAD_SEGMENT=True
### pipecat version

0.0.91

### Python version

3.12.2

### Operating System

macOS 26.0.1

### Question

I am confused as to why `USE_ONLY_LAST_VAD_SEGMENT` is set to `True` in pipecat.audio.turn.smart_turn.base_smart_turn.

More specifically I am confused as to why this is the case and non configurable for the `LocalSmartTurnAnalyzerV3`.

From what I can see the only affect this has is to enforce the audio buffer to clear when the VAD goes to quiet even if the turn analyzer determined the turn to not be over.

In the smart turn repository it specifically says

```
Smart Turn takes 16kHz PCM audio as input. Up to 8 seconds of audio is supported, and we recommend providing the full audio of the user's current turn.
```

and

```
If additional speech is detected from the user before Smart Turn has finished executing, re-run Smart Turn on the entire turn recording, including the new audio, rather than just the new segment. Smart Turn works best when given sufficient context, and is not designed to run on very short audio segments.
```

This seems to be in direct contradiction to this advice. Could someone explain what I am missing?

### What I've tried

I have looked through the documentation and not found any mention.

In the code it has the comment

```
# not exposing this for now yet until the model can handle it.
# use_only_last_vad_segment: bool = USE_ONLY_LAST_VAD_SEGMENT
```

But again the documentation suggests that smart turn v3 model is meant to work with it set to False.

### Context

_No response_

marcus-daily:
`USE_ONLY_LAST_VAD_SEGMENT` seems to be a historical mistake -- Smart Turn is designed to always use the full audio from the turn.

Fixes https://github.com/pipecat-ai/pipecat/issues/3094

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -40,6 +40,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Updated Smart Turn model weights to v3.1.
 
+- Smart Turn analyzer now uses the full context of the turn rather than just the
+  audio since VAD last triggered.
+
 ### Deprecated
 
 - Package `pipecat.sync` is deprecated, use `pipecat.utils.sync` instead.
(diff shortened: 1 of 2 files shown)


marcus-daily:
Thanks @markbackman!

PR #3184, opened 2025-12-04
Added new languages support for AsyncAI

ashotbagh:
Added new languages for AsyncAI in `AsyncAITTSService` and `AsyncAIHttpTTSService`.
New languages:  `pt`, `nl`, `ar`, `ru`, `ro`, `ja`, `he`, `hy`, `tr`, `hi`, `zh`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,8 +8,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Added
-- Added new languages for AsyncAI in `AsyncAITTSService` and `AsyncAIHttpTTSService`.
+- Added new languages to AsyncAI in `AsyncAITTSService` and `AsyncAIHttpTTSService`.
   New `languages`:  `pt`, `nl`, `ar`, `ru`, `ro`, `ja`, `he`, `hy`, `tr`, `hi`, `zh`.
+  
+- Updated the default AsyncAI TTS model to asyncflow_multilingual_v1.0 for improved accuracy and broader language coverage.
 
 - Added `wait_for_all` argument to the base `LLMService`. When enabled, this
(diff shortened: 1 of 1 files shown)


src/pipecat/services/asyncai/tts.py
```diff
@@ -74,7 +85,7 @@ class InputParams(BaseModel):
             language: Language to use for synthesis.
         """
 
-        language: Optional[Language] = Language.EN
+        language: Optional[Language] = None
```
markbackman, line 88: Is the goal to pass no language at all? If so, `self._settings["language"]` is still set to `"en"` when no language is provided. Is that expected?

Also, does the new model detect language, but benefit from a specific language being provided?

CHANGELOG.md
```diff
@@ -8,6 +8,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Added
+- Added new languages for AsyncAI in `AsyncAITTSService` and `AsyncAIHttpTTSService`.
```
markbackman: You'll also want to mention that the default model is updated.

src/pipecat/services/asyncai/tts.py
ashotbagh: I’ve updated the code so that no language value is passed at all when the user does not specify one. Previously, self._settings["language"] defaulted to "en", which wasn’t the intended behavior.

Regarding the model behavior:
In Async, if no language is provided, the multilingual model automatically detects the language.
If a language is provided, then that language is explicitly forced during TTS generation.

Thanks for the note!

CHANGELOG.md
ashotbagh: I’ve updated the changelog to include the default model change as well.

markbackman (approved):
LGTM! Thanks for the quick update 🙇

PR #3185, opened 2025-12-04
fix(websocket): handle CancelledError to prevent reconnection on shutdown

jamsea:
## Problem

The `_receive_task_handler` in `WebsocketService` was catching `asyncio.CancelledError` in its generic `except Exception` block. This caused websocket-based services (like `ElevenLabsTTSService`) to attempt reconnection when they should have been shutting down cleanly.

### Symptoms
- When `EndFrame` was received, the TTS service would log: `error receiving messages: no close frame received or sent`
- The service would then attempt to reconnect up to 3 times with exponential backoff
- This delayed pipeline shutdown and caused sessions to hang

### Why This Works Most of the Time

There's a race condition in the shutdown sequence. When `_disconnect()` is called:

1. `cancel_task(self._receive_task)` cancels the receive task
2. `_disconnect_websocket()` closes the websocket

**When it works:** If the websocket closes cleanly before the cancellation propagates, the `websockets` library raises `ConnectionClosedOK`, which is already handled correctly:

```python
except ConnectionClosedOK as e:
    logger.debug(f"{self} connection closed normally: {e}")
    break  # Clean exit
```

**When it fails (the bug):** If the task is cancelled while the websocket is still open (e.g., mid-stream, network latency, server slow to acknowledge close), a different exception is raised (`CancelledError` or a wrapped exception like "no close frame received or sent"). This falls through to `except Exception`, triggering reconnection attempts.

### Root Cause

When the pipeline receives an `EndFrame`:
1. `stop()` is called on the TTS service
2. `_disconnect()` calls `cancel_task(self._receive_task)`
3. The receive task gets cancelled, raising `asyncio.CancelledError`
4. **Bug**: This was caught by `except Exception as e`, which logged it as an error and triggered reconnection logic when `reconnect_on_error=True`

## Solution

Treat `CancelledError` the same as `ConnectionClosedOK` - both represent intentional/normal closure:

```python
except (asyncio.CancelledError, ConnectionClosedOK):
    # Normal closure or task cancelled (e.g., during shutdown), exit cleanly
    logger.debug(f"{self} receive task ended")
    break
```

This eliminates the race condition by ensuring both shutdown paths lead to a clean exit.

## Testing

- Verified that `EndFrame` now properly terminates websocket services without reconnection attempts
- Normal reconnection behavior on actual connection errors is preserved

### DIFF
--- a/src/pipecat/services/websocket_service.py
+++ b/src/pipecat/services/websocket_service.py
@@ -12,7 +12,7 @@ from typing import Awaitable, Callable, Optional
 
 import websockets
 from loguru import logger
-from websockets.exceptions import ConnectionClosedOK
+from websockets.exceptions import ConnectionClosedError, ConnectionClosedOK
 from websockets.protocol import State
 
 from pipecat.frames.frames import ErrorFrame
@@ -136,6 +136,10 @@ class WebsocketService(ABC):
             except ConnectionClosedOK as e:
                 # Normal closure, don't retry
                 logger.debug(f"{self} connection closed normally: {e}")
+                break
+            except ConnectionClosedError as e:
+                # Error closure, don't retry
+                logger.warning(f"{self} connection closed, but with an error: {e}")
                 break
             except Exception as e:
                 message = f"{self} error receiving messages: {e}"
--- a/changelog/3185.fixed.md
+++ b/changelog/3185.fixed.md
@@ -1,0 +1,2 @@
+- Fixed an issue that caused `WebsocketService` instances to attempt
+  reconnection during shutdown.


src/pipecat/services/websocket_service.py
```diff
@@ -133,6 +133,10 @@ async def _receive_task_handler(self, report_error: Callable[[ErrorFrame], Await
         while True:
             try:
                 await self._receive_messages()
+            except asyncio.CancelledError:
```
jamsea: @aconchillo perhaps we should do the same thing as line 140? Is there any case where we'd want to retry if the task was explicitly cancelled?

aconchillo:
> ## Problem
> The `_receive_task_handler` in `WebsocketService` was catching `asyncio.CancelledError` in its generic `except Exception` block. This caused websocket-based services (like `ElevenLabsTTSService`) to attempt reconnection when they should have been shutting down cleanly.
> 

Thank you @jamsea ! Good catch! There's certainly an issue but it's not related to `asyncio.CancelledError`. `asyncio.CancelledError` is not a subclass of `Exception`, it's a `BaseException` so it would not be caught in the `Exception` block.

The problem is that a `ConnectionClosedError` is triggered which is the one causing `no close frame received or sent`, so we need to capture that instead.

I took the liberty to update this PR with a different commit.

PR #3188, opened 2025-12-04
Auto-generate changelog from fragments

markbackman:

I'm proposing a new changelog process to eliminate the constant merge conflicts that we hit. In this process, we'll create changelog fragments in the format of `<PR_number>.<type>.md`. We'll use [towncrier](https://towncrier.readthedocs.io/) to build our changelog by running a github workflow. We'll run this workflow at release time.

You can view a preview of the changelog by running:
```
uv sync --group dev
uv run towncrier build --draft --version Unreleased
```

Note: the workflow takes two inputs
- release tag
- release date

The workflow creates a PR that:
- generates the changelog
- removes the processed fragments

Let me know what you think!

### DIFF
--- a/changelog/3120.added.md
+++ b/changelog/3120.added.md
@@ -1,1 +1,0 @@
-- Added `wait_for_all` argument to the base `LLMService`. When enabled, this ensures all function calls complete before returning results to the LLM (i.e., before running a new inference with those results).


markbackman:
The created Changelog doesn't include links to PRs, but we could include that as well if we think it's helpful. This would be automated, of course. I left it out for now since we don't include them currently.

.github/workflows/generate-changelog.yml
```diff
@@ -0,0 +1,126 @@
+name: Generate Changelog for Release
+
+on:
+  workflow_dispatch:
+    inputs:
+      version:
+        description: "Release version (e.g., 0.0.97)"
```
kompfner, line 7: I suppose we could make determining the release version automatic, but that could introduce a bunch of complexity and require us to reason about edge cases. So I agree with the decision to just have it manual for now

kompfner: Ditto for release date.

markbackman: Yeah, these are things we manually add now, so it's not a big deal to do when we run the action. At least that was my thinking.

I think we'll always run this right before releasing, so maybe we just automate the date generation?

.github/workflows/generate-changelog.yml
```diff
@@ -0,0 +1,126 @@
...
+          else
+            echo "✓ Found $FRAGMENT_COUNT changelog fragment(s)"
+            echo "has_fragments=true" >> $GITHUB_OUTPUT
+          fi
+
+      - name: Preview changelog
+        run: |
+          echo "## Preview of changelog for version ${{ inputs.version }}"
+          echo ""
+          uv run towncrier build --draft --version "${{ inputs.version }}" --date "${{ inputs.date }}"
```
kompfner: Gotta hand it to the Python community for making a dedicated tool for everything under the sun

.github/workflows/generate-changelog.yml
markbackman: Ok, now the date will generate automatically with an optional override:
https://github.com/pipecat-ai/pipecat/pull/3188/commits/6d1d17a48d36a05d5b872cabe77d9e52af21d035

.github/workflows/generate-changelog.yml
```diff
@@ -0,0 +1,126 @@
...
+
+          if [ "$FRAGMENT_COUNT" -eq "0" ]; then
+            echo "⚠️ Warning: No changelog fragments found"
+            echo "has_fragments=false" >> $GITHUB_OUTPUT
+          else
+            echo "✓ Found $FRAGMENT_COUNT changelog fragment(s)"
+            echo "has_fragments=true" >> $GITHUB_OUTPUT
+          fi
+
+      - name: Preview changelog
```
kompfner, line 121: What does the preview step get us?

.github/workflows/generate-changelog.yml
```diff
@@ -0,0 +1,126 @@
...
+          fi
+
+      - name: Check for changelog fragments
+        id: check_fragments
+        run: |
+          FRAGMENT_COUNT=$(find changelog -name "*.md" ! -name "README.md" ! -name "_template.md.j2" | wc -l | tr -d ' ')
+          echo "fragment_count=$FRAGMENT_COUNT" >> $GITHUB_OUTPUT
+
+          if [ "$FRAGMENT_COUNT" -eq "0" ]; then
+            echo "⚠️ Warning: No changelog fragments found"
```
kompfner: Is there then a way to skip the subsequent steps?

.github/workflows/generate-changelog.yml
```diff
@@ -0,0 +1,126 @@
...
+          # Validate date format
+          if ! date -d "${{ inputs.date }}" >/dev/null 2>&1; then
+            # Try macOS date format
+            if ! date -j -f "%Y-%m-%d" "${{ inputs.date }}" >/dev/null 2>&1; then
+              echo "Error: Date must be in YYYY-MM-DD format (e.g., 2025-12-04)"
+              exit 1
+            fi
+          fi
+
+      - name: Check for changelog fragments
```
kompfner, line 74: Do we want to also validate that the fragment types ("added", "changed", etc) are in the set we expect?

.github/workflows/generate-changelog.yml
markbackman: This will add a preview output in the github workflow logs. Maybe it's not terribly useful. You can create a preview manually as well.

It also doesn't hurt.

.github/workflows/generate-changelog.yml
kompfner: Or, more broadly, that they follow the required naming conventions?

.github/workflows/generate-changelog.yml
markbackman: Maybe we should just fail the workflow in this case? WDYT?

.github/workflows/generate-changelog.yml
kompfner: Actually, that validation probably makes sense in our pre-commit check as well.

kompfner: Oh, is the towncrier tool strict? (i.e. does it validate its inputs and output errors?) Is that the purpose of the "preview" step, below?

If so—then we could leverage the towncrier tool for the pre-commit check too.

.github/workflows/generate-changelog.yml
markbackman: Ok, just made this change. I think failing is the right thing. That means no PR is created and you'll get an error message warning about no fragments being found.

.github/workflows/generate-changelog.yml
markbackman: Good idea. Yes, let's validate the fragment types match our towncrier config.

.github/workflows/generate-changelog.yml
```diff
@@ -0,0 +1,144 @@
...
+
+      - name: Preview changelog
+        run: |
+          echo "## Preview of changelog for version ${{ inputs.version }}"
+          echo ""
+          uv run towncrier build --draft --version "${{ inputs.version }}" --date "${{ steps.set_date.outputs.release_date }}"
+
+      - name: Build changelog
+        run: |
+          uv run towncrier build --version "${{ inputs.version }}" --date "${{ steps.set_date.outputs.release_date }}" --yes
```
kompfner, line 129: Is this the step also responsible for deleting the fragments?

.github/workflows/generate-changelog.yml
markbackman: I thought about including checks for changelog entries in the pre-commit (or a new workflow) but I opted to not include it. There will be, at least on occasion, PRs that don't have changelog entries.

WDYT?

changelog/README.md
kompfner, line 1: Having a README here (or at least a detailed one) in addition to what you've added to CONTRIBUTING.md feels like overkill (and just creates an opportunity for some info to get out of sync or out of date). CONTRIBUTING.md feels like the right place for it.

.github/workflows/generate-changelog.yml
markbackman: Yes, my understanding is that `towncrier build` deletes the fragments.

changelog/README.md
markbackman: 👍

.github/workflows/generate-changelog.yml
kompfner: I was saying I agreed with your decision to keep these manual. But great! Happy for there to be an auto-date feature :)

.github/workflows/generate-changelog.yml
kompfner: 👍

kompfner: I guess it would be better for the workflow to finish without failing (since there's no failure, just no work to do) but I'm not familiar enough with GH workflows to say if there's a way of doing that off the top of my head.

kompfner: Failure seems fine! ™️

.github/workflows/generate-changelog.yml
kompfner: > let's validate the fragment types match our towncrier config

hopefully towncrier has an option/mode to automagically do that work for us 🤞

kompfner: > I thought about including checks for changelog entries in the pre-commit (or a new workflow) but I opted to not include it. There will be, at least on occasion, PRs that don't have changelog entries.
> 
> WDYT?

Agree with you to **not** require changelog entries for every PR (and certainly not for every commit)

changelog/3120.added.md
kompfner, line 1: This can now be removed.

.github/workflows/generate-changelog.yml
markbackman: I think failure is a good idea because it's a clear sign that something was not right. Then, you'll go and look at the logs to see that no fragments existed. That's better than creating a PR with no changes which you have to close.

changelog/3120.added.md
markbackman: Removed!

PR #3189, opened 2025-12-04
introduce uninterruptible frames

aconchillo:

In this PR we introduce the `UninterruptibleFrame` mixin. This mixin is used to mark certain data and control frames as uninterruptible. Sometimes we need to use ordered frames (data and control) but we don't want those frames to be discarded or cancelled if there's an interruption, this is what this new frame type is useful for.

```
@dataclass
class TestFrame(DataFrame, UninterruptibleFrame):
   data: str
```

We also change the behavior of `FunctionCallInProgressFrame` and `FunctionCallResultFrame`, bot have now changed from system frames to control and data frames, respectively. Both are set to `UninterruptibleFrame`.

### DIFF
--- a/src/pipecat/frames/frames.py
+++ b/src/pipecat/frames/frames.py
@@ -711,6 +711,44 @@ class LLMConfigureOutputFrame(DataFrame):
 
 
 @dataclass
+class FunctionCallResultProperties:
+    """Properties for configuring function call result behavior.
+
+    Parameters:
+        run_llm: Whether to run the LLM after receiving this result.
+        on_context_updated: Callback to execute when context is updated.
+    """
+
+    run_llm: Optional[bool] = None
+    on_context_updated: Optional[Callable[[], Awaitable[None]]] = None
+
+
+@dataclass
+class FunctionCallResultFrame(DataFrame, UninterruptibleFrame):
+    """Frame containing the result of an LLM function call.
+
+    This is an uninterruptible frame because once a result is generated we
+    always want to update the context.
+
+    Parameters:
+        function_name: Name of the function that was executed.
+        tool_call_id: Unique identifier for the function call.
+        arguments: Arguments that were passed to the function.
+        result: The result returned by the function.
+        run_llm: Whether to run the LLM after this result.
+        properties: Additional properties for result handling.
+
+    """
+
+    function_name: str
+    tool_call_id: str
+    arguments: Any
+    result: Any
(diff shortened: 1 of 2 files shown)


src/pipecat/frames/frames.py
filipi87, line 1: Shouldn't we also change `FunctionCallsStartedFrame` and `FunctionCallCancelFrame`? Otherwise, it feels like a race condition could happen.

kompfner:
Tested the changes in this PR against the problem scenario described in #3175 and it resolves it!

src/pipecat/frames/frames.py
aconchillo: In the case of `FunctionCallsStartedFrame` we might want to notify right away that function calls are about to start . For `FunctionCallCancelFrame` it doesn't really matter. These two don't modify the context and don't trigger any inference. So, I didn't see a reason to change them.

filipi87: Yeah, I was wondering if we might have a race condition, but I think we’re okay. 👍

src/pipecat/processors/frame_processor.py
```diff
@@ -902,10 +907,26 @@ def __reset_process_task(self):
...
+            item = self.__process_queue.get_nowait()
+            if isinstance(item, UninterruptibleFrame):
+                new_queue.put_nowait(item)
             self.__process_queue.task_done()
 
+        # Put back UninterruptibleFrame frames into our process queue.
+        while not new_queue.empty():
+            item = new_queue.get_nowait()
+            self.__process_queue.put_nowait(item)
+            new_queue.task_done()
```
filipi87, line 928: Wouldn’t it be enough to simply do `self.__process_queue = new_queue` here ?

src/pipecat/processors/frame_processor.py
```diff
@@ -805,8 +809,12 @@ async def _start_interruption(self):
                 # interruption). Instead we just drain the queue because this is
                 # an interruption.
                 self.__reset_process_task()
+            elif isinstance(self.__process_current_frame, UninterruptibleFrame):
```
filipi87, line 812: Makes sense. 👍

src/pipecat/processors/frame_processor.py
aconchillo: No, because the process task is using the old queue still and we are not cancelling the task.

PR #3190, opened 2025-12-05
LLMService: let's not introduce wait_for_all for now

aconchillo:

Let's not introduce `wait_for_all` for now. We can always do it if we really have a use case for it.


markbackman (approved):
LGTM. Makes sense to hold off.

PR #3194, opened 2025-12-05
feat: add Gemini client identification.

### LINKED ISSUE #3172: Client identification for Gemini API
### Problem Statement

Hey team, would it be possible to add the following client identification to the library to help us track usage origins better? 

### Proposed Solution

`X-Goog-Api-Client`: API client identification. The value is a space-separated list of NAME "/" SEMVER strings, where the NAME should only contain lowercase letters, digits, and "-", and the SEMVER should be a semantic version string. For example: X-Goog-Api-Client: python/3.5.0 grpc-google-pubsub-v1/0.1.0-beta2 linux/2.7.0.

See system parameter docs here: https://docs.cloud.google.com/apis/docs/system-parameters 

```
Example: 
## Python:

from google import genai

client = genai.Client(
    api_key="...",
    http_options={
        "headers": {
            "x-goog-api-client": "my-sdk-py/1.23",
        }
    }
)

## JavaScript
import { GoogleGenAI } from '@google/genai';

const ai = new GoogleGenAI({
    apiKey: '...',
    httpOptions: {
        headers: {
            'x-goog-api-client': 'my-sdk-js/1.23'
        }
    }
});

## Direct API
curl https://generativelanguage.googleapis.com/v1beta/models/gemini-2.5-flash:generateContent?key=$GOOGLE_API_KEY \
    -H 'Content-Type: application/json'
    -H 'X-Goog-Api-Client: my-sdk/1.23'
    -d '{...}'
```

You can see an example implementation of that client identification in the SDK here: https://github.com/google/adk-python/blob/main/src/google/adk/models/google_llm.py#L262-L272

### Alternative Solutions

_No response_

### Additional Context

_No response_

- [x] No, I'm just suggesting

thorwebdev:
Add Gemini API client identification headers. Closes #3172

### DIFF
--- a/tests/test_google_utils.py
+++ b/tests/test_google_utils.py
@@ -1,0 +1,60 @@
+import importlib.util
+import sys
+import unittest
+from unittest.mock import MagicMock
+
+# Mock Pipecat package
+sys.modules["pipecat"] = MagicMock()
+sys.modules["pipecat"].__version__ = "0.0.0-test"
+
+# Load the module directly from source
+spec = importlib.util.spec_from_file_location(
+    "pipecat.services.google.utils", "src/pipecat/services/google/utils.py"
+)
+utils_module = importlib.util.module_from_spec(spec)
+sys.modules["pipecat.services.google.utils"] = utils_module
+spec.loader.exec_module(utils_module)
+
+update_google_client_http_options = utils_module.update_google_client_http_options
+pipecat_version = "0.0.0-test"
+
+
+class TestGoogleUtils(unittest.TestCase):
+    def test_update_google_client_http_options_none(self):
+        options = update_google_client_http_options(None)
+        self.assertEqual(options, {"headers": {"x-goog-api-client": f"pipecat/{pipecat_version}"}})
+
+    def test_update_google_client_http_options_dict_empty(self):
+        options = update_google_client_http_options({})
+        self.assertEqual(options, {"headers": {"x-goog-api-client": f"pipecat/{pipecat_version}"}})
+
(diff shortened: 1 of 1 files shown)


src/pipecat/services/google/gemini_live/llm.py
```diff
@@ -682,6 +683,21 @@ def __init__(
...
+        elif isinstance(self._http_options, dict):
+            if "headers" in self._http_options:
+                self._http_options["headers"].update(client_header)
+            else:
+                self._http_options["headers"] = client_header
+        elif hasattr(self._http_options, "headers"):
+            if self._http_options.headers is None:
+                self._http_options.headers = client_header
+            else:
+                self._http_options.headers.update(client_header)
```
aconchillo: Since it's always the same code, it'd be awesome if we could add a `services/google/utils.py` file and just do:

```
        self._http_options = update_google_client_http_options(http_options)
```

or something like that.

thorwebdev: Good, call, updated. PTAL

aconchillo: Nice! Some formatting issues. We have some pre-commit hooks, or simply run `ruff`.

```
uv run pre-commit install
```

Not sure about the failing tests... :thinking:

thorwebdev: Done ✅

aconchillo:
Superseded by #3208. I took all the commits and fixed the UT. Mocking the whole pipecat module was causing issues. I ended up adding a `version()` function so it's easy to mock.

PR #3195, opened 2025-12-05
Add tracking headers to Hume service

ivaaan:
This adds a tracking header to all requests made to the Hume service, which helps Hume with Support and Analytics by clearly identifying the requests made via Pipecat.

### DIFF
--- a/changelog/3195.changed.md
+++ b/changelog/3195.changed.md
@@ -1,0 +1,3 @@
+- Added tracking headers (`X-Hume-Client-Name` and `X-Hume-Client-Version`) to all requests made by `HumeTTSService` to the Hume API for better usage tracking and analytics.
+- Added `stop()` and `cancel()` cleanup methods to `HumeTTSService` to properly close the HTTP client and prevent resource leaks.
+


src/pipecat/services/hume/tts.py
```diff
@@ -104,7 +108,11 @@ def __init__(
             **kwargs,
         )
 
-        self._client = AsyncHumeClient(api_key=api_key)
+        # Create a custom httpx.AsyncClient with tracking headers
+        # Headers are included in all requests made by the Hume SDK
+        custom_http_client = httpx.AsyncClient(headers=DEFAULT_HEADERS)
```
markbackman: Do we need to close this client in `stop()` or `cancel()`?

ivaaan: Yes! Added, please review

ivaaan:
> LGTM!
> 
> Care to add a changelog? If so, we now use changelog fragments, which you can add to `changelog/`. Format is <pr_number>..md, so 3195.changed.md for this one.

Sure, added!

Issue #3191, opened 2025-12-05
User Interruption not working properly


hardiktulsiani1110:
### pipecat version
0.0.91

### Python version
3.11

### Operating System
macOS 15.6.1

### Issue description
Upon interrupting the bot, the bot still keeps on speaking. It either doesn't work or takes very long time for interruption to be detected.

### Reproduction steps

**1) Configuration Overview**
- Strategy: MinWordsInterruptionStrategy(min_words=2)
- Serializer: TwilioFrameSerializer
- STT: DeepgramSTTService
- LLM: GroqLLMService (or OpenAILLMService, GoogleLLMService)
- TTS: CartesiaTTSService (or ElevenLabsTTSService)

**2) Transport config**
```python
# Create Twilio frame serializer
serializer = TwilioFrameSerializer(
    stream_sid=call_data['stream_id'],
    call_sid=call_data['call_id'],
    account_sid=configs['telephony_config']['credentials']['account_sid'],
    auth_token=configs['telephony_config']['credentials']['auth_token'],
)

# Create FastAPI WebSocket transport
transport = FastAPIWebsocketTransport(
    websocket=websocket,
    params=FastAPIWebsocketParams(
        audio_in_enabled=True,
        audio_out_enabled=True,
        audio_in_passthrough=True,
        add_wav_header=False,
        vad_analyzer=SileroVADAnalyzer(
            params=VADParams(
                confidence=0.7, 
                start_secs=0.15,
                stop_secs=0.8,
                min_volume=0.6,
            ),
        ),
        serializer=serializer,
    ),
)
```

**3) Pipeline config**
```python
context = LLMContext(messages)
context_aggregator = LLMContextAggregatorPair(context)

# Create pipeline
pipeline = Pipeline(
    [
        transport.input(),  # Audio input from Twilio
        stt,  # Speech-to-Text
        context_aggregator.user(),  # Add user message to context
        llm,  # LLM processing
        tts,  # Text-to-Speech
        transport.output(),  # Audio output to Twilio
        context_aggregator.assistant(),  # Add assistant response to context
    ]
)

task = PipelineTask(
    pipeline,
    params=PipelineParams(
        audio_in_sample_rate=8000,
        audio_out_sample_rate=8000,
        enable_metrics=True,
        allow_interruptions=True,
        interruption_strategies=[MinWordsInterruptionStrategy(min_words=2)],
    ),
    idle_timeout_secs=8,
)

# Run pipeline
[...]

Issue #2922, opened 2025-10-27
Interrupting a bot with MinWordsInterruptionStrategy specified causes the context to be incorrectly ordered


markbackman:
### pipecat version

0.0.90

### Python version

Any

### Operating System

Any

### Issue description

When specifying a Pipeline with `MinWordsInterruptionStrategy`, interrupting the bot results in the user and assistant messages being incorrectly ordered in the context.

### Reproduction steps

1. Run foundational example 42: https://github.com/pipecat-ai/pipecat/blob/main/examples/foundational/42-interruption-config.py
2. Interrupt the bot on the first turn
3. Inspect the context to verify the messages aren't user/assistant alternating.

### Expected behavior

User/assistant messages alternate

### Actual behavior

User and assistant messages don't alternate when an interruption occurs.

### Logs

```shell

```

omChauhanDev:
@markbackman i tested this & it looks like this is already fixed by PR #3045.

Side note for anyone coming across this later: `MinWordsInterruptionStrategy` is deprecated. Please use `MinWordsUserTurnStartStrategy` instead, via the new `PipelineTask.turn_start_strategies` parameter.

markbackman:
Indeed! This is fixed with the new user turn strategies which will be released next week. This is already available on `main`, so closing.

Issue #2785, opened 2025-10-04
Using `InterimTranscription` to trigger Interruptions


lshaun:
### Problem Statement

I have enabled emulated interruptions on my bot however these interruptions are triggered too late.
Often at the end of user speech.

### Proposed Solution

It seems such interruptions are triggered only on transcripts.
If we trigger these interruptions based on interim transcripts. This would make the bot more responsive in terms of handling such interruptions.

Currently `InterimTranscriptionFrame` is not being used much but we can instead use this to trigger `EmulatedUserSpeakingFrame`.

Although I'd like to know if any other interactions that I'm not seeing here.

### Would you be willing to help implement this feature?

- [x] Yes, I'd like to contribute
- [ ] No, I'm just suggesting

lshaun:
@markbackman @kompfner 
Does this seem valid ?

kompfner:
Thanks for raising this concern!

Today the `EmulateUserStartedSpeakingFrame` mechanism is really more of a backup system for the (hopefully relatively rare) cases that your “first-line” voice activity detector (VAD) system doesn’t pick up the start of a user utterance. [Check out our examples for usage of ￼`SileroVADAnalyzer`￼](https://github.com/pipecat-ai/pipecat/blob/fce6f55ddb31f0b0ec66c32e88efe18c4e2a2a65/examples/foundational/07-interruptible.py#L40)—this locally-running VAD should be able to detect the start of user speech pretty quickly and reliably.

If there are reasons you’d prefer to not configure a VAD in your pipeline, though, we’d be interested to hear about them!

(Another note: given that `EmulateUserStartedSpeakingFrame` is a backup mechanism, we opted to base it on a more reliable signal—a transcription—as opposed to one prone to false positives—an interim transcription).

poseneror:
I think this is resolved in v0.0.99 - seems that intermediate transcriptions are now supported as user turn start strategy
[https://docs.pipecat.ai/server/utilities/user-turn-strategies#transcriptionuserturnstartstrategy](https://docs.pipecat.ai/server/utilities/user-turn-strategies#transcriptionuserturnstartstrategy)

markbackman:
We did add support for interims but in closer testing, we're seeing issues. You can follow in this PR:
https://github.com/pipecat-ai/pipecat/pull/3456

aconchillo:
[...]

PR #3204, opened 2025-12-06
feat: add interim transcription interruptions for lower latency

divyanshjain-git:
- Add enable_interim_transcription_interruptions param to LLMUserAggregatorParams
- Evaluate interruption strategies on interim transcriptions when enabled
- Process InterruptionFrame synchronously to fix context ordering race condition
- Add guard against re-entry during interruption processing

This reduces interruption latency by triggering interruptions as soon as the threshold is met on interim transcriptions, without waiting for the final transcription. Default is True for lower latency.

#3191 #2922 #2785


markbackman:
Thanks for sharing this solution. We're in the process of refactoring the interruption strategies, which will improve how all of this works.
https://github.com/pipecat-ai/pipecat/pull/3045

I'm going to close this PR now since it will be superseded by the PR I linked.

PR #3209, opened 2025-12-09
Update system prompt in Gemini example to be more instructive

jamsea:
Changed the on_client_connected system message from a direct greeting to an instruction that tells the AI to introduce itself, giving the LLM more flexibility in how it starts the conversation.

### DIFF
--- a/examples/foundational/07n-interruptible-gemini.py
+++ b/examples/foundational/07n-interruptible-gemini.py
@@ -136,7 +136,7 @@ async def run_bot(transport: BaseTransport, runner_args: RunnerArguments):
         messages.append(
             {
                 "role": "system",
-                "content": "Hello! I'm your AI assistant. I can help you with a variety of tasks. What would you like to know?",
+                "content": "You are an AI assistant. You can help with a variety of tasks. Introduce yourself and ask the user what they would like to know.",
             }
         )
         await task.queue_frames([LLMRunFrame()])


examples/foundational/07n-interruptible-gemini.py
```diff
@@ -136,7 +136,7 @@ async def on_client_connected(transport, client):
         messages.append(
             {
                 "role": "system",
-                "content": "Hello! I'm your AI assistant. I can help you with a variety of tasks. What would you like to know?",
+                "content": "You are an AI assistant. You can help with a variety of tasks. Introduce yourself and ask the user what they would like to know.",
```
markbackman, line 139: This will be converted to a `user` message, so perhaps something like this?

```suggestion
                "content": "Greet the user by saying, 'Hello! I'm your AI assistant. I can help you with a variety of tasks. What would you like to know?'",
```

kompfner: I'm personally more into @jamsea's wording, which reads much more like the kind of system instruction we have everywhere else in our foundational examples. The developer shouldn't have to necessarily be that aware of the fact that under the hood Gemini system instructions will be converted to user messages—if it's a reasonably well-crafted system instruction it should just work™️

markbackman: Fine by me.

Wording aside, the interesting part is that we're saying that you have to know the transformation of the message to write the prompt correctly. That just feels off. I haven't seen the issue that Vanessa is reporting, but I can imagine how it would happen. Perhaps it's unavoidable, as prompting will LLM provider and model specific...

markbackman (approved):

PR #3212, opened 2025-12-09
Nova 2 Sonic support

kompfner:
For another PR: cross-modal input (switching between voice and text input).

### DIFF
--- a/examples/foundational/40-aws-nova-sonic.py
+++ b/examples/foundational/40-aws-nova-sonic.py
@@ -44,7 +44,7 @@ async def fetch_weather_from_api(params: FunctionCallParams):
     # You can continue chatting while waiting for this to complete.
     # With Nova 2 Sonic (the default model), the assistant will respond
     # appropriately once the function call is complete.
-    await asyncio.sleep(10)
+    await asyncio.sleep(5)
     await params.result_callback(
         {
             "conditions": "nice",


examples/foundational/40-aws-nova-sonic.py
```diff
@@ -33,7 +35,16 @@
...
-    temperature = 75 if params.arguments["format"] == "fahrenheit" else 24
+    temperature = (
+        random.randint(60, 85)
+        if params.arguments["format"] == "fahrenheit"
+        else random.randint(15, 30)
+    )
+    # Simulate a long network delay.
+    # You can continue chatting while waiting for this to complete.
+    # With Nova 2 Sonic (the default model), the assistant will respond
+    # appropriately once the function call is complete.
```
kompfner, line 46: I _think_ I'm witnessing a bit of buggy behavior in the model.

When there are multiple in-flight tool calls, the model seems prone to getting confused and "mixing up" the results when they come back.

For example, I see things like this:
```
User: "Hey, can you tell me the weather in San Diego?"

--> Tool call kicks off with ID "foo"

User: "Actually, hang on. I'm in Washington, D.C."

--> Tool call kicks of with ID "bar"

--> Tool result arrives with ID "foo" (corresponding to San Diego) and is reported to the model.

Assistant: "The weather in Washington, D.C. is nice, with a temperature of 65 degrees."

--> Tool result arrives with ID "bar" (corresponding to Washington, D.C.) and is reported to the model.

Assistant: "Actually in Washington, D.C. the temperature is 70 degrees."
```

src/pipecat/services/aws/nova_sonic/llm.py
```diff
@@ -591,11 +618,33 @@ def _create_client(self) -> BedrockRuntimeClient:
...
+        # Nova Sonic (the older model) is identified by "amazon.nova-sonic-v1:0"
+        return self._model == "amazon.nova-sonic-v1:0"
+
+    def _is_endpointing_sensitivity_supported(self) -> bool:
+        # endpointing_sensitivity is only supported with Nova 2 Sonic (and,
+        # presumably, future models)
+        return not self._is_first_generation_sonic_model()
+
+    def _is_assistant_response_trigger_needed(self) -> bool:
+        # Assistant response trigger audio is only needed with the older model
```
kompfner, line 631: Very glad about this 😄

src/pipecat/services/aws/nova_sonic/llm.py
```diff
@@ -192,8 +201,8 @@ def __init__(
         access_key_id: str,
         session_token: Optional[str] = None,
         region: str,
-        model: str = "amazon.nova-sonic-v1:0",
-        voice_id: str = "matthew",  # matthew, tiffany, amy
+        model: str = "amazon.nova-2-sonic-v1:0",
```
kompfner, line 204: One thing I've noticed in testing this new model is that some guardrails are _very_ strong. Some of the things I usually do during testing are occasionally treated as prohibited.

For example, a few times when I asked the model to tell me a short story it told me it didn't want to accidentally infringe on intellectual property so it couldn't fulfill my request.

Another time, when I asked for suggestions of things to do for fun while in Washington, D.C. it said it didn't want to answer to avoid suggesting criminal or violent activities.

In both cases, once the guardrail was "triggered", it became very hard to continue the conversation without completely changing topics.

In the short-story scenario, for example, a follow-up request for a 3-sentence story was also blocked. The model then suggested "safer" topics, like general story arcs or story-writing tips. Each time I said "OK, let's discuss that", it still said it couldn't discuss those topics, for the same reason (not infringing on intellectual property).

In the fun-things-to-do-around-town scenario, I followed up with things like "what about, like, going to museums?" and it still said it couldn't answer, for the same reason (it didn't want to suggest something harmful).

changelog/3212.added.md
```diff
@@ -0,0 +1,6 @@
+- Added to `AWSNovaSonicLLMService` functionality related to the new (and now
+  default) Nova 2 Sonic model (`"amazon.nova-2-sonic-v1:0"`):
+  - Added the `endpointing_sensitivity` parameter to control how quickly the
+    model decides the user has stopped speaking.
+  - Made the assistant-response-trigger hack a no-op. It's only needed for the
+    older Nova Sonic model.
```
filipi87, line 6: Cool. So they have fixed this. 🙌🎉

examples/foundational/40-aws-nova-sonic.py
filipi87: Interesting. In theory, the `tool_call_id` should be enough to handle this.
Have you tried invoking different functions at the same time to check whether it works correctly in that case?

src/pipecat/services/aws/nova_sonic/llm.py
filipi87: Yeah, it looks like there are too many guardrails in this case now.

filipi87 (approved):
LGTM. 🚀

examples/foundational/40-aws-nova-sonic.py
kompfner: Hm, if I fire all of the tool calls at the exact same time (e.g. "tell me the weather in San Diego, Washington, D.C., and New York"), `tool_call_id` _does_ seem to be enough to tell the model which is which...the problem seems to arise only when they're offset...

kompfner: Ah, another clue! Maybe it has something to do with whether you interact with the model while the function calls are in-flight...so, going back to the previous example ("tell me the weather in San Diego, Washington, D.C., and New York"), if I then ask the model to tell me a fun fact while I'm waiting for the weather results, when the weather results arrive they might get scrambled (the model might think San Diego's weather is New York's, etc)

kompfner: Another thing I'm running into: if there are multiple in-flight tool calls at the same time, then when they come back, it triggers in the model another unnecessary duplicate tool call.

And one last thing: if I hammer the model too hard with interrupting in-flight tool calls with new tool calls, it occasionally crashes with an "System instability detected" error.

kompfner: But this is all somewhat aggressive testing. In the "normal" case of one tool call at a time (probably most real-world usage falls under this umbrella), things seem to be working reasonably well.

filipi87: Yeah, I think so. And in this case, since they are handling the context, I’m not sure there’s much we can do to try to prevent it.

PR #3214, opened 2025-12-09
Update run_inference to use the provided LLM configuration params

markbackman:

The LLMs previously used the default values. With this change, they'll use the InputParams values specified at initialization time. This more closely aligns with expectations.

We could add another step to provide custom params, but since this is a convenience, this seems like a good first change.

### DIFF
--- a/src/pipecat/services/anthropic/llm.py
+++ b/src/pipecat/services/anthropic/llm.py
@@ -303,7 +303,7 @@ class AnthropicLLMService(LLMService):
         # LLM completion
         response = await self._client.beta.messages.create(**params)
 
-        return response.content[0].text
+        return next((block.text for block in response.content if hasattr(block, "text")), None)
 
     def create_context_aggregator(
         self,


src/pipecat/services/anthropic/llm.py
```diff
@@ -238,26 +238,38 @@ async def run_inference(self, context: LLMContext | OpenAILLMContext) -> Optiona
...
+            messages = invocation_params["messages"]
+            system = invocation_params["system"]
+            tools = invocation_params["tools"]
         else:
             context = AnthropicLLMContext.upgrade_to_anthropic(context)
             messages = context.messages
             system = getattr(context, "system", NOT_GIVEN)
+            tools = context.tools or []
+
+        # Build params using the same method as streaming completions
```
kompfner, line 285: Since this PR was opened, we have new params:
- "thinking"
- "betas"

src/pipecat/services/google/llm.py
```diff
@@ -779,6 +787,42 @@ async def run_inference(self, context: LLMContext | OpenAILLMContext) -> Optiona
...
+        Args:
+            system_instruction: Optional system instruction to use.
+            tools: Optional list of tools to include.
+            tool_config: Optional tool configuration.
+
+        Returns:
+            Dictionary of generation parameters with None values filtered out.
+        """
+        # Filter out None values
+        generation_params = {
```
kompfner, line 853: Since this PR was opened, we have a new "thinking" param.

kompfner (approved):
Looks good!

Just missing the new thinking parameters from Anthropic and Google.

In the future, we really ought to make it so both the one-off `run_inference` and the "normal" inference code paths share a function in each LLM service to construct the inference parameters. Otherwise, we're just going to keep having the problem of the the two pathways diverging.

src/pipecat/services/google/llm.py
markbackman: I added thinking, which needs to be handled as a Pydantic model here and as a dict in _update settings.

src/pipecat/services/anthropic/llm.py
markbackman: @kompfner I'm not familiar with how these work. Mind pushing a commit to update those?

kompfner: Sure thing

kompfner: Done

kompfner:
Hm, testing Pipecat Flows summary generation (which leverages `run_inference`), and it's failing when extended thinking is enabled with Anthropic. Investigating.

UPDATE: think I found the problem. Testing a fix.

UPDATE: fixed.

src/pipecat/services/anthropic/llm.py
```diff
@@ -303,7 +303,7 @@ async def run_inference(self, context: LLMContext | OpenAILLMContext) -> Optiona
         # LLM completion
         response = await self._client.beta.messages.create(**params)
 
-        return response.content[0].text
+        return next((block.text for block in response.content if hasattr(block, "text")), None)
```
kompfner, line 306: When extended thinking is enabled, `content[0]` doesn't necessarily contain `text`. It may be the thinking block.

PR #3215, opened 2025-12-09
user-bot-latency log observer internal var change

vipyne:

I am creating a custom observer that inherits from multiple observers like so:
```
class MyCustomObserver(TurnTrackingObserver, UserBotLatencyLogObserver):
```
Both [`TurnTrackingObserver`](https://github.com/pipecat-ai/pipecat/blob/main/src/pipecat/observers/turn_tracking_observer.py#L66) and [`UserBotLatencyLogObserver`](https://github.com/pipecat-ai/pipecat/blob/main/src/pipecat/observers/loggers/user_bot_latency_log_observer.py#L39) use an internal `self._processed_frames` variable. The TurnTrackingObserver's `self._processed_frames` clobbers the UserBotLatencyLogObserver's version. We should scope the variable name to avoid such collisions.

### DIFF
--- a/src/pipecat/observers/loggers/user_bot_latency_log_observer.py
+++ b/src/pipecat/observers/loggers/user_bot_latency_log_observer.py
@@ -36,7 +36,7 @@ class UserBotLatencyLogObserver(BaseObserver):
         to calculate response latencies.
         """
         super().__init__()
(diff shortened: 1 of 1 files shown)


markbackman (approved):
LGTM

PR #3219, opened 2025-12-10
deprecate: FalSmartTurnAnalyzer and LocalSmartTurnAnalyzer

markbackman:

I'm not sure which model is maintained by Fal, but I think it has fallen behind the latest. Also, running locally on CPU with `LocalSmartTurnAnalyzerV3` is now recommended. Given that, I think a deprecation is the right path here.

@marcus-daily does it make sense to deprecate any of the older local smart turn versions?


markbackman:
> Makes sense! Yes I think it would be fine to deprecate the original LocalSmartTurnAnalyzer.

Sounds good. I've added the LocalSmartTurnAnalyzer deprecation to this PR as well.

PR #3224, opened 2025-12-12
Clean up logic related to applying Gemini thought signatures to conte…

kompfner:
…xt messages.

Also, fix Gemini image generation (see CHANGELOG entries for description).


src/pipecat/services/google/llm.py
```diff
@@ -1007,49 +998,59 @@ async def _process_context(self, context: OpenAILLMContext | LLMContext):
...
+                                # can later be sure we've put it in the right
+                                # place in context when sending the context
+                                # back to the LLM to continue the conversation.
                                 bookmark = {}
-                                if part.inline_data and part.inline_data.data:
-                                    bookmark["inline_data"] = {"inline_data": part.inline_data}
+                                if part.function_call:
+                                    bookmark["function_call"] = function_call_id
+                                elif part.inline_data and part.inline_data.data:
+                                    # NOTE: missing feature: we don't store
```
kompfner: Recording model-generated images in context is a feature we'll need to support if we want Gemini 3 Pro image generation to work properly.

Currently, the `07n-interruptible-gemini-image` example works because the model always also generates some text after generating an image. If you switch the model to "gemini-3-pro-image-preview", though, you'll see that the example stops working after the first time the model generates an image, since it doesn't also generate text. The response containing the image is never recorded to context, so the model doesn't know it's already done the work. Every future turn of the conversation just causes the model to generate a new image.

kompfner: Happy to tackle this next.

kompfner: In tackling this work, I realized that this inline_data bookmark logic isn't quite right. Working support for Gemini 3 Pro Image into this PR. Stay tuned.

kompfner: Fixed it in [the latest commit](https://github.com/pipecat-ai/pipecat/pull/3224/commits/7454a4eac1233146e85bb1bae58427f91037ec61) in this PR.

But it feels a bit brittle, esp. w/r/t assuming images will be JPGs. Going to continue tinkering with it.

kompfner: Made it a little better...maybe it can be further improved in a separate PR. It works quite well for Gemini Pro 3 Image right now, though.

This fix also makes it so Gemini 2.5 Flash Image _can_ iterate on an image (i.e. "can you change the color palette, can you move this element over here, etc."—prior to this fix, any images it generated wouldn't be written to context so each generated image would be totally unrelated to the previous.

The problem is that since our universal context uses JPEG-encoded images as the universal format, and Gemini 2.5 Flash Image doesn't give us JPEGs, we end up encoding those images further and further with each turn, causing a subtle degradation. This isn't ideal. But hey—image context wasn't working at all before, so this change is an upgrade.

In another PR we should do the deeper fix and address the underlying problem:
- We should store images in context in the format they're given to us, along with a flag saying what format that is
- When reading them out of the context to run the LLM, we can convert to the format we need, as needed (and only _if_ needed, which may be rare?)
- For user images captured from camera, maybe we can keep storing them as JPEG as we do today

But I would lobby for going forward with the change in this PR to not bite off too much in one PR.

kompfner: Actually working on making this change this morning. Stay tuned.

kompfner: OK I feel significantly better about the state of things now.

src/pipecat/services/google/llm.py
```diff
@@ -975,81 +981,81 @@ async def _process_context(self, context: OpenAILLMContext | LLMContext):
...
+                                    logger.warning("Thought signature found on unhandled Part type")
+                                if bookmark:
+                                    await self.push_frame(
+                                        LLMMessagesAppendFrame(
+                                            [
+                                                self.get_llm_adapter().create_llm_specific_message(
+                                                    {
+                                                        "type": "thought_signature",
+                                                        "signature": part.thought_signature,
+                                                        "bookmark": bookmark,
```
aconchillo, line 1053: Nice! Much better than `append_extra_context_messages`!

aconchillo:
LGTM!

PR #3227, opened 2025-12-12
Add Ultravox service

### LINKED ISSUE #2245: Ultravox API service report
### Problem Statement

Pipecat have already add  ultravox sevice as local deployment.
But we find that, the local deployment can not support tool registry and context management, also lack of some audio process method compare to the api sevice. 
1. can pipecat support  API service for ultravox.
2. can pipecat support function tool registry and context managemnet. 

### Proposed Solution

Here is the repo: 
- https://github.com/fixie-ai/ultravox/tree/main/ultravox

1. I think we may implement like open AI context for ultravox.
2. the ultravox API doc https://docs.ultravox.ai/api-reference/introduction

### Alternative Solutions

_No response_

### Additional Context

_No response_

- [ ] No, I'm just suggesting

mdepinet:
Adds support for using Ultravox Realtime as a speech-to-speech service. (Fixes #2245.)

Also removes the deprecated Ultravox speech-to-text vllm model integration to avoid confusion.

### DIFF
--- a/examples/foundational/49-ultravox-realtime.py
+++ b/examples/foundational/49-ultravox-realtime.py
@@ -36,17 +36,14 @@ transport_params = {
     "daily": lambda: DailyParams(
         audio_in_enabled=True,
         audio_out_enabled=True,
-        vad_enabled=False,
     ),
     "twilio": lambda: FastAPIWebsocketParams(
         audio_in_enabled=True,
         audio_out_enabled=True,
-        vad_enabled=False,
     ),
     "webrtc": lambda: TransportParams(
         audio_in_enabled=True,
         audio_out_enabled=True,
-        vad_enabled=False,
     ),
 }
 
--- a/src/pipecat/services/ultravox/__init__.py
+++ b/src/pipecat/services/ultravox/__init__.py
@@ -1,1 +1,0 @@
-from .llm import UltravoxRealtimeLLMService
--- a/src/pipecat/services/ultravox/llm.py
+++ b/src/pipecat/services/ultravox/llm.py
@@ -26,8 +26,6 @@ from pipecat.adapters.schemas.tools_schema import ToolsSchema
 from pipecat.audio.utils import create_stream_resampler
 from pipecat.frames.frames import (
     AggregationType,
-    BotStartedSpeakingFrame,
-    BotStoppedSpeakingFrame,
     CancelFrame,
     EndFrame,
     Frame,
@@ -44,6 +42,7 @@ from pipecat.frames.frames import (
     TTSStartedFrame,
     TTSStoppedFrame,
     TTSTextFrame,
+    UserAudioRawFrame,
 )
 from pipecat.processors.aggregators.llm_context import LLMContext
 from pipecat.processors.aggregators.llm_response import (
@@ -115,7 +114,7 @@ class OneShotInputParams(BaseModel):
     api_key: str
     system_prompt: Optional[str] = None
     temperature: float = Field(default=0.0, ge=0.0, le=1.0)
-    model: str = "fixie-ai/ultravox"
+    model: Optional[str] = None
     voice: Optional[uuid.UUID] = None
     metadata: Dict[str, str] = Field(default_factory=dict)
     max_duration: datetime.timedelta = Field(
@@ -177,6 +176,7 @@ class UltravoxRealtimeLLMService(LLMService):
         self._receive_task: Optional[asyncio.Task] = None
         self._disconnecting = False
         self._bot_responding: Literal[None, "text", "voice"] = None
+        self._last_user_id: Optional[str] = None
 
         self._sample_rate = 48000
         self._resampler = create_stream_resampler()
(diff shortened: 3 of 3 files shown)


src/pipecat/services/ultravox/__init__.py
```diff
@@ -1,13 +1 @@
...
-#
-
-import sys
-
-from pipecat.services import DeprecatedModuleProxy
-
-from .stt import *
-
-sys.modules[__name__] = DeprecatedModuleProxy(globals(), "ultravox", "ultravox.stt")
+from .llm import UltravoxRealtimeLLMService
```
aconchillo: I think it is safe to remove everything from this file (including the new line), we usually force it to indicate the full package name.

examples/foundational/49-ultravox-realtime.py
```diff
@@ -0,0 +1,224 @@
...
+
+
+# We store functions so objects (e.g. SileroVADAnalyzer) don't get
+# instantiated. The function will be called when the desired transport gets
+# selected.
+transport_params = {
+    "daily": lambda: DailyParams(
+        audio_in_enabled=True,
+        audio_out_enabled=True,
+        vad_enabled=False,
```
aconchillo: `vad_enabled` is deprecated and shouldn't be used.

src/pipecat/services/ultravox/llm.py
```diff
@@ -0,0 +1,534 @@
...
+                } | self._params.extra
+                async with aiohttp.ClientSession() as session:
+                    async with session.post(
+                        f"https://api.ultravox.ai/api/agents/{self._params.agent_id}/calls",
+                        headers={"X-Api-Key": self._params.api_key},
+                        json=request_body,
+                    ) as response:
+                        if response.status != 201:
+                            error_text = await response.text()
+                            raise Exception(f"Ultravox API error {response.status}: {error_text}")
```
aconchillo: Use `push_error()` instead and pass the exception instead (and return)

src/pipecat/services/ultravox/llm.py
```diff
@@ -0,0 +1,534 @@
...
+                } | self._params.extra
+                async with aiohttp.ClientSession() as session:
+                    async with session.post(
+                        "https://api.ultravox.ai/api/calls",
+                        headers={"X-Api-Key": self._params.api_key},
+                        json=request_body,
+                    ) as response:
+                        if response.status != 201:
+                            error_text = await response.text()
+                            raise Exception(f"Ultravox API error {response.status}: {error_text}")
```
aconchillo: Use `push_error()` instead and pass the exception instead.

src/pipecat/services/ultravox/llm.py
```diff
@@ -0,0 +1,534 @@
...
+                    "templateContext": self._params.template_context,
+                    "metadata": self._params.metadata,
+                    "maxDuration": f"{self._params.max_duration.total_seconds():3f}s",
+                    "medium": {
+                        "serverWebSocket": {
+                            "inputSampleRate": self._sample_rate,
+                        }
+                    },
+                } | self._params.extra
+                async with aiohttp.ClientSession() as session:
```
aconchillo: we probably should add a try/except to catch any issues when connecting (and `push_error()`).

maybe it would be simpler to have a few functions to make it simpler and add the try/except in each one

```
        match self._params:
            case JoinUrlInputParams():
                join_url = self._start_join_url()
            case AgentInputParams():
                join_url = self._start_agent()

     if join_url:
        logger.info(f"Joining Ultravox Realtime call via URL: {join_url}")
        self._socket = await websocket_client.connect(join_url)
        self._receive_task = self.create_task(self._receive_messages())
```

src/pipecat/services/ultravox/llm.py
```diff
@@ -0,0 +1,534 @@
...
+                await self.push_error("Ultravox websocket receive error", e, fatal=True)
+
+    async def _handle_audio(self, audio: bytes):
+        """Handle incoming audio bytes from Ultravox Realtime."""
+        if not audio:
+            return
+        if not self._bot_responding:
+            await self.push_frame(BotStartedSpeakingFrame())
+            await self.push_frame(TTSStartedFrame())
+            await self.push_frame(LLMFullResponseStartFrame())
```
aconchillo, line 465: - No need to push `BotStartedSpeakngFrame` this is automatically push by the output transport when output audio frames get there.
- Push `LLMFullResponseStartFrame` before `TTSStartedFrame`. It should be fine, but usually that's what happens.

So
```
            await self.push_frame(TTSStartedFrame())
```

src/pipecat/services/ultravox/llm.py
```diff
@@ -0,0 +1,534 @@
...
+            self._bot_responding = "voice"
+        await self.push_frame(TTSAudioRawFrame(audio, self._sample_rate, 1))
+
+    async def _handle_response_end(self):
+        if self._bot_responding == "voice":
+            await self.push_frame(BotStoppedSpeakingFrame())
```
aconchillo: Same, don't push `BotStoppedSpeakingFrame`

src/pipecat/services/ultravox/llm.py
```diff
@@ -0,0 +1,534 @@
...
+                    arguments=parameters,
+                    context=None,
+                )
+            ]
+        )
+
+    async def _handle_user_transcript(self, text: str):
+        await self.push_frame(
+            TranscriptionFrame(
+                user_id="",
```
aconchillo: We could save the `user_id` in `_send_user_audio` (if available) and then use it here.

aconchillo: Not super important right now though.

src/pipecat/services/ultravox/llm.py
mdepinet: I moved the try/catch into start to make sure exceptions are handled consistently, but otherwise done. :)

aconchillo:
LGTM! Thank you for fixing the comments!

PR #3228, opened 2025-12-12
Update GeminiLiveLLMService model to gemini-2.5-flash-native-audio-pr…

markbackman:
…eview-12-2025

In addition, I fixed an unrelated unit test failure in test_transcript_processor.py's `TestThoughtTranscription` test. The issue appears to be a missing delay, which I've fixed by introducing a `SleepFrame()` between transcripts. cc @kompfner.

### DIFF
--- a/tests/test_transcript_processor.py
+++ b/tests/test_transcript_processor.py
@@ -730,9 +730,6 @@ class TestThoughtTranscription(unittest.IsolatedAsyncioTestCase):
         self.assertIsInstance(second_message, ThoughtTranscriptionMessage)
         self.assertEqual(second_message.content, "Second consideration")
 
-        # Verify timestamps are different
-        self.assertNotEqual(first_message.timestamp, second_message.timestamp)
-
     async def test_empty_thought_handling(self):
         """Test that empty thoughts are not emitted"""
         processor = AssistantTranscriptProcessor(process_thoughts=True)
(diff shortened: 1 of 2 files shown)


kompfner (approved):

tests/test_transcript_processor.py
```diff
@@ -696,6 +696,7 @@ async def handle_update(proc, frame: TranscriptionUpdateFrame):
             LLMThoughtStartFrame(),
             LLMThoughtTextFrame(text="First consideration"),
             LLMThoughtEndFrame(),
+            SleepFrame(),  # Delay to ensure different timestamps
```
kompfner: Oh hm just saw this...wut? Shouldn't these frames guaranteed to be processed in order without this?

kompfner:
Looks like `GeminiLiveVertexLLMService`'s default model, `"google/gemini-2.0-flash-live-preview-04-09"`, is still supported, so that's good 👍

But it might be worth considering whether we want to bump it to `"google/gemini-2.5-flash-native-audio"` for parity between the Pipecat Gemini API and Vertex API offerings.

tests/test_transcript_processor.py
kompfner: Let's get rid of the timestamps-are-different check instead. That check doesn't need to exist. It's valid for two thoughts to have the same timestamp.

PR #3229, opened 2025-12-12
Asterisk transport

NikolayShakin:
Asterisk websocket channel transport and serializer implementation. 
It allows us to connect [Asterisk WebSocket media channels](https://docs.asterisk.org/Configuration/Channel-Drivers/WebSocket/) to Pipecat. 

- The transport supports  codecs: `ulaw/alaw/slin/slin12/slin16...`
- The transport implements auto detection of codec/bit rate(for slin)/websocket sub protocol (`plain-text/json`)
- The transport implements an initial jitter buffer to address the audio interruptions at the very beginning of a dialog
- The transport implements flow control and uses a remote(on Asterisk) media buffer 
- The transport allows flushing the buffered media in Asterisk's media buffer (for immediate interruptions) 
- The transport implements media exchange and some basic `channel_websocket` signaling 

### Serializer:
The serializer is designed to handle both binary and text websocket messages. It detects channel parameters based on `MEDIA_START` event, and resamples audio data if needed. 

### Transport:
The transport implements a local audio buffer to provide flow control. The local audio buffer is effectively an `async.Queue` that stores properly sampled binary audio chunks.  
The flow control logic is:
1. Every audio packet goes to this buffer first
2. If we have any data in the local audio buffer, we send it until we fill out the remote audio buffer (on Asterisk) or run out of data in the local buffer. After that, we put sending on pause
3. We resume sending if we have any data in the local buffer and the remote buffer utilization drops below 50% until we fill out the remote buffer or run out of data in the local buffer

On the Asterisk side, configuration is very simple:
```
;;; /etc/asterisk/extensions.conf
[your_context]
exten = 999,1,Dial(WebSocket/pipecat/c(ulaw)f(json))
exten = 000,1,Dial(WebSocket/pipecat/c(slin16)f(json))


;;; /etc/asterisk/websocket_client.conf
[pipecat]
type = websocket_client
uri = ws://127.0.0.1:8765
protocols = media
connection_type = per_call_config
connection_timeout = 1000
reconnect_interval = 1000
reconnect_attempts = 5
tls_enabled = no
``` 

On pipecat's side:
```
...
from pipecat.transports.asterisk.chan_ws_server import (
    AsteriskWSServerParams,
    AsteriskWSServerTransport,
)

...
    transport = AsteriskWSServerTransport(
        params=AsteriskWSServerParams(
            audio_in_enabled=True,
            audio_out_enabled=True,
            serializer=AsteriskWsFrameSerializer(),
        ),
    )
...
```


temp144p (commented):
Great work on this! We've been running Asterisk + Pipecat in production and have some learnings that might help make this even more robust.

## 1. XOFF/XON Flow Control is Critical

The current implementation logs XOFF/XON but doesn't actually pause sending. Without this, audio **will** be dropped when the buffer fills (~900 frames high water mark). We solved this with an `asyncio.Event`:

```python
# In output transport
self._flow_ready = asyncio.Event()
self._flow_ready.set()

def set_flow_control(self, paused: bool):
    if paused:
        self._flow_ready.clear()  # Pause sending
    else:
        self._flow_ready.set()    # Resume sending

async def write_audio_frame(self, frame):
    # Wait if paused by XOFF
    if not self._flow_ready.is_set():
        await asyncio.wait_for(self._flow_ready.wait(), timeout=5.0)
    # ... then send audio
```

## 2. Generation Counter for Clean Interruptions

After `FLUSH_MEDIA`, old audio frames still in the pipeline continue to be sent, immediately refilling Asterisk's buffer. This causes XOFF/XON cycling and delays. We use a generation counter to drop stale frames:

```python
self._flush_generation = 0

async def _flush_asterisk_buffer(self):
    self._flush_generation += 1
    await self._send_command("FLUSH_MEDIA")

async def write_audio_frame(self, frame):
    my_generation = self._flush_generation
    
    # ... wait for flow control if needed ...
    
    # Drop stale audio from before the flush
    if my_generation != self._flush_generation:
        return False  # Skip this frame
    
    await self._websocket.send_bytes(audio)
```

## 3. 1x Real-Time Pacing

Even with `START_MEDIA_BUFFERING`, sending audio faster than real-time immediately fills the buffer after XON, causing XOFF/XON cycling. Pacing at 1x real-time prevents this:

```python
# 20ms chunks = 160 samples at 8kHz = 320 bytes PCM
self._send_interval_sec = 0.02  # 20ms

async def write_audio_frame(self, frame):
    now = time.monotonic()
    if self._next_send_time > now:
        await asyncio.sleep(self._next_send_time - now)
    
    await self._websocket.send_bytes(audio)
    self._next_send_time = time.monotonic() + self._send_interval_sec
```

## 4. Chunk Size Matching ptime

For cleanest audio, output chunks should match Asterisk's `ptime` (usually 20ms). At 8kHz that's 160 samples = 320 bytes PCM. Mismatched chunk sizes can cause subtle audio glitches.

## 5. Recommended Dialplan Addition

For best audio quality, we recommend documenting this jitterbuffer setting:

```ini
same => n,Set(JITTERBUFFER(adaptive)=60,300,40)
same => n,Dial(WebSocket/...)
```

Parameters: `target_extra=60ms`, `max_size=300ms`, `resync_threshold=40ms`

---

## Results

With these additions, we achieved:
- **~1.2-1.5s** end-to-end response latency
- **Clean interruptions** - no stale audio after user speaks
- **No XOFF/XON cycling** - stable buffer management
- **Smooth audio** - no glitches or dropouts

Tested with Asterisk 22.7, ulaw codec.

---

Happy to contribute these changes or discuss further! Let me know if you'd like me to open a follow-up PR with the flow control and pacing additions.

NikolayShakin:
@temp144p Thanks for your inputs. Actually, I expected to have the buffer overflow issue you mentioned, and that's why I've created a handler(`_handle_media_xon`) for MEDIA_XON/OFF events, but in my tests, I could not reach the high water mark, so I've never received `MEDIA_XOFF`. I used  deepgram STT->chat-gpt LLM->Catesia TTS with default settings.
Could you describe the scenario when you hit the buffer's  high-watter limit?

temp144p:
We hit XOFF consistently with ElevenLabs TTS in a couple scenarios:

**Long responses without pacing:** Our bot explains a security incident in one long turn (~15-20 seconds of audio). ElevenLabs streams audio faster than real-time, so without 1x pacing, we'd send ~20s of audio in ~5-8s, immediately hitting the high water mark.

**Background audio mixing:** We also have ambient background noise playing continuously during calls (mixed at 30% volume). This means audio is always being sent to Asterisk, not just during TTS - which makes hitting the buffer limit more likely.

**After interruption + FLUSH_MEDIA:** This was the trickier case. When the user interrupts, we send FLUSH_MEDIA, but audio frames already in Pipecat's pipeline continue sending. These "stale" frames immediately refill Asterisk's buffer, causing XOFF. Then buffer drains, XON fires, more stale frames send, XOFF again. This cycling caused 500ms-1000ms (or longer) delays before new audio actually played. The generation counter solved this - we drop any frame from before the flush.

**Your setup might naturally pace slower:** Cartesia might stream slower than ElevenLabs, or ChatGPT's token generation could be the bottleneck. If your TTS output rate ≤ 1x real-time and you're not mixing background audio, you might never hit XOFF.

In our case, we had issues with every response - both long and short - until we added proper pacing and flow control. Even short "How are you doing today?" responses would get delayed or glitchy without it.

NikolayShakin:
Thanks @temp144p, yep, that totally makes sense, if you send more than 20s of audio at once, you will definitely hit the limit. Addressing these issues would definitely require creating a new transport. And this is what I was trying to avoid indeed :). 
Here are my thoughts on the subject.
For a robust transport implementation, I would go this way:

1. When we receive `MEDIA_START` event in the output transport, we need to extract `optimal_frame_size` from the message and use this value as a size for the following websocket binary messages. Probably it can be done on TTS level, but it's better to have it in the transport to not rely on some external services. Ideally, we should receive properly framed audio chunks from TTS, but I think it's better to have re-framing mechanics within the transport.  Especially if we are creating an internal buffer anyway.

2. By default, always send `START_MEDIA_BUFFERING` and `REPORT_QUEUE_DRAINED` after we received  `MEDIA_START` in output transport (as I understood `set(JITTERBUFFER...)` on the channel does the same job, it buffers audio before sending it to Asterisk's core). After `REPORT_QUEUE_DRAINED` asterisk will notify us when/if the buffer is empty. 

3. Implement flow control and a jitter buffer to not rely on Asterisk's jitter buffer and `MEDIA_XOFF/ON` events. 
    3.1. First, we need to create a jitter buffer on Pipecat's side to handle inconsistent delays in the very beginning of speech generation before we fill up the asterisk's buffer. We need to have at least `N=200` ms of audio before we send it to Asterisk. Because if we send audio right away, we may have interruptions in case TTS generates an audio stream slower than real-time (a frame every 20ms/s for ptime 20) and asterisk's media buffer is drained. I guess it's a very plausible event. And the same buffer can be used for 3.2 implementation. 
 
    3.2. We need to keep track of how much audio we have already sent over real-time flow. It's kind of a high-water mark on pipcat's side. It can be implemented as a counter, every time we send some audio to the websocet we add audio duration to the counter, every `ptime` milliseconds we substitute `ptime` from the counter. And we stop sending audio as soon as we reach `M=5000ms` and start buffering them into the internal buffer of the transport.  With a reasonable value of `M` it would prevent us from overflowing asterisk's buffer. 
 
    3.3. Handling exceptions: 
        3.3.1. We received `MEDIA_XOFF` - something went wrong this 3.2, I don't think we need to do anything except show a respective error message. It indicates an implementation bug.  
        3.3.2. We receive `QUEUE_DRAINED` - it might not be an indication of an issue, e.g. user was speaking for a while and we didn't generate any audio. So it's not an error, but it's good to know. Effectively, it tells pipecat when asterisk stopped playing any audio to the user. This handler should create a log message and send another `REPORT_QUEUE_DRAINED`, so we can get another `QUEUE_DRAINED` event next time. Optionally, push a respective text frame to the pipeline. 
 
 5. `FLUSH` handler. I liked your idea with the flush generations, it's quite clever. But I think it can be implemented even in a simpler way,  we just drop the internal buffer of the transport and send `FLUSH` and then `REPORT_QUEUE_DRAINED` to asterisk instead. So there is nothing to playback the the user from the pipecat's or asterisk's buffers
 

 I would appreciate feedback

temp144p:
Great breakdown! A few thoughts:

**On pre-buffering (3.1):** The 200ms buffer makes sense for consistency. We didn't need it with ElevenLabs (streams faster than real-time), but it would help with slower TTS providers. The tradeoff is 200ms added latency on first speech - acceptable for most telephony use cases.

**On Pipecat-side high-water mark (3.2):** I like this approach better than reacting to XOFF. It's more predictable and self-contained. The 5000ms limit is conservative (Asterisk's default high-water is 800ms), which gives plenty of margin.

**On flush handling:** Our generation counter was solving a slightly different problem - frames that were already *in Pipecat's pipeline* (between TTS processor and our transport). With your internal buffer approach, those frames would land in the buffer and get dropped when you clear it, so the counter becomes unnecessary. Cleaner solution.

**On REPORT_QUEUE_DRAINED:** We didn't use this - good catch. It's useful for knowing when Asterisk actually finished playing audio (vs just when we finished sending). Could be valuable for analytics or end-of-utterance timing.

**One addition I'd suggest:** Handle the `optimal_frame_size` gracefully - if MEDIA_START doesn't include it (older Asterisk?), fall back to calculating from sample_rate * ptime * bytes_per_sample. In our testing with 22.x, we saw `ptime=20` in the message which gave us 320 bytes for 8kHz/16-bit.

NikolayShakin:
Internal jitter buffer size and internal high-water marks should be configurable; the numbers I mentioned are preliminary default values. And I need to check how the media buffer size is defined on the Asterisk side, because it might be limited by a number of bytes or a number of frames. We will need to follow the same logic in our buffer.
As for `optimal_frame_size`, it was added almost from the very beginning of the chan_websocket module. So it's safe to rely on it. However, it's a good idea to specify the minimal supported Asterisk version in the documentation.

NikolayShakin:
I started working on transport. I'm going to create a new PR when it's done. Too many changes have to be made; it's easier to start from scratch in a new branch

NikolayShakin:
I checked the media buffer implementation in asterisk code. The buffer size is limited by a number of frames (=1000). 
When asterisk receives binary audio data more than `optimal_frame_size`, it will reframe it to 20ms-long frames and store them into the buffer, saving leftovers if any. 
For pipecat this means that we don't need to reframe audio, can safely send arbitrary framed audio unless we overflow the buffer (1000x`optimal_frame_size` Bytes)

jcbjoe:
Hey @NikolayShakin Thanks for the contribution here! We have our own Asterisk transport. Something that I feel would be useful is the option to use the raw codec rather than decoding it. Most STT/TTS providers actually support ULAW so it doesnt make sense to convert to PCM/L16. This is how we use it and it works great. Also saves a little bit on CPU. I think defaulting to decoding would be fine then for people who want to use ULAW have that as an InputParam.

NikolayShakin:
Hi @jcbjoe, you are right, I didn't think about that. TBH, in my head, the default scenario was using `slin16` with Asterisk. The reasoning was that in Asterisk, it's relatively cheap to transcode `slin16` to something else, and Asterisk can transcode it to some HD codec e.g. g722 or regular g711[a|u] automatically, depending on the endpoint's codecs set.
I'll review this part. And I would appreciate it a lot if you could share some related code snippets from your implementation

jcbjoe:
> Hi @jcbjoe, you are right, I didn't think about that. TBH, in my head, the default scenario was using `slin16` with Asterisk. The reasoning was that in Asterisk, it's relatively cheap to transcode `slin16` to something else, and Asterisk can transcode it to some HD codec e.g. g722 or regular g711[a|u] automatically, depending on the endpoint's codecs set. I'll review this part. And I would appreciate it a lot if you could share some related code snippets from your implementation

For us, traffic hits asterisk from our carriers in ULAW(G711). Which is why Id like to persist ULAW through Pipecat, togglable of course. When handling tens of thousands of concurrent calls, every little bit of CPU helps! Plus when you set ULAW for the format for some STT providers they have dedicated telephony models.

Im going to try out your transport this week and if I have any suggestions ill be sure to post them

NikolayShakin:
@jcbjoe, Yep, you're absolutely right, in most cases we will have `a/ulaw` on the second leg of Asterisk, if a TTS provider can produce `u/alaw` it's the most straightforward case, no need to transcode

Mickaelh51:
Hi, I started to use Asterisk with pipecat, and I think use only slin16, because in my case, many TTS, STT providers use by default slin16 instead of telephony code as G711A or U.
And I prefer ask to Asterisk to transcode than pipecat though python and its GIL.
and I think than VAD and smart turn of pipecat prefer slin16.

I'm right or not for you ?

NikolayShakin:
Hi, @Mickaelh51, I agree that `slin16` is more versatile, and it's better to transcode in Asterisk's C code instead of Pipecat's Python code. I played with the slin16 on Pipecat and g722 on asterisk side and it gives nice HD-voice. However, when we deal with real PSTN, it's `g711a/u` in 99% of the cases, so for telephony applications, probably we will always have g711 in the second leg. And even if we use `slin16` in the websocket channel, effectively we only have 64Kbit/s of useful media data from the original `g711` stream re-packed to `slin16`, it doesn't improve voice quality or anything. 
So if we can use `g711` in TTS and STT providers, it looks like a reasonable optimization; we don't need to transcode the media at all.
But TBH I didn't check if it's feasible in pipecat yet

Mickaelh51:
Thanks for your answer.
My goal it's not to enhance quality.
Sure 99% will be g711. But for me if I use slin16, the flow from asterisk will be compatible with lot of TTS / STT providers.
And pipecat internal tools.

NikolayShakin:
Yep, @Mickaelh51, I agree, and all the flavors of `slin` are already supported along with a/u-law.
But currently, the serializer would transcode`g711` to `slin` before sending the audio to the pipeline.  
My point is that having "transparent-g711" options is a good idea that might be useful in many cases, and it seems to be relatively easy to implement. 
As I said, I didn't have a chance to dive deeper into the subject yet, but based on what @jcbjoe said, it should work.

Mickaelh51:
I don't have time to check your asterisk transport yet. But in mine, (very simple), it's asterisk that transcode codecs to slin16. And there is no transcoding in pipecat (it uses many CPU than asterisk, as you said).
and yes it's a good idea to add transprant processes (without transcoding), in this case, TTS or STT provider will transcode in his side (if needed).

Tanmay120703:
Will this work for Twilio where the codec is ulaw and the sample rate is 8000. I just want to know the configuration of extension.conf and also what changes might required?

NikolayShakin:
Hi @Tanmay120703, Asterisk is a provider-agnostic application, so yes, it will work with any SIP provider, including Twilio. However, asterisk configuration is out of the scope of this discussion

abalashov:
Beautiful contribution, thank you so much for solving this problem, @NikolayShakin!

However, unfortunately there seems to be an underlying API incompatibility with Pipecat, or at least, the Pipecat `master`:

```
pipecat.processors.frame_processor:push_error_frame:745 - AsteriskWSServerOutputTransport#0 exception (/usr/src/pipecat/src/pipecat/transports/asterisk/chan_ws_server.py:405): Error processing frame: 'AsteriskWSServerOutputTransport' object has no attribute '_write_frame'
```

I do not pretend to know the architecture of this codebase or Pipecat internals, so all I can offer is Claude's analysis:

```
⏺ The issue is clear. AsteriskWSServerOutputTransport.send_message (line 405) calls self._write_frame(frame), but _write_frame is
  defined on WebsocketServerOutputTransport — not on BaseOutputTransport or AsteriskWSServerOutputTransport. The Asterisk transport was
   copied/ported from the generic WebSocket server but _write_frame was never added.

  The fix is to implement the method inline, matching the pattern already used throughout the Asterisk transport (serialize →
  websocket.send):
  
 405 -        await self._write_frame(frame)
 405 +        if not self._params.serializer:
 406 +            return
 407 +        try:
 408 +            payload = await self._params.serializer.serialize(frame)
 409 +            if payload and self._websocket:
 410 +                await self._websocket.send(payload)
 411 +        except Exception as e:
 412 +            logger.error(f"{self} exception sending message: {e.__class__.__name__} ({e})")
```

But I am not in a position to evaluate the merit of this suggestion. However, I am curious how it ever worked for you when you tested it, given this hindrance? Is there a specific release tag of Pipecat against which this was built?

NikolayShakin:
Thanks @abalashov, well, it worked with the version I initially forked(1000k + commits ago), so you can clone the project from there if you'd like to test. I guess it will work.
Which version did you use when you got this error?

abalashov:
@NikolayShakin I used it with the current `main`, whose `HEAD` is, as of this moment today (late evening 21 February 2026) at `b67af19d`. :-) 

I got it to work, but had to make two changes that were generating console errors. One was the one mentioned above:

```
diff --git a/src/pipecat/transports/asterisk/chan_ws_server.py b/src/pipecat/transports/asterisk/chan_ws_server.py
index 26d87d258..b1774abbf 100644
--- a/src/pipecat/transports/asterisk/chan_ws_server.py
+++ b/src/pipecat/transports/asterisk/chan_ws_server.py
@@ -402,7 +402,14 @@ class AsteriskWSServerOutputTransport(BaseOutputTransport):
         Args:
             frame: The transport message frame to send.
         """
-        await self._write_frame(frame)
+        if not self._params.serializer:
+            return
+        try:
+            payload = await self._params.serializer.serialize(frame)
+            if payload and self._websocket:
+                await self._websocket.send(payload)
+            logger.error(f"{self} exception sending message: {e.__class__.__name__} ({e})")

     async def write_audio_frame(self, frame: OutputAudioRawFrame) -> bool:
         """Write an audio frame to the WebSocket client with timing control.
```

The other was a type cast to explicitly force `optimal_frame_size` and `ptime` values to `int`, to sort out this problem:

```
pipecat.processors.frame_processor:push_error_frame:745 - AsteriskWSServerOutputTransport#0 exception (/usr/src/pipecat/src/pipecat/transports/asterisk/chan_ws_server.py:446): Error processing frame: unsupported operand type(s) for /: 'str' and 'int'
```

```
diff --git a/src/pipecat/transports/asterisk/chan_ws_server.py b/src/pipecat/transports/asterisk/chan_ws_server.py
index 26d87d258..aa4380ed3 100644
--- a/src/pipecat/transports/asterisk/chan_ws_server.py
+++ b/src/pipecat/transports/asterisk/chan_ws_server.py
@@ -432,6 +432,9 @@ class AsteriskWSServerOutputTransport(BaseOutputTransport):
             logger.warning(f"{self} MEDIA_START message missing ptime")
             raise ValueError("MEDIA_START message missing ptime")

+        optimal_frame_size = int(optimal_frame_size)
+        ptime = int(ptime)
+
         # Update internal parameters
         self._optimal_frame_size = optimal_frame_size
         logger.debug(f"{self} optimal frame size set to {self._optimal_frame_size} bytes")
```

abalashov:
Another question for you: is any of the event payload from Asterisk's wire protocol exposed downstream in the Pipecat pipeline?

For example, with a `MEDIA_START` payload:

```
{
    "event": "MEDIA_START",
    "connection_id": "e226e283-c90a-4ea9-9e37-389000b9ef47",
    "channel": "WebSocket/connectionid",
    "channel_id": "pbx1-123456789.999",
    "format": "ulaw",
    "optimal_frame_size": 160,
    "ptime": 20,
    "channel_variables": {
        "SOME_CHANNEL_VARIABLE": "some value",
        "ANOTHER_CHANNEL_VARIABLE": "some other value"
    }
}
```

I'd like to get some of those channel variable values and inject them into a system prompt, but there isn't an obvious way to do that. 

Edit: this was a stupid question -- I just need to read your comments. :-)

> So we send MEDIA_START event object to the pipeline as InputTransportMessageFrame.

Saving comment here for posterity.

abalashov:
How does this standalone WS server architecture work vis-a-vis multiple concurrent calls? It seems to me this `transport` is created as a global server, so this doesn't fit into the FastAPI pattern where the connection comes from somewhere else. What's the right way to structure a pipeline runner for this transport?

NikolayShakin:
Initially, I started with FastAPI, but FastAPI (starlett) WebSocket implementation doesn't allow you to work with both binary and text data in the same ws connection, which is required for the asterisk web-socket-channel.
So I used the regular WebSocket transport as the base for this transport. TBH, I wrote this transport more like an exercise out of curiosity, and I didn't think about production aspects like scaling or concurrent calls; I assumed that it was considered/solved for the original WebSocket transport.
  
Currently, I see a few options for simultaneous calls. If the scale of the project is small, I'd try to run multiple instances on different ports (one instance per call). If the scale is large, I see two main options:
1. Kubernetes
2. Add ws sessions management to this transport (to make it similar to starlett's ws server)
But I'm not a big expert in pipecat, maybe there are some better ways to do it. 

In my tests, I started the pipeline right from the ARI application on the same machine after the call arrived; it seems to be easy to run multiple servers in this scenario

abalashov:
It seems to work fine with `uvicorn` / FastAPI, just using your serialiser and not the transport:

```
import uvicorn
from fastapi import FastAPI, WebSocket

...

app = FastAPI()

@app.websocket("/ws")
async def ws_endpoint(websocket: WebSocket, caller_id: str):
    await websocket.accept()

    transport = FastAPIWebsocketTransport(
        websocket=websocket,
        params=FastAPIWebsocketParams(
            audio_in_enabled=True,
            audio_out_enabled=True,
            add_wav_header=False,
            serializer=AsteriskWsFrameSerializer(),
        ),
    )
    
[...]

if __name__ == "__main__":
    uvicorn.run(app, host="0.0.0.0", port=8765)
```

NikolayShakin:
ahah, that's an interesting side effect :), I didn't even try to use it with this transport
However, the main reason I've created the transport is to utilize Asterisk's audio_buffer on ws_channel. Without this buffer, I had an inconsistent bot voice quality

NikolayShakin:
@abalashov, I've checked the code, and you were right, it was not functional indeed. I guess I messed up when I was cleaning up the code, or forgot to push the latest changes :( 
Besides the problems you mentioned, I found a few more and fixed them. Hopefully, I'll have time to test it thoroughly next week

abalashov:
@NikolayShakin Thank you so much for your work on this! I declare this the most important PR of 2026, and I think the project should incorporate it promptly. There is absolutely no reason why the concept of PSTN and/or telephony interfaces should be narrowly defined in the limited mental categories of Telnyx, Twilio, etc.

It is very nice to have a clean and sophisticated reference implementation of the `chan_websocket` wire protocol for flow control, and saves an incredible amount of work, troubleshooting and needless duplication. It is deeply appreciated, and I will spread the good word.

abalashov:
Sadly, this branch no longer merges cleanly onto `main`. :( 

```
[sasha@BardzoDużyMac repos]$ cd pipecat
[sasha@BardzoDużyMac pipecat]$ git fetch origin pull/3229/head:pr-branch
remote: Enumerating objects: 57, done.
remote: Counting objects: 100% (25/25), done.
remote: Compressing objects: 100% (8/8), done.
remote: Total 57 (delta 18), reused 17 (delta 17), pack-reused 32 (from 2)
Unpacking objects: 100% (57/57), 36.15 KiB | 420.00 KiB/s, done.
From https://github.com/pipecat-ai/pipecat
 * [new ref]             refs/pull/3229/head -> pr-branch
[sasha@BardzoDużyMac pipecat]$ GIT_COMMITTER_NAME="automated" GIT_COMMITTER_EMAIL="noreply@localhost" git rebase pr-branch
Auto-merging src/pipecat/frames/frames.py
CONFLICT (content): Merge conflict in src/pipecat/frames/frames.py
Auto-merging src/pipecat/services/ai_service.py
Auto-merging src/pipecat/services/anthropic/llm.py
Auto-merging src/pipecat/services/asyncai/tts.py
Auto-merging src/pipecat/services/cartesia/tts.py
Auto-merging src/pipecat/services/deepgram/stt.py
Auto-merging src/pipecat/services/deepgram/stt_sagemaker.py
Auto-merging src/pipecat/services/elevenlabs/tts.py
CONFLICT (content): Merge conflict in src/pipecat/services/elevenlabs/tts.py
Auto-merging src/pipecat/services/gradium/tts.py
CONFLICT (content): Merge conflict in src/pipecat/services/gradium/tts.py
Auto-merging src/pipecat/services/inworld/tts.py
CONFLICT (content): Merge conflict in src/pipecat/services/inworld/tts.py
Auto-merging src/pipecat/services/llm_service.py
Auto-merging src/pipecat/services/openai/base_llm.py
Auto-merging src/pipecat/services/playht/tts.py
Auto-merging src/pipecat/services/resembleai/tts.py
Auto-merging src/pipecat/services/rime/tts.py
CONFLICT (content): Merge conflict in src/pipecat/services/rime/tts.py
Auto-merging src/pipecat/services/sarvam/stt.py
CONFLICT (content): Merge conflict in src/pipecat/services/sarvam/stt.py
Auto-merging src/pipecat/services/speechmatics/stt.py
Auto-merging src/pipecat/services/stt_service.py
CONFLICT (content): Merge conflict in src/pipecat/services/stt_service.py
Auto-merging src/pipecat/services/tts_service.py
error: could not apply 8a4ab611b... Broad service settings refactor, with the primary aim of making service settings discoverable and strongly-typed. Service settings can be updated at runtime with `*UpdateSettingsFrame`s.
hint: Resolve all conflicts manually, mark them as resolved with
hint: "git add/rm <conflicted_files>", then run "git rebase --continue".
hint: You can instead skip this commit: run "git rebase --skip".
hint: To abort and get back to the state before "git rebase", run "git rebase --abort".
Could not apply 8a4ab611b... Broad service settings refactor, with the primary aim of making service settings discoverable and strongly-typed. Service settings can be updated at runtime with `*UpdateSettingsFrame`s.
```

abalashov:
It would probably be helpful to my testing if I knew the last release tag of `pipecat` to which this branch could cleanly apply.

NikolayShakin:
@abalashov, please try again. I synced my branch with the main project, and it worked for me
```
MacBook-Pro:original_pipecat shakinnn$ git fetch origin pull/3229/head:pr-branch
Warning: Permanently added 'github.com' (ED25519) to the list of known hosts.
remote: Enumerating objects: 63, done.
remote: Counting objects: 100% (31/31), done.
remote: Compressing objects: 100% (10/10), done.
remote: Total 63 (delta 22), reused 22 (delta 20), pack-reused 32 (from 2)
Unpacking objects: 100% (63/63), 37.41 KiB | 1008.00 KiB/s, done.
From github.com:pipecat-ai/pipecat
 * [new ref]             refs/pull/3229/head -> pr-branch
MacBook-Pro:original_pipecat shakinnn$ git rebase pr-branch
Successfully rebased and updated refs/heads/main.
```

NikolayShakin:
> It would probably be helpful to my testing if I knew the last release tag of `pipecat` to which this branch could cleanly apply.

The latest version or the main branch that I merged successfully into my branch is  `f558105c3c0f5fbcc48c812fe4df0659da8fbeb4`, but I haven't checked if it works yet.

abalashov:
Sadly, after these fixes to your own PR branch, I'm getting failures streaming the TTS to Asterisk:

```
pipecat-mediator  | 2026-02-26 01:49:42.781 | DEBUG    | pipecat.processors.metrics.frame_processor_metrics:stop_ttfb_metrics:141 - HumeTTSService#0 TTFB: 0.785s
pipecat-mediator  | 2026-02-26 01:49:42.786 | DEBUG    | pipecat.transports.base_output:_bot_started_speaking:620 - Bot started speaking
asterisk          | [Feb 26 01:49:42] WARNING[113][C-00000001]: chan_websocket.c:1131 read_from_ws_and_queue: WebSocket/pipecat_conn/0xfffe90005118: WebSocket frame type 0 not supported
asterisk          | [Feb 26 01:49:42]     -- Channel WebSocket/pipecat_conn/0xfffe90005118 left 'simple_bridge' basic-bridge <814389d7-94ed-4a96-8f2a-9870159c1aa4>
asterisk          | [Feb 26 01:49:42]     -- Channel PJSIP/100-00000000 left 'simple_bridge' basic-bridge <814389d7-94ed-4a96-8f2a-9870159c1aa4>
asterisk          | [Feb 26 01:49:42]   == Spawn extension (internal-dial, 999, 3) exited non-zero on 'PJSIP/100-00000000'
pipecat-mediator  | 2026-02-26 01:49:42.786 | ERROR    | pipecat.transports.websocket.fastapi:send:152 - <pipecat.transports.websocket.fastapi.FastAPIWebsocketClient object at 0xffff2e3c16a0> exception sending data: TypeError (data contains inconsistent types), application_state: WebSocketState.CONNECTED
pipecat-mediator  | 2026-02-26 01:49:42.791 | ERROR    | pipecat.transports.websocket.fastapi:send:152 - <pipecat.transports.websocket.fastapi.FastAPIWebsocketClient object at 0xffff2e3c16a0> exception sending data: WebSocketDisconnect (), application_state: WebSocketState.DISCONNECTED
```

As a reminder, I am continuing to use your serialiser with FastAPI:

```
    transport = FastAPIWebsocketTransport(
        websocket=websocket,
        params=FastAPIWebsocketParams(
            audio_in_enabled=True,
            audio_out_enabled=True,
            add_wav_header=False,
            serializer=AsteriskWsFrameSerializer(),
        ),
    )
```

Your `serialize()` method returns a tuple, which the FastAPI side doesn't accept. It's possible to fix it there:

```
diff --git a/src/pipecat/transports/websocket/fastapi.py b/src/pipecat/transports/websocket/fastapi.py
index f52123e52..733728d95 100644
--- a/src/pipecat/transports/websocket/fastapi.py
+++ b/src/pipecat/transports/websocket/fastapi.py
@@ -498,6 +498,14 @@ class FastAPIWebsocketOutputTransport(BaseOutputTransport):
         try:
             payload = await self._params.serializer.serialize(frame)
             if payload:
+                # Some serializers (e.g. AsteriskWsFrameSerializer) may return a tuple of
+                # (str, bytes) when they need to send a control message followed by audio data
+                # in the same serialize() call. Send each element individually.
+                if isinstance(payload, tuple):
+                    for item in payload:
+                        await self._client.send(item)
+                    return
+
                 # Optional protocol-level audio packetization:
                 # If a downstream WebSocket media endpoint requires fixed-size PCM frames,
                 # configure params.fixed_audio_packet_size (e.g. 640 for 20ms @ 16kHz PCM16 mono).
```

... but I'm not sure this is optimal. Is it possible to have `serialize()` return what FastAPI expects? It worked before...

NikolayShakin:
Ah, this tuple is a hack from one of the first iterations, before I even started the transport implementation. Now I'm 100% sure that I've lost quite some code :(. 
But, good there is good news:
1. This tuple-based approach will not work with my transport either, so I need to fix it anyway 
2. When I started implementation, pipecat's FastAPI WebSocket client could only work with bin or text data and had a `_is_binary` property, but I checked the current version of this transport, and there is no such limitation anymore. That explains why the serializer can now work with the transport.

I'll fix the serializer, and try to rebuild the transport on FastAPI's webscoket, the only question is finding time for that.

Rickaym:
Hi, this is simply amazing! I think it will be one of the best additions to Pipecat ever. Over the weekend, I had the chance to test this implementation. It works pretty much off the bat for me with minimal changes. There is only one thing I encountered and it happens when the conversation gets longer, I started hearing 
popping noises in the audio, and I'm curious whether if this is because of my setup. Here are the details:

Audio recording:

[03-03-2026 11.50.mp3](https://github.com/user-attachments/files/25707029/03-03-2026.11.50.mp3)

The language is Turkish, you can hear the "pop" or "stutters" quite clearly in the background.

This is my TTS:
```
             tts = ElevenLabsTTSService(
                 api_key=_get_api_key("ELEVEN_API_KEY"),
                 voice_id="db0e7kB1Ok43TcwAEuyl",
                 model="eleven_turbo_v2_5",
                 text_filters=text_filters,
             )
```

The audio works perfectly using daily transport.

My extension:

```
exten => 101,1,NoOp(Voice Agent)
 same => n,Answer()
 same => n,Wait(1)
 same => n,Set(_call_uuid=${UUID()})
 same => n,Set(_did=${EXTEN})
 same => n,Set(_ani=${CALLERID(num)})
 same => n,Set(_linkedid=${CHANNEL(linkedid)})
 same => n,Dial(WebSocket/ai_media/c(ulaw)f(json))
 same => n,Hangup()
```

I saw the suggestion earlier to add a ` same => n,Set(JITTERBUFFER(adaptive)=60,300,40)` to the extension. I tried this and found that my call would no longer connect (I placed the line above the dial string), and gave up.

Any pointers on what this issue might be is well appreciated. I would love to help contribute if it is something I can.

NikolayShakin:
Merhaba @Rickaym, thanks for testing! First of all, could you please show the transport definition?
It's unlikely that the issues you have are related to TTS, especially with 11-labs. The transport (`AsteriskWSServerTransport`) has audio buffers management logic, and it should work, but not `FastAPIWebsocketTransport` which will deliver audio data, but doesn't care about buffers and flow control

AndreTeixeira1998:
Greetings, any news here?

is this PR in a good state to use?

NikolayShakin:
Hi @AndreTeixeira1998, it worked last time I checked. Please let me know if you have any specific issues

abalashov:
What are the prospects for this getting merged into the mainline of the project? I've pinned the pr-branch rebase against `main:a8cb2a26d15ad0b02761ed474f2f79f5e3159478` (from 25 Feb) for now, but of course, it would be nice if it could apply cleanly.

markbackman:
This is really great work! Congrats on collaborating to build it out.

The fastest way to get this into a form that can be used is to submit it as a community integration:

It would require a maintainer (or maintainers) and ideally a PyPI package.

As a community integration, this will be listed in docs so that it's more easily discovered.

Please let me know if you have any questions about this approach. Once we have time to review and maintain, we can consider pulling it into Pipecat. But, I think this is a great first step to make it easier to use.

NikolayShakin:
Hi @markbackman, thanks for your feedback!
Yes, I'm considering the community integration option. And yes, I have some questions :)
To use the asterisk-side media buffer, we need to implement a buffer and flow control on pipecat side. At first, I tried to implement these features in the serializer, but it was messy, and I moved them to transport. Effectively, I've created a new transport from the websocket transport. I don't like this approach, because I duplicated the code, and if tomorrow some changes happen in the original websocket transport, I will need to move them manually to my websocket_asterisk transport and so on. There is an option to use the original websocket implementation as a parent class, but in this case, we still have another transport specifically for Asterisk, which is almost the same as the parent transport. Another option is to change the original websocket transport to make it work with the asterisk serializer in a special way, but save the logic for all other cases. I like this approach better, but I don't understand how to do that in "community integration" mode.   
How would you go in this case?

abalashov:
@NikolayShakin I don't know if it helps, but as I mentioned, I've been using it with the FastAPI WS transport, and it works very well. I am yet to run into any problems or drawbacks with this. It raises the question of why the standalone WebSocket transport is even needed. Am I missing something? Is there some downside to using your serialiser this way?

Example:

```
app = FastAPI()

@app.websocket("/tts")
async def tts_endpoint(websocket: WebSocket, particle_ref: str, kv_context: str = None):
    await websocket.accept()

    ...

    transport = FastAPIWebsocketTransport(
        websocket=websocket,
        params=FastAPIWebsocketParams(
            audio_in_enabled=False,
            audio_out_enabled=True,
            add_wav_header=False,
            serializer=AsteriskWsFrameSerializer(),
        ),
    )

    pipeline = Pipeline([
        transport.input(),
        tts,
        transport.output(),
    ])

    task = PipelineTask(
        pipeline,
        params=PipelineParams(
            allow_interruptions=True,
            audio_in_sample_rate=8000,
            audio_out_sample_rate=8000,
            enable_metrics=True,
            enable_usage_metrics=True,
        )
    )

    await task.queue_frames([
        TTSSpeakFrame(final_particle_text),
    ])
    await task.stop_when_done()

    runner = PipelineRunner(handle_sigint=False)
    await runner.run(task)
```

It's really just the fact that `main` keeps changing that has kept this painful:

```
# 25 Feb 2026 or so
ARG last_pipecat_head_that_works="a8cb2a26d15ad0b02761ed474f2f79f5e3159478"

...

RUN     git clone https://github.com/pipecat-ai/pipecat && \
    cd pipecat && \
    git checkout ${last_pipecat_head_that_works} && \
    git fetch origin pull/3229/head:pr-branch && \
    GIT_COMMITTER_NAME="automated" GIT_COMMITTER_EMAIL="noreply@localhost" git rebase pr-branch 
```

NikolayShakin:
@abalashov FastAPI might work, but I think it's not production-ready; the transport doesn't support flow control(as far as I know). My main concern is that, in this case, the WS_channel audio buffer is not enabled on the Asterisk side [more here](https://docs.asterisk.org/Configuration/Channel-Drivers/WebSocket/#media-transfer). Without the buffer, I expect that TCP-based WebSocket connections will only work well in very good environments and network conditions. For instance,   what would happen when you send a lot of audio at once, e.g., TTS-generated 30-second-long audio, pipecat can deliver it to Asterisk in < 1 second. I'm not sure what Asterisk will do with those packets with the media buffer disabled

Or maybe I'm missing something

markbackman:
@NikolayShakin The `FastAPIWebsocketTransport` is used heavily in production—it's probably one of the most battle tested transports.

I would say that it's the best fit for what you're trying to do. If there are features missing, please let us know, but it seems to work based on @abalashov's findings.

NikolayShakin:
@markbackman, asterisk's websocket channels are designed to be used with audio buffering on the Asterisk side. 
Simply speaking, Asterisk expects that audio arrives in bursts, arbitrarily framed and timed. Asterisk will take care of framing and timing, but the audio buffer size is hardcoded (10000 frames). 
This means that in Pipecat's transport, we need to implement flow control logic(to avoid overflowing asterisk's buffer) and the internal buffer to save audio when asterisk's buffer is full. 
For `FastAPIWebsocketTransport` we will need to add audiobuffer, flow-control, bitrate/codec autodetection, and they are very specific to Asterisk; probably, but it makes no sense to add them in the main project if we are talking about community integration.
I'll try to implement these features and see

markbackman:
@NikolayShakin To answer your question about how to structure this as a community integration: subclass `FastAPIWebsocketOutputTransport` rather than ship a parallel transport. The hooks you need are already in place:

- `process_frame` is an override-and-super point. Catch `InputTransportMessageFrame` (the input side already pushes these for `MEDIA_START`) to grab `optimal_frame_size`/`ptime`, launch your consumer task, and send `START_MEDIA_BUFFERING`. Same hook handles `InterruptionFrame` so you can drop your local queue and send `FLUSH_MEDIA` before calling super.
- `write_audio_frame` is fully overridable. Skip super and enqueue into your own `asyncio.Queue`. That also skips `_write_audio_sleep`, which you don't want anyway since Asterisk does its own pacing.
- `_write_frame` on the base does serialize + send with error handling, so your consumer task can reuse it.
- `start`/`stop` for consumer and monitor task lifecycle.

The input side needs no changes. `_receive_messages` already deserializes and broadcasts events as `InputTransportMessageFrame`, which land in your output's `process_frame`.

On your concern that FastAPI doesn't do flow control: you're right, and it shouldn't. The buffering, XON/XOFF awareness, and `optimal_frame_size` reframing are Asterisk-specific and belong in the subclass, not in the base FastAPI transport. @abalashov's setup works for short utterances but will drop audio once you exceed Asterisk's ~1000-frame queue or send oddly-sized messages (the docs are explicit that non-aligned bytes are discarded when buffering isn't started). So your architectural instinct is correct, it's just a question of where the code lives.

As a PyPI package you'd export: the serializer, an `AsteriskFastAPIWebsocketTransport` that wraps FastAPI input + your custom output, and a params class extending `FastAPIWebsocketParams`. Likely a few hundred lines instead of duplicating the full transport.

Happy to provide an opinion if you have any other questions. And if a new hook in the base FastAPI transport would make this cleaner, open an issue here and we can discuss it.

abalashov:
> @abalashov's setup works for short utterances but will drop audio once you exceed Asterisk's ~1000-frame queue or send oddly-sized messages (the docs are explicit that non-aligned bytes are discarded when buffering isn't started). So your architectural instinct is correct, it's just a question of where the code lives.

I'm curious what constitutes a long utterance for this purpose. To be honest, I'm yet to encounter an instance of this being a problem, but I admit I haven't specially sought out the longest completion I can.

NikolayShakin:
> I'm curious what constitutes a long utterance for this purpose. To be honest, I'm yet to encounter an instance of this being a problem, but I admit I haven't specially sought out the longest completion I can.

for u/alaw it's 1000 frames x 20ms = 20secs

abalashov:
Ah, I see. This buffer size really should be configurable on the Asterisk side. Then it's up to people to shoot themselves in the foot with that if they so choose...

But no, none of this is to dispute that flow control is necessary in order to regulate the inflow in an Asterisk-compliant way. I see it now more clearly after @markbackman 's explanation.

NikolayShakin:
Thanks for the guidance @markbackman. It's clear now

NikolayShakin:
Hi everyone, I re-wrote the serializer and the transport based on Fastapi websocket transport. The package [pipecat-asterisk](https://pypi.org/project/pipecat-asterisk/) is now available on pypi. I would appreciate it a lot if you could find time to test it and provide your feedback

markbackman:
@NikolayShakin that's great! Want to submit a docs PR to make the transport and serializers as an official integration?

NikolayShakin:
Hi @markbackman, not yet, I'd like to test it first. I only tested it with the example pipeline in the code, and my experience tells me that the first iteration never works outside my dev environment :)
If someone can test it in a more production-like environment, it would be a great help
BTW, here is the link to the new implementation:
https://github.com/NikolayShakin/pipecat-asterisk

abalashov:
I will be contributing to this testing imminently in a production-like environment. :-)

jcolp:
*falls into this PR* Would making the queue size configurable on the Websocket channel instead of hardcoded (it would still default to 1000) be desirable, alongside putting queue parameters into MEDIA_START? We'd enforce a minimum, but probably not a maximum so one could shoot themselves in the foot in memory usage...

(Disclaimer: Seeing if there is interest, I also can't guarantee this happening soon)

abalashov:
> _falls into this PR_ Would making the queue size configurable on the Websocket channel instead of hardcoded (it would still default to 1000) be desirable, alongside putting queue parameters into MEDIA_START? We'd enforce a minimum, but probably not a maximum so one could shoot themselves in the foot in memory usage...
> 
> (Disclaimer: Seeing if there is interest, I also can't guarantee this happening soon)

I wasn't aware you were following this PR! A very pleasantly unexpected surprise, and I appreciate your engagement with it and your willingness to consider helpful ways to meet this constituency in the middle. 

My vote is just one, but I would wholeheartedly endorse this configurable queue depth idea.

NikolayShakin:
@jcolp, nice to see you here, thanks for coming :)

> making the queue size configurable

It's good to have; however, for this specific case, I don't think we need more than 1000. At the end of the day, it's a trade-off between storing the audio buffer in RAM on the machine running pipecat or on the machine running Asterisk. 
I guess in most large-scale architectures, Asterisk will run on one machine, and multiple pipcats in containers on some kind of Kubernetes or Docker Swarm-like environments. In this case, we will probably have a lot of RAM on the container orchestrator side and limited RAM on the asterisk side. So  1000 is a reasonable default assumption to avoid abusing memory on the asterisk side; it will work fine in 99% cases, in my opinion. 

> putting queue parameters into MEDIA_START

That's a good one. I was thinking it would be very convenient to have the buffering configurable in the websocket channel configuration when I wrote the transport, like `yes/no/dynamic`, where `dynamic` is the default, allowing using `START_MEDIA_BUFFERING`.
If that were the case, yes, sending the queue parameters in the `MEDIA_START` event would simplify the transport implementation a little bit. But currently, it's not important for this specific project, as it's already handled by the transport

Overall, I think there are lots more useful features to implement.
My votes are for:
- adding Transfer command, so we can transfer the call from pipecat through the websocket signalling, without ARI/AMI, which is required very often with AI-agents, something like `transfer: extension@context`
- adding DTMF commands, so we can send DTMFs as commands from pipecat to asterisk
- adding CLI command to show websocket channels
- adding "rtp set debug on" for websocket channels, in Asterisk 22.8.2 I could only see outgoing packets to the websocket `Writing websocket binary frame, length 320`  but not incoming, and it's only available with general debug enabled, which is too verbose.
- add a configurable warning when the last frame from the buffer ended up with a non-zero value, which is an indicator that "click" happened, and it would typically indicate some problems on the pipecat(or other app) side. It would simplify debugging a lot
- support asymmetric codec/sampling in the websocket channel, this one is a little bit exotic, but still, very often we have different codecs/sampling in audio from TTS and audio expected by STT. So it might be a good feature to have for a wide range of STT-LLM-TTS scenarios

And I noticed that when I send audio in large chunks close to the max size(<= 65500) asterisk kills the websocket connection. I didn't have time to debug the issue thoroughly; it might be my fault, but the error was `DEBUG[22367][C-00000050]: chan_websocket.c:1107 read_from_ws_and_queue: WebSocket/pipecat/0xfffee4009f58: WebSocket read error`

jcolp:
I've copied your response over to our side so we can keep track of it, we'll probably pull items from it over time.

NikolayShakin:
> And I noticed that when I send audio in large chunks close to the max size(<= 65500) asterisk kills the websocket connection. I didn't have time to debug the issue thoroughly; it might be my fault, but the error was DEBUG[22367][C-00000050]: chan_websocket.c:1107 read_from_ws_and_queue: WebSocket/pipecat/0xfffee4009f58: WebSocket read error

A smallish update, the issue I mentioned only happens with asterisk 22.8.2 running in docker on mac, I tested it with asterisk [23.3.0-rc2](https://github.com/asterisk/asterisk/commit/a423a5f3eaa8e3e6fafc69ed60a7193cbfa8cbc7) on a virtual machine with debian13, no issues were observed.

NikolayShakin:
Thanks, everyone! I'm closing the PR. 
The transport is available as a PyPI package https://pypi.org/project/pipecat-asterisk/
And and in the repo https://github.com/NikolayShakin/pipecat-asterisk
A pull request is added to make it a registered community integration https://github.com/pipecat-ai/docs/pull/810

PR #3233, opened 2025-12-15
Improve Elevenlabs realtime transcription service

jaydamani:
- improve error handling to log all error types

Add error frame for all error types from https://elevenlabs.io/docs/developers/guides/cookbooks/speech-to-text/streaming#error-handling 

- fix issue with infinite loop when websocket disconnects
Explanation: the [_receive_messages](https://github.com/pipecat-ai/pipecat/blob/7e424d750ebee48d8ff8af6f009b8257acd41dc5/src/pipecat/services/elevenlabs/stt.py#L693) method is executed inside a loop with attempt to reconnect on error but the eleven labs implementation had a try catch due to which, if the websocket randomly disconnects then this part goes into an infinite loop which blocks the whole process.

https://github.com/pipecat-ai/pipecat/blob/7e424d750ebee48d8ff8af6f009b8257acd41dc5/src/pipecat/services/websocket_service.py#L133-L154

### DIFF
--- a/changelog/3233.fixed.md
+++ b/changelog/3233.fixed.md
@@ -1,0 +1,2 @@
+- Improved error handling in `ElevenLabsRealtimeSTTService`
+- Fixed an issue in `ElevenLabsRealtimeSTTService` causing an infinite loop that blocks the process if the websocket disconnects due to an error


vmfullyfaltu:
I had this exact problem where i got this infinite loop and this PR fixed the issue for me. Looking forward to seeing this integrated into the main branch.
thanks @jaydamani 

2025-12-16 22:22:50.856 | WARNING  | pipecat.services.elevenlabs.stt:_receive_messages:709 - ElevenLabsRealtimeSTTService#0 WebSocket connection closed: no close frame received or sent
2025-12-16 22:22:50.856 | WARNING  | pipecat.services.elevenlabs.stt:_receive_messages:709 - ElevenLabsRealtimeSTTService#0 WebSocket connection closed: no close frame received or sent
2025-12-16 22:22:50.857 | WARNING  | pipecat.services.elevenlabs.stt:_receive_messages:709 - ElevenLabsRealtimeSTTService#0 WebSocket connection closed: no close frame received or sent
2025-12-16 22:22:50.857 | WARNING  | pipecat.services.elevenlabs.stt:_receive_messages:709 - ElevenLabsRealtimeSTTService#0 WebSocket connection closed: no close frame received or sent
2025-12-16 22:22:50.858 | WARNING  | pipecat.services.elevenlabs.stt:_receive_messages:709 - ElevenLabsRealtimeSTTService#0 WebSocket connection closed: no close frame received or sent

src/pipecat/services/elevenlabs/stt.py
```diff
@@ -738,12 +733,14 @@ async def _process_response(self, data: dict):
             error_msg = data.get("error", "Authentication error")
             logger.error(f"ElevenLabs auth error: {error_msg}")
             await self.push_error(error_msg=f"Auth error: {error_msg}")
-
         elif message_type == "quota_exceeded_error":
             error_msg = data.get("error", "Quota exceeded")
             logger.error(f"ElevenLabs quota exceeded: {error_msg}")
             await self.push_error(error_msg=f"Quota exceeded: {error_msg}")
-
+        elif message_type in error_types:
```
markbackman: Good catch!

We can make this more concise and more future proof if we just check for `_error` in the message_type:
```
        elif message_type and message_type.endswith("_error"):
            # Handle all error types
            error_msg = data.get("error", "Unknown error")
            await self.push_error(error_msg=f"ElevenLabs {message_type}: {error_msg}")
```

This would replace your proposal and the individual checks for `"error"`, `"auth_error"`, and `"quota_exceeded_error"`.

markbackman:
Also, can you please:

- Add a changelog fragment?
- Lint the code (run `./scripts/fix-ruff.sh`) or install the pre-commit hook (`uv run pre-commit install`)

src/pipecat/services/elevenlabs/stt.py
jaydamani: We also have error types like queue_overflow, which do not have error at the end. Should we use a combined approach replacing only condition for types that have _error in them with a common condition.

markbackman: From the reference docs, I only see errors ending in `_error`.
https://elevenlabs.io/docs/api-reference/speech-to-text/v-1-speech-to-text-realtime#receive

Are you an 11Labs engineer or can you point me to the docs with the other error types?

jaydamani: https://elevenlabs.io/docs/developers/guides/cookbooks/speech-to-text/streaming#error-handling

markbackman: Check out the reference docs that I linked. The same errors are listed, but all have `_error` appended. This is a docs inconsistency. I would imagine that the reference docs are accurate, but I can ask the 11Labs team to be sure.

jaydamani: Yeah, that would be helpful.
We recently had issues with `queue_overflow` error which led to this PR but I am not sure about other error types.

markbackman: Cool. I wrote to the 11Labs team. I'll let you know what I find out. That can inform the implementation.

markbackman: Ok! The team confirmed that the errors in the guide are correct (not the reference docs). The reference docs are now updated.

We can probably use something like this:
```
        elif message_type in (
            "error",
            "auth_error",
            "quota_exceeded",
            "commit_throttled",
            "unaccepted_terms",
            "rate_limited",
            "queue_overflow",
            "resource_exhausted",
            "session_time_limit_exceeded",
            "input_error",
            "chunk_size_exceeded",
            "insufficient_audio_activity",
            "transcriber_error",
        ):
            # Handle all error types
            error_msg = data.get("error", "Unknown error")
            await self.push_error(error_msg=f"ElevenLabs {message_type}: {error_msg}")
```

Note, no logger.error is needed as `push_error()` calls `push_error_frame()` which calls `logger.error()` and pushes the `ErrorFrame`.

jaydamani:
@markbackman Can you please review again.
Also, after testing I noticed that if the websocket closes due a [ConnectionClosedError](https://websockets.readthedocs.io/en/stable/faq/common.html#what-does-connectionclosederror-no-close-frame-received-or-sent-mean) which may happen due to something like network/server issue then the [_receive_task_handler](https://github.com/jaydamani/pipecat/blob/614d5e0d19efa6382aa3d528b1ceeca2276f0e76/src/pipecat/services/websocket_service.py#L123) completes and when the websocket is connected again after receiving new audio, the _receive_task is not None so the receving task handler is not started again which stops the transcription. 
What are your thoughts on how to resolve this?

I checked other services and they all seem to have their own logic for reconnection. I think one of the below solutions can work:
Solution 1:
- When (re)connecting to a websocket, add condition to check if _receive_task is complete

https://github.com/pipecat-ai/pipecat/blob/614d5e0d19efa6382aa3d528b1ceeca2276f0e76/src/pipecat/services/elevenlabs/stt.py#L604-L608

Solution  2:
- Update WebsocketSTTService to reconnect on ConnectionClosedError

https://github.com/pipecat-ai/pipecat/blob/c46a20328d532b35d5175f9b51bb018328536455/src/pipecat/services/websocket_service.py#L133-L143

markbackman:
Let's keep this PR scoped to this change.

As for the `ConnectionClosedError`, the base WebsocketService handles this error:
https://github.com/pipecat-ai/pipecat/blob/main/src/pipecat/services/websocket_service.py#L140

We just added a PR to warn on this situation, which was part of 0.0.97:
https://github.com/pipecat-ai/pipecat/pull/3185

This also eliminates the spamming of `error receiving messages: no close frame received or sent`, so it's possible that this is already fixed for you. Which version are you running?

changelog/3233.fixed.md
```diff
@@ -0,0 +1,2 @@
+- Improved error handling in `ElevenLabsRealtimeSTTService`
+- Fixed an issue in `ElevenLabsRealtimeSTTService` causing an infinite loop that blocks the process if the websocket disconnects due to an error
```
markbackman, line 2: Remove?

```suggestion
```

jaydamani:
> Let's keep this PR scoped to this change.
> 
> As for the `ConnectionClosedError`, the base WebsocketService handles this error: https://github.com/pipecat-ai/pipecat/blob/main/src/pipecat/services/websocket_service.py#L140
> 
> We just added a PR to warn on this situation, which was part of 0.0.97: #3185
> 
> This also eliminates the spamming of `error receiving messages: no close frame received or sent`, so it's possible that this is already fixed for you. Which version are you running?

We are already running 0.0.97 and it still spams the logs and also blocks the process because  [_receive_messages](https://github.com/pipecat-ai/pipecat/blob/7e424d750ebee48d8ff8af6f009b8257acd41dc5/src/pipecat/services/elevenlabs/stt.py#L693) has it's own exception handler so the exception never reaches the base WebsocketService. 
To fix the above issue, I made this [change](https://github.com/pipecat-ai/pipecat/pull/3233/files#diff-95dc73ca4604c22047d84b1eb8bfc54d7a6d171cc2d4f0019282d2b461764e15L693-L709) after which the base WebsocketService correctly handles the exception.
But, when there is ConnectionClosedError specifically due to  [this](https://websockets.readthedocs.io/en/stable/faq/common.html#what-does-connectionclosederror-no-close-frame-received-or-sent-mean) the receiver task handler has ended so transcription is not received even when the websocket reconnects.
Also, it is not restarted on websocket reconnect because `_receive_task` is no longer none.
https://github.com/pipecat-ai/pipecat/blob/614d5e0d19efa6382aa3d528b1ceeca2276f0e76/src/pipecat/services/elevenlabs/stt.py#L608-L609

markbackman (approved):
> it still spams the logs and also blocks the process because [_receive_messages](https://github.com/pipecat-ai/pipecat/blob/7e424d750ebee48d8ff8af6f009b8257acd41dc5/src/pipecat/services/elevenlabs/stt.py#L693) has it's own exception handler so the exception never reaches the base WebsocketService.

Thanks for clarifying. This looks good then!  Approving.

For this issue (https://github.com/pipecat-ai/pipecat/pull/3233#issuecomment-3666811350), please raise a new PR. Improving the `WebsocketService` would be the preferred change since it will improve this case for all services inheriting from the `WebsocketService` (both TTS and STT).

PR #3234, opened 2025-12-15
Fix LLM context tool audio content handling

hwuiwon:
Description

- Build audio message content before use and thread the encoder correctly to avoid undefined content and ensure audio payload is appended.

### DIFF
--- a/src/pipecat/processors/aggregators/llm_context.py
+++ b/src/pipecat/processors/aggregators/llm_context.py
@@ -87,19 +87,10 @@ class LLMContext:
         # Convert tools to ToolsSchema if needed.
         # If the tools are already a ToolsSchema, this is a no-op.
         # Otherwise, we wrap them in a shim ToolsSchema.
-        converted_tools: ToolsSchema | NotGiven
-        raw_tools = openai_context.tools
-        if isinstance(raw_tools, list):
+        converted_tools = openai_context.tools
+        if isinstance(converted_tools, list):
             converted_tools = ToolsSchema(
-                standard_tools=[], custom_tools={AdapterType.SHIM: raw_tools}
-            )
-        elif isinstance(raw_tools, ToolsSchema):
-            converted_tools = raw_tools
-        elif raw_tools is NOT_GIVEN:
-            converted_tools = NOT_GIVEN
-        else:
-            raise TypeError(
-                f"Unsupported tools type when converting OpenAI context: {type(raw_tools)}"
+                standard_tools=[], custom_tools={AdapterType.SHIM: converted_tools}
             )
         return LLMContext(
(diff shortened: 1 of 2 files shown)


src/pipecat/processors/aggregators/llm_context.py
```diff
@@ -87,10 +87,19 @@ def from_openai_context(openai_context: "OpenAILLMContext") -> "LLMContext":
...
             converted_tools = ToolsSchema(
-                standard_tools=[], custom_tools={AdapterType.SHIM: converted_tools}
+                standard_tools=[], custom_tools={AdapterType.SHIM: raw_tools}
+            )
+        elif isinstance(raw_tools, ToolsSchema):
+            converted_tools = raw_tools
+        elif raw_tools is NOT_GIVEN:
+            converted_tools = NOT_GIVEN
+        else:
+            raise TypeError(
```
kompfner: Don't we essentially do this type-checking already in the `LLMContext` initializer, invoked on line 104?

What's the issue you're addressing with this change? Is it that, if the validation failed in the `LLMContext` initializer, the message might be confusing to the user because they were never explicitly initializing an `LLMContext` object?

kompfner: In general: I'd be hesitant to add more smarts to this shim, since it's, well, just a shim for older code (and probably won't be long lived, as we'll be deprecating `OpenAILLMContext` before long in favor of `LLMContext`).

src/pipecat/processors/aggregators/llm_context.py
```diff
@@ -195,7 +203,7 @@ async def encode_audio():
                     wf.setframerate(sample_rate)
                     wf.writeframes(data)
 
-            encoded_audio = base64.b64encode(buffer.getvalue()).decode("utf-8")
+                encoded_audio = base64.b64encode(buffer.getvalue()).decode("utf-8")
```
kompfner, line 197: Yup, I agree with your two fixes to `create_audio_message()`, which does seem to be broken today. cc @aconchillo, who last touched this code, in case he wants to double-check

aconchillo: Yeah, this looks good! Thanks!

src/pipecat/processors/aggregators/llm_context.py
hwuiwon: I fixed it because type checking was complaining about invalid-argument-type but I'm fine with leaving it as is.

Screenshot 2025-12-16 at 3 54 56 PM

hwuiwon:
@kompfner Hi, I have added the changelog and reverted the change for the tool conversion!

PR #3240, opened 2025-12-16
WordTTSService: make sure word timestamps are always started

aconchillo:

It's possible that the service sends word timestamps before receiving any audio. So, we need to make sure word timestamps are properly started.

### DIFF
--- a/changelog/3240.changed.md
+++ b/changelog/3240.changed.md
@@ -1,0 +1,2 @@
+- ⚠️ Breaking change: `WordTTSService.start_word_timestamps()` and
+  `WordTTSService.reset_word_timestamps()` are now async.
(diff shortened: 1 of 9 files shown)


markbackman (approved):
Looks good. The breaking change should only affect our service classes.

Speaking of which, there are two new service classes that use `start_word_timestamps`: InworldTTSService, InworldHttpTTSService. Can you update those too? (Might need to rebase `main` to get them. They were merged this afternoon.

aconchillo:
> Looks good. The breaking change should only affect our service classes.
> 
> Speaking of which, there are two new service classes that use `start_word_timestamps`: InworldTTSService, InworldHttpTTSService. Can you update those too? (Might need to rebase `main` to get them. They were merged this afternoon.

Done!

PR #3247, opened 2025-12-17
SimpleTextAggregator: Strip whitespace in the returned aggregation

markbackman:

### DIFF
--- a/src/pipecat/utils/text/simple_text_aggregator.py
+++ b/src/pipecat/utils/text/simple_text_aggregator.py
@@ -40,7 +40,7 @@ class SimpleTextAggregator(BaseTextAggregator):
         Returns:
             The text that has been accumulated in the buffer.
         """
-        return Aggregation(text=self._text.strip(), type=AggregationType.SENTENCE)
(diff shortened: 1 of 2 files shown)


src/pipecat/utils/text/simple_text_aggregator.py
```diff
@@ -97,7 +97,7 @@ async def _check_sentence_with_lookahead(self, char: str) -> Optional[Aggregatio
                     # NLTK confirmed a sentence - return it
                     result = self._text[:eos_marker]
                     self._text = self._text[eos_marker:]
-                    return Aggregation(text=result, type=AggregationType.SENTENCE)
+                    return Aggregation(text=result.strip(), type=AggregationType.SENTENCE)
```
mattieruth: it just occurred to me that maybe we should specify to only strip spaces?

markbackman: We discussed stripping only spaces. I've updated the commit to do that for the `text` property, `aggregate()` method, and `flush()` method.

PR #3248, opened 2025-12-17
Bump the RTVI version to 1.1.0 and add pipecat versioning to the botR…

mattieruth:
…eady about field

The pipecat library and version will now be included by default in the `about` field in the `bot-ready` message. I have also bumped the RTVI version to 1.1.0. Sadly, I'd been forgetting to do this along the way...

### DIFF
--- a/src/pipecat/processors/frameworks/rtvi.py
+++ b/src/pipecat/processors/frameworks/rtvi.py
@@ -31,6 +31,7 @@ from typing import (
 from loguru import logger
 from pydantic import BaseModel, Field, PrivateAttr, ValidationError
 
+from pipecat import version as pipecat_version
 from pipecat.audio.utils import calculate_audio_volume
 from pipecat.frames.frames import (
     AggregatedTextFrame,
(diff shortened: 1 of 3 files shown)


src/pipecat/processors/frameworks/rtvi.py
```diff
@@ -15,6 +15,7 @@
 import base64
 import time
 from dataclasses import dataclass
+from importlib.metadata import version as lib_version
```
markbackman: We already have version exported, which you can access from this import:
```
from pipecat import __version__ as pipecat_version
```

src/pipecat/processors/frameworks/rtvi.py
```diff
@@ -1873,14 +1879,21 @@ async def _handle_action(self, request_id: Optional[str], data: RTVIActionRun):
...
+
+        Args:
+            about: Optional information about the bot to include in the ready message.
+                   If left as None, the pipecat library and version will be used.
+        """
         config = None
         if self._client_version and self._client_version[0] < 1:
             config = self._config.config
+        if not about:
+            about = {"library": "pipecat-ai", "library_version": f"{lib_version('pipecat-ai')}"}
```
markbackman: Then use the pipecat version here:
```suggestion
            about = {"library": "pipecat-ai", "library_version": pipecat_version}
```

markbackman (approved):
LGTM

PR #3252, opened 2025-12-17
add vision response and text frames

aconchillo:

Currently, the vision services only push `TextFrame`s. The new aggregators being used in the TTS services flush any aggregated text when receiving an LLM end response. However, vision service don't do that.

In this PR we introduce three new frames: `VisionFullResponseStartFrame`, `VisionFullResponseEndFrame` and `VisionTextFrame`. So, vision services should push:

```
VisionFullResponseStartFrame VisionTextFrame.... VisionFullResponseEndFrame
```

This matches the LLM behavior.


markbackman (approved):
👨‍🍳 💋

PR #3255, opened 2025-12-17
Update linting scripts to use ruff version installed by uv

markbackman:

I stumbled on an issue where my local ruff check was passing but CI was not. The root cause is that my system version of ruff was different than what uv installs. The solution is to make the fix-ruff and pre-commit scripts use the uv installed ruff version. This will ensure linting works uniformly for everyone.

Also, some housekeeping on our github workflows; now they all use Python 3.12.

No changelog needed since these changes are unrelated to source.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -21,20 +21,20 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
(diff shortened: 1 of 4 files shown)


kompfner (approved):
Good catch! Stumbling into the edge case of a system-uv mismatch sounds...ruff.

PR #3256, opened 2025-12-17
Inworld: docstring and websockets fixes

cshape:

Updates to Inworld TTS plugin beyond original PR: https://github.com/pipecat-ai/pipecat/pull/3239

- Letting client control if they want timestamps or not (defaulting to disabled as new 1.5 models won't support timestamps initially)
- Setting buffer character threshold to support longer outputs
- Updating docstrings

### DIFF
--- a/src/pipecat/services/inworld/tts.py
+++ b/src/pipecat/services/inworld/tts.py
@@ -24,6 +24,7 @@ from loguru import logger
 from pydantic import BaseModel
 
 try:
+    import websockets
     from websockets.asyncio.client import connect as websocket_connect
     from websockets.protocol import State
 except ModuleNotFoundError as e:
@@ -489,6 +490,7 @@ class InworldTTSService(AudioContextWordTTSService):
         }
 
         self._receive_task = None
+        self._keepalive_task = None
         self._context_id = None
         self._started = False
         self._flush_complete_event: Optional[asyncio.Event] = None
@@ -667,6 +669,47 @@ class InworldTTSService(AudioContextWordTTSService):
             return self._websocket
         raise Exception("Websocket not connected")
 
+    async def _keepalive_task_handler(self):
+        """Send periodic keepalive messages to maintain WebSocket connection.
+
+        Inworld closes WebSocket connections after 10 minutes of inactivity.
+        This sends messages every 30 seconds to prevent timeout.
+        """
+        KEEPALIVE_SLEEP = 30
+        while True:
+            await asyncio.sleep(KEEPALIVE_SLEEP)
+            try:
+                if self._websocket and self._websocket.state is State.OPEN:
+                    if self._context_id:
+                        # Send keepalive with active context ID
+                        msg = {"send_text": {"text": ""}, "contextId": self._context_id}
+                        await self._websocket.send(json.dumps(msg))
+                    else:
+                        # No active context - create temporary one, send text, close it
+                        keepalive_ctx = str(uuid.uuid4())
+                        # Create context
+                        create_msg = {
+                            "create": {
+                                "voiceId": self._settings["voiceId"],
+                                "modelId": self._settings["modelId"],
+                                "audioConfig": self._settings["audioConfig"],
+                            },
+                            "contextId": keepalive_ctx,
+                        }
+                        await self._websocket.send(json.dumps(create_msg))
+                        # Send empty text
+                        text_msg = {"send_text": {"text": ""}, "contextId": keepalive_ctx}
+                        await self._websocket.send(json.dumps(text_msg))
+                        # Close the temporary context
+                        close_msg = {"close_context": {}, "contextId": keepalive_ctx}
+                        await self._websocket.send(json.dumps(close_msg))
+            except websockets.ConnectionClosed as e:
+                logger.warning(f"{self} keepalive error: {e}")
+                break
+            except Exception as e:
+                logger.warning(f"{self} keepalive error: {e}")
+                break
+
     async def _connect(self):
         """Connect to the Inworld WebSocket TTS service.
 
@@ -677,12 +720,19 @@ class InworldTTSService(AudioContextWordTTSService):
         if self._websocket and not self._receive_task:
             self._receive_task = self.create_task(self._receive_task_handler(self._report_error))
 
+        if self._websocket and not self._keepalive_task:
+            self._keepalive_task = self.create_task(self._keepalive_task_handler())
+
     async def _disconnect(self):
         """Disconnect from the Inworld WebSocket TTS service.
 
         Returns:
             The websocket.
         """
+        if self._keepalive_task:
+            await self.cancel_task(self._keepalive_task)
+            self._keepalive_task = None
+
         if self._receive_task:
             await self.cancel_task(self._receive_task)
             self._receive_task = None
--- a/changelog/3256.changed.md
+++ b/changelog/3256.changed.md
@@ -3,3 +3,4 @@
 
   - Improved timestamp functionality to be accurate for long generations
 - Updated websockets implementation to better flush and close context with long generations
+- Added keep alive ping to prevent session timeouts
--- a/examples/foundational/07ab-interruptible-inworld.py
+++ b/examples/foundational/07ab-interruptible-inworld.py
@@ -107,7 +107,7 @@ async def run_bot(transport: BaseTransport, runner_args: RunnerArguments):
                 }
             ),
         ],
-        idle_timeout_secs=runner_args.pipeline_idle_timeout_secs,
+        idle_timeout_secs=3000,
     )
 
     @transport.event_handler("on_client_connected")


src/pipecat/services/inworld/tts.py
```diff
@@ -57,10 +59,13 @@ class InputParams(BaseModel):
         Parameters:
             temperature: Temperature for speech synthesis.
             speaking_rate: Speaking rate for speech synthesis.
+            enable_word_timestamps: Whether to enable word-level timestamps.
+                Defaults to False (disabled).
         """
 
         temperature: Optional[float] = None
         speaking_rate: Optional[float] = None
+        enable_word_timestamps: bool = False
```
markbackman: I would strongly encourage that you set the default to True. Emitting words is very important for voice AI app and is worth a slight latency penalty.

You also mentioned that the newest model doesn't yet support word timestamps. If you set `enable_word_timestamps` to True, does the model just ignore the input or does it result in an error?

src/pipecat/services/inworld/tts.py
```diff
@@ -520,15 +529,15 @@ async def cancel(self, frame: CancelFrame):
         await self._disconnect()
 
     async def flush_audio(self):
-        """Flush any pending audio from the Inworld WebSocket TTS service.
+        """Flush any pending audio without closing the context.
```
markbackman: `flush_audio()` is called in three situations:

1. You a `TTSSpeakFrame` is pushed. This is a predetermined input to the model sent by the application (not streamed from the LLM).
2. An `LLMFullResponseEndFrame` is received indicating that the final content streamed from the LLM has been received.
3. An `EndFrame` is received indicating that the conversation is over.

In all cases, it makes sense to close the context. The same is true in the Cartesia and ElevenLabs services, so I'd recommend following the same pattern. This ensures that a new context is opened for the subsequent bot turn.

src/pipecat/services/inworld/tts.py
```diff
@@ -520,15 +529,15 @@ async def cancel(self, frame: CancelFrame):
...
 
-        Args:
-            frame: The flush frame.
+        This triggers synthesis of all accumulated text in the buffer while
+        keeping the context open for subsequent text. The context is only
+        closed on interruption, disconnect, or end of session.
```
markbackman: The context should be closed on:
- Interruption
- End of bot turn
- End of session (e.g. disconnect)

src/pipecat/services/inworld/tts.py
```diff
@@ -538,16 +547,14 @@ async def push_frame(self, frame: Frame, direction: FrameDirection = FrameDirect
             direction: The direction to push the frame.
         """
         await super().push_frame(frame, direction)
-        if isinstance(frame, TTSStoppedFrame):
+        if isinstance(frame, (TTSStoppedFrame, InterruptionFrame)):
```
markbackman, line 573: Nice catch 🙏

src/pipecat/services/inworld/tts.py
```diff
@@ -678,6 +689,21 @@ async def _process_messages(self):
...
+            msg_types = []
+            if "contextCreated" in result:
+                msg_types.append("contextCreated")
+            if "audioChunk" in result:
+                msg_types.append("audioChunk")
+            if "flushCompleted" in result:
+                msg_types.append("flushCompleted")
+            if "contextClosed" in result:
+                msg_types.append("contextClosed")
+            logger.debug(
```
markbackman: Maybe trace logging as this could get noisy, especially in logging the `audioChunk` messages.

```suggestion
            logger.trace(
```

src/pipecat/services/inworld/tts.py
```diff
@@ -573,16 +580,21 @@ async def _handle_interruption(self, frame: InterruptionFrame, direction: FrameD
             frame: The interruption frame.
             direction: The direction of the interruption.
         """
+        old_context_id = self._context_id
```
markbackman, line 644: Are you adding this for logging purposes only?

Regarding logging, I think this is too much logging. You have three debug logs that will appear when an interruption occurs. Interruptions are not something that developers generally care about in their application logs. That's why this is a trace log. I'd recommend the current code unless there's subtle functional difference that I'm missing.

src/pipecat/services/inworld/tts.py
```diff
@@ -689,15 +715,26 @@ async def _process_messages(self):
                 await self.push_error(error_msg=str(msg["error"]))
                 continue
 
-            # Skip messages for unavailable contexts
+            # Check if this message belongs to an available context.
```
markbackman, line 830: This makes sense and matches the pattern we use for ElevenLabsTTSService. Good catch.

src/pipecat/services/inworld/tts.py
```diff
@@ -706,8 +743,6 @@ async def _process_messages(self):
                 frame = TTSAudioRawFrame(audio, self.sample_rate, 1)
 
                 if ctx_id:
```
markbackman: You may be able to drop the `if ctx_id:` check all together as we already have a new context which we created above. That should allow you to safely call:
```
await self.append_to_audio_context(ctx_id, frame)
```

src/pipecat/services/inworld/tts.py
```diff
@@ -717,12 +752,25 @@ async def _process_messages(self):
                     if word_times:
                         await self.add_word_timestamps(word_times)
 
-            # Handle context completion
-            if "flushCompleted" in result or "contextClosed" in result:
+            # Handle context created confirmation
+            if "contextCreated" in result:
+                logger.debug(f"{self}: Context created on server: {ctx_id}")
```
markbackman: Should be trace

```suggestion
                logger.trace(f"{self}: Context created on server: {ctx_id}")
```

src/pipecat/services/inworld/tts.py
```diff
@@ -717,12 +752,25 @@ async def _process_messages(self):
...
+
+            # Handle flush completion - context is still valid, just acknowledge it
+            if "flushCompleted" in result:
+                logger.debug(f"{self}: Flush completed for context {ctx_id}")
```
markbackman: Same here
```suggestion
                logger.trace(f"{self}: Flush completed for context {ctx_id}")
```

src/pipecat/services/inworld/tts.py
```diff
@@ -717,12 +752,25 @@ async def _process_messages(self):
...
+
+            # Handle context closed - context no longer exists on server
+            if "contextClosed" in result:
+                logger.debug(f"{self}: Context closed on server: {ctx_id}")
```
markbackman:
```suggestion
                logger.trace(f"{self}: Context closed on server: {ctx_id}")
```

src/pipecat/services/inworld/tts.py
```diff
@@ -689,15 +715,26 @@ async def _process_messages(self):
...
+                    continue
 
             # Process audio chunk
             audio_chunk = result.get("audioChunk", {})
             audio_b64 = audio_chunk.get("audioContent")
 
             if audio_b64:
+                logger.debug(f"{self}: Processing audio chunk for context {ctx_id}")
                 await self.stop_ttfb_metrics()
                 self.start_word_timestamps()
```
markbackman: This must be awaited. I think you're behind `main`. This was caught by my colleague already and fixed on `main`.

```suggestion
                await self.start_word_timestamps()
```

src/pipecat/services/inworld/tts.py
```diff
@@ -689,15 +715,26 @@ async def _process_messages(self):
...
+                    # This is a message from an old/closed context - skip it
+                    logger.debug(f"{self}: Skipping message from unavailable context: {ctx_id}")
+                    continue
 
             # Process audio chunk
             audio_chunk = result.get("audioChunk", {})
             audio_b64 = audio_chunk.get("audioContent")
 
             if audio_b64:
+                logger.debug(f"{self}: Processing audio chunk for context {ctx_id}")
```
markbackman: I'd say to remove or set to trace.

src/pipecat/services/inworld/tts.py
```diff
@@ -752,14 +800,26 @@ async def _send_context(self, context_id: str):
...
+        else:
+            # Default to 250 chars threshold
+            create_config["bufferCharThreshold"] = 250
 
-        create_config["timestampType"] = self._timestamp_type
+        if self._enable_word_timestamps:
+            create_config["timestampType"] = "WORD"
 
         msg = {"create": create_config, "contextId": context_id}
+        logger.debug(f"{self}: Sending context create: {create_config}")
```
markbackman: Trace?

src/pipecat/services/inworld/tts.py
```diff
@@ -811,12 +871,19 @@ async def run_tts(self, text: str) -> AsyncGenerator[Frame, None]:
                     await self.start_ttfb_metrics()
                     yield TTSStartedFrame()
                     self._started = True
+
                     if not self._context_id:
                         self._context_id = str(uuid.uuid4())
-                    if not self.audio_context_available(self._context_id):
+                        logger.debug(f"{self}: Creating new context {self._context_id}")
```
markbackman: trace?

src/pipecat/services/inworld/tts.py
```diff
@@ -811,12 +871,19 @@ async def run_tts(self, text: str) -> AsyncGenerator[Frame, None]:
...
+                        await self.create_audio_context(self._context_id)
+                        await self._send_context(self._context_id)
+                        logger.debug(
+                            f"{self}: Context created and sent to server: {self._context_id}"
+                        )
+                    elif not self.audio_context_available(self._context_id):
+                        # Context exists on server but local tracking was removed
+                        logger.debug(f"{self}: Recreating local audio context {self._context_id}")
```
markbackman: trace?

src/pipecat/services/inworld/tts.py
```diff
@@ -811,12 +871,19 @@ async def run_tts(self, text: str) -> AsyncGenerator[Frame, None]:
...
                     yield TTSStartedFrame()
                     self._started = True
+
                     if not self._context_id:
                         self._context_id = str(uuid.uuid4())
-                    if not self.audio_context_available(self._context_id):
+                        logger.debug(f"{self}: Creating new context {self._context_id}")
+                        await self.create_audio_context(self._context_id)
+                        await self._send_context(self._context_id)
+                        logger.debug(
```
markbackman: trace?

markbackman (commented):
In testing 07ab-interruptible-inworld.py, this change breaks the text output from the TTS service. Let me see what change specifically is causing this...

You can see this too. Run the example and in the console you should see the TTSTextFrame containing each word as it's spoken. I don't see anything outputted which means the context has no update after the bot's turn. The frames should be be synchronized to the speech.

UPDATE:

I found the issue very quickly. I was running the new service which disables the word timestamps. The issue is that the `AudioContextWordTTSService` _must_ push text downstream. I think the safest route to go is to just enable word timestamps by default.

The alternative would be a separate class that is just an `InterruptibleTTSService`, but I think that's not ideal since developers will have a hard time figuring out which class to select.

src/pipecat/services/inworld/tts.py
cshape: For the new models (1.5-mini and 1.5-max) they will return an error if timestamps are enabled. This should be verifiable if they're selected and used in Pipecat.

But I think since we're delaying the model release, we can have timestamps enabled. Our timestamp solution for 1.5 should be ready by release.

src/pipecat/services/inworld/tts.py
ianbbqzy: Hi Mark, thanks for the thorough review!

I want to get some clarification here to have a better understanding. It makes sense to close a context if Pipecat uses contexts for some sort of interaction/turn management. So I think I see why we want to close it when interruption happens or when it's end of turn.

But I'm curious if there are cases where multiple `flush_context` will be sent to Inworld server automatically by Pipecat in one agent's turn. 

In my mind there are 3 ways to interact with Inworld WS endpoint assuming LLM sends back a one or a few tokens in each response chunk.

1. Pipecat has a sentence buffer/chunker. when it detects a sentence in the buffer, it calls `run_tts` which triggers `send_text` . Finally, Pipecat calls flush_audio on end of LLM response or interruption
2. Pipecat calls `run_tts` on each LLM response chunk, which triggers `send_text` . Finally, Pipecat calls `flush_audio` on end of LLM response or interruption
3. Pipecat calls `run_tts` on each LLM response chunk, which triggers `send_text` . Then the sentence buffer/chunker detects a sentence in the buffer, Pipecat calls `flush_audio` or something else that sends `flush_context` to Inworld server

Originally, I expected 3, but based on your comments, I assume that's not the case. So is it 1 or 2 or I have a completely wrong idea? Thanks!

markbackman:
This issue was filed:
https://github.com/pipecat-ai/pipecat/issues/3269

My assumption is that there's no input for a period of time which causes the session to disconnect (or go stale). Is there a keepalive message that can be sent to keep the websocket connected?

src/pipecat/services/inworld/tts.py
markbackman: > It makes sense to close a context if Pipecat uses contexts for some sort of interaction/turn management.

Yes, this is exactly it. Pipecat does use the context for turn tracking. Since there is no message to explicitly stop the generation, the context is used help Pipecat know which audio belongs to which turn. This is how interruptions work in the context of a websocket.

> But I'm curious if there are cases where multiple flush_context will be sent to Inworld server automatically by Pipecat in one agent's turn.

There would be one `flush_audio()` per turn, which normally comes at the end of the turn.

> In my mind there are 3 ways to interact with Inworld WS endpoint assuming LLM sends back a one or a few tokens in each response chunk.

Let me clarify the workflow:
- LLM streams output in this sequence:
  - `LLMFullResponseStartFrame` to signify the beginning of a turn (e.g. LLM stream)
  - `LLMTextFrame`s are send containing the streamed content from the LLM. These are pushed downstream as they're available.
  - `LLMFullResponseEndFrame` signifies the end of a turn (e.g. LLM stream or can be caused by an interruption).
- The TTS contains aggregation logic that looks for the LLM frame sequence and assembles content into sentences. As sentences are formed, they're provided to `run_tts()` for audio generation. This occurs until all text is received.
- For (`WordTTSService`s, which `InworldTTSService` inherits from) the TTS service flushes audio when the `LLMFullResponseEndFrame` is received. This signifies the end of the turn.

There's one other scenario that exists, which is when a `TTSSpeakFrame` is pushed to the TTS. This signifies an "ad-hoc" TTS generation created by the application. In this case, a frame with all of the text to be spoken is pushed. This results in `flush_audio()` being called. This also acts, in a way, as a single turn, but is different from the LLM streamed text scenario in that all audio is predetermined.

Does that help?

cshape:
Hi @markbackman -

I've got another commit here which addresses your PR feedback. I had to update the timestamp logic as well as it had some issues with long generations.

All seems to be functioning well now and I'm going to take a look at the best keepalive solution tomorrow regarding https://github.com/pipecat-ai/pipecat/issues/3269. Should have a commit addressing that tomorrow.

markbackman:
@cshape providing `enable_word_timestamps` is not an option. When set to False, the InworldTTSService produces no text output meaning that the conversation's context is not updated. This would mean that from the perspective of the bot, the assistant never says anything. To use the `AudioContextWordTTSService` class, you need to include word timestamps.

You have a few options:
1. Remove the `enable_word_timestamps` flag, always producing word-timestamp aligned data (RECOMMENDED!)
2. Never produce word-timestamp alignment data; this would mean changing to the `InterruptibleTTSService`, but this is a substantially inferior base class.
3. Write two different classes—one for word timestamps, one without word timestamps. This is just confusing for developers.

Our recommendations to teams is to always seek out TTS services that provide word-timestamp alignment data. These are the best services. I'd recommend option 1, so that `InworldTTSService` falls into that category.

src/pipecat/services/inworld/tts.py
```diff
@@ -594,6 +669,47 @@ def _get_websocket(self):
             return self._websocket
         raise Exception("Websocket not connected")
 
+    async def _keepalive_task_handler(self):
```
markbackman, line 672: Something about the keepalive task is causing the outputted text to be incomplete. My hunch is that the context ID is changing on the generation at some point, causing outputted json data to appear to be from a different turn. This is especially prone on longer generations. I tested this case by lowering the KEEPALIVE_SLEEP to 10 seconds to see what happens on this boundary condition.

Ideally, there would be a means to provide a keepalive that doesn't require a new context. I think, for now, we probably need to omit this as it will cause issues with the assistant context.

src/pipecat/services/inworld/tts.py
```diff
@@ -538,33 +570,69 @@ async def push_frame(self, frame: Frame, direction: FrameDirection = FrameDirect
...
+        if isinstance(frame, (TTSStoppedFrame, InterruptionFrame)):
             self._started = False
-            await self.add_word_timestamps([("Reset", 0)])
+            self._reset_timestamp_tracking()
+            if isinstance(frame, TTSStoppedFrame):
+                await self.add_word_timestamps([("Reset", 0)])
 
-    def _calculate_word_times(self, timestamp_info: Dict[str, Any]) -> List[Tuple[str, float]]:
+    def _reset_timestamp_tracking(self):
+        """Reset cumulative timestamp tracking state."""
```
markbackman, line 580: Why is cumulative time needed?

The timestamps returned by Inworld already reflect the cumulative time. And, I can see that the word-timestamp alignment is working well. This added logic just adds complexity to the code, so ideally we wouldn't need to include it.

markbackman: Ok, I've looked at this closer. I see that you have a combination of aggregate timestamps between chunks but then at some point reset the timing back to zero. Is there a reason why this is the case?

From an API perspective, this inconsistency makes handling the word-timestamp data more difficult. In fact, in testing, I see the words and audio become misaligned at points, which I would guess may be related to the timestamp resetting the internal aggregation.

The two options for timestamp aggregation would preferably be:
1. Always provide aggregate timestamps for a given context. Only reset the timestamps upon the start of a new context.
2. Always reset the timestamp for each chunk. With this approach, the application can maintain a consistent aggregation across turns.

Are either 1 or 2 possible? That would make this much easier for developers to work with.

markbackman:
Can we break up this PR into multiple PRs? I've taken the first few commits and broke them out into a smaller cleanup PR:
https://github.com/pipecat-ai/pipecat/pull/3288

These are your changes through ff3f8bb1d6d59a979a870557823577948cd71590, with some very minor clean up for simpler Python code. I added a second change which I mention in the PR.

These two commits we both agree on and will be beneficial for the service right away.

That leaves the following to be sorted out separately:
- timestamps: how to handle the proposal to enable/disable them
- keepalive: how can we send a keepalive without impacting the context aggregation?
- timestamp aggregation challenges: do we need to handle the hybrid timestamp aggregation?

cshape:
@markbackman 

> Can we break up this PR into multiple PRs? I've taken the first few commits and broke them out into a smaller cleanup PR: #3288
>
> These are your changes through [ff3f8bb](https://github.com/pipecat-ai/pipecat/commit/ff3f8bb1d6d59a979a870557823577948cd71590), with some very minor clean up for simpler Python code. I added a second change which I mention in the PR.
> 
> These two commits we both agree on and will be beneficial for the service right away.

Sounds good - thanks.

> That leaves the following to be sorted out separately:
> 
> * timestamps: how to handle the proposal to enable/disable them
> * keepalive: how can we send a keepalive without impacting the context aggregation?
> * timestamp aggregation challenges: do we need to handle the hybrid timestamp aggregation?

Timestamps: we can leave them enabled as we're actually cooking up a new timestamp solution which will work for our new 1.5 models and won't add latency.

Keepalive and timestamp aggregation: I'll work on cleaner implementations and submit new PRs.

markbackman:
> Timestamps: we can leave them enabled as we're actually cooking up a new timestamp solution which will work for our new 1.5 models and won't add latency.

Sounds good.

> Keepalive and timestamp aggregation: I'll work on cleaner implementations and submit new PRs.

Great. Ideally, you can build off of my in progress PR. I hope to have it merged down soon, though staffing is a bit light with the holidays.

markbackman:
FYI: This cleanup PR was merged:
https://github.com/pipecat-ai/pipecat/pull/3288

markbackman:
Closing since this has been superseded by two PRs:
- https://github.com/pipecat-ai/pipecat/pull/3288
- https://github.com/pipecat-ai/pipecat/pull/3403

PR #3261, opened 2025-12-18
Krisp VIVA SDK Filter and Turn support.

realgarik:
* The existing VIVA filter implementation doesn't handle the SDK global instantiation properly which can cause some memory leaks and unexpected behavior.
* Added Krisp VIVA Turn processor.
* Cover the case where 2 technologies can run in same process and will require sharing the single Krisp SDK instance.
* Added tests for both Filter and Turn.

### DIFF
--- a/src/pipecat/audio/filters/krisp_viva_filter.py
+++ b/src/pipecat/audio/filters/krisp_viva_filter.py
@@ -58,12 +58,6 @@ class KrispVivaFilter(BaseAudioFilter):
         """
         super().__init__()
 
-        # Acquire SDK reference (will initialize on first call)
-        try:
-            KrispVivaSDKManager.acquire()
-        except Exception as e:
-            raise RuntimeError(f"Failed to initialize Krisp SDK: {e}")
-
         try:
             # Set model path, checking environment if not specified
             if model_path:
@@ -92,47 +86,26 @@ class KrispVivaFilter(BaseAudioFilter):
                 raise FileNotFoundError(f"Model file not found: {self._model_path}")
 
             self._session = None
-            self._preload_session = None
             self._samples_per_frame = None
             self._noise_suppression_level = noise_suppression_level
             self._frame_duration_ms = frame_duration
             self._audio_buffer = bytearray()
-            self._sdk_acquired = True
             self._filtering = True
-
-            # Adding the model preload mechanism with default sample rate
-            # improves the latency of the session creation in the start() method.
-            self._preload_model()
 
         except Exception:
             # If initialization fails, release the SDK reference
             KrispVivaSDKManager.release()
             raise
 
-    def __del__(self):
-        """Release SDK reference when filter is destroyed."""
-        if hasattr(self, "_sdk_acquired") and self._sdk_acquired:
-            try:
-                if hasattr(self, "_session") and self._session is not None:
-                    self._session = None
-
-                if hasattr(self, "_preload_session") and self._preload_session is not None:
-                    self._preload_session = None
-
-                KrispVivaSDKManager.release()
-                self._sdk_acquired = False
-            except Exception as e:
-                logger.error(f"Error in __del__: {e}", exc_info=True)
-
     def _create_session(self, sample_rate: int, frame_duration: int):
-        """Preload the model with a specific sample rate.
+        """Create a Krisp session with a specific sample rate.
 
         Args:
-            sample_rate: Sample rate for preloading
+            sample_rate: Sample rate for the session
             frame_duration: Frame duration in milliseconds
 
         Raises:
-            Exception: If preloading fails
+            Exception: If session creation fails
         """
         try:
             model_info = krisp_audio.ModelInfo()
@@ -152,14 +125,6 @@ class KrispVivaFilter(BaseAudioFilter):
             logger.error(f"Failed to create Krisp session: {e}", exc_info=True)
             raise RuntimeError(f"Failed to create Krisp processing session: {e}") from e
 
-    def _preload_model(self):
-        """Preload the model with a specific sample rate."""
-        try:
-            self._preload_session = self._create_session(16000, self._frame_duration_ms)
-        except Exception as e:
-            logger.error(f"Failed to preload Krisp model: {e}", exc_info=True)
-            raise RuntimeError(f"Failed to preload Krisp model: {e}") from e
-
     async def start(self, sample_rate: int):
         """Initialize the Krisp processor with the transport's sample rate.
 
@@ -167,6 +132,8 @@ class KrispVivaFilter(BaseAudioFilter):
             sample_rate: The sample rate of the input transport in Hz.
         """
         try:
+            # Acquire SDK reference (will initialize on first call)
+            KrispVivaSDKManager.acquire()
             self._session = self._create_session(sample_rate, self._frame_duration_ms)
         except Exception as e:
             logger.error(f"Failed to start Krisp session: {e}", exc_info=True)
@@ -175,8 +142,13 @@ class KrispVivaFilter(BaseAudioFilter):
 
     async def stop(self):
         """Clean up the Krisp processor when stopping."""
-        self._session = None
-        self._audio_buffer.clear()
+        try:
+            self._session = None
+            self._audio_buffer.clear()
+            KrispVivaSDKManager.release()
+        except Exception as e:
+            logger.error(f"Error in stop: {e}", exc_info=True)
+            raise RuntimeError(f"Failed to stop Krisp processor: {e}") from e
 
     async def process_frame(self, frame: FilterControlFrame):
         """Process control frames to enable/disable filtering.
--- a/tests/test_krisp_viva_filter.py
+++ b/tests/test_krisp_viva_filter.py
@@ -108,8 +108,8 @@ class TestKrispVivaFilter(unittest.IsolatedAsyncioTestCase):
         """Test filter initialization with explicit model path."""
         filter_instance = KrispVivaFilter(model_path=self.model_path)
 
-        # Verify SDK was acquired
-        self.mock_sdk_manager.acquire.assert_called_once()
+        # Verify SDK was NOT acquired during initialization (happens in start())
+        self.mock_sdk_manager.acquire.assert_not_called()
 
         # Verify filter attributes
         self.assertEqual(filter_instance._model_path, self.model_path)
@@ -122,7 +122,8 @@ class TestKrispVivaFilter(unittest.IsolatedAsyncioTestCase):
         with patch.dict(os.environ, {"KRISP_VIVA_FILTER_MODEL_PATH": self.model_path}):
             filter_instance = KrispVivaFilter()
 
-            self.mock_sdk_manager.acquire.assert_called_once()
+            # Verify SDK was NOT acquired during initialization (happens in start())
+            self.mock_sdk_manager.acquire.assert_not_called()
             self.assertEqual(filter_instance._model_path, self.model_path)
 
     async def test_initialization_without_model_path(self):
@@ -132,7 +133,9 @@ class TestKrispVivaFilter(unittest.IsolatedAsyncioTestCase):
                 KrispVivaFilter()
 
             self.assertIn("Model path", str(context.exception))
-            # Verify SDK was released on error
+            # SDK acquire not called during initialization (happens in start())
+            # But release() is called in exception handler even though acquire() wasn't called
+            self.mock_sdk_manager.acquire.assert_not_called()
             self.mock_sdk_manager.release.assert_called_once()
 
     async def test_initialization_with_invalid_extension(self):
@@ -146,6 +149,9 @@ class TestKrispVivaFilter(unittest.IsolatedAsyncioTestCase):
                 KrispVivaFilter(model_path=tmp_path)
 
             self.assertIn(".kef extension", str(context.exception))
+            # SDK acquire not called during initialization (happens in start())
+            # But release() is called in exception handler even though acquire() wasn't called
+            self.mock_sdk_manager.acquire.assert_not_called()
             self.mock_sdk_manager.release.assert_called_once()
         finally:
             os.unlink(tmp_path)
@@ -155,6 +161,9 @@ class TestKrispVivaFilter(unittest.IsolatedAsyncioTestCase):
         with self.assertRaises(FileNotFoundError):
             KrispVivaFilter(model_path="/nonexistent/path/model.kef")
 
+        # SDK acquire not called during initialization (happens in start())
+        # But release() is called in exception handler even though acquire() wasn't called
+        self.mock_sdk_manager.acquire.assert_not_called()
         self.mock_sdk_manager.release.assert_called_once()
 
     async def test_initialization_with_custom_noise_level(self):
@@ -174,14 +183,17 @@ class TestKrispVivaFilter(unittest.IsolatedAsyncioTestCase):
 
         await filter_instance.start(16000)
 
+        # Verify SDK was acquired during start()
+        self.mock_sdk_manager.acquire.assert_called_once()
+
         # Verify session was created
         self.assertIsNotNone(filter_instance._session)
         self.assertEqual(filter_instance._current_sample_rate, 16000)
         self.assertEqual(filter_instance._samples_per_frame, 160)  # 16000 * 10ms / 1000
 
         # Verify NcSessionConfig was created and configured
-        # Note: Called twice - once for preload session, once for start()
-        self.assertEqual(self.mock_krisp_audio.NcSessionConfig.call_count, 2)
+        # Note: Called once in start() (no preload session anymore)
+        self.assertEqual(self.mock_krisp_audio.NcSessionConfig.call_count, 1)
         # Verify frame duration was set (hardcoded to 10ms in filter)
         self.assertEqual(self.mock_nc_cfg.inputFrameDuration, "10ms")
         # inputSampleRate and outputSampleRate are now set to the enum value
@@ -396,15 +408,16 @@ class TestKrispVivaFilter(unittest.IsolatedAsyncioTestCase):
             # Verify correct processing
             self.assertEqual(len(output_audio), len(input_audio))
 
-    async def test_destructor_cleanup(self):
-        """Test that destructor properly releases SDK reference."""
+    async def test_stop_releases_sdk(self):
+        """Test that stop() properly releases SDK reference."""
         filter_instance = KrispVivaFilter(model_path=self.model_path)
+        await filter_instance.start(16000)
 
-        # Call destructor
-        filter_instance.__del__()
+        # Stop the filter
+        await filter_instance.stop()
 
         # Verify SDK was released
-        self.mock_sdk_manager.release.assert_called()
+        self.mock_sdk_manager.release.assert_called_once()
 
     async def test_int_to_sample_rate_conversion(self):
         """Test sample rate conversion using the shared utility function."""
@@ -437,15 +450,24 @@ class TestKrispVivaFilter(unittest.IsolatedAsyncioTestCase):
         call_args = self.mock_session.process.call_args
         self.assertEqual(call_args[0][1], 75)  # Second argument should be the level
 
-    async def test_preload_session_created(self):
-        """Test that preload session is created during initialization."""
+    async def test_start_acquires_sdk(self):
+        """Test that start() acquires SDK reference and creates session."""
         filter_instance = KrispVivaFilter(model_path=self.model_path)
 
-        # Verify preload session was created
-        self.assertIsNotNone(filter_instance._preload_session)
+        # Verify no session exists before start
+        self.assertIsNone(filter_instance._session)
 
-        # Verify NcSessionConfig was created for preload and frame duration was set
-        self.mock_krisp_audio.NcSessionConfig.assert_called()
+        # Start the filter
+        await filter_instance.start(16000)
+
+        # Verify SDK was acquired
+        self.mock_sdk_manager.acquire.assert_called_once()
+
+        # Verify session was created
+        self.assertIsNotNone(filter_instance._session)
+
+        # Verify NcSessionConfig was created and frame duration was set
+        self.mock_krisp_audio.NcSessionConfig.assert_called_once()
         # Verify frame duration was set to 10ms (hardcoded in filter)
         self.assertEqual(self.mock_nc_cfg.inputFrameDuration, "10ms")
 


markbackman:
Tagging @filipi87 for when he's back from vacation.

src/pipecat/audio/filters/krisp_viva_filter.py
```diff
@@ -9,133 +9,168 @@
...
-        """Convert integer sample rate to krisp_audio SamplingRate enum.
+        # Acquire SDK reference (will initialize on first call)
+        try:
+            KrispVivaSDKManager.acquire()
+        except Exception as e:
+            raise RuntimeError(f"Failed to initialize Krisp SDK: {e}")
+
+        try:
+            # Set model path, checking environment if not specified
+            self._model_path = model_path or os.getenv("KRISP_VIVA_FILTER_MODEL_PATH")
```
filipi87: We should continue to support `KRISP_VIVA_MODEL_PATH` and, in this case, add a warning advising users to use `KRISP_VIVA_FILTER_MODEL_PATH` instead, so we don’t break backward compatibility for now.

I believe we will also need to update the Pipecat documentation to mention this. I’ll add it to my todo list.

examples/foundational/07p-interruptible-krisp-viva.py
```diff
@@ -44,19 +65,22 @@
         audio_in_enabled=True,
         audio_out_enabled=True,
         vad_analyzer=SileroVADAnalyzer(params=VADParams(stop_secs=0.2)),
-        audio_in_filter=KrispVivaFilter(),
+        turn_analyzer=KrispVivaTurn(params=KrispTurnParams()),
```
filipi87: With the recent changes inside Pipecat we should remove it from here, and add inside the `context_aggregator`. 

```suggestion
```

examples/foundational/07p-interruptible-krisp-viva.py
```diff
@@ -44,19 +65,22 @@
...
+        audio_in_filter=KrispVivaFilter(noise_suppression_level=KRISP_NOISE_SUPPRESSION_LEVEL),
     ),
     "twilio": lambda: FastAPIWebsocketParams(
         audio_in_enabled=True,
         audio_out_enabled=True,
         vad_analyzer=SileroVADAnalyzer(params=VADParams(stop_secs=0.2)),
-        audio_in_filter=KrispVivaFilter(),
+        turn_analyzer=KrispVivaTurn(params=KrispTurnParams()),
```
filipi87:
```suggestion
```

examples/foundational/07p-interruptible-krisp-viva.py
```diff
@@ -44,19 +65,22 @@
...
+        audio_in_filter=KrispVivaFilter(noise_suppression_level=KRISP_NOISE_SUPPRESSION_LEVEL),
     ),
     "webrtc": lambda: TransportParams(
         audio_in_enabled=True,
         audio_out_enabled=True,
         vad_analyzer=SileroVADAnalyzer(params=VADParams(stop_secs=0.2)),
-        audio_in_filter=KrispVivaFilter(),
+        turn_analyzer=KrispVivaTurn(params=KrispTurnParams()),
```
filipi87:
```suggestion
```

examples/foundational/07p-interruptible-krisp-viva.py
filipi87: Inside the `context_aggregator`, you will need to replace:
`bot=[TurnAnalyzerBotTurnStartStrategy(turn_analyzer=LocalSmartTurnAnalyzerV3())]`
for this:
`bot=[TurnAnalyzerBotTurnStartStrategy(turn_analyzer=KrispVivaTurn(params=KrispTurnParams()))]`

src/pipecat/audio/filters/krisp_viva_filter.py
```diff
@@ -9,133 +9,168 @@
...
-        # Audio buffer to accumulate samples for complete frames
-        self._audio_buffer = bytearray()
-
-    def _int_to_sample_rate(self, sample_rate):
-        """Convert integer sample rate to krisp_audio SamplingRate enum.
+        # Acquire SDK reference (will initialize on first call)
+        try:
+            KrispVivaSDKManager.acquire()
+        except Exception as e:
+            raise RuntimeError(f"Failed to initialize Krisp SDK: {e}")
```
filipi87: Cool. We had to do something similar with the previous `KrispFilter`.

For this one, we created this filter based on an example from Krisp, so I assumed Krisp was already handling this inside `globalInit`.

Nice improvement. Thank you for catching this.

src/pipecat/audio/filters/krisp_viva_filter.py
```diff
@@ -9,133 +9,168 @@
...
+            session = krisp_audio.NcInt16.create(nc_cfg)
+            return session
+        except Exception as e:
+            logger.error(f"Failed to create Krisp session: {e}", exc_info=True)
+            raise RuntimeError(f"Failed to create Krisp processing session: {e}") from e
+
+    def _preload_model(self):
+        """Preload the model with a specific sample rate."""
+        try:
+            self._preload_session = self._create_session(16000)
```
filipi87: It looks like we’re not using `self._preload_session` anywhere. So, I guess it is only to improve the latency when creating a new session at `start`, is that it ?

If the `sample_rate` remains the same as the default used to create `self._preload_session`, could we reuse that session in start instead of creating a new one ?

src/pipecat/audio/filters/krisp_viva_filter.py
```diff
@@ -9,133 +9,168 @@
...
-
-        self._samples_per_frame = int((sample_rate * self.FRAME_SIZE_MS) / 1000)
-        self._session = krisp_audio.NcInt16.create(nc_cfg)
+        try:
+            self._session = self._create_session(sample_rate)
+            self._state = self.State.STARTED
+        except Exception as e:
+            logger.error(f"Failed to start Krisp session: {e}", exc_info=True)
+            self._session = None
+            raise RuntimeError(f"Failed to create Krisp processing session: {e}") from e
```
filipi87, line 141: Do we need to also change the `self._state` again here ?

src/pipecat/audio/filters/krisp_viva_filter.py
```diff
@@ -144,7 +179,13 @@ async def process_frame(self, frame: FilterControlFrame):
             frame: The control frame containing filter commands.
         """
         if isinstance(frame, FilterEnableFrame):
-            self._filtering = frame.enable
+            if frame.enable == True:
```
filipi87:
```suggestion
            if frame.enable:
```

src/pipecat/audio/filters/krisp_viva_filter.py
```diff
@@ -144,7 +179,13 @@ async def process_frame(self, frame: FilterControlFrame):
...
+                if not self._state == self.State.STARTED:
+                    self._state = self.State.STARTING
+
+            elif frame.enable == False:
```
filipi87:
```suggestion
            elif not frame.enable:
```

src/pipecat/audio/filters/krisp_viva_filter.py
```diff
@@ -144,7 +179,13 @@ async def process_frame(self, frame: FilterControlFrame):
             frame: The control frame containing filter commands.
         """
         if isinstance(frame, FilterEnableFrame):
```
filipi87, line 159: The goal of `FilterEnableFrame` is to enable or disable audio filters at runtime. I think we have two possible approaches here:
- 01) Keep a simple approach, as we did before, by tracking whether filtering is enabled using a variable. In this case we are not changing going to destroy or recreate the `self._session`.
- 02) Explicitly delete and recreate the self._session when enabling or disabling the filter.

Right now, we’re changing the state to `STOPPING`, and then inside the filter we try to stop the session while also recreating a new one.

When the filter is enabled, we only change the state to STARTING, and it seems to work because a new session is created during STOPPING.

So, this flow is a bit confusing, and I guess we could simplify it using one of the approaches that I mentioned above.

src/pipecat/audio/filters/krisp_viva_filter.py
```diff
@@ -155,39 +196,64 @@ async def filter(self, audio: bytes) -> bytes:
...
 
-        # Calculate how many complete frames we can process
-        total_samples = len(self._audio_buffer) // 2  # 2 bytes per int16 sample
-        num_complete_frames = total_samples // self._samples_per_frame
+            async def stop():
+                if self._state == self.State.STOPPING:
+                    self._state = self.State.STOPPED
+                    self._session = None
+                    self._audio_buffer.clear()
+                    self._session = self._create_session(self._current_sample_rate)
```
filipi87: I don't think we should handle this here, based [on this](https://github.com/pipecat-ai/pipecat/pull/3261/files#r2665380070).

src/pipecat/audio/turn/krisp_viva_turn.py
```diff
@@ -0,0 +1,368 @@
...
+            self._frame_probabilities = []  # Store all probabilities from last append_audio call
+
+            # Create session with provided sample rate or default to 16000 Hz
+            # This preloads the model to improve latency when set_sample_rate is called later
+            preload_sample_rate = sample_rate if sample_rate else 16000
+            try:
+                self._preload_tt_session = self._create_tt_session(preload_sample_rate)
+                self._tt_session = self._preload_tt_session
+                # Calculate samples per frame for the preloaded session
+                if preload_sample_rate:
```
filipi87: We can remove this `if`, since we are always initializing `preload_sample_rate` with a default value above.

src/pipecat/audio/turn/krisp_viva_turn.py
```diff
@@ -0,0 +1,368 @@
...
+                    self._samples_per_frame = int(
+                        (preload_sample_rate * self._params.frame_duration_ms) / 1000
+                    )
+            except Exception as e:
+                logger.error(f"Failed to create turn detection session: {e}", exc_info=True)
+                self._preload_tt_session = None
```
filipi87, line 126: Shouldn't we `raise` the exception here as well ?

src/pipecat/audio/turn/krisp_viva_turn.py
```diff
@@ -0,0 +1,368 @@
...
+                    speech_duration = (
+                        current_time - self._speech_start_time if self._speech_start_time else 0
+                    )
+                    eot_latency = current_time - self._eot_start_time if self._eot_start_time else 0
+
+                    # logger.debug(
+                    #     f"Turn complete: prob={prob:.3f}, speech={speech_duration:.1f}s, "
+                    #     f"eot_latency={eot_latency:.3f}s"
+                    # )
+
```
filipi87: It looks like `speech_duration` and `eot_latency` are not being used anywhere, so we can probably remove all of this code.
If we do that, we can also remove all code related to `self._eot_start_time` and `self._speech_start_time`. 🙂

src/pipecat/audio/turn/krisp_viva_turn.py
```diff
@@ -0,0 +1,368 @@
...
+        """
+        # For real-time processing, the state is determined in append_audio
+        # This method can be used for periodic checks if needed
+        return EndOfTurnState.INCOMPLETE, None
+
+    def clear(self):
+        """Reset the turn analyzer to its initial state."""
+        self._clear(EndOfTurnState.COMPLETE)
+
+    def _clear(self, turn_state: EndOfTurnState):
```
filipi87, line 341: It looks like we never invoke `_clear` with a state other than `EndOfTurnState.COMPLETE`. Do we really need this ? Could we just implement `clear` and reset the state assuming it’s always `COMPLETE`?

scripts/krisp/audio_file_utils.py
filipi87, line 1: Since we’re not using this as a test, but instead as a standalone script, I think we should move it to `/scripts/krisp`.

That would mean moving these files to there:
- /tests/integration/audio_file_utils.py
- /tests/integration/test_krisp_viva_filter_audiofile.py
- /tests/integration/test_krisp_viva_turn_audiofile.py

Copying @markbackman for a second opinion on this.

src/pipecat/audio/turn/krisp_viva_turn.py
```diff
@@ -0,0 +1,368 @@
...
+
+    async def analyze_end_of_turn(self) -> Tuple[EndOfTurnState, Optional[MetricsData]]:
+        """Analyze the current audio state to determine if turn has ended.
+
+        Returns:
+            Tuple containing the end-of-turn state and optional metrics data.
+        """
+        # For real-time processing, the state is determined in append_audio
+        # This method can be used for periodic checks if needed
+        return EndOfTurnState.INCOMPLETE, None
```
filipi87: Would it make sense to save the last `state` from `appendAudio` and return it here instead of always returning it as `INCOMPLETE` ?

tests/integration/test_krisp_viva_filter_integration.py
```diff
@@ -0,0 +1,430 @@
...
+# Note: krisp_audio is a private/proprietary module, so these tests are skipped
+# in CI/CD environments where it's not available. Developers with access to
+# krisp_audio can run these tests locally.
+try:
+    import krisp_audio
+
+    KRISP_AVAILABLE = True
+except (ImportError, ModuleNotFoundError):
+    KRISP_AVAILABLE = False
+
```
filipi87: Considering that we’re essentially mocking the entire `krisp_audio` module below, do we still need this availability check ?
If not, we could run these tests on CI, or even on machines where the module isn’t installed.

tests/integration/test_krisp_viva_turn_integration.py
```diff
@@ -0,0 +1,489 @@
...
+# Check if krisp_audio is available - skip tests if not
+# Note: krisp_audio is a private/proprietary module, so these tests are skipped
+# in CI/CD environments where it's not available. Developers with access to
+# krisp_audio can run these tests locally.
+try:
+    import krisp_audio
+
+    KRISP_AVAILABLE = True
+except (ImportError, ModuleNotFoundError):
+    KRISP_AVAILABLE = False
```
filipi87: Same here:
Considering that we’re essentially mocking the entire `krisp_audio` module below, do we still need this availability check ?
If not, we could run these tests on CI, or even on machines where the module isn’t installed.

filipi87: Also, I believe that due to recent changes in Pipecat, where the turn analyzer is now used inside `TurnStartStrategies`, and executed inside `context_aggregator`, this test is no longer working.

tests/test_krisp_sdk_manager.py
```diff
@@ -0,0 +1,86 @@
...
+
+
+class TestKrispVivaSDKManager:
+    """Tests for KrispVivaSDKManager singleton."""
+
+    def test_reference_counting(self):
+        """Test that SDK manager properly tracks references."""
+        # Initial state
+        initial_count = KrispVivaSDKManager.get_reference_count()
+        initial_initialized = KrispVivaSDKManager.is_initialized()
```
filipi87: It looks like this isn’t used anywhere.
```suggestion
```

tests/test_krisp_viva_filter.py
filipi87, line 1: Same here:
Considering that we’re essentially mocking the entire `krisp_audio` module below, do we still need this availability check ?
If not, we could run these tests on CI, or even on machines where the module isn’t installed.

tests/test_krisp_viva_turn.py
filipi87, line 1: Same here:
Considering that we’re essentially mocking the entire `krisp_audio` module below, do we still need this availability check ?
If not, we could run these tests on CI, or even on machines where the module isn’t installed.

scripts/krisp/audio_file_utils.py
markbackman: +1 for moving to `/scripts/krisp`.

src/pipecat/audio/filters/krisp_viva_filter.py
realgarik: self._preload_session is used to improve the latency of creating the session.
If sample rate remains the same, in general we can reuse it, but once we used it, it will remain the state inside and on the next new audio stream processing it will start with invalid state impacting the current processing, so batter to keep it for pre-load only.

src/pipecat/audio/turn/krisp_viva_turn.py
realgarik: This was done for debugging, will remove.

src/pipecat/audio/filters/krisp_viva_filter.py
realgarik: I think doesn't make sense, as it was failed to start and the state still not updated to started.

src/pipecat/audio/filters/krisp_viva_filter.py
realgarik: this was done to reduce the latency on start

tests/test_krisp_viva_filter.py
```diff
@@ -0,0 +1,752 @@
...
+    # If import fails due to missing krisp_audio, mark as unavailable
+    if "krisp_audio" in str(e) or "Missing module" in str(e):
+        KRISP_AVAILABLE = False
+        KrispVivaFilter = None
+        FilterEnableFrame = None
+    else:
+        raise
+
+
+@unittest.skipIf(not KRISP_AVAILABLE, "krisp_audio module not available (private dependency)")
```
filipi87: It is skipping all the tests. I guess we need to remove all these lines that check whether `KRISP_AVAILABLE`.

filipi87: The same thing is happening for `test_krisp_viva_turn_integration.py` and `test_krisp_viva_turn.py`.

Also, I believe that due to recent changes in Pipecat, where the turn analyzer is now used inside `TurnStartStrategies`, and executed inside `context_aggregator`, the `test_krisp_viva_turn_integration.py` test is no longer working. So, it would be nice to execute and double check it is working as expected.

examples/foundational/07p-interruptible-krisp-viva.py
```diff
@@ -4,14 +4,32 @@
...
+"""
 
 import os
 
 from dotenv import load_dotenv
 from loguru import logger
 
 from pipecat.audio.filters.krisp_viva_filter import KrispVivaFilter
-from pipecat.audio.turn.smart_turn.local_smart_turn_v3 import LocalSmartTurnAnalyzerV3
+from pipecat.audio.interruptions.turn_analyzer_bot_turn_start_strategy import TurnAnalyzerBotTurnStartStrategy
```
filipi87: This import doesn't exist anymore.
```suggestion
```

examples/foundational/07p-interruptible-krisp-viva.py
```diff
@@ -82,7 +104,7 @@ async def run_bot(transport: BaseTransport, runner_args: RunnerArguments):
         context,
         user_params=LLMUserAggregatorParams(
             user_turn_strategies=UserTurnStrategies(
-                stop=[TurnAnalyzerUserTurnStopStrategy(turn_analyzer=LocalSmartTurnAnalyzerV3())]
+                stop=[TurnAnalyzerUserTurnStopStrategy(turn_analyzer=KrispVivaTurn(params=KrispTurnParams())]
```
filipi87: It is missing one ")"
```suggestion
                stop=[TurnAnalyzerUserTurnStopStrategy(turn_analyzer=KrispVivaTurn(params=KrispTurnParams()))]
```

filipi87:
Hi @realgarik, you need to execute this script: ./scripts/fix-ruff.sh

It validates the expected Pipecat format and docstrings for the files. Currently, there are 18 errors when I run it against this branch.

scripts/krisp/test_krisp_viva_filter_audiofile.py
```diff
@@ -0,0 +1,284 @@
...
+        output_path: Path to save filtered audio
+        noise_suppression_level: Noise suppression level (0-100)
+        frame_duration_ms: Frame duration in milliseconds (for chunking input)
+        verbose: Show detailed processing information
+    """
+    # Read and convert audio file
+    audio_data, sample_rate = read_audio_file(input_path, verbose=True)
+
+    # Load filter to get supported sample rates
+    FilterClass, model_path = load_filter()
```
filipi87: Is there any reason you can’t create the `KrispVivaFilter` class directly here instead of loading it this way ? 
It feels like this dynamic loading is just adding unnecessary complexity. It would simplify this script a lot.🙂

scripts/krisp/test_krisp_viva_turn_audiofile.py
```diff
@@ -0,0 +1,357 @@
...
+
+# Add src directory to Python path for development environment
+script_dir = Path(__file__).parent
+project_root = script_dir.parent.parent
+src_dir = project_root / "src"
+if src_dir.exists() and str(src_dir) not in sys.path:
+    sys.path.insert(0, str(src_dir))
+
+
+def load_turn_analyzer():
```
filipi87: The same thing here, maybe we could remove this dynamic loading.

src/pipecat/audio/filters/krisp_viva_filter.py
```diff
@@ -9,133 +9,180 @@
...
+                if hasattr(self, "_session") and self._session is not None:
+                    self._session = None
+
+                if hasattr(self, "_preload_session") and self._preload_session is not None:
+                    self._preload_session = None
+
+                KrispVivaSDKManager.release()
+                self._sdk_acquired = False
+            except Exception as e:
+                logger.error(f"Error in __del__: {e}", exc_info=True)
```
filipi87: In Pipecat, we usually clean up all resources inside the `stop` method. In all services, that’s where we add this kind of logic.

It would be nice to move the logic for releasing `KrispVivaSDKManager` there and remove this code.

Copying @aconchillo, @markbackman for a second opinion.

src/pipecat/audio/turn/krisp_viva_turn.py
```diff
@@ -0,0 +1,342 @@
...
+                # Clean up session first
+                if hasattr(self, "_tt_session") and self._tt_session is not None:
+                    self._tt_session = None
+                if hasattr(self, "_preload_tt_session") and self._preload_tt_session is not None:
+                    self._preload_tt_session = None
+
+                KrispVivaSDKManager.release()
+                self._sdk_acquired = False
+            except Exception as e:
+                logger.error(f"Error in __del__: {e}", exc_info=True)
```
filipi87, line 149: In Pipecat, we usually clean up all resources inside the `stop` method. In all services, that’s where we add this kind of logic.

It would be nice to move the logic for releasing `KrispVivaSDKManager` there and remove this code.

Copying @aconchillo, @markbackman for a second opinion.

src/pipecat/audio/turn/krisp_viva_turn.py
```diff
@@ -0,0 +1,342 @@
...
+            for frame in frames:
+                if is_speech:
+                    # Track speech start time
+                    if not self._speech_triggered:
+                        logger.trace("Speech detected, turn analysis started")
+                    self._speech_triggered = True
+                else:
+                    # Track when speech stops (potential end-of-turn)
+                    if self._speech_triggered:
+                        state = EndOfTurnState.COMPLETE
```
filipi87: It feels like this `else` should exist. Otherwise, we are effectively changing the state to complete as soon as we’ve previously received some speech and then stop receiving it. Or am I understanding it wrong ?

It looks like it treats the turn as complete as soon as there is silence after prior speech.

src/pipecat/audio/filters/krisp_viva_filter.py
filipi87: Hi @realgarik, I don’t think we’re gaining any real advantage from introducing all these extra states in this PR:

```
STOPPING = "STOPPING"
STARTING = "STARTING"
STARTED = "STARTED"
STOPPED = "STOPPED"
```

In Pipecat, the filter is effectively either started (when `start` is called) or stopped (when `stop` is called).

Only after `start` is invoked does the `filter` method begin receiving audio.

And Pipecat will first start all the services in the Pipeline before it is ready for processing the audio. So I am not even sure if the the `self._preload_session` would be needed.

Inside the `filter` itself, we really just need to decide whether or not we should be filtering audio based on the `FilterEnableFrame` received in `process_frame`.

If the goal here is to reduce latency, and if pausing or resuming the filter via `FilterEnableFrame` requires creating a new `self._session` to avoid carrying over any remaining state, I think we could simplify this considerably with something like:

```
async def process_frame(self, frame: FilterControlFrame):
    """Process control frames to enable/disable filtering.

    Args:
        frame: The control frame containing filter commands.
    """
    if isinstance(frame, FilterEnableFrame):
        self._filtering = frame.enable
        if not self._filtering:
            self._audio_buffer.clear()
            self._session = self._create_session(self._current_sample_rate)
```


This would guarantee that a new session is preloaded every time the audio filter is paused, while keeping the filter logic essentially the same as before, only checking whether filtering is enabled:

```
if not self._filtering:
    return audio
```

This approach feels simpler and avoids the additional state complexity.

src/pipecat/audio/filters/krisp_viva_filter.py
realgarik: @filipi87 We call KrispVivaSDKManager.acquire() on __init__() and have to release it in __delete__().
If I move all this logic in stop() we have to move all the initialization logic into start() which will add more delay and break the model pre-load logic which reduces load time ~100ms -> ~15ms on start.

src/pipecat/audio/turn/krisp_viva_turn.py
realgarik: @filipi87 We call KrispVivaSDKManager.acquire() on init() and have to release it in delete().
If I move all this logic in stop() we have to move all the initialization logic into start() which will add more delay and break the model pre-load logic which reduces load time ~100ms -> ~15ms on start.

src/pipecat/audio/filters/krisp_viva_filter.py
realgarik: Will revert this logic and make it similar to current main branch solution using:
```
if isinstance(frame, FilterEnableFrame):
            self._filtering = frame.enable
```

src/pipecat/audio/filters/krisp_viva_filter.py
realgarik: @filipi87 will revert this part to be the same as on main branch.
The pre-load model itself doesn't break the logic, so I would suggest to keep it. Even though, all the services should be started before the stream processing begins, I think it's good to have a quick starting filter.

scripts/krisp/test_krisp_viva_filter_audiofile.py
realgarik: no, will change!

scripts/krisp/test_krisp_viva_turn_audiofile.py
realgarik: will change.

tests/test_krisp_viva_filter.py
realgarik: I just removed these tests.

tests/test_krisp_viva_turn.py
realgarik: removed the test

tests/integration/test_krisp_viva_turn_integration.py
realgarik: TT is low priority for us, so removed some tests.

src/pipecat/audio/filters/krisp_viva_filter.py
filipi87: Hi @realgarik, as it is now, in all my tests the `__del__` method is never invoked, so we always end up using the fallback and see the warning "Force cleaning up Krisp SDK at exit" inside the `krisp_instance`.

When we start the Pipeline, I don’t think it will make much difference whether this load time happens inside `__init__` or inside `start`, the total time for the Pipeline to be ready to receive audio will probably be the same.

If we move the logic into `start` and `stop`, we can ensure that resources are always released properly, without having to rely on the fallback mechanism. And it will match the behavior that we already have for all the other Pipecat services.

The code would be like this:
```
    async def start(self, sample_rate: int):
        """Initialize the Krisp processor with the transport's sample rate.

        Args:
            sample_rate: The sample rate of the input transport in Hz.
        """
        try:
            # Acquire SDK reference (will initialize on first call)
            KrispVivaSDKManager.acquire()
            self._session = self._create_session(sample_rate, self._frame_duration_ms)
        except Exception as e:
            logger.error(f"Failed to start Krisp session: {e}", exc_info=True)
            self._session = None
            raise RuntimeError(f"Failed to create Krisp processing session: {e}") from e

    async def stop(self):
        """Clean up the Krisp processor when stopping."""
        self._session = None
        self._audio_buffer.clear()
        KrispVivaSDKManager.release()
```

And we wouldn't even need the `self._preload_session` anymore, since we are not going to be able to actually create a session until we have started the SDK.

But tagging @markbackman, @kompfner or @aconchillo for a second opinion on this.

src/pipecat/audio/turn/krisp_viva_turn.py
filipi87: Let's keep this discussion [here](https://github.com/pipecat-ai/pipecat/pull/3261#discussion_r2673533551), so what we decide we apply for both, the krisp filter and the turn.

examples/foundational/07p-interruptible-krisp-viva.py
filipi87, line 1: I believe that after the last merge, you accidentally removed the KrispVivaTurn😅.

src/pipecat/audio/turn/krisp_viva_turn.py
realgarik: Sure!

src/pipecat/audio/filters/krisp_viva_filter.py
realgarik: @filipi87 thanks for detailed explanation.
Based on your feedback, I'll remove the preload logic and move the init/del logic into start/stop.

examples/foundational/07p-interruptible-krisp-viva.py
realgarik: @filipi87 I removed some tests.

src/pipecat/audio/turn/krisp_viva_turn.py
realgarik: @filipi87 for TT there is no start/stop API. Do we need move the SDK init into other place than __init__/__del__?

PR #3263, opened 2025-12-19
Deprecate `OpenAILLMContext` and associated things

kompfner:
Still need to actually test these changes, but going ahead and opening the PR for review.

### DIFF
--- a/examples/foundational/14e-function-calling-google.py
+++ b/examples/foundational/14e-function-calling-google.py
@@ -104,7 +104,7 @@ async def run_bot(transport: BaseTransport, runner_args: RunnerArguments):
         voice_id="71a7ad14-091c-4e8e-a314-022ece01c121",  # British Reading Lady
     )
 
-    llm = GoogleLLMService(api_key=os.getenv("GOOGLE_API_KEY"), model="gemini-2.0-flash-001")
+    llm = GoogleLLMService(api_key=os.getenv("GOOGLE_API_KEY"))
     llm.register_function("get_weather", get_weather)
     llm.register_function("get_image", get_image)
     llm.register_function("get_restaurant_recommendation", fetch_restaurant_recommendation)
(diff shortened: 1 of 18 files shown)


kompfner:
Seeing if I can simplify this change, adding the deprecation warning in just a couple targeted superclasses rather than sprinkling it everywhere. Then when we finally get around to removing the superclasses, we can just follow the inheritance chain and get rid of all subclasses too.

src/pipecat/services/openai_realtime_beta/openai.py
```diff
@@ -780,7 +780,21 @@ def create_context_aggregator(
             OpenAIContextAggregatorPair: A pair of context aggregators, one for
             the user and one for the assistant, encapsulated in an
             OpenAIContextAggregatorPair.
+
+        .. deprecated:: 0.0.99
```
kompfner: maybe we don't bother with this deprecation as the whole OpenAI Realtime Beta class is deprecated...when we finally remove the OpenAILLMContext we remove OpenAI Realtime Beta too

examples/foundational/14e-function-calling-google.py
```diff
@@ -104,7 +104,7 @@ async def run_bot(transport: BaseTransport, runner_args: RunnerArguments):
         voice_id="71a7ad14-091c-4e8e-a314-022ece01c121",  # British Reading Lady
     )
 
-    llm = GoogleLLMService(api_key=os.getenv("GOOGLE_API_KEY"), model="gemini-2.0-flash-001")
+    llm = GoogleLLMService(api_key=os.getenv("GOOGLE_API_KEY"))
```
kompfner, line 107: Just a thing I found along the way; in a brief test 2.0 wasn't really working well

kompfner: Might be worth updating this everywhere (maybe in another PR)

kompfner:
OK, greatly _reduced_ the number of places where we print a deprecation warning, and greatly _increased_ the number of places where we mention deprecation in a docstring.

PR #3265, opened 2025-12-19
Add support for a VonageFrameSerializer

markbackman:

Updates the FastAPIWebsocketTransport to support a new `MIXED` mode where websocket messages may be a mix of text or binary messages, as is require by Vonage.

Adds a new VonageFrameSerializer to connect a Vonage Voice API call over a websocket.

Updates the development runner to handle the websocket messages, identify the call, parse query params in the websocket URL to make the to and from number available to the bot.

FYI: Example implementation here:
https://github.com/pipecat-ai/pipecat-examples/pull/133

### DIFF
--- a/changelog/3265.changed.2.md
+++ b/changelog/3265.changed.2.md
@@ -1,0 +1,1 @@
+- Updated the development runner utility to support parsing `VonageFrameSerializer` websocket messages and handling/passing the to and from numbers via the `RunnerArguments`.
--- a/src/pipecat/runner/utils.py
+++ b/src/pipecat/runner/utils.py
@@ -88,6 +88,11 @@ def _detect_transport_type_from_message(message_data: dict) -> str:
         logger.trace("Auto-detected: EXOTEL")
         return "exotel"
 
+    # Vonage detection
+    if message_data.get("event") == "websocket:connected" and "content-type" in message_data:
+        logger.trace("Auto-detected: VONAGE")
+        return "vonage"
+
     logger.trace("Auto-detection failed - unknown format")
     return "unknown"
 
@@ -135,6 +140,16 @@ async def parse_telephony_websocket(websocket: WebSocket):
                 "to": str,
             }
 
+        - Vonage::
+
+            {
+                "content_type": str,  # e.g., "audio/l16;rate=16000"
+                "from": str,
+                "to": str,
+                "call_uuid": str,
+                "conversation_uuid": str,
+            }
+
     Example usage::
 
         transport_type, call_data = await parse_telephony_websocket(websocket)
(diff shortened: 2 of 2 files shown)


markbackman:
Questions I have:
- What should the default sample rate be for Vonage? Via phone, audio is reduce to 8khz for the most part. This is the case on my calls too. Some carriers support higher sample rates though. Pipecat will resample as needed to match the desired sample rate, so we just need to figure out what that is.
- Just to make 100% sure, the NCCO can't be hosted on Vonage, right? In my example, I have a simple FastAPI server that handles the webhook (which is also used to get the to and from phone number info) and returns the NCCO to initiate the webhook.
- Is there any way to get identifying information from the webhook messages themselves? I don't think so, but just wanted to be certain. In my example, I use the connection webhook to parse the calling information and pass that to the bot via query params.

To dos:
- [ ] Resampling is choppy. Need to revisit. @aconchillo maybe you could spare some time to help me sort this out?
- [x] Add DTMF support

varunps2003:
> Questions I have:
> 
> * What should the default sample rate be for Vonage? Via phone, audio is reduce to 8khz for the most part. This is the case on my calls too. Some carriers support higher sample rates though. Pipecat will resample as needed to match the desired sample rate, so we just need to figure out what that is.
> * Just to make 100% sure, the NCCO can't be hosted on Vonage, right? In my example, I have a simple FastAPI server that handles the webhook (which is also used to get the to and from phone number info) and returns the NCCO to initiate the webhook.
> * Is there any way to get identifying information from the webhook messages themselves? I don't think so, but just wanted to be certain. In my example, I use the connection webhook to parse the calling information and pass that to the bot via query params.
> 
> To dos:
> 
> * [ ]  Resampling is choppy. Need to revisit. @aconchillo maybe you could spare some time to help me sort this out?
> * [x]  Add DTMF support

Hi @markbackman below are some answers to your questions:
1) For Vonage Video Audio Connector, the default sample rate should be 16 kHz. The Vonage docs and WebSocket examples default to audio/l16;rate=16000 with 20 ms / 640-byte frames.
While PSTN audio is often 8 kHz, Audio Connector streams audio from WebRTC video sessions, where 16 kHz is the default. We can support 8 kHz via configuration when explicitly requested, and Pipecat can resample as needed.
2) NCCO is a Vonage Voice API concept and doesn’t apply to Vonage Video Audio Connector. In the Audio Connector flow, there is no NCCO or Voice webhook — the WebSocket is initiated via the /connect REST API against an existing Video session. - https://developer.vonage.com/en/video/guides/audio-connector

varunps2003:
Hi @markbackman , 
if a few additional changes can be incorporated into fastapi.py, we can avoid maintaining a separate custom Vonage transport altogether. With those changes in place, Vonage Audio Connector integrations can be built and run using the existing FastAPIWebsocketTransport only.

I’ve already validated this approach in an example PR here:
https://github.com/pipecat-ai/pipecat-examples/pull/146

Once the additional updates are applied to fastapi.py, that example should work cleanly without any custom transport code.

The proposed changes to be included in fastapi.py are:
```
diff --git a/src/pipecat/transports/websocket/fastapi.py b/src/pipecat/transports/websocket/fastapi.py
index ef7f6fc0..f3cb11d3 100644
--- a/src/pipecat/transports/websocket/fastapi.py
+++ b/src/pipecat/transports/websocket/fastapi.py
@@ -376,6 +376,13 @@ class FastAPIWebsocketOutputTransport(BaseOutputTransport):
         self._send_interval = 0
         self._next_send_time = 0
 
+        # Buffer for protocol-level audio packetization.
+        # Some serializers (e.g. Vonage) may return arbitrary-sized PCM blobs,
+        # but the downstream protocol requires fixed-size audio frames.
+        # This buffer accumulates outgoing audio until a full protocol frame
+        # (e.g. 20ms / 640 bytes for 16kHz PCM16 mono) can be sent.
+        self._audio_send_buffer = bytearray()
+
         # Whether we have seen a StartFrame already.
         self._initialized = False
 
@@ -496,6 +503,19 @@ class FastAPIWebsocketOutputTransport(BaseOutputTransport):
         try:
             payload = await self._params.serializer.serialize(frame)
             if payload:
+                # Protocol safety: some WebSocket audio backends (e.g. Vonage Audio Connector)
+                # require audio to be sent as fixed-size PCM frames (20ms @ 16kHz = 640 bytes).
+                if isinstance(payload, (bytes, bytearray)):
+                    FRAME_BYTES = 640  # 20ms @ 16kHz mono PCM16
+                    self._audio_send_buffer.extend(payload)
+
+                    # send only full 20ms frames
+                    while len(self._audio_send_buffer) >= FRAME_BYTES:
+                        chunk = bytes(self._audio_send_buffer[:FRAME_BYTES])
+                        del self._audio_send_buffer[:FRAME_BYTES]
+                        await self._client.send(chunk)
+                    return
+
                 await self._client.send(payload)
         except Exception as e:
             logger.error(f"{self} exception sending data: {e.__class__.__name__} ({e})")
```

markbackman:
Closing for now. The FastAPIWebsocketTransport change was superseded by: https://github.com/pipecat-ai/pipecat/pull/3379. Holding off for now on a VonageFrameSerializer.

PR #3270, opened 2025-12-19
Fix Inworld WebSocket TTS recovery on inactivity pause

hwuiwon:
- Detect SESSION_PAUSED from Inworld and treat it as a recoverable disconnect.
- Restart the receive loop on closed/error connections and reconnect after inactivity.
- Guard close_context to only send when the websocket is open to avoid cascading close errors.

### DIFF
--- a/changelog/3270.fixed.md
+++ b/changelog/3270.fixed.md
@@ -1,0 +1,2 @@
+- Fixed `InworldTTSService` recovery when the websocket session is paused due to inactivity.
+- Improved reconnect behavior for Inworld WebSocket TTS after unexpected close.


hwuiwon:
@markbackman would appreciate a review on this one!

cc @cshape to make sure that this is the correct fix from the inworld side

markbackman:
FYI: We're already working through improvements here:
https://github.com/pipecat-ai/pipecat/pull/3256

Rather than respond to the error, preemptively sending a keepalive would be the right solution. I'm not sure if that's available from Inworld, but if it is, that's a better solution.

hwuiwon:
@markbackman Oh I missed it! Sounds great, I'll close this one then

PR #3276, opened 2025-12-20
GrokRealtimeLLMService cleanup

markbackman:

Improvements to how the SessionProperty handling is done:
- sample rates:
  - remove the `sample_rate` arg
  - Add logic to detect sample_rate from StartFrame or explicitly setting in SessionProperties
- voice: remove from args and rely on SessionProperties
- turn_detection: make `"server_vad"` the default as it's the only available value and must be set

Implement `response.function_call_arguments.delta` and corresponding class to avoid a warning when functions are called.

Added 51 foundational to evals.

Update README's services list with Grok Voice Agent.

No changelog needed as this hasn't been released yet.


aconchillo:
LGTM

kompfner:
I've recently noticed that using Grok Realtime with SmallWebrtc (i.e. running `uv run python ./examples/foundational/51-grok-realtime.py`) isn't working due to seemingly a sample rate issue.

As soon as the user starts speaking, we get this error:
```
{"type":"error","event_id":"e5845cc2-189e-4f56-80fd-567a7cfa9b47","params":"Multiple frame rates in the concatenated frames: {24000, 16000}","code":"internal_error","message":"ValueError","previous_item_id":null}
```

Currently investigating; stay tuned.

markbackman:
> I've recently noticed that using Grok Realtime with SmallWebrtc (i.e. running `uv run python ./examples/foundational/51-grok-realtime.py`) isn't working due to seemingly a sample rate issue.
> 
> As soon as the user starts speaking, we get this error:
> 
> ```
> {"type":"error","event_id":"e5845cc2-189e-4f56-80fd-567a7cfa9b47","params":"Multiple frame rates in the concatenated frames: {24000, 16000}","code":"internal_error","message":"ValueError","previous_item_id":null}
> ```
> 
> Currently investigating; stay tuned.

Good catch! I noticed this last night. Thanks for looking into it 🙏

kompfner:
OK, what I discovered so far is that the issue existed even before this PR, but it was being masked by the fact that we weren't enforcing always sending an `AudioConfiguration` to the server. After you added that enforcement to this PR, it seems like any deviation of audio rate from 24000 triggers an error in Grok.

What was happening before this PR is that user-specified `sample_rate` just wasn't being received/honored by the server.

kompfner:
It seems to me that we have a deeper issue possibly in the flow of how the service is even being initialized. I suspect internally it's starting out expecting an input sample rate of 24000 and then when we tell it otherwise it panics.

kompfner:
> any deviation of audio rate from 24000 triggers an error in Grok

Hmm, this isn't quite right, either—when using the Daily transport the default sample rate is also 16000, and there are no errors when use that transport...

kompfner:
OK I think I figured it out!

kompfner:
> OK I think I figured it out!

https://github.com/pipecat-ai/pipecat/pull/3480

PR #3277, opened 2025-12-21
fix: Separate LLMContextAssistantTimestampFrame from OpenAILLMContext…

markbackman:
…AssistantTimestampFrame

With the addition of the deprecation warning, using the LLMContext results in:
```
DeprecationWarning: OpenAILLMContextAssistantTimestampFrame is deprecated and will be removed in a future version. Use LLMContextAssistantTimestampFrame with the universal LLMContext and LLMContextAggregatorPair instead. See OpenAILLMContext docstring for migration guide.
```

The root cause is that the `LLMContextAssistantTimestampFrame` is aliased from the `OpenAILLMContextAssistantTimestampFrame`, so the deprecation from `OpenAILLMContextAssistantTimestampFrame` is passed on. Instead, these should be two separate frames without any inheritance. Doing so eliminates the deprecation warning and creates a clear path for removing the old path.

### DIFF
--- a/src/pipecat/frames/frames.py
+++ b/src/pipecat/frames/frames.py
@@ -514,9 +514,15 @@ class OpenAILLMContextAssistantTimestampFrame(DataFrame):
             )
 
 
-# A more universal (LLM-agnostic) name for
-# OpenAILLMContextAssistantTimestampFrame, matching LLMContext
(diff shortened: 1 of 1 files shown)


aconchillo:
LGTM

PR #3287, opened 2025-12-23
Fix TTFB metric and add multi-context WebSocket support for Async TTS

ashotbagh:
### Summary
This PR improves the Async TTS integration by fixing TTFB metric calculation and making WebSocket-based interruption handling more robust.

### Changes
- Fixed incorrect TTFB metric calculation in `AsyncAIHttpTTSService`.
- Updated `AsyncAITTSService` to support multi-context WebSocket sessions, improving reliability when handling interruptions and rapid context switches.

### Design Notes
In `tts_service.py`, the only abstract class that provides built-in audio context management is `AudioContextWordTTSService`, which assumes the availability of word-level timestamps.

Since Async does not currently provide word timestamps, the Async TTS implementation cannot derive from this class. Instead, it derives from `WebsocketTTSService`, and the required context-management logic is implemented directly in the Async service.

It may be beneficial to introduce a base class for audio context management that does not depend on word timestamps, which would allow services like Async to reuse this logic more cleanly.

### DIFF
--- a/src/pipecat/services/asyncai/tts.py
+++ b/src/pipecat/services/asyncai/tts.py
@@ -9,9 +9,9 @@
 import asyncio
 import base64
 import json
-from typing import AsyncGenerator, Optional, Dict
-
 import uuid
+from typing import AsyncGenerator, Optional
+
 import aiohttp
 from loguru import logger
 from pydantic import BaseModel
@@ -127,10 +127,6 @@ class AsyncAITTSService(AudioContextTTSService):
             **kwargs,
         )
 
-        self._contexts: Dict[str, asyncio.Queue] = {}
-        self._audio_context_task = None
-        self._context_id = None
-
         params = params or AsyncAITTSService.InputParams()
 
         self._api_key = api_key
@@ -153,6 +149,30 @@ class AsyncAITTSService(AudioContextTTSService):
         self._receive_task = None
         self._keepalive_task = None
         self._started = False
+        self._context_id = None
+
+    def can_generate_metrics(self) -> bool:
+        """Check if this service can generate processing metrics.
+
+        Returns:
+            True, as Async service supports metrics generation.
+        """
+        return True
+
+    def language_to_service_language(self, language: Language) -> Optional[str]:
+        """Convert a Language enum to Async language format.
+
+        Args:
+            language: The language to convert.
+
+        Returns:
+            The Async-specific language code, or None if not supported.
+        """
+        return language_to_async_language(language)
+
+    def _build_msg(self, text: str = "", context_id: str = "", force: bool = False) -> str:
+        msg = {"transcript": text, "context_id": context_id, "force": force}
+        return json.dumps(msg)
 
     async def start(self, frame: StartFrame):
         """Start the Async TTS service.
@@ -181,29 +201,6 @@ class AsyncAITTSService(AudioContextTTSService):
         """
         await super().cancel(frame)
         await self._disconnect()
-
-    def can_generate_metrics(self) -> bool:
-        """Check if this service can generate processing metrics.
-
-        Returns:
-            True, as Async service supports metrics generation.
-        """
-        return True
-
-    def language_to_service_language(self, language: Language) -> Optional[str]:
-        """Convert a Language enum to Async language format.
-
-        Args:
-            language: The language to convert.
-
-        Returns:
-            The Async-specific language code, or None if not supported.
-        """
-        return language_to_async_language(language)
-
-    def _build_msg(self, text: str = "", context_id: str = "", force: bool = False) -> str:
-        msg = {"transcript": text, "context_id": context_id, "force": force}
-        return json.dumps(msg)
 
     async def _connect(self):
         await super()._connect()
@@ -264,7 +261,7 @@ class AsyncAITTSService(AudioContextTTSService):
                 await self._websocket.close()
                 logger.debug("Disconnected from Async")
         except Exception as e:
-            logger.error(f"{self} error closing websocket: {e}")
+            await self.push_error(error_msg=f"Unknown error occurred: {e}", exception=e)
         finally:
             self._websocket = None
             self._context_id = None
@@ -338,7 +335,7 @@ class AsyncAITTSService(AudioContextTTSService):
                 if self._websocket and self._websocket.state is State.OPEN:
                     if self._context_id:
                         keepalive_message = {
-                            "transcript": " ", 
+                            "transcript": " ",
                             "context_id": self._context_id,
                         }
                         logger.trace("Sending keepalive message")
@@ -397,24 +394,22 @@ class AsyncAITTSService(AudioContextTTSService):
                     if not self.audio_context_available(self._context_id):
                         await self.create_audio_context(self._context_id)
 
-                    msg = self._build_msg(text=" ", context_id=self._context_id)
-                    await self._get_websocket().send(msg)
                     msg = self._build_msg(text=text, force=True, context_id=self._context_id)
                     await self._get_websocket().send(msg)
                     await self.start_tts_usage_metrics(text)
                 else:
                     if self._websocket and self._context_id:
                         msg = self._build_msg(text=text, force=True, context_id=self._context_id)
-                        await self._get_websocket().send(msg)                    
+                        await self._get_websocket().send(msg)
 
             except Exception as e:
-                logger.error(f"{self} error sending message: {e}")
+                yield ErrorFrame(error=f"Unknown error occurred: {e}")
                 yield TTSStoppedFrame()
                 self._started = False
                 return
             yield None
         except Exception as e:
-            logger.error(f"{self} exception: {e}")
+            yield ErrorFrame(error=f"Unknown error occurred: {e}")
 
 
 class AsyncAIHttpTTSService(TTSService):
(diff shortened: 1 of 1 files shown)


markbackman:
FYI: I'm proposing an AudioContextTTSService base class here: https://github.com/pipecat-ai/pipecat/pull/3289. If we decide to include that, it will greatly simplify changes to this class.

CHANGELOG.md
```diff
@@ -6,6 +6,15 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 <!-- towncrier release notes start -->
+## [Unreleased]
```
markbackman: We now use changelog fragements. You can learn how to provide changelog updates here:

ashotbagh:
I’ve pushed an update adapting the Async TTS implementation to use the newly merged AudioContextTTSService. Thanks a lot for introducing this base class — it’s a great addition and it significantly simplifies context handling for services like Async.
One small question regarding the inheritance: since AudioContextTTSService is intended to serve as a TTS base, would it make sense for it to inherit from WebsocketTTSService (or otherwise include TTSService) so that downstream implementations automatically get the standard TTS helpers, metrics, and behavior?
For Async, I can adapt either way, but at the moment I still rely on several utilities that currently live in WebsocketTTSService / TTSService. Just wanted to flag this in case it’s useful for future refinements.

markbackman:
> One small question regarding the inheritance: since AudioContextTTSService is intended to serve as a TTS base, would it make sense for it to inherit from WebsocketTTSService (or otherwise include TTSService) so that downstream implementations automatically get the standard TTS helpers, metrics, and behavior?

You're right. This was an oversight. I'm making the suggested change here:
https://github.com/pipecat-ai/pipecat/pull/3363

Hopefully this will be reviewed and on `main` at some point today.

markbackman:
Ok, merged! If you update this PR, I'll be ready to review and test.

ashotbagh:
Thanks a lot for the very fast fix and follow-up - much appreciated.

I’ve rebased on the latest main and updated the Async TTS implementation to align with the updated AudioContextTTSService.

Happy to adjust further if needed.

src/pipecat/services/asyncai/tts.py
```diff
@@ -249,11 +254,16 @@ async def _disconnect_websocket(self):
...
             if self._websocket:
                 logger.debug("Disconnecting from Async")
+                # Close all contexts and the socket
+                if self._context_id:
+                    await self._websocket.send(json.dumps({"terminate": True}))
                 await self._websocket.close()
+                logger.debug("Disconnected from Async")
         except Exception as e:
-            await self.push_error(error_msg=f"Unknown error occurred: {e}", exception=e)
+            logger.error(f"{self} error closing websocket: {e}")
```
markbackman: This needs to be `await self.push_error` in the format we recommend. If you want to change anything, you can change the error message. (Don't include `{self}` as that's already included; also there's a logger.error called as a result of push_error, so no need to call that either. This allows developers to handle service failures.)

src/pipecat/services/asyncai/tts.py
```diff
@@ -126,6 +127,10 @@ def __init__(
             **kwargs,
         )
 
+        self._contexts: Dict[str, asyncio.Queue] = {}
```
markbackman: Unused, remove:
```suggestion
```

src/pipecat/services/asyncai/tts.py
```diff
@@ -126,6 +127,10 @@ def __init__(
...
+        self._audio_context_task = None
```
markbackman: Unused, remove:
```suggestion
```

src/pipecat/services/asyncai/tts.py
```diff
@@ -200,6 +182,29 @@ async def cancel(self, frame: CancelFrame):
         await super().cancel(frame)
         await self._disconnect()
 
+    def can_generate_metrics(self) -> bool:
```
markbackman: Nit: any reason to move these three functions?

src/pipecat/services/asyncai/tts.py
```diff
@@ -126,6 +127,10 @@ def __init__(
             **kwargs,
         )
 
+        self._contexts: Dict[str, asyncio.Queue] = {}
+        self._audio_context_task = None
+        self._context_id = None
```
markbackman: Nit: collocate this with:
```
        self._receive_task = None
        self._keepalive_task = None
        self._started = False
        self._context_id = None
```

src/pipecat/services/asyncai/tts.py
```diff
@@ -9,8 +9,9 @@
 import asyncio
 import base64
 import json
-from typing import AsyncGenerator, Optional
+from typing import AsyncGenerator, Optional, Dict
```
markbackman: Remove `Dict`, which is now unused.
```suggestion
from typing import AsyncGenerator, Optional
```

src/pipecat/services/asyncai/tts.py
```diff
@@ -332,25 +382,35 @@ async def run_tts(self, text: str) -> AsyncGenerator[Frame, None]:
...
+                    await self.start_ttfb_metrics()
+                    yield TTSStartedFrame()
+                    self._started = True
+
+                    if not self._context_id:
+                        self._context_id = str(uuid.uuid4())
+                    if not self.audio_context_available(self._context_id):
+                        await self.create_audio_context(self._context_id)
+
+                    msg = self._build_msg(text=" ", context_id=self._context_id)
```
markbackman: Why is this required? Do you need to prepend a `" "`? Is it required to send a blank character for some reason?

src/pipecat/services/asyncai/tts.py
```diff
@@ -332,25 +382,35 @@ async def run_tts(self, text: str) -> AsyncGenerator[Frame, None]:
...
+                    await self._get_websocket().send(msg)
+                    await self.start_tts_usage_metrics(text)
+                else:
+                    if self._websocket and self._context_id:
+                        msg = self._build_msg(text=text, force=True, context_id=self._context_id)
+                        await self._get_websocket().send(msg)                    
+
             except Exception as e:
-                yield ErrorFrame(error=f"Unknown error occurred: {e}")
+                logger.error(f"{self} error sending message: {e}")
```
markbackman: You'll want to revert this too. You need to yield ErrorFrame here:
```
```

You can change the message.

src/pipecat/services/asyncai/tts.py
```diff
@@ -332,25 +382,35 @@ async def run_tts(self, text: str) -> AsyncGenerator[Frame, None]:
...
                 yield TTSStoppedFrame()
-                await self._disconnect()
-                await self._connect()
+                self._started = False
                 return
             yield None
         except Exception as e:
-            yield ErrorFrame(error=f"Unknown error occurred: {e}")
+            logger.error(f"{self} exception: {e}")
```
markbackman: Same here—yield ErrorFrame:
```
```

markbackman (commented):
For the TTFB change, there's a more minimal change that doesn't require tracking `first_byte_seen`. Since it's an HTTP request, each request has a single response.

So, in your `run_tts()`, you can just do:
```
    @traced_tts
    async def run_tts(self, text: str) -> AsyncGenerator[Frame, None]:
        """Generate speech from text using Async's HTTP streaming API.

        Args:
            text: The text to synthesize into speech.

        Yields:
            Frame: Audio frames containing the synthesized speech.
        """
        logger.debug(f"{self}: Generating TTS [{text}]")

        try:
            voice_config = {"mode": "id", "id": self._voice_id}
            payload = {
                "model_id": self._model_name,
                "transcript": text,
                "voice": voice_config,
            }
            yield TTSStartedFrame()
            headers = {
                "version": self._api_version,
                "x-api-key": self._api_key,
                "Content-Type": "application/json",
            }
            url = f"{self._base_url}/text_to_speech/streaming"

            async with self._session.post(url, json=payload, headers=headers) as response:
                if response.status != 200:
                    error_text = await response.text()
                    await self.push_error(error_msg=f"Async API error: {error_text}")
                    raise Exception(f"Async API returned status {response.status}: {error_text}")

                # Read streaming bytes; stop TTFB on the *first* received chunk
                buffer = bytearray()
                async for chunk in response.content.iter_chunked(64 * 1024):
                    if not chunk:
                        continue
                    buffer.extend(chunk)
                audio_data = bytes(buffer)

            frame = TTSAudioRawFrame(
                audio=audio_data,
                sample_rate=self.sample_rate,
                num_channels=1,
            )

            yield frame

        except Exception as e:
        finally:
            yield TTSStoppedFrame()
```

Note: the only change is:
```
-                audio_data = await response.read()
+                # Read streaming bytes; stop TTFB on the *first* received chunk
+                buffer = bytearray()
+                async for chunk in response.content.iter_chunked(64 * 1024):
+                    if not chunk:
+                    buffer.extend(chunk)
+                audio_data = bytes(buffer)
```

With this, I see sane TTFB values and correct character usage data.

markbackman (commented):
Also, we just merged this PR:
https://github.com/pipecat-ai/pipecat/pull/3392

To get the benefits of it, for `AsyncAITTSService`, you should:
- In `_connect()` call `await super()._connect()` first
- In `_disconnect()`, call `await super()._disconnect()` first

This will add better websocket handling in the event of exceptions. It will handle reconnects, etc.

ashotbagh:
Hi @markbackman,

All comments resolved, rebased on main, and linting is clean.
Ready for re-review. Thanks!

PR #3289, opened 2025-12-23
Add AudioContextTTSService base class, update AudioContextWordTTSServ…

markbackman:
…ice inheritance

We've had some vendors with services that support audio contexts but not word timestamps. This base class update makes it easier to build a TTS integration that matches this scenario.


aconchillo:
LGTM. We should plan to cleanup TTS services in the future. I believe everything should be based on audio contexts, even HTTP services or services that don't even have audio contexts. I believe it will be much simpler to think about.

PR #3292, opened 2025-12-24
initial user mute strategies

aconchillo:
## Introduction

This PR introduces a set of **user mute strategies** for Pipecat. These strategies control when user input is muted based on the current system state (e.g., bot speaking, function calls in progress), rather than relying on speech detection.

## Design Goals

* Replace the legacy [STTMuteFilter](https://docs.pipecat.ai/server/utilities/filters/stt-mute) with a more flexible approach.
* Provide clear, modular strategies that can be combined or swapped depending on the conversation scenario.

## Configuration

The new mute strategies are implemented as subclasses of `BaseUserMuteStrategy`:

1. **`FirstSpeechUserMuteStrategy`**
   Mutes user frames **only during the bot’s first speech**. Input is allowed before the bot starts speaking.

2. **`MuteUntilFirstBotCompleteUserMuteStrategy`**
   Mutes user frames **from the start of the interaction until the bot finishes its first speech**.

3. **`AlwaysUserMuteStrategy`**
   Mutes user frames **for the entire duration of each bot speech**.

4. **`FunctionCallUserMuteStrategy`**
   Mutes user frames while a **function call is in progress**, unmuting after completion or cancellation.

## Migration Notes

To migrate from the legacy `STTMuteFilter`, replace it with one of the new user mute strategies when initializing the `LLMContextAggregatorPair`. Example:

```python
    context,
        user_mute_strategies=[FirstSpeechUserMuteStrategy()]
    ),
)
```

### DIFF
--- a/examples/foundational/24-user-mute-strategy.py
+++ b/examples/foundational/24-user-mute-strategy.py
@@ -35,8 +35,8 @@ from pipecat.transports.base_transport import BaseTransport, TransportParams
 from pipecat.transports.daily.transport import DailyParams
 from pipecat.transports.websocket.fastapi import FastAPIWebsocketParams
 from pipecat.turns.bot.turn_analyzer_bot_turn_start_strategy import TurnAnalyzerBotTurnStartStrategy
-from pipecat.turns.mute.function_call_user_mute_strategy import FunctionCallUserMuteStrategy
-from pipecat.turns.mute.mute_until_first_bot_complete_user_mute_strategy import (
+from pipecat.turns.mute import (
+    FunctionCallUserMuteStrategy,
     MuteUntilFirstBotCompleteUserMuteStrategy,
 )
 from pipecat.turns.turn_start_strategies import TurnStartStrategies
--- a/src/pipecat/turns/mute/__init__.py
+++ b/src/pipecat/turns/mute/__init__.py
@@ -1,0 +1,13 @@
+#
+# Copyright (c) 2024–2025, Daily
+#
+# SPDX-License-Identifier: BSD 2-Clause License
+#
+
+from pipecat.turns.mute.always_user_mute_strategy import AlwaysUserMuteStrategy
+from pipecat.turns.mute.base_user_mute_strategy import BaseUserMuteStrategy
+from pipecat.turns.mute.first_speech_user_mute_strategy import FirstSpeechUserMuteStrategy
+from pipecat.turns.mute.function_call_user_mute_strategy import FunctionCallUserMuteStrategy
+from pipecat.turns.mute.mute_until_first_bot_complete_user_mute_strategy import (
+    MuteUntilFirstBotCompleteUserMuteStrategy,
+)


src/pipecat/turns/mute/function_call_user_mute_strategy.py
```diff
@@ -0,0 +1,59 @@
...
+
+        Args:
+            frame: The frame to be processed.
+
+        Returns:
+            Whether the strategy is muted.
+        """
+        await super().process_frame(frame)
+
+        if isinstance(frame, FunctionCallsStartedFrame):
```
aconchillo, line 50: We use `FunctionCallsStartedFrame` instead of `FunctionCallInProgressFrame` because `FunctionCallInProgressFrame` is not a system frame anymore, so a VAD event could come quicker.

src/pipecat/turns/mute/function_call_user_mute_strategy.py
```diff
@@ -0,0 +1,59 @@
...
+    function execution. While a function call is active, all user frames are
+    muted. Once the function call completes or is canceled, user input is
+    allowed again.
+
+    """
+
+    def __init__(self):
+        """Initialize the function call user mute strategy."""
+        super().__init__()
+        self._function_call_in_progress: Set[str] = set()
```
aconchillo, line 33: We keep track of all function calls now, not just one.

src/pipecat/turns/mute/function_call_user_mute_strategy.py
markbackman: Makes sense. This is a change we should probably make to the STTMuteFilter too, right?

markbackman:
Given this change, should we deprecate the STTMuteFilter?

examples/foundational/24-user-mute-strategy.py
```diff
@@ -0,0 +1,178 @@
+#
```
markbackman, line 1: We now have two 24 examples. Maybe one is 24a?

src/pipecat/turns/mute/function_call_user_mute_strategy.py
aconchillo: Done! Also added support for multiple simultaneous function calls.

examples/foundational/24-user-mute-strategy.py
aconchillo: I keep is as `24` because the STTMuteFilter will be removed at some point.

aconchillo:
> LGTM! I also tested the scenarios manually and things look good.
> 
> Thoughts:
> 
> * Should we deprecate the STTMuteFilter?

Done. Also enabled function calls UT and userd FunctionCallsStartedFrame and supported multiple simulateanous function calls.

> * Imports: We've required complete import paths for things in Pipecat to date. This can be a little annoying with things in turns because each individual strategy comes from a different path. Maybe this is fine, but we could also consider making imports easier by providing them via **init**.py

That might not be a bad idea.

aconchillo:
Added mute strategies to `turns.mute.__init__`. Certainly looks nicer. I will do the same for turn start strategies as well.

markbackman:
> Added mute strategies to `turns.mute.__init__`. Certainly looks nicer. I will do the same for turn start strategies as well.

Much better! Yes, I agree that the turn strategies should have the same. We should figure out where else this makes sense, but not a top priority.

src/pipecat/processors/filters/stt_mute_filter.py
```diff
@@ -120,6 +120,16 @@ def __init__(self, *, config: STTMuteConfig, **kwargs):
         self._function_call_in_progress = set()
         self._is_muted = False
 
+        import warnings
```
markbackman, line 135: Can you also add a deprecation directive to the STTMuteFilter class's docstring?

markbackman (approved):
Very nice! 👏

src/pipecat/processors/filters/stt_mute_filter.py
aconchillo: Done!

PR #3288, opened 2025-12-23
Inworld TTS service clean up

markbackman:
#### Please describe the changes in your PR. If it is addressing an issue, please reference that as well.

Improvements from https://github.com/pipecat-ai/pipecat/pull/3256, through commit ff3f8bb1d6d59a979a870557823577948cd71590.

Also, added 5038ebf2054ffb4ed10cba354b63904b86baae82, which uses the WebsocketService base class for reconnects instead of trying to manage them within this class.

Fixes https://github.com/pipecat-ai/pipecat/issues/3302

Issue #3269, opened 2025-12-19
Inworld TTS session pauses due to inactivity and leaves pipeline in a dangling state


hwuiwon:
### pipecat version

0.0.98

### Python version

3.13

### Operating System

linux x86_64

### Issue description

During an active pipeline run, the InworldTTSService receives a SESSION_PAUSED error (code: 10, "Paused due to inactivity"). Although the error is non-fatal, the TTS websocket closes without a proper close frame and the pipeline does not recover or terminate cleanly.

As a result:

The pipeline continues running until the full input timeout (~1200s)

Multiple Pipecat tasks remain pending/dangling (TTS, STT, gates, aggregators, transports)

Modal eventually force-cancels the input and reports task cleanup warnings and gRPC 504 errors

The call is marked as call_complete with pipeline_failed=False, despite the degraded state

### Reproduction steps

Using inworld websocket service example in pipecat examples should also produce the similar output

### Expected behavior

On SESSION_PAUSED from Inworld TTS, the pipeline should either reconnect or explicitly fail/short-circuit and cleanly cancel all downstream tasks.

### Actual behavior

The pipeline stalls, leaks tasks, and relies on global timeout for cleanup.

### Logs

```shell
2025-12-19 19:33:57.967 | ERROR    | function_call_id=fcc | conversation_id=abc | InworldTTSService#1 error: {'code': 10, 'message': 'Paused due to inactivity', 'details': [{'@type': 'type.googleapis.com/ai.inworld.common.InworldStatus', 'errorType': 'SESSION_PAUSED', 'reconnectType': 'IMMEDIATE', 'reconnectTime': None, 'maxRetries': 0}]}
2025-12-19 19:33:57.985 | WARNING  | function_call_id=fcc | conversation_id=abc | PipelineTask#1: Something went wrong: ErrorFrame#0(error: {'code': 10, 'message': 'Paused due to inactivity', 'details': [{'@type': 'type.googleapis.com/ai.inworld.common.InworldStatus', 'errorType': 'SESSION_PAUSED', 'reconnectType': 'IMMEDIATE', 'reconnectTime': None, 'maxRetries': 0}]}, fatal: False)
2025-12-19 19:33:59.968 | WARNING  | function_call_id=fcc | conversation_id=abc | InworldTTSService#1 connection closed, but with an error: no close frame received or sent
Task's current input(s) ["in-abc"] cancelled because: Task's current input in-abc hit its timeout of 1200s
[...]

PR #3295, opened 2025-12-24
Add Keepalive to Inworld TTS

cshape:
Continuing from the Inworld TTS service improvements in #3288, adding a keepalive function to prevent the timeout issues noted in #3269.

This follows similar patterns from other providers like Elevenlabs.

Note that Inworld will return an error when there's no context (eg. when the keepalive runs during a user interruption), but this doesn't affect functionality and works to keep the connection alive.

### DIFF
--- a/changelog/3295.added.md
+++ b/changelog/3295.added.md
@@ -1,0 +1,1 @@
+- Added Keepalive Function to Inworld TTS to prevent timeouts after 10 minutes of inactivity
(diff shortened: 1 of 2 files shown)


markbackman:
Closing since this has been superseded by two PRs:
- https://github.com/pipecat-ai/pipecat/pull/3288
- https://github.com/pipecat-ai/pipecat/pull/3403

PR #3297, opened 2025-12-24
deprecate allow interruptions

aconchillo:

This PR adds the following deprecations:

- PipelineParams.allow_interruptions
- FrameProcessor.interruptions_allowed

Users should use the new user mute strategies instead.

### DIFF
--- a/src/pipecat/transports/base_output.py
+++ b/src/pipecat/transports/base_output.py
@@ -505,10 +505,12 @@ class BaseOutputTransport(FrameProcessor):
             await self._cancel_audio_task()
             await self._cancel_clock_task()
             await self._cancel_video_task()
+
             # Create tasks.
             self._create_video_task()
             self._create_clock_task()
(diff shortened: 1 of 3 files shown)


src/pipecat/transports/base_output.py
```diff
@@ -498,17 +498,16 @@ async def handle_interruptions(self, _: InterruptionFrame):
             Args:
                 _: The start interruption frame (unused).
             """
-            if not self._transport.interruptions_allowed:
```
markbackman, line 501: Removing this check is breaking. I think we need to leave it in until `interruptions_allowed()` and `allow_interruptions` are removed.

By removing this check, `allow_interruptions` has no impact and interruptions occur regardless of how `allow_interruptions` is set.

aconchillo: Oh, yes! Duh. I thought this was another function. Updated!

PR #3298, opened 2025-12-24
push UserStartedSpeakingFrame before interruption

aconchillo:

Originally, the base input transport has always pushed `UserStartedSpeakingFrame` before `InterruptionFrame`. We need to follow this pattern for real-time services as well. The idea is: the user starts speaking and this causes an interruption.

### DIFF
--- a/examples/foundational/07c-interruptible-deepgram-vad.py
+++ b/examples/foundational/07c-interruptible-deepgram-vad.py
@@ -98,7 +98,7 @@ async def run_bot(transport: BaseTransport, runner_args: RunnerArguments):
 
     @stt.event_handler("on_speech_started")
     async def on_speech_started(stt, *args, **kwargs):
-        await task.queue_frames([InterruptionFrame(), UserStartedSpeakingFrame()])
+        await task.queue_frames([UserStartedSpeakingFrame(), InterruptionFrame()])
 
     @stt.event_handler("on_utterance_end")
     async def on_utterance_end(stt, *args, **kwargs):
--- a/src/pipecat/services/deepgram/flux/stt.py
+++ b/src/pipecat/services/deepgram/flux/stt.py
@@ -591,8 +591,8 @@ class DeepgramFluxSTTService(WebsocketSTTService):
         """
         logger.debug("User started speaking")
         self._user_is_speaking = True
-        await self.push_interruption_task_frame_and_wait()
         await self.broadcast_frame(UserStartedSpeakingFrame)
+        await self.push_interruption_task_frame_and_wait()
         await self.start_metrics()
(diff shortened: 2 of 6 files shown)


markbackman (approved):
This seem logical. Is there any impact to making this change other than the frame ordering being different?

aconchillo:
> This seem logical. Is there any impact to making this change other than the frame ordering being different?

It shouldn't unless someone does some reset of a flag (e.g. user is speaking) when an interruption occurs, but I've never seen that.

kompfner:
> We need to follow this pattern for real-time services as well

OK, here's a wrinkle: two of the real-time APIs, Gemini Live and AWS Nova Sonic, do _not_, AFAICT, tell us when they think the user has started or stopped speaking.

But they _do_ tell us when a user barge-in has occurred, which folks in the community are finding useful for timely interruptions (see #3429 and #2431).

If we're cool with those services _not_ acting as turn controllers (i.e. not sending user started/stopped), but we _do_ want to benefit from their timely interruption detection, then...are we OK with the ordering (`UserStartedSpeaking` always preceding the interruption frame) not being strict? Because in Gemini Live and AWS Nova Sonic, I think I'd prefer that we have code like:

```
if barge_in_detected:
  # no UserStartedSpeakingFrame here since it'd be inconsistent,
  # as this service is *not* otherwise a reliable turn controller
```

markbackman:
> > We need to follow this pattern for real-time services as well
> 
> OK, here's a wrinkle: two of the real-time APIs, Gemini Live and AWS Nova Sonic, do _not_, AFAICT, tell us when they think the user has started or stopped speaking.
> 
> But they _do_ tell us when a user barge-in has occurred, which folks in the community are finding useful for timely interruptions (see #3429 and #2431).
> 
> If we're cool with those services _not_ acting as turn controllers (i.e. not sending user started/stopped), but we _do_ want to benefit from their timely interruption detection, then...are we OK with the ordering (`UserStartedSpeaking` always preceding the interruption frame) not being strict? Because in Gemini Live and AWS Nova Sonic, I think I'd prefer that we have code like:
> 
> ```
> if barge_in_detected:
>   # no UserStartedSpeakingFrame here since it'd be inconsistent,
>   # as this service is *not* otherwise a reliable turn controller
>   await self.push_interruption_task_frame_and_wait()
> ```

That's a tricky one. I guess if you need conversational events, we could say that you need the aggregators. This is probably more likely if you're building a web or mobile app.

On the other hand, if you're just building for the phone, events are likely less useful, so you might just want interruptions.

Given that, what you're proposing might be OK, but it comes with the caveat of requiring aggregators if you want the speaking events, right?

Issue #1564, opened 2025-04-11
Bug: Long-Running Function Calls Break Tool Response Processing in Gemini Live Realtime [Websockets]


rizified:
Environment
pipecat-ai version: 0.0.62
python version: 3.11.11
OS: Windows

Issue description
When using Google Gemini Live Realtime with Websocket Implementation, function calls that take more than 6-7 seconds break the normal flow of processing. After a long-running function completes, the LLM is unable to process the tool result because the user transcription finishes first and becomes the last message in the context.

This happens because in GeminiMultimodalLiveLLMService, the condition `elif context.messages and context.messages[-1].get("role") == "tool"` fails since the last role becomes "user" instead of "tool" which doesn't executes the LLM. 

![Image](https://github.com/user-attachments/assets/73d1af35-70e2-4da1-88a3-038f9c5e0981)

Repro steps
1. Run the realtime Gemini function call demo with websockets from the samples
2. Add a sleep(8) in the function method
3. Observe that the function result doesn't get processed by the LLM

Expected behavior
Long-running function calls should still have their results properly processed by the LLM, regardless of timing. The tool message should be the last one processed.

Actual behavior
The user transcription message is placed after the tool result message, which prevents the LLM from processing the tool result.

Logs
see screenshot.

Proposed Fix
I've found a workaround by modifying `pipecat.services.gemini_multimodal_live.gemini` to detect this condition and rearrange the messages:

Updated the code to - 

```
 elif isinstance(frame, OpenAILLMContextFrame):
             context: GeminiMultimodalLiveContext = GeminiMultimodalLiveContext.upgrade(
                 frame.context
             )
             # For now, we'll only trigger inference here when either:
             #   1. We have not seen a context frame before
             #   2. The last message is a tool call result
             if (self._context and len(context.messages) > 3 and
                 context.messages[-1].get('role') == 'user' and
                 context.messages[-2].get('role') == 'tool' and
                 context.messages[-3].get('tool_calls', [{}])[0].get('type') == 'function'
             ):
                 logger.info("Rearranging messages to move user role dictionary before LLM call for function call result")
                 # Extract the last element (user role dictionary)
                 user_element = context.messages.pop(-1)
                 # Insert it before the third-to-last element (type:function)
                 context.messages.insert(len(context.messages) - 2, user_element)
             else:
                 pass
             if not self._context:
                 self._context = context
                 if frame.context.tools:
                     self._tools = frame.context.tools
                 await self._create_initial_response()
[...]

PR #569, opened 2024-10-11
Fixed parallel async function calls bug

chadbailey59:
This PR fixes a bug that caused garbled and/or repeated output when using parallel function calling with RTVI. Instead of requesting LLM inference when the final item of the function call list completes, we're now actually waiting for all of the function calls to complete (so they can be processed in parallel by the client).

PR #3301, opened 2025-12-26
Fixes #3273: Await parallel function call tasks to prevent race condition

### LINKED ISSUE #3273: Race condition in _run_parallel_function_calls: Tasks created but not awaited
### pipecat version

0.0.85+ (affects all versions with parallel tool execution)

### Python version

3.10+

### Operating System

All platforms

### Issue description

There is a critical race condition in `LLMService._run_parallel_function_calls()` that causes parallel tool execution to fail. The method creates asynchronous tasks for function calls but **never awaits them**, causing the LLM to continue processing and go idle before tool results are available.

This results in tool calls being executed but their results never being processed by the LLM, breaking the agentic loop.

### Root cause

In `src/pipecat/services/llm_service.py`, the `_run_parallel_function_calls` method:

```python
async def _run_parallel_function_calls(self, runner_items: Sequence[FunctionCallRunnerItem]):
    tasks = []
    for runner_item in runner_items:
        task = self.create_task(self._run_function_call(runner_item))
        tasks.append(task)
        self._function_call_tasks[task] = runner_item
        task.add_done_callback(self._function_call_task_finished)
    # BUG: Method returns here without awaiting any tasks!
```

The method:
1. ✅ Creates tasks for each function call
2. ✅ Adds them to `_function_call_tasks` for tracking
3. ✅ Registers completion callbacks
4. ❌ **Returns immediately without awaiting task completion**

This means:
- Tasks are created and start executing in the background
- The LLM immediately continues processing
- The LLM may go idle before tool results arrive
- Tool results arrive but the LLM is no longer waiting for them
- The agentic loop breaks

### Reproduction steps

1. Create an LLM service with `run_in_parallel=True` (the default)
2. Configure a tool/function that the LLM can call
3. Trigger a scenario where the LLM calls the tool
4. Observe that the tool executes but the LLM doesn't process the result

**Example code:**
```python
llm = GroqLLMService(
    api_key=os.getenv("GROQ_API_KEY"),
    model="moonshotai/kimi-k2-instruct",
    run_in_parallel=True,  # Default behavior - triggers the bug
)

# Register a tool
async def my_tool():
    await asyncio.sleep(0.1)  # Simulate async work
    return {"result": "data"}

llm.register_function("my_tool", my_tool)
```

### Expected behavior

When `run_in_parallel=True`:
1. LLM calls one or more tools in parallel
2. `_run_parallel_function_calls` creates tasks for all tool calls
3. **The method awaits all tasks** (e.g., using `asyncio.gather()`)
4. Tool results are collected
5. Results are sent back to the LLM for processing
6. LLM continues the conversation with tool results

### Actual behavior

1. LLM calls one or more tools in parallel
2. `_run_parallel_function_calls` creates tasks for all tool calls
3. **Method returns immediately without waiting**
4. LLM continues processing and goes idle
5. Tool tasks complete in the background
6. Results arrive but LLM is no longer listening
7. Agentic loop breaks - tool results are lost

### Suggested fix

The method should await all tasks before returning:

```python
async def _run_parallel_function_calls(self, runner_items: Sequence[FunctionCallRunnerItem]):
    tasks = []
    for runner_item in runner_items:
        task = self.create_task(self._run_function_call(runner_item))
        tasks.append(task)
        self._function_call_tasks[task] = runner_item
        task.add_done_callback(self._function_call_task_finished)
    
    # FIX: Await all tasks before returning
    await asyncio.gather(*tasks, return_exceptions=True)
```

### Workaround

Until this is fixed, users must set `run_in_parallel=False`:

```python
llm = GroqLLMService(
    api_key=os.getenv("GROQ_API_KEY"),
    model="moonshotai/kimi-k2-instruct",
    run_in_parallel=False,  # Workaround: disable parallel execution
)
```

This forces sequential execution via `_sequential_runner_handler()`, which properly awaits each task.

### Impact

This bug affects all users who:
- Use `run_in_parallel=True` (the default)
- Have function/tool calling enabled
- Rely on the LLM to process tool results

The bug is particularly severe because:
- It's enabled by default
- It silently fails (tools execute but results are ignored)
- It breaks the core agentic loop pattern

### Additional context

This issue is distinct from:
- #424 (function name concatenation bug)
- #2043 (cancellation in sequential mode)
- #2787 (context management for mixed content/tool_calls responses)

The root cause is a missing `await` statement in the parallel execution path.

### Logs

```
# Tool is called but LLM doesn't process the result
# No error is thrown - the task just completes in the background
# LLM continues as if no tool was called
```

devbyteai:
## Summary

Fixes #3273

This PR fixes a critical race condition in `_run_parallel_function_calls()` where async tasks are created but never awaited, causing the LLM to continue processing before tool execution results arrive.

---

## The Problem

When `run_in_parallel=True` (the default), the `_run_parallel_function_calls()` method creates async tasks for each function call but **returns immediately without awaiting them**:

```python
async def _run_parallel_function_calls(self, runner_items):
    tasks = []
    for runner_item in runner_items:
        task = self.create_task(self._run_function_call(runner_item))
        tasks.append(task)
        self._function_call_tasks[task] = runner_item
        task.add_done_callback(self._function_call_task_finished)
    # BUG: Method returns here without awaiting tasks!
```

**Impact:**
- The LLM continues processing before tool results are ready
- Function call results are silently discarded
- The agentic loop pattern is completely broken
- Users report "tools execute but results are ignored"

**Current workaround:** Set `run_in_parallel=False` to use sequential execution (which properly awaits tasks).

---

## The Solution

Add `await asyncio.gather(*tasks, return_exceptions=True)` at the end of the method to properly synchronize all parallel function executions:

```python
async def _run_parallel_function_calls(self, runner_items):
    tasks = []
    for runner_item in runner_items:
        task = self.create_task(self._run_function_call(runner_item))
        tasks.append(task)
        self._function_call_tasks[task] = runner_item
        task.add_done_callback(self._function_call_task_finished)

    # FIX: Wait for all parallel function calls to complete before returning
    if tasks:
        await asyncio.gather(*tasks, return_exceptions=True)
```

---

## Why `return_exceptions=True`?

Using `return_exceptions=True` ensures that:
1. One failing function call doesn't cancel the others
2. All function calls complete before the method returns
3. Exceptions are captured but don't propagate (they're already handled by the callback)

This matches the behavior expected for parallel execution - all tasks should complete independently.

---

## Edge Cases Handled

1. **Empty runner_items** - The `if tasks:` guard prevents calling `gather()` on an empty list
2. **One task fails** - `return_exceptions=True` ensures other tasks still complete
3. **Callback cleanup** - `add_done_callback` still fires after `gather()` completes, maintaining proper cleanup

---

## Comparison with Sequential Execution

The sequential runner (`_sequential_runner_handler`) properly awaits each task:

```python
async def _sequential_runner_handler(self):
    while True:
        runner_item = await self._sequential_runner_queue.get()
        task = self.create_task(self._run_function_call(runner_item))
        # ...
        await task  # Properly awaited!
```

This fix brings parallel execution in line with the same synchronization guarantees.

---

## Testing

- Ran core test suite: 21 tests passed
- Verified the fix doesn't break existing callback cleanup logic
- The change is minimal and focused on the specific race condition

---

## Related Issues

- Issue #3273: Race condition in `_run_parallel_function_calls`
- PR #569: Previous fix for parallel async function calls (similar pattern)
- Issue #1564: Long-running function calls break tool response processing

### DIFF
--- a/src/pipecat/services/llm_service.py
+++ b/src/pipecat/services/llm_service.py
@@ -555,6 +555,11 @@ class LLMService(AIService):
             self._function_call_tasks[task] = runner_item
             task.add_done_callback(self._function_call_task_finished)
 
+        # Wait for all parallel function calls to complete before returning.
+        # Using return_exceptions=True to prevent one failing task from canceling others.
+        if tasks:
+            await asyncio.gather(*tasks, return_exceptions=True)
+
     async def _run_sequential_function_calls(self, runner_items: Sequence[FunctionCallRunnerItem]):
         # Enqueue all function calls for background execution.
         for runner_item in runner_items:


Maurotb:
Hi — wanted to flag something I ran into while testing this PR on our end before it lands.

The `await asyncio.gather(*tasks, return_exceptions=True)` at the end of `_run_parallel_function_calls` ends up deadlocking the `activate_worker(result_callback=...)` handoff pattern (the one shown in `LLMWorker.activate_worker`'s docstring for use from a `@tool`).

The chain is:

- `_run_parallel_function_calls` runs inside the LLM service's frame-processing task (via `process_frame` → `_process_context`). Awaiting the tasks there parks that task until the tools finish.
- A tool calling `activate_worker(..., result_callback=params.result_callback)` reaches `_finish_function_call` → `_flush_pipeline()` and waits for `_flush_done`.
- `_flush_pipeline` queues a `PipelineFlushFrame` and waits for it to round-trip back. That round-trip has to go through the LLM service's frame-processing task — which is the one currently parked in `gather`, waiting for the tool. Circular wait → permanent hang until the call is torn down.

Today this is hidden because `_run_parallel_function_calls` returns immediately (detached), so the flush frame flows. This PR removes that property.

### Minimal repro (deterministic)

Two bridged `LLMWorker`s, a `@tool` that calls `activate_worker(result_callback=...)`. `PATCH=1` applies this PR's `await gather`, `PATCH=0` is current `main`:

```
PATCH=1  → tool: activate_worker(result_callback=...) → *** DEADLOCK ***
PATCH=0  → tool: activate_worker(result_callback=...) → DONE (1ms)
```

<details>
<summary>repro script</summary>

```python
import asyncio, os, time
from pipecat.bus.local.async_queue import AsyncQueueBus
from pipecat.frames.frames import ControlFrame, FunctionCallFromLLM
from pipecat.services.llm_service import LLMService
from pipecat.workers.llm.llm_worker import LLMWorker
from pipecat.workers.llm.tool_decorator import tool
from pipecat.workers.runner import WorkerRunner

T0 = time.monotonic()
def log(*a): print(f"[{time.monotonic()-T0:6.3f}]", *a, flush=True)

if os.environ.get("PATCH", "1") == "1":
    async def _await_gather(self, runner_items):
        tasks = []
        for ri in runner_items:
            t = self.create_task(self._run_function_call(ri))
            tasks.append(t); self._function_call_tasks[t] = ri
            t.add_done_callback(self._function_call_task_finished)
        if tasks:
            await asyncio.gather(*tasks, return_exceptions=True)  # <-- this PR
    LLMService._run_parallel_function_calls = _await_gather


@dataclass
class TriggerFrame(ControlFrame):
    pass


class StubLLM(LLMService):
        if isinstance(frame, TriggerFrame):
            await self.run_function_calls([FunctionCallFromLLM(
                function_name="go", tool_call_id="1", arguments={}, context=None)])
        else:


class A(LLMWorker):
    def __init__(self):
        super().__init__("A", llm=StubLLM(), active=True, bridged=())

    @tool
    async def go(self, params):
        log("tool: activate_worker(result_callback=...)")
        try:
            await asyncio.wait_for(self.activate_worker(
                "B", deactivate_self=True,
                result_callback=params.result_callback), timeout=8)
            log("tool: DONE")
        except asyncio.TimeoutError:
            log("tool: *** DEADLOCK ***")
        return {"ok": True}


class B(LLMWorker):
    def __init__(self):
        super().__init__("B", llm=StubLLM(), active=False, bridged=())


async def main():
    bus = AsyncQueueBus()
    r = WorkerRunner(bus=bus, handle_sigint=False)
    a, b = A(), B()
    await r.add_workers(a, b)
    ready = asyncio.Event()
    r.event_handler("on_ready")(lambda _r: ready.set())
    rt = asyncio.create_task(r.run(auto_end=False))
    await asyncio.wait_for(ready.wait(), 10); await asyncio.sleep(1)
    log(f"PATCH={os.environ.get('PATCH','1')} -> inject")
    await a.queue_frame(TriggerFrame())
    await asyncio.sleep(10)
    await r.cancel()
    try: await asyncio.wait_for(rt, 5)
    except (asyncio.TimeoutError, asyncio.CancelledError): pass

asyncio.run(main())
```
</details>

### One possible alternative

The goal of #3273 (don't silently swallow exceptions escaping `_run_function_call`) seems achievable without blocking the caller — surface them from the done-callback instead:

```python
def _function_call_task_finished(self, task: asyncio.Task):
    self._function_call_tasks.pop(task, None)
    if task.cancelled():
        return
    exc = task.exception()
    if exc is not None:
        logger.warning(f"Function call task failed: {exc!r}")
```

Same exceptions get surfaced (both `gather` and `task.exception()` read the task's terminal state), execution stays detached, `_flush_pipeline` keeps working. I'm running this shape internally and it's been stable — happy to open a PR with it + a regression test if you'd like.

(Small caveat: `_run_function_call` already wraps the handler in `except Exception: push_error(...)`, so in practice this catches only `BaseException`s escaping outside that try — but the non-blocking property is the part that matters here.)

Hope it helps — and apologies if you've already considered this angle.

filipi87:
Hi @devbyteai, thanks for this PR! The issue it references (#3273) is already closed, so this looks like it may be stale or superseded by other work.

PR #3304, opened 2025-12-26
fix: kept parallel function calls together in single message

### LINKED ISSUE #3290: GoogleLLMService throws 400 error when using gemini 3 flash and executing multiple tool calls in one LLM turn
### pipecat version

0.0.98

### Python version

3.12.11

### Operating System

macOS 15

### Issue description

When the LLM executes multiple tool calls in a single turn, the following error is thrown:
```
GoogleLLMService#0 exception (~/src/pipecat/.venv/lib/python3.12/site-packages/google/genai/errors.py:210): Unknown error occurred: 400 Bad Request. {'message': '{\n  "error": {\n    "code": 400,\n    "message": "Function call is missing a thought_signature in functionCall parts. This is required for tools to work correctly, and missing thought_signature may lead to degraded model performance. Additional data, function call `default_api:get_restaurant_recommendation` , position 7. Please refer to https://ai.google.dev/gemini-api/docs/thought-signatures for more details.",\n    "status": "INVALID_ARGUMENT"\n  }\n}\n', 'status': 'Bad Request'}
```

### Reproduction steps

1. Run https://github.com/pipecat-ai/pipecat/blob/main/examples/foundational/14e-function-calling-google.py
2. Say, "Hi. I'm in Atlanta. Can you give me the weather and a restaurant recommendation?"
3. Bot says "Let me check on that", then thee 400 error.

### Expected behavior

Bot functions are called.

### Actual behavior

400 error

### Logs

```shell

```

omChauhanDev:
Fixes #3290 

## Solution
Modified `LLMAssistantAggregator` to keep parallel function calls together in a single message:

1. Track all function calls from `FunctionCallsStartedFrame` in a batch
2. Create one message with all `tool_calls` when the first `FunctionCallInProgressFrame` arrives
3. This matches how Gemini returns parallel calls and satisfies its validation requirements

Also updated `GeminiLLMAdapter._apply_thought_signatures_to_messages()` to check all parts in a message for bookmark matching (not just the last part), since parallel calls can have bookmarks on any part.

This pr also improves correctness for all LLM providers, as they all expect parallel tool calls in one message per the standard format.

Below are screenshots showing the issue before and after the fix.
## Before 
image

## After 
image

### DIFF
--- a/src/pipecat/adapters/services/gemini_adapter.py
+++ b/src/pipecat/adapters/services/gemini_adapter.py
@@ -477,7 +477,9 @@ class GeminiLLMAdapter(BaseLLMAdapter[GeminiLLMInvocationParams]):
         # Apply thought signatures to the corresponding assistant messages.
         # Thought signatures are already in message order.
         thought_signatures_applied = 0
-        message_start_index = 0  # Track where to start searching for the next matching message.
+        message_start_index = (
+            0  # Track where to start searching for the next matching message part.
+        )
         for thought_signature_dict in thought_signature_dicts:
             signature = thought_signature_dict.get("signature")
             bookmark = thought_signature_dict.get("bookmark")
@@ -494,12 +496,15 @@ class GeminiLLMAdapter(BaseLLMAdapter[GeminiLLMInvocationParams]):
                 thought_signature_found = False
                 for part in message.parts:
                     if self._thought_signature_bookmark_matches_part(bookmark, part):
-                        # Apply the thought signature to the matching part.
+                        # Apply the thought signature to the matching part
                         part.thought_signature = signature
                         thought_signatures_applied += 1
 
                         # Update the start index and stop searching for a match
-                        message_start_index = i + 1
+                        # (Note that we don't set message_start_index = i + 1
+                        # here, to support multiple thought signatures in the
+                        # same message; this is a theoretical concern for now)
+                        message_start_index = i
                         thought_signature_found = True
                         break
 


src/pipecat/adapters/services/gemini_adapter.py
```diff
@@ -490,18 +490,19 @@ def _apply_thought_signatures_to_messages(
...
+                        # Apply the thought signature to the matching part.
+                        part.thought_signature = signature
+                        thought_signatures_applied += 1
+
+                        # Update the start index and stop searching for a match
+                        message_start_index = i + 1
+                        break
+                else:
+                    continue
+                break
```
aconchillo: This works, but just to avoid confusion with people not familiar with for/else and to make it more clear we might want to use a variable:

```
thought_signature_found = False
for part in message.parts:
    if self._thought_signature_bookmark_matches_part(bookmark, part):
       ...
        thought_signature_found  = True
        break

if thought_signature_found:
    break
```

src/pipecat/processors/aggregators/llm_response_universal.py
```diff
@@ -725,22 +729,22 @@ async def _handle_function_call_in_progress(self, frame: FunctionCallInProgressF
...
+                            "arguments": json.dumps(fc.arguments),
                         },
                         "type": "function",
                     }
-                ],
-            }
-        )
+                )
+            self._context.add_message({"role": "assistant", "tool_calls": tool_calls})
+            self._function_calls_message_added = True
```
aconchillo: It feels we could add the `tool_calls` message inside `_handle_function_calls_started`. `FunctionCallsStartedFrame` is a system frame while `FunctionCallInProgressFrame` is a data frame, not sure if that would have an impact. 

If we could do that (which would be great) we would not need `self._pending_function_calls` and `self._function_calls_message_added`.

I'm not sure I'll have time to look at this in details (I'll be out for a few days). If I don't have time I'll leave it to @kompfner.

aconchillo: Side note: we need to keep `FunctionCallsStartedFrame` as a system frame because this is how the system gets notified quickly function calls will be called.

src/pipecat/adapters/services/gemini_adapter.py
omChauhanDev: agreed, the flag approach is clearer...updated!

src/pipecat/processors/aggregators/llm_response_universal.py
omChauhanDev: Thanks for the suggestion! 
moving tool_calls message creation to `_handle_function_calls_started` makes more sense, updated!

@kompfner please have a look at the final approach whenever you’re back - no rush. Thanks!

kompfner:
Pushed up one tiny change: https://github.com/pipecat-ai/pipecat/pull/3304/changes/9a41b8dc871763ad8a8eb71986b2752bee44bd7d

kompfner:
In testing this change out...it seems to break parallel function calling in Anthropic :/

Screenshot 2026-01-12 at 3 22 48 PM

kompfner:
And OpenAI, looks like:

Screenshot 2026-01-12 at 3 25 11 PM

kompfner:
And AWS Bedrock:

Screenshot 2026-01-12 at 3 26 17 PM

kompfner:
To avoid rocking the boat too much, I'm wondering if we can do a more targeted fix for just Gemini...

kompfner:
> To avoid rocking the boat too much, I'm wondering if we can do a more targeted fix for just Gemini...

Here's an option for a more surgical fix: https://github.com/pipecat-ai/pipecat/pull/3420

kompfner:
> Here's an option for a more surgical fix: https://github.com/pipecat-ai/pipecat/pull/3420

Ended up merging the above

PR #3308, opened 2025-12-28
turns: add external user and bot turn start strategies

aconchillo:

External strategies are strategies where the logic for user turn start and turn end come from different processors (e.g. an STT). This works great for Deepgram's Flux or other STT services that have turn features.

```
        context,
        user_params=LLMUserAggregatorParams(turn_start_strategies=ExternalTurnStartStrategies()),
    )
```

The external turn start strategies are also used to support backwards compatibility with turn analyzers being set in the transport.

### DIFF
--- a/src/pipecat/transports/base_input.py
+++ b/src/pipecat/transports/base_input.py
@@ -477,7 +477,7 @@ class BaseInputTransport(FrameProcessor):
             )
 
             # Make sure we notify about interruptions quickly out-of-band.
-            if should_push_immediate_interruption and self.interruptions_allowed:
+            if should_push_immediate_interruption and self._allow_interruptions:
                 await self.push_interruption_task_frame_and_wait()
             elif self.interruption_strategies and self._bot_speaking:
                 logger.debug(
--- a/src/pipecat/transports/base_output.py
+++ b/src/pipecat/transports/base_output.py
@@ -498,7 +498,7 @@ class BaseOutputTransport(FrameProcessor):
             Args:
                 _: The start interruption frame (unused).
             """
-            if not self._transport.interruptions_allowed:
+            if not self._transport._allow_interruptions:
                 return
 
             # Cancel tasks.


src/pipecat/processors/aggregators/llm_response.py
```diff
@@ -81,16 +81,11 @@ class LLMUserAggregatorParams:
...
-        enable_user_speaking_frames: [DO NOT USE] added for temporary backwards
-            compatibility.
-
     """
 
     aggregation_timeout: float = 0.5
     turn_emulated_vad_timeout: float = 0.8
     enable_emulated_vad_interruptions: bool = False
-    # Added for backwards compatibility.
-    enable_user_speaking_frames: bool = True
```
markbackman: Should we warn when used?

src/pipecat/turns/bot/external_bot_turn_start_strategy.py
```diff
@@ -0,0 +1,127 @@
...
+class ExternalBotTurnStartStrategy(BaseBotTurnStartStrategy):
+    """Bot turn start strategy controlled by an external processor.
+
+    This strategy does not determine when a user turn ends on its own, it relies
+    on a different processor in the pipeline which is responsible for emitting
+    `UserStoppedSpeakingFrame` frames.
+
+    """
+
+    def __init__(self, *, timeout: float = 0.2):
```
markbackman: Why 0.2 sec?

markbackman:
Conceptually, I think this makes sense. It explicitly tells the user context aggregator to trust the signals coming from other processors in the Pipeline.

In testing the backwards compatibility path, I'm seeing this error just before the StartFrame reaches the end of the Pipeline (sharing the logline before and after for context):
```
2025-12-28 21:17:53.046 | DEBUG    | pipecat.services.cartesia.tts:_connect_websocket:494 - Connecting to Cartesia TTS
/Users/mbackman/.local/share/uv/python/cpython-3.12.11-macos-aarch64-none/lib/python3.12/asyncio/events.py:88: RuntimeWarning: coroutine 'TranscriptionBotTurnStartStrategy._task_handler' was never awaited
  self._context.run(self._callback, *self._args)
RuntimeWarning: Enable tracemalloc to get the object allocation traceback
2025-12-28 21:17:53.490 | DEBUG    | pipecat.pipeline.task:_wait_for_pipeline_start:610 - PipelineTask#0: StartFrame#0 reached the end of the pipeline, pipeline is now ready.
```

I can hit this by running the 07-interruptible example with:
1. transport:
```
transport_params = {
    "daily": lambda: DailyParams(
        audio_in_enabled=True,
        audio_out_enabled=True,
        vad_analyzer=SileroVADAnalyzer(params=VADParams(stop_secs=0.2)),
        turn_analyzer=LocalSmartTurnAnalyzerV3(),
    ),
}
```

2. user context agg:
```
    context = LLMContext(messages)
```

I only see this runtime warning once (whereas I was previously seeing it after every speaking turn).

src/pipecat/turns/bot/external_bot_turn_start_strategy.py
aconchillo: I was just testing, put it back to 0.5. We can discuss tomorrow.

aconchillo:
> Conceptually, I think this makes sense. It explicitly tells the user context aggregator to trust the signals coming from other processors in the Pipeline.
> 
> In testing the backwards compatibility path, I'm seeing this error just before the StartFrame reaches the end of the Pipeline (sharing the logline before and after for context):
>

I think this is fine and should be harmless. The reason is that we cancel the task from the default strategies right away while Python hasn't even had time to start them. That's why it only happens once now: just at the beggining we start the task and we cancel right away once we detect there's a turn analyzer in the transport. Before, we were causing 2 interruptions back to back that was causing this issue all the time (that was a FrameProcessor task).

aconchillo:
> > Conceptually, I think this makes sense. It explicitly tells the user context aggregator to trust the signals coming from other processors in the Pipeline.
> > In testing the backwards compatibility path, I'm seeing this error just before the StartFrame reaches the end of the Pipeline (sharing the logline before and after for context):
> 
> I think this is fine and should be harmless. The reason is that we cancel the task from the default strategies right away while Python hasn't even had time to start them. That's why it only happens once now: just at the beggining we start the task and we cancel right away once we detect there's a turn analyzer in the transport. Before, we were causing 2 interruptions back to back that was causing this issue all the time (that was a FrameProcessor task).

I tested this theory by simply adding a small `asyncio.sleep(0.5)` before canceling that task `TranscriptionBotTurnStartStrategy` and the warning doesn't appear anymore.

markbackman:
Two PRs that should be merged into this one:
- Deepgram STT to push frames: https://github.com/pipecat-ai/pipecat/pull/3314
- Fix VoicemailDetector to work with external turn start strategies: https://github.com/pipecat-ai/pipecat/pull/3312

With these two fixes, I think I'll be ready to approve this one.

markbackman (approved):
LGTM! I've tested this a bunch already.

PR #3310, opened 2025-12-29
fix: preserve asyncio.Task return value in create_task

### LINKED ISSUE #2874: TaskManager does not return coroutine result
### pipecat version

0.0.86

### Python version

3.13

### Operating System

MacOS 15.6.1

### Issue description

When running `result = await self.create_task(some_coro())` in a `FrameProcessor`, the `result` is always `None`, even when `some_coro()` returns something else.

I suspect [this line in `TaskManager`](https://github.com/pipecat-ai/pipecat/blob/3c4807d7d4c56730d1da6b830a72abac7d7d480b/src/pipecat/utils/asyncio/task_manager.py#L159) should change from `await coroutine` to `return await coroutine`.

### Reproduction steps

```py
async def coro():
    return "foo"
result = await self.create_task(coro())
print(result)
```

in some `FrameProcessor`

### Expected behavior

`foo` is printed

### Actual behavior

`None` is printed

### Logs

```shell

```

omChauhanDev:
Fixes #2874 

Issue: `TaskManager` wraps coroutine but awaits it without returning the result. As a result, tasks created via create_task (including those in `FrameProcessor`) always resolve to None when awaited, breaking the expected asyncio.Task behavior.

Fix: Return the result of the awaited coroutine from the run_coroutine wrapper.

### DIFF
--- a/src/pipecat/utils/asyncio/task_manager.py
+++ b/src/pipecat/utils/asyncio/task_manager.py
@@ -157,7 +157,7 @@ class TaskManager(BaseTaskManager):
 
         async def run_coroutine():
             try:
-                await coroutine
+                return await coroutine
             except asyncio.CancelledError:
                 logger.trace(f"{name}: task cancelled")
                 # Re-raise the exception to ensure the task is cancelled.


aconchillo:
LGTM! Thank you!

aconchillo (approved):

PR #3312, opened 2025-12-29
Update VoicemailDetector to use ExternalTurnStartStrategies

markbackman:

This fixes the VoicemailDetector.

There are two user context aggregators in the Pipeline. It seems like a bad idea to have both generate UserStarted/StoppedSpeakingFrames and InterruptionFrames. Instead, I'm making the user context aggregator in the VoicemailDetector ParallelPipeline use ExternalTurnStrategies to treat the main user context aggregator as the source of truth.

To support that, the ClassificationProcessor needs to not block the UserStarted/StoppedSpeakingFrames, which previously didn't matter.

Note: @aconchillo this is a branch off of your `aleix/external-turn-start-strategies` branch, so feel free to approve and merge if you agree.


aconchillo:
Yep! This is it! LGTM!

PR #3314, opened 2025-12-29
DeepgramSTTService pushes user started/stopped speaking and interrupt…

markbackman:
…ion frames

This aligns DeepgramSTTService with other services. For anyone using  DeepgramSTTService with `vad_enabled=True`, this could mean extra frames are pushed. But, I can't think of a way around it. It's better to align the service to Pipecat's norms. This makes it easier to work with.


aconchillo:
Makes sense. LGTM!

PR #3319, opened 2025-12-30
BaseInputTransport: fix OpenAILLMContext backwards compatibility

aconchillo:

This PR fixes backwards compatibility when using the deprecated `OpenAILLMContext` and `LLMUserContextAggregator`.

### DIFF
--- a/src/pipecat/turns/user/external_user_turn_start_strategy.py
+++ b/src/pipecat/turns/user/external_user_turn_start_strategy.py
@@ -25,7 +25,7 @@ class ExternalUserTurnStartStrategy(BaseUserTurnStartStrategy):
         Args:
             **kwargs: Additional keyword arguments.
         """
-        super().__init__(enable_user_speaking_frames=False, **kwargs)
+        super().__init__(enable_interruptions=False, enable_user_speaking_frames=False, **kwargs)
 
     async def process_frame(self, frame: Frame):
         """Process an incoming frame to detect user turn start.


src/pipecat/turns/user/external_user_turn_start_strategy.py
```diff
@@ -25,7 +25,7 @@ def __init__(self, **kwargs):
         Args:
             **kwargs: Additional keyword arguments.
         """
-        super().__init__(enable_user_speaking_frames=False, **kwargs)
+        super().__init__(enable_interruptions=False, enable_user_speaking_frames=False, **kwargs)
```
aconchillo, line 28: This one is important! Since I split it into two variable we need to set both now.

src/pipecat/pipeline/task.py
```diff
@@ -885,3 +887,26 @@ def _print_dangling_tasks(self):
...
+        # NOTE(aleix): Remove when OpenAILLMContext/LLMUserContextAggregator is removed.
+        if self._find_deprecated_openai_llm_context(self._pipeline):
+            start_metadata["deprecated_openaillmcontext"] = True
+
+        # Update with user provided metadata.
+        start_metadata.update(self._params.start_metadata)
+
+        return start_metadata
+
+    def _find_deprecated_openai_llm_context(self, processor: FrameProcessor) -> bool:
```
markbackman: For naming consistency, maybe it should be:
```suggestion
    def _find_deprecated_openaillmcontext(self, processor: FrameProcessor) -> bool:
```

In frame_processor.py, you have `_deprecated_openaillmcontext`. (Or, update frame_processor.py to match).

src/pipecat/turns/user/external_user_turn_start_strategy.py
markbackman: Good catch!

src/pipecat/pipeline/task.py
aconchillo: Done!

Issue #3259, opened 2025-12-18
RNNoise issue: av.error.MemoryError: [Errno 12] Cannot allocate memory,


wuxuedaifu:
https://github.com/pipecat-ai/pipecat/blob/7f53483f6b636d892eaf3d76f31c1d214f6296f5/src/pipecat/audio/filters/rnnoise_filter.py#L29

When the input chunk is "bx000". mute sound, the resampler will return b"" empty byte audio_samples.

So the next step  self._rnnoise.denoise_chunk(audio_samples) will have can not allocate memory Error.

  File "/app/voice_services/RNNoiseFilter.py", line 153, in filter
    for speech_prob, denoised_frame in self._rnnoise.denoise_chunk(audio_samples):
                                       ~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^
  File "/app/.venv/lib/python3.13/site-packages/pyrnnoise/pyrnnoise.py", line 103, in denoise_chunk
    self.in_graph.push(chunk)
    ~~~~~~~~~~~~~~~~~~^^^^^^^
  File "/app/.venv/lib/python3.13/site-packages/audiolab/av/graph.py", line 85, in push
    super().push(frame)
    ~~~~~~~~~~~~^^^^^^^
  File "av/filter/graph.pyx", line 198, in av.filter.graph.Graph.push
  File "av/filter/context.pyx", line 98, in av.filter.context.FilterContext.push
  File "av/error.pyx", line 424, in av.error.err_check
av.error.MemoryError: [Errno 12] Cannot allocate memory

markbackman:
#3322 is merged. Closing.

PR #3309, opened 2025-12-29
RNNoiseFilter: don't filter silence

aconchillo:
#### Please describe the changes in your PR. If it is addressing an issue, please reference that as well.

Fixes #3259

PR #3322, opened 2025-12-30
Fix/rnnoise filter handle empty audio

gui217:
Fixes [https://github.com/pipecat-ai/pipecat/issues/3259](url)


gui217:
@aconchillo your fix on https://github.com/pipecat-ai/pipecat/pull/3309/files looks legit - all i did was adding a unit test to make sure that it fails before the implementation and pass after it 🙏

PR #3328, opened 2025-12-31
Update to SpeechmaticsSTTService for `0.0.99`

sam-s10s:
## What's Changed
- Fixed `SpeechmaticsSTTService` to listen for `VADUserStoppedSpeakingFrame` in order to finalize transcription.
- Default to `TurnDetectionMode.FIXED` for Pipecat-controlled end of turn detection.
- Only emit VAD + interruption frames if VAD is enabled within the plugin (modes other than `TurnDetectionMode.FIXED` or `TurnDetectionMode.EXTERNAL`).

### DIFF
--- a/src/pipecat/services/speechmatics/stt.py
+++ b/src/pipecat/services/speechmatics/stt.py
@@ -29,7 +29,6 @@ from pipecat.frames.frames import (
     TranscriptionFrame,
     UserStartedSpeakingFrame,
     UserStoppedSpeakingFrame,
-    VADUserStartedSpeakingFrame,
     VADUserStoppedSpeakingFrame,
 )
 from pipecat.processors.frame_processor import FrameDirection
@@ -615,7 +614,7 @@ class SpeechmaticsSTTService(STTService):
 
         The service will:
         - Send a BotInterruptionFrame upstream to stop bot speech
-        - Send a VADUserStartedSpeakingFrame downstream to notify other components
+        - Send a UserStartedSpeakingFrame downstream to notify other components
         - Start metrics collection for measuring response times
 
         Args:
@@ -636,7 +635,7 @@ class SpeechmaticsSTTService(STTService):
 
         The service will:
         - Stop processing metrics collection
-        - Send a VADUserStoppedSpeakingFrame to signal turn completion
+        - Send a UserStoppedSpeakingFrame to signal turn completion
 
         Args:
             message: the message payload.


src/pipecat/services/speechmatics/stt.py
```diff
@@ -660,10 +668,10 @@ async def process_frame(self, frame: Frame, direction: FrameDirection):
             self._bot_speaking = False
 
         # Force finalization
-        if isinstance(frame, UserStoppedSpeakingFrame):
+        if isinstance(frame, VADUserStoppedSpeakingFrame):
```
markbackman, line 680: This is the only change needed to fix the issue.

src/pipecat/services/speechmatics/stt.py
```diff
@@ -597,17 +597,21 @@ async def _handle_start_of_turn(self, message: dict[str, Any]) -> None:
...
         Args:
             message: the message payload.
         """
         logger.debug(f"{self} StartOfTurn received")
-        await self.broadcast_frame(UserStartedSpeakingFrame)
-        await self.push_interruption_task_frame_and_wait()
         # await self.start_processing_metrics()
 
+        # Emit VAD events if enabled
+        if self._enable_vad:
```
markbackman: This change isn't required. If using the now deprecated `end_of_utterance_mode`, the frames pushed are correct. You would use with with:
```
        context,
        user_params=LLMUserAggregatorParams(user_turn_strategies=ExternalUserTurnStrategies()),
    )
```

The same is true of the `_handle_end_of_turn` change. It's not needed for the same reason.

changelog/3328.fixed.md
```diff
@@ -0,0 +1,3 @@
+- Fix to `SpeechmaticsSTTService` for version `0.0.99+`:
```
markbackman: Should be something like this:

- Fixed `SpeechmaticsSTTService` to listen for `VADUserStoppedSpeakingFrame` in order to finalize transcription.

src/pipecat/services/speechmatics/stt.py
```diff
@@ -618,14 +638,18 @@ async def _handle_end_of_turn(self, message: dict[str, Any]) -> None:
...
 
         Args:
             message: the message payload.
         """
         logger.debug(f"{self} EndOfTurn received")
         # await self.stop_processing_metrics()
-        await self.broadcast_frame(UserStoppedSpeakingFrame)
+
+        # Emit VAD events if enabled
+        if self._enable_vad:
```
markbackman: This if statement isn't needed. This handler only runs when self._enable_vad is true, which is controlled earlier in the code:
```
        # Add listeners for VAD
        if self._enable_vad:
            self._client.on(AgentServerMessageType.START_OF_TURN, add_message)
            self._client.on(AgentServerMessageType.END_OF_TURN, add_message)
```

src/pipecat/services/speechmatics/stt.py
```diff
@@ -618,14 +638,18 @@ async def _handle_end_of_turn(self, message: dict[str, Any]) -> None:
...
+            logger.debug(f"{self} sending VADUserStoppedSpeakingFrame")
+            await self.broadcast_frame(VADUserStoppedSpeakingFrame)
```
markbackman: When Speechmatics is acting as the VAD, it should emit the `UserStoppedSpeakingFrame`. The previous implementation was correct.

src/pipecat/services/speechmatics/stt.py
markbackman: Also, the if check is also redundant here.

src/pipecat/services/speechmatics/stt.py
```diff
@@ -597,17 +613,21 @@ async def _handle_start_of_turn(self, message: dict[str, Any]) -> None:
...
         logger.debug(f"{self} StartOfTurn received")
-        await self.broadcast_frame(UserStartedSpeakingFrame)
-        await self.push_interruption_task_frame_and_wait()
         # await self.start_processing_metrics()
 
+        # Emit VAD events if enabled
+        if self._enable_vad:
+            await self.push_interruption_task_frame_and_wait()
+            logger.debug(f"{self} sending VADUserStartedSpeakingFrame")
+            await self.broadcast_frame(VADUserStartedSpeakingFrame)
```
markbackman: Same comment as for the _handle_end_of_turn function.

And, the frame ordering should be:
- First: `await self.broadcast_frame(UserStartedSpeakingFrame)`
- Then: `await self.push_interruption_task_frame_and_wait()`

So, no change recommended for this function.

src/pipecat/services/speechmatics/stt.py
```diff
@@ -230,6 +236,7 @@ class InputParams(BaseModel):
         end_of_utterance_max_delay: float | None = None
         punctuation_overrides: dict | None = None
         include_partials: bool | None = None
+        split_sentences: bool | None = None
```
markbackman, line 240: Is this a new arg you're supporting? If so, I'd recommend adding to the changelog.

src/pipecat/services/speechmatics/stt.py
```diff
@@ -201,7 +207,7 @@ class InputParams(BaseModel):
         language: Language | str = Language.EN
 
         # Endpointing mode
-        turn_detection_mode: TurnDetectionMode = TurnDetectionMode.EXTERNAL
+        turn_detection_mode: TurnDetectionMode = TurnDetectionMode.FIXED
```
markbackman, line 211: 👍

PR #3334, opened 2025-12-31
Add word-level timestamp support to Azure TTS with race condition fix

### LINKED ISSUE #2918: Feature Request: AzureTTSService should emit word boundaries via synthesis_word_boundary
### Problem Statement

Currently, Pipecat’s AzureTTSService does NOT emit word boundaries, even though the Azure SDK supports this via the `synthesis_word_boundary` event.

### Proposed Solution

Subscribe to the `synthesis_word_boundary` event in `AzureTTSService.start()`
- Handle each word boundary event and emit a `TTSTextFrame` with word-level text and the corresponding audio offset (timing)
- Use `evt.text` for the word and `evt.audio_offset` (divide by 10000 for ms) for synchronization

**Example implementation:**
```python
def _handle_word_boundary(self, evt):
    # evt.text contains the word
    # evt.audio_offset contains timing (in ticks, divide by 10000 for ms)
    pass

    await super().start(frame)
    # ... existing setup code ...
    self._speech_synthesizer.synthesis_word_boundary.connect(self._handle_word_boundary)
```

### Alternative Solutions

_No response_

### Additional Context

**References:**
- [Azure SDK sample for word boundaries](https://github.com/Azure-Samples/cognitive-services-speech-sdk/blob/master/samples/python/console/speech_synthesis_sample.py)

- [ ] No, I'm just suggesting

yuki901:

Closes #2918

Adds word-level timestamp support to AzureTTSService, enabling downstream components to synchronize with spoken words.

##  Credits
This PR completes the work started by https://github.com/remisharrock:
  - https://github.com/remisharrock/pipecat/pull/2

## Problem
Currently, Pipecat's AzureTTSService does NOT emit word boundaries, even though the Azure SDK supports this via the synthesis_word_boundary event.

## Solution

Based on @markbackman's suggestion in #2918 and @remisharrock's investigation:

- Changed base class to WordTTSService (not InterruptibleWordTTSService) because Azure SDK handles WebSocket connections internally, unlike Cartesia/ElevenLabs/Rime
- Subscribe to synthesis_word_boundary event in start()
- Handle word boundaries via _handle_word_boundary() 
- converting evt.audio_offset from ticks to seconds

## Additional fixes

During testing with multi-sentence responses, I encountered a race condition where word boundaries from concurrent sentences would interleave, causing scrambled transcripts. Fixed with:

- asyncio.Lock for sequential sentence processing
- Cumulative audio offset tracking for monotonically increasing PTS
- Missing await on start_word_timestamps()

## Test plan

- Tested with multi-sentence responses
- Verified word timestamps are correctly ordered
- Verified interruption handling works correctly

### DIFF
--- a/src/pipecat/services/azure/tts.py
+++ b/src/pipecat/services/azure/tts.py
@@ -13,6 +13,8 @@ from loguru import logger
 from pydantic import BaseModel
 
 from pipecat.frames.frames import (
+    CancelFrame,
+    EndFrame,
     ErrorFrame,
     Frame,
     InterruptionFrame,
@@ -25,7 +27,6 @@ from pipecat.processors.frame_processor import FrameDirection
 from pipecat.services.azure.common import language_to_azure_language
 from pipecat.services.tts_service import TTSService, WordTTSService
 from pipecat.transcriptions.language import Language
-from pipecat.utils.time import seconds_to_nanoseconds
 from pipecat.utils.tracing.service_decorators import traced_tts
 
 try:
@@ -279,6 +280,8 @@ class AzureTTSService(WordTTSService, AzureBaseTTSService):
         self._speech_config = None
         self._speech_synthesizer = None
         self._audio_queue = asyncio.Queue()
+        self._word_boundary_queue = asyncio.Queue()
+        self._word_processor_task = None
         self._started = False
         self._cumulative_audio_offset: float = 0.0  # Cumulative audio duration in seconds
 
@@ -316,8 +319,31 @@ class AzureTTSService(WordTTSService, AzureBaseTTSService):
         self._speech_synthesizer.synthesizing.connect(self._handle_synthesizing)
         self._speech_synthesizer.synthesis_completed.connect(self._handle_completed)
         self._speech_synthesizer.synthesis_canceled.connect(self._handle_canceled)
-        # Add word boundary event handler for word-level timestamps
         self._speech_synthesizer.synthesis_word_boundary.connect(self._handle_word_boundary)
+
+        # Start word processor task
+        if not self._word_processor_task:
+            self._word_processor_task = self.create_task(self._word_processor_task_handler())
+
+    async def stop(self, frame: EndFrame):
+        """Stop the Azure TTS service.
+
+        Args:
+            frame: End frame signaling service stop.
+        """
+        await super().stop(frame)
+        await self.cancel_task(self._word_processor_task)
+        self._word_processor_task = None
+
+    async def cancel(self, frame: CancelFrame):
+        """Cancel the Azure TTS service.
+
+        Args:
+            frame: Cancel frame signaling service cancellation.
+        """
+        await super().cancel(frame)
+        await self.cancel_task(self._word_processor_task)
+        self._word_processor_task = None
 
     def _handle_word_boundary(self, evt):
         """Handle word boundary events from Azure SDK.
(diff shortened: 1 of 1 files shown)


markbackman:
Do you mind linking to the Azure docs so I can use those for my review?

yuki901:
Hi @markbackman,

I investigated this with Claude Code and gathered the relevant Azure documentation for your review:

### Core Azure Speech SDK Documentation

1. **How to Synthesize Speech (Main Guide)** - Shows event subscription patterns including `WordBoundary`:
   - https://learn.microsoft.com/en-us/azure/ai-services/speech-service/how-to-speech-synthesis

2. **Python SpeechSynthesizer Class** - API reference showing `synthesis_word_boundary` signal:
   - https://learn.microsoft.com/en-us/python/api/azure-cognitiveservices-speech/azure.cognitiveservices.speech.speechsynthesizer?view=azure-python

3. **Python SpeechSynthesisWordBoundaryEventArgs** - Event args with `audio_offset`, `text`, `duration`, `boundary_type`:
   - https://learn.microsoft.com/en-us/python/api/azure-cognitiveservices-speech/azure.cognitiveservices.speech.speechsynthesiswordboundaryeventargs?view=azure-python

### Code Samples

4. **Official Python Sample** - `speech_synthesis_word_boundary_event()` function:
   - https://github.com/Azure-Samples/cognitive-services-speech-sdk/blob/master/samples/python/console/speech_synthesis_sample.py

5. **Azure TTS Wiki - Word Timestamps**:
   - https://github.com/Azure-Samples/Cognitive-Speech-TTS/wiki/How-to-get-word-time-stamp-events-using-Azure-TTS

### Key Points from Documentation

The Azure SDK handles WebSocket connections internally via `SpeechSynthesizer`. Word boundaries are accessed through high-level callbacks:

```python
speech_synthesizer.synthesis_word_boundary.connect(lambda evt: 
    print(f"Word: {evt.text}, Offset: {evt.audio_offset / 10000}ms"))
```

This is why `WordTTSService` fits better than `InterruptibleWordTTSService` - Azure doesn't expose WebSocket methods like `_connect_websocket()`, `_disconnect_websocket()`, or `_receive_messages()` that `InterruptibleWordTTSService` requires.

Let me know if you need any additional documentation!

markbackman:
Thanks for the links and the submission!

I spent some time looking at this and have some feedback:

- First, let's preserve the AzureBaseTTSService class. Perhaps we can treat it as more of a mixin since the two Azure service classes will require a different base TTSService class.
- Next, there's no need to use a lock for the run_tts. Pipecat already guarantees that audio will play out correctly. You just need to pass text to run_tts and the TTSAudioRawFrame playout is handled by the base class.
- I noticed some state issues related to handling the timestamps. We should be able to listen to audio and see each word printed out as the word is said.

Here's a WIP version I played around with while reviewing, which includes fixes for all of these items (plus a few other minor things):
[tts.py](https://github.com/user-attachments/files/24442928/tts.py)

This is nearly working as I would like with one exception. For some reason, the timing for the first bot turn is slightly off.

To test these files, I use a DebugLogObserver to print TTSTextFrames to the console. These frames contain words which should be timed to the audio playout:
[07f-interruptible-azure.py](https://github.com/user-attachments/files/24442935/07f-interruptible-azure.py)

I'd like to get the initial bot turn timing working correctly, but I thought I would share my progress in case you want to finish out the final changes. I'm not sure what the root cause is yet...

Lastly, don't forget to add a changelog fragment:

yuki901:
Thanks for the detailed review and the WIP version!

I replaced my implementation with your WIP version and made the following modification to fix the first turn timing issue:

The fix: Call start_word_timestamps() when the first audio chunk arrives, not before synthesis starts.
```python
if not self._started:
    yield TTSStartedFrame()
    self._started = True
    self._first_chunk = True
    self._cumulative_audio_offset = 0.0

ssml = self._construct_ssml(text)
self._speech_synthesizer.speak_ssml_async(ssml)

while True:
    chunk = await self._audio_queue.get()
    if chunk is None:
        break

    # Start word timestamps when first chunk arrives
    if self._first_chunk:
        await self.start_word_timestamps()
        self._first_chunk = False

    yield TTSAudioRawFrame(...)
```
[tts_fixed.py](https://github.com/user-attachments/files/24464029/tts_fixed.py)

This ensures the base timestamp is captured at the exact moment audio starts flowing downstream, aligning word timing with actual playback.

Also added the changelog fragment.

markbackman:
Ok, this was a tricky one! The base WordTTSService requires that you use the `add_word_timestamp()` method as the input is cached and timing is important. Without this, the timing just won't be right.

The solution is to use a asyncio.Queue() to hold the words and an asyncio task (which Pipecat has its own TaskManager class that's used) to turn the synchronous response to async, matching what's required by the add_word_timestamp method.

Here's the solution:
[tts.py](https://github.com/user-attachments/files/24476872/tts.py)

I've tested this with the foundational file I shared and see the word timing line up correctly now. With this, I think we're ready to approve/merge.

Sorry, this was a tricky issue due to the Azure SDK being synchronous.

yuki901:
Thanks for the final solution! I updated the file accordingly.

remisharrock:
WOW congratulations and thank you very much for this work everybody !

PR #3343, opened 2026-01-03
fix: keeping the Aggregator and Service states synchronized.

### LINKED ISSUE #3342: If a function call handler does not await its FunctionCallParams.result_callback, it will perpetually remain IN_PROGRESS in LLMAssistantAggregator
### pipecat version

0.0.97

### Python version

3.12.10

### Operating System

MacOS

### Issue description

When implementing a function call handler, if one doesn't need the function call to return a result for the LLM, it feels intuitive to just return from the function handler vs calling `FunctionCallParams.result_callback`.

However, if the function is implemented without an await on the `result_callback`, the LLM service running the function will never receive the `function_call_result_callback` call and will not broadcast a corresponding `FunctionCallResultFrame` for the function call.

If the pipeline contains an `LLMAssistantAggregator`, that aggregator will add the function call to its `_function_calls_in_progress` dict when the function call starts. However, in this case it will never receive the `FunctionCallResultFrame` and so the tool call will never be removed from its `_function_calls_in_progress`.

This has multiple side effects.

One such undesirable effect is that future function calls that _do_ return a result will not trigger an automatic response from the LLM, because `has_function_calls_in_progress` will perpetually return True for the `LLMAssistantAggregator`.

### Reproduction steps

1. Set up a pipeline that includes an LLM and an LLMAssistantAggregator
2. Add a tool implementation to the pipeline that does not await `FunctionCallParams.result_callback`.
3. Run the pipeline and speak dialog so that the tool will trigger.
4. See that the tool's status in the context remains IN_PROGRESS for the remainder of execution.
5. Via debugger, inspect `LLMAssistantAggregator` and see that the function call remains in its `_function_calls_in_progress`.

### Expected behavior

If a function call implementation doesn't await `FunctionCallParams.result_callback`, the tool status should be marked as COMPLETED after execution. 

This is the equivalent of having the tool do:

`await params.result_callback(None)` 

in the current pipecat implementation, which will trigger all appropriate cleanup and lead to the tool call status being updated to COMPLETED.

### Actual behavior

Actual behavior is that the function call remains in the IN_PROGRESS status for the remainder of execution.

### Logs

```shell

```

omChauhanDev:

Fixes #3342 

Issue : 
Currently, if a function handler returns without awaiting `FunctionCallParams.result_callback`:
- `LLMService` task finishes & is removed from tracking (breaking cancellation).
- `LLMAssistantAggregator` never receives a result frame and remains stuck in `IN_PROGRESS`.

This causes the pipeline to silently deadlock for that specific tool context, preventing future interaction as described [here](https://github.com/pipecat-ai/pipecat/issues/3342) in detail.

Fix : 
- If the handler returns and the callback was not invoked, the service now automatically awaits `result_callback(None)`.
This ensures the `FunctionCallResultFrame` is always sent, keeping the Aggregator and Service states synchronized.

Tested using : examples/foundational/14e-function-calling-google.py
image

#### Before (freezed - no response)
image

#### After 
image

get_weather (the "void" handler) now correctly unblocks the pipeline instead of freezing it.

### DIFF
--- a/src/pipecat/services/llm_service.py
+++ b/src/pipecat/services/llm_service.py
@@ -595,10 +595,14 @@ class LLMService(AIService):
             cancel_on_interruption=item.cancel_on_interruption,
         )
 
+        callback_executed = False
+
         # Define a callback function that pushes a FunctionCallResultFrame upstream & downstream.
         async def function_call_result_callback(
             result: Any, *, properties: Optional[FunctionCallResultProperties] = None
         ):
+            nonlocal callback_executed
+            callback_executed = True
             await self.broadcast_frame(
                 FunctionCallResultFrame,
                 function_name=runner_item.function_name,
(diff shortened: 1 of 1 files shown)


aconchillo:
LGTM! Thank you @omChauhanDev !

PR #3345, opened 2026-01-04
Add trailing space to DeepgramTTSService text generation

### LINKED ISSUE #3242: DeepgramTTSService (WebSocket): Missing space between consecutive Speak messages causes mispronunciation
## Description

When using `DeepgramTTSService` (WebSocket-based), consecutive sentences are mispronounced because Deepgram's internal buffer concatenates `Speak` messages without spaces between them.

For example, when the LLM generates:
- Sentence 1: `"Perfect, 12345."`
- Sentence 2: `"Let me check that we're in your area."`

Deepgram TTS pronounces this as: *"Perfect, 12345 **dot l e t** me check that we're in your area"* - spelling out "let" because it sees `12345.Let` as a single token.

## Root Cause

Two factors combine to cause this:

### 1. `SimpleTextAggregator` strips trailing spaces

In `pipecat/utils/text/simple_text_aggregator.py:68`:
```python
return Aggregation(text=result.strip(), type=AggregationType.SENTENCE)
```

This removes the trailing space after sentence-ending punctuation, so `"Perfect, 12345. "` becomes `"Perfect, 12345."`.

### 2. `DeepgramTTSService` doesn't add spacing between Speak messages

In `pipecat/services/deepgram/tts.py:286-287`:
```python
speak_msg = {"type": "Speak", "text": text}
await self._get_websocket().send(json.dumps(speak_msg))
```

Consecutive `Speak` messages are sent to the same WebSocket connection. Deepgram accumulates these in an internal buffer, concatenating them directly without adding spaces.

The result: Deepgram's buffer contains `"Perfect, 12345.Let me check..."` (no space after the period).

## Reproduction

1. Use `DeepgramTTSService` (WebSocket-based, not HTTP)
2. Have the LLM generate a response with multiple sentences where one ends with a number followed by a period
3. Observe the TTS output - the first word of the next sentence will be mispronounced

Logs showing the issue:
```
2025-12-14 23:46:11.281 | DEBUG | Generating TTS [Perfect, 12345.]
2025-12-14 23:46:11.320 | DEBUG | Generating TTS [Let me check that we're in your area.]
```

## Suggested Fix

Either:

1. **Don't strip trailing spaces in the aggregator** - Modify `SimpleTextAggregator` to preserve trailing whitespace after sentence-ending punctuation
2. **Add spacing in the TTS service** - Have `DeepgramTTSService` prepend a space to text or send a space between consecutive Speak messages
3. **Document the behavior** - If this is intentional, document that users should add trailing spaces to TTS text

## Environment

- Pipecat version: latest (installed via pip)
- TTS Service: `DeepgramTTSService` (WebSocket)
- Deepgram voice: aura-2 series

markbackman:

In some cases, `DeepgramTTSService` outputs a "Dot" when a period occurs. For example, this is reproducible when a sentence ends in a number and `.` (e.g. The secret code is 1234.)

Deepgram expects spaces to be preserved, but Pipecat strips them in the aggregation logic. This fix is targeted to the DeepgramTTSService class to add a trailing space. DeepgramTTSService is a TTSService class, so no word-timestamp alignment data exists. That means this added space is only for audio generation and is not added to the context.

Fixes #3242

### DIFF
--- a/changelog/3345.fixed.md
+++ b/changelog/3345.fixed.md
@@ -1,0 +1,1 @@
+- Fixed an issue with DeepgramTTSService where the model would output "Dot" instead of a period in some circumstances.
--- a/src/pipecat/services/deepgram/tts.py
+++ b/src/pipecat/services/deepgram/tts.py
@@ -287,7 +287,9 @@ class DeepgramTTSService(WebsocketTTSService):
         Yields:
             Frame: Audio frames containing the synthesized speech, plus start/stop frames.
         """
-        logger.debug(f"{self}: Generating TTS [{text}]")
+        # Append trailing space to prevent TTS from vocalizing trailing periods as "dot"
+        text_with_trailing_space = text + " "
+        logger.debug(f"{self}: Generating TTS [{text_with_trailing_space}]")
 
         try:
             # Reconnect if the websocket is closed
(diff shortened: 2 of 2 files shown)


mattieruth:
Actually, I'm thinking we should modify this so that we only add a space if the last character is a dot. Unless it's an issue for other characters too?

markbackman:
> Actually, I'm thinking we should modify this so that we only add a space if the last character is a dot. Unless it's an issue for other characters too?

Deepgram effectively concatenates the received text, so a space after all inputs is probably the safest call and will yield the right output.

PR #3352, opened 2026-01-05
fix: prevent unnecessary reconnections for unchanged config

mhussam-ai:
fix: prevent unnecessary reconnections for unchanged config

Previously, the client forced a reconnect whenever `system_instruction` or `tools` were present. This change adds equality checks against the initialization values to ensure we only reconnect if the configuration has actually changed.

### DIFF
--- a/src/pipecat/services/google/gemini_live/llm.py
+++ b/src/pipecat/services/google/gemini_live/llm.py
@@ -975,7 +975,22 @@ class GeminiLiveLLMService(LLMService):
                 logger.warning(
                     "Tools provided both at init time and in context; using context-provided value."
                 )
-            if system_instruction or tools:
+            # if system_instruction or tools:
+            #     await self._reconnect()
+
+            should_reconnect = False
+            
+            if system_instruction and system_instruction != self._system_instruction_from_init:
+                should_reconnect = True
+            
+            if tools:
+                # Check if tools have changed by comparing the underlying ToolsSchema from context
+                # against the ToolsSchema provided at initialization.
+                context_tools = self._context.tools if self._context else None
+                if context_tools != self._tools_from_init:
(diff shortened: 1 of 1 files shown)


markbackman:
@kompfner can you please take a look when you get a moment?

src/pipecat/services/google/gemini_live/llm.py
kompfner, line 971: Here we log warnings about providing system instructions and/or tools at _both_ init time and via the context object, which we're considering a misconfiguration.

I personally don't think we need or ought to do anything to optimize the misconfiguration case.

Unless, of course, there's a scenario I'm not thinking of other than the misconfiguration called out here, in which case do let me know!

kompfner: Ah, OK, now I remember. You are right that we are actually doing a reconnection in all cases where system instruction or tools are passed into the context object rather than as `GeminiLiveLLMService` init params.

The challenge is that the connection to Gemini is established when the pipeline boots up, _before_ the `GeminiLiveLLMService` in the pipeline receives the context object.

We actually went back and forth on this issue for a while, and, after some testing, decided that the one-time reconnection at the start of the call was worth it (it's usually negligible) to ensure that the same pattern that worked when using other LLM services (providing system instruction and tools through the context object) worked with Gemini Live as well.

Today, if you really want to avoid that initial reconnection, you can simply pass system instruction and tools through as `GeminiLiveLLMService` init params. Is there some reason that option doesn't work for you, and you need to have the system instruction and tools also specified in the context?

markbackman:
Closing this. The underlying issue has been addressed on main. We now compare the system instruction against the init-provided value before reconnecting (see `_handle_context` in `gemini_live/llm.py`). Thanks for flagging this!

PR #3355, opened 2026-01-05
Add `user_bot_latency_seconds` to OpenTelemetry turn spans

itsderek23:
This is intended to provide more insights into potential problems during a conversation by adding a `user_bot_latency_seconds` attribute to OpenTelemetry turn spans.

I created `UserBotLatencyObserver` to abstract the actual latency tracking so it can be reused by the existing `UserBotLatencyLogObserver` and included directly in `TurnTraceObserver`, following the same pattern as `TurnTrackingObserver`.

Below is a screenshot showing the `user_bot_latency_seconds` being rendered in [Finchvox](https://github.com/itsderek23/finchvox/):

image

Here is the raw span as well:

```json
{
  "trace_id": "hKgttwU0ynjgpQw6q0OgeA==",
  "span_id": "+0pOdMKLd3k=",
  "parent_span_id": "pbym7ya48ko=",
  "name": "turn",
  "kind": "SPAN_KIND_INTERNAL",
  "start_time_unix_nano": "1767551104418988000",
  "end_time_unix_nano": "1767551126286140000",
  "attributes": [
    {
      "key": "turn.number",
      "value": {
        "int_value": "3"
      }
    },
    {
      "key": "turn.type",
      "value": {
        "string_value": "conversation"
      }
    },
    {
      "key": "conversation.id",
      "value": {
        "string_value": "212b72b0-91a8-47d4-b2cb-cd4c32304d92"
      }
    },
    {
      "key": "turn.user_bot_latency_seconds",
      "value": {
        "double_value": 4.258035898208618
      }
    },
    {
      "key": "turn.duration_seconds",
      "value": {
        "double_value": 21.867765208
      }
    },
    {
      "key": "turn.was_interrupted",
      "value": {
        "bool_value": true
      }
    }
  ],
  "status": {},
  "flags": 256,
  "trace_id_hex": "84a82db70534ca78e0a50c3aab43a078",
  "span_id_hex": "fb4a4e74c28b7779",
  "parent_span_id_hex": "a5bca6ef26b8f24a",
  "resource": {
    "attributes": [
      {
        "key": "telemetry.sdk.language",
        "value": {
          "string_value": "python"
        }
      },
      {
        "key": "telemetry.sdk.name",
        "value": {
          "string_value": "opentelemetry"
        }
      },
      {
        "key": "telemetry.sdk.version",
        "value": {
          "string_value": "1.39.1"
        }
      },
      {
        "key": "service.name",
        "value": {
          "string_value": "switch-voices"
        }
      },
      {
        "key": "service.instance.id",
        "value": {
          "string_value": "unknown"
        }
      },
      {
        "key": "deployment.environment",
        "value": {
          "string_value": "development"
        }
      }
    ]
  },
  "instrumentation_scope": {
    "name": "pipecat.turn"
  }
}
```

### DIFF
--- a/changelog/3355.added.md
+++ b/changelog/3355.added.md
@@ -1,0 +1,1 @@
+- Added `UserBotLatencyObserver` for tracking user-to-bot response latency. When tracing is enabled, latency measurements are automatically recorded as `turn.user_bot_latency_seconds` attributes on OpenTelemetry turn spans.
--- a/changelog/3355.deprecated.md
+++ b/changelog/3355.deprecated.md
@@ -1,0 +1,1 @@
+- Deprecated `UserBotLatencyLogObserver`. Use `UserBotLatencyObserver` directly with its `on_latency_measured` event handler instead.
--- a/examples/foundational/29-turn-tracking-observer.py
+++ b/examples/foundational/29-turn-tracking-observer.py
@@ -15,6 +15,7 @@ from pipecat.audio.vad.silero import SileroVADAnalyzer
 from pipecat.audio.vad.vad_analyzer import VADParams
 from pipecat.frames.frames import LLMRunFrame
 from pipecat.observers.loggers.user_bot_latency_log_observer import UserBotLatencyLogObserver
+from pipecat.observers.user_bot_latency_observer import UserBotLatencyObserver
 from pipecat.pipeline.pipeline import Pipeline
 from pipecat.pipeline.runner import PipelineRunner
 from pipecat.pipeline.task import PipelineParams, PipelineTask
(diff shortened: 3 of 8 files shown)


markbackman (approved):
This is really great! It's been on our to do list for quite some time, so this is majorly helpful.

Rather than keep both classes around, I pushed a commit that reverts the `UserBotLatencyLogObserver` to the original code and deprecates it. Your new observer is more capable, so we should keep it and let it serve both purposes: 1) for tracing, 2) for debug logging.

I also rebased on `main`, so this is ready to go. Thanks again for this! 🙌

itsderek23:
Awesome - thanks! I verified this works again with a few examples. Looks good!

The chart below is derived from the new `turn.user_bot_latency_seconds` span attribute:

image

markbackman:
Retested before merging. All looks well with the latency measurement.

Issue #3350, opened 2026-01-05
GeminiLiveLLMService does not return user transcription


nipun-dezoysa:
### pipecat version

0.0.96

### Python version

3.12.3

### Operating System

Ubuntu 24.04.3 LTS

### Issue description

I used Gemini Live model without using stt. when i enabled transcription option only the assistant transcription is only return from the gemini model. but sometimes it works fine. I also tried to capture user transaction using custom frame processor but looks like it not return at all.
Is there anyway to enable transcribe audio option like previous gemini multimodel library version

### Reproduction steps

Codes that I used in my project:

```
  llm = GeminiLiveLLMService(
        api_key=gemini_api_key,
        voice_id=voice_id,
        model=os.getenv("GEMINI_MODEL", "models/gemini-2.5-flash-native-audio-preview-09-2025"),
        system_instruction=system_instructions,
    )
```

```
transcript = TranscriptProcessor()

    pipeline = Pipeline(
        [
            transport.input(),  # Transport user input
            rtvi,  # RTVI processor
            context_aggregator.user(),  # User responses
            transcript.user(),  # User transcript
            llm,  # LLM
            transcript.assistant(),  # Assistant transcript
            transport.output(),  # Transport bot output
            context_aggregator.assistant(),  # Assistant spoken responses
        ]
    )
```

```
@transcript.event_handler("on_transcript_update")
    async def on_transcript_update(processor, frame):
        for msg in frame.messages:
            print(f"[{msg.role.upper()}]: {msg.content}")


    @transport.event_handler("on_client_disconnected")
    async def on_client_disconnected(transport, client):
        print("\n--- client disconnected ---")

        user_msgs = transcript.user()._processed_messages
        assistant_msgs = transcript.assistant()._processed_messages
        full_history = user_msgs + assistant_msgs
        
        # Sort by timestamp
        full_history.sort(key=lambda m: m.timestamp if m.timestamp else "")

        for entry in full_history:
            print(f"({entry.timestamp}) {entry.role.upper()}: {entry.content}")
        
        await task.cancel()
```

### Expected behavior

--- client disconnected ---
(2026-01-05T07:51:24.794+00:00) ASSISTANT: Hello! How can I help you?
(2026-01-05T07:51:29.582+00:00) USER: Hello, what is your name?
(2026-01-05T07:51:32.692+00:00) ASSISTANT: I am an AI assistant.

### Actual behavior

--- client disconnected ---
[...]

Issue #2206, opened 2025-07-15
GeminiMultiModal not emitting TranscriptionFrames correctly


zialtawi:
### pipecat version

0.0.76

### Python version

3.12

### Operating System

Ubuntu 24.04.2 

### Issue description

In the GeminiMultimodalLiveLLMService class, the _handle_evt_input_transcription method buffers incoming transcribed text and uses the "match_endofsentence" function from the String library to detect sentence boundaries. This approach is unreliable because the transcriber often does not append punctuation marks such as "." or "!" at the end of a user's speech segment. As a result, the method may not push a TranscriptionFrame as expected and continues buffering, which can cause multiple user turns (sometimes 3-4) to be merged into a single TranscriptionFrame. This leads to mismatches in conversation logging and inaccurate representation of user input turns.

### Reproduction steps

1. Run a pipeline using GeminiMultiModal and TranscriptProcessor() (foundational example : 26a-gemini-multimodal-live-transcription.py)

2. Speak normally you'll notice some transcriptions aren't showing or logging

### Expected behavior

The transcription should be outputted even if it didn't contain a "." (period) , "!" , "?" for reliable logging and observation of transcription.

### Actual behavior

No transcriptions, no logs , Multiple user turns in one transcription  

### Logs

```shell

```

markbackman:
Gemini's `evt.serverContent.inputTranscription` event emits individual words, like this:
```
Received transcription:  Hey
Received transcription: ,
Received transcription:  I've
Received transcription:  go
Received transcription: t a
Received transcription:  question
Received transcription: .
Received transcription:  Um
Received transcription: ,
Received transcription:  can
Received transcription:  you
Received transcription:  tell
Received transcription:  me
Received transcription:  uh story?
```

This is an output from Gemini and used only as an artifact for creating a local context for Pipecat. Rather than include each individual word as a message, we collect words by sentences, so that the context contains complete input from the user. (The same goes for the TranscriptProcessor, which uses this data.)

In my experience testing Gemini, which is fairly extensive, the model always punctuates. Though, I wouldn't be surprised if it randomly decides to do something different, like output a foreign language as a transcript (I've seen that!). Would you mind sharing the settings that you're using to hit this case? Also, how frequently are you impacted by this?

For the record, here's the file I tested with before writing this response:
https://github.com/pipecat-ai/pipecat/blob/main/examples/foundational/26a-gemini-multimodal-live-transcription.py

zialtawi:
[...]

PR #3356, opened 2026-01-05
Add timeout for handling user transcript messages

markbackman:

Gemini Live doesn't always output punctuated messages. As a result, that means no TranscriptionFrame is pushed. This is more common with non-English languages and incomplete English sentences.

In this branch, I'm adding a timeout (0.5 sec) to allow for buffering transcription messages and emitting a TranscriptionFrame after the timeout.

I'm open to other ways of solving this one, too!

Fixes: 
- https://github.com/pipecat-ai/pipecat/issues/1994
- https://github.com/pipecat-ai/pipecat/issues/2206
- https://github.com/pipecat-ai/pipecat/issues/3350

### DIFF
--- a/src/pipecat/services/google/gemini_live/llm.py
+++ b/src/pipecat/services/google/gemini_live/llm.py
@@ -1569,6 +1569,27 @@ class GeminiLiveLLMService(LLMService):
         """Handle a transcription result with tracing."""
         pass
 
+    async def _push_user_transcription(self, text: str, result: Optional[LiveServerMessage] = None):
+        """Push a user transcription frame upstream.
+
+        Helper method to ensure consistent handling of user transcriptions
+        from both punctuation-based and timeout-based paths.
+
+        Args:
+            text: The transcription text to push
+            result: Optional LiveServerMessage that triggered this transcription
+        """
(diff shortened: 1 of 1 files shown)


src/pipecat/services/google/gemini_live/llm.py
```diff
@@ -1564,12 +1569,49 @@ async def _handle_user_transcription(
...
+
+            # If we still have buffered text after timeout, flush it
+            if self._user_transcription_buffer:
+                logger.trace(
+                    f"[Transcription:user:timeout] Flushing buffer: [{self._user_transcription_buffer}]"
+                )
+                complete_sentence = self._user_transcription_buffer
+                self._user_transcription_buffer = ""
+
+                await self._handle_user_transcription(
```
kompfner: Optional: factor out the call to `_handle_user_transcription` followed by pushing the transcription frame into a helper method used by both `_transcription_timeout_handler` and `_handle_msg_input_transcription`, since there's juuuust enough logic that it might be worth guaranteeing they don't get out of sync someday.

markbackman: Good call! Adding a utility function to handle this.

PR #3360, opened 2026-01-06
Add video input (e.g. image input) support for OpenAI Realtime

markbackman:

OpenAI supports image input, which this PR adds support for. This borrows heavily from the GeminiLiveLLMService implementation.

### DIFF
--- a/examples/foundational/19c-openai-realtime-live-video.py
+++ b/examples/foundational/19c-openai-realtime-live-video.py
@@ -5,7 +5,6 @@
 #
 
 
-import asyncio
 import os
 
 from dotenv import load_dotenv
@@ -131,8 +130,8 @@ Remember, your responses should be short. Just one or two sentences, usually. Re
     async def on_client_connected(transport, client):
         logger.info(f"Client connected: {client}")
 
-        await maybe_capture_participant_camera(transport, client, framerate=1)
-        await maybe_capture_participant_screen(transport, client, framerate=1)
+        await maybe_capture_participant_camera(transport, client, framerate=0.5)
+        await maybe_capture_participant_screen(transport, client, framerate=0.5)
 
         await task.queue_frames([LLMRunFrame()])
 
--- a/src/pipecat/services/openai/realtime/llm.py
+++ b/src/pipecat/services/openai/realtime/llm.py
@@ -109,8 +109,8 @@ class OpenAIRealtimeLLMService(LLMService):
         base_url: str = "wss://api.openai.com/v1/realtime",
         session_properties: Optional[events.SessionProperties] = None,
         start_audio_paused: bool = False,
-        start_image_paused: bool = False,
-        image_detail: str = "auto",
+        start_video_paused: bool = False,
+        video_frame_detail: str = "auto",
         send_transcription_frames: Optional[bool] = None,
         **kwargs,
     ):
@@ -127,8 +127,9 @@ class OpenAIRealtimeLLMService(LLMService):
                 These are session-level settings that can be updated during the session
                 (except for voice and model). If None, uses default SessionProperties.
             start_audio_paused: Whether to start with audio input paused. Defaults to False.
-            start_image_paused: Whether to start with image input paused. Defaults to False.
-            image_detail: Detail level for image processing. Can be "auto", "low", or "high".
+            start_video_paused: Whether to start with video input paused. Defaults to False.
+            video_frame_detail: Detail level for video processing. Can be "auto", "low", or "high".
+                This sets the image_detail parameter in the OpenAI Realtime API.
                 "auto" lets the model decide, "low" is faster and uses fewer tokens,
                 "high" provides more detail. Defaults to "auto".
             send_transcription_frames: Whether to emit transcription frames.
@@ -165,9 +166,9 @@ class OpenAIRealtimeLLMService(LLMService):
             session_properties or events.SessionProperties()
         )
         self._audio_input_paused = start_audio_paused
-        self._image_input_paused = start_image_paused
-        self._image_detail = image_detail
-        self._last_image_sent_time = 0
+        self._video_input_paused = start_video_paused
+        self._video_frame_detail = video_frame_detail
+        self._last_sent_time = 0
         self._websocket = None
         self._receive_task = None
         self._context: LLMContext = None
@@ -205,24 +206,24 @@ class OpenAIRealtimeLLMService(LLMService):
         """
         self._audio_input_paused = paused
 
-    def set_image_input_paused(self, paused: bool):
-        """Set whether image input is paused.
+    def set_video_input_paused(self, paused: bool):
+        """Set whether video input is paused.
 
         Args:
-            paused: True to pause image input, False to resume.
+            paused: True to pause video input, False to resume.
         """
-        self._image_input_paused = paused
+        self._video_input_paused = paused
 
-    def set_image_detail(self, detail: str):
-        """Set the detail level for image processing.
+    def set_video_frame_detail(self, detail: str):
+        """Set the detail level for video processing.
 
         Args:
             detail: Detail level - "auto", "low", or "high".
         """
         if detail not in ["auto", "low", "high"]:
-            logger.warning(f"Invalid image detail '{detail}', must be 'auto', 'low', or 'high'")
+            logger.warning(f"Invalid video detail '{detail}', must be 'auto', 'low', or 'high'")
             return
-        self._image_detail = detail
+        self._video_frame_detail = detail
 
     def _is_modality_enabled(self, modality: str) -> bool:
         """Check if a specific modality is enabled, "text" or "audio"."""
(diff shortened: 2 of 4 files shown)


src/pipecat/services/openai/realtime/llm.py
```diff
@@ -847,6 +881,49 @@ async def _send_user_audio(self, frame):
...
+        """Send user image frame to OpenAI Realtime API.
+
+        Args:
+            frame: The input image frame to send.
+        """
+        if self._image_input_paused or self._disconnecting or not self._websocket:
+            return
+
+        now = time.time()
+        if now - self._last_image_sent_time < 1:
```
markbackman: Same rate limit that Gemini Live applies.

src/pipecat/services/openai/realtime/llm.py
```diff
@@ -106,6 +109,8 @@ def __init__(
         base_url: str = "wss://api.openai.com/v1/realtime",
         session_properties: Optional[events.SessionProperties] = None,
         start_audio_paused: bool = False,
+        start_image_paused: bool = False,
+        image_detail: str = "auto",
```
markbackman: I'm opting to include an arg for `image_detail`, so that this can be controlled by developers.

src/pipecat/services/openai/realtime/llm.py
```diff
@@ -106,6 +109,8 @@ def __init__(
         base_url: str = "wss://api.openai.com/v1/realtime",
         session_properties: Optional[events.SessionProperties] = None,
         start_audio_paused: bool = False,
+        start_image_paused: bool = False,
```
markbackman: Gemini Live calls this `start_video_paused`. Maybe we should do the same? I used "image" because OpenAI Realtime talks about image input support. WDYT?

src/pipecat/services/openai/realtime/llm.py
```diff
@@ -193,6 +205,25 @@ def set_audio_input_paused(self, paused: bool):
...
 
+    def set_image_input_paused(self, paused: bool):
+        """Set whether image input is paused.
+
+        Args:
+            paused: True to pause image input, False to resume.
+        """
+        self._image_input_paused = paused
+
+    def set_image_detail(self, detail: str):
```
markbackman: Following the pattern of including methods to control these settings, but we should really use an `UpdateSettingsFrame`. I suppose that's a future TODO.

src/pipecat/services/openai/realtime/llm.py
```diff
@@ -847,6 +881,49 @@ async def _send_user_audio(self, frame):
...
+        """
+        if self._image_input_paused or self._disconnecting or not self._websocket:
+            return
+
+        now = time.time()
+        if now - self._last_image_sent_time < 1:
+            return  # Ignore if less than 1 second has passed
+
+        self._last_image_sent_time = now  # Update last sent time
+        logger.debug(f"Sending image frame to OpenAI Realtime: {frame}")
```
markbackman: This log is a bit chatty, but I've included because Gemini Live did. Maybe we should set both to trace?

examples/foundational/19c-openai-realtime-live-video.py
```diff
@@ -0,0 +1,163 @@
...
+You are participating in a voice conversation. Keep your responses concise, short, and to the point
+unless specifically asked to elaborate on a topic.
+
+Remember, your responses should be short. Just one or two sentences, usually. Respond in English.""",
+    )
+
+    llm = OpenAIRealtimeLLMService(
+        api_key=os.getenv("OPENAI_API_KEY"),
+        session_properties=session_properties,
+        start_audio_paused=False,
```
kompfner: Nit: this is the default, I believe. Feels odd to have to explicitly say "start with audio". I realize this is an issue with our other `OpenAIRealtimeLLMService` examples, too.

examples/foundational/19c-openai-realtime-live-video.py
```diff
@@ -0,0 +1,163 @@
...
+    async def on_client_connected(transport, client):
+        logger.info(f"Client connected: {client}")
+
+        await maybe_capture_participant_camera(transport, client, framerate=1)
+        await maybe_capture_participant_screen(transport, client, framerate=1)
+
+        await task.queue_frames([LLMRunFrame()])
+        await asyncio.sleep(3)
+        logger.debug("Unpausing audio and video")
+        llm.set_audio_input_paused(False)
```
kompfner: Why do you need to unpause audio? Doesn't it start off in the right state?

src/pipecat/services/openai/realtime/llm.py
kompfner: I was wondering why the divergence from the Gemini Live terminology. One argument for referring to it as "video": it sounds odd to talk about "pausing" an image, which is a single-shot, static thing.

kompfner: Though...if OpenAI wants you to only send up to 1 image per second, it really is more like sending a series of static images...

How exactly does the model handle these images? It doesn't generate a response directly after receiving an image, right, it still expects the user to ask it something about the image it received? If it also handles this input similarly to Gemini Live, that could be another argument for aligning on terminology.

src/pipecat/services/openai/realtime/llm.py
kompfner: Agree with this approach—consistency for now, update everything at once later.

src/pipecat/services/openai/realtime/llm.py
kompfner: No strong opinion on this one.

src/pipecat/services/openai/realtime/llm.py
kompfner: Makes sense! Just read the OpenAI docs on this, and it seems valuable to expose this to developers.

src/pipecat/services/openai/realtime/llm.py
kompfner: > How exactly does the model handle these images? It doesn't generate a response directly after receiving an image, right, it still expects the user to ask it something about the image it received?

My assumption is that it handles image inputs in the same way as Gemini Live.

src/pipecat/services/openai/realtime/llm.py
markbackman: To clarify, I can't find an actual applied rate limit by OpenAI Realtime, but the higher the frame rate of the video, the slower the  response time from the model. I think 1 per sec is reasonable.

src/pipecat/services/openai/realtime/llm.py
markbackman: > How exactly does the model handle these images? It doesn't generate a response directly after receiving an image, right, it still expects the user to ask it something about the image it received?

Correct, a user prompt is required for the image to be used. I believe this is the same as Gemini, right?

But, OpenAI clearly refers to this as an image and treats it as such. Including setting `image_detail`. Maybe this is an acceptable deviation between the two, as it more closely aligns with how OpenAI works?

src/pipecat/services/openai/realtime/llm.py
markbackman: I think I'll set to trace because it will add up to too much log storage space.

examples/foundational/19c-openai-realtime-live-video.py
markbackman: Agreed. I'm removing`start_audio_paused=False` from all demos. It doesn't need to be set as `start_audio_paused=False` is the default behavior.

examples/foundational/19c-openai-realtime-live-video.py
markbackman: We don't need to unpause anything. This was carried over from 26c. I'm removing since it's confusing. Do you know why it's there in 26c?

src/pipecat/services/openai/realtime/llm.py
markbackman: You mentioned:
> (If we change it to "video", then we might need a different name for image_detail, like video_frame_detail?)

Do you think it's better to be consistent within Pipecat or consistent with the OpenAI API? I think that would decide which way to go. I've opted to be more similar with OpenAI since it will be more helpful to developers. WDYT?

kompfner: > since it will be more helpful to developers

It depends. If we go one way, we'll add a bit of friction to developers who are familiar with OpenAI terminology. If we go the other way, it's a bit of friction for developers switching between services in Pipecat. I don't feel strongly. 

Might ever so slight preference for referring it to as video input simply because "pausing" image input makes less sense to me than pausing video input.

Your call.

examples/foundational/19c-openai-realtime-live-video.py
kompfner: > Do you know why it's there in 26c

Nope!

examples/foundational/19c-openai-realtime-live-video.py
```diff
@@ -131,8 +130,8 @@ async def run_bot(transport: BaseTransport, runner_args: RunnerArguments):
     async def on_client_connected(transport, client):
         logger.info(f"Client connected: {client}")
 
-        await maybe_capture_participant_camera(transport, client, framerate=1)
-        await maybe_capture_participant_screen(transport, client, framerate=1)
+        await maybe_capture_participant_camera(transport, client, framerate=0.5)
```
markbackman, line 133: Slowing down the input because OpenAI seems to struggle with a lot of images.

PR #3364, opened 2026-01-07
Add support for specifying sip provider.

rajneeshksoni:
optional "provider" field in the RoomSipParams


src/pipecat/transports/daily/utils.py
```diff
@@ -27,13 +27,15 @@ class DailyRoomSipParams(BaseModel):
         num_endpoints: Number of allowed SIP endpoints.
         codecs: Codecs to support for audio and video. If None, uses Daily defaults.
             Example: {"audio": ["OPUS"], "video": ["H264"]}
+        provider: (Optional) SIP provider name (daily).
```
aconchillo: What does `(daily)` mean? Is this an example of a possible value? If so, may be `(e.g. "daily")` or does it default to `daily` somehow?

Also, instead of `(Optional)` maybe `An optional SIP provider name (e.g. "daily")` or `An optional SIP provider name (defaults to "daily").`

aconchillo:
Thank you! LGTM!

PR #3365, opened 2026-01-07
Fix function calls staying IN_PROGRESS when result_callback not called

### LINKED ISSUE #3342: If a function call handler does not await its FunctionCallParams.result_callback, it will perpetually remain IN_PROGRESS in LLMAssistantAggregator

lukepayyapilli:
## Summary

Fixes #3342

When a function handler doesn't call `result_callback`, the function call stays `IN_PROGRESS` forever in `_function_calls_in_progress`. This causes:
- [x] Future function calls that return results won't trigger automatic LLM responses.
- [x] `has_function_calls_in_progress` perpetually returns `True`.

This fix auto-completes the function call with `None` if the handler returns without calling the callback.

Every existing usage follows the same pattern - immediate await. Even "fire and forget" cases explicitly call `await params.result_callback(None)`. No evidence of deferred callback patterns where the callback is stored and called later.

This fix aligns with the existing mental model: if you don't explicitly call the callback, the system treats it as `result_callback(None)`.

### DIFF
--- a/src/pipecat/services/llm_service.py
+++ b/src/pipecat/services/llm_service.py
@@ -596,9 +596,14 @@ class LLMService(AIService):
         )
 
         # Define a callback function that pushes a FunctionCallResultFrame upstream & downstream.
+        callback_invoked = False
+
         async def function_call_result_callback(
             result: Any, *, properties: Optional[FunctionCallResultProperties] = None
         ):
+            nonlocal callback_invoked
+            callback_invoked = True
+
             await self.broadcast_frame(
                 FunctionCallResultFrame,
                 function_name=runner_item.function_name,
@@ -643,6 +648,9 @@ class LLMService(AIService):
                     result_callback=function_call_result_callback,
                 )
                 await item.handler(params)
+
+        if not callback_invoked:
+            await function_call_result_callback(None)
 
     async def _cancel_function_call(self, function_name: Optional[str]):
         cancelled_tasks = set()


lukepayyapilli:
Closing as duplicate of #3343 which implements the same fix.

PR #3366, opened 2026-01-07
Allow language=None in CartesiaTTSService for auto-detection

### LINKED ISSUE #3353: CartesiaTTSService forces language="en" instead of allowing auto-detection
### pipecat version

0.0.98

### Service Name

CartesiaTTSService

### Service or model version

sonic-3

### Issue Description

I’m using CartesiaTTSService with Pipecat and noticed that when language=None is passed, the service automatically replaces it with "en".

However, according to Cartesia’s documentation, language is optional and when omitted, the TTS engine performs automatic language detection, which works correctly for multilingual input.

This behavior is inconsistent with:

Cartesia’s own API

```
from pipecat.services.cartesia.tts import CartesiaTTSService

CartesiaTTSService(
    api_key=settings.tts.cartesia.CARTESIA_API_KEY,
    model=model,
    voice_id=voice_id,
    params=CartesiaTTSService.InputParams(language=None),
)
```

### Reproduction Steps

1. Use CartesiaTTSService 
2. Can't speak multi-languages in the same conversation. 

### Expected Behavior

If language=None, do not send the language field at all (or allow it to be None)

Let Cartesia perform automatic language detection

This matches this documentation:

“Sonic-3 supports 27 languages with native pronunciation. While auto-detection works well for most cases, selecting a specific language improves accuracy for specialized vocabulary or accent requirements.”
— https://vapi.ai/blog/Cartesia-free-sonic-3-TTS

And the official WebSocket API reference:
https://docs.cartesia.ai/api-reference/tts/websocket

(language is an enum but not required)

Proof / Test

I verified this by subclassing CartesiaTTSService and explicitly sending "language": None:
```
class TestClass(CartesiaTTSService):
    def _build_msg(
        self, text: str = "", continue_transcript: bool = True, add_timestamps: bool = True
    ):
        voice_config = {
            "mode": "id",
            "id": self._voice_id,
        }

        msg = {
            "transcript": text,
            "continue": continue_transcript,
            "context_id": self._context_id,
            "model_id": self.model_name,
            "voice": voice_config,
            "language": None,
            "add_timestamps": add_timestamps,
            "use_original_timestamps": False if self.model_name == "sonic" else True,
        }

        if self._settings["speed"]:
            msg["speed"] = self._settings["speed"]

        return json.dumps(msg)
```

When speaking French and English in the same conversation, Cartesia correctly detected and synthesized both languages.

Proposed Fix

Change the initialization logic to allow None:

```
self._settings = {
    "output_format": {
        "container": container,
        "encoding": encoding,
        "sample_rate": 0,
    },
    "language": self.language_to_service_language(params.language)
    if params.language
    else None,
    "speed": params.speed,
    "emotion": params.emotion,
}
```

This would align CartesiaTTSService behavior with:

from pipecat.services.elevenlabs.tts import ElevenLabsTTSService

Where language=None is already supported and not replaced with "en".

### Actual Behavior

In CartesiaTTSService.__init__:
```
self._settings = {
    "output_format": {
        "container": container,
        "encoding": encoding,
        "sample_rate": 0,
    },
    "language": self.language_to_service_language(params.language)
    if params.language
    else "en",
    "speed": params.speed,
    "emotion": params.emotion,
}
```

If params.language is None, it is forced to "en", which disables Cartesia’s auto-detection.

### Error Logs

```shell

```

lukepayyapilli:
Fixes #3353

When `language=None` is passed to CartesiaTTSService, it was being replaced with `"en"`. This prevented Cartesia's built-in language auto-detection from working.

**Changes:**
1. Allow `None` as a valid language setting (instead of defaulting to `"en"`).
2. Conditionally include `language` in API payloads only when set (omit the field entirely when `None`, rather than sending `"language": null`).

This matches the pattern used for other optional fields like `speed` and `generation_config`, and aligns with ElevenLabsTTSService behavior.

### DIFF
--- a/src/pipecat/services/cartesia/tts.py
+++ b/src/pipecat/services/cartesia/tts.py
@@ -392,10 +392,10 @@ class CartesiaTTSService(AudioContextWordTTSService):
         Returns:
             List of (word, start_time) tuples processed for the language.
         """
-        current_language = self._settings.get("language", "en")
+        current_language = self._settings.get("language")
 
-        # Check if this is a CJK language
-        if self._is_cjk_language(current_language):
+        # Check if this is a CJK language (if language is None, treat as non-CJK)
+        if current_language and self._is_cjk_language(current_language):
             # For CJK languages, combine all characters in this message into one word
             # using the first character's start time
             if words and starts:


lukepayyapilli:
Thanks @markbackman! Updated `_process_word_timestamps_for_language` to handle None language as requested.

markbackman (approved):
Nice job! Thanks for the contribution 🙌 

Some housekeeping before this is ready:
- Can you add a changelog fragment?

PR #3367, opened 2026-01-07
Updated Smart Turn model weights to v3.2

marcus-daily:
Updating the weights to the latest version!

https://www.daily.co/blog/smart-turn-v3-2-handling-noisy-environments-and-short-responses/


markbackman:
Merging so we can test on `main`! 🎉

PR #3372, opened 2026-01-08
add new UserTurnController and UserTurnProcessor

aconchillo:

In this PR we introduce the following:

- `UserTurnController`: this is the class that now manages user turn strategies and emits events `on_user_turn_started`, `on_user_turn_stopped` and `on_user_turn_stop_timeout`.
- `LLMUserAggregator` now implements user turn logic using a `UserTurnController` (composition).
- `UserTurnProcesspr`: this is a processor that can be useful in advanced cases. It can be used as the pipeline user turn manager and then multiple user aggregators can be added in a parallel pipelines using external strategies (since `UserTurnProcessor` is the turn manager).

We also add a new example `53-concurrent-llm-evaluation.py` to show how to use the new `UserTurnProcessor`.

### DIFF
--- a/scripts/evals/run-release-evals.py
+++ b/scripts/evals/run-release-evals.py
@@ -239,6 +239,10 @@ TESTS_51 = [
     ("51-grok-realtime.py", EVAL_WEATHER),
 ]
 
+TESTS_53 = [
+    ("53-concurrent-llm-evaluation.py", EVAL_SIMPLE_MATH),
+]
+
 TESTS = [
     *TESTS_07,
     *TESTS_12,
@@ -254,6 +258,7 @@ TESTS = [
     *TESTS_49,
     *TESTS_50,
     *TESTS_51,
+    *TESTS_53,
 ]
 
 


src/pipecat/processors/aggregators/llm_response_universal.py
```diff
@@ -312,12 +323,6 @@ async def process_frame(self, frame: Frame, direction: FrameDirection):
         elif isinstance(frame, CancelFrame):
             await self._cancel(frame)
             await self.push_frame(frame, direction)
-        elif isinstance(frame, VADUserStartedSpeakingFrame):
-            await self._handle_vad_user_started_speaking(frame)
-            await self.push_frame(frame, direction)
```
markbackman, line 317: By moving this logic to the UserTurnController, it looks like the `VADUserStartedSpeakingFrame` (and stop frame) are no longer being pushed. Is that right?

markbackman: Ahh, no, these are pushed in the `else:` condition. Ok!

src/pipecat/processors/aggregators/llm_response_universal.py
```diff
@@ -238,11 +239,11 @@ class LLMUserAggregator(LLMContextAggregator):
     Example::
 
         @aggregator.event_handler("on_user_turn_started")
-        async def on_user_turn_started(aggregator, Optional[strategy]):
+        async def on_user_turn_started(aggregator, strategy: BaseUserTurnStartStrategy]):
```
markbackman, line 242: In the class docstring for `LLMUserAggregator` it could be useful to mention the relationship to the UserTurnController, similar to how you mention in UserTurnProcessor.

src/pipecat/processors/aggregators/llm_response_universal.py
aconchillo: Yep.

src/pipecat/processors/aggregators/llm_response_universal.py
aconchillo: Done! I realize there were still reference to bot start strategies, fixed that as well.

PR #3374, opened 2026-01-08
External turn controllers improvements

filipi87:
External turn controllers improvements:
- Allows defining whether `Deepgram` should send an interruption when the user’s turn has started.
- Allows defining whether `Flux` should send an interruption when the user’s turn has started.
- Allows defining whether `Speechmatics` should send an interruption when the user’s turn has started.

### DIFF
--- a/src/pipecat/services/speechmatics/stt.py
+++ b/src/pipecat/services/speechmatics/stt.py
@@ -299,7 +299,7 @@ class SpeechmaticsSTTService(STTService):
                 or defaults to `wss://eu2.rt.speechmatics.com/v2`.
             sample_rate: Optional audio sample rate in Hz.
             params: Optional[InputParams]: Input parameters for the service.
-            should_interrupt: Determine whether the bot should be interrupted when Speechmatics detects that the user is speaking.
+            should_interrupt: Determine whether the bot should be interrupted when Speechmatics turn_detection_mode is configured to detect user speech.
             **kwargs: Additional arguments passed to STTService.
         """
         super().__init__(sample_rate=sample_rate, **kwargs)


src/pipecat/services/deepgram/flux/stt.py
```diff
@@ -592,7 +595,8 @@ async def _handle_start_of_turn(self, transcript: str):
         logger.debug("User started speaking")
         self._user_is_speaking = True
         await self.broadcast_frame(UserStartedSpeakingFrame)
-        await self.push_interruption_task_frame_and_wait()
+        if self._should_interrupt:
+            await self.push_interruption_task_frame_and_wait()
```
kompfner, line 599: Is the thinking that we'll keep using this pattern (`push_interruption_task_frame_and_wait()`) since that's the general current code pattern until/if we definitely decide on the new way of triggering whole-pipeline interruptions?

Though...after @aconchillo's introduction of turn strategies, `LLMUserAggregator` broadcasts the interruption (with a TODO comment above it to revisit).

src/pipecat/services/speechmatics/stt.py
```diff
@@ -298,6 +299,7 @@ def __init__(
                 or defaults to `wss://eu2.rt.speechmatics.com/v2`.
             sample_rate: Optional audio sample rate in Hz.
             params: Optional[InputParams]: Input parameters for the service.
+            should_interrupt: Determine whether the bot should be interrupted when Speechmatics detects that the user is speaking.
```
kompfner: Maybe also mention that this only applies if Speechmatic VAD is enabled (which seems to be based on the `end_of_utterance_mode`), like you do with Deepgram

src/pipecat/services/deepgram/flux/stt.py
kompfner: Not sure which pattern we want to align on for now until we do a pass an make it consistent throughout.

aconchillo: For now, I also changed to `push_interruption_task_frame_and_wait()` until we have a better way.

markbackman (approved):
LGTM!

src/pipecat/services/speechmatics/stt.py
filipi87: Yep, make sense. I will improve this comment.

PR #3376, opened 2026-01-08
headers update

dhruvladia-sarvam:
New sdk_headers() helper that reports:
 X-SDK-Source: Pipecat
 X-SDK-Version: <pipecat-ai version>
 SDK-Language: Python
 sdk-language-version: <python version>

Sarvam STT (SarvamSTTService)
Best-effort injection of headers into sarvamai.AsyncSarvamAI across SDK versions (supports different internal header fields + different connect() kwarg names like headers / additional_headers / extra_headers).

Sarvam TTS
HTTP TTS (SarvamHttpTTSService) now sends the SDK headers on the /text-to-speech request.
WebSocket TTS (SarvamTTSService) now includes the SDK headers in the websocket handshake headers.

### DIFF
--- a/src/pipecat/services/sarvam/_sdk.py
+++ b/src/pipecat/services/sarvam/_sdk.py
@@ -4,8 +4,6 @@
 # SPDX-License-Identifier: BSD 2-Clause License
 #
 
-from __future__ import annotations
-
 import platform
 from typing import Dict
 


src/pipecat/services/sarvam/stt.py
```diff
@@ -125,7 +126,8 @@ def __init__(
 
         self.set_model_name(model)
         self._api_key = api_key
-        self._language_code = params.language
+        self._language_code: Optional[Language] = params.language
+        self._language_string: Optional[str]
```
markbackman: This is unnecessary since the assignment happens after. If you want to keep it, then you should explicitly assign it to `self._language_string: Optional[str] = None`.

src/pipecat/services/sarvam/tts.py
```diff
@@ -1,5 +1,5 @@
 #
-# Copyright (c) 2024-2026, Daily
+# Copyright (c) 2024–2025, Daily
```
```suggestion
# Copyright (c) 2024–2026, Daily
```

src/pipecat/services/sarvam/stt.py
```diff
@@ -394,6 +416,10 @@ async def _handle_message(self, message):
                     logger.debug("User started speaking")
                     await self._call_event_handler("on_speech_started")
 
+                elif signal == "END_SPEECH":
```
markbackman, line 418: When vad_signal are enabled, is the idea to use Sarvam as a VAD and not include a local VAD like Silero? If so, there are new changes in 0.0.99 that we should account for here. See the updates to Deepgram we made recently:
https://github.com/pipecat-ai/pipecat/pull/3314/files

We can take this up in a separate PR since it's out of scope here. Just something I noticed.

src/pipecat/services/sarvam/_sdk.py
```diff
@@ -0,0 +1,25 @@
+#
+# Copyright (c) 2024–2025, Daily
+#
+# SPDX-License-Identifier: BSD 2-Clause License
+#
+
+from __future__ import annotations
+
+import platform
+from importlib.metadata import version as lib_version
```
markbackman: We make the version available. Use this instead:
```
from pipecat import version as pipecat_version
```

src/pipecat/services/sarvam/_sdk.py
```diff
@@ -0,0 +1,25 @@
...
+
+
+def sdk_headers() -> dict[str, str]:
```
markbackman: By using the Pipecat import, you should be able to simplify this function to:
```
def sdk_headers() -> dict[str, str]:
    """SDK identification headers for upstream providers."""
    return {
        "X-SDK-Source": "Pipecat",
        "X-SDK-Version": pipecat_version(),
        "SDK-Language": "Python",
        "sdk-language-version": platform.python_version(),
    }
```

src/pipecat/services/sarvam/_sdk.py
```diff
@@ -0,0 +1,25 @@
+#
+# Copyright (c) 2024–2025, Daily
+#
+# SPDX-License-Identifier: BSD 2-Clause License
+#
+
+from __future__ import annotations
```
markbackman: I don't think this is needed. Pipecat supports 3.10+. Instead, we should use Dict from typing.

That would make this file:
```
#
# Copyright (c) 2024–2026, Daily
#
#

import platform
from typing import Dict

from pipecat import version as pipecat_version


def sdk_headers() -> Dict[str, str]:
    """SDK identification headers for upstream providers."""
    return {
        "X-SDK-Source": "Pipecat",
        "X-SDK-Version": pipecat_version(),
        "SDK-Language": "Python",
        "sdk-language-version": platform.python_version(),
    }
```

src/pipecat/services/sarvam/_sdk.py
```diff
@@ -0,0 +1,25 @@
+#
+# Copyright (c) 2024–2025, Daily
```
markbackman:
```suggestion
# Copyright (c) 2024–2026, Daily
```

src/pipecat/services/sarvam/_sdk.py
dhruvladia-sarvam: From what you are suggesting I am getting version as: 0.0.0.dev6918
From what I implemented, I am getting version as: 0.0.99.dev80

Which one would be correct here?

src/pipecat/services/sarvam/_sdk.py
```diff
@@ -0,0 +1,22 @@
+#
+# Copyright (c) 2024–2026, Daily
+#
+# SPDX-License-Identifier: BSD 2-Clause License
+#
+
+from __future__ import annotations
```
markbackman: This can be removed.
```suggestion
```

src/pipecat/services/sarvam/tts.py
```diff
@@ -1,5 +1,5 @@
 #
-# Copyright (c) 2024-2026, Daily
+# Copyright (c) 2024–2026, Daily
```
markbackman, line 2: Do you know what this change is? Maybe it's a different `-` char?

dhruvladia-sarvam: This change was by mistake, doesn't mean anything

src/pipecat/services/sarvam/stt.py
dhruvladia-sarvam: Noted, will look into this. Thanks Mark!

src/pipecat/services/sarvam/stt.py
dhruvladia-sarvam: Ah, got it. Done

src/pipecat/services/sarvam/_sdk.py
dhruvladia-sarvam: @markbackman

markbackman (approved):

PR #3377, opened 2026-01-08
Take into account VAD start_secs when passing audio data to Smart Turn

### LINKED ISSUE #3375: VAD_START_SECS causing audio provided to Smart Turn to be truncated
### pipecat version

Latest main (0.0.99.dev177)

### Python version

3.12.0

### Operating System

Ubuntu 24.04

### Issue description

I've been investigating why using Pipecat with Smart Turn produces less accurate results in some cases than just using Smart Turn on its own (e.g. with `record_and_predict.py`).

I've found that the beginning of each audio sample provided to Smart Turn is truncated -- for example, if I say "one two three four five", Pipecat might only send the audio for "three four five" to the model.

This can cause issues with short utterances where no speech data at all gets sent to the model, and so a random result gets returned.

After some more experimentation, I think this is caused by `VAD_START_SECS` -- increasing this value causes more of the data to be truncated.

Looking into a fix now.

### Reproduction steps

Run the quickstart example, and output to a file the data which gets passed to Smart Turn.

### Expected behavior

The start of each of the user's utterances is cut off.

### Actual behavior

The full turn context should be provided to Smart Turn

### Logs

```shell

```

marcus-daily:
Fixes #3375

### DIFF
--- a/src/pipecat/audio/turn/smart_turn/base_smart_turn.py
+++ b/src/pipecat/audio/turn/smart_turn/base_smart_turn.py
@@ -163,6 +163,7 @@ class BaseSmartTurn(BaseTurnAnalyzer):
         return state, result
 
     def on_vad_start_secs_updated(self, vad_start_secs: float):
+        """Store the new vad_start_secs value."""
         self._vad_start_secs = vad_start_secs
 
     def clear(self):


src/pipecat/turns/user_stop/turn_analyzer_user_turn_stop_strategy.py
```diff
@@ -102,6 +105,24 @@ async def _start(self, frame: StartFrame):
         self._turn_analyzer.set_sample_rate(frame.audio_in_sample_rate)
         await self.broadcast_frame(SpeechControlParamsFrame, turn_params=self._turn_analyzer.params)
 
+    async def _handle_speech_control_params(self, frame: SpeechControlParamsFrame):
+        """Sync Smart Turn pre-speech buffering with VAD start delay.
+
+        In the new user-turn-strategies pipeline, `VADUserStartedSpeakingFrame`
```
kompfner: So prior to the user turn strategies work, smart turn was getting that frame before VAD confirmed speech? I would've thought `VADUserStartedSpeakingFrame` was always emitted only once VAD confirmed speech...

kompfner (approved):

examples/quickstart/bot.py
```diff
@@ -79,7 +87,18 @@ async def run_bot(transport: BaseTransport, runner_args: RunnerArguments):
...
+        user_params=LLMUserAggregatorParams(
+            user_turn_strategies=UserTurnStrategies(
+                stop=[
+                    TurnAnalyzerUserTurnStopStrategy(
+                        turn_analyzer=LocalSmartTurnAnalyzerV3(params=SmartTurnParams())
+                    )
+                ],
+            )
+        ),
+    )
```
aconchillo: We can't update this just yet. We need to wait for a release, because quickstart is used in docker images that people are using.

src/pipecat/audio/turn/smart_turn/base_smart_turn.py
```diff
@@ -35,11 +35,15 @@ class SmartTurnParams(BaseTurnParams):
     Parameters:
         stop_secs: Maximum silence duration in seconds before ending turn.
         pre_speech_ms: Milliseconds of audio to include before speech starts.
+        vad_start_secs: Seconds VAD waits before confirming speech start (e.g. VAD STARTING window).
```
kompfner: Should we add a recommendation here that the developer *not* explicitly set this value in their application code, since it'll be automatically set under the hood to the right value at runtime?

markbackman: Perhaps we shouldn't expose `vad_start_secs` as a `SmartTurnParams`. Maybe it should only be settable via the VADParams? This might be preferred anyway as it would be set in one place only.

filipi87: > Perhaps we shouldn't exposevad_start_secs as a SmartTurnParams. Maybe it should only be settable via the VADParams? This might be preferred anyway as it would be set in one place only.

I think it makes sense. 

Inside the base input, when we change (or set the first time) the VAD parameters, we broadcast a `SpeechControlParamsFrame` containing the `VADParams` in use. 

Maybe we should handle this frame inside `TurnAnalyzerUserTurnStopStrategy` so we can pass the same VAD values used by the VAD to the turn analyzer ?

src/pipecat/turns/user_stop/turn_analyzer_user_turn_stop_strategy.py
```diff
@@ -102,6 +105,24 @@ async def _start(self, frame: StartFrame):
...
+        add it at inference slicing time (preserving `pre_speech_ms` semantics).
+        """
+        if not frame.vad_params:
+            return
+
+        params = self._turn_analyzer.params
+        if not isinstance(params, SmartTurnParams):
+            return
+
+        params.vad_start_secs = frame.vad_params.start_secs
```
filipi87: It looks like we are already doing this. 😅 So, feel free to ignore my previous comment.

src/pipecat/audio/turn/smart_turn/base_smart_turn.py
```diff
@@ -181,7 +185,8 @@ def _process_speech_segment(self, audio_buffer) -> Tuple[EndOfTurnState, Optiona
             return state, None
 
         # Extract recent audio segment for prediction
-        start_time = self._speech_start_time - (self._params.pre_speech_ms / 1000)
+        effective_pre_speech_ms = self._params.pre_speech_ms + (self._params.vad_start_secs * 1000)
+        start_time = self._speech_start_time - (effective_pre_speech_ms / 1000)
```
aconchillo, line 190: Should we leave this as it was and change this instead?

```
            if self._speech_start_time == 0:
                self._speech_start_time = time.time()
```
to

```
            if self._speech_start_time == 0:
                self._speech_start_time = time.time() - (self._params.vad_start_secs * 1000)
```

src/pipecat/turns/user_stop/turn_analyzer_user_turn_stop_strategy.py
marcus-daily: The comment is a bit misleading, this was also an issue before the strategies work.

src/pipecat/audio/turn/smart_turn/base_smart_turn.py
marcus-daily: Could do, I put it here though because it's where the existing `pre_speech_ms` calculation is done, and I think it makes sense to have all the adjustments in one place.

marcus-daily:
Thanks for the comments everyone! I've updated this to:

* Revert the quickstart to how it is on `main`
* Removed the new field from the turn analyzer params, and instead added a new `on_vad_start_secs_updated` method to the base class, which means the value doesn't have to be publicly exposed
* Made the comment in `_handle_speech_control_params` clearer

I've also updated the changelog.

There's a failure in `test_sieit` however this seems to be unrelated to this PR and intermittently fails elsewhere too (seems to be a timing-dependent test).

marcus-daily:
Thanks guys!

PR #3379, opened 2026-01-08
Fix FastAPIWebsocketTransport to handle both binary and text messages

### LINKED ISSUE #3347: FastAPIWebsocketInputTransport blocks text turns and crashes on voice due to strict JSON 'text' assumption
### pipecat version

0.0.98

### Python version

3.12.10

### Operating System

Windows 11 

## Use Case Description

I am building a browser-based conversational AI assistant using Pipecat with both
**text chat and voice chat** support.

The system architecture is:

- Frontend: Browser UI using WebSocket
  - Sends:
    - JSON `{ "text": "..." }` for text chat
    - Binary PCM audio frames for voice chat
- Backend: FastAPI + Pipecat
- STT: Deepgram
- LLM: Google Gemini
- TTS: Cartesia

The expected behavior is:

- When the user types text:
  - The text should create a user turn
  - The LLM should generate a response
  - The response should be returned as text (and optionally audio)

- When the user speaks:
  - Audio is streamed to Deepgram
  - After VAD + final transcription
  - The transcription should create a user turn
  - The LLM should generate a response
  - The response should be returned as text and audio

This use case matches a typical real-time browser voice assistant workflow.

[main.py](https://github.com/user-attachments/files/24422502/main.py)

### Current Approach

```python
## Current Approach

I am using `FastAPIWebsocketInputTransport` to handle a single WebSocket connection
that carries both:

- Binary audio frames (for voice input)
- JSON text messages (for text input)

Pipeline structure (simplified):
```

When using `FastAPIWebsocketInputTransport` with a browser WebSocket client:

- **Text input** causes the pipeline to stall after STT initialization
- **Voice input** crashes the pipeline with `KeyError('text')`

Both issues stem from the transport assuming *all JSON WebSocket messages contain a top-level `"text"` key*.

## Text Chat Behavior

- `{ "text": "Hello" }` reaches FastAPI correctly
- Pipeline initializes
- Deepgram STT connects
- **No LLM output occurs**
- Pipeline appears to stop after Deepgram

This happens because no valid user turn is created despite valid text input.

## Voice Chat Behavior

- Audio frames stream normally
- Deepgram connects successfully
- While streaming, transport raises:

```text
FastAPIWebsocketInputTransport exception receiving data: KeyError ('text')


### Additional Context

_No response_

### PR DESCRIPTION
**Pull request by lukepayyapilli**
## Summary
- [x] Create `_WebSocketMessageIterator` to handle both binary and text WebSocket messages.
- [x] Simplify `FastAPIWebsocketClient` by removing the `is_binary` configuration flag.
- [x] Route outgoing messages based on data type (`bytes` vs `str`) instead of configuration.
- [x] Add unit tests for the new iterator.
- [x] Remove `FrameSerializerType` enum and `type` property from all serializers (no longer needed).
- [x] Fix pre-existing `test_protobuf_serializer` test failure (compare attributes instead of frame objects).
- [x] Add `--extra websocket` to CI workflow for FastAPI test coverage.

## Approach

The previous implementation used Starlette's `iter_bytes()` or `iter_text()` based on serializer type, which meant messages of the other type were silently dropped.

The fix makes the transport format-agnostic:
- **Receive**: A new `_WebSocketMessageIterator` uses the low-level `receive()` method to yield both binary and text messages.
- **Send**: Routes based on actual data type (`isinstance(data, bytes)`) rather than configuration.
- **Serializer cleanup**: Since the transport no longer queries `serializer.type`, the `FrameSerializerType` enum and abstract `type` property are removed from all serializers.

## Alternatives Considered

1. **Add MIXED enum to FrameSerializerType**: Extend the enum with a MIXED type and update the transport to handle it. Trade-off: Adds complexity to the serializer contract and requires transport changes to switch on three types instead of two.

2. **Convert receive() to async generator**: Change from returning an iterator to an async generator function. Trade-off: Breaking change to the method signature that could affect downstream code.

3. **Add mixed_mode flag to client**: Add a boolean flag alongside `is_binary`. Trade-off: Proliferates configuration options and doesn't address the root issue that the transport shouldn't need to know about message formats.

The chosen approach is the simplest - it removes configuration entirely and lets the data type drive behavior, following the single responsibility principle.

Fixes #3347

### DIFF
--- a/.github/workflows/coverage.yaml
+++ b/.github/workflows/coverage.yaml
@@ -33,7 +33,7 @@ jobs:
 
       - name: Install dependencies
         run: |
-          uv sync --group dev --extra anthropic --extra aws --extra google --extra langchain
+          uv sync --group dev --extra anthropic --extra aws --extra google --extra langchain --extra websocket
 
       - name: Run tests with coverage
         run: |

### CODE REVIEW

**Comment by markbackman** [comment]
In this PR, I have a proposed fix for this:
https://github.com/pipecat-ai/pipecat/pull/3265

It hasn't been reviewed because we've been busy, but in it, I propose a MIXED mode. Commits:
- https://github.com/pipecat-ai/pipecat/pull/3265/commits/186f76db469b91ed980de41492d5cc0e9b5ce32f
- https://github.com/pipecat-ai/pipecat/pull/3265/commits/024625fc1522cb96ebfe21dd11543c2b98c1b7c9

I could extract this change if it would solve your issues. WDYT?

**Comment by lukepayyapilli** [comment]
Thanks for the pointer @markbackman! I took a look at your commits in #3265.

Our approaches differ in one key way:
- **#3265**: Adds opt-in MIXED mode via a new enum value, letting serializers explicitly declare mixed message support.
- This PR makes the transport format-agnostic by default, removing the `is_binary` configuration.

Both solve the underlying issue. The explicit MIXED mode gives more control, while the format-agnostic approach is simpler. Happy to defer to what the maintainers think is the better fit for the codebase as either solution works for me.

**Comment by markbackman** [comment]
@lukepayyapilli the base `FrameSerializer` class has:
```
    @property
    @abstractmethod
    def type(self) -> FrameSerializerType:
        """Get the serialization type supported by this serializer.

        Returns:
            The FrameSerializerType indicating binary or text format.
        """
        pass
```

Which requires that subclassed `FrameSerializers` declare their `FrameSerializerType`. This is why I opted to go with the explicit typing.

@aconchillo might have the strongest opinion on which direction to go. I have a slight preference for the explicit typing since that's a preference in the base class.

lukepayyapilli:
> @lukepayyapilli the base `FrameSerializer` class has:
> 
> ```
>     @property
>     @abstractmethod
>     def type(self) -> FrameSerializerType:
>         """Get the serialization type supported by this serializer.
> 
>         Returns:
>             The FrameSerializerType indicating binary or text format.
>         """
>         pass
> ```
> 
> Which requires that subclassed `FrameSerializers` declare their `FrameSerializerType`. This is why I opted to go with the explicit typing.

Ah that makes sense. Okay happy to close this one as needed just lmk. I'll wait for confirmation from @aconchillo - thanks!

aconchillo:
> > @lukepayyapilli the base `FrameSerializer` class has:
> > ```
> >     @property
> >     @abstractmethod
> >     def type(self) -> FrameSerializerType:
> >         """Get the serialization type supported by this serializer.
> > 
> >         Returns:
> >             The FrameSerializerType indicating binary or text format.
> >         """
> >         pass
> > ```
> >   
> > Which requires that subclassed `FrameSerializers` declare their `FrameSerializerType`. This is why I opted to go with the explicit typing.
> 
> Ah that makes sense. Okay happy to close this one as needed just lmk. I'll wait for confirmation from @aconchillo - thanks!

I like this PR approach better. We should actually get rid of `FrameSerializerType` and `def type(self)`. The only reason this was added was to use `iter_bytes()` or `iter_text()`. Since we can get that info from `receive()` data type, I don't think we need `FrameSerializerType` and therefore `type()` anymore.

lukepayyapilli:
Thanks @aconchillo! Updated to remove `FrameSerializerType` enum and the `type` property from all serializers.

If this expands the scope too much for this PR, I can revert the serializer changes and raise a separate issue to remove the unused enum/property.

markbackman:
@aconchillo I thought about this more just now on a run and had the same conclusion. With the MIXED type, we're using `websocket.receive()` to handle different types. This makes needing to declare the types unnecessary.

@lukepayyapilli so yes, I agree with your proposal. I think this will simplify things. I think we can work through this in a single PR. No worries.

aconchillo:
LGTM! @markbackman I'll leave the merge to you. Thank you @lukepayyapilli !

lukepayyapilli:
Thanks @aconchillo and @markbackman for the review and guidance!

Since the last update, I also made a few additional changes:
- Fixed a pre-existing test failure in `test_protobuf_serializer` (was comparing frame objects which have auto-incrementing IDs, now compares attributes).
- Added `--extra websocket` to the CI test and coverage workflows so the new FastAPI tests run with the required dependencies.

Ready for merge whenever you are.

PR #3386, opened 2026-01-09
Deprecate support for vad_events in DeepgramSTTService

markbackman:

The Deepgram team recommends using either a local VAD, like Silero, or Flux over the server-side vad_events. Due to performance issues, we're deprecated support.

@filipi87 this includes also deprecating the newly added `support_interruptions` feature. I don't love this sequence, but it works given the changes in 0.0.99. Any other ideas?

I'm also removing the 07c VAD variant from the release evals.


filipi87 (approved):
I think deprecating both makes sense, and we can probably remove this code in version 1.0.

LGTM!

markbackman:
> I think deprecating both makes sense, and we can probably remove this code in version 1.0.
> 
> LGTM!

Agreed. Given the performance of the local VAD, I can't see how anyone is using it since it issues so many false positives. Removing in 1.0 will work fine.

PR #3391, opened 2026-01-09
Krisp VIVA follow-up improvements

filipi87:
Krisp VIVA follow-up improvements:
- Added changelog entries
- Updated the Krisp Viva example to use Krisp turn model.
- Fixed `KrispVivaTurn` to properly release the Krisp SDK.

### DIFF
--- a/examples/foundational/07p-interruptible-krisp-viva.py
+++ b/examples/foundational/07p-interruptible-krisp-viva.py
@@ -99,11 +99,7 @@ async def run_bot(transport: BaseTransport, runner_args: RunnerArguments):
         context,
         user_params=LLMUserAggregatorParams(
             user_turn_strategies=UserTurnStrategies(
-                stop=[
-                    TurnAnalyzerUserTurnStopStrategy(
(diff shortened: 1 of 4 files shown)


examples/foundational/07p-interruptible-krisp-viva.py
```diff
@@ -82,7 +99,7 @@ async def run_bot(transport: BaseTransport, runner_args: RunnerArguments):
         context,
         user_params=LLMUserAggregatorParams(
             user_turn_strategies=UserTurnStrategies(
-                stop=[TurnAnalyzerUserTurnStopStrategy(turn_analyzer=LocalSmartTurnAnalyzerV3())]
+                stop=[TurnAnalyzerUserTurnStopStrategy(turn_analyzer=KrispVivaTurn(params=KrispTurnParams()))]
```
markbackman: Do you need to set `KrispTurnParams()`?

filipi87: No, we don't. Fixed.

PR #3392, opened 2026-01-09
Add reconnect logic to WebsocketService in the event of ConnectionClo…

markbackman:
…sedError

A user in Discord reported challenges with the current approach:
https://discord.com/channels/1239284677165056021/1458802211453796462

The post clearly communicates the challenges of just breaking with a warning. Instead, it seems that we should retry if the service is configured to retry. In the event that retrying works, then the service will recover. If not, it will emit an ErrorFrame, which developers can handle via `on_error` to switch services.

### DIFF
--- a/src/pipecat/services/websocket_service.py
+++ b/src/pipecat/services/websocket_service.py
@@ -121,6 +121,39 @@ class WebsocketService(ABC):
             else:
                 logger.error(f"{self} send failed; unable to reconnect")
 
+    async def _maybe_try_reconnect(
+        self,
+        error: Exception,
+        error_message: str,
+        report_error: Callable[[ErrorFrame], Awaitable[None]],
+    ) -> bool:
+        """Check if reconnection should be attempted and try if appropriate.
+
+        Args:
+            error: The exception that occurred.
+            error_message: Human-readable error message for logging.
+            report_error: Callback function to report connection errors.
+
+        Returns:
+            True if should continue the receive loop, False if should break.
+        """
+        # Don't reconnect if we're intentionally disconnecting
+        if self._disconnecting:
+            logger.warning(f"{self} error during disconnect: {error}")
+            return False
+
+        # Log the error
+        logger.warning(error_message)
+
+        # Try to reconnect if enabled
+        if self._reconnect_on_error:
+            success = await self._try_reconnect(report_error=report_error)
+            return success
+        else:
+            # Reconnection disabled
+            await report_error(ErrorFrame(error_message))
+            return False
+
     async def _receive_task_handler(self, report_error: Callable[[ErrorFrame], Awaitable[None]]):
         """Handle websocket message receiving with automatic retry logic.
 
@@ -139,33 +172,17 @@ class WebsocketService(ABC):
                 logger.debug(f"{self} connection closed normally: {e}")
                 break
             except ConnectionClosedError as e:
-                # Don't reconnect if we're intentionally disconnecting
-                if self._disconnecting:
-                    logger.warning(f"{self} connection closed with an error during disconnect: {e}")
-                    break
-
                 # Connection closed with error (e.g., no close frame received/sent)
                 # This often indicates network issues, server problems, or abrupt disconnection
                 message = f"{self} connection closed, but with an error: {e}"
-                logger.warning(message)
-
-                if self._reconnect_on_error:
-                    success = await self._try_reconnect(report_error=report_error)
-                    if not success:
-                        break
-                else:
-                    await report_error(ErrorFrame(message))
+                should_continue = await self._maybe_try_reconnect(e, message, report_error)
+                if not should_continue:
                     break
             except Exception as e:
+                # General error during message receiving
                 message = f"{self} error receiving messages: {e}"
-                logger.error(message)
-
-                if self._reconnect_on_error:
-                    success = await self._try_reconnect(report_error=report_error)
-                    if not success:
-                        break
-                else:
-                    await report_error(ErrorFrame(message))
+                should_continue = await self._maybe_try_reconnect(e, message, report_error)
+                if not should_continue:
                     break
 
     async def _connect(self):


aconchillo:
> …sedError
> 
> A user in Discord reported challenges with the current approach: https://discord.com/channels/1239284677165056021/1458802211453796462
> 
> The post clearly communicates the challenges of just breaking with a warning. Instead, it seems that we should retry if the service is configured to retry. In the event that retrying works, then the service will recover. If not, it will emit an ErrorFrame, which developers can handle via `on_error` to switch services.

Yeah... the issue (see #3185 ) was that in some cases disconnecting would cause these errors and then we would try to reconnect when we were really trying to disconnect. It seems we should know if we are trying to disconnect because we are done (so there's no need to reconnect) or if we just got an error and we want to reconnect.

markbackman:
@aconchillo that makes sense. We could instead track disconnecting to filter out the case where an actual disconnect is taking place. Make sense?

markbackman:
@aconchillo I just pushed a new commit that adds a `_disconnecting` status to the `WebsocketService`. With this, we can track whether the service class is disconnecting as a result of an `EndFrame` or `CancelFrame` or if there was an unexpected disconnect. WDYT?

src/pipecat/services/websocket_service.py
```diff
@@ -138,9 +139,23 @@ async def _receive_task_handler(self, report_error: Callable[[ErrorFrame], Await
                 logger.debug(f"{self} connection closed normally: {e}")
                 break
             except ConnectionClosedError as e:
-                # Error closure, don't retry
-                logger.warning(f"{self} connection closed, but with an error: {e}")
-                break
+                # Don't reconnect if we're intentionally disconnecting
+                if self._disconnecting:
+                    logger.debug(f"{self} connection closed during disconnect: {e}")
```
aconchillo: Maybe make it a warning (so it's easy to see) and mention there was an error, otherwise it would be easy to miss that something went wrong (service might need to fix something).

```
logger.warning(f"{self} connection closed with an error during disconnect: {e}")
```

aconchillo:
> @aconchillo I just pushed a new commit that adds a `_disconnecting` status to the `WebsocketService`. With this, we can track whether the service class is disconnecting as a result of an `EndFrame` or `CancelFrame` or if there was an unexpected disconnect. WDYT?

I think it's good but this is a breaking change for integrations since they will not call `super()._connect()` or `super()._disconnect()`. For example: https://github.com/murf-ai/pipecat-murf-tts/blob/main/src/pipecat_murf_tts/tts.py

markbackman:
> > @aconchillo I just pushed a new commit that adds a `_disconnecting` status to the `WebsocketService`. With this, we can track whether the service class is disconnecting as a result of an `EndFrame` or `CancelFrame` or if there was an unexpected disconnect. WDYT?
> 
> I think it's good but this is a breaking change for integrations since they will not call `super()._connect()` or `super()._disconnect()`. For example: https://github.com/murf-ai/pipecat-murf-tts/blob/main/src/pipecat_murf_tts/tts.py

Any other thoughts of how to work around this? If not, I can post PRs to the integration repos for this change. That might be the best course of action. WDYT?

src/pipecat/services/websocket_service.py
markbackman: Agreed. Just made this update.

markbackman:
> > @aconchillo I just pushed a new commit that adds a `_disconnecting` status to the `WebsocketService`. With this, we can track whether the service class is disconnecting as a result of an `EndFrame` or `CancelFrame` or if there was an unexpected disconnect. WDYT?
> 
> I think it's good but this is a breaking change for integrations since they will not call `super()._connect()` or `super()._disconnect()`. For example: https://github.com/murf-ai/pipecat-murf-tts/blob/main/src/pipecat_murf_tts/tts.py

Thinking about this more, I don't think this is a breaking change. They just don't get the benefit of managing the self._disconnecting` flag.

We could update Murf with super()._connect(), etc. and that would also be non-breaking because there's a `_connect()` defined as an abstract method in WebsocketService. So, I think this is a safe change to make. I wrote some basic tests to confirm this. If you agree, I'll post a PR to Murf, which is the only impacted Community Integration.

aconchillo:
> > > @aconchillo I just pushed a new commit that adds a `_disconnecting` status to the `WebsocketService`. With this, we can track whether the service class is disconnecting as a result of an `EndFrame` or `CancelFrame` or if there was an unexpected disconnect. WDYT?
> > 
> > 
> > I think it's good but this is a breaking change for integrations since they will not call `super()._connect()` or `super()._disconnect()`. For example: https://github.com/murf-ai/pipecat-murf-tts/blob/main/src/pipecat_murf_tts/tts.py
> 
> Thinking about this more, I don't think this is a breaking change. They just don't get the benefit of managing the self._disconnecting` flag.
> 
> We could update Murf with super()._connect(), etc. and that would also be non-breaking because there's a `_connect()` defined as an abstract method in WebsocketService. So, I think this is a safe change to make. I wrote some basic tests to confirm this. If you agree, I'll post a PR to Murf, which is the only impacted Community Integration.

Yes! I actually woke with this thought. It's not a breaking change, they simply would not get this benefit. I think provigin PR to other projects would work. That said, I'm not sure what would happened if someone wanted to use the external service with an older version of Pipecat, I think that would break. So the change could check the Python version and call `super()...` if Pipecat is >= 0.0.99.

markbackman:
> That said, I'm not sure what would happened if someone wanted to use the external service with an older version of Pipecat, I think that would break. So the change could check the Python version and call super()... if Pipecat is >= 0.0.99.

I checked this too.

The updated `MurfTTSService` class would have:

tts.py:
```
    async def _connect(self):
        super._connect()

        await self._connect_websocket()

        if self._websocket and not self._receive_task:
            self._receive_task = self.create_task(self._receive_task_handler(self._report_error))
```

For older Pipecat (< 0.0.99), `MurfTTSService` is an `AudioContextWordTTSService`, which inherits `WebsocketService` (through a number of layers). `WebsocketService` has:
```
    @abstractmethod
    async def _connect(self):
        """Connect to the service.

        Implement service-specific connection logic including websocket connection
        via _connect_websocket() and any additional setup required.
        """
        pass
```

So, the call to `super._connect()` would be satisfied by the `@abstractmethod`. It just doesn't do anything. I think we're good. I tested this path too.

markbackman:
Submitted the change to Murf:
https://github.com/murf-ai/pipecat-murf-tts/pull/4

src/pipecat/services/websocket_service.py
```diff
@@ -138,9 +139,23 @@ async def _receive_task_handler(self, report_error: Callable[[ErrorFrame], Await
...
+                message = f"{self} connection closed, but with an error: {e}"
+                logger.warning(message)
+
+                if self._reconnect_on_error:
+                    success = await self._try_reconnect(report_error=report_error)
+                    if not success:
+                        break
+                else:
+                    await report_error(ErrorFrame(message))
+                    break
```
aconchillo, line 180: Should we do the same if there's an Exception? It seems the logic is kind of the same, maybe we can extract it into a `_maybe_try_reconnect` method.

markbackman: Good idea. Just pushed that change, which makes the logic the same. I can't think of a case where they should be different. I left the exception handling separate in case we need to modify either path in the future.

Does this align to what you're thinking?

aconchillo: Great. One last thing.... I was thinking that we added `ConnectionClosedError` to avoid reconnection. Now that we have added it again (with reconnection) we might just go with the `Exception` block, since it's the same.. The only difference is the error message, so we might just go with one case.

aconchillo: Aside from that LGTM!

markbackman: We discussed keeping these separate for better logging. I'll do that. This can help developers discover what the root cause is.

PR #3393, opened 2026-01-09
fix: prevent duplicate on_pipeline_started events

### LINKED ISSUE #3341: Two on_pipeline_started event getting triggered on starting pipeline
### pipecat version

0.0.96

### Python version

3.11

### Operating System

macOS

### Issue description

Two on_pipeline_started events are coming on starting pipeline

`@task.event_handler("on_pipeline_started")
        async def on_pipeline_started(task, frame):
            try:
                if self.websocket.client_state.value == 1:
                    await self.websocket.send_json({
                        "type": "state",
                        "state": "ready"
                    })
                self.log.info("Pipeline started")
            except Exception as e:
                self.log.error(f"Failed to send pipeline ready state event, error: {e}")
        `

### Reproduction steps

`async def run(self):
        """Run the bot session"""
        
        # check for doro device or parent type
        is_welcome_msg, welcome_msg = welcome_msg_required(self.device_type, self.user_type, self.user_name, self.language)
        # Setup transport
        transport = self._setup_transport()
        
        # Create services
        services = ServiceManager.create_services(
            pipeline_config, self.stt_service, self.llm_service, self.tts_service,
            self.llm_model, self.metadata, self.log
        )
        
        # Setup context aggregator
        context_aggregator, system_prompt = self._setup_context_aggregator(services['llm'])
        
        # setup welcome canceller
        welcome_canceller = None
        send_welcome_after_timeout = None
        if pipeline_config.welcome_detection_enabled and not is_welcome_msg:
            async def send_welcome_after_timeout():
                try:
                    await asyncio.sleep(self.welcome_timeout_secs)
                    if not self.welcome_state.user_has_spoken and not self.welcome_state.welcome_sent:
                        self.log.info(f"silent for {self.welcome_timeout_secs}s, sending welcome message")
                        self.welcome_state.welcome_sent = True

                        if context_aggregator:
                            await task.queue_frames([
                                context_aggregator.user().get_context_frame()
                            ])
                    self.log.debug("Welcome timeout cancelled - user spoke first")

            welcome_canceller = FirstSpeechDetection(self.welcome_state, metadata=self.metadata)
        else:
            self.log.warning("Welcome detection disabled - no automatic greeting")
        
        
        # Create processors
        recorder = CallRecorder(self.call_id, self.user_id)
        processors = ProcessorManager.create_processors(
            pipeline_config, self.metadata, recorder, self.websocket, self.queue_service,
            self.call_id, self.user_id, self.stt_enable_vad, self.stt_enable_partials, 
            self.welcome_state, context_aggregator, system_prompt
        )
        
        # Create additional components
        user_transcript_recorder = CustomTranscriptProcessor()
        text_cleaner = TextCleanerProcessor(metadata=self.metadata)
        emotion_filler = EmotionFillerProcessor(
            sample_rate=16000, min_audio_duration=0.5, confidence_threshold=0.1,
            language=self.language.lower(), metadata=self.metadata,
            notifier=EventNotifier(), recorder=recorder
        )
        
        exit_intent_processor = maybe_create(
            context_aggregator is not None,
            lambda: ExitIntentProcessor(
                transport=transport, context_aggregator=context_aggregator,
                websocket=self.websocket, metadata=self.metadata
            )
        )
        self.exit_intent_processor = exit_intent_processor
        
        # Build pipeline
        pipeline_components = PipelineBuilder.build_component_chain(
            transport, services, processors, context_aggregator,
            user_transcript_recorder, text_cleaner, emotion_filler,
            exit_intent_processor, self.filler_mode
        )
        
        pipeline = PipelineBuilder.create_pipeline(pipeline_components)
        
        # Setup event handlers
        self._setup_event_handlers(transport, user_transcript_recorder, recorder, send_welcome_after_timeout, is_welcome_msg)
        
        # Create and run task
        task = PipelineTask(
            pipeline,
            params=PipelineParams(
                allow_interruptions=True,
                enable_metrics=True,
                interruption_strategies=[MinWordsInterruptionStrategy(min_words=1)],
            )
        )
        self.task = task

        @task.event_handler("on_pipeline_started")
        async def on_pipeline_started(task, frame):
            try:
                if self.websocket.client_state.value == 1:
                    await self.websocket.send_json({
                        "type": "state",
                        "state": "ready"
                    })
                self.log.info("Pipeline started")
            except Exception as e:
                self.log.error(f"Failed to send pipeline ready state event, error: {e}")
        
        start_frame = StartFrame(allow_interruptions=True)
        if is_welcome_msg:
            await task.queue_frames([start_frame, TTSSpeakFrame(welcome_msg)])
        else:
            await task.queue_frame(start_frame)
        
        runner = PipelineRunner()
        self.runner = runner
        try:
            await runner.run(task)
        finally:
            # Always attempt cleanup even if the pipeline exits unexpectedly.
            await self._finalize_session(recorder, reason="session_finished")`

### Expected behavior

Ideally only one should be created and one more question only one startFrame is passed right during the whole pipeline run session? Correct?

### Actual behavior

Getting two events instead of one. And In case if user starts speaking then the second one is passed after a while

### Logs

```shell

```

lukepayyapilli:
## Summary
- [x] Logs warning when duplicate `StartFrame` detected (without blocking the event)
- [x] Logs warning when duplicate `EndFrame` detected (without blocking the event)
- [x] Added tracking flags `_started` and `_ended` for duplicate detection
- [x] Added test to verify the warning is logged

Per maintainer feedback, this change only emits warnings rather than blocking duplicate events.

Fixes #3341

### DIFF
--- a/changelog/3393.fixed.md
+++ b/changelog/3393.fixed.md
@@ -1,0 +1,1 @@
+- Fixed duplicate `on_pipeline_started` events being fired when pipeline is started multiple times.
--- a/src/pipecat/pipeline/task.py
+++ b/src/pipecat/pipeline/task.py
@@ -160,9 +160,7 @@ class PipelineTask(BasePipelineTask):
     - on_frame_reached_upstream: Called when upstream frames reach the source
     - on_frame_reached_downstream: Called when downstream frames reach the sink
     - on_idle_timeout: Called when pipeline is idle beyond timeout threshold
-    - on_pipeline_started: Called once when the pipeline starts. The pipeline
-          automatically sends a StartFrame when ``run()`` is called, so you should
-          not queue a StartFrame manually.
+    - on_pipeline_started: Called when pipeline starts with StartFrame
     - on_pipeline_stopped: [deprecated] Called when pipeline stops with StopFrame
 
             .. deprecated:: 0.0.86
@@ -293,6 +291,7 @@ class PipelineTask(BasePipelineTask):
         self._finished = False
         self._cancelled = False
         self._started = False
+        self._ended = False
 
         # This task maneger will handle all the asyncio tasks created by this
         # PipelineTask and its frame processors.
@@ -517,11 +516,6 @@ class PipelineTask(BasePipelineTask):
 
         Args:
             frame: The frame to be processed.
-
-        Note:
-            Do not queue a StartFrame manually. The pipeline automatically sends
-            a StartFrame when ``run()`` is called. Queueing a StartFrame will
-            result in a warning being logged.
         """
         await self._push_queue.put(frame)
 
@@ -530,11 +524,6 @@ class PipelineTask(BasePipelineTask):
 
         Args:
             frames: An iterable or async iterable of frames to be processed.
-
-        Note:
-            Do not queue a StartFrame manually. The pipeline automatically sends
-            a StartFrame when ``run()`` is called. Queueing a StartFrame will
-            result in a warning being logged.
         """
         if isinstance(frames, AsyncIterable):
             async for frame in frames:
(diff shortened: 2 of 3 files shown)


kedar389:
Hi, I am not a reviewer. But i noticed that same problem happens for on_pipeline_finished event. It can be triggered multiple times if there are for example multiple End Frames emitted from processors. Do you think you could fix that as well?

lukepayyapilli:
Hi @kedar389 - I think the [original issue](https://github.com/pipecat-ai/pipecat/issues/3341) was closed by the individual who raised it. If you open a new issue I can add the same fix for the `on_pipeline_finished` event. Just comment here when its open - thanks!

markbackman:
@lukepayyapilli thanks for opening this PR. The only case I can think of that would cause multiple `on_pipeline_started` events to be emitted is the case where the application code pushes a separate `StartFrame`. Given this, I think the best thing to do is to emit a warning. (The firing of two events is a side effect, but shouldn't be prevented. A warning is an effective mechanism to alert the user of the issue.

Do you agree? If so, can you modify this PR to warn in the event of this situation?

lukepayyapilli:
Updated - now just emits warnings without blocking the events. Also added a warning for duplicate EndFrame per @kedar389's feedback.

markbackman:
I chatted with @aconchillo about this and he had some thoughts. Let's see what he has to say.

aconchillo:
> I chatted with @aconchillo about this and he had some thoughts. Let's see what he has to say.

Thank you @lukepayyapilli! I will take a look in detail after the release (we are trying to get it out ASAP!). But I was just thinking that it might be also useful to let the user send an `StartFrame()` which mean we should probably disable the automatic one (and avoid the warning). This doesn't mean we shouldn't have these warnings. But I just want to give it a little bit more thought before guiding you into any direction. Thank you again for your latest PRs!

lukepayyapilli:
@aconchillo Good luck with the release and that makes sense. Let me know your preferred solution and I can adapt this work to fit. Thanks!

lukepayyapilli:
@aconchillo @markbackman just bumping this so it doesn't get stale. Let me know what you'd like me to do - thanks!

PR #3397, opened 2026-01-10
PipelineTask: add external pipeline task setup files

aconchillo:

In this PR we replace `PIPECAT_OBSERVER_FILES` for `PIPECAT_SETUP_FILES`. This makes it clearer that this is not only about observers. It also doesn't force us to return a list of observers, they can dynamically be added with `task.add_observer()`.

Here's an example:

```
from pipecat_whisker import WhiskerObserver

from pipecat.pipeline.task import PipelineTask

async def setup_pipeline_task(task: PipelineTask):
    task.add_observer(WhiskerObserver(task.pipeline))
```

### DIFF
--- a/src/pipecat/pipeline/task.py
+++ b/src/pipecat/pipeline/task.py
@@ -654,6 +654,9 @@ class PipelineTask(BasePipelineTask):
 
     async def _setup(self, params: PipelineTaskParams):
         """Set up the pipeline task and all processors."""
+        # Do any additional pipeline task setup externally.
+        await self._load_setup_files()
+
         # Load additional observers.
         await self._load_observer_files()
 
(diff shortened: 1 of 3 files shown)


src/pipecat/pipeline/task.py
```diff
@@ -860,10 +863,51 @@ async def _idle_timeout_detected(self) -> bool:
...
+                logger.error(f"{self} error running external setup from {f}: {e}")
+
     async def _load_observer_files(self):
         """Dynamically load observers from files listed in PIPECAT_OBSERVER_FILES."""
-        observer_files = os.environ.get("PIPECAT_OBSERVER_FILES", "").split(":")
+        observer_files = [f for f in os.environ.get("PIPECAT_OBSERVER_FILES", "").split(":") if f]
         for f in observer_files:
+            import warnings
+
+            with warnings.catch_warnings():
```
markbackman, line 904: Deprecation is probably safest. I was thinking we can just remove. Perhaps we remove in 1.0.0.

aconchillo: Yeah, I don't think anyone is using that yet. But better be safe.

PR #3403, opened 2026-01-10
InworldTTSService: Add keepalive task

markbackman:

### DIFF
--- a/src/pipecat/services/inworld/tts.py
+++ b/src/pipecat/services/inworld/tts.py
@@ -13,12 +13,14 @@ Contains two TTS services:
 Inworld’s text-to-speech (TTS) models offer ultra-realistic, context-aware speech synthesis and precise voice cloning capabilities, enabling developers to build natural and engaging experiences with human-like speech quality at an accessible price point.
 """
 
+import asyncio
 import base64
 import json
 import uuid
 from typing import Any, AsyncGenerator, Dict, List, Optional, Tuple
 
 import aiohttp
+import websockets
 from loguru import logger
 from pydantic import BaseModel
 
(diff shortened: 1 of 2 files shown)


src/pipecat/services/inworld/tts.py
```diff
@@ -757,6 +777,27 @@ async def _receive_messages(self):
...
+            try:
+                if self._websocket and self._websocket.state is State.OPEN:
+                    if self._context_id:
+                        keepalive_message = {
+                            "send_text": {"text": ""},
+                            "contextId": self._context_id,
+                        }
+                        logger.trace(f"Sending keepalive for context {self._context_id}")
+                    else:
+                        keepalive_message = {"send_text": {"text": ""}}
```
markbackman, line 794: Sending a keepalive message without a context results in an error (code 5). I'm handling that error in the code above, but there's an open question if this is the right thing to do.

It should be an infrequent case to hit and likely benign. You can have a no context but the keepalive run when there's a user interruption but no transcript generated. This would have to happen when the task runs (every 60 sec), so it would likely be of no consequence.

Do you have a preference on what to do?

markbackman: cc @cshape

markbackman:
> LGTM!
> 
> As a future improvement, it would be nice to have a specific keep alive message to send, where sending the context ID is not even needed. However, this would require changes on the service provider side.

I agree. cc @cshape from Inworld, who may be the most interested in that feedback.

markbackman:
Keeping this implementation for now, so we can get it into the release. If @cshape has a different POV on the keepalive messages, we can open a separate PR.

cshape:
Thanks @filipi87 and @markbackman -

I'll see internally if we can have a more elegant keepalive (perhaps without requiring context id) and will submit a PR to update our service when that's available.

For now this setup should be fine.

PR #3404, opened 2026-01-11
feat(gladia): add VAD events support

### LINKED ISSUE #3405: Gladia STT is missing some utterances
### pipecat version

0.0.93

### Python version

3.12

### Operating System

macOS

### Issue description

Gladia STT service uses an internal VAD analyzer that is used for setting utterance boundaries.
In the documentation, it is used with Pipecat's Silero VAD analyzer implementation.
This creates a redundant layer which truncates silence audio frames before they arrive at Gladia.
In turn, this prevents Gladia from detecting the end of the User Speak segment.

### Reproduction steps

1. Run a pipeline with both Gladia and Silero
2. Shortening silero stop_secs parameter, or playing with Gladia endpointing parameter result in unexpected behaviors.

### Expected behavior

Gladia STT service should produce transcriptions naturally as the user stopped speaking.

### Actual behavior

Gladia STT service will swallow short utterances and sometimes batch multiple user speak segments after a long delay.

### Logs

```shell

```

poseneror:
Add support for Gladia's speech_start/speech_end events to emit UserStartedSpeakingFrame and UserStoppedSpeakingFrame frames.

When enable_vad=True in GladiaInputParams:
- speech_start triggers interruption and pushes UserStartedSpeakingFrame
- speech_end pushes UserStoppedSpeakingFrame
- Tracks speaking state to prevent duplicate events

This allows using Gladia's built-in VAD instead of a separate VAD in the pipeline.

Fixes #3405

### DIFF
--- a/src/pipecat/services/gladia/stt.py
+++ b/src/pipecat/services/gladia/stt.py
@@ -204,6 +204,7 @@ class GladiaSTTService(WebsocketSTTService):
         model: str = "solaria-1",
         params: Optional[GladiaInputParams] = None,
         max_buffer_size: int = 1024 * 1024 * 20,  # 20MB default buffer
+        should_interrupt: bool = True,
         **kwargs,
     ):
         """Initialize the Gladia STT service.
@@ -222,6 +223,8 @@ class GladiaSTTService(WebsocketSTTService):
             model: Model to use for transcription. Defaults to "solaria-1".
             params: Additional configuration parameters for Gladia service.
             max_buffer_size: Maximum size of audio buffer in bytes. Defaults to 20MB.
+            should_interrupt: Determine whether the bot should be interrupted when
+                Gladia VAD detects user speech. Defaults to True.
             **kwargs: Additional arguments passed to the STTService parent class.
         """
         super().__init__(sample_rate=sample_rate, **kwargs)
@@ -270,6 +273,7 @@ class GladiaSTTService(WebsocketSTTService):
 
         # VAD state tracking
         self._is_speaking = False
+        self._should_interrupt = should_interrupt
 
     def __str__(self):
         return f"{self.name} [{self._session_id}]"
@@ -515,25 +519,28 @@ class GladiaSTTService(WebsocketSTTService):
     async def _on_speech_started(self):
         """Handle speech start event from Gladia.
 
-        Triggers interruption and emits UserStartedSpeakingFrame when VAD is enabled.
+        Broadcasts UserStartedSpeakingFrame and optionally triggers interruption
+        when VAD is enabled.
         """
         if not self._params.enable_vad or self._is_speaking:
             return
+
         logger.debug(f"{self} User started speaking")
         self._is_speaking = True
-        # Push interruption first to stop the bot, then notify about user speaking
-        await self.push_interruption_task_frame_and_wait()
-        await self.push_frame(UserStartedSpeakingFrame())
+
+        await self.broadcast_frame(UserStartedSpeakingFrame)
+        if self._should_interrupt:
+            await self.push_interruption_task_frame_and_wait()
 
     async def _on_speech_ended(self):
         """Handle speech end event from Gladia.
 
-        Emits UserStoppedSpeakingFrame when VAD is enabled.
+        Broadcasts UserStoppedSpeakingFrame when VAD is enabled.
         """
         if not self._params.enable_vad or not self._is_speaking:
             return
         self._is_speaking = False
-        await self.push_frame(UserStoppedSpeakingFrame())
+        await self.broadcast_frame(UserStoppedSpeakingFrame)
         logger.debug(f"{self} User stopped speaking")
 
     async def _send_audio(self, audio: bytes):


src/pipecat/services/gladia/stt.py
```diff
@@ -523,6 +528,30 @@ async def _handle_transcription(
...
+
+        Triggers interruption and emits UserStartedSpeakingFrame when VAD is enabled.
+        """
+        if not self._params.enable_vad or self._is_speaking:
+            return
+        logger.debug(f"{self} User started speaking")
+        self._is_speaking = True
+        # Push interruption first to stop the bot, then notify about user speaking
+        await self.push_interruption_task_frame_and_wait()
+        await self.push_frame(UserStartedSpeakingFrame())
```
markbackman: Per the comment above, this should be:
```
    async def _on_speech_started(self):
        """Handle speech start event from Gladia.

        Triggers interruption and emits UserStartedSpeakingFrame when VAD is enabled.
        """
        if not self._params.enable_vad or self._is_speaking:
            return

        logger.debug(f"{self} User started speaking")
        self._is_speaking = True
        
        if self._should_interrupt:
```

src/pipecat/services/gladia/stt.py
```diff
@@ -278,6 +280,9 @@ def __init__(
         self._connection_task = None
         self._should_reconnect = True
 
+        # VAD state tracking
```
markbackman, line 274: We've made changes in 0.0.99 that improve the user's speech input handling. In short, the user turn is broken into two phases, user started speaking and user stopped speaking. The decision making of these phases is done by a turn controller, which is a new concept. In this case, when Gladia acts as the turn controller, it needs to do a few things:

1. Add a new `__init__` arg called `should_interrupt:
```
should_interrupt: bool = True,
```

Docstring:
```
should_interrupt: Determine whether the bot should be interrupted when Speechmatics turn_detection_mode is configured to detect user speech.
```

Then initialize as:
```
self._should_interrupt = should_interrupt
```

2. During the user started speaking phase (in this order):
  - Broadcast `UserStartedSpeakingFrame` (which means push up and downstream)
  - push_interruption_task_frame_and_wait()

3. During the user stopped speaking phase:
  - Broadcast `UserStoppedSpeakingFrame`

src/pipecat/services/gladia/stt.py
```diff
@@ -523,6 +528,30 @@ async def _handle_transcription(
...
+
+    async def _on_speech_ended(self):
+        """Handle speech end event from Gladia.
+
+        Emits UserStoppedSpeakingFrame when VAD is enabled.
+        """
+        if not self._params.enable_vad or not self._is_speaking:
+            return
+        self._is_speaking = False
+        await self.push_frame(UserStoppedSpeakingFrame())
```
markbackman: Need to broadcast this frame.

```suggestion
        await self.broadcast_frame(UserStoppedSpeakingFrame())
```

src/pipecat/services/gladia/stt.py
markbackman: Note: in 0.0.99, the user context aggregator defaults to the turn controller. You can cede control to an external turn controller, by setting the context aggregator as follows:
```
        context,
        user_params=LLMUserAggregatorParams(user_turn_strategies=ExternalUserTurnStrategies()),
    )
```

markbackman (commented):
Thanks for the proposal! We're in the midst of a big change regarding speech input. I've included comments you need to account for. You can see DeepgramSTTService for how this all comes together:
https://github.com/pipecat-ai/pipecat/blob/main/src/pipecat/services/deepgram/stt.py

Note: for Deepgram, their external VAD is unreliable, so we're deprecating it. In general, relying on an external VAD is inferior to a local VAD, so we don't recommend it. But, we don't want to discourage use, if that's better suited for your use case. Anyway, just sharing context.

gui217:
Big thanks @markbackman for the quick review here!

Can you please elaborate more on why the external VADs are below expectations vs the internal ones? Also, are you comparing to Silero or some other internal VAD?

markbackman:
> Big thanks @markbackman for the quick review here!
> 
> Can you please elaborate more on why the external VADs are below expectations vs the internal ones? Also, are you comparing to Silero or some other internal VAD?

I'm comparing remote VADs to Silero running locally. The issue with remote VADs is that they take longer to provide a result. The main driver is the network roundtrip latency. Running Silero locally is almost always the best outcome; the compute required to run a local VAD is essentially insignificant (see hosting costs: https://docs.google.com/spreadsheets/d/1-B3nv7fhwEoFmDs-phm280XK9phTep0qfyLKRNsEwVE/edit?gid=0#gid=0)

poseneror:
I just want to mention that I've discussed this with the Gladia team, and they indeed use Silero internally.
Currently their API doesn't provide a way to disable it, so we'll have to use it for now.
With that being said, I feel like some services might provide better VAD then what you can use locally (specifically this might not be the case), so allowing that seems right.

src/pipecat/services/gladia/stt.py
poseneror: Version 0.0.99 sounds really exciting! These feel like great conceptual changes 👍

poseneror: I've added the should_interrupt bool and updated the user events to be broadcasted.

One question I have:
It's hard for me to figure out when a parameter should be in the function constructor vs when it should be in the internal params argument (Gladia's "params" arg).
Are there any guidelines here? - I've placed enable_vad in the "params" argument, and the should_interrupt as an argument of it's own (as seen in deepgrams implementation)

markbackman:
> Currently their API doesn't provide a way to disable it, so we'll have to use it for now.

Interesting—if they're both using Silero, performance can be identical. Why not run it locally and say ~150-200ms of latency?

Also, one additional consideration: with Pipecat's new user turn strategies (releasing in 0.0.99), a user turn is broken into two parts: user started speaking and user stopped speaking. You can assign strategies for how to handle both. With this, you're ceding control to Gladia, which prevents you tailoring the control for how interruptions and turn completion work.

For example, you're foregoing use of a smart-turn model to determine end of turn. I fully expect all STT providers to add this support over the next 12 months, so perhaps this is a short lived gap. Just something worth calling out so you (and another others finding this PR) are aware.

> Currently their API doesn't provide a way to disable it, so we'll have to use it for now.

True, but Pipecat can just ignore the VAD events.

Anyway, I'm not opposed to this change. I'm happy to leave the choice to developers.

src/pipecat/services/gladia/stt.py
markbackman: Args in `__init__` are usually required for initialization. The convention for `Params`/`InputParams` is to leave this to optional configuration items for the service. `enable_vad` belongs in `Params`.

`should_interrupt` is an initialization time arg that tells the service how to operate and won't be changed at runtime.

Generally, I agree with the confusion. This is on the to do list to improve.

PR #3406, opened 2026-01-11
fix(openrouter): handle multiple system messages for Gemini models

### LINKED ISSUE #3362: OpenRouter: Multiple system messages cause issues with Gemini models
### pipecat version

0.0.98

### Python version

3.11.8

### Operating System

Ubuntu 24.04

## Problem
When using pipecat-flows with Gemini models via OpenRouter, multiple system 
prompts are injected at different points in the conversation. Gemini doesn't 
support multiple system messages inline, only the last one "wins".

This causes the role prompt to be overwritten by task prompts.

Related: [pipecat-ai/pipecat-flows#171
](https://github.com/pipecat-ai/pipecat-flows/pull/171)

## Proposed Solution
Add an `OpenRouterLLMAdapter` that converts mid-conversation system messages 
to user messages, keeping only the first system message as an actual system 
message. This matches the behavior of the native Gemini adapter.

### Reproduction steps

1. Use OpenRouterLLMService with a Gemini model (e.g., "google/gemini-2.5-flash")
2. Use pipecat-flows with a node configuration that has both role_messages and task_messages:

return NodeConfig(
        name="initial",
        role_messages=[
            {
                "role": "system",
                "content": "some role message.",
            }
        ],
        task_messages=[
            {
                "role": "system",
                "content": "For this step, ask the user if they want pizza or sushi, and wait for them to use a function to choose. Start off by greeting them. Be friendly and casual; you're taking an order for food over the phone.",
            }
        ],
        pre_actions=[
            {
                "type": "function",
                "handler": check_kitchen_status,
            },
        ],
        functions=[choose_pizza, choose_sushi],
    )

3. Navigate through multiple nodes (each injecting new task_messages as system messages)
4. Observe the messages sent to the LLM

### Expected behavior

Multiple system messages should be handled gracefully. The first system message 
(role_messages) should remain as the system instruction, and subsequent system 
messages (task_messages) should be converted to user messages to preserve context.

Example expected output to API:
[
  {"role": "system", "content": "You are a friendly insurance agent..."},
  {"role": "assistant", "content": "Hello!"},
  {"role": "user", "content": "I need insurance"},
  {"role": "user", "content": "Ask for the customer's age."},  ← converted from system
]

### Actual behavior

Multiple system messages are sent directly to OpenRouter/Gemini, which doesn't 
handle them correctly. Gemini only supports a single system instruction, so 
subsequent system messages either:
- Overwrite the original role_messages
- Cause unexpected behavior in the conversation

Example actual output to API:
[
  {"role": "system", "content": "You are a friendly insurance agent..."},
  {"role": "assistant", "content": "Hello!"},
  {"role": "user", "content": "I need insurance"},
  {"role": "system", "content": "Ask for the customer's age."},  ← breaks Gemini
]

### Logs

```shell

```

omChauhanDev:

Fixes #3362 

### Problem
Gemini models strictly support a single system instruction. When `pipecat` sends multiple system messages (e.g., Role + Task), Gemini often overwrites the initial role with the last task, causing the bot to lose its persona. 

### Solution
Overrides `build_chat_completion_params` in `OpenRouterLLMService` to:

if the model is a Gemini variant.
-    Keep the first system message as `system`.
-    Convert all subsequent system messages to `user` messages.

### DIFF
--- a/src/pipecat/services/openrouter/llm.py
+++ b/src/pipecat/services/openrouter/llm.py
@@ -10,7 +10,7 @@ This module provides an OpenAI-compatible interface for interacting with OpenRou
 extending the base OpenAI LLM service functionality.
 """
 
-from typing import Optional
+from typing import Any, Dict, Optional
 
 from loguru import logger
 
@@ -61,3 +61,35 @@ class OpenRouterLLMService(OpenAILLMService):
         """
         logger.debug(f"Creating OpenRouter client with api {base_url}")
         return super().create_client(api_key, base_url, **kwargs)
+
+    def build_chat_completion_params(self, params_from_context: Dict[str, Any]) -> Dict[str, Any]:
+        """Builds chat parameters, handling model-specific constraints.
+
+        Args:
+            params_from_context: Parameters from the LLM context.
+
+        Returns:
+            Transformed parameters ready for the API call.
+        """
+        params = super().build_chat_completion_params(params_from_context)
+        model = getattr(self, "model_name", getattr(self, "model", "")).lower()
+        if "gemini" in model:
+            messages = params.get("messages", [])
+            if not messages:
+                return params
+            transformed_messages = []
+            system_message_seen = False
+            for msg in messages:
+                if msg.get("role") == "system":
+                    if not system_message_seen:
+                        transformed_messages.append(msg)
+                        system_message_seen = True
+                    else:
+                        new_msg = msg.copy()
+                        new_msg["role"] = "user"
+                        transformed_messages.append(new_msg)
+                else:
+                    transformed_messages.append(msg)
+            params["messages"] = transformed_messages
+
+        return params


src/pipecat/services/openrouter/llm.py
```diff
@@ -61,3 +61,35 @@ def create_client(self, api_key=None, base_url=None, **kwargs):
         """
         logger.debug(f"Creating OpenRouter client with api {base_url}")
         return super().create_client(api_key, base_url, **kwargs)
+
+    def build_chat_completion_params(self, params_from_context: Dict[str, Any]) -> Dict[str, Any]:
```
kompfner, line 65: This logic seems reasonable!

I think it might be cleaner if it could be moved into an `OpenRouterLLMAdapter` instead, which could just inherit from `OpenAILLMAdapter`. Typically we prefer to keep the work of "translating" standard context messages into LLM-specific messages, adhering to LLM-specific requirements, inside adapter classes.

This is also the approach recommended in the the issue, #3362.

But...looking at going that route, it might not be straightforward. You'd want to override `OpenAILLMAdapter.get_llm_invocation_params()`...but that method never gets any information about which model is being used (because it doesn't matter at all to the base class, `OpenAILLMAdapter`). And I'm not sure it's worth updating the base `OpenAILLMAdapter`, which is used everywhere, just for this.

So, I think the approach you took is reasonable.

kompfner (approved):
Thanks for the fix!

omChauhanDev:
@kompfner Is this good to merge, or adding a `changelog` fragment is recommended for such a fix?

kompfner:
Ah right, pardon me, forgot to hit merge. Will do so. Happy to stick in a changelog fragment for this before doing so.

kompfner:
> Ah right, pardon me, forgot to hit merge. Will do so. Happy to stick in a changelog fragment for this before doing so.

Oops, will add it in a follow-up PR :)

kompfner:
> > Ah right, pardon me, forgot to hit merge. Will do so. Happy to stick in a changelog fragment for this before doing so.
> 
> Oops, will add it in a follow-up PR :)

https://github.com/pipecat-ai/pipecat/pull/3546

yuki901:
Thanks for taking care of this fix and pushing it through!

PR #3411, opened 2026-01-12
Adding DailyParams to the 41b example.

filipi87:
Adding DailyParams to the 41b example, so I could use it to test with RN, this issue:
- https://github.com/pipecat-ai/pipecat-client-react-native-transports/issues/17


filipi87:
Makes sense. I was only using this example because it was the only foundational example where we were sending `RTVIServerMessageFrame`, but I can add support for this in the other examples.

PR #3415, opened 2026-01-12
Rtvi send file support

mattieruth:
This PR introduces first-class file support across Pipecat's frame, context, adapter, and RTVI layers, allowing clients to send files (PDFs, images, and other documents) to the bot and have them incorporated into LLM context.

#### Core abstractions

- New `FileRawFrame` / `InputFileRawFrame` / `UserFileRawFrame` frame hierarchy (parallel to the existing image frame hierarchy), carrying a file payload as either bytes or a URL, a MIME type, and a FileSourceType ("bytes", "url", "id").
- `ImageRawFrame.image` widened from bytes to bytes | str to accommodate base64 strings arriving from RTVI clients.

#### Universal context messages

Two new universal message content types flow through LLMContext and get translated per-adapter:
  - `file` — file as a base64 data URL (data:<mime>;base64,...), stored in file.file_data
  - `file_url` — file as a plain URL, stored in file.url

New LLMContext methods: `create_file_message()`, `create_file_url_message()`, `add_file_frame_message()`.

#### Log redaction

`LLMContext._truncate_large_values_from_messages` now redacts `file.file_data` in file-type content items, keeping filename and mime_type visible. The Bedrock adapter's get_messages_for_logging also redacts `document.source.bytes`.

#### Aggregator

`LLMAssistantAggregator` now handles `UserFileRawFrame`, appending it to context and triggering a context push the same way images do.

#### Adapter translations

| Adapter | file | file_url |
| :--- | :--- | :--- |
| Anthropic  | document w/ base64 source (PDF only)       | document or image w/ URL source (PDF/image) |
| AWS Bedrock   | document w/ raw bytes source (PDF/CSV/DOC/DOCX/XLS/XLSX/HTML/TXT/MD) | document w/ S3 URI (S3 URLs only)           |
| Gemini      | Part(inline_data=Blob(...))                               | not supported (warning logged)    |
| OpenAI Chat      | file w/ file_data (PDF only)                     | not supported (warning logged)    |
| OpenAI Responses | input_file w/ file_data                        | input_file w/ file_url                         |

Unsupported MIME types or source combinations log a warning and skip the item rather than crashing.

### Error recovery

`LLMContext.maybe_remove_invalid_message()` removes the most recent file-containing message when OpenAI returns an `invalid_request_error` with param: "file_id", preventing the context from getting permanently stuck. Called from `BaseOpenAILLMService` on any LLM exception.

#### RTVI send-file message

New RTVI message type send-file with models `SendFileData`, `File`, `FileBytes`, `FileUrl`, `FileId`, `SendFileOptions`. Three source variants are resolved in `RTVIProcessor._handle_send_file`:
  - `FileBytes` — base64 string passed directly from the client
  - `FileUrl` public — URL forwarded as-is
  - `FileId (pipecat:<uuid>)` — read from the uploads folder on disk, deleted after reading

Image files are dispatched as `UserImageRawFrame`; all others as `UserFileRawFrame`. `SendFileOptions` controls `run_immediately` (triggers a bot interruption), `audio_response` (toggles TTS skip), and `custom_options` (provider-specific hints like OpenAI detail or Bedrock citations). 

#### Development runner

  - New POST `/files` endpoint saves uploads under UUID names (not original filenames) and returns {"source": {"type": "id", "id": "pipecat:<uuid>"}, "format": "<mime>", "name": "<original>"}.
  - New -u/--uploads-folder CLI flag enables the endpoint; without it the endpoint is not mounted.
  - New --uploads-folder-max-files (default 10) trims oldest files when the cap is exceeded.
  - -f/--folder renamed to --downloads-folder; --folder kept as a deprecated alias.
  - New runner_uploads_folder() accessor for use in RTVIProcessor(uploads_folder=...).

#### Dependency

Bumped aiohttp minimum from 3.11.12 to 3.14.0 — required by google-genai ≥ 2.8.0, which calls StreamReader.readline(max_line_length=...), a parameter added in aiohttp 3.14.

#### Testing

Check out the new example here: https://github.com/pipecat-ai/pipecat-examples/pull/237

TODOS:
- the server-specific configs are currently dropped. they need to be added to the universal message to be handled by the adapter


src/pipecat/frames/frames.py
```diff
@@ -244,6 +244,26 @@ class ImageRawFrame:
     format: Optional[str]
 
 
+FileSourceType = Literal["bytes", "url", "id"]
+
+
+@dataclass
+class FileRawFrame:
```
mattieruth, line 199: these frames follow the pattern set by the image frames

src/pipecat/processors/aggregators/llm_context.py
```diff
@@ -173,6 +174,11 @@ async def create_image_message(
             image: Raw image bytes.
             text: Optional text to include with the image.
         """
+        # Format is a data URL: data:<mime type>;base64,<data> already provided
+        if format.startswith("url/"):
```
mattieruth: i wrote this, sort of following the pattern of the 'image/' check below to bypass the encoded logic since the data's good to go, but it feels hacky. suggestions welcome.  (the 'url/' is prepended down in rtvi/processory handling.)

src/pipecat/processors/frameworks/rtvi/processor.py
```diff
@@ -555,6 +569,91 @@ async def _handle_send_text(self, data: RTVI.SendTextData):
             output_frame = LLMConfigureOutputFrame(skip_tts=cur_llm_skip_tts)
             await self.push_frame(output_frame)
 
+    async def _handle_send_file(self, data: RTVI.SendFileData):
```
mattieruth: this is probably the best place to start this review :)

src/pipecat/processors/frameworks/rtvi/processor.py
```diff
@@ -555,6 +569,91 @@ async def _handle_send_text(self, data: RTVI.SendTextData):
...
+                    return
+                # read bytes from file system, encode to base64, then delete the file
+                type = "bytes"
+                file_path = os.path.join(self._folder, file.source.id.removeprefix("pipecat:"))
+                with open(file_path, "rb") as f:
+                    raw_bytes = f.read()
+                    encoded_file = base64.b64encode(raw_bytes).decode("utf-8")
+                    source = f"data:{file.format};base64,{encoded_file}"
+                try:
+                    os.remove(file_path)
```
mattieruth: i'm torn on whether this handling belongs here or in the llm frame handling

src/pipecat/runner/run.py
```diff
@@ -230,18 +262,30 @@ async def root_redirect():
...
     @app.get("/files/{filename:path}")
     async def download_file(filename: str):
-        """Handle file downloads."""
-        if not args.folder:
-            logger.warning(f"Attempting to dowload {filename}, but downloads folder not setup.")
-            return
-
-        file_path = Path(args.folder) / filename
-        if not os.path.exists(file_path):
+        """Serve a file from uploads or downloads folder (uploads checked first)."""
```
mattieruth: i think all this was done by Claude to support getting files from either folder. Happy to undo it.

src/pipecat/services/openai/base_llm.py
```diff
@@ -568,6 +568,7 @@ async def process_frame(self, frame: Frame, direction: FrameDirection):
                 await self.push_error(error_msg="LLM completion timeout", exception=e)
             except Exception as e:
                 await self.push_error(error_msg=f"Error during completion: {e}", exception=e)
+                context.maybe_remove_invalid_message(e)
```
mattieruth: **Bug: `NameError` — `context` is not defined in this scope**

`context` is not a local variable in `process_frame`. The method only has `self`, `frame`, and `direction` in scope. Any non-timeout LLM exception will raise a `NameError` here, masking the real error.

The correct reference (matching the call two lines above) is `frame.context`.

```suggestion
            frame.context.maybe_remove_invalid_message(e)
```

src/pipecat/processors/frameworks/rtvi/processor.py
```diff
@@ -492,6 +506,95 @@ async def _handle_send_text(self, data: RTVI.SendTextData):
...
+                if not self._folder:
+                    logger.warning(
+                        "Send-file with a pipecat id requires uploads_folder on RTVIProcessor "
+                        "(e.g. uploads_folder=runner_uploads_folder())."
+                    )
+                    await self._send_error_response(message_id, "Uploads folder not set")
+                    return
+                # read bytes from file system, encode to base64, then delete the file
+                type = "bytes"
+                file_path = os.path.join(self._folder, fs.id.removeprefix("pipecat:"))
```
mattieruth: **Security: Path traversal — `FileId` source allows arbitrary file read and deletion**

`fs.id.removeprefix("pipecat:")` is client-controlled and passed directly to `os.path.join` with no containment check. A client can send `pipecat:../../../etc/passwd` (or an absolute path like `pipecat:/etc/passwd`, which causes `os.path.join` to discard `self._folder` entirely) to read and then `os.remove` any file the process can access.

The upload endpoint at `run.py` already follows the correct pattern via `_resolve_download_path` (which asserts `is_relative_to(allowed_base)`). The same validation should be applied here. The simplest fix is to enforce that the id suffix matches the UUID hex format produced by the upload endpoint:

```python
import re
suffix = fs.id.removeprefix("pipecat:")
if not re.fullmatch(r"[0-9a-f]{32}", suffix):
    logger.error(f"Invalid file id: {fs.id}")
    return
file_path = os.path.join(self._folder, suffix)
```

See the [existing safe pattern](https://github.com/pipecat-ai/pipecat/blob/90383e2a79384f3b67efe9c9a66d213b4e7a38c0/src/pipecat/runner/run.py#L761-L769) in `run.py`.

src/pipecat/processors/frameworks/rtvi/processor.py
```diff
@@ -492,6 +506,95 @@ async def _handle_send_text(self, data: RTVI.SendTextData):
...
+        source = None
+        type = file.source.type
+        opts = data.options if data.options is not None else RTVI.SendFileOptions()
+
+        logger.debug(f"Handling file from RTVI: format={file.format}, type={type}")
+
+        match file.source:
+            case RTVI.FileBytes() as fs:
+                source = fs.bytes
+            case RTVI.FileUrl() as fs:
```
mattieruth, line 527: **Security: SSRF — `FileUrl(public=False)` fetches a fully client-controlled URL with no restrictions**

`fs.url` is passed directly to `aiohttp.ClientSession().get()` with:
- No scheme allowlist (allows `file://`, `gopher://`, etc.)
- No host validation (allows `169.254.169.254` cloud metadata, RFC1918 addresses, and loopback)
- No redirect controls (aiohttp follows redirects by default, enabling open-redirect-to-internal attacks)
- No timeout or body size limit (enables DoS/memory exhaustion via slow or infinite responses)
- No HTTP status check — error responses (4xx/5xx bodies, HTML error pages) are silently base64-encoded and sent to the LLM

At minimum, the response status should be checked (`response.raise_for_status()`) and a reasonable timeout and body size cap should be applied. Host/scheme validation should also be added for production deployments.

src/pipecat/processors/frameworks/rtvi/processor.py
```diff
@@ -492,6 +506,95 @@ async def _handle_send_text(self, data: RTVI.SendTextData):
...
+        """Handle a send-file message from the client."""
+        file = data.file
+        source = None
+        type = file.source.type
+        opts = data.options if data.options is not None else RTVI.SendFileOptions()
+
+        logger.debug(f"Handling file from RTVI: format={file.format}, type={type}")
+
+        match file.source:
+            case RTVI.FileBytes() as fs:
```
mattieruth, line 521: **Bug: `FileBytes` source is not wrapped in a data URL, causing `IndexError` in all adapters**

For `FileBytes`, `source` is set to the raw base64 string from `fs.bytes`. But the `FileUrl(public=False)` and `FileId` branches (lines 527 and 549) both wrap their output as `f"data:{file.format};base64,{base64_string}"`. All three adapters downstream call `.split(",")[1]` on the stored value:

- `anthropic_adapter.py:392` — `f_data["file_data"].split(",")[1]`
- `bedrock_adapter.py:317` — `file_data_url.split(",")[1]`
- `gemini_adapter.py:470` — `file_data_url.split(",")[1]`

Raw base64 strings never contain a comma, so `[1]` always raises `IndexError`. Sending any file via `FileBytes` is broken end-to-end for all LLM providers.

Fix: normalize `FileBytes` to a data URL the same way the other branches do:

```suggestion
                source = f"data:{file.format};base64,{fs.bytes}"
```

src/pipecat/frames/frames.py
```diff
@@ -178,16 +174,36 @@ class ImageRawFrame:
...
+    """A frame containing a raw file.
+
+    Parameters:
+        file: Raw file bytes.
+        type: Type of the file ('bytes', 'url', or 'id'),
+        name: Optional name of the file.
```
mattieruth: **Bug: `FileRawFrame.name` is shadowed by `Frame.name` and cannot be set**

`Frame` (the base class reached via `SystemFrame`) declares `name: str = field(init=False)` and unconditionally overwrites it in `__post_init__` with `f"{class_name}#{count}"`. When `FileRawFrame` is mixed into `InputFileRawFrame(SystemFrame, FileRawFrame)`, the merged dataclass inherits the `init=False` semantics from `Frame.name`, so `name` is excluded from `__init__` entirely. Passing `name=` raises `TypeError`; after construction `frame.name` always returns the auto-generated identifier, never the filename.

This means the filename provided by the RTVI client (`file.name`) can never be stored in the frame, and adapters that consume a `filename` field (OpenAI, Bedrock) always receive an empty string.

Fix: rename the field to avoid the conflict — e.g. `filename: str | None = None` — and update `_handle_send_file` and `_handle_user_file_frame` to propagate it into `add_file_frame_message`.

src/pipecat/processors/frameworks/rtvi/models.py
```diff
@@ -241,6 +242,66 @@ class SendTextData(BaseModel):
...
+
+class File(BaseModel):
+    """File data structure for RTVI file sending."""
+
+    format: str  # Mime format of the file, e.g., 'application/pdf'
+    name: str | None = None
+    source: FileBytes | FileUrl | FileId
+
+
+class SendFileOptions(BaseModel):
```
mattieruth, line 293: **Incorrect docstring — copy-pasted from `SendTextOptions`**

This docstring describes "sending text input to the LLM" but the class is `SendFileOptions`. Violates the [CLAUDE.md](https://github.com/pipecat-ai/pipecat/blob/90383e2a79384f3b67efe9c9a66d213b4e7a38c0/CLAUDE.md) rule: "Classes describe purpose."

```suggestion
    """Options for sending a file to the LLM."""
```

src/pipecat/processors/aggregators/llm_response_universal.py
```diff
@@ -1869,6 +1872,23 @@ async def _handle_user_image_frame(self, frame: UserImageRawFrame):
         if image_appended:
             await self.push_context_frame(FrameDirection.UPSTREAM)
 
+    async def _handle_user_file_frame(self, frame: UserFileRawFrame):
```
mattieruth, line 2015: **Work-in-progress TODO and commented-out code left in production source**

Two artifacts that violate the [CLAUDE.md](https://github.com/pipecat-ai/pipecat/blob/90383e2a79384f3b67efe9c9a66d213b4e7a38c0/CLAUDE.md) rule "Leave the current moment out of it":

1. `# TODO: Should this have a similar function-call check like _handle_user_image_frame?` — a developer question from during implementation, not a constraint a future reader needs.
2. `#            options=frame.custom_options,` — commented-out code describing an incomplete feature. Note that simply uncommenting it would also fail since `add_file_frame_message` has no `options=` parameter.

Both should be removed; if `custom_options` propagation is intended, it needs a full implementation.

changelog/3415.added.md
```diff
@@ -0,0 +1,6 @@
+- Added support for clients sending files and images to the bot via the new RTVI `send-file` message, incorporating them into the LLM context.
+  - New frames: `FileRawFrame`, `InputFileRawFrame`, and `UserFileRawFrame` (parallel to the image frame hierarchy), carrying a file payload as bytes, a URL, or an upload ID.
```
filipi87: nit: not sure if we need to mention this: "(parallel to the image frame hierarchy)"

changelog/3415.added.md
```diff
@@ -0,0 +1,6 @@
...
+  - New `LLMContext` methods: `create_file_message()`, `create_file_url_message()`, and `add_file_frame_message()`, producing universal `file` (base64 data URL) and `file_url` content items translated per adapter (Anthropic, AWS Bedrock, Gemini, OpenAI Chat, OpenAI Responses). Unsupported MIME types or source combinations log a warning and are skipped.
+  - `RTVIProcessor` takes a new `uploads_folder` parameter for resolving `pipecat:<id>` file references produced by the upload endpoint.
+  - The development runner gains a `POST /files` upload endpoint (50 MB cap), enabled with the new `-u/--uploads-folder` flag; `--uploads-folder-max-files` (default 10) trims the oldest uploads.
```
filipi87: In general, the idea behind our development runner is to mimic the Pipecat Cloud behavior.

I wonder if we should put those file endpoints behind `/sessions/{session_id}/`, so:
* `/sessions/{session_id}/files`

This way, we would be able to make the same thing work for PCC. We should probably also remove all the files once the session has ended.

What do you think ?

src/pipecat/adapters/services/anthropic_adapter.py
```diff
@@ -379,8 +384,45 @@ def _from_standard_message(self, message: LLMStandardMessage) -> MessageParam:
...
+                        }
+                        del item["file"]
+                    else:
+                        logger.warning(
+                            f"Unsupported 'file_url' MIME type: {f_data['mime_type']} for URL: {f_data['url']}"
+                        )
+                        continue
+                if item["type"] == "file":
+                    f_data = item["file"]
+                    if f_data["mime_type"] != "application/pdf":
```
filipi87, line 427: Are we not able to handle `image/*` in this case ?  Anthropic doesn't support base64 encoded images?

src/pipecat/adapters/services/anthropic_adapter.py
```diff
@@ -379,8 +384,45 @@ def _from_standard_message(self, message: LLMStandardMessage) -> MessageParam:
...
+                            "type": "url",
+                            "url": f_data["url"],
+                        }
+                        del item["file"]
+                    else:
+                        logger.warning(
+                            f"Unsupported 'file_url' MIME type: {f_data['mime_type']} for URL: {f_data['url']}"
+                        )
+                        continue
+                if item["type"] == "file":
```
filipi87: Also, since the other type is explicitly called `file_url`, wouldn't it make sense to be explicit here as well and call this one `file_base64`?

Does that make sense?

src/pipecat/adapters/services/bedrock_adapter.py
```diff
@@ -321,6 +335,44 @@ def _from_standard_message(self, message: LLMStandardMessage) -> dict[str, Any]:
...
+                            {
+                                "document": {
+                                    "name": f_data["filename"] or "document",
+                                    "format": bedrock_format,
+                                    "source": {"s3Location": {"uri": url}},
+                                }
+                            }
+                        )
+                    else:
+                        logger.warning(f"Bedrock only supports S3 URLs for file sources: {url}")
```
filipi87: Interesting.

src/pipecat/adapters/services/gemini_adapter.py
```diff
@@ -459,6 +459,20 @@ def _from_standard_message(
                     input_audio = c["input_audio"]
                     audio_bytes = base64.b64decode(input_audio["data"])
                     parts.append(Part(inline_data=Blob(mime_type="audio/wav", data=audio_bytes)))
+                elif c["type"] == "file":
+                    f_data = c["file"]
+                    mime_type = f_data["mime_type"]
```
filipi87, line 521: Does this mean that Gemini basically works with all file types ?

src/pipecat/adapters/services/open_ai_adapter.py
```diff
@@ -204,11 +205,37 @@ def _from_universal_context_messages(
...
+                            f_data = item["file"]
+                            mime_type = f_data["mime_type"]
+                            if mime_type == "application/pdf":
+                                item = {
+                                    "type": "file",
+                                    "file": {
+                                        "filename": f_data["filename"],
+                                        "file_data": f_data["file_data"],
+                                    },
+                                }
```
filipi87, line 264: The `item` structure seems to be the same here.
 Are we recreating it just to remove the `mime_type`?

src/pipecat/processors/aggregators/llm_context.py
```diff
@@ -456,6 +524,33 @@ async def add_image_frame_message(
         )
         self.add_message(message)
 
+    async def add_file_frame_message(
```
filipi87, line 516: It looks like we're invoking this method from `_handle_user_file_frame`, where the type is `FileSourceType`, which means it could also be an `"id"`. Do we need to handle that case here?

Also, wouldn't it be safer to make the type an enum here?

src/pipecat/processors/aggregators/llm_context.py
```diff
@@ -184,6 +184,71 @@ def encode_image():
...
+            type: The type of the file (e.g., "bytes" or "url").
+            format: MIME type of the file (e.g., "application/pdf").
+            file: Base64 data URL (``data:<mime>;base64,...``) or plain URL string.
+            name: Optional name of the file.
+            text: Optional text to include with the file.
+        """
+        # Format is a public url. It is expected that the llm service will fetch the file from the url.
+        # The url is passed as is to the llm service.
+        # TODO: Deprecate support in adapters for passing a non-data url as part of the image_url interface.
+        #       all urls should be passed as a file_url message instead.
```
filipi87: nit: Is this TODO needed? It feels a bit misplaced.

src/pipecat/processors/aggregators/llm_context.py
```diff
@@ -184,6 +184,71 @@ def encode_image():
...
+        if type == "url":
+            return LLMContext.create_file_url_message(
+                role=role, format=format, url=file, filename=name, text=text
+            )
+
+        content: list[dict[str, Any]] = []
+        if text:
+            content.append({"type": "text", "text": text})
```
filipi87, line 214: Question: does it make any difference to the LLM whether we add the `text` before or after the actual file ?

src/pipecat/processors/aggregators/llm_context.py
```diff
@@ -468,6 +563,38 @@ async def add_audio_frames_message(
         message = await LLMContext.create_audio_message(audio_frames=audio_frames, text=text)
         self.add_message(message)
 
+    def maybe_remove_invalid_message(self, exception: Exception):
```
filipi87: Should we give it a name that specifically indicates it's removing invalid files from the context?

src/pipecat/processors/aggregators/llm_response_universal.py
```diff
@@ -1869,6 +1872,24 @@ async def _handle_user_image_frame(self, frame: UserImageRawFrame):
...
```
filipi87, line 2015: `UserFileRawFrame` can support `FileSourceType.id`, but none of the methods after this one support it.

I saw that you're handling this inside the RTVI processor, but maybe we should also handle it here in case we receive a `UserFileRawFrame` with `FileSourceType.id`, since it wouldn't be supported downstream.

Or maybe `FileSourceType.id` shouldn't be an option for this frame at all, and should instead be something specific to RTVI?

src/pipecat/processors/frameworks/rtvi/models.py
```diff
@@ -241,6 +242,63 @@ class SendTextData(BaseModel):
     options: SendTextOptions | None = None
 
 
+class FileSource(BaseModel):
+    """Base class for RTVI file sources."""
+
+    type: FileSourceType
```
filipi87: Like I mentioned above, maybe this type shouldn't come from the `FileSourceType` on the frame, since it looks like RTVI needs to support:
* `"bytes"`
* `"url"`
* `"id"`

But inside `FileRawFrame`, everywhere else in Pipecat, we're expecting the source type to always be either `"bytes"` or `"url"`.

What do you think ?

src/pipecat/processors/frameworks/rtvi/processor.py
```diff
@@ -59,6 +65,7 @@ def __init__(
         self,
         *,
         transport: BaseTransport | None = None,
+        uploads_folder: str | None = None,
```
filipi87: When creating the processor, this would need to match the argument passed to the runner, right (`args.uploads_folder`)?

So, does that mean we'd need to pass the same path twice ? Once when creating the runner and once when creating the `RTVIProcessor`?

I wonder if we should have something like a `FileStorageUtil` that relies on a single environment variable, which both the runner and the `RTVIProcessor` could use to save and retrieve files instead.

filipi87: Even extrapolating on this, having a base storage abstraction shared by both the runner and RTVIProcessor could allow users to override the "development storage" we provide by default with something more production ready.

They could then plug it into a proper storage backend, such as a custom S3 implementation, instead of relying solely on this limited local storage.

src/pipecat/processors/frameworks/rtvi/processor.py
```diff
@@ -492,6 +508,134 @@ async def _handle_send_text(self, data: RTVI.SendTextData):
...
+        match file.source:
+            case RTVI.FileBytes() as fs:
+                # `bytes` is raw base64, but tolerate clients that send a full data URL.
+                if fs.bytes.startswith("data:"):
+                    source = fs.bytes
+                else:
+                    source = f"data:{file.format};base64,{fs.bytes}"
+            case RTVI.FileUrl() as fs:
+                if not fs.public:
+                    if not fs.url.startswith(("http://", "https://")):
```
filipi87, line 529: I guess it would be nice to isolate all this logic into a helper method whose only responsibility is to download the file.

Maybe we could allow users to override this as well ?

For example, they might want to download a file from S3 or from some other storage that requires credentials.

src/pipecat/processors/frameworks/rtvi/processor.py
```diff
@@ -492,6 +508,134 @@ async def _handle_send_text(self, data: RTVI.SendTextData):
...
+                        logger.warning(f"Timed out fetching file from URL: {fs.url}")
+                        await self._send_error_response(message_id, "Timed out fetching file")
+                        return
+                    except aiohttp.ClientError as e:
+                        logger.warning(f"Failed to fetch file from URL: {e}")
+                        await self._send_error_response(message_id, f"Failed to fetch file: {e}")
+                        return
+                else:
+                    source = fs.url
+            case RTVI.FileId() as fs:
```
filipi87, line 559: It will be impacted in case we wish to move with the File storage approach which I mentioned above.

src/pipecat/runner/run.py
```diff
@@ -1401,6 +1432,45 @@ async def websocket_endpoint_with_token(websocket: WebSocket, token: str):
         await _handle_telephony_ws(websocket, path_token=token)
 
 
+_MAX_UPLOAD_BYTES = 50 * 1024 * 1024
+
+
+def _setup_file_uploads_route(app: FastAPI, args: argparse.Namespace):
```
filipi87: Same here. I added two comments above regarding the route, using `sessionId` to match Pipecat Cloud, and the storage approach, which could impact how we've implemented things here in the runner.

filipi87:
@mattieruth, I also asked Claude to review it. Here’s what it found:

There are a couple of things we can probably ignore, depending on what we decide in the discussions above.

## Bugs & Correctness

- [x] **High** — `src/pipecat/runner/run.py:929` and `src/pipecat/runner/run.py:933`

  The `-f/--folder` → `-f/--downloads-folder` rename gave both argparse arguments `dest="downloads_folder"`, and no argument declares `dest="folder"` any more. The WebRTC download route still reads `args.folder`:

  ```python
  @app.get("/files/{filename:path}")
  async def download_file(filename: str):
      """Handle file downloads."""
      if not args.folder:                                        # line 929
          ...
      file_path = _resolve_download_path(args.folder, filename)  # line 933
  ```

  The route is registered from `_setup_webrtc_routes`, called unconditionally from `main()` and gated only on `_transport_routes_enabled("webrtc")` — which is on by default when `-t` is omitted or set to `webrtc`. The first `GET /files/<name>` raises `AttributeError: 'Namespace' object has no attribute 'folder'` (HTTP 500) instead of serving the file or returning 404, regardless of whether a downloads folder was configured. Runner file downloads are broken.

  Suggested fix: change both references to `args.downloads_folder`, matching `RUNNER_DOWNLOADS_FOLDER = args.downloads_folder` at line 1900.

- [ ] **Medium** — `src/pipecat/processors/frameworks/rtvi/processor.py:611`

  `SendFileOptions.run_immediately` does not prevent LLM inference. In `_handle_send_file` it gates only `interrupt_bot()` (line 625); both frame constructions hardcode `append_to_context=True` (lines 612 and 622), and `LLMAssistantAggregator._handle_user_file_frame` (`llm_response_universal.py:1890`) plus the image path via `_maybe_append_image_to_context` end with an unconditional `push_context_frame(FrameDirection.UPSTREAM)`, which makes the LLM generate a response.

  This diverges from the identically-named, identically-defaulted `SendTextOptions.run_immediately`, which `_handle_send_text` propagates as `LLMMessagesAppendFrame(..., run_llm=opts.run_immediately)`; `_handle_llm_messages_append` then pushes the context frame only when `run_llm` is truthy. A client sending a file with `run_immediately: false` expects the file to be staged in context but still gets an immediate bot response.

  Suggested fix: carry the flag on the frame (e.g. a `run_llm` field on `UserFileRawFrame`, and reuse the equivalent plumbing on the image branch) and push the upstream context frame only when it is set.

- [ ] **Medium** — `src/pipecat/processors/frameworks/rtvi/processor.py:628`

  `audio_response=False` is not reliably honored, because the file frame can overtake the `LLMConfigureOutputFrame`s that bracket it. `LLMConfigureOutputFrame` is a `DataFrame` (`frames.py:744`) while `UserFileRawFrame` and `UserImageRawFrame` are `SystemFrame`s. Two mechanisms in `frame_processor.py` break the intended ordering:

  - `FrameProcessorQueue` (`frame_processor.py:119-167`) is an `asyncio.PriorityQueue` that assigns `HIGH_PRIORITY` to `SystemFrame` and `LOW_PRIORITY` to everything else.
  - `__input_frame_task_handler` (`frame_processor.py:1029-1036`) processes system frames inline, while non-system frames are deferred to a separate `__process_queue` task.

  So the `skip_tts=True` frame travels the deferred data path while the file frame races ahead through the inline system path to the assistant aggregator, which immediately pushes `LLMContextFrame` upstream. The LLM can ingest the context before it applies `skip_tts=True`, and symmetrically the restore frame at line 636 can land before the response finishes. `_handle_send_text` is unaffected because `LLMMessagesAppendFrame` is also a `DataFrame`.

  Suggested fix: don't rely on ordering across the system/data frame boundary — carry the TTS preference on the payload frame itself and have the LLM apply it when ingesting the context.

- [x] **Medium** — `src/pipecat/processors/frameworks/rtvi/processor.py:527`

  Server-side request forgery via client-supplied URLs. For `FileUrl` with `public=False`, the processor issues a server-side `GET` to an arbitrary client-supplied URL (line 538). The only validation is the `http://`/`https://` scheme check at line 529 — there is no host or IP restriction, no allowlist, and no opt-in gate on `RTVIProcessor`. A client can make the bot's host fetch internal endpoints (`http://localhost:*`, `http://169.254.169.254/…`, RFC1918 ranges) and have up to 50 MB of the response base64-inlined into the LLM context, where it is readable back through the bot's reply. `aiohttp` follows redirects by default, so validating only the initial URL would be insufficient.

  Fetching a URL the LLM provider cannot reach is the stated purpose of `public=False`, so the fix is to bound who can be reached rather than to forbid the fetch.

  Suggested fix: gate server-side fetching behind an explicit `RTVIProcessor` option (default off) and/or an operator-supplied host allowlist; resolve the hostname and reject loopback, link-local, and private/ULA targets; pass `allow_redirects=False` or re-validate the host on each hop.

- [ ] **Low** — `src/pipecat/processors/aggregators/llm_response_universal.py:1890`

  `_handle_user_file_frame` calls `push_aggregation()` *after* `add_file_frame_message()` (line 1881), so a pending assistant turn is committed behind the user's file message. With `run_immediately=True` the preceding `interrupt_bot()` drains the aggregation first and the call is a no-op; with `run_immediately=False` there is no interruption, so a file sent mid-response inverts the two messages in context. There is no state in which the current placement is the correct one. The sibling `_handle_user_image_frame` (lines 1854-1873) does not call `push_aggregation()` at all, so the two handlers disagree.

  Suggested fix: move `await self.push_aggregation()` above the `add_file_frame_message()` call, and align `_handle_user_image_frame` with it.

## Performance

- [ ] **Medium** — `src/pipecat/adapters/services/bedrock_adapter.py:346`, `src/pipecat/adapters/services/gemini_adapter.py:470`, `src/pipecat/adapters/services/anthropic_adapter.py:420`

  A file's base64 payload is re-processed from scratch on every LLM turn for the life of the session. `LLMContext.create_file_message` (`llm_context.py:220`) stores the base64 data URL permanently in the message history, and `_process_context` re-runs `_from_universal_context_messages` over the *entire* history on each generation (`services/aws/llm.py:432`, `services/google/llm.py:443`, `services/anthropic/llm.py:377`). Bedrock and Gemini therefore run `base64.b64decode` on the same payload — up to 50 MB — synchronously on the event loop on every conversational turn. Anthropic's impact is smaller: it only re-slices the string (`f_data["file_data"].split(",")[1]`) rather than decoding, but still allocates a full copy per turn.

  Suggested fix: decode/slice once when the file frame first becomes a context message and cache the result on the content item, and/or offload the decode via `asyncio.to_thread` if it must stay per-call.

- [ ] **Medium** — `src/pipecat/processors/frameworks/rtvi/processor.py:547` and `src/pipecat/processors/frameworks/rtvi/processor.py:586`

  Redundant base64 round-trip for image files arriving via URL or upload ID. In the non-public `FileUrl` and `FileId` branches the raw bytes are already in hand as `raw_bytes`, but are base64-encoded into a data URL string to build a uniform `source`. When the file turns out to be an image, that same string is immediately decoded back to bytes at line 603 to construct `UserImageRawFrame`, even though `raw_bytes` is still in scope and unused — two full CPU-bound base64 passes over up to 50 MB on the event loop.

  Suggested fix: use `raw_bytes` directly for the image-frame path in those two branches.

- [ ] **Medium** — `src/pipecat/runner/run.py:1456` and `src/pipecat/runner/run.py:524`

  Synchronous filesystem I/O inside an async route handler. `upload_file` is `async def` (line 1440) but calls `file_path.write_bytes(contents)` synchronously for up to 50 MB (line 1456), then `_trim_uploads_folder` (lines 524-540, called at line 1461) performs synchronous `folder.iterdir()`, `p.stat()`, and `p.unlink()`. The development runner launches bots with `asyncio.create_task(bot_module.bot(runner_args))` in the same process and loop as `uvicorn.run(app, ...)` (line 1909) — there is no subprocess or worker-pool isolation — so a large upload or a trim over many files blocks the loop shared with any running pipeline and can stall audio processing.

  Suggested fix: use `aiofiles` (already a dependency, used at `processor.py:580`) for the write, and wrap the trim in `asyncio.to_thread(...)`.

## Repository Conventions & Docstrings

- [ ] **Low** — `src/pipecat/frames/frames.py:196`

  In the newly added `FileRawFrame`, the `type:` entry of the `Parameters:` block ends with a comma instead of a period, while every sibling entry (lines 194, 197, 198) and the adjacent `ImageRawFrame` docstring (lines 176-179) use periods.

  Rule:

  > "**Docstrings**: Google-style. Classes describe purpose; `__init__` has `Args:` section; dataclasses use `Parameters:` section."

  Source: `AGENTS.md`

  Suggested fix: `type: Type of the file ('bytes', 'url', or 'id').`

src/pipecat/adapters/services/anthropic_adapter.py
mattieruth: images are sent as a different type and are already handled above

src/pipecat/adapters/services/bedrock_adapter.py
mattieruth: yup.

mattieruth: https://docs.aws.amazon.com/bedrock/latest/APIReference/API_runtime_DocumentSource.html

src/pipecat/adapters/services/gemini_adapter.py
mattieruth: good question. the format would indicate so, but looking back at the docs, it only talks about pdfs. so maybe it just supports pdfs? https://ai.google.dev/gemini-api/docs/document-processing

i'll test.

src/pipecat/processors/aggregators/llm_response_universal.py
mattieruth: this is probably one of those cases where i added something without full support. most apis support a way to upload files and then address them by ID. Right now, though. I'm not supporting that use case. I'm overriding the format i imagine we would use for it for our own similar file upload support, using an id that makes it clear it's a local file. maybe you're right that I should remove it from this type for now since it's not supported. maybe i'll also look to see if it's not that hard to support. But users would have to do their own implementation of the upload portion of that API. This would just assume that the file had already been uploaded and the id is known.

src/pipecat/adapters/services/open_ai_adapter.py
mattieruth: it's because the messages is a shallow copy, so this ensures the original context stays in tact.

src/pipecat/processors/aggregators/llm_context.py
mattieruth: i think this is named correctly. right now it only removes files but i don't see why it couldn't be extended in the future.

src/pipecat/adapters/services/anthropic_adapter.py
mattieruth: yup! renaming..

src/pipecat/processors/aggregators/llm_context.py
mattieruth: it is, but i've moved it to the image handling method for better placement. the idea is that i think there should only be one path for urls and as this one is more generic, the preferred one.

src/pipecat/processors/aggregators/llm_context.py
mattieruth: I've been sending before, following the pattern of images. However, anthropic does document that it's best practice is to send the text after. So i'm leaving the default as-is, but will fix anthropic to swap the order (again, following image's lead)

changelog/3415.added.md
mattieruth: This got lost in my recent changes, but I wanted to address it.  This makes total sense. However, I think I want to do this in a follow-up. The changes are already so vast and my PCC knowledge is minimal, so I want to tackle it separately if that is ok.

src/pipecat/processors/aggregators/llm_context.py
mattieruth: you're right. i've removed "id" as an option for FileSourceType for now. We should consider supporting File APIs and their ids in a follow-up PR and consider how/if that relates to RTVI. But I did update the 'type' type

changelog/3415.added.md
filipi87: Yep, we can do that. Maybe we’ll need to deprecate the current routes in case we’re not able to do it before the next release, but I think it should be fine.

Also, probably `@app.get("/files/{filename:path}")` should be inside `_setup_file_uploads_route` instead of `_setup_webrtc_routes`, but we can change that as well in this follow-up PR.

src/pipecat/adapters/services/open_ai_adapter.py
filipi87: Makes sense. 👍

changelog/3415.added.md
```diff
@@ -0,0 +1,7 @@
+- Added support for clients sending files and images to the bot via the new RTVI `send-file` message, incorporating them into the LLM context.
+  - New frames: `FileRawFrame`, `InputFileRawFrame`, and `UserFileRawFrame`, carrying a file payload as bytes, a URL, or an upload ID.
+  - New `LLMContext` methods: `create_file_message()`, `create_file_url_message()`, and `add_file_frame_message()`, producing universal `file_base64` (base64 data URL) and `file_url` content items translated per adapter (Anthropic, AWS Bedrock, Gemini, OpenAI Chat, OpenAI Responses). Unsupported MIME types or source combinations raise during conversion rather than being silently dropped, so the invalid message can be cleaned up automatically (see below). AWS Bedrock document names are sanitized to satisfy Bedrock's character restrictions.
+  - New `FileStorage` interface (`pipecat.utils.file_storage`) with a `LocalFileStorage` default implementation, shared by the development runner's upload endpoint and `RTVIProcessor` so both resolve the same `pipecat:<id>` file references without duplicating configuration. Implement `FileStorage` to back uploads with something other than local disk in production.
```
filipi87, line 4: Nice. 🚀

filipi87:
Hi @mattieruth, I also asked Claude to review it, and it found a couple of issues that feel legit. For example, regarding the one saying that we currently try to download from any source, maybe we should consider some approach to restrict it, similar to what we did with `PIPECAT_ALLOWED_ORIGINS`, but in this case, we would need to define from what places we should be allowed to download.


## Bugs & Correctness

- [x] **High** — `src/pipecat/runner/run.py:991`, `src/pipecat/runner/run.py:995`

  The `-f/--folder` → `-f/--downloads-folder` rename set `dest="downloads_folder"` on both the new flag (lines 1799-1804) and the deprecated `--folder` alias (lines 1806-1811), so the namespace only ever carries `downloads_folder`. The WebRTC download route still reads `args.folder`. Nothing else in the repository assigns that attribute — `main()` was updated at line 2078, this handler was not.

  Failure: any `GET /files/{filename}` against the runner raises `AttributeError: 'Namespace' object has no attribute 'folder'` and returns 500, even when `--downloads-folder` is set. The endpoint is unconditionally broken, not just in the unconfigured case.

  Suggested fix: read `args.downloads_folder` at both lines.

- [x] **High** — `src/pipecat/processors/frameworks/rtvi/processor.py:538`

  The `FileUrl` branch with `public=False` makes the bot process perform a server-side GET on a URL supplied verbatim by the client (`_handle_send_file`, lines 527-551). The only validation is the `http://`/`https://` scheme check — no host or IP restriction, and aiohttp follows redirects by default, so even an allowed public URL can redirect into internal space. The response body is base64-encoded into the LLM context, where the model can read it back to the caller. The `send-file` message is dispatched unconditionally (line 378); unlike the file-ID path, this needs no configuration to reach.

  Failure: a client sends `{"content": "summarize this", "file": {"format": "text/plain", "source": {"type": "url", "url": "http://169.254.169.254/latest/meta-data/iam/security-credentials/", "public": false}}}`. The bot fetches cloud instance metadata — or any internal service reachable from the bot host — and the model repeats it. The distinct error responses ("Unsupported URL scheme" / "Failed to fetch file" / success) also make the endpoint usable for internal port and service probing.

  Suggested fix: resolve the host and reject loopback, link-local, private, unique-local, and metadata addresses, re-checking after each redirect (or disabling redirects), and/or gate server-side fetching behind an explicit opt-in such as a host allowlist on `RTVIProcessor`.

- [x] **Medium** — `src/pipecat/services/anthropic/llm.py:612`, `src/pipecat/services/aws/llm.py:638`, `src/pipecat/services/google/llm.py:851`

  All three services treat the entire 4xx range as evidence that a file was bad and call `remove_invalid_file_message()`. But 4xx also covers failures unrelated to the payload, and each SDK routes them through the same exception: Anthropic's `RateLimitError`, `AuthenticationError`, `PermissionDeniedError`, and `NotFoundError` are all `APIStatusError` subclasses; a Bedrock `ThrottlingException` arrives as a `ClientError` with `HTTPStatusCode` 429; and `google.genai.errors.ClientError` is raised for every 4xx. The method's docstring scopes its best-effort caveat to *which* file message is the culprit, and states the precondition that the service has "determined (from a provider-specific error) that a request was rejected due to unsupported or malformed file content" — a 401/403/404/429 does not meet it. The OpenAI path in the same change (`base_llm.py:617-622`) keys precisely on `type == "invalid_request_error"` and `param == "file_id"`, which is the intended precision.

  Failure: the user uploads a PDF and the next completion hits a rate limit (429). The PDF message is popped from context and never returns; the retry succeeds and the model answers confidently about a document it can no longer see. This is silent apart from a `logger.warning`, and worst under load, when 429s cluster.

  Suggested fix: restrict removal to payload-related statuses (400/413/415/422), or key off provider error codes — Anthropic `invalid_request_error`, Bedrock `ValidationException`, Gemini `INVALID_ARGUMENT` — as the OpenAI path already does.

- [x] **Medium** — `src/pipecat/processors/frameworks/rtvi/processor.py:611`

  `SendFileOptions.run_immediately` never reaches the pipeline. `_handle_send_text` threads it into the frame (`run_llm=opts.run_immediately`, line 504) and the aggregator gates inference on it (`llm_response_universal.py:1738`). `_handle_send_file` uses the flag only to decide whether to call `interrupt_bot()`; `UserFileRawFrame` carries no equivalent, and `_handle_user_file_frame` calls `push_context_frame(FrameDirection.UPSTREAM)` unconditionally (`llm_response_universal.py:2031`) — the same mechanism `_handle_llm_run` uses to trigger a completion. The image branch of `_handle_send_file` has the same gap.

  Failure: a client sends a file with `options: {"run_immediately": false}` to stage it for the user's next turn. The bot is not interrupted, so it keeps speaking, but inference fires anyway — producing an overlapping, unwanted response. The pre-existing image path behaves this way too, but it has no such option to honor; here a documented public RTVI option is silently ignored.

  Suggested fix: add a `run_llm`-style field to `UserFileRawFrame`, set it from `opts.run_immediately`, and gate the upstream push on it.

- [x] **Medium** — `src/pipecat/adapters/services/gemini_adapter.py:531`

  The Gemini adapter logs a warning for a `file_url` item and appends no `Part`, so the file vanishes from the request while the accompanying text still goes through. Because nothing raises, `LLMContextConversionError` never fires and `remove_invalid_file_message()` is never reached — the dead content item stays in the context and is silently dropped on every subsequent turn. The adapter already handles non-data URLs four lines earlier via `Part.from_uri(file_uri=url)` (line 512), so the capability exists. This also contradicts `changelog/3415.added.md`: "Unsupported MIME types or source combinations raise during conversion rather than being silently dropped, so the invalid message can be cleaned up automatically."

  Failure: a client sends a PDF or image by public URL to a Gemini bot. The model answers as if only the text existed, no error reaches the client, and the situation repeats every turn thereafter.

  Suggested fix: map `file_url` to `Part.from_uri(file_uri=url, mime_type=f_data["mime_type"])`, or raise so the documented cleanup path engages.

- [x] **Medium** — `src/pipecat/adapters/services/open_ai_responses_adapter.py:297`, `src/pipecat/adapters/services/open_ai_adapter.py:271`

  `_handle_send_file` routes to `UserImageRawFrame` only when `file.format.startswith("image/") and type != "url"` (`processor.py:583`), so an image sent as a *public* URL becomes a `file_url` content item carrying an `image/*` MIME type. The Responses adapter emits it as `{"type": "input_file", "file_url": ...}` — the document shape — while emitting the correct `input_image` for `image_url` items at lines 279-287; the MIME type is available in `f_data["mime_type"]` and goes unused. In the OpenAI Chat adapter the same input is worse: *any* `file_url` raises `LLMContextConversionError`, including an image URL OpenAI Chat supports natively via `image_url`, so the turn always fails — and, per the finding above, the message is then popped from context.

  Failure: a client calls `sendFile()` with `{"format": "image/png", "source": {"type": "url", "url": "https://example.com/x.png", "public": true}}`. On OpenAI Chat the turn errors out; on Responses the image is delivered in a document-shaped part rather than as an image.

  Suggested fix: branch on `mime_type.startswith("image/")` in both adapters and emit `input_image` / `image_url` respectively — or fix it one level up by routing image-plus-URL to `create_image_url_message`, which corrects every adapter at once.

- [x] **Low** — `src/pipecat/adapters/services/bedrock_adapter.py:395`

  `_MIME_TO_BEDROCK_IMAGE_FORMAT` is consulted only in the `file_url`/`s3://` branch (line 432). The `file_base64` branch checks video, then the document table, then raises — so `image/png` as `file_base64` raises "Unsupported 'file' MIME type for Bedrock" even though Bedrock accepts inline image bytes and the mapping sits in the same file. Not reachable through RTVI today, since inline images route to `UserImageRawFrame`, but `LLMContext.create_file_message()` and `add_file_frame_message()` are public API.

  Suggested fix: consult `_MIME_TO_BEDROCK_IMAGE_FORMAT` in the `file_base64` branch before the document lookup and emit an `image` block with `source.bytes`.

## Performance

- [x] **Medium** — `src/pipecat/utils/file_storage.py:99`

  `LocalFileStorage.save` calls `self._trim()` directly rather than offloading it. `_trim()` (lines 126-142) does a synchronous `iterdir()`, a `is_file()`/`stat()` per entry, a sort, and one or more `unlink()` calls, all on the event loop. The sibling `delete()` wraps its `unlink` in `asyncio.to_thread`, so the inconsistency is unlikely to be deliberate. In the development runner the upload route shares the loop with running bot pipelines, so this blocks in-flight audio and frame processing. The cost is bounded — trimming keeps the folder near `max_files` (default 10) — but the blocking is avoidable. `self._folder.mkdir(...)` on line 97 has the same shape at negligible cost.

  Suggested fix: `await asyncio.to_thread(self._trim)`, matching the pattern already used in `delete()`.

- [x] **Medium** — `src/pipecat/processors/frameworks/rtvi/processor.py:589`

  For a `FileId` source the bytes are loaded from storage and base64-encoded into a data URL (line 575); for a private `FileUrl` they are fetched and encoded the same way (line 547). Both branches then set `type = "bytes"`, so an image file reaches the image branch at line 583, which immediately base64-decodes that very string back into bytes for `UserImageRawFrame.image`. That is a full decode pass over a ~66 MB string that was just produced from bytes already in hand, run synchronously inside a real-time pipeline frame handler.

  Suggested fix: keep `raw_bytes` in those branches and pass it straight to `UserImageRawFrame`, building the base64 data URL only for the non-image `UserFileRawFrame` path that needs it.

- [ ] **Medium** — `src/pipecat/adapters/services/bedrock_adapter.py:399`, `src/pipecat/adapters/services/gemini_adapter.py:519`

  Both adapters `base64.b64decode()` a `file_base64` item's full payload inside the message-conversion loop, which runs synchronously from async pipeline code (`_process_context` → `get_llm_invocation_params` → `_from_universal_context_messages`) on every LLM inference. Nothing prunes file content items from the context — `remove_invalid_file_message()` only handles error recovery — so a large file is re-decoded on every turn for the rest of the conversation. The Anthropic and OpenAI adapters forward the base64 string without decoding. This extends the pre-existing `image_url` decode pattern to a new content type rather than introducing it, but files are capped at 50 MB where inline images are frame-sized, so the recurring cost is materially larger.

  Suggested fix: cache the decoded bytes on the context item so repeat turns skip the decode, or offload the conversion to a worker thread.

## Repository Conventions & Docstrings

- [x] **Low** — `src/pipecat/frames/frames.py:204`

  In `FileRawFrame`'s `Parameters:` block, the `type:` entry ends with a comma where a period is expected. Of the 126 single-line `Parameters:` entries in this file, 125 end with a period; this is the sole exception, and it is new in this branch. Ruff's docstring rules do not check punctuation inside Google-style section bodies, so nothing catches it automatically.

  Rule:

  > "**Dataclasses:** Class docstring describes the purpose and documents all fields in a `Parameters:` section"

  Source: `CONTRIBUTING.md` (Docstring Conventions)

  Suggested fix: `type: Type of the file ('bytes' or 'url').`

- [x] **Low** — `src/pipecat/frames/frames.py:1586`

  `UserFileRawFrame.custom_options`'s continuation line is indented 24 spaces to align under the description text, rather than the 4-space hanging indent (12 spaces here) used by all 80 other multi-line `Parameters:` continuations in this file. It is the only exception, and it is new in this branch.

  Rule:

  > "**Dataclasses:** Class docstring describes the purpose and documents all fields in a `Parameters:` section"

  Source: `CONTRIBUTING.md` (Docstring Conventions)

  Suggested fix: re-indent the continuation line to 12 spaces.

src/pipecat/processors/aggregators/llm_context.py
```diff
@@ -125,6 +132,9 @@ def create_image_url_message(
             url: The URL of the image.
             text: Optional text to include with the image.
         """
+        # TODO: Deprecate support for a non-data `url` here now that `create_file_url_message`
+        #       exists — a URL-referenced image and a URL-referenced file should go through the
+        #       same path instead of two.
```
filipi87, line 137: Do we intend to add the deprecation in this PR or in a follow-up one ?

src/pipecat/processors/aggregators/llm_response_universal.py
```diff
@@ -2009,6 +2012,24 @@ async def _handle_user_image_frame(self, frame: UserImageRawFrame):
...
+        # TODO: Should this have a similar function-call check like _handle_user_image_frame?
```
filipi87, line 2016: I believe this could make sense. This way we could provide a file as a result of a function call, the same way we do with the image.

src/pipecat/runner/run.py
```diff
@@ -1483,6 +1495,38 @@ async def websocket_endpoint_with_token(websocket: WebSocket, token: str):
         await _handle_telephony_ws(websocket, path_token=token)
 
 
+_MAX_UPLOAD_BYTES = 50 * 1024 * 1024
+
+
+def _setup_file_uploads_route(app: FastAPI):
```
filipi87, line 1501: I believe the `@app.get("/files")`, should be here as well. Instead of inside the `_setup_webrtc_routes`

filipi87: Also, the I believe the logic inside `@app.get("/files")` should try to retrieve the file using `FileStorage`, instead of trying to download it directly from the disk.

filipi87:
I believe we need to change the way we create and handle the `FileStorage`.

Right now, the only way to get the storage that the runner's upload route uses is `runner_file_storage()`, which reads a module global set in `main()`. But that will not work with Pipecat Cloud, and the user has no way to provide their own storage. Unless they also reimplement all the routes to download and upload, but I think we can provide that in the base image.

The only approach I've thought of so far is to let the bot module declare its storage the same way it declares `bot()`, and have the host inject it:

```
  # bot.py — optional, alongside your bot() function
  def create_file_storage() -> FileStorage:
      return S3FileStorage(bucket=os.environ["UPLOADS_BUCKET"])                            


  # The bot can then read it from the runner arguments
      rtvi = RTVIProcessor(file_storage=runner_args.file_storage)
```

This way both the dev runner today and Pipecat Cloud later can use it for their own upload route. 

In our development runner that route is POST `/files`, and I think it should become POST `/sessions/{session_id}/files`, which matches the `/sessions/{session_id}/{path:path}` proxy route we already have, and matches what is expected inside Pipecat Cloud. This way we can implement those routes inside the Pipecat Cloud base image, as we do for all the other routes.

For the development runner we can still create a `LocalFileStorage` when `uploads_folder` is set and the bot hasn't defined `create_file_storage`. But at least this gives the user a way to override the `FileStorage`.

Probably worth getting a second opinion from @markbackman or @kompfner, in case I am overcomplicating here.

src/pipecat/adapters/services/gemini_adapter.py
mattieruth: i forgot to loop back here. I tested and it does work with at least most of the file types i tried.

src/pipecat/processors/aggregators/llm_context.py
mattieruth: i was thinking follow-up, unless you disagree.

mattieruth:
OK. I've addressed all the rest of the feedback EXCEPT the big PipecatCloud -- endpointing question. 

One exception, here's claude's response to your claude's suggestion to cache the decoded files 😂 
> #-3 under Performance — Cache decoded base64 bytes per turn (bedrock_adapter.py, gemini_adapter.py): I disagree this is worth doing. The decode is a small fraction of per-turn cost — the full file bytes get re-serialized and re-sent to the provider over the network every turn regardless, which is inherent to how LLMContext works and isn't fixed by caching the decode step alone. A correct cache would need real design (content-hash keys, bounded/LRU eviction, invalidation across context copies/serialization) to avoid becoming its own source of bugs. This reads as premature optimization without a measured problem, so I left it alone — flag if you'd like me to revisit with profiling data.

I am likely going to rebase all these changes to make reviewing clearer and add the new endpointing/file upload work we discussed offline. I'll likely close this PR and replace it with the cleaner rebase.

PR #3417, opened 2026-01-12
frames awaited

dhruvladia-sarvam:
Respective frames awaited on start speech and end speech events to enable user interruption and eliminate use of local vad for turn detection strategy


markbackman:
Superseded by https://github.com/pipecat-ai/pipecat/pull/3495.

See this commit:
https://github.com/pipecat-ai/pipecat/pull/3495/commits/4491267f3e24ebd6777242f313d7794dc8f12d85

PR #3420, opened 2026-01-12
Fix parallel function calling with Gemini 3.

kompfner:
Gemini expects parallel function calls to be passed in as a single multi-part `Content` block. This is important because only one of the function calls in a batch of parallel function calls gets a thought signature—if they're passed in as separate `Content` blocks, there'd be one or more missing thought signatures, which would result in a Gemini error.

Fixes https://github.com/pipecat-ai/pipecat/issues/3290

### DIFF
--- a/src/pipecat/adapters/services/gemini_adapter.py
+++ b/src/pipecat/adapters/services/gemini_adapter.py
@@ -256,7 +256,7 @@ class GeminiLLMAdapter(BaseLLMAdapter[GeminiLLMInvocationParams]):
         self._apply_thought_signatures_to_messages(thought_signature_dicts, messages)
 
         # When thinking is enabled, merge parallel tool calls into single messages
-        messages = self._merge_parallel_tool_calls_for_thinking(messages)
+        messages = self._merge_parallel_tool_calls_for_thinking(thought_signature_dicts, messages)
 
         # Check if we only have function-related messages (no regular text)
         has_regular_messages = any(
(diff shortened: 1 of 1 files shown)


aconchillo:
LGTM!

kompfner:
Hm, thought of one more thing I want to test here before merging—the case of back-to-back tool calls that aren't parallel, but just don't have any assistant response in between...

src/pipecat/adapters/services/gemini_adapter.py
```diff
@@ -433,6 +436,89 @@ def _from_standard_message(
...
+
+                # Output merged calls, then collected other messages
+                merged_messages.append(Content(role="model", parts=merged_parts))
+                merged_messages.extend(other_messages)
+                i = j
+            else:
+                merged_messages.append(current)
+                i += 1
+
+        return merged_messages
```
aconchillo, line 534: Could we just check `if not thought_signature_dicts` and only call `_merge_parallel_tool_calls_for_thinking` otherwise don't do anything because it seems that doesn't matter.

so

```
if thought_signature_dicts:
    messages = self._merge_parallel_tool_calls_for_thinking(messages)  
```
if thought_signature_dicts is empty, maybe `_apply_thought_signatures_to_messages` could return a True or False?


would that be posssible?

kompfner: Good idea! Will do.

PR #3422, opened 2026-01-13
fix: openai llm model name is unknown

monster-anshu:

This PR fixes an issue in `traced_llm` where the model name was reported as `"unknown"` in OpenTelemetry even when a model was correctly configured.

The root cause was incorrect model resolution logic, which failed to consistently pick the available model value. The logic has now been updated to resolve the model name in a clear priority order, ensuring the correct value is always propagated to OpenTelemetry.

## Fix

The model name is now resolved using the following fallback order:

1. `_full_model_name`
2. `model_name`
3. `params["model"]`
4. `"unknown"` (final fallback)

```python
"model": (
    getattr(self, "_full_model_name", None)
    or getattr(self, "model_name", None)
    or params.get("model")
    or "unknown"
)
```

### DIFF
--- a/src/pipecat/utils/tracing/service_decorators.py
+++ b/src/pipecat/utils/tracing/service_decorators.py
@@ -483,7 +483,10 @@ def traced_llm(func: Optional[Callable] = None, *, name: Optional[str] = None) -
                             # Add all available attributes to the span
                             attribute_kwargs = {
                                 "service_name": service_class_name,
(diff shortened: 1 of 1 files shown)


markbackman:
I was able to push the linting fix to your branch. I've approved and will merge so this gets in today's release. Thanks again!

monster-anshu:
thanks @markbackman

Issue #3421, opened 2026-01-12
Rime TTS Punctuation Spacing Error


tfvjr:
### pipecat version

0.0.96+

### Python version

3.13.7

### Operating System

macOS 26.1

### Issue description

Same issue reported and fixed here, but for Rime not Deepgram.

https://github.com/pipecat-ai/pipecat/issues/3242
https://github.com/pipecat-ai/pipecat/pull/3345

### Reproduction steps

Typically occurs in phrases with more than one period.

### Expected behavior

Not pronounce punctuation like "dot".

### Actual behavior

Sent to TTS with "dot".

### Logs

```shell

```

markbackman:
Can you provide more information on how to reproduce? On `main`, I ran the same example I ran for Deepgram and cannot get the TTS to say "dot".

Here's that file:
[07q-interruptible-rime.py](https://github.com/user-attachments/files/24575212/07q-interruptible-rime.py)

Rime is configured for:
```
    tts = RimeTTSService(
        api_key=os.getenv("RIME_API_KEY", ""),
        voice_id="rex",
    )
```

which means the `mistv2` model is in use.

tfvjr:
Still working on being able to replicate consistently, but here are some of the logs from one of our faulty calls.

Master log shows: "2026-01-12 21:11:58.984 | INFO     | llm:llm_response_complete | text="Perfect, Teddy. I've got you down for PPF on your Toyota Camry. Someone from our team will give you a call soon. Thanks." duration_ms=975.466 text_length=120 transport_type="twilio"
2026-01-12 16:11:59.023 | DEBUG    | metrics.frame_processor_metrics:stop_ttfb_metrics:131 | RimeTTSService#0 TTFB: 0.400036096572876
2026-01-12 16:11:59.042 | DEBUG    | transports.base_output:_bot_started_speaking:604 | Bot started speaking
2026-01-12 16:11:59.107 | DEBUG    | rime.tts:_receive_messages:409 | Updated cumulative time to: 0.7778684807256236
2026-01-12 16:11:59.458 | DEBUG    | rime.tts:_receive_messages:409 | Updated cumulative time to: 3.2391836734693875
2026-01-12 16:11:59.634 | DEBUG    | rime.tts:_receive_messages:409 | Updated cumulative time to: 5.816598639455782
[...]

PR #3424, opened 2026-01-13
Add append_trailing_space to TTSService to prevent vocalizing trailin…

markbackman:
…g punctuation; update DeepgramTTSService and RimeTTSService to use the arg

Fixes https://github.com/pipecat-ai/pipecat/issues/3421

### DIFF
--- a/changelog/3424.changed.md
+++ b/changelog/3424.changed.md
@@ -1,0 +1,1 @@
+- `DeepgramTTSService` and `RimeTTSService` now set `append_trailing_space` to `True` to prevent punctuation (e.g., “dot”) from being pronounced.
--- a/src/pipecat/services/tts_service.py
+++ b/src/pipecat/services/tts_service.py
@@ -101,6 +101,9 @@ class TTSService(AIService):
         silence_time_s: float = 2.0,
         # if True, we will pause processing frames while we are receiving audio
         pause_frame_processing: bool = False,
+        # if True, append a trailing space to text before sending to TTS
+        # (helps prevent some TTS services from vocalizing trailing punctuation)
+        append_trailing_space: bool = False,
         # TTS output sample rate
         sample_rate: Optional[int] = None,
         # Text aggregator to aggregate incoming tokens and decide when to push to the TTS.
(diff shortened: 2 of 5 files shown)


changelog/3424.changed.md
```diff
@@ -0,0 +1 @@
+- `DeepgramTTSService` and `RimeTTSService` now use the base class `append_trailing_space` to add trailing spaces in their `run_tts()` methods in order to prevent pronouncing punctuation (e.g. "dot").
```
filipi87: I guess this "base class" part is a little confusing. Maybe something like this:
```suggestion
- `DeepgramTTSService` and `RimeTTSService` now set `append_trailing_space` to `True` to prevent punctuation (e.g., “dot”) from being pronounced.
```

tfvjr (approved):

src/pipecat/services/tts_service.py
```diff
@@ -273,6 +279,19 @@ def language_to_service_language(self, language: Language) -> Optional[str]:
...
+    def _prepare_text_for_tts(self, text: str) -> str:
+        """Prepare text for TTS by applying any necessary transformations.
+
+        Args:
+            text: The text to prepare.
+
+        Returns:
+            The prepared text with transformations applied.
+        """
+        if self._append_trailing_space and not text.endswith(" "):
```
mattieruth, line 291: in deepgram, you had purposely chosen to always add a space, no matter the ending. is there a reason you went back on that decision here?

src/pipecat/services/tts_service.py
```diff
@@ -273,6 +279,19 @@ def language_to_service_language(self, language: Language) -> Optional[str]:
         """
         return Language(language)
 
+    def _prepare_text_for_tts(self, text: str) -> str:
+        """Prepare text for TTS by applying any necessary transformations.
```
mattieruth: nit: i think this comment should make it clear that these transformations are service specific or built-in and not to be confused with the set of custom transformations someone can add. maybe something like:

```suggestion
        """Prepare text for TTS by applying any transformations required by the service.
```

src/pipecat/services/tts_service.py
markbackman: I hadn't thought of this case, and I'm not sure it will exist, given how the aggregator works. But, the issue is that some TTS services require a space between sentences. If there's already a space, we shouldn't add a second one. Think that makes sense?

markbackman:
> two minor things. but also: In passing, I think you mentioned that Fuse might also need this?
> 

I thought Hume might need it, but in my testing, I couldn't find a case where the punctuation was vocalized. If someone reports the issue, the fix is now easy to make.

src/pipecat/services/tts_service.py
markbackman: Makes sense. Fixed.

changelog/3424.changed.md
markbackman: Fixed.

PR #3425, opened 2026-01-13
Add Gemini 3 Flash-specific thinking levels

kompfner:


pyproject.toml
```diff
@@ -62,7 +62,7 @@ fal = [ "fal-client~=0.5.9" ]
 fireworks = []
 fish = [ "ormsgpack~=1.7.0", "pipecat-ai[websockets-base]" ]
 gladia = [ "pipecat-ai[websockets-base]" ]
-google = [ "google-cloud-speech>=2.33.0,<3", "google-cloud-texttospeech>=2.31.0,<3", "google-genai>=1.51.0,<2", "pipecat-ai[websockets-base]" ]
+google = [ "google-cloud-speech>=2.33.0,<3", "google-cloud-texttospeech>=2.31.0,<3", "google-genai>=1.57.0,<2", "pipecat-ai[websockets-base]" ]
```
kompfner, line 65: I had to. update `google-genai` because it does some validation itself, and the older version didn't recognize the new thinking levels.

filipi87:
@kompfner, I can see some tests failing. Is this related to any changes in this PR? Maybe the new dependency?

kompfner:
> @kompfner, I can see some tests failing. Is this related to any changes in this PR? Maybe the new dependency?

Hmm, seems related to `av`.

Just found this: https://github.com/PyAV-Org/PyAV/issues/2082.

They had to delete old wheels to free up space in their project recently.

Given that `uv.lock` has been pinned to `av` 14.4.0 for a while, my guess is these tests will fail for all new PRs.

kompfner:
OK looks like `aiortc` is the culprit pinning us to `av` 14.4.0. Updating to a newer version of `aiortc`, which will relax this constraint.

PR #3428, opened 2026-01-13
Fix TTS, realtime LLM services could return unknown for model_name

markbackman:

We had a trickle of community PRs finding issues:
- LLM: https://github.com/pipecat-ai/pipecat/pull/3422
- STT: https://github.com/pipecat-ai/pipecat/pull/3351

This applies the same fix for TTS, OpenAI Realtime, Gemini Live.

Updating changelogs to be consistent, too.

### DIFF
--- a/src/pipecat/utils/tracing/service_decorators.py
+++ b/src/pipecat/utils/tracing/service_decorators.py
@@ -186,7 +186,7 @@ def traced_tts(func: Optional[Callable] = None, *, name: Optional[str] = None) -
                     add_tts_span_attributes(
                         span=span,
                         service_name=service_class_name,
-                        model=getattr(self, "model_name", "unknown"),
+                        model=getattr(self, "model_name") or "unknown",
                         voice_id=getattr(self, "_voice_id", "unknown"),
                         text=text,
                         settings=getattr(self, "_settings", {}),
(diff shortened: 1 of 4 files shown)


src/pipecat/utils/tracing/service_decorators.py
```diff
@@ -585,8 +585,10 @@ async def wrapper(self, *args, **kwargs):
                 ) as current_span:
                     try:
                         # Base service attributes
-                        model_name = getattr(
-                            self, "model_name", getattr(self, "_model_name", "unknown")
+                        model_name = (
```
filipi87, line 588: Shouldn't we also check for this ?
`or params.get("model")`

filipi87: The same thing for the code below in case this makes sense.

filipi87 (approved):
I have added a comment to check whether we should also consider an extra condition, but I am already leaving my pre approval here.

src/pipecat/utils/tracing/service_decorators.py
markbackman: `params` is unique to `BaseOpenAILLMService`, so it only applies here.

PR #3430, opened 2026-01-13
Fix request_image_frame and usage

kompfner:

### DIFF
--- a/changelog/3430.fixed.md
+++ b/changelog/3430.fixed.md
@@ -1,0 +1,1 @@
+- Fixed `request_image_frame` (for backwards compatibility) and restored function-call–related fields in `UserImageRequestFrame` and `UserImageRawFrame`, preventing a case where adding a non-LLM message to the context could trigger duplicate LLM inferences (on image arrival and on function-call result), potentially causing an infinite inference loop.


src/pipecat/processors/aggregators/llm_response_universal.py
```diff
@@ -856,31 +866,24 @@ async def _handle_function_call_cancel(self, frame: FunctionCallCancelFrame):
...
-            format=frame.format,
-            size=frame.size,
-            image=frame.image,
-            text=frame.text,
-        )
+        image_appended = False
+
+        # Check if this image is a result of a function call if so, let's cache.
+        # TODO(aleix): The function call might have already been executed
+        # because FunctionCallResultFrame was just fater, in that case we just
```
kompfner: typo: "faster"

examples/foundational/14e-function-calling-google.py
```diff
@@ -209,10 +226,15 @@ async def on_client_connected(transport, client):
 
         await maybe_capture_participant_camera(transport, client)
 
-        global client_id
         client_id = get_transport_client_id(transport, client)
 
         # Kick off the conversation.
+        messages.append(
```
kompfner, line 231: I can't remember—we handle things gracefully when you have two system messages in a row at the start of the context, right? I think so...

Asking because we already have a `system_prompt` earlier in this file.

Oh wait...we also have a user message earlier in the context...

Either way, I think this message will just get converted to a user message under the hood, by the Gemini adapter.

But I think we can get rid of the earlier user message ("Say hello."), as it's no longer needed.

aconchillo: Ah, right. Removed `Say hello`.

src/pipecat/processors/aggregators/llm_response_universal.py
aconchillo: Done.

kompfner:
LGTM! I'm not allowed to approve this PR since I opened it, but I think you can @aconchillo!

PR #3431, opened 2026-01-13
Update GeminiLiveLLMService to push thought frames, update 26a for new transcript events

markbackman:

In porting the realtime LLM examples to use the new transcription events from the context aggregators, I found that the transcript showed both the thinking and spoken data. This means that the context includes that data as well, which is not desired.

This update handles the `part.thought` content and pushes `LLMThoughtStartFrame`, `LLMThoughtTextFrame`, and `LLMThoughtEndFrame`.

Also, 26a is updated to use the new transcription events.

Notes:

- The new Gemini Live models only have an audio+text mode; no text only mode. This means that you'll get these frames by default, unless the thinking budget is set to 0.
  - LLMFullResponseStartFrame
  - TTSTextFrame
  - LLMFullResponseEndFrame
  - LLMThoughtStartFrame
  - LLMThoughtTextFrame
  - LLMThoughtEndFrame

- If the thinking budget is set to 0, you'll get these frames, which matches the previous behavior:
  - LLMFullResponseStartFrame
  - TTSTextFrame
  - LLMFullResponseEndFrame

### DIFF
--- a/examples/foundational/26a-gemini-live-transcription.py
+++ b/examples/foundational/26a-gemini-live-transcription.py
@@ -12,13 +12,16 @@ from loguru import logger
 
 from pipecat.audio.vad.silero import SileroVADAnalyzer
 from pipecat.audio.vad.vad_analyzer import VADParams
-from pipecat.frames.frames import LLMRunFrame, TranscriptionMessage
+from pipecat.frames.frames import LLMRunFrame
 from pipecat.pipeline.pipeline import Pipeline
 from pipecat.pipeline.runner import PipelineRunner
 from pipecat.pipeline.task import PipelineParams, PipelineTask
 from pipecat.processors.aggregators.llm_context import LLMContext
-from pipecat.processors.aggregators.llm_response_universal import LLMContextAggregatorPair
-from pipecat.processors.transcript_processor import TranscriptProcessor
+from pipecat.processors.aggregators.llm_response_universal import (
+    AssistantTurnStoppedMessage,
+    LLMContextAggregatorPair,
+    UserTurnStoppedMessage,
+)
 from pipecat.runner.types import RunnerArguments
 from pipecat.runner.utils import create_transport
 from pipecat.services.google.gemini_live.llm import GeminiLiveLLMService
@@ -93,17 +96,16 @@ async def run_bot(transport: BaseTransport, runner_args: RunnerArguments):
     )
     context_aggregator = LLMContextAggregatorPair(context)
 
-    transcript = TranscriptProcessor()
+    user_aggregator = context_aggregator.user()
+    assistant_aggregator = context_aggregator.assistant()
 
     pipeline = Pipeline(
         [
             transport.input(),
-            context_aggregator.user(),
-            transcript.user(),
+            user_aggregator,
             llm,
             transport.output(),
-            transcript.assistant(),
-            context_aggregator.assistant(),
+            assistant_aggregator,
         ]
     )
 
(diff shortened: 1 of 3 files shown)


src/pipecat/services/google/gemini_live/llm.py
```diff
@@ -1455,10 +1458,19 @@ async def _handle_msg_model_turn(self, msg: LiveServerMessage):
...
-            self._search_result_buffer += text  # Also accumulate for grounding
-            frame = LLMTextFrame(text=text)
-            await self.push_frame(frame)
+            # Check if this is a thought
+            if part.thought:
+                # Gemini Live emits fully-formed thoughts rather than chunks,
+                # so bracket each thought in start/end frames
+                await self.push_frame(LLMThoughtStartFrame())
+                await self.push_frame(LLMThoughtTextFrame(text))
+                await self.push_frame(LLMThoughtEndFrame(signature=part.thought_signature))
```
markbackman: @kompfner I see the `LLMThoughtEndFrame` definition includes the ability to set the thought signature. I'm assuming this is acceptable.

I can see that the signature is always None, at least for now. I'm not sure if/when that will change. The Gemini Live API has been changing a lot and I expect that it will continue to.

examples/foundational/26a-gemini-live-transcription.py
```diff
@@ -127,14 +129,17 @@ async def on_client_disconnected(transport, client):
...
+    async def on_user_turn_stopped(aggregator, strategy, message: UserTurnStoppedMessage):
+        timestamp = f"[{message.timestamp}] " if message.timestamp else ""
+        line = f"{timestamp}user: {message.content}"
+        logger.info(f"Transcript: {line}")
+
+    @assistant_aggregator.event_handler("on_assistant_turn_stopped")
+    async def on_assistant_turn_stopped(aggregator, message: AssistantTurnStoppedMessage):
+        timestamp = f"[{message.timestamp}] " if message.timestamp else ""
+        line = f"{timestamp}assistant: {message.content}"
+        logger.info(f"Transcript: {line}")
```
aconchillo, line 142: Oh man, I thought I updated everything :disappointed:. Thanks!

aconchillo:
LGTM! But I will leave the `thought_signature` to @kompfner

src/pipecat/services/google/gemini_live/llm.py
kompfner: Gemini doesn't couple thought signatures to thought messages. This field is essentially only for Anthropic at the moment. I'd leave it blank.

src/pipecat/services/google/gemini_live/llm.py
```diff
@@ -1455,10 +1458,19 @@ async def _handle_msg_model_turn(self, msg: LiveServerMessage):
                 await self._set_bot_is_responding(True)
                 await self.push_frame(LLMFullResponseStartFrame())
 
-            self._bot_text_buffer += text
-            self._search_result_buffer += text  # Also accumulate for grounding
-            frame = LLMTextFrame(text=text)
-            await self.push_frame(frame)
+            # Check if this is a thought
```
kompfner, line 1461: Wait so prior to this PR we were treating thoughts as regular responses? Why haven't we always been hearing strange long inner monologue when talking to Gemini Live?

src/pipecat/services/google/gemini_live/llm.py
markbackman: Good to know. Updating.

src/pipecat/services/google/gemini_live/llm.py
markbackman: This is Gemini Live, so audio is produced by the model itself. The issue is that the thinking text was being added to the LLMTextFrame. This was revealed by the new transcription events which use the context to create the assistant transcription. The transcript included thoughts and the spoken text.

This change breaks them into two distinct groups:
- LLMThoughtTextFrame for thoughts
- TTSTextFrame for spoken text

(Once Google adds text output mode back, then LLMText will be for non-thinking output, but that's currently None and not pushed.)

kompfner: Ah right, yes, makes sense. So the thoughts were always polluting the transcript but somehow we didn't really pay close enough attention to notice?

markbackman: On 0.0.98, the TranscriptProcessor produces the assistant transcript from the TTSTextFrame. So, the thought parts weren't displayed. (They were included in the context though, which might have created problems.)

Starting in 0.0.99, the TranscriptProcessor is deprecated; we now emit transcription events from the context aggregator instances. For the assistant, the context itself is used to generate the assistant transcript. Due to that change, we found the issue.

PR #3432, opened 2026-01-13
Update/plugins metrics

dhruvladia-sarvam:
STT: Previously, metrics were being tracked between receiving consecutive server messages (START_SPEECH to data), which only measured local message processing time rather than actual latency. 
-> Now: ttfb is processing latency of END_SPEECH to first transcript received.

TTS: stop_ttfb_metrics() was being called on every audio chunk received, rather than only on the first audio chunk.
-> Now: Added _ttfb_recorded flag to ensure TTFB is only recorded once per utterance and now accurately measures from first text sent to first audio byte received.


src/pipecat/services/sarvam/tts.py
```diff
@@ -1,5 +1,5 @@
 #
-# Copyright (c) 2024–2026, Daily
+# Copyright (c) 2024–202, Daily
```
```suggestion
# Copyright (c) 2024–2026, Daily
```

src/pipecat/services/sarvam/stt.py
```diff
@@ -441,6 +441,7 @@ async def _handle_message(self, message):
                     await self.push_interruption_task_frame_and_wait()
 
                 elif signal == "END_SPEECH":
+                    await self.start_ttfb_metrics()
```
markbackman, line 444: No changes are needed to the STT service. Metrics are now handled by the base class. See https://github.com/pipecat-ai/pipecat/pull/3495 for details

src/pipecat/services/sarvam/stt.py
```diff
@@ -472,8 +473,6 @@ async def _handle_message(self, message):
                         )
                     )
 
-                await self.stop_processing_metrics()
```
markbackman, line 475: This is required. You need a `stop_processing_metrics()` after a start. I would leave this.

markbackman (commented):
No changes to STT are required. See my comments.

No changes are required for TTS. The call to `stop_ttfb_metrics()` has an early return if `self._start_ttfb_time = 0`. The pattern in SarvamTTSService matches that of other classes.

I'm going to close this PR as no changes are needed.

PR #3433, opened 2026-01-13
Update examples to use transcription events from context aggregators

markbackman:

- Added to Nova Sonic example: 40
- Realtime LLM examples updated to use transcription events from context aggregators: 19, 51
- Removed use, as it's not necessary: 19b*
- Replaced with an observer: 42

Also, I'm removing 28-transcription-processor.py since it's deprecated in 0.0.99.


aconchillo:
Amazing... Maybe I updated them in my dreams! LGTM! :pray:

PR #3441, opened 2026-01-14
Add example demonstrating dynamic interruption toggling

jamsea:
## Summary

This example shows how to disable interruptions while still transcribing user speech using the `enable_interruptions` parameter on user turn start strategies (introduced in 0.0.99).

## Key features demonstrated

- `VADUserTurnStartStrategy` and `TranscriptionUserTurnStartStrategy` with `enable_interruptions=False` to prevent bot interruption
- Dynamic toggling via `user_turn_controller.update_strategies()` at runtime
- Transcription continues in both modes, only interruption behavior changes

## Use case

When you want to capture what the user says during bot speech without interrupting the bot's response, then optionally re-enable interruptions later.

## Testing

Run the example and speak while the bot is talking. Watch the logs to see when interruptions toggle between enabled/disabled every 5 seconds.

### DIFF
--- a/examples/foundational/07a-non-interruptible.py
+++ b/examples/foundational/07a-non-interruptible.py
@@ -13,9 +13,10 @@ transcribing user speech. Every 5 seconds, the bot toggles between:
 This is useful when you want to capture what the user says during bot speech
 without interrupting the bot's response, and then re-enable interruptions later.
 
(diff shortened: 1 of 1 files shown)


markbackman:
Closing. Perhaps pipecat-examples is a better place.

PR #3442, opened 2026-01-14
feat: implement LLMMessagesAppendFrame for OpenAI Realtime services

sheen4n:
Implement the _handle_messages_append method for OpenAIRealtimeLLMService and OpenAIRealtimeBetaLLMService to support programmatic message injection and bot response triggering.

This enables use cases like:
- Making the bot speak first by injecting a prompt message
- Adding context messages to the conversation programmatically
- Triggering responses without audio input

The implementation:
- Converts messages from LLMMessagesAppendFrame to ConversationItem objects
- Uses "input_text" content type for user messages, "text" for others
- Sends each message via ConversationItemCreateEvent
- Triggers ResponseCreateEvent if frame.run_llm is True

### DIFF
--- a/src/pipecat/services/openai/realtime/llm.py
+++ b/src/pipecat/services/openai/realtime/llm.py
@@ -448,8 +448,56 @@ class OpenAIRealtimeLLMService(LLMService):
             # Send results for newly-completed function calls, if any.
             await self._process_completed_function_calls(send_new_results=True)
 
-    async def _handle_messages_append(self, frame):
-        logger.error("!!! NEED TO IMPLEMENT MESSAGES APPEND")
+    async def _handle_messages_append(self, frame: LLMMessagesAppendFrame):
+        """Handle appending messages to the conversation and optionally triggering a response.
+
+        Converts messages from the frame to OpenAI Realtime conversation items and sends
+        them to the API. If `frame.run_llm` is True, triggers a response generation.
+
+        Args:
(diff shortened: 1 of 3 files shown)


amandesai01:
Came here to literally raise same PR after I finished extending in my own codebase!

src/pipecat/services/openai/realtime/llm.py
```diff
@@ -448,8 +448,56 @@ async def _handle_context(self, context: LLMContext):
...
+            await self.send_client_event(events.ConversationItemCreateEvent(item=item))
+
+        # Trigger response if requested
+        if frame.run_llm:
+            await self.send_client_event(
+                events.ResponseCreateEvent(
+                    response=events.ResponseProperties(
+                        output_modalities=self._get_enabled_modalities()
+                    )
+                )
```
amandesai01, line 499: reuse `await self._create_response()`

src/pipecat/services/openai/realtime/llm.py
```diff
@@ -448,8 +448,56 @@ async def _handle_context(self, context: LLMContext):
...
+            else:
+                content_type = "text"
+
+            # Create conversation item
+            item = events.ConversationItem(
+                type="message",
+                role=role,
+                content=[events.ItemContent(type=content_type, text=content)],
+            )
+
```
amandesai01, line 488: Also add this line:
```
self._messages_added_manually[item.id] = True
```

PR #3449, opened 2026-01-14
fix: Record correct system_instruction in LLM spans for LLM services

kingster:
### Problem
The `@traced_llm` decorator in LLM services records the `system_instruction` attribute at span creation time, but this captures the stale `self._system_instruction` value rather than the actual system instruction that will be sent to the API.

When using dynamic system prompts (e.g., updating language context via `LLMContext.set_messages()`), the traced spans show the **old** system instruction while the LLM actually receives the **updated** one. This causes significant confusion during debugging, as the traces suggest the context update isn't working when it actually is.

### Root Cause
In `GoogleLLMService._process_context()` /  others, the system instruction from context is extracted and potentially updates `self._system_instruction` **after** the tracing span has already captured the attribute. The `@traced_llm` decorator reads `self._system_instruction` at decoration/span-start time, before `_stream_content_universal_context()` has a chance to update it from the context's messages.

**Timeline of events:**
1. `@traced_llm` decorator captures `self._system_instruction` (OLD value) → recorded in span
2. `_stream_content_universal_context()` calls adapter to extract params from context
3. Adapter extracts system_instruction from context messages (NEW value)
4. `self._system_instruction` is updated to NEW value
5. API call is made with NEW value ✅
6. Span shows OLD value ❌

### Impact
- Developers debugging dynamic prompt updates see incorrect data in traces
- Hours of debugging wasted investigating "why context updates aren't working" when they actually are
- Misleading observability data in production systems

### Solution
Update the tracing logic to capture the system_instruction **after** it has been resolved from the context, ensuring the traced value matches what is actually sent to the API.

### Testing (Done)
1. Create a processor that dynamically updates `LLMContext` with different system messages
2. Trigger LLM generation
3. Verify Jaeger/OpenTelemetry traces show the **updated** system instruction, not the initial one


markbackman (approved):
LGTM! This is a nice improvement 🙌 

Also, two small things:
1. Can you lint the code, either (`uv run scripts/fix-ruff.sh` or install the pre-commit hook `uv run pre-commit install`)
2. Add a changelog fragment: https://github.com/pipecat-ai/pipecat/blob/main/CONTRIBUTING.md#changelog-entries

With those two things, this will be ready to be merged.

kingster:
@markbackman Done, have linted and added the changelog.

markbackman:
A few things have changed since I reviewed. I've pushed some changes to your branch that take a different approach. Instead of going through the adapter (which has signature/key-name issues across providers), we read directly from `_settings.system_instruction`, which is now the preferred way to do things, with a context message fallback.

Also, I took the opportunity to align the tracing attribute names with the latest OpenTelemetry GenAI semantic conventions (`gen_ai.provider.name`, `gen_ai.system_instructions`, etc.).

GenAI OTEL spec: https://opentelemetry.io/docs/specs/semconv/gen-ai/gen-ai-spans/

PR #3454, opened 2026-01-15
fix to make on_user_turn_stop_timeout work with ExternalUserTurnStrat…

markbackman:
…egies

In testing 07c-interruptible-deepgram-flux.py, if you speak for a long time, you'll see this logged mid output:
```
LLMUserAggregator#0: User stopped speaking (strategy: None)
```

This is the `on_user_turn_stop_timeout` firing. The issue is that the timeout looks for speaking as measured only by VADUserStarted/StoppedSpeakingFrames. This fix adds logic to allow UserStarted/StoppedSpeakingFrames to block the timeout from firing as well.

### DIFF
--- a/src/pipecat/turns/user_turn_controller.py
+++ b/src/pipecat/turns/user_turn_controller.py
@@ -12,6 +12,8 @@ from typing import Optional, Type
 from pipecat.frames.frames import (
     Frame,
     TranscriptionFrame,
+    UserStartedSpeakingFrame,
+    UserStoppedSpeakingFrame,
     VADUserStartedSpeakingFrame,
     VADUserStoppedSpeakingFrame,
 )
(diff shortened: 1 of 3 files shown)


src/pipecat/turns/user_turn_controller.py
```diff
@@ -81,6 +83,7 @@ def __init__(
         self._task_manager: Optional[BaseTaskManager] = None
 
         self._vad_user_speaking = False
+        self._user_speaking = False
```
aconchillo, line 85: I think we can probably eliminate `self._vad_user_speaking` the idea is just to have a flag we don't really need two

markbackman: Good call!

aconchillo:
LGTM! Just added a small comment.

PR #3456, opened 2026-01-15
MinWordsUserTurnStartStrategy: avoid using interim transcriptions

aconchillo:

Using a mix of interim and final transcriptions caused incorrect behavior; using only interim transcriptions alone also produced undesired results.

This approach is a bit slower because we need to wait on final transcriptions but it's much more reliable.


aconchillo:
Closing in favor of #3462

PR #3457, opened 2026-01-15
Reduce Tool Result Context Size by Using UTF-8 for JSON Serialization

ahoshaiyan:
This PR changes tool result serialization to prevent UTF-8 code points from being escaped during serialization. This drastically reduces the context size when returning a response that contains languages other than English.

We have been running a monkey-patched version in production and it helped us improve the agent accuracy and control cost better.

Screenshot 2026-01-15 at 10 08 03 AM


smokyabdulrahman:
Hey @markbackman 👋 
We would appreciate a review here. 🙏

smokyabdulrahman:
We would appreciate your feedback @markbackman 🙏

markbackman (approved):
I pushed an update to this branch. There were a few services missing and the changelog needed to match the contributing guidelines. With this, we can go ahead and merge.

PR #3461, opened 2026-01-15
Allow WebsocketClientTransport to send custom headers

glennpow:

The `additional_headers` param has been added to allow `WebsocketClientTransport` to send custom headers, for things like authentication.

### DIFF
--- a/src/pipecat/transports/websocket/client.py
+++ b/src/pipecat/transports/websocket/client.py
@@ -20,7 +20,6 @@ from typing import Awaitable, Callable, Optional
 import websockets
 from loguru import logger
 from pydantic.main import BaseModel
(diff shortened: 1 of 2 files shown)


src/pipecat/transports/websocket/client.py
```diff
@@ -20,6 +20,7 @@
 import websockets
 from loguru import logger
 from pydantic.main import BaseModel
+from websockets import HeadersLike
```
markbackman: Remove?

glennpow: Ah yes.  I'd imported that as a type for `additional_headers`, but it didn't play nicely with Pydantic.  I'll remove.

Issue #3459, opened 2026-01-15
TTSSpeakFrame text is added to LLM context


Viking5274:
### pipecat version

  - pipecat-ai: 0.0.98

### Python version

3.13

### Operating System

Mac OS 15.7.3

### Issue description

TTSSpeak frame and tts_say from flow are added into LLM context

### Reproduction steps

1. Use pipecat flow
2. At 2nd or 3d node add tts_say or push manually TTSSpekFrame
3. this TTSSpeakFrame will be added into LLM context (If it's not first agent message)

### Expected behavior

TTSSpeakFrame text should not be added to LLM contexg

### Actual behavior

It's added

### Logs

```shell

```

markbackman:
TTSSpeakFrames are added to the context if it occurs between the `LLMFullResponseStartFrame` and `LLMFullResponseEndFrame`. This behavior hasn't changed.

I'm interested in making this controllable; changing this issue to an Enhancement request.

markbackman:
@Viking5274 which TTS service class are you using?

Viking5274:
Soniox

markbackman:
> Soniox

Soniox is the STT service. Which TTS service?

markbackman:
TTSSpeakFrame can now be configured to be added to the context (or not) but specifying append_to_context when creating the TTSSpeakFrame.

PR #3465, opened 2026-01-15
Update so that TTSSpeakFrame can be configured to append_to_context

markbackman:

Whether `TTSSpeakFrame` text should be added to the LLM context has been debated. Previously, text was only added to context if it occurred between `LLMFullResponseStartFrame` and `LLMFullResponseEndFrame`, which was limiting and non-obvious.

This change adds an `append_to_context` parameter to `TTSSpeakFrame` (defaults to `False`), giving developers explicit control over whether the spoken text should be added to the LLM context.

**Implementation details:**

- **`TTSSpeakFrame`**: Added `append_to_context: bool = False` parameter

- **`TTSService`**: Modified `_push_tts_frames()` to accept and pass through the `append_to_context` parameter. When `append_to_context` is explicitly set, it's applied to the generated `TTSTextFrame`s; otherwise, the `TTSTextFrame` uses its default (`True`).

- **`WordTTSService`**: This required a slightly awkward workaround. `WordTTSService` classes (like `CartesiaTTSService`) collect word/timestamp tuples asynchronously and emit `TTSTextFrame`s via a task handler (`_words_task_handler`). To propagate the `append_to_context` setting:
  - Override `_push_tts_frames()` to capture the `append_to_context` value into `self._current_append_to_context`
  - The task handler applies this stored value to each word's `TTSTextFrame`
  - **Future consideration**: In a future refactor, merging `TTSService` and `WordTTSService` into a single class hierarchy would make this more natural.

- **`LLMAssistantAggregator`**: Simplified `_handle_text()` to rely solely on the `append_to_context` parameter:
  - Removed the `self._started` check (which required text to be between `LLMFullResponseStartFrame` and `LLMFullResponseEndFrame`)
  - Now any `TextFrame` with `append_to_context=True` will be aggregated and added to context
  - This allows `TTSSpeakFrame`-generated text to be added to context regardless of LLM response boundaries
  - **Note**: This is a simplification that makes the aggregator more flexible - text frames now control their own context inclusion via the `append_to_context` flag

Addresses https://github.com/pipecat-ai/pipecat/issues/3459.

### DIFF
--- a/src/pipecat/frames/frames.py
+++ b/src/pipecat/frames/frames.py
@@ -341,6 +341,11 @@ class TextFrame(DataFrame):
 
     Parameters:
         text: The text content.
+        skip_tts: Whether this text should be skipped by the TTS service.
+        includes_inter_frame_spaces: Whether any necessary inter-frame (leading/trailing) spaces are already
+            included in the text.
+        append_to_context: Whether this text should be appended to the LLM context.
+            Defaults to True.
     """
 
     text: str
@@ -918,9 +923,12 @@ class TTSSpeakFrame(DataFrame):
 
     Parameters:
         text: The text to be spoken.
+        append_to_context: Whether this text should be appended to the LLM context.
+            Defaults to False.
     """
 
     text: str
+    append_to_context: bool = False
 
 
 @dataclass
--- a/src/pipecat/processors/aggregators/llm_response_universal.py
+++ b/src/pipecat/processors/aggregators/llm_response_universal.py
@@ -49,6 +49,7 @@ from pipecat.frames.frames import (
     StartFrame,
     TextFrame,
     TranscriptionFrame,
+    TranslationFrame,
     UserImageRawFrame,
     UserStartedSpeakingFrame,
     UserStoppedSpeakingFrame,
@@ -639,7 +640,6 @@ class LLMAssistantAggregator(LLMContextAggregator):
                     DeprecationWarning,
                 )
 
-        self._started = 0
         self._function_calls_in_progress: Dict[str, Optional[FunctionCallInProgressFrame]] = {}
         self._function_calls_image_results: Dict[str, UserImageRawFrame] = {}
         self._context_updated_tasks: Set[asyncio.Task] = set()
@@ -758,7 +758,6 @@ class LLMAssistantAggregator(LLMContextAggregator):
 
     async def _handle_interruptions(self, frame: InterruptionFrame):
         await self._trigger_assistant_turn_stopped()
-        self._started = 0
         await self.reset()
 
     async def _handle_function_calls_started(self, frame: FunctionCallsStartedFrame):
@@ -904,15 +903,17 @@ class LLMAssistantAggregator(LLMContextAggregator):
             )
 
     async def _handle_llm_start(self, _: LLMFullResponseStartFrame):
-        self._started += 1
         await self._trigger_assistant_turn_started()
 
     async def _handle_llm_end(self, _: LLMFullResponseEndFrame):
-        self._started -= 1
         await self._trigger_assistant_turn_stopped()
 
     async def _handle_text(self, frame: TextFrame):
-        if not self._started or not frame.append_to_context:
+        # Skip TextFrame types not intended to build the assistant context
+        if isinstance(frame, (TranscriptionFrame, TranslationFrame, InterimTranscriptionFrame)):
+            return
+
+        if not frame.append_to_context:
             return
 
         # Make sure we really have text (spaces count, too!)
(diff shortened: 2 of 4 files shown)


src/pipecat/frames/frames.py
```diff
@@ -341,6 +341,11 @@ class TextFrame(DataFrame):
 
     Parameters:
         text: The text content.
+        skip_tts: Whether this text should be skipped by the TTS service.
```
markbackman, line 344: Noticed these were missing, so I added them.

src/pipecat/services/tts_service.py
```diff
@@ -690,6 +701,20 @@ async def start_word_timestamps(self):
...
+        src_frame,
+        includes_inter_frame_spaces: Optional[bool] = False,
+        append_to_context: Optional[bool] = None,
+    ):
+        """Override to capture append_to_context for word-by-word frames."""
+        # Store the append_to_context value for use in _words_task_handler
+        # If append_to_context is explicitly specified, use it; otherwise default to True
+        if append_to_context is not None:
+            self._current_append_to_context = append_to_context
+        await super()._push_tts_frames(src_frame, includes_inter_frame_spaces, append_to_context)
```
aconchillo: I'm not sure we need this. Word timestamps are a consequence of the aggregated text in the base TTS service class and I believe we always add that to the context by default (see `_process_text_frame`). So, to make things simple, maybe we don't need this?

src/pipecat/processors/aggregators/llm_response_universal.py
```diff
@@ -904,15 +902,17 @@ async def _handle_assistant_image_frame(self, frame: AssistantImageRawFrame):
...
         await self._trigger_assistant_turn_started()
 
     async def _handle_llm_end(self, _: LLMFullResponseEndFrame):
-        self._started -= 1
         await self._trigger_assistant_turn_stopped()
 
     async def _handle_text(self, frame: TextFrame):
-        if not self._started or not frame.append_to_context:
+        # Skip transcription frames - those are handled by the user aggregator
+        if isinstance(frame, (TranscriptionFrame, InterimTranscriptionFrame)):
```
markbackman: The TranscriptionFrame is blocked by the `LLMUserAggregator`, but the `InterimTranscriptionFrame` is pushed down the pipeline. To be safe, I think we want to guard against both being added to the assistant aggregator.

src/pipecat/services/tts_service.py
```diff
@@ -207,6 +207,9 @@ def __init__(
 
         self._processing_text: bool = False
 
+        # Track append_to_context for the current TTS generation (used by WordTTSService subclasses)
+        self._current_append_to_context: Optional[bool] = None
```
markbackman, line 211: @aconchillo based on your feedback, I think we can sidestep overridding _push_tts_frame by just tracking the `_current_append_to_context` in the base class. Then, the WordTTSService can just use the status.

src/pipecat/services/tts_service.py
```diff
@@ -639,7 +652,15 @@ async def _push_tts_frames(
             # or transformations.
             frame = TTSTextFrame(text, aggregated_by=type)
             frame.includes_inter_frame_spaces = includes_inter_frame_spaces
+            # If append_to_context was explicitly specified (e.g., from TTSSpeakFrame),
+            # use that value; otherwise, use TTSTextFrame's default (True).
+            if self._current_append_to_context is not None:
```
markbackman, line 657: At this point in time, only the TTSSpeakFrame calls `_push_tts_frame()` with an `append_to_context` arg. Given that, it's the only one that will conditionally modify this setting.

This update applies to non-word based TTS services.

src/pipecat/services/tts_service.py
```diff
@@ -783,6 +805,10 @@ async def _words_task_handler(self):
                 # we can rely on the default includes_inter_frame_spaces=False
                 frame = TTSTextFrame(word, aggregated_by=AggregationType.WORD)
                 frame.pts = self._initial_word_timestamp + timestamp
+                # Apply the append_to_context setting from the original TTSSpeakFrame
+                # if it was explicitly set; otherwise rely on TTSTextFrame's default (True).
+                if self._current_append_to_context is not None:
+                    frame.append_to_context = self._current_append_to_context
```
markbackman, line 811: Now we can just read the state from the TTSService class.

src/pipecat/processors/aggregators/llm_response_universal.py
markbackman: Oh, and the TranslationFrame, for anyone using it.

src/pipecat/services/tts_service.py
aconchillo: Ooops. this might be more tricky actually. How do we know the words received correspond to the `TTSSpeakFrame`? I think we will need the context ID idea I was talking about, where everything pushed to the TTS service has an ID and we could then associate the append_to_context to that ID.

So, the current approach will only work for HTTP services and for services without context IDs where the TTS service pauses processing incoming frames while still waiting for audio.

aconchillo: In any case, we can go with this for now. I'll fix it with the TTS services work.

filipi87: Just for future reference, I believe I have addressed all these context id concerns in this PR:
- https://github.com/pipecat-ai/pipecat/pull/3584

markbackman:
Closing in favor of https://github.com/pipecat-ai/pipecat/pull/3584

PR #3466, opened 2026-01-15
Fix realtime (speech-to-speech) services' RTVI event compatibility

kompfner:


kompfner:
@markbackman this is the change we discussed on the phone. Good idea! Seems to work well.

I just hope there's not some unintended side-effect of now pushing `LLMTextFrame`s when we weren't before...for example, it might affect any custom processors folks may have downstream of these services that weren't expecting two sets of text frames.

kompfner:
There's a bigger question posed by these changes: if we _are_ OK with this pattern of letting `LLMTextFrame`s with `append_to_context=False` continuing downstream from the TTS service (or TTS equivalent, in this case), then perhaps we could consider _always_ letting `LLMTextFrame`s proceed past TTS services rather than swallowing them, but setting `append_to_context=False` on them as they go by.

I didn't want to rock that boat yet, as that would introduce a bigger likelihood of side-effects and unintentional breakage.

But it _would_ have the nice benefit of making frame output the same between realtime services and cascaded LLM + TTS services.

PR #3349, opened 2026-01-04
Add Camb.ai TTS integration with MARS models

eRuaro:
## Summary

- Add Camb.ai text-to-speech integration using MARS models
- Support for mars-flash (default, fastest), mars-pro, and mars-instruct models
- 140+ languages supported
- 24kHz audio output with HTTP streaming
- Custom instructions support for mars-instruct model

## Files Added

- `src/pipecat/services/camb/__init__.py` - Module exports
- `src/pipecat/services/camb/tts.py` - CambTTSService implementation
- `tests/test_camb_tts.py` - Unit tests
- `examples/foundational/07zb-interruptible-camb-local.py` - Local audio example
- `changelog/3349.added.md` - Changelog entry

## Usage

```python
from pipecat.services.camb.tts import CambTTSService

tts = CambTTSService(
    api_key=os.getenv("CAMB_API_KEY"),
    aiohttp_session=session,
    voice_id=147320,
    model="mars-flash",
)
```

## Test plan

- [x] Unit tests pass (`uv run pytest tests/test_camb_tts.py -v`)
- [x] Local audio example works with mic/speakers

## Notes

I work at Camb.ai and will help maintain this integration.

PR #3468, opened 2026-01-15
Clean up CambTTSService

markbackman:

Small clean up on CambTTSService after https://github.com/pipecat-ai/pipecat/pull/3349.

Also, found some additional copyright dates that were wrong.

### DIFF
--- a/src/pipecat/services/camb/tts.py
+++ b/src/pipecat/services/camb/tts.py
@@ -248,18 +248,10 @@ class CambTTSService(TTSService):
             frame: The start frame containing initialization parameters.
         """
         await super().start(frame)
+
         # Use model-specific sample rate if not explicitly specified
         if not self._init_sample_rate:
-            self._sample_rate = MODEL_SAMPLE_RATES.get(self._model_name, 22050)
-        self._settings["sample_rate"] = self._sample_rate
-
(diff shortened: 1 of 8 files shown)


src/pipecat/services/camb/tts.py
```diff
@@ -248,18 +248,10 @@ async def start(self, frame: StartFrame):
...
-        self._settings["sample_rate"] = self._sample_rate
-
-        # Warn if sample rate doesn't match model's supported rate
-        if self._sample_rate != MODEL_SAMPLE_RATES.get(self._model_name):
-            logger.warning(
-                f"Camb.ai's {self._model_name} model requires "
-                f"{MODEL_SAMPLE_RATES.get(self._model_name)}Hz sample rate. "
-                f"Current rate of {self._sample_rate}Hz may cause issues."
-            )
+            self._sample_rate = MODEL_SAMPLE_RATES.get(self.model_name, 22050)
```
markbackman, line 254: This seems like a good compromise since only a single sample rate can be supported per model. The audio will be up/down sampled as needed.

The warning isn't needed because the case never occurs.

aconchillo (approved):

aconchillo:
LGTM

PR #3471, opened 2026-01-16
Revert pydantic 2.12 extra type annotation

markbackman:

This annotation isn't necessary, so reverting.

Note: there's some type of issue between sphinx / autodoc / pydantic which doesn't yet understand or like the pydantic 2.12 extra. This means two docs pages aren't being generated:

- runner/daily
- daily/utils

I think this is OK for now. We'll have to revisit in the future.


aconchillo:
LGTM

PR #3472, opened 2026-01-16
Add whisker_setup.py setup file to .gitignore

markbackman:

I wanted to make pipecat-ai-whisker a dev dependency of pipecat-ai, but this would create a circular dependency :/

Instead, we'll have to just install pip install pipecat-ai-whisker.

This commit adds whisker_setup.py to .gitignore. I'm open to other names.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -52,3 +52,6 @@ docs/api/api
 
 # uv
 .python-version
+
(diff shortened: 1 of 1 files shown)


aconchillo:
That works. I use the same file name.

PR #3474, opened 2026-01-16
Fix Pylance reportOptionalMemberAccess in _handle_function_call_cancel

jamsea:
## Summary
Fix Pylance type error in `_handle_function_call_cancel` method of `LLMAssistantContextAggregator`.

## Changes
Extract the dictionary value into a local variable and check for `None` before accessing `cancel_on_interruption` attribute. This is needed because `_function_calls_in_progress` is typed as `Dict[str, Optional[FunctionCallInProgressFrame]]`, so values can be `None`.

## Before
```python
if self._function_calls_in_progress[frame.tool_call_id].cancel_on_interruption:
```

## After
```python
function_call = self._function_calls_in_progress[frame.tool_call_id]
if function_call and function_call.cancel_on_interruption:
```

### DIFF
--- a/src/pipecat/processors/aggregators/llm_response.py
+++ b/src/pipecat/processors/aggregators/llm_response.py
@@ -1024,10 +1024,8 @@ class LLMAssistantContextAggregator(LLMContextResponseAggregator):
         logger.debug(
             f"{self} FunctionCallCancelFrame: [{frame.function_name}:{frame.tool_call_id}]"
         )
-        if frame.tool_call_id not in self._function_calls_in_progress:
-            return
-
-        if self._function_calls_in_progress[frame.tool_call_id].cancel_on_interruption:
+        function_call = self._function_calls_in_progress.get(frame.tool_call_id)
+        if function_call and function_call.cancel_on_interruption:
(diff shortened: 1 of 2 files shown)


src/pipecat/processors/aggregators/llm_response.py
```diff
@@ -1027,7 +1027,8 @@ async def _handle_function_call_cancel(self, frame: FunctionCallCancelFrame):
         if frame.tool_call_id not in self._function_calls_in_progress:
             return
 
-        if self._function_calls_in_progress[frame.tool_call_id].cancel_on_interruption:
+        function_call = self._function_calls_in_progress[frame.tool_call_id]
+        if function_call and function_call.cancel_on_interruption:
```
aconchillo, line 1028: I think this should be:

```
function_call = self._function_calls_in_progress.get(frame.tool_call_id)
if function_call and function_call.cancel_on_interruption:
```

without get() you might get an exception.

markbackman: You're right. We also need this fix in `llm_response_universal.py`, too. I added it there.

aconchillo (approved):

PR #3483, opened 2026-01-16
throttle user speaking frame

aconchillo:

No need to send UserSpeakingFrame so frequently.


src/pipecat/transports/base_input.py
```diff
@@ -444,6 +450,14 @@ async def _audio_task_handler(self):
                     else:
                         await self.push_frame(VADUserStoppedSpeakingFrame())
 
+    async def _user_currently_speaking(self):
+        """Handle user speaking frame."""
+        await self.broadcast_frame(UserSpeakingFrame)
```
markbackman: Why broadcast here?

Shouldn't this be handled only inside the if condition?

aconchillo: mistake. updated.

Issue #3388, opened 2026-01-09
`AWSNovaSonicLLMService` Fails to Recover from "Invalid event bytes" due to Premature `push_error`


msmsriram:
### pipecat version

0.0.98

### Python version

3.13.3

### Operating System

windows

### Issue description

When using `AWSNovaSonicLLMService`, requesting long responses (e.g., `max_tokens=1024`) frequently causes a `smithy_aws_event_stream.events.InvalidEventBytesError` from the underlying AWS SDK.

While the service has built-in auto-recovery logic via `_wants_connection` and `reset_conversation()`, it currently fails to recover the pipeline because it calls `push_error()` **before** attempting recovery. This causes the pipeline to propagate the `ErrorFrame` and terminate/cancel operations before the recovery can successfully re-establish the connection.

### Reproduction steps

1. Initialize `AWSNovaSonicLLMService` with a high `max_tokens` value (e.g., 1024).
2. Prompt the model with a request that generates a long response.
3. Observe the pipeline crash with `InvalidEventBytesError`.

```python
import asyncio
import os
import logging
from dotenv import load_dotenv

from pipecat.pipeline.pipeline import Pipeline
from pipecat.pipeline.runner import PipelineRunner
from pipecat.pipeline.task import PipelineParams, PipelineTask
from pipecat.frames.frames import LLMMessagesFrame
from pipecat.services.aws.nova_sonic.llm import AWSNovaSonicLLMService
from pipecat.processors.aggregators.llm_context import LLMContext

# Configure logging
logging.basicConfig(level=logging.INFO)
logger = logging.getLogger(__name__)

load_dotenv()

async def main():
    if not os.getenv("AWS_ACCESS_KEY_ID") or not os.getenv("AWS_SECRET_ACCESS_KEY"):
        logger.error("AWS credentials missing.")
        return

    # Trigger issue with high max_tokens
    llm = AWSNovaSonicLLMService(
        access_key_id=os.getenv("AWS_ACCESS_KEY_ID"),
        secret_access_key=os.getenv("AWS_SECRET_ACCESS_KEY"),
        region=os.getenv("AWS_REGION", "us-east-1"),
        voice_id="tiffany",
        max_tokens=1024, 
    )

    context = LLMContext(
        messages=[
            {"role": "system", "content": "You are a helpful assistant."},
            {
                "role": "user", 
                "content": "Tell me a very long story about the history of the universe. Go into extreme detail."
            },
        ]
    )

[...]

PR #3486, opened 2026-01-17
Fix `AWSNovaSonicLLMService.reset_conversation()`

kompfner:
Still to do: test with AWS Nova 1 Sonic.

Fixes https://github.com/pipecat-ai/pipecat/issues/3388.

### DIFF
--- a/src/pipecat/services/aws/nova_sonic/llm.py
+++ b/src/pipecat/services/aws/nova_sonic/llm.py
@@ -533,20 +533,29 @@ class AWSNovaSonicLLMService(LLMService):
         if system_instruction:
             await self._send_text_event(text=system_instruction, role=Role.SYSTEM)
 
-        # Send conversation history
+        # Send conversation history (except for the last message if it's from the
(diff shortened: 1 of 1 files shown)


examples/foundational/20e-persistent-context-aws-nova-sonic.py
```diff
@@ -114,6 +114,15 @@ async def _reset():
                 #         "content": f"{AWSNovaSonicLLMService.AWAIT_TRIGGER_ASSISTANT_RESPONSE_INSTRUCTION}",
                 #     }
                 # )
+                # If the last message isn't from the user, add a message asking for a recap
+                if messages and messages[-1].get("role") != "user":
+                    print("[pk] adding recap message to loaded conversation")
```
filipi87: nit: I guess even inside the examples we are using `logger` instead of `print`.

kompfner: whoops, didn't meant to leave this debugging statement in. thanks for the catch!

Issue #3458, opened 2026-01-15
Notifying observers when the pipeline is muted


a6kme:
### pipecat version

0.0.99

### Question

In the now deprecated `STTMuteFilter` we used to send `STTMuteFrame` which the observers could listen to and understand that the pipeline is muted. It could ignore frames like `InterruptionFrame` when the STT is muted. But, now that muting of pipeline is handled in User Aggregators, is there any recommended way for observers to know whether the pipeline is currently muted or not? 

One example of where this could be useful is - in `TurnTrackingObserver`. The observer will be notified of `UserStartedSpeakingFrame` even if the pipeline is muted, and they will keep changing the turns.

markbackman:
Added new user mute events in PR #3490. Hope this helps!

PR #3490, opened 2026-01-17
Add on_user_mute_started and on_user_mute_stopped events

markbackman:

Addressing https://github.com/pipecat-ai/pipecat/issues/3458.

I'm opting to include the event for observability and not pass through the strategy. I think this makes sense to keep the logic simple.


aconchillo:
Makes sense. LGTM!

PR #3492, opened 2026-01-18
Remove unused imports

markbackman:

I took a two part approach:

1. First, I temporarily updated `[tool.ruff.lint]` in pyproject.toml to include a check for unused imports:
```
[tool.ruff.lint]
select = [
    "D", # Docstring rules
    "I", # Import rules
    "F401", # Unused imports
]
```

With that in place, I ran `uv run ruff check --fix .`, which automatically resolves unused imports.

2. `ruff check --fix .` only removes safe to remove exports. That leaves others to be cleaned up manually. See the second commit. I have two questions for @filipi87.

---

In general, I think we should not enable F401 (unused imports) since it creates friction during development. Instead, we should run it periodically to clean up.

### DIFF
--- a/src/pipecat/runner/run.py
+++ b/src/pipecat/runner/run.py
@@ -263,7 +263,7 @@ def _setup_webrtc_routes(
         """Handle WebRTC offer requests via SmallWebRTCRequestHandler."""
 
         # Prepare runner arguments with the callback to run your bot
-        async def webrtc_connection_callback(connection):
+        async def webrtc_connection_callback(connection: SmallWebRTCConnection):
             bot_module = _get_bot_module()
 
             runner_args = SmallWebRTCRunnerArguments(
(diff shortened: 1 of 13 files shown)


src/pipecat/runner/run.py
```diff
@@ -206,7 +206,6 @@ def _setup_webrtc_routes(
     try:
         from pipecat_ai_small_webrtc_prebuilt.frontend import SmallWebRTCPrebuiltUI
 
-        from pipecat.transports.smallwebrtc.connection import SmallWebRTCConnection
```
markbackman, line 209: @filipi87 just making sure these imports can be removed.

src/pipecat/transports/smallwebrtc/connection.py
```diff
@@ -23,7 +23,7 @@
 
 try:
     from aiortc import (
-        MediaStreamTrack,
+        MediaStreamTrack,  # noqa: F401
```
markbackman: @filipi87 is this used or can it be removed?

src/pipecat/runner/run.py
filipi87: Instead of removing it, maybe we should start using it here:
`async def webrtc_connection_callback(connection):`

We were not specifying the type, which is why it was removed. Instead, it should be written like this:
`async def webrtc_connection_callback(connection: SmallWebRTCConnection):`

src/pipecat/transports/smallwebrtc/connection.py
filipi87: Yes, we can remove it. 👍

filipi87 (approved):
Nice cleanup. 🚀

src/pipecat/runner/run.py
markbackman: Good idea. Making that update now.

PR #3493, opened 2026-01-18
fix: make SmallWebRTCConnection pc_id globally unique

### LINKED ISSUE #3485: SmallWebRTCConnection pc_id is not globally unique, can be reused across servers
### pipecat version

0.0.99

### Python version

3.12

### Operating System

All

### Issue description

The peer connection id set by SmallWebRTCConnection (here)[https://github.com/pipecat-ai/pipecat/blob/main/src/pipecat/transports/smallwebrtc/connection.py#L281] uses a process level incrementing counter which is nice for a human to look at, but can cause conflicts across multiple server processes.

This can be problematic when a peer attempts reconnection, connects to a difference process (e.g. to a websocket signaling server behind a load balancer), and then reuses a different pc due to this ID conflict.

This can trivially fixed by using a random pc_id instead.

### Reproduction steps

1. Have 2 webrtc servers running.
2. Client 1 connects to server 1, gets pc_id `SmallWebRTCConnection#1`.
3. Client 2 connects to server 2, gets pc_id `SmallWebRTCConnection#1`.
4. Shut down server 2.
5. Watch client 2 attempt reconnection with server 1 and reuse client 1's peer connection.

### Expected behavior

No improper reuse of peer connections.

### Actual behavior

Improper reuse of peer connections.

### Logs

```shell

```

omChauhanDev:

Fixes #3485 

### Problem
The `pc_id` was previously derived from `self.name` (e.g., `SmallWebRTCConnection#1`), which uses a process-local counter. In multi-process deployments behind a load balancer, different processes generate identical `pc_id` values, causing connection conflicts during client reconnection.

### Solution
Append a full UUIDv4 hex suffix to `pc_id` to guarantee global uniqueness:
- **Before:** `SmallWebRTCConnection#1`
- **After:** `SmallWebRTCConnection#1-550e8400e29b41d4a716446655440000`

Preserves a human-readable class name for debugging, while eliminating collision risks at scale.

### DIFF
--- a/src/pipecat/transports/smallwebrtc/connection.py
+++ b/src/pipecat/transports/smallwebrtc/connection.py
@@ -14,6 +14,7 @@ for real-time communication applications.
 import asyncio
 import json
 import time
+import uuid
 from typing import Any, List, Literal, Optional, Union
 
 from loguru import logger
@@ -278,7 +279,7 @@ class SmallWebRTCConnection(BaseObject):
 
         self._answer: Optional[RTCSessionDescription] = None
         self._pc = RTCPeerConnection(rtc_config)
-        self._pc_id = self.name
+        self._pc_id = f"{self.name}-{uuid.uuid4().hex}"
         self._setup_listeners()
         self._data_channel = None
         self._renegotiation_in_progress = False


markbackman:
Tagging @filipi87 to take a look.

PR #3494, opened 2026-01-18
fix: preserve UninterruptibleFrames in __reset_process_queue

### LINKED ISSUE #3478: UninterruptibleFrames not preserved in __reset_process_queue
### pipecat version

master

### Python version

3.12

### Operating System

linux

### Issue description

In the code here:
https://github.com/pipecat-ai/pipecat/blob/main/src/pipecat/processors/frame_processor.py#L952
 
__process_queue is an ordinary asyncio.Queue of tuples.  You can see the tuples are put here:
https://github.com/pipecat-ai/pipecat/blob/main/src/pipecat/processors/frame_processor.py#L1005

So then item is a tuple `(frame, direction, callback)`.  But the check is `isinstance(item, UninterruptibleFrame)`, which will always be False.  In this case, the logic won't do anything.

### Reproduction steps

We found this issue due to static analysis and believe it to be accurate, but aren't using this functionality

### Expected behavior

The check should check for a frame.

### Actual behavior

the check is impossible.

### Logs

```shell

```

omChauhanDev:

Fixes #3478 

#### Problem 
The `__reset_process_queue()` method was checking `isinstance(item, UninterruptibleFrame)` but item is actually a `tuple (frame, direction, callback)` - not the frame itself. This meant the check always returned False, so all queued frames got tossed during interruptions, including the ones that should've been kept.

#### Solution
Extracted frame from the tuple before checking.

### DIFF
--- a/src/pipecat/processors/frame_processor.py
+++ b/src/pipecat/processors/frame_processor.py
@@ -950,7 +950,8 @@ class FrameProcessor(BaseObject):
         # Process current queue and keep UninterruptibleFrame frames.
         while not self.__process_queue.empty():
             item = self.__process_queue.get_nowait()
-            if isinstance(item, UninterruptibleFrame):
+            frame = item[0]
+            if isinstance(frame, UninterruptibleFrame):
                 new_queue.put_nowait(item)
             self.__process_queue.task_done()
 


markbackman:
@aconchillo to take a look.

aconchillo:
LGTM! Thank you!

aconchillo (approved):

PR #3495, opened 2026-01-19
Add TTFB metrics for STT services

### LINKED ISSUE #1933: Add usage metrics for STT services
### Problem Statement

We can't track the usage of STT services in the same way I can for LLM and TTS services. This makes it challenging to monitor resource consumption, optimize performance, and manage costs effectively.

### Proposed Solution

Extend the current usage metrics functionality to include STT services, similar to what is already available for LLM and TTS services. This could involve logging the duration, number of calls, and source (e.g., service used, model name) for each STT request.

### Alternative Solutions

_No response_

### Additional Context

See [Pipecat Usage Metrics Guide](https://docs.pipecat.ai/guides/features/metrics#enabling-llm/tts-usage-metrics): it would be helpful to know if there's any plan to expand support to STT.

- [x] No, I'm just suggesting

markbackman:

Adds Time To First Byte (TTFB) measurement for all STT services, measuring the time from when the user stops speaking to when the final transcription is received.

## How it works

Since STT services receive continuous audio input, we can't measure response time from a discrete request. Instead, we measure the time between when a user finishes speaking and when the corresponding transcription arrives.

The calculation uses the following signals:
- `VADUserStartedSpeakingFrame`: Indicates when a user started speaking
- `VADUserStoppedSpeakingFrame`: Indicates when a user stopped speaking
- `VADParams.stop_secs`: The VAD silence duration, which allows us to calculate the actual speech end time (`_speech_end_time = VADUserStoppedSpeakingFrame timestamp - VADParams.stop_secs`)
- `TranscriptionFrame`: The final transcription; TTFB is calculated as `TranscriptionFrame receipt time - _speech_end_time`

## Why we measure from the _last_ TranscriptionFrame

For LLM and TTS services, there's a discrete input with a corresponding output, making TTFB straightforward to measure. Streaming STT services work differently; user audio is continuously streamed to the STT provider:

stt-ttfb-diagram

This continuous stream can generate multiple `TranscriptionFrame`s. When there's more than one, we have no way to know when the STT service finalized the audio to produce the first transcription, making that timing unmeasurable.

Instead, we rely on the *last* `TranscriptionFrame` received. We can use this final transcript along with the VAD's stop-speaking event as two points in time to calculate a meaningful TTFB value. This is why determining which `TranscriptionFrame` is the final one for an utterance is critical.

## Determining the final transcript

A key challenge is knowing which `TranscriptionFrame` is the *final* one for an utterance. We use two mechanisms:

### 1. `TranscriptionFrame.finalized` field

A new `finalized: bool` field on `TranscriptionFrame` (default `False`) explicitly marks transcripts as final, as it's the result of a finalized / commit / or flush action taken by the STT class (or is inherent in the service, in some cases):

| Service | Approach |
|---------|----------|
| **Deepgram** | Calls `finalize()` on VAD stop; checks `from_finalize` in response |
| **ElevenLabs Realtime** | Uses manual commit strategy; marks committed transcripts |
| **Cartesia** | Sends finalize command on VAD stop |
| **Sarvam** | Sends flush command on VAD stop |
| **Soniox** | Always finalized (end token = complete transcript) |
| **Flux** | Always finalized (`EndOfTurn` = complete transcript) |
| **SegmentedSTTService** | Always finalized (processes complete audio segments) |

When a `TranscriptionFrame` arrives with `finalized=True`, TTFB is reported immediately and the timeout is cancelled.

### 2. Timeout fallback

For services that don't support finalization or when the signal isn't available, a configurable timeout (`stt_ttfb_timeout`, default 2 seconds) starts when `VADUserStoppedSpeakingFrame` is received. When the timeout fires, TTFB is calculated using the most recent `TranscriptionFrame` timestamp.

## Handling transcript timing

The `TranscriptionFrame` can arrive either before or after the `VADUserStoppedSpeakingFrame`:

1. **Transcript arrives after `VADUserStoppedSpeakingFrame`:** We calculate TTFB as `TranscriptionFrame receipt time - _speech_end_time`. If positive, we report it. If the transcript is marked `finalized=True`, we report immediately; otherwise we wait for the timeout.

2. **Transcript arrives before `VADUserStoppedSpeakingFrame`:** We store the timestamp of the most recent `TranscriptionFrame`. When the timeout fires (or a finalized transcript arrives), we use the stored timestamp to calculate TTFB.

## Other considerations

- A local VAD is required to calculate STT TTFB. If no VAD is configured or no VAD events are received, no TTFB value will be reported.
- We reset tracking state on `VADUserStartedSpeakingFrame` (new utterance) and `InterruptionFrame` (cancelled utterance) to prevent stale values and cancel pending timeouts.
- Negative TTFB values (transcript arrived before calculated speech end) are discarded.
- The timeout is cancelled when a new utterance starts, ensuring we don't report stale TTFB values for interrupted speech.

Fixes #1933


kompfner:
> For services that don't support finalization or when the signal isn't available

Without having read the code yet: I'm guessing we assume that when the signal _is_ available, we expect the `finalized` frame to arrive before the timeout anyway, so we don't have to bake in any "smarts" about which STT service is in used and whether it has this signal.

If so: that'd make sense to me 👍

kompfner:
> A local VAD is required to calculate STT TTFB

Hmm, that doesn't feel ideal, especially since services like Flux are almost _expected_ to do the VAD work themselves and not require a local VAD...

(I count 4 services that can emit their own `UserStoppedSpeakingFrame` rather than relying on a local VAD to assist with end-of-turn detection: Deepgram (vanilla), Deepgram Flux, Gladia, and Speechmatics.

Is there maybe a signal other than VAD stop that we could use to mean "we've just sent the last request related to an utterance to the STT service?"

Maybe there's no good way to do this, but just wanted to raise this question.

changelog/3495.changed.md
```diff
@@ -0,0 +1 @@
+- Improved the Time To First Byte (TTFB) measurement for STT services, measuring the time from when the user stops speaking to when the final transcription is received. Added `finalized` field to `TranscriptionFrame` to indicate when a transcript is the final result for an utterance.
```
kompfner: Maybe a note that we know it's not _technically_ TTFB, but it's getting at what we really care about when we look at TTFB: the delay between sending data to the server and getting a response.

src/pipecat/services/assemblyai/stt.py
```diff
@@ -161,7 +161,7 @@ async def process_frame(self, frame: Frame, direction: FrameDirection):
         """
         await super().process_frame(frame, direction)
         if isinstance(frame, VADUserStartedSpeakingFrame):
-            await self.start_ttfb_metrics()
+            pass  # TTFB handled by base class
```
kompfner: Just curious: any reason you left a comment here? Looks like in the other STT services you just don't mention the base class magic

markbackman:
> > For services that don't support finalization or when the signal isn't available
> 
> Without having read the code yet: I'm guessing we assume that when the signal _is_ available, we expect the `finalized` frame to arrive before the timeout anyway, so we don't have to bake in any "smarts" about which STT service is in used and whether it has this signal.
> 
> If so: that'd make sense to me 👍

@kompfner, yes that's right. If we get a TranscriptionFrame with `finalized=True`, then the MetricsFrame is generate right away.

src/pipecat/services/stt_service.py
```diff
@@ -61,6 +68,8 @@ def __init__(
         audio_passthrough=True,
         # STT input sample rate
         sample_rate: Optional[int] = None,
+        # STT TTFB timeout - time to wait after VAD stop before reporting TTFB
```
kompfner, line 71: Again might be worth leaving a note somewhere in code (not saying it necessarily needs to be here) that what we're calling "TTFB" is not technically TTFB, but rather the delay between the user stopping speaking and receiving the last transcript, which gets at the "spirit" of TTFB.

kompfner: (This is just to avoid future reader confusion)

markbackman:
> Hmm, that doesn't feel ideal, especially since services like Flux are almost expected to do the VAD work themselves and not require a local VAD...

In doing this work, I came to realize how important two things are:
1. A local VAD configured with a very short `stop_secs` value. In every case I've tested, this creates a faster TranscriptionFrame by 100-300ms, in most cases.
2. Calling finalize/flush/commit when the audio is done. This relates to 1 but helps with the latency as well.

Given that, I think the default for all services (except for Flux) should be to use a local VAD and finalize where possible.

Flux is a different beast. I'm not sure what we can do there. There is no way to provide a good TTFB value. I've removed it from my benchmarking. It's its own blackbox, so there's no good way to get a TTFB value from it.

src/pipecat/services/assemblyai/stt.py
markbackman: No reason. I should remove it for consistency.

src/pipecat/services/stt_service.py
```diff
@@ -144,6 +163,11 @@ async def start(self, frame: StartFrame):
         self._sample_rate = self._init_sample_rate or frame.audio_in_sample_rate
         self._tracing_enabled = frame.enable_tracing
 
+    async def cleanup(self):
```
kompfner, line 210: Huh, was this just missing before?

src/pipecat/services/stt_service.py
```diff
@@ -204,14 +228,157 @@ async def process_frame(self, frame: Frame, direction: FrameDirection):
...
+        Args:
+            frame: The VAD user stopped speaking frame.
+        """
+        self._user_speaking = False
+
+        # Skip TTFB measurement if we don't have VAD params
+        if self._vad_stop_secs is None:
+            return
+
+        # Calculate the actual speech end time (current time minus VAD stop delay)
```
kompfner: Maybe mention that this is an approximation of the "request send time" of the last bit of user audio, to correlate to the eventual transcription response.

src/pipecat/services/stt_service.py
kompfner: Oh maybe there was just nothing that needed cancelling before...

kompfner (approved):

src/pipecat/services/stt_service.py
markbackman: It was, but it's there through inheritance:

STTService -> AIService -> FrameProcessor

The `cleanup()` method is in the FrameProcessor class. We had no reason to override it before. With the timeout requiring a task, now we do.

changelog/3495.changed.md
markbackman: I'll add a note to the changelog to clarify this.

src/pipecat/services/stt_service.py
markbackman: Adding a clarifying comment to the docstring.

src/pipecat/services/stt_service.py
markbackman: Done.

src/pipecat/services/sarvam/stt.py
```diff
@@ -411,16 +439,18 @@ async def _handle_message(self, message):
                 logger.debug(f"VAD Signal: {signal}, Occurred at: {timestamp}")
 
                 if signal == "START_SPEECH":
-                    await self.start_metrics()
+                    await self._start_metrics()
                     logger.debug("User started speaking")
                     await self._call_event_handler("on_speech_started")
+                    await self.broadcast_frame(UserStartedSpeakingFrame)
+                    await self.push_interruption_task_frame_and_wait()
```
filipi87: Should we check here if we should or not interrupt ?

That means, we should add a new parameter to the class, like we did for the other STT services:
`should_interrupt: bool = True, `

markbackman: Good catch! Yes, this should be included. I just added it.

filipi87 (approved):
LGTM!

src/pipecat/services/cartesia/stt.py
```diff
@@ -222,10 +224,13 @@ async def process_frame(self, frame: Frame, direction: FrameDirection):
         await super().process_frame(frame, direction)
 
         if isinstance(frame, VADUserStartedSpeakingFrame):
-            await self.start_metrics()
+            # Reset finalize state for new utterance
+            self._finalize_pending = False
```
aconchillo: If we move this to the base class, then we could set this in the base `STTService` class in `super().process_frame()` would do it.

src/pipecat/services/cartesia/stt.py
```diff
@@ -172,6 +172,9 @@ def __init__(
         self._base_url = base_url or "api.cartesia.ai"
         self._receive_task = None
 
+        # Track when finalize has been sent to mark the next TranscriptionFrame
+        self._finalize_pending: bool = False
```
aconchillo: It feels we should do this in the base class since we are repeating it in many subclasses. Then, we could have a method to set it

```
self.set_final_transcription(True | False)
```

then, we could override `STTService.push_frame()` and automatically set `finalized`. Something like

```
async def push_frame(self, frame, direction):
    if isinstance(frame, TranscriptionFrame):
        frame.finalized = self._finalize_pending

````

File: src/pipecat/services/cartesia/stt.py [hunk 9]
```diff
@@ -222,10 +224,13 @@ async def process_frame(self, frame: Frame, direction: FrameDirection):
...
 
         if isinstance(frame, VADUserStartedSpeakingFrame):
-            await self.start_metrics()
+            # Reset finalize state for new utterance
+            self._finalize_pending = False
+            await self._start_metrics()
         elif isinstance(frame, VADUserStoppedSpeakingFrame):
             # Send finalize command to flush the transcription session
             if self._websocket and self._websocket.state is State.OPEN:
+                self._finalize_pending = True
```
aconchillo: This would also be in the base class. It doesn't matter if the websocket is closed or not.

src/pipecat/services/cartesia/stt.py
```diff
@@ -351,8 +355,10 @@ async def _on_transcript(self, data):
                         time_now_iso8601(),
                         language,
                         result=data,
+                        finalized=self._finalize_pending,
                     )
                 )
+                self._finalize_pending = False
```
aconchillo: And these two changes would go away.

src/pipecat/services/deepgram/flux/stt.py
```diff
@@ -659,13 +659,16 @@ async def _handle_end_of_turn(self, transcript: str, data: Dict[str, Any]):
...
+            # EndOfTurn means Flux has determined the turn is complete,
+            # so this TranscriptionFrame is always finalized
             await self.push_frame(
                 TranscriptionFrame(
                     transcript,
                     self._user_id,
                     time_now_iso8601(),
                     self._language,
                     result=data,
+                    finalized=True,
```
aconchillo, line 671: If we do the changes above we would just set the base class `set_finalize_pending` to True in the constructor and no need to set it here.

src/pipecat/services/soniox/stt.py
```diff
@@ -374,12 +374,15 @@ async def _receive_messages(self):
...
                 text = "".join(map(lambda token: token["text"], self._final_transcription_buffer))
+                # Soniox only pushes TranscriptionFrame when an end token is received,
+                # so every TranscriptionFrame is inherently finalized
                 await self.push_frame(
                     TranscriptionFrame(
                         text=text,
                         user_id=self._user_id,
                         timestamp=time_now_iso8601(),
                         result=self._final_transcription_buffer,
+                        finalized=True,
```
aconchillo, line 385: Same as with Deepgram Flux

src/pipecat/services/stt_service.py
```diff
@@ -204,14 +232,159 @@ async def process_frame(self, frame: Frame, direction: FrameDirection):
...
+    async def _handle_vad_user_started_speaking(self, frame: VADUserStartedSpeakingFrame):
+        """Handle VAD user started speaking frame to start tracking transcriptions.
+
+        Cancels any pending TTFB timeout, resets TTFB tracking state, and marks user as speaking.
+
+        Args:
+            frame: The VAD user started speaking frame.
+        """
+        await self._reset_stt_ttfb_state()
+        self._user_speaking = True
```
aconchillo, line 367: here's where we would handle the `_finalize_pending`

aconchillo (approved):

aconchillo:
LGTM! This is great!

PR #3496, opened 2026-01-19
fix: add type annotations to event_handler decorator

jamsea:
Add TypeVar annotations to event_handler decorator to fix pyright reportUnusedFunction warnings


markbackman:
We'll roll out pyright fixes at some point...

PR #3498, opened 2026-01-19
AzureTTSService 8khz workaround

markbackman:

When using AzureTTSService at 8kHz sample rate, word timestamps from consecutive sentences could become interleaved, causing words to appear out of order. For example:
```
'Why'
"don't"
'skeletons'
'fight'
'each'
'They'
'other?'
"don't"
'have'
'the'
'guts!'
```

**Root Cause**

At 8kHz sample rate, Azure's `audio_duration` from the completion callback doesn't match the `audio_offset` values from word boundary events. For example, the last word boundary reports an offset of 1.93s, but `audio_duration` reports only 1.35s. This causes the next sentence's word timestamps to overlap with the previous sentence.
This appears to be an Azure SDK issue specific to 8kHz; other sample rates work correctly.

**Solution**

- Track the maximum word boundary offset during synthesis
- At 8kHz: use max_word_offset for cumulative timing (workaround for Azure inconsistency)
- At other sample rates: use audio_duration (original behavior)

**Additional change**

- The base Azure TTS class's prosody `rate` was set to `1.05`. This was changed to `None`, so it can inherit the APIs default value.
- Bump `azure-cognitiveservices-speech` to 1.47.0.

**Related**

I filed this issue against Azure's speech SDK for this:
https://github.com/Azure-Samples/cognitive-services-speech-sdk/issues/2983


src/pipecat/services/azure/tts.py
```diff
@@ -627,6 +650,16 @@ async def run_tts(self, text: str) -> AsyncGenerator[Frame, None]:
                     )
                     yield frame
 
+                # Update cumulative offset for next sentence
+                # At 8kHz, Azure's audio_duration doesn't match word boundary offsets,
+                # so we use max_word_offset as a workaround. At other sample rates,
+                # audio_duration is accurate.
```
filipi87, line 656: At other sample rates, will audio_duration always match `self._current_sentence_max_word_offset` ?

If so, could we always use `self._current_sentence_max_word_offset` for all sample rates ? If this works consistently, we could remove `self._current_sentence_duration` to simplify the code.

filipi87 (approved):
One small suggestion to maybe simplify the code, but based on the explanation, the fixes make sense.

src/pipecat/services/azure/tts.py
markbackman: `max_word_offset` is when the last word starts, while `audio_duration` is when the audio ends. Using `max_word_offset` everywhere would cut off the last word's duration from the timeline, causing timing drift.
At 8kHz we use `max_word_offset` because Azure's `audio_duration` is buggy there (reports values less than the word offsets). At other sample rates, `audio_duration` is correct and more accurate.

I have an open issue against the Azure python SDK, so hopefully they resolve this issue and we can remove this workaround.

PR #3499, opened 2026-01-19
fix(livekit): prevent memory leak when video_in_enabled is False

### LINKED ISSUE #3116: Memory leak on Ubuntu with Pipecat 0.0.85
### pipecat version

0.0.85

### Python version

3.12.8

### Operating System

Ubuntu 24.04.3 LTS

### Use Case Description

I am running pipecat (for a year), with a pipeline of deepgram, openai, elevenlabs through livekit.

### Current Approach

```python
This issue occurs only when running in the pods on the cluster. 
I can run version 0.0.85 just fine from a macOS.

The memory leak occurs with 0.0.86 and 0.0.92. (Those are the only two other version I've checked.

I incremented from v0.0.80, where I had had it, up one at a time, and .81, .82, .83 and .84 are all fine.

With the exact same code, except for the change in pipecat version, the memory leak occurs.

I updated the deepgram, openai, elevenlabs, and livekit packages from where they were to the latest, and there's still the leak.
```

When I go from pipecat version 0.0.84 to 0.0.85 there is a memory issue that makes the instance increase its memory usage about 3GB/minute.

### Additional Context

**Pipeline Details**
Services:
- livekit
- deepgram
- openai
- elevenlabs
- krisp noise filter
- silero vad analyzer

**I wonder if anyone here has experienced anything similar, or know what could be happening**

lukepayyapilli:
## Summary

- [x] Only start `_process_video_stream` when `video_in_enabled=True` to prevent unbounded queue growth.
- [x] Added livekit to dev dependencies so tests always run in CI.
- [x] Added behavior-focused unit tests for the fix.

## Approach

As identified by @draftin1 in #3116, the root cause of the memory leak is a producer/consumer mismatch:

- `_process_video_stream` (producer) was started unconditionally when a video track was subscribed, putting frames into `_video_queue`.
- `_video_in_task_handler` (consumer) only runs when `video_in_enabled=True`.

When `video_in_enabled=False`, frames pile up in the queue with no consumer, causing unbounded memory growth (~3GB/min as reported).

The fix adds a check for `video_in_enabled` before starting `_process_video_stream`, aligning producer behavior with consumer behavior.

## Tests

- [x] `test_disabled_video_input_does_not_start_queue_producer` - Verifies that when video input is disabled, no producer fills the queue (preventing the memory leak). Also confirms that track metadata is still recorded and callbacks still fire.
- [x] `test_enabled_video_input_starts_queue_producer` - Verifies that when video input is enabled, the producer starts normally.

Fixes #3116


pyproject.toml
```diff
@@ -125,6 +125,7 @@ dev = [
     "build~=1.2.2",
     "coverage~=7.9.1",
     "grpcio-tools~=1.67.1",
+    "livekit~=1.0.13",
```
markbackman: Recommend to remove this and just install the optional project dep instead.
```suggestion
```

markbackman:
Tagging @aconchillo and @filipi87 to take a look.

pyproject.toml
lukepayyapilli: Thanks for the review @markbackman! I switched livekit to be an optional. I wasn't sure if we wanted to add tests for it or not so if you feel it's overkill I can also remove them.

aconchillo:
LGTM! Thank you @lukepayyapilli ! Can we add a changelog file?

aconchillo:
Merging, so we can include it in today's release. I'll add the changelog file.

lukepayyapilli:
> LGTM! Thank you @lukepayyapilli ! Can we add a changelog file?

> Merging, so we can include it in today's release. I'll add the changelog file.

Thank you @aconchillo and my apologies - will add this going forward for future contributions.

PR #3500, opened 2026-01-19
Feature/genesys serializer

ssillerom:
# New Serializer: Genesys AudioHook WebSocket Protocol

This PR introduces a new serializer for the **Genesys AudioHook WebSocket protocol**, enabling bidirectional audio streaming between Pipecat pipelines and Genesys Cloud contact center.

The serializer handles audio streaming in **PCMU (μ-law)** and **L16** formats at **8kHz**, supporting mono (external/internal) and stereo channels, while managing session lifecycle, interruption events, and control messages.

## 🚀 Motivation and Context
Genesys Cloud is a widely used contact center platform, and the AudioHook protocol allows connecting custom voice bots to real-time calls.

* **Goal:** Facilitates the creation of advanced interactive voice response (IVR) applications (chatbots with ASR/TTS).
* **Integration:** Direct integration with the Pipecat ecosystem.
* **Benefit:** Reduces manual implementation complexity and minimizes error risks for developers.

---

### New File: `genesys.py`
* **Class:** Implements `GenesysAudioHookSerializer` (inherits from `FrameSerializer`).
* **Control Messages:** Supports `open`, `close`, `ping`, `pause`, `update`, `error`, `dtmf`, etc.
* **Audio Conversion:** Handles PCM to μ-law (and vice versa) using **SOXR resamplers**.
* **Events:** Includes custom events like barge-in, resume, and interruptions.
* **Configurability:** Support for multiple audio channels (external, internal, or both).
* **Frame Integration:** Works with `AudioRawFrame`, `InputAudioRawFrame`, `InputDTMFFrame`, etc.

### Dependencies
* Uses `SOXRStreamAudioResampler`, `ulaw_to_pcm`, `pcm_to_ulaw`.
* Requires `pydantic` for configuration and `loguru` for logging.

---

## 📈 Project Impact
* **Compatibility:** No breaking changes; modular addition.
* **Usage:** Can be imported into WebSocket transports. *Usage example included in docstrings.*
* **Performance:** Optimized for low-latency real-time streaming.
* **Documentation:** Includes references to the official Genesys protocol.

---

## 🔗 References
* [Official Genesys AudioHook Documentation](https://developer.genesys.cloud/devapps/audiohook/)

---

## ✅ Checklist
- [x] Code tested and functional.
- [x] Documentation updated.
- [x] No new critical dependencies.
- [x] Compatible with recent Pipecat versions.

> This PR enhances Pipecat's interoperability with contact center platforms, expanding its utility in enterprise scenarios. **Please review and comment!** 🚀


ssillerom:
im going to add during this weekend unit testing, mocking events from genesys, docs and a bot.py example with FastAPIWebsocketTransport

src/pipecat/serializers/genesys.py
```diff
@@ -0,0 +1,945 @@
...
+        media_format: AudioHookMediaFormat = AudioHookMediaFormat.PCMU
+        process_external: bool = True
+        process_internal: bool = False
+        enable_interruption_events: bool = True
+
+    def __init__(
+        self,
+        session_id: Optional[str] = None,
+        params: Optional[InputParams] = None,
+        send_message_callback: Optional[Callable[[str], Awaitable[None]]] = None,
```
aconchillo: What is the purpose of `send_message_callback`? Shouldn't pushing an output transport frame be enough?

src/pipecat/serializers/genesys.py
```diff
@@ -0,0 +1,945 @@
...
+                    logger.debug("📡 Mono mode: internal channel")
+        
+        # Log participant info for debugging
+        ani = self._participant.get("ani", "unknown") if self._participant else "unknown"
+        logger.info(
+            f"AudioHook open request: session={self._session_id}, "
+            f"conversation={self._conversation_id}, ani={ani}"
+        )
+        
+        # Note: Application should call create_opened_response() to respond
```
aconchillo: how does the application know when to call `create_opened_response()`? i'm thinking this could be an event, so you would do:

```
        serializer = GenesysAudioHookSerializer(
            session_id="abc-123",
            params=GenesysAudioHookSerializer.InputParams(
                channel=AudioHookChannel.BOTH,
            )
        )
        
        @serializer.event_handler("on_open")
        async def on_open(serializer, open_message: AudioHookOpenMessage):
           ....
```

aconchillo: So, in here, you would:

```
await self._call_event_handler("on_open", open_message)
```

aconchillo: This would need a `class AudioHookOpenMessage(BaseModel)` with all the fields coming from the open messsage.

src/pipecat/serializers/genesys.py
```diff
@@ -0,0 +1,945 @@
...
+        # Send closed response via callback if available
+        if self._send_message_callback:
+            try:
+                closed_response = self.create_closed_response()
+                await self._send_message_callback(closed_response)
+            except Exception as e:
+                logger.error(f"Failed to send closed response: {e}")
+        
+        # Return EndFrame to close the pipeline and WebSocket
+        return EndFrame()
```
aconchillo: Return an `OutputTransportMessageUrgentFrame` instead with the close response.

src/pipecat/serializers/genesys.py
aconchillo: As we discussed, no need for `send_message_callback` just use `OutputTransportMessageUrgentFrame`.

src/pipecat/serializers/genesys.py
```diff
@@ -0,0 +1,945 @@
...
+        sample_rate: Optional[int] = None
+        channel: AudioHookChannel = AudioHookChannel.EXTERNAL
+        media_format: AudioHookMediaFormat = AudioHookMediaFormat.PCMU
+        process_external: bool = True
+        process_internal: bool = False
+        enable_interruption_events: bool = True
+
+    def __init__(
+        self,
+        session_id: Optional[str] = None,
```
aconchillo: No need to pass the `session_id` we get it from the open message and we can save it internally.

src/pipecat/serializers/genesys.py
```diff
@@ -0,0 +1,945 @@
...
+        self._media_info: Optional[List[Dict[str, Any]]] = None
+        self._input_variables: Optional[Dict[str, Any]] = None  # Custom input from Genesys
+
+    def set_send_message_callback(self, callback: Callable[[str], Awaitable[None]]):
+        """Set the callback for sending urgent messages directly.
+        
+        Args:
+            callback: An async function that takes a string message and sends it.
+        """
+        self._send_message_callback = callback
```
aconchillo: Remove.

src/pipecat/serializers/genesys.py
```diff
@@ -0,0 +1,945 @@
...
+            parameters={
+                "entities": [
+                    {"type": "barge_in", "data": {}}
+                ]
+            },
+        )
+        
+        logger.debug("🔇 Barge-in event sent to Genesys")
+        
+        return json.dumps(msg)
```
aconchillo: Implement barge in when receiving `InterruptionFrame` (note that `StartInterruptionFrame` is deprecated).

src/pipecat/serializers/genesys.py
ssillerom: done

src/pipecat/serializers/genesys.py
ssillerom: done :)

src/pipecat/serializers/genesys.py
ssillerom: going to remove it when commiting event handler "on open"

src/pipecat/serializers/genesys.py
ssillerom: done

src/pipecat/serializers/genesys.py
ssillerom: done :)

src/pipecat/serializers/genesys.py
ssillerom: WIP

src/pipecat/serializers/genesys.py
```diff
@@ -0,0 +1,814 @@
...
+            msg["parameters"] = parameters
+            
+        return msg
+
+    def create_opened_response(
+        self,
+        start_paused: bool = False,
+        supported_languages: Optional[List[str]] = None,
+        selected_language: Optional[str] = None,
+    ) -> str:
```
aconchillo: I think it will be better for the `create_*` functions to return a `Dict[str, Any]`, this way you don't have to do `json.dumps()` and then `json.loads()` again to add use it in the `OutputTransportMessageUrgentFrame`.

aconchillo: That is, in general, always use `json.dumps()` at the very end when you are sure you need the string. Before, just pass the data structure around (dictionary in this case).

aconchillo:
This looks great! Just added one final comment. And also missing the use of the event as you mentioned.

ssillerom:
aleix added all things that you told me during this morning!

aconchillo (approved):

aconchillo:
> aleix added all things that you told me during this morning!

LGTM! Thank you @ssillerom !

aconchillo:
@ssillerom A couple of last things. Do you mind running `ruff check --fix` to fix formatting?

Also, it's probably a good idea to register events for the main protocol messages: `on_open`, `on_close`, `on_pause`, `on_update` and `on_error`.

Should be as simple as:

At the end of the constructor:

```
self._register_event_handler("on_open")
...
```

And then before the return of each `_handle_xxxxx`:

```
await self._call_event_handler("on_open", message)
```

where `message` is the full dictionary you receive from Genesys.

For example:

```
    async def _handle_error(self, message: Dict[str, Any]) -> Frame | None:
        """Handle an 'error' message from Genesys.
        
        Args:
            message: The error message.
            
        Returns:
            None.
        """
        params = message.get("parameters", {})
        code = params.get("code", 0)
        error_msg = params.get("message", "Unknown error")
        
        logger.error(f"AudioHook error from Genesys: {code} - {error_msg}")
        
        await self._call_event_handler("on_error", message).    <---------- ADD THIS
        
        return None
```

I think this will give developers notifications and more control of when these things happen with any info coming from Genesys.

src/pipecat/serializers/genesys.py
```diff
@@ -0,0 +1,956 @@
...
+
+        Args:
+            frame: The Pipecat frame to serialize.
+
+        Returns:
+            Serialized data as string (JSON) or bytes (audio), or None if
+            the frame type is not handled or session is not open.
+        """
+        if isinstance(frame, (EndFrame, CancelFrame)):
+            return json.dumps(self.create_disconnect_message(reason="completed"))
```
aconchillo: shouldn't we use `self._output_variables` here? That is, the developer will close the connection with `EndFrame` and `CancelFrame`, so they might call `self.set_output_variables(...)` before that.

So:
```
            return json.dumps(self.create_disconnect_message(output_variables=self._output_variables, reason="completed"))
```

PR #3503, opened 2026-01-20
AIService: handle StartFrame/EndFrame/CancelFrame exceptions

aconchillo:

If `AIService` subclasses implement start()/stop()/cancel() and exceptions are not handled, execution will not continue and therefore the originator frames will not be pushed. This would cause the pipeline to not be started (i.e. StartFrame would not be pushed downstream) or stopped properly.


aconchillo:
> I think these changes make sense, but I guess we could still have an issue where the originator frames are not pushed, because we are handling the exception here but not pushing the frame. So it is still up to the subclasses to push that frame, right ?
> 
> In any case, I think this is already an improvement.

Yes, subclasses are the ones pushing the frame. It usually looks like this:

```
await super().process_frame(frame, direction)  <----- exception was thrown here.

    await self.push_frame(frame)
    await self._start(frame)
```

PR #3506, opened 2026-01-20
feat: Add LiveKit Agents Worker Support

### LINKED ISSUE #3505: Add LiveKitWorkerRunner for livekit-agents compatibility
### Problem Statement

Right now, Pipecat works great as a client - you connect to a room with a URL/token, run your pipeline, and you're good to go. 

But if you want to deploy to **LiveKit Cloud's Agent hosting** (or self-hosted livekit-agents), you hit a wall.  LiveKit Agents expects a "Worker" that listens for room assignments rather than a script that dials in like a participant.  

There's currently no native way to wrap a Pipecat pipeline inside the `livekit-agents` worker pattern.

This means **we can't use LiveKit Cloud's managed hosting, auto-scaling, or the `livekit-agents` CLI** with Pipecat bots without significant custom wrappers.

### Proposed Solution

I've been prototyping a solution that adds Worker pattern support via two changes:

### 1. **New `LiveKitWorkerRunner`**
A runner class (e.g., `src/pipecat/runner/livekit_worker.py`) that:
- Wraps `livekit.agents.WorkerOptions` and `cli.run_app`
- Listens for job assignments from LiveKit
- Takes a `pipeline_factory` callback to spin up pipelines when rooms are assigned

### 2. **Update `LiveKitTransport`** 
Allow injecting an existing `rtc.Room` connection: 
- **Current behavior:** Transport always creates its own connection
- **Proposed:** Accept an optional `room` parameter so the Worker can pass in its pre-existing connection

This keeps the transport flexible for both client-mode (existing behavior) and worker-mode (new).

### Alternative Solutions

I thought about maintaining a separate wrapper library, but:

1. It would duplicate Runner logic
2. Every Pipecat user would reinvent this wheel
3. It feels like a natural fit for core Pipecat

## Outcome

-  Deploying Pipecat to LiveKit Cloud's managed agents
-  Using LiveKit's auto-scaling and monitoring
-  Standard `livekit-agents` CLI tooling
-  Production-grade deployments without custom glue code

## Livekit Docs

- LiveKit Agents docs: https://docs.livekit.io/agents/overview/

- [ ] No, I'm just suggesting

xprabhudayal:
# feat: Add LiveKit Agents Worker Support

Closes #3505

## Summary

This PR adds native support for running Pipecat pipelines as **LiveKit Agents (Workers)**. Previously, Pipecat could only operate as a client that dials into rooms. Now it can run as a Worker that listens for job assignments from LiveKit's agent infrastructure.

### 1. New `LiveKitWorkerRunner` class

**File:** `src/pipecat/runner/livekit_worker.py`

A new runner that wraps `livekit-agents` to handle:

- Job requests from LiveKit server
- Room connections via `ctx.connect()`
- Pipeline creation through a factory pattern
- Standard `livekit-agents` CLI integration (`dev`, `start`, etc.)

### 2. Updated `LiveKitTransport` constructor

**File:** `src/pipecat/transports/livekit/transport.py`

- Made `url`, `token`, `room_name` optional when `room` is provided
- Added validation: requires either `room` (worker mode) OR connection params (client mode)
- Room name auto-derived from injected room object

**Before (verbose):**

```python
transport = LiveKitTransport(
    url="ws://localhost:7880",
    token="...",
    room_name=room.name,
    room=room,
)
```

**After (clean):**

```python
transport = LiveKitTransport(room=room, params=LiveKitParams(...))
```

### 3. Example implementation

**File:** `examples/livekit_agent_bot.py`

Complete working example showing:

- Pipeline factory pattern
- Room injection
- VAD + STT + LLM + TTS pipeline
- Event handlers for participant join/disconnect

## Usage

```python
from pipecat.runner.livekit_worker import LiveKitWorkerRunner
from pipecat.transports.livekit.transport import LiveKitTransport, LiveKitParams

async def pipeline_factory(room):
    transport = LiveKitTransport(room=room, params=LiveKitParams(...))
    # ... build pipeline
    return transport, task

if __name__ == "__main__":
    runner = LiveKitWorkerRunner()
    runner.set_pipeline_factory(pipeline_factory)
    runner.run()  # Listens for jobs
```

Run with:

```bash
python examples/livekit_agent_bot.py dev
```

## Testing

Tested locally with:

- `livekit-server --dev`
- LiveKit Agents Playground (https://agents-playground.livekit.io)
- Full voice conversation loop working (STT -> LLM -> TTS)

## Output 
image

image_1

image_2

### DIFF
--- a/examples/livekit_agent_bot.py
+++ b/examples/livekit_agent_bot.py
@@ -137,14 +137,14 @@ async def pipeline_factory(room):
         voice_id="abhilash",
         model="bulbul:v2",
         aiohttp_session=session,
-        aggregate_sentences=False,
+        aggregate_sentences=True,  # Buffer text to avoid single-character sends
     )
     logger.info(f"✅ Sarvam TTS created (API key set: {bool(os.getenv('SARVAM_API_KEY'))})")
 
     messages = [
         {
             "role": "system",
-            "content": "You are a helpful AI assistant running as a LiveKit Agent. Respond politely and concisely.",
+            "content": "You are a helpful AI assistant running as a LiveKit Agent. Always respond in English, regardless of what language the user speaks. Respond politely and concisely.",
         },
     ]
 


examples/livekit_agent_bot.py
aipdv, line 1: - this file is used for testing this out - it works great (kinda verbose)
- i used soniox, sarvamai and groq cloud for voice pipeline

markbackman:
See #3505 for more information.

PR #3508, opened 2026-01-20
Log Daily participant and meeting session IDs upon successful join in…

sunahsuh:
… Daily Transport

Logs Daily participant and meeting IDs upon join, if available. Looking over the latest contributor guidelines, I'm not sure if this warrants a changelog entry – I'm thinking no..?

Tested in a staging pipecat cloud deployment and it looks as expected

### DIFF
--- a/src/pipecat/transports/daily/transport.py
+++ b/src/pipecat/transports/daily/transport.py
@@ -759,7 +759,11 @@ class DailyTransportClient(EventHandler):
             # Increment leave counter if we successfully joined.
             self._leave_counter += 1
 
-            logger.info(f"Joined {self._room_url}")
+            participant_id = data.get("participants", {}).get("local", {}).get("id")
+            meeting_id = data.get("meetingSession", {}).get("id")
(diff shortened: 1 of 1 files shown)


src/pipecat/transports/daily/transport.py
```diff
@@ -759,7 +759,11 @@ async def join(self):
             # Increment leave counter if we successfully joined.
             self._leave_counter += 1
 
-            logger.info(f"Joined {self._room_url}")
+            participant_id = data.get("participants", {}).get("local", {}).get("id")
```
markbackman, line 762: This will be the Pipecat bot's participant_id, is that right?

markbackman (approved):
Looks good! Just one clarifying question.

No changelog needed.

src/pipecat/transports/daily/transport.py
sunahsuh: yes! which will be very useful to have codified in the logs

PR #3510, opened 2026-01-21
feat(task): add additive filter methods for frame monitoring

### LINKED ISSUE #3507: Is there a reason `set_reached_downstream_filter` is destructive?
### pipecat version

0.0.95

### Python version

3.12

### Operating System

_No response_

### Question

The event listener for monitoring which frames have reached downstream must first populate which frames should trigger this event via `task.set_reached_downstream_filter`. 

However, this currently just completely overwrites the internal `_reached_downstream_types` list and there is no method for simply appending to this list. I've ran into some odd interactions setting this since I use Pipecat Flow's as well and Pipecat Flows also defines their own tuple of frames to monitor and I think I'm overwriting their frames.

A simple alternative is to just set them like so, but this is just a work-around:
```
new_filter = (EndFrame,)
already_set_filter = task._reached_downstream_types
task.set_reached_downstream_filter(already_set_filter + new_filter)
```

Are there any plans to make overwriting here less likely?

### What I've tried

_No response_

### Context

_No response_

omChauhanDev:
### Problem

There's no public API to **read** or **append to** the frame monitoring filters - user can only replace them via `set_reached_downstream_filter()`.

This limits : when multiple components (e.g., Pipecat Flows & user code) need to monitor different frame types, they must either:
- Access private `_reached_downstream_types` directly (fragile)
- Risk overwriting filters set by other components

### Solution

Added public APIs for reading and appending:
- `reached_upstream_types` / `reached_downstream_types` - read-only properties  
- `add_reached_upstream_filter()` / `add_reached_downstream_filter()` - append without replacing

Closes #3507

### DIFF
--- a/changelog/3510.added.md
+++ b/changelog/3510.added.md
@@ -1,0 +1,1 @@
+- Added `reached_upstream_types` and `reached_downstream_types` read-only properties to `PipelineTask` for inspecting current frame filters.
--- a/src/pipecat/pipeline/task.py
+++ b/src/pipecat/pipeline/task.py
@@ -15,7 +15,7 @@ import asyncio
 import importlib.util
 import os
 from pathlib import Path
-from typing import Any, AsyncIterable, Dict, Iterable, List, Optional, Tuple, Type
+from typing import Any, AsyncIterable, Dict, Iterable, List, Optional, Set, Tuple, Type
 
 from loguru import logger
 from pydantic import BaseModel, ConfigDict, Field
@@ -366,8 +366,8 @@ class PipelineTask(BasePipelineTask):
         # in. This is mainly for efficiency reason because each event handler
         # creates a task and most likely you only care about one or two frame
         # types.
-        self._reached_upstream_types: Tuple[Type[Frame], ...] = ()
-        self._reached_downstream_types: Tuple[Type[Frame], ...] = ()
+        self._reached_upstream_types: Set[Type[Frame]] = set()
+        self._reached_downstream_types: Set[Type[Frame]] = set()
         self._register_event_handler("on_frame_reached_upstream")
         self._register_event_handler("on_frame_reached_downstream")
         self._register_event_handler("on_idle_timeout")
@@ -434,7 +434,7 @@ class PipelineTask(BasePipelineTask):
         Returns:
             Tuple of frame types that trigger the on_frame_reached_upstream event.
         """
-        return self._reached_upstream_types
+        return tuple(self._reached_upstream_types)
 
     @property
     def reached_downstream_types(self) -> Tuple[Type[Frame], ...]:
@@ -443,7 +443,7 @@ class PipelineTask(BasePipelineTask):
         Returns:
             Tuple of frame types that trigger the on_frame_reached_downstream event.
         """
-        return self._reached_downstream_types
+        return tuple(self._reached_downstream_types)
 
     def event_handler(self, event_name: str):
         """Decorator for registering event handlers.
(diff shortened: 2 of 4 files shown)


src/pipecat/pipeline/task.py
```diff
@@ -451,6 +469,30 @@ def set_reached_downstream_filter(self, types: Tuple[Type[Frame], ...]):
...
+        """Add frame types to trigger the on_frame_reached_upstream event.
+
+        Args:
+            types: Tuple of frame types to add to upstream monitoring.
+        """
+        if not types:
+            return
+        current = set(self._reached_upstream_types)
+        current.update(types)
+        self._reached_upstream_types = tuple(current)
```
aconchillo: A bit shorter:

```
self._reached_upstream_types.update(types)
```

changelog/3510.added.md
```diff
@@ -0,0 +1,4 @@
+### Added
+
+- Added `reached_upstream_types` and `reached_downstream_types` read-only properties to `PipelineTask` for inspecting current frame filters.
+- Added `add_reached_upstream_filter()` and `add_reached_downstream_filter()` methods to `PipelineTask` for appending frame types.
```
aconchillo: no need for `### Added` since it's automatically added by towncrier.

split the two lines into two files 3510.added.md and 3510.added.2.md. All the additional lines added in one file will be indented which is useful in some cases.

src/pipecat/pipeline/task.py
```diff
@@ -451,6 +469,30 @@ def set_reached_downstream_filter(self, types: Tuple[Type[Frame], ...]):
...
+        """Add frame types to trigger the on_frame_reached_downstream event.
+
+        Args:
+            types: Tuple of frame types to add to downstream monitoring.
+        """
+        if not types:
+            return
+        current = set(self._reached_downstream_types)
+        current.update(types)
+        self._reached_downstream_types = tuple(current)
```
aconchillo: same here.

src/pipecat/pipeline/task.py
omChauhanDev: I considered that, but `self._reached_upstream_types` is a `tuple` (immutable), so it doesn't have an `.update()` method. 

Also the in-between conversion to `set` handles deduplication when the same frame type is added multiple times, though we can still shorten it like this : 

```python
self._reached_upstream_types = tuple(set(self._reached_upstream_types) | set(types))
``` 

kindly, lemme know if this is preferred!

changelog/3510.added.md
omChauhanDev: sure, done

src/pipecat/pipeline/task.py
aconchillo: Oh! I forgot it was a tuple. Then, I think we should probably change the `_reached_upstream_types`  type to a set:

```
self._reached_upstream_types: Set[Type[Frame]] = set()
self._reached_downstream_types: Set[Type[Frame]] = set()
```

so then you can just do:

```
self._reached_upstream_types.update(types)
```

and also

```
if isinstance(frame, tuple(self._reached_upstream_types))
```

I believe this should work well and probably cleaner.

omChauhanDev: make sense, done

src/pipecat/pipeline/task.py
omChauhanDev: done

aconchillo (approved):

aconchillo:
LGTM! Thanks!

itissid:
@aconchillo

Is it documented that FlowManager initialization should precede any `add_reached_downstream_filter` calls? right now in bot.py's we have to do 

```python
def run_bot():
   ...
    task = PipelineTask(pipeline, params,...)
    flow_manager = FlowManager(task=task...)
   
    # this must be below flow_manager else ActionManager will reset it in its `__init__`
    task.add_reached_downstream_filter((MyFrame,))
   
```

PR #3516, opened 2026-01-21
refactor(user_mute): remove unnecessary _bot_speaking assignment in _handle_bot_stopped_speaking

okue:

The `_bot_speaking` flag does not need to be set in this method, so the redundant assignment has been removed.


markbackman (approved):
Nice catch! 🕵️ Thank you.

PR #3519, opened 2026-01-21
automatically add RTVI to the pipeline

aconchillo:
## Summary

- **Auto-enable RTVI**: `PipelineTask` now automatically adds the RTVI processor and observer when `enable_rtvi=True` (default), simplifying pipeline setup
- **Improved `broadcast_frame()` API**: `FrameProcessor.broadcast_frame()` now accepts a frame instance (consistent with `push_frame()`), with `broadcast_frame_class()` available for the class+kwargs pattern
- **Transport message broadcasting**: WebSocket and other transports now properly broadcast `InputTransportMessageFrame` frames both upstream and downstream, which allow us to add RTVI processor anywhere in the pipeline

### RTVI Auto-Configuration
- `PipelineTask` automatically inserts `RTVIProcessor` and registers `RTVIObserver` when `enable_rtvi` is enabled
- Added `RTVIProcessor.create_rtvi_observer()` factory method
- Removed manual RTVI setup from foundational examples (now handled automatically)

### FrameProcessor API
- `broadcast_frame(frame: Frame)` - takes a frame instance, extracts class and init fields, creates two copies for upstream/downstream
- `broadcast_frame_class(frame_cls: Type[Frame], **kwargs)` - previous behavior, takes class + kwargs

### Transport Improvements  
- WebSocket transports (client, server, FastAPI) now broadcast `InputTransportMessageFrame` frames instead of only pushing downstream

### DIFF
--- a/src/pipecat/transports/websocket/fastapi.py
+++ b/src/pipecat/transports/websocket/fastapi.py
@@ -26,6 +26,7 @@ from pipecat.frames.frames import (
     EndFrame,
     Frame,
     InputAudioRawFrame,
+    InputTransportMessageFrame,
     InterruptionFrame,
     OutputAudioRawFrame,
     OutputTransportMessageFrame,
@@ -311,6 +312,8 @@ class FastAPIWebsocketInputTransport(BaseInputTransport):
 
                 if isinstance(frame, InputAudioRawFrame):
                     await self.push_audio_frame(frame)
+                elif isinstance(frame, InputTransportMessageFrame):
+                    await self.broadcast_frame(frame)
                 else:
                     await self.push_frame(frame)
         except Exception as e:
--- a/changelog/3519.added.2.md
+++ b/changelog/3519.added.2.md
@@ -1,0 +1,1 @@
+- Added `RTVIProcessor.create_rtvi_observer()` factory method for creating RTVI observers.
--- a/changelog/3519.added.3.md
+++ b/changelog/3519.added.3.md
@@ -1,0 +1,1 @@
+- Added `FrameProcessor.broadcast_frame_instance(frame)` method to broadcast a frame instance by extracting its fields and creating new instances for each direction.
--- a/changelog/3519.added.md
+++ b/changelog/3519.added.md
@@ -1,0 +1,1 @@
+- `PipelineTask` now automatically adds `RTVIProcessor` and registers `RTVIObserver` when `enable_rtvi=True` (default), simplifying pipeline setup.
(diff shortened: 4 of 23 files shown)


markbackman (approved):
😍 I love this change!!!

One question: when should `broadcast_frame()` be used vs `broadcast_frame_class()`? The latter naming (`broadcast_frame_class`) doesn't make it obvious what it accomplishes, but that could just be me.

Anyway, approving because this is a really great change.

aconchillo:
> 
> One question: when should `broadcast_frame()` be used vs `broadcast_frame_class()`? The latter naming (`broadcast_frame_class`) doesn't make it obvious what it accomplishes, but that could just be me.
> 

You use `broadcat_frame` one you have a frame instance. In some cases, like when you call a deserialzier, you  end up with a frame instance and you would use `broadcast_frame` for that. What `broadcast_frame` does is  unpack the frame type and fields and then call `broadcast_frame_class`, so `broadcast_frame` is actually less efficient than `broadcast_frame_class`. 

Now that I'm thinking more about it. I think `broadcast_frame` should stay as before because we want people to use that actually (no unpacking of fields, etc.). So, I'm adding a `broadcast_frame_instance` instead which will be the slower version.

markbackman:
I like that plan. The more concise name being the more commonly used one seems like a good idea!

aconchillo:
> I like that plan. The more concise name being the more commonly used one seems like a good idea!

Also added a note to the new function `broadcast_frame_instance` to indicate it should only be sued in certain cases and that `broadcast_frame` is preferred.

```
        Note:
             Prefer using `broadcast_frame()` when possible, as it is more
             efficient. This method should only be used when you are not the
             creator of the frame and need to broadcast an existing instance.
```

mattieruth:
i know i'm late to this party, but i still wanted to say, whoopie!

src/pipecat/transports/websocket/fastapi.py
```diff
@@ -311,6 +312,8 @@ async def _receive_messages(self):
 
                 if isinstance(frame, InputAudioRawFrame):
                     await self.push_audio_frame(frame)
+                elif isinstance(frame, InputTransportMessageFrame):
+                    await self.broadcast_frame(frame)
```
okue, line 316: @aconchillo 
Hi, I found type mismatch happens here. (`type[Frame]` vs. `Frame` instance)

`FastAPIWebsocketTransport` fails when I run the Pipecat with main branch.

PR #3529, opened 2026-01-22
feat: handle exceptions for BaseOpenAILLMService

### LINKED ISSUE #3481: Allow passing timeout to LLM services (for FAILING, not for retry)
## Use Case

We just noticed the new'ish LLMSwitcher and started to use it - awesome addition! So we wanted to have timeouts (~2s or so) trigger an error that we could handle in our code to call upon the LLM Switcher to do its thing, but then we noticed that the only way to do this was to either:

1. Make changes to the private _client member of the LLM processor
2. Add a bunch of code to monitor and actively interrupt the LLM if it takes too long

## Request

We'd like to be able to pass the timeout to the LLM service constructor.

Thanks!
-Al

### Proposed Solution

We only use OpenAI-like LLMs so I'm not sure how hard this is to do across all LLM services.

But for OpenAI-like's, we also noticed the timeout can have more granular settings because of the underlying HTTPX timeout settings - bonus points if we can even specify those too!

But honestly, even being able to pass a single timeout that will make it fail at any point in the response if there's `>timeout` seconds of inactivity would be great.

### Alternative Solutions

If it's too tricky to make this work for all LLM Services in a standard way, just exposing the underlying client on a per-llm-service basis

### Additional Context

_No response_

- [x] No, I'm just suggesting

lukepayyapilli:
## Summary

Enables `retry_timeout_secs` to cause failure (not just retry) when `retry_on_timeout=False`, allowing integration with `LLMSwitcher` for slow response handling.

Closes #3481

## Changes

- [x] When `retry_timeout_secs` is explicitly set, the timeout now applies regardless of `retry_on_timeout`.
- [x] When timeout occurs and `retry_on_timeout=False`, the `TimeoutError` propagates (enabling LLMSwitcher).
- [x] Backwards compatible: default behavior unchanged.

## Behavior Matrix

| `retry_timeout_secs` | `retry_on_timeout` | Behavior |
|---------------------|-------------------|----------|
| `None` (default) | `False` (default) | No timeout |
| `None` | `True` | 5.0s timeout + retry |
| Explicit (e.g., 5.0) | `False` | Timeout + **fail** (NEW) |
| Explicit (e.g., 5.0) | `True` | Timeout + retry |

## Usage

```python
# For LLMSwitcher integration - fail on slow responses
llm = OpenAILLMService(
    model="gpt-4",
    retry_timeout_secs=2.0,  # Fail after 2 seconds
    retry_on_timeout=False,   # Don't retry, let LLMSwitcher handle it
)
```

## Design Considerations

The parameter name `retry_timeout_secs` is admittedly misleading since it now controls non-retry behavior too. I considered renaming to `timeout_secs` but opted for backwards compatibility. A future cleanup could:

- [ ] Deprecate `retry_timeout_secs` in favor of clearer `timeout_secs`.
- [ ] Simplify the parameter interaction.

### DIFF
--- a/src/pipecat/services/openai/base_llm.py
+++ b/src/pipecat/services/openai/base_llm.py
@@ -495,6 +495,8 @@ class BaseOpenAILLMService(LLMService):
             except httpx.TimeoutException as e:
                 await self._call_event_handler("on_completion_timeout")
                 await self.push_error(error_msg="LLM completion timeout", exception=e)
+            except Exception as e:
+                await self.push_error(error_msg=f"Error during completion: {e}", exception=e)
             finally:
                 await self.stop_processing_metrics()
                 await self.push_frame(LLMFullResponseEndFrame())
--- a/src/pipecat/services/openai/base_llm.py
+++ b/src/pipecat/services/openai/base_llm.py
@@ -492,8 +492,9 @@ class BaseOpenAILLMService(LLMService):
                 await self.push_frame(LLMFullResponseStartFrame())
                 await self.start_processing_metrics()
                 await self._process_context(context)
-            except httpx.TimeoutException:
+            except httpx.TimeoutException as e:
                 await self._call_event_handler("on_completion_timeout")
+                await self.push_error(error_msg="LLM completion timeout", exception=e)
             finally:
                 await self.stop_processing_metrics()
                 await self.push_frame(LLMFullResponseEndFrame())
--- a/changelog/3529.fixed.md
+++ b/changelog/3529.fixed.md
@@ -1,0 +1,1 @@
+- Fixed OpenAI LLM services to emit `ErrorFrame` on completion timeout, enabling proper error handling and LLMSwitcher failover.
--- a/tests/test_openai_llm_timeout.py
+++ b/tests/test_openai_llm_timeout.py
@@ -1,0 +1,127 @@
+#
+# Copyright (c) 2024-2026, Daily
+#
+# SPDX-License-Identifier: BSD 2-Clause License
+#
+
+"""Unit tests for OpenAI LLM timeout handling."""
+
+from unittest.mock import AsyncMock, patch
+
+import httpx
+import pytest
+
+from pipecat.frames.frames import (
+    LLMContextFrame,
+    LLMFullResponseEndFrame,
+    LLMFullResponseStartFrame,
+)
+from pipecat.processors.aggregators.llm_context import LLMContext
+from pipecat.processors.frame_processor import FrameDirection
+from pipecat.services.openai.llm import OpenAILLMService
+
+
+@pytest.mark.asyncio
+async def test_openai_llm_emits_error_frame_on_timeout():
+    """Test that OpenAI LLM service emits ErrorFrame when a timeout occurs.
+
+    This enables LLMSwitcher to trigger failover to backup LLMs when the
+    primary LLM times out.
+    """
+    with patch.object(OpenAILLMService, "create_client"):
+        service = OpenAILLMService(model="gpt-4")
+        service._client = AsyncMock()
+
+        # Track pushed frames and errors
+        pushed_frames = []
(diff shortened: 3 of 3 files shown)


markbackman (commented):
Good catch!

It sounds like you want an exception raised (or in Pipecat terms, an ErrorFrame pushed) when an exception is caught as part of a standard completion. That makes sense. I don't think I'd integrate it with the retry system though.

We could just update the `process_frame()` method to handle the exception:
```
        """Process frames for LLM completion requests.

        Handles OpenAILLMContextFrame, LLMContextFrame, LLMMessagesFrame,
        and LLMUpdateSettingsFrame to trigger LLM completions and manage
        settings.

        Args:
            frame: The frame to process.
            direction: The direction of frame processing.
        """

        context = None
        if isinstance(frame, OpenAILLMContextFrame):
            # Handle OpenAI-specific context frames
            context = frame.context
        elif isinstance(frame, LLMContextFrame):
            # Handle universal (LLM-agnostic) LLM context frames
            context = frame.context
            # NOTE: LLMMessagesFrame is deprecated, so we don't support the newer universal
            # LLMContext with it
            context = OpenAILLMContext.from_messages(frame.messages)
        elif isinstance(frame, LLMUpdateSettingsFrame):
            await self._update_settings(frame.settings)
        else:

        if context:
            try:
                await self._process_context(context)
            except httpx.TimeoutException:
                await self._call_event_handler("on_completion_timeout")
            except Exception as e:
                await self.push_error(error_msg=f"Error during completion: {e}", exception=e)
            finally:
```

The new code being:
```
            except Exception as e:
                await self.push_error(error_msg=f"Error during completion: {e}", exception=e)
```

That is, an exception would be raised as a result of the completion failure. That would be caught in process_frame, emitting an ErrorFrame which will allow application code to catch the error. GoogleLLMService follows a similar pattern.

WDYT?

lukepayyapilli (commented):
@markbackman Thanks for the review and feedback! Your suggested approach of catching exceptions in `process_frame()` and pushing an `ErrorFrame` is a better fix than my original proposal.

I've intentionally kept the scope narrow to address the original issue (#3481) which specifically requested timeout handling for `LLMSwitcher` failover. The change only catches `httpx.TimeoutException` and emits an error for that case. Happy to broaden it to handle other exceptions (e.g., API errors, rate limits) if you'd like.

Also added tests to verify:
- [x] `ErrorFrame` is pushed with correct message on timeout.
- [x] `LLMFullResponseEndFrame` is still pushed in the finally block.
- [x] `on_completion_timeout` handler is still called.

src/pipecat/services/openai/base_llm.py
```diff
@@ -492,8 +492,9 @@ async def process_frame(self, frame: Frame, direction: FrameDirection):
                 await self.push_frame(LLMFullResponseStartFrame())
                 await self.start_processing_metrics()
                 await self._process_context(context)
-            except httpx.TimeoutException:
+            except httpx.TimeoutException as e:
                 await self._call_event_handler("on_completion_timeout")
+                await self.push_error(error_msg="LLM completion timeout", exception=e)
```
markbackman, line 497: Good call on pushing an error frame here!

Let's also add a catch all for other exceptions after handling the TimeoutException:
```
            except Exception as e:
                await self.push_error(error_msg=f"Error during completion: {e}", exception=e)
```

This will be more robust and will allow developers to handle different exception cases.

markbackman (commented):
Very nice! I like your idea about TimeoutException. Just one suggestion about broadening the handling for `Exception`.

Also, thanks for tests 🙏

lukepayyapilli (commented):
@markbackman - done! Added the catch-all `Exception` handler as suggested. Also added a test to verify the general exception handling works correctly. Please let me know if you'd like me to change anything else and thank you for the suggestions!

markbackman (approved):
Perfect! 🙌

Issue #3535, opened 2026-01-22
pipecat context corruption in non-english languages


surapuramakhil:
### pipecat version

0.0.98

### Python version

3.12

### Operating System

macOS

### Issue description

when working with non english languages - LLM context becomes unreliable 

### Reproduction steps

1. configure pipecat ot use a non english language
2. form logs unicode chars are converstion into asciii - which AI will find hard to undersand - especially when you task / system prompts in unicode

### Expected behavior

context shouldn't be corrupted. 

### Actual behavior

you see something like this in context

**"\\u5f0a\\u793e\\u300c\\u30ad\\u30e3\\u30da\\u30c3\\u30af\\u30b9\\u300d\\u306f\\u3001\\u5bfe\\u8a71\\u30b7\\u30b9\\u30c6\\u30e0\\uff08\\u30c1\\u30e3\\u30c3\\u30c8\\u30dc\\u30c3\\u30c8\\u3084\\u97f3\\u58f0\\u30a2\\u30b7\\u30b9\\u30bf\\u30f3\\u30c8\\uff09\\u3092\\u6d3b\\u7528\\u3057\\u3001\\u30c7\\u30b8\\u30bf\\u30eb\\u30de\\u30fc\\u30b1\\u30c6\\u30a3\\u30f3\\u30b0\\u3084\\u55b6\\u696d\\u6d3b\\u52d5\\u3092\\u30c8\\u30fc\\u30bf\\u30eb\\u306b\\u30b5\\u30dd\\u30fc\\u30c8\\u3059\\u308b\\u4f1a\\u793e\\u3067\\u3059\\u3002\\n\\u6700\\u5148\\u7aef\\u306e\\u30b3\\u30df\\u30e5\\u30cb\\u30b1\\u30fc\\u30b7\\u30e7\\u30f3\\u6280\\u8853\\u3092\\u99c6\\u4f7f\\u3057\\u3066\\u3001\\u9867\\u5ba2\\u3068\\u306e\\u6df1\\u3044\\u95a2\\u4fc2\\u69cb\\u7bc9\\u3092\\u5b9f\\u73fe\\u3057\\u3001\\u5fa1\\u793e\\u306e\\u30d3\\u30b8\\u30cd\\u30b9\\u6210\\u9577\\u306b\\u8ca2\\u732e\\u3044\\u305f\\u3057\\u307e\\u3059\\u3002"}'**

```
[...]

PR #3536, opened 2026-01-22
LLMAssistantAggregator: preserve non-ASCII characters in JSON output

surapuramakhil:
Add ensure_ascii=False to json.dumps() calls for tool call arguments and function call results to prevent unnecessary unicode escaping.

solves #3535


markbackman (approved):
LGTM! Nice catch.

PR #3537, opened 2026-01-22
Fix the /start and /offer/api proxy endpoints for smallWebRTC to matc…

mattieruth:
…h pipecat cloud behavior WRT requestData

⚠️  **IMPORTANT** Note that this change affects the format for sending `requestData` when sending any custom parameters they want passed through to the runner args for SmallWebRTC. Any custom args must now be embedded in a `body: {}` field of `requestData` in the `/start` endpoint. This aligns expectations with the Daily /start endpoint as well as pipecat cloud.


filipi87 (approved):
Yep. Thank you for fixing it. 🚀

Issue #2851, opened 2025-10-14
AudioBufferProcessor miscomputing silence?


glennpow:
### pipecat version

0.0.90

### Python version

3.11.11

### Operating System

macOS 15.6.1

### Issue description

I was noticing what seemed like misalignment of the user and bot audio clips in the merged output from `AudioBufferProcessor`, and after looking at the code I'm wondering if it's because the computed silence might be incorrect.  It is currently using properties to track the last frame times and then subtracting from current time to determine how much silence to add to the output streams.  However, it's also adding the actual audio clip to the stream, so wouldn't that mean that the silence should also take into account the duration of the audio clip?  I.e. for the user audio stream:
```
            frame_time = frame.num_frames / frame.sample_rate  # This should be added to the last timestamp?
            self._last_user_frame_at = time.time() + frame_time
```

### Reproduction steps

Save merged audio output.

### Expected behavior

User and bot audio should be properly aligned.

### Actual behavior

Audio isn't aligned.

### Logs

```shell

```

markbackman:
Can you share a simple repro case demonstrating the issue?

We recently fixed an alignment issue and our testing confirms all is well. It's possible that you're using the AudioBufferProcessor in a way that we haven't tested, so sharing a repro case will help.

glennpow:
@markbackman Perhaps I wasn't using the latest code, but doesn't my calculation still stand?  I made a diagram to show what I believe to be the currently calculated silence time versus what I'd expect.  Currently, it doesn't take into account the duration of the audio clip, so this would be accurate, right?  (The image may be wrong to use the audio start times as the event triggers, rather than the audio stop which I believe is actually correct, but it should still work for explanatory purposes)

<img width="964" height="300" alt="Image" src="https://github.com/user-attachments/assets/f729755a-60a7-48b7-9ceb-b750afed2891" />

ParthShindenovus:
@glennpow i am also having this kind of behavior from 4 to 5 days, I am having user's audio delayed in the recording

In actual scenerio I have answered all the bot question immediately once bot is stopped speaking

But In Recordings i am getting user's audio way to delayed

I am attaching my audio recording below please hear this for once

[merged_20251016_101014.wav](https://github.com/user-attachments/files/22963180/merged_20251016_101014.wav)

glennpow:
[...]

PR #3541, opened 2026-01-23
Refactoring AudioBufferProcessor to fix audio track synchronization.

filipi87:
We were trying to use `time.time()` to compute how much silence to insert. But for some reason, this was not reliable enough, and the audio was appended at the wrong position relative to the other track.

So, in this PR we are simplifying the logic of how we are keeping both tracks in sync. We are now using buffer positions as the source of truth.
- When we add audio to one buffer, we are immediately ensuring the other buffer is at the same position (by padding with silence if needed).
- When user audio arrives, we first pad the bot buffer with silence up to where the user buffer currently is.
- When bot audio arrives, we first pad the user buffer with silence up to where the bot buffer currently is.

This way, audio is always added at the correct synchronized position.

I have tested it with `Daily`, `SmallWebRTC`, and `FastApiWebsocket`. Using all these transports, it feels like it is working as expected.

### DIFF
--- a/changelog/3541.fixed.md
+++ b/changelog/3541.fixed.md
@@ -1,0 +1,1 @@
+- Fixed how audio tracks are synchronized inside the `AudioBufferProcessor` to fix timing issues where silence and audio were misaligned between user and bot buffers.
--- a/src/pipecat/processors/audio/audio_buffer_processor.py
+++ b/src/pipecat/processors/audio/audio_buffer_processor.py
@@ -11,7 +11,6 @@ of audio from both user input and bot output sources, with support for various a
 configurations and event-driven processing.
 """
 
-import time
 from typing import Optional
 
 from pipecat.audio.utils import create_stream_resampler, interleave_stereo_audio, mix_audio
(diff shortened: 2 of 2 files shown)


markbackman:
@filipi87 with the current code, were you able to reproduce the issue where there could be a small drift when using a websocket transport (with or without intermittent stream disruptions from something like a user mute strategy)?

filipi87:
This PR fixes https://github.com/pipecat-ai/pipecat/issues/2851

filipi87:
> @filipi87 with the current code, were you able to reproduce the issue where there could be a small drift when using a websocket transport (with or without intermittent stream disruptions from something like a user mute strategy)?

Hi @markbackman, yes, I can easily reproduce the issue on main.

If I change the `34-audio-recording.py`, to use the audio input and output required by Twilio:
```
params=PipelineParams(
            enable_metrics=True,
            enable_usage_metrics=True,
            audio_in_sample_rate=8000,
            audio_out_sample_rate=8000,
        ),
```

Then start the example using Twilio as transport:
`python examples/foundational/34-audio-recording.py -t twilio`

And connect using the client which we provide here:
- https://github.com/pipecat-ai/pipecat-examples/tree/main/twilio-chatbot/inbound/client/typescript 

I can easily reproduce the issue. As you can see in the image below. That is when I interrupted the bot:

Screenshot 2026-01-23 at 17 14 22

markbackman (approved):
Very nice and clean! Thanks for the clear comments on how this is being solved 🙏

PR #3543, opened 2026-01-23
Fix await usage in transcription timeout task

kompfner:
Looks like this `await` was missing from https://github.com/pipecat-ai/pipecat/pull/3356


kompfner:
Oh wait, this isn't right (I opened this PR with zero thought after seeing warnings in the console)

kompfner:
Here was the warning

Screenshot 2026-01-23 at 11 55 12 AM

PR #3544, opened 2026-01-23
Fix: Handle both camelCase and snake_case for request_data field in SmallWebRTC requests

SRai22:
The Pipecat /api/offer endpoint fails to receive custom configuration data sent from JavaScript clients due to a field name mismatch between JavaScript's camelCase convention (requestData) and Python's snake_case convention (request_data). This PR converts SmallWebRTCRequest from a @dataclass to a Pydantic BaseModel with field aliasing to accept both naming conventions.

Changes
- Converted SmallWebRTCRequest from @dataclass to Pydantic BaseModel in src/pipecat/transports/smallwebrtc/request_handler.py
- Added alias="requestData" to the request_data field to support camelCase from JavaScript clients
- Added Config class with populate_by_name = True to maintain backwards compatibility with existing Python code using request_data
- Updated from_dict() classmethod to use Pydantic's model_validate() for consistent field aliasing behavior 

Backwards Compatibility
- Function signatures remain unchanged
- Existing Python code using request_data continues to work
- Existing code using from_dict() continues to work
- Already includes Pydantic as a dependency (no new dependencies)

### DIFF
--- a/src/pipecat/runner/run.py
+++ b/src/pipecat/runner/run.py
@@ -327,10 +327,10 @@ def _setup_webrtc_routes(
                 if request.method == HTTPMethod.POST.value:
                     request_data = await request.json()
                     webrtc_request = SmallWebRTCRequest.model_validate(request_data)
-                    
+
                     if webrtc_request.request_data is None:
                         webrtc_request.request_data = active_session
-                    
+
                     return await offer(webrtc_request, background_tasks)
                 elif request.method == HTTPMethod.PATCH.value:
                     patch_request = SmallWebRTCPatchRequest(
--- a/src/pipecat/transports/smallwebrtc/request_handler.py
+++ b/src/pipecat/transports/smallwebrtc/request_handler.py
@@ -13,10 +13,11 @@ import asyncio
 from dataclasses import dataclass
 from enum import Enum
 from typing import Any, Awaitable, Callable, Dict, List, Optional
-from pydantic import BaseModel, Field
+
 from aiortc.sdp import candidate_from_sdp
 from fastapi import HTTPException
 from loguru import logger
+from pydantic import BaseModel, Field
 
 from pipecat.transports.smallwebrtc.connection import IceServer, SmallWebRTCConnection
 
@@ -40,6 +41,12 @@ class SmallWebRTCRequest(BaseModel):
     request_data: Optional[Any] = Field(None, alias="requestData")
 
     class Config:
+        """Pydantic configuration for the model.
+
+        `populate_by_name` allows population of fields by their Python
+        attribute names as well as any aliases (e.g. camelCase).
+        """
+
         populate_by_name = True
 
     @classmethod
--- a/changelog/3544.fixed.md
+++ b/changelog/3544.fixed.md
@@ -1,0 +1,1 @@
+- Fixed `SmallWebRTCRequest` to accept both camelCase (`requestData`) and snake_case (`request_data`) for cross-platform compatibility between JavaScript and Python clients.


markbackman:
cc @filipi87 to take a look.

@SRai22 you'll also need:
- lint your code (either `uv run scripts/fix-ruff.sh` or install pre-commit hook via `uv run pre-commit install`)
- Add a changelog fragment: https://github.com/pipecat-ai/pipecat/blob/main/CONTRIBUTING.md#changelog-entries

SRai22:
> cc @filipi87 to take a look.
> 
> @SRai22 you'll also need:
> 
> * lint your code (either `uv run scripts/fix-ruff.sh` or install pre-commit hook via `uv run pre-commit install`)
> * Add a changelog fragment: https://github.com/pipecat-ai/pipecat/blob/main/CONTRIBUTING.md#changelog-entries

@markbackman : made the linting fixes and added the changelog fragment. Thanks for the feedback

src/pipecat/runner/run.py
```diff
@@ -325,14 +325,12 @@ async def proxy_request(
...
                 if request.method == HTTPMethod.POST.value:
-                    webrtc_request = SmallWebRTCRequest(
-                        sdp=request_data["sdp"],
-                        type=request_data["type"],
-                        pc_id=request_data.get("pc_id"),
-                        restart_pc=request_data.get("restart_pc"),
-                        request_data=request_data.get("request_data")
-                        or request_data.get("requestData"),
-                    )
+                    request_data = await request.json()
```
filipi87, line 328: We already have the same line above the `if`, so you should have access to `request_data` at this point.

src/pipecat/transports/smallwebrtc/request_handler.py
```diff
@@ -17,12 +17,13 @@
...
 from fastapi import HTTPException
 from loguru import logger
+from pydantic import BaseModel, Field
 
 from pipecat.transports.smallwebrtc.connection import IceServer, SmallWebRTCConnection
 
 
 @dataclass
-class SmallWebRTCRequest:
+class SmallWebRTCRequest(BaseModel):
```
filipi87, line 26: I don't think we should mix @dataclass and BaseModel.

pydantic.BaseModel already provides its own data handling, validation, and initialization, and @dataclass also generates methods that can conflict with Pydantic’s machinery.

Since we are already using @dataclass, I think we should remove BaseModel and keep it as a pure dataclass.

src/pipecat/runner/run.py
```diff
@@ -325,14 +325,12 @@ async def proxy_request(
...
+                    webrtc_request = SmallWebRTCRequest.model_validate(request_data)
```
filipi87, line 329: This was already possible with the current code, but instead you need to do this:
`webrtc_request = SmallWebRTCRequest.from_dict(request_data)`

src/pipecat/runner/run.py
```diff
@@ -325,14 +325,12 @@ async def proxy_request(
             try:
                 request_data = await request.json()
                 if request.method == HTTPMethod.POST.value:
-                    webrtc_request = SmallWebRTCRequest(
-                        sdp=request_data["sdp"],
-                        type=request_data["type"],
-                        pc_id=request_data.get("pc_id"),
-                        restart_pc=request_data.get("restart_pc"),
-                        request_data=request_data.get("request_data")
-                        or request_data.get("requestData"),
```
filipi87, line 334: The current code already handles both camelCase and snake_case. 

Refactoring it to use `from_dict`, as mentioned above, would not change the behavior but could make the code cleaner.

markbackman:
Hey @SRai22, thanks for taking the time to contribute this PR! We appreciate you identifying the camelCase/snake_case issue.

After reviewing, it looks like the existing code on `main` already handles both conventions — the runner at `run.py:324-325` checks for both `request_data` and `requestData`, and `SmallWebRTCRequest.from_dict()` also normalizes camelCase input. So the fix you're addressing is already covered.

Additionally, as @filipi87 noted in his review, mixing `@dataclass` with Pydantic `BaseModel` can cause conflicts, so we'd want to keep `SmallWebRTCRequest` as a pure dataclass.

Going to close this one out, but thanks again for the contribution!

PR #3518, opened 2026-01-21
Add TurnCompletionMixin

markbackman:
#### Please describe the changes in your PR. If it is addressing an issue, please reference that as well.

This change involves the main conversational LLM in deciding whether a user's input is semantically complete. When complete, the user's input results in a bot response. When incomplete, the bot waits for the user to continue.

This capability is meant to complement the user turn stop strategy (like Smart Turn). It helps prevent the bot from responding too quickly when the user is mid-thought or needs time to formulate their response.

## Summary

- Added `TurnCompletionMixin` to involve the conversational LLM in semantic turn completion
- When complete (✓), the bot responds normally
- When incomplete, the bot waits with two modes:
  - Short incomplete (○): User was cut off mid-sentence, wait briefly (default 5s)
  - Long incomplete (◐): User needs time to think ("let me think...", "hmm..."), wait longer (default 10s)
- After timeout expires, the LLM is prompted to re-engage the user with a contextual follow-up

## Configuration

Enable via `LLMUserAggregatorParams`:
- `filter_incomplete_user_turns=True` - Enables the feature and auto-injects turn completion instructions into the system prompt
- `turn_completion_config` - Optional `TurnCompletionConfig` for customization:
  - `instructions` - Custom turn completion prompt for the LLM
  - `incomplete_short_timeout` - Seconds to wait after ○ (default: 5.0)
  - `incomplete_long_timeout` - Seconds to wait after ◐ (default: 15.0)
  - `incomplete_short_prompt` - Custom prompt when short timeout expires
  - `incomplete_long_prompt` - Custom prompt when long timeout expires

All LLM services (OpenAI, Anthropic, Google, AWS Bedrock) support this capability.

## Testing

```bash
uv run examples/foundational/22-filter-incomplete-turns.py
```

When testing, try a complete phrase, an incomplete/dangling phrase, and also ask for more time.

## Miscellaneous

I'm removing the old 22 series of foundational examples as they're now outdated.

PR #3553, opened 2026-01-24
Fix: OutputGate SystemFrame condition for EndFrame

cyh0530:
The `OutputGate` in the example will cause `EndFrame` never go through properly.

`EndFrame` is a `ControlFrame`, so `if isinstance(frame, SystemFrame)` will never let `EndFrame` to pass in.

### DIFF
--- a/examples/foundational/22b-natural-conversation-proposal.py
+++ b/examples/foundational/22b-natural-conversation-proposal.py
@@ -145,7 +145,7 @@ class OutputGate(FrameProcessor):
         await super().process_frame(frame, direction)
 
         # We must not block system frames.
-        if isinstance(frame, SystemFrame):
+        if isinstance(frame, (SystemFrame, EndFrame)):
             if isinstance(frame, StartFrame):
                 await self._start()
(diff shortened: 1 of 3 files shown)


markbackman:
Thanks for submitting this 🙏 

I'm going to close it though since we're removing the 22 foundational examples in favor of a new, simpler approach that's integrated directly into Pipecat:
https://github.com/pipecat-ai/pipecat/pull/3518

Let us know what you think about the new approach and example.

cyh0530:
This looks interesting! Can you clarify what you mean by wait n seconds then prompt, does it mean the tts part will wait for n seconds hoping the user can finish their sentences? How will it coordinate with smart turn stop secs?

```
- ○ (incomplete short): User was cut off, wait ~5s then prompt
- ◐ (incomplete long): User needs time to think, wait ~15s then prompt
```

markbackman:
There's a built in timeout in the LLMService which will wait `n` seconds based on the LLM deciding there should be a short or long for the incomplete answer. These timeouts—both for short and long—are configurable. This will be available in the 0.0.101 release, which is due out ASAP.

PR #3554, opened 2026-01-24
Simplify STT finalize handling

markbackman:
### Summary

Simplifies the STT finalization methods used for TTFB measurement. Now, only services with explicit server confirmation of finalization (e.g., Deepgram's from_finalize field) should use `request_finalize()` + `confirm_finalize()`. Services without server confirmation rely on the TTFB timeout fallback.

### Rationale

For services without a finalize confirmation message, using the timeout is:
- Simpler: No need for extra method calls in subclasses
- More robust:  If a service sends multiple transcripts after finalize, the timeout waits for the last one
- Equally accurate: Timeliness of TTFB reporting is less important than correctness


aconchillo (approved):

aconchillo:
Nice!

PR #3555, opened 2026-01-24
Align Speechmatics STT TTFB metrics with STT classes

markbackman:

Brings Speechmatics metrics in line with other STT services, making it easier to do like for like comparisons.

cc @sam-s10s


aconchillo:
LGTM!

PR #3560, opened 2026-01-25
FrameSerializer: subclass from BaseObject so we can add events

aconchillo:

Simple change to subclass `FrameSerializer` from `BaseObject`. This allows frame serializer to register and trigger events.


markbackman (approved):
LGTM

PR #3562, opened 2026-01-26
Support TTFS for Speechmatics STT

sam-s10s:
Add the `finalized=True` to the last `TranscriptionFrame` in a turn, as determined by external VAD (Pipecat) or internal (`ADAPTIVE` or `SMART_TURN` modes).


src/pipecat/services/speechmatics/stt.py
```diff
@@ -738,16 +736,33 @@ def attr_from_segment(segment: dict[str, Any]) -> dict[str, Any]:
 
         # If final, then re-parse into TranscriptionFrame
         if finalized:
+            # Do any segments have `is_eou` set to True?
+            if (
+                any(segment.get("is_eou", False) for segment in segments)
+                and self._finalize_requested
+            ):
+                self.confirm_finalize()
```
markbackman, line 745: There are two patterns for detecting the final transcript frame:
1. Use a timeout, which is already part of the base class. This timeout will wait to receive the last TranscriptionFrame (2 second timeout window). Once received, it will emit the MetricsFrame with the TTFB value. This is fine, but just a little slow. Though, slow is OK and not a problem.
2. TTFB values can be emitted faster. The pattern is to call `self.request_finalize()` when the audio is finalized and the STT service is flushing to return a transcript. Then, you receive the final transcript along with some type of state (id, field, etc) that indicates that _this_ is the finalized transcript. Once received, you call `self.confirm_finalize()`, which triggers the MetricsFrame sooner. This is ideal because you can get better fidelity on the transcription frames (and more timely data).

If you have the data available to use approach #2, then you need to make a small update. You should first call `self.request_finalize()` on line 697 after calling `self._client.finalize()`.

sam-s10s: I have done this. LGTM!

markbackman (approved):
Looks great! I see a nice performance improvement 👏

Issue #2106, opened 2025-07-01
invalid_request_error with OpenAI Realtime API from 0.0.69 onward


crro:
### pipecat version

0.0.69

### Python version

3.12.8

### Operating System

linux from dailyco/pipecat-base:latest

### Issue description

For some reason this does not happen to me in 0.0.68, but basically when I'm having a conversation with a bot that uses OpenAI Realtime API I get this error and the bot dies and leaves the room:

`2025-07-01 20:14:03.908 | ERROR | pipecat.pipeline.task:_process_up_queue:568 | 0533c952-88f0-4e10-a7f8-f45fccd33600 - A fatal error occurred: ErrorFrame#1(error: Error: event_id='event_BobedUl333Nb1OtoyABG1' type='error' error=RealtimeError(type='invalid_request_error', code='invalid_value', message='Audio content of 2450ms is already shorter than 2535ms', param=None, event_id=None), fatal: True)`

I have not found a specific pattern that triggers it, I thought it was interruption or function calls but I've seen it happen several times in various scenarios

### Reproduction steps

1. Create a bot with
```
session_properties = SessionProperties(
        input_audio_transcription=InputAudioTranscription(
            model="gpt-4o-transcribe",
            language=language_code,  # Set language for better transcription accuracy
            prompt="This is an interview, stay professional and friendly."
        ),
        turn_detection=SemanticTurnDetection(
            type="semantic_vad",
            eagerness="auto",
            create_response=True,
            interrupt_response=True
        ),
        input_audio_noise_reduction=InputAudioNoiseReduction(type="near_field"),
        instructions=discussion_guide,
        voice=realtime_voice,
    )
    llm = OpenAIRealtimeBetaLLMService(
        api_key=os.getenv("OPENAI_API_KEY"),
        session_properties=session_properties,
        start_audio_paused=False,
        send_transcription_frames=True,
        voice=realtime_voice,
    )
```

3. Register a function
4. Make sure you are using pipecat-ai 0.0.69
5. Have a conversation for at least a few minutes and see it fail
6. Try it with 0.0.68 and it should not happen

### Expected behavior

The interview should complete normally.

### Actual behavior

you get this fatal erro and the interview ends: 

[...]

PR #3567, opened 2026-01-26
Fixed race condition in OpenAIRealtimeBetaLLMService

filipi87:
Fixed race condition in `OpenAIRealtimeBetaLLMService` that could cause an error when truncating the conversation.

### DIFF
--- a/src/pipecat/services/openai_realtime_beta/openai.py
+++ b/src/pipecat/services/openai_realtime_beta/openai.py
@@ -525,6 +525,14 @@ class OpenAIRealtimeBetaLLMService(LLMService):
         # note: ttfb is faster by 1/2 RTT than ttfb as measured for other services, since we're getting
         # this event from the server
         await self.stop_ttfb_metrics()
+
+        if self._current_audio_response and self._current_audio_response.item_id != evt.item_id:
+            logger.warning(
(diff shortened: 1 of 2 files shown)


src/pipecat/services/openai_realtime_beta/openai.py
```diff
@@ -525,6 +525,14 @@ async def _handle_evt_audio_delta(self, evt):
         # note: ttfb is faster by 1/2 RTT than ttfb as measured for other services, since we're getting
         # this event from the server
         await self.stop_ttfb_metrics()
+
+        if self._current_audio_response and self._current_audio_response.item_id != evt.item_id:
```
filipi87, line 529: We were only setting `self._current_audio_response` to `None` when we received the `BotStoppedSpeakingFrame` or when we attempted to truncate the audio.

However, if in the meantime we received a new audio delta, it would be added to the previous `self._current_audio_response`, causing its duration to increase when it shouldn’t.

As a result, when we later tried to truncate it, we would receive an error because `audio_end_ms` was greater than the actual audio duration.

In short, we should only increment the audio duration for audio responses with the same `item_id`.

filipi87:
This PR fixes https://github.com/pipecat-ai/pipecat/issues/2106

markbackman (approved):
Really nice catch! 🙌

PR #3571, opened 2026-01-27
Function call improvements

filipi87:
Function call improvements:
- Removed an overridden method that was identical to the parent implementation.
- Provided a way to defer the function call results.
- Updated the examples which use `UserImageRequestFrame` to defer the function call result.


src/pipecat/services/google/llm.py
```diff
@@ -199,22 +198,6 @@ async def _update_function_call_result(
                     if part.function_response and part.function_response.id == tool_call_id:
                         part.function_response.response = {"value": json.dumps(result)}
 
-    async def handle_user_image_frame(self, frame: UserImageRawFrame):
```
filipi87, line 202: `GoogleAssistantContextAggregator` extends `OpenAIAssistantContextAggregator`. This method that was identical to the parent implementation.

examples/foundational/14d-function-calling-anthropic-video.py
```diff
@@ -48,30 +53,32 @@ async def fetch_user_image(params: FunctionCallParams):
...
 
-    await params.result_callback(None)
-
-    # Instead of None, it's possible to also provide a tool call answer to
-    # tell the LLM that we are grabbing the image to analyze.
-    # await params.result_callback({"result": "Image is being captured."})
+    # Call result_callback with deferred=True to signal that the result will
+    # be provided later (when the image is actually retrieved). This prevents
+    # the function call from completing immediately.
+    await params.result_callback(None, properties=FunctionCallResultProperties(deferred=True))
```
aconchillo: I was thinking this might be a bit complicated for the user. That is, they need to know:

1. Pass `result_callback` to `UserImageRequestFrame`
2. Call  `params.result_callback` with `properties=FunctionCallResultProperties(deferred=True)`

What if we just did (1). So:

```
await params.llm.push_frame(
        UserImageRequestFrame(
            user_id=user_id,
            text=question,
            append_to_context=True,
            function_name=params.function_name,
            tool_call_id=params.tool_call_id,
            result_callback=params.result_callback,
        ),
        FrameDirection.UPSTREAM,
    )
```

Then, in `LLMService` we always wait a certain timeout for the `params.result_callback()` to be called. Not only for the image, but always.

Because today we already have the problem that a user can forget to call `params.result_callback()` and we have no way to know that.

So, my suggestion:

- You can push `UserImageRequestFrame` with `result_callback` and that's it.
- Or you can call `params.result_callback()` as we have always done. But not both.
- We add a timeout to the `LLMService` constructor, maybe `function_call_timeout`? Set to something big like 10 or 20 seconds.
- We always wait for that timeout.
- If the timeout expires we send a FunctionCallResultFrame with `TIMEOUT`.
- We log a warning saying the function call timed out.

src/pipecat/frames/frames.py
```diff
@@ -881,10 +881,14 @@ class FunctionCallResultProperties:
...
         run_llm: Whether to run the LLM after receiving this result.
         on_context_updated: Callback to execute when context is updated.
+        deferred: Whether this result is deferred and will be completed later.
+        deferred_timeout: Timeout in seconds for deferred results (auto-completes if exceeded).
     """
 
     run_llm: Optional[bool] = None
     on_context_updated: Optional[Callable[[], Awaitable[None]]] = None
+    deferred: Optional[bool] = None
+    deferred_timeout: float = 10  # 10 seconds default timeout for deferred results
```
aconchillo: These changes would not be needed with my proposal. The `deferred_timeout` would be the `function_call_timeout` in `LLMService`.

examples/foundational/14d-function-calling-anthropic-video.py
filipi87: I think it makes sense. I had thought about an approach like that, but I tried to respect what we already had as much as possible.

If we think that a single, global timeout for all function calls would be enough, and that we could defer all function calls to respect this timeout, that would make things a lot easier.

filipi87: Another possible approach I considered was to specify in the `ToolsSchema` whether a function call should be deferred or not, along with the timeout, when defining the function. We also have the direct function call spec, so we would need to add support for this there as well.

In this case, we would have more flexibility to specify if the function call is deferred or not, and a different timeout for each kind of function call.

src/pipecat/frames/frames.py

examples/foundational/14d-function-calling-anthropic-video.py
filipi87: Done!

aconchillo (approved):

aconchillo:
Looks great! We need a changelog file.

PR #3573, opened 2026-01-27
feat(google): add image_url support for Gemini adapter

maximizeschneider:
## Summary

- Enable external URLs for images in Gemini API using `Part.from_uri()`
- Previously `image_url` content type was logged as unsupported, but Google now allows external URLs for file input

Tested with Gemini from Vertex AI.

Reference: https://ai.google.dev/gemini-api/docs/file-input-methods


aconchillo (approved):

aconchillo:
LGTM!

PR #3574, opened 2026-01-27
Fix WebsocketService infinite loop on graceful server disconnect

### LINKED ISSUE #3539: event loop deadlock on WebsocketService when remote server closes connection
### pipecat version

latest master

### Python version

3.12

### Operating System

linux

### Issue description

When the remote server for a processor that uses WebsocketService disconnects gracefully, this triggers an infinite loop that doesn't yield, stopping up the entire event loop.

In WebsocketService we have the following code:
https://github.com/pipecat-ai/pipecat/blob/main/src/pipecat/services/websocket_service.py#L169
```
        while True:
            try:
                await self._receive_messages()
            except ...
```

In this case, when self._receive_messages() completes normally, it will be called again immediately.  But when we look at the implementation of _receive_messages, we see (equivalent across most subclasses of WebsocketService):

https://github.com/pipecat-ai/pipecat/blob/main/src/pipecat/services/elevenlabs/stt.py#L706

```
    async def _receive_messages(self):
        async for message in self._get_websocket():
            ....
```

But if we look at the [docs for the websockets package](https://websockets.readthedocs.io/en/stable/reference/asyncio/client.html#websockets.asyncio.client.ClientConnection.__aiter__)

We see

> It exits when the connection is closed normally. It raises a [ConnectionClosedError](https://websockets.readthedocs.io/en/stable/reference/exceptions.html#websockets.exceptions.ConnectionClosedError) exception after a protocol error or a network failure.

This means that if the remote server closed the connection gracefully, ```async for message in self._get_websocket():``` will simply exit immediately, exiting _receive_messages as well.  This entire loop doesn't yield control back to the event loop either, and has no way to reconnect.

### Reproduction steps

We saw the issue specifically on the elevenlabs frameprocessor but it's likely similarly possible on other processors as well.

We don't have a runnable repro to share, but this should cover the steps
```
# create a task to check if the event loop is still live
async def monitor_routine():
    while True:
        print('we are alive still')
        await asyncio.sleep(1)
monitor_task = asyncio.create_task(monitor_routine())

# create a pipeline
stt = ElevenLabsRealtimeSTTService(
    api_key=api_key,
    sample_rate=8000,
    audio_passthrough=False,
    params=ElevenLabsRealtimeSTTService.InputParams(),
)

# create the pipeline
pipeline = Pipeline([stt])

# run the pipeline and start it
```

After connecting to the elevenlabs servers, if we have a gap where we don't send in audio frames, the elevenlabs server will close the connection.

When this happens the deadlock is triggered, and the asyncio event loop stops working.

### Expected behavior

It seems the expected behavior is that the websocket should reconnect

### Actual behavior

asyncio event loop is broken, and websocket does not connect.

### Logs

```shell

```

markbackman:
## Summary

- Fixed infinite loop in `WebsocketService` that blocked the event loop when a remote server closed the connection gracefully
- When a websocket server sends a close frame, the `async for` loop exits normally without raising an exception. The `while True` loop in `_receive_task_handler` would immediately call `_receive_messages()` again, creating a tight loop with no `await` that starved the event loop.
- The fix detects normal return from `_receive_messages()` and either breaks (if intentionally disconnecting) or attempts reconnection

## Fixes

- Fixes #3539, which includes some more description as well as a repro case that I used to test the fix.

---

I also opted to consolidate the reconnecting logic into the `_maybe_try_reconnect`. This is an internal API, so it shouldn't cause any issues (though adds minor complexity).

### DIFF
--- a/changelog/3574.fixed.md
+++ b/changelog/3574.fixed.md
@@ -1,0 +1,1 @@
+- Fixed an infinite loop in `WebsocketService` that blocked the event loop when a remote server closed the connection gracefully.
--- a/src/pipecat/services/websocket_service.py
+++ b/src/pipecat/services/websocket_service.py
@@ -123,26 +123,29 @@ class WebsocketService(ABC):
 
     async def _maybe_try_reconnect(
         self,
-        error: Exception,
         error_message: str,
         report_error: Callable[[ErrorFrame], Awaitable[None]],
+        error: Optional[Exception] = None,
     ) -> bool:
         """Check if reconnection should be attempted and try if appropriate.
 
         Args:
-            error: The exception that occurred.
             error_message: Human-readable error message for logging.
             report_error: Callback function to report connection errors.
+            error: The exception that occurred (optional, may be None for graceful closes).
 
         Returns:
             True if should continue the receive loop, False if should break.
         """
         # Don't reconnect if we're intentionally disconnecting
         if self._disconnecting:
-            logger.warning(f"{self} error during disconnect: {error}")
+            if error:
+                logger.warning(f"{self} error during disconnect: {error}")
+            else:
+                logger.debug(f"{self} receive loop ended during disconnect")
             return False
 
-        # Log the error
+        # Log the message
         logger.warning(error_message)
 
         # Try to reconnect if enabled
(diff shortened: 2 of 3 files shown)


aconchillo (approved):

aconchillo:
LGTM!

PR #3575, opened 2026-01-27
Emit on_assistant_turn_stopped and on_user_turn_stopped from EndFrame…

### LINKED ISSUE #3564: Turn events not firing on task cancellation & Missing timestamps in AssistantTurnStoppedMessage
### pipecat version

0.0.100

### Python version

3.13.2

### Operating System

Docker Image: FROM python:3.13.2-slim

### Issue description

I am using **pipecat-ai version 0.0.100** with **FastAPIWebsocketTransport** for a Voice Agent using **Twilio**. I have encountered two distinct issues regarding the `LLMContextAggregatorPair`:

1. **Missing Timestamp**  
   The `message.timestamp` in `AssistantTurnStoppedMessage` is filled an empty string sometimes.

2. **Events Dropped on Disconnect**  
   If the call is manually disconnected while the user or AI is speaking, the final turn events (`on_user_turn_stopped` or `on_assistant_turn_stopped`) are never fired.  
   This results in the last spoken segment being recorded in audio but missing from transcription logs.

### Setup details

- **Library:** `pipecat-ai==0.0.100`
- **Transport:** `FastAPIWebsocketTransport`
- **Aggregator Strategy:** `LocalSmartTurnAnalyzerV3`

### My Event Handlers

```python
@user_aggregator.event_handler("on_user_turn_stopped")
async def on_user_turn_stopped(aggregator, strategy, message: UserTurnStoppedMessage):
    print("💸 User turn stopped:", message)


@assistant_aggregator.event_handler("on_assistant_turn_stopped")
async def on_assistant_turn_stopped(aggregator, message: AssistantTurnStoppedMessage):
    print("💥 Assistant turn stopped:", message)
````

These are the handlers I'm using that fail to capture the final turn on disconnect (and where I see the empty timestamp for the assistant).

---

## Steps for Issue 1 (Timestamp)

1. Run a normal conversation flow.
2. Inspect the `message` object inside `on_assistant_turn_stopped`.

---

## Steps for Issue 2 (Disconnect)

1. Start a voice session.
2. While the User or the Assistant is actively speaking, trigger a client disconnect.
3. The transport handler cancels the running task, effectively killing the pipeline before the aggregator processes the final partial turn:

```python
@transport.event_handler("on_client_disconnected")
async def on_client_disconnected(
    transport: FastAPIWebsocketTransport,
    client: WebSocket
):
    await task.cancel()
```

---

## Aggregator Configuration

```python
if use_external_strategies:
    turn_strategies = ExternalUserTurnStrategies()
else:
    # Using LocalSmartTurnAnalyzerV3
    turn_strategies = UserTurnStrategies(
        stop=[
                turn_analyzer=LocalSmartTurnAnalyzerV3()
            )
        ]
    )

    context,
        user_turn_strategies=turn_strategies,
        user_mute_strategies=[
            USER_MUTE_STRATEGY_MAP[strategy]()
            for strategy in (user_mute_strategies or [])
        ],
        user_idle_timeout=user_idle_timeout,
    ),
)
```

### Expected behavior

* **Timestamp:**
  `AssistantTurnStoppedMessage` should always contain a valid timestamp string.

* **Disconnect:**
  When `task.cancel()` is called or the pipeline shuts down, the aggregator should flush any incomplete turns and fire the corresponding `on_user_turn_stopped` or `on_assistant_turn_stopped` events for the final segment.

### Actual behavior

* **Timestamp:**
  The timestamp value is often an empty string `""`.

* **Disconnect:**
  The event handlers are not triggered for the final turn. The task cancellation seems to stop execution before the aggregator can process or report the final incomplete turn.

### Logs

```shell

```

markbackman:
… or CancelFrame

### Summary

- Fixed `LLMUserAggregator` and `LLMAssistantAggregator` not emitting pending transcripts when the conversation ends (`EndFrame`) or is cancelled (`CancelFrame`)
- Added `_maybe_emit_user_turn_on_end` helper method to consolidate user turn emission logic
- Added `_handle_end_or_cancel` handler for assistant aggregator to emit pending assistant transcripts

### Testing

Run [foundational example 28](https://github.com/pipecat-ai/pipecat/blob/main/examples/foundational/28-user-assistant-turns.py) and leave the session while either the bot or user is speaking; confirm that the events are emitted with user or assistant information, respectively.

This commit also updates the `uv.lock` for speechmatics and ai-coustics. I can separate out, but for simplicity, I just left it in this commit.

Fixes #3564

### DIFF
--- a/src/pipecat/processors/aggregators/llm_response_universal.py
+++ b/src/pipecat/processors/aggregators/llm_response_universal.py
@@ -464,9 +464,11 @@ class LLMUserAggregator(LLMContextAggregator):
             await s.setup(self.task_manager)
 
     async def _stop(self, frame: EndFrame):
+        await self._maybe_emit_user_turn_stopped(on_session_end=True)
         await self._cleanup()
 
     async def _cancel(self, frame: CancelFrame):
+        await self._maybe_emit_user_turn_stopped(on_session_end=True)
         await self._cleanup()
 
     async def _cleanup(self):
@@ -602,20 +604,33 @@ class LLMUserAggregator(LLMContextAggregator):
         if params.enable_user_speaking_frames:
             await self.broadcast_frame(UserStoppedSpeakingFrame)
 
-        # Always push context frame.
-        aggregation = await self.push_aggregation()
-
-        message = UserTurnStoppedMessage(
-            content=aggregation, timestamp=self._user_turn_start_timestamp
-        )
-        await self._call_event_handler("on_user_turn_stopped", strategy, message)
-        self._user_turn_start_timestamp = ""
+        await self._maybe_emit_user_turn_stopped(strategy)
 
     async def _on_user_turn_stop_timeout(self, controller):
         await self._call_event_handler("on_user_turn_stop_timeout")
 
     async def _on_user_turn_idle(self, controller):
         await self._call_event_handler("on_user_turn_idle")
+
+    async def _maybe_emit_user_turn_stopped(
+        self,
+        strategy: Optional[BaseUserTurnStopStrategy] = None,
+        on_session_end: bool = False,
+    ):
+        """Maybe emit user turn stopped event.
+
+        Args:
+            strategy: The strategy that triggered the turn stop.
+            on_session_end: If True, only emit if there's unemitted content
+                (avoids duplicate events when session ends).
+        """
+        aggregation = await self.push_aggregation()
+        if not on_session_end or aggregation:
+            message = UserTurnStoppedMessage(
+                content=aggregation, timestamp=self._user_turn_start_timestamp
+            )
+            await self._call_event_handler("on_user_turn_stopped", strategy, message)
+            self._user_turn_start_timestamp = ""
 
 
 class LLMAssistantAggregator(LLMContextAggregator):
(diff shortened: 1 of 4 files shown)


src/pipecat/processors/aggregators/llm_response_universal.py
```diff
@@ -464,11 +464,23 @@ async def _start(self, frame: StartFrame):
...
 
+    async def _maybe_emit_user_turn_on_end(self):
+        """Emit pending user turn on end/cancel if there's unemitted content."""
+        if len(self._aggregation) > 0:
+            aggregation = await self.push_aggregation()
+            message = UserTurnStoppedMessage(
+                content=aggregation, timestamp=self._user_turn_start_timestamp
+            )
+            await self._call_event_handler("on_user_turn_stopped", None, message)
+            self._user_turn_start_timestamp = ""
```
aconchillo: I think this could be `_maybe_emit_user_turn_stopped` (and move it at the end of the class) and also call it from `_on_user_turn_stopped` as well. And probably:

```
            aggregation = await self.push_aggregation()
            if aggregation:
                message = UserTurnStoppedMessage(
                    content=aggregation, timestamp=self._user_turn_start_timestamp
                )
                await self._call_event_handler("on_user_turn_stopped", None, message)
                self._user_turn_start_timestamp = ""
```

aconchillo (approved):

aconchillo:
LGTM! Just one comment

src/pipecat/processors/aggregators/llm_response_universal.py
markbackman: I tried this exact approach before, but we need to emit the `on_user_turn_stopped` event even when there is no transcription available.

Specifically, during the normal flow of conversation, the `on_user_turn_stopped` event is always emitted, regardless of the existence of a transcript. At the end of the session, we only want to emit it when there is an aggregation to push. That would mean that we'd have to give `_maybe_emit_user_turn_stopped` some state from the caller. I didn't love that approach, but if you want to consolidate the logic, we can do that.

PR #3581, opened 2026-01-28
Fixed race condition in OpenAIRealtimeLLMService

filipi87:
Fixed race condition in `OpenAIRealtimeLLMService` that could cause an error when truncating the conversation.


filipi87:
Same fix as we did here for the OpenAIRealtimeBetaLLMService: https://github.com/pipecat-ai/pipecat/pull/3567/

markbackman (approved):
LGTM

PR #3582, opened 2026-01-28
rename DAILY_SAMPLE_ROOM_URL to DAILY_ROOM_URL

aconchillo:

This has been bothering me for a while.


markbackman (approved):
LGTM

This is a breaking change for things, but it's a test utility, so maybe not as big of a deal?

aconchillo:
> LGTM
> 
> This is a breaking change for things, but it's a test utility, so maybe not as big of a deal?

I think we are the only ones that use this :-D

markbackman:
> > LGTM
> > This is a breaking change for things, but it's a test utility, so maybe not as big of a deal?
> 
> I think we are the only ones that use this :-D

Probably not, but I think it's OK. It's not breaking for Pipecat code. It's an env change.

PR #3583, opened 2026-01-28
VAD analyzer is now passed to LLMUserAggregator

aconchillo:
## Summary

- Move VAD analysis out of the transport layer and into the pipeline by introducing `VADController` and `VADProcessor`
- `LLMUserAggregatorParams` now accepts a `vad_analyzer` parameter, allowing the `LLMUserAggregator` to handle VAD internally via `VADController`
- Added `VADProcessor` as a standalone pipeline processor that pushes `VADUserStartedSpeakingFrame`, `VADUserStoppedSpeakingFrame`, and `UserSpeakingFrame` — useful for advanced pipelines with multiple aggregators (e.g. parallel LLM evaluation)
- Deprecated `vad_analyzer` on `BaseInputTransport`
- Updated all foundational examples to use the new approach

## Breaking Changes

- `vad_analyzer` on `BaseInputTransport` (and transport params) is now deprecated. Pass `vad_analyzer` to `LLMUserAggregatorParams` instead, or use `VADProcessor` in the pipeline for advanced use cases.

## Testing

- `uv run pytest tests/test_vad_controller.py`
- `uv run pytest tests/test_vad_processor.py`

### DIFF
--- a/src/pipecat/processors/audio/vad_processor.py
+++ b/src/pipecat/processors/audio/vad_processor.py
@@ -1,0 +1,100 @@
+#
+# Copyright (c) 2024-2026, Daily
+#
+# SPDX-License-Identifier: BSD 2-Clause License
+#
+
+"""Voice Activity Detection processor for detecting speech in audio streams.
+
+This module provides a VADProcessor that wraps a VADController to process
+audio frames and push VAD-related frames into the pipeline.
+"""
+
+from typing import Type
+
+from loguru import logger
+
+from pipecat.audio.vad.vad_analyzer import VADAnalyzer
+from pipecat.audio.vad.vad_controller import VADController
+from pipecat.frames.frames import (
+    Frame,
+    UserSpeakingFrame,
+    VADUserStartedSpeakingFrame,
+    VADUserStoppedSpeakingFrame,
+)
+from pipecat.processors.frame_processor import FrameDirection, FrameProcessor
+
+
+class VADProcessor(FrameProcessor):
+    """Processes audio frames through voice activity detection.
+
+    This processor wraps a VADController to detect speech in audio streams
+    and push VAD frames into the pipeline:
+
+    - ``VADUserStartedSpeakingFrame``: Pushed when speech begins.
+    - ``VADUserStoppedSpeakingFrame``: Pushed when speech ends.
+    - ``UserSpeakingFrame``: Pushed periodically while speech is detected.
+
+    Example::
+
+        vad_processor = VADProcessor(vad_analyzer=SileroVADAnalyzer())
+    """
+
+    def __init__(
+        self,
+        *,
+        vad_analyzer: VADAnalyzer,
+        speech_activity_period: float = 0.2,
+        **kwargs,
+    ):
+        """Initialize the VAD processor.
+
+        Args:
+            vad_analyzer: The VADAnalyzer instance for processing audio.
+            speech_activity_period: Minimum interval in seconds between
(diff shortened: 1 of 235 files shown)


src/pipecat/transports/base_input.py
```diff
@@ -48,7 +48,7 @@
 from pipecat.processors.frame_processor import FrameDirection, FrameProcessor
 from pipecat.transports.base_transport import TransportParams
 
-AUDIO_INPUT_TIMEOUT_SECS = 0.5
+AUDIO_INPUT_TIMEOUT_SECS = 1.0
```
markbackman, line 51: Why the value change?

src/pipecat/transports/base_input.py
```diff
@@ -373,70 +405,56 @@ async def _cancel_audio_task(self):
...
+        # Skip timeout handling until the first audio frame arrives (e.g. client
+        # not yet connected).
+        audio_received = False
         while True:
             try:
                 frame: InputAudioRawFrame = await asyncio.wait_for(
                     self._audio_in_queue.get(), timeout=AUDIO_INPUT_TIMEOUT_SECS
                 )
 
+                # From now one, timeout should warn if there's no audio.
```
markbackman: Nit: Should be `From now on`.

src/pipecat/processors/audio/vad_processor.py
```diff
@@ -0,0 +1,100 @@
...
+        super().__init__(**kwargs)
+        self._vad_controller = VADController(
+            vad_analyzer, speech_activity_period=speech_activity_period
+        )
+
+        # Push VAD frames when speech events are detected
+        @self._vad_controller.event_handler("on_speech_started")
+        async def on_speech_started(_controller):
+            logger.debug(f"{self}: User started speaking")
+            await self.push_frame(VADUserStartedSpeakingFrame())
```
markbackman: broadcast_frame()? Same for others in this class.

markbackman (commented):
Generally looks good. Testing:

- Tested the 07 example as is and works well.
- Tested a modified 07 with a VADProcessor and I'm not seeing interruptions. From a quick investigation, it looks like the VAD frames aren't making it into the user turn strategies when the VADProcessor is used. The root cause appears to be that the `_handle_speech_control_params` function is called when the `SpeechControlParamsFrame` frame is handled. That incorrectly sets the turn strategy to External and emits this warning:

```
2026-01-29 17:02:56.705 | WARNING  | pipecat.processors.aggregators.llm_response_universal:_handle_speech_control_params:554 - LLMUserAggregator#0: `turn_analyzer` in base input transport is deprecated. Use `LLMUserAggregator`'s new `user_turn_strategies` parameter with `TurnAnalyzerUserTurnStopStrategy` instead:

        context,
            ...,
                stop=[
                        turn_analyzer=LocalSmartTurnAnalyzerV3(params=SmartTurnParams())
                    )
                ],
            )
        ),
    )
```

src/pipecat/transports/base_input.py
aconchillo: I thought 0.5 is too short if there's a network blip.

src/pipecat/processors/audio/vad_processor.py
aconchillo: oops! done!

src/pipecat/transports/base_input.py
aconchillo: Done!

markbackman (approved):
LGTM! Those last fixes have this working well. Thanks for the clean up 👏

Issue #1842, opened 2025-05-19
TTS and LLM function calls are not synched


itsikbelson-spotlight:
### pipecat version

0.0.67

### Python version

3.12.9

### Operating System

macOS 14.4.1

### Issue description

When my LLM response includes text and function call, the function call execution time will not be correlated with the TTS output.
While there are cases that this may be expected (e.g. when function is intended to fetch data for the LLM to use), there are cases where it would be better to execute the function in sync with the TTS (e.g. when you want trigger an operation, such as turning on lights, after the bot **said** "Let there be light !"

From looking [here](https://github.com/pipecat-ai/pipecat/blob/2b66eddaa1a0e93049c3942cfea69168b0e8c2f5/src/pipecat/services/tts_service.py#L180), it seems the TTS layer should pause frame processing exactly for this reason.
However `FunctionCallInProgressFrame`, `FunctionCallCancelFrame` and `FunctionCallResultFrame` are defined as `SystemFrame` and do not wait for TTS to speak.

I wonder if there is a good reason to always treat functions that way.

### Reproduction steps

Create a call to the LLM which returns text and a function call (e.g. "Let there be light", followed by a function call to turn on the light)
You can use the code below and say "Turn on lights" to see the output is printed before the bots finished speaking.

```
async def turn_on_light(params: FunctionCallParams):
    print("!!!!! Lights are on !!!!!")
    await params.result_callback(f"Done")


async def main():
    async with aiohttp.ClientSession() as session:
        (room_url, _) = await configure(session) 

        # Initialize services
        transport = DailyTransport(
            room_url,
            None,
            "Test",
            DailyParams(
                audio_in_enabled=True,
                audio_out_enabled=True,
                audio_in_passthrough=True,
                vad_analyzer=SileroVADAnalyzer(),
            ),
        )

        stt = DeepgramSTTService(api_key=os.getenv("DEEPGRAM_API_KEY"))
        tts = ElevenLabsTTSService(
            api_key=os.getenv("ELEVENLABS_API_KEY"), voice_id="XrExE9yKIg1WjnnlVkGX"
        )
        turn_on_light_function = FunctionSchema(
            name="turn_on_light",
            description="Turn on light",
            properties={},
            required=[],
        )
        tools = ToolsSchema(standard_tools=[turn_on_light_function])

        llm = GoogleLLMService(
            api_key=os.getenv("GOOGLE_API_KEY"),
            model="gemini-2.0-flash-001",
[...]

PR #3584, opened 2026-01-28
TTS services improvements.

filipi87:
**Summary**                                                                                                                                    
- Added TTS context tracking system to trace audio generation through the pipeline
- Introduced `TTSContext` class and unique `context_id` for each TTS request
- Added `context_id` field to all TTS-related frames (`TTSAudioRawFrame`, `TTSStartedFrame`, `TTSStoppedFrame`, `AggregatedTextFrame`, `TTSTextFrame`)
- Added `append_to_context` parameter to `TTSSpeakFrame` for conditional LLM context addition
- Created `on_tts_request` event handler for monitoring TTS requests
- Refactored all 30+ TTS service implementations to support context tracking
- Removed unnecessary `self._started` fields from TTS services
- Simplified context aggregators, `_handle_text()` to only check `frame.append_to_context` instead of also checking `self._started`

**Breaking Changes**
- `TTSService.run_tts()` now requires a `context_id` parameter
- Subclasses implementing `run_tts()` must update their method signature to include `context_id: str`

  ---


markbackman:
@filipi87 I started work on this here:
https://github.com/pipecat-ai/pipecat/pull/3465

Aleix raised this issue that prompted me to stop work on this:
https://github.com/pipecat-ai/pipecat/pull/3465#discussion_r2696581006

filipi87:
@markbackman, @aconchillo,

I am halfway through validating that all the TTS services are working after this refactor. I am only missing tests for 12 more TTS services.

However, I believe this PR already addresses the main concern Aleix mentioned in the other PR:
https://github.com/pipecat-ai/pipecat/pull/3465/changes#r2696581006

If you could already provide a review for this one, it would be greatly appreciated.

I will also bring in any other commits that Mark made there which I haven’t addressed in this PR yet.

filipi87:
@markbackman, @aconchillo, I think it’s now fully ready for review.
I’ll add the changelog later, once we’re happy with all the changes here.

src/pipecat/services/tts_service.py
```diff
@@ -255,13 +257,14 @@ def set_voice(self, voice: str):
 
     # Converts the text to audio.
     @abstractmethod
-    async def run_tts(self, text: str) -> AsyncGenerator[Frame, None]:
+    async def run_tts(self, text: str, context_id: str) -> AsyncGenerator[Frame, None]:
```
markbackman, line 291: Should `context_id` be `Optional` here? Without that, this is a breaking change for `run_tts()`. While you've covered all cases, there are developers that have built their own services that we should support.

filipi87: Makes sense. 👍

filipi87: @Mark, to be honest, I am having second thoughts about making this optional. 

I know this would allow TTS services in third party projects to remain compatible with the current signature, but it would also introduce a downside: all the TTS services here would need to add extra logic to handle an optional `context_id` even though, in practice, it would never be optional.

So I think we should introduce this breaking change now, before 1.0.0. All the TTS services inside Pipecat will already be refactored to handle this anyway.

cc: @aconchillo for thoughts.

src/pipecat/services/resembleai/tts.py
filipi87, line 1: This one I have not been able to test. I am always receiving this error, even on `main`:
`Error in receive loop: no close frame received or sent`

filipi87: It is working with the new api keys.

itissid:
I think one thing where its not stated and should be added to the review title is some User Stories this can support? 

Perhaps its so obvious to everyone but not me, is how this will help in correlating TTSSpeakFrames from an observer or a frame processor with what was actually spoken. The closest thing is AggregateTextFrame, but those are pushed before anything is queued to a TTS service like cartesia. The best application of this I have are filler phrases emitted after a user's speech is complete is to reduce perceived latency very typically due to waiting for LLM inference or the subsequent function calls. 

1. User says "Hi I need an appointment"
2. LLMContextFrame sends this to LLM
3. LLM provide starts completion. 
4. LLM service streams tool calls and executes any number of functions for those calls, adds the results of those to the assistant context aggregator. Function Frames are emitted.
5. LLM service uses the results of the executed functions to queue LLMTextFrames which are processed into AggregatedTextFrames and sent to the TTSService. 
6. TTSTextFrames, TTSStart/Stop frames are are emmited when as the bot speech occurs.


Between 1, 6 it could be several seconds of delay in some cases. Fillers can be really good UX. Several implementations I have seen have various drawbacks:

- Pre Actions aren't a good fit because they are executed before the LLM inference but delays can still happen in the LLM and functions. 
- My current "safest" solution to this is creating AggregateTextFrames like so to track them from originiating TTSpeak:
```python
class CorrelatingCartesiaTTSService(CartesiaTTSService):
    async def process_frame(self, frame, direction: FrameDirection):
        # Preserve metadata across TTSSpeakFrame -> AggregatedTextFrame
        if isinstance(frame, TTSSpeakFrame):
            logger.debug(
                f"[CorrelatingCartesiaTTSService] Processing TTSSpeakFrame id={frame.id} with text: {frame.text}"
            )
            # Mirror base behavior: store/restore processing state, pause, flush
            processing_text = getattr(self, "_processing_text", False)

            agg = AggregatedTextFrame(frame.text, AggregationType.SENTENCE)

            # Copy any metadata (correlation ID, kind, etc.)
            if getattr(frame, "metadata", None):
                agg.metadata.update(frame.metadata)

            # Optional: record original speak frame id for debugging
            agg.metadata.setdefault("lf_src_speak_id", getattr(frame, "id", None))

            # Now do what base TTSService would do, but with metadata-preserving agg
            await self._push_tts_frames(agg)
            await self._maybe_pause_frame_processing()
            await self.flush_audio()
            self._processing_text = processing_text
            return

        # Default path for everything else
        return await super().process_frame(frame, direction)
```

- People have been queuing TTSSpeakFrame from functions itself, but this is also not perfect because the LLMResponse has to be awaited.

src/pipecat/services/tts_service.py
```diff
@@ -653,9 +673,17 @@ async def _push_tts_frames(
             if aggregation_type == type or aggregation_type == "*":
                 transformed_text = await transform(transformed_text, type)
 
+        # Create context ID and store metadata
+        context_id = str(uuid.uuid4())
```
itissid: Actually I think if you expose this context_id creation as an overidable method/callback people can add their own context IDs. IIUC if you use uuid directly here it here there is no way to know externally what this ID was for for the text that was asked to be emitted. Right?

itissid: I think it would also solve this core limitation that @markbackman mentioned here: https://github.com/pipecat-ai/pipecat/issues/1842#issuecomment-2901648231 and make that feature easily implementable in a number of ways. e.g. I could specify  the thing that executes LLM function calls can be given an a optional async task which makes it wait on this ID generation function to be called. Once called the LLM function execution logic can capture that ID and wait for those TTSTextFrame, TTSStopFrame with that ID for a configurable timeout and if there is no speech they can  resume execution.

itissid: I would be interested in knowing what do you think about passing this id to the aggregate context frame above? I think for both the sources of the AggregateTextFrame — created by the LLMTextProcesssor  or from TTSpeakFrames —  it would make the client's job easier to correlate what was sent to be spoken and what actually was by these frames if you stuffed the context_id in the AggregateTextFrame before emitting it?

src/pipecat/processors/aggregators/llm_response_universal.py
```diff
@@ -809,7 +810,6 @@ def __init__(
                     DeprecationWarning,
                 )
 
-        self._started = 0
```
aconchillo, line 798: why are we removing this?

src/pipecat/services/asyncai/tts.py
aconchillo: we might not need this now.

src/pipecat/services/azure/tts.py
```diff
@@ -617,9 +625,10 @@ async def run_tts(self, text: str) -> AsyncGenerator[Frame, None]:
             try:
                 if not self._started:
                     await self.start_ttfb_metrics()
-                    yield TTSStartedFrame()
+                    yield TTSStartedFrame(context_id=context_id)
                     self._started = True
                     self._first_chunk = True
+                    self._current_context_id = context_id
```
aconchillo: I think we can get rid of `_started` and check `self._current_context_id` instead.

src/pipecat/services/asyncai/tts.py
aconchillo: Check other services as well.

src/pipecat/services/tts_service.py
aconchillo: @filipi87 and I have discussed. It certainly is a breaking change and it was a concern for me as well, but I think TTS integrations will be eager to update. Developers using previous Pipecat versions they can pin previous versions of their TTS integration.

src/pipecat/services/tts_service.py
aconchillo: @itissid That's a good idea. @filipi87 can you add a function to create the UUID, something like `def create_context_id() -> str`? 🙏 

Also, it might be a good idea to also trigger an event, for example `on_tts_request` and pass both the context ID and the text. So people would get:

```
@tts.event_handler("on_tts_request")
async def on_tts_request(tts, context_id: str, text: str):
    ....
```

@itissid I believe you could implement the logic you are mentioning with this without needing to control the UUID. But the function I think makes sense as well.

aconchillo: Yep, I think this is a good idea as well. 🙌

src/pipecat/services/tts_service.py
```diff
@@ -209,6 +210,7 @@ def __init__(
         self._stop_frame_queue: asyncio.Queue = asyncio.Queue()
 
         self._processing_text: bool = False
+        self._tts_contexts: Dict[str, dict] = {}
```
aconchillo: Create a dataclass as we spoke.

src/pipecat/services/deepgram/tts.py
aconchillo: As we discussed, save `context_id` so we can always pass it here. In theory this works, because we are pausing frame processing and there shouldn't be more than one context id at a time.

src/pipecat/services/azure/tts.py
aconchillo: Also, double check other services.

aconchillo:
> I think one thing where its not stated and should be added to the review title is some User Stories this can support?
> 

The goals of this PR are:

- Be able to conditionally add `TTSSpeakFrame` to the context, without needing to sandwich it between `LLMFullResponseStartFrame` and `LLMFullResponseEndFrame`.
- Everything should have a context ID, whether the TTS service supports it or not. We are planning to do more internal work/cleanup and this PR will make that easier.
- Context ID and output frame correlation.
- Concurrent TTS calls. This is only possible on the services that support context IDs, but even in those cases we still currently don't support it as we assume just one internal ID, and that make us need to pause frames. This will come in a separate PR.

src/pipecat/services/tts_service.py
filipi87: Yep, I think it makes sense, and like you said, I believe this solves the needs that @itissid mentioned. 

I’ve implemented both. Thank you for the suggestions. 🙂

src/pipecat/services/tts_service.py
filipi87: Done.

src/pipecat/services/deepgram/tts.py
filipi87: Yep, done.

src/pipecat/processors/aggregators/llm_response_universal.py
filipi87: Before, text was only added to the context if it occurred between `LLMFullResponseStartFrame` and `LLMFullResponseEndFrame`, and that was why we had `self._started`.

Now we are relying only on `frame.append_to_context` to decide whether the text should be appended to the context.

src/pipecat/services/asyncai/tts.py
filipi87: Done.

src/pipecat/services/azure/tts.py
filipi87: Done.

src/pipecat/services/tts_service.py
filipi87: Ok, I also created the `context_id` in the `AggregatedTextFrame`, and I’m setting it to the same `context_id`. 👍

src/pipecat/processors/aggregators/llm_response_universal.py
aconchillo: Can you double-check that things still work if we push TTSSpeakFrame inside a function call?

aconchillo (approved):

aconchillo:
LGTM! Nice work @filipi87 ! :rocket:

PR #3585, opened 2026-01-29
improve Piper TTS support

aconchillo:
## Summary

- Added local `PiperTTSService` for offline TTS using Piper voice models (auto-downloads models on first use)
- Renamed the existing HTTP-based service to `PiperHttpTTSService` and added `voice_id` parameter support
- Added automatic audio resampling in `TTSService._stream_audio_frames_from_iterator()` — parses sample rate from WAV headers or accepts an explicit `in_sample_rate`, resampling when the source rate differs from the output rate
- Fixed `PiperHttpTTSService` to correctly resample audio based on the model's native sample rate

## Breaking Changes

- `PiperTTSService` has been renamed to `PiperHttpTTSService`. The `PiperTTSService` name is now used for the new local Piper service.


markbackman (approved):
LGTM!

PR #3590, opened 2026-01-29
runner: allow custom CLI arguments

aconchillo:
## Summary

- `main()` now accepts an optional `argparse.ArgumentParser`, allowing bots to define custom CLI arguments that are accessible via `runner_args.cli_args`
- Added `cli_args` field to `RunnerArguments` so all runner argument types carry parsed CLI args
- Refactored internal route setup functions to pass `args: Namespace` instead of unpacking individual keyword arguments
- Updated `18-gstreamer-filesrc.py` example to use the new custom parser pattern

## Example

```python
    input_file = runner_args.cli_args.input
    ...

if __name__ == "__main__":
    from pipecat.runner.run import main

    parser = argparse.ArgumentParser()
    parser.add_argument("-i", "--input", required=True)
    main(parser)
```

### DIFF
--- a/changelog/3590.added.md
+++ b/changelog/3590.added.md
@@ -1,0 +1,1 @@
+- `main()` in `pipecat.runner.run` now accepts an optional `argparse.ArgumentParser`, allowing bots to define custom CLI arguments accessible via `runner_args.cli_args`.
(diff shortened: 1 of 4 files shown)


markbackman (approved):
LGTM

Issue #3538, opened 2026-01-23
Inworld TTS


vsridharan6:
### pipecat version

0.0.99

### Python version

3.13.2

### Operating System

MacOS 15.4.1

### Issue description

Noticed this a couple times, 
Once it did not stream TTS at all. 
This time it just clipped in the middle. 
```

2026-01-22 21:23:52.538 | DEBUG    | pipecat.services.deepgram.stt:_disconnect:268 - Disconnecting from Deepgram
2026-01-22 21:23:52.539 | DEBUG    | pipecat.services.inworld.tts:run_tts:226 - InworldHttpTTSService#0: Generating TTS [Thank you for your help—have a great day. Goodbye.] (streaming=True)
2026-01-22 21:23:52.573 | DEBUG    | pipecat.processors.aggregators.llm_response_universal:_handle_function_call_in_progress:771 - LLMAssistantAggregator#0 FunctionCallInProgressFrame: [end_call:call_8Dkflofyt6Hua8q7cSH3BC81]
2026-01-22 21:23:52.573 | DEBUG    | pipecat.processors.aggregators.llm_response_universal:_handle_function_call_result:802 - LLMAssistantAggregator#0 FunctionCallResultFrame: [end_call:call_8Dkflofyt6Hua8q7cSH3BC81]
2026-01-22 21:23:53.581 | DEBUG    | pipecat.processors.metrics.frame_processor_metrics:stop_ttfb_metrics:131 - InworldHttpTTSService#0 TTFB: 1.0423510074615479
2026-01-22 21:23:53.613 | DEBUG    | pipecat.transports.base_output:_bot_started_speaking:608 - Bot started speaking


 --> TTS just cut off Thank you of your. - 

**2026-01-22 21:23:53.731 | ERROR    | pipecat.processors.frame_processor:push_error_frame:725 - InworldHttpTTSService#0 exception (/Users/vensridh/Dev/leka-labs/voice-pipecat/.venv/lib/python3.13/site-packages/pipecat/services/inworld/tts.py:294): Unknown error occurred: 'utf-8' codec can't decode byte 0xc9 in position 1023: unexpected end of data**


2026-01-22 21:23:53.731 | DEBUG    | pipecat.processors.metrics.frame_processor_metrics:stop_processing_metrics:152 - InworldHttpTTSService#0 processing time: 1.1921789646148682
[...]

PR #3591, opened 2026-01-29
Fix WebSocket reconnection races and Inworld TTS context restore

hwuiwon:
This PR improves reliability for websocket-based services during transient disconnects.

  - WebsocketService: serialize reconnect attempts with an asyncio.Lock and short-circuit if the connection is already
    healthy (reduces “reconnect attempt aborted / unable to reconnect” races).
  - InworldTTSService: if a reconnect happens while a context is active, re-send the create message for the existing
    contextId, and only close_context / clear contextId on intentional disconnects (helps prevent “connected but no
    audio / context missing” after reconnects).


hwuiwon:
@markbackman @cshape would appreciate a review!

markbackman:
@hwuiwon we haven't seen this case. We've also just made changes to the WebsocketService to make it more robust to reconnects. I'm hesitant to make further changes for this release unless we have a documented issue and repro case that occurs on `main`. If you have that, then we're happy to take a look. I just want to be cautious because changes here affect so many services.

hwuiwon:
@markbackman I’ve been seeing multiple cases like https://github.com/pipecat-ai/pipecat/issues/3538#issuecomment-3819636305 while running hundreds of calls through the Inworld WebSocket service. It doesn’t happen consistently, but it reliably shows up under higher call volume. I would suggest making 1000+ calls that uses inworld websocket service to reproduce this. Not so sure how testing is set up for testing this service in particular from pipecat side.

It is critical because this prevents the speech from getting generated and users won't get any feedbacks.

markbackman:
> @markbackman I’ve been seeing multiple cases like [#3538 (comment)](https://github.com/pipecat-ai/pipecat/issues/3538#issuecomment-3819636305) while running hundreds of calls through the Inworld WebSocket service. It doesn’t happen consistently, but it reliably shows up under higher call volume. I would suggest making 1000+ calls that uses inworld websocket service to reproduce this. Not so sure how testing is set up for testing this service in particular from pipecat side.
> 
> It is critical because this prevents the speech from getting generated and users won't get any feedbacks.

Have you been running from `main`? Also, is the issue with the Inworld configuration or others? We have customers that do 10K+ calls daily on services like Cartesia and ElevenLabs and they haven't reported issues.

hwuiwon:
I don't believe it's the configuration issue with Inworld because it works normally with the http service. I can rollback the changes in the `src/pipecat/services/websocket_service.py` but I think we need fix in the `src/pipecat/services/inworld/tts.py`

ianbbqzy:
> if a reconnect happens while a context is active, re-send the create message for the existing
contextId

I think this is worth a try. I noticed that Cartesia doesn't have a separate CreateContext message because they expect the configs in every message. On the other hand, 11lab might also not necessitate a createContext message because a lot of configs were already sent as part of query params in the URL. 

In the case of Inworld, on Websocket reconnection, it's important to re create a context first

hwuiwon:
@markbackman rolled back the changes in `websocket_service.py`. Fix only touches Inworld Websocket TTS now

ianbbqzy:
@hwuiwon @markbackman , I took a peak at this issue myself. I'm curious if the extra complexity is because `InworldAITtsService` uses `send_with_retry` which isn't used by elevenlabs or Cartesia. Maybe in case of disconnect, we should just create a new context instead of trying to replay messages that failed to send (just like the other providers do). Then we don't have to worry about re-creating the previous context?

markbackman:
> Maybe in case of disconnect, we should just create a new context instead of trying to replay messages that failed to send (just like the other providers do). Then we don't have to worry about re-creating the previous context?

That sounds like a nice simplification!

markbackman:
Tagging @filipi87 for this one since I'll be out for a few days. And, @filipi87 has been working on context improvements for TTS services. This may already be handled or he may have ideas.

filipi87 (commented):
In version `0.0.102`, we refactored how we handle TTS contexts.

`TTSService.run_tts()` now requires a `context_id` parameter for context tracking.

All 30+ TTS service implementations were updated to support context tracking via `context_id`.

We also have another improvement in progress, refactoring all TTS services that use `AudioContextTTSService`, which is the case for `InworldTTSService`. You can take a look at it here:
- https://github.com/pipecat-ai/pipecat/pull/3732/

Even after these changes land, I believe we may still face the issue reported here. If there is a connection problem, we may not receive the audio back from the Inworld, becauseas you mentioned, unlike Cartesia, this implementation requires a separate message to create the context.

Additionally, even with the changes in this PR, if an issue occurs inside `_send_text`, we would not receive audio for that text, and it would be lost.

At this point, we have two possible options:
- Accept this behavior and simply throw an error, which I believe is the current approach here in this PR.
- Keep `_send_text` using `send_with_retry`, but also check upon reconnection whether there is a context in progress. If there is, we would need to recreate the context in Inworld using the same `context_id`. This way it will match the same `context_id` inside the message that we are trying to send.

In any case, I believe we need to wait for this PR (https://github.com/pipecat-ai/pipecat/pull/3732/) to land first. Since it is already refactoring this same code. 

After that, we can address the reported issue using one of the alternatives mentioned above.

markbackman:
@filipi87 is the issue raised here still relevant? If not, is it included in the Inworld refactoring that you're doing with the Inworld team? Just wondering if this is still relevant or if it should be closed out.

hwuiwon:
I haven't seen this recently, but not sure if it was because there was some fix from the inworld side or is it because of the override that I'm doing from my side

markbackman:
I'm closing this out. We worked closely with the Inworld team and made some solid improvements. If there's still an issue, please let us know!

PR #3593, opened 2026-01-29
Add auto_mode support for inworld plugin

ianbbqzy:
Inworld auto_mode to improve first token latency.

### DIFF
--- a/changelog/3593.added.md
+++ b/changelog/3593.added.md
@@ -1,0 +1,2 @@
+- Added support for Inworld TTS Websocket Auto Mode for improved latency
+- Updated timestamps to be cumulative within an agent turn, using flushCompleted message as an indication of when timestamps from the server are reset to 0
--- a/changelog/3593.added.md
+++ b/changelog/3593.added.md
@@ -1,2 +1,1 @@
 - Added support for Inworld TTS Websocket Auto Mode for improved latency
-- Updated timestamps to be cumulative within an agent turn, using flushCompleted message as an indication of when timestamps from the server are reset to 0
--- a/src/pipecat/services/inworld/tts.py
+++ b/src/pipecat/services/inworld/tts.py
@@ -414,6 +414,11 @@ class InworldTTSService(AudioContextWordTTSService):
             apply_text_normalization: Whether to apply text normalization.
             max_buffer_delay_ms: Maximum buffer delay in milliseconds.
             buffer_char_threshold: Buffer character threshold.
+            auto_mode: Whether to use auto mode. Recommended when texts are sent
+                in full sentences/phrases. When enabled, the server controls
+                flushing of buffered text to achieve minimal latency while
+                maintaining high quality audio output. If None (default),
+                automatically set based on aggregate_sentences.
         """
 
         temperature: Optional[float] = None
@@ -421,6 +426,7 @@ class InworldTTSService(AudioContextWordTTSService):
         apply_text_normalization: Optional[str] = None
         max_buffer_delay_ms: Optional[int] = None
         buffer_char_threshold: Optional[int] = None
+        auto_mode: Optional[bool] = None
 
     def __init__(
         self,
@@ -432,6 +438,7 @@ class InworldTTSService(AudioContextWordTTSService):
         sample_rate: Optional[int] = None,
         encoding: str = "LINEAR16",
         params: InputParams = None,
+        aggregate_sentences: bool = True,
         **kwargs: Any,
     ):
         """Initialize the Inworld WebSocket TTS service.
@@ -444,6 +451,7 @@ class InworldTTSService(AudioContextWordTTSService):
             sample_rate: Audio sample rate in Hz.
             encoding: Audio encoding format.
             params: Input parameters for Inworld WebSocket TTS configuration.
+            aggregate_sentences: Whether to aggregate sentences before synthesis.
             **kwargs: Additional arguments passed to the parent class.
         """
         super().__init__(
@@ -451,6 +459,7 @@ class InworldTTSService(AudioContextWordTTSService):
             push_stop_frames=True,
             pause_frame_processing=True,
             sample_rate=sample_rate,
+            aggregate_sentences=aggregate_sentences,
             **kwargs,
         )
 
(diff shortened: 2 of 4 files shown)


markbackman:
Tagging @cshape for review.

@ianbbqzy can you link to the docs for this? I don't see this feature listed.

cshape:
> Tagging @cshape for review.
> 
> @ianbbqzy can you link to the docs for this? I don't see this feature listed.

It's a new feature which will be going out next week. @ianbbqzy and I will link the docs, review/test and ping you once we're ready then. Thanks!

markbackman:
> It's a new feature which will be going out next week. @ianbbqzy and I will link the docs, review/test and ping you once we're ready then. Thanks!

Sounds good!

cshape:
Hi @markbackman -

Chatted with @ianbbqzy and this feature (basically a smart auto-flushing mechanism to get optimal latency and prosody from our TTS with websockets) is good to merge now from our perspective.

It's currently slated to be rolled out on our side later this week, but will be safely ignored if the Pipecat SDK includes it before then. Our preference is to merge it into the Pipecat SDK ASAP so users can benefit as soon as our release goes out.

Thanks!

markbackman (approved):
Thanks for confirming, @cshape!

Can you please add a changelog to this PR and then it'll be ready to go:

ianbbqzy:
>Can you please add a changelog to this PR and then it'll be ready to go:
@markbackman Done! thanks!

ianbbqzy:
@markbackman sorry, I realized that this auto_mode could exacerbate the timestamps resetting issue. I will work on a fix and report back

ianbbqzy:
Implemented timestamp cumulation using `flushCompleted` response message as the indication that timestamps will reset to 0.

src/pipecat/services/inworld/tts.py
```diff
@@ -414,13 +414,19 @@ class InputParams(BaseModel):
             apply_text_normalization: Whether to apply text normalization.
             max_buffer_delay_ms: Maximum buffer delay in milliseconds.
             buffer_char_threshold: Buffer character threshold.
+            auto_mode: Whether to use auto mode. Recommended when texts are sent
```
markbackman, line 417: How does this parameter work with manual flush requests from application code?

The thing is: Pipecat knows before Inworld that all text has been sent to Inworld. It's the best party to say when to flush. The reason being that Pipecat receives a stream of tokens from the LLM, including a token that indicates the end of stream. Shouldn't this be the default mechanism to flush?

ianbbqzy: It depends on what you mean by "manual flush":
1. pipecat's `flush_audio()` method
2. Sending `{flush_context: {}}` to Inworld

Previously doing 2 is what lets Inworld know that a complete sentence/phrase has been sent so that Inworld can start TTS generation. But this has the side effect of creating flush boundaries between sentences, causing there to be some discontinuity. With `auto_mode`, pipecat no longer has to do 2, and Inworld can have the liberty to combine sentences in the same generation to achieve better continuity while maintaining the lowest latency possible.

But when pipecat calls `flush_audio` at the end of turn, we send a `flush_context` to inworld anyway just as a safety measure.

IIUC Inworld's `flush` is implemented in a similar way as that of elevenlabs and Cartesia. Please let me know if you think this is not true and I have a wrong understanding.

markbackman: I mean Pipecat sending a flush message to Inworld, e.g. `{"flush_context": {}, "contextId": context_id}`.

Let me briefly explain how things work so we can figure out the right thing to do there:

- The LLM streams tokens to the TTS service.
- The TTS service assembles those tokens into sentences and sends them sentence-by-sentence to Inworld.
- Once the LLM stream is finished, the LLM notifies the TTS service of this fact and the TTS service will flush audio by sending `{"flush_context": {}, "contextId": context_id}` to Inworld.

So, that means the messages to Inworld are as follows:
- Sentence 1: `Hello!`
- Sentence 2: `I'm an AI assistant.`
- Sentence 3: `How can I help you today?`
- Flush: `{"flush_context": {}, "contextId": context_id}`

The flush message acts as a final indicator to Inworld to tell it to send everything it has.

Given this, what's the right thing to do? It seems helpful that Pipecat indicates to Inworld that the audio generations for a given context are finished and that audio can be finalized and sent.

ianbbqzy: From Inworld's perspective, wed on't really want to wait for the end of the LLM stream to finish to "finalize" an audio. With auto_mode, the idea is that we want to start TTS generation ASAP for low latency, So in this example, Inworld might decide to start generating "Hello!" immediately. We expect pipecat to send complete sentence so we don't try to generate an incomplete phrase like "I'm an...".

Then, While "Hello" is being generated, we can aggregate the later sentences for better continuity between sentences. 

We are working on improving continuity between generations as well. But currently we want to help Pipecat users achieve low latency first. While continuity will be improved overtime. auto_mode helps with subsequent sentences aggregation for now.

markbackman: Ok, I see. This is more of an internal setting. So, this isn't exclusive of manual flushing, but is used in addition to it. Is that right? If so, like you said, this is similar to the same feature in 11Labs, right?

changelog/3593.added.md
```diff
@@ -0,0 +1,2 @@
+- Added support for Inworld TTS Websocket Auto Mode for improved latency
```
markbackman, line 1: Mind breaking this into two changelog fragments.

3593.added.md:
```
- Added support for Inworld TTS Websocket Auto Mode for improved latency
```

3593.change.md:
```
- Updated timestamps to be cumulative within an agent turn, using flushCompleted message as an indication of when timestamps from the server are reset to 0
```

markbackman:
Thanks for patiently educating me on this. This does sound like a nice improvement. In running the example, I see really solid TTFB performance now. Kudos!

Before this is ready, just two small things:
1. Lint the code (either `uv run scripts/fix-ruff.sh` or install pre-commit hook: `uv run pre-commit install`)
2. Update the changelog per my comment.

ianbbqzy:
>Thanks for patiently educating me on this. This does sound like a nice improvement. In running the example, I see really solid TTFB performance now. Kudos!

No problem at all and thanks for your team's and your personal insights into Voice AI orchestration in general. There might still be some rough edges with the Inworld API, but we will keep improving it to truly show the potentials of our model.

>solid TTFB performance now

Thanks! but we also know it's not there yet, we have an even bigger latency improvement regarding timestamps coming soon next week, stay tuned!

PR #3594, opened 2026-01-29
UserTurnController: reset user turn timeout with interim transcriptions

aconchillo:
## Summary

- Fixed `UserTurnController` to also reset the user turn timeout on `InterimTranscriptionFrame`, not just `TranscriptionFrame`. Previously, if a user was speaking and only interim transcriptions arrived, the turn timeout could fire prematurely.


markbackman (approved):
LGTM

PR #3600, opened 2026-01-30
LLMService: make sure function call timeout handler is started

aconchillo:
## Summary

- Fixed "coroutine was never awaited" warning for the function call timeout handler in `LLMService`
- Moved timeout task creation after `function_call_result_callback` definition to avoid forward-reference issues
- Added `await asyncio.sleep(0)` after task creation to ensure the timeout coroutine is entered by the event loop before it could be cancelled
- Cancel the timeout task on exception in the function call handler to prevent it from firing after an error


markbackman (approved):
LGTM

PR #3605, opened 2026-01-31
GeminiLiveLLMService: let the transcription timeout handler be scheduled

aconchillo:
## Summary

- Fixed `GeminiLiveLLMService` transcription timeout handler not being properly scheduled by adding an `await asyncio.sleep(0)` after task creation, allowing the event loop to schedule the task before it potentially gets cancelled.


markbackman:
LGTM. Thanks for the fix.

PR #3607, opened 2026-01-31
Optimize PatternPairAggregator with incremental parsing

KYash03:
Replaced the regex loop in PatternPairAggregator with a single-pass parser. It’s significantly faster and consumes less memory. Verified backward compatibility with existing tests!


markbackman:
This change is pretty significant. Is there any way to make it more narrowly focused on this?
- Improved `PatternPairAggregator` performance by using incremental delimiter detection.

KYash03:
I've trimmed the diff to remove unrelated changes. The core change is replacing regex-based re.finditer scanning (O(n²)) with incremental delimiter detection (O(n)) — that's inherently a full rewrite of the matching logic, so the diff can't get much smaller without artificially splitting it into separate PRs. Happy to do that if you'd prefer.

Issue #3552, opened 2026-01-24
GeminiLiveLLMService: EndFrame gets stuck when called from function call with run_llm=False


markbackman:
### pipecat version

0.0.100

### Python version

Any

### Operating System

Any

### Issue description

When an EndTaskFrame is pushed upstream in a function call where the result_callback is configured for run_llm=False, the EndFrame is blocked.

### Reproduction steps

Run example 26i with the end_conversation function configured as follows:
```
async def end_conversation(params: FunctionCallParams):
    properties = FunctionCallResultProperties(run_llm=False)
    await params.result_callback({"success": True}, properties=properties)
    await params.llm.push_frame(EndTaskFrame(), FrameDirection.UPSTREAM)
```

### Expected behavior

Session should end and all processes shutdown.

### Actual behavior

The EndFrame is blocked by the GeminiLiveLLMService, resulting in a hung pipeline.

### Logs

```shell

```

erikrubinov:
In some cases, it also works despite setting run_llm=False by adjusting the system prompt. In the case of example 26i, it works if I remove these lines from the system prompt:


After you've responded to the user three times, do two things, in order:
1. Politely let them know that that's all the time you have today and say goodbye.
2. *WITHOUT WAITING FOR THE USER TO RESPOND*, call the end_conversation tool to gracefully end the conversation.

markbackman:
Resolved here:
https://github.com/pipecat-ai/pipecat/pull/4125

PR #3609, opened 2026-01-31
gemini-live: end response segment before running tool calls

KartikVashishta:
Addressing #3552, 

Gemini Live pauses generation while a tool call is outstanding and does (may) not emit turn_complete until tool responses are provided. If a tool returns run_llm=False and ends the task, we could keep the bot marked as responding and defer EndFrame indefinitely.

Close out the current response segment before dispatching tool calls (clear buffers, stop TTS when appropriate, emit LLMFullResponseEndFrame) so EndFrame can propagate and the pipeline shuts down cleanly


markbackman:
Thanks for contributing! Sorry we didn't get to this. Issue #3552 was resolved by https://github.com/pipecat-ai/pipecat/pull/4125. Closing.

PR #3610, opened 2026-01-31
PipelineTask: don't add RTVIObserver if already there

aconchillo:
## Summary

- `PipelineTask` now skips adding a default `RTVIObserver` if the provided observers list already contains one, preventing duplicate RTVI observers.


markbackman (approved):
LGTM

aconchillo:
We will now have error and warning logs.

PR #3612, opened 2026-02-01
KokoroTTSService: use kokoro-onnx instead of kokoro

aconchillo:
## Summary

- Switched `KokoroTTSService` from the `kokoro` package to `kokoro-onnx` for local TTS synthesis
- `kokoro` was using loguru and removing any handlers which made it hard for applications to integrate
- Model files are now auto-downloaded from GitHub releases to `~/.cache/kokoro-onnx/` instead of using Hugging Face
- Uses the async streaming API from `kokoro-onnx` for audio generation
- Re-enabled Kokoro eval test (`07zj-interruptible-kokoro.py`)

## Breaking Changes

- The `repo_id` parameter in `KokoroTTSService.__init__()` has been replaced with `model_path` and `voices_path` parameters

## Testing

- Run the Kokoro interruptible example: `python examples/07zj-interruptible-kokoro.py`


markbackman (approved):
LGTM

PR #3617, opened 2026-02-02
Fix sentence splitting for CJK and other non-Latin languages

jamsea:
## Summary

- NLTK's `sent_tokenize()` only supports ~15 European languages and defaults to English. For Japanese, Chinese, Korean, Hindi, Arabic, and other non-Latin scripts, NLTK fails to recognize sentence boundaries (e.g., `。`, `？`, `！`), causing TTS text to accumulate until flush instead of being emitted sentence-by-sentence.
- Added a fallback in `match_endofsentence()` that scans for unambiguous non-Latin sentence-ending punctuation when NLTK returns the entire text as a single sentence. Latin punctuation (`. ! ? ; …`) is excluded from the fallback since NLTK handles those correctly and they can be ambiguous (abbreviations, decimals, etc.).
- Added unit tests for CJK/Indic/Arabic with lookahead characters, Latin regression safety, and full aggregator integration tests for Japanese and Chinese streaming.
## Test plan

- [x] All existing `test_utils_string.py` tests pass (no regression for English abbreviations, decimals, etc.)
- [x] All existing `test_simple_text_aggregator.py` tests pass (no regression for English streaming)
- [x] New `test_endofsentence_cjk_with_lookahead` covers Japanese, Chinese, Korean, Indic, Arabic
- [x] New `test_endofsentence_latin_not_affected_by_fallback` confirms no Latin regression
- [x] New aggregator tests cover Japanese multi-sentence, Japanese lookahead, Chinese token streaming, Japanese single-sentence flush
- [x] Manual test with `GoogleHttpTTSService` using Japanese text to confirm sentence-by-sentence TTS output

### DIFF
--- a/changelog/3617.fixed.md
+++ b/changelog/3617.fixed.md
@@ -1,0 +1,5 @@
+- Fixed sentence splitting for Japanese, Chinese, Korean, and other non-Latin
+  languages in TTS pipeline. NLTK's sentence tokenizer does not support CJK
+  languages, causing text to accumulate until flush instead of being split at
+  sentence boundaries. Added fallback detection for unambiguous non-Latin
+  sentence-ending punctuation (e.g., `。`, `？`, `！`).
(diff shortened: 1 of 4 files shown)


markbackman (approved):
LGTM! Nice solution to this problem 🙌

PR #3620, opened 2026-02-02
A couple of features made while exploring Gemini Live <-> Pipecat Flows interop

kompfner:
This PR adds a couple of features needed while exploring support for Gemini Live <-> Pipecat Flows interop. They are:

* Adding `LLMMessagesTransformFrame` to facilitate programmatically editing context in a frame-based way.
* Added support to Gemini Live for programmatically swapping tools or editing context at runtime, e.g. with `LLMMessagesAppendFrame`, `LLMMessagesUpdateFrame`, `LLMMessagesTransformFrame`, and `LLMSetToolsFrame`

See CHANGELOG entries for more detail.

Note that we're actually not (or not yet) officially supporting Gemini Live in Pipecat Flows, but these features are generally useful regardless.

Pipecat Flows companion PR, made during the exploration: https://github.com/pipecat-ai/pipecat-flows/pull/232.

Why not already officially support Gemini Live in Pipecat Flows?
* The need to reconnect on context and tools updates makes Gemini Live not the most natural fit for Flows, performing notably slower than cascaded text-to-text pipelines during node transitions
* Because of this, the remaining investment to fully support Gemini Live with Flows (rethinking Flows node action timings for realtime models, in addition to ongoing long-term maintenance costs) doesn't—at least today—clear the necessary priority bar


changelog/3620.added.md
```diff
@@ -0,0 +1 @@
+- Added support to Gemini Live for programmatically swapping tools or editing context at runtime; now you can use `LLMMessagesAppendFrame`, `LLMMessagesUpdateFrame`, and `LLMSetToolsFrame` with Gemini Live, like you would with text-to-text services. Importantly, this change enables you to use Gemini Live with Pipecat Flows, if you're using Pipecat Flows 0.0.23 or above.
```
kompfner: TODO: update the Pipecat Flows version mentioned here if it turns out the necessary Flows changes don't make it in that version.

markbackman (commented):
## Code review

Found 1 issue. Checked for bugs and CLAUDE.md compliance.

src/pipecat/services/google/gemini_live/llm.py
```diff
@@ -826,6 +833,9 @@ async def set_context(self, context: OpenAILLMContext):
             logger.error("Context already set. Can only set up Gemini Live context once.")
             return
         self._context = GeminiLiveContext.upgrade(context)
+        self._context_snapshot = copy.deepcopy(
+            self._context
```
markbackman, line 837: **Bug:** When the deprecated `set_context()` method is used (which creates a `GeminiLiveContext` via `GeminiLiveContext.upgrade()`), `self._context_snapshot` becomes a deep copy of a `GeminiLiveContext` instance. `GeminiLiveContext` extends `OpenAILLMContext`, which does **not** have a `diff()` method — `diff()` is only defined on [`LLMContext`](https://github.com/pipecat-ai/pipecat/blob/b77604953512f240c2301f557074577bc2a4e0b0/src/pipecat/processors/aggregators/llm_context.py#L459-L531).

When a subsequent `LLMContextFrame` arrives through the pipeline and hits the `else` branch of [`_handle_context()`](https://github.com/pipecat-ai/pipecat/blob/b77604953512f240c2301f557074577bc2a4e0b0/src/pipecat/services/google/gemini_live/llm.py#L1033-L1040), the call to `self._context_snapshot.diff(self._context)` on [line 1037](https://github.com/pipecat-ai/pipecat/blob/b77604953512f240c2301f557074577bc2a4e0b0/src/pipecat/services/google/gemini_live/llm.py#L1036-L1038) will raise `AttributeError: 'GeminiLiveContext' object has no attribute 'diff'`.

This only affects the deprecated `set_context()` path (since the normal pipeline flow via `_handle_context()` stores an `LLMContext` snapshot which does have `diff()`), and the PR description notes the new functionality only works with `LLMContext`/`LLMContextAggregatorPair`. However, the crash is still reachable at runtime if someone uses the deprecated API.

src/pipecat/adapters/services/gemini_adapter.py
```diff
@@ -668,3 +681,43 @@ def _thought_signature_inline_data_bookmark_matches_part(
...
+
+        Returns:
+            List of Content messages with function-related parts converted to text.
+        """
+        converted_messages = []
+        for msg in messages:
+            if msg.parts:
+                converted_parts = []
+                for part in msg.parts:
+                    if func_call := getattr(part, "function_call", None):
```
filipi87, line 699: I guess the function call also has an ID. Should we try to preserve it as well so we can properly match each response to its corresponding call ?

src/pipecat/frames/frames.py
```diff
@@ -667,9 +667,16 @@ class LLMContextFrame(Frame):
 
     Parameters:
         context: The LLM context containing messages, tools, and configuration.
+        messages_programmatically_edited: Whether the context messages were
```
filipi87, line 670: I wonder whether this should be part of the frame, or instead live inside the context. 

In that case, we could check whether the context has been programmatically modified since a specific timestamp, for example.

As you said, this is something we could improve both here and in the context summarization. 

Another related thought is whether we should add some kind of lock to the context, or another mechanism, to prevent race conditions. Like, preventig both user aggregator and system aggregator changing the context at the same time.

src/pipecat/frames/frames.py
```diff
@@ -822,6 +829,25 @@ class LLMMessagesUpdateFrame(DataFrame):
     run_llm: Optional[bool] = None
 
 
+@dataclass
+class LLMMessagesTransformFrame(DataFrame):
```
filipi87, line 833: Are we already pushing this new frame anywhere?

src/pipecat/processors/aggregators/llm_context.py
```diff
@@ -341,6 +374,19 @@ def set_messages(self, messages: List[LLMContextMessage]):
         """
         self._messages[:] = messages
 
+    def transform_messages(
```
filipi87, line 377: I believe maybe we will need to add some kind of lock to the context, or another mechanism to prevent race conditions.

What do you think ?

If it makes sense, we can add it in a follow up PR.

src/pipecat/processors/aggregators/llm_context.py
```diff
@@ -410,3 +456,137 @@ def _normalize_and_validate_tools(tools: ToolsSchema | NotGiven) -> ToolsSchema
             raise TypeError(
                 f"In LLMContext, tools must be a ToolsSchema object or NOT_GIVEN. Got type: {type(tools)}",
             )
+
+    def diff(self, other: "LLMContext") -> LLMContextDiff:
```
filipi87, line 460: Inside `utils/context/`, I’ve added `llm_context_summarization.py`.

I wonder if we should do the same here and move all the diff logic into something like `llm_context_diff.py`.

src/pipecat/processors/aggregators/llm_response_universal.py
```diff
@@ -231,7 +232,16 @@ def _get_context_frame(self) -> LLMContextFrame:
...
-        return LLMContextFrame(context=self._context)
+        # Check if messages were programmatically edited since the last push.
+        # This flag is stored as a runtime attribute on the shared context
+        # object so that both user and assistant aggregators can see it.
+        messages_programmatically_edited = getattr(
+            self._context, "_pipecat_messages_programmatically_edited", False
+        )
+        return LLMContextFrame(
+            context=self._context,
+            messages_programmatically_edited=messages_programmatically_edited,
```
filipi87, line 243: Not needed if we decide to check the context instead to determine whether it has been programmatically changed.

src/pipecat/processors/aggregators/llm_response_universal.py
```diff
@@ -458,13 +482,17 @@ async def process_frame(self, frame: Frame, direction: FrameDirection):
             await self._handle_llm_messages_append(frame)
         elif isinstance(frame, LLMMessagesUpdateFrame):
             await self._handle_llm_messages_update(frame)
+        elif isinstance(frame, LLMMessagesTransformFrame):
+            await self._handle_llm_messages_transform(frame)
```
filipi87, line 486: Is there a reason we decided to add this here instead of inside the Assistant Aggregator?

I was having a similar discussion with @aconchillo yesterday, and we concluded that the UserAggregator should only handle user related stuff, such as turn strategies, idle, and mute.

Given that, this logic feels like it should live inside the Assistant Aggregator instead. I’ve moved all the summarization logic there as well. 😅

src/pipecat/processors/aggregators/llm_response_universal.py
```diff
@@ -573,11 +601,28 @@ async def _handle_llm_run(self, frame: LLMRunFrame):
 
     async def _handle_llm_messages_append(self, frame: LLMMessagesAppendFrame):
         self.add_messages(frame.messages)
+        # Mark the context as programmatically edited. This flag is stored as a
+        # runtime attribute on the shared context object so that both user and
+        # assistant aggregators can see it.
+        self._context._pipecat_messages_programmatically_edited = True
```
filipi87, line 607: If we handle this inside the context, every time we invoke `add_messages`, `set_messages`, or `transform_messages`, we could store a timestamp of the last time the context was programmatically edited. 

We could then expose a method to check whether the context has been programmatically edited since a given point in time.

tropxy:
Hi, I know I am pressing a bit, but this is quite an urgent feature for us, at the moment I believe given the lack of this, the agent hallucinates quite easily.
If the team can put some prio on this, would be great! 
Thank you

kompfner:
I'm going to close this PR and instead open a more targeted one containing _just_ the `LLMMessagesTransformFrame`, to unblock [this Pipecat Flows change](https://github.com/pipecat-ai/pipecat-flows/pull/232) fix.

kompfner:
> given the lack of this, the agent hallucinates quite easily

@tropxy, I'd like to better understand what you mean by this—what feature in particular do you feel you need to address hallucinations?

I'm _guessing_ (but correct me if I'm wrong) that you're referring to the [associated Pipecat Flows work](https://github.com/pipecat-ai/pipecat-flows/pull/232) that depends on this PR? That work is meant to help LLMs better understand what tools are actually available to it at any given time, to avoid tool-related hallucinations.

kompfner:
Pardon the confusion, but we've actually closed the [Pipecat Flows PR](https://github.com/pipecat-ai/pipecat-flows/pull/232) dependent on this PR. We're going to instead pursue a fix that lives in base Pipecat itself, as "deactivating" functions is not just a Pipecat Flows concern.

PR #3621, opened 2026-02-02
Context summarization feature implementation

filipi87:
## Summary

- Added automatic context summarization to compress conversation history when token limits are approached, enabling efficient long running conversations
- Implemented token estimation using character-count heuristic (1 token ≈ 4 characters)
- Added smart message selection that preserves system messages, recent context, and incomplete function call sequences
- Configured via `enable_context_summarization=True` in `LLMUserAggregatorParams` with customizable thresholds and behavior
- Runs summarization asynchronously in background tasks to avoid blocking the pipeline

## Key Features

- **Automatic triggering**: Summarization triggers when context exceeds 80% of max tokens (default 8000) or after 20 unsummarized messages
- **Function call awareness**: Never summarizes incomplete tool call sequences, preserving request-response pairing integrity
- **Interruption handling**: Cancels pending summarizations when user interrupts to avoid stale results
- **Configurable preservation**: Keeps configurable number of recent messages (default: 4) uncompressed for immediate context

## Configuration

```python
from pipecat.processors.aggregators.llm_response_universal import (
    LLMUserAggregatorParams,
)
from pipecat.utils.context.llm_context_summarization import (
    LLMContextSummarizationConfig,
)

user_aggregator_params = LLMUserAggregatorParams(
    enable_context_summarization=True,
    context_summarization_config=LLMContextSummarizationConfig(
        max_context_tokens=8000,           # Maximum context size
        summarization_threshold=0.8,        # Trigger at 80% of max
        max_unsummarized_messages=20,       # Or after 20 new messages
        min_messages_after_summary=4,       # Keep last 4 messages
        summarization_prompt=None,          # Optional custom prompt
    )
)
```

## Testing

Run the new test suite:
```bash
uv run pytest tests/test_context_summarization.py
```

Try the examples:
```bash
# OpenAI example with function calling
uv run examples/foundational/54-context-summarization-openai.py

# Google Gemini example
uv run examples/foundational/54a-context-summarization-google.py
```

## Implementation Details

**New Components:**
- `src/pipecat/utils/context/llm_context_summarization.py`: Core utility with token estimation, message selection, and formatting
- `LLMContextSummaryRequestFrame` and `LLMContextSummaryResultFrame`: New control frames for async summarization flow
- `LLMContextSummarizationConfig`: Configuration dataclass with validation

**Modified Components:**
- `LLMUserAggregator`: Added summarization trigger logic, state tracking, and result handling
- `LLMService`: Added async summary generation using `run_inference()` with max_tokens override

### DIFF
--- a/.claude/skills/cleanup/SKILL.md
+++ b/.claude/skills/cleanup/SKILL.md
@@ -1,0 +1,307 @@
+# Code Cleanup Skill
+
+The **Code Cleanup Skill** reviews, refactors, and documents code changes in your current branch, ensuring alignment with **Pipecat’s architecture, coding standards, and example patterns**.
+It focuses on **readability, correctness, performance, and consistency**, while avoiding breaking changes.
+
+---
+
+## Skill Overview
+
+This skill analyzes all changes introduced in your branch and performs the following actions:
+
+1. **Analyze Branch Changes**
+   - Review uncommitted changes and outgoing commits
+2. **Refactor for Readability**
+   - Improve clarity, naming, structure, and modern Python usage
+3. **Enhance Performance**
+   - Identify safe, conservative optimization opportunities
+4. **Add Documentation**
+   - Apply Pipecat-style, Google-format docstrings
+5. **Ensure Pattern Consistency**
+   - Match existing Pipecat services, pipelines, and examples
+6. **Validate Examples**
+   - Ensure examples follow foundational patterns (e.g. `07-interruptible.py`)
+
+---
+
+## Usage
+
+Invoke the skill using any of the following commands:
+
+- “Clean up my branch code”
+- “Refactor the changes in my branch”
+- “Review and improve my branch code”
+- `/cleanup`
+
+---
+
+## What This Skill Does
+
+### 1. Analyze Branch Changes
+
+The skill retrieves all uncommitted changes and outgoing commits to understand:
+
+- New files added
+- Modified files
+- Code additions and deletions
+- Overall scope and intent of changes
+
+---
+
+### 2. Code Refactoring
+
+#### Readability Improvements
+
+- Replace tuples with named classes or dataclasses
+- Improve variable, method, and class naming
+- Extract complex logic into well-named helper methods
+- Add missing type hints
+- Simplify nested or complex conditionals
+- Replace deprecated methods and features
+- Normalize formatting to match Pipecat style
+
+#### Performance Enhancements
+
+- Identify inefficient loops or repeated work
+- Suggest appropriate data structures
+- Optimize async workflows and I/O
+- Remove redundant operations
+
+> Performance changes are conservative and non-breaking.
+
+---
+
+### 3. Documentation
+
+Documentation follows **Google-style docstrings**, consistent with Pipecat conventions.
+
+#### Class Documentation
+
+```python
+class ExampleService:
+    """Brief one-line description.
+
+    Detailed explanation of the class purpose, responsibilities,
+    and important behaviors.
+
+    Supported features:
+
+    - Feature 1
+    - Feature 2
+    - Feature 3
+    """
+```
+
+#### Method Documentation
+
+```python
+def process_data(self, data: str, options: Optional[dict] = None) -> bool:
+    """Process incoming data with optional configuration.
+
+    Args:
+        data: The input data to process.
+        options: Optional configuration dictionary.
+
+    Returns:
+        True if processing succeeded, False otherwise.
+
+    Raises:
+        ValueError: If data is empty or invalid.
+    """
+```
+
+#### Pydantic Model Parameters
+
+```python
+class InputParams(BaseModel):
+    """Configuration parameters for the service.
+
+    Parameters:
+        timeout: Request timeout in seconds.
+        retry_count: Number of retry attempts.
+        enable_logging: Whether to enable debug logging.
+    """
+
+    timeout: Optional[float] = None
+    retry_count: int = 3
+    enable_logging: bool = False
+```
+
+---
+
+### 4. Pattern Consistency Checks
+
+#### Service Classes
+
+- Correct inheritance (`TTSService`, `STTService`, `LLMService`)
+- Consistent constructor signatures
+- Frame emission patterns
+- Metrics support:
+  - `can_generate_metrics()`
+  - TTFB metrics
+  - Usage metrics
+- Alignment with similar existing services
+
+#### Examples
+
+Validated against `examples/foundational/07-interruptible.py`:
+
+- Proper `create_transport()` usage
+- Correct pipeline structure
+- Task setup and observers
+- Event handler registration
+- Runner and bot entrypoint consistency
+
+---
+
+### 5. Specific Implementation Patterns
+
+#### Service Implementation
+
+```python
+class ExampleTTSService(TTSService):
+
+    def __init__(self, *, api_key: Optional[str] = None, **kwargs):
+        super().__init__(**kwargs)
+        self._api_key = api_key or os.getenv("SERVICE_API_KEY")
+
+    def can_generate_metrics(self) -> bool:
+        return True
+
+    async def run_tts(self, text: str) -> AsyncGenerator[Frame, None]:
+        try:
+            await self.start_ttfb_metrics()
+            yield TTSStartedFrame()
+            # ... processing ...
+            yield TTSAudioRawFrame(...)
+        finally:
+            await self.stop_ttfb_metrics()
+```
+
+---
+
+#### Example Structure Pattern
+
+```python
+transport_params = {
+    "daily": lambda: DailyParams(...),
+    "twilio": lambda: FastAPIWebsocketParams(...),
+    "webrtc": lambda: TransportParams(...),
+}
+
+async def run_bot(transport: BaseTransport, runner_args: RunnerArguments):
+    stt = DeepgramSTTService(...)
+    tts = SomeTTSService(...)
+    llm = OpenAILLMService(...)
+
+    context = LLMContext(messages)
+    user_aggregator, assistant_aggregator = LLMContextAggregatorPair(...)
+
+    pipeline = Pipeline([...])
+    task = PipelineTask(pipeline, params=..., observers=[...])
+
+    @transport.event_handler("on_client_connected")
+    async def on_client_connected(transport, client):
+        await task.queue_frames([LLMRunFrame()])
+
+    runner = PipelineRunner(handle_sigint=runner_args.handle_sigint)
+    await runner.run(task)
+
+async def bot(runner_args: RunnerArguments):
+    """Main bot entry point compatible with Pipecat Cloud."""
+    transport = await create_transport(runner_args, transport_params)
+    await run_bot(transport, runner_args)
+```
+
+---
+
+## Execution Flow
+
+1. Fetch uncommitted and outgoing changes
+2. Categorize files (services, examples, tests, utilities)
+3. Analyze each file:
+   - Readability
+   - Performance
+   - Documentation
+   - Pattern consistency
+4. Generate actionable recommendations
+5. Apply Pipecat standards
+
+---
+
+## Examples
+
+### Before: Tuple Usage
+
+```python
+def get_audio_info(self) -> Tuple[int, int]:
+    return (48000, 1)
+```
+
+### After: Named Class
+
+```python
+class AudioInfo:
+    """Audio configuration information.
+
+    Parameters:
+        sample_rate: Sample rate in Hz.
+        num_channels: Number of audio channels.
+    """
+
+    sample_rate: int
+    num_channels: int
+
+def get_audio_info(self) -> AudioInfo:
+    return AudioInfo(sample_rate=48000, num_channels=1)
+```
+
+---
+
+### Before: Missing Documentation
+
+```python
+class NewTTSService(TTSService):
+    def __init__(self, api_key: str, voice: str):
+        self._api_key = api_key
+        self._voice = voice
+```
+
+### After: Fully Documented
+
+```python
+class NewTTSService(TTSService):
+    """Text-to-speech service using NewProvider API.
+
+    Streams PCM audio and emits TTSAudioRawFrame frames compatible
+    with Pipecat transports.
+
+    Supported features:
+    - Text-to-speech synthesis
+    - Streaming PCM audio
+    - Voice customization
+    - TTFB metrics
+    """
+
+    def __init__(self, *, api_key: str, voice: str, **kwargs):
+        """Initialize the NewTTSService.
+
+        Args:
+            api_key: API key for authentication.
+            voice: Voice identifier to use.
+            **kwargs: Additional arguments passed to the parent service.
+        """
+        super().__init__(**kwargs)
+        self._api_key = api_key
+        self.set_voice(voice)
+```
+
+---
+
+## Notes
+
+- Non-breaking improvements only
+- Backward compatibility preserved
+- Conservative performance changes
+- Google-style docstrings
+- Pattern checks follow recent Pipecat code
--- a/changelog/3621.added.md
+++ b/changelog/3621.added.md
@@ -1,0 +1,5 @@
+- Added context summarization feature to automatically compress conversation history when conversation length limits (by token or message count) are reached, enabling efficient long-running conversations.
+  - Configure via `enable_context_summarization=True` in `LLMAssistantAggregatorParams`
+  - Customize behavior with `LLMContextSummarizationConfig` (max tokens, thresholds, etc.)
+  - Automatically preserves incomplete function call sequences during summarization
+  - See new examples: `examples/foundational/54-context-summarization-openai.py` and `examples/foundational/54a-context-summarization-google.py`
--- a/examples/foundational/54-context-summarization-openai.py
+++ b/examples/foundational/54-context-summarization-openai.py
@@ -1,0 +1,188 @@
+#
+# Copyright (c) 2024-2026, Daily
+#
+# SPDX-License-Identifier: BSD 2-Clause License
+#
+
+"""Example demonstrating context summarization feature.
+
+This example shows how to enable and configure context summarization to automatically
+compress conversation history when token limits are approached. It also demonstrates
+that summarization correctly handles function calls, preserving incomplete function
+call sequences.
+"""
+
+import asyncio
+import os
+
+from dotenv import load_dotenv
+from loguru import logger
+
+from pipecat.adapters.schemas.function_schema import FunctionSchema
+from pipecat.adapters.schemas.tools_schema import ToolsSchema
+from pipecat.audio.turn.smart_turn.local_smart_turn_v3 import LocalSmartTurnAnalyzerV3
+from pipecat.audio.vad.silero import SileroVADAnalyzer
+from pipecat.audio.vad.vad_analyzer import VADParams
+from pipecat.frames.frames import LLMRunFrame
+from pipecat.pipeline.pipeline import Pipeline
+from pipecat.pipeline.runner import PipelineRunner
+from pipecat.pipeline.task import PipelineParams, PipelineTask
+from pipecat.processors.aggregators.llm_context import LLMContext
+from pipecat.processors.aggregators.llm_response_universal import (
+    LLMAssistantAggregatorParams,
+    LLMContextAggregatorPair,
+    LLMUserAggregatorParams,
+)
+from pipecat.runner.types import RunnerArguments
+from pipecat.runner.utils import create_transport
+from pipecat.services.cartesia.tts import CartesiaTTSService
+from pipecat.services.deepgram.stt import DeepgramSTTService
+from pipecat.services.llm_service import FunctionCallParams
+from pipecat.services.openai.llm import OpenAILLMService
+from pipecat.transports.base_transport import BaseTransport, TransportParams
+from pipecat.transports.daily.transport import DailyParams
+from pipecat.transports.websocket.fastapi import FastAPIWebsocketParams
+from pipecat.turns.user_stop import TurnAnalyzerUserTurnStopStrategy
+from pipecat.turns.user_turn_strategies import UserTurnStrategies
+from pipecat.utils.context.llm_context_summarization import LLMContextSummarizationConfig
+
+load_dotenv(override=True)
+
+# We use lambdas to defer transport parameter creation until the transport
+# type is selected at runtime.
+transport_params = {
+    "daily": lambda: DailyParams(
+        audio_in_enabled=True,
+        audio_out_enabled=True,
+    ),
+    "twilio": lambda: FastAPIWebsocketParams(
+        audio_in_enabled=True,
+        audio_out_enabled=True,
+    ),
+    "webrtc": lambda: TransportParams(
+        audio_in_enabled=True,
+        audio_out_enabled=True,
+    ),
+}
+
+
+# Tool functions for the LLM
+async def get_current_weather(params: FunctionCallParams):
+    """Get the current time in a readable format."""
+    logger.info("Tool called: get_current_weather")
+    await asyncio.sleep(1)  # Simulate some processing
+    await params.result_callback({"conditions": "nice", "temperature": "75"})
+
+
+async def run_bot(transport: BaseTransport, runner_args: RunnerArguments):
+    logger.info("Starting bot")
+
+    stt = DeepgramSTTService(api_key=os.getenv("DEEPGRAM_API_KEY"))
+
+    tts = CartesiaTTSService(
+        api_key=os.getenv("CARTESIA_API_KEY"),
+        voice_id="71a7ad14-091c-4e8e-a314-022ece01c121",  # British Reading Lady
+    )
+
+    llm = OpenAILLMService(api_key=os.getenv("OPENAI_API_KEY"))
+
+    # Register tool functions
+    llm.register_function("get_current_weather", get_current_weather)
+
+    weather_function = FunctionSchema(
+        name="get_current_weather",
+        description="Get the current weather",
+        properties={
+            "location": {
+                "type": "string",
+                "description": "The city and state, e.g. San Francisco, CA",
+            },
+            "format": {
+                "type": "string",
+                "enum": ["celsius", "fahrenheit"],
+                "description": "The temperature unit to use. Infer this from the user's location.",
+            },
+        },
+        required=["location", "format"],
+    )
+    tools = ToolsSchema(standard_tools=[weather_function])
+
+    messages = [
+        {
+            "role": "system",
+            "content": "You are a helpful LLM in a WebRTC call. Your goal is to demonstrate your capabilities in a succinct way. Your output will be spoken aloud, so avoid special characters that can't easily be spoken, such as emojis or bullet points. Respond to what the user said in a creative and helpful way. You have access to tools to get the current weather - use them when relevant.",
+        },
+    ]
+
+    context = LLMContext(messages, tools=tools)
+
+    # Create aggregators with summarization enabled
+    user_aggregator, assistant_aggregator = LLMContextAggregatorPair(
+        context,
+        user_params=LLMUserAggregatorParams(
+            user_turn_strategies=UserTurnStrategies(
+                stop=[TurnAnalyzerUserTurnStopStrategy(turn_analyzer=LocalSmartTurnAnalyzerV3())]
+            ),
+            vad_analyzer=SileroVADAnalyzer(params=VADParams(stop_secs=0.2)),
+        ),
+        assistant_params=LLMAssistantAggregatorParams(
+            enable_context_summarization=True,
+            # Optional: customize context summarization behavior
+            # Using low limits to demonstrate the feature quickly
+            context_summarization_config=LLMContextSummarizationConfig(
+                max_context_tokens=1000,  # Trigger summarization at 1000 tokens
+                target_context_tokens=800,  # Target context size for the summarization
+                max_unsummarized_messages=10,  # Or when 10 new messages accumulate
+                min_messages_after_summary=2,  # Keep last 2 messages uncompressed
+            ),
+        ),
+    )
+
+    pipeline = Pipeline(
+        [
+            transport.input(),  # Transport user input
+            stt,
+            user_aggregator,  # User responses
+            llm,  # LLM
+            tts,  # TTS
+            transport.output(),  # Transport bot output
+            assistant_aggregator,  # Assistant spoken responses
+        ]
+    )
+
+    task = PipelineTask(
+        pipeline,
+        params=PipelineParams(
+            enable_metrics=True,
+            enable_usage_metrics=True,
+        ),
+        idle_timeout_secs=runner_args.pipeline_idle_timeout_secs,
+    )
+
+    @transport.event_handler("on_client_connected")
+    async def on_client_connected(transport, client):
+        logger.info("Client connected")
+        # Kick off the conversation.
+        messages.append({"role": "system", "content": "Please introduce yourself to the user."})
+        await task.queue_frames([LLMRunFrame()])
+
+    @transport.event_handler("on_client_disconnected")
+    async def on_client_disconnected(transport, client):
+        logger.info("Client disconnected")
+        await task.cancel()
+
+    runner = PipelineRunner(handle_sigint=runner_args.handle_sigint)
+
+    await runner.run(task)
+
+
+async def bot(runner_args: RunnerArguments):
+    """Main bot entry point compatible with Pipecat Cloud."""
+    transport = await create_transport(runner_args, transport_params)
+    await run_bot(transport, runner_args)
+
+
+if __name__ == "__main__":
+    from pipecat.runner.run import main
+
+    main()
--- a/examples/foundational/54a-context-summarization-google.py
+++ b/examples/foundational/54a-context-summarization-google.py
@@ -1,0 +1,188 @@
+#
+# Copyright (c) 2024-2026, Daily
+#
+# SPDX-License-Identifier: BSD 2-Clause License
+#
+
+"""Example demonstrating context summarization feature.
+
+This example shows how to enable and configure context summarization to automatically
+compress conversation history when token limits are approached. It also demonstrates
+that summarization correctly handles function calls, preserving incomplete function
+call sequences.
+"""
+
+import asyncio
+import os
+
+from dotenv import load_dotenv
+from loguru import logger
+
+from pipecat.adapters.schemas.function_schema import FunctionSchema
+from pipecat.adapters.schemas.tools_schema import ToolsSchema
+from pipecat.audio.turn.smart_turn.local_smart_turn_v3 import LocalSmartTurnAnalyzerV3
+from pipecat.audio.vad.silero import SileroVADAnalyzer
+from pipecat.audio.vad.vad_analyzer import VADParams
+from pipecat.frames.frames import LLMRunFrame
+from pipecat.pipeline.pipeline import Pipeline
+from pipecat.pipeline.runner import PipelineRunner
+from pipecat.pipeline.task import PipelineParams, PipelineTask
+from pipecat.processors.aggregators.llm_context import LLMContext
+from pipecat.processors.aggregators.llm_response_universal import (
+    LLMAssistantAggregatorParams,
+    LLMContextAggregatorPair,
+    LLMUserAggregatorParams,
+)
+from pipecat.runner.types import RunnerArguments
+from pipecat.runner.utils import create_transport
+from pipecat.services.cartesia.tts import CartesiaTTSService
+from pipecat.services.deepgram.stt import DeepgramSTTService
+from pipecat.services.google import GoogleLLMService
+from pipecat.services.llm_service import FunctionCallParams
+from pipecat.transports.base_transport import BaseTransport, TransportParams
+from pipecat.transports.daily.transport import DailyParams
+from pipecat.transports.websocket.fastapi import FastAPIWebsocketParams
+from pipecat.turns.user_stop import TurnAnalyzerUserTurnStopStrategy
+from pipecat.turns.user_turn_strategies import UserTurnStrategies
+from pipecat.utils.context.llm_context_summarization import LLMContextSummarizationConfig
+
+load_dotenv(override=True)
+
+# We use lambdas to defer transport parameter creation until the transport
+# type is selected at runtime.
+transport_params = {
+    "daily": lambda: DailyParams(
+        audio_in_enabled=True,
+        audio_out_enabled=True,
+    ),
+    "twilio": lambda: FastAPIWebsocketParams(
+        audio_in_enabled=True,
+        audio_out_enabled=True,
+    ),
+    "webrtc": lambda: TransportParams(
+        audio_in_enabled=True,
+        audio_out_enabled=True,
+    ),
+}
+
+
+# Tool functions for the LLM
+async def get_current_weather(params: FunctionCallParams):
+    """Get the current time in a readable format."""
+    logger.info("Tool called: get_current_weather")
+    await asyncio.sleep(1)  # Simulate some processing
+    await params.result_callback({"conditions": "nice", "temperature": "75"})
+
+
+async def run_bot(transport: BaseTransport, runner_args: RunnerArguments):
+    logger.info("Starting bot")
+
+    stt = DeepgramSTTService(api_key=os.getenv("DEEPGRAM_API_KEY"))
+
+    tts = CartesiaTTSService(
+        api_key=os.getenv("CARTESIA_API_KEY"),
+        voice_id="71a7ad14-091c-4e8e-a314-022ece01c121",  # British Reading Lady
+    )
+
+    llm = GoogleLLMService(api_key=os.getenv("GOOGLE_API_KEY"))
+
+    # Register tool functions
+    llm.register_function("get_current_weather", get_current_weather)
+
+    weather_function = FunctionSchema(
+        name="get_current_weather",
+        description="Get the current weather",
+        properties={
+            "location": {
+                "type": "string",
+                "description": "The city and state, e.g. San Francisco, CA",
+            },
+            "format": {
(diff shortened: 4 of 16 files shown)


src/pipecat/processors/aggregators/llm_response_universal.py
```diff
@@ -723,6 +769,209 @@ async def _maybe_emit_user_turn_stopped(
...
+
+        # Find first system message (if any)
+        first_system_message = next((msg for msg in messages if msg.get("role") == "system"), None)
+
+        # Get messages after the last summarized index
+        # These are the recent messages that were not included in the summary
+        recent_messages = messages[last_summarized_index + 1 :]
+
+        # Create summary message using provider-formatted summary
+        summary_message = {"role": summary.role, "content": summary.content}
```
filipi87: `LLMSummaryMessage` is not needed, and the role should always be "System".

src/pipecat/frames/frames.py
```diff
@@ -1871,6 +1872,56 @@ def __post_init__(self):
...
+            async responses.
+        summary: The formatted summary message ready to be inserted into context.
+            Contains provider-appropriate role and formatted content.
+        last_summarized_index: Index (0-based) of the last message that was
+            included in the summary. Messages after this index are preserved.
+        error: Error message if summarization failed, None on success.
+    """
+
+    request_id: str
+    summary: "LLMSummarizedMessage"
```
filipi87: This could be only a String.

src/pipecat/processors/aggregators/llm_response_universal.py

src/pipecat/frames/frames.py

.claude/skills/cleanup/SKILL.md
```diff
@@ -0,0 +1,307 @@
...
+    "webrtc": lambda: TransportParams(...),
+}
+
+async def run_bot(transport: BaseTransport, runner_args: RunnerArguments):
+    stt = DeepgramSTTService(...)
+    tts = SomeTTSService(...)
+    llm = OpenAILLMService(...)
+
+    context = LLMContext(messages)
+    user_agg, assistant_agg = LLMContextAggregatorPair(...)
```
markbackman:
```suggestion
    user_aggregator, assistant_aggregator = LLMContextAggregatorPair(...)
```

.claude/skills/cleanup/SKILL.md
```diff
@@ -0,0 +1,307 @@
+# Code Cleanup Skill
```
markbackman, line 1: I think this is fine to start. We will probably want to iterate until we land on something optimized.

examples/foundational/54-context-summarization-openai.py
```diff
@@ -0,0 +1,185 @@
...
+        user_params=LLMUserAggregatorParams(
+            user_turn_strategies=UserTurnStrategies(
+                stop=[TurnAnalyzerUserTurnStopStrategy(turn_analyzer=LocalSmartTurnAnalyzerV3())]
+            ),
+            vad_analyzer=SileroVADAnalyzer(params=VADParams(stop_secs=0.2)),
+            enable_context_summarization=True,
+            # Optional: customize context summarization behavior
+            # Using low limits to demonstrate the feature quickly
+            context_summarization_config=LLMContextSummarizationConfig(
+                max_context_tokens=1000,  # Keep context within 500 tokens
```
markbackman:
```suggestion
                max_context_tokens=1000,  # Keep context within 1000 tokens
```

examples/foundational/54-context-summarization-openai.py
```diff
@@ -0,0 +1,185 @@
...
+                summarization_threshold=0.8,  # Trigger at 80% (400 tokens)
```
markbackman:
```suggestion
                summarization_threshold=0.8,  # Trigger at 80% (800 tokens)
```

examples/foundational/54a-context-summarization-google.py
```diff
@@ -0,0 +1,185 @@
...
+        user_params=LLMUserAggregatorParams(
+            user_turn_strategies=UserTurnStrategies(
+                stop=[TurnAnalyzerUserTurnStopStrategy(turn_analyzer=LocalSmartTurnAnalyzerV3())]
+            ),
+            vad_analyzer=SileroVADAnalyzer(params=VADParams(stop_secs=0.2)),
+            enable_context_summarization=True,
+            # Optional: customize context summarization behavior
+            # Using low limits to demonstrate the feature quickly
+            context_summarization_config=LLMContextSummarizationConfig(
+                max_context_tokens=1000,  # Keep context within 500 tokens
```
markbackman: Same comment fixes from the other example.

src/pipecat/frames/frames.py
```diff
@@ -1871,6 +1871,55 @@ def __post_init__(self):
...
+    """Frame requesting context summarization from an LLM service.
+
+    Sent by aggregators to LLM services when conversation context needs to be
+    compressed. The LLM service generates a summary of older messages while
+    preserving recent conversation history.
+
+    Parameters:
+        request_id: Unique identifier to match this request with its response.
+            Used to handle async responses and avoid race conditions.
+        context: The full LLM context containing all messages to analyze.
```
markbackman: Nit:

```suggestion
        context: The full LLM context containing all messages to analyze and summarize.
```

src/pipecat/utils/context/llm_context_summarization.py
```diff
@@ -0,0 +1,437 @@
...
+            summarization. This ensures regular compression even if token
+            limits are not reached.
+        min_messages_after_summary: Number of recent messages to preserve
+            uncompressed after each summarization. These messages maintain
+            immediate conversational context.
+        summarization_prompt: Custom prompt for the LLM to use when generating
+            summaries. If None, uses DEFAULT_SUMMARIZATION_PROMPT.
+    """
+
+    max_context_tokens: int = 8000
```
markbackman, line 78: Is there a reason to include `max_context_tokens` and `summarization_threshold`? Would it be more straightforward to just include one param, say `max_context_tokens`, which serves as the limit. Adding the threshold makes setting the value a little less straightforward.

src/pipecat/utils/context/llm_context_summarization.py
```diff
@@ -0,0 +1,437 @@
...
+
+
+@dataclass
+class LLMContextSummarizationConfig:
+    """Configuration for context summarization behavior.
+
+    Controls when and how conversation context is automatically compressed
+    to manage token limits in long-running conversations.
+
+    Attributes:
```
markbackman: For docstrings, this should be:
```suggestion
    Parameters:
```

src/pipecat/utils/context/llm_context_summarization.py
```diff
@@ -0,0 +1,437 @@
...
+            The custom prompt if set, otherwise the default summarization prompt.
+        """
+        return self.summarization_prompt or DEFAULT_SUMMARIZATION_PROMPT
+
+
+@dataclass
+class LLMMessagesToSummarize:
+    """Result of get_messages_to_summarize operation.
+
+    Attributes:
```
markbackman:
```suggestion
    Parameters:
```

.claude/skills/cleanup/SKILL.md
```diff
@@ -0,0 +1,307 @@
...
+    return (48000, 1)
+```
+
+### After: Named Class
+
+```python
+class AudioInfo:
+    """Audio configuration information.
+
+    Attributes:
```
markbackman:
```suggestion
    Parameters:
```

src/pipecat/processors/aggregators/llm_response_universal.py
```diff
@@ -723,6 +768,211 @@ async def _maybe_emit_user_turn_stopped(
...
+        if self._summarization_in_progress:
+            logger.debug(f"{self}: Summarization already in progress")
+            return False
+
+        # Estimate tokens in context
+        total_tokens = LLMContextSummarizationUtil.estimate_context_tokens(self._context)
+        num_messages = len(self._context.messages)
+
+        # Check if we've reached the token threshold
+        token_threshold = int(
```
markbackman: I made a comment below, in the `LLMContextSummarizationConfig`, but we might want to set the threshold just from a `max_context_tokens` param to make this more straightforward.

src/pipecat/processors/aggregators/llm_response_universal.py
```diff
@@ -723,6 +768,211 @@ async def _maybe_emit_user_turn_stopped(
...
+        )
+        token_threshold_exceeded = total_tokens >= token_threshold
+
+        # Check if we've exceeded max unsummarized messages
+        messages_since_summary = len(self._context.messages) - 1
+        message_threshold_exceeded = (
+            messages_since_summary >= self._summarization_config.max_unsummarized_messages
+        )
+
+        logger.debug(
```
markbackman: Maybe set to `trace` since this will be a lot of info to log for every completion.
```suggestion
        logger.trace(
```

src/pipecat/processors/aggregators/llm_response_universal.py
```diff
@@ -723,6 +768,211 @@ async def _maybe_emit_user_turn_stopped(
...
+            - Summarization is enabled
+            - No summarization currently in progress
+            - AND either:
+              - Token count exceeds threshold (max_context_tokens * summarization_threshold)
+              - OR message count exceeds max_unsummarized_messages since last summary
+        """
+        logger.debug(f"{self}: Checking if context summarization is needed")
+
+        if not self._summarization_enabled:
+            logger.debug(f"{self}: Summarization disabled")
```
markbackman: Trace or remove

```suggestion
            logger.trace(f"{self}: Summarization disabled")
```

src/pipecat/processors/aggregators/llm_response_universal.py
```diff
@@ -723,6 +768,211 @@ async def _maybe_emit_user_turn_stopped(
...
+            return False
+
+        if self._summarization_in_progress:
+            logger.debug(f"{self}: Summarization already in progress")
```
markbackman: Trace or remove

```suggestion
            logger.trace(f"{self}: Summarization already in progress")
```

src/pipecat/processors/aggregators/llm_response_universal.py
```diff
@@ -723,6 +768,211 @@ async def _maybe_emit_user_turn_stopped(
...
+            f"~{total_tokens} tokens (token threshold: {token_threshold}, "
+            f"{self._summarization_config.summarization_threshold * 100}% of "
+            f"{self._summarization_config.max_context_tokens}), "
+            f"{messages_since_summary} messages since last summary "
+            f"(message threshold: {self._summarization_config.max_unsummarized_messages})"
+        )
+
+        # Trigger if either threshold is exceeded
+        if not token_threshold_exceeded and not message_threshold_exceeded:
+            logger.debug(
```
markbackman: Trace or remove

```suggestion
            logger.trace(
```

src/pipecat/processors/aggregators/llm_response_universal.py
```diff
@@ -723,6 +768,211 @@ async def _maybe_emit_user_turn_stopped(
...
+
+        reason = []
+        if token_threshold_exceeded:
+            reason.append(f"~{total_tokens} tokens (>{token_threshold} threshold)")
+        if message_threshold_exceeded:
+            reason.append(
+                f"{messages_since_summary} messages (>{self._summarization_config.max_unsummarized_messages} threshold)"
+            )
+
+        logger.info(f"{self}: ✓ Summarization needed - {', '.join(reason)}")
```
markbackman: Debug

```suggestion
        logger.debug(f"{self}: ✓ Summarization needed - {', '.join(reason)}")
```

src/pipecat/processors/aggregators/llm_response_universal.py
```diff
@@ -723,6 +768,211 @@ async def _maybe_emit_user_turn_stopped(
...
+        first_system_message = next((msg for msg in messages if msg.get("role") == "system"), None)
+
+        # Get messages after the last summarized index
+        # These are the recent messages that were not included in the summary
+        recent_messages = messages[last_summarized_index + 1 :]
+
+        # Create summary message using provider-formatted summary
+        summary_message = {
+            "role": "system",
+            "content": f"Here's a summary of the conversation so far:\n{summary}",
```
markbackman: Could this be more concise, like `Conversation summary:\n`? (Saves 5 tokens, which is negligible, but I think would create an equivalent outcome.)

```suggestion
            "content": f"Conversation summary:\n{summary}",
```

src/pipecat/services/llm_service.py
```diff
@@ -372,6 +383,127 @@ async def _handle_interruptions(self, _: InterruptionFrame):
...
+            logger.debug(f"{self}: No messages to summarize")
+            raise RuntimeError("No messages to summarize")
+
+        logger.debug(
+            f"{self}: Generating summary for {len(result.messages)} messages "
+            f"(index 0 to {result.last_summarized_index})"
+        )
+
+        # Calculate max_tokens for the summary using utility method
+        max_summary_tokens = LLMContextSummarizationUtil.calculate_max_summary_tokens(
```
markbackman: Why is this needed? Doesn't the user specify the max tokens param for the completion in their  `LLMContextSummarizationConfig`?

src/pipecat/services/llm_service.py
```diff
@@ -372,6 +383,127 @@ async def _handle_interruptions(self, _: InterruptionFrame):
...
+        # Create summary context
+        transcript = LLMContextSummarizationUtil.format_messages_for_summary(result.messages)
+        prompt_messages = [
+            {
+                "role": "system",
+                "content": frame.summarization_prompt,
+            },
+            {
+                "role": "user",
+                "content": f"The maximum context size is {max_summary_tokens} tokens. "
```
markbackman: I think you may be able to distill this prompt down to just `Conversation history:\n{transcript}`. The system prompt and the completion parameters should carry the weight of telling the LLM how to summarize. The job of the user message is to just provide the conversation history up to that point. This should save a few tokens.

src/pipecat/services/llm_service.py
```diff
@@ -372,6 +383,127 @@ async def _handle_interruptions(self, _: InterruptionFrame):
...
+                f"Generate a summary that keeps the total context within this limit.\n\n"
+                f"Conversation history:\n{transcript}",
+            },
+        ]
+        summary_context = LLMContext(messages=prompt_messages)
+
+        # Generate summary using run_inference
+        # This will be overridden by each LLM service implementation
+        try:
+            summary_text = await self.run_inference(summary_context, max_tokens=max_summary_tokens)
```
markbackman: Shouldn't this be set from the `LLMContextSummarizationConfig` param?

src/pipecat/services/llm_service.py
```diff
@@ -372,6 +383,127 @@ async def _handle_interruptions(self, _: InterruptionFrame):
...
+        except NotImplementedError:
+            raise RuntimeError(
```
markbackman, line 489: 👍 This will guard against realtime LLMs.

src/pipecat/utils/context/llm_context_summarization.py
```diff
@@ -0,0 +1,437 @@
...
+
+from pipecat.processors.aggregators.llm_context import LLMContext
+
+# Token estimation constants
+TOKEN_OVERHEAD_PER_MESSAGE = 10  # Estimated structural overhead per message
+IMAGE_TOKEN_ESTIMATE = 100  # Rough estimate for image content
+SUMMARY_TOKEN_BUFFER = 0.8  # Keep summary at 80% of available space for safety
+MIN_SUMMARY_TOKENS = 100  # Minimum tokens to allocate for summary
+
+DEFAULT_SUMMARIZATION_PROMPT = """You are summarizing a conversation between a user and an AI assistant.
```
markbackman, line 27: 👍 

This will probably be overridden, but it seems like a good default, general purpose summary prompt.

src/pipecat/utils/context/llm_context_summarization.py
```diff
@@ -0,0 +1,437 @@
...
+
+        Args:
+            text: Text to estimate tokens for
+
+        Returns:
+            Estimated token count (characters // 4)
+        """
+        if not text:
+            return 0
+        return len(text) // 4
```
markbackman: Maybe make `4` a constant a the top of the file too?

src/pipecat/utils/context/llm_context_summarization.py
```diff
@@ -0,0 +1,437 @@
...
+from dataclasses import dataclass
+from typing import List, Optional
+
+from loguru import logger
+
+from pipecat.processors.aggregators.llm_context import LLMContext
+
+# Token estimation constants
+TOKEN_OVERHEAD_PER_MESSAGE = 10  # Estimated structural overhead per message
+IMAGE_TOKEN_ESTIMATE = 100  # Rough estimate for image content
```
markbackman: I have no idea how to think about images, but I think they're really big. Probably bigger than this. But, we don't want to give them too big of a size because a single image would cause compression, which for a stream of images may cause things to really break down.

markbackman: Claude shared this:
```
- OpenAI GPT-4o: Low-res is 85 tokens, but high-res ranges from ~170 to 1,105+ tokens depending on tile count (each 512x512 tile = 170 tokens + 85 base). Most real images will be high-res.
- Anthropic Claude: Based on image dimensions — a small 200x200 image is ~250 tokens, while 1024x1024 can be ~1,600 tokens.
- Google Gemini: Images typically cost ~258 tokens.
```

Maybe something larger, ~400-500 tokens?

src/pipecat/utils/context/llm_context_summarization.py
```diff
@@ -0,0 +1,437 @@
...
+        summary_end = len(messages) - min_messages_to_keep
+
+        if summary_start >= summary_end:
+            return LLMMessagesToSummarize(messages=[], last_summarized_index=-1)
+
+        # Check for function calls in progress in the range we want to summarize
+        function_call_start = LLMContextSummarizationUtil._get_function_calls_in_progress_index(
+            messages, summary_start
+        )
+        if function_call_start >= 0 and function_call_start < summary_end:
```
markbackman, line 321: This makes sense. Summarize up to the point where the function call is in progress. That helps with some compression, though not optimal. This is good though; much safer.

markbackman (commented):
This looks really great! Very clean and easy to understand. Nice work 👏 

I'm going to do some testing and will let you know if I find anything.

src/pipecat/utils/context/llm_context_summarization.py
```diff
@@ -0,0 +1,437 @@
...
+
+        messages_to_summarize = messages[summary_start:summary_end]
+        last_summarized_index = summary_end - 1
+
+        return LLMMessagesToSummarize(
+            messages=messages_to_summarize, last_summarized_index=last_summarized_index
+        )
+
+    @staticmethod
+    def format_messages_for_summary(messages: List[dict]) -> str:
```
markbackman, line 347: Just curious if this format is needed or if it's just for token efficiency.

src/pipecat/utils/context/llm_context_summarization.py
filipi87: Yeah, my original plan was to trigger it when we were getting close to the threshold, but I think you’re right. It does feel redundant. 
I will remove the `summarization_threshold`. I think that makes sense. Nice catch.

examples/foundational/54-context-summarization-openai.py
filipi87: Done.

examples/foundational/54-context-summarization-openai.py
filipi87: Done.

examples/foundational/54a-context-summarization-google.py
filipi87: Done.

src/pipecat/frames/frames.py
filipi87: Fixed.

src/pipecat/utils/context/llm_context_summarization.py
filipi87: Done. I have removed the `summarization_threshold`.

src/pipecat/utils/context/llm_context_summarization.py
filipi87: Fixed.

src/pipecat/utils/context/llm_context_summarization.py
filipi87: Fixed.

.claude/skills/cleanup/SKILL.md
filipi87: Fixed.

src/pipecat/processors/aggregators/llm_response_universal.py
filipi87: Yep, I have removed the `summarization_threshold`. 👍

src/pipecat/processors/aggregators/llm_response_universal.py

src/pipecat/processors/aggregators/llm_response_universal.py

src/pipecat/processors/aggregators/llm_response_universal.py
filipi87: I don’t think this will happen that much. Maybe we should keep it for a while for debugging ?

src/pipecat/processors/aggregators/llm_response_universal.py

src/pipecat/processors/aggregators/llm_response_universal.py

src/pipecat/processors/aggregators/llm_response_universal.py

src/pipecat/services/llm_service.py
filipi87: Yep, that is the plan.

src/pipecat/utils/context/llm_context_summarization.py
filipi87: Agreed. They will probably override it to be more specific. Makes sense.

src/pipecat/utils/context/llm_context_summarization.py
filipi87: Done. Created a new constant CHARS_PER_TOKEN

src/pipecat/utils/context/llm_context_summarization.py
filipi87: That was my thought. Better to be conservative for now.

src/pipecat/utils/context/llm_context_summarization.py
filipi87: Kwin was using a similar approach on Gradient Bang, and it felt more natural to format all the messages as a single user message. It also saves some tokens, so it felt like a good solution for this.

src/pipecat/utils/context/llm_context_summarization.py
filipi87: Exactly. I just considered some random size here, as I mentioned in the comment, a totally rough estimate.

Right now, we are not even using the image when compressing the context. Maybe in the future that’s something we should allow.

For now, though, I thought it was better to keep it out. If we need it later, we can add a property to define whether we should keep the images or try to compress them somehow.

src/pipecat/utils/context/llm_context_summarization.py
markbackman: Makes sense. I was curious was inspired the approach. Thanks!

src/pipecat/utils/context/llm_context_summarization.py
markbackman: That's fine. Images are so infrequent. I think this is fine.

src/pipecat/processors/aggregators/llm_response_universal.py
markbackman: That's fine. Yes, debug is OK.

markbackman (approved):
LGTM! Thanks for the clean up.

From the first review, the only item I still see open is this one (regarding the base summary prompt):
https://github.com/pipecat-ai/pipecat/pull/3621/changes#r2775817727

Aside from that, this looks good to go. It probably makes sense to get input from someone else too since this is such a key feature and will get a ton of use.

filipi87:
I have missed that one. Fixed. Thank you for the review @markbackman. 🙌

changelog/3621.added.md
```diff
@@ -0,0 +1,5 @@
+- Added context summarization feature to automatically compress conversation history when token limits are approached, enabling efficient long-running conversations.
```
kompfner:
```suggestion
- Added context summarization feature to automatically compress conversation history when conversation length limits (by token or message count) are approached, enabling efficient long-running conversations.
```

examples/foundational/54-context-summarization-openai.py
```diff
@@ -0,0 +1,184 @@
...
+        user_params=LLMUserAggregatorParams(
+            user_turn_strategies=UserTurnStrategies(
+                stop=[TurnAnalyzerUserTurnStopStrategy(turn_analyzer=LocalSmartTurnAnalyzerV3())]
+            ),
+            vad_analyzer=SileroVADAnalyzer(params=VADParams(stop_secs=0.2)),
+            enable_context_summarization=True,
+            # Optional: customize context summarization behavior
+            # Using low limits to demonstrate the feature quickly
+            context_summarization_config=LLMContextSummarizationConfig(
+                max_context_tokens=1000,  # Trigger summarization at 1000 tokens
```
kompfner, line 133: Ah looks like we've moved away from `summarization_threshold`? If so: agreed! It was one extra layer of "math" 👍 I don't see a problem with developers just directly specifying the number directly instead of a max and a threshold, which would simply get multiplied.

src/pipecat/frames/frames.py
```diff
@@ -1890,6 +1890,55 @@ def __post_init__(self):
         self.skip_tts = None
 
 
+@dataclass
+class LLMContextSummaryRequestFrame(ControlFrame):
+    """Frame requesting context summarization from an LLM service.
+
+    Sent by aggregators to LLM services when conversation context needs to be
```
kompfner, line 1998: Might also be worth describing what the LLMs then do with that summary (i.e. push a `LLMContextSummaryResultFrame`, right?)

src/pipecat/frames/frames.py
```diff
@@ -1890,6 +1890,55 @@ def __post_init__(self):
...
+    compressed. The LLM service generates a summary of older messages while
+    preserving recent conversation history.
+
+    Parameters:
+        request_id: Unique identifier to match this request with its response.
+            Used to handle async responses and avoid race conditions.
+        context: The full LLM context containing all messages to analyze and summarize.
+        min_messages_to_keep: Number of recent messages to preserve uncompressed.
+            These messages will not be included in the summary.
+        max_context_tokens: Maximum allowed context size in tokens. The LLM should
```
kompfner: Feedback applies throughout: should we transparent about the fact that it's _approximate_ tokens, by maybe calling it something like `max_approx_context_tokens` (and updating docstring comments accordingly to let developers know what they're dealing with)

src/pipecat/frames/frames.py
```diff
@@ -1890,6 +1890,55 @@ def __post_init__(self):
         self.skip_tts = None
 
 
+@dataclass
+class LLMContextSummaryRequestFrame(ControlFrame):
```
kompfner, line 1995: Q: do we want to consume these frames in the LLMs, or let them continue down the pipeline, just in case anyone wants to handle them in custom processors?

My inclination is that these seem fine to consume in the LLMs...

src/pipecat/frames/frames.py
```diff
@@ -1890,6 +1890,55 @@ def __post_init__(self):
...
+    min_messages_to_keep: int
+    max_context_tokens: int
+    summarization_prompt: str
+
+
+@dataclass
+class LLMContextSummaryResultFrame(ControlFrame):
+    """Frame containing the result of context summarization.
+
+    Sent by LLM services back to aggregators after generating a summary.
```
kompfner, line 2026: This one shouldn't be swallowed by the aggregator right?

src/pipecat/processors/aggregators/llm_response_universal.py
```diff
@@ -98,6 +105,12 @@ class LLMUserAggregatorParams:
         user_turn_completion_config: Configuration for turn completion behavior including
             custom instructions, timeouts, and prompts. Only used when
             filter_incomplete_user_turns is True.
+        enable_context_summarization: Enable automatic context summarization when token
+            limits are approached (disabled by default). When enabled, older conversation
+            messages are automatically compressed into summaries to manage context size.
```
kompfner: `token limits` -> `conversation length limits (by token or message count)`

kompfner: Wait, is it when the limits are _approached_, or when the limits are _reached_? With the removal of the threshold parameter, it's now when they're reached, right?

This feedback also applies to the changelog entry.

src/pipecat/processors/aggregators/llm_response_universal.py
```diff
@@ -417,6 +432,14 @@ def __init__(
         # pushed by strategies.
         self._self_queued_frames = set()
 
+        # Context summarization state
+        self._summarization_enabled = self._params.enable_context_summarization
+        self._summarization_config = (
+            self._params.context_summarization_config or LLMContextSummarizationConfig()
+        )
+        self._summarization_in_progress = False
+        self._pending_summary_request_id: Optional[str] = None
```
kompfner: do we need both of the above state variables? could the presence of `_pending_summary_request_id` indicate that a summarization is in progress?

src/pipecat/processors/aggregators/llm_response_universal.py
```diff
@@ -534,6 +563,12 @@ async def _cleanup(self):
         for s in self._params.user_mute_strategies:
             await s.cleanup()
 
+    async def _clear_summarization(self):
```
kompfner: nit: `_clear_summarization_state` might be clearer

src/pipecat/processors/aggregators/llm_response_universal.py
```diff
@@ -729,6 +774,206 @@ async def _maybe_emit_user_turn_stopped(
...
+
+        # Validate context state
+        if not self._validate_summary_context(frame.last_summarized_index):
+            logger.warning(f"{self}: Context state changed, skipping summary application")
+            return
+
+        # Apply summary
+        await self._apply_summary(frame.summary, frame.last_summarized_index)
+
+    def _validate_summary_context(self, last_summarized_index: int) -> bool:
```
kompfner: Is this function safeguarding against the possibility of programmatic edits to the context, like from `LLMMessageUpdateFrame`s and the like? If so, then in a [PR](https://github.com/pipecat-ai/pipecat/pull/3620) I worked on recently (which maybe you've looked through already) I added some mechanisms for tracking with more certainty whether a context has been edited...wonder if we could join forces and use those here to determine with more certainty whether a summary still applies.

src/pipecat/processors/aggregators/llm_response_universal.py
```diff
@@ -729,6 +774,206 @@ async def _maybe_emit_user_turn_stopped(
...
+        as determined by the LLM service's adapter.
+
+        Args:
+            summary: The formatted summary message from the LLM service adapter.
+            last_summarized_index: The index of the last message included in
+                the summary. Messages after this index are preserved.
+        """
+        messages = self._context.messages
+
+        # Find first system message (if any)
```
kompfner: Probably a very fringe case, but should we handle the possibility of the first system message appearing later than the last summarized index? There's technically no hard requirement that a "system"-role message has to appear at or near the beginning of the conversation, esp. with providers like OpenAI.

Or...does the summarization process already exclude the first system message? (I should probably just read on to find out, but wanted to jot this note down here).

kompfner: When using OpenAI (where "system"-role messages can appear anywhere), forcing the system message to appear at the beginning on summarization might mess with conversation flow. But on the other hand, summarization does needs to work universally, and other providers don't handle "system"-role messages anywhere...

With Gemini, we "pull" the first system instruction out of the messages and use it as the overall system instruction (which it seems like the logic here is modeled after). But with AWS Bedrock, we only pull a "system"-role message our of messages and use it as system instruction if it's the _first_ message. We're inconsistent, which isn't ideal...

As I think about it, two approaches come to mind:
* What you have here
* Only checking the very first message in `messages` for a system message

_Almost_ always, those two are the same. So in practice I don't know if this makes much of a difference.

But it's a good reminder that we should probably do a consistency pass on how we translate "system"-role messages for different providers.

src/pipecat/services/llm_service.py
```diff
@@ -372,6 +383,125 @@ async def _handle_interruptions(self, _: InterruptionFrame):
...
+            logger.debug(f"{self}: No messages to summarize")
+            raise RuntimeError("No messages to summarize")
+
+        logger.debug(
+            f"{self}: Generating summary for {len(result.messages)} messages "
+            f"(index 0 to {result.last_summarized_index})"
+        )
+
+        # Calculate max_tokens for the summary using utility method
+        max_summary_tokens = LLMContextSummarizationUtil.calculate_max_summary_tokens(
```
kompfner: For my understanding: the `max_context_tokens` that the developer specifies as the point where a summary should be triggered is the same number used to compute how big the summary should be?

It seems like how big you want the summary to be should be (at least somewhat) independent—you might want to ask the summary to be relatively compact so you don't have to do it as often, rather than letting it take up all the remaining space, no?

src/pipecat/processors/aggregators/llm_response_universal.py
```diff
@@ -729,6 +774,206 @@ async def _maybe_emit_user_turn_stopped(
...
+        # Estimate tokens in context
+        total_tokens = LLMContextSummarizationUtil.estimate_context_tokens(self._context)
+        num_messages = len(self._context.messages)
+
+        # Check if we've reached the token limit
+        token_limit = self._summarization_config.max_context_tokens
+        token_limit_exceeded = total_tokens >= token_limit
+
+        # Check if we've exceeded max unsummarized messages
+        messages_since_summary = len(self._context.messages) - 1
```
kompfner: nit: do you want to count the possible initial system message towards the limit? if not, you might have to subtract 2, no?

examples/foundational/54-context-summarization-openai.py
filipi87: Yeah, initially I thought it would make sense to have both, but since both you and Mark mentioned this, it really just looked unnecessary. So I’ve removed the `summarization_threshold`.

src/pipecat/frames/frames.py
filipi87: I was describing what they do inside `LLMContextSummaryResultFrame`.

src/pipecat/frames/frames.py
filipi87: I think the name `max_context_tokens` is correct here, because this is what we pass to the LLM when running inference to enforce the maximum number of tokens.

The one inside `LLMContextSummarizationConfig`, in that case, what the user specifies is only an approximation, since the way we calculate the tokens is approximate. Even so, in my opinion we should keep the same name there, but make it clear in the docstring that the token calculation is an approximation.

What do you think ?

src/pipecat/frames/frames.py
filipi87: As we discussed in yesterday’s meeting, in this case, since we’re actually handling the frame and doing something with it, creating the summary, it feels more natural to consume the frame here rather than let it continue.

src/pipecat/frames/frames.py
filipi87: Yep, as per our discussion, we should let this one through. 🙂

src/pipecat/processors/aggregators/llm_response_universal.py
filipi87: Exactly, since we removed the threshold, now it is when they are _reached_. Nice catch.

src/pipecat/processors/aggregators/llm_response_universal.py
filipi87: Yeah, that could work. 👍

src/pipecat/processors/aggregators/llm_response_universal.py
filipi87: What I am doing in this PR is finding the index of the first system message and defining `summary_start = first_system_index + 1`.

I then summarize only the messages that come after the first system message, or everything if there is no system message.

src/pipecat/services/llm_service.py
filipi87: It basically is, but when calculating the available space to define the `max_tokens` that I pass to the LLM, I always apply a 0.8 buffer to keep the summary at a maximum of 80% of the available space.

But I think you’re right, we should probably create a:
- `target_context_tokens`: the target maximum context size in tokens after summarization.

What do you think?

src/pipecat/frames/frames.py
filipi87: Ok, for now I have improved the description of `max_context_tokens`, inside `LLMContextSummarizationConfig`, explaining how the tokens are calculated.

src/pipecat/processors/aggregators/llm_response_universal.py
```diff
@@ -722,6 +767,206 @@ async def _maybe_emit_user_turn_stopped(
...
 
+    async def _handle_interruptions(self, frame: InterruptionFrame):
+        # If an interruption happened before we received the summary response
+        # from the LLM, we need to cancel the pending summarization.
+        # Because maybe we will never receive the response frame again
+        await self._clear_summarization()
+
+    # Context summarization methods
+
+    def _should_summarize(self) -> bool:
```
aconchillo: We should have probably moved all this logic into it's own class instead of making LLMUserAggregator bigger. Same we have done with all the controllers. Something that can be also tested separately from the user aggregator.

src/pipecat/processors/aggregators/llm_response_universal.py
```diff
@@ -110,6 +123,8 @@ class LLMUserAggregatorParams:
     vad_analyzer: Optional[VADAnalyzer] = None
     filter_incomplete_user_turns: bool = False
     user_turn_completion_config: Optional[UserTurnCompletionConfig] = None
+    enable_context_summarization: bool = False
+    context_summarization_config: Optional[LLMContextSummarizationConfig] = None
```
aconchillo: What is the reason for adding this into the user aggregator instead of the assistant one? Just curious.

src/pipecat/frames/frames.py
filipi87: Fixed.

src/pipecat/processors/aggregators/llm_response_universal.py
filipi87: Fixed.

src/pipecat/processors/aggregators/llm_response_universal.py
aconchillo: That is the LLMUserAggreagtor should just have a class `LLMContextSummarizer`

src/pipecat/services/llm_service.py
```diff
@@ -372,6 +383,125 @@ async def _handle_interruptions(self, _: InterruptionFrame):
...
+        according to the provider's requirements. Broadcasts the result back
+        to the aggregator for context reconstruction.
+
+        Args:
+            frame: The summary request frame containing context and parameters.
+        """
+        logger.debug(f"{self}: Processing summarization request {frame.request_id}")
+
+        # Create a background task to generate the summary without blocking
+        self.create_task(self._generate_summary_task(frame))
```
aconchillo: we should save a reference to the task and cancel it on cleanup if necessary. we should also call

```
await asyncio.sleep(0)
```

to schedule the task in the event loop.

src/pipecat/processors/aggregators/llm_response_universal.py
aconchillo: It actually, seems to be this should be done in the assistant, feels more natural, I think.

filipi87: I decided to include it there because each time the user started a new turn, and we pushed a new context frame, I thought it was a nice moment to also push, as a follow up, a frame requesting summarization if needed.

This way, the LLM would have time to process it while the TTS was generating the previous answer.

So it felt like a good spot to add this logic without impacting performance.

But like we discussed on slack, we can achieve something similar using the assistant aggregator, if we use the `LLMFullResponseStartFrame` to trigger if we should or not request the context summarization.

src/pipecat/processors/aggregators/llm_response_universal.py

src/pipecat/processors/aggregators/llm_response_universal.py
filipi87: Yep, done.

src/pipecat/processors/aggregators/llm_response_universal.py

src/pipecat/services/llm_service.py
aconchillo: we still need to keep track of this task. and, since we don't await it, we need to call `async asyncio.sleep(0)`.

filipi87: Yep, I have just pushed the fix to keep track of it.

filipi87: But I am not sure why we need this ?
`asyncio.sleep(0)`

filipi87: I havent added this one yet. Is it really needed ?

aconchillo (approved):

src/pipecat/services/llm_service.py
aconchillo: If we don't cancel the task during interruption, then no, not needed. The reason is that if you cancel a task before the task is started by the event loop (different than created) you will get RuntimeWarnings saying that the task was never awaited. Since we don't cancel the task during interruptions, I think we should be ok.

aconchillo:
LGTM! :clap:

src/pipecat/services/llm_service.py
filipi87: Done. I have created `target_context_tokens`

PR #3625, opened 2026-02-03
[inworld] Allow Async delivery of timestamps info

ianbbqzy:
* speed up first audio chunk latency

There might be requirements to keep the timestamp info synced with the audio chunks. E.g. if a response chunk `{audio_bytes: asdflksdafja}` has the audio for "Hello", then the same response must also have `{audio_bytes: asdflksdafja, timestamp_info: {... "Hello" ...}}`, such as in some latency sensitive lipsync use cases.

But IIUC, Pipecat doesn't really care about this because the timestamps and audios are emitted to two independent channels anyway. If my understanding is incorrect, I will refactor the PR to allow SDK users to configure `SYNC` vs `ASYNC` in case some have this requirement for backward compatibility

Official docs for this mode will be up shortly.

### DIFF
--- a/src/pipecat/services/inworld/tts.py
+++ b/src/pipecat/services/inworld/tts.py
@@ -17,7 +17,7 @@ import asyncio
 import base64
 import json
 import uuid
-from typing import Any, AsyncGenerator, Dict, List, Optional, Tuple
+from typing import Any, AsyncGenerator, Dict, List, Literal, Optional, Tuple
 
 import aiohttp
 import websockets
@@ -65,10 +65,12 @@ class InworldHttpTTSService(WordTTSService):
         Parameters:
             temperature: Temperature for speech synthesis.
             speaking_rate: Speaking rate for speech synthesis.
+            timestamp_transport_strategy: The strategy to use for timestamp transport.
         """
 
         temperature: Optional[float] = None
         speaking_rate: Optional[float] = None
+        timestamp_transport_strategy: Optional[Literal["ASYNC", "SYNC"]] = None
 
     def __init__(
         self,
@@ -128,6 +130,8 @@ class InworldHttpTTSService(WordTTSService):
             self._settings["temperature"] = params.temperature
         if params.speaking_rate is not None:
             self._settings["audioConfig"]["speakingRate"] = params.speaking_rate
+        if params.timestamp_transport_strategy is not None:
+            self._settings["timestampTransportStrategy"] = params.timestamp_transport_strategy
 
         self._cumulative_time = 0.0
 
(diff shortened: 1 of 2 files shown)


ianbbqzy:
cc @cshape

src/pipecat/services/inworld/tts.py
```diff
@@ -123,6 +125,8 @@ def __init__(
             self._settings["temperature"] = params.temperature
         if params.speaking_rate is not None:
             self._settings["audioConfig"]["speakingRate"] = params.speaking_rate
+        if params.timestamp_transport_strategy is not None:
```
markbackman, line 133: You're setting `timestamp_transport_strategy` to `ASYNC`, so this will always have a value. If that's expected, then you can simplify to:
```
self._settings["timestampTransportStrategy"] = params.timestamp_transport_strategy
```

If you want it to be optional, you should update the type and set to None.

This is applies to both classes.

src/pipecat/services/inworld/tts.py
markbackman: Unrelated to this change, I found an issue that we should fix. The TTFB value is the time to first byte for a given context. Currently, it's being emitted for each individual input.

I would recommend moving `await self.start_ttfb_metrics()` inside the `self._context_id` check:
```
            try:
                yield TTSStartedFrame(context_id=context_id)

                if not self._context_id:
```

markbackman (commented):
Generally looks good! Just a few findings.

Also, you'll need to rebase to get the latest changes (should only affect imports).

src/pipecat/services/inworld/tts.py
```diff
@@ -419,14 +424,16 @@ class InputParams(BaseModel):
                 flushing of buffered text to achieve minimal latency while
                 maintaining high quality audio output. If None (default),
```
markbackman, line 434: `None` is no longer the default.

```suggestion
                maintaining high quality audio output. If None,
```

src/pipecat/services/inworld/tts.py
ianbbqzy: You are right, currently the check is unnecessary. But I made it actually optional instead as in developers can modify it to be None. But we will still default to ASYNC as we want users to have a better first impression

src/pipecat/services/inworld/tts.py
ianbbqzy: Looks like it's been updated https://github.com/pipecat-ai/pipecat/commit/136732afae720857f80a849f8b6221836d2481ac

but looks like it would conflict with my other PR

markbackman (approved):
LGTM!

PR #3626, opened 2026-02-03
added tool handling & triggered reconnection

### LINKED ISSUE #3549: GeminiLiveVertexLLMService Vertex AI issue with update_settings
### pipecat version

pipecat-ai==0.0.100 pipecat-ai-flows==0.0.22

### Python version

3.13

### Operating System

Darwin MacBook-Pro-2.local 24.1.0 Darwin Kernel Version 24.1.0: Thu Oct 10 21:03:15 PDT 2024; root:xnu-11215.41.3~2/RELEASE_ARM64_T6000 arm64

### Issue description

There is some issue with the update of a frame received, which seems to affect the tool calling:
https://github.com/pipecat-ai/pipecat/blob/main/src/pipecat/services/google/gemini_live/llm.py#L950

### Reproduction steps

1. Setup a project with Google Vertex and do a call
2. Interact with the LLM so it attempts to do a node transition and calls a tool

### Expected behavior

1. Tool call should work

### Actual behavior

Agent wont call a tool

### Logs

```shell
2026-01-24 03:22:07.893 | DEBUG    | pipecat_flows.manager:_execute_transition:602 - Transition to function-returned node: clarify_intent
2026-01-24 03:22:07.893 | DEBUG    | pipecat_flows.manager:_set_node:769 - Setting node: clarify_intent
2026-01-24 03:22:07.895 | DEBUG    | pipecat_flows.manager:_update_llm_context:973 - Updated LLM context using LLMMessagesAppendFrame with strategy ContextStrategy.APPEND
2026-01-24 03:22:07.895 | DEBUG    | pipecat_flows.manager:_set_node:861 - Updated LLM context
2026-01-24 03:22:07.895 | DEBUG    | pipecat_flows.manager:_set_node:880 - Successfully set node: clarify_intent
2026-01-24 03:22:07.896 | ERROR    | pipecat.processors.frame_processor:push_error_frame:725 - GeminiLiveVertexLLMService#0 exception (/Users/andre/Devel/amp_assist_agent_test_py313/.venv/lib/python3.13/site-packages/pipecat/services/google/gemini_live/llm.py:953): Error processing frame: AIService._update_settings() missing 1 required positional argument: 'settings'
2026-01-24 03:22:07.897 | DEBUG    | pipecat.adapters.base_llm_adapter:from_standard_tools:127 - Retrieving the tools using the adapter: <class 'pipecat.adapters.services.gemini_adapter.GeminiLLMAdapter'>
2026-01-24 03:22:07.898 | WARNING  | pipecat.pipeline.task:_source_push_frame:781 - PipelineTask#0: Something went wrong: ErrorFrame#1(error: Error processing frame: AIService._update_settings() missing 1 required positional argument: 'settings', fatal: False)
```

omChauhanDev:

Fixes #3549                                                                                                                           
                                                                                                                                        
When `LLMSetToolsFrame` was processed, the code called `_update_settings()` without the required argument, causing a TypeError. Since Gemini Live does not currently support mid-session tool updates, replaced the broken call with a warning log.

Before : 
<img width="2902" height="1718" alt="image" src="https://github.com/user-attachments/assets/f42b0021-1608-431e-ba67-2192477d8c19" />


After : 
<img width="2362" height="180" alt="image" src="https://github.com/user-attachments/assets/e1555101-3d14-4eba-b21a-4124e71882cb" />

### DIFF
--- a/src/pipecat/services/google/gemini_live/llm.py
+++ b/src/pipecat/services/google/gemini_live/llm.py
@@ -950,10 +950,7 @@ class GeminiLiveLLMService(LLMService):
         elif isinstance(frame, LLMUpdateSettingsFrame):
             await self._update_settings(frame.settings)
         elif isinstance(frame, LLMSetToolsFrame):
-            if frame.tools is not None:
-                self._tools_from_init = frame.tools
-                if self._session:
-                    await self._reconnect()
+            logger.warning(f"{self}: Updating tools is not currently supported in Gemini Live")
         else:
             await self.push_frame(frame, direction)
 
--- a/src/pipecat/services/google/gemini_live/llm.py
+++ b/src/pipecat/services/google/gemini_live/llm.py
@@ -950,7 +950,10 @@ class GeminiLiveLLMService(LLMService):
         elif isinstance(frame, LLMUpdateSettingsFrame):
             await self._update_settings(frame.settings)
         elif isinstance(frame, LLMSetToolsFrame):
-            await self._update_settings()
+            if frame.tools is not None:
+                self._tools_from_init = frame.tools
+                if self._session:
+                    await self._reconnect()
         else:
             await self.push_frame(frame, direction)
 


src/pipecat/services/google/gemini_live/llm.py
```diff
@@ -950,7 +950,10 @@ async def process_frame(self, frame: Frame, direction: FrameDirection):
         elif isinstance(frame, LLMUpdateSettingsFrame):
             await self._update_settings(frame.settings)
         elif isinstance(frame, LLMSetToolsFrame):
-            await self._update_settings()
+            if frame.tools is not None:
+                self._tools_from_init = frame.tools
+                if self._session:
+                    await self._reconnect()
```
aconchillo: We might not want to do this for now, reconnects are bad and slow. I believe we will wait until Gemini supports it. In the meantime we probably might want to remove the `update_settings` and probably add a warning instead saying that this is not supported in Gemini Live.

@kompfner has been looking into this and can comment.

tropxy: what I did as a workaround was to monkey patch the Gemini class locally
```python
class GeminiLiveVertexLLMService(BaseGeminiLiveVertexLLMService):
    """Patched Gemini Live service that fixes bugs in the base implementation.

    Fixes:
    1. _update_settings bug: https://github.com/pipecat-ai/pipecat/blob/main/src/pipecat/services/google/gemini_live/llm.py#L953
       The base class calls _update_settings() without the required 'settings' argument.
    2. Transcription timeout task not being properly awaited when cancelled.
    """

        """Process frames with fix for LLMSetToolsFrame handling."""
        # Fix the bug: pass current settings when handling LLMSetToolsFrame
        if isinstance(frame, LLMSetToolsFrame):
            await self._update_settings(self._settings)
            # Continue processing through parent for other frame handling
            return

        # All other frames go through normal processing

    async def cancel_task(self, task, timeout: float = 1.0):
        """Override to properly await cancelled transcription timeout tasks."""
        if task and not task.done():
            task.cancel()
            try:
                # Properly await the cancelled task to avoid RuntimeWarning
                await asyncio.wait_for(task, timeout=timeout)
            except (asyncio.CancelledError, asyncio.TimeoutError):
                # Expected when task is cancelled
                pass
            except Exception as e:
                logger.debug(f"Error cancelling task: {e}")
```

kompfner: I have an [open PR](https://github.com/pipecat-ai/pipecat/pull/3620) that adds the following (per the changelog entry):

```
- Added support to Gemini Live (`GeminiLiveLLMService`) for programmatically swapping tools or editing context at runtime; now you can use `LLMMessagesAppendFrame`, `LLMMessagesUpdateFrame`, `LLMMessagesTransformFrame`, and `LLMSetToolsFrame` with Gemini Live, like you would with text-to-text services. Note that this new functionality only works if you're using `LLMContext` and `LLMContextAggregatorPair` rather than the deprecated `OpenAILLMContext` and associated aggregators.
```

So we probably won't need this PR when that one lands.

Good catch, though, that the `LLMSetToolsFrame` handling in Gemini Live was never quite working as intended 🤦 .

Do note, however, that—as @aconchillo pointed out—a reconnect is necessary to make tool or conversation history changes at runtime. It's not _too_ slow, but it might be noticeable (though not always).

omChauhanDev: make sense, i will close this one then

PR #3628, opened 2026-02-03
Fix: Broadcast SpeechControlParamsFrame from VADController

markbackman:

Found this regression in doing some local testing. This frame needs to be broadcasted to allow for the STT TTFS measurement to work.

### DIFF
--- a/src/pipecat/processors/audio/vad_processor.py
+++ b/src/pipecat/processors/audio/vad_processor.py
@@ -93,8 +93,10 @@ class VADProcessor(FrameProcessor):
         """
         await super().process_frame(frame, direction)
 
+        # Forward the frame first, then let VAD controller process. This ensures:
+        # 1. StartFrame reaches downstream before SpeechControlParamsFrame is broadcast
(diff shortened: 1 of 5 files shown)


src/pipecat/processors/audio/vad_processor.py
```diff
@@ -93,8 +94,13 @@ async def process_frame(self, frame: Frame, direction: FrameDirection):
         """
         await super().process_frame(frame, direction)
 
-        # Let the VAD controller handle the frame
-        await self._vad_controller.process_frame(frame)
-
-        # Always forward the frame
-        await self.push_frame(frame, direction)
+        if isinstance(frame, StartFrame):
```
markbackman: @aconchillo there might be a better way to do this. I had to first push the StartFrame downstream so that the broadcasted SpeechControlParamsFrame reaches the next process _after_ the StartFrame. Otherwise, we want to push the frame after the VADController in all other cases (at least that was the prior logic).

Any other ideas?

aconchillo (approved):

aconchillo:
LGTM!

PR #3630, opened 2026-02-03
Add native RTVI function call lifecycle messages

markbackman:

### DIFF
--- a/src/pipecat/processors/frameworks/rtvi.py
+++ b/src/pipecat/processors/frameworks/rtvi.py
@@ -1585,33 +1585,6 @@ class RTVIProcessor(FrameProcessor):
         message = RTVILLMFunctionCallMessage(data=fn)
         await self.push_transport_message(message, exclude_none=False)
 
-    async def handle_function_call_start(
-        self, function_name: str, llm: FrameProcessor, context: OpenAILLMContext
-    ):
-        """Handle the start of a function call from the LLM.
-
-        .. deprecated:: 0.0.66
-            This method is deprecated and will be removed in a future version.
-            Use `RTVIProcessor.handle_function_call()` instead.
-
-        Args:
-            function_name: Name of the function being called.
-            llm: The LLM processor making the call.
-            context: The LLM context.
-        """
-        import warnings
-
-        with warnings.catch_warnings():
-            warnings.simplefilter("always")
-            warnings.warn(
-                "Function `RTVIProcessor.handle_function_call_start()` is deprecated, use `RTVIProcessor.handle_function_call()` instead.",
-                DeprecationWarning,
-            )
-
-        fn = RTVILLMFunctionCallStartMessageData(function_name=function_name)
-        message = RTVILLMFunctionCallStartMessage(data=fn)
-        await self.push_transport_message(message, exclude_none=False)
-
     async def process_frame(self, frame: Frame, direction: FrameDirection):
         """Process incoming frames through the RTVI processor.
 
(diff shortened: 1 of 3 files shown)


src/pipecat/processors/frameworks/rtvi.py
```diff
@@ -1500,33 +1585,6 @@ async def handle_function_call(self, params: FunctionCallParams):
         message = RTVILLMFunctionCallMessage(data=fn)
         await self.push_transport_message(message, exclude_none=False)
 
-    async def handle_function_call_start(
```
markbackman, line 1503: This was dead code, so just removing.

src/pipecat/processors/frameworks/rtvi.py
```diff
@@ -671,6 +676,54 @@ class RTVILLMFunctionCallResultData(BaseModel):
...
 
 
+class RTVILLMFunctionCallInProgressMessageData(BaseModel):
+    """Data for LLM function call in-progress notification.
+
+    Contains function call details including name, ID, and arguments.
+    Fields may be omitted based on the configured function_call_report_level for security.
+    """
+
+    tool_call_id: str
```
mattieruth, line 690: is there any reason this should be behind a security level?

src/pipecat/processors/frameworks/rtvi.py
```diff
@@ -943,6 +1012,11 @@ class RTVIObserverParams:
             transformed text. To register, provide a list of tuples of
             (aggregation_type | '*', transform_function).
         audio_level_period_secs: How often audio levels should be sent if enabled.
+        function_call_enabled: Indicates if function call messages should be sent.
+        function_call_report_level: Controls what information is exposed in function call
+            events for security. Use NONE to emit events without function names or metadata
+            (most secure), NAME to include function names but hide arguments and results,
+            or FULL to include all details. Defaults to FULL.
```
mattieruth: i think this should be a Map of levels to function names, with a reserved name for "*", allowing the level to apply to all function calls.  This way you can set the metadata level per function call. You might have multiple, but the client only needs (or can securely have) the args/results for one specific one.

it might be overkill to make the `function_call_enabled` a list of function names to enable it for, but IMO i think it, too, should be possible to enable on a per function call basis.

src/pipecat/processors/frameworks/rtvi.py
```diff
@@ -1500,33 +1618,6 @@ async def handle_function_call(self, params: FunctionCallParams):
         message = RTVILLMFunctionCallMessage(data=fn)
         await self.push_transport_message(message, exclude_none=False)
```
mattieruth, line 1679: This should be marked deprecated and give pointers to use `enable_function_call` and a `FULL` report level on the processor to achieve this.

src/pipecat/processors/frameworks/rtvi.py
mattieruth: We will need to bump the version with this change to 1.2.0

src/pipecat/processors/frameworks/rtvi.py
markbackman: My understanding is that `tool_call_id` is just a random ID and has no information encoded to it. So I think it's fine to be exposed as part of `NONE.

src/pipecat/processors/frameworks/rtvi.py
markbackman: Good idea! We can probably get rid of `function_call_enabled` if we go this route and just use `function_call_report_level` to control everything.

mattieruth (approved):

PR #3632, opened 2026-02-03
Flush assistant text to context before processing tool calls

### LINKED ISSUE #3631: Assistant text not in context when tool calls are processed
# Assistant text not in context when tool calls are processed

## Problem

When an LLM returns both text content and a tool call, the text content is only added to context after `LLMFullResponseEndFrame` arrives. However, tool call handlers are invoked earlier via `FunctionCallInProgressFrame`, meaning the assistant's spoken text is not yet in context when the tool executes.

This causes issues for frameworks like pipecat-flows where tool calls trigger node transitions that rebuild context - the assistant's acknowledgment text is missing because it hasn't been added yet.

## Example Logs

```
15:33:14 | DEBUG | ElevenLabsTTSService#1: Generating TTS [Selvä Pekka, laitan soittopyynnön tällä numerolla.]
15:33:15 | DEBUG | Bot started speaking
15:33:15 | DEBUG   |
=== LLM RESPONSE ===
Content: Selvä Pekka, laitan soittopyynnön tällä numerolla. Yhdistän järjestelmään hetkeksi.
Tool calls (1):
  - callback_end_call: {}
=== END RESPONSE ===
15:33:15 | DEBUG | SointuOpenAILLMService#0 Calling function [callback_end_call:call_iwLjUqehLxzB1fTeStXDFROu]
15:33:15 | DEBUG | Function called: callback_end_call
15:33:15 | DEBUG | Transition-only function called for callback_end_call   ← node transition happens here
15:33:15 | DEBUG | LLM response complete, captured: 'Selvä Pekka...'        ← text captured AFTER transition
15:33:15 | DEBUG | LLMAssistantAggregator#0 FunctionCallsStartedFrame: ['callback_end_call:...']
```

Notice the timing: the function executes and triggers a node transition **before** the text is added to context. The text "Selvä Pekka, laitan soittopyynnön tällä numerolla." is spoken but not in context when the transition rebuilds it.

## Current Flow

```
1. LLMFullResponseStartFrame    → _handle_llm_start()
2. TextFrame, TextFrame, ...    → _handle_text() accumulates in self._aggregation
3. FunctionCallInProgressFrame  → _handle_function_call_in_progress()
                                  → adds tool_calls message to context
                                  → tool handler executes (may rebuild context)
4. LLMFullResponseEndFrame      → _handle_llm_end()
                                  → push_aggregation() adds text to context
                                  → BUT context may have already been rebuilt!
```

The text accumulated in step 2 isn't added to context until step 4, but the tool call in step 3 may have already triggered actions that need that context.

## Proposed Solution

Flush accumulated text to context before processing tool calls:

```python
async def _handle_function_call_in_progress(self, frame: FunctionCallInProgressFrame):
    # Flush accumulated text to context BEFORE processing tool call
    await self.push_aggregation()

    # Then add the tool call message
    self._context.add_message({
        "role": "assistant",
        "tool_calls": [...]
    })
    # ... rest unchanged
```

This ensures context contains:
```
{"role": "assistant", "content": "Selvä Pekka, laitan soittopyynnön tällä numerolla."}
{"role": "assistant", "tool_calls": [...]}
{"role": "tool", "content": "IN_PROGRESS", ...}
```

## Trade-off

**Current behavior**: Context only contains text that was fully spoken (added after `LLMFullResponseEndFrame`)

**Proposed behavior**: Context contains text as soon as it's generated, even if speech is still in progress or gets interrupted

This means if the user interrupts mid-speech, the context would contain text that wasn't fully spoken to the user.

**Why this trade-off is acceptable**:

**Reduced perceived latency** - Tool calls can execute immediately without waiting for speech to complete, which can significantly reduce response time for multi-step interactions

## Environment

- Pipecat version: 0.0.99
- Discovered while using pipecat-flows with node transitions

OlliTapio:
## Summary
- Fixes #3631: Assistant text is now added to context before tool calls are processed
- Ensures frameworks like pipecat-flows have correct context when tool handlers trigger node transitions
- Adds tests for text ordering with function calls and cancellation scenarios

Closes #3631

### DIFF
--- a/src/pipecat/processors/aggregators/llm_response.py
+++ b/src/pipecat/processors/aggregators/llm_response.py
@@ -968,6 +968,12 @@ class LLMAssistantContextAggregator(LLMContextResponseAggregator):
     async def _handle_function_calls_started(self, frame: FunctionCallsStartedFrame):
         function_names = [f"{f.function_name}:{f.tool_call_id}" for f in frame.function_calls]
         logger.debug(f"{self} FunctionCallsStartedFrame: {function_names}")
+
+        # Flush any pending text to context BEFORE tool calls are added.
+        # This ensures assistant text appears before tool_calls in context,
+        # even if LLMFullResponseEndFrame hasn't been processed yet.
+        await self.push_aggregation()
+
         for function_call in frame.function_calls:
             self._function_calls_in_progress[function_call.tool_call_id] = None
 
@@ -975,10 +981,6 @@ class LLMAssistantContextAggregator(LLMContextResponseAggregator):
         logger.debug(
             f"{self} FunctionCallInProgressFrame: [{frame.function_name}:{frame.tool_call_id}]"
         )
-
-        # Flush any accumulated text to context BEFORE processing the tool call.
-        # This ensures the assistant's text is in context when the tool executes.
-        await self.push_aggregation()
 
         await self.handle_function_call_in_progress(frame)
         self._function_calls_in_progress[frame.tool_call_id] = frame
@@ -1014,6 +1016,8 @@ class LLMAssistantContextAggregator(LLMContextResponseAggregator):
                 run_llm = not bool(self._function_calls_in_progress)
 
         if run_llm:
+            # Flush any pending text before triggering LLM to ensure complete context
+            await self.push_aggregation()
             await self.push_context_frame(FrameDirection.UPSTREAM)
 
         # Call the `on_context_updated` callback once the function call result
--- a/src/pipecat/processors/aggregators/llm_response_universal.py
+++ b/src/pipecat/processors/aggregators/llm_response_universal.py
@@ -929,6 +929,12 @@ class LLMAssistantAggregator(LLMContextAggregator):
     async def _handle_function_calls_started(self, frame: FunctionCallsStartedFrame):
         function_names = [f"{f.function_name}:{f.tool_call_id}" for f in frame.function_calls]
         logger.debug(f"{self} FunctionCallsStartedFrame: {function_names}")
+
+        # Flush any pending text to context BEFORE tool calls are added.
+        # This ensures assistant text appears before tool_calls in context,
+        # even if LLMFullResponseEndFrame hasn't been processed yet.
+        await self.push_aggregation()
+
         for function_call in frame.function_calls:
             self._function_calls_in_progress[function_call.tool_call_id] = None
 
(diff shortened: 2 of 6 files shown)


src/pipecat/processors/aggregators/llm_response.py
```diff
@@ -975,6 +975,11 @@ async def _handle_function_call_in_progress(self, frame: FunctionCallInProgressF
         logger.debug(
             f"{self} FunctionCallInProgressFrame: [{frame.function_name}:{frame.tool_call_id}]"
         )
+
+        # Flush any accumulated text to context BEFORE processing the tool call.
+        # This ensures the assistant's text is in context when the tool executes.
+        await self.push_aggregation()
```
OlliTapio: We are using OpenAi so this affects us.

src/pipecat/processors/aggregators/llm_response_universal.py
```diff
@@ -937,6 +937,10 @@ async def _handle_function_call_in_progress(self, frame: FunctionCallInProgressF
             f"{self} FunctionCallInProgressFrame: [{frame.function_name}:{frame.tool_call_id}]"
         )
 
+        # Flush any accumulated text to context BEFORE adding the tool call.
+        # This ensures the assistant's text is in context when the tool executes.
+        await self.push_aggregation()
+
```
OlliTapio: Don't know if this is necessary for this. We use OpenAi so don't know if this is necessary or beneficial change.

OlliTapio:
With additional testing this was not as simple as this. There are cases where the message spoken by TTS is now added twice to the context.

OlliTapio:
Closing this PR - further investigation revealed that this approach doesn't fully solve the issue.

## Why This Doesn't Work

The root cause is **pipeline ordering and frame routing**, not just when we flush the aggregator:

1. `LLMTextFrame` uses `push_frame()` → travels sequentially through the pipeline
2. `FunctionCallsStartedFrame` uses `broadcast_frame()` → takes a different/faster path

When the assistant aggregator is positioned after TTS (common for capturing "actually spoken" text), text frames must travel through TTS before reaching the aggregator. Meanwhile, `broadcast_frame` delivers `FunctionCallsStartedFrame` more directly.

By the time the aggregator receives `FunctionCallsStartedFrame` and tries to flush, **the text frames haven't arrived yet** - the aggregation buffer is empty.

See updated analysis in #3631.

PR #3637, opened 2026-02-04
Improve user turn stop timing by triggering timeout from VAD stop, push STT metadata to user aggregator

markbackman:
## Summary

This PR improves user turn stop timing by leveraging `TranscriptionFrame` finalization status and STT service latency metrics (P99 time-to-final-segment) to dynamically configure timeouts. Both changes result in more predictable and optimized turn detection.

## Key Changes

- **Improved turn stop timing**: Timeout now starts on VADUserStoppedSpeakingFrame instead of transcript events, providing tighter and more predictable timing
- **STT metadata broadcasting**: STT services now broadcast their P99 latency via `STTMetadataFrame`, allowing turn strategies to auto-configure timeouts
- **Finalized transcript support**: Services that support transcript finalization can trigger turn stop earlier
- **ServiceSwitcher integration**: Proper handling of STT metadata when switching between services

## Breaking Changes

- ⚠️ Renamed `timeout` parameter to `user_resume_speaking_timeout` in `TranscriptionUserTurnStopStrategy`
- ⚠️ Removed `timeout` parameter from `TurnAnalyzerUserTurnStopStrategy` (now managed internally based on STT latency)

### Turn Stop Strategy Improvements

- Timeout starts on `VADUserStoppedSpeakingFrame` for tighter timing
- Added support for `TranscriptionFrame.finalized=True` to trigger earlier
- Added handling for `SpeechControlParamsFrame` to adjust timeout based on VAD `stop_secs`
- Added fallback timeout for edge cases where transcripts arrive without VAD events
- Removed `InterimTranscriptionFrame` handling (no longer affects timing)

### STT Metadata Broadcasting

- New `STTMetadataFrame` (inherits from `ServiceMetadataFrame`) broadcasts P99 time-to-final-segment latency
STT services push this metadata at pipeline start, after `StartFrame`
- Turn strategies automatically configure their internal `_stt_timeout` from this metadata
- Measured P99 values added for: Deepgram (0.39s), Cartesia (0.55s), ElevenLabs Realtime (0.41s) (PLACEHOLDER. NEED TO UPDATE ALL SERVICES!)
- Services without explicit values fall back to 1.0s with a warning

### ServiceSwitcher Support

Added `RequestMetadataFrame` and metadata handling for `ServiceSwitcher` to ensure STT services correctly emit `STTMetadataFrame` when used with `ServiceSwitcher`. Only the active service's metadata is propagated downstream, switching services triggers the newly active service to re-emit its metadata, and proper frame ordering is maintained at startup.

**Problems addressed during implementation**

When STT services are used within a `ServiceSwitcher`, there were several issues with `STTMetadataFrame` handling:
- Both services emit metadata at startup: Since `StartFrame` passes through to all services (lifecycle frame), both active and inactive services would push `STTMetadataFrame`, causing downstream processors to receive metadata from both.
- Ordering issues at startup: Due to `ParallelPipeline` synchronizing `StartFrame` release across branches while pushing other frames immediately, `STTMetadataFrame` could arrive at downstream processors before `StartFrame`, causing errors.
- No metadata on service switch: When switching from one STT service to another, downstream processors (like turn strategies) wouldn't receive the new service's metadata.

**New ControlFrame**

Added a `RequestMetadataFrame`, a new control frame that can be used to request a `ServiceMetadataFrame`.

**ServiceSwitcher Changes**

The ServiceSwitcherFilter now:
- Blocks `ServiceMetadataFrame` from inactive services: Only the active service's metadata passes through
- Blocks metadata during startup: Until `StartFrame` is released by the `ParallelPipeline`
- Requests metadata after startup completes: Sends `RequestMetadataFrame` to the active service after `StartFrame` is released
- Requests metadata on service switch: When a service becomes active, requests its metadata

The key architectural insight is that two filters "sandwich" each service (upstream and downstream), and the upstream filter for the active service sends `RequestMetadataFrame` after updating its state.

**STTService Changes**

Handles `RequestMetadataFrame` by re-emitting `STTMetadataFrame` via `_push_stt_metadata()`.

## TODOs

- [ ] Add final P99 values for services

Also, fixes: https://github.com/pipecat-ai/pipecat/issues/3643

### DIFF
--- a/src/pipecat/pipeline/service_switcher.py
+++ b/src/pipecat/pipeline/service_switcher.py
@@ -6,11 +6,10 @@
 
 """Service switcher for switching between different services at runtime, with different switching strategies."""
 
-from dataclasses import dataclass
+from abc import abstractmethod
 from typing import Any, Generic, List, Optional, Type, TypeVar
 
 from pipecat.frames.frames import (
-    ControlFrame,
     Frame,
     ManuallySwitchServiceFrame,
     RequestMetadataFrame,
@@ -20,14 +19,25 @@ from pipecat.frames.frames import (
 from pipecat.pipeline.parallel_pipeline import ParallelPipeline
 from pipecat.processors.filters.function_filter import FunctionFilter
 from pipecat.processors.frame_processor import FrameDirection, FrameProcessor
+from pipecat.utils.base_object import BaseObject
 
 
-class ServiceSwitcherStrategy:
+class ServiceSwitcherStrategy(BaseObject):
     """Base class for service switching strategies.
 
     Note:
         Strategy classes are instantiated internally by ServiceSwitcher.
         Developers should pass the strategy class (not an instance) to ServiceSwitcher.
+
+    Event handlers available:
+
+    - on_service_switched: Called when the active service changes.
+
+    Example::
+
+        @strategy.event_handler("on_service_switched")
+        async def on_service_switched(strategy, service):
+            ...
     """
 
     def __init__(self, services: List[FrameProcessor]):
@@ -39,20 +49,42 @@ class ServiceSwitcherStrategy:
         Args:
             services: List of frame processors to switch between.
         """
-        self.services = services
-        self.active_service: Optional[FrameProcessor] = None
+        super().__init__()
 
-    def handle_frame(self, frame: ServiceSwitcherFrame, direction: FrameDirection):
+        if len(services) == 0:
+            raise Exception(f"ServiceSwitcherStrategy needs at least one service")
+
+        self._services = services
+        self._active_service = services[0]
+
+        self._register_event_handler("on_service_switched")
+
+    @property
+    def services(self) -> List[FrameProcessor]:
+        """Return the list of available services."""
+        return self._services
+
+    @property
+    def active_service(self) -> FrameProcessor:
+        """Return the currently active service."""
+        return self._active_service
+
+    @abstractmethod
+    async def handle_frame(
+        self, frame: ServiceSwitcherFrame, direction: FrameDirection
+    ) -> Optional[FrameProcessor]:
         """Handle a frame that controls service switching.
 
-        This method can be overridden by subclasses to implement specific logic
-        for handling frames that control service switching.
+        Subclasses implement this to decide whether a switch should occur.
 
         Args:
             frame: The frame to handle.
             direction: The direction of the frame (upstream or downstream).
+
+        Returns:
+            The newly active service if a switch occurred, or None otherwise.
         """
-        raise NotImplementedError("Subclasses must implement this method.")
+        pass
 
 
 class ServiceSwitcherStrategyManual(ServiceSwitcherStrategy):
@@ -69,31 +101,24 @@ class ServiceSwitcherStrategyManual(ServiceSwitcherStrategy):
         )
     """
 
-    def __init__(self, services: List[FrameProcessor]):
-        """Initialize the manual service switcher strategy with a list of services.
-
-        Note:
-            This is called internally by ServiceSwitcher. Do not instantiate directly.
-
-        Args:
-            services: List of frame processors to switch between.
-        """
-        super().__init__(services)
-        self.active_service = services[0] if services else None
-
-    def handle_frame(self, frame: ServiceSwitcherFrame, direction: FrameDirection):
+    async def handle_frame(
+        self, frame: ServiceSwitcherFrame, direction: FrameDirection
+    ) -> Optional[FrameProcessor]:
         """Handle a frame that controls service switching.
 
         Args:
             frame: The frame to handle.
             direction: The direction of the frame (upstream or downstream).
+
+        Returns:
+            The newly active service if a switch occurred, or None otherwise.
         """
         if isinstance(frame, ManuallySwitchServiceFrame):
-            self._set_active_if_available(frame.service)
-        else:
-            raise ValueError(f"Unsupported frame type: {type(frame)}")
+            return await self._set_active_if_available(frame.service)
 
-    def _set_active_if_available(self, service: FrameProcessor):
+        return None
+
+    async def _set_active_if_available(self, service: FrameProcessor) -> Optional[FrameProcessor]:
         """Set the active service to the given one, if it is in the list of available services.
 
         If it's not in the list, the request is ignored, as it may have been
@@ -101,16 +126,35 @@ class ServiceSwitcherStrategyManual(ServiceSwitcherStrategy):
 
         Args:
             service: The service to set as active.
+
+        Returns:
+            The newly active service, or None if the service was not found.
         """
         if service in self.services:
-            self.active_service = service
+            self._active_service = service
+            await self._call_event_handler("on_service_switched", service)
+            return service
+        return None
 
 
 StrategyType = TypeVar("StrategyType", bound=ServiceSwitcherStrategy)
 
 
 class ServiceSwitcher(ParallelPipeline, Generic[StrategyType]):
-    """A pipeline that switches between different services at runtime."""
+    """Parallel pipeline that routes frames to one active service at a time.
+
+    Wraps each service in a pair of filters that gate frame flow based on
+    which service is currently active. Switching is controlled by
+    `ServiceSwitcherFrame` frames and delegated to a pluggable
+    `ServiceSwitcherStrategy`.
+
+    Example::
+
+        switcher = ServiceSwitcher(
+            services=[stt_1, stt_2],
+            strategy_type=ServiceSwitcherStrategyManual,
+        )
+    """
 
     def __init__(self, services: List[FrameProcessor], strategy_type: Type[StrategyType]):
         """Initialize the service switcher with a list of services and a switching strategy.
@@ -119,101 +163,20 @@ class ServiceSwitcher(ParallelPipeline, Generic[StrategyType]):
             services: List of frame processors to switch between.
             strategy_type: The strategy class to use for switching between services.
         """
-        strategy = strategy_type(services)
-        super().__init__(*self._make_pipeline_definitions(services, strategy))
-        self.services = services
-        self.strategy = strategy
+        _strategy = strategy_type(services)
+        super().__init__(*self._make_pipeline_definitions(services, _strategy))
+        self._services = services
+        self._strategy = _strategy
 
-    class ServiceSwitcherFilter(FunctionFilter):
-        """An internal filter that gates frame flow based on active service.
+    @property
+    def strategy(self) -> StrategyType:
+        """Return the active switching strategy."""
+        return self._strategy
 
-        Two filters "sandwich" each service, allowing frames through only
-        when the wrapped service is active. The pipeline layout is::
-
-            DownstreamFilter → Service → UpstreamFilter
-
-        The filter names refer to which *direction* of frame flow they
-        filter, not their physical position: the downstream filter sits
-        *before* the service (filtering frames flowing into it) and the
-        upstream filter sits *after* it (filtering frames flowing back out).
-        """
-
-        def __init__(
-            self,
-            wrapped_service: FrameProcessor,
-            active_service: FrameProcessor,
-            direction: FrameDirection,
-        ):
-            """Initialize the service switcher filter with a strategy and direction.
-
-            Args:
-                wrapped_service: The service that this filter wraps.
-                active_service: The currently active service.
-                direction: The direction of frame flow this filter gates
-                    (DOWNSTREAM for the filter before the service,
-                    UPSTREAM for the filter after it).
-            """
-            self._wrapped_service = wrapped_service
-            self._active_service = active_service
-
-            async def filter(_: Frame) -> bool:
-                return self._wrapped_service == self._active_service
-
-            super().__init__(filter, direction, filter_system_frames=True)
-
-        async def process_frame(self, frame, direction):
-            """Process a frame through the filter, handling special internal filter-updating frames."""
-            if isinstance(frame, ServiceSwitcher.ServiceSwitcherFilterFrame):
-                old_active = self._active_service
-                self._active_service = frame.active_service
-                # Two ServiceSwitcherFilters "sandwich" a service. The
-                # frame enters via the downstream filter first. Push it
-                # through so the upstream filter also updates its state.
-                if direction == self._direction:
-                    await self.push_frame(frame, direction)
-                # This is the upstream filter (the second to update). At
-                # this point both filters know the new active service, so
-                # it's safe to request metadata — the resulting
-                # ServiceMetadataFrame will pass both filters on its way
-                # out. Only do this for the newly active service's sandwich.
-                elif (
-                    self._direction == FrameDirection.UPSTREAM
-                    and self._wrapped_service == frame.active_service
-                    and old_active != self._wrapped_service
-                ):
-                    await self.push_frame(RequestMetadataFrame(), FrameDirection.UPSTREAM)
-                return
-
-            # RequestMetadataFrame is pushed upstream by the upstream filter
-            # (above) and consumed by the service. Guard against services
-            # that don't consume it: only forward in the filter's own
-            # direction (so it can reach the service) and only for the
-            # active service. Block in all other cases to prevent it from
-            # escaping the sandwich.
-            if isinstance(frame, RequestMetadataFrame):
-                if direction == self._direction and self._wrapped_service == self._active_service:
-                    await self.push_frame(frame, direction)
-                return
-
-            # Block ServiceMetadataFrame from inactive services.
-            if isinstance(frame, ServiceMetadataFrame):
-                if self._wrapped_service != self._active_service:
-                    return
-                await self.push_frame(frame, direction)
-                return
-
-            await super().process_frame(frame, direction)
-
-    @dataclass
-    class ServiceSwitcherFilterFrame(ControlFrame):
-        """An internal frame used to update filter state on service switch.
-
-        Sent when a service switch occurs to update the active service in
-        the sandwich filters and trigger metadata emission from the newly
-        active service.
-        """
-
-        active_service: FrameProcessor
+    @property
+    def services(self) -> List[FrameProcessor]:
+        """Return the list of available services."""
+        return self._services
 
     @staticmethod
     def _make_pipeline_definitions(
@@ -228,20 +191,52 @@ class ServiceSwitcher(ParallelPipeline, Generic[StrategyType]):
     def _make_pipeline_definition(
         service: FrameProcessor, strategy: ServiceSwitcherStrategy
     ) -> Any:
-        # Layout: DownstreamFilter → Service → UpstreamFilter
+        async def filter(_: Frame) -> bool:
+            return service == strategy.active_service
+
+        # Layout: Filter → Service → Filter
+        #
+        # filter_system_frames: we want to run filter functions also on system
+        # frames.
+        #
+        # enable_direct_mode: filter functions are quick so we don't need
+        # additional tasks.
         return [
-            ServiceSwitcher.ServiceSwitcherFilter(
-                wrapped_service=service,
-                active_service=strategy.active_service,
+            FunctionFilter(
+                filter=filter,
                 direction=FrameDirection.DOWNSTREAM,
+                filter_system_frames=True,
+                enable_direct_mode=True,
             ),
             service,
-            ServiceSwitcher.ServiceSwitcherFilter(
-                wrapped_service=service,
-                active_service=strategy.active_service,
+            FunctionFilter(
+                filter=filter,
                 direction=FrameDirection.UPSTREAM,
+                filter_system_frames=True,
+                enable_direct_mode=True,
             ),
         ]
+
+    async def push_frame(self, frame: Frame, direction: FrameDirection = FrameDirection.DOWNSTREAM):
+        """Push a frame out of the service switcher.
+
+        Suppresses `RequestMetadataFrame` (internal to the switcher) and
+        `ServiceMetadataFrame` from inactive services so only the active
+        service's metadata reaches downstream processors. One case this happens
+        is with `StartFrame` since all the filters let it pass, and `StartFrame`
+        causes the service to generate `ServiceMetadataFrame`.
+
+        """
+        # Don't let RequestMetadataFrame out.
+        if isinstance(frame, RequestMetadataFrame):
+            return
+
+        # Only let metadata from the active service escape.
+        if isinstance(frame, ServiceMetadataFrame):
+            if frame.service_name != self.strategy.active_service.name:
+                return
+
+        await super().push_frame(frame, direction)
 
     async def process_frame(self, frame: Frame, direction: FrameDirection):
         """Process a frame, handling frames which affect service switching.
@@ -250,11 +245,16 @@ class ServiceSwitcher(ParallelPipeline, Generic[StrategyType]):
             frame: The frame to process.
             direction: The direction of the frame (upstream or downstream).
         """
-        await super().process_frame(frame, direction)
-
         if isinstance(frame, ServiceSwitcherFrame):
-            self.strategy.handle_frame(frame, direction)
-            service_switcher_filter_frame = ServiceSwitcher.ServiceSwitcherFilterFrame(
-                active_service=self.strategy.active_service
-            )
-            await super().process_frame(service_switcher_filter_frame, direction)
+            service = await self.strategy.handle_frame(frame, direction)
+
+            # If we don't switch to a new service we need to keep processing the
+            # frame. If we switched, we just swallow the frame.
+            if not service:
+                await super().process_frame(frame, direction)
+
+            # If we switched to a new service, request its metadata.
+            if service:
+                await service.queue_frame(RequestMetadataFrame())
+        else:
+            await super().process_frame(frame, direction)
--- a/src/pipecat/pipeline/llm_switcher.py
+++ b/src/pipecat/pipeline/llm_switcher.py
@@ -44,7 +44,7 @@ class LLMSwitcher(ServiceSwitcher[StrategyType]):
         return self.services
 
     @property
-    def active_llm(self) -> Optional[LLMService]:
+    def active_llm(self) -> LLMService:
         """Get the currently active LLM.
 
         Returns:
--- a/src/pipecat/processors/filters/function_filter.py
+++ b/src/pipecat/processors/filters/function_filter.py
@@ -10,10 +10,12 @@ This module provides a processor that filters frames based on a custom function,
 allowing for flexible frame filtering logic in processing pipelines.
 """
 
-from typing import Awaitable, Callable
+from typing import Awaitable, Callable, Optional
 
 from pipecat.frames.frames import CancelFrame, EndFrame, Frame, StartFrame, SystemFrame
 from pipecat.processors.frame_processor import FrameDirection, FrameProcessor
+
+FilterType = Callable[[Frame], Awaitable[bool]]
 
 
 class FunctionFilter(FrameProcessor):
@@ -26,9 +28,10 @@ class FunctionFilter(FrameProcessor):
 
     def __init__(
         self,
-        filter: Callable[[Frame], Awaitable[bool]],
-        direction: FrameDirection = FrameDirection.DOWNSTREAM,
+        filter: FilterType,
+        direction: Optional[FrameDirection] = FrameDirection.DOWNSTREAM,
         filter_system_frames: bool = False,
+        **kwargs,
     ):
         """Initialize the function filter.
 
@@ -36,10 +39,13 @@ class FunctionFilter(FrameProcessor):
             filter: An async function that takes a Frame and returns True if the
                 frame should pass through, False otherwise.
             direction: The direction to apply filtering. Only frames moving in
-                this direction will be filtered. Defaults to DOWNSTREAM.
+                this direction will be filtered; frames in the other direction
+                pass through unfiltered. If None, frames in both directions
+                are filtered. Defaults to DOWNSTREAM.
             filter_system_frames: Whether to filter system frames. Defaults to False.
+            **kwargs: Additional arguments passed to parent class.
         """
-        super().__init__()
+        super().__init__(**kwargs)
         self._filter = filter
         self._direction = direction
         self._filter_system_frames = filter_system_frames
@@ -51,7 +57,7 @@ class FunctionFilter(FrameProcessor):
     def _should_passthrough_frame(self, frame, direction):
         """Check if a frame should pass through without filtering."""
         # Always passthrough frames in the wrong direction
-        if direction != self._direction:
+        if self._direction and direction != self._direction:
             return True
 
         # Always passthrough lifecycle frames
--- a/tests/test_filters.py
+++ b/tests/test_filters.py
@@ -14,10 +14,12 @@ from pipecat.frames.frames import (
     UserStartedSpeakingFrame,
     UserStoppedSpeakingFrame,
 )
+from pipecat.pipeline.pipeline import Pipeline
 from pipecat.processors.filters.frame_filter import FrameFilter
 from pipecat.processors.filters.function_filter import FunctionFilter
 from pipecat.processors.filters.identity_filter import IdentityFilter
 from pipecat.processors.filters.wake_check_filter import WakeCheckFilter
+from pipecat.processors.frame_processor import FrameDirection, FrameProcessor
 from pipecat.tests.utils import run_test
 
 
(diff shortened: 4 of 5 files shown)


src/pipecat/services/stt_latency.py
```diff
@@ -0,0 +1,50 @@
+#
+# Copyright (c) 2024-2026, Daily
+#
+# SPDX-License-Identifier: BSD 2-Clause License
+#
+
+"""STT service latency defaults.
+
+This module contains P99 time-to-final-segment (TTFS) latency values for STT
```
markbackman, line 9: Is P99 too conservative?

P99 means that we'll wait for 99% of transcripts received, leaving 1% of cases where the transcripts come in later than the allocated time. This feels better than P95, where 5% of cases fail. So, while conservative, it feels like the right call. Thoughts?

Note: for services that finalize, they don't wait these values; the P99 time is a fallback. For other services that don't finalize, they will wait to ensure all transcripts are received.

filipi87: Agreed, P99 sounds a lot safer than using P95.

changelog/3637.added.md
```diff
@@ -0,0 +1,6 @@
+- Added `STTMetadataFrame` to broadcast STT service latency information at pipeline start.
+  - STT services broadcast P99 time-to-final-segment (`ttfs_p99_latency`) to downstream processors
+  - Turn stop strategies automatically configure their STT timeout from this metadata
+  - Developers can override `ttfs_p99_latency` via constructor argument for custom deployments
+  - Added measured P99 values for STT providers.
+  - See [stt-benchmark](https://github.com/pipecat-ai/stt-benchmark) to measure latency for your configuration
```
filipi87, line 6: 🙌💯

src/pipecat/services/stt_latency.py
```diff
@@ -0,0 +1,50 @@
...
+
+# Conservative fallback for services without measured values
+DEFAULT_TTFS_P99: float = 1.0
+
+# Measured P99 TTFS latency values (in seconds)
+CARTESIA_TTFS_P99: float = 0.55
+DEEPGRAM_TTFS_P99: float = 0.39
+ELEVENLABS_REALTIME_TTFS_P99: float = 0.41
+
+# Placeholder values - TODO: Replace with measured values
```
filipi87: Are we already going to include the measured values for these services ?

markbackman: Yes, I've nearly finished all runs for these. I'll push updates for them now.

src/pipecat/turns/user_stop/transcription_user_turn_stop_strategy.py
```diff
@@ -23,34 +25,47 @@
...
-                slightly delayed transcriptions.
+            user_resume_speaking_timeout: Time to wait for the user to potentially
+                say more after they pause speaking. Defaults to 0.6 seconds.
             **kwargs: Additional keyword arguments.
         """
         super().__init__(**kwargs)
-        self._timeout = timeout
+        self._user_resume_speaking_timeout = user_resume_speaking_timeout
+        self._stt_timeout: float = 0.0  # STT P99 latency from STTMetadataFrame
+        self._stop_secs: float = 0.0  # VAD stop_secs from SpeechControlParamsFrame
```
filipi87: What if `VADUserStoppedSpeakingFrame` included the `time` when the user stopped speaking, as well as the `stop_secs` value it used internally ?

It feels like it would make more sense to have this information directly in the frame, so it’s clear how it was triggered. Then we could simply use it when we receive the `VADUserStoppedSpeakingFrame`.

That way, we wouldn’t need to handle `SpeechControlParamsFrame` or keep the `self._stop_secs` variable here. 

Instead, we’d get this information from `VADUserStoppedSpeakingFrame` and use it inside `_handle_vad_user_stopped_speaking`.

What do you think ?

src/pipecat/turns/user_stop/transcription_user_turn_stop_strategy.py
```diff
@@ -78,50 +92,123 @@ async def process_frame(self, frame: Frame):
...
-        # We just got a final result, so let's reset interim results.
-        self._seen_interim_results = False
-        # Reset aggregation timer.
-        self._event.set()
+        if frame.finalized:
+            self._transcript_finalized = True
+            # For finalized transcripts, check if we can trigger early
+            await self._maybe_trigger_user_turn_stopped()
+
+        # Fallback: handle transcripts when no VAD stop was received.
```
filipi87: Nice catch on this one.

src/pipecat/turns/user_stop/transcription_user_turn_stop_strategy.py
```diff
@@ -78,50 +92,123 @@ async def process_frame(self, frame: Frame):
...
+          and user_resume_speaking_timeout has elapsed
+        """
+        if self._vad_user_speaking or not self._text:
+            return
+
+        # For finalized transcripts, check if user_resume_speaking_timeout has elapsed.
+        # If elapsed, trigger user turn stopped immediately. Else, wait for user resume
+        # speaking timeout.
+        if self._transcript_finalized and self._vad_stopped_time is not None:
+            elapsed = time.time() - self._vad_stopped_time
```
filipi87: Shouldn’t we also adjust the elapsed time here by reducing `stop_secs`, since that time has already elapsed ?

src/pipecat/turns/user_stop/turn_analyzer_user_turn_stop_strategy.py
```diff
@@ -33,32 +34,39 @@ class TurnAnalyzerUserTurnStopStrategy(BaseUserTurnStopStrategy):
...
         Args:
             turn_analyzer: The turn detection analyzer instance to detect end of user turn.
-            timeout: Short delay used internally to handle frame timing and event triggering.
             **kwargs: Additional keyword arguments.
         """
         super().__init__(**kwargs)
         self._turn_analyzer = turn_analyzer
-        self._timeout = timeout
+        self._stt_timeout: float = 0.0  # STT P99 latency from STTMetadataFrame
+        self._stop_secs: float = 0.0  # VAD stop_secs from SpeechControlParamsFrame
```
filipi87: Same here in case we add this to the `VADUserStoppedSpeakingFrame`.

src/pipecat/services/stt_latency.py
markbackman: The fact that this is configurable allows developers to specify their own value based on their config and risk tolerance/use case.

src/pipecat/turns/user_stop/transcription_user_turn_stop_strategy.py
markbackman: Really great idea. That would simplify things. I'm going to try this out now.

src/pipecat/turns/user_stop/turn_analyzer_user_turn_stop_strategy.py
markbackman: Done

src/pipecat/turns/user_stop/transcription_user_turn_stop_strategy.py
markbackman: I'm not sure. I guess there's a question of when is the user finished speaking? In this case, it might be best for the VAD to be the thing that indicates when speaking is finished. WDYT? (Honestly, I could make a case for either one, so it's just a matter of figure out the convention and documenting it.)

src/pipecat/turns/user_stop/transcription_user_turn_stop_strategy.py
markbackman: I've done the same for `VADUserStartedSpeakingFrame` and `start_secs`, too.

markbackman:
@filipi87 I've made changes. I'm running the last few benchmarks now and will have the values soon. Hoping to have this PR ready for the release today, if everyone agrees to the change.

src/pipecat/turns/user_stop/transcription_user_turn_stop_strategy.py
filipi87: Yeah, I feel the same. My only point in favor of considering it here is that we already account for it when reducing the timeout defined for the STT services, so this would be more consistent.

On the other hand, from the user’s perspective, it might be easier to understand if `user_resume_speaking_timeout` only starts being considered once we receive the `UserStoppedSpeakingFrame`. In that case, the current approach makes sense.

I’ll leave it up to you to decide. 😅

src/pipecat/frames/frames.py
```diff
@@ -1255,16 +1255,28 @@ def __post_init__(self):
 
 @dataclass
 class VADUserStartedSpeakingFrame(SystemFrame):
```
filipi87, line 1273: Would it also make sense to include the `time` here, in both frames ?

filipi87 (approved):
@markbackman , this looks great. I have tested with a couple of different STT providers, and the results are awesome.
Really nice job.  🚀

src/pipecat/frames/frames.py
markbackman: Great idea! Added.

src/pipecat/frames/frames.py
```diff
@@ -1255,16 +1256,32 @@ def __post_init__(self):
...
-    pass
+    Parameters:
+        start_secs: The VAD start_secs duration that was used to confirm the user
+            started speaking. This represents the speech duration that had to
+            elapse before the VAD determined speech began.
+        timestamp: Wall-clock time when the VAD made its determination.
+    """
+
+    start_secs: float = 0.0
+    timestamp: float = field(default_factory=time.time)
```
aconchillo, line 1284: Nice

src/pipecat/frames/frames.py
```diff
@@ -1636,6 +1653,49 @@ class SpeechControlParamsFrame(SystemFrame):
...
+    Broadcast by STT services to inform downstream processors (like turn
+    strategies) about STT latency characteristics.
+
+    Parameters:
+        ttfs_p99_latency: Time to final segment P99 latency in seconds.
+            This is the expected time from when speech ends to when the
+            final transcript is received, at the 99th percentile.
+    """
+
+    ttfs_p99_latency: float
```
aconchillo, line 1698: 👌

src/pipecat/observers/user_bot_latency_observer.py
```diff
@@ -73,7 +73,7 @@ async def on_push_frame(self, data: FramePushed):
             self._user_stopped_time = None
         elif isinstance(data.frame, VADUserStoppedSpeakingFrame):
             # Record timestamp when user stops speaking
-            self._user_stopped_time = time.time()
+            self._user_stopped_time = data.frame.timestamp
```
aconchillo: More accurate! 👌

src/pipecat/frames/frames.py
```diff
@@ -1636,6 +1653,49 @@ class SpeechControlParamsFrame(SystemFrame):
...
+
+
+@dataclass
+class RequestMetadataFrame(ControlFrame):
```
aconchillo, line 1702: Should this be a `SystemFrame`? An upstream `ControlFrame` will probably be processed fast. But it seems we want this to happen as soon as possible because the service is already active in the ServiceSwitcher.

src/pipecat/pipeline/service_switcher.py
```diff
@@ -192,6 +235,25 @@ def _make_pipeline_definition(
...
         ]
 
+    async def _parallel_push_frame(self, frame: Frame, direction: FrameDirection):
+        """Push frames while handling StartFrame completion for metadata ordering.
+
+        When StartFrame is released (after all branches have processed it),
+        we send a filter frame to mark startup as complete and trigger metadata
+        emission. This ensures ServiceMetadataFrame arrives after StartFrame.
+        """
+        await super()._parallel_push_frame(frame, direction)
```
aconchillo: This is the reason for making small functions for almost everything, so we can reuse them.

src/pipecat/frames/frames.py
markbackman: I think it will work. Let me try.

markbackman: Yup, no issue. I'll make this change.

src/pipecat/turns/user_stop/transcription_user_turn_stop_strategy.py
```diff
@@ -23,34 +24,47 @@
 class TranscriptionUserTurnStopStrategy(BaseUserTurnStopStrategy):
     """User turn stop strategy based on transcriptions.
```
kompfner: In my mind, the *main* defining feature of this strategy—what distinguishes it from the TurnAnalyzer one—is this:

```
User turn stop strategy that uses a configurable simple timeout to determine if the user is done speaking.
```

as opposed to the TurnAnalyzer strategy, which can be described in these terms:

```
User turn stop strategy that uses a turn detection model to determine if the user is done speaking.
```

The transcript stuff is secondary and similar between the strategies. I think it's important to talk about it in terms like this because it's a simple mental model that'll be easier for developers to understand and make choices about. This is why I've also been pretty insistent on the name change: easy to compare `TimeoutUserTurnStopStrategy` v `TurnAnalyzerUserTurnStopStrategy`.

kompfner: In this docstring I'd also talk about the transcript stuff secondarily to the configurable "user resume speaking" timeout.

src/pipecat/turns/user_stop/transcription_user_turn_stop_strategy.py
```diff
@@ -23,34 +24,47 @@
...
+    The timeout starts when VADUserStoppedSpeakingFrame is received. For services
+    that support finalization (TranscriptionFrame.finalized=True), the turn can
+    be triggered immediately once the finalized transcript is received and the
+    user resume speaking timeout has elapsed.
 
     """
 
-    def __init__(self, *, timeout: float = 0.5, **kwargs):
+    def __init__(self, *, user_resume_speaking_timeout: float = 0.6, **kwargs):
         """Initialize the transcription-based user turn stop strategy.
```
kompfner: Same feedback re configurable timeout being the "headline" feature of this strategy

src/pipecat/services/stt_latency.py
aconchillo: Very nice to put this in the same place. Beautiful! 👌

src/pipecat/services/cartesia/stt.py
```diff
@@ -146,10 +148,12 @@ def __init__(
             base_url: Custom API endpoint URL. If empty, uses default.
             sample_rate: Audio sample rate in Hz. Defaults to 16000.
             live_options: Configuration options for transcription service.
+            ttfs_p99_latency: P99 latency from speech end to final transcript in seconds.
+                Defaults to 0.55s (measured). Override for custom deployments.
```
aconchillo: Remove the default values in docstrings. This is very hard to maintain.

aconchillo: You can point to the benchmark like you did for AseemblyAI.

src/pipecat/transports/base_input.py
```diff
@@ -464,7 +464,12 @@ async def _audio_task_handler(self):
                     if self._params.turn_analyzer:
                         await self._deprecated_handle_user_interruption(VADState.QUIET)
                     else:
-                        await self.push_frame(VADUserStoppedSpeakingFrame())
+                        stop_secs = (
+                            self._params.vad_analyzer.params.stop_secs
+                            if self._params.vad_analyzer
+                            else 0.0
+                        )
+                        await self.push_frame(VADUserStoppedSpeakingFrame(stop_secs=stop_secs))
```
aconchillo, line 472: Good things this is all deprecated.

src/pipecat/turns/user_stop/transcription_user_turn_stop_strategy.py
```diff
@@ -78,50 +91,113 @@ async def process_frame(self, frame: Frame):
...
+        if not self._vad_user_speaking and self._vad_stopped_time is None:
+            # Cancel existing fallback timeout if any
+            if self._timeout_task:
+                await self.task_manager.cancel_task(self._timeout_task)
+            timeout = self._calculate_timeout()
+            self._timeout_task = self.task_manager.create_task(
+                self._timeout_handler(timeout), f"{self}::_timeout_handler"
+            )
+
+    def _calculate_timeout(self) -> float:
```
kompfner: We had originally planned on a two-timer/two-independent-state approach here, where the two states were the following (ignoring these non-final names):

* `waiting_for_user_resume`
* `waiting_for_transcription`

Here it's been simplified to a single timer/single state, effectively `waiting_for_transcription` _falling back to_ `waiting_for_user_resume` if transcription has been finalized.

I'm not sure we can merge these two timers into one...they _do_ represent independent state, as far as I can tell.

In the case that our STT services doesn't set `finalized` on the frame, *that's* when we need to set (or reset) the timeout for the `waiting_for_transcription` state, it seems (which might be missing in this implementation).

So as much as I like the idea of trying to "compress" the state management here into a single timer, I do think we still need bookkeeping for the 2 independent states:

* `waiting_for_user_resume`: timer starts on VAD stop
* `waiting_for_transcription`: timer starts on transcription receipt, potentially resolves earlier if receive `finalized` transcription frame

And on each timer completion, we check if all conditions are met to say "user turn has ended".

kompfner: I could be missing something, though...this stuff is a bit tricky to reason through!

src/pipecat/turns/user_stop/transcription_user_turn_stop_strategy.py
aconchillo: I see your point, my issue with `TimeoutUserTurnStopStrategy` is that it might get confused with the default stop timeout which is triggered when non of the stop strategies work. There's a `user_turn_stop_timeout`. Looking for a better name....

aconchillo: I think `TranscriptionUserTurnStopStrategy` is actually pretty accurate. It's true that both use transcriptions, but this one is solely based on transcriptions, the other one uses the model + transcriptions. Actually, the timeout might be just an implementation details. We could just say:

```
User turn stop strategy that uses only transcriptions to determine if the user is done speaking.
```

changelog/3637.changed.2.md
```diff
@@ -0,0 +1,5 @@
+- Improved user turn stop timing in `TranscriptionUserTurnStopStrategy` and `TurnAnalyzerUserTurnStopStrategy`.
+  - Timeout now starts on `VADUserStoppedSpeakingFrame` for tighter, more predictable timing
```
kompfner, line 2: Note: this will need updating depending on outcome of my other comments...

src/pipecat/turns/user_stop/transcription_user_turn_stop_strategy.py
aconchillo: I'm not set on what I said though... finishing reading the PR 😅

aconchillo: What about `SpeechTimeoutUserTurnStopStrategy`? I think I like it!

src/pipecat/frames/frames.py
```diff
@@ -1651,6 +1668,49 @@ class SpeechControlParamsFrame(SystemFrame):
...
+        ttfs_p99_latency: Time to final segment P99 latency in seconds.
+            This is the expected time from when speech ends to when the
+            final transcript is received, at the 99th percentile.
+    """
+
+    ttfs_p99_latency: float
+
+
+@dataclass
+class RequestMetadataFrame(SystemFrame):
```
kompfner: Slick

src/pipecat/turns/user_stop/transcription_user_turn_stop_strategy.py
```diff
@@ -23,34 +24,47 @@
...
 
-    def __init__(self, *, timeout: float = 0.5, **kwargs):
+    def __init__(self, *, user_resume_speaking_timeout: float = 0.6, **kwargs):
         """Initialize the transcription-based user turn stop strategy.
 
         Args:
-            timeout: A short delay used internally to handle consecutive or
-                slightly delayed transcriptions.
+            user_resume_speaking_timeout: Time to wait for the user to potentially
+                say more after they pause speaking. Defaults to 0.6 seconds.
```
aconchillo: This is the `(old vad stops secs) 0.8 - (needed vad stop secs )0.2 = 0.6`.

src/pipecat/frames/frames.py
kompfner: Hmm...is there a case for ever wanting to use this anywhere other than ServiceSwitcher, like to trigger all services in a pipeline to re-broadcast metadata?

Because it looks like ServiceSwitcher will block this frame from traveling past it up or down the pipeline (if I'm reading it right)...

kompfner: If it *is* meant to be ServiceSwitcher-specific, then maybe it could have a ServiceSwitcher-specific name...

kompfner: (And then maybe some other day if/when we need the more "general" one of these, we can add it then)

src/pipecat/pipeline/service_switcher.py
```diff
@@ -149,19 +157,54 @@ async def filter(_: Frame) -> bool:
...
                 # Two ServiceSwitcherFilters "sandwich" a service. Push the
                 # frame only to update the other side of the sandwich, but
                 # otherwise don't let it leave the sandwich.
                 if direction == self._direction:
                     await self.push_frame(frame, direction)
+                # If this is the upstream filter for the active service, request
+                # metadata when: (1) startup just completed, or (2) service became
+                # active. Startup triggers metadata request because ParallelPipeline
+                # synchronizes StartFrame release, so we must request metadata after
+                # to maintain ordering for downstream processors.
```
kompfner: Minor clarity thing

```suggestion
                # If this is the upstream filter for the active service, request
                # a metadata broadcast from the service when: (1) startup just completed, or (2) service became
                # active. Startup triggers metadata request because ParallelPipeline
                # synchronizes StartFrame release, so we must request metadata after
                # to maintain ordering for downstream processors.
```

Also, for my understanding: we could've done this work on either the upstream or downstream filter, right? We just had to pick one arbitrarily to be the one to request the new metadata broadcast from the wrapped service? Am I understanding this right?

src/pipecat/services/stt_service.py
```diff
@@ -314,14 +324,13 @@ async def push_frame(self, frame: Frame, direction: FrameDirection = FrameDirect
...
-        """
-        if frame.vad_params is not None:
-            self._vad_stop_secs = frame.vad_params.stop_secs
+    async def _push_stt_metadata(self):
+        """Push STT metadata frame for downstream processors (e.g., turn strategies)."""
+        ttfs = self._ttfs_p99_latency
+        if ttfs is None:
+            ttfs = DEFAULT_TTFS_P99
+            logger.warning(f"{self.name}: ttfs_p99_latency not set, using default {ttfs}s")
+        await self.push_frame(STTMetadataFrame(service_name=self.name, ttfs_p99_latency=ttfs))
```
kompfner: Should this be broadcast? Theoretically any service upstream or downstream could care about receiving new metadata about pipeline services

src/pipecat/pipeline/service_switcher.py
kompfner: Ah wait maybe it *had* to be done on the upstream filter because we want the `StartFrame` to have already passed through the service before asking it emit metadata, right

src/pipecat/turns/user_stop/transcription_user_turn_stop_strategy.py
kompfner: If we take that name, then maybe we can update the init-time arg to match? Maybe `speech_timeout` or `user_speech_timeout`? (I don't feel strongly about the arg name change; I think it also works as-is)

src/pipecat/services/cartesia/stt.py
markbackman: Good catch!

src/pipecat/transports/base_input.py
markbackman: Agreed!

src/pipecat/turns/user_stop/transcription_user_turn_stop_strategy.py
markbackman: Yes, that's right.

src/pipecat/services/stt_service.py
markbackman: Yes. I had debated about it, but you're right. Let's broadcast this.

src/pipecat/frames/frames.py
markbackman: It's currently ServerSwitcher specific but I could see how it could be generalized. Not sure about where to put it... I opted for general to future proof.

src/pipecat/turns/user_stop/transcription_user_turn_stop_strategy.py
kompfner: I did miss something—I realize now that both pieces of state are now driven off of the VAD stop...the time delta between transcripts is irrelevant now.

src/pipecat/pipeline/service_switcher.py
```diff
@@ -192,6 +236,25 @@ def _make_pipeline_definition(
...
+
+        if isinstance(frame, StartFrame):
+            # StartFrame has been released. Send filter frame to mark startup
+            # complete and trigger metadata from the active service.
+            await super().process_frame(
+                ServiceSwitcher.ServiceSwitcherFilterFrame(
+                    active_service=self.strategy.active_service
+                ),
+                FrameDirection.DOWNSTREAM,
+            )
```
aconchillo: This should probably be:

```
await self.queue_frame(ServiceSwitcherFrame(active_service=self.strategy.active_service))
```

Which will endup in `process_frame()` below. So we just have one place where we send the filter frame.

aconchillo: Or also

```
await self.process_frame(
    ServiceSwitcherFrame(active_service=self.strategy.active_service),
    FrameDirection.DOWNSTREAM
)
```

src/pipecat/turns/user_stop/transcription_user_turn_stop_strategy.py
kompfner: So yes, this stuff can be "collapsed" into a single timer.

src/pipecat/frames/frames.py
kompfner: Per our discussion:
- make it specific, renaming to something like `ServiceSwitcherRequestMetadataFrame`
- no need to swallow the frame from the STT services; the ServiceSwitcher itself should be responsible for swallowing, since it's a ServiceSwitcher-specific mechanism

src/pipecat/pipeline/service_switcher.py
markbackman: I removed the need to override the function :)

changelog/3637.changed.2.md
```diff
@@ -0,0 +1,5 @@
+- Improved user turn stop timing in `TranscriptionUserTurnStopStrategy` and `TurnAnalyzerUserTurnStopStrategy`.
```
aconchillo, line 1: Should we mention `SpeechTimeoutUserTurnStopStrategy`?

```
Improved user turn stop timing in SpeechTimeoutUserTurnStopStrategy` (previously `TranscriptionUserTurnStopStrategy`)...
```

src/pipecat/pipeline/service_switcher.py
aconchillo: Sorry to go back to this. I know I'm missing something, but I'm not sure I understand why we need all the logic above. Couldn't we process a `RequestMetadataFrame` from here? If we do that, it feels the upstream/downstream filter would just do it's thing?

```
            await super().process_frame(service_switcher_filter_frame, direction)
            await super().process_frame(RequestMetadataFrame(), direction)
```

src/pipecat/pipeline/service_switcher.py
```diff
@@ -149,19 +164,54 @@ async def filter(_: Frame) -> bool:
...
+            if isinstance(frame, RequestMetadataFrame):
+                if direction == self._direction and self._wrapped_service == self._active_service:
+                    await self.push_frame(frame, direction)
+                return
+
+            # Block ServiceMetadataFrame from inactive services.
+            if isinstance(frame, ServiceMetadataFrame):
+                if self._wrapped_service != self._active_service:
+                    return
+                await self.push_frame(frame, direction)
```
aconchillo: And could we just add this to the filter function instead?

src/pipecat/pipeline/service_switcher.py
markbackman: The issue is a race condition. We have to wait until the `ServiceSwitcherFilterFrame` reaches the lower FunctionFilter in the ServiceSwitcher pipeline before we can emit the RequestMetadataFrame. Without this, the STTMetadataFrame will be blocked, as the lower FunctionFilter doesn't know about the active processor switch yet. At least, that's what I can tell in my testing. (I tested a very similar thing to this.)

src/pipecat/pipeline/service_switcher.py
```diff
@@ -123,7 +125,18 @@ def __init__(self, services: List[FrameProcessor], strategy_type: Type[StrategyT
...
     class ServiceSwitcherFilter(FunctionFilter):
-        """An internal filter that allows frames to pass through to the wrapped service only if it's the active service."""
+        """An internal filter that gates frame flow based on active service.
+
+        Two filters "sandwich" each service, allowing frames through only
+        when the wrapped service is active. The pipeline layout is::
+
+            DownstreamFilter → Service → UpstreamFilter
+
+        The filter names refer to which *direction* of frame flow they
```
kompfner: Thank you, this is a really handy reference comment. I've forgotten several time which one "upstream" refers to hehe.

src/pipecat/frames/frames.py
kompfner: @markbackman were you still hoping to do the suggestions in the above comment?

src/pipecat/pipeline/service_switcher.py
```diff
@@ -149,19 +164,54 @@ async def filter(_: Frame) -> bool:
...
                 self._active_service = frame.active_service
-                # Two ServiceSwitcherFilters "sandwich" a service. Push the
-                # frame only to update the other side of the sandwich, but
-                # otherwise don't let it leave the sandwich.
+                # Two ServiceSwitcherFilters "sandwich" a service. The
+                # frame enters via the downstream filter first. Push it
+                # through so the upstream filter also updates its state.
                 if direction == self._direction:
                     await self.push_frame(frame, direction)
+                # This is the upstream filter (the second to update). At
```
kompfner: Hmm...I actually don't think the assumption that the upstream filter is 2nd to update holds. In Pipecat Flows we even have an example pushing a `ManuallySwitchServiceFrame` _upstream_.

markbackman: We discussed this in our sync. The ServiceSwitcher refactor handle this case. All unit tests are passing.

danish-wani-haptik:
@markbackman 
Thank you so much for this.
We are currently on the latest version and are facing this issue where it takes 5 seconds, can you please guide us on how to use this change?

stt_provider being used is SONIOX

This is our current setup

```
        stop_strategies = [
                turn_analyzer=LocalSmartTurnAnalyzerV3(
                    params=SmartTurnParams(stop_secs=SMART_TURN_STOP_SECS)
                ),
            )
        ]
        # user_turn_stop_timeout should account for SMART_TURN_STOP_SECS + STT latency buffer
        user_turn_stop_timeout = DEFAULT_USER_TURN_STOP_TIMEOUT_SECS # set to 5 seconds

    params = LLMUserAggregatorParams(
            start=[
                MinWordsUserTurnStartStrategy(min_words=USER_TURN_START_MIN_WORDS),
            ],
            stop=stop_strategies,
        ),
        user_turn_stop_timeout=user_turn_stop_timeout,
        vad_analyzer=SileroVADAnalyzer(params=VAD_USER_TURN_PARAMS),
        user_mute_strategies=(
            [
                MuteUntilFirstBotCompleteUserMuteStrategy(),
            ]
            if is_callgenie
            else []
        ),
    )
    ```

Any help would be highly appreciated.

**Comment by markbackman** [comment]
@danish-wani-haptik would you mind opening an troubleshooting issue for this, so we can track it?

**Comment by danish-wani-haptik** [comment]
> @danish-wani-haptik would you mind opening an troubleshooting issue for this, so we can track it?

Sure, let me do that.

Done, issue is raised here: https://github.com/pipecat-ai/pipecat/issues/4250

File: src/pipecat/turns/user_stop/__init__.py [hunk 29]
```diff
@@ -6,13 +6,13 @@
...
-from .transcription_user_turn_stop_strategy import TranscriptionUserTurnStopStrategy
+from .speech_timeout_user_turn_stop_strategy import SpeechTimeoutUserTurnStopStrategy
 from .turn_analyzer_user_turn_stop_strategy import TurnAnalyzerUserTurnStopStrategy
 
 __all__ = [
     "BaseUserTurnStopStrategy",
     "ExternalUserTurnStopStrategy",
+    "SpeechTimeoutUserTurnStopStrategy",
     "UserTurnStoppedParams",
-    "TranscriptionUserTurnStopStrategy",
```
010gvr, line 16: understand this was deprecated but I think this removal was a bit pre-mature as it made into even patch releases of [0.0.107](https://github.com/pipecat-ai/pipecat/commits/v0.0.107/src/pipecat/turns/user_stop/__init__.py) and hence breaking builds. The [0.0.107 changelog](https://github.com/pipecat-ai/pipecat/releases/tag/v0.0.107) doesn't list this breaking change either. Looks like the removal was planned only in [1.0.0](https://github.com/pipecat-ai/pipecat/blob/main/CHANGELOG.md#100---2026-04-14)?

markbackman: Apologies about that. That breaking change wasn't intended. You can still use it in 0.0.107, but you need to use the full module path:

```
pipecat.turns.user_stop.transcription_user_turn_stop_strategy
```

Issue #3639, opened 2026-02-04
LLM services missing stream cleanup on cancellation 


lukepayyapilli:
## Summary

Multiple LLM services iterate over async streams without ensuring cleanup on cancellation/exception. This can lead to socket leaks when `CancelledError` is raised during stream iteration.

## Affected Services

| Service | File | Pattern |
|---------|------|---------|
| BaseOpenAILLMService | `openai/base_llm.py` | `async for chunk in chunk_stream` |
| SambaNovaLLMService | `sambanova/llm.py` | Same pattern (overrides `_process_context`) |
| GoogleLLMOpenAIBetaService | `google/llm_openai.py` | Same pattern (overrides `_process_context`) |
| AnthropicLLMService | `anthropic/llm.py` | `async for event in response` |

## Root Cause

When iterating over an async stream without a context manager, the stream's `close()` method is not called if:
- `CancelledError` is raised (e.g., user interruption)
- Any other exception occurs during iteration

The OpenAI SDK's `AsyncStream` is designed to be used with `async with` to ensure cleanup via `__aexit__`.

## Current State

- `SambaNovaLLMService` and `GoogleLLMOpenAIBetaService` copy-paste the entire `_process_context` method from `BaseOpenAILLMService` rather than calling the base implementation
- Google has a TODO: "Remove the _process_context method once Google resolves the bug"
- No shared utility for stream cleanup

### Option 1: Fix each service individually
Add `async with chunk_stream:` wrapper in each `_process_context` method.

**Pros:** Simple, minimal changes per file
**Cons:** Repetitive, doesn't address code duplication

### Option 2: Create a shared utility in LLMService
```python
@asynccontextmanager
async def _managed_stream(self, stream):
    try:
        yield stream
    finally:
        if hasattr(stream, 'close'):
            await stream.close()
```

**Pros:** DRY, standardized pattern
**Cons:** Still need to update callers

### Option 3: Refactor to eliminate duplication
Address why `SambaNovaLLMService` and `GoogleLLMOpenAIBetaService` copy-paste `_process_context` instead of extending it.

**Pros:** Fixes underlying technical debt
**Cons:** Larger refactor

## Recommendation

1. Fix `BaseOpenAILLMService` first (#3640 addresses #3589)
2. Apply same fix to `SambaNovaLLMService` and `GoogleLLMOpenAIBetaService`
3. Investigate `AnthropicLLMService` (different SDK, may have different cleanup semantics)
[...]

PR #3640, opened 2026-02-04
fix: close stream on cancellation to prevent socket leaks

### LINKED ISSUE #3589: Missing Close() call for  connection used by  LLMService of type openai
Hi @aconchillo  i have been looking at how openai client was used in the service. 

When you iterate on the chunk, the context manager is not used. 
However only in the aexit() function of the context manager the function close() is called which enables to close the socket in any case (especially when there is an exception, like a CancelledError). 
As no context manager is used, when cancelledError is triggered, i believe that the socket will not be properly closed.

```
async for chunk in chunk_stream: <-------- HERE, no context manager used
            if chunk.usage:
                cached_tokens = (
                    chunk.usage.prompt_tokens_details.cached_tokens
                    if chunk.usage.prompt_tokens_details
                    else None
                )
                reasoning_tokens = (
                    chunk.usage.completion_tokens_details.reasoning_tokens
                    if chunk.usage.completion_tokens_details
                    else None
                )
                tokens = LLMTokenUsage(
                    prompt_tokens=chunk.usage.prompt_tokens,
                    completion_tokens=chunk.usage.completion_tokens,
                    total_tokens=chunk.usage.total_tokens,
                    cache_read_input_tokens=cached_tokens,
                    reasoning_tokens=reasoning_tokens,
                )
                await self.start_llm_usage_metrics(tokens)

```
https://github.com/pipecat-ai/pipecat/blob/f3b72e9263fafebabc00b4f6d4ace8fa7e3e668c/src/pipecat/services/openai/base_llm.py#L

```
class AsyncStream(Generic[_T]):
    . ...
    async def __aenter__(self) -> Self:
        return self

    async def __aexit__(
        self,
        exc_type: type[BaseException] | None,
        exc: BaseException | None,
        exc_tb: TracebackType | None,
    ) -> None:
        await self.close().  <------ ONLY CALLED if used in contet manager is being used
```

Possible fix:
```
async with chunk_stream:  # <-- ADD THIS
      async for chunk in chunk_stream:
          # ... rest of the code 
```

Am i missing something ?

lukepayyapilli:
## Summary

Fixes #3589 - OpenAI LLM stream not being closed on cancellation/exception.

When iterating over the `AsyncStream` without a context manager, the stream's `close()` method is not called if a `CancelledError` or other exception occurs. This leaves the underlying socket open.

## Changes

- [x] Wrap `chunk_stream` iteration in `async with chunk_stream:` context manager.
- [x] Add test verifying stream closes on `CancelledError`.
- [x] Add changelog entry.

## Approach

The OpenAI SDK's `AsyncStream` is designed to be used as a context manager. Its `__aexit__` calls `close()` which releases the connection. Using `async with` ensures cleanup happens even on cancellation.

```python
# Before
async for chunk in chunk_stream:
    ...

# After  
async with chunk_stream:
    async for chunk in chunk_stream:
        ...
```

## Related

Issue #3639 documents that `SambaNovaLLMService`, `GoogleLLMOpenAIBetaService`, and `AnthropicLLMService` have the same pattern and may need similar fixes.

### DIFF
--- a/changelog/3589.fixed.md
+++ b/changelog/3589.fixed.md
@@ -1,0 +1,1 @@
+- Fixed OpenAI LLM stream not being closed on cancellation/exception, which could leak sockets.
--- a/src/pipecat/services/openai/base_llm.py
+++ b/src/pipecat/services/openai/base_llm.py
@@ -362,74 +362,77 @@ class BaseOpenAILLMService(LLMService):
             else self._stream_chat_completions_universal_context(context)
         )
 
-        async for chunk in chunk_stream:
-            if chunk.usage:
-                cached_tokens = (
-                    chunk.usage.prompt_tokens_details.cached_tokens
-                    if chunk.usage.prompt_tokens_details
-                    else None
-                )
-                reasoning_tokens = (
-                    chunk.usage.completion_tokens_details.reasoning_tokens
-                    if chunk.usage.completion_tokens_details
-                    else None
-                )
-                tokens = LLMTokenUsage(
-                    prompt_tokens=chunk.usage.prompt_tokens,
-                    completion_tokens=chunk.usage.completion_tokens,
-                    total_tokens=chunk.usage.total_tokens,
-                    cache_read_input_tokens=cached_tokens,
-                    reasoning_tokens=reasoning_tokens,
-                )
-                await self.start_llm_usage_metrics(tokens)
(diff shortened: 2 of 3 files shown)


markbackman (approved):
LGTM! Thanks for the submission! 🙌

PR #3642, opened 2026-02-04
Update RimeTTSService for arcana and mistv2 model support

chadbailey59:
## Summary

- Updated `RimeTTSService` default endpoint to `wss://users-ws.rime.ai/ws3` and default model to `arcana`, following [Rime Arcana v3 JSON WebSocket support](https://docs.rime.ai/docs/changelog#2026-02-04-2)
- Overhauled `InputParams` with model-specific parameters: arcana (`repetition_penalty`, `temperature`, `top_p`) and mistv2 (`no_text_normalization`, `save_oovs`, `segment`), all defaulting to `None` so only explicitly-set values are sent as query params
- Added `_build_settings()` to dynamically construct WebSocket query params based on the current model
- `set_model()` and `_update_settings()` now trigger WebSocket reconnection when settings change, since all params are URL query parameters
- Deprecated `RimeNonJsonTTSService` now that arcana supports the JSON WebSocket endpoint with word-level timestamps

## Breaking Changes

- `RimeTTSService` default model changed from `mistv2` to `arcana`
- Default WebSocket URL changed from `wss://users.rime.ai/ws2` to `wss://users-ws.rime.ai/ws3`
- `InputParams` defaults changed from mistv2-specific values (`speed_alpha=1.0`, `reduce_latency=False`, etc.) to `None` — only explicitly-set params are now sent as query params

## Testing

- Run `uv run python examples/foundational/07q-interruptible-rime.py` with arcana model
- Verify mistv2 still works by passing `model="mistv2"` with mistv2-specific params

### DIFF
--- a/src/pipecat/services/rime/tts.py
+++ b/src/pipecat/services/rime/tts.py
@@ -98,8 +98,8 @@ class RimeTTSService(AudioContextWordTTSService):
         *,
         api_key: str,
         voice_id: str,
-        url: str = "wss://users.rime.ai/ws2",
-        model: str = "mistv2",
+        url: str = "wss://users-ws.rime.ai/ws3",
+        model: str = "arcana",
         sample_rate: Optional[int] = None,
         params: Optional[InputParams] = None,
         text_aggregator: Optional[BaseTextAggregator] = None,
@@ -383,7 +383,7 @@ class RimeTTSService(AudioContextWordTTSService):
         async for message in self._get_websocket():
             msg = json.loads(message)
 
-            if not msg or not self.audio_context_available(msg["contextId"]):
+            if not msg or not self.audio_context_available(msg.get("contextId")):
                 continue
 
             context_id = msg["contextId"]
@@ -626,20 +626,18 @@ class RimeHttpTTSService(TTSService):
 class RimeNonJsonTTSService(InterruptibleTTSService):
     """Pipecat TTS service for Rime's non-JSON WebSocket API.
 
+    .. deprecated:: 0.0.102
+        Arcana now supports JSON WebSocket with word-level timestamps via the
+        ``wss://users-ws.rime.ai/ws3`` endpoint. Use :class:`RimeTTSService`
+        with ``model="arcana"`` instead.
+
     This service enables Text-to-Speech synthesis over WebSocket endpoints
     that require plain text (not JSON) messages and return raw audio bytes.
-    It is designed for use with TTS models like Arcana, which currently do
-    not support JSON-based WebSocket protocols (though this may change in
-    the future).
 
     Limitations:
         - Does not support word-level timestamps or context IDs.
         - Intended specifically for integrations where the TTS provider only
           accepts and returns non-JSON messages.
-
-    Note:
-        - Arcana and similar models may add JSON WebSocket support in the
-          future. This service focuses on the current plain text protocol.
     """
 
     class InputParams(BaseModel):
--- a/src/pipecat/services/rime/tts.py
+++ b/src/pipecat/services/rime/tts.py
@@ -81,17 +81,31 @@ class RimeTTSService(AudioContextWordTTSService):
 
         Parameters:
             language: Language for synthesis. Defaults to English.
-            speed_alpha: Speech speed multiplier. Defaults to 1.0.
-            reduce_latency: Whether to reduce latency at potential quality cost.
-            pause_between_brackets: Whether to add pauses between bracketed content.
-            phonemize_between_brackets: Whether to phonemize bracketed content.
+            segment: Text segmentation mode ("immediate", "bySentence", "never").
+            repetition_penalty: Token repetition penalty (arcana only).
+            temperature: Sampling temperature (arcana only).
+            top_p: Cumulative probability threshold (arcana only).
+            speed_alpha: Speech speed multiplier (mistv2 only).
+            reduce_latency: Whether to reduce latency at potential quality cost (mistv2 only).
+            pause_between_brackets: Whether to add pauses between bracketed content (mistv2 only).
+            phonemize_between_brackets: Whether to phonemize bracketed content (mistv2 only).
+            no_text_normalization: Whether to disable text normalization (mistv2 only).
+            save_oovs: Whether to save out-of-vocabulary words (mistv2 only).
         """
 
         language: Optional[Language] = Language.EN
-        speed_alpha: Optional[float] = 1.0
-        reduce_latency: Optional[bool] = False
-        pause_between_brackets: Optional[bool] = False
-        phonemize_between_brackets: Optional[bool] = False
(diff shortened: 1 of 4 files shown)


MaCaki (approved):
thank you!

markbackman (changes requested):
The `ws3` endpoint has a few breaking changes that need to be adapted. There's an open question in our Slack to decide what to do before this can be merged.

The audio works great. The issue pertains to timestamp messages being different between `ws2` and `ws3`.

0xjyk:
@chadbailey59 Just making the changes above doesn't seem to work for me. I might be mistaken, but some more changes are required to get this up and running - 
1. Mistv2 and arcana accept different variable parameters (InputParams in the Pipecat service)
2. `contextId` isn't passed in the timestamp responses. As a result the spoken words aren't added to the conversation log.

@markbackman Given this, what's the quickest way to integrate Arcana V3? Happy to implement it and raise a PR.

markbackman:
> @chadbailey59 Just making the changes above doesn't seem to work for me. I might be mistaken, but some more changes are required to get this up and running -
> 
> 1. Mistv2 and arcana accept different variable parameters (InputParams in the Pipecat service)
> 2. `contextId` isn't passed in the timestamp responses. As a result the spoken words aren't added to the conversation log.
> 
> @markbackman Given this, what's the quickest way to integrate Arcana V3? Happy to implement it and raise a PR.

@0xjyk In testing this PR, I reported three issues to the Rime team, which they're working on:

- Timestamp unit consistency between the `ws2` and `ws3` endpoints
- Timestamp missing `contextId` in `ws3`
- Missing punctuation in the `ws3` endpoint

Once these are resolved, I'll revisit this PR.

We'll also add support for the new `InputParams`.

0xjyk:
Thanks a lot for clarifying! Also - I've really enjoyed going through the codebase. The interfaces are super intuitive and very well documented. Really appreciate the work you guys have put into this project. Lots of love <3

markbackman:
> Thanks a lot for clarifying! Also - I've really enjoyed going through the codebase. The interfaces are super intuitive and very well documented. Really appreciate the work you guys have put into this project. Lots of love <3

Thanks for the kind words! It means a lot :)

0xjyk:
Spoke to Rime's team recently. They claim to have fixed the above mentioned issues.

markbackman:
> Spoke to Rime's team recently. They claim to have fixed the above mentioned issues.

Cool! I'll take a look.

markbackman:
Ok, the three issues I outlined are fixed, but the word level timestamps have incorrect timing—they're slower than the audio play out. I've alerted them of this issue. I think we can proceed with this change as the API now works correctly.

I'm in touch with the Rime team who are working on timestamp improvements.

markbackman:
Ok! This PR has been updated.

Maybe @filipi87 can take a look when he gets a chance.

- Arcana Websocket JSON docs: https://docs.rime.ai/api-reference/arcana/websockets-json
- Mist v2 Websocket JSON docs: https://docs.rime.ai/api-reference/endpoint/websockets-json

markbackman (approved):
Approving to dismiss the request changes.

Looking for a separate review since I made the changes (and don't want to approve my own PR).

0xjyk:
Thanks a lot @markbackman and @chadbailey59!

src/pipecat/services/rime/tts.py
```diff
@@ -81,25 +81,39 @@ class InputParams(BaseModel):
...
+        repetition_penalty: Optional[float] = None
+        temperature: Optional[float] = None
+        top_p: Optional[float] = None
+        # Mistv2 params
+        speed_alpha: Optional[float] = None
+        reduce_latency: Optional[bool] = None
+        pause_between_brackets: Optional[bool] = None
+        phonemize_between_brackets: Optional[bool] = None
+        no_text_normalization: Optional[bool] = None
+        save_oovs: Optional[bool] = None
```
filipi87, line 108: Hi @markbackman, since Arcana and MistV2 support completely different parameters, would it make sense to structure the input params as model specific classes, for example:

```
class InputParams(BaseModel):
    language: Optional[Language] = Language.EN

class ArcanaInputParams(InputParams):
    repetition_penalty: float
    temperature: float
    top_p: float

class MistV2InputParams(InputParams):
    peed_alpha: Optional[float] = None
    reduce_latency: Optional[bool] = None
    pause_between_brackets: Optional[bool] = None
    phonemize_between_brackets: Optional[bool] = None
    no_text_normalization: Optional[bool] = None
    save_oovs: Optional[bool] = None
```

It feels like this would make it easier for developers to know exactly which parameters are supported.

src/pipecat/services/rime/tts.py
```diff
@@ -223,12 +271,68 @@ def INLINE_SPEED(self, text: str, speed: float) -> str:
...
+            needs_reconnect = True
+
+        if "language" in settings:
+            new_lang = self.language_to_service_language(settings["language"])
+            if new_lang and new_lang != prev_settings.get("lang"):
+                logger.info(f"Updating language to: [{new_lang}]")
+                self._settings["lang"] = new_lang
+                needs_reconnect = True
+
+        # Arcana params
```
filipi87, line 301: This could also be improved in the future once the ongoing `update_settings` work is in place. But for now the way it is makes sense to me.

src/pipecat/services/rime/tts.py
```diff
@@ -626,20 +730,18 @@ async def run_tts(self, text: str, context_id: str) -> AsyncGenerator[Frame, Non
 class RimeNonJsonTTSService(InterruptibleTTSService):
     """Pipecat TTS service for Rime's non-JSON WebSocket API.
 
+    .. deprecated:: 0.0.102
```
filipi87, line 733: 🔥

src/pipecat/services/rime/tts.py
markbackman: I had the same thought. I'm just going to leave it as is for now (though it's very messy!)

src/pipecat/services/rime/tts.py
markbackman: I'm torn on whether this makes sense. I do like that it documents that feature compatibility, but it breaks from the other patterns we have in place.

I'm curious where @kompfner's settings refactor will land as it may change things. Perhaps we should start with a flat structure and move towards using different BaseModels for the different TTS models. WDYT?

filipi87: Yep, we can land it as it is and improve it later once we’re aware of all the refactors Paul is doing. 👍

filipi87 (approved):
LGTM! 🚀

src/pipecat/services/rime/tts.py
filipi87: Makes sense. 👍

PR #3645, opened 2026-02-05
fix(smallwebrtc): respect audio_out_10ms_chunks parameter in RawAudioTrack

jamsea:
## Summary

Fixed an issue where `SmallWebRTCTransport` was not respecting the `audio_out_10ms_chunks` parameter from `TransportParams`. The `RawAudioTrack` class was hardcoded to always produce 10ms audio frames, causing firmware clients to receive incorrect chunk sizes (e.g., receiving 20ms chunks when 40ms was configured).

## Problem

When using `SmallWebRTCTransport` on Pipecat Cloud with `audio_out_10ms_chunks=4` (40ms), firmware clients were still receiving 20ms audio chunks. This happened because:

1. The base transport correctly chunked audio into 40ms pieces based on `audio_out_10ms_chunks`
2. However, `RawAudioTrack` in SmallWebRTC was re-chunking the audio back to 10ms for WebRTC transmission
3. The `recv()` method always produced 10ms frames regardless of configuration

## Solution

Updated `RawAudioTrack` to accept and respect a configurable chunk size:

- Added `num_10ms_chunks` parameter to `RawAudioTrack` constructor (default 1 for backward compatibility)
- Updated `add_audio_bytes()` to validate and chunk audio using the configured chunk size
- Updated `recv()` to produce audio frames of the configured size and advance timestamps correctly
- Modified `_handle_client_connected()` to pass `audio_out_10ms_chunks` from `TransportParams` when creating the track

## Testing

- Verified that setting `audio_out_10ms_chunks=4` now correctly produces 40ms audio frames over WebRTC
- Backward compatible: default behavior (10ms chunks) remains unchanged when parameter is not specified

### DIFF
--- a/changelog/3645.fixed.md
+++ b/changelog/3645.fixed.md
@@ -1,0 +1,1 @@
+- Fixed `SmallWebRTCTransport` not respecting `TransportParams.audio_out_10ms_chunks` parameter. The transport was hardcoded to produce 10ms audio frames regardless of the configured chunk size.
--- a/src/pipecat/transports/smallwebrtc/transport.py
+++ b/src/pipecat/transports/smallwebrtc/transport.py
@@ -84,7 +84,12 @@ class RawAudioTrack(AudioStreamTrack):
         Args:
             sample_rate: The audio sample rate in Hz.
             num_10ms_chunks: Number of 10ms chunks per output frame (default 1).
+
+        Raises:
+            ValueError: If num_10ms_chunks is not a positive integer.
         """
+        if num_10ms_chunks < 1:
+            raise ValueError(f"num_10ms_chunks must be a positive integer, got {num_10ms_chunks}")
         super().__init__()
         self._sample_rate = sample_rate
         self._num_10ms_chunks = num_10ms_chunks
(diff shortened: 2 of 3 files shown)


markbackman:
Requesting that @filipi87 take a look at this one when he has a moment.

jamsea:
@filipi87 I went through line by line and this all _looks_ right to me, however my SmallWebRTC Vibe isn't strong enough to 100% confirm 🙏

filipi87 (commented):
The `audio_out_10ms_chunks` is used inside `BaseOutputTransport` to chunk the audio we receive from the TTS, store it in a queue, and optionally mix it with other audio, and then send it to the appropriate transport (Daily, SmallWebRTC, WebSocket ....).

As far as I remember, creating 40 ms chunks improved CPU usage inside Pipecat. But I am not sure that when we implemented this, we intended for transports to preserve that chunk size when actually sending audio to WebRTC. I don’t think that was the original intention.

For example, in Daily we always send 10 ms of audio, regardless of what is passed to DailyTransport.

And the reason is, WebRTC works best when sending 10–20 ms audio frames to keep latency low:
- 10 ms: ultra low latency
- 20 ms: low latency

This is why we implemented the same logic in SmallWebRTC.

My final concern is that if we send 40 ms frames to WebRTC, this will impact latency. So, If we want to support that behavior, we should introduce a new property to explicitly control it.

cc: @aconchillo for thoughts on this, who implemented similar logic inside Daily.

PR #3647, opened 2026-02-05
Fix function call cancel before in-progress causing stuck state

OlliTapio:
## Summary

- Fixes a bug where function calls cancelled before `FunctionCallInProgressFrame` was sent would get stuck in tracking forever
- When interrupted early, `has_function_calls_in_progress` would return `True` indefinitely, potentially blocking subsequent LLM runs
- Makes `FunctionCallResultFrame` and `FunctionCallInProgressFrame` system frames again (they were originally SystemFrames but were changed to DataFrame/ControlFrame + UninterruptibleFrame when the UninterruptibleFrame mixin was introduced)

### 1. Cancel before in-progress (aggregator bug)

When `FunctionCallsStartedFrame` is received, the aggregator adds an entry to `_function_calls_in_progress` with value `None`. Later, `FunctionCallInProgressFrame` updates this to the actual frame.

If an interruption cancels the function call **before** `FunctionCallInProgressFrame` arrives:
1. `_handle_function_call_cancel` receives `FunctionCallCancelFrame`
2. It checks `if function_call and function_call.cancel_on_interruption`
3. Since `function_call` is `None`, this condition is `False`
4. The entry is never deleted from `_function_calls_in_progress`
5. `has_function_calls_in_progress` returns `True` forever

### 2. Cancel jumping ahead of result (frame priority bug)

`FunctionCallCancelFrame` was a `SystemFrame` (high priority) while `FunctionCallResultFrame` was a `DataFrame` (low priority). In the frame processor's priority queue, the cancel would be dequeued before the result even if the result was enqueued first, causing a completed function call result to be dropped.

### Aggregator fixes
1. **Cancel before in-progress**: When `function_call` is `None`, discard the function call to prevent stuck state
2. **Late in-progress after cancel**: Check if entry exists before processing `FunctionCallInProgressFrame`

### Frame type fix
Make all function call lifecycle frames `SystemFrame` so they maintain FIFO ordering:
- `FunctionCallsStartedFrame` — already SystemFrame
- `FunctionCallInProgressFrame` — changed from `ControlFrame, UninterruptibleFrame` to `SystemFrame`
- `FunctionCallResultFrame` — changed from `DataFrame, UninterruptibleFrame` to `SystemFrame`
- `FunctionCallCancelFrame` — already SystemFrame

These frames were originally SystemFrames but were changed when `UninterruptibleFrame` was introduced (Dec 2025). However, `UninterruptibleFrame` only prevents frames from being dropped during queue clearing — it doesn't prevent reordering by SystemFrames in the priority queue.

## Test plan

- [x] `test_function_call_cancel_before_in_progress` - verifies cancel before in-progress is handled
- [x] `test_function_call_cancel_after_in_progress` - verifies normal cancellation still works
- [x] `test_function_call_not_cancelled_when_cancel_on_interruption_false` - verifies functions with `cancel_on_interruption=False` continue despite cancel frame
- [x] `test_function_call_cancel_then_in_progress_frame_arrives` - verifies late in-progress frame doesn't re-add cancelled function
- [x] All existing aggregator, mute strategy, idle, and pipeline tests pass

🤖 Generated with [Claude Code](https://claude.ai/claude-code)


OlliTapio:
This bug seemed to happen a lot when OpenAI sent content and tool_call in the same request. So when someone talks on top of the function being executed in flows. Even if the issue does not stop transferring it keeps the tool call as in progress in the LLM context which makes the LLM less likely to call the function again. 

This seems to fix the issue at least on local testing, but skipping the function interruption rules if we manage to cancel it before it is progress is maybe not wanted behaviour.

filipi87 (commented):
The reason `FunctionCallInProgressFrame` and `FunctionCallResultFrame` are `UninterruptibleFrames` rather than `SystemFrames` is to ensure that function call results respect the previous LLM context, mainly in cases where the LLM is instructed to both generate text and invoke a function call at the same time.

In those cases, the context would not be updated before the function call result arrived, causing the bot to repeat itself. All the details are described in this [issue](https://github.com/pipecat-ai/pipecat/issues/3023).

I believe we should instead provide a solution that keeps `FunctionCallInProgressFrame` and `FunctionCallResultFrame` as non `SystemFrames`.

cc: @aconchillo for thoughts.

OlliTapio:
Closed this as there is better approach as filipi mentioned

namanbansal013:
Hi @filipi87 — commenting here because we have hit the same race in production and wanted to ask how you are thinking about landing a fix.

**Our case (pipecat 1.3.0, voice bot with `cancel_on_interruption=True` tool calls):** A user interrupted during a bare tool-only LLM turn (no spoken text in that response). In the logs, `FunctionCallCancelFrame` reached `LLMAssistantAggregator` **1ms before** `FunctionCallInProgressFrame`:

- `FunctionCallCancelFrame` processed
- `FunctionCallInProgressFrame` processed 1ms later

An earlier tool call in the same session (cancel ~300ms after InProgress) correctly ended up as `CANCELLED` in context. The fast-cancelled one ended up as `IN_PROGRESS` because `_handle_function_call_cancel` no-ops when the map entry is still `None`, and the late InProgress frame then wrote `IN_PROGRESS` to context.

We traced this through 1.4 → 1.7 and `main` (as of today): the aggregator handlers in `llm_response_universal.py` still use `if function_call and function_call.cancel_on_interruption`, and `_handle_function_call_in_progress` has no guard for a late arrival after cancel. We also do not see the aggregator-only logic from the later revisions of this PR (ignore late InProgress when the entry was already removed; delete the `None` placeholder on cancel).

We saw related fixes land ([#4435](https://github.com/pipecat-ai/pipecat/pull/4435) TTS uninterruptible preservation, [#4824](https://github.com/pipecat-ai/pipecat/pull/4824) user-mute orphan frames, etc.), but not this specific cancel-before-InProgress path — which matches your earlier feedback here about keeping `FunctionCallInProgressFrame` / `FunctionCallResultFrame` as non-`SystemFrame`s ([#3023](https://github.com/pipecat-ai/pipecat/issues/3023)).

**Question:** Is the aggregator-only approach from the later commits on this PR still the direction you prefer? If so, would you be open to reviving that patch (without the `SystemFrame` type changes), or is there a different fix in flight we should wait for?

Happy to share full pipecat logs / a minimal repro if useful. Thanks for the earlier review on the frame-type tradeoff — it helped us understand why the SystemFrame approach was rejected.

Issue #3634, opened 2026-02-03
enable_rtvi=True default in 0.0.101 triggers Twilio warnings (31951 Stream - Protocol - Invalid message)


haayhappen:
### pipecat version

0.0.101

### Python version

3.13.5

### Operating System

macOS 15.7.3

### Issue description

In pipecat-ai 0.0.101, enable_rtvi=True became the default for PipelineTask. This causes Twilio warning 31951: Stream - Protocol - Invalid Message because RTVIObserver automatically sends RTVI-formatted messages through the WebSocket that Twilio doesn't recognize.

### Reproduction steps

1. Use pipecat-ai 0.0.101 with FastAPIWebsocketTransport and TwilioFrameSerializer
2. Create a PipelineTask without explicitly setting enable_rtvi=False
3. Make a call - Twilio logs show error 31951

### Expected behavior

- Rtvi messages should be validated for certain transports
- Add documentation warning about RTVI + Twilio incompatibility

### Actual behavior

Twilio logs show warning 31951 https://www.twilio.com/docs/api/errors/31951

### Logs

```shell

```

markbackman:
This will be addressed in the next release: https://github.com/pipecat-ai/pipecat/pull/3650

filipi87:
Fixed in the PR mentioned above.

PR #3650, opened 2026-02-05
Ignoring RTVI messages inside the Serializers by default.

filipi87:
Ignoring RTVI messages inside TwilioSerializer by default.

Fixes: https://github.com/pipecat-ai/pipecat/pull/3650

### DIFF
--- a/src/pipecat/serializers/base_serializer.py
+++ b/src/pipecat/serializers/base_serializer.py
@@ -7,8 +7,17 @@
 """Frame serialization interfaces for Pipecat."""
 
 from abc import ABC, abstractmethod
+from typing import Optional
 
-from pipecat.frames.frames import Frame, StartFrame
+from pydantic import BaseModel
+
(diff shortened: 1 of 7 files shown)


filipi87:
This PR fixes https://github.com/pipecat-ai/pipecat/issues/3634

markbackman:
@filipi87 do you know why RTVI messages were a factor here?

Also, should we just move the logic to ignore RTVI messages into the base FrameSerializer class so that all FrameSerializers benefit? I would imagine others are impacted too. (Or is this a Twilio only issue for some reason?)

filipi87:
@markbackman, when we enabled RTVI by default, we started sending these RTVI messages, and as mentioned in the issue, Twilio doesn’t recognize them, so we receive a warning from Twilio.

And yeah, I will apply this for all the serializers, I think it makes sense. 👍

markbackman (approved):
Much better! ☺️ 

LGTM

PR #3653, opened 2026-02-05
HeyGen improvements.

filipi87:
## Summary                                                                                                         
- Added support for sandbox mode in HeyGen LiveAvatar integration via `is_sandbox` parameter
- Added configurable video settings (encoding type and quality level) via `video_settings` parameter
- Changed default session mode from "CUSTOM" to "LITE" with VP8 as default encoding
- Updated examples to use sandbox mode with the official sandbox avatar

### DIFF
--- a/src/pipecat/services/heygen/api_liveavatar.py
+++ b/src/pipecat/services/heygen/api_liveavatar.py
@@ -9,6 +9,7 @@
 API to communicate with LiveAvatar Streaming API.
 """
 
+from enum import Enum
 from typing import Any, Dict, Optional
 
 import aiohttp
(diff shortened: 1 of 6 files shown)


src/pipecat/services/heygen/api_liveavatar.py
```diff
@@ -46,17 +47,43 @@ class CustomSDKLiveKitConfig(BaseModel):
...
-        mode (str): Session mode (default: "CUSTOM").
+        mode (str): Session mode (default: "LITE").
         avatar_id (str): Unique identifier for the avatar.
+        is_sandbox (bool): Enable sandbox mode (default: False).
         avatar_persona (AvatarPersona): Avatar persona configuration.
     """
 
     mode: str = "CUSTOM"
     avatar_id: str
+    video_settings: Optional[VideoSettings] = VideoSettings(encoding=VideoEncoding.VP8)
```
filipi87, line 87: I suspect that HeyGen changed their API and started sending video using H.264 instead of VP8 by default.

While this works on other platforms, it looks like the LiveKit Python SDK on macOS is unable to decode H.264, so we don’t receive any video frames in Pipecat.

We will use VP8 as the default codec, while still allowing users to switch codecs in case H.264 works on other platforns.

src/pipecat/services/heygen/api_liveavatar.py
```diff
@@ -46,17 +47,43 @@ class CustomSDKLiveKitConfig(BaseModel):
...
+    encoding: VideoEncoding
+    quality: VideoQuality = VideoQuality.high
+
+
 class LiveAvatarNewSessionRequest(BaseModel):
     """Request model for creating a LiveAvatar session token.
 
     Parameters:
-        mode (str): Session mode (default: "CUSTOM").
+        mode (str): Session mode (default: "LITE").
```
markbackman, line 77: You've changed the docstring to a default of `"LITE"` but the actual mode default on line 83 is still `"CUSTOM"`.

src/pipecat/services/heygen/api_liveavatar.py
markbackman, line 87: Add `video_settings` to the docstring above.

Also, `livekit_config` is missing from the docstring too.

markbackman (commented):
Just two small items then this is good to go.

src/pipecat/services/heygen/api_liveavatar.py
filipi87: True, nice catch.

filipi87: Fixed.

src/pipecat/services/heygen/api_liveavatar.py
filipi87: Fixed.

markbackman (approved):
LGTM!

PR #3657, opened 2026-02-05
Changing the ‘no audio received’ log from warning to debug.

filipi87:
Changing the ‘no audio received’ log from warning to debug.

This warning was introduced in version 101, when we deprecated the `vad_analyzer` parameter on `BaseInputTransport`.

Aside from seeing this `warning`, users do not appear to be experiencing any call issues. So changing this log to debug to avoid confusion.

### DIFF
--- a/src/pipecat/transports/base_input.py
+++ b/src/pipecat/transports/base_input.py
@@ -449,9 +449,7 @@ class BaseInputTransport(FrameProcessor):
                 if not audio_received:
                     continue
 
-                logger.warning(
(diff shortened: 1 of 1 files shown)


markbackman (approved):
LGTM, but it might be worth getting @aconchillo's say since he just made this change. I want to make sure he doesn't have something else in mind.

markbackman:
@filipi87 Actually, we should merge this so it can make it into the release, then @aconchillo can weigh in after if he agrees with this move. I'll merge this now.

PR #3659, opened 2026-02-05
Set VADParams stop_secs to 0.2 by default

markbackman:

This change needs to ship along with https://github.com/pipecat-ai/pipecat/pull/3637.

The new stop strategy breaks up the old 0.8 sec VAD stop_secs into a fixed 0.2 sec value for STT finalization and a developer configurable timeout for how long to wait for the user to stop speaking.

TODOs:
- [x] Confirm the variable and class names pending #3637. Update any names in the changelog before merging
- [ ] Update the quickstart example to align with this pattern after this feature is released


filipi87 (approved):
Makes sense.

changelog/3659.change.md
```diff
@@ -0,0 +1,5 @@
+- ⚠️ The default `VADParams` `stop_secs` default is changing from `0.8` seconds to `0.2` seconds.
+
+  - This change is being made to improve the performance of STT services. With a shorter `stop_secs` value, STT services using a local VAD can finalize sooner which results in receiving transcripts sooner. This ultimately reduces latency in the user to bot portion of the Pipeline.
+  - For those using the `TranscriptionUserTurnStopStrategy`, you can control how long to wait for the user to speak using the new `user_resume_speaking_timeout`.
```
filipi87: I guess this is the only thing we might need to adjust if we change the variable name, right ?

In any case, it would probably still be fine to merge this PR as is, to avoid having to resolve conflicts across all these files.

kompfner: We're still planning on renaming `TranscriptionUserTurnStopStrategy` to `TimeoutUserTurnStopStrategy`, right?

changelog/3659.change.md
```diff
@@ -0,0 +1,5 @@
...
+  - For those using the `TurnAnalyzerUserTurnStopStrategy`, no timeout value needs to be set. The turn analyzer dynamically adjust the user wait time based on the audio input.
```
kompfner:
```suggestion
  - For those using the `TurnAnalyzerUserTurnStopStrategy`, no timeout value needs to be set. The turn analyzer dynamically adjusts the user wait time based on the audio input.
```

changelog/3659.change.md
```diff
@@ -0,0 +1,5 @@
+- ⚠️ The default `VADParams` `stop_secs` default is changing from `0.8` seconds to `0.2` seconds.
+
+  - This change is being made to improve the performance of STT services. With a shorter `stop_secs` value, STT services using a local VAD can finalize sooner which results in receiving transcripts sooner. This ultimately reduces latency in the user to bot portion of the Pipeline.
```
kompfner: The main motivation for this change was so that folks should've have to touch the `stop_secs` depending on which user turn stop strategy they were using, not performance, right?

For folks leaving `stop_secs` and the turn strategies at their default value, they're still getting an effective 0.8s wait for the user to resume speaking. Any improvements to performance have come from how we handle transcriptions, not this.

kompfner: So then actually we can assuage the reader that, most likely, this change with the ⚠️ won't require any action on their part!

examples/foundational/04-transports-small-webrtc.py
```diff
@@ -90,7 +89,7 @@ async def run_example(webrtc_connection: SmallWebRTCConnection):
             user_turn_strategies=UserTurnStrategies(
                 stop=[TurnAnalyzerUserTurnStopStrategy(turn_analyzer=LocalSmartTurnAnalyzerV3())]
             ),
-            vad_analyzer=SileroVADAnalyzer(params=VADParams(stop_secs=0.2)),
+            vad_analyzer=SileroVADAnalyzer(),
```
kompfner, line 92: Love seeing all these melt away! 🫠

kompfner (commented):
Feedback only on the changelog! Let me know if a synchronous discussion would help!

changelog/3659.change.md
markbackman: No, there's an actual benefit to reducing the value. Finalization happens sooner which means the transcripts are generated much faster.

changelog/3659.change.md
markbackman: Done!

changelog/3659.change.md
markbackman: Thank you

aconchillo (approved):

aconchillo:
Yes, please!

PR #3660, opened 2026-02-06
Attach asyncio.Event to InterruptionFrame for completion signaling

aconchillo:
## Summary

- Moves the interruption wait event from per-processor instance state (`_wait_interruption_event`) to the frame itself (`InterruptionFrame.event`). The event is created in `push_interruption_task_frame_and_wait()`, threaded through `InterruptionTaskFrame` → `InterruptionFrame`, and set when the frame reaches the pipeline sink via `InterruptionFrame.complete()`.
- Scopes each event to a single interruption flow rather than sharing mutable state on the processor, avoiding potential issues with concurrent interruptions.
- Adds a 2s timeout warning to help diagnose cases where `InterruptionFrame.complete()` is never called (e.g. a processor consuming the frame before it reaches the sink).
- Handles the edge case where the user is muted in `LLMUserAggregator` — since the frame won't propagate to the sink, `complete()` is called explicitly.

## Important: InterruptionFrames must always complete

`InterruptionFrame` now carries an optional `asyncio.Event` that signals when the interruption has fully traversed the pipeline. Any code that consumes or blocks an `InterruptionFrame` before it reaches the pipeline sink **must** call `frame.complete()` to avoid hanging `push_interruption_task_frame_and_wait()` callers indefinitely.

The framework handles this automatically in two places:
- **Pipeline sink** (`PipelineTask`): calls `complete()` when the frame reaches the end of the pipeline (the normal path).
- **`LLMUserAggregator`**: calls `complete()` when the user is muted, since the frame won't propagate further.

If you write a custom processor that stops an `InterruptionFrame` from flowing downstream, you must call `frame.complete()` explicitly. A warning is logged after 2 seconds if an `InterruptionFrame` has not completed, to help identify these cases.


src/pipecat/frames/frames.py
```diff
@@ -1121,12 +1122,27 @@ class InterruptionFrame(SystemFrame):
...
+    Parameters:
+        event: Optional event set when the frame has fully traversed the
+            pipeline.
+
     """
 
-    pass
+    event: Optional[asyncio.Event] = None
+
+    def complete(self):
```
filipi87, line 1136: I guess this is the first frame where we’ve actually added a method to it, right ? I think that was a clever idea for this case. 

I think we already don’t do that for `InterruptionFrame`, but in this case it means we should never invoke `broadcast_frame`. It might be worth adding a comment explicitly stating that we should never do that for the `InterruptionFrame`.

Also, if you improve the comments, it’s still mentioned that this frame is emitted by `BaseInputTransport`. 😅

filipi87: Actually, I think it would still be able to broadcast, since even if we pass an event, both frames would have the same event.

filipi87 (approved):
I think this is a nice improvement. Clever idea to solve this. 🚀

src/pipecat/frames/frames.py
aconchillo: That's right.

aconchillo:
> I think this is a nice improvement. Clever idea to solve this. 🚀

I attempted the same in the past, but didn't realize (well, I kind of did) we just need to trigger the event where the event is blocked. In that case it was the STTMuteFilter. Which, btw, I just updated the PR with a new commit to fix STTMuteFilter.

src/pipecat/frames/frames.py
aconchillo: > 
> Also, if you improve the comments, it’s still mentioned that this frame is emitted by `BaseInputTransport`. 😅

Done!

src/pipecat/processors/filters/stt_mute_filter.py
```diff
@@ -234,6 +234,8 @@ async def process_frame(self, frame: Frame, direction: FrameDirection):
                 await self.push_frame(frame, direction)
             else:
                 logger.trace(f"{frame.__class__.__name__} suppressed - STT currently muted")
+                if isinstance(frame, InterruptionFrame):
+                    frame.complete()
```
filipi87, line 242: Yep, nice catch.

PR #3663, opened 2026-02-06
fix: close stream on cancellation for SambaNova and Google OpenAI services

### LINKED ISSUE #3639: LLM services missing stream cleanup on cancellation

lukepayyapilli:
## Summary

- [x] Apply the same stream cleanup fix to `SambaNovaLLMService` and `GoogleLLMOpenAIBetaService`.
- [x] These services override `_process_context` and had the same issue where `CancelledError` during stream iteration would leave the underlying socket open.
- [x] Add tests verifying stream closes on cancellation.

## Changes

Wrap stream iteration in `async with chunk_stream:` to ensure proper cleanup on cancellation/exception.

## Related

- [x] Fixes #3639.
- [x] Follow-up to #3640 (merged fix for `BaseOpenAILLMService`).
- [x] Per [feedback](https://github.com/pipecat-ai/pipecat/issues/3639#issuecomment-3856886350) from @markbackman.

### DIFF
--- a/changelog/3663.fixed.md
+++ b/changelog/3663.fixed.md
@@ -1,0 +1,1 @@
+- Fixed `SambaNovaLLMService` and `GoogleLLMOpenAIBetaService` streams not being closed on cancellation/exception, which could leak sockets.
--- a/src/pipecat/services/google/llm_openai.py
+++ b/src/pipecat/services/google/llm_openai.py
@@ -91,52 +91,55 @@ class GoogleLLMOpenAIBetaService(OpenAILLMService):
             ChatCompletionChunk
         ] = await self._stream_chat_completions_specific_context(context)
 
-        async for chunk in chunk_stream:
-            if chunk.usage:
-                tokens = LLMTokenUsage(
-                    prompt_tokens=chunk.usage.prompt_tokens or 0,
-                    completion_tokens=chunk.usage.completion_tokens or 0,
-                    total_tokens=chunk.usage.total_tokens or 0,
-                )
-                await self.start_llm_usage_metrics(tokens)
+        # Use context manager to ensure stream is closed on cancellation/exception.
+        # Without this, CancelledError during iteration leaves the underlying socket open.
+        async with chunk_stream:
+            async for chunk in chunk_stream:
+                if chunk.usage:
+                    tokens = LLMTokenUsage(
+                        prompt_tokens=chunk.usage.prompt_tokens or 0,
+                        completion_tokens=chunk.usage.completion_tokens or 0,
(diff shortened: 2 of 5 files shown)


markbackman (approved):
Looks great! Thanks again 🙇

PR #3665, opened 2026-02-06
[inworld] Open a new context per agent turn

ianbbqzy:

* opening the context on LLMFullResponseStartFrame to minimize the impact context creation overhead

### DIFF
--- a/src/pipecat/services/inworld/tts.py
+++ b/src/pipecat/services/inworld/tts.py
@@ -37,6 +37,7 @@ from pipecat.frames.frames import (
     ErrorFrame,
     Frame,
     InterruptionFrame,
+    LLMFullResponseStartFrame,
     StartFrame,
     TTSAudioRawFrame,
     TTSStartedFrame,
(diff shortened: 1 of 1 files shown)


markbackman:
I'm not sure that I understand this PR and why it's needed. Can you elaborate? Are you seeing any race conditions you're trying to handle? The same pattern that InworldTTSService uses (on main) is used in other services. This works well in those cases.

ianbbqzy:
Hi @markbackman I think there are two issues regarding context management:

1. Should we close a context per agent turn?
For both Inworld and 11lab SDKs, we don't close a context on the end of an agent turn, unless it's interruption. On the other hand, [Cartesia](https://github.com/pipecat-ai/pipecat/blob/4f77c532fb9d2b69ef3adfb3b008f913b7658cf3/src/pipecat/services/cartesia/tts.py#L548) does close a context and create a new one each turn 

From our [conversation](https://inworldai.slack.com/archives/C095V4EA23T/p1770162125744579?thread_ts=1768064186.468669&cid=C095V4EA23T), you mentioned that Cartesia's pattern is preferred 

2. Regardless of whether we start following Cartesia's pattern, at least on the very first turn, Inworld TTS will need to create a context when one isn't available. And if we do refactor to follow Cartesia's pattern, as I did in this PR, every agent turn, when tts is requested, we need to create a new context because one was closed at the end of previous turn.

Creating a context with Inworld TTS websocket could introduce ~10-30ms additional latency. That's why also in this PR, I want to start the context creation as soon as possible (when LLM response starts, but before the first sentence is aggregated) such that the context creation overlaps with waiting for the first sentence, effectively reducing latency by 10-30ms in most cases.

Please let me know if you have more questions!

markbackman:
I'm going to tag in @filipi87 who has worked most closely with the TTS services and base TTSService classes over the last few days. He'll have the best recommendation of how proceed here.

He's also continuing refactoring in places, so it could be helpful to coordinate anyway.

ianbbqzy:
superceded by https://github.com/pipecat-ai/pipecat/pull/4013

PR #3667, opened 2026-02-06
[inworld] aggregate_sentence mode needs trailing space

ianbbqzy:

Spaces from the LLM generations are stripped in the text aggregator. Inworld server does not append additional space in its own aggregation, resulting in words to be joined, such as in `assistant.How` in `Hello! I’m your friendly AI assistant. How can I assist you today?`. The model could still synthesize `assistant.How` correctly but it causes the timestamps alignment to have this joined word instead.

### DIFF
--- a/changelog/3667.fixed.md
+++ b/changelog/3667.fixed.md
@@ -1,0 +1,1 @@
+- Fixed an issue in `InworldTTSService` where punctuation was pronounced. Now, the `InworldTTSService` ensures proper spacing between sentences, resolving pronunciation issues.
--- a/src/pipecat/services/inworld/tts.py
+++ b/src/pipecat/services/inworld/tts.py
@@ -439,6 +439,7 @@ class InworldTTSService(AudioContextWordTTSService):
         encoding: str = "LINEAR16",
         params: InputParams = None,
         aggregate_sentences: bool = True,
+        append_trailing_space: bool = True,
         **kwargs: Any,
     ):
         """Initialize the Inworld WebSocket TTS service.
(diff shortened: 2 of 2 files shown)


changelog/3667.fixed.md
```diff
@@ -0,0 +1 @@
+- Inworld TTS requires space to not be stripped with the simple text aggregator
```
markbackman: Can you change the entry to something like this?

```suggestion
- Fixed an issue in `InworldTTSService` where punctuation was pronounced. Now, the `InworldTTSService` ensures proper spacing between sentences, resolving pronunciation issues.
```

src/pipecat/services/inworld/tts.py
```diff
@@ -460,6 +460,8 @@ def __init__(
             pause_frame_processing=True,
             sample_rate=sample_rate,
             aggregate_sentences=aggregate_sentences,
+            # always append trailing space for Inworld when aggregating sentences
```
markbackman: Nit: remove comment.

src/pipecat/services/inworld/tts.py
```diff
@@ -460,6 +460,8 @@ def __init__(
...
+            append_trailing_space=aggregate_sentences,
```
markbackman: It might be better to explicitly set this rather than it be set from another variable.

If a developer is not sending in sentences, they'll be doing things more manually and should think about the settings.

```suggestion
            append_trailing_space=True,
```

ianbbqzy:
Thanks @markbackman, addressed the comments

markbackman (approved):
LGTM!

PR #3668, opened 2026-02-06
Buffer internal frames during ParallelPipeline lifecycle sync

aconchillo:
## Summary
- Fixes a bug where processors inside parallel sub-pipelines could push frames that escape the `ParallelPipeline` before all branches finish processing a lifecycle frame (`StartFrame`/`EndFrame`/`CancelFrame`).
- Adds a `_synchronizing` flag and `_buffered_frames` list to `ParallelPipeline`. During lifecycle frame synchronization, non-lifecycle frames from internal processors are buffered and flushed only after all branches complete.
- Existing pause/resume mechanism is kept to prevent external lifecycle frames from entering during synchronization.

## Test plan
- [x] Added `test_parallel_internal_frames_buffered_during_start` to verify internal frames are buffered and released after `StartFrame` synchronization
- [x] All existing pipeline tests pass

### DIFF
--- a/.claude/skills/changelog/SKILL.md
+++ b/.claude/skills/changelog/SKILL.md
@@ -7,23 +7,30 @@ Create changelog files for the important commits in this PR. The PR number is pr
 
 ## Instructions
 
-1. First, check what commits are on the current branch compared to main:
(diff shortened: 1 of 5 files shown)


markbackman (approved):
Makes sense! Thanks for the quick fix.

PR #3672, opened 2026-02-07
Filter RTVIObserver to downstream frames only and broadcast FunctionCallCancelFrame

markbackman:

@mattieruth found that we were sending multiple events. This started as we moved everything to broadcasting unique frames. I audited current frames and everything goes downstream (and some upstream). I'm adding filtering for the RTVIObserver to only emit events on downstream frames as a convention.

I don't know of a better solution to this, but I think it works for now.

### DIFF
--- a/changelog/3672.changed.md
+++ b/changelog/3672.changed.md
@@ -1,0 +1,1 @@
+- Changed `FunctionCallCancelFrame` to broadcast in both directions for consistency with other function call frames.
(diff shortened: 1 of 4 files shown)


filipi87 (approved):
Agreed. I think this approach makes sense for now. 👍

mattieruth (approved):
I approve, though, in testing, it looks like the comment on line 1026 is wrong. The events aren't disabled by default, they are NONE by default.

markbackman:
> I approve, though, in testing, it looks like the comment on line 1026 is wrong. The events aren't disabled by default, they are NONE by default.

Good catch. Updated the comment.

PR #3675, opened 2026-02-08
Add silence-based keepalive to WebsocketSTTService

markbackman:
## Summary
- Adds opt-in `keepalive_timeout` and `keepalive_interval` params to `WebsocketSTTService` that send silent audio when idle, preventing server-side timeout disconnects (e.g. behind a `ServiceSwitcher`)
- Subclasses override `_send_keepalive(silence)` to wrap silence in their service-specific protocol; the default sends raw PCM bytes
- Enables keepalive for ElevenLabs (10s timeout), Cartesia (120s), Gladia (20s), and Soniox (1s, protocol-level message)
- Removes per-service custom keepalive tasks in favor of the shared base class infrastructure

## Test plan
- [x] `uv run pytest` — 529 passed, 47 skipped
- [x] `uv run ruff check && uv run ruff format --check` — all clean
- [x] Manual test with ServiceSwitcher: verify idle ElevenLabs/Cartesia/Gladia/Soniox STT connections stay alive when switched away

### DIFF
--- a/src/pipecat/services/elevenlabs/stt.py
+++ b/src/pipecat/services/elevenlabs/stt.py
@@ -459,6 +459,8 @@ class ElevenLabsRealtimeSTTService(WebsocketSTTService):
         super().__init__(
             sample_rate=sample_rate,
             ttfs_p99_latency=ttfs_p99_latency,
+            keepalive_timeout=10,
+            keepalive_interval=5,
             **kwargs,
         )
 
@@ -611,9 +613,9 @@ class ElevenLabsRealtimeSTTService(WebsocketSTTService):
 
     async def _connect(self):
         """Establish WebSocket connection to ElevenLabs Realtime STT."""
-        await super()._connect()
-
         await self._connect_websocket()
+
+        await super()._connect()
 
         if self._websocket and not self._receive_task:
             self._receive_task = self.create_task(self._receive_task_handler(self._report_error))
@@ -627,6 +629,21 @@ class ElevenLabsRealtimeSTTService(WebsocketSTTService):
             self._receive_task = None
 
         await self._disconnect_websocket()
+
+    async def _send_keepalive(self, silence: bytes):
+        """Send silent audio wrapped in ElevenLabs' JSON protocol.
+
+        Args:
+            silence: Silent 16-bit mono PCM audio bytes.
+        """
+        audio_base64 = base64.b64encode(silence).decode("utf-8")
+        message = {
+            "message_type": "input_audio_chunk",
+            "audio_base_64": audio_base64,
+            "commit": False,
+            "sample_rate": self.sample_rate,
+        }
+        await self._websocket.send(json.dumps(message))
 
     async def _connect_websocket(self):
         """Connect to ElevenLabs Realtime STT WebSocket endpoint."""
--- a/src/pipecat/services/stt_service.py
+++ b/src/pipecat/services/stt_service.py
@@ -14,6 +14,7 @@ from abc import abstractmethod
 from typing import Any, AsyncGenerator, Dict, Mapping, Optional
 
 from loguru import logger
+from websockets.protocol import State
 
 from pipecat.frames.frames import (
     AudioRawFrame,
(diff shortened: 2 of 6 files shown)


markbackman:
## Code review

src/pipecat/services/elevenlabs/stt.py
```diff
@@ -617,8 +627,48 @@ async def _disconnect(self):
             await self.cancel_task(self._receive_task)
             self._receive_task = None
 
+        if self._keepalive_task:
+            await self.cancel_task(self._keepalive_task)
+            self._keepalive_task = None
+
         await self._disconnect_websocket()
 
+    async def _keepalive_task_handler(self):
```
filipi87: Would it make sense to add it inside `WebsocketSTTService` instead? That way, any STT that uses WebSockets could create a keepalive if needed.

kompfner:
Is the service billed based on how long the connection remains? If so...might it not be desirable for the connection to drop when it's not the active service in a service switcher? (But, of course, not be stuck in a cycle of dropping the connection then reconnecting...)

markbackman:
> Is the service billed based on how long the connection remains? If so...might it not be desirable for the connection to drop when it's not the active service in a service switcher? (But, of course, not be stuck in a cycle of dropping the connection then reconnecting...)

The thing is that Pipecat will automatically reconnect. You'll still be billed for the same amount of time, you just see a lot of reconnect warnings in your logs.

I've thought about the ServiceSwitcher case for this—I wondered if the service should actually disconnect until used. But, in thinking through it, I'm not sure if that's possible without some data loss. In the case of an STTService, it needs to be able to start receiving AudioRawFrames as soon as the switch happens. That switch requires that a connection be established prior to the switch itself. In a way, the contract is that a ServiceSwitcher requires a warm reserve.

This is very performant, but it increases the cost. Any thoughts on how this could work otherwise?

kompfner:
> Any thoughts on how this could work otherwise?

Not off the top of my head. What you implemented here seems preferable to an infinite reconnect loop.

In the ServiceSwitcher case specifically, it _might_ be worth a momentary reconnect hiccup in the (hopefully rare) case of having to switch services, though...

One way of mitigating it is through a mechanism telling services whether or not they're active in a switcher, and having them reconnect only when activated. Giving services awareness of whether they're active in a switcher, or even whether they're inside of a switcher, is something we've avoided so far. But there could be other benefits of communicating to a service whether it's been activated or deactivated—for example, in the STT metadata work you were just doing in the other PR, services could re-broadcast metadata in response to being activated.

markbackman:
We talked about this on Slack. The charge per connection is on audio sent to the service, not on the connection time. So, this keepalive strategy is a good balance between cost and performance. I'm going to proceed with this change.

markbackman:
Ok, I moved the change the WebsocketSTTService and enabled it for services which I know support a keepalive. The others don't use it, since the keepalive isn't initialized. If we find out it's needed, it's simple to add the functionality to those services now.

filipi87 (approved):
LGTM!

src/pipecat/services/stt_service.py
```diff
@@ -543,18 +547,100 @@ class WebsocketSTTService(STTService, WebsocketService):
...
+            )
+
+    async def _disconnect(self):
+        """Disconnect and cancel keepalive task."""
+        await super()._disconnect()
+        if self._keepalive_task:
+            await self.cancel_task(self._keepalive_task)
+            self._keepalive_task = None
+
+    async def _keepalive_task_handler(self):
```
aconchillo, line 629: we might want to implement `_reconnect_websocket` to start keepalive task again.

aconchillo: something like:

```
    async def _reconnect_websocket(self, attempt_number: int) -> bool:
        result = super()._reconnect_websocket(attempt_number)
        if result:
            await self._cancel_keepalive_task()
            self._create_keepalive_task()
        return result
```

aconchillo (approved):

aconchillo:
LGTM!

PR #3678, opened 2026-02-08
Enable pyright basic type checking for core framework

markbackman:
## Summary

- Add `[tool.pyright]` configuration to `pyproject.toml` (basic mode, Python 3.10, excluding `services/`and `adapters/`)
- Fix all 276 type errors in the core framework across 64 files
- Establish a CI-ready type checking baseline as Pipecat approaches 1.0

## Details

Pipecat ships a `py.typed` marker and has `pyright` as a dev dependency, but had no static type checking enabled. Running pyright at basic level showed 1,306 errors total — 810 in services/adapters (out of scope), 276 in the core framework. This PR fixes all core framework errors.

**Errors fixed by directory:**

| Directory | Errors | Key Fix Patterns |
|-----------|--------|-----------------|
| `processors/` | 79 | `cast()` for OpenAI/LLM type boundaries, RTVI data types, Optional guards |
| `transports/` | 68 | Optional member guards, SDK type ignores, assertions |
| `audio/` | 43 | Optional dep import guards, None assertions, ML API ignores |
| `serializers/` | 31 | `_params` type narrowing in subclasses, protobuf ignores |
| `runner/` | 16 | Optional→concrete param types, None guards, HTTPMethod compat |
| `pipeline/` | 11 | List covariance casts, Optional guards |
| `turns/` | 10 | `TYPE_CHECKING` base class pattern for mixin |
| `utils/` | 10 | `getattr()` for dynamic access, return type fixes |
| Other | 8 | Type annotations, bug fixes (`Set[str]`→`Set[int]`) |

**Fix strategies used:**
- `assert x is not None` / `if x is not None:` guards for Optional member access
- `cast()` at type boundaries (OpenAI ChatCompletion types ↔ LLMContextMessage)
- `# type: ignore` for optional dependency imports and SDK type mismatches (livekit, websockets, ML models)
- Type narrowing annotations (`self._params: SubclassInputParams`) for serializers
- `TYPE_CHECKING` conditional base class for mixin pattern (turns/)

## Test plan

- [x] `uv run pyright src/pipecat` — 0 errors, 0 warnings, 0 informations
- [x] `uv run ruff check` — all checks passed
- [x] `uv run ruff format --check` — all files formatted
- [x] `uv run pytest` — 569 passed, 2 pre-existing failures, 1 skipped

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -212,6 +212,14 @@ ignore = [
 [tool.ruff.lint.pydocstyle]
 convention = "google"
 
+[tool.pyright]
+typeCheckingMode = "basic"
+pythonVersion = "3.10"
+exclude = [
+    "src/pipecat/services/",
+    "src/pipecat/adapters/",
+]
+
 [tool.coverage.run]
 command_line = "--module pytest"
 source = ["src"]
--- a/src/pipecat/processors/aggregators/llm_response_universal.py
+++ b/src/pipecat/processors/aggregators/llm_response_universal.py
@@ -20,7 +20,7 @@ from typing import Any, Dict, List, Literal, Optional, Set, Type, cast
 
 from loguru import logger
 
-from pipecat.adapters.schemas.tools_schema import ToolsSchema
+from pipecat.adapters.schemas.tools_schema import AdapterType, ToolsSchema
 from pipecat.audio.vad.vad_analyzer import VADAnalyzer
 from pipecat.audio.vad.vad_controller import VADController
 from pipecat.frames.frames import (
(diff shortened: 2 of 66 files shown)


markbackman:
Considerations:

- There are a number of assertions in the code, which we might want to avoid. We could instead use local variable narrowing as an option, e.g.

Is:
```
assert self._mixer is not None
```

To:
```
mixer = self._mixer
  if mixer:
      frame.audio = await mixer.mix(frame.audio)
```

Or, we could just guard and return an error, e.g.
```
  if self._mixer is None:
      raise RuntimeError("mixer is not initialized")
```

This would clean up a number of things. I don't think we have a precedent set yet. I think narrowing is a good choice.

- Ignores: There's are a fair number of these. I tried to reduce the number to just import ignores, but there are still more...

src/pipecat/processors/aggregators/llm_response_universal.py
```diff
@@ -272,7 +280,9 @@ def set_tool_choice(self, tool_choice: Literal["none", "auto", "required"] | dic
         Args:
             tool_choice: Tool choice configuration for the context.
         """
-        self._context.set_tool_choice(tool_choice)
+        from pipecat.processors.aggregators.llm_context import LLMContextToolChoice
+
+        self._context.set_tool_choice(cast(LLMContextToolChoice, tool_choice))
```
filipi87, line 285: It looks like in a couple of different places we’re using `cast`. Is there any reason not to change the signature instead, like this?
`def set_tool_choice(self, tool_choice: LLMContextToolChoice)`

filipi87: Same for the other places where we’re using cast. I guess we should check whether changing the signature would be possible instead.

pyproject.toml
```diff
@@ -212,6 +212,14 @@ ignore = [
 [tool.ruff.lint.pydocstyle]
 convention = "google"
 
+[tool.pyright]
+typeCheckingMode = "basic"
+pythonVersion = "3.10"
+exclude = [
+    "src/pipecat/services/",
```
filipi87, line 219: I guess we’re excluding them for now to avoid too many changes?

markbackman:
I'm going to close this. We can add in smaller chunks.

PR #3684, opened 2026-02-09
AIC model caching

gkmngrgn:
# Summary

- **AIC model caching:** Read-only AIC models are now shared across `AICFilter` instances via a singleton `AICModelManager` in `aic_filter.py`. Multiple filters using the same `model path` or (`model_id`, `model_download_dir`) share one loaded `Model` and avoid duplicate loads.
- **Reference counting:** The manager acquires a reference when a filter calls `acquire()` in `start()` and releases it in `stop()`. When the last reference is dropped, the model is evicted from the cache.
- **Concurrent load deduplication:** If several filters request the same uncached model at once, only one load runs; the rest await the same in-flight task and then use the cached result.
- **Internal helpers:** Cache key is built by `_get_cache_key()`; reference updates use `_increment_reference()` and `_store_new_reference()` so logic is centralized and model is never optional where it is used.

# Testing

- `uv run pytest tests/test_aic_filter.py tests/test_aic_vad.py -v` — all 28 tests pass, including:
  - `test_model_manager_reference_count`: ref count increments on get_model, decrements on release, model evicted at 0.
  - `test_model_manager_concurrent_load_deduplication`: concurrent get_model calls for the same key trigger a single `Model.from_file` and all callers receive the same model.

**No breaking changes:** AICFilter’s public API (constructor and start/stop) is unchanged.

### DIFF
--- a/src/pipecat/audio/filters/aic_filter.py
+++ b/src/pipecat/audio/filters/aic_filter.py
@@ -77,16 +77,19 @@ class AICModelManager:
         """Run the actual load (file or download). Separate to allow create_task and deduplication."""
         if model_path is not None:
             logger.debug(f"Loading AIC model from file: {model_path}")
-            return Model.from_file(str(model_path))
+            model_path_str = str(model_path)
 
-        if model_id is not None and model_download_dir is not None:
+        elif model_id is not None and model_download_dir is not None:
             logger.debug(f"Downloading AIC model: {model_id}")
             model_download_dir.mkdir(parents=True, exist_ok=True)
-            path = await Model.download_async(model_id, str(model_download_dir))
-            logger.debug(f"Model downloaded to: {path}")
-            return Model.from_file(path)
+            model_path_str = await Model.download_async(model_id, str(model_download_dir))
+            logger.debug(f"Model downloaded to: {model_path_str}")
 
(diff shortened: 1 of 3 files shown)


steckes (approved):

src/pipecat/audio/filters/aic_filter.py
```diff
@@ -33,6 +36,174 @@
...
+        cache_key: str,
+        *,
+        model_path: Optional[Path] = None,
+        model_id: Optional[str] = None,
+        model_download_dir: Optional[Path] = None,
+    ) -> Model:
+        """Run the actual load (file or download). Separate to allow create_task and deduplication."""
+        if model_path is not None:
+            logger.debug(f"Loading AIC model from file: {model_path}")
+            return Model.from_file(str(model_path))
```
Fl1tzi: I think this could block (especially IO) inside the async function. Not sure if this is really an issue.

gkmngrgn: It won't be a problem because there's one more layer for keeping the concurrency deduplication, that's why we have `_loading` definition.

When we added async support in our sdk, it will help us to make the code simpler here.

related -> https://github.com/pipecat-ai/pipecat/pull/3684/changes/2036757b84062dc0804f77ecf2466c39b3932461

markbackman (approved):
LGTM

PR #3686, opened 2026-02-09
Fix LLMAssistantAggregator aggregating STT transcription frames

### LINKED ISSUE #3638: LLMAssistantAggregator aggregates InterimTranscriptionFrames
### pipecat version

0.0.101

### Python version

3.12.11

### Operating System

### Issue description

After the bot finished speaking, the `LLMAssistantAggregator` sometimes aggregates the first few words of a user as an additional assistant message. 

```json
{
  role: "assistant",
  content: "Hi, how can I help you today?"
},
{
  role: "assistant",
  content: "I like"
},
{
  role: "user",
  content: "I like chocolate."
}
```

We observed that the LLMAssistantAggregator handles all Frames of type `TextFrame` equally: https://github.com/pipecat-ai/pipecat/blob/6bcfb40d1230966cfb3affc4748d28ad804f6d76/src/pipecat/processors/aggregators/llm_response_universal.py#L855C1-L856C43
This includes all child classes including the `InterimTranscriptionFrame`, emitted by STT services.

Our transport implements custom emission logic for `Bot<Started, Stopped, ''>SpeakingFrame`s that might be part of this issue. But wouldn't it be safer for the assistant aggregator to only aggregate text from appropriate frames?
The `TTSService`, for example, checks for transcription-related subclasses before processing the text of a TextFrame: https://github.com/pipecat-ai/pipecat/blob/6bcfb40d1230966cfb3affc4748d28ad804f6d76/src/pipecat/services/tts_service.py#L435C1-L439C11

We fixed the issue for now by wrapping the assistant aggregator in a class that deflects InterimTranscriptionFrames:
```python
class CustomLLMAssistantAggregator(LLMAssistantAggregator):
    """Custom LLMAssistantAggregator that stops InterimTranscriptionFrames from being processed by pipecat's assistant aggregator.

    Without this fix, parts of the user's spoken text end up in assistant messages in the LLM's context.
    """

        """Process frames for assistant response aggregation and function call management.

        Args:
            frame: The frame to process.
        """
        if isinstance(frame, InterimTranscriptionFrame):
        else:

```

### Reproduction steps

Start a pipeline with:
- Deepgram STT
- the new universal LLMContextAggregatorPair
- ElevenLabs TTS
- talk for multiple turns
- observe LLM context

### Expected behavior

Spoken user text should not appear in assistant messages.

```json
{
  role: "assistant",
  content: "Hi, how can I help you today?"
},
{
  role: "user",
  content: "I like chocolate."
}
```

### Actual behavior

Parts of spoken user text are included in additional assistant messages:

```json
{
  role: "assistant",
  content: "Hi, how can I help you today?"
},
{
  role: "assistant",
  content: "I like"
},
{
  role: "user",
  content: "I like chocolate."
}
```

### Logs

```shell

```

lukepayyapilli:
## Summary
- [x] Narrowed `LLMAssistantAggregator.process_frame` type check from `isinstance(frame, TextFrame)` to `isinstance(frame, LLMTextFrame)`.
- [x] Updated `_handle_text` type annotation from `TextFrame` to `LLMTextFrame` and removed unused import.
- [x] Fixed `LangchainProcessor` to emit `LLMTextFrame` instead of plain `TextFrame`, consistent with all other LLM services.
- [x] Updated legacy assistant aggregator tests to use `LLMTextFrame` (subclass of `TextFrame`, so legacy aggregators still match).
- [x] Added regression test verifying `InterimTranscriptionFrame` and `TranscriptionFrame` pass through without being aggregated.
- [x] Added changelog.

## Approach

`LLMAssistantAggregator` used `isinstance(frame, TextFrame)` to catch text for aggregation. Since `InterimTranscriptionFrame` and `TranscriptionFrame` are `TextFrame` subclasses, partial STT output leaked into the LLM context as assistant messages corrupting conversation history.

**Allow-list (`LLMTextFrame`) over deny-list:** Rather than adding `and not isinstance(frame, InterimTranscriptionFrame) and not isinstance(frame, TranscriptionFrame)`, this PR switches to checking `isinstance(frame, LLMTextFrame)` directly. 

## Rationale

- Every LLM service in the codebase already emits `LLMTextFrame`, never plain `TextFrame`. This was verified across all 11 service files.
- Allow-list is closed by default so future `TextFrame` subclasses won't accidentally leak in.
- Simpler code (one type check vs. three) and makes intent explicit: the assistant aggregator aggregates LLM text.

**`LangchainProcessor` fix:** This was the one processor emitting plain `TextFrame` instead of `LLMTextFrame`. Since it wraps a language model chain and already emits `LLMFullResponseStartFrame`/`LLMFullResponseEndFrame`, it should emit `LLMTextFrame` to be consistent with other LLM services. `LLMTextFrame` also sets `includes_inter_frame_spaces = True` by default, so the manual assignment is no longer needed.

## Test Updates

The legacy `BaseTestAssistantContextAggregator` tests were updated from `TextFrame` to `LLMTextFrame`. Since `LLMTextFrame` is a subclass of `TextFrame`, the legacy provider aggregators (OpenAI, Anthropic, Google, AWS) that still check `isinstance(frame, TextFrame)` continue to match. The `test_multiple_text_stripped` test (which relies on the aggregator inserting spaces between frames) is overridden in the universal subclass because `LLMTextFrame` always includes inter-frame spaces so that scenario doesn't apply.

Closes #3638

### DIFF
--- a/src/pipecat/processors/aggregators/llm_response_universal.py
+++ b/src/pipecat/processors/aggregators/llm_response_universal.py
@@ -44,13 +44,13 @@ from pipecat.frames.frames import (
     LLMRunFrame,
     LLMSetToolChoiceFrame,
     LLMSetToolsFrame,
+    LLMTextFrame,
     LLMThoughtEndFrame,
     LLMThoughtStartFrame,
     LLMThoughtTextFrame,
     LLMUpdateSettingsFrame,
     SpeechControlParamsFrame,
     StartFrame,
-    TextFrame,
     TranscriptionFrame,
     UserImageRawFrame,
     UserSpeakingFrame,
@@ -852,7 +852,7 @@ class LLMAssistantAggregator(LLMContextAggregator):
             await self._handle_llm_start(frame)
         elif isinstance(frame, LLMFullResponseEndFrame):
             await self._handle_llm_end(frame)
-        elif isinstance(frame, TextFrame):
+        elif isinstance(frame, LLMTextFrame):
             await self._handle_text(frame)
         elif isinstance(frame, LLMThoughtStartFrame):
             await self._handle_thought_start(frame)
@@ -1076,7 +1076,7 @@ class LLMAssistantAggregator(LLMContextAggregator):
         self._started -= 1
         await self._trigger_assistant_turn_stopped()
 
-    async def _handle_text(self, frame: TextFrame):
+    async def _handle_text(self, frame: LLMTextFrame):
         if not self._started or not frame.append_to_context:
             return
 
--- a/changelog/3638.fixed.md
+++ b/changelog/3638.fixed.md
@@ -1,0 +1,1 @@
+- Fixed LLMAssistantAggregator incorrectly aggregating STT transcription frames as assistant messages.
--- a/src/pipecat/processors/frameworks/langchain.py
+++ b/src/pipecat/processors/frameworks/langchain.py
@@ -15,7 +15,7 @@ from pipecat.frames.frames import (
     LLMContextFrame,
     LLMFullResponseEndFrame,
     LLMFullResponseStartFrame,
-    TextFrame,
+    LLMTextFrame,
 )
 from pipecat.processors.aggregators.openai_llm_context import OpenAILLMContextFrame
 from pipecat.processors.frame_processor import FrameDirection, FrameProcessor
@@ -107,8 +107,7 @@ class LangchainProcessor(FrameProcessor):
                 {self._transcript_key: text},
                 config={"configurable": {"session_id": self._participant_id}},
             ):
-                frame = TextFrame(self.__get_token_value(token))
-                frame.includes_inter_frame_spaces = True
+                frame = LLMTextFrame(self.__get_token_value(token))
                 await self.push_frame(frame)
         except GeneratorExit:
             logger.warning(f"{self} generator was closed prematurely")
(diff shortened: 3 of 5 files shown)


src/pipecat/processors/aggregators/llm_response_universal.py
```diff
@@ -858,7 +852,7 @@ async def process_frame(self, frame: Frame, direction: FrameDirection):
             await self._handle_llm_start(frame)
         elif isinstance(frame, LLMFullResponseEndFrame):
             await self._handle_llm_end(frame)
-        elif isinstance(frame, TextFrame):
+        elif isinstance(frame, LLMTextFrame):
```
markbackman, line 855: It's intentionally aggregating the `TextFrame` here. This is important because many different `TextFrame` subclasses make it here, e.g. `LLMTextFrame`, `TTSTextFrame`, etc.

markbackman: I've asked this developer to indicate what his Pipeline looks like. That will be an important clue.

lukepayyapilli: that makes sense - thanks @markbackman!

filipi87:
Hi @lukepayyapilli, thanks for this PR! The issue it references (#3638) is already closed, so this looks like it may be stale or superseded by other work.

PR #3689, opened 2026-02-09
Use TurnAnalyzerUserTurnStopStrategy as default stop strategy

aconchillo:
## Summary
- Change the default user turn stop strategy from `TranscriptionUserTurnStopStrategy` to `TurnAnalyzerUserTurnStopStrategy` with `LocalSmartTurnAnalyzerV3`
- Reduce `AUDIO_INPUT_TIMEOUT_SECS` from 1.0s to 0.5s and remove its debug log

## Test plan
- [ ] Verify default turn detection works correctly with the new strategy
- [ ] Verify audio input timeout behavior with the reduced timeout

### DIFF
--- a/.github/workflows/coverage.yaml
+++ b/.github/workflows/coverage.yaml
@@ -39,6 +39,7 @@ jobs:
             --extra google \
             --extra langchain \
             --extra livekit \
+            --extra local-smart-turn-v3 \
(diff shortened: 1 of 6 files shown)


markbackman (approved):
LGTM!

PR #3692, opened 2026-02-09
Rename RequestMetadataFrame to ServiceSwitcherRequestMetadataFrame with service targeting

markbackman:
## Summary
- Renamed `RequestMetadataFrame` to `ServiceSwitcherRequestMetadataFrame` and added a `service` field to target a specific service
- `ServiceSwitcher.push_frame` now only consumes the frame when the targeted service matches the active service (instead of blocking all instances)
- `STTService` and test mocks now push the frame downstream after handling instead of silently consuming it

## Test plan
- [x] All 12 existing `test_service_switcher.py` tests pass
- [x] Ruff lint and format checks pass
- [x] Grepped for any remaining references to old `RequestMetadataFrame` name — none found

### DIFF
--- a/src/pipecat/pipeline/service_switcher.py
+++ b/src/pipecat/pipeline/service_switcher.py
@@ -12,9 +12,9 @@ from typing import Any, Generic, List, Optional, Type, TypeVar
 from pipecat.frames.frames import (
     Frame,
     ManuallySwitchServiceFrame,
-    RequestMetadataFrame,
     ServiceMetadataFrame,
     ServiceSwitcherFrame,
+    ServiceSwitcherRequestMetadataFrame,
 )
 from pipecat.pipeline.parallel_pipeline import ParallelPipeline
 from pipecat.processors.filters.function_filter import FunctionFilter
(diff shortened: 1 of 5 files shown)


src/pipecat/pipeline/service_switcher.py
```diff
@@ -220,16 +220,19 @@ async def filter(_: Frame) -> bool:
...
+        generate `ServiceMetadataFrame`.
 
         """
-        # Don't let RequestMetadataFrame out.
-        if isinstance(frame, RequestMetadataFrame):
-            return
+        # Consume ServiceSwitcherRequestMetadataFrame once the targeted service
+        # has handled it (i.e. the active service).
+        if isinstance(frame, ServiceSwitcherRequestMetadataFrame):
+            if frame.service == self.strategy.active_service:
```
aconchillo, line 234: Good catch. I guess this will allow requesting metadata for switchers down the pipeline.

aconchillo (approved):

aconchillo:
LGTM!

src/pipecat/pipeline/service_switcher.py
kompfner: 👍

kompfner (commented):
👍

PR #3696, opened 2026-02-10
Improve streaming TTS input support, add TextAggregationMetricsData

markbackman:
## Context

This PR has two separate changes that are both part of my goal to speed up the Pipeline and make everything measurable.

First, the `TextAggregationMetricsData`. In looking at where Pipeline time is spent, I can see that a meaningful amount of time is spent waiting for tokens to stream from the LLM to aggregate. This is a function of how long the sentence is and how fast the LLM streams tokens. Short sentences can be <100ms, but on average, sentences can take around 200ms to aggregate. This is a large enough time period that we should log it as a metrics frame.

Second, STT services now offer modes to stream tokens directly from LLMs. This helps to avoid aggregation time. While you can turn on this mode for every TTS service, not all provide consistent audio quality. `CartesiaTTSService` is an example that does this very well! Since that's a popular service, we want to make it easier to use `CartesiaTTSService` in streaming mode.

## Summary

- Added `TextAggregationMetricsData` metric measuring time from first LLM token to first complete sentence, representing the latency cost of sentence aggregation in the TTS pipeline
- Added `text_aggregation_mode` parameter to `TTSService` and all TTS subclasses, with a new `TextAggregationMode` enum (`SENTENCE`, `TOKEN`) as the user-facing API
- Deprecated `aggregate_sentences` boolean in favor of `text_aggregation_mode`
- Moved aggregation mode into text aggregators so all text flows through them regardless of mode — enables pattern detection (`PatternPairAggregator`) and tag handling (`SkipTagsAggregator`) to work in TOKEN mode
- Merged `TTSService._process_text_frame()` into a single codepath instead of separate sentence/token branches
- Reduced debug log spam when streaming tokens: accumulates text and logs once per response instead of per-token
- Skips per-token processing metrics in TOKEN mode (TTFB already captures meaningful timing)

## Breaking Changes

- `aggregate_sentences` parameter is deprecated. Use `text_aggregation_mode=TextAggregationMode.SENTENCE` or `text_aggregation_mode=TextAggregationMode.TOKEN` instead. The old parameter still works but emits a `DeprecationWarning`.

## Testing

- `uv run pytest tests/test_simple_text_aggregator.py tests/test_skip_tags_aggregator.py tests/test_pattern_pair_aggregator.py` — TOKEN mode tests for all three aggregator types
- `uv run pytest` — full test suite passes

### DIFF
--- a/changelog/3696.changed.md
+++ b/changelog/3696.changed.md
@@ -1,0 +1,1 @@
+- Added `text_aggregation_mode` parameter to `TTSService` and all TTS subclasses with a new `TextAggregationMode` enum (`SENTENCE`, `TOKEN`). All text now flows through text aggregators regardless of mode, enabling pattern detection and tag handling in TOKEN mode.
--- a/examples/foundational/07-interruptible.py
+++ b/examples/foundational/07-interruptible.py
@@ -24,6 +24,7 @@ from pipecat.runner.utils import create_transport
 from pipecat.services.cartesia.tts import CartesiaTTSService
 from pipecat.services.deepgram.stt import DeepgramSTTService
 from pipecat.services.openai.llm import OpenAILLMService
+from pipecat.services.tts_service import TextAggregationMode
 from pipecat.transports.base_transport import BaseTransport, TransportParams
 from pipecat.transports.daily.transport import DailyParams
 from pipecat.transports.websocket.fastapi import FastAPIWebsocketParams
@@ -56,6 +57,7 @@ async def run_bot(transport: BaseTransport, runner_args: RunnerArguments):
     tts = CartesiaTTSService(
         api_key=os.getenv("CARTESIA_API_KEY"),
         voice_id="71a7ad14-091c-4e8e-a314-022ece01c121",  # British Reading Lady
+        text_aggregation_mode=TextAggregationMode.TOKEN,
     )
 
     llm = OpenAILLMService(api_key=os.getenv("OPENAI_API_KEY"))
--- a/examples/foundational/07-interruptible.py
+++ b/examples/foundational/07-interruptible.py
@@ -57,7 +57,9 @@ async def run_bot(transport: BaseTransport, runner_args: RunnerArguments):
     tts = CartesiaTTSService(
         api_key=os.getenv("CARTESIA_API_KEY"),
         voice_id="71a7ad14-091c-4e8e-a314-022ece01c121",  # British Reading Lady
-        text_aggregation_mode=TextAggregationMode.TOKEN,
+        # Alternatively, you can use TextAggregationMode.TOKEN to stream tokens instead of
+        # sentencesfor faster response times.
+        # text_aggregation_mode=TextAggregationMode.TOKEN,
     )
 
     llm = OpenAILLMService(api_key=os.getenv("OPENAI_API_KEY"))
--- a/src/pipecat/processors/frame_processor.py
+++ b/src/pipecat/processors/frame_processor.py
@@ -485,6 +485,18 @@ class FrameProcessor(BaseObject):
             if frame:
                 await self.push_frame(frame)
 
+    async def start_text_aggregation_metrics(self):
+        """Start text aggregation time metrics collection."""
+        if self.can_generate_metrics() and self.metrics_enabled:
+            await self._metrics.start_text_aggregation_metrics()
+
+    async def stop_text_aggregation_metrics(self):
+        """Stop text aggregation time metrics collection and push results."""
+        if self.can_generate_metrics() and self.metrics_enabled:
+            frame = await self._metrics.stop_text_aggregation_metrics()
+            if frame:
+                await self.push_frame(frame)
+
     async def stop_all_metrics(self):
         """Stop all active metrics collection."""
         await self.stop_ttfb_metrics()
--- a/src/pipecat/processors/frame_processor.py
+++ b/src/pipecat/processors/frame_processor.py
@@ -501,6 +501,7 @@ class FrameProcessor(BaseObject):
         """Stop all active metrics collection."""
         await self.stop_ttfb_metrics()
         await self.stop_processing_metrics()
+        await self.stop_text_aggregation_metrics()
 
     def create_task(self, coroutine: Coroutine, name: Optional[str] = None) -> asyncio.Task:
         """Create a new task managed by this processor.
--- a/src/pipecat/processors/metrics/frame_processor_metrics.py
+++ b/src/pipecat/processors/metrics/frame_processor_metrics.py
@@ -17,6 +17,7 @@ from pipecat.metrics.metrics import (
     LLMUsageMetricsData,
     MetricsData,
     ProcessingMetricsData,
+    TextAggregationMetricsData,
     TTFBMetricsData,
     TTSUsageMetricsData,
 )
@@ -211,3 +212,27 @@ class FrameProcessorMetrics(BaseObject):
         )
         logger.debug(f"{self._processor_name()} usage characters: {characters.value}")
         return MetricsFrame(data=[characters])
+
+    async def start_text_aggregation_metrics(self):
+        """Start measuring text aggregation time (first token to first sentence)."""
+        self._start_text_aggregation_time = time.time()
+
+    async def stop_text_aggregation_metrics(self):
+        """Stop text aggregation measurement and generate metrics frame.
+
+        Returns:
+            MetricsFrame containing text aggregation time, or None if not measuring.
+        """
+        if (
+            not hasattr(self, "_start_text_aggregation_time")
+            or self._start_text_aggregation_time == 0
+        ):
+            return None
+
+        value = time.time() - self._start_text_aggregation_time
+        logger.debug(f"{self._processor_name()} text aggregation time: {value}")
+        aggregation = TextAggregationMetricsData(
+            processor=self._processor_name(), value=value, model=self._model_name()
+        )
+        self._start_text_aggregation_time = 0
+        return MetricsFrame(data=[aggregation])
--- a/src/pipecat/processors/metrics/frame_processor_metrics.py
+++ b/src/pipecat/processors/metrics/frame_processor_metrics.py
@@ -44,6 +44,7 @@ class FrameProcessorMetrics(BaseObject):
         self._task_manager = None
         self._start_ttfb_time = 0
         self._start_processing_time = 0
+        self._start_text_aggregation_time = 0
         self._last_ttfb_time = 0
         self._should_report_ttfb = True
 
@@ -223,10 +224,7 @@ class FrameProcessorMetrics(BaseObject):
         Returns:
             MetricsFrame containing text aggregation time, or None if not measuring.
         """
-        if (
-            not hasattr(self, "_start_text_aggregation_time")
-            or self._start_text_aggregation_time == 0
-        ):
+        if self._start_text_aggregation_time == 0:
             return None
 
         value = time.time() - self._start_text_aggregation_time
--- a/src/pipecat/services/asyncai/tts.py
+++ b/src/pipecat/services/asyncai/tts.py
@@ -28,7 +28,7 @@ from pipecat.frames.frames import (
 )
 from pipecat.processors.frame_processor import FrameDirection
 from pipecat.services.settings import NOT_GIVEN, TTSSettings, _NotGiven
-from pipecat.services.tts_service import AudioContextTTSService, TTSService
+from pipecat.services.tts_service import AudioContextTTSService, TextAggregationMode, TTSService
 from pipecat.transcriptions.language import Language, resolve_language
 from pipecat.utils.tracing.service_decorators import traced_tts
 
@@ -128,7 +128,8 @@ class AsyncAITTSService(AudioContextTTSService):
         encoding: str = "pcm_s16le",
         container: str = "raw",
         params: Optional[InputParams] = None,
-        aggregate_sentences: Optional[bool] = True,
+        aggregate_sentences: Optional[bool] = None,
+        text_aggregation_mode: Optional[TextAggregationMode] = None,
         **kwargs,
     ):
         """Initialize the Async TTS service.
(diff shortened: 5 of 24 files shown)


src/pipecat/services/asyncai/tts.py
```diff
@@ -115,11 +116,17 @@ def __init__(
             encoding: Audio encoding format.
             container: Audio container format.
             params: Additional input parameters for voice customization.
-            aggregate_sentences: Whether to aggregate sentences within the TTSService.
+            aggregate_sentences: Deprecated. Use text_aggregation_mode instead.
+
+                .. deprecated:: 0.0.102
```
markbackman: Revisit when this actually ships. It'll likely be 0.0.103.

changelog/3696.changed.md
```diff
@@ -0,0 +1 @@
+- Added `text_aggregation_mode` parameter to `TTSService` and all TTS subclasses with a new `TextAggregationMode` enum (`SENTENCE`, `TOKEN`). All text now flows through text aggregators regardless of mode, enabling pattern detection and tag handling in TOKEN mode.
```
filipi87, line 1: This makes things a lot more clear.

examples/foundational/07-interruptible.py
```diff
@@ -56,6 +57,7 @@ async def run_bot(transport: BaseTransport, runner_args: RunnerArguments):
     tts = CartesiaTTSService(
         api_key=os.getenv("CARTESIA_API_KEY"),
         voice_id="71a7ad14-091c-4e8e-a314-022ece01c121",  # British Reading Lady
+        text_aggregation_mode=TextAggregationMode.TOKEN,
```
filipi87: Cool. Would it be worth making this the default mode for Cartesia? Or could this be a breaking change somehow?

src/pipecat/processors/metrics/frame_processor_metrics.py
```diff
@@ -211,3 +212,27 @@ async def start_tts_usage_metrics(self, text: str):
         )
         logger.debug(f"{self._processor_name()} usage characters: {characters.value}")
         return MetricsFrame(data=[characters])
+
+    async def start_text_aggregation_metrics(self):
+        """Start measuring text aggregation time (first token to first sentence)."""
+        self._start_text_aggregation_time = time.time()
```
filipi87, line 219: Should we also initialized (or define) in `__init__` ?

src/pipecat/processors/metrics/frame_processor_metrics.py
```diff
@@ -211,3 +212,27 @@ async def start_tts_usage_metrics(self, text: str):
...
+
+    async def stop_text_aggregation_metrics(self):
+        """Stop text aggregation measurement and generate metrics frame.
+
+        Returns:
+            MetricsFrame containing text aggregation time, or None if not measuring.
+        """
+        if (
+            not hasattr(self, "_start_text_aggregation_time")
```
filipi87: If we initialize there (in __init__), I believe we don't need this. Basically the same as we do for the other metrics.

src/pipecat/processors/frame_processor.py
```diff
@@ -485,6 +485,18 @@ async def start_tts_usage_metrics(self, text: str):
...
+            await self._metrics.start_text_aggregation_metrics()
+
+    async def stop_text_aggregation_metrics(self):
+        """Stop text aggregation time metrics collection and push results."""
+        if self.can_generate_metrics() and self.metrics_enabled:
+            frame = await self._metrics.stop_text_aggregation_metrics()
+            if frame:
+                await self.push_frame(frame)
+
     async def stop_all_metrics(self):
```
filipi87, line 500: Should this also invoke `stop_text_aggregation_metrics` ?

src/pipecat/services/cartesia/tts.py
```diff
@@ -254,20 +255,27 @@ def __init__(
...
+
             **kwargs: Additional arguments passed to the parent service.
         """
-        # Aggregating sentences still gives cleaner-sounding results and fewer
-        # artifacts than streaming one word at a time. On average, waiting for a
-        # full sentence should only "cost" us 15ms or so with GPT-4o or a Llama
-        # 3 model, and it's worth it for the better audio quality.
+        # By default, we aggregate sentences before sending to TTS. This adds
+        # ~200-300ms of latency per sentence (waiting for the sentence-ending
+        # punctuation token from the LLM). Setting aggregate_sentences=False
```
filipi87: Instead of mentioning `aggregate_sentences`, which is now deprecated, shouldn’t we mention `text_aggregation_mode` instead?

src/pipecat/services/cartesia/tts.py
```diff
@@ -616,7 +629,9 @@ async def run_tts(self, text: str, context_id: str) -> AsyncGenerator[Frame, Non
 
             try:
                 await self._get_websocket().send(msg)
-                await self.start_tts_usage_metrics(text)
+                # Usage metrics are aggregated at flush time when streaming tokens.
+                if not self._is_streaming_tokens:
+                    await self.start_tts_usage_metrics(text)
```
filipi87: Shouldn’t we include this same check in all other TTS services that support it? Or maybe just override `start_tts_usage_metrics` inside `tts_service.py` and do the check there?

src/pipecat/services/elevenlabs/tts.py
```diff
@@ -286,13 +287,19 @@ def __init__(
...
+
             **kwargs: Additional arguments passed to the parent service.
         """
-        # Aggregating sentences still gives cleaner-sounding results and fewer
-        # artifacts than streaming one word at a time. On average, waiting for a
-        # full sentence should only "cost" us 15ms or so with GPT-4o or a Llama
-        # 3 model, and it's worth it for the better audio quality.
+        # By default, we aggregate sentences before sending to TTS. This adds
+        # ~200-300ms of latency per sentence (waiting for the sentence-ending
+        # punctuation token from the LLM). Setting aggregate_sentences=False
```
filipi87: Same comment I made on Cartesia, regarding mentioning `text_aggregation_mode` instead

src/pipecat/services/tts_service.py
```diff
@@ -178,7 +206,33 @@ def __init__(
...
+                    if aggregate_sentences
+                    else TextAggregationMode.TOKEN
+                )
+
+        if text_aggregation_mode is None:
+            text_aggregation_mode = TextAggregationMode.SENTENCE
+
+        self._text_aggregation_mode: TextAggregationMode = text_aggregation_mode
+        # Keep for backward compat with subclasses that read self._aggregate_sentences
+        self._aggregate_sentences: bool = text_aggregation_mode != TextAggregationMode.TOKEN
```
filipi87: I guess that inside Pipecat we don’t have any subclass that reads this. Shouldn’t we remove it? It was supposed to be an internal variable anyway.

src/pipecat/services/tts_service.py
```diff
@@ -465,6 +528,9 @@ async def process_frame(self, frame: Frame, direction: FrameDirection):
             and not isinstance(frame, InterimTranscriptionFrame)
             and not isinstance(frame, TranscriptionFrame)
         ):
+            if not self._is_streaming_tokens and not self._aggregation_logged:
+                await self.start_text_aggregation_metrics()
+                self._aggregation_logged = True
```
filipi87: I guess `self._aggregation_logged` means that we’ve started the aggregation metrics. Is that correct?

Couldn’t we override `start_text_aggregation_metrics` and `stop_text_aggregation_metrics` here and set this variable there? And maybe rename it to something that makes it clearer that we’re using it to handle metrics?

src/pipecat/services/tts_service.py
```diff
@@ -478,8 +544,18 @@ async def process_frame(self, frame: Frame, direction: FrameDirection):
...
+                # If this is the first (and only) sentence, stop the aggregation metric.
+                await self.stop_text_aggregation_metrics()
                 await self._push_tts_frames(AggregatedTextFrame(remaining.text, remaining.type))
 
+            self._aggregation_logged = False
+
+            # Log accumulated streamed text and emit aggregated usage metric.
+            if self._streaming_text_log:
+                logger.debug(f"{self}: Generating TTS [{self._streaming_text_log}]")
+                await self.start_tts_usage_metrics(self._streaming_text_log)
```
filipi87: I know we didn’t introduce this in this PR, but `start_tts_usage_metrics` feels a bit weird since we don’t have a corresponding `stop_tts_usage_metrics`. Should we rename it in a follow up PR?

examples/foundational/07-interruptible.py
markbackman: I think we can make the switch in 1.0. We've made the aggregators all handle this case (e.g. not aggregate) but you can still use them. Since this is a performance change, maybe 1.0 is the opportune time to change. WDYT?

filipi87: Yep, I think it makes sense. 👍

src/pipecat/processors/metrics/frame_processor_metrics.py
markbackman: Yup!

src/pipecat/processors/metrics/frame_processor_metrics.py
markbackman: Good point.

src/pipecat/processors/frame_processor.py
markbackman: Yes, good catch.

src/pipecat/services/cartesia/tts.py
markbackman: Good idea. I'll move this to TTSService.

src/pipecat/services/tts_service.py
markbackman: Correct. Good catch. We're converting to self._text_aggregation_mode, so we don't need this internal var. It's dead code. Removing.

src/pipecat/services/cartesia/tts.py
markbackman: Fixed.

src/pipecat/services/elevenlabs/tts.py
markbackman: Fixed.

src/pipecat/services/tts_service.py
markbackman: Fixed!

src/pipecat/services/tts_service.py
markbackman: Yes, in a follow-up

filipi87 (approved):
LGTM! Nice improvement. 🚀

markbackman (commented):
## Code Review

Found 3 bugs in `src/pipecat/services/tts_service.py`. All relate to the new streaming tokens / metrics functionality.

src/pipecat/services/tts_service.py
```diff
@@ -526,9 +617,17 @@ async def process_frame(self, frame: Frame, direction: FrameDirection):
...
             remaining = await self._text_aggregator.flush()
+            # Stop the aggregation metric (no-op if already stopped on first sentence).
+            await self.stop_text_aggregation_metrics()
             if remaining:
                 await self._push_tts_frames(AggregatedTextFrame(remaining.text, remaining.type))
 
+            # Log accumulated streamed text and emit aggregated usage metric.
+            if self._streamed_text:
+                logger.debug(f"{self}: Generating TTS [{self._streamed_text}]")
+                await self.start_tts_usage_metrics(self._streamed_text)
```
markbackman: **Bug:** `start_tts_usage_metrics` override blocks the flush-time aggregated call.

The override at [line 326](https://github.com/pipecat-ai/pipecat/blob/660ca3f84529855e68b0dc91f64489aebb6fe264/src/pipecat/services/tts_service.py#L325-L329) unconditionally returns early when `_is_streaming_tokens` is `True`. However, the flush-time call here also routes through `self.start_tts_usage_metrics(...)` and gets suppressed by the same guard. This means TTS usage metrics are **never emitted** in TOKEN mode.

The override is correct for suppressing per-token calls from individual TTS service implementations (e.g., ElevenLabs, Cartesia), but the aggregated flush-time call needs to bypass it.

```suggestion
            if self._streamed_text:
                logger.debug(f"{self}: Generating TTS [{self._streamed_text}]")
                await FrameProcessor.start_tts_usage_metrics(self, self._streamed_text)
                self._streamed_text = ""
```

Alternatively, call `super()` from `TTSService` explicitly or add a `_force` parameter to distinguish flush-time from per-token calls.

src/pipecat/services/tts_service.py
```diff
@@ -240,6 +294,8 @@ def __init__(
 
         self._processing_text: bool = False
         self._tts_contexts: Dict[str, TTSContext] = {}
+        self._streamed_text: str = ""
+        self._text_aggregation_metrics_started: bool = False
```
markbackman, line 298: **Bug:** `_streamed_text` and `_text_aggregation_metrics_started` not reset on interruption.

`_handle_interruption` (line 773) resets `_processing_text` and `_llm_response_started` but does not reset these two new fields:

1. **`_streamed_text`**: If an interruption occurs mid-response in TOKEN mode, the accumulated text is not cleared. On the next LLM response, new tokens get appended to the stale text, producing incorrect debug logs and inflated TTS usage metrics at flush time.

2. **`_text_aggregation_metrics_started`**: If an interruption occurs after `start_text_aggregation_metrics()` but before `stop_text_aggregation_metrics()`, the flag stays `True`. On the next response, `start_text_aggregation_metrics()` sees the flag is already `True` and skips, so text aggregation timing is not collected.

**Fix:** Add these two lines to [`_handle_interruption`](https://github.com/pipecat-ai/pipecat/blob/660ca3f84529855e68b0dc91f64489aebb6fe264/src/pipecat/services/tts_service.py#L773-L781), after `self._llm_response_started = False` (line 779):

```python
self._streamed_text = ""
self._text_aggregation_metrics_started = False
```

markbackman: I think this is inconsequential, but logically it makes sense. I'm adding the fix.

src/pipecat/services/tts_service.py
markbackman: Nice catch! This issue meant that character usage data wasn't emitted. Now it is with the fix.

PR #3697, opened 2026-02-10
Update SonioxSTTService default model to stt-rt-v4

markbackman:

See the question about the default config for Soniox.

Update:
Also, updating the `vad_force_turn_endpoint` arg to `True` to better align with Pipecat norms.


examples/foundational/07za-interruptible-soniox.py
```diff
@@ -52,6 +53,11 @@ async def run_bot(transport: BaseTransport, runner_args: RunnerArguments):
 
     stt = SonioxSTTService(
         api_key=os.getenv("SONIOX_API_KEY"),
+        vad_force_turn_endpoint=True,
```
markbackman: This configuration was used for the STT benchmark. It's very, very fast. But, it relies on an external smart-turn model, which matches Pipecat patterns.

Soniox has it's own internal smart-turn capability built into the model. But, it doesn't have the same end of turn events to be a turn controller. This makes it a bit of an outlier.

Given this, what is the best configuration for this service. 🤔 

If you're using the internal smart-turn, you probably have to write your own strategy that just pushes the transcript immediately once you get the end of transcript stream token. We don't have one of these out of the box.

Perhaps that means we should change the defaults to match Pipecat's strengths, which is external smart-turn model and/or timeout based end of turn.

Thoughts?

filipi87: > Perhaps that means we should change the defaults to match Pipecat's strengths, which is external smart-turn model and/or timeout based end of turn.

I think this makes sense. By default, we should make the service as optimized as possible within Pipecat. 👍

filipi87 (approved):
LGTM!

examples/foundational/07za-interruptible-soniox.py
markbackman: Yes, agree. I just made this change and added a new changelog entry explaining the decision.

PR #2654, opened 2025-09-15
transports: on_client_disconnected only if remote client disconnects

aconchillo:
#### Please describe the changes in your PR. If it is addressing an issue, please reference that as well.

With this PR we unify `on_client_disconnected` behavior among all transports. `on_client_disconnected` is only triggered if the remote client actually disconnects. If the bot disconnects (e.g. `EndFrame` or `CancelFrame`) `on_client_disconnected` will *not* be called.

PR #3712, opened 2026-02-11
fix: on_client_disconnected not firing when send exception precedes receive exit

### LINKED ISSUE #3710: Potential bug in FastAPIWebsocketClient
### pipecat version

latest

### Python version

NA

### Operating System

NA

### Issue description

Hello,

When there is a send exception in `FastAPIWebsocketClient`, we set `_is_closing` as True. https://github.com/pipecat-ai/pipecat/blob/0574167fbddcf00dfded13c11c7146c3e95eb94b/src/pipecat/transports/websocket/fastapi.py#L162

This means, when there is a receive exception in _receive_messages https://github.com/pipecat-ai/pipecat/blob/0574167fbddcf00dfded13c11c7146c3e95eb94b/src/pipecat/transports/websocket/fastapi.py#L324, we wont trigger on_client_disconnected. 

So, if there is a user hangup, and we receive send exception before receive exception, we will not trigger on_client_disconnected. 

Also, is_closing was supposed to mean that we have closed the pipeline, which is being set in disconnect() method. 
https://github.com/pipecat-ai/pipecat/blob/0574167fbddcf00dfded13c11c7146c3e95eb94b/src/pipecat/transports/websocket/fastapi.py#L171

IMO, this is a bug.

### Reproduction steps

NA

### Expected behavior

NA

### Actual behavior

NA

### Logs

```shell

```

lukepayyapilli:
## Summary
- [x] Removed `self._closing = True` from `FastAPIWebsocketClient.send()` exception handler so `_closing` only gets set by intentional `disconnect()` calls.
- [x] Removed the `and not self.is_closing` guard which only existed to avoid setting the flag twice.
- [x] Added tests verifying `_closing` flag behavior for both `send()` exceptions and `disconnect()`.

Closes #3710

## Approach

When a client hangs up, a send exception can fire before the receive loop exits. Previously, the `send()` exception handler set `self._closing = True` when it detected a `DISCONNECTED` application state. This caused `_receive_messages()` to skip calling `on_client_disconnected`, since it checks `if not self._client.is_closing`.

The fix distinguishes between two meanings of `_closing`:
- **Intentional disconnect** (set in `disconnect()`): we initiated the close, so the receive loop should not fire `on_client_disconnected`.
- **Observed connection death** (previously set in `send()`): the connection is already gone, but the receive loop should still notify via `on_client_disconnected`.

The `_closing = True` in `send()` was redundant because every downstream consumer (`_can_send()`, `write_audio_frame()`, `_write_frame()`) already independently checks `is_connected`, so removing it does not allow stale sends. The only behavioral change is that the receive loop now correctly fires `on_client_disconnected` in this race condition.

### DIFF
--- a/changelog/3710.fixed.md
+++ b/changelog/3710.fixed.md
@@ -1,0 +1,1 @@
+- Fixed `FastAPIWebsocketTransport` issue where `on_client_disconnected` was not triggered when a send exception occurred before the receive loop exited.
--- a/src/pipecat/transports/websocket/fastapi.py
+++ b/src/pipecat/transports/websocket/fastapi.py
@@ -154,12 +154,8 @@ class FastAPIWebsocketClient:
             )
             # For some reason the websocket is disconnected, and we are not able to send data
             # So let's properly handle it and disconnect the transport if it is not already disconnecting
-            if (
-                self._websocket.application_state == WebSocketState.DISCONNECTED
-                and not self.is_closing
-            ):
+            if self._websocket.application_state == WebSocketState.DISCONNECTED:
                 logger.warning("Closing already disconnected websocket!")
-                self._closing = True
 
     async def disconnect(self):
         """Disconnect the WebSocket client."""
--- a/tests/test_fastapi_websocket.py
+++ b/tests/test_fastapi_websocket.py
@@ -5,9 +5,17 @@
 #
 
 import unittest
-from unittest.mock import AsyncMock
+from unittest.mock import AsyncMock, MagicMock
 
-from pipecat.transports.websocket.fastapi import _WebSocketMessageIterator
+import pytest
+from starlette.websockets import WebSocketState
+
+from pipecat.frames.frames import StartFrame
+from pipecat.transports.websocket.fastapi import (
+    FastAPIWebsocketCallbacks,
+    FastAPIWebsocketClient,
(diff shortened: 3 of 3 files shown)


filipi87 (commented):
Hi @lukepayyapilli,

The current behavior is intentional. We decided to trigger `on_client_disconnected` only when we receive confirmation from the client that it has disconnected.
- https://github.com/pipecat-ai/pipecat/pull/2654

As you can see, inside FastAPI we check two different WebSocket states:
- `application_state`: used to determine whether the bot has accepted or closed the connection.
- `client_state`: used to determine whether the client is still connected.

So, if an exception occurs during a send operation and the `application_state` is `DISCONNECTED`, it means something inside Pipecat caused the WebSocket to close. However, the `client_state` may still be `CONNECTED`, and attempting to disconnect again in this case could cause side effects.

If you want an event that is always triggered when the pipeline finishes, you should use instead:

```
@task.event_handler("on_pipeline_finished")
async def on_pipeline_finished(task, frame):
```

You can use this event for cleanup, logging, or post processing tasks.

aconchillo:
We fixed an issue regarding `on_client_disconnected`: https://github.com/pipecat-ai/pipecat/pull/4186.

Closing. Please, re-open or create a new ticket if you still have issues.

PR #3713, opened 2026-02-11
Fixing smallwebrtc transport input audio resampling logic.

filipi87:
## Summary                                                                                                                     
                                                                                                                                 
  - Fixed SmallWebRTC transport input audio resampling to properly handle all sample rates                                       
  - Previously hardcoded resampler to 16kHz, which failed to resample 8kHz input audio (common in telephony)                   
  - Now configures resampler with the actual target sample rate during transport setup                                           
  - Simplified resampling logic to handle both upsampling and downsampling correctly                                             
                                                                                                                                 
  ## Changes                                                                                                                     
                                                                                                                                 
  - Removed hardcoded `_pipecat_resampler` that always targeted 16000 Hz                                                         
  - Added `_audio_in_resampler` configured with `self._in_sample_rate` during setup                                              
  - Fixed condition to resample whenever `frame.sample_rate != self._in_sample_rate`
  - Unified code path for both resampled and non-resampled frames

### DIFF
--- a/src/pipecat/transports/smallwebrtc/transport.py
+++ b/src/pipecat/transports/smallwebrtc/transport.py
@@ -233,9 +233,8 @@ class SmallWebRTCClient:
         self._out_sample_rate = None
         self._leave_counter = 0
 
-        # We are always resampling it for 16000 if the sample_rate that we receive is bigger than that.
-        # otherwise we face issues with Silero VAD
-        self._pipecat_resampler = AudioResampler("s16", "mono", 16000)
+        # Audio resampler - will be configured during setup with target sample rate
+        self._audio_in_resampler = None
 
         @self._webrtc_connection.event_handler("connected")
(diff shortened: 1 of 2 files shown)


src/pipecat/transports/smallwebrtc/transport.py
```diff
@@ -233,9 +233,8 @@ def __init__(self, webrtc_connection: SmallWebRTCConnection, callbacks: SmallWeb
         self._out_sample_rate = None
         self._leave_counter = 0
 
-        # We are always resampling it for 16000 if the sample_rate that we receive is bigger than that.
-        # otherwise we face issues with Silero VAD
-        self._pipecat_resampler = AudioResampler("s16", "mono", 16000)
```
filipi87, line 238: This was basically the issue, we were always converting it to 16khz.

markbackman (approved):
LGTM!

PR #3714, opened 2026-02-11
Broad refactor of service settings and how they’re updated at runtime

kompfner:
Maybe the best way to understand these changes is to check out the [COMMUNITY_INTEGRATIONS.md changes](https://github.com/pipecat-ai/pipecat/pull/3714/changes#diff-99229f900513f524670cd58fb71049ac712c3addab6987ebf443046b8aef8291).

Does not (yet) touch `InputParams`, to avoid scope creep, but what to do with those is an open question. `InputParams` represents what you can specify at init time, and `*Settings` represents what can (theoretically) be updated at runtime. These often overlap _heavily_. In the pattern introduced in this PR, `self._settings` is the source of truth, and the init-time `InputParams` are used just to seed `self._settings` (which was _kind of_ the pattern before, but never really formalized or adhered to especially strictly)

NOTE: manual testing of every service in progress, there may be some more changes that creep in.


src/pipecat/services/asyncai/tts.py
```diff
@@ -143,7 +143,7 @@ def __init__(
         }
 
         self.set_model_name(model)
-        self.set_voice(voice_id)
+        self._voice_id = voice_id
```
kompfner: Previously it wasn't clear whether `set_voice`/`set_model`/`set_language` were meant to be public APIs that users could directly invoke—in which case they should handle applying the change by reconnecting if necessary, etc.—or whether they were meant to only be "bookkeeping" methods, invokable only by subclasses, to update the underlying instance variables (`_voice_id`, etc).

I've made the decision here to make `set_voice`/`set_model`/`set_language` the "public" methods (even though we probably would suggest that users to use `*UpdateSettingsFrame`s instead) and if we just need to update bookkeeping, access the instance variable directly.

kompfner: We do use `set_voice()` directly in one example.

src/pipecat/services/google/stt.py
```diff
@@ -355,6 +358,44 @@ def language_to_google_stt_language(language: Language) -> Optional[str]:
...
+class GoogleSTTSettings(STTSettings):
+    """Typed settings for Google Cloud Speech-to-Text V2.
+
+    Parameters:
+        languages: List of ``Language`` enums for recognition
+            (e.g. ``[Language.EN_US]``). Preferred over ``language_codes``.
+        language_codes: List of Google STT language code strings
+            (e.g. ``["en-US"]``).
+
+            .. deprecated:: 0.0.103
```
kompfner, line 371: TODO: update this (and all other occurrences of this) if it looks like these changes will land in a different version

src/pipecat/services/settings.py
```diff
@@ -0,0 +1,313 @@
...
+        extra: Overflow dict for service-specific keys that don't map to a
+            declared field.
+    """
+
+    # -- common fields -------------------------------------------------------
+
+    model: Any = field(default_factory=lambda: NOT_GIVEN)
+    """AI model identifier (e.g. ``"gpt-4o"``, ``"eleven_turbo_v2_5"``)."""
+
+    extra: Dict[str, Any] = field(default_factory=dict)
```
kompfner, line 120: Note that this PR doesn't go through every service and add handling for the `extra` field. It will be up to each service to add it when it's a priority. This can be done piecemeal.

src/pipecat/services/deepgram/stt.py
```diff
@@ -185,37 +195,57 @@ def can_generate_metrics(self) -> bool:
...
-    async def set_model(self, model: str):
-        """Set the Deepgram model and reconnect.
+    async def _update_settings(self, update: STTSettings) -> set[str]:
+        """Apply a settings update, keeping ``live_options`` in sync.
 
-        Args:
-            model: The Deepgram model name to use.
+        Top-level ``model`` and ``language`` are the source of truth.  When
+        they are given in *update* their values are propagated into
+        ``live_options``.  When only ``live_options`` is given, its ``model``
```
markbackman, line 214: Developers will want to change more than just model and language. Is it possible to make any live_options change apply, as well?

kompfner: That's already the case. Model and language are just given some special treatment in terms of bookkeeping

kompfner: Note, though, that before this PR deepgram didn't actually handle runtime changes to live_options

kompfner: And, actually...I'm working on a commit that will stub out most of the changes in this PR that add _new_ runtime handling of settings updates, as I think it's way too much testing scope for this PR. The new handling will be replaced with a warning to the effect of "unhandled settings updates" and a commented-out code stub. Then in future PRs we can add proper handling of settings updates.

kompfner: Note that I *will* keep the reconnect logic in this service, though, since this service already handled reconnecting on language and model change (but, as previously mentioned, not other live_options)

src/pipecat/services/settings.py
```diff
@@ -0,0 +1,319 @@
...
+        model: TTS model identifier.
+        voice: Voice identifier or name.
+        language: Language for speech synthesis. The union type reflects the
+            *input* side: callers may pass a ``Language`` enum or a raw string.
+            However, the **stored** value is always a service-specific string
+            — ``TTSService._update_settings`` converts ``Language`` enums via
+            ``language_to_service_language()`` before writing, and ``__init__``
+            methods do the same at construction time. Code that reads
+            ``self._settings.language`` after initialisation can treat it as
+            ``str``.
```
kompfner, line 306: This input-versus-storage quirk is a bit ugly, but I couldn't find a nice elegant way around it.

The benefits of storing language as the service-specific string (in each service's `self._settings`)  are:
- Caching (we don't have to convert enum to service-specific string JIT at the point of use)
- Less error prone (service implementors don't have to remember to do all the JIT conversions)
- Truer "diffing" when determining whether language has changed (in a way that requires a service reconnection, say)
- Backwards compatibility (we need to support users sending in dict-based—as opposed to class-based—`*UpdateSettingsFrame`s updates that specify language as a service-specific strings, which used to work before); we *could* choose to implement "backwards" mappings from service-specific strings to language enum values, but that would be a big undertaking and a maintenance pain going forward

This all comes from the tension of using these `*Settings` objects both for per-service settings storage and as settings update objects. But the alternative would be introducing a whole slew of new `*SettingsUpdate` types, and that felt not worth it.

kompfner:
Hm...wonder if we should make the `*UpdateSettingsFrame`s uninterruptible? In my testing, if the settings update is queued while the bot is speaking, and then I interrupt the bot, the update never takes effect. As a user, that might not be the desired outcome.

filipi87:
> Hm...wonder if we should make the *UpdateSettingsFrames uninterruptible? In my testing, if the settings update is queued while the bot is speaking, and then I interrupt the bot, the update never takes effect. As a user, that might not be the desired outcome.

I think this makes sense. We did the same for the summarization result. Otherwise, most of the time it wouldn’t be applied.

.claude/skills/cleanup/SKILL.md
```diff
@@ -293,7 +293,7 @@ class NewTTSService(TTSService):
         """
         super().__init__(**kwargs)
         self._api_key = api_key
-        self.set_voice(voice)
+        self._voice_id = voice
```
filipi87: Is this change intentional? I think that most of the time our TTS service invokes `self.set_voice` instead of `self._voice_id`.

filipi87: I see, we have deprecated the `set_voice`. 😅

examples/foundational/35-pattern-pair-voice-switching.py
```diff
@@ -117,7 +117,7 @@ async def on_voice_tag(match: PatternMatch):
             # First flush any existing audio to finish the current context
             await tts.flush_audio()
             # Then set the new voice
-            tts.set_voice(VOICE_IDS[voice_name])
+            await tts.set_voice(VOICE_IDS[voice_name])
```
filipi87, line 120: Should we refactor this example so we don’t keep using something deprecated?

.claude/skills/cleanup/SKILL.md
kompfner: Yes, it's intentional. It was confusing before, being a "public" method. It now works the same way as `set_language`, or `set_model`, in that it updates the setting *and* causes the service to react to the new setting (doing whatever it needs to do, like reconnect). Here we're being more transparent that if all we want to do is update private bookkeeping, we should just set the private variable.

But the fact of the private `_voice_id` living in a parent class *is* rather confusing—maybe in a future pass we can update this pattern?

examples/foundational/35-pattern-pair-voice-switching.py
kompfner: Yes, we should :)

kompfner:
OK, phew, I've gone through every single example file for manual testing.

Still more to do! I think the example files aren't exhaustive enough, there were some services that were missed. And I discovered a few more things along the way...

Tomorrow at our sync we should talk about how to proceed with this PR. If it stays open for much longer, it'll be unmaintainable due to merge conflicts. But we really really don't want to sneak this in before a release, as it's huge.

examples/foundational/35-pattern-pair-voice-switching.py
filipi87, line 1: Would it make sense to include any of these new examples in evals ? I guess this would probably require some changes in the eval itself to support this.

src/pipecat/services/asyncai/tts.py
filipi87: But I see that it is deprecated now, should we keep using it in our example ?

src/pipecat/services/settings.py
filipi87: Makes sense. 👍

.claude/skills/cleanup/SKILL.md
filipi87: Agreed, this is really confusing. 

In the superclass, we have two methods: `set_model_name` and `set_model`. What if we did something similar for the voice id and created a `set_voice_id` method ? 

I also think the most straightforward approach would be to pass `voice_id` when initializing the parent class (`super().__init__`). However, in that case, the same should apply to the model name as well.

src/pipecat/frames/frames.py
```diff
@@ -2115,13 +2113,21 @@ class TTSStoppedFrame(ControlFrame):
 class ServiceUpdateSettingsFrame(ControlFrame):
```
filipi87, line 2121: I think make it `Uninterruptible` as you suggest would make sense.

src/pipecat/frames/frames.py
```diff
@@ -2115,13 +2113,21 @@ class TTSStoppedFrame(ControlFrame):
...
+            .. deprecated:: 0.0.103
+                Use ``update`` with a typed settings object instead.
+
+        update: :class:`~pipecat.services.settings.ServiceSettings` object
+            describing the delta to apply.
     """
 
-    settings: Mapping[str, Any]
+    settings: Mapping[str, Any] = field(default_factory=dict)
+    update: Optional["ServiceSettings"] = None
```
filipi87, line 2138: nit: The name `update` feels a bit generic. Wouldn't it be confusing since the frame itself is called `ServiceUpdateSettingsFrame` ?

What if we called it `new_settings`, `delta_settings` or `service_settings` ?

src/pipecat/services/anthropic/llm.py
```diff
@@ -69,6 +70,31 @@
     raise Exception(f"Missing module: {e}")
 
 
+@dataclass
+class AnthropicLLMSettings(LLMSettings):
```
filipi87, line 93: This is so much better. Great work, Paul! 🔥

src/pipecat/services/assemblyai/stt.py
```diff
@@ -165,6 +184,37 @@ def can_generate_metrics(self) -> bool:
         """
         return True
 
+    async def _update_settings(self, update: STTSettings) -> dict[str, Any]:
```
filipi87, line 189: Just to understand: why are we overriding `_update_settings` and logging a `warning` here, but not doing the same in the Anthropic implementation ?

src/pipecat/services/settings.py
```diff
@@ -0,0 +1,319 @@
...
+            delta = TTSSettings(voice="bob")
+            changed = current.apply_update(delta)
+            # changed == {"voice": "alice"}
+            # current.voice == "bob", current.language == "en"
+        """
+        changed: Dict[str, Any] = {}
+        for f in fields(self):
+            if f.name == "extra":
+                continue
+            new_val = getattr(update, f.name)
```
filipi87, line 183: I believe that `getattr` without a default argument raises `AttributeError` when it doesn't exist. Probably safe do something like this:

```suggestion
            new_val = getattr(update, f.name, NOT_GIVEN)
```

src/pipecat/services/settings.py
```diff
@@ -0,0 +1,319 @@
...
+            if not is_given(new_val):
+                continue
+            old_val = getattr(self, f.name)
```
filipi87, line 186: Maybe the same here:

```suggestion
            old_val = getattr(self, f.name, NOT_GIVEN)
```

src/pipecat/services/asyncai/tts.py
```diff
@@ -132,19 +161,19 @@ def __init__(
...
-                "encoding": encoding,
-                "sample_rate": 0,
-            },
-            "language": self.language_to_service_language(params.language)
+        self._settings = AsyncAITTSSettings(
+            model=model,
+            voice=voice_id,
+            output_container=container,
+            output_encoding=encoding,
+            output_sample_rate=0,
```
filipi87, line 169: Considering that the API expects to receive the nested `output_format`, wouldn’t it be easier to understand to keep the nested structure close to what the service API expects ?

Keep `output_format` inside `AsyncAITTSSettings` instead of flattening the fields. Would it make sense ?

src/pipecat/services/cartesia/tts.py
```diff
@@ -290,22 +329,20 @@ def __init__(
...
         self._cartesia_version = cartesia_version
         self._url = url
-        self._settings = {
-            "output_format": {
-                "container": container,
-                "encoding": encoding,
-                "sample_rate": 0,
-            },
-            "language": self.language_to_service_language(params.language)
+        self._settings = CartesiaTTSSettings(
```
filipi87, line 332: Should we also include the voice here ?
`voice=voice_id,`

And the same for the other TTS services, the settings object doesn't reflect the initial configured voice.

src/pipecat/services/cartesia/tts.py
```diff
@@ -290,22 +329,20 @@ def __init__(
...
-            "pronunciation_dict_id": params.pronunciation_dict_id,
-        }
+            speed=params.speed,
+            emotion=params.emotion,
+            generation_config=params.generation_config,
+            pronunciation_dict_id=params.pronunciation_dict_id,
+        )
         self.set_model_name(model)
-        self.set_voice(voice_id)
+        self._voice_id = voice_id
```
filipi87: @kompfner, I was wondering, since `self._settings` already contains the `model` and `voice`, and all services now have `self._settings`, do we still need to keep the fields `self._voice_id` and `self._model_name` in the services ?

src/pipecat/services/google/tts.py
```diff
@@ -1159,14 +1222,14 @@ def __init__(
         self._location = location
         self._model = model
```
filipi87, line 1231: Settings updates go through `AIService._update_settings` which calls `set_model_name()` (syncing `self._model_name`), but `self._model` is never updated.

Maybe we shouldn't need this `self._model` here and should instead use `self._model_name` when needed ?

src/pipecat/services/hathora/stt.py
filipi87, line 97: > Settings updates go through `AIService._update_settings` which calls `set_model_name()` (syncing self._model_name), but self._model is never updated.

I guess it is the same here ?

src/pipecat/services/hathora/tts.py
filipi87, line 114: > Settings updates go through `AIService._update_settings` which calls `set_model_name()` (syncing self._model_name), but self._model is never updated.

And here. Maybe there are also other services with the same behavior.

filipi87 (commented):
This looks great, @kompfner. 👏

I added a couple of suggestions, but my main question is, since `self._settings` already contains the model and voice, and all services now have `self._settings`, do we still need to keep the fields `self._voice_id` and `self._model_name` in the services ?

Another question is that some services implement `_update_settings` and show a warning, while others do not. What is the criteria for deciding whether a service should implement this method and display the warning ?

src/pipecat/services/ai_service.py
```diff
@@ -41,29 +42,25 @@ def __init__(self, **kwargs):
...
-        self._settings: Dict[str, Any] = {}
+        self._settings: ServiceSettings = ServiceSettings(model="")
         self._session_properties: Dict[str, Any] = {}
-        self._tracing_enabled: bool = False
-        self._tracing_context = None
 
-    @property
-    def model_name(self) -> str:
-        """Get the current model name.
+    def _sync_model_name_to_metrics(self):
```
kompfner, line 50: Got rid of `self._voice_id` and `self._model_name`, as you suggested, @filipi87. Feels like a nice step in the right direction! Now `self._settings.voice` and `self._settings.model` are the source of truth.

See the comment in this method for how things could be cleaned up even more (maybe in a follow-up PR).

src/pipecat/services/ai_service.py
```diff
@@ -41,29 +42,25 @@ def __init__(self, **kwargs):
...
+        We don't store model name here because there's already a single source
+        of truth for it in `self._settings.model`. This method is just for
+        syncing the model name to the metrics data.
 
-    def set_model_name(self, model: str):
-        """Set the AI model name and update metrics.
+        TODO: as a next step we should make it so that service classes pass
+        model into `super().__init__` and `AIService` can be responsible for
+        syncing its initial value to metrics, just as it's responsible for
+        syncing any updates to its value to metrics via `_update_settings`.
```
kompfner, line 60: NOTE: service classes should also pass other superclass-handled settings like `language` and `voice` into `super().__init__`, too, as part of this change.

kompfner:
NOTE: the last merge commit resolving conflicts is NOT finished. Needs some local work.

vmfullyfaltu:
@kompfner @filipi87 
this looks great, would this work with service switcher with 2 different STT services? would you have to push two frames for each or figure out the active service and push to that?

kompfner:
> would this work with service switcher with 2 different STT services?

Great question! I actually think today you'd have to figure out the active service and push to that one, since the service switcher blocks most inbound frames—including settings update frames—to inactive services.

This is a really great consideration, though—I think we can come up with something better than that. We'll tackle this work soon. Thanks!

examples/foundational/55q-update-settings-deepgram-sagemaker-tts.py
kompfner, line 1: Note: this example isn't working for me—it stalls out the pipeline. I assumed I had broken Deepgram Sagemaker TTS in this PR, but when I tested the 07c example in `main`, I'm seeing the same stall. cc @markbackman

vmfullyfaltu:
> > would this work with service switcher with 2 different STT services?
> 
> Great question! I actually think today you'd have to figure out the active service and push to that one, since the service switcher blocks most inbound frames—including settings update frames—to inactive services.
> 
> This is a really great consideration, though—I think we can come up with something better than that. We'll tackle this work soon. Thanks!

@kompfner understood and thank you for your reply. would this work in flows to swap deepgram keyterms as you navigate from node to node through the settings frame?

filipi87 (approved):
Hi Paul, this PR is huge, amazing job. 👏🔥

There are still a couple of comments open that I believe haven’t been addressed yet.

I am not sure whether you’ve already added them to your TODO list to be implemented as follow up PRs, or if they’re not actually needed, but I believe these are the main ones:
- Move `_settings` to the initializer, as Aleix suggested, so we only need to invoke `self._sync_model_name_to_metrics()` inside AiService.
- Make `ServiceUpdateSettingsFrame` uninterruptible.
- The name `update` inside `ServiceUpdateSettingsFrame` feels a bit generic. Would it make more sense to call it `new_settings`, `delta_settings`, or `service_settings`?
- `GeminiTTSService` still holds a reference to `self._model`, which is never updated. I believe the same might be true for `HathoraSTTService` and `HathoraTTSService`.
- Inside `ServiceSettings.apply_update`, I added two comments suggesting defaulting to `NOT_GIVEN`. It might be worth applying that change.

Other considerations:
- Inside `AsyncAITTSSettings`, since the API expects a nested `output_format`, would it be clearer to keep that nested structure closer to what the service API expects ? For example, keeping `output_format` inside `AsyncAITTSSettings` instead of flattening the fields. Maybe the same apply for other services.

In any case, I am leaving my approval, as I agree we should merge this as soon as possible. Any improvements can be implemented in follow up PRs, which should be much easier to maintain. 

Again, great work. 🔥🚀

src/pipecat/services/soniox/stt.py
```diff
@@ -134,6 +136,17 @@ def _prepare_language_hints(
...
 
+@dataclass
+class SonioxSTTSettings(STTSettings):
+    """Settings for Soniox STT service.
+
+    Parameters:
+        input_params: Soniox ``SonioxInputParams`` for detailed configuration.
+    """
+
+    input_params: SonioxInputParams | _NotGiven = field(default_factory=lambda: NOT_GIVEN)
```
kompfner, line 147: TODO: this doesn't follow our new pattern. We should include the fields of the input params directly in the settings object. Audit the other services for this mistake, too.

src/pipecat/services/deepgram/stt.py
```diff
@@ -139,12 +154,18 @@ def __init__(
         if "language" in merged_options and isinstance(merged_options["language"], Language):
             merged_options["language"] = merged_options["language"].value
 
-        self.set_model_name(merged_options["model"])
-        self._settings = merged_options
+        merged_live_options = LiveOptions(**merged_options)
+        self._settings = DeepgramSTTSettings(
```
kompfner, line 158: TODO: previously looks like we stored all the `LiveOptions` fields directly in `self._settings`, meaning that users could specify those fields directly in the update dictionary, not nested under `live_options`. Fix this broken backward compatibility (use `from_mapping`), and audit other services for the same kind of mistake. To audit, maybe look for:
- other uses of `LiveOptions`
- `to_dict()` in service initializers
- `*Settings` objects that contain any 3rd-party SDK objects in their fields

kompfner:
Thanks @filipi87! Will go ahead and merge just to make conflicts more manageable, but will open various follow-up PRs in a hopefully more targeted and rapid-fire manner.

kompfner:
> The name update inside ServiceUpdateSettingsFrame feels a bit generic. Would it make more sense to call it new_settings, delta_settings, or service_settings?

Renamed in https://github.com/pipecat-ai/pipecat/pull/3812 @filipi87

src/pipecat/services/soniox/stt.py
kompfner: Found that Soniox STT and Gladia STT weren't following the new pattern. Updating in https://github.com/pipecat-ai/pipecat/pull/3812.

src/pipecat/services/ai_service.py
kompfner: @aconchillo had the better idea (that @filipi87 reminded me of, too) of just passing settings wholesale into `super().__init__()`. Will explore that approach.

src/pipecat/services/asyncai/tts.py
kompfner: Hmm...actually, in the example (example 35), we kind of need the voice change to take effect immediately. It can't be queued. The voice switches occur partway through the TTS service processing an LLM response. So using an `STTUpdateSettingsFrame` is not really a drop-in replacement for `set_voice()`...

But also...that example doesn't quite work? The timing of the voice switches is always wrong. I've confirmed that this was the case before the refactor, too.

With these things in mind, I'm going to punt on updating that example for now.

src/pipecat/services/deepgram/stt.py
kompfner: @markbackman [this new PR](https://github.com/pipecat-ai/pipecat/pull/3817) adds back users' ability to request granular field-level updates to `LiveOptions` settings without wholesale replacing `LiveOptions`. I had broken backward compatibility with this PR.

src/pipecat/services/settings.py
kompfner: [This follow-up PR](https://github.com/pipecat-ai/pipecat/pull/3812) hopefully cleared up this ugliness a bit.

kompfner:
> > Hm...wonder if we should make the *UpdateSettingsFrames uninterruptible? In my testing, if the settings update is queued while the bot is speaking, and then I interrupt the bot, the update never takes effect. As a user, that might not be the desired outcome.
> 
> I think this makes sense. We did the same for the summarization result. Otherwise, most of the time it wouldn’t be applied.

https://github.com/pipecat-ai/pipecat/pull/3819

examples/foundational/35-pattern-pair-voice-switching.py
kompfner: See my response to this comment: https://github.com/pipecat-ai/pipecat/pull/3714#discussion_r2849742123

examples/foundational/35-pattern-pair-voice-switching.py
kompfner: Exactly. For the moment, leaving it out of evals.

src/pipecat/frames/frames.py
kompfner: https://github.com/pipecat-ai/pipecat/pull/3819

src/pipecat/frames/frames.py
kompfner: It's now `delta` 👍

src/pipecat/services/assemblyai/stt.py
kompfner: We probably already discussed this in person, but: it's because there any updated settings actually _are_ applied. No reconnect is needed there, the settings are just picked up on the next run. This is true for various services.

src/pipecat/services/settings.py
kompfner: Ah is this to protect against the case of trying to `apply_update` say, a generic `TTSSettings` object into a specific `RimeTTSSettings` object? Because if they're actually the same type, I don't think this error can ever occur. Will test.

kompfner: Great catch! Fixing in a PR. Stay tuned.

src/pipecat/services/settings.py
kompfner: No need here, I believe. It's guaranteed, because it's `self`, and `f` comes from looping over `fields(self)`.

src/pipecat/services/asyncai/tts.py
kompfner: I believe it used to be nested, and I flattened it. The reasoning behind flattening:
- I wanted strong types. Either we could introduce a new type, or have a looser dict, which we're trying to move away from.
- I wanted to make it simple to do sparse updates, updating one field without touching the others.

src/pipecat/services/cartesia/tts.py
kompfner: taken care of :)

src/pipecat/services/google/tts.py
kompfner: https://github.com/pipecat-ai/pipecat/pull/3827

src/pipecat/services/hathora/stt.py
kompfner: Also fixed here: https://github.com/pipecat-ai/pipecat/pull/3827

src/pipecat/services/hathora/tts.py
kompfner: Also fixed here: https://github.com/pipecat-ai/pipecat/pull/3827

kompfner:
> > > would this work with service switcher with 2 different STT services?
> > 
> > 
> > Great question! I actually think today you'd have to figure out the active service and push to that one, since the service switcher blocks most inbound frames—including settings update frames—to inactive services.
> > This is a really great consideration, though—I think we can come up with something better than that. We'll tackle this work soon. Thanks!
> 
> @kompfner understood and thank you for your reply. would this work in flows to swap deepgram keyterms as you navigate from node to node through the settings frame?

Though I did not test that specific thing, in theory, yes, the `*UpdateSettingsFrame`s should be useful for exactly things like that. In terms of updating deepgram settings specifically, and updating things in a way that user interruptions shouldn't interfere with, there are a couple of follow-up PRs to this one that may be of interest:
- https://github.com/pipecat-ai/pipecat/pull/3817
- https://github.com/pipecat-ai/pipecat/pull/3819

kompfner:
@filipi87 I think I addressed all the comments except the suggestion of passing settings to `super().__init__` to let `AIService` be the sole handler of syncing model name to metrics, which will be the biggest change. Will tackle tomorrow.

kompfner:
> @filipi87 I think I addressed all the comments except the suggestion of passing settings to `super().__init__` to let `AIService` be the sole handler of syncing model name to metrics, which will be the biggest change. Will tackle tomorrow.

And here it is: https://github.com/pipecat-ai/pipecat/pull/3834

PR #3719, opened 2026-02-11
Add SIP transfer and SIP REFER frames to Daily transport

aconchillo:
## Summary

- Added `DailySIPTransferFrame` and `DailySIPReferFrame` to queue SIP operations with audio, so they execute only after the bot finishes its current utterance
- Added `write_transport_frame()` hook to `BaseOutputTransport` so transport subclasses can handle custom frame types flowing through the audio queue
- Undeprecated `DailyUpdateRemoteParticipantsFrame` and changed it from `ControlFrame` to `DataFrame` so it is also queued with audio via `write_transport_frame`
- Switched error handling in `DailyOutputTransport` to use `push_error` instead of `logger.error`

## Testing

- `uv run ruff check` and `uv run ruff format --check` pass
- `uv run pytest` passes (607/607, 1 pre-existing integration failure)

### DIFF
--- a/src/pipecat/transports/daily/transport.py
+++ b/src/pipecat/transports/daily/transport.py
@@ -15,7 +15,7 @@ import asyncio
 import time
 from concurrent.futures import CancelledError as FuturesCancelledError
 from concurrent.futures import ThreadPoolExecutor
-from dataclasses import dataclass
+from dataclasses import dataclass, field
 from typing import Any, Awaitable, Callable, Dict, Mapping, Optional, Tuple
 
 import aiohttp
@@ -26,6 +26,7 @@ from pipecat.audio.vad.vad_analyzer import VADAnalyzer, VADParams
 from pipecat.frames.frames import (
     CancelFrame,
     ControlFrame,
+    DataFrame,
     EndFrame,
     Frame,
     InputAudioRawFrame,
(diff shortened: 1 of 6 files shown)


src/pipecat/transports/daily/transport.py
```diff
@@ -1968,7 +1966,7 @@ async def send_message(
         """
         error = await self._client.send_message(frame)
         if error:
-            logger.error(f"Unable to send message: {error}")
+            await self.push_error(f"{self}: Unable to send message: {error}")
```
markbackman: `push_error()` takes care of the logging, include `{self}`.

```suggestion
            await self.push_error(error_msg=f"Unable to send message: {error}")
```

src/pipecat/transports/daily/transport.py
```diff
@@ -2011,6 +2009,25 @@ async def write_video_frame(self, frame: OutputImageRawFrame) -> bool:
...
+    async def write_transport_frame(self, frame: Frame):
+        """Handle queued SIP frames after preceding audio has been sent.
+
+        Args:
+            frame: The frame to handle.
+        """
+        if isinstance(frame, DailySIPTransferFrame):
+            error = await self._client.sip_call_transfer(frame.settings)
+            if error:
+                await self.push_error(f"{self}: Unable to transfer SIP call: {error}")
```
markbackman: Same here and below for `push_error`.

markbackman (approved):
LGTM. Just a few small changes to `push_error()`.

filipi87 (approved):
LGTM!

src/pipecat/transports/daily/transport.py
aconchillo: Done

src/pipecat/transports/daily/transport.py
aconchillo: Done.

PR #3722, opened 2026-02-12
Fix race condition in finalized transcript early trigger

jamsea:
## Summary
  - Fixed a race condition in SpeechTimeoutUserTurnStopStrategy where a finalized transcript arriving after
  user_speech_timeout elapsed from VAD stop would immediately trigger a turn stop, even if the user was still
  mid-sentence
  - The root cause: when a finalized transcript arrived, the strategy checked `time.time() - vad_stopped_time >=
  user_speech_timeout and triggered immediately` — but STT processing latency consumed the user_speech_timeout window, leaving zero actual time for the user to resume speaking
  - Removed the early trigger path in _maybe_trigger_user_turn_stopped() and replaced it with smarter timeout
  recalculation in _handle_transcription(): when the finalized transcript arrives and user_speech_timeout still has
  time remaining, the timeout is shortened (no more STT wait needed); when it has elapsed, the original timeout
  continues running, providing a buffer for VAD to detect resumed speech
  - Added regression test reproducing the race condition with high STT latency

## Test plan
- [x] All 15 existing `test_user_turn_stop_strategy.py` tests pass
- [x] New `test_finalized_transcript_does_not_trigger_early_with_slow_stt` test passes
- [ ] Manual testing with high-latency STT service to verify no premature interruptions

### DIFF
--- a/changelog/3722.fixed.md
+++ b/changelog/3722.fixed.md
@@ -1,0 +1,5 @@
+- Fixed a race condition in `SpeechTimeoutUserTurnStopStrategy` where a finalized
+  transcript arriving after `user_speech_timeout` elapsed from VAD stop would
+  immediately trigger a turn stop, even if the user was still speaking. STT
+  processing latency was consuming the `user_speech_timeout` window, leaving no
+  time for the user to resume speaking.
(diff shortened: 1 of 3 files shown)


jamsea:
This might have already been fixed in 0.0.102, need to do additional testing to confirm 👀

jamsea:
Reported fixed post v0.0.102

PR #3725, opened 2026-02-12
feat(sarvam): add keepalive to SarvamSTTService to prevent idle timeout

### LINKED ISSUE #3699: SarvamSTTService: WebSocket dies during call with no reconnection or keepalive
## Description

`SarvamSTTService` WebSocket connections die after ~60-70 seconds of silence during live phone calls. Once dead, every subsequent `run_stt()` call fails with:

```
Error sending audio to Sarvam: no close frame received or sent
```

The connection is permanently broken for the rest of the call — there is no reconnection logic and no keepalive mechanism.

### Logs from production (pipecat 0.0.100, sarvamai SDK, websockets 15.0.1)

```
2026-02-10T15:51:59.239 | ERROR | pipecat.processors.frame_processor:push_error_frame:726 -
    SarvamSTTService#11 exception (websockets/legacy/protocol.py:921):
    Error sending audio to Sarvam: no close frame received or sent

2026-02-10T15:51:59.240 | WARNING | pipecat.pipeline.task:_source_push_frame:844 -
    PipelineTask#33: Something went wrong:
    ErrorFrame#4574(error: Error sending audio to Sarvam: no close frame received or sent, fatal: False)
```

This error repeats every ~20ms for every audio frame — the connection never recovers.

### Root cause

The `"no close frame received or sent"` means `close_rcvd=None, close_sent=None` in websockets — the TCP connection was severed without any WebSocket close handshake. This points to a server-side or proxy idle timeout killing the connection during silence periods.

### Why it doesn't recover

1. **`run_stt()`** catches the exception and yields an `ErrorFrame`, but does NOT reconnect:
   ```python
   except Exception as e:
       yield ErrorFrame(error=f"Error sending audio to Sarvam: {e}", exception=e)
   ```

2. **No connection health check** — the only guard is `if not self._socket_client:` which checks for `None`, not a dead connection. The `_socket_client` object still exists but points to a closed WebSocket.

3. **`_receive_task_handler()`** also exits when the connection dies (SDK's `start_listening()` raises `WebSocketException`), but nothing triggers a reconnect.

4. **No keepalive** — unlike `SarvamTTSService` which sends `{"type": "ping"}` every 20s to prevent idle timeout, `SarvamSTTService` has no keepalive mechanism.

## Comparison with other services

| Service | Keepalive | Reconnection |
|---------|-----------|-------------|
| **SarvamTTSService** | ✅ `{"type": "ping"}` every 20s | Via parent `WebsocketService` |
| **SonioxSTTService** | ✅ `{"type": "keepalive"}` every ~30s | — |
| **GladiaSTTService** | ✅ Empty audio bytes every 20s | — |
| **DeepgramSTTService** | ✅ `options={"keepalive": "true"}` | — |
| **SarvamSTTService** | ❌ None | ❌ None |

## Sarvam TTS docs confirm idle timeout

From [Sarvam TTS Streaming docs](https://docs.sarvam.ai/api-reference-docs/api-guides-tutorials/text-to-speech/streaming-api), under Input Message Types → Ping Message:

> **Ping Message** — Keeps the WebSocket connection alive; send periodically to avoid timeout. **The connection automatically closes after one minute of inactivity.**
> ```json
> {"type": "ping"}
> ```

The STT docs don't mention keepalive or idle timeout, but the observed ~60-70s timeout suggests the same infrastructure is used.

## Expected behavior

1. **Keepalive**: `SarvamSTTService` should send periodic keepalive messages (like the TTS service does) to prevent the connection from being killed during silence.

2. **Reconnection**: When the WebSocket dies unexpectedly, `SarvamSTTService` should attempt to reconnect automatically, similar to how other pipecat STT services handle connection failures.

3. **Connection health check**: `run_stt()` should validate the connection state before attempting to send, rather than just checking `if not self._socket_client`.

## Questions for Sarvam team

cc @shreyas-sarvam @pratham-sarvam @vinayak-sarvam

1. Does the STT WebSocket endpoint have an idle timeout? (We observe ~60-70s, matching the TTS documented 1-minute timeout)
2. Does the STT endpoint accept `{"type": "ping"}` messages like TTS does?
3. What is the recommended approach to keep STT connections alive during silence periods in voice agent scenarios?

## Environment

- pipecat: 0.0.100
- sarvamai SDK: latest
- websockets: 15.0.1 (legacy mode, used by sarvamai SDK)
- Model: `saaras:v3`
- Use case: Live phone calls with Asterisk (8kHz PCM audio)

disolaterX:
## Summary

Add a keepalive mechanism to `SarvamSTTService` that sends silent audio every 20 seconds to prevent Sarvam's WebSocket server from closing STT connections due to idle timeout (~60s).

**Resolves #3699**

## Problem

Sarvam's WebSocket server closes connections after ~60 seconds of inactivity. In voice AI pipelines, natural pauses are common — the user listens to the bot's response, thinks before replying, or the bot plays a long TTS response. During these pauses, no audio reaches the STT WebSocket, causing the server to kill the connection. Once dead, all subsequent `run_stt()` calls fail.

This was discussed in #3699 with @markbackman and @dhruvladia-sarvam, who suggested sending silent audio chunks as keepalive.

## Solution

Mirrors the existing keepalive pattern from `SarvamTTSService` (tts.py), adapted for the SDK-based STT client:

- **`_keepalive_task_handler()`**: Sends silent audio every 20s (same interval as TTS keepalive)
- **`_send_keepalive()`**: Sends 100ms of silent PCM through the SDK's public `transcribe()`/`translate()` API
- Task lifecycle managed via `create_task()` / `cancel_task()` in `_connect()` / `_disconnect()`

### Why silent audio instead of `{"type": "ping"}`?

`SarvamTTSService` uses `{"type": "ping"}` because it owns the raw WebSocket. `SarvamSTTService` goes through the Sarvam SDK (`self._socket_client.transcribe()`), which doesn't expose direct WebSocket access. Silent audio goes through the SDK's **public API**, which is stable across SDK versions and keeps the STT session alive at the application level.

## Changes

| File | Change |
|------|--------|
| `src/pipecat/services/sarvam/stt.py` | Add `import asyncio`, `_keepalive_task` field, keepalive start/stop in `_connect()`/`_disconnect()`, new `_keepalive_task_handler()` and `_send_keepalive()` methods |

## Testing

Tested in production telephony pipeline (Asterisk → Pipecat → Sarvam STT) where idle periods of 60+ seconds are common. Without keepalive, connections die silently. With keepalive, connections survive indefinitely through idle periods.

### DIFF
--- a/src/pipecat/services/sarvam/stt.py
+++ b/src/pipecat/services/sarvam/stt.py
@@ -11,6 +11,7 @@ API. It supports real-time transcription with Voice Activity Detection (VAD) and
 can handle multiple audio formats for Indian language speech recognition.
 """
 
+import asyncio
 import base64
 from dataclasses import dataclass
 from typing import AsyncGenerator, Dict, Literal, Optional
@@ -239,6 +240,7 @@ class SarvamSTTService(STTService):
         self._websocket_context = None
         self._socket_client = None
         self._receive_task = None
+        self._keepalive_task = None
 
         if self._vad_signals:
             self._register_event_handler("on_speech_started")
@@ -463,6 +465,10 @@ class SarvamSTTService(STTService):
             # Start receive task using Pipecat's task management
             self._receive_task = self.create_task(self._receive_task_handler())
 
+            # Start keepalive to prevent server idle timeout (~60s).
+            # Mirrors the pattern used in SarvamTTSService.
+            self._keepalive_task = self.create_task(self._keepalive_task_handler())
+
             logger.info("Connected to Sarvam successfully")
 
         except ApiError as e:
@@ -476,6 +482,10 @@ class SarvamSTTService(STTService):
 
     async def _disconnect(self):
         """Disconnect from Sarvam WebSocket API using SDK."""
+        if self._keepalive_task:
+            await self.cancel_task(self._keepalive_task)
+            self._keepalive_task = None
+
         if self._receive_task:
             await self.cancel_task(self._receive_task)
             self._receive_task = None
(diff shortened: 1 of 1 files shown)


markbackman:
A keepalive is the right move, but let's make it more generic and usable by all classes. We already had this in the WebsocketSTTService class. This PR moves it to STTService so that SarvamSTTService can use it:
https://github.com/pipecat-ai/pipecat/pull/3730

PR #3726, opened 2026-02-12
initial commit

dhruvladia-sarvam:
- Add silent-audio keepalive to Sarvam STT to prevent idle websocket timeouts.
- Rework STT websocket lifecycle to support reconnect attempts with backoff.
- Harden send path by routing audio sends through a shared helper and reconnecting on failures.


markbackman:
A keepalive is the right move, but let's make it more generic and usable by all classes. We already had this in the WebsocketSTTService class. This PR moves it to STTService so that SarvamSTTService can use it:
https://github.com/pipecat-ai/pipecat/pull/3730

PR #3727, opened 2026-02-12
Fix ElevenLabs TTS word timestamp interleaving across sentences

### LINKED ISSUE #3723: assistant context sentences are getting jumbled up
### pipecat version

0.0.102

### Python version

3.13

### Operating System

Mac

### Issue description

@filipi87 - I believe #3584 has broken something in the way we aggregate context for the assistant. The sentences that the assistant is speaking is getting jumbled up. 

I believe it might be because we are setting `self._cumulative_time = 0` every time in `run_tts` method. 

https://github.com/pipecat-ai/pipecat/blob/d99a25671591386128b99496156251169dd4ea23/src/pipecat/services/elevenlabs/tts.py#L703

### Reproduction steps

Run a pipeline with ElevenLabsTTSService

### Expected behavior

The assistant context should not have jumbled aggregation

### Actual behavior

The assistant context has jumbled aggregation

### Logs

```shell

```

markbackman:
## Summary

- Fixed `ElevenLabsTTSService` word timestamps being interleaved across sentences, causing garbled word ordering in the assistant context
- `_cumulative_time` and partial word state were being reset to zero on every `run_tts` call, but with the multi-stream WebSocket API multiple sentences share the same context -- so each sentence's word timestamps started from zero instead of accumulating after the previous sentence
- Moved the timing state reset into the new-context creation block so it only resets when a fresh context begins (e.g., after an interruption)

## Testing

- Run the ElevenLabs interruptible example and verify multi-sentence responses have words in the correct order in the assistant context

Fixes #3723

### DIFF
--- a/changelog/3727.fixed.md
+++ b/changelog/3727.fixed.md
@@ -1,0 +1,1 @@
+- Fixed `ElevenLabsTTSService` word timestamps being interleaved across sentences within the same context, causing incorrect word ordering.
--- a/src/pipecat/services/elevenlabs/tts.py
+++ b/src/pipecat/services/elevenlabs/tts.py
@@ -700,9 +700,6 @@ class ElevenLabsTTSService(AudioContextWordTTSService):
             try:
                 await self.start_ttfb_metrics()
                 yield TTSStartedFrame(context_id=context_id)
-                self._cumulative_time = 0
-                self._partial_word = ""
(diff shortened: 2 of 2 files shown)


markbackman:
Closing in favor of https://github.com/pipecat-ai/pipecat/pull/3729

PR #3729, opened 2026-02-12
TTS services fixes.

### LINKED ISSUE #3723: assistant context sentences are getting jumbled up

filipi87:
## Summary                                                                                                                     

- Fixed word timestamp interleaving issue in `ElevenLabsTTS` when processing multiple sentences
- Added `create_context_id()` override to 6 TTS services to properly reuse context IDs across multiple `run_tts` invocations   
- Fixed initialization logic to only emit `TTSStartedFrame` and start metrics once per turn, not once per sentence

### The Problem

When an LLM generates multiple sentences in a single turn (before `LLMFullResponseEndFrame`), the base `TTSService.create_context_id()` creates a new UUID for each sentence. This caused:

1. **Word timestamp interleaving** - timestamps tracked in separate contexts instead of one continuous stream
2. **Multiple TTSStartedFrame emissions** - incorrect lifecycle signaling
3. **Context tracking issues** - base class tracking multiple contexts when only one is actively used

The root cause: frames are only paused when receiving `LLMFullResponseEndFrame`, but by that point `run_tts()` has been invoked multiple times (once per sentence). Some services need to reuse the same `context_id` across all invocations within a turn.

  ### The Solution

Services that extend `AudioContextTTSService` or `AudioContextWordTTSService` and maintain `self._context_id` for turn tracking now override `create_context_id()` to:

1. Return existing `self._context_id` if one is already in progress
2. Create a new UUID only when starting a new turn
3. Reset `self._context_id` on interruptions or turn completion

### Services Fixed

1. **ElevenLabsTTSService** - Context reuse + Word timestamp interleaving across sentences
2. **InworldTTSService** - Context reuse + Moved initialization inside guard
3. **RimeTTSService** - Context reuse
4. **CartesiaTTSService** - Context reuse
5. **AsyncAITTSService** - Context reuse + Moved initialization inside guard
6. **PlayHTTTSService** - Context reuse

Fixes #3723

### DIFF
--- a/src/pipecat/services/elevenlabs/tts.py
+++ b/src/pipecat/services/elevenlabs/tts.py
@@ -13,6 +13,7 @@ with support for streaming audio, word timestamps, and voice customization.
 import asyncio
 import base64
 import json
+import uuid
 from typing import Any, AsyncGenerator, Dict, List, Literal, Mapping, Optional, Tuple, Union
 
 import aiohttp
@@ -680,6 +681,20 @@ class ElevenLabsTTSService(AudioContextWordTTSService):
             msg = {"text": text, "context_id": self._context_id}
             await self._websocket.send(json.dumps(msg))
 
+    def create_context_id(self) -> str:
+        """Generate a unique context ID for a TTS request in case we don't have one already in progress.
+
+        Returns:
+            A unique string identifier for the TTS context.
+        """
+        # If a context ID does not exist, create a new one.
+        # If an ID exists, continue using the current ID.
+        # When interruptions happens, user speech results in
+        # an interruption, which resets the context ID.
+        if not self._context_id:
+            return str(uuid.uuid4())
+        return self._context_id
+
     @traced_tts
     async def run_tts(self, text: str, context_id: str) -> AsyncGenerator[Frame, None]:
         """Generate speech from text using ElevenLabs' streaming WebSocket API.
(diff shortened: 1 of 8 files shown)


src/pipecat/services/elevenlabs/tts.py
```diff
@@ -680,6 +681,20 @@ async def _send_text(self, text: str):
...
+        """Generate a unique context ID for a TTS request in case we don't have one already in progress.
+
+        Returns:
+            A unique string identifier for the TTS context.
+        """
+        # If a context ID does not exist, create a new one.
+        # If an ID exists, continue using the current ID.
+        # When interruptions happens, user speech results in
+        # an interruption, which resets the context ID.
+        if not self._context_id:
```
markbackman, line 694: Could we just move this logic to the base class?

filipi87: I don’t think we should for now. Most of the TTS services don’t save the context_id, and we already have a reference to it in each TTS service. 

So maybe, in a follow up refactor, we can check which services are currently saving the context_id and refactor them. But for now, I would keep it like this.

filipi87: @markbackman, I’ve added this to our 1.0 wishlist, along with the other TTS improvements, so we can review how the classes are handling `self._context_id`. 👍

markbackman (approved):
LGTM.

This works as a temporary fix, but I agree that we'll want to make the API easier to work with in the future.

PR #3732, opened 2026-02-12
Refactored audio context management in TTS services

filipi87:
## Summary                                                                                                                                           
   
- Refactored audio context management in TTS services to improve encapsulation and reduce code duplication                                           
- Moved context ID tracking from individual TTS service implementations to the base `AudioContextTTSService` class
- Added `reuse_context_id_within_turn` parameter to control whether a TTS service should reuse the same context_id during a turn.
- Introduced helper methods (`has_active_audio_context()`, `get_active_audio_context_id()`, `remove_active_audio_context()`, `reset_active_audio_context()`) for managing the active audio context
- Updated context creation logic to reuse context IDs when concurrent contexts are not supported
- Simplified TTS service implementations (Cartesia, ElevenLabs, Inworld, Rime, AsyncAI) by removing duplicate context management code

## Details

**Context Management Improvements:**
- The base class now maintains `_context_id` to track the active context
- Services that don't support concurrent contexts (default behavior) will reuse the same context ID until an interruption occurs
- Context ID is automatically reset on interruption, allowing a new context to be created for the next TTS request

**Code Simplification:**
- Removed redundant `_context_id` instance variables from 5 TTS service implementations
- Removed duplicate `create_context_id()` implementations from individual services
- Unified context lifecycle management in the base class

**Affected Services:**
- `src/pipecat/services/cartesia/tts.py`
- `src/pipecat/services/elevenlabs/tts.py`
- `src/pipecat/services/inworld/tts.py`
- `src/pipecat/services/rime/tts.py`
- `src/pipecat/services/asyncai/tts.py`
- `src/pipecat/services/resembleai/tts.py`

### DIFF
--- a/src/pipecat/services/asyncai/tts.py
+++ b/src/pipecat/services/asyncai/tts.py
@@ -9,7 +9,6 @@
 import asyncio
 import base64
 import json
-import uuid
 from typing import AsyncGenerator, Optional
 
 import aiohttp
@@ -148,7 +147,6 @@ class AsyncAITTSService(AudioContextTTSService):
 
         self._receive_task = None
         self._keepalive_task = None
-        self._context_id = None
 
     def can_generate_metrics(self) -> bool:
         """Check if this service can generate processing metrics.
@@ -255,7 +253,7 @@ class AsyncAITTSService(AudioContextTTSService):
             if self._websocket:
                 logger.debug("Disconnecting from Async")
                 # Close all contexts and the socket
-                if self._context_id:
+                if self.has_active_audio_context():
                     await self._websocket.send(json.dumps({"terminate": True}))
                 await self._websocket.close()
                 logger.debug("Disconnected from Async")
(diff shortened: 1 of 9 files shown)


src/pipecat/services/asyncai/tts.py
```diff
@@ -362,19 +349,16 @@ async def _keepalive_task_handler(self):
...
             try:
                 await self._websocket.send(
-                    json.dumps(
-                        {"context_id": self._context_id, "close_context": True, "transcript": ""}
-                    )
+                    json.dumps({"context_id": context_id, "close_context": True, "transcript": ""})
                 )
             except Exception as e:
                 logger.error(f"Error closing context on interruption: {e}")
-            self._context_id = None
```
filipi87, line 377: The superclass is now the one responsible for setting it to `None` when an interruption happens.

src/pipecat/services/asyncai/tts.py
```diff
@@ -400,10 +384,10 @@ async def run_tts(self, text: str, context_id: str) -> AsyncGenerator[Frame, Non
 
                     self._context_id = context_id
```
filipi87: This is already happening inside `create_audio_context`, so we don't need it here.

aconchillo:
LGTM!

aconchillo (approved):

PR #3733, opened 2026-02-12
Deprecate unused class_decorators tracing module and fix stale comments

markbackman:
## Context

This is just dead code. We aren't using it and aren't maintaining. I'm deprecating and we should remove in 1.0.

## Summary

- Added tests for `TracingContext` and `TurnTraceObserver`, covering conversation/turn span lifecycle, parent-child relationships, interruption handling, concurrent pipeline isolation, and `TracingContext` state management
- Enabled the `tracing` extra in CI test and coverage workflows so tracing tests run with OpenTelemetry installed
- Deprecated the unused `class_decorators.py` module (`Traceable`, `@traceable`, `@traced`, `AttachmentStrategy`) which is not imported anywhere in the codebase
- Fixed stale comments in `service_decorators.py` that incorrectly described `Traceable` as active; added TODO for future removal

## Testing

```bash
uv run pytest tests/test_tracing_context.py tests/test_turn_trace_observer.py -v
```

### DIFF
--- a/changelog/3733.deprecated.md
+++ b/changelog/3733.deprecated.md
@@ -1,0 +1,1 @@
+- Deprecated unused `Traceable`, `@traceable`, `@traced`, and `AttachmentStrategy` in `pipecat.utils.tracing.class_decorators`. This module will be removed in a future release.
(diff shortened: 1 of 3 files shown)


filipi87 (approved):
Yep, makes sense to me. It will be nice to remove a lot of dead code before releasing 1.0. 🙂

PR #3735, opened 2026-02-13
Fix double execution of service functions on tracing errors

markbackman:
## Summary

- Fixed a bug where all five tracing service decorators (`traced_tts`, `traced_stt`, `traced_llm`, `traced_gemini_live`, `traced_openai_realtime`) would execute the wrapped function **twice** when the function itself raised an exception (e.g., LLM rate limit, TTS provider timeout). The outer `try/except` intended to catch tracing setup failures was also catching application errors, triggering an untraced fallback re-execution.


filipi87 (approved):
Makes sense. Nice catch!

PR #3738, opened 2026-02-13
Fix LLMUserAggregator broadcasting mute events before StartFrame

### LINKED ISSUE #3737: LLMUserAggregator broadcasts UserMuteStartedFrames before pushing StartFrames
### pipecat version

0.0.102

### Python version

3.12.11

### Operating System

### Issue description

When using the `MuteUntilFirstBotCompleteUserMuteStrategy` the `LLMUserAggregator` broadcasts a `UserMuteStartedFrame` before pushing the `StartFrame` to downstream processors (e.g. LLMService), causing an ERROR Log: 
```
| ERROR    | pipecat.processors.frame_processor:_check_started:946 - YourLLMService#2 Trying to process UserMuteStartedFrame#6 but StartFrame not received yet
```

### Reproduction steps

Configure a pipeline with `MuteUntilFirstBotCompleteUserMuteStrategy`:

```python
context = LLMContext()
    context,
    assistant_params=LLMAssistantAggregatorParams(),
        user_mute_strategies=[
            MuteUntilFirstBotCompleteUserMuteStrategy(),
        ],
        vad_analyzer=SileroVADAnalyzer(
            params=VADParams(
                stop_secs=0.2,
            )
        ),
        user_idle_timeout=12.0,
    ),
)
pipeline = Pipeline(
    [
        transport.input(),
        stt,
        context_aggregator.user(),
        llm,
        tts,
        transport.output(),
    ]
)
```

Start the pipeline and observe logs.

### Expected behavior

`StartFrame` reaches all processors before `UserMuteStartedFrame`.

### Actual behavior

`UserMuteStartedFrame` reaches downstream processors before `StartFrame`

### Logs

```shell
| ERROR    | pipecat.processors.frame_processor:_check_started:946 - YourLLMService#2 Trying to process UserMuteStartedFrame#6 but StartFrame not received yet
```

lukepayyapilli:
## Summary

Fixes #3737

- [x] Skip control frames (`StartFrame`, `EndFrame`, `CancelFrame`) in `_maybe_mute_frame()` so mute strategies never process them.
- [x] Add regression test verifying `StartFrame` reaches downstream before `UserMuteStartedFrame`.

## Approach/Rationale

The root cause is a frame-ordering race in `LLMUserAggregator.process_frame()`. When `StartFrame` arrives:

1. `super().process_frame()` sets `__started = True` on the aggregator.
2. `_maybe_mute_frame(StartFrame)` feeds the frame to mute strategies.
3. `MuteUntilFirstBotCompleteUserMuteStrategy` returns `True` (muted) for any frame before the bot's first speech.
4. Mute state changes - `on_user_mute_started` fires as an async task - `broadcast_frame(UserMuteStartedFrame)` pushes to all downstream processors.
5. Only after all of that does `push_frame(StartFrame)` send `StartFrame` downstream.

Downstream processors (e.g. `LLMService`) receive `UserMuteStartedFrame` before `StartFrame`, triggering `_check_started` errors.

The fix adds an early return at the top of `_maybe_mute_frame` for control frames. This prevents mute strategies from seeing frames they have no business processing, and keeps `process_frame()` unchanged.

Note: the default `TurnAnalyzerUserTurnStopStrategy` broadcasts a `SpeechControlParamsFrame` when it processes `StartFrame` (during `_user_turn_controller.process_frame()`). This frame gets re-queued back to the aggregator via `_queued_broadcast_frame` and `_internal_queue_frame`. When it arrives at `_maybe_mute_frame`, it legitimately triggers the mute state change and `UserMuteStartedFrame` broadcast - but by that point `StartFrame` has already reached downstream, so the ordering is correct.

## Tradeoffs

- **Guard in `_maybe_mute_frame` vs restructuring `process_frame`**: Moving control frame handling above `_maybe_mute_frame()` in `process_frame` would also fix the issue, but adds a nesting level to an already long method and moves the fix away from the actual defect. The guard in `_maybe_mute_frame` fixes it at the source.
- **Explicit frame type list vs `SystemFrame` check**: The guard checks for `StartFrame`, `EndFrame`, and `CancelFrame` explicitly rather than checking `isinstance(frame, SystemFrame)`. This is more conservative - it only skips the specific control frames that must flow unconditionally, rather than broadly skipping all system frames which might have legitimate interactions with mute strategies in the future.

### DIFF
--- a/src/pipecat/processors/aggregators/llm_response_universal.py
+++ b/src/pipecat/processors/aggregators/llm_response_universal.py
@@ -549,6 +549,15 @@ class LLMUserAggregator(LLMContextAggregator):
             await s.cleanup()
 
     async def _maybe_mute_frame(self, frame: Frame):
+        # Control frames must flow unconditionally — never feed them to mute
+        # strategies.  Without this guard, strategies like
+        # MuteUntilFirstBotCompleteUserMuteStrategy fire on_user_mute_started
+        # and broadcast UserMuteStartedFrame before StartFrame is pushed
+        # downstream, causing downstream processors to receive frames before
+        # StartFrame and log errors.
+        if isinstance(frame, (StartFrame, EndFrame, CancelFrame)):
+            return False
+
         should_mute_frame = self._user_is_muted and isinstance(
             frame,
             (
--- a/changelog/3737.fixed.md
+++ b/changelog/3737.fixed.md
@@ -1,0 +1,1 @@
+- Fixed `LLMUserAggregator` broadcasting mute events before `StartFrame` reaches downstream processors.
--- a/tests/test_context_aggregators_universal.py
+++ b/tests/test_context_aggregators_universal.py
@@ -24,7 +24,9 @@ from pipecat.frames.frames import (
     LLMThoughtEndFrame,
     LLMThoughtStartFrame,
     LLMThoughtTextFrame,
+    StartFrame,
     TranscriptionFrame,
+    UserMuteStartedFrame,
     UserStartedSpeakingFrame,
     UserStoppedSpeakingFrame,
     VADUserStartedSpeakingFrame,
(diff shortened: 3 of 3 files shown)


src/pipecat/processors/aggregators/llm_response_universal.py
```diff
@@ -549,6 +549,15 @@ async def _cleanup(self):
             await s.cleanup()
 
     async def _maybe_mute_frame(self, frame: Frame):
+        # Control frames must flow unconditionally — never feed them to mute
```
markbackman, line 552: Mind changing this comment to something like this?
```
# Lifecycle frames should never be muted and should not trigger mute
# state changes. Evaluating mute strategies on StartFrame would
# broadcast UserMuteStartedFrame before StartFrame reaches downstream
# processors.
```

markbackman (approved):
LGTM! Thank you. Just a small request to update the comment.

markbackman:
Merging and I'll update the comment separately.

PR #3744, opened 2026-02-14
Redesign UserIdleController to use BotStoppedSpeakingFrame

### LINKED ISSUE #3742: Should UserIdleController respect user mute and user turn stopped turn ?
### pipecat version

0.0.101

### Python version

_No response_

### Operating System

_No response_

### Question

Currently, the UserIdleController only responds to UserSpeakingFrame and BotSpeakingFrame. There is also special handling for FunctionCallsStartedFrame to prevent timeouts while waiting for the LLM. However, this feels like a band-aid solution, as there are gaps between these frames. For example, network latency or CPU overload could cause the UserIdleController to trigger a timeout unexpectedly.

I think that once we detect that a user turn has stopped, we should no longer expect user speech. This should also apply when a user is muted. 

Should UserIdleController stop waiting for user speech once a user turn ends or the user is muted?

### What I've tried

_No response_

### Context

I encountered this issue while trying to implement a DTMF authentication processor. Consider the following scenario:
A user is muted from the start. They remain muted until they are authenticated via a set of DTMF tones. This is collected in custom Frame processor.

In the current version of Pipecat: This workflow is not possible without encountering the bugs described above, or without using non-standard approaches.

Normally, DTMF tones from DTMF aggregator would be filtered because of mute, I created a custom DTMF aggregator to handle this, but UserIdleController still triggers even when the user is muted because it does not register any user activity but also it does not know about user's mute.

Regardless of what i was trying to achieve I’d love to hear your thoughts on this matter.

markbackman:
## Context

Normally, the `UserIdleController` works well—the user to bot latency is lower than `user_idle_timeout`. But, there are cases, where the LLM can be slow to respond; we've seen cases where this latency can be greater than 5 seconds in some extreme cases. Longer delay like this would cause a false trigger of the `UserIdleController`.

To make this more resilient, I'm changing the implementation to use a timeout, triggered from the BotStoppedSpeakingFrame.

## Summary

- Fixed `UserIdleController` false idle triggers caused by gaps between user and bot activity frames (e.g. during LLM/TTS processing latency)
- Replaced the continuous heartbeat-based timer (`UserSpeakingFrame`/`BotSpeakingFrame` + `asyncio.Event` loop) with a one-shot timer that starts on `BotStoppedSpeakingFrame` and cancels on `UserStartedSpeakingFrame` or `BotStartedSpeakingFrame`
- Added guards to suppress the timer during active user turns (interruption case) and pending function calls (where `FunctionCallsStartedFrame` can race with `BotStoppedSpeakingFrame`)
- Forwarded `UserStartedSpeakingFrame`/`UserStoppedSpeakingFrame` to the idle controller in `UserTurnProcessor` and `LLMUserAggregator`, since broadcast frames don't flow through the host's own `process_frame`

## Testing

- `uv run pytest tests/test_user_idle_controller.py` — 9 tests covering: idle after bot stops, user/bot speaking cancels timer, no idle before bot speaks, interruption suppression, idle cycling, cleanup cancellation, function call cancellation (normal ordering), and function call suppression (race condition ordering)

Fixes #3742

### DIFF
--- a/changelog/3744.fixed.md
+++ b/changelog/3744.fixed.md
@@ -1,0 +1,1 @@
+- Fixed `UserIdleController` false idle triggers caused by gaps between user and bot activity frames. The idle timer now starts only after `BotStoppedSpeakingFrame` and is suppressed during active user turns and function calls.
--- a/examples/foundational/17-detect-user-idle.py
+++ b/examples/foundational/17-detect-user-idle.py
@@ -5,11 +5,14 @@
 #
 
 
+import asyncio
 import os
 
 from dotenv import load_dotenv
 from loguru import logger
 
+from pipecat.adapters.schemas.function_schema import FunctionSchema
+from pipecat.adapters.schemas.tools_schema import ToolsSchema
 from pipecat.audio.vad.silero import SileroVADAnalyzer
 from pipecat.frames.frames import (
     EndTaskFrame,
@@ -30,6 +33,7 @@ from pipecat.runner.types import RunnerArguments
 from pipecat.runner.utils import create_transport
 from pipecat.services.cartesia.tts import CartesiaTTSService
 from pipecat.services.deepgram.stt import DeepgramSTTService
+from pipecat.services.llm_service import FunctionCallParams
 from pipecat.services.openai.llm import OpenAILLMService
 from pipecat.transports.base_transport import BaseTransport, TransportParams
 from pipecat.transports.daily.transport import DailyParams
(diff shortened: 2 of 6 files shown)


kedar389:
I think this is move into the right direction and seems like a good work.
 Prove me wrong but, would there be another possible bug/race condition if for example we would play filler before llm generation. This would finish with BotSpeaking and UserIdleController would start before generation ? I know that chances for this are really low and for this to happen we would need to have huge latencies either in LLM or CPU and idle timeout would need to be low.
 
 
 But I generally think this is more problem of the framework not having explicit modeling of turns for bot and user. But it seems that recent versions of pipecat seems to partly head that direction.

aconchillo (approved):

aconchillo:
> I think this is move into the right direction and seems like a good work. Prove me wrong but, would there be another possible bug/race condition if for example we would play filler before llm generation. This would finish with BotSpeaking and UserIdleController would start before generation ? I know that chances for this are really low and for this to happen we would need to have huge latencies either in LLM or CPU and idle timeout would need to be low.
> 

I think this should be fine. This is what would happen:

- User asks something.
- Bot replies. User idle timer starts after `BotStoppedSpeakingFrame`.
- User asks something else. User idle timer stops.
- LLM is processing for a long time. User idle timer is still sopped.
- User says "Hmmm". User idle timer is still sopped.
- The filler would actually interrupt the slow LLM and trigger a new one.
- Bot starts speaking and stops. User idle timer starts again.

So, as long as the bot hasn't started and stopped speaking, no timer is started.

> But I generally think this is more problem of the framework not having explicit modeling of turns for bot and user. But it seems that recent versions of pipecat seems to partly head that direction.

Yep, we now do have user and bot turns well separated after we introduce user turn strategies.

PR #3745, opened 2026-02-14
Serializer for Tata Smartflo Bi-Directional Audio Streaming WebSocket…

harisuryaa:
… protocol.

Smartflo's protocol is based on the Twilio Media Streams format with these differences:
    - Outbound media messages require an incrementing `chunk` field
    - Audio payloads should be multiples of 160 bytes to avoid audio gaps
    - Start event includes additional metadata (from, to, direction, bitRate, bitDepth)
    
    Reference: https://docs.smartflo.tatatelebusiness.com/docs/bi-directional-audio-streaming-integration-document

### DIFF
--- a/src/pipecat/serializers/tata_serializer.py
+++ b/src/pipecat/serializers/tata_serializer.py
@@ -1,0 +1,28 @@
+import json
+from pipecat.serializers.twilio import TwilioFrameSerializer
+
+class SmartfloFrameSerializer(TwilioFrameSerializer):
+    """Serializer for Tata Smartflo Bi-Directional Audio Streaming WebSocket protocol.
+    
+    Smartflo's protocol is based on the Twilio Media Streams format with these differences:
+    - Outbound media messages require an incrementing `chunk` field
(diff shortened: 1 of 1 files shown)


markbackman:
Thanks for the contribution! We've been evaluating which integrations to maintain in core vs. as community packages, and given that the Smartflo protocol is very close to Twilio (the main difference being the `chunk` field on outbound media), we think this would be a better fit as a community-maintained package.

A couple of notes on the current implementation that would need to be addressed either way:

- `_build_media_message()` isn't a method on `TwilioFrameSerializer` — the media message is built inline in `serialize()`, so the override here would never be called. You'd need to override `serialize()` instead.
- The 160-byte payload alignment mentioned in the Smartflo docs isn't handled.

If you'd like to publish this as a standalone package (e.g. `pipecat-serializer-smartflo`), we're happy to help point people to it. Take a look at our [Community Integrations](https://github.com/pipecat-ai/pipecat/blob/main/COMMUNITY_INTEGRATIONS.md) guide for more details. And if we see growing demand for Smartflo support, we'd be open to revisiting inclusion in core down the road.

PR #3748, opened 2026-02-14
Make UserIdleController always-on with dynamic timeout updates

markbackman:
## Context

In thinking about the issue filed in #3742, I realized that there are likely cases where you want to enable and disable the UserIdleController conditionally. Or, change the response time. I'm surprised we haven't had this request before, but it will make the UserIdleController more flexible to include in application code.

## Summary
- Always create `UserIdleController` with `timeout=0` (disabled) by default, removing all `Optional` type annotations and `if self._user_idle_controller:` guards from `UserTurnProcessor` and `LLMUserAggregator`
- Add `UserIdleTimeoutUpdateFrame` to allow enabling/disabling idle detection at runtime by updating the timeout dynamically
- Add tests for disabled-by-default, enable-via-frame, and disable-via-frame scenarios

## Test plan
- [x] `uv run pytest tests/test_user_idle_controller.py` — all 12 tests pass (9 existing + 3 new)
- [x] `uv run ruff check` — no lint issues

### DIFF
--- a/changelog/3748.added.md
+++ b/changelog/3748.added.md
@@ -1,0 +1,1 @@
+- Added `UserIdleTimeoutUpdateFrame` to enable or disable user idle detection at runtime by updating the timeout dynamically.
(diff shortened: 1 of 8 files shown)


aconchillo (approved):

aconchillo:
Nice!

PR #3751, opened 2026-02-14
Fix RTVIObserver missing transcriptions from Realtime LLMs

### LINKED ISSUE #3749: RTVIObserver drops upstream-only TranscriptionFrame from Gemini Live (and other realtime LLMs)
## Environment

- **pipecat version:** 0.0.102 (and current `main` at commit c682a44b)
- **Python version:** 3.13.5
- **Operating System:** macOS 15.4 (Darwin 25.2.0)

## Issue description

Commit `5f64dae0c` ("Filter RTVIObserver to downstream frames only") re-added a `direction != FrameDirection.DOWNSTREAM: return` guard to `RTVIObserver.on_push_frame()`. This was intended to prevent duplicate RTVI messages for frames that are broadcast in **both** directions (e.g. `UserStartedSpeakingFrame`, `FunctionCallResultFrame`).

However, `GeminiLiveLLMService` pushes `TranscriptionFrame` **only upstream** (`FrameDirection.UPSTREAM`) — it is never broadcast downstream. As a result, the RTVIObserver silently drops every user transcription, and RTVI clients never receive `user-transcription` messages.

The same issue likely affects `OpenAIRealtimeLLMService` and any other realtime/multimodal LLM service that pushes transcription frames upstream only.

**Relevant code:**

- `RTVIObserver.on_push_frame` direction guard: `src/pipecat/processors/frameworks/rtvi.py` (line ~1226)
- `GeminiLiveLLMService._push_user_transcription` pushing upstream: `src/pipecat/services/google/gemini_live/llm.py` (line ~1615)

```python
# RTVIObserver — drops ALL upstream frames including TranscriptionFrame
if direction != FrameDirection.DOWNSTREAM:
    return
```

```python
# GeminiLiveLLMService — pushes TranscriptionFrame upstream only
await self.push_frame(
    TranscriptionFrame(text=text, user_id="", timestamp=time_now_iso8601(), result=result),
    FrameDirection.UPSTREAM,
)
```

## Reproduction steps

1. Create a bot using `GeminiLiveLLMService` with any pipeline (e.g. the `26b-gemini-live-function-calling.py` example)
2. Connect an RTVI client (e.g. `@pipecat-ai/client-react`)
3. Speak to the bot
4. Observe that the bot responds with audio, but the RTVI client never receives `user-transcription` messages

## Expected behavior

RTVI clients should receive `user-transcription` messages containing the user's speech transcribed by the Gemini Live API (via `input_audio_transcription`).

This worked correctly before commit `5f64dae0c` when the direction filter was not present in `RTVIObserver`.

## Actual behavior

`RTVIObserver.on_push_frame()` returns early for all upstream frames. Since `TranscriptionFrame` from Gemini Live is only pushed upstream, it is never processed, and no `user-transcription` RTVI message is ever sent to the client.

The Gemini Live service **does** receive and process transcriptions from the API (visible in logs as `[Transcription:user] [...]` at DEBUG level), but the RTVIObserver discards them before they can be forwarded.

## Logs

```
# Transcription IS received from Gemini API:
2026-02-14 19:06:28.685 | DEBUG | pipecat.services.google.gemini_live.llm:_handle_msg_input_transcription:1683 - [Transcription:user] [maybe problem solve neighbor yeah.]

# But NO corresponding RTVI user-transcription message is ever sent to the client
```

## Suggested fix

Either:
1. **Broadcast `TranscriptionFrame` in both directions** from `GeminiLiveLLMService` (and other realtime LLM services), so the downstream copy reaches RTVIObserver — OR
2. **Allowlist `TranscriptionFrame` and `InterimTranscriptionFrame`** in the RTVIObserver's direction check, so upstream-only transcription frames are still processed

markbackman:
## Context

We added a filter for downstream frames only in 0.0.102 because of double messages, which in general works. The exception is realtime LLMs, which push transcription frames upstream. This change removes the direction restriction for `InterimTranscriptionFrame` and `TranscriptionFrame`.

In surveying other frames, these are the only two that need to be allow listed like this.

There are other ways to solve this, but this will patch the issue. I'm open to other ideas of how to solve this one.

## Summary

- Fixed `RTVIObserver` not sending `user-transcription` RTVI messages when using Realtime LLMs (OpenAI Realtime, Gemini Live)
- Realtime LLMs push `TranscriptionFrame` and `InterimTranscriptionFrame` upstream only, but the observer previously filtered to downstream-only frames
- Transcription frames are now allowed regardless of direction, while all other frame types still use the downstream-only filter to prevent duplicates

Fixes #3749

### DIFF
--- a/src/pipecat/processors/frameworks/rtvi.py
+++ b/src/pipecat/processors/frameworks/rtvi.py
@@ -1220,11 +1220,15 @@ class RTVIObserver(BaseObserver):
         frame = data.frame
         direction = data.direction
 
-        # Only process downstream frames. Some frames are broadcast in both
-        # directions (e.g. UserStartedSpeakingFrame, FunctionCallResultFrame),
-        # and we only want to send one RTVI message per event.
-        if direction != FrameDirection.DOWNSTREAM:
-            return
+        # Many frames are broadcast in both directions (e.g.
+        # UserStartedSpeakingFrame, FunctionCallResultFrame), so we default
+        # to only processing downstream copies to avoid duplicate RTVI
+        # messages. However, some frame types (like transcriptions) may only
+        # flow upstream (e.g. from Realtime LLMs), so we allow those
+        # regardless of direction.
+        if not isinstance(frame, (TranscriptionFrame, InterimTranscriptionFrame)):
+            if direction != FrameDirection.DOWNSTREAM:
+                return
 
         # If we have already seen this frame, let's skip it.
         if frame.id in self._frames_seen:
--- a/changelog/3751.fixed.md
+++ b/changelog/3751.fixed.md
@@ -1,0 +1,1 @@
+- Fixed `RTVIObserver` not sending user transcription messages when using Realtime LLMs (OpenAI Realtime, Gemini Live) that push transcription frames upstream.


src/pipecat/processors/frameworks/rtvi.py
```diff
@@ -1220,11 +1220,15 @@ async def on_push_frame(self, data: FramePushed):
...
-        # and we only want to send one RTVI message per event.
-        if direction != FrameDirection.DOWNSTREAM:
-            return
+        # Many frames are broadcast in both directions (e.g.
+        # UserStartedSpeakingFrame, FunctionCallResultFrame), so we default
+        # to only processing downstream copies to avoid duplicate RTVI
+        # messages. However, some frame types (like transcriptions) may only
+        # flow upstream (e.g. from Realtime LLMs), so we allow those
+        # regardless of direction.
+        if not isinstance(frame, (TranscriptionFrame, InterimTranscriptionFrame)):
```
filipi87, line 1229: Mark, regarding different ideas on how to solve this, another possible approach would be to filter only the frames that have been `broadcasted`. In this case, we would check the `DOWNSTREAM` direction.

Inside our base `Frame` class, we could add a new boolean field called `broadcasted`, which would be set to `True` when the `broadcast_frame` method is invoked.

Then, we would filter only `DOWNSTREAM` for frames that have been broadcasted.

What do you think ?

markbackman: Good idea!

markbackman:
Closing in favor of https://github.com/pipecat-ai/pipecat/pull/3774

PR #3753, opened 2026-02-15
Fix concurrency race condition in tracing and resource leaks in aggregators

RinZ27:

Working with the codebase revealed several stability issues in high-concurrency environments and multimodal use cases.

Safety improvements were prioritized in `service_decorators.py` to handle `traced_llm` method restoration. Current monkey-patching logic can lead to race conditions when multiple tasks share a service instance, so I added checks to ensure methods are only restored if they match the patched versions.

Memory leaks in `LLMAssistantContextAggregator` are now mitigated by cancelling pending `on_context_updated` tasks during reset. This ensures long-running bots don't accumulate stale tasks over time.

Robustness for `LangchainProcessor` has been enhanced to handle list-based message content. Previously, multimodal inputs would cause a crash during the `.strip()` call on non-string content.

Redundant imports in `tk.py` have also been removed for general cleanup.

Summary of changes:
- Mitigated race condition in `traced_llm` monkey-patching logic.
- Added resource cleanup (task cancellation) in `LLMAssistantContextAggregator.reset()`.
- Enhanced `LangchainProcessor` robustness for multimodal content.
- Removed redundant `import tkinter` in `tk.py`.

### DIFF
--- a/src/pipecat/processors/aggregators/llm_response.py
+++ b/src/pipecat/processors/aggregators/llm_response.py
@@ -845,6 +845,15 @@ class LLMAssistantContextAggregator(LLMContextResponseAggregator):
         """
         return bool(self._function_calls_in_progress)
 
+    async def reset(self):
+        """Reset the aggregation state."""
+        await super().reset()
+        # Cancel any pending context updated tasks
+        for task in list(self._context_updated_tasks):
+            if not task.done():
+                task.cancel()
+        self._context_updated_tasks.clear()
+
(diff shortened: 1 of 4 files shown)


markbackman:
We've done some of this work already.

- Tracing moved away from a singleton pattern
- Removed unused imports

The others are too unclear to accept as a fix. We would need more info on the root cause or issue you're solving. For now, I'm closing.

PR #3756, opened 2026-02-16
feat: add acefone and smartflo frame serializer

monster-anshu:

Added the support for the Acefone and TATA Smartflo Frame Serializer for websocket. Both platform works on same events so I have added the docs for both platforms.

https://docs.acefone.in/docs/bi-directional-audio-streaming-integration-document
https://docs.smartflo.tatatelebusiness.com/docs/bi-directional-audio-streaming-integration-document

### DIFF
--- a/changelog/3756.added.md
+++ b/changelog/3756.added.md
@@ -1,0 +1,1 @@
+- Added `AcefoneFrameSerializer` serializer support to websocket stream from Acefone and Smartlfo.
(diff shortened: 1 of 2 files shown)


markbackman:
You're welcome to submit these as community integrations:

I'll close out this PR for now since we're not planning to add support inside of Pipecat at this moment. But, a community integration is a great place to start, if you're looking to maintain it!

PR #3759, opened 2026-02-16
Switch Gradium TTS to AudioContextWordTTSService for multiplexing

markbackman:
## Context

Collab notebook from Gradium, showing how this works:
https://colab.research.google.com/drive/1MX1GqHPkU7FFKX7JQPcRgfwfbbS_FnoT#scrollTo=mxu7t_ngo_18

## Summary

- Switched `GradiumTTSService` from `InterruptibleWordTTSService` to `AudioContextWordTTSService`, eliminating the websocket disconnect/reconnect cycle on every interruption
- Uses Gradium's `client_req_id` field to multiplex multiple TTS sessions on a single websocket connection, following the same pattern already used by Cartesia, ElevenLabs, and other services
- Sends `close_ws_on_eos: false` in the setup message to keep the connection alive after `end_of_stream`
- Demultiplexes incoming messages by `client_req_id`, silently dropping audio for cancelled contexts instead of tearing down the connection
- Removed no-op `push_frame()` override

## Testing

- Verify TTS audio plays correctly
- Verify interruptions work without websocket reconnection (check logs for absence of "connecting"/"disconnecting" during interruptions)
- Verify word timestamps still work
- Verify voice change reconnects properly

### DIFF
--- a/src/pipecat/services/gradium/tts.py
+++ b/src/pipecat/services/gradium/tts.py
@@ -6,6 +6,7 @@
 
 import base64
 import json
+import uuid
 from typing import Any, AsyncGenerator, Mapping, Optional
 
 from loguru import logger
@@ -74,6 +75,7 @@ class GradiumTTSService(AudioContextWordTTSService):
         """
         super().__init__(
             push_stop_frames=True,
+            push_text_frames=False,
             pause_frame_processing=True,
             sample_rate=SAMPLE_RATE,
             **kwargs,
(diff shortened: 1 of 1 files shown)


markbackman:
@filipi87 I'm not sure if this is fully aligned with the latest context ID changes. If there are any issues, feel free to push changes.

filipi87 (approved):
Looks good. I have just pushed some fixes to reuse context IDs across multiple `run_tts` calls and prevent the parent class from pushing text frames.

PR #3764, opened 2026-02-17
feat: Add support for private endpoint in Azure STT

kedar389:
Added support for Azure STT that is hosted in private networks behind firewall.


kedar389:
@markbackman Could you have look at this ? It is a very short PR.

markbackman (approved):
LGTM! Thank you. 🙇

PR #3766, opened 2026-02-17
Fix missing TTFB metrics in STT OTel spans

itsderek23:
It appears that in pipecat v101+, opentelemetry STT spans are missing the `metrics.ttfb` attribute. This is a one-line change in STTService.push_frame() to store the calculated TTFB value before emitting the metric.

### DIFF
--- a/changelog/3766.fixed.md
+++ b/changelog/3766.fixed.md
@@ -1,0 +1,1 @@
+- Fixed missing `metrics.ttfb` attribute in STT OpenTelemetry spans.


aconchillo:
Thank you @itsderek23 ! I realized this could be done a bit differently reusing the already existing `start_ttb_metrics/stop_ttfb_metrics` with some changes. It makes it even simpler. Closing in favor of #3776

itsderek23:
Thanks @aconchillo !

Issue #3767, opened 2026-02-18
incorrect sample_rate in TavusInputTransport._on_participant_audio_data


tanmayc25:
### pipecat version

0.0.102

### Python version

3.12.8

### Operating System

macOs 26.2

### Issue description

In TavusInputTransport._on_participant_audio_data (src/pipecat/transports/tavus/transport.py, line 522), the InputAudioRawFrame is constructed with sample_rate=audio.audio_frames instead of sample_rate=audio.sample_rate.

This means the sample rate is set to the raw audio byte data rather than the actual sample rate integer, which would cause incorrect audio processing downstream.

### Reproduction steps

1. Use a Tavus transport with `TavusInputTransport`
2. A participant sends audio data, triggering `_on_participant_audio_data`
3. The `InputAudioRawFrame` is created with `sample_rate=audio.audio_frames` (raw audio bytes) instead of `sample_rate=audio.sample_rate`

### Expected behavior

```
  frame = InputAudioRawFrame(
      audio=audio.audio_frames,
      sample_rate=audio.sample_rate, # <-- changed here
      num_channels=audio.num_channels,
  )
```

### Actual behavior

```
  frame = InputAudioRawFrame(
      audio=audio.audio_frames,
      sample_rate=audio.audio_frames,  # bug: should be audio.sample_rate
      num_channels=audio.num_channels,
  )
```

### Logs

```shell

```

PR #3768, opened 2026-02-18
fix: use audio.sample_rate instead of audio.audio_frames in TavusInputTransport

tanmayc25:
Fixes issue: https://github.com/pipecat-ai/pipecat/issues/3767 

In TavusInputTransport._on_participant_audio_data (src/pipecat/transports/tavus/transport.py, line 522), the InputAudioRawFrame is constructed with sample_rate=audio.audio_frames
  instead of sample_rate=audio.sample_rate.

This means the sample rate is set to the raw audio byte data rather than the actual sample rate integer, which would cause incorrect audio processing downstream.
```
  # Current (incorrect)
  frame = InputAudioRawFrame(
      audio=audio.audio_frames,
      sample_rate=audio.audio_frames,  # bug: should be audio.sample_rate
      num_channels=audio.num_channels,
  )
 ```

### DIFF
--- a/changelog/3768.fixed.md
+++ b/changelog/3768.fixed.md
@@ -1,0 +1,1 @@
+- Fixed incorrect `sample_rate` assignment in `TavusInputTransport._on_participant_audio_data` (was using `audio.audio_frames` instead of `audio.sample_rate`).


aconchillo:
LGTM!

PR #3771, opened 2026-02-19
[Cartesia TTS] Clear `_context_id` when audio context completes

sauhardjain:
When Cartesia sends a `"done"` message, `_context_id` is not cleared. In non-standard flows (e.g. direct `TTSSpeakFrame` usage), subsequent TTS requests reuse the stale context ID, which the API may error on due to a race condition.

This adds a guard in `_process_messages()` to clear `_context_id` when the completed context matches. It's a no-op in the normal LLM pipeline flow where `flush_audio()` already clears `_context_id` before `"done"` arrives.


sauhardjain:
cc @markbackman @aconchillo for review 🙏🏼

markbackman:
@filipi87 has been working on these things.

sauhardjain:
Nice! Closing in favor of https://github.com/pipecat-ai/pipecat/pull/3732

PR #3774, opened 2026-02-19
Fix RTVIObserver missing upstream-only frames

markbackman:
## Summary

- Fixed `RTVIObserver` silently ignoring upstream-only frames. The observer previously filtered out all upstream frames to prevent duplicate RTVI messages from broadcasted frames (frames pushed in both directions). This meant any frame that only traveled upstream was never converted to an RTVI client message.
- Added a `broadcasted` boolean field to the base `Frame` class, automatically set to `True` by `broadcast_frame()` and `broadcast_frame_instance()`.
- Updated `RTVIObserver` to only skip the upstream copy of broadcasted frames, allowing upstream-only frames to be observed normally.

## Testing

- All existing tests pass (576 passed, 55 skipped)
- Verify upstream-only frames are now observed by RTVIObserver
- Verify broadcasted frames still only produce one RTVI message (downstream copy)

### DIFF
--- a/src/pipecat/frames/frames.py
+++ b/src/pipecat/frames/frames.py
@@ -131,6 +131,7 @@ class Frame:
     id: int = field(init=False)
     name: str = field(init=False)
     pts: Optional[int] = field(init=False)
+    broadcasted: bool = field(init=False)
     metadata: Dict[str, Any] = field(init=False)
     transport_source: Optional[str] = field(init=False)
     transport_destination: Optional[str] = field(init=False)
@@ -139,6 +140,7 @@ class Frame:
         self.id: int = obj_id()
         self.name: str = f"{self.__class__.__name__}#{obj_count(self)}"
         self.pts: Optional[int] = None
+        self.broadcasted: bool = False
         self.metadata: Dict[str, Any] = {}
         self.transport_source: Optional[str] = None
         self.transport_destination: Optional[str] = None
--- a/src/pipecat/frames/frames.py
+++ b/src/pipecat/frames/frames.py
@@ -123,7 +123,7 @@ class Frame:
         id: Unique identifier for the frame instance.
         name: Human-readable name combining class name and instance count.
         pts: Presentation timestamp in nanoseconds.
-        broadcasted_sibling_id: ID of the paired frame when this frame was
+        broadcast_sibling_id: ID of the paired frame when this frame was
             broadcast in both directions. Set automatically by
             ``broadcast_frame()`` and ``broadcast_frame_instance()``.
         metadata: Dictionary for arbitrary frame metadata.
@@ -134,7 +134,7 @@ class Frame:
     id: int = field(init=False)
     name: str = field(init=False)
     pts: Optional[int] = field(init=False)
-    broadcasted_sibling_id: Optional[int] = field(init=False)
+    broadcast_sibling_id: Optional[int] = field(init=False)
     metadata: Dict[str, Any] = field(init=False)
     transport_source: Optional[str] = field(init=False)
     transport_destination: Optional[str] = field(init=False)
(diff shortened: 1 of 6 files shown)


src/pipecat/frames/frames.py
```diff
@@ -131,6 +131,7 @@ class Frame:
     id: int = field(init=False)
     name: str = field(init=False)
     pts: Optional[int] = field(init=False)
+    broadcasted: bool = field(init=False)
```
filipi87: @markbackman, after discussing this with @kompfner and @aconchillo, we agreed that instead we should store `broadcasted_sibling_id`, which will reference the frame ID of the other frame that has been broadcast. This will be useful for debugging in the future.

So, I will apply this change so we can include this PR in the next release.

kwindla:
Related: I had an issue with ParallelPipeline sending multiple RTVI events, because of how observers work. I wonder if it makes sense to fix this behavior in this PR, too.

Claude came up with this very hacky workaround that unsets the "extra" RTVI observers: https://github.com/pipecat-ai/gradient-bang/blob/8d3c573259887eb7a6ea78ea223391b9284f5795/src/gradientbang/pipecat_server/bot.py#L361

filipi87:
> Related: I had an issue with ParallelPipeline sending multiple RTVI events, because of how observers work. I wonder if it makes sense to fix this behavior in this PR, too.

Hi @kwindla, in this case it wasn’t that the RTVI events were duplicated, but that you wanted to filter out RTVI events from a specific branch (source). Is that right ?

src/pipecat/frames/frames.py
```diff
@@ -123,6 +123,9 @@ class Frame:
         id: Unique identifier for the frame instance.
         name: Human-readable name combining class name and instance count.
         pts: Presentation timestamp in nanoseconds.
+        broadcasted_sibling_id: ID of the paired frame when this frame was
```
aconchillo: Searching for the correct past tense, it seems `broadcast` is more common that `broadcasted`.

aconchillo: Also, wondering if this should be a string instead. 🤔 . I'm just thinking mainly for Whisker. The id is hard to track, but not the name. For example, we could have id 1376434 but frame name `UserSpeakingFrame#56`. So, it would be `broadcast_sibling_name`, WDYT? I'm not sure though.

filipi87: I think that right now this is mostly for debugging, so if we think that using the name would make things easier to debug, I can do that.

However, the `id` still feels more natural for referencing something than the name. Maybe Whisker could simply create a link between the frames using the ID and retrieve any other information as needed ?

filipi87: I have renamed to `broadcast_sibling_id`, but let me know in case you think we should use the name instead.

aconchillo: OK, let's use id for now.

src/pipecat/processors/frameworks/rtvi.py
```diff
@@ -1220,10 +1220,9 @@ async def on_push_frame(self, data: FramePushed):
         frame = data.frame
         direction = data.direction
 
-        # Only process downstream frames. Some frames are broadcast in both
-        # directions (e.g. UserStartedSpeakingFrame, FunctionCallResultFrame),
-        # and we only want to send one RTVI message per event.
-        if direction != FrameDirection.DOWNSTREAM:
+        # For broadcasted frames (pushed in both directions), only process
```
aconchillo: Broadcast

aconchillo (approved):

aconchillo:
LGTM

filipi87:
@kwindla, I will push a follow up PR to filter out RTVI events from a specific branch, so we can include this one in today’s release.

src/pipecat/processors/frameworks/rtvi.py
filipi87: Done.

filipi87:
Follow up PR for filtering RTVI events from a specific source:
- https://github.com/pipecat-ai/pipecat/pull/3779

PR #3775, opened 2026-02-19
Fix Cartesia TTS dropping audio on null context_id

sauhardjain:
## Summary

- Use safe `msg.get("context_id")` instead of direct dict access
- Fall back to the current active context when `context_id` is missing
- Add warning/debug logging for diagnostics


markbackman:
@sauhardjain is this PR still relevant or can we close it out?

filipi87:
Closing this since it has been stale for a while, and the logic in Pipecat should already handle this correctly.

If there's still something that needs to be fixed, feel free to open a new issue and link to the PR with the fix.

PR #3779, opened 2026-02-19
Allowing to define the list of frame processors whose frames should be silently ignored by the RTVI observer.

filipi87:
## Summary
                                                                                                                
- Added `ignored_sources` parameter to `RTVIObserverParams` to specify a list of frame processors whose frames should be silently skipped by the RTVI observer
- Added `add_ignored_source()` and `remove_ignored_source()` methods to `RTVIObserver` for dynamically managing the ignored sources at runtime
- Useful for suppressing RTVI messages from secondary pipeline branches (e.g. a silent evaluation LLM) that should not be visible to clients

### DIFF
--- a/examples/foundational/53-concurrent-llm-rtvi-ignored-sources.py
+++ b/examples/foundational/53-concurrent-llm-rtvi-ignored-sources.py
@@ -156,9 +156,7 @@ async def run_bot(transport: BaseTransport, runner_args: RunnerArguments):
             enable_metrics=True,
             enable_usage_metrics=True,
         ),
-        rtvi_observer_params=RTVIObserverParams(
-            ignored_sources=[evaluator_llm]
-        ),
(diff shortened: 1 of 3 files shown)


examples/foundational/53-concurrent-llm-rtvi-ignored-sources.py
```diff
@@ -0,0 +1,193 @@
...
+
+    task = PipelineTask(
+        pipeline,
+        params=PipelineParams(
+            enable_metrics=True,
+            enable_usage_metrics=True,
+        ),
+        rtvi_observer_params=RTVIObserverParams(
+            ignored_sources=[evaluator_llm]
+        ),
```
filipi87: We can use like this.

aconchillo (approved):

aconchillo:
LGTM!

PR #3781, opened 2026-02-20
Fix race condition where context times out after sending second transcript

dakshdua:
One of our customers has seen an issue where audio from a context can sometimes be dropped. We're reasonably confident that this was the problem:
  1. LLM emits sentence 1 and `run_tts` creates new context and sends it to TTS service
  2. On audio, `append_to_audio_context` is called and frames go on the queue
  3. Audio for sentence 1 finishes — nothing more on the queue                                                                                
  4. LLM takes ~2.5s and generates a second sentence. `run_tts` is called again but `self._context_id` is still set so it skips creation
  5. `_handle_audio_context` times out since it's been > 3s since the last audio on the context and deletes the context 
  7. TTS service returns audio for the second sentence in the deleted context and `append_to_audio_context` drops the audio

The fix in this PR is to call `refresh_audio_context` whenever `run_tts` is called on an already existing context, which resets the 3s timeout so that the audio context is not deleted for at least 3s after the last text sent to the TTS service.

### DIFF
--- a/src/pipecat/services/asyncai/tts.py
+++ b/src/pipecat/services/asyncai/tts.py
@@ -402,6 +402,8 @@ class AsyncAITTSService(AudioContextTTSService):
 
                     if not self.audio_context_available(self._context_id):
                         await self.create_audio_context(self._context_id)
+                else:
+                    self.refresh_audio_context(self._context_id)
 
                 msg = self._build_msg(text=text, force=True, context_id=self._context_id)
                 await self._get_websocket().send(msg)
--- a/src/pipecat/services/cartesia/tts.py
+++ b/src/pipecat/services/cartesia/tts.py
@@ -627,6 +627,8 @@ class CartesiaTTSService(AudioContextWordTTSService):
                 yield TTSStartedFrame(context_id=context_id)
                 self._context_id = context_id
                 await self.create_audio_context(self._context_id)
+            else:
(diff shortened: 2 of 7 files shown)


markbackman:
@filipi87 can take a look. This may already be handled in https://github.com/pipecat-ai/pipecat/pull/3732

filipi87:
Hi @dakshdua, we have an open PR that refactors how we handle audio context across all TTS services that extend `AudioContextTTSService`, which includes Cartesia.

With this PR, the `context_id` and the audio context will be properly kept in sync, so the issue you reported above should no longer occur.

Could you give it a try and let us know if this PR already solves your issue?
https://github.com/pipecat-ai/pipecat/pull/3732

dakshdua:
I don't see how #3732 will solve the issue I described, as it does not seem like it does anything to change the timeout logic in `_handle_audio_context`.

filipi87:
Hi @dakshdua, you are right, we still need to reset the timeout when reusing the context.

Considering all the refactors we did, the only thing required to solve this now is the following:
https://github.com/pipecat-ai/pipecat/pull/3787

I’ve added you as the author of the commit.  Thank you for catching and fixing the issue.

I am closing this one in favor of https://github.com/pipecat-ai/pipecat/pull/3787

PR #3786, opened 2026-02-20
Refactoring the services using the WordTTSService

filipi87:
## Summary                                                                                                                      
                                                         
  - Merged word timestamp support from `WordTTSService` into the base `TTSService` class via a new `supports_word_timestamps`     
  parameter (defaults to `False`)                                           
  - Refactored all TTS services that previously extended `WordTTSService`, `AudioContextWordTTSService`, or                       
  `WebsocketWordTTSService` to instead extend their non-word counterparts and pass `supports_word_timestamps=True`
  - Deprecated `WordTTSService`, `WebsocketWordTTSService`, `AudioContextWordTTSService`, and `InterruptibleWordTTSService` — they
   remain as backward-compatible stubs

  ## Breaking Changes

  None. All deprecated classes remain functional and delegate to the new parameter-based approach.

  ## Deprecated

  - `WordTTSService` → use `TTSService(supports_word_timestamps=True)`
  - `WebsocketWordTTSService` → use `WebsocketTTSService(supports_word_timestamps=True)`
  - `AudioContextWordTTSService` → use `AudioContextTTSService(supports_word_timestamps=True)`
  - `InterruptibleWordTTSService` → use `InterruptibleTTSService(supports_word_timestamps=True)`

### DIFF
--- a/changelog/3786.changed.md
+++ b/changelog/3786.changed.md
@@ -1,0 +1,1 @@
+- Word timestamp support has been moved from `WordTTSService` into `TTSService` via a new `supports_word_timestamps` parameter. Services that previously extended `WordTTSService`, `AudioContextWordTTSService`, or `WebsocketWordTTSService` now pass `supports_word_timestamps=True` to their parent `__init__` instead.
--- a/src/pipecat/services/tts_service.py
+++ b/src/pipecat/services/tts_service.py
@@ -128,6 +128,8 @@ class TTSService(AIService):
         append_trailing_space: bool = False,
         # TTS output sample rate
         sample_rate: Optional[int] = None,
+        # if True, enables word-level timestamp tracking and synchronization
+        supports_word_timestamps: bool = False,
         # Text aggregator to aggregate incoming tokens and decide when to push to the TTS.
         text_aggregator: Optional[BaseTextAggregator] = None,
         # Types of text aggregations that should not be spoken.
@@ -160,6 +162,9 @@ class TTSService(AIService):
             append_trailing_space: Whether to append a trailing space to text before sending to TTS.
                 This helps prevent some TTS services from vocalizing trailing punctuation (e.g., "dot").
             sample_rate: Output sample rate for generated audio.
+            supports_word_timestamps: Whether this service supports word-level timestamp tracking.
+                When True, enables synchronization of audio with spoken words so only spoken words
+                are added to the conversation context.
             text_aggregator: Custom text aggregator for processing incoming text.
 
                 .. deprecated:: 0.0.95
@@ -230,6 +235,13 @@ class TTSService(AIService):
 
         self._processing_text: bool = False
         self._tts_contexts: Dict[str, TTSContext] = {}
+
+        # Word timestamp state (active when supports_word_timestamps=True)
+        self._supports_word_timestamps: bool = supports_word_timestamps
+        self._initial_word_timestamp: int = -1
+        self._initial_word_times: List[Tuple[str, float, Optional[str]]] = []
+        self._words_task: Optional[asyncio.Task] = None
+        self._llm_response_started: bool = False
 
         self._register_event_handler("on_connected")
         self._register_event_handler("on_disconnected")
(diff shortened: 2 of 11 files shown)


changelog/3786.changed.md
```diff
@@ -0,0 +1 @@
+- Word timestamp support has been moved from `WordTTSService` into `TTSService` via a new `supports_word_timestamps` parameter. Services that previously extended `WordTTSService`, `AudioContextWordTTSService`, or `WebsocketWordTTSService` now pass `supports_word_timestamps=True` to their parent `__init__` instead.
```
kompfner, line 1: In the PR description you have a really nice bulleted list showing "old" v "new". That's a lot more readable—I'd recommend using that here, too.

kompfner: Or, at least use that list in the "deprecated" changelog fragment.

src/pipecat/services/tts_service.py
```diff
@@ -157,6 +159,9 @@ def __init__(
             append_trailing_space: Whether to append a trailing space to text before sending to TTS.
                 This helps prevent some TTS services from vocalizing trailing punctuation (e.g., "dot").
             sample_rate: Output sample rate for generated audio.
+            supports_word_timestamps: Whether this service supports word-level timestamp tracking.
```
kompfner, line 165: nice explanation 👍

src/pipecat/services/tts_service.py
```diff
@@ -358,6 +372,8 @@ async def stop(self, frame: EndFrame):
         if self._stop_frame_task:
             await self.cancel_task(self._stop_frame_task)
             self._stop_frame_task = None
+        if self._words_task:
```
kompfner, line 394: nit: don't need `if self._words_task` since that's the first check in `_stop_words_task`

src/pipecat/services/tts_service.py
```diff
@@ -978,53 +963,26 @@ def __init__(self, *, reconnect_on_error: bool = True, **kwargs):
...
+    .. deprecated::
+        Word timestamp functionality has been moved to TTSService. Pass
+        ``supports_word_timestamps=True`` to InterruptibleTTSService instead.
     """
 
     def __init__(self, **kwargs):
         """Initialize the Interruptible Word TTS service.
 
         Args:
-            **kwargs: Additional arguments passed to the parent WebsocketWordTTSService.
```
kompfner, line 1036: just checking: this removed logic is just stuff that's already in `InterruptibleTTSService` right?

src/pipecat/services/tts_service.py
```diff
@@ -978,53 +963,26 @@ def __init__(self, *, reconnect_on_error: bool = True, **kwargs):
...
 
 
-class InterruptibleWordTTSService(WebsocketWordTTSService):
-    """Websocket-based TTS service with word timestamps that handles interruptions.
+class InterruptibleWordTTSService(InterruptibleTTSService):
+    """Deprecated. Use InterruptibleTTSService with supports_word_timestamps=True instead.
 
-    For TTS services that support word timestamps but can't correlate generated
-    audio with requested text. Handles interruptions by reconnecting when needed.
+    .. deprecated::
```
kompfner: We usually specify as of which version something is deprecated. In my latest change, I used 0.0.104, though I realize that might change. (I'll add a note to our notion checklist to check whether we need to update any deprecation version numbers when the time comes.)

kompfner (approved):
Much cleaner!

src/pipecat/services/tts_service.py
filipi87: I kept it here to follow the same pattern we were using elsewhere. But yeah, we don’t actually need it. 

In my follow up PR, I will be completely refactoring this anyway, so I’ll leave it as is for now to avoid unnecessary conflicts in the next PR.

src/pipecat/services/tts_service.py
filipi87: Makes sense.

src/pipecat/services/tts_service.py
filipi87: Yep, we had the same logic there.

changelog/3786.changed.md
filipi87: Done.

PR #3789, opened 2026-02-20
Fix missing await and interruption timeout hang

aconchillo:
## Summary

- Fixed `push_interruption_task_frame_and_wait()` hanging indefinitely when the `InterruptionFrame` does not reach the pipeline sink within the timeout. The event is now forcibly set after the timeout so the processor can continue.
- Added a `timeout` keyword argument to `push_interruption_task_frame_and_wait()` so callers can customize the wait duration (defaults to `INTERRUPTION_COMPLETION_TIMEOUT`).
- Fixed missing `await` on `add_audio_frames_message()` calls in Google audio-in examples (`07s`, `25`), which silently discarded the coroutine.


markbackman (approved):
LGTM!

PR #3790, opened 2026-02-20
Lazy import LocalSmartTurnAnalyzerV3 to suppress PyTorch warning

markbackman:
## Context

We've had reports of users seeing the warning:
```
PyTorch was not found. Models won't be available and only tokenizers, configuration and file/data utilities can be used.
```

even when they're not using the `TurnAnalyzerUserTurnStopStrategy`. This change lazy loads `LocalSmartTurnAnalyzerV3`.

We should also consider suppressing the PyTorch warning, which we can either add here or do separately.

## Summary
- Moves the `LocalSmartTurnAnalyzerV3` import in `user_turn_strategies.py` from module level into `__post_init__`, so it only loads when the default stop strategy is actually needed
- Eliminates the spurious "PyTorch was not found" warning that appeared whenever `llm_response_universal` was imported (e.g. in the quickstart example)

## Test plan
- [x] Verified `from pipecat.processors.aggregators.llm_response_universal import LLMContextAggregatorPair` no longer triggers the warning
- [x] Verified direct import of `LocalSmartTurnAnalyzerV3` still works
- [x] Existing unit tests pass

### DIFF
--- a/changelog/3790.fixed.md
+++ b/changelog/3790.fixed.md
@@ -1,0 +1,1 @@
+- Fixed spurious "PyTorch was not found" warning on startup by lazy-importing `LocalSmartTurnAnalyzerV3` in `UserTurnStrategies`.


markbackman:
Closing this. There's some type of version corruption happening with tqdm. Instead, we should just suppress the warning. I'll open another PR for that next week.

PR #3791, opened 2026-02-21
Adding the LemonSlice transport integration

jp-lemon:
Adding the LemonSliceTransport layer to easily integrate the LemonSlice Avatar with any Pipecat pipeline.

Users need to pass in a LemonSlice API key and a LemonSlice agent ID or publicly accessible avatar image url to start the session.
Users can pass in a Daily room to use or our API will create one for them under the LemonSlice domain.
This transport layer will add a "pipecat user" to the room to feed audio to LemonSlice through a special audio stream. Then a LemonSlice avatar will join the call and play the synced audio and video.

LemonSlice API docs: https://lemonslice.com/docs/api-reference/create-session


src/pipecat/transports/lemonslice/transport.py
```diff
@@ -0,0 +1,801 @@
...
+        params: LemonSliceParams = LemonSliceParams(),
+        callbacks: LemonSliceCallbacks,
+        api_key: str,
+        agent_image_url: Optional[str] = None,
+        agent_id: Optional[str] = None,
+        agent_prompt: Optional[str] = None,
+        idle_timeout: Optional[int] = None,
+        daily_room_url: Optional[str] = None,
+        daily_token: Optional[str] = None,
+        lemonslice_properties: Optional[dict] = None,
```
filipi87, line 96: It looks like we’re only using all these parameters inside the API, so maybe we could wrap them in a new class, `LemonSliceNewSessionRequest`.

src/pipecat/transports/lemonslice/transport.py
```diff
@@ -0,0 +1,801 @@
...
+        """
+        self._bot_name = bot_name
+        self._api = LemonSliceApi(api_key, session)
+        self._agent_id = agent_id
+        self._agent_image_url = agent_image_url
+        self._agent_prompt = agent_prompt
+        self._idle_timeout = idle_timeout
+        self._daily_room_url = daily_room_url
+        self._daily_token = daily_token
+        self._lemonslice_properties = lemonslice_properties
```
filipi87, line 123: We can keep a single variable here, `_session_request`.

src/pipecat/transports/lemonslice/transport.py
```diff
@@ -0,0 +1,801 @@
...
+    async def _initialize(self) -> str:
+        """Initialize the conversation and return the room URL."""
+        response = await self._api.create_session(
+            agent_image_url=self._agent_image_url,
+            agent_id=self._agent_id,
+            agent_prompt=self._agent_prompt,
+            idle_timeout=self._idle_timeout,
+            daily_room_url=self._daily_room_url,
+            daily_token=self._daily_token,
+            properties=self._lemonslice_properties,
```
filipi87, line 139: And here we would pass only a single parameter, `session_request`.

src/pipecat/transports/lemonslice/transport.py
```diff
@@ -0,0 +1,801 @@
...
+
+    def __init__(
+        self,
+        bot_name: str,
+        session: aiohttp.ClientSession,
+        api_key: str,
+        agent_image_url: Optional[str] = None,
+        agent_id: Optional[str] = None,
+        agent_prompt: Optional[str] = None,
+        idle_timeout: Optional[int] = None,
```
filipi87, line 681: Same here, we could wrap them in a new class, `LemonSliceNewSessionRequest`. As well as daily_room_url, daily_token and lemonslice_properties

src/pipecat/transports/lemonslice/transport.py
```diff
@@ -0,0 +1,801 @@
...
+
+    def __init__(
+        self,
+        *,
+        bot_name: str,
+        params: LemonSliceParams = LemonSliceParams(),
+        callbacks: LemonSliceCallbacks,
+        api_key: str,
+        agent_image_url: Optional[str] = None,
+        agent_id: Optional[str] = None,
```
filipi87, line 91: If we try to run the example without providing an `agent_id`, we receive the following error:
`ValueError: Provide either agent_id or agent_image_url`

Is there a default value we can set here?

examples/foundational/55-lemonslice-transport.py
```diff
@@ -0,0 +1,117 @@
...
+            ),
+        )
+
+        stt = DeepgramSTTService(api_key=os.getenv("DEEPGRAM_API_KEY"))
+
+        llm = GroqLLMService(api_key=os.getenv("GROQ_API_KEY"))
+
+        tts = ElevenLabsTTSService(
+            api_key=os.getenv("ELEVENLABS_API_KEY", ""),
+            voice_id="ys3XeJJA4ArWMhRpcX1D",
```
filipi87, line 56:
```suggestion
            voice_id=os.getenv("ELEVENLABS_VOICE_ID", ""),
```

filipi87 (approved):
This looks good. I’ve tested it, and it’s working nicely. 🙌
There are a couple of improvements I mentioned here, but I’ll push them in a follow-up PR.

filipi87:
I have pushed this [PR](https://github.com/pipecat-ai/pipecat/pull/3900), resolving the conflicts and implementing the improvements I mentioned above.

filipi87:
Thank you @jp-lemon for doing all this. 🙌🔥

markbackman:
Closing in favor of https://github.com/pipecat-ai/pipecat/pull/3900, which has been merged and will be in the upcoming release.

Excited to have LemonSlice support in Pipecat!

PR #3792, opened 2026-02-21
Update default Anthropic model to claude-sonnet-4-6

aconchillo:
## Summary

- Updated default Anthropic model from `claude-sonnet-4-5-20250929` to `claude-sonnet-4-6`.
- Removed explicit model override from the Anthropic function calling example since it now matches the default.


markbackman (commented):
👍

PR #3795, opened 2026-02-21
fix(realtime): handle response_cancel_not_active as non-fatal

### LINKED ISSUE #3755: OpenAIRealtimeLLMService: response_cancel_not_active error fatally kills WebSocket
### pipecat version

0.0.98

### Python version

3.12

### Operating System

macOS (darwin 25.2.0)

### Issue description

When using `OpenAIRealtimeLLMService` with `turn_detection=False` for push-to-talk, sending an `InterruptionFrame` when the bot is not actively responding causes a fatal WebSocket disconnection.

When you send a `ResponseCancelEvent`, the OpenAI API returns a `response_cancel_not_active` error, and `_receive_task_handler` treats all unhandled errors as fatal — exiting the receive loop and tearing down the connection. `response_cancel_not_active` should not be a fatal error.

### Reproduction steps

1. Configure `OpenAIRealtimeLLMService` with `turn_detection=False` in AudioInput
2. Start a conversation and let the bot finish its initial response
3. Send an InterruptionFrame
4. The InterruptionFrame triggers _handle_interruption() which sends ResponseCancelEvent even though no response is active

### Expected behavior

The `response_cancel_not_active` error should be treated as non-fatal. The WebSocket connection should remain open and the conversation should continue normally.

### Actual behavior

The receive loop in _receive_task_handler treats the error as fatal, exits the loop, and the WebSocket connection is destroyed.

### Logs

```shell

```

omChauhanDev:

Fixes #3755 

**Change** : `response_cancel_not_active` errors from the API are now logged as warnings instead of fatally killing the ws connection in realtime services.

### DIFF
--- a/src/pipecat/services/grok/realtime/llm.py
+++ b/src/pipecat/services/grok/realtime/llm.py
@@ -512,7 +512,7 @@ class GrokRealtimeLLMService(LLMService):
                 await self._handle_evt_function_call_arguments_done(evt)
             elif evt.type == "error":
                 if evt.error.code == "response_cancel_not_active":
-                    logger.warning(f"Non-fatal API error: {evt.error.message}")
+                    logger.debug(f"{self} {evt.error.message}")
                 else:
                     await self._handle_evt_error(evt)
                     return
--- a/src/pipecat/services/grok/realtime/llm.py
+++ b/src/pipecat/services/grok/realtime/llm.py
@@ -511,8 +511,11 @@ class GrokRealtimeLLMService(LLMService):
             elif evt.type == "response.function_call_arguments.done":
                 await self._handle_evt_function_call_arguments_done(evt)
             elif evt.type == "error":
-                await self._handle_evt_error(evt)
-                return
+                if evt.error.code == "response_cancel_not_active":
+                    logger.warning(f"Non-fatal API error: {evt.error.message}")
+                else:
+                    await self._handle_evt_error(evt)
+                    return
 
     async def _handle_evt_conversation_created(self, evt):
         """Handle conversation.created event - first event after connecting."""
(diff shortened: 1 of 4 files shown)


kollaikal-rupesh (commented):
Nice fix! The approach is correct — `response_cancel_not_active` should definitely not be fatal.

One minor suggestion: consider using `logger.debug()` instead of `logger.warning()` here. This error is an **expected, benign condition** in push-to-talk mode — every time the user presses the button while the bot is idle, this will fire. Using `warning` could create log noise in production for something that's completely normal. `debug` would be more appropriate for an expected code path.

Also, minor style nit: the pipecat convention for log messages typically prefixes with `{self}` for service identification, e.g.:
```python
logger.debug(f"{self} {evt.error.message}")
```

src/pipecat/services/grok/realtime/llm.py
```diff
@@ -511,8 +511,11 @@ async def _receive_task_handler(self):
             elif evt.type == "response.function_call_arguments.done":
                 await self._handle_evt_function_call_arguments_done(evt)
             elif evt.type == "error":
-                await self._handle_evt_error(evt)
-                return
+                if evt.error.code == "response_cancel_not_active":
+                    logger.warning(f"Non-fatal API error: {evt.error.message}")
```
markbackman: This was suggested already, but this should be debug. If it's a warning, it'll draw undue attention since it's a normal outcome of using the API.

markbackman (approved):
LGTM! Just a small comment to resolve before merging.

src/pipecat/services/grok/realtime/llm.py
omChauhanDev: make sense, done

PR #3796, opened 2026-02-21
Fix asyncio.Event race conditions in idle processors

### LINKED ISSUE #3402: Race Conditions and Lost Notifications Due to Unsafe asyncio.Event Reset Patterns
### pipecat version

main branch

### Python version

### Operating System

### Issue description

**Location:**

```python
async def wait(self):
    await self._event.wait()
    self._event.clear()
```

---

**Locations:**

```python
while True:
    try:
    except asyncio.TimeoutError:
        await self._callback(self)
    finally:
        self._idle_event.clear()
```

* Before `clear()` executes,

### Reproduction steps

### Expected behavior

-

### Actual behavior

-

### Logs

```shell

```

iamjr15:
## Summary

Fixes #3402

- Fixed race conditions in `IdleFrameProcessor` and `UserIdleProcessor` where `asyncio.Event` signals set during async timeout callbacks were unconditionally cleared by a `finally` block, causing false-positive idle detection.
- Moved `event.clear()` from the `finally` block to the success path only (after `event.wait()` returns), matching the correct pattern already used in `PipelineTask._idle_monitor_handler()`.
- Updated `EventNotifier` docstrings to document safety guarantees: single-consumer repeated cycles and multi-consumer one-shot usage are safe; multi-consumer repeated-notify cycles are not (no current codebase usage hits this case).
- Added comprehensive tests for `IdleFrameProcessor`, `UserIdleProcessor`, and `EventNotifier`, including deterministic white-box regression tests that fail before the fix and pass after.

### The bug

Both `_idle_task_handler()` methods used this pattern:

```python
# BEFORE (vulnerable):
try:
except asyncio.TimeoutError:
    await self._callback(self)  # yields control — activity signal can arrive here
finally:
    self._idle_event.clear()  # BUG: clears signal set during callback
```

If activity occurs during the async callback (which yields control via `await`), the signal is lost when `finally` runs, causing the processor to miss valid activity and incorrectly fire idle callbacks again.

### The fix

```python
# AFTER (safe):
try:
    self._idle_event.clear()  # Only clear on activity — not after timeout callbacks
except asyncio.TimeoutError:
    await self._callback(self)
```

## Test plan

- [x] `pytest tests/test_idle_frame_processor.py -v` — 4 tests (basic idle, activity reset, signal preservation regression, frame type filtering)
- [x] `pytest tests/test_user_idle_processor.py -v` — 7 tests (6 existing + 1 new regression test)
- [x] `pytest tests/test_event_notifier.py -v` — 3 tests (multi-consumer one-shot, coalescing, repeated cycles)
- [x] `ruff check` passes on all modified files
- [x] 5x stress test iterations pass with no flakiness


iamjr15:
Closing in favor of a new PR with clean commit history.

PR #3797, opened 2026-02-21
Fix asyncio.Event race conditions in idle processors

### LINKED ISSUE #3402: Race Conditions and Lost Notifications Due to Unsafe asyncio.Event Reset Patterns

iamjr15:
## Summary

Fixes #3402

- Fixed race conditions in `IdleFrameProcessor` and `UserIdleProcessor` where `asyncio.Event` signals set during async timeout callbacks were unconditionally cleared by a `finally` block, causing false-positive idle detection.
- Moved `event.clear()` from the `finally` block to the success path only (after `event.wait()` returns), matching the pattern used in `PipelineTask._idle_monitor_handler()`.

### The bug

```python
try:
except asyncio.TimeoutError:
    await self._callback(self)  # yields control — activity signal can arrive here
finally:
    self._idle_event.clear()  # clears signal set during callback
```

### The fix

```python
try:
    self._idle_event.clear()
except asyncio.TimeoutError:
    await self._callback(self)
```

## Test plan

- [x] `pytest tests/test_idle_frame_processor.py -v` — 4 tests pass
- [x] `pytest tests/test_user_idle_processor.py -v` — 7 tests pass (6 existing + 1 regression)
- [x] `ruff check` passes on all modified files

### DIFF
--- a/changelog/3796.fixed.md
+++ b/changelog/3796.fixed.md
@@ -1,0 +1,1 @@
+- Fixed `IdleFrameProcessor` where `asyncio.Event` was unconditionally cleared in a `finally` block instead of only on the success path.
--- a/src/pipecat/processors/idle_frame_processor.py
+++ b/src/pipecat/processors/idle_frame_processor.py
@@ -85,7 +85,6 @@ class IdleFrameProcessor(FrameProcessor):
         while True:
             try:
                 await asyncio.wait_for(self._idle_event.wait(), timeout=self._timeout)
+                self._idle_event.clear()
             except asyncio.TimeoutError:
                 await self._callback(self)
-            finally:
-                self._idle_event.clear()
--- a/tests/test_idle_frame_processor.py
+++ b/tests/test_idle_frame_processor.py
@@ -1,0 +1,104 @@
+#
+# Copyright (c) 2024-2026, Daily
+#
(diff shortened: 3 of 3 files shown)


iamjr15:
@aconchillo Would you be able to review this when you get a chance? This fixes the asyncio.Event race condition reported in #3402 — the `finally: event.clear()` pattern in both idle processors was clearing signals set during async callbacks, causing false-positive idle detection. Includes regression tests that deterministically reproduce the bug. Thanks!

aconchillo:
I think the changes are fine, but it's not actually a fix for any false-positive. There are not false-positive, worst case you would just get a delayed idle notification but not a premature one.

That is, if there's activity while the callback is running then we might get delayed notifications, because we would clear the event after the callback. But that's about it.

aconchillo (approved):

PR #3803, opened 2026-02-23
Inline local-smart-turn-v3 deps for Poetry compatibility

markbackman:
## Summary

- Replaced self-referential `pipecat-ai[local-smart-turn-v3]` extra in core dependencies with inlined `transformers` and `onnxruntime` packages for Poetry compatibility
- Self-referential extras work in uv/pip but are not supported by Poetry, causing dependency resolution failures
- These packages are required by `TurnAnalyzerUserTurnStopStrategy`, the default user turn stop strategy, so they belong in core dependencies
- Removed the `local-smart-turn-v3` optional extra from `[project.optional-dependencies]`
- Removed the `try/except ModuleNotFoundError` guard in `local_smart_turn_v3.py` since the imports are now always available
- Removed `--extra local-smart-turn-v3` from CI workflows (no longer needed)

## Breaking Changes

- The `local-smart-turn-v3` pip extra is removed. `pip install pipecat-ai[local-smart-turn-v3]` will no longer work, but the dependencies are now always installed with the base package.

## Testing

- [x] Verify `uv sync` installs `transformers` and `onnxruntime` correctly
- [x] Verify CI tests pass without the `--extra local-smart-turn-v3` flag
- [x] Verify Poetry can resolve dependencies for a project depending on `pipecat-ai`

### DIFF
--- a/.github/workflows/coverage.yaml
+++ b/.github/workflows/coverage.yaml
@@ -40,7 +40,6 @@ jobs:
             --extra google \
             --extra langchain \
             --extra livekit \
-            --extra local-smart-turn-v3 \
             --extra piper \
             --extra tracing \
             --extra websocket
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -44,7 +44,6 @@ jobs:
             --extra google \
             --extra langchain \
             --extra livekit \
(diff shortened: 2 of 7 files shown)


filipi87 (approved):
LGTM!

PR #3804, opened 2026-02-23
Allowing concurrent audio contexts

filipi87:
## Summary
 - Merged audio context management from `AudioContextTTSService` into the `TTSService` base class, making audio context support available to all TTS services.
- Added `_turn_context_id` to group all sentences from a single LLM response under one audio context, and `_playing_context_id` to track what is currently being played back (renamed from `_context_id`)
- Replaced the separate word-timestamp queue/task (`_words_task`) with inline processing via `_WordTimestampEntry` objects routed through the existing audio context queue.
- Moved `start_word_timestamps()` call into `_handle_audio_context` (triggered on the first audio chunk per context), removing the need for each provider to call it in their receive loops
- Removed `supports_word_timestamps` parameter, word timestamp logic is now always active; `flush_audio()` and `reset_word_timestamps()` are no-ops for services that don't implement them, so non-timestamp services are unaffected
- Enabled `pause_frame_processing=False` on `CartesiaTTSService`, allowing it to begin synthesizing the next sentence while the previous one is still playing
- Updated all WebSocket providers (`cartesia`, `elevenlabs`, `asyncai`, `inworld`, `rime`, `gradium`, `resembleai`) to inherit from `WebsocketTTSService` directly; `AudioContextTTSService` and `AudioContextWordTTSService` are kept as deprecated shims for backwards compatibility
- Fixed per-context guards in each provider's `run_tts()` (`has_active_audio_context()` → `audio_context_available(context_id)`) so a new context can start even when another is already active
- Updated `flush_audio()` signatures across all providers to accept an optional `context_id`, enabling callers to flush a specific context rather than always using the global active context
- Introduced `push_start_frame` parameter to `TTSService.__init__` so subclasses no longer need to manually call `create_audio_context` or push `TTSStartedFrame` — the base class handles it automatically
- Refactored all 25+ TTS service implementations to use the new `push_start_frame=True` pattern, reducing boilerplate across providers (AWS, Azure, Camb, Cartesia, Deepgram, ElevenLabs, Fish, Google, Gradium, Groq, Hathora, Hume, Inworld, Kokoro, LMNT, Minimax, Neuphonic, Nvidia, OpenAI, Piper, Rime, Sarvam, Speechmatics, Xtts, and others)
- Fixed `BotStoppedSpeakingFrame` emission: now emitted as soon as `TTSStoppedFrame` is received, with a fallback silence-based timeout increased to reduce false positives
- Fixed handling of `TTSSpeakFrame` to correctly treat it as a standalone utterance (not reusing the current turn context ID)

## Breaking Changes

- `supports_word_timestamps` parameter has been removed from `TTSService.__init__()`. Remove it from any custom subclass `super().__init__()` calls.
- `AudioContextTTSService` and `AudioContextWordTTSService` now emit `DeprecationWarning`. Subclass `WebsocketTTSService` directly instead.
- `WordTTSService`, `WebsocketWordTTSService`, and `InterruptibleWordTTSService` no longer pass `supports_word_timestamps=True` (the flag is gone); their behaviour is unchanged since the logic is always active
  now.


examples/foundational/07-interruptible.py
```diff
@@ -98,8 +98,16 @@ async def run_bot(transport: BaseTransport, runner_args: RunnerArguments):
     async def on_client_connected(transport, client):
         logger.info(f"Client connected")
         # Kick off the conversation.
-        messages.append({"role": "system", "content": "Please introduce yourself to the user."})
-        await task.queue_frames([LLMRunFrame()])
+        await task.queue_frames(
```
filipi87: TODO: remove it. Only for testing.

src/pipecat/services/cartesia/tts.py
```diff
@@ -308,8 +308,7 @@ def __init__(
         super().__init__(
             aggregate_sentences=aggregate_sentences,
             push_text_frames=False,
-            pause_frame_processing=True,
-            supports_word_timestamps=True,
+            pause_frame_processing=False,
```
filipi87, line 337: To test this feature, we are no longer pausing frame processing for Cartesia. Everything is now running in parallel. We can do the same for the other TTS providers that support context.

src/pipecat/services/tts_service.py
```diff
@@ -1286,12 +1163,173 @@ async def on_audio_context_completed(self, context_id: str):
         pass
 
 
+class WordTTSService(TTSService):
+    """Deprecated. Use TTSService directly instead.
+
+    .. deprecated::
```
markbackman: Add the version to the deprecation. Same for all other class deprecations below.

```suggestion
    .. deprecated:: 0.0.104
```

src/pipecat/services/tts_service.py
filipi87: TODO: we also need to pass the `context_id` here when invoking `flush_audio`

src/pipecat/services/tts_service.py
```diff
@@ -1337,47 +1274,56 @@ async def _audio_context_task_handler(self):
...
                 # audio available (i.e. we find None).
                 await self._handle_audio_context(context_id)
 
                 # We just finished processing the context, so we can safely remove it.
-                del self._contexts[context_id]
+                del self._audio_contexts[context_id]
                 await self.on_audio_context_completed(context_id=context_id)
                 self.reset_active_audio_context()
-
-                # Append some silence between sentences.
```
filipi87, line 1352: If we want to keep sending silence, I will need to move this logic to a different place.

Let me know what you think.

cc: @markbackman

src/pipecat/transports/base_output.py
```diff
@@ -44,12 +44,15 @@
...
     TTSAudioRawFrame,
+    TTSStoppedFrame,
 )
 from pipecat.processors.frame_processor import FrameDirection, FrameProcessor
 from pipecat.transports.base_transport import TransportParams
 from pipecat.utils.time import nanoseconds_to_seconds
 
 BOT_VAD_STOP_SECS = 0.35
+# Only used as a fallback
+BOT_VAD_STOP_FALLBACK_SECS = 3
```
filipi87, line 55: - `BOT_VAD_STOP_SECS` is still used when we are continuously receiving audio, in the “speech stream” case, for example, with Tavus and HeyGen.

- `BOT_VAD_STOP_FALLBACK_SECS`: a new value used only as a fallback. It is used inside `_next_frame`, but we should rarely (if ever) hit this path now. We are increasing the timeout to allow enough time to receive the `TTSStoppedSpeakingFrame` instead of relying on this fallback. This help preventing from receiving multiple bot started and stopped speaking events in the same turn.

src/pipecat/services/tts_service.py
filipi87: Done.

src/pipecat/services/elevenlabs/tts.py
```diff
@@ -514,20 +513,15 @@ async def _update_settings(self, delta: TTSSettings) -> dict[str, Any]:
...
-                if self._websocket:
-                    await self._websocket.send(
-                        json.dumps({"context_id": context_id, "close_context": True})
-                    )
-            except Exception as e:
-                await self.push_error(error_msg=f"Unknown error occurred: {e}", exception=e)
-            self.reset_active_audio_context()
+            audio_contexts = self.get_audio_contexts()
+            if audio_contexts:
+                for ctx_id in audio_contexts:
```
aconchillo, line 568: nit: simplification:

```
for ctx_id in self.get_audio_contexts():
    await self._close_context(ctx_id)
```

aconchillo: same in other places.

aconchillo:
Looks awesome @filipi87 !

aconchillo (approved):

PR #3806, opened 2026-02-23
Align Ultravox Realtime service with OpenAI/Gemini patterns

markbackman:
## Context

I made some spot fixes to UltravoxRealtimeLLMService. The big missing pieces were:
1. TTSTextFrames and proper LLMTextFrame handling.
2. TTFB metrics
3. Text or voice modality handling (and an example showing how to use it)

There's still more work to do, but this makes the class more closely aligned with the frame flow of other Realtime LLM services.

## Summary

- Added `output_medium` parameter to `AgentInputParams` and `OneShotInputParams` to control initial output medium (text vs voice) at call creation time
- Restructured agent transcript handling to properly separate voice and text modality flows:
  - **Voice mode**: audio lifecycle managed by `_handle_audio()`; transcripts push `LLMTextFrame` (`append_to_context=False`) and `TTSTextFrame` for deltas only
  - **Text mode**: transcript manages its own response lifecycle with `LLMFullResponseStartFrame`/`EndFrame`; pushes `LLMTextFrame` (appends to context); no `TTSTextFrame`
- Added `InterruptionFrame` handling with `stop_all_metrics()` cleanup
- Added processing metrics (`start_processing_metrics`/`stop_processing_metrics`) at response boundaries
- Improved TTFB measurement accuracy using `VADUserStoppedSpeakingFrame` speech end time instead of `UserStoppedSpeakingFrame`
- Enabled metrics generation (`can_generate_metrics()`)
- Updated voice example with proper user turn strategies (VAD + speech timeout) and transcript event handlers
- Added text-only output example (`50a-ultravox-realtime-text.py`)

## Testing

- Test voice mode: `uv run examples/foundational/50-ultravox-realtime.py` — verify audio output with text transcripts
- Test text mode: `uv run examples/foundational/50a-ultravox-realtime-text.py` — verify `LLMTextFrame` output without duplicate TTS
- Test interruption handling: interrupt the bot mid-response, verify metrics cleanup
- Run `uv run pytest tests/`

### DIFF
--- a/src/pipecat/services/ultravox/llm.py
+++ b/src/pipecat/services/ultravox/llm.py
@@ -31,6 +31,7 @@ from pipecat.frames.frames import (
     Frame,
     InputAudioRawFrame,
     InputTextRawFrame,
+    InterruptionFrame,
     LLMContextFrame,
     LLMFullResponseEndFrame,
     LLMFullResponseStartFrame,
@@ -42,7 +43,7 @@ from pipecat.frames.frames import (
     TTSStoppedFrame,
     TTSTextFrame,
     UserAudioRawFrame,
-    UserStoppedSpeakingFrame,
+    VADUserStoppedSpeakingFrame,
 )
 from pipecat.processors.aggregators.llm_context import LLMContext
 from pipecat.processors.aggregators.llm_response import (
@@ -90,6 +91,9 @@ class AgentInputParams(BaseModel):
         template_context: Context variables to use when instantiating a call with the
             agent. Defaults to an empty dict.
         metadata: Metadata to attach to the call. Default to an empty dict.
+        output_medium: The initial output medium for the agent. Use "text" for text
+            responses or "voice" for audio responses. Defaults to None, which uses the
+            agent's default.
         max_duration: The maximum duration of the call. Defaults to None, which will
             use the agent's default maximum duration.
         extra: Extra parameters to include in the agent call creation request. Defaults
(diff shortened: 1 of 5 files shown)


src/pipecat/services/ultravox/llm.py
```diff
@@ -528,22 +579,29 @@ async def _handle_user_transcript(self, text: str):
...
+                frame.append_to_context = False
+                await self.push_frame(frame)
+            if delta:
+                tts_frame = TTSTextFrame(text=delta, aggregated_by=AggregationType.WORD)
+                tts_frame.includes_inter_frame_spaces = True
+                await self.push_frame(tts_frame)
+        elif medium == "text":
+            if final:
+                await self.stop_processing_metrics()
                 await self.push_frame(LLMFullResponseEndFrame())
```
filipi87, line 610: Is there a chance that we receive only a single message that is already "final" ? If so, would we never trigger `LLMFullResponseStartFrame`, or is this not possible?

filipi87 (approved):
The changes make sense to me. Just one question. 🚀

src/pipecat/services/ultravox/llm.py
markbackman: This is as short as it gets, results from:
```
print(f"medium: {medium}, text: {text}, delta: {delta}, final: {final}")
```

```
medium: voice, text: , delta: None, final: False
medium: voice, text: None, delta: Order, final: False
medium: voice, text: None, delta:  , final: False
medium: voice, text: Order , delta: None, final: True
```

It looks like there's always a delta before the final, so this should be safe for now unless they break the API.

PR #3809, opened 2026-02-24
Add Krisp API key support and debug logging

markbackman:
## Context

The motivation for this PR is to provide transparency to the metrics data for `KrispVivaTurn`. I took the opportunity to also remove dead code fields in the `SmartTurnMetricsData` and created a generic `TurnMetricsData` class.

While working on that, I found that the new Krisp SDK requires an API key, so I added support for that too, in 3bb9b07d460aa876de19df003ca80a547ce4a9f8.

## Summary

- Added `TurnMetricsData` as a generic metrics class for turn detection, replacing the service-specific `SmartTurnMetricsData`. `KrispVivaTurn` now emits `TurnMetricsData` with `e2e_processing_time_ms` measuring the interval from VAD speech-to-silence transition to model threshold crossing.
- `BaseSmartTurn` now emits `TurnMetricsData` instead of `SmartTurnMetricsData`. Deprecated `SmartTurnMetricsData` (still available as a subclass for backwards compatibility).
- `MetricsLogObserver` normalizes deprecated `SmartTurnMetricsData` in `include_metrics` filters and logs all turn metrics via a unified `_log_turn` handler.
- Metrics are consumed in the strategy `_handle_input_audio` path when `append_audio` returns COMPLETE, so they are pushed immediately while fresh.
- Added `api_key` parameter to `KrispVivaSDKManager`, `KrispVivaTurn`, and `KrispVivaFilter` for Krisp SDK v1.8.0 licensing. Falls back to `KRISP_VIVA_API_KEY` environment variable.
- Updated examples (`07p`, `38b`) to use `TurnMetricsData` and `MetricsLogObserver`.

## Testing

- Run `07p-interruptible-krisp-viva.py` with `KRISP_VIVA_API_KEY` set and verify turn metrics are logged with e2e processing time
- Run `38b-smart-turn-local.py` and verify `TurnMetricsData` metrics appear in logs
- Verify `include_metrics={SmartTurnMetricsData}` still works (with deprecation warning)


scripts/evals/run-release-evals.py
```diff
@@ -123,6 +123,7 @@ def EVAL_VISION_IMAGE(*, eval_speaks_first: bool = False):
     ("07n-interruptible-google.py", EVAL_SIMPLE_MATH),
     ("07n-interruptible-google-http.py", EVAL_SIMPLE_MATH),
     ("07o-interruptible-assemblyai.py", EVAL_SIMPLE_MATH),
+    ("07p-interruptible-krisp-viva.py", EVAL_SIMPLE_MATH),
```
markbackman, line 126: @aconchillo I can provide the Krisp SDK and model for running release evals.

src/pipecat/audio/turn/smart_turn/base_smart_turn.py
```diff
@@ -222,27 +222,18 @@ def _process_speech_segment(self, audio_buffer) -> Tuple[EndOfTurnState, Optiona
                 # Calculate processing time
                 e2e_processing_time_ms = (end_time - start_time) * 1000
 
-                # Extract metrics from the nested structure
-                metrics = result.get("metrics", {})
-                inference_time = metrics.get("inference_time", 0)
```
markbackman, line 227: These fields weren't populated, so removing them along with SmartTurnMetricsData.

src/pipecat/audio/filters/krisp_viva_filter.py
```diff
@@ -48,6 +52,8 @@ def __init__(
                 If None, uses KRISP_VIVA_FILTER_MODEL_PATH environment variable.
             frame_duration: Frame duration in milliseconds.
             noise_suppression_level: Noise suppression level.
+            api_key: Krisp SDK API key. If empty, falls back to
+                the KRISP_API_KEY environment variable.
```
filipi87: @markbackman, is this something new? As far as I remember, we never needed an API key for Krisp before. Is it required now?

If it is required, we should probably update our docs as well:
- https://docs.pipecat.ai/guides/features/krisp-viva

src/pipecat/audio/krisp_instance.py
```diff
@@ -88,25 +89,46 @@ class KrispVivaSDKManager:
...
         """
         with cls._lock:
             # Initialize SDK on first acquire
             if cls._reference_count == 0:
                 try:
-                    krisp_audio.globalInit("", cls._log_callback, krisp_audio.LogLevel.Off)
+                    key = api_key or os.environ.get("KRISP_API_KEY", "")
+                    try:
+                        # New SDK signature (requires license key)
+                        krisp_audio.globalInit(
```
filipi87, line 122: Even in Krisp’s own documentation, I don’t see an API key being required:
- https://sdk-docs.krisp.ai/docs/getting-started-server#voice-isolation-model

env.example
```diff
@@ -104,6 +104,7 @@ INWORLD_API_KEY=...
 KRISP_MODEL_PATH=...
 
 # Krisp Viva
+KRISP_API_KEY=...
```
filipi87: Should we call it `KRISP_VIVA_API_KEY` to match the other environment variables?

filipi87 (approved):
Thanks for confirming with Krisp. LGTM!

src/pipecat/audio/filters/krisp_viva_filter.py
markbackman: Good call. Docs PR: https://github.com/pipecat-ai/docs/pull/548

src/pipecat/audio/krisp_instance.py
markbackman: I know, this isn't well documented, but it is required. You can find it in the SDK ref docs when you download the SDK. The Krisp team confirmed version 1.6.1 and newer require it.

env.example
markbackman: Good idea. Yes. I'll do that now.

PR #3812, opened 2026-02-24
Make clearer the distinction between "storage-mode" and "delta-mode" …

kompfner:
…usage of `*Settings` objects

- Storage mode: for use in `self._settings`. All fields should be specified, i.e. should not be `NOT_GIVEN`.
- Delta mode: for use in `*UpdateSettingsFrame`.

In service of this, this commit:
- Adds a runtime check that all fields are specified in storage mode
- Updates all services to specify all fields in stored settings
- Updates all services to no longer check for `is_given` in stored settings (not necessary anymore)
- Updates relevant docstrings
- Renames `update` to `delta` in `*UpdateSettingsFrame`
- Updates community integrations guide

### DIFF
--- a/src/pipecat/frames/frames.py
+++ b/src/pipecat/frames/frames.py
@@ -2127,7 +2127,7 @@ class ServiceUpdateSettingsFrame(ControlFrame):
     Parameters:
         settings: Dictionary of setting name to value mappings.
 
-            .. deprecated:: 0.0.103
+            .. deprecated:: 0.0.104
                 Use ``delta`` with a typed settings object instead.
 
         delta: :class:`~pipecat.services.settings.ServiceSettings` delta-mode
--- a/src/pipecat/frames/frames.py
+++ b/src/pipecat/frames/frames.py
@@ -2121,21 +2121,21 @@ class TTSStoppedFrame(ControlFrame):
 class ServiceUpdateSettingsFrame(ControlFrame):
     """Base frame for updating service settings.
 
-    Supports both a ``settings`` dict (for backward compatibility) and an
-    ``update`` object.  When both are provided, ``update`` takes precedence.
+    Supports both a ``settings`` dict (for backward compatibility) and a
+    ``delta`` object.  When both are provided, ``delta`` takes precedence.
 
     Parameters:
         settings: Dictionary of setting name to value mappings.
 
             .. deprecated:: 0.0.103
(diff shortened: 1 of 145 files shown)


src/pipecat/frames/frames.py
```diff
@@ -2121,21 +2121,21 @@ class TTSStoppedFrame(ControlFrame):
...
-    ``update`` object.  When both are provided, ``update`` takes precedence.
+    Supports both a ``settings`` dict (for backward compatibility) and a
+    ``delta`` object.  When both are provided, ``delta`` takes precedence.
 
     Parameters:
         settings: Dictionary of setting name to value mappings.
 
-            .. deprecated:: 0.0.103
-                Use ``update`` with a typed settings object instead.
+            .. deprecated:: 0.0.104
```
kompfner, line 2130: Found various mentions of version 0.0.103 in deprecation messages from https://github.com/pipecat-ai/pipecat/pull/3714 that should've been updated to 0.0.104 (or...maybe another version?!) Updating them here.

filipi87 (commented):
> Storage mode: for use in self._settings. All fields should be specified, i.e. should not be NOT_GIVEN.

@kompfner, regarding this, I can see now that we are specifying the fields when creating `self._settings`, but there is nothing enforcing it, right ? Because the `*Settings` classes still allow `_NotGiven`.

Is this expected?

src/pipecat/services/ai_service.py
```diff
@@ -73,6 +73,7 @@ async def start(self, frame: StartFrame):
         Args:
             frame: The start frame containing initialization parameters.
         """
+        self._settings.validate_complete()
```
filipi87, line 76: I guess this is how we are enforcing that all fields have been populated, right ?

The only downside is that we would only catch this at runtime. IDEs would never show a warning that a field is missing while we are developing.

kompfner: > I guess this is how we are enforcing that all fields have been populated, right ?

Correct!

> The only downside is that we would only catch this at runtime.

I figured this would be worth the runtime v development-time downside because:
- It happens *immediately* at runtime, so the feedback cycle is fast
- The alternative, I believe, is doubling the settings classes, as discussed yesterday (storage classes v delta classes), with the downsides that that brings

filipi87 (approved):
LGTM! One more step on refactoring all the settings. 🚀

PR #3813, opened 2026-02-24
Fix STT TTFB metrics for Soniox and AWS Transcribe

markbackman:
## Summary

- Fixed STT TTFB (time-to-final-segment) metrics not being reported for `SonioxSTTService` and `AWSTranscribeSTTService`
- Regression from 859cd7c9 which refactored TTFB measurement to use the base class `start_ttfb_metrics`/`stop_ttfb_metrics` — these are gated behind `can_generate_metrics()`, which defaults to `False`. Soniox and AWS Transcribe never overrode it, so TTFB calls became silent no-ops.
- Adds `can_generate_metrics() -> True` to both services, matching all other STT service implementations

## Testing

- [ ] Verify Soniox STT reports TTFB metrics after VAD stop
- [ ] Verify AWS Transcribe STT reports TTFB metrics after VAD stop

### DIFF
--- a/changelog/3813.fixed.md
+++ b/changelog/3813.fixed.md
@@ -1,0 +1,1 @@
+- Fixed STT TTFB metrics not being reported for `SonioxSTTService` and `AWSTranscribeSTTService` due to missing `can_generate_metrics()` override.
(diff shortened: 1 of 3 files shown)


aconchillo (approved):

aconchillo:
LGTM

PR #3814, opened 2026-02-24
Fixed an issue where the TTS providers did not close the context after the audio context finished processing all audio.

filipi87:
## Summary                                                                                                           
                              
- Added two new lifecycle callbacks to `AudioContextTTSService`: `on_audio_context_interrupted()` and `on_audio_context_completed()`
- Fixed a bug where TTS providers (AsyncAI, ElevenLabs, Inworld, Rime) did not close or clean up their server side audio contexts after normal playback completion, only on interruption
- Refactored all `AudioContextTTSService` based providers to override the new callbacks instead of `_handle_interruption()`, making provider-specific cleanup cleaner and more explicit
- `AudioContextTTSService` subclasses that previously overrode `_handle_interruption()` for cleanup should now override `on_audio_context_interrupted()` instead. The old approach still works but the new callbacks are the recommended pattern.

### DIFF
--- a/src/pipecat/services/resembleai/tts.py
+++ b/src/pipecat/services/resembleai/tts.py
@@ -18,13 +18,11 @@ from pipecat.frames.frames import (
     EndFrame,
     ErrorFrame,
     Frame,
-    InterruptionFrame,
     StartFrame,
     TTSAudioRawFrame,
     TTSStartedFrame,
     TTSStoppedFrame,
 )
-from pipecat.processors.frame_processor import FrameDirection
 from pipecat.services.settings import NOT_GIVEN, TTSSettings, _NotGiven
 from pipecat.services.tts_service import AudioContextTTSService
 from pipecat.utils.tracing.service_decorators import traced_tts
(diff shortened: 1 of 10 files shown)


ianbbqzy (approved):

src/pipecat/services/resembleai/tts.py
```diff
@@ -245,15 +243,18 @@ def _get_websocket(self):
...
+    async def on_audio_context_completed(self, context_id: str):
+        """Stop metrics after the Resemble AI context finishes playing.
+
+        No close message is needed: Resemble AI signals completion with an
+        ``audio_end`` message (handled in ``_process_messages``), after which
+        the server-side context is already closed.
         """
-        await super()._handle_interruption(frame, direction)
-        await self.stop_all_metrics()
+        pass
```
markbackman, line 259: Why `pass` here?

markbackman (commented):
Looks good. Just one question.

src/pipecat/services/resembleai/tts.py
filipi87: Just to state explicitly that we are not doing anything in this case.
Since `stop_ttfb_metrics` is already invoked inside `audio_end`, I believe there’s nothing additional we need to do here.
What do you think?

markbackman: Makes sense.

markbackman (approved):
LGTM

PR #3816, opened 2026-02-24
Move skills to pipecat-ai/skills repo, add README instructions

aconchillo:

Remove bundled Claude Code skills (changelog, cleanup, code-review, docstring, pr-description, pr-submit) that now live in https://github.com/pipecat-ai/skills. Add a section to the README with installation instructions. The update-docs skill remains as it is specific to this repository.


kompfner:
Some of these skills to me really seem like they're more geared specifically towards working _on_ Pipecat, as opposed to what I assume the new repo is for, which is for working _with_ (i.e. using) Pipecat (this can include general-purpose skills, too).

/changelog: a fragment-style changelog like we have in this repo is not that common, so it feels like it should stay in this repo.

Also, /cleanup is pretty specific to working on Pipecat services. Maybe there's an argument to be made that external-to-this-repo community services will be more and more common, but even then I think it ought to be somehow distinguished from the set of skills generally useful for those building on Pipecat.

/docstring is also filled with some pretty project-specific conventions, so might be more at home here.

kompfner:
> this can include general-purpose skills

Now that I think about this more, I don't actually think we ought to publish extremely general-purpose skills like code review and PR management skills in a Pipecat-specific repo...in a world where more and more skills are available and people are loading up their own sets of skills to customize their own Claude Code experience, it feels like muddying what our Pipecat skills repo is for, and risks stepping on other folks' toes by duplicating their skills. At the very least, in that repo we should separate the skills that are specific to working with PIpecat from general-purpose stuff (though I think my vote would be to clear that repo of general-purpose stuff entirely, leaving it very clear what the point of that repo is).

PR #3817, opened 2026-02-24
Flatten `LiveOptions` into individual fields on `DeepgramSTTSettings`…

kompfner:
… and `DeepgramSageMakerSTTSettings` for backward-compatible dict-style updates via `STTUpdateSettingsFrame`; during the big service settings refactor, we accidentally got rid of the ability to update individual `LiveOptions` fields with a sparse update

### DIFF
--- a/examples/foundational/55a-update-settings-deepgram-sagemaker-stt.py
+++ b/examples/foundational/55a-update-settings-deepgram-sagemaker-stt.py
@@ -112,6 +112,13 @@ async def run_bot(transport: BaseTransport, runner_args: RunnerArguments):
             STTUpdateSettingsFrame(delta=DeepgramSageMakerSTTSettings(language=Language.ES))
         )
 
+        # Old-style dict update (for backward-compat testing):
+        # await asyncio.sleep(10)
+        # logger.info("Updating Deepgram SageMaker STT settings via dict: punctuate=False, filler_words=True")
+        # await task.queue_frame(
+        #     STTUpdateSettingsFrame(settings={"punctuate": False, "filler_words": True})
+        # )
+
     @transport.event_handler("on_client_disconnected")
     async def on_client_disconnected(transport, client):
         logger.info(f"Client disconnected")
--- a/examples/foundational/55a-update-settings-deepgram-sagemaker-stt.py
+++ b/examples/foundational/55a-update-settings-deepgram-sagemaker-stt.py
@@ -7,6 +7,7 @@
 import asyncio
 import os
 
+from deepgram import LiveOptions
 from dotenv import load_dotenv
 from loguru import logger
 
@@ -106,10 +107,16 @@ async def run_bot(transport: BaseTransport, runner_args: RunnerArguments):
         messages.append({"role": "system", "content": "Please introduce yourself to the user."})
         await task.queue_frames([LLMRunFrame()])
 
+        # NOTE: after this change, the bot will only respond if you speak Spanish
         await asyncio.sleep(10)
-        logger.info("Updating Deepgram SageMaker STT settings: language=es")
+        logger.info("Updating Deepgram SageMaker STT settings: language=es, punctuate=False")
         await task.queue_frame(
-            STTUpdateSettingsFrame(delta=DeepgramSageMakerSTTSettings(language=Language.ES))
+            STTUpdateSettingsFrame(
+                delta=DeepgramSageMakerSTTSettings(
+                    language=Language.ES,
+                    live_options=LiveOptions(punctuate=False),
+                )
+            )
         )
 
         # Old-style dict update (for backward-compat testing):
--- a/examples/foundational/55a-update-settings-deepgram-stt.py
+++ b/examples/foundational/55a-update-settings-deepgram-stt.py
@@ -106,6 +106,13 @@ async def run_bot(transport: BaseTransport, runner_args: RunnerArguments):
             STTUpdateSettingsFrame(delta=DeepgramSTTSettings(language=Language.ES))
         )
 
+        # Old-style dict update (for backward-compat testing):
+        # await asyncio.sleep(10)
+        # logger.info("Updating Deepgram STT settings via dict: punctuate=False, filler_words=True")
+        # await task.queue_frame(
+        #     STTUpdateSettingsFrame(settings={"punctuate": False, "filler_words": True})
+        # )
+
     @transport.event_handler("on_client_disconnected")
     async def on_client_disconnected(transport, client):
         logger.info(f"Client disconnected")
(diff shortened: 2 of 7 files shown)


examples/foundational/55a-update-settings-deepgram-sagemaker-stt.py
```diff
@@ -112,6 +112,13 @@ async def on_client_connected(transport, client):
             STTUpdateSettingsFrame(delta=DeepgramSageMakerSTTSettings(language=Language.ES))
         )
 
+        # Old-style dict update (for backward-compat testing):
+        # await asyncio.sleep(10)
+        # logger.info("Updating Deepgram SageMaker STT settings via dict: punctuate=False, filler_words=True")
+        # await task.queue_frame(
+        #     STTUpdateSettingsFrame(settings={"punctuate": False, "filler_words": True})
+        # )
```
filipi87, line 127: Do we need to keep this here or can we remove ?

examples/foundational/55a-update-settings-deepgram-stt.py
```diff
@@ -106,6 +106,13 @@ async def on_client_connected(transport, client):
             STTUpdateSettingsFrame(delta=DeepgramSTTSettings(language=Language.ES))
         )
 
+        # Old-style dict update (for backward-compat testing):
+        # await asyncio.sleep(10)
+        # logger.info("Updating Deepgram STT settings via dict: punctuate=False, filler_words=True")
+        # await task.queue_frame(
+        #     STTUpdateSettingsFrame(settings={"punctuate": False, "filler_words": True})
+        # )
```
filipi87, line 121: Same here.

filipi87 (commented):
I am not sure about this one. After your other [PR](https://github.com/pipecat-ai/pipecat/pull/3834), we are going to start receiving the `settings` inside `__init__`, and we are also receiving `live_options` there as well.

It feels like it should be one or the other. Using `live_options` had the advantage of being explicit about which fields the Deepgram API supports. Now we’re kind of remapping those fields into `settings`.

Do we have any other service where we follow this same pattern ?

examples/foundational/55a-update-settings-deepgram-sagemaker-stt.py
kompfner: I added this commented-out stuff just a manual testing aid. Doesn't need to be here long-term but is helpful to me in the near term at least. Deepgram is a particularly tricky case in terms of back-compat due to the nested LiveOptions field.

kompfner:
> we are going to start receiving the settings inside __init__, and we are also receiving live_options there as well

Not sure I follow. The Deepgram STT services take a user-provided `LiveOptions`, which specifies the initial settings, but it never directly takes a settings object.

> Using live_options had the advantage of being explicit about which fields the Deepgram API supports

Hm, this is a good point. Letting users pass in a 3rd-party `LiveOptions` object directly makes it so that they'll get type hints for all the options available to them...

The problem is that these services *used* to always support passing in individual fields of a `LiveOptions` as a dictionary (like `{"punctuate": False, "filler_words": True}`) as a "delta".

Without the changes in this PR, we won't preserve that behavior and our next release will be a breaking change. If we kept the settings object as-is, so the public API is:
```
STTUpdateSettingsFrame(delta=DeepgramSTTSettings(live_options=LiveOptions(...)))
```
then I think it might be surprising/unexpected to the user if `live_options` _was_ treated as a delta, and not swapped wholesale. Do you agree? Maybe we could set expectations through appropriate commenting?

The alternative would be to abandon that behavior (supporting per-field updates) and either:
1. be OK with breakage of backward compatibility, or
2. come up with a fancier solution so that:
  a. newer usage swaps out `live_options` wholesale (`STTUpdateSettingsFrame(delta=DeepgramSTTSettings(live_options=...)`)
  b. legacy usage updates fields piecemeal (`STTUpdateSettingsFrame(settings={"punctuate": False})`)

WDYT?

kompfner:
> I think it might be surprising/unexpected to the user if live_options was treated as a delta, and not swapped wholesale. Do you agree? Maybe we could set expectations through appropriate commenting?

Let's actually go ahead and collapse settings back into one `live_options` field, treat it as a delta, and just add appropriate commenting to make it crystal clear that that's how it works.

The observation pushing us in this direction: `live_options` is *already* treated as a delta in the initializer, as a delta from a bag of default live options.

tests/test_settings.py
```diff
@@ -21,6 +18,18 @@
     is_given,
 )
 
+try:
+    from deepgram import LiveOptions
```
kompfner: In CI this module isn't available. This is a quick workaround.

kompfner:
> Let's actually go ahead and collapse settings back into one `live_options` field, treat it as a delta, and just add appropriate commenting to make it crystal clear that that's how it works.

Updated @filipi87 @markbackman. Mark, I'll need your help again tomorrow testing the SageMaker variant.

tests/test_settings.py
kompfner: Actually just going ahead and updating what we install in CI for the test run.

src/pipecat/services/deepgram/stt.py
```diff
@@ -292,6 +373,21 @@ async def run_stt(self, audio: bytes) -> AsyncGenerator[Frame, None]:
...
+
+        Returns:
+            A fully-populated ``LiveOptions`` ready for the Deepgram SDK.
+        """
+        opts: dict[str, Any] = self._settings.live_options.to_dict()
+
+        # Overlay model/language from top-level settings and sample_rate from service.
+        opts["model"] = self._settings.model
+        opts["language"] = self._settings.language
+        opts["sample_rate"] = self.sample_rate
```
filipi87: Could this be out of sync between what we have in self._settings.live_options and what we have here? Is that why we need to do this?

filipi87 (approved):
I think the approach you implemented makes sense. 🚀

src/pipecat/services/deepgram/stt.py
kompfner: I'm actually now wondering if these lines are necessary (at least the model/language part of it), since we already do the work in `DeepgramSTTSettings.apply_update` to keep the top-level settings model and language fields in sync with the values in live_options...

kompfner: Fixed. Good catch.

kompfner:
> I think the approach you implemented makes sense. 🚀

I'm going to spend a moment exploring whether this logic can be tidied up a bit before merging.

kompfner:
> > I think the approach you implemented makes sense. 🚀
> 
> I'm going to spend a moment exploring whether this logic can be tidied up a bit before merging.

OK, I think the logic cleaned up quite nicely: https://github.com/pipecat-ai/pipecat/pull/3817/changes/faed775d9075064af3804be89f94322c36af2c8e

PR #3819, opened 2026-02-24
Make `ServiceUpdateSettingsFrame` uninterruptible—settings updates ar…

kompfner:
…e generally independent of specific utterances.

Before this change, settings updates were often not applied. For example, a `TTSUpdateSettingsFrame` queued while the bot was speaking would only have an effect at the end of the bot's reply, and any interruption before the end of the reply would "cancel" the update.

This change was discussed in [this PR comment](https://github.com/pipecat-ai/pipecat/pull/3714#issuecomment-3923470697).


filipi87 (approved):
Agreed! 🚀

PR #3820, opened 2026-02-24
Fix breakage when using a generic settings update (e.g. a `TTSSetting…

kompfner:
…s`) instead of a more specific one (e.g. a `RimeTTSSettings`). Both should work, assuming you're only changing fields present in the generic settings.

@filipi87 raised this issue [here](https://github.com/pipecat-ai/pipecat/pull/3714#discussion_r2827940250).


filipi87 (approved):
LGTM!

PR #3776, opened 2026-02-19
Refactor STT TTFB metrics to use base class start/stop pattern

aconchillo:
## Summary

- Added `start_time` and `end_time` parameters to `start_ttfb_metrics()`, `stop_ttfb_metrics()`, `start_processing_metrics()`, and `stop_processing_metrics()` in `FrameProcessor` and `FrameProcessorMetrics`, allowing custom timestamps for metrics measurement.
- Refactored `STTService` to use base class TTFB metrics instead of custom `_emit_stt_ttfb_metric()` and manual timestamp tracking (`_speech_end_time`, `_last_transcription_time`).
- STT TTFB now follows chronological start→stop ordering: `start_ttfb_metrics(start_time=speech_end_time)` at VAD stop, `stop_ttfb_metrics()` at transcription finalize or timeout.
- Improved metrics debug log formatting (`.3f` precision, `s` suffix).

## Testing

- `uv run pytest tests/`

PR #3822, opened 2026-02-25
Fix STT TTFB timeout measuring to use transcript arrival time

markbackman:
## Summary

- Fixed STT TTFB metrics reporting timeout expiry time instead of actual transcript arrival time. PR #3776 replaced manual timestamp tracking with `stop_ttfb_metrics()` in the timeout handler, but without an `end_time` it used `time.time()` at timeout expiry — producing TTFB = timeout + stop_secs (~2.2s) instead of the actual transcript latency.
- Restored `_last_transcript_time` tracking so the timeout handler measures to when the transcript arrived, and skips reporting if none arrived.

## Test plan

- [x] Verify STT TTFB metrics report actual transcript latency, not timeout duration
- [x] Verify no TTFB is reported when no transcript arrives before timeout

### DIFF
--- a/changelog/3822.fixed.md
+++ b/changelog/3822.fixed.md
@@ -1,0 +1,1 @@
+- Fixed STT TTFB metrics measuring timeout expiry time instead of actual transcript arrival time.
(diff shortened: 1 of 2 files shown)


filipi87 (approved):
Makes sense.

PR #3823, opened 2026-02-25
Fix assistant aggregator incorrectly aggregating transcription frames

### LINKED ISSUE #3638: LLMAssistantAggregator aggregates InterimTranscriptionFrames

kollaikal-rupesh:
## Description

Fixes #3638

`TranscriptionFrame` and `InterimTranscriptionFrame` were caught by the generic `TextFrame` handler in both `LLMAssistantContextAggregator` and `LLMAssistantAggregator`, causing user speech to appear as assistant messages in the LLM context.

Added explicit handlers before the `TextFrame` check to push transcription frames through instead of aggregating them. Removed the redundant guard in the universal aggregator's `_handle_text()` that was silently dropping these frames.

## Test plan

- Added `test_transcription_frames_not_aggregated` across all 5 provider aggregators (OpenAI, Anthropic, AWS, Google, Universal)
- Added separate `TranscriptionFrame` and `InterimTranscriptionFrame` tests for the universal aggregator
- All 151 tests pass

### DIFF
--- a/src/pipecat/processors/aggregators/llm_response.py
+++ b/src/pipecat/processors/aggregators/llm_response.py
@@ -901,6 +901,8 @@ class LLMAssistantContextAggregator(LLMContextResponseAggregator):
             await self._handle_llm_start(frame)
         elif isinstance(frame, LLMFullResponseEndFrame):
             await self._handle_llm_end(frame)
+        elif isinstance(frame, (TranscriptionFrame, InterimTranscriptionFrame)):
+            await self.push_frame(frame, direction)
         elif isinstance(frame, TextFrame):
             await self._handle_text(frame)
         elif isinstance(frame, LLMRunFrame):
--- a/src/pipecat/processors/aggregators/llm_response_universal.py
+++ b/src/pipecat/processors/aggregators/llm_response_universal.py
@@ -879,6 +879,8 @@ class LLMAssistantAggregator(LLMContextAggregator):
             await self._handle_llm_start(frame)
         elif isinstance(frame, LLMFullResponseEndFrame):
             await self._handle_llm_end(frame)
+        elif isinstance(frame, (TranscriptionFrame, InterimTranscriptionFrame, TranslationFrame)):
+            await self.push_frame(frame, direction)
         elif isinstance(frame, TextFrame):
             await self._handle_text(frame)
         elif isinstance(frame, LLMThoughtStartFrame):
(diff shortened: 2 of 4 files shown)


markbackman:
@rupesh-svg thanks for the proposal. I think we want a slightly different direction. Check out https://github.com/pipecat-ai/pipecat/pull/3825. Closing this in favor of 3825.

kollaikal-rupesh:
Makes sense, consuming at the user aggregator is cleaner than passing through the assistant side. Thanks for the alternative approach.

Issue #1006, opened 2025-01-16
Wrong API Key or Voice ID in ElevenLabs doesn't end the pipeline


Vaibhav-Lodha:
# Description - BUG
For any error in ElevenLabs, the pipeline doesn't end and keeps on logging the error on console.

If reporting a bug, please fill out the following:

### Environment
- pipecat-ai version: 0.0.52
- python version: 3.12
- OS: macos/linux

### Issue description
ElevenLabs websocket connection keep on trying to either connect via `_disconnect_websocket` for the first time but later attempts are made via `_keepalive_task_handler`, even in cases where the config were wrong and hence the process is forever running with no system level execptions.

### Repro steps
List the steps to reproduce the issue.
Put a wrong api key or voice id

### Expected behavior
Handle config related issues and end the pipeline

### Logs
```
2025-01-16 15:01:26.726 | WARNING  | pipecat.services.elevenlabs:_reconnect_websocket:381 - ElevenLabsFlashTTSService reconnecting (attempt: 2)
2025-01-16 15:01:26.727 | DEBUG    | pipecat.processors.metrics.frame_processor_metrics:stop_ttfb_metrics:50 - ElevenLabsFlashTTSService TTFB: 3.5634312629699707
2025-01-16 15:01:26.727 | DEBUG    | pipecat.services.elevenlabs:_disconnect_websocket:356 - Disconnecting from ElevenLabs
2025-01-16 15:01:26.727 | ERROR    | pipecat.services.elevenlabs:_disconnect_websocket:363 - ElevenLabsFlashTTSService error closing websocket: received 1008 (policy violation) A voice with voice_id s does not exist.; then sent 1008 (policy violation) A voice with voice_id s does not exist.
2025-01-16 15:01:26.727 | DEBUG    | pipecat.services.elevenlabs:_connect_websocket:318 - Connecting to ElevenLabs
2025-01-16 15:01:27.686 | DEBUG    | pipeline.event_handlers:on_webrtc_app_message_callback:41 - Received app message: data={'event': 'request-chat-history'} - participant='ec3075b1-183a-4fc3-be09-d3abbb498897'
2025-01-16 15:01:30.568 | ERROR    | pipecat_overrides.services.elevenlabs:_keepalive_task_handler:317 - ElevenLabsFlashTTSService exception: received 1008 (policy violation) A voice with voice_id s does not exist.; then sent 1008 (policy violation) A voice with voice_id s does not exist.
[...]

PR #3824, opened 2026-02-25
Fix WebsocketService infinite reconnection loop on immediate server close

### LINKED ISSUE #3711: WebsocketService enters infinite reconnection loop when server accepts handshake but immediately closes connection
### pipecat version

0.0.100

### Python version

3.12

### Operating System

macOS 26.2 

## Bug

`WebsocketService._try_reconnect` enters an infinite loop when the server accepts the WebSocket handshake but immediately closes the connection with an error (e.g. `1008 policy violation - Invalid API key`).

## What happens

The reconnection succeeds (handshake + ping verification pass), so `_try_reconnect` returns `True` on attempt 1. The `_receive_task_handler` loop resumes, `_receive_messages` runs, the server immediately kicks the connection with a `ConnectionClosedError`, and the cycle repeats — forever.

```
reconnected successfully on attempt 1
connection closed, but with an error: received 1008 (policy violation) Invalid API key
reconnecting, attempt 1
reconnected successfully on attempt 1
connection closed, but with an error: received 1008 (policy violation) Invalid API key
reconnecting, attempt 1
... (infinite)
```

## Why it happens

1. `_try_reconnect` only counts **failed handshakes** toward its 3-attempt limit. If the handshake succeeds, the counter resets.
2. `_verify_connection` only checks `websocket.ping()` — which succeeds in the brief window before the server sends its close frame.
3. No `ErrorFrame` is ever emitted — neither non-fatal (only sent on failed reconnect attempts) nor fatal (only sent after 3 exhausted attempts). The `on_pipeline_error` handler is never called.

## Impact

- The service silently burns CPU reconnecting ~1x/second indefinitely
- No `ErrorFrame` (fatal or non-fatal) is ever propagated to the pipeline
- Downstream consumers (e.g. `on_pipeline_error` handlers) have no way to detect or react to this state
- Affects all services inheriting from `WebsocketService` (ElevenLabs TTS, etc.)

## Suggested fixes

Any of the following would address this:

1. **Track connection stability**: If a connection lasts less than N seconds before being closed, count it as a failed attempt toward the retry limit.
2. **Respect non-recoverable close codes**: Close codes like `1008` (policy violation), `1003` (unsupported data), `4xxx` (application-level) should not trigger reconnection — they indicate permanent errors.
3. **Global cycle detection**: Track the number of reconnection cycles within a time window (e.g. 5 reconnects in 30 seconds = fatal), regardless of whether individual attempts "succeed."

- **Affected services**: ElevenLabs TTS,  (any `WebsocketService` subclass)

### Reproduction steps

1. Provide wrong token to one of the services (TTS...)

### Expected behavior

After a small number of failed reconnection cycles, `WebsocketService` should emit a fatal `ErrorFrame` and stop retrying — allowing the pipeline's `on_pipeline_error` handler to detect the failure and take action (e.g. redirect to a human agent).

### Actual behavior

The service enters an infinite reconnection loop (~1 reconnect/second), never emits any `ErrorFrame`, and the pipeline has no way to detect or react to the failure. The loop continues until the process is manually killed.

### Logs

```shell

```

kollaikal-rupesh:
Fixes #3711

`_try_reconnect` only counts failed handshakes toward its retry limit. When a server accepts the handshake but immediately closes (e.g., `1008` for invalid API key), reconnection "succeeds" every time and the loop never exits.

### Changes

- Detect non-recoverable close codes (1002, 1003, 1008, 1009, 1010, 1015, 4000-4999) and emit a fatal `ErrorFrame` without retrying
- Track rapid failures: if a connection drops within 5s of establishment, count it toward a 3-strike limit before emitting a fatal `ErrorFrame`
- Reset failure tracking on `_connect()`

### DIFF
--- a/src/pipecat/services/websocket_service.py
+++ b/src/pipecat/services/websocket_service.py
@@ -7,6 +7,7 @@
 """Base websocket service with automatic reconnection and error handling."""
 
 import asyncio
+import time
 from abc import ABC, abstractmethod
 from typing import Awaitable, Callable, Optional
 
@@ -17,6 +18,21 @@ from websockets.protocol import State
 
 from pipecat.frames.frames import ErrorFrame
 from pipecat.utils.network import exponential_backoff_time
+
+# Close codes that indicate permanent errors where reconnection will never succeed.
+# See RFC 6455 Section 7.4.1 and https://www.iana.org/assignments/websocket/websocket.xhtml
+_NON_RECOVERABLE_CLOSE_CODES = {
+    1002,  # Protocol error
+    1003,  # Unsupported data
+    1008,  # Policy violation (e.g., invalid API key)
+    1009,  # Message too big
+    1010,  # Mandatory extension
+    1015,  # TLS handshake failure
+}
+
+# Minimum connection duration (seconds) to consider a reconnection "stable".
+# If the connection drops faster than this, it counts toward the rapid failure limit.
+_MIN_STABLE_CONNECTION_SECS = 5.0
 
 
 class WebsocketService(ABC):
@@ -38,6 +54,8 @@ class WebsocketService(ABC):
         self._reconnect_on_error = reconnect_on_error
         self._reconnect_in_progress: bool = False
         self._disconnecting: bool = False
+        self._rapid_failure_count: int = 0
+        self._last_connect_time: float = 0.0
 
     async def _verify_connection(self) -> bool:
         """Verify the websocket connection is active and responsive.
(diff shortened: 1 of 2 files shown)


smokyabdulrahman:
Hey @markbackman 👋 ,
We would really appreciate a review here, as this is happening to us also on production.

markbackman:
I've been working through older issues and was thinking about #1006 last week. I didn't see this PR until your ping. I just posted my proposal for this issue in #4201.

The two main departures from this PR are:
- No code allowlist: I'm worried about this assumption as different services uses these codes in different ways.
- No fatal errors: Pipecat has a ServiceSwitcher and emits ErrorFrames so that applications can failover to backup services.

Check out the PR I posted for a similar approach to the problem. Thanks @kollaikal-rupesh for the hard work 🙇

PR #3826, opened 2026-02-25
Add user turn gate hooks for better interruptability

johnwhoyou:

This PR adds LLM turn strategy gating hooks for user turn start/stop decisions without introducing LLM dependencies. It introduces a shared gate context, timeout/error handling (deny on timeout, configurable allow-on-error), logging, and serialized transitions to avoid duplicate events. It also adds tests for start/stop denial and timeout/error fallback, and includes a changelog fragment (`changelog/3826.added.md`).

#### Usage example
In our production app we use LLM-based gating to decide when to interrupt on interim transcripts and when to close on final transcripts (currently implemented via `LLMInterimTurnStartStrategy` + `LLMTurnStopStrategy` in our repo). This PR lets us implement the same pattern using core hooks:

```python
async def start_gate(ctx):
    # Gate interruptions on interim text while the bot is speaking.
    if ctx.transcription_is_interim and ctx.transcription_text:
        return await llm_is_complete_interrupt(ctx.transcription_text)
    return True

async def stop_gate(ctx):
    # Gate end-of-turn on final transcriptions.
    if ctx.transcription_is_interim:
        return False
    if ctx.transcription_text:
        return await llm_is_complete_turn(ctx.transcription_text)
    return False

user_turn_strategies = UserTurnStrategies(
    start=[VADUserTurnStartStrategy(), TranscriptionUserTurnStartStrategy()],
    stop=[TurnAnalyzerUserTurnStopStrategy(turn_analyzer=LocalSmartTurnAnalyzerV3())],
    start_gate=start_gate,
    stop_gate=stop_gate,
)
```

## Test Plan
- [x] `pytest tests/test_user_turn_gates.py -v`

### DIFF
--- a/changelog/3826.added.md
+++ b/changelog/3826.added.md
@@ -1,0 +1,1 @@
+- Added LLM turn gate hooks to control user turn start/stop decisions with timeouts and error handling.


markbackman:
Hi, thanks for the proposal. I don't think we're ready to add this additional customization layer. We've made the user start and user stop strategies available to allow this level of control. Rather than introduce this layer, we would like developers to write custom strategies to handle different start and stop strategies.

I'm going to close out this PR, but thanks again for offering it. You're free to fork Pipecat and add this to your own project if it's a good fit for what you're looking to accomplish.

PR #3827, opened 2026-02-25
Remove unnecessary `_model` ivar from `GeminiTTSService`, using `_set…

kompfner:
…tings.model` instead

Based on @filipi87's comment [here](https://github.com/pipecat-ai/pipecat/pull/3714#discussion_r2828400468).


src/pipecat/services/google/tts.py
```diff
@@ -621,6 +621,7 @@ def __init__(
             google_style=params.google_style,
             voice=voice_id,
         )
+        self._sync_model_name_to_metrics()
```
filipi87: I believe that once we send the setting through `__init__`, we’ll be able to remove this, right ?

filipi87 (approved):
LGTM!

PR #3828, opened 2026-02-25
Skip empty audio frames after filter buffering

### LINKED ISSUE #3517: Use of RNNoiseFilter throws warnings regarding empty audio frames
### pipecat version

0.0.100

### Python version

3.12.12 / 3.12.7

### Operating System

macOS 15.5 (24F74) / Amazon Linux 2023

### Issue description

When running the server with `WebsocketServerParams(audio_in_enabled=True, audio_in_filter=RNNoiseFilter())`, warning messages are displayed like `Empty audio frame received for STT service`, despite the setup working as expected. These warnings are misleading and may cause confusion about the actual functionality.

We didn't test using other transports than the WebsocketTransport in combination with the TwilioSerializer.

### Reproduction steps

1. Start the server using the following parameters:
   ```python
   WebsocketServerParams(
       audio_in_enabled=True,
       audio_in_filter=RNNoiseFilter()
   )
   ```
3. Use DeepgramSTTService as STT not sure if related)
4. Connect an audio stream/client as usual.
5. Observe the logs and warning output.

### Expected behavior

No `Empty audio frame received for STT service` warnings should be displayed when the feature is functioning correctly.

### Actual behavior

Warning messages are shown in the logs referencing possible issues with `WARNING  | pipecat.services.stt_service:process_audio_frame:184 | Empty audio frame received for STT service: DeepgramSTTService#4 160`, even though the filter is working as intended.

### Logs

```shell
WARNING  | pipecat.services.stt_service:process_audio_frame:184 | Empty audio frame received for STT service: DeepgramSTTService#4 160
```

markbackman:
## Summary

- Audio filters like `RNNoiseFilter`, `KrispVivaFilter`, and `AICFilter` return empty bytes (`b""`) while buffering audio to accumulate their required frame size. These empty frames were flowing downstream through VAD and into the STT service, causing misleading "Empty audio frame received for STT service" warnings.
- Skip the frame in `BaseInputTransport` when the audio filter returns empty audio, preventing unnecessary processing in VAD and downstream processors.

## Fixes

- Fixes #3517

### DIFF
--- a/src/pipecat/transports/base_input.py
+++ b/src/pipecat/transports/base_input.py
@@ -424,6 +424,11 @@ class BaseInputTransport(FrameProcessor):
                 if self._params.audio_in_filter:
                     frame.audio = await self._params.audio_in_filter.filter(frame.audio)
 
+                # Skip frames with no audio data (e.g. filter is buffering).
+                if not frame.audio:
+                    self._audio_in_queue.task_done()
+                    continue
+
                 ###################################################################
                 # DEPRECATED.
                 #
--- a/changelog/3828.fixed.md
+++ b/changelog/3828.fixed.md
@@ -1,0 +1,1 @@
+- Fixed misleading "Empty audio frame received for STT service" warnings when using audio filters (e.g. `RNNoiseFilter`, `KrispVivaFilter`, `AICFilter`) that buffer audio internally.


src/pipecat/transports/base_input.py
```diff
@@ -423,6 +423,11 @@ async def _audio_task_handler(self):
                 # If an audio filter is available, run it before VAD.
                 if self._params.audio_in_filter:
                     frame.audio = await self._params.audio_in_filter.filter(frame.audio)
+                    # Skip frames with no audio data (e.g. filter is
+                    # buffering).
+                    if not frame.audio:
+                        self._audio_in_queue.task_done()
+                        continue
```
aconchillo: we can actually move this after the audio in filter.

```
                # If an audio filter is available, run it before VAD.
                if self._params.audio_in_filter:
                    frame.audio = await self._params.audio_in_filter.filter(frame.audio)

                # Skip frames with no audio data (e.g. filter is buffering).
                if not frame.audio:
                    self._audio_in_queue.task_done()
                    continue
```

probably more generic.

aconchillo (approved):

aconchillo:
LGTM. Added a comment.

src/pipecat/transports/base_input.py
markbackman: 👍

PR #3831, opened 2026-02-25
Replace VirtualCameraDevice with CustomVideoTrack + custom video track support

aconchillo:
## Summary

NOTE: Requires upcoming daily-python 0.24.0

- Replaced `VirtualCameraDevice` with `CustomVideoSource`/`CustomVideoTrack` for the default camera output, mirroring how audio already uses `CustomAudioSource`/`CustomAudioTrack`
- Added custom video destination support via `video_out_destinations` in `DailyParams`, enabling multiple simultaneous video tracks
- New methods on `DailyTransportClient`: `register_video_destination`, `add_custom_video_track`, `remove_custom_video_track`
- Video frames are routed to the correct track based on `transport_destination`, matching the audio pattern
- Added foundational example (`56-custom-video-track.py`) demonstrating dual video track output with a gradient pattern and blue-tinted variant

## Testing

- `uv run pytest` — all existing tests pass (677 passed)
- `uv run ruff check` and `uv run ruff format --check` — clean
- Run `python examples/foundational/56-custom-video-track.py -t daily` to verify dual track output

[Screencast from 2026-02-25 00-24-44.webm](https://github.com/user-attachments/assets/70118e52-7af1-47cf-a45a-81d94919ebb6)

### DIFF
--- a/changelog/3831.changed.md
+++ b/changelog/3831.changed.md
@@ -1,0 +1,1 @@
+- Daily transport now uses `CustomVideoSource`/`CustomVideoTrack` instead of `VirtualCameraDevice` for the default camera output, mirroring how audio already works with `CustomAudioSource`/`CustomAudioTrack`.
--- a/examples/foundational/56-custom-video-track.py
+++ b/examples/foundational/56-custom-video-track.py
@@ -1,0 +1,210 @@
+#
+# Copyright (c) 2024-2026, Daily
+#
+# SPDX-License-Identifier: BSD 2-Clause License
+#
+
+"""Example demonstrating custom video tracks output with Daily transport.
+
+This example outputs two video track simultaneously:
+  - The default camera track with an animated color gradient pattern.
+  - A custom "blue" track with the same pattern but with a blue tint applied.
+
+The pattern generator pushes frames to the default camera. A second processor
+(BlueTintProcessor) duplicates each frame, applies a blue tint, and pushes it
+to the "blue" custom video destination.
+
+Run with: python examples/foundational/56-custom-video-track.py -t daily
+"""
+
+import asyncio
+import math
+import time
+
+import numpy as np
+from loguru import logger
+
+from pipecat.frames.frames import (
+    CancelFrame,
+    EndFrame,
+    Frame,
+    OutputImageRawFrame,
+    StartFrame,
+    SystemFrame,
+)
+from pipecat.pipeline.pipeline import Pipeline
+from pipecat.pipeline.runner import PipelineRunner
+from pipecat.pipeline.task import PipelineTask
+from pipecat.processors.frame_processor import FrameDirection, FrameProcessor
+from pipecat.runner.types import RunnerArguments
+from pipecat.runner.utils import create_transport
+from pipecat.transports.base_transport import BaseTransport
+from pipecat.transports.daily.transport import DailyCustomVideoTrackParams, DailyParams
+
+WIDTH = 320
+HEIGHT = 240
+FPS = 30
+
+transport_params = {
+    "daily": lambda: DailyParams(
+        video_out_enabled=True,
+        video_out_width=WIDTH,
+        video_out_height=HEIGHT,
+        video_out_framerate=FPS,
+        video_out_destinations=["blue"],
+        custom_video_track_params={
+            "blue": DailyCustomVideoTrackParams(
+                width=WIDTH,
+                height=HEIGHT,
+                send_settings={
+                    "maxQuality": "low",
+                    "encodings": {
+                        "low": {
+                            "maxBitrate": 500_000,
+                            "maxFramerate": FPS,
+                        }
+                    },
+                },
+            ),
+        },
+    ),
+}
+
+
+def generate_gradient_frame(width: int, height: int, t: float) -> np.ndarray:
+    """Generate an animated gradient pattern.
+
+    Creates a smooth color gradient that shifts over time using sine waves
+    for each RGB channel at different frequencies.
+    """
+    x = np.linspace(0, 1, width)
+    y = np.linspace(0, 1, height)
+    xv, yv = np.meshgrid(x, y)
+
+    r = ((np.sin(2 * math.pi * (xv + t * 0.3)) + 1) / 2 * 255).astype(np.uint8)
+    g = ((np.sin(2 * math.pi * (yv + t * 0.5)) + 1) / 2 * 255).astype(np.uint8)
+    b = ((np.sin(2 * math.pi * (xv + yv + t * 0.7)) + 1) / 2 * 255).astype(np.uint8)
+
+    return np.stack([r, g, b], axis=-1)
+
+
+class VideoPatternGenerator(FrameProcessor):
+    """Generates an animated gradient pattern and pushes it as video frames."""
+
+    def __init__(self, width: int, height: int, fps: int):
+        super().__init__()
+        self._width = width
+        self._height = height
+        self._fps = fps
+        self._generate_task = None
+
+    async def process_frame(self, frame: Frame, direction: FrameDirection):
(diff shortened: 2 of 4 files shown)


examples/foundational/56-custom-video-track.py
```diff
@@ -0,0 +1,197 @@
+#
+# Copyright (c) 2024-2026, Daily
+#
+# SPDX-License-Identifier: BSD 2-Clause License
+#
+
+"""Example demonstrating custom video tracks with Daily transport.
```
kompfner: Nit: you may want to make it clearer to the reader that we're demonstrating _output_ from the pipeline, not input to the pipeline.

Examples of updates:

- "Example demonstrating custom video track output with Daily transport"
- "This example outputs two video track simultaneously"
- "The pattern generator pushes frames to the default virtual "camera""

I always get turned around when we're talking about "sending" video, or "camera"s (since the user is the only one with an actual camera), so I'm guessing other readers might face similar confusion

examples/foundational/56-custom-video-track.py
```diff
@@ -0,0 +1,197 @@
...
+    async def process_frame(self, frame: Frame, direction: FrameDirection):
+        await super().process_frame(frame, direction)
+
+        if isinstance(frame, StartFrame):
+            await self.push_frame(frame, direction)
+            await self._start()
+        elif isinstance(frame, (EndFrame, CancelFrame)):
+            await self._stop()
+            await self.push_frame(frame, direction)
+        elif isinstance(frame, SystemFrame):
```
kompfner: nit: do we need special handling of system frames v other frames here? they're all getting pushed through regardless

src/pipecat/transports/daily/transport.py
```diff
@@ -626,6 +637,31 @@ async def register_audio_destination(self, destination: str):
...
         self._client.update_publishing({"customAudio": {destination: True}})
 
+    async def register_video_destination(self, destination: str):
+        """Register a custom video destination for multi-track output.
+
+        Args:
+            destination: The destination identifier to register.
+        """
+        self._custom_video_tracks[destination] = await self.add_custom_video_track(destination)
+        send_settings = {
```
kompfner: Isn't one of the possible reasons for having custom tracks for sending, for example, higher-fidelity video in addition to the main "camera" video? Here we're just using the same settings from the main "camera" track without a way to customize.

This actually seems to differ a bit from what we do in `register_audio_destination`, which is to just use the default publishing settings for "customAudio" as opposed to the "microphone" publishing settings specified in `_join`.

Although...wait a minute...
* Do we ever actually make use of the "microphone" publishing settings specified in `_join`? We're always publishing audio as a "customAudio" track, right? (In other words, are the `_join`-specified publishing settings "orphaned"?)
* With the change in this PR, won't we be similarly be "orphaning" the `_join`-specified "camera" publishing settings?

src/pipecat/transports/daily/transport.py
```diff
@@ -1215,6 +1268,51 @@ async def remove_custom_audio_track(self, track_name: str) -> Optional[CallClien
...
+        Args:
+            track_name: Name for the custom video track.
+
+        Returns:
+            The created DailyVideoTrack instance.
+        """
+        future = self._get_event_loop().create_future()
+
+        video_source = CustomVideoSource(
+            self._params.video_out_width,
```
kompfner: See previous comment about whether we ought to have the flexibility to, well, customize custom tracks, as opposed to having them all read a single bag of settings.

changelog/3831.changed.md
filipi87, line 1: Makes a lot of sense. Nice improvement.

src/pipecat/transports/daily/transport.py
```diff
@@ -731,13 +777,14 @@ async def start(self, frame: StartFrame):
                 self._callback_task_handler(self._video_queue),
                 f"{self}::video_callback_task",
             )
-        if self._params.video_out_enabled and not self._camera:
-            self._camera = Daily.create_camera_device(
```
filipi87, line 737: Should we still provide a way to use the virtual camera device instead of a custom video source, as we did for audio? Is it worth it here?

examples/foundational/56-custom-video-track.py
aconchillo: Yes, not needed.

src/pipecat/transports/daily/transport.py
aconchillo: You are right. I think we should use defaults for now, I just updated it. We might add custom video/audio settings in the future.

Regarding `microphone` settings, aren't `sendSettings` applied if we have a custom track on the `microphone`  itself? It's the same stream but with a differen track. I thought those were independent things, but I might be mistaken. Same for the `camera` track.

src/pipecat/transports/daily/transport.py
aconchillo: For audio we only use custom tracks, no more virtual microphone. I don't think it makes sense to keep the virtual camera.

src/pipecat/transports/daily/transport.py
aconchillo: I just went ahead and provided `send_settings` and other per custom track. The `camera` and `microphone` hopefully use their settings even if they point to a custom track id.

src/pipecat/transports/daily/transport.py
aconchillo: The only virtual thing we keep is the speaker which allows us to receive mixed audio from all participants. But you could use `capture_participant_audio(....)`.

src/pipecat/transports/daily/transport.py
aconchillo: Updated

examples/foundational/56-custom-video-track.py
aconchillo: Done.

src/pipecat/transports/daily/transport.py
```diff
@@ -304,15 +305,53 @@ class DailyTranscriptionSettings(BaseModel):
...
+    """Configuration for a custom video track.
+
+    If ``send_settings`` is not provided, the track will use the default video
+    publishing settings (framerate, bitrate, codec, etc.).
+
+    Parameters:
+        width: Video width in pixels.
+        height: Video height in pixels.
+        color_format: Video color format (e.g., "RGB", "RGBA", "BGRA").
+        send_settings: Optional Daily sendSettings dict for this track.
```
kompfner, line 318: Maybe we could reference the docs here? https://reference-python.daily.co/types.html#videopublishingsettings

src/pipecat/transports/daily/transport.py
```diff
@@ -304,15 +305,53 @@ class DailyTranscriptionSettings(BaseModel):
...
+class DailyCustomAudioTrackParams(BaseModel):
+    """Configuration for a custom audio track.
+
+    If ``send_settings`` is not provided, the track will use the default audio
+    publishing settings (bitrate, channel config, etc.).
+
+    Parameters:
+        sample_rate: Audio sample rate in Hz. Defaults to transport's output sample rate.
+        channels: Number of audio channels.
+        send_settings: Optional Daily sendSettings dict for this track.
```
kompfner, line 337: Maybe we could reference the docs here? https://reference-python.daily.co/types.html#audiopublishingsettings

kompfner (approved):
👏  (left some minor suggestions re referencing documentation in docstrings)

src/pipecat/transports/daily/transport.py
aconchillo: Yep! Added

src/pipecat/transports/daily/transport.py
aconchillo: Yep. Added

PR #3834, opened 2026-02-25
Make it so that `AIService` is the exclusive "syncer" of model name t…

kompfner:
…o metrics.

This change has the side-effect of bringing model name to metrics for a number of services that were accidentally omitting it before.

### DIFF
--- a/src/pipecat/services/ai_service.py
+++ b/src/pipecat/services/ai_service.py
@@ -35,14 +35,21 @@ class AIService(FrameProcessor):
     this base infrastructure.
     """
 
-    def __init__(self, **kwargs):
+    def __init__(self, settings: ServiceSettings | None = None, **kwargs):
         """Initialize the AI service.
 
         Args:
+            settings: The runtime-updatable settings for the AI service.
             **kwargs: Additional arguments passed to the parent FrameProcessor.
         """
         super().__init__(**kwargs)
-        self._settings: ServiceSettings = ServiceSettings()  # Here in case subclass doesn't implement more specific settings (hopefully shouldn't happen)
+        self._settings: ServiceSettings = (
+            settings
+            # Here in case subclass doesn't implement more specific settings
+            # (which hopefully should be rare)
+            or ServiceSettings()
+        )
+        self._sync_model_name_to_metrics()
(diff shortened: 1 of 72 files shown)


src/pipecat/services/ai_service.py
```diff
@@ -54,15 +61,12 @@ def _sync_model_name_to_metrics(self):
...
-        model into `super().__init__` and `AIService` can be responsible for
-        syncing its initial value to metrics, just as it's responsible for
-        syncing any updates to its value to metrics via `_update_settings`.
-
         Args:
             model: The name of the AI model to use.
         """
-        self.set_core_metrics_data(MetricsData(processor=self.name, model=self._settings.model))
+        self.set_core_metrics_data(
+            MetricsData(processor=self.name, model=self._settings.model or "")
```
kompfner, line 68: Added back "" as fallback for model name, since I believe that's what we used as a fallback before the big service settings refactor.

kompfner: One question (not necessary to answer for this PR, as it's not related to a regression but rather forward-looking change) is whether we should fall back to "unknown", since that's the string we use when we encounter an unknown model in service_decorators.py. Using it for the core metrics data, too, might be nice for consistency. cc @markbackman

src/pipecat/services/ai_service.py
```diff
@@ -35,14 +35,21 @@ class AIService(FrameProcessor):
     this base infrastructure.
     """
 
-    def __init__(self, **kwargs):
+    def __init__(self, settings: ServiceSettings | None = None, **kwargs):
```
filipi87, line 38: It does feel a lot better passing the setting here. 🔥

filipi87: nit I guess in most places we are using the syntax: `Optional[ServiceSettings] = None`

filipi87 (approved):
LGTM. 🚀

src/pipecat/services/ai_service.py
kompfner: Yes, but for some reason in this file we use `| None` elsewhere (`Optional` is not imported). I decided to fit the norms of the file 🤷

PR #3837, opened 2026-02-25
Add `append_trailing_space` to all Rime websocket services

alts:
This was added in 31daa889e83b960fab79d66b2ab014d930e15a2e, but only to `RimeTTSService`, not to `RimeNonJsonTTSService. Bringing these to parity means that users switching between the two, with the same inputs, have more consistent vocalization behaviors.

(changelog entry to be added once I know the PR number I'm assigned)


markbackman (approved):
LGTM

PR #3840, opened 2026-02-25
Add together STT and TTS to pipecat

blainekasten:
# Summary

Add support for [TogetherAI](https://www.together.ai/) as a service provider for llm, stt and tts inference.

Docs for our APIs can be found here: https://docs.together.ai/reference/audio-transcriptions-realtime#parameter-input-audio-format


examples/foundational/07d-interruptible-togetherai.py
```diff
@@ -0,0 +1,128 @@
+#
```
markbackman: There's an opening at 07e, which you can take. Rename to: `07e-interruptible-togetherai.py`.

src/pipecat/services/together/stt.py
```diff
@@ -0,0 +1,306 @@
...
+                        time_now_iso8601(),
+                        Language(self._language),
+                    )
+                )
+
+        elif msg_type == "conversation.item.input_audio_transcription.completed":
+            # Final transcription
+            transcript = data.get("transcript", "").strip()
+            if transcript:
+                await self.stop_ttfb_metrics()
```
markbackman: Same, remove.

PR #3841, opened 2026-02-25
A few Groq-related tweaks:

kompfner:
- Wire up passing speed setting to Groq, even though only a value of 1.0 is supported today
- Update the 55y example to switch voices instead of changing speed
- Add a 55zn example to exercise runtime updates of Groq STT


filipi87 (approved):
LGTM!

PR #3794, opened 2026-02-21
skipping provider-specific messages during summarization

omChauhanDev:
#### Please describe the changes in your PR. If it is addressing an issue, please reference that as well.

Fixes #3778 

**Issue** : `LLMContextSummarizationUtil` crashes with `AttributeError: 'LLMSpecificMessage' object has no attribute 'get'` when 

**Cause**: Context contains `LLMSpecificMessage` objects (used by non-OpenAI models like Gemini with thinking)
  
**Fix**: Added `isinstance(message, LLMSpecificMessage)` guards in all 4 affected methods.

PR #3843, opened 2026-02-26
Fix context summarization splitting tool-call/response pairs

### LINKED ISSUE #3832: enable_context_summarization truncates message history between tool_calls and tool role messages, causing OpenAI 400 Error
### pipecat version

0.0.103

### Python version

3.13

### Operating System

Docker

### Issue description

When using `LLMAssistantAggregatorParams` with `enable_context_summarization = True`, an issue occurs when the context window reaches its limit and the older messages are pruned/summarized.

If the cutoff point for the kept window lands exactly between an `assistant` message that contains `tool_calls` and the subsequent `tool` message containing the function's result, the message history array is split improperly. This leaves an orphaned `tool` message in the history without its preceding `assistant` request.

When this broken message history is passed to the next LLM generation call, OpenAI rejects the payload because the schema is invalid, crashing the pipeline.

**Proposed Solution:** The truncation/summarization logic needs to be aware of tool call sequences. When cutting off the message history, it should never cut between an `assistant` message with `tool_calls` and the corresponding `tool` messages. It must either keep the entire tool call sequence or truncate the whole sequence.

`Before Summarization`:
Image

`After Summarization`:
Image

### Reproduction steps

1. Initialize a Pipecat pipeline using OpenAI as the LLM.
2. Configure the aggregator with `LLMAssistantAggregatorParams(enable_context_summarization=True)`.
3. Provide the LLM with tools/functions and trigger a conversation where a tool is called.
4. Continue the conversation until the context window limit is reached, triggering the summarization and message pruning.
5. Ensure the pruning boundary falls immediately after an `assistant` tool call but before the `tool` response.
6. Trigger the next LLM generation. The application will crash.

### Expected behavior

When context summarization reduces the message array, the slicing logic should check the roles and contents of the messages at the boundary. It should preserve the structural integrity of tool calls, ensuring a `tool` role message is always strictly preceded by an `assistant` message containing `tool_calls`.

### Actual behavior

The message history is truncated blindly based on window size. This strips away the `assistant` message that initiated the tool call, but leaves the `tool` response message in the active array. The subsequent API call to OpenAI fails.

### Logs

```shell
openai.BadRequestError: Error code: 400 - {'error': {'message': "Invalid parameter: messages with role 'tool' must be a response to a preceeding message with 'tool_calls'.", 'type': 'invalid_request_error', 'param': 'messages.[2].role', 'code': None}}
```

HamadAlrashid:
## Summary

Fixes #3832.

When the summary boundary in `get_messages_to_summarize()` falls inside a tool-call/tool-response group, the reconstructed context ends up with orphaned tool responses that have no corresponding `assistant` `tool_calls` message. The OpenAI API (and compatible APIs) rejects this with a 400 error.

- Added `_find_safe_summary_boundary()` that walks the boundary backward past any `tool` response messages and `assistant` messages with `tool_calls`, so complete groups are either fully summarized or fully preserved.
- Called the new helper in `get_messages_to_summarize()` after all existing boundary adjustments.

## Test plan

- [x] `TestToolCallBoundarySafety::test_boundary_on_assistant_tool_call_moves_back` — boundary landing on an assistant `tool_calls` message shifts back so its responses stay with it
- [x] `TestToolCallBoundarySafety::test_boundary_on_tool_response_moves_back` — boundary landing on a `tool` response walks all the way back past the paired `tool_calls` message
- [x] `TestToolCallBoundarySafety::test_boundary_on_multiple_tool_responses_moves_back` — multiple consecutive tool responses all move back together
- [x] `TestToolCallBoundarySafety::test_boundary_safe_when_no_tool_messages` — boundary is unchanged when no tool messages are near it
- [x] `TestToolCallBoundarySafety::test_boundary_returns_empty_when_all_messages_are_tool_related` — returns an empty result rather than an invalid slice when every summarizable message is tool-related

All 28 tests in `tests/test_context_summarization.py` pass.

### DIFF
--- a/changelog/3843.fixed.md
+++ b/changelog/3843.fixed.md
@@ -1,0 +1,1 @@
+- Fixed context summarization splitting tool-call/response pairs. When the summary boundary landed on an `assistant` message with `tool_calls` or a `tool` response message, the kept context contained orphaned tool responses without their corresponding tool call, causing OpenAI API 400 errors.


src/pipecat/utils/context/llm_context_summarization.py
```diff
@@ -272,6 +272,42 @@ def _get_function_calls_in_progress_index(messages: List[dict], start_idx: int)
...
+            messages: The full list of context messages.
+            summary_end: The proposed exclusive end index (messages before this
+                index will be summarized).
+
+        Returns:
+            A safe exclusive end index that does not split tool-call sequences.
+        """
+        while summary_end > 0:
+            last_summarized = messages[summary_end - 1]
+            if isinstance(last_summarized, LLMSpecificMessage):
```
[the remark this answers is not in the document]
HamadAlrashid: https://github.com/pipecat-ai/pipecat/pull/3794

markbackman:
cc @filipi87 to take a look

SwAt1563:
Q. Can this happen with other providers, ex Google (cuz google doesn't have "tool" role, but idk if this can happen with it LLMs)? if yes, the logic should consider them.

src/pipecat/utils/context/llm_context_summarization.py
```diff
@@ -272,6 +272,42 @@ def _get_function_calls_in_progress_index(messages: List[dict], start_idx: int)
 
         return -1
 
+    @staticmethod
+    def _find_safe_summary_boundary(messages: List[dict], summary_end: int) -> int:
```
filipi87, line 276: `_find_safe_summary_boundary` moves back any time it sees a tool message at the boundary, even when the tool call and ALL its responses are fully within the summarized range (no orphan).                                                                                                            

The check should mirror `_get_function_calls_in_progress_index's` intent: only move back when the kept range would contain orphaned tool responses. 

For a tool message at the boundary, it's only a problem when there are MORE tool messages right after summary_end (the group is split).

src/pipecat/utils/context/llm_context_summarization.py
filipi87, line 231: This logic should no longer be necessary if we already guarantee that no orphan function calls are kept inside `_find_safe_summary_boundary`.

filipi87, line 1: We need to rebase with the recent changes from main.

filipi87: Maybe we still need this I guess, depending on how we implement the fix inside `_find_safe_summary_boundary`.

filipi87:
Closing this in favor of https://github.com/pipecat-ai/pipecat/pull/3937.

The root cause was that `_get_function_calls_in_progress_index` scanned the full message list, so a tool response in the kept range would resolve its tool call, allowing the assistant message with tool_calls to be summarized while its response remained, causing an API rejection.

Fixed by bounding the scan to `summary_end`, so any tool response outside the summarized range is treated as absent, keeping the entire tool call group together in the kept context.

PR #3845, opened 2026-02-26
Add TTSSpeakFrame.push_assistant_aggregation to force context flush after TTS.

filipi87:
## Summary                                                                                                           
- Fixed `TTSSpeakFrame` not automatically committing spoken text to the conversation context when used outside of an LLM response (e.g., for bot greeting messages or injected speech)
- Added a new `LLMAssistantPushAggregationFrame` control frame that signals `LLMAssistantAggregator` to immediately flush its text buffer to the conversation context
- When a `TTSSpeakFrame` is processed while an LLM response is **not** in progress, the TTS service now automatically emits an `LLMAssistantPushAggregationFrame` after the text has been spoken, ensuring the spoken text is committed to context without waiting for an `LLMFullResponseEndFrame`
- Updated the `52-live-translation.py` example to demonstrate the fix

### DIFF
--- a/examples/foundational/52-live-translation.py
+++ b/examples/foundational/52-live-translation.py
@@ -11,6 +11,7 @@ from dotenv import load_dotenv
 from loguru import logger
 
 from pipecat.audio.vad.silero import SileroVADAnalyzer
+from pipecat.frames.frames import TTSSpeakFrame
 from pipecat.pipeline.pipeline import Pipeline
 from pipecat.pipeline.runner import PipelineRunner
 from pipecat.pipeline.task import PipelineParams, PipelineTask
@@ -110,6 +111,14 @@ async def run_bot(transport: BaseTransport, runner_args: RunnerArguments):
     @transport.event_handler("on_client_connected")
     async def on_client_connected(transport, client):
         logger.info(f"Client connected")
+        await task.queue_frames(
+            [
+                TTSSpeakFrame(
+                    text="Hello, welcome to live translation. Everything you say will be automatically translated to Spanish. Let's begin!",
+                    append_to_context=True,
+                ),
+            ]
+        )
 
     @transport.event_handler("on_client_disconnected")
     async def on_client_disconnected(transport, client):
--- a/src/pipecat/services/tts_service.py
+++ b/src/pipecat/services/tts_service.py
@@ -39,6 +39,7 @@ from pipecat.frames.frames import (
     Frame,
     InterimTranscriptionFrame,
     InterruptionFrame,
+    LLMAssistantPushAggregationFrame,
     LLMFullResponseEndFrame,
     LLMFullResponseStartFrame,
     StartFrame,
@@ -67,10 +68,16 @@ class TTSContext:
     """Context information for a TTS request.
 
     Attributes:
-        append_to_context: Whether this TTS output should be appended to the conversation context.
+        append_to_context: Whether this TTS output should be appended to the
+            conversation context after it is spoken.
+        push_assistant_aggregation: Whether to push an
+            ``LLMAssistantPushAggregationFrame`` after the TTS has finished
+            speaking, forcing the assistant aggregator to commit its current
+            text buffer to the conversation context.
     """
 
     append_to_context: bool = True
+    push_assistant_aggregation: Optional[bool] = False
 
 
 class TextAggregationMode(str, Enum):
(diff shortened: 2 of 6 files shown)


markbackman:
@filipi87 can you explain why this is needed? I would have expected the TTSSpeakFrame's append_to_context field to cause it to be added to the context by the LLMAssistantAggregator when the corresponding TTSTextFrames are handled. Are the TTSTextFrames still bound by the LLMFullResponseStartFrame and LLMFullResponseEndFrame?

filipi87:
> Are the TTSTextFrames still bound by the LLMFullResponseStartFrame and LLMFullResponseEndFrame?

They are aggregated by the `LLMAssistantAggregator`, but the aggregated text is only pushed when either an `InterruptionFrame` or an `LLMFullResponseEndFrame` is received.

So, in a scenario where we pause interruptions in `UserTurnStartStrategy`, the `TTSSpeakFrames` would not be added to the context.

You can test this using the example I updated by changing it to **NOT** `push_assistant_aggregation`. You’ll see that, in this case, the `TTSSpeakFrames` is only added to the context after we receive the `LLMFullResponseEndFrame`, that is, after the user has spoken, instead of before.

markbackman:
> > Are the TTSTextFrames still bound by the LLMFullResponseStartFrame and LLMFullResponseEndFrame?
> 
> They are aggregated by the `LLMAssistantAggregator`, but the aggregated text is only pushed when either an `InterruptionFrame` or an `LLMFullResponseEndFrame` is received.
> 
> So, in a scenario where we pause interruptions in `UserTurnStartStrategy`, the `TTSSpeakFrames` would not be added to the context.
> 
> You can test this using the example I updated by changing it to **NOT** `push_assistant_aggregation`. You’ll see that, in this case, the `TTSSpeakFrames` is only added to the context after we receive the `LLMFullResponseEndFrame`, that is, after the user has spoken, instead of before.

I see. So while the TTSSpeakFrame is added to the context, it's out of order. Interesting!

markbackman:
Requiring the user to set both `push_assistant_aggregation` and `append_to_context` seems like we're overburdening the user. Could we instead remove `push_assistant_aggregation` from the frame and use `append_to_context` to determine if the user wants the text added to the context automatically?

Essentially, three cases:
1. TTSSpeakFrame pushed outside of LLM response with `append_to_context=True`: Flush and add text to context right away
2. TTSSpeakFrame pushed inside of an LLM response with `append_to_context=True`: Add to the context when the text is received
3. `append_to_context=False`: Skip adding

That would be more clear from an API / developer perspective. Is that possible?

filipi87:
> 2. TTSSpeakFrame pushed inside of an LLM response with append_to_context=True: Add to the context when the text is received

For this case, you mean respect the current behavior, right ? Only add to the context after we receive the LLMFullResponseEndFrame or an interruption. Is that it ?

kompfner:
> Could we instead remove push_assistant_aggregation from the frame and use append_to_context to determine if the user wants the text added to the context automatically?

+1, if it's possible

filipi87:
@markbackman , @kompfner , I have applied the changes. Please could you review again ?

examples/foundational/52-live-translation.py
```diff
@@ -110,6 +111,14 @@ async def run_bot(transport: BaseTransport, runner_args: RunnerArguments):
     @transport.event_handler("on_client_connected")
     async def on_client_connected(transport, client):
         logger.info(f"Client connected")
+        await task.queue_frames(
+            [
+                TTSSpeakFrame(
+                    text="Hello, welcome to live translation. Everything you say will be automatically translated to spanish. We can begin!",
```
kompfner: nit: 
```suggestion
                    text="Hello, welcome to live translation. Everything you say will be automatically translated to Spanish. Let's begin!",
```

src/pipecat/services/tts_service.py
```diff
@@ -641,10 +648,13 @@ async def process_frame(self, frame: Frame, direction: FrameDirection):
         elif isinstance(frame, TTSSpeakFrame):
             # Store if we were processing text or not so we can set it back.
             processing_text = self._processing_text
+            # If we are not receiving text from the LLM, we can assume that the SpeakFrame should be automatically added to the context
```
kompfner, line 651: Nice. 👍

kompfner (approved):
Nice!

markbackman (commented):
Very nice! Thanks for the simplification 🙇

PR #3849, opened 2026-02-26
Flatten LiveOptions fields into DeepgramSTTSettings

filipi87:
Flatten LiveOptions fields into DeepgramSTTSettings

### DIFF
--- a/src/pipecat/services/deepgram/stt.py
+++ b/src/pipecat/services/deepgram/stt.py
@@ -177,56 +177,27 @@ class DeepgramSTTSettings(STTSettings):
     params.
 
     Parameters:
-        callback: Callback URL for async transcription delivery.
-        callback_method: HTTP method for the callback (``"GET"`` or ``"POST"``).
         channels: Number of audio channels.
-        detect_entities: Enable named entity detection.
         diarize: Enable speaker diarization.
-        dictation: Enable dictation mode.
         encoding: Audio encoding (e.g. ``"linear16"``).
         endpointing: Endpointing sensitivity in ms, or ``False`` to disable.
         interim_results: Whether to emit interim transcriptions.
-        keyterm: Keyterms to boost (str or list of str).
-        keywords: Keywords to boost (str or list of str).
-        mip_opt_out: Opt out of model improvement program.
-        multichannel: Enable per-channel transcription for multi-channel audio.
-        numerals: Convert spoken numbers to numerals.
         profanity_filter: Filter profanity from transcripts.
         punctuate: Add punctuation to transcripts.
-        redact: Redact sensitive information (str or list).
-        replace: Word replacement rules (str or list).
-        search: Search terms to highlight (str or list of str).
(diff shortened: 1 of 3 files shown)


src/pipecat/services/deepgram/stt.py
```diff
@@ -402,10 +302,10 @@ def __init__(
                 )
             base_url = url
 
-        default_options = LiveOptions(
-            encoding="linear16",
-            language=Language.EN,
+        settings = DeepgramSTTSettings(
```
kompfner, line 276: I wonder if we can away with including _just_ the below fields in `DeepgramSTTSettings` and ask users to put any additional settings they want into `extra`...it feels to me like we ought not to sign up for the task of forever maintaining an exhaustive "mirror" of the deepgram options...

kompfner: (Paring down the fields would, of course, mean that once we deprecate init-time `live_options` in favor of init-time `settings`, those migrating their code would have to move from providing strongly-typed to stringly-typed options. It'll only affect folks using some of these "long-tail" fields. Might be worth it)

kompfner: it might be worth a quick pass to decide which other fields we might want to "promote" from being loose `extra` fields to those worth first-class typed fields? Maybe fields we expect to be relatively stable (i.e. will stick around awhile) and expect to have high usage?

filipi87: I think this could make sense. Maybe we could keep: `model`, `language`, `encoding`, `channels`, `interim_results`, `smart_format`, `punctuate`, `profanity_filter`, `vad_events`, `diarize` and `endpointing`.

I believe multi speaker transcription will probably become a very common use case, so keeping `diarize` makes sense. 

And `endpointing` is usefull to control silence detection sensitivity.

filipi87: What do you think ?

kompfner: Sounds really good!

We can always "promote" more in the future if we feel the ergonomics gains of adding those fields is worth it, right? If/when we "promote" a field, it wouldn't break those passing in that same field in `extra`, would it (i.e. `settings.extra["foo"]` would still work, even if `settings.foo` existed?)

Actually, that brings to mind one thing we should validate: if someone _does_ provide `settings.foo` _and_ `settings.extra["foo"]`, `settings.foo` should take precedence.

filipi87: Ok, cool. Keeping only the fields which I mentioned above.

filipi87: Also applied the changes so the extra fields have lowest precedence.

filipi87:
@kompfner, I will merge this PR into the other one, so we can review all the changes in a single place now that we’ve agreed on what we want.
- https://github.com/pipecat-ai/pipecat/pull/3848

PR #3852, opened 2026-02-27
Deprecate processing metrics (ProcessingMetricsData)

markbackman:
## Summary

- Add deprecation warnings and docstrings to `ProcessingMetricsData`, `start_processing_metrics()`, and `stop_processing_metrics()` on both `FrameProcessor` and `FrameProcessorMetrics`
- Warning is emitted once at runtime (class-level flag) when any service calls `start_processing_metrics()`
- No behavioral changes — all existing metrics continue to be collected and reported as before

## Context

Processing metrics are inconsistently implemented across services and overlap with the better-defined TTFB metric. This deprecation prepares for full removal in a future version.

## Test plan

- [x] All tests pass (`uv run pytest` — 702 passed)
- [x] Verified warning appears once at runtime when running an example
- [x] Verified no behavioral changes — metrics still collected and reported

### DIFF
--- a/changelog/3852.deprecated.md
+++ b/changelog/3852.deprecated.md
@@ -1,0 +1,1 @@
+- Deprecated `ProcessingMetricsData` and `start_processing_metrics()`/`stop_processing_metrics()` on `FrameProcessor` and `FrameProcessorMetrics`. These metrics don't accurately depict a service's performance. Instead, TTFB metrics are recommended. Processing metrics will be removed in the 1.0.0 version.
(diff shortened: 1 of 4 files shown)


examples/foundational/07-interruptible.py
```diff
@@ -92,8 +92,8 @@ async def run_bot(transport: BaseTransport, runner_args: RunnerArguments):
     task = PipelineTask(
         pipeline,
         params=PipelineParams(
-            enable_metrics=True,
-            enable_usage_metrics=True,
+            # enable_metrics=True,
+            # enable_usage_metrics=True,
```
filipi87: Do we need this ?

filipi87 (approved):
Makes sense. 👍

examples/foundational/07-interruptible.py
markbackman: Oops! No. That was from testing. Reverting that now.

PR #3855, opened 2026-02-27
Improve context summarization with dedicated LLM, timeout, and observability

markbackman:
## Summary

- Fixed summary message role to use `user` instead of `assistant`, preventing consecutive assistant messages that some LLM providers reject
- Added `summary_message_template` to `LLMContextSummarizationConfig` for customizing how summaries are formatted when injected into context (e.g., wrapping in XML tags)
- Added optional `llm` field to `LLMContextSummarizationConfig` for routing summarization to a dedicated LLM service (e.g., Gemini Flash) instead of the primary pipeline model
- Added `summarization_timeout` (default 120s) to prevent hung LLM calls from permanently blocking future summarizations
- Added `on_summary_applied` event to `LLMContextSummarizer` for observability, providing message counts before and after summarization
- Added example `54b` demonstrating all new summarization options together

## Testing

- `uv run python examples/foundational/54b-context-summarization-custom.py` to verify dedicated LLM, custom template, timeout, and event handling

### DIFF
--- a/examples/foundational/54-context-summarization-openai.py
+++ b/examples/foundational/54-context-summarization-openai.py
@@ -20,14 +20,13 @@ from loguru import logger
 
 from pipecat.adapters.schemas.function_schema import FunctionSchema
 from pipecat.adapters.schemas.tools_schema import ToolsSchema
-from pipecat.audio.turn.smart_turn.local_smart_turn_v3 import LocalSmartTurnAnalyzerV3
 from pipecat.audio.vad.silero import SileroVADAnalyzer
-from pipecat.audio.vad.vad_analyzer import VADParams
 from pipecat.frames.frames import LLMRunFrame
 from pipecat.pipeline.pipeline import Pipeline
 from pipecat.pipeline.runner import PipelineRunner
 from pipecat.pipeline.task import PipelineParams, PipelineTask
 from pipecat.processors.aggregators.llm_context import LLMContext
+from pipecat.processors.aggregators.llm_context_summarizer import SummaryAppliedEvent
 from pipecat.processors.aggregators.llm_response_universal import (
     LLMAssistantAggregatorParams,
     LLMContextAggregatorPair,
@@ -42,8 +41,6 @@ from pipecat.services.openai.llm import OpenAILLMService
 from pipecat.transports.base_transport import BaseTransport, TransportParams
 from pipecat.transports.daily.transport import DailyParams
 from pipecat.transports.websocket.fastapi import FastAPIWebsocketParams
-from pipecat.turns.user_stop import TurnAnalyzerUserTurnStopStrategy
-from pipecat.turns.user_turn_strategies import UserTurnStrategies
 from pipecat.utils.context.llm_context_summarization import LLMContextSummarizationConfig
 
 load_dotenv(override=True)
@@ -120,10 +117,7 @@ async def run_bot(transport: BaseTransport, runner_args: RunnerArguments):
     user_aggregator, assistant_aggregator = LLMContextAggregatorPair(
         context,
         user_params=LLMUserAggregatorParams(
-            user_turn_strategies=UserTurnStrategies(
-                stop=[TurnAnalyzerUserTurnStopStrategy(turn_analyzer=LocalSmartTurnAnalyzerV3())]
-            ),
-            vad_analyzer=SileroVADAnalyzer(params=VADParams(stop_secs=0.2)),
+            vad_analyzer=SileroVADAnalyzer(),
         ),
         assistant_params=LLMAssistantAggregatorParams(
             enable_context_summarization=True,
@@ -137,6 +131,19 @@ async def run_bot(transport: BaseTransport, runner_args: RunnerArguments):
             ),
         ),
     )
+
+    # Listen for summarization events
+    summarizer = assistant_aggregator._summarizer
+    if summarizer:
+
+        @summarizer.event_handler("on_summary_applied")
+        async def on_summary_applied(summarizer, event: SummaryAppliedEvent):
+            logger.info(
+                f"Context summarized: {event.original_message_count} messages -> "
+                f"{event.new_message_count} messages "
+                f"({event.summarized_message_count} summarized, "
+                f"{event.preserved_message_count} preserved)"
+            )
 
     pipeline = Pipeline(
         [
--- a/examples/foundational/54c-context-summarization-dedicated-llm.py
+++ b/examples/foundational/54c-context-summarization-dedicated-llm.py
@@ -4,12 +4,14 @@
 # SPDX-License-Identifier: BSD 2-Clause License
 #
 
-"""Example demonstrating context summarization feature.
+"""Example demonstrating advanced context summarization configuration.
 
-This example shows how to enable and configure context summarization to automatically
-compress conversation history when token limits are approached. It also demonstrates
-that summarization correctly handles function calls, preserving incomplete function
-call sequences.
+This example shows how to customize context summarization with:
+- A dedicated cheap/fast LLM for generating summaries (Gemini Flash)
+- A custom summary message template (XML tags)
+- A custom summarization prompt
+- A summarization timeout
+- The on_summary_applied event for observability
 """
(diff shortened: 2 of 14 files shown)


changelog/3855.fixed.md
filipi87, line 1: Hi @markbackman, when we discussed this back then with @kompfner, I thought we had said that this was already handled by the universal context adapter, which would change this when the LLM didn’t support it. Wasn’t this the case?

examples/foundational/54-context-summarization-openai.py
```diff
@@ -138,6 +132,19 @@ async def run_bot(transport: BaseTransport, runner_args: RunnerArguments):
         ),
     )
 
+    # Listen for summarization events
+    summarizer = assistant_aggregator._summarizer
+    if summarizer:
+
+        @summarizer.event_handler("on_summary_applied")
+        async def on_summary_applied(summarizer, event: SummaryAppliedEvent):
```
filipi87, line 140: Pretty cool. 🔥

examples/foundational/54c-context-summarization-dedicated-llm.py
filipi87, line 1: Maybe rename, `54b-context-summarization-dedicated-llm.py` ? ‎

src/pipecat/processors/aggregators/llm_context_summarizer.py
```diff
@@ -296,8 +326,10 @@ async def _apply_summary(self, summary: str, last_summarized_index: int):
         # Get recent messages to keep
         recent_messages = messages[last_summarized_index + 1 :]
 
-        # Create summary message as an assistant message
-        summary_message = {"role": "assistant", "content": f"Conversation summary: {summary}"}
+        # Create summary message as a user message (the summary is context
+        # provided *to* the assistant, not something the assistant said)
+        summary_content = self._config.summary_message_template.format(summary=summary)
+        summary_message = {"role": "user", "content": summary_content}
```
filipi87, line 401: Just double checking on this one if we really are going to keep it as "user" or "assistant".

src/pipecat/processors/aggregators/llm_response_universal.py
```diff
@@ -82,6 +83,9 @@
 from pipecat.utils.string import TextPartForConcatenation, concatenate_aggregated_text
 from pipecat.utils.time import time_now_iso8601
 
+if TYPE_CHECKING:
+    from pipecat.services.llm_service import LLMService
```
filipi87: Is there any specific reason to use `TYPE_CHECKING` here instead of a regular `import` ? I am always a little bit confusing when we actually need to use this.

src/pipecat/processors/aggregators/llm_response_universal.py
```diff
@@ -1245,13 +1249,71 @@ async def _on_request_summarization(
...
         """
-        await self.push_frame(frame, FrameDirection.UPSTREAM)
+        summarization_llm = (
+            self._params.context_summarization_config.llm
+            if self._params.context_summarization_config
+            else None
+        )
+
+        if summarization_llm:
+            self.create_task(self._generate_summary_with_dedicated_llm(summarization_llm, frame))
```
filipi87: I think that if we’re using a dedicated LLM, `_on_request_summarization` wouldn’t even need to be invoked, and all of this logic could live inside `llm_context_summarizer.py`. 

That way, we wouldn’t need any changes in the assistant aggregator. What do you think?

src/pipecat/services/llm_service.py
```diff
@@ -437,7 +437,17 @@ async def _generate_summary_task(self, frame: LLMContextSummaryRequestFrame):
         error = None
 
         try:
-            summary, last_index = await self._generate_summary(frame)
+            if frame.summarization_timeout:
```
filipi87: Maybe it would be worth to always include a `timeout` by default, even if not provided by the user ?

changelog/3855.fixed.md
markbackman: I think this case is different. The universal context adapter translates system messages into user messages. In this case, we were explicitly setting the summary message to an `assistant` role, which means that the assistant "said" the summary. But, in this case we want this to be a `user` role message.

markbackman: Fix in question is here 6244088f06d82678ef98c1a7bd0e6fcbeb12d6b4.

examples/foundational/54-context-summarization-openai.py
markbackman: I stole this idea from Gradient Bang. It should be helpful for observability (and will let us plug this functionality into Gradient Bang).

examples/foundational/54c-context-summarization-dedicated-llm.py
markbackman: I'll make this 54c, since you have a 54b in your other PR. Good idea on the name.

src/pipecat/processors/aggregators/llm_context_summarizer.py
markbackman: Yes, I believe `user` is correct. This is how we do it in Flows and Gradient Bang. I think, conceptually, this is right. We (the `user`) are telling the assistant what the conversation history is.

changelog/3855.fixed.md
kompfner: LLM-specific adapters _should_ already concatenate consecutive `user` messages (or consecutive `assistant` messages) into single multipart messages if the LLM requires it...

Maybe it's the right call to make summaries `user` messages, but that should be a separate consideration/debate than preventing consecutive message.

src/pipecat/processors/aggregators/llm_response_universal.py
markbackman: It's not strictly needed here, but can prevent any future issues.

markbackman: And, yes, it's confusing! I agree.

changelog/3855.fixed.md
markbackman: @kompfner great to know! Then, I'll make this a "change" to be clear that this is making the conversation history more semantically correct. Thank you.

src/pipecat/processors/aggregators/llm_response_universal.py
markbackman: Good point. Yes, this could simplify. Let me see what this looks like.

The biggest benefit is to keep the LLMAssistantAggregator simpler. This logic arguably should not live there.

markbackman: Ok, great suggestion. This is so much cleaner now 🙇

src/pipecat/services/llm_service.py
markbackman: Yes, that will simplify the code (eliminating the if/else) and provide a safety net for developers.

markbackman:
@filipi87 great suggestions. This one is ready for review again. Thanks for helping to simplify 🙌

src/pipecat/utils/context/llm_context_summarization.py
```diff
@@ -73,13 +79,29 @@ class LLMContextSummarizationConfig:
...
     """
 
     max_context_tokens: int = 8000
     target_context_tokens: int = 6000
     max_unsummarized_messages: int = 20
     min_messages_after_summary: int = 4
     summarization_prompt: Optional[str] = None
+    summary_message_template: str = "Conversation summary: {summary}"
+    llm: Optional["LLMService"] = None
+    summarization_timeout: float = DEFAULT_SUMMARIZATION_TIMEOUT
```
filipi87, line 104: @markbackman, @kompfner, I believe that with the changes in this PR:
- https://github.com/pipecat-ai/pipecat/pull/3863/changes

All these settings will be moved inside `LLMContextSummaryConfig`, right ?

filipi87 (approved):
LGTM! Nice improvements @markbackman 🚀

src/pipecat/utils/context/llm_context_summarization.py
markbackman: Yes, that's right.

PR #3857, opened 2026-02-27
Fix Smart Turn v3 producing incorrect predictions at non-16kHz sample rates

### LINKED ISSUE #3844: Smart Turn v3 silently breaks when audio_in_sample_rate=8000 (Twilio/telephony)
## Summary

Setting `audio_in_sample_rate=8000` in `PipelineParams` (as recommended by the [Twilio WebSocket integration guide](https://docs.pipecat.ai/guides/telephony/twilio-websockets#key-features)) silently breaks Smart Turn v3 end-of-turn detection. The model receives 8kHz audio but its internal `WhisperFeatureExtractor` hardcodes `sampling_rate=16000`, causing it to interpret speech at 2x speed with shifted pitch. This leads to aggressive/incorrect turn predictions with no errors or warnings.

## The Documentation Conflict

The **Twilio guide** recommends:

```python
task = PipelineTask(
    pipeline,
    params=PipelineParams(
        audio_in_sample_rate=8000,
        audio_out_sample_rate=8000,
    ),
)
```

The **Smart Turn v3 README** states:

> Smart Turn takes 16kHz PCM audio as input.

Since Smart Turn is auto-injected as the default stop strategy via `UserTurnStrategies.__post_init__`, anyone following the Twilio docs will unknowingly break turn detection.

## Root Cause

1. `TwilioFrameSerializer` upsamples 8kHz μ-law to the pipeline's configured `audio_in_sample_rate`
2. When `audio_in_sample_rate=16000` (the default), the serializer upsamples to 16kHz — Smart Turn works correctly
3. When `audio_in_sample_rate=8000`, the serializer outputs 8kHz PCM — Smart Turn receives 8kHz audio
4. `LocalSmartTurnAnalyzerV3._predict_endpoint` uses `WhisperFeatureExtractor(sampling_rate=16000)` with no internal resampling
5. The model processes 8kHz audio as if it were 16kHz — no error, no warning, just wrong classifications

## Evidence

We ran A/B tests comparing Smart Turn predictions on the same utterances at 8kHz vs 16kHz:

- **6 out of 20 utterances** had classification flips (Complete ↔ Incomplete)
- **Max probability delta**: 0.9391 (a confident "complete" at 8kHz flipped to confident "incomplete" at 16kHz)
- **Mean turn duration dropped 51%** in production (2.33s → 1.14s) when using 8kHz
- **Phone numbers were fragmented across multiple turns** — the model couldn't recognize digit sequences as incomplete at 8kHz

## Community Impact

Multiple Discord threads and issues show users hitting this without identifying the root cause:

- "Chipmunk audio" reports with Twilio + OpenAI Realtime (audio at 2x speed)
- "Issues with Local Smart Turn Analyzer" — Smart Turn marking turns COMPLETE incorrectly
- "Premature turn finalization" — bot replying twice per utterance
- Telnyx users finding mysterious "2x mismatch" workarounds
- Support staff giving the correct workaround ("just don't set it") without explaining the Smart Turn connection

## Suggested Fixes

1. **Update Twilio/telephony docs** to warn against setting `audio_in_sample_rate=8000` when using Smart Turn, or recommend only setting `audio_out_sample_rate=8000`
2. **Add a runtime warning** in `LocalSmartTurnAnalyzerV3` when `set_sample_rate` is called with a value != 16000
3. **Add internal resampling** in the Smart Turn analyzer to handle non-16kHz input gracefully

## Workaround

Leave `audio_in_sample_rate` at the default (16000). Only set `audio_out_sample_rate=8000` for telephony. The `TwilioFrameSerializer` handles the 8kHz → 16kHz upsampling internally.

```python
task = PipelineTask(
    pipeline,
    params=PipelineParams(
        audio_out_sample_rate=8000,  # Twilio needs 8kHz output
        # Do NOT set audio_in_sample_rate=8000 — breaks Smart Turn v3
    ),
)
```

## Environment

- pipecat-ai 0.0.103
- LocalSmartTurnAnalyzerV3 (bundled ONNX weights)
- TwilioFrameSerializer
- Silero VAD with `stop_secs=0.2`

kollaikal-rupesh:
## Summary

- Resample audio to 16 kHz before Whisper feature extraction when the pipeline uses a different sample rate (e.g. 8 kHz for Twilio telephony)
- Replace five hardcoded `16000` references with a `_MODEL_SAMPLE_RATE` constant
- Fix WAV debug logger to write files with the correct sample rate header
- Use `soxr.resample` with VHQ quality for high-fidelity resampling (per @marcus-daily's review feedback)

## Root Cause

The `_predict_endpoint` method in `LocalSmartTurnAnalyzerV3` assumed 16 kHz audio in five places without checking the actual pipeline sample rate. When running at 8 kHz (common for Twilio/telephony transports), the audio was fed to the Whisper feature extractor at the wrong rate, causing the model to perceive speech at 2x speed with shifted formant frequencies. This produced incorrect end-of-turn predictions with no error or warning.

## Approach

Added a `_resample_to_model_rate()` method that uses `soxr.resample` with VHQ quality for high-quality sample rate conversion from the actual sample rate to 16 kHz. This is called at the start of `_predict_endpoint()` before any model processing. A one-time warning is logged when resampling is needed, so users know it's happening.

Initially used `numpy.interp` for linear interpolation, but switched to `soxr` based on @marcus-daily's review suggestion for better audio quality. `soxr` is already a core dependency in pipecat (`soxr~=0.5.0`) and is used by `SOXRAudioResampler` and `SOXRStreamAudioResampler` throughout the project.

## Fixes

Fixes #3844

## Testing

- All existing tests pass (559 passed)
- The resampling path is exercised whenever `sample_rate != 16000` (e.g. Twilio at 8 kHz)

### DIFF
--- a/src/pipecat/audio/turn/smart_turn/local_smart_turn_v3.py
+++ b/src/pipecat/audio/turn/smart_turn/local_smart_turn_v3.py
@@ -46,7 +46,6 @@ class LocalSmartTurnAnalyzerV3(BaseSmartTurn):
         super().__init__(**kwargs)
 
         self._log_data = env_truthy("PIPECAT_SMART_TURN_LOG_DATA", default=False)
-        self._resample_warned = False
 
         if not smart_turn_model_path:
             # Load bundled model
@@ -136,13 +135,6 @@ class LocalSmartTurnAnalyzerV3(BaseSmartTurn):
         actual_rate = self._sample_rate or _MODEL_SAMPLE_RATE
         if actual_rate == _MODEL_SAMPLE_RATE:
             return audio_array
-
-        if not self._resample_warned:
-            logger.warning(
-                f"Smart Turn v3 model expects {_MODEL_SAMPLE_RATE}Hz audio but received "
-                f"{actual_rate}Hz. Audio will be resampled automatically."
-            )
-            self._resample_warned = True
 
         return soxr.resample(audio_array, actual_rate, _MODEL_SAMPLE_RATE, quality="VHQ")
 
--- a/src/pipecat/audio/turn/smart_turn/local_smart_turn_v3.py
+++ b/src/pipecat/audio/turn/smart_turn/local_smart_turn_v3.py
@@ -20,6 +20,9 @@ from transformers import WhisperFeatureExtractor
 from pipecat.audio.turn.smart_turn.base_smart_turn import BaseSmartTurn
 from pipecat.utils.env import env_truthy
 
+# The Whisper-based ONNX model expects 16 kHz audio input.
+_MODEL_SAMPLE_RATE = 16000
+
 
 class LocalSmartTurnAnalyzerV3(BaseSmartTurn):
     """Local turn analyzer using the smart-turn-v3 ONNX model.
@@ -42,6 +45,7 @@ class LocalSmartTurnAnalyzerV3(BaseSmartTurn):
         super().__init__(**kwargs)
 
         self._log_data = env_truthy("PIPECAT_SMART_TURN_LOG_DATA", default=False)
+        self._resample_warned = False
 
         if not smart_turn_model_path:
             # Load bundled model
@@ -77,7 +81,7 @@ class LocalSmartTurnAnalyzerV3(BaseSmartTurn):
         logger.debug("Loaded Local Smart Turn v3.x")
 
     def _write_audio_to_wav(
-        self, audio_array: np.ndarray, sample_rate: int = 16000, suffix: str = ""
+        self, audio_array: np.ndarray, sample_rate: int = _MODEL_SAMPLE_RATE, suffix: str = ""
     ) -> None:
         """Write audio data to a WAV file in a background thread.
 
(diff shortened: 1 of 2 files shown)


src/pipecat/audio/turn/smart_turn/local_smart_turn_v3.py
```diff
@@ -119,10 +123,39 @@ def write_wav():
...
+                f"{actual_rate}Hz. Audio will be resampled automatically."
+            )
+            self._resample_warned = True
+
+        num_output_samples = int(len(audio_array) * _MODEL_SAMPLE_RATE / actual_rate)
+        return np.interp(
+            np.linspace(0, len(audio_array), num_output_samples, endpoint=False),
+            np.arange(len(audio_array)),
+            audio_array,
+        )
```
marcus-daily: This PR mostly looks good, but using linear interpolation for resampling isn't great for audio quality. Maybe we can use `soxr.resample` instead?

marcus-daily: If so, we'll need to ensure the project dependencies are correct (I think Pipecat already uses soxr somewhere).

kollaikal-rupesh: Good suggestion! `soxr` is already a core dependency (`soxr~=0.5.0` in pyproject.toml) and pipecat uses it in `SOXRAudioResampler` and `SOXRStreamAudioResampler`. Switching to `soxr.resample` with VHQ quality will give much better audio fidelity than `np.interp` linear interpolation. Will update.

kollaikal-rupesh:
Updated in 3f1769b. Switched from `np.interp` linear interpolation to `soxr.resample` with VHQ quality. Also extracted the resampling into a `_resample_to_model_rate()` helper for clarity. @marcus-daily ready for another look when you get a chance!

src/pipecat/audio/turn/smart_turn/local_smart_turn_v3.py
```diff
@@ -119,10 +124,34 @@ def write_wav():
...
+            audio_array: Audio data as a float32 numpy array.
+
+        Returns:
+            Resampled audio array at 16 kHz.
+        """
+        actual_rate = self._sample_rate or _MODEL_SAMPLE_RATE
+        if actual_rate == _MODEL_SAMPLE_RATE:
+            return audio_array
+
+        if not self._resample_warned:
```
markbackman: I don't think we want to warn on resampling. We don't do that elsewhere in the code; it happens transparently, which makes sense here, too.

markbackman (commented):
This looks good! Thanks for the contribution. Just a small comment about the warning.

src/pipecat/audio/turn/smart_turn/local_smart_turn_v3.py
kollaikal-rupesh: oh yes good point, updated now

markbackman:
LGTM, but I'll leave the final ✅ to @marcus-daily.

marcus-daily (approved):

marcus-daily:
Great work, thanks again for this @kollaikal-rupesh. Now merged in!

PR #3859, opened 2026-02-27
feat(transports): add FreeSWITCH SIP/RTP transport (UAS, LAN dial-in)

rasonyang:
## Summary

- Add a FreeSWITCH-scoped SIP UAS transport to Pipecat enabling PBX/SBC SIP calls to enter pipelines directly
- Implements G.711 PCMU codec with numpy LUT tables, 20ms RTP packetization, and RFC 2833 DTMF detection
- Follows Pipecat transport architecture: `FreeSwitchSIPServerTransport` (BaseObject) + per-call `FreeSwitchSIPCallTransport` (BaseTransport) with `input()`/`output()` processors

### Constraints

- **LAN-only:** No NAT traversal, STUN, or ICE. Both FreeSWITCH and the bot must share a reachable network.
- **No SIP REGISTER:** Dial-in only — FreeSWITCH routes calls to the bot's IP:port directly.
- **UAS role only:** The bot accepts incoming INVITEs; it does not originate outbound calls.

## Architecture

```
┌──────────────────────────────────────────────────┐
│        FreeSwitchSIPServerTransport              │
│  (UDP SIP listener, port manager)                │
│                                                  │
│  on INVITE → create FreeSwitchSIPCallTransport   │
│  on BYE    → teardown call                       │
└──────────────┬───────────────────────────────────┘
               │ emits per call
    ┌──────────▼──────────────────────┐
    │  FreeSwitchSIPCallTransport     │
    │   (extends BaseTransport)       │
    │                                 │
    │  .input()  → FreeSwitchSIPInputTransport  (BaseInputTransport)
    │  .output() → FreeSwitchSIPOutputTransport (BaseOutputTransport)
    │  .session  → FreeSwitchSIPSession (call metadata)
    └──────────┬──────────────────────┘
               │ owns
    ┌──────────▼──────────────┐
    │      RTPSession         │
    │  (UDP send/recv, 20ms)  │
    │  rx_queue → input       │
    │  tx_queue ← output      │
    └─────────────────────────┘
```

**Data flow:**
- Inbound: RTP UDP → G.711 decode → resample 8k→16k → `InputAudioRawFrame(16kHz)` → pipeline
- Outbound: pipeline → `OutputAudioRawFrame` → resample to 8k → G.711 encode → RTP UDP

## Supported Codecs

| Codec | PT | Rate | Status |
|-------|-----|------|--------|
| PCMU  | 0   | 8000 | ✅ Implemented |
| PCMA  | 8   | 8000 | 🔮 Future |

## New Files

```
src/pipecat/transports/sip/
├── __init__.py      # Exports FreeSwitchSIPServerTransport, FreeSwitchSIPCallTransport, FreeSwitchSIPParams
├── params.py        # FreeSwitchSIPParams (extends TransportParams)
├── transport.py     # FreeSwitchSIPServerTransport, FreeSwitchSIPCallTransport, FreeSwitchSIPInput/OutputTransport
├── signaling.py     # SIP message parsing, response building
├── sdp.py           # SDP parse/generate
├── rtp.py           # RTPSession (send/recv loops, pre-buffering, DTMF)
├── codecs.py        # G.711 PCMU codec (numpy LUT) + resampling
```

## Recent Improvements (latest commits)

- **Rename to FreeSwitchSIP\***: All public classes renamed from generic `SIP*` to `FreeSwitchSIP*` to signal implementation constraints (LAN-only, no NAT/STUN, no REGISTER, dial-in only)
- **Constraints documented**: Explicit constraints added to changelog, example docstring, and design doc summary
- **RTP dead timeout detection**: `rtp_dead_timeout_ms` param now actively monitored — calls are torn down when no RTP is received within the configured timeout
- **SIP OPTIONS handling**: Responds to SBC keepalive probes with `200 OK` (with `Allow` header) to prevent SBCs from marking us as dead
- **Wired `rtp_prebuffer_frames`**: `FreeSwitchSIPParams.rtp_prebuffer_frames` and `dtmf_enabled` are now forwarded through `FreeSwitchSIPSession` to `RTPSession`
- **Error propagation**: Input transport `_rx_loop` and `_dtmf_loop` now push errors upstream via `push_error()` instead of silently logging; RTP send loop failures trigger call cleanup with `on_call_ended`
- **22 new transport-level tests**: Covering session lifecycle, port allocation/release, max calls enforcement, ACK timeout, duplicate INVITE rejection, OPTIONS handling, RTP dead timeout, and server shutdown

## Test Plan

- [x] G.711 PCMU encode/decode roundtrip, known vectors (8 tests)
- [x] Integer-ratio resampling up/down, DC preservation (9 tests)
- [x] SDP parse/generate, single/multiple codecs, missing fields (8 tests)
- [x] RTP header pack/unpack, sequence/timestamp wraparound (7 tests)
- [x] RFC 2833 DTMF parsing, digits 0-9/*/# (6 tests)
- [x] SIP message parsing INVITE/BYE/ACK, response building (12 tests)
- [x] Transport: session create/stop, prebuffer/dtmf forwarding (6 tests)
- [x] Transport: port allocation/release/exhaustion (2 tests)
- [x] Transport: INVITE, duplicate rejection, max calls, OPTIONS, BYE (8 tests)
- [x] Transport: ACK timeout, RTP dead timeout (3 tests)
- [x] Transport: FreeSwitchSIPCallTransport input/output, server stop (3 tests)
- [x] All 72 tests passing, ruff lint/format clean

## Local Verification

```bash
# Run all SIP tests
uv run pytest tests/test_sip_*.py -v

# Lint
uv run ruff check src/pipecat/transports/sip/
uv run ruff format --check src/pipecat/transports/sip/ tests/test_sip_transport.py

# Test with pjsua (CLI SIP client)
python examples/foundational/56-freeswitch-sip-transport.py
pjsua --null-audio sip:bot@<HOST>:5060
```

## Future Roadmap

- SRTP / TLS
- SIP CANCEL, REGISTER, REFER
- G.711 PCMA (A-law) codec
- Opus codec support
- Full jitter buffer with reordering
- SIP authentication (digest)


rasonyang:
Hi maintainers 👋

Just a quick note that this PR has been **validated end-to-end in a local LAN environment using FreeSWITCH as a SIP endpoint** .  All tests described above are passing locally.

When convenient, could someone with write access please approve the pending workflows so CI can run? An initial review whenever you have time would be greatly appreciated.

Thanks for your time!

markbackman:
Hey! Thanks for the contribution 🙇 

I've tagged two of my colleagues to take a look this week.

In the meantime, can you:
- Lint this code: `uv run scripts/fix-ruff.sh` or install pre-commit hook: `uv run pre-commit install`
- Update the changelog to reflect this PR

examples/foundational/55zzo-sip-transport.py
```diff
@@ -0,0 +1,83 @@
...
+# SPDX-License-Identifier: BSD 2-Clause License
+#
+
+"""SIP Transport Example - Echo Bot
+
+Accepts incoming SIP calls and echoes received audio back to the caller.
+Useful for testing SIP connectivity and verifying audio round-trip.
+
+Requirements:
+    pip install "pipecat-ai[sip]"
```
markbackman: I don't see a `sip` optional-dependency. Is this needed or is the dependency missing?

examples/foundational/55zzo-sip-transport.py
```diff
@@ -0,0 +1,83 @@
+#
```
markbackman: Can you rename this to 56-sip-transport.py?

examples/foundational/55zzo-sip-transport.py
```diff
@@ -0,0 +1,83 @@
...
+    )
+
+    server = SIPServerTransport(params=params)
+
+    @server.event_handler("on_call_started")
+    async def on_call(server, call_transport: SIPCallTransport):
+        logger.info("Call started: %s", call_transport.session.call_id)
+
+        # Echo pipeline: input -> output (echoes audio back)
+        pipeline = Pipeline([call_transport.input(), call_transport.output()])
```
markbackman: This pipeline doesn't have an agent. It just makes a call. It would be ideal to use it to call.

In fact, it may be better to build out a full example in the `pipecat-examples` repo:
https://github.com/pipecat-ai/pipecat-examples

This example could show how to get set up on a provider like FreeSwitch or Asterisk. This would significantly improve the odds of developers using this integration. If we got this route, then a foundational example might not be required. I think this is better.

vr000m:
Thank you!

Below is just a stream of consciousness of things for us to check in the review, next week!

I think at a quick glance we could easily scope this as FreeswitchSIPtransport and that already reduces the burden of testing all sip features and incompatibility outside of Freeswitch’s settings/config and various SIP flavours. 

For freeswitch:

- Confirm that it can both receive calls and make calls (dial-in vs dial-out)

- Confirm STUN, IP discovery. when freeswitch and pipecat are on different networks/subnets, it's a very likely consideration that pipecat is within a K8s cluster or not a globally routable IP address (behind a firewall). And needs to figure out IPs. Also need to check if pipecat is advertising as "controlled" or "controlling" (vis-a-vis freeswitch configuration). 

- Confirm encryption, I think freeswitch handles the complexity with the external, need to check what's supported here. 

- Confirm how to register, i.e., process of creating the username/password (maybe out of scope), but how to pass that into pipecat bot the register with SIP network.

You've already indicated the codecs, dtmf send and receive over rfc2833 (should consider adding SIP INFO, since that would be needed unless Freeswitch handles both rtp and sip flavours of dtmf). 

- confirm what rtp dead timeout does -- confirm it's a rtp circuit breaker/keep alive using lack of rtp/rtcp as an indicator

- confirm SIP INVITEs as keep alives, ie what happens if pipecat gets the same invite again and again. Many systems do this aggressively, every 30s, every few minutes, some in 10a of minutes timescale. (If there are assumptions in the code, we should document those). 

Since I see _INVITE, duplicate rejection, max calls, OPTIONS, BYE (8 tests)_ not sure if this is freeswitch config (maybe this impacts generic sip and not freeswitch behaviour). 


On the topic of generic SIP support. Currently the remote sip user connects to freeswitch which independently handles the sip connection for that leg from the freeswitch-pipecat connection. This constraint reduces the risk of feature bloat but also limits the integration SIP to freeswitch features. 

A simple test with just a sip proxy and not a B2BUA would be useful to see what passes/fails. In this case, direct end to end sip tests without freeswitch acting as a B2BUA using something like kamailio or similar. 

This is not a blocker right now, especially if we reactor this as freeswitch sip, and then consider a path to enhance the scope.

examples/foundational/55zzo-sip-transport.py
rasonyang: Good catch — there's actually no `sip` optional dependency needed. The SIP transport is pure Python and only uses `numpy` (already a core dependency) for audio resampling. Removed the `pip install` line from the example.

examples/foundational/55zzo-sip-transport.py
rasonyang: Done — renamed to `56-sip-transport.py`.

examples/foundational/55zzo-sip-transport.py
rasonyang: Great point. I've replaced the echo bot with a full STT → LLM → TTS conversational agent (Deepgram + OpenAI + Cartesia with Silero VAD turn management), following the pattern in `07-interruptible.py`.

Agreed that a complete example with FreeSWITCH setup instructions belongs in `pipecat-examples` — I'll prepare that as a follow-up once the core transport is merged.

rasonyang:
Thanks @markbackman and @vr000m for the thorough review! Addressed all feedback in the latest push:

**Changes made:**
- **Lint fixed** — ran `ruff format` on `tests/test_sip_signaling.py`
- **Example renamed** to `56-sip-transport.py`
- **Example rewritten** with a full STT → LLM → TTS conversational agent pipeline (Deepgram + OpenAI + Cartesia + Silero VAD), following the `07-interruptible.py` pattern
- **Removed `pip install "pipecat-ai[sip]"`** — no optional dependency needed, it's pure Python + numpy (already a core dep)
- **Changelog updated** with PR number and scoped description

---

**Responding to @vr000m's scope questions:**

> Dial-in vs dial-out

This PR is **dial-in only** (UAS). Dial-out is possible via FreeSWITCH's `originate` command, where FreeSWITCH handles the outbound leg and bridges the call to Pipecat. Native UAC `INVITE` origination is out of scope for now.

> STUN, IP discovery, NAT traversal

This transport is **LAN-only** — no STUN/TURN, no ICE. It's designed for the common deployment where FreeSWITCH and Pipecat are on the same network/subnet. FreeSWITCH handles NAT traversal on its external-facing leg. Pipecat advertises the local IP from `sip_listen_host` in its SDP.

> Encryption

No SRTP/TLS in this PR. FreeSWITCH terminates encryption on the external leg and forwards plain RTP to Pipecat. This is the standard pattern for media servers on trusted internal networks.

> Registration

No SIP REGISTER support — Pipecat acts as a static UAS endpoint. FreeSWITCH is configured to route calls to Pipecat's IP:port directly (e.g., via a dialplan `bridge` action). No username/password registration needed.

> RTP dead timeout

Correct — it's a circuit breaker. If no RTP packets arrive for `rtp_dead_timeout_ms` (default 5s), we tear down the call. This handles the case where the remote party disappears without sending BYE.

> Re-INVITE handling

Currently, re-INVITEs from the same Call-ID are rejected as duplicates. This is a known limitation. FreeSWITCH typically doesn't send session-refresh re-INVITEs to backends, but if needed, this can be enhanced in a follow-up.

> Generic SIP / proxy-only (Kamailio/OpenSIPS)

The SIP implementation is actually generic — it's not FreeSWITCH-specific. It should work with any SIP proxy/B2BUA that sends standard INVITE → ACK → BYE flows with G.711 RTP. I've scoped the docs/changelog as "FreeSWITCH" because that's what's been tested, but it would work with OpenSIPS/Kamailio routing calls directly as well. Happy to add proxy-only testing as a follow-up.

> DTMF

RFC 2833 (RTP-based) DTMF detection is implemented. SIP INFO DTMF is not — FreeSWITCH can be configured to use RFC 2833 on its internal leg, which is the standard approach.

vr000m:
Hi @rasonyang, thanks for addressing the feedback so quickly!

I’m really excited about getting SIP support into the framework. To ensure we can merge this quickly without 'boiling the ocean' of a universal SIP implementation, I have a suggestion regarding the scope:

While much of this logic is technically generic, the current implementation relies significantly on FreeSWITCH’s B2BUA functionality to handle the heavy lifting (e.g., NAT traversal via local network assumptions and the lack of SIP REGISTER support).

To manage user expectations and get this landable ASAP, would you be open to prefixing the transport as `FreeSwitchSIPTransport` (or similar)?

My personal reason to suggest the path of the prefix is the following: it immediately signals the specific configuration and constraints (LAN-only, dial-plan dependency) to other developers. It bypasses the need for this PR to test and solve for every SIP edge case or proxy-only (Kamailio/OpenSIPS) scenario right now.

Future work by you or other developers can absolutely treat the SIP message parsing and state-machine logic in this PR to then build the missing features. 

By scoping it to FreeSWITCH now, we can get this into the hands of developers much faster. What do you think?

rasonyang:
Thanks @vr000m — I'm happy to scope this to FreeSWITCH to set expectations and land it faster.

I'll rename the public transport classes to `FreeSwitchSIPServerTransport` / `FreeSwitchSIPCallTransport` (and update exports, example, docs + changelog) to make the current constraints explicit (LAN-scoped, FreeSWITCH dialplan/B2BUA dependency, no REGISTER; not targeting proxy-only Kamailio/OpenSIPS scenarios yet).

The SIP parsing / state-machine pieces can remain generic internally so we (or others) can build a more universal SIP transport incrementally later.

vr000m:
just ran /spec-compliance skill on rtp.py

  Spec Compliance Report

  Code: rtp.py (303 lines)
  Specs checked: RFC 3550 §5.1, RFC 4585 §6.2.1, RFC 5104 §4.3.1, RFC 8083 §4

  ---
  RFC 3550, Section 5.1 — "RTP Fixed Header Fields"

  Source: RFC 3550 §5.1
  Requirements found: 10 (5 MUST, 4 SHOULD, 1 MAY)

  MUST Requirements

  1. ✅ Met — "If the extension bit is set, the fixed header MUST be followed by exactly one header extension"
  Evidence: line 55 — 0x80 sets V=2 with no extension bit, so the requirement is satisfied by not setting the bit.
  2. ➖ N/A — "A receiver MUST ignore packets with payload types that it does not understand"
  Line 193 checks pt == DTMF_PAYLOAD_TYPE and lines 205-207 assume G.711 for all other PTs. Unknown PTs are decoded as G.711 rather than ignored — but this is a narrow SIP/G.711 implementation, not a general-purpose RTP stack.
  3. ⚠️ Partial — "The sampling instant MUST be derived from a clock that increments monotonically"
  Line 263 — timestamp increments by SAMPLES_PER_FRAME (160) each frame, which is monotonic. However, it's derived from a counter, not from an actual sampling clock.
  4. ✅ Met — "The resolution of the clock MUST be sufficient for the desired synchronization accuracy" 
  8kHz clock with 160 samples/frame = 20ms resolution, appropriate for G.711 telephony.
  6. ⚠️ Partial — "All RTP implementations must be prepared to detect and resolve collisions"
  No SSRC collision detection. Line 133 picks a random SSRC but never checks for collisions from incoming packets.

  SHOULD Requirements

  7. ✅ Met — "The initial value of the sequence number SHOULD be random (unpredictable)"
  Evidence: line 134 — self._seq = random.randint(0, 0xFFFF)
  8. ✅ Met — "The initial value of the timestamp SHOULD be random"
  Evidence: line 135 — self._timestamp = random.randint(0, 0xFFFFFFFF)
  9. ✅ Met — "The SSRC identifier SHOULD be chosen randomly"
  Evidence: line 133 — self._ssrc = random.randint(0, 0xFFFFFFFF)
  10. ⚠️ Partial — "If a source changes its source transport address, it must also choose a new SSRC
  identifier"
  start() accepts a new remote_addr but reuses the same SSRC. No new SSRC is generated on address change.

```
Summary — RFC 3550 §5.1

  ┌────────┬─────┬─────────┬─────────┬─────┬───────┐
  │ Level  │ Met │ Missing │ Partial │ N/A │ Total │
  ├────────┼─────┼─────────┼─────────┼─────┼───────┤
  │ MUST   │ 2   │ 0       │ 2       │ 1   │ 5     │
  ├────────┼─────┼─────────┼─────────┼─────┼───────┤
  │ SHOULD │ 3   │ 0       │ 1       │ 0   │ 4     │
  ├────────┼─────┼─────────┼─────────┼─────┼───────┤
  │ MAY    │ 0   │ 0       │ 0       │ 0   │ 0     │
  └────────┴─────┴─────────┴─────────┴─────┴───────┘
```

This code does not implement RTCP feedback at all. No NACK/RED either

  ---
  Overall Summary
```
  ┌─────────────────┬─────┬─────────┬─────────┬─────┬───────┐
  │      Spec       │ Met │ Missing │ Partial │ N/A │ Total │
  ├─────────────────┼─────┼─────────┼─────────┼─────┼───────┤
  │ RFC 3550 §5.1   │ 5   │ 0       │ 3       │ 1   │ 9     │
  ├─────────────────┼─────┼─────────┼─────────┼─────┼───────┤
  │ RFC 4585 §6.2.1 │ 0   │ 0       │ 0       │ 3   │ 3     │
  ├─────────────────┼─────┼─────────┼─────────┼─────┼───────┤
  │ RFC 5104 §4.3.1 │ 0   │ 0       │ 0       │ 6   │ 6     │
  ├─────────────────┼─────┼─────────┼─────────┼─────┼───────┤
  │ RFC 8083 §4     │ 0   │ 10      │ 0       │ 0   │ 10    │
  ├─────────────────┼─────┼─────────┼─────────┼─────┼───────┤
  │ Total           │ 5   │ 10      │ 3       │ 10  │ 28    │
  └─────────────────┴─────┴─────────┴─────────┴─────┴───────┘
```
  Key Findings

RFC 3550 — Good basic compliance. Random SSRC/seq/timestamp initialization is correct. Missing: SSRC collision detection (common gap in minimal implementations) and unknown PT handling.

This is a basic G.711 audio RTP session — no RTCP feedback, no video codec control. These specs would be relevant if the code grew to support AVPF or video.


  Sources:
  - RFC 3550

rasonyang:
Thanks for running the spec compliance check @vr000m! Pushed a commit addressing the three RFC 3550 §5.1 partial-compliance items:

1. **Unknown payload types** (MUST) — `_handle_packet()` now ignores packets with PTs other than PCMU (0) and DTMF (101), instead of blindly decoding as G.711.
2. **SSRC collision detection** (MUST) — If an incoming packet carries our own SSRC, we regenerate. Minimal implementation suitable for 1:1 SIP calls.
3. **New SSRC on address change** (SHOULD) — `start()` now regenerates the SSRC when the remote transport address changes.

For the **RFC 8083 (RTCP circuit breaker)** gap — this is intentionally out of scope for a LAN-only FreeSWITCH transport where congestion control is handled at the network level. I've added a "Known limitations" docstring to `RTPSession` documenting this along with the no-NACK/AVPF and frame-counter timestamp limitations.

All changes have tests in `TestRTPSessionCompliance` (`tests/test_sip_rtp.py`).

vr000m:
As long as the rtp session doesn't timeout due to lack of rtcp, we should be okay. 

I didn't run a spec check on sip because there are quite many sip specs. 3261 probably makes sense, but it would be good to know what conformance freeswitch needs (since it's B2BUA) we only need whatever is needed on the internal FS side. Not on the external SIP.

rasonyang:
This transport intentionally targets a minimal FreeSWITCH → Pipecat dial-in flow in a LAN environment.
Since FreeSWITCH is a B2BUA, we only implement the subset required on the internal FS side rather than a full SIP stack.
Happy to expand or fix issues if the community runs into interoperability problems.

rasonyang (commented):
## Code review

### Bug: PCMA codec advertised but never actually used

The `codec` field is stored in `FreeSwitchSIPSession` from the negotiated codec but is never passed to `RTPSession` or used anywhere in the RTP layer.

**Evidence:**

1. `transport.py:573` passes `codec=negotiated_codec` to `FreeSwitchSIPSession`
2. `transport.py:99-103` creates `RTPSession` without passing the codec
3. `rtp.py:280` calls `pack_rtp_header(...)` without specifying `payload_type`, defaulting to `0` (PCMU)
4. `sdp.py:77-78` hardcodes PCMU only in the SDP answer regardless of `codec_preferences`

**Impact:** If a client requests PCMA, the SDP answer will only advertise PCMU, and the RTP layer will always send PCMU regardless of what was negotiated.

**Suggested fix:** Either remove PCMA from `CODEC_PT` and `codec_preferences` until it's implemented, or wire up the codec through to `RTPSession`.

---

vr000m:
Thank you for the quick turnaround. I ran the spec compliance on the rest of the files. I think these are known limitations (no RTCP, no ICE/STUN, no SIP REGISTER) but within the deployment framework that you have in mind.

Leaving this as a future note as a good starting point

| File | Requirement | RFC | Status | Notes |
|---|---|---|---|---|
| rtp.py | Marker bit handling | RFC 3550 §5.1 | Not implemented | Marker bit is always 0. RFC 3551 §4.1 says the marker bit SHOULD be set on the first packet after silence suppression. Acceptable for continuous audio, only needed in case of video. |
| rtp.py | CSRC handling | RFC 3550 §5.1 | N/A | CC=0, no mixers — correct for point-to-point. not needed until multiple participants in the call|
| rtp.py | RTCP | RFC 3550 §6 | Not implemented | Explicitly documented as out-of-scope (LAN-only). This is a known deviation. RFC 3550 says RTCP "SHOULD" be used. For LAN-only deployments with FreeSWITCH this is pragmatically fine. This is okay for telephony cases unless the remote party times out the connection when no RTCP is received. |
| rtp.py | Dynamic PT negotiation via SDP | RFC 4733 §5 | Hardcoded | PT=101 is hardcoded, not negotiated from SDP `a=rtpmap:101 telephone-event/8000`. Works with FreeSWITCH defaults but technically should be negotiated. |
| sdp.py | SDP answer matches offer codecs | RFC 3264 §6.1 | Simplified | Answer always offers PCMU regardless of what's in the offer. If the offerer doesn't support PCMU, the call will fail. Acceptable for FreeSWITCH (always supports PCMU). |
| signaling.py | Branch parameter in Via (UAS BYE) | RFC 3261 §8.1.1.7 | Simplified | Uses `z9hG4bK{call_id[:8]}` — the magic cookie is correct but the uniqueness comes from truncated call-id. Should be fine in practice but not cryptographically unique per spec. |
| signaling.py | Header parsing (case sensitivity) | RFC 3261 §7.3.1 | Case-sensitive | The parser does exact-case matching for header names. RFC 3261 says header names are case-insensitive. E.g., `call-id:` would not match `Call-ID`. FreeSWITCH uses standard casing, so this works in practice. |

rasonyang:
Thanks — this is very helpful. Agreed these are known scope boundaries, as the implementation is optimized for controlled FreeSWITCH / LAN deployments rather than full RFC completeness.

The SIP transport is already production-proven in our environment. I’ll track these as follow-up issues — feel free to open issues and assign them to me.

Appreciate you documenting this so clearly.

vr000m:
To run or recreate your set up. Is there docs or docker image that we can run and test locally for example.

rasonyang:
Not yet — no Docker setup in the repo currently. I'll prepare a `docker-compose` environment in `pipecat-examples` as a follow-up.

In the meantime, here's the minimal FreeSWITCH config to route calls to the Pipecat SIP transport:

**1. Gateway** — `sip_profiles/external/pipecat.xml`
```xml
<include>
  <gateway name="pipecat">
    <param name="proxy" value="sip:PIPECAT_IP:6060"/>
    <param name="register" value="false"/>
  </gateway>
</include>
```

**2. Dialplan** — add to `dialplan/public.xml`
```xml
<extension name="pipecat-bot">
  <condition field="destination_number" expression="^8888$">
    <action application="bridge" data="sofia/gateway/pipecat/8888"/>
  </condition>
</extension>
```

**3. vars.xml** — ensure codec is set to PCMU (only codec supported currently):
```xml
<X-PRE-PROCESS cmd="set" data="global_codec_prefs=PCMU"/>
```

Then run: `SIP_PORT=6060 python examples/foundational/56-freeswitch-sip-transport.py`

Dial `8888` from a SIP phone registered to FreeSWITCH and it will bridge to the Pipecat bot.

hdduytran:
@rasonyang when will you support REGISTER?

filipi87 (commented):
Hi, I believe this would be great as a community integration! That way, it could be maintained as a separate repository and package, making it easier for others to use.

- https://docs.pipecat.ai/api-reference/server/services/community-integrations
- https://github.com/pipecat-ai/pipecat/blob/main/COMMUNITY_INTEGRATIONS.md

cc: @mark for thoughts.

markbackman:
Agreed! This would be a great community integration.

rasonyang:
> @rasonyang when will you support REGISTER?
Thanks for asking!

REGISTER is not planned for this PR. This transport is intentionally scoped as a LAN/internal SIP endpoint: Pipecat acts as a static UAS and receives direct INVITEs from FreeSWITCH or another SIP server configured to route calls to its IP:port.

In our target deployment, SIP registration, authentication, NAT traversal, keepalives, and external SIP trunk handling should be managed by a dedicated SIP server such as OpenSIPS, Kamailio, or FreeSWITCH. That SIP server can then route the call to Pipecat via a direct SIP INVITE on the internal network.

Pipecat only handles the internal signaling/media leg needed for the bot pipeline, so no username/password registration is required in this architecture.

REGISTER support would move this transport toward a more generic SIP client/UAC implementation, which is a larger scope. I think it would be better tracked as a separate follow-up feature if there is enough demand.

hdduytran:
@rasonyang  How do you think about outbound call from pipecat to human?  Can I do this in scope of this integration?

rasonyang:
@hdduytran 

For outbound calls, I would keep Pipecat scoped as an inbound-only UAS.

The flow would be:

1. FreeSWITCH acts as the B2BUA and originates the outbound call to the human first.
2. If the human answers, FreeSWITCH then bridges or invites Pipecat as the internal SIP endpoint.
3. Pipecat only accepts inbound SIP INVITEs and handles the bot media leg.

This has two advantages:

1. Call control stays in FreeSWITCH/API, including dialing, timeout, retry, transfer, conference, hangup, and state tracking.
2. If the human does not answer, we do not start or reserve Pipecat/AI resources.

So I do not think Pipecat needs to originate outbound calls directly. Keeping Pipecat as UAS-only should be simpler and cleaner for this integration.

filipi87:

Closing this out for now.

PR #3863, opened 2026-02-27
Manual context summarization

filipi87:
## Summary

- Added `LLMSummarizeContextFrame`: push this frame anywhere in the pipeline to trigger on-demand context summarization (e.g. from a function call tool).

- Refactored `LLMContextSummarizationConfig` into two focused classes:
  - `LLMContextSummaryConfig` — summary *generation* params (`target_context_tokens`, `min_messages_after_summary`, `summarization_prompt`); shared by both auto and manual modes.
  - `LLMAutoContextSummarizationConfig` — auto-trigger *thresholds* (`max_context_tokens`, `max_unsummarized_messages`) plus a nested `summary_config: LLMContextSummaryConfig`.

- Renamed `LLMAssistantAggregatorParams` fields for clarity:
  - `enable_context_summarization` → `enable_auto_context_summarization`
  - `context_summarization_config` → `auto_context_summarization_config`

- `LLMContextSummarizationConfig` is now deprecated (emits `DeprecationWarning`); old `LLMAssistantAggregatorParams` field names are kept with deprecation warnings and auto-migration for one release cycle.

- Added example `54b-context-summarization-manual-openai.py` demonstrating on-demand summarization triggered via a function call tool.

## Breaking Changes

- `LLMAssistantAggregatorParams.enable_context_summarization` renamed to `enable_auto_context_summarization`.

- `LLMAssistantAggregatorParams.context_summarization_config` renamed to `auto_context_summarization_config` and now accepts `LLMAutoContextSummarizationConfig` instead of `LLMContextSummarizationConfig`.

### Migration

```python
# Before
LLMAssistantAggregatorParams(
    enable_context_summarization=True,
    context_summarization_config=LLMContextSummarizationConfig(
        max_context_tokens=8000,
        target_context_tokens=6000,
        max_unsummarized_messages=20,
        min_messages_after_summary=4,
    ),
)

# After
LLMAssistantAggregatorParams(
    enable_auto_context_summarization=True,
    auto_context_summarization_config=LLMAutoContextSummarizationConfig(
        max_context_tokens=8000,
        max_unsummarized_messages=20,
        summary_config=LLMContextSummaryConfig(
            target_context_tokens=6000,
            min_messages_after_summary=4,
        ),
    ),
)
```

##  Testing

Run the new example to see manual summarization in action:

`uv run examples/foundational/54b-context-summarization-manual-openai.py`

Ask the bot to "summarize the conversation" — it will call the `summarize_conversation` function, which pushes an `LLMContextSummaryRequestFrame` into the pipeline. The LLM service generates the summary in a background task and the assistant aggregator compresses the context when the result arrives.

### DIFF
--- a/tests/test_context_summarization.py
+++ b/tests/test_context_summarization.py
@@ -14,8 +14,10 @@ from pipecat.frames.frames import LLMContextSummaryRequestFrame, LLMContextSumma
 from pipecat.processors.aggregators.llm_context import LLMContext, LLMSpecificMessage
 from pipecat.services.llm_service import LLMService
 from pipecat.utils.context.llm_context_summarization import (
+    LLMAutoContextSummarizationConfig,
     LLMContextSummarizationConfig,
     LLMContextSummarizationUtil,
+    LLMContextSummaryConfig,
 )
 
 
@@ -167,41 +169,107 @@ class TestContextSummarizationMixin(unittest.TestCase):
         self.assertIn("USER: First part Second part", transcript)
 
 
-class TestLLMContextSummarizationConfig(unittest.TestCase):
-    """Tests for LLMContextSummarizationConfig."""
+class TestLLMContextSummaryConfig(unittest.TestCase):
+    """Tests for LLMContextSummaryConfig."""
 
     def test_default_config(self):
         """Test default configuration values."""
-        config = LLMContextSummarizationConfig()
+        config = LLMContextSummaryConfig()
 
-        self.assertEqual(config.max_context_tokens, 8000)
-        self.assertEqual(config.max_unsummarized_messages, 20)
+        self.assertEqual(config.target_context_tokens, 6000)
         self.assertEqual(config.min_messages_after_summary, 4)
         self.assertIsNone(config.summarization_prompt)
 
     def test_custom_config(self):
         """Test custom configuration."""
-        config = LLMContextSummarizationConfig(
-            max_context_tokens=2500,
+        config = LLMContextSummaryConfig(
             target_context_tokens=2000,
-            max_unsummarized_messages=15,
             min_messages_after_summary=4,
             summarization_prompt="Custom prompt",
         )
 
-        self.assertEqual(config.max_context_tokens, 2500)
         self.assertEqual(config.target_context_tokens, 2000)
-        self.assertEqual(config.max_unsummarized_messages, 15)
         self.assertEqual(config.min_messages_after_summary, 4)
         self.assertEqual(config.summary_prompt, "Custom prompt")
 
     def test_summary_prompt_property(self):
         """Test summary_prompt property uses default when None."""
-        config = LLMContextSummarizationConfig()
+        config = LLMContextSummaryConfig()
         self.assertIn("summarizing a conversation", config.summary_prompt.lower())
 
-        config_with_custom = LLMContextSummarizationConfig(summarization_prompt="Custom")
+        config_with_custom = LLMContextSummaryConfig(summarization_prompt="Custom")
         self.assertEqual(config_with_custom.summary_prompt, "Custom")
+
+
+class TestLLMAutoContextSummarizationConfig(unittest.TestCase):
+    """Tests for LLMAutoContextSummarizationConfig."""
+
+    def test_default_config(self):
+        """Test default configuration values."""
+        config = LLMAutoContextSummarizationConfig()
+
+        self.assertEqual(config.max_context_tokens, 8000)
+        self.assertEqual(config.max_unsummarized_messages, 20)
+        self.assertEqual(config.summary_config.target_context_tokens, 6000)
+        self.assertEqual(config.summary_config.min_messages_after_summary, 4)
+
+    def test_custom_config(self):
+        """Test custom configuration."""
+        config = LLMAutoContextSummarizationConfig(
+            max_context_tokens=2500,
+            max_unsummarized_messages=15,
+            summary_config=LLMContextSummaryConfig(
+                target_context_tokens=2000,
+                min_messages_after_summary=4,
+                summarization_prompt="Custom prompt",
+            ),
+        )
+
+        self.assertEqual(config.max_context_tokens, 2500)
+        self.assertEqual(config.max_unsummarized_messages, 15)
+        self.assertEqual(config.summary_config.target_context_tokens, 2000)
+        self.assertEqual(config.summary_config.min_messages_after_summary, 4)
+        self.assertEqual(config.summary_config.summary_prompt, "Custom prompt")
+
+    def test_target_tokens_auto_adjusted(self):
+        """Test that target_context_tokens is auto-adjusted when it exceeds max."""
+        config = LLMAutoContextSummarizationConfig(
+            max_context_tokens=1000,
+            summary_config=LLMContextSummaryConfig(target_context_tokens=9000),
+        )
+        self.assertLessEqual(config.summary_config.target_context_tokens, config.max_context_tokens)
+
+
+class TestLLMContextSummarizationConfigDeprecated(unittest.TestCase):
+    """Tests for deprecated LLMContextSummarizationConfig."""
+
+    def test_emits_deprecation_warning(self):
+        """Test that instantiating the deprecated config emits a DeprecationWarning."""
+        with self.assertWarns(DeprecationWarning):
+            LLMContextSummarizationConfig()
+
+    def test_to_auto_config(self):
+        """Test conversion to the new LLMAutoContextSummarizationConfig."""
+        import warnings
+
+        with warnings.catch_warnings():
+            warnings.simplefilter("ignore", DeprecationWarning)
+            old_config = LLMContextSummarizationConfig(
+                max_context_tokens=2500,
+                target_context_tokens=2000,
+                max_unsummarized_messages=15,
+                min_messages_after_summary=4,
+                summarization_prompt="Custom",
+            )
+
+        new_config = old_config.to_auto_config()
+
+        self.assertIsInstance(new_config, LLMAutoContextSummarizationConfig)
+        self.assertEqual(new_config.max_context_tokens, 2500)
+        self.assertEqual(new_config.max_unsummarized_messages, 15)
+        self.assertEqual(new_config.summary_config.target_context_tokens, 2000)
+        self.assertEqual(new_config.summary_config.min_messages_after_summary, 4)
+        self.assertEqual(new_config.summary_config.summarization_prompt, "Custom")
 
 
 class TestFunctionCallHandling(unittest.TestCase):
@@ -670,10 +738,12 @@ class TestDedicatedLLMSummarization(unittest.IsolatedAsyncioTestCase):
                 {"role": "user", "content": f"Test message {i} that adds tokens to context."}
             )
 
-        config = LLMContextSummarizationConfig(
+        config = LLMAutoContextSummarizationConfig(
             max_context_tokens=50,  # Very low to trigger easily
-            llm=dedicated_llm,
-            summarization_timeout=5.0,
+            summary_config=LLMContextSummaryConfig(
+                llm=dedicated_llm,
+                summarization_timeout=5.0,
+            ),
         )
         return context, config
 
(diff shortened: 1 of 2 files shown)


examples/foundational/54b-context-summarization-manual-openai.py
```diff
@@ -0,0 +1,193 @@
...
+        audio_out_enabled=True,
+    ),
+}
+
+
+async def summarize_conversation(params: FunctionCallParams):
+    """Trigger manual context summarization via a pipeline frame."""
+    logger.info("Tool called: summarize_conversation")
+
+    summarization_config = LLMContextSummarizationConfig()
```
kompfner: Having the user instantiate an `LLMContextSummarizationConfig` with no arguments just to then grab default values off of it feels a little awkward.

I wonder if `LLMContextSummaryRequestFrame` could internally create a `LLMContextSummarizationConfig` as a defaults provider in the case the user hasn't provided all fields?

examples/foundational/54b-context-summarization-manual-openai.py
```diff
@@ -0,0 +1,193 @@
...
+    request_frame = LLMContextSummaryRequestFrame(
+        request_id=str(uuid.uuid4()),
+        context=params.context,
```
kompfner: Our public frame-based APIs that affect the context (like `LLMMessagesAppendFrame`, etc) don't generally require the user to have to pass the context. For one, it's already clear which context they intend to update. Secondly, since it's a frame-based API, the effects occur at the time the frame is processed; an API that takes a context might imply that it will act on a "snapshot" of the context given to it at that time, which isn't how we want users thinking about it.

I understand why you chose to make `LLMContextSummaryRequestFrame` the public interface for this new manual summarization functionality: it already exists, and is used internally. But the use-case we want to enable is subtly different.
* `LLMContextSummaryRequestFrame` is used ask an LLM to generate a summary of a given context
* What we need is a new way to trigger the pipeline (the aggregator? the summarizer?) to then trigger the same code path that an auto summarization would trigger. What if we had a new `LLMSummarizeContext` frame that `LLMContextSummarizer.process_frame` handled, causing it to trigger a request summarization (with frame-provided options)? This approach would have a few advantages:
  * it would make greater re-use of more existing code paths
  * it would not require users to generate a `request_id` either
  * it would update the `LLMContextSummarizer`'s internal bookkeeping (like `_summarization_in_progress`), which I assume is valuable

src/pipecat/processors/aggregators/llm_context_summarizer.py
```diff
@@ -30,9 +30,16 @@
...
 
-    This class manages automatic context summarization when token or message
-    limits are reached. It monitors the LLM context size, triggers
-    summarization requests, and applies the results to compress conversation history.
+    This class manages context summarization, either automatically when token or
+    message limits are reached, or on-demand when a manual summarization is
+    requested. It monitors the LLM context size, triggers summarization requests,
+    and applies the results to compress conversation history.
+
+    When ``auto_trigger=True`` (the default), summarization is triggered
```
kompfner, line 65: Do these need to be exclusive? I could very easily see a desire to auto-summarize and support manual summaries.

examples/foundational/54b-context-summarization-manual-openai.py
markbackman: +1 to this feeback

src/pipecat/processors/aggregators/llm_context_summarizer.py
```diff
@@ -56,17 +63,23 @@ def __init__(
         *,
         context: LLMContext,
         config: Optional[LLMContextSummarizationConfig] = None,
```
kompfner: I think `LLMContextSummarizationConfig` could use a refactor to separate out params related to triggering auto-summarization (e.g. `max_context_tokens`) and those relevant to the summary generation itself and therefore relevant to both auto and manual summarization (e.g. `summarization_prompt`, `min_messages_after_summary`). Otherwise this will be messy to maintain.

Maybe we have something like:
* `LLMAutoContextSummarizationConfig`, which contains the auto trigger related fields, as well as a nested field that is of type:
* `LLMContextSummaryConfig`, which contains the summary generation related fields

kompfner: I would then rename this field `auto_summary_config` or something to distinguish it from the new summary config, which will be able to differ between each manual summarization request

src/pipecat/processors/aggregators/llm_response_universal.py
```diff
@@ -824,16 +824,18 @@ def __init__(
...
-                context=self._context,
-                config=self._params.context_summarization_config,
-            )
-            self._summarizer.add_event_handler(
-                "on_request_summarization", self._on_request_summarization
-            )
+        # Context summarization — always create the summarizer so that manually
+        # pushed LLMContextSummaryRequestFrame results are always handled.
+        # Auto-triggering based on thresholds is only enabled when
+        # enable_context_summarization is True.
```
kompfner: I think we should rename to `auto_context_summarization`, and rename `context_summarization_config` to `auto_context_summarization_config`, to disambiguate between the per-summarization-request configs the user can provide.

src/pipecat/processors/aggregators/llm_context_summarizer.py
kompfner: Actually, another benefit of the above suggested refactor: we could then use `LLMContextSummaryConfig` _directly_ in the new frame that users use to request a summary 💡 .

examples/foundational/54b-context-summarization-manual-openai.py
kompfner: See a related idea in another comment about refactoring `LLMContextSummarizationConfig`, below.

src/pipecat/processors/aggregators/llm_context_summarizer.py
kompfner: Ah, looks like maybe they're not exclusive.

examples/foundational/54b-context-summarization-manual-openai.py
kompfner: Ah...the more I think about it, we really do need to make sure  `LLMContextSummarizer`'s bookkeeping is updated on a manual summarization, since you can have _both_ automatic and manual summarization going on in an app.

examples/foundational/54b-context-summarization-manual-openai.py
filipi87: Agreed. 👍

src/pipecat/processors/aggregators/llm_context_summarizer.py
filipi87: Refactored how we were handling it.

examples/foundational/54b-context-summarization-manual-openai.py
filipi87: Done.

examples/foundational/54b-context-summarization-manual-openai.py
filipi87: Yep. I believe I have implemented all the suggestions that you did here.

src/pipecat/processors/aggregators/llm_context_summarizer.py
filipi87: I believe I have implemented all the suggestions that you did here. Let me know if this is what you had in mind.

src/pipecat/processors/aggregators/llm_response_universal.py
filipi87: I have renamed it to `enable_auto_context_summarization` and `auto_context_summarization_config`. 

We usually use the `enable_` prefix for features that we are turning on, so I thought it would be better to do the same here.

filipi87:
@markbackman, @kompfner, I think it’s ready for another round of review. 

I have also already merged in the latest changes from main, which include Mark’s updates to context summarization.

changelog/3863.added.md
```diff
@@ -0,0 +1 @@
+- Added `LLMSummarizeContextFrame` to trigger on-demand context summarization from anywhere in the pipeline (e.g. a function call tool) without needing to supply a context reference or generate a request ID. Accepts an optional `config: LLMContextSummaryConfig` to override summary generation settings per request.
```
kompfner: I don't think we need to leak that we were ever "supplying a context reference or generating a request id"—that was a previous draft of this work

changelog/3863.deprecated.md
```diff
@@ -0,0 +1 @@
+- Deprecated `LLMContextSummarizationConfig`. Use `LLMAutoContextSummarizationConfig` with a nested `LLMContextSummaryConfig` instead. The old class emits a `DeprecationWarning` and provides a `.to_auto_config()` helper for migration.
```
kompfner: Wonder if `.to_auto_config()` will actually hurt us more than help, long-term (in that folks won't really migrate off the old class). The migration should be pretty trivial.

examples/foundational/54b-context-summarization-manual-openai.py
```diff
@@ -0,0 +1,179 @@
...
+        audio_out_enabled=True,
+    ),
+}
+
+
+async def summarize_conversation(params: FunctionCallParams):
+    """Trigger manual context summarization via a pipeline frame."""
+    logger.info("Tool called: summarize_conversation")
+    await params.result_callback({"status": "summarization_requested"})
+    await params.llm.queue_frame(LLMSummarizeContextFrame())
```
kompfner, line 73: yassss

src/pipecat/frames/frames.py
```diff
@@ -2000,6 +2001,26 @@ class LLMAssistantPushAggregationFrame(ControlFrame):
     """
 
 
+@dataclass
+class LLMSummarizeContextFrame(ControlFrame):
+    """Frame requesting on-demand context summarization.
+
+    Push this frame into the pipeline to trigger a manual context summarization.
+    The ``LLMContextSummarizer`` handles it internally — no need to supply a
```
kompfner: Again, feels like we're leaking details about a previous draft of this work. Thanks, Claude 😛

src/pipecat/processors/aggregators/llm_context_summarizer.py
```diff
@@ -88,18 +101,26 @@ def __init__(
...
-        config: Optional[LLMContextSummarizationConfig] = None,
+        config: Optional[LLMAutoContextSummarizationConfig] = None,
+        auto_trigger: bool = True,
     ):
         """Initialize the context summarizer.
 
         Args:
             context: The LLM context to monitor and summarize.
-            config: Configuration for summarization behavior. If None, uses default config.
+            config: Auto-summarization configuration controlling both trigger
```
kompfner, line 111: nit: do we want to call this `auto_config` or something other than `config`, which might imply `LLMContextSummaryConfig`? (I don't feel super strong about this, as it's internal)

src/pipecat/processors/aggregators/llm_context_summarizer.py
```diff
@@ -151,12 +174,24 @@ async def _handle_llm_response_start(self, frame: LLMFullResponseStartFrame):
...
+        so bookkeeping (``_summarization_in_progress``,
+        ``_pending_summary_request_id``) is always updated correctly.
 
         Args:
-            frame: The interruption frame.
+            frame: The manual summarization request frame, optionally carrying
+                a per-request :class:`~pipecat.utils.context.llm_context_summarization.LLMContextSummaryConfig`.
         """
+        if self._summarization_in_progress:
+            logger.debug(f"{self}: Summarization already in progress, ignoring manual request")
```
kompfner, line 189: 👍

src/pipecat/frames/frames.py
markbackman: Claude loves this.

changelog/3863.deprecated.md
markbackman: I wasn't as bothered by this. It seems more of an internal convenience.

kompfner: Ohhh wait, just realized—we need this *internally*, to continue supporting folks providing `LLMContextSummarizationConfig`s...

kompfner: OK, but then I wouldn't advertise it as something users should actually use in _their_ code.

filipi87: Exactly, we need this for now.

src/pipecat/frames/frames.py
filipi87: Fixed.

changelog/3863.added.md
filipi87: Fixed.

changelog/3863.deprecated.md
filipi87: Fixed the message.

filipi87:
Thank you @kompfner , @markbackman for the great review. 🙌🚀

PR #3868, opened 2026-02-27
fix: Update numba version pin from == to >=0.61.2

wollerman:
Previously the numba version was pinned to `0.61.2`. [However, reading the commit and corresponding changelog information](https://github.com/pipecat-ai/pipecat/commit/3a306dae90c2c77bfd9fdde4ecdb3a531005e753), it looks like it was intended to be `>=`.

This PR updates the pin to `>=`.  This allows wider version support for numba and therefore numpy versions for other custom processing.

### DIFF
--- a/changelog/3868.changed.md
+++ b/changelog/3868.changed.md
@@ -1,0 +1,1 @@
+- Updated numba version pin from == to >=0.61.2


markbackman (approved):
LGTM! Thank you.

PR #3869, opened 2026-02-27
Fix tracing crashes on dataclass settings (.items() called on non-dict objects)

### LINKED ISSUE #3858: Tracing crashes on dataclass settings: .items() called on non-dict objects
## Bug Description

The tracing utilities call `.items()` on service settings objects (`GoogleLLMSettings`, `InworldTTSSettings`, etc.), but these are `@dataclass` instances that don't have an `.items()` method. This causes repeated warnings/errors on every LLM and TTS call.

## Error Messages

```
WARNING:root:Error setting up LLM tracing: 'GoogleLLMSettings' object has no attribute 'items'
WARNING:root:Error in TTS tracing: 'InworldTTSSettings' object has no attribute 'items'
ERROR:root:Error in TTS tracing (continuing without tracing): 'InworldTTSSettings' object has no attribute 'items'
```

## Version

`pipecat-ai==0.0.104.dev111`

## Root Cause

**4 call sites** assume settings is a dict when it's actually a dataclass:

### 1. `pipecat/utils/tracing/service_decorators.py` ~line 513
```python
# In the LLM tracing decorator
if hasattr(self, "_settings"):
    for key, value in self._settings.items():  # ← AttributeError
```

### 2. `pipecat/utils/tracing/service_attributes.py` ~line 110 (`add_tts_span_attributes`)
```python
if settings:
    for key, value in settings.items():  # ← AttributeError
```

### 3. `pipecat/utils/tracing/service_attributes.py` ~line 174 (`add_stt_span_attributes`)
```python
for key, value in settings.items():  # ← same pattern
```

### 4. `pipecat/utils/tracing/service_attributes.py` ~line 362
```python
# Same .items() pattern on settings
```

## Suggested Fix

The `ServiceSettings` base class already provides a `given_fields()` method that returns a dict. The tracing code should either:

1. Use `settings.given_fields()` if available, or
2. Use `dataclasses.asdict(settings)` as a fallback, or
3. Check `hasattr(settings, 'items')` before calling it

Example:
```python
if settings:
    fields = settings.given_fields() if hasattr(settings, 'given_fields') else (settings if isinstance(settings, dict) else {})
    for key, value in fields.items():
        if isinstance(value, (str, int, float, bool)):
            span.set_attribute(f"settings.{key}", value)
```

## Impact

- **Not a blocker** — tracing errors are caught and execution continues
- **Very noisy** — every LLM and TTS call logs WARNING + ERROR lines
- Affects all services that use dataclass-based settings (Google LLM, Google STT, Inworld TTS, likely others)

kollaikal-rupesh:
## Summary

- Added a shared `_settings_to_dict()` helper that safely converts service settings to a plain dict, whether the settings object is a dict or a dataclass
- Fixed all four call sites where `.items()` was called directly on dataclass settings objects
- Tracing no longer crashes on services that use dataclass-based settings (Google LLM, Google STT, Inworld TTS, and others)

## Root Cause

The tracing utilities assumed `self._settings` is always a dict and called `.items()` on it directly. However, many services use dataclass instances for their settings (e.g. `GoogleLLMSettings`, `InworldTTSSettings`). Dataclass objects do not have an `.items()` method, so this raised `AttributeError` on every traced LLM, TTS, and STT call.

The errors were caught and logged as warnings, so execution continued, but the log noise was significant in production:

```
WARNING:root:Error setting up LLM tracing: 'GoogleLLMSettings' object has no attribute 'items'
WARNING:root:Error in TTS tracing: 'InworldTTSSettings' object has no attribute 'items'
```

## Approach

Added a `_settings_to_dict()` helper in `service_attributes.py` that normalises settings objects to plain dicts:

1. If the object is already a `dict`, return it as-is
2. If it has `given_fields()` (the `ServiceSettings` API), use that
3. If it is a dataclass, fall back to `dataclasses.asdict()`
4. Otherwise return an empty dict

This follows the existing `ServiceSettings.given_fields()` pattern already used elsewhere in the codebase (e.g. `fal/stt.py`).

### Files changed

| File | Change |
|------|--------|
| `src/pipecat/utils/tracing/service_attributes.py` | Added `_settings_to_dict()` helper, fixed 3 call sites (`add_tts_span_attributes`, `add_stt_span_attributes`, `add_gemini_live_span_attributes`) |
| `src/pipecat/utils/tracing/service_decorators.py` | Fixed 1 call site (LLM tracing decorator), imported the helper |

## Fixes

Fixes #3858

## Testing

- All existing tests pass (575 passed)
- The fix handles all three cases: plain dicts, `ServiceSettings` subclasses, and generic dataclasses

cc @markbackman @aconchillo

### DIFF
--- a/changelog/3858.fixed.md
+++ b/changelog/3858.fixed.md
@@ -1,0 +1,1 @@
+- Fixed tracing utilities crashing with `AttributeError` when service settings are dataclass instances (e.g. `GoogleLLMSettings`, `InworldTTSSettings`) instead of plain dicts. The four affected call sites now use a shared `_settings_to_dict()` helper that handles both dataclass and dict settings objects.
--- a/src/pipecat/utils/tracing/service_attributes.py
+++ b/src/pipecat/utils/tracing/service_attributes.py
@@ -23,6 +23,34 @@ if is_tracing_available():
     from opentelemetry.trace import Span
 
 
+def _settings_to_dict(settings: Any) -> Dict[str, Any]:
+    """Convert a settings object to a dictionary for span attributes.
+
+    Service settings may be a dict or a dataclass (e.g. ``GoogleLLMSettings``,
+    ``InworldTTSSettings``).  Dataclass instances do not have an ``.items()``
+    method, so calling it directly raises ``AttributeError``.
+
+    This helper normalises the input: if the object exposes ``given_fields()``
+    (the ``ServiceSettings`` API) that is used; otherwise ``dataclasses.asdict``
+    is tried; plain dicts are returned as-is.
+
+    Args:
+        settings: A dict or dataclass settings object.
+
+    Returns:
+        A plain dictionary of settings fields.
+    """
+    if isinstance(settings, dict):
+        return settings
+    if hasattr(settings, "given_fields"):
+        return settings.given_fields()
+    import dataclasses
+
+    if dataclasses.is_dataclass(settings):
+        return dataclasses.asdict(settings)
+    return {}
+
+
 def _get_gen_ai_system_from_service_name(service_name: str) -> str:
     """Extract the standardized gen_ai.system value from a service class name.
 
@@ -107,7 +135,7 @@ def add_tts_span_attributes(
 
     # Add settings if provided
     if settings:
-        for key, value in settings.items():
+        for key, value in _settings_to_dict(settings).items():
             if isinstance(value, (str, int, float, bool)):
                 span.set_attribute(f"settings.{key}", value)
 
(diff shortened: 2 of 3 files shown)


markbackman:
Nice find! 🕵️ 

This is happening on `main` only, right? These changes have not been released (so thanks for the early catch!).

kollaikal-rupesh:
Yep, main only. The service-settings refactor changed settings from dicts to dataclasses, which broke the .items() calls in tracing.

markbackman:
Rather than introduce a conversion function, we should just use the new ServiceSettings API. I made a PR to do this:
https://github.com/pipecat-ai/pipecat/pull/3879

Closing this PR in favor of the one I just created. Thanks for working on this though 🙇

PR #3870, opened 2026-02-27
Add ServiceSwitcherStrategyFailover for automatic failover on service errors

### LINKED ISSUE #3861: New automatic strategy for `ServiceSwitcher` that works on service failures
### Problem Statement

Currently, there's only one service switcher strategy, which is the [`ServiceSwitcherStrategyManual`](https://docs.pipecat.ai/server/utilities/service-switchers/service-switcher#serviceswitcherstrategymanual). As the name implies, it only works if service is switched explicitly. I believe this could be improved by adding a new strategy that _automatically_ switches to another service when the primary one is down, keeps track of the services' state in the background and switches to the main one once it is healthy again.

### Proposed Solution

I think something similar to what Livekit does with their [`FallbackAdapter`](https://docs.livekit.io/reference/other/events/#fallbackadapter) could be done for pipecat as well.

> When in use, FallbackAdapter handles the following: 
> - Automatically resubmits the failed request to backup providers when the primary provider fails.
> - Marks the failed provider as unhealthy and stops sending requests to it.
> - Continues to use the backup providers until the primary provider recovers.
> - Periodically checks the primary provider's status in the background.

### Alternative Solutions

_No response_

### Additional Context

_No response_

- [x] No, I'm just suggesting

kollaikal-rupesh:
## Summary

- Added `ServiceSwitcherStrategyFailover` that automatically switches to the next service when the active service reports a non-fatal error
- Moved manual switching (`ManuallySwitchServiceFrame`) and `handle_error()` into the base `ServiceSwitcherStrategy` class, making both available to all strategies
- `strategy_type` now defaults to `ServiceSwitcherStrategy`, so `ServiceSwitcher(services=[...])` works without specifying a strategy for the common manual-switching case
- Deprecated `ServiceSwitcherStrategyManual` — it is now an empty subclass of the base strategy
- `ServiceSwitcher.push_frame()` delegates non-fatal `ErrorFrame` instances to the strategy via `handle_error()`, enabling failover strategies
- Updated `LLMSwitcher` with the same default `strategy_type`
- Updated example `48-service-switcher.py` to use the simplified API

## Testing

- `uv run pytest tests/test_service_switcher.py` — 20 tests covering base strategy, deprecated manual strategy, default strategy in ServiceSwitcher, failover, and metadata handling

## Fixes

- Fixes #3861

### DIFF
--- a/changelog/3861.changed.md
+++ b/changelog/3861.changed.md
@@ -1,1 +1,1 @@
-- `ServiceSwitcherStrategy` base class now handles `ManuallySwitchServiceFrame` and `handle_error()` directly, making manual switching and error hooks available to all strategies.
+- `ServiceSwitcherStrategy` base class now provides a `handle_error()` hook for subclasses to implement error-based switching. `ServiceSwitcher` defaults to `ServiceSwitcherStrategyManual` and `strategy_type` is now optional.
--- a/changelog/3861.changed.md
+++ b/changelog/3861.changed.md
@@ -1,1 +1,1 @@
-- Added `handle_error()` hook to `ServiceSwitcherStrategy` base class. `ServiceSwitcher` now calls this method when a non-fatal `ErrorFrame` is pushed upstream, allowing strategies to implement automatic failover.
+- `ServiceSwitcherStrategy` base class now handles `ManuallySwitchServiceFrame` and `handle_error()` directly, making manual switching and error hooks available to all strategies.
--- a/changelog/3861.changed.md
+++ b/changelog/3861.changed.md
@@ -1,0 +1,1 @@
+- Added `handle_error()` hook to `ServiceSwitcherStrategy` base class. `ServiceSwitcher` now calls this method when a non-fatal `ErrorFrame` is pushed upstream, allowing strategies to implement automatic failover.
--- a/src/pipecat/pipeline/llm_switcher.py
+++ b/src/pipecat/pipeline/llm_switcher.py
@@ -9,7 +9,11 @@
 from typing import Any, List, Optional, Type
 
 from pipecat.adapters.schemas.direct_function import DirectFunction
-from pipecat.pipeline.service_switcher import ServiceSwitcher, ServiceSwitcherStrategy, StrategyType
+from pipecat.pipeline.service_switcher import (
+    ServiceSwitcher,
+    ServiceSwitcherStrategyManual,
+    StrategyType,
+)
 from pipecat.processors.aggregators.llm_context import LLMContext
 from pipecat.services.llm_service import LLMService
 
@@ -23,14 +27,16 @@ class LLMSwitcher(ServiceSwitcher[StrategyType]):
     """
 
     def __init__(
-        self, llms: List[LLMService], strategy_type: Type[StrategyType] = ServiceSwitcherStrategy
+        self,
+        llms: List[LLMService],
+        strategy_type: Type[StrategyType] = ServiceSwitcherStrategyManual,
     ):
         """Initialize the service switcher with a list of LLMs and a switching strategy.
 
         Args:
             llms: List of LLM services to switch between.
             strategy_type: The strategy class to use for switching between LLMs.
-                Defaults to ``ServiceSwitcherStrategy`` (the base strategy; manual switching only).
+                Defaults to ``ServiceSwitcherStrategyManual``.
         """
         super().__init__(llms, strategy_type)
 
--- a/src/pipecat/pipeline/llm_switcher.py
+++ b/src/pipecat/pipeline/llm_switcher.py
@@ -30,7 +30,7 @@ class LLMSwitcher(ServiceSwitcher[StrategyType]):
         Args:
             llms: List of LLM services to switch between.
             strategy_type: The strategy class to use for switching between LLMs.
-                Defaults to ``ServiceSwitcherStrategy`` (manual switching).
+                Defaults to ``ServiceSwitcherStrategy`` (the base strategy; manual switching only).
         """
         super().__init__(llms, strategy_type)
 
--- a/src/pipecat/pipeline/llm_switcher.py
+++ b/src/pipecat/pipeline/llm_switcher.py
@@ -9,7 +9,7 @@
 from typing import Any, List, Optional, Type
 
 from pipecat.adapters.schemas.direct_function import DirectFunction
-from pipecat.pipeline.service_switcher import ServiceSwitcher, StrategyType
+from pipecat.pipeline.service_switcher import ServiceSwitcher, ServiceSwitcherStrategy, StrategyType
 from pipecat.processors.aggregators.llm_context import LLMContext
 from pipecat.services.llm_service import LLMService
 
@@ -19,18 +19,18 @@ class LLMSwitcher(ServiceSwitcher[StrategyType]):
 
     Example::
 
-        llm_switcher = LLMSwitcher(
-            llms=[openai_llm, anthropic_llm],
-            strategy_type=ServiceSwitcherStrategyManual
-        )
+        llm_switcher = LLMSwitcher(llms=[openai_llm, anthropic_llm])
     """
 
-    def __init__(self, llms: List[LLMService], strategy_type: Type[StrategyType]):
+    def __init__(
+        self, llms: List[LLMService], strategy_type: Type[StrategyType] = ServiceSwitcherStrategy
+    ):
         """Initialize the service switcher with a list of LLMs and a switching strategy.
 
         Args:
             llms: List of LLM services to switch between.
             strategy_type: The strategy class to use for switching between LLMs.
+                Defaults to ``ServiceSwitcherStrategy`` (manual switching).
         """
         super().__init__(llms, strategy_type)
 
--- a/src/pipecat/pipeline/service_switcher.py
+++ b/src/pipecat/pipeline/service_switcher.py
@@ -6,9 +6,8 @@
 
 """Service switcher for switching between different services at runtime, with different switching strategies."""
 
-import asyncio
 from abc import abstractmethod
-from typing import Any, Dict, Generic, List, Optional, Set, Type, TypeVar
+from typing import Any, Generic, List, Optional, Type, TypeVar
 
 from loguru import logger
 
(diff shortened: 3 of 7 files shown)


markbackman:
Thanks for the PR and the thorough test coverage!
                                                                                                                                       
I think the basic failover (i.e. switch to next service on error) is great and belongs in the framework. The part I'd like to simplify is the recovery policy (timers, `prefer_primary`, health tracking). That tends to be application-specific: some apps want health checks, some want exponential backoff, some want manual recovery, etc.

Could we slim this down to just the switch-on-error behavior and let apps handle recovery via the `on_service_switched` event? Something like:

```
@switcher.strategy.event_handler("on_service_switched")
async def on_switched(strategy, service):
    # App decides when/how to recover the failed service
    ...
```

That gives people automatic failover out of the box while keeping recovery policy in application code. WDYT?

kollaikal-rupesh:
Good call. Stripped out the recovery policy (timers, prefer_primary, health tracking) and kept just the switch-on-error + on_service_switched event. Apps can handle recovery however they want. Tests updated too. @markbackman

markbackman:
@rupesh-svg this is great! I made one architectural change, which is to pull the ServiceSwitcherStrategyManual functionality into the base ServiceSwitcherStrategy class. This eliminates duplication and means that manual switching is always available in addition to whatever custom strategy is used. I think this is a better ergonomic choice, which we didn't think of before.

I thought about strategies being composable, but I think that gets messy with tracking which service is active.

Anyway, thanks for doing the work on this one. Hopefully it's OK that I'm collaborating on this change. I think this is a nice addition to Pipecat. Much appreciated!

kollaikal-rupesh:
That makes a lot of sense, pulling manual switching into the base class is cleaner. Absolutely fine with you collaborating on this, appreciate the improvement @markbackman!

kollaikal-rupesh:
Hey @markbackman, anything else needed on this one or is it good to merge?

markbackman:
> Hey @markbackman, anything else needed on this one or is it good to merge?

I've requested a review from my team. Just waiting for their feedback.

src/pipecat/pipeline/service_switcher.py
```diff
@@ -137,6 +125,80 @@ async def _set_active_if_available(self, service: FrameProcessor) -> Optional[Fr
         return None
 
 
+class ServiceSwitcherStrategyManual(ServiceSwitcherStrategy):
+    """A strategy for switching between services manually.
+
+    .. deprecated:: 0.0.104
+        Manual switching is now the default behavior of ``ServiceSwitcher``.
+        Pass ``services`` without a ``strategy_type`` instead.
```
kompfner: I wouldn't mention other init args (i.e. `services`) here, lest this comment get out of date. Maybe just:

```
Omit `strategy_type` when initializing.
```

src/pipecat/pipeline/llm_switcher.py
```diff
@@ -19,18 +19,18 @@ class LLMSwitcher(ServiceSwitcher[StrategyType]):
...
-    def __init__(self, llms: List[LLMService], strategy_type: Type[StrategyType]):
+    def __init__(
+        self, llms: List[LLMService], strategy_type: Type[StrategyType] = ServiceSwitcherStrategy
+    ):
         """Initialize the service switcher with a list of LLMs and a switching strategy.
 
         Args:
             llms: List of LLM services to switch between.
             strategy_type: The strategy class to use for switching between LLMs.
+                Defaults to ``ServiceSwitcherStrategy`` (manual switching).
```
kompfner:
```suggestion
                Defaults to ``ServiceSwitcherStrategy`` (the base strategy; manual switching only).
```

src/pipecat/pipeline/service_switcher.py
```diff
@@ -150,18 +212,20 @@ class ServiceSwitcher(ParallelPipeline, Generic[StrategyType]):
...
+        self,
+        services: List[FrameProcessor],
+        strategy_type: Type[StrategyType] = ServiceSwitcherStrategy,
+    ):
         """Initialize the service switcher with a list of services and a switching strategy.
 
         Args:
             services: List of frame processors to switch between.
             strategy_type: The strategy class to use for switching between services.
+                Defaults to ``ServiceSwitcherStrategy`` (manual switching).
```
kompfner: See minor suggestion in llm_switcher.py

src/pipecat/pipeline/service_switcher.py
kollaikal-rupesh: Updated, thanks!

src/pipecat/pipeline/llm_switcher.py
kollaikal-rupesh: Applied, thanks!

src/pipecat/pipeline/service_switcher.py
kollaikal-rupesh: Updated, all three addressed now!

changelog/3861.changed.md
```diff
@@ -0,0 +1 @@
+- `ServiceSwitcherStrategy` base class now handles `ManuallySwitchServiceFrame` and `handle_error()` directly, making manual switching and error hooks available to all strategies.
```
kompfner: This comes at a cost - that means that any future strategies' bookkeeping will have to contend with the possibility that, at any point, the user might choose to manually swap out the active service. Maybe it's worth it, though...I can't really imagine a strategy where you wouldn't want at least the _option_ to switch manually.

kollaikal-rupesh: Yeah, that was the reasoning, hard to think of a strategy where you'd want to block manual override. And if one comes up, a subclass can always override handle_frame to restrict it.

src/pipecat/pipeline/service_switcher.py
```diff
@@ -239,6 +303,12 @@ async def push_frame(self, frame: Frame, direction: FrameDirection = FrameDirect
             if frame.service_name != self.strategy.active_service.name:
                 return
 
+        # Let the strategy react to non-fatal errors from the active service.
```
kompfner, line 315: Why do this logic here in `push_frame` rather than in `process_frame`, where we do the other strategy invocation (i.e. `handle_frame`)?

kompfner: Ah wait it's because the error is coming *from* the service, not going *to* the service. Answered my own question.

src/pipecat/pipeline/service_switcher.py
```diff
@@ -239,6 +303,12 @@ async def push_frame(self, frame: Frame, direction: FrameDirection = FrameDirect
...
+        if isinstance(frame, ErrorFrame) and not frame.fatal:
+            service = await self.strategy.handle_error(frame)
+            if service:
```
kompfner: In the `ServiceSwitcherFrame`-handling logic in `process_frame`, we also see the same pattern:
```
if service:
    await service.queue_frame(ServiceSwitcherRequestMetadataFrame(service=service))
```
If we will always need to request metadata from the service we just switched to, perhaps this logic should live inside `_set_active_if_available` in the base strategy class?

kompfner: (I recognize that in `process_frame` today there is a teeny bit of logic between `service = await self.strategy.handle_frame(frame, direction)` and the metadata-requesting logic, but looking at it I don't _think_ it necessarily has to come before, though I'm not 100% sure)

changelog/3861.changed.md
kompfner: Part of me is still wondering if we shouldn't instead have it so that strategies that need to inherit manual-switching behavior inherit from `ServiceSwitcherStrategyManual`...

It would also solve an (admittedly minor) quirk that's nagging at me. It feels odd for `ServiceSwitcher` to default to a base `ServiceSwitcherStrategy` because it's a base class, and sounds/feels like a non-strategy:

```python
# Before
switcher.strategy # ServiceSwitcherStrategyManual - cool, got it, makes sense

# Now
switcher.strategy # ServiceSwitcherStrategy - huh? but what is it?
```

We could solve both problems
1. The (again, admittedly minor) quirk above
2. The (probably minor) issue of every strategy having to contend with manual overrides

by:
1. Putting the manual switching logic back in `ServiceSwitcherStrategyManual`
2. Having the failover strategy inherit from `ServiceSwitcherStrategyManual`
3. Having `ServiceSwitcher` default to `ServiceSwitcherStrategyManual` (I like that it has a sensible default now, and agree that "manual" feels like the sensible default)

cc @markbackman - thoughts?

kompfner: (I think it still makes sense to have `_set_active_if_available` be shared in the base class, too; per another of my comments, we could have more shared "set active" logic coming)

yuki901:
How is it going? @kollaikal-rupesh

src/pipecat/pipeline/service_switcher.py
kollaikal-rupesh: Thanks for catching this! Makes sense to consolidate the metadata request into `_set_active_if_available`. Working on it.

changelog/3861.changed.md
kollaikal-rupesh: This makes sense to me having `ServiceSwitcherStrategyManual` as the explicit default reads much clearer than using the base class directly. Happy to make the change once we align. @markbackman

kollaikal-rupesh:
Thanks for checking in! Working on the remaining review feedback, addressing the metadata dedup suggestion now. Waiting on @markbackman for input on the architecture question about where manual switching logic should live (base class vs `ServiceSwitcherStrategyManual`).

markbackman:
I talked with @kompfner about this. @kollaikal-rupesh if you don't mind, I'll finish this up.

kollaikal-rupesh:
Sounds good, all yours @markbackman!

src/pipecat/pipeline/service_switcher.py
```diff
@@ -118,23 +153,54 @@ async def handle_frame(
 
         return None
 
-    async def _set_active_if_available(self, service: FrameProcessor) -> Optional[FrameProcessor]:
-        """Set the active service to the given one, if it is in the list of available services.
 
-        If it's not in the list, the request is ignored, as it may have been
-        intended for another ServiceSwitcher in the pipeline.
+class ServiceSwitcherStrategyFailover(ServiceSwitcherStrategyManual):
```
aconchillo, line 158: Is this because we still want to support switching manually?

aconchillo (approved):

aconchillo:
LGTM!

src/pipecat/pipeline/service_switcher.py
markbackman: Yes, that's right. That's what @kompfner and I were discussing. how to continue supporting the manual strategy along with a custom strategy.

PR #3871, opened 2026-02-27
Fix PipelineTask double-inserting RTVIProcessor with custom RTVIObserver

### LINKED ISSUE #3867: PipelineTask double-inserts RTVIProcessor when custom RTVIObserver subclass is provided in observers
### pipecat version

0.0.103

### Python version

3.12.10

### Operating System

macOS 15.7.2

### Issue description

When using a custom RTVIObserver subclass, PipelineTask.__init__ correctly detects both the RTVIProcessor in the pipeline and the observer subclass, logging "RTVIProcessor and RTVIObserver found, skipping default ones." However, it then unconditionally inserts self._rtvi before the user pipeline (tasks.py, row 391):

```python
processors = [self._rtvi, pipeline] if self._rtvi else [pipeline]
```

This causes RTVIProcessor to appear twice in the frame chain — once inside the user pipeline and once wrapping it.

There is also no way to provide a custom RTVIObserver subclass via constructor parameters. rtvi_observer_params only accepts params for the default observer, not a custom instance.

### Reproduction steps

Test script:
```python
from pipecat.pipeline.task import PipelineTask, PipelineParams
from pipecat.processors.frameworks.rtvi import (
    RTVIConfig, RTVIObserver, RTVIObserverParams, RTVIProcessor,
)
from pipecat.processors.frame_processor import FrameProcessor


class DummyProcessor(FrameProcessor):
    pass


class CustomRTVIObserver(RTVIObserver):
    """Custom subclass, e.g. for filtering emoji from transcriptions."""
    pass


def count_rtvi(processor):
    count = 1 if isinstance(processor, RTVIProcessor) else 0
    for p in getattr(processor, "processors", []):
        count += count_rtvi(p)
    return count


rtvi = RTVIProcessor(config=RTVIConfig(config=[]))
pipeline = Pipeline([rtvi, DummyProcessor()])
observer = CustomRTVIObserver(rtvi, params=RTVIObserverParams(metrics_enabled=False))

task = PipelineTask(
    pipeline,
    params=PipelineParams(),
    enable_rtvi=False,
    observers=[observer],
)

print(f"RTVIProcessor count: {count_rtvi(task._pipeline)}")
# Output: 2 (expected: 1)
```

### Expected behavior

When PipelineTask detects that the user has already placed both RTVIProcessor in the pipeline and an RTVIObserver (sub)class in observers, it should skip re-inserting the processor. Additionally, a rtvi_observer parameter (accepting a custom RTVIObserver instance) would allow clean custom observer usage without manually placing both processor and observer.

### Actual behavior

RTVIProcessor is inserted twice into the frame chain.

### Logs

```shell

```

kollaikal-rupesh:
## Summary

- Fixed `PipelineTask` prepending `RTVIProcessor` to the pipeline even when the user already placed it inside their own pipeline
- Added a `_rtvi_external` flag to distinguish between an externally provided RTVIProcessor and an internally created one

## Root Cause

When the user provides both an `RTVIProcessor` inside their pipeline and a custom `RTVIObserver` subclass in observers, `PipelineTask.__init__` correctly detects both and logs "RTVIProcessor and RTVIObserver found, skipping default ones." It then sets `self._rtvi = external_rtvi`.

However, the pipeline construction at line 391 unconditionally prepends `self._rtvi`:

```python
processors = [self._rtvi, pipeline] if self._rtvi else [pipeline]
```

Since `self._rtvi` is set (it references the processor already inside the user pipeline), the processor ends up in the frame chain twice.

## Approach

Track whether the `RTVIProcessor` was found externally (already in the user pipeline) with a `_rtvi_external` flag. Only prepend it when it was created internally by `PipelineTask`.

```python
if self._rtvi and not self._rtvi_external:
    processors = [self._rtvi, pipeline]
else:
    processors = [pipeline]
```

## Fixes

Fixes #3867

## Testing

- All 575 existing tests pass

cc @markbackman @aconchillo

### DIFF
--- a/changelog/3867.fixed.md
+++ b/changelog/3867.fixed.md
@@ -1,0 +1,1 @@
+- Fixed `PipelineTask` double-inserting `RTVIProcessor` into the frame chain when the user provides both an `RTVIProcessor` in the pipeline and a custom `RTVIObserver` subclass in observers.
--- a/src/pipecat/pipeline/task.py
+++ b/src/pipecat/pipeline/task.py
@@ -330,7 +330,7 @@ class PipelineTask(BasePipelineTask):
 
         # RTVI support
         self._rtvi = None
-        self._rtvi_external = False
+        prepend_rtvi = False
         external_rtvi = self._find_processor(pipeline, RTVIProcessor)
         external_observer_found = any(isinstance(o, RTVIObserver) for o in observers)
 
@@ -350,10 +350,10 @@ class PipelineTask(BasePipelineTask):
                 "They are both added by default, no need to add them yourself."
             )
             self._rtvi = external_rtvi
-            self._rtvi_external = True
         elif enable_rtvi:
             self._rtvi = rtvi_processor or RTVIProcessor()
             observers.append(self._rtvi.create_rtvi_observer(params=rtvi_observer_params))
+            prepend_rtvi = True
 
         if self._rtvi:
             # Automatically call RTVIProcessor.set_bot_ready()
@@ -393,10 +393,7 @@ class PipelineTask(BasePipelineTask):
         # Only prepend the RTVIProcessor if we created it ourselves. When the
         # user already placed it inside their pipeline we must not insert it
         # again or it will appear twice in the frame chain.
-        if self._rtvi and not self._rtvi_external:
-            processors = [self._rtvi, pipeline]
-        else:
-            processors = [pipeline]
+        processors = [self._rtvi, pipeline] if prepend_rtvi else [pipeline]
         self._pipeline = Pipeline(processors, source=source, sink=sink)
 
         # The task observer acts as a proxy to the provided observers. This way,
--- a/src/pipecat/pipeline/task.py
+++ b/src/pipecat/pipeline/task.py
@@ -330,6 +330,7 @@ class PipelineTask(BasePipelineTask):
 
         # RTVI support
         self._rtvi = None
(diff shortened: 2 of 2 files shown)


markbackman:
Thanks for picking this one up!

One suggestion: since `_rtvi_external` is only used during `__init__` to decide whether to prepend, we can avoid the instance variable and use a local instead. This keeps the construction-time decision out of instance state and communicates the intent ("should we prepend?") more directly:

```python
# At the top of the RTVI section (Line 331):
prepend_rtvi = False

# In the `elif enable_rtvi:` branch (the only case where we create it ourselves):
elif enable_rtvi:
    self._rtvi = rtvi_processor or RTVIProcessor()
    observers.append(self._rtvi.create_rtvi_observer(params=rtvi_observer_params))
    prepend_rtvi = True

# Then at pipeline construction:
processors = [self._rtvi, pipeline] if prepend_rtvi else [pipeline]
   ```

This way `self._rtvi` still gets set for the external case (so the `rtvi` property and `on_client_ready` handler work), but we only prepend when we created the processor ourselves. No new instance variable needed.

Could you update the PR with this approach?

kollaikal-rupesh:
Updated, using a local prepend_rtvi instead of the instance variable. Thanks for the suggestion @markbackman.

markbackman (approved):
LGTM! Thanks for the fix 🙌

PR #3873, opened 2026-02-27
[RimeTTS] Add `speedAlpha` parameter support to the `arcana` model 

MaCaki:
## Summary

The `speedAlpha` param can now be sent to the `arcana` model. This change allows a User set value for `speedAlpha` to be passed as a url param when opening up a websockets conn with Rime's APIs for any model requested.


markbackman (approved):
LGTM!

PR #3874, opened 2026-02-28
Changelog for PR 3873, docstrings change

markbackman:

Follow up PR to https://github.com/pipecat-ai/pipecat/pull/3873

`speed_alpha` is now available for `arcana` and `mistv2` models.

### DIFF
--- a/src/pipecat/services/rime/tts.py
+++ b/src/pipecat/services/rime/tts.py
@@ -147,10 +147,10 @@ class RimeTTSService(AudioContextTTSService):
         Parameters:
             language: Language for synthesis. Defaults to English.
             segment: Text segmentation mode ("immediate", "bySentence", "never").
+            speed_alpha: Speech speed multiplier.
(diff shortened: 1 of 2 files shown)


src/pipecat/services/rime/tts.py
```diff
@@ -147,10 +147,10 @@ class InputParams(BaseModel):
         Parameters:
             language: Language for synthesis. Defaults to English.
             segment: Text segmentation mode ("immediate", "bySentence", "never").
+            speed_alpha: Speech speed multiplier (mistv2 only).
```
filipi87: The PR description mentions that it is now available in both models, arcana and mistv2. But here the docstring is only mentioning "mistv2 only".

filipi87 (approved):
LGTM! Just one comment regarding the docstring, which I believe needs to be fixed.

src/pipecat/services/rime/tts.py
markbackman: Good catch 🕵️

PR #3877, opened 2026-02-28
Add `settings` as canonical init arg for all AIService descendants, d…

kompfner:
…eprecate redundant `model`/`voice`/`params` args

ServiceSettings types were introduced for runtime updates via ServiceUpdateSettingsFrame, but there was tension between init-time and runtime APIs: overlapping-but-different InputParams vs ServiceSettings classes, and runtime-updatable fields like `model` and `voice` scattered as direct init args rather than living in a settings object. This unifies them so developers use the same settings type at both init and runtime, improving ergonomics and consistency.

Every concrete AIService subclass (LLM, TTS, STT, ImageGen, Vision, Video) now accepts a `settings` parameter for runtime-updatable config. Old init args (`model`, `voice_id`, `params`/`InputParams`) still work but emit DeprecationWarnings pointing to the new API. When both are provided, `settings` takes precedence. Leaf classes emit warnings; base classes do not, avoiding double warnings in inheritance chains.

---

# IMPORTANT

While this PR is already open, it has **not** been thoroughly vetted yet, even by the author. It still needs a once-over, followed by service-by-service careful validation (maybe done in parallel by multiple members of the team)

For context to the reader about the bigger picture, here is the Claude plan—carefully crafted and iterated on—that resulted in these changes: 
[service-init-settings-alignment.md](https://github.com/user-attachments/files/25618290/service-init-settings-alignment.md).

### DIFF
--- a/examples/foundational/20e-persistent-context-aws-nova-sonic.py
+++ b/examples/foundational/20e-persistent-context-aws-nova-sonic.py
@@ -24,7 +24,7 @@ from pipecat.processors.aggregators.llm_context import LLMContext
 from pipecat.processors.aggregators.llm_response_universal import LLMContextAggregatorPair
 from pipecat.runner.types import RunnerArguments
 from pipecat.runner.utils import create_transport
-from pipecat.services.aws.nova_sonic.llm import AWSNovaSonicLLMService
+from pipecat.services.aws.nova_sonic.llm import AWSNovaSonicLLMService, AWSNovaSonicLLMSettings
 from pipecat.services.llm_service import FunctionCallParams
 from pipecat.transports.base_transport import BaseTransport, TransportParams
 from pipecat.transports.daily.transport import DailyParams
@@ -222,9 +222,10 @@ async def run_bot(transport: BaseTransport, runner_args: RunnerArguments):
         secret_access_key=os.getenv("AWS_SECRET_ACCESS_KEY"),
         access_key_id=os.getenv("AWS_ACCESS_KEY_ID"),
         region=os.getenv("AWS_REGION"),  # as of 2025-05-06, us-east-1 is the only supported region
-        voice_id="tiffany",  # matthew, tiffany, amy
-        # you could choose to pass instruction here rather than via context
-        # system_instruction=system_instruction,
+        settings=AWSNovaSonicLLMSettings(
+            voice="tiffany",  # matthew, tiffany, amy
+            system_instruction=system_instruction,
+        ),
         # you could choose to pass tools here rather than via context
         # tools=tools
     )
@@ -236,7 +237,6 @@ async def run_bot(transport: BaseTransport, runner_args: RunnerArguments):
 
     context = LLMContext(
         messages=[
-            {"role": "system", "content": f"{system_instruction}"},
             {"role": "user", "content": "Hello!"},
         ],
         tools=tools,
--- a/examples/foundational/40-aws-nova-sonic.py
+++ b/examples/foundational/40-aws-nova-sonic.py
@@ -28,7 +28,7 @@ from pipecat.processors.aggregators.llm_response_universal import (
 )
 from pipecat.runner.types import RunnerArguments
 from pipecat.runner.utils import create_transport
-from pipecat.services.aws.nova_sonic.llm import AWSNovaSonicLLMService
+from pipecat.services.aws.nova_sonic.llm import AWSNovaSonicLLMService, AWSNovaSonicLLMSettings
 from pipecat.services.llm_service import FunctionCallParams
 from pipecat.transports.base_transport import BaseTransport, TransportParams
 from pipecat.transports.daily.transport import DailyParams
@@ -130,9 +130,10 @@ async def run_bot(transport: BaseTransport, runner_args: RunnerArguments):
         #   - ap-northeast-1
         region=os.getenv("AWS_REGION"),
         session_token=os.getenv("AWS_SESSION_TOKEN"),
-        voice_id="tiffany",
-        # you could choose to pass instruction here rather than via context
-        # system_instruction=system_instruction
+        settings=AWSNovaSonicLLMSettings(
+            voice="tiffany",
+            system_instruction=system_instruction,
+        ),
         # you could choose to pass tools here rather than via context
         # tools=tools
     )
@@ -147,7 +148,6 @@ async def run_bot(transport: BaseTransport, runner_args: RunnerArguments):
     # Set up context and context management.
     context = LLMContext(
         messages=[
-            {"role": "system", "content": f"{system_instruction}"},
             {
                 "role": "user",
                 "content": "Tell me a fun fact!",


src/pipecat/services/anthropic/llm.py
```diff
@@ -225,41 +230,65 @@ def __init__(
...
-                presence_penalty=None,
-                seed=None,
-                filter_incomplete_user_turns=False,
-                user_turn_completion_config=None,
-                thinking=params.thinking,
-                extra=params.extra if isinstance(params.extra, dict) else {},
-            ),
-            **kwargs,
+        if model is not None:
+            _warn_deprecated_param("model", "AnthropicLLMSettings", "model")
```
kompfner: I don't like that we have to pass a string for the class name here...

src/pipecat/services/anthropic/llm.py
```diff
@@ -225,41 +230,65 @@ def __init__(
 
         Args:
             api_key: Anthropic API key for authentication.
-            model: Model name to use. Defaults to "claude-sonnet-4-6".
+            model: Model name to use.
+
+                .. deprecated::
+                    Use ``settings=AnthropicLLMSettings(model=...)`` instead.
```
kompfner, line 235: Where are all the version numbers? These docstring deprecation notices should all say...what? 1.0.0? Or should we target 0.0.104?

markbackman: 0.0.104 for now, but we can update to 105 or 1.0.0 depending on the timing.

kompfner:
Also needed: community guidelines update. It should express something to the effect of: "if an init-time setting is runtime updateable _or ought to be (implementation pending)_ then it should go into the settings object, otherwise it should go as a direct init-time arg; see Dynamic Settings Updates section for more info"

kompfner:
@markbackman, based on our convo here's a slight tweak to the template/example from the Claude plan:

```python
def __init__(
    self,
    *,
    api_key: str,                                  # Fixed: stays as direct arg
    voice_id: Optional[str] = None,                # DEPRECATED → settings.voice
    model: Optional[str] = None,                   # DEPRECATED → settings.model
    params: Optional[InputParams] = None,          # DEPRECATED → settings
    settings: Optional[MySettings] = None,         # NEW: the canonical API
    **kwargs,
):
    # 1. initialize default_settings
    default_settings = MySettings(
        model="actual default", # moved from init default
        voice="actual default", # moved from init default
        speed="actual default", # moved from InputParams() default
        ...
    )

    # 2. apply direct init arg overrides, warning if any are specified
    if voice_id is not None:
        _warn_deprecated_param("voice_id", "MySettings", "voice")
        # apply value
        default_settings.voice = voice_id
    if model is not None:
        _warn_deprecated_param("model", "MySettings", "model")
        # apply value
        default_settings.model = model

    # 3. if settings arg is not provided, apply input params overrides
    # (i.e. ignore input params wholesale if settings is provided, no benefit of "mixing")
    if params is not None:
        _warn_deprecated_param("params", "MySettings")
        # only apply params values if settings is not specified - settings takes precedence *wholesale*
        if not settings:
            # apply value
            default_settings.model = params.model

    # 4. apply settings overrides (the new way of doing things!)
    if settings is not None:
        default_settings.apply_update(settings)

    # 5. Call super with merged settings
    super().__init__(settings=default_settings, **kwargs)
    self._api_key = api_key
```

Important things to point out to Claude:
- Slight change to the setting priority/fallback logic (settings takes precedence over params wholesale, not field-by-field)
- Clarity is more important than conciseness here - the priority fallback logic is clearly laid out step by step as separate blocks/chunks in the init logic
- A goal is to be able to someday easily delete params and individual direct init args with confidence; with the new template above you'd just delete, say, all code under the `#2` comment or `#3` comment

src/pipecat/services/anthropic/llm.py
```diff
@@ -170,6 +170,10 @@ class AnthropicLLMService(LLMService):
     class InputParams(BaseModel):
         """Input parameters for Anthropic model inference.
 
+        .. deprecated::
```
kompfner, line 173: Looks like we're still missing some version numbers...

COMMUNITY_INTEGRATIONS.md
kompfner, line 1: Nice update to this doc!

COMMUNITY_INTEGRATIONS.md
```diff
@@ -231,49 +231,102 @@ def can_generate_metrics(self) -> bool:
...
+# Overrides just the voice
+svc = MyTTSService(
+    api_key="sk-xxx",
+    settings=MyTTSSettings(voice="custom-voice"),
+)
 ```
 
+#### Reacting to runtime changes
+
+STT, LLM, and TTS services support runtime configuration changes via `*UpdateSettingsFrame`s (e.g. `STTUpdateSettingsFrame`, `TTSUpdateSettingsFrame`, `LLMUpdateSettingsFrame`).
```
kompfner, line 323: As I was reading through this just now, it struck me that earlier in this doc we mention image-generation services, and they're not mentioned here. This may be a bit confusing to some readers.

COMMUNITY_INTEGRATIONS.md
```diff
@@ -231,49 +231,102 @@ def can_generate_metrics(self) -> bool:
     return True
 ```
 
-### Dynamic Settings Updates
+### Service Settings
 
-STT, LLM, and TTS services support runtime configuration changes via `*UpdateSettingsFrame`s (e.g. `STTUpdateSettingsFrame`, `TTSUpdateSettingsFrame`, `LLMUpdateSettingsFrame`).
+Every STT, LLM, TTS, and image-generation service exposes a **Settings dataclass** that serves two roles:
```
kompfner, line 236: Maybe we should say something more along the lines of "every `AIService` (e.g. STT, LLM, TTS, image-generation service)" to leave things open for other service types?

COMMUNITY_INTEGRATIONS.md
```diff
@@ -231,49 +231,102 @@ def can_generate_metrics(self) -> bool:
...
+
+1. **Store mode** — the service's `self._settings` holds the current value of every runtime-updatable field.
+2. **Delta mode** — an update frame carries only the fields that changed; unset fields remain `NOT_GIVEN`.
+
+#### Defining your Settings class
 
-Each service declares a settings dataclass that extends the appropriate base (`STTSettings`, `TTSSettings`, `LLMSettings`). Fields default to `NOT_GIVEN` so that update objects can represent sparse deltas:
+Extend `STTSettings`, `TTSSettings`, `LLMSettings`, or `ImageGenSettings`. The base classes already provide common fields (e.g. `model`, `voice`, `language`). You only need to add **service-specific knobs that should be runtime-updatable**:
```
kompfner, line 243: Maybe "Extend `STTSettings`, `TTSSettings`, `LLMSettings` or `ImageGenSettings` (or, if your service directly subclasses `AIService`, `ServiceSettings`)"?

COMMUNITY_INTEGRATIONS.md
```diff
@@ -231,49 +231,102 @@ def can_generate_metrics(self) -> bool:
...
 ```
 
-### Dynamic Settings Updates
+### Service Settings
 
-STT, LLM, and TTS services support runtime configuration changes via `*UpdateSettingsFrame`s (e.g. `STTUpdateSettingsFrame`, `TTSUpdateSettingsFrame`, `LLMUpdateSettingsFrame`).
+Every STT, LLM, TTS, and image-generation service exposes a **Settings dataclass** that serves two roles:
+
+1. **Store mode** — the service's `self._settings` holds the current value of every runtime-updatable field.
+2. **Delta mode** — an update frame carries only the fields that changed; unset fields remain `NOT_GIVEN`.
```
kompfner, line 239:
```suggestion
2. **Delta mode** — an update frame (e.g. `TTSSettingsUpdateFrame`) specifies only the fields that should change; unspecified fields remain `NOT_GIVEN`.
```

PR #3879, opened 2026-02-28
Fix tracing to use ServiceSettings API instead of dict access

### LINKED ISSUE #3858: Tracing crashes on dataclass settings: .items() called on non-dict objects

markbackman:
## Summary

- Fixed tracing code to use `ServiceSettings` dataclass API instead of dict-style access (`.items()`, `in`, subscript), which broke after the ServiceSettings refactor (PR #3714)
- In `service_attributes.py`: changed `settings` parameter type to `ServiceSettings` and use `given_fields().items()` for iteration (TTS, STT, Gemini Live)
- In `service_decorators.py`: changed `getattr` defaults from `{}` to `None`, replaced `self._settings.items()` with `given_fields().items()` in LLM decorator, replaced dict containment/subscript with attribute access in Gemini Live decorator
- Removed unnecessary `extra` key skip and `NOT_GIVEN` checks since `given_fields()` already handles both

## Fixes

- Fixes #3858

### DIFF
--- a/src/pipecat/utils/tracing/service_attributes.py
+++ b/src/pipecat/utils/tracing/service_attributes.py
@@ -17,6 +17,8 @@ from typing import TYPE_CHECKING, Any, Dict, List, Optional
 if TYPE_CHECKING:
     from opentelemetry.trace import Span
 
+    from pipecat.services.settings import ServiceSettings
+
 from pipecat.utils.tracing.setup import is_tracing_available
 
 if is_tracing_available():
@@ -68,7 +70,7 @@ def add_tts_span_attributes(
     model: str,
     voice_id: str,
     text: Optional[str] = None,
-    settings: Optional[Dict[str, Any]] = None,
+    settings: Optional["ServiceSettings"] = None,
     character_count: Optional[int] = None,
     operation_name: str = "tts",
     ttfb: Optional[float] = None,
@@ -107,7 +109,7 @@ def add_tts_span_attributes(
 
     # Add settings if provided
     if settings:
-        for key, value in settings.items():
+        for key, value in settings.given_fields().items():
             if isinstance(value, (str, int, float, bool)):
                 span.set_attribute(f"settings.{key}", value)
 
@@ -126,7 +128,7 @@ def add_stt_span_attributes(
     is_final: Optional[bool] = None,
     language: Optional[str] = None,
     user_id: Optional[str] = None,
-    settings: Optional[Dict[str, Any]] = None,
+    settings: Optional["ServiceSettings"] = None,
     vad_enabled: bool = False,
     ttfb: Optional[float] = None,
     **kwargs,
@@ -171,7 +173,7 @@ def add_stt_span_attributes(
 
     # Add settings if provided
     if settings:
-        for key, value in settings.items():
+        for key, value in settings.given_fields().items():
             if isinstance(value, (str, int, float, bool)):
                 span.set_attribute(f"settings.{key}", value)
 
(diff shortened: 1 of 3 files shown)


filipi87 (approved):
LGTM!

src/pipecat/utils/tracing/service_attributes.py
```diff
@@ -107,7 +109,7 @@ def add_tts_span_attributes(
 
     # Add settings if provided
     if settings:
-        for key, value in settings.items():
+        for key, value in settings.given_fields().items():
```
kompfner, line 112: Nit: we shouldn't need `given_fields()`, as they should all be given (settings is in "storage mode" here, not in "delta mode"). But this is safer, in case a developer forgets to initialize all values in `_settings` in their service. Might be worth a comment to this effect though

src/pipecat/utils/tracing/service_decorators.py
```diff
@@ -510,15 +510,10 @@ async def wrapped_start_llm_usage_metrics(tokens):
                             # Get settings from the service
                             params = {}
                             if hasattr(self, "_settings"):
-                                for key, value in self._settings.items():
-                                    if key == "extra":
-                                        continue
-                                    # Add value directly if it's a basic type
+                                for key, value in self._settings.given_fields().items():
```
kompfner, line 513: ditto, just a comment like "we shouldn't need to get given fields (they should all be given), but juuuuust in case"

PR #3881, opened 2026-02-28
Add StartupTimingObserver

markbackman:
## Context

The goal is to characterize how long it takes for the infrastructure and application code to be ready to run. The important events to measure are:

- Startup time for processors in the pipeline
- Transport readiness (bot connection + client connection)

startup-illustration

## Summary

- Added `StartupTimingObserver` for measuring how long each processor's `start()` method takes during pipeline startup
- Tracks `StartFrame` arrive/leave deltas per processor, emits an `on_startup_timing_report` event with the full timing report
- Internal pipeline processors (`PipelineSource`, `PipelineSink`, `Pipeline`) are excluded from reports by default
- Optional `processor_types` filter to measure only specific processor types (e.g. STT and TTS services)
- Added `ClientConnectedFrame`, a new `SystemFrame` pushed downstream by all transports (Daily, LiveKit, FastAPI WebSocket, WebSocket Server, SmallWebRTC, HeyGen, Tavus) when a client connects
- Added `BotConnectedFrame`, a new `SystemFrame` pushed downstream by SFU transports (Daily, LiveKit, HeyGen, Tavus) when the bot joins the room — non-SFU transports don't emit this frame
- `StartupTimingObserver` uses both frames to measure transport timing via an `on_transport_timing_report` event containing `bot_connected_secs` (Optional, SFU only) and `client_connected_secs`
- Observer data models use Pydantic `BaseModel` with `start_time` (Unix timestamp) fields for serialization and timeline plotting
- Updated the foundational example `29-turn-tracking-observer.py` to demonstrate startup timing and transport readiness

## Testing

```bash
uv run pytest tests/test_startup_timing_observer.py -v
```

Example output (Daily):
```
Total startup: 0.359s
  DeepgramSTTService#0: 0.260s
  CartesiaTTSService#0: 0.099s
Bot connected: 0.420s
Client connected: 2.672s
```

Example output (FastAPI WebSocket):
```
Total startup: 0.477s
  DeepgramSTTService#0: 0.297s
  CartesiaTTSService#0: 0.179s
Client connected: 0.001s
```

### DIFF
--- a/examples/foundational/29-turn-tracking-observer.py
+++ b/examples/foundational/29-turn-tracking-observer.py
@@ -12,6 +12,7 @@ from loguru import logger
 
 from pipecat.audio.vad.silero import SileroVADAnalyzer
 from pipecat.frames.frames import LLMRunFrame
+from pipecat.observers.startup_timing_observer import StartupTimingObserver
 from pipecat.observers.user_bot_latency_observer import UserBotLatencyObserver
 from pipecat.pipeline.pipeline import Pipeline
 from pipecat.pipeline.runner import PipelineRunner
@@ -87,8 +88,8 @@ async def run_bot(transport: BaseTransport, runner_args: RunnerArguments):
         ]
     )
 
-    # Create latency tracking observer
     latency_observer = UserBotLatencyObserver()
+    startup_observer = StartupTimingObserver()
 
     task = PipelineTask(
         pipeline,
@@ -97,13 +98,18 @@ async def run_bot(transport: BaseTransport, runner_args: RunnerArguments):
             enable_usage_metrics=True,
         ),
         idle_timeout_secs=runner_args.pipeline_idle_timeout_secs,
-        observers=[latency_observer],
+        observers=[latency_observer, startup_observer],
     )
 
-    # Log latency measurements using the event handler
     @latency_observer.event_handler("on_latency_measured")
     async def on_latency_measured(observer, latency_seconds):
         logger.info(f"⏱️ User-to-bot latency: {latency_seconds:.3f}s")
+
+    @startup_observer.event_handler("on_startup_timing_report")
+    async def on_startup_timing_report(observer, report):
+        logger.info(f"Total startup: {report.total_duration_secs:.3f}s")
+        for timing in report.processor_timings:
+            logger.info(f"  {timing.processor_name}: {timing.duration_secs:.3f}s")
 
     turn_observer = task.turn_tracking_observer
     if turn_observer:
--- a/examples/foundational/29-turn-tracking-observer.py
+++ b/examples/foundational/29-turn-tracking-observer.py
@@ -111,6 +111,10 @@ async def run_bot(transport: BaseTransport, runner_args: RunnerArguments):
         for timing in report.processor_timings:
             logger.info(f"  {timing.processor_name}: {timing.duration_secs:.3f}s")
 
+    @startup_observer.event_handler("on_transport_readiness_measured")
+    async def on_transport_readiness_measured(observer, report):
+        logger.info(f"Transport readiness: {report.readiness_secs:.3f}s")
+
     turn_observer = task.turn_tracking_observer
     if turn_observer:
 
--- a/examples/foundational/29-turn-tracking-observer.py
+++ b/examples/foundational/29-turn-tracking-observer.py
@@ -111,9 +111,11 @@ async def run_bot(transport: BaseTransport, runner_args: RunnerArguments):
         for timing in report.processor_timings:
             logger.info(f"  {timing.processor_name}: {timing.duration_secs:.3f}s")
 
-    @startup_observer.event_handler("on_transport_readiness_measured")
-    async def on_transport_readiness_measured(observer, report):
-        logger.info(f"Transport readiness: {report.readiness_secs:.3f}s")
+    @startup_observer.event_handler("on_transport_timing_report")
+    async def on_transport_timing_report(observer, report):
+        if report.bot_connected_secs is not None:
+            logger.info(f"Bot connected: {report.bot_connected_secs:.3f}s")
+        logger.info(f"Client connected: {report.client_connected_secs:.3f}s")
 
     turn_observer = task.turn_tracking_observer
     if turn_observer:
--- a/src/pipecat/frames/frames.py
+++ b/src/pipecat/frames/frames.py
@@ -1911,6 +1911,18 @@ class StopFrame(ControlFrame, UninterruptibleFrame):
 
 
 @dataclass
+class BotConnectedFrame(SystemFrame):
+    """Frame indicating the bot has connected to the transport service.
+
+    Pushed downstream by SFU transports (Daily, LiveKit, HeyGen, Tavus)
+    when the bot successfully joins the room. Non-SFU transports do not
+    emit this frame.
+    """
+
+    pass
+
+
+@dataclass
 class ClientConnectedFrame(SystemFrame):
     """Frame indicating that a client has connected to the transport.
 
--- a/src/pipecat/frames/frames.py
+++ b/src/pipecat/frames/frames.py
@@ -1911,6 +1911,17 @@ class StopFrame(ControlFrame, UninterruptibleFrame):
 
 
 @dataclass
+class ClientConnectedFrame(SystemFrame):
+    """Frame indicating that a client has connected to the transport.
+
+    Pushed downstream by the input transport when a client (participant)
+    connects. Used by observers to measure transport readiness timing.
+    """
+
+    pass
+
+
+@dataclass
 class OutputTransportReadyFrame(ControlFrame):
     """Frame indicating that the output transport is ready.
 
--- a/src/pipecat/observers/startup_timing_observer.py
+++ b/src/pipecat/observers/startup_timing_observer.py
@@ -1,0 +1,232 @@
+#
+# Copyright (c) 2024-2026, Daily
+#
+# SPDX-License-Identifier: BSD 2-Clause License
+#
+
+"""Observer for tracking pipeline startup timing.
+
+This module provides an observer that measures how long each processor's
+``start()`` method takes during pipeline startup. It works by tracking
+when a ``StartFrame`` arrives at a processor (``on_process_frame``) versus
+when it leaves (``on_push_frame``), giving the exact ``start()`` duration
+for each processor in the pipeline.
+
+Example::
+
+    observer = StartupTimingObserver()
+
+    @observer.event_handler("on_startup_timing_report")
+    async def on_report(observer, report):
+        for t in report.processor_timings:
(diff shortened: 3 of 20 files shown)


examples/foundational/29-turn-tracking-observer.py
```diff
@@ -97,14 +98,23 @@ async def run_bot(transport: BaseTransport, runner_args: RunnerArguments):
...
     async def on_latency_measured(observer, latency_seconds):
         logger.info(f"⏱️ User-to-bot latency: {latency_seconds:.3f}s")
 
+    @startup_observer.event_handler("on_startup_timing_report")
+    async def on_startup_timing_report(observer, report):
+        logger.info(f"Total startup: {report.total_duration_secs:.3f}s")
+        for timing in report.processor_timings:
+            logger.info(f"  {timing.processor_name}: {timing.duration_secs:.3f}s")
+
+    @startup_observer.event_handler("on_transport_readiness_measured")
```
filipi87: I wonder if we shouldn’t call it `on_startup_transport_connected_report`, to match the other event and to make it clear that we evaluate it only at startup, not in the case of reconnections.

src/pipecat/frames/frames.py
```diff
@@ -1910,6 +1910,17 @@ class StopFrame(ControlFrame, UninterruptibleFrame):
     pass
 
 
+@dataclass
+class ClientConnectedFrame(SystemFrame):
```
filipi87, line 1926: Now I’m even wondering whether this is the right indicator for transport readiness. 

For example, with Daily, we could consider the bot ready as soon as the bot has already joined to a Daily room. Because the other participant might take several seconds to actually join. 

In that case, the transport would be ready as soon as we have joined the room, not when the first client joins, since we don’t have control over that part.

For SmallWebRTC and WebSocket, things are a bit different, since the transport will only try to establish a connection once the other participant joins. However, this is not true for Daily, LiveKit, HeyGen, and Tavus.

Or maybe we need two indicators: 
- on_connected: Called when bot successfully joined a room. This would be true for all transports which use SFU (Daily, LiveKit, HeyGen, and Tavus).
- on_client_connected: Called when a client (participant) connects.

What do you think ?

markbackman: @filipi87 great catch! I've updated the PR to account for this.

I've updated the PR to reflect this. There are now two events from the StartupTimingObserver:
- `on_startup_timing_report` which reports the processor startup times
- `on_transport_timing_report` which reports the bot and client connection timings from the receipt of the StartFrame

This will make things much cleaner.

markbackman: Note for the SmallWebRTCTransport and FastAPIWebsocketTransport, there is no bot connected timing because in these cases the bot is the server; there is no connection time. Do you think this makes sense?

filipi87: Yeah, I think it makes sense. Another option is if the bot connection timing is the same as the client connection time, but I’m not sure if this is worth it.

examples/foundational/29-turn-tracking-observer.py
```diff
@@ -97,14 +98,25 @@ async def run_bot(transport: BaseTransport, runner_args: RunnerArguments):
...
+    async def on_startup_timing_report(observer, report):
+        logger.info(f"Total startup: {report.total_duration_secs:.3f}s")
+        for timing in report.processor_timings:
+            logger.info(f"  {timing.processor_name}: {timing.duration_secs:.3f}s")
+
+    @startup_observer.event_handler("on_transport_timing_report")
+    async def on_transport_timing_report(observer, report):
+        if report.bot_connected_secs is not None:
+            logger.info(f"Bot connected: {report.bot_connected_secs:.3f}s")
+        logger.info(f"Client connected: {report.client_connected_secs:.3f}s")
```
filipi87, line 118: 💯

src/pipecat/frames/frames.py
markbackman: That wouldn't be accurate though; the bot connects immediately so the connection time reported is None. We could make that 0s, but that might also be confusing. Perhaps it's fine as is for the initial release.

src/pipecat/observers/startup_timing_observer.py
```diff
@@ -0,0 +1,326 @@
...
+
+        # Timestamp (ns) when we first see a StartFrame arrive at a processor.
+        self._start_frame_arrival_ns: Optional[int] = None
+
+        # Bot connected timing (stored for inclusion in the transport report).
+        self._bot_connected_secs: Optional[float] = None
+
+        # Wall clock reference for converting monotonic ns to Unix timestamps.
+        self._wall_clock_ref: Optional[float] = None
+        self._mono_clock_ref_ns: Optional[int] = None
```
filipi87: I guess we could remove these variables and the `_mono_to_wall` method.

All reported values are relative durations from the first `StartFrame`, so the `_start_frame_arrival_ns` timestamp already provides what we need, I guess there’s no need for wall-clock conversion. What do you think ?

src/pipecat/observers/startup_timing_observer.py
```diff
@@ -0,0 +1,326 @@
...
+        Args:
+            processor_types: Optional tuple of processor types to measure.
+                If None, all non-internal processors are measured.
+            **kwargs: Additional arguments passed to parent class.
+        """
+        super().__init__(**kwargs)
+        self._processor_types = processor_types
+
+        # Map processor ID -> (processor, arrival_timestamp_ns)
+        self._arrivals: Dict[int, Tuple[FrameProcessor, int]] = {}
```
filipi87: nit: What if we create a class for this, instead of relying on Tuple ?

Something like this:
```
@dataclass
class ArrivalInfo:
    processor: FrameProcessor
    arrival_ts_ns: int
```

src/pipecat/observers/startup_timing_observer.py
```diff
@@ -0,0 +1,326 @@
...
+            self._start_frame_arrival_ns = data.timestamp
+            self._wall_clock_ref = time.time()
+            self._mono_clock_ref_ns = data.timestamp
+        elif data.frame.id != self._start_frame_id:
+            return
+
+        # When the StartFrame reaches a PipelineSink, all processors have
+        # completed start(). PipelineSinks use direct mode so the outermost
+        # sink fires last within the same synchronous call chain.
+        if isinstance(data.processor, PipelineSink):
```
filipi87: Have you tested this with `ParallelPipeline`? I guess we also include a `PipelineSink` in that case. I’m just wondering if it works as expected when using a `ParallelPipeline`.

src/pipecat/transports/daily/transport.py
```diff
@@ -2578,6 +2580,8 @@ async def _on_joined(self, data):
             if error:
                 await self._on_error(f"Unable to start transcription: {error}")
         await self._call_event_handler("on_joined", data)
```
filipi87, line 2585: To make all transports consistent, as we did with `on_client_connected`, shouldn’t we invoke `on_connected` here?
The same applies to the other transports, such as Tavus, where we currently invoke `on_joined`.

src/pipecat/utils/tracing/service_attributes.py
```diff
@@ -70,7 +68,7 @@ def add_tts_span_attributes(
     model: str,
     voice_id: str,
     text: Optional[str] = None,
-    settings: Optional["ServiceSettings"] = None,
+    settings: Optional[Dict[str, Any]] = None,
```
filipi87: Are these changes intentional?

src/pipecat/utils/tracing/service_decorators.py
```diff
@@ -219,7 +219,7 @@ async def tracing_context(self, text):
             tracer = trace.get_tracer("pipecat")
             with tracer.start_as_current_span(span_name, context=parent_context) as span:
                 try:
-                    settings = getattr(self, "_settings", None)
+                    settings = getattr(self, "_settings", {})
```
filipi87: Same here. Maybe there was some conflict when merging changes from `main`?

src/pipecat/utils/tracing/service_attributes.py
markbackman: No. Removed.

src/pipecat/utils/tracing/service_decorators.py
markbackman: Yes, conflict from main. Removed.

examples/foundational/29-turn-tracking-observer.py
markbackman: Naming changed to `on_transport_timing_report`.

src/pipecat/observers/startup_timing_observer.py
markbackman: I'm envisioning a case where we want to build a dashboard UI showing the events. We want events to be relative to the exact timing that occurs during the call. So, while we have the relative timings, we do need a top-level wall clock time to ensure the events are correctly displayed. I've made a small change to simplify. Please take a look.

src/pipecat/observers/startup_timing_observer.py
```diff
@@ -54,12 +54,13 @@ class ProcessorStartupTiming(BaseModel):
...
         processor_name: The name of the processor.
-        start_time: Unix timestamp when the processor's start() began.
+        start_offset_secs: Offset in seconds from the StartFrame to when this
+            processor's start() began.
         duration_secs: How long the processor's start() took, in seconds.
     """
 
     processor_name: str
-    start_time: float
+    start_offset_secs: float
```
markbackman, line 72: I'm including a `start_offset_secs` here to make it easier to plot the data. Each measurement of the processor is the duration for the processor to start. The `start_offset_secs` is the time offset from the actual start time. Without this, developers will have to add up times to get the right timing.

src/pipecat/observers/startup_timing_observer.py
markbackman: Good idea!

src/pipecat/observers/startup_timing_observer.py
markbackman: We talked about using the `PipelineTask` to inform the `TaskObserver` that the pipeline has started. I've made this change in https://github.com/pipecat-ai/pipecat/pull/3881/changes/389d0c3fb6adbbf7d926a3e3a6fea49a8aaae3d0. I think this is a clean way to handle things.

src/pipecat/transports/daily/transport.py
markbackman: I renamed Tavus to `on_connected`. `on_joined` already exists for Daily, so we'd have to add an additional event that does the same thing. It's probably worth doing that in a separate PR.

src/pipecat/observers/startup_timing_observer.py
filipi87: Makes sense. 👍

filipi87 (approved):
This will be a great improvement Mark. 👏🚀

PR #3885, opened 2026-03-01
Add latency breakdown to UserBotLatencyObserver

markbackman:
## Context

The goal is to measure the user to bot latency and to display the contributing factors to that latency number. The possible contributors are:
- VAD stop_secs
- STT latency
- Turn detection
- LLM latency
- Pre-TTS text aggregation
- TTS latency
- Function call execution time

The sequence of events looks something like this:
user-bot-events

We are breaking the measurement down into the following categories:
- User turn secs: This measures from when the user actually stops speaking (VADStoppedSpeakingFrame - VADParams stop_secs) to when the UserStoppedSpeakingFrame occurs. There are two main contributors to this phase: STT latency and turn detection. When an STT latency is available, we display that correspond TTFB value. We do not explicitly display the turn detection latency because it is difficult to characterize, but it is generally going to be the main contributor to the long user_turn_secs values.
- TTFB latencies: STT, LLM, TTS. The LLM and TTS portions directly impact the response time. The STT is a function of more complex logic, which I mention above.
- Text aggregation latency: This is the time to build a sentence from the LLM streamed token. If sentence aggregation is in place, we measure it to help the developer understand the impact.
- Function call latency: Time spent executing tool/function calls between LLM roundtrips. Measured from `FunctionCallInProgressFrame` to `FunctionCallResultFrame`, making previously invisible API call time visible in the waterfall.

Note: we do not provide a measurement for the turn result because it does not have a direct impact on response time. The result itself is used by the LLMUserAggregator logic to determine how to respond, when the user stop strategy includes a turn analyzer. The turn result latency is available in a MetricsFrame for those that want access, perhaps for troubleshooting or debugging.

## Summary

- Added `on_latency_breakdown` event to `UserBotLatencyObserver` with a `LatencyBreakdown` model providing per-service TTFB, text aggregation, user turn duration, and function call latency metrics
- Added `on_first_bot_speech_latency` event measuring time from `ClientConnectedFrame` to first `BotStartedSpeakingFrame`, with a latency breakdown including per-service metrics. Skipped when user speaks first (only meaningful for greetings).
- Added `FunctionCallMetrics` tracking to `LatencyBreakdown`, measuring execution time of each function call via `FunctionCallInProgressFrame` to `FunctionCallResultFrame`
- Measures `user_turn_secs` from actual user silence (VAD timestamp minus stop_secs) to turn release, capturing VAD silence detection, STT finalization, and turn analyzer wait time
- Resets metric accumulators on `InterruptionFrame` to discard stale metrics from cancelled LLM/TTS cycles
- All breakdown data models are Pydantic `BaseModel` with `start_time` (Unix timestamp) and `duration_secs` fields, enabling `.model_dump()` serialization and timeline plotting
- Updated example 29 with waterfall-style latency breakdown display including function calls

## Testing

```bash
uv run pytest tests/test_user_bot_latency_observer.py -v
```


examples/foundational/29-turn-tracking-observer.py
```diff
@@ -131,6 +182,33 @@ async def on_turn_ended(observer, turn_number, duration, was_interrupted):
...
+        for ttfb in non_stt:
+            logger.info(f"  {ttfb.processor}: TTFB {ttfb.duration_secs:.3f}s")
+            if not fc_shown and breakdown.function_calls:
+                for fc in breakdown.function_calls:
+                    logger.info(f"  {fc.function_name}: {fc.duration_secs:.3f}s")
+                fc_shown = True
+
+        if breakdown.text_aggregation:
+            ta = breakdown.text_aggregation
+            logger.info(f"  {ta.processor}: text aggregation {ta.duration_secs:.3f}s")
```
aconchillo: Since this might be common. I'm wondering if we could have a function that we could call that returns us a chronological list of strings. Like:

```
events = observer.chronological_events()
```

So people could just iterate and print.

aconchillo (approved):

aconchillo:
Approving! This is great. Just added a comment to simplify examples/apps a bit.

examples/foundational/29-turn-tracking-observer.py
markbackman: Good idea!

markbackman: Ok, pushing a commit with this function.

PR #3886, opened 2026-03-01
fix(sarvam): standardize STT/TTS User-Agent headers

dhruvladia-sarvam:
This PR standardizes how Sarvam services send the `User-Agent` header so backend/proxy systems reliably receive Pipecat's SDK identity instead of provider SDK defaults.

What changed:

- **Sarvam STT**
  - Ensures Pipecat headers are passed through the SDK-supported path (`headers` at client init + connect-time request options).
  - Adds debug logging for outbound STT connect headers to verify handshake header values during testing.

- **Sarvam TTS**
  - Aligns TTS websocket header behavior with STT format by sending `User-Agent` inside `additional_headers`.
  - Avoids ambiguous websocket-client default behaviour by explicitly controlling header fields.
  - Adds debug logging for outbound TTS websocket connect headers.

### DIFF
--- a/src/pipecat/services/sarvam/stt.py
+++ b/src/pipecat/services/sarvam/stt.py
@@ -266,15 +266,10 @@ class SarvamSTTService(STTService):
 
         # Initialize Sarvam SDK client
         self._sdk_headers = sdk_headers()
-        # NOTE: We avoid passing non-standard kwargs here because different sarvamai
-        # versions expose different constructor signatures (static type checkers
-        # complain otherwise). We instead inject headers best-effort below.
-        self._sarvam_client = AsyncSarvamAI(api_subscription_key=api_key)
-        for attr in ("default_headers", "_default_headers", "headers", "_headers"):
-            d = getattr(self._sarvam_client, attr, None)
-            if isinstance(d, dict):
-                d.update(self._sdk_headers)
-                break
+        # Pass Pipecat SDK headers directly at client construction time so they are
+        # merged by the Sarvam SDK's client wrapper and consistently applied to
+        # WebSocket handshake requests.
+        self._sarvam_client = AsyncSarvamAI(api_subscription_key=api_key, headers=self._sdk_headers)
         self._websocket_context = None
         self._socket_client = None
         self._receive_task = None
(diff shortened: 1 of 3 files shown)


src/pipecat/services/sarvam/tts.py
```diff
@@ -1013,12 +912,17 @@ async def _connect_websocket(self):
             if self._websocket and self._websocket.state is State.OPEN:
                 return
 
+            ws_additional_headers = {
+                "api-subscription-key": self._api_key,
+                **sdk_headers(),
+            }
+            logger.debug(
+                f"Sarvam TTS websocket connect headers: additional_headers={ws_additional_headers}"
```
markbackman: This leaks the API key in plain text, so not a good idea to debug log...

markbackman (commented):
Change generally seems fine. You'll need to rebase on `main` because a number of changes have happened in this past week regarding settings.

Also, please ensure no API keys are leaked. See my comment and test the debug logs you're sending.

Also, can you please amend the commit name from `ruff fix` to something more accurate.

Lastly, please add a changelog entry:
https://github.com/pipecat-ai/pipecat/blob/main/CONTRIBUTING.md

src/pipecat/services/sarvam/stt.py
```diff
@@ -517,20 +512,29 @@ async def _connect(self):
                 connect_kwargs["prompt"] = self._settings.prompt
 
             def _connect_with_sdk_headers(connect_fn, **kwargs):
-                # Different SDK versions may use different kwarg names.
                 # If prompt is unsupported at connect-time, retry without it.
+                # Headers are supplied through request_options because this is a
+                # documented SDK parameter that survives SDK signature changes.
+                request_options = {"additional_headers": self._sdk_headers}
+                logger.debug(
```
markbackman: Is this log necessary? I would recommend removing unless it needs to be there.

markbackman (commented):
A few more comments before this one is ready.

src/pipecat/services/sarvam/tts.py
```diff
@@ -1013,12 +1013,15 @@ async def _connect_websocket(self):
...
+            }
+
             self._websocket = await websocket_connect(
                 self._websocket_url,
-                additional_headers={
-                    "api-subscription-key": self._api_key,
-                    **sdk_headers(),
-                },
+                additional_headers=ws_additional_headers,
+                user_agent_header=None,
```
markbackman: Is this necessary?

markbackman (approved):
LGTM! Thanks for the quick cleanup.

PR #3887, opened 2026-03-01
Fix LLM response duplication on interleaved tool calling

yuki901:
## Summary
- Flush the `_aggregation` buffer before adding tool calls to context in both assistant context aggregators (universal and legacy), fixing text duplication during interleaved tool calling.

## Problem

When models produce interleaved responses (text → tool_call → text in a single response), the `_aggregation` buffer is not flushed before `FunctionCallInProgressFrame` is processed. This causes two issues:

1. **The pre-tool text is already spoken via TTS** — the user hears "Let me check the weather..." in real-time as it streams
2. **The unflushed buffer leaks into the post-tool assistant message** — the same pre-tool text appears again in context after the tool result, so the user hears it a second time when the LLM continues its response

## Reproduction

1. Use an Anthropic model (e.g. Claude 4.6 Sonnet) with registered tools
2. Ask a question that triggers a tool call, e.g. "What is the weather in New York?"
3. The model produces interleaved output: text ("Let me check...") → `get_weather` tool call → text ("The weather is...")
4. The user hears "Let me check the weather" spoken twice — once when it streams, and again when it leaks into the post-tool response

**Without fix — context after tool result (from actual logs):**
```python
# Pre-tool text is MISSING from the tool call message (already spoken via TTS but not committed to context)...
{"role": "assistant", "content": [
    {"type": "tool_use", "id": "toolu_015g...", "name": "get_weather", "input": {"location": "New York, NY"}}
]},
{"role": "user", "content": [
    {"type": "tool_result", "tool_use_id": "toolu_015g...", "content": "...72 degrees and sunny."}
]},
# ...and LEAKED here — the unflushed buffer is dumped alongside the post-tool response,
# causing the user to hear the pre-tool text AGAIN via TTS
{"role": "assistant", "content": [
    {"type": "text", "text": "Let me check the weather in New York for you right away!"},
    {"type": "text", "text": "The current weather in New York, NY is 72°F and sunny! ☀️ ..."}
]},
```

**With fix — correct context:**
```python
# Pre-tool text correctly flushed and committed to context alongside the tool_use
{"role": "assistant", "content": [
    {"type": "text", "text": "Sure! Let me check the current weather in New York for you right away!"},
    {"type": "tool_use", "id": "toolu_01Ci...", "name": "get_weather", "input": {"location": "New York, NY"}}
]},
{"role": "user", "content": [
    {"type": "tool_result", "tool_use_id": "toolu_01Ci...", "content": "...72 degrees and sunny."}
]},
# Post-tool response contains only new text — no duplication
```

## Fix

Add `await self.push_aggregation()` at the start of `_handle_function_call_in_progress` in both:
- `LLMAssistantAggregator` (universal — `llm_response_universal.py`)
- `LLMAssistantContextAggregator` (legacy — `llm_response.py`)

This ensures any accumulated text is committed to context before the tool call message is added, so it is not carried over into the next LLM turn.

## Test plan
- [x] Existing aggregator unit tests pass (19/19)
- [x] Manual test: with fix — pre-tool text spoken once, correctly grouped with tool_use in context
- [x] Manual test: without fix — pre-tool text spoken twice (once live, once leaked into post-tool response)


markbackman:
The change is missing. I only see a changelog entry. Can you push the changes to this branch?

yuki901:
Oh sorry for that. Added!

yuki901:
If you guys have time can you review this one?

PR #3888, opened 2026-03-01
Re-inject turn completion instructions after LLM context reset

markbackman:
## Summary

- Fixed turn completion instructions being lost when `LLMMessagesUpdateFrame` replaces the LLM context via `set_messages()`
- When `filter_incomplete_user_turns` is enabled, the turn completion system message is now re-injected after context replacement, preventing the LLM from stopping marker emission
- This fixes the warning "filter_incomplete_user_turns is enabled but LLM response did not contain turn completion markers" that occurs when using Pipecat Flows (or any component that sends `LLMMessagesUpdateFrame`)

## Testing

- `uv run pytest tests/test_context_aggregators_universal.py tests/test_user_turn_completion_mixin.py`


filipi87 (approved):
Nice catch! LGTM!

PR #3889, opened 2026-03-02
AIC Voice Focus version update & concurrency safety issue on audio buffer.

gkmngrgn:
# Summary

- **Fix BufferError:** Existing exports of data: object cannot be re-sized that occurs in production when concurrent filter() or stop() calls interleave with an in-flight filter() holding a memoryview on the mutable _audio_buffer bytearray across await yield points.
- The fix snapshots the needed audio blocks into immutable bytes and trims the buffer before any await, so no memoryview (or any reference to the mutable bytearray) is held across async yield points.
- Add two regression tests that reproduce the exact concurrency scenarios: concurrent filter()/filter() interleaving and stop() during filter().

# Problem

`AICFilter.filter()` created a `memoryview` on `self._audio_buffer` (a `bytearray`) and held it alive across await `self._processor.process_async(...)`. In production with concurrent calls, another task could run `filter()` (calling `_audio_buffer.extend()`) or `stop()` (calling `_audio_buffer.clear()`) while the `memoryview` still exported the buffer, causing Python to raise:

```
    BufferError: Existing exports of data: object cannot be re-sized
```

It's not easy to reproduce locally because it requires concurrent async interleaving but wrote a test for simulate it.

# Fix

Replace the `memoryview` on the mutable `bytearray` with a `bytes` snapshot taken before any await:

```
    blocks_data = bytes(self._audio_buffer[:total_size])
    self._audio_buffer = self._audio_buffer[total_size:]
    # all awaits below only reference the immutable blocks_data
```

The extra copy is negligible — typical block size is 320 bytes (10ms of 16kHz 16-bit audio).

### DIFF
--- a/changelog/3889.changed.md
+++ b/changelog/3889.changed.md
@@ -1,0 +1,3 @@
+- Support for Voice Focus 2.0 models.
+  - Updated `aic-sdk` to `~=2.1.0` to support Voice Focus 2.0 models.
+  - Cleaned unused `ParameterFixedError` exception handling in `AICFilter` parameter setup. 
--- a/changelog/3889.fixed.md
+++ b/changelog/3889.fixed.md
@@ -1,0 +1,1 @@
+- Fixed `BufferError: Existing exports of data: object cannot be re-sized` in `AICFilter` caused by holding a `memoryview` on the mutable audio buffer across async yield points.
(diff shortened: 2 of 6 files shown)


markbackman (approved):
LGTM! Thank you 🙇

PR #3896, opened 2026-03-02
Add broadcast_interruption() to FrameProcessor

aconchillo:
## Summary

- Added `broadcast_interruption()` to `FrameProcessor` — pushes `InterruptionFrame` both upstream and downstream directly from the calling processor, eliminating the round-trip through the pipeline task
- Deprecated `push_interruption_task_frame_and_wait()` — now delegates to `broadcast_interruption()` and emits a `DeprecationWarning`
- Removed `event` field and `complete()` from `InterruptionFrame` / `InterruptionTaskFrame`, and all associated special-case logic (`_wait_for_interruption` flag, queue bypass, sink `frame.complete()`)

## Motivation

`push_interruption_task_frame_and_wait()` sends an `InterruptionTaskFrame` upstream to the pipeline task, which creates an `InterruptionFrame` and sends it back downstream. This round-trip causes:

1. **Swallowed events** — the frame can be blocked before reaching the sink, the event never gets set, and a 5s timeout fires
2. **Race conditions** — on a short VAD utterance, the transcription arrives before the `InterruptionFrame` comes back downstream, then gets wiped out
3. **Complexity** — `_wait_for_interruption` flag, queue bypass logic, `frame.complete()` obligations on every processor that stops an `InterruptionFrame`

`broadcast_interruption()` solves all three by pushing the `InterruptionFrame` in both directions immediately, with no event, no waiting, and no round-trip.

## Test plan

- [x] `uv run pytest tests/test_frame_processor.py -v` — all 9 tests pass
- [x] `uv run pytest tests/test_pipeline.py -v` — all 24 tests pass
- [x] `uv run pytest tests/test_stt_mute_filter.py -v` — all 8 tests pass
- [x] `uv run pytest tests/ -x --ignore=tests/integration` — all 727 tests pass (1 skipped)


src/pipecat/processors/frame_processor.py
```diff
@@ -790,43 +776,32 @@ async def push_frame(self, frame: Frame, direction: FrameDirection = FrameDirect
...
-        and sends it downstream through the pipeline. An `asyncio.Event` is
-        attached to both frames so the caller can wait until the interruption
-        has fully traversed the pipeline. The event is set when the
-        `InterruptionFrame` reaches the pipeline sink. If the frame does
-        not complete within the given timeout, a warning is logged and the
-        event is forcibly set so the caller is unblocked.
-
-        Args:
-            timeout: Maximum seconds to wait for the interruption to complete.
+        .. deprecated::
```
markbackman: Include the deprecated version

```suggestion
        .. deprecated:: 0.0.104
```

markbackman (approved):
Change looks good. I've done some testing and it works as expected.

This approach is much simpler to implement and reason about, so I do like that!

src/pipecat/processors/frame_processor.py
aconchillo: Done

PR #3900, opened 2026-03-02
Adding the LemonSlice transport integration

filipi87:
Adding the LemonSliceTransport layer to easily integrate the LemonSlice Avatar with any Pipecat pipeline.

Users need to pass in a LemonSlice API key and a LemonSlice agent ID or publicly accessible avatar image url to start the session.
Users can pass in a Daily room to use or our API will create one for them under the LemonSlice domain.
This transport layer will add a "pipecat user" to the room to feed audio to LemonSlice through a special audio stream. Then a LemonSlice avatar will join the call and play the synced audio and video.

LemonSlice API docs: https://lemonslice.com/docs/api-reference/create-session

Bringing the work from this [PR](https://github.com/pipecat-ai/pipecat/pull/3791), so we can solve the merge conflicts.


markbackman (approved):
LGTM!

(Pushed a commit with some minor cleanup, but it looks good and the demo works well.)

PR #3903, opened 2026-03-03
fix: race conditions in asyncio.Event usage patterns

### LINKED ISSUE #3402: Race Conditions and Lost Notifications Due to Unsafe asyncio.Event Reset Patterns

DukeDeSouth:
Fixes #3402.

Two fixes for race conditions caused by unsafe `asyncio.Event` reset patterns:

**EventNotifier** — replaced `asyncio.Event` with `asyncio.Condition` to prevent notification loss in multi-consumer scenarios. The old `wait()` → `clear()` sequence allowed a new `notify()` to be silently erased when multiple waiters interleaved.

**IdleFrameProcessor / UserIdleProcessor** — moved `Event.clear()` from the `finally` block to the `try` block (success path only). The `finally` pattern unconditionally cleared signals that arrived during async callback execution, causing false-positive idle triggers even when the user was actively speaking. This matches the already-correct pattern in `pipeline/task.py`.

Added regression tests that confirm both issues. All 33 existing + new tests pass.

Made with [Cursor](https://cursor.com)

### DIFF
--- a/src/pipecat/processors/idle_frame_processor.py
+++ b/src/pipecat/processors/idle_frame_processor.py
@@ -85,7 +85,6 @@ class IdleFrameProcessor(FrameProcessor):
         while True:
             try:
                 await asyncio.wait_for(self._idle_event.wait(), timeout=self._timeout)
+                self._idle_event.clear()
             except asyncio.TimeoutError:
                 await self._callback(self)
-            finally:
-                self._idle_event.clear()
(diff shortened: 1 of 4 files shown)


aconchillo:
Closing.`IdleFrameProcessor` fix in #3797. The `EventNotifier` is supposed to be a single-consusmer (docstring needs update), no current usage of multi-consumer.

PR #3910, opened 2026-03-03
TTS improvements + BotStoppedSpeakingFrame without using VAD timeout

filipi87:
## Summary
- Introduced `push_start_frame` parameter to `TTSService.__init__` so subclasses no longer need to manually call `create_audio_context` or push `TTSStartedFrame` — the base class handles it automatically
- Refactored all 25+ TTS service implementations to use the new `push_start_frame=True` pattern, reducing boilerplate across providers (AWS, Azure, Camb, Cartesia, Deepgram, ElevenLabs, Fish, Google, Gradium, Groq, Hathora, Hume, Inworld, Kokoro, LMNT, Minimax, Neuphonic, Nvidia, OpenAI, Piper, Rime, Sarvam, Speechmatics, Xtts, and others)
- Fixed `BotStoppedSpeakingFrame` emission: now emitted as soon as `TTSStoppedFrame` is received, with a fallback silence-based timeout increased to reduce false positives
- Fixed handling of `TTSSpeakFrame` to correctly treat it as a standalone utterance (not reusing the current turn context ID)


src/pipecat/services/tts_service.py
```diff
@@ -1210,16 +1210,6 @@ async def _audio_context_task_handler(self):
                 del self._audio_contexts[context_id]
                 await self.on_audio_context_completed(context_id=context_id)
                 self.reset_active_audio_context()
-
-                # Append some silence between sentences.
```
filipi87, line 1214: If we want to keep sending silence, we should do this after `TTSStoppedFrame`, but I’m not sure this is needed.

filipi87: Doing it here would trigger `BotStartedSpeaking` again.

src/pipecat/transports/base_output.py
```diff
@@ -44,12 +44,15 @@
...
     TTSAudioRawFrame,
+    TTSStoppedFrame,
 )
 from pipecat.processors.frame_processor import FrameDirection, FrameProcessor
 from pipecat.transports.base_transport import TransportParams
 from pipecat.utils.time import nanoseconds_to_seconds
 
 BOT_VAD_STOP_SECS = 0.35
+# Only used as a fallback
+BOT_VAD_STOP_FALLBACK_SECS = 3
```
filipi87, line 55: - `BOT_VAD_STOP_SECS` is still used when we are continuously receiving audio, in the “speech stream” case, for example, with Tavus and HeyGen.

- `BOT_VAD_STOP_FALLBACK_SECS`: a new value used only as a fallback. It is used inside `_next_frame`, but we should rarely (if ever) hit this path now. We are increasing the timeout to allow enough time to receive the `TTSStoppedSpeakingFrame` instead of relying on this fallback. This help preventing from receiving multiple bot started and stopped speaking events in the same turn.

filipi87:
@markbackman, @aconchillo, I will merge the changes from this PR into the previous one so you can review everything at once, since they are related and modify the same files.

You can review everything here:
- https://github.com/pipecat-ai/pipecat/pull/3804

PR #3914, opened 2026-03-04
Make max_context_tokens and max_unsummarized_messages independently optional

markbackman:
## Context

I bumped into this while porting Gradient Bang to use the built-in context summarization. This makes the experience better for developers (otherwise, they have to set one of the values to a high number to "disable" either token or messages.

## Summary

- Allow `max_context_tokens` and `max_unsummarized_messages` in `LLMAutoContextSummarizationConfig` to be independently set to `None` to disable that summarization threshold
- Previously both thresholds were always active (OR'd together), forcing users to set a very large number to effectively disable one — now setting `None` cleanly disables it
- Validate that at least one threshold must remain set (raises `ValueError` if both are `None`)
- Apply the same changes to the deprecated `LLMContextSummarizationConfig` for consistency

## Testing

```bash
uv run pytest tests/test_context_summarization.py tests/test_llm_context_summarizer.py -v
```

### DIFF
--- a/changelog/3914.changed.md
+++ b/changelog/3914.changed.md
@@ -1,0 +1,1 @@
+- `max_context_tokens` and `max_unsummarized_messages` in `LLMAutoContextSummarizationConfig` (and deprecated `LLMContextSummarizationConfig`) can now be set to `None` independently to disable that summarization threshold. At least one must remain set.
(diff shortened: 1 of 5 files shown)


filipi87 (approved):
Makes sense. Nice improvement. 🚀

kompfner (commented):
Makes a lot of sense!

PR #3915, opened 2026-03-04
Add per-tool function call timeout_secs

markbackman:
## Context

We added a global `function_call_timeout_secs` param set to 10 seconds, but this is too coarse to be effective. What if I have two function calls, A and B. A runs fast, expected to complete in <1s; B runs slow, expected to complete in >10s. We need a way to set the timeout more granularly—on a per function call basis. That's what I'm adding here. When you register a function, you can now override the global fallback.

I'm also improving the warning message so developers can understand why a timeout occurred and can fix it, if needed.

This was inspired by a Discord help thread where a user was surprised to get a timeout error when running a long running function call.

## Summary

- Added optional `timeout_secs` parameter to `register_function()` and `register_direct_function()` for per-tool function call timeout control
- When set, overrides the global `function_call_timeout_secs` from the LLM constructor for that specific function
- Improved the timeout warning message to mention both the per-tool `timeout_secs` and the global `function_call_timeout_secs` options

## Testing

```python
# Per-tool timeout (30s for a slow tool)
llm.register_function("slow_database_query", handler, timeout_secs=30)

# Uses global default (10s)
llm.register_function("fast_lookup", handler)
```

- `uv run ruff check src/pipecat/services/llm_service.py` — passes
- `uv run pytest` — no regressions

### DIFF
--- a/changelog/3915.added.md
+++ b/changelog/3915.added.md
@@ -1,0 +1,1 @@
+- Added optional `timeout_secs` parameter to `register_function()` and `register_direct_function()` for per-tool function call timeout control, overriding the global `function_call_timeout_secs` default.
--- a/src/pipecat/services/llm_service.py
+++ b/src/pipecat/services/llm_service.py
@@ -120,12 +120,15 @@ class FunctionCallRegistryItem:
         function_name: The name of the function (None for catch-all handler).
         handler: The handler for processing function call parameters.
         cancel_on_interruption: Whether to cancel the call on interruption.
+        timeout_secs: Optional per-tool timeout in seconds. Overrides the global
+            ``function_call_timeout_secs`` for this specific function.
     """
 
     function_name: Optional[str]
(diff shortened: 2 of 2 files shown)


kompfner (approved):
Nice!

We probably also want to open a follow-up PR in Pipecat Flows adding `timeout_secs` to both `FlowsFunctionSchema` and the `flows_direct_function` decorator, exactly like we did [here](https://github.com/pipecat-ai/pipecat-flows/pull/225/changes) for `cancel_on_interruption`.

markbackman:
> Nice!
> 
> We probably also want to open a follow-up PR in Pipecat Flows adding `timeout_secs` to both `FlowsFunctionSchema` and the `flows_direct_function` decorator, exactly like we did [here](https://github.com/pipecat-ai/pipecat-flows/pull/225/changes) for `cancel_on_interruption`.

Indeed! Someone asked for this. I'll try to knock this out today.

PR #3916, opened 2026-03-04
daily-transport: add cloud-audio-only recording option

vipyne:

as per daily docs here: https://docs.daily.co/guides/products/live-streaming-recording/recording-calls-with-the-daily-api#cloud-audio-only


markbackman (approved):
Nice! LGTM

vr000m (approved):

PR #3918, opened 2026-03-04
Wire up system_instruction in OpenAI, Anthropic, and AWS Bedrock

aconchillo:
## Summary

- Wire up `system_instruction` in `BaseOpenAILLMService`, `AnthropicLLMService`, and `AWSBedrockLLMService` so it works as a default system prompt, matching the behavior of the Google services.
- This enables sharing a single `LLMContext` across multiple LLM services, where each service provides its own system instruction independently. The shared context does not need to contain a system message — each service injects its own via the `system_instruction` constructor parameter.
- For each service, `system_instruction` acts as a fallback: if the context already provides a system message, the context value takes precedence.

Code now looks like this:

  ```python
  llm = OpenAILLMService(
      api_key=os.getenv("OPENAI_API_KEY"),
      system_instruction="You are a helpful assistant.",
  )

  context = LLMContext()

      context.add_message({"role": "user", "content": "Please introduce yourself."})
  ```

## Changes

- **OpenAI** (`BaseOpenAILLMService`): Prepend `system_instruction` as a `{"role": "system", ...}` message in `build_chat_completion_params()`.
- **Anthropic** (`AnthropicLLMService`): Inject `system_instruction` when the universal `LLMContext` has no system prompt (`NotGiven`).
- **AWS Bedrock** (`AWSBedrockLLMService`): Inject `system_instruction` as `[{"text": "..."}]` in the Bedrock system format.

## Testing

- [x] All 707 tests pass (`uv run pytest tests/ --ignore=tests/integration`)
- [x] Ruff lint + format pass
- [ ] Manual test: verify `system_instruction` is prepended for OpenAI
- [ ] Manual test: verify `system_instruction` works as fallback for Anthropic
- [ ] Manual test: verify `system_instruction` works as fallback for AWS Bedrock

### DIFF
--- a/src/pipecat/services/anthropic/llm.py
+++ b/src/pipecat/services/anthropic/llm.py
@@ -17,7 +17,7 @@ import io
 import json
 import re
 from dataclasses import dataclass, field
-from typing import Any, ClassVar, Dict, List, Literal, Optional, Union
+from typing import Any, Dict, List, Literal, Optional, Union
 
 import httpx
 from loguru import logger
@@ -38,7 +38,6 @@ from pipecat.frames.frames import (
     LLMFullResponseEndFrame,
     LLMFullResponseStartFrame,
     LLMMessagesFrame,
-    LLMTextFrame,
     LLMThoughtEndFrame,
     LLMThoughtStartFrame,
     LLMThoughtTextFrame,
@@ -219,6 +218,7 @@ class AnthropicLLMService(LLMService):
         client=None,
         retry_timeout_secs: Optional[float] = 5.0,
         retry_on_timeout: Optional[bool] = False,
+        system_instruction: Optional[str] = None,
         **kwargs,
     ):
         """Initialize the Anthropic LLM service.
@@ -230,6 +230,7 @@ class AnthropicLLMService(LLMService):
             client: Optional custom Anthropic client instance.
             retry_timeout_secs: Request timeout in seconds for retry logic.
             retry_on_timeout: Whether to retry the request once if it times out.
+            system_instruction: Optional system instruction to use as the system prompt.
             **kwargs: Additional arguments passed to parent LLMService.
         """
         params = params or AnthropicLLMService.InputParams()
@@ -265,6 +266,9 @@ class AnthropicLLMService(LLMService):
         )  # if the client is provided, use it and remove it, otherwise create a new one
         self._retry_timeout_secs = retry_timeout_secs
         self._retry_on_timeout = retry_on_timeout
+        self._system_instruction = system_instruction
+        if self._system_instruction:
+            logger.debug(f"{self}: Using system instruction: {self._system_instruction}")
 
     def can_generate_metrics(self) -> bool:
         """Check if this service can generate usage metrics.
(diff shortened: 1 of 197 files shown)


src/pipecat/services/anthropic/llm.py
```diff
@@ -395,9 +398,11 @@ def _get_llm_invocation_params(
         # Universal LLMContext
         if isinstance(context, LLMContext):
             adapter: AnthropicLLMAdapter = self.get_llm_adapter()
-            params = adapter.get_llm_invocation_params(
+            params: AnthropicLLMInvocationParams = adapter.get_llm_invocation_params(
                 context, enable_prompt_caching=self._settings.enable_prompt_caching
             )
+            if self._system_instruction and params["system"] is NotGiven:
```
markbackman: Don't we want to give `self._system_instruction` the priority here? Maybe just:
```
if self._system_instruction:
    params["system"] = self._system_instruction
```

so system_instruction overrides the system message obtained from the context.

markbackman (commented):
We also need to make the same change to `AWSBedrockLLMService`.

src/pipecat/services/aws/llm.py
```diff
@@ -1019,7 +1022,9 @@ def _get_llm_invocation_params(
         # Universal LLMContext
         if isinstance(context, LLMContext):
             adapter: AWSBedrockLLMAdapter = self.get_llm_adapter()
-            params = adapter.get_llm_invocation_params(context)
+            params: AWSBedrockLLMInvocationParams = adapter.get_llm_invocation_params(context)
+            if self._system_instruction and not params["system"]:
```
markbackman: Same comment about AWS. I think we want this to just check for the existence of `self._system_instruction`.

aconchillo:
> We also need to make the same change to `AWSBedrockLLMService`.

Done

src/pipecat/services/anthropic/llm.py
aconchillo: Yeah, maybe that's what we want.

markbackman (approved):
LGTM! Nice improvement 🙌

src/pipecat/services/anthropic/llm.py
aconchillo: Done!

src/pipecat/services/aws/llm.py
aconchillo: Done!

kompfner:
Looks like several services that inherit from `OpenAILLMService` override `build_chat_completion_params()` (e.g. Mistral). We need to add the system instruction there, too. I'll work it into the changes in https://github.com/pipecat-ai/pipecat/pull/3877.

kompfner:
> Looks like several services that inherit from `OpenAILLMService` override `build_chat_completion_params()` (e.g. Mistral). We need to add the system instruction there, too. I'll work it into the changes in #3877.

Fixed in [this commit](https://github.com/pipecat-ai/pipecat/pull/3877/commits/560d2306e8d5d47f4bb5c98e6d4c80cacd9c55c9) in that PR.

PR #3926, opened 2026-03-05
Update dependency version ranges for flexibility

markbackman:
## Summary

- Widened version ranges across core, optional, and dev dependencies to reduce unnecessary version conflicts while maintaining compatibility
- Removed `cartesia` SDK dependency — `CartesiaHttpTTSService` now uses direct `aiohttp` calls with optional session injection
- Removed `fal-client` SDK dependency — both `FalSTTService` and `FalImageGenService` now use direct HTTP calls to `fal.run`, fixing a ~4x latency regression introduced in fal-client 0.7.0+
- Emptied `silero` extra since `onnxruntime` is now a core dependency
- Bumped `pyloudnorm` to 0.2.0, `soxr` to 1.0.0, `faster-whisper` to 1.2.1, and several other packages to latest compatible versions
- Added upper bounds to packages that were missing them (`numba`, `wait_for2`, `transformers`, `hume`, `camb`, `livekit-api`, `pyjwt`, `tracing`, `anthropic`, etc.)
- Bumped dev dependencies: `build`, `coverage`, `pip-tools`, `pre-commit`

## Testing

- Run `uv lock && uv sync` to verify dependency resolution
- Run `uv run pytest` to verify tests pass
- Test Fal STT with `examples/foundational/07w-interruptible-fal.py`
- Test Cartesia HTTP TTS with `examples/foundational/07-interruptible-cartesia-http.py`


examples/foundational/07c-interruptible-deepgram-sagemaker.py
```diff
@@ -110,7 +110,7 @@ async def run_bot(transport: BaseTransport, runner_args: RunnerArguments):
     async def on_client_connected(transport, client):
         logger.info(f"Client connected")
         # Kick off the conversation.
-        context.add_message({"role": "system", "content": "Please introduce yourself to the user."})
+        context.add_message({"role": "user", "content": "Please introduce yourself to the user."})
```
markbackman, line 113: AWS was unhappy with back-to-back system messages.

markbackman:
I removed the `cartesia` and `fal-client` packages because they were just simple HTTP interfaces. This simplies the API and in the case of Fal, appears to speed up performance.

I made the `aiohttp_session` `__init__` arg optional for both, so this isn't breaking.

Also, in general, I was careful with upgrades and tested changes to make sure all was well. The evals run before releasing will be a good final test.

pyproject.toml
```diff
@@ -20,27 +20,27 @@ classifiers = [
...
     "aiohttp>=3.11.12,<4",
     "audioop-lts~=0.2.1; python_version>='3.13'",
-    "docstring_parser~=0.16",
+    "docstring_parser>=0.16,<1",
     "loguru~=0.7.3",
     "Markdown>=3.7,<4",
     "nltk>=3.9.3,<4",
     "numpy>=1.26.4,<3",
     "Pillow>=11.1.0,<13",
     "protobuf~=5.29.6",
```
markbackman, line 32: I want to bump the `protobuf` version, but the `nvidia-riva-client` package pins a specific `grpcio` version, which `protobuf` uses. Here's the relevant issue where this is tracked:
https://github.com/nvidia-riva/python-clients/issues/165

Also, they're pinning `setuptools` which constrains our dev tooling. Ideally, that's removed from requirements.txt since it puts an unnecessary constraint on Pipecat.

aconchillo (approved):

PR #3927, opened 2026-03-05
feat(assemblyai): add vad_threshold parameter for U3 Pro

zkleb-aai:
Add vad_threshold parameter to AssemblyAIConnectionParams to support voice activity detection threshold configuration for the u3-rt-pro model.

This parameter allows users to align AssemblyAI's VAD threshold with their external VAD systems (e.g., Silero VAD) to avoid the "dead zone" where AssemblyAI transcribes speech that the external VAD hasn't detected yet, which can delay interruption handling.

- Range: 0.0 to 1.0 (lower = more sensitive)
- Default: 0.3 (API default when not sent)
- Only applicable to u3-rt-pro model
- Automatically included in WebSocket query parameters

Recommended usage: Set vad_threshold to match your VAD's activation threshold (e.g., both at 0.3) for optimal performance.


markbackman (approved):
LGTM!

Please add a changelog entry too:

PR #3930, opened 2026-03-05
Add `push_empty_transcripts` param to `BaseWhisperSTTService` to push received empty transcripts downstream

dakshdua:
- Adds `push_empty_transcripts` param to `BaseWhisperSTTService` to push received empty transcripts downstream


markbackman (approved):
LGTM! Thanks for the docstrings to explain the case. Cool solution!

PR #3932, opened 2026-03-06
Warn when both system_instruction and context system message are set

aconchillo:
## Summary

- Added warnings in `BaseOpenAILLMService`, `AnthropicLLMService`, and `AWSBedrockLLMService` when both `system_instruction` and a system message in the context are set. The constructor's `system_instruction` always takes precedence.


markbackman (approved):
LGTM

PR #3933, opened 2026-03-06
Fix Daily transport log level and eval script import

aconchillo:
## Summary

- Changed `participant_updated` log level from `debug` to `trace` in Daily transport (too noisy at debug level)
- Fixed `LiveOptions` import in eval script to use the re-export from `pipecat.services.deepgram.stt` instead of importing directly from `deepgram`

### DIFF
--- a/src/pipecat/transports/daily/transport.py
+++ b/src/pipecat/transports/daily/transport.py
@@ -2747,7 +2747,7 @@ class DailyTransport(BaseTransport):
 
     async def _on_participant_updated(self, participant):
         """Handle participant updated events."""
-        logger.debug(f"{self} participant updated: {participant}")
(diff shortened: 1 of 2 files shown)


src/pipecat/transports/daily/transport.py
```diff
@@ -2747,7 +2747,7 @@ async def _on_participant_left(self, participant, reason):
 
     async def _on_participant_updated(self, participant):
         """Handle participant updated events."""
-        logger.debug(f"{self} participant updated: {participant}")
+        logger.trace(f"{self} participant updated: {participant}")
```
aconchillo, line 2750: `debug` was quite annoying.

markbackman (approved):

PR #3936, opened 2026-03-06
Fixed TTS context not being appended to the assistant message history

filipi87:
## Summary

- Fixed TTSService to push LLMAssistantPushAggregationFrame at the correct point: when TTSStoppedFrame is processed  and the TTS context is cleaned up, rather than after each TTSAudioRawFrame                                           
- Fixed _stop_frame_handler to track context_id from TTSStartedFrame so that timeout-triggered TTSStoppedFrame frames carry the correct context ID — enabling the aggregation push to work correctly in that path                         
- Consolidated LLMAssistantPushAggregationFrame emission from multiple locations into a single canonical place (_process_tts_stopped_frame), removing duplicate pushes from _tts_audio_queue_handler and tts_process_generator
- Fixed is_yielding_frames to only be set to True when a TTSAudioRawFrame is received (not for any frame type)

### DIFF
--- a/changelog/3936.fixed.md
+++ b/changelog/3936.fixed.md
@@ -1,1 +1,1 @@
-- Fixed `TTSService` to push `LLMAssistantPushAggregationFrame` when the base class is responsible for pushing `TTSStoppedFrame`.
+- Fixed TTS context not being appended to the assistant message history when using `TTSSpeakFrame` with `append_to_context=True` with some TTS providers.
--- a/changelog/3936.fixed.md
+++ b/changelog/3936.fixed.md
@@ -1,0 +1,1 @@
+- Fixed `TTSService` to push `LLMAssistantPushAggregationFrame` when the base class is responsible for pushing `TTSStoppedFrame`.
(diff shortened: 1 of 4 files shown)


changelog/3936.fixed.md
```diff
@@ -0,0 +1 @@
+- Fixed `TTSService` to push `LLMAssistantPushAggregationFrame` when the base class is responsible for pushing `TTSStoppedFrame`.
```
kompfner: Might want to describe this in terms of the user-facing effect...was the problem that `TTSSpeakFrame`s with `append_to_context=True` weren't being honored by some TTS services?

kompfner (approved):
If I understand the problem (which I _think_ I do, after reading some code; read my comment to confirm), then LGTM (once the comment has been addressed).

changelog/3936.fixed.md
filipi87: Exactly, for TTS services like `ElevenLabs`, where the base TTSService was responsible for emitting `TTSStoppedFrame`, this was not happening. I will improve the comment.

filipi87: Done.

PR #3937, opened 2026-03-06
Fix context summarization leaving orphaned tool responses in kept context.

### LINKED ISSUE #3832: enable_context_summarization truncates message history between tool_calls and tool role messages, causing OpenAI 400 Error

filipi87:
## Summary                                                                                                              
 - Fixed context summarization leaving orphaned tool responses in the kept context when the matching tool call was moved to the summarized portion

The root cause was that `_get_function_calls_in_progress_index` scanned the full message list, so a tool response in the kept range would resolve its tool call, allowing the assistant message with `tool_calls` to be summarized while its response remained, causing an API rejection
 
Fixed by bounding the scan to `summary_end`, so any tool response outside the summarized range is treated as absent, keeping the entire tool call group together in the kept context.

Fixes https://github.com/pipecat-ai/pipecat/issues/3832.

### DIFF
--- a/src/pipecat/utils/context/llm_context_summarization.py
+++ b/src/pipecat/utils/context/llm_context_summarization.py
@@ -382,25 +382,33 @@ class LLMContextSummarizationUtil:
         return total
 
     @staticmethod
-    def _get_function_calls_in_progress_index(messages: List[dict], start_idx: int) -> int:
+    def _get_function_calls_in_progress_index(
+        messages: List[dict], start_idx: int, summary_end: int
+    ) -> int:
         """Find the earliest message index with incomplete function calls.
 
-        Scans messages to identify function/tool calls that haven't received
-        their results yet. This prevents summarizing incomplete tool interactions
-        which would break the request-response pairing.
+        Scans messages from ``start_idx`` up to (but not including)
+        ``summary_end`` to identify tool calls whose responses either don't
+        exist yet or fall in the kept portion of the context (>= summary_end).
+        This prevents summarizing tool call requests when their responses would
+        remain in the kept context as orphans, which the OpenAI API rejects.
 
         Args:
             messages: List of messages to check.
             start_idx: Index to start checking from.
+            summary_end: Exclusive upper bound for the scan (the first kept
+                message index). Only tool responses within this range count as
+                completing a call; responses beyond it are treated as absent,
+                leaving the call "in progress".
 
         Returns:
             Index of first message with function call in progress, or -1 if all
-            function calls are complete.
+            function calls are complete within the scanned range.
         """
         # Track tool call IDs mapped to their message index
         pending_tool_calls: dict[str, int] = {}
 
-        for i in range(start_idx, len(messages)):
+        for i in range(start_idx, summary_end):
             msg = messages[i]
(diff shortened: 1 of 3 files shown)


src/pipecat/utils/context/llm_context_summarization.py
```diff
@@ -382,25 +382,33 @@ def estimate_context_tokens(context: LLMContext) -> int:
         return total
 
     @staticmethod
-    def _get_function_calls_in_progress_index(messages: List[dict], start_idx: int) -> int:
+    def _get_function_calls_in_progress_index(
```
kompfner: Method name/signature is unclear, as it's no longer true that it only looks for in-progress function calls. (Also, it shouldn't mention plural "calls")

How about something along the lines of `_get_earliest_function_call_not_resolved_in_range(messages, start, end)`. Then it would be a little more self-documenting.

kompfner (approved):
Nice catch! One comment.

src/pipecat/utils/context/llm_context_summarization.py
filipi87: Makes sense, renamed. 👍

PR #3938, opened 2026-03-06
fix: replace bare except handlers with specific exception types

koriyoshi2041:
## Summary

Replaces 7 bare `except:` handlers across the codebase with specific exception types, following Python best practices (PEP 8, B001/E722).

**Changes:**
- `examples/foundational/39-mcp-stdio.py` — `except:` → `except (json.JSONDecodeError, KeyError, TypeError):`
- `examples/foundational/39c-multiple-mcp.py` — same
- `src/pipecat/adapters/services/bedrock_adapter.py` — `except:` → `except (json.JSONDecodeError, ValueError):`
- `src/pipecat/serializers/telnyx.py` — `except:` → `except Exception:`
- `src/pipecat/serializers/twilio.py` — `except:` → `except Exception:`
- `src/pipecat/services/aws/llm.py` — `except:` → `except (json.JSONDecodeError, ValueError):`
- `src/pipecat/services/mcp_service.py` — `except:` → `except Exception:`

Each replacement was chosen based on the specific context: JSON parsing uses `JSONDecodeError`/`ValueError`, while generic error logging uses `Exception` (which still excludes `SystemExit`, `KeyboardInterrupt`, and `GeneratorExit`).

## Motivation

Bare `except:` catches everything including `SystemExit` and `KeyboardInterrupt`, which can mask critical errors and prevent proper process shutdown. This is flagged by flake8 (E722) and ruff (B001).

## Test plan

- [x] No behavioral change — each handler already uses `pass` or fallback logic
- [x] Verified exception types match the operations in each try block

### DIFF
--- a/src/pipecat/adapters/services/bedrock_adapter.py
+++ b/src/pipecat/adapters/services/bedrock_adapter.py
@@ -209,7 +209,7 @@ class AWSBedrockLLMAdapter(BaseLLMAdapter[AWSBedrockLLMInvocationParams]):
                     tool_result_content = [{"json": content_json}]
                 else:
                     tool_result_content = [{"text": message["content"]}]
-            except (json.JSONDecodeError, ValueError):
+            except (json.JSONDecodeError, ValueError, AttributeError):
                 tool_result_content = [{"text": message["content"]}]
 
             return {
--- a/src/pipecat/adapters/services/bedrock_adapter.py
+++ b/src/pipecat/adapters/services/bedrock_adapter.py
@@ -209,7 +209,7 @@ class AWSBedrockLLMAdapter(BaseLLMAdapter[AWSBedrockLLMInvocationParams]):
                     tool_result_content = [{"json": content_json}]
                 else:
                     tool_result_content = [{"text": message["content"]}]
-            except:
+            except (json.JSONDecodeError, ValueError):
                 tool_result_content = [{"text": message["content"]}]
 
             return {
--- a/src/pipecat/services/aws/llm.py
+++ b/src/pipecat/services/aws/llm.py
@@ -369,7 +369,7 @@ class AWSBedrockLLMContext(OpenAILLMContext):
                     tool_result_content = [{"json": content_json}]
                 else:
                     tool_result_content = [{"text": message["content"]}]
-            except (json.JSONDecodeError, ValueError):
+            except (json.JSONDecodeError, ValueError, AttributeError):
                 tool_result_content = [{"text": message["content"]}]
 
             return {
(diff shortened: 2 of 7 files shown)


src/pipecat/services/mcp_service.py
```diff
@@ -298,7 +298,7 @@ async def _list_tools_helper(self, session):
 
         try:
             logger.debug(f"Found {len(available_tools)} available tools")
-        except:
+        except Exception:
```
markbackman: This try/except around a debug log seems unnecessary. If `len(available_tools)` fails, the code continues to iterate `available_tools.tools` on line 304 anyway.

I'd suggest removing the try/except entirely and changing this to `len(available_tools.tools)` to match what's actually used below.
```
logger.debug(f"Found {len(available_tools.tools)} available tools")
```

src/pipecat/adapters/services/bedrock_adapter.py
```diff
@@ -209,7 +209,7 @@ def _from_standard_message(self, message: LLMStandardMessage) -> dict[str, Any]:
                     tool_result_content = [{"json": content_json}]
                 else:
                     tool_result_content = [{"text": message["content"]}]
-            except:
+            except (json.JSONDecodeError, ValueError):
```
markbackman: The try block calls `message["content"].strip()`, which would raise `AttributeError` if content is `None`. The bare `except:` previously caught that, but `(json.JSONDecodeError, ValueError)` won't. Since the fallback in the except handler also accesses `message["content"]`, I'd suggest either using except `Exception:` here or adding `AttributeError` to the tuple.

src/pipecat/services/aws/llm.py
```diff
@@ -371,7 +369,7 @@ def from_standard_message(self, message):
                     tool_result_content = [{"json": content_json}]
                 else:
                     tool_result_content = [{"text": message["content"]}]
-            except:
+            except (json.JSONDecodeError, ValueError):
```
markbackman: Same comment from bedrock_adapter.py applies here. Either using except `Exception:` here or adding `AttributeError` to the tuple.

markbackman (commented):
Thanks for the submission. Just a few comments.

src/pipecat/services/mcp_service.py
koriyoshi2041: Good catch\! Removed the try/except entirely and switched to `len(available_tools.tools)` to match the iteration below.

src/pipecat/adapters/services/bedrock_adapter.py
koriyoshi2041: You're right — added `AttributeError` to the tuple so `None` content is handled the same way the bare except did.

src/pipecat/services/aws/llm.py
koriyoshi2041: Done — same fix applied here, `AttributeError` added to the tuple.

markbackman (approved):
LGTM! Thanks for the contribution and quick fixes 🙌

PR #3939, opened 2026-03-06
Adopt the `settings` pattern for OpenAI Realtime session properties

kompfner:
Move `session_properties` into `OpenAIRealtimeLLMSettings`, making `settings` the canonical way to configure OpenAI Realtime — matching the pattern used across the rest of the codebase. The `session_properties` init arg is now deprecated in favor of `settings=OpenAIRealtimeLLMSettings(session_properties=...)`.

`model` and `system_instruction` are synced bidirectionally between the top-level settings fields and `session_properties.model`/`.instructions`, with top-level taking precedence on conflict.

### DIFF
--- a/src/pipecat/services/openai/realtime/llm.py
+++ b/src/pipecat/services/openai/realtime/llm.py
@@ -10,8 +10,9 @@ import base64
 import io
 import json
 import time
-from dataclasses import dataclass
-from typing import Any, Optional
+from dataclasses import dataclass, field
+from dataclasses import fields as dataclass_fields
+from typing import Any, Dict, Mapping, Optional, Type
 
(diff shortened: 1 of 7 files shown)


src/pipecat/services/openai/realtime/llm.py
```diff
@@ -140,6 +245,10 @@ def __init__(
                 Defaults to "wss://api.openai.com/v1/realtime".
             session_properties: Configuration properties for the realtime session.
                 If None, uses default SessionProperties.
+
+                .. deprecated::
```
markbackman, line 249: Add version.

markbackman (approved):
LGTM

PR #3940, opened 2026-03-06
Adopt the `settings` pattern for Grok Realtime session properties

kompfner:
Move `session_properties` into `GrokRealtimeLLMSettings`, making `settings` the canonical way to configure Grok Realtime — matching the pattern used across the rest of the codebase. The `session_properties` init arg is now deprecated in favor of `settings=GrokRealtimeLLMSettings(session_properties=...)`.

`system_instruction` is synced bidirectionally between the top-level settings field and `session_properties.instructions`, with top-level taking precedence on conflict. (Unlike OpenAI Realtime, Grok's `SessionProperties` has no `model` field, so no model sync is needed.)

### DIFF
--- a/src/pipecat/services/grok/realtime/llm.py
+++ b/src/pipecat/services/grok/realtime/llm.py
@@ -13,8 +13,9 @@ https://docs.x.ai/docs/guides/voice/agent
 import base64
 import json
 import time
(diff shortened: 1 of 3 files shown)


markbackman (approved):
LGTM

PR #3941, opened 2026-03-06
STT services: settings and examples fixes

markbackman:

Primarily changes to STT services to align to STTSettings norms.

Also includes:
- Examples fixes for services
- Update to COMMUNITY_INTEGRATIONS.md guide per feedback in https://github.com/pipecat-ai/pipecat/pull/3877
- Update examples to avoid repeated `system` messages in the context
- Add versions to deprecation docstrings

### DIFF
--- a/COMMUNITY_INTEGRATIONS.md
+++ b/COMMUNITY_INTEGRATIONS.md
@@ -233,14 +233,14 @@ def can_generate_metrics(self) -> bool:
 
 ### Service Settings
 
-Every STT, LLM, TTS, and image-generation service exposes a **Settings dataclass** that serves two roles:
+Every AI service (STT, LLM, TTS, image generation, etc.) exposes a **Settings dataclass** that serves two roles:
 
 1. **Store mode** — the service's `self._settings` holds the current value of every runtime-updatable field.
-2. **Delta mode** — an update frame carries only the fields that changed; unset fields remain `NOT_GIVEN`.
+2. **Delta mode** — an update frame (e.g. `TTSUpdateSettingsFrame`) specifies only the fields that should change; unspecified fields remain `NOT_GIVEN`.
 
 #### Defining your Settings class
 
-Extend `STTSettings`, `TTSSettings`, `LLMSettings`, or `ImageGenSettings`. The base classes already provide common fields (e.g. `model`, `voice`, `language`). You only need to add **service-specific knobs that should be runtime-updatable**:
+Extend `STTSettings`, `TTSSettings`, `LLMSettings`, or `ImageGenSettings` (or, if your service directly subclasses `AIService`, `ServiceSettings`). The base classes already provide common fields (e.g. `model`, `voice`, `language`). You only need to add **service-specific knobs that should be runtime-updatable**:
 
 ```python
 from dataclasses import dataclass, field
@@ -320,7 +320,7 @@ svc = MyTTSService(
 
 #### Reacting to runtime changes
 
-STT, LLM, and TTS services support runtime configuration changes via `*UpdateSettingsFrame`s (e.g. `STTUpdateSettingsFrame`, `TTSUpdateSettingsFrame`, `LLMUpdateSettingsFrame`).
+AI services support runtime configuration changes via `*UpdateSettingsFrame`s (e.g. `STTUpdateSettingsFrame`, `TTSUpdateSettingsFrame`, `LLMUpdateSettingsFrame`).
 
 To react to runtime setting changes, override `_update_settings`. The base implementation applies the delta to `self._settings` and returns a `dict` mapping each changed field name to its **pre-update** value. Your override should call `super()` first, then act on the changed fields. A common implementation might look like:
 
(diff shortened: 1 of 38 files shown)


src/pipecat/services/assemblyai/stt.py
```diff
@@ -651,7 +657,7 @@ async def _handle_speech_started(self, message: SpeechStartedMessage):
         await self.start_processing_metrics()
         await self.broadcast_frame(UserStartedSpeakingFrame)
         if self._should_interrupt:
-            await self.push_interruption_task_frame_and_wait()
+            await self.broadcast_interruption()
```
markbackman, line 660: This was missing from a previous PR.

src/pipecat/services/aws/stt.py
```diff
@@ -99,13 +99,13 @@ def __init__(
         # 1. Initialize default_settings with hardcoded defaults
         default_settings = AWSTranscribeSTTSettings(
             model=None,
-            language=self.language_to_service_language(Language.EN) or "en-US",
+            language=self.language_to_service_language(Language.EN),
```
markbackman, line 102: No need for a fallback. It's being set to a specific value.

src/pipecat/processors/aggregators/llm_context.py
```diff
@@ -255,7 +255,7 @@ def get_messages_for_persistent_storage(self) -> List[LLMContextMessage]:
         this method, which is part of the public API of OpenAILLMContext but
         doesn't need to be for LLMContext.
 
-        .. deprecated::
+        .. deprecated:: 0.0.92
```
markbackman, line 258: I looked up these older deprecations in the CHANGELOG and picked the corresponding version.

src/pipecat/services/google/tts.py
```diff
@@ -514,7 +514,7 @@ class GoogleHttpTTSSettings(TTSSettings):
 
 
 @dataclass
-class GoogleStreamTTSSettings(TTSSettings):
+class GoogleTTSSettings(TTSSettings):
```
markbackman, line 515: Aligning the naming with the class it's likely to be used with.

markbackman: Note: other changes in this file weren't included in the previous PR. I think they weren't saved by me.

src/pipecat/services/speechmatics/stt.py
```diff
@@ -507,6 +500,13 @@ def __init__(
             **kwargs,
         )
 
+        # Build SDK config from settings, then resolve model from operating_point
```
markbackman: This fixes a bug I introduced in the previous PR. The client and config need to happen after the default_settings are set.

COMMUNITY_INTEGRATIONS.md
```diff
@@ -233,14 +233,14 @@ def can_generate_metrics(self) -> bool:
 
 ### Service Settings
 
-Every STT, LLM, TTS, and image-generation service exposes a **Settings dataclass** that serves two roles:
+Every AI service (STT, LLM, TTS, image generation, etc.) exposes a **Settings dataclass** that serves two roles:
```
markbackman, line 236: Clean up from comments in https://github.com/pipecat-ai/pipecat/pull/3877

src/pipecat/processors/aggregators/llm_context.py
kompfner: Way to go the extra mile and do your homework! 📚

src/pipecat/services/google/stt.py
```diff
@@ -653,7 +653,7 @@ async def _reconnect_if_needed(self):
     async def set_languages(self, languages: List[Language]):
         """Update the service's recognition languages.
 
-        .. deprecated::
+        .. deprecated:: 0.0.104
```
kompfner, line 656: 👍

src/pipecat/services/speechmatics/stt.py
kompfner: Wait why can't you move this block right above `super().__init__` (but after step 4 above)? Then you wouldn't have to do the hacky call to `self._sync_model_name_to_metrics()`, as super would handle it...right?

src/pipecat/services/google/tts.py
kompfner: Wait, is the name change technically breaking? Didn't we already ship `GoogleStreamTTSSettings`?

kompfner (approved):
Couple minor comments to address. Pre-approving. Nice!

src/pipecat/services/speechmatics/stt.py
markbackman: Good point. I'll move this above to save the sync'ing.

src/pipecat/services/google/tts.py
markbackman: Oof. I think you're right.

markbackman: Ok, I'm going to alias it:
```
#: .. deprecated:: 0.0.105
#:     Use ``GoogleTTSSettings`` instead.
GoogleStreamTTSSettings = GoogleTTSSettings
```

PR #3942, opened 2026-03-06
Add AudioConfig class to AWSNovaSonicLLMService for non-deprecated au…

kompfner:
…dio configuration

The audio fields (sample rates, sample sizes, channel counts) on the deprecated `Params` class had no non-deprecated equivalent. This adds an `AudioConfig` class and `audio_config` init arg so users can specify audio configuration without relying on the deprecated `params` parameter.


markbackman (approved):
Nice and clean ✨

PR #3946, opened 2026-03-07
Review TTS settings

markbackman:

A variety of things, including bug fixes, recategorizing settings, updating Fish Audio because it now works different, updated examples to resolve issue, and added information to the TTSSettings docstring indicating to which class the settings class applies.

### DIFF
--- a/examples/foundational/07m-interruptible-aws-strands.py
+++ b/examples/foundational/07m-interruptible-aws-strands.py
@@ -104,7 +104,7 @@ async def run_bot(transport: BaseTransport, runner_args: RunnerArguments):
 
     # Create Strands agent processor
     try:
-        agent = build_agent(model_id="us.anthropic.claude-3-5-haiku-20241022-v1:0", max_tokens=8000)
+        agent = build_agent(model_id="us.anthropic.claude-sonnet-4-6", max_tokens=8000)
         llm = StrandsAgentsProcessor(agent=agent)
         logger.info("Successfully created Strands agent for NAB customer service coaching")
     except Exception as e:
@@ -152,7 +152,7 @@ async def run_bot(transport: BaseTransport, runner_args: RunnerArguments):
                     messages=[
                         {
                             "role": "user",
-                            "content": f"Greet the user and introduce yourself.",
+                            "content": f"Greet the user and introduce yourself. Don't use emojis.",
                         }
                     ],
                     run_llm=True,
--- a/src/pipecat/services/aws/stt.py
+++ b/src/pipecat/services/aws/stt.py
@@ -107,7 +107,7 @@ class AWSTranscribeSTTService(WebsocketSTTService):
             _warn_deprecated_param("language", AWSTranscribeSTTSettings, "language")
             default_settings.language = self.language_to_service_language(language)
 
-        # 3. No params to apply
+        # 3. (No step 3, as there's no params object to apply)
 
         # 4. Apply settings delta (canonical API, always wins)
         if settings is not None:
@@ -161,7 +161,7 @@ class AWSTranscribeSTTService(WebsocketSTTService):
         """Apply a settings delta and reconnect if anything changed."""
         changed = await super()._update_settings(delta)
 
-        if changed:
+        if changed and self._websocket:
             await self._disconnect()
             await self._connect()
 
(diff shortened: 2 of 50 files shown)


examples/foundational/07m-interruptible-aws-strands.py
```diff
@@ -152,7 +152,7 @@ async def on_client_connected(transport, client):
                     messages=[
                         {
                             "role": "user",
-                            "content": f"Greet the user and introduce yourself.",
+                            "content": f"Greet the user and introduce yourself. Don't use emojis.",
```
kompfner, line 155: 🤔

src/pipecat/services/aws/stt.py
```diff
@@ -158,22 +158,12 @@ def get_service_encoding(self, encoding: str) -> str:
...
-            return changed
-
-        # TODO: someday we could reconnect here to apply updated settings.
-        # Code might look something like the below:
-        # if changed and self._websocket:
-        #     await self._disconnect()
-        #     await self._connect()
-
-        self._warn_unhandled_updated_settings(changed)
+        if changed:
```
kompfner: You didn't want to go with the `changed and self._websocket` check, to prevent disconnecting/reconnecting in the case we weren't yet connected in the first place? Maybe that was only a theoretical concern...I don't quite remember why the commented-out code was the way it was.

kompfner: Glad this "just worked" though!

src/pipecat/services/azure/stt.py
```diff
@@ -159,23 +159,15 @@ def language_to_service_language(self, language: Language) -> Optional[str]:
...
-        #         # language is applied at recognizer initialization
-        #         await self._disconnect()
-        #         await self._connect()
-
-        self._warn_unhandled_updated_settings(changed)
+        if "language" in changed:
+            self._speech_config.speech_recognition_language = (
+                self._settings.language or language_to_azure_language(Language.EN_US)
+            )
+            await self._disconnect()
```
kompfner: Same comment - we don't need to first check whether disconnect/reconnect even needed?

Btw, cool that you're going through and actually doing all the things that I put in "TODOs" and punted on to avoid exploding the scope of the initial PR. These TODOs represent runtime settings updates that never previously worked, so this is all new functionality!

...so, actually, maybe worth shouting out in a CHANGELOG entry that we actually implemented runtime settings updates to a number of services?

src/pipecat/services/gradium/stt.py
```diff
@@ -151,19 +154,18 @@ def __init__(
         default_settings = GradiumSTTSettings(
             model=None,
             language=None,
+            delay_in_frames=None,
         )
 
-        # 2. (no deprecated direct args for this service)
-
-        # 3. Apply params overrides — only if settings not provided
+        # 2. Apply params overrides — only if settings not provided
```
kompfner: Oh huh I took the opposite approach to these comments here: https://github.com/pipecat-ai/pipecat/pull/3943. We maybe ought to align.

src/pipecat/services/groq/tts.py
```diff
@@ -34,7 +34,7 @@
 
 @dataclass
 class GroqTTSSettings(TTSSettings):
-    """Settings for the Groq TTS service.
+    """Settings for GroqTTSService.
```
kompfner, line 37: I'll go ahead and make this style of update to the LLM services too

kompfner (approved):

examples/foundational/07m-interruptible-aws-strands.py
markbackman: Sonnet 4.6 littered the response with emojis for some reason. Just a simple suggestion to the model to not do that.

src/pipecat/services/aws/stt.py
markbackman: I went through and added checks for each service. This is now more consistent.

src/pipecat/services/azure/stt.py
markbackman: Same, added checks.

Also, added a changelog entry.

src/pipecat/services/gradium/stt.py
markbackman: I went with your approach to include the blank step.

PR #3952, opened 2026-03-07
Add Settings class attribute alias to all service classes

markbackman:
## Summary

- Added a `Settings` class-level alias on every service class (STT, LLM, TTS, image, vision, video) pointing to its corresponding settings dataclass
- Developers can now discover the right settings class via the service class itself (e.g. `GoogleSTTService.Settings(...)`) without needing to know or import the separate `GoogleSTTSettings` class name
- Both `ServiceClass.Settings(...)` and the direct `ServiceNameSettings(...)` import continue to work — this is additive convenience, not a replacement

## Testing

```python
# Before: need to know both names
from pipecat.services.google import GoogleSTTService, GoogleSTTSettings
stt = GoogleSTTService(settings=GoogleSTTSettings(languages=[Language.EN_US]))

# After: discoverable from the service class
from pipecat.services.google import GoogleSTTService
stt = GoogleSTTService(settings=GoogleSTTService.Settings(languages=[Language.EN_US]))
```

### DIFF
--- a/src/pipecat/services/anthropic/llm.py
+++ b/src/pipecat/services/anthropic/llm.py
@@ -159,6 +159,7 @@ class AnthropicLLMService(LLMService):
     Can use custom clients like AsyncAnthropicBedrock and AsyncAnthropicVertex.
     """
 
+    Settings = AnthropicLLMSettings
     _settings: AnthropicLLMSettings
 
     # Overriding the default adapter to use the Anthropic one.
--- a/COMMUNITY_INTEGRATIONS.md
+++ b/COMMUNITY_INTEGRATIONS.md
@@ -271,12 +271,15 @@ The rule of thumb: if a caller might send an update frame to change it at runtim
 
 #### Wiring settings into `__init__`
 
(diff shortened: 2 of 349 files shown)


aconchillo:
Makes sense. LGTM!

aconchillo (approved):

filipi87 (approved):
LGTM!

src/pipecat/services/anthropic/llm.py
```diff
@@ -159,6 +159,7 @@ class AnthropicLLMService(LLMService):
     Can use custom clients like AsyncAnthropicBedrock and AsyncAnthropicVertex.
     """
 
+    Settings = AnthropicLLMSettings
```
kompfner, line 162: Can we also update the COMMUNITY_INTEGRATIONS.md doc that shows the `_settings: MySettings` pattern?

kompfner (approved):
One comment. Otherwise, LGTM! Nice convenience.

src/pipecat/services/anthropic/llm.py
markbackman: Done!

PR #3956, opened 2026-03-08
Fix turn completion mixin not resetting state when no `InterruptionFrame` is emitted

radhikagpt1208:
## Summary

Fixed `UserTurnCompletionLLMServiceMixin` not resetting turn completion state between LLM responses.

## Problem

- When `filter_incomplete_user_turns=True` is enabled, the mixin uses `_turn_complete_found` to track whether the current LLM response started with a `✓` (complete) marker. Once set to `True`, this flag should be cleared before the next LLM response so marker detection runs fresh.
- The flag was never reset between responses. The `process_frame` method had an `elif isinstance(frame, LLMFullResponseEndFrame)` branch intended to handle this, but it was dead code - `LLMFullResponseEndFrame` is created inside the LLM service and pushed outward via `push_frame`, so it never arrives at `process_frame` from upstream.
- The flag was reset by `process_frame` handling an `InterruptionFrame`, which fires whenever the user starts speaking. This meant normal conversational turns worked correctly. However, in scenarios where no user speech occurs (e.g. idle timeout prompts that trigger a new LLM run directly), no `InterruptionFrame` is ever emitted and `_turn_complete_found` stays `True` indefinitely.
- **Result:** in these no-speech scenarios, subsequent LLM responses bypass marker detection entirely, hitting the early return in `_push_turn_text`:

```python
if self._turn_complete_found:
    await self.push_frame(LLMTextFrame(text))  # raw ✓ leaks to TTS
    return
```

- The ✓ character is pushed directly to TTS and spoken aloud or appears in the transcript.

## Fix

Added a push_frame override to `UserTurnCompletionLLMServiceMixin` that intercepts the outgoing `LLMFullResponseEndFrame` and calls `_turn_reset()` before passing it along. This ensures `_turn_complete_found` is cleared after every LLM response regardless of whether the user speaks.
- Also removed the now-redundant `elif isinstance(frame, LLMFullResponseEndFrame)` dead code branch from `process_frame`.

## Test plan

- [x] Added regression test `test_turn_state_reset_after_llm_full_response_end_frame` that verifies `_turn_complete_found` is reset to `False` after `LLMFullResponseEndFrame` is pushed
- [x] All existing `test_user_turn_completion_mixin.py` tests continue to pass


markbackman (approved):
Really nice find! Thanks for catching and fixing this. I pushed a commit to this branch to make the docstring and changelog more concise (plus squashed branches).

Approving and merging!

PR #3957, opened 2026-03-08
Move turn completion instructions to system_instruction

markbackman:
## Summary

- Fixed turn completion instructions being injected as a system message in the LLM context. This caused warning spam when `system_instruction` was also set, didn't persist across full context updates, and broke LLMs that don't support consecutive system messages.
- Turn completion instructions are now composed into the LLM service's `system_instruction` field via `_compose_system_instruction()`, with the original value saved in `_base_system_instruction` for restoration when turn completion is disabled.
- Handles mid-session changes: enabling/disabling turn completion, updating `system_instruction` while active, and changing `user_turn_completion_config` while active.

## Testing

```bash
uv run pytest tests/test_user_turn_completion_mixin.py tests/test_context_aggregators_universal.py -v
```

### DIFF
--- a/changelog/3957.fixed.md
+++ b/changelog/3957.fixed.md
@@ -1,0 +1,1 @@
+- Fixed turn completion instructions being injected as a context system message instead of using `system_instruction`. This caused warning spam when `system_instruction` was also set and didn't persist across full context updates.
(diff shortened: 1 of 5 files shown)


aconchillo (approved):

aconchillo:
LGTM!

kompfner (commented):
👍

PR #3958, opened 2026-03-08
Route Deepgram WebSocket TTS audio through audio context queue

### LINKED ISSUE #1695: Deepgram’s WebSocket TTS
### Problem Statement

I'm experiencing a 3 to 6-second delay between the user's last spoken word and the bot's first spoken response. I've looked into Deepgram's latest WebSocket TTS feature, which may help reduce latency in real-time speech generation. I'm currently using Twilio, Deepgram's TTS and STT, and the OpenAI `gpt-4o` model to generate responses. The application is deployed on an EC2 instance of type `c6i.x2large`.

### Proposed Solution

Implement Deepgram's WebSocket TTS service to reduce the delay in converting text to speech.

### Alternative Solutions

_No response_

### Additional Context

_No response_

- [ ] No, I'm just suggesting

markbackman:
## Summary

- Fixed `DeepgramTTSService` bypassing pipecat's audio context queue system, which caused stale audio frames to leak into the pipeline after interruptions
- Audio frames are now routed through `append_to_audio_context()` instead of `push_frame()`, with context availability checks to discard stale post-interruption frames
- `Flushed` responses now properly signal end-of-context by appending `TTSStoppedFrame` and calling `remove_audio_context()`
- Replaced `_handle_interruption()` override with `on_audio_context_interrupted()` hook, matching the pattern used by ElevenLabs and Cartesia
- Removed redundant `process_frame()` override that caused double-flush (base class already handles this via `on_turn_context_completed()`)
- Removed redundant `start_tts_usage_metrics()` call in `run_tts()` (base class handles aggregated usage metrics)

## Testing

- Manual testing with a Deepgram API key: verify audio plays normally, interruptions clear audio promptly, and no stale audio leaks after interruption

Fixes https://github.com/pipecat-ai/pipecat/issues/1695

### DIFF
--- a/src/pipecat/services/deepgram/tts.py
+++ b/src/pipecat/services/deepgram/tts.py
@@ -22,12 +22,10 @@ from pipecat.frames.frames import (
     EndFrame,
     ErrorFrame,
     Frame,
-    InterruptionFrame,
-    LLMFullResponseEndFrame,
     StartFrame,
     TTSAudioRawFrame,
+    TTSStoppedFrame,
 )
-from pipecat.processors.frame_processor import FrameDirection
 from pipecat.services.settings import TTSSettings, _warn_deprecated_param
 from pipecat.services.tts_service import TTSService, WebsocketTTSService
 from pipecat.utils.tracing.service_decorators import traced_tts
@@ -119,7 +117,7 @@ class DeepgramTTSService(WebsocketTTSService):
         super().__init__(
             sample_rate=sample_rate,
             pause_frame_processing=True,
-            push_stop_frames=True,
+            push_stop_frames=False,
             push_start_frame=True,
             append_trailing_space=True,
             settings=default_settings,
@@ -166,19 +164,6 @@ class DeepgramTTSService(WebsocketTTSService):
         """
         await super().cancel(frame)
         await self._disconnect()
-
-    async def process_frame(self, frame: Frame, direction: FrameDirection):
-        """Process frames with special handling for LLM response end.
-
-        Args:
-            frame: The frame to process.
-            direction: The direction of frame processing.
-        """
-        await super().process_frame(frame, direction)
-
-        # When the LLM finishes responding, flush any remaining text in Deepgram's buffer
-        if isinstance(frame, (LLMFullResponseEndFrame, EndFrame)):
-            await self.flush_audio()
 
     async def _connect(self):
         """Connect to Deepgram WebSocket and start receive task."""
@@ -276,19 +261,19 @@ class DeepgramTTSService(WebsocketTTSService):
             return self._websocket
         raise Exception("Websocket not connected")
 
-    async def _handle_interruption(self, frame: InterruptionFrame, direction: FrameDirection):
-        """Handle interruption by sending Clear message to Deepgram.
+    async def on_audio_context_interrupted(self, context_id: str):
+        """Send Clear message to Deepgram when an audio context is interrupted.
 
         The Clear message will clear Deepgram's internal text buffer and stop
         sending audio, allowing for a new response to be generated.
-        """
-        await super()._handle_interruption(frame, direction)
 
-        # Send Clear message to stop current audio generation
+        Args:
+            context_id: The ID of the audio context that was interrupted.
+        """
+        await self.stop_all_metrics()
         if self._websocket:
             try:
-                clear_msg = {"type": "Clear"}
-                await self._websocket.send(json.dumps(clear_msg))
+                await self._websocket.send(json.dumps({"type": "Clear"}))
             except Exception as e:
                 logger.error(f"{self} error sending Clear message: {e}")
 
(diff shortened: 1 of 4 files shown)


src/pipecat/services/deepgram/tts.py
```diff
@@ -167,19 +166,6 @@ async def cancel(self, frame: CancelFrame):
...
-
-        Args:
-            frame: The frame to process.
-            direction: The direction of frame processing.
-        """
-        await super().process_frame(frame, direction)
-
-        # When the LLM finishes responding, flush any remaining text in Deepgram's buffer
-        if isinstance(frame, (LLMFullResponseEndFrame, EndFrame)):
-            await self.flush_audio()
```
filipi87, line 182: Nice catch!

src/pipecat/services/deepgram/tts.py
```diff
@@ -130,6 +128,7 @@ def __init__(
         self._base_url = base_url
         self._encoding = encoding
 
+        self._current_context_id = None
```
filipi87: We should not need to save it in a new variable. 

If you need to access the current context, you should be able to invoke `self.get_active_audio_context_id()`. 

What issue were you encountering here that you had to implement it like this ?

src/pipecat/services/deepgram/tts.py
```diff
@@ -297,11 +284,12 @@ async def _receive_messages(self):
         async for message in self._get_websocket():
             if isinstance(message, bytes):
                 # Binary message contains audio data
-                await self.stop_ttfb_metrics()
-                frame = TTSAudioRawFrame(
-                    message, self.sample_rate, 1, context_id=self.get_active_audio_context_id()
```
filipi87, line 303: What problem were you seeing with `get_active_audio_context_id`?

src/pipecat/services/deepgram/tts.py
```diff
@@ -312,12 +300,18 @@ async def _receive_messages(self):
...
                         logger.trace(f"Received Flushed: {msg}")
-                        # Flushed indicates the end of audio generation for the current buffer
-                        # This happens after flush_audio() is called
+                        ctx_id = self._current_context_id
+                        if ctx_id and self.audio_context_available(ctx_id):
+                            if self._push_stop_frames:
+                                await self.append_to_audio_context(
+                                    ctx_id, TTSStoppedFrame(context_id=ctx_id)
+                                )
+                            await self.remove_audio_context(ctx_id)
```
filipi87: Deepgram currently sets the property: `push_stop_frames`, which makes the base class responsible for pushing the `TTSStoppedFrame`.

So, if we are going to start pushing the stop frames here, we should change that property to False.

And I guess all the code that we should need here is:
```
await self.append_to_audio_context(
   ctx_id, TTSStoppedFrame(context_id=context_id=self.get_active_audio_context_id())
)
await self.remove_audio_context(ctx_id)
```

src/pipecat/services/deepgram/tts.py
```diff
@@ -297,11 +284,12 @@ async def _receive_messages(self):
         async for message in self._get_websocket():
             if isinstance(message, bytes):
                 # Binary message contains audio data
-                await self.stop_ttfb_metrics()
-                frame = TTSAudioRawFrame(
-                    message, self.sample_rate, 1, context_id=self.get_active_audio_context_id()
-                )
-                await self.push_frame(frame)
```
filipi87, line 305: I believe this was probably the actual issue. 
Inside `await self.push_frame(frame)`, it should be `await self.append_to_audio_context(ctx_id, frame)`.

filipi87: But other than this, you should not need `self._current_context_id`.

markbackman:
@filipi87 I pushed a commit (2fc75a5dfccd25601b246fc9232da13f0bbc0f88) which I believe follows your guidance.

I need to fully get up to speed on the new base class methods available. I'll take a closer look next time I work on this.

src/pipecat/services/deepgram/tts.py
```diff
@@ -297,11 +282,12 @@ async def _receive_messages(self):
...
+                ctx_id = self.get_active_audio_context_id()
+                if ctx_id and self.audio_context_available(ctx_id):
+                    frame = TTSAudioRawFrame(message, self.sample_rate, 1, context_id=ctx_id)
+                    await self.append_to_audio_context(ctx_id, frame)
+                else:
+                    logger.trace(f"{self}: discarding audio for stale context {ctx_id}")
```
filipi87: Just this would be enough:

```suggestion
                ctx_id = self.get_active_audio_context_id()
                frame = TTSAudioRawFrame(message, self.sample_rate, 1, context_id=ctx_id)
                await self.append_to_audio_context(ctx_id, frame)
```

Internally, `append_to_audio_context` will already check if the context is available, whether it should be recreated if it is still part of the same turn, or whether the audio should simply be discarded.

src/pipecat/services/deepgram/tts.py
```diff
@@ -312,12 +298,16 @@ async def _receive_messages(self):
...
                     elif msg_type == "Flushed":
                         logger.trace(f"Received Flushed: {msg}")
-                        # Flushed indicates the end of audio generation for the current buffer
-                        # This happens after flush_audio() is called
+                        ctx_id = self.get_active_audio_context_id()
+                        if ctx_id and self.audio_context_available(ctx_id):
+                            await self.append_to_audio_context(
+                                ctx_id, TTSStoppedFrame(context_id=ctx_id)
+                            )
+                            await self.remove_audio_context(ctx_id)
```
filipi87: Same here, this should be enough:
```suggestion
                        ctx_id = self.get_active_audio_context_id()
                        await self.append_to_audio_context(
                            ctx_id, TTSStoppedFrame(context_id=ctx_id)
                        )
                        await self.remove_audio_context(ctx_id)
```

filipi87 (approved):
Looks good. Thank you for catching this @markbackman . 🚀

src/pipecat/services/deepgram/tts.py
markbackman: Updated

src/pipecat/services/deepgram/tts.py
markbackman: Updated

PR #3959, opened 2026-03-08
fix(livekit): remove redundant self arg in on_call_state_updated

ajmeraharsh:
## Summary
- `_on_call_state_updated` passes `(self, state)` to `_call_event_handler`, but `_run_handler` already prepends `self` when invoking the handler. This causes event handlers registered via `@transport.event_handler("on_call_state_updated")` to receive 3 positional arguments `(transport, transport, state)` instead of the expected 2 `(transport, state)`, making the event unusable.
- This aligns the behavior with `_on_first_participant_joined`, which correctly passes only `participant_id` without `self`.

## Test plan
- [x] Register an `on_call_state_updated` event handler with signature `(transport, state)` and verify it receives the correct arguments
- [x] Verify SIP call state changes (dialing → active → hangup) are correctly reported

### DIFF
--- a/changelog/3959.fixed.md
+++ b/changelog/3959.fixed.md
@@ -1,0 +1,1 @@
+- Fixed `on_call_state_updated` event handler in LiveKit transport receiving incorrect number of arguments due to redundant `self` passed to `_call_event_handler`.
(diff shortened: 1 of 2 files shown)


aconchillo (approved):

aconchillo:
@ajmeraharsh Thank you! Approved. Can you please add a changelog entry following the guidelines? :pray:

aconchillo (approved):

ajmeraharsh:
@aconchillo totally forgot the changelog my bad. Added now.

PR #3960, opened 2026-03-08
Treat conversation_already_has_active_response as non-fatal in Realtime API

sysradium:
## Summary

-  After sending a function call result via conversation.item.create, the Realtime API sometimes auto-creates a response before pipecat sends response.create, producing a conversation_already_has_active_response error
-  This error is harmless (the model is already responding), but it's currently treated as fatal — killing the WebSocket connection and ending the call
-  Fix: handle it the same way as the existing response_cancel_not_active — log at debug level and continue
-  Applied to all three Realtime API services: OpenAI, Grok, and OpenAI Realtime Beta

## How to reproduce

1.   Set up a Realtime API voice bot with function calling (tools)
2.  Trigger a tool call that takes more than a few hundred milliseconds
3.  The response.create sent by pipecat after the tool result races with the API's auto-created response
4.  Connection dies with conversation_already_has_active_response followed by a keepalive ping timeout

### DIFF
--- a/changelog/3960.fixed.md
+++ b/changelog/3960.fixed.md
@@ -1,0 +1,1 @@
+- Fixed OpenAI Realtime, OpenAI Realtime Beta, and Grok realtime services treating `conversation_already_has_active_response` as a fatal error. These services now log it as a non-fatal debug event when a response is already in progress.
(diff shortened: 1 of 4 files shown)


markbackman (approved):
LGTM. Thanks for the fix 🙇 

The repo also has a Claude Code skill for this: `/changelog`.

sysradium:
@markbackman added a changelog item.

PR #3961, opened 2026-03-09
Add enhancement_level support to `AICFilter`.

gkmngrgn:
# Summary

This PR re-introduces `enhancement_level` support in `AICFilter` with safe model compatibility behavior, and restores runtime enable/disable control through `FilterEnableFrame`.

# What changed

- Re-added optional `enhancement_level` to `AICFilter(...)`.
- Added strict init-time validation for `enhancement_level` (`0.0 <= value <= 1.0`).
- Applied enhancement level through SDK `ProcessorParameter.EnhancementLevel`.
- Restored runtime control:
  - `FilterEnableFrame(enable=False)` sets enhancement level to `0.0`.
  - `FilterEnableFrame(enable=True)` restores the configured enhancement level.
- Kept behavior safe for unsupported/fixed parameter cases:
  - catches `ParameterOutOfRangeError`
  - logs and ignores further enhancement-level application by clearing configured value.
- Added startup logging when enhancement level is unset (model default behavior used).
- Clarified internal naming by using `_is_filter_enabled`.

# Tests

Updated `tests/test_aic_filter.py` to cover:

- initialization validation for `enhancement_level` (valid/None/invalid)
- startup behavior when enhancement is configured vs not configured
- unsupported enhancement parameter handling
- runtime frame control behavior (disable -> `0.0`, enable -> restore configured level)

# Verification

**Ran:** `uv run pytest tests/test_aic_filter.py`
**Result:** 41 passed

### DIFF
--- a/src/pipecat/audio/filters/aic_filter.py
+++ b/src/pipecat/audio/filters/aic_filter.py
@@ -336,13 +336,13 @@ class AICFilter(BaseAudioFilter):
 
     def _apply_enhancement_level(self):
         """Apply enhancement_level if configured and supported by the active model."""
-        if self._enhancement_level is None or self._processor_ctx is None:
+        if self._processor_ctx is None or self._enhancement_level is None:
             return
 
-        level = float(self._enhancement_level if not self._bypass else 0.0)
-
         try:
-            self._processor_ctx.set_parameter(ProcessorParameter.EnhancementLevel, level)
+            self._processor_ctx.set_parameter(
+                ProcessorParameter.EnhancementLevel, self._enhancement_level
+            )
         except ParameterOutOfRangeError as e:
             logger.warning(f"AIC EnhancementLevel set_parameter out-of-range: {e}")
             self._enhancement_level = None
--- a/src/pipecat/audio/filters/aic_filter.py
+++ b/src/pipecat/audio/filters/aic_filter.py
@@ -412,8 +412,7 @@ class AICFilter(BaseAudioFilter):
         logger.debug(f"  Sample rate: {self._sample_rate} Hz")
         logger.debug(f"  Frames per chunk: {self._frames_per_block}")
         if self._enhancement_level is not None:
-            level = self._enhancement_level if not self._bypass else 0.0
-            logger.debug(f"  Enhancement strength: {int(level * 100)}%")
+            logger.debug(f"  Enhancement level: {self._enhancement_level}")
         else:
             logger.debug("  Enhancement level not configured; using the model's default behavior.")
         logger.debug(f"  Optimal sample rate: {self._model.get_optimal_sample_rate()} Hz")
--- a/src/pipecat/audio/filters/aic_filter.py
+++ b/src/pipecat/audio/filters/aic_filter.py
@@ -23,6 +23,7 @@ from typing import List, Optional, Tuple
 import numpy as np
 from aic_sdk import (
     Model,
+    ParameterOutOfRangeError,
     ProcessorAsync,
     ProcessorConfig,
     ProcessorParameter,
@@ -220,6 +221,7 @@ class AICFilter(BaseAudioFilter):
         model_id: Optional[str] = None,
         model_path: Optional[Path] = None,
         model_download_dir: Optional[Path] = None,
+        enhancement_level: Optional[float] = None,
     ) -> None:
         """Initialize the AIC filter.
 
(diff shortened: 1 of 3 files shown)


src/pipecat/audio/filters/aic_filter.py
```diff
@@ -232,9 +233,12 @@ def __init__(
                 model_id is ignored and no download occurs.
             model_download_dir: Directory for downloading models as a Path object.
                 Defaults to a cache directory in user's home folder.
+            enhancement_level: Optional overall enhancement strength (0.0..1.0).
+                If None, the model default is used.
```
andresovela, line 237:
```suggestion
                If None, the model's default is used.
```

src/pipecat/audio/filters/aic_filter.py
```diff
@@ -326,6 +335,27 @@ def create_vad_analyzer(
...
+        if self._enhancement_level is None or self._processor_ctx is None:
+            return
+
+        level = float(self._enhancement_level if self._is_filter_enabled else 0.0)
+
+        try:
+            self._processor_ctx.set_parameter(ProcessorParameter.EnhancementLevel, level)
+        except ParameterOutOfRangeError as e:
+            logger.warning("AIC enhancement_level was rejected as out-of-range; ignoring it.")
+            logger.debug(f"AIC EnhancementLevel set_parameter out-of-range: {e}")
```
andresovela: Is this common practice? To log the same thing twice with different log levels?

src/pipecat/audio/filters/aic_filter.py
```diff
@@ -326,6 +335,27 @@ def create_vad_analyzer(
             sensitivity=sensitivity,
         )
 
+    def _apply_enhancement_level(self):
+        """Apply enhancement_level if configured and supported by the active model."""
+        if self._enhancement_level is None or self._processor_ctx is None:
+            return
+
+        level = float(self._enhancement_level if self._is_filter_enabled else 0.0)
```
andresovela: You shouldn't change the enhancement level when the bypass is toggled. Setting the bypass parameter is enough.

src/pipecat/audio/filters/aic_filter.py
```diff
@@ -423,12 +459,12 @@ async def process_frame(self, frame: FilterControlFrame):
             None
         """
         if isinstance(frame, FilterEnableFrame):
+            self._is_filter_enabled = frame.enable
```
andresovela: This seems to track the same thing as `_bypass`. We can remove either of them.

andresovela (changes requested):
Left some comments. Note that the changes to the enhancement_level parameter only apply for aic_sdk 2.1.0. Is that already considered here? Not sure how the pipecat dependency is set up.

src/pipecat/audio/filters/aic_filter.py
```diff
@@ -374,14 +404,20 @@ async def start(self, sample_rate: int):
...
+        self._apply_bypass()
+        self._apply_enhancement_level()
 
         # Log processor information
         logger.debug(f"ai-coustics filter started:")
         logger.debug(f"  Model ID: {self._model.get_id()}")
         logger.debug(f"  Sample rate: {self._sample_rate} Hz")
         logger.debug(f"  Frames per chunk: {self._frames_per_block}")
+        if self._enhancement_level is not None:
+            level = self._enhancement_level if self._is_filter_enabled else 0.0
```
steckes: The same here, you don't need to do anything with the enhancement-level.
Btw Enhancement-level = 0.0 can even be something very different than bypass = True :)

steckes (changes requested):

src/pipecat/audio/filters/aic_filter.py
gkmngrgn: addressed -> bc11bf96731342a236bb2322190a48aa425395be
It was a rollback issue, I removed bypass first, then rollbacked it.

src/pipecat/audio/filters/aic_filter.py
gkmngrgn: These logs are also visible in UI, I thought the json-like structured data will not be seem good but seems fine. I dropped the one -> 483f6689ed97c7dcd233ddf424bcf90d411ba94f

gkmngrgn:
> Left some comments. Note that the changes to the enhancement_level parameter only apply for aic_sdk 2.1.0. Is that already considered here? Not sure how the pipecat dependency is set up.

@andresovela yes, starting from the version 0.0.105, aic min sdk version will be v2.1.0, it's considered.

src/pipecat/audio/filters/aic_filter.py
gkmngrgn: done -> 8e1c8a38e487d12272b09eedd1d8726be326f5a2

src/pipecat/audio/filters/aic_filter.py
gkmngrgn: done -> 780559dc32a771db0462cb2efe9614f32afa550f

andresovela (approved):
LGTM

markbackman:
@gkmngrgn Please let me know when this is ready for review.

There's also one requested change, so please resolve that before we can merge.

Fl1tzi (approved):

steckes (approved):

gkmngrgn:
@markbackman it's ready! thank you.

markbackman (approved):
LGTM! Thanks.

PR #3962, opened 2026-03-09
Queuing the messages received before the data channel is ready

filipi87:
## Summary                                                                                                           
- Messages sent to the data channel before it is ready are now queued instead of silently discarded, ensuring no early application messages are lost during connection setup.
- The queue is bounded (`MAX_MESSAGE_QUEUE_SIZE = 50`) to prevent unbounded memory growth if messages accumulate before the channel opens.
- A 10-second timeout (`DATA_CHANNEL_TIMEOUT_SECS`) starts when the peer connection is established. If the data channel does not open within that window, the queue is cleared and future messages fall back to "discard" mode to avoid holding stale data indefinitely.
- Added `_flush_message_queue()` to drain buffered messages as soon as the data channel opens.


aconchillo (approved):

aconchillo:
LGTM!

PR #3964, opened 2026-03-09
Add missing 55-* update-settings examples for Piper TTS, Kokoro TTS, …

kompfner:
…Whisper STT, and Whisper MLX STT

Also fix 13e-whisper-mlx.py to pass MLXModel.LARGE_V3_TURBO.value instead of the enum directly.


kompfner:
Note that there may be more missing 55-* examples. These are just ones I happened to notice.

PR #3966, opened 2026-03-09
Add missing 55-* update-settings examples for OpenPipe LLM and XTTS TTS

kompfner:


examples/foundational/55zzp-update-settings-xtts-tts.py
kompfner, line 1: Note that I actually haven't been able to test on this one, as I think running the XTTS server (per the instructions [here](https://docs.pipecat.ai/server/services/tts/xtts)) requires Linux (I'm on Mac). @aconchillo have you run XTTS before?

markbackman (approved):

PR #3967, opened 2026-03-09
Fix Azure STT SpeechConfig with private endpoint

markbackman:
## Summary
- Fix `SpeechConfig` initialization failing when `private_endpoint` is provided
- `region` and `endpoint` are mutually exclusive in the Azure Speech SDK — passing both raises `ValueError`
- Now conditionally passes either `endpoint` (private endpoint) or `region` (standard), never both

## Test plan
- [x] Verify Azure STT works without `private_endpoint` (standard `region`-based config)
- [x] Verify Azure STT works with `private_endpoint` URL (e.g., `wss://my-custom.cognitiveservices.azure.com`)


aconchillo (approved):

aconchillo:
LGTM

PR #3968, opened 2026-03-09
Add system_instruction parameter to run_inference

markbackman:
## Summary

- Added `system_instruction` parameter to `run_inference` across all LLM services (OpenAI, Anthropic, Google, AWS Bedrock, LLMSwitcher), allowing callers to override the system prompt for one-shot inference calls
- Updated `_generate_summary` to use `system_instruction` instead of embedding the summarization prompt as a system message in the context messages list
- Fixed OpenAI `run_inference` to replace (not just prepend) leading system messages when `system_instruction` is provided, preventing duplicate system messages from the service-level `system_instruction` setting

## Testing

```bash
uv run pytest tests/ -k "summar or openai" -v
```

### DIFF
--- a/src/pipecat/pipeline/llm_switcher.py
+++ b/src/pipecat/pipeline/llm_switcher.py
@@ -52,17 +52,19 @@ class LLMSwitcher(ServiceSwitcher[StrategyType]):
         """
         return self.strategy.active_service
 
-    async def run_inference(self, context: LLMContext) -> Optional[str]:
+    async def run_inference(self, context: LLMContext, **kwargs) -> Optional[str]:
         """Run a one-shot, out-of-band (i.e. out-of-pipeline) inference with the given LLM context, using the currently active LLM.
 
         Args:
             context: The LLM context containing conversation history.
+            **kwargs: Additional arguments forwarded to the active LLM's run_inference
+                (e.g. max_tokens, system_instruction).
 
         Returns:
             The LLM's response as a string, or None if no response is generated.
         """
         if self.active_llm:
-            return await self.active_llm.run_inference(context=context)
+            return await self.active_llm.run_inference(context=context, **kwargs)
         return None
 
     def register_function(
--- a/src/pipecat/services/anthropic/llm.py
+++ b/src/pipecat/services/anthropic/llm.py
@@ -345,7 +345,10 @@ class AnthropicLLMService(LLMService):
             return response
 
     async def run_inference(
-        self, context: LLMContext | OpenAILLMContext, max_tokens: Optional[int] = None
+        self,
+        context: LLMContext | OpenAILLMContext,
+        max_tokens: Optional[int] = None,
+        system_instruction: Optional[str] = None,
     ) -> Optional[str]:
         """Run a one-shot, out-of-band (i.e. out-of-pipeline) inference with the given LLM context.
 
@@ -353,6 +356,8 @@ class AnthropicLLMService(LLMService):
             context: The LLM context containing conversation history.
             max_tokens: Optional maximum number of tokens to generate. If provided,
                 overrides the service's default max_tokens setting.
+            system_instruction: Optional system instruction to use for this inference.
+                If provided, overrides any system instruction in the context.
 
         Returns:
             The LLM's response as a string, or None if no response is generated.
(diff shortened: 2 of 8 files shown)


src/pipecat/services/openai/base_llm.py
```diff
@@ -371,6 +375,15 @@ async def run_inference(
         params["stream"] = False
         params.pop("stream_options", None)
 
+        # Replace system instruction if provided (removes any service-level system message
```
markbackman: `build_chat_completion_params` always prepends the service-level `system_instruction` to messages list. When `system_instruction` is passed to `run_inference`, we need to strip that before adding the override. Without this, we end up with both the system prompt from the run_inference arg and the one from the context.

kompfner:
> Fixed OpenAI run_inference to replace (not just prepend) leading system messages when system_instruction is provided, preventing duplicate system messages from the service-level system_instruction setting

Shouldn't we follow the same decision for the one-shot `run_inference` as we made for "regular" pipeline inference, which is to simply emit a warning, rather than try to be "smart"?

src/pipecat/pipeline/llm_switcher.py
```diff
@@ -52,17 +52,19 @@ def active_llm(self) -> LLMService:
         """
         return self.strategy.active_service
 
-    async def run_inference(self, context: LLMContext) -> Optional[str]:
+    async def run_inference(self, context: LLMContext, **kwargs) -> Optional[str]:
         """Run a one-shot, out-of-band (i.e. out-of-pipeline) inference with the given LLM context, using the currently active LLM.
 
         Args:
             context: The LLM context containing conversation history.
+            **kwargs: Additional arguments forwarded to the active LLM's run_inference
```
kompfner, line 60: Wouldn't we get better auto-complete by actually matching `LLMService`'s `run_inference` signature in this method?

src/pipecat/services/anthropic/llm.py
```diff
@@ -346,14 +345,19 @@ async def _create_message_stream(self, api_call, params):
...
+        system_instruction: Optional[str] = None,
     ) -> Optional[str]:
         """Run a one-shot, out-of-band (i.e. out-of-pipeline) inference with the given LLM context.
 
         Args:
             context: The LLM context containing conversation history.
             max_tokens: Optional maximum number of tokens to generate. If provided,
                 overrides the service's default max_tokens setting.
+            system_instruction: Optional system instruction to use for this inference.
+                If provided, overrides any system instruction in the context.
```
kompfner, line 360: Per my other suggestion: should only take precedence over the system instruction in `self._settings.sytem_instruction`, not the one in the context.

kompfner: I'm actually surprised that we were never previously using `self._settings.system_instruction` for this inference, as I thought that that was the entire source of the bug that this PR was meant to address (two source of system instructions during summarization)...

...unless that was indeed the bug in other LLM services (I should read on to find out).

kompfner: > Per my other suggestion: should only take precedence over the system instruction in `self._settings.sytem_instruction`, not the one in the context.

Ah nvm! You *are* actually following the previous pattern established for each service's "regular" pipeline inference.

kompfner: > I'm actually surprised that we were never previously using self._settings.system_instruction for this inference, as I thought that that was the entire source of the bug that this PR was meant to address (two source of system instructions during summarization)...

OK, I'm guessing we only saw this bug with OpenAI. Still, making the change everywhere is the right call.

src/pipecat/services/openai/base_llm.py
kompfner: I think, even if it means `run_inference` being slightly inconsistent between services, we should maybe still match the behavior of the "regular" pipeline inference, which is:
- For OpenAI, allow service-level system instructions _alongside_ an initial context system message (because OpenAI allows it, and we don't want to guess about the user's intent). A warning should suffice if both are provided.
- For others, do the override.

src/pipecat/services/google/llm.py
```diff
@@ -908,6 +913,15 @@ async def run_inference(
             system = getattr(context, "system_message", None)
             tools = context.tools or []
 
+        # Override system instruction if provided
```
kompfner, line 916: The prioritization of service-level system instruction (in this case `system_instruction`, and in the "regular" pipeline inference case `self._settings.system_instruction`) over context-provided system message is not consistent here between Google's `run_inference` and Google's "regular" pipeline inference (in `_stream_content`).

I actually think we should update `_stream_content` to bring it in line with what the other non-OpenAI services do, which is print a warning and give `self._settings.system_instruction` precedence.

Not sure if you wanted to include that work in this PR. One of us could also take that as a follow-up item.

kompfner: @aconchillo was there a reason you didn't update Google when you worked on https://github.com/pipecat-ai/pipecat/pull/3932?

markbackman: @aconchillo will do this separately.

kompfner (approved):
👏 

We could update Flows to use this new arg, once this change is shipped.

PR #3973, opened 2026-03-10
Update Fish Audio default model from s1 to s2-pro

yuki901:
## Summary
- Update the default TTS model for `FishAudioTTSService` from `s1` to `s2-pro`, aligning with Fish Audio's latest recommended model.

## Test plan
- [x] Verify Fish Audio TTS works with default settings (no explicit model_id)
- [x] Verify explicit model_id override still works


markbackman (approved):
LGTM!

PR #3980, opened 2026-03-10
Move Google Vertex and OpenAI LLM modules to subpackages

aconchillo:
## Summary

- Moved `google/llm_vertex.py` → `google/vertex/llm.py`
- Moved `google/llm_openai.py` → `google/openai/llm.py`
- Moved `google/gemini_live/llm_vertex.py` → `google/gemini_live/vertex/llm.py`
- Old import paths still work but emit a `DeprecationWarning`
- Updated all examples, tests, and internal imports to use the new paths


markbackman (approved):
LGTM

kompfner (commented):
I like it! Not least because it'll be easier to make changes across all LLM files if they're all named llm.py :)

PR #3981, opened 2026-03-10
Sarvam LLM Integration

dhruvladia-sarvam:
## Summary

- Added a new OpenAI-compatible Sarvam LLM integration via `SarvamLLMService` in `src/pipecat/services/sarvam/llm.py`.
- Added supported model validation for:
  - `sarvam-30b`
  - `sarvam-30b-16k`
  - `sarvam-105b`
  - `sarvam-105b-32k`
- Added Sarvam-specific request handling:
  - supports `wiki_grounding` and `reasoning_effort`
  - filters unsupported request fields (`stream_options`, `max_completion_tokens`, `service_tier`)
  - validates tool constraints (`tool_choice` requires non-empty `tools`)
- Exported the new service from `src/pipecat/services/sarvam/__init__.py`.
- Added foundational example `examples/foundational/55zzq-update-settings-sarvam-llm.py`.
- Added changelog entry in `changelog/3978.added.md`.


src/pipecat/services/sarvam/llm.py
```diff
@@ -0,0 +1,235 @@
...
+    - SDK User-Agent propagation on every API call
+    - raw Sarvam server error passthrough
+    """
+
+    SUPPORTED_MODELS = frozenset({"sarvam-30b", "sarvam-30b-16k", "sarvam-105b", "sarvam-105b-32k"})
+    TOOL_CALLING_MODELS = frozenset(
+        {"sarvam-30b", "sarvam-30b-16k", "sarvam-105b", "sarvam-105b-32k"}
+    )
+
+    class InputParams(OpenAILLMService.InputParams):
```
markbackman: Check out the latest patterns in https://github.com/pipecat-ai/pipecat/blob/main/COMMUNITY_INTEGRATIONS.md#defining-your-settings-class.

The InputParams class is no longer used. You'll define a ServiceSettings class. See the other modules for guidance. Let me know if you have questions.

The guidances is:
- `__init__` args: init-time only parameters
- Settings params: params that will be updated at runtime. This can include those that are required at init time.

The base LLMServiceSettings class and OpenAILLMSettings classes already have many variables, which you'll get through inheritance.

src/pipecat/services/sarvam/llm.py
```diff
@@ -0,0 +1,235 @@
...
+        if "reasoning_effort" in extra and extra["reasoning_effort"] is not None:
+            params["reasoning_effort"] = extra["reasoning_effort"]
+
+        return params
+
+    async def get_chat_completions(self, params_from_context: OpenAILLMInvocationParams):
+        """Get streaming chat completions with Sarvam raw error passthrough."""
+        try:
+            return await super().get_chat_completions(params_from_context)
+        except (APITimeoutError, asyncio.TimeoutError, httpx.TimeoutException):
```
markbackman: This is redundant with the `BaseOpenAILLMService` class from which this class inherits:
https://github.com/pipecat-ai/pipecat/blob/main/src/pipecat/services/openai/base_llm.py#L268-L296

src/pipecat/services/sarvam/llm.py
```diff
@@ -0,0 +1,235 @@
...
+            raise
+        except Exception as e:
+            raise RuntimeError(self._format_raw_server_error(e)) from e
+
+    async def run_inference(
```
markbackman: Same comment here and for other functions about being redundant. Make sure these aren't already covered in the base class, which I believe all are. That will make your class simpler and easier to maintain.

markbackman (commented):
Thanks! See my comments about settings and the base class inheritance. You should be able to simplify this code even more. Only override the base class when you have a deviation.

src/pipecat/services/sarvam/llm.py
dhruvladia-sarvam: Fair point on structure, this method intentionally delegates to base for request execution.
The override exists only to normalize Sarvam error payloads (error.body/response.json/text) into a raw user-visible message, while re-raising timeout exceptions unchanged so timeout handling remains intact.
Without this wrapper, we lose provider-specific error extraction expected by our Sarvam tests.

src/pipecat/services/sarvam/llm.py
dhruvladia-sarvam: Agreed, please re-review.

src/pipecat/services/sarvam/llm.py
dhruvladia-sarvam: Agree this mirrors base flow. I kept it only to enforce Sarvam raw server error passthrough for out-of-band inference as well. Have refactored to reduce duplication and keep full base signature compatibility.

src/pipecat/services/sarvam/llm.py
```diff
@@ -0,0 +1,278 @@
...
+        # 2. Apply direct init arg overrides (deprecated)
+        if model is not None:
+            _warn_deprecated_param("model", SarvamLLMSettings, "model")
+            default_settings.model = model
+
+        # 3. Apply settings delta (canonical API, always wins)
+        if settings is not None:
+            default_settings.apply_update(settings)
+
+        # BaseOpenAILLMService stores settings as OpenAILLMSettings, so keep
```
markbackman: This is unnecessary. You can remove L104-107. The `SarvamLLMSettings` are always available, so no need to duplicate.

src/pipecat/services/sarvam/llm.py
```diff
@@ -0,0 +1,278 @@
...
+        request fields and applies Sarvam-specific options.
+        """
+        self._validate_tool_parameters(params_from_context)
+
+        params = super().build_chat_completion_params(params_from_context)
+        params.pop("stream_options", None)
+        params.pop("max_completion_tokens", None)
+        params.pop("service_tier", None)
+
+        extra = self._settings.extra if isinstance(self._settings.extra, dict) else {}
```
markbackman: Remove L161-165. Instead, let's just use self._settings to access the settings directly:
```
        if is_given(self._settings.wiki_grounding) and self._settings.wiki_grounding is not None:
            params["wiki_grounding"] = self._settings.wiki_grounding
        if (
            is_given(self._settings.reasoning_effort)
            and self._settings.reasoning_effort is not None
        ):
            params["reasoning_effort"] = self._settings.reasoning_effort
```

src/pipecat/services/sarvam/llm.py
```diff
@@ -0,0 +1,278 @@
...
+    - SDK User-Agent propagation on every API call
+    - raw Sarvam server error passthrough
+    """
+
+    SUPPORTED_MODELS = frozenset({"sarvam-30b", "sarvam-30b-16k", "sarvam-105b", "sarvam-105b-32k"})
+    TOOL_CALLING_MODELS = frozenset(
+        {"sarvam-30b", "sarvam-30b-16k", "sarvam-105b", "sarvam-105b-32k"}
+    )
+    Settings = SarvamLLMSettings
+    _settings: SarvamLLMSettings
```
markbackman: The convention is to use an internal `Settings` alias so that you can reference as `SarvamLLMService.Settings`. This makes discoverability easier.

```suggestion
    _settings: Settings
```

src/pipecat/services/sarvam/llm.py
```diff
@@ -0,0 +1,278 @@
...
+
+    def __init__(
+        self,
+        *,
+        api_key: str,
+        base_url: str = "https://api.sarvam.ai/v1",
+        model: Optional[str] = None,
+        settings: Optional[SarvamLLMSettings] = None,
```
markbackman:
```suggestion
        settings: Optional[Settings] = None,
```

src/pipecat/services/sarvam/llm.py
```diff
@@ -0,0 +1,278 @@
...
+        """Initialize Sarvam LLM service.
+
+        Args:
+            api_key: Sarvam API key used for both OpenAI auth and Sarvam subscription header.
+            base_url: Sarvam OpenAI-compatible base URL.
+            model: Sarvam model identifier. Supported values: ``sarvam-30b``,
+                ``sarvam-30b-16k``, ``sarvam-105b``, ``sarvam-105b-32k``.
+
+                .. deprecated:: 0.0.105
+                    Use ``settings=SarvamLLMSettings(model=...)`` instead.
```
markbackman:
```suggestion
                    Use ``settings=SarvamLLMService.Settings(model=...)`` instead.
```

src/pipecat/services/sarvam/llm.py
```diff
@@ -0,0 +1,278 @@
...
+
+            settings: Runtime-updatable settings. When provided alongside deprecated
+                parameters, ``settings`` values take precedence.
+            default_headers: Additional HTTP headers to include in requests.
+            **kwargs: Additional keyword arguments passed to ``OpenAILLMService``.
+        """
+        # 1. Initialize default_settings with hardcoded defaults
+        default_settings = SarvamLLMSettings(model="sarvam-30b")
```
markbackman:
```suggestion
        default_settings = self.Settings(model="sarvam-30b")
```

src/pipecat/services/sarvam/llm.py
```diff
@@ -0,0 +1,278 @@
...
+
+        extra = self._settings.extra if isinstance(self._settings.extra, dict) else {}
+        if "wiki_grounding" in extra and extra["wiki_grounding"] is not None:
+            params["wiki_grounding"] = extra["wiki_grounding"]
+        if "reasoning_effort" in extra and extra["reasoning_effort"] is not None:
+            params["reasoning_effort"] = extra["reasoning_effort"]
+
+        return params
+
+    async def _update_settings(self, delta: OpenAILLMSettings) -> dict[str, Any]:
```
markbackman: Remove `_update_settings()`. The base class handles this already if you use self._settings, which I'm recommending above.

src/pipecat/services/sarvam/llm.py
```diff
@@ -0,0 +1,278 @@
...
+                system_instruction=system_instruction,
+            )
+        )
+
+    def _validate_model(self, model: str):
+        if model not in self.SUPPORTED_MODELS:
+            allowed = ", ".join(sorted(self.SUPPORTED_MODELS))
+            raise ValueError(f"Unsupported Sarvam LLM model '{model}'. Allowed values: {allowed}.")
+
+    def _extract_sarvam_extra_from_settings(self, settings_obj: Any) -> dict[str, Any]:
```
markbackman: Remove `_extract_sarvam_extra_from_settings()`, as this is now unused with the recommendations above.

src/pipecat/services/sarvam/llm.py
```diff
@@ -0,0 +1,278 @@
...
+                parameters, ``settings`` values take precedence.
+            default_headers: Additional HTTP headers to include in requests.
+            **kwargs: Additional keyword arguments passed to ``OpenAILLMService``.
+        """
+        # 1. Initialize default_settings with hardcoded defaults
+        default_settings = SarvamLLMSettings(model="sarvam-30b")
+
+        # 2. Apply direct init arg overrides (deprecated)
+        if model is not None:
+            _warn_deprecated_param("model", SarvamLLMSettings, "model")
```
markbackman: No longer needed when using self.Settings

```suggestion
            _warn_deprecated_param("model", "model")
```

src/pipecat/services/sarvam/llm.py
```diff
@@ -0,0 +1,278 @@
...
+from pipecat.adapters.services.open_ai_adapter import OpenAILLMInvocationParams
+from pipecat.processors.aggregators.llm_context import LLMContext
+from pipecat.processors.aggregators.openai_llm_context import OpenAILLMContext
+from pipecat.services.openai.base_llm import OpenAILLMSettings
+from pipecat.services.openai.llm import OpenAILLMService
+from pipecat.services.sarvam._sdk import sdk_headers
+from pipecat.services.settings import NOT_GIVEN as _NOT_GIVEN
+from pipecat.services.settings import _NotGiven, _warn_deprecated_param, is_given
+
+__all__ = ["SarvamLLMService", "SarvamLLMSettings"]
```
markbackman: Not needed. These classes are already available from this module.

examples/foundational/55zzq-update-settings-sarvam-llm.py
```diff
@@ -0,0 +1,146 @@
...
+    if not value:
+        raise ValueError(f"Environment variable `{name}` is required.")
+    return value
+
+
+async def run_bot(transport: BaseTransport, runner_args: RunnerArguments):
+    logger.info("Starting bot")
+
+    stt = SarvamSTTService(
+        model="saaras:v3",
```
markbackman: Initialize services using settings, including an update to context:
```
    stt = SarvamSTTService(
        api_key=_require_env("SARVAM_API_KEY"),
        settings=SarvamSTTService.Settings(
            model="bulbul:v3",
        ),
    )

    tts = SarvamTTSService(
        api_key=_require_env("SARVAM_API_KEY"),
        settings=SarvamTTSService.Settings(
            model="bulbul:v3",
        ),
    )

    llm = SarvamLLMService(
        api_key=_require_env("SARVAM_API_KEY"),
        settings=SarvamLLMService.Settings(
            model="sarvam-30b",
            system_instruction="You are a helpful LLM in a WebRTC call. Your goal is to demonstrate your capabilities in a succinct way. Your output will be spoken aloud, so avoid special characters that can't easily be spoken, such as emojis or bullet points. Respond to what the user said in a creative and helpful way.",
        ),
    )

    context = LLMContext()
```

examples/foundational/55zzq-update-settings-sarvam-llm.py
```diff
@@ -0,0 +1,146 @@
...
+        model="bulbul:v3",
+        api_key=_require_env("SARVAM_API_KEY"),
+    )
+
+    llm = SarvamLLMService(
+        api_key=_require_env("SARVAM_API_KEY"),
+        model="sarvam-30b",
+    )
+
+    messages: list[Any] = [
```
markbackman: Remove L77-88

examples/foundational/55zzq-update-settings-sarvam-llm.py
```diff
@@ -0,0 +1,146 @@
...
+                "You are a helpful LLM in a WebRTC call. Your goal is to "
+                "demonstrate your capabilities in a succinct way. Your output "
+                "will be spoken aloud, so avoid special characters that can't "
+                "easily be spoken, such as emojis or bullet points. Respond to "
+                "what the user said in a creative and helpful way."
+            ),
+        },
+    ]
+
+    context = LLMContext(messages)
```
markbackman:
```suggestion
    context = LLMContext()
```

examples/foundational/55zzq-update-settings-sarvam-llm.py
```diff
@@ -0,0 +1,146 @@
...
+            enable_metrics=True,
+            enable_usage_metrics=True,
+        ),
+        idle_timeout_secs=runner_args.pipeline_idle_timeout_secs,
+    )
+
+    @transport.event_handler("on_client_connected")
+    async def on_client_connected(transport, client):
+        logger.info("Client connected")
+        messages.append({"role": "system", "content": "Please introduce yourself to the user."})
```
markbackman:
```suggestion
        context.add_message({"role": "user", "content": "Please introduce yourself to the user."})
```

examples/foundational/55zzq-update-settings-sarvam-llm.py
```diff
@@ -0,0 +1,146 @@
...
+        await task.queue_frames([LLMRunFrame()])
+
+        await asyncio.sleep(10)
+        logger.info("Updating Sarvam LLM settings: temperature=0.1")
+        await task.queue_frame(LLMUpdateSettingsFrame(delta=OpenAILLMSettings(temperature=0.1)))
```
markbackman:
```suggestion
        await task.queue_frame(LLMUpdateSettingsFrame(delta=SarvamLLMService.Settings(temperature=0.1)))
```

markbackman (commented):
Can you make a handful of updates to align with the latest pattern for Settings?

markbackman:
I'd recommend removing the error handling overrides. The base class already catches exceptions and calls `push_error()`, which surfaces the error message to the application. Wrapping all exceptions feels like the wrong pattern here. What do these errors provide on your end? Is there another way to handle them? I'm raising this because no other OpenAILLMService subclasses are doing this.

src/pipecat/services/sarvam/__init__.py
```diff
@@ -4,5 +4,5 @@
 # SPDX-License-Identifier: BSD 2-Clause License
 #
 
-
+from .llm import SarvamLLMService
```
markbackman: Remove the .llm export. This shouldn't be needed. Our current pattern is to not export in the module (but we may change that in the future).
```suggestion
```

src/pipecat/services/sarvam/llm.py
```diff
@@ -0,0 +1,278 @@
...
+    This service extends ``OpenAILLMService`` while adding Sarvam-specific behavior:
+
+    - model allow-list validation
+    - request shaping for Sarvam-compatible parameters
+    - Sarvam auth header wiring (``api-subscription-key``)
+    - SDK User-Agent propagation on every API call
+    - raw Sarvam server error passthrough
+    """
+
+    SUPPORTED_MODELS = frozenset({"sarvam-30b", "sarvam-30b-16k", "sarvam-105b", "sarvam-105b-32k"})
```
markbackman: Will `SUPPORTED_MODELS` and `TOOL_CALLING_MODELS` ever be different? Also, are these internal. If so, recommend changing to `_SUPPORTED_MODELS` and `_TOOL_CALLING_MODELS`.

src/pipecat/services/sarvam/llm.py
```diff
@@ -0,0 +1,278 @@
...
+        if is_given(wiki_grounding):
+            updates["wiki_grounding"] = wiki_grounding
+
+        reasoning_effort = getattr(settings_obj, "reasoning_effort", _NOT_GIVEN)
+        if is_given(reasoning_effort):
+            updates["reasoning_effort"] = reasoning_effort
+
+        return updates
+
+    def _validate_tool_parameters(self, params_from_context: OpenAILLMInvocationParams):
```
markbackman, line 148: This seems unnecessary. The OpenAI client already returns an API error if the combination is invalid.

src/pipecat/services/sarvam/llm.py
dhruvladia-sarvam: needed to preserve Sarvam-only settings because base currently stores OpenAILLMSettings

src/pipecat/services/sarvam/llm.py
dhruvladia-sarvam: Not adopted for same reason above; Sarvam-only fields must be bridged through extra right now.

src/pipecat/services/sarvam/llm.py
dhruvladia-sarvam: Already compliant (Settings = SarvamLLMSettings remains)

src/pipecat/services/sarvam/llm.py
dhruvladia-sarvam: Retained, required for runtime Sarvam field updates

src/pipecat/services/sarvam/llm.py
dhruvladia-sarvam: Retained for backward-compatible deprecated model arg support.

src/pipecat/services/sarvam/llm.py
dhruvladia-sarvam: Accepted for cleanup, removed __all__

dhruvladia-sarvam:
> I'd recommend removing the error handling overrides. The base class already catches exceptions and calls `push_error()`, which surfaces the error message to the application. Wrapping all exceptions feels like the wrong pattern here. What do these errors provide on your end? Is there another way to handle them? I'm raising this because no other OpenAILLMService subclasses are doing this.



> I'd recommend removing the error handling overrides. The base class already catches exceptions and calls `push_error()`, which surfaces the error message to the application. Wrapping all exceptions feels like the wrong pattern here. What do these errors provide on your end? Is there another way to handle them? I'm raising this because no other OpenAILLMService subclasses are doing this.

Retained, wrappers normalize raw Sarvam server payloads for direct helper usage (run_inference / get_chat_completions), not just frame pipeline paths

src/pipecat/services/sarvam/llm.py
dhruvladia-sarvam: Accepted and implemented

src/pipecat/services/sarvam/llm.py
dhruvladia-sarvam: Retained, early deterministic validation/error messaging before network call remains intentional

markbackman:
@dhruvladia-sarvam I've opened https://github.com/pipecat-ai/pipecat/pull/4081 which contains the required changes. I also have a final question about the error override.

Note: the required changes are to correctly use the ServiceSettings API. I've validated that these work correctly. Let's continue discussing in the PR I opened.

src/pipecat/services/sarvam/llm.py
```diff
@@ -0,0 +1,209 @@
...
+
+    This service extends ``OpenAILLMService`` while adding Sarvam-specific behavior:
+
+    - model allow-list validation
+    - request shaping for Sarvam-compatible parameters
+    - Sarvam auth header wiring (``api-subscription-key``)
+    - SDK User-Agent propagation on every API call
+    """
+
+    _SUPPORTED_MODELS = frozenset(
```
markbackman, line 45: This is no longer used. You can remove.

src/pipecat/services/sarvam/llm.py
```diff
@@ -0,0 +1,209 @@
...
+
+        Args:
+            api_key: Sarvam API key used for both OpenAI auth and Sarvam subscription header.
+            base_url: Sarvam OpenAI-compatible base URL.
+            settings: Runtime-updatable settings.
+            default_headers: Additional HTTP headers to include in requests.
+            **kwargs: Additional keyword arguments passed to ``OpenAILLMService``.
+        """
+        # Initialize defaults with concrete values for Sarvam-specific fields.
+        default_settings = self.Settings(
```
markbackman, line 71: You don't need to initialize the full set. The full set is set through inheritance. So, just use:
```
            model="sarvam-30b",
            wiki_grounding=None,
            reasoning_effort=None,
        )
```

src/pipecat/services/sarvam/llm.py
```diff
@@ -0,0 +1,209 @@
...
+            default_settings.apply_update(settings)
+
+        super().__init__(
+            api_key=api_key,
+            base_url=base_url,
+            settings=default_settings,
+            default_headers=default_headers,
+            **kwargs,
+        )
+        # Keep Sarvam-specific settings object so runtime updates include
```
markbackman: Remove L103-105. This is not needed. When you call super().__init__(settings=default_settings), you're passing settings up the inheritance change so that self._settings now contains the values you specified above. This base class that uses this is AIServices.

src/pipecat/services/sarvam/llm.py
```diff
@@ -0,0 +1,209 @@
...
+            params["wiki_grounding"] = self._settings.wiki_grounding
+        if (
+            is_given(self._settings.reasoning_effort)
+            and self._settings.reasoning_effort is not None
+        ):
+            params["reasoning_effort"] = self._settings.reasoning_effort
+
+        return params
+
+    async def _call_with_raw_sarvam_errors(self, awaitable: Awaitable[_T]) -> _T:
```
markbackman: I'm confused. I thought we agreed to remove the overrides?

Remove `_call_with_raw_sarvam_errors`, `get_chat_completions`, and `run_inference`.

markbackman (commented):
I thought we agreed to drop error overrides: https://github.com/pipecat-ai/pipecat/pull/4081#issuecomment-4109803628.

Also, you removed `_validate_model` which could be useful, though not required. Maybe your API returns useful error messages if the wrong model string is provided. Anyway, what I shared in #4081 is mergeable. You can just copy changes from there to here. Please don't overlook the examples, README, evals, etc files.

Or, I can just merge the PR I posted. Either way, let's get this cleaned up so we can merge it.

src/pipecat/services/sarvam/llm.py
dhruvladia-sarvam: Yeah I was in process of pushing that particular change, have pushed with the new ones

markbackman (approved):
LGTM. I pushed a few additional changes to these PRs to fix some issues with the examples, added the missing 14y example, removed unnecessary settings dictionary things.

I tested the examples, including using the settings to make sure all is well ✅

PR #3984, opened 2026-03-11
Update onnxruntime to 1.24.3

aconchillo:
## Summary

- Update onnxruntime from 1.23.2 to 1.24.3
- This version adds support for Python 3.14

### DIFF
--- a/uv.lock
+++ b/uv.lock
@@ -1,12 +1,11 @@
 version = 1
 revision = 3
-requires-python = ">=3.10"
+requires-python = ">=3.11"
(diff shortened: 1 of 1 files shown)


aconchillo:
> This drops 3.10 support. If we do that, it should be for the 1.0.0 release.
> 
> Are we ready for that?

There's a still considerable amount of downloads for 3.10: https://pypistats.org/packages/pipecat-ai

But they will need to update anyways in a few months. Also, they can still use Pipecat < 1.0 and update when they are ready.

PR #3991, opened 2026-03-11
Fix out of date docstrings

kompfner:
...plus a bonus change, which is moving `OpenAIRealtimeSTTService`'s `noise_reduction` into its settings (I figured it could be useful to update this at runtime).

### DIFF
--- a/src/pipecat/services/openai/stt.py
+++ b/src/pipecat/services/openai/stt.py
@@ -187,9 +187,15 @@ class OpenAIRealtimeSTTSettings(STTSettings):
 
     Parameters:
         prompt: Optional prompt text to guide transcription style.
+        noise_reduction: Noise reduction mode. ``"near_field"`` for close
+            microphones, ``"far_field"`` for distant microphones, or ``None``
+            to disable.
     """
 
     prompt: str | None | _NotGiven = field(default_factory=lambda: NOT_GIVEN)
+    noise_reduction: Literal["near_field", "far_field"] | None | _NotGiven = field(
+        default_factory=lambda: NOT_GIVEN
+    )
 
 
 class OpenAIRealtimeSTTService(WebsocketSTTService):
@@ -220,8 +226,10 @@ class OpenAIRealtimeSTTService(WebsocketSTTService):
 
         stt = OpenAIRealtimeSTTService(
             api_key="sk-...",
-            model="gpt-4o-transcribe",
-            noise_reduction="near_field",
+            settings=OpenAIRealtimeSTTService.Settings(
+                model="gpt-4o-transcribe",
+                noise_reduction="near_field",
+            ),
         )
     """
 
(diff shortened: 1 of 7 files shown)


src/pipecat/services/openai/stt.py
```diff
@@ -274,6 +282,9 @@ def __init__(
             noise_reduction: Noise reduction mode. ``"near_field"`` for
                 close microphones, ``"far_field"`` for distant
                 microphones, or ``None`` to disable.
+
+                .. deprecated:: 0.0.106
```
kompfner, line 286: Note: we already have a TODO in our 1.0 wishlist to fix any mentions of a version which will turn out not to exist.

src/pipecat/services/openai/stt.py
```diff
@@ -220,8 +226,10 @@ class OpenAIRealtimeSTTService(WebsocketSTTService):
 
         stt = OpenAIRealtimeSTTService(
             api_key="sk-...",
-            model="gpt-4o-transcribe",
-            noise_reduction="near_field",
+            settings=OpenAIRealtimeSTTSettings(
```
markbackman: We're mixing patterns with the standalone class and the internal class. We should pick one pattern to use.

src/pipecat/services/openai/stt.py
markbackman, line 286: I had assumed this would be fixed for the session, but that was just an opinion.

markbackman (commented):
Just a few comments.

src/pipecat/services/openai/stt.py
kompfner: Wait, I thought this PR had all of them with the internal class...an oversight

src/pipecat/services/openai/stt.py
kompfner: Yup, just a judgment call. It struck me as something that someone might reasonably want to toggle. Changing it at runtime shouldn't have any effect on the rest of the pipeline, AFAICT.

src/pipecat/services/openai/stt.py
markbackman: Oh, I mean in general across the entire codebase. I left them all to use the external class only because that's how it was originally written.

In fact, the `settings` arg in `__init__` still references the external class. So that might be a clean up task, if we're interested in it.

src/pipecat/services/openai/stt.py
markbackman: Makes sense.

src/pipecat/services/openai/stt.py
```diff
@@ -274,6 +282,9 @@ def __init__(
             noise_reduction: Noise reduction mode. ``"near_field"`` for
                 close microphones, ``"far_field"`` for distant
                 microphones, or ``None`` to disable.
+
+                .. deprecated:: 0.0.106
+                    Use ``settings=OpenAIRealtimeSTTSettings(noise_reduction=...)`` instead.
```
kompfner, line 287: Hmm...looking at this line, I wonder if we should go through all of our deprecated warnings that point people to using `settings` and change it to point people to use the nested alias to the settings...

kompfner: Will do that in a follow-up PR, all in one fell swoop

kompfner: Ah just saw that your other comment on this. Sounds like we're in agreement. (Ah yes, and the `settings` arg type itself...gotcha)

markbackman (approved):

PR #3995, opened 2026-03-11
LemonSlice transport updates - new events, extra params

jp-lemon:
Fixing the LemonSlice example, since the current voice ID is not accessible to all accounts.
Adding the ability to specify a URL for the LemonSlice API.
Adding the ability to pass arbitrary named variables to the LemonSlice API endpoint.
Adding new `on_avatar_connected` and `on_avatar_disconnected` events


examples/foundational/56-lemonslice-transport.py
```diff
@@ -114,6 +114,14 @@ async def on_client_disconnected(transport, participant):
             logger.info("Client disconnected")
             await task.cancel()
 
+        @transport.event_handler("on_avatar_connected")
+        async def on_avatar_connected(transport, participant):
+            logger.info(f"Avatar connected")
```
markbackman:
```suggestion
            logger.info("Avatar connected")
```

examples/foundational/56-lemonslice-transport.py
```diff
@@ -114,6 +114,14 @@ async def on_client_disconnected(transport, participant):
...
+
+        @transport.event_handler("on_avatar_disconnected")
+        async def on_avatar_disconnected(transport, participant):
+            logger.info(f"Avatar disconnected")
```
markbackman:
```suggestion
            logger.info("Avatar disconnected")
```

src/pipecat/transports/lemonslice/api.py
```diff
@@ -64,16 +66,14 @@ async def create_session(
             ValueError: If neither agent_id nor agent_image_url is provided.
         """
         if not agent_id and not agent_image_url:
-            # Fallback to a default agent if none is provided
-            logger.debug("No agent_id or agent_image_url provided, using default agent")
-            agent_id = "agent_080308d8b6e99f47"
+            raise ValueError("Provide an agent_id or agent_image_url")
```
markbackman, line 71: Do we no longer need the fallback agent_id?

src/pipecat/transports/lemonslice/transport.py
```diff
@@ -721,11 +731,15 @@ def __init__(
...
         self._register_event_handler("on_client_connected")
         self._register_event_handler("on_client_disconnected")
+        self._register_event_handler("on_avatar_connected")
+        self._register_event_handler("on_avatar_disconnected")
 
     async def _on_participant_left(self, participant, reason):
         """Handle participant left events."""
         ls_bot_name = await self._client.get_bot_name()
-        if participant.get("info", {}).get("userName", "") != ls_bot_name:
+        if participant.get("info", {}).get("userName", "") == ls_bot_name:
```
markbackman, line 741: Nice catch!

Should we forward the `reason` parameter to the event handler?

markbackman (commented):
Just a few comments. I'll have Claude post a review too.

src/pipecat/transports/lemonslice/transport.py
```diff
@@ -668,6 +676,8 @@ class LemonSliceTransport(BaseTransport):
 
     - on_client_connected(transport, participant): Participant connected to the session
     - on_client_disconnected(transport, participant): Participant disconnected from the session
+    - on_avatar_connected(transport, participant): LemonSlice avatar connected to the session
+    - on_avatar_disconnected(transport, participant, reason): LemonSlice avatar disconnected from the session
```
markbackman, line 681: The docstring documents `on_avatar_disconnected(transport, participant, reason)` with a `reason` parameter, but the [implementation](https://github.com/pipecat-ai/pipecat/blob/b442b99902477ecea41ae49a1f0d9d32f76e426e/src/pipecat/transports/lemonslice/transport.py#L802-L806) only passes `(transport, participant)`:

```python
async def _on_avatar_disconnected(self, participant: Any):
    await self._call_event_handler("on_avatar_disconnected", participant)
```

The `reason` from [`_on_participant_left`](https://github.com/pipecat-ai/pipecat/blob/b442b99902477ecea41ae49a1f0d9d32f76e426e/src/pipecat/transports/lemonslice/transport.py#L737-L743) is silently dropped. Users who write handlers matching the documented signature will get a `TypeError` at runtime. Either forward `reason` through the call chain or remove it from the docstring.

src/pipecat/transports/lemonslice/api.py
```diff
@@ -82,16 +82,17 @@ async def create_session(
...
             properties_dict["daily_url"] = daily_room_url
         if daily_token is not None:
             properties_dict["daily_token"] = daily_token
         if properties_dict:
             payload["properties"] = properties_dict
-        async with self._session.post(
-            self.LEMONSLICE_URL, headers=self._headers, json=payload
-        ) as r:
+        if properties:
+            payload.update(properties)
```
markbackman: `payload.update(properties)` merges `lemonslice_properties` (and `model_extra`) at the **top level** of the API payload. Previously, values from `lemonslice_properties` were [nested inside `payload["properties"]`](https://github.com/pipecat-ai/pipecat/blob/6eb988b720f7b2b7e8b8445f2e2e8a52c3fcce16/src/pipecat/transports/lemonslice/api.py#L84-L91). This is a behavioral change — keys from `lemonslice_properties` that were previously sent under `"properties"` are now top-level payload keys.

Additionally, if `properties` happens to contain a key like `"transport_type"` or `"agent_id"`, it would silently overwrite the values set above.

Consider separating the handling: keep `lemonslice_properties` nested under `payload["properties"]` (the old behavior), and only apply `model_extra` at the top level for the new arbitrary named parameters feature.

src/pipecat/transports/lemonslice/api.py
jp-lemon: We currently don't have the concept of a public agent_id that anyone can use. The api key must be linked to the owner of the agent being used to give people control over who is using their agents.
I think this was added incorrectly during the initial PR process.

src/pipecat/transports/lemonslice/api.py
jp-lemon: Separated them out here: 5e83ce6ecd9a814ea5398e104b3f1d46e2f82350

I updated arg names to  designat the old lemonslice_properties as connection properties as that is what we use it for while model_extra as top level extra props.

src/pipecat/transports/lemonslice/transport.py
jp-lemon: passed through `reason` to both `_left` callbacks: 5e83ce6ecd9a814ea5398e104b3f1d46e2f82350

examples/foundational/56-lemonslice-transport.py
jp-lemon: We are passing through reason now so logging it in this string

examples/foundational/56-lemonslice-transport.py
jp-lemon: We are passing through reason now so logging it in this string

examples/foundational/56-lemonslice-transport.py
```diff
@@ -110,10 +110,18 @@ async def on_client_connected(transport, participant):
             await task.queue_frames([LLMRunFrame()])
 
         @transport.event_handler("on_client_disconnected")
-        async def on_client_disconnected(transport, participant):
-            logger.info("Client disconnected")
+        async def on_client_disconnected(transport, participant, reason):
+            logger.info(f"Client disconnected. Reason: {reason}")
```
filipi87, line 114: We use the same signature for `on_client_connected` and `on_client_disconnected` across all transports. So, in theory, users can keep the same Pipeline and simply change the transport.

With that in mind, we shouldn’t include a reason in this case.

filipi87:
I’ve pushed this [PR](https://github.com/pipecat-ai/pipecat/pull/4268), rebased on the latest main, and restored `on_client_disconnected` for consistency across transports.

@jp-lemon, I’ve kept you as the author of the commit.

I’m doing this because we intend to include these fixes in today’s release.

Closing this PR in favor of #4268.

PR #3996, opened 2026-03-11
Prefer nested settings alias

kompfner:
Based on our discussion in https://github.com/pipecat-ai/pipecat/pull/3991.

One task remains: updating `_warn_deprecated_param` to output our latest recommended pattern.


markbackman (approved):
Nice clean up! 🧹

PR #4003, opened 2026-03-11
Fix deprecated vad_analyzer usage in examples

kompfner:
## Summary
- Migrated example files from the deprecated `TransportParams.vad_analyzer` parameter to the modern approach (skipping examples that use deprecated services)
- Broadened the deprecation warning to mention `VADProcessor` as an alternative when no `LLMUserAggregator` is needed

## Test plan
- [x] All 1018 tests pass
- [x] `ruff check` and `ruff format` pass on all modified files
- [x] Verified no deprecated `vad_analyzer` transport param usage remains in non-deprecated examples


markbackman (approved):
Nice catches! 🕵️

PR #4004, opened 2026-03-12
Add optional `service` field to `ServiceUpdateSettingsFrame` for targ…

kompfner:
…eting a specific service instance

When `service` is set and doesn't match, the service forwards the frame instead of consuming it. This allows targeting a specific service when multiple services of the same type exist in the pipeline.

### DIFF
--- a/src/pipecat/services/openai_realtime_beta/openai.py
+++ b/src/pipecat/services/openai_realtime_beta/openai.py
@@ -386,11 +386,7 @@ class OpenAIRealtimeBetaLLMService(LLMService):
         # fields, not our Settings fields, so we construct SessionProperties
         # directly. The frame.delta path falls through to super, which calls
         # _update_settings → our override handles the rest.
-        if (
-            isinstance(frame, LLMUpdateSettingsFrame)
-            and frame.delta is None
-            and (frame.service is None or frame.service is self)
-        ):
+        if isinstance(frame, LLMUpdateSettingsFrame) and frame.delta is None:
             self._session_properties = events.SessionProperties(**frame.settings)
             await self._send_session_update()
             await self.push_frame(frame, direction)


src/pipecat/services/openai_realtime_beta/openai.py
```diff
@@ -386,7 +386,11 @@ async def process_frame(self, frame: Frame, direction: FrameDirection):
         # fields, not our Settings fields, so we construct SessionProperties
         # directly. The frame.delta path falls through to super, which calls
         # _update_settings → our override handles the rest.
-        if isinstance(frame, LLMUpdateSettingsFrame) and frame.delta is None:
+        if (
```
kompfner: not sure we even strictly need to update this file, as the service is deprecated...we could just have it not support the new `service` targeting field...

src/pipecat/frames/frames.py
```diff
@@ -2154,10 +2154,15 @@ class ServiceUpdateSettingsFrame(ControlFrame, UninterruptibleFrame):
...
             object describing the fields to change.
+
+        service: Optional target service instance. When provided, only that
+            service will apply the settings; other services will forward the
+            frame unchanged.
     """
 
     settings: Mapping[str, Any] = field(default_factory=dict)
     delta: Optional["ServiceSettings"] = None
+    service: Optional["FrameProcessor"] = None
```
kompfner, line 2165: This follows the pattern of how we do service-specific targeting in `ServiceSwitcherRequestMetadataFrame` and `ManuallySwitchServiceFrame`.

src/pipecat/services/openai_realtime_beta/openai.py
kompfner: actually tempted to do that, just to avoid having to test this stuff...what do y'all think?

markbackman: Yes, I would not update this since it's deprecated. OpenAI is shutting it down soon too: May 7, 2026.

PR #4006, opened 2026-03-12
handle EndTaskFrame, StopTaskFrame and CancelTaskFrame downstream

aconchillo:
## Summary

- Fixed a race condition where `EndTaskFrame` (as a `SystemFrame`) could overtake in-flight frames and shut down the pipeline before pending work completed — e.g. an LLM function call response racing against `EndFrame`
- `EndTaskFrame` and `StopTaskFrame` are now `ControlFrame`s, so they flow through the pipeline behind pending frames, ensuring a full flush before shutdown
- Added `TaskSystemFrame(SystemFrame)` base for `CancelTaskFrame` and `InterruptionTaskFrame`, which still need immediate propagation
- The pipeline sink now catches `EndTaskFrame` and `StopTaskFrame` downstream and re-queues them upstream to the task source

## Breaking Changes

- `TaskFrame` is now a `ControlFrame` instead of `SystemFrame`. Custom code subclassing `TaskFrame` will see changed frame propagation behavior.
- `CancelTaskFrame` and `InterruptionTaskFrame` now extend `TaskSystemFrame` instead of `TaskFrame`.


markbackman (approved):
LGTM. Works cleanly when pushing EndTaskFrame upstream or downstream.

PR #4009, opened 2026-03-12
Add PerplexityLLMAdapter for message ordering strictness

kompfner:
## Summary

- Added `PerplexityLLMAdapter` that transforms conversation messages to satisfy Perplexity's stricter API constraints before sending requests. Perplexity requires strict role alternation (user/tool ↔ assistant), no non-initial system messages, and the last message to be user/tool — conversation histories that work fine with OpenAI can cause errors with Perplexity (`PerplexityLLMService` subclasses `OpenAILLMService`).
- The adapter converts non-initial system messages to user, merges consecutive same-role messages (preserving consecutive initial system messages, which Perplexity allows), and removes trailing assistant messages.
- Initial system messages are intentionally never converted to "user" — Perplexity appears to have statefulness within a conversation, so a message that was "user" in one call but becomes "system" in the next causes errors.
- `PerplexityLLMService` now uses this adapter via `adapter_class = PerplexityLLMAdapter`.
- Added missing dual-system-instruction warnings to Cerebras, Fireworks, Mistral, Perplexity, and SambaNova services, matching the existing `BaseOpenAILLMService` behavior.

## Testing

```bash
uv run pytest tests/test_get_llm_invocation_params.py -v
```

11 test cases cover: standard passthrough, initial system preservation, multiple initial system preservation, consecutive same-role merging, non-initial system conversion, trailing assistant removal, system-only context preserved, system exposed after trailing assistant removal, consecutive assistant merge+removal, tool message preservation, and empty messages.


aconchillo (approved):

aconchillo:
LGTM!

PR #4011, opened 2026-03-12
Fix for compatibility with `transformers>=5.0.0`

stefanwebb:
This PR makes the custom model `_Wav2Vec2ForEndpointing` that's part of `SmartTurn v2` work with Transformers versions >= 5.0.0.

That version introduced a breaking change that requires a `self.post_init()` at the end of `self.__init__()`.

See this issue for more discussions: https://github.com/huggingface/transformers/issues/42832


markbackman:
Thanks for the fix\! However, `LocalSmartTurnAnalyzerV2` is deprecated and will be removed in the upcoming major version release, so we're not going to be making further changes to it. We appreciate you taking the time to investigate the `transformers` 5.0 breaking change though.

PR #4013, opened 2026-03-12
[inworld] Pre-open WebSocket TTS context on LLM response start

ianbbqzy:
## Summary
- Pre-opens the Inworld TTS context on the server when `LLMFullResponseStartFrame` arrives, overlapping server-side context creation with sentence aggregation time
- Makes `_send_context` idempotent so `run_tts` skips the redundant create if the context was already prewarmed
- Cleans up unused prewarmed contexts on new turns and disconnect to avoid orphaned contexts

## Motivation
Without prewarming, context creation happens inside `run_tts` — after sentence aggregation completes — putting the server-side setup time on the TTFB critical path. By sending the context config as soon as the LLM turn starts, the server can allocate resources in parallel with text aggregation.

## Testing performed
- [x] Pipecat WebSocket TTFB benchmark passes shows a 30ms improvement for 50pct latency
- [x] Pipecat quickstart voice bot works e2e

### DIFF
--- a/src/pipecat/services/inworld/tts.py
+++ b/src/pipecat/services/inworld/tts.py
@@ -56,7 +56,6 @@ from pipecat.frames.frames import (
     ErrorFrame,
     Frame,
     InterruptionFrame,
-    LLMFullResponseStartFrame,
     StartFrame,
     TTSAudioRawFrame,
     TTSStartedFrame,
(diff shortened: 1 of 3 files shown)


src/pipecat/services/inworld/tts.py
```diff
@@ -726,6 +732,29 @@ async def push_frame(self, frame: Frame, direction: FrameDirection = FrameDirect
...
+                try:
+                    await self._send_close_context(self._prewarmed_context_id)
+                except Exception as e:
+                    logger.warning(f"{self}: Failed to close previous prewarmed context: {e}")
+                self._prewarmed_context_id = None
+            try:
+                await self._send_context(self._turn_context_id)
+                self._prewarmed_context_id = self._turn_context_id
+            except Exception as e:
+                logger.warning(f"{self}: Failed to pre-open context: {e}")
```
filipi87: Instead of doing this, I believe we should define `on_turn_context_created` inside `tts_service.py` so it can be overridden here, allowing the context to be created at that point.

Also, I don’t think you should close any context at this stage. For example, if you have an LLM response with a function call and you decide to send a `TTSSpeakFrame` when it starts, you might briefly have two contexts with different context IDs, and that should be fine.

src/pipecat/services/inworld/tts.py
```diff
@@ -1001,9 +1018,16 @@ async def _keepalive_task_handler(self):
     async def _send_context(self, context_id: str):
```
filipi87, line 1018: In theory, we shouldn't need to invoke `_send_context` inside `run_tts`, but since you're already checking here whether it has been created, this is fine.

filipi87 (approved):
LGTM. Thank you for doing this @ianbbqzy. 🚀

PR #3897, opened 2026-03-02
Add Smallest AI STT and TTS service integrations

harshitajain165:
## Summary

- Add Smallest AI speech-to-text and text-to-speech service integrations with both HTTP and WebSocket variants:
  - **SmallestSTTService** — HTTP-based segmented STT using the Pulse model
  - **SmallestRealtimeSTTService** — WebSocket-based real-time streaming STT with interim/final transcripts
  - **SmallestTTSService** — WebSocket-based streaming TTS with interruption support
  - **SmallestHttpTTSService** — HTTP-based TTS
- Add `smallest` optional dependency group in `pyproject.toml` (`httpx`, `numpy`, `soundfile`, `websockets`)
- Add `SMALLEST_TTFS_P99` latency constant in `stt_latency.py`
- Add two foundational example scripts demonstrating full conversational bot pipelines using Smallest AI services

## Changes

| File | Description |
|------|-------------|
| `src/pipecat/services/smallest/__init__.py` | Package init with `DeprecatedModuleProxy` |
| `src/pipecat/services/smallest/stt.py` | HTTP (`SmallestSTTService`) and WebSocket (`SmallestRealtimeSTTService`) STT implementations |
| `src/pipecat/services/smallest/tts.py` | WebSocket (`SmallestTTSService`) and HTTP (`SmallestHttpTTSService`) TTS implementations |
| `src/pipecat/services/stt_latency.py` | Added `SMALLEST_TTFS_P99` constant |
| `pyproject.toml` | Added `smallest` extras group |
| `uv.lock` | Auto-updated lockfile |
| `examples/foundational/07zl-interruptible-smallest.py` | Example bot using WebSocket STT + TTS |
| `examples/foundational/07zl-interruptible-smallest-http.py` | Example bot using HTTP STT + TTS |

## Test plan

- [x] Verified all four services against the live Smallest AI API (HTTP STT, Realtime STT, HTTP TTS, WebSocket TTS)
- [x] Confirmed no linting errors (`ruff check` and `ruff format`)
- [x] Structure follows existing service integration patterns (Deepgram, ElevenLabs, Sarvam, etc.)

PR #4014, opened 2026-03-13
Add Smallest AI STT and TTS service integrations

markbackman:
## Summary

Adds Smallest AI WebSocket-based STT and TTS services, aligned with current Pipecat conventions.

Based on and supersedes #3897 by @harshitajain165. Key changes from the original PR:

- **WebSocket-only**: Removed HTTP variants in favor of WebSocket services for real-time use
- **Settings pattern**: Replaced InputParams with SmallestSTTSettings/SmallestTTSSettings dataclasses
- **Init-only vs runtime-updatable**: encoding is init-only; voice/language/speed/etc. are runtime-updatable
- **_update_settings overrides**: STT reconnects on any change; TTS reconnects only on model change
- **TTS keepalive**: Added flush keepalive every 30s to prevent 60s idle timeout
- **Removed redundant overrides**: _handle_interruption and duplicate TTFB metrics
- **TTS settings default to None**: API applies its own defaults for speed/consistency/similarity/enhancement
- **Full STT language support**: All 32 documented languages with resolve_language fallback
- **Model enums for defaults**: SmallestSTTModel.PULSE and SmallestTTSModel.LIGHTNING_V3_1

## Test plan
- [ ] Verify STT transcription with Smallest API key
- [ ] Verify TTS synthesis with Smallest API key
- [ ] Confirm keepalive prevents idle timeout on both services
- [ ] Test runtime settings update triggers reconnect appropriately


markbackman:
@harshitajain165 I started from your branch and focused on the services best suited for real-time. That is, the websocket services.

The TTS service is working well and would be ready for production use.

I'm struggling with the STT service—I see that the transcriptions are usually missing words and I get hallucinations that appear ~3 seconds after the final transcript arrives. Have you seen this with other implementations and do you have any other recommendations? As it stands, the STT service is no fit for production, so I'd like to iron out the issues before merging this.

harshitajain165:
@markbackman Thanks for the detailed feedback and for refactoring the integration!

Regarding the STT issues:

- Explicit language vs "multi"

If you're testing with "multi" (auto-detect), that could be a contributing factor. The auto-detect path runs language identification on every utterance before transcribing, which both adds latency and can occasionally misroute to the wrong language model, leading to dropped words or garbled output. For single-language use cases, an explicit language code bypasses that entirely and goes straight to the language-specific model.

Worth trying if you haven't already:

```
stt = SmallestSTTService(
    api_key="...",
    settings=SmallestSTTSettings(language=Language.EN),
)
```

Happy to be wrong on this though: just thought it might be a quick thing to rule out!

- Reproducing the hallucinations

The ~3 second delay you're describing sounds like it could be the model processing a trailing audio buffer after the finalize signal, possibly silence or low-level mic noise being interpreted as speech. I'm going to run some tests on our end with detailed logging around the finalize flow to see if I can reproduce this and identify whether it's the model generating from noise or a buffering issue on the WebSocket side.

I'll follow up here

harshitajain165:
Hey @markbackman 

Given that the TTS service is production-ready and the STT issues are still being investigated, would it make sense to split this into two PRs - one for TTS only (ready to merge now) and a follow-up for STT once the hallucination/finalize behaviour is resolved? Happy to do the legwork on the split if that helps move TTS forward sooner.

markbackman:
Yes, great idea!

TTS was working well, so we could merge that sooner, getting it into the upcoming release. Please let me know once you're ready for a review.

harshitajain165:
[@markbackman](https://github.com/markbackman) TTS-only PR is up for review: #4092 

Took your refactor from this PR as-is - no changes to tts.py, settings pattern, keepalive, or any of the conventions you established. The only delta is stt.py removed and DeepgramSTTService used as a stand-in in the example while the STT issue is being investigated.

Happy to address any feedback there!

markbackman:
Closing in favor of https://github.com/pipecat-ai/pipecat/pull/4092, which adds `SmallestTTSService`.

PR #4015, opened 2026-03-13
feat(nova-sonic): add proactive session continuation for conversation…

prettyprettyprettygood:
…s >8min

Nova Sonic sessions have an AWS-imposed ~8-minute time limit. This adds transparent session continuation that rotates sessions in the background before the limit is reached, preserving conversation context with no user-perceptible interruption.

Implementation follows the AWS reference architecture:
- Monitor loop detects when session age exceeds threshold
- On assistant AUDIO contentStart: start buffering user audio, create next session (sessionStart + promptStart + system instruction)
- Track SPECULATIVE/FINAL text counts as completion signal
- On completion signal: send conversation history + audioInputStart + buffered audio to next session, then promote immediately
- Close old session in background (non-blocking)
- Dead session detection: recreate next session if idle >30s

Key design decisions:
- Session continuation enabled by default (fundamental for long conversations)
- Conversation history tracked in real-time via _sc_conversation_history (independent of pipeline context aggregator which updates asynchronously)
- Completion signal check in _handle_content_end_event (after history update) to ensure latest text is included in handoff
- Rolling audio buffer (default 3s) captures user audio during transition
- transition_threshold_seconds capped at 420s (7min) for safety margin
- Unified event methods (_send_text_event, _send_client_event, etc.) accept optional stream/prompt_name params, eliminating duplicate SC methods

Also adds:
- SessionContinuationParams config (enabled, threshold, buffer, timeout)

Nova Sonic sessions have an AWS-imposed ~8-minute time limit. This adds transparent session continuation that rotates sessions in the background before the limit is reached, preserving conversation context with no user-perceptible interruption.

 ## Summary

Nova Sonic sessions have an AWS-imposed ~8-minute time limit. This adds transparent
session continuation that rotates sessions in the background before the limit is
reached — preserving conversation context with no user-perceptible interruption.

Enabled by default. Users can opt out via `SessionContinuationParams(enabled=False)`.

##### Reference

Implementation follows the [AWS reference architecture](https://github.com/aws-samples/amazon-nova-samples/tree/main/speech-to-speech/amazon-nova-2-sonic/repeatable-patterns/session-continuation/console-python).

   ## Test plan
   - [X] Run `40a-aws-nova-sonic-session-continuation.py` with `transition_threshold_seconds=20` and verify seamless session rotation
   - [X] Verify conversation context is preserved across transitions (ask follow-up questions referencing earlier context)
   - [X] Test barge-in during transition (interrupt assistant while session is rotating)
   - [X] Test tool calling across session boundaries
   - [X] Verify `SessionContinuationParams(enabled=False)` disables session continuation
   - [X] Run existing `40-aws-nova-sonic.py` example to verify no regression

### DIFF
--- a/src/pipecat/services/aws/nova_sonic/llm.py
+++ b/src/pipecat/services/aws/nova_sonic/llm.py
@@ -12,6 +12,7 @@ bidirectional audio streaming, text generation, and function calling capabilitie
 
 import asyncio
 import base64
+import concurrent.futures
 import json
 import time
 import uuid
@@ -50,6 +51,10 @@ from pipecat.frames.frames import (
 )
 from pipecat.processors.aggregators.llm_context import LLMContext
 from pipecat.processors.frame_processor import FrameDirection
+from pipecat.services.aws.nova_sonic.session_continuation import (
+    SessionContinuationHelper,
+    SessionContinuationParams,
+)
 from pipecat.services.llm_service import LLMService
 from pipecat.services.settings import NOT_GIVEN, LLMSettings, _NotGiven
 from pipecat.utils.time import time_now_iso8601
@@ -257,6 +262,7 @@ class AWSNovaSonicLLMService(LLMService):
         settings: Settings | None = None,
         system_instruction: str | None = None,
         tools: ToolsSchema | None = None,
+        session_continuation: SessionContinuationParams | None = None,
         **kwargs,
     ):
         """Initializes the AWS Nova Sonic LLM service.
@@ -300,6 +306,9 @@ class AWSNovaSonicLLMService(LLMService):
                 .. deprecated:: 0.0.105
                     Use ``settings=AWSNovaSonicLLMService.Settings(system_instruction=...)`` instead.
             tools: Available tools/functions for the model to use.
+            session_continuation: Configuration for automatic session continuation.
+                When enabled (the default), sessions are seamlessly rotated before
+                the AWS time limit (~8 minutes) with no user-perceptible interruption.
             **kwargs: Additional arguments passed to the parent LLMService.
         """
         # 1. Initialize default_settings with hardcoded defaults
@@ -406,6 +415,18 @@ class AWSNovaSonicLLMService(LLMService):
         self._user_text_buffer = ""
         self._completed_tool_calls = set()
         self._audio_input_started = False
+
+        # Session continuation helper. The service itself implements the
+        # NovaSonicSessionSender protocol (see methods below) so the helper can
+        # target either the current or next session without coupling to the
+        # service's internal config.
+        sc_params = session_continuation or SessionContinuationParams()
+        self._sc = SessionContinuationHelper(
+            sc_params,
+            sender=self,
+            create_task=lambda coro: self.create_task(coro),
+            cancel_task=lambda task, timeout: self.cancel_task(task, timeout=timeout),
+        )
         self._pending_speculative_text: str | None = None
 
         file_path = files("pipecat.services.aws.nova_sonic").joinpath("ready.wav")
@@ -533,6 +554,16 @@ class AWSNovaSonicLLMService(LLMService):
         if self._triggering_assistant_response:
             return
 
+        # Session continuation: let the helper buffer audio during the transition window
+        self._sc.on_audio_input(frame.audio)
+
+        # Stop sending audio to the old stream once a handoff is in progress.
+        # Audio is still being buffered above and will be replayed to the new
+        # session. Matches reference: old session state set to CLOSING stops
+        # audio routing before close events are sent.
+        if self._sc.handoff_in_progress:
+            return
+
         await self._send_user_audio_event(frame.audio)
 
     async def _handle_interruption_frame(self):
@@ -555,7 +586,7 @@ class AWSNovaSonicLLMService(LLMService):
             self._input_audio_content_name = str(uuid.uuid4())
 
             # Create the client
-            self._client = self._create_client()
+            self._client = self.create_client()
 
             # Start the bidirectional stream
             self._stream = await self._client.invoke_model_with_bidirectional_stream(
@@ -643,13 +674,21 @@ class AWSNovaSonicLLMService(LLMService):
                 text=last_user_message.text, role=last_user_message.role, interactive=True
             )
 
-        # Start receiving events
-        self._receive_task = self.create_task(self._receive_task_handler())
+        # Start receiving events (bound to the current stream)
+        self._receive_task = self.create_task(self._receive_task_handler(stream=self._stream))
 
         # Record finished connecting time (must be done before sending assistant response trigger)
         self._connected_time = time.time()
 
         logger.info("Finished connecting")
+
+        # Notify session continuation helper of connection and start monitoring
+        self._sc.set_connected(self._connected_time)
+        # Seed the helper's history with initial context messages (these wouldn't be
+        # captured via real-time FINAL text events since they pre-date the session)
+        for message in llm_connection_params["messages"]:
+            self._sc.seed_history(message.role.value, message.text)
+        self._sc.start_monitor()
 
         # If we need to, send assistant response trigger (depends on self._connected_time)
         if self._triggering_assistant_response:
@@ -705,11 +744,17 @@ class AWSNovaSonicLLMService(LLMService):
             self._audio_input_started = False
             self._pending_speculative_text = None
 
+            # Stop session continuation monitor and notify of disconnect
+            await self._sc.stop_monitor()
+            await self._sc.cleanup_next_session()
+            self._sc.set_disconnected()
+
             logger.info("Finished disconnecting")
         except Exception as e:
             await self.push_error(error_msg=f"Error disconnecting: {e}", exception=e)
 
-    def _create_client(self) -> BedrockRuntimeClient:
+    def create_client(self) -> BedrockRuntimeClient:
+        """Create a new Bedrock runtime client (NovaSonicSessionSender protocol)."""
         config = Config(
             endpoint_uri=f"https://bedrock-runtime.{self._region}.amazonaws.com",
             region=self._region,
@@ -720,6 +765,11 @@ class AWSNovaSonicLLMService(LLMService):
             auth_schemes={"aws.auth#sigv4": SigV4AuthScheme(service="bedrock")},
         )
         return BedrockRuntimeClient(config=config)
+
+    @property
+    def audio_config(self) -> AudioConfig:
+        """Return the audio configuration (NovaSonicSessionSender protocol)."""
+        return self._audio_config
 
     def _is_first_generation_sonic_model(self) -> bool:
         # Nova Sonic (the older model) is identified by "amazon.nova-sonic-v1:0"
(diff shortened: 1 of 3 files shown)


examples/foundational/40a-aws-nova-sonic-session-continuation.py
```diff
@@ -0,0 +1,189 @@
...
+        "two or three sentences for chatty scenarios."
+    )
+
+    llm = AWSNovaSonicLLMService(
+        secret_access_key=os.getenv("AWS_SECRET_ACCESS_KEY"),
+        access_key_id=os.getenv("AWS_ACCESS_KEY_ID"),
+        region=os.getenv("AWS_REGION"),
+        session_token=os.getenv("AWS_SESSION_TOKEN"),
+        voice_id="tiffany",
+        # Session continuation is enabled by default to allow conversations
```
kompfner: How would you feel about moving these lines showing `SessionContinuationParams` into 40-aws-nova-sonic.py? I don't think we need a separate example for this, as longer-than-8-minute conversations ought to be covered by the base one.

examples/foundational/40a-aws-nova-sonic-session-continuation.py
```diff
@@ -0,0 +1,189 @@
...
+        # longer than the ~8-minute AWS session limit. The service seamlessly
+        # rotates sessions in the background. You can tune the parameters or
+        # disable it with: session_continuation=SessionContinuationParams(enabled=False)
+        session_continuation=SessionContinuationParams(
+            # Start preparing the next session after 6 minutes (default).
+            # For testing, you can lower this to e.g. 20 seconds.
+            transition_threshold_seconds=20,
+            # Buffer 1 second of user audio during the transition window
```
kompfner: Here you may also want to mention what the default is.

Or, actually...maybe omit showing `audio_buffer_duration_seconds` at all in the example? I can see why it's helpful to show setting a custom `transition_threshold_seconds` because doing so could be useful for testing. I'm not entirely clear on why we necessarily need to highlight setting a custom `audio_buffer_duration_seconds` (especially one that's _lower_ than the default, which I assume is a reasonable tried-and-true value). I imagine that needing to stray from the default would be rare.

WDYT?

prettyprettyprettygood: make sense, I'll do that.

examples/foundational/40a-aws-nova-sonic-session-continuation.py
kompfner: Nit: "allow _seamless_ conversations longer than..."

_Technically_ even with `session_continuation` disabled you can have longer conversations (assuming you are using a `LLMContextAggregatorPair` to record context), but you might notice a hiccup during reconnection, as the bot will lose some of your audio and respond to whatever the last message was in the freshly re-seeded context.

src/pipecat/services/aws/nova_sonic/llm.py
```diff
@@ -189,54 +184,48 @@ class Params(BaseModel):
...
-    input_sample_rate: Optional[int] = Field(default=16000)
-    input_sample_size: Optional[int] = Field(default=16)
-    input_channel_count: Optional[int] = Field(default=1)
-
-    # Output
-    output_sample_rate: Optional[int] = Field(default=24000)
-    output_sample_size: Optional[int] = Field(default=16)
-    output_channel_count: Optional[int] = Field(default=1)
+    enabled: bool = Field(default=True)
+    transition_threshold_seconds: float = Field(default=360.0, le=420.0)
```
kompfner: If there's a chance the hard 8-minute limit will change on the AWS server side, it might be worth avoiding a hard "less than 7 minutes" validation here as a bit of future-proofing (so anyone on this version of Pipecat can extend `transition_threshold_seconds` if needed).

src/pipecat/services/aws/nova_sonic/llm.py
```diff
@@ -435,35 +343,39 @@ def __init__(
...
         self._completed_tool_calls = set()
-        self._audio_input_started = False
+
+        # Session continuation state
+        self._session_continuation = session_continuation or SessionContinuationParams()
+        self._next_session: Optional[_NextSession] = None
+        self._sc_audio_buffer: Deque[bytes] = deque()
+        self._sc_audio_buffer_max_bytes: int = int(
+            self._session_continuation.audio_buffer_duration_seconds
+            * self._params.input_sample_rate
```
kompfner: This will conflict with changes in `main`. I think you can grab these values that were previously in `self._params` from `self._audio_config` instead.

src/pipecat/services/aws/nova_sonic/llm.py
```diff
@@ -685,40 +611,27 @@ async def _finish_connecting_if_context_available(self):
...
+        for message in llm_connection_params["messages"]:
             # logger.debug(f"Seeding conversation history with message: {message}")
-            is_last_message = i == len(messages) - 1
-            if is_last_message and message.role == Role.USER:
-                # Save for sending after audio input starts
-                last_user_message = message
-            else:
-                await self._send_text_event(text=message.text, role=message.role)
+            await self._send_text_event(text=message.text, role=message.role)
+            # Seed session continuation history with initial context messages
```
kompfner: I think this change might conflict with [this fix](https://github.com/pipecat-ai/pipecat/commit/06b3ecd2d69d387a52a550672147e0d82468ba87), which went in a few weeks ago

src/pipecat/services/aws/nova_sonic/llm.py
```diff
@@ -862,76 +763,76 @@ async def _send_prompt_start_event(self, tools: List[Any]):
...
         '''
-        await self._send_client_event(prompt_start)
+        await self._send_client_event(prompt_start, stream)
 
-    async def _send_audio_input_start_event(self):
-        if not self._prompt_name:
+    async def _send_audio_input_start_event(
+        self, *, prompt_name=None, content_name=None, stream=None
+    ):
+        """Send audio input contentStart. Defaults to current session."""
```
kompfner: Nit: I think it might be worth a quick explainer in all these docstrings of what you mean by "current session". Something along the lines of "(as opposed to the next session, in the case of a session continuation handoff)". A bit clunky, but may be worth it for future readers trying to understand this stuff.

src/pipecat/services/aws/nova_sonic/llm.py
```diff
@@ -1403,14 +1439,439 @@ async def _report_user_transcription_ended(self):
...
+                if llm_params["tools"]
+                else adapter.from_standard_tools(self._tools)
+            )
+            await self._send_prompt_start_event(tools, prompt_name=ns.prompt_name, stream=ns.stream)
+
+            # Send system instruction
+            system_instruction = (
+                llm_params["system_instruction"]
+                if llm_params["system_instruction"]
+                else self._system_instruction
```
kompfner: This may also conflict with `main` (`_system_instruction` moved under `self._settings`).

src/pipecat/services/aws/nova_sonic/llm.py
```diff
@@ -1403,14 +1439,439 @@ async def _report_user_transcription_ended(self):
...
+            ns = self._next_session
+            if not ns:
+                logger.warning("Session continuation: no next session available for handoff")
+                return
+
+            logger.info("Session continuation: executing handoff")
+
+            # Send conversation history to the next session. This uses our own
+            # real-time history (updated as FINAL text events arrive), not the
+            # pipeline context which may lag.
```
kompfner: Did you happen to confirm that through session swaps the pipeline context still records an accurate history of the conversation?

One way to test this is with example 20e - we should be able to save conversation history files to disk (where you can validate it) and load it in future conversations.

src/pipecat/services/aws/nova_sonic/llm.py
```diff
@@ -1403,14 +1439,439 @@ async def _report_user_transcription_ended(self):
...
+        self._sc_next_session_created_time = time.time()
+
+        ns = self._next_session
+
+        # Send sessionStart
+        await self._send_client_event(self._build_session_start_event_json(), ns.stream)
+
+        # Build context params from current context
+        if self._context:
+            adapter: AWSNovaSonicLLMAdapter = self.get_llm_adapter()
```
kompfner: This connection logic overlaps a reasonable amount with the logic in `_finish_connecting_if_context_available`...is it worth DRYing up somehow, or did you find it diverged enough to be worth maintaining the two paths separately?

src/pipecat/services/aws/nova_sonic/llm.py
```diff
@@ -435,35 +343,39 @@ def __init__(
...
+        self._session_continuation = session_continuation or SessionContinuationParams()
+        self._next_session: Optional[_NextSession] = None
+        self._sc_audio_buffer: Deque[bytes] = deque()
+        self._sc_audio_buffer_max_bytes: int = int(
+            self._session_continuation.audio_buffer_duration_seconds
+            * self._params.input_sample_rate
+            * (self._params.input_sample_size / 8)
+            * self._params.input_channel_count
+        )
+        self._sc_is_buffering = False
```
kompfner: Is there a way we can encapsulate all (or most) of this session continuation state, as well as the `_sc_*` methods, into some sort of "session continuation helper"—with a clean API—that we then import into this file? There's enough complexity related to this feature that it might be worth walling it off somewhat. Another benefit of encapsulation would be that if/when very long-running connections become supported fully—or much more simply—on the AWS server side, we could more easily excise this logic.

prettyprettyprettygood: That was my initial thought but I wanted to ensure to respect the structure you already have. As long as you are okay with it, I'm happy to move this to a helper class.

examples/foundational/40a-aws-nova-sonic-session-continuation.py
prettyprettyprettygood: Done. Removed the separate 40a example. Session continuation params are now shown in the main realtime-aws-nova-sonic.py example.

examples/foundational/40a-aws-nova-sonic-session-continuation.py
prettyprettyprettygood: Done. The example now only shows transition_threshold_seconds with a comment noting the default (360s). audio_buffer_duration_seconds is omitted.

src/pipecat/services/aws/nova_sonic/llm.py
prettyprettyprettygood: SessionContinuationHelper now receives input_sample_rate, input_sample_size, and input_channel_count from self._audio_config.

src/pipecat/services/aws/nova_sonic/llm.py
prettyprettyprettygood: done.

src/pipecat/services/aws/nova_sonic/llm.py
prettyprettyprettygood: Done. All session continuation state and logic lives in session_continuation.py as SessionContinuationHelper. It communicates with the LLM service through a NovaSonicSessionSender protocol, keeping the coupling minimal. The LLM file only has thin self._sc.*  call sites. This should make it straightforward to remove if/when long-running sessions are natively supported server-side.

src/pipecat/services/aws/nova_sonic/llm.py
prettyprettyprettygood: The shared wire-format serialization is now DRY via the NovaSonicSessionSender protocol — both paths use the same build_session_start_json, send_prompt_start, send_text, and send_audio_input_start methods. The next-session setup in _prepare_next_session calls these through the protocol. The remaining divergence is intentional: _finish_connecting_if_context_available handles tool call bookkeeping, the interactive last-user-message trigger, assistant response trigger, and receive task creation — none of which apply to next-session prep (where history and audio are deferred to handoff time).

src/pipecat/services/aws/nova_sonic/llm.py
```diff
@@ -996,6 +920,248 @@ async def _send_client_event(self, event_json: str):
...
+            "event": {{
+                "contentStart": {{
+                    "promptName": "{prompt_name}",
+                    "contentName": "{content_name}",
+                    "type": "AUDIO",
+                    "interactive": true,
+                    "role": "USER",
+                    "audioInputConfiguration": {{
+                        "mediaType": "audio/lpcm",
+                        "sampleRateHertz": {self._audio_config.input_sample_rate},
```
kompfner, line 1042: Feels rather fragile that we pass individual audio config values to the session continuation helper at its init time, but it requires those _same_ values to implicitly be used by methods in its `sender` (like we do here).

One way to avoid this "two things must magically be synced" problem might be to have the `sender` vend an `audio_config` getter property as part of the sender protocol, that the session continuation helper reads from, avoiding the need for the audio config init args.

src/pipecat/services/aws/nova_sonic/llm.py
```diff
@@ -643,14 +677,22 @@ async def _finish_connecting_if_context_available(self):
...
+        self._receive_task = self.create_task(self._receive_task_handler(stream=self._stream))
 
         # Record finished connecting time (must be done before sending assistant response trigger)
         self._connected_time = time.time()
 
         logger.info("Finished connecting")
 
+        # Notify session continuation helper of connection and start monitoring
+        self._sc.set_connected(self._connected_time)
+        # Seed the helper's history with initial context messages (these wouldn't be
```
kompfner, line 687: Just to confirm: the new "pending" connection will also be configured with the right system instruction and tools, right?

kompfner: Ah wait, pardon the noise - it does. All good.

src/pipecat/services/aws/nova_sonic/llm.py
```diff
@@ -996,6 +920,248 @@ async def _send_client_event(self, event_json: str):
...
+                "audioType": "SPEECH"
+              }}{tools_config}
+            }}
+          }}
+        }}
+        '''
+        await self.send_event(event_json, stream)
+
+    def get_setup_params(self):
+        """Return ``(system_instruction, tools, messages)`` for the next session setup."""
```
kompfner: Why do we need to send back `messages` here? Looks like it's not used, in favor of another separate mechanism (`seed_history`) for getting the conversation history into the next session.

src/pipecat/services/aws/nova_sonic/llm.py
prettyprettyprettygood: Good call. Added audio_config as a property on the NovaSonicSessionSender protocol. The helper now reads self._sender.audio_config directly instead of storing its own copies. Single source of truth.

src/pipecat/services/aws/nova_sonic/llm.py
prettyprettyprettygood: Stale from the pre-refactor code. History now comes from the helper's real-time tracking. Removed messages from the return value and get_setup_params now returns just (system_instruction, tools).

kompfner (approved):
Looks good! Nice work!

PR #4022, opened 2026-03-13
Draft Implementation for Krisp VIVA VAD.

realgarik:


filipi87 (commented):
@realgarik  Should we also change `examples/foundational/07p-interruptible-krisp-viva.py` to use this new `KrispVivaVadAnalyzer` instead of `SileroVADAnalyzer` ?

src/pipecat/audio/vad/krisp_viva_vad.py
```diff
@@ -0,0 +1,217 @@
...
+            # Process through VAD session
+            voice_probability = self._session.process(audio_float32)
+
+            return voice_probability
+
+        except Exception as e:
+            logger.error(f"Error analyzing audio with Krisp VIVA VAD: {e}", exc_info=True)
+            return 0.0
+
+    def __del__(self):
```
filipi87: Inside `VADController`, we should override:
`async def cleanup(self):`

Then call `self._vad_analyzer.cleanup()
`
So, instead of `def __del__(self):`, we would define `async def cleanup(self):`.

Happy to push these changes if needed.

filipi87 (commented):
One more thing, we should also include a changelog file, describing the changes that we are doing here. Like [this one](https://github.com/pipecat-ai/pipecat/pull/3391/changes#diff-4f96ff7d5b62d727841e44710fb8175059caf14d47dbc3a021cdc8d4faf61037) for example.

If you are using Claude, there is a skill inside Pipecat /changelog. So, just point to this PR and Claude will create it for you.
`/changelog  4022`

examples/foundational/07p-interruptible-krisp-viva.py
```diff
@@ -63,17 +64,20 @@
     "daily": lambda: DailyParams(
         audio_in_enabled=True,
         audio_out_enabled=True,
-        audio_in_filter=krisp_viva_filter,
+        vad_analyzer=SileroVADAnalyzer(params=VADParams(stop_secs=0.2)),  # or KrispVivaVadAnalyzer
+        audio_in_filter=KrispVivaFilter(),
```
filipi87, line 68: You should revert all these changes. This usage is deprecated. If you want to reference `KrispVivaVadAnalyzer`, do it where we create the `vad_analyzer` inside `LLMContextAggregatorPair` (line 107).

filipi87 (approved):
LGTM! I will push the final fixes to the examples, along with the `VadController` changes, in a follow-up PR. Thanks for this, @realgarik!

PR #4024, opened 2026-03-13
Fix Language enum conversion at init time in base TTS/STT services

markbackman:
## Summary

- Fixed `Language` enum values not being converted to service-specific codes when passed via `settings=Service.Settings(language=Language.ES)` at init time. This caused API errors (e.g. 400 from Rime) because the raw enum was sent instead of the expected language code (e.g. `"spa"`). Runtime updates via `UpdateSettingsFrame` were unaffected.
- Centralized language conversion in the base `TTSService.__init__` and `STTService.__init__`, using the subclass's `language_to_service_language()` via normal Python method resolution.
- Removed redundant per-service language conversion from 22 subclass `__init__` methods (hardcoded defaults, deprecated `params` paths, deprecated direct arg paths). Subclasses now just pass raw `Language` enums and let the base class convert.
- Found two missing deprecations for `language` arg. See https://github.com/pipecat-ai/pipecat/pull/4024/changes/0ec5f5e5ac528c89872c49a5fb9c8a3fde8fd72c

## Testing

- `uv run ruff check` — passes
- `uv run pytest` — all 1030 tests pass


aconchillo (approved):

aconchillo:
Much cleaner!

kompfner (commented):
Just got around to seeing this. This is a very nice improvement. The init-time conversions scattered everywhere really bugged me.

PR #4026, opened 2026-03-13
Fix DeepgramSTTService base_url forcing HTTPS/WSS schemes

### LINKED ISSUE #4019: DeepgramSTTService Forcing HTTPS or WSS
### pipecat version

0.0.105

### Python version

3.11.5

### Operating System

MacOS 26.2

### Issue description

The new DeepgramSTTService implementation requires the base_url to be https or wss, if it is not it will prepend those protocols. When running Deepgram in an airgapped or private environment using https or wss is not necessary and adds additional overhead. So requiring https or wss is redundant and should be removed. The user should be able to pass the exact url without it being modified.

```
        if base_url:
            try:
                from deepgram import DeepgramClientEnvironment

                ws_url = base_url if base_url.startswith("wss://") else f"wss://{base_url}"
                http_url = base_url if base_url.startswith("https://") else f"https://{base_url}"
                environment = DeepgramClientEnvironment(
                    base=http_url,
                    production=ws_url,
                    agent=ws_url,
                )
                self._client = AsyncDeepgramClient(api_key=api_key, environment=environment)
```

### Reproduction steps

Create DeepgramSTTService and set the base_url argument as http:// or ws://

### Expected behavior

Allows the base_url that you pass without modification

### Actual behavior

Guarantees the URL starts with https:// or wss://

### Logs

```shell

```

markbackman:
## Summary

- Fixed `DeepgramSTTService` `base_url` parameter silently forcing `wss://` / `https://` schemes, breaking air-gapped or private deployments using `ws://` or `http://`
- Extracted URL derivation into a `_derive_deepgram_urls()` helper that respects the developer's scheme choice and consistently derives both the WebSocket and HTTP URLs the Deepgram SDK requires
- Unrecognized schemes now log a warning and fall back to secure defaults

## Testing

```
uv run pytest tests/test_deepgram_stt.py -v
```

## Fixes

- Fixes #4019

### DIFF
--- a/changelog/4026.fixed.md
+++ b/changelog/4026.fixed.md
@@ -1,0 +1,1 @@
+- Fixed `DeepgramSTTService` ignoring the `base_url` scheme when using `ws://` or `http://`. Previously these were silently overwritten with `wss://` / `https://`, breaking air-gapped or private deployments that don't use TLS. All scheme choices (`wss://`, `https://`, `ws://`, `http://`, or bare hostname) are now respected.
--- a/src/pipecat/services/deepgram/stt.py
+++ b/src/pipecat/services/deepgram/stt.py
@@ -247,6 +247,45 @@ class DeepgramSTTSettings(STTSettings):
                 del self.extra[key]
 
 
+def _derive_deepgram_urls(base_url: str) -> tuple[str, str]:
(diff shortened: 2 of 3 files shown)


filipi87 (approved):
Makes sense. 👍

PR #4028, opened 2026-03-13
fix(inworld): close context at end of turn instead of relying on idle timeout

ianbbqzy:
The Inworld WS TTS plugin previously relied on the base TTS service's 3-second AUDIO_CONTEXT_TIMEOUT to detect when audio was done, then sent close_context in on_audio_context_completed. This added unnecessary latency before TTSStoppedFrame was emitted.  This also reduces concurrent contexts by closing contexts as soon as they are not needed anymore

The original implementation likely borrowed this idea from the 11labs' impelementation. But it's likely better to mirror the Cartesia plugin pattern where on_audio_context_completed is a no-op because the server signals completion directly.

Now close_context is sent in on_turn_context_completed (right after flush_context), so the server responds with contextClosed immediately after the last audio byte. The existing receive handler already calls remove_audio_context on contextClosed, which exits the audio context handler cleanly.

### DIFF
--- a/src/pipecat/services/inworld/tts.py
+++ b/src/pipecat/services/inworld/tts.py
@@ -774,12 +774,18 @@ class InworldTTSService(WebsocketTTSService):
         self._cumulative_time = 0.0
         self._generation_end_time = 0.0
 
+    async def on_turn_context_completed(self):
+        """Close the server-side context at end of turn.
+
+        Sends close_context so contextClosed arrives immediately after the
+        last audio byte.
+        """
+        ctx_id = self._turn_context_id
+        await super().on_turn_context_completed()
+        await self._close_context(ctx_id)
+
     async def on_audio_context_interrupted(self, context_id: str):
         """Callback invoked when an audio context has been interrupted."""
-        await self._close_context(context_id)
-
-    async def on_audio_context_completed(self, context_id: str):
-        """Callback invoked when an audio context has been completed."""
         await self._close_context(context_id)
 
     def _get_websocket(self):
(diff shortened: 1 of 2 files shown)


src/pipecat/services/inworld/tts.py
```diff
@@ -774,14 +774,20 @@ async def _close_context(self, context_id: str):
         self._cumulative_time = 0.0
         self._generation_end_time = 0.0
 
+    async def on_turn_context_completed(self):
```
filipi87, line 777: If you override it here instead of inside `on_audio_context_completed`, both `on_turn_context_completed` and `on_audio_context_interrupted` could be called, and you might end up invoking `self._close_context` twice for the same `context_id`. Would that be an issue?

ianbbqzy: the duplicate `close_context` is harmless: the server returns code 5 "Context not found", the client logs it and moves on. I believe this scenario you are describing is rare correct? It only happens if interruption happens between `on_turn_context_completed` and before a `context_closed` message from server arrives.

So i think the same already exists, we will try to close the contexts twice if the interruption happens between `on_audio_context_interrupted` and `context_closed`.

we can add a set to keep track of `close_context` messages sent to server, but I feel that might be extra complexity without clear benefits. WDYT?

filipi87: It’s not that rare, because on_turn_context_completed is triggered as soon as we receive the full response from the LLM. In the meantime, however, we still need to receive all the audio from the TTS service. Even after that, there’s an additional timeout before we mark the audio context as closed.

So, if the user interrupts a couple of seconds after we’ve received the full LLM response, this situation can occur.

If the server handles this correctly, I think it’s fine. I don’t think we should add an extra set for this.

filipi87 (approved):
LGTM!

PR #4029, opened 2026-03-13
`SyncParallelPipeline` and related fixes

kompfner:
## Summary

- Fixed `SyncParallelPipeline` race condition where concurrent `SystemFrame` processing (e.g. from RTVI) could corrupt sink queues and cause deadlocks. SystemFrames now take a fast path that passes them through without draining queued output.
- Fixed `SyncParallelPipeline` breaking the Whisker debugger by returning the wrong type from `pipeline_processors()`.
- Fixed TTS frame ordering when an audio context is active. Pass-through frames are now routed through the audio context queue so they stay properly ordered with TTS audio frames.
- Added `sync_with_audio` field to `OutputImageRawFrame`. When set to `True`, the output transport queues image frames with audio so they are displayed only after all preceding audio has been sent.
- Added `frame_order` parameter to `SyncParallelPipeline`. Set `frame_order=FrameOrder.PIPELINE` to push synchronized output frames in pipeline definition order (all frames from the first pipeline, then the second, etc.) instead of the default arrival order.
- Updated `05-sync-speech-and-image` example to use `FrameOrder.PIPELINE` for synchronized audio/image playback.

## Testing

- `uv run pytest tests/test_sync_parallel_pipeline.py`
- Run `05-sync-speech-and-image.py` example to verify synchronized image/audio playback

### DIFF
--- a/changelog/4029.added.2.md
+++ b/changelog/4029.added.2.md
@@ -1,0 +1,1 @@
+- Added `frame_order` parameter to `SyncParallelPipeline`. Set `frame_order=FrameOrder.PIPELINE` to push synchronized output frames in pipeline definition order (all frames from the first pipeline, then the second, etc.) instead of the default arrival order.
--- a/changelog/4029.fixed.2.md
+++ b/changelog/4029.fixed.2.md
@@ -1,0 +1,1 @@
+- Fixed TTS frame ordering when an audio context is active. Pass-through frames are now routed through the audio context queue so they stay properly ordered with TTS audio frames. Previously, a frame that preceded some incoming text to a TTS service could "jump the queue", getting ahead of the outgoing audio.
--- a/changelog/4029.fixed.2.md
+++ b/changelog/4029.fixed.2.md
@@ -1,1 +1,0 @@
-- Fixed TTS frame ordering when an audio context is active. Pass-through frames are now routed through the audio context queue so they stay properly ordered with TTS audio frames. Previously, a frame that preceded some incoming text to a TTS service could "jump the queue", getting ahead of the outgoing audio.
--- a/changelog/4029.fixed.md
+++ b/changelog/4029.fixed.md
@@ -1,0 +1,1 @@
+- Fixed `SyncParallelPipeline` race condition where concurrent SystemFrame processing (e.g. from RTVI) could corrupt sink queues and cause deadlocks. SystemFrames now take a fast path that passes them through without draining queued output.
--- a/examples/foundational/05-sync-speech-and-image.py
+++ b/examples/foundational/05-sync-speech-and-image.py
@@ -31,6 +31,7 @@ from pipecat.runner.utils import create_transport
 from pipecat.services.cartesia.tts import CartesiaHttpTTSService
 from pipecat.services.fal.image import FalImageGenService
 from pipecat.services.openai.llm import OpenAILLMService
+from pipecat.services.tts_service import TextAggregationMode
 from pipecat.transports.base_transport import BaseTransport, TransportParams
 from pipecat.transports.daily.transport import DailyParams
 
@@ -114,6 +115,10 @@ async def run_bot(transport: BaseTransport, runner_args: RunnerArguments):
             settings=CartesiaHttpTTSService.Settings(
                 voice="71a7ad14-091c-4e8e-a314-022ece01c121",  # British Reading Lady
             ),
+            # No need to aggregate by sentences (the default), as we already know we're getting full sentences
+            # (Otherwise the service will unnecessarily wait for follow-up input to confirm the sentence is complete,
+            #  which, sadly, actually breaks the synchronization mechanism)
+            text_aggregation_mode=TextAggregationMode.TOKEN,
         )
 
         imagegen = FalImageGenService(
--- a/examples/foundational/05-sync-speech-and-image.py
+++ b/examples/foundational/05-sync-speech-and-image.py
@@ -12,21 +12,16 @@ from dotenv import load_dotenv
 from loguru import logger
 
 from pipecat.frames.frames import (
-    AggregatedTextFrame,
     DataFrame,
     Frame,
     LLMContextFrame,
     LLMFullResponseStartFrame,
     OutputImageRawFrame,
     TextFrame,
-    TTSAudioRawFrame,
-    TTSStartedFrame,
-    TTSStoppedFrame,
-    TTSTextFrame,
 )
 from pipecat.pipeline.pipeline import Pipeline
 from pipecat.pipeline.runner import PipelineRunner
-from pipecat.pipeline.sync_parallel_pipeline import SyncParallelPipeline
+from pipecat.pipeline.sync_parallel_pipeline import FrameOrder, SyncParallelPipeline
 from pipecat.pipeline.task import PipelineTask
 from pipecat.processors.aggregators.llm_context import LLMContext
 from pipecat.processors.aggregators.sentence import SentenceAggregator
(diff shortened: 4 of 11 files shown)


src/pipecat/services/tts_service.py
```diff
@@ -758,7 +759,17 @@ async def process_frame(self, frame: Frame, direction: FrameDirection):
             await self._maybe_resume_frame_processing()
             await self.push_frame(frame, direction)
         else:
-            await self.push_frame(frame, direction)
+            # If there's an active audio context, route the frame through it
```
kompfner: This is the part of these changes that will have the widest impact. We should make sure to test this one thoroughly. cc @filipi87

kompfner:
TODO: also fix 05a-local-sync-speech-and-image.py

changelog/4029.fixed.md
```diff
@@ -0,0 +1 @@
+- Fixed `SyncParallelPipeline` race condition where concurrent SystemFrame processing (e.g. from RTVI) could corrupt sink queues and cause deadlocks. SystemFrames now take a fast path that passes them through without draining queued output.
```
kompfner, line 1: This actually was a longstanding issue with `SyncParallelPipeline`, but we happened not to notice it until on-by-default RTVI sent a flood of system frames through it—that's why the 05- example seemed to break between versions 0.0.100 and 0.0.101.

kompfner:
Wait, I have a different/better idea for this PR. Putting it back into draft state while I work on the changes.

kompfner:
Not sure what to do about example 05a (the `TkLocalTransport` variant)...it doesn't seem to work, even on `main`. Is it important enough to keep around and invest time into? If not, we could remove it as part of this PR. Thoughts?

changelog/4029.added.2.md
```diff
@@ -0,0 +1 @@
+- Added `frame_order` parameter to `SyncParallelPipeline`. Set `frame_order=FrameOrder.PIPELINE` to push synchronized output frames in pipeline definition order (all frames from the first pipeline, then the second, etc.) instead of the default arrival order.
```
filipi87, line 1: Was this also a long standing issue?

examples/foundational/05-sync-speech-and-image.py
```diff
@@ -44,6 +46,18 @@ def __str__(self):
...
 
 
+class MarkImageForPlaybackSync(FrameProcessor):
+    """Marks output image frames to be synchronized with audio playback."""
+
+    async def process_frame(self, frame: Frame, direction: FrameDirection):
+        await super().process_frame(frame, direction)
+
+        if isinstance(frame, OutputImageRawFrame):
+            frame.sync_with_audio = True
```
filipi87, line 56: I wonder if this logic is something we should include inside `SyncParallelPipeline` instead. 

In that case, when creating it, we could add a flag (`sync_image_with_audio`) and automatically set `sync_with_audio` for `OutputImageRawFrame`. Would that make sense ?

src/pipecat/pipeline/sync_parallel_pipeline.py
```diff
@@ -215,66 +245,113 @@ async def process_frame(self, frame: Frame, direction: FrameDirection):
...
         Args:
             frame: The frame to process.
             direction: The direction of frame flow.
         """
         await super().process_frame(frame, direction)
 
+        # SystemFrames (but not EndFrame) are simply passed through all
+        # internal pipelines without draining queued output. This avoids
+        # the race condition where a SystemFrame's wait_for_sync steals
+        # frames from a concurrent non-SystemFrame's wait_for_sync.
```
filipi87, line 269: So, `SyncParallelPipeline` is only supposed to work with non-`SystemFrames`, is that correct?

Are we documenting this somewhere?

src/pipecat/services/tts_service.py
filipi87: @kompfner, I actually believe we’ll need to do even more than that. We should handle all the frames inside the audio queue so we don’t need to pause frame processing.

For example, with Cartesia, where `pause_frame_processing` is `False` and multiple contexts can exist at the same time, this fix alone wouldn’t be enough to guarantee frame ordering.

I will work on this in a separate PR, and we can test it together with this one.

filipi87: I believe this is the full fix:
- https://github.com/pipecat-ai/pipecat/pull/4075

changelog/4029.added.2.md
kompfner: My guess is that when the 05- example was being developed, maybe the order in which image v audio frames arrived at the output transport didn't matter as much for some reason? But I'm not entirely sure why that would be the case. Or there was always a subtle bug where _sometimes_ (but not always) an image would be shown after the corresponding audio was played.

examples/foundational/05-sync-speech-and-image.py
kompfner: Hmm, to me this falls clearly outside the scope of things `SyncParallelPipeline` should be aware of. It has no idea what kind of frames are going through it, only that all frames generated in response to a certain inbound frame should be released at the same time coming out. This field is related to how media is queued in the output transport.

src/pipecat/pipeline/sync_parallel_pipeline.py
kompfner: That's my proposal, yes. 

We _already_ expect `SystemFrame`s to be able to "race ahead" of other frames, so I don't think we should expect them to drain the `SyncParallelPipeline`'s queue just for the sake of keeping their relative order with non-system frames (which, again, isn't something that matters anyway!) Would also love @aconchillo's feedback on this particular change, since I believe he wrote `SyncParallelPipeline` (back in the day hehe).

But yeah, `SystemFrame`s aren't going to be synchronized by the `SyncParallelPipeline`. That should for sure be documented clearly. I'll do a pass on the docstrings and comments in this file.

kompfner: Pushed a commit updating docstrings for clarity: https://github.com/pipecat-ai/pipecat/pull/4029/commits/216d7bd4baeb0ab5acd1c37491f1fd3080abb0fa

changelog/4029.added.2.md
filipi87: In any case, this does feel like a nice addition to the `SyncParallelPipeline`.

src/pipecat/pipeline/sync_parallel_pipeline.py
filipi87: Makes sense to me as well. 👍

filipi87 (commented):
@kompfner, I have rebased the TTS frame order [PR](https://github.com/pipecat-ai/pipecat/pull/4075), and pushed some extra changes there. 

So, you will probably need a rebase here as well.

Once that PR lands, I think we are good to merge this one 🚀.

kompfner:
> @kompfner, I have rebased the TTS frame order [PR](https://github.com/pipecat-ai/pipecat/pull/4075), and pushed some extra changes there.
> 
> So, you will probably need a rebase here as well.
> 
> Once that PR lands, I think we are good to merge this one 🚀.

Hm, I opted to use Github's "resolve conflicts" tool but it screwed up this PR pretty badly. Fixing.

changelog/4029.fixed.2.md
filipi87, line 1: I guess this changelog it is not needed anymore.

filipi87 (approved):
Looks good. There is only one changelog entry that I believe we need to remove. Other than that, it looks great. 🚀

changelog/4029.fixed.2.md
kompfner: good catch!

src/pipecat/pipeline/sync_parallel_pipeline.py
```diff
@@ -209,72 +251,114 @@ async def cleanup(self):
...
+        if isinstance(frame, SystemFrame) and not isinstance(frame, EndFrame):
```
aconchillo, line 270: `and not isinstance(frame, EndFrame):` is not needed because `EndFrame`s are not system frames.

src/pipecat/pipeline/sync_parallel_pipeline.py
aconchillo: Yeah, I think it makes sense to let system frames go through. I don't think there's any need to synchronize them

src/pipecat/pipeline/sync_parallel_pipeline.py
kompfner: Ah good catch! Will open a tiny PR fixing this.

kompfner: https://github.com/pipecat-ai/pipecat/pull/4109

PR #4031, opened 2026-03-13
Add xAI TTS service

niczy:
## Summary

- Add `XAIHttpTTSService` backed by xAI's HTTP text-to-speech API
- Support all 20 xAI-supported languages via explicit `LANGUAGE_MAP` using `resolve_language()`
- Configurable output `encoding` (pcm, mp3, wav, mulaw, alaw) as a constructor parameter
- Add `07e-interruptible-xai.py` foundational example; update `14g-function-calling-grok.py` to use xAI TTS
- Register `07e` in `run-release-evals.py`

## Testing

- `uv run pytest tests/test_xai_tts.py`
- Test live with `examples/foundational/07e-interruptible-xai.py` using a valid `GROK_API_KEY`


src/pipecat/services/xai/__init__.py
```diff
@@ -0,0 +1,13 @@
+#
+# Copyright (c) 2024-2026, Daily
+#
+# SPDX-License-Identifier: BSD 2-Clause License
+#
+
+import sys
```
markbackman: You can leave this file blank. At the moment, the pattern in Pipecat is to use the full module path.

markbackman:
Thanks for the contribution! Here are some changes needed before we can merge:                                                       

1. Rename XAITTSService to XAIHttpTTSService — This is an HTTP-based service, and we want to leave room for a future WebSocket implementation. This follows the pattern of ElevenLabsHttpTTSService / CartesiaHttpTTSService. Update the settings class name to XAIHttpTTSSettings as well.
2. Add language map — language_to_service_language currently does return str(language) which passes through any Language enum value verbatim. xAI only supports 20 specific language codes (some regional like ar-EG, pt-BR, es-MX). Please add a language_to_xai_language() function with an explicit LANGUAGE_MAP using resolve_language(), like other services do (see cartesia/tts.py or elevenlabs/tts.py for examples). Also needs from pipecat.transcriptions.language import Language, resolve_language.
3. Remove deprecated InputParams, params, voice, language init params — Since this is a brand new service that has never shipped, there's nothing to be backward-compatible with.
4. Remove class constants XAI_DEFAULT_SAMPLE_RATE and XAI_PCM_CODEC - instead, we want to include encoding as a constructor parameter and set sample_rate to None, so that it can be set either by default from the PipelineParams or by the user via a constructor parameter.
5. Use Language.EN instead of string "en" for default language.
6. README link was missing; docs will be automatically generated after this is merged. Location will be https://docs.pipecat.ai/server/services/tts/xai
7. Linting fixes.
8. Examples: Added 07e foundational example; updated 14g foundational example.

I made a branch while I was reviewing these changes. I've pushed these to origin for you to use as a starting point:
https://github.com/pipecat-ai/pipecat/pull/4121

Also, once this change is in, I think we'll move the GrokLLMService to the new xai module, so that services are consolidate. We can do that in a separate PR though.

markbackman (approved):
LGTM!

I rebased on origin/main and pushed a few tiny clean up items, but this looks great. Thanks for adding this! 🙌

PR #4040, opened 2026-03-16
Add reproduction tests for TurnAnalyzerUserTurnStopStrategy interim deadlock

### LINKED ISSUE #3988: TurnAnalyzerUserTurnStopStrategy ignores InterimTranscriptionFrame, causing deadlock when STT delays finalized transcriptions
## Summary

`TurnAnalyzerUserTurnStopStrategy` only processes `TranscriptionFrame` (finalized) to set `_text`, but ignores `InterimTranscriptionFrame`. When an STT service delays finalized transcriptions for short utterances until more speech arrives, the stop strategy can never trigger — `_turn_complete` gets set by VAD stop, but `_text` stays empty.

This causes a deadlock where the bot goes silent for as long as the user stays quiet.

## Reproduction scenario

1. User says "I need help with my order" → turn completes → LLM triggers
2. User says "Billing." (short word) while LLM is generating → `InterimTranscriptionFrame("Billing")` arrives
3. `TranscriptionUserTurnStartStrategy` fires (from the interim) → new turn starts, LLM interrupted
4. VAD stop fires ~0.4s later → `TurnAnalyzerUserTurnStopStrategy` sets `_turn_complete = True`
5. But `_text` is still `""` because `_handle_transcription` only processes `TranscriptionFrame`:

```python
# turn_analyzer_user_turn_stop_strategy.py line 106
elif isinstance(frame, TranscriptionFrame):
    await self._handle_transcription(frame)
```

6. `_maybe_trigger_user_turn_stopped()` returns early:

```python
# line 213
if not self._text or not self._turn_complete:
    return
```

7. The STT holds the finalized transcription for "Billing." until the next sentence arrives — which could be 15+ seconds later
8. Bot sits silent until the user speaks again

## Root cause

`InterimTranscriptionFrame` and `TranscriptionFrame` are separate classes (both inherit from `TextFrame`), so `isinstance(frame, TranscriptionFrame)` doesn't match interims. The stop strategy has no way to populate `_text` from interim transcriptions.

## STT behavior context

Some STT services don't finalize short utterances immediately. A single word like "Billing." may only produce an `InterimTranscriptionFrame`, with the finalized `TranscriptionFrame` arriving only when the user speaks again (e.g., "Billing. Hello?" as a single finalized transcription 15+ seconds later).

This may also be an STT configuration issue on our side — any guidance on expected STT finalization behavior would be helpful.

## Suggested fix

Have `TurnAnalyzerUserTurnStopStrategy.process_frame` also handle `InterimTranscriptionFrame` to set `_text`, similar to how it handles `TranscriptionFrame`:

```python
elif isinstance(frame, InterimTranscriptionFrame):
    text = frame.text.strip()
    if text:
        self._text = text
        await self._maybe_trigger_user_turn_stopped()
        # Fallback path for no-VAD-stop scenario
        if not self._vad_user_speaking and self._vad_stopped_time is None:
            self._turn_complete = True
            ...
```

## Workaround

We currently work around this by setting `TranscriptionUserTurnStartStrategy(enable_interruptions=False)` and increasing VAD `stop_secs` to give users more time to finish multi-word answers within a single turn.

## Environment

- pipecat 0.0.104
- Turn analyzer: `LocalSmartTurnAnalyzerV3` with `SmartTurnParams(stop_secs=2)`

OlliTapio:
## Summary

Adds reproduction tests for the deadlock described in #3988, covering three layers of the turn management system.

### 1. Strategy level: `TurnAnalyzerUserTurnStopStrategy` (4 tests)

- **`test_sie_no_transcription_deadlock`** — Core reproduction: user says a short word, STT emits only `InterimTranscriptionFrame`, VAD stops, turn analyzer says COMPLETE, but `_text` stays empty because `process_frame` only handles `TranscriptionFrame`. Turn never completes.
- **`test_sie_t_basic_flow`** — Happy path with interim + finalized transcription
- **`test_se_no_transcription_timeout`** — No transcription at all (VAD false positive)
- **`test_sie_delayed_t_recovers`** — Delayed finalized transcription eventually recovers (but 15+ seconds too late)

### 2. Controller level: `UserTurnController` timeout (2 tests)

Addresses the suggested workaround of using `on_user_turn_stop_timeout`:

- **`test_controller_timeout_fires_when_stt_stops`** — Timeout works if the STT goes completely silent
- **`test_controller_timeout_blocked_by_continuous_interims`** — **But**: `InterimTranscriptionFrame` resets the timeout via `_handle_transcription` → `event.set()`. When the STT keeps emitting interims (refining its hypothesis), the 5s timeout never fires. This explains why `on_user_turn_stop_timeout` doesn't help in practice with some STT providers.

### 3. Idle controller: `UserIdleController` (2 tests)

- **`test_idle_timer_fires_when_no_turn_in_progress`** — Baseline: works normally
- **`test_idle_timer_blocked_by_user_turn_in_progress`** — The ghost turn sends `UserStartedSpeakingFrame` → `_user_turn_in_progress=True`. Since the turn never stops, `UserStoppedSpeakingFrame` never clears it, and the idle timer can never start.

### Key findings

1. **Root cause**: `TurnAnalyzerUserTurnStopStrategy.process_frame` ignores `InterimTranscriptionFrame`, so `_text` stays empty
2. **`user_turn_stop_timeout` is not a reliable safety net**: `InterimTranscriptionFrame` resets the controller's timeout clock, so continuous interims from the STT prevent it from ever firing
3. **`user_idle_timeout` is blocked**: the stuck turn holds `_user_turn_in_progress=True`, preventing the idle timer from starting

All tests assert the current (buggy) behavior and document what should change when fixed.

Closes #3988

## Test plan

- [x] All 8 tests pass on current main
- [x] Tests document bugs with comments explaining expected vs actual behavior

### DIFF
--- a/tests/test_turn_analyzer_user_turn_stop_strategy.py
+++ b/tests/test_turn_analyzer_user_turn_stop_strategy.py
@@ -290,18 +290,16 @@ class TestControllerTimeoutWithInterimDeadlock(unittest.IsolatedAsyncioTestCase)
         await controller.setup(self.task_manager)
         return controller
 
-    async def test_controller_timeout_fires_for_interim_only_turn(self):
-        """Full scenario: Turn 1 completes normally, then a delayed interim
-        starts a ghost turn that the stop strategy can't close.
+    async def test_controller_timeout_fires_when_stt_stops(self):
+        """Controller timeout fires if the STT stops emitting frames entirely.
 
-        The controller's user_turn_stop_timeout should fire as a safety net,
-        ending the stuck turn. This is the fallback Mark described.
+        When a ghost turn starts and the STT goes quiet, the controller's
+        user_turn_stop_timeout fires after CONTROLLER_STOP_TIMEOUT seconds of
+        inactivity. This is the fallback Mark described.
 
-        Sequence:
-        1. Turn 1: S E T → completes normally via stop strategy
-        2. Delayed I arrives → TranscriptionUserTurnStartStrategy fires new turn
-        3. No VAD, no finalized T → stop strategy is stuck
-        4. Controller timeout should fire after CONTROLLER_STOP_TIMEOUT
+        However, this only works if the STT actually stops emitting. See
+        test_controller_timeout_blocked_by_continuous_interims for the
+        realistic scenario where it doesn't help.
         """
         controller = await self._create_controller()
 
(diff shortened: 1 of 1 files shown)


OlliTapio:
## Context

These reproduction tests came from investigating #3988. After digging deeper, we believe the root cause in our case was Soniox STT configuration.

### STT configuration issue

Soniox defaults to `vad_force_turn_endpoint=True`, which sends a finalize message on VAD stop and disables Soniox's own endpoint detection. This works well for clean audio, but when background noise prevents VAD from firing a clean stop (we traced this to two production calls — one with the caller multitasking, one with heavy breathing), Soniox never finalizes the transcription and keeps emitting interims.

When using `TranscriptionUserTurnStartStrategy`, the STT can start turns via interim/finalized transcriptions, but with `vad_force_turn_endpoint=True` it can't reliably *end* them — it depends on VAD to trigger finalization. If VAD misses the endpoint, the finalized transcription never arrives.

We're testing with `vad_force_turn_endpoint=False` to let Soniox handle its own endpoint detection. This should be the right configuration when relying on transcription-based turn strategies.

### Safety net gap

While building the tests we noticed that `InterimTranscriptionFrame` resets `user_turn_stop_timeout` via `_handle_transcription` in `UserTurnController`. When the STT keeps emitting interims (as Soniox does with background noise), the 5s safety timeout never fires because each interim resets the clock. The `test_controller_timeout_blocked_by_continuous_interims` test documents this.

Should we open a separate issue for the timeout reset behavior?

OlliTapio:
Closing this PR — the root cause was on our side: we never registered the `on_user_turn_stop_timeout` event handler, so the 5s safety net existed but we weren't using it.

With the handler enabled (3.5s timeout) + Soniox `vad_force_turn_endpoint=False`, the deadlock is resolved without needing upstream changes.

Thanks Mark for pointing us in the right direction!

PR #4042, opened 2026-03-16
Improvements for Nova Sonic LLM and TTS output frames

markbackman:
## Summary

- Fixed empty user transcriptions causing spurious interruptions of the assistant's response. When the bot starts the conversation (text-only prompt, no user speech), the empty transcription no longer triggers an interruption.
- Nova Sonic assistant text transcripts now use speculative text events instead of final text events. Speculative text arrives before each audio chunk, providing real-time text synchronized with what the bot is saying. Final text arrives seconds after all audio is done and caused laggy transcripts in client UIs.
- Deferred TTSTextFrame emission to audio contentEnd so the text frame arrives after the corresponding audio chunks have been sent, aligning it with audio playout timing rather than firing at text contentEnd (before any audio).
- Moved LLMFullResponseStartFrame to fire at completionStart (before speculative text) so TTSTextFrames are always sandwiched between start/end frames.
- Removed dead code related to FINAL assistant text handling: the 1.25s delay in bot-stopped-speaking that waited for trailing FINAL text blocks, the background task machinery around it, the interruption re-push hack.

## Testing

- Run the Nova Sonic example: `uv run examples/foundational/40-aws-nova-sonic.py`
- Verify assistant text appears in real-time as the bot speaks (not delayed until after audio finishes)
- Verify context is built correctly after each turn (user and assistant messages present)
- Test interruptions: interrupt the bot mid-sentence and verify the conversation continues normally


kompfner:
> Nova Sonic assistant text transcripts now use speculative text events instead of final text events.

This definitely used to be a no-go at the time this service was first written, as we couldn't find a way to reliably use the speculative chunks to write to context. If my memory serves me right (it's been a while), it's because there wasn't a good way to "sync up" what was sent as speculative text events to what was then actually uttered by the bot. For example, if we received something like:
- Speculative 1
- Audio 1
- Speculative 2 (after Audio 1 playback delay)
- Audio 2
- ...
Then we could easily determine, on an interruption, which chunks of speculative text could go ahead and be written to context. But instead what we were seeing was more like:
- Speculative 1
- Audio 1
- Speculative 2
- Audio 2 (immediate; not after playback delay)
- Speculative 3
- ...

The fact that you could get large batches of speculative text at once meant that on an interruption we'd end up writing to potentially *way way* more than what the bot actually ended up speaking. We decided it was better to just wait for the final text chunks, at which point we at least knew the bot actually spoke the corresponding text. Of the two kinds of context inaccuracies - jotting down potentially way more than the bot spoke, or a little less - we decided that less is the preferable one.

But, again, it's been a while. I can try to find the original discussion for more context.

It's also possible that the API has improved substantially, and that this limitation went away!

kompfner:
In other words: assuming nothing's changed on the AWS API side from when this service was written:
- I agree with you that speculative text seems like the right source for "karaoke-style" displayed text
- But it's *not* the right source for storing in recorded context

changelog/4042.changed.md
```diff
@@ -0,0 +1 @@
+- Nova Sonic assistant text transcripts are now delivered in real-time using speculative text events instead of delayed final text events. Previously, assistant text only arrived after all audio had finished playing, causing laggy transcripts in client UIs. Speculative text arrives before each audio chunk, providing text synchronized with what the bot is saying. This also simplifies the internal text handling by removing the interruption re-push hack and assistant text buffer.
```
kompfner, line 1: > assistant text only arrived after all audio had finished playing

Wait doesn't "final" text arrive in chunks? In other words, during very long bot responses, didn't we previously get partial transcripts along the way?

markbackman:
@kompfner and I talked and are planning to use only speculative text. First to push an AggregatedTextFrame, which is used by the bot-output event to display the to be spoken text right away. Then, use the contentStart and contentEnd events to determine the corresponding speculative text for a given audio content block. Once the content audio block is finished, we'll push a TTSTextFrame with the speculative text. This will allow the context to be more accurately updated with the text that was just spoken. IT will also allow for a better client UI event to display text chunk by chunk as it's spoken.

src/pipecat/services/aws/nova_sonic/llm.py
```diff
@@ -1132,7 +1127,7 @@ async def _receive_task_handler(self):
                 await self.reset_conversation()
 
     async def _handle_completion_start_event(self, event_json):
-        pass
+        await self.push_frame(LLMFullResponseStartFrame())
```
kompfner: "Completion start" in Nova Sonic land doesn't quite mean what we might think it means (the start of a single assistant response being generated). The completion start event is what we receive at the start of the session. That's why `LLMFullResponseStartFrame` was previously only sent in `_report_assistant_response_started`.

src/pipecat/services/aws/nova_sonic/llm.py
```diff
@@ -1257,83 +1250,41 @@ async def _handle_completion_end_event(self, event_json):
...
-        # Report the start of the assistant response.
-        await self.push_frame(LLMFullResponseStartFrame())
-
         # Report that equivalent of TTS (this is a speech-to-speech model) started
         await self.push_frame(TTSStartedFrame())
 
-    async def _report_assistant_response_text_added(self, text):
-        if not self._context:  # should never happen
-            return
+    async def _report_assistant_speculative_text(self, text):
```
kompfner: The pattern for all of the `_report_` function naming was: `_report_<occurrence_in_pipecat_pipeline_terms>`, since it's "reporting" to the rest of the pipeline that something occurred.

So in this case, it would be something like `_report_llm_text_generated` (not set on that name, it was just the first thing that popped into my mind).

Also...looks like this code still pushes the `TTSTextFrame` immediately after the `LLMTextFrame`...I'm not seeing the change we discussed, where pushing the `TTSTextFrame` is deferred to after pushing the corresponding audio frames...did it not make it into the PR yet?

When the change does come in, same notes on function naming: perhaps the `TTSTextFrame`s could be pushed in a `_report_tts_text_spoken` function.

src/pipecat/services/aws/nova_sonic/llm.py
kompfner: After your changes, I'd imagine we'd want to push the `LLMFullResponseStartFrame` only when we detected that the assistant started generating a response (determined now based on the arrival of speculative text).  Perhaps we need to update the `_assistant_is_responding` bookkeeping so that `_assistant_is_responding = True` starts a bit earlier, on the arrival of the first speculative text rather on the first audio chunk, so that `LLMFullResponseStartFrame` properly precedes the `LLMTextFrame`s.

markbackman: Good catch! We can push the `LLMFullResponseStartFrame` when the first `contentStart` for text/speculative is received. That should clean this up.

src/pipecat/services/aws/nova_sonic/llm.py
markbackman: Good call!

src/pipecat/services/aws/nova_sonic/llm.py
```diff
@@ -1232,11 +1203,12 @@ async def _handle_content_end_event(self, event_json):
...
-                if content.text_stage == TextStage.FINAL and stop_reason != "INTERRUPTED":
-                    if self._assistant_is_responding:
-                        # Text added to the ongoing assistant response
-                        await self._report_assistant_response_text_added(content.text_content)
+                if stop_reason != "INTERRUPTED":
+                    if content.text_stage == TextStage.SPECULATIVE:
+                        await self._report_llm_text_generated(content.text_content)
+            elif content.type == ContentType.AUDIO:
+                # Emit deferred TTSTextFrame after all audio chunks have been sent
+                await self._report_tts_text_spoken()
```
kompfner: Now that I'm looking at this name, it feels off...it hasn't been spoken yet. Maybe we can simplify to just `_report_llm_text_generated` and `_report_tts_text`?

src/pipecat/services/aws/nova_sonic/llm.py
```diff
@@ -569,49 +569,20 @@ async def _handle_input_audio_frame(self, frame: InputAudioRawFrame):
...
-                    # text that is quite delayed but not waiting so long that user text comes in
-                    # first and results in a bit of context message order scrambling.
-                    await asyncio.sleep(1.25)
-                self._assistant_is_responding = False
-                await self._report_assistant_response_ended()
-
-            self._handling_bot_stopped_speaking = False
+        if self._assistant_is_responding:
+            self._assistant_is_responding = False
+            await self._report_assistant_response_ended()
```
kompfner: Is it still the case that we can't really know that Nova Sonic has finished generating its response until the bot actually has stopped speaking? Otherwise, I'm wondering if there's some signal that we can listen to to call this a lot sooner...

Maybe this consideration is out of scope for this PR.

We also may not have a reliable signal to go off, because, IIRC, Nova Sonic didn't provide a "single assistant response generation complete" event.

Though...it might be worth quickly checking if the last speculative block in an assistant response is guaranteed to have `stopReason: "END_TURN" | "INTERRUPTED"` (and never `"PARTIAL_TURN"`). If so, that could be our reliable signal to fire `_report_assistant_response_ended` (which sends the LLM and TTS stopped frames).

kompfner (approved):
Looks good! One tiny naming nit, and another improvement suggestion (feel free to call that one out of scope, and one of us could take it in a follow-up PR)

src/pipecat/services/aws/nova_sonic/llm.py
kompfner: Whoops, sorry, above meant `_report_llm_text` and `_report_tts_text`, so they match.

markbackman: Updated the naming.

src/pipecat/services/aws/nova_sonic/llm.py
markbackman: Good idea! We do get:
```
Content end: {'contentEnd': {'completionId': '179efba6-b7ae-457a-974d-23206bb75f09', 'contentId': '0c2a130b-d05e-4335-9d48-0afcd2e72673', 'promptName': '03dccfed-a520-4ea4-bad9-c626ef88c25a', 'sessionId': '58ddd438-4600-415c-9951-5e0f27c8da8f', 'stopReason': 'END_TURN', 'type': 'AUDIO'}}
```

When all of the audio for a given turn has been received. We should be able to use that to more accurately reflect an end of turn. I'll make that update.

markbackman: An interruption looks different:
```
Content end: {'contentEnd': {'completionId': 'd754f891-f3fc-4cc8-9bc0-cff5a6753829', 'contentId': 'd8e4308f-fed5-4417-adbc-660f90474b5a', 'promptName': 'c1ba22af-fe1e-42f2-ad87-42cc4229f881', 'sessionId': 'cc1ff7db-19d3-43ef-a573-04e838edef09', 'stopReason': 'INTERRUPTED', 'type': 'TEXT'}}
```

We can handle this too.

kompfner (approved):
This service looks much more robust now!

PR #4046, opened 2026-03-16
Fix SonioxSTTService crash when language_hints contains plain strings…

### LINKED ISSUE #4045: SonioxSTTService: exception processing StartFrame: 'str' object has no attribute 'value'
### pipecat version

0.0.105

### Python version

3.13

### Operating System

Docker

### Issue description

```python
from enum import StrEnum


class Language(StrEnum):
    EN = "en"

def language_to_soniox_language(language: Language) -> str:
    """Pipecat Language enum uses same ISO 2-letter codes as Soniox, except with added regional variants.

    For a list of all supported languages, see: https://soniox.com/docs/speech-to-text/core-concepts/supported-languages
    """
    lang_str = str(language.value).lower()
    if "-" in lang_str:
        return lang_str.split("-")[0]
    return lang_str
```

Here we should remove `.value` cuz it StrEnum, and maybe we pass normal string which will raise the error in the logs below.

### Reproduction steps

```python

SonioxSTTService(
      api_key=settings.stt.soniox.SONIOX_API_KEY,
      settings=SonioxSTTSettings(
          language_hints=["en"],
          language_hints_strict=True,
          model="stt-rt-v4",
      )
)

```

### Expected behavior

<br>

### Actual behavior

<br>

### Logs

```shell
<LogRecord: pipecat.services.ai_service, 40, /usr/local/lib/python3.13/site-packages/pipecat/services/ai_service.py, 185, "2026-03-15 21:20:11.928 | ERROR    | pipecat.services.ai_service:_start:185 - SonioxSTTService#0: exception processing StartFrame#0: 'str' object has no attribute 'value'">
```

markbackman:
## Summary

- Fixed `SonioxSTTService` crash (`'str' object has no attribute 'value'`) when `language_hints` contains plain strings (e.g. `["en"]`) instead of `Language` enum values
- Refactored `language_to_soniox_language()` to use the `resolve_language()` + `LANGUAGE_MAP` pattern consistent with other services, with an explicit map of all ~60 Soniox-supported languages
- Fixed the same crash in `OpenAIRealtimeSTTService._language_to_code()` which also called `language.value` directly
- Fixed `resolve_language()` fallback path to use `str(language)` instead of `language.value`, making it robust for all services when an unmapped language is passed as a plain string

## Fixes

- Fixes #4045

### DIFF
--- a/changelog/4046.fixed.md
+++ b/changelog/4046.fixed.md
@@ -1,0 +1,1 @@
+Fixed `SonioxSTTService` and `OpenAIRealtimeSTTService` crash when language parameters contain plain strings instead of `Language` enum values.
--- a/changelog/4045.fixed.md
+++ b/changelog/4045.fixed.md
@@ -1,1 +1,0 @@
-Fixed `SonioxSTTService` crash when `language_hints` contains plain strings instead of `Language` enum values.
--- a/changelog/4045.fixed.md
+++ b/changelog/4045.fixed.md
@@ -1,0 +1,1 @@
+Fixed `SonioxSTTService` crash when `language_hints` contains plain strings instead of `Language` enum values.
--- a/src/pipecat/services/openai/stt.py
+++ b/src/pipecat/services/openai/stt.py
@@ -358,8 +358,8 @@ class OpenAIRealtimeSTTService(WebsocketSTTService):
         Returns:
             Two-letter ISO-639-1 language code.
         """
-        # Language.value is e.g. "en", "en-US", "fr", "zh".
-        return language.value.split("-")[0].lower()
+        # Language value is e.g. "en", "en-US", "fr", "zh".
+        return str(language).split("-")[0].lower()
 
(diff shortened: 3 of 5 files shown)


changelog/4046.fixed.md
```diff
@@ -0,0 +1 @@
+Fixed `SonioxSTTService` and `OpenAIRealtimeSTTService` crash when language parameters contain plain strings instead of `Language` enum values.
```
filipi87, line 1: Just confirming, is this fix still needed even after this [PR](https://github.com/pipecat-ai/pipecat/pull/4058) ?

filipi87 (approved):
LGTM.

changelog/4046.fixed.md
markbackman: Yes, the checks for language.value() are the problem when the language is string and not an Enum.

PR #4054, opened 2026-03-17
Add together STT and TTS services

blainekasten:


markbackman:
@blainekasten, we've changed enough with the base TTS service class and how settings are handled, that it's probably worth collaborating on my branch here:
https://github.com/pipecat-ai/pipecat/pull/3904

I've updated this branch to align with those recent changes.

In revisiting it today, I see that the TTS service is not returning any audio. I've poked around a bit and have written some standalone tests where I can't get audio to be returned in following examples in your docs. Was there an API change or perhaps this is something to do with my account or API key. Can you help?

blainekasten:
@markbackman sorry - let me put this in draft mode. I'm actively working on this

examples/foundational/07e-interruptible-together.py
```diff
@@ -55,12 +54,17 @@ async def run_bot(transport: BaseTransport, runner_args: RunnerArguments):
...
+        settings=TogetherTTSService.Settings(
+            voice="tara",
+        ),
+    )
 
     llm = TogetherLLMService(
         api_key=os.getenv("TOGETHER_API_KEY"),
         settings=TogetherLLMService.Settings(
-            model="Qwen/Qwen3.5-9B",
+            model="openai/gpt-oss-120b",
```
markbackman: I found the Qwen model to be really slow and have issues with producing inference. I've used the gpt-oss-120b model for another project and it's worked well. Seems to work well here too.

examples/transcription/transcription-together.py
```diff
@@ -0,0 +1,83 @@
+#
```
markbackman, line 1: This file lets you test STT as a standalone service. It displays TranscriptionFrames in the console.

src/pipecat/services/together/llm.py
```diff
@@ -56,7 +56,9 @@ def __init__(
             **kwargs: Additional keyword arguments passed to OpenAILLMService.
         """
         # 1. Initialize default_settings with hardcoded defaults
-        default_settings = self.Settings(model=model)
+        default_settings = self.Settings(
+            model="openai/gpt-oss-120b",
```
markbackman: Making `"openai/gpt-oss-120b"` the default. This is the pattern, where we initialize settings.

src/pipecat/services/together/stt.py
```diff
@@ -40,6 +41,9 @@
 from pipecat.utils.time import time_now_iso8601
 from pipecat.utils.tracing.service_decorators import traced_stt
 
+# Together requires 16 kHz 16-bit mono PCM input.
+_TOGETHER_SAMPLE_RATE = 16000
```
markbackman: Together only supports 16khz, so we're setting a constant and then using a resampler. This will help the service work in the event that a user sets a different sample rate via the PipelineParams.

src/pipecat/services/together/stt.py
```diff
@@ -50,55 +54,69 @@ class TogetherSTTSettings(STTSettings):
         language: Language of the audio input.
     """
 
-    model: str = "openai/whisper-large-v3"
-    language: Language = Language.EN
+    pass
```
markbackman, line 48: model and language are set in the base STTSettings / ServiceSettings classes. No need to respecify here.

src/pipecat/services/together/stt.py
```diff
@@ -50,55 +54,69 @@ class TogetherSTTSettings(STTSettings):
...
+            api_key="...",
+            settings=TogetherSTTService.Settings(
+                model="openai/whisper-large-v3",
+            ),
+        )
     """
 
-    _settings: TogetherSTTSettings
+    Settings = TogetherSTTSettings
+    _settings: Settings
```
markbackman, line 69: We're using `Settings` as the preferred class since you can use `ClassName.Settings`, making dev ex better.

src/pipecat/services/together/stt.py
```diff
@@ -50,55 +54,69 @@ class TogetherSTTSettings(STTSettings):
...
-            sample_rate: Audio sample rate (default: 16000). Together AI requires 16kHz input.
             base_url: The URL of the Together AI WebSocket API.
+            settings: Runtime-updatable settings for model and language configuration.
             ttfs_p99_latency: P99 latency from speech end to final transcript in seconds.
                 Override for your deployment. See https://github.com/pipecat-ai/stt-benchmark
             **kwargs: Additional arguments passed to the parent WebsocketSTTService.
         """
+        # Hardcoded defaults
+        default_settings = self.Settings(
+            model="openai/whisper-large-v3",
```
markbackman, line 93: Need to init all defaults for the store here.

src/pipecat/services/together/stt.py
```diff
@@ -108,6 +126,39 @@ def can_generate_metrics(self) -> bool:
         """
         return True
 
+    async def _update_settings(self, delta: STTSettings) -> dict[str, Any]:
```
markbackman: This allows for runtime updates, where supported.

src/pipecat/services/together/tts.py
```diff
@@ -63,40 +54,42 @@ class TogetherTTSService(WebsocketTTSService):
     Supports streaming synthesis with configurable voice and model options.
     """
 
-    _settings: TogetherTTSSettings
+    Settings = TogetherTTSSettings
```
markbackman, line 73: Same settings patterns here.

src/pipecat/services/together/tts.py
```diff
@@ -231,12 +249,22 @@ async def _ws_send(self, message: dict):
...
+
         Args:
-            context_id: Pipecat TTS context (unused for Together; required for
-                compatibility with :meth:`TTSService.on_turn_context_completed`).
+            context_id: Pipecat TTS context to flush.
         """
-        logger.trace(f"{self}: flushing audio (context_id={context_id})")
-        await self._ws_send({"type": "input_text_buffer.commit"})
+        ctx_id = context_id or self._context_id
+        if not ctx_id or not self.audio_context_available(ctx_id):
```
markbackman: Lots of context changes, which are now required. I applied the latest in this class.

markbackman (commented):
I pushed a number of changes to these classes and have a working 07e version.

I think the issue was three-fold:
1. The STT service seems like it takes a while to warm up before it will yield a transcript. Do you see this?
2. The Qwen model is _really_ slow to produce inference.
3. The STT is very sensitive and prone to outputting false positives. This is a class symptom of Whisper models—lots of "You" or "Thank you" produced. This is actually a pretty big problem for production use, so if you can tune this to reduce false positives, it will really help with adoption.

Also, I rebased on the latest main.

src/pipecat/services/together/stt.py
```diff
@@ -0,0 +1,452 @@
...
+            if self._websocket and self._websocket.state is State.OPEN:
+                return
+
+            url = (
+                f"{self._base_url}/realtime?intent=transcription"
+                f"&model={self._settings.model}"
+                f"&input_audio_format=pcm_s16le_16000"
+            )
+            headers = {
+                "Authorization": f"Bearer {self._api_key}",
```
markbackman, line 221: This previous had something about a beta header, which if you're using the OpenAI Realtime pattern is no longer needed. I removed it, but wanted to alert you of it.

src/pipecat/services/stt_latency.py
```diff
@@ -46,6 +46,7 @@
 SARVAM_TTFS_P99: float = 1.17
 SONIOX_TTFS_P99: float = 0.35
 SPEECHMATICS_TTFS_P99: float = 0.74
+TOGETHER_TTFS_P99: float = 2.028
```
markbackman: I wonder if this value is so high due to the warm up problem I think I observed in testing. SOTA services are around p50 at 0.3 sec with p99 around 0.4 sec. Ideally, the p99 latency is lower to remain competitive.

src/pipecat/services/together/stt.py
```diff
@@ -0,0 +1,396 @@
...
+                "OpenAI-Beta": "realtime=v1",
```
markbackman, line 222: Is this required?

src/pipecat/services/together/__init__.py
```diff
@@ -0,0 +1,15 @@
+#
```
markbackman: Remove all contents. The `DeprecatedModuleProxy` was removed in the 1.0.0 release.

examples/voice/voice-together.py
```diff
@@ -0,0 +1,123 @@
+#
```
markbackman, line 1: The pattern for this file has shifted. You might want to pull a voice example from main and just add your services to it.

examples/transcription/transcription-together.py
markbackman, line 1: Same here, the pattern is different. Pull from main and replace your service.

src/pipecat/services/together/stt.py
```diff
@@ -0,0 +1,397 @@
...
+from dataclasses import dataclass
+from typing import Any, AsyncGenerator, Optional
+
+from loguru import logger
+
+from pipecat.services.settings import STTSettings
+from pipecat.services.stt_latency import TOGETHER_TTFS_P99
+
+try:
+    from websockets.asyncio.client import connect as websocket_connect
```
markbackman: The `websockets` python package is now a core dependency, so remove the try/except here.

src/pipecat/services/together/tts.py
```diff
@@ -0,0 +1,446 @@
...
+import base64
+import json
+from dataclasses import dataclass, field
+from typing import Any, AsyncGenerator, Optional
+
+from loguru import logger
+
+from pipecat.services.settings import NOT_GIVEN, TTSSettings, _NotGiven
+
+try:
```
markbackman: `websockets` is a core dependency, so remove the try/except.

src/pipecat/services/together/stt.py
```diff
@@ -0,0 +1,397 @@
...
+This module provides a STT service using Together AI's WebSocket API:
+
+- ``TogetherSTTService``: WebSocket-based real-time STT using an
+  OpenAI-compatible realtime transcription endpoint.
+"""
+
+import base64
+import json
+from dataclasses import dataclass
+from typing import Any, AsyncGenerator, Optional
```
markbackman: `Optional` is no longer supported. We're applying the `"UP"` linting rule, so you'll need to run the linter to clean this up: `uv run ruff check --fix`. Same in other files.

markbackman (commented):
Few comments to start. I'll put in some more time soon.

markbackman (approved):
LGTM! Pushed some clean up changes. Tested and it works well. Congrats 👏

PR #4057, opened 2026-03-17
Fix stale state in user turn stop strategies between turns

### LINKED ISSUE #4053: Stale _text in TurnAnalyzerUserTurnStopStrategy causes 'one turn behind' bug
## Bug description

When a late `TranscriptionFrame` arrives between turns (after `push_aggregation` clears `_aggregation`), it sets `strategy._text` to stale content. On the next turn, `_handle_vad_user_started_speaking` does **not** reset `_text`, so `timeout(0)` fires with the stale `_text` and triggers a premature turn stop before the current `TranscriptionFrame` arrives.

**The cycle:**
1. Turn N ends → `push_aggregation` → `_aggregation` cleared
2. Late `TranscriptionFrame` arrives → accumulated in `_aggregation`, sets `_text`
3. Turn N+1 starts → `_aggregation` NOT cleared, `_text` NOT reset
4. VAD stop → turn analyzer COMPLETE → `timeout(0)` fires
5. Timeout sees non-empty `_text` → `trigger_user_turn_stopped` → pushes stale `_aggregation`
6. Strategy `reset()` clears `_turn_complete`
7. Current `TranscriptionFrame` arrives → accumulated but `_turn_complete=False` → no push
8. Current text becomes stale for turn N+2 → cycle repeats

**Contributing factors:**
- `_handle_vad_user_started_speaking` does NOT reset `self._text` (only `reset()` does, called after turn stop)
- `timeout = max(0, stt_p99 - vad_stop_secs)` = 0 when `stop_secs >= stt_p99` (common with Soniox where TTFS P99 = 0.35s and VAD stop_secs >= 0.35s)

## Suggested fix

Reset `_text` in `_handle_vad_user_started_speaking`:
```python
async def _handle_vad_user_started_speaking(self, frame):
    self._text = ""  # ← add this
    self._turn_complete = False
    ...
```

`_text` is only used as a gate condition in `_maybe_trigger_user_turn_stopped` (`if not self._text`). It doesn't carry content downstream — the actual transcription flows through the aggregator. Clearing it on turn start means "wait for this turn's transcript before allowing a stop".

## Production evidence

Two separate production calls (different languages, different code versions) show the identical pattern — every LLM call has the previous turn's transcription as the latest user message:

| LLM Call | STT transcribed | LLM context had (user msg) |
|----------|----------------|---------------------------|
| Call 3 | "John Smith" | "I'll wait on site" |
| Call 4 | "John Smith" (retry) | "John Smith" |
| Call 6 | "No thanks" | "John Smith" (retry) |
| Call 7 | "No thanks" (retry) | "No thanks" |

The pattern repeats indefinitely until the call ends.

## PR with fix and tests

#4051

markbackman:
## Context

The issue filed in #4053 highlights a case where state can bleed over between one turn to the next. This PR attempts to clean up the state between turns to ensure we don't have any state carrying over.

## Summary

- Fixed premature user turn stops caused by late transcriptions arriving between turns. When a `TranscriptionFrame` arrived after a turn ended but before the next one started, its text would persist as stale state into the next turn, triggering an immediate stop before the current turn's real transcript arrived.
- Stop strategies are now reset at both turn **start** and turn **stop**, ensuring no state leaks across turn boundaries. The `if self._user_turn: return` guard in `_trigger_user_turn_start` ensures this only fires once per turn, not on mid-turn VAD restarts (speak-pause-speak).
- Pending timeout tasks in `SpeechTimeoutUserTurnStopStrategy` and `TurnAnalyzerUserTurnStopStrategy` are now cancelled during `reset()` to prevent stale timeouts from firing after a turn boundary.

## Testing

```bash
uv run pytest tests/test_user_turn_stop_strategy.py tests/test_user_turn_controller.py -v
```

## Fixes

- Fixes #4053

### DIFF
--- a/changelog/4057.fixed.md
+++ b/changelog/4057.fixed.md
@@ -1,0 +1,1 @@
+- Fixed premature user turn stops caused by late transcriptions arriving between turns. A stale transcript from the previous turn could persist into the next turn and trigger a stop before the current turn's real transcript arrived. Stop strategies are now reset at both turn start and turn stop to prevent state from leaking across turn boundaries.
--- a/src/pipecat/turns/user_stop/speech_timeout_user_turn_stop_strategy.py
+++ b/src/pipecat/turns/user_stop/speech_timeout_user_turn_stop_strategy.py
@@ -64,6 +64,9 @@ class SpeechTimeoutUserTurnStopStrategy(BaseUserTurnStopStrategy):
         self._vad_user_speaking = False
         self._transcript_finalized = False
         self._vad_stopped_time = None
+        if self._timeout_task:
+            await self.task_manager.cancel_task(self._timeout_task)
+            self._timeout_task = None
 
     async def setup(self, task_manager: BaseTaskManager):
         """Initialize the strategy with the given task manager.
--- a/src/pipecat/turns/user_stop/turn_analyzer_user_turn_stop_strategy.py
+++ b/src/pipecat/turns/user_stop/turn_analyzer_user_turn_stop_strategy.py
@@ -68,6 +68,9 @@ class TurnAnalyzerUserTurnStopStrategy(BaseUserTurnStopStrategy):
         self._vad_user_speaking = False
         self._vad_stopped_time = None
         self._transcript_finalized = False
+        if self._timeout_task:
+            await self.task_manager.cancel_task(self._timeout_task)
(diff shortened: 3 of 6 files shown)


filipi87 (approved):
Nice catch! LGTM!

PR #4058, opened 2026-03-17
fix: resolve raw language strings through Language enum for proper service conversion

markbackman:
## Summary

- Fixed raw language strings like `"de-DE"` silently failing when passed to TTS/STT services. Previously, raw strings bypassed the `Language` enum resolution logic and were sent directly to the provider API (e.g. ElevenLabs expects `"de"`, not `"de-DE"`), causing no audio output with no error.
- Raw strings are now converted to `Language` enums first so they go through the same `resolve_language()` path as enum values, properly resolving regional codes to service-expected formats.
- Unrecognized language strings now log a warning instead of failing silently.
- The fix applies to both `TTSService` and `STTService` base classes, covering both init-time and runtime update paths.

## Testing

```bash
uv run pytest tests/test_service_language.py -v
```

23 tests covering all permutations of:
- Input types: `Language` enum (base), `Language` enum (regional), raw string (base), raw string (regional), unrecognized string, `None`
- Code paths: init-time settings, runtime `_update_settings()`
- Service types: TTS and STT


filipi87 (approved):
LGTM!

PR #4060, opened 2026-03-17
fix(openai): handle tool calls with empty/null arguments

### LINKED ISSUE #4059: Tool calls with empty/null arguments silently dropped (affects vLLM, Ollama, and other OpenAI-compatible providers)
### pipecat version

Latest main branch (also verified in 0.0.76+)

### Python version

3.12

### Operating System

Linux (CI) / macOS

### Issue description

When an LLM returns a streaming tool call with **no arguments** (`arguments=null` in the delta chunks), the tool call is silently dropped — no error, no warning. The tool handler is never invoked.

This happens in `BaseOpenAILLMService._process_context` ([base_llm.py](https://github.com/pipecat-ai/pipecat/blob/main/src/pipecat/services/openai/base_llm.py)):

1. **Line 544**: `if tool_call.function and tool_call.function.arguments:` — when `arguments` is `None`, this check is falsy, so nothing is accumulated. The local `arguments` variable stays as `""`.

2. **Line 561**: `if function_name and arguments:` — the empty string `""` is falsy, so the entire tool call execution block is skipped. `run_function_calls` is never called.

**OpenAI masks this bug** because it always sends `arguments: "{}"` (empty JSON object string) even for tools with zero parameters. But other OpenAI-compatible providers behave differently:

- **vLLM** (with Qwen, Llama, Mistral, etc.) omits `arguments` entirely when the tool schema has no required properties
- **Ollama** and other local providers may exhibit the same behavior

This is a significant issue because:
- Tools with no required parameters (e.g., `end_call`, `hang_up`, `get_status`) are common
- The failure is completely silent — no error logged, no exception raised
- It's hard to debug because the LLM *does* correctly identify and call the tool; pipecat just ignores it

### Reproduction steps

1. Serve any model via vLLM with `--enable-auto-tool-choice --tool-call-parser hermes`
2. Register a tool with no required parameters:
   ```python
   tools = [{"type": "function", "function": {"name": "end_call", "description": "End the call", "parameters": {"type": "object", "properties": {}, "required": []}}}]
   ```
3. Send a prompt that triggers the tool call (e.g., "End the call")
4. The streaming response includes `tool_calls` with `function.name = "end_call"` and `finish_reason = "tool_calls"`, but **no arguments chunks**
5. The tool handler is never invoked

### Expected behavior

Tool calls with no arguments should be treated as having `arguments = {}` (empty dict) and the registered handler should be invoked.

### Actual behavior

The tool call is silently dropped. `run_function_calls` is never called. No error or warning is logged.

### Suggested fix

Three lines in `base_llm.py`:

| Line | Before | After |
|------|--------|-------|
| 535 | `arguments_list.append(arguments)` | `arguments_list.append(arguments or "{}")` |
| 561 | `if function_name and arguments:` | `if function_name:` |
| 564 | `arguments_list.append(arguments)` | `arguments_list.append(arguments or "{}")` |

This defaults empty arguments to `"{}"` so `json.loads` produces an empty dict. No behavior change for OpenAI (which already sends `"{}"`). Fixes all OpenAI-compatible providers.

alpsencer:
Fixes #4059

## Problem

When an LLM returns a streaming tool call with no arguments (`arguments=null` in the delta chunks), the tool call is silently dropped. This affects all OpenAI-compatible providers that don't send `arguments: "{}"` for parameterless tools, including vLLM (with Qwen, Llama, Mistral), Ollama, and others.

**Root cause in `BaseOpenAILLMService._process_context`:**

1. `tool_call.function.arguments` is `None` → nothing accumulated → `arguments` stays `""`
2. `if function_name and arguments:` → `""` is falsy → tool call execution skipped entirely
3. No error or warning logged — completely silent failure

OpenAI masks this because it always sends `arguments: "{}"` even for tools with zero parameters.

## Fix

Three lines changed:

- **Line 535**: `arguments_list.append(arguments)` → `arguments_list.append(arguments or "{}")` (intermediate tool calls in multi-tool responses)
- **Line 561**: `if function_name and arguments:` → `if function_name:` (final tool call check)
- **Line 564**: `arguments_list.append(arguments)` → `arguments_list.append(arguments or "{}")` (final tool call append)

Empty arguments default to `"{}"` so `json.loads` produces `{}`. No behavior change for OpenAI or any provider that already sends `"{}"`.

## Testing

Verified with vLLM serving `Qwen3-VL-235B` with `--enable-auto-tool-choice --tool-call-parser hermes`. Before this fix, `end_call` tool calls were silently dropped. After, they execute correctly with empty arguments.

### DIFF
--- a/src/pipecat/services/openai/base_llm.py
+++ b/src/pipecat/services/openai/base_llm.py
@@ -532,7 +532,7 @@ class BaseOpenAILLMService(LLMService):
                     tool_call = chunk.choices[0].delta.tool_calls[0]
                     if tool_call.index != func_idx:
                         functions_list.append(function_name)
-                        arguments_list.append(arguments)
+                        arguments_list.append(arguments or "{}")
                         tool_id_list.append(tool_call_id)
                         function_name = ""
                         arguments = ""
@@ -558,10 +558,10 @@ class BaseOpenAILLMService(LLMService):
         # a registered handler. If so, run the registered callback, save the result to
         # the context, and re-prompt to get a chat answer. If we don't have a registered
         # handler, raise an exception.
-        if function_name and arguments:
+        if function_name:
             # added to the list as last function name and arguments not added to the list
             functions_list.append(function_name)
-            arguments_list.append(arguments)
+            arguments_list.append(arguments or "{}")
             tool_id_list.append(tool_call_id)
 
             function_calls = []


markbackman (approved):
Thanks for the well-written issue and clean fix.

PR #4063, opened 2026-03-17
fix: pass list-type Deepgram settings as lists instead of stringifying

markbackman:
List-valued settings like keyterm, keywords, search, redact, and replace were being converted to strings before being passed to the SDK connect() method. The SDK expects lists so its encode_query can produce repeated query params (keyterm=a&keyterm=b).


aconchillo (approved):

aconchillo:
LGTM

PR #4064, opened 2026-03-17
Add WakePhraseUserTurnStartStrategy

markbackman:
## Summary

- Added `WakePhraseUserTurnStartStrategy`, a user turn start strategy that gates interaction behind wake phrase detection. Supports a configurable inactivity timeout (re-requires the wake phrase after a period of silence) and a `single_activation` mode (requires the wake phrase before every turn, using the timeout as a keepalive window).
- Added `default_user_turn_start_strategies()` and `default_user_turn_stop_strategies()` helper functions for composing custom strategy lists.
- Deprecated `WakeCheckFilter` in favor of the new strategy and updated the `10-wake-phrase` foundational example.
- Extended `ProcessFrameResult` to stop strategies, enabling a stop strategy to short-circuit evaluation of subsequent strategies by returning `STOP`.
- Fixed `MinWordsUserTurnStartStrategy` including text below the word threshold in the output by resetting aggregation when the minimum word count is not met.

## Testing

- `uv run pytest tests/test_wake_phrase_user_turn_start_strategy.py`
- `uv run pytest` — all existing tests pass


src/pipecat/turns/user_turn_controller.py
```diff
@@ -161,7 +165,9 @@ async def process_frame(self, frame: Frame):
             await self._handle_transcription(frame)
 
         for strategy in self._user_turn_strategies.start or []:
-            await strategy.process_frame(frame)
+            result = await strategy.process_frame(frame)
+            if result in (ProcessFrameResult.TRIGGERED, ProcessFrameResult.BREAK):
+                break
```
aconchillo, line 171: It seems that triggered and break do the same thing. I'm wondering if we should return CONTINUE and STOP only.

src/pipecat/turns/types.py
```diff
@@ -0,0 +1,24 @@
+#
```
markbackman, line 1: I wasn't sure where to put this. I wanted to avoid duplication, so I created this new location. Open to other suggestions though.

src/pipecat/turns/user_turn_controller.py
markbackman: Resolved. We're just using CONTINUE and STOP now.

src/pipecat/turns/user_stop/base_user_turn_stop_strategy.py
```diff
@@ -89,14 +90,18 @@ async def reset(self):
         """Reset the strategy to its initial state."""
         pass
 
-    async def process_frame(self, frame: Frame):
+    async def process_frame(self, frame: Frame) -> Optional[ProcessFrameResult]:
```
aconchillo: We probably should not make it optional. Without Optional it should be backwards compatible, it's just the linting will fail, but at runtime it should be fine I think

src/pipecat/turns/user_start/base_user_turn_start_strategy.py
```diff
@@ -100,14 +102,18 @@ async def reset(self):
         """Reset the strategy to its initial state."""
         pass
 
-    async def process_frame(self, frame: Frame):
+    async def process_frame(self, frame: Frame) -> Optional[ProcessFrameResult]:
```
aconchillo: Same here. We probably don't want optional

src/pipecat/turns/types.py
aconchillo: That's fine I think. Or maybe `type.py` in case we want to add other stuff?

aconchillo (approved):

aconchillo:
Looks great. Some comments, but approving.

PR #4066, opened 2026-03-18
GradiumSTTService improvements

markbackman:
## Summary

- Improved `GradiumSTTService` transcription completeness by switching from silence-frame flushing to the flush API with text accumulation. Previously, trailing words could be dropped when the server's `flushed` response arrived before all text tokens were delivered.
- Added a transcript aggregation delay (100ms) after flush to capture trailing tokens before finalizing the transcription.
- Decoupled audio encoding from sample rate. The `encoding` parameter now takes a base type (`"pcm"`, `"wav"`, `"opus"`) and the sample rate is derived from the pipeline's `audio_in_sample_rate`, assembled dynamically via `input_format_from_encoding()`. This fixes the mismatch where `SAMPLE_RATE=24000` was passed to the base class while encoding defaulted to `"pcm_16000"`.
- Added `model_name` to the WebSocket setup message.
- Aligned VAD handling with base class patterns and removed duplicate reconnection logic.

## Breaking Changes

- `GradiumSTTService` `encoding` parameter default changed from `"pcm_16000"` to `"pcm"`. If you were passing `encoding="pcm_16000"` explicitly, change it to `encoding="pcm"` or omit it entirely.

## Testing

- Run `uv run python examples/foundational/07zf-interruptible-gradium.py`
- Verify complete utterances appear in LLM context (no dropped trailing words)
- Verify the WebSocket stays connected during pauses in speech

### DIFF
--- a/src/pipecat/services/gradium/stt.py
+++ b/src/pipecat/services/gradium/stt.py
@@ -46,6 +46,9 @@ except ModuleNotFoundError as e:
     raise Exception(f"Missing module: {e}")
 
 SAMPLE_RATE = 24000
+# Seconds to wait after a "flushed" message for trailing text tokens to arrive
+# before finalizing the transcription.
+TRANSCRIPT_AGGREGATION_DELAY = 0.1
 
 
 def language_to_gradium_language(language: Language) -> Optional[str]:
@@ -202,7 +205,6 @@ class GradiumSTTService(WebsocketSTTService):
         # and pushed as a TranscriptionFrame.
         self._accumulated_text: list[str] = []
         self._flush_counter = 0
-        self._transcript_aggregation_delay = 0.1  # seconds to wait after flushed
         self._transcript_aggregation_task: Optional[asyncio.Task] = None
 
     def can_generate_metrics(self) -> bool:
@@ -384,6 +386,9 @@ class GradiumSTTService(WebsocketSTTService):
             await self.cancel_task(self._transcript_aggregation_task)
             self._transcript_aggregation_task = None
 
+        self._accumulated_text.clear()
+        self._flush_counter = 0
+
         if self._receive_task:
             await self.cancel_task(self._receive_task)
             self._receive_task = None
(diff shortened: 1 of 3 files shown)


src/pipecat/services/gradium/stt.py
```diff
@@ -190,10 +196,14 @@ def __init__(
...
-        # These values are used for flushing transcription.
-        self._delay_in_frames = 0
-        self._frame_size = 0
+        # Accumulates text fragments within a turn. Each "text" message
+        # appends to this list. On "flushed" a short aggregation delay
+        # allows trailing tokens to arrive before the full text is joined
+        # and pushed as a TranscriptionFrame.
+        self._accumulated_text: list[str] = []
+        self._flush_counter = 0
+        self._transcript_aggregation_delay = 0.1  # seconds to wait after flushed
```
filipi87: Is this something we would like to allow users to change ? Otherwise, I think this could be a constant.

src/pipecat/services/gradium/stt.py
```diff
@@ -190,10 +196,14 @@ def __init__(
...
-        # Set from the ready message when connecting to the service.
-        # These values are used for flushing transcription.
-        self._delay_in_frames = 0
-        self._frame_size = 0
+        # Accumulates text fragments within a turn. Each "text" message
+        # appends to this list. On "flushed" a short aggregation delay
+        # allows trailing tokens to arrive before the full text is joined
+        # and pushed as a TranscriptionFrame.
+        self._accumulated_text: list[str] = []
+        self._flush_counter = 0
```
filipi87, line 241: Do we need to reset these values when we disconnect the webSocket? For example, in case of a reconnection.

filipi87 (approved):
Looks good. I just added a few possible improvements, but nothing major.

src/pipecat/services/gradium/stt.py
markbackman: Yes, updating.

src/pipecat/services/gradium/stt.py
markbackman: Yup!

PR #4067, opened 2026-03-18
fix: default thinking config for Gemini 3+ Flash models

### LINKED ISSUE #3993: GoogleLLMService: apply model-aware default thinking config for Gemini 3+
### pipecat version

0.0.104

### Python version

3.12

### Operating System

macOS

### Issue description

`GoogleLLMService` currently applies a default thinking override only for `gemini-2.5-flash*` via `_maybe_unset_thinking_budget()` in `pipecat/services/google/llm.py`.

That works for Gemini 2.5 Flash, but Gemini 3+ uses a different thinking surface (`thinking_level`, not `thinking_budget`). If a caller uses `GoogleLLMService` or `GoogleVertexLLMService` with a Gemini 3 Flash model and does not pass an explicit `params.thinking`, Pipecat leaves `thinking_config` unset and the model falls back to provider defaults.

For latency-sensitive voice use cases this is surprising because:

- `gemini-2.5-flash*` gets an automatic low-latency default (`thinking_budget=0`)
- `gemini-3*` gets no automatic default at all, even though the model family requires a different control surface

This is especially easy to hit now that newer Gemini 3 / 3.1 Flash models are becoming default choices.

Related Google docs:
- https://ai.google.dev/gemini-api/docs/thinking

### Reproduction steps

1. Initialize `GoogleLLMService` or `GoogleVertexLLMService` with a Gemini 3 Flash model, for example `gemini-3.1-flash-lite-preview`.
2. Do not pass `params.thinking`.
3. Inspect the generation params or run a normal call.
4. Observe that Pipecat does not populate `thinking_config` at all for Gemini 3 models.

Minimal example:

```python
from pipecat.services.google.llm_vertex import GoogleVertexLLMService

llm = GoogleVertexLLMService(
    credentials_path="...",
    project_id="...",
    location="europe-west2",
    model="gemini-3.1-flash-lite-preview",
    system_instruction="You are a voice assistant.",
)

params = llm._build_generation_params(system_instruction="x", tools=[], tool_config=None)
llm._maybe_unset_thinking_budget(params)
print(params)
# No thinking_config is added for Gemini 3 models.
```

### Expected behavior

If the caller did not specify `params.thinking`, Pipecat should apply model-aware defaults:

- `gemini-2.5-flash*` -> `thinking_budget=0`, `include_thoughts=False`
- `gemini-3-flash*` / `gemini-3.1-flash*` -> `thinking_level="minimal"`, `include_thoughts=False`
- do not override explicit user-provided thinking config
- probably leave Pro-family models untouched unless there is a clear safe default

### Actual behavior

Only the Gemini 2.5 Flash family gets a default low-thinking config. Gemini 3+ models get provider-default thinking behavior unless every application remembers to configure it explicitly.

### Suggested fix

Replace `_maybe_unset_thinking_budget()` with a model-aware default-thinking helper, for example:

- detect the model family
- apply the appropriate control surface for that family
- skip entirely when `thinking_config` is already present

This would make latency behavior much more predictable across Gemini model upgrades.

omChauhanDev:

Fixes #3993 

- Gemini 2.5 Flash gets a default `thinking_budget=0` for low latency, but Gemini 3+ Flash uses a different API surface (`thinking_level`) & was getting no default at all, falling back to provider defaults.         
                                            
- Now Gemini 3.x Flash models get `thinking_level="minimal"` automatically when no explicit thinking config is provided.

### DIFF
--- a/src/pipecat/services/google/llm.py
+++ b/src/pipecat/services/google/llm.py
@@ -995,18 +995,20 @@ class GoogleLLMService(LLMService):
 
     def _maybe_unset_thinking_budget(self, generation_params: Dict[str, Any]):
         try:
-            # There's no way to introspect on model capabilities, so
-            # to check for models that we know default to thinkin on
-            # and can be configured to turn it off.
-            if not self._settings.model.startswith("gemini-2.5-flash"):
-                return
-            # If we have an image model, we don't use a budget either.
+            # If we have an image model, we don't apply a thinking default.
             if "image" in self._settings.model:
                 return
             # If thinking_config is already set, don't override it.
             if "thinking_config" in generation_params:
                 return
-            generation_params.setdefault("thinking_config", {})["thinking_budget"] = 0
+            # Apply model-aware low-latency thinking defaults.
+            # Gemini 2.5 Flash: disable thinking via thinking_budget.
+            # Gemini 3+ Flash: use minimal thinking via thinking_level.
+            model = self._settings.model
+            if model.startswith("gemini-2.5-flash"):
+                generation_params["thinking_config"] = {"thinking_budget": 0}
+            elif model.startswith("gemini-3") and "flash" in model:
+                generation_params["thinking_config"] = {"thinking_level": "minimal"}
         except Exception as e:
             logger.error(f"Failed to unset thinking budget: {e}")
 
--- a/changelog/4067.fixed.md
+++ b/changelog/4067.fixed.md
@@ -1,0 +1,1 @@
+- GoogleLLMService now applies a low-latency thinking default (`thinking_level="minimal"`) for Gemini 3+ Flash models.


src/pipecat/services/google/llm.py
```diff
@@ -995,18 +995,20 @@ def _build_generation_params(
...
             if "thinking_config" in generation_params:
                 return
-            generation_params.setdefault("thinking_config", {})["thinking_budget"] = 0
+            # Apply model-aware low-latency thinking defaults.
+            # Gemini 2.5 Flash: disable thinking via thinking_budget.
+            # Gemini 3+ Flash: use minimal thinking via thinking_level.
+            model = self._settings.model
+            if model.startswith("gemini-2.5-flash"):
+                generation_params["thinking_config"] = {"thinking_budget": 0}
+            elif model.startswith("gemini-3") and "flash" in model:
```
kompfner, line 1010: Technically this more general check (as opposed to explicitly checking for the string "gemini-3-flash") would cause use to set this thinking config unnecessarily in "gemini-3.1-flash-lite" (where minimal thinking is default), *but* I think it's absolutely worth it for future-proofing. So, good call 👍.

kompfner (approved):
Thanks! Nice improvement.

Issue #4027, opened 2026-03-13
ElevenLabs error causes permanent TTS pipeline deadlock in specific timing conditions


skheyfets-asapp:
### pipecat version

0.0.105

### Python version

3.12.11

### Operating System

MacOS 15.7.4

### Issue description

Under adverse network conditions with high latency on the Internet connection the following sequence of events results in ElevenLabsTTSService pausing the frame processing of anything behind the TTSSpeakFrame . Several things have to happen (observed once, with fully running pipline) under right conditions.

**Production pipeline**
```python   
    pipeline = Pipeline(
        [
            ws_transport.input(), # fastapi websocket
            stt,
            user_aggregator,
            llm,  # LLM
            tts,  # TTS
            interruption_processor, # custom processor that pushes all frames in all directions, just tracks interruptions between BotStartedSpeaking and BotStoppedSpeaking events
            ws_transport.output(),
            assistant_aggregator,
        ]
    )

    task = PipelineTask(
        pipeline,
        params=PipelineParams(
            audio_in_sample_rate=16000,
            enable_metrics=True,
            enable_usage_metrics=True,
        ),
    )
```

**Log from production**
```json
{"timestamp": "2026-03-12T23:07:02.207720Z", "level": "debug", "message": "OpenAILLMService#0 Calling function [send_customer_request_or_update:call_nGE6SxeSt3dD0S7IaugB4ILC] with arguments {'customer_request_or_update': 'Customer wants a quote for insuring a motorcycle.'}", "location": "pipecat.services.llm_service:_run_function_call:841"}
{"timestamp": "2026-03-12T23:07:02.208852Z", "level": "debug", "message": "Sending customer request to Supervisor API: 'Customer wants a quote for insuring a motorcycle.'", "location": "my_agent.functions.manager:callback:106"}
{"timestamp": "2026-03-12T23:07:02.211566Z", "level": "debug", "message": "LLMAssistantAggregator#0 FunctionCallsStartedFrame: ['send_customer_request_or_update:call_nGE6SxeSt3dD0S7IaugB4ILC']", "location": "pipecat.processors.aggregators.llm_response_universal:_handle_function_calls_started:1015"}
[...]

PR #4071, opened 2026-03-18
Fixed the ordering of `_maybe_pause_frame_processing` call in `TTSService`

### LINKED ISSUE #4041: ElevenLabs TTS: WebSocket context limit (5) exceeded during rapid tool-call turns
## Summary

When the LLM produces rapid tool-call chains (e.g. function call -> instant response -> new LLM turn -> new TTS context), ElevenLabs closes the WebSocket with:

```
received 1008 (policy violation) Maximum simultaneous contexts per WebSocket connection exceeded (5).
Please close an existing context before opening a new one.
```

## Root Cause

Each `LLMFullResponseStartFrame` allocates a **new ElevenLabs context** (`tts_service.py:687`), but the previous context is only closed server-side in `on_audio_context_completed` — **after the audio finishes playing**. During fast tool-call flows the new context is opened while the old one is still playing, and the count quickly reaches the ElevenLabs limit of 5 concurrent contexts per WebSocket.

When several contexts overlap, downstream aggregation can also interleave text from different contexts and scramble the final assistant message on larger responses.

### Reproduction timeline (from real logs)

| Time | Event | Open contexts |
|------|-------|:---:|
| 17:46:33.642 | cleanup previous context | 1 |
| 17:46:37.169 | `run_tts` "Отлично." (new LLM turn after `record_tire_preferences`) | 2 |
| 17:46:37.253 | `run_tts` "Тогда ищу летние шины…" (same turn, reused ctx) | 2 |
| 17:46:37.372 | `run_tts` "Без конкретных предпочтений…" (same turn) | 2 |
| 17:46:37.373 | `run_tts` "Всё верно?" (same turn, flush) | 2 |
| 17:46:39.755 | cleanup for the above turn | 1 |
| 17:46:48.707 | user answers -> LLM turn -> tool call `search_catalog` | — |
| 17:46:50.740 | `run_tts` for post-tool LLM response (new context) | accumulates |
| 17:46:50.742 | `run_tts` second sentence | accumulates |
| 17:46:51.714 | **1008 policy violation** — limit exceeded | >5 |

The tool call (`search_catalog`) returns instantly, so the LLM fires a new response while previous contexts are still draining audio.

## Environment

- pipecat-ai 0.0.105
- ElevenLabs Creator plan (5 concurrent contexts per WebSocket)
- `reuse_context_id_within_turn = True` (default) — contexts are reused within a single LLM turn, but **not across turns**

## Working Mitigation Strategies

The earlier idea of simply enforcing a local `max_concurrent_contexts` / `get_audio_contexts()` cap was not sufficient for us. Pipecat's local audio-context map can drift from what ElevenLabs still counts as open during reconnects, delayed final messages, and ghost flushes.

### 1. Context-limited WebSocket wrapper

- Keep a small bounded number of server-side contexts on one WebSocket.
- Track context closure until the server-finalized event, not only until local playback finishes.
- On overflow, explicitly close an older context and wait for server final before freeing the slot.
- Treat reconnect as a full reset and skip `flush_audio` for untracked / ghost contexts.

Trade-off: preserves more overlap and lower latency, but the lifecycle bookkeeping is more complex.

### 2. Strict lifecycle WebSocket wrapper

- Allow only one WebSocket/context lifecycle at a time.
- Before switching to a new context, disconnect the WebSocket first.
- On context completion or interruption, close the active ElevenLabs context and disconnect the socket.
- If Pipecat has already removed the local audio context, reset the stale active context before disconnect so cleanup does not act on a context that no longer exists locally.

Trade-off: simpler and more robust; it also avoids cross-context text scrambling, but may add a small amount of TTS latency between turns.

filipi87:
## Summary
- Fixed the ordering of `_maybe_pause_frame_processing` call in `TTSService`. It now runs after flushing remaining text on `LLMFullResponseEndFrame`/`EndFrame`, not before.                                                                                           
- Previously, frame processing was paused before remaining TTS text was flushed, which could cause audio overlap or dropped content when ElevenLabs (and other TTS services) hit context limits under rapid consecutive turns 

Fixes #4041, #4027

### DIFF
--- a/examples/foundational/07d-interruptible-elevenlabs.py
+++ b/examples/foundational/07d-interruptible-elevenlabs.py
@@ -5,13 +5,19 @@
 #
 
 
+import asyncio
 import os
 
 from dotenv import load_dotenv
 from loguru import logger
 
 from pipecat.audio.vad.silero import SileroVADAnalyzer
-from pipecat.frames.frames import LLMRunFrame
+from pipecat.frames.frames import (
+    LLMFullResponseEndFrame,
+    LLMFullResponseStartFrame,
+    LLMRunFrame,
+    TextFrame,
+)
 from pipecat.pipeline.pipeline import Pipeline
 from pipecat.pipeline.runner import PipelineRunner
 from pipecat.pipeline.task import PipelineParams, PipelineTask
@@ -102,6 +108,16 @@ async def run_bot(transport: BaseTransport, runner_args: RunnerArguments):
         # Kick off the conversation.
         context.add_message({"role": "user", "content": "Please introduce yourself to the user."})
         await task.queue_frames([LLMRunFrame()])
+        # Rapid consecutive assistant turns to try to force the ElevenLabs maximum context limit exceeded error.
+        #for idx in range(1, int(10) + 1):
+        #    await task.queue_frames(
+        #        [
+        #            LLMFullResponseStartFrame(),
+        #            TextFrame(f"✓ [{idx}] test"),
+        #            LLMFullResponseEndFrame(),
+        #        ],
+        #    )
+        #    await asyncio.sleep(0)
 
     @transport.event_handler("on_client_disconnected")
     async def on_client_disconnected(transport, client):
--- a/examples/foundational/07d-interruptible-elevenlabs.py
+++ b/examples/foundational/07d-interruptible-elevenlabs.py
@@ -5,19 +5,13 @@
 #
 
 
-import asyncio
 import os
 
 from dotenv import load_dotenv
 from loguru import logger
 
 from pipecat.audio.vad.silero import SileroVADAnalyzer
-from pipecat.frames.frames import (
-    LLMFullResponseEndFrame,
-    LLMFullResponseStartFrame,
-    LLMRunFrame,
-    TextFrame,
-)
+from pipecat.frames.frames import LLMRunFrame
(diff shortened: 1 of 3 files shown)


examples/foundational/07d-interruptible-elevenlabs.py
```diff
@@ -102,6 +108,16 @@ async def on_client_connected(transport, client):
...
+        # Rapid consecutive assistant turns to try to force the ElevenLabs maximum context limit exceeded error.
+        #for idx in range(1, int(10) + 1):
+        #    await task.queue_frames(
+        #        [
+        #            LLMFullResponseStartFrame(),
+        #            TextFrame(f"✓ [{idx}] test"),
+        #            LLMFullResponseEndFrame(),
+        #        ],
+        #    )
+        #    await asyncio.sleep(0)
```
filipi87: This is all that was needed to reproduce the issue on `main`. 

Thank you, @yydrift-code for the repro case.

filipi87: TODO: remove this before merging this PR.

filipi87:
This PR fixes: https://github.com/pipecat-ai/pipecat/issues/4041 and https://github.com/pipecat-ai/pipecat/issues/4027

markbackman (approved):
LGTM!

Reminder to remove the lines from 07d before merging.

PR #4073, opened 2026-03-18
Suppress InvalidState log spam from audio mixer during interruptions in LiveKit transport

joachimchauvet:
When using an audio mixer with the LiveKit transport, the with_mixer() generator continuously yields frames (background audio mixed with silence) even when no TTS audio is queued. During interruptions, the audio task is cancelled and recreated, but there is a brief window where the native LiveKit AudioSource rejects capture_frame() with an InvalidState error. This produced dozen of ERROR log lines per interruption in some cases.

These errors are transient and harmless — the mixer produces a new frame within milliseconds — so we silently drop InvalidState errors while still logging other unexpected errors.


markbackman (approved):
LGTM. Thank you.

PR #3164, opened 2025-11-28
OpenAI Responses API LLM service implementation

vmfullyfaltu:
## Overview
Complete migration from OpenAI Chat Completions API to the new Responses API with stateful conversation support and incremental mode implementation.

## Motivation
The OpenAI Responses API provides several advantages over the Chat Completions API:
- **Stateful conversations** with server-side storage (when `store=True`)
- **Incremental updates** using `previous_response_id` (90% token savings)
- **Event-based streaming** with `ResponseStreamEvent`
- **Improved tool calling** support
- **Better multi-turn** conversation handling

### Core Service: `OpenAIResponsesLLMService`
A new LLM service that inherits from `BaseOpenAILLMService` and implements the Responses API:

```python
class OpenAIResponsesLLMService(BaseOpenAILLMService):
    """OpenAI Responses API LLM service implementation.
    
    Uses client.responses.create() instead of client.chat.completions.create()
    """
```

#### 1. API Endpoint Migration
- **Old**: `client.chat.completions.create()`
- **New**: `client.responses.create()`

#### 2. Message Format Conversion
Converts Chat Completions message format to Responses API input format:

**Chat Completions format:**
```python
{"role": "user", "content": "Hello"}
```

**Responses API format:**
```python
{
    "type": "message",
    "role": "user", 
    "content": [{"type": "input_text", "text": "Hello"}]
}
```

Key conversions handled:
- `"text"` → `"input_text"` (user messages)
- `"text"` → `"output_text"` (assistant messages)
- `"image_url"` → `"input_image"` (with base64 data extraction)
- Tool calls: Nested structure → Flat structure with `call_id` at top level
- Tool results: Added as `function_call_output` items

#### 3. Stateful Conversations (`store=True`)
Enables server-side conversation storage with incremental updates:

```python
params = OpenAIResponsesLLMService.InputParams(
    temperature=0.7,
    store=True,  # Enable stateful mode
    metadata={"call_sid": "CA1234"}
)
```

Benefits:
- **90% token savings**: Send only new messages, not entire history
- **50% latency reduction**: Smaller requests, faster processing
- **Metadata tracking**: Associate conversations with business context

#### 4. Incremental Mode
Implements robust context reset detection using **dual-signal detection**:

**Tracking Dictionaries:**
```python
self._response_ids: Dict[int, str] = {}           # Maps context_id → response_id
self._sent_message_counts: Dict[int, int] = {}    # Tracks messages sent per context
[...]

PR #4074, opened 2026-03-18
feat: add OpenAI Responses API LLM service

kompfner:
## Summary

- Added `OpenAIResponsesLLMService`, a new LLM service using the OpenAI Responses API. Supports streaming text, function calling, usage metrics, `service_tier`, and out-of-band inference (e.g. context summarization). Works with the universal `LLMContext` and `LLMContextAggregatorPair`.
- Added `OpenAIResponsesLLMAdapter` to convert LLMContext messages to Responses API input items (system→developer role, tool_calls→function_call items, tool→function_call_output items, multimodal content conversion, and tools schema flattening).
- Sets `store=False` by default since we send full conversation history as input items and don't use `previous_response_id`.
- Added foundational examples:
  - `07-interruptible-openai-responses.py` — basic voice conversation
  - `12-describe-image-openai-responses.py` — image input
  - `14-function-calling-openai-responses.py` — function calling
  - `14d-function-calling-openai-responses-video.py` — video + function calling
  - `20a-persistent-context-openai-responses.py` — persistent context
  - `55zi-update-settings-openai-responses-llm.py` — runtime settings update
- Added eval entries for all applicable examples.

## Testing

- `uv run pytest tests/test_openai_responses_adapter.py` — 19 unit tests for adapter conversion logic
- `uv run pytest tests/test_run_inference.py -k openai_responses` — 5 run_inference tests (real adapter, mocked HTTP client)
- `uv run pytest` — full test suite (no regressions)
- `uv run python scripts/evals/run-release-evals.py --pattern "openai-responses"` — evals for all examples

---

## Follow-up work

This PR is intended to get `OpenAIResponsesLLMService` to rough parity with `OpenAILLMService` as a starting point. It's already usable and useful and familiar.

But, of course, the Responses API is not the same as the Completions API! Things to still follow up on:

- [x] A Websocket variant or mode
- [ ] New settings
  - [x] `truncation` (UPDATE: deprecated)
  - [ ] `parallel_tool_calls`
  - [ ] `reasoning`
  - [x] `user` (UPDATE: deprecated)
- [ ] What we should do re invocation patterns that leverage OpenAI server-side convo storage (how to expose them, how to handle them in a scenario with service switching, etc)
  - [ ] `previous_response_id`
  - [ ] `conversation_id`
- [ ] How we should allow users to provide new Responses API `ResponseInputItem`s in conversation history, in a way that doesn't break context compatibility with other LLM services in the case of service switching
  - [ ] `LLMSpecificMessage` is the obvious initial way...
  - [ ] ...but do we want to stay wedded to the aging Completions message type as the "standard" across all services? Maybe it still makes sense for now, as it _has_ ingrained itself as somewhat of a standard...
- [ ] New features
  - [ ] `tool_search`
  - [ ] computer use
  - [ ] stored prompts
  - [ ] ...
- [ ] An [Azure Responses](https://learn.microsoft.com/en-us/azure/foundry/openai/how-to/responses?tabs=python-key) LLM Service?

### DIFF
--- a/src/pipecat/adapters/services/open_ai_responses_adapter.py
+++ b/src/pipecat/adapters/services/open_ai_responses_adapter.py
@@ -79,9 +79,19 @@ class OpenAIResponsesLLMAdapter(BaseLLMAdapter[OpenAIResponsesLLMInvocationParam
             # message when instructions are provided. Contexts that worked with
             # OpenAILLMService (system_instruction + empty messages) need the
             # instructions converted to an initial developer message.
-            # NOTE: once we support `previous_response_id`, we need to revisit
+            #
+            # NOTE: if/when we support `previous_response_id`, we'll need to revisit
             # this logic, as it'll be legit to provide instructions without input
-            # items if `previous_response_id` is provided.
+            # items if `previous_response_id` is provided. Though...OpenAI's docs +
+            # ChatGPT suggests that `previous_response_id` is primarily for
+            # development convenience, not performance (other than minor bandwidth
+            # savings from not transferring the full context), as the model still
+            # processes the full context from the previous response. The tradeoff
+            # of using `previous_response_id` is that it requires enabling OpenAI-side
+            # 30-day conversation storage (meaning we couldn't do `store=False`
+            # in the API call), which may not be desirable for all users. So,
+            # my guess is we won't need to support `previous_response_id` in the
+            # immediate future.
             if not input_items:
                 params["input"] = [{"role": "developer", "content": system_instruction}]
             else:
--- a/src/pipecat/adapters/services/open_ai_responses_adapter.py
+++ b/src/pipecat/adapters/services/open_ai_responses_adapter.py
@@ -79,6 +79,9 @@ class OpenAIResponsesLLMAdapter(BaseLLMAdapter[OpenAIResponsesLLMInvocationParam
             # message when instructions are provided. Contexts that worked with
             # OpenAILLMService (system_instruction + empty messages) need the
             # instructions converted to an initial developer message.
+            # NOTE: once we support `previous_response_id`, we need to revisit
+            # this logic, as it'll be legit to provide instructions without input
+            # items if `previous_response_id` is provided.
             if not input_items:
                 params["input"] = [{"role": "developer", "content": system_instruction}]
             else:
--- a/src/pipecat/adapters/services/open_ai_responses_adapter.py
+++ b/src/pipecat/adapters/services/open_ai_responses_adapter.py
@@ -80,18 +80,16 @@ class OpenAIResponsesLLMAdapter(BaseLLMAdapter[OpenAIResponsesLLMInvocationParam
             # OpenAILLMService (system_instruction + empty messages) need the
             # instructions converted to an initial developer message.
             #
-            # NOTE: if/when we support `previous_response_id`, we'll need to revisit
-            # this logic, as it'll be legit to provide instructions without input
-            # items if `previous_response_id` is provided. Though...OpenAI's docs +
-            # ChatGPT suggests that `previous_response_id` is primarily for
-            # development convenience, not performance (other than minor bandwidth
-            # savings from not transferring the full context), as the model still
-            # processes the full context from the previous response. The tradeoff
-            # of using `previous_response_id` is that it requires enabling OpenAI-side
-            # 30-day conversation storage (meaning we couldn't do `store=False`
-            # in the API call), which may not be desirable for all users. So,
-            # my guess is we won't need to support `previous_response_id` in the
-            # immediate future.
+            # NOTE: if/when we support `previous_response_id` and/or
+            # `conversation_id`, we'll need to revisit this logic, as it'll
+            # be legit to provide instructions without input items. Worth
+            # noting that OpenAI's docs suggest these parameters are primarily
+            # for development convenience rather than performance (the model
+            # still processes the full context), and come with the tradeoff
+            # of requiring OpenAI-side 30-day conversation storage, which may
+            # not be desirable for many users. But it could give folks an easy
+            # way to store/switch between conversations without needing to
+            # manage that storage themselves.
             if not input_items:
                 params["input"] = [{"role": "developer", "content": system_instruction}]
             else:
--- a/src/pipecat/adapters/services/open_ai_responses_adapter.py
+++ b/src/pipecat/adapters/services/open_ai_responses_adapter.py
@@ -246,6 +246,9 @@ class OpenAIResponsesLLMAdapter(BaseLLMAdapter[OpenAIResponsesLLMInvocationParam
                     }
                 )
             else:
-                # Pass through unknown types as-is
+                # Pass through other types as-is. Note: "input_audio" is not
+                # yet supported by the Responses API (coming soon per OpenAI
+                # docs) but the LLMContext format already matches the expected
+                # shape, so it should work once support is enabled.
                 result.append(part)
         return result
--- a/src/pipecat/adapters/services/open_ai_responses_adapter.py
+++ b/src/pipecat/adapters/services/open_ai_responses_adapter.py
@@ -1,0 +1,240 @@
+#
+# Copyright (c) 2024-2026, Daily
+#
+# SPDX-License-Identifier: BSD 2-Clause License
+#
+
+"""OpenAI Responses API adapter for Pipecat."""
+
+import copy
+from typing import Any, Dict, List, Optional, TypedDict
+
+from loguru import logger
+from openai._types import NotGiven as OpenAINotGiven
+from openai.types.responses import FunctionToolParam, ResponseInputItemParam
+
+from pipecat.adapters.base_llm_adapter import BaseLLMAdapter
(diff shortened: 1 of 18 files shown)


src/pipecat/adapters/services/open_ai_responses_adapter.py
```diff
@@ -0,0 +1,253 @@
...
+            "tools": self.from_standard_tools(context.tools),
+        }
+
+        if system_instruction:
+            # Compatibility: The Responses API requires at least one input
+            # message when instructions are provided. Contexts that worked with
+            # OpenAILLMService (system_instruction + empty messages) need the
+            # instructions converted to an initial developer message.
+            #
+            # NOTE: if/when we support `previous_response_id`, we'll need to revisit
```
kompfner: Just wanted to call out this bit of reasoning, for reviewers

vmfullyfaltu:
@kompfner if this helps you in your development

https://github.com/pipecat-ai/pipecat/pull/3164

markbackman (commented):
This looks really good! I ran the examples and all seems well. I also had Claude perform a review, which flagged a few things. Up to you about what you want to include.

markbackman (commented):
## Review

Overall this is a well-structured PR that cleanly ports the key functionality from `BaseOpenAILLMService` (Chat Completions) to a new `OpenAIResponsesLLMService` (Responses API). The adapter pattern is solid, test coverage is good, and the code follows existing codebase conventions. A few items to flag:

---

### Bugs / Issues

**1. `run_inference` double-applies instructions for empty context + explicit `system_instruction`**
`responses/llm.py:377-397`

When `run_inference(context, system_instruction="X")` is called with an empty context:
1. The adapter sees empty input + system_instruction, creates `input=[{"role": "developer", "content": "X"}]`, no `instructions` key
2. `_build_response_params` passes that through
3. Then line 389 fires: `params["instructions"] = system_instruction` -- adding it again

Result: the API receives **both** a developer message and an `instructions` param with the same content. Functionally probably harmless (OpenAI will handle it), but it is a semantic duplication the adapter was trying to avoid.

**Fix**: Either skip the line 389 override when `params["input"]` already contains the developer-message fallback, or restructure so the adapter empty-input logic and the run_inference override are coordinated.

**2. `_full_model_name` set without initialization or accessor**
`responses/llm.py:295`

`self._full_model_name = model` is set directly as an instance attribute. In `BaseOpenAILLMService`, this is initialized in `__init__` and has proper `get_full_model_name()`/`set_full_model_name()` methods. `OpenAIResponsesLLMService` extends `LLMService` directly, so these do not exist.

This works at runtime because the tracing decorator (`service_decorators.py:55`) uses `getattr(service, "_full_model_name", None)`, but it is fragile. Consider initializing in `__init__` and/or adding the accessors, or adding this to the `LLMService` base class.

**3. Example inconsistency: `14-function-calling-openai-responses.py:151-153`**

```python
context.add_message(
    {"role": "developer", "content": "Please introduce yourself to the user."}
)
```

The non-Responses version (`14-function-calling.py:151`) does not add any message -- it just queues `LLMRunFrame()` and lets the system instruction drive the greeting. This Responses version adds an explicit developer message, which is a behavioral divergence from the original example.

---

### Design Concerns

**4. Missing `cached_tokens` and `reasoning_tokens` in usage metrics**
`responses/llm.py:286-291`

The Chat Completions service extracts `cache_read_input_tokens` and `reasoning_tokens` from usage (`base_llm.py:490-505`). The Responses API also provides these (via `usage.input_tokens_details.cached_tokens` and `usage.output_tokens_details.reasoning_tokens`), but the new service does not extract them. Usage metrics will be less detailed for Responses API calls than for Chat Completions calls.

**5. `_closing` context manager is copy-pasted**
`responses/llm.py:232-243` is identical to `base_llm.py:472-485`. Consider extracting to a shared utility. Not blocking, but worth noting for follow-up.

**6. Defensive `getattr` on typed SDK objects**
`responses/llm.py:254-294` uses `getattr(item, "type", None)`, `getattr(item, "id", "")`, etc. on typed OpenAI SDK event objects. The Chat Completions service accesses attributes directly (e.g. `chunk.usage.prompt_tokens`). Minor inconsistency.

---

### Minor / Nits

**7. Unused base settings fields** -- `responses/llm.py:108-114` initializes `frequency_penalty`, `presence_penalty`, `seed`, `top_k`, `max_tokens` in default settings but `_build_response_params` never reads them. If someone sets `frequency_penalty=0.5`, it will silently be ignored.

**8. No `input_audio` multimodal conversion** -- `_convert_multimodal_content` handles `text` and `image_url` but not `input_audio` (unknown types fall through via pass-through on line 249-250, so it works, but explicit handling would be more robust).

**9. Example typo (pre-existing)** -- `20a-persistent-context-openai-responses.py:119`: `"conversatione"` should be `"conversation"`. Exists in the non-Responses version too.

**10. `responses/__init__.py` is empty** -- Consider exporting `OpenAIResponsesLLMService`, `OpenAIResponsesLLMSettings` for cleaner imports.

---

### Positives

- Clean adapter pattern following established `BaseLLMAdapter` interface
- Good test coverage: 19 adapter unit tests + 5 `run_inference` integration tests
- `store=False` default is well-reasoned and well-documented
- Thoughtful `system_instruction` to `instructions` / developer-message fallback logic
- `service_tier` support
- Examples cover key use cases with evals wired up

kompfner:
> 3. Example inconsistency: 14-function-calling-openai-responses.py:151-153

I believe I added the

```python
context.add_message(
    {"role": "developer", "content": "Please introduce yourself to the user."}
)
```

part before implementing a change to the `OpenAIResponsesLLMAdapter` that would detect an empty conversation history and "move" the service-settings-provided system instruction into it as an initial "developer"-role message.

So technically now we could remove the above lines and let the adapter do its magic.

But...I'm not actually sure that providing only a system instruction and nothing in the conversation history is still a pattern we want to be promoting, now that we're (correctly, I would argue) conceptually treating the system instruction as distinct from conversation history.

I'd actually argue for updating _other_ examples to be more like this one.

kompfner:
> 1. run_inference double-applies instructions for empty context + explicit system_instruction
responses/llm.py:377-397

Fixed! Good catch. https://github.com/pipecat-ai/pipecat/pull/4074/changes/348df9d4ced25de0f6c36fb0fb38e1b31743d28b

kompfner:
> 2. _full_model_name set without initialization or accessor

This is actually OK. Added a comment for clarity, and fixed a tracing issue discovered while looking into this: https://github.com/pipecat-ai/pipecat/pull/4074/changes/a3431d3b0176c4740dc5f27fe5f59977ae1d54bc

kompfner:
> 4. Missing cached_tokens and reasoning_tokens in usage metrics

Fixed: https://github.com/pipecat-ai/pipecat/pull/4074/changes/4ec7be88507eaae006d41c283d80ce1f805de920

kompfner:
> 5. _closing context manager is copy-pasted

Not worried about this. But actually, I just noticed some helpful explanatory comments *weren't* copy/pasted. Porting them over. https://github.com/pipecat-ai/pipecat/pull/4074/changes/05e344b9ec67fe94ab22f5fe8047bd9a41b40ed6

kompfner:
Ah! Just noticed that `model` is an initializer arg! We don't want that (users should provide `settings.model`). Fixed: https://github.com/pipecat-ai/pipecat/pull/4074/changes/6424c36666bd7755c7fe5c26d342e5431bb5d3c4

kompfner:
> 6. Defensive getattr on typed SDK objects

Hehe, already fixed before I got to this bit of feedback 👍

kompfner:
> 7. Unused base settings fields

Not unique to this service. It's OK.

kompfner:
> 8. No input_audio multimodal conversion

This is somewhat expected - we don't support audio context messages in OpenAILLMService either.

We potentially could, though...looks like we're using an out-of-date openai Python library version that doesn't have types for audio messages, though...investigating.

Looks like we'd need to update other libraries that are constraining the openai Python library to an old version 1.97.0. BUT just because types are missing for audio messages doesn't mean they're not already supported. The good news is that "input_audio" messages are *already* in the right format in `LLMContext`. Quickly checking if they work.

UPDATE: looks like, even though newer openai Python library versions show that "input_audio" is supported with the Responses API, it's not:

Screenshot 2026-03-19 at 3 32 10 PM

API calls fail if they contain input_audio fields, with an explicit message saying that input_audio is not a supported message type.

UPDATE: here's a comment update addressing this: https://github.com/pipecat-ai/pipecat/pull/4074/commits/ea1534f9f8e757ee0fa86c46aea79eafdcb86a95

markbackman (approved):
Thanks for the fixes! This looks really good. I'm in favor of merging and adding new functionality later. Nice work 👏

kompfner:
> 9. Example typo (pre-existing)

Fixed: https://github.com/pipecat-ai/pipecat/pull/4074/changes/dafbb2eb6628692b3cba8e7659c99abdf124e97d

kompfner:
> 10. responses/__init__.py is empty

Not addressing now. Maybe someday we can consider this work across all the services.

PR #4075, opened 2026-03-18
Fixing TTS frame order

filipi87:
## Summary                                                                                                                                   
- Replaced the internal `_contexts_queue` (audio-context-only FIFO) with a unified `_serialization_queue` that carries three item types: audio context IDs (str), arbitrary non-system downstream frames (Frame), and a shutdown sentinel (None)                                        
- Non-system downstream frames pushed via `push_frame` are now routed through the serialization queue instead of being emitted directly, ensuring they respect the order of any already-queued audio contexts                                                                      
- `AggregatedTextFrame` is now enqueued in the serialization queue (when its audio context doesn't yet exist) or appended directly to the existing context queue, so it always appears immediately before the `TTSStartedFrame` of the audio context it describes                     
- `TTSTextFrame` is now unconditionally appended to the audio context, removing the branch that could push it out-of-band
                                                                                                                                            
## Issue fixed                                                                                                                                 
- non-system frames were pushed directly to the pipeline, causing them to race ahead of (or fall behind) the `TTSStartedFrame` / `TTSAudioRawFrame` / `TTSStoppedFrame` sequence produced by a audio context. This
  resulted in incorrect downstream frame ordering.

### DIFF
--- a/src/pipecat/services/tts_service.py
+++ b/src/pipecat/services/tts_service.py
@@ -1060,11 +1060,8 @@ class TTSService(AIService):
             # Only override append_to_context if explicitly set
             if append_tts_text_to_context is not None:
                 frame.append_to_context = append_tts_text_to_context
-            # For services using the audio context we are appending to the context, so it preserves the ordering.
-            if self.audio_context_available(context_id):
-                await self.append_to_audio_context(context_id, frame)
-            else:
-                await self.push_frame(frame)
+            # Appending to the context, so it preserves the ordering.
+            await self.append_to_audio_context(context_id, frame)
 
     async def tts_process_generator(
         self, context_id: str, generator: AsyncGenerator[Frame | None, None]
--- a/src/pipecat/services/tts_service.py
+++ b/src/pipecat/services/tts_service.py
@@ -43,6 +43,7 @@ from pipecat.frames.frames import (
     LLMFullResponseEndFrame,
     LLMFullResponseStartFrame,
     StartFrame,
+    SystemFrame,
     TextFrame,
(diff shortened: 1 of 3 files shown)


src/pipecat/services/tts_service.py
```diff
@@ -1295,7 +1309,14 @@ def _refresh_audio_context(self, context_id: str):
 
     def _create_audio_context_task(self):
         if not self._audio_context_task:
-            self._contexts_queue: asyncio.Queue = asyncio.Queue()
+            # Single FIFO queue that serializes everything the TTS service emits downstream.
```
kompfner, line 1312: Nice comment! This makes a lot of sense

src/pipecat/services/tts_service.py
```diff
@@ -1295,7 +1309,14 @@ def _refresh_audio_context(self, context_id: str):
 
     def _create_audio_context_task(self):
```
kompfner, line 1310: Do today's names (`_create_audio_context_task`, `_audio_context_task_handler`, maybe others) still make sense? They're concerned with the serialization queue, which is broader than just the audio contexts

kompfner (approved):

aconchillo (approved):

aconchillo:
Yes! :raised_hands:

src/pipecat/services/tts_service.py
filipi87: We are still mainly handling the audio context there, but I see your point, since it now uses the serialization queue.
I don’t have a strong opinion on this, but since these are all internal, we can keep it as it is for now.

PR #4076, opened 2026-03-18
Fixed SimliVideoService pushing near-silent audio constantly preventing BotStoppedSpeaking 

phamtrung0633:
Fixed SimliVideoService forwarding near-silent audio frames (quantization noise) as TTSAudioRawFrame. Added RMS threshold check to skip frames below 1.0, which is well above the observed noise floor (~0.41) but far below real speech. Without this fix - BotStoppedSpeaking will never trigger.


markbackman:
If I run [example 27](https://github.com/pipecat-ai/pipecat/blob/main/examples/foundational/27-simli-layer.py) using [Whisker](https://github.com/pipecat-ai/whisker) to inspect frames, I see `BotStartedSpeakingFrame` and `BotStoppedSpeakingFrame` for each bot turn. Can you share a repro case where this isn't working?

markbackman:
Closing as this PR has gone stale and the issue could not be reproduced. Please request to reopen if this is still of interest and you have a repro case.

PR #4078, opened 2026-03-18
Updates for Gemini Live 

chadbailey59:
This PR updates the google-genai dependency, and makes a few small changes for compatibility in future Gemini Live models.


chadbailey59:
(note to self: I need to check turning off automatic VAD)

chadbailey59:
@kompfner I decided deleting the example without a context aggregator made the most sense.

src/pipecat/services/google/gemini_live/llm.py
```diff
@@ -1522,6 +1527,14 @@ async def _create_single_response(self, messages_list):
         if self._disconnecting or not self._session:
             return
 
+        model = self._settings.model or ""
+        if "gemini-3" in model:
+            await self.push_error(
+                f"LLMMessagesAppendFrame is not supported with model '{model}'. "
```
kompfner: Out of curiosity: this `_create_single_response` code path doesn't work with gemini 3 even for the "conversation kick-off" usage demonstrated in today's 26- example?

kompfner: Note: not sure we necessarily want to discourage use of `LLMMessagesAppendFrame` per se (you may want to use it, with `run_llm=True`, in lieu of `LLMRunFrame`, to start a conversation). 

Maybe (depending on the answer to the above question) the thing to warn about is either:
- using `LLMMessagesAppendFrame` mid-conversation (i.e. for anything other than conversation kick-off)
- using `LLMMessagesAppendFrame` without context aggregators (and even then, only using it for conversation kick-off)

chadbailey59: When I removed the gemini-3 check from `_create_single_response` and tried running the deleted 26-gemini-live.py example with `gemini-3.1-flash-audio-eap` as the model, I can see this in the logs:

```
Creating response: [Content(
  parts=[
    Part(
      text='Greet the user and introduce yourself.'
    ),
  ],
  role='user'
)]
```

But the bot doesn't greet me. If I speak, the bot will respond, though. I _presume_ what's happening is that as soon as the transport connects, it starts sending realtime audio input, and once that's happened, either the send_client_content call is getting silently ignored, or pipecat is eating an error from the SDK somewhere.

examples/foundational/26-gemini-live.py
```diff
@@ -4,26 +4,32 @@
 # SPDX-License-Identifier: BSD 2-Clause License
```
markbackman, line 4: This change simply moves 26a-gemini-live-transcription.py to 26-gemini-live.py. No change otherwise.

markbackman (commented):
When a bot is configured for a user to speak first, e.g. don't set up a context message and push `LLMRunFrame`, then I see an error:
```
2026-03-23 15:54:12.207 | DEBUG    | pipecat.services.google.gemini_live.llm:_create_initial_response:1505 -   
  Creating initial response: [Content(                                                                           
    parts=[                                                                                                      
      Part(                                                                                                      
        text='Hey, can you hear me?'                                                                             
      ),                                                                                                         
    ],                                                                                                           
    role='user'                                                                                                  
  )]                                                                                                             
  2026-03-23 15:54:13.199 | WARNING  | pipecat.services.google.gemini_live.llm:_handle_connection_error:1383 -   
  Connection error (failure 1/3): 1007 None. Request contains an invalid argument.
```

From what I can tell, we're calling `send_realtime_input()` is called with user audio data, but subsequently, `process_frame()` is handling an LLMContextFrame which results in calling `_create_initial_response()`.

examples/foundational/26-gemini-live.py
```diff
@@ -33,56 +39,59 @@
...
-    """
-
     llm = GeminiLiveLLMService(
         api_key=os.getenv("GOOGLE_API_KEY"),
         settings=GeminiLiveLLMService.Settings(
-            system_instruction=system_instruction,
-            voice="Puck",  # Aoede, Charon, Fenrir, Kore, Puck
+            voice="Aoede",  # Puck, Charon, Kore, Fenrir, Aoede
+            # system_instruction="Talk like a pirate."
+            # inference_on_context_initialization=False,
```
kompfner: `inference_on_context_initialization` is a direct argument to `GeminiLiveLLMService` (not to `GeminiLiveLLMService.Settings`)

changelog/4078.changed.md
```diff
@@ -0,0 +1 @@
+- Upcoming Gemini Live models don't support replacing the context during a session, so creating a single response (by using an `LLMMessagesAppendFrame` without a context manager, for example) will return an error with those models.
```
kompfner: Nit: I'm not sure "replacing context during a session" ever _truly_ worked, so I wouldn't go with this framing of the fix.

The fix in this PR is more just a general "tweaks to properly support Gemini Live 3.1" or something like that, because it just doesn't work with the new model without the fix.

markbackman (approved):
LGTM! Thanks for improving this.

PR #4081, opened 2026-03-19
Sarvam LLM: Settings refactor and cleanup

markbackman:
## Summary

Follow-up to #3981 applying Settings pattern refactoring and cleanup to the Sarvam LLM integration.

- Use `self.Settings` / `Settings` alias instead of `SarvamLLMSettings` directly
- Remove deprecated `model` init param — use `settings=SarvamLLMService.Settings(model=...)` instead
- Remove `settings.extra` dict bridge — access `self._settings.wiki_grounding` / `self._settings.reasoning_effort` directly
- Remove `_update_settings()` override — base class handles this
- Remove `_extract_sarvam_extra_from_settings()` — no longer needed
- Remove unreachable tool-model validation check (`_TOOL_CALLING_MODELS == _SUPPORTED_MODELS`)

## Open question: error handling overrides

The remaining ~95 lines of error wrapping code (`_call_with_raw_sarvam_errors`, `get_chat_completions`, `run_inference`, `_format_raw_server_error`, `_extract_raw_server_message`, `_payload_to_message`) override base class behavior. The base class already catches exceptions and calls `push_error()`, and the `openai` library extracts error bodies from API responses.

@dhruvladia-sarvam — Can you provide a concrete example of a Sarvam error payload that gets lost without this wrapper? Also note that wrapping every non-timeout exception in `RuntimeError` replaces the original exception type (`APIStatusError`, `APIConnectionError`, etc.), which makes it harder for application code to handle specific error cases.

## Test plan

- [ ] Verify Sarvam LLM service initializes correctly with `Settings`
- [ ] Verify `wiki_grounding` and `reasoning_effort` are passed in chat completion params
- [ ] Verify `LLMUpdateSettingsFrame` correctly updates Sarvam-specific settings at runtime


dhruvladia-sarvam:
Thanks mark, I validated this and agree we should remove the Sarvam error-wrapping overrides.

I couldn’t reproduce a case where Sarvam payload details are lost without the wrapper. With the current OpenAI SDK, a Sarvam-like 400 response raises a typed APIStatusError subclass (e.g., BadRequestError), `str(e)` includes the server payload, and `e.body` preserves parsed error fields. Also, BaseOpenAILLMService already propagates errors via push_error for pipeline execution.

Given that, the wrapper’s RuntimeError conversion is not ideal because it erases exception types (APIStatusError/APIConnectionError/etc.) that callers may want to handle explicitly.

I’m aligned with removing these overrides in this follow-up.

markbackman:
@dhruvladia-sarvam sounds great! I just pushed the changes to this branch to remove the function wrapping. Please take a look.

dhruvladia-sarvam:
> @dhruvladia-sarvam sounds great! I just pushed the changes to this branch to remove the function wrapping. Please take a look.

Got it, please re-review #3981

PR #4082, opened 2026-03-19
Improvements to SarvamTTSService.

filipi87:
## Summary                                                                                                                                
                                                                                                                                            
- Removed redundant `push_frame` and `process_frame` overrides from `SarvamTTSService` that duplicated parent class behavior              
- Removed manual `flush_audio()` call on `LLMFullResponseEndFrame`/`EndFrame` (handled by the parent `TTSService`)           
- Switched from directly pushing frames via `push_frame()` to using `append_to_audio_context()` for audio and error frames, enabling proper audio context support

### DIFF
--- a/src/pipecat/services/sarvam/tts.py
+++ b/src/pipecat/services/sarvam/tts.py
@@ -1156,9 +1156,7 @@ class SarvamTTSService(InterruptibleTTSService):
                     # Check for interruption before processing audio
                     await self.stop_ttfb_metrics()
                     audio = base64.b64decode(msg["data"]["audio"])
-                    frame = TTSAudioRawFrame(
-                        audio, self.sample_rate, 1, context_id=context_id
-                    )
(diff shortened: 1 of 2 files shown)


src/pipecat/services/sarvam/tts.py
```diff
@@ -1031,23 +1031,6 @@ async def flush_audio(self, context_id: Optional[str] = None):
...
-        """
-        await super().push_frame(frame, direction)
-
-    async def process_frame(self, frame: Frame, direction: FrameDirection):
-        """Process a frame and flush audio if it's the end of a full response."""
-        await super().process_frame(frame, direction)
-
-        # When the LLM finishes responding, flush any remaining text in Sarvam's buffer
-        if isinstance(frame, (LLMFullResponseEndFrame, EndFrame)):
-            await self.flush_audio()
```
filipi87, line 1049: The parent class already does that now.

markbackman (approved):
LGTM!

PR #4084, opened 2026-03-19
Refactoring the way we automatically push TTSStoppedFrame.

filipi87:
## Summary                                                                                                                                
- Removed the separate `_stop_frame_handler` background task, `_stop_frame_queue`, and the `push_frame` override logic that fed it      
- Moved the "auto push `TTSStoppedFrame`" logic into `_handle_audio_context`, which already owns the full audio lifecycle for each context
- Since all TTS services now route audio through audio contexts, the separate handler was redundant                                       
- Changed `stop_frame_timeout_s` default from `2.0` → `3.0` to align with the context idle timeout previously hardcoded in `_handle_audio_context`

### DIFF
--- a/src/pipecat/services/tts_service.py
+++ b/src/pipecat/services/tts_service.py
@@ -156,7 +156,7 @@ class TTSService(AIService):
         # if True, TTSService will push TTSStartedFrames and create audio contexts automatically
         push_start_frame: bool = False,
         # if push_stop_frames is True, wait for this idle period before pushing TTSStoppedFrame
-        stop_frame_timeout_s: float = 2.0,
+        stop_frame_timeout_s: float = 3.0,
         # if True, TTSService will push silence audio frames after TTSStoppedFrame
         push_silence_after_stop: bool = False,
         # if push_silence_after_stop is True, send this amount of audio silence
@@ -332,9 +332,6 @@ class TTSService(AIService):
             self._text_filters = [text_filter]
 
         self._resampler = create_stream_resampler()
-
-        self._stop_frame_task: Optional[asyncio.Task] = None
-        self._stop_frame_queue: asyncio.Queue = asyncio.Queue()
 
         self._processing_text: bool = False
         self._tts_contexts: Dict[str, TTSContext] = {}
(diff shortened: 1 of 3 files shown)


src/pipecat/services/tts_service.py
```diff
@@ -156,7 +156,7 @@ def __init__(
         # if True, TTSService will push TTSStartedFrames and create audio contexts automatically
         push_start_frame: bool = False,
         # if push_stop_frames is True, wait for this idle period before pushing TTSStoppedFrame
-        stop_frame_timeout_s: float = 2.0,
+        stop_frame_timeout_s: float = 3.0,
```
filipi87, line 159: Changing the default to 3 to match the previous `AUDIO_CONTEXT_TIMEOUT`.

src/pipecat/services/tts_service.py
```diff
@@ -1389,15 +1352,25 @@ async def _handle_audio_context(self, context_id: str):
...
                     else:
                         await self.push_frame(frame)
             except asyncio.TimeoutError:
                 # We didn't get audio, so let's consider this context finished.
                 logger.trace(f"{self} time out on audio context {context_id}")
+                if should_push_stop_frame and self._push_stop_frames:
+                    await self.push_frame(TTSStoppedFrame(context_id=context_id))
                 break
 
+        if should_push_stop_frame and self._push_stop_frames:
```
kompfner, line 1371: Checking my understanding: this is for the _non-timeout_ case where we need to send `TTSStoppedFrame`, right? When we _know_ that the audio context has ended?

kompfner (approved):

src/pipecat/services/tts_service.py
filipi87: This applies when the TTS service doesn’t receive any signal indicating that audio generation has fully completed, so it doesn’t emit the `TTSStopped` frames on its own.

filipi87: Now that all TTS services use the audio context, we no longer need an extra queue or task to emit `TTSStoppedFrame` in these cases.

PR #4087, opened 2026-03-20
enable_dialout should not depend on sip_caller_phone being set

vr000m:
Decouple `enable_dialout` from SIP dial-in configuration in `configure()` and allowing dial-out without needing to create a SIP dial-in

### DIFF
--- a/src/pipecat/runner/daily.py
+++ b/src/pipecat/runner/daily.py
@@ -225,6 +225,9 @@ async def configure(
         if room_geo:
             room_properties.geo = room_geo
 
+        if enable_dialout:
+            room_properties.enable_dialout = True
+
         # Add SIP configuration if enabled
         if sip_enabled:
             sip_params = DailyRoomSipParams(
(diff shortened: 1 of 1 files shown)


src/pipecat/runner/daily.py
```diff
@@ -207,7 +207,7 @@ async def configure(
         return DailyRoomConfig(room_url=room_url, token=token)
 
     # Create a new room
-    room_prefix = "pipecat-sip" if sip_enabled else "pipecat"
+    room_prefix = "pipecat-sip" if (sip_enabled or enable_dialout) else "pipecat"
```
markbackman: This just sets the room name for convenience. I don't think we want this change since you can set enable_dialout for a PSTN dial-out call.

`sip_enabled` is set when `sip_caller_phone` is set. If this isn't a requirement, we should update to be the definitive set to check for when a SIP call is happening.

src/pipecat/runner/daily.py
```diff
@@ -225,6 +225,9 @@ async def configure(
         if room_geo:
             room_properties.geo = room_geo
 
+        if enable_dialout:
```
markbackman, line 228: Yes! Good catch 🙇

src/pipecat/runner/daily.py
vr000m:
```suggestion
    room_prefix = "pipecat-telephony" if (sip_enabled or enable_dialout) else "pipecat"
```

markbackman (approved):
LGTM!

PR #4089, opened 2026-03-20
Centralize system message handling in adapters; add developer message support

kompfner:
## Summary

- **Centralizes system message handling in adapters** for maintainability and clarity. Logic for resolving `system_instruction` vs context system messages was duplicated across ~16 locations in service `llm.py` files — separately for in-pipeline (`_process_context`) and out-of-band (`run_inference`) code paths. Each adapter is now the single source of truth for how system messages are handled for its provider.
- **Adds support for "developer" role messages** in conversation context across all adapters:
  - Non-OpenAI adapters (Anthropic, Google, Bedrock, Perplexity, realtime adapters): "developer" messages are converted to "user". They are NOT promoted to the system instruction — use `system_instruction` for that.
  - OpenAI adapter: "developer" messages pass through in conversation history without triggering warnings.
  - Responses adapter: "developer" messages are kept as "developer" role (matching the existing "system" → "developer" conversion).
- **Updates ~220 example files** to use "developer" role for programmatic conversation-kickoff messages and other developer-intent messages that were incorrectly using "user" role.
- **Fixes Gemini Live** (`GoogleGeminiLiveLLMService`) not honoring `settings.system_instruction` — it was reading from a deprecated constructor parameter instead of the settings object.
- **Fixes Bedrock** sending an empty message list when the only message was a system message.
- Adds `07e-interruptible-anthropic.py` example and registers it in release evals.
- Shared helpers on `BaseLLMAdapter`: `_extract_initial_system()` and `_resolve_system_instruction()`.

## Breaking Changes

- `GeminiLLMAdapter` now only treats `messages[0]` as the initial system message, matching all other adapters. Previously it searched for the first "system" message anywhere in the conversation history. A "system" message appearing later in the list will now be converted to "user" instead of being extracted as the system instruction.

## Testing

- Run `uv run pytest tests/test_get_llm_invocation_params.py -v` — 104 tests covering all adapters including realtime (system_instruction handling, developer messages, base helpers)
- Run `uv run pytest tests/test_run_inference.py -v` — updated existing tests
- Run full test suite: `uv run pytest`
- Verify no "Both system_instruction" warnings remain in service files: `rg "Both system_instruction" src/pipecat/services/` should return nothing


kompfner:
Still need:
- Manual reviewing
- Testing
- Follow up work of converting examples to use "developer" in place of various "user" messages

kompfner:
I have a remaining task, which is to validate the other `OpenAILLMService`-descended services. Perplexity did not take "developer" messages, so I'm wondering whether other services will have the same limitation.

kompfner:
Noting a follow-up task: we really should make sure that realtime services follow the same prioritization as non-realtime services in that they should prefer `settings.system_instruction` to context-provided system instructions (and log similar warnings about what they're doing)

src/pipecat/services/together/llm.py
```diff
@@ -46,7 +51,7 @@ def __init__(
         Args:
             api_key: The API key for accessing Together.ai's API.
             base_url: The base URL for Together.ai API. Defaults to "https://api.together.xyz/v1".
-            model: The model identifier to use. Defaults to "meta-llama/Meta-Llama-3.1-8B-Instruct-Turbo".
+            model: The model identifier to use. Defaults to "openai/gpt-oss-20b".
```
kompfner, line 54: See [commit msg](https://github.com/pipecat-ai/pipecat/pull/4089/commits/d7e8334cc8a5731f93f0c18356f475aeafa38264) for explanation

src/pipecat/services/google/gemini_live/llm.py
```diff
@@ -788,7 +788,7 @@ def __init__(
         self._last_sent_time = 0
         self._base_url = base_url
 
-        self._system_instruction_from_init = system_instruction
+        self._system_instruction_from_init = self._settings.system_instruction
```
kompfner, line 797: This is an important bugfix

aconchillo (approved):

aconchillo:
So much cleaner! LGTM!

PR #4091, opened 2026-03-20
feat: send per-context setup in Gradium TTS multiplexing

markbackman:
## Context

Requested by the Gradium team. Docs:
https://docs.gradium.ai/guides/multiplexing#multiplexing

## Summary

- Send a per-context `setup` message with `client_req_id` in `GradiumTTSService`, following Gradium's multiplexing protocol for `close_ws_on_eos=False`
- Previously a single setup was sent at connection time without `client_req_id`, preventing Gradium from associating requests with sessions
- Pipecat's `context_id` is now used as `client_req_id` in setup messages, matching the text and `end_of_stream` messages that already use it
- Setup is sent once per context (tracked via `_setup_context_ids`), not on every `run_tts` call

### DIFF
--- a/src/pipecat/services/gradium/tts.py
+++ b/src/pipecat/services/gradium/tts.py
@@ -140,6 +140,7 @@ class GradiumTTSService(WebsocketTTSService):
 
         # State tracking
         self._receive_task = None
+        self._setup_context_ids: set[str] = set()
 
     def can_generate_metrics(self) -> bool:
         """Check if this service can generate processing metrics.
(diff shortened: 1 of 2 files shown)


src/pipecat/services/gradium/tts.py
```diff
@@ -363,8 +371,12 @@ async def run_tts(self, text: str, context_id: str) -> AsyncGenerator[Frame, Non
                 await self._connect()
 
             try:
+                ws = self._get_websocket()
+                if context_id not in self._setup_context_ids:
+                    await ws.send(json.dumps(self._build_setup_msg(context_id)))
+                    self._setup_context_ids.add(context_id)
                 msg = self._build_msg(text=text, context_id=context_id)
```
kompfner: nit: do we want to now call this function `_build_text_msg`, to keep it consistent with `_build_setup_msg`?

kompfner (approved):

src/pipecat/services/gradium/tts.py
markbackman: yup!

PR #4092, opened 2026-03-20
Add Smallest AI TTS service integration

harshitajain165:
## Summary

Adds Smallest AI WebSocket-based TTS service, aligned with current Pipecat conventions.

This PR has the following history:
- #3897 by @harshitajain165 — original integration with both STT and TTS (HTTP + WebSocket variants)
- #4014 by @markbackman — refactored to WebSocket-only, aligned with Pipecat service conventions (superseded #3897)
- This PR — TTS-only split from #4014, raised because the STT service has a known hallucination/finalize behaviour issue that needs to be resolved before it is production-ready. Splitting allows TTS (which is working well) to move forward independently.

The TTS implementation is taken directly from #4014 with no changes. The only difference is:
- `stt.py` removed — will be submitted as a follow-up PR once the STT issue is resolved
- `stt_latency.py` unchanged — `SMALLEST_TTFS_P99` will be added with the STT PR
- Example (`07zl-interruptible-smallest.py`) uses `DeepgramSTTService` as a stand-in for STT

Everything else — settings pattern, keepalive, `_update_settings` overrides, model enum — is identical to #4014.

## Key details

* **WebSocket-only**: WebSocket service for real-time use
* **Settings pattern**: `SmallestTTSSettings` dataclass with runtime-updatable fields
* **Init-only vs runtime-updatable**: model is init-only; voice/language/speed/etc. are runtime-updatable
* **`_update_settings` override**: TTS reconnects only on model change
* **TTS keepalive**: Flush keepalive every 30s to prevent 60s idle timeout
* **TTS settings default to None**: API applies its own defaults for speed/consistency/similarity/enhancement
* **Model enum**: `SmallestTTSModel.LIGHTNING_V3_1`


markbackman (approved):
LGTM! I pushed a few additional changes to this branch to align with some TTS service changes we've made. Everything looks good now. Thanks for the help on this one 🙌

PR #4093, opened 2026-03-20
fix: always include parameters in Genesys AudioHook messages

poislagarde:
## Summary

The AudioHook protocol requires every message to include a `parameters` object (even if empty `{}`). The `_create_message` method conditionally included it only when parameters were truthy:

```python
if parameters:
    msg["parameters"] = parameters
```

This meant `pong` responses and `closed` responses without `outputVariables` were sent without a `parameters` field at all.

The Genesys reference client validates every incoming message with `isMessageBase`, which requires `isJsonObject(msg['parameters'])`. Without the field, the message fails validation and triggers a fatal protocol error. This causes a cascade:

1. Client rejects the `pong` → doesn't update its `serverseq` tracker
2. Next message (`closed`) arrives with a higher-than-expected `seq` → protocol error
3. Client rejects `closed` → **`outputVariables` never reach the Architect flow**

### Fix

```python
msg["parameters"] = parameters if parameters is not None else {}
```

## Testing

- All 28 existing Genesys serializer unit tests pass (updated 2 to match new behavior)
- Tested against the [Genesys AudioHook reference implementation](https://github.com/purecloudlabs/audiohook-reference-implementation) test client (`client/`). Before the fix, the client logged `Protocol error: Current serverseq=1 and incoming message has seq=3` and the session ended in `CLOSING-ERROR`. After the fix, the session completes cleanly:

```
INFO: onOpenedMessage - Session open! Parameters: {"startPaused":false,"media":[...],"selectedLanguage":"en-US"}
INFO: onPongMessage - rtt: 3.8ms
INFO: Source stream ended (duration: PT3S), closing session.
INFO: onClosedMessage - Session closed (state: CLOSING)
INFO: All sessions disconnected, exiting
```

## Changelog

Included as `changelog/0000.fixed.md` — happy to rename to the PR number once assigned.


markbackman (approved):
LGTM! Thanks for the contribution 🙇

PR #4098, opened 2026-03-21
fix: re-check turn stop when transcript arrives after expired timeout

namanbansal013:
## Summary

`TurnAnalyzerUserTurnStopStrategy._handle_transcription` only called `_maybe_trigger_user_turn_stopped()` inside the `if frame.finalized:` block. When a **non-finalized** `TranscriptionFrame` arrives after the strategy's internal STT timeout has already expired, there is no code path to re-evaluate turn stop conditions. The turn stalls for up to 5 seconds (`user_turn_stop_timeout` safety-net).

This happens reliably when STT endpointing is faster than VAD `stop_secs` — for example, Deepgram Nova-3 (default 10ms endpointing) paired with Silero VAD (`stop_secs=0.5`). The internal timeout becomes `max(0, 0.35 - 0.5) = 0.0s`, fires instantly before the `TranscriptionFrame` is processed, and the late-arriving transcript has no re-check path.

### Timeline from production logs

```
13:37:51.164  User started speaking (Turn 6)
13:37:51.358  SmartTurn: End of Turn result: COMPLETE (97.19%, e2e: 96.9ms)
13:37:51.363  Deepgram is_final transcript: "Hi, Sarah."  (non-finalized)
              ← timeout = max(0, 0.35 - 0.5) = 0.0s → fires instantly
              ← _maybe_trigger: _text is empty → returns early
              ← transcript arrives → _text set, but no re-check → STUCK
13:37:56.363  User turn stopped — strategy: timeout  ← 5 SECONDS LATER
```

### Fix

Move `await self._maybe_trigger_user_turn_stopped()` out of the `if frame.finalized:` block so it runs on every `TranscriptionFrame`. The `_maybe_trigger_user_turn_stopped()` method already has comprehensive guards that prevent premature triggers:

- Returns early if `_text` is empty or `_turn_complete` is False
- Returns early if timeout is still running (`_timeout_task is not None`) and transcript is not finalized
- Triggers immediately only when finalized OR the timeout has already expired

No logic changes to `_maybe_trigger_user_turn_stopped()` itself.

## Testing

- `uv run ruff check` — no lint errors
- `uv run ruff format --check` — no formatting issues
- `uv run pytest tests/test_user_turn_stop_strategy.py -v` — all 20 tests pass (5 new + 15 existing)
- `uv run pytest tests/` — 903 passed, 70 skipped (2 pre-existing collection errors in `test_deepgram_stt.py` and `test_settings.py` due to Deepgram SDK import mismatch — unrelated)

### New tests

- `test_transcript_after_expired_timeout_triggers_immediately` — reproduces the race condition: VAD stop with 0.0s timeout, then late transcript → now triggers immediately
- `test_set_finalized` — finalized transcript after VAD stop triggers immediately (unchanged behavior)
- `test_set_non_finalized_triggers_after_timeout` — non-finalized transcript before timeout waits correctly (unchanged behavior)
- `test_transcript_while_speaking_does_not_trigger` — transcript during speech does not trigger (unchanged behavior)
- `test_transcript_before_vad_stop_waits_for_timeout` — transcript before VAD stop waits for timeout (unchanged behavior)


markbackman:
We'll take a look, but a quick comment is that we strongly recommend not modifying the `VADParams` `stop_secs` value. If you do, you will need to re-run the [stt-benchmark](https://github.com/pipecat-ai/stt-benchmark) and determine a new p99 latency number which will be meaningfully slower. We found 0.2 secs to be an optimal value across providers that allows the VAD to meaningfully pad the speech with silence for STT accuracy, but also provide fast final transcripts.

The `stop_secs` is now more of a fixed value which is used by STT services to tell it when to generate the transcript after silence and the turn control logic is deferred to the start and stop strategies in the `LLMUserAggregator`. For the `SpeechTimeoutUserTurnStopStrategy` you have a static timeout you can set which works like the old `stop_secs` value. If you're using the `TurnAnalyzerUserTurnStopStrategy`, there is no timeout; you're relying on the smart-turn model to make this decision for you.

markbackman:
Ok, I've had a chance to look at this more closely, and the change your suggesting would create problems unfortunately.

The issue is that a single user turn can generate 1-n TranscriptionFrames which have a variety of timing possibilities. A bad case would be:

- t=0: TranscriptionFrame
- t=0.1 VADUserStoppedSpeakingFrame
- t=0.15: TranscriptionFrame
- t=0.5: TranscriptionFrame
 
In this case, the TranscriptionFrame received at 0.15 seconds would result in the turn finalizing, but the issue is it's not the last transcript from the user turn. That means the final text would interrupt the bot mid-response.

The current system takes into account a p99 latency value which was calculated using this time to final segment STT benchmark: https://github.com/pipecat-ai/stt-benchmark. These benchmarks assume 0.2sec VAD stop_secs (see my comment above). You can modify the VAD stop_secs (though not optimal); if you do, you need to modify the p99 latency time, which you can derive by running the benchmark.

The goal of all of this is to provide optimal latency between the user stop and bot started phase of the conversation. This takes careful orchestration.

I'm adding warnings to the code to prevent others from hitting this case in this PR:
https://github.com/pipecat-ai/pipecat/pull/4115

I'll close out this change, but thanks so much for raising this to our attention! 🙇

namanbansal013:
Thanks @markbackman for looking into this. I understand that we shouldn't play around with stop_secs now. The reason why we were trying to change this was because our bot was constantly interrupting the user while they were speaking. We are using Silero VAD with Smart Turn v3.1 and also using user_turn_completion_mixin. Even after this, many times the bot interrupts the user as the Smart Turn model is falsely identifying that the user has stopped speaking and the LLM as well gives the ✓ marker indicating that the user has stopped speaking when we are lucky and the TTFT of the LLM is very low. The result of which is that the bot interrupts the user while they were still speaking, which brings out a very bad experience. 

We were able to fix this slightly by increasing stop_secs. Could you help us out with some other alternate that you think might work better? We have two ideas in mind:
1. Increase the value of max_duration_secs for Smart Turn while figuring out end of turn
2. Migrate to Deepgram Flux for turn detection

markbackman:
Interesting. Which language are you using? Maybe there is lower accuracy for your language.

Also, how frequently do you see this?

Regardless, I will say that this, in general, is an unsolved problem. No matter which technology you use, you'll have a long tail of incorrect end of turn predictions. The goal is about reducing the likelihood, not eliminating it. (Think about human conversations—we talk over one another despite all of our experience in the real world. It's a hard problem!)

Options I can think of:
- Slow down transcript generation, in general, by changes the `stop_secs` value. This means that you're increasing the user to bot latency for every turn in hopes to avoid some bot interruptions. If you do this, you'll want to run the TTFS analysis from the [stt-benchmark](https://github.com/pipecat-ai/stt-benchmark) repo to determine your p99 transcript latency number. Make sure to configure based on your set up, including VAD stop_secs.
- Experiment with a provider that has turn detection built in. The three great options that support this are Deepgram Flux, Soniox (set `vad_force_turn_endpoint`: False), AssemblyAI.
- Write your own user stop strategy. User turn start and stop strategies are meant to be easily extensible. You can write your own to experiment with ways to do something better. If you find a good pattern, we'd love to hear about it!

namanbansal013:
Hi @markbackman, we are doing this for English only. Right now, in a session of 20 minutes, we see the bot interrupting the user due to this roughly 7-8 times, leading to a very bad experience. 

The reason why we are trying to play around with stop_secs is because in release 0.0.104 the default value of stop_secs changed from 0.8 to 0.2 because of which all these interruptions are coming in. That's why we wanted to try out by changing it to 0.5 but as you suggested, doing that is not recommended. 

For us the latency of the bot matters less compared to how accurate it is in determining when the user stopped speaking. We are fine to take a hit of 500 ms and in the worst case one second as well in order to get very accurate turn detection. We will try out the above approaches you mentioned see what works the best for us. 

Thanks for taking the time and looking into this. 

BTW, there were two more bugs we found and have raised a PR [[4102](https://github.com/pipecat-ai/pipecat/pull/4102) and [4095](https://github.com/pipecat-ai/pipecat/pull/4095)] for the same. Can you please look into them as well whenever you get time.

markbackman:
If the longer timeout works, then I'd suggest using it and re-running the STT benchmark. If you don't want to re-run the STT benchmark, the conservative thing to do would be to increase the `ttfs_p99_latency` by the same delta you're applying to the stop_secs.

E.g.
- stop_secs: 0.2 -> 0.5 (0.3 delta)
- ttfs_p99_latency: 0.35 -> 0.65 (0.3 delta)

This is overly conservative and not optimized because Deepgram does its own endpointing and will return a transcript faster than the VAD finalizing after 0.5 seconds of silence. Hence, running the benchmark is worth it to get more precision.

namanbansal013:
Hi @markbackman , wanted to follow up with some new evidence. The logs below are from a session running `stop_secs=0.2`. I reverted from 0.5 after your feedback. The issue still reproduces.

For session_id: 166237c8-65dd-4b0d-b467-15a63425f57b running on pipecat version 0.0.104 had 4 out of 29 turns hit the `user_turn_stop_timeout`. 

Root cause: `finalized` is always `False` for Deepgram

After digging into the logs, I found that the issue is more fundamental than `stop_secs` timing. Deepgram's STT service never sets `TranscriptionFrame.finalized = True`:

```python
# deepgram/stt.py is_final path:
await self.push_frame(
    TranscriptionFrame(transcript, self._user_id, time_now_iso8601(), language, result=message)
)
# finalized defaults to False never explicitly set
```

In `TurnAnalyzerUserTurnStopStrategy._handle_transcription`, `_maybe_trigger_user_turn_stopped()` is only called inside `if frame.finalized:` which never executes for Deepgram. The only trigger path is the STT timeout (150ms with defaults). If the `TranscriptionFrame` arrives even 1ms late, the turn is dead for 5 seconds.

Looking at this again, the existing guard in `_maybe_trigger_user_turn_stopped()` already prevents premature triggering for multi-transcript turns:

```python
# For non-finalized, only trigger if timeout task has completed
if self._timeout_task is None:
    await self.trigger_user_turn_stopped()
```

With the PR fix (calling `_maybe_trigger` for all `TranscriptionFrame`s), in the multi-transcript scenario:

- `t=0.10`: VAD stop → timeout starts (0.15s)
- `t=0.15`: TranscriptionFrame arrives → `_maybe_trigger` → `_timeout_task is not None` → **does not trigger** 
- `t=0.25`: timeout expires → triggers normally 

The guard prevents any trigger before the timeout expires so no premature finalization possible.

Would you be open to re-examining this fix? Happy to open a fresh PR with the same change if that's easier.

markbackman:
A few thoughts:

- Deepgram most definitely returns finalized transcripts. The issue is: when you finalize, you won't always get a finalized transcript back. That's why the logic includes a finalized path but also a timeout that uses the p99 latency value. The system is built to handle this.
- Where in the world are you located? The p99 latency value is calculated from a US location to the Deepgram US endpoints. If you're located elsewhere, then you should run the stt-benchmark from your location to determine what the possible TTFB values are. Then, set the `ttfs_p99_latency` on the Deepgram service when you initialize.

The whole point of the p99 timeout is to wait long enough for transcripts to arrive. If you want to be conservative, then you can just set the value with margin for safety. But, the optimal path is to do the measurement for your geography and settings and then set the value. With this, you'll find more success.

namanbansal013:
Hi @markbackman, wanted to share some results. We deployed the exact fix from this PR (moving `_maybe_trigger_user_turn_stopped()` outside `if frame.finalized:`) as a monkey-patch and ran multiple sessions (20 mins each) with **`stop_secs=0.2`** on pipecat `0.0.107`.

**Results (Pipecat Cloud session `bea4df21-82fb-437d-9b80-1c9840544f32`):**

- **155** times the STT timeout (150ms) expired before the `TranscriptionFrame` arrived
- **60** of those had the turn analyzer already at `COMPLETE`. Our fix caught the late transcript and triggered the turn stop immediately
- **0** instances of the 5-second `user_turn_stop_timeout` safety-net firing
- **0** premature finalization incidents. No user was ever cut off mid-sentence
- Median race delay: **~60ms** (transcript arrived 60ms after the 150ms timeout expired)

The multi-transcript concern you raised doesn't apply because `_maybe_trigger_user_turn_stopped()` already guards against it:

```python
# This check prevents triggering while the timeout is still running
if self._timeout_task is None:
    await self.trigger_user_turn_stopped()
```

If a `TranscriptionFrame` arrives at t=0.15 while the timeout task is still active (expires at t=0.25), `_timeout_task is not None` → no trigger. Only after the timeout expires and sets `_timeout_task = None` will a subsequent transcript trigger the turn stop.

We are running this from North India and I agree that tuning `ttfs_p99_latency` for geography helps, but we feel it might not fully solve this: the timeout is a **fixed budget** while Deepgram latency has a tail distribution. Even at p99, 1% of transcripts arrive late. When they do, the current code has no recovery path. The turn is dead for 5 seconds. The fix adds that recovery path at zero cost to the happy path.

Would you be open to re-examining this? Happy to open a fresh PR if that's easier.

markbackman:
@namanbansal013 I just posted https://github.com/pipecat-ai/pipecat/pull/4283, which handles the 1% case. This should significantly reduce the number of cases you're hitting.

PR #4090, opened 2026-03-20
fix: route TTS audio through audio context queue in Fish, LMNT, Neuph…

markbackman:
## Summary

- Fixed audio frame ordering and interruption handling in **Fish Audio**, **LMNT**, **Neuphonic**, and **Rime NonJson** TTS services
- These services were pushing audio frames directly via `push_frame()` in their WebSocket receive loops, bypassing the base `TTSService` audio context serialization queue — causing incorrect frame ordering and broken interruptions
- Follows the same pattern applied in #4082 (Sarvam) and #4083 (Deepgram SageMaker)

### Changes per service

- **Fish Audio**: route audio through `append_to_audio_context()`, replace `_handle_interruption()` with `on_audio_context_interrupted()`
- **LMNT**: route audio through `append_to_audio_context()`, remove redundant `push_frame()` override
- **Neuphonic**: route audio through `append_to_audio_context()`, remove redundant `push_frame()` and `process_frame()` overrides (base class handles pause/resume via `pause_frame_processing=True`)
- **Rime NonJson**: route audio through `append_to_audio_context()`, remove redundant `push_frame()` override

## Testing

- `uv run ruff check` — no lint errors
- `uv run ruff format --check` — no formatting issues
- `uv run pytest tests/` — all 1117 tests pass

PR #4099, opened 2026-03-21
use audio context hook for InterruptibleTTSService

### LINKED ISSUE #3986: InterruptibleTTSService: _bot_speaking guard causes interruption to fail when TTS audio hasn't reached output transport yet
## Description

`InterruptibleTTSService._handle_interruption()` only disconnects/reconnects the WebSocket when `_bot_speaking=True`. However, interruption can occur **before** `BotStartedSpeakingFrame` is emitted — when TTS audio is still being synthesized or in transit to the output transport. In this case, `_bot_speaking` is `False`, the WebSocket is **not** disconnected, and the TTS server continues sending audio that gets played after the interruption.

This is a regression from the fix in #950 / PR #1272, which was later generalized into `InterruptibleTTSService`. The `_bot_speaking` guard was added as an optimization but introduces a race condition.

## Root Cause

In `tts_service.py` lines 1429-1433:

```python
async def _handle_interruption(self, frame: InterruptionFrame, direction: FrameDirection):
    if self._bot_speaking:          # <-- This guard is the problem
        await self._disconnect()
        await self._connect()
```

`_bot_speaking` is set to `True` only when `BotStartedSpeakingFrame` is received (line 1444), which happens when audio reaches the output transport and starts playing. If the user interrupts before that point, the guard fails and the WebSocket stays connected.

## Affected Services

All TTS services inheriting from `InterruptibleTTSService` that don't override `on_audio_context_interrupted()`:

- **Fish Audio** (`fish/tts.py`)
- **LMNT** (`lmnt/tts.py`)
- **Neuphonic** (`neuphonic/tts.py`)
- **Sarvam** (`sarvam/tts.py`)

## Services NOT Affected

These services correctly override `on_audio_context_interrupted()` to cancel server-side synthesis regardless of `_bot_speaking` state:

- **ElevenLabs** — sends `close_context` message via WebSocket
- **Rime** — sends clear message to server
- **Deepgram** — sends `{"type": "Clear"}` to server

## Steps to Reproduce

1. Configure a pipeline with any affected TTS service (e.g., Fish Audio) and `allow_interruptions=True`
2. Send a prompt that triggers a multi-sentence LLM response
3. Speak (trigger `UserStartedSpeakingFrame`) **before** the first sentence's TTS audio starts playing (before `BotStartedSpeakingFrame`)
4. Observe that `InterruptionFrame` fires, but the TTS audio still arrives and plays afterward

### Log Evidence

```
49.216  [OUTPUT] UserStartedSpeakingFrame DOWNSTREAM    ← User interrupts
49.237  [OUTPUT] InterruptionFrame DOWNSTREAM            ← Interruption fires
49.239  on_assistant_turn_stopped                        ← LLM response complete
49.419  [OUTPUT] BotStartedSpeakingFrame DOWNSTREAM      ← Audio plays AFTER interruption (bug)
```

Note: `BotStartedSpeakingFrame` (49.419) occurs **after** `InterruptionFrame` (49.237), meaning `_bot_speaking` was `False` when `_handle_interruption` ran.

## Suggested Fix

**Option A**: Remove the `_bot_speaking` guard — always disconnect/reconnect on interruption:

```python
async def _handle_interruption(self, frame: InterruptionFrame, direction: FrameDirection):
    await self._disconnect()
    await self._connect()
```

**Option B**: Override `on_audio_context_interrupted()` in each affected service (similar to ElevenLabs/Rime/Deepgram) to cancel server-side synthesis. This is called from `super()._handle_interruption()` regardless of `_bot_speaking`.

## Environment

- Pipecat version: 0.0.dev8273
- TTS: Fish Audio (`FishAudioTTSService`)
- Transport: Twilio Media Streams (WebSocket) via Daily
- Python: 3.12

omChauhanDev:

Fixes #3986 

### Issue :
The `_bot_speaking` guard in `InterruptibleTTSService._handle_interruption()` skips websocket disconnect/reconnect when `BotStartedSpeakingFrame` hasn't reached the TTS processor yet. If a user interrupts while audio is still being synthesized or in-transit, the TTS server keeps streaming stale audio into the next response. 

Note: #4090 partially addressed this by routing audio through append_to_audio_context(), so stale audio is discarded when no active context exists. However, the server still continues synthesizing unused audio (wasted cost/bandwidth), and old audio can leak into the next response once a new audio context becomes active.                                                          
                                                                                                                             
### Approach : 
Replaced the `_bot_speaking` guard with an `on_audio_context_interrupted()` override - the same hook ElevenLabs, Rime, & Deepgram already use. Audio contexts exist from synthesis start to playback end, so this fires exactly when needed & stays silent when the bot is idle (preserving the original optimization against unnecessary reconnects from VAD noise).                 

**Changes** : 
- tts_service.py: removed _bot_speaking, _handle_interruption, process_frame override; added on_audio_context_interrupted with disconnect/reconnect              
- fish/tts.py: added super() call in existing on_audio_context_interrupted override

### DIFF
--- a/changelog/4099.fixed.md
+++ b/changelog/4099.fixed.md
@@ -1,0 +1,1 @@
+- Fixed `InterruptibleTTSService` interruption handling by replacing the `_bot_speaking` guard with `on_audio_context_interrupted`, so the websocket is always reconnected when audio is in-transit. Affects Fish Audio, LMNT, Neuphonic, and Sarvam TTS services.
--- a/src/pipecat/services/fish/tts.py
+++ b/src/pipecat/services/fish/tts.py
@@ -362,6 +362,7 @@ class FishAudioTTSService(InterruptibleTTSService):
 
     async def on_audio_context_interrupted(self, context_id: str):
         """Stop all metrics when audio context is interrupted."""
+        await super().on_audio_context_interrupted(context_id)
         await self.stop_all_metrics()
 
     async def _receive_messages(self):
--- a/src/pipecat/services/tts_service.py
+++ b/src/pipecat/services/tts_service.py
@@ -31,7 +31,6 @@ from pipecat.audio.utils import create_stream_resampler
 from pipecat.frames.frames import (
     AggregatedTextFrame,
     AggregationType,
-    BotStartedSpeakingFrame,
     BotStoppedSpeakingFrame,
     CancelFrame,
     EndFrame,
@@ -1478,42 +1477,15 @@ class WebsocketTTSService(TTSService, WebsocketService):
 class InterruptibleTTSService(WebsocketTTSService):
     """Websocket-based TTS service that handles interruptions without word timestamps.
 
-    Designed for TTS services that don't support word timestamps. Handles interruptions
-    by reconnecting the websocket when the bot is speaking and gets interrupted.
+    Designed for TTS services that don't support word timestamps or native cancel
+    APIs. Handles interruptions by reconnecting the websocket when an active audio
+    context gets interrupted.
     """
 
-    def __init__(self, **kwargs):
-        """Initialize the Interruptible TTS service.
-
-        Args:
-            **kwargs: Additional arguments passed to the parent WebsocketTTSService.
-        """
-        super().__init__(**kwargs)
-
-        # Indicates if the bot is speaking. If the bot is not speaking we don't
-        # need to reconnect when the user speaks. If the bot is speaking and the
-        # user interrupts we need to reconnect.
-        self._bot_speaking = False
-
-    async def _handle_interruption(self, frame: InterruptionFrame, direction: FrameDirection):
-        await super()._handle_interruption(frame, direction)
-        if self._bot_speaking:
-            await self._disconnect()
-            await self._connect()
-
-    async def process_frame(self, frame: Frame, direction: FrameDirection):
-        """Process frames with bot speaking state tracking.
-
-        Args:
-            frame: The frame to process.
-            direction: The direction of frame processing.
-        """
-        await super().process_frame(frame, direction)
-
-        if isinstance(frame, BotStartedSpeakingFrame):
-            self._bot_speaking = True
-        elif isinstance(frame, BotStoppedSpeakingFrame):
-            self._bot_speaking = False
+    async def on_audio_context_interrupted(self, context_id: str):
+        """Disconnect and reconnect the websocket to stop server-side synthesis."""
+        await self._disconnect()
+        await self._connect()
 
 
 class WebsocketWordTTSService(WebsocketTTSService):


yuki901:
Thank you very much!!
Btw, since #4090, all `InterruptibleTTSService` subclasses route audio through `append_to_audio_context()`. After an interruption, `_handle_interruption` replaces `_audio_contexts` with a fresh empty dict, so stale audio from `_receive_messages()` is silently discarded — the user-facing bug in #3986 seems already fixed regardless of the `_bot_speaking` guard.

This PR's fix would still prevent the server from continuing to synthesize and stream unused audio (cost/bandwidth), but is that the intended scope? Does the PR description account for #4090?

omChauhanDev:
Hey @yuki901, nice catch - you're right that #4090 handles the immediate window. After interruption, `_create_audio_context_task()` replaces `_audio_contexts` with a fresh dict and `_playing_context_id` is reset to `None`, so stale audio from `_receive_messages()` hits `append_to_audio_context(None, ...)` and gets silently dropped. The audio-plays-right-after-interruption symptom is largely gone.

That said, this PR still covers two things #4090 doesn't:

1. **Server-side waste** - without disconnecting, the TTS server keeps synthesizing and streaming audio nobody will use. That's wasted compute, bandwidth, and API cost.

2. **Audio crossover into the next response** - there's a subtler window where old audio can leak. Once the next LLM response starts and a new audio context is created, `_playing_context_id` gets set to the new context ID. If old audio from the still-connected server arrives at that point, `get_active_audio_context_id()` returns the new ID, and `append_to_audio_context` routes old audio into the new context. The disconnect/reconnect prevents this by clearing server-side state entirely.

Happy to update the PR description to reference #4090 & clarify the scope. Thanks for flagging it!

markbackman:
Tagging @filipi87 to take a look.

filipi87 (changes requested):
Hi @omChauhanDev,

As discussed above, all TTS services now route audio through the audio context. When an interruption occurs, all audio contexts are canceled. As a result, even if audio arrives afterward, `append_to_audio_context` may be called with an invalid or undefined `context_id`, and the audio is effectively discarded. In this case, we log a debug message indicating that the audio was dropped.

The behavior of reconnecting only when the bot is speaking was introduced as an optimization, since reestablishing the connection can sometimes take a couple of seconds. Keeping this behavior helps maintain a better user experience.

For that reason, I believe it still makes sense to preserve the current approach, only reconnecting when the bot is actively speaking.

That said, I do see a small window where stale audio could leak through. However, this would likely only occur in the case where `run_tts` has been invoked but the `BotStartedSpeakingFrame` has not yet been received. This seems to be the only scenario where the issue could arise.

If that’s the case, we could address it more directly within the `InterruptibleTTSService`, by treating the presence of any audio context as an indication that the bot has started speaking. Something along these lines should be sufficient to prevent the race condition:

```
    if isinstance(frame, TTSStartedFrame):
        self._bot_speaking = True
```

filipi87:
> If that’s the case, we could address it more directly within the InterruptibleTTSService, by treating the presence of any audio context as an indication that the bot has started speaking. Something along these lines should be sufficient to prevent the race condition:

This has been fixed in this PR:
- https://github.com/pipecat-ai/pipecat/pull/4145

PR #4100, opened 2026-03-22
Add failover strategy for service switcher with TTS error handling

yuki901:
#### Description

This PR adds a new `ServiceSwitcherStrategyFailover` class that enables automatic retry and failover behavior for TTS services. When a TTS service fails, the switcher will retry the same service up to a configurable limit before switching to the next available service.

**Key changes:**

1. **New `ServiceSwitcherStrategyFailover` class** (`service_switcher.py`):
   - Implements per-service retry tracking with configurable `max_retries` parameter
   - Automatically switches to the next service after retries are exhausted
   - Tracks failed services to prevent re-attempting them
   - Fires `on_service_switched` event when failover occurs
   - Supports manual service switching via `ManuallySwitchServiceFrame`

2. **TTS error handling improvements**:
   - New `TTSErrorFrame` class (`frames.py`) that extends `ErrorFrame` with `text` and `tts_context_id` fields, allowing the service switcher to retry with the original text
   - Modified `TTSService.tts_process_generator()` to promote plain `ErrorFrame` to `TTSErrorFrame` with original text context
   - Updated `ServiceSwitcher.push_frame()` to intercept `TTSErrorFrame` and trigger retry/failover logic

3. **ServiceSwitcher API enhancement**:
   - Added `strategy_kwargs` parameter to `ServiceSwitcher.__init__()` to pass configuration to strategy constructors

4. **Comprehensive test coverage**:
   - Unit tests for failover strategy initialization, retry logic, and service exhaustion
   - Integration tests with `MockFailingTTSService` covering retry-then-succeed, failover, error propagation, and text preservation scenarios
   - Test for non-TTS error pass-through behavior

The implementation ensures that transient TTS failures are handled gracefully without losing the original text, enabling robust multi-service TTS configurations.

https://claude.ai/code/session_01RpHvxvN9aR4469Jv29pDcR

### DIFF
--- a/src/pipecat/services/fish/tts.py
+++ b/src/pipecat/services/fish/tts.py
@@ -392,6 +392,9 @@ class FishAudioTTSService(InterruptibleTTSService):
                                 )
                             else:
                                 logger.debug(f"Fish Audio session finished: {reason}")
+                            context_id = self.get_active_audio_context_id()
+                            if context_id:
+                                await self.remove_audio_context(context_id)
 
             except Exception as e:
                 await self.push_error(error_msg=f"Unknown error occurred: {e}", exception=e)


markbackman:
FYI: We added a failover strategy in v0.0.105:
https://github.com/pipecat-ai/pipecat/pull/3870

You can see the code here:
https://github.com/pipecat-ai/pipecat/blob/main/src/pipecat/pipeline/service_switcher.py#L158

You can see the history in the PR that we opted to keep logic, like retry logic out of the switcher and instead add event handlers that application code can hook into.

yuki901:
Hi! Thank you for the clarification. Appreciate a lot.

Actually I've been trying to implement TTS failover with text re-dispatch using the existing callbacks (Fish TTS is unstable lately), and I hit a gap I wanted to flag.

When a TTS service fails, the `ErrorFrame` that reaches `handle_error` doesn't carry the original text that was being synthesized. This means application code in `on_service_switched` has no way to know what to resend to the new service.

I tried tracking text via `on_tts_request`, but it only fires inside `_synthesize_text` — so if multiple `TTSSpeakFrame`s are already queued in the service's internal `__process_queue`, only the one currently being synthesized has been recorded. The rest haven't hit `on_tts_request` yet and are invisible to application code.

Would it make sense to include the original text in the error path — e.g. a `TTSErrorFrame(text=...)` subclass, or passing the text to the `on_error` event handler? That would let apps implement retry/re-dispatch without any framework-level retry logic, which seems aligned with the design intent here.

Happy to contribute a PR if there's an approach that fits.

PR #4101, opened 2026-03-22
enable pause_frame_processing for CartesiaTTSService

### LINKED ISSUE #4038: post_actions: FunctionActionFrame executes before TTS finishes due to concurrent audio context
### pipecat version

v0.0.105

### Python version

3.13

### Operating System

Ubuntu

### Issue description

After upgrading to pipecat v0.0.105, `post_actions` with `{"type": "function"}` execute while TTS is still speaking the LLM-generated response. The FE receives the event (e.g. "conversation ended") and closes the screen before the user hears the bot's closing message.

### Reproduction steps

1. Create a pipecat-flows node where the LLM generates speech based on `role_messages`/`task_messages`
2. Add `post_actions` with `{"type": "function", "handler": my_handler}` — this handler sends a frontend event signaling the conversation is over
3. Use `CartesiaTTSService` on pipecat v0.0.105
4. Observe that `my_handler` executes while TTS is still speaking the LLM-generated response — the `FunctionActionFrame` passes through TTS without waiting for audio to finish

### Expected behavior

Post-action `FunctionActionFrame` should be processed after TTS finishes speaking the LLM response, not concurrently. I fixed the issue by setting the private `_pause_frame_processing`, which resolved the issue on our end. However, this relies on a private API that could change without notice.

### Actual behavior

After upgrading to pipecat v0.0.105, `post_actions` with `{"type": "function"}` execute while TTS is still speaking the LLM-generated response. The FE receives the event (e.g. "conversation ended") and closes the screen before the user hears the bot's closing message.

### Logs

```shell

```

omChauhanDev:

Fixes: #4038 

**Change** : Enabled `pause_frame_processing` for `CartesiaTTSService` so downstream frames wait until audio playback completes before reaching the pipeline sink.

### DIFF
--- a/changelog/4101.fixed.md
+++ b/changelog/4101.fixed.md
@@ -1,0 +1,1 @@
+- Fixed `CartesiaTTSService` allowing downstream frames to execute before audio playback completes by enabling `pause_frame_processing`.
--- a/src/pipecat/services/cartesia/tts.py
+++ b/src/pipecat/services/cartesia/tts.py
@@ -334,7 +334,7 @@ class CartesiaTTSService(WebsocketTTSService):
             text_aggregation_mode=text_aggregation_mode,
             aggregate_sentences=aggregate_sentences,
             push_text_frames=False,
-            pause_frame_processing=False,
+            pause_frame_processing=True,
             sample_rate=sample_rate,
             push_start_frame=True,
             text_aggregator=text_aggregator,


markbackman:
@omChauhanDev can you double check the issue you linked?

@filipi87 I believe you intentionally made it such that CartesiaTTSService doesn't pause frame processing, correct?

filipi87:
Hi @omChauhanDev,

The reason we had this `pause_frame_processing`  before is that we weren’t able to handle multiple contexts at the same time, and enabling it would cause the frames and words to become completely out of order.

This is no longer the case with Cartesia and the base `TTSService`, which now supports multiple contexts. We can request them simultaneously, improving latency while still respecting all requirements.

I am closing this PR since the current behavior is expected. If anything doesn’t work as intended, please let us know so we can look into it.

omChauhanDev:
oh I see...saw someone flagging that with CartesiaTTSService, post_actions were executing before TTS finishes, so I thought it might be bcz we weren't pausing frame processing until audio completes. But since that's intentional for concurrent context support & post_actions are a part of pipecat-flows, that makes sense. Thanks for the clarification!

PR #4103, opened 2026-03-22
fix: merge Gemini parallel tool call responses

### LINKED ISSUE #3992: GeminiLLMAdapter merges functionCall messages without merging matching functionResponse messages
### pipecat version

0.0.104

### Python version

3.12

### Operating System

macOS

### Issue description

`GeminiLLMAdapter._merge_parallel_tool_calls_for_thinking()` currently merges `model` messages containing `function_call` parts when thought signatures are present, but it does **not** merge the matching `user` messages containing `function_response` parts.

This can corrupt otherwise valid sequential tool history into an invalid Gemini/Vertex function-call turn.

Related code path:
- `pipecat/adapters/services/gemini_adapter.py`
- `_merge_parallel_tool_calls_for_thinking()`

Related, but different, from #3557:
- #3557 is about missing `thought_signature`
- this issue happens even when signatures are preserved, because the adapter re-groups messages incorrectly

Google docs:
- https://ai.google.dev/gemini-api/docs/thought-signatures

### Reproduction steps

Take a message history shaped like this:

1. `model(functionCall fc1 + thought_signature)`
2. `user(functionResponse fr1)`
3. `model(functionCall fc2)`
4. `user(functionResponse fr2)`

These are two sequential single-tool turns.

Today, `_merge_parallel_tool_calls_for_thinking()` can turn that into:

1. `model(functionCall fc1, functionCall fc2)`
2. `user(functionResponse fr1)`
3. `user(functionResponse fr2)`

The next request to Gemini / Vertex then fails with:

```text
400 Bad Request
Please ensure that the number of function response parts is equal to the number of function call parts of the function call turn.
```

We hit this with `GoogleVertexLLMService` after otherwise normal single-tool turns.

### Expected behavior

If Pipecat decides to merge parallel tool calls into a single `model` turn, it should also merge the corresponding `user function_response` messages into a single `user` turn, preserving the same grouping expected by Gemini.

For example:

1. `model(functionCall fc1, functionCall fc2)`
2. `user(functionResponse fr1, functionResponse fr2)`

Alternatively, if Pipecat cannot safely prove the calls are truly parallel, it should not merge them at all.

### Actual behavior

Pipecat merges only the `function_call` side and leaves `function_response` messages split, producing an invalid call/response structure for Gemini.

### Suggested fix

In `GeminiLLMAdapter._merge_parallel_tool_calls_for_thinking()`:

- only merge contiguous tool-call groups
- when merging `model` `function_call` parts, also merge the immediately corresponding `user` `function_response` parts
- do not scan through arbitrary text messages and then retroactively regroup tool turns

### Suggested regression tests

1. Parallel case:
   - `M(fc1+sig), U(fr1), M(fc2), U(fr2)`
   - expected -> `M(fc1,fc2), U(fr1,fr2)`

2. Sequential case:
   - `M(fc1+sigA), U(fr1), M(fc2+sigB), U(fr2)`
   - expected -> no cross-turn merge

3. Mixed text boundary:
   - `M(fc1+sig), U(fr1), M(text), M(fc2), U(fr2)`
   - expected -> no merge across the text boundary

omChauhanDev:

Fixes #3992 

### Issue :     
When thinking is enabled, `_merge_parallel_tool_calls_for_thinking()` merges model `function_call` messages into a single turn but left the corresponding user `function_response` messages split - Gemini rejects this with a 400                                                   

### Change : 
The inner scan loop now also collects & merges `function_response` parts, & stops at any non-tool message boundary instead of scanning past it. Also, added tests covering parallel, sequential, text boundary, batch ordering, & mixed parallel+sequential cases


markbackman:
Tagging @kompfner to take a look.

kompfner:
> left the corresponding user function_response messages split - Gemini rejects this with a 400

Huh, wonder if this is a new Gemini requirement? This was almost certainly not the case when we first introduced thinking support for Gemini...

src/pipecat/adapters/services/gemini_adapter.py
kompfner: This docstring is a bit out of date now. Let's update it throughout to discuss the parallel-tool-call-responses-must-be-grouped requirement as well, rather than only mentioning it at the end.

src/pipecat/adapters/services/gemini_adapter.py
```diff
@@ -457,9 +457,9 @@ def _merge_parallel_tool_calls_for_thinking(
         - Sequential tool calls: each have their own thought_signature
 
         Algorithm: A tool call message with a thought_signature starts a new
-        parallel group. Any tool call messages after it without a
-        thought_signature get merged into that group, regardless of what
-        messages appear in between.
+        parallel group. Subsequent tool call messages without a
```
kompfner: We had built this algorithm initially trying to make the _fewest_ assumptions about how our Pipecat context was structured. That meant allowing for the possibility that other kinds of messages were interleaved between any tool call messages (even though Gemini, I _believe_, doesn't allow that).

I think we can aim to keep that flexibility, even with the new requirement of bundling tool results together into a single message.

I think we can (and maybe have to) assume that we'll get all the tool results for tool request group A before a group B starts. But we can limit ourselves to _just_ that assumption.

Maybe we could tweak your algorithm just a bit:
- Loop through the messages (no change)
- If we hit a tool call with a thought signature, start a group (no change)
- Scan forward (no change), and...
  - If you encounter a tool call with no thought signature, add it to `merged_parts` (no change)
  - If you hit a tool response, add it to `merged_response_parts` (no change)
  - If you hit any other non-tool-call or non-tool-response message, add it to `other_messages` and keep scanning (changed; similar to previous code)
- At the end of the group, output `merged_parts`, then `merged_response_parts`, then other messages (changed; similar to previous code)

This is mostly academic/defensive—it's sort of "just in case" our context has other stuff interleaved between tool calls and responses (even though, again, I don't think Gemini supports that).

kompfner: Went ahead and pushed a change that addresses this.

kompfner:
> > left the corresponding user function_response messages split - Gemini rejects this with a 400
> 
> Huh, wonder if this is a new Gemini requirement? This was almost certainly not the case when we first introduced thinking support for Gemini...

As noted in the issue (#3992), this doesn't happen with any of today's Gemini models when using the Gemini API, but *does* occur when using the stricter Vertex.

kompfner (approved):

PR #4104, opened 2026-03-22
Allow defining whether to insert silence in the output transport.

filipi87:
## Summary                             
- Added `audio_out_insert_silence` parameter to `TransportParams` (defaults to `True`) to control behavior when the audio output queue is empty.                                                                                                                            
- When `True` (default), the transport inserts silence frames, preserving existing behavior.
- When `False`, the transport waits for audio data to become available instead of inserting silence, which is useful for scenarios where uninterrupted audio playback is required without artificial silence gaps.   
  - For example, when using `TavusVideoService`, we are always receiving audio from Tavus (or silence). In this case, the transport should never inject additional silence, because doing so can cause small glitches when the bot is speaking.


src/pipecat/transports/smallwebrtc/transport.py
```diff
@@ -79,14 +79,17 @@ class RawAudioTrack(AudioStreamTrack):
     supporting queued audio data with proper synchronization.
     """
 
-    def __init__(self, sample_rate):
+    def __init__(self, sample_rate: int, insert_silence: bool = True):
```
filipi87: We will need to apply similar changes to the `DailyTransport`, but we may need to modify `daily-python` to do so, if I’m not mistaken.

src/pipecat/transports/daily/transport.py
```diff
@@ -831,7 +835,14 @@ async def start(self, frame: StartFrame):
...
 
         if self._params.audio_out_enabled and not self._microphone_track:
-            audio_source = CustomAudioSource(self._out_sample_rate, self._params.audio_out_channels)
+            logger.debug(
+                f"Creating custom audio source, auto silence {self._params.audio_out_insert_silence}"
+            )
+            audio_source = CustomAudioSource(
+                self._out_sample_rate,
+                self._params.audio_out_channels,
+                self._params.audio_out_insert_silence,
```
filipi87: This will require a new version of `daily-python`.

examples/foundational/21a-tavus-video-service.py
```diff
@@ -50,6 +51,7 @@
         video_out_is_live=True,
         video_out_width=1280,
         video_out_height=720,
+        audio_out_auto_silence=False,
```
aconchillo, line 54: I don't think we need this here.

examples/foundational/21a-tavus-video-service.py
```diff
@@ -42,6 +42,7 @@
...
```
markbackman, line 45: This isn't critical for right now, but this is a random, but critical value to set on the transport. It would be easy to overlook setting it. Ideally, we could auto-configure the transport to do this.

I think it's OK for now, but maybe we can figure out a way to improve the dev ex in a follow up PR. (The tricky thing is that the transport is configured first in the pipeline 🤔.)

src/pipecat/transports/base_transport.py
```diff
@@ -144,6 +146,7 @@ class TransportParams(BaseModel):
     audio_out_mixer: Optional[BaseAudioMixer | Mapping[Optional[str], BaseAudioMixer]] = None
     audio_out_destinations: List[str] = Field(default_factory=list)
     audio_out_end_silence_secs: int = 2
+    audio_out_auto_silence: bool = True
```
aconchillo, line 149: It seems this is Daily specific. SO maybe just add it to `DailyParams`

examples/foundational/21a-tavus-video-service.py
aconchillo: Oh, I see we were adding silence in SmallWebRTC. Nevermind.

src/pipecat/transports/base_transport.py
aconchillo: Ignore.

src/pipecat/transports/smallwebrtc/transport.py
aconchillo: Do you want to rename this to `auto_silence` as well? Not really necessary

aconchillo (approved):

examples/foundational/21a-tavus-video-service.py
filipi87: Agreed. We can figure out how to improve and change this default behavior in a follow up PR.

src/pipecat/transports/daily/transport.py
filipi87: Done.

examples/foundational/21a-tavus-video-service.py
aconchillo: Yeah, I thought about the same, but not sure how we would do this, since the transport is not really tied to the service, so you need to configure it the way the service needs it.

Maybe one way would be for the service to have some static value like: `TavusVideoService.transport_params`.

markbackman: In other cases, we push a frame to configure the processor, but in this case, the transport would have already started when the TavusVideoService is running. That makes this harder.

src/pipecat/transports/smallwebrtc/transport.py
filipi87: done.

PR #4107, opened 2026-03-23
fix: omit unsupported output_modalities from Azure Realtime response.create

### LINKED ISSUE #4106: AzureRealtimeLLMService sends response.output_modalities which Azure doesn't support
There's an error using AzureRealtimeLLMService with Azure's GPT realtime API due to an unsupported parameter 'response.output_modalities' being sent, causing request failure.

Related thread: https://discord.com/channels/1239284677165056021/1484272187010715719/1484272187010715719

voidborne-d:
## Summary

Fixes #4106.

`AzureRealtimeLLMService` inherits `_create_response` from `OpenAIRealtimeLLMService`, which sends `response.output_modalities` in every `response.create` event. Azure's Realtime API does not support this parameter and rejects the request with an error.

## Changes

Override `_create_response` in `AzureRealtimeLLMService` to send a plain `ResponseCreateEvent` without the `output_modalities` field, while preserving all other behavior (conversation setup, session update, metrics).

## Testing

- Verified the fix addresses the exact error described in #4106
- The override follows the same pattern as the existing `_connect` override in the same class
- All other `response.create` behavior (conversation setup, metrics, session update) is preserved identically

### DIFF
--- a/src/pipecat/services/azure/realtime/llm.py
+++ b/src/pipecat/services/azure/realtime/llm.py
@@ -10,6 +10,7 @@ from dataclasses import dataclass
 
 from loguru import logger
 
+from pipecat.frames.frames import LLMFullResponseStartFrame
 from pipecat.services.openai.realtime import events
 from pipecat.services.openai.realtime.llm import OpenAIRealtimeLLMService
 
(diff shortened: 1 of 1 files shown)


jamsea:
@markbackman at the moment no, but I can do some digging on Friday, or rope in the Discord person too 👀

PR #4113, opened 2026-03-23
fix(inworld): fallback to full text when TTS timestamps are not received

ianbbqzy:
- When timestamps are not returned by the TTS server, no TTSTextFrames reach the assistant context aggregator, so the agent's turn is committed as empty in the LLM conversation context.
- This adds a fallback for both the HTTP and WebSocket services: if a TTS request completes without any timestamps, the full input text is pushed as a single TTSTextFrame so the conversation context reflects what the agent spoke.
- Note: without word-level timestamps, interrupted turns commit the full text rather than only the portion that was spoken. This is a workaround until all supported languages have full timestamp support in the coming weeks

Depends on #4028.

### DIFF
--- a/src/pipecat/services/inworld/tts.py
+++ b/src/pipecat/services/inworld/tts.py
@@ -28,6 +28,7 @@ from typing import (
     Mapping,
     Optional,
     Self,
+    Set,
     Tuple,
 )
 
@@ -51,6 +52,7 @@ except ModuleNotFoundError as e:
     raise Exception(f"Missing module: {e}")
 
 from pipecat.frames.frames import (
+    AggregationType,
     CancelFrame,
     EndFrame,
     ErrorFrame,
@@ -60,6 +62,7 @@ from pipecat.frames.frames import (
     TTSAudioRawFrame,
     TTSStartedFrame,
     TTSStoppedFrame,
+    TTSTextFrame,
 )
 from pipecat.processors.frame_processor import FrameDirection
 from pipecat.services.tts_service import TextAggregationMode, TTSService, WebsocketTTSService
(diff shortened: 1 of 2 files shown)


src/pipecat/services/inworld/tts.py
```diff
@@ -776,12 +807,35 @@ async def _close_context(self, context_id: str):
...
     async def on_audio_context_interrupted(self, context_id: str):
         """Callback invoked when an audio context has been interrupted."""
+        await self._push_fallback_text(context_id)
         await self._close_context(context_id)
 
     async def on_audio_context_completed(self, context_id: str):
         """Callback invoked when an audio context has been completed."""
         await self._close_context(context_id)
 
+    async def _push_fallback_text(self, context_id: str):
```
filipi87: Nit: rename it to `_maybe_push_fallback_text`, since we always invoke it and the logic that decides whether to push the fallback text lives here.

src/pipecat/services/inworld/tts.py
```diff
@@ -955,6 +1011,8 @@ async def _receive_messages(self):
             if timestamp_info:
                 word_times = self._calculate_word_times(timestamp_info)
                 if word_times:
+                    if ctx_id:
+                        self._contexts_with_timestamps.add(ctx_id)
```
filipi87, line 1046: Is there a scenario where you might receive timestamps for some words but not others, and end up adding duplicated text to the context ?

ianbbqzy: If it hypothetically did happen (server bug), the current behavior is the safe choice: once we see any timestamp, we mark the context as "had timestamps" and skip the fallback entirely. So there would never be duplicated text. The context might have incomplete text but it's unlikely.

src/pipecat/services/inworld/tts.py
ianbbqzy: SG, renamed!

filipi87 (approved):
LGTM!

PR #4120, opened 2026-03-24
Added cleanup() method to VADAnalyzer base class

filipi87:
## Summary
- Added `cleanup()` method to `VADAnalyzer` base class (no-op by default) and `VADController`, ensuring VAD analyzers are properly cleaned up when no longer needed                                                                                                     
- Fixed the Krisp Viva example.


markbackman (approved):
LGTM

PR #4122, opened 2026-03-24
Invoking on_turn_context_created when we receive a TTSSpeakFrame.

filipi87:
This is a follow-up PR, for the changes introduced by Inworld here:
- https://github.com/pipecat-ai/pipecat/pull/4013


markbackman (approved):
Makes sense. This is nice as it will allow TTSSpeakFrames to be spoken more naturally. Today, they're treated as one-offs. All TTS providers that use a context ID will benefit from this, right?

filipi87:
> All TTS providers that use a context ID will benefit from this, right?

Yep, this is already how it is working for all the TTS providers. 👍

PR #4125, opened 2026-03-24
Gemini Live: fix EndFrame-deferral hang

kompfner:
## Summary

We observed Gemini Live's `EndFrame` deferral hanging indefinitely: when an `EndFrame` arrives while the bot is mid-response, it is deferred until `turn_complete` is received — but `turn_complete` never arrived, so the pipeline hung forever.

Two fixes:

- **Possible root-cause fix**: `turn_complete` messages were previously only handled if they also contained `usage_metadata`. Now `turn_complete` is always handled, with `usage_metadata` processing only when available. (We have not actually observed a `turn_complete` without `usage_metadata`, so this is a theoretical fix.)
- **Fallback timeout**: The deferred `EndFrame` now has a 30-second safety timeout. If `turn_complete` hasn't arrived by then, the `EndFrame` is released anyway with a warning log.

## Testing

- Run any Gemini Live example that ends the conversation via `EndFrame` while the bot is responding, and verify the pipeline shuts down cleanly
- Verify normal conversation flow is unaffected (`turn_complete` with `usage_metadata` still processed correctly)

### DIFF
--- a/changelog/4125.fixed.md
+++ b/changelog/4125.fixed.md
@@ -1,0 +1,1 @@
+- Fixed Gemini Live pipeline hanging indefinitely when an `EndFrame` was deferred while waiting for the bot to finish responding and `turn_complete` never arrived. As a possible root-cause fix, `turn_complete` messages are now handled even if they lack `usage_metadata`. As a fallback, the deferred `EndFrame` now has a 30-second safety timeout.
--- a/src/pipecat/services/google/gemini_live/llm.py
+++ b/src/pipecat/services/google/gemini_live/llm.py
@@ -1181,6 +1181,7 @@ class GeminiLiveLLMService(LLMService):
         """Release a deferred EndFrame and cancel the deferral timeout."""
         if self._end_frame_pending_bot_turn_finished:
             self._cancel_end_frame_deferral_timeout()
+            self._bot_is_responding = False
             frame = self._end_frame_pending_bot_turn_finished
             self._end_frame_pending_bot_turn_finished = None
             await self.queue_frame(frame)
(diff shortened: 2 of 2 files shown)


vipyne (approved):
lgtm.
not blocking, but just curious why 30 seconds for the timeout?

kompfner:
> lgtm. not blocking, but just curious why 30 seconds for the timeout?

Long enough to not cut the bot off during an intentionally-very-long response, but not so long that an errant bot stays a zombie for too long.

I figured it should be a fairly long timeout...there's not _too_ much downside of an occasionally stray bot sticking around for awhile, compared with a bot being cut off while it's saying its parting words.

kompfner:
> > lgtm. not blocking, but just curious why 30 seconds for the timeout?
> 
> Long enough to not cut the bot off during an intentionally-very-long response, but not so long that an errant bot stays a zombie for too long.
> 
> I figured it should be a fairly long timeout...there's not _too_ much downside of an occasionally stray bot sticking around for awhile, compared with a bot being cut off while it's saying its parting words.

We could maybe make this configurable someday, if the need arises, i.e. if a developer finds that they're likely to have really really long "parting words". But this is the kind of thing I'd vote not to build if/until we need it.

vipyne:
> > > lgtm. not blocking, but just curious why 30 seconds for the timeout?
> > 
> > 
> > Long enough to not cut the bot off during an intentionally-very-long response, but not so long that an errant bot stays a zombie for too long.
> > I figured it should be a fairly long timeout...there's not _too_ much downside of an occasionally stray bot sticking around for awhile, compared with a bot being cut off while it's saying its parting words.
> 
> We could maybe make this configurable someday, if the need arises, i.e. if a developer finds that they're likely to have really really long "parting words". But this is the kind of thing I'd vote not to build if/until we need it.

sounds reasonable to me.  honestly, more than 30 seconds of "parting words" is probably too much for a human to pay attention to toward the end of a call, so this may encourage good behavior on the part of developers.

and like you said, build configurability if/when needed.

PR #4126, opened 2026-03-24
Fix TTS flush creating phantom contexts on ElevenLabs

### LINKED ISSUE #4114: ElevenLabsTTSService WebSocket Error: Maximum simultaneous contexts exceeded (1008 policy violation)
### pipecat version

0.0.106

### Service Name

ElevenLabs

### Service or model version

eleven_flash_v2_5

### Issue Description

When using `ElevenLabsTTSService`, the pipeline eventually crashes with a WebSocket `1008 (policy violation)` error returned by ElevenLabs. The specific error message from the API is: `"Maximum simultaneous contexts per WebSocket connection exceeded (5). Please close an existing context before opening a new one."` 

This appears to happen when active TTS contexts pile up on the same WebSocket connection. It seems that during rapid user interruptions (or high concurrency), `pipecat` might not be sending the proper cancellation/close signals to ElevenLabs to terminate the old TTS contexts. As a result, the connection hits the hard limit of 5 active contexts, and ElevenLabs forcefully drops the WebSocket connection.

Error Trace: https://into-ai.sentry.io/share/issue/177b482be84b447d9aa77f75e185b281/

### Reproduction Steps

1. Initialize a voice pipeline using `ElevenLabsTTSService` for TTS.
2. Start a conversation where the bot begins generating and streaming TTS audio.
3. Rapidly interrupt the bot (triggering a flush/cancel) multiple times in a row before the TTS audio can fully complete.
4. Observe the pipeline crash after roughly 5 rapid interruptions when the WebSocket context limit is reached.

### Expected Behavior

When an interruption occurs and audio is flushed, `pipecat` should proactively send the necessary signal to ElevenLabs to close or cancel that specific active context over the WebSocket. The concurrent context count should remain well below the 5-context limit, keeping the WebSocket connection stable.

### Actual Behavior

Interrupted or abandoned contexts appear to remain open on the ElevenLabs server. Once 5 contexts stack up, ElevenLabs terminates the WebSocket connection with a `1008` code, resulting in an `ErrorFrame` being pushed and the pipeline crashing.

### Error Logs

```shell
2026-03-23 16:13:21.889 | ERROR | pipecat.processors.frame_processor:push_error_frame:744 - ElevenLabsTTSService#2 exception (/usr/local/lib/python3.13/site-packages/websockets/asyncio/connection.py:957): Unknown error occurred: received 1008 (policy violation) Maximum simultaneous contexts per WebSocket connection exceeded (5). Please close an existing context before opening a n; then sent 1008 (policy violation) Maximum simultaneous contexts per WebSocket connection exceeded (5). Please close an existing context before opening a n

exception=ConnectionClosedError(Close(code=1008, reason='Maximum simultaneous contexts per WebSocket connection exceeded (5). Please close an existing context before opening a n'), Close(code=1008, reason='Maximum simultaneous contexts per WebSocket connection exceeded (5). Please close an existing context before opening a n'), True))
```

markbackman:
## Summary

- Fixed ElevenLabs WebSocket 1008 policy violation ("Maximum simultaneous contexts exceeded") caused by rapid user interruptions
- When an interruption arrives before any LLM text reaches `run_tts`, the turn context ID exists but was never registered via `create_audio_context`. The subsequent `LLMFullResponseEndFrame` triggers `on_turn_context_completed` → `flush_audio`, which sends a `{"context_id": "<uuid>", "flush": true}` message to ElevenLabs for a context it has never seen. ElevenLabs implicitly creates a server-side context for each such flush, and since no `close_context` is ever sent, they accumulate until the 5-context limit is exceeded.
- The fix guards `flush_audio` in base `TTSService.on_turn_context_completed` with `audio_context_available` so it only fires when text actually reached the TTS provider and the context was opened

## Testing

- Connect to an ElevenLabs-backed voice agent and rapidly interrupt (speak over the bot) 10-20 times in quick succession
- Previously this would cause a 1008 WebSocket disconnection; with the fix the connection remains stable

## Fixes

- Fixes #4114

### DIFF
--- a/changelog/4126.fixed.md
+++ b/changelog/4126.fixed.md
@@ -1,0 +1,1 @@
+- Fixed ElevenLabs WebSocket disconnections (1008 "Maximum simultaneous contexts exceeded") caused by rapid user interruptions. When interruptions arrived before any TTS text was generated, phantom contexts were created on the ElevenLabs server that were never closed, eventually exceeding the 5-context limit.
--- a/src/pipecat/services/tts_service.py
+++ b/src/pipecat/services/tts_service.py
@@ -682,7 +682,12 @@ class TTSService(AIService):
             await self.remove_audio_context(self._turn_context_id)
 
         # Flush any pending audio so the TTS service closes the current context.
-        await self.flush_audio(context_id=self._turn_context_id)
+        # Only flush if the context was actually opened (text reached run_tts).
+        # When an interruption arrives before any text flows, the turn context ID
+        # exists but was never registered via create_audio_context, so flushing
+        # would send a message for a context the provider never opened.
+        if self._turn_context_id and self.audio_context_available(self._turn_context_id):
+            await self.flush_audio(context_id=self._turn_context_id)
 
         # Reset the turn context ID
         self._turn_context_id = None


filipi87 (approved):
Really nice catch. The fix also makes a lot of sense, we don’t need to flush audio if the context was never actually opened. 🔥🙌

PR #4128, opened 2026-03-24
AssemblyAI: Add on_end_of_turn event handler

### LINKED ISSUE #4008: TranscriptionFrame And UserStoppedSpeakingFrame Ordering Not Preserved
### pipecat version

0.0.105

### Python version

3.11.5

### Operating System

MacOS 26.2 (25C56)

### Issue description

For Deepgram Flux and Assembly AI it seems the ordering of TranscriptionFrame and UserStoppedSpeakingFrame is not Preserved. There is even a comment in the AssemblyAI processor saying it should be: 
```
                # AAI is authoritative — emit UserStoppedSpeakingFrame immediately.
                # broadcast_frame pushes downstream (same queue as TranscriptionFrame
                # above, so ordering is preserved) and upstream.
```

However on testing I find that the UserStoppedSpeakingFrame reaches my subsequent processors first then the TranscriptionFrame gets received later. 

I have done some testing on this and I believe its down to UserStoppedSpeakingFrame being a SystemFrame. I tried this with another frame that I made "EndOfTurnDetectedFrame"
```
@dataclass
class EndOfTurnDetectedFrame(DataFrame):
    pass
```

Which seems to work fine. 

Here is a snippet from the DeepgramFluxProcessor and you can see that the TranscriptionFrame is first then the UserStoppedSpeakingFrame is broadcast after

```
        if not self._settings.min_confidence or average_confidence > self._settings.min_confidence:
            # EndOfTurn means Flux has determined the turn is complete,
            # so this TranscriptionFrame is always finalized
            await self.push_frame(
                TranscriptionFrame(
                    transcript,
                    self._user_id,
                    time_now_iso8601(),
                    self._settings.language,
                    result=data,
                    finalized=True,
                )
            )
        else:
            logger.warning(
                f"Transcription confidence below min_confidence threshold: {average_confidence}"
            )

        await self._handle_transcription(transcript, True, self._settings.language)
        await self.broadcast_frame(UserStoppedSpeakingFrame)
        await self._call_event_handler("on_end_of_turn", transcript)
```

### Reproduction steps

1. Make a test processor, in my case I called it OrderingProcessor
```
from loguru import logger
from pipecat.frames.frames import UserStoppedSpeakingFrame, Frame, TranscriptionFrame
from pipecat.processors.frame_processor import FrameProcessor, FrameDirection

class OrderingProcessor(FrameProcessor):
        if isinstance(frame, TranscriptionFrame):
            logger.info("Received TranscriptionFrame")
        if isinstance(frame, UserStoppedSpeakingFrame):
            logger.info("Received UserStoppedSpeakingFrame")
```
2. Then make a simple pipeline with both Deepgram and Ordering Processor
For example:
```
        pipeline = Pipeline(
            [
                transport.input(), 
                DeepgramFluxSTTService(....),
                OrderingProcessor(),
                transport.output(), 
            ]
        )
```
3. Check the logs and you will see that UserStoppedSpeakingFrame reaches first

{"timestamp": 1773314786.697549, "level": "INFO", "message": "Received UserStoppedSpeakingFrame"}
{"timestamp": 1773314786.698449, "level": "INFO", "message": "Received TranscriptionFrame"}

### Expected behavior

Expected behaviour is that TranscriptionFrame will reach the following processors first

### Actual behavior

Actual behaviour is that UserStoppedSpeakingFrame will reach the following processors first

### Workaround

Deepgram has a on_end_of_turn event I can use to send my own EndOfTurnDetectedFrame that isnt a system frame which works. However AssemblyAI does not have the same event sadly
```
await self._call_event_handler("on_end_of_turn", transcript)
```

markbackman:

Also correct the Deepgram Flux event handler docstrings.

Resolves #4008

### DIFF
--- a/changelog/4128.added.md
+++ b/changelog/4128.added.md
@@ -1,0 +1,1 @@
+- Added `on_end_of_turn` event handler to `AssemblyAISTTService`. This fires after the final transcript is pushed, providing a reliable hook for end-of-turn logic that doesn't race with `TranscriptionFrame`. Works in both Pipecat and AssemblyAI turn detection modes.
--- a/src/pipecat/services/assemblyai/stt.py
+++ b/src/pipecat/services/assemblyai/stt.py
@@ -129,6 +129,16 @@ class AssemblyAISTTService(WebsocketSTTService):
     Provides real-time speech transcription using AssemblyAI's WebSocket API.
     Supports both interim and final transcriptions with configurable parameters
     for audio processing and connection management.
+
+    Event handlers available (in addition to WebsocketSTTService events):
+
+    - on_end_of_turn(service, transcript): Called when AssemblyAI detects end of turn.
+
+    Example::
+
+        @service.event_handler("on_end_of_turn")
+        async def on_end_of_turn(service, transcript):
+            ...
     """
 
     Settings = AssemblyAISTTSettings
@@ -302,6 +312,8 @@ class AssemblyAISTTService(WebsocketSTTService):
         self._chunk_size_bytes = 0
 
         self._user_speaking = False
+
+        self._register_event_handler("on_end_of_turn")
 
     def _configure_pipecat_turn_mode(self, settings: Settings, is_u3_pro: bool):
         """Configure settings for Pipecat turn detection mode.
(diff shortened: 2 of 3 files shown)


filipi87 (approved):
LGTM!

PR #4130, opened 2026-03-24
Prefer init-provided system instructions in realtime services

kompfner:
## Summary

- Realtime services (Gemini Live, OpenAI Realtime, Grok Realtime, Nova Sonic) now delegate system instruction conflict resolution to their adapters via `_resolve_system_instruction()`, matching the pattern established for non-realtime services.
- Init-provided `system_instruction` now takes precedence over context-provided system messages (previously it was the opposite). A warning is logged when both are set.
- Fix unnecessary reconnects in Gemini Live when the resolved system instruction hasn't changed from the initial connection.
- Update Grok Realtime example to use `settings.system_instruction` instead of `session_properties.instructions`.

## Breaking Changes

- Realtime services (Gemini Live, OpenAI Realtime, Grok Realtime, Nova Sonic) now prefer `system_instruction` from service settings over an initial system message in the LLM context. Previously, context-provided system instructions took precedence. If you rely on context system messages overriding init-provided `system_instruction`, you will need to adjust your setup.

## Testing

```bash
uv run pytest tests/test_get_llm_invocation_params.py -v
```

### DIFF
--- a/changelog/4130.changed.md
+++ b/changelog/4130.changed.md
@@ -1,0 +1,1 @@
+- ⚠️ Realtime services (Gemini Live, OpenAI Realtime, Grok Realtime, Nova Sonic) now prefer `system_instruction` from service settings over an initial system message in the LLM context, matching the behavior of non-realtime services. Previously, context-provided system instructions took precedence. A warning is now logged when both are set.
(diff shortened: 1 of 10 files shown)


kompfner:
Manually validated all the services touched in this PR, by providing system instructions in:
- [ ] init only
- [ ] context only
- [ ] both places (conflicting)

filipi87 (approved):
LGTM!

Issue #4132, opened 2026-03-25
ElevenLabs TTS: fire-and-forget close_context causes 1008 disconnect during rapid interruptions


OlliTapio:
## Bug Description

`ElevenLabsTTSService._close_context()` sends `close_context: True` over the WebSocket without waiting for server-side acknowledgment. During rapid conversational turns where the user interrupts the bot every few seconds, contexts accumulate server-side faster than they're released, eventually exceeding ElevenLabs' 5-context-per-connection limit and causing a 1008 (policy violation) disconnect.

The bot goes silent after the disconnect until it reconnects.

## Reproduction Test

PR with self-contained repro: #4133

Uses a mock WebSocket server with delayed context cleanup. Two tests:
1. Fire-and-forget close (current behavior) → 1008 after 5+ cycles ✅
2. Wait for `isFinal` ack before opening next context → never exceeds 1 open context ✅

## Steps to Reproduce

1. Run a voice bot with ElevenLabs TTS
2. Have a fast conversation with frequent interruptions (~9 turns in 56 seconds)
3. Each interruption triggers `_handle_interruption` → `on_audio_context_interrupted` → `_close_context` (fire-and-forget)
4. New contexts are created by `run_tts` before server-side cleanup of old contexts completes
5. After accumulating 5+ unclosed contexts server-side → 1008 disconnect

## Root Cause

In `src/pipecat/services/elevenlabs/tts.py`:

```python
async def _close_context(self, context_id: str):
    if context_id and self._websocket:
        await self._websocket.send(
            json.dumps({"context_id": context_id, "close_context": True})
        )
```

This is fire-and-forget — no waiting for ElevenLabs server to acknowledge the close. The pipecat-side `_audio_contexts` dict is cleaned up immediately, but the server-side context remains open until it finishes async cleanup.

## Sentry Breadcrumbs

```
09:49:57.797 [warn] ElevenLabsTTSService#1 connection closed, but with an error: received 1008 (policy violation) Maximum simultaneous contexts per WebSocket connection exceeded (5). Please close an existing context before opening a new one
09:49:57.797 [warn] ElevenLabsTTSService#1 reconnecting, attempt 1
09:49:57.799 [error] ElevenLabsTTSService#1 exception: Unknown error occurred: received 1008 (policy violation) Maximum simultaneous contexts per WebSocket connection exceeded (5)
```

## Environment

- pipecat: 0.0.106 (25.9.23)
[...]

PR #4133, opened 2026-03-25
test: reproduce ElevenLabs 5-context limit bug (#4132)

OlliTapio:
## Summary

Reproduction test for #4132 — ElevenLabs TTS `_close_context()` is fire-and-forget, causing 1008 policy violation during rapid interruptions.

## How it works

Drives the **real `ElevenLabsTTSService`** through a pipecat `Pipeline` with `PipelineTask`, alternating between LLM response frames and `InterruptionFrame`, against a mock WebSocket server that enforces the 5-context limit with realistic cleanup delay.

### Mock server behavior
- Tracks open contexts (max 5)
- `close_context` requests trigger delayed cleanup (1s) — simulating real ElevenLabs async processing
- Sends 1008 (policy violation) when a 6th context is opened

### What the test proves

1. **Server sends 1008** — contexts accumulate past the 5-context limit
2. **`_close_context` was called** — fire-and-forget close requests were sent
3. **TTS works initially** — audio frames produced before the disconnect
4. **Bot goes silent** — some turns produce zero audio after the 1008 (the user-facing bug)

This PR is just the repro test — feel free to close after confirming the bug.

### DIFF
--- a/tests/test_elevenlabs_context_limit_repro.py
+++ b/tests/test_elevenlabs_context_limit_repro.py
@@ -4,25 +4,22 @@
 # SPDX-License-Identifier: BSD 2-Clause License
 #
 
-"""Reproduction test for ElevenLabs TTS 5-context limit bug.
+"""Reproduction test for ElevenLabs TTS 5-context-per-connection limit bug.
 
-The bug: ElevenLabsTTSService._close_context() sends close_context:True over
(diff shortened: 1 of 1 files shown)


markbackman:
This issue was fixed in: https://github.com/pipecat-ai/pipecat/pull/4126

Issue #3758, opened 2026-02-16
AudioBufferProcessor produces laggy/distorted recordings when background audio is present (8kHz Twilio call)


SwAt1563:
### pipecat version

0.0.102

### Python version

3.13.2

### Operating System

python:3.13.2-slim

### Use Case Description

I'm using Pipecat to record conversations from Twilio telephony calls. Twilio audio streams run at **8kHz sample rate**, so I'm configuring `AudioBufferProcessor` accordingly.

My goal is to record clean conversation audio from calls while optionally handling background audio separately.

### Current Approach

```python
I'm creating the audio buffer processor like this:


from pipecat.processors.audio.audio_buffer_processor import AudioBufferProcessor

def get_audiobuffer_processor(
    enable_chunking: bool,
    *,
    sample_rate: int = 8000,
    chunk_duration: int = 15
) -> AudioBufferProcessor:
    if enable_chunking:
        buffer_size = sample_rate * 2 * chunk_duration
    else:
        buffer_size = 0

    return AudioBufferProcessor(
        num_channels=1,
        sample_rate=sample_rate,
        buffer_size=buffer_size
    )







import os
from pipecat.audio.mixers.soundfile_mixer import SoundfileMixer

# NOTE: Make sure the audio files are 8kHz mono WAV files for compatibility
OFFICE_SOUND_FILE = os.path.join(
    ROOT_PATH, "assets", "office-ambience-8000-mono.wav"
)

audio_out_mixer=(
    SoundfileMixer(
        sound_files={
            "office_ambience": OFFICE_SOUND_FILE
        },
        default_sound=background_sound,
        volume=0.2,
    )
    if not is_muted and background_sound
    else None
),


I also tested with a background sound audio file (WAV, 8kHz) played at volume = 0.2 during calls. (enable_chunking=True)
```

### Errors or Unexpected Behavior

When **no background sound is present**, recordings are clean.

When **background sound is present**, the recorded audio quality becomes poor:

* noticeable lag
* interruptions
* speech distortion while the user is talking

This affect only recordings, not the live stream itself.

I initially suspected background sound might be excluded from recording, but it appears to be included and may be affecting buffering or processing timing.

### Additional Context

* This behavior was not noticeable before (unsure which version introduced it).
* Currently tested on latest release (0.0.102).
* Issue also reproduced on 0.0.101.
* Twilio stream sample rate: **8kHz**
* Background audio sample rate: **8kHz WAV**
[...]

PR #4135, opened 2026-03-25
Fixed audio crackling and popping artifacts in AudioBufferProcessor

filipi87:
## Summary
- Fixed audio crackling and popping artifacts in `AudioBufferProcessor` caused by silence being injected into an active audio stream mid-utterance                                                                                 
- Silence injection into the bot buffer is now skipped while the bot is actively speaking, and vice-versa for the user buffer

### DIFF
--- a/changelog/4135.fixed.md
+++ b/changelog/4135.fixed.md
@@ -1,0 +1,1 @@
+- Fixed audio crackling and popping in recordings when both user and bot are speaking. `AudioBufferProcessor` no longer injects silence into a track's buffer while that track is actively producing audio, preventing mid-utterance interruptions in the recorded output.
--- a/src/pipecat/processors/audio/audio_buffer_processor.py
+++ b/src/pipecat/processors/audio/audio_buffer_processor.py
@@ -13,6 +13,8 @@ configurations and event-driven processing.
 
 from typing import Optional
 
+from loguru import logger
+
 from pipecat.audio.utils import create_stream_resampler, interleave_stereo_audio, mix_audio
 from pipecat.frames.frames import (
     BotStartedSpeakingFrame,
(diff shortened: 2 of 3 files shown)


filipi87:
Fixes https://github.com/pipecat-ai/pipecat/issues/3758

examples/foundational/34-audio-recording-plus-soundfile.py
```diff
@@ -0,0 +1,229 @@
+#
```
markbackman: Let's remove this file before merging.

markbackman (commented):
LGTM!

Is there any test coverage that we can add for this? I'm only asking because this has code has been tricky and we've iterated a bunch to get it right. I want to make sure we're not regressing functionality. Not sure how easy that would be though. We could tackle that later.

Also, remove the example and 8khz audio file. I don't think those need to be added to the repo.

I had Claude review locally. It flagged one thing regarding performance:

> Performance bug: value equality on bytearrays instead of identity check
> 
> audio_buffer_processor.py:280:
> user_buffer = self._user_audio_buffer == buffer
> 
> This does a byte-by-byte value comparison (O(n)) of potentially large audio buffers on every
> audio frame. With a mixer producing continuous audio at 16kHz, this could be called hundreds of
> times per second with megabyte-sized buffers. Should be an identity check:
> 
> user_buffer = buffer is self._user_audio_buffer

examples/foundational/34-audio-recording-plus-soundfile.py
filipi87: Done.

filipi87:
@markbackman, I have removed the section of code you flagged, as it was only used for tracing logs and is no longer needed.

I have also added some simple automated tests, at least to validate the silence injection guards:
- test_no_silence_injected_into_bot_buffer_while_bot_speaking
- test_no_silence_injected_into_user_buffer_while_user_speaking
- test_silence_resumes_into_bot_buffer_after_bot_stops_speaking
- test_silence_resumes_into_user_buffer_after_user_stops_speaking

The changes are ready for another round of review.

markbackman (approved):
LGTM! Thanks for the fix 🙌

PR #4137, opened 2026-03-25
Downgrade unrecognized language string log from warning to debug

### LINKED ISSUE #4134: Deepgram Nova-3 multilingual STT: mismatch with pipecat Language enum causing warning
### pipecat version

0.0.106

### Python version

3.13

### Operating System

Docker

### Use Case Description

We are using Deepgram Nova-3 STT models with multilingual detection enabled.

According to Deepgram documentation:
[https://developers.deepgram.com/docs/models-languages-overview](https://developers.deepgram.com/docs/models-languages-overview)

To enable multi-language detection, the `language` parameter should be set to `"multi"`.

### Current Approach

```python
We pass `"multi"` as the language value when initializing the STT service.

However, in pipecat, the `Language` enum does not support `"multi"` as a valid value. This results in a warning during initialization:


WARNING | pipecat.services.stt_service:__init__:135 - Language string 'multi' is not a recognized Language code. It will be passed to the service as-is.


Previously, this worked without warning when using `LiveOptions`, which supported `"multi"` directly.
```

* Warning is raised when using `"multi"`:

  ```
  Language string 'multi' is not a recognized Language code
  ```

### Additional Context

To align with pipecat’s `Language` enum while still supporting Deepgram multilingual detection:

* Allow `language=None` in STT settings (do not default to `"EN"` or any specific language)
* Before sending the request to Deepgram:

  * Convert `None` → `"multi"`
* This avoids enum conflicts while preserving intended behavior

markbackman:
## Summary

- Downgraded the "not a recognized Language code" log message from `warning` to `debug` in both `STTService` and `TTSService`
- Service-specific language strings like Deepgram's `"multi"` are valid pass-through values that work correctly, so the warning was unnecessarily alarming

## Fixes

- Fixes #4134

### DIFF
--- a/changelog/4137.changed.md
+++ b/changelog/4137.changed.md
@@ -1,0 +1,1 @@
+- Unrecognized language strings (e.g. Deepgram's `"multi"`) no longer produce a warning at startup. The log message has been downgraded to debug level since these are valid service-specific values that are passed through correctly.
--- a/src/pipecat/services/stt_service.py
+++ b/src/pipecat/services/stt_service.py
@@ -132,7 +132,7 @@ class STTService(AIService):
             try:
                 self._settings.language = Language(self._settings.language)
             except ValueError:
-                logger.warning(
+                logger.debug(
                     f"Language string '{self._settings.language}' is not a recognized "
                     f"Language code. It will be passed to the service as-is."
                 )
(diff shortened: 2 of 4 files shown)


filipi87 (approved):
LGTM!

PR #4140, opened 2026-03-25
Add Inworld Realtime Service

cshape:
Add Inworld's [Realtime Speech to Speech API](https://inworld.ai/realtime-api) as a Pipecat service.


oleksa-source (approved):

examples/realtime/realtime-inworld.py
filipi87, line 1: We have reorganized all our examples in Pipecat. This should now go inside `examples/realtime`.

It would be great if you could also rebase this branch with the latest Pipecat changes.

examples/foundational/19zb-inworld-realtime.py
```diff
@@ -0,0 +1,193 @@
...
+
+Always be helpful and proactive in offering assistance.""",
+        ),
+    )
+
+    llm.register_function("get_restaurant_recommendation", fetch_restaurant_recommendation)
+
+    # Create context with initial message and tools
+    context = LLMContext(
+        [{"role": "user", "content": "Say hello and introduce yourself!"}],
```
filipi87: Use the role "developer", instead of "user".

examples/foundational/19zb-inworld-realtime.py
```diff
@@ -0,0 +1,193 @@
...
+            transport.output(),
+            assistant_aggregator,
+        ]
+    )
+
+    task = PipelineTask(
+        pipeline,
+        params=PipelineParams(
+            audio_in_sample_rate=24000,
+            audio_out_sample_rate=24000,
```
filipi87: Are these specific sample rates required, or can we remove this and let Pipecat use the default values?

src/pipecat/services/inworld/realtime/llm.py
```diff
@@ -0,0 +1,1064 @@
...
+        self._api_session_ready = True
+        if self._run_llm_when_api_session_ready:
+            self._run_llm_when_api_session_ready = False
+            await self._create_response()
+
+    async def _handle_evt_audio_delta(self, evt):
+        """Handle audio delta event — streaming audio from assistant."""
+        await self.stop_ttfb_metrics()
+
+        if not self._current_audio_response:
```
filipi87, line 776:
```suggestion
        await self.stop_ttfb_metrics()

        if self._current_audio_response and self._current_audio_response.item_id != evt.item_id:
            logger.warning(
                f"Received a new audio delta for an already completed audio response before receiving the BotStoppedSpeakingFrame."
            )
            logger.debug("Forcing previous audio response to None")
            self._current_audio_response = None

        if not self._current_audio_response:
```

We had to include this change for `OpenAIRealtimeLLMService` to fix a race condition that could cause an error when truncating the conversation.

src/pipecat/adapters/services/inworld_realtime_adapter.py
filipi87, line 1: Here we have essentially the same logic as the `open_ai_realtime_adapter.py`. 

However, some features are still missing, such as converting "developer" messages when needed, which we added recently.

I was thinking we could create a new `base_llm_openai_realtime_adapter.py` that both OpenAI and Inworld extend. This would help reduce code duplication.

That said, I can implement this refactor in a follow-up PR if you'd prefer not to introduce larger changes right now. In that case, we would at least align this with the most recent features added to the OpenAI realtime adapter.

filipi87: It looks like we already have this code duplication with XAI (Grok), so you can just match the features here. I’ll push a follow-up PR later with the necessary refactors. 👍

src/pipecat/services/inworld/realtime/llm.py
```diff
@@ -0,0 +1,1064 @@
...
+    async def _handle_user_stopped_speaking(self, frame):
+        """Handle user stopped speaking event.
+
+        When server-side turn detection is disabled, pipecat's local VAD
+        drives commit + response. When enabled, the server handles it.
+        """
+        if not self._is_turn_detection_enabled():
+            await self.send_client_event(events.InputAudioBufferCommitEvent())
+            await self.start_processing_metrics()
+            await self.start_ttfb_metrics()
```
filipi87: Are these also needed here? It looks like we’re already invoking them inside `_handle_evt_speech_stopped`. Is it needed in both places ?

filipi87:
Hi @cshape, when trying to run the example `examples/foundational/19zb-inworld-realtime.py`, I’m encountering the following error:

```
2026-04-08 10:22:58.909 | DEBUG    | pipecat.services.inworld.realtime.llm:_send_user_audio:1017 - Streaming audio to Inworld: 24000Hz, 1ch, 960B/frame
2026-04-08 10:22:59.123 | ERROR    | pipecat.processors.frame_processor:push_error_frame:744 - InworldRealtimeLLMService#0 error: Inworld Realtime Error: Failed to read content stream. INTERNAL: HTTP error: 400
2026-04-08 10:22:59.124 | WARNING  | pipecat.pipeline.task:_source_push_frame:903 - PipelineTask#0: Something went wrong: ErrorFrame#0(error: Inworld Realtime Error: Failed to read content stream. INTERNAL: HTTP error: 400, fatal: False)
```

Could you double check whether everything is working as expected ?

filipi87 (commented):
Thanks for pushing this PR, @cshape.

Overall, it looks good. I’ve left a few comments that we should address.

I noticed that we’re reusing a lot of code from `OpenAIRealtime`. But the same pattern exists with XAI (Grok), so once this lands, I’ll work on a broader refactor to introduce a shared base class that all three can extend. 

For now, it’s probably best to handle that in a follow-up so we don’t block this PR.

Let me know once you’ve addressed the comments so I can review it again 👍.

src/pipecat/services/inworld/realtime/llm.py
cshape: These were for a code path where we turn off Inworld STT/VAD and use Pipecat's local VAD, but this actually isn't supported well in Inworld Realtime now.

So I've removed and simplified so there's just one code path here for Inworld Realtime which is using the server-side VAD in our audio-to-audio pipeline.

examples/foundational/19zb-inworld-realtime.py
cshape: Not required - removed and let the service handle it.

src/pipecat/adapters/services/inworld_realtime_adapter.py
```diff
@@ -0,0 +1,263 @@
...
+            List of tool definitions in Inworld Realtime format.
+        """
+        # Convert standard function tools
+        functions_schema = tools_schema.standard_tools
+        standard_tools = [self._to_inworld_function_format(func) for func in functions_schema]
+
+        # Support shimmed custom tools for backward compatibility
+        shimmed_tools = []
+        if tools_schema.custom_tools:
+            shimmed_tools = tools_schema.custom_tools.get(AdapterType.SHIM, [])
```
filipi87: Is this code needed? If so, you would need to create an `AdapterType.SHIM`, currently it doesn't exist. 

You should also consider renaming it to something more specific to the provider, like `INWORLD_SHIM` or `OPENAI_SHIM`, to make it easier to understand what it is.

cshape: Good catch! Yes we don't need this (unlike Gemini) as the tools are OpenAI-compatible already.

filipi87 (approved):
LGTM! Thank you for putting this together, @cshape 🚀

PR #4141, opened 2026-03-25
feat: add WebSocket-based OpenAI Responses LLM service

### LINKED ISSUE #3935: Support OpenAI Responses WebSocket mode
### Service Name

OpenAI WebSocket Mode

### Service Website

https://developers.openai.com/api/docs/guides/websocket-mode/

### Service Description

Is it possible to integrate with websocket mode released by OpenAI to reduce TTFT latency?

### API Information

_No response_

### Would you be willing to help implement this service?

- [ ] No, I'm just suggesting

kompfner:
## Summary

- Added a WebSocket variant of the OpenAI Responses API service (`OpenAIResponsesLLMService`) that maintains a persistent connection to `wss://api.openai.com/v1/responses` for lower-latency inference
- Implements `previous_response_id` optimization to send only incremental context when the conversation prefix hasn't changed, with automatic fallback to full context on reconnection or cache miss
- Skips resending the server's own response output (assistant messages, function calls) by loosely matching it against the items following the matched prefix
- Handles interruptions by sending `response.cancel` and draining stale events before the next inference — the API doesn't provide a reliable way to correlate events to a specific response, so we drain rather than filter
- Renames the existing HTTP variant to `OpenAIResponsesHttpLLMService`; both share a private base class with common settings, parameter building, and `run_inference` logic
- Documents why the HTTP variant doesn't use `previous_response_id` (requires `store=True` / 30-day OpenAI-side storage)
- Adds HTTP-specific example variants and corresponding eval entries

## Breaking Changes

- `OpenAIResponsesLLMService` now refers to the WebSocket variant. Users who need the HTTP streaming behavior should switch to `OpenAIResponsesHttpLLMService`. The public API (constructor args, settings, `run_inference`) is identical between both variants.

## Testing

- `uv run pytest tests/test_run_inference.py` — 32 tests (existing + 6 new HTTP variant)
- `uv run pytest tests/test_openai_responses_websocket.py` — 34 new WebSocket-specific tests
- `uv run pytest tests/test_get_llm_invocation_params.py` — 115 adapter tests unaffected
- Manual: run `07-interruptible-openai-responses.py` to verify WebSocket e2e
- Manual: run `07-interruptible-openai-responses-http.py` to verify HTTP e2e

Resolves #3935


kompfner:
Resolves #3935

src/pipecat/services/openai/responses/llm.py
```diff
@@ -173,6 +211,682 @@ def can_generate_metrics(self) -> bool:
...
+# ---------------------------------------------------------------------------
+
+
+class OpenAIResponsesLLMService(_BaseOpenAIResponsesLLMService):
+    """OpenAI Responses API LLM service using WebSocket transport.
+
+    Maintains a persistent WebSocket connection to ``wss://api.openai.com/v1/responses``
+    for lower-latency inference, especially beneficial for tool-call-heavy workflows.
+    Automatically uses ``previous_response_id`` to send only incremental context when
+    possible, and falls back to full context on reconnection or cache miss.
```
kompfner, line 347: Note that we will often have cache misses when the user interrupts the bot mid-response, since the context after an interruption won't quite match the API-side cached context (which includes the full untruncated bot response)

src/pipecat/services/openai/responses/llm.py
```diff
@@ -173,6 +211,682 @@ def can_generate_metrics(self) -> bool:
...
+
+    The ``previous_response_id`` optimization works with ``store=False`` (the default)
```
kompfner, line 349: Reviewers: see this comment for an explanation of why I decided to make the `previous_response_id` optimization default here but not in the HTTP variant of the service

src/pipecat/services/openai/responses/llm.py
kompfner: But keep in mind that the `previous_response_id` optimization is just that—an optimization. Even if we have a lot of cache misses and have to send full context frequently, it's really OK. The fallback performance is no worse than that of the HTTP variant, or than any other of our LLM services—still good!

changelog/4141.added.md
```diff
@@ -0,0 +1 @@
+- ⚠️ Added WebSocket-based `OpenAIResponsesLLMService` as the new default for the OpenAI Responses API. It maintains a persistent connection to `wss://api.openai.com/v1/responses` and automatically uses `previous_response_id` to send only incremental context, falling back to full context on reconnection or cache miss. The previous HTTP-based implementation is now available as `OpenAIResponsesHttpLLMService`.
```
kompfner, line 1: The reason to make this default is that WebSocket is recommended for real-time, conversational uses (which is the vast majority of Pipecat usage). The downside is that it's necessarily quite a bit trickier/more stateful/more complex so it's likely to be slightly buggier than HTTP.

kompfner (commented):
## Code review

Found 3 issues. Checked for bugs and CLAUDE.md compliance.

kompfner:
## Code review

Found 3 issues:

### 1. `_disconnect` can strand `_disconnecting = True` on error

https://github.com/pipecat-ai/pipecat/blob/8756fb8678af5f258a4361febd0fe0f0ac4cb015/src/pipecat/services/openai/responses/llm.py#L444-L456

If `self._websocket.close()` (or anything else in the `try` block) raises, `_disconnecting` stays `True` and the `except` handler doesn't reset it. After this, every `_ws_send` call silently no-ops. Currently mitigated by accident (`_connect` resets it, and `_reconnect` calls `_connect` after `_disconnect`), but fragile. Fix: move `self._disconnecting = False` to a `finally` block.

### 2. `_drain_cancelled_response` doesn't handle `ConnectionClosed`

https://github.com/pipecat-ai/pipecat/blob/8756fb8678af5f258a4361febd0fe0f0ac4cb015/src/pipecat/services/openai/responses/llm.py#L709-L712

Only `asyncio.TimeoutError` is caught. If `recv()` raises `ConnectionClosed` during drain, it propagates uncaught, `_needs_drain` stays `True`, and every subsequent inference attempt fails the same way — the service is bricked. Fix: catch `ConnectionClosed` alongside `TimeoutError` and reconnect.

### 3. WebSocket sends `"Bearer None"` when `api_key` is not passed

https://github.com/pipecat-ai/pipecat/blob/8756fb8678af5f258a4361febd0fe0f0ac4cb015/src/pipecat/services/openai/responses/llm.py#L163-L167

`self._api_key = api_key` stores `None` when the user doesn't pass an API key (relying on the env var convention). The `AsyncOpenAI` HTTP client handles this fallback internally, but the WebSocket `_connect` uses `self._api_key` directly in the `Authorization` header, producing `"Bearer None"`. Fix: resolve from the environment, e.g. `self._api_key = api_key or os.environ.get("OPENAI_API_KEY")`.

src/pipecat/services/openai/responses/llm.py
filipi87: Makes sense, otherwise, we would always send the full context, so any optimization when possible is a gain. 👍

src/pipecat/services/openai/responses/llm.py
```diff
@@ -173,6 +212,786 @@ def can_generate_metrics(self) -> bool:
...
+            )
+        return fc_list
+
+
+# ---------------------------------------------------------------------------
+# WebSocket variant (default / recommended)
+# ---------------------------------------------------------------------------
+
+
+class OpenAIResponsesLLMService(_BaseOpenAIResponsesLLMService):
```
filipi87, line 341: Wouldn’t it make sense to use the `WebsocketService` here instead of reimplementing all the logic?

Maybe we should even have a `WebsocketLLMService`, like we do for STT and TTS. What do you think ?

src/pipecat/services/openai/responses/llm.py
```diff
@@ -173,6 +212,786 @@ def can_generate_metrics(self) -> bool:
...
+            Hex digest of the SHA-256 hash.
+        """
+        return hashlib.sha256(json.dumps(items, sort_keys=True).encode()).hexdigest()
+
+    def _apply_previous_response_optimization(self, params: dict, full_input: list) -> dict:
+        """Try to use previous_response_id to send only new input items.
+
+        If the prefix of ``full_input`` matches the stored hash from the
+        previous inference call, only new items are sent along with
+        ``previous_response_id``. Otherwise the full input is sent.
```
filipi87, line 508: 💯

src/pipecat/services/openai/responses/llm.py
```diff
@@ -173,6 +212,786 @@ def can_generate_metrics(self) -> bool:
...
+                    except Exception:
+                        pass
+                else:
+                    logger.debug(
+                        f"{self}: Cancelled before response.created "
+                        f"— will cancel on next response.created"
+                    )
+                    self._cancel_pending_response = True
+                self._current_response_id = None
+                self._needs_drain = True
```
filipi87, line 765: Shouldn’t we drain it right away here? Maybe in a different task so we don’t delay the interruption?

src/pipecat/services/openai/responses/llm.py
```diff
@@ -173,6 +212,786 @@ def can_generate_metrics(self) -> bool:
...
+        Raises:
+            _PreviousResponseNotFoundError: Server couldn't find previous response.
+            _ConnectionLimitReachedError: 60-minute connection limit reached.
+            ConnectionClosed: WebSocket connection was closed unexpectedly.
+        """
+        function_calls: Dict[str, Dict[str, str]] = {}
+        current_arguments: Dict[str, str] = {}
+
+        while True:
+            raw = await self._websocket.recv()
```
filipi87, line 870: I see that we only try to call `recv` when processing the context, but is there any chance we could receive messages from OpenAI that we should handle even when we’re not processing a response ?

src/pipecat/services/openai/responses/llm.py
kompfner: Yes it probably would make sense!

src/pipecat/services/openai/responses/llm.py
kompfner: > is there any chance we could receive messages from OpenAI that we should handle even when we’re not processing a response

I don't believe so. Despite being a WebSocket connection, the usage pattern is very much "transactional" - you send a request and it gives a response, just like an HTTP API. There's nothing going on outside the transaction.

kompfner: I'll do a bit of research and double-check this assumption, though

src/pipecat/services/openai/responses/llm.py
kompfner: I could also take this as a follow-up/cleanup task to this PR to prevent it from getting too big. Re: `WebsocketLLMService`—there's currently there's no other WebSocket-based LLM services, so making it "shareable" would be strictly a gift to our future selves.

src/pipecat/services/openai/responses/llm.py
kompfner: I think doing it this way—just setting the `_needs_drain` flag, to be processed on the next `_process_context`—simplifies our code quite a bit, eliminating the need for another task (and another queue to ensure the next-processed-context is processed only after the drain is done).

The websocket implementation will keep receiving messages and putting them in an internal buffer, so when we next need to `_drain_cancelled_response` at the start of a `_process_context`, it should be negligibly fast—it's just popping messages off of a queue in memory, with no network involvement.

src/pipecat/services/openai/responses/llm.py
kompfner: Yeah as far as I can tell that's the case.

filipi87: Cool. So handling it like this should be good enough. 👍

filipi87 (approved):
LGTM! Nice improvement. 🚀

Let’s plan to use the WebSocket service, but do this in a follow-up PR, as you suggested, so we don’t block this one. 👍

PR #4143, opened 2026-03-25
Add Deepgram Flux STT service for AWS SageMaker

chadbailey59:
I (well, Claude) basically combined the existing DeepgramSageMakerSTTService and DeepgramFluxService to make a DeepgramFluxSageMakerSTTService. I ran the included new example, and everything worked as it should. There weren't any existing automated tests for Deepgram Flux, and it's additionally impractical to run tests on these SageMaker services because of setup and teardown.

## Summary

- Adds `DeepgramFluxSageMakerSTTService` that combines SageMaker's HTTP/2 bidirectional transport (`SageMakerBidiClient`) with Deepgram Flux's JSON turn detection protocol
- Supports advanced turn events: StartOfTurn, EndOfTurn, EagerEndOfTurn, TurnResumed
- Includes mid-stream `Configure` support for updatable fields (keyterm, eot_threshold, eager_eot_threshold, eot_timeout_ms)
- Adds silence watchdog to prevent dangling turns when audio stops flowing
- Adds example bot (`07c-interruptible-deepgram-flux-sagemaker.py`) using Flux STT + SageMaker TTS + AWS Bedrock LLM with `ExternalUserTurnStrategies`

## Test plan

- [x] Verified import resolves: `from pipecat.services.deepgram.sagemaker.flux_stt import DeepgramFluxSageMakerSTTService`
- [x] Passed `ruff check` and `ruff format --check`
- [x] Existing tests pass (`uv run pytest tests/ -x`)
- [x] Tested end-to-end with SageMaker Flux endpoint via Daily transport


src/pipecat/services/deepgram/flux/base.py
filipi87, line 1: Would it make sense to instead add it inside the `deepgram/flux/sagemaker` module and keep it there as `stt.py` ?

I don’t have a strong opinion, though. 

Keeping it here has the advantage that everything related to `SageMaker` is in one place, while keeping it inside `deepgram/flux/sagemaker` has the advantage that everything related to `Flux` is in one place. 🤷‍♂️

cc: @markbackman for thoughts.

src/pipecat/services/deepgram/sagemaker/flux_stt.py
```diff
@@ -0,0 +1,620 @@
...
+    EndFrame,
+    ErrorFrame,
+    Frame,
+    InterimTranscriptionFrame,
+    StartFrame,
+    TranscriptionFrame,
+    UserStartedSpeakingFrame,
+    UserStoppedSpeakingFrame,
+)
+from pipecat.processors.frame_processor import FrameDirection
```
filipi87: nit: Unused import.

src/pipecat/services/deepgram/flux/base.py
markbackman: I had a similar thought about this. I agree that we should probably follow the same pattern so this is an stt.py file.

filipi87 (approved):
@chadbailey59 this works great.

As a suggestion, I have a follow-up [PR](https://github.com/pipecat-ai/pipecat/pull/4158) where:
- We create a base class, `DeepgramFluxSTTBase`, so we can reuse Deepgram Flux logic across both services.
- We move `flux_stt.py` to `deepgram/flux/sagemaker/stt.py`.

markbackman:
@filipi87 is this ready to be merged now?

filipi87:
> @filipi87 is this ready to be merged now?

Hi @markbackman, I think so. 👍

PR #4145, opened 2026-03-25
TTS improvements.

filipi87:
## Summary                                                                              
- Removed `"Reset"` and `"TTSStoppedFrame"` string sentinels from `add_word_timestamps`                                                                          
- Services that previously called `add_word_timestamps([("TTSStoppedFrame", 0), ("Reset", 0)], ctx_id)` now call `append_to_audio_context(ctx_id, TTSStoppedFrame(...))` directly, and reset logic is handled uniformly at the end of `_handle_audio_context`
- `append_to_audio_context` signature broadened to accept `Frame | _WordTimestampEntry | None`, and `remove_audio_context` now routes through it for consistency                                                                                                          
- All `on_audio_context_interrupted` and `on_audio_context_completed` overrides across TTS services now call `super()`, ensuring base class lifecycle hooks are consistently invoked                                                                                      
- Fixed a race condition in `InterruptibleTTSService` where, if `run_tts` had been invoked but `BotStartedSpeakingFrame` had not yet been received, a user interruption could allow stale audio to leak through.
- Fixed websocket TTS word timestamps so interrupted contexts cannot leak stale words or backward PTS values into later turns.


markbackman:
## Code review

markbackman (approved):
Change looks good.

I tested the services modified and all sound good.

I noticed issues with word timestamp / audio alignment for Inworld and Rime. I think we can follow up on those separately as they may be service related.

Nice cleanup! 👏

filipi87:
Hi @markbackman, yes, I had noticed that for `Inworld` as well. I just pushed a PR with some fixes:
- https://github.com/pipecat-ai/pipecat/pull/4167

I will take a look on what is happening for Rime.

Thank you for testing this. 🙌

PR #4146, opened 2026-03-25
Fix Gemini Live local VAD by sending correct activity events to server

markbackman:
## Summary

- Fixed Gemini Live local VAD mode (`GeminiVADParams(disabled=True)`) not working. The service now correctly detects user speech via VAD frames and sends `ActivityStart`/`ActivityEnd` signals to the Gemini API to indicate turn boundaries.
- Updated Gemini Live examples to default to server-side VAD (no local Silero VAD). Added a new dedicated example (`26a-gemini-live-local-vad.py`) demonstrating local VAD configuration.

## Testing

- Run the local VAD example: `python examples/foundational/26a-gemini-live-local-vad.py`
- Confirm the bot responds after the user stops speaking and that interruptions work.

### DIFF
--- a/src/pipecat/services/google/gemini_live/llm.py
+++ b/src/pipecat/services/google/gemini_live/llm.py
@@ -54,8 +54,8 @@ from pipecat.frames.frames import (
     TTSStoppedFrame,
     TTSTextFrame,
     UserImageRawFrame,
-    UserStartedSpeakingFrame,
-    UserStoppedSpeakingFrame,
+    VADUserStartedSpeakingFrame,
+    VADUserStoppedSpeakingFrame,
 )
 from pipecat.metrics.metrics import LLMTokenUsage
 from pipecat.processors.aggregators.llm_context import LLMContext
@@ -88,6 +88,8 @@ try:
     from google.genai import Client
     from google.genai.live import AsyncSession
     from google.genai.types import (
+        ActivityEnd,
+        ActivityStart,
         AudioTranscriptionConfig,
         AutomaticActivityDetection,
         Blob,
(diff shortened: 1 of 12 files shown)


src/pipecat/services/google/gemini_live/llm.py
```diff
@@ -522,7 +524,7 @@ class GeminiVADParams(BaseModel):
     """Voice Activity Detection parameters for Gemini Live.
 
     Parameters:
-        disabled: Whether to disable VAD. Defaults to None.
+        disabled: Whether to disable VAD. Defaults to None (server-side VAD is enabled).
```
kompfner, line 527: Was server-side VAD enabled always the default and you're just calling it out explicitly here? Or is this a change?

kompfner: Follow-up question: if so, prior to your changes to the examples in this PR, the examples attempted to have _both_ local and server-side VAD going?

kompfner: And follow-up to that: haven't we long maintained that local VAD is faster/more reliable than server-side, and that we recommend treating server-side signals as "supplementary"? (I remember that's what we recommended for AWS Nova Sonic at least)

markbackman: I'm not sure if it has always been, but it's definitely now the default. In looking at the Gemini docs, it was unclear what the behavior is, but in testing it, it's very clear that the default is to use the server-side VAD.

markbackman: > the examples attempted to have both local and server-side VAD going?

Correct. The local VAD was running but doing nothing, AFAICT.

markbackman: > haven't we long maintained that local VAD is faster/more reliable than server-side, and that we recommend treating server-side signals as "supplementary"?

Yes, that is the case for most things. Though, for Gemini Live, the server-side VAD yields useful user transcripts whereas using the local VAD yields garbage for the user transcripts. I'm not sure why this is; perhaps their STT model requires a specific amount of silence padding. We'll have to ask the Google team.

kompfner (approved):

filipi87 (approved):
LGTM!

PR #4147, opened 2026-03-25
Fix Gemini Live to handle bundled server_content fields

chadbailey59:
## Summary
- Gemini 3.x can bundle multiple fields (e.g. `model_turn` and `output_transcription`) on the same `server_content` message
- The previous `elif` chain only processed the first matching field, silently dropping the rest
- Switched to independent `if` checks so every field on a message is handled

## Test plan
- [ ] Test with Gemini Live using transcription enabled to verify both model audio and output transcription are processed from bundled messages
- [ ] Verify interruption handling still works correctly
- [ ] Confirm tool calls and session resumption updates are unaffected


src/pipecat/services/google/gemini_live/llm.py
```diff
@@ -1360,7 +1360,12 @@ async def _connection_task_handler(self, config: LiveConnectConfig):
                         # Reset failure counter if connection has been stable
                         self._check_and_reset_failure_counter()
 
-                        if message.server_content and message.server_content.interrupted:
+                        # server_content fields are NOT mutually exclusive —
```
kompfner, line 1363: Haven't dug too deep into this yet, but if this is broadly the case I'd be a bit surprised if the fix were as simple as just changing `elif` to `if` (in other words, I'd expect that there'd be some subtle implicit assumptions in our message handling logic about messages being mutually exclusive).

Do we know which `server_content` fields are allowed together? If it's only `model_turn` and `output_transcription` maybe we can be a bit more targeted in our change.

kompfner: > I'd expect that there'd be some subtle implicit assumptions in our message handling logic about messages being mutually exclusive)

Indeed: https://github.com/pipecat-ai/pipecat/pull/4148

tests/test_google_gemini_live_llm.py
kompfner, line 1: these tests feel a little too "reach-into-internals-y". pretty much any internal refactoring or name changes will break these.

kwindla: I agree. I told @chadbailey59 that I thought we should just cherry-pick the actual fix and not include the test. (I just let Codex write the test it wanted to write, and then committed that change. But I shouldn't have committed the test.) But I think I told Chad that in a DM so it got lost in the shuffle. Apologies.

kompfner:
Waiting to test/validate this once we can (once our usage of `history_config` is unblocked)

kompfner (approved):
> Waiting to test/validate this once we can (once our usage of history_config is unblocked)

Gemini 3.1 Flash Live is working for basic conversation-and-tool-call usage (26, 26b, 26e, 26g). We'll continue to iron out the kinks in the others.

PR #4151, opened 2026-03-26
fix(livekit): clear AudioSource buffer on interruption

ajmeraharsh:
## Summary

- When an `InterruptionFrame` arrives, Pipecat cancels its Python-side audio task but frames already submitted to LiveKit's `rtc.AudioSource` via `capture_frame()` remain in its internal buffer and continue playing — causing the bot to keep speaking for several seconds (up to minutes if the TTS had streamed far ahead) after being interrupted
- Fix overrides `process_frame` in `LiveKitOutputTransport` to call `audio_source.clear_queue()` on `InterruptionFrame`, immediately flushing the buffered audio

## Test plan

- [ ] Start a LiveKit voice session with a verbose bot response
- [ ] Interrupt the bot mid-speech — bot should stop speaking immediately
- [ ] Confirm normal conversation flow continues after interruption

### DIFF
--- a/changelog/4151.fixed.md
+++ b/changelog/4151.fixed.md
@@ -1,0 +1,1 @@
+- Fixed `LiveKitOutputTransport` not clearing the `rtc.AudioSource` internal buffer on interruption, causing the bot to continue speaking for several seconds after being interrupted.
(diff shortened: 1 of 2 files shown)


markbackman (approved):
LGTM! Thank you 🙌

PR #4152, opened 2026-03-26
Fix audio transcript check in base_llm.py

vpalmisano:
In some cases the openai provider could answer with a `chunk.choices[0].delta.audio = None`, so the process context fails with error:
```
pipecat/services/openai/base_llm.py:552): Error during completion: 'NoneType' object has no attribute 'get'
```


filipi87 (approved):
Thank you for fixing this @vpalmisano. I’ve fixed the format and included it in the changelog so we can include it in our next release.

PR #4153, opened 2026-03-26
Handle Deepgram SDK 6.x send_media() exceptions

markbackman:
## Summary

- Fixed error floods in `DeepgramSTTService` when the WebSocket connection drops mid-stream. Deepgram SDK 6.x intentionally surfaces connection errors from `send_media()` rather than silently swallowing them (as SDK 4.x did), causing every queued audio frame (~20ms interval) to hit the dead connection and log an exception.
- Now `send_media()` failures are caught gracefully — a single warning is logged and `self._connection` is set to `None` so subsequent audio frames are skipped until the existing `_connection_handler` reconnects.

## Testing

- `ruff check` and `ruff format --check` pass
- All 1223 tests pass (`uv run pytest tests/ -x`)


src/pipecat/services/deepgram/stt.py
```diff
@@ -577,7 +577,11 @@ async def run_stt(self, audio: bytes) -> AsyncGenerator[Frame, None]:
             Frame: None (transcription results come via WebSocket callbacks).
         """
         if self._connection:
-            await self._connection.send_media(audio)
+            try:
+                await self._connection.send_media(audio)
+            except Exception as e:
+                logger.warning(f"{self}: send_media failed, connection will reconnect: {e}")
```
markbackman, line 583: I think a warning is the right call here. If it's too noisy, we can make it debug.

filipi87: Agreed. 👍

src/pipecat/services/deepgram/stt.py
```diff
@@ -577,7 +577,11 @@ async def run_stt(self, audio: bytes) -> AsyncGenerator[Frame, None]:
...
+                self._connection = None
```
filipi87, line 584: @markbackman, Deepgram extends `STTService`, not `WebsocketSTTService`.

So, would setting it to `None` here be enough to reestablish the connection ?

markbackman: Setting `self._connection = None` doesn't trigger reconnection. It stops the flood of error messages that would happen.

The reconnection is driven by `_connection_handler()`, which is an internal loop:
```
    async def _connection_handler(self):
        """Manages the full WebSocket lifecycle inside a single async with block.

        Reconnects automatically after transient errors. Exits cleanly when
        the task is cancelled (i.e. on stop/cancel).
        """
        while True:
            connect_kwargs = self._build_connect_kwargs()
            try:
                async with self._client.listen.v1.connect(**connect_kwargs) as connection:
                    self._connection = connection
                    connection.on(EventType.MESSAGE, self._on_message)
                    connection.on(EventType.ERROR, self._on_error)

                    logger.debug(f"{self}: Websocket connection initialized")

                    keepalive_task = self.create_task(
                        self._keepalive_handler(), f"{self}::keepalive"
                    )
                    try:
                        await connection.start_listening()
                    finally:
                        await self.cancel_task(keepalive_task)
                raise
            except Exception as e:
                logger.warning(f"{self}: Connection lost, will retry: {e}")
            finally:
                self._connection = None
```

When the WebSocket dies, `start_listening()` exits/raises, the `except Exception` block catches it, and the `while True` loop iterates to open a fresh connection.

I think this should do the trick. We could try to force a disconnection, but if the websocket is broken already, that will likely return an error as well.

WDYT?

I'll add a test for this.

filipi87: True. We are already handling it there. Thank you for the explanation.

filipi87 (approved):
LGTM!

PR #4156, opened 2026-03-26
fix(mem0): improve Mem0 service reliability and add get_memories() method

### LINKED ISSUE #1741: Mem0 `self._store_messages` blocks the conversation flow
### pipecat version

0.0.65

### Service Name

Mem0

### Service or model version

_No response_

### Issue Description

When using the Mem0 memory service the `self.store_messages(context_messages)` (line 194) seems to be blocking the conversation. This should be refactored to to be called async (or made non-blocking some other way).

Also, the `position` input param seems to be unused. 

### Reproduction Steps

Add `Mem0MemoryService`. Observe that there is a noticeable long pause after logs for storing messages in Mem0 and the LLM service generating chat.

### Expected Behavior

The storing of messages should not be blocking, only the retrieving and enhancing of context.

### Actual Behavior

Storing the messages seems to be blocking, seems like it's waiting for responses from all the API calls.

### Error Logs

```shell

```

markbackman:
## Summary

- Fixed `Mem0MemoryService` blocking the event loop — all Mem0 API calls (`add`, `search`, `get_all`) now run in background threads via `asyncio.to_thread()`, and message storage is fire-and-forget via `create_task()`
- Fixed memory storage failing with 400 errors when context contained system/developer role messages — these are now filtered to only user/assistant roles before calling the Mem0 API
- Added `Mem0MemoryService.get_memories()` convenience method for retrieving all stored memories outside the pipeline (e.g. for personalized greetings at connection time)
- Memory messages are now inserted at the configured `position` in the context instead of always being appended
- Bumped `mem0ai` dependency from `~=0.1.94` to `>=1.0.8,<2`
- Updated the `37-mem0.py` example to use `get_memories()` and simplified the greeting logic

## Breaking Changes

- `mem0ai` dependency updated from `~=0.1.94` to `>=1.0.8,<2`. Users of the `mem0` extra will need to update their mem0ai package.

## Testing

- `uv run ruff check src/pipecat/services/mem0/memory.py examples/foundational/37-mem0.py`
- Manual: run `examples/foundational/37-mem0.py` and verify personalized greeting includes memory content

Fixes #1741


kompfner (approved):

filipi87 (approved):
Looks good. 👍

PR #4167, opened 2026-03-27
InworldTTSService improvements.

filipi87:
InworldTTSService improvements:
- Fixed a word timestamp interleaving issue when processing multiple sentences.
- Disabled the property that pushes stop frames, since the service already does this when closing the context.
- Avoided closing the context inside `on_audio_context_completed`, as it is already closed in `on_turn_context_completed`.
- Fixed an issue where, in cases of fast interruption, we would continue receiving audio from the previous context.

### DIFF
--- a/src/pipecat/services/inworld/tts.py
+++ b/src/pipecat/services/inworld/tts.py
@@ -646,7 +646,7 @@ class InworldTTSService(WebsocketTTSService):
 
         super().__init__(
             push_text_frames=False,
-            push_stop_frames=True,
+            push_stop_frames=False,
             pause_frame_processing=True,
             sample_rate=sample_rate,
             aggregate_sentences=aggregate_sentences,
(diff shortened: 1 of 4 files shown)


src/pipecat/services/inworld/tts.py
```diff
@@ -837,10 +838,6 @@ async def on_audio_context_interrupted(self, context_id: str):
         await self._maybe_push_fallback_text(context_id)
         await self._close_context(context_id)
 
-    async def on_audio_context_completed(self, context_id: str):
-        """Callback invoked when an audio context has been completed."""
-        await self._close_context(context_id)
```
filipi87, line 839: Not needed, we are already closing it inside `on_turn_context_completed`.

src/pipecat/services/inworld/tts.py
```diff
@@ -819,8 +822,6 @@ async def _close_context(self, context_id: str):
             except Exception as e:
                 await self.push_error(error_msg=f"Unknown error occurred: {e}", exception=e)
         self._sent_context_ids.discard(context_id)
-        self._cumulative_time = 0.0
-        self._generation_end_time = 0.0
```
filipi87, line 819: This was causing the word timestamp interleaving issue because we were resetting it while still receiving audio from Inworld.

src/pipecat/services/inworld/tts.py
```diff
@@ -648,7 +648,7 @@ def __init__(
 
         super().__init__(
             push_text_frames=False,
-            push_stop_frames=True,
+            push_stop_frames=False,
```
filipi87, line 649: Not needed, as Inworld already pushes the stop frame when the context is closed.

src/pipecat/services/inworld/tts.py
```diff
@@ -1015,10 +997,6 @@ async def _receive_messages(self):
             # Handle context created confirmation
             if "contextCreated" in result:
                 logger.trace(f"{self}: Context created on server: {ctx_id}")
-            # If the context isn't available recreate it (handles race conditions during interruption recovery).
```
filipi87, line 1018: We should not try to recreate the context. If it doesn’t exist, it was due to an interruption, and any received audio should be ignored.

Ideally, we would have a message to send to Inworld to cancel any audio from that specific context when an interruption occurs.

markbackman (approved):
LGTM! Works much better.

The remaining thing to do is make:
```
logger.warning(f"{self} unable to append audio to context {context_id}")
```

In `TTSService` a debug log.

filipi87:
@markbackman, done, just changed the log to debug. 👍

PR #4171, opened 2026-03-27
Gate Gemini Live sending real-time input messages to the API until it…

kompfner:
…'s ready, i.e. after we've sent the initial conversation history (or determined that we don't need to).

This fixes the 26c example when using Gemini 3.1 Flash Live, which seems to be more strict about not receiving real-time input (at least, video messages) before conversation history.

### DIFF
--- a/src/pipecat/services/google/gemini_live/llm.py
+++ b/src/pipecat/services/google/gemini_live/llm.py
@@ -803,6 +803,7 @@ class GeminiLiveLLMService(LLMService):
 
         self._audio_input_paused = start_audio_paused
         self._video_input_paused = start_video_paused
+        self._ready_for_realtime_input = False
         self._context = None
         self._api_key = api_key
         self._http_options = update_google_client_http_options(http_options)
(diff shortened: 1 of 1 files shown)


src/pipecat/services/google/gemini_live/llm.py
```diff
@@ -996,7 +997,7 @@ async def _handle_interruption(self):
 
     async def _handle_user_started_speaking(self, frame):
         self._user_is_speaking = True
-        if self._vad_disabled and self._session:
+        if self._vad_disabled and self._session and self._ready_for_realtime_input:
```
kompfner, line 1000: Validated that in the quite unlikely case that we weren't yet ready for realtime input when the user started speaking but we were ready when the user stopped speaking, we would recover when the user repeated themselves.

markbackman (approved):
This seems like a safe change.

Generally, it seems like it's harder to interact with the Gemini Live API after this change. More rules about when input is accepted. Do they have events or signaling that we could hook into that would help out here?

kompfner:
> Generally, it seems like it's harder to interact with the Gemini Live API after this change.

I think their checking has just gotten a wee bit stricter. This was always a bug/misuse on our part; their docs are clear on this: send convo history, then start real-time input.

kompfner:
> Do they have events or signaling that we could hook into that would help out here?

Wouldn't really help in this case, as it's not a question of API-driven timing (we don't need the API to be "ready"). It's a question of client-side ordering of messages.

PR #4172, opened 2026-03-27
Fix Rime TTS stop-frame handling and handle done message

markbackman:
## Summary

- **Rime TTS: handle `done` message for faster utterance completion.** `RimeTTSService` now handles Rime's `done` WebSocket message to immediately complete audio contexts, eliminating the 3-second idle timeout that previously added latency at the end of each utterance. This mirrors how Cartesia handles its completion signal.
- **Fix duplicate `TTSStoppedFrame` in TTS service timeout path.** When `push_stop_frames=True` was set and the stop-frame timeout fired, a second `TTSStoppedFrame` could be pushed after the normal one at context completion. Fixed by setting `should_push_stop_frame = False` after pushing in the timeout branch.

## Testing

- `uv run pytest` — full test suite passes (1227 passed)
- Manual test with Rime TTS to verify reduced latency at end of utterances

### DIFF
--- a/src/pipecat/services/tts_service.py
+++ b/src/pipecat/services/tts_service.py
@@ -1403,6 +1403,7 @@ class TTSService(AIService):
                 logger.trace(f"{self} time out on audio context {context_id}")
                 if should_push_stop_frame and self._push_stop_frames:
                     await self.push_frame(TTSStoppedFrame(context_id=context_id))
+                    should_push_stop_frame = False
                 break
 
         if should_push_stop_frame and self._push_stop_frames:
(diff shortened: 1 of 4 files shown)


src/pipecat/services/tts_service.py
```diff
@@ -1403,6 +1403,7 @@ async def _handle_audio_context(self, context_id: str):
                 logger.trace(f"{self} time out on audio context {context_id}")
                 if should_push_stop_frame and self._push_stop_frames:
                     await self.push_frame(TTSStoppedFrame(context_id=context_id))
+                    should_push_stop_frame = False
```
filipi87, line 1406: Nice catch. I totally missed this one.

filipi87 (approved):
LGTM!

PR #4173, opened 2026-03-27
Removing the models from the Inworld example so we can use the default model.

filipi87:
Removing the models from the Inworld example so we can use the default model.


examples/foundational/07zb-interruptible-inworld.py
```diff
@@ -56,7 +56,6 @@ async def run_bot(transport: BaseTransport, runner_args: RunnerArguments):
         api_key=os.getenv("INWORLD_API_KEY", ""),
         settings=InworldTTSService.Settings(
             voice="Ashley",
-            model="inworld-tts-1",
             temperature=1.1,
```
ianbbqzy, line 59: let's also change temperature to 1.0, which we recently evaluated to be better

PR #4175, opened 2026-03-27
Warn when TEXT modality is set for Gemini Live, and remove 26d text example

kompfner:
All recent Gemini Live models (including the default gemini-2.5-flash-native-audio-preview-12-2025, and going at least as far back as gemini-2.5-flash-native-audio-preview-09-2025) only support AUDIO as a response modality. We considered using `modalities=TEXT` as a Pipecat-level signal to suppress audio output frames (so developers could pair Gemini Live with an external TTS), but the output transcription from the API arrives too late relative to the audio to be useful for driving an external TTS service.

For now, just log a warning when a TEXT modality is configured (at init or via set_model_modalities) and proceed as normal. The 26d text-modality example is removed since it no longer represents a viable configuration.


markbackman (approved):
LGTM. Hopefully we can bring this back sometime in the future.

PR #4178, opened 2026-03-27
Keeping audio and clock queue in sync.

filipi87:
## Summary                              
- When a `TTSStoppedFrame` carries a presentation timestamp (`pts`), it is ow enqueued in both the `audio_queue` and the `clock_queue` so that timed frames (e.g. word-boundary events) and audio playback stay in sync.
- When the audio task finishes sending all preceding audio and dequeues the `TTSStoppedFrame`, it signals the clock task via `_clock_flush_event` to stop waiting on timestamps and drain all pending frames immediately.                                                   
- The clock task signals back via `_clock_drained_event` once the drain is complete, so the audio task can push `TTSStoppedFrame` downstream.                                                                       
- A `BOT_VAD_STOP_FALLBACK_SECS` timeout guards the wait on `_clock_drained_event` to prevent a deadlock if the clock task fails before completing the drain.

### DIFF
--- a/changelog/4178.fixed.md
+++ b/changelog/4178.fixed.md
@@ -1,0 +1,1 @@
+- Fixed timed frames (e.g. word-boundary events) arriving out of order or too late relative to TTS audio playback. When a `TTSStoppedFrame` carries a presentation timestamp, the clock queue now flushes all pending timed frames immediately once the audio task finishes sending the preceding audio, ensuring timed events always reach downstream processors before the stop signal.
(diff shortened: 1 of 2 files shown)


filipi87:
I tested all TTS services that provide word timestamps:
- Rime
- ElevenLabs
- Inworld
- Gradium
- Hume
- ResembleAI
- Azure
- Cartesia


Only Rime is returning incorrect timestamps.

However, the fix in the PR is not specifically addressing the Rime issue. Instead, it acts as a safeguard to ensure both queues stay in sync by using the TTSStopped frame as a gatekeeper.

But closing it for now, since Rime is currently the only service that would benefit from this change.

PR #4179, opened 2026-03-27
Don't send history_config for Gemini Live Vertex (unsupported)

kompfner:


filipi87 (approved):
LGTM!

kompfner:
I'm guessing this will be supported *soon*, since I think it will have to be once Gemini 3.1 Flash Live makes its way to Vertex (and Gemini 3.1 Flash Live makes more use of this config, if I'm not mistaken). But as is, it just crashes.

PR #4180, opened 2026-03-27
scripts(evals): run evals in parallel

aconchillo:

### DIFF
--- a/scripts/evals/eval.py
+++ b/scripts/evals/eval.py
@@ -16,6 +16,7 @@ from pathlib import Path
 from typing import Any, List, Optional, Tuple
 
 import aiofiles
+import aiohttp
 from loguru import logger
 from PIL.ImageFile import ImageFile
 from utils import (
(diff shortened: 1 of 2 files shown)


JiwaniZakir (commented):
In `run_eval`, the Daily room is created near the top of the function but deleted only at the bottom — outside any `try/finally` block. If an exception escapes after room creation (e.g., during task setup or the `asyncio.wait_for` call), the room will leak. Wrapping the body after `room = await helper.create_room(...)` in a `try/finally` that always calls `delete_room_by_name` would make the cleanup unconditional.

Additionally, two separate `aiohttp.ClientSession` instances are opened per eval (one for creation, one for deletion). Since these sessions are very short-lived and the client is stateless, this works correctly, but it's worth either reusing a single session within `run_eval` or sharing a session at the `EvalRunner` level across all concurrent evals to reduce connection overhead when many evals run in parallel.

The refactor to `_create_eval_callbacks` returning closures over a per-eval `result_future` is clean and correctly removes the shared `self._result_future` state that would have caused race conditions under parallel execution. The `result_future.done()` guard in `assert_eval` is a good addition for the case where both `EndFrame` and `CancelFrame` could fire close together.

aconchillo:
Closing. Doing this differently.

PR #4183, opened 2026-03-28
Yield after create_task to ensure timer tasks are scheduled

aconchillo:
## Summary

- Fixed a timing issue where turn detection timer tasks could miss their first tick because the newly created asyncio task had not yet been scheduled when the caller continued
- Affected components: `UserIdleController`, `SpeechTimeoutUserTurnStopStrategy`, `TurnAnalyzerUserTurnStopStrategy`, and `UserTurnCompletionLLMServiceMixin`
- Each `create_task()` call is now followed by `await asyncio.sleep(0)` to yield to the event loop and ensure the timer task is scheduled before proceeding

## Testing

- [x] Verify turn detection timing is consistent in voice conversations
- [x] Test idle controller, speech timeout, and turn analyzer strategies


markbackman (approved):
LGTM

PR #4186, opened 2026-03-28
Fix FastAPI WebSocket disconnect race condition

### LINKED ISSUE #3912: FastAPIWebsocketClient exception sending data and WebSocketDisconnect
### pipecat version

0.0.103

### Python version

3.13

### Operating System

Docker

### Issue description

Here is the sentry trace: https://into-ai.sentry.io/share/issue/c3f09cb56fe34c1f845c8017e9b31acc

Even though this error was raised, we didn’t encounter any actual problems. The workflow completed successfully, although sometimes it took longer than usual to close. The task eventually finished, and we were able to complete the operation after it ended.

This error occurred around 10 times out of 200 calls, but it didn’t break or affect anything. We’re not sure about the exact reason, so I’m just sharing it here to keep you updated on the behavior.

PIPELINE: STT(soniox), LLM(openai), TTS(elevenlabs)

### Reproduction steps

Twilio + FastAPI WebSocket

### Expected behavior

This error should not be raised.

### Actual behavior

Actual behavior: https://into-ai.sentry.io/share/issue/c3f09cb56fe34c1f845c8017e9b31acc

### Logs

```shell
2026-03-03 13:32:07.066 | ERROR    | pipecat.transports.websocket.fastapi:send:152 - <pipecat.transports.websocket.fastapi.FastAPIWebsocketClient object at 0x7d02cf45b950> exception sending data: WebSocketDisconnect (), application_state: WebSocketState.DISCONNECTED
```

markbackman:
## Summary

- Fix race condition where `send()` setting `_closing=True` on remote disconnect prevented the receive loop from firing `on_client_disconnected`, causing the pipeline to hang
- Add `application_state` check to `_can_send()` to suppress subsequent sends after a failed send without poisoning the `_closing` flag
- Downgrade send failure log from error to warning since this is a benign race condition in WebSocket teardown

## Details

When the remote side (e.g. Twilio) disconnects while TTS audio is being sent, there's a race between `send()` and the receive loop:

- **Before**: `send()` caught the exception and set `_closing=True`. If this happened before the receive loop checked `is_closing`, the `on_client_disconnected` callback was skipped, and the pipeline hung waiting for a disconnect signal that never came.
- **After**: `send()` no longer touches `_closing` (that flag means "we initiated the close"). Instead, `_can_send()` checks Starlette's `application_state` to suppress subsequent sends after a failure. The receive loop always sees `is_closing=False` for remote disconnects, so the callback fires correctly.

Fixes #3912

## Test plan

- [x] Added regression test that deterministically reproduces the race condition using `asyncio.Event` to enforce the exact ordering
- [x] Added tests verifying `send()` does not set `_closing` on remote disconnect
- [x] Added test verifying subsequent sends are suppressed via `application_state`
- [x] Verified fix with 30+ consecutive manual Twilio hangup tests with zero occurrences

### DIFF
--- a/changelog/4186.fixed.md
+++ b/changelog/4186.fixed.md
@@ -1,0 +1,1 @@
+- Fixed `FastAPIWebsocketTransport` intermittently hanging on shutdown when the remote side (e.g. Twilio) disconnects while audio is being sent. A race condition between the send and receive paths could cause the `on_client_disconnected` callback to be skipped, leaving the pipeline waiting for a disconnect signal that never came.
--- a/src/pipecat/transports/websocket/fastapi.py
+++ b/src/pipecat/transports/websocket/fastapi.py
@@ -150,17 +150,9 @@ class FastAPIWebsocketClient:
                 else:
                     await self._websocket.send_text(data)
         except Exception as e:
-            logger.error(
+            logger.warning(
                 f"{self} exception sending data: {e.__class__.__name__} ({e}), application_state: {self._websocket.application_state}"
             )
-            # For some reason the websocket is disconnected, and we are not able to send data
-            # So let's properly handle it and disconnect the transport if it is not already disconnecting
-            if (
-                self._websocket.application_state == WebSocketState.DISCONNECTED
-                and not self.is_closing
-            ):
-                logger.warning("Closing already disconnected websocket!")
-                self._closing = True
 
     async def disconnect(self):
         """Disconnect the WebSocket client."""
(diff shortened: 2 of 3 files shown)


markbackman:
I was able to reproduce the issue reported in ~1/5-10 tests before the fix. After the fix, I ran 30 tests without any issues. This gives me confidence that this is the right fix.

aconchillo (approved):

PR #4187, opened 2026-03-28
Fix heartbeat monitor timeout not respecting custom heartbeat interval

OmerCohenAviv:
## Summary

- Fixed a bug where setting `heartbeats_period_secs` to a value greater than 10 seconds would cause the heartbeat monitor to emit false "heartbeat frame not received" warnings, because the monitor timeout was hardcoded to 10 seconds regardless of the configured push interval.
- Added `heartbeats_monitor_secs` parameter to `PipelineParams` so users can configure the monitor timeout independently. Defaults to 10 seconds.
- Note: the current docs state "The monitor window is always 10x the interval," which is incorrect — it was hardcoded as `HEARTBEAT_SECS * 10` (10x the *default* constant of 1.0), so it was always 10 seconds regardless of what `heartbeats_period_secs` was set to. Docs should be updated to reflect the new `heartbeats_monitor_secs` parameter.


src/pipecat/pipeline/task.py
```diff
@@ -964,9 +967,9 @@ async def _heartbeat_monitor_handler(self):
         the time that a heartbeat frame takes to processes, that is how long it
         takes for the heartbeat frame to traverse all the pipeline.
         """
-        wait_time = HEARTBEAT_MONITOR_SECS
         while True:
             try:
+                wait_time = self._params.heartbeats_monitor_secs
```
markbackman: Can we move this outside of the `while True:` loop?

markbackman (commented):
Thanks for submitting! Looks good. Just one comment.

src/pipecat/pipeline/task.py
OmerCohenAviv: Done, thanks!

JiwaniZakir (commented):
The decoupling of `HEARTBEAT_MONITOR_SECS` from `HEARTBEAT_SECS * 10` introduces a subtle usability issue: if a caller sets `heartbeats_period_secs` to a large value (e.g., `5.0`) without also setting `heartbeats_monitor_secs`, the monitor will fire after only 2 missed heartbeats instead of the implied 10x ratio the original code suggested. Consider defaulting `heartbeats_monitor_secs` to `None` and resolving it in `_heartbeat_monitor_handler` as `self._params.heartbeats_monitor_secs or (self._params.heartbeats_period_secs * 10)`, which would preserve the original proportional relationship while still allowing explicit overrides.

In `tests/test_pipeline.py`, the call to `asyncio.get_event_loop()` inside the test (line `task.run(PipelineTaskParams(loop=asyncio.get_event_loop()))`) is deprecated since Python 3.10 and will emit a `DeprecationWarning` in async contexts; `asyncio.get_running_loop()` is the correct replacement here.

The test only validates that the warning message appears but doesn't assert a *timing* bound — it's conceivable the warning fires much later than `custom_monitor_secs` (e.g., due to scheduling delays) without the test catching it. Adding a rough upper-bound check (e.g., asserting the log is non-empty within `custom_monitor_secs * 2`) would make the test more meaningful as a regression guard for the timeout behavior specifically.

src/pipecat/pipeline/task.py
markbackman: Thank you!

markbackman (approved):
LGTM! Thanks for the contribution. 🙇

PR #4193, opened 2026-03-29
feat: Implement ACS Frame serializer

kedar389:
- Added ACSFrameSerializer for Azure Communication Services (ACS) Media Streaming WebSocket protocol                                                                                                                                                                                                                  
  - Supports bidirectional audio streaming with automatic PCM resampling between pipeline and ACS sample rates (16000 or 24000 Hz)
  - Handles DTMF input events, interruptions (InterruptionFrame → StopAudio), and automatic call hang-up on EndFrame/CancelFrame
  - Added acs-serializer optional dependency group: pip install pipecat-ai[acs-serializer]
  -  unit tests covering serialization, deserialization, hang-up logic, DTMF handling, and error cases
  
  Note: I know that this should probably be a community integration, but can we reconsider it? Azure communication services is a big telephony provider and pipecat already supports Azure STT, TTS and even LLM and its Realtime counterparts.
 This serializer has all the functionality implemented and has good test suites to cover it.


filipi87:

Closing this out for now.

kedar389:
@filipi87 I do not see why other serializers for other providers get mantained in the core and this one would not.  I guess it did not get considered.

markbackman:
@kedar389 one of the biggest factors is developer demand. We haven't had any requests for ACS, which is why we haven't added it yet. But, if you see demand, you're welcome to contribute as an integration and maintain it. Thanks again for proposing it.

PR #4199, opened 2026-03-30
Add VAD based user stop strategies

markbackman:
## Context

The changes I made in 0.0.108 hooked up Gemini Live to the local VAD directly, bypassing Pipecat's turn control logic.

This change makes it possible to write your own strategy. I've added a VAD/Turn analyzer stop strategy which takes action on the turn analyzer result without considering the transcript. This is necessary because Gemini Live is so slow in providing a transcript in the local VAD case that you need to forego the transcript check.

My general takeaway is that Gemini Live is inferior in the local VAD case and developers should always use the server-side VAD until Gemini Live better accommodates an external VAD. Given this, we should consider if this approach makes sense or not. I think it does as this gives developers more control, in exchange for slightly worse performance. WDYT?

## Summary

- Added `VADTurnAnalyzerUserTurnStopStrategy`, a turn stop strategy that triggers immediately when the turn analyzer reports COMPLETE without waiting for STT transcriptions. This reduces end-of-turn latency for speech-to-speech pipelines (e.g. Gemini Live) where audio goes directly to the LLM.
- Fixed `GeminiLiveLLMService` not responding to user speaking events when using local VAD turn management. The service was listening for `VADUserStartedSpeakingFrame`/`VADUserStoppedSpeakingFrame` instead of `UserStartedSpeakingFrame`/`UserStoppedSpeakingFrame`, which are the frames emitted by the user turn controller.
- Updated the `26a-gemini-live-local-vad` example to use the new `VADTurnAnalyzerUserTurnStopStrategy` with `VADUserTurnStartStrategy` (no transcription-based start strategy needed).

## Testing

```bash
uv run pytest tests/test_vad_turn_analyzer_user_turn_stop_strategy.py -v
```

### DIFF
--- a/changelog/4199.added.md
+++ b/changelog/4199.added.md
@@ -1,0 +1,1 @@
+- Added `VADTurnAnalyzerUserTurnStopStrategy`, a turn stop strategy that triggers immediately when the turn analyzer reports COMPLETE without waiting for STT transcriptions. This reduces end-of-turn latency for speech-to-speech pipelines (e.g. Gemini Live) where audio goes directly to the LLM.
--- a/examples/realtime/realtime-gemini-live-local-vad.py
+++ b/examples/realtime/realtime-gemini-live-local-vad.py
@@ -10,6 +10,7 @@ import os
 from dotenv import load_dotenv
 from loguru import logger
 
+from pipecat.audio.turn.smart_turn.local_smart_turn_v3 import LocalSmartTurnAnalyzerV3
 from pipecat.audio.vad.silero import SileroVADAnalyzer
 from pipecat.frames.frames import LLMRunFrame
 from pipecat.pipeline.pipeline import Pipeline
@@ -28,6 +29,9 @@ from pipecat.services.google.gemini_live.llm import GeminiLiveLLMService, Gemini
 from pipecat.transports.base_transport import BaseTransport, TransportParams
 from pipecat.transports.daily.transport import DailyParams
 from pipecat.transports.websocket.fastapi import FastAPIWebsocketParams
+from pipecat.turns.user_start import VADUserTurnStartStrategy
+from pipecat.turns.user_stop import VADTurnAnalyzerUserTurnStopStrategy
+from pipecat.turns.user_turn_strategies import UserTurnStrategies
 
 load_dotenv(override=True)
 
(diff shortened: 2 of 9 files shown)


changelog/4199.added.md
```diff
@@ -0,0 +1 @@
+- Added `VADTurnAnalyzerUserTurnStopStrategy`, a turn stop strategy that triggers immediately when the turn analyzer reports COMPLETE without waiting for STT transcriptions. This reduces end-of-turn latency for speech-to-speech pipelines (e.g. Gemini Live) where audio goes directly to the LLM.
```
markbackman, line 1: Not sure what the best name is... The current stop strategies all assume that a transcript is present. In hindsight, we might have wanted to include transcript somewhere in those name 🤷

examples/realtime/realtime-gemini-live-local-vad.py
```diff
@@ -73,6 +77,12 @@ async def run_bot(transport: BaseTransport, runner_args: RunnerArguments):
     user_aggregator, assistant_aggregator = LLMContextAggregatorPair(
         context,
         user_params=LLMUserAggregatorParams(
+            user_turn_strategies=UserTurnStrategies(
```
kompfner, line 80: I think if we actually wanted Gemini Live to base `ActivityStart`/`ActivityEnd` off of the turn decisions from this strategy, we'd also need to update `GeminiLiveLLMService` to go back to using `UserStarted/StoppedSpeakingFrame`s rather than the VAD ones.

Oh wait!! You snuck that change back in in https://github.com/pipecat-ai/pipecat/pull/4198/changes!!

changelog/4199.added.md
kompfner: An alternative that comes to mind: adding a parameter to the existing stop strategies for controlling `_stt_timeout`. The default would be "auto" (where the value comes from `STTMetadataFrame.ttfs_p99_latency`) but it could be overridable. Then we wouldn't need a new strategy, and the names we have—`SpeechTimeoutUserTurnStopStrategy`, `TurnAnalyzerUserTurnStopStrategy`—would still make sense.

kompfner: If you don't want to wait for a transcript, you could just set the timeouts to 0.

kompfner: Another alternative that comes to mind: we have Gemini Live push a `STTMetadataFrame` with a `ttfs_p99_latency` of 0 on `start`.

This solution is a bit awkward since it it's technically _possible_ that other listeners might care about a more meaningful `ttfs_p99_latency`. In other words, this solution requires that we assume `ttfs_p99_latency` is *only* meaningful for controlling turn strategy behavior.

The upside, though, is that it would be zero configuration for the user—they could simply enable VAD and things would work.

kompfner: Wait, hold on a sec...just realized that we *are* using an `stt_timeout` of 0 in the `TurnAnalyzerUserTurnStopStrategy` when there's no STT service in the pipeline...so for realtime services it appears we're *already* be doing the right thing and not waiting around for a transcript to declare the user turn finished!

kompfner: Shoot, just realized that we *also* need transcription text, in `TurnAnalyzerUserTurnStopStrategy._maybe_trigger_user_turn_stopped`...

kompfner: OK, so then maybe the thing that could be configurable in the stop strategies is not necessarily the timeout but whether a transcript is required?

markbackman:
Closing as we have other plans.

PR #4201, opened 2026-03-30
Fix WebsocketService infinite reconnection loop

markbackman:
## Summary

- Fixed `WebsocketService` infinite reconnection loop when a server accepts the WebSocket handshake but immediately closes the connection (e.g. invalid API key, close code 1008). The existing exponential backoff only handles handshake failures -- it cannot detect that a connection keeps dying immediately after being established.
- Added quick failure detection: tracks how long each connection survives and stops retrying after 3 consecutive connections that die within 5 seconds of being established.
- Changed all `WebsocketService` reconnection errors from fatal to non-fatal, allowing application-level failover (e.g. `ServiceSwitcher`) instead of killing the pipeline.

## Testing

```bash
uv run pytest tests/test_websocket_service.py -v
```

## Fixes

- Fixes:
  - #1006
  - #3711

### DIFF
--- a/changelog/4201.changed.md
+++ b/changelog/4201.changed.md
@@ -1,0 +1,1 @@
+- `WebsocketService` reconnection errors are now non-fatal. When a websocket service exhausts its reconnection attempts (either via exponential backoff or quick failure detection), it emits a non-fatal `ErrorFrame` instead of a fatal one. This allows application-level failover (e.g. `ServiceSwitcher`) to handle the failure instead of killing the entire pipeline.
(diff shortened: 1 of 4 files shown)


smokyabdulrahman (approved):

smokyabdulrahman:
We would appreciate having this fix in the next release 🙏
Anyway we could help?

markbackman:
Yes, for sure. This will be reviewed.

filipi87 (approved):
The changes look good. Nice improvement to the websocket service. 

I will merge so we can have a couple of days to bake it before the next release.

PR #4209, opened 2026-03-31
Change GrokLLMService default model to grok-3

markbackman:
## Summary
- Updates the default model for `GrokLLMService` from `grok-3-beta` to `grok-3` now that the model is generally available

## Test plan
- [x] Verify `GrokLLMService()` uses `grok-3` as the default model
- [x] Verify explicit model override still works via `settings` parameter


markbackman:
There are grok-4 models, but they're too slow for real-time use cases.

filipi87 (approved):
LGTM!

PR #4211, opened 2026-03-31
Introduce WebsocketLLMService and refactor OpenAIResponsesLLMService …

kompfner:
…to use it

Add WebsocketLLMService as a base class for WebSocket-based LLM services, parallel to WebsocketTTSService/WebsocketSTTService but codifying a transactional request-response model rather than a continuous background receive loop.

WebsocketLLMService provides:
- Connection lifecycle (start/stop/cancel → connect/disconnect)
- _ws_send/_ws_recv with transparent ConnectionClosed handling (auto-reconnect via exponential backoff → WebsocketReconnectedError)
- _ensure_connected with retry via _try_reconnect

OpenAIResponsesLLMService now inherits from WebsocketLLMService, removing duplicated connection management code (_connect, _disconnect, _reconnect, _ensure_connected, _ws_send, start, stop, cancel) and simplifying _process_context from a loop with attempt tracking to a flat try/except with a single retry.

---

This is a follow-up to #4141, [suggested](https://github.com/pipecat-ai/pipecat/pull/4141#discussion_r3009736631) by @filipi87

### DIFF
--- a/src/pipecat/services/llm_service.py
+++ b/src/pipecat/services/llm_service.py
@@ -1051,6 +1051,7 @@ class WebsocketLLMService(LLMService, WebsocketService):
         """
         LLMService.__init__(self, **kwargs)
         WebsocketService.__init__(self, reconnect_on_error=reconnect_on_error, **kwargs)
+        self._register_event_handler("on_connection_error")
 
     # -- lifecycle ------------------------------------------------------------
 
--- a/src/pipecat/services/llm_service.py
+++ b/src/pipecat/services/llm_service.py
@@ -1012,8 +1012,13 @@ class WebsocketLLMService(LLMService, WebsocketService):
 
     Provides connection lifecycle management (connect on start, disconnect
     on stop/cancel), automatic reconnection with exponential backoff, and
-    helpers for the per-inference exchange (``_ws_send``, ``_ws_recv``,
-    ``_ensure_connected``).
+    three helpers for running each inference:
+
+    1. ``_ensure_connected()`` — verify the websocket is alive, reconnect
+       with exponential backoff if not.
+    2. ``_ws_send(message)`` — send the inference request as JSON.
+    3. ``_ws_recv()`` — receive and parse response events one at a time
+       until the caller sees a terminal event.
 
     ``_ws_send`` and ``_ws_recv`` catch ``ConnectionClosed`` transparently,
     auto-reconnect via ``_try_reconnect``, and raise
--- a/src/pipecat/services/llm_service.py
+++ b/src/pipecat/services/llm_service.py
@@ -994,31 +994,31 @@ class WebsocketReconnectedError(Exception):
     """Raised by ``_ws_send``/``_ws_recv`` after a transparent reconnection.
 
     Signals that the WebSocket connection was lost and automatically
-    re-established.  Callers should treat this as a prompt to restart
-    the current transaction — any connection-local state on the server
-    (e.g. cached responses) is gone.
+    re-established.  The current inference should be restarted — any
+    connection-local state on the server (e.g. cached responses) is gone.
     """
 
     pass
 
 
 class WebsocketLLMService(LLMService, WebsocketService):
-    """Base class for websocket-based LLM services using a transactional model.
+    """Base class for websocket-based LLM services.
 
+    Each inference is a self-contained exchange: send a request, receive
+    events inline until a terminal event, then move on to the next.
     Unlike ``WebsocketTTSService`` / ``WebsocketSTTService`` which run a
-    continuous background receive loop (``_receive_task_handler``), LLM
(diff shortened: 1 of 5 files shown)


kompfner:
Still doing some final testing but wanted to get this open for review.

kompfner (commented):
## Code review

Found 2 issues. Checked for bugs and CLAUDE.md compliance.

src/pipecat/services/openai/responses/llm.py
```diff
@@ -711,9 +655,9 @@ async def _drain_cancelled_response(self):
                     )
                     self._clear_cancellation_state()
                     return
-        except (asyncio.TimeoutError, ConnectionClosed) as e:
-            logger.warning(f"{self}: Error draining cancelled response: {e} — reconnecting")
-            await self._reconnect()
+        except (asyncio.TimeoutError, WebsocketReconnectedError, ConnectionClosed) as e:
```
kompfner, line 662: **Bug:** `_drain_cancelled_response` can crash with `AttributeError` when `self._websocket` is `None`, causing the service to become permanently broken.

**Scenario:** `CancelledError` during inference → `_ws_send` of `response.cancel` hits a dead connection → reconnection fails → `_disconnect_websocket` clears `_needs_drain` but then `process_frame` sets `_needs_drain = True` again → next `_process_context` enters `_drain_cancelled_response` with `self._websocket = None` → line 628 (`self._websocket.recv()`) raises `AttributeError`.

Since this `except` clause doesn't catch `AttributeError`, it propagates without clearing `_needs_drain`, so every subsequent `_process_context` call hits the same error.

**Fix:** Add a guard at the top of the method body (after the docstring):

```python
if not self._websocket:
    self._clear_cancellation_state()
    return
```

src/pipecat/services/llm_service.py
```diff
@@ -978,3 +983,189 @@ async def _cancel_function_call(self, function_name: Optional[str]):
...
+
+    def __init__(self, *, reconnect_on_error: bool = True, **kwargs):
+        """Initialize the Websocket LLM service.
+
+        Args:
+            reconnect_on_error: Whether to automatically reconnect on websocket errors.
+            **kwargs: Additional arguments passed to parent classes.
+        """
+        LLMService.__init__(self, **kwargs)
+        WebsocketService.__init__(self, reconnect_on_error=reconnect_on_error, **kwargs)
```
kompfner, line 1053: **Bug:** `on_connection_error` event handler is documented but never registered.

`_report_error` calls `self._call_event_handler("on_connection_error", ...)`, but `_register_event_handler("on_connection_error")` is never called in `__init__`. Without registration, users who try `@llm.event_handler("on_connection_error")` will get a warning and their handler will be silently ignored.

The analogous `STTService` and `TTSService` both register this handler in their `__init__` methods ([stt_service.py L168](https://github.com/pipecat-ai/pipecat/blob/f2b3f8766127f075287299f9923694877b2f8daa/src/pipecat/services/stt_service.py#L167-L169), [tts_service.py L372](https://github.com/pipecat-ai/pipecat/blob/f2b3f8766127f075287299f9923694877b2f8daa/src/pipecat/services/tts_service.py#L371-L373)).

```suggestion
        LLMService.__init__(self, **kwargs)
        WebsocketService.__init__(self, reconnect_on_error=reconnect_on_error, **kwargs)
        self._register_event_handler("on_connection_error")
```

kompfner:
Note: I discovered that example 55zi doesn't work. In that example, we change the value of "temperature" on the fly. It appears that maybe "temperature" is not a supported parameter for the API.

kompfner:
> Note: I discovered that example 55zi doesn't work. In that example, we change the value of "temperature" on the fly. It appears that maybe "temperature" is not a supported parameter for the API.

Ah. It's an open bug in the OpenAI Python library: https://github.com/openai/openai-python/issues/2919.

Made a change to 55zi which documents the bug.

src/pipecat/services/openai/responses/llm.py
```diff
@@ -796,60 +750,61 @@ async def _process_context(self, context: LLMContext):
...
-                    return
+        async def send_and_receive(params: dict):
+            await self._ensure_connected()
+            await self.start_ttfb_metrics()
+            await self._ws_send({"type": "response.create", **params})
+            await self._receive_response_events(context, full_input)
+
+        async def cleanup():
+            self._clear_previous_response_state()
+            await self.stop_ttfb_metrics()
```
filipi87, line 769: I am just wondering whether these methods should stay in `_process_context` or belong to the class instead. 

Up to you, I don’t have strong opinion on this.

filipi87 (approved):
I think providing different behavior for the websocket LLM service makes sense for now, since, like you said, OpenAI is the only one using it.

LGTM! Thank you for doing this. 🚀

PR #4215, opened 2026-03-31
Remove deprecated `OpenAILLMContext` as well as everything (code path…

kompfner:
…s or whole types) dependent on it (all of which were also deprecated)


.pre-commit-config.yaml
kompfner, line 1: This is kind of an unrelated change, made in response to an annoyance I was hitting while working: explicitly running `uv run ruff...` was changing my files in a way that _conflicted_ with what the pre-commit hook was doing.

It turns out the pre-commit hook configuration here was pinning `ruff` to a _different_ version that what you'd get from a `uv sync --group dev`, which is the version you'd run locally as well as the version that GitHub actions will use. This brings the pre-commit hook into alignment (though it's possible you might need to run `uv sync --group dev` before this hook works?)

scripts/pre-commit.sh
kompfner, line 1: While I was investigating https://github.com/pipecat-ai/pipecat/pull/4215/changes#r3018339195, I noticed that we have this pre-commit script that, as far as I can tell, isn't used? We have .pre-commit-config.yaml already.

src/pipecat/frames/frames.py
```diff
@@ -706,44 +676,6 @@ def __str__(self):
         return f"{self.name}(pts: {pts}, signature: {self.signature})"
 
 
-@dataclass
-class LLMMessagesFrame(DataFrame):
```
kompfner, line 710: Also went ahead and removed this one, which has been deprecated for a loooong time.

src/pipecat/processors/aggregators/gated_open_ai_llm_context.py
kompfner, line 1: Ugh, we missed marking this one as deprecated before...I want to remove it...but understand if we need to keep it 😢 . Maybe it's not that big of a deal to have this little breaking change? It should be *very* easy for folks to fix their import.

kompfner:
I'm assuming [this change](https://github.com/pipecat-ai/pipecat/pull/4215/changes/19e521b75afc4059d45babf26e818ca04c6c8385) is safe—`LLMContextFrame` should always contain a context. The reason for the previous pattern was that multiple different kinds of frames could be carrying a context. But juuuuust in case the previous pattern was unwittingly also serving as a guard against context frames with missing contexts, I've kept this change in its own commit.

kompfner:
A fair bit of manual testing needs to be done, but going ahead and opening this for review

scripts/pre-commit.sh
```diff
@@ -1,27 +0,0 @@
-#!/bin/bash
```
markbackman, line 1: I think this was removed by mistake.

kompfner (commented):
## Code review

1 issue found. Checked for bugs and CLAUDE.md compliance.

scripts/pre-commit.sh
kompfner: Not by mistake - see [my comment](https://github.com/pipecat-ai/pipecat/pull/4215#discussion_r3018351486)

kompfner: But I may have misunderstood something

kompfner (commented):
**Bug: `UnboundLocalError` for `system_message`**

The old code initialized `system_message = None` before the conditional block. The refactoring removed that initialization when collapsing the `if isinstance(context, LLMContext)` / `elif` branches, but the reference on [line 543](https://github.com/pipecat-ai/pipecat/blob/ebab75765dfb6fb827dfbb36d1191c5f19f7c3dc/src/pipecat/utils/tracing/service_decorators.py#L542-L544) remains.

If none of the inner conditionals assign `system_message` (e.g. no `system_instruction` in settings AND context messages are empty or don't start with a system role), this will raise `UnboundLocalError`.

**Fix:** Add `system_message = None` before this line.

scripts/pre-commit.sh
markbackman: I see, yes, you're right. This script would only apply if run manually, which it's probably fine to assume doesn't happen. Ok, I retract my comment!

src/pipecat/utils/tracing/service_decorators.py
```diff
@@ -459,40 +458,30 @@ async def wrapped_start_llm_usage_metrics(tokens):
...
 
-                            if isinstance(context, OpenAILLMContext):
-                                # OpenAILLMContext and subclasses have their own method
-                                messages = context.get_messages_for_logging()
-                            elif isinstance(context, LLMContext):
-                                # Universal LLMContext - use adapter for provider-native format
-                                if hasattr(self, "get_llm_adapter"):
-                                    adapter = self.get_llm_adapter()
-                                    messages = adapter.get_messages_for_logging(context)
+                            # Use adapter for provider-native format
```
kompfner, line 466: I went back and forth over whether to keep `if isinstance(context, LLMContext)` juuuuust in case `context` is some other random thing...

...but I really don't think that's possible, unless someone fundamentally misuses the `traced_llm` decorator...

So I opted to take out this defensiveness in the code, to simplify and avoid misleading readers into thinking there's a good reason to expect that `context` could be something else.

src/pipecat/processors/aggregators/llm_response.py
```diff
@@ -4,106 +4,17 @@
...
-        expect_stripped_words: Whether to expect and handle stripped words
-            in text frames by adding spaces between tokens. This parameter is
-            ignored when used with the newer LLMAssistantAggregator, which
-            handles word spacing automatically.
-    """
-
-    expect_stripped_words: bool = True
 
 
 class LLMFullResponseAggregator(FrameProcessor):
```
markbackman, line 20: Kind of a bummer that this class lives on in this module, but I guess it's OK. We can't move it without breaking things. So, I guess we just let it be.

markbackman (approved):
LGTM!

I also had Claude review and it found no issues with the review.

Once you've done manual testing, this should be good to go!

Also, what a nice clean-up. This really simplifies the LLMService classes. 👏

src/pipecat/utils/tracing/service_decorators.py
markbackman: Makes sense.

src/pipecat/processors/aggregators/llm_response.py
kompfner: Yeah, I know, right...

src/pipecat/processors/aggregators/gated_open_ai_llm_context.py
markbackman: I think we should proceed with removing it. It's heavily related to OpenAILLMContext which is going away, so it makes sense to remove.

PR #4217, opened 2026-03-31
Supporting async function calls.

filipi87:
## Summary
- Added async function call support to `register_function()` and `register_direct_function()` via `cancel_on_interruption=False`. When set to `False`, the LLM continues the conversation immediately without waiting for the function result. The result is injected back into the context as a `developer` message once available, triggering a new LLM inference at that point.
- Added `group_id` to function calls so all calls originating from the same LLM response share an identifier. The LLM is triggered exactly once when the last call in the group completes.
- Fixed `MediaSender.handle_interruptions` to preserve and deliver pending `UninterruptibleFrame` items instead of discarding them on interruption
- Prevented spurious LLM inference: context frame is not pushed when a function call result arrives while the user is actively speaking
- Updated context summarization to treat `IN_PROGRESS` tool messages as unresolved, preventing async function calls from being summarized away before their results arrive.
- Added `group_parallel_tools` parameter to `LLMService` (default `True`). When `True`, all function calls from the same LLM response batch share a group ID and the LLM is triggered exactly once after the last call completes. Set to `False` to trigger inference independently for each function call result as it arrives.


src/pipecat/processors/frame_processor.py
```diff
@@ -865,9 +924,14 @@ async def __resume(self, frame: FrameProcessorResumeFrame | FrameProcessorResume
     async def _start_interruption(self):
         """Start handling an interruption by cancelling current tasks."""
         try:
-            if isinstance(self.__process_current_frame, UninterruptibleFrame):
-                # We don't want to cancel UninterruptibleFrame, so we simply
-                # cleanup the queue.
+            current_is_uninterruptible = isinstance(
+                self.__process_current_frame, UninterruptibleFrame
+            )
+            if current_is_uninterruptible or self.__process_queue.has_uninterruptible:
```
filipi87, line 825: We were not previously checking whether there was an `UninterruptibleFrame` in the queue and discarding it.

src/pipecat/transports/base_output.py
```diff
@@ -520,15 +524,21 @@ async def handle_interruptions(self, _: InterruptionFrame):
             if not self._transport._allow_interruptions:
                 return
 
-            # Cancel tasks.
-            await self._cancel_audio_task()
+            # Cancel all tasks.
             await self._cancel_clock_task()
             await self._cancel_video_task()
 
+            if self._audio_queue.has_uninterruptible:
```
filipi87, line 525: The same applies, if an `UninterruptibleFrame` was in the `_audio_queue`, it would be discarded.

src/pipecat/processors/frame_processor.py
aconchillo: We were doing it in the `__reset_process_queue` process no? Or was there a bug?

aconchillo: If the frame was interruptible, we were cancelling the task and calling `__reset_process_task()` which called `__reset_process_queue()`. So always keep uninterruptible frames.

aconchillo: OK. I understand now. This is just to have a queue that we can re-use.

In that case, I think we should move this to `utils`, and we also might want to call is `FrameQueue`. We assume that frame queues handle interruptible frames already.

filipi87: There was a bug as well, we were only checking whether the currently executing frame was an `UninterruptibleFrame`, but not whether there was an `UninterruptibleFrame` anywhere in the process queue.

filipi87: I will move this to utils. 👍

filipi87: Done.

filipi87:
@aconchillo, @markbackman, @kompfner, I believe I’ve addressed everything we discussed during our meeting. It would be great if you could take another look.

examples/function-calling/function-calling-anthropic.py
```diff
@@ -35,9 +35,10 @@
 load_dotenv(override=True)
 
 
-async def get_weather(params: FunctionCallParams):
-    location = params.arguments["location"]
-    await params.result_callback(f"The weather in {location} is currently 72 degrees and sunny.")
+async def fetch_weather_from_api(params: FunctionCallParams):
+    # Simulate a long-running API call, so we can test async function calls.
+    await asyncio.sleep(20)
```
aconchillo: I think we should create a new example. Maybe `function-calling-anthropic-delayed.py`, or maybe not even change the example. This will run in evalss, we can't wait for 20 seconds.

examples/function-calling/function-calling-openai.py
```diff
@@ -35,6 +39,8 @@
 
 
 async def fetch_weather_from_api(params: FunctionCallParams):
+    # Simulate a long-running API call, so we can test async function calls.
+    await asyncio.sleep(20)
```
aconchillo: Same here.

examples/function-calling/function-calling-anthropic.py
filipi87: Makes sense. I’ll create just one new example for OpenAI and revert the changes here.

filipi87: Done!

examples/function-calling/function-calling-openai.py
filipi87: Done!

src/pipecat/processors/aggregators/llm_response_universal.py
```diff
@@ -1047,13 +1055,24 @@ async def _handle_function_call_in_progress(self, frame: FunctionCallInProgressF
...
-                "tool_call_id": frame.tool_call_id,
-            }
-        )
+
+        is_async = not frame.cancel_on_interruption
+        if is_async:
+            self._context.add_message(
+                {
+                    "role": "tool",
+                    "content": json.dumps({"type": "tool", "status": "started"}),
```
filipi87: @aconchillo , @kompfner , I think this is the last thing we need to decide, whether we are going to call it "tool" or "async_tool".

kompfner:
> Added group_parallel_tools parameter to LLMService (default True). When True, all function calls from the same LLM response batch share a group ID and the LLM is triggered exactly once after the last call completes. Set to False to trigger inference independently for each function call result as it arrives.

Do we need this parameter? Based on a bit of research, to me it seems like opting out would mean going against best practice and violating LLM expectations. We could choose to not expose this, and add it if/when we have a clear need for it.

src/pipecat/processors/aggregators/llm_response_universal.py
kompfner: Or distinguish _even further_ from the language of "tool calling" or "function calling", and go with "task" or "job" or something like that—in my understanding of things, the more distinct we make this mechanism from "vanilla" tool calling, the easier it will be for LLMs (and honestly for human readers of context) to understand. 

There's no downside, in my mind, to underscoring in bold that when you (the LLM) are calling `get_weather`, you're not getting the weather, you're actually kicking off a job to eventually get the weather.

kompfner: I have a suspicion that if we tried to maximize clarity (and honesty about what's going on) then we may be able to fix issues like the LLM trying to re-invoke an already-running async tool on a subsequent turn. We need more natural language. This is a prompt engineering issue, I think.

For example, Claude goes pretty far and suggests messages that look like this:

```
"content": "Task dispatched asynchronously. job_id: job_xyz789. Result not yet available — it will be provided in a subsequent user message. Do not speculate about or act on the result until then."
```

I don't think we need to abandon the result being JSON (which has its benefits, as @aconchillo pointed out a while back, and which is leveraged in the changes in this PR), but the above is a lot closer to what I've had in mind. Maybe something like:

```
"content": json.dumps({ 
  "type": "async_task",
  "status": "dispatched",
  "task_id": frame.tool_call_id,
  "message": "Task dispatched asynchronously. Result not yet available. It will be provided in a subsequent message with matching task ID. Do not speculate about or act on the result until then."
})
```

(And of course we can adjust the `"message"` to add support for multi-result streaming tasks)

filipi87:
```
"content": json.dumps({ 
  "type": "async_tool",
  "status": "started",
  "tool_call_id": frame.tool_call_id,
  "description": "The tool with this tool_call_id is in progress and the result is not yet available. It will be provided in a subsequent message with matching tool_call_id."
})
```

filipi87: Try with the `CerebrasLLMService` to see if it works fine.

filipi87: I have applied the changes we discussed. 👍

changelog/4217.changed.md
```diff
@@ -0,0 +1 @@
+- When multiple function calls are returned in a single LLM response, the LLM is now triggered exactly once after the last call in the batch completes, rather than waiting for all function calls.
```
kompfner: Do we want to change this wording to account for the new `group_parallel_tools` parameter?

"...by default (when `group_parallel_tools=True`) the LLM is now triggered once after the last call in the batch completes, rather than waiting for all function calls (which could've included ongoing calls from previous batches)."

examples/function-calling/function-calling-anthropic-delayed.py
```diff
@@ -0,0 +1,174 @@
...
+from pipecat.services.llm_service import FunctionCallParams
+from pipecat.transports.base_transport import BaseTransport, TransportParams
+from pipecat.transports.daily.transport import DailyParams
+from pipecat.transports.websocket.fastapi import FastAPIWebsocketParams
+
+load_dotenv(override=True)
+
+
+async def fetch_weather_from_api(params: FunctionCallParams):
+    # Simulate a long-running API call, so we can test async function calls.
```
kompfner: Might be helpful to the reader to say what you mean by "async function call" here, to disambiguate from the Python function definition, which has `async` in it.

```suggestion
    # Simulate a long-running API call, so we can test async function calls (cancel_on_interruption=False).
```

examples/function-calling/function-calling-anthropic-async.py
kompfner, line 1: Nit: do we want to name these files `*-async.py`, in keeping with the terminology "async function call" that you use elsewhere?

examples/function-calling/function-calling-openai-async.py
kompfner, line 1: You mentioned that other LLMs also handle async functioning well, too. Should we get at least Google in this set of examplees, too?

kompfner: Also...I know that `OpenAIResponsesLLMService` is still fairly new, but we should eventually start treating it as the "main" or "recommended" OpenAI service. We're maybe not quite there yet, but it might be worth testing it and including it as an example this group of example. WDYT?

src/pipecat/frames/frames.py
```diff
@@ -1642,12 +1642,19 @@ class FunctionCallInProgressFrame(ControlFrame, UninterruptibleFrame):
         tool_call_id: Unique identifier for this function call.
         arguments: Arguments passed to the function.
         cancel_on_interruption: Whether to cancel this call if interrupted.
+            When ``False`` the call is treated as asynchronous: the LLM
```
kompfner, line 1645: 👍 Nice clear and concise description here

tests/test_context_aggregators_universal.py
```diff
@@ -806,7 +806,7 @@ async def test_function_call(self):
                 function_name="get_weather",
                 tool_call_id="1",
                 arguments={"location": "Los Angeles"},
-                cancel_on_interruption=False,
+                cancel_on_interruption=True,
```
kompfner, line 809: Why this change?

src/pipecat/frames/frames.py
```diff
@@ -1642,12 +1642,19 @@ class FunctionCallInProgressFrame(ControlFrame, UninterruptibleFrame):
...
+            result, and the result is injected later via a developer message.
+        group_id: Identifier shared by all function calls originating from the
+            same LLM response batch. Used to determine when the last call in a
+            group completes so the LLM can be triggered exactly once.
     """
 
     function_name: str
     tool_call_id: str
     arguments: Any
     cancel_on_interruption: bool = False
```
kompfner, line 1656: Not to revisit an old decision, but: why does this field have a default value? Won't we always specify it?

kompfner (approved):
Had a few minor questions/suggestions. Pre-approving, given all your testing. Nice work!

changelog/4217.changed.md
filipi87: Done.

tests/test_context_aggregators_universal.py
filipi87: Because the test was expecting the old syntax, it was trying to read the context from the replaced result message.
`assert json.loads(context.messages[-1]["content"]) == {"conditions": "Sunny"}`

src/pipecat/frames/frames.py
filipi87: Yeah, as far as I can tell, we always do.

examples/function-calling/function-calling-anthropic-async.py
filipi87: Yep, I think it makes sense. 
I have just renamed the OpenAI and Anthropic examples to “async” instead of “delayed.” 👍

examples/function-calling/function-calling-openai-async.py
filipi87: Done. I have added two new examples, one for Google and the other one using OpenAI Response. 👍

PR #4223, opened 2026-04-01
Add mistv3 model support to Rime TTS services

mcullan:
Adds `mistv3` as a supported model in `RimeTTSService` and `RimeHttpTTSService`.

`mistv3` supports `speedAlpha`,`inlineSpeedAlpha`, `pauseBetweenBrackets`, and `phonemizeBetweenBrackets`. 

`reduceLatency`, `noTextNormalization`, and `saveOovs` are mistv2-only and are excluded.

### DIFF
--- a/src/pipecat/services/rime/tts.py
+++ b/src/pipecat/services/rime/tts.py
@@ -353,6 +353,8 @@ class RimeTTSService(WebsocketTTSService):
                 params["pauseBetweenBrackets"] = json.dumps(self._settings.pauseBetweenBrackets)
             if self._settings.phonemizeBetweenBrackets is not None:
                 params["phonemizeBetweenBrackets"] = json.dumps(self._settings.phonemizeBetweenBrackets)
+            if self._settings.inlineSpeedAlpha is not None:
+                params["inlineSpeedAlpha"] = self._settings.inlineSpeedAlpha
         else:  # mistv2/mist
             if self._settings.reduceLatency is not None:
                 params["reduceLatency"] = self._settings.reduceLatency
--- a/src/pipecat/services/rime/tts.py
+++ b/src/pipecat/services/rime/tts.py
@@ -348,6 +348,8 @@ class RimeTTSService(WebsocketTTSService):
                 params["temperature"] = self._settings.temperature
             if self._settings.top_p is not None:
                 params["top_p"] = self._settings.top_p
+        elif self._settings.model == "mistv3":
+            pass  # only speedAlpha applies, already added above
         else:  # mistv2/mist
             if self._settings.reduceLatency is not None:
                 params["reduceLatency"] = self._settings.reduceLatency
--- a/src/pipecat/services/rime/tts.py
+++ b/src/pipecat/services/rime/tts.py
@@ -75,10 +75,10 @@ class RimeTTSSettings(TTSSettings):
 
     Parameters:
         segment: Text segmentation mode ("immediate", "bySentence", "never").
-        speedAlpha: Speech speed multiplier (mistv2 only).
+        speedAlpha: Speech speed multiplier (mist, mistv2, mistv3).
         reduceLatency: Whether to reduce latency at potential quality cost (mistv2 only).
-        pauseBetweenBrackets: Whether to add pauses between bracketed content (mistv2 only).
-        phonemizeBetweenBrackets: Whether to phonemize bracketed content (mistv2 only).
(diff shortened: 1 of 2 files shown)


aresgeo (approved):

aresgeo (approved):

markbackman:
Please add a changelog entry. See CONTRIBUTING.md for details:
https://github.com/pipecat-ai/pipecat/blob/main/CONTRIBUTING.md

mcullan:
@markbackman changelog entry added!

markbackman:
Is this ready for review?

markbackman:
Based on the code, the PR description is out of date. Is that right?

mcullan:
@markbackman apologies for the discrepancy/changes there -- we ended up landing more features for Mist v3 than expected yesterday. With the latest commit and the update to the description, this is now 100% ready for review and the description is accurate to the code.

src/pipecat/services/rime/tts.py
```diff
@@ -348,6 +348,13 @@ def _build_ws_params(self) -> dict[str, Any]:
                 params["temperature"] = self._settings.temperature
             if self._settings.top_p is not None:
                 params["top_p"] = self._settings.top_p
+        elif self._settings.model == "mistv3":
+            if self._settings.pauseBetweenBrackets is not None:
+                params["pauseBetweenBrackets"] = json.dumps(self._settings.pauseBetweenBrackets)
+            if self._settings.phonemizeBetweenBrackets is not None:
+                params["phonemizeBetweenBrackets"] = json.dumps(self._settings.phonemizeBetweenBrackets)
```
markbackman, line 355: This line is 101 characters — will fail the ruff 100-char limit.

Can you run `uv run scripts/fix-ruff.sh` or install the pre-commit hook (`uv run pre-commit install`)?

src/pipecat/services/rime/tts.py
```diff
@@ -348,6 +348,13 @@ def _build_ws_params(self) -> dict[str, Any]:
                 params["temperature"] = self._settings.temperature
             if self._settings.top_p is not None:
                 params["top_p"] = self._settings.top_p
+        elif self._settings.model == "mistv3":
```
markbackman, line 351: Suggestion: consider restructuring to avoid duplicating `pauseBetweenBrackets` and `phonemizeBetweenBrackets` handling (same logic exists in the `else` branch). Something like:

```python
if self._settings.model == "arcana":
    # arcana-only params
    ...
else:
    # shared mist/mistv2/mistv3 params
    if self._settings.pauseBetweenBrackets is not None:
        params["pauseBetweenBrackets"] = json.dumps(self._settings.pauseBetweenBrackets)
    if self._settings.phonemizeBetweenBrackets is not None:
        params["phonemizeBetweenBrackets"] = json.dumps(
            self._settings.phonemizeBetweenBrackets
        )

    if self._settings.model == "mistv3":
        if self._settings.inlineSpeedAlpha is not None:
            params["inlineSpeedAlpha"] = self._settings.inlineSpeedAlpha
    else:  # mist/mistv2
        if self._settings.reduceLatency is not None:
            params["reduceLatency"] = self._settings.reduceLatency
        ...
```

This way the shared params are written once, reducing the chance of them drifting apart.

markbackman (changes requested):
Nice addition — just two items to address.

JiwaniZakir (commented):
The docstring update on line ~77 correctly notes that `speedAlpha` now supports `mist, mistv2, mistv3`, but the new `elif self._settings.model == "mistv3":` block in `_build_ws_params()` omits `speedAlpha` entirely—only `pauseBetweenBrackets`, `phonemizeBetweenBrackets`, and `inlineSpeedAlpha` are added. If `speedAlpha` is genuinely supported by mistv3, it needs to be wired up in that block; if not, the docstring update is incorrect and should remain `mistv2 only`. Conversely, `inlineSpeedAlpha` is included in the mistv3 block but its docstring doesn't mention mistv3 support, so the documentation is inconsistent in both directions. It's also worth double-checking whether `reduceLatency`, `noTextNormalization`, and `saveOovs` truly remain mistv2-only or if mistv3 picks up any of those as well, since leaving them in the `else` branch means they'd silently be sent for mistv2 but not mistv3.

markbackman:
Closing as this PR has gone stale. `coda` is now the latest model. The Rime team pushed updates for the upcoming release.

PR #4224, opened 2026-04-01
Make function_call_timeout_secs optional

aconchillo:
## Summary

- `LLMService.function_call_timeout_secs` now defaults to `None` instead of `10.0`, allowing deferred function calls to run without a timeout by default
- The timeout task is only created when a timeout is actually configured (service-level or per-call)

## Breaking Changes

- `function_call_timeout_secs` default changed from `10.0` to `None`. If you relied on the implicit 10-second timeout for deferred function calls, pass `function_call_timeout_secs=10.0` explicitly.


filipi87 (approved):
LGTM!

PR #4230, opened 2026-04-02
Support for streaming multiple responses via function calls

filipi87:
## Summary
- Added support for streaming intermediate results from async function calls — handlers can now call `result_callback` multiple times with `properties=FunctionCallResultProperties(is_final=False)` before sending the final result                             
- Fixed duplicate LLM replies that could occur when multiple async function call results arrived while an LLM request was already queued.                                                            
- Added new examples: `function-calling-openai-async-stream.py` and `function-calling-openai-responses-async-stream.py`

### DIFF
--- a/examples/function-calling/function-calling-openai-async-stream.py
+++ b/examples/function-calling/function-calling-openai-async-stream.py
@@ -1,0 +1,205 @@
+#
+# Copyright (c) 2024-2026, Daily
+#
+# SPDX-License-Identifier: BSD 2-Clause License
+#
+
+"""Example: async function call with intermediate updates.
+
+The ``track_current_location`` tool simulates a GPS tracker reporting the
+device's position during a road trip from San Francisco to San Diego.  It
+sends two intermediate updates (via ``params.result_callback`` with
+``is_final=False``) as the vehicle passes through cities along the way, then
+delivers the final destination (via ``params.result_callback``).  Each update
+returns the same structure with a different city:
+
+  Update 1 – {gps, city: "San Francisco"}   ← trip start
+  Update 2 – {gps, city: "Los Angeles"}     ← passing through
+  Final     – {gps, city: "San Diego"}      ← destination reached
+
+Because the function is registered with ``cancel_on_interruption=False``, the
+LLM can keep talking while the trip is in progress; each position update
+arrives as a developer message so the LLM can narrate the journey to the user.
+"""
+
+import asyncio
+import os
+
+from dotenv import load_dotenv
+from loguru import logger
+
+from pipecat.adapters.schemas.function_schema import FunctionSchema
+from pipecat.adapters.schemas.tools_schema import ToolsSchema
+from pipecat.audio.vad.silero import SileroVADAnalyzer
+from pipecat.frames.frames import (
+    FunctionCallResultProperties,
+    LLMRunFrame,
+    TTSSpeakFrame,
+)
+from pipecat.pipeline.pipeline import Pipeline
+from pipecat.pipeline.runner import PipelineRunner
+from pipecat.pipeline.task import PipelineParams, PipelineTask
+from pipecat.processors.aggregators.llm_context import LLMContext
+from pipecat.processors.aggregators.llm_response_universal import (
+    LLMContextAggregatorPair,
+    LLMUserAggregatorParams,
+)
+from pipecat.runner.types import RunnerArguments
+from pipecat.runner.utils import create_transport
+from pipecat.services.cartesia.tts import CartesiaTTSService
+from pipecat.services.deepgram.stt import DeepgramSTTService
+from pipecat.services.llm_service import FunctionCallParams
+from pipecat.services.openai.llm import OpenAILLMService
+from pipecat.transports.base_transport import BaseTransport, TransportParams
+from pipecat.transports.daily.transport import DailyParams
+from pipecat.transports.websocket.fastapi import FastAPIWebsocketParams
+
+load_dotenv(override=True)
+
+
+async def track_current_location(params: FunctionCallParams):
+    """Simulate a GPS tracker reporting position during a road trip.
+
+    Step 1 – San Francisco (trip start)     (update)
+    Step 2 – Los Angeles   (passing through) (update)
+    Step 3 – San Diego     (destination)     (final result)
+    """
+
+    # First update: initial city estimate.
+    gps = {"lat": 37.7310, "lng": -122.4527}
+    await params.result_callback(
+        {"gps": gps, "city": "San Francisco"},
+        properties=FunctionCallResultProperties(is_final=False),
+    )
+
+    # Second update: revised city estimate.
+    await asyncio.sleep(10)
+    gps = {"lat": 33.96003, "lng": -118.40639}
+    await params.result_callback(
+        {"gps": gps, "city": "Los Angeles"},
+        properties=FunctionCallResultProperties(is_final=False),
+    )
+
+    # Final result: confirmed city.
+    await asyncio.sleep(10)
+    gps = {"lat": 32.743569, "lng": -117.20466}
+    await params.result_callback({"gps": gps, "city": "San Diego"})
+
+
+# We use lambdas to defer transport parameter creation until the transport
+# type is selected at runtime.
+transport_params = {
+    "daily": lambda: DailyParams(
+        audio_in_enabled=True,
+        audio_out_enabled=True,
+    ),
+    "twilio": lambda: FastAPIWebsocketParams(
+        audio_in_enabled=True,
+        audio_out_enabled=True,
+    ),
+    "webrtc": lambda: TransportParams(
+        audio_in_enabled=True,
+        audio_out_enabled=True,
+    ),
+}
+
+
+async def run_bot(transport: BaseTransport, runner_args: RunnerArguments):
+    logger.info(f"Starting bot")
+
+    stt = DeepgramSTTService(api_key=os.getenv("DEEPGRAM_API_KEY"))
+
+    tts = CartesiaTTSService(
+        api_key=os.getenv("CARTESIA_API_KEY"),
+        settings=CartesiaTTSService.Settings(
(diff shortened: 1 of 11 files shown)


src/pipecat/services/llm_service.py
```diff
@@ -776,6 +802,32 @@ async def function_call_result_callback(
...
+                    " update_callback is only valid for async function calls"
+                    " (cancel_on_interruption=False)."
+                )
+                return
+            # Force is_final=False so the aggregator treats this as an update.
+            update_properties = FunctionCallResultProperties(
+                run_llm=properties.run_llm if properties else None,
+                on_context_updated=properties.on_context_updated if properties else None,
+                is_final=False,
+            )
```
aconchillo: I guess this makes it possible for updates to run inference. Makes sense.

examples/function-calling/function-calling-openai-async-stream.py
```diff
@@ -0,0 +1,206 @@
...
+        cancel_on_interruption=False,
+        timeout_secs=30,
+    )
+
+    @llm.event_handler("on_function_calls_started")
+    async def on_function_calls_started(service, function_calls):
+        await tts.queue_frame(TTSSpeakFrame("Sure, tracking your location now."))
+
+    location_function = FunctionSchema(
+        name="get_current_location",
```
kompfner: Since this function represents a request to track location, as opposed to a one-time lookup, maybe it should be "track_current_location"

examples/function-calling/function-calling-openai-async-stream.py
kompfner, line 1: Do we also want examples for Google and Anthropic, like we do for one-shot (non-streaming) async function calling?

src/pipecat/processors/aggregators/llm_response_universal.py
```diff
@@ -1137,6 +1123,66 @@ async def _handle_function_call_result(self, frame: FunctionCallResultFrame):
...
+
+        result = json.dumps(frame.result, ensure_ascii=False)
+        self._context.add_message(
+            {
+                "role": "developer",
+                "content": json.dumps(
+                    {
+                        "type": "async_tool",
+                        "tool_call_id": frame.tool_call_id,
+                        "status": "updated",
```
kompfner: The "updated" nomenclature, used throughout this PR, doesn't feel _quite_ right to me...it's a bit ambiguous whether "update" means "modification" or "new non-final result"...

kompfner: One idea: just expose `is_final` to users.

```python
await params.result_callback(value, is_final=False)
```

That would also address another minor concern I had when reading this code, which is the mismatch between the names `update_callback` and `FunctionCallResultProperties` (which is used by `update_callback`, too).

Here we could change the message content to:
```python
"result": result,
"is_final_result": False # "is_final" might be slightly ambiguous here
```

Thoughts?

kompfner: Ah, actually...I see why you want to use "status". Maybe "received_partial_results" or something?

kompfner: Or maybe even clearer:

Initial message:
```
"status": "running"
```

Non-final result:
```
"status": "running", # no change
"result": result,
"is_result_final": False # redundant, since "status" is still "running", but is sometimes helpful to be extra extra clear to LLMs
```

Final result:
```
"status": "finished",
"result": result,
"is_result_final": True
```

That also decouples the status of the task (whether it's running or not) from results, which might be helpful later on for task cancellation or handling task failure.

src/pipecat/processors/aggregators/llm_response_universal.py
kompfner, line 1: We may want to update the "description" of the initial "async_tool" message to account for the new streaming behavior.

kompfner: We also may want to add a "description" to the non-final (and maybe even final) result messages in context, for additional clarity.

examples/function-calling/function-calling-openai-async-stream.py
filipi87: Done.

examples/function-calling/function-calling-openai-async-stream.py
```diff
@@ -0,0 +1,206 @@
...
+    )
+
+    llm = OpenAILLMService(
+        api_key=os.getenv("OPENAI_API_KEY"),
+        settings=OpenAILLMService.Settings(
+            system_instruction=(
+                "You are a helpful assistant in a voice conversation. "
+                "Your responses will be spoken aloud, so avoid emojis, bullet points, or other "
+                "formatting that can't be spoken. "
+                "When you receive an tool update message, tell the user the current city "
```
kompfner: It'd be nice if we didn't have to describe the whole streaming results mechanism in the system prompt, just domain-specific info, like:

"You have access to a function that starts tracking a moving device's location and provides regular updates on it"

examples/function-calling/function-calling-openai-async-stream.py
filipi87: Yep, I can create them once we’re happy with the OpenAI one. 👍

src/pipecat/processors/aggregators/llm_response_universal.py
filipi87: Ok, I did both. Let me know what you think.

src/pipecat/processors/aggregators/llm_response_universal.py
filipi87: Ok, I have implemented it like this:

- Non-final result:
```
"status": "running",
"description": "Intermediate update. More updates or a final 'finished' message will follow.",
"result": result,
````

- Final result:
```
"status": "finished",
"description": "Final result. No further messages for this tool_call_id.",
"result": result,
```

I also tried adding `is_result_final` at first, but removing it doesn’t seem to make any difference. 
So, I thought it would be better not to duplicate the information in the final message (`is_result_final=True` and `"status": "finished"`).

Let me know what you think, if should we keep `is_result_final` or not ?

File: examples/function-calling/function-calling-openai-async-stream.py (hunk 4 above)
Comment (filipi87) [reply examples/function-calling/function-calling-openai-async-stream.py]: It looks like after the recent changes that we did in this PR, I have been able to simplify the prompt.

So, it is just like this now:

```
system_instruction=(
    "You are a helpful assistant in a voice conversation. "
    "Your responses will be spoken aloud, so avoid emojis, bullet points, or other "
    "formatting that can't be spoken. "
    "You have access to a function that starts tracking a moving device's location and "
    "provides regular updates on it. When you receive the final location, tell the user "
    "the destination has been reached and announce the final city."
 ),
```

File: src/pipecat/processors/aggregators/llm_response_universal.py [hunk 5]
```diff
@@ -284,7 +284,14 @@ def _get_context_frame(self) -> LLMContextFrame:
...
         """
-        return LLMContextFrame(context=self._context)
+        # We pass a snapshot of the context rather than a live reference.
+        # LLMContextFrame processing is asynchronous: by the time the LLM
+        # service reads the frame, new messages may have already been added to
+        # the original context (e.g. a function-call result arriving while the
+        # LLM is still queued). Without a snapshot, all in-flight frames share
+        # the same object and see the same ever-growing message list, causing
+        # the LLM to generate redundant or duplicate replies.
+        return LLMContextFrame(context=self._context.copy())
```
filipi87: @kompfner, this is fixing the LLM duplicated replies.

kompfner: This solution, I think, makes sense. With this solution in place, if we have the scenario

- user starts speaking
- function results comes in (`A`)
- user finishes speaking (`B`)

then the LLM would run two inferences back to back:

- response to context `[A]` -> `A'`
- response to context `[A, B]`

Right? Or would its second response be with context `[A, A', B]`?

If not...actually, isn't that still a problem? The model's second response wouldn't take into account the fact that it *already* started responding in a way that addressed the first message (`A`)? So the second response might overlap/duplicate a portion of the first response?

kompfner: The alternative would be to keep the context as a "live reference" and somehow try to only run inference once if the context has not changed, so that in the above scenario we'd get the LLM running a single inference with `[A, B]`.

kompfner: Hmm, yeah, as expected in my testing I'm finding that the fact of the context not being "live" means that some responses are using out-of-date context. Here's the behavior I witnessed:

- inference: `[<arrived in san diego>]` -> "You've arrived at your destination, San Diego!"
- user interrupts with "tell me a joke"
- inference: `[<arrived in san diego>, <user: "tell me a joke">]` -> "Here's a joke: <joke>. Oh, and by the way, you're at your destination, San Diego!"

In the above scenario, we're dropping assistant responses on subsequent inferences after interruptions. That's a pretty big breakage.

I wouldn't be surprised if there were other ways, too, that we were implicitly relying on the fact of context being "live". This change feels a smidge dangerous to me.

filipi87: If the user is speaking, we do not send a context frame, so the LLM only runs with the context [A, A', B].

kompfner: @filipi87 and I chatted, and he pointed out that the issue I was experiencing, above, was simply due to the fact that the example was using an STT service that was doesn't support word timestamps, so the assistant response was lost in its entirety on the interruption.

kompfner: Also discussed in our chat: the issue of async stream tool results arriving close together (say, one or two seconds apart)...with only the fix in this code in place, the bot would generate a response to every single result, in sequence. It could take a long time for it to "catch up" to the latest results.

Probably a better default behavior would be to "bundle" the streamed results so that the bot can describe batches of them all together, then when it's done it can move on to describing the next set of results received in the meantime, while it was talking.

@filipi87 already has a plan in mind for this!

kompfner (approved):
Looks great!

I think the only thing left is adding and testing the other non-OpenAI examples.

PR #4231, opened 2026-04-02
Add a `LLMMessagesTransformFrame` to facilitate programmatically edit…

kompfner:
…ing context in a frame-based way.

The previous approach required the caller to directly grab a reference to the context object, grab a "snapshot" of its messages *at that point in time*, transform the messages, and then push an `LLMMessagesUpdateFrame` with the transformed messages. This approach can lead to problems: what if there had already been a change to the context queued in the pipeline? The transformed messages would simply overwrite it without consideration.

---

This PR salvages just one change from closed PR #3620.

### DIFF
--- a/src/pipecat/frames/frames.py
+++ b/src/pipecat/frames/frames.py
@@ -594,8 +594,6 @@ class LLMMessagesTransformFrame(DataFrame):
     A frame containing a transform function that takes the context's current list
     of LLM messages and returns a modified list.
 
-    Only compatible with LLMContext and not the deprecated OpenAILLMContext.
-
     Parameters:
         transform: A function that takes a list of messages and returns a
             modified list.
--- a/src/pipecat/frames/frames.py
+++ b/src/pipecat/frames/frames.py
@@ -38,7 +38,7 @@ from pipecat.utils.time import nanoseconds_to_str
 from pipecat.utils.utils import obj_count, obj_id
 
 if TYPE_CHECKING:
-    from pipecat.processors.aggregators.llm_context import LLMContext, NotGiven
+    from pipecat.processors.aggregators.llm_context import LLMContext, LLMContextMessage, NotGiven
     from pipecat.processors.frame_processor import FrameProcessor
     from pipecat.services.settings import ServiceSettings
     from pipecat.utils.context.llm_context_summarization import LLMContextSummaryConfig
@@ -584,6 +584,25 @@ class LLMMessagesUpdateFrame(DataFrame):
     """
 
     messages: List[dict]
+    run_llm: Optional[bool] = None
+
+
+@dataclass
+class LLMMessagesTransformFrame(DataFrame):
+    """Frame containing a transform function to modify the current context's LLM messages.
+
+    A frame containing a transform function that takes the context's current list
+    of LLM messages and returns a modified list.
+
+    Only compatible with LLMContext and not the deprecated OpenAILLMContext.
+
+    Parameters:
+        transform: A function that takes a list of messages and returns a
+            modified list.
+        run_llm: Whether the context update should be sent to the LLM.
+    """
+
+    transform: Callable[[List["LLMContextMessage"]], List["LLMContextMessage"]]
     run_llm: Optional[bool] = None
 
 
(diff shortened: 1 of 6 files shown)


src/pipecat/processors/aggregators/llm_context.py
```diff
@@ -266,6 +266,17 @@ def set_messages(self, messages: List[LLMContextMessage]):
         """
         self._messages[:] = messages
 
+    def transform_messages(
```
kompfner, line 269: @filipi87, in the other closed PR that this PR's changes came from,. you asked:

> I believe maybe we will need to add some kind of lock to the context, or another mechanism to prevent race conditions.
> 
> What do you think ?
> 
> If it makes sense, we can add it in a follow up PR.

I'm not sure what we're doing in this function is much different from a concurrency point of view from what we're doing in `add_messages` or `set_messages` (the methods defined just above here). `transform` is synchronous.

src/pipecat/processors/aggregators/llm_response_universal.py
```diff
@@ -512,6 +524,8 @@ async def process_frame(self, frame: Frame, direction: FrameDirection):
             await self._handle_llm_messages_append(frame)
         elif isinstance(frame, LLMMessagesUpdateFrame):
             await self._handle_llm_messages_update(frame)
+        elif isinstance(frame, LLMMessagesTransformFrame):
```
kompfner, line 527: @filipi87, in the other closed PR that this PR's changes came from, you asked:

> Is there a reason we decided to add this here instead of inside the Assistant Aggregator?
>
> I was having a similar discussion with @aconchillo yesterday, and we concluded that the UserAggregator should only handle user related stuff, such as turn strategies, idle, and mute.
> 
> Given that, this logic feels like it should live inside the Assistant Aggregator instead. I’ve moved all the summarization logic there as well. 😅

`LLMMessagesTransformFrame` is exactly analogous to `LLMMessagesAppendFrame` or `LLMMessagesUpdateFrame`, and so follows the same pattern: we account for it being sent either up or downstream, so we're prepared to "catch" it and handle it in either the user or assistant aggregator.

src/pipecat/frames/frames.py
```diff
@@ -587,6 +587,25 @@ class LLMMessagesUpdateFrame(DataFrame):
...
 
 
+@dataclass
+class LLMMessagesTransformFrame(DataFrame):
+    """Frame containing a transform function to modify the current context's LLM messages.
+
+    A frame containing a transform function that takes the context's current list
+    of LLM messages and returns a modified list.
+
+    Only compatible with LLMContext and not the deprecated OpenAILLMContext.
```
filipi87: > and not the deprecated OpenAILLMContext

I guess we have already removed the deprecated `OpenAILLMContext`, so we probably don't need to mention it here.

src/pipecat/processors/aggregators/llm_context.py
filipi87: Yep, and we probably needed that there as well, since both the `AssistantAggregator` and the `UserAggregator` have access to the same context. 

That is why I was wondering if we need some kind of lock to prevent a race condition.

src/pipecat/processors/aggregators/llm_response_universal.py
filipi87: Makes sense. 👍

filipi87 (approved):
LGTM!

Regarding locking the context before modifying it, we haven’t encountered any issues so far, so we don’t need to include it in this PR. 

But it’s probably something to keep an eye on in case it becomes necessary in the future. 🙂

src/pipecat/processors/aggregators/llm_context.py
kompfner: > Yep, and we probably needed that there as well, since both the AssistantAggregator and the UserAggregator have access to the same context.

But they shouldn't be actually be running in different threads, right? Even if they're running in different asyncio tasks, we wouldn't have a problem with the context-updating operations in this file because they run to completion without ever hitting an `await` (so they should be atomic). Right?

src/pipecat/frames/frames.py
kompfner: Good catch, thanks!

PR #4234, opened 2026-04-03
Export FastAPI app from runner for custom routes

aconchillo:
## Summary

- Export the FastAPI `app` instance at module level from `pipecat.runner.run` so other packages can import it and add custom routes before calling `main()`
- `main()` now configures the existing app with transport-specific routes instead of creating a new one

Example usage:

```python
from pipecat.runner.run import app, main

@app.get("/my-route")
async def my_route():
    return {"hello": "world"}

if __name__ == "__main__":
    main()
```


markbackman (approved):
LGTM!

PR #4241, opened 2026-04-03
Enable async tool cancellation feature.

filipi87:
Enable async tool cancellation feature.

### DIFF
--- a/examples/function-calling/function-calling-anthropic-async-stream.py
+++ b/examples/function-calling/function-calling-anthropic-async-stream.py
@@ -118,6 +118,7 @@ async def run_bot(transport: BaseTransport, runner_args: RunnerArguments):
 
     llm = AnthropicLLMService(
         api_key=os.getenv("ANTHROPIC_API_KEY"),
+        enable_async_tool_cancellation=True,
         settings=AnthropicLLMService.Settings(
             system_instruction=(
                 "You are a helpful assistant in a voice conversation. "
@@ -138,10 +139,6 @@ async def run_bot(transport: BaseTransport, runner_args: RunnerArguments):
         cancel_on_interruption=False,
         timeout_secs=30,
     )
-
-    @llm.event_handler("on_function_calls_started")
-    async def on_function_calls_started(service, function_calls):
-        await tts.queue_frame(TTSSpeakFrame("Sure, tracking your location now."))
 
     location_function = FunctionSchema(
         name="track_current_location",
--- a/examples/function-calling/function-calling-anthropic-async-stream.py
+++ b/examples/function-calling/function-calling-anthropic-async-stream.py
@@ -140,6 +140,11 @@ async def run_bot(transport: BaseTransport, runner_args: RunnerArguments):
         timeout_secs=30,
     )
 
+    @llm.event_handler("on_function_calls_cancelled")
+    async def on_function_calls_cancelled(service, cancelled):
+        for item in cancelled:
+            logger.info(f"Function call cancelled: {item.function_name} [{item.tool_call_id}]")
+
     location_function = FunctionSchema(
         name="track_current_location",
         description="Start tracking the user's current GPS location, reporting position updates until the user reaches their destination.",
--- a/examples/function-calling/function-calling-anthropic-async-stream.py
+++ b/examples/function-calling/function-calling-anthropic-async-stream.py
@@ -141,8 +141,8 @@ async def run_bot(transport: BaseTransport, runner_args: RunnerArguments):
     )
 
     @llm.event_handler("on_function_calls_cancelled")
-    async def on_function_calls_cancelled(service, cancelled):
-        for item in cancelled:
+    async def on_function_calls_cancelled(service, function_calls):
+        for item in function_calls:
             logger.info(f"Function call cancelled: {item.function_name} [{item.tool_call_id}]")
 
     location_function = FunctionSchema(
--- a/src/pipecat/adapters/base_llm_adapter.py
+++ b/src/pipecat/adapters/base_llm_adapter.py
@@ -15,6 +15,7 @@ from typing import Any, Dict, Generic, List, Optional, TypeVar
 
 from loguru import logger
 
+from pipecat.adapters.schemas.function_schema import FunctionSchema
 from pipecat.adapters.schemas.tools_schema import ToolsSchema
 from pipecat.processors.aggregators.llm_context import (
     LLMContext,
@@ -48,6 +49,20 @@ class BaseLLMAdapter(ABC, Generic[TLLMInvocationParams]):
     def __init__(self):
         """Initialize the adapter."""
         self._warned_system_instruction = False
+        self._builtin_tools: List[FunctionSchema] = []
+
+    @property
+    def builtin_tools(self) -> List[FunctionSchema]:
+        """Built-in tools automatically merged into every inference request.
+
+        Mixins (e.g. ``AsyncToolCancellationLLMServiceMixin``) append their
+        tool schemas here so that the tools are injected transparently without
+        the user having to add them to their ``ToolsSchema``.
+
+        Returns:
+            Mutable list of ``FunctionSchema`` instances.
+        """
+        return self._builtin_tools
 
     @property
     @abstractmethod
@@ -122,6 +137,9 @@ class BaseLLMAdapter(ABC, Generic[TLLMInvocationParams]):
     def from_standard_tools(self, tools: Any) -> List[Any] | NotGiven:
         """Convert tools from standard format to provider format.
 
+        Built-in tools are automatically merged into the schema before conversion so that every
+        inference request receives them without the user having to declare them explicitly.
+
         Args:
             tools: Tools in standard format or provider-specific format.
 
@@ -129,8 +147,26 @@ class BaseLLMAdapter(ABC, Generic[TLLMInvocationParams]):
             List of tools converted to provider format, or original tools
             if not in standard format.
         """
+        if self._builtin_tools:
+            if isinstance(tools, ToolsSchema):
+                tools = ToolsSchema(
+                    standard_tools=tools.standard_tools + self._builtin_tools,
+                    custom_tools=tools.custom_tools,
+                )
+            else:
+                # User supplied tools in a legacy/provider-specific format;
+                # we cannot safely merge — build a schema from builtins only.
+                if tools is not None:
+                    logger.warning(
+                        "Built-in tools could not be merged because the supplied tools are not"
+                        " a ToolsSchema instance. Only built-in tools will be sent."
+                    )
+                tools = ToolsSchema(standard_tools=self._builtin_tools)
+
         if isinstance(tools, ToolsSchema):
             logger.debug(f"Retrieving the tools using the adapter: {type(self)}")
+            tool_names = [tool.name for tool in tools.standard_tools]
+            logger.debug(f"Tool names: {tool_names}")
             return self.to_provider_tools_format(tools)
         # Fallback to return the same tools in case they are not in a standard format
         return tools
--- a/src/pipecat/adapters/base_llm_adapter.py
+++ b/src/pipecat/adapters/base_llm_adapter.py
@@ -10,6 +10,7 @@ This module provides the abstract base class for implementing LLM provider-speci
 adapters that handle tool format conversion and standardization.
 """
 
+import warnings
 from abc import ABC, abstractmethod
 from typing import Any, Dict, Generic, List, Optional, TypeVar
 
@@ -49,18 +50,19 @@ class BaseLLMAdapter(ABC, Generic[TLLMInvocationParams]):
     def __init__(self):
         """Initialize the adapter."""
         self._warned_system_instruction = False
-        self._builtin_tools: List[FunctionSchema] = []
+        self._builtin_tools: Dict[str, FunctionSchema] = {}
 
     @property
-    def builtin_tools(self) -> List[FunctionSchema]:
+    def builtin_tools(self) -> Dict[str, FunctionSchema]:
         """Built-in tools automatically merged into every inference request.
 
-        Mixins (e.g. ``AsyncToolCancellationLLMServiceMixin``) append their
-        tool schemas here so that the tools are injected transparently without
-        the user having to add them to their ``ToolsSchema``.
+        Keyed by tool name for O(1) lookup, insertion, and removal.  The
+        service injects tools here so they are sent transparently on every
+        inference request without the user having to add them to their
+        ``ToolsSchema``.
 
         Returns:
-            Mutable list of ``FunctionSchema`` instances.
+            Mutable dict mapping tool name to ``FunctionSchema``.
         """
         return self._builtin_tools
 
@@ -150,23 +152,28 @@ class BaseLLMAdapter(ABC, Generic[TLLMInvocationParams]):
         if self._builtin_tools:
             if isinstance(tools, ToolsSchema):
                 tools = ToolsSchema(
-                    standard_tools=tools.standard_tools + self._builtin_tools,
+                    standard_tools=tools.standard_tools + list(self._builtin_tools.values()),
                     custom_tools=tools.custom_tools,
                 )
             else:
-                # User supplied tools in a legacy/provider-specific format;
-                # we cannot safely merge — build a schema from builtins only.
+                # User supplied tools in a legacy/provider-specific format.
+                # Built-in tools cannot be safely merged, so they will not be injected.
+                # Migrate to ToolsSchema to enable built-in tool support; use custom_tools
+                # as an escape hatch for any provider-specific tools that don't fit the
+                # standard schema.
                 if tools is not None:
-                    logger.warning(
-                        "Built-in tools could not be merged because the supplied tools are not"
-                        " a ToolsSchema instance. Only built-in tools will be sent."
+                    warnings.warn(
+                        "Built-in tools (e.g. async tool cancellation) could not be injected "
+                        "because the supplied tools are not a ToolsSchema instance. "
+                        "Migrate to ToolsSchema to enable built-in tool support. "
+                        "Use ToolsSchema(custom_tools=...) as an escape hatch for any "
+                        "provider-specific tools that don't fit the standard schema.",
+                        DeprecationWarning,
+                        stacklevel=2,
                     )
-                tools = ToolsSchema(standard_tools=self._builtin_tools)
+                # Fall through and return the original tools unchanged.
 
         if isinstance(tools, ToolsSchema):
-            logger.debug(f"Retrieving the tools using the adapter: {type(self)}")
-            tool_names = [tool.name for tool in tools.standard_tools]
-            logger.debug(f"Tool names: {tool_names}")
             return self.to_provider_tools_format(tools)
         # Fallback to return the same tools in case they are not in a standard format
         return tools
--- a/src/pipecat/services/llm_service.py
+++ b/src/pipecat/services/llm_service.py
@@ -195,12 +195,12 @@ class LLMService(UserTurnCompletionLLMServiceMixin, AIService):
             logger.warning("LLM completion timed out")
 
         @task.event_handler("on_function_calls_started")
-        async def on_function_calls_started(service, function_calls):
+        async def on_function_calls_started(service, function_calls: List[FunctionCallFromLLM]):
             logger.info(f"Starting {len(function_calls)} function calls")
 
         @task.event_handler("on_function_calls_cancelled")
-        async def on_function_calls_cancelled(service, cancelled):
-            logger.info(f"Cancelled {len(cancelled)} async function calls")
+        async def on_function_calls_cancelled(service, function_calls: List[FunctionCallFromLLM]):
+            logger.info(f"Cancelled {len(function_calls)} function calls")
     """
 
     _settings: LLMSettings
@@ -987,7 +987,14 @@ class LLMService(UserTurnCompletionLLMServiceMixin, AIService):
                     FunctionCallCancelFrame, function_name=name, tool_call_id=tool_call_id
                 )
 
-                cancelled_items.append(runner_item)
+                cancelled_items.append(
+                    FunctionCallFromLLM(
+                        function_name=runner_item.function_name,
+                        tool_call_id=runner_item.tool_call_id,
+                        arguments=runner_item.arguments,
+                        context=runner_item.context,
+                    )
+                )
                 logger.debug(f"{self} Async function call [{name}:{tool_call_id}] cancelled")
 
         for task in cancelled_tasks:
@@ -1018,7 +1025,14 @@ class LLMService(UserTurnCompletionLLMServiceMixin, AIService):
                     FunctionCallCancelFrame, function_name=name, tool_call_id=tool_call_id
                 )
 
-                cancelled_items.append(runner_item)
+                cancelled_items.append(
+                    FunctionCallFromLLM(
+                        function_name=runner_item.function_name,
+                        tool_call_id=runner_item.tool_call_id,
+                        arguments=runner_item.arguments,
+                        context=runner_item.context,
+                    )
+                )
                 logger.debug(f"{self} Function call [{name}:{tool_call_id}] has been cancelled")
 
         # Remove all cancelled tasks from our set.
--- a/src/pipecat/services/llm_service.py
+++ b/src/pipecat/services/llm_service.py
@@ -16,6 +16,7 @@ from typing import (
     Awaitable,
     Callable,
     Dict,
+    List,
     Mapping,
     Optional,
     Protocol,
@@ -60,6 +61,11 @@ from pipecat.services.ai_service import AIService
 from pipecat.services.settings import LLMSettings
 from pipecat.services.websocket_service import WebsocketService
 from pipecat.turns.user_turn_completion_mixin import UserTurnCompletionLLMServiceMixin
+from pipecat.utils.async_tool_cancellation import (
+    ASYNC_TOOL_CANCELLATION_INSTRUCTIONS,
+    CANCEL_ASYNC_TOOL_NAME,
+    CANCEL_ASYNC_TOOL_SCHEMA,
+)
 from pipecat.utils.context.llm_context_summarization import (
     DEFAULT_SUMMARIZATION_TIMEOUT,
     LLMContextSummarizationUtil,
@@ -230,6 +236,7 @@ class LLMService(UserTurnCompletionLLMServiceMixin, AIService):
         self._group_parallel_tools = group_parallel_tools
         self._function_call_timeout_secs = function_call_timeout_secs
         self._filter_incomplete_user_turns: bool = False
+        self._async_cancellation_enabled: bool = False
         self._base_system_instruction: Optional[str] = None
         self._adapter = self.adapter_class()
         self._functions: Dict[Optional[str], FunctionCallRegistryItem] = {}
@@ -291,6 +298,8 @@ class LLMService(UserTurnCompletionLLMServiceMixin, AIService):
         await super().start(frame)
         if not self._run_in_parallel:
             await self._create_sequential_runner_task()
+        if self._has_async_functions():
+            self._setup_async_tool_cancellation()
 
     async def stop(self, frame: EndFrame):
         """Stop the LLM service.
@@ -315,17 +324,20 @@ class LLMService(UserTurnCompletionLLMServiceMixin, AIService):
         await self._cancel_summary_task()
 
     def _compose_system_instruction(self):
-        """Compose system_instruction by appending turn completion instructions.
+        """Compose system_instruction from the base and all active addon instructions.
 
         Combines the base system instruction with turn completion instructions
-        and writes the result to ``self._settings.system_instruction``.
+        (when enabled) and async tool cancellation instructions (when enabled),
+        writing the result to ``self._settings.system_instruction``.
         """
         base = self._base_system_instruction
-        completion_instructions = self._user_turn_completion_config.completion_instructions
-        if base:
-            self._settings.system_instruction = f"{base}\n\n{completion_instructions}"
-        else:
-            self._settings.system_instruction = completion_instructions
+        parts = [base] if base else []
+        if self._filter_incomplete_user_turns:
+            parts.append(self._user_turn_completion_config.completion_instructions)
+        if self._async_cancellation_enabled:
+            parts.append(ASYNC_TOOL_CANCELLATION_INSTRUCTIONS)
+        composed = "\n\n".join(p for p in parts if p)
+        self._settings.system_instruction = composed or None
 
     async def _update_settings(self, delta: LLMSettings) -> dict[str, Any]:
         """Apply a settings delta, handling turn-completion fields.
@@ -361,10 +373,10 @@ class LLMService(UserTurnCompletionLLMServiceMixin, AIService):
 
         if (
             "system_instruction" in changed
-            and self._filter_incomplete_user_turns
+            and (self._filter_incomplete_user_turns or self._async_cancellation_enabled)
             and "filter_incomplete_user_turns" not in changed
         ):
-            # system_instruction changed while turn completion is active.
+            # system_instruction changed while composition is active.
             # Treat the new value as the new base and recompose.
             self._base_system_instruction = self._settings.system_instruction
             self._compose_system_instruction()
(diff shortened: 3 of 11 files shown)


src/pipecat/processors/aggregators/llm_response_universal.py
```diff
@@ -1061,7 +1061,13 @@ async def _handle_function_call_in_progress(self, frame: FunctionCallInProgressF
             self._context.add_message(
                 {
                     "role": "tool",
-                    "content": json.dumps({"type": "tool", "status": "started"}),
+                    "content": json.dumps(
+                        {
+                            "type": "tool",
+                            "status": "started",
+                            "tool_call_id": frame.tool_call_id,
```
filipi87, line 1068: Adding this was the key to making it work, together with a prompt instructing it to explicitly look at this field.

filipi87:
@aconchillo, @kompfner, @markbackman, should we invoke `on_function_calls_started` when a function call is used to cancel another function call? It feels like we shouldn’t.

JiwaniZakir (commented):
In `base_llm_adapter.py`, the fallback branch in `from_standard_tools` (lines ~155–163) silently drops the user's legacy/provider-specific tools when `_builtin_tools` is non-empty—the original behavior was `return tools` which passed them through unchanged. Replacing them entirely with only built-in tools is a silent breaking change for anyone using a non-`ToolsSchema` format; a better approach would be to still pass the original tools through and emit the warning, rather than discarding them.

In `_compose_system_instruction` within `llm_service.py`, the old implementation unconditionally appended `completion_instructions` (i.e., even when `_filter_incomplete_user_turns` was `False`), whereas the new version gates this behind the flag. This is likely a bug fix, but it's worth verifying there aren't existing users or tests relying on the old unconditional behavior—a regression test covering the "turn completion off, async cancellation on" composition path would help here.

In `llm_response_universal.py`, `tool_call_id` is now duplicated—once inside the JSON `content` payload and once as a top-level message key. If this is intentional (e.g., for a provider that reads it from the content body), a brief comment explaining why would prevent future readers from removing it as redundant.

src/pipecat/services/llm_service.py
```diff
@@ -291,6 +298,8 @@ async def start(self, frame: StartFrame):
         await super().start(frame)
         if not self._run_in_parallel:
             await self._create_sequential_runner_task()
+        if self._has_async_functions():
```
filipi87: We should probably also add a property to the class so users can disable this behavior if they wish.

src/pipecat/adapters/base_llm_adapter.py
```diff
@@ -122,15 +137,36 @@ def get_messages(self, context: LLMContext) -> List[LLMContextMessage]:
...
             if not in standard format.
         """
+        if self._builtin_tools:
+            if isinstance(tools, ToolsSchema):
+                tools = ToolsSchema(
+                    standard_tools=tools.standard_tools + self._builtin_tools,
+                    custom_tools=tools.custom_tools,
+                )
+            else:
+                # User supplied tools in a legacy/provider-specific format;
```
kompfner: Shoot, wondering if we should've already deprecated the ability to use the non-standard format...

src/pipecat/adapters/base_llm_adapter.py
```diff
@@ -122,15 +137,36 @@ def get_messages(self, context: LLMContext) -> List[LLMContextMessage]:
...
+                # we cannot safely merge — build a schema from builtins only.
+                if tools is not None:
+                    logger.warning(
+                        "Built-in tools could not be merged because the supplied tools are not"
+                        " a ToolsSchema instance. Only built-in tools will be sent."
```
kompfner: I don't agree that we should be throwing away the user's provided tools in favor of just our built-in tools in this case...

Maybe we should take this opportunity to finally deprecate legacy formats in favor of `ToolsSchema` (which is pretty much all we test with these days anyway), and say that built-in tools are not supported with legacy formats.

WDYT?

kompfner: In the deprecation warning, in addition to generally pointing readers to `ToolsSchema`, we should make sure to mention that `custom_tools` exists as an escape hatch for any LLM-specific tools that don't fit the standard schema.

src/pipecat/adapters/base_llm_adapter.py
```diff
@@ -122,15 +137,36 @@ def get_messages(self, context: LLMContext) -> List[LLMContextMessage]:
...
+                    )
+                tools = ToolsSchema(standard_tools=self._builtin_tools)
+
         if isinstance(tools, ToolsSchema):
             logger.debug(f"Retrieving the tools using the adapter: {type(self)}")
+            tool_names = [tool.name for tool in tools.standard_tools]
+            logger.debug(f"Tool names: {tool_names}")
```
kompfner: Did you mean to keep this logging in, or was it meant to just be temporary while you worked on these changes?

src/pipecat/adapters/base_llm_adapter.py
kompfner: If you go this route, I think the logic in this function can be simplified to a single `if`:

```python
if isinstance(tools, ToolsSchema):
    tools = ToolsSchema(
        standard_tools=tools.standard_tools + self._builtin_tools, # no-op if _builtin_tools is empty
        custom_tools=tools.custom_tools,
    )
    return self.to_provider_tools_format(tools)
# deprecation warning here
return tools
```

src/pipecat/services/llm_service.py
```diff
@@ -230,6 +236,7 @@ def __init__(
         self._group_parallel_tools = group_parallel_tools
         self._function_call_timeout_secs = function_call_timeout_secs
         self._filter_incomplete_user_turns: bool = False
+        self._async_cancellation_enabled: bool = False
```
kompfner: nit
```suggestion
        self._async_tool_cancellation_enabled: bool = False
```

src/pipecat/services/llm_service.py
kompfner: nit: we seem to switch between referring to them as "async functions" and "async tools". since we've _generally_ been calling them "async tools", maybe we should call this

```suggestion
        if self._has_async_tools():
```

src/pipecat/services/llm_service.py
```diff
@@ -315,17 +324,20 @@ async def cancel(self, frame: CancelFrame):
...
+        writing the result to ``self._settings.system_instruction``.
         """
         base = self._base_system_instruction
-        completion_instructions = self._user_turn_completion_config.completion_instructions
-        if base:
-            self._settings.system_instruction = f"{base}\n\n{completion_instructions}"
-        else:
-            self._settings.system_instruction = completion_instructions
+        parts = [base] if base else []
+        if self._filter_incomplete_user_turns:
```
kompfner, line 349: Looks like we weren't previously checking for this before appending `self._user_turn_completion_config.completion_instructions`...was there a reason for that?

src/pipecat/services/llm_service.py
kompfner: Just wondering...would it ever make sense to make async tool cancellation a per-tool configuration as opposed to an all-or-nothing configuration?

kompfner: Curious why we're running this logic in `start()` as opposed to the end of `init()`? In `init()` we already have access to the service's initial `settings` and everything else I think we need to run this.

src/pipecat/utils/async_tool_cancellation.py
```diff
@@ -0,0 +1,49 @@
...
+
+When an ``LLMService`` has functions registered with
+``cancel_on_interruption=False`` (async tools), it automatically injects the
+``cancel_async_tool_call`` tool and the instructions below into every inference
+request so the LLM can cancel stale in-progress calls.
+"""
+
+from pipecat.adapters.schemas.function_schema import FunctionSchema
+
+CANCEL_ASYNC_TOOL_NAME = "cancel_async_tool_call"
```
kompfner, line 17: Do we want to add a safety check somewhere for the (unlikely, but possible) case where the user provides a tool with this name?

src/pipecat/services/llm_service.py
```diff
@@ -849,6 +861,91 @@ async def timeout_handler():
...
+        logger.debug(f"{self}: Enabling async tool cancellation")
+
+        self._async_cancellation_enabled = True
+
+        if self._base_system_instruction is None:
+            self._base_system_instruction = self._settings.system_instruction
+
+        self._compose_system_instruction()
+
+        if not any(t.name == CANCEL_ASYNC_TOOL_NAME for t in self._adapter.builtin_tools):
```
kompfner: When would `self._adapter.builtin_tools` already contain the cancel tool?

src/pipecat/services/llm_service.py
```diff
@@ -849,6 +861,91 @@ async def timeout_handler():
...
 
+    def _has_async_functions(self) -> bool:
+        """Return True if at least one non-builtin async function is registered."""
+        return any(
+            not item.cancel_on_interruption
+            for name, item in self._functions.items()
+            if name != CANCEL_ASYNC_TOOL_NAME
+        )
+
+    def _setup_async_tool_cancellation(self):
```
kompfner, line 905: Do we ever want to "tear down" async tool cancellation, say when the LLM receives a context with an updated set of tools that no longer contain async tools?

src/pipecat/utils/async_tool_cancellation.py
```diff
@@ -0,0 +1,49 @@
...
+
+ASYNC_TOOL_CANCELLATION_INSTRUCTIONS = """ASYNC TOOL CANCELLATION:
+Some tool calls run asynchronously in the background. When one starts, a tool response \
+is added to the conversation whose content is a JSON object with \
+"type": "tool", "status": "started", and a "tool_call_id" field containing the \
```
kompfner: Slightly out of date. Off the top of my head:
- "type" is now "async_tool"
- "status" is now "running"
- we now have language more like "some tool calls kick of tasks that run asynchronously in the background"
- (maybe more)

src/pipecat/utils/async_tool_cancellation.py
```diff
@@ -0,0 +1,49 @@
...
+result would clearly be stale, call cancel_async_tool_call. \
+To find the correct tool_call_id: locate the most recent tool response in the conversation \
+whose content has "status": "started" and whose call has NOT already been cancelled, \
+then copy the "tool_call_id" value from that content exactly as-is. \
+Never invent or guess a tool_call_id."""
+
+CANCEL_ASYNC_TOOL_SCHEMA = FunctionSchema(
+    name=CANCEL_ASYNC_TOOL_NAME,
+    description=(
+        "Cancel a single async tool call that is no longer needed. "
```
kompfner:
```suggestion
        "Cancel a single async tool call whose results are no longer needed. "
```

src/pipecat/utils/async_tool_cancellation.py
```diff
@@ -0,0 +1,49 @@
...
+        "Use this when the user changes topic, indicates a pending result is "
+        "no longer relevant, or when processing the result would produce a "
+        "stale or confusing response. "
+        "The tool_call_id must be the exact 'id' value from the assistant's "
```
kompfner: Now that "tool_call_id" is explicitly in the initial result we can probably clear up this language

src/pipecat/services/llm_service.py
filipi87: I think we can start with an all-or-nothing approach and make it a per-tool cancellation if needed.

src/pipecat/adapters/base_llm_adapter.py
filipi87: Probably yes. But for this PR, I just realized I am being too aggressive, we should not include the built-in tools in this case.

src/pipecat/adapters/base_llm_adapter.py
filipi87: Yep, I think this is the right way to move forward. Nice catch!

src/pipecat/adapters/base_llm_adapter.py
filipi87: Only for testing. Need to remove it.

src/pipecat/services/llm_service.py
filipi87: Before, `_compose_system_instruction` was only invoked inside `_update_settings`, and we had this check there. 

But now, since we also invoke it to inject the async tool cancellation instructions, we need to add this check here as well.

src/pipecat/services/llm_service.py
filipi87: The functions are registered when `register_function` is invoked, which happens after initialization. 

And we need access to the functions inside `self._functions` to determine whether any are async and inject what’s needed.

That is why I am doing it here inside `start`, at this point all the functions must be registered.

src/pipecat/utils/async_tool_cancellation.py
filipi87: We can do that. 👍

src/pipecat/services/llm_service.py
filipi87: I think you’re right, this case should never happen unless the user has provided a function with the same name. 

In that case, we shouldn’t change the prompt either. I’ll improve this.

src/pipecat/services/llm_service.py
kompfner: Is the thinking that we'll add this param in a follow-up PR?

kompfner: > The functions are registered when register_function is invoked

Ah right, of course! Brain fart.

src/pipecat/services/llm_service.py
filipi87: I believe the right way to do this would be to validate, when `unregister_function` is invoked, whether any async tools remain, and if not, then we should remove that. 

What do you think ?

src/pipecat/services/llm_service.py
kompfner: Ah makes sense!

src/pipecat/utils/async_tool_cancellation.py
filipi87: Yeah, nice catch. I rebased this branch early on the latest changes, but missed this.

src/pipecat/services/llm_service.py
filipi87: I was planning to add this parameter in this PR so we can keep it disabled by default, preserving the current behavior unless the user specifies otherwise.

src/pipecat/services/llm_service.py
kompfner: > unless the user has provided a function with the same name

Actually, even if they did, it wouldn't end up in `self._adapter.builtin_tools`...

But maybe it doesn't hurt to check for it here, especially if, as [suggested in another PR comment](https://github.com/pipecat-ai/pipecat/pull/4241/changes/BASE..772fb57090c8aa6eb4c9e07511e5c94cba0a2efb#r3059130030), we end up registering and unregistering the cancellation tool from `builtin_tools`.

Actually...if we _do_ end up implementing that suggestion and thus develop the need to unregister tools from `builtin_tools`, we'd be better served with a different data structure for `builtin_tools`, like a dictionary keyed by tool name. Then we wouldn't need this check at all.

src/pipecat/services/llm_service.py
kompfner: > disabled by default

Oh, huh. I think I was imagining that it would be _enabled_ by default (but honestly, hadn't thought about it too much). The nice thing is we haven't shipped async tool support yet, so there's no "shipped behavior" to consider and we can make this decision before we ship any of it.

What was your reasoning for having it disabled by default? Are LLMs already pretty good at simply ignoring tool results that come back after they're no longer relevant, i.e. after the conversation has already moved onto another topic and it's clear that the user is no longer interested in the result?

I guess a simple test would be "get me the weather in san francisco...oh wait, sorry, i meant san diego".

kompfner: > at simply ignoring tool results that come back after they're no longer relevant

Hmm, actually, thinking about this a bit more, our architecture doesn't really always let the LLM make that judgment call...when the result comes back it will trigger an LLM response.

filipi87: The idea was to avoid injecting anything for the user by default, unless they explicitly ask for it.

However, LLMs are not yet good at ignoring function call results, at least in my tests. 🤕

kompfner: > LLMs are not yet good at ignoring function call results, at least in my tests

We've actually set it up so they *can't*, I think—we ask them to run inference with those results as the latest message(s).

I wonder if the way to get them to reliably ignore irrelevant function call results is with some prompting akin to what @markbackman did with `USER_TURN_COMPLETION_INSTRUCTIONS`: asking the LLM to disregard "async_tool" results that are no longer relevant to the conversation, including in the case where the result is the only message after the last assistant message (in which case we could ask the LLM to output a symbol that will be ignored, essentially making it "skip" responding).

src/pipecat/services/llm_service.py
kompfner: Oh, hmm...not sure. `unregister_function` is pretty optional, in terms of telling the LLM which set of tools is available to it...the user could choose to keep all tools "registered" with the Pipecat service but at runtime set `LLMSetToolsFrame`s to change which tools are available for each inference.

Kind of makes me think we probably want to reconsider when/where we run the logic that determines if we have async tools and sets up the tool cancellation mechanism.

kompfner:
I think we'll really want an example file to demonstrate where this behavior would be useful, with some helpful commenting explaining how the user might exercise it.

src/pipecat/services/llm_service.py
filipi87: I have added a new property for now: `enable_async_tool_cancellation`

src/pipecat/adapters/base_llm_adapter.py
filipi87: Fixed.

src/pipecat/adapters/base_llm_adapter.py
filipi87: Fixed.

src/pipecat/services/llm_service.py
filipi87: Fixed.

src/pipecat/utils/async_tool_cancellation.py
filipi87: Done!

src/pipecat/services/llm_service.py
filipi87: Yep. I have refactored that. 👍

src/pipecat/utils/async_tool_cancellation.py
filipi87: Fixed.

kompfner:
Thoughts after talking with @filipi87...

It seems like async tool cancellation is sort of being designed to be two things:

1. A way to disregard stale results that come in after they're no longer needed in the convo
2. As a task control function, stopping long-running "subagent-like" or "process-like" async tools that might have some external side effects and may be sending a stream of results into the conversation

Re 1: I'm not sure the cancellation mechanism as it exists in this PR will excel at this without some modification, as tool calls are "expensive"—requiring a back and forth—and the LLM will respond to cancellation tool call results coming in (the latter point which maybe we can address with additional prompt engineering in the cancellation tool call result message). It's possible an entirely different approach like [this one](https://github.com/pipecat-ai/pipecat/pull/4241#discussion_r3060225540) or something we're not yet considering, would be better suited to the task.

Re 2: I feel a smidge uneasy using a hidden tool cancellation mechanism as a control interface for long-running subagent-like tasks - if the developer is adding something like "start_location_tracking", it would be more natural for them to provide a "stop_location_tracking" (and possibly even an "update_location_tracking", if they so desire). This then makes me wonder if these sorts of tasks are better suited for the WIP subagents and their corresponding control systems...which then makes me wonder whether we're making a mistake by bundling up long-running, streamed-response agent-like things into our conception of "tools" in the first place.

Maybe I'm overthinking things and spiraling out...but I think it's worth a bit of more discussion, as we really want to get this right since it's a pretty foundational pattern. Thanks for bearing with me!

cc @markbackman @aconchillo

kompfner:
Re 1: talked to @filipi87 and he'll try addressing the issue of

> the LLM will respond to cancellation tool call results coming in

🤞 hopefully that'll be enough to make things work fairly well for both the disregard-result case as well as the cancel-long-running-task case (we talked about the validation scenarios that make sense for this).

Re 2: after talking to @aconchillo and @filipi87, my existential unease about longer-running streamed-result tools have been assuaged. I'm back on board.

> Maybe I'm overthinking things and spiraling out...

Yes, probably! For those who know me: very much classic me.

> but I think it's worth a bit of more discussion

I think it was worth it 👍.

examples/function-calling/function-calling-anthropic-async-stream.py
```diff
@@ -139,10 +140,6 @@ async def run_bot(transport: BaseTransport, runner_args: RunnerArguments):
         timeout_secs=30,
     )
 
-    @llm.event_handler("on_function_calls_started")
```
aconchillo, line 142: It would be great to add an `on_function_call_cancelled` event.

aconchillo:
This looks great to me @filipi87 ! My only concern would be the cancellation prompt. Worst case, if the prompt doesn't work for some use case/LLM, users are able to modify the module variables.

examples/function-calling/function-calling-anthropic-async-stream.py
filipi87: Yep. Done.

filipi87:
> Worst case, if the prompt doesn't work for some use case/LLM, users are able to modify the module variables.

Or they can simply leave the mechanism disabled and handle it on their own.

filipi87:
@kompfner , it looks like we were already respecting when the run_llm is False:
- https://github.com/pipecat-ai/pipecat/blob/891f00cb5fe79eb3ab5ad074df7a154f45927d5c/src/pipecat/processors/aggregators/llm_response_universal.py#L1122-L1127

filipi87:
@kompfner, it looks like we need to keep `run_llm` set to `True`, otherwise, the LLM doesn’t reply. With it set to `True`, it seems to be working well.

For example:
```
[
   {
      "role":"assistant",
      "content":"How can I help you today?"
   },
   {
      "role":"user",
      "content":"Can you tell me the current weather in Florianópolis?"
   },
   {
      "role":"assistant",
      "tool_calls":[
         {
            "id":"call_RsG3xBDlDIgxhNz57aNxvAbY",
            "function":{
               "name":"get_current_weather",
               "arguments":"{\"location\": \"Florianópolis, SC\", \"format\": \"celsius\"}"
            },
            "type":"function"
         }
      ]
   },
   {
      "role":"tool",
      "content":"{\"type\": \"async_tool\", \"status\": \"running\", \"tool_call_id\": \"call_RsG3xBDlDIgxhNz57aNxvAbY\", \"description\": \"An asynchronous task associated with this tool_call_id has started running. Expect results to arrive later as developer messages that look roughly like this one (with \\'type=async_tool\\' and a matching tool_call_id) but with a \\'result\\' field. Note that there *may* be more than one result (i.e., a stream of results), but there doesn\\'t have to be (there may be only one). The last result will come in a message with \\'status=finished\\'.\"}",
      "tool_call_id":"call_RsG3xBDlDIgxhNz57aNxvAbY"
   },
   {
      "role":"assistant",
      "content":"Let me check on that."
   },
   {
      "role":"user",
      "content":"Actually, would you mind? Tell me the weather in Curitiba instead."
   },
   {
      "role":"assistant",
      "tool_calls":[
         {
            "id":"call_zduKgNDnbjxqbk9pBWeVPJom",
            "function":{
               "name":"cancel_async_tool_call",
               "arguments":"{\"tool_call_id\": \"call_RsG3xBDlDIgxhNz57aNxvAbY\"}"
            },
            "type":"function"
         }
      ]
   },
   {
      "role":"tool",
      "content":"{\"cancelled\": \"call_RsG3xBDlDIgxhNz57aNxvAbY\"}",
      "tool_call_id":"call_zduKgNDnbjxqbk9pBWeVPJom"
   },
   {
      "role":"assistant",
      "tool_calls":[
         {
            "id":"call_x5TCeVP8aUgKADUvabYvQXFq",
            "function":{
               "name":"get_current_weather",
               "arguments":"{\"location\": \"Curitiba, PR\", \"format\": \"celsius\"}"
            },
            "type":"function"
         }
      ]
   },
   {
      "role":"tool",
      "content":"{\"type\": \"async_tool\", \"status\": \"running\", \"tool_call_id\": \"call_x5TCeVP8aUgKADUvabYvQXFq\", \"description\": \"An asynchronous task associated with this tool_call_id has started running. Expect results to arrive later as developer messages that look roughly like this one (with \\'type=async_tool\\' and a matching tool_call_id) but with a \\'result\\' field. Note that there *may* be more than one result (i.e., a stream of results), but there doesn\\'t have to be (there may be only one). The last result will come in a message with \\'status=finished\\'.\"}",
      "tool_call_id":"call_x5TCeVP8aUgKADUvabYvQXFq"
   },
   {
      "role":"developer",
      "content":"{\"type\": \"async_tool\", \"tool_call_id\": \"call_x5TCeVP8aUgKADUvabYvQXFq\", \"status\": \"finished\", \"description\": \"This is the final result for the asynchronous task associated with this tool_call_id. The task has completed. No further results will arrive for this tool_call_id.\", \"result\": \"{\\\"conditions\\\": \\\"nice\\\", \\\"temperature\\\": \\\"75\\\"}\"}"
   }
]
```

Screenshot 2026-04-10 at 08 15 00

filipi87:
And this is an example with multiple stream responses:

```
[
   {
      "role":"developer",
      "content":"Please introduce yourself to the user."
   },
   {
      "role":"assistant",
      "content":"Hello! I’m your helpful assistant. How can I assist you today?"
   },
   {
      "role":"user",
      "content":"Could you track my current location?"
   },
   {
      "role":"assistant",
      "content":"Sure, tracking your location now."
   },
   {
      "role":"assistant",
      "tool_calls":[
         {
            "id":"call_J4LO1WCW2Re1ow99rxb9ya3H",
            "function":{
               "name":"track_current_location",
               "arguments":"{}"
            },
            "type":"function"
         }
      ]
   },
   {
      "role":"tool",
      "content":"{\"type\": \"async_tool\", \"status\": \"running\", \"tool_call_id\": \"call_J4LO1WCW2Re1ow99rxb9ya3H\", \"description\": \"An asynchronous task associated with this tool_call_id has started running. Expect results to arrive later as developer messages that look roughly like this one (with \\'type=async_tool\\' and a matching tool_call_id) but with a \\'result\\' field. Note that there *may* be more than one result (i.e., a stream of results), but there doesn\\'t have to be (there may be only one). The last result will come in a message with \\'status=finished\\'.\"}",
      "tool_call_id":"call_J4LO1WCW2Re1ow99rxb9ya3H"
   },
   {
      "role":"developer",
      "content":"{\"type\": \"async_tool\", \"tool_call_id\": \"call_J4LO1WCW2Re1ow99rxb9ya3H\", \"status\": \"running\", \"description\": \"This is an intermediate result for the asynchronous task associated with this tool_call_id. The task is still running. More intermediate results may follow, or the next result may be the final one with \\'status=finished\\'.\", \"result\": \"{\\\"gps\\\": {\\\"lat\\\": 37.731, \\\"lng\\\": -122.4527}, \\\"city\\\": \\\"San Francisco\\\"}\"}"
   },
   {
      "role":"assistant",
      "content":"I’ve started tracking your current location. You’re currently in San Francisco. Let me know if you"
   },
   {
      "role":"developer",
      "content":"{\"type\": \"async_tool\", \"tool_call_id\": \"call_J4LO1WCW2Re1ow99rxb9ya3H\", \"status\": \"running\", \"description\": \"This is an intermediate result for the asynchronous task associated with this tool_call_id. The task is still running. More intermediate results may follow, or the next result may be the final one with \\'status=finished\\'.\", \"result\": \"{\\\"gps\\\": {\\\"lat\\\": 33.96003, \\\"lng\\\": -118.40639}, \\\"city\\\": \\\"Los Angeles\\\"}\"}"
   },
   {
      "role":"user",
      "content":"Actually, we can cancel that now."
   },
   {
      "role":"assistant",
      "tool_calls":[
         {
            "id":"call_ARVAj8QVyoSIhtdugrgtKcHA",
            "function":{
               "name":"cancel_async_tool_call",
               "arguments":"{\"tool_call_id\": \"call_J4LO1WCW2Re1ow99rxb9ya3H\"}"
            },
            "type":"function"
         }
      ]
   },
   {
      "role":"tool",
      "content":"{\"cancelled\": \"call_J4LO1WCW2Re1ow99rxb9ya3H\"}",
      "tool_call_id":"call_ARVAj8QVyoSIhtdugrgtKcHA"
   },
   {
      "role":"assistant",
      "content":"Location tracking has been cancelled. If you need help with anything else, just let me know!"
   },
   {
      "role":"user",
      "content":"Okay. Thank you."
   }
]
```

Screenshot 2026-04-10 at 08 28 29

filipi87:
And this one, is one test with multiple function calls:

```
[
   {
      "role":"developer",
      "content":"Please introduce yourself to the user."
   },
   {
      "role":"assistant",
      "content":"Hello! I'm your helpful voice assistant. How can I assist you today?"
   },
   {
      "role":"user",
      "content":"Could you tell me how is the weather in"
   },
   {
      "role":"user",
      "content":"San Diego?"
   },
   {
      "role":"user",
      "content":"And track my current location?"
   },
   {
      "role":"assistant",
      "content":"Sure, tracking your location now."
   },
   {
      "role":"assistant",
      "tool_calls":[
         {
            "id":"call_MWF6W3GyUb7rqEq0rTehNsEQ",
            "function":{
               "name":"get_current_weather",
               "arguments":"{\"location\": \"San Diego, CA\", \"format\": \"fahrenheit\"}"
            },
            "type":"function"
         }
      ]
   },
   {
      "role":"tool",
      "content":"{\"type\": \"async_tool\", \"status\": \"running\", \"tool_call_id\": \"call_MWF6W3GyUb7rqEq0rTehNsEQ\", \"description\": \"An asynchronous task associated with this tool_call_id has started running. Expect results to arrive later as developer messages that look roughly like this one (with \\'type=async_tool\\' and a matching tool_call_id) but with a \\'result\\' field. Note that there *may* be more than one result (i.e., a stream of results), but there doesn\\'t have to be (there may be only one). The last result will come in a message with \\'status=finished\\'.\"}",
      "tool_call_id":"call_MWF6W3GyUb7rqEq0rTehNsEQ"
   },
   {
      "role":"assistant",
      "tool_calls":[
         {
            "id":"call_kqGHoNfdvpJBnY6cUu0rFMgH",
            "function":{
               "name":"track_current_location",
               "arguments":"{}"
            },
            "type":"function"
         }
      ]
   },
   {
      "role":"tool",
      "content":"{\"type\": \"async_tool\", \"status\": \"running\", \"tool_call_id\": \"call_kqGHoNfdvpJBnY6cUu0rFMgH\", \"description\": \"An asynchronous task associated with this tool_call_id has started running. Expect results to arrive later as developer messages that look roughly like this one (with \\'type=async_tool\\' and a matching tool_call_id) but with a \\'result\\' field. Note that there *may* be more than one result (i.e., a stream of results), but there doesn\\'t have to be (there may be only one). The last result will come in a message with \\'status=finished\\'.\"}",
      "tool_call_id":"call_kqGHoNfdvpJBnY6cUu0rFMgH"
   },
   {
      "role":"developer",
      "content":"{\"type\": \"async_tool\", \"tool_call_id\": \"call_kqGHoNfdvpJBnY6cUu0rFMgH\", \"status\": \"running\", \"description\": \"This is an intermediate result for the asynchronous task associated with this tool_call_id. The task is still running. More intermediate results may follow, or the next result may be the final one with \\'status=finished\\'.\", \"result\": \"{\\\"gps\\\": {\\\"lat\\\": 37.731, \\\"lng\\\": -122.4527}, \\\"city\\\": \\\"San Francisco\\\"}\"}"
   },
   {
      "role":"assistant",
      "content":"I'm checking the weather in San Diego for you. Meanwhile, your current GPS coordinates place you in San Francisco."
   },
   {
      "role":"developer",
      "content":"{\"type\": \"async_tool\", \"tool_call_id\": \"call_kqGHoNfdvpJBnY6cUu0rFMgH\", \"status\": \"running\", \"description\": \"This is an intermediate result for the asynchronous task associated with this tool_call_id. The task is still running. More intermediate results may follow, or the next result may be the final one with \\'status=finished\\'.\", \"result\": \"{\\\"gps\\\": {\\\"lat\\\": 33.96003, \\\"lng\\\": -118.40639}, \\\"city\\\": \\\"Los Angeles\\\"}\"}"
   },
   {
      "role":"user",
      "content":"Okay. We can cancel tracking my location."
   },
   {
      "role":"assistant",
      "tool_calls":[
         {
            "id":"call_ezcvw1BvgSrK2bxspZV0EIwc",
            "function":{
               "name":"cancel_async_tool_call",
               "arguments":"{\"tool_call_id\": \"call_kqGHoNfdvpJBnY6cUu0rFMgH\"}"
            },
            "type":"function"
         }
      ]
   },
   {
      "role":"tool",
      "content":"{\"cancelled\": \"call_kqGHoNfdvpJBnY6cUu0rFMgH\"}",
      "tool_call_id":"call_ezcvw1BvgSrK2bxspZV0EIwc"
   },
   {
      "role":"assistant",
      "content":"Location tracking has been cancelled. If you need to start it again or have any other requests, just let me know!"
   },
   {
      "role":"developer",
      "content":"{\"type\": \"async_tool\", \"tool_call_id\": \"call_MWF6W3GyUb7rqEq0rTehNsEQ\", \"status\": \"finished\", \"description\": \"This is the final result for the asynchronous task associated with this tool_call_id. The task has completed. No further results will arrive for this tool_call_id.\", \"result\": \"{\\\"conditions\\\": \\\"nice\\\", \\\"temperature\\\": \\\"75\\\"}\"}"
   },
   {
      "role":"assistant",
      "content":"The weather in San Diego is nice, with a temperature around 75 degrees Fahrenheit. If you need anything else, feel free to ask!"
   },
   {
      "role":"user",
      "content":"Thank you."
   }
]
```

Screenshot 2026-04-10 at 08 47 53

src/pipecat/utils/async_tool_cancellation.py
kompfner: I agree with your decision to do the simple thing and disallow usage of that name even if no async tools are currently registered—they _may_ be registered at any point, and going from allowing a name to disallowing one would be confusing.

src/pipecat/services/llm_service.py
```diff
@@ -849,8 +894,111 @@ async def timeout_handler():
...
+                )
+
+                cancelled_items.append(runner_item)
+                logger.debug(f"{self} Async function call [{name}:{tool_call_id}] cancelled")
+
+        for task in cancelled_tasks:
+            self._function_call_task_finished(task)
+
+        if cancelled_items:
+            await self._call_event_handler("on_function_calls_cancelled", cancelled_items)
```
aconchillo, line 1004: Shouldn't there be a single tool call that matches `tool_call_id`? So, this should be `on_function_call_cancelled` (singular) and we don't need to keep track of a list, I think.

src/pipecat/services/llm_service.py
kompfner: > we probably want to reconsider when/where we run the logic that determines if we have async tools and sets up the tool cancellation mechanism

Exploring what that would look like, in case it would be too gnarly...

We would have the "hook in" point for setting up the async tool cancellation mechanism be right before running inference, once we've resolved the actual set of tools available to the LLM for that inference run (not the set of tools `register_function`ed with the Pipecat LLM service). So...completely internal to the adapter `get_llm_invocation_params` function.

Advantages of this approach:
- It only adds the async tool cancellation mechanism if we really need it (in the case of more tools being `register`ed than are being used)
- Correctly omits the tool cancellation mechanism if the user doesn't `unregister_function`

Disadvantage of this approach:
- There are some LLMs where `tools` can also be specified directly in the init, and the resolution of the "active set" of tools happens *after* `get_llm_invocation_params`.
  - Though...I'm not sure async tools even makes sense for realtime LLMs, since we don't really have a good way for all of them to report async result updates mid-conversation...maybe we should
- It would require an update to all the `get_llm_invocation_params` implementations in the adapters
- It _might_ be beneficial to the LLM to have the async-tool-cancellation-related system instructions in place continually throughout the conversation...

Taking a step back: the reason to change your approach would be to make the cancellation mechanism's lifespan be "tighter" around when async tools are actually available to the LLM. But it doesn't _hurt_ to have the mechanism in place for longer than that. So...maybe let's do nothing.

kompfner:
Thanks for sharing the outcome of running the examples, @filipi87!

> it looks like we need to keep run_llm set to True, otherwise, the LLM doesn’t reply

I see: if we hadn't set `run_llm=True`, the assistant wouldn't have fired off the next tool call to get the weather for the right place.

Our of curiosity: what was the assistant's response at the end of the ignore-my-first-weather-request example? I assume the assistant did the right thing and didn't feel the need to mention canceling the first request.

kompfner (approved):

kompfner (approved):
Thanks for bearing with all my comments! This is a tricky feature! Nice work.

src/pipecat/services/llm_service.py
filipi87: I created `on_function_calls_cancelled` to match the behavior of `on_function_calls_started`, and this callback is invoked even when there are no async function calls.

So, for example, if you have two function calls in progress (`non-async`) and you interrupt the bot, both will be cancelled.

filipi87:
> Our of curiosity: what was the assistant's response at the end of the ignore-my-first-weather-request example? I assume the assistant did the right thing and didn't feel the need to mention canceling the first request.

Exactly. In that case, after we provided the result of the cancelled function, the assistant simply invoked `get_weather` again, but this time for Curitiba. It didn’t mention it had cancelled the `get_weather` for `Florianopolis`.

src/pipecat/services/llm_service.py
```diff
@@ -188,6 +197,10 @@ async def on_completion_timeout(service):
         @task.event_handler("on_function_calls_started")
         async def on_function_calls_started(service, function_calls):
             logger.info(f"Starting {len(function_calls)} function calls")
+
+        @task.event_handler("on_function_calls_cancelled")
+        async def on_function_calls_cancelled(service, cancelled):
```
aconchillo: Update to:
```
 async def on_function_calls_cancelled(service, function_calls: List[FunctionCallFromLLM]):
```

same in the started case.

filipi87: Done.

aconchillo (approved):

aconchillo:
LGTM!

PR #4242, opened 2026-04-03
Fix Gemini Live session resumption hanging after reconnect

kompfner:
## Summary

- Fixed Gemini Live bot hanging after session resumption reconnect
- After reconnect, `_ready_for_realtime_input` was never set back to `True`, causing all audio/video/text input to be silently dropped
- Root cause: `_disconnect()` resets `_ready_for_realtime_input = False`, but on reconnect nobody set it back — `_create_initial_response()` (which normally sets the flag) is only called on initial connection
- Fix: in `_handle_session_ready`, detect reconnects via `_session_resumption_handle` (primary) or existing `_context` (fallback), and set the flag directly

## Testing

- Tested with simulated disconnect across multiple Pipecat versions (v0.0.96 through main) to bisect the regression
- Confirmed bug reproduces on v0.0.108+ (when `_ready_for_realtime_input` was introduced)
- Confirmed fix works with both session-resumption and non-resumption reconnect paths
- Verified initial connection (non-reconnect) still works correctly

### DIFF
--- a/src/pipecat/services/google/gemini_live/llm.py
+++ b/src/pipecat/services/google/gemini_live/llm.py
@@ -1355,14 +1355,28 @@ class GeminiLiveLLMService(LLMService):
     async def _handle_session_ready(self, session: AsyncSession):
         """Handle the session being ready."""
         self._session = session
-        # If we were just waititng for the session to be ready to run the LLM,
-        # do that now.
         if self._run_llm_when_session_ready:
+            # Initial connection: context arrived before session was ready.
             self._run_llm_when_session_ready = False
             await self._create_initial_response()
-        elif self._context:
-            # Reconnect case: context already exists, no need for _create_initial_response
+        elif self._session_resumption_handle:
+            # Reconnect with session resumption: the server will restore
+            # session state, so we can accept realtime input right away.
             self._ready_for_realtime_input = True
+        elif self._context:
+            # Reconnect without session resumption (e.g. error occurred
+            # before server sent a resumption handle).
+            # TODO: ideally we'd re-send conversation history here via
+            # _create_initial_response(), but that currently doesn't handle
+            # the reconnect case properly. This should be very rare — the
+            # connection would have to drop before we've received our first
+            # session_resumption_handle from the server.
+            self._ready_for_realtime_input = True
+        else:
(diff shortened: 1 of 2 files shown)


src/pipecat/services/google/gemini_live/llm.py
```diff
@@ -1360,6 +1360,9 @@ async def _handle_session_ready(self, session: AsyncSession):
         if self._run_llm_when_session_ready:
             self._run_llm_when_session_ready = False
             await self._create_initial_response()
+        elif self._context:
```
kompfner, line 1366: gonna use a different signal here instead...

kompfner: i think i'll check for `self._session_resumption_handle`, which should be a reliable signal that the conversation has already been started

kompfner: Hm, that's not quite right either. It's possible for a reconnect to occur before we get our first session resumption handle (not likely, but possible). Working on it.

src/pipecat/services/google/gemini_live/llm.py
```diff
@@ -1355,11 +1355,28 @@ async def _tool_result(
...
+            self._ready_for_realtime_input = True
+        elif self._context:
+            # Reconnect without session resumption (e.g. error occurred
+            # before server sent a resumption handle).
+            # TODO: ideally we'd re-send conversation history here via
+            # _create_initial_response(), but that currently doesn't handle
+            # the reconnect case properly. This should be very rare — the
+            # connection would have to drop before we've received our first
+            # session_resumption_handle from the server.
+            self._ready_for_realtime_input = True
```
filipi87, line 1374: Just to confirm, there’s nothing else we need to do here, just set the same flag as above?

filipi87 (approved):
LGTM!

src/pipecat/services/google/gemini_live/llm.py
kompfner: For now - this scenario isn't quite working today either, so I left a TODO to fix it properly (re-seeding convo history) as a separate task. But leaving it out of scope for this PR.

PR #4244, opened 2026-04-06
fix VAD stuck in SPEAKING state when audio stops mid-speech

### LINKED ISSUE #4124: VAD stuck in SPEAKING state when audio input stops (mute mid-speech) — no turn stop or idle trigger
### pipecat version

0.0.107

### Python version

3.12

### Operating System

ALL

### Issue description

When a user mutes their microphone mid-speech (or audio frames stop arriving), Pipecat’s pipeline gets stuck in an inconsistent state:

VAD remains in SPEAKING
VADUserStoppedSpeakingFrame is never emitted
STT never finalizes transcription
User turn never ends
Idle detection does not trigger
System appears frozen until user unmutes

This happens when using the recommended architecture (VAD on LLMUserAggregator, not transport).

### Reproduction steps

Setup
 
- Use Pipecat with:
- WebSocket transport (e.g. FastAPIWebsocketTransport)
- VAD on LLMUserAggregator (default / recommended setup)
- No vad_analyzer on transport
- Ensure STT + pipeline is working normally (speech → transcript → response)

Steps to Reproduce

- Start a voice session
- Begin speaking (e.g. say: “How are you…”)
- While still speaking, mute the microphone (or stop sending audio frames from client)
- Wait for a few seconds (≥ 5s)

### Expected behavior

When audio input stops unexpectedly (mute, network drop, etc.):

VAD should eventually transition to QUIET
A VADUserStoppedSpeakingFrame should be emitted
User turn should close
STT should finalize transcription
Idle timer should start

### Actual behavior

If audio frames stop mid-speech:

VAD stays in SPEAKING indefinitely
No stop frame is emitted
UserTurnController remains stuck (_user_speaking=True)
UserIdleController is blocked (_user_turn_in_progress=True)
No recovery until new audio arrives

### Logs

```shell

```

omChauhanDev:

Fixes: #4124 

#### VAD gets stuck in `SPEAKING` state when audio frames stop arriving mid-speech (e.g. user mutes mic)

**Cause**  : `VADAnalyzer` only transitions states when it receives audio frames to analyze. If frames stop coming while in `SPEAKING` state, there's nothing to trigger the transition back to `QUIET`, so everything downstream (turn management, LLM pipeline) stays frozen waiting for speech to end.

There was actually a starvation guard for this exact scenario, added by @filipi87 inside `BaseInputTransport`. It worked well. But when the deprecated transport-level VAD was cleaned up as part of the architecture migration to `VADController`, that guard naturally went with it. 

This pr adds that back : 
A lightweight starvation watchdog inside `VADController` that detects the audio gap and forces a clean `SPEAKING → QUIET` transition after a configurable timeout (default 1s, disable with 0). The pattern mirrors what `UserTurnController` already does for its turn stop timeout : asyncio.Event + wait_for.

### DIFF
--- a/src/pipecat/audio/vad/vad_controller.py
+++ b/src/pipecat/audio/vad/vad_controller.py
@@ -39,8 +39,8 @@ class VADController(BaseObject):
     Event handlers available:
 
     - on_speech_started: Called when speech begins.
-    - on_speech_stopped: Called when speech ends, including forced stop on
-      audio starvation (no frames received while speaking).
+    - on_speech_stopped: Called when speech ends, including forced stop when
+      the audio stream goes idle (no frames received while speaking).
     - on_speech_activity: Called periodically while speech is detected.
     - on_push_frame: Called when the controller wants to push a frame.
     - on_broadcast_frame: Called when the controller wants to broadcast a frame.
@@ -73,7 +73,7 @@ class VADController(BaseObject):
         vad_analyzer: VADAnalyzer,
         *,
         speech_activity_period: float = 0.2,
-        audio_starvation_timeout: float = 1.0,
+        audio_idle_timeout: float = 1.0,
     ):
         """Initialize the VAD controller.
 
@@ -81,7 +81,7 @@ class VADController(BaseObject):
             vad_analyzer: The `VADAnalyzer` instance for processing audio.
             speech_activity_period: Minimum interval in seconds between
                 `on_speech_activity` events. Defaults to 0.2.
-            audio_starvation_timeout: Timeout in seconds to force speech stop
+            audio_idle_timeout: Timeout in seconds to force speech stop
                 when no audio frames are received while in SPEAKING state.
                 This handles cases like mic mute mid-speech.
                 Set to 0 to disable. Defaults to 1.0.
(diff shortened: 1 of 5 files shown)


src/pipecat/audio/vad/vad_controller.py
```diff
@@ -128,6 +168,8 @@ async def _handle_audio(self, frame: InputAudioRawFrame):
         Args:
             frame: Audio frame to process.
         """
+        self._audio_received_event.set()
```
aconchillo: This is too CPU intensive. It wakes up the starvation handler too often. A simple solution is to simply store a timestamp here and then in the starvation handler do:

```
while True:
   await asyncio.sleep(self._audio_starvation_timeout)
   diff_time = time.time() - self._last_audio_timestamp
   if diff_time >= self._audio_starvation_timeout:
       self._vad_state = VADState.QUIET
       await self._call_event_handler("on_speech_stopped")
```

This will not be as accurate, but you could decrease the default to 0.5 and worst case it would be 1.0s. Or you could leave 1.0 and the worst case would be 2.0s.

aconchillo:
> The pattern mirrors what `UserTurnController` already does for its turn stop timeout : asyncio.Event + wait_for.

In that specific case, the event is set on the turn which happens occasionally. But in this case we can't set the event at every audio frame.

src/pipecat/audio/vad/vad_controller.py
aconchillo: Actually, there's a better solution, just sleep for the remaining time (i.e. (last_time + timeout) - current_time). This was the timeout works fine.

aconchillo (approved):

src/pipecat/audio/vad/vad_controller.py
```diff
@@ -149,6 +192,29 @@ async def _handle_vad(self, audio: bytes, vad_state: VADState) -> VADState:
             vad_state = new_vad_state
         return vad_state
 
+    async def _audio_idle_handler(self):
```
filipi87, line 195: Yep, this makes a lot of sense. 👍

filipi87 (approved):
LGTM!

aconchillo:
Thank you @omChauhanDev ! I have updated the PR, renamed starvation to idle since this is the convention we usually use in Pipecat.

src/pipecat/audio/vad/vad_controller.py
omChauhanDev: agreed, that's more efficient.

PR #4245, opened 2026-04-06
Stop audio mixer on pipeline cancellation

kollaikal-rupesh:
`cancel()` was missing `mixer.stop()` that `stop()` already had - added for consistency and custom mixer implementations.

### DIFF
--- a/changelog/2955.fixed.md
+++ b/changelog/2955.fixed.md
@@ -1,0 +1,1 @@
+- Fixed audio mixer not being stopped on pipeline cancellation, causing 100% CPU usage after call disconnect.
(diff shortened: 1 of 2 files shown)


markbackman:
Adding @filipi87 to look at this.

filipi87 (changes requested):
Hi @kollaikal-rupesh,

From what I can see, the current implementation of `SoundfileMixer` doesn’t actually perform any cleanup:

```
async def stop(self):
    """Clean up mixer resources.

    Currently performs no cleanup as sound data is managed by garbage collection.
    """
    pass
```

So unless there’s a custom mixer in use, calling this method likely won’t have any effect on CPU usage.

That said, I think it still makes sense to keep the call here, since it could be relevant for custom implementations.

Given that, would you mind removing the changelog entry so it doesn’t suggest a fix, and unlinking it from this PR in relation to issue 2955 ?

Thank you.

filipi87 (approved):
LGTM! Thank you @kollaikal-rupesh.

PR #4249, opened 2026-04-07
Add stitching support and enhancements for NvidiaTTSService

sathwikareddy02:
- Add stitching support and parameter enhancements for NvidiaTTSService
- Add cross-sentence stitching via persistent SynthesizeOnline gRPC stream
- Add custom_dictionary, encoding, zero_shot_audio_prompt_file parameters
- Make api_key optional for local NIM deployments
- Fix _update_settings to correctly indicate settings take effect on next turn
- Add can_generate_metrics, stop_all_metrics on interruption
- Add gRPC error handling for synthesis config retrieval
- Rename Riva references to Nemotron Speech

### DIFF
--- a/src/pipecat/services/nvidia/tts.py
+++ b/src/pipecat/services/nvidia/tts.py
@@ -13,11 +13,6 @@ gRPC API for high-quality speech synthesis.
 Refer to the NVIDIA TTS NIM documentation for usage, customization,
 and local deployment steps:
 https://docs.nvidia.com/nim/speech/latest/tts/
-
-For zero-shot voice cloning, request access to the Magpie TTS Zero-Shot model:
-https://developer.nvidia.com/riva-tts-zeroshot-models
-
-Local or private cloud deployment is recommended for best zero-shot performance.
 """
 
 import asyncio
@@ -26,7 +21,6 @@ import queue
 import textwrap
 import threading
 from dataclasses import dataclass, field
-from pathlib import Path
 from typing import Any, AsyncGenerator, Mapping, Optional
 
 from pipecat.utils.tracing.service_decorators import traced_tts
@@ -74,6 +68,19 @@ class NvidiaTTSSettings(TTSSettings):
     quality: int | _NotGiven = field(default_factory=lambda: NOT_GIVEN)
 
 
+@dataclass
+class _SynthesisStreamState:
+    """Runtime state for one active synthesis stream."""
+
+    context_id: str
+    text_queue: queue.Queue
+    response_queue: asyncio.Queue
+    stop_event: threading.Event
+    rpc_call: Any = None
+    synth_task: Optional[asyncio.Task] = None
+    response_task: Optional[asyncio.Task] = None
+
+
 class NvidiaTTSService(TTSService):
     """NVIDIA Nemotron Speech text-to-speech service.
 
@@ -116,8 +123,6 @@ class NvidiaTTSService(TTSService):
         use_ssl: bool = True,
         custom_dictionary: Optional[dict] = None,
         encoding: Optional[AudioEncoding] = AudioEncoding.LINEAR_PCM,
-        zero_shot_audio_prompt_file: Optional[Path] = None,
-        audio_prompt_encoding: Optional[AudioEncoding] = AudioEncoding.ENCODING_UNSPECIFIED,
         **kwargs,
     ):
         """Initialize the NVIDIA Nemotron Speech TTS service.
@@ -149,13 +154,6 @@ class NvidiaTTSService(TTSService):
                 https://docs.nvidia.com/nim/speech/latest/tts/phoneme-support.html
                 for the list of supported IPA phonemes.
             encoding: Output audio encoding format. Defaults to ``AudioEncoding.LINEAR_PCM``.
-            zero_shot_audio_prompt_file: Path to audio prompt file for zero-shot voice
-                cloning. Audio length should be between 3-10 seconds. The file
-                is read once at init time and cached in memory. Requires the
-                Magpie TTS Zero-Shot model. See
-                https://docs.nvidia.com/nim/speech/latest/tts/voice-cloning.html
-            audio_prompt_encoding: Encoding of the zero-shot audio prompt file.
-                Defaults to ``AudioEncoding.ENCODING_UNSPECIFIED``.
             **kwargs: Additional arguments passed to parent TTSService.
         """
         # 1. Initialize default_settings with hardcoded defaults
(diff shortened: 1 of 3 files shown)


markbackman:
@filipi87 can you take a look?

filipi87 (commented):
Hi @sathwikareddy02, thanks for the PR! A couple of suggestions:
                                                                                                                                                  
- Thread simplification
  - Instead of managing the thread manually, you could replace `threading.Thread` with `asyncio.to_thread(self._synthesis_thread_handler)` and wrap it with `self.create_task()`, which is how we handle blocking calls elsewhere in the codebase (for example [here for the nvidia stt](https://github.com/pipecat-ai/pipecat/blob/27a8a973b11f0865d3067db03cd6282fd02497f3/src/pipecat/services/nvidia/stt.py#L351)). 
  - This way you don't need to manage the thread lifecycle yourself, and the task gets tracked and cleaned up automatically by TaskManager.

- `flush_audio` vs `on_audio_context_interrupted`
  - Both currently call `_close_synthesis_stream()`, but they need different behaviors:
    - `flush_audio`: the turn is over and we want all remaining audio delivered. Closing the stream and waiting for it to finish is correct here.
    - `on_audio_context_interrupted`: the user interrupted, so we want to stop immediately and discard any pending audio.

So, I believe these are the two biggest changes that we need. Once they are done, and you think the PR is ready for review, please mark it as ready for review and I can take another look.

Thank you again for submitting these improvements.

sathwikareddy02:
@filipi87 Thanks for the detailed review and suggestions. I’ve incorporated both changes.

For the thread lifecycle, I replaced manual threading. Thread management with self.create_task(...) + asyncio.to_thread(...), so synth/response tasks are tracked and cancelled via TaskManager.

For behaviour split, flush_audio, and interruption now use separate paths:

flush_audio performs a graceful close and waits for the remaining audio to drain.
on_audio_context_interrupted performs immediate abort (stop flag, clear pending text queue, best-effort gRPC call cancel, cancel response/synth tasks), so pending audio is discarded immediately.

One nuance from runtime testing: a strict blocking flush wait can delay interruption handling in this pipeline, so flush waiting is implemented as an interruptible wait (it still drains normally, but exits immediately if interruption preempts). This preserves the intended flush semantics while keeping interruption responsive.

I marked the PR ready for review. I would appreciate another review pass.

src/pipecat/services/nvidia/tts.py
```diff
@@ -216,14 +273,21 @@ def _create_synthesis_config(self):
...
+                riva.client.proto.riva_tts_pb2.RivaSynthesisConfigRequest()
+            )
+            return config
+        except grpc.RpcError as e:
+            status = e.code().name if hasattr(e, "code") else "UNKNOWN"
+            details = e.details() if hasattr(e, "details") else str(e)
+            logger.error(
+                f"{self} failed to get synthesis config from server (gRPC {status}): {details}"
+            )
+            return None
```
filipi87: It looks like `_create_synthesis_config` may return `None`. Is this valid ? Should the service still work in this case?

If so, we should remove the assertion inside `run_tts`. This one:
`assert self._config is not None, "Synthesis configuration not created"`

src/pipecat/services/nvidia/tts.py
```diff
@@ -233,65 +297,336 @@ async def start(self, frame: StartFrame):
...
+        """Consume gRPC responses and append audio to the active audio context."""
+        while True:
+            item = await state.response_queue.get()
+            if item is None:
+                break
+            if isinstance(item, Exception):
+                # Ignore stale exceptions from interrupted streams.
+                if self._stream_state is state:
+                    await self.push_error(f"{self} synthesis error: {item}")
+                break
```
filipi87, line 425: If we receive an exception, should we always break? Or is it possible to receive both errors and audio intermixed? In that case, should we push the error without breaking, and only break when we receive `None`?

src/pipecat/services/nvidia/tts.py
```diff
@@ -233,65 +297,336 @@ async def start(self, frame: StartFrame):
...
+    async def stop(self, frame: EndFrame):
+        """Stop the NVIDIA Nemotron Speech TTS service.
 
         Args:
-            text: The text to synthesize into speech.
-            context_id: The context ID for tracking audio frames.
+            frame: The end frame.
+        """
+        await super().stop(frame)
+        await self._close_synthesis_stream()
```
filipi87: Here, you should close any active connections. You shouldn’t wait for the task to finish producing audio, just close any connections or tasks.

src/pipecat/services/nvidia/tts.py
```diff
@@ -233,65 +297,336 @@ async def start(self, frame: StartFrame):
...
-        Yields:
-            Frame: Audio frames containing the synthesized speech data.
+    async def cancel(self, frame: CancelFrame):
+        """Cancel the NVIDIA Nemotron Speech TTS service.
+
+        Args:
+            frame: The cancel frame.
         """
+        await super().cancel(frame)
+        await self._close_synthesis_stream()
```
filipi87: Same here. You should close any active connections. You shouldn’t wait for the task to finish producing audio, just close any connections or tasks.

src/pipecat/services/nvidia/tts.py
```diff
@@ -233,65 +297,336 @@ async def start(self, frame: StartFrame):
...
-                yield item
+        # Finalize ownership once the stream drains naturally.
+        if self._stream_state is state and not state.stop_event.is_set():
+            self._stream_state = None
+
+    def _signal_synthesis_close(self, state: _SynthesisStreamState):
+        """Signal the active synthesis request generator to close."""
+        state.text_queue.put(None)
+
+    async def _close_synthesis_stream(self):
```
filipi87: So, I guess this method can be completely removed.

src/pipecat/services/nvidia/tts.py
```diff
@@ -233,65 +297,336 @@ async def start(self, frame: StartFrame):
...
+        """Flush any pending audio and finalize the current context.
+
+        Args:
+            context_id: The specific context to flush. If None, falls back to the
+                currently active context.
+        """
+        state = self._stream_state
+        if state is not None:
+            self._signal_synthesis_close(state)
+            await self._wait_for_synthesis_close_interruptibly(state)
```
filipi87: Inside `flush_audio`, all you need to do is signal the server to send any remaining audio.

In your case, this is essentially what you’re doing with the `_signal_synthesis_close` approach, where adding `None` to the `text_queue` will make you stop sending audio and receive the remaining responses.

I don’t think `_wait_for_synthesis_close_interruptibly` is necessary, this should happen naturally.

It would be worth testing by completely removing `_wait_for_synthesis_close_interruptibly` as well.

In case of interruption, your existing logic already handles it correctly.

src/pipecat/services/nvidia/tts.py
sathwikareddy02: Changed to fail-fast startup: _create_synthesis_config() now raises on failure, and run_tts no longer asserts _config.

src/pipecat/services/nvidia/tts.py
sathwikareddy02: Kept exception as terminal for that stream (break), because once SynthesizeOnline errors, the stream is no longer considered reliable.
Added an additional guard to suppress noisy error emission during intentional interruption cases:
only push error when the stream is still active and not already interrupted.

src/pipecat/services/nvidia/tts.py
sathwikareddy02: Updated stop() to call immediate stream abort path (_abort_synthesis_stream) rather than graceful drain.

src/pipecat/services/nvidia/tts.py
sathwikareddy02: Updated cancel() to use the same immediate abort path.

src/pipecat/services/nvidia/tts.py
sathwikareddy02: Removed graceful-close helper that was no longer needed after switching stop/cancel behavior.
Also removed the related interruptible wait helper used only by that path.

src/pipecat/services/nvidia/tts.py
sathwikareddy02: Simplified flush_audio() to only signal request-generator close (None sentinel).
Removed explicit wait logic; remaining audio delivery proceeds naturally via the existing pipeline/context flow.

filipi87 (approved):
This looks great.

I have pushed one final improvement to remove the audio context when there is no more audio to receive.

Thank you for doing this @sathwikareddy02.

PR #4251, opened 2026-04-07
Add Mistral Voxtral streaming TTS service

markbackman:
## Summary

- Added `MistralTTSService` for streaming text-to-speech using Mistral's Voxtral TTS API (`voxtral-mini-tts-2603`) via HTTP with Server-Sent Events
- Converts base64-encoded float32 PCM from the API to int16 for the Pipecat pipeline, with automatic resampling from the native 24kHz to any requested sample rate
- Added `mistralai>=2.0.0,<3` dependency to the `mistral` optional extra
- Added voice example at `examples/voice/voice-mistral.py`

## Testing

- `uv run ruff check src/pipecat/services/mistral/tts.py` — passes
- `uv run ruff format --check src/pipecat/services/mistral/tts.py` — passes
- `uv run pytest` — all 1133 tests pass
- Test with a Mistral API key: `uv run python examples/voice/voice-mistral.py`


src/pipecat/services/mistral/tts.py
```diff
@@ -0,0 +1,183 @@
...
+        **kwargs,
+    ):
+        """Initialize Mistral TTS service.
+
+        Args:
+            api_key: Mistral API key for authentication. If None, uses
+                MISTRAL_API_KEY environment variable.
+            voice_id: Voice ID to use for synthesis.
+
+                .. deprecated:: 0.0.105
```
kompfner: This is a new service. We shouldn't have any deprecations.

src/pipecat/services/mistral/tts.py
```diff
@@ -0,0 +1,183 @@
...
+            **kwargs: Additional keyword arguments passed to TTSService.
+        """
+        # 1. Initialize default_settings with hardcoded defaults
+        default_settings = self.Settings(
+            model="voxtral-mini-tts-2603",
+            voice=None,
+            language=None,
+        )
+
+        # 2. Apply direct init arg overrides (deprecated)
```
kompfner: See above.

kompfner (approved):
One comment. Pre-approving!

src/pipecat/services/mistral/tts.py
markbackman: Thanks. I had Claude write this then focused on a few issues, but lost sight of the forest for the trees. Cleaned up now.

PR #4252, opened 2026-04-07
VIVA SDK TT v3 support

realgarik:

Krisp is going to release the new Turn Taking v3 model which requires slight API changes.
Please review, but don't merge before Krisp VIVA SDK release with the mentioned changes will not be released


markbackman (approved):
LGTM!

@realgarik please let us know when we should merge this. Making this DRAFT for now so it's not merged by mistake.

src/pipecat/audio/turn/krisp_viva_turn.py
```diff
@@ -307,12 +311,7 @@ def append_audio(self, buffer: bytes, is_speech: bool) -> EndOfTurnState:
                 # Instead, we wait for the model's probability check below to confirm
                 # end-of-turn based on the threshold.
 
-                prob = self._tt_session.process(frame.tolist())
-
-                # Negative values indicate the model is not ready yet (working with 100ms data)
-                # Skip processing until we get positive probabilities
-                if prob < 0:
-                    continue
+                prob = self._tt_session.process(frame.tolist(), is_speech, False)
```
filipi87, line 314: What does the third `False` argument mean in `self._tt_session.process(frame.tolist(), is_speech, False)` ?

filipi87 (commented):
Hi @realgarik, the changes look good.

We just need to rebase with the latest changes to resolve the conflicts, and I think we’re good to go.

Thank you for doing this!

src/pipecat/audio/turn/krisp_viva_turn.py
realgarik: third argument is bool flag for botSpeaking.

src/pipecat/turns/user_start/__init__.py
filipi87, line 28: We should use try/catch approach, which is how we handle everywhere in Pipecat.

 Something like this:
```
from .base_user_turn_start_strategy import BaseUserTurnStartStrategy, UserTurnStartedParams
from .external_user_turn_start_strategy import ExternalUserTurnStartStrategy
from .min_words_user_turn_start_strategy import MinWordsUserTurnStartStrategy
from .transcription_user_turn_start_strategy import TranscriptionUserTurnStartStrategy
from .vad_user_turn_start_strategy import VADUserTurnStartStrategy
from .wake_phrase_user_turn_start_strategy import WakePhraseUserTurnStartStrategy

try:
    from .krisp_viva_ip_user_turn_start_strategy import KrispVivaIPUserTurnStartStrategy
except ImportError:  # or ModuleNotFoundError
    KrispVivaIPUserTurnStartStrategy = None  # or omit entirely

__all__ = [
    "BaseUserTurnStartStrategy",
    "ExternalUserTurnStartStrategy",
    "MinWordsUserTurnStartStrategy",
    "TranscriptionUserTurnStartStrategy",
    "UserTurnStartedParams",
    "VADUserTurnStartStrategy",
    "WakePhraseUserTurnStartStrategy",
]

if KrispVivaIPUserTurnStartStrategy is not None:
    __all__.append("KrispVivaIPUserTurnStartStrategy")
```

src/pipecat/turns/user_turn_strategies.py
```diff
@@ -9,7 +9,6 @@
 from dataclasses import dataclass
 from typing import List, Optional
 
-from pipecat.audio.turn.smart_turn.local_smart_turn_v3 import LocalSmartTurnAnalyzerV3
```
filipi87, line 12: We can keep it here, as it was originally.

tests/test_krisp_sdk_manager.py
```diff
@@ -62,8 +62,15 @@ class TestKrispVivaSDKManager:
...
         mock_krisp_audio.reset_mock()
+        mock_krisp_audio.globalInit.side_effect = None
         mock_krisp_audio.getVersion.return_value = mock_version
 
+        # Ensure krisp_instance module uses THIS test's mock, not a stale
+        # reference cached from a different test file's sys.modules entry.
+        import pipecat.audio.krisp_instance as _ki
+
+        _ki.krisp_audio = mock_krisp_audio
+
```
filipi87, line 73: Why was this change needed ?

src/pipecat/turns/user_turn_strategies.py
filipi87: The failing tests may be related to this, since all the tests are working fine on `main`.

realgarik: ok, will revert

tests/test_krisp_sdk_manager.py
realgarik: At least in my local environment when pytest runs multiple test files in one process, test_krisp_ip_user_turn_start_strategy.py executes before test_krisp_sdk_manager.py (alphabetical order). Both files create their own MagicMock() for krisp_audio and install it in sys.modules["krisp_audio"].

However, Python only executes import krisp_audio inside krisp_instance.py once -- the first time the module is loaded. After that, the module-level name krisp_audio in krisp_instance.py is permanently bound to whichever mock was in sys.modules at that time (the first test file's mock).

When test_krisp_sdk_manager.py later sets sys.modules["krisp_audio"] to its own mock, that has no effect -- krisp_instance.krisp_audio still points to the old one. So KrispVivaSDKManager.acquire() calls globalInit() on the other test's mock, while this test's assertions check this mock. Result: assert_called_once() sees 0 calls.

The fix _ki.krisp_audio = mock_krisp_audio directly rebinds the module-level name so all code in krisp_instance.py uses the correct mock for each test.

filipi87 (approved):
LGTM!
Thank you for doing this @realgarik . 🚀

PR #4253, opened 2026-04-07
Add Mistral Voxtral Realtime STT service

markbackman:
## Summary

- Added `MistralSTTService` for real-time speech-to-text using Mistral's Voxtral Realtime API (`voxtral-mini-transcribe-realtime-2602`), with streaming interim transcriptions, automatic language detection, and VAD-driven utterance lifecycle
- Extends `STTService` directly (not `WebsocketSTTService`) since the Mistral SDK manages its own WebSocket via `RealtimeConnection`
- Added `MISTRAL_TTFS_P99` latency constant to `stt_latency.py`
- Updated `voice-mistral` example to use Mistral STT instead of Deepgram
- Added `transcription-mistral` example for standalone transcription

## Testing

- `python -c "from pipecat.services.mistral.stt import MistralSTTService; print('OK')"`
- `uv run ruff check src/pipecat/services/mistral/stt.py`
- Run `examples/voice/voice-mistral.py` with a Mistral API key
- Run `examples/transcription/transcription-mistral.py` with a Mistral API key


filipi87 (approved):
LGTM!

PR #4254, opened 2026-04-07
Prompting tweaks for streaming async tools

kompfner:


filipi87 (approved):
The results feel really close to what we had before, but I think the descriptions here are a lot better.
Thank you, @kompfner.

PR #4257, opened 2026-04-08
Async tools stream context improvements

filipi87:
Async tools stream context improvements:
- Deferring pushing the context frame if the bot is speaking.
- Deferring pushing the context frame if we have another `FunctionCallResultFrames` queued.

### DIFF
--- a/src/pipecat/processors/aggregators/llm_response_universal.py
+++ b/src/pipecat/processors/aggregators/llm_response_universal.py
@@ -25,6 +25,8 @@ from pipecat.audio.vad.vad_analyzer import VADAnalyzer
 from pipecat.audio.vad.vad_controller import VADController
 from pipecat.frames.frames import (
     AssistantImageRawFrame,
+    BotStartedSpeakingFrame,
+    BotStoppedSpeakingFrame,
     CancelFrame,
     EndFrame,
     Frame,
@@ -790,6 +792,22 @@ class LLMAssistantAggregator(LLMContextAggregator):
     The aggregator manages function calls in progress and coordinates between
     text generation and tool execution phases of LLM responses.
 
+    **Deferred context frame push (bot-speaking guard)**
+
+    When a function call result arrives while the bot is already speaking (i.e., a
+    ``BotStartedSpeakingFrame`` has been seen but no matching ``BotStoppedSpeakingFrame``
+    yet), pushing a context frame immediately would cause the LLM to start a new inference
+    pass while the current TTS output is still playing. If several function call results
+    arrive in rapid succession during that window they would each trigger a separate LLM
+    run, leading to duplicated or overlapping responses.
+
+    To prevent this, the aggregator sets an internal ``_pending_context_frame`` flag
+    instead of pushing immediately. When ``BotStoppedSpeakingFrame`` is finally received
+    — meaning TTS has finished — the aggregator checks the flag and, if set, pushes a
+    single context frame upstream. Because all results that arrived during the speaking
+    window have already been added to the context, the LLM sees them in one combined
+    call rather than multiple staggered ones.
+
     Event handlers available:
 
     - on_assistant_turn_started: Called when the assistant turn starts
@@ -839,6 +857,13 @@ class LLMAssistantAggregator(LLMContextAggregator):
         self._context_updated_tasks: Set[asyncio.Task] = set()
 
         self._user_speaking: bool = False
+        self._bot_speaking: bool = False
+
+        # When a function call result arrives while the bot is speaking, we defer the LLM
+        # re-invocation until the bot stops speaking. This flag is set to True in that case
+        # so that `BotStoppedSpeakingFrame` knows to push a context frame. Multiple results
+        # arriving in the same speaking window are bundled into a single deferred push.
+        self._pending_context_frame: bool = False
 
         self._assistant_turn_start_timestamp = ""
 
@@ -879,6 +904,7 @@ class LLMAssistantAggregator(LLMContextAggregator):
         """Reset the aggregation state."""
         await super().reset()
         await self._reset_thought_aggregation()  # Just to be safe
+        self._pending_context_frame = False
 
     async def _reset_thought_aggregation(self):
         """Reset the thought aggregation state."""
@@ -950,6 +976,16 @@ class LLMAssistantAggregator(LLMContextAggregator):
         elif isinstance(frame, UserStoppedSpeakingFrame):
             self._user_speaking = False
             await self.push_frame(frame, direction)
+        elif isinstance(frame, BotStartedSpeakingFrame):
+            self._bot_speaking = True
+            await self.push_frame(frame, direction)
+        elif isinstance(frame, BotStoppedSpeakingFrame):
+            self._bot_speaking = False
+            await self.push_frame(frame, direction)
+            if self._pending_context_frame and not self._user_speaking:
+                self._pending_context_frame = False
+                logger.debug(f"{self}: Bot stopped speaking — pushing deferred context frame!")
+                await self.push_context_frame(FrameDirection.UPSTREAM)
         else:
             await self.push_frame(frame, direction)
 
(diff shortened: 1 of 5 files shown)


src/pipecat/processors/aggregators/llm_response_universal.py
```diff
@@ -790,6 +785,22 @@ class LLMAssistantAggregator(LLMContextAggregator):
...
     text generation and tool execution phases of LLM responses.
 
+    **Deferred context frame push (bot-speaking guard)**
+
+    When a function call result arrives while the bot is already speaking (i.e., a
+    ``BotStartedSpeakingFrame`` has been seen but no matching ``BotStoppedSpeakingFrame``
+    yet), pushing a context frame immediately would cause the LLM to start a new inference
+    pass while the current TTS output is still playing. If several function call results
+    arrive in rapid succession during that window they would each trigger a separate LLM
+    run, leading to duplicated or overlapping responses.
```
kompfner: The problem isn't necessarily duplication or overlap, right, because of your other fix ensuring that each pushed context frame takes a snapshot of the context rather than holding a live reference to the context? Isn't the problem more than we'd just be generating way more responses than are necessarily needed?

filipi87: Yes, but in this PR I also removed that other fix, since we agreed it was too risky and could cause unknown side effects. 

Without that protection, this could lead to duplication or overlap.

src/pipecat/processors/aggregators/llm_response_universal.py
```diff
@@ -790,6 +785,22 @@ class LLMAssistantAggregator(LLMContextAggregator):
     The aggregator manages function calls in progress and coordinates between
     text generation and tool execution phases of LLM responses.
 
+    **Deferred context frame push (bot-speaking guard)**
+
+    When a function call result arrives while the bot is already speaking (i.e., a
```
kompfner: Do we want this mechanism to be specific to results from the new async tools?

src/pipecat/processors/aggregators/llm_response_universal.py
```diff
@@ -790,6 +785,22 @@ class LLMAssistantAggregator(LLMContextAggregator):
     The aggregator manages function calls in progress and coordinates between
     text generation and tool execution phases of LLM responses.
 
+    **Deferred context frame push (bot-speaking guard)**
```
kompfner: Nit: feels out of place for this quite detailed description of this mechanism to live in this high-level docstring.

src/pipecat/processors/aggregators/llm_response_universal.py
```diff
@@ -1122,8 +1159,24 @@ async def _handle_function_call_result(self, frame: FunctionCallResultFrame):
                     run_llm = True
 
         if run_llm and not self._user_speaking:
-            logger.debug(f"{self}: Pushing context frame!")
-            await self.push_context_frame(FrameDirection.UPSTREAM)
+            if self.has_queued_frame(FunctionCallResultFrame):
+                # Another FunctionCallResultFrame is already waiting in the queue. Defer
```
kompfner: Ah, I see, so we're bundling in two distinct cases:
1. If we're about to process more function call results (in which case it makes sense to wait for all of them before triggering an inference)
2. If the bot is currently speaking (in which case it makes sense to wait for the bot to finish speaking before triggering inference, in case more results come in)

src/pipecat/processors/aggregators/llm_response_universal.py
filipi87: Right now, it’s implemented for `FunctionCallResultFrame` in general, which I think makes sense. It will just be harder to hit this path in other cases.

src/pipecat/processors/aggregators/llm_response_universal.py
filipi87: I have shorter descriptions in other places as well. Do you think I should remove this one ?

src/pipecat/processors/aggregators/llm_response_universal.py
kompfner: In scenario 1, by setting `self._pending_context_frame = True` wouldn't we be unintentionally deferring until the next time the bot finishes speaking (as opposed to what we want, which is deferring only until the last result arrives)?

filipi87: Yep, we’re handling both cases now.

Since `FunctionCallResultFrame` is not a `SystemFrame`, there’s a good chance it will be queued until the bot finishes speaking. As a result, we can have multiple `FunctionCallResultFrames` in the queue.

filipi87: So, for scenario 1, we are only deferring until we process the next `FunctionCallResultFrame` from the queue, which should happen almost instantaneously.

In this case, we will push a single context frame.

src/pipecat/processors/aggregators/llm_response_universal.py
kompfner: IMO, yeah. Maybe make sure the other ones are clear enough to stand on their own.

src/pipecat/processors/aggregators/llm_response_universal.py
```diff
@@ -1122,8 +1159,24 @@ async def _handle_function_call_result(self, frame: FunctionCallResultFrame):
...
+                # the context push so all results are bundled into a single LLM call
+                # instead of triggering one inference pass per result.
+                logger.debug(
+                    f"{self}: More FunctionCallResultFrames queued — deferring context frame push."
+                )
+                self._pending_context_frame = True
```
kompfner: Ah, I understand now. Instead of setting `_pending_context_frame` here (which _should_ be a no-op) we could instead just rely on the comment to do the explaining of what's happening here.

src/pipecat/processors/aggregators/llm_response_universal.py
```diff
@@ -1122,8 +1159,23 @@ async def _handle_function_call_result(self, frame: FunctionCallResultFrame):
...
-            await self.push_context_frame(FrameDirection.UPSTREAM)
+            if self.has_queued_frame(FunctionCallResultFrame):
+                # Another FunctionCallResultFrame is already waiting in the queue. Defer
+                # the context push so all results are bundled into a single LLM call
+                # instead of triggering one inference pass per result.
+                logger.debug(
+                    f"{self}: More FunctionCallResultFrames queued — deferring context frame push."
+                )
+            elif self._bot_speaking:
+                # Defer the context frame push until the bot finishes speaking. If multiple
```
kompfner, line 1155: Per our discussion: this is really quite an edge case, since it would require a `FunctionCallResultFrame` being queued *between* an LLM response start and end frame.

src/pipecat/processors/aggregators/llm_response_universal.py
```diff
@@ -1122,8 +1159,23 @@ async def _handle_function_call_result(self, frame: FunctionCallResultFrame):
...
+                # function call results arrive while the bot is speaking, they all accumulate
+                # in the context and a single push is performed once speaking stops, preventing
+                # the LLM from running multiple times and producing duplicated responses.
+                logger.debug(f"{self}: Bot is speaking — deferring context frame push.")
+                self._pending_context_frame = True
```
kompfner: nit: `_pending_context_frame` feels a bit too generic/not quite descriptive...

ideas:
* `_push_deferred_context_frame`
* `_push_context_frame_on_bot_stopped_speaking`
* `_push_context_on_bot_stopped_speaking`

kompfner (approved):

kompfner (approved):
After the couple remaining nits (name change, comment removal), this looks good!

src/pipecat/processors/aggregators/llm_response_universal.py

src/pipecat/processors/aggregators/llm_response_universal.py

src/pipecat/processors/aggregators/llm_response_universal.py
filipi87: I have mentioned that in the comment. 👍

src/pipecat/processors/aggregators/llm_response_universal.py
filipi87: Renamed to `_push_context_on_bot_stopped_speaking`.

filipi87:
Thank you for the review @kompfner. 🙌

PR #4265, opened 2026-04-09
Using the correct default for auto_mode based on text_aggregation_mode.

filipi87:
## Summary
- Fixed `ElevenLabsTTSService` defaulting `auto_mode` to `True` regardless of the text aggregation strategy.
- `auto_mode` now defaults to `True` for `SENTENCE` aggregation and `False` for `TOKEN` aggregation.
- Callers can still override by passing `auto_mode` explicitly


markbackman (approved):
LGTM

PR #4267, opened 2026-04-09
Fix observer cleanup ordering to stop proxy tasks before closing resources

### LINKED ISSUE #4195: TaskObserver.cleanup() closes observer resources before stopping proxy tasks
## Bug Description

During pipeline shutdown, `TaskObserver.cleanup()` calls `observer.cleanup()` (which closes resources like files) **before** the proxy tasks are cancelled. This causes a race condition where proxy tasks still try to write frames to already-closed resources.

## Error Message

```
ERROR - TaskObserver::WhiskerObserver#0::_proxy_task_handler unexpected exception: write to closed file
```

## Reproduction

1. Use `WhiskerObserver` with `file_name=` parameter (session recording to disk)
2. Run a normal conversation through the pipeline
3. End the conversation (normal `EndFrame` flow)
4. Observe the error in logs during shutdown

The session file is saved correctly despite the error — data is flushed before close. The error comes from late-arriving frames hitting the closed file handle.

## Root Cause

In `pipecat/pipeline/task_observer.py`, the `cleanup()` method (line ~128) iterates over proxies and calls `proxy.cleanup()` which triggers the observer's `cleanup()` (closing files, connections etc.). However, the proxy tasks created via `_proxy_task_handler` are still running at this point — they are only cancelled later when `stop()` is called from `PipelineTask.run()`'s finally block.

The sequence is:
1. `PipelineTask._cleanup()` → `TaskObserver.cleanup()` → `WhiskerObserver.cleanup()` → **file closed**
2. Proxy task dequeues remaining frame → `observer.on_process_frame()` → `_queue_data()` → `await self._file.write(msg)` → **ValueError: write to closed file**
3. Later: `PipelineTask.run()` finally → `TaskObserver.stop()` → proxy tasks cancelled (too late)

## Suggested Fix

In `TaskObserver.cleanup()`, stop the proxy tasks before cleaning up observers:

```python
async def cleanup(self):
    await super().cleanup()
    # Stop proxy tasks FIRST so no more frames arrive
    await self.stop()
    # THEN clean up observers
    if not self._proxies:
        return
    for proxy in self._proxies:
        await proxy.cleanup()
```

## Environment

- pipecat-ai: 0.0.107
- pipecat-ai-whisker: 0.0.11
- Python: 3.11
- Observer: WhiskerObserver with file recording enabled

aconchillo:
## Summary

- Fixed `ValueError: write to closed file` during pipeline shutdown when observers were active. Proxy tasks are now explicitly stopped before observer resources are cleaned up.
- Moved `TaskObserver.start()` to after external setup files are loaded, ensuring all observers are registered before proxy tasks begin.
- Removed deprecated `_load_observer_files()` and `PIPECAT_OBSERVER_FILES` env var (use `PIPECAT_SETUP_FILES` instead).
- Fixed misleading variable name in `TaskObserver.cleanup()` iteration.

## Breaking Changes

- Removed `PIPECAT_OBSERVER_FILES` environment variable support. Use `PIPECAT_SETUP_FILES` instead (deprecated since previous release).

## Fixes

- Fixes #4195

### DIFF
--- a/changelog/4267.fixed.md
+++ b/changelog/4267.fixed.md
@@ -1,0 +1,1 @@
+- Fixed `ValueError: write to closed file` during pipeline shutdown when observers were active. Observer proxy tasks are now cancelled before observer resources are cleaned up.
--- a/changelog/4267.removed.md
+++ b/changelog/4267.removed.md
@@ -1,0 +1,1 @@
+- ⚠️ Removed deprecated `PIPECAT_OBSERVER_FILES` environment variable support. Use `PIPECAT_SETUP_FILES` instead.
--- a/src/pipecat/pipeline/task.py
+++ b/src/pipecat/pipeline/task.py
@@ -613,9 +613,6 @@ class PipelineTask(BasePipelineTask):
         self._process_push_task = self._task_manager.create_task(
             self._process_push_queue(), f"{self}::_process_push_queue"
         )
-
-        await self._observer.start()
-
         return self._process_push_task
 
     def _maybe_start_heartbeat_tasks(self):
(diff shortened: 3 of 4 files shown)


filipi87 (approved):
LGTM!

PR #4269, opened 2026-04-10
Update NVIDIA STT services for Nemotron Speech defaults and config parity

sathwikareddy02:
Align the NVIDIA STT integrations with Nemotron Speech naming, defaults, and runtime behaviour.

This updates docs and logs from Riva to Nemotron Speech terminology, expands STT settings
(max alternatives, word offsets, diarization), and adds constructor options for endpointing/custom
configuration. It also improves streaming audio duration tracking and tightens segmented recognition
error handling and config construction.


sathwikareddy02:
@filipi87 and @markbackman, can you help review this PR?

filipi87:
Hi @sathwikareddy02 , I will take a look. Thank you for pushing this.

src/pipecat/services/nvidia/stt.py
```diff
@@ -143,32 +171,52 @@ class InputParams(BaseModel):
     def __init__(
         self,
         *,
-        api_key: str,
+        api_key: Optional[str] = None,
```
filipi87, line 174: We have modernized Python typing in this https://github.com/pipecat-ai/pipecat/pull/4319. We should follow this new pattern now.

So for example here it should be:

```suggestion
        api_key: str | None,
```

filipi87: I think if you rebase with the latest changes from `main`, and run `./scripts/fix-ruff.sh`, it will apply those changes automatically, if I’m not mistaken.

src/pipecat/services/nvidia/stt.py
```diff
@@ -470,7 +583,7 @@ class InputParams(BaseModel):
...
```
filipi87, line 586: Same here. And probably in a couple of other places in this class.

filipi87 (commented):
The changes look good. We just need to rebase onto the latest changes from `main` and make sure we are using modernized Python typing.

src/pipecat/services/nvidia/stt.py
sathwikareddy02: Rebased the branch and updated the python styles

src/pipecat/services/nvidia/stt.py
sathwikareddy02: Rebased here too

src/pipecat/services/nvidia/stt.py
```diff
@@ -370,30 +472,36 @@ async def _handle_response(self, response):
...
                 if result.is_final:
                     await self.stop_processing_metrics()
+                    if hasattr(result, "audio_processed") and result.audio_processed:
+                        server_lag = self._audio_duration_counter - result.audio_processed
+                        logger.debug(
+                            f"{self} ASR server-side lag: {server_lag:.3f}s "
+                            f"(audio sent: {self._audio_duration_counter:.3f}s, "
+                            f"audio processed: {result.audio_processed:.3f}s)"
+                        )
+                    logger.debug(f"Transcription: [{transcript}]")
```
filipi87: Do we need to keep this ? It feels like it would generate a lot of logs. If we remove it, we can also remove the `self._audio_duration_counter` variable.

src/pipecat/services/nvidia/stt.py
```diff
@@ -370,30 +472,36 @@ async def _handle_response(self, response):
...
                     await self.push_frame(
                         TranscriptionFrame(
                             transcript,
                             self._user_id,
                             time_now_iso8601(),
-                            self._settings.language,
```
filipi87, line 378: It would be nice to preserve the `language` when creating the `TranscriptionFrame` and `InterimTranscriptionFrame`.

src/pipecat/services/nvidia/stt.py
```diff
@@ -422,7 +530,10 @@ def __next__(self) -> bytes:
 
         try:
             future = asyncio.run_coroutine_threadsafe(self._queue.get(), self.get_event_loop())
-            return future.result()
+            audio = future.result()
+            samples = len(audio) // (2 * self._audio_channel_count)
+            self._audio_duration_counter += samples / self.sample_rate
+            return audio
```
filipi87, line 529: I think these changes can be reverted if we remove that log which I mentioned above.

src/pipecat/services/nvidia/stt.py
sathwikareddy02: Now removed the processing latency log and its related variables

src/pipecat/services/nvidia/stt.py
sathwikareddy02: reverted this change

src/pipecat/services/nvidia/stt.py
sathwikareddy02: yes these are removed now

filipi87 (approved):
LGTM!

PR #4270, opened 2026-04-10
Enhance NVIDIA LLM reasoning tokens handling and allow keyless local …

sathwikareddy02:
Improve NvidiaLLMService to better support modern NIM model behaviour and local deployments.

This adds reasoning extraction for both `reasoning_content` deltas and `<think>...</think>` content, emitting `LLMThought*Frame`s so reasoning stays observable without leaking into normal assistant text. It also updates the default model, allows optional `api_key` for local endpoints (with cloud-endpoint warning), and keeps NVIDIA token usage accounting behaviour intact with safer per-response state/reset handling.

### DIFF
--- a/src/pipecat/services/nvidia/llm.py
+++ b/src/pipecat/services/nvidia/llm.py
@@ -2,20 +2,37 @@
 # Copyright (c) 2024-2026, Daily
 #
 # SPDX-License-Identifier: BSD 2-Clause License
+# Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES.
 #
 
 """NVIDIA NIM API service implementation.
 
 This module provides a service for interacting with NVIDIA's NIM (NVIDIA Inference
 Microservice) API while maintaining compatibility with the OpenAI-style interface.
+
+Refer to the NVIDIA NIM LLM API documentation for available models and usage:
+https://docs.api.nvidia.com/nim/reference/llm-apis
 """
 
 from dataclasses import dataclass
+from typing import AsyncIterator, Optional
 
+from loguru import logger
+from openai import AsyncStream
+from openai.types.chat import ChatCompletionChunk
+
+from pipecat.frames.frames import (
+    LLMThoughtEndFrame,
+    LLMThoughtStartFrame,
+    LLMThoughtTextFrame,
+)
 from pipecat.metrics.metrics import LLMTokenUsage
 from pipecat.processors.aggregators.llm_context import LLMContext
 from pipecat.services.openai.base_llm import BaseOpenAILLMService
 from pipecat.services.openai.llm import OpenAILLMService
+
+_THINK_OPEN = "<think>"
+_THINK_CLOSE = "</think>"
 
 
 @dataclass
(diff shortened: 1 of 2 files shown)


sathwikareddy02:
@filipi87 / @markbackman, can you help review this PR too?

filipi87:
Hi @sathwikareddy02 , I will take a look. Thank you for pushing this.

src/pipecat/services/nvidia/llm.py
```diff
@@ -40,7 +65,7 @@ class NvidiaLLMService(OpenAILLMService):
     def __init__(
         self,
         *,
-        api_key: str,
+        api_key: Optional[str] = None,
```
filipi87, line 68: We have modernized Python typing in this [PR](https://github.com/pipecat-ai/pipecat/pull/4319). We should follow this new pattern now. 
So for example here it should be:


```suggestion
        api_key: str | None,
```

filipi87: I think if you rebase with the latest changes from `main`, and run `./scripts/fix-ruff.sh`, it will apply those changes automatically, if I’m not mistaken.

src/pipecat/services/nvidia/llm.py
```diff
@@ -76,33 +103,191 @@ def __init__(
...
         self._is_processing = True
 
+        # <think> tag detection: "detecting" → "in_thought" | "content"
+        self._think_tag_state = "detecting"
+        self._think_tag_buffer = ""
+
+        # reasoning_content field tracking
+        self._has_reasoning_field = False
+
+    async def _push_llm_text(self, text: str):
```
filipi87, line 140: It would be better to create a new method to handle the NVIDIA responses and invoke it inside `_process_context`, instead of overriding this method from the parent class.

src/pipecat/services/nvidia/llm.py
```diff
@@ -76,33 +103,191 @@ def __init__(
...
         """
-        # Reset all counters and flags at the start of processing
         self._prompt_tokens = 0
         self._completion_tokens = 0
         self._total_tokens = 0
         self._has_reported_prompt_tokens = False
         self._is_processing = True
 
+        # <think> tag detection: "detecting" → "in_thought" | "content"
+        self._think_tag_state = "detecting"
```
filipi87, line 134: It might be better to create an enum to represent all these possible states.

src/pipecat/services/nvidia/llm.py
filipi87: Oh, I guess this might not be possible, right? Since it’s the `base_llm` class that invokes it.

filipi87: Ok, but maybe a cleaner alternative is to handle this inside `_handle_reasoning_content`. Then you can drop the `_push_llm_text` override entirely, and you will handle both reasoning and thinking in one place.

filipi87:
We also need to rebase onto the latest changes from `main` and make sure we are using modernized Python typing.

src/pipecat/services/nvidia/llm.py
sathwikareddy02: Rebased and updated now

src/pipecat/services/nvidia/llm.py
sathwikareddy02: Done, we are now handling reasoning content inside single method.

src/pipecat/services/nvidia/llm.py
sathwikareddy02: Created an enum class for this state management

filipi87 (approved):
Looks good. Thank you for the improvements @sathwikareddy02. 🚀

PR #4275, opened 2026-04-13
 Fix Inworld TTS word timestamps dropping punctuation from assistant context

omChauhanDev:

Fixes: #4261 

**Issue :** Inworld's word alignment API returns bare words, these become assistant context via TTSTextFrame, so the LLM sees unpunctuated history and imitates that style   
                                                                     
**Fix :** Added `_restore_punctuation()` to map aligned words back to original text tokens, recovering punctuation. Falls back to bare word on mismatch. Applied to both HTTP and WebSocket services

### DIFF
--- a/changelog/4261.fixed.md
+++ b/changelog/4261.fixed.md
@@ -1,0 +1,1 @@
+- Fixed `InworldTTSService` word timestamps returning bare words without punctuation, causing the assistant context to lose punctuation and the LLM to imitate that style on subsequent turns.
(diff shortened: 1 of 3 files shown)


markbackman:
Thanks @omChauhanDev but I'm going to close this. The root cause is that Inworld doesn't return punctuation. We should request that Inworld fix this issue instead of hacking around it.

omChauhanDev:
Agreed, this should ideally be fixed on Inworld’s side like other services that already return punctuated text.

PR #4280, opened 2026-04-13
Update uv.lock resolving langchain-core and cryptography vulnerabilities

markbackman:

No changelog needed, as these versions are already supported in the pyproject.toml version ranges.

Note: There's one more open security vulnerability, but it's been blocked by Speechmatics as they depend on `transformers` <5. I've reached out to them directly to request that they bump the supported range to <6. This will only affect developers if they're using Speechmatics.

### DIFF
--- a/uv.lock
+++ b/uv.lock
@@ -1047,61 +1047,61 @@ wheels = [
 
 [[package]]
 name = "cryptography"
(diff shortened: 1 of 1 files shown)


filipi87 (approved):
LGTM!

PR #4282, opened 2026-04-13
Reconnect Cartesia STT websocket on settings change

markbackman:
## Summary

- `CartesiaSTTService` now supports runtime settings updates (e.g. changing `language` or `model` via `STTUpdateSettingsFrame`). The service disconnects and reconnects with the new query parameters. Previously, settings updates were silently ignored.
- Send the Cartesia `done` command before closing the websocket for clean session shutdown
- Fix `run_stt` websocket state check to catch `CLOSING` state (not just `CLOSED`), preventing `ConnectionClosedOK` errors during reconnection
- Fix race condition in `_disconnect_websocket` where a concurrent reconnect could have its new websocket reference nulled by a stale `finally` block

## Testing

- Run `examples/update-settings/stt/stt-cartesia.py` and verify transcriptions work after language switch
- Verify no `ConnectionClosedOK` errors in logs during settings update

### DIFF
--- a/src/pipecat/services/cartesia/stt.py
+++ b/src/pipecat/services/cartesia/stt.py
@@ -285,8 +285,8 @@ class CartesiaSTTService(WebsocketSTTService):
         Yields:
             None - transcription results are handled via WebSocket responses.
         """
-        # If the connection is closed, due to timeout, we need to reconnect when the user starts speaking again
-        if not self._websocket or self._websocket.state is State.CLOSED:
+        # If the connection is not open (closed or closing), reconnect
+        if not self._websocket or self._websocket.state is not State.OPEN:
             await self._connect()
 
         await self._websocket.send(audio)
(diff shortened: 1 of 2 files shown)


src/pipecat/services/cartesia/stt.py
```diff
@@ -285,8 +285,8 @@ async def run_stt(self, audio: bytes) -> AsyncGenerator[Frame, None]:
         Yields:
             None - transcription results are handled via WebSocket responses.
         """
-        # If the connection is closed, due to timeout, we need to reconnect when the user starts speaking again
-        if not self._websocket or self._websocket.state is State.CLOSED:
+        # If the connection is not open (closed or closing), reconnect
+        if not self._websocket or self._websocket.state is not State.OPEN:
```
markbackman, line 289: Catches websockets in CLOSING state (not just CLOSED), which can happen when the receive task cancellation triggers a close handshake.

src/pipecat/services/cartesia/stt.py
```diff
@@ -351,14 +350,18 @@ async def _connect_websocket(self):
             await self.push_error(error_msg=f"Unknown error occurred: {e}", exception=e)
 
     async def _disconnect_websocket(self):
+        ws = self._websocket
         try:
-            if self._websocket and self._websocket.state is State.OPEN:
+            if ws and ws.state is State.OPEN:
```
markbackman, line 355: Operates on a local `ws` reference so the `finally` block only nulls `self._websocket` if it hasn't been replaced by a concurrent `_connect` (which can happen when `run_stt` sees the closing websocket and reconnects during the `close()` await).

aconchillo (approved):

PR #4288, opened 2026-04-14
feat: add Replicate image generation service

### LINKED ISSUE #4287: Add Replicate image generation service
## Summary

Add a `ReplicateImageGenService` to Pipecat's image generation integrations, following the same `ImageGenService` contract used by `FalImageGenService` and `GoogleImageGenService`.

## Motivation

Pipecat already supports image generation providers like fal and Google Imagen, but Replicate hosts a large set of image models behind a consistent prediction API. Adding a Replicate-backed image service would make it easier to use models like `black-forest-labs/flux-schnell` inside existing Pipecat image pipelines.

## Proposed scope

- Add `src/pipecat/services/replicate/image.py`
- Support official Replicate models via `owner/name`
- Support versioned models via `owner/name:version`
- Generate `URLImageRawFrame` results from text prompts
- Handle sync prediction responses with polling fallback when output is not immediately available
- Add tests, environment wiring, example usage, and changelog entry

## Notes

- This should mirror the existing service patterns in the repo rather than introduce a new image-service abstraction.
- Runtime image-settings update frames appear to be out of scope for now since Pipecat currently only has LLM/TTS/STT update frames.

bnovik0v:
## Summary

Adds `ReplicateImageGenService`, a new `ImageGenService` integration for Replicate-hosted image generation models.

- Supports official Replicate models via `owner/name`
- Supports versioned models via `owner/name:version`
- Returns generated images as `URLImageRawFrame`
- Handles sync prediction requests with polling fallback when output is not immediately available
- Adds tests, environment wiring, a local example, and a changelog entry

Closes #4287

## Motivation

Pipecat already supports image generation providers like fal and Google Imagen, but Replicate hosts a large catalog of image models behind a consistent prediction API. This adds the missing Replicate path while preserving the existing `ImageGenService` contract used elsewhere in the framework.

## Design

- Mirrors the existing `FalImageGenService` structure instead of adding a new abstraction
- Uses a Replicate-specific `Settings` dataclass with complete store-mode defaults so it passes the framework's settings-completeness checks
- Routes `owner/name` models to the official model prediction endpoint
- Routes `owner/name:version` models to the generic versioned prediction endpoint
- Polls prediction status only when the initial sync response does not yet contain output
- Emits `URLImageRawFrame` on success and `ErrorFrame` on API or download failures
- Keeps runtime image-settings update frames out of scope, matching the current framework surface

## Usage

```python
    imagegen = ReplicateImageGenService(
        aiohttp_session=session,
        api_token=os.getenv("REPLICATE_API_TOKEN"),
        settings=ReplicateImageGenService.Settings(
            model="black-forest-labs/flux-schnell",
            aspect_ratio="1:1",
        ),
    )
```

## Test plan

- [x] Added focused tests for official model requests
- [x] Added focused tests for versioned model requests
- [x] Added focused tests for polling fallback when sync output is delayed
- [x] Added focused tests for non-200 API failures
- [x] Verified `ReplicateImageGenSettings` delta defaults via `tests/test_service_init.py`
- [x] Verified `ReplicateImageGenService` store-mode completeness via `tests/test_service_init.py`
- [x] Ran `uv run ruff check src/pipecat/services/replicate/image.py tests/test_replicate_image.py examples/getting-started/03b-local-still-frame-replicate.py`
- [x] Ran `uv lock`
- [x] Ran `uv run pytest tests/test_replicate_image.py tests/test_service_init.py`


markbackman:
Thanks for submitting! We'd like to request you add Replicate as a community integration:

For new services, this is the best way to get started. This acknowledges that we can only maintain so many services inside Pipecat and helps new services get added as quickly as possible. As we have time and see demand from the community, we can collaborate to move the integration to a service the Pipecat maintainers support directly.

Let me know if you have any questions! (Closing this as community integrations reside in a separate repo.)

bnovik0v:
Thanks for the guidance, @markbackman! Makes total sense

PR #4291, opened 2026-04-14
fix(openai-responses): clear previous_response_id state on system_instruction update

yuki901:
## Summary

`OpenAIResponsesLLMService` did not honor dynamic `system_instruction` updates sent via `LLMUpdateSettingsFrame` on subsequent turns.

## The bug

The Responses API ignores the `instructions` field on any request that also carries `previous_response_id`. `_apply_previous_response_optimization` attaches `previous_response_id` whenever the conversation prefix hash still matches — which is the case when only `system_instruction` (not the messages) has changed.

As a result:

```python
await task.queue_frame(
    LLMUpdateSettingsFrame(
        delta=OpenAIResponsesLLMSettings(system_instruction="New prompt"),
    )
)
```

…would update `self._settings.system_instruction` locally, but the next turn would reuse the server-side conversation state tied to the old instructions. In practice this breaks multi-agent handoffs, language-switching prompts, and any other pattern that changes the system prompt mid-conversation.

The docs explicitly promise this should work:

- https://docs.pipecat.ai/guides/fundamentals/service-settings#updating-settings-at-runtime — *"You can change service settings while the pipeline is running by pushing an update settings frame."*
- https://docs.pipecat.ai/guides/learn/context-management

## The fix

Override `_update_settings` on `OpenAIResponsesLLMService`. When the base class reports that `system_instruction` was among the changed fields, call `_clear_previous_response_state()` so the next turn falls back to sending the full context with fresh instructions.

Only the `system_instruction` change triggers the invalidation; `temperature`, `max_tokens`, etc. do not affect `instructions` and therefore do not clear the optimization state.

## Test plan

- [ ] Update `system_instruction` mid-conversation — next turn uses new prompt
- [ ] Update `temperature` mid-conversation — `previous_response_id` still used (no false-positive invalidation)
- [ ] Multi-agent handoff scenario (two agents with different prompts) — second agent's prompt actually applies

## Related

No open issue for this specific behavior, but the contract is in the docs linked above. Happy to file an issue first if preferred.


yuki901:
Closing this PR. After local verification on `main` without the fix, the Responses API actually does honor updated `instructions` even when `previous_response_id` is attached — `system_instruction` updates via `LLMUpdateSettingsFrame` are reflected on the next turn. The premise of this PR no longer reproduces. Will reopen with a deterministic repro if the issue surfaces again.

PR #4294, opened 2026-04-14
Fix on_assistant_turn_stopped not firing for tool-call-only responses

### LINKED ISSUE #4292: 0 completion tokens pipeline bug
### pipecat version

0.0.102

### Python version

3.12.3

### Operating System

Ubuntu 24.04

### Issue description

In rare occasions the LLM returns a response without any text content or function calls. I gathered the logs and traces for Google and OpenAI below. Because the response contains neither text or function call, the pipeline does not continue until the user intercepts.

### Reproduction steps

1. Standard pipecat installation
2. Override the OpenAILLService class to return empty content

### Expected behavior

I believe the best way to deal with this is to run an internal retry.

### Actual behavior

The pipeline does not continue until the user intercepts.

### Logs

```shell
21:03:04.298 | DEBUG    | pipecat.processors.metrics.frame_processor_metrics:start_llm_usage_metrics:173 | 47be5310-c4ad-4132-b18c-221acd0c7c98 - GoogleLLMService#1 prompt tokens: 4337, completion tokens: 0
```

It also occurs in OpenAI gpt4.1. I was able to capture the trace below.

Image

See `-> 0 completion`

aconchillo:
## Summary

- `on_assistant_turn_stopped` now always fires at the end of an assistant turn, even when the LLM returned zero text tokens. Previously it was gated behind `push_aggregation()` returning a non-empty string, so consumers waiting for this event would never hear back.
- Updated `AssistantTurnStoppedMessage.content` docstring to clarify it may be empty.

## Fixes

- Fixes #4292

### DIFF
--- a/changelog/4294.fixed.md
+++ b/changelog/4294.fixed.md
@@ -1,0 +1,1 @@
+- Fixed `on_assistant_turn_stopped` not resetting internal state when the LLM returned no text tokens. Added `interrupted` field to `AssistantTurnStoppedMessage` to indicate whether the assistant turn was interrupted.
--- a/src/pipecat/processors/aggregators/llm_response_universal.py
+++ b/src/pipecat/processors/aggregators/llm_response_universal.py
@@ -209,12 +209,16 @@ class AssistantTurnStoppedMessage:
     content. This is the aggregated transcript that is then used in the context.
 
     Parameters:
-        content: The message content/text.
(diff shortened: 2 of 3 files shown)


markbackman (approved):
Makes sense! LGTM.

Approving, assuming test issues are resolved.

Issue #4165, opened 2026-03-27
Orphan function call entries block pipecat-flows transitions after user interruption


OlliTapio:
## Bug Description

When a user interrupts the bot after `FunctionCallsStartedFrame` is emitted but before `FunctionCallInProgressFrame` arrives, no `FunctionCallCancelFrame` is emitted. This leaves an orphan entry in `_function_calls_in_progress` (with value `None`) that is never cleaned up.

This orphan blocks pipecat-flows transitions: `_check_and_execute_transition` checks `has_function_calls_in_progress`, which returns `True` because of the orphan entry, so the transition callback never fires and the flow gets permanently stuck.

## Steps to Reproduce

1. Set up a pipecat-flows application with an edge function that triggers a node transition
2. Have the LLM call the function → `FunctionCallsStartedFrame` is emitted, entry added to `_function_calls_in_progress` with value `None`
3. User interrupts (speaks over the bot) before `FunctionCallInProgressFrame` arrives
4. No `FunctionCallCancelFrame` is emitted for the orphan
5. LLM retries on the next turn and calls the same function again
6. The retry completes successfully (`FunctionCallsStartedFrame` → `FunctionCallInProgressFrame` → `FunctionCallResultFrame`)
7. `_check_and_execute_transition` runs but `has_function_calls_in_progress` is `True` because the orphan from step 2 is still there
8. **The node transition never fires** — the flow is stuck

## Minimal Reproduction

```python
import asyncio
from pipecat.frames.frames import (
    FunctionCallFromLLM,
    FunctionCallInProgressFrame,
    FunctionCallResultFrame,
    FunctionCallsStartedFrame,
    InterruptionFrame,
)
from pipecat.processors.aggregators.llm_context import LLMContext
from pipecat.processors.aggregators.llm_response_universal import LLMAssistantAggregator

async def reproduce():
    context = LLMContext()
    aggregator = LLMAssistantAggregator(context)

    # Turn 1: FunctionCallsStartedFrame fires, then user interrupts
    call_1 = FunctionCallFromLLM(
        context=context,
        tool_call_id="call_1",
        function_name="save_notes",
        arguments={},
    )
    # Simulate: FunctionCallsStartedFrame adds orphan entry
    aggregator._function_calls_in_progress[call_1.tool_call_id] = None
    
    # Simulate: InterruptionFrame arrives — no FunctionCallCancelFrame emitted
    # aggregator._function_calls_in_progress still has {"call_1": None}
    
[...]

Issue #3661, opened 2026-02-06
Function calls get stuck when cancelled before FunctionCallInProgressFrame arrives


OlliTapio:
### pipecat version

0.99

### Python version

3.12.12

### Operating System

MacOS 26.2

### Issue description

When we manage to interrupt the function call before the FunctionCallInProgressFrame has been sent, the function call gets stuck IN_PROGRESS and does not ever complete. In Pipecat Flows this stops node transitions and also makes the LLM more hesitant to call the function.

### Reproduction steps

Easiest to reproduce when using LLM that can send text and function call in same response. 

Instruct LLM to create content and tool call. Interrupt the speach when it has been generated. 

Interruption frame is sent before the FunctionCallInProgressFrame. The function gets stuck in progress.

Seems like the FunctionCallInProgressFrame is sent after all of the text frames have arrived to TTS. 

### Expected behavior

Either cancel the function or let it complete. Not leave it in_progress.

### Actual behavior

Function is stuck IN_PROGRESS. Flows node transitions are blocked.

### Logs

```shell

```

OlliTapio:
Attempted fix. Works locally, but is cancels all functions even if the interruption is set to false:
https://github.com/pipecat-ai/pipecat/pull/3647

OlliTapio:
Closing — we've worked around this on our side.

PR #4301, opened 2026-04-14
Fail missing tool calls cleanly

### LINKED ISSUE #4300: Missing tool handlers can leave tool-call state hanging
## Problem
If the model calls a tool that is not registered, Pipecat logs a warning and skips it.

That sounds harmless, but by that point Pipecat may already have told other parts of the system that a tool call started.

So the tool call can start without ever getting a matching finish/error result.

## Example
The model calls `lookup_customer`, but no handler was registered.

Current behavior:
- Pipecat logs that the tool is not registered
- no terminal tool result is sent
- parts of the system can stay stuck as if a tool is still running
- the turn can go silent

Expected behavior:
- Pipecat should fail the tool call cleanly
- it should emit a normal final tool result, for example:
  - `Error: function 'lookup_customer' is not registered.`
- downstream state should be able to clean up normally
- the model should be able to continue the conversation

## Why this matters
The main issue is lifecycle consistency.

If Pipecat starts a tool call, it should always end it with a normal terminal event.

bnovik0v:
## Summary
Fixes missing tool handlers so they fail cleanly instead of leaving tool-call state hanging.

Before this change:
- the model could call an unregistered tool
- Pipecat would log a warning and skip it
- the tool call might never get a normal final result
- the turn could go silent and downstream state could stay stuck

After this change:
- missing tools still produce a normal final tool result
- Pipecat returns a clear error message
- downstream state can clean up normally
- the model can continue the conversation

## Example
If the model calls `lookup_customer` but no handler is registered:

Before:
- Pipecat logs a warning
- no terminal tool result is sent
- parts of the system may still think a tool is running

After:
- Pipecat sends a normal final tool result:
  - `Error: function 'lookup_customer' is not registered.`

## Testing
- `uv run pytest tests/test_llm_service.py tests/test_user_mute_strategy.py tests/test_user_idle_controller.py`

Closes #4300

## Related
Once a tool call starts, it should always reach a terminal state. Several issues have hit variants of this invariant being broken:
- #4165 — orphan function call entries block transitions after interruption
- #3661 — function calls stuck when cancelled before InProgressFrame arrives
- #4260 — TTS queue drops function call frames during interruption, causing permanent silence

This PR closes another gap: when no handler is registered (e.g. model hallucinates a tool name), the call now still gets a proper terminal result.

### DIFF
--- a/src/pipecat/services/llm_service.py
+++ b/src/pipecat/services/llm_service.py
@@ -725,10 +725,8 @@ class LLMService(UserTurnCompletionLLMServiceMixin, AIService):
                 logger.warning(
                     f"{self} is calling '{function_call.function_name}', but it's not registered."
                 )
-                item = FunctionCallRegistryItem(
-                    function_name=function_call.function_name,
-                    handler=self._missing_function_call_handler,
-                    cancel_on_interruption=True,
+                item = self._build_missing_function_call_registry_item(
+                    function_call.function_name
                 )
 
             runner_items.append(
@@ -786,7 +784,21 @@ class LLMService(UserTurnCompletionLLMServiceMixin, AIService):
             await self._sequential_runner_queue.put(runner_item)
 
     async def _run_function_call(self, runner_item: FunctionCallRunnerItem):
-        item = runner_item.registry_item
+        # Re-resolve the registry item at execution time. The function may have
+        # been unregistered between queuing and execution, in which case we
+        # fall back to the missing-function handler so the call still terminates
+        # with a normal tool result.
+        if runner_item.function_name in self._functions.keys():
+            item = self._functions[runner_item.function_name]
+        elif None in self._functions.keys():
+            item = self._functions[None]
+        elif runner_item.registry_item.handler is self._missing_function_call_handler:
+            item = runner_item.registry_item
+        else:
+            logger.warning(
+                f"{self} is calling '{runner_item.function_name}', but it was just unregistered."
+            )
+            item = self._build_missing_function_call_registry_item(runner_item.function_name)
 
         logger.debug(
             f"{self} Calling function [{runner_item.function_name}:{runner_item.tool_call_id}] with arguments {runner_item.arguments}"
@@ -892,6 +904,16 @@ class LLMService(UserTurnCompletionLLMServiceMixin, AIService):
         finally:
             if timeout_task and not timeout_task.done():
                 await self.cancel_task(timeout_task)
+
+    def _build_missing_function_call_registry_item(
+        self, function_name: str
+    ) -> FunctionCallRegistryItem:
+        """Build a registry item that routes to the missing-function handler."""
+        return FunctionCallRegistryItem(
+            function_name=function_name,
+            handler=self._missing_function_call_handler,
+            cancel_on_interruption=True,
+        )
 
     async def _missing_function_call_handler(self, params: FunctionCallParams):
         """Return a terminal tool result when the LLM calls an unknown function."""
--- a/src/pipecat/services/llm_service.py
+++ b/src/pipecat/services/llm_service.py
@@ -725,7 +725,11 @@ class LLMService(UserTurnCompletionLLMServiceMixin, AIService):
                 logger.warning(
                     f"{self} is calling '{function_call.function_name}', but it's not registered."
                 )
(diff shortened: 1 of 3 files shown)


src/pipecat/services/llm_service.py
```diff
@@ -725,7 +725,11 @@ async def run_function_calls(self, function_calls: Sequence[FunctionCallFromLLM]
                 logger.warning(
```
kompfner, line 725: Here I think we'll eventually want to handle the developer error case (forgetting to register a function) differently than the LLM hallucination case, but that can be done in a future change (not sure yet exactly how that will look, but see [my current thinking here](https://github.com/pipecat-ai/pipecat/issues/4300#issuecomment-4322875600)). The code as it existed here before this PR was focused on the former.

src/pipecat/services/llm_service.py
```diff
@@ -782,12 +786,7 @@ async def _run_sequential_function_calls(self, runner_items: Sequence[FunctionCa
             await self._sequential_runner_queue.put(runner_item)
 
     async def _run_function_call(self, runner_item: FunctionCallRunnerItem):
-        if runner_item.function_name in self._functions.keys():
-            item = self._functions[runner_item.function_name]
-        elif None in self._functions.keys():
-            item = self._functions[None]
-        else:
-            return
+        item = runner_item.registry_item
```
kompfner: Hmm, I'm guessing the reason we were previously re-looking-up the `item` here rather than just relying on `runner_item.registry_item` is just in case the item was *just* deleted from the registry (`self._functions`), in between when the function call was queued up (in `run_function_calls`) and now, here. If so, we should be careful not to introduce a regression.

Maybe the logic here should mostly still mirror what we have in `run_function_calls`, with a small adjustment, i.e.

```python
# Same logic as run_function_calls lookup
if runner_item.function_name in self._functions.keys():
    item = self._functions[runner_item.function_name]
elif None in self._functions.keys():
    item = self._functions[None]
else:
    # Here's where logic differs (the "not found" case)
    if item.handler == self._missing_function_call_handler:
        # This item already represents a missing function call, use it
        item = runner_item.registry_item
    else:
        # We need to create the missing function call item just in time (the function must've *just* been unregistered)
        logger.warning(
            f"{self} is calling '{function_call.function_name}', but it was just unregistered."
        )
        item = FunctionCallRegistryItem(
            function_name=runner_item.function_name,
            handler=self._missing_function_call_handler,
            cancel_on_interruption=True,
        )
```

kompfner: And then, since we'd be creating the "missing function" registry item in two places, it could be worth factoring that logic out into a helper.

src/pipecat/services/llm_service.py
kompfner: ^ To be clear, there's nothing actionable in the above comment, just making a note :)

kompfner (commented):
Looks good, and works well! Just the one change needed to continue handling the edge case we were previously trying to handle.

src/pipecat/services/llm_service.py
bnovik0v: thanks for the catch! pushed a fix

src/pipecat/services/llm_service.py
```diff
@@ -782,12 +784,21 @@ async def _run_sequential_function_calls(self, runner_items: Sequence[FunctionCa
...
+        # Re-resolve the registry item at execution time. The function may have
+        # been unregistered between queuing and execution, in which case we
+        # fall back to the missing-function handler so the call still terminates
+        # with a normal tool result.
         if runner_item.function_name in self._functions.keys():
             item = self._functions[runner_item.function_name]
         elif None in self._functions.keys():
             item = self._functions[None]
+        elif runner_item.registry_item.handler is self._missing_function_call_handler:
+            item = runner_item.registry_item
```
kompfner, line 794: Weird, in my testing we maybe don't seem to be hitting this branch? I see in my logs *both* warnings ("it's not registered" and "was just unregistered"), which would be confusing to the user...

bnovik0v: My bad, used `is` instead of `==`

kompfner (approved):
Looks good! Thanks for the contribution!

PR #4303, opened 2026-04-14
Improve interruption recovery and MCP tool errors

### LINKED ISSUE #4302: Interrupt recovery and MCP tool errors should be cleaner
## Problem
There are two related cleanup problems around tool calls.

### 1. Interruption does not fully clear stale tool-call state
If a bad tool-call lifecycle leaves stale `tool still running` state behind, interrupting the conversation does not fully clear that assistant-side state.

Example:
- a tool call gets stuck
- the user interrupts to recover
- assistant text state resets
- but Pipecat can still remember an old tool call as in progress

Expected behavior:
- interruption/reset should clear assistant-local tool-call tracking so recovery is clean

### 2. MCP failures lose the real error message
When an MCP tool call fails, Pipecat logs the real error but sends back only a generic message.

Example:
- MCP call fails with `upstream unavailable`
- current result is only `Sorry, could not call the mcp tool`

Expected behavior:
- Pipecat should return a useful tool result such as:
  - `Error calling mcp tool demo_tool: upstream unavailable`

## Why this matters
Both problems make recovery harder:
- interruption is less reliable than it should be
- the model loses the actual reason an MCP tool failed

bnovik0v:
## Summary
Fixes two recovery-path problems around tool calls.

Before this change:
- interruption could leave stale assistant-side tool-call state behind
- MCP failures could lose the real error and return only a generic message

After this change:
- interruption/reset clears stale assistant-local tool-call tracking
- MCP failures return the real error message through the tool result path

## Examples
If a bad tool call leaves stale `tool still running` state behind:
- before: interrupting the conversation may still leave that stale state in memory
- after: interruption clears it

If an MCP tool fails with `upstream unavailable`:
- before: the model gets `Sorry, could not call the mcp tool`
- after: the model gets `Error calling mcp tool demo_tool: upstream unavailable`

## Testing
- `uv run pytest tests/test_context_aggregators_universal.py tests/test_mcp_service.py tests/test_llm_service.py tests/test_user_mute_strategy.py tests/test_user_idle_controller.py`

Closes #4302

### DIFF
--- a/changelog/4303.fixed.md
+++ b/changelog/4303.fixed.md
@@ -1,0 +1,1 @@
+- Fixed tool-call recovery by clearing stale assistant-side tool state on interruption and returning the original MCP tool error message instead of a generic fallback.
--- a/src/pipecat/processors/aggregators/llm_response_universal.py
+++ b/src/pipecat/processors/aggregators/llm_response_universal.py
@@ -897,6 +897,8 @@ class LLMAssistantAggregator(LLMContextAggregator):
         """Reset the aggregation state."""
         await super().reset()
         await self._reset_thought_aggregation()  # Just to be safe
+        self._function_calls_in_progress = {}
+        self._function_calls_image_results = {}
         self._push_context_on_bot_stopped_speaking = False
 
     async def _reset_thought_aggregation(self):
--- a/src/pipecat/services/mcp_service.py
+++ b/src/pipecat/services/mcp_service.py
@@ -242,6 +242,8 @@ class MCPClient(BaseObject):
         except Exception as e:
             error_msg = f"Error calling mcp tool {function_name}: {str(e)}"
             logger.error(error_msg)
+            await result_callback(error_msg)
+            return
 
         response = ""
         if results:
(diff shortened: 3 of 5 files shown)


stackdumper:
Thanks for this one. It covers both parts of #4302: the stale tool call state in the aggregator, and the fixed error text in `MCPClient`.

I opened #5734 for the `MCPClient` part only. It returns the error the tool call raised, and also drops the MCP session the call found dead, so the next call connects again. Your aggregator change is the other part, and #5734 does not touch it.

Notes from a cherry-pick of this branch onto current `main` (f67c18a), in case a rebase is on your list:

- `main` split `_call_tool` into `_call_tool` and `_call_tool_text`. The mcp hunk auto-merges into `_call_tool_text`, where `result_callback` is not in scope, so each failed tool call raises `NameError: name 'result_callback' is not defined`. `tests/test_mcp_service.py` on `main` does not catch it: 33 passed.
- The aggregator hunk fails 2 tests that pass on `main`: `test_default_user_turn_strategies` and `test_async_function_call_after_model_response_is_deferred`. `tests/test_context_aggregators_universal.py` gives 2 failed, 99 passed; stock gives 101 passed.

stackdumper:
Update: #5734 now carries the error text fix alone. The dead-session drop moved to #5778.

PR #4304, opened 2026-04-14
Add missing daily-python dependency for tavus extra

markbackman:
## Summary

- Fixed `pipecat-ai[tavus]` not installing the required `daily-python` dependency
- The Tavus service and transport both import from `daily.daily`, so the `tavus` extra now correctly depends on `pipecat-ai[daily]`


filipi87 (approved):
LGTM!

PR #4306, opened 2026-04-14
fix(azure-tts): Route completion through word boundary queue to prevent last word from being missed

YFortin:
## Problem

The last word spoken by the Azure TTS service is intermittently missed by observers (RTVI, custom observers) and the RTVI playground UI (word stays gray instead of turning white).

### Root Cause

In `AzureTTSService._handle_completed()`, the audio stream completion signal (`None`) was put directly into `_audio_queue`, while the last word was still pending in `_word_boundary_queue` waiting to be processed by `_word_processor_task_handler`.

This caused a race condition:

1. Azure SDK fires `synthesis_completed` → `_handle_completed` flushes last word to `_word_boundary_queue` and puts `None` in `_audio_queue`
2. `run_tts` sees `None` in `_audio_queue` and returns
3. The caller proceeds to emit `TTSStoppedFrame` via audio context
4. `_word_processor_task_handler` finally processes the last word and calls `add_word_timestamps` → `TTSTextFrame` is emitted **after** `TTSStoppedFrame`

Since observers and the RTVI UI expect all `TTSTextFrame`s to arrive before `TTSStoppedFrame`, the last word appears unspoken.

## Fix

Route the completion signal through `_word_boundary_queue` as a `None` sentinel instead of putting it directly in `_audio_queue`. The word processor task recognizes this sentinel and only signals `_audio_queue` after all pending words have been drained.

This guarantees the ordering: `TTSTextFrame` (last word) → `TTSStoppedFrame`.

The cancellation/interruption path (`_handle_canceled`) is unchanged — it still signals `_audio_queue` directly, which is correct since word ordering doesn't matter when speech is interrupted.

## Changes

- `_handle_completed`: Puts `None` in `_word_boundary_queue` instead of `_audio_queue`
- `_word_processor_task_handler`: Handles `None` sentinel by forwarding completion to `_audio_queue`

## Testing

- Existing `test_tts_frame_ordering.py` tests pass (4/4)
- Verified the fix addresses the issue observed in the playground UI where the last word stays gray even though it has been said by the bot.


markbackman (approved):
LGTM! Nice fix 👏

PR #4309, opened 2026-04-15
feat(observers): add LatencyWatchdogObserver for proactive latency monitoring

yro7:
this is my first PR to an open-source project. Apologies in advance if any inconvenience, I needed this feature myself, so I figured others might too. I'll be happy to iterate on feedback.

## Summary

Adds `LatencyWatchdogObserver` — a **proactive** latency monitor. Unlike the existing metrics (`TTFBMetricsData`, `ProcessingMetricsData`) which report *after* a slow operation completes, the watchdog fires callbacks **while** the operation is still in flight, as soon as a threshold is crossed.

## Why

Reactive metrics are great for dashboards, useless for real-time UX mitigations. Concrete use case: an LLM takes 3s to respond. With reactive metrics, you learn this *after* the user has already been staring at silence. With the watchdog, you trigger a TTS filler ("Still working on it...") at 1s, or swap to a fallback LLM at 3s — while the slow call is still pending.

## API

```python
watchdog = LatencyWatchdogObserver()

@watchdog.subscribe(
    start_processor=LLMService,       # optional — class filter
    start_frame=LLMContextFrame,      # required — frame that starts the timer
    stop_frame=TTFBMetricsData,       # optional — defaults to TTFBMetricsData
    stop_processor=...,               # optional — smart default = start_processor
    threshold_secs=1.0,               # required
)
async def on_slow_llm(processor_name, elapsed_secs):
    ...
```

Three modes emerge naturally from how you combine the args:
- **Per-processor** (`start_processor == stop_processor`): slow LLM TTFB, slow TTS first audio, etc.
- **Spanning** (`start_processor != stop_processor`): end-to-end latency across services (e.g. user-speech → LLM-done).
- **Global** (no `start_processor`): matches any processor; useful for pipeline-wide SLOs.

## Design notes (for reviewers)

- **Smart default** `stop_processor ← start_processor` is *not* a coupling to specific services — it just means "by default, same processor for start and stop". You override it for spanning.
- **Handler errors are isolated**: if a subscriber's callback raises, we log and continue — other subscriptions keep working. See `_fire()`.
- **Keying** `(sub_idx, proc_id)`: `proc_id=0` for spanning/global (single timer per subscription), `id(proc)` for per-processor mode (parallel timers across multiple instances of the same class).

## Testing

- 11 unit tests in `tests/test_latency_watchdog_observer.py` — cover per-processor, spanning, global, no-reset, handler-throws-continues, concurrent processors.
- New runnable example at `examples/observability/observability-latency-watchdog.py` centralizing the 3 modes in a `register_latency_observers()` helper.
- `ruff check` ✅, `ruff format --check` ✅.

## Related Issues

None.


markbackman:
Thanks for offering this PR. I don't think we want to add this feature to Pipecat at this time. This feels more closely aligned with an application concern, which developers can add. Thanks again for submitting!

PR #4311, opened 2026-04-15
New approach to reconnect STT services after updating settings.

filipi87:
## Summary                                                                                                              
- STT services now reconnect safely when settings are updated at runtime: if the user is currently speaking, the reconnect is deferred until `UserStoppedSpeakingFrame` is received, preventing mid-speech disruption                    
- Audio frames received while a reconnect is in progress are buffered and replayed once the new connection is ready, preventing audio loss                                                                                                   
- Added `_request_reconnect()`, `_reconnect()`, and `_do_reconnect()` hooks to `STTService` / `WebsocketSTTService` so subclasses can participate in the reconnect lifecycle without duplicating the buffering logic                           
- `CartesiaSTTService` and `DeepgramSTTService` updated to use the new reconnect mechanism instead of calling `_disconnect()` / `_connect()` directly on settings change
- `DeepgramSTTService` adds a `_connection_ready` event so `_do_reconnect()` can wait until the async WebSocket handshake completes before replaying buffered audio


src/pipecat/services/stt_service.py
```diff
@@ -163,6 +164,16 @@ def __init__(
...
 
+        # VAD-aware reconnect state
+        # Whether it is safe to reconnect right now (False while the user is speaking).
+        self._can_reconnect: bool = True
+        # Whether a reconnect has been requested but deferred until speaking ends.
+        self._need_reconnect: bool = False
+        # Whether a reconnect cycle is currently in progress.
+        self._reconnecting: bool = False
+        # Audio frames received while _reconnecting is True, replayed after reconnect.
+        self._reconnect_audio_buffer: list[tuple[AudioRawFrame, FrameDirection]] = []
```
filipi87, line 175: We need to add this logic here, inside the parent `STTService`, because of services like `DeepgramSTTService`, which do not extend `WebsocketSTTService`.

aconchillo: I think this is fine.

src/pipecat/services/deepgram/stt.py
```diff
@@ -611,16 +632,15 @@ async def _connection_handler(self):
                     keepalive_task = self.create_task(
                         self._keepalive_handler(), f"{self}::keepalive"
                     )
-                    try:
-                        await connection.start_listening()
-                    finally:
-                        await self.cancel_task(keepalive_task)
+                    await connection.start_listening()
```
aconchillo, line 635: Just saw this. I don't think we need the `except asyncio.CancelledError` this is handled by the task manager already.

aconchillo (approved):

aconchillo:
LGTM! Just added one minor comment about a couple of lines I think we don't need.

src/pipecat/services/deepgram/stt.py
filipi87: Agreed. Removed.

PR #4312, opened 2026-04-15
Improve HeyGen LiveAvatar plugin reliability and performance

dyi1:
- Add keep-alive mechanism: send session.keep_alive every 2.5 min to prevent 5-minute inactivity timeout
- Optimize audio chunking: 600ms first chunk for faster initial response, 1s subsequent chunks for efficient streaming
- Fix audio buffer flush: send remaining buffered audio on utterance end instead of discarding it
- Fix WS state cleanup: properly reset connected/ready state when WebSocket drops unexpectedly
- Add livekit_config passthrough in LiveAvatar session token creation
- Replace stray print() with logger.debug()

- Add WebSocket ready gate: wait for session.state_updated connected event before sending commands (prevents silently dropped messages) -> Not doing this for now. For some reason something is off with the gating so removing for time being.


markbackman:
@filipi87 to take a look

src/pipecat/services/heygen/video.py
```diff
@@ -212,9 +212,11 @@ async def start(self, frame: StartFrame):
             frame: The start frame containing initialization parameters.
         """
         await super().start(frame)
-        # 40 ms of audio, match the default behavior from the output transport
-        self._audio_chunk_size = int((HEY_GEN_SAMPLE_RATE * 2) / 25)
-        await self._client.start(frame, self._audio_chunk_size)
+        # First chunk: 600ms for faster initial response
+        self._first_chunk_size = int(HEY_GEN_SAMPLE_RATE * 2 * 0.6)  # 28800 bytes
```
edwHeygen: first chunk let's do 400ms

src/pipecat/services/heygen/client.py
```diff
@@ -225,22 +226,22 @@ async def cleanup(self) -> None:
...
+                self._ws_ready.clear()
 
             if self._event_task and self._task_manager:
                 await self._task_manager.cancel_task(self._event_task)
                 self._event_task = None
         except Exception as e:
             logger.error(f"Exception during cleanup: {e}")
 
-    async def start(self, frame: StartFrame, audio_chunk_size: int) -> None:
+    async def start(self, frame: StartFrame) -> None:
```
filipi87, line 235: You will also need to update `src/pipecat/transports/heygen/transport.py` to match this new signature.

filipi87 (commented):
Hi @dyi1, thank you for working on this. There are two things we should fix:
- When I try to run the example `examples/video-avatar/video-avatar-heygen-video-service.py`, I consistently get the following error:
    ```
    2026-04-16 09:57:15.097 | ERROR    | pipecat.services.heygen.client:start:257 - HeyGenClient: Timed out waiting for WebSocket session.state_updated connected
    ```
- We need to apply similar fixes to `HeyGenTransport` as you did for `HeyGenVideoService` (see `src/pipecat/transports/heygen/transport.py`)

dyi1:
Thanks @filipi87 . I've updated to fix the issues mentioned. There's something off with the gating for the time being so I'm going to remove it and investigate adding it back in later.

filipi87 (approved):
LGTM! Thank you @dyi1 . 🚀

PR #4314, opened 2026-04-16
Log system instruction once at composition time, not on every LLM call

markbackman:
## Context

I'm making this change because system instructions are often very verbose, which fills up debug logs quickly. Instead, let's separate the system instruction into its own log that's made a composition time. This should help to keep debug logs more concise and readable.

## Summary

- Removed the system instruction from per-call "Generating chat from context" debug logs in Google, Anthropic, and AWS LLM services (OpenAI already omitted it)
- Added a one-time debug log in `LLMService._compose_system_instruction()` so the composed system instruction is logged once when turn completion or async tool cancellation is enabled
- This reduces console noise significantly, especially when `filter_incomplete_user_turns` is enabled (the turn completion prompt is ~80 lines that was previously logged on every LLM call)


markbackman:
Also updating uv.lock.

filipi87 (approved):
Makes sense. 🚀

PR #4319, opened 2026-04-16
Modernize Python typing across the codebase

aconchillo:
## Summary

- Replace `typing.List`, `Dict`, `Tuple`, `Set`, `FrozenSet`, `Type` with built-in equivalents (`list`, `dict`, `tuple`, `set`, `frozenset`, `type`)
- Replace `typing.Optional[X]` with `X | None` (PEP 604)
- Replace `typing.Union[X, Y]` with `X | Y` (PEP 604)
- Move `Mapping`, `Sequence`, `Callable`, `Awaitable`, `MutableMapping`, `MutableSequence`, `Iterator`, `AsyncIterator`, `AsyncGenerator` imports from `typing` to `collections.abc` (PEP 585)
- Remove now-unused `typing` imports
- Add `from __future__ import annotations` to 5 files that use forward-reference strings in `X | "Y"` annotations (required because `|` between a type and a string doesn't work at runtime)

Automated via ruff rules `UP006`, `UP007`, `UP035`, `UP045` targeting Python 3.11+.

## Testing

- Full test suite passes (1193 passed, 1 skipped)
- `uv run ruff check --select UP006,UP007,UP035,UP045 --target-version py311 src/` reports zero violations

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -194,6 +194,7 @@ line-length = 100
 select = [
     "D", # Docstring rules
     "I", # Import rules
+    "UP", # Modern Python
 ]
 ignore = [
     "D105",  # Missing docstring in magic methods (__str__, __repr__, etc.)
--- a/examples/persistent-context/persistent-context-anthropic.py
+++ b/examples/persistent-context/persistent-context-anthropic.py
@@ -84,7 +84,7 @@ async def load_conversation(params: FunctionCallParams):
     filename = params.arguments["filename"]
     logger.debug(f"loading conversation from {filename}")
     try:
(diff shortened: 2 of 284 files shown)


pyproject.toml
```diff
@@ -194,6 +194,7 @@ line-length = 100
 select = [
     "D", # Docstring rules
     "I", # Import rules
+    "UP", # Modern Python
```
markbackman, line 197: This is amazing! I didn't know about this.

markbackman (approved):
So good! Thanks for doing this.

PR #4320, opened 2026-04-16
Fix Smallest AI TTS WebSocket endpoint URL and remove unsupported flush

harshitajain165:
## Summary
- Fix WebSocket base URL from `wss://waves-api.smallest.ai` → `wss://api.smallest.ai` and path prefix from `/api/v1/` → `/waves/v1/` to match the [v4.0.0 API docs](https://docs.smallest.ai/waves/v-4-0-0/documentation/text-to-speech-lightning/quickstart)
- Remove `flush_audio`, `_keepalive_task_handler`, and `_send_keepalive` — the Smallest AI WebSocket API does not support the `{"flush": True}` message type

## Test plan
- [ ] Run `examples/voice/voice-smallest.py -t webrtc -v` and verify WebSocket connects without error
- [ ] Confirm no `unknown message status` warnings in logs
- [ ] Confirm TTS audio plays back correctly

### DIFF
--- a/src/pipecat/services/smallest/tts.py
+++ b/src/pipecat/services/smallest/tts.py
@@ -125,8 +125,8 @@ class SmallestTTSService(InterruptibleTTSService):
         *,
         api_key: str,
         base_url: str = "wss://api.smallest.ai",
-        sample_rate: Optional[int] = None,
-        settings: Optional[Settings] = None,
+        sample_rate: int | None = None,
+        settings: Settings | None = None,
         **kwargs,
     ):
         """Initialize the Smallest AI WebSocket TTS service.
--- a/src/pipecat/services/smallest/tts.py
+++ b/src/pipecat/services/smallest/tts.py
@@ -124,7 +124,7 @@ class SmallestTTSService(InterruptibleTTSService):
         self,
         *,
         api_key: str,
-        base_url: str = "wss://waves-api.smallest.ai",
+        base_url: str = "wss://api.smallest.ai",
         sample_rate: Optional[int] = None,
         settings: Optional[Settings] = None,
         **kwargs,
@@ -216,7 +216,7 @@ class SmallestTTSService(InterruptibleTTSService):
 
     def _build_websocket_url(self) -> str:
         """Build the WebSocket URL from base URL and model."""
-        return f"{self._base_url}/api/v1/{self._settings.model}/get_speech/stream"
+        return f"{self._base_url}/waves/v1/{self._settings.model}/get_speech/stream"
 
     async def start(self, frame: StartFrame):
         """Start the Smallest TTS service.
--- a/src/pipecat/services/smallest/tts.py
+++ b/src/pipecat/services/smallest/tts.py
@@ -10,7 +10,6 @@ This module provides a WebSocket-based integration with Smallest AI's
 Waves API for real-time text-to-speech synthesis.
 """
 
-import asyncio
 import base64
 import json
 from dataclasses import dataclass, field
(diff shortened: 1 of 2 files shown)


src/pipecat/services/smallest/tts.py
```diff
@@ -342,25 +333,6 @@ def _get_websocket(self):
...
-            await asyncio.sleep(KEEPALIVE_INTERVAL)
-            await self._send_keepalive()
-
-    async def _send_keepalive(self):
-        """Send a flush message to keep the connection alive."""
-        if self._websocket and self._websocket.state is State.OPEN:
-            msg = {"flush": True}
-            await self._websocket.send(json.dumps(msg))
-
-    async def flush_audio(self, context_id: str | None = None):
```
markbackman, line 358: `flush_audio()` is used by the base TTSService class in `on_turn_context_completed()`. This should remain.

src/pipecat/services/smallest/tts.py
```diff
@@ -125,9 +124,9 @@ def __init__(
         self,
         *,
         api_key: str,
-        base_url: str = "wss://waves-api.smallest.ai",
-        sample_rate: int | None = None,
-        settings: Settings | None = None,
+        base_url: str = "wss://api.smallest.ai",
+        sample_rate: Optional[int] = None,
```
markbackman: We just added the `UP` linting rule to modernize python types. Can you revert to
```suggestion
        sample_rate: int | None = None,
```

src/pipecat/services/smallest/tts.py
```diff
@@ -125,9 +124,9 @@ def __init__(
...
+        settings: Optional[Settings] = None,
```
markbackman: Same here:
```suggestion
        settings: Settings | None = None,
```

markbackman (commented):
Few comments.

About the keepalive, is there a reason to drop it? It would be preferred to maintain the connection than drop it.

src/pipecat/services/smallest/tts.py
harshitajain165: Done!

src/pipecat/services/smallest/tts.py
harshitajain165: Done

src/pipecat/services/smallest/tts.py
harshitajain165: Thanks for pointing @markbackman. `flush_audio()` has been restored

harshitajain165:
Thanks for pointing to the keepalive @markbackman 

The original keepalive used {"flush": True} which Smallest's API doesn' support — that's why it was removed in the previous commit. I've restored it but changed the message to {"text": " ", "voice_id": ..., "language": ...}, which keeps the connection alive.

src/pipecat/services/smallest/tts.py
```diff
@@ -174,6 +174,10 @@ def can_generate_metrics(self) -> bool:
         """
         return True
 
+    async def flush_audio(self, context_id: str | None = None):
+        """Flush any pending audio data."""
+        logger.trace(f"{self}: flushing audio")
```
markbackman, line 179: Is there a mechanism to flush audio? If so, it should go here.

harshitajain165: Hey @markbackman, we don’t currently have a mechanism to flush audio: so this just logs when a flush is attempted.

markbackman: Ok, this makes sense. We can replace the trace log if/when you add a flush. Thanks.

markbackman (commented):
Looks good.

harshitajain165:
> Looks good.
> 
> Can you add a changelog entry? https://github.com/pipecat-ai/pipecat/blob/main/CONTRIBUTING.md#changelog-entries

Thanks @markbackman, made the changelog entry.

markbackman (approved):
LGTM!

PR #4321, opened 2026-04-16
Remove redundant duplicate type checks in direct_function.py

aconchillo:
## Summary

- After the typing modernization (#4319), `type_hint is dict or type_hint is Dict` became `type_hint is dict or type_hint is dict` (redundant). Same for `list`. Removed the duplicates.


markbackman (approved):
LGTM

PR #4323, opened 2026-04-16
fix: when aggregating by tokens, allow inter-token whitespace once non-whitespace has been sent

dakshdua:
In token-streaming mode, `_push_tts_frames` previously stripped only leading newlines and dropped any pure-whitespace frame. That silently discarded meaningful inter-word whitespace (e.g. a standalone `" "` chunk between `"hello"` and `"world"`), losing prosody cues and any downstream sentence-boundary semantics.

When streaming token-by-token, `TTSService` now tracks whether a non-whitespace character has been sent in the current context. While the flag is false, strip all leading whitespace; once true, let whitespace tokens flow through. Reset the flag on `LLMFullResponseEndFrame`/`EndFrame` and on interruption, and save/restore it around `TTSSpeakFrame` since each utterance is its own context.

Sentence-aggregation mode preserves the existing behavior.

### DIFF
--- a/src/pipecat/services/tts_service.py
+++ b/src/pipecat/services/tts_service.py
@@ -292,6 +292,7 @@ class TTSService(AIService):
         self._processing_text: bool = False
         self._tts_contexts: dict[str, TTSContext] = {}
         self._streamed_text: str = ""
+        self._sent_non_whitespace_in_context: bool = False
         self._text_aggregation_metrics_started: bool = False
 
         # Word timestamp state
@@ -683,6 +684,7 @@ class TTSService(AIService):
 
             # Reset aggregator state
             self._processing_text = False
+            self._sent_non_whitespace_in_context = False
             if isinstance(frame, LLMFullResponseEndFrame):
                 if self._push_text_frames:
                     # Route through the serialization queue so the frame is
@@ -697,6 +699,8 @@ class TTSService(AIService):
         elif isinstance(frame, TTSSpeakFrame):
             # Store if we were processing text or not so we can set it back.
             processing_text = self._processing_text
+            saved_sent_non_whitespace = self._sent_non_whitespace_in_context
+            self._sent_non_whitespace_in_context = False
             # TTSSpeakFrame is independent — temporarily clear the turn context
             # so create_context_id() generates a fresh UUID for this utterance.
             saved_turn_context_id = self._turn_context_id
@@ -717,6 +721,7 @@ class TTSService(AIService):
             # the TTS. We pause to avoid audio overlapping.
             await self._maybe_pause_frame_processing()
             self._turn_context_id = saved_turn_context_id
+            self._sent_non_whitespace_in_context = saved_sent_non_whitespace
             self._processing_text = processing_text
         elif isinstance(frame, TTSUpdateSettingsFrame):
             if frame.service is not None and frame.service is not self:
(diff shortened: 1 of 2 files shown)


markbackman:
@filipi87 can you please take a look?

src/pipecat/services/tts_service.py
```diff
@@ -900,13 +906,22 @@ async def _push_tts_frames(
             await self.push_frame(src_frame)
             return
 
-        # Remove leading newlines only
-        text = text.lstrip("\n")
```
filipi87, line 904: @dakshdua , shouldn't this still be valid for both modes ?

dakshdua: In sentence aggregation mode, this exact behavior is preserved. When streaming tokens, all leading whitespace is stripped by the call to `lstrip` below (including any newlines). However, if we _always_ call `lstrip("\n")` even after sending a non-whitespace chunk, chunks like `["Hello.", "\n", "I am your agent."]` get sent as `Hello.I am your agent.`, which will likely harm prosody. Skipping `lstrip("\n")` ensures that we send `Hello.\nI am your agent.`, which should be played back better.

src/pipecat/services/tts_service.py
```diff
@@ -900,13 +906,22 @@ async def _push_tts_frames(
...
-
-        # Don't send only whitespace. This causes problems for some TTS models. But also don't
-        # strip all whitespace, as whitespace can influence prosody.
-        if not text.strip():
-            return
```
filipi87, line 909: And in this case, I wonder if we could just check whether `self._is_streaming_tokens` is `False` as well.

Basically, at this point we’re not stripping anything, just returning if the text contains only whitespace.

So we don’t need this check when we are streaming tokens. In that case, we should always send what we receive, so no check should be needed.

I guess it should be like this:
```
if not text.strip() and not self._is_streaming_tokens:
            return
```

filipi87: And if this is the case, all the rest of the logic I believe should not be needed.

What do you think ?

dakshdua: I see, is the suggestion that we can simplify and simply send `TextFrame` frames as-is if `self._is_streaming_tokens` is `True`? Makes sense, I had assumed that it was necessary to remove any leading whitespace, but if that's not necessary on your end, then we can simplify.

filipi87: At this point, we were not actually removing whitespace, only checking whether the `text` contained only whitespace and stopping if that was the case. 

But as you mentioned, we shouldn’t do this when streaming tokens.

So yes, the idea is to simplify and just send the text as is when `self._is_streaming_tokens` is `True`. We will need to test it with Cartesia and ElevenLabs, but I believe it will work.

dakshdua: `text.lstrip("\n")` was removing leading newlines, so this PR generalized that to remove all leading whitespace when streaming tokens. Sending leading whitespace as its own chunk will cause errors with Cartesia, but I am unsure about ElevenLabs.

src/pipecat/services/tts_service.py
filipi87: Ok, makes sense. 👍

filipi87 (approved):
LGTM! Thank you for doing this @dakshdua. 🚀

PR #4324, opened 2026-04-17
Add pyright type checking: step 1

markbackman:
## Summary

Enables pyright type checking project-wide using an incremental approach. A `pyrightconfig.json` with an explicit `include` list starts with zero-error modules and expands as modules are fixed. CI enforces the checked set from day one.

### Configuration

- `typeCheckingMode: "basic"` — catches real bugs without requiring exhaustive annotations
- `reportMissingImports: false` — permanent setting since services have optional dependencies
- `exclude` for `*_pb2.py` — generated protobuf files should never be type-checked
- `include` grows over time — each phase adds modules to this list

### Modules added in this PR

| Module | Errors fixed | Notes |
|---|---|---|
| `clocks` | 0 | Already clean |
| `metrics` | 0 | Already clean |
| `transcriptions` | 0 | Already clean |
| `frames` | 0 | Already clean (protobuf excluded) |
| `observers` | 2 | Grouped co-assigned fields into `_StartFrameInfo` dataclass; fixed `str \| None` → `int \| None` on frame ID |
| `extensions` | 2 | Tightened `LLMMessagesUpdateFrame`/`LLMMessagesAppendFrame` from `list[dict]` to `list[LLMContextMessage]` |
| `turns` | 6 | Widened strategy return types to `ProcessFrameResult \| None`; gave mixin a `FrameProcessor` base class |
| `pipeline` | 6 | Used `Sequence[FrameProcessor]` for pipeline/switcher params; added `None` guards |
| `runner` | 10 | Made required params non-optional; used `os.environ[]` for required env vars |

### What's left (future PRs)

Remaining modules ordered by error count:

| Phase | Module | Errors |
|---|---|---|
| 2 | `utils` | ~21 |
| 2 | `serializers` | ~32 |
| 2 | `audio` | ~39 |
| 3 | `processors` | ~64 |
| 3 | `transports` | ~83 |
| 4 | `adapters` | ~295 |
| 5 | `services` | ~1,126 |

Services alone account for 67% of all errors and should be added sub-module by sub-module. The `_settings` override pattern (~89 errors) and `create_client` pattern (~18 errors) need structural fixes with generics before most services can pass.

## Test plan

- [x] `uv run pyright` exits 0 with no errors
- [x] Full test suite passes (1193 tests)
- [x] Verified pyright catches deliberately introduced type errors in checked modules


src/pipecat/turns/user_turn_completion_mixin.py
```diff
@@ -292,7 +292,7 @@ async def _incomplete_timeout_handler(
 
             # Push through pipeline to trigger LLM response
             await self.push_frame(
-                LLMMessagesAppendFrame(messages=[{"role": "system", "content": prompt}])
+                LLMMessagesAppendFrame(messages=[{"role": "developer", "content": prompt}])
```
markbackman, line 295: Found this stray "system" message along the way. This should be a developer message.

changelog/4324.added.md
```diff
@@ -0,0 +1 @@
+- Added incremental `pyright` type checking. A `pyrightconfig.json` at the repo root uses `typeCheckingMode: "basic"` with an explicit `include` list of modules that pass cleanly (`clocks`, `metrics`, `transcriptions`, `frames`, `observers`, `extensions`, `turns`, `pipeline`, `runner`). Remaining modules will be added in subsequent PRs. CI enforces the checked set via `uv run pyright` in the format workflow.
```
filipi87, line 1: This incremental approach makes a lot of sense. 👍

src/pipecat/pipeline/service_switcher.py
```diff
@@ -42,7 +43,7 @@ async def on_service_switched(strategy, service):
             ...
     """
 
-    def __init__(self, services: list[FrameProcessor]):
+    def __init__(self, services: Sequence[FrameProcessor]):
```
filipi87: Was there a particular reason to use `Sequence` instead of `list` here ? 

And we are always casting it back to `list`. So, wouldn't be better just keep it as `list` ?

src/pipecat/runner/utils.py
```diff
@@ -416,9 +416,9 @@ def _get_transport_params(transport_key: str, transport_params: dict[str, Callab
 
 async def _create_telephony_transport(
     websocket: WebSocket,
-    params: Any | None = None,
-    transport_type: str = None,
-    call_data: dict = None,
+    params: Any,
+    transport_type: str,
+    call_data: dict,
```
filipi87, line 421: Just double-checking, we don’t need to support None in these cases?

filipi87 (approved):
I think it’s worth checking my comment in `service_switcher.py`, but otherwise this looks great. Nice improvements. 👏🚀

src/pipecat/runner/utils.py
markbackman: Good callout. I traced all three:

- **`transport_type`** and **`call_data`** both come from `parse_telephony_websocket()` (the only caller, `create_transport` at line 582). That function always returns a non-None `str` for the type (falling back to the literal `"unknown"`, which we then reject with `ValueError`) and always returns a `dict` for call data (empty `{}` on the unknown path). So `None` isn't reachable for those.
- **`params`** comes from `_get_transport_params()`, which calls a user-supplied factory. Technically a user could write `lambda: None`, but since the type is `Any`, pyright wouldn't catch that anyway, and `params.add_wav_header = False` on the next line would immediately raise `AttributeError`.

I removed the stale `if params is None` check since its error message was written for the old `params: Any | None = None` signature and is no longer accurate now that the parameter is required.

src/pipecat/pipeline/service_switcher.py
markbackman: Good question. The reason is variance: `list` is invariant in its type parameter, while `Sequence` is covariant. That matters because of subclasses like `LLMSwitcher`, which takes a narrower `list[LLMService]` and forwards it to the parent:

```python
class LLMSwitcher(ServiceSwitcher[StrategyType]):
    def __init__(self, llms: list[LLMService], ...):
        super().__init__(llms, strategy_type)
```

If `ServiceSwitcher.__init__` took `services: list[FrameProcessor]`, that `super().__init__(llms, ...)` call fails pyright:

> `"list[LLMService]"` is not assignable to `"list[FrameProcessor]"`. Type parameter `_T@list` is invariant... Consider switching from `list` to `Sequence` which is covariant.

(Pyright literally suggests the fix.) `Sequence[FrameProcessor]` is read-only and covariant, so `list[LLMService]` assigns cleanly. It also accepts tuples and other sequence types at no extra cost.

The `self._services = list(services)` is unrelated to the signature: it's just an internal defensive copy so we own a mutable list for indexing (`self._services[next_idx]` in failover) and aren't affected by outside mutation of the caller's list.

Alternatives I considered:
- Widen `LLMSwitcher.llms` to `list[FrameProcessor]` — loses the nice narrow type for users.
- Sprinkle `cast(list[FrameProcessor], llms)` at each `super().__init__` call — ugly, and we'd repeat it for any future switcher subclass.
- Make `ServiceSwitcher`/`ServiceSwitcherStrategy` generic in the service type (e.g. `Generic[T]` with `T = TypeVar("T", bound=FrameProcessor)`). This is the most principled option — it would preserve the concrete service type through the whole class and let us drop the `cast()` calls in `LLMSwitcher.llms`/`active_llm`. But it's a bigger refactor and worth its own PR, not part of the initial pyright rollout.

`Sequence` is the minimum-churn fix that keeps the subclass types clean. Happy to switch if you see a cleaner angle I missed.

filipi87: What if we fix like this:
- https://github.com/pipecat-ai/pipecat/pull/4327/changes

Just  `cast(list[FrameProcessor], llms)` in `llm_switcher.py`.

It looks like has fixed the issue.

What do you think ?

src/pipecat/runner/utils.py
markbackman: Went with your approach via #4327. We can revisit the generic type if we add another subclass.

PR #4326, opened 2026-04-17
Add multilingual support for Deepgram Flux STT

markbackman:
## Summary
- Adds `language_hints: list[Language]` to `DeepgramFluxSTTSettings` for use with the new `flux-general-multi` model. Hints are sent as repeatable `language_hint=<code>` URL params at connect and via the Configure control message when updated mid-stream (detect-then-lock). Warn-and-skip when set on any other model.
- `TranscriptionFrame.language` / `InterimTranscriptionFrame.language` now come from the per-turn `languages` field in Flux's `TurnInfo`, so emitted frames reflect what Flux actually detected.
- Removes reliance on the inherited `STTSettings.language` field in Flux (follows the Soniox pattern).

## Test plan
- [x] `uv run pytest -k flux` — 4 passing (service-init tests covering both WebSocket and SageMaker variants).
- [x] `uv run pytest -k deepgram` — 44 passing.
- [x] `uv run ruff check` + `ruff format --check` on modified files — clean.
- [x] Manual wire sanity: `flux-general-multi` with `[EN, ES, EN_GB]` → URL contains `language_hint=en&language_hint=es` (EN_GB dedupes to `en`); multi with no hints → no hint params; `flux-general-en` with hints → warning logged and hints skipped.
- [x] Live run against Deepgram confirmed by user: `TurnInfo` carries `languages: ["en"]` / `languages_hinted: ["en", "es"]` and frame language tracks detected language per turn.
- [ ] Live mid-stream Configure path exercised via the updated `examples/update-settings/stt/stt-deepgram-flux.py` (narrow hints to `[ES]` after 20s).


src/pipecat/services/deepgram/flux/sagemaker/stt.py
```diff
@@ -112,12 +111,13 @@ def __init__(
         # Initialize default settings
         default_settings = self.Settings(
             model="flux-general-en",
-            language=Language.EN,
```
markbackman, line 115: For the `flux-general-en` model, language wasn't a property, so this isn't really a change.

examples/update-settings/stt/stt-deepgram-flux.py
```diff
@@ -115,10 +125,12 @@ async def on_client_connected(transport, client):
...
         )
 
+        # Detect-then-lock: narrow the hints to a single language mid-stream.
+        # Sent as a Configure message — no reconnect needed.
         await asyncio.sleep(10)
-        logger.info("Updating Deepgram Flux STT settings: language=es")
+        logger.info("Updating Deepgram Flux STT settings: language_hints=[es]")
         await task.queue_frame(
-            STTUpdateSettingsFrame(delta=DeepgramFluxSTTService.Settings(language=Language.ES))
+            STTUpdateSettingsFrame(delta=DeepgramFluxSTTSettings(language_hints=[Language.ES]))
```
filipi87: What is our preferred way to define this, `DeepgramFluxSTTService.Settings` or `DeepgramFluxSTTSettings` ?

src/pipecat/services/deepgram/flux/base.py
```diff
@@ -543,6 +641,7 @@ async def _handle_end_of_turn(self, transcript: str, data: dict[str, Any]):
 
         # Compute the average confidence
         average_confidence = self._calculate_average_confidence(data)
+        detected_language = self._primary_detected_language(data)
```
filipi87, line 648: Before, we were always populating `language=Language.EN`, and here we were using `self._settings.language`, which would always be `English`.

Right now, as far as I understand, `_primary_detected_language` only returns a language when using the `flux-general-multi model`.

Shouldn’t we also make it return English when using the `flux-general-en` model, which is still our default ?

Otherwise, it seems we could receive `None` here instead of a language when using this the default model `flux-general-en`.

markbackman: Good point! I'll make this change.

examples/update-settings/stt/stt-deepgram-flux.py
markbackman: Good catch. Yes, `DeepgramFluxSTTService.Settings` is right.

filipi87 (commented):
LGTM! 🚀

filipi87 (approved):

PR #4327, opened 2026-04-17
Fixing typecheck for service switcher.

filipi87:
Fixing typecheck for service switcher.


markbackman (approved):
Thanks, agreed this is the cleaner call for one subclass. If we add more switchers later we can revisit with a generic refactor.

PR #4328, opened 2026-04-17
fix: re-seed Gemini Live context on reconnect without session resumption

denxxs:
## Problem
When Gemini Live reconnects without session resumption (e.g. connection dropped before 
the server sent a resumption handle), the new session starts with blank context — losing 
all conversation history.

The TODO comment in the existing `elif self._context` branch acknowledges this gap.

## Fix
Resolves the TODO by re-seeding conversation history via `send_client_content` before 
marking the session ready for input. The session resumption path is unchanged.

## Testing
Tested with a multi-node voice agent where system instructions change on node transitions,
triggering reconnects. Context is preserved across reconnects when resumption is unavailable.


denxxs:
@kompfner  — you added the TODO in the elif self._context branch in #4242. This PR resolves it by re-seeding conversation history via send_client_content before marking the session ready. Would appreciate your review. Thanks.

src/pipecat/services/google/gemini_live/llm.py
```diff
@@ -1363,12 +1363,21 @@ async def _handle_session_ready(self, session: AsyncSession):
             self._ready_for_realtime_input = True
         elif self._context:
             # Reconnect without session resumption (e.g. error occurred
-            # before server sent a resumption handle).
-            # TODO: ideally we'd re-send conversation history here via
-            # _create_initial_response(), but that currently doesn't handle
-            # the reconnect case properly. This should be very rare — the
-            # connection would have to drop before we've received our first
-            # session_resumption_handle from the server.
+            # before server sent a resumption handle): re-seed conversation
```
kompfner: I suppose I had been thinking that we'd somehow re-use `_create_initial_response()` in order to avoid duplicating this logic. Is your thinking that it's distinct enough to not necessarily warrant re-use?

Looks like the main distinction here—and the judgment call that you made—is that in this reconnection-without-session-resumption case, we just won't bother triggering the bot into speaking first (as that involves a bunch of other logic).

That seems like a very reasonable simplifying decision to me: this whole reconnection-without-session-resumption should hopefully be quite rare anyway, so in that case having the user start the conversation with "hello?" isn't the end of the world.

It might be worth documenting that decision here real quick. Happy to tag on a commit with that.

changelog/4328.fixed.md
```diff
@@ -0,0 +1,3 @@
+Fixed Gemini Live losing conversation history on WebSocket reconnect. When the session
```
kompfner, line 1: Nit: this makes it sound like we would lose convo history on WebSocket reconnects generally, whereas this would only happen quite rarely, if the reconnect was needed before we got our first session resumption handle. Happy to tweak wording in the commit I'll tag onto this PR.

src/pipecat/services/google/gemini_live/llm.py
```diff
@@ -1363,12 +1363,21 @@ async def _handle_session_ready(self, session: AsyncSession):
...
+            # history so the new session retains full context before
+            # accepting input.
+            try:
```
kompfner: Hmm...in my testing, this isn't working for some reason...the bot is responding as though it has no clue what was in the seeded convo history...investigating now.

kompfner: OK, found the issue. Going to push a fix.

kompfner: Still not quite working great...this might be more complicated than initially thought

denxxs: Thanks for digging into this! Happy to test on my end too if that helps! I've been running this against a multi-node voice agent where system instructions change on each node transition, which consistently triggers the reconnect-without-resumption path. Let me know if you want me to try any specific variation.

denxxs:
Verified on both `gemini-2.5-flash-native-audio-preview-12-2025` and `gemini-3.1-flash-live-preview`. Before this fix: "I don't actually know your name." After: "Sure thing, Dennis!" on both models.

<details>
<summary>Test setup & notes</summary>

**Setup:** the WebRTC realtime example (`examples/realtime/realtime-gemini-live.py`), with a SIGUSR1 handler that nulls the resumption handle and closes the session to force the non-resumption reconnect path.

**Flow:**
1. Connect, tell the bot "my name is Dennis"
2. `kill -USR1 <pid>` to force a reconnect without a resumption handle
3. Speak again and ask it to repeat my name

**What changed vs. the first pass:** `send_client_content(turn_complete=False)` silently drops history when `initial_history_in_client_content=True`, which is why the first seeding attempt didn't actually land. Committing with `turn_complete=True` works, but the server rejects a trailing `model` turn, so we drop any trailing model turns before sending. Gemini 3.x also needs the `send_realtime_input(text=" ")` nudge to trigger inference, mirroring the initial-connection path.

</details>

Hope this helps! @kompfner

denxxs:
Force-pushed to squash a duplicated block and apply `ruff format`. No behavior change from `916e3e2`. Diff is whitespace and cleanup.

src/pipecat/services/google/gemini_live/llm.py
```diff
@@ -1280,23 +1284,42 @@ async def _create_initial_response(self):
...
+            await self.start_ttfb_metrics()
 
         try:
-            await self._session.send_client_content(
-                turns=messages, turn_complete=self._inference_on_context_initialization
-            )
-            # Gemini 3.x wants turn_complete=True, but also won't run inference without a realtime input
-            if self._is_gemini_3 and self._inference_on_context_initialization:
-                await self._session.send_realtime_input(text=" ")
+            if reconnect:
```
kompfner, line 1291: Thanks for the fixes!

You know what, in my testing I was actually finding that Gemini Live wasn't quite working properly on the first connection (let alone on reconnection) when `inference_on_context_initialization=False`. So I think we might not need to bifurcate this logic so much between the `reconnect` case and the first connection case—they _both_ need fixing.

kompfner: Looks like we can't really fix the first connection case...there may be a real Gemini Live 2.5 issue, sadly...(which is hopefully short-lived...perhaps we'll make Gemini 3.1 the default soon...)

I opened an [alternate PR](https://github.com/pipecat-ai/pipecat/pull/4355) that builds on top of the first couple of commits of this one, if you'd like to take a look. Mostly the same, but took a slightly different direction in a few ways:
- Adding a new blank user message rather than removing the latest model message to appease Gemini 2.5's must-end-in-user-message requirement
- Taking advantage of Gemini 3.1's improvements to avoid the assistant having to utter a "recap" message on reconnect (which, as you found out, makes sense for Gemini 2.5)

Also added a bunch of comments for posterity and tweaked the changelog entry a tad.

Let me know what you think! Would love to know if this alternate PR also addresses your needs.

denxxs:
Tested #4355 against the same setup. It addresses and covers our use case for both models. Happy to close #4328. Thanks @kompfner!

kompfner:
Thanks for the contribution! The changes here were rolled up into #4355.

PR #4330, opened 2026-04-17
feat(tts): add includes_inter_frame_spaces flag to word-timestamp API

ianbbqzy:
Some TTS providers return verbatim tokens where spaces and punctuation are already embedded (e.g. Inworld). Joining these with an extra space produces `hello , world` instead of `hello, world`.

**Changes:**
- `TTSService.add_word_timestamps` / `_add_word_timestamps` / `_WordTimestampEntry` gain an opt-in `includes_inter_frame_spaces: bool = False` flag, threaded through to every emitted `TTSTextFrame`. Defaults to `False` — no change for existing services.
- `InworldTTSService` passes `includes_inter_frame_spaces=True` and returns tokens verbatim from `_calculate_word_times`.
- Tests in `test_tts_frame_ordering.py` cover HTTP and WebSocket paths: verbatim text, PTS ordering, text-before-audio ordering, and the Inworld punctuation-token scenario.

### DIFF
--- a/changelog/4330.changed.md
+++ b/changelog/4330.changed.md
@@ -1,0 +1,2 @@
+- added `includes_inter_frame_spaces` to `TTSService` and `WebsocketTTSService`, with default to `false` to preserve existing behaviors.
+- `InworldTTSService` to set `includes_inter_frame_spaces` to `true` to stop inserting spaces between words from server, which already include spaces.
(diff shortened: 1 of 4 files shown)


filipi87 (approved):
Hi @ianbbqzy, these changes look good. 

I have a follow-up [PR](https://github.com/pipecat-ai/pipecat/pull/4354) where I’m making `includes_inter_frame_spaces` `optional` so we can keep the default frame behavior.

I’m also bumping to `small-webrtc-prebuilt` **2.5.0** to fix karaoke highlighting.

Thank you for doing this.

ianbbqzy:
Regarding making `includes_inter_frame_spaces`  optional. From my perspective, it already is because it defaults to false, so the default behavior won't change. But feel free to make it explicit if that aligns with Pipecat's code styles better, thanks!

PR #4332, opened 2026-04-19
fix: handle Claude 4.6+ no-prefill constraint for Anthropic and Bedrock

### LINKED ISSUE #4020: Tool calls fail on Claude 4.6 models
### pipecat version

0.0.105

### Python version

3.11.5

### Operating System

MacOS 26.2

### Issue description

The new Anthropic Sonnet 4.6 does not support assistant prefilling. This causes tool calls to fail with the following error:

```
PipelineTask#1: Something went wrong: ErrorFrame#0(error: Unknown error occurred: Error code: 400 - {'type': 'error', 'error': {'type': 'invalid_request_error', 'message': 'This model does not support assistant message prefill. The conversation must end with a user message.'}, 'request_id': ''}, fatal: False)
```

A workaround for this is to append a blank message to the llm request before it is sent and is the solution the LiveKit team opted for: https://github.com/livekit/agents/pull/4973

### Reproduction steps

Use a tool call with a claud 4.6 model (Opus or Sonnet)

### Expected behavior

No errors and the flow work as expected

### Actual behavior

Pipeline crashes

### Logs

```shell
PipelineTask#1: Something went wrong: ErrorFrame#0(error: Unknown error occurred: Error code: 400 - {'type': 'error', 'error': {'type': 'invalid_request_error', 'message': 'This model does not support assistant message prefill. The conversation must end with a user message.'}, 'request_id': 'req_011CYzCWrrmTBzStvAz2XWKE'}, fatal: False)
```

wirjo:
## Summary

Claude 4.6+ models (Sonnet and Opus) do not support assistant message prefilling — the API rejects requests where the conversation ends with an assistant turn:

```
Error code: 400 - {"type": "error", "error": {"type": "invalid_request_error",
"message": "This model does not support assistant message prefill.
The conversation must end with a user message."}}
```

This can occur during tool call flows when the model streams text (e.g. "Let me check that...") alongside a `tool_use` block. The streamed text is appended to the context as a trailing assistant message after tool results have been added, causing the next inference request to fail.

## Root Cause

In `_process_context()`, the flow is:
1. LLM streams text + `tool_use` blocks
2. `run_function_calls()` executes tools, adds `assistant(tool_calls)` + `user(tool_result)` to context
3. Tool result triggers re-inference (context pushed upstream)
4. `finally:` block fires `LLMFullResponseEndFrame` → `push_aggregation()` adds streamed text as a trailing `assistant` message

The trailing assistant message from step 4 causes Claude 4.6+ to reject the request.

**Why this is hard to reproduce:** If the model returns a bare tool call with no preamble text, `push_aggregation` has nothing to push — no trailing assistant message, no error. The bug only triggers when the model says something *before* calling the tool.

## Changes

- **Adapters:** Add `ensure_last_message_is_user()` static method to both `AnthropicLLMAdapter` and `AWSBedrockLLMAdapter`. When the message list ends with an assistant turn, a minimal `(continue)` user message is appended.
- **LLM Services:** Add `_model_disables_prefill()` to both `AnthropicLLMService` and `AWSBedrockLLMService` to detect Claude 4.6+ model identifiers. Applied in both `_get_llm_invocation_params()` (streaming) and `run_inference()` (non-streaming) code paths.
- **Tests:** 8 unit tests covering the adapter methods for both providers.

For Bedrock, model ID matching uses `in` instead of `startswith` since Bedrock IDs include region prefixes (e.g. `us.anthropic.claude-sonnet-4-6-v1:0`).

## References

- Similar fix shipped by LiveKit: https://github.com/livekit/agents/pull/4973
- Same issue reported across the ecosystem: langchain, strands-agents, crush, oh-my-openagent, opencode

Fixes #4020

### DIFF
--- a/src/pipecat/adapters/services/anthropic_adapter.py
+++ b/src/pipecat/adapters/services/anthropic_adapter.py
@@ -178,6 +178,27 @@ class AnthropicLLMAdapter(BaseLLMAdapter[AnthropicLLMInvocationParams]):
 
         return self.ConvertedMessages(messages=messages, system=system)
 
+    @staticmethod
+    def ensure_last_message_is_user(messages: list[MessageParam]) -> list[MessageParam]:
+        """Ensure the message list does not end with an assistant message.
+
+        Claude 4.6+ models do not support assistant message prefilling — the
+        conversation must end with a user turn.  When the last message has
+        ``role="assistant"``, a minimal user message is appended so that the
+        API request is accepted.
+
+        Args:
+            messages: The converted message list (may be mutated in-place).
+
+        Returns:
+            The same list, possibly with an appended user message.
+        """
+        if messages and messages[-1]["role"] == "assistant":
+            messages.append(
+                {"role": "user", "content": [{"type": "text", "text": "(continue)"}]}
+            )
+        return messages
+
     def _from_universal_context_message(self, message: LLMContextMessage) -> MessageParam:
         if isinstance(message, LLMSpecificMessage):
             return self._from_anthropic_specific_message(message)
--- a/src/pipecat/adapters/services/bedrock_adapter.py
+++ b/src/pipecat/adapters/services/bedrock_adapter.py
@@ -168,6 +168,25 @@ class AWSBedrockLLMAdapter(BaseLLMAdapter[AWSBedrockLLMInvocationParams]):
 
         return self.ConvertedMessages(messages=messages, system=system)
 
+    @staticmethod
+    def ensure_last_message_is_user(messages: list[dict[str, Any]]) -> list[dict[str, Any]]:
+        """Ensure the message list does not end with an assistant message.
+
+        Claude 4.6+ models on Bedrock do not support assistant message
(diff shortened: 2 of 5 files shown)


filipi87:
Hi @wirjo, thanks for this PR! The issue it references (#4020) is already closed, so this looks like it may be stale or superseded by other work.

markbackman:
Update: this PR had the right shape — it was closed only because #4020 lacked a reproduction at the time. We've since found and confirmed the repro (see https://github.com/pipecat-ai/pipecat/issues/5036#issuecomment-4982700149 — the same trailing-assistant 400 also surfaces via `filter_incomplete_user_turns`, #5035/#5036).

Rather than reopen (maintainer edits aren't enabled on the fork branch), we've cherry-picked this PR's commit as the base of #5041 — @wirjo's authorship is preserved in the history — with one design change on top: the model gate is inverted to a frozen allow-list of prefill-*supporting* models, because the deny-list approach was already stale (claude-opus-4-8 also rejects prefill). Thanks @wirjo for the groundwork!

PR #4197, opened 2026-03-30
support added for aws transcribe via pod role

abhishek-kumar41:
This PR contains the change to get AWS credentials via boto3 from pod role. Currently, aws transcribe stt.py supports static credentials or as environment variables.

PR #4333, opened 2026-04-19
feat(aws): add shared credential resolver with boto3 chain fallback for Transcribe STT

wirjo:
## Summary

AWS Transcribe STT previously only supported credentials via explicit parameters or environment variables. Services running with IAM roles (EKS pod roles, IRSA, ECS task roles, EC2 instance profiles) or SSO could not use Transcribe without exporting static credentials.

This PR adds a shared `resolve_credentials()` utility that provides the same credential chain that `aioboto3.Session()` uses for the Bedrock LLM and Polly TTS services.

## Motivation

The Bedrock LLM and Polly TTS services use `aioboto3.Session()` which automatically supports the full boto3 credential provider chain (instance profiles, IRSA, ECS task roles, SSO, credential files, etc.). However, the Transcribe STT service uses presigned WebSocket URLs that require explicit credentials — there was no fallback to the boto3 chain.

This creates an inconsistency: Bedrock and Polly work out-of-the-box on EKS/ECS/EC2 with IAM roles, but Transcribe requires manually exporting `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY`.

### `src/pipecat/services/aws/utils.py`
- Add `AWSCredentials` dataclass for type-safe credential passing
- Add `resolve_credentials()` function with a 3-step fallback chain:
  1. Explicit parameters
  2. Environment variables (`AWS_ACCESS_KEY_ID`, `AWS_SECRET_ACCESS_KEY`, `AWS_SESSION_TOKEN`, `AWS_REGION`)
  3. Default boto3 credential provider chain (instance profiles, IRSA, pod roles, SSO, credential files, etc.)
- boto3 is imported lazily inside the function to avoid hard dependency for services that don't need the fallback
- The boto3 fallback is only attempted when *both* access key and secret key are still unresolved, avoiding replacement of explicitly provided credentials

### `src/pipecat/services/aws/stt.py`
- Update `AWSTranscribeSTTService.__init__()` to use `resolve_credentials()` instead of manual `os.getenv()` calls
- Update docstrings to document the fallback behaviour
- Remove unused `import os`

### `tests/test_aws_credentials.py`
- 7 unit tests covering explicit params, env var fallback, boto3 chain fallback, default region, and edge cases

## Non-Changes

The Bedrock LLM and Polly TTS services already support the full credential chain via `aioboto3.Session()` and are **not** modified. The shared `resolve_credentials()` utility is available for any future AWS service that needs explicit credentials.

## Testing

All tests pass:
```
tests/test_aws_credentials.py: 7 passed
```

Related to #4197 (which takes a similar approach but without the shared utility).

### DIFF
--- a/src/pipecat/services/aws/stt.py
+++ b/src/pipecat/services/aws/stt.py
@@ -11,7 +11,6 @@ speech-to-text transcription with support for multiple languages and audio forma
 """
 
 import json
-import os
 import random
 import string
 from collections.abc import AsyncGenerator
@@ -29,7 +28,7 @@ from pipecat.frames.frames import (
     StartFrame,
     TranscriptionFrame,
 )
-from pipecat.services.aws.utils import build_event_message, decode_event, get_presigned_url
+from pipecat.services.aws.utils import build_event_message, decode_event, get_presigned_url, resolve_credentials
 from pipecat.services.settings import STTSettings
 from pipecat.services.stt_latency import AWS_TRANSCRIBE_TTFS_P99
 from pipecat.services.stt_service import WebsocketSTTService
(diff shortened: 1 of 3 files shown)


markbackman:
Thanks for this change. I rebased it onto current `main`, resolved the conflict in the AWS STT service, and added one small follow-up fix to avoid mixing partial explicit credentials with ambient boto3 credentials.

I pushed the result here for review/merge:
#4416

The branch preserves your commit authorship and includes the original PR changes plus the regression test for the credential fallback behavior.

PR #4334, opened 2026-04-20
Sarvam - VAD parameters configurable on saaras:v3

dhruvladia-sarvam:
## Summary

- Exposes 10 fine-grained VAD (Voice Activity Detection) parameters in the Sarvam STT plugin, gated to the `saaras:v3` model only. These parameters were recently added to the Sarvam SDK (`sarvamai` 0.1.28) and allow callers to tune speech detection behaviour at the server level.
- Bumps the `sarvamai` SDK dependency from `0.1.26` to `0.1.28` to pick up the new VAD parameter support.

## New parameters

All parameters are optional (default `None`, which defers to server defaults) and only accepted when `model="saaras:v3"`. Passing them to `saarika:v2.5` or `saaras:v2.5` raises `ValueError`.

| Parameter | Type | Description |
|---|---|---|
| `positive_speech_threshold` | `float` | VAD probability (0.0–1.0) above which a frame is speech |
| `negative_speech_threshold` | `float` | VAD probability (0.0–1.0) below which a frame is silence |
| `min_speech_frames` | `int` | Min consecutive speech frames to start a segment |
| `first_turn_min_speech_frames` | `int` | Min speech frames for the first user turn |
| `negative_frames_count` | `int` | Silence frames within window to end a segment |
| `negative_frames_window` | `int` | Sliding window size for counting negative frames |
| `start_speech_volume_threshold` | `float` | Volume (dB) below which audio is too quiet |
| `interrupt_min_speech_frames` | `int` | Min speech frames for barge-in/interruption |
| `pre_speech_pad_frames` | `int` | Audio frames to prepend before speech onset |
| `num_initial_ignored_frames` | `int` | Leading frames to skip at connection start |

## Example usage

```python
stt = SarvamSTTService(
    api_key="...",
    settings=SarvamSTTService.Settings(
        model="saaras:v3",
        positive_speech_threshold=0.7,
        negative_speech_threshold=0.3,
        min_speech_frames=3,
        first_turn_min_speech_frames=5,
        negative_frames_count=3,
        negative_frames_window=15,
        start_speech_volume_threshold=-40.0,
        interrupt_min_speech_frames=2,
        pre_speech_pad_frames=3,
        num_initial_ignored_frames=5,
    ),
)
```

## Changes

- **`src/pipecat/services/sarvam/stt.py`** — Added `supports_vad_params` to `ModelConfig` (only `True` for `saaras:v3`). Added 10 VAD fields to `SarvamSTTSettings` and deprecated `InputParams`. Wired through `__init__` (with validation), `_connect` (converted to strings per SDK spec), and `_update_settings` (triggers reconnect on change).
- **`pyproject.toml`** — Bumped `sarvamai` from `0.1.26` to `0.1.28`.

## Test plan

- [x] `uv run pytest tests/test_sarvam_vad_params.py` — 20 passed, 130 subtests passed
- [x] `uv run ruff check src/pipecat/services/sarvam/stt.py` — all checks passed
- [x] `uv run ruff check tests/test_sarvam_vad_params.py` — all checks passed
- [x] Manual integration test with a live `saaras:v3` endpoint to verify parameters reach the server


src/pipecat/services/sarvam/stt.py
```diff
@@ -190,14 +227,44 @@ class InputParams(BaseModel):
             mode: Mode of operation for saaras:v3 models only. Options: transcribe, translate,
                 verbatim, translit, codemix. Defaults to "transcribe" for saaras:v3.
             vad_signals: Enable VAD signals in response. Defaults to None.
-            high_vad_sensitivity: Enable high VAD (Voice Activity Detection) sensitivity. Defaults to None.
+            high_vad_sensitivity: Enable high VAD sensitivity. Defaults to None.
+            positive_speech_threshold: VAD probability threshold (0.0-1.0) above which
```
markbackman: The `InputParams` are deprecated, so I would recommend adding new params to the `SarvamSTTSettings` only, which is the supported way to do things going forward.

src/pipecat/services/sarvam/stt.py
markbackman: Unrelated question, but is `saarika:v2.5` the right default model? Or, should it be `saaras:v3`? This is up to you, so feel free to change.

src/pipecat/services/sarvam/stt.py
markbackman: Remove the related logic below too.

markbackman (commented):
Looks good. Just a few comments.

src/pipecat/services/sarvam/stt.py
dhruvladia-sarvam: Kept the new VAD params on SarvamSTTSettings only and removed them from the deprecated InputParams path. Intentionally trying to preserve backward compatibility for callers already using the deprecated params object. But point noted, I'll make this change shortly post this PR, once I've communicated the same to our users in order to prevent them from facing unexpected errors.

markbackman (approved):
LGTM

PR #4335, opened 2026-04-20
Krisp tt demo tool

realgarik:
Fixes for Krisp VIVA Interrupt Prediction integration.


scripts/krisp/audio_file_utils.py
```diff
@@ -90,6 +90,45 @@ def read_audio_file(input_path: str, verbose: bool = False) -> tuple[np.ndarray,
     return audio_data, sample_rate
 
 
+def resample_audio(
```
filipi87: I believe we are not using this new function anywhere in the scripts. Can we remove it ?

src/pipecat/turns/user_start/krisp_viva_ip_user_turn_start_strategy.py
```diff
@@ -167,7 +168,7 @@ def _ensure_session(self, sample_rate: int):
     def _reset_state(self):
         """Reset speech tracking state for the next candidate interruption."""
         self._speech_active = False
-        self._audio_buffer.clear()
```
filipi87, line 170: Why don’t we need to clear the buffer here anymore? It looks like `_reset_state` is invoked from within `reset`, where the purpose is to reset the strategy to its initial state. 

It feels like we should clear the buffer at this point.

src/pipecat/turns/user_start/krisp_viva_ip_user_turn_start_strategy.py
```diff
@@ -121,6 +121,7 @@ def __init__(
 
         # State tracking
         self._speech_active = False
+        self._vad_flag = False
```
filipi87: It looks like `self._speech_active` and `self._vad_flag` are doing exactly the same job, which is tracking whether the user is speaking or not. I guess we don’t need this new flag here.

src/pipecat/turns/user_start/krisp_viva_ip_user_turn_start_strategy.py
```diff
@@ -43,7 +43,7 @@
     logger.error(
         "In order to use KrispVivaIPUserTurnStartStrategy, you need to install krisp_audio."
     )
-    raise Exception(f"Missing module: {e}")
+    raise ImportError(f"Missing module: {e}") from e
```
filipi87, line 46: Nice catch!

src/pipecat/turns/user_start/krisp_viva_ip_user_turn_start_strategy.py
```diff
@@ -252,9 +260,9 @@ async def _handle_audio(self, frame: InputAudioRawFrame) -> ProcessFrameResult:
         frames = audio_float32.reshape(-1, self._samples_per_frame)
 
         for ip_frame in frames:
-            ip_prob = self._ip_session.process(ip_frame.tolist(), self._speech_active)
+            ip_prob = self._ip_session.process(ip_frame.tolist(), self._vad_flag)
```
filipi87: `self._speech_active` and `self._vad_flag` always have exactly the same state, as far as I can see in the code. I believe we can keep using `self._speech_active` and avoid introducing a new flag.

tests/test_krisp_ip_user_turn_start_strategy.py
filipi87, line 1: The tests will probably need to be updated to reflect the changes above.

scripts/krisp/audio_file_utils.py
pogaram: Yes, I think this function was used in the demo scripts and should be removed.

src/pipecat/turns/user_start/krisp_viva_ip_user_turn_start_strategy.py
pogaram: IMHO it was deleted accidentally. Restored it.

src/pipecat/turns/user_start/krisp_viva_ip_user_turn_start_strategy.py
pogaram: Deleted the self._vad_flag

src/pipecat/turns/user_start/krisp_viva_ip_user_turn_start_strategy.py
pogaram: Restored the self._speech_active

tests/test_krisp_ip_user_turn_start_strategy.py
pogaram: Added some tests.

pogaram (commented):
Fixed last comments

filipi87:
Hi @pogaram, the formatting lint is failing.

If you run `./scripts/fix-ruff-and-typecheck.sh`, it will fix the files with incorrect formatting, so you can push the changes here.

filipi87 (approved):
LGTM!

PR #4337, opened 2026-04-20
Split user-turn stop timeout into independent speech and STT timers

markbackman:
## Context

There was a logic flaw in the initial code that skipped the user_stop_timeout in the case where finalization occurs. This remedies that issue by splitting the timeout into two timeouts, allowing us to keep the finalization logic for the STT timeout.

## Summary

- `SpeechTimeoutUserTurnStopStrategy` previously collapsed two distinct waits into `max(stt_timeout, user_speech_timeout)`, which over-waited for finalizing STT services and, in a legacy code path, could end the turn early.
- Split into two independent timers after VAD stop:
  - `user_speech_timeout`: policy floor, always runs to completion.
  - `stt_timeout`: latency safety net, short-circuited by `TranscriptionFrame(finalized=True)` since STT has signaled it has nothing more to send.
- The no-VAD fallback now waits only `user_speech_timeout` rather than `max(stt_timeout, user_speech_timeout)`. `stt_timeout` is defined relative to VAD stop and has no meaning when no VAD event occurred. This shortens the fallback wait for users who set `stt_timeout` greater than `user_speech_timeout`.

## Test plan

- [x] `uv run pytest tests/test_user_turn_stop_strategy.py -v` — all 24 tests pass, including three new ones
  - `test_finalized_short_circuits_stt_wait` — turn stops at `user_speech_timeout`, not `stt_timeout`, when finalization arrives mid-wait
  - `test_non_finalized_waits_full_stt_timeout` — safety net runs when STT doesn't finalize
  - `test_fallback_uses_only_user_speech_timeout` — no-VAD path ignores `stt_timeout`
- [x] `uv run pytest tests/test_user_turn_processor.py tests/test_user_turn_controller.py tests/test_context_aggregators_universal.py tests/test_langchain.py` — 46 tests pass
- [x] `uv run ruff check` and `uv run ruff format --check` — clean


src/pipecat/turns/user_stop/speech_timeout_user_turn_stop_strategy.py
```diff
@@ -59,8 +62,16 @@ def __init__(self, *, user_speech_timeout: float = 0.6, **kwargs):
         self._vad_user_speaking = False
         self._transcript_finalized = False
         self._vad_stopped_time: float | None = None
-        self._timeout_task: asyncio.Task | None = None
-        self._timeout_expired: bool = False
+
+        # VAD-driven timers and completion flags.
+        self._user_speech_timeout_task: asyncio.Task | None = None
+        self._stt_timeout_task: asyncio.Task | None = None
+        self._user_speech_expired: bool = False
```
kompfner: super-nit: why not `_user_speech_wait_done` to match `_stt_wait_done`

markbackman: Good recommendation. Changed.

src/pipecat/turns/user_stop/speech_timeout_user_turn_stop_strategy.py
```diff
@@ -150,59 +160,98 @@ async def _handle_vad_user_stopped_speaking(self, frame: VADUserStoppedSpeakingF
                     f"user_turn_stop_timeout parameter in the LLMUserAggregatorParams."
                 )
 
-        # Start the timeout task
-        timeout = self._calculate_timeout()
-        self._timeout_task = self.task_manager.create_task(
-            self._timeout_handler(timeout), f"{self}::_timeout_handler"
+        # Any prior fallback timer is superseded by the VAD-driven path.
```
kompfner: I think we might be overcomplicating things with a 3rd timer...(though I might be missing something)

Rather than have a "fallback timer", couldn't we just restart the user speech timer as a fallback? Conceptually, it's still a timer waiting for the user to maybe start speaking again. We just need to start it at a different time in the fallback case since we maybe "missed" the VAD stop event.

markbackman: You're right. I think that simplifies things. I pushed a commit with this change.

src/pipecat/turns/user_stop/speech_timeout_user_turn_stop_strategy.py
```diff
@@ -150,59 +154,85 @@ async def _handle_vad_user_stopped_speaking(self, frame: VADUserStoppedSpeakingF
...
-            # Cancel existing fallback timeout if any
-            if self._timeout_task:
-                await self.task_manager.cancel_task(self._timeout_task)
-            timeout = self._calculate_timeout()
-            self._timeout_task = self.task_manager.create_task(
-                self._timeout_handler(timeout), f"{self}::_timeout_handler"
-            )
-            # Make sure the task is scheduled.
-            await asyncio.sleep(0)
+            self._stt_wait_done = True
```
markbackman, line 201: Artificially setting self._stt_wait_done = True, so that the check in `_maybe_trigger_user_turn_stopped()` reads the state of the two timers:
```
if self._user_speech_wait_done and self._stt_wait_done:
```

We could opt to break up `self._stt_wait_done` into additional state, but this keeps things simple.

kompfner (approved):

PR #4338, opened 2026-04-20
Include examples in type checking

markbackman:
## Summary

- Remove `examples/` from the `pyrightconfig.json` ignore list so example files participate in type checking.
- Fix the resulting type errors across all example files.

## Common fixes

- **Required API keys:** `os.getenv("X")` -> `os.environ["X"]` so the return type is `str` rather than `str | None`, and misconfiguration fails fast with a clear `KeyError`.
- **`LLMContextMessage` unions:** narrow with `isinstance(..., dict)` before dict-style access, since the union includes both the OpenAI-style `TypedDict` and `LLMSpecificMessage` dataclass.
- **Service-specific LLM methods:** `assert isinstance(params.llm, ...)` before calling methods that aren't on the base `LLMService` (e.g. `AWSNovaSonicLLMService.reset_conversation`).
- **Optional frame fields:** guard before use (e.g. `LLMSearchResponseFrame.search_result`, which is `str | None`).

## Test plan

- [x] CI pyright passes on `examples/`
- [x] Spot-check a representative example runs end-to-end (e.g. `voice/voice-soniox.py`, `transcription/transcription-gladia-translation.py`)


markbackman:
Switching up a few patterns in examples so that we don't have to ignore type checking for examples. While we're not applying pyright rules to examples, it's a pain to have type detection totally turned off. This provides a balance where we don't include in checking, but we can still be notified if something is incorrect.

pyrightconfig.json
```diff
@@ -28,7 +25,6 @@
     "src/pipecat/transports",
     "src/pipecat/utils",
     "tests",
-    "examples",
```
kompfner, line 31: Ahh I see...removing something from "ignore" causes type-checking to happen in your editor if the file is open, but leaving it out of "include" still causes it to be avoided when running in the CLI (and therefore in CI).

kompfner (approved):

PR #4348, opened 2026-04-21
Fix type errors in scripts and add to pyright checked set

markbackman:
## Summary

- Fix type errors across `scripts/` and include the directory in pyright checks.
- Replaces `from daily import *` with explicit imports in the Tavus transport test.
- Fixes a genuinely broken call site in `scripts/evals/run-eval.py` where `run_eval` was receiving 3 positional args against a 2-arg signature.

## Test plan

- [x] `uv run pyright` reports 0 errors, 0 warnings
- [x] Pre-commit hooks (ruff lint + format) pass

### DIFF
--- a/pyrightconfig.json
+++ b/pyrightconfig.json
@@ -13,7 +13,8 @@
     "src/pipecat/runner",
     "src/pipecat/tests",
     "src/pipecat/transcriptions",
-    "src/pipecat/turns"
+    "src/pipecat/turns",
+    "src/pipecat/utils"
   ],
   "exclude": ["**/*_pb2.py", "**/__pycache__"],
   "ignore": [
@@ -23,7 +24,6 @@
     "src/pipecat/serializers",
     "src/pipecat/services",
     "src/pipecat/transports",
-    "src/pipecat/utils",
     "tests"
   ],
   "reportMissingImports": false
--- a/pyrightconfig.json
+++ b/pyrightconfig.json
@@ -11,7 +11,8 @@
     "src/pipecat/extensions",
     "src/pipecat/turns",
     "src/pipecat/pipeline",
-    "src/pipecat/runner"
+    "src/pipecat/runner",
+    "scripts"
   ],
   "exclude": ["**/*_pb2.py", "**/__pycache__"],
   "ignore": [
@@ -24,9 +25,7 @@
     "src/pipecat/tests",
     "src/pipecat/transports",
     "src/pipecat/utils",
-    "tests",
-    "scripts",
-    "docs"
+    "tests"
   ],
   "reportMissingImports": false
 }
--- a/pyrightconfig.json
+++ b/pyrightconfig.json
@@ -21,7 +21,6 @@
     "src/pipecat/processors",
     "src/pipecat/serializers",
     "src/pipecat/services",
-    "src/pipecat/sync",
     "src/pipecat/tests",
     "src/pipecat/transports",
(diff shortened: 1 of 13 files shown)


scripts/daily/test_tavus_transport.py
```diff
@@ -33,8 +40,8 @@ def __new__(cls, *args, **kwargs):
     def __init__(self, sample_rate: int):
         super().__init__()
         self._sample_rate = sample_rate
-        self._loop = None
-        self._audio_queue: asyncio.Queue | None = None
+        self._loop = asyncio.new_event_loop()
```
markbackman, line 43: Avoids asserts

scripts/evals/run-eval.py
```diff
@@ -33,7 +33,8 @@ async def main(args: argparse.Namespace):
 
     runner = EvalRunner(examples_dir=script_path, record_audio=args.audio, log_level=log_level)
 
-    await runner.run_eval(script_file, args.prompt, args.eval)
+    eval_config = EvalConfig(prompt=args.prompt, eval=args.eval)
+    await runner.run_eval(script_file, eval_config)
```
markbackman, line 37: run_eval takes 2 positional args, so this fixes a bug.

pyrightconfig.json
```diff
@@ -20,13 +21,10 @@
     "src/pipecat/processors",
     "src/pipecat/serializers",
     "src/pipecat/services",
-    "src/pipecat/sync",
```
markbackman, line 23: module doesn't exist 🤷 

Not sure why it was added.

pyrightconfig.json
```diff
@@ -3,15 +3,17 @@
...
+    "src/pipecat/extensions",
     "src/pipecat/frames",
+    "src/pipecat/metrics",
     "src/pipecat/observers",
-    "src/pipecat/extensions",
-    "src/pipecat/turns",
     "src/pipecat/pipeline",
-    "src/pipecat/runner"
+    "src/pipecat/runner",
+    "src/pipecat/tests",
```
markbackman, line 14: Newly added.

pyrightconfig.json
```diff
@@ -3,15 +3,17 @@
   "pythonVersion": "3.11",
   "pythonPlatform": "All",
   "include": [
+    "scripts",
```
markbackman, line 6: Newly added. Alphabetized the rest.

pyrightconfig.json
```diff
@@ -20,13 +22,9 @@
     "src/pipecat/processors",
     "src/pipecat/serializers",
     "src/pipecat/services",
-    "src/pipecat/sync",
-    "src/pipecat/tests",
     "src/pipecat/transports",
     "src/pipecat/utils",
-    "tests",
-    "scripts",
-    "docs"
```
markbackman, line 29: No type issues, but we don't want to track for types.

pyrightconfig.json
kompfner: I take it everything in this folder already passed pyright checking...doesn't look like it contains much

markbackman: Yes, that's right.

pyrightconfig.json
kompfner: > Not sure why it was added.

It used to exist (there's still an empty folder in my checkout), so I'm guessing that's why

kompfner (approved):

PR #4349, opened 2026-04-21
Fix type errors in serializers and add to pyright checked set

markbackman:
## Summary

- Moves `src/pipecat/serializers` into pyright's `include` list. Narrows `self._params` to each subclass's `InputParams` in exotel, vonage, plivo, twilio, genesys, and telnyx so subclass-only fields type-check.
- In `protobuf.py`, renames the reassigned `frame` local to avoid clobbering its `Frame` type and silences two dynamic attribute accesses on the generated `frames_pb2` module.
- Aligns telnyx and plivo hangup validation with twilio: if `auto_hang_up=True` (the default) but required credentials are missing, `__init__` now raises `ValueError` instead of silently logging a warning at call-end time.

## Test plan

- [ ] `uv run pyright` is clean
- [ ] `uv run pyright src/pipecat/serializers/` is clean
- [ ] `uv run ruff check src/pipecat/serializers/` passes
- [ ] `uv run pytest tests/ -k serializer` (29 tests) passes
- [ ] Manual: construct `PlivoFrameSerializer`/`TelnyxFrameSerializer`/`TwilioFrameSerializer` with `auto_hang_up=True` and no credentials → expect `ValueError` at construction
- [ ] Manual: construct the same serializers with `auto_hang_up=False` and no credentials → expect successful construction


src/pipecat/serializers/plivo.py
```diff
@@ -71,7 +72,24 @@ def __init__(
             auth_token: Plivo auth token (required for auto hang-up).
             params: Configuration parameters.
         """
-        super().__init__(params or PlivoFrameSerializer.InputParams())
+        params = params or PlivoFrameSerializer.InputParams()
+        super().__init__(params)
+        self._params: PlivoFrameSerializer.InputParams = params
+
+        # Validate hangup-related parameters if auto_hang_up is enabled
```
markbackman, line 79: Aligning with Twilio, which validates and notifies in the constructor, not at runtime.

src/pipecat/serializers/plivo.py
```diff
@@ -152,23 +170,11 @@ async def _hang_up_call(self):
...
-                if not auth_token:
-                    missing.append("auth_token")
-
-                logger.warning(
-                    f"Cannot hang up Plivo call: missing required parameters: {', '.join(missing)}"
-                )
-                return
+            # __init__ guarantees these are non-None whenever auto_hang_up is True,
+            # which is the only path that reaches this method.
+            auth_id = cast(str, self._auth_id)
```
markbackman, line 175: Validation happens in the constructor, so we can cast as str. We could also assert, but this feels cleaner.

Open to other options.

src/pipecat/serializers/telnyx.py
```diff
@@ -82,10 +82,26 @@ def __init__(
             api_key: Your Telnyx API key (required for auto hang-up).
             params: Configuration parameters.
         """
+        params = params or TelnyxFrameSerializer.InputParams()
+        super().__init__(params)
+        self._params: TelnyxFrameSerializer.InputParams = params
+
+        # Validate hangup-related parameters if auto_hang_up is enabled
```
markbackman, line 89: Align with Twilio.

src/pipecat/serializers/telnyx.py
```diff
@@ -46,7 +46,7 @@ class TelnyxFrameSerializer(FrameSerializer):
     credentials to be provided.
     """
 
-    class InputParams(BaseModel):
+    class InputParams(FrameSerializer.InputParams):
```
markbackman, line 49: This was a bug.

src/pipecat/serializers/telnyx.py
```diff
@@ -82,10 +82,26 @@ def __init__(
             api_key: Your Telnyx API key (required for auto hang-up).
             params: Configuration parameters.
         """
+        params = params or TelnyxFrameSerializer.InputParams()
+        super().__init__(params)
```
markbackman, line 86: Also a bug given that this was missing.

pyrightconfig.json
```diff
@@ -11,6 +11,7 @@
     "src/pipecat/observers",
     "src/pipecat/pipeline",
     "src/pipecat/runner",
+    "src/pipecat/serializers",
```
kompfner, line 14: this will be superseded by https://github.com/pipecat-ai/pipecat/pull/4350

kompfner (approved):

pyrightconfig.json
markbackman: Indeed. I'll merge this first then clean up in #4350

PR #4350, opened 2026-04-21
Expand pyright coverage to full src/pipecat with per-file ignores

markbackman:
## Summary

- Switches the pyright `include` list to `src/pipecat` so every module (and any new module) is type-checked by default.
- Replaces the six directory-level ignores (`adapters`, `audio`, `processors`, `serializers`, `services`, `transports`) with the 140 specific files that still have type errors.
- Puts 189 previously-untyped files under pyright coverage immediately, and turns the remaining work into a concrete per-file TODO list that can only shrink.

## Why

Wholesale directory ignores hid the fact that ~57% of files in those modules already pass type checking. There was no regression protection for clean files, and no visibility into what actually needs fixing. Per-file ignores flip that: new files default to being checked, contributors can grab individual entries from the ignore list, and base-class fixes (e.g. `llm_service.py`) will cascade and let us drop multiple entries at once.

## Test plan

- [x] `uv run pyright` reports 0 errors, 0 warnings
- [x] CI pyright job passes


kompfner (approved):
Great idea!

PR #4351, opened 2026-04-22
feat(tts): VertexGeminiTTSService — route Gemini TTS through aiplatform.googleapis.com

arnavXgupta:
Rewrites `VertexGeminiTTSService` in `pipecat/services/google/vertex/tts.py` to                                                          
  use `google.genai.Client(vertexai=True)` instead of `google.cloud.texttospeech_v1`.                                                      
                                                                                                                                           
  **Before**: synthesis requests went to `{location}-texttospeech.googleapis.com`
  (regional Cloud TTS endpoint) — the same infrastructure as the non-Vertex
  `GeminiTTSService`, just region-pinned.

  **After**: synthesis requests go to `{location}-aiplatform.googleapis.com` via
  the same genai client pattern already used by `GeminiLiveVertexLLMService`. This
  is the correct Vertex AI platform endpoint and is required for workloads covered
  under a Vertex AI BAA (e.g. HIPAA healthcare).

  ## Changes

  - `VertexGeminiTTSService` no longer extends `GoogleVertexBaseTTSService`
    (which is wired to `texttospeech_v1`). It now extends `TTSService` directly.
  - Auth follows the same `_get_credentials` + token-refresh pattern as
    `GeminiLiveVertexLLMService`: service-account JSON string, file path, or ADC.
  - Streaming via `client.aio.models.stream_generate_content` with
    `response_modalities=["AUDIO"]` and `SpeechConfig`. Audio extracted from
    `response.candidates[0].content.parts[0].inline_data.data` (bytes, SDK
    decodes base64 automatically). Output: PCM 16-bit mono at 24 kHz.
  - Default model changed from `gemini-3.1-flash-tts-preview` to
    `gemini-2.5-flash-preview-tts` — the former is not available on the
    aiplatform endpoint.
  - `VertexTTSService` and `GoogleVertexBaseTTSService` are **untouched** —
    they remain on `texttospeech_v1` which is correct for Chirp/Journey voices.

  ## Why not just use GeminiTTSService?

  `GeminiTTSService` uses `texttospeech.googleapis.com` (global endpoint).
  `VertexGeminiTTSService` is the right class for users who have a Vertex AI
  agreement and need data to stay within that contractual boundary.

### DIFF
--- a/src/pipecat/services/google/vertex/__init__.py
+++ b/src/pipecat/services/google/vertex/__init__.py
@@ -3,3 +3,11 @@
 #
 # SPDX-License-Identifier: BSD 2-Clause License
 #
+
+from pipecat.services.google.vertex.llm import GoogleVertexLLMService, GoogleVertexLLMSettings
+from pipecat.services.google.vertex.tts import (
+    VertexGeminiTTSService,
+    VertexGeminiTTSSettings,
+    VertexTTSService,
+    VertexTTSSettings,
+)
--- a/src/pipecat/services/google/vertex/__init__.py
+++ b/src/pipecat/services/google/vertex/__init__.py
@@ -4,7 +4,6 @@
 # SPDX-License-Identifier: BSD 2-Clause License
 #
 
-from pipecat.services.google.vertex.llm import GoogleVertexLLMService, GoogleVertexLLMSettings
 from pipecat.services.google.vertex.tts import (
     VertexGeminiTTSService,
     VertexGeminiTTSSettings,
--- a/src/pipecat/services/google/vertex/tts.py
+++ b/src/pipecat/services/google/vertex/tts.py
@@ -24,9 +24,9 @@ Both services require a GCP project ID, a region, and service-account credential
 import json
 import os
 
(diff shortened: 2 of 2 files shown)


arnavXgupta:
@markbackman @filipi87 @aconchillo  hey can you guys look into this. Thankyou!

markbackman:
Thanks for putting this together. A few comments first:

- Gemini TTS is too slow for real-time right now. Last I checked, audio generation took 2-3 seconds. If you're seeing different performance, let me know.
- If the goal is keeping data in the US, `GoogleTTSService` already supports this via its `location` parameter. It's a Cloud TTS offering with US regional endpoints, real-time latency, and Chirp 3 voice support.

Is there a reason you specifically need Vertex AI (`aiplatform.googleapis.com`) rather than regional Cloud TTS? If so, I'd suggest narrowing this PR to add a `GoogleVertexTTSService` that subclasses `GoogleTTSService`, overriding just auth and the endpoint. That keeps the Settings, voices, streaming loop, and language mapping shared with the parent.

PR #4352, opened 2026-04-22
More pyright fixes

markbackman:
## Summary

This PR started as a pyright-cleanup pass to chip away at the per-file ignore list in `pyrightconfig.json`, and in the process surfaced a handful of real bugs. The ignore list drops from 135 entries to 116.

### Real fixes worth leading with

- **WebSocket `run_stt` no longer crashes the pipeline on transient send failures** in `AssemblyAISTTService`, `CartesiaSTTService`, `GradiumSTTService`, and `SonioxSTTService`. Each sent audio over the socket without a `try/except`, so a single network hiccup mid-stream raised an uncaught exception through `process_frame` and ended the pipeline. Every other push-based STT (Deepgram, xAI, Azure, Smallest, etc.) already guarded its sends; this PR brings these four in line.
- **Direct-function registration no longer crashes for functions without a docstring.** `DirectFunctionWrapper` passed `inspect.getdoc()`'s result straight to `docstring_parser.parse()`, which raises on `None`.
- **Duplicate `ResponseCreated` class removed** from `xai/realtime/events.py`. Dead code from an earlier refactor; one of the two definitions was shadowed.

### Type-system and structural cleanup

- Widened `run_stt` / `run_tts` to `AsyncGenerator[Frame | None, None]` across the STT/TTS base classes and all 36 subclasses that legitimately `yield None` (results arrive on a separate WebSocket receive channel). The consumer (`process_generator`) already accepted `Frame | None`; this aligns the producer side with reality.
- `ToolsSchema(standard_tools=...)` now accepts any `Sequence` rather than requiring the exact union-typed `list`, so callers can pass a narrower `list[FunctionSchema]` without a covariance error.
- Moved Mistral's three message-shape quirks (tool→assistant insertion, non-initial system→user conversion, trailing-assistant `prefix=True`) into a new `MistralLLMAdapter` on the Perplexity template. Same behavior, but the quirks now live in the adapter layer alongside every other provider's.
- `FrameProcessorMetrics.task_manager` accessed before `setup()` now raises a clear `RuntimeError` instead of an `AttributeError` on a None-derived attribute.
- `SarvamLLMService` rejects a missing/non-string model at init with a clear `ValueError` instead of validating the string `"NOT_GIVEN"` against its supported-models set.
- Replaced asserts with structural fixes: FastAPI's `_monitor_websocket` takes the timeout as a method argument; `SOXRStreamAudioResampler._maybe_initialize_sox_stream` returns the initialized stream; callers no longer need to narrow an optional field outside the init site.
- Abstract `run_image_gen` and `run_vision` now contain a `raise NotImplementedError; yield` marker so they're real async generators (not coroutines returning one) and surface a clear error if a subclass calls through to the base.
- Various smaller fixes: `inspect.getdoc() or ""` coercion, `isinstance` narrowing for `_settings.model` (replacing truthiness-based narrowing that relied on `_NotGiven.__bool__`), and a scoped `# type: ignore[attr-defined]` on the Tkinter GC-reference idiom.

## Testing

- `uv run pytest`
- `uv run pyright` passes with zero errors (down from 135 ignored files to 116)
- `uv run ruff check && uv run ruff format --check` pass
- Smoke-test a WebSocket STT (e.g. Cartesia) surviving a mid-stream socket hiccup
- Smoke-test a Mistral chat completion end-to-end — tool call flow, trailing assistant prefix, non-initial system messages should all still work
- Register a direct function without a docstring and confirm wrapper construction succeeds


src/pipecat/adapters/services/mistral_adapter.py
kompfner, line 1: hehe this isn't just a pyright fix

markbackman: True. Doing clean up on things as I go. I broke everything out commit by commit.

src/pipecat/services/mistral/llm.py
```diff
@@ -92,60 +94,6 @@ def create_client(self, api_key=None, base_url=None, **kwargs):
         logger.debug(f"Creating Mistral client with api {base_url}")
         return super().create_client(api_key, base_url, **kwargs)
 
-    def _apply_mistral_fixups(
```
kompfner, line 95: hooray!

kompfner (approved):

PR #4354, opened 2026-04-22
feat(tts): add includes_inter_frame_spaces flag to word-timestamp API - follow-up

filipi87:
Applying the changes from this [PR](https://github.com/pipecat-ai/pipecat/pull/4330), with some extra cleanup to make `includes_inter_frame_spaces` optional while keeping the default frame behavior.

It also bumps to `small-webrtc-prebuilt` 2.5.0 to fix karaoke highlighting.


markbackman (approved):
Makes sense!

PR #4355, opened 2026-04-22
Re-seed Gemini Live context on reconnect without session resumption

kompfner:
## Summary

Fixes Gemini Live losing conversation history in the (rare) case of a WebSocket reconnect before any session resumption handle has been received. When the session reconnects (e.g. on system instruction change, or on an unexpected connection error early in a session), conversation history is now re-seeded into the new session before it is marked ready for input.

This PR builds on @denxxs's original fix (#4328) and hardens it to work cleanly across Gemini Live 2.5 and Gemini Live 3.x, which have meaningfully different behaviors around seeded context.

## Details

- Unifies the initial-connection and reconnect code paths through a single `_create_initial_response(for_reconnect=...)` method, with a docstring that lays out the behavior for all four cases.
- On Gemini 2.5 reconnect, forces `turn_complete=True` on the seed so the model produces a recap-style response immediately. Without this, the user's next utterance would briefly get a "forgetful" response before recovering on the following turn — jarring mid-conversation. This works around Gemini 2.5's [documented audio-input / history-recall limitation](https://discuss.ai.google.dev/t/audio-input-cannot-trigger-history-recall-in-gemini-live-api-only-text-input-works/111617).
- Enforces Gemini 2.5's requirement that the seed end with a user turn, appending a blank user turn when it doesn't (e.g. when the bot had finished speaking before the disconnect).
- Gemini 3.x needs neither workaround — its seed path stays clean.

## Test plan

Tested the full matrix on `gemini-2.5-flash-live-preview` (default) and `gemini-3.1-flash-live-preview`:

- [x] Gemini 3:
  - [x] Initial connection, `inference_on_context_initialization=True` (default), history ending in user *or* assistant message
  - [x] Initial connection, `inference_on_context_initialization=False`, history ending in user *or* assistant message
  - [x] Reconnection without session resumption
- [x] Gemini 2.5:
  - [x] Initial connection, `inference_on_context_initialization=True` (default), history ending in user *or* assistant message
  - [x] Initial connection, `inference_on_context_initialization=False`, history ending in user *or* assistant message
  - [x] Reconnection without session resumption


kompfner:
This is my (slightly) alternate approach to the [nice PR](https://github.com/pipecat-ai/pipecat/pull/4328) that @denxxs contributed. First two commits are theirs.

denxxs:
Thanks for taking this further @kompfner, really appreciate it. LGTM from me!

filipi87 (approved):
LGTM!

PR #4358, opened 2026-04-23
Fixed SmallWebRTC data channel silently stalling on networks with a 1280-byte MTU

filipi87:
Fixed SmallWebRTC data channel silently stalling on networks with a 1280-byte MTU (IPv6, Tailscale overlays, many consumer VPNs). aiortc's default SCTP chunk size of 1200 bytes produces ~1305-byte UDP datagrams after headers, which the kernel rejects with EMSGSIZE; aiortc has no path-MTU discovery so it retransmits forever at the same oversized size. 

The chunk size is now clamped to `1100` bytes (~1205-byte datagrams, ~75 bytes of slack). Override with `PIPECAT_SCTP_MAX_CHUNK_SIZE` if your path MTU requires a different value.


env.example
```diff
@@ -214,4 +214,26 @@ WHATSAPP_PHONE_NUMBER_ID=...
 WHATSAPP_APP_SECRET=...
 
 # xAI / Grok
-XAI_API_KEY=...
\ No newline at end of file
+XAI_API_KEY=...
+
+# SmallWebRTC / aiortc
```
markbackman: Perhaps we can put these details in docs and link out to them?

markbackman (approved):
LGTM. Just one non-blocking comment.

env.example
filipi87: @markbackman , what do you think about this ?
- https://github.com/pipecat-ai/docs/pull/769

I have also explained about ice server which were missing in our docs.

markbackman: Docs look good!

I think you could merge down the docs so that you get a URL to link to. That means the docs will be available ~1 day before the release, but I think it's OK for this case.

PR #4360, opened 2026-04-23
Add Soniox real-time TTS service

markbackman:
## Summary

- Adds `SonioxTTSService`, a WebSocket-based TTS provider that streams text in and audio out over a persistent connection, multiplexing up to 5 concurrent streams per socket via Soniox's `stream_id`.
- Handles Soniox's two-phase end-of-stream (`audio_end` followed by `terminated`), client-initiated cancellation on pipeline interruption (`{"cancel": true}` per `stream_id`), per-stream error routing that keeps sibling streams alive, and a protocol-level keepalive to stay inside Soniox's 20-30s idle timeout.
- Updates `examples/voice/voice-soniox.py` to use the new TTS alongside `SonioxSTTService`, and adds Soniox to the TTS services row in the README.

## Example usage

```python
from pipecat.services.soniox.tts import SonioxTTSService

tts = SonioxTTSService(
    api_key=os.environ["SONIOX_API_KEY"],
    settings=SonioxTTSService.Settings(
        voice="Maya",
        language=Language.EN,
    ),
)
```

## Test plan

- [x] `uv run ruff check src/pipecat/services/soniox/tts.py && uv run ruff format --check src/pipecat/services/soniox/tts.py`
- [x] `uv run pytest tests/test_service_init.py -k Soniox`
- [x] Live smoke test: set `SONIOX_API_KEY`, run `examples/voice/voice-soniox.py`, and confirm Soniox TTS audio plays end-to-end with no leftover audio after an interruption.


src/pipecat/services/soniox/tts.py
```diff
@@ -0,0 +1,520 @@
...
+
+        Args:
+            frame: The start frame containing initialization parameters.
+        """
+        await super().start(frame)
+        if self._audio_format.startswith("pcm_") and self.sample_rate not in VALID_SAMPLE_RATES:
+            logger.warning(
+                f"{self}: sample_rate={self.sample_rate} is not in Soniox supported rates "
+                f"{sorted(VALID_SAMPLE_RATES)}; the server may reject the stream."
+            )
```
filipi87, line 254: Is a warning enough in this case? Shouldn't we raise an error if the sample rate is not supported?

src/pipecat/services/soniox/tts.py
```diff
@@ -0,0 +1,520 @@
...
+        super().__init__(
+            text_aggregation_mode=text_aggregation_mode,
+            # Soniox doesn't expose alignment data, so TTSTextFrames can be
+            # pushed immediately by the base class.
+            push_text_frames=True,
+            # We push TTSStoppedFrame ourselves when Soniox sends `terminated`.
+            push_stop_frames=False,
+            # Let the base class create audio contexts and emit TTSStartedFrame.
+            push_start_frame=True,
+            pause_frame_processing=False,
```
filipi87, line 205: Just for double checking that it works fine with this, that Soniox is able to handle multiple contexts at the same time.

src/pipecat/services/soniox/tts.py
markbackman: The one case for the warning is that Soniox may add a sample rate before Pipecat updates our list of sample rates. Then, we leave Soniox to reject. I don't have strong feelings on this though.

src/pipecat/services/soniox/tts.py
markbackman: Things worked well in my testing, but it's possible that I didn't hit a multiple contexts path. How do you test it?

src/pipecat/services/soniox/tts.py
```diff
@@ -0,0 +1,520 @@
...
+
+        self._api_key = api_key
+        self._url = url
+
+        # Init-only audio format (not runtime-updatable).
+        self._audio_format = audio_format
+
+        # Tracks which context_ids have had their per-stream config sent.
+        # Soniox rejects duplicate config for the same stream_id.
+        self._configured_contexts: set[str] = set()
```
filipi87, line 219: Instead of tracking this, you could override this method:
`async def on_turn_context_created(self, context_id: str):`

This allows you to send the configuration for a new context when it is created, which is what we do in the `InworldTTSService`.

This way, you can completely remove any logic related to `self._configured_contexts`.

src/pipecat/services/soniox/tts.py
```diff
@@ -0,0 +1,520 @@
...
+        try:
+            if not self._websocket or self._websocket.state is State.CLOSED:
+                await self._connect()
+
+            try:
+                if context_id not in self._configured_contexts:
+                    config = self._build_config_msg(context_id)
+                    await self._get_websocket().send(json.dumps(config))
+                    self._configured_contexts.add(context_id)
+                    logger.trace(f"{self}: opened Soniox stream {context_id}")
```
filipi87: You should do this inside: `on_turn_context_created`

src/pipecat/services/soniox/tts.py
filipi87: Makes sense. We can leave like this. 👍

src/pipecat/services/soniox/tts.py
filipi87: >  How do you test it?

I was testing sending multiple `TTSSpeakFrame`.  For example like this:

```
        # Kick off the conversation.
        await task.queue_frames([TTSSpeakFrame("Hello, I'm a bot."), TTSSpeakFrame("Can you hear me fine ?")])
```

markbackman: Ahh! Good to know. I'll test this. Thank you.

markbackman: This works as expected. I hear two sentences consecutively.

@filipi87 if it failed, I would expect to hear audio playing out at the same time, is that right?

filipi87: Or the sentences would be out of order, like the words were mixed up.

filipi87: I have just tested here as well, it seems to be working as expected. 🎉

src/pipecat/services/soniox/tts.py
markbackman: I added this override and hit an edge case. I'm not sure if it's specific to Soniox.

The issue arises when the base TTSService receives `LLMFullResponseStartFrame` and creates a context (e.g. `on_turn_context_created`). But the user interrupts before the text reaches `run_tts` to send to Soniox. This means that we open a context with Soniox and never send text. After 5 seconds, we get a 408: request timeout error from Soniox. This doesn't affect how Soniox works, but it does cause an unhandled error.

I don't get this issue if I don't override `on_turn_context_created`. This is because the context creation occurs concurrently with the text processing in run_tts.

Does the same edge case exist for Inworld? What is the motivation to break up the context creation from run_tts? Is there a latency improvement?

We could work around this issue, closing the context if an interruption occurs. WDYT?

markbackman: Ok, I refactored a bit to make this class look a little more like Inworld, adding on_turn_context_completed to handle the case I mention above.

filipi87 (approved):
LGTM! 🚀

PR #4361, opened 2026-04-23
Some pyright fixes

kompfner:
## Summary

A targeted pass at pyright errors centered on the `NotGiven` / `_NotGiven` sentinel patterns. Net: **~377 errors cleared** from a full pyright scan (1317 → 940), including all flavor-mismatch errors across the OpenAI/Anthropic LLM services. No runtime behavior changes intended.

**New typing primitives:**
- `is_given(...)` TypeGuard helpers, co-located with each `NotGiven` flavor: pipecat's `_NotGiven` (in `services/settings.py`), LLMContext's `NotGiven` (in `processors/aggregators/llm_context.py`), OpenAI's, and Anthropic's. Pyright cannot narrow on identity checks against module-level sentinels (they aren't typed as singletons), so callers use these guards instead of `x is NOT_GIVEN`.
- `assert_given(...)` in `services/settings.py` for reads from store-mode settings, where `_NotGiven` should never appear in practice (per `validate_complete()`'s invariant) but the field type still includes it. Returns the value narrowed, raises on the invariant violation.

**Type fixes across services:**
- Replaced direct `is NOT_GIVEN` checks with `is_given(...)` at call sites pyright had flagged.
- Used `assert_given(...)` at ~150 store-mode settings read sites across 43 services.
- Widened settings field types to admit `None` where services initialize them with `None` to mean "no override / use the vendor default" (~30 fields across 11 files).
- Widened LLM Settings fields in OpenAI/Anthropic services to also accept the SDK's `NotGiven`, since those services intentionally pass the SDK sentinel through to API calls. Each override is in the service-specific Settings subclass to keep the base class SDK-agnostic.
- Narrowed `voice` in `SpeechmaticsTTSSettings` to disallow `None` (Speechmatics requires a voice).

**Documenting the LLMContext/OpenAI boundary:**
- LLMContext's `NotGiven`, `LLMContextToolChoice`, and `LLMStandardMessage` are currently aliased to their OpenAI equivalents, so values flowing between the two sides type-check implicitly. Added two module-private cast helpers in `open_ai_adapter.py` — `_openai_from_llm_context_tool_choice(...)` and `_openai_from_llm_standard_message(...)` — that are typed no-ops today (via `typing.cast`) but name and document every boundary crossing. If LLMContext ever diverges from OpenAI's types, every crossing becomes a grep away rather than a silent mismatch.

**Other:**
- Widened `TLLMInvocationParams` TypeVar bound from `dict[str, Any]` to `Mapping[str, Any]` so `TypedDict` invocation-param classes satisfy it (per PEP 589, `TypedDict` is not a subtype of `dict`).
- Dropped 5 files from the pyright ignore list as they no longer have errors.

## Testing

- `uv run pytest tests/test_service_init.py` passes (one pre-existing `NvidiaSegmentedSTTService` failure on `speaker_diarization` is unrelated to this PR).
- `uv run pyright` (with the existing ignore list) passes clean.


markbackman (approved):
Nice clean up! 👏

Issue #3985, opened 2026-03-11
Proposal: graceful interruption release for TTS/output transports


golbin:
### Problem Statement

Pipecat interruption handling is very responsive today, but for real-time voice UX it can sound too abrupt.

When an `InterruptionFrame` arrives, the current behavior is generally optimized for immediate stop:
- output transports reset/cancel audio processing right away
- TTS services/providers also tend to clear/cancel immediately

This is technically correct and responsive, but assistant speech can sound hard-cut instead of naturally released.

In practice, this is difficult to solve cleanly at the application layer because both the transport and the TTS side may clear state aggressively. Apps often end up monkeypatching internal behavior just to avoid a harsh cutoff.

### Proposed Solution

Expose an explicit interruption release policy at the framework level.

Example shape:

- `interruption_release_mode`
  - `immediate_cut`
  - `bounded_drain`
- `interruption_max_release_drain_ms`
  - drain a small amount of already-generated, not-yet-played audio
  - suggested range: `40-80ms`
- `interruption_fade_out_ms`
  - apply a short fade-out to the drained tail
  - suggested range: `10-30ms`

#### `immediate_cut`
- current behavior
- interruption clears/stops immediately

#### `bounded_drain`
- stop new synthesis immediately
- allow the output transport to play a small bounded amount of already-generated audio
- apply a short fade-out at the end
- discard anything beyond the configured bound

This keeps interruption responsive while avoiding a harsh cutoff.

This does not need to change the default behavior. `immediate_cut` can remain the default, and graceful release can be opt-in.

A reasonable starting point for opt-in defaults might be:
- `interruption_release_mode = bounded_drain`
- `interruption_max_release_drain_ms = 60`
- `interruption_fade_out_ms = 20`

If this direction sounds useful, I’d be happy to help implement it.

### Would you be willing to help implement this feature?

- [x] Yes, I'd like to contribute
- [ ] No, I'm just suggesting

markbackman:
I've done some experimentation with this, but I haven't found a good solution. @golbin if you have a proposal, we're happy to collaborate on it. I can't guarantee that we'll accept it, but it's an interesting route to pursue that we're happy to collaborate on if you're interested.

Stoic-Angel:
Hi! Raised this a couple weeks ago, does this need any other changes?

swayamg20:
Hi @markbackman — I reviewed the current interruption path and the earlier attempt in #4364, including your `voice-cartesia.py` listening result.

[...]

PR #4364, opened 2026-04-24
Add audio fade-out functionality and interruption handling

Stoic-Angel:
## Graceful interruption - addresses #3985

- Introduced `apply_fade_out_to_pcm16` function for linear fade-out on PCM audio.
- Updated `BaseOutputTransport` to support `bounded drain` mode for audio interruptions, allowing a controlled release of audio with optional fade-out.
- Enhanced `TransportParams` to include settings for interruption release mode, maximum drain duration, and fade-out duration.
- Added a unit test for the new functionality, ensuring correct behaviour of fade-out and bounded drain logic.


markbackman:
I'm sorry it took so long to look at this. In testing, I still hear an audio pop on hard cutoffs. I tried this running https://github.com/pipecat-ai/pipecat/blob/main/examples/voice/voice-cartesia.py.

markbackman:
Closing this as stale. See my previous comment if you think there's a better solution to this. Thanks for the attempt!

Issue #4118, opened 2026-03-24
gemini live function calling options : Addition of SILENT, WHEN_IDLE, and INTERRUPT modes + other improvements


remisharrock:
### Question

I've received this email from Google today and was wondering if pipecat is ready for the Addition of SILENT, WHEN_IDLE, and INTERRUPT modes ?

```
You’re receiving this communication because you’re a user of Gemini 2.5 Live API Native Audio (General Availability) model ID: gemini-live-2.5-flash-native-audio.

We’re implementing a scheduled product update on April 28, 2026, to add new product features and improvements.

We’ve provided additional information below to help you understand the details of this update.

What you need to know

New features:

Function call modes: Addition of SILENT, WHEN_IDLE, and INTERRUPT modes for more granular control over how and when the AI triggers functions.
Vertex AI Studio integration: The ability to define and test function calls directly in Vertex AI Studio.
Improved transcription quality: New support for a language-specific flag to enhance transcription accuracy.
Product improvements:

Function calling improvements:
“No Response” issues: Minimized instances where the model remains silent during long-running tool calls.
“Trigger Repeat”issues: Minimized the issue where the model sends the same function call multiple times before receiving a response.
Note: We still recommend client-side filtering for duplicate function calls as a best practice.
Transcription quality: Resolved the transcription spacing issues for a higher quality output.
```

remisharrock:
official documentation is 

https://ai.google.dev/gemini-api/docs/live-api/tools#async-function-calling

especially that part:

NON-BLOCKING ensures the function runs asynchronously while you can continue interacting with the model.

Then you need to tell the model how to behave when it receives the FunctionResponse using the scheduling parameter. It can either:

Interrupt what it's doing and tell you about the response it got right away (scheduling="INTERRUPT"),
Wait until it's finished with what it's currently doing (scheduling="WHEN_IDLE"),
Or do nothing and use that knowledge later on in the discussion (scheduling="SILENT")


```python
# for a non-blocking function definition, apply scheduling in the function response:
  function_response = types.FunctionResponse(
      id=fc.id,
      name=fc.name,
      response={
          "result": "ok",
          "scheduling": "INTERRUPT" # Can also be WHEN_IDLE or SILENT
      }
  )
```

elliottventures:
Adding a data point from a reproduction today (2026-04-21) in case it's useful for prioritization.

[...]

PR #4366, opened 2026-04-24
fix: send realtime input after tool_result so Gemini 3.x runs inference

elliottventures:
## What

After `send_tool_response`, send `send_realtime_input(text=" ")` for Gemini 3.x models. Mirrors the workaround already applied in `_create_initial_response` and `_create_single_response` in the same file.

## Why

Gemini 3.x won't run inference on content it receives unless a realtime input follows. The tool-result path in `GeminiLiveLLMService` was the only remaining send-path missing this workaround. The in-file comment on the existing two sites already states *"Gemini 3.x wants turn_complete=True, but also won't run inference without a realtime input."*

**Symptom:** every tool-call turn stalls silently after the tool returns, until the user speaks again (their speech itself is a realtime input that wakes inference back up). TTFB on affected turns ranges from single-digit seconds to multiple minutes depending on when the user gives up waiting. Reproducible on `gemini-3.1-flash-live-preview` + Daily transport + any function-calling tool. Not reproducible on Gemini 2.5 (which doesn't require the kick).

## The fix

3-line addition to `_tool_result` — the same guarded `send_realtime_input(text=" ")` call that already exists in the two other send paths in the same file, with the identical comment.

## Testing

Verified against a production voice agent using `gemini-3.1-flash-live-preview` + Daily direct mode + a 7-tool function-calling setup (begin_survey, advance_step, record_field, take_photo, flag_issue, go_back, end_survey):

- **Before this fix** (applied locally as a subclass monkeypatch): 20–130s stalls on every tool-result turn.
- **After this fix**: sub-second TTFB on every tool-result turn across 20+ test turns in multiple 6-step survey walkthroughs.

No unit test added — the Gemini Live service's `_session` is an `AsyncSession` from google-genai's bidi WebSocket SDK and the existing test suite doesn't have a mocking pattern for it. Happy to add one if a maintainer has a preferred shape.

## Related

I'll separately open a PR for a related `LLMAssistantAggregator` bug that I found while diagnosing this, and add a summary data point on #4118.

### DIFF
--- a/changelog/4366.fixed.md
+++ b/changelog/4366.fixed.md
@@ -1,0 +1,1 @@
+- Fixed Gemini Live tool calls stalling silently on Gemini 3.x models. After `_tool_result` sent the function response, the service omitted the `send_realtime_input` kick that Gemini 3.x needs to resume inference — the same kick already applied in `_create_initial_response` and `_create_single_response`. Previously, every tool-call turn stalled until the user spoke again.
--- a/src/pipecat/services/google/gemini_live/llm.py
+++ b/src/pipecat/services/google/gemini_live/llm.py
@@ -1416,6 +1416,9 @@ class GeminiLiveLLMService(LLMService):
 
         try:
             await self._session.send_tool_response(function_responses=response)
+            # Gemini 3.x wants turn_complete=True, but also won't run inference without a realtime input
+            if self._is_gemini_3:
+                await self._session.send_realtime_input(text=" ")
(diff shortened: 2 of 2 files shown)


elliottventures:
@markbackman Wanted to make sure you guys saw this PR

pankajagarwal:
+1 to get this fix into pipecat main as this significantly impacts the quality.

PR #4368, opened 2026-04-24
Fix issue where STTService unintentionally created a method with the same name as SegmentedSTTService.

filipi87:
Fix issue where STTService unintentionally created a method with the same name as SegmentedSTTService.

### DIFF
--- a/src/pipecat/services/stt_service.py
+++ b/src/pipecat/services/stt_service.py
@@ -408,7 +408,7 @@ class STTService(AIService):
             await self._handle_vad_user_stopped_speaking(frame)
             await self.push_frame(frame, direction)
         elif isinstance(frame, UserStoppedSpeakingFrame):
-            await self._handle_user_stopped_speaking(frame)
+            await self._maybe_reconnect()
             await self.push_frame(frame, direction)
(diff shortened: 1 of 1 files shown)


src/pipecat/services/stt_service.py
```diff
@@ -508,8 +508,8 @@ async def _handle_vad_user_started_speaking(self, frame: VADUserStartedSpeakingF
         self._finalize_pending = False
         self._last_transcript_time = 0
 
-    async def _handle_user_stopped_speaking(self, frame: UserStoppedSpeakingFrame):
-        """Handle user stopped speaking frame.
+    async def _maybe_reconnect(self):
```
kompfner: nit: if it's specifically meant only to maybe trigger a reconnection after the user stops speaking (as the docstring says), give it a more verbose name to reflect that. perhaps `_maybe_reconnect_on_user_stopped_speaking`

kompfner (approved):
This fixes the SegmentedSTTService breakage in my testing 👍 One suggestion, but otherwise looks good to me!

src/pipecat/services/stt_service.py
filipi87: Done.

PR #4370, opened 2026-04-24
feat(daily): support screenVideo destination and configurable camera send settings

aconchillo:
## Summary

- Adds support for Daily's built-in `screenVideo` destination in `DailyTransport`. When `"screenVideo"` is included in `video_out_destinations`, a dedicated screen video track is created at join time and frames with `transport_destination="screenVideo"` are routed to it.
- Adds `DailyParams.camera_out_send_settings`, a dict passed verbatim to the Daily client's camera publishing settings so applications can fully control encoding, codec, bitrate, and framerate.
- Deprecates `TransportParams.video_out_bitrate` for the Daily transport (use `camera_out_send_settings` instead) and changes its default to `None`.

## Testing

- Configure a `DailyTransport` with `video_out_destinations=["screenVideo"]` and verify the screen video track is published; send `OutputImageRawFrame` with `transport_destination="screenVideo"` and confirm it lands on the screen track.
- Configure both `"camera"` and `"screenVideo"` destinations and verify frames are routed correctly based on `transport_destination`.
- Configure `camera_out_send_settings` with custom encodings (e.g., higher bitrate / different codec) and verify the published track honors the values.
- Verify existing camera-only setups still work unchanged.

### DIFF
--- a/src/pipecat/transports/daily/transport.py
+++ b/src/pipecat/transports/daily/transport.py
@@ -555,6 +555,7 @@ class DailyTransportClient(EventHandler):
         self._camera_track: DailyVideoTrack | None = None
         self._microphone_track: DailyAudioTrack | None = None
         self._custom_audio_tracks: dict[str, DailyAudioTrack] = {}
+        # Custom video tracks will also include `screenVideo`.
         self._custom_video_tracks: dict[str, DailyVideoTrack] = {}
 
     def _speaker_name(self):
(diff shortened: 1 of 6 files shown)


src/pipecat/transports/daily/transport.py
```diff
@@ -922,6 +952,22 @@ async def _join(self):
...
+                                {"preferredCodec": self._params.video_out_codec}
+                                if self._params.video_out_codec
+                                else {}
+                            ),
+                            "encodings": {
+                                "low": {
+                                    "maxBitrate": self._params.video_out_bitrate,
+                                    "maxFramerate": self._params.video_out_framerate,
+                                }
+                            },
```
filipi87: I see that we did the same for the camera, but is there any reason we decided to send only a single layer instead of relying on our default behavior of sending three layers ? 

Basically, we could rely on the default settings provided by our SDK unless the user overrides them.

filipi87 (approved):
LGTM! 🚀

src/pipecat/transports/daily/transport.py
aconchillo: Yes, agreed. This is very old. Adding a commit.

src/pipecat/transports/daily/transport.py
```diff
@@ -1246,61 +1248,6 @@ async def capture_participant_video(
             color_format=color_format,
         )
 
-    async def add_custom_audio_track(
```
filipi87, line 1249: I am not sure I understand why we’re removing `add_custom_audio_track` and `remove_custom_audio_track`. I believe we still need them, don’t we?

PR #4371, opened 2026-04-24
Add a global context for tool calls: tool_resources

Stoic-Angel:
## Attempting to solve #4256 
- Introduced a new optional parameter, tool_resources, to the PipelineTask and FrameProcessorSetup classes. 
- The `tool_resources` parameter allows for the sharing of application-defined resources across tool handlers and processors during pipeline execution. 
- Additionally, the LLMService class has been updated to cache these resources during setup and pass them to function call parameters. 
- A new test suite has been added to validate the functionality of tool_resources integration.


Stoic-Angel:
@kompfner looks good?

kompfner (approved):
This makes sense to me! Clean and minimal and I believe gets the job done. Thanks!

One thing we might want to add is an example file showing how this might be used, maybe as `examples/function-calling/function-calling-tool-resources`. Happy to add that example post-merge.

PR #4136, opened 2026-03-25
Upgrade protobuf to 6.x for nvidia-riva-client compatibility

markbackman:
## Summary

- Upgraded `protobuf` from 5.x to 6.x (`>=6.31.1,<7`) to fix compatibility with `nvidia-riva-client>=2.25.1`, which ships gencode compiled against protobuf 6.31.1
- Upgraded `grpcio-tools` (dev dependency) from `~=1.67.1` to `~=1.78.0` (required for protobuf 6.x)
- Bumped `nvidia-riva-client` lower bound to `>=2.25.1`
- Regenerated `frames_pb2.py` with new protobuf 6.31.1 compiler

## Testing

- `uv run pytest tests/test_protobuf_serializer.py` — roundtrip serialization passes
- `python -c "import riva.client"` — no version error

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #4372, opened 2026-04-25
Relax protobuf pin to support both 5.x and 6.x runtimes

markbackman:
## Summary

Pipecat 1.0.8 hard-required protobuf 6.x via the base `protobuf>=6.31.1,<7` pin, blocking users whose dependency graph already constrains protobuf to the 5.x line. The original bump (#4136) was only needed because `nvidia-riva-client>=2.25.1` ships gencode compiled with protoc 6.31.1 — pipecat's own schema is plain proto3 with no version-specific features.

This PR widens the pin so 5.x users are unblocked, while keeping 6.x as the floor when the `nvidia` extra is installed.

## Changes

- **Widen base pin** to `protobuf>=5.29.6,<7`.
- **Regenerate `frames_pb2.py`** with `grpcio-tools~=1.67.1` (protoc 5.x). Per [Google's cross-version runtime guarantee](https://protobuf.dev/support/cross-version-runtime-guarantee), 5.x gencode runs on both 5.x and 6.x runtimes — one artifact serves all users.
- **Loosen `grpcio-tools`** dev pin to `>=1.67.1,<2` so contributors can install `pipecat-ai[dev,nvidia]` without resolver conflict. Comment in `frames.proto` documents the 1.67.x requirement for future regeneration.
- **Add explicit `protobuf>=6.31.1,<7` to the `nvidia` extra.** This compensates for an upstream packaging gap: `nvidia-riva-client 2.25.1` ships protobuf 6.x gencode but does not declare protobuf in its `Requires-Dist`. Filed as https://github.com/nvidia-riva/python-clients/issues/172. Once NVIDIA fixes their metadata, the explicit `protobuf` entry in the `nvidia` extra can be removed.

## How resolution works after this PR

| Install | protobuf range pip sees | Resolves to |
|---|---|---|
| `pip install pipecat-ai` | `>=5.29.6,<7` | latest in range (5.x or 6.x both work) |
| `pip install pipecat-ai[nvidia]` | `>=5.29.6,<7` ∩ `>=6.31.1,<7` | 6.x |
| `pip install "protobuf~=5.29.6" pipecat-ai` | pinned to 5.29.x | 5.29.6 |

## Test plan

- [x] `tests/test_protobuf_serializer.py` passes against the regenerated gencode
- [x] `import riva.client` succeeds when installing `pipecat-ai[nvidia]` (verified resolver picks protobuf 6.33.6)
- [x] Clean venv pinned to `protobuf~=5.29.6`: pipecat installs, `frames_pb2` imports, and `ProtobufFrameSerializer` roundtrips a `TextFrame` cleanly
- [ ] Reviewer to consider whether other extras shipping protobuf gencode (livekit, heygen, google) need similar explicit-protobuf entries


filipi87 (approved):
LGTM!

PR #4374, opened 2026-04-26
Expire runner-created Daily rooms after 4h

markbackman:
I'm making this change because I noticed that I have _many_ Daily rooms piled up in my account. We should instead be expiring them when the local dev runner is used.

## Summary

- Runner-created Daily rooms had no expiration when callers posted partial `dailyRoomProperties` (e.g. `{"start_video_off": true}`), so they accumulated and never got cleaned up by Daily's cron.
- Encode Pipecat Cloud's 4-hour max session policy in the runner: `PIPECAT_CLOUD_ROOM_EXP_HOURS = 4.0`, applied via `setdefault` for user-supplied dicts and `room_exp_duration` for all four `configure()` call sites.

## Test plan

- [ ] `python bot.py -t daily`, open `http://localhost:7860/`, fetch the created room via Daily REST: `exp` ~4h out, `eject_at_room_exp=true`.
- [ ] `POST /start` with `{"createDailyRoom": true, "dailyRoomProperties": {"start_video_off": true}}`: room has `exp` ~4h, `eject_at_room_exp=true`, `start_video_off=true` preserved.
- [ ] `POST /start` with explicit `exp` + `eject_at_room_exp=false`: caller values win.
- [ ] `python bot.py -t daily -d` (direct mode): created room has `exp` ~4h, `eject_at_room_exp=true`.


src/pipecat/runner/run.py
```diff
@@ -106,6 +107,9 @@ async def bot(runner_args: RunnerArguments):
 
 TELEPHONY_TRANSPORTS = ["twilio", "telnyx", "plivo", "exotel"]
 
+# Mirror Pipecat Cloud's 4-hour max session limit so dev rooms get cleaned up.
+PIPECAT_CLOUD_ROOM_EXP_HOURS = 4.0
```
filipi87: Nit: Maybe we should rename it to `PIPECAT_ROOM_EXP_HOURS` instead of naming it after Pipecat Cloud.

filipi87 (approved):
LGTM!

src/pipecat/runner/run.py
markbackman: Good point. Updating now.

PR #4376, opened 2026-04-27
Add OpenAI responses to readme

kompfner:
Two things:
1. Add OpenAI Responses to the README's list of LLM services
2. Update docs URLs in README to reflect new docs site structure and avoid redirects


markbackman (approved):
Thanks for cleaning up the links too 🙇

PR #4378, opened 2026-04-27
More pyright fixes

kompfner:
## Summary

A grab-bag pyright cleanup PR. Ten substantive themes plus a structural typing improvement:

### LLM adapters — TypedDict / Optional cleanup (10 files cleaned, removed from ignore list)

A common pattern across many adapter files: the code mutates values typed as TypedDict union members (`MessageParam`, `ChatCompletionMessageParam`, etc.) using dict-style key/index access that those TypedDicts don't permit, or accesses Optional fields without guards. Applied a uniform fix shape (cast to `dict[str, Any]` for the mutation block, cast back at return sites; narrow Optionals where they slip through).

Cleaned & removed from the pyright ignore list:

- `anthropic_adapter.py` (115 → 0)
- `perplexity_adapter.py` (23 → 0)
- `bedrock_adapter.py` (19 → 0)
- `gemini_adapter.py` (25 → 0)
- `open_ai_adapter.py` (5 → 0)
- `open_ai_responses_adapter.py` (5 → 0)
- `open_ai_realtime_adapter.py` (28 → 0)
- `aws_nova_sonic_adapter.py` (12 → 0)
- `grok_realtime_adapter.py` (21 → 0)
- `inworld_realtime_adapter.py` (21 → 0)
- Also `processors/aggregators/llm_context.py` (13 → 0).

Most changes are pure typing. A handful are small latent bug fixes flagged in their commits — defensive None guards on Pydantic Optional fields (gemini), and a few real fixes in `open_ai_realtime_adapter.py` (the bulk-packing path was constructing raw dicts where a Pydantic `ConversationItem` was expected; an unhandled-message fallthrough was implicitly returning None despite a non-Optional return annotation).

### AWS Nova Sonic — `aws/nova_sonic/llm.py`

Several disconnected pyright errors that turned out to map to small real bugs or undocumented assumptions:

- `get_setup_params` was unannotated and could return `NotGiven` in the tools position, breaking the `NovaSonicSessionSender` protocol it claims to implement. Added the matching return annotation and coerced the `NotGiven` case to an empty list.
- `_process_completed_function_calls` was iterating context messages without guarding `self._context` (typed `LLMContext | None`), didn't skip `LLMSpecificMessage` items (which lack `.get`), and was using a truthy `role` check where the intent was `role == "tool"`. Tightened all three.
- `reset_conversation` (public API) silently passed `None` to `_handle_context` when no initial context had been received. Now logs a warning and bails — treating that as developer error.
- The connection-startup tools-coercion path had the same `NotGiven` issue as `get_setup_params`.

### `LLMService` generic refactor

Made `LLMService` generic over its adapter type:

```python
TAdapter = TypeVar("TAdapter", bound=BaseLLMAdapter, default=BaseLLMAdapter)

class LLMService(..., Generic[TAdapter]):
    def get_llm_adapter(self) -> TAdapter: ...
```

In-tree subclasses opted in (`AnthropicLLMService(LLMService[AnthropicLLMAdapter])`, etc.). Callers now get the precise adapter type from `get_llm_adapter()` automatically, which lets us drop **17 redundant `adapter: SomeAdapter = self.get_llm_adapter()`** workarounds and **7 redundant `params: SomeLLMInvocationParams = adapter.get_llm_invocation_params(...)`** annotations.

Backward-compatible for third-party providers: `class MyService(LLMService):` (no bracket) still works. The `default=BaseLLMAdapter` on the TypeVar means pyright resolves `TAdapter` to `BaseLLMAdapter` for unparameterized usage, matching the pre-refactor return type. `WebsocketLLMService` is also generic so the multi-inheritance case (`OpenAIResponsesLLMService`) keeps both bases agreeing on `TAdapter`. Added a `TestUnparameterizedSubclass` test exercising this contract.

### Optional-websocket/client access — 19 sites across services and transports

A common pattern: `await self._<connection>.send/recv/...` where the receiver is typed `X | None`. Each call site got a fix matched to its semantics rather than a blanket suppression:

- **Streaming/user-facing paths** (e.g. `cartesia/stt.py`, `elevenlabs/tts.py`, `deepgram/sagemaker/tts.py`, lemonslice/tavus `send_message`): early return + warn — drop and warn is the right runtime fail-safe when reconnect didn't succeed.
- **"Should never happen" cases** (e.g. `deepgram/flux/stt.py` transport methods, `_send_keepalive` in `stt_service.py`, `_ws_recv` in `WebsocketLLMService`): early return with a comment explaining the upstream gate, no warn — caller is already guarding on `_is_active`/`_is_keepalive_ready`/`self._connected`.
- **Just-assigned-above** (e.g. `lmnt/tts.py`, `gradium/stt.py`, `fish/tts.py`): use a local variable so pyright keeps the narrowing across statements.
- `websocket_service.send_with_retry` raises `ConnectionError` from inside its existing try-block when `self._websocket` is None, preserving the prior behaviour where the broad `except` triggered reconnect.

Removed `cartesia/stt.py`, `elevenlabs/stt.py`, and `soniox/stt.py` from the pyright ignore list as a side effect.

### `language_to_<service>_language` return types

These provider-specific helpers are all thin wrappers around `resolve_language(...)`, which itself returns `str` — never `None`. The `str | None` annotations were misleading and were producing spurious pyright errors at the call sites that assigned the result into a `str` field. Tightened each helper to return `str`, and rewrote the `Returns:` docstrings to describe the actual fallback behaviour (resolve to base or full code, with a warning).

Importantly, the per-class `language_to_service_language(...)` methods on `STTService`/`TTSService` subclasses keep `str | None` as their return type. That signature is an extension hook for future and/or third-party subclasses that may genuinely not be able to produce a code for some languages.

### `_settings` `_NotGiven` leaks + Google STT Liskov fix

Six sites where a value flowed out of `self._settings.X` (typed `T | _NotGiven`) into a context that wanted the plain `T`. Each got an `assert_given(...)` wrap at the boundary:

- `aws/nova_sonic/llm.py`: `_settings.model` and `_settings.system_instruction`.
- `deepgram/flux/base.py`: iterating `_settings.keyterm`.
- `google/stt.py`: iterating `_settings.languages`.
- `google/tts.py`: iterating `_settings.speaker_configs`.
- `openai/base_llm.py`: `_settings.system_instruction`.

Also took a deeper pass on a related Google STT issue: the override of `language_to_service_language` had been broadened to take `Language | list[Language]` and return `str | list[str]`, a Liskov violation against the base's `Language -> str | None` contract. Restored it to a single-Language signature with `_get_language_codes` iterating; the override is tightened to return `str` (narrower than `str | None`, LSP-compatible) since it always falls back to `"en-US"` rather than returning None.

### TTS/STT services and one-offs — 19 files cleared from the ignore list

A second pass over services with low single-digit error counts. Two dominant patterns plus a handful of one-offs:

**Voice / `context_id` `None` handling.** `assert_given(self._settings.voice)` strips `_NotGiven` but not `None`, and `get_active_audio_context_id()` legitimately returns `None` when no context is active. Two shapes of fix:

- For services where a missing voice means the request can't proceed (`hume`, `openai`, `xtts`, `groq`, `kokoro`, `piper`), added an explicit `None` check. Inside `run_tts` we yield an `ErrorFrame` and return — matching each service's existing error-emission style (some wrap `Exception` broadly so `raise` is fine; openai/hume/xtts had narrower-or-no try blocks where a bare raise would have escaped). Piper validates in `__init__`. OpenAI also gained a `voice not in VALID_VOICES` guard with a clear message listing supported voices.
- For services where a missing audio context just means "skip this message" (`fish`, `lmnt`, `smallest`, `sarvam`, `neuphonic`), widened `TTSService.append_to_audio_context`'s `context_id` parameter to `str | None`. The body already explicitly handled `None` with a debug log + early return, so the `str` annotation was a lie. Same shape applied to `inworld._close_context`.

**`google.genai` imports — IDE autocomplete win.** Switched `from google import genai` to `import google.genai as genai` in `google/image.py` and `google/llm.py`. The dotted form sidesteps a PEP 420 namespace-package stub gap (the `google` namespace stubs come from a different distribution and don't declare `genai`), so pyright now resolves `genai` to the real module rather than `Unknown`. **IDE autocomplete on `genai.<x>` works for the first time.** In `image.py` this surfaced three latent bugs the `Unknown` resolution had been hiding (`model: str | _NotGiven | None` not narrowed before passing to the SDK; two `.image_bytes` accesses on `Image | None` without a guard) — all fixed. `llm.py`'s switch surfaced 8 errors (Content-list typing nuances, internal `_api_client` access, a few small Optionals) — deferred to a future pass; the file stays in the ignore list with the dotted import in place.

**Latent bug fixes spotted along the way:**

- `resembleai/tts.py` was calling `push_error(ErrorFrame(...))` — but `push_error` takes a string; there's a separate `push_error_frame` for the frame case. Switched to the right method.
- `openai/base_llm.py`: `max_completion_tokens` was the only sibling field on `OpenAILLMSettings` missing `| None` in its type, which caused the assignment from `params.max_completion_tokens` (`int | None`) to fail. Added `| None` for consistency with `max_tokens` etc.
- `heygen/base_api.py`: `livekit_url: str = None` and `ws_url: str = None` — declared `str` while defaulting to `None`. Removed the bogus defaults; both fields are required at construction in every in-tree call site.

**Other small ones:** gladia `_session_url` None guard before `websocket_connect`; openrouter's `build_chat_completion_params` override had widened to `dict[str, Any]` diverging from the parent's `OpenAILLMInvocationParams` (restored); neuphonic guarded the receive loop's `async for message in self._websocket` with the local-variable narrowing pattern from #4359.

**`groq/tts.py` `output_format` typing — hint-not-constraint.** Tightened to `Literal["flac","mp3","mulaw","ogg","wav"] | str = "wav"`. The literal side gives IDE autocomplete hints for the currently-supported set; the `| str` side keeps callers unblocked if groq adds a new format before this list is updated. A `cast` at the API boundary satisfies groq's stricter `Literal` parameter type. The literal alias mirrors the inlined Literal on `groq.resources.audio.speech.AsyncSpeech.create`'s `response_format` (the SDK doesn't export it as a named symbol).

**`websocket_service.py`:** scoped `# pyright: ignore[reportAttributeAccessIssue]` on `websockets.WebSocketClientProtocol`. That alias is now a deprecated re-export from the legacy submodule and pyright doesn't surface it on the top-level `websockets` namespace; runtime is fine. Migrating to `websockets.ClientConnection` is a separate piece of work (`transports/websocket/client.py` uses the same alias four times) and left for a future commit.

**Files dropped from the ignore list (96 → 77):** `fish/tts.py`, `gladia/stt.py`, `google/image.py`, `groq/tts.py`, `heygen/base_api.py`, `hume/tts.py`, `inworld/tts.py`, `kokoro/tts.py`, `lmnt/tts.py`, `neuphonic/tts.py`, `openai/llm.py`, `openai/tts.py`, `openrouter/llm.py`, `piper/tts.py`, `resembleai/tts.py`, `sarvam/tts.py`, `smallest/tts.py`, `websocket_service.py`, `xtts/tts.py`.

### Second pass over low-error-count files — 10 more cleared

A follow-up sweep over the remaining single-digit-error files in the ignore list. Three coherent shapes plus one-offs:

**`Language | str | None` → `Language | None` at STT frame boundaries.** `assert_given(self._settings.language)` strips `_NotGiven` but leaves `Language | str | None`, while `TranscriptionFrame.language` expects `Language | None`. In practice both `_settings.language` and SDK-supplied codes resolve to a `Language` enum value, but technically they could be raw strings — and `Language` is a StrEnum, so downstream consumers (which mostly compare/serialize as strings) handle either. Used `cast("Language | None", ...)` at each call site rather than a runtime-validating helper, so an unrecognised code (e.g. one we haven't added to the enum yet) still flows through unchanged. Cleared `azure/stt.py`, `aws/stt.py`, `gradium/stt.py`. `mistral/stt.py` got the same cast (at the SDK boundary, storing under `_detected_language: Language | None`) but stays in the ignore list because of two unrelated Optional-access errors.

**aiobotocore `async with` stub gap.** `aioboto3.Session().client(...)` is an async context manager at runtime but its stubs don't advertise `__aenter__`/`__aexit__` to pyright. Scoped `# pyright: ignore[reportGeneralTypeIssues]` on the two affected sites: `aws/agent_core.py` and `aws/tts.py`. `aws/tts.py` also had a latent bug on the no-`AudioStream` path: the original code set `audio_data = None` and then crashed in `resample(...)` and `len(audio_data)` below; replaced with an early `return` after logging — matches the convention elsewhere (OpenAI TTS, etc.) of not recording usage metrics on the error path.

**heygen `event_id: str | None` → `str` at transport→client boundary.** Three call sites in `transports/heygen/transport.py` passed `self._event_id` (`str | None`) into client methods that take `str`. Added a guard at each: `agent_speak_end` and `interrupt` only fire when `_event_id` is set; `write_audio_frame` warn-and-drops when there's no active bot event rather than sending a malformed message.

**`OpenAIResponsesLLMInvocationParams` TypedDict.** `get_llm_invocation_params` always sets both `input` and `tools` in the same dict literal, but the TypedDict was `total=False` so direct subscript access (`invocation_params["input"]`) tripped `reportTypedDictNotRequiredAccess` in `services/openai/responses/llm.py`. Marked both keys `Required[...]`; `instructions` stays non-required since it's only added when a system instruction is present.

**Latent bug in `heygen/api_interactive_avatar.py`:** the code accessed `request_data.voice.voiceId` and `request_data.voice.elevenlabsSettings`, but those names are Pydantic *aliases*; the actual attribute names (used for attribute access) are `voice_id` and `elevenlabs_settings`. Switched to the field names — those camelCase accesses would have raised `AttributeError` at runtime if `voice` was set.

**Other small fixes:**

- `assemblyai/stt.py`: the deprecated `connection_params=` init path was reading `formatted_finals` and `word_finalization_max_wait_time` off `AssemblyAIConnectionParams`, but those fields were never on the deprecated input model — they were added to `Settings` later. Removed the reads (with a comment noting they're only available via the canonical `settings=...` API).
- `rtvi/processor.py`: two `about: Mapping[str, Any] = None` parameter signatures — declared `Mapping`, defaulted to `None`, and both function bodies already handled the None case. Widened to `Mapping[str, Any] | None = None`.
- `aws/stt.py`: `subprotocols=["mqtt"]` failed against websockets' `Sequence[Subprotocol] | None` (Subprotocol is a NewType wrapper). Wrapped: `subprotocols=[Subprotocol("mqtt")]`.

**Files dropped from the ignore list (77 → 67):** `processors/frameworks/rtvi/processor.py`, `services/assemblyai/stt.py`, `services/aws/agent_core.py`, `services/aws/stt.py`, `services/aws/tts.py`, `services/azure/stt.py`, `services/gradium/stt.py`, `services/heygen/api_interactive_avatar.py`, `services/openai/responses/llm.py`, `transports/heygen/transport.py`.

### Third pass over low-error-count files — 10 more cleared

Another follow-up sweep. Most-cohesive shapes plus one-offs:

**Optional access guards.** Same fix shape as #4359: a receiver typed `X | None` accessed without a guard. Each site got the smallest behaviour-preserving narrowing.

- `mistral/stt.py`: `_connection.send_audio` could crash if `_connect()` swallowed an exception and left `_connection` unset; drop the audio chunk with a warning instead. `_receive_events` got the same defensive narrowing.
- `deepgram/flux/stt.py`: `_websocket_url` is set in `_connect` before the helper is called — assert at the use site so pyright tracks it across methods. `websocket.response` is `Response | None` in the websockets stubs even though it's always populated post-handshake; guarded with a fallback.
- `audio/filters/rnnoise_filter.py`: the module-level import sets `RNNoise` to `None` if `pyrnnoise` isn't installed; raise `ImportError` explicitly instead of relying on `None(...)` to crash inside the existing try-block.
- `transports/smallwebrtc/request_handler.py`: `get_answer()` legitimately returns `None`; raise instead of crashing on three subscript accesses.

**`TTSService.remove_audio_context` signature widening.** Mirroring the `append_to_audio_context` fix from the previous pass: parameter widened to `str | None` with the `None` handling living in the function body (early debug log + return). `audio_context_available` *keeps* its narrow `str` signature — asking "is `None` available?" isn't a meaningful question (`_audio_contexts` is `dict[str, asyncio.Queue]`) — and the one internal call site that previously passed `str | None` got an explicit narrowing. Side effect: `deepgram/tts.py`'s `remove_audio_context` error clears without local changes there.

**`deepgram/tts.py` (4 errors → 0).** Same `push_error(ErrorFrame(...))` latent bug we fixed in `resembleai/tts.py` earlier in this PR — `push_error` takes a string; there's a separate `push_error_frame` for frames. Two sites switched. The Optional `_websocket.response` access is guarded the same way as `deepgram/flux/stt.py`. The `remove_audio_context` error was cleared by the parent-class tightening above.

**`aws/utils.py` (3 errors → 0).** `AWSTranscribePresignedURL` declared `session_token: str` but the dict source is `str | None` (AWS supports long-term IAM creds without a session token). Same for `vocabulary_name`/`vocabulary_filter_name` on `get_request_url`, which were typed `str = ""` even though the body uses truthy checks to skip them. Widened to `str | None = None` — matches actual runtime semantics.

**`audio/dtmf/utils.py` — latent bug for zipped distributions.** `files("...").joinpath(...)` returns a `Traversable`, but `aiofiles.open` wants a real path. For regular pip installs this worked in practice (Traversable was a `Path`), but it would fail for zipped distributions (zipapp, zipimport) where the resource isn't on disk. Wrapped in `importlib.resources.as_file(...)` — the canonical bridge that extracts to a temp file when the resource isn't already on the filesystem. Validated end-to-end (regular install still reads bytes; ad-hoc zipapp test confirmed `as_file` extracts and yields a real `Path`).

**`openai/image.py`.** The `size` arg to `images.generate` is `Literal[...] | None` in the SDK but our settings field is `str | None`. Mirrored the `groq/tts.py` hint-not-constraint pattern from the previous pass: defined a module-level `OpenAIImageSize = Literal[...]` alias with a comment attributing the upstream symbol and documenting the cast contract (callers can pass any string; invalid values surface as an OpenAI API error). Also guarded `image.data[0]` (`response.data` is `list[Image] | None`).

**`processors/frameworks/{langchain,strands_agents}.py`.** Both processors do `messages[-1]["content"]` on a value typed `LLMStandardMessage | LLMSpecificMessage` (the latter is a dataclass, not a dict). Historically these only handled plain-text user messages, so the fix is two explicit guards (skip if the last message isn't a dict; skip if `content` isn't a string) plus a TODO noting that other shapes (multi-modal content, provider-specific messages) aren't supported yet. langchain's `__get_token_value` also got a small fix where `AIMessageChunk.content` is `str | list[parts]` but the function declares `-> str`; stringify the list case. strands_agents' surfaced two unrelated narrows: `graph_exit_node: str | None` gated by an `__init__`-time assert, and `agent.stream_async` reached only when we're not in graph mode.

**Files dropped from the ignore list (67 → 57):** `audio/dtmf/utils.py`, `audio/filters/rnnoise_filter.py`, `processors/frameworks/langchain.py`, `processors/frameworks/strands_agents.py`, `services/aws/utils.py`, `services/deepgram/flux/stt.py`, `services/deepgram/tts.py`, `services/mistral/stt.py`, `services/openai/image.py`, `transports/smallwebrtc/request_handler.py`.

### Fourth pass over low-error-count files — 8 more cleared

Another follow-up sweep. Most-cohesive shape plus a few one-offs:

**Optional access on transport/client receivers — `assert` for "should-never-happen" cases.** Same fix shape as #4359, but where the caller's lifecycle genuinely guarantees the field is non-None when the method runs, used `assert` rather than silent early-return. This makes the invariant explicit and surfaces real bugs loudly:

- `services/openai/stt.py`: traced the call chain — `_receive_messages` is started by `_connect()` only when `self._websocket` is set, and `WebsocketService._receive_task_handler`'s reconnect loop re-establishes the websocket before each retry. `assert` at entry. Plus the import-fallback pattern (`websocket_connect` and `State` typed as `<type> | None` due to `try/except ModuleNotFoundError`) — `__init__` already raises if either is None, so the `_connect_websocket` use site uses an `assert`. Also a `Language | str` cast (same shape as last batch).
- `transports/websocket/client.py`: `_client_task_handler`'s call chain — `connect()` only starts the task after `_websocket` is assigned, and `disconnect()` cancels the task before clearing the field. `assert` at entry. Plus four `WebSocketClientProtocol` deprecated-alias errors — same `# pyright: ignore[reportAttributeAccessIssue]` as the `services/websocket_service.py` fix from earlier in this PR. The full migration to `websockets.ClientConnection` is left for a separate piece of work.
- `transports/whatsapp/client.py`: `_validate_whatsapp_webhook_request` was typed `bytes`/`str` but called with `bytes | None`/`str | None`; widened the helper signature and pushed the explicit None-check inside (matching its existing empty-string check). Also handled `pipecat_connection.get_answer()` returning `None` — would have crashed at `.get("sdp")` before.
- `services/deepgram/flux/base.py`: `event = data.get("event")` flowed into `_handle_turn_resumed(event: str)` as `Any | None`; added an `isinstance(event, str)` guard before the `FluxEventType(event)` lookup. Plus a latent `TypeError` crash where `average_confidence > min_confidence` was evaluated even when `average_confidence` was `None` — restored the original `not min_confidence` (which treats both `None` and `0.0` as "no filter") and added an explicit drop-on-missing-confidence-data branch.

**`gemini_live` deprecated-InputParams `None` translation.** The deprecated `InputParams` declares `modalities: GeminiModalities | None` and `media_resolution: GeminiMediaResolution | None`, but their downstream usage at `services/google/gemini_live/llm.py:952,959` calls `.value` on each — `None` would crash. Rather than touching the deprecated input model, translate `None` to the canonical defaults (`GeminiModalities.AUDIO`, `GeminiMediaResolution.UNSPECIFIED`) at the assignment site in `vertex/llm.py`. Also fixed an unrelated annotation bug: `_get_credentials` was annotated `-> str` but actually returns `service_account.Credentials` (used correctly by the caller — only the annotation was wrong).

**`moondream/vision.py`.** `frame.format is None` early-return; the other two errors are pyright thinking the moondream2-custom `encode_image`/`query` methods are `Tensor` (they're provided by the model code via `trust_remote_code=True` and aren't visible on the base `AutoModelForCausalLM` type). Scoped `# pyright: ignore[reportCallIssue]` on those.

**`transports/base_output.py`.** Two `self._mixer.mix(...)` calls inside the `with_mixer` closure (invoked only when `self._mixer` is truthy) — captured the mixer to a local with an `assert` inside the closure. Third was the PIL `Image.frombytes(mode, ...)` shape — `frame.format is None` early-return guard at the top of `resize_frame` so the main resize logic reads cleanly.

**`elevenlabs/tts.py`.** The payload-building dicts at L926/L1271 were typed `dict[str, str | dict[str, float | bool]]` — an aspirational shape that matched only the first few assignments. Subsequent code assigned `list[dict[...]]` (pronunciation locators) and bools, all violating the annotation. Widened to `dict[str, Any]` — the honest shape for a JSON request payload and what similar code uses elsewhere.

**Files dropped from the ignore list (57 → 49):** `services/deepgram/flux/base.py`, `services/elevenlabs/tts.py`, `services/google/gemini_live/vertex/llm.py`, `services/moondream/vision.py`, `services/openai/stt.py`, `transports/base_output.py`, `transports/websocket/client.py`, `transports/whatsapp/client.py`.

## Pyright error count

Measured with `pyrightconfig.full.json` (the regular config but with no ignore list, so all source files are checked):

| Run | Errors |
|-----|--------|
| Pre-PR (main) | 947 |
| Post-PR | 496 |
| **Net** | **−451** |

## Pyright ignore-list size

Files in `pyrightconfig.json`'s `ignore` array (excluding `"tests"`):

| Run | Ignored files |
|-----|---------------|
| Pre-PR (main) | 110 |
| Post-PR | 49 |
| **Net** | **−61** |

## Test plan

- [x] `uv run pytest tests/test_llm_service.py` — all 4 tests pass, including the new `TestUnparameterizedSubclass.test_unparameterized_subclass_instantiates`.
- [x] `uv run pytest tests/test_piper_tts.py tests/test_websocket_service.py` — covers the two tweaked services that have unit tests in-repo.
- [x] `uv run pyright` (with the regular ignore list) is clean.
- [x] `uv run ruff check` clean.
- [x] Smoke-test against live Anthropic and AWS Nova Sonic services to confirm no runtime regressions.


src/pipecat/services/aws/nova_sonic/llm.py
```diff
@@ -606,9 +612,18 @@ async def _start_connecting(self):
...
+        if not self._context:  # should never happen
+            return
         # Check for set of completed function calls in the context
         for message in self._context.get_messages():
-            if message.get("role") and message.get("content") not in ["IN_PROGRESS", "CANCELLED"]:
+            # LLMSpecificMessages are opaque provider-specific payloads, not
+            # standard tool-result messages — skip them.
+            if isinstance(message, LLMSpecificMessage):
+                continue
+            if message.get("role") == "tool" and message.get("content") not in [
```
kompfner, line 623: Forgetting to check that role is "tool" was a mistake before. A mostly harmless one (since we then check for "tool_call_id"), but still, not intended. Fixed now.

src/pipecat/services/resembleai/tts.py
```diff
@@ -409,7 +409,9 @@ async def _process_messages(self):
 
                 await self.push_frame(TTSStoppedFrame(context_id=context_id))
                 await self.stop_all_metrics()
-                await self.push_error(ErrorFrame(error=f"{self} error: {error_name} - {error_msg}"))
+                await self.push_error_frame(
```
kompfner, line 412: this was an actual bug.

there are various more minor fixes like this scattered throughout this pr.

src/pipecat/services/websocket_service.py
```diff
@@ -42,7 +42,7 @@ def __init__(self, *, reconnect_on_error: bool = True, **kwargs):
             reconnect_on_error: Whether to automatically reconnect on connection errors.
             **kwargs: Additional arguments (unused, for compatibility).
         """
-        self._websocket: websockets.WebSocketClientProtocol | None = None
+        self._websocket: websockets.WebSocketClientProtocol | None = None  # pyright: ignore[reportAttributeAccessIssue]
```
kompfner, line 45: `WebSocketClientProtocol` is deprecated and so isn't formally exposed through `websockets` (though it still works at runtime). Updating to the more modern `CilentConnection` and validating might be a slightly bigger task which can be done in a future commit.

src/pipecat/services/assemblyai/stt.py
```diff
@@ -233,10 +233,11 @@ def __init__(
                 sample_rate = connection_params.sample_rate
                 encoding = connection_params.encoding
                 default_settings.model = connection_params.speech_model
-                default_settings.formatted_finals = connection_params.formatted_finals
-                default_settings.word_finalization_max_wait_time = (
-                    connection_params.word_finalization_max_wait_time
-                )
+                # Note: `formatted_finals` and `word_finalization_max_wait_time`
```
kompfner, line 236: this was an actual bug

src/pipecat/services/heygen/api_interactive_avatar.py
```diff
@@ -210,11 +210,11 @@ async def new_session(self, request_data: NewSessionRequest) -> StandardSessionR
             "quality": request_data.quality,
             "avatar_id": request_data.avatar_id,
             "voice": {
-                "voice_id": request_data.voice.voiceId if request_data.voice else None,
+                "voice_id": request_data.voice.voice_id if request_data.voice else None,
```
kompfner, line 213: these were actual bugs

kompfner:
Whoops, looks like we have a few "active" pyright errors from files that were maybe prematurely removed from the "ignore" list.

kompfner:
> Whoops, looks like we have a few "active" pyright errors from files that were maybe prematurely removed from the "ignore" list.

Oh, I bet this has to do with the fact of a mismatch between my local pyright version and the one being used in CI...

kompfner:
> > Whoops, looks like we have a few "active" pyright errors from files that were maybe prematurely removed from the "ignore" list.
> 
> Oh, I bet this has to do with the fact of a mismatch between my local pyright version and the one being used in CI...

Nope, it's because we weren't installing extra deps in our CI job. Running pyright on all our source files requires installing everything.

.github/workflows/format.yaml
```diff
@@ -32,7 +32,9 @@ jobs:
         run: uv python install 3.12
 
       - name: Install development dependencies
-        run: uv sync --group dev --extra daily --extra tracing
+        # `--all-extras` (matching the dev setup in README.md) so pyright can
+        # resolve types from various optional dependencies.
+        run: uv sync --group dev --all-extras --no-extra gstreamer --no-extra local
```
kompfner, line 37: Worth noting this change

markbackman:
I had Codex take a look and it found:
```
• The patch introduces a backward-incompatible runtime failure in Google STT's public language conversion helper for list inputs that were previously supported.
  Other inspected changes appear mostly type-checking or guard improvements.

  Review comment:

  - [P2] Preserve list inputs for Google STT language conversion — /Users/mbackman/src/pipecat/src/pipecat/services/google/stt.py:637-637
    When external code calls the previously documented language_to_service_language([Language.EN_US, ...]) overload, this now passes the list into
    language_to_google_stt_language, which reaches resolve_language and raises TypeError: unhashable type: 'list' instead of returning a list of Google language
    codes. This method is public and its prior implementation explicitly supported multi-language conversion, so removing the list branch is a backward-incompatible
    regression even though _get_language_codes() handles the internal settings path.
```

Now having Claude take a look. I'm curious how the two compare...

markbackman:
## Code review

markbackman (approved):
Amazing! I looked through but didn't see anything jump out at me. I also had both Claude and Codex take a look. Claude didn't find anything and Codex flagged one item. Approving, assuming you'll look at the Codex finding.

kompfner:
> Amazing! I looked through but didn't see anything jump out at me. I also had both Claude and Codex take a look. Claude didn't find anything and Codex flagged one item. Approving, assuming you'll look at the Codex finding.

Thanks! Looking into the potential regression now...

kompfner:
> > Amazing! I looked through but didn't see anything jump out at me. I also had both Claude and Codex take a look. Claude didn't find anything and Codex flagged one item. Approving, assuming you'll look at the Codex finding.
> 
> Thanks! Looking into the potential regression now...

OK, so prior to this PR we had this signature:

```python
def language_to_service_language(self, language: Language | list[Language]) -> str | list[str]:
```

This PR brings it into alignment* with its superclass's definition, and all other `language_to_service_language` definitions:

```python
def language_to_service_language(self, language: Language) -> str:
```

(*it's actually a bit _more_ strict (it omits `| None` from the return value), which is A-OK from a typing standpoint when overriding a superclass method.)

In our internal Pipecat code, we were *not* relying on Google STT's `language_to_service_language` being able to take a list of languages.

Externally...well, since it is a public method (though not one I'd suspect would get much usage), it's technically _possible_ that someone was relying on Google STT's `language_to_service_language` going against the `language_to_service_language` contract, but...I think I'm OK with the small risk. It was always akin to relying on a bug rather than a feature.

kompfner:
> > > Amazing! I looked through but didn't see anything jump out at me. I also had both Claude and Codex take a look. Claude didn't find anything and Codex flagged one item. Approving, assuming you'll look at the Codex finding.
> > 
> > 
> > Thanks! Looking into the potential regression now...
> 
> OK, so prior to this PR we had this signature:
> 
> ```python
> def language_to_service_language(self, language: Language | list[Language]) -> str | list[str]:
> ```
> 
> This PR brings it into alignment* with its superclass's definition, and all other `language_to_service_language` definitions:
> 
> ```python
> def language_to_service_language(self, language: Language) -> str:
> ```
> 
> (*it's actually a bit _more_ strict (it omits `| None` from the return value), which is A-OK from a typing standpoint when overriding a superclass method.)
> 
> In our internal Pipecat code, we were _not_ relying on Google STT's `language_to_service_language` being able to take a list of languages.
> 
> Externally...well, since it is a public method (though not one I'd suspect would get much usage), it's technically _possible_ that someone was relying on Google STT's `language_to_service_language` going against the `language_to_service_language` contract, but...I think I'm OK with the small risk. It was always akin to relying on a bug rather than a feature.

Looks like `language_to_service_language` was updated to take a list of languages in [this commit](https://github.com/pipecat-ai/pipecat/pull/1197/changes/32d8f6153f2f6a619563271795c01f2b546e732f), but, looking at the commit carefully, it looks like it wasn't even necessary then...

Looking through the comments in the [associated PR](https://github.com/pipecat-ai/pipecat/pull/1197), it doesn't seem like there was any specific ask for this functionality, either...

PR #4380, opened 2026-04-27
Smart Text Handling

filipi87:
## Summary                                                                                                                                      
            
- Introduced ordered slot tracking (`_AggregatedFrameSlot`) for every `AggregatedTextFrame` that passes through the TTS pipeline, ensuring skipped frames (e.g. code blocks) are emitted downstream only after all preceding spoken frames complete — preserving correct assistant context ordering.                                                                                                                    
- Added `WordCompletionTracker` integration to map incoming word timestamps back to their source `AggregatedTextFrame`.                                                                                        
- Propagated `raw_text` (from `PatternMatch.full_match`) through `AggregatedTextFrame`         
- Fixed `BaseOutputTransport` frame scheduling to preserve insertion order when two frames share the same timestamp                             
- Stripped Cartesia SSML tags (`spell`, `emotion`, `break`, `volume`, `speed`) from word timestamps before processing.                                              
- Fixed `_includes_inter_frame_spaces` handling and trailing punctuation consumption in word timestamp processing.


src/pipecat/processors/frameworks/rtvi/observer.py
```diff
@@ -510,13 +510,16 @@ async def _send_aggregated_llm_text(self, frame: AggregatedTextFrame):
         ):
             return
 
-        text = frame.text
+        text = frame.raw_text or frame.text
```
filipi87: Should the RTVI receive the raw version or just the text ? Or both ?

src/pipecat/services/tts_service.py
```diff
@@ -1155,6 +1261,26 @@ async def _add_word_timestamps(
...
+                )
+                if active and active.tracker.add_word_and_check_complete(word):
+                    active.complete = True
+                    logger.debug(
+                        f"{self} Last word before flushing skipped aggregated frames: {frame.text}"
+                    )
+                    await self._flush_aggregated_text_frame_sequence(last_word_pts=frame.pts)
+                    # TODO: here now we should emit a new frame, to use the raw_content in the context
+                    # We need to also consider the case of interruption, but in this case, we need to be aware
+                    # until how far we have spoken, to maybe replace only part of the context.
```
filipi87: TODO: Still need to implement this piece.

src/pipecat/services/tts_service.py
```diff
@@ -1106,6 +1213,7 @@ async def add_word_timestamps(
         """
         if context_id and self.audio_context_available(context_id):
             for word, timestamp in word_times:
+                logger.info(f"Adding word timestamp to context: {word} @ {timestamp}")
```
filipi87: TODO: only for testing. Remove it.

src/pipecat/utils/context/word_completion_tracker.py
```diff
@@ -0,0 +1,64 @@
...
+    def add_word_and_check_complete(self, word: str) -> bool:
+        """Record a spoken word from a word-timestamp event.
+
+        Args:
+            word: A single word token returned by the TTS service.
+
+        Returns:
+            True when all expected content has been covered.
+        """
+        logger.info(f"WordCompletionTracker add_word_and_check_complete: {word}")
```
filipi87: TODO: only for testing. Remove it.

src/pipecat/processors/aggregators/llm_response_universal.py
```diff
@@ -1321,9 +1322,14 @@ async def _handle_text(self, frame: TextFrame):
         if len(frame.text) == 0:
             return
 
+        text = (
+            frame.raw_text
+            if isinstance(frame, AggregatedTextFrame) and frame.raw_text
+            else frame.text
```
filipi87, line 1503: We are using the `raw_text` to add to the context when it is available.

filipi87:
@markbackman , @mattieruth, it is still in progress, I need to run more tests across all TTS systems, but it should be ready for initial review.

src/pipecat/processors/aggregators/llm_text_processor.py
```diff
@@ -85,7 +86,9 @@ async def _handle_llm_text(self, in_frame: LLMTextFrame):
             out_frame = AggregatedTextFrame(
                 text=aggregation.text,
                 aggregated_by=aggregation.type,
+                raw_text=aggregation.full_match if isinstance(aggregation, PatternMatch) else None,
```
mattieruth: is there a reason to have raw_text be set to `None` vs always setting it to the `full_match` or `text`. then downstream would just always use one or the other and not have to check if it exists.

src/pipecat/processors/frameworks/rtvi/observer.py
mattieruth: It should be the text. The tags are not needed as the information they might communicate is already communicated via the `aggregated_by` field.

```
llm: "You can email support at <email>help@daily.co<email>"
aggregations: [{
    text: "You can email support at",
    raw_text: "You can email support at",
    aggregated_by: sentence
  },{
    text: "help@daily.co",
    raw_text: <email>help@daily.co<email>,
    aggregated_by: email
  }
]
tts: speak all, based on `text`
     "You can email support at help@daily.co"
context: add all, based on `raw_text`
      You can email support at <email>help@daily.co<email>
client: renders based on `text` and `aggregated_by`:
       <p>bot: "You can email support at <a href="mailto:help@daily.co">help@dialy.co</a>.
```

filipi87: Maybe also the accumulated spoken text ?

src/pipecat/services/tts_service.py
```diff
@@ -901,10 +993,13 @@ async def _push_tts_frames(
         type = src_frame.aggregated_by
         text = src_frame.text
 
+        # Create context ID and store metadata
+        context_id = self.create_context_id()
```
mattieruth, line 984: side note: i know you haven't changed the behavior here, but does this mean that we currently do not implement the context_id in the way it's intended? My understanding from @markbackman was that the TTS services want you to use this sort of in a way that I would think would parallel the user's turn. i.e. we would only generate and start a new context id when the user stopped speaking vs. how this is, which is at the start of each aggregated section (most commonly, per sentence).

src/pipecat/services/tts_service.py
```diff
@@ -1122,6 +1234,72 @@ async def add_word_timestamps(
...
+        raw_overflow_word: str | None,
+        pts: int,
+    ):
+        """Emit a TTSTextFrame for the overflow portion of a word that straddles two slots.
+
+        When a TTS word spans the boundary between two AggregatedTextFrames (e.g.
+        ``"<spell>1234</spell>And"`` where ``"1234"`` closes one frame and ``"And"``
+        opens the next), the overflow suffix is fed into the next slot's tracker and
+        emitted as its own TTSTextFrame. If the overflow word also completes the next
+        slot, that slot is marked complete and any waiting skipped frames are flushed.
```
mattieruth: this totally makes sense. But in theory, this is sort of a work-around for a TTS bug, right? The TTS really shouldn't do this. Should we maybe mention that in this comment to make it clear that this is a work-around and could maybe go away if the TTS providers all fixed these issues? (and by TTS providers, I think we really just mean Cartesia. or do other providers do this too?)

src/pipecat/services/tts_service.py
```diff
@@ -1146,16 +1324,43 @@ async def _add_word_timestamps(
...
+                )
                 self._word_last_pts = frame.pts
                 await self.push_frame(frame)
 
+                # When all expected chars are covered the slot is marked complete and
+                # any skipped frames queued behind it (e.g. code blocks) are flushed.
+                if is_complete:
+                    active.complete = True
+                    await self._flush_aggregated_text_frame_sequence(last_word_pts=pts)
+                    if raw_overflow_word:
```
mattieruth: can a raw_overflow_word only exist at the end of a complete slot? I would think it could happen mid-slot and shouldn't be nested under `is_complete`

mattieruth (commented):
i think i'm grasping what's happening. I'm realizing I missed the changes recently wrt audio_context which I'm still trying to understand and makes this a little hand wavy, but i think this is sound. just a few comments and now i need to test.

src/pipecat/processors/aggregators/llm_text_processor.py
filipi87: I thought the only case where it would be needed is when we have a full match. This way, the text and the full match will be different.
But there is no particular reason. I think we can do what you’re suggesting. 👍

src/pipecat/processors/frameworks/rtvi/observer.py
filipi87: Makes sense. I have already fixed it to send just the text.

src/pipecat/services/tts_service.py
filipi87: The actual implementation on `create_context_id` is this:
 ```
   def create_context_id(self) -> str:
        """Generate or reuse a context ID based on concurrent TTS support.

        Returns:
            A context ID string for the TTS request.
        """
        if self._reuse_context_id_within_turn and self._turn_context_id:
            self._refresh_audio_context(self._turn_context_id)
            return self._turn_context_id
        return str(uuid.uuid4())
```

We usually reuse the same `context_id` within a single user turn. After some refactoring, the method name has become a bit misleading.

File: src/pipecat/services/tts_service.py (hunk 8 above)
Comment (filipi87) [reply src/pipecat/services/tts_service.py]: I have noticed this with Cartesia, but I am not 100% sure whether other TTS providers have the same issue.

I can improve the comments. 👍

File: src/pipecat/services/tts_service.py (hunk 9 above)
Comment (filipi87) [reply src/pipecat/services/tts_service.py]: I believe this comment is outdated after all the refactors that I did. 😅

File: src/pipecat/services/tts_service.py (hunk 7 above)
Comment (mattieruth) [reply src/pipecat/services/tts_service.py:984]: ah. nice.

File: src/pipecat/processors/aggregators/llm_text_processor.py (hunk 6 above)
Comment (filipi87) [reply src/pipecat/processors/aggregators/llm_text_processor.py]: Done!

File: src/pipecat/processors/aggregators/llm_response_universal.py (hunk 5 above)
Comment (mattieruth) [reply src/pipecat/processors/aggregators/llm_response_universal.py:1503]: with the most recent change, I think now you can simply do:
```
text = frame.raw_text
```
(that was kind of the main point. so that the consumers of the text frames can grab which version they want and not think about if it's there or not)

mattieruth (approved):
minor last thing. but this is crazy cool. never thought about approaching it on a character length level. still some word/pattern matching but then character lengths after. nice. hopefully this really gets us to a cool new world.

filipi87:
> with the most recent change, I think now you can simply do:
text = frame.raw_text
(that was kind of the main point. so that the consumers of the text frames can grab which version they want and not think about if it's there or not)

@mattieruth, I thought about this, but I believe we need to keep `raw_text` optional because people may be creating and pushing `AggregatedFrame` instances themselves. In that case, making it required would break the API.

What do you think?

mattieruth:
> > with the most recent change, I think now you can simply do:
> > text = frame.raw_text
> > (that was kind of the main point. so that the consumers of the text frames can grab which version they want and not think about if it's there or not)
> 
> @mattieruth, I thought about this, but I believe we need to keep `raw_text` optional because people may be creating and pushing `AggregatedFrame` instances themselves. In that case, making it required would break the API.
> 
> What do you think?

oh hm. that's unfortunate. should we go back to not setting it then? the whole point of that change and my suggestion was to clean up the consuming end. i dunno. i doubt anyone is making their own yet, but who knows. tossing back to you...

PR #4385, opened 2026-04-29
feat(runner): add session_id to RunnerArguments

markbackman:
This is another step towards deprecating the types in the pipecatcloud python SDK and use the types vended by Pipecat's runner. This will better align local and deployed development. The last missing piece was the session_id, which wasn't supported by the local runner. There's a follow-up pipecatcloud PR which uses these types.

## Summary

- Adds a `session_id: str | None` field to `RunnerArguments` so bots can log/trace a per-session identifier in local development the same way they can in Pipecat Cloud (where it is provided via the `x-daily-session-id` header).
- The local runner now mints a UUID at every `*RunnerArguments` construction site. For paths that already returned a `sessionId` to the caller (Daily `/start`, dial-in webhook), a single UUID is now generated and shared between `runner_args.session_id` and the response body instead of being thrown away.
- The SmallWebRTC `/api/offer` endpoint accepts an optional `session_id` so the `/sessions/{session_id}/...` proxy can thread it through.

This is the prerequisite step for collapsing pipecat-cloud's `SessionArguments` / `*SessionArguments` hierarchy onto the upstream runner types.


filipi87 (approved):
LGTM!

PR #4386, opened 2026-04-29
feat(soniox): update default TTS model to tts-rt-v1

markbackman:
## Summary

- Promotes the Soniox TTS default model from `tts-rt-v1-preview` to the generally available `tts-rt-v1`.

## Test plan

- [x] Run a Pipecat agent using `SonioxTTSService` without overriding `model` and confirm audio is synthesized end to end.


filipi87 (approved):
LGTM!

PR #4387, opened 2026-04-29
feat(serializers): add BandwidthFrameSerializer

kush2shah:
## Summary

Adds `BandwidthFrameSerializer` for [Bandwidth's](https://www.bandwidth.com/) Programmable Voice bidirectional WebSocket media streaming, established via the BXML `<StartStream>` verb. This brings Bandwidth to parity with the existing telephony serializers (Twilio, Telnyx, Plivo, Exotel, Vonage, Genesys).

## What it does

- Parses Bandwidth's `start` / `media` / `stop` event envelopes (`eventType`, `track`, top-level base64 `payload`, `sequenceNumber`)
- Filters to `track="inbound"` so the bot doesn't echo itself back when BXML uses `tracks="both"`
- Encodes outbound audio as `playAudio` events with a nested `media` object (`contentType` + base64 `payload`)
- Emits `{"eventType": "clear"}` on `InterruptionFrame` to flush buffered outbound audio
- Terminates the call via the Bandwidth Voice API on `EndFrame` / `CancelFrame` when `auto_hang_up` is enabled

## Differentiator vs other telephony serializers

Bandwidth supports linear PCM outbound at higher rates than μ-law's 8 kHz. The serializer exposes this via `outbound_encoding="PCM"` + `outbound_pcm_sample_rate` (8000/16000/24000), which can noticeably improve TTS audio quality. Default is PCMU 8 kHz for parity with Twilio/Telnyx.

## Authentication

Uses OAuth 2.0 client credentials, **not** the legacy API username/password Basic Auth (which Bandwidth has deprecated; sunset 2026-12-02). Constructor takes `client_id` and `client_secret`. The serializer mints a Bearer token from `https://api.bandwidth.com/api/v1/oauth2/token` only when needed for hangup (per-call, no caching).

## DTMF

Bandwidth does not deliver DTMF events over the media-stream WebSocket — they go through the BXML `<Gather>` verb webhook. This is documented in the docstring; applications handle DTMF in their own webhook handler.

## Testing

- 20 unit tests covering init/validation, deserialize (start/media/stop/inbound-track-filter/outbound-track-rejection/invalid JSON), serialize (PCMU + PCM 24kHz + interruption + passthrough), and the OAuth-then-hangup flow with a regression test for the token-fetch failure path
- Validated against the live Bandwidth API: OAuth token mint succeeds, hangup endpoint accepts auth/path/method/body (404 with Bandwidth-shaped validation error against a fake call ID — clean signal that everything is correct)
- **End-to-end validated on a real phone call**: BXML response, WebSocket protocol, audio in/out, and call termination all work. Tested with multi-turn conversations.

## Companion PRs

- Example: [pipecat-ai/pipecat-examples#207](https://github.com/pipecat-ai/pipecat-examples/pull/207) — `bandwidth-chatbot/inbound`
- Docs: [pipecat-ai/docs#780](https://github.com/pipecat-ai/docs/pull/780) — `BandwidthFrameSerializer` API reference page

## Notes for reviewers

- Wire-format field names verified against [Bandwidth's StartStream BXML docs](https://dev.bandwidth.com/docs/voice/programmable-voice/bxml/startStream/) and confirmed against real `start` events from a live call
- Hangup endpoint and OAuth flow verified against [Bandwidth's official Python SDK](https://github.com/Bandwidth/python-sdk) (which is generated from their canonical OpenAPI spec)


markbackman:
Thanks for submitting! We'd like to request you add Bandwidth as a community integration:

For new services, this is the best way to get started. This acknowledges that we can only maintain so many services inside Pipecat and helps new services get added as quickly as possible. As we have time and see demand from the community, we can collaborate to move the integration to a service the Pipecat maintainers support directly.

Let me know if you have any questions! (Closing this as community integrations reside in a separate repo.)

PR #4393, opened 2026-04-30
fix(turns): defer LocalSmartTurnAnalyzerV3 import to fix transformers warning

### LINKED ISSUE #4392: user_turn_strategies.py unconditionally imports LocalSmartTurnAnalyzerV3, pulling in transformers/PyTorch at startup
## Summary

`pipecat/turns/user_turn_strategies.py` has an unconditional top-level import:

```python
from pipecat.audio.turn.smart_turn.local_smart_turn_v3 import LocalSmartTurnAnalyzerV3
```

`local_smart_turn_v3.py` in turn imports at module level:

```python
from transformers import WhisperFeatureExtractor
```

This means **any code that imports `UserTurnStrategies` — even just to use it as a dataclass — pulls in the entire `transformers` library**, which prints a warning if PyTorch/TensorFlow/Flax are not installed:

```
None of PyTorch, TensorFlow >= 2.0, or Flax have been found.
Models won't be available and only tokenizers, configuration and file/data utilities can be used.
```

## Import chain

```
pipecat.processors.aggregators.llm_response_universal   (line 81)
  → pipecat.turns.user_turn_strategies
      → pipecat.audio.turn.smart_turn.local_smart_turn_v3
          → transformers.WhisperFeatureExtractor
```

`llm_response_universal` is imported by many pipecat services (e.g. `pipecat.services.google.gemini_live.llm`), so the blast radius is wide — importing almost any LLM service brings in `transformers`.

## Impact

- Applications that don't use smart-turn detection are forced to install `transformers` (and implicitly a deep-learning backend) just to use pipecat's LLM aggregation layer.
- In test environments without PyTorch/TF/Flax, the warning pollutes test output.
- It significantly inflates the effective dependency footprint of pipecat for users who only need VAD- or transcription-based turn detection.

## Potential fixes

**Option 1 — Lazy import (minimal change):**  
Move the `LocalSmartTurnAnalyzerV3` import inside the function/class body that actually uses it, or guard it with `TYPE_CHECKING`:

```python
from typing import TYPE_CHECKING
if TYPE_CHECKING:
    from pipecat.audio.turn.smart_turn.local_smart_turn_v3 import LocalSmartTurnAnalyzerV3
```

And import at runtime only in the specific strategy constructor that needs it.

**Option 2 — Optional extras:**  
Gate `transformers` behind an optional extras group (e.g. `pipecat-ai[smart-turn]`) and raise a clear `ImportError` with install instructions if the user tries to use `LocalSmartTurnAnalyzerV3` without it.

**Option 3 — Deferred import in `local_smart_turn_v3.py`:**  
Move `from transformers import WhisperFeatureExtractor` inside the `__init__` of `LocalSmartTurnAnalyzerV3` so it only loads when the class is actually instantiated.

Option 3 is the smallest diff and eliminates the problem for all callers immediately.

## Environment

- pipecat-ai version: 0.0.104 (issue still present in current `main`)
- Python 3.12

markbackman:
## Summary

- Defers the `LocalSmartTurnAnalyzerV3` import from the top of `user_turn_strategies.py` into `default_user_turn_stop_strategies()` so `transformers` and `onnxruntime` only load when the default smart-turn strategy is actually constructed.

## Why

`pipecat.turns.user_turn_strategies` is imported by `llm_response_universal`, which is imported by most LLM services. The previous top-level import meant any LLM service paid the cost of loading `transformers` (and `onnxruntime`) at startup, and triggered the noisy `transformers` missing-backend warning in environments without PyTorch/TensorFlow/Flax (e.g. test envs).

## Test plan

- [x] `import pipecat.processors.aggregators.llm_response_universal` no longer loads `transformers` or `onnxruntime` (verified via `sys.modules`).
- [x] `default_user_turn_stop_strategies()` still constructs a working `LocalSmartTurnAnalyzerV3` and pulls in `transformers`/`onnxruntime` at that point, as expected.
- [x] `ruff check` and `ruff format --check` pass.

## Fixes

- Fixes #4392

### DIFF
--- a/src/pipecat/turns/user_turn_strategies.py
+++ b/src/pipecat/turns/user_turn_strategies.py
@@ -8,7 +8,6 @@
 
 from dataclasses import dataclass
 
-from pipecat.audio.turn.smart_turn.local_smart_turn_v3 import LocalSmartTurnAnalyzerV3
 from pipecat.turns.user_start import (
     BaseUserTurnStartStrategy,
     ExternalUserTurnStartStrategy,
@@ -44,6 +43,8 @@ def default_user_turn_stop_strategies() -> list[BaseUserTurnStopStrategy]:
     Returns ``[TurnAnalyzerUserTurnStopStrategy(LocalSmartTurnAnalyzerV3)]``.
     Useful when building a custom strategy list that extends the defaults.
     """
+    from pipecat.audio.turn.smart_turn.local_smart_turn_v3 import LocalSmartTurnAnalyzerV3
+
     return [TurnAnalyzerUserTurnStopStrategy(turn_analyzer=LocalSmartTurnAnalyzerV3())]
 
 
--- a/changelog/4393.fixed.md
+++ b/changelog/4393.fixed.md
@@ -1,0 +1,1 @@
+- Fixed an issue where `LocalSmartTurnAnalyzerV3` was imported unconditionally for user turn stop strategies. It is now only imported when `default_user_turn_stop_strategies()` is called. This improves startup time and removes the `transformers` "PyTorch/TensorFlow/Flax not found" warning when the default stop strategies are not used.
--- a/tests/test_krisp_ip_user_turn_start_strategy.py
+++ b/tests/test_krisp_ip_user_turn_start_strategy.py
@@ -12,10 +12,6 @@ from unittest.mock import MagicMock, patch
 
 import numpy as np
 
-# Mock package version check before importing pipecat (development mode)
-_version_patcher = patch("importlib.metadata.version", return_value="0.0.0-dev")
-_version_patcher.start()
-
 # Mock krisp_audio before any pipecat import that loads krisp_instance / VIVA IP strategy
 mock_krisp_audio = MagicMock()
 mock_krisp_audio.SamplingRate.Sr8000Hz = 8000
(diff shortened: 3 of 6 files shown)


src/pipecat/turns/user_turn_strategies.py
```diff
@@ -44,6 +43,85 @@ def default_user_turn_stop_strategies() -> list[BaseUserTurnStopStrategy]:
...
+        print(f"DEBUG version source:\n{src}", flush=True)
+    except Exception as e:
+        print(f"DEBUG version source error: {e}", flush=True)
+    print(f"DEBUG importlib.metadata module file: {importlib.metadata.__file__!r}", flush=True)
+
+    print("DEBUG meta_path finders:", flush=True)
+    for finder in sys.meta_path:
+        print(f"  - {type(finder).__module__}.{type(finder).__name__}", flush=True)
+
+    from pipecat.audio.turn.smart_turn.local_smart_turn_v3 import LocalSmartTurnAnalyzerV3
```
filipi87, line 46: I think this line is all we need for this PR, right ? Can we keep just this one ?

markbackman: I'm trying to get to the bottom of why CI is failing... Hang tight. Found the culprit.

I'll clean up the code and commits.

markbackman: Ok, ready! Commits are all clean now. Sorry for the noise. The CI failure was unexpected.

filipi87 (approved):
LGTM!

PR #4395, opened 2026-04-30
Broaden tool_resources to app_resources

kompfner:
## Summary

Broaden `tool_resources` to `app_resources` for easy access not just in tool handlers but in other places like custom `FrameProcessor`s.

Involves 3 changes:

- A rename: `tool_resources` → `app_resources`
- A new property on `PipelineTask`: `app_resources`
- A new property on `FrameProcessor`: `pipeline_task`

Usage in tool handler:

```python
async def get_weather(params: FunctionCallParams):
    resources = cast(MyAppResources, params.app_resources)
    ...
```

Usage in custom `FrameProcessor`:

```python
class MyProcessor(FrameProcessor):
        if self.pipeline_task is not None:
            resources = cast(MyAppResources, self.pipeline_task.app_resources)
            ...
```

The previous `tool_resources` aliases (on `PipelineTask`, `FunctionCallParams`, and `FrameProcessorSetup`) keep working but are deprecated as of 1.2.0 and emit `DeprecationWarning`s.

See the new `examples/features/features-app-resources.py` for an end-to-end example exercising both read sites against a shared `AppResources` container.

## Test plan

- [x] New tests in `tests/test_app_resources.py` cover: `app_resources` getters/defaults; deprecated `tool_resources` aliases (kwarg, property, FrameProcessorSetup field) all warn and still work; `FrameProcessor.pipeline_task` returns `None` before setup and the live task during; LLMService function-call wiring (handler + direct-function paths); legacy custom processors reading `setup.tool_resources` still receive the value (whether the user passed `app_resources=` or the deprecated `tool_resources=` to `PipelineTask`).
- [x] Full test suite passes locally (modulo the pre-existing `NvidiaSegmentedSTTService` failure on main).
- [x] `ruff check` and `ruff format --check` clean on touched files.


aconchillo (approved):

aconchillo:
LGTM!

PR #4399, opened 2026-05-01
feat(deepgram): expose mip_opt_out on TTS services

vr000m:
## Summary

`DeepgramSTTService` already supports `mip_opt_out` for opting out of Deepgram's Model Improvement Program. The two TTS services didn't, forcing users to subclass. This PR adds the field to both for parity:

- `DeepgramTTSSettings` gains `mip_opt_out: bool | None | _NotGiven` (default `NOT_GIVEN`).
- `DeepgramTTSService._connect_websocket` appends `mip_opt_out=true|false` to the WebSocket URL when set.
- `DeepgramHttpTTSService.run_tts` adds the same key to the aiohttp `params` dict when set.
- `_update_settings` already triggers reconnect on any changed field, so runtime updates are covered.

## Why settings-only (not a constructor kwarg)

STT exposes `mip_opt_out` as a constructor kwarg, but TTS in this file is mid-migration to settings-only configuration (see the deprecated `voice` kwarg with the "use `settings=...` instead" warning). Adding `mip_opt_out` as a constructor kwarg here would extend the deprecation surface. STT hasn't started that migration yet, so it still uses constructor kwargs. The asymmetry will resolve when STT migrates; until then, prefer the canonical path on TTS rather than locking in another deprecation.

## Test plan

- [x] `uv run pytest tests/test_deepgram_tts.py tests/test_deepgram_stt.py` — 21 passed
- [x] `uv run ruff check src/pipecat/services/deepgram/tts.py tests/test_deepgram_tts.py` — clean
- [x] WS URL includes `mip_opt_out=true` / `=false` when set
- [x] HTTP `params` dict includes `mip_opt_out=true` / `=false` when set
- [x] Default `settings=None` and explicit `Settings()` (no kwargs) both omit the param — regression test for the NOT_GIVEN sentinel path

### DIFF
--- a/changelog/4399.added.md
+++ b/changelog/4399.added.md
@@ -1,0 +1,1 @@
+- Added `mip_opt_out` to `DeepgramTTSSettings` (used by both `DeepgramTTSService` and `DeepgramHttpTTSService`) for opting out of Deepgram's Model Improvement Program. Pass it via `settings=DeepgramTTSService.Settings(mip_opt_out=True)` to mirror the existing flag on `DeepgramSTTService`.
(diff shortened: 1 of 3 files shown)


markbackman:
Superseded by https://github.com/pipecat-ai/pipecat/pull/4400.

PR #4401, opened 2026-05-01
fix(xai/realtime): pass model as query param on connect

markbackman:
## Summary

- Fixed `GrokRealtimeLLMService` ignoring the configured model. xAI's Voice Agent API selects the model via a `?model=` query parameter on the WebSocket URL and does not honor a model field in `session.update`, so every session was silently falling back to the deprecated `grok-voice-fast-1.0`.
- `_connect()` now appends `?model=<model>` (URL-encoded) to `base_url` when opening the WebSocket.
- Default `Settings.model` is now `grok-voice-think-fast-1.0`, the model xAI currently recommends.

## Testing

- Run a Grok Realtime example and confirm the WebSocket connects to `wss://api.x.ai/v1/realtime?model=grok-voice-think-fast-1.0`.
- Pass a custom model via `settings=GrokRealtimeLLMService.Settings(model="...")` and confirm the URL reflects it.


filipi87 (approved):
LGTM!

PR #4404, opened 2026-05-01
Mitigate tool-call-related hallucination

kompfner:
## Summary

When tools change mid-conversation, LLMs can produce a few different flavors of tool-call-related hallucination: calling tools that have been removed, avoiding tools that have been re-added, or hallucinating output (made-up answers or tool-call-shaped non-tool-calls) when tools are unavailable.

This PR introduces an opt-in **`add_tool_change_messages`** flag on the LLM aggregators (preferred entry point: `LLMContextAggregatorPair(..., add_tool_change_messages=True)`) that appends a developer-role message to the context whenever `LLMSetToolsFrame` changes the set of advertised standard tools. Helps the LLM stay coherent across tool changes by spelling out exactly what just became available or unavailable. Both aggregators participate; whichever handles the frame first wins, and the other (if any) sees an empty diff against the shared context and stays silent — order-independent regardless of whether the frame flows downstream or upstream.

It also tightens the existing missing-handler path that #4301 introduced:

- Reworded the terminal tool result returned to the LLM to a neutral `"The function \`X\` is not currently available."` (overridable via `LLMService.MISSING_FUNCTION_CALL_MESSAGE_TEMPLATE`). Previously read `"Error: function 'X' is not registered."`
- Logs at the call site now distinguish developer error (tool advertised but no handler registered → `logger.error`) from hallucination (tool not advertised → `logger.warning`).

Two examples accompany the change:

- `examples/features/features-add-tool-change-messages.py` — manual validation harness for the new `add_tool_change_messages` mitigation. Defaults the flag to ON; flips tool availability on a turn counter so its effect can be observed end-to-end with the flag on vs. off.
- `examples/function-calling/function-calling-missing-handler.py` — deterministic demo of the missing-handler path. Deliberately advertises a tool without registering its handler (what happens when a developer forgets to call `register_function`).

## Test plan

- [x] `uv run pytest tests/test_llm_service.py tests/test_context_aggregators_universal.py`
- [x] Full suite: `uv run pytest` (passes apart from one pre-existing failure unrelated to this change)
- [x] Manual validation with the `add_tool_change_messages` harness, A/B with the flag on vs. off — flag-on noticeably improves LLM coherence after tool removal (no made-up answers, no tool-call-shaped JSON masquerading as text)


kompfner:
To be perfectly honest: in my testing, I was not actually able to coax the LLM into hallucinating calls to unavailable tools, or avoiding calling an available tool.

However, I *was* able to see it hallucinate in all sorts of other ways when tools that were previously available became unavailable (like just making up answers, or answering with weird JSON blobs that *looked* like tool calls but weren't), so I was able to validate that the `add_tool_change_messages` mechanism really had a huge impact in mitigating those hallucinations.

src/pipecat/processors/aggregators/llm_response_universal.py
```diff
@@ -248,20 +275,87 @@ class LLMContextAggregator(FrameProcessor):
     common functionality for context-based conversation management.
     """
 
-    def __init__(self, *, context: LLMContext, role: str, **kwargs):
+    # Developer-role messages appended to the context when tools are added/
+    # removed via ``LLMSetToolsFrame`` (only when ``add_tool_change_messages``
+    # is enabled on the aggregator's params). ``{function_names}`` is
+    # substituted with a sorted, comma-separated, backtick-wrapped list.
+    TOOL_ACTIVATION_MESSAGE_TEMPLATE = (
```
markbackman, line 282: Out of curiosity, have you found it to be more effective to notify about the both the active an inactive messages? Or, would just reinforcing what's active be sufficient? I'm assuming both are needed but am curious about what you learned in working on this.

markbackman (approved):
Looks really good. I haven't had a chance to test it, but I reviewed and had Claude take a look and everything looks good.

The one thing I was wondering about is why does the active/inactive logic live on the LLMContextAggregator base, but since handling the LLMSetToolsFrame is on the LLMContextAggregator base, this makes sense.

kompfner:
> Looks really good.

Thanks! ❤️ 

> The one thing I was wondering about is why does the active/inactive logic live on the LLMContextAggregator base, but since handling the LLMSetToolsFrame is on the LLMContextAggregator base, this makes sense.

Yup! Someone could technically send a `LLMSetToolsFrame` from either direction, so both user and assistant context aggregators need to be prepared to handle it.

Admittedly it's a bit awkward that you could feasibly add `add_tool_change_messages` on one of your aggregators and forget to do so on the other. But the `add_tool_change_messages` flag on the aggregator pair (the recommended way of doing it) hopefully mitigates that. The alternative I considered was putting `add_tool_change_messages` directly on the `LLMContext`, but that felt like a violation of the (implicit) precedent of keeping `LLMContext`'s initializer args "dumb" and laser-focused on *content* and not behavior.

src/pipecat/processors/aggregators/llm_response_universal.py
kompfner: Honestly, I didn't test injecting a message simply listing the new, changed set of available tools—I was a bit concerned about the case of many many tools being in the list, in which case the "diff" felt both more compact and more narrowly focused on what matters. The LLM needs to know what changed, to start or stop behaving a certain way. If it's looking at a long list, that's a lot of noise to sift through and the crux of the message (e.g. "function foo went away") might be lost.

kompfner:
@aconchillo raised a good point today - we ought to mention this new flag in the Flows docs, as Flows does a lot of adding/removing tools on the fly.

PR #4405, opened 2026-05-01
Split user-turn-stop into inference-triggered and finalized events

aconchillo:
## Summary

- Fixes premature `on_user_turn_stopped` firing when `filter_incomplete_user_turns` is enabled — the event now waits on the LLM's completion verdict instead of firing on the smart-turn detector's tentative stop.
- Splits stop into two events: `on_user_turn_inference_triggered` (kicks LLM inference) and `on_user_turn_stopped` (semantic finalization). Default strategies still fire both at the same call site.
- Adds `LLMTurnCompletionUserTurnStopStrategy`, `UserTurnInferenceCompletedFrame`, and a `deferred(...)` wrapper so a detector can drive inference without owning finalization.
- Extracts `ExternalUserTurnCompletionStopStrategy` as a reusable base for any external completeness signal (STT, end-of-turn classifiers, custom code) — `LLMTurnCompletionUserTurnStopStrategy` is now a thin subclass that adds the LLM marker-protocol setup.
- Adds `FilterIncompleteUserTurnStrategies` — a `UserTurnStrategies` specialization that wraps the detector chain with `deferred(...)` and appends `LLMTurnCompletionUserTurnStopStrategy` as the finalizer. Common case: `user_turn_strategies=FilterIncompleteUserTurnStrategies()`.
- Adds `LLMMarkerFrame` (sideband marker frame) and routes turn-completion markers through it instead of `LLMTextFrame(skip_tts=True)`. `append_to_context_immediately` controls whether the marker is committed as its own assistant message (○ / ◐) or merged with the upcoming aggregation as a prefix (✓). Fixes a case where ○ / ◐ markers were aggregated by the assistant aggregator but never committed to the context because the assistant turn lifecycle didn't run to completion.
- Preserves the full user transcript when multiple inferences fire in one turn. Each inference's segment is concatenated into `_full_user_turn_aggregation` (instead of overwriting `_pending_user_turn_aggregation`), so the eventual `on_user_turn_stopped` event surfaces everything the user said across the whole turn.
- Renames `UserTurnCompletedFrame` to `UserTurnInferenceCompletedFrame` to avoid semantic overlap with `UserStoppedSpeakingFrame` — the former is the post-judgment "inference about the turn is now complete" signal; the latter is the acoustic stop signal.
- Deprecates `LLMUserAggregatorParams.filter_incomplete_user_turns` (auto-translated to `FilterIncompleteUserTurnStrategies` with a `DeprecationWarning`).

## Test plan

- [x] `tests/test_context_aggregators_universal.py` — covers default-flow event ordering, the deprecation warning, deprecation auto-installation, gated-finalization end-to-end (turn waits for `UserTurnInferenceCompletedFrame` before stopping), and multi-inference transcript preservation.
- [x] `tests/test_user_turn_controller.py` — covers new event propagation and `deferred(...)` suppressing finalization.
- [x] `tests/test_user_turn_completion_mixin.py` — covers `LLMMarkerFrame` emission for ✓ / ○ / ◐ and `UserTurnInferenceCompletedFrame` broadcast on ✓ only.
- [x] Full test suite passes.
- [x] `uv run ruff check` and `uv run ruff format --check` clean.
- [x] Manual smoke test in a real voice pipeline with the new `FilterIncompleteUserTurnStrategies()` helper.


markbackman (commented):
I'll take a closer look tomorrow.

One thing while I'm thinking... This split makes `on_user_turn_stopped` mean “semantically finalized,” which can now happen after LLM completion rather than when inference starts. That seems right for this feature, but it may affect latency observers or custom handlers that previously treated user-turn-stop as the start of bot latency. It might be worth checking whether those should bind to `on_user_turn_inference_triggered` instead.

aconchillo:
> I'll take a closer look tomorrow.
> 
> One thing while I'm thinking... This split makes `on_user_turn_stopped` mean “semantically finalized,” which can now happen after LLM completion rather than when inference starts. That seems right for this feature, but it may affect latency observers or custom handlers that previously treated user-turn-stop as the start of bot latency. It might be worth checking whether those should bind to `on_user_turn_inference_triggered` instead.

I think we have always expected `on_user_turn_stopped` to really be the end of user turn. The problem is that we never got it right. Whether we compute that mechanically or semantically it doesn't really matter, is when we really think the user is done talking. I see the filter incomplete turn as an optimization because we use the same LLM that will output the answer (which is the cool part), but we could have done this in a smaller separate LLM (inside a new turn analyzer).

src/pipecat/frames/frames.py
```diff
@@ -970,6 +970,24 @@ class UserSpeakingFrame(SystemFrame):
     pass
 
 
+@dataclass
+class UserTurnCompletedFrame(SystemFrame):
```
markbackman: This is great. In this PR, we should also update the STT services that are pushing UserStoppedSpeakingFrame with this new frame. (or perhaps in a fast follow). This feature (filter incomplete user turns) breaks with this change for STT services like Flux because they are currently only compatible with External turn strategies. It would be best to both ship this feature and update the STT services in the same release.

markbackman:
> I think we have always expected on_user_turn_stopped to really be the end of user turn. The problem is that we never got it right. Whether we compute that mechanically or semantically it doesn't really matter, is when we really think the user is done talking. I see the filter incomplete turn as an optimization because we use the same LLM that will output the answer (which is the cool part), but we could have done this in a smaller separate LLM (inside a new turn analyzer).

True. @filipi87 and I talked about this. If we weren't trying to optimize latency, the end of turn determination would have been done by a separate LLM before the conversation inference occurs. I think this is OK. The only case where it's weird is when the user has a complete response. That is: they were actually done speaking at a time before the LLM said so. We don't backdate the time (and I don't think we should). I'm just stating what we lose by using this.

Anyway, I'm onboard with this approach now.

markbackman:
I see a regression in running this. For an incomplete result, I see:
```
{'role': 'user', 'content': 'Well,'}, {'role': 'user', 'content': "I'd go to France."}, {'role': 'assistant', 'content': '◐✓ France is a fantastic choice!...
```

The issue being that the incomplete message is not maintained as its own message. Running from `main` yields:
```
{'role': 'user', 'content': 'Well, I would'}, {'role': 'assistant', 'content': '◐'}, {'role': 'user', 'content': 'Go to France.'}
```

You can see how there's an assistant message when the turn marker is emitted. This is the expected behavior.

This is likely due to the same turn semantics. I'm not sure of the best way of dealing with this. We need the turn marker to always be emitted downstream and added to the context. Without it, the LLM will learn from the context and break the system_instruction rules.

markbackman:
I'm also seeing that the incomplete timeout created by ◐ survives after the user continues and after a later ✓ completes the turn. That stale timeout injects the “user has been quiet” prompt after it is no longer valid.

src/pipecat/frames/frames.py
aconchillo: Yep. I was planning to do it in a followup PR, but it might makes sense to do it in this one.

src/pipecat/frames/frames.py
```diff
@@ -970,6 +1004,24 @@ class UserSpeakingFrame(SystemFrame):
     pass
 
 
+@dataclass
+class UserTurnInferenceCompletedFrame(SystemFrame):
```
markbackman, line 1008: Much clearer naming 🙇

markbackman:
## Code review

markbackman (approved):
LGTM! I pushed two small clean up commits:
- One to fix an old typo in the duration of the `incomplete_long_timeout`, which was propagated in this PR
- One to add deprecation directives to docstrings + a warning about the `user_turn_completion_config`

Nice clean up 👏

PR #4407, opened 2026-05-02
feat(rtvi): add UI Agent Protocol as first-class RTVI message types

markbackman:
## Summary

Adds the **UI Agent Protocol** as first-class RTVI message types so server-side AI agents can observe and drive a GUI app on the client through structured RTVI messages.

The data layer is the canonical authority for the wire format. Higher-level frameworks (`pipecat-ai-subagents`'s `UIAgent`) build agent abstractions on top, and single-LLM Pipecat apps can target the same wire format directly via custom tools that emit these typed messages. The matching client-side support lives in `@pipecat-ai/client-js` and `@pipecat-ai/client-react`.

### Five new top-level RTVI message types

In kebab-case, in line with the rest of the protocol:

| Type | Direction | Purpose |
| --- | --- | --- |
| `ui-event` | client → server | Named event with payload (e.g. `{name: "nav_click", payload: {...}}`) |
| `ui-snapshot` | client → server | Accessibility tree snapshot of the page |
| `ui-cancel-task` | client → server | Cancel an in-flight user task group |
| `ui-command` | server → client | Named command with payload (e.g. toast/navigate/scroll/highlight/...) |
| `ui-task` | server → client | Task lifecycle envelope (group_started / task_update / task_completed / group_completed) |

Each ships paired `*Data` / `*Message` pydantic models in `rtvi.models`, following the existing RTVI envelope convention (`BotReady` / `BotReadyData`, `Error` / `ErrorData`, etc.). Built-in command payload models (`Toast`, `Navigate`, `ScrollTo`, `Highlight`, `Focus`, `Click`, `SetInputValue`, `SelectText`) ship alongside; the matching default React handlers live in `@pipecat-ai/client-react`.

### Frame and event dispatch (mirrors `client-message`)

The `RTVIProcessor` registers a new `on_ui_message` event handler that fires for inbound `ui-event` / `ui-snapshot` / `ui-cancel-task` with the parsed `Message` envelope, the same way `on_client_message` works for `client-message`.

Five new pipeline frames let observers and processors see UI traffic the same way they see other RTVI messages:

- **Outbound** (downstream code pushes; observer wraps into the matching envelope and sends to the transport):
  - `RTVIUICommandFrame(command_name, payload)` → `UICommandMessage`
  - `RTVIUITaskFrame(data: UITaskData)` → `UITaskMessage` (`UITaskData` is a discriminated union over the four lifecycle kinds)
- **Inbound** (`RTVIProcessor._handle_message` pushes when the message arrives, alongside firing `on_ui_message`):
  - `RTVIUIEventFrame(msg_id, event_name, payload)`
  - `RTVIUISnapshotFrame(msg_id, tree)`
  - `RTVIUICancelTaskFrame(msg_id, task_id, reason)`

Pipeline observers and processors can match on the frame; subscribers like the `pipecat-ai-subagents` bridge keep using the event handler.

### Protocol version bump

`PROTOCOL_VERSION` 1.2.0 → 1.3.0. Purely additive: only new top-level message types are introduced; no existing wire shapes changed. The major-version compatibility check the server runs on `client-ready` still passes for older 1.x clients/servers, so cross-version connections continue to work; older endpoints simply do not exercise the new types.

### Convention notes

- Wire-format strings (`"ui-event"`, `"ui-command"`, etc.) are pinned exclusively via `Literal[...]` field defaults on the `*Message` classes, the same way every other RTVI message type does it. No standalone `*_MESSAGE_TYPE` constants.
- The `kind` discriminator on `ui-task` envelopes is similarly pinned via `Literal[...]` on each `UITask*Data` class.

## Test plan

- [x] New `tests/test_rtvi_ui.py` (18 tests) pins the on-the-wire shape of each envelope and each command payload model. `model_dump()` is asserted against the exact dict every client implementation has to interop with, so any accidental rename or type drift fails loudly.
- [x] Full test suite passes locally.
- [x] `ruff check` and `ruff format --check` clean on touched files.
- [x] Wire-format compatibility verified end-to-end with a working app (`pipecat-ai-subagents`'s `UIAgent` plus `@pipecat-ai/client-react`).

### DIFF
--- a/src/pipecat/processors/frameworks/rtvi/frames.py
+++ b/src/pipecat/processors/frameworks/rtvi/frames.py
@@ -10,6 +10,7 @@ from dataclasses import dataclass
 from typing import Any
 
 from pipecat.frames.frames import SystemFrame
+from pipecat.processors.frameworks.rtvi.models import UITaskData
 
 
 @dataclass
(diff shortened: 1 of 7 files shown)


src/pipecat/processors/frameworks/rtvi/frames.py
```diff
@@ -27,6 +28,134 @@ def __str__(self):
         return f"{self.name}(data: {self.data})"
 
 
+@dataclass
+class RTVIUICommandFrame(SystemFrame):
```
mattieruth, line 32: what do you think about subclassing from ServerMessageFrame and simply adding `command_name` as a parameter (use `data` instead of `payload`). Just as the UIAgentClient wrapping/extending the client, the `ui-event` and `ui-command` messages feel like an optional built-in extension of the core client<->server messages.

src/pipecat/processors/frameworks/rtvi/frames.py
```diff
@@ -27,6 +28,134 @@ def __str__(self):
...
+    data: UITaskData | None = None
+
+    def __str__(self):
+        """String representation of the UI task frame."""
+        kind = getattr(self.data, "kind", "?")
+        return f"{self.name}(kind: {kind})"
+
+
+@dataclass
+class RTVIUIEventFrame(SystemFrame):
```
mattieruth, line 84: you know what i'm going to say.. i think this should be a ClientMessageFrame. It wouldn't even need to define any new paramaters. This really is just a ClientMessage with slightly more specific intentions.

src/pipecat/processors/frameworks/rtvi/models.py
```diff
@@ -549,3 +549,401 @@ class SystemLogMessage(BaseModel):
...
+#
+# where N is the number of workers in the group. The kind strings are
+# pinned as ``Literal[...]`` defaults on the matching ``UITask*Data``
+# class below.
+
+
+# -- UI envelope data classes --
+
+
+class UIEventData(BaseModel):
```
mattieruth, line 594: this could be a `ClientMessage`

src/pipecat/processors/frameworks/rtvi/models.py
```diff
@@ -549,3 +549,401 @@ class SystemLogMessage(BaseModel):
...
+    Parameters:
+        name: App-defined event name.
+        payload: App-defined payload, schemaless by design.
+    """
+
+    name: str
+    payload: Any | None = None
+
+
+class UICommandData(BaseModel):
```
mattieruth, line 606: this could be a `ServerMessage` with an added `name` -- though, maybe `type` would be more consistent?? the client version of these messages enforces the data be it's own type:/data: combo and is sent over the wire under t:/d: -- wondering if we want to parallel that convention here.

src/pipecat/processors/frameworks/rtvi/models.py
```diff
@@ -549,3 +549,401 @@ class SystemLogMessage(BaseModel):
...
+
+    The accessibility snapshot tree is opaque on the server side.
+    The client owns its shape; the server stores it as-is for
+    rendering into the LLM context.
+
+    Parameters:
+        tree: The serialized accessibility tree.
+    """
+
+    tree: Any | None = None
```
mattieruth: in the web-client, this is fully typed and not an `Any`. I would think you could provide proper types here as well.

src/pipecat/processors/frameworks/rtvi/frames.py
markbackman: We talked about changing `command_name` to `command` and not subclassing. I've made those updates.

We also talked about not changing `payload` to `data`, which I've maintained.

src/pipecat/processors/frameworks/rtvi/frames.py
markbackman: Same as above, but `event_name` should be `event`. I made that change.

src/pipecat/processors/frameworks/rtvi/models.py
markbackman: Skipping this suggestion based on the changes above.

src/pipecat/processors/frameworks/rtvi/models.py
markbackman: Same as above.

src/pipecat/processors/frameworks/rtvi/models.py
```diff
@@ -617,18 +617,91 @@ class UICommandData(BaseModel):
...
+    The accessibility snapshot tree mirrors the client-side
+    ``A11ySnapshot`` wire shape and is kept forward-compatible by
+    allowing extra fields on the snapshot models.
 
     Parameters:
         tree: The serialized accessibility tree.
     """
 
-    tree: Any | None = None
+    tree: A11ySnapshot
```
mattieruth, line 704: this was optional before. now it's not? (seems like it shouldn't be, but just making sure this was an intentional change)

mattieruth (approved):
ignore the second comment. you already fixed it and i can't figure out how to get rid of it.

this is looking good. the only high-level remaining question is if any of the ui commands are too web-centric and if they are, is that fine? I'm thinking yes, but worth a second thought.

src/pipecat/processors/frameworks/rtvi/models.py
markbackman: Confirmed that the client will always deliver a snapshot and there are no null/None conditions.

src/pipecat/processors/frameworks/rtvi/models.py
markbackman: I was able to add proper types for the tree. Updated on this branch.

markbackman:
## Code review

PR #4413, opened 2026-05-04
Add Google STT adaptation support

ddallabenetta:
This PR adds native Google Speech-to-Text v2 `adaptation` support to `GoogleSTTService`.

Changes included:
- add `adaptation` to `GoogleSTTSettings`
- support `adaptation` in deprecated `InputParams`
- normalize adaptation payloads from native v2 `phrase_sets` and v1-style `phrase_set_references`
- inject adaptation into the Google STT v2 `RecognitionConfig`
- warn and ignore adaptation for unsupported `telephony` model
- support runtime updates for adaptation settings
- add unit coverage for adaptation normalization and service settings completeness

Validation:
- `uv build`
- `uv run ruff format --check src/pipecat/services/google/stt.py tests/test_google_stt.py`
- `uv run ruff check src/pipecat/services/google/stt.py tests/test_google_stt.py`
- `uv run pyright src/pipecat/services/google/stt.py`
- `uv run pytest tests/test_google_stt.py tests/test_service_init.py`


2010b9:
This would be relevant for our team as well 🙂
@markbackman is this something you are planning on adding?

pficardoso:
This would be an huge improvement. @markbackman any thoughts about merging this feature?

src/pipecat/services/google/stt.py
```diff
@@ -823,24 +908,38 @@ async def _connect(self):
...
+                enable_automatic_punctuation=self._settings.enable_automatic_punctuation,
+                enable_spoken_punctuation=self._settings.enable_spoken_punctuation,
+                enable_spoken_emojis=self._settings.enable_spoken_emojis,
+                profanity_filter=self._settings.profanity_filter,
+                enable_word_time_offsets=self._settings.enable_word_time_offsets,
+                enable_word_confidence=self._settings.enable_word_confidence,
             ),
+        }
+
+        speech_adaptation = self._get_speech_adaptation()
```
markbackman, line 963: `adaptation` dicts are first validated here, inside `_connect()` — which runs *after* the existing stream has already been torn down. A malformed payload leaves the service permanently unable to transcribe.

Failure path for a runtime update via `STTUpdateSettingsFrame`:

1. `apply_update` commits the value into `self._settings` with no validation, and never rolls back.
2. `changed` is non-empty, so `_reconnect_if_needed()` runs — it calls `await self._disconnect()` **first**, which sets `self._streaming_task = None`.
3. `_connect()` reaches this line and the proto constructor raises.
4. `__process_frame` catches it and pushes a **non-fatal** error frame, so the pipeline keeps running.
5. `_streaming_task` stays `None` forever. `run_stt` guards on it and silently discards all audio, and `_reconnect_if_needed` guards on it too, so nothing can ever recover.

A plausible payload does raise — and note it's a `TypeError`, so a caller guarding `except ValueError` wouldn't even catch it:

```
{"phrase_sets": [{"phrases": ["hello"]}]}
# TypeError: Parameter to initialize message field must be dict or instance
# of same class: expected <class 'Phrase'> got <class 'str'>
```

(The required shape is `{"value": "hello"}`, as the new tests use — the bare-string-phrase case is untested.)

The same path also affects startup: a bad `adaptation` passed at construction fails only once the pipeline starts, not at construction like every other `InputParams` field.

Suggested fix: normalize the payload into a `SpeechAdaptation` *before* it is committed — in `__init__`, and in the `_update_settings` override ahead of `super()._update_settings(delta)` — so a malformed payload raises without tearing down a working stream.

Worth noting the "failed `_connect` leaves the service deaf" property is pre-existing (a bad `model` would do the same); what's new here is a user-supplied, structurally complex field whose only validation runs after the teardown.

src/pipecat/services/google/stt.py
```diff
@@ -359,6 +359,57 @@ def language_to_google_stt_language(language: Language) -> str | None:
...
+    if isinstance(adaptation, cloud_speech.SpeechAdaptation):
+        return adaptation
+
+    normalized = dict(adaptation)
+
+    references = normalized.pop("phrase_set_references", None)
+    phrase_sets = list(normalized.get("phrase_sets", []))
+
+    if references:
+        phrase_sets.extend({"phrase_set": ref} for ref in references)
```
markbackman: A bare string here is iterated character-by-character, silently producing one bogus phrase set per character. Both entry points are affected:

```
{"phrase_set_references": "projects/p/locations/global/phraseSets/x"}
# -> 40 phrase_sets: {phrase_set: "p"}, {phrase_set: "r"}, {phrase_set: "o"}, ...

{"phrase_sets": "some/ref"}
# -> 8 phrase_sets: "s", "o", "m", "e", "/", "r", "e", "f"
```

This is an easy slip because the function deliberately accepts a bare string for an individual `phrase_sets` *entry* just below, which teaches the caller that "a phrase set is a string." It's also inconsistent with the function's own contract: `{"phrase_sets": [123]}` raises a clear `ValueError`, but a top-level string produces garbage that only surfaces later as an opaque `InvalidArgument` from Google — or silently does nothing.

Wrapping a bare string matches the leniency already extended to individual entries:

```suggestion
    references = normalized.pop("phrase_set_references", None)
    if isinstance(references, str):
        references = [references]

    raw_phrase_sets = normalized.get("phrase_sets", [])
    if isinstance(raw_phrase_sets, str):
        raw_phrase_sets = [raw_phrase_sets]
    phrase_sets = list(raw_phrase_sets)

    if references:
        phrase_sets.extend({"phrase_set": ref} for ref in references)
```

changelog/4413.changed.md
```diff
@@ -0,0 +1 @@
+- Added Google Speech-to-Text v2 `adaptation` support to `GoogleSTTService`, including phrase set normalization, runtime settings support, and model compatibility handling.
```
markbackman: This fragment should be named `4413.changed.md`.

[CONTRIBUTING.md](https://github.com/pipecat-ai/pipecat/blob/f82b059b3f8637e972f258207f65b657cd284e5b/CONTRIBUTING.md#L76-L78) specifies the naming pattern `<PR_number>.<type>.md`, and [`issue_format`](https://github.com/pipecat-ai/pipecat/blob/f82b059b3f8637e972f258207f65b657cd284e5b/pyproject.toml#L268-L270) renders that number as a link to `/pull/{issue}`.

As named, the released changelog will link to [#4387](https://github.com/pipecat-ai/pipecat/pull/4387) — an unrelated closed PR ("feat(serializers): add BandwidthFrameSerializer") — instead of this one.

markbackman (commented):
## Code review

Checked for bugs and CLAUDE.md compliance. Three issues below, one of which can take the STT service permanently offline.

The payload shape itself checks out against the Google STT v2 protos — `SpeechAdaptation.phrase_sets`/`custom_classes`, `AdaptationPhraseSet.phrase_set`/`inline_phrase_set`, and `phrase_set_references` is the correct v1 spelling for the compat shortcut.

ddallabenetta:
Thanks @markbackman for the review. I've addressed the points and rebased the branch on the latest main since there were some conflicts

changelog/4413.changed.md
ddallabenetta: Fixed, the changelog file name has been renamed

src/pipecat/services/google/stt.py
ddallabenetta: Fixed using the suggested approach

src/pipecat/services/google/stt.py
ddallabenetta: Fixed. Adaptation is now normalized during initialization and before `super()._update_settings(delta)` commits a runtime update.

markbackman:
@ddallabenetta I opened a PR on your fork with some additional clean to this PR:
https://github.com/teleniasoftware/pipecat/pull/1

I suggest reviewing commit by commit to see the rationale for each change. Let me know if you think any of that doesn't make sense. If you agree, feel free to approve/merge. It's based off of this branch, so it will be added here.

Once we iron that out, this PR will be ready to go. Thanks for collaborating 🙌

markbackman (approved):
LGTM! Thanks for collaborating on this one 🙌

pficardoso:
Thanks for the feature and release :fireworks:

PR #4415, opened 2026-05-04
Fix ElevenLabs alignment chunk spacing for Flash models

### LINKED ISSUE #4391: ElevenLabsTTSService: _strip_leading_space causes incorrect word merging across chunks (e.g. "que quieras" → "quequieras")
### Version

- `pipecat-ai==1.1.0`
- ElevenLabs WebSocket TTS, Flash v2.5 model, Spanish output

### Summary

Adjacent words that should be space-separated are being concatenated in the emitted `TTSTextFrame` stream and end up merged in the assistant transcript that is appended to the LLM context.

Observed example (Spanish):

> "...algo que **que**quieras contarme..."

Expected: `"que quieras"` — Actual: `"quequieras"`.

### Root cause

`_strip_leading_space` in `pipecat/services/elevenlabs/tts.py` removes the first character of every `normalizedAlignment` chunk if it is a space. The docstring justifies this as "marks the prosody/chunk boundary" and notes that leaving it in place "would prematurely terminate a partial word carried over from the previous chunk".

In practice ElevenLabs emits chunks of two distinct shapes, and the strip cannot distinguish between them:

| Case | Chunk N ends | Chunk N+1 starts | Without strip | With strip |
|---|---|---|---|---|
| A — word split mid-word | `"que"` | `" uieras"` | `"queuieras"` ✓ | `"queuieras"` ✓ |
| B — words separated normally | `"...que"` | `" quieras..."` | space terminates `_partial_word` → `"que" + "quieras"` ✓ | strip removes the separator → `_partial_word=\"que\"` absorbs `\"quieras\"` → `\"quequieras\"` ✗ |

The strip fixes case A and regresses case B. Pre-1.1.0 (no strip) handled case B correctly because the leading space terminated the carried-over `_partial_word` in `calculate_word_times`.

### Impact

The bug surfaces in the assistant text appended to the LLM context (via `TTSTextFrame.append_to_context=True` flowing through the universal LLM response aggregator), so subsequent turns see corrupted history, not just mis-rendered subtitles.

### Reproduction

Any Spanish (or other language with frequent short function words) prompt that produces consecutive short words across an ElevenLabs WebSocket chunk boundary will reproduce. Flash v2.5 emits two `normalizedAlignment` messages per sentence, which makes boundary cases more likely.

### Suggested directions

The strip is too eager — it cannot tell a "chunk-marker space" from a real word-separator space using only the leading char. Possible mitigations:

1. Only strip the leading space of the *first* chunk of an utterance (the "leading whitespace stripped before first token" wording in the 1.1.0 release notes suggests this was the intent).
2. Decide per-chunk based on whether the previous chunk's last char was a space or whether `_partial_word` is empty — strip only when continuing a real partial word.
3. Cross-check against the non-normalized `alignment` field (when present) to know whether the boundary is intra- or inter-word.

markbackman:
## Summary

- Fixes word merging (e.g. `bookLook`) in `ElevenLabsTTSService` and `ElevenLabsHttpTTSService` when using Flash models.
- The previous fix unconditionally stripped a leading space from every normalized alignment chunk to avoid a stray space at the start of an utterance. Flash often splits sentences mid-stream into chunks that start with a real inter-word space, so stripping it concatenated adjacent words.
- Now only strip leading spaces on the first alignment chunk of an utterance. Subsequent chunks keep their leading space so `calculate_word_times` flushes any partial word carried over from the previous chunk. Applied to both the WebSocket and HTTP paths.

## Testing

- `uv run pytest tests/test_elevenlabs_tts.py` covers the Flash chunking pattern and verifies only utterance-leading spaces are stripped.

Fixes #4391


filipi87 (approved):
LGTM!

PR #4416, opened 2026-05-04
feat(aws): unify credential resolution across all AWS services

markbackman:
## Summary

AWS services in Pipecat had inconsistent credential handling: Transcribe STT only honored explicit kwargs and `AWS_*` env vars, while Polly TTS, Bedrock LLM, and the Bedrock AgentCore processor relied on aioboto3 falling through to the boto3 chain — but only when *both* access key and secret key were unset. None of them supported IAM roles cleanly, and a half-set environment (`AWS_ACCESS_KEY_ID` without `AWS_SECRET_ACCESS_KEY`) caused hard failures across the aioboto3 services.

This PR adds a shared `resolve_credentials()` helper and routes every AWS service in the module that owns auth through it.

## Motivation

Services running with IAM roles (EKS pod roles, IRSA, ECS task roles, EC2 instance profiles) or SSO previously had to export static credentials to use Transcribe. The aioboto3 services masked the same gap less obviously — they "worked" on instance roles only when the environment was completely empty, and silently broke when an operator had partial AWS env vars set in a shell.

A single resolution chain (`explicit → env → boto3 chain`) used by every service makes credential behavior consistent and matches what users get from the AWS CLI.

### `src/pipecat/services/aws/utils.py`
- Add `AWSCredentials` dataclass for type-safe credential passing.
- Add `resolve_credentials()` with the three-step fallback chain (explicit params → env vars → boto3 provider chain). The chain is only consulted when both access key and secret key are still unresolved, so explicit credentials are never replaced.
- `boto3` is imported lazily and the chain step is wrapped in try/except so services that don't need it don't take a hard dependency.
- Add `AWSCredentials.to_boto_kwargs()` to bridge the dataclass field names to the `aws_access_key_id` / `aws_secret_access_key` / `region_name` kwargs that `aioboto3.Session().client()` expects.

### `src/pipecat/services/aws/stt.py`
- `AWSTranscribeSTTService` now resolves credentials through `resolve_credentials()` instead of manual `os.getenv()` calls.
- Docstrings updated to document the fallback behavior.

### `src/pipecat/services/aws/tts.py`, `llm.py`, `agent_core.py`
- `AWSPollyTTSService`, `AWSBedrockLLMService`, and `AWSAgentCoreProcessor` all swap their `arg or os.getenv(...)` blocks for `resolve_credentials(...).to_boto_kwargs()`.
- Public constructor signatures are unchanged.
- Behavior is identical for fully-explicit and fully-env-var configurations. Partial env vars now correctly fall through to the boto3 chain instead of erroring.

### `tests/test_aws_credentials.py`
- 8 unit tests covering explicit params, env var fallback, boto3 chain fallback, default region, partial-credential safety, and the no-credentials-available case.

## Out of scope

Two services in the AWS module use the smithy-based AWS SDKs (not boto3/aioboto3) and are intentionally **not** modified here:

- `src/pipecat/services/aws/nova_sonic/llm.py` — requires static credentials via `StaticCredentialsResolver`. Wiring it through `resolve_credentials()` is a follow-up because it would change the constructor signature (currently `secret_access_key`/`access_key_id` are required positional kwargs).
- `src/pipecat/services/aws/sagemaker/bidi_client.py` — uses smithy's `EnvironmentCredentialsResolver` which is env-only. Adding chain support is a follow-up that swaps it for `StaticCredentialsResolver` populated from `resolve_credentials()`.

## Testing

```
uv run pytest tests/test_aws_credentials.py
# 8 passed
uv run pyright src/pipecat/services/aws/
# 0 errors
```

Related to #4333.

### DIFF
--- a/src/pipecat/services/aws/stt.py
+++ b/src/pipecat/services/aws/stt.py
@@ -11,7 +11,6 @@ speech-to-text transcription with support for multiple languages and audio forma
 """
 
 import json
-import os
 import random
 import string
 from collections.abc import AsyncGenerator
@@ -29,7 +28,12 @@ from pipecat.frames.frames import (
     StartFrame,
     TranscriptionFrame,
 )
-from pipecat.services.aws.utils import build_event_message, decode_event, get_presigned_url
+from pipecat.services.aws.utils import (
+    build_event_message,
+    decode_event,
+    get_presigned_url,
+    resolve_credentials,
+)
 from pipecat.services.settings import STTSettings, assert_given
 from pipecat.services.stt_latency import AWS_TRANSCRIBE_TTFS_P99
 from pipecat.services.stt_service import WebsocketSTTService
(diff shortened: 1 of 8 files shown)


src/pipecat/services/aws/stt.py
filipi87, line 1: Should we do the same for `aws/tts.py`, `aws/llm.py` and `aws/agent_core.py` ? I believe they are all using aioboto3 if I am not mistaken.

markbackman: Good idea. Let me extend this change.

filipi87 (approved):
LGTM!

PR #4417, opened 2026-05-04
fix: validate download path containment in runner

markbackman:
## Summary

- Fixes a path traversal in the development runner's `/files/{filename:path}` download endpoint, where `%2F`-encoded separators bypassed Starlette's path normalisation and let an unauthenticated request read any file the `pipecat` process could access.
- The endpoint now resolves the joined path and rejects any filename that escapes `--folder` with a 403, and returns 404 (instead of an implicit `null` 200) when `--folder` is unset.
- Adds tests covering literal `../`, `%2F`-decoded traversal, absolute paths, and symlink escape.

## Testing

- `uv run pytest tests/test_runner_downloads.py`
- Manual: start the runner with `--folder /tmp/bot_media`, confirm legitimate files still download and that `curl http://127.0.0.1:7860/files/..%2F..%2Fetc%2Fpasswd` returns 403.


filipi87 (approved):
Nice improvement! 🚀

PR #4423, opened 2026-05-05
feat: add Inception LLM service with Mercury 2 support

joycech333:
Adds InceptionLLMService, an OpenAI-compatible service for Inception's [Mercury 2](https://www.inceptionlabs.ai/blog/introducing-mercury-2) diffusion reasoning model. Supports reasoning_effort (instant/low/medium/high) and realtime mode for reduced TTFT.


markbackman (approved):
Thanks for the addition! I pushed a few changes to align to Pipecat norms and cleaned up the example a bit. Since speed is one differentiator for your model, I switched reasoning to `instant`; this makes the model a great fit for realtime use cases.

PR #4424, opened 2026-05-05
fix(elevenlabs): only use normalizedAlignment when pronunciation dict is set

### LINKED ISSUE #4316: ElevenLabs TTS: duplicate alignment chunks cause garbled assistant transcript
## Bug Description

When using `ElevenLabsTTSService` (WebSocket, Flash v2.5) with the `multi-stream-input` endpoint, the assistant transcript in `LLMAssistantAggregator` contains duplicated and garbled text. Each sentence in the LLM response appears twice — once truncated mid-word, once more complete — producing output like:

```
"Hola Antoine, soy MHola Antoine, soy Maria, la talent manager vir Te llamo para continTe llamo para continuar con tu candidatura..."
```

**This affects all versions** (confirmed on v0.0.102 and v1.0.0). The bug is more severe in v1.0.0 due to a `run_tts` reset regression (see below), but the root cause exists in all versions.

## Root Cause

ElevenLabs sends **two alignment messages per sentence** over the WebSocket: first a partial alignment (covering part of the sentence), then a second alignment restarting from character 0.

`calculate_word_times()` in `elevenlabs/tts.py` carries `_partial_word` state across alignment chunks, assuming they are **sequential segments** of a single pass. When the second alignment restarts from character 0, the leftover partial from the end of pass 1 merges with the beginning of pass 2:

```
Pass 1 alignment: "Buenos días ¿hablo co..."
  → words: "Buenos", "días", "¿hablo"
  → partial_word = "co"  (tail of "con", cut mid-word)

Pass 2 alignment: "Buenos días ¿hablo con Antoine?"  (restarts from beginning)
  → partial_word "co" + "Buenos" → word "coBuenos" emitted!
  → then "día", etc. — entire sentence emitted again
```

This produces doubled word-aligned `TTSTextFrame`s, all with `append_to_context: True`, which the `LLMAssistantAggregator` faithfully appends — resulting in garbled assistant messages in the LLM context.

## Pronunciation dictionaries amplify the issue

The duplicate alignment behavior is **more visible when pronunciation dictionaries are configured**. When `pronunciation_dictionary_locators` are passed in the WebSocket message, ElevenLabs applies text normalization that changes the surface form of words. The two alignment messages then contain noticeably different text:

- **Pass 1** (`alignment`): alignment for the original text — e.g., `"Buenos días\n ¿hablo co"` (23 chars)
- **Pass 2**: alignment for the normalized text — e.g., `"Buenos día "` (11 chars, shorter, different word forms)

Without pronunciation dictionaries the two passes are more similar, so the duplication is less obvious in the final output.

## Debug Evidence

Frame-level trace showing the duplicated word stream (word timestamps from `_add_word_timestamps`):

```
# Pass 1 — partial alignment
adding word timestamp: 'Buenos'    at 0.00s
adding word timestamp: 'días\n'    at 0.34s
adding word timestamp: '¿hablo'    at 0.56s
# partial_word = "co"

# Pass 2 — full alignment, restarts from beginning
adding word timestamp: 'coBuenos'  at 0.81s   ← GARBLED: partial "co" + "Buenos"
adding word timestamp: 'día'       at 1.30s    ← duplicate
```

Raw alignment messages from ElevenLabs WebSocket:
```
RAW alignment chunk: text='Buenos días\n ¿hablo co' chars_len=23 first_time_ms=0 last_time_ms=848 partial_word='' cumulative_time=0.000
RAW alignment chunk: text='Buenos día '            chars_len=11 first_time_ms=0 last_time_ms=534 partial_word='co' cumulative_time=0.929
```

## Impact

- Every assistant message in `LLMContext` is garbled (doubled sentences with merged partial words)
- LLM receives corrupted conversation history, degrading response quality
- Affects all non-trivial responses (multiple sentences)
- Worse with pronunciation dictionaries (normalized vs original text differ more)

## Environment

- pipecat-ai 1.0.0 (also confirmed on 0.0.102)
- ElevenLabs Flash v2.5 (WebSocket `multi-stream-input` endpoint)
- μ-law 8kHz telephony output
- Spanish language (`es-es`)
- Observed with `aggregate_sentences=True` (default)

markbackman:
## Summary

- PR #4344 unconditionally switched both ElevenLabs TTS services to `normalizedAlignment` to fix #4316, but `normalizedAlignment` returns the *post-normalized* form of what was spoken. For non-Latin input, that means romanization: Chinese ends up in the assistant's LLM context as pinyin (`Ni Hao ! Wo Shi ...`), which the LLM then anchors on for subsequent turns.
- Gate the switch on `pronunciation_dictionary_locators` being configured. Without a dictionary the original `alignment` field is fine (the issue #4316 garbling only manifests with dictionaries). With a dictionary, keep using `normalizedAlignment` so #4316 stays fixed.
- Extract a `_select_alignment(...)` helper used by both WebSocket and HTTP services. Reads with preferred-with-fallback semantics since both fields are nullable per the API schema.

## Testing

- `uv run pytest tests/test_elevenlabs_tts.py -v` covers default mode, dictionary mode, fallback when preferred is missing/null, both-null, and the HTTP key variant.
- Manual: `examples/voice/voice-elevenlabs.py` in Chinese (no dictionary) -> confirm assistant context shows Chinese characters, not pinyin.
- Manual: same example with a pronunciation dictionary configured -> confirm word timestamps still match spoken audio.


src/pipecat/services/elevenlabs/tts.py
```diff
@@ -248,17 +248,57 @@ class ElevenLabsHttpTTSSettings(TTSSettings):
...
 ) -> Mapping[str, Any]:
     """Return alignment with utterance-leading space chars removed, if requested.
 
-    Normalized alignment chunks from ElevenLabs often begin with a space. On the
-    first chunk of an utterance, that space is leading whitespace and should not
-    become a text token. On subsequent chunks, however, a leading space can be a
-    real inter-word separator (Flash models commonly split sentences this way),
-    so it must be preserved for ``calculate_word_times`` to flush any partial
-    word carried over from the previous chunk.
+    ElevenLabs Flash normalized alignment chunks can begin with a leading space
```
markbackman, line 294: Improving the docstring for clarity.

filipi87 (approved):
I will trust you on this one. 😅

LGTM! 🚀

markbackman:
Yeah, this is not ideal. I originally made the change to `normalizedAlignment` because of the pronunciation dictionaries, but found out that some languages appear as the phonetical spelling. For example Chinese characters are returned as [pinyin](https://en.wikipedia.org/wiki/Pinyin).

So, instead, we're reverting back to the `alignment` data by default, only using the `normalizedAlignment` when pronunciation dictionaries are used. I think this is the best we can do provided what ElevenLabs' API offers.

PR #4425, opened 2026-05-06
feat: support new async-tool mechanism in AWS Nova Sonic

kompfner:
## Summary

Add support to `AWSNovaSonicLLMService` for the new "async tool call" mechanism activated by `cancel_on_interruption=False`, which includes:

- delivering results asynchronously
- delivering result streams
- cancelling running async tools

Note that the introduction of the new mechanism had actually caused a regression in AWS Nova Sonic, which previously supported `cancel_on_interruption=False` with the old mechanism (simply avoiding discarding tool calls on interruptions).

Support for the other major realtime services (`GeminiLiveLLMService`, `OpenAIRealtimeLLMService`) will follow in a separate PR, or separate PRs — Gemini Live, for instance, definitely needs more work before it can support long-running tool calls reliably. This PR also lays the groundwork for those follow-ups: the new provider-agnostic helper module `pipecat.processors.aggregators.async_tool_messages` (with unit tests) and the placeholder-as-formal-tool-result + intermediate/final-as-text-injection pattern established here are intended to be reused by the Gemini Live and OpenAI Realtime implementations.

(Replaces #4402, which was inadvertently closed by a branch rename via the GitHub REST API.)

## Test plan

- [x] Helper unit tests pass (`uv run pytest tests/test_async_tool_messages.py`)
- [x] Existing test suite still passes
- [x] AWS Nova Sonic — confirmed working at the "full" tier (single-shot async + streaming async)

### DIFF
--- a/changelog/4425.added.md
+++ b/changelog/4425.added.md
@@ -1,0 +1,1 @@
+- Added support to `AWSNovaSonicLLMService` for the new "async tool call" mechanism activated by `cancel_on_interruption=False`, which includes delivering results asynchronously, delivering result streams, and cancelling running async tools. Support for the other major realtime services (`GeminiLiveLLMService`, `OpenAIRealtimeLLMService`) will be added in a follow-up PR.
--- a/examples/function-calling/function-calling-openai-async-stream.py
+++ b/examples/function-calling/function-calling-openai-async-stream.py
@@ -74,6 +74,7 @@ async def track_current_location(params: FunctionCallParams):
 
     # Second update: revised city estimate.
     await asyncio.sleep(10)
+    # await asyncio.sleep(20)
     gps = {"lat": 33.96003, "lng": -118.40639}
     await params.result_callback(
         {"gps": gps, "city": "Los Angeles"},
@@ -82,6 +83,7 @@ async def track_current_location(params: FunctionCallParams):
 
     # Final result: confirmed city.
     await asyncio.sleep(10)
+    # await asyncio.sleep(20)
     gps = {"lat": 32.743569, "lng": -117.20466}
     await params.result_callback({"gps": gps, "city": "San Diego"})
 
--- a/src/pipecat/processors/aggregators/async_tool_messages.py
+++ b/src/pipecat/processors/aggregators/async_tool_messages.py
@@ -289,95 +289,82 @@ def parse_message(message: LLMStandardMessage) -> AsyncToolMessagePayload | None
 # --- Realtime preparation ----------------------------------------------------
 
 
-# Natural-language reminder grafted onto the ``description`` field of in-flight
-# payloads (started / intermediate) when they're sent to a realtime LLM
-# service. Realtime services receive these mid-stream while the model is
-# still talking with the user, which is the moment the model is most likely
-# to mistakenly re-issue the same tool call. Keeping this reminder out of the
-# canonical payload descriptions (and confined to the realtime path) avoids
-# influencing non-realtime consumers of the same context. We don't graft it
-# onto ``final`` payloads, because at that point the task is done and
-# re-invocation by the model is no longer a mistake.
-#
-# The reminder is appended *after* the canonical description so the model
-# first reads the protocol-level explanation of what async-tool messages are
-# and how they work, and only then encounters the behavioral directive,
-# which now flows naturally from that context.
-_REALTIME_REINVOCATION_REMINDER = (
-    "While this task is in flight, do not call the same tool with the same "
-    "arguments again; you would just kick off a duplicate task."
-)
-
-
-def prepare_message_payload_for_realtime(payload: AsyncToolMessagePayload) -> str:
+def prepare_message_payload_for_realtime(
+    payload: AsyncToolMessagePayload,
+    *,
+    template: str | None = None,
+) -> str:
     """Prepare an async-tool message payload for sending to a realtime LLM service.
 
-    Returns a wire-ready JSON string. Realtime services that fully honor the
-    async-tool mechanism send the ``started`` payload via the formal
-    tool-result channel and the subsequent ``intermediate`` / ``final``
-    payloads as text injected mid-conversation; this function returns the
-    string to send in either case, and callers route it to the appropriate
-    channel.
-
-    The exact transformation depends on the payload kind. Each kind is
-    handled by its own private helper, so per-kind tweaks can be added later
-    without entangling the others. Today:
-
-    - ``started`` / ``intermediate``: a natural-language reminder
-      discouraging the model from re-invoking the in-flight tool is grafted
-      onto the ``description`` field, then the payload is re-serialized.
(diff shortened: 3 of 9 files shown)


examples/realtime/realtime-aws-nova-sonic.py
```diff
@@ -46,11 +45,6 @@ async def fetch_weather_from_api(params: FunctionCallParams):
         if params.arguments["format"] == "fahrenheit"
         else random.randint(15, 30)
     )
-    # Simulate a long network delay.
```
kompfner, line 49: Now that we have a dedicated async tool example, this one is updated to demonstrate the "base" case.

kompfner:
Going ahead and opening this for review, though there's one remaining task, which is testing async tool cancellation (`enable_async_tool_cancellation=True`).

src/pipecat/services/aws/nova_sonic/llm.py
```diff
@@ -235,6 +237,26 @@ class AWSNovaSonicLLMSettings(LLMSettings):
     endpointing_sensitivity: str | None | _NotGiven = field(default_factory=lambda: NOT_GIVEN)
 
 
+# Bracketed plain-text template Nova Sonic uses when injecting async-tool
+# result updates onto the cross-modal user-text channel.
+#
+# Note that this template intentionally drops the payload's ``description``
```
kompfner, line 243: The findings in this comment were mind-numbing to discover. I think I spent hours fiddling with the message format to coax the model to mostly behave how we want it to.

kompfner:
Here's the major design decision worth documenting (though @filipi87 might already be aware, as it was in a discussion with him that we decided on this decision).

Broadly, there were two overall approaches we could've taken to support async tool calling in our realtime services:

1. Continue letting `cancel_on_interruption=False` work as it does today: by adding a "started" tool call result right away to the context, then sending "result" updates via subsequent developer-role context messages. Add the "translation" at the service level as it reads context updates.

2. Somehow override how `cancel_on_interruption=False` works on a service-by-service basis, so that we can take advantage of each realtime service's (AWS Nova Sonic, Gemini Live, OpenAI Realtime, etc) _native_ support for async function calling.

As @filipi87 is aware, we decided to go with Option 1 for now, for a couple of reasons:
- (The big one) So we can aim for uniformity in behavior: the centralized `cancel_on_interruption=False` mechanism means every service should get the same features instead of a service-by-service patchwork:
  - async tool calls
  - async streamed results
  - async tool cancellation
- To keep aggregators agnostic about which LLM services they're sharing a pipeline with (though it _might_ be possible for this to be achievable with option 2...not sure yet)

The huge downside, of course, is that we're *not* fully taking advantage of native async tool calling mechanisms in each service, and all the testing and effort and updates that provider pour into those.

This was not an easy choice to make. And, honestly, I'm still open to debate this choice.

filipi87:
> The huge downside, of course, is that we're not fully taking advantage of native async tool calling mechanisms in each service, and all the testing and effort and updates that provider pour into those.

For those real-time services, we usually need to provide the function call result through some specific message that we send back to the service, right ?

So, if I remember correctly, the difference is that we now need to retrieve the function call result from the developer message in the context, instead of from the specific tool result message.

But real-time services usually don’t rely on that context anyway. For them, what really matters is the result we send back through that specific message mechanism, isn’t it ?

So I’m not entirely sure how changing this would let us take greater advantage of native async tool-calling mechanisms. But maybe I am still missing something. 🤕

changelog/4425.added.md
filipi87, line 1: Do we need to include this last part: "Support for the other major realtime services (`GeminiLiveLLMService`, `OpenAIRealtimeLLMService`) will be added in a follow-up PR." ?

examples/function-calling/function-calling-openai-async-stream.py
filipi87, line 1: nit: I think we can probably remove the changes in this file.

examples/realtime/realtime-aws-nova-sonic.py
filipi87: Yep. Makes sense. 👍

src/pipecat/processors/aggregators/async_tool_messages.py
```diff
@@ -0,0 +1,370 @@
...
+    *,
+    template: str | None = None,
+) -> str:
+    """Prepare an async-tool message payload for sending to a realtime LLM service.
+
+    Realtime services that fully honor the async-tool mechanism send the
+    ``started`` payload via the formal tool-result channel and the subsequent
+    ``intermediate`` / ``final`` payloads as text injected mid-conversation;
+    this function returns the string to send in either case, and callers
+    route it to the appropriate channel.
```
filipi87, line 303: @kompfner, I believe this is the difference. I thought that for the realtime services, we would not support async streaming, like sending multiple messages, and instead we would only send the result using the tool-result message. 

This way, we would be taking advantage of the async mechanism that these realtime services already support natively.

Of course, the downside would be not supporting async streaming for those cases. But I guess the tradeoff would be worth it.

What do you think ?

src/pipecat/processors/aggregators/async_tool_messages.py
filipi87, line 1: Nice abstraction. 🎉

src/pipecat/services/aws/nova_sonic/llm.py
```diff
@@ -620,6 +643,13 @@ async def _process_completed_function_calls(self, send_new_results: bool):
             # standard tool-result messages — skip them.
             if isinstance(message, LLMSpecificMessage):
                 continue
+
+            async_info = async_tool_messages.parse_message(message)
+            if async_info is not None:
+                # Async-tool message — dispatch per the configured support tier.
+                await self._dispatch_async_tool_message(async_info, send_new_results)
+                continue
```
filipi87, line 651: I thought that for the real-time services, we would not support async streaming, and instead we would only send the result using the `_send_tool_result` message.

So basically, most of the changes in this file would not be needed. 

We would only retrieve the result from a different place in the context. And this way, we would keep using the native mechanism provided by the realtime services for async tool calls.

What do you think?

kompfner: Ah! You know, that decision sounds familiar now that you're describing it! Somehow I had misremembered our decision, or my thinking had "drifted" away from it.

So, per your recollection, we had decided:

1. To just not bother with supporting streamed async tool results with realtime services.
2. To leverage each realtime providers' native async tool calling capabilities.
3. To achieve that by just ignoring the initial "started" result and then reporting the eventual result message as the formal result.

This actually makes a lot of sense to me...it was feeling off to me to be "fighting" the realtime providers' native capabilities. Better to just support a bit less, and make it clear through logging and documentation what we don't support.

filipi87: Yep, this is exactly what I thought we had decided.

kompfner:
Closing in favor of (still-draft) PR #4441

PR #4428, opened 2026-05-06
chore(audio): deprecate ResampyResampler

markbackman:
## Summary

- Deprecates `ResampyResampler` in favor of the SOXR resamplers (`SOXRAudioResampler`, `create_file_resampler()`, `create_stream_resampler()`)
- Instantiating `ResampyResampler` now emits a `DeprecationWarning`; the class will be removed in Pipecat 2.0 along with the default `resampy` and `numba` dependencies
- Adds a unit test verifying the warning is emitted

## Testing

- `uv run pytest tests/test_resampy_resampler.py`


markbackman:
Once removed, we can drop meaningful dependencies: `resampy` and `numba`, which will allow us to drop `llvmlite`, which is a large import.

We can't remove those dependencies now since that would be breaking :/

filipi87 (approved):
LGTM!

PR #4430, opened 2026-05-06
Implementing dynamic watchdog timeout for Deepgram Flux STT

filipi87:
## Summary                                                                                                                                   
                                  
- Changed the Deepgram Flux STT watchdog silence threshold from a fixed 500 ms to a dynamic `max(chunk_duration * 2, watchdog_min_timeout)`. This prevents false silence injections when large audio chunks are sent at a lower frequency.      
- Added `watchdog_min_timeout` parameter (default `0.5` s) to `DeepgramFluxSTT` and `DeepgramFluxSageMakerSTT` to allow overriding the minimum threshold when needed.

### DIFF
--- a/src/pipecat/services/deepgram/flux/sagemaker/stt.py
+++ b/src/pipecat/services/deepgram/flux/sagemaker/stt.py
@@ -245,6 +245,7 @@ class DeepgramFluxSageMakerSTTService(DeepgramFluxSTTBase):
         if self._client and self._client.is_active:
             try:
                 self._last_stt_time = time.monotonic()
+                self._last_audio_chunk_duration = len(audio) / (self.sample_rate * 2)
                 await self._client.send_audio_chunk(audio)
             except Exception as e:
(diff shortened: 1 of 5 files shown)


src/pipecat/services/deepgram/flux/sagemaker/stt.py
```diff
@@ -105,6 +106,8 @@ def __init__(
             tag: Tags to label requests for identification during usage reporting.
             should_interrupt: Whether to interrupt the bot when Flux detects that
                 the user is speaking. Defaults to True.
+            watchdog_min_timeout: Minimum silence duration in seconds before the watchdog
```
markbackman, line 109: Maybe use wording like "minimum idle timeout before sending silence"

markbackman (approved):
Change looks good. Just a small nit on the docstring descriptions.

src/pipecat/services/deepgram/flux/sagemaker/stt.py
filipi87: Done.

Issue #4418, opened 2026-05-05
TTS with `pause_frame_processing=True` permanently deadlocks when user barges in mid-TTFB while a `FunctionCallResultFrame` is queued


tfvjr:
### pipecat version

1.0.0

### Python version

3.13

### Operating System

Ubuntu 24.04

### Issue description

When an LLM produces text + a tool call in the same response, and the user barges in within the TTFB window of the TTS filler, the affected TTS service (any service that sets `pause_frame_processing=True` -- Rime, ElevenLabs, Cartesia, etc.) goes permanently silent for the rest of the call. Every subsequent `run_tts` invocation never fires; subsequent `LLMTextFrame`s pile up in `__process_queue` but the process task is parked forever on `__process_event.wait()`.

### Reproduction steps

1. Configure a pipeline with `RimeTTSService`, `AnthropicLLMService`, `enable_async_tool_cancellation=True`, and a tool that takes 200-500ms to complete (e.g. a Supabase round-trip). Disable `user_mute_strategies` so the user can speak during tool calls.
2. The LLM emits a short filler text plus a tool call in one response (e.g. `"One sec."` + `collect_customer_info`).
3. User barges in within the TTFB window of the filler -- around 200-700ms after `tts_started`, before any audio chunks have come back from Rime.
4. Tool result fires roughly at the same time, which pushes a `FunctionCallResultFrame` (an `UninterruptibleFrame`) into the TTS process queue.
5. From this point onward, the TTS service no longer logs `Generating TTS [...]` for any subsequent text. The bot goes silent for the rest of the call.

### Expected behavior

After the interruption, subsequent LLM turns should reach `run_tts` normally and produce audio. The pause flag set by `_maybe_pause_frame_processing` after the previous `LLMFullResponseEndFrame` should be cleared so the process task can drain the queue.

### Actual behavior

The process task stays parked on `__process_event.wait()` indefinitely. `_start_interruption` takes the `has_uninterruptible -> __reset_process_queue()` branch (because `__process_current_frame` is the queued `FunctionCallResultFrame`), which doesn't cancel/recreate the task and doesn't set the event. `BotStoppedSpeakingFrame` never fires because no audio reached the output transport, so `_maybe_resume_frame_processing` is never called either.

### Logs

```shell
{"text": "Yeah.", "event": "transcription_final", "timestamp": "2026-05-04T21:39:13.611017+00:00"}
[...]

PR #4431, opened 2026-05-06
Fixing TTSService deadlock.

filipi87:
## Summary                      
                                                                                                                                               
- Fixed a deadlock in `TTSService` that permanently stalled pipeline processing in services using `pause_frame_processing=True` (ElevenLabs, Rime, AsyncAI, Gradium, ResembleAI).                                                                                               
- The deadlock required three simultaneous conditions: an interruption arrived before any TTS audio was played, an `UninterruptibleFrame` (e.g. `TTSUpdateSettingsFrame`, `FunctionCallResultFrame`) was in the processing queue, and `pause_frame_processing=True`.                   

## Testing
                                                                                                                                               
Two regression tests added to `tests/test_tts_frame_ordering.py`:
                                                                                                                                               
- `test_no_deadlock_on_interrupt_before_audio_simple` — verifies the common case (no `UninterruptibleFrame`) continues to work via  `_start_interruption()`.                                                                                                                     
- `test_no_deadlock_on_interrupt_before_audio_with_uninterruptible` — reproduces the exact deadlock condition (times out without the fix, passes with it).


filipi87:
Fixes https://github.com/pipecat-ai/pipecat/issues/4418#event-25157836688

markbackman (approved):
LGTM! Nice catch.

PR #4433, opened 2026-05-06
Refactoring ElevenLabs to send close_context as soon as the turn context is complete.

filipi87:
## Summary
                                                                                                                                               
- Refactored `ElevenLabsTTSService` to send `close_context` earlier — at the end of the turn (via `on_turn_context_completed`) rather than waiting for all audio to finish playing back                                                                                                 
- `isFinal` message from ElevenLabs now signals `TTSStoppedFrame` and cleans up the audio context, removing the need for Pipecat to manage that explicitly                                                                                                                              
- Disabled `push_stop_frames` in the parent class since stop signaling is now driven by the `isFinal` message from the server

### DIFF
--- a/src/pipecat/services/elevenlabs/tts.py
+++ b/src/pipecat/services/elevenlabs/tts.py
@@ -558,7 +558,7 @@ class ElevenLabsTTSService(WebsocketTTSService):
             text_aggregation_mode=text_aggregation_mode,
             aggregate_sentences=aggregate_sentences,
             push_text_frames=False,
-            push_stop_frames=True,
+            push_stop_frames=False,
             pause_frame_processing=True,
             sample_rate=sample_rate,
             settings=default_settings,
@@ -825,15 +825,15 @@ class ElevenLabsTTSService(WebsocketTTSService):
         await self._close_context(context_id)
         await super().on_audio_context_interrupted(context_id)
 
-    async def on_audio_context_completed(self, context_id: str):
-        """Close the ElevenLabs context after all audio has been played.
+    async def on_turn_context_completed(self):
+        """Close the server-side context at end of turn.
 
-        ElevenLabs does not send a server-side signal when a context is
-        exhausted, so Pipecat must explicitly close it with
-        ``close_context: True`` to free server-side resources.
+        Sends close_context so isFinal arrives immediately after the last audio byte.
         """
-        await self._close_context(context_id)
-        await super().on_audio_context_completed(context_id)
+        context_id = self._turn_context_id
+        await super().on_turn_context_completed()
+        if context_id:
+            await self._close_context(context_id)
 
     async def _receive_messages(self):
         """Handle incoming WebSocket messages from ElevenLabs."""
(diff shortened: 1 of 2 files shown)


src/pipecat/services/elevenlabs/tts.py
```diff
@@ -843,25 +843,15 @@ async def _receive_messages(self):
...
             if msg.get("isFinal") is True:
-                logger.trace(f"Received final message for context {received_ctx_id}")
-                continue
-
-            # Check if this message belongs to the current context.
-            if not self.audio_context_available(received_ctx_id):
-                if self.get_active_audio_context_id() == received_ctx_id:
-                    logger.debug(
-                        f"Received a delayed message, recreating the context: {received_ctx_id}"
+                logger.debug(f"Received final message for context {received_ctx_id}")
```
markbackman, line 856: Maybe this should remain as trace?

markbackman (commented):
LGTM. I have some questions about the sequencing of things, so I'm asking Claude for a quick review.

src/pipecat/services/elevenlabs/tts.py
```diff
@@ -825,15 +825,15 @@ async def on_audio_context_interrupted(self, context_id: str):
...
-        exhausted, so Pipecat must explicitly close it with
-        ``close_context: True`` to free server-side resources.
+        Sends close_context so isFinal arrives immediately after the last audio byte.
         """
-        await self._close_context(context_id)
-        await super().on_audio_context_completed(context_id)
+        context_id = self._turn_context_id
+        await super().on_turn_context_completed()
+        if context_id:
+            await self._close_context(context_id)
```
markbackman, line 845: **`_close_context` resets per-turn timing state mid-turn**

This call moves the close from end-of-playback to end-of-LLM-turn ([new `on_turn_context_completed`](https://github.com/pipecat-ai/pipecat/blob/fda18a9afa9bb305270b927c4cc495af25a94da2/src/pipecat/services/elevenlabs/tts.py#L827-L837)), but [`_close_context`](https://github.com/pipecat-ai/pipecat/blob/fda18a9afa9bb305270b927c4cc495af25a94da2/src/pipecat/services/elevenlabs/tts.py#L800-L822) does two things at once: it sends `close_context` to the server **and** unconditionally resets `_cumulative_time`, `_partial_word`, `_partial_word_start_time`, and discards `context_id` from `_alignment_started_context_ids`.

The PR description and the new `isFinal` handler both confirm that audio + alignment chunks keep arriving for that same context until `isFinal` (that's the point — overlap the tail with the next turn). The [receive loop](https://github.com/pipecat-ai/pipecat/blob/fda18a9afa9bb305270b927c4cc495af25a94da2/src/pipecat/services/elevenlabs/tts.py#L860-L901) reads and writes those exact fields when processing trailing alignment, so on every normal turn:

1. **Wrong word PTS for trailing chunks.** `calculate_word_times` is called with `self._cumulative_time` already reset to `0`, so word timestamps for the back half of the response collapse to chunk-relative offsets and feed bogus `pts` into `add_word_timestamps`, `_word_last_pts`, `TTSStoppedFrame.pts`, and `LLMFullResponseEndFrame.pts`.
2. **Mangled inter-word spacing.** Because the `context_id` was discarded from `_alignment_started_context_ids`, the next trailing chunk re-enters the "utterance-leading" branch of [`_strip_utterance_leading_spaces`](https://github.com/pipecat-ai/pipecat/blob/fda18a9afa9bb305270b927c4cc495af25a94da2/src/pipecat/services/elevenlabs/tts.py#L289-L321) and strips a leading space that Flash uses as a real inter-word separator on subsequent chunks, merging adjacent words.

Pre-PR this was safe because `_close_context` only ran from `on_audio_context_completed`, after the audio queue had drained.

**Suggested fix:** split `_close_context` so the network send and the local state reset are separable.
- `on_turn_context_completed` should only send the `close_context` WS message.
- The state reset (`_cumulative_time`, `_partial_word*`, `_alignment_started_context_ids.discard(context_id)`) should run on end-of-context — either in the `isFinal` branch of `_receive_messages` or in `on_audio_context_completed` (still wired up via the parent class).
- `on_audio_context_interrupted` can keep the immediate full reset since the dead context should be forgotten right away.

markbackman (commented):
## Code review

One issue flagged. Checked for bugs and CLAUDE.md compliance.

src/pipecat/services/elevenlabs/tts.py
filipi87: Yeah, I was kind of unsure about this one. 

I thought about leaving it as `debug` for now to make things easier to identify in case we find any issues, just to be sure that we are receiving this event from ElevenLabs. 

But in all my tests, we’ve always received it, so maybe it’s okay to change it back to trace.

src/pipecat/services/elevenlabs/tts.py
filipi87: Nice catch!

filipi87: Fixed!

src/pipecat/services/elevenlabs/tts.py
markbackman: It will get logged on every turn, so it's pretty noisy and is mainly for development. Maybe trace is good?

markbackman (approved):
LGTM!

PR #4434, opened 2026-05-06
Fix interruption blocked by slow non-uninterruptible frame in queue

### LINKED ISSUE #4412: Potential regression in the way we handle Interruption in FrameProcessor
### pipecat version

1.1.0

### Python version

3.13.2

### Operating System

MacOS

### Issue description

Hello @filipi87 / @aconchillo 

[In this commit](https://github.com/pipecat-ai/pipecat/pull/4217/changes#diff-69b82d1a260a49ffa45e16d103bc437a24949b15b02a457d3b86c92af38e2a5c) we made a change where we don't cancel process task if there is an `UninterruptibleFrame` **in the queue**. 

I believe it will cause an issue in the below scenario

1. A processor is stuck processing an interruptible frame for instance an LLMRunFrame (and the provider has higher than usual latency)
2. There is an uninterruptible frame in the queue, it can be either `LLMContextSummaryResultFrame` or any other `UninterruptibleFrame`. 
3. The user notices the silence and starts speaking again causing the pipeline to broadcast an `InterruptionFrame`

The earlier behaviour was that we would cancel process task IF the frame being processed is interruptible allowing the interruption frame to be propagated down the pipeline. 

IMO, this is a regression. The interruption frame must cancel process task if its not processing an UninterruptibleFrame. It should also preserve the `UninterruptibleFrame` which was existing behaviour. 

### Reproduction steps

```
"""Standalone reproduction for pipecat PR #4217 interruption regression.

Scenario:
    1. A processor is slow on an interruptible frame (LLMRunFrame).
    2. An UninterruptibleFrame (LLMContextSummaryResultFrame) is queued behind it.
    3. An InterruptionFrame is broadcast.

Expected (pre-#4217): the in-flight LLMRunFrame is cancelled because it is
interruptible. The queued UninterruptibleFrame is preserved.

Actual (post-#4217): _start_interruption() takes the
`has_uninterruptible -> reset_process_queue()` branch and never cancels the
process task, so the slow LLMRunFrame runs to completion.
"""

import asyncio
import sys
import time

from loguru import logger

    EndFrame,
    Frame,
    InterruptionFrame,
    LLMContextSummaryResultFrame,
    LLMRunFrame,
)
from pipecat.pipeline.runner import PipelineRunner
from pipecat.pipeline.task import PipelineTask

logger.remove()
logger.add(sys.stderr, level="TRACE")

LATENCY = 5.0


class SlowLLM(FrameProcessor):
    """Mimics an LLM service stuck on a high-latency provider response."""

    def __init__(self):
        super().__init__()
        self.run_finished_naturally = False

        if isinstance(frame, LLMRunFrame):
            await asyncio.sleep(LATENCY)
            # Reached only if the process task was NOT cancelled.
            self.run_finished_naturally = True


async def main() -> int:
    proc = SlowLLM()
    pipeline = Pipeline([proc])
    task = PipelineTask(pipeline, cancel_on_idle_timeout=False)

    summary = LLMContextSummaryResultFrame(
        request_id="x",
        summary="...",
        last_summarized_index=0,
    )

    async def driver():
        await asyncio.sleep(0.05)                 # let runner start
        await task.queue_frame(LLMRunFrame())     # slow, interruptible
        await asyncio.sleep(0.2)                  # let LLMRunFrame begin
        await task.queue_frame(summary)           # uninterruptible, sits in queue
        await asyncio.sleep(0.2)                  # Let summary reach SlowLLM processor
        await task.queue_frame(InterruptionFrame())  # should cancel LLMRunFrame
        await asyncio.sleep(0.5)                  # window for cancellation
        await task.queue_frame(EndFrame())

    runner = PipelineRunner()
    started = time.monotonic()
    await asyncio.gather(runner.run(task), driver())
    elapsed = time.monotonic() - started

    print()
    print(f"elapsed: {elapsed:.2f}s  (LATENCY={LATENCY}s)")
    print(f"LLMRunFrame ran to completion: {proc.run_finished_naturally}")

    if proc.run_finished_naturally or elapsed >= LATENCY:
        print(
            "[FAIL] regression reproduced — InterruptionFrame did NOT cancel "
            "the slow LLMRunFrame because an UninterruptibleFrame was queued."
        )
        return 1

    print("[PASS] InterruptionFrame cancelled the slow LLMRunFrame.")
    return 0


if __name__ == "__main__":
    raise SystemExit(asyncio.run(main()))
```

### Expected behavior

I have give the expected behaviour and actual behaviour in repro code

### Actual behavior

I have give the expected behaviour and actual behaviour in repro code

### Logs

```shell

```

filipi87:
When a non-uninterruptible frame was being processed slowly and an uninterruptible frame was waiting in the queue, _start_interruption skipped task cancellation. This caused interruptions to stall until the slow frame finished, even though it had no reason to block them.

Fixes: https://github.com/pipecat-ai/pipecat/issues/4412

### DIFF
--- a/changelog/4434.fixed.md
+++ b/changelog/4434.fixed.md
@@ -1,0 +1,1 @@
+- Fixed interruptions being delayed when a slow non-uninterruptible frame was processing and an uninterruptible frame was waiting in the queue. The bot would stall until the slow frame finished instead of cancelling it immediately on interruption.
--- a/src/pipecat/processors/frame_processor.py
+++ b/src/pipecat/processors/frame_processor.py
@@ -877,14 +877,19 @@ class FrameProcessor(BaseObject):
             current_is_uninterruptible = isinstance(
                 self.__process_current_frame, UninterruptibleFrame
             )
-            if current_is_uninterruptible or self.__process_queue.has_uninterruptible:
-                # We don't want to cancel an UninterruptibleFrame (either the
-                # one currently being processed or one waiting in the queue),
-                # so we simply cleanup the queue keeping only
-                # UninterruptibleFrames.
+            if current_is_uninterruptible:
+                # The frame currently being processed is uninterruptible, so we
+                # must not cancel it. Just flush non-uninterruptible frames from
+                # the queue; any uninterruptible ones will be kept and processed
+                # after the current frame finishes.
                 self.__reset_process_queue()
             else:
-                # Cancel and re-create the process task.
+                # Cancel and re-create the process task. Previously this branch
+                # was skipped when the queue contained an uninterruptible frame,
+                # which caused slow non-uninterruptible frames to block
+                # interruptions. Uninterruptible queued frames are safe here
+                # because __create_process_task calls __reset_process_queue
+                # internally, which always preserves them.
                 await self.__cancel_process_task()
                 self.__create_process_task()
         except Exception as e:


markbackman (approved):
LGTM

PR #4441, opened 2026-05-07
fix: restore cancel_on_interruption=False support in AWS Nova Sonic and OpenAI Realtime

kompfner:
## Summary

Restores `cancel_on_interruption=False` support for `AWSNovaSonicLLMService` and `OpenAIRealtimeLLMService`.

These services previously honored the flag by simply not cancelling in-flight function calls on interruption — the eventual result then flowed through the same channel as any synchronous tool result. The new async-tool mechanism (which threads started/intermediate/final messages through the LLM context as the underlying task progresses) broke that path: the realtime services didn't know how to interpret those messages, and the eventual result was never delivered to the provider.

This PR teaches both services to detect async-tool messages in the context and route them appropriately:

- `started` → skipped silently (the provider already issued the tool call and natively awaits a result).
- `final` → delivered via the formal tool-result channel, same path as a synchronous tool result, just delayed.

A new shared module `pipecat.processors.aggregators.async_tool_messages` is the single source of truth for the on-the-wire payload shape — the aggregator uses its `build_*_message` functions when injecting messages, and the realtime services use `parse_message` when scanning the context.

**Limitation:** streamed intermediate results (reported via `FunctionCallResultProperties(is_final=False)`) are not supported on these realtime services. An intermediate result is dropped and an error is raised. Use a non-realtime LLM service if your tool needs to stream intermediate results. (Docstrings on `register_function`, `register_direct_function`, and `FunctionCallResultProperties.is_final` updated to call this out.)

Similar fixes for other realtime services are forthcoming.

## Test plan

- [x] Helper unit tests pass (`uv run pytest tests/test_async_tool_messages.py`)
- [x] Existing test suite still passes (`uv run pytest tests/test_context_aggregators_universal.py`)
- [x] AWS Nova Sonic — empirical validation with `examples/realtime/realtime-aws-nova-sonic-async-tool.py`
- [x] OpenAI Realtime — empirical validation with `examples/realtime/realtime-openai-async-tool.py`


examples/realtime/realtime-gemini-live-async-tool.py
kompfner, line 1: The rename you see here doesn't really explain what happened.

In reality: I _removed_ realtime-gemini-live-function-calling.py and made the "base" realtime-gemini-live.py also exercise tool calling, in keeping with the other realtime services' examples. And then I *added* a realtime-gemini-live-async-tool.py.

kompfner:
Originally in this PR I had wanted to include Gemini Live, but it looks like more work is still needed there - I believe that our Gemini Live service sadly never properly supported tools that took some time to run...

The silver lining of it never properly working is that our change to the `cancel_on_interruption=False` mechanism isn't as much of a regression there...

src/pipecat/frames/frames.py
```diff
@@ -695,6 +695,11 @@ class FunctionCallResultProperties:
         is_final: Whether this is the final result for the function call. When
             ``False`` the result is treated as an intermediate update. Defaults to ``True``.
             Only meaningful for async function calls (``cancel_on_interruption=False``).
+            Note: realtime LLM services do not support streamed intermediate
+            results; they deliver only the final result to the provider. An
+            intermediate result reported to a realtime service is dropped
+            and an error is raised. Use a non-realtime LLM service if your
+            tool needs to stream intermediate results.
```
filipi87, line 702: Yep! 👍

src/pipecat/services/aws/nova_sonic/llm.py
```diff
@@ -620,6 +621,38 @@ async def _process_completed_function_calls(self, send_new_results: bool):
...
+                    # awaits a result; nothing to send for the started marker.
+                    continue
+                if async_payload.kind == "intermediate":
+                    logger.error(
+                        f"{self}: Nova Sonic does not support streamed async "
+                        f"tool results; dropping intermediate result for "
+                        f"tool_call_id={async_payload.tool_call_id}. Use a "
+                        f"non-realtime LLM service if your tool needs to "
+                        f"stream intermediate results."
+                    )
```
filipi87, line 643: nit: I believe the `push_error` method already logs the error message. Do we want to log both messages ?

src/pipecat/services/openai/realtime/llm.py
```diff
@@ -1039,6 +1040,43 @@ async def _process_completed_function_calls(self, send_new_results: bool):
...
+                        f"non-realtime LLM service if your tool needs to "
+                        f"stream intermediate results."
+                    )
+                    await self.push_error(
+                        error_msg="OpenAI Realtime does not support streamed async tool results.",
+                    )
+                    continue
+                # kind == "final": deliver via the formal tool-result channel
+                # — same path as a synchronous tool result, just delayed.
+                if send_new_results:
```
filipi87: Should we also check whether `async_payload.kind == "final"` here, just to be extra careful?

src/pipecat/services/aws/nova_sonic/llm.py
```diff
@@ -620,6 +621,38 @@ async def _process_completed_function_calls(self, send_new_results: bool):
...
+                    await self.push_error(
+                        error_msg="Nova Sonic does not support streamed async tool results.",
+                    )
+                    continue
+                # kind == "final": deliver via the formal toolResult channel
+                # — same path as a synchronous tool result, just delayed.
+                if send_new_results:
```
filipi87: Same here. Should we also check whether async_payload.kind == "final" here, just to be extra careful?

filipi87 (approved):
LGTM. Thank you for fixing this @kompfner. Nice job!  🚀

src/pipecat/services/aws/nova_sonic/llm.py
kompfner: To me it makes sense that the more verbose instructional message be logged but not be pushed, but happy to change it if you feel strongly!

filipi87: No strong feelings either way. We can keep like this. 👍

src/pipecat/services/openai/realtime/llm.py
kompfner: Good idea. Will update in this PR and in the PRs that depend on this one.

kompfner: Done

src/pipecat/services/aws/nova_sonic/llm.py
kompfner: Done

PR #4442, opened 2026-05-07
Unified start route to make all transports available

filipi87:
## Summary
  
- Unified the `/start` endpoint so a single runner server supports all transports (WebRTC, Daily, telephony, and plain WebSocket) simultaneously — clients pick their transport via a `"transport"` field in the request body instead of requiring the server to be restarted for each transport type
- Added plain WebSocket transport support via a new `/ws-client` endpoint, enabling non-telephony WebSocket clients (e.g., browser apps using protobuf framing) to connect alongside other transports
- Added a `GET /status` endpoint that reports which transports the running instance accepts
- Made `-t/--transport` optional — omitting it enables all transports at once; passing it restricts `/start` to that transport only and sets it as the default
- Added `transport_type` field to `WebSocketRunnerArguments` so bots can distinguish plain WebSocket from telephony WebSocket connections
  
## Breaking Changes
  
- The Daily browser redirect route moved from `GET /` to `GET /daily`


examples/voice/voice-elevenlabs.py
```diff
@@ -47,6 +48,12 @@
         audio_in_enabled=True,
         audio_out_enabled=True,
     ),
+    "websocket": lambda: FastAPIWebsocketParams(
+        audio_in_enabled=True,
+        audio_out_enabled=True,
+        add_wav_header=False,
+        serializer=ProtobufFrameSerializer(),
+    ),
 }
```
filipi87, line 50: Should we change all our examples to support this ?

filipi87:
You can test it together with this voice-ui-kit PR:
- https://github.com/pipecat-ai/voice-ui-kit/pull/137

examples/voice/voice-elevenlabs.py
filipi87: I mean to support the websocket transport.

aconchillo: Does the Prebuilt UI support it?

aconchillo: I don't think we should include this for now. Websocket transport with protobuf is not something we give too much love to.

aconchillo (approved):

aconchillo:
LGTM! Awesome. I'd get rid of the elevenlabs example changes for now.

examples/voice/voice-elevenlabs.py
filipi87: Yep, the Prebuilt UI supports it.

markbackman (approved):
LGTM!

Should we also update the `runner` optional dependency to use the new `pipecat-ai-prebuilt` package?

examples/voice/voice-elevenlabs.py
markbackman: I agree with Aleix. We don't need this in our foundational examples. Maybe the better route is to add a websocket transports to examples/transports.

PR #4446, opened 2026-05-07
[inworld] default to using PCM encoding

ianbbqzy:
* server returns audio bytes without headers

https://docs.inworld.ai/api-reference/ttsAPI/texttospeech/synthesize-speech-stream#body-audio-config-audio-encoding


markbackman (approved):
LGTM!

markbackman (approved):
LGTM! Thanks for the update.

PR #4447, opened 2026-05-08
fix: extend cancel_on_interruption=False regression fix to remaining realtime services

kompfner:
## Summary

Extends the `cancel_on_interruption=False` regression fix from #4441 to additional realtime LLM services.

| Service | Status |
|---|---|
| `AWSNovaSonicLLMService` | Fixed in #4441 |
| `OpenAIRealtimeLLMService` | Fixed in #4441 |
| `GrokRealtimeLLMService` (xAI Realtime) | **Fixed here** |
| `AzureRealtimeLLMService` | **Free fix** — inherits `OpenAIRealtimeLLMService` |
| `UltravoxRealtimeLLMService` | **Omitted with warning** — see "Why Ultravox is omit-and-warn" |
| `GeminiLiveLLMService` | ⏸ Deferred — async-tool path needs deeper investigation |
| `InworldRealtimeLLMService` | ⏸ Deferred — see "Why Inworld is excluded" |

`GrokRealtimeLLMService`'s `_process_completed_function_calls` now matches the canonical pattern from #4441:

- Skip `LLMSpecificMessage` up top.
- Detect async-tool messages via `async_tool_messages.parse_message` and route them: `started` → skipped silently, `intermediate` → logged as error + surfaced via `push_error`, `final` → delivered via the formal tool-result channel.
- Fall through to the regular sync tool-result handling for non-async-tool messages.

**Limitation** carried forward from #4441: streamed intermediate results (`FunctionCallResultProperties(is_final=False)`) are not supported on these realtime services. Use a non-realtime LLM service if your tool needs to stream intermediate results.

## Why Ultravox is omit-and-warn

Per [Ultravox's documentation](https://docs.ultravox.ai/tools/async-tools#custom-tool-timeouts), the Ultravox API freezes the conversation during tool execution: "During tool execution, conversations are essentially frozen. Users can continue talking, but the agent won't respond until the tool completes." The whole *point* of `cancel_on_interruption=False` is to let the conversation continue while a tool runs in the background — that intent is structurally not achievable on Ultravox, so applying the same code pattern there would mislead users. (We confirmed empirically that even with the per-tool `timeout` bumped to its 40s maximum, the bot stays silent until the tool returns.)

Instead of silently misbehaving, `UltravoxRealtimeLLMService` now logs a one-time error and pushes an `ErrorFrame` whenever it sees async-tool messages in the context (i.e. whenever a function registered with `cancel_on_interruption=False` runs), explaining that the flag isn't supported and pointing users at `cancel_on_interruption=True` or a non-realtime LLM service.

If/when Ultravox grows true async tool support, we can revisit.

## Why Inworld is excluded

Inworld appears to have a pre-existing problem with even simple (synchronous) tool calling on its Realtime API — independent of any of this PR's work. Empirically:

- A `session.update` containing `tools` is accepted by the server (`session.updated` arrives back).
- The very next `response.create` returns `response.done` with `status="failed"` and a generic `status_details={'type': 'failed', 'reason': 'server_error'}`.
- This reproduces across multiple model names (`xai/grok-4-1-fast-non-reasoning`, `openai/gpt-4.1-mini`, `gemini-2.5-flash`, `openai/gpt-4o-mini`).
- The same example *without* tools works fine. The same wire format that we send is documented as a working tools example in [Inworld's WebSocket docs](https://docs.inworld.ai/docs/realtime/connect/websocket), and we're matching it field-for-field on the tool shape.

Since async-tool support for Inworld is meaningless until simple sync tool calling works, Inworld is excluded from this PR. Will investigate separately — suspect candidates are extra fields we send (`provider_data`, audio `format`/`transcription` defaults) that Inworld's docs example omits.

## Test plan

- [x] Existing test suite still passes (`uv run pytest tests/`)
- [ ] Grok Realtime — empirical validation with `examples/realtime/realtime-grok-async-tool.py`
- [x] Azure Realtime — empirical validation with `examples/realtime/realtime-azure-async-tool.py` (Azure Realtime service not actually available yet for testing, so we're just including this example in the hopes it will someday, like we do with the "base" example)

## Notes

This PR is stacked on top of #4441 — once that merges, the base will be retargeted to `main`.


changelog/4447.fixed.md
filipi87, line 1: Just wondering whether, for `UltravoxRealtimeLLMService`, it would be worth providing the same approach we used for the other **non** realtime LLMs, since it doesn't seem to support async tool calls anyway.

What do you think? Would it be worth it, or do you think the results would be pretty bad in this case?

src/pipecat/services/xai/realtime/llm.py
```diff
@@ -913,6 +914,43 @@ async def _process_completed_function_calls(self, send_new_results: bool):
...
+                        f"non-realtime LLM service if your tool needs to "
+                        f"stream intermediate results."
+                    )
+                    await self.push_error(
+                        error_msg="Grok Realtime does not support streamed async tool results.",
+                    )
+                    continue
+                # kind == "final": deliver via the formal tool-result channel
+                # — same path as a synchronous tool result, just delayed.
+                if send_new_results:
```
filipi87: The same comment that I did in the other PR, should we also check the kind == "final" here, just to be extra carefull ?

filipi87 (approved):
LGTM!

changelog/4447.fixed.md
kompfner: My hopes would be pretty low of this working successfully...but we could give it a shot.

kompfner: Actually...looks like Ultravox docs even suggest doing this! So good call! Let's do it.

src/pipecat/services/xai/realtime/llm.py
kompfner: Done

changelog/4447.fixed.md
kompfner: Works pretty well! Still tinkering a little bit, but this should be good to go soon.

kompfner:
Going ahead and merging to unblock the next PR (#4448), which is dependent on this one, but @filipi87 you might want to take a look at the latest Ultravox changes.

PR #4448, opened 2026-05-08
feat: support cancel_on_interruption=False on Gemini Live (Gemini 2.x)

kompfner:
## Summary

Adds `cancel_on_interruption=False` support for `GeminiLiveLLMService` on models that support Gemini's `NON_BLOCKING` tool mechanism (currently Gemini 2.x). The flag wasn't really working before — pipecat never set Gemini's `NON_BLOCKING` `behavior` on function declarations or paired it with `scheduling: WHEN_IDLE` on responses, so the conversation would stall waiting for tool results regardless of how the function was registered. This PR wires both up so registering a function with `cancel_on_interruption=False` actually delivers the intended UX: the model keeps talking while the tool runs, and the result lands at a graceful pause.

### `_supports_non_blocking_tools` gate

A new property gates the entire NON_BLOCKING path. Currently returns `not self._is_gemini_3` — Gemini 3.x has not yet shipped support for the `NON_BLOCKING` `behavior` field on function declarations or for the `scheduling` field on FunctionResponse. When Gemini 3.x grows that support, the gate's body will be the only thing that needs to change.

### Per-tool NON_BLOCKING (only `cancel_on_interruption=False` tools)

In the connection-config setup, every function declaration whose corresponding registered handler has `cancel_on_interruption=False` is tagged with `behavior: NON_BLOCKING`. Synchronous (default) tools stay BLOCKING — applying NON_BLOCKING uniformly was producing filler responses like "let me look that up for you" on regular calls, since the model knew it would have an opportunity to keep talking while waiting. The lookup mirrors `LLMService.run_function_calls` (name-specific entry preferred, falls back to the `None`-keyed catch-all) via a small `_function_is_async` helper.

### `scheduling: WHEN_IDLE` on async tool results

In `_tool_result`, NON_BLOCKING tools' `FunctionResponse` carries `scheduling: WHEN_IDLE` so the result is integrated at a natural pause rather than mid-sentence. Synchronous tools' responses don't get this — the model isn't talking when their results arrive anyway, so it's a no-op there.

### Service-side refactor (foundational)

`_process_completed_function_calls` is now a single-pass forward iteration over raw context messages, matching the canonical pattern from #4441 / #4447. The two-pass adapter-based scan is gone; tool name lookups use a service-level `_tool_call_id_to_name` map populated when the model issues calls. A new `GeminiLLMAdapter.to_function_response_dict` static method holds the dict-coercion rule (non-dict tool returns get wrapped as `{value: <result>}`) so this Gemini-format knowledge stays in the adapter where it belongs.

### Warning on unsupported models

On Gemini 3.x, the service surfaces a one-time `logger.error` + `push_error` the moment async-tool messages first appear in the context, explaining that `cancel_on_interruption=False` isn't properly supported on the current model.

### Example consolidation

- Folds `realtime-gemini-live-function-calling.py` into the base `realtime-gemini-live.py` example so the base exercises function calling out of the box (matching `realtime-openai.py` and `realtime-aws-nova-sonic.py`).
- Renames `realtime-gemini-live-vertex-function-calling.py` → `realtime-gemini-live-vertex.py`.
- Adds `realtime-gemini-live-async-tool.py`.
- Updates `scripts/evals/run-release-evals.py` for the renames.

## Known caveat (Gemini 2.5)

An intermittent server-side 1008 "Operation is not implemented or supported or enabled" error can fire when realtime input arrives during a pending tool call. We auto-reconnect from these but the user may need to repeat what they were saying. Per [this Google AI dev forum thread](https://discuss.ai.google.dev/t/gemini-live-api-websocket-error-1008-operation-is-not-implemented-or-supported-or-enabled/114644/56), the suggested mitigation is to gate realtime input during pending tool calls — that's fundamentally incompatible with the whole point of NON_BLOCKING tool calling, so we don't apply it. Tracking pending upstream resolution.

## Test plan

- [x] Existing test suite still passes (`uv run pytest tests/`)
- [x] Gemini Live (Gemini 2.5 default model) — empirical validation with `examples/realtime/realtime-gemini-live-async-tool.py` confirms async tool calling works as intended

## Notes

This PR is stacked on top of #4441 — once that merges, the base will be retargeted to `main`.


kompfner:
One final thing to look into on this one: Vertex.

src/pipecat/services/google/gemini_live/llm.py
```diff
@@ -372,6 +374,27 @@ def _is_gemini_3(self) -> bool:
...
         return "gemini-3" in (assert_given(self._settings.model) or "")
 
+    @property
+    def _supports_non_blocking_tools(self) -> bool:
+        """Whether the current model supports the NON_BLOCKING tool behavior + scheduling hints.
+
+        Gemini 3.x has not yet shipped support for NON_BLOCKING function
+        declarations or for the ``scheduling`` field on FunctionResponse.
+        """
+        return not self._is_gemini_3
```
filipi87, line 384: Just to double check, Gemini 2.x supports non blocking function calls, but not Gemini 3 ?

src/pipecat/services/google/gemini_live/llm.py
```diff
@@ -1047,6 +1140,27 @@ async def _connect(self, session_resumption_handle: str | None = None):
...
+                            continue
+                        decls = tool.get("function_declarations")
+                        if not isinstance(decls, list):
+                            continue
+                        for decl in decls:
+                            if not isinstance(decl, dict):
+                                continue
+                            name = decl.get("name")
+                            if isinstance(name, str) and self._function_is_async(name):
+                                decl["behavior"] = "NON_BLOCKING"
```
filipi87, line 1151: So, we didn’t use to support async tools with Gemini before ?

filipi87 (approved):
LGTM!

src/pipecat/services/google/gemini_live/llm.py
kompfner: Yes. Admittedly a bit counter-intuitive.

src/pipecat/services/google/gemini_live/llm.py
kompfner: Not properly, at least. They wouldn't behave quite right.

src/pipecat/services/google/gemini_live/llm.py
```diff
@@ -1047,6 +1128,27 @@ async def _connect(self, session_resumption_handle: str | None = None):
                 logger.debug(f"Setting system instruction: {system_instruction}")
                 config.system_instruction = system_instruction
             if tools:
+                # Tag function declarations registered with
+                # cancel_on_interruption=False as NON_BLOCKING so Gemini
```
kompfner, line 1132: This highlights that we should always be recommending that folks flag functions that may take a while as `cancel_on_interruption=False`

kompfner:
> One final thing to look into on this one: Vertex.

Vertex doesn't support the NON_BLOCKING mechanism. Excluding it by overriding `_supports_non_blocking_tools`.

PR #4450, opened 2026-05-08
Default OpenAI Realtime transcription to gpt-realtime-whisper

markbackman:
## Summary

- Switch the default OpenAI Realtime input audio transcription model from `gpt-4o-transcribe` to `gpt-realtime-whisper` in both `OpenAIRealtimeSTTService` and `OpenAIRealtimeLLMService`.
- `gpt-realtime-whisper` does not accept a `prompt` parameter, so transcription prompts are dropped automatically (with a warning) when that model is selected — both at construction time and when settings are updated at runtime.
- Update the OpenAI examples (`function-calling-openai*`, `transcription-openai`, `voice-openai`) to use the new default.

## Breaking Changes

- Users who relied on the old default (`gpt-4o-transcribe`) for OpenAI Realtime transcription will now use `gpt-realtime-whisper`. To keep the previous behavior, explicitly pin `model="gpt-4o-transcribe"` (or `"gpt-4o-mini-transcribe"`). Prompt hints are silently dropped when used with `gpt-realtime-whisper`.

## Testing

- `uv run pytest tests/test_settings.py`
- Manually exercise `examples/transcription/transcription-openai.py` and `examples/voice/voice-openai.py` against the OpenAI Realtime API.


examples/voice/voice-openai.py
```diff
@@ -72,7 +71,7 @@ async def run_bot(transport: BaseTransport, runner_args: RunnerArguments):
     llm = OpenAILLMService(
         api_key=os.environ["OPENAI_API_KEY"],
         settings=OpenAILLMService.Settings(
-            system_instruction="You are very knowledgable about dogs. Your output will be spoken aloud, so avoid special characters that can't easily be spoken, such as emojis or bullet points. Respond to what the user said in a creative and helpful way.",
+            system_instruction="You are a helpful assistant in a voice conversation. Your responses will be spoken aloud, so avoid emojis, bullet points, or other formatting that can't be spoken. Respond to what the user said in a creative, helpful, and brief way.",
```
torontodeveloper, line 67: just curious, is this not hardcoding in the middle of the file, wondering should we define variable at top of the file and reuse it? i don't know, just thinking out loude

src/pipecat/services/openai/realtime/llm.py
```diff
@@ -650,8 +674,11 @@ async def _ws_send(self, realtime_message):
     async def _update_settings(self, delta):
         """Apply a settings delta, sending a session update when needed."""
         changed = await super()._update_settings(delta)
+        prompt_omitted = self._omit_unsupported_input_audio_transcription_prompt(
+            assert_given(self._settings.session_properties)
+        )
         handled = {"session_properties", "system_instruction"}
-        if changed.keys() & handled:
+        if changed.keys() & handled or prompt_omitted:
```
torontodeveloper: since we are using bitwise AND operator, so changed.keys() returns bits and so handled and prompt_omitted?

src/pipecat/services/openai/stt.py
```diff
@@ -227,7 +229,7 @@ class OpenAIRealtimeSTTService(WebsocketSTTService):
         stt = OpenAIRealtimeSTTService(
             api_key="sk-...",
             settings=OpenAIRealtimeSTTService.Settings(
-                model="gpt-4o-transcribe",
+                model="gpt-realtime-whisper",
```
filipi87, line 229: nit: Should we use the constant (`_OPENAI_REALTIME_WHISPER_MODEL`) instead ?

src/pipecat/services/openai/stt.py
```diff
@@ -303,7 +308,7 @@ def __init__(
 
         # --- 1. Hardcoded defaults ---
         default_settings = self.Settings(
-            model="gpt-4o-transcribe",
+            model="gpt-realtime-whisper",
```
filipi87: Same here. Should we use the constant ?

examples/function-calling/function-calling-openai-async.py
```diff
@@ -69,11 +69,10 @@ async def fetch_restaurant_recommendation(params: FunctionCallParams):
...
     logger.info(f"Starting bot")
 
-    stt = OpenAISTTService(
+    stt = OpenAIRealtimeSTTService(
         api_key=os.environ["OPENAI_API_KEY"],
-        settings=OpenAISTTService.Settings(
-            model="gpt-4o-transcribe",
-            prompt="Expect words related weather, such as temperature and conditions. And restaurant names.",
+        settings=OpenAIRealtimeSTTService.Settings(
+            model="gpt-realtime-whisper",
```
filipi87: Should we pass the model here if we’re already using the default one ? Or could we just remove it ?

examples/function-calling/function-calling-openai.py
```diff
@@ -63,11 +63,10 @@ async def fetch_restaurant_recommendation(params: FunctionCallParams):
...
     logger.info(f"Starting bot")
 
-    stt = OpenAISTTService(
+    stt = OpenAIRealtimeSTTService(
         api_key=os.environ["OPENAI_API_KEY"],
-        settings=OpenAISTTService.Settings(
-            model="gpt-4o-transcribe",
-            prompt="Expect words related weather, such as temperature and conditions. And restaurant names.",
+        settings=OpenAIRealtimeSTTService.Settings(
+            model="gpt-realtime-whisper",
```
filipi87: Same here. Should we pass the model here if we’re already using the default one ?

filipi87: And in the other examples.

filipi87 (approved):
LGTM! Just a couple of nits.

src/pipecat/services/openai/stt.py
torontodeveloperonit: probably, i would agree to it too.

examples/voice/voice-openai.py
markbackman: The idea is that these files are runnable without additional dependencies, so a prompt in plain text, while not ideal does make this possible.

src/pipecat/services/openai/stt.py
markbackman: This is a docstring, so being explicit is helpful.

src/pipecat/services/openai/stt.py
markbackman: Yes, this should be the constant. Will update.

src/pipecat/services/openai/realtime/llm.py
markbackman: Good callout. The original expression was using set intersection (`changed.keys() & handled`), not bitwise math over flags, so it was functionally correct. I updated it to assign the intersection result to an explicit boolean first so the intent is clearer.

src/pipecat/services/openai/tts.py
```diff
@@ -85,8 +86,6 @@ class OpenAITTSService(TTSService):
     Settings = OpenAITTSSettings
     _settings: Settings
 
-    OPENAI_SAMPLE_RATE = 24000  # OpenAI TTS always outputs at 24kHz
```
markbackman, line 88: Before I merge @filipi87 or @aconchillo, I'm consolidating constants into a private _constants.py in this directory.

I'm also removing `OPENAI_SAMPLE_RATE`, which wasn't meant to be public—never documented or in any changelog. I think this is safe to remove, but if you have issues, I can keep. Let me know.

filipi87: Hi @markbackman, I am ok with this approach. Since we are using the constants in a couple of different classes, consolidating them into a single file makes sense.

And I agree that removing `OPENAI_SAMPLE_RATE` should be safe enough. 👍

filipi87 (approved):

PR #4454, opened 2026-05-11
Fix Korean Cartesia timestamp spacing

seongho-park-playmore:
## What changed
- Keep Cartesia Chinese and Japanese timestamp groups joined without spaces.
- Join Cartesia Korean timestamp groups with spaces so assistant transcripts preserve Korean word spacing.
- Move the timestamp normalization into a small Cartesia utility so the behavior is unit tested without websocket extras.
- Added the required changelog fragment: changelog/4454.fixed.md.

## Why
Cartesia groups CJK timestamp tokens before Pipecat rebuilds assistant transcript text. Korean uses spaces between words, so handling it like Chinese/Japanese can collapse Korean spacing in displayed assistant transcripts.

## Validation
- uv run pytest tests/test_cartesia_tts.py
- uv run --extra websockets-base pytest tests/test_tts_frame_ordering.py tests/test_context_aggregators_universal.py
- uv run ruff format --diff
- uv run ruff check
- uv run pyright src/pipecat/services/cartesia/tts.py src/pipecat/services/cartesia/utils.py tests/test_cartesia_tts.py
- uv run towncrier build --draft --version Unreleased

Note: full uv run pyright currently reports pre-existing errors outside this PR in scripts/daily/test_tavus_transport.py and src/pipecat/utils/tracing/turn_trace_observer.py; changed files pass targeted pyright.

Supersedes #4453, which was closed when the fork branch was renamed.


markbackman:
Hey! Thanks for offering a solution. In looking closer two things are happening:
1. Korean doesn't need to be special cased like Chinese and Japanese. Cartesia returns words, so we can just use it straight away.
2. Something recently regressed the Chinese and Japanese spacing.

I made this PR which fixes both:
https://github.com/pipecat-ai/pipecat/pull/4475

I'm going to close out this PR, but thank you for raising this 🙏

PR #4455, opened 2026-05-11
Fix PipelineTask cancellation before StartFrame reaches sink

### LINKED ISSUE #4276: PipelineTask deadlock: cancel() before StartFrame reaches sink hangs worker permanently
# PipelineTask deadlock: cancel() before StartFrame reaches sink hangs worker permanently

## Summary

When `task.cancel()` is called while `_process_push_queue` is blocked in `_wait_for_pipeline_start()`, the queued `CancelFrame` can never be consumed and the pipeline worker hangs indefinitely in a busy state. This regression was introduced in v0.0.108 and is still present in `main`.

## Environment

- **pipecat-ai version:** 0.0.108 (latest; confirmed present in `main` as of 2026-04-13)
- **Transport:** Daily
- **Python:** 3.13

## Steps to Reproduce

1. Start a `PipelineTask` with a Daily transport.
2. Call `task.cancel()` within the first ~1–2 seconds of session start (i.e. before the `StartFrame` has finished propagating to the sink — typically during the client connection handshake or cold-start initialisation).
3. Observe that the task never finishes: `has_finished()` remains `False` and the worker process stays in a `BUSY` state permanently.

A concrete trigger: a user clicks a "Stop session" / disconnect button immediately after initiating a session, before the pipeline has fully started.

## Root Cause

`_process_push_queue` blocks at line 777 (v0.0.108 / `main`):

```python
await self._wait_for_pipeline_start(start_frame)  # blocks here until _pipeline_start_event is set

while running:
    frame = await self._push_queue.get()           # CancelFrame is stuck here — never reached
    ...
```

`_pipeline_start_event` is only set when the `StartFrame` reaches the sink (`_sink_push_frame`, line 853). If `task.cancel()` is called before that happens, `_cancel()` enqueues a `CancelFrame` via `queue_frame()`:

```python
async def _cancel(self, *, reason=None):
    ...
    await self.queue_frame(CancelFrame(reason=reason))  # → puts CancelFrame in _push_queue
```

But `_process_push_queue` is still blocked waiting for `_pipeline_start_event.wait()`. The `CancelFrame` sits in `_push_queue` forever. `_pipeline_end_event` is never set, `_pipeline_finished_event` is never set, and `run()` waits on `_wait_for_pipeline_finished()` indefinitely. The worker is permanently exhausted.

## Why This Is a Regression

Prior to v0.0.108, `_process_push_queue` did not call `_wait_for_pipeline_start()` before entering the frame loop. The `_wait_for_pipeline_start` guard was added in v0.0.108 to ensure frames aren't processed before the pipeline is ready — a valid goal — but the cancellation path was not updated to account for the queue being blocked.

Notably, the codebase already handles a similar "blocked push queue" scenario for interruptions (line 821–823):

```python
# InterruptionTaskFrame bypasses the push queue directly
await self._pipeline.queue_frame(InterruptionFrame())
```

The same bypass pattern is not applied for the cancel path.

## Workaround

Manually set `_pipeline_start_event` before calling `task.cancel()` to unblock the queue worker:

```python
async def _safe_cancel(task: PipelineTask) -> None:
    try:
        if not task.has_finished():
            ev = getattr(task, "_pipeline_start_event", None)
            if ev and not ev.is_set():
                ev.set()  # unblock _wait_for_pipeline_start so CancelFrame can be consumed
            await task.cancel()
    except Exception as e:
        logger.error(f"Error cancelling task: {e}")

asyncio.create_task(_safe_cancel(task))
```

This works but relies on a private attribute (`_pipeline_start_event`), making it fragile across future pipecat versions.

## Suggested Fix

In `_cancel()`, check whether `_pipeline_start_event` has been set and set it if not, before queuing the `CancelFrame`:

```python
async def _cancel(self, *, reason: Optional[str] = None):
    if not self._cancelled:
        logger.debug(f"Cancelling pipeline task {self}")
        self._cancelled = True
        # Unblock _process_push_queue if StartFrame hasn't reached the sink yet,
        # otherwise CancelFrame would be enqueued but never consumed.
        if not self._pipeline_start_event.is_set():
            self._pipeline_start_event.set()
        await self.queue_frame(CancelFrame(reason=reason))
```

Alternatively, the `CancelFrame` could bypass the push queue entirely (as `InterruptionFrame` does), though that may have different ordering implications.

## Impact

Any downstream application that allows users to cancel a session during the first few seconds of start-up (before the pipeline is fully initialised) will permanently exhaust its worker pool. No recovery is possible without restarting the worker process.

eecczz:
Fixes #4276.

This prevents `PipelineTask.cancel()` from hanging when cancellation is requested before the initial `StartFrame` reaches the pipeline sink.

Previously, `_process_push_queue()` could be blocked in `_wait_for_pipeline_start()`, while the `CancelFrame` was queued behind that wait and never consumed. This change unblocks the startup wait during cancellation so the queued `CancelFrame` can be processed normally.

Added a regression test that blocks `StartFrame` before it reaches the sink, calls `task.cancel()`, and verifies the task finishes instead of hanging.

Testing:
- `uv run pytest tests/test_pipeline.py::TestPipelineTask::test_task_cancel_before_start_reaches_sink`
- `uv run pytest tests/test_pipeline.py`
- `uv run ruff check src/pipecat/pipeline/task.py tests/test_pipeline.py`

### DIFF
--- a/changelog/4455.fixed.md
+++ b/changelog/4455.fixed.md
@@ -1,0 +1,1 @@
+- Fixed `PipelineTask.cancel()` hanging when cancellation is requested before the initial `StartFrame` reaches the pipeline sink.
--- a/src/pipecat/pipeline/task.py
+++ b/src/pipecat/pipeline/task.py
@@ -650,6 +650,8 @@ class PipelineTask(BasePipelineTask):
         if not self._cancelled:
             logger.debug(f"Cancelling pipeline task {self}")
             self._cancelled = True
+            if not self._pipeline_start_event.is_set():
+                self._pipeline_start_event.set()
             await self.queue_frame(CancelFrame(reason=reason))
 
     async def _create_tasks(self):
--- a/tests/test_pipeline.py
+++ b/tests/test_pipeline.py
@@ -625,6 +625,33 @@ class TestPipelineTask(unittest.IsolatedAsyncioTestCase):
         except asyncio.CancelledError:
             assert cancelled
 
+    async def test_task_cancel_before_start_reaches_sink(self):
+        class StartBlocker(FrameProcessor):
+            def __init__(self, *, start_received: asyncio.Event, **kwargs):
+                super().__init__(**kwargs)
+                self._start_received = start_received
+                self._block = asyncio.Event()
+
+            async def process_frame(self, frame: Frame, direction: FrameDirection):
+                await super().process_frame(frame, direction)
+
+                if isinstance(frame, StartFrame):
+                    self._start_received.set()
+                    await self._block.wait()
+
+                await self.push_frame(frame, direction)
+
+        start_received = asyncio.Event()
+        pipeline = Pipeline([StartBlocker(start_received=start_received)])
+        task = PipelineTask(pipeline, cancel_timeout_secs=0.1)
+
+        run_task = asyncio.create_task(task.run(PipelineTaskParams(loop=asyncio.get_event_loop())))
+        await start_received.wait()
+        await task.cancel()
+        await asyncio.wait_for(run_task, timeout=1.0)
+
+        assert task.has_finished()
+
     async def test_task_error(self):
         class ErrorProcessor(FrameProcessor):
             def __init__(self, **kwargs):


eecczz:
Note: I did not add a changelog fragment yet because the repository uses PR-numbered fragment filenames. Happy to add `changelog/<PR_NUMBER>.fixed.md` if desired.

markbackman:
Thanks! This looks like the right fix, but getting a second opinion.

@eecczz, please add a changelog with the filename of `4455.fixed.md` in the `changelog` folder.

@aconchillo, @filipi87 I think this is fine—forces a shutdown in the event an immediate CancelFrame is pushed before the StartFrame reaches the PipelineTask sink.

eecczz:
Thanks! Added `changelog/4455.fixed.md` in 96e0c76.

filipi87 (approved):
Makes sense. Nice catch!

PR #4463, opened 2026-05-12
feat(deepgram-stt): add filler_words init parameter

akash-dutta-dev:
## Summary

Adds `filler_words: bool | None = None` as a typed init parameter on`DeepgramSTTService`, matching the existing `mip_opt_out` pattern.

## Why a new init param instead of `Settings.extra`?

`filler_words` is supported by Deepgram's REST API, but the installed `deepgram-sdk`'s typed `AsyncV1Client.connect()` signature does not expose it as a kwarg. 
Passing it via
`Settings(extra={"filler_words": True})` 
currently crashes: AsyncV1Client.connect() got an unexpected keyword argument 'filler_words'

### DIFF
--- a/src/pipecat/services/deepgram/stt.py
+++ b/src/pipecat/services/deepgram/stt.py
@@ -304,6 +304,7 @@ class DeepgramSTTService(STTService):
         callback_method: str | None = None,
         tag: Any | None = None,
         mip_opt_out: bool | None = None,
+        filler_words: bool | None = None,
         live_options: LiveOptions | None = None,
         addons: dict | None = None,
         settings: Settings | None = None,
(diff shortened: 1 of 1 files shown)


markbackman:
Thanks for digging into this. I don't think we should add this to `DeepgramSTTService`.

Deepgram's `filler_words` feature appears to be pre-recorded only. The feature page marks both `Streaming: Nova` and `Streaming: Flux` as unsupported, and the example uses the POST `/v1/listen` endpoint rather than the WSS streaming endpoint:

https://developers.deepgram.com/docs/filler-words

Since `DeepgramSTTService` uses live streaming, exposing this setting here would likely add an unsupported/no-op parameter. Closing for now unless we can confirm from Deepgram that `filler_words=true` is supported on WSS live audio.

markbackman:
Closing because Deepgram documents `filler_words` as unsupported for streaming, and `DeepgramSTTService` uses WSS live audio.

PR #4464, opened 2026-05-12
NVidia sagemaker - TTS and STT services

filipi87:
## Summary
- Add NVIDIA Nemotron ASR STT service (`NvidiaSageMakerWebsocketSTTService`) backed by an AWS SageMaker bidirectional-stream endpoint, which proxies to NVIDIA NIM's realtime WebSocket
- Add NVIDIA Magpie TTS services backed by AWS SageMaker:
  - `NvidiaSageMakerHTTPTTSService` — synchronous HTTP `invoke_endpoint` call, streams raw PCM audio back
  - `NvidiaSageMakerWebsocketTTSService` — persistent HTTP/2 bidi-stream session with interruption support via `InterruptibleTTSService`


src/pipecat/services/nvidia/sagemaker/stt.py
```diff
@@ -0,0 +1,340 @@
...
+                        await self.stop_processing_metrics()
+
+                elif event_type in (
+                    "conversation.item.input_audio_transcription.failed",
+                    "error",
+                ):
+                    await self.push_error(error_msg=f"NIM ASR error: {msg}")
+                    # In case of error we need to reconnect, otherwise we are not going to receive from the STT service anymore
+                    await self._disconnect()
+                    await self._connect()
```
markbackman: **Broken reconnect-on-error: `_process_responses` cancels its own task.**

`_process_responses()` *is* the body of `self._response_task` (set in [`_connect`](https://github.com/pipecat-ai/pipecat/blob/7f98dba925dc9c0bc00b3321aaa2e35b9c645fe9/src/pipecat/services/nvidia/sagemaker/stt.py#L196-L213)). Here, on a NIM ASR `error` / `...transcription.failed` event, it calls `await self._disconnect()` — but [`_disconnect()`](https://github.com/pipecat-ai/pipecat/blob/7f98dba925dc9c0bc00b3321aaa2e35b9c645fe9/src/pipecat/services/nvidia/sagemaker/stt.py#L215-L228) does `await self.cancel_task(self._response_task)`, i.e. the running task cancels and awaits *itself*.

`cancel_task()` calls `task.cancel()` then `await task`; the resulting `CancelledError` is raised at that `await` and swallowed inside `cancel_task` (`except asyncio.CancelledError: pass`), so this `_process_responses` invocation does **not** terminate — it continues past `_disconnect()` into `await self._connect()`, which spawns a *second* `_process_responses` task. You end up with two tasks racing on the same bidi client (duplicated `TranscriptionFrame`/`InterimTranscriptionFrame`, duplicated metrics calls, and the old task leaked since `self._response_task` now points only at the new one). On every subsequent error it compounds.

The reconnect must happen off the response task — e.g. `break` out of the loop here and have a separate supervisor (or `_disconnect`/`_connect` driven from outside the task) restart it. Compare the TTS service's `_receive_messages`, which just `break`s instead of calling `_disconnect()` from within itself.

markbackman (commented):
## Code review

Reviewed for bugs and CLAUDE.md compliance. Two issues, posted inline.

src/pipecat/services/nvidia/sagemaker/stt.py
```diff
@@ -0,0 +1,340 @@
...
+@dataclass
+class NvidiaSageMakerWSSTTSettings(STTSettings):
+    """Settings for NvidiaSageMakerWebsocketSTTService.
+
+    Parameters:
+        language: ISO-639-1 language code passed to NIM (e.g. ``en-US``).
+    """
+
+
+class NvidiaSageMakerWebsocketSTTService(STTService):
```
markbackman: For other WSS services, we don't include "Websocket" in the class name. Maybe we should do the same here? This would help to shorten the name.

```suggestion
class NvidiaSageMakerSTTService(STTService):
```

src/pipecat/services/nvidia/sagemaker/stt.py
```diff
@@ -0,0 +1,340 @@
...
+from pipecat.processors.frame_processor import FrameDirection
+from pipecat.services.aws.sagemaker.bidi_client import SageMakerBidiClient
+from pipecat.services.settings import STTSettings
+from pipecat.services.stt_service import STTService
+from pipecat.utils.time import time_now_iso8601
+from pipecat.utils.tracing.service_decorators import traced_stt
+
+
+@dataclass
+class NvidiaSageMakerWSSTTSettings(STTSettings):
```
markbackman: Drop `WS` to shorten the name.

```suggestion
class NvidiaSageMakerSTTSettings(STTSettings):
```

src/pipecat/services/nvidia/sagemaker/tts.py
```diff
@@ -0,0 +1,505 @@
...
+        settings: NvidiaSageMakerTTSSettings | None = None,
+        **kwargs,
+    ):
+        """Initialize the SageMaker HTTP TTS service.
+
+        Args:
+            endpoint_name: Name of the deployed SageMaker endpoint.
+            region: AWS region where the endpoint lives.
+            sample_rate: Output sample rate in Hz. Defaults to bot's pipeline rate.
+            params: Deprecated — use ``settings`` instead.
```
markbackman: Remove `params` since deprecation doesn't apply.

```suggestion
```

src/pipecat/services/nvidia/sagemaker/tts.py
```diff
@@ -0,0 +1,505 @@
...
+                    )
+        except Exception as e:
+            logger.error(f"{self}: SageMaker TTS error: {e}")
+            yield ErrorFrame(error=f"SageMaker TTS error: {e}")
+        finally:
+            await self.stop_ttfb_metrics()
+
+
+@dataclass
+class NvidiaSageMakerWSTTSSettings(TTSSettings):
```
markbackman: We might be able to remove this class since the settings are the same as `NvidiaSageMakerTTSSettings`.

src/pipecat/services/nvidia/sagemaker/tts.py
```diff
@@ -0,0 +1,505 @@
...
+    """Settings for NvidiaSageMakerWebsocketTTSService.
+
+    Parameters:
+        voice: NIM voice name (e.g. ``Magpie-Multilingual.EN-US.Aria``).
+        language: BCP-47 language code passed to NIM (e.g. ``en-US``).
+    """
+
+
+class NvidiaSageMakerWebsocketTTSService(InterruptibleTTSService):
```
markbackman: Drop `Websocket` to align with the naming convention of similar services.

```suggestion
class NvidiaSageMakerTTSService(InterruptibleTTSService):
```

src/pipecat/services/nvidia/sagemaker/tts.py
```diff
@@ -0,0 +1,505 @@
...
+            endpoint_name=os.getenv("SAGEMAKER_MAGPIE_ENDPOINT_NAME"),
+            region=os.getenv("AWS_REGION", "us-west-2"),
+            settings=NvidiaSageMakerWebsocketTTSService.Settings(
+                voice="Magpie-Multilingual.EN-US.Aria",
+                language="en-US",
+            ),
+        )
+    """
+
+    Settings = NvidiaSageMakerWSTTSSettings
```
markbackman: Use a common settings class.

```suggestion
    Settings = NvidiaSageMakerTTSSettings
```

src/pipecat/services/nvidia/sagemaker/tts.py
```diff
@@ -0,0 +1,505 @@
...
+
+    def __init__(
+        self,
+        *,
+        endpoint_name: str,
+        region: str = "us-west-2",
+        sample_rate: int | None = None,
+        settings: NvidiaSageMakerWSTTSSettings | None = None,
```
markbackman: Same here:
```suggestion
        settings: NvidiaSageMakerTTSSettings | None = None,
```

src/pipecat/utils/text/tts_text_sanitizer.py
```diff
@@ -0,0 +1,131 @@
+#
+# Copyright (c) 2024-2026, Daily
+#
+# SPDX-License-Identifier: BSD 2-Clause License
+#
+
+"""Utility for stripping non-speakable characters and markdown formatting from text.
```
markbackman: Interesting. Why is this needed? I'm assuming the NVIDIA docs point this out? Maybe these characters break the TTS in some way.

markbackman (commented):
Looks good! Just a few comments.

Should we add the simple examples here? I know that requires more work, like setting up the SageMaker endpoint, but including here will be handy for our testing, I think.

src/pipecat/services/nvidia/sagemaker/stt.py
filipi87: Done!

src/pipecat/services/nvidia/sagemaker/stt.py
filipi87: Done!

src/pipecat/services/nvidia/sagemaker/tts.py
filipi87: Done!

src/pipecat/services/nvidia/sagemaker/tts.py
filipi87: Done!

src/pipecat/services/nvidia/sagemaker/tts.py
filipi87: Done!

src/pipecat/services/nvidia/sagemaker/tts.py
filipi87: Done!

src/pipecat/services/nvidia/sagemaker/tts.py
filipi87: Done!

examples/voice/voice-nvidia-sagemaker.py
filipi87, line 1: @markbackman, a new example using `NvidiaSageMakerSTTService` and `NvidiaSageMakerTTSService`.

I’ve already updated our `.env` file to include the required credentials.

src/pipecat/utils/text/tts_text_sanitizer.py
filipi87: To be honest, I brought this over from Kwin’s first example, and I kept it while developing this integration because the LLM was generating many special characters, which did not work well with TTS.

However, the only real requirement is that we have some valid characters.

What do you think? Should we keep it or remove it? Or maybe hide it behind a flag?

src/pipecat/services/nvidia/sagemaker/stt.py
```diff
@@ -0,0 +1,348 @@
...
+
+        Args:
+            frame: The cancel frame.
+        """
+        await super().cancel(frame)
+        await self._disconnect()
+
+    # ── Audio input ───────────────────────────────────────────────────────────
+
+    async def run_stt(self, audio: bytes) -> AsyncGenerator[Frame | None, None]:
```
aconchillo, line 159: Missing `@traced_stt`

filipi87: It is inside `_handle_transcription`:

```
@traced_stt
async def _handle_transcription(self, transcript: str, is_final: bool, language=None):
  pass
```

src/pipecat/utils/text/tts_text_sanitizer.py
aconchillo: I'd remove it. If someone has issues with the TTS, they can use the text filters and transformers we have in the TTS (like the markdown one) that this is one is doing as well.

filipi87: Done.

src/pipecat/services/nvidia/sagemaker/stt.py
filipi87: Fixed. Nice catch!

examples/voice/voice-nvidia-sagemaker.py
```diff
@@ -0,0 +1,129 @@
...
+        audio_out_enabled=True,
+    ),
+}
+
+
+async def run_bot(transport: BaseTransport, runner_args: RunnerArguments):
+    logger.info(f"Starting bot")
+
+    stt = NvidiaSageMakerSTTService(
+        endpoint_name=os.getenv("SAGEMAKER_ASR_ENDPOINT_NAME"),
```
markbackman: Nit: conforming with the other examples:
```suggestion
        endpoint_name=os.environ["SAGEMAKER_ASR_ENDPOINT_NAME"],
```

examples/voice/voice-nvidia-sagemaker.py
```diff
@@ -0,0 +1,129 @@
...
+    llm = NvidiaLLMService(
+        api_key=os.environ["NVIDIA_API_KEY"],
+        settings=NvidiaLLMService.Settings(
+            model="meta/llama-3.3-70b-instruct",
+            system_instruction="You are a helpful assistant in a voice conversation. Your responses will be spoken aloud, so avoid emojis, bullet points, or other formatting that can't be spoken. Respond to what the user said in a creative, helpful, and brief way.",
+        ),
+    )
+
+    tts = NvidiaSageMakerTTSService(
+        endpoint_name=os.getenv("SAGEMAKER_MAGPIE_ENDPOINT_NAME"),
```
markbackman: Same here
```suggestion
        endpoint_name=os.environ["SAGEMAKER_MAGPIE_ENDPOINT_NAME"],
```

markbackman (approved):
LGTM! Just some tiny nits on the example.

PR #4465, opened 2026-05-12
Handle gpt-realtime-2 multi-output-item audio responses

kompfner:
## Summary

Properly handle new behavior observed with `gpt-realtime-2`:

- A single response from the API can now contain more than one audio item (we've seen two in practice).
- The `audio.done` event for the first item can arrive after deltas of the second item have started arriving.
- Deltas still arrive strictly in playback order (all of item A, then all of item B), so we keep queueing them for playback as soon as they arrive. This matches OpenAI's own reference implementation (`openai-agents-python`), which forwards deltas immediately with no `output_index`-based ordering.
- The new behavior required changes so that:
  - We don't fire spurious warnings on item-switch deltas.
  - Truncation always applies to the newest audio item (last-item-wins).
  - The whole response is bookended by a single `TTSStartedFrame` / `TTSStoppedFrame` pair (the Stopped is now pushed on `response.done`).

## Test plan

- [x] Unit tests in `tests/test_openai_realtime_audio_frames.py`:
  - Single-item happy path and single-item interrupt (regression guards).
  - Multi-item with overlapping `audio.done` (matches observed `gpt-realtime-2` behavior).
  - Multi-item with non-overlapping `audio.done` (theoretical future-proofing).
  - Interrupt during multi-item truncates against the latest item.
- [x] `uv run pytest tests/ -k openai` — 138 passed locally.
- [x] Verify in a live `gpt-realtime-2` session that the spurious warning no longer fires.


src/pipecat/services/openai/realtime/llm.py
```diff
@@ -752,20 +766,39 @@ async def _handle_evt_audio_delta(self, evt):
         # this event from the server
         await self.stop_ttfb_metrics()
 
-        if self._current_audio_response and self._current_audio_response.item_id != evt.item_id:
-            logger.warning(
-                f"Received a new audio delta for an already completed audio response before receiving the BotStoppedSpeakingFrame."
```
kompfner, line 757: @filipi87 wanted to understand what you were protecting against here before removing this...

Based on the wording of the warning, it sounds like you were protecting against the possibility that deltas were arriving *after* the "done" message for a particular audio item?

In my testing with gpt-realtime-2, this warning was getting triggered reasonably often for a different reason: we'd start getting deltas for a second audio item before the "done" message was received for the first item.

...is it possible that that's what you were seeing all along? I want to make sure that in this PR we're not regressing by removing a necessary protection.

changelog/4465.fixed.md
kompfner, line 1: Note that this PR doesn't yet bump the default model to `gpt-realtime-2`

src/pipecat/services/openai/realtime/llm.py
kompfner: ([Here's the PR](https://github.com/pipecat-ai/pipecat/pull/3581) where this protection introduced)

markbackman: This comment describes what we were seeing:
https://github.com/pipecat-ai/pipecat/pull/3567#discussion_r2729097098

src/pipecat/services/openai/realtime/llm.py
```diff
@@ -478,6 +480,18 @@ async def _handle_user_stopped_speaking(self, frame):
             await self.send_client_event(events.ResponseCreateEvent())
 
     async def _handle_bot_stopped_speaking(self):
+        if self._current_audio_response:
+            current = self._current_audio_response
+            elapsed_ms = int(time.time() * 1000) - current.start_time_ms
+            logger.debug(
```
markbackman: Is this logging for development only? Maybe set both to trace or remove?

src/pipecat/services/openai/realtime/llm.py
```diff
@@ -752,20 +766,39 @@ async def _handle_evt_audio_delta(self, evt):
...
-
         if not self._current_audio_response:
+            # First delta of a new assistant turn.
             self._current_audio_response = CurrentAudioResponse(
                 item_id=evt.item_id,
                 content_index=evt.content_index,
                 start_time_ms=int(time.time() * 1000),
+                response_id=evt.response_id,
+            )
+            logger.debug(
```
markbackman: Trace or remove before merging?

src/pipecat/services/openai/realtime/llm.py
markbackman: In reading your new change, perhaps the issue is that we had a bad assumption about how audio would be received. What you're proposing appears to work around the problem by expecting multiple audio chunks. That should eliminate this case, if I understand correctly.

markbackman (approved):
LGTM, if I understand the new case correctly. I'll have Claude look too. Probably worth getting a definitive answer from @filipi87 on the truncation race condition you asked about.

src/pipecat/services/openai/realtime/llm.py
kompfner: Yeah, agreed, I think we can remove this, it's not adding much value.

src/pipecat/services/openai/realtime/llm.py
kompfner: I think I'll remove this one (which happens all the time) but keep the one below (which logs the slightly rarer/more interesting case of multiple audio items per response)

src/pipecat/services/openai/realtime/llm.py
```diff
@@ -776,10 +809,14 @@ async def _handle_evt_audio_delta(self, evt):
         await self.push_frame(frame)
 
     async def _handle_evt_audio_done(self, evt):
-        if self._current_audio_response:
-            await self.push_frame(TTSStoppedFrame())
-            # Don't clear the self._current_audio_response here. We need to wait until we
-            # receive a BotStoppedSpeakingFrame from the output transport.
+        # Debug-trace only; TTSStoppedFrame is emitted on response.done so that
```
kompfner: I think we can probably get rid of this one, too

kompfner:
> LGTM, if I understand the new case correctly. I'll have Claude look too. Probably worth getting a definitive answer from @filipi87 on the truncation race condition you asked about.

Spoke to @filipi87 in person and it sounds like the thing he was protecting against was the same as what we're addressing in this PR: a new batch of audio item deltas arriving in the same response.

PR #4467, opened 2026-05-12
Fix metrics.ttfb and partial output on TTS/STT/LLM OpenTelemetry spans

aconchillo:
## Summary

Three related fixes to the OpenTelemetry tracing decorators in `service_decorators.py`. All driven by the same root pattern — span lifetimes were tied to per-call handler invocations rather than the actual operations they were meant to trace — and all confined to `service_decorators.py`. `tts_service.py`, `stt_service.py`, and every per-service file are unchanged.

### TTS (`@traced_tts`)

- Previously: span scoped to `run_tts()`, which for streaming TTS (Cartesia, Deepgram WS, ElevenLabs WS, etc.) returns as soon as the WS message is sent.
- Now: span lifetime is owned by the audio-context methods (`create_audio_context` / `append_to_audio_context` / `on_audio_context_completed` / `remove_audio_context` / `reset_active_audio_context`). `metrics.ttfb` is recorded on the first `TTSAudioRawFrame` (after `stop_ttfb_metrics` has produced a real value). The span closes naturally on `TTSStoppedFrame` / `on_audio_context_completed`, or with `tts.interrupted=true` on interruption.

### STT (`@traced_stt`)

- Previously: span scoped to the per-transcript handler call, so `metrics.ttfb` was read before `stop_ttfb_metrics` had run for the finalized frame. The span was also sometimes parented to the previous turn — `TurnTraceObserver` sets the current turn's context via an `asyncio.create_task`-scheduled handler (`_call_event_handler` with `sync=False`), which races with any span opened on a speaking-frame.
- Now: one STT span per finalized `TranscriptionFrame`, anchored at speech start (`VADUserStartedSpeakingFrame.timestamp - start_secs`) but lazy-opened on the first `TranscriptionFrame` so the turn context is reliably set by the time the span exists. Closes on `finalized=True` (success), on `stop_ttfb_metrics(end_time=...)` with `stt.timed_out=true` (TTFB-timeout path), or on `UserStoppedSpeakingFrame` with `stt.incomplete=true` (orphan).

### LLM (`@traced_llm`)

- Previously: the aggregated `output` attribute was attached to the span only after the wrapped LLM call returned cleanly. On mid-stream interruption (user starts speaking while LLM is generating), the call was cancelled and the accumulated text was discarded.
- Now: the `output` attribute is written in the `finally` block so partial output is preserved on cancellation.

### Mechanism

`@traced_tts` and `@traced_stt` use a `__set_name__` descriptor to install per-instance method wrappers on the owning class at class-definition time. `@traced_llm` keeps its per-call wrapper but moves attribute writes into `finally`.

## Testing

- `uv run pytest tests/test_tracing_context.py tests/test_turn_trace_observer.py tests/test_elevenlabs_tts.py tests/test_xai_tts.py tests/test_piper_tts.py tests/test_tts_frame_ordering.py`
- Manually verify a streaming TTS session: `metrics.ttfb` matches `stop_ttfb_metrics` log; interrupted synthesis ends with `tts.interrupted=true`.
- Manually verify an STT session in Jaeger: trace tree shows `Turn (STT, LLM, TTS)` with STT under the correct turn; multi-final services produce multiple STT spans per turn; spans with no finalized transcript carry `stt.incomplete=true`; timeout-only spans carry `stt.timed_out=true`.
- Manually verify an interrupted LLM generation: the LLM span has a partial `output` attribute reflecting whatever text the model emitted before the cancellation.

### DIFF
--- a/src/pipecat/utils/tracing/service_decorators.py
+++ b/src/pipecat/utils/tracing/service_decorators.py
@@ -889,10 +889,6 @@ def traced_llm(func: Callable | None = None, *, name: str | None = None) -> Call
                         fn_called = True
                         result = await f(self, context, *args, **kwargs)
 
-                        # Add aggregated output after function completes, if available
-                        if output_text:
-                            current_span.set_attribute("output", output_text)
-
                         return result
 
                     finally:
@@ -904,6 +900,15 @@ def traced_llm(func: Callable | None = None, *, name: str | None = None) -> Call
                             and original_start_llm_usage_metrics
                         ):
                             self.start_llm_usage_metrics = original_start_llm_usage_metrics
+
+                        # Attach whatever output text we accumulated so
+                        # far. Doing this in finally captures partial
+                        # output when ``f`` is cancelled or raises mid-
+                        # stream (e.g. interruption during LLM
+                        # generation), rather than only on clean
+                        # completion.
+                        if output_text:
+                            current_span.set_attribute("output", output_text)
 
                         # Update TTFB metric
                         ttfb: float | None = getattr(getattr(self, "_metrics", None), "ttfb", None)
--- a/src/pipecat/utils/tracing/service_decorators.py
+++ b/src/pipecat/utils/tracing/service_decorators.py
@@ -23,9 +23,16 @@ if TYPE_CHECKING:
     from opentelemetry import context as context_api
     from opentelemetry import trace
 
-from pipecat.frames.frames import MetricsFrame, TTSStoppedFrame
+from pipecat.frames.frames import (
+    MetricsFrame,
+    TranscriptionFrame,
+    TTSStoppedFrame,
+    UserStoppedSpeakingFrame,
+    VADUserStartedSpeakingFrame,
+)
 from pipecat.metrics.metrics import TTFBMetricsData
 from pipecat.processors.aggregators.llm_context import NOT_GIVEN
+from pipecat.processors.frame_processor import FrameDirection
 from pipecat.utils.tracing.service_attributes import (
     add_gemini_live_span_attributes,
     add_llm_span_attributes,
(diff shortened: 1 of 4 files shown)


src/pipecat/utils/tracing/service_decorators.py
```diff
@@ -190,102 +204,198 @@ def traced_tts(func: Callable | None = None, *, name: str | None = None) -> Call
...
-                        cartesia_version=getattr(self, "_cartesia_version", None),
-                        context_id=getattr(self, "_context_id", None),
-                    )
+            async def traced_append_to_audio_context(context_id, frame):
+                entry = service._tts_spans.get(context_id)
+                if entry and frame is not None:
+                    try:
+                        span = entry["span"]
+                        if isinstance(frame, TTSAudioRawFrame) and not entry["ttfb_recorded"]:
+                            ttfb = getattr(getattr(service, "_metrics", None), "ttfb", None)
```
markbackman: From my Codex review:

This still looks like it can record the pre-finalized TTFB value. `append_to_audio_context()` runs when the first `TTSAudioRawFrame` is queued, but the canonical TTFB metric is stopped later in `_handle_audio_context()` when that frame is consumed. While `_metrics.ttfb` is in progress it returns elapsed time so far, so this span attribute can be lower than the emitted `MetricsFrame`, especially if this context waits behind earlier queued audio.

Can we attach `metrics.ttfb` after `stop_ttfb_metrics()` runs, or use the generated metric value, so the span matches the existing TTFB metric?

aconchillo: Reading from the MetricsFrame just works.

aconchillo: Should be fixed now. It now just shows the same as the metrics frame.

markbackman (approved):
LGTM! Thanks for the clean up here 🙌

PR #4414, opened 2026-05-04
fix: TTSSpeakFrame greetings drop trailing words and skip on_assistant_turn_stopped

markbackman:
## Summary

- `TTSSpeakFrame(append_to_context=True)` greetings could split across two assistant messages in the LLM context and never surface in `on_assistant_turn_stopped` (regression since v0.0.105).
- Root cause 1: `LLMAssistantPushAggregationFrame` had no PTS, so it traveled through the transport's audio queue while word-level `TTSTextFrame`s travel through the clock queue and could overtake them, leaving trailing words orphaned in the aggregator buffer. Now stamped with `_word_last_pts + 1` when word timestamps are present.
- Root cause 2: `LLMAssistantAggregator`'s handler called `push_aggregation()` directly and bypassed `_trigger_assistant_turn_stopped()`. Since TTS-only flows never see `LLMFullResponseStartFrame`, the turn-start timestamp stayed empty and the event never fired. The handler now opens a turn (if needed) and triggers stopped, restoring the v0.0.104 behavior.

## Testing

- `uv run pytest tests/test_tts_frame_ordering.py tests/test_context_aggregators_universal.py`
- New `test_push_aggregation_pts_after_last_word` and `test_push_aggregation_no_pts_without_word_timestamps` cover the PTS routing.
- New `test_push_aggregation_fires_turn_stopped_for_tts_speak` and `test_push_aggregation_does_not_double_fire_in_llm_response` cover the event firing.

## Fixes

- Fixes #4264

PR #4468, opened 2026-05-12
fix: deliver already-spoken clock-queued frames before InterruptionFrame cancels the queue

### LINKED ISSUE #4466: Interruption drops already-spoken TTSTextFrames from the output transport's clock queue
### pipecat version

1.1.0

### Python version

3.13

### Operating System

Linux (Modal)

### Issue description

`MediaSender.handle_interruptions` calls `_cancel_clock_task()` (`base_output.py:558`, `:960-964`), which discards every `TTSTextFrame` still sitting in `_clock_queue` — including frames whose `pts <= transport.get_clock().get_time()`. By Pipecat's own PTS semantics, those frames represent words the caller has already heard, so dropping them is silent data loss.

The aggregator then sees nothing for the interrupted turn:

- `push_aggregation()` returns `""` (`llm_response_universal.py:995-996`)
- `on_assistant_turn_stopped.message.content == ""`
- `LLMContext` is not updated

Application code following [docs/guides/fundamentals/saving-transcripts](https://docs.pipecat.ai/guides/fundamentals/saving-transcripts) persists nothing for calls where the caller heard audible speech.

### Reproduction steps

1. Use ElevenLabs WS TTS (any service that emits per-word `TTSTextFrame`s through the clock queue reproduces).
2. `await task.queue_frames([TTSSpeakFrame(text=<long greeting>, append_to_context=True)])` on `on_client_connected`.
3. Subscribe to `on_assistant_turn_stopped`.
4. Inject an `InterruptionFrame` ~1s after `BotStartedSpeakingFrame`.

### Expected behavior

`message.content` contains words whose `pts` had already elapsed at interruption time.

### Actual behavior

`message.content == ""`. Every clock-queued word is dropped.

### Suggested fix

Drain already-elapsed frames from `_clock_queue` before cancelling the task:

```python
async def _drain_already_spoken_clock_frames(self):
    if not self._clock_queue:
        return
    current_time = self._transport.get_clock().get_time()
    while True:
        try:
            timestamp, _, frame = self._clock_queue.get_nowait()
        except asyncio.QueueEmpty:
            break
        if timestamp <= current_time:
            await self._transport.push_frame(frame)
        self._clock_queue.task_done()
```

Call from `handle_interruptions` before `_cancel_clock_task()`. Happy to put up a PR.

### Related

PR #4414 fixed the analogous drop on clean completion; the same invariant ("aggregator must see every committed word") is violated here on the interruption path. Adjacent races: #4308, #4111, #1361, #3986.

### Logs

N/A

nateschickler0:
## Summary

- Closes #4466.
- When `InterruptionFrame` arrives at `BaseOutputTransport`, `MediaSender.handle_interruptions` calls `_cancel_clock_task()` which discards every frame still in `_clock_queue` — including frames whose `pts <= transport.get_clock().get_time()` (words whose playout instant has already passed). For TTS services that emit per-word `TTSTextFrame`s (ElevenLabs WS, Cartesia, Rime), the assistant aggregator's `_aggregation` ends up missing words the caller has already heard, and `on_assistant_turn_stopped.message.content` is empty for the interrupted turn.
- Fix: in `process_frame`, before forwarding the `InterruptionFrame` downstream, drain any clock-queued frame whose PTS has already elapsed and push it through. Words whose PTS has not yet arrived stay dropped (consistent with the existing cancellation semantics).
- Mirrors the invariant established by #4414 on the clean-completion path: the assistant aggregator must see every word whose PTS has been committed, regardless of how the turn ends.

## Testing

- `uv run pytest tests/test_tts_frame_ordering.py` — 16/16 pass, including the two new tests:
  - `test_drain_already_spoken_clock_frames_delivers_past_pts_drops_future_pts` — regression case (past-PTS frames pushed in chronological order, future-PTS dropped).
  - `test_drain_already_spoken_clock_frames_noop_without_clock_task` — short-circuit when no clock task was created.
- `uv run pytest tests/test_context_aggregators_universal.py` — 57/57 pass.
- Full suite passes apart from 4 unrelated pre-existing failures (Anthropic-API integration test, deprecation-warning test, elevenlabs/xai service init).

## Fixes

- Fixes #4466


nateschickler0:
Closing this PR per author request.

PR #4474, opened 2026-05-12
Extend cancel_on_interruption=False to Inworld Realtime (best-effort + warning)

kompfner:
## Summary

- Extends the `cancel_on_interruption=False` regression fix to `InworldRealtimeLLMService`, using the same async-tool routing approach as #4441: detect async-tool messages in the LLM context, route the final result to the formal tool-result channel.
- **Caveat:** as of this writing, Inworld Realtime doesn't appear to handle the resulting delayed tool result reliably. The routing is best-effort, and the service emits a one-time warning when async-tool messages are seen.
- Streamed intermediate results (`FunctionCallResultProperties(is_final=False)`) remain unsupported.
- Adds function-calling demonstration to `examples/realtime/realtime-inworld.py`.

## Background

Inworld was excluded from the equivalent fix in #4447 because we'd hit a pre-existing tool-calling issue on the Inworld Realtime API. That turned out to be an account-level matter rather than a service-level one, so basic tool calling now works on Inworld Realtime — but `cancel_on_interruption=False` in particular still misbehaves at the upstream layer. The routing code is in place so that if/when Inworld resolves the upstream behavior, no further pipecat changes are required.

## Test plan

- [x] Run `examples/realtime/realtime-inworld.py` with `--transport webrtc` and confirm synchronous function calling works (ask for the weather).
- [x] Register a function with `cancel_on_interruption=False` against `InworldRealtimeLLMService` and verify the one-time warning fires the first time the async-tool envelope hits the context.
- [x] Confirm `tests/test_async_tool_messages.py` still passes.


filipi87 (approved):
LGTM!

PR #4475, opened 2026-05-12
Fix Cartesia CJK timestamp spacing

markbackman:
Note: there are two fixes in this PR:
1. Korean doesn't require special handling. Cartesia emits word chunks with timing info.
2. Chinese and Japanese had a regression where the inter frame spacing was causing problems.

## What changed
- Treat Cartesia Chinese and Japanese timestamp groups as space-preserving text so downstream aggregation does not insert artificial spaces between groups.
- Let Korean timestamp words use the normal path so word spacing and per-word timestamps are preserved.
- Add focused tests for Chinese, Japanese, and Korean timestamp normalization and reassembly.

## Validation
- uv run ruff format --check src/pipecat/services/cartesia/tts.py tests/test_cartesia_tts.py
- uv run ruff check src/pipecat/services/cartesia/tts.py tests/test_cartesia_tts.py
- uv run pytest tests/test_cartesia_tts.py

### DIFF
--- a/src/pipecat/services/cartesia/tts.py
+++ b/src/pipecat/services/cartesia/tts.py
@@ -419,30 +419,30 @@ class CartesiaTTSService(WebsocketTTSService):
         """Convenience method to create a speed tag."""
         return f'<speed ratio="{speed}" />'
 
-    def _is_cjk_language(self, language: str) -> bool:
-        """Check if the given language is CJK (Chinese, Japanese, Korean).
+    def _is_chinese_or_japanese_language(self, language: str) -> bool:
+        """Check if the given language is Chinese or Japanese.
 
         Args:
             language: The language code to check.
 
         Returns:
-            True if the language is Chinese, Japanese, or Korean.
+            True if the language is Chinese or Japanese.
         """
-        cjk_languages = {"zh", "ja", "ko"}
         base_lang = language.split("-")[0].lower()
-        return base_lang in cjk_languages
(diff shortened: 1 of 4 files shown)


src/pipecat/services/cartesia/tts.py
```diff
@@ -419,30 +419,30 @@ def SPEED_TAG(speed: float) -> str:
         """Convenience method to create a speed tag."""
         return f'<speed ratio="{speed}" />'
 
-    def _is_cjk_language(self, language: str) -> bool:
-        """Check if the given language is CJK (Chinese, Japanese, Korean).
+    def _is_chinese_or_japanese_language(self, language: str) -> bool:
+        """Check if the given language is Chinese or Japanese.
```
filipi87, line 423: Are we not handling the Korean language anymore? We removed it from all the comments here, but we kept it in the changelog. I believe we should fix one of those two places.

filipi87 (approved):
LGTM! Just one comment regarding the Korean language.

src/pipecat/services/cartesia/tts.py
markbackman: There were two bug fixes in this PR:
1. Korean doesn't require special handling. Cartesia emits word chunks with timing info.
2. Chinese and Japanese had a regression where the inter frame spacing was causing problems

Both are fixed. I added two changelog entries for the fixes.

PR #4477, opened 2026-05-13
NVidia TTS Sagemaker: Buffering audio to avoid glitches.

filipi87:
Recording audio in the nvidia sagemaker example.

In progress. Only for testing for now...


examples/voice/voice-nvidia-sagemaker.py
filipi87, line 1: TODO: remove this before merging. Only for testing.

markbackman (approved):
Buffering works well to avoid the initial audio blip that I'm hearing on the integration on main.

Approving—reminder to remove the changes to the demo before merging.

PR #4480, opened 2026-05-13
Skip user-transcript wait when local turn detection drives a realtime service

kompfner:
## Summary
- Adds `wait_for_transcript_to_end_user_turn` on `LLMUserAggregatorParams` for pipelines where local turn detection drives a realtime service like Gemini Live or OpenAI Realtime. Set it to False to avoid unnecessary latency from transcript delay — the realtime service consumes user audio directly, so we don't need user transcripts in context before it can respond.
- With this option:
  - Turn strategies do not consider user transcripts, so the user turn ends sooner. (Under the hood: the aggregator drops `TranscriptionUserTurnStartStrategy` from the start strategies and flips `wait_for_transcript=False` on any stop strategy that supports it. Strategy mutations are logged: loudly when the user passed their own strategies, quietly otherwise.)
  - User transcripts are then handled by the aggregator: a simple post-turn transcript wait gives them time to arrive after the user turn ends, and once the wait completes, the aggregator emits a new `on_user_turn_message_finalized` event with the new user context message.
- The post-turn transcript wait timer is sized to the STT service's reported P99 TTFS (`STTMetadataFrame.ttfs_p99_latency`), falling back to `DEFAULT_TTFS_P99` (1.0s) if no STT service has reported one. The simple-timer approach assumes user transcripts don't arrive too late and that the bot response won't finish before the timer.
- The new `on_user_turn_message_finalized` event also fires in the default mode (coinciding with `on_user_turn_stopped`), so consumers that want the populated user transcript can subscribe to it uniformly.
- The strategy-level `wait_for_transcript` parameter on `TurnAnalyzerUserTurnStopStrategy` and `SpeechTimeoutUserTurnStopStrategy` remains exposed for advanced cases.

## Realtime services covered
- **Gemini Live**: pair `wait_for_transcript_to_end_user_turn=False` with `GeminiVADParams(disabled=True)`. See `examples/realtime/realtime-gemini-live-local-vad.py`.
- **OpenAI Realtime**: pair `wait_for_transcript_to_end_user_turn=False` with `turn_detection=False` inside `OpenAIRealtimeLLMService.Settings.session_properties.audio.input`. The service already wires `UserStoppedSpeakingFrame` to `input_audio_buffer.commit` + `response.create` when `turn_detection=False`. See `examples/realtime/realtime-openai-local-vad.py`.

## Test plan
- [x] Gemini Live + local VAD: confirm `ActivityEnd` fires promptly after the local turn detection signals end of turn
- [x] OpenAI Realtime + local VAD: confirm `input_audio_buffer.commit` + `response.create` fire promptly after the local turn detection signals end of turn
- [x] Confirm the user message lands in `LLMContext` with the gathered transcript (via `on_user_turn_message_finalized`) for both services
- [x] Confirm conversational order in context: user message before assistant message
- [x] Run a cascaded STT pipeline: confirm existing transcript-wait behavior is unchanged; `on_user_turn_stopped` still carries content
- [x] `uv run pytest`


kompfner:
This PR is a zero-configuration (from the user's perspective) alternative to #4199.

kompfner:
> This PR is a zero-configuration (from the user's perspective) alternative to #4199.

Correction: this PR makes it so the user _stop_ strategy requires no configuration. The user may still need to configure the start strategy.

kompfner:
Hmm, I think we actually have potentially a serious problem with this approach (and with the alternative, #4199): with both approaches, we're intentionally decoupling user start/stop detection from transcripts, which is the right thing to do for sending `ActivityStart`/`ActivityEnd` (they shouldn't wait around for transcripts, or be triggered by transcript). 

Here's the fundamental problem: user start/stop detection *also drives how we flush transcripts to context*. If we don't wait for the user transcript to be captured by the aggregator before declaring the user's turn over, we'll mangle the user messages. So, as things are architected today, we *can't* decouple transcripts from user start/stop detection.

Put another way: with Gemini Live (and other realtime services), first the turns happen and then *later* the transcripts corresponding to those turns happen. This violates the cascade-model assumption that transcripts are inherently part of the turn.

Put yet another way: if we do anything (like this PR, or #4199) to solve the slowness problem with Gemini Live + local turn detection, then we'll necessarily create the other problem (mangled transcripts in context).

It's almost like what we need is:
* One mechanism for determining "realtime turns" - when to send Gemini Live `ActivityStart`/`ActivityEnd`
* Another separate mechanism for determining "transcript turns" - when to write transcripts to our context

Oof.

examples/realtime/realtime-gemini-live-local-vad.py
```diff
@@ -70,10 +71,25 @@ async def run_bot(transport: BaseTransport, runner_args: RunnerArguments):
             },
         ],
     )
+    # `wait_for_transcript_to_end_user_turn=False` is the right setting
```
kompfner, line 74: Here's how you use the new flag.

examples/realtime/realtime-gemini-live-local-vad.py
```diff
@@ -107,8 +123,23 @@ async def on_client_disconnected(transport, client):
         logger.info(f"Client disconnected")
         await task.cancel()
 
+    # `on_user_turn_stopped` fires at the end of the user turn. With
```
kompfner, line 126: And in the below two handlers you can see the effect of the flag.

kompfner:
After a discussion with @filipi87, @markbackman and @aconchillo, I'm planning on exploring a slightly different direction. Closing this PR for now.

PR #4481, opened 2026-05-13
SmallWebRTC: Preventing injecting silence while we are receiving audio from TTS.

filipi87:
Preventing injecting silence while we are receiving audio from TTS.

In progress. Just for testing.


filipi87:
Closing it for now.

PR #4482, opened 2026-05-13
Propagate Soniox token language

### LINKED ISSUE #4476: SonioxSTTService drops per-token language field from TranscriptionFrame
### pipecat version

1.1.0

### Python version

3.12.8

### Operating System

Mac

### Issue description

SonioxSTTService._receive_messages builds TranscriptionFrame(...) at [services/soniox/stt.py L575-583](https://github.com/pipecat-ai/pipecat/blob/main/src/pipecat/services/soniox/stt.py) without setting language=. Even with enable_language_identification=True in SonioxSTTService.Settings, the per-token language returned by Soniox is silently dropped on its way to downstream processors.

Soniox API contract (per [the WebSocket API docs](https://soniox.com/docs/api-reference/stt/websocket-api)):

Each token in the tokens array contains a language field (string, ISO code like "en", "nl")
Observed:

TranscriptionFrame.language is always None even when Soniox returns valid per-token language info
Downstream processors that route by detected language (TranslationProcessor, language-aware aggregators) can't function

Proposed fix (one-liner):
```
# In send_endpoint_transcript():
_langs = [t.get("language") for t in self._final_transcription_buffer if t.get("language")]
language = None
if _langs:
    from collections import Counter
    _lang_str = Counter(_langs).most_common(1)[0][0]
    try:
        language = Language(_lang_str)
    except ValueError:
        pass  # Soniox returned a code Pipecat's enum doesn't cover

await self.push_frame(
    TranscriptionFrame(
        text=text,
        user_id=self._user_id,
        timestamp=time_now_iso8601(),
        result=self._final_transcription_buffer,
        finalized=True,
        language=language,  # NEW
    )
)

```

### Reproduction steps

- Configure SonioxSTTService with multiple language_hints and enable_language_identification=True
- Speak in one of the configured languages
- Inspect the emitted TranscriptionFrame.language — always None

### Expected behavior

TranscriptionFrame.language should be set

### Actual behavior

TranscriptionFrame.language is not set

### Logs

```shell

```

markbackman:
## Summary

- Fixed `SonioxSTTService` final transcriptions so token-level language annotations from Soniox populate `TranscriptionFrame.language`.
- Passed the detected language through STT tracing alongside the final transcript.
- Enabled language identification in the Soniox voice example so users can see language metadata when Soniox returns it.

## Testing

- `uv run pytest tests/test_soniox_stt.py`
- `uv run ruff check src/pipecat/services/soniox/stt.py tests/test_soniox_stt.py`
- `uv run ruff format --check src/pipecat/services/soniox/stt.py tests/test_soniox_stt.py`

## Fixes

- Fixes #4476

### DIFF
--- a/changelog/4482.fixed.md
+++ b/changelog/4482.fixed.md
@@ -1,0 +1,1 @@
+- Fixed `SonioxSTTService` final transcription frames missing detected language metadata when Soniox returns token-level language annotations.
--- a/examples/voice/voice-soniox.py
+++ b/examples/voice/voice-soniox.py
@@ -58,6 +58,7 @@ async def run_bot(transport: BaseTransport, runner_args: RunnerArguments):
             # Add strict mode to enforce the language hints
             language_hints=[Language.EN],
             language_hints_strict=True,
+            enable_language_identification=True,
         ),
     )
 
--- a/src/pipecat/services/soniox/stt.py
+++ b/src/pipecat/services/soniox/stt.py
@@ -201,6 +201,19 @@ def _prepare_language_hints(
     return list(set(prepared_languages))
 
 
+def _language_from_tokens(tokens: list[dict]) -> Language | None:
+    for token in reversed(tokens):
+        language = token.get("language")
+        if not language:
+            continue
+        try:
+            return Language(language)
+        except ValueError:
+            pass
+
+    return None
+
+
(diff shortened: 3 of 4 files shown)


filipi87 (approved):
LGTM!

PR #4489, opened 2026-05-13
Expose LiveKit transport options and subscription failures

snejati86:
During testing with LiveKit and Pipecat, we saw SIP audio tracks publish successfully but Pipecat fail to subscribe with track not bound, so this adds LiveKitParams(single_peer_connection=True) as a possible mitigation and on_track_subscription_failed(participant_id, track_sid, error) as an API hook for detecting the failure and triggering recovery.

## Summary
- add LiveKitParams.single_peer_connection and forward it to rtc.RoomOptions
- register LiveKit track_subscription_failed events and expose an on_track_subscription_failed transport callback
- add LiveKit transport tests for both behaviors

## Tests
- uv run ruff format --check src/pipecat/transports/livekit/transport.py tests/test_livekit_transport.py
- uv run ruff check src/pipecat/transports/livekit/transport.py tests/test_livekit_transport.py
- uv run --extra livekit pytest tests/test_livekit_transport.py

### DIFF
--- a/changelog/4489.added.md
+++ b/changelog/4489.added.md
@@ -1,0 +1,1 @@
+- Added `single_peer_connection` configuration and track subscription failure callbacks to the LiveKit transport.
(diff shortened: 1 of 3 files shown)


markbackman:
This has been in draft for a while. Is this still in progress or should we close it?

snejati86:
Hi @markbackman , I marked it as draft to go review the code base guidelines first, then never got back to it. I will reopen it this week.
Thanks

PR #4491, opened 2026-05-13
fix(smallwebrtc): clear dead track ref on MediaStreamError

### LINKED ISSUE #4490: SmallWebRTCClient.read_audio_frame/read_video_frame loops forever on MediaStreamError
### pipecat version

1.1.0 (and `main` as of 2026-05-14)

### Python version

3.11

### Operating System

Linux (Docker, Debian slim)

### Issue description

`SmallWebRTCClient.read_audio_frame` (and the analogous `read_video_frame`) in `src/pipecat/transports/smallwebrtc/transport.py` catches `MediaStreamError` inside its `while True:` loop, logs a warning, sets `frame = None`, sleeps 10 ms, and **continues**. Once the underlying `aiortc` track is dead — for example after a WebRTC peer connection error mid-session — every subsequent `track.recv()` raises `MediaStreamError` again, so the async generator never terminates. The caller (`SmallWebRTCInputTransport._receive_audio` at ~line 654) is stuck inside `async for audio_frame in audio_iterator:` forever, and its task never finishes.

In practice this also pins one CPU core indefinitely via a downstream anyio `CancelScope` busy-loop (anyio#695-style, separate bug), because pipecat's pipeline cancellation can't actually reap the stuck receive task. We caught a single session pegging 100 %+ for 28 minutes until manual restart.

### Reproduction steps

Self-contained repro that doesn't require WebRTC plumbing (only standard-library `asyncio`):

\`\`\`python
import asyncio

class MediaStreamError(Exception): ...

class DeadTrack:
    async def recv(self):
        raise MediaStreamError("track ended")

async def read_audio_frame_buggy(track):
    """Pipecat 1.1.0 SmallWebRTCClient.read_audio_frame control flow."""
    while True:
        if track is None:
            await asyncio.sleep(0.01); continue
        try:
            frame = await asyncio.wait_for(track.recv(), timeout=2.0)
        except asyncio.TimeoutError:
            frame = None
        except MediaStreamError:
            frame = None   # ← bug: logs + continues
        if frame is None:
            await asyncio.sleep(0.01); continue
        yield frame

async def drive(factory):
    async def consume():
        async for _ in factory(): pass
    task = asyncio.create_task(consume())
    try:
        await asyncio.wait_for(asyncio.shield(task), timeout=1.0)
        print("exited")
    except asyncio.TimeoutError:
        print("STUCK after 1.0s (had to cancel)")
        task.cancel()
        try: await task
        except BaseException: pass

asyncio.run(drive(lambda: read_audio_frame_buggy(DeadTrack())))
\`\`\`

Output:

\`\`\`
STUCK after 1.0s (had to cancel)
\`\`\`

### Expected behavior

When the track raises `MediaStreamError`, the async generator should exit cleanly so the downstream `async for` returns and the receive task can finish. A dead track is unrecoverable — the next `recv()` will raise the same error.

### Actual behavior

Generator loops at ~100 Hz indefinitely on a permanently-failed track.

### Suggested fix

Replace `frame = None` with `break` in both `MediaStreamError` handlers (audio: ~line 381, video: ~line 327). Caller `_receive_audio` already handles iterator exit via `async for`.

Happy to send a PR with the fix + regression tests for both iterators.

bnovik0v:
Fixes #4490. Updated in response to @filipi87's feedback that a plain `break` would regress renegotiation.

## Summary

The async generators `SmallWebRTCClient.read_audio_frame` / `read_video_frame` in `src/pipecat/transports/smallwebrtc/transport.py` previously caught `MediaStreamError`, slept 10 ms, and re-entered `recv()` on the same track. Once the underlying aiortc track is permanently dead, every subsequent `recv()` raises the same error — so the loop busy-spins at ~100 Hz forever and the caller's `_receive_audio` task stays wedged on `async for audio_frame in audio_iterator:`. In production this pinned one CPU core for ~28 min via a downstream anyio `CancelScope` busy-loop.

## Fix

On `MediaStreamError`, clear the affected track reference (`_audio_input_track` / `_video_input_track` / `_screen_video_track`) and let the loop park on its existing `is None` sleep gate — the same gate `_handle_peer_disconnected` already uses. When `_handle_client_connected` reassigns a fresh track during renegotiation, the loop resumes from the new track. No new state machine.

This replaces an earlier revision of this PR that used `break`, which would have terminated the iterator on transient mid-renegotiation errors (#4490 review feedback).

## Changes

- `src/pipecat/transports/smallwebrtc/transport.py`: in both `MediaStreamError` handlers, null the relevant track ref instead of leaving it dangling. Video path picks `_video_input_track` vs `_screen_video_track` from `video_source`.
- `tests/test_smallwebrtc_transport.py`: 5 regression tests — park-on-dead-track and renegotiation-resumes for audio + camera-video, plus screen-share park.
- `changelog/4491.fixed.md`: updated to describe nulling approach.

## Test plan

- [x] `pytest tests/test_smallwebrtc_transport.py -v` → 5 passed.
- [x] Reverted the transport fix locally and re-ran → all 5 fail. In particular, `test_renegotiation_resumes` catches the regression the `break`-based fix would have caused.

## Verification table

Driving the real `SmallWebRTCClient.read_audio_frame` against fake aiortc tracks for 500 ms each:

| Scenario | `track.recv()` calls | Frames yielded |
|---|---|---|
| Current upstream (buggy), dead track | **49** (busy-spin at ~100 Hz) | 0 |
| This PR, dead track | **1** (parks on `is None`) | 0 |
| This PR, dead track swapped to fresh @100 ms | dead: 1, fresh: 7550 | 7549 |

### DIFF
--- a/changelog/4491.fixed.md
+++ b/changelog/4491.fixed.md
@@ -1,5 +1,7 @@
-- Fixed `SmallWebRTCClient.read_audio_frame` and `read_video_frame` looping
-  forever after `MediaStreamError`. On a permanently-failed track, the
-  generators now exit cleanly so the receive task can finish — previously they
-  caught the error, slept 10 ms, and retried `recv()` indefinitely, leaving
-  the receive task wedged.
+- Fixed `SmallWebRTCClient.read_audio_frame` and `read_video_frame` busy-looping
+  on `MediaStreamError`. When a track raises `MediaStreamError`, the generator
+  now clears the track reference (`_audio_input_track` / `_video_input_track` /
+  `_screen_video_track`) so the loop parks on the existing `is None` gate
+  instead of re-entering `recv()` at ~100 Hz on a permanently-failed track.
+  Renegotiation still resumes seamlessly: when `_handle_client_connected`
+  reassigns a fresh track, the loop picks up frames from the new track.
--- a/changelog/4491.fixed.md
+++ b/changelog/4491.fixed.md
@@ -1,0 +1,5 @@
+- Fixed `SmallWebRTCClient.read_audio_frame` and `read_video_frame` looping
+  forever after `MediaStreamError`. On a permanently-failed track, the
+  generators now exit cleanly so the receive task can finish — previously they
+  caught the error, slept 10 ms, and retried `recv()` indefinitely, leaving
+  the receive task wedged.
--- a/src/pipecat/transports/smallwebrtc/transport.py
+++ b/src/pipecat/transports/smallwebrtc/transport.py
@@ -15,7 +15,8 @@ import asyncio
 import fractions
 import time
 from collections import deque
-from typing import Any, Awaitable, Callable, List, Optional
+from collections.abc import Awaitable, Callable
+from typing import Any
 
 import numpy as np
 from loguru import logger
@@ -235,9 +236,9 @@ class SmallWebRTCClient:
 
         self._audio_output_track = None
         self._video_output_track = None
-        self._audio_input_track: Optional[AudioStreamTrack] = None
-        self._video_input_track: Optional[VideoStreamTrack] = None
-        self._screen_video_track: Optional[VideoStreamTrack] = None
+        self._audio_input_track: AudioStreamTrack | None = None
+        self._video_input_track: VideoStreamTrack | None = None
+        self._screen_video_track: VideoStreamTrack | None = None
 
         self._params = None
         self._audio_in_channels = None
@@ -314,7 +315,7 @@ class SmallWebRTCClient:
 
             try:
                 frame = await asyncio.wait_for(video_track.recv(), timeout=2.0)
-            except asyncio.TimeoutError:
+            except TimeoutError:
                 if (
                     self._webrtc_connection.is_connected()
                     and video_track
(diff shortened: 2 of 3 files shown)


bnovik0v:
Force-pushed: switched the fix from `break` to clearing the track reference, per @filipi87's renegotiation concern in #4490 ([comment](https://github.com/pipecat-ai/pipecat/issues/4490#issuecomment-4500158055)).

The new behaviour piggybacks on the existing `is None: sleep; continue` gate — no new state machine. Renegotiation resumes seamlessly when `_handle_client_connected` reassigns a fresh track; truly dead tracks park at ~0% CPU instead of ~100%. Tests now cover both the park path and the renegotiation path for audio, camera-video, and screen-video iterators; reverting the fix locally fails all 5.

filipi87:
Hi @bnovik0v, yes, I think this approach makes more sense. Could you just double-check why the tests are failing? After that, I believe we should be good to merge.

Also, make sure our lint formatting is correct. You can run:
`./scripts/fix-ruff-and-typecheck.sh`

Thank you for doing this.

bnovik0v:
Thanks @filipi87. The failing tests were a collection-time `ModuleNotFoundError`: the test module imported `aiortc` and `av` at top level, but the unit-test job doesn't install the `webrtc` extra. Guarded both with `pytest.importorskip` in `1906d4b42`. With the extra installed all 5 tests still pass.

Also re-ran `scripts/fix-ruff-and-typecheck.sh` against current `main`: ruff clean, pyright 0 errors on changed files.

filipi87 (approved):
LGTM! Thank you for fixing this @bnovik0v . 🚀

PR #4493, opened 2026-05-14
Port pipecat-subagents into pipecat core

aconchillo:
## Summary

Folds the standalone `pipecat-subagents` package into pipecat core as a first-class task model. Pipelines, tools, and inter-task RPC now share one `BaseTask` contract and one `TaskBus`.

- **`pipecat.registry`** — `TaskRegistry` for tracking local + remote tasks across runners.
- **`pipecat.bus`** — `TaskBus` (abstract) + `AsyncQueueBus` (in-process), the bus message hierarchy, `BusSubscriber`, plus `BusBridgeProcessor` and an internal `_BusEdgeProcessor` that wraps bridged pipelines.
- **Job RPC** in `pipecat.pipeline` — `JobContext` / `JobGroupContext`, the `JobGroup` types, and the `@job` / `@task_ready` decorators.
- **`BaseTask`** replaces the old abstract `BasePipelineTask`. It owns activation, end / cancel, job RPC, and a default `run()` for bus-only tasks.
- **`PipelineTask`** now extends `BaseTask`. Optional `bridged=` wraps the pipeline with bus edge processors so it can exchange frames with other bridged tasks.
- **`PipelineRunner`** owns the shared `TaskBus` and `TaskRegistry`. `spawn(task)` registers a fire-and-forget task; `run(task)` drives the main task through the same path.
- **`pipecat.tasks`** — `LLMTask` (LLM pipeline + `@tool` collection + tool-call deferral), `LLMContextTask` (LLM + `LLMContextAggregatorPair`), and the `@tool` decorator. Also ships `pipecat.tasks.proxy.websocket` client/server stubs that need a follow-up port.
- **Tighter contract** — `FrameProcessorSetup.pipeline_task` and `FunctionCallParams.pipeline_task` are now required; the `pipeline_task` property raises if accessed before setup. `BaseTask.attach(*, registry, bus)` replaces `set_registry()` and is the single hook the runner uses to hand the runner-provided context to each task.


aconchillo:
This is all the same code as Pipecat Subagents with a few main differences:

- Before: `BaseAgent` HAS A `PipelineTask` / After: `PipelineTask` IS A `BaseTask` (renamed from `BaseAgent`). So composition vs inheritance.
- A `PipelineTask` is now an "agent" and they are now all connected to the bus. So, all the existing examples are now "agent" capable.
- Previous agent tasks are called "jobs" and "job groups".
- Event `on_ready`, `on_error` don't exist anymore, because since we are now using `PipelineTask`s we have `on_pipeline_started` and `on_pipeline_error`.

The rest is mostly the same (main renames, like `@task` is now `@job` and `on_task_request` is `on_job_request`, etc.).

markbackman (approved):

filipi87 (commented):
Hi @aconchillo, I really like the idea of integrating subagents into core Pipecat. Bringing all the features from subagents, like delegation, orchestration, shared bus, and making existing examples “agent-capable”, is great.

I am just not completely sure about the current approach, because it feels like we may be collapsing a few abstractions that were previously very cleanly separated in both Pipecat and Subagents.

Today, Pipecat has a very natural conceptual model:
* pipelines and `PipelineTask` = realtime execution / connecting processors together
* `PipelineRunner` = lifecycle/runtime supervisor

And originally, Subagents introduced another layer, bringing orchestration, delegating tasks to other agents, and a way for different agents to communicate.

In the current PR, these concepts seem to get merged, like you mentioned, right now the concept is that "PipelineTask IS Agent"

Which makes things a little harder to distinguish conceptually, tasks, jobs, and agents all start looking very similar. It feels like we lose some of the layering/separation we had before.

I wonder if instead of: "**_PipelineTask IS an Agent_**", we could do something more like: "**_PipelineTask can participate AS an Agent_**".

Something like:

```python
assistant = task.as_agent("voice_assistant")
runner.add_agent(assistant)
```

Where:
* `PipelineTask` remains the realtime runtime primitive
* `Agent` remains the orchestration abstraction
* the runner supervises both tasks and agents

And the full example could look something like this:

```python
runner = PipelineRunner()

task = PipelineTask(...)

assistant = task.as_agent(
    name="voice_assistant"
)

runner.add_agent(assistant)

runner.add_agent(
    CodeWorkerAgent(
        name="code_worker",
        project_path=PROJECT_PATH,
    )
)

await runner.run()
```

This would still achieve the really nice goal of making all existing examples “agent-capable”, but without needing to make `PipelineTask` itself become the primary agent abstraction.

What do you think ? Would something like this make sense / be viable to implement ?

aconchillo:
> 
> What do you think ? Would something like this make sense / be viable to implement ?

This is exactly what we are trying to avoid. We don't need to think about another abstraction because there is none. Adding a new abstraction (even conceptual) means you need to know about it. For example, there shouldn't be any difference in the way you think about a single PipelineTask vs two PipelineTasks. That is, let's say you start with one PIpelineTask (as all our examples) but months later you want to add a second one. With your proposal, now you need to think that the first PipelineTask needs to participate as an agent (so we need this concept of an agent) and then we need to add a second PipelineTask which is also an agent and needs to participate as well. With the current proposal, that friction disappears, there's no notion of an agent or anything similar.  Pipecat just becomes a multi-pipeline framework instead of a single-pipeline framework. You don't need to think about any other layer except for the communication between pipelines. Pipelines are just participants of this new system. If you have a single pipeline, fine, everything keeps working, if you add more pipelines you just worry about communications (messages, jobs, etc.).

With Pipecat Subagents things were even worst in the sense that it was really forcing you to learn how the new agent abstraction worked (new methods you have to override). We made that decision because it was the only way to have something completely separate from Pipecat.

At the beginning I was hesitating as well, but @kwindla pushed for a more friction-less integration. I believe he is totally right. Initially, I was not sure how to get there until I had this realization about PipelineTasks being the actual "agents", which actually simplified quite a few things. Happy to keep discussing about it!

filipi87:
I am totally in favor of this frictionless integration. What I am still not sure about is whether we should keep overloading what `PipelineTask` is doing, or if creating this new layer, where we would hide all these new features we are bringing from subagents, like delegation, orchestration, and a shared bus, would make things easier to maintain and evolve.

I also wonder whether, for users, having this new layer would make the "multiple-agents" concept simpler or harder to understand.

But I know you have spent a lot more time than I have thinking about this. I was just sharing how my mental model has been working around this to see whether it makes sense or not. Because we could achieve a frictionless integration either way I believe.

elnygren:
We are using Pipecat Subagents in production and liking it! It helped us structure our code and separate some concerns with agent handovers. Absolutely in favor of bringing that to core. 

Interested to hear why call it "Task" (or "LLMTask") instead of "Agent" ("LLMAgent", "FlowAgent" etc) ? Even some of the examples are "two_**agents**" in filename but then it is two `AcmeLLMTask`s in the code. Is it just because vanilla Pipecat already has a "PipelineTask" and these "agents-that-we-call-tasks" override that class? Or is your thinking that these are not really "agents" but some lower level building blocks for agents?

Pipecat Subagents call these Agents and it _feels_ good to have a class `FooAgent(SomeBaseAgent)` that declares the LLM, LLM tools etc in the class. It's mentally a nice model to structure code and your thoughts. We are building agents that can handover to other agents. 

Might be a bad argument, but other players seem to call these agents, for example:
- https://docs.livekit.io/agents/logic/agents-handoffs/
- https://mastra.ai/docs/agents/overview

examples/multi-task/local-handoff/local-handoff-two-agents.py
```diff
@@ -0,0 +1,242 @@
...
+        audio_out_enabled=True,
+    ),
+    "webrtc": lambda: TransportParams(
+        audio_in_enabled=True,
+        audio_out_enabled=True,
+    ),
+}
+
+
+class AcmeLLMTask(LLMTask):
```
markbackman: I've thought more about the naming and do think that `BaseTask` is right. But, in our examples, especially this one, we might want to lean into "agent".
```suggestion
class AcmeLLMAgent(LLMTask):
```

examples/multi-worker/local-handoff/local-handoff-two-agents.py
```diff
@@ -0,0 +1,242 @@
...
+        """
+        logger.info(f"Task '{self.name}': ending conversation ({reason})")
+        await self.end(
+            reason=reason,
+            messages=[{"role": "developer", "content": reason}],
+            result_callback=params.result_callback,
+        )
+
+
+def build_greeter() -> AcmeLLMTask:
```
markbackman, line 112: Similarly, return an `AcmeLLMAgent` here.

```suggestion
def build_greeter_agent() -> AcmeLLMAgent:
```

examples/multi-worker/local-handoff/local-handoff-two-agents.py
```diff
@@ -0,0 +1,242 @@
...
+                "Do not answer product questions yourself. If the user says goodbye, "
+                "call the end_conversation tool. Do not mention transferring — just do it "
+                "seamlessly. Keep responses brief — this is a voice conversation."
+            ),
+        ),
+    )
+    return AcmeLLMTask("greeter", llm=llm, bridged=())
+
+
+def build_support() -> AcmeLLMTask:
```
markbackman, line 132: And here:
```suggestion
def build_support_agent() -> AcmeLLMAgent:
```

markbackman (commented):
Some quick thoughts on naming.

examples/multi-worker/sensor-controller/sensor-controller.py
```diff
@@ -0,0 +1,334 @@
+#
```
markbackman, line 1: Nice example! 🚀

src/pipecat/bus/adapters/__init__.py
```diff
@@ -0,0 +1,21 @@
+#
```
markbackman, line 1: We haven't defined module APIs like this in other places. We've opted to use full paths. Just highlighting this as a change, which can be fine, but we should standardize on how we do this.

pyproject.toml
```diff
@@ -119,7 +121,7 @@ tavus = [ "pipecat-ai[daily]" ]
 together = []
 tracing = [ "opentelemetry-sdk>=1.33.0,<2", "opentelemetry-api>=1.33.0,<2", "opentelemetry-instrumentation>=0.54b0,<1" ]
 ultravox = [ "pipecat-ai[websockets-base]" ]
-vonage-video-connector = [ "vonage-video-connector~=0.2.3b0; python_full_version>='3.13' and python_full_version<'3.14' and platform_system=='Linux'" ]
+vonage-video-connector = [ "vonage-video-connector>=1.0.0,<2; python_version >= '3.13'" ]
```
markbackman, line 124: I'd recommend reverting this line. In discussing with the Vonage engineer yesterday, the `0.2.3b0` version is needed. He's going to push changes for the 1.0.0 package later.

pyproject.toml
```diff
@@ -99,8 +99,10 @@ openai = [ "pipecat-ai[websockets-base]" ]
 rnnoise = [ "pyrnnoise~=0.4.1" ]
 openrouter = []
 perplexity = []
+pgmq = ["pgmq>=1.0.6,<1.1", "asyncpg>=0.30.0"]
 piper = [ "piper-tts>=1.3.0,<2", "requests>=2.32.5,<3" ]
 qwen = []
+redis = ["redis[hiredis]>=5.0.0"]
```
markbackman, line 105: Current redis version is 7.4.0. Maybe this should be:
```suggestion
redis = ["redis[hiredis]>=7,<8"]
```

pyproject.toml
```diff
@@ -99,8 +99,10 @@ openai = [ "pipecat-ai[websockets-base]" ]
 rnnoise = [ "pyrnnoise~=0.4.1" ]
 openrouter = []
 perplexity = []
+pgmq = ["pgmq>=1.0.6,<1.1", "asyncpg>=0.30.0"]
```
markbackman, line 102: Do we need the `asyncpg` pin?

Do we also need the tight version constaint on pgmq?

changelog/4493.added.md
```diff
@@ -0,0 +1 @@
+- Added `pipecat.tasks`, a task-based agent framework folded in from the standalone `pipecat-subagents` package. Tasks inherit from `BaseTask`, share a `TaskBus`, register in a `TaskRegistry`, and exchange typed work via `@job` handlers. `LLMTask` and `LLMContextTask` provide ready-made LLM-driven tasks. `PipelineRunner.spawn(task)` registers fire-and-forget tasks alongside the main pipeline task.
```
kompfner: IIRC, you had wanted to rename `spawn`

changelog/4493.changed.md
```diff
@@ -0,0 +1 @@
+- `PipelineTask` now inherits from `BaseTask`, so every pipeline task is also a bus participant. It accepts a new optional `bridged=()` parameter that auto-wraps the pipeline with bus edge processors, letting the task exchange frames with other bridged tasks over the shared `TaskBus`. The bus is supplied by `PipelineRunner` via `task.attach(registry=..., bus=...)` instead of through the constructor.
```
kompfner: This framing of it—it simply being a bus participant now—is the angle that I find particularly persuasive when I think about why we don't need another level of abstraction, @filipi87.

If you think of it as a physical device, the pipeline task is exactly the same as before. The only thing they changed in the new model is they put in a new port that you can use to connect to other pipeline tasks!

examples/multi-worker/code-assistant/code-assistant.py
```diff
@@ -0,0 +1,179 @@
...
+from pipecat.services.cartesia.tts import CartesiaTTSService
+from pipecat.services.deepgram.stt import DeepgramSTTService
+from pipecat.services.llm_service import FunctionCallParams
+from pipecat.services.openai.llm import OpenAILLMService
+from pipecat.transports.base_transport import BaseTransport, TransportParams
+from pipecat.transports.daily.transport import DailyParams
+
+load_dotenv(override=True)
+
+PROJECT_PATH = os.getenv("PROJECT_PATH", os.getcwd())
```
kompfner, line 56: Maybe this example warrants a README explaining how to use it?

And/or it could be a bit more self-documenting by logging something when you haven't set PROJECT_PATH.

examples/multi-worker/code-assistant/code-assistant.py
```diff
@@ -0,0 +1,179 @@
...
+    explore the project by reading files, searching code, and running
+    commands. It remembers previous questions for follow-ups.
+
+    Args:
+        question (str): The question about code, files, structure,
+            dependencies, or anything in the project.
+    """
+    logger.info(f"Asking code worker: '{question}'")
+    async with params.pipeline_task.job("code_worker", payload={"question": question}) as job:
+        await params.llm.queue_frame(
```
kompfner, line 81: Wanted to jot down a thought unrelated to subagents, but is relevant to this line of code. It's about our `cancel_on_interruption=False` mechanism. It's been a recurring thought for me for a little while.

Here you added a `LLMMessagesAppendFrame` to give the user audible feedback while the function is running.

Because of how `cancel_on_interruption=False` works under the hood, shouldn't we already be able to trigger a meaningful assistant response without adding anything new to the context?

Taking a step back, `cancel_on_interruption=False` can be used for either cases where:
- Functions that *can* take a little while to run, but not so long that we expect the user to wonder "what's going on?"; we just want to ensure that if the user does happen to say something in the meantime, the function isn't stopped
- Functions that are _known_ to take a long time up front, where we probably want something like a "stay tuned while I work on this" message

For the latter case (like the case in this file)...would it make sense to have a flag to supplement `cancel_on_interruption=False` that tells it to respond to the user when the function starts? Something like `speak_on_function_start` (placeholder name, not meant as a real suggestion).

src/pipecat/bus/bus.py
```diff
@@ -0,0 +1,184 @@
...
+            subscriber: The `BusSubscriber` to register.
+        """
+        sub = BusSubscription(subscriber=subscriber)
+        if self._running:
+            self._start_dispatch_task(sub)
+            # Schedule task right away.
+            await asyncio.sleep(0)
+        if subscriber.name in self._subscriptions:
+            raise ValueError(f"Subscriber '{subscriber.name}' is already registered on the bus")
+        self._subscriptions[subscriber.name] = sub
```
markbackman, line 107: **Bug: the duplicate-subscriber check runs *after* the dispatch tasks are started, leaking them.**

When the bus is already running, `_start_dispatch_task(sub)` creates the router and data dispatch tasks before the duplicate-name check raises. On the `raise` path `sub` is never stored in `self._subscriptions`, so neither `stop()` nor `unsubscribe()` (both only iterate `self._subscriptions`) can ever cancel those two tasks — they leak and block forever on their queues. Run the check first:

```suggestion
        if subscriber.name in self._subscriptions:
            raise ValueError(f"Subscriber '{subscriber.name}' is already registered on the bus")
        sub = BusSubscription(subscriber=subscriber)
        if self._running:
            self._start_dispatch_task(sub)
            # Schedule task right away.
            await asyncio.sleep(0)
        self._subscriptions[subscriber.name] = sub
```

src/pipecat/pipeline/base_worker.py
```diff
@@ -4,98 +4,1349 @@
...
+            The created ``JobGroup``.
+
+        Raises:
+            JobGroupError: If tasks are not ready within the timeout.
+        """
+        all_ready = await self._wait_tasks_ready(task_names)
+        try:
+            await asyncio.wait_for(all_ready, timeout=timeout)
+        except TimeoutError:
+            raise JobGroupError("tasks not ready within timeout")
```
markbackman: **Bug: watch handlers registered by `_wait_tasks_ready` are never removed, so each timed-out job request leaks one.**

For each task name, `_wait_tasks_ready` appends an `_on_ready` handler via `await self._registry.watch(name, _on_ready)` ([base_task.py#L1297-L1307](https://github.com/pipecat-ai/pipecat/blob/0cdb7d47ac6ec0cd18e0a064798e50bff0991b47/src/pipecat/pipeline/base_task.py#L1297-L1307)). When this `asyncio.wait_for(...)` times out and raises `JobGroupError`, those handlers are never removed: `TaskRegistry.watch` only ever appends to `self._watches[task_name]` ([registry.py#L88-L92](https://github.com/pipecat-ai/pipecat/blob/0cdb7d47ac6ec0cd18e0a064798e50bff0991b47/src/pipecat/registry/registry.py#L88-L92)) and there is no unwatch path, nor are handlers one-shot. Every `request_job`/`request_job_group` call that supplies a `timeout` for a task that is slow or never registers permanently accumulates a handler that also re-fires on every future registration of that name.

Consider adding an `unwatch(task_name, handler)` to `TaskRegistry` and wrapping the wait in `try/finally` so the handlers are removed on both the timeout and success paths.

src/pipecat/bus/serializers/json.py
```diff
@@ -0,0 +1,212 @@
...
+    Returns:
+        The resolved class, or None if it cannot be found.
+    """
+    module_path, _, class_name = qualified_name.rpartition(".")
+    if not module_path:
+        return None
+    try:
+        module = importlib.import_module(module_path)
+        return getattr(module, class_name, None)
+    except ImportError:
```
markbackman, line 211: **Security: `_resolve_type` imports arbitrary modules named in the wire payload, with no allowlist.**

`_deserialize_typed` passes the wire-supplied `__type__` straight to `_resolve_type`, which calls `importlib.import_module()` + `getattr` on any dotted path. This runs on raw network bytes — before any target/authorization filtering — in `RedisBus._reader_loop`, `PgmqBus._reader_loop`, and the WebSocket proxy receive loops. A peer that can publish to the channel/queue/socket can force the receiver to import any importable module (running its top-level import side effects — a DoS/side-effect vector) and to construct arbitrary dataclasses/`BaseModel`s/`Enum`s with attacker-chosen fields.

Construction is bounded to dataclass/BaseModel/Enum/registered adapters (no arbitrary-callable invocation), so this is not a clean RCE — but it is a real deserialization-trust gap. Consider gating `_resolve_type` behind an explicit type registry / module allowlist. At minimum, the trusted-peer assumption should be documented for the Redis and WebSocket transports (only the pgmq backend currently states it).

src/pipecat/workers/llm/llm_worker.py
```diff
@@ -0,0 +1,342 @@
...
+    """Probe frame used to flush all in-flight frames from the pipeline."""
+
+    pass
+
+
+@dataclass
+class LLMTaskActivationArgs(TaskActivationArgs):
+    """Activation arguments for LLM tasks.
+
+    Attributes:
```
markbackman, line 50: [AGENTS.md#L106-L108](https://github.com/pipecat-ai/pipecat/blob/0cdb7d47ac6ec0cd18e0a064798e50bff0991b47/AGENTS.md#L106-L108) requires dataclass docstrings to use a `Parameters:` section; this `@dataclass` uses `Attributes:` (the parent `TaskActivationArgs` already uses `Parameters:`).

```suggestion
    Parameters:
```

src/pipecat/services/llm_service.py
```diff
@@ -126,6 +131,7 @@ class FunctionCallParams:
     # treat it invariantly, rejecting `LLMService[XAdapter]` at the call
     # sites that build FunctionCallParams.
     llm: LLMService[Any]
+    pipeline_task: PipelineTask
```
markbackman: The new `pipeline_task` field isn't documented in the `Parameters:` section above ([L109-L122](https://github.com/pipecat-ai/pipecat/blob/0cdb7d47ac6ec0cd18e0a064798e50bff0991b47/src/pipecat/services/llm_service.py#L109-L122) lists every other field). [AGENTS.md#L106-L108](https://github.com/pipecat-ai/pipecat/blob/0cdb7d47ac6ec0cd18e0a064798e50bff0991b47/AGENTS.md#L106-L108) requires dataclass docstrings to use a `Parameters:` section covering the fields. Add an entry, e.g. between `llm:` and `context:`:

```
        llm: The LLMService instance being used.
        pipeline_task: The PipelineTask running the function call.
        context: The LLM context.
```

src/pipecat/pipeline/runner.py
```diff
@@ -4,36 +4,107 @@
...
+from pipecat.registry.types import TaskReadyData, TaskRegistryEntry
+from pipecat.utils.asyncio.task_manager import TaskManager, TaskManagerParams
 from pipecat.utils.base_object import BaseObject
 
 
-class PipelineRunner(BaseObject):
-    """Manages the execution of pipeline tasks with lifecycle and signal handling.
+@dataclass
+class _TaskEntry:
+    """A task registered on the runner and its background asyncio task."""
```
markbackman: This `@dataclass` documents its purpose but omits the `Parameters:` section required by [AGENTS.md#L106-L108](https://github.com/pipecat-ai/pipecat/blob/0cdb7d47ac6ec0cd18e0a064798e50bff0991b47/AGENTS.md#L106-L108) (and used by the other dataclasses added in this PR, e.g. `BusSubscription`):

```suggestion
    """A task registered on the runner and its background asyncio task.

    Parameters:
        task: The registered task.
        runner_task: The background asyncio task driving it, if any.
    """
```

src/pipecat/bus/network/pgmq_backends.py
```diff
@@ -0,0 +1,347 @@
...
+    return safe[:_MAX_CHANNEL_LEN] or "pipecat_bus"
+
+
+@dataclass
+class BackendMessage:
+    """A message returned by a backend's ``read`` call.
+
+    Backends normalize whatever shape they get off the wire into this
+    minimal record so the bus orchestrator can stay backend-agnostic.
+    """
```
markbackman, line 79: This `@dataclass` has fields (`msg_id`, `message`) but no `Parameters:` section, required by [AGENTS.md#L106-L108](https://github.com/pipecat-ai/pipecat/blob/0cdb7d47ac6ec0cd18e0a064798e50bff0991b47/AGENTS.md#L106-L108):

```suggestion
    """A message returned by a backend's ``read`` call.

    Backends normalize whatever shape they get off the wire into this
    minimal record so the bus orchestrator can stay backend-agnostic.

    Parameters:
        msg_id: Backend-specific id used to acknowledge/archive the message.
        message: The decoded message payload.
    """
```

markbackman (commented):
## Code review

Reviewed for bugs and CLAUDE.md (AGENTS.md) compliance. Flagged 3 correctness/security issues and 5 docstring-convention violations as inline comments below.

src/pipecat/bus/network/pgmq_backends.py
```diff
@@ -0,0 +1,347 @@
...
+    The provided ``PGMQueue`` must already be initialized (``await pgmq.init()``).
+    The backend does not own the client's lifetime.
+
+    Use this when bus peers trust each other. The channel name appears in
+    queue names visible to any role that can read ``pg_class``, so channels
+    are not secret.
+    """
+
+    def __init__(self, pgmq: PGMQueue):
+        """Initialize the backend with an already-initialized PGMQueue client."""
```
markbackman, line 157: [AGENTS.md#L106-L108](https://github.com/pipecat-ai/pipecat/blob/0cdb7d47ac6ec0cd18e0a064798e50bff0991b47/AGENTS.md#L106-L108) requires `__init__` to have an `Args:` section; the sibling `IsolatedPgmqBackend.__init__` already has one. Document the `pgmq` parameter:

```suggestion
        """Initialize the backend with an already-initialized PGMQueue client.

        Args:
            pgmq: An initialized ``PGMQueue`` client.
        """
```

filipi87 (approved):
Amazing work @aconchillo . 👏🚀

jptaylor (approved):
Great work!

PR #4495, opened 2026-05-14
Use majority language for Soniox transcripts

markbackman:
## Summary

- Select the most frequent recognized token language for Soniox final transcription frames.
- Ignore missing or unsupported Soniox token language values.
- Update Soniox tests for majority-language, tie, and final frame behavior.

## Testing

- uv run pytest tests/test_soniox_stt.py
- uv run ruff check src/pipecat/services/soniox/stt.py tests/test_soniox_stt.py
- uv run ruff format --check src/pipecat/services/soniox/stt.py tests/test_soniox_stt.py


filipi87 (approved):
LGTM!

Do we need a changelog for this ?

PR #4497, opened 2026-05-14
Fall back to _turn_context_id in get_active_audio_context_id

aconchillo:
## Summary

- Fixed dropped audio in streaming TTS services whose wire protocol doesn't echo `context_id` back on incoming audio (Sarvam, Smallest, Soniox, Inworld, …). `TTSService.get_active_audio_context_id()` now falls back to the synthesis-side `_turn_context_id` when the playback cursor isn't set yet, covering the race between context creation and the playback loop popping it off the serialization queue.
- Moved `model=` into `QwenLLMService.Settings(...)` in the qwen function-calling example to match the deprecation in the constructor.

## Testing

- Run a Sarvam-backed bot and verify `unable to append audio to context: no context ID provided` warnings no longer appear at the start of a turn or between rapid turns.


filipi87 (approved):
LGTM!

PR #4499, opened 2026-05-15
fix(elevenlabs): send voice_settings only on the first context init per WS

niklas-pernter:
## Problem

After upgrading to pipecat 1.2.0, every TTS turn after the first one in a session goes silent and ElevenLabs closes the WebSocket with code **1008**:

> `voice_settings field must be provided in the first message and then either be not provided or not change.`

Reproduced consistently across multiple production sessions: turn 1 plays audio fine, turn 2+ TTFB completes (~0.8s) and then the WS dies before any audio reaches the client.

## Root cause

PR #4433 (ElevenLabs ``close_context`` per turn) made each LLM turn open a brand-new audio context, so ``audio_context_available(context_id)`` is now always ``False`` on turn N+1 and the ``run_tts`` "create new context" branch fires every turn:

https://github.com/pipecat-ai/pipecat/blob/v1.2.0/src/pipecat/services/elevenlabs/tts.py#L974-L983

```python
msg: dict[str, Any] = {"text": " ", "context_id": context_id}
if self._voice_settings:
    msg["voice_settings"] = self._voice_settings
```

That re-sends ``voice_settings`` on every context-init. ElevenLabs' multi-context WebSocket protocol only allows ``voice_settings`` on the **first** message of the connection — see the official reference snippet at https://elevenlabs.io/docs/eleven-api/guides/how-to/websockets/multi-context-web-socket:

```js
function sendTextInContext(websocket, text, contextId, voiceSettings = null) {
  const message = { text, context_id: contextId };
  // Only include voice_settings for the first message in a context
  if (voiceSettings) message.voice_settings = voiceSettings;
  websocket.send(JSON.stringify(message));
}
```

## Fix

Track whether ``voice_settings`` has been sent on the current WebSocket. The flag is reset on every fresh ``_connect_websocket`` so reconnects re-send. Subsequent context-init messages on the same connection now omit the field.

## Tests

Two new regression tests in ``tests/test_elevenlabs_tts.py`` drive ``run_tts`` against a fake WebSocket:

- ``test_voice_settings_only_in_first_context_init`` — three back-to-back ``run_tts`` calls; only the first init carries ``voice_settings``.
- ``test_voice_settings_resent_after_ws_reconnect`` — after replacing the WS and resetting the flag, the next init carries ``voice_settings`` again.

Full ``tests/test_elevenlabs_tts.py`` suite is green (10 passed).


niklas-pernter:
Closing again. I just patched it myself

lencshu:
Thanks for the PR. I ran into the same issue as well. Could we reopen this and consider merging it?

markbackman:
@lencshu I can't repro this issue. The error is:
```
voice_settings field must be provided in the first message and then either be not provided or not change.
```

If `voice_settings` are initially set then not changed then the rule isn't violated. And, if there is a change, Pipecat disconnects/reconnects the websocket to abide by the rules. The latter is tested in this example:
https://github.com/pipecat-ai/pipecat/blob/main/examples/update-settings/tts/tts-elevenlabs.py

I tried by initializing to:
```
tts = ElevenLabsTTSService(
    api_key=os.environ["ELEVENLABS_API_KEY"],
    settings=ElevenLabsTTSService.Settings(
        voice=os.getenv(
            "ELEVENLABS_VOICE_ID",
            "Xb7hH8MSUJpSbSDYk0k2",
        ),
        speed=1.2,
        stability=0.5,
        similarity_boost=0.5,
        style=0.5,
        use_speaker_boost=True,
    ),
)
```

Then changing by pushing:
```
await task.queue_frame(
    TTSUpdateSettingsFrame(delta=ElevenLabsTTSService.Settings(speed=0.7, stability=0.6, similarity_boost=0.6, style=0.6, use_speaker_boost=False))
)
```

This works correctly.

It's possible that we're not testing this in the way you're using the API, so please let us know how to repro.

PR #4500, opened 2026-05-15
Update Gradium STT/TTS endpoints to region-neutral URLs

markbackman:
## Summary
- Switch the default WebSocket endpoints for `GradiumSTTService` and `GradiumTTSService` from the EU-specific URLs (`wss://eu.api.gradium.ai/...`) to the region-neutral `wss://api.gradium.ai/...`.
- Drop the now-redundant explicit endpoint overrides from the Gradium examples so they exercise the new defaults.

## Test plan
- [x] Run the updated Gradium STT example end-to-end against a live API key.
- [x] Run the updated Gradium TTS example end-to-end against a live API key.
- [x] Confirm the voice and transcription examples still connect and stream successfully.


filipi87 (approved):
LGTM!

PR #4501, opened 2026-05-15
Fix filter-incomplete + function-calling deadlock

aconchillo:
## Summary

- Fixed a hang when `filter_incomplete_user_turns` is enabled and the LLM responds to a user turn with a tool call. Before this PR the user turn never finalized, the assistant aggregator gated the tool-result context push on `_user_speaking`, and the LLM continuation never ran.
- Broadcasting `UserTurnInferenceCompletedFrame` now happens on `FunctionCallsStartedFrame` (idempotent with the existing ✓-marker path), so the turn finalizes the moment the LLM commits to a tool call — before the function dispatches.
- Added `turn-management-filter-incomplete-turns-function-calling.py` example to exercise the path, and wired it into the release eval suite.
- Drive-by: moved `OpenAITTSService` `instructions` into `Settings` in the openai function-calling example to match the canonical Settings API.

## Testing

- `uv run examples/turn-management/turn-management-filter-incomplete-turns-function-calling.py` and ask "what's the weather in Paris?" — the bot should call the tool and speak the result instead of hanging.

### DIFF
--- a/examples/turn-management/turn-management-filter-incomplete-turns-function-calling.py
+++ b/examples/turn-management/turn-management-filter-incomplete-turns-function-calling.py
@@ -1,0 +1,201 @@
+#
+# Copyright (c) 2024-2026, Daily
+#
+# SPDX-License-Identifier: BSD 2-Clause License
+#
+
+"""Example 22: Filter Incomplete Turns
+
+Demonstrates LLM-based turn completion detection to suppress bot responses when
+the user was cut off mid-thought. The LLM outputs one of three markers:
+- ✓ (complete): User finished their thought, respond normally
+- ○ (incomplete short): User was cut off, wait ~5s then prompt
+- ◐ (incomplete long): User needs time to think, wait ~10s then prompt
+
+When incomplete is detected, the bot's response is suppressed. After the timeout
+expires, the LLM is automatically prompted to re-engage the user.
+"""
+
+import os
+
+from dotenv import load_dotenv
+from loguru import logger
+
+from pipecat.adapters.schemas.tools_schema import ToolsSchema
+from pipecat.audio.vad.silero import SileroVADAnalyzer
+from pipecat.frames.frames import LLMRunFrame
+from pipecat.pipeline.pipeline import Pipeline
+from pipecat.pipeline.runner import PipelineRunner
+from pipecat.pipeline.task import PipelineParams, PipelineTask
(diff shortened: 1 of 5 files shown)


examples/turn-management/turn-management-filter-incomplete-turns-function-calling.py
```diff
@@ -0,0 +1,201 @@
+#
```
markbackman, line 1: Should we just add function calling to the current turn-management-filter-incomplete-turns.py example? That will eliminate a near duplicate example.

scripts/evals/run-release-evals.py
```diff
@@ -242,6 +242,7 @@ def EVAL_VISION_IMAGE(*, eval_speaks_first: bool = False):
 
 TESTS_TURN_MANAGEMENT = [
     ("turn-management/turn-management-filter-incomplete-turns.py", EVAL_COMPLETE_TURN),
+    ("turn-management/turn-management-filter-incomplete-turns-function-calling.py", EVAL_WEATHER),
```
markbackman, line 245: Any reason to keep these two?

markbackman (commented):
Works well. My only questions are:
1. Do we need the new example?
2. Is there a race condition between the `UserStoppedSpeakingFrame` and `FunctionCallResultFrame`? I guess it's unlikely to be realized since the `UserStoppedSpeakingFrame` will be much faster than the function call result being returned.

Having Claude look too.

markbackman:
## Code review

scripts/evals/run-release-evals.py
aconchillo: One will trigger with the complete marker, the function call does not I believe.

aconchillo:
> Works well. My only questions are:
> 
> 1. Do we need the new example?
> 2. Is there a race condition between the `UserStoppedSpeakingFrame` and `FunctionCallResultFrame`? I guess it's unlikely to be realized since the `UserStoppedSpeakingFrame` will be much faster than the function call result being returned.
> 
> Having Claude look too.

That why we broadcast `UserTurnInferenceCompletedFrame` just when the `FunctionCallsStartFrame` is pushed. Before, the final fixed I tried pushing in the turn reset. but that might be too late. I also thought about the race condition. This is the best we can do. I think it's fine.

markbackman (approved):
👍

PR #4504, opened 2026-05-16
feat(dtmf): add opt-in finalize_transcripts to DTMFAggregator

Salman778:
## Summary
Adds an opt-in `finalize_transcripts` parameter (default `False`) to `DTMFAggregator`. When enabled, the flushed `TranscriptionFrame` is marked `finalized=True` so user turn-stop strategies close the turn from the DTMF transcript alone.

## The bug this fixes

In a typical telephony flow, the agent prompts the caller to enter digits and the caller responds via keypad. What actually happens today:

1. Call starts. Caller speaks: *"Hi, I want to pay my bill."*
2. VAD fires `UserStartedSpeakingFrame` → STT produces a transcript → VAD fires `UserStoppedSpeakingFrame` → turn-stop strategy closes the turn → LLM responds, asks for account number.
3. Caller presses `1`, `2`, `3`, `4`, `#` on the keypad. `DTMFAggregator` aggregates the digits and pushes `TranscriptionFrame(text="DTMF: 1234#")` downstream.
4. **The agent does nothing.** The transcript sits buffered in `LLMUserContextAggregator` waiting for a turn-stop signal that never comes.
5. Caller, confused, says *"hello?"* → VAD fires → turn-stop strategy finally closes the turn → LLM is invoked with the buffered DTMF transcript + the spoken word.

The agent ignored the keypad entry until the caller also spoke.

## Why this happens

Turn-stop strategies (`TurnAnalyzerUserTurnStopStrategy`, `SpeechTimeoutUserTurnStopStrategy`) close a turn from a transcript alone only when one of these is true:

- The transcript is marked `finalized=True` — STT has signalled "nothing more coming."
- A fallback path runs: `_vad_user_speaking is False and _vad_stopped_time is None`, i.e. VAD has *never* fired in this session.

DTMF transcripts are produced with the default `finalized=False`. Once the caller has spoken once, `_vad_stopped_time` is set, so the fallback is permanently disabled. The buffered DTMF transcript has no path to trigger inference, and the strategy continues waiting for the next VAD `UserStoppedSpeakingFrame` — which only arrives when the caller speaks again.

## The fix

Add a `finalize_transcripts: bool = False` constructor parameter. When `True`, the flushed `TranscriptionFrame` is emitted with `finalized=True`, which routes through the strategies' "finalized transcript triggers turn stop" path and runs the LLM immediately on the DTMF input alone.

This is semantically correct: by the time `_flush_aggregation` runs, the DTMF sequence is definitionally complete (terminator pressed, idle timeout expired, or `EndFrame` received). There is nothing more coming from this synthetic transcript source, so marking it final is accurate — not just convenient.

The default stays `False` to preserve existing behavior. Users opt in with:

```python
dtmf = DTMFAggregator(finalize_transcripts=True)

### DIFF
--- a/changelog/0000.added.md
+++ b/changelog/0000.added.md
@@ -1,0 +1,7 @@
+- Added a `finalize_transcripts` option to `DTMFAggregator`. When enabled, the
+  flushed `TranscriptionFrame` is marked `finalized=True` so user turn-stop
+  strategies (e.g. `TurnAnalyzerUserTurnStopStrategy`,
+  `SpeechTimeoutUserTurnStopStrategy`) close the turn from the DTMF transcript
+  alone and the LLM runs immediately. Without this, once VAD has fired earlier
+  in the session, DTMF-only input does not trigger inference until the caller
+  also speaks. Defaults to `False` to preserve existing behavior.
--- a/src/pipecat/processors/aggregators/dtmf_aggregator.py
+++ b/src/pipecat/processors/aggregators/dtmf_aggregator.py
@@ -37,6 +37,12 @@ class DTMFAggregator(FrameProcessor):
     - EndFrame or CancelFrame is received
 
     Emits TranscriptionFrame for compatibility with existing LLM context aggregators.
+
+    By default the emitted ``TranscriptionFrame`` is not finalized to preserve
+    the original behavior. Pass ``finalize_transcripts=True`` so user turn-stop
+    strategies close the turn from the DTMF transcript alone — necessary when
+    the caller has already spoken in the session, because turn-stop strategies
+    otherwise wait for VAD events that DTMF does not produce.
     """
 
     def __init__(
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

**Comment by markbackman** [comment]
Hi — thanks for the PR. I tested `DTMFAggregator` in a simple voice pipeline and the DTMF transcript both pushes and triggers inference at the expected time — the LLM runs on the `DTMF: ...` input on its own, no second utterance needed. So I'm not able to reproduce the "agent does nothing" behavior, and unless there's a demonstrable issue I don't think we should add this option.

My setup — `DTMFAggregator` before the `LLMUserAggregator`, default turn strategies:

```python
pipeline = Pipeline([
    transport.input(),
    stt,
    dtmf_aggregator,
    user_aggregator,
    llm,
    tts,
    transport.output(),
    assistant_aggregator,
])
```

Could you share a repro? Specifically:

- What Pipecat version are you on? The transcript fallback that drives a turn without a VAD event landed in **v0.0.102** — on anything older, a DTMF-only turn wouldn't close on its own, which would match what you describe.
- What are your `user_turn_strategies`? With the defaults, `TranscriptionUserTurnStartStrategy` opens the turn on the DTMF transcript and the stop strategy closes it. If you've dropped that (e.g. VAD-only start, or external turn control), the turn never opens — but `finalize_transcripts` wouldn't address that either, since it only affects turn *stop*.

PR #4507, opened 2026-05-17
Fix/elevenlabs stt service crash language

BismeetSingh:
Elevenlabs STT supports passing language as None after which it should auto detect the language of the user, but while passing the language as None, that gets set as the lang code which crashes the STT service.


markbackman (approved):

PR #4510, opened 2026-05-17
Feat/OpenAI realtime codex auth

drew-royster:
* add support for codex auth for realtime (recently added in stealth to codex cli)
* prefers api key over codex auth, enable with `use_codex_auth` flag
* finds keys using same logic from codex-cli

image


markbackman:
Thanks for putting this together. I don't think we should add this to Pipecat right now.

The main concern is product/API fit: this adds support for reusing local Codex/ChatGPT auth instead of an OpenAI API key, but Pipecat's OpenAI service should stick to publicly documented OpenAI auth mechanisms. API keys already cover the normal local, CI, and production cases, and relying on Codex's private auth file/header behavior would make Pipecat depend on undocumented implementation details that may change without notice.

Because this is undocumented and entitlement-dependent, I don't think it is something we can confidently expose as part of the Pipecat service API. If OpenAI documents this auth flow for Realtime, we can revisit with a supported contract.

drew-royster:
fair @markbackman  thanks for looking it over. it's really handy for me so I may need to maintain a private fork for now.

PR #4511, opened 2026-05-17
feat(rime): add coda model support; default RimeTTSService model to coda

MaCaki:
## Summary

Adds support for Rime's `coda` TTS model across all three Rime services in `src/pipecat/services/rime/tts.py`:

- `RimeTTSService` (WebSocket JSON)
- `RimeHttpTTSService` (HTTP chunked)
- `RimeNonJsonTTSService` (deprecated raw WS — included for parity)

Coda mirrors Arcana on the wire but intentionally does **not** accept `temperature`, `top_p`, or `repetition_penalty`. The implementations skip those params for coda even when they're set on the settings object.

[Official Coda Docs](https://docs.rime.ai/docs/models#coda)

### Default model change for `RimeTTSService`

This PR also changes the default model for `RimeTTSService` from `arcana` to `coda`. **This is a behavior change** for callers that relied on the implicit default. To preserve previous behavior, pass `model="arcana"` explicitly (or via `settings=RimeTTSService.Settings(model="arcana", ...)`).

`RimeHttpTTSService` (default `mistv2`) and `RimeNonJsonTTSService` (deprecated, default `arcana`) are unchanged.

## Changes

- `RimeTTSService._build_ws_params()`: add a `coda` branch that contributes no model-specific WS query params.
- `RimeHttpTTSService.run_tts()`: no code change — the existing `if modelId == "arcana"` WAV-format gate already correctly excludes coda, which uses `audio/pcm`.
- `RimeNonJsonTTSService._connect_websocket()`: gate `repetition_penalty` / `temperature` / `top_p` on `model == "arcana"` so coda doesn't receive them.
- Docstring tweaks on `RimeNonJsonTTSSettings` / `RimeNonJsonTTSService.InputParams` to label those three params as `arcana only` (matching `RimeTTSSettings` style).
- Default model for `RimeTTSService` flipped from `arcana` to `coda`.
- [ ] Add support for the `timescalefactor` parameter 

## Test plan

- [x] `ruff check` and `ruff format --check` pass on the modified file.
- [x] Static smoke checks pass:
  - Default model for `RimeTTSService` is `coda`
  - WS params for `model="coda"` exclude `temperature` / `top_p` / `repetition_penalty` / Mist-only params
  - Arcana branch unchanged: still sends the three sampling params
  - Mistv2 branch unchanged: still sends `reduceLatency` / `pauseBetweenBrackets`
  - HTTP service: coda stays on `audio/pcm` (only arcana strips WAV header)
  - Deprecated NonJson service: Arcana params are gated on `model == "arcana"`
- [x] Live API path validated indirectly via a parallel change to `livekit-plugins-rime` ([livekit/agents#5748](https://github.com/livekit/agents/pull/5748)) which confirmed Rime's `/v1/rime-tts` and `/ws3` endpoints accept `modelId=coda` and return `audio/pcm` on both transports.


markbackman (commented):
The implementation for `RimeTTSService` looks good. Should the `RimeHttpTTSService` be updated similarly?

In testing, I notice that the word-timestamps lag the spoken audio by a similar margin to the `arcana` model on the `/ws3` endpoint. I'm assuming that this is a known issue. For example, in the screenshot, the white text shows the word-timestamps outputted by the time the audio finishes playing:
Image

You can see that this gap can be sizable for longer outputs. This isn't blocking since the issue exists for the `/ws3` endpoint already and your team is aware.

I'm not approving for now, awaiting on your response regarding `RimeHttpTTSService`.

MaCaki:
> Should the RimeHttpTTSService be updated similarly?

Yes!  Updating this class now. 

Regarding WLTs for `coda` - This is still a known discrepancy.  We're still using estimates for `coda`, `arcana` and `mistv3`, though making iterative improvements. 

On the other side of this launch however we're going to re-prioritize dialing in the WLT estimates as well as extracting and delivering model-native WLT's from `mistv3`

MaCaki:
`RimeHttpTTSService` updated to default to coda

src/pipecat/services/rime/tts.py
markbackman: This should be `coda`
```suggestion
            model="coda",
```

changelog/4511.changed.md
```diff
@@ -0,0 +1 @@
+- Changed the default model for `RimeTTSService` from `arcana` to `coda`. Code that relied on the implicit default should set `model="arcana"` explicitly to preserve previous behavior.
```
markbackman: Also, update the changelog to reflect this.

```suggestion
- Changed the default model for `RimeTTSService` and `RimeHttpTTSService` from `arcana` to `coda`. Code that relied on the implicit default should set `model="arcana"` explicitly to preserve previous behavior.
```

src/pipecat/services/rime/tts.py
MaCaki: whoops.  Updated w/ latest commit.

markbackman:
@MaCaki sorry for all of the "one more things". Can you lint the code?
`uv run scripts/fix-ruff-and-typecheck.sh`

MaCaki:
> @MaCaki sorry for all of the "one more things". Can you lint the code? `uv run scripts/fix-ruff-and-typecheck.sh`

yea I'll re-run that, ha I keep rushing through this, but now that we've completed all the release work I'll dot all the i's on this. 

I also need to add support for a new model parameter in this PR, so yet another detail on my end to address.

src/pipecat/services/rime/tts.py
```diff
@@ -881,9 +914,9 @@ class InputParams(BaseModel):
         Args:
             language: Language for synthesis. Defaults to English.
             segment: Text segmentation mode ("immediate", "bySentence", "never").
-            repetition_penalty: Token repetition penalty (1.0-2.0).
-            temperature: Sampling temperature (0.0-1.0).
-            top_p: Cumulative probability threshold (0.0-1.0).
+            repetition_penalty: Token repetition penalty (arcana only, 1.0-2.0).
```
markbackman, line 917: `RimeNonJsonTTSService` is deprecated and we traditionally don't update these classes. Perhaps we should leave out the update for this one?

markbackman (commented):
Looks good! Just one question about `RimeNonJsonTTSService`.

markbackman (approved):
Approving. I'll do some clean up in a separate PR. Thanks for the changes 🙇

src/pipecat/services/rime/tts.py
MaCaki: Sorry just seeing this - 

I was thinking along the same lines, but I'm not 100% that this class is actually deprecated, as I have _heard_ of some customers sill preferring the non-json, binary interface.  

However I guess that implies we need to make a distinction between _what_ is deprecated: 

* the  `pipecat` class `RimeNonJsonTTSService`
* the rime websockets `/ws` api serving non-json-wrapped-audio?

The latter is probably never going to be fully deprecated and decommissioned due to some customers preference for this format.  

Might this mean that `RimeNonJsonTTSService` should stay alive indefinitely? Or is the deprecation of non-json websockets support more a matter of the `pipepcat` library philosophy, where we'd _only_ want to support json-based websocket apis to minimize `pipecat` complexity as well as switching costs between tts ws apis for the end user?

markbackman: The Pipecat class `RimeNonJsonTTSService` was deprecated in 0.0.102. It will be removed, likely in the 2.0.0 release.

This class only existed as a gap filler when the `arcana` model didn't support the JSON response. Now that it does, this class no longer has meaning.

In Pipecat, we aim to maintain the best performing APIs as first class integrations. The `RimeNonJsonTTSService` lacks word timestamps and is fully replaced by `RimeTTSService`. Customers can easily migrate by just changing the class name and they automatically get better performance. Make sense?

PR #4512, opened 2026-05-18
fix: improve NVIDIA STT stream shutdown and reconnect handling

sathwikareddy02:
- move NVIDIA streaming audio iteration into a per-stream iterator
- unblock shutdown cleanly by closing the iterator with a sentinel
- add reconnect handling for dropped gRPC streaming sessions

### DIFF
--- a/src/pipecat/services/nvidia/stt.py
+++ b/src/pipecat/services/nvidia/stt.py
@@ -168,7 +168,6 @@ class AudioChunkIterator:
         """
         self._event_loop = event_loop
         self._queue: asyncio.Queue[bytes | object] = asyncio.Queue()
-        self._broken = False
         self._closed = False
 
     @property
@@ -182,7 +181,7 @@ class AudioChunkIterator:
         Args:
             audio: Raw PCM audio bytes to send to the server.
         """
-        if self._closed or self._broken:
+        if self._closed:
             return
         await self._queue.put(audio)
 
@@ -350,7 +349,6 @@ class NvidiaSTTService(STTService):
 
         self._asr_service = None
         self._audio_iterator: AudioChunkIterator | None = None
-        self._pending_audio_chunks: list[bytes] = []
         self._config = None
         self._thread_task = None
 
@@ -475,7 +473,6 @@ class NvidiaSTTService(STTService):
         await super().start(frame)
         self._initialize_client()
         self._config = self._create_recognition_config()
-        self._pending_audio_chunks.clear()
         self._audio_iterator = AudioChunkIterator(self.get_event_loop())
 
         if not self._thread_task:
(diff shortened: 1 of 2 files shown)


sathwikareddy02:
@filipi87 can you help review this fix?

src/pipecat/services/nvidia/stt.py
```diff
@@ -148,6 +148,80 @@ class NvidiaSegmentedSTTSettings(_NvidiaBaseSTTSettings):
...
+        """Initialize the iterator for a single streaming session.
+
+        Args:
+            event_loop: Event loop used to await queue operations from the
+                worker thread.
+        """
+        self._event_loop = event_loop
+        self._queue: asyncio.Queue[bytes | object] = asyncio.Queue()
+        self._broken = False
+        self._closed = False
```
filipi87: Do we need both states? I see that we are directly accessing `iterator._broken = True` inside `_response_handler`. Couldn't we invoke `iterator.close()` there instead and remove this _broken state ?

src/pipecat/services/nvidia/stt.py
```diff
@@ -275,7 +349,8 @@ def __init__(
         self._function_id = model_function_map.get("function_id")
 
         self._asr_service = None
-        self._queue = None
+        self._audio_iterator: AudioChunkIterator | None = None
+        self._pending_audio_chunks: list[bytes] = []
```
filipi87: It also seems that, now that we have a dedicated iterator with a queue to cache the audio, `self._pending_audio_chunks` should not be needed. Can we remove it ?

filipi87 (commented):
Hi @sathwikareddy02, the changes make sense. There are two points that I think we can simplify, but overall it looks good. Let me know when it's ready for review again.

src/pipecat/services/nvidia/stt.py
sathwikareddy02: I simplified this flow so we no longer need a separate _broken state. Instead of marking the iterator as broken, we now keep a single iterator alive across unexpected server-side stream drops and reattach that same iterator to the new gRPC stream during reconnect. We only close the iterator on client-initiated stop/cancel, so the shutdown path and reconnect path stay distinct.

src/pipecat/services/nvidia/stt.py
sathwikareddy02: I also removed _pending_audio_chunks. Since we now keep the same iterator and queue alive across reconnect, any audio that arrives while the server-side stream is down stays buffered in that iterator and is consumed by the new stream once reconnect completes. That lets us handle the handoff with a single queue instead of splitting it across iterator state plus a second pending buffer.

filipi87 (approved):
LGTM! Thank you for the improvements @sathwikareddy02 .

PR #4513, opened 2026-05-18
Handle OpenRouter developer messages conservatively

markbackman:
## Context

I found this one in supporting a customer who was using `google/gemini-2.5-flash`. Function calling wasn't working correctly with the `developer` role set. This feels like a missing feature or bug in OpenRouter. I think the safer path is to just disable the developer message and have developers override as needed.

## Summary

- Convert OpenRouter `developer` messages to `user` messages by default so async tool result messages work across more OpenRouter-backed models.
- Update the OpenRouter default model to `openai/gpt-4.1`.
- Align function-calling examples by adding the initial context message on client connect and letting the OpenRouter example use the service default model.
- Add a changelog entry for the OpenRouter behavior changes.

## Testing

- `uv run pytest tests/test_run_inference.py::test_openrouter_run_inference_converts_developer_messages_to_user`
- `uv run ruff check src/pipecat/services/openrouter/llm.py tests/test_run_inference.py`
- `uv run ruff check examples/function-calling/function-calling-azure.py examples/function-calling/function-calling-cerebras.py examples/function-calling/function-calling-deepseek.py examples/function-calling/function-calling-direct.py examples/function-calling/function-calling-groq.py examples/function-calling/function-calling-mistral.py examples/function-calling/function-calling-novita.py examples/function-calling/function-calling-nvidia.py examples/function-calling/function-calling-ollama.py examples/function-calling/function-calling-openai-responses-async-stream.py examples/function-calling/function-calling-openai.py examples/function-calling/function-calling-openrouter.py examples/function-calling/function-calling-qwen.py examples/function-calling/function-calling-sambanova.py`

### DIFF
--- a/changelog/4513.changed.2.md
+++ b/changelog/4513.changed.2.md
@@ -1,0 +1,1 @@
+- OpenRouter LLM requests now convert `developer` messages to `user` messages by default for broader model compatibility. Override this by subclassing `OpenRouterLLMService` or setting `llm.supports_developer_role = True` for models that support the `developer` role.
--- a/changelog/4513.changed.md
+++ b/changelog/4513.changed.md
@@ -1,0 +1,1 @@
+- OpenRouter LLM service now defaults to `openai/gpt-4.1`.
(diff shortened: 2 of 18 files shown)


filipi87 (approved):
Makes sense. 👍🚀

PR #4514, opened 2026-05-18
Align websocket STT connection failures

markbackman:
## Context

On connection failure, several `WebsocketSTTService` implementations previously pushed an error and then re-raised the exception. Raising is undesirable because `ServiceSwitcher` can fail over when services emit non-fatal `ErrorFrame`s. The better path is to clear stale connection state and push the error without re-raising. This change updates websocket STT services to follow that pattern.

## Summary

- Fixed websocket STT connection setup failures so services clear stale websocket state and emit non-fatal error frames.
- Avoided re-raising connect setup exceptions so `ServiceSwitcher` failover can handle active service failures and keep agents running.
- Added focused Cartesia and Soniox regression coverage, plus a changelog entry for the fix.

## Testing

- `uv run ruff check src/pipecat/services/aws/stt.py src/pipecat/services/gladia/stt.py src/pipecat/services/gradium/stt.py src/pipecat/services/xai/stt.py src/pipecat/services/elevenlabs/stt.py src/pipecat/services/assemblyai/stt.py src/pipecat/services/soniox/stt.py src/pipecat/services/cartesia/stt.py tests/test_cartesia_stt.py tests/test_soniox_stt.py`
- `uv run pytest tests/test_cartesia_stt.py tests/test_soniox_stt.py tests/test_elevenlabs_stt.py tests/test_service_switcher.py tests/test_websocket_service.py`
- `uv run towncrier build --draft --version Unreleased`

### DIFF
--- a/changelog/4514.fixed.md
+++ b/changelog/4514.fixed.md
@@ -1,0 +1,1 @@
+- Fixed websocket STT connection setup failures so services clear stale websocket state and emit non-fatal error frames, allowing `ServiceSwitcher` failover to keep agents running.
--- a/src/pipecat/services/assemblyai/stt.py
+++ b/src/pipecat/services/assemblyai/stt.py
@@ -586,9 +586,9 @@ class AssemblyAISTTService(WebsocketSTTService):
             await self._call_event_handler("on_connected")
             logger.debug(f"{self} Connected to AssemblyAI WebSocket")
         except Exception as e:
+            self._websocket = None
             self._connected = False
             await self.push_error(error_msg=f"Unable to connect to AssemblyAI: {e}", exception=e)
-            raise
 
     async def _disconnect_websocket(self):
         """Close the websocket connection to AssemblyAI."""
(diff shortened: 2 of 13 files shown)


changelog/4514.fixed.md
filipi87, line 1: Hi Mark, with this behavior change, shouldn’t we also update `_reconnect_websocket` inside `WebsocketService` ? I believe it currently expects the service to raise an exception in case of failure.

markbackman: Nice catch. Before, we were relying on `_connect_websocket()` to raise when a connection issue occurred, but I think that was a mistake because it entangles two concepts.

Instead, we want:

1. When an initial connection fails, we don’t want to raise. We push a non-fatal ErrorFrame via push_error() so ServiceSwitcher can switch to a working service.
2. When an initial connection succeeds and there’s an unexpected websocket closure, `_try_reconnect()` is responsible for retrying. `_reconnect_websocket()` performs each reconnect attempt, and now uses `_verify_connection()` to turn an unsuccessful attempt into an exception that `_try_reconnect()` catches for retry/backoff.

Here’s a commit to ensure that 2 works correctly:
https://github.com/pipecat-ai/pipecat/pull/4514/changes/9586db5b50d91f45228f274a4689e8a9f02be000

filipi87 (approved):
LGTM!

PR #4517, opened 2026-05-18
Fix ElevenLabs Chinese/Japanese timestamp spacing

### LINKED ISSUE #4515: ElevenLabsTTSService: missing includes_inter_frame_spaces hint for Japanese produces stray spaces in assistant transcript
## Bug summary

When `ElevenLabsTTSService` is used with a non-space-separated language (tested with Japanese; same root cause likely affects other non-space-separated scripts but I have not verified them), the assistant transcript stored in the LLM context contains spurious spaces between word-level chunks. For Japanese this turns `こんにちは、ご時間` into `こんにちは 、ご時間` — visible spaces between every chunk that ElevenLabs returned in a separate WS alignment message.

**Pipecat version:** 1.2.1
**ElevenLabs model used:** `eleven_turbo_v2_5` / `eleven_flash_v2_5` (multi-stream-input WS endpoint)

## Reproduction

1. Configure `ElevenLabsTTSService` with `Settings(language="ja", ...)`.
2. Generate any multi-chunk Japanese TTS response.
3. Read the persisted assistant message from `LLMContext.messages`.

Observed:
```
こんにちは 、ご時間 をいただきありがとうございます。
```

Expected:
```
こんにちは、ご時間をいただきありがとうございます。
```

## Root cause

`pipecat/services/elevenlabs/tts.py:893` calls:

```python
await self.add_word_timestamps(word_times, received_ctx_id)
```

without passing `includes_inter_frame_spaces`. Default propagates as `None` → `_add_word_timestamps` (`tts_service.py:1193-1194`) leaves it on the `TTSTextFrame`'s post-init default of `False` (from `TextFrame.__post_init__`).

Downstream, `LLMAssistantAggregator._handle_text` (`processors/aggregators/llm_response_universal.py:1499-1503`) builds:

```python
TextPartForConcatenation(frame.text, includes_inter_part_spaces=frame.includes_inter_frame_spaces)
```

`concatenate_aggregated_text` then injects an inter-part space between every word frame, because the parts claim **not** to embed their own inter-word spaces.

This is correct for space-separated scripts (English, Spanish, etc.) — ElevenLabs returns words without leading/trailing spaces, and the aggregator's space injection produces well-formed text. For Japanese, where the source chars contain no whitespace, every chunk arrives as a single "word" and the aggregator concatenates them with stray spaces.

Example WS message (`alignment.chars`) for Japanese, observed in production:
```python
{'alignment': {
    'chars': ['こ', 'ん', 'に', 'ち', 'は', '\n', '\n'],
    'charStartTimesMs': [0, 128, 186, 325, 464, 639, 743],
    'charDurationsMs': [128, 58, 139, 139, 175, 104, 0],
 },
 'normalizedAlignment': {
    'chars': [' ', 'k', 'o', 'n', 'n', 'i', 'c', 'h', 'i', 'h', 'a', ' ', '.'],
    ...
 },
 ...}
```

`calculate_word_times` splits on `" "`, so this chunk emits a single word `こんにちは\n\n`. The next chunk emits the next phrase as another single word. The aggregator joins them with an injected space.

## Suggested fix

In `ElevenLabsTTSService._receive_messages`, pass `includes_inter_frame_spaces=True` to `add_word_timestamps` when the configured language is a non-space-separated script:

```python
_NON_SPACE_SEPARATED = {"ja"}  # extend as needed (zh, th likely apply too but I have not tested)

# inside _receive_messages, where add_word_timestamps is called:
lang = (self._settings.language or "").split("-")[0].lower()
ifs = True if lang in _NON_SPACE_SEPARATED else None
await self.add_word_timestamps(word_times, received_ctx_id, includes_inter_frame_spaces=ifs)
```

The signal is essentially \"the tokens emitted for this script already constitute the full text — do not inject extra separators between them,\" which matches the existing semantics of `TextFrame.includes_inter_frame_spaces`.

Rationale for inferring from `settings.language` rather than detecting at runtime per chunk:
- The TTS service already knows the synthesis language.
- Per-chunk Unicode-range sniffing would also work but is heavier and would handle mixed-script content marginally better; happy to take either direction.

A similar consideration applies to the HTTP TTS path (`_handle_message` around `tts.py:1421`), which has the same `add_word_timestamps` call without the hint.

## Workaround

For anyone hitting this before a fix lands, subclass and override `add_word_timestamps`:

```python
class FixedElevenLabsTTSService(ElevenLabsTTSService):
    def __init__(self, *args, language: str = "", **kwargs):
        super().__init__(*args, **kwargs)
        self._words_already_spaced = language.split("-")[0].lower() in {"ja"}

    async def add_word_timestamps(self, word_times, context_id=None, includes_inter_frame_spaces=None):
        if includes_inter_frame_spaces is None and self._words_already_spaced:
            includes_inter_frame_spaces = True
        return await super().add_word_timestamps(word_times, context_id, includes_inter_frame_spaces)
```

markbackman:

Fixes #4515.

## What changed
- Treat ElevenLabs Chinese and Japanese timestamp text as space-preserving so downstream aggregation does not inject artificial spaces between provider timestamp chunks.
- Apply the same spacing hint to WebSocket timestamps, HTTP streaming timestamps, and the final partial-word flush.
- Add focused tests for Japanese, Chinese, English, and the language selection helper.

## Validation
- uv run ruff format --check src/pipecat/services/elevenlabs/tts.py tests/test_elevenlabs_tts.py
- uv run ruff check src/pipecat/services/elevenlabs/tts.py tests/test_elevenlabs_tts.py
- uv run pytest tests/test_elevenlabs_tts.py


filipi87 (approved):
LGTM!

It would be nice if you could merge the changes from `main` into this branch, given all the smart text handling changes, and make sure everything is still working as expected.

markbackman:
@filipi87 — flagging this for you since it overlaps with the word-timestamp refactor you just landed in #4380.

This PR fixes an extra space that shows up between Chinese/Japanese sentences in the assistant context (e.g. `…です。 何か…` instead of `…です。何か…`). The fix works by having ElevenLabs pass `includes_inter_frame_spaces=True` for `zh`/`ja`, which the assistant aggregator relies on: `concatenate_aggregated_text()` inserts a space between two consecutive parts only when **both** are `includes_inter_part_spaces=False` (`utils/string.py`). CJK sentences arrive as one whole-sentence word-timestamp frame each, so without the flag they get joined with a bogus space.

The problem: #4380 removed the mechanism this fix depends on. In the old path the value was written onto each emitted frame:

```python
frame = TTSTextFrame(word, aggregated_by=AggregationType.WORD)
if includes_inter_frame_spaces is not None:
    frame.includes_inter_frame_spaces = includes_inter_frame_spaces
```

After #4380, the word path goes `add_word_timestamps()` → `_add_word_timestamps()` → `AggregatedFrameSequencer.process_word()` → `_build_word_frame()`, and:

- `add_word_timestamps()` (`services/tts_service.py`) now only uses `includes_inter_frame_spaces` for `merge_punct_tokens()` and never forwards it.
- `_add_word_timestamps()` no longer takes the parameter at all.
- `_build_word_frame()` (`utils/context/aggregated_frame_sequencer.py`) constructs the `TTSTextFrame` setting `pts`, `context_id`, `append_to_context`, and `raw_text` — but not `includes_inter_frame_spaces`, so it stays at the `TextFrame` default `False`.

Net effect: once this PR rebases on main, the `True` we pass from ElevenLabs is dropped before the frame is built, so the per-word/sentence `TTSTextFrame` is always `False` regardless of language and the CJK space comes back. I confirmed the regression is live on `main` — the failing trace shows `WordCompletionTracker` running (so it's the #4380 path) together with the extra space in the context message.

Could you take a look at how you'd want to handle this in the new sequencer/tracker design? The most natural fix seems to be restoring propagation of `includes_inter_frame_spaces` through `add_word_timestamps → _add_word_timestamps → process_word → _build_word_frame` (and re-adding it to `_WordTimestampEntry` for the audio-context path), so a service can still mark a frame's text as already-spaced. But since this is your new code, I'd rather you decide the cleanest approach — the slot/tracker may be a better place to carry it than the per-word call. Happy to pair on it. Worth noting the CJK tests added in #4380 don't cover the two-sentence join, which is why this slipped through.

filipi87 (approved):
Hi @markbackman, I have pushed a fix that separates the two concerns: a new `pre_merge_tokens` parameter now controls token pre-merging (used by Inworld, which emits punctuation as separate tokens), while `includes_inter_frame_spaces` is restored to its original purpose of flowing through the word-timestamp pipeline all the way to `TTSTextFrame.includes_inter_frame_spaces`

I have also created some automated tests to make sure they are working.

Thank you for catching this.

Approved! 🚀

markbackman:
> Hi @markbackman, I have pushed a fix that separates the two concerns: a new `pre_merge_tokens` parameter now controls token pre-merging (used by Inworld, which emits punctuation as separate tokens), while `includes_inter_frame_spaces` is restored to its original purpose of flowing through the word-timestamp pipeline all the way to `TTSTextFrame.includes_inter_frame_spaces`
> 
> I have also created some automated tests to make sure they are working.
> 
> Thank you for catching this.
> 
> Approved! 🚀

In testing ElevenLabs with this latest branch, I still see the issue. You can repro with ElevenLabs using Japanese:
```
    tts = ElevenLabsTTSService(
        api_key=os.getenv("ELEVENLABS_API_KEY", ""),
        settings=ElevenLabsTTSService.Settings(
            voice="H8ZPDxbrPcks5hEsi2fq",
            language=Language.JA,
            model="eleven_flash_v2_5",
        ),
    )
```

Then, just have the bot initiate the conversation. I see:

```
2026-05-25 17:51:40.972 | DEBUG    | pipecat.services.elevenlabs.tts:run_tts:990 - ElevenLabsTTSService#0: Generating TTS [こんにちは、私はあなたのお手伝いをするAIアシスタントです。]
2026-05-25 17:51:41.064 | DEBUG    | pipecat.services.elevenlabs.tts:run_tts:990 - ElevenLabsTTSService#0: Generating TTS [どんなことでも気軽に話しかけてくださいね。]
```

Actual output:
```
こんにちは、私はあなたのお手伝いをするAIアシスタントです。 どんなことでも気 軽に話しかけてくださいね。
```

Expected output:
```
こんにちは、私はあなたのお手伝いをするAIアシスタントです。どんなことでも気 軽に話しかけてくださいね。
```

This is very, very subtle and hard to see, but there's an extra space between the first and second sentences (between the `。` and `ど`). That space shouldn't be there.

filipi87:
Hi @markbackman , thank you for catching this, there was one more thing missing. I fixed it here:
- https://github.com/pipecat-ai/pipecat/pull/4517/changes/d56eb811c5c06c4a61b1bc83667b742fee57125e#diff-0dfe1ceebb2564cfbce9ae613fbb365b06f921213b775d09f5e7993c5ba3154bR197

Now I believe everything looks correct:
```
2026-05-26 11:35:03.916 | DEBUG    | pipecat.services.elevenlabs.tts:run_tts:990 - ElevenLabsTTSService#0: Generating TTS [こんにちは、私はあなたのAIアシスタントです。]
2026-05-26 11:35:04.280 | DEBUG    | pipecat.services.elevenlabs.tts:run_tts:990 - ElevenLabsTTSService#0: Generating TTS [日常の質問や雑談、困りごとがあれば、なんでもサポートします。]
2026-05-26 11:35:04.330 | DEBUG    | pipecat.services.elevenlabs.tts:run_tts:990 - ElevenLabsTTSService#0: Generating TTS [お気軽に話しかけてくださいね。]

2026-05-26 11:35:24.916 | DEBUG    | pipecat.services.openai.base_llm:get_chat_completions:296 - OpenAILLMService#0: Generating chat from context [{'role': 'developer', 'content': 'ユーザーに自己紹介をしてください。'}, {'role': 'assistant', 'content': 'こんにちは、私はあなたのAIアシスタントです。日常の質問や雑談、困りごとがあれば、なんでもサポートします。お気軽に話しかけてくださいね。'}]
```

I have also added some extra tests to catch this.

markbackman:
> Hi @markbackman , thank you for catching this, there was one more thing missing. I fixed it here:
> 
> * [d56eb81#diff-0dfe1ceebb2564cfbce9ae613fbb365b06f921213b775d09f5e7993c5ba3154bR197](https://github.com/pipecat-ai/pipecat/commit/d56eb811c5c06c4a61b1bc83667b742fee57125e#diff-0dfe1ceebb2564cfbce9ae613fbb365b06f921213b775d09f5e7993c5ba3154bR197)
> 
> Now I believe everything looks correct:
> 
> ```
> 2026-05-26 11:35:03.916 | DEBUG    | pipecat.services.elevenlabs.tts:run_tts:990 - ElevenLabsTTSService#0: Generating TTS [こんにちは、私はあなたのAIアシスタントです。]
> 2026-05-26 11:35:04.280 | DEBUG    | pipecat.services.elevenlabs.tts:run_tts:990 - ElevenLabsTTSService#0: Generating TTS [日常の質問や雑談、困りごとがあれば、なんでもサポートします。]
> 2026-05-26 11:35:04.330 | DEBUG    | pipecat.services.elevenlabs.tts:run_tts:990 - ElevenLabsTTSService#0: Generating TTS [お気軽に話しかけてくださいね。]
> 
> 2026-05-26 11:35:24.916 | DEBUG    | pipecat.services.openai.base_llm:get_chat_completions:296 - OpenAILLMService#0: Generating chat from context [{'role': 'developer', 'content': 'ユーザーに自己紹介をしてください。'}, {'role': 'assistant', 'content': 'こんにちは、私はあなたのAIアシスタントです。日常の質問や雑談、困りごとがあれば、なんでもサポートします。お気軽に話しかけてくださいね。'}]
> ```
> 
> I have also added some extra tests to catch this.

Looks great now! Merging 🚢

PR #4522, opened 2026-05-19
Add P99 latency for Smallest AI, Mistral, XAI STT

markbackman:
## Summary

- Replaces the conservative default P99 time-to-first-speech estimate for Smallest AI STT with a benchmarked value (1.59s), giving consumers of `stt_latency` a more accurate latency budget for that provider.
- Also adding Mistral and XAI P99 latency numbers.


filipi87 (approved):
LGTM!

PR #4524, opened 2026-05-19
Improve runner optional transport handling

markbackman:
## Context

In testing out the new websocket endpoint, I found that the unified start endpoint in the local dev runner requires all optional dependencies to be installed. In pulling that thread, I ended up making a few improvements:

- Add checks for optional deps and return early from routes that aren't enabled. This avoids the primary issue.
- In the information that's displayed with the unified start endpoint, we lost the simplicity of a single URL to click. This will make the quickstart harder and doesn't funnel users to the Pipecat Prebuilt. So, the update is to refine the start information to show:

```
🚀 Bot ready!
   → Open: http://localhost:7860
   → Enabled transports: daily, telephony, websocket
   → Disabled transports: webrtc (install pipecat-ai[webrtc])
```

This concisely shows the user what to open, which transports are supported, which are disabled, and how to install them. This only reports which runner routes are available in the current environment; it does not imply that the bot implementation supports every enabled transport.

- In addition, when the client attempts to start a transport that isn't available because an optional dependency is missing, the runner returns a clear `400` response before trying to start the bot.

## Summary

- Fixed the development runner so missing optional transport dependencies disable only their related routes instead of failing startup in all-transport mode
- Clarified the startup banner to show one prebuilt client URL and concise enabled/disabled transport status with install hints
- Added changelog entries and focused tests for optional route registration, disabled starts, dependency checks, and banner output

## Testing

- uv run pytest tests/test_runner_run.py
- uv run ruff check src/pipecat/runner/run.py tests/test_runner_run.py

### DIFF
--- a/src/pipecat/runner/run.py
+++ b/src/pipecat/runner/run.py
@@ -90,6 +90,7 @@ To run locally:
 
 import argparse
 import asyncio
+import importlib.util
 import mimetypes
 import os
 import sys
@@ -131,6 +132,11 @@ load_dotenv(override=True)
 os.environ["ENV"] = "local"
 
 TELEPHONY_TRANSPORTS = ["twilio", "telnyx", "plivo", "exotel"]
+TRANSPORT_ROUTE_DEPENDENCIES = {
+    "daily": ("daily",),
+    "webrtc": ("aiortc",),
+    "websocket": ("fastapi", "websockets"),
+}
 
 # Mirror Pipecat Cloud's 4-hour max session limit so dev rooms get cleaned up.
 PIPECAT_ROOM_EXP_HOURS = 4.0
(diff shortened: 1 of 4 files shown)


src/pipecat/runner/run.py
```diff
@@ -156,6 +168,121 @@ async def my_route():
...
+
+
+def _transport_status_lists() -> tuple[list[str], list[str]]:
+    """Return enabled and disabled transport labels for the startup banner."""
+    transports = ["daily", "webrtc", "telephony", "websocket"]
+    enabled = []
+    disabled = []
+
+    for label in transports:
+        transport = TELEPHONY_TRANSPORTS[0] if label == "telephony" else label
```
filipi87: In the case of telephony, this would be available for all providers, not just Twilio, right?

The only difference is that we can only connect to Twilio through the web interface. I believe that’s the only distinction.

src/pipecat/runner/run.py
```diff
@@ -908,6 +1051,9 @@ def _setup_telephony_routes(app: FastAPI, args: argparse.Namespace):
     specific telephony transport is chosen via ``-t`` because the XML template
     is provider-specific and requires a proxy hostname (``--proxy``).
     """
+    if not _transport_routes_enabled("twilio"):
+        return
```
filipi87: So, this is where we define the route for all telephony providers, which makes it feel a bit odd to check only for “twilio” here.

filipi87 (approved):
LGTM! Nice catch!

src/pipecat/runner/run.py
```diff
@@ -156,6 +169,120 @@ async def my_route():
...
+    """Check whether a module can be imported without importing it.
+
+    Args:
+        module: Fully-qualified module name to check.
+
+    Returns:
+        ``True`` if Python can resolve the module, ``False`` otherwise.
+    """
+    try:
+        return importlib.util.find_spec(module) is not None
```
kompfner, line 182: In my testing, the changes in this PR don't seem to be working...and I think it's because this check is wrong. The modules of the various transports (e.g. the Daily transport) are _always_ available; but they might fail during import due to _their_ dependencies being unavailable (e.g. `daily-python`, brought in by the `pipecat-ai[daily]` extra).

kompfner: Ah, I think I see why this mechanism *appeared* work in testing...it was relying on another bug, which I squashed in some in-progress work. Stay tuned.

markbackman: Thanks for looking. Please feel free to raise a PR for this one 🙏

kompfner: Hmm, actually, maybe I'm misunderstanding something. We might be good here.

kompfner: Ohhh we're not checking for the modules themselves, we're already checking the transitive dependencies. Sorry, should've noticed that right away!

PR #4525, opened 2026-05-19
Raise ImportError instead of Exception for missing optional deps

markbackman:
Other options:
- We could re-raise the `ModuleNotFoundError`, but the `ImportError` feels like a better fit. Open to opinions. I ran into this when working on the local dev runner.

## Summary

- Across 84 files, the optional-dependency guard at module load did `raise Exception(f"Missing module: {e}")`. Replaced with `raise ImportError(f"Missing module: {e}") from e`.
- `ImportError` is the semantically correct type (`ModuleNotFoundError` is already a subclass), so callers can `except ImportError:` cleanly instead of being forced into a bare `except Exception:`.
- `from e` preserves the original `ModuleNotFoundError` traceback that was previously dropped.
- Two files in the repo (`audio/turn/krisp_viva_turn.py`, `turns/user_start/krisp_viva_ip_user_turn_start_strategy.py`) already used the correct pattern; no changes needed there.

Scope is intentionally narrow: only the `raise` line was touched. The two preceding `logger.error` lines and other unrelated `raise Exception(...)` sites are left alone.

## Test plan

- [x] `uv run ruff check` — passes
- [x] `uv run ruff format --check` — clean
- [x] `uv run pytest` — 1481 passed, 1 skipped. One unrelated pre-existing failure on `main` (`test_service_settings_complete[NvidiaSegmentedSTTService]` — `_settings.speaker_diarization is NOT_GIVEN`), not caused by this change
- [x] `rg 'raise Exception\(f"Missing module' src/` returns zero hits

### DIFF
--- a/changelog/4525.changed.md
+++ b/changelog/4525.changed.md
@@ -1,0 +1,1 @@
+- Services and transports with missing optional dependencies now raise `ImportError` instead of a bare `Exception` when their module is imported without the required extra installed. The original `ModuleNotFoundError` is preserved as `__cause__`, so code that wraps these imports can now use `except ImportError:` cleanly instead of `except Exception:`.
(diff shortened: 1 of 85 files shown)


filipi87 (approved):
Makes sense to me. 🚀

aconchillo (approved):

PR #4526, opened 2026-05-20
fix(sarvam): set finalized=True on terminal TranscriptionFrame

sandeep-equal:
## TL;DR

Anyone using `SarvamSTTService` in pipecat is silently paying **~1 second of latency on every user turn**, because Sarvam's final transcript is emitted without the `finalized=True` flag. Pipecat's user-turn-stop strategies need that flag to forward the transcript to the LLM immediately; without it they wait for a defensive 1.17 s timeout.

This PR adds the missing flag — a one-line code change that removes the wait. No behavior change for non-Sarvam consumers.

## Impact

**Theoretical**: this removes a `max(0, SARVAM_TTFS_P99 − vad.stop_secs)` wait per user turn. At pipecat's default `vad.stop_secs=0.3`, that's **0.87 s** of pure latency eliminated by the strategy's fast-path firing instead of the timeout fallback.

**Measured in a real-world voice-agent pipeline** (p50 over a multi-turn sample, before vs after applying the patch):

| Metric | Before | After |
|---|---:|---:|
| End-of-speech → LLM invocation | **1134 ms** | **~10 ms** |
| End-of-speech → bot speaking (TTFB) | **2037 ms** | **~1200 ms** |

That's a **~40 % drop in TTFB** purely from this one-line fix. The theoretical 0.87 s saving matches the observed reduction within measurement noise.

## What

`SarvamSTTService._handle_message` constructs the terminal `TranscriptionFrame` without `finalized=True` (see `src/pipecat/services/sarvam/stt.py`, the `message.type == \"data\"` branch). This PR adds the missing kwarg.

## Why

The user-turn-stop strategies that ship with the universal aggregator fast-path through to LLM invocation only when `frame.finalized=True`:

- `TurnAnalyzerUserTurnStopStrategy._handle_transcription` (`src/pipecat/turns/user_stop/turn_analyzer_user_turn_stop_strategy.py`)
- `SpeechTimeoutUserTurnStopStrategy._maybe_trigger_user_turn_stopped` (`src/pipecat/turns/user_stop/speech_timeout_user_turn_stop_strategy.py`)

Because Sarvam never sets the flag, both strategies fall through to a timeout derived from `SARVAM_TTFS_P99` (1.17 s, in `src/pipecat/services/stt_latency.py`) minus `vad.stop_secs`:

\`\`\`
wait = max(0, SARVAM_TTFS_P99 - vad.stop_secs)
     = max(0, 1.17           - 0.3)
     = 0.87s per user turn
\`\`\`

Sarvam's data event is the explicit terminal transcript (it follows \`END_SPEECH\` in the events stream), so \`finalized=True\` is the semantically correct flag.

## How

One-line addition: \`finalized=True\` on the \`TranscriptionFrame(...)\` constructor in the data-event branch. The adjacent \`self._handle_transcription(transcript, True, language)\` call already declared the frame is final — this just propagates the flag through to the actual emitted frame.

## Risk

Minimal.

- Behavior change: downstream consumers that read \`TranscriptionFrame.finalized\` (currently always seeing \`False\` from Sarvam) will now correctly see \`True\` on terminal transcripts. No consumer should ever assume final Sarvam transcripts are non-final, so this is upgrade-safe.
- The flag is opt-in for consumers; non-using consumers see no change.
- All other Sarvam STT behavior (VAD events, metrics, tracing, language mapping) preserved exactly.

## Tests

Adds \`tests/test_sarvam_stt.py\` with 4 tests:

1. \`test_data_event_emits_finalized_transcription_frame\` — pins the functional delta
2. \`test_empty_transcript_is_dropped\` — non-regression on whitespace-drop behavior
3. \`test_language_falls_back_to_hi_in_when_message_lacks_code\` — non-regression on language fallback
4. \`test_events_message_does_not_push_transcription_frame\` — VAD events still use \`broadcast_frame\`

All pass locally with \`pytest tests/test_sarvam_stt.py\`.

## Changelog

Added \`changelog/+sarvam-finalized.fixed.md\` (happy to rename to \`<PR_NUMBER>.fixed.md\` once this PR is assigned a number).


markbackman:
Thanks for the proposal.

There's a subtle issue here worth pointing out. While Sarvam supports flushing, they don't confirm in the response which transcript is the finalized result of that flush. To set `finalized=True`, Pipecat needs a confirmed finalize transcript, a signal that the last transcript chunk has been received.

Without that confirmation, Pipecat can close the user's turn prematurely. A late transcript chunk then opens a new turn that interrupts the bot's response.

So yes, Sarvam supports flushing the transcription, but until they add metadata to the returned transcript data, we can't reliably set finalization information on the `TranscriptionFrame`.

I'd love to see this functionality come to Sarvam, so feel free to reach out to the Sarvam team to encourage adding this metadata.

Closing this PR for now, since this can't be applied to Pipecat without a negative impact.

PR #4527, opened 2026-05-20
Fix ElevenLabs keepalive racing context-init (1008 disconnects)

### LINKED ISSUE #4520: `ElevenLabsTTSService` reinitializes `voice_settings` often, causing Websocket disconnects with code 1008
### pipecat version

1.2.0

### Python version

3.12

### Operating System

Debian 13, Windows

### Issue description

This appears to be directly related to https://github.com/pipecat-ai/pipecat/pull/4499
After upgrading to 1.2.0 we noticed extreme spike in ElevenLabs API errors with codes 1008 and 1006
Pipecat logs did show the exact reason of 1008: 
```
Something went wrong: ErrorFrame#0(error: Unknown error occurred: received 1008 (policy violation) voice_settings field must be provided in the first message and then either be not provided or not change.; then sent 1008 (policy violation) voice_settings field must be provided in the first message and then either be not provided or not change., fatal: False)
```
1006 errors are possibly directly caused by 1008.
It appears to be related with recreating context on each request, see the linked PR.
While in most cases WS managed to reconnect, we did have a few calls which apparently crashed unrecoverably due to this issue.

Implementing local patch from the PR does seem to resolve the issue.

### Reproduction steps

Upgrading to 1.2.0 and using `ElevenLabsTTSService` for more than a few dozen calls seem to always surface the issue.

### Expected behavior

No 1008 disconnects.

### Actual behavior

Many 1008 and 1006 status code disconnects.

### Logs

```shell

```

markbackman:
## Summary

- Fixes a race in `ElevenLabsTTSService` where the periodic keepalive could be sent for a new turn's context **before** that context's `voice_settings` context-init, causing ElevenLabs to close the WebSocket with a **1008 policy violation** (`voice_settings field must be provided in the first message and then either be not provided or not change.`).
- A turn's `context_id` is assigned on `LLMFullResponseStartFrame` (before the LLM produces any tokens), but `run_tts` doesn't send the context-init until text flows — a gap that spans the whole LLM TTFB. A 10s keepalive landing in that window becomes the context's first message with no `voice_settings`, so the later init looks like a change and is rejected. This is why it's intermittent and more likely when the LLM is slow to produce its first token.
- The keepalive now only stamps a `context_id` once that context's context-init has been sent (tracked in a new `_context_init_sent` set, cleaned up on context complete/interrupt and on disconnect). In the pre-init window it sends a context-less keepalive — there's no server-side context to keep alive yet anyway, and this keeps the real init as the context's first message.

## Testing

- `uv run pytest tests/test_elevenlabs_tts.py` — new regression tests covering: keepalive does not stamp a context before its init, keepalive stamps the context after its init, and keepalive with no active context sends a plain empty message.
- `uv run pytest tests/test_tts_frame_ordering.py` passes (no TTS frame-ordering regressions).
- `uv run ruff check` / `ruff format --check` clean.

## Fixes

- Fixes #4520


filipi87 (approved):
Makes sense. Nice catch!

PR #4530, opened 2026-05-20
Tavus faster than realtime.

filipi87:
## Summary
  
- Enables faster-than-realtime audio delivery for the Tavus transport: when the Tavus API returns a `stream_declared_sample_rate` (e.g. 48 kHz) in the conversation response, Pipecat declares that rate to `CustomAudioSource` while writing true rate PCM (e.g. 24 kHz), causing `write_frames()` to unblock ~2x faster and pre fill the avatar's buffer ahead of playback
- Disables WebRTC silence injection (auto_silence=False) on the Tavus audio destination so the buffer only grows from real TTS audio
- Moves conversation creation from `setup()` to `start()` so the true audio sample rate from `StartFrame` is available when calling the Tavus API
- Updates the `test_tavus_transport.py` proxy script to match: receives at the declared rate, buffers with a 100 ms pre-buffer, drains at the true rate, detects and filters WebRTC injected silence, and handles interrupts by flushing the buffer
- Documents the `TAVUS_SAMPLE_ROOM_URL` env var in `env.example`

## Testing   

Start a pipeline using `TavusTransport` or `TavusVideoService` with `TAVUS_SAMPLE_ROOM_URL` set to a Daily room URL, then run the proxy script to simulate Tavus avatar behavior: 

`uv run scripts/daily/test_tavus_transport.py`

The proxy should log that it is receiving at 48 kHz (declared) / 24 kHz (true content) and buffer status should show audio arriving faster than it drains.

## Possible limitations

**Packet Loss Concealment (PLC):** When a packet is lost, WebRTC's jitter buffer generates synthetic concealment audio at the declared rate (e.g. 48 kHz). Because each sample is played at the true rate, PLC artifacts play at half the expected pitch and double the expected duration. Packet loss sounds noticeably worse than in a normal stream.


src/pipecat/transports/tavus/transport.py
```diff
@@ -69,27 +75,37 @@ def __init__(self, api_key: str, session: aiohttp.ClientSession):
...
+                "stream_declared_sample_rate": _STREAM_DECLARED_SAMPLE_RATE,
             }
 
         logger.debug(f"Creating Tavus conversation: replica={replica_id}, persona={persona_id}")
         url = f"{self.BASE_URL}/conversations"
         payload = {
             "replica_id": replica_id,
             "persona_id": persona_id,
+            # TODO: start to send it when Tavus start to support it.
+            # "sample_rate": sample_rate,
```
filipi87, line 108: TODO: we will need to send this to Tavus so they are aware of the real sample rate.

filipi87:
@markbackman, @aconchillo, if you could test this and share your thoughts on the results, that would be great.

examples/video-avatar/video-avatar-tavus-video-service.py
filipi87, line 1: TODO: only for testing, remove the changes before merging.

src/pipecat/services/tavus/video.py
filipi87, line 1: TODO: the recording part is only for testing, remove the changes before merging.

filipi87: The prebuffer is needed.

scripts/daily/test_tavus_transport.py
filipi87, line 1: We had to add logic to the client so it knows how to handle:
- Buffering and playback adjustment: Because the audio arrives at 2× speed, the client must compensate for the rate mismatch before playback.
- Silence handling: WebRTC injects all-zero silence frames between speech bursts, and because the audio arrives at 2× speed, that silence accumulates twice as fast. The client needs to discard silence frames when the playback buffer is healthy.
- Interruption handling: Clear the playback buffer and prepare for when new audio starts arriving.

PR #4533, opened 2026-05-20
Decouple context management from turn frames for realtime services

kompfner:
## Problems

Realtime-service pipelines suffer four problems at the boundary between context management, turn frames, and turn events:

1. **No turn frames at all.** Gemini Live, AWS Nova Sonic, and Ultravox don't emit `UserStarted/StoppedSpeakingFrame`, and the aggregator — which writes user messages on those frames — doesn't write to context correctly without them.
2. **Local VAD as workaround is fragile and confusing.** Enabling local VAD lets the aggregator generate its own turn frames, but those boundaries are a guess that doesn't match the provider's server-side ground truth, and the per-service "do I need it?" rule is hard for users to keep straight.
3. **Transcript-bound turn-end adds latency.** When local turn detection is the *intended* driver of a realtime conversation, the default turn-end strategies still wait for transcripts before signalling end-of-turn — adding latency on the critical path. (PR #4480 tried to solve this in isolation.)
4. **Per-service user-turn-strategy configuration is brittle.** To ensure the `on_user_turn_started` and `on_user_turn_stopped` events are fired, and without duplication, the user has to know which strategies to wire — external-frame strategies for services that emit `UserStarted/StoppedSpeakingFrame` from server-side VAD (OpenAI Realtime, Azure, Grok, Inworld, when configured with server-side turns), local-VAD-driven strategies for services that don't (Gemini Live, Nova Sonic, Ultravox, or any of the other services when configured with locally-driven turns). This is a lot to know, and hard to get right.

Problems #1 and #2 stem from coupling context writes to turn frames. Problem #3 stems from coupling turn-end decisions to transcripts. Problem #3 can't be fixed in isolation — dropping the transcript wait would leave the aggregator writing empty or incomplete user messages — so the two couplings have to be untangled together. Problem #4 is a configuration-ergonomics issue rather than a bug — the user could always wire the right strategies manually — but the natural place to remove that footgun is the same place: when the user opts into realtime mode.

## Solution

A new `realtime_service_mode: bool = False` kwarg on `LLMContextAggregatorPair`. When `True`:

- **Context writes decouple from the `UserStoppedSpeakingFrame` signal.** Instead, the assistant aggregator uses the assistant response start as the equivalent signal, telling the user aggregator to flush its message then.
    - Context-writing now works whether or not the realtime service emits user turn frames.
    - When turns are locally driven, `UserStoppedSpeakingFrame` can be configured to fire earlier without waiting for user transcripts, reducing latency (see next bullet).
- **`UserStoppedSpeakingFrame` fires earlier, without waiting for user transcripts** — reducing latency on the critical path when local turn detection drives the conversation.
- **The default user turn strategies are replaced with `ExternalUserTurnStartStrategy` and `ExternalUserTurnStopStrategy`** when the realtime service advertises that it emits its own turn frames (OpenAI Realtime, Azure, Grok, Inworld in their default server-VAD modes) — so `on_user_turn_started` / `on_user_turn_stopped` fire from the server-emitted `UserStarted/StoppedSpeakingFrame`. For services that don't emit them — either because they never do (Gemini Live, Nova Sonic, Ultravox) or because server-side turn detection is disabled at runtime (e.g. OpenAI Realtime with `turn_detection=False`, in locally-driven-turns setups) — the defaults stay in place so locally-driven turn detection (e.g. local VAD) can fire the events. Passing custom `user_turn_strategies` opts out of the swap.

## API

```python
#
# Recommended aggregator setups
#

# 1. Cascade (default — unchanged from today):
user_agg, assistant_agg = LLMContextAggregatorPair(context)

# 2. Realtime service with server-driven turns and turn frames (e.g. OpenAI Realtime):
user_agg, assistant_agg = LLMContextAggregatorPair(
    context,
    realtime_service_mode=True,
)

# 3. Realtime service with either:
# - Locally-driven turns (e.g. OpenAI Realtime with turn_detection=False)
# - Server-driven turns supplemented with optional local turn frames (which you may
#   want to use with, for example, Gemini Live, which doesn't emit turn frames)
#   - This can be useful for other processors who expect turn frames, like RTVI,
#     or for triggering on_user_turn_* events
#   - WARNING: local supplemental turn frames may not align with server turns; this
#     was a common risk before this PR
user_agg, assistant_agg = LLMContextAggregatorPair(
    context,
    realtime_service_mode=True,
    user_params=LLMUserAggregatorParams(vad_analyzer=SileroVADAnalyzer()),
)

#
# Event handler setup
#

# Fired without message content when realtime_service_mode enabled on aggregator; 
# get message from on_user_turn_message_added instead.
# Fires if either:
# - Service emits turn frames
# - We're supplementing with local turn frames
@user_agg.event_handler("on_user_turn_stopped")
async def on_user_turn_stopped(aggregator, strategy, message):
    ...

# Source of truth for new user context messages when realtime_service_mode enabled on 
# aggregator
@user_agg.event_handler("on_user_turn_message_added")
async def on_user_turn_message_added(aggregator, message):
    ...
```

## ⚠️ Standard usage change ⚠️

Note that, with the changes in this PR, we're now uniformly demonstrating in our "base" example files the pattern of using **realtime services *without* local VAD/turn detection**.

We're recommending local VAD/turn detection only for cases where:
- You're choosing to drive the conversation from local turn detection (e.g. OpenAI Realtime with turn_detection=False)
- You need supplemental turn frames to drive other processors that expect them, (e.g. RTVI) or to trigger on_user_turn_* events (WARNING: in this case you should consider local turn frames as approximate, as they may not always align with server turns)

Prior to this PR, we were inconsistent about recommending the no-local-VAD pattern—it depended on the service. Also, prior to this PR, we were usually recommending enabling local VAD with no caveat; now we're at least giving the user a heads up that local turns aren't ground truth, and may not accurately reflect server turns.

## Test plan

- [ ] Existing test suite passes (aggregator + turn-related tests green locally)
- [ ] `TestRealtimeServiceModeAggregator` covers trailing-write ordering, interruption mid-assistant, strategy mutations (including the external-strategy auto-append and its opt-out via custom strategies), pair validation, metadata recommendation log, and function-call no-context-push
- [ ] Smoke-test each realtime service example end-to-end (OpenAI Realtime, Azure Realtime, Inworld, Grok/xAI, Gemini Live, Nova Sonic, Ultravox)
- [ ] Verify `InterruptionFrame` is now surfaced when interrupting Nova Sonic and Ultravox without local VAD wired
- [ ] Confirm `realtime-openai-locally-driven-turns.py`, `realtime-gemini-live-locally-driven-turns.py`, `realtime-grok-locally-driven-turns.py`, and `realtime-inworld-locally-driven-turns.py` variants drive turns locally as expected


src/pipecat/processors/aggregators/llm_response_universal.py
```diff
@@ -606,8 +690,19 @@ def __init__(
             )
             self._params.user_turn_strategies = user_turn_strategies
 
+        # Realtime mutation: when turns shouldn't wait for transcripts,
+        # drop the transcription-based start strategy and flip the
+        # wait_for_transcript flag on stop strategies that expose it. The
+        # set of strategies that support it intentionally stays narrow —
+        # the flag was reintroduced specifically for this realtime path.
```
kompfner: TODO: not "reintroduced"; the reader doesn't know about PR 4480, which was never merged

src/pipecat/processors/aggregators/llm_response_universal.py
```diff
@@ -606,8 +690,19 @@ def __init__(
...
+        if not self._turns_await_transcripts:
+            self._apply_realtime_strategy_mutations(user_turn_strategies)
+
         self._user_is_muted = False
         self._user_turn_start_timestamp = ""
+        # Tracks whether the §3.6 recommendation log has already fired
```
kompfner: TODO: reader doesn't know about the Claude plan; it won't be committed. We shouldn't reference it anywhere.

src/pipecat/processors/aggregators/llm_response_universal.py
```diff
@@ -748,13 +853,138 @@ async def _start(self, frame: StartFrame):
             await s.setup(self.task_manager)
 
     async def _stop(self, frame: EndFrame):
-        await self._maybe_emit_user_turn_stopped(on_session_end=True)
+        if not self._context_writes_await_turns:
```
kompfner: TODO: maybe we should:
- reference the full `self._realtime_service_mode.context_writes_await_turns` for clarity (and same feedback goes for `self.realtime_service_mode.turns_await_transcripts`)
- rather than just refer to "realtime" in comments, refer to "realtime mode"

src/pipecat/processors/aggregators/llm_response_universal.py
```diff
@@ -748,13 +853,138 @@ async def _start(self, frame: StartFrame):
...
+                self._context_writes_await_turns != self._paired_half._context_writes_await_turns
+                or self._turns_await_transcripts != self._paired_half._turns_await_transcripts
+            ):
+                raise RuntimeError(
+                    f"{self}: realtime-mode config mismatch between user and "
+                    "assistant halves. Use LLMContextAggregatorPair to construct "
+                    "the pair so both halves share the same configuration."
+                )
+
+    def _apply_realtime_strategy_mutations(self, user_turn_strategies: UserTurnStrategies) -> None:
```
kompfner: TODO: `_apply_realtime_strategy_mutations` -> `_apply_realtime_mode_strategy_mutations`

tests/test_context_aggregators_universal.py
```diff
@@ -1651,5 +1659,314 @@ async def test_pair_default_respects_per_params(self):
...
+            _realtime_service_mode=RealtimeServiceModeConfig(turns_await_transcripts=True),
+            _pair_lock=lock,
+        )
+        user._paired_half = assistant
+        assistant._paired_half = user
+        with self.assertRaises(RuntimeError):
+            user._validate_realtime_pairing()
+
+    async def test_function_call_no_context_push_in_realtime_mode(self):
+        # Realtime services consume function results directly via
```
kompfner: TODO: wait...this isn't true. I believe all realtime services receive new context frames and look for new function call result messages in them. If they find any, they send the resuilts to the realtime API.

tests/test_context_aggregators_universal.py
```diff
@@ -1651,5 +1659,314 @@ async def test_pair_default_respects_per_params(self):
         self.assertFalse(pair.assistant()._add_tool_change_messages)
 
 
+class TestRealtimeServiceModeConfig(unittest.TestCase):
+    def test_default_fields_are_realtime(self):
+        cfg = RealtimeServiceModeConfig()
+        self.assertFalse(cfg.context_writes_await_turns)
+        self.assertFalse(cfg.turns_await_transcripts)
+
+    def test_keep_transcripts_keep_writes_on_turn(self):
```
kompfner: TODO: `test_keep_transcripts_keep_writes_on_turn` and `test_keep_transcripts_trailing_writes` seem to be of very low value. Maybe not even worth keeping around.

src/pipecat/processors/aggregators/llm_response_universal.py
kompfner: TODO: if we don't already have unit tests validating that these mutations are applied correctly, it could be worth adding them.

src/pipecat/processors/aggregators/llm_response_universal.py
```diff
@@ -935,6 +1204,17 @@ async def _on_user_turn_stopped(
 
         await self._user_idle_controller.process_frame(UserStoppedSpeakingFrame())
 
+        if not self._context_writes_await_turns:
+            # Realtime: turn frames are supplemental. The user message
```
kompfner: (Here's another example where "realtime mode" might be clearer than just "realtime". Note that I won't be adding GitHub comments pointing out every example of this, though.)

src/pipecat/processors/aggregators/llm_response_universal.py
```diff
@@ -1184,6 +1494,10 @@ async def process_frame(self, frame: Frame, direction: FrameDirection):
             if self._push_context_on_bot_stopped_speaking and not self._user_speaking:
                 logger.debug(f"{self}: Bot stopped speaking — pushing deferred context frame!")
                 await self.push_context_frame(FrameDirection.UPSTREAM)
+        elif isinstance(frame, RealtimeServiceMetadataFrame):
+            # The user half logs the §3.6 recommendation; the assistant
```
kompfner: (Here's another example where we're referencing a section of the plan, which won't make sense to the reader. Note that I won't be adding GitHub comments pointing out every example of this, though.)

src/pipecat/processors/aggregators/llm_response_universal.py
```diff
@@ -1379,6 +1708,42 @@ async def _handle_function_call_result(self, frame: FunctionCallResultFrame):
             self._context_updated_tasks.add(task)
             task.add_done_callback(self._context_updated_task_finished)
 
+    async def _maybe_push_context_after_function_result(self) -> None:
+        """Decide whether to push a context frame after a function-call result.
+
+        Dispatched by mode. Cascade re-runs LLM inference by pushing an
+        ``LLMContextFrame`` upstream (with care to avoid duplicate pushes
+        while results are queued or the bot is still speaking). Realtime
```
kompfner: TODO: see feedback I left in the test file. This assumption isn't true, I believe.

src/pipecat/processors/aggregators/llm_response_universal.py
```diff
@@ -1469,6 +1834,20 @@ async def _handle_assistant_image_frame(self, frame: AssistantImageRawFrame):
             )
 
     async def _handle_llm_start(self, _: LLMFullResponseStartFrame):
+        if not self._context_writes_await_turns:
```
kompfner: TODO: worth a code comment here, like we do elsewhere we have a "realtime mode" code bifurcation at the top of a function. This feedback applies everywhere this pattern occurs.

src/pipecat/processors/aggregators/llm_response_universal.py
```diff
@@ -1620,9 +2003,86 @@ async def _trigger_assistant_turn_stopped(self, *, interrupted: bool = False):
...
+            await self._call_event_handler("on_assistant_message_added", message)
 
         self._assistant_turn_start_timestamp = ""
 
+    async def _realtime_trigger_assistant_turn_stopped(self, *, interrupted: bool):
+        """Realtime variant: defer the context write or flush on interruption.
+
+        Normal end-of-turn (``interrupted=False``, from
+        ``LLMFullResponseEndFrame``) parks the message text in a pending
+        slot — it isn't written to context until the next user transcript
```
kompfner: NOTE: why, again, is it beneficial to not write the assistant message until the next user transcript arrives?

I understand why it makes sense for the assistant turn to start before writing the user message - we simply have no reliable signal from realtime services that the user has finished speaking, other than the assistant starting to reply (which, importantly, is expected to happen quite shortly after the user has finished speaking).

But in the case of assistant messages:
- Don't we *always* have a reliable `LLMFullResponseEndFrame` to go off of?
- The next user reply can be *quite* a long time later...it's still better to write to context sooner, if we can

If we make the above change, we don't need each aggregator to hold on to the other as a `_paired_half`, right? Doesn't it become a one-way thing, with the assistant aggregator holding on to the user aggregator but not vice-versa?

kompfner:
@aconchillo, @markbackman, @filipi87, still doing some additional testing and clean up but wanted to already open this up for review since it's a lot to take in, both conceptually and code-wise. Happy to discuss if it'd be helpful! Also, any testing help would be greatly appreciated. I've mostly validated OpenAI, Gemini Live, and Nova Sonic, but the others still need some scrutiny.

changelog/4533.added.2.md
filipi87, line 1: I also wonder if we should have an `on_context_updated`, which we could invoke whenever the context is updated, regardless of whether it was updated by the user or the assistant.

changelog/4533.added.4.md
filipi87, line 1: If local VAD is enabled, are we still emitting the warning ? Maybe in this case we should emit a different message, considering that we’ll be emitting `UserStartedSpeakingFrame` / `UserStoppedSpeakingFrame` ?

What do you think ?

filipi87: * Local turn analyser.

changelog/4533.added.2.md
kompfner: Hmm, with this work I was only looking for a new drop-in replacement/equivalent to `on_user_turn_stopped` that would work in the case that the user message actually _trailed_ the turn stop.

If we implemented `on_context_updated` _instead_, it wouldn't quite be a drop-in replacement (the user's code would have to check what changed).

If we implemented `on_context_updated` _additionally_, we could do it as a follow-up item. It would have a broader area of effect than this PR.

filipi87: Yep, I was thinking about that additionally. Just sharing it here because I thought we already had it. And yes, we can do it in a follow-up PR. 👍

kompfner: Thinking about it a bit more now, though, the name `on_user_message_added` sounds like it promises a lot more than it's doing in this PR...

In this PR, we're firing it when the user aggregator flushes a user message to context. That's it. We don't fire when, say, a user message is programmatically added to context.

I'm open to either finding a narrower name for this event to reflect what it actually is, or looking into the more generic `on_context_updated`. It may be OK that it's not a simple drop-in replacement, as long as we have code examples showing how to use it (the user will have to write new code to opt into the new event anyway).

changelog/4533.added.md
```diff
@@ -0,0 +1,6 @@
+- Added a `realtime_service_mode: bool` kwarg on `LLMContextAggregatorPair`, for opting into a set of behaviors tailored for use with realtime (speech-to-speech) services. Setting `realtime_service_mode=True` does three things:
```
filipi87, line 1: Cool. I thought we would need a new aggregator, but it looks like you managed to handle all the different cases in the same one. 🎉

changelog/4533.added.2.md
kompfner: Thinking a little bit about this now...a generic `on_context_updated` might be too general for the purpose of logging user transcripts. It might be tricky to thread through a "why" to the context update payload. A "why" seems important for making user transcript logging code robust (it would need to ignore, say, programmatic context updates).

kompfner: So, if we were to go with the narrower name...how about `on_user_turn_message_added`? This name implies that it's tethered to a user turn.

changelog/4533.changed.2.md
filipi87, line 1: Just adding this as something to think about, probably for a follow-up PR. 

But should we do the same for Flux (and any other  STT services which provide turn control)? If no custom `user_turn_strategies` are provided, should the user aggregator use `ExternalUserTurnStartStrategy` and `ExternalUserTurnStopStrategy` in the default strategy chain ?

changelog/4533.fixed.2.md
filipi87, line 1: Nice catch! 🎉

changelog/4533.added.4.md
kompfner: > If local VAD is enabled, are we still emitting the warning

Yes; we don't really have a good way today of knowing how the rest of the pipeline is set up and whether something will provide those frames. (We could try to implement a new mechanism for this, but it would be a lot of additional complexity for this one thing).

Additionally, we also wanted to use this warning to convey something else: that local VAD-driven turns may actually subtly disagree with the "ground truth" server-decided turns, so they should be treated as "approximate".

Finally, this warning can be a nice visible nudge/reminder/(nag?) that we should encourage, however we can, realtime service providers to provide their own turn signals, as they're very useful for app developers!

kompfner: Let's definitely update this changelog to reflect the above additional reasons for the warning, and also maybe reword the warning itself to accomplish all of its goals better.

changelog/4533.added.2.md
filipi87: I like the name `on_user_turn_message_added`, but it gives me the impression that I will receive this message only once per user turn. Is that true ?

Will `on_user_turn_message_added` be invoked only once per turn? Or, if the user turn has already finished and we receive a new transcription message, will it theoretically be triggered again for the same turn ? Like if I receive more than one transcription after the turn has finished.

changelog/4533.added.md
kompfner: A whole new aggregator would've been really really nice from a code-separation/logical-complexity point of view. But a little analysis revealed that the _vast_ majority of the aggregator code would've had to be ported over or somehow shared between the current aggregator and the new one. The differences were very small compared to the commonalities.

changelog/4533.added.4.md
filipi87: Agreed, not worth the extra complexity only for this. Updating the changelog and reword the warning will be more than enough. 👍

changelog/4533.changed.2.md
kompfner: Oho, interesting thought! You mean, if it doesn't _hurt_ to have the external strategies there (if they simply no-op if they're not needed), then why not just have them there by default?

That's a really interesting thought. 🤔 Worth bringing to the team. @aconchillo, specifically, may have some thoughts.

changelog/4533.added.2.md
kompfner: > it gives me the impression that I will receive this message only once per user turn. Is that true?

That's the intent, yes!

src/pipecat/processors/aggregators/llm_response_universal.py
```diff
@@ -748,13 +860,157 @@ async def _start(self, frame: StartFrame):
...
+        assistant starts responding, so when no aggregated transcript
+        is present yet we defer the flush by ``ttfs_p99_latency`` —
+        the same wait the cascade pipeline applies after
+        ``UserStoppedSpeakingFrame``. If a transcript is already
+        aggregated (chunk-streaming services), flush immediately.
+        """
+        # Cancel any prior deferred flush task — a new turn supersedes it.
+        await self._cancel_realtime_handoff_flush_task()
+
+        if self._aggregation:
```
kompfner: Hmm, I actually think this is wrong. We should *always* do the deferred flush—simply having `self._aggregation` doesn't mean it's ready for flushing; it may be incomplete.

kompfner: (Note to self: remember to update the docstring accordingly once this is fixed)

changelog/4533.added.2.md
kompfner: Renamed.

changelog/4533.changed.2.md
filipi87: Not exactly, because if the user has not defined `user_turn_strategies`, we will use our default `UserTurnStrategies()`. And I believe we can't simply include the external strategies there, otherwise, we would emit the "user started speaking" and "user stopped speaking" events twice.

Instead, we would need logic similar to what you implemented for the realtime services.

We should be able to determine that, for these STT services, the default strategy should be the external one rather than our current default.

examples/realtime/realtime-aws-nova-sonic.py
```diff
@@ -147,11 +145,33 @@ async def run_bot(transport: BaseTransport, runner_args: RunnerArguments):
...
+    #     LLMUserAggregatorParams,
+    #     UserTurnStoppedMessage,
+    # )
+    # from pipecat.turns.user_stop import BaseUserTurnStopStrategy
+
     context = LLMContext(tools=tools)
     user_aggregator, assistant_aggregator = LLMContextAggregatorPair(
         context,
-        user_params=LLMUserAggregatorParams(vad_analyzer=SileroVADAnalyzer()),
+        realtime_service_mode=True,
```
filipi87, line 173: @kompfner, if we are already emitting the `RealtimeServiceMetadataFrame`, couldn’t we use it to define whether `realtime_service_mode` is true or false, instead of adding this new property ?

Or is there any case where we would be using a realtime service but wouldn’t want to set the `realtime_service_mode` flag on the aggregator ?

examples/realtime/realtime-ultravox.py
```diff
@@ -178,18 +174,31 @@ async def run_bot(transport: BaseTransport, runner_args: RunnerArguments):
...
-    # to produce user and assistant turn stopped events.
+    # Ultravox drives the conversation server-side.
+    #
+    # It does not, however, emit turn frames (UserStartedSpeakingFrame,
+    # UserStoppedSpeakingFrame). realtime_service_mode ensures that context
+    # aggregation will work without those frames, but you can add supplemental
+    # local turn frames for consumption by other pipeline processors that
+    # expect them (like RTVI), or to trigger on_user_turn_* events. WARNING:
+    # you should consider supplemental local turn frames approximate, as they
+    # may not always align with server turns.
```
filipi87, line 185: These comments are great. They make it really easy to understand the limitations and the available alternatives. 💯

changelog/4533.changed.2.md
kompfner: > otherwise, we would emit the "user started speaking" and "user stopped speaking" events twice

I think we have this same possible misconfiguration in this PR. 

With the code as it is here, if you (the developer) set `realtime_service_mode=True`, we automatically append the external strategies to pick up possible turn frames from the realtime service, but we keep the other strategies in place...

My thinking was that, if you have a service that emits its own turn frames, it's a misconfiguration anyway for you to *also* enable local VAD/turn detection in your pipeline. But...that might be a lot for you to have to be aware of...

Also, my misconfiguration assumption might be wrong: if you have a realtime service that emits its own frames and that you want to *drive* from local turn detection, I think you might get duplicate events in that case: both the service and the turn strategies will be emitting turn frames.

examples/realtime/realtime-aws-nova-sonic.py
kompfner: > couldn’t we use it to define whether realtime_service_mode is true or false, instead of adding this new property

That was my initial approach, but that would be a big backward-compatibility break (people would need to update their event handlers, for example). We (unfortunately, I think) need to have this new mode be opt-in.

src/pipecat/frames/frames.py
```diff
@@ -1439,6 +1439,27 @@ class STTMetadataFrame(ServiceMetadataFrame):
     ttfs_p99_latency: float
 
 
+@dataclass
+class RealtimeServiceMetadataFrame(ServiceMetadataFrame):
```
filipi87, line 1443: I mentioned this in another comment, but couldn’t this frame determine whether the context aggregator has its `realtime_service_mode` property set to true or false?

We could use a similar approach for services like Flux, where ideally we should include the external strategies by default instead of `UserTurnStrategies()`.

examples/realtime/realtime-aws-nova-sonic.py
kompfner: The middle-ground solution I came up with was to log a prominent warning when someone is using a realtime service but not setting `realtime_service_mode=True` (in llm_response_universal.py)

src/pipecat/frames/frames.py
kompfner: See my response to the other comment.

changelog/4533.changed.2.md
kompfner: OK, maaaybe what we need to do is this:

- If `realtime_service_mode=True` and the service advertises that it provides its own frames (using `RealtimeServiceMetadataFrame.emits_user_turn_frames`), we make the default strategies be *only* the external strategies (and we print an info log explaining this)
- We make sure that, if a realtime service is configured to be locally-driven, it does *not* also emit its own turn frames (and accordingly, advertises itself as `emits_user_turn_frames=False`)
  - Alternatively: it still *does* emit its own turn frames, and continues to advertise itself as `emits_user_turn_frames=True`. The downside of this alternative is that turn frames may be a smidge slower coming from the service than from the local strategies...though arguably more "accurate"?

WDYT?

kompfner: Hmm, it does look like my change to automatically append the external turn strategies isn't working for the locally-driven-turns scenario...will have to rethink it.

src/pipecat/processors/aggregators/llm_response_universal.py
```diff
@@ -606,8 +669,43 @@ def __init__(
...
+        # assistant response start as the "user turn ended" signal for
+        # context-writing purposes, then waits up to this many seconds
+        # for any in-flight ``TranscriptionFrame`` to land before
+        # flushing — mirroring the cascade-mode pattern of waiting up
+        # to ``ttfs_p99_latency`` after ``UserStoppedSpeakingFrame``.
+        # Captured from the most recent ``STTMetadataFrame`` (which a
+        # realtime LLM service may broadcast for its internal STT
+        # pass), falling back to ``DEFAULT_TTFS_P99``.
+        self._ttfs_p99_latency: float | None = None
+        self._realtime_handoff_flush_task: asyncio.Task | None = None
```
filipi87, line 702: Hi @kompfner, if the goal of this new mode is to invoke `on_user_turn_message_added` only once, couldn't we invoke `_realtime_handoff_flush_immediate` when we receive the first message from the next assistant turn. At that point, we would know that all user transcriptions have already been received.

If that’s the case, we could potentially remove all the new tasks and p99 logic added here. Because, as far as I understand, the only moment when we can be sure that no additional user transcriptions will arrive is when we receive the next assistant message.

Or am I missing something ?

filipi87: Like, maybe invoke it inside `_handle_text`, in the assistant aggregator  ?

src/pipecat/services/inworld/realtime/llm.py
```diff
@@ -923,12 +956,14 @@ async def _handle_evt_function_call_arguments_done(self, evt):
 
     async def _handle_evt_speech_started(self, evt):
         """Handle speech started event from server-side VAD."""
+        print(f"[pk] Speech started event received: {evt}")
```
filipi87: nit: remove the `print`.

src/pipecat/services/inworld/realtime/llm.py
```diff
@@ -923,12 +956,14 @@ async def _handle_evt_function_call_arguments_done(self, evt):
...
         await self._truncate_current_audio_response()
         await self.broadcast_frame(UserStartedSpeakingFrame)
         await self.broadcast_interruption()
 
     async def _handle_evt_speech_stopped(self, evt):
         """Handle speech stopped event from server-side VAD."""
+        print(f"[pk] Speech stopped event received: {evt}")
```
filipi87: nit: remove the `print`.

src/pipecat/services/xai/realtime/llm.py
```diff
@@ -680,9 +694,22 @@ async def _receive_task_handler(self):
...
+                if (
+                    evt.error.code
+                    in (
+                        "response_cancel_not_active",
+                        "conversation_already_has_active_response",
+                    )
+                    or "no active response" in msg
+                    or "already has an active response" in msg
                 ):
                     logger.debug(f"{self} {evt.error.message}")
```
filipi87, line 719: Just double-checking: all this logic exists because, in those cases, `self._handle_evt_error` should not be invoked. Is that correct?

src/pipecat/services/llm_service.py
```diff
@@ -495,6 +549,23 @@ async def push_frame(self, frame: Frame, direction: FrameDirection = FrameDirect
...
+        if (
+            self._realtime_service_info is not None
+            and isinstance(frame, StartFrame)
+            and direction == FrameDirection.DOWNSTREAM
+        ):
+            await self.broadcast_frame(
+                RealtimeServiceMetadataFrame,
+                service_name=self.name,
+                emits_user_turn_frames=self._realtime_service_info.emits_user_turn_frames,
+            )
```
filipi87: nit: wouldn't be better if we moved this inside `async def start(self, frame: StartFrame):` ?

src/pipecat/services/inworld/realtime/llm.py
```diff
@@ -928,12 +961,20 @@ async def _handle_evt_function_call_arguments_done(self, evt):
 
     async def _handle_evt_speech_started(self, evt):
         """Handle speech started event from server-side VAD."""
+        if self._is_manual_turn_detection():
+            # In manual mode, the client is responsible for broadcasting user turn frames
```
kompfner, line 970: In my testing I found that Inworld Realtime still sent speech started/stopped events when configured with manual turn detection, so this fix was necessary. OpenAI Realtime does not send those events with its manual turn detection mode enabled.

A TODO is to check what Grok Realtime does...I haven't been able to test it for the last day or so due to 429 errors.

changelog/4533.added.4.md
kompfner: Done.

changelog/4533.changed.2.md
kompfner: Here's what I ended up going with: https://github.com/pipecat-ai/pipecat/pull/4533/changes/bca7642ada1779e21df92188dda801395e36671a (plus a little follow-up here: https://github.com/pipecat-ai/pipecat/pull/4533/changes/c9cf1aca1749c7698ec609c0ecb18221fb2ac509). Let me know what you think!

The new mechanism is also now explained in 4533.added.md.

examples/realtime/realtime-aws-nova-sonic.py
kompfner: We discussed this in person and I _think_ were in agreement about the middle-ground approach (warning). Let me know if you disagree!

src/pipecat/frames/frames.py
kompfner: Resolving this comment thread - see [the other one](https://github.com/pipecat-ai/pipecat/pull/4533#discussion_r3311877829) for discussion.

src/pipecat/processors/aggregators/llm_response_universal.py
kompfner: As discussed in person: sadly, we *can* receive bits of user transcript even after we've started receiving the assistant response 🤦. They can be interleaved. Protecting against that scenario is the purpose of this mechanism.

Essentially, the mechanism is:
- Assistant response received -> we now know definitively that the user turn has ended on the server side, for the purposes of context-writing -> we start a timer to provide breathing room for late transcript chunks to trickle in
- Timer expires -> flush aggregated transcript to context
- Assistant response ended -> juuuuust in case the timer hasn't expired, end it and flush aggregated transcript to context (it has to land in the context before the assistant response)

src/pipecat/processors/aggregators/llm_response_universal.py
kompfner: Fixed.

src/pipecat/processors/aggregators/llm_response_universal.py
kompfner: Let me know if this still makes sense to you, then I'll resolve this comment thread.

src/pipecat/services/xai/realtime/llm.py
kompfner: It's mostly to avoid the `return`, which exits the receive loop and causes the conversation to end. We were seeing this happen. 

Honestly, this was just a quick, relatively thoughtless fix to coax this code to match the clear *intent* of this code, which was to ignore these two errors. Grok doesn't seem to be using meaningful error codes like OpenAI Realtime (which the Grok Realtime code is based on), so we're falling back to checking for message strings (while keeping the code checks in place just in case they're used going forward—after all, they probably _should_ be, since I believe Grok Realtime is meant to be compatible with the OpenAI Realtime interface).

src/pipecat/services/inworld/realtime/llm.py
kompfner: Was able to run Grok Realtime today (thanks, @markbackman!), and discovered that it needed the same fix.

src/pipecat/services/llm_service.py
kompfner: An LLM service's `start()` runs *before* it does `push_frame(StartFrame)`. We want to ensure that the `RealtimeServiceMetadataFrame` broadcast happens *after* the `StartFrame` is pushed.

That said...this pattern here is a bit wonky. Maybe we should follow the metadata-frame-push pattern established in `STTService`...

kompfner: Ah, here's the problem: the `LLMService` "leaf" implementations all do a blanket `push_frame` of every frame that the service doesn't handle. `STTService` implementations don't do this, so the parent class `STTService` can properly handle the push-start-frame-then-broadcast-metadata pattern in its `process_frame`. If we wanted to do it in `process_frame` here, we'd either have to do it in every `LLMService` subclass, or we'd end up with duplicate frames.

kompfner: Improved the comment.

changelog/4533.changed.2.md
filipi87: Yep. I think the approach that you implemented makes sense. 🚀

examples/realtime/realtime-aws-nova-sonic.py
filipi87: I think it’s fine. Like you said, maintaining backward compatibility is more important right now.

src/pipecat/processors/aggregators/llm_response_universal.py
filipi87: Yes, I added those suggestions because, based on previous discussions, we thought interleaved messages should not be possible. It’s really weird that this can happen. 🤕

And yes, this approach makes sense considering that.

filipi87 (approved):
Amazing work @kompfner! 👏🚀

PR #4535, opened 2026-05-21
Update Smallest AI TTS plugin for Waves v4.0.0

harshitajain165:
## Summary

- **New WebSocket endpoint**: `/waves/v1/tts/live` (previously `/waves/v1/{model}/get_speech/stream`)
- **Model moved to message payload**: model is no longer part of the URL path — it's now sent as a field in each request message, removing the need to reconnect on model change
- **Updated model names**: `lightning_v3.1` and `lightning_v3.1_pro` (underscore convention); removed deprecated `lightning-v2` and its exclusive fields (`consistency`, `similarity`, `enhancement`)
- **New `output_format` setting**: supports `pcm`, `mp3`, `wav`, `ulaw`, `alaw`
- **Default model is now `lightning_v3.1_pro`**: with `meher` as its default voice; `lightning_v3.1` defaults to `sophia`; explicit voice in user settings always takes precedence

## Breaking changes

- `SmallestTTSModel.LIGHTNING_V2` removed — users on `lightning-v2` must migrate to `lightning_v3.1` or `lightning_v3.1_pro`
- `SmallestTTSSettings` fields `consistency`, `similarity`, `enhancement` removed (were v2-only)
- `SmallestTTSModel.LIGHTNING_V3_1` value changed from `"lightning-v3.1"` to `"lightning_v3.1"` — users hardcoding the string value will need to update
- Default model changed from `lightning_v3.1` to `lightning_v3.1_pro`

## Test plan

- [X] Unit tests covering default voice selection per model, WebSocket URL, message payload fields
- [X] Integration tests verified against live Waves v4.0.0 API for both `lightning_v3.1` and `lightning_v3.1_pro`


src/pipecat/services/smallest/tts.py
```diff
@@ -47,8 +47,14 @@
 class SmallestTTSModel(StrEnum):
     """Available Smallest AI TTS models."""
 
-    LIGHTNING_V2 = "lightning-v2"
-    LIGHTNING_V3_1 = "lightning-v3.1"
+    LIGHTNING_V3_1 = "lightning_v3.1"
```
markbackman, line 50: Just noting that the model string changed. I'm assuming this is expected?

src/pipecat/services/smallest/tts.py
```diff
@@ -90,16 +96,13 @@ class SmallestTTSSettings(TTSSettings):
     """Settings for SmallestTTSService.
 
     Parameters:
-        speed: Speech speed multiplier.
-        consistency: Consistency level for voice generation (0-1).
-        similarity: Similarity level for voice generation (0-1).
-        enhancement: Enhancement level for voice generation (0-2).
+        speed: Speech speed multiplier (0.5–2.0).
+        output_format: Audio output format. One of ``pcm``, ``mp3``, ``wav``,
```
markbackman: The output_format should be in __init__ as it's an init time config only. I would recommend defaulting to `pcm` which is what Pipecat expects.

markbackman (commented):
In addition, please add a changelog entry:

src/pipecat/services/smallest/tts.py
harshitajain165: Yes @markbackman, this is expected

src/pipecat/services/smallest/tts.py
harshitajain165: output_format has been moved into __init__ as a fixed parameter, defaulting to pcm

markbackman (approved):

PR #4536, opened 2026-05-21
Vendor Whisper log-mel feature extractor in Smart Turn v3

Salman778:
Smart Turn v3 used `transformers.WhisperFeatureExtractor` only for the log-mel features it feeds to the ONNX model. The underlying numpy path is small and self-contained, so this change vendors a minimal numpy implementation into pipecat and removes the transformers import from the v3 module.

We noticed Smart Turn v3 was using significantly more memory than expected at import time, almost entirely from pulling in the full `transformers` package for a single feature extractor. Measured on this branch (Python 3.12, Linux):

  | Metric                            | Before    | After    |
  | --------------------------------- | --------- | -------- |
  | `import local_smart_turn_v3` time |  5,044 ms |   300 ms |
  | Peak RSS                          |    566 MB |    60 MB |

That is ~17x faster cold-start and ~506 MB less resident memory per process, which is most impactful under pod-per-call deployments where every container pays the import cost.

Per-call CPU and event-loop behavior are unaffected. `_predict_endpoint` remains synchronous and runs off the main loop via the analyzer thread pool, exactly as before; only the implementation of the feature extractor changed.

Behavior is verified numerically equivalent: the vendored output matches the reference numpy code path within 1e-5 absolute tolerance across silence/noise/sine-sweep/partially-padded audio (both `do_normalize=True` and `=False`), and the ONNX model produces bit-identical probabilities on representative inputs.

Smart Turn v3 no longer imports `transformers` at module load. The `transformers` package remains a project dependency because the deprecated v2/CoreML analyzers and the Moondream service still use it; making it optional is out of scope here.

### DIFF
--- a/changelog/+smart-turn-v3-vendor-whisper-features.changed.md
+++ b/changelog/+smart-turn-v3-vendor-whisper-features.changed.md
@@ -1,0 +1,4 @@
+- Replaced the `transformers.WhisperFeatureExtractor` dependency in `LocalSmartTurnAnalyzerV3` with a vendored numpy-only implementation, reducing peak RSS at import from ~566 MB to ~60 MB and cold-start time from ~5.0 s to ~0.3 s. Behavior is numerically equivalent (matches the reference numpy code path within 1e-5 absolute tolerance; ONNX model output is bit-identical on representative inputs).
+  - Smart Turn v3 no longer imports `transformers` at module load.
+  - Prepares the ground for making `transformers` an optional dependency in a future release.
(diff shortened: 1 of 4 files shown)


markbackman:
@marcus-daily I believe you've thought about removing this dependency as well. Can you please review / add your thoughts?

marcus-daily:
Thanks @Salman778, this is exactly what we've been intending to do with the feature extractor. I'll take a closer look at this soon so we can get it merged.

marcus-daily (approved):

marcus-daily:
Now merged, thanks again @Salman778!

PR #4537, opened 2026-05-21
fix(tracing): accumulate text across run_tts calls when context is re…

monster-anshu:

With reuse_context_id_within_turn=True (the default), multiple run_tts calls within a turn share the same context_id and therefore the same TTS span. attach_run_tts_attributes was calling span.set_attribute("text") on every call, overwriting the previous chunk so only the last synthesised sentence appeared in the trace.

Fix: store accumulated text in the span entry dict and append each new chunk before setting the span attribute, so the span captures the full turn text regardless of how many run_tts calls were made.

### DIFF
--- a/src/pipecat/utils/tracing/service_decorators.py
+++ b/src/pipecat/utils/tracing/service_decorators.py
@@ -364,7 +364,14 @@ def traced_tts(func: Callable | None = None, *, name: str | None = None) -> Call
             owner.setup = patched_setup
 
         def attach_run_tts_attributes(service, text, args, kwargs):
-            """Attach text-specific attributes to the in-flight TTS span."""
+            """Attach text-specific attributes to the in-flight TTS span.
+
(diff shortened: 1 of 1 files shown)


filipi87:
Hi @monster-anshu, thanks for this PR!

Looks like Pipecat v1.6.0 already includes this — see the changelog:

> Fixed TTS spans in OpenTelemetry traces missing some or all of the spoken text (regression in 1.2.0): with sentence aggregation, a turn's span only kept the last sentence's `text` and `metrics.character_count`, and for services that create their audio context inside `run_tts` (e.g. the ElevenLabs websocket path) the first sentence was dropped — leaving single-sentence turns with no text at all. Text from every `run_tts` call in an audio context now accumulates onto the span (joined text, summed character count), including text spoken before an interruption. (PR #5049)

Closing since this looks like it's already covered. Feel free to reopen if this PR does something additional not covered there.

PR #4539, opened 2026-05-21
Update inception realtime setting

markbackman:

Per Inception team, the recommendation is to enable the `realtime` setting for lower TTFT latency. Also, use `reasoning` as `low` for the function-calling-inception.py example for more reliable function calling.


filipi87 (approved):
LGTM!

PR #4542, opened 2026-05-22
Smart turn v3 perf fixes

Salman778:
### Summary

Three small, independent changes to Smart Turn v3 targeting the per-frame hot path and the per-turn resample step.

| Commit | Change | Effect |
| --- | --- | --- |
| `e59ac39` | `soxr` quality `VHQ` → `HQ` in `LocalSmartTurnAnalyzerV3._resample_to_model_rate` | 30–50% less resample CPU per turn on non-16 kHz inputs (~3–10 ms). No-op at 16 kHz — the existing fast path returns early without calling soxr. |
| `7a48f52` | `time.time()` → `time.monotonic()` at three sites in `BaseSmartTurn.append_audio` | Uses the correct primitive for interval math (buffer trim window, speech-start window). Wall-clock time can step under NTP correction; monotonic cannot. Existing `time.perf_counter()` sites are already monotonic and unchanged. |
| `6bef0f3` | Defer int16 → float32 conversion until segment extraction | Buffer holds int16 PCM views; conversion runs once per turn on the concatenated segment instead of once per ~20 ms frame. Eliminates ~50 numpy allocations/sec per analyzer. Bit-identical output (int16 → float32 / 32768 distributes over concatenation; verified by test). |

Plus a fourth commit with three towncrier changelog fragments.

### Rationale

**`soxr` VHQ → HQ.** VHQ is soxr's highest preset (~26-tap polyphase). For a binary classifier on Whisper log-mel features the 80-band filterbank already smears spectral detail, so VHQ's marginal gain over HQ (~16-tap) sits below the noise floor of the feature representation. Most pipecat deployments deliver 16 kHz audio and are unaffected; the win applies to 24/48 kHz sources.

**`time.monotonic()`.** Both affected sites compute durations that never leave the process: the pre-speech trim loop pops entries older than `now - max_buffer_time`, and `speech_start_time` is later used to compute the segment window. Both are pure interval math, which is the case `time.monotonic()` exists for.

**Deferred int16 → float32.** `append_audio` previously allocated a fresh float32 ndarray per frame; at 16 kHz / 20 ms-per-frame that's ~50 allocations/sec per analyzer, all held in `self._audio_buffer` until the turn closes. Keeping the buffer in int16 and casting once at segment time is bit-identical because the cast and the `/ 32768.0` divide distribute over concatenation. Also drops a spurious `np.frombuffer(audio_int16, dtype=np.int16)` re-wrap on the line below — it was a no-op view-of-a-view.

### Subclass impact

All `BaseSmartTurn` subclasses (`LocalSmartTurnAnalyzerV3`, `LocalCoreMLSmartTurnAnalyzer`, `HttpSmartTurnAnalyzer`) receive `audio_array` through the parent's `_process_speech_segment`, which converts the concatenated segment to float32 before calling `_predict_endpoint`. The contract that `audio_array` is float32 in `[-1, 1]` is preserved exactly. None of the subclasses read `self._audio_buffer` directly.

### Scope

Deliberately small — per-frame allocation and per-turn resample only. Larger architectural changes are out of scope here.

### Test plan

- `uv run pytest tests/test_local_smart_turn_v3_features.py` — Whisper feature parity unchanged (19 cases)
- `uv run pytest tests/test_local_smart_turn_v3_resample.py` — new: HQ vs VHQ output equivalence on a synthesized 48 kHz signal; 16 kHz fast-path returns input unchanged; 48 kHz end-to-end smoke (3 cases)
- `uv run pytest tests/test_base_smart_turn_buffer.py` — new: bit-identical equivalence between eager and deferred float32 conversion; buffer holds int16 after `append_audio`; segment reaching `_predict_endpoint` is normalized float32; pre-speech trim still bounds growth (4 cases)
- `uv run pytest tests/` — full suite: 1389 passing, 4 skipped

One unrelated pre-existing failure in `tests/test_service_init.py::test_service_settings_complete[NvidiaSegmentedSTTService]` also fails on `main` — flagged separately.

### Risk

Low. The two changes touching shared state in `base_smart_turn.py` preserve the float32 contract observed by all three subclasses. The `soxr` change is a no-op on 16 kHz inputs. The monotonic change is a strict primitive correction.


markbackman:
cc @marcus-daily for an opinion.

marcus-daily:
Looks good! Looks like minor fixes needed for the CI.

markbackman:
Looks good to me! Leaving this to @marcus-daily for approval.

Salman778:
Hi @marcus-daily, gentle ping on this when you have a moment. CI is green after the ruff I001 fix in 44745cf, and @markbackman signed off pending your approval.

marcus-daily (approved):

marcus-daily:
Thanks @Salman778, I've taken a more detailed look now. Approved and merged!

PR #4544, opened 2026-05-22
Allowing to change the declared sample rate to send audio faster than realtime.

filipi87:
## Summary
  
- Adds `audio_out_declared_sample_rate` to `DailyParams`, allowing the Daily `CustomAudioSource` to be declared at a different rate than the true TTS output rate
- When set, writing true-rate PCM (e.g. 24 kHz) into a source declared at a higher rate (e.g. 48 kHz) causes write_frames() to return faster than real-time, delivering audio to the WebRTC peer ahead of schedule
- Sends an `audio.faster_than_realtime` RTVI app message to each joining participant carrying `true_sample_rate` and `declared_sample_rate`, so the web client can set up the correct playback pipeline

## Notes

This is a proof of concept for faster-than-realtime audio delivery. The web client must be configured with `fasterThanRealtime: true` on `DailyTransport` to decode the stream at the correct pitch. Pair with `audio_out_prebuffer_secs` (on the base TransportParams) to absorb TTS jitter before the first frame is sent.

## Related PRs
- Daily web transport [PR](https://github.com/pipecat-ai/pipecat-client-web-transports/pull/129).
- New example [PR](https://github.com/pipecat-ai/pipecat-examples/pull/213).

### DIFF
--- a/src/pipecat/transports/daily/transport.py
+++ b/src/pipecat/transports/daily/transport.py
@@ -324,6 +324,14 @@ class DailyParams(TransportParams):
         api_url: Daily API base URL.
         api_key: Daily API authentication key.
         audio_in_user_tracks: Receive users' audio in separate tracks
+        audio_out_declared_sample_rate: When set, the microphone CustomAudioSource is declared
+            at this rate instead of the true output sample rate. Writing true-rate PCM bytes
+            into a source declared at a higher rate (e.g. 48000 when content is 24000 Hz) causes
+            write_frames() to return faster than real-time, delivering audio to the WebRTC peer
+            ahead of schedule. The web client must be configured with ``fasterThanRealtime: true``
+            to receive and play back the audio at the correct pitch. Pair with
+            ``audio_out_prebuffer_secs`` to absorb TTS jitter before the first frame is sent.
+            Defaults to ``None`` (disabled).
         camera_out_enabled: Whether to enable the main camera output track.
         camera_out_send_settings: Camera output track publishing settings.
         custom_audio_track_params: Per-destination configuration for custom audio tracks.
@@ -337,6 +345,7 @@ class DailyParams(TransportParams):
     api_url: str = "https://api.daily.co/v1"
     api_key: str = ""
     audio_in_user_tracks: bool = True
+    audio_out_declared_sample_rate: int | None = None
     camera_out_enabled: bool = True
     camera_out_send_settings: dict[str, Any] | None = None
     custom_audio_track_params: Mapping[str, DailyCustomAudioTrackParams] | None = None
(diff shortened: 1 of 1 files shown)


mattieruth (commented):
what is the benefit of sending audio faster? you still play it out at the expected rate. i know i'm missing something.

as for the actual changes here, my only main concern is the format of the rtvi message. we have not used the format of `x.y` for things so it feels a bit inconsistent. I think a higher-level type like: `track-settings-update` might make sense with data being:

```
type: 'track-settings-update'
data: { <track name>: <settings> }
```

ex. (theoretical, to show how other settings could be sent):
```
{
  audio: { true_sample_rate: 96000, declared_sample_rate: 44100 },
  screenVideo: { screenPixelRatio: .5 },
  myCustomAudio: { true_sample_rate: 96000, declared_sample_rate: 44100 }
}
```

what do you think?

src/pipecat/transports/daily/transport.py
```diff
@@ -2949,6 +2964,18 @@ async def _on_participant_joined(self, participant):
         logger.info(f"Participant joined {id}")
 
         if self._input and self._params.audio_in_enabled and self._params.audio_in_user_tracks:
+            if self._params.audio_out_declared_sample_rate:
+                self._client._client.send_app_message(
```
mattieruth, line 2968: this would ideally be run through the RTVI Observer and the message would be defined and typed with the RTVI messages

mattieruth: not only would that be consistent with all other messaging, but it would allow us to check the client RTVI version for compatibility as well.

mattieruth: also, this would require a minor version bump to the RTVI version

filipi87:
Hi @mattieruth, this was more of a proof of concept. The idea originated from a requirement in Pipecat → `daily-python`, where the `daily-python` client ideally needed to receive audio faster than real time in order to buffer it and pre process video avatar images, so that both the image and audio would be ready to send back to the user at the playback rate.

This would give us the advantage of being able to pre process the audio before using it downstream.

Then, after discussing it with Kwin, we thought it would be interesting to test this in the browser to see whether it could help compensate for temporary poor network conditions.

In the end, though, there are a couple of limitations, which I detailed in the example:
- https://github.com/pipecat-ai/pipecat-examples/tree/filipi/daily_faster_than_realtime.py/faster-than-realtime#known-limitations

That is also why I haven’t put much effort into designing a proper RTVI protocol for it. I left it as a draft so we would know that there is a viable way to make it work, along with what it would require.

I also explained all the findings in detail in a Slack thread, I will share with you. 🙂

mattieruth:
> Hi @mattieruth, this was more of a proof of concept. The idea originated from a requirement in Pipecat → `daily-python`, where the `daily-python` client ideally needed to receive audio faster than real time in order to buffer it and pre process video avatar images, so that both the image and audio would be ready to send back to the user at the playback rate.
> 
> This would give us the advantage of being able to pre process the audio before using it downstream.
> 
> Then, after discussing it with Kwin, we thought it would be interesting to test this in the browser to see whether it could help compensate for temporary poor network conditions.
> 
> In the end, though, there are a couple of limitations, which I detailed in the example:
> 
> * https://github.com/pipecat-ai/pipecat-examples/tree/filipi/daily_faster_than_realtime.py/faster-than-realtime#known-limitations
> 
> That is also why I haven’t put much effort into designing a proper RTVI protocol for it. I left it as a draft so we would know that there is a viable way to make it work, along with what it would require.
> 
> I also explained all the findings in detail in a Slack thread, I will share with you. 🙂

thank you for the explanation. i had just gotten to the comments in the example and my next question was going to be... "yikes. that's a lot of limitations. is this really something we should support at such a core level?" 😆

PR #4545, opened 2026-05-22
fix: make test_task_heartbeats robust to timing jitter

markbackman:
This test keeps failing in CI. This PR is to make the timing more deterministic.

## Summary

- Fixes flaky `tests/test_pipeline.py::TestPipelineTask::test_task_heartbeats` (`assert 4 == 5.0` in CI, passes locally).
- Root cause: the test asserted an **exact** count of timer-driven heartbeats over a fixed 1.0s wall-clock window, so the result depended on event-loop scheduling precision.
- Test-only change — no library code is affected.

## Details

The test asserted an exact heartbeat count inside a fixed 1.0s window:

```python
expected_heartbeats = 1.0 / 0.2   # == 5.0
await asyncio.wait_for(task.run(...), timeout=1.0)
assert heartbeats_counter == expected_heartbeats
```

The heartbeat handler pushes one `HeartbeatFrame` immediately, then `asyncio.sleep(0.2)` between each, so heartbeats land at roughly `ε, ε+0.2, …, ε+0.8, ε+1.0` where `ε` is startup latency. Whether exactly 5 fall inside the 1.0s window depends entirely on scheduling precision — fine locally, but under CI load a tick slips past the cutoff and only 4 are observed.

The fix stops coupling the assertion to the wall-clock window. It waits until the expected number of heartbeats has actually been observed (with a generous 5s safety-net timeout that only triggers if heartbeats never fire), then stops the pipeline and asserts:

- `heartbeats_counter >= expected_heartbeats` — they fire and reach the observer.
- `elapsed >= (expected_heartbeats - 1) * period_secs` — they're properly paced. Since `asyncio.sleep` is a guaranteed lower bound, N heartbeats span at least `(N−1)` periods. This is immune to scheduling jitter (which can only delay, never speed up) while still catching heartbeats that fire too fast.

This preserves the original two-sided intent (right number, right rate) without depending on timer precision.

## Testing

- Passes consistently (~1.3s), including under heavy CPU contention where the original off-by-one surfaced.
- Full `TestPipelineTask` class passes (19 tests).

### DIFF
--- a/tests/test_pipeline.py
+++ b/tests/test_pipeline.py
@@ -353,11 +353,16 @@ class TestPipelineTask(unittest.IsolatedAsyncioTestCase):
         assert "Second" in upstream_texts
 
     async def test_task_heartbeats(self):
+        period_secs = 0.2
+        expected_heartbeats = 5
         heartbeats_counter = 0
+        received_expected = asyncio.Event()
 
         async def heartbeat_received(processor: FrameProcessor, heartbeat: HeartbeatFrame):
             nonlocal heartbeats_counter
             heartbeats_counter += 1
+            if heartbeats_counter >= expected_heartbeats:
+                received_expected.set()
 
         identity = IdentityFilter()
         pipeline = Pipeline([identity])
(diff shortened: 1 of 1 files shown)


filipi87 (approved):
LGTM!

PR #4549, opened 2026-05-22
feat(azure): add private_endpoint support to Azure TTS services

brunomartinez-lmi:
## Summary

Add `private_endpoint` parameter to `AzureTTSService` and `AzureHttpTTSService`, enabling connection via Private Link or custom domain endpoints.

## Motivation

`AzureSTTService` already supports a `private_endpoint` parameter (added in PR #3967 and refined in PR #3974), but the TTS services still require a region-based endpoint. Users deploying Azure Speech resources with custom domains or Private Link must currently subclass and monkey-patch the `SpeechConfig` after initialization.

This PR brings TTS to parity with STT by passing `private_endpoint` directly to Azure SDK's `SpeechConfig(endpoint=...)` constructor — the same pattern used in the STT service.

## Changes

- Added `private_endpoint` parameter to `AzureBaseTTSService._init_azure_base()`
- Added `private_endpoint` parameter to `AzureTTSService.__init__()`
- Added `private_endpoint` parameter to `AzureHttpTTSService.__init__()`
- Updated `start()` methods to create `SpeechConfig` with `endpoint=` when `private_endpoint` is provided (matching the STT implementation pattern)
- Made `region` optional (required only when `private_endpoint` is not provided)
- Added validation: raises `ValueError` if neither `region` nor `private_endpoint` is given
- Added warning when both are provided (region is ignored, matching STT behavior)

## Usage

```python
# Using private endpoint (Private Link or custom domain)
tts = AzureTTSService(
    api_key="...",
    private_endpoint="https://my-resource.cognitiveservices.azure.com/",
    settings=AzureTTSService.Settings(voice="en-US-SaraNeural"),
)

# Existing region-based usage continues to work unchanged
tts = AzureTTSService(
    api_key="...",
    region="eastus",
    settings=AzureTTSService.Settings(voice="en-US-SaraNeural"),
)
```

## References

- Related: PR #3967 (Fix Azure STT SpeechConfig with private endpoint)
- Related: PR #3974 (Make Azure STT region optional when private_endpoint is used)
- Azure docs: https://learn.microsoft.com/en-us/azure/ai-services/speech-service/speech-services-private-link

🤖 generated by copilot, pre-reviewed by me


src/pipecat/services/azure/tts.py
```diff
@@ -374,10 +389,20 @@ async def start(self, frame: StartFrame):
...
         # Now self.sample_rate is properly initialized
-        self._speech_config = SpeechConfig(
-            subscription=self._api_key,
-            region=self._region,
-        )
+        if self._private_endpoint:
+            if self._region:
+                logger.warning(
+                    "Both 'region' and 'private_endpoint' provided; 'region' will be ignored."
+                )
```
markbackman: Consider moving this warning to `__init__` (right next to the `ValueError` check), so it fires exactly once at construction time instead of every time `start()` is called. The check only depends on `region` and `private_endpoint`, both known at init time, and this matches the pattern used in `AzureSTTService.__init__` where both the `ValueError` and the warning live together.

src/pipecat/services/azure/tts.py
```diff
@@ -861,10 +895,20 @@ async def start(self, frame: StartFrame):
...
 
-        self._speech_config = SpeechConfig(
-            subscription=self._api_key,
-            region=self._region,
-        )
+        if self._private_endpoint:
+            if self._region:
+                logger.warning(
+                    "Both 'region' and 'private_endpoint' provided; 'region' will be ignored."
+                )
```
markbackman: Same suggestion as above: move this warning to `__init__` alongside the `ValueError` for parity with `AzureSTTService` and so it only fires once per instance.

src/pipecat/services/azure/tts.py
markbackman: Maybe something like:
```
if not region and not private_endpoint:
    raise ValueError("Either 'region' or 'private_endpoint' must be provided.")
if region and private_endpoint:
    logger.warning(
        "Both 'region' and 'private_endpoint' provided; 'region' will be ignored."
    )
```

markbackman (approved):
LGTM!

PR #4551, opened 2026-05-22
Honor SOXR resampler quality setting

MHamdyK:
## Summary

This makes the SOXR resamplers honor the configured quality setting instead of always using VHQ.

The default remains VHQ, so existing callers keep the same behavior unless they explicitly pass another quality preset.

## Details

RNNoiseFilter already passes resampler_quality into SOXRStreamAudioResampler, defaulting to QQ for lower-latency real-time processing. However, the stream resampler ignored the argument and always initialized SOXR with VHQ.

This updates both SOXR resampler implementations to store and pass through the selected quality preset.

## Tests

- uv run pytest tests/test_soxr_resamplers.py
- uv run pytest tests/test_soxr_resamplers.py tests/test_rnnoise_resampling.py tests/test_rnnoise_filter.py tests/test_resampy_resampler.py
- uv run ruff check src/pipecat/audio/resamplers/soxr_resampler.py src/pipecat/audio/resamplers/soxr_stream_resampler.py src/pipecat/audio/resamplers/soxr_utils.py tests/test_soxr_resamplers.py

### DIFF
--- a/changelog/4551.fixed.md
+++ b/changelog/4551.fixed.md
@@ -1,0 +1,5 @@
+- Fixed `SOXRAudioResampler` and `SOXRStreamAudioResampler` ignoring the
+  configured quality setting. Both resamplers were hardcoded to `VHQ`, which
+  meant `RNNoiseFilter`'s `resampler_quality` argument (defaulting to `QQ`
+  for low-latency real-time use) had no effect. The resamplers now honor
+  the configured quality, with `VHQ` retained as the default.
--- a/src/pipecat/audio/filters/rnnoise_filter.py
+++ b/src/pipecat/audio/filters/rnnoise_filter.py
@@ -14,6 +14,7 @@ import numpy as np
 from loguru import logger
 
 from pipecat.audio.filters.base_audio_filter import BaseAudioFilter
+from pipecat.audio.resamplers.base_audio_resampler import SoxrQuality
 from pipecat.frames.frames import FilterControlFrame, FilterEnableFrame
 
 try:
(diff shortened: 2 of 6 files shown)


markbackman:
Nice catch on the root cause. Threading `quality` through instead of hardcoding `VHQ` is exactly right.

One suggestion: soxr already validates the preset for both entry points, so `validate_soxr_quality()` mostly duplicates it:

```python
>>> soxr.resample(x, 8000, 16000, quality="BAD")
ValueError: Quality must be one of [QQ, LQ, MQ, HQ, VHQ]
```

Two downsides to the hand-rolled check: it's stricter than soxr (soxr accepts lowercase like `"hq"`, the validator rejects it), and it duplicates the preset list between the `Literal` and `SOXR_QUALITY_PRESETS`.

I'd keep the `Literal` for the type hint and drop `validate_soxr_quality` + `soxr_utils.py`. The `Literal` gives static/IDE safety; soxr gives runtime safety with a clear message. The shared `SoxrQuality` alias can just live in `base_audio_resampler.py`. That also lets you drop `test_soxr_resampler_rejects_invalid_quality` (it'd be testing soxr, not us); the forwarding/default tests are the ones worth keeping.

Orthogonal: `RNNoiseFilter.resampler_quality` is typed `str`, so it skips the `Literal`'s static check. Retyping it to `SoxrQuality` buys more than the runtime validator does.

---

Also, please lint the code and typecheck:
```
uv run scripts/fix-ruff-and-typecheck.sh
```

MHamdyK:
Good points. Dropped validate_soxr_quality and removed soxr_utils.py, and moved the SoxrQuality alias into base_audio_resampler.py. soxr handles the runtime check and Literal covers the static side, so the hand-rolled validator was just adding a stricter, inconsistent path. Also removed the invalid-preset test since it was really exercising soxr rather than our code, and retyped RNNoiseFilter.resampler_quality to SoxrQuality so the static check reaches that call site. Ran ruff and pyright (clean on the changed files) plus the resampler tests.

markbackman (approved):
LGTM! Thanks 🙌

PR #4552, opened 2026-05-23
Add Cartesia Turns streaming STT service

dakshdua:
`CartesiaTurnsSTTService` speaks the Cartesia Streaming ASR v2 turn-based WebSocket protocol at `/stt/turns/websocket`. The server drives turn boundaries via `turn.start` / `turn.update` / `turn.eager_end` / `turn.resume` / `turn.end` messages, which the service translates into `UserStartedSpeakingFrame`, `InterimTranscriptionFrame`, finalized `TranscriptionFrame`, and `UserStoppedSpeakingFrame`. Eager end-of-turn predictions are surfaced as `InterimTranscriptionFrame` frames for opt-in speculative downstream processing.


src/pipecat/services/cartesia/turns/stt.py
```diff
@@ -0,0 +1,471 @@
...
+try:
+    from websockets.asyncio.client import connect as websocket_connect
+    from websockets.protocol import State
+except ModuleNotFoundError as e:
+    logger.error(f"Exception: {e}")
+    logger.error("In order to use Cartesia, you need to `pip install pipecat-ai[cartesia]`.")
+    raise Exception(f"Missing module: {e}")
+
+
+CARTESIA_VERSION = "2026-03-01"
```
charlotte-zhuang: Can you inline this variable or make it private?

```suggestion
_CARTESIA_VERSION = "2026-03-01"
```

charlotte-zhuang (approved):

src/pipecat/services/cartesia/turns/stt.py
dakshdua: inlined!

examples/transcription/transcription-cartesia-manual.py
markbackman, line 1: In addition to these, you should think about adding an example to `examples/voice` for the new `CartesiaTurnsSTTService`. I'd suggest `examples/voice/voice-cartesia-turns.py` and you could add it to the evals, located at `scripts/evals/run-release-evals.py`.

src/pipecat/services/cartesia/turns/stt.py
```diff
@@ -0,0 +1,468 @@
...
+        if self._should_interrupt:
+            await self.broadcast_interruption()
+        await self.start_processing_metrics()
+        await self._call_event_handler("on_turn_start", transcript)
+
+    async def _handle_turn_update(self, data: dict):
+        transcript = data.get("transcript", "")
+        if transcript:
+            logger.trace(f"Cartesia Ink-2 ASR turn.update: {transcript}")
+        await self.push_frame(
```
markbackman, line 476: I see that there's an `InterimTranscriptionFrame` pushed for both `turn.update` and `turn.eager_end`. How are these different? Does it make sense to push one for each event?

src/pipecat/services/cartesia/turns/stt.py
```diff
@@ -0,0 +1,468 @@
...
+        )
+        await self._handle_transcription(transcript, True, self._language)
+        await self.stop_processing_metrics()
+        await self.broadcast_frame(UserStoppedSpeakingFrame)
+        await self._call_event_handler("on_turn_end", transcript)
+
+    async def _handle_server_error(self, data: dict):
+        message = data.get("message", "Unknown error")
+        error_code = data.get("error_code", "unknown")
+        logger.error(f"Cartesia Ink-2 ASR error [{error_code}]: {message}")
```
markbackman: We can remove this line:
```suggestion
```

`self.push_error` emits a `logger.error()` message.

markbackman (commented):
Looks good! Just a few items to clean up.

In addition to the inline comments, two more:
- After `turn.start`, if the client stops sending audio, does the server still emit `turn.end` on its own, or does it wait for more audio? (Trying to decide whether we need to send keepalive silence to avoid dangling turns, which we added for the Deepgram Flux service.)
- Also, does it make sense to update the model for the `CartesiaSTTService` to `ink-2`?

src/pipecat/services/cartesia/turns/stt.py
```diff
@@ -0,0 +1,468 @@
...
+        await self.push_error(error_msg=f"Cartesia Ink-2 ASR error: {message}")
```
markbackman:
```suggestion
        await self.push_error(error_msg=f"Cartesia Ink-2 ASR error [{error_code}]: {message}")
```

src/pipecat/services/cartesia/turns/stt.py
dakshdua: dropped!

examples/transcription/transcription-cartesia-manual.py
dakshdua: added!

src/pipecat/services/cartesia/turns/stt.py
dakshdua: just checked with the team and all updates should be included in `turn.update`, so we're not missing anything by not pushing an interim transcription on `turn.eager_end`, so I've dropped it and kept only the one for `turn.update`, which is sent periodically on transcript updates.

dakshdua:
@markbackman addressed all your inline comments.

For the question regarding `turn.end` when no audio is arriving, our STT works similarly to Deepgram Flux -- we do not send a `turn.end` without either audio or a `close` command. I've adapted the same logic as the Flux service to insert silence when there are gaps in client audio.

For `CartesiaSTTService`, we may not want to change the default model to `ink-2` as `ink-whisper` supports non-English languages and `ink-2` does not, which may surprise some users. Perhaps if there's a simple way to make the default `ink-2` only if the language is English?

markbackman:
## Code review

markbackman (commented):
LGTM! I have a few small things I'll clean up in a separate PR. If anything needs your attention, I'll ping you @dakshdua.

Thanks and congrats on the launch!

markbackman (approved):

PR #4557, opened 2026-05-25
Use the moq libraries instead of hand-rolling the wire codec

kixelated:
A pass on top of the POC swapping the hand-rolled wire codec on both sides for the upstream [`moq-rs`](https://pypi.org/project/moq-rs/) (Python) and [`@moq/publish`](https://www.npmjs.com/package/@moq/publish) (browser) libraries. Net **~−1,500 lines** vs. the POC.

## Benefits

- **Opus-encoded audio on the wire.** No more raw 16-bit PCM streaming. Bandwidth drops from ~384 kbps (24 kHz mono PCM) to ~32 kbps for the bot's audio, ~−12×. libopus also handles packet loss concealment, so dropouts sound like brief artifacts instead of clicks.

- **Battle-tested Rust QUIC stack.** [`moq-rs`](https://pypi.org/project/moq-rs/) wraps [moq-native](https://crates.io/crates/moq-native), which gets built-in caching, fan-out, prioritization, congestion control, and WebTransport ALPN negotiation. We were re-implementing slivers of all of these in Python on top of `aioquic`. Gone.

- **`max_latency_ms` SLA on the subscriber.** Default bumped to 500 ms (was 100 in the previous draft). Under healthy networks frames flow with no buffering; under congestion the consumer waits up to 500 ms for late groups to fill, then skips ahead. This is the structural win over WebRTC — there's no \"choose audio loss vs. stall,\" you set a deadline and the library obeys it. Tunable via `MOQParams.audio_in_max_latency_ms`.

- **No `moq-relay` process for local dev.** New `--moq-serve` flag binds the bot to a local UDP socket via `moq.Server`. The bot **is** the relay. `scripts/moq-dev-setup.sh` is deleted. One terminal, not two.

- **Self-signed certs minted in-process.** `tls_generate=[\"localhost\"]` produces a cert on startup; `server.cert_fingerprints()` exposes the SHA-256, the runner threads it back through `/start`, and `/api/config` hands it to the browser for WebTransport pinning. No PEM files, no symlinks across repos.

- **`@moq/publish` does browser audio capture.** Replaces ~250 lines of AudioWorklet + Int16 PCM + framing with `Publish.Broadcast({ audio: { source: mic.source } })`. We get Opus encoding via WebCodecs, catalog publishing, and the subscribe-fulfilment loop for free. AudioWorklet code now lives entirely upstream.

- **WebSocket fallback for free.** `@moq/net` races WebTransport and WebSocket — Firefox (no WebTransport yet) just works without the demo knowing.

- **Catalog-driven track discovery.** The bot writes a catalog from `publish_audio`'s `AudioEncoderInput`; the browser reads it from `subscribe(\"catalog.json\")` and uses whatever rate/channel count it advertises. Track names aren't pinned in the config — adding a video track or a screen-share later is a one-liner.

- **Better resampling.** moq-rs uses [rubato](https://docs.rs/rubato/) inside its `publish_audio` / `subscribe_audio` so you can hand it any sample rate and it does sinc-interpolation to/from libopus's native rates. The earlier draft used a hand-rolled 4-point linear interpolator that was fine for speech and dubious for anything else.

- **Single dependency.** `pyproject.toml` `[moq]` extra is now just `moq-rs>=0.2.13` and `cryptography`. `aioquic` and `opuslib` are both gone; libopus is statically bundled into the moq-rs wheel so there's no `brew install opus` / `apt-get install libopus0`.

- **Clean separation of concerns.** `transport.py` is down to ~700 lines — basically a thin glue layer between the pipecat Frame pipeline and the moq library. No varint codec, no stream-state machine, no codec helpers.

## What's still hand-rolled

The browser-side playback loop — `@moq/hang`'s `Container.Consumer` + WebCodecs `AudioDecoder` + Web Audio scheduling, about 60 lines. Could swap for `@moq/watch.Audio.Decoder` later, but that pulls in Sync/Source/jitter-buffer machinery that's overkill for a single-track playback path.

## Things to know before merging

1. The pipecat Client SDK transport plugin (`@pipecat-ai/moq-transport`) is still a future item, per [`moq_prebuilt/PLAN-moq-transport-package.md`](https://github.com/pipecat-ai/pipecat/blob/vp-moq-vibe/moq_prebuilt/PLAN-moq-transport-package.md). The browser demo here is still ad-hoc.

2. End-to-end testing requires `moq-rs 0.2.13` installed (auto via `uv sync --extra moq`). I haven't run the live mic-to-speaker round-trip in this branch — the smoke test is `uv run python examples/transports/transports-moq.py -t moq --moq-serve` and opening localhost:7860.

3. The browser loads `@moq/net`, `@moq/publish`, `@moq/hang`, `@moq/signals` from esm.sh. If that bothers you for production, the path is a tiny Vite build dropped into `moq_prebuilt/client/`. For dev / demo it works as-is.


aconchillo:
Hi @kixelated ! Awesome! Nice to see you around here! :rocket:

vipyne:
hey @kixelated, thanks for keeping up with this.  I'm going to try to port these changes to [https://github.com/pipecat-ai/pipecat/pull/4614](https://github.com/pipecat-ai/pipecat/pull/4614). 
(well, actually, I think they'd go in [pipecat-client-web-transports](https://github.com/pipecat-ai/pipecat-client-web-transports/tree/vp-add-moq-transport/transports/moq-transport) and/or [voice-ui-kit](https://github.com/pipecat-ai/voice-ui-kit/tree/vp-add-moq-transport), but yeah)

kixelated:
> hey @kixelated, thanks for keeping up with this. I'm going to try to port these changes to #4614. (well, actually, I think they'd go in [pipecat-client-web-transports](https://github.com/pipecat-ai/pipecat-client-web-transports/tree/vp-add-moq-transport/transports/moq-transport) and/or [voice-ui-kit](https://github.com/pipecat-ai/voice-ui-kit/tree/vp-add-moq-transport), but yeah)

want to wait a second? Claude did a PR pass. I still haven't looked at it or tried to run it though.

kixelated:
Anyway feel free to merge these in, but again I want to stress I haven't tested them yet. I'm primarily using this PR as a way of knowing how the API could be improved.

vipyne:
> Anyway feel free to merge these in, but again I want to stress I haven't tested them yet. I'm primarily using this PR as a way of knowing how the API could be improved.

copy that. 

I'm in the process of splitting out the moq transport into the appropriate FE libraries... I'll stick with the hand rolled protocol version for now (bc it works for me in local dev, which is all that matters, right 🙃).

once that is all in place, I'll have claude try subbing in these changes to see what happens.  in general, it makes sense to use these somewhat standardized libs instead of artisanal claude code.

kixelated:
Yeah and let me know if `moq/watch` could help on the front-end side too. There's currently support for streaming audio playback (via max_latency) but there's currently no support for buffered audio playback (min_latency?) like TTS.

vipyne:
cc @kixelated https://github.com/pipecat-ai/pipecat/pull/4629
most of the interesting stuff for you is probably in https://github.com/pipecat-ai/pipecat-client-web-transports/pull/139

Issue #4561, opened 2026-05-26
AudioBufferProcessor collapses wall-clock silence between bot utterances when no user audio is flowing


achalringg:
### pipecat version

1.2.1

### Python version

3.12.13

### Operating System

macOS 26.0 (logic-level bug, platform-independent — also reproduces on macOS 15 + Python 3.14)

### Issue description

`AudioBufferProcessor` advances the bot recording buffer only when an `OutputAudioRawFrame` arrives, and pads the bot buffer with silence only when an `InputAudioRawFrame` arrives (via `_sync_buffer_to_position`). If both are absent for a stretch of wall-clock time, the bot buffer never grows. The next bot utterance is then appended directly after the previous one, and the recording loses the real-time gap — two utterances spoken seconds apart sound concatenated.

This is observable any time user-side audio is not delivered to the processor during a bot silence: push-to-talk transports, server-side input gating during function/tool calls, agents that explicitly mute or suppress user audio between turns, etc.

**Relevant code** — [`src/pipecat/processors/audio/audio_buffer_processor.py`](https://github.com/pipecat-ai/pipecat/blob/main/src/pipecat/processors/audio/audio_buffer_processor.py):
- `_process_recording` (~line 188) — bot buffer only extends on `OutputAudioRawFrame`; bot buffer is only padded with silence inside the `InputAudioRawFrame` branch.
- `_sync_buffer_to_position` (~line 249) — the only silence-padding routine, and it is never called on a wall-clock timer.

There is no third source of advancement, so a quiet bot + absent user input freezes both buffers in time.

**Prior art:** PR #3541 intentionally removed wall-clock-based silence padding, so any fix here probably shouldn't reintroduce a timer inside `AudioBufferProcessor` itself. A separate downstream processor that emits silent `OutputAudioRawFrame` chunks during the hold window resolves it without touching `AudioBufferProcessor`. Happy to send a PR if a sibling processor (e.g. `BotRecordingSilencePadder`) is the preferred shape; an opt-in flag on `AudioBufferProcessor` is also fine.

### Reproduction steps

1. `pip install pipecat-ai==1.2.1`
2. Save the script below as `repro_audio_buffer_gap.py` and run it:

   ```python
   import asyncio
   import math
   import wave

   from pipecat.clocks.system_clock import SystemClock
   from pipecat.frames.frames import (
       BotStartedSpeakingFrame,
       BotStoppedSpeakingFrame,
       OutputAudioRawFrame,
[...]

PR #4562, opened 2026-05-26
Add BotRecordingSilencePadder to preserve wall-clock gaps in recordings

achalringg:
## Problem

`AudioBufferProcessor` advances the bot recording buffer only when an `OutputAudioRawFrame` arrives, and pads the bot buffer with silence only when an `InputAudioRawFrame` arrives (via `_sync_buffer_to_position`). If both are absent for a stretch of wall-clock time, the bot buffer never grows. The next bot utterance is then appended directly after the previous one and the saved recording loses the real-time gap — two utterances spoken seconds apart sound concatenated.

This is observable any time user-side audio is not delivered to the processor during a bot silence: push-to-talk transports, server-side input gating during function/tool calls, agents that explicitly mute or suppress user audio between turns, etc.

## Approach

Add a small sibling processor, `BotRecordingSilencePadder`, in `src/pipecat/processors/audio/`. Users insert it between `transport.output()` and `AudioBufferProcessor` when they need wall-clock fidelity in recordings. While the bot is not speaking, the padder pushes silent `OutputAudioRawFrame` chunks **downstream only**, so the recording processor keeps advancing the bot buffer in sync with wall-clock time. The upstream output transport never sees the padded frames — nothing is played out the line and no extra `BotStarted/Stopped` events are emitted.

### Why a sibling processor instead of changing `AudioBufferProcessor`

PR #3541 intentionally removed wall-clock-based silence padding from `AudioBufferProcessor` itself. This change respects that constraint — the new behavior lives in a separate, opt-in processor and `AudioBufferProcessor` is untouched.

## Changes

- **`src/pipecat/processors/audio/bot_recording_silence_padder.py`** (new): the processor. Configurable `chunk_ms` (default 100, min 20). Uses `self.create_task` / `self.cancel_task` for the silence loop and cleans up on `EndFrame` / `CancelFrame` / `cleanup()`.
- **`tests/test_bot_recording_silence_padder.py`** (new): two unit tests
  - two bot utterances separated by a wall-clock gap with no user audio in between are recorded with the gap preserved
  - the padder stops emitting silence while the bot is actually speaking

## Verification

```
$ pytest tests/test_bot_recording_silence_padder.py tests/test_audio_buffer_processor.py -v
...
tests/test_audio_buffer_processor.py ............ 6 passed
tests/test_bot_recording_silence_padder.py ...... 2 passed
8 passed
```

New tests pass; the existing `AudioBufferProcessor` suite is unaffected.

## Minimal repro of the underlying bug (without this fix)

Same script I'll attach to the bug-report issue — drives `AudioBufferProcessor` directly with two synthetic bot utterances separated by `await asyncio.sleep(3.0)` and no input audio. With pipecat 1.2.1 the bot buffer ends up 2.00s long when wall-clock was 5.00s. Inserting `BotRecordingSilencePadder` in front of `AudioBufferProcessor` makes the buffer match wall-clock.

Happy to iterate on naming, API surface, or move things around if a different shape is preferred.


achalringg:
this fixes: https://github.com/pipecat-ai/pipecat/issues/4561

markbackman:
Closing as this was resolved in https://github.com/pipecat-ai/pipecat/pull/4567.

PR #4563, opened 2026-05-26
fix: handle ContextWindowCompressionParams object in Gemini Live settings

### LINKED ISSUE #4560: GeminiLive LLM connection error 1007 on reconnect — "Request contains an invalid argument" with ContextWindowCompressionParams trigger_tokens
### pipecat version

1.2.1

### Python version

3.12.x

### Operating System

Ubuntu 24.04

### Issue description

When using GeminiLiveLLMService, a WebSocket connection error with code 1007 is thrown during connection attempts.

WARNING::pipecat::pipecat.services.google.gemini_live.llm::1258::Connection error (failure 1/3): 1007 None. Request contains an invalid argument.

Currently We are using Gemini Live S2S Service and Adding ContextWindowCompressionParams trigger_tokens to API.

When Trigger token are passed then pipecat is throwing the above error.

I am passing like below:
context_window_compression={'enabled': True, 'trigger_tokens': 12000} 

based on below PR the value is INT and i am passing same for trigger_tokens.

https://github.com/pipecat-ai/pipecat/pull/1701/changes

if i remove trigger_tokens then everything works fine...I want to add this value to compress tokens..

Can you please add fix or propose proper usage for this.

### Reproduction steps

1.Use Gemini Live S2S Service
2.Add https://docs.pipecat.ai/api-reference/server/services/s2s/gemini-live#contextwindowcompressionparams
both parameters and try a call
3.Pipecat shows error eventhough right values are passed trigger_tokens

### Expected behavior

Pipecat framework should accept trigger_tokens values..If removed then everything working properly with just ContextWindowCompressionParams--> enabled:True

### Actual behavior

Somehow trigger_tokens is not working as expected eventhough INT value is passed correctly..

### Logs

```shell
context_window_compression={'enabled': True, 'trigger_tokens': 8000} and 26-05-26 06:48:34.126::------::WARNING::pipecat::pipecat.services.google.gemini_live.llm::1258::Connection error (failure 1/3): 1007 None. Request contains an invalid argument.
```

markbackman:
## Summary

Fixes a crash when supplying `context_window_compression` to `GeminiLiveLLMService` (and `GeminiVertexLiveLLMService`) through the canonical `settings` API.

The setting can arrive in two shapes:
- a `ContextWindowCompressionParams` Pydantic object — via `settings=Settings(...)`, where `apply_update` stores it as-is
- a `dict` — via the legacy `params=InputParams(...)` path, which stores a `model_dump()`

The session-config consumer only handled the dict shape (calling `.get()` on the value), so the canonical path raised:

```
'ContextWindowCompressionParams' object has no attribute 'get'
```

This normalizes the value to the params model before reading it, so both input shapes work. `GeminiVertexLiveLLMService` inherits the same consumer and is covered too.

This is not a Google API change — the API still supports context window compression. The bug was purely in how Pipecat read the setting after the settings-API migration.

## How to reproduce

```python
llm = GeminiLiveLLMService(
    api_key=os.environ["GOOGLE_API_KEY"],
    settings=GeminiLiveLLMService.Settings(
        context_window_compression=ContextWindowCompressionParams(
            enabled=True,
            trigger_tokens=1000,
        ),
    ),
)
```

## Test plan

- [x] `settings=Settings(context_window_compression=ContextWindowCompressionParams(...))` resolves to enabled with the given `trigger_tokens` (previously crashed)
- [x] default (unset) resolves to disabled
- [x] legacy `params=InputParams(context_window_compression=...)` (dict shape) still resolves correctly
- [x] `ruff check` passes

Fixes https://github.com/pipecat-ai/pipecat/issues/4560


filipi87 (approved):
LGTM!

PR #4565, opened 2026-05-26
chore: support deepgram-sdk 6.x and 7.x and drop websockets upper bound

lencshu:
#### Summary

Widens the `deepgram` extra to `deepgram-sdk>=6.1.1,<8` (was `>=6.1.1,<7`) so installations can resolve to either deepgram-sdk 6.x or 7.x, and drops the upper bound on the `websockets-base` extra (`websockets>=13.1`, was `<16.0`) so downstream deployments can resolve `websockets` to 16.x and beyond.

The shipped lockfile still resolves to deepgram-sdk 6.1.1 (matching `main`); the change just lets downstream apps pull in 7.x if they prefer. A small compatibility shim in `DeepgramSTTService` keeps `base_url` overrides working under both 6.x and 7.x.

#### Files changed

- `pyproject.toml`: widen `deepgram-sdk` upper bound from `<7` to `<8`, drop `websockets<16.0` upper bound (now `websockets>=13.1`).
- `src/pipecat/services/deepgram/stt.py`: pass `agent_rest` to `DeepgramClientEnvironment` only when the installed SDK accepts it (deepgram-sdk 7.2.0 added that kwarg as required; 6.x rejects it).
- `uv.lock`: two specifier lines updated; deepgram-sdk stays locked to 6.1.1.
- `changelog/+upgrade-deepgram-sdk-7-websockets-16.changed.md`: towncrier fragment.

Total diff: 17 insertions, 9 deletions across 4 files.

##### deepgram-sdk: now spans 6.1.1 through 7.x

Pipecat's Deepgram integrations (`services/deepgram/stt.py`, `services/deepgram/tts.py`, `services/deepgram/flux/stt.py`) use:

- `AsyncDeepgramClient`, `DeepgramClientEnvironment`
- `EventType.MESSAGE`, `EventType.ERROR`
- `ListenV1Results`, `ListenV1CloseStream`, `ListenV1Finalize`, `ListenV1KeepAlive`
- `client.listen.v1.connect()` and the resulting connection's `send_media` / `send_finalize` / `send_close_stream` / `send_keep_alive` / `start_listening()` / `on(EventType, ...)` methods.

All of these symbols exist in deepgram-sdk 6.1.1 (the fully generated SDK introduced in 6.0.0) and remain available through 7.2.0. The 7.0.0 / 7.1.x / 7.2.0 releases are generator regenerations whose breaking changes are confined to schema renames Pipecat does not touch.

The only relevant difference is that deepgram-sdk 7.2.0 added a required `agent_rest` kwarg to `DeepgramClientEnvironment.__init__`. Without a shim, `DeepgramSTTService(api_key=..., base_url=...)` would raise `TypeError` on 7.2.0 and silently fall through Pipecat's `except Exception` to the default URL. This PR runtime-introspects the constructor and passes `agent_rest=http_url` only when present, so `base_url` overrides keep working on both 6.x and 7.x without code branches per SDK version.

##### websockets: upper bound dropped

Pipecat's actual `websockets` API usage:

- 28 call sites of `websockets.asyncio.client.connect` (modern API, recommended since 13.x, fully retained in 16.0).
- 1 call site of `websockets.asyncio.server.serve` (same).
- 24 import sites of `websockets.protocol.State`, plus `ConnectionClosed{,Error,OK}` from `websockets.exceptions`. All retained.
- Three modules (`services/websocket_service.py`, `transports/websocket/client.py`, `transports/websocket/server.py`) reference the legacy aliases `websockets.WebSocketClientProtocol` and `websockets.WebSocketServerProtocol` in type hints only. These aliases are still importable in 16.0 (resolved via `websockets.legacy.*`) and emit a `DeprecationWarning`. The five-year deprecation policy means they will not be removed until around 2030. The hints already carry `# pyright: ignore[reportAttributeAccessIssue]` and have no runtime effect.

`websockets` 16.0 itself has only one breaking change relative to 15.x: it requires Python >= 3.10. Pipecat already requires Python >= 3.11, so this is a no-op.

##### nvidia-riva-client constraint (why the lock still shows websockets 15.0.1)

`nvidia-riva-client==2.25.1` declares `websockets>=15.0.1,<16`. Because the `nvidia` extra participates in pipecat's lock resolution, `uv lock` keeps `websockets==15.0.1` in the shipped lockfile. This is unchanged from main. Downstream consumers that do not install the `nvidia` extra (or that install a future `nvidia-riva-client` which relaxes the bound) will be free to resolve `websockets` to 16.x or later.

This is purely an upstream NVIDIA limitation; pipecat cannot work around it here.

#### Verification

Under the locked environment (`deepgram-sdk 6.1.1`, `websockets 15.0.1`):

- `uv lock` and `uv sync --group dev --all-extras --no-extra gstreamer --no-extra local` succeed.
- `uv run pytest tests/test_deepgram_stt.py tests/test_service_init.py tests/test_websocket_service.py tests/test_fastapi_websocket.py -q`: 239 passed, 1 failed. The single failure is `NvidiaSegmentedSTTService` and is reproducible on `main` without this change (pre-existing, unrelated).
- `uv run ruff check` and `uv run ruff format --check`: clean.
- Smoke-checked `DeepgramSTTService(api_key=..., base_url=...)`: constructs without falling back to the default URL.

Cross-version check by overriding the venv to `deepgram-sdk==7.2.0` (then restored):

- `uv run pytest tests/test_deepgram_stt.py tests/test_service_init.py -q`: 220 passed, 1 failed (same pre-existing `NvidiaSegmentedSTTService`).
- `DeepgramSTTService(api_key=..., base_url=...)` smoke check: constructs without the fallback warning, confirming the `agent_rest` shim works.

Cross-version check by overriding the venv to `websockets==16.0` (then restored):

- All ws-using pipecat modules import successfully. The only diagnostics are `DeprecationWarning`s for the legacy type-hint aliases noted above.
- 66 / 66 relevant tests pass against `websockets==16.0`:
  - `tests/test_websocket_service.py`: 11/11
  - `tests/test_openai_responses_websocket.py`: 31/31
  - `tests/test_websocket_transport.py`: 3/3 (others skipped for missing `webrtc` extra)
  - `tests/test_fastapi_websocket.py`: 8/8
  - `tests/test_deepgram_stt.py`: 13/13

#### Notes for reviewers

- Lockfile diff is tiny on purpose: only the two `requires-dist` specifier lines change. The previously-large lock churn (pydantic / pydantic-core / fastapi-cloud-cli collapse) only happened when the deepgram-sdk lower bound forced a jump to 7.x; with the bound widened, `uv lock` keeps the existing 6.1.1 resolution and the transitive graph is untouched.
- `deepgram-sdk<8` is kept because every prior deepgram-sdk major bump shipped a rewritten public client. Happy to relax this in a follow-up release once 8.x is out and validated, but left in here as a precaution that reflects observed upstream behavior rather than speculation.


Natim:
IMHO, a Python lib shouldn't be protective of the upper dependency limit; it should state the minimum supported version and then evaluate whether the new version is compatible. It is the responsibility of the project using pipecat-ai to assess and restrict, if necessary.

pyproject.toml
```diff
@@ -62,7 +62,7 @@ cartesia = [ "pipecat-ai[websockets-base]" ]
 camb = [ "camb-sdk>=1.5.4,<2" ]
 cerebras = []
 daily = [ "daily-python~=0.28.0" ]
-deepgram = [ "deepgram-sdk>=6.1.1,<7", "pipecat-ai[websockets-base]" ]
+deepgram = [ "deepgram-sdk>=7.2.0,<8", "pipecat-ai[websockets-base]" ]
```
Natim:
```suggestion
deepgram = [ "deepgram-sdk>=6.1.1", "pipecat-ai[websockets-base]" ]
```

Natim: Why enforce 7.2.0 if 6.1.1 still works?

lencshu:
> IMHO, a Python lib shouldn't be protective of the upper dependency limit; it should state the minimum supported version and then evaluate whether the new version is compatible. It is the responsibility of the project to asses and restrict if necessary.

Thanks a lot for pointing out! @Natim  This has definitely been a long standing debate in the community:

E.g.,
> [By pinning against an unreleased version, it causes toil (maintenance burden) for each and every project that does the pinning (and the consumers that use it) and increases the risk of erosion if maintenance is unsustained](https://setuptools.pypa.io/en/latest/userguide/dependency_management.html)

I agree with the underlying principle and I have applied that to this PR with one small carve-out:

- `websockets`: dropped the upper bound entirely. The upstream `websockets` 5-year deprecation policy makes future API removal in 17.0 unlikely as well

- `deepgram-sdk`: kept the `<8` cap. Unlike `websockets`, this is not speculative: deepgram-sdk v3 -> v6 -> v7 each shipped a rewritten public client, so an unconstrained spec would very likely break downstream users the day deepgram-sdk 8.0 ships. I would rather flip the cap off in a follow-up release that has actually validated 8.x than have downstream apps silently break.

If we want to extend this policy further, I am happy to open a separate issue for that broader policy discussion so we can tackle it consistently rather than ad hoc in this PR

Pushed an amended commit reflecting the change and the PR title and description have been updated too.

Natim (approved):
Good Job !

markbackman (approved):
LGTM! Thanks for the clean up.

PR #4566, opened 2026-05-26
fix: avoid tracing-context AttributeError in service decorators

AlessioSpallino:
**Summary**
This PR fixes a tracing-related crash in service decorators during function-call handling.

**Problem**
In some function-call paths, tracing context access could raise:
`'NoneType' object has no attribute 'get_turn_context'`
This can happen when `_tracing_context` is truthy at the call-site check, but the underlying getter path raises an `AttributeError.`

**Root Cause**
_get_turn_context() and _get_parent_service_context() used direct getter calls after a truthiness check.
That check does not protect against AttributeError raised inside get_turn_context() / get_conversation_context().

**How to reproduce**
- Disable tracing (enable_tracing=False).
- Trigger a tool/function call (runs through LLMService._run_function_call). Tracing decorator helpers can raise AttributeError: `'NoneType' object has no attribute 'get_turn_context'` when tracing context isn’t available.

**Changes**
Updated `_get_turn_context` to return None on AttributeError.
Updated `_get_parent_service_context` to return None conversation context on AttributeError, then keep the existing fallback behaviour.

**Tests**
Added targeted regression tests in `tests/test_service_tracing_decorators.py`:
- test_get_turn_context_handles_none_context
- test_get_turn_context_handles_truthy_broken_proxy
- test_get_parent_service_context_handles_none_context
- test_get_parent_service_context_handles_truthy_broken_proxy

Validation run: `uv run pytest tests/test_service_tracing_decorators.py`


markbackman:
Thanks for digging into this. I tried to reproduce it and don't think this path can crash on current `main`.

`_tracing_context` only ever holds `None` (`AIService.__init__`, `StartFrame` default) or a `TracingContext` from `PipelineWorker`. Its getters are one-line returns of fields set in `__init__`, so they can't raise `AttributeError`, and `None` is already covered by the existing guard.

On the repro: with `enable_tracing=False` these helpers are never reached — every call site sits behind `if not getattr(self, "_tracing_enabled", False): return await f(...)` (`service_decorators.py:928, 1148, 1433`, and the TTS/STT closures at `:366, :632`). And `llm_service.py` has no tracing code, so there's no path from `_run_function_call` into them.

The two "broken proxy" tests construct a getter that is literally `inner = None; return inner.get_turn_context()` — they raise the error themselves rather than reaching a real state. Against unmodified `main`, the two tests using real state (`_tracing_context=None`) pass; only the fabricated proxies fail.

I'd also push back on the shape: a blanket `except AttributeError` would swallow genuine bugs in a future `TracingContext` and silently disable tracing.

That said, it sounds like you hit a real traceback. Could you share it plus your pipecat version? If it predates the `TracingContext` refactor, the old singleton `TurnContextProvider`/`ConversationContextProvider` are a plausible source and the fix would look different.

PR #4567, opened 2026-05-26
Fixed a mute gap bug in AudioBufferProcessor where silence was missing

### LINKED ISSUE #4561: AudioBufferProcessor collapses wall-clock silence between bot utterances when no user audio is flowing

filipi87:
## Summary
  
- Fixed `AudioBufferProcessor` concatenating utterances spoken seconds apart when the microphone is muted between them. When no user audio arrives for more than 200 ms, silence proportional to the wall-clock gap is now inserted into the user buffer so the two utterances remain temporally separated in the recording.
- Bot audio received during a mute period correctly advances the internal timestamp so silence is not double-counted (i.e., the gap is measured from the last bot sync, not from the last user audio frame).
- Added 8 automated tests covering: gap threshold boundary, proportional silence insertion, the concatenation regression, double-counting prevention, and buffer state after flush and recording reset.

Fixes https://github.com/pipecat-ai/pipecat/issues/4561


achalringg:
@filipi87 I think we will need same side bot buffer as well.
can you review this PR for the same: https://github.com/pipecat-ai/pipecat/pull/4571

filipi87:
Thank @achalringg, I have merged your changes in this PR. 🎉

aconchillo (approved):

aconchillo:
LGTM

PR #4571, opened 2026-05-27
Symmetric bot-side silence gap fill in AudioBufferProcessor

achalringg:
## Summary

Extends the user-side fix in #4567 with a mirror on the bot side.

Without it, two bot utterances separated by a silent wall-clock gap — e.g. progressive hold messages spoken while a slow function call (20–30 s) runs — are appended to the bot buffer with no padding between them, so the recording sounds *concatenated* even though the caller heard them seconds apart. We hit this in our deployment: the bot says \"one moment\", waits 7 s, says \"still working on it\", waits 7 s, says \"almost there\", returns the answer — and the recording stitches all the hold lines together at the start.

## Changes

- Added `_fill_bot_silence_gap` (exact mirror of `_fill_user_silence_gap`).
- Tracks `_last_bot_buffer_update_time` alongside the existing `_last_user_buffer_update_time`.
- User audio that syncs the bot buffer now advances the bot timestamp, so silence injected by the sync isn't double-counted by the next bot frame (symmetric to what this PR already does for the user side on bot-audio sync).
- `_reset_recording` and `_reset_primary_audio_buffers` reset/refresh both timestamps.

## Tests

Added `TestBotSilenceGapInsertion` — 8 tests mirroring `TestMuteGapSilenceInsertion`:

- no silence inserted when there's no prior timestamp
- no silence below the 200 ms threshold
- proportional silence for a 1 s idle gap
- the concatenation-regression test (two utterances 1 s apart)
- user-frame-during-pause advances the bot timestamp (no double-counting)
- user buffer is synced to bot position after the gap fill
- `stop_recording` clears the bot timestamp
- buffer flush sets the bot timestamp to flush time

All 22 audio_buffer_processor tests pass locally (14 existing + 8 new).

## Note

Targets `filipi/audio_buffer` so it can be folded into #4567 rather than landing as a separate PR. Happy to rebase / split if you'd prefer it as its own PR against `main` once #4567 merges.


filipi87 (approved):
Thank you for catching this. This is an even more specific edge case, because the user also needs to be muted for this to happen, but it’s nice to include it anyway. Nice catch!

PR #4573, opened 2026-05-27
Inworld Realtime: default STT to inworld-stt-1, quiet unhandled server events

cshape:
### Summary

Two small, related quality-of-life improvements to `InworldRealtimeLLMService`:

1. **Default STT model → `inworld/inworld-stt-1`.** The service previously
   defaulted input transcription to `assemblyai/u3-rt-pro`. It now defaults to
   the Inworld-native `inworld/inworld-stt-1`. Passing `stt_model=` still
   overrides. The TTS default (`inworld-tts-2`) and the voice/LLM-model defaults
   are unchanged.

2. **Quiet logging for unrecognized realtime server events.** The current
   Inworld Realtime AsyncAPI emits newer events (e.g. `response.output_text.done`)
   that the service doesn't have a model for, producing a noisy
   `WARNING | Failed to parse server event: Unimplemented server event type: ...`
   on every turn. These are benign. `parse_server_event()` now returns `None`
   for unrecognized event types (instead of raising), and the service logs them
   at `DEBUG` and skips them. Known event types that genuinely fail validation
   still raise and warn as before. This also subsumes the previous hardcoded
   skip for `conversation.item.done`.

### Notes

- No new Pydantic event models are added — unrecognized events are handled by
  the generic `DEBUG` path rather than by registering each new type.
- Changelog fragments included.
- Also updated the LLM to a better one in the Inworld realtime example

### Verification

- `uv run ruff check` and `uv run ruff format --check` pass.
- `uv run towncrier build --draft --version Unreleased` renders both fragments.
- A constructed `InworldRealtimeLLMService` now reports
  `inworld/inworld-stt-1` for input transcription and `inworld-tts-2` for
  output in its session properties.
- Exercised against a local Inworld Realtime server via
  `examples/realtime/realtime-inworld.py`.


filipi87 (approved):
LGTM!

PR #4574, opened 2026-05-27
Subagents improvements: rename proxy classes, variadic add/watch_workers, default-watch children

aconchillo:
## Summary

Follow-up cleanups on top of the subagents port (#4493):

- **`cancel_runner_on_idle_timeout` on `PipelineWorker`** (default `True`). When the main `PipelineWorker` of a multi-worker bot hit its idle timeout, only its own pipeline was cancelled — sibling root workers (LLM helpers, sidecars) were never told to stop and the runner hung. With the new flag, the worker also emits `BusCancelMessage` so the runner broadcasts cancellation to every other root worker. The local cancel still happens too, so standalone usage (no runner) keeps working. Set the flag to `False` for sidecar pipelines that should self-cancel on idle without bringing down their peers.
- **Thread a cancel reason through `BaseWorker.cancel`**. `BusCancelMessage` already had an optional `reason` and the runner already propagated it onto every `BusCancelWorkerMessage`, but `BaseWorker.cancel()` discarded it — asymmetric with `PipelineWorker.cancel(reason=...)`. Both paths now accept it; the idle-timeout-triggered cancel is tagged `"idle timeout"` end-to-end so operators see in logs why the bot went down.
- **Fix a subscribe-vs-ready race in the runner.** When a worker was added via `PIPECAT_SETUP_FILES` it landed last in the runner's start loop. Previously the earlier workers subscribed to the bus *inside* `run()` and then immediately emitted `BusWorkerReadyMessage`, so the later-added subscriber missed those messages. `BaseWorker.attach` is now async and subscribes the worker at attach time — every subscriber is listening before any worker emits. `WorkerBus.subscribe` is also now idempotent (and no longer leaks dispatch tasks on a duplicate subscribe).
- **Rename** `WebSocketProxyClientTask` / `WebSocketProxyServerTask` to `WebSocketProxyClient` / `WebSocketProxyServer`. The `Task` suffix was a holdover from before the `BaseWorker` rename; `Client`/`Server` already describe the role and the `pipecat.workers.*` module path conveys worker-ness. Updated all examples and tests.
- **Variadic worker APIs**: `BaseWorker.add_worker(worker)` → `add_workers(*workers)` and `watch_worker(name)` → `watch_workers(*names)`. Matches the earlier `PipelineRunner.add_worker` → `add_workers` rename and collapses several existing for-loops in the distributed-handoff examples.
- **`add_workers(..., watch=True)` default**: adding a child usually implies wanting `on_worker_ready` for it. The default now watches each newly added child (only the ones actually taken ownership of — workers skipped for having an existing parent are not watched). Pass `watch=False` to opt out.
- **Split the multi-worker README** into per-example READMEs (matching the ui-worker layout). The top-level README is now a short orientation page: setup, shared env vars, a conceptual intro to the multi-worker patterns, and three index tables linking to each example subfolder.
- **Polish**: name main `PipelineWorker`s explicitly (`code-assistant`, `parallel-debate`, `voice-agent`), kebab-case worker names (`code-worker`, `sensor-controller`), and call `runner.cancel()` (not `worker.cancel()`) from disconnect handlers in the distributed-handoff examples so the LLM helpers shut down too.
- **Fix** a regression from an earlier rename sweep: `BackgroundTasks.add_task(...)` in `transports-small-webrtc.py` had been turned into `add_worker(...)`, which doesn't exist on FastAPI's `BackgroundTasks`. Restored.


markbackman (approved):
LGTM! Nice clean up. Having Claude check for anything I missed, too.

markbackman (commented):
## Code review

Two findings below. Renamed/removed worker symbols were checked against the deprecation policy and are fine (they were introduced after v1.2.1 and never shipped).

src/pipecat/pipeline/worker.py
```diff
@@ -1214,11 +1222,19 @@ async def _idle_timeout_detected(self) -> bool:
...
-        if self._cancel_on_idle_timeout:
-            logger.warning(f"Idle pipeline detected, cancelling pipeline worker...")
-            await self.cancel()
-            return False
-        return True
+        if not (self._cancel_on_idle_timeout or self._cancel_runner_on_idle_timeout):
+            return True
+
+        logger.warning("Idle pipeline detected, cancelling pipeline worker...")
+        await self.cancel(reason="idle timeout")
```
markbackman, line 1234: **`cancel_on_idle_timeout=False` no longer keeps an idle worker alive.**

Line 1229 cancels the local pipeline whenever *either* flag is set (the guard on line 1225 is an `or`). Because `cancel_runner_on_idle_timeout` defaults to `True`, a worker built with `cancel_on_idle_timeout=False` and the new flag left at its default now gets cancelled on idle — the opposite of what `cancel_on_idle_timeout` documents (“Whether the pipeline worker should be cancelled if the idle timeout is reached”). On `main`, `cancel_on_idle_timeout=False` returned early and left the pipeline running.

This is a backward-incompatible change beyond the one in the PR description (which only mentions sidecars passing `cancel_runner_on_idle_timeout=False`). The tests corroborate it: every spot that used `cancel_on_idle_timeout=False` had to add `cancel_runner_on_idle_timeout=False` to stay green, and no test covers the `(False, default-True)` combination.

Consider gating the local cancel on `cancel_on_idle_timeout` (broadcasting only under `cancel_runner_on_idle_timeout`) — or, since a worker can't realistically stay alive while its runner is torn down, short-circuiting the whole method when `cancel_on_idle_timeout=False` and updating the `cancel_on_idle_timeout` docstring to note the interaction.

https://github.com/pipecat-ai/pipecat/blob/8045fa401488f000598e431ebf05501fd0437328/src/pipecat/pipeline/worker.py#L1225-L1237

src/pipecat/pipeline/base_worker.py
```diff
@@ -543,22 +548,35 @@ async def send_bus_error_message(self, error: str) -> None:
...
+            if worker._parent is not None:
+                logger.warning(
+                    f"Worker '{worker.name}' already has parent '{worker._parent}', skipping"
+                )
+                continue
+            worker._parent = self.name
+            self._children.append(worker)
+            added_names.append(worker.name)
+            await self.send_bus_message(BusAddWorkerMessage(source=self.name, worker=worker))
+        if watch and added_names:
```
markbackman, line 578: **Default `watch=True` double-registers the readiness callback when combined with `@worker_ready`.**

With the new `watch=True` default, `add_workers(child)` auto-calls `watch_workers(child.name)` (line 579), which registers `self._on_watched_worker_ready` in the registry. If the same parent also declares a `@worker_ready(name=child.name)` handler, `_watch_decorated_workers()` registers `self._on_watched_worker_ready` for that name too. `WorkerRegistry.watch` appends handlers without de-duping and `_notify` invokes every entry, so when the child registers, the parent's `@worker_ready` handler, its `on_worker_ready` override, and the `on_worker_ready` event each fire **twice** (verified with a repro).

`add_workers(child)` plus a readiness handler for that child is exactly the documented idiom for a parent reacting to a child it created, so this is reachable in normal use (just not exercised by any in-tree example yet). On `main` it couldn't happen because `add_worker` didn't auto-watch.

Consider de-duping handlers in `WorkerRegistry.watch`, or having `add_workers` skip auto-watching names already covered by `@worker_ready` handlers.

https://github.com/pipecat-ai/pipecat/blob/8045fa401488f000598e431ebf05501fd0437328/src/pipecat/pipeline/base_worker.py#L577-L580

aconchillo: Fixed!

src/pipecat/pipeline/worker.py
aconchillo: Fixed!

PR #4577, opened 2026-05-28
Revert PR 4511 changes to the deprecated RimeNonJsonTTSService

markbackman:
Reverts the `coda` model changes to `RimeNonJsonTTSService` and `RimeNonJsonTTSSettings`, which are both deprecated classes.


filipi87 (approved):
Makes sense.

PR #4578, opened 2026-05-28
fix(base_output): bound _send_silence write timeout

### LINKED ISSUE #4547: Pipeline hangs on LiveKit disconnect: MediaSender._send_silence blocks on capture_frame when audio source is no longer drained
### pipecat version

0.0.103

### Python version

3.12.13

### Operating System

Debian 13 (slim) in container

## Summary

When a LiveKit session ends after the remote peer has disconnected, `MediaSender.stop()` blocks indefinitely on `await self._audio_task`. The pipeline never finishes shutting down and the process can only exit via SIGKILL after the grace period.

The hang reproduces reliably with `audio_out_enabled=True` and does NOT reproduce with `audio_out_enabled=False`, which isolates the bug cleanly to the audio output path on disconnect.

## Root cause

`MediaSender.stop()` (`src/pipecat/transports/base_output.py`) waits for `_audio_task` to drain:

```python
    await self._clock_queue.put((float("inf"), frame.id, frame))
    await self._audio_queue.put(frame)
    if self._audio_task:
        await self._audio_task   # blocks here
```

`_audio_task_handler` sees the `EndFrame` and calls `_send_silence` (default 2 seconds) before breaking out of the loop:

```python
if isinstance(frame, EndFrame):
    await self._send_silence(self._params.audio_out_end_silence_secs)
    break
```

`_send_silence` calls `write_audio_frame` → `_client.publish_audio` → `self._audio_source.capture_frame` (from the `livekit-rtc` Python SDK). `capture_frame` has backpressure — it awaits when the internal buffer cannot accept more data.

When the remote peer has already disconnected:

1. The LiveKit audio source is no longer being drained by any subscriber.
2. The 2 seconds of silence (~96 KB at 24 kHz mono 16-bit) fills the internal buffer.
3. `capture_frame` awaits forever waiting for buffer space that will never free up.
4. `_audio_task` never reaches `break`.
5. `MediaSender.stop()` never returns.

There is no timeout or cancellation path on this branch.

## Why some deployments don't hit this

We run two configurations against the same LiveKit transport, which made the bug easy to isolate:

- **Video mode** (`audio_out_enabled=False`): an external avatar service composes audio+video and pushes to LiveKit directly. Pipecat's audio output is never used. The buffer stays empty for the whole session, so the 2s silence on disconnect enqueues instantly and cleanup proceeds normally.
- **Voice mode** (`audio_out_enabled=True`): pipecat pushes TTS audio through its own LiveKit output for the entire session. The buffer is under constant backpressure (drained by the subscriber in real-time). The moment the peer disconnects, draining stops — the trailing silence write blocks.

## Workaround

Setting `audio_out_end_silence_secs=0` in `LiveKitParams` resolves the hang completely — `_send_silence` has an `if secs <= 0: return` early-return, so the blocking call is skipped entirely. The theoretical cost (clipping the bot's last word) is not an issue in practice when disconnect is user-initiated.

```python
params=LiveKitParams(
    audio_in_enabled=True,
    audio_out_enabled=True,
    audio_out_end_silence_secs=0,  # workaround for the hang
)
```

## Suggested fix

The underlying issue is that the pipeline cleanup path has no timeout for downstream sink writes that may never complete. A few options:

1. **Wrap `_send_silence` in `asyncio.wait_for`** with a timeout (e.g. `max(audio_out_end_silence_secs * 2, 1)` seconds). Most general — would protect against similar issues in other transports.
2. **Check connection state in the LiveKit transport's `write_audio_frame`** — skip the write if the room is closing or disconnected. Most targeted.
3. **Make `_audio_task` cancellable from `MediaSender.stop()`** after a timeout, so cleanup can always make progress regardless of which sink is misbehaving.

Option 2 is the smallest patch; option 1 is the safest general guard.

## Related issues

- #3757 — `_wait_for_pipeline_end` has no timeout for EndFrame; same class of problem (no timeout on cleanup wait) in the pipeline task layer.
- #4466 — also `MediaSender` lifecycle behavior on shutdown / interruption.
- #3872 — different `MediaSender` issue (audio buffer not cleared on reconnect in PTT mode), but illustrates the same area of the codebase.

### Reproduction steps

1. Create a `LiveKitTransport` with `audio_out_enabled=True` (default `audio_out_end_silence_secs=2`).
2. Run a pipeline that produces TTS audio continuously (any STT/LLM/TTS chain works).
3. While the bot is actively streaming audio, have the remote peer disconnect from the LiveKit room (e.g. user closes the UI, network drop, or `room.disconnect()` from the client side).
4. Trigger graceful pipeline shutdown — e.g. queue an `EndFrame`, call `task.stop_when_done()`, or rely on an `on_participant_left` handler to end the session.
5. Observe that `MediaSender.stop()` never returns. The pipeline task hangs on `await self._audio_task`. The process only exits when killed externally (e.g. K8s SIGKILL after the termination grace period).

### Expected behavior

`MediaSender.stop()` should return within a bounded time regardless of whether the LiveKit peer is still connected and consuming audio. Pipeline cleanup should complete and the process should exit cleanly through the normal shutdown path.

### Actual behavior

`MediaSender.stop()` blocks indefinitely on `await self._audio_task`, because `_audio_task_handler` is itself blocked inside `_send_silence` → `capture_frame`, waiting for buffer space in a LiveKit audio source that is no longer being drained. Cleanup never completes. In a Kubernetes deployment this means the pod waits the full `terminationGracePeriodSeconds` and then gets SIGKILL'd, which prevents any `finally`-block cleanup (including trace span finalization) from running.

### Logs

```shell

```

omChauhanDev:

Fixes: #4547 

**Issue :**   When the remote peer disconnects mid-stream on LiveKit & the pipeline shuts down gracefully, `MediaSender._send_silence` calls `write_audio_frame` to flush the trailing silence. LiveKit's `AudioSource.capture_frame` has buffer backpressure, with no consumer draining, it awaits forever. `MediaSender.stop()` blocks on the audio task, pipeline shutdown hangs.

**Change :**  Wrap the silence write in `asyncio.wait_for(timeout=secs + 1)`. On timeout we log a warning & move on, shutdown proceeds normally.

### DIFF
--- a/changelog/4578.fixed.md
+++ b/changelog/4578.fixed.md
@@ -1,0 +1,1 @@
+- Fixed pipeline shutdown hanging on LiveKit when the remote peer disconnected mid-stream. The trailing `audio_out_end_silence_secs` write is now bounded by a timeout.
--- a/src/pipecat/transports/base_output.py
+++ b/src/pipecat/transports/base_output.py
@@ -819,7 +819,13 @@ class BaseOutputTransport(FrameProcessor):
             silence_frame = OutputAudioRawFrame(
                 audio=silence, sample_rate=self.sample_rate, num_channels=1
             )
-            await self._transport.write_audio_frame(silence_frame)
+            try:
+                await asyncio.wait_for(
+                    self._transport.write_audio_frame(silence_frame),
+                    timeout=secs + 1,
+                )
+            except TimeoutError:
+                logger.warning(f"{self} timed out writing end-frame silence")
 
         async def _audio_task_handler(self):
             """Main audio processing task handler."""


filipi87 (approved):
This seems like a good safeguard to add. LGTM.
Thank you for fixing this.

PR #4581, opened 2026-05-28
Cartesia STT followup: tighten Turns service

markbackman:
## Summary

Follow-up to #4552.

- Stop `CartesiaTurnsSTTService` from emitting empty `TranscriptionFrame` / `InterimTranscriptionFrame` when a `turn.end` / `turn.update` carries an empty transcript (e.g. when the dangling-turn watchdog forces a turn end on silence). Lifecycle frames, metrics, and event handlers still fire so the turn closes cleanly.
- Smaller cleanups: raise `ImportError` instead of bare `Exception` for the missing optional extra; reword a stale receive-task reconnect comment.
- Restore the canonical `voice-cartesia.py` / `transcription-cartesia.py` example names, and have the transcription examples print interim frames. Evals updated to match.


aconchillo (approved):

aconchillo:
LGTM

PR #4585, opened 2026-05-28
Let turn-based STT services opt out of TTFS

markbackman:
## Summary

- Turn-based STT services (`CartesiaTurnsSTTService`, `DeepgramFluxSTTService`) have no meaningful "speech end → final transcript" interval — the server defines turn boundaries directly. Omitting `ttfs_p99_latency` on these previously produced a spurious `ttfs_p99_latency not set, using default 1.0s` warning at pipeline start.
- Adds the `STTService.supports_ttfs` property (default `True`). When a subclass returns `False`, `_push_stt_metadata` broadcasts `STTMetadataFrame` with `ttfs_p99_latency=0.0` and skips the warning. Downstream turn-stop strategies already treat `0.0` as "no extra wait."
- Overrides `supports_ttfs` to return `False` on the two turn-based services.

## Testing

- Run `examples/voice/voice-cartesia-turns.py` and `examples/voice/voice-deepgram-flux.py`; confirm the `ttfs_p99_latency not set` warning no longer appears at pipeline start.

### DIFF
--- a/src/pipecat/services/stt_service.py
+++ b/src/pipecat/services/stt_service.py
@@ -108,6 +108,8 @@ class STTService(AIService):
                 This is broadcast via STTMetadataFrame at pipeline start for downstream
                 processors (e.g., turn strategies) to optimize timing. Subclasses provide
                 measured defaults; pass a value here to override for your deployment.
+                Turn-based services where the server defines turn boundaries should
+                override :attr:`supports_ttfs` to return False instead of supplying a value.
             keepalive_timeout: Seconds of no audio before sending silence to keep the
                 connection alive. None disables keepalive. Useful for services that
                 close idle connections (e.g. behind a ServiceSwitcher).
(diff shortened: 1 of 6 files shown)


src/pipecat/services/stt_service.py
```diff
@@ -468,8 +470,23 @@ async def push_frame(self, frame: Frame, direction: FrameDirection = FrameDirect
 
         await super().push_frame(frame, direction)
 
+    def supports_ttfs(self) -> bool:
```
aconchillo, line 474: would be nicer as a property i think.

aconchillo (approved):

aconchillo:
LGTM. Small comment.

src/pipecat/services/stt_service.py
markbackman: Done!

PR #4586, opened 2026-05-28
Migrate the eval script to the worker API

aconchillo:
## Summary

`scripts/evals/eval.py` still used the pre-subagents API. Bring it in line with the worker API merged in #4493:

- `PipelineTask` / `PipelineParams` import switched to `PipelineWorker` from `pipecat.pipeline.worker`.
- `task` renamed to `worker`; `run(task)` replaced with `add_workers(worker)` + `run()`.
- Client disconnect now calls `runner.cancel()` (instead of `task.cancel()`) so the runner shuts down cleanly.

Internal dev tooling only (`scripts/evals/`), no user-facing change, so no changelog entry.


markbackman (approved):
LGTM!

PR #4588, opened 2026-05-28
Add AICQuailVADAnalyzer for standalone Quail VAD 2.0; deprecate AICVADAnalyzer

gkmngrgn:
## Summary

- Add `AICQuailVADAnalyzer` (`pipecat.audio.vad.aic_quail_vad`), a noise-robust VAD analyzer powered by the standalone Quail VAD 2.0 model from the ai-coustics SDK. The analyzer owns its own `Processor` (loaded with a dedicated VAD-only model — `quail-vad-2.0-xxs-16khz` by default), so it can sit anywhere in the pipeline, independent of `AICFilter`.
- Deprecate `AICVADAnalyzer` and `AICFilter.create_vad_analyzer()` in favor of the new analyzer. Both surfaces emit `DeprecationWarning` on use and are scheduled for removal in v1.2.0.
- Bump `aic-sdk` to `~=2.3.0` (the release that ships the standalone Quail VAD model and the new model-specific `VadParameter.Sensitivity` semantics — probability threshold in `[0.0, 1.0]` on dedicated VAD models).
- Rename the license environment variable from `AIC_LICENSE_KEY` to `AIC_SDK_LICENSE` so it matches the SDK's canonical name; users must update their `.env` files.

## Test plan

- [x] New unit tests for `AICQuailVADAnalyzer` (mock-based, skip when `aic_sdk` isn't installed) — 16 tests, 98% coverage on the new module.
- [x] New deprecation tests for `AICVADAnalyzer` and `AICFilter.create_vad_analyzer()`.
- [x] Full AIC + neighbor test suites pass against the released `aic-sdk` 2.3.0 (103/103: `test_aic_quail_vad.py`, `test_aic_vad.py`, `test_aic_vad_deprecation.py`, `test_aic_filter_vad_factory_deprecation.py`, `test_aic_filter.py`, `test_krisp_viva_vad.py`).
- [x] `uv run ruff check` and `uv run ruff format --check` pass on touched files.
- [x] `uv run towncrier build --draft --version Unreleased` renders the new fragments under Added / Deprecated / Changed.
- [ ] Smoke-run `examples/voice/voice-aicoustics.py` against a Daily room with `AIC_SDK_LICENSE` set.


gkmngrgn:
Hi @markbackman, I have two questions about the same thing:

1. What's the next release of pipecat including this PR? → I added a deprecation message with a pipecat version, that's why I'm asking it. I set it v1.4.0 for now.

2. Is Pipecat 2.0 an acceptable removal target? We're also ok to remove deprecated module in the next two major version, like 1.6.0.

markbackman:
1. Next release will be this week—v1.4.0.
2. Yes, 2.0 is an acceptable removal.

Is this PR ready for review?

gkmngrgn:
> Is this PR ready for review?

@markbackman not yet, I'll let you know when we completed the tests.

One more question, is there a prevention about making the removal after the next two minor release, for example v1.6.0? We also want to drop it asap and force the users to use the newer vad analyzer implementation.

markbackman:
> One more question, is there a prevention about making the removal after the next two minor release, for example v1.6.0? We also want to drop it asap and force the users to use the newer vad analyzer implementation.

We try to keep breaking changes to major version bumps. That rule primarily applies to the core framework, and not as much to plugins. So, if you have a change that _must_ happen, you can do it before 2.0, since our release schedules are likely not perfectly aligned. Other plugins have had to break in minor version bumps. Just be sure to call this out in the changelog with ⚠️ and mention the breaking change.

src/pipecat/audio/filters/aic_filter.py
```diff
@@ -303,6 +309,10 @@ def create_vad_analyzer(
     ):
         """Return an analyzer that will lazily instantiate the AIC VAD when ready.
 
+        .. deprecated:: <NEXT_RELEASE>
```
andresovela: Is this a placeholder?

src/pipecat/audio/vad/aic_quail_vad.py
```diff
@@ -0,0 +1,365 @@
...
+                ``model_id`` when set.
+            model_download_dir: Directory for downloaded models. Defaults to
+                ``~/.cache/pipecat/aic-models``.
+            speech_hold_duration: Optional override for the SDK's
+                ``VadParameter.SpeechHoldDuration``.
+            minimum_speech_duration: Optional override for the SDK's
+                ``VadParameter.MinimumSpeechDuration``.
+            sensitivity: Optional override for the SDK's
+                ``VadParameter.Sensitivity``. On dedicated Quail VAD models this
+                is a probability in ``[0.0, 1.0]``.
```
andresovela: This class only deals with a dedicated Quail VAD model so this comment is redundant.
```suggestion
                ``VadParameter.Sensitivity``. This is a probability threshold in ``[0.0, 1.0]``. Values above this threshold are considered speech.
```

src/pipecat/audio/vad/aic_quail_vad.py
```diff
@@ -0,0 +1,365 @@
...
+                self._initialize_processor(sample_rate)
+        except Exception:
+            try:
+                self._executor.shutdown(wait=False)
+            except Exception as e:  # noqa: BLE001 - executor cleanup is best-effort
+                logger.debug(f"AICQuailVADAnalyzer executor shutdown failed: {e}")
+            raise
+
+    def _ensure_model_loaded(self) -> None:
+        # We bypass ``AICModelManager`` (used by ``AICFilter``) because each
```
andresovela: I don't understand how is AICFilter relevant here?

src/pipecat/audio/vad/aic_vad.py
```diff
@@ -1,12 +1,17 @@
 """AIC-integrated VAD analyzer that lazily binds to the AIC SDK backend.
 
 This module provides VAD analyzer implementations that query the AIC SDK's
-is_speech_detected() and map it to a float confidence (1.0/0.0).
+``is_speech_detected()`` and map it to a float confidence (1.0/0.0).
+
+.. deprecated:: <NEXT_RELEASE>
```
andresovela: Placeholder?

andresovela (changes requested):

src/pipecat/audio/filters/aic_filter.py
gkmngrgn: Yes — was a placeholder. Per @markbackman's confirmation that v1.4.0 is the next release (and our team's decision to remove these surfaces ahead of 2.0), it's now `.. deprecated:: 1.4.0` with removal targeted at v1.6.0. Fixed in 590dae69f.

src/pipecat/audio/vad/aic_quail_vad.py
gkmngrgn: Applied your suggestion (slightly reflowed for line length). The 'other VAD models' wording was a leftover from the old `AICVADAnalyzer` docstring — this class only handles the Quail VAD model. Fixed in 590dae69f.

src/pipecat/audio/vad/aic_quail_vad.py
gkmngrgn: Fair point — the parenthetical was meant to preempt a 'why not use the shared manager like AICFilter does?' question, but on re-read it just adds noise. Removed the comment in 590dae69f; the implementation reads fine on its own.

src/pipecat/audio/vad/aic_vad.py
gkmngrgn: Yes — same placeholder situation. Now `.. deprecated:: 1.4.0` with removal in v1.6.0 per @markbackman's response in the PR thread. Fixed in 590dae69f.

andresovela (approved):
LGTM

markbackman (approved):
LGTM!

@gkmngrgn anything to change or is this ready to be merged?

gkmngrgn:
@markbackman thanks for patience and helps with your comments. I think that's ready to merge!

PR #4592, opened 2026-05-29
Remove redundant inline [DEPRECATED] tags from LLMUserAggregatorParams docstring

Mudit-Arora:

Per `CONTRIBUTING.md`/`AGENTS.md`, deprecations should use the `.. deprecated::` Sphinx directive — "never inline tags like `[DEPRECATED]`". The two `LLMUserAggregatorParams` fields (`filter_incomplete_user_turns` and `user_turn_completion_config`) carried both an inline `[DEPRECATED]` tag **and** the proper directive.

This removes the redundant inline tags while keeping the `.. deprecated:: 1.2.0` directives and all descriptive text intact. These were the only two remaining violations in the codebase. No behavior change.

A changelog fragment is included.


markbackman (approved):
Nice catch! Thanks 🙇

PR #4593, opened 2026-05-29
feat(assemblyai): add continuous_partials and interruption_delay streaming params

dlange-aai:
#### Summary

Adds two AssemblyAI v3 streaming connection parameters to `AssemblyAISTTService.Settings`, bringing the plugin to parity with the current Universal-Streaming / Universal-3 Pro streaming API:

- **`continuous_partials`** (`bool`, default `True`): emit partial transcripts at a steady cadence during long turns instead of just one early partial near the turn start. Defaulted to `True` so voice agents receive continuous interim updates.
- **`interruption_delay`** (`int`, `0–1000` ms, optional): per-stream override of how soon the first partial is emitted (the server adds 256 ms `MIN_TURN_DURATION_MS` on top). Unset → server default.

Both are **`u3-rt-pro`-only**: gated on the model in `_build_ws_url` and omitted for `universal-streaming-*`. `interruption_delay` is range-validated at construction, with a warning if set on a non-`u3-rt-pro` model.

Notes:
- Added only to the canonical `AssemblyAISTTService.Settings`; the deprecated `AssemblyAIConnectionParams` is intentionally left unchanged (consistent with how `formatted_finals` / `word_finalization_max_wait_time` were handled).
- These are connection params with no response-schema impact, so `models.py` is untouched.

#### Testing
- New `tests/test_assemblyai_stt.py` (10 tests): default-on for `u3-rt-pro`, override to `False`, omission for `universal-streaming`, and `interruption_delay` send/omit/boundary/range-error. All pass; `ruff` clean.
- Manually verified against the live streaming API via the examples: `transcription-assemblyai.py`, `voice-assemblyai.py`, and `voice-assemblyai-turn-detection.py`.

### DIFF
--- a/changelog/4593.added.md
+++ b/changelog/4593.added.md
@@ -1,0 +1,6 @@
+- Added `continuous_partials` and `interruption_delay` connection parameters to
+  the AssemblyAI streaming STT service (`u3-rt-pro` only). `continuous_partials`
+  defaults to `True` so voice agents receive interim transcripts at a steady
+  cadence during long turns; `interruption_delay` (0–1000 ms) overrides how soon
+  the first partial is emitted. Both are exposed via
+  `AssemblyAISTTService.Settings` and are omitted for non-`u3-rt-pro` models.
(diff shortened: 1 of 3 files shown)


markbackman (approved):
LGTM! Thanks for the update 🙇

PR #4597, opened 2026-05-29
Improve user transcripts when Gemini Live is configured for locally-d…

kompfner:
…riven turns

Does two things:
- Gives Gemini Live a "pre-roll" buffer so user transcripts aren't cut off at the beginning
- Updates the Gemini Live locally-driven-turn example to increase VAD `stop_secs`; this breathing room improves transcript accuracy at the end

This PR is the equivalent of the OpenAI-Realtime-focused #4599.


filipi87 (approved):
LGTM!

PR #4598, opened 2026-05-29
fix(gemini): keep tool responses grouped

### LINKED ISSUE #3992: GeminiLLMAdapter merges functionCall messages without merging matching functionResponse messages

he-yufeng:
Fixes #3992.

Gemini thinking mode uses the first function call's thought signature to identify a parallel tool-call group. The current merge code combines later unsigned `function_call` messages into the first model message, but leaves the matching `function_response` messages split as separate user messages. That can produce a Gemini request with one model turn containing multiple calls followed by separate response turns, which Gemini rejects because the function-call turn and response turn no longer match.

This patch keeps the merge local to the current tool-call/response group:

- merge unsigned follow-up `function_call` messages into the signed model turn
- merge the immediately matching `function_response` messages into one user turn
- stop at a new thought signature or an unrelated text/message boundary

Regression coverage covers the parallel case, the sequential case with a fresh signature, and a text boundary that should not be crossed.

To verify:

```bash
uv run --extra google --extra anthropic --extra aws --extra aws-nova-sonic --extra xai --extra perplexity pytest tests/test_get_llm_invocation_params.py -q
uv run ruff check src/pipecat/adapters/services/gemini_adapter.py tests/test_get_llm_invocation_params.py
uv run ruff format --check src/pipecat/adapters/services/gemini_adapter.py tests/test_get_llm_invocation_params.py
git diff --check
```


2010b9:
@markbackman is this something that you are considering merging?
I had the same problem as @he-yufeng, so I cherry-picked their fix and it solved the problem.

AndreTeixeira1998:
> @markbackman is this something that you are considering merging? I had the same problem as @he-yufeng, so I cherry-picked their fix and it solved the problem.

I’m seeing the same issue as well

markbackman:
cc @kompfner to take a look

kompfner:
Pardon the delay, and thanks for the contribution!

There had actually been a previous PR (#4103) fixing this issue, which somehow got left behind in the day-to-day shuffle. I just went ahead and got that one past the finish line, so the issue should be resolved now.

PR #4599, opened 2026-05-29
Improve user transcripts when OpenAI Realtime is configured for locally-driven turns

kompfner:
Does two things:
- Gives OpenAI Realtime a "pre-roll" buffer so user transcripts aren't cut off at the beginning
- Updates the OpenAI Realtime locally-driven-turn example to increase VAD `stop_secs`; this breathing room improves transcript accuracy at the end

This PR is the equivalent of the Gemini-Live-focused #4597. Note that the behavioral improvement here isn't _nearly_ as dramatic—the pre-roll and elongated VAD `stop_secs` seem to help a _bit_, but in my testing there are still wonky transcripts that clearly don't match what the model "hears" and responds to.


examples/realtime/realtime-openai-locally-driven-turns.py
```diff
@@ -200,7 +201,12 @@ async def run_bot(transport: BaseTransport, runner_args: RunnerArguments):
         # correct and (by default) drops the transcript wait on turn-end so
         # local VAD can drive turn boundaries on the latency critical path.
         realtime_service_mode=True,
-        user_params=LLMUserAggregatorParams(vad_analyzer=SileroVADAnalyzer()),
+        user_params=LLMUserAggregatorParams(
+            # stop_secs is intentionally longer than Pipecat's 0.2s default so
+            # brief mid-utterance pauses don't prematurely commit the user turn
+            # and trigger a response while the user is still speaking.
+            vad_analyzer=SileroVADAnalyzer(params=VADParams(stop_secs=0.5)),
```
filipi87, line 208: Wouldn't the `LocalSmartTurnAnalyzerV3` already prevent this ?

filipi87 (approved):
LGTM! Just one question regarding changing the default value in the example. 🚀

examples/realtime/realtime-openai-locally-driven-turns.py
kompfner: The explanation here maybe isn't totally accurate. It's not *only* about avoiding mid-utterance pauses (which this tweak still helps with a bit, as smart turn isn't 100%)—it's also about guaranteeing that the audio we end up sending has enough "breathing room" at the end. This is something that the Gemini Live folks recommended, and at least appears to help somewhat helpful, based on my unscientific observations. So we're applying the same tweak here, too, where, again, based on unscientific messing around, it might help.

I'll update this comment for accuracy.

kompfner: Updated: https://github.com/pipecat-ai/pipecat/pull/4599/changes/0872ed842bbd73822834366a0141450894c7b4d4

filipi87: Oo, I see, makes more sense now. Thank you Paul.

Issue #4409, opened 2026-05-03
DeepgramTTSService WebSocket not fully torn down after pipeline ends


ValleZ:
### pipecat version

0.0.108

### Python version

3.13.5

### Operating System

Linux (Debian 13, x86_64) 

### Issue description


**File:** `pipecat/services/deepgram/tts.py` — `stop()` / `_disconnect()` teardown path  
**Related:** `pipecat/services/websocket_service.py` — `_maybe_try_reconnect`

**Severity:** Low — no functional impact on completed calls, but leaks an open WebSocket connection and generates hourly error noise.

### What happens

After a pipeline runs to completion (EndFrame processed, `runner.run()` returns), the `DeepgramTTSService`'s underlying WebSocket connection remains open. Deepgram times out idle connections with a 1011 internal error (`NET-0003`) after ~60 minutes. Because `_disconnecting` is `False` at that point (suggesting `_disconnect()` was either not called or its effect didn't persist to the receive task), `_maybe_try_reconnect` treats this as a recoverable error, reconnects, and the cycle repeats every hour until the process restarts.

#### Suspected cause

Possible race between the receive task being cancelled and Deepgram sending its 1011 — or `_disconnect()` completing but `_disconnecting` being reset before the 1-hour timeout fires. Needs reproduction with a controlled sleep/disconnect sequence.



### Reproduction steps

  1. Create a pipeline with DeepgramTTSService and RTVIProcessor                                                                                                                                                                                                                                                   
  2. Run a call to completion (pipeline processes EndFrame naturally — e.g. agent says goodbye and hangs up)                                                                                                                                                                                                       
  3. Confirm runner.run() returns and the "Finally block" is reached in logs
  4. Wait ~60 minutes
  5. Observe logs 

### Expected behavior


`stop(EndFrame)` → `_disconnect()` sets `_disconnecting = True` and cancels `_receive_task`. After that, any close event from Deepgram should hit the `_disconnecting` guard in `_maybe_try_reconnect` and exit the receive loop without reconnecting.


### Actual behavior

  After runner.run() returns, DeepgramTTSService's WebSocket connection remains open.
  Deepgram times out the idle connection after ~60 minutes with a 1011 internal error (NET-0003).
  Because _disconnecting is False at that point, _maybe_try_reconnect treats it as a recoverable
  error, reconnects successfully, and the cycle repeats every hour indefinitely until the process
  restarts.



### Logs

```shell
[...]

PR #4601, opened 2026-05-29
fix: harden Deepgram TTS disconnect cleanup

he-yufeng:
﻿## Summary
- close the Deepgram TTS websocket before awaiting receive-task cancellation
- keep receive-task cancellation and reference cleanup in a finally block
- add regression coverage for cancellation during disconnect cleanup

Addresses #4409.

## To verify
- python -m pytest tests\test_deepgram_tts.py -q
- python -m py_compile src\pipecat\services\deepgram\tts.py tests\test_deepgram_tts.py
- python -m ruff check src\pipecat\services\deepgram\tts.py tests\test_deepgram_tts.py
- python -m ruff format --check src\pipecat\services\deepgram\tts.py tests\test_deepgram_tts.py
- git diff --check

### DIFF
--- a/src/pipecat/services/deepgram/tts.py
+++ b/src/pipecat/services/deepgram/tts.py
@@ -184,11 +184,16 @@ class DeepgramTTSService(WebsocketTTSService):
         """Disconnect from Deepgram WebSocket and clean up tasks."""
         await super()._disconnect()
 
-        if self._receive_task:
-            await self.cancel_task(self._receive_task)
-            self._receive_task = None
-
-        await self._disconnect_websocket()
+        receive_task = self._receive_task
+        try:
+            await self._disconnect_websocket()
+        finally:
+            if receive_task:
+                try:
+                    await self.cancel_task(receive_task)
+                finally:
+                    if self._receive_task is receive_task:
+                        self._receive_task = None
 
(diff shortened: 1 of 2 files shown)


he-yufeng:
First nudge on this one. The disconnect path can wedge: the receive task is awaited before the websocket is closed, so a hung receive blocks cleanup indefinitely. The patch closes the socket first and moves cancellation/reference cleanup into a finally block, with a regression test for cancellation mid-disconnect. CI green.

markbackman:
@he-yufeng is this still an issue. The corresponding issue was closed with this rationale:
https://github.com/pipecat-ai/pipecat/issues/4409#issuecomment-4909217935

Can you repro the issue still on 1.5.0 or later?

markbackman:
Thanks for the follow-up, @he-yufeng. I don't think this one is needed anymore — here's where things landed:

The original leak (#4409) is fixed by #4902, released in v1.5.0. `cleanup()` is now the guaranteed teardown hook, and `WebsocketTTSService.cleanup()` calls `_disconnect()`, so the socket is released even when the frame-driven `stop()`/`cancel()` path gets interrupted — which was the race in that issue.

Closing for now — if you can still reproduce a stale connection on 1.6.0, please reopen with the repro and we'll dig in.

PR #3038, opened 2025-11-12
Deepgram Flux improvements

filipi87:
Deepgram Flux improvements:
- Extracted the logic for retrying connections, and create a new `send_with_retry` method inside `WebSocketService`.
- Refactored `DeepgramFluxSTTService` to automatically reconnect if sending a message fails.
- Added a watchdog to `Flux` to prevent dangling tasks.
- Introduced a minimum confidence parameter in `DeepgramFluxSTTService` to avoid generating transcriptions below a defined threshold.

PR #4608, opened 2026-06-01
Fix unhandled exception in WebsocketService.send_with_retry retry path Closes #4618

### LINKED ISSUE #4618: WebsocketService.send_with_retry: retry send after reconnect is unhandled and crashes services on the audio hot path
### pipecat version

1.3.0

### Python version

3.11.15

### Operating System

macOS 15.3

### Issue description

In WebsocketService.send_with_retry (src/pipecat/services/websocket_service.py), when an initial send fails and a reconnect succeeds, the retry send is performed outside any try/except. If that retry also fails — a likely outcome right after a
flaky reconnect (server immediately closes the socket, second transient error) — the exception propagates out of send_with_retry unhandled, violating the method's "handle retries internally" contract.

Affected code (before fix):
async def send_with_retry(self, message, report_error):
try:
if self._websocket is None:
raise ConnectionError(f"{self} no websocket connected")
await self._websocket.send(message)
except Exception as e:
logger.error(f"{self} send failed: {e}, will try to reconnect")
success = await self._try_reconnect(report_error=report_error)
if success and self._websocket is not None:
logger.info(f"{self} reconnected successfully, will retry send the message")
await self._websocket.send(message) # ← not inside try/except
else:
logger.error(f"{self} send failed; unable to reconnect")

Why it's a bug

send_with_retry is contractually expected to either succeed or report the failure via the report_error callback (the same channel _try_reconnect uses). The unguarded retry on line 138 breaks that: a second send failure raises a raw exception
out of a method named "with retry."

Impact

    Callers run on the hot audio path: inworld/tts.py (4 call sites), cartesia/turns/stt.py (audio + silence sends), deepgram/flux/stt.py.
    An unhandled exception there propagates into the service's send task and can tear down STT/TTS mid-conversation instead of degrading gracefully.
    The failure triggers precisely during unstable-network conditions — the exact scenario the retry logic exists to handle — so it bites when reliability matters most.

### Reproduction steps

 Subclass WebsocketService with a fake _websocket whose send raises on the first call, whose _reconnect_websocket succeeds, and whose send raises again on the second (retry) call.
    await service.send_with_retry(msg, report_error).
    Observe the exception propagates out instead of being caught and reported.

### Expected behavior

send_with_retry() should never let an exception escape to its caller. Its
contract is to either send the message successfully or report the failure via
the report_error callback (an ErrorFrame). After a reconnect succeeds, if the
single retry send also fails, that failure should be caught and reported the
same way — so the calling STT/TTS service can degrade gracefully and keep the
pipeline alive.

### Actual behavior

When the initial send fails, send_with_retry() reconnects and retries the send.
That retry is performed OUTSIDE the try/except block, so if it also fails (e.g.
the server immediately closes the freshly reconnected socket, or a second
transient error occurs), the exception propagates unhandled out of
send_with_retry().

Because callers invoke this on the audio hot path (inworld/tts.py,
cartesia/turns/stt.py, deepgram/flux/stt.py), the unhandled exception crashes
the service's send task mid-conversation instead of being reported as an
ErrorFrame — and it happens precisely during unstable-network conditions, which
is the exact scenario the retry logic exists to handle.

If the form also has a Steps to reproduce field, use:

    Subclass WebsocketService with a fake _websocket whose send() raises on the
    first call, whose _reconnect_websocket() succeeds, and whose send() raises
    again on the second (retry) call.
    await service.send_with_retry(message, report_error)
    The exception from the retry send propagates out instead of being caught and
    reported via report_error.

### Logs

```shell

```

kavya-shree-s:
Fix unhandled exception in WebsocketService.send_with_retry retry path

  The retry send after a successful reconnect was outside the try/except,
  so a second send failure (e.g. server closing the freshly reconnected
  socket) propagated unhandled out of send_with_retry, crashing services
  on the audio hot path (Inworld TTS, Cartesia STT, Deepgram Flux STT).

  Wrap the retry send in its own try/except and report the failure via the
  report_error callback, consistent with the other error paths.

### DIFF
--- a/src/pipecat/services/websocket_service.py
+++ b/src/pipecat/services/websocket_service.py
@@ -134,10 +134,21 @@ class WebsocketService(ABC):
             success = await self._try_reconnect(report_error=report_error)
             if success and self._websocket is not None:
                 logger.info(f"{self} reconnected successfully, will retry send the message")
-                # trying to send the message one more time
-                await self._websocket.send(message)
+                # Trying to send the message one more time. This retry must stay
+                # inside its own try/except: if the send fails again (e.g. the
+                # server closes the freshly reconnected socket), the exception
+                # would otherwise escape send_with_retry unhandled and crash the
+                # calling service on the audio hot path.
+                try:
+                    await self._websocket.send(message)
+                except Exception as retry_error:
+                    logger.error(f"{self} send failed after reconnect: {retry_error}")
+                    await report_error(
+                        ErrorFrame(f"{self} send failed after reconnect: {retry_error}")
+                    )
             else:
                 logger.error(f"{self} send failed; unable to reconnect")
+                # _try_reconnect already reported the failure via report_error.
 
     async def _maybe_try_reconnect(
         self,
--- a/tests/test_websocket_service.py
+++ b/tests/test_websocket_service.py
@@ -256,7 +256,25 @@ async def test_stable_connection_resets_quick_failure_counter(service, report_er
 # ---------------------------------------------------------------------------
 # Lifecycle and guards
 # ---------------------------------------------------------------------------
+@pytest.mark.asyncio
+async def test_send_with_retry_reports_retry_send_failure(service, report_error):
+    """A failed send after reconnect is reported instead of escaping."""
+    websocket = type("FakeWebsocket", (), {})()
+    websocket.send = AsyncMock(
+        side_effect=[ConnectionError("initial send failed"), RuntimeError("retry failed")]
+    )
+    service._websocket = websocket
+    service._try_reconnect = AsyncMock(return_value=True)
 
+    await service.send_with_retry("message", report_error)
+
+    assert websocket.send.call_count == 2
+    service._try_reconnect.assert_called_once()
+    report_error.assert_called_once()
+    error_frame = report_error.call_args[0][0]
+    assert isinstance(error_frame, ErrorFrame)
+    assert error_frame.fatal is False
+    assert "retry failed" in error_frame.error
 
 @pytest.mark.asyncio
 async def test_disconnect_prevents_reconnection(service, report_error):


kavya-shree-s:
@aconchillo @markbackman  @kompfner 

Please review this PR

kavya-shree-s:
fix: #4618

kompfner (changes requested):
Based on my reading of `report_error` (including how it's used throughout the codebase, and how it was used [when first introduced](https://github.com/pipecat-ai/pipecat/pull/3038/changes)), the intent/contract is for it to be only for reporting *connection* errors (i.e. failure to connect).

For message-send errors, the contract seems to be: the Exception will bubble up and should be handled by the caller. Check out the call pattern in Deepgram Flux [here](https://github.com/pipecat-ai/pipecat/blob/5e87c5e2f9037f4f1a844b8a5c984d4704f5895b/src/pipecat/services/cartesia/turns/stt.py#L362-L368) or in Cartesia Turns [here](https://github.com/pipecat-ai/pipecat/blob/5e87c5e2f9037f4f1a844b8a5c984d4704f5895b/src/pipecat/services/cartesia/turns/stt.py#L362-L368).

As is, this change would cause message-send errors to be erroneously treated as connection errors.

I think the right approach would be to:
- Add proper error handling wherever it's missing, in the callers' code (in the services themselves)
- Probably rename the `report_error` param in this method to `report_connection_error`, to avoid confusing readers in the future
  - Also probably rename it in `_receive_task_handler`
  - It doesn't need to be renamed in the other `WebsocketService` methods that have the same parameter, because those methods are fundamentally _about_ establishing a connection

ianbbqzy:
Thanks for this fix! One related issue worth considering while you're in this area: when `_receive_task_handler` and `send_with_retry` both detect a disconnect simultaneously, `_try_reconnect` returns `False` immediately for the second caller (the current "reconnect attempt aborted: already in progress" path). Since `send_with_retry` treats any `False` as a hard failure, it logs `"send failed; unable to reconnect"` and drops the message — even though the background reconnect may succeed moments later.

A minimal fix would be to wait on an `asyncio.Event` instead of returning `False` immediately:

```python
def __init__(self, ...):
    self._reconnect_in_progress: bool = False
    self._reconnect_complete: asyncio.Event = asyncio.Event()
    self._reconnect_complete.set()  # "idle" initially

async def _try_reconnect(self, ...):
    if self._reconnect_in_progress:
        logger.warning(f"{self} reconnect already in progress, waiting for completion")
        await self._reconnect_complete.wait()
        return self._websocket is not None  # report the outcome

    self._reconnect_in_progress = True
    self._reconnect_complete.clear()
    try:
        # ... existing retry loop ...
    finally:
        self._reconnect_in_progress = False
        self._reconnect_complete.set()
```

This is race-free in asyncio since there's no `await` between the `_reconnect_in_progress` check and the flag assignment. Happy to open a separate PR for this if it sounds useful.

filipi87:
Hi @kavya-shree-s, thanks for this PR! The issue it references (#4618) is already closed, so this looks like it may be stale or superseded by other work.

PR #4612, opened 2026-06-01
feat(smallest): add word_timestamps support to TTS service

harshitajain165:
## Summary

- Added `word_timestamps` opt-in setting to `SmallestTTSService` — sends `word_timestamps: true` in the WebSocket request payload when enabled, receiving interleaved per-word timing events (`id`, `word`, `start`, `end`) alongside audio chunks
- Updated `voice-smallest.py` example to use `lightning_v3.1_pro` model with `meher` voice

## word_timestamps feature

Smallest AI's Lightning v3.1 and v3.1 Pro WebSocket streaming now supports per-word timing events (shipped 2026-06-01). Opt in by setting `word_timestamps=True` in the service settings:

```python
tts = SmallestTTSService(
    api_key=os.environ["SMALLEST_API_KEY"],
    settings=SmallestTTSService.Settings(
        model="lightning_v3.1_pro",
        voice="meher",
        word_timestamps=True,
    ),
)
```

When enabled, the server interleaves `status: "word_timestamp"` frames with audio chunks. Each frame carries `data: { id, word, start, end }` where `start`/`end` are floats in seconds relative to the start of the audio stream, and `word` is the verbatim input text substring (un-normalized — `"$100"` stays `"$100"`).

Supported voices: base-queue English + Hindi voices (`meher`, `devansh`, `kartik`, `maithili`, `liam`, `avery`). Other voice families silently emit no word events — audio works normally.

Defaults to `None` (disabled) — purely opt-in, no behavior change for existing integrations.

## Voice/model fix

`lightning_v3.1_pro` is now the default model and uses `meher` as its default voice. The example was still using `sophia` (the default for the older `lightning_v3.1` model), creating a mismatch. Both `model` and `voice` are now set explicitly.

### DIFF
--- a/changelog/4612.added.md
+++ b/changelog/4612.added.md
@@ -1,0 +1,8 @@
+- Added word-level timestamp support to `SmallestTTSService`. Enabled by default
+  via the `word_timestamps` constructor argument, it emits per-word
+  `TTSTextFrame`s aligned to audio playback so downstream consumers (captions,
+  lip-sync, RTVI) receive word timing. Timestamps from each TTS request are
+  offset onto the turn's continuous playback timeline, so multi-sentence turns
+  stay correctly ordered. Available on Smallest's word-timestamp-capable voices;
+  other voices simply emit no word events, so leaving it on is safe. Pass
+  `word_timestamps=False` to fall back to whole-text frames.
--- a/examples/voice/voice-smallest.py
+++ b/examples/voice/voice-smallest.py
@@ -61,7 +61,8 @@ async def run_bot(transport: BaseTransport, runner_args: RunnerArguments):
     tts = SmallestTTSService(
         api_key=os.environ["SMALLEST_API_KEY"],
         settings=SmallestTTSService.Settings(
-            voice="sophia",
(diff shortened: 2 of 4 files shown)


markbackman (approved):
This wasn't working quite right, so I spent a few minutes getting things configured correctly. The key changes:
- `word_timestamps` was moved out of `SmallestTTSSettings` and into `__init__`. It's an init time only config, as the base class needs to know whether the class itself is pushing timestamps.
- Default `word_timestamps` to True, which is superior for interruption handling.
- The `push_text_frames` needs to be set according to `word_timestamps`.
- Correctly the timestamp handling.

One important thing I noticed. The timing of words and audio are not synchronized. The word timestamps are slightly ahead. See this video:
https://github.com/user-attachments/assets/80b711d5-9182-46ab-8d0c-8fe03966d1c0

You can replicate by running the example and watching the pacing of words vs audio.

Nonetheless, approving since I think this is a nice improvement. The model changes can be made on your end, independent of this feature.

markbackman:
I'm going to merge this down, but please let me know if you disagree with any of the changes I made.

PR #4615, opened 2026-06-01
create_transport() refactor

markbackman:
### My overview

The goal of this change is to simplify specifying the transport down to a single utility: `create_transport()`. This was largely possible already with the exception of two cases:
1. Daily PSTN Dial-in: A special Daily Param arg is required, which was being special cased. That's been folded in.
2. Websocket transports required a different pattern if you wanted to parse the websocket message for call data (e.g. to and from numbers).

Now, you can write your bot() function like this for essentially all transports:
```python
async def bot(runner_args: RunnerArguments, testing: bool | None = False):

    transport_params = {
        "twilio": lambda: FastAPIWebsocketParams(
            audio_in_enabled=True,
            audio_out_enabled=True,
        ),
    }

    transport = await create_transport(runner_args, transport_params)

    await run_bot(transport, runner_args)
```

1 remaining caveat:
- Daily dial-out and SIP use cases require a server to create a room in a different way, so I'm leaving those two cases out for now...

---

### Claude's overview

Makes `create_transport()` the single, feature-complete way to build any transport, so bots don't hand-roll transport construction or telephony parsing. Public signatures are unchanged (`create_transport -> BaseTransport`, `parse_telephony_websocket -> (type, call_data)`).

### `create_transport` feature-complete for telephony and eval

- **Typed `CallData` model** for the parsed telephony handshake: attribute access (`call_data.to_number`, with `from`/`to` aliased off the Python-keyword wire keys) that stays dict-compatible (`call_data["call_id"]`, `.get(...)`) so bots written against the old dict keep working. `parse_telephony_websocket` returns it; the Twilio branch promotes `from`/`to` custom params for a uniform API.
- **`call_data` on the base `RunnerArguments`**, next to `body`, so every bot reads `runner_args.call_data` uniformly. `transport_type` is overwritten with the detected provider so bots can personalize without re-parsing the single-use stream.
- **Idempotent `parse_telephony_websocket`** — caches the result on the websocket; only successful parses are cached, so failures stay retryable and a verbose bot calling it directly coexists with `create_transport`'s internal call.
- **Transparent Daily PSTN dial-in** via `_maybe_apply_daily_dialin`: applies `dialin_settings`/`api_key`/`api_url` from the request body in the Daily branch; no-op when the body carries no dial-in.

### Mount the Daily dial-in webhook by default for `-t daily`

- `--dialin` becomes a `BooleanOptionalAction` defaulting `True` (so `bot.py -t daily` supports inbound PSTN testing out of the box); `--no-dialin` opts out. The webhook is mounted only inside `_setup_daily_routes`, so it's a no-op for non-Daily transports — the old "`--dialin` only works with `-t daily`" validation is dropped.

Local-runner convenience only; the bot contract, `runner_args` shape, and Cloud behavior are unchanged.

### Tests

Adds coverage in `tests/test_runner_utils.py` for `CallData` attribute/dict access and aliasing, `call_data` on the base, idempotent parse, and dial-in body application.

### DIFF
--- a/src/pipecat/runner/run.py
+++ b/src/pipecat/runner/run.py
@@ -1207,7 +1207,8 @@ def main(parser: argparse.ArgumentParser | None = None):
        - -x/--proxy: Public proxy hostname for telephony webhooks
        - -d/--direct: Connect directly to Daily room (automatically sets transport to daily)
        - -f/--folder: Path to downloads folder
-       - --dialin: Enable Daily PSTN dial-in webhook handling
+       - --dialin/--no-dialin: Mount the Daily PSTN dial-in webhook for -t daily
+         (on by default; --no-dialin disables it)
        - --esp32: Enable SDP munging for ESP32 compatibility (requires --host with IP address)
        - --whatsapp: Ensure required WhatsApp environment variables are present
        - -v/--verbose: Increase logging verbosity
@@ -1250,9 +1251,12 @@ def main(parser: argparse.ArgumentParser | None = None):
     )
     parser.add_argument(
         "--dialin",
-        action="store_true",
-        default=False,
-        help="Enable Daily PSTN dial-in webhook handling",
+        action=argparse.BooleanOptionalAction,
+        default=True,
+        help=(
+            "Mount the Daily PSTN dial-in webhook for -t daily. On by default (a local "
+            "stand-in for Pipecat Cloud's dial-in handler); use --no-dialin to disable."
+        ),
     )
     parser.add_argument(
         "--esp32",
@@ -1286,10 +1290,8 @@ def main(parser: argparse.ArgumentParser | None = None):
         logger.error("For ESP32, you need to specify `--host IP` so we can do SDP munging.")
         return
 
-    # Validate dial-in requirements
-    if args.dialin and args.transport is not None and args.transport != "daily":
-        logger.error("--dialin flag only works with Daily transport (-t daily)")
-        return
+    # The dial-in webhook is mounted only inside _setup_daily_routes, so --dialin is a
+    # no-op for non-Daily transports; nothing to validate here.
 
     # Log level
     logger.remove()
--- a/src/pipecat/runner/types.py
+++ b/src/pipecat/runner/types.py
@@ -15,7 +15,7 @@ from dataclasses import dataclass, field
 from typing import Any
 
 from fastapi import WebSocket
-from pydantic import BaseModel
+from pydantic import BaseModel, ConfigDict, Field
 
 
 class DialinSettings(BaseModel):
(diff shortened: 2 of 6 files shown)


markbackman:
@aconchillo I rebased this on `origin/main` given our new direction for evals. We can make any subsequent changes need after this is merged.

src/pipecat/runner/run.py
```diff
@@ -1250,9 +1251,12 @@ def main(parser: argparse.ArgumentParser | None = None):
     )
     parser.add_argument(
         "--dialin",
-        action="store_true",
-        default=False,
-        help="Enable Daily PSTN dial-in webhook handling",
+        action=argparse.BooleanOptionalAction,
```
filipi87, line 1254: Oo, I see, this is what makes the `no-dialin` option. Cool.

src/pipecat/runner/types.py
```diff
@@ -56,6 +56,63 @@ class DailyDialinRequest(BaseModel):
...
+    Fields are populated per provider; absent ones stay ``None``. Provider-specific keys
+    not modeled here remain accessible (``extra="allow"``).
+
+    Parameters:
+        stream_id: Provider media-stream identifier.
+        call_id: Provider call identifier (Twilio/Plivo/Exotel; Telnyx uses
+            ``call_control_id``).
+        from_number: Caller's number. Wire key ``from``.
+        to_number: Dialed number. Wire key ``to``.
+        call_control_id: Telnyx call control id.
```
filipi87: Even if Telnyx internally uses `call_control_id`, in this object, which is intended to normalize data across all telephony providers, couldn't we call it `call_id` ? And remove this `call_control_id` ?

src/pipecat/runner/types.py
```diff
@@ -56,6 +56,63 @@ class DailyDialinRequest(BaseModel):
...
+        outbound_encoding: Telnyx outbound media encoding.
+        account_sid: Exotel account sid.
+        body: Custom parameters sent by the provider (e.g. Twilio TwiML stream
+            parameters).
+        custom_parameters: Exotel custom parameters.
```
filipi87: What if we isolated provider specific fields into their own subclasses? For example:

```
class TelnyxCallData(CallData):
    outbound_encoding: str | None = None

class ExotelCallData(CallData):
    account_sid: str | None = None
    custom_parameters: str | dict | None = None
```

What do you think ?

src/pipecat/runner/utils.py
```diff
@@ -147,9 +151,23 @@ async def parse_telephony_websocket(websocket: WebSocket):
...
+    function again (e.g. once inside ``create_transport`` and once in bot code)
+    returns the same ``(transport_type, call_data)`` without re-consuming it.
     """
+    # Return the cached parse if this websocket has already been parsed — the
+    # message stream below can only be consumed once. The cache is always a
+    # (transport_type, call_data) tuple; isinstance keeps this robust against
+    # mock/auto-attr websockets in tests.
+    cached = getattr(websocket, "_pipecat_parsed_telephony", None)
+    if isinstance(cached, tuple):
+        return cached
```
filipi87, line 171: Is this only for testing, or can this function also be called multiple times in production such that we need this ?

If it's only for testing, maybe we should remove it from here and have the tests that need repeated access store the result themselves.

What do you think?

filipi87 (commented):
The changes look good. I have added a couple of questions.

Also, it may be worth double checking whether any of these changes require updates to our Pipecat Cloud base image. At first glance, I don't think they do, but I wanted to mention it just to be safe.

src/pipecat/runner/utils.py
markbackman: Good question, but it is not test-only. `parse_telephony_websocket` is public and was the documented pre-`create_transport` pattern, and the underlying message stream is single-use. `create_transport` now calls it internally, so if a bot also calls it directly (or it gets called twice for any reason), the second call would hit an exhausted stream without this cache. It is a cheap getattr/setattr that keeps the public function idempotent. Happy to drop it if we decide direct calls are no longer supported (bots can read `runner_args.call_data` instead) — let me know your preference.

markbackman:
Thanks for the review!

On the base image: confirmed, no Pipecat Cloud base image changes are needed. These changes are additive — a new optional `call_data` field on the base `RunnerArguments` (default `None`), the new `CallData` model, the idempotent parse, and transparent dial-in. The base image constructs the `*SessionArguments` (runner-type subclasses) and never touches `CallData` / `parse_telephony_websocket`, so nothing there is affected.

Also addressed the `CallData` feedback:
- Normalized Telnyx `call_control_id` onto the common `call_id` field and dropped the separate field (eecf217).
- Moved provider-specific fields into `TelnyxCallData` / `ExotelCallData` subclasses; `parse_telephony_websocket` builds the right one per provider (c928324).

Left a reply on the `parse_telephony_websocket` caching question.

src/pipecat/runner/types.py
markbackman: Done!

src/pipecat/runner/types.py
markbackman: Agreed. Done!

src/pipecat/runner/utils.py
filipi87: If this was the recommended approach before, it's probably best to keep the cache for now, so it still works if the method is called multiple times. 👍

filipi87 (approved):
LGTM! 🚀

PR #4617, opened 2026-06-02
fix: report websocket retry send failures

### LINKED ISSUE #4607: WebsocketService.send_with_retry: retry send after reconnect is unhandled and crashes services on the audio hot path
### pipecat version

1.3.0

### Python version

3.11.15

### Operating System

macOS 15.3

### Issue description

In WebsocketService.send_with_retry (src/pipecat/services/websocket_service.py), when an initial send fails and a reconnect succeeds, the retry send is performed outside any try/except. If that retry also fails — a likely outcome right after a
  flaky reconnect (server immediately closes the socket, second transient error) — the exception propagates out of send_with_retry unhandled, violating the method's "handle retries internally" contract.

  Affected code (before fix):
  async def send_with_retry(self, message, report_error):
      try:
          if self._websocket is None:
              raise ConnectionError(f"{self} no websocket connected")
          await self._websocket.send(message)
      except Exception as e:
          logger.error(f"{self} send failed: {e}, will try to reconnect")
          success = await self._try_reconnect(report_error=report_error)
          if success and self._websocket is not None:
              logger.info(f"{self} reconnected successfully, will retry send the message")
              await self._websocket.send(message)      # ← not inside try/except
          else:
              logger.error(f"{self} send failed; unable to reconnect")

  Why it's a bug

  send_with_retry is contractually expected to either succeed or report the failure via the report_error callback (the same channel _try_reconnect uses). The unguarded retry on line 138 breaks that: a second send failure raises a raw exception
  out of a method named "with retry."

Impact

  - Callers run on the hot audio path: inworld/tts.py (4 call sites), cartesia/turns/stt.py (audio + silence sends), deepgram/flux/stt.py.
  - An unhandled exception there propagates into the service's send task and can tear down STT/TTS mid-conversation instead of degrading gracefully.
  - The failure triggers precisely during unstable-network conditions — the exact scenario the retry logic exists to handle — so it bites when reliability matters most.

### Reproduction steps

 1. Subclass WebsocketService with a fake _websocket whose send raises on the first call, whose _reconnect_websocket succeeds, and whose send raises again on the second (retry) call.
  2. await service.send_with_retry(msg, report_error).
  3. Observe the exception propagates out instead of being caught and reported.

### Expected behavior

send_with_retry() should never let an exception escape to its caller. Its
  contract is to either send the message successfully or report the failure via
  the report_error callback (an ErrorFrame). After a reconnect succeeds, if the
  single retry send also fails, that failure should be caught and reported the
  same way — so the calling STT/TTS service can degrade gracefully and keep the
  pipeline alive.

### Actual behavior

When the initial send fails, send_with_retry() reconnects and retries the send.
  That retry is performed OUTSIDE the try/except block, so if it also fails (e.g.
  the server immediately closes the freshly reconnected socket, or a second
  transient error occurs), the exception propagates unhandled out of
  send_with_retry().

  Because callers invoke this on the audio hot path (inworld/tts.py,
  cartesia/turns/stt.py, deepgram/flux/stt.py), the unhandled exception crashes
  the service's send task mid-conversation instead of being reported as an
  ErrorFrame — and it happens precisely during unstable-network conditions, which
  is the exact scenario the retry logic exists to handle.

  If the form also has a Steps to reproduce field, use:
  1. Subclass WebsocketService with a fake _websocket whose send() raises on the
     first call, whose _reconnect_websocket() succeeds, and whose send() raises
     again on the second (retry) call.
  2. await service.send_with_retry(message, report_error)
  3. The exception from the retry send propagates out instead of being caught and
     reported via report_error.

### Logs

```shell

```

he-yufeng:
﻿## Summary
- catch failures from the retry send after a successful reconnect
- report that second failure through the existing ErrorFrame path instead of letting it escape the caller
- add a WebsocketService regression test for the double-send-failure path

## To verify
- uv run pytest tests\test_websocket_service.py -q --basetemp .tmp\pytest
- uv run ruff check src\pipecat\services\websocket_service.py tests\test_websocket_service.py
- uv run ruff format --check src\pipecat\services\websocket_service.py tests\test_websocket_service.py
- uv run python -m py_compile src\pipecat\services\websocket_service.py tests\test_websocket_service.py
- git diff --check

Fixes #4607

### DIFF
--- a/src/pipecat/services/websocket_service.py
+++ b/src/pipecat/services/websocket_service.py
@@ -135,7 +135,12 @@ class WebsocketService(ABC):
             if success and self._websocket is not None:
                 logger.info(f"{self} reconnected successfully, will retry send the message")
                 # trying to send the message one more time
-                await self._websocket.send(message)
+                try:
+                    await self._websocket.send(message)
+                except Exception as retry_error:
+                    msg = f"{self} send failed after reconnect: {retry_error}"
+                    logger.error(msg)
+                    await report_error(ErrorFrame(msg))
             else:
                 logger.error(f"{self} send failed; unable to reconnect")
 
--- a/tests/test_websocket_service.py
+++ b/tests/test_websocket_service.py
@@ -259,6 +259,27 @@ async def test_stable_connection_resets_quick_failure_counter(service, report_er
 
 
 @pytest.mark.asyncio
+async def test_send_with_retry_reports_retry_send_failure(service, report_error):
+    """A failed send after reconnect is reported instead of escaping."""
+    websocket = type("FakeWebsocket", (), {})()
+    websocket.send = AsyncMock(
+        side_effect=[ConnectionError("initial send failed"), RuntimeError("retry failed")]
+    )
+    service._websocket = websocket
+    service._try_reconnect = AsyncMock(return_value=True)
+
+    await service.send_with_retry("message", report_error)
+
+    assert websocket.send.call_count == 2
+    service._try_reconnect.assert_called_once()
+    report_error.assert_called_once()
+    error_frame = report_error.call_args[0][0]
+    assert isinstance(error_frame, ErrorFrame)
+    assert error_frame.fatal is False
+    assert "retry failed" in error_frame.error
+
+
+@pytest.mark.asyncio
 async def test_disconnect_prevents_reconnection(service, report_error):
     """After _disconnect(), errors exit the loop without reconnecting or emitting errors."""
     await service._disconnect()


filipi87:
Hi @he-yufeng, thanks for this PR! The issue it references (#4607) is already closed, so this looks like it may be stale or superseded by other work.

he-yufeng:
Checking in on this one. Retry sends that fail inside the backoff loop get swallowed today, so a dead websocket looks like a slow one. The patch surfaces the failure on the first failed send instead, with coverage for the retry path.

PR #4619, opened 2026-06-02
bus: stop relaying emitter-local output frames across the bus

Maurotb:
## Summary

`BusBridgeProcessor` and `_BusEdgeProcessor` currently relay every frame across the bus by default, including frames that are semantically local to the emitting worker. In a multi-worker setup with `BusBridgeProcessor`, sibling tasks end up processing the same frame instance multiple times.

Observed in production (pipecat 1.3.0 + 1.3.1.dev, verified the bug exists on `main` HEAD as well):

- **`FunctionCallsStartedFrame` / `FunctionCallResultFrame`** duplicated in every active task's context aggregator → LLM context contains the same `assistant.tool_calls` + `tool` result twice → flow blocks at the next `set_node`. Reproduced deterministically (100% hit rate, 2/2 runs): caller `charlie` post-resume from sub-bot `carrozzerie`, user dictates a location, `data_nearby` completes successfully with `outcome=found`, but `set_node(present_result)` never fires; bot silent for 90 s until timeout. Log evidence (room id elided):

  ```
  12:04:39.420  LLMAssistantAggregator#13 FunctionCallResultFrame: [call_bot:E7hhTOCiz]  ← 1×
  12:04:39.420  LLMAssistantAggregator#13 FunctionCallResultFrame: [call_bot:E7hhTOCiz]  ← 2× (DUPLICATE, same ms)
  ```

  The duplicate has the **same `frame.id`** as the original — it's the same Python instance routed through the bus a second time, not a fresh emission.

- **`TTSSpeakFrame`** queued by a pre_action plays twice on the audio path (Azure TTS synthesises the same utterance twice in rapid succession on the SIP voice channel; on web chat we mute downstream synthesis with a sink so the text relay forwards the frame twice to the widget → same chat bubble appears twice).

- **`MetricsFrame`** counters inflated ~N× (N = number of active sibling tasks) → CDR token billing wrong. Observed on a 5-turn voice call: 30 metric samples, 3 distinct values (13× 644.5 + 13× 396.8 + 4× 888.1) — confirmed each frame seen 13× via the pipeline's 13 processors.

The pattern affects only frames that have no legitimate cross-task meaning. Pipecat already uses class-based exclusion (`_LIFECYCLE_FRAMES`, `_PASSTHROUGH_FRAMES`). This PR adds `_LOCAL_ONLY_FRAMES` for the six emitter-local frame types.

## Fix

Add a new module-level constant:

```python
_LOCAL_ONLY_FRAMES = (
    TTSSpeakFrame,
    FunctionCallsStartedFrame,
    FunctionCallInProgressFrame,
    FunctionCallResultFrame,
    FunctionCallCancelFrame,
    MetricsFrame,
)
```

…and short-circuit in both `BusBridgeProcessor.process_frame` and `_BusEdgeProcessor.process_frame`: when the frame is an instance of `_LOCAL_ONLY_FRAMES`, continue downstream locally and skip the `bus.send`. Same shape as the existing `_LIFECYCLE_FRAMES` exclusion two lines above.

## Test plan

- [x] **New tests** in `tests/test_bridge_processor.py`:
  - `test_local_only_frames_not_sent_to_bus` — instantiates one of each local-only frame type, asserts none reach the bus.
  - `test_local_only_frames_tuple_pin` — pins the exact set membership so any future add/remove of a local-only frame type is a deliberate code change.
- [x] Full regression: `pytest tests/test_bridge_processor.py tests/test_bus.py tests/test_bus_network.py tests/test_base_worker.py` → **80 passed, 0 regressions**.
- [x] `ruff check` + `ruff format --check` clean on both touched files.

## Compat

This is a behaviour change: a subscriber that relied on seeing one of these six frame types cross the bus will stop seeing them. We are not aware of any legitimate consumer — these frames are produced by the LLM/TTS/metrics services in the emitting task and consumed by the aggregator/sink/TTS service in the **same** task. If a use case for cross-task relay exists, a follow-up could expose an `include_local_frames=False` kwarg with opt-in restore — happy to add it in this PR if maintainers prefer.

## Related

This PR consolidates 3 separate downstream workarounds we've been carrying for the same root cause (bus broadcast not scoped to emitter):
- bounded-id dedup in our `CallMetricsObserver` for the `MetricsFrame` 13× duplication
- `set.remove` → `set.discard` monkey-patch on `FunctionCallUserMuteStrategy` for `FunctionCallsStartedFrame` double-broadcast
- bounded-id dedup sink + relay for `TTSSpeakFrame` double-broadcast

Happy to test any review/rebase against current `main`.

### DIFF
--- a/src/pipecat/bus/bridge_processor.py
+++ b/src/pipecat/bus/bridge_processor.py
@@ -25,9 +25,15 @@ from pipecat.frames.frames import (
     CancelFrame,
     EndFrame,
     Frame,
+    FunctionCallCancelFrame,
+    FunctionCallInProgressFrame,
+    FunctionCallResultFrame,
+    FunctionCallsStartedFrame,
+    MetricsFrame,
     OutputTransportMessageUrgentFrame,
     StartFrame,
     StopFrame,
+    TTSSpeakFrame,
 )
 from pipecat.processors.frame_processor import FrameDirection, FrameProcessor, FrameProcessorSetup
 
@@ -36,6 +42,21 @@ if TYPE_CHECKING:
 
 _LIFECYCLE_FRAMES = (StartFrame, EndFrame, CancelFrame, StopFrame)
 _PASSTHROUGH_FRAMES = (OutputTransportMessageUrgentFrame,)
+
+# Frames that are semantically local to the emitting worker's pipeline and
+# must never be relayed across the bus. Broadcasting them would cause the
+# same frame instance to be processed once per sibling task: duplicate
+# entries in each task's context aggregator for the function-call frames,
+# double TTS playback for ``TTSSpeakFrame``, inflated counters for
+# ``MetricsFrame``. None of these have legitimate cross-task semantics.
+_LOCAL_ONLY_FRAMES = (
+    TTSSpeakFrame,
+    FunctionCallsStartedFrame,
+    FunctionCallInProgressFrame,
+    FunctionCallResultFrame,
+    FunctionCallCancelFrame,
+    MetricsFrame,
+)
 
 
 class BusBridgeProcessor(FrameProcessor, BusSubscriber):
@@ -103,6 +124,13 @@ class BusBridgeProcessor(FrameProcessor, BusSubscriber):
         # Urgent transport frames pass through directly. They need to
         # reach the transport even when no child worker is active yet.
         if isinstance(frame, _PASSTHROUGH_FRAMES):
+            await self.push_frame(frame, direction)
+            return
+
+        # Emitter-local frames never cross the bus (see comment on
+        # ``_LOCAL_ONLY_FRAMES`` above). They continue downstream in this
+        # worker's pipeline only.
+        if isinstance(frame, _LOCAL_ONLY_FRAMES):
             await self.push_frame(frame, direction)
             return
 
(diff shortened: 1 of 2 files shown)


Maurotb:
**Heads up — DO NOT MERGE yet.** Tried to land this patch downstream in our worker and discovered the design is wrong.

Empirical finding from production: when I install the equivalent class-method monkey-patch on a real multi-worker setup (transport host `PipelineWorker` + N `LLMWorker` children via `BusBridgeProcessor`), all child workers register correctly but the entry agent's pre-action `TTSSpeakFrame` (greeting) never reaches the `TTSService`. Bot is silent from t=0. Reproduces deterministically.

Diagnosis: in the canonical multi-worker topology, the `TTSService` (and audio output sinks in general) live in the **main** `PipelineWorker` — not in the LLM child workers. When a child queues a `TTSSpeakFrame` via `flow_manager.task.queue_frame(...)`, that frame **must** cross the bus to reach the main's `TTSService`. Blocking it at `BusBridgeProcessor.process_frame` (the "send" side) breaks the legitimate child → main path.

The duplication bug is real but asymmetric: a child emits `TTSSpeakFrame` → bus broadcasts to **all** subscribers → main receives it (correct) **and** sibling children receive it too (the bug). Same shape for `FunctionCallResultFrame` etc.

A correct fix should be **receive-side, child-only**: drop local-only frames in `BusBridgeProcessor.on_bus_message` (the children's receiver), and leave `_BusEdgeProcessor.on_bus_message` (the main's receiver) untouched. Or, better: scope the broadcast at source with an explicit "main-only target" semantics that the bus already supports via `message.target`.

I'll rework this as a draft once I've worked out the canonical shape — closing for now to avoid a wrong merge.

Maurotb:
Closing per the analysis above — the patch's design is wrong for the multi-worker topology. Will reopen with a corrected receive-side fix once verified in production.

PR #4620, opened 2026-06-02
azure/stt: expose profanity kwarg + mark RecognizedSpeech transcripts finalized

Maurotb:
## Summary

Two small, independent fixes on `AzureSTTService` that downstream consumers currently work around by reaching into `_speech_config` and patching `_on_handle_recognized`.

### 1. Expose `profanity` as a public init kwarg

`SpeechConfig.set_profanity(ProfanityOption.{Raw,Masked,Removed})` is the canonical Azure switch for the profanity filter, but it isn't reachable from `AzureSTTService.__init__`. The default `Masked` mode is over-eager on non-English transcripts: Italian deployments routinely see common names (e.g. "Toro", surnames containing substrings the filter dislikes) come back as `****`, which breaks downstream fuzzy matching and LLM reasoning on the censored token.

This PR adds:

```python
AzureSTTService(api_key=..., region="eastus", profanity="raw")
```

`Literal["raw", "masked", "removed"] | None`, default `None` = keep the Azure SDK default (Masked). Backward compatible.

### 2. Mark `RecognizedSpeech` transcripts as `finalized=True`

`AzureSTTService._on_handle_recognized` constructs the `TranscriptionFrame` for an utterance's final recognition but doesn't set `finalized=True`. Azure's `RecognizedSpeech` event is by definition the final recognition for an utterance — leaving the flag unset is a silent misclassification.

User-visible impact: `SpeechTimeoutUserTurnStopStrategy` (and any strategy that reads `finalized`) takes its finalized fast-path on Deepgram (where pipecat already sets the flag) but falls through to a VAD-only fallback on Azure. On short or quiet replies ("Sì.", "Ok.") where the local Silero VAD doesn't fire, the user turn never stops → LLM never sees the transcript → bot looks deaf and idle nudges fire ~7s later.

One-line fix.

## Test plan

- [x] New `tests/test_azure_stt.py`:
  - 7 tests for the `profanity` kwarg (default no-op, all 3 values accepted, invalid value rejected fast, `set_profanity` actually called when set, not called when omitted).
  - 1 test pinning the `finalized=True` invariant on the constructed `TranscriptionFrame` (spy on `__init__`).
- [x] `ruff check` + `ruff format --check` clean on touched files.
- [x] `pytest tests/test_azure_stt.py` → 8 passed.

## Compat

- `profanity=None` default = previous behaviour unchanged.
- `finalized=True` is additive on the frame — any downstream consumer that branched on `if frame.finalized` previously took the non-final path on Azure; on the recognized handler that branch was always wrong (the event IS the final recognition). Aligns Azure with Deepgram's existing behaviour for this path.

## Out of scope

There's a separate, related issue in `UserTurnController._trigger_user_turn_stop` (silently swallows stop when `_user_turn=False`, i.e. when local VAD missed a short utterance). Fixing that requires design discussion — happy to open a separate issue.


src/pipecat/services/azure/stt.py
```diff
@@ -79,6 +80,7 @@ def __init__(
         sample_rate: int | None = None,
         private_endpoint: str | None = None,
         endpoint_id: str | None = None,
+        profanity: Literal["raw", "masked", "removed"] | None = None,
```
markbackman: __init__ args are for settings that will never change in a session. `profanity` feels like a setting that should be runtime settable, as well. I'd recommend moving it to `AzureSTTSettings`.

markbackman (commented):
Finalize fix looks good. Azure is atypical in that it doesn't finalize, but as you point out, the API mechanism is such that it will always deliver the transcript in the final event. This seems like a safe change.

markbackman (approved):
Ok, pushed a commit to move profanity to AzureSTTSettings and add a changelog.

Looks good to merge! Thanks for the contribution 🙇

PR #4622, opened 2026-06-02
Passing the WhatsApp call metadata to the runner arguments.

filipi87:
## Summary
  
  - Registered the `RequestValidationError` handler globally in `_configure_server_app()` so 422 validation errors from any transport (WhatsApp, WebRTC, telephony, etc.) log both the error details and the raw request body — previously this handler only existed for WhatsApp routes
  - Passed the `WhatsAppConnectCall` metadata (caller phone number, call ID, direction, timestamp) as `runner_args.body` in the WhatsApp `connection_callback`, giving bot code typed access to call context via `isinstance(runner_args.body, WhatsAppConnectCall)`
  - Updated `WhatsAppClient.handle_webhook_request` to support both the new two-argument callback signature `(connection, call)` and the legacy single-argument `(connection,)` — the old form still works but emits a `DeprecationWarning`

### DIFF
--- a/src/pipecat/runner/run.py
+++ b/src/pipecat/runner/run.py
@@ -118,6 +118,7 @@ try:
     import uvicorn
     from dotenv import load_dotenv
     from fastapi import BackgroundTasks, FastAPI, Header, HTTPException, Request, WebSocket
+    from fastapi.encoders import jsonable_encoder
     from fastapi.exceptions import RequestValidationError
     from fastapi.middleware.cors import CORSMiddleware
     from fastapi.responses import HTMLResponse, JSONResponse, RedirectResponse
@@ -384,8 +385,11 @@ def _configure_server_app(args: argparse.Namespace):
     async def validation_exception_handler(request: Request, exc: RequestValidationError):
         body = await request.body()
         logger.error(f"422 Validation error on {request.url.path}: {exc.errors()}")
-        logger.error(f"Raw body: {body.decode()}")
-        return JSONResponse(status_code=422, content={"detail": exc.errors()})
+        logger.error(
+            "Raw body: %s",
+            body.decode(errors="replace")[:5000],
+        )
+        return JSONResponse(status_code=422, content=jsonable_encoder({"detail": exc.errors()}))
 
     # Shared session store: session_id -> body data. Used by the WebRTC /start
     # flow and the /sessions/{session_id}/... proxy routes.
--- a/src/pipecat/runner/run.py
+++ b/src/pipecat/runner/run.py
@@ -118,8 +118,9 @@ try:
     import uvicorn
     from dotenv import load_dotenv
     from fastapi import BackgroundTasks, FastAPI, Header, HTTPException, Request, WebSocket
+    from fastapi.exceptions import RequestValidationError
     from fastapi.middleware.cors import CORSMiddleware
-    from fastapi.responses import HTMLResponse, RedirectResponse
+    from fastapi.responses import HTMLResponse, JSONResponse, RedirectResponse
 except ImportError as e:
     logger.error(f"Runner dependencies not available: {e}")
     logger.error("To use Pipecat runners, install with: pip install pipecat-ai[runner]")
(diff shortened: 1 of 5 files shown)


markbackman (commented):
## Code review

Reviewed for bugs and CLAUDE.md/AGENTS.md compliance. Two issues found (one bug, one deprecation-convention).

src/pipecat/runner/run.py
```diff
@@ -375,6 +376,17 @@ def _configure_server_app(args: argparse.Namespace):
...
+    # FastAPI returns 422 Unprocessable Entity for Pydantic validation failures by default, but
+    # swallows the raw request body in the error response. This handler overrides that behavior to
+    # log both the validation errors and the raw body, making it much easier to debug malformed
+    # payloads from any transport (WhatsApp, WebRTC, telephony, etc.).
+    @app.exception_handler(RequestValidationError)
+    async def validation_exception_handler(request: Request, exc: RequestValidationError):
+        body = await request.body()
+        logger.error(f"422 Validation error on {request.url.path}: {exc.errors()}")
+        logger.error(f"Raw body: {body.decode()}")
+        return JSONResponse(status_code=422, content={"detail": exc.errors()})
```
markbackman: **Bug:** `exc.errors()` can contain non-JSON-serializable objects, so this handler can raise inside itself and surface as a 500 — the opposite of its intended purpose.

In Pydantic v2, when a request model uses a custom `@field_validator`/`@model_validator` that raises `ValueError`, the error entry's `ctx["error"]` holds the raw exception instance. `JSONResponse` serializes with plain `json.dumps` (no `jsonable_encoder`), so this raises `TypeError: Object of type ValueError is not JSON serializable`. Because it's thrown inside the exception handler, Starlette converts it into an unhandled 500. This is a regression vs. FastAPI's default validation handler, which wraps errors in `jsonable_encoder(...)`.

Wrap the content with `jsonable_encoder` (and add `from fastapi.encoders import jsonable_encoder` to the imports):

```python
return JSONResponse(status_code=422, content=jsonable_encoder({"detail": exc.errors()}))
```

https://github.com/pipecat-ai/pipecat/blob/7ed16cdd53c571757b5ff6cc1fa153ab9c22253b/src/pipecat/runner/run.py#L385-L388

src/pipecat/transports/whatsapp/client.py
```diff
@@ -197,9 +199,12 @@ async def handle_webhook_request(
...
             request: The webhook request from WhatsApp containing call events
-            connection_callback: Optional callback function to invoke when a new
-                               WebRTC connection is established. The callback
-                               receives the SmallWebRTCConnection instance.
+            connection_callback: Optional callback invoked when a new WebRTC
+                               connection is established. Accepts two signatures:
+                               ``(connection, call)`` — preferred, receives the
+                               SmallWebRTCConnection and the WhatsAppConnectCall
+                               with caller metadata; or the legacy
+                               ``(connection,)`` — still supported but deprecated.
```
markbackman: **AGENTS.md adherence:** This deprecates the single-argument `connection_callback(connection)` signature, but the docstring describes the deprecation in prose instead of using the required `.. deprecated::` Sphinx directive:

> Use the `.. deprecated:: <version>` Sphinx directive in docstrings (never inline tags like `[DEPRECATED]`), and pair it with a runtime `warnings.warn(..., DeprecationWarning)` at the call site.

— [AGENTS.md L108](https://github.com/pipecat-ai/pipecat/blob/7ed16cdd53c571757b5ff6cc1fa153ab9c22253b/AGENTS.md#L107-L109)

The runtime `warnings.warn(...)` half is already in place; add the directive to the docstring, e.g.:

```
    .. deprecated:: 1.4.0
        The single-argument ``(connection,)`` callback signature is deprecated.
        Update it to accept ``(connection, call: WhatsAppConnectCall)``.
```

(substitute the actual upcoming release version)

https://github.com/pipecat-ai/pipecat/blob/7ed16cdd53c571757b5ff6cc1fa153ab9c22253b/src/pipecat/transports/whatsapp/client.py#L202-L207

src/pipecat/runner/run.py
filipi87: Done.

src/pipecat/transports/whatsapp/client.py
filipi87: Done.

src/pipecat/transports/whatsapp/client.py
```diff
@@ -197,9 +199,17 @@ async def handle_webhook_request(
...
-                               WebRTC connection is established. The callback
-                               receives the SmallWebRTCConnection instance.
+            connection_callback: Optional callback invoked when a new WebRTC
+                               connection is established. Accepts two signatures:
+                               ``(connection, call)`` — preferred, receives the
+                               SmallWebRTCConnection and the WhatsAppConnectCall
+                               with caller metadata; or the legacy
+                               ``(connection,)`` signature.
+
+                               .. deprecated:: 1.3.0
```
markbackman: Next release will be 1.4.0:

```suggestion
                               .. deprecated:: 1.4.0
```

src/pipecat/transports/whatsapp/client.py
```diff
@@ -197,9 +199,17 @@ async def handle_webhook_request(
...
+                                   The single-argument ``(connection,)`` signature is
+                                   deprecated. Use ``(connection, call: WhatsAppConnectCall)``
+                                   instead.
```
markbackman, line 212: I think you need a newline after a deprecation block, too.

markbackman (approved):
LGTM! Just some small clean up.

src/pipecat/transports/whatsapp/client.py
filipi87: True. Nice catch!

src/pipecat/transports/whatsapp/client.py
filipi87: Done.

PR #4624, opened 2026-06-02
bus: stop sibling workers from receiving emitter-local frames

Maurotb:
## Summary

In a multi-worker setup with bridged `PipelineWorker` instances (one main `PipelineWorker` + N `LLMWorker` children via `BusBridgeProcessor` / `_BusEdgeProcessor`), the bus broadcasts every frame to every subscriber with no per-frame scoping. Some frame types are semantically local to the emitting worker and have no meaning for siblings — but they're delivered there anyway, causing real bugs:

- **`TTSSpeakFrame`**: a sub-bot queues TTS via `flow_manager.task.queue_frame(TTSSpeakFrame(...))`. The main `PipelineWorker` (which hosts the `TTSService`) is the legit destination, but a sibling `LLMWorker` also receives the frame, and its own TTS path synthesises the audio a second time (or, on a chat-only DataChannel sink, emits the same `bot_text` twice).

- **`FunctionCallsStartedFrame` / `FunctionCallInProgressFrame` / `FunctionCallResultFrame` / `FunctionCallCancelFrame`**: the emitting worker's `LLMAssistantAggregator` records the tool round-trip. A sibling that picks up the same frames via the bus appends duplicate `assistant.tool_calls` + `tool` result pairs to its own LLM context. The very next `set_node` can block on the inconsistent context — reproduced deterministically in production as a sub-bot return where the caller's flow stalls 90 s after a successful tool result.

- **`MetricsFrame`**: the emitting worker's metrics observer is the legit consumer. A sibling that double-counts inflates token / duration / chars counters that downstream consumers (CDR / billing dashboards) use.

## Fix

Add a receive-side filter on `_BusEdgeProcessor.on_bus_message` that drops these 6 frame types. The send side is unchanged so the main worker — which uses `BusBridgeProcessor` (not the edge variant) — still receives them; that path is intentional (a sub-bot's `TTSSpeakFrame` *must* still reach the main's `TTSService`).

```python
_LOCAL_ONLY_FRAMES = (
    TTSSpeakFrame,
    FunctionCallsStartedFrame,
    FunctionCallInProgressFrame,
    FunctionCallResultFrame,
    FunctionCallCancelFrame,
    MetricsFrame,
)
```

## Empirical verification

Verified against a 3-worker production setup (transport host + N child `LLMWorker`s):

**Before (pristine pipecat 1.3.0):**

```
12:04:39.420  LLMAssistantAggregator#13 FunctionCallResultFrame: [call_bot:E7hhTOCiz]   ← 1×
12:04:39.420  LLMAssistantAggregator#13 FunctionCallResultFrame: [call_bot:E7hhTOCiz]   ← 2× (DUPLICATE, same frame.id)
```

The duplicate carries the **same `frame.id`** as the original — it's the same Python instance routed through the bus a second time.

**After (this PR applied):**

```
15:11:11.477  LLMAssistantAggregator#1  FunctionCallResultFrame: [data_nearby:biz8r6pqx]   ← 1× only
```

Zero duplicates in the assistant aggregator across the full scenario.

## Test plan

- [x] `tests/test_bridge_processor.py` — 4 new tests:
  - `test_local_only_frames_from_bus_dropped_at_edge`: `_BusEdgeProcessor` drops all 6 frame types from a sibling
  - `test_non_local_only_frames_from_bus_pass_at_edge`: sanity, `TextFrame` still injected
  - `test_local_only_frames_from_bus_pass_at_bridge`: sanity, `BusBridgeProcessor` still passes these frames (legit child → main route preserved)
  - `test_local_only_frames_tuple_pin`: drift detection on `_LOCAL_ONLY_FRAMES` membership
- [x] Full regression: `pytest tests/test_bridge_processor.py tests/test_bus.py tests/test_bus_network.py tests/test_base_worker.py` → **82 passed, 0 regressions**.
- [x] `ruff check` + `ruff format --check` clean.
- [x] Live worker smoke (3-worker setup, sub-bot return after `end_call`): duplicates eliminated, no regression on baseline scenarios.

## Compat

This is a behaviour change: a sibling worker that relied on receiving one of these 6 frame types from another sibling will stop seeing them. We're not aware of any legitimate consumer — these frames are emitter-local and the bus broadcast was an unintended side-effect of the lack of per-frame scoping. If a use case exists, a follow-up could expose an `include_local_frames` knob on `_BusEdgeProcessor.__init__`.

## History

This is the receive-side child-only variant of a fix that was originally attempted in PR #4619 (closed) using a send-side filter on `BusBridgeProcessor.process_frame`. The send-side approach broke the legit child → main routing for `TTSSpeakFrame` (sub-bot's greeting never reached the main's TTSService → bot silent). Identified the topology asymmetry — `BusBridgeProcessor` lives in the main `PipelineWorker` (consumes frames from children), `_BusEdgeProcessor` lives in child `LLMWorker`s (sends to bus + receives sibling broadcasts) — and moved the filter to the edge.

Verified locally with the 5-step workflow (apply to `.venv` of a real worker, disable downstream dedup workarounds, run real smoke scenarios with the production bug repro, revert, then submit) before opening this PR.


src/pipecat/bus/bridge_processor.py
```diff
@@ -230,4 +252,14 @@ async def on_bus_message(self, message: BusMessage) -> None:
...
+        # Emitter-local frames from a sibling worker must never enter
+        # this worker's pipeline (see comment on ``_LOCAL_ONLY_FRAMES``
+        # above). Letting a sibling's ``FunctionCallResultFrame`` reach
+        # this worker's assistant aggregator duplicates the tool result
+        # in the LLM context; letting its ``TTSSpeakFrame`` reach this
+        # worker's TTS service synthesises the audio twice; letting its
+        # ``MetricsFrame`` reach this worker's metrics observer inflates
+        # the counters.
+        if isinstance(message.frame, _LOCAL_ONLY_FRAMES):
+            return
```
aconchillo, line 264: I believe this should go in the sender. If the frames are never used in the receiving side, then there's no need to even put those frames in the bus.

aconchillo: That is, the `exclude_frames` is used in `process_frame` which is for frames that go out. The receiving frames will still be received by `BusBridgeProcessor` in `on_bus_message`

Maurotb: Thanks for taking a look. I actually tried the send-side approach in the original PR (#4619, closed) and it broke our setup — TTS goes silent because we host the TTSService in the main worker, and child LLMWorkers emit TTSSpeak frames that need to cross the bus to get there.

Our topology is the one from examples/multi-worker/local-handoff/local-handoff-two-agents.py: main pipeline is `transport.input → stt → user_agg → BusBridge → tts → transport.output`, child workers are LLM-only. When a pipecat-flows pre_action calls `task.queue_frame(TTSSpeakFrame(...))` on the child side, the frame has to travel via the bus to reach the main's TTSService. Excluding it at the source kills the synthesis.

Same idea for MetricsFrame if the observer sits in the main worker.

The tts-in-each-child variant (local-handoff-two-agents-tts.py) wouldn't have this problem — there the frames stay local — so exclude_frames at the source works.

I guess it comes down to which topology you want to support. If only the tts-in-each-child pattern is expected to scale to multi-worker, the filter here doesn't really do anything for that case and I can close. But if the TTS-in-main topology is also meant to work, then dropping at the source breaks the legit child→main path.

Let me know how you'd prefer to handle it. Thanks.

aconchillo: I see, the issue is when there is more than 1 children. In theory, in a handoff scenario the other bridged children should not be active and if they are not active all frames should be discarded. Is it possible that you have more than 1 active bridged children? If so, that's probably the issue. You can have as many children as you want that receive jobs etc. the issue is when they are in the same bridge. You could have multiple bridge children in separate bridges, but if more than 1 is active in the same bridge, that's the issue.

Maurotb: Got curious and added some instrumentation to track frame ids end-to-end. Turns out the duplicate isn't two different sibling frames from `broadcast_frame` reaching the aggregator — it's the same Python instance (same `frame.id`) getting processed twice. So `broadcast_frame` is fine. The issue is an echo loop in `_BusEdgeProcessor` triggered by a brief window where 2 bridged workers are both active.

Trace for a tool handler that triggers a worker handoff (worker_a calls a tool that activates worker_b, then returns the result dict so flows fires `result_callback`):

```
worker_a  EDGE_TX_BUS (down)    frame=585     ← initial bus send
main      BRIDGE_RX_BUS         frame=585
worker_a  EDGE_TX_BUS (up)      frame=586     ← upstream sibling, separate id
main      BRIDGE_RX_BUS         frame=586

worker_b  EDGE_TX_BUS (down)    frame=585     ← ECHO: worker_b re-broadcasts frame 585
worker_b  EDGE_TX_BUS (up)      frame=586     ← ECHO: and 586
main      BRIDGE_RX_BUS         frame=585     ← main receives 585 again
main      BRIDGE_RX_BUS         frame=586     ← and 586 again

AGG_RX    frame=585  (1st)
AGG_RX    frame=585  (2nd, same Python instance)
```

For a plain tool that runs entirely inside one worker, the same `broadcast_frame` happens (down+up siblings) but only one EDGE_TX_BUS sequence, no echo, the aggregator sees the result exactly once.

The mechanism: when worker_b becomes active during the handoff, the in-flight `FunctionCallResultFrame` from worker_a still gets delivered to it via the bus. `BusBridgeProcessor.on_bus_message` injects the frame into worker_b's pipeline via `push_frame`, and the frame travels through to worker_b's `_BusEdgeProcessor`, which calls `bus.send` again — same instance, same id, re-broadcast. Main's bridge then receives the same frame twice.

So the "more than 1 active children in the same bridge" condition you mentioned is real, but it's not a misconfiguration — it's a transient race during the activate/deactivate transition itself. `activate_worker(deactivate_self=True)` doesn't make the two transitions atomic with respect to in-flight frames.

A fix on the edge side seems natural: mark frames received via `on_bus_message` with a "from_bus" flag (or track ids in a bounded set) so `_BusEdgeProcessor.process_frame` knows not to re-broadcast them. Same idea you'd use to break any rebroadcast loop.

Given this, the receive-side filter in this PR is masking a symptom and isn't the right fix. Closing the PR. Happy to put up a new one with the echo-prevention if you'd like, otherwise feel free to handle it however fits best.

Maurotb:
Closing per the analysis in the inline review — this PR was filtering at the wrong layer. Root cause is an echo loop in `_BusEdgeProcessor` when 2 bridged workers are briefly both active during a handoff (the second one re-broadcasts the in-flight frame it received via the bus). Will follow up with a different patch focused on the actual loop, or defer to maintainers if there's a preferred shape.

aconchillo:
> Closing per the analysis in the inline review — this PR was filtering at the wrong layer. Root cause is an echo loop in `_BusEdgeProcessor` when 2 bridged workers are briefly both active during a handoff (the second one re-broadcasts the in-flight frame it received via the bus). Will follow up with a different patch focused on the actual loop, or defer to maintainers if there's a preferred shape.

Thank you @Maurotb ! I think the real issue is "when 2 bridged workers are briefly both active" (briefly might be the key). Let me know if this is easily reproducible or if it was an issue in your application (e.g. not deactivating at the right time). Definitely we'd love to fix this.

PR #4626, opened 2026-06-02
feat(mem0): support mem0ai 2.x

markbackman:
## Summary

Bumps the `mem0` extra to `mem0ai>=2,<3` and updates the `Mem0MemoryService` integration for the [mem0 2.0.0 breaking changes](https://github.com/mem0ai/mem0/releases/tag/v2.0.0).

### What changed in mem0 2.0.0 (relevant subset)

- Entity IDs (`user_id`/`agent_id`/`run_id`) must be passed inside `filters=` on `search()`/`get_all()` — top-level kwargs now raise `ValueError`.
- `api_version`/`version` and `output_format` parameters were removed.
- `get_all()` returns a paginated envelope (`{count, next, previous, results}`).
- `add()` is now async server-side (returns `{status: "PENDING", event_id}`).

### Integration changes

- **Local `Memory` path** (`get_memories()`, `_retrieve_memories()`): pass entity IDs via `filters={...}` (flat top-level keys) instead of top-level kwargs; use `top_k` instead of the removed `limit`.
- **Cloud `MemoryClient` path**: drop the removed `version`/`output_format` args from `search()` and `add()`.
- **`InputParams.api_version`**: deprecated (no longer used); emits a `DeprecationWarning` when set.
- **`examples/rag/rag-mem0.py`**: remove `api_version="v2"`.

## Breaking Changes

- The `mem0` extra now requires `mem0ai>=2,<3` (was `>=1.0.8,<2`). Users on mem0 1.x must upgrade.
- `Mem0MemoryService.InputParams.api_version` is deprecated and no longer used; setting it emits a `DeprecationWarning`.
- Behavior inherited from mem0 2.0.0: `rerank` defaults to `False` (was `True`), and `add()` is async server-side, so newly stored memories are only queryable once processed.

## Notes

- `rerank` is left at the new mem0 default (`False`) to favor latency in voice contexts.
- The `mem0ai[nlp]` hybrid-search extra only ships wheels for Python 3.10–3.12; the base package (and the cloud path used by the example) is unaffected on newer Pythons.

Resolves https://github.com/pipecat-ai/pipecat/security/dependabot/194


filipi87 (approved):
Yeah, I think we can't avoid a breaking change in those cases.
LGTM!

markbackman:
Unfortunately, true. mem0ai has a vulnerability that requires a bump to >=2, so we have to break. Not much we can do. The good new is that it's quite benign and essential no one will hit it.

PR #4629, opened 2026-06-02
Feat: Add MoQ transport

vipyne:

MoQ transport - now using proper Pipecat Prebuilt UI

Involves changes in the following repos:
- [voice-ui-kit](https://github.com/pipecat-ai/voice-ui-kit/pull/143)
- [pipecat-prebuilt](https://github.com/pipecat-ai/pipecat-prebuilt/pull/45)
- [pipecat-client-web-transport](https://github.com/pipecat-ai/pipecat-client-web-transports/pull/139)

To run/test locally (assumes those repos^ and pipecat are all sibling directories):
`~/pipecat $ ./scripts/moq-dev-setup.sh`
this will make local changes to link repos, then print instructions for running locally like so:
```
==========================================
 Done. Two terminals to run the stack:
==========================================

# 1. Bot (self-serves as the MoQ server)
cd ~/pipecat
uv run python examples/transports/transports-moq.py \
    -t moq --moq-serve --moq-tls-generate localhost

# 2. Prebuilt client dev server (Vite, usually port 5173)
cd ~/pipecat-prebuilt/client
npm run dev

Open the Vite dev URL (http://localhost:5173/ NOT http://localhost:7860/client/, 
which serves the published pipecat-ai-prebuilt wheel without MoQ). Pick "MoQ" in the
transport dropdown.
```
Be sure to use `http://localhost:5173/` and to choose `Media over QUIC` in the top right dropdown menu.
Checkout `MOQ_DEV.md` for more details.

NOTE: `MOQ_DEV.md` and `scripts/moq-dev-setup.sh` not intended to be merged- will remove them once this is approved.


.coverage
markbackman, line 1: Was this an intentional addition?

MOQ_DEV.md
```diff
@@ -0,0 +1,255 @@
+# MoQ local dev setup
```
markbackman: Is there a better location for this--perhaps docs?

pyproject.toml
```diff
@@ -102,6 +102,7 @@ mistral = ["mistralai>=2.0.0,<3"]
 mlx-whisper = [ "mlx-whisper~=0.4.2" ]
 moondream = [ "accelerate~=1.10.0", "einops~=0.8.0", "pyvips[binary]~=3.0.0", "timm~=1.0.13", "transformers>=4.48.0,<6" ]
 moonshine = [ "moonshine-voice>=0.0.62" ]
+moq = [ "moq-rs>=0.2.13,<0.3", "cryptography>=43.0.0" ]
```
markbackman: 0.3 is the current latest. Is there a reason to not make this `<1`? Not sure how closely this package follows semantic versioning.

markbackman: Also `cryptography` of 43 is very old, July 20, 2024. Is there a reason for the minimum to be that old?

examples/transports/transports-moq.py
```diff
@@ -0,0 +1,186 @@
...
+from pipecat.services.openai.llm import OpenAILLMService
+from pipecat.transports.base_transport import BaseTransport, TransportParams
+from pipecat.transports.daily.transport import DailyParams
+from pipecat.transports.moq import MOQParams
+
+load_dotenv(override=True)
+
+# Transport-specific parameters using lambdas for deferred creation
+transport_params = {
+    "daily": lambda: DailyParams(
```
markbackman: Shall we make this run the moq transport only?

examples/transports/transports-moq.py
```diff
@@ -0,0 +1,186 @@
...
+        audio_out_enabled=True,
+    ),
+}
+
+
+async def run_bot(transport: BaseTransport, runner_args: RunnerArguments):
+    """Run the bot with the given transport."""
+    logger.info("Starting bot")
+
+    stt = DeepgramSTTService(api_key=os.getenv("DEEPGRAM_API_KEY"))
```
markbackman: Nit: align with pattern in examples to satisfy pyright:

```suggestion
    stt = DeepgramSTTService(api_key=os.environ["DEEPGRAM_API_KEY"])
```

examples/transports/transports-moq.py
```diff
@@ -0,0 +1,186 @@
...
+
+    tts = CartesiaTTSService(
+        api_key=os.getenv("CARTESIA_API_KEY"),
```
markbackman: Same, re: pyright:
```suggestion
        api_key=os.environ["CARTESIA_API_KEY"],
```

examples/transports/transports-moq.py
```diff
@@ -0,0 +1,186 @@
...
+        voice_id="71a7ad14-091c-4e8e-a314-022ece01c121",  # British Reading Lady
+    )
+
+    llm = OpenAILLMService(api_key=os.getenv("OPENAI_API_KEY"), model="gpt-4o")
```
markbackman: Pattern is:
```
        llm = OpenAILLMService(
            api_key=os.environ["OPENAI_API_KEY"],
            settings=OpenAILLMService.Settings(
                system_instruction="You are a helpful assistant in a voice conversation. Your responses will be spoken aloud, so avoid emojis, bullet points, or other formatting that can't be spoken. Respond to what the user said in a creative, helpful, and brief way.",
            ),
        )
```

examples/transports/transports-moq.py
```diff
@@ -0,0 +1,186 @@
...
+
+    messages = [
```
markbackman: Remove `messages`. This is now a `system_instruction` in the LLM.

examples/transports/transports-moq.py
```diff
@@ -0,0 +1,186 @@
...
+            "role": "system",
+            "content": "You are a helpful assistant in a real-time voice call. "
+            "Your goal is to demonstrate your capabilities in a succinct way. "
+            "Your output will be spoken aloud, so avoid special characters that can't easily be "
+            "spoken, such as emojis or bullet points. Respond to what the user said in a creative "
+            "and helpful way.",
+        },
+    ]
+
+    context = LLMContext(messages)
```
markbackman: No need to provide since the `system_instruction` lives on the LLM now.

```suggestion
    context = LLMContext()
```

examples/transports/transports-moq.py
```diff
@@ -0,0 +1,186 @@
...
+            stt,
+            user_aggregator,  # User responses
+            llm,  # LLM
+            tts,  # TTS
+            transport.output(),  # Transport bot output
+            assistant_aggregator,  # Assistant spoken responses
+        ]
+    )
+
+    task = PipelineTask(
```
markbackman: You'll want to update for the PipelineWorker pattern. Check out the other examples for this.
```suggestion
    worker = PipelineWorker(
```

examples/transports/transports-moq.py
```diff
@@ -0,0 +1,186 @@
...
+
+    # For MOQ, we need to handle connection and events differently.
+    # create_transport() already wires the runner's ready_event +
+    # cert_fingerprints back-channel for us, so we don't need to do that here.
+    if isinstance(runner_args, MOQRunnerArguments):
+
+        @transport.event_handler("on_client_connected")
+        async def on_client_connected(transport):
+            logger.info("Client subscribed — starting conversation")
+            messages.append({"role": "system", "content": "Please introduce yourself to the user."})
```
markbackman: Latest pattern is:
```
context.add_message(
    {"role": "developer", "content": "Please introduce yourself to the user."}
)
```

examples/transports/transports-moq.py
```diff
@@ -0,0 +1,186 @@
...
+            logger.info("Disconnected from MOQ relay")
+            await task.cancel()
+
+        @transport.event_handler("on_error")
+        async def on_error(transport, message, exception):
+            logger.error(f"MOQ error: {message}")
+
+        # MOQInputTransport.start() auto-connects to the relay when the
+        # pipeline starts, so we don't dial transport.connect() here.
+        runner = PipelineRunner(handle_sigint=runner_args.handle_sigint)
```
markbackman: New pattern is:
```
        runner = WorkerRunner()

        await runner.add_workers(worker)
        await runner.run()
```

src/pipecat/runner/run.py
```diff
@@ -56,8 +56,9 @@ async def bot(runner_args: RunnerArguments):
 Supported transports:
 
 - Daily - Creates rooms and tokens, runs bot as participant
-- WebRTC - Provides local WebRTC interface with prebuilt UI
+- MOQ - Media over QUIC, connects to a MOQ relay for pub/sub streaming
```
markbackman, line 59: Thanks for the clean up! 🧹

src/pipecat/runner/run.py
```diff
@@ -1181,6 +1248,83 @@ async def handle_dialin_webhook(request: Request):
             }
 
 
+def _cert_hash_from_pem(path: str) -> str | None:
```
markbackman: Can we move to runner/moq.py, where we'll keep helper functions?

src/pipecat/runner/run.py
```diff
@@ -1181,6 +1248,83 @@ async def handle_dialin_webhook(request: Request):
...
+            cert = x509.load_pem_x509_certificate(f.read())
+        der_bytes = cert.public_bytes(serialization.Encoding.DER)
+        digest = hashlib.sha256(der_bytes).digest()
+        return base64.b64encode(digest).decode()
+    except Exception as e:
+        logger.warning(f"Could not compute cert fingerprint from {path}: {e}")
+        return None
+
+
+def _hex_to_b64(hex_str: str) -> str | None:
```
markbackman: Can we move to runner/moq.py, where we'll keep helper functions?

src/pipecat/runner/run.py
```diff
@@ -1181,6 +1248,83 @@ async def handle_dialin_webhook(request: Request):
...
+    try:
+        import base64
+
+        return base64.b64encode(bytes.fromhex(hex_str)).decode()
+    except Exception as e:
+        logger.warning(f"Could not convert cert hash {hex_str!r}: {e}")
+        return None
+
+
+def _build_moq_client_config(
```
markbackman: Can we move to runner/moq.py, where we'll keep helper functions?

src/pipecat/runner/run.py
```diff
@@ -1475,6 +1619,90 @@ def main(parser: argparse.ArgumentParser | None = None):
         ),
     )
 
+    # MOQ-specific arguments
```
markbackman: I know nothing about moq, but are all of these always needed? This is a lot!

src/pipecat/transports/moq/__init__.py
```diff
@@ -0,0 +1,30 @@
+#
```
markbackman: We've avoided re-vending types in __init__.py, so perhaps we just leave this blank.

markbackman (commented):
Sweet!

Bunch of comments, many nitpicks.

I don't know anything about moq, so someone else other than me should review the transport changes. Perhaps we can reach out to moq experts if we want their review? Not required.

src/pipecat/transports/moq/transport.py
```diff
@@ -0,0 +1,825 @@
...
+                    f"audio: {self._params.audio_out_track!r})"
+                )
+                await self._call_event_handler("on_connected")
+
+                # In serve mode, drive the accept loop in the background.
+                # serve() holds each session task until the session closes,
+                # so memory doesn't grow with past connections.
+                serve_task: asyncio.Task | None = None
+                if self._params.serve:
+                    serve_task = asyncio.create_task(cast(moq.Server, transport).serve())
```
markbackman: This uses raw `asyncio.create_task()`, but [AGENTS.md](https://github.com/pipecat-ai/pipecat/blob/e26b3ea596a4f451746d7e10aff94ec9bc93bfdd/AGENTS.md#L103) says: "Always use `self.create_task(coroutine, name)` instead of raw `asyncio.create_task()`. The `TaskManager` automatically tracks tasks and cleans them up on processor shutdown."

Swapping just this line isn't enough, though: `MOQTransport` never receives a task manager (neither `MOQInputTransport.setup()` nor `MOQOutputTransport.setup()` forwards it), so `self.create_task()` here would raise `TaskManager is not initialized`. The established pattern is what the Daily transport does — forward setup from the input/output processors to the shared object ([daily/transport.py#L1836-L1844](https://github.com/pipecat-ai/pipecat/blob/e26b3ea596a4f451746d7e10aff94ec9bc93bfdd/src/pipecat/transports/daily/transport.py#L1836-L1844)) — and then use `self.create_task(...)` here and `await self.cancel_task(serve_task)` in the `finally` block below ([context](https://github.com/pipecat-ai/pipecat/blob/e26b3ea596a4f451746d7e10aff94ec9bc93bfdd/src/pipecat/transports/moq/transport.py#L605-L622)).

Reason: CLAUDE.md adherence (via AGENTS.md include).

src/pipecat/transports/moq/transport.py
```diff
@@ -0,0 +1,825 @@
...
+class MOQInputTransport(BaseInputTransport):
+    """MOQ input transport: subscribes to the peer's audio track."""
+
+    def __init__(
+        self,
+        transport: "MOQTransport",
+        params: MOQParams,
+        **kwargs,
+    ):
+        """Initialize the MOQ input transport."""
```
markbackman: Per [AGENTS.md#L128](https://github.com/pipecat-ai/pipecat/blob/e26b3ea596a4f451746d7e10aff94ec9bc93bfdd/AGENTS.md#L128) — "Docstrings: Google-style. Classes describe purpose; `__init__` has `Args:` section" — this `__init__` is missing its `Args:` section (compare `MOQTransport.__init__` in this same file, or [`DailyInputTransport.__init__`](https://github.com/pipecat-ai/pipecat/blob/e26b3ea596a4f451746d7e10aff94ec9bc93bfdd/src/pipecat/transports/daily/transport.py#L1779-L1793)).

```suggestion
        """Initialize the MOQ input transport.

        Args:
            transport: The parent MOQTransport managing the MoQ session.
            params: MOQ transport configuration parameters.
            **kwargs: Additional arguments passed to the parent class.
        """
```

Reason: CLAUDE.md adherence (via AGENTS.md include).

markbackman (commented):
## Code review

Checked for bugs and CLAUDE.md compliance. No bugs found; flagged 3 [AGENTS.md](https://github.com/pipecat-ai/pipecat/blob/e26b3ea596a4f451746d7e10aff94ec9bc93bfdd/AGENTS.md) convention issues inline (1 task-management, 2 docstring).

src/pipecat/transports/moq/transport.py
```diff
@@ -0,0 +1,825 @@
...
+class MOQOutputTransport(BaseOutputTransport):
+    """MOQ output transport: writes audio + transcript frames to MOQ tracks."""
+
+    def __init__(
+        self,
+        transport: "MOQTransport",
+        params: MOQParams,
+        **kwargs,
+    ):
+        """Initialize the MOQ output transport."""
```
markbackman: Same as `MOQInputTransport.__init__` above — missing `Args:` section per [AGENTS.md#L128](https://github.com/pipecat-ai/pipecat/blob/e26b3ea596a4f451746d7e10aff94ec9bc93bfdd/AGENTS.md#L128).

```suggestion
        """Initialize the MOQ output transport.

        Args:
            transport: The parent MOQTransport managing the MoQ session.
            params: MOQ transport configuration parameters.
            **kwargs: Additional arguments passed to the parent class.
        """
```

Reason: CLAUDE.md adherence (via AGENTS.md include).

.coverage
vipyne: will remove

MOQ_DEV.md
vipyne: ah, this was intended only for review purposes- not to merge. I should have explicitly said that.

src/pipecat/runner/run.py
vipyne: thoughts @kixelated ?

kixelated: I have separate arguments depending on if we're running as a [client](https://github.com/moq-dev/moq/blob/dad2e8d1afb2db41ba4c224d07c7ebf9c3d8ce33/rs/moq-native/src/client.rs#L11) or [server](https://github.com/moq-dev/moq/blob/dad2e8d1afb2db41ba4c224d07c7ebf9c3d8ce33/rs/moq-native/src/server.rs#L21), but there's a lot of overlap. And a LOT of optional stuff.

src/pipecat/runner/run.py
```diff
@@ -1475,6 +1619,90 @@ def main(parser: argparse.ArgumentParser | None = None):
...
+    parser.add_argument(
+        "--moq-host",
```
kixelated: I would get rid of `--moq-host` and `--moq-port` because they're ambiguous. For example, clients need a host/port to connect to, and a host/port to bind to.

src/pipecat/runner/run.py
```diff
@@ -1475,6 +1619,90 @@ def main(parser: argparse.ArgumentParser | None = None):
...
+        help="MOQ relay host address (default: localhost)",
+    )
+    parser.add_argument(
+        "--moq-port",
+        type=int,
+        default=4080,
+        help="MOQ relay port (default: 4080)",
+    )
+    parser.add_argument(
+        "--moq-path",
```
kixelated: This doesn't make sense for a server, I would fold it into:

`--moq-connect https://host:port/path`

src/pipecat/runner/run.py
```diff
@@ -1475,6 +1619,90 @@ def main(parser: argparse.ArgumentParser | None = None):
...
+        type=str,
+        default="/moq",
+        help="MOQ endpoint path (default: /moq)",
+    )
+    parser.add_argument(
+        "--moq-namespace",
```
kixelated, line 1627: Note that MoQ automatically gossips namespaces (based on a prefix).

So you could actually discover any namespaces that match `pipecat` or `pipecat/**`. That way you could support multiple web clients with random IDs or something.

src/pipecat/runner/run.py
```diff
@@ -1475,6 +1619,90 @@ def main(parser: argparse.ArgumentParser | None = None):
...
+        type=str,
+        default="pipecat",
+        help="MOQ namespace/room (default: pipecat)",
+    )
+    parser.add_argument(
+        "--moq-bot-id",
```
kixelated, line 1633: You can make this part of the namespace. My libraries automatically split based on `/`

src/pipecat/runner/run.py
```diff
@@ -1475,6 +1619,90 @@ def main(parser: argparse.ArgumentParser | None = None):
...
+        type=str,
+        default="bot0",
+        help="This bot's participant id; broadcasts under <namespace>/<bot-id> (default: bot0)",
+    )
+    parser.add_argument(
+        "--moq-client-id",
```
kixelated, line 1639: IDK how this is different from `--moq-bot-id`

src/pipecat/runner/run.py
```diff
@@ -1475,6 +1619,90 @@ def main(parser: argparse.ArgumentParser | None = None):
...
+        type=str,
+        default="client0",
+        help="Peer client's participant id the bot subscribes to (default: client0)",
+    )
+    parser.add_argument(
+        "--moq-insecure",
```
kixelated: I would add `-tls` in front of TLS stuff.

```
--moq-tls-insecure <bool>
--moq-tls-cert <.pem>
--moq-tls-key <.pem>
--moq-tls-generate <hostname>
```

And clients can also present their own TLS certs, but by default they don't.

src/pipecat/runner/run.py
```diff
@@ -1475,6 +1619,90 @@ def main(parser: argparse.ArgumentParser | None = None):
...
+    )
+    parser.add_argument(
+        "--moq-serve",
+        action="store_true",
+        default=False,
+        help=(
+            "Run the bot as a MOQ server — the bot binds its own UDP socket "
+            "and accepts the browser's direct connection. Removes the need "
+            "for a separate moq-relay process. If --moq-cert/--moq-key are "
+            "not provided, a self-signed cert is generated on startup."
```
kixelated: I would make this explicit btw. So the server doesn't silently use a self-signed certificate only viable for local development.

src/pipecat/runner/run.py
```diff
@@ -1475,6 +1619,90 @@ def main(parser: argparse.ArgumentParser | None = None):
...
+        ),
+    )
+    parser.add_argument(
+        "--moq-key",
+        type=str,
+        default=None,
+        help="Path to the PEM-encoded private key matching --moq-cert (serve mode).",
+    )
+    parser.add_argument(
+        "--moq-serve",
```
kixelated: Fun fact, the same QUIC endpoint can be both a client and a server if you want. No other notes.

src/pipecat/runner/run.py
```diff
@@ -1475,6 +1619,90 @@ def main(parser: argparse.ArgumentParser | None = None):
...
+        default=False,
+        help=(
+            "Run the bot as a MOQ server — the bot binds its own UDP socket "
+            "and accepts the browser's direct connection. Removes the need "
+            "for a separate moq-relay process. If --moq-cert/--moq-key are "
+            "not provided, a self-signed cert is generated on startup."
+        ),
+    )
+    parser.add_argument(
+        "--moq-serve-bind",
```
kixelated: I would make `--moq-bind`. A client can also choose an address to bind to, but usually it uses an ephemeral port `[::]` while servers pick a poor `[::]:443`

pyproject.toml
kixelated: I'm trying to follow semver, yeah. There's automation for Rust to enforce it but not Python yet.

src/pipecat/runner/run.py
```diff
@@ -1181,6 +1248,83 @@ async def handle_dialin_webhook(request: Request):
...
+    In serve mode the bot just minted (or loaded) its own cert; we use
+    the fingerprint it reported (passed via ``cert_fingerprints``).
+    Otherwise we fall back to the PEM file at ``--moq-cert``.
+
+    Track names aren't pinned here — the bot publishes a catalog at
+    runtime and the browser reads whatever it advertises (codec, sample
+    rate, channel count, track name). Lets us add tracks (video,
+    screen-share) without a server-side config update.
+    """
+    cert_hash: str | None = None
```
kixelated: note this must be set ONLY if a self-signed certificate is used. It purposely doesn't work with valid TLS certficates.

src/pipecat/runner/run.py
kixelated: Oh I see, this is for the bot replies.

Yeah if you want to support discovering namespaces, this would be something like `{namespace}/bot` or `bot/{namespace}`. See: https://github.com/moq-dev/moq/blob/28505fc46640c8f38d18b518bd03e3590c92975a/js/moq-boy/src/index.ts#L20

Otherwise it's fine to hard code a single one.

kixelated:
> Sweet!
> 
> Bunch of comments, many nitpicks.
> 
> I don't know anything about moq, so someone else other than me should review the transport changes. Perhaps we can reach out to moq experts if we want their review? Not required.

yee these are my libraries, and I made MoQ

src/pipecat/transports/moq/transport.py
```diff
@@ -0,0 +1,918 @@
...
+MOQ session, announcement discovery, subscription routing, group/frame
+framing, codec-specific catalog management, and Opus encode/decode +
+resampling for raw audio tracks. This module just wires it into the
+pipecat Frame pipeline.
+
+Each participant publishes under a per-participant broadcast path
+``<namespace>/<participant_id>`` (e.g. ``pipecat/bot0``); the bot
+subscribes to the peer at ``<namespace>/<peer_id>``. Audio rides on a
+single Opus track; RTVI JSON rides on a fixed-name ``transcript.json``
+raw byte track (one message per group). The transcript is a side-channel,
```
kixelated: The problem with one message per group is that all of the message become lossy. You probably want them to be reliable?

vipyne: thanks for pointing this out. wdyt about handling this in a future PR? feels like the right thing to do, but maybe isn't a show-stopper for an mvp.

kixelated: I'm publishing a fixed Python version now. Adds a `Stream` functionality and renames the previous functionality to `Snapshot`. Will make a PR.

kixelated: @vipyne I would do https://github.com/pipecat-ai/pipecat/pull/5015 (and the JS changes it links to).

If you don't like the gzip compression, it's pretty trivial to remove (flip the boolean), but it reduces the bitrate by ~90%.

I'm also fine if you want to keep the existing lossy JSON messages, but you should change the JS side to not use `Json.Producer`. It uses JSON Merge Patch if the delta is smaller than making a new snapshot, but the Python consumer currently set up doesn't read them.

vipyne: done

filipi87:
Hi @vipyne, you can run `./scripts/fix-ruff-and-typecheck.sh` to fix the formatting check that's failing.

filipi87:
@vipyne, I also asked Claude to review it, it found a couple of issues probably worth checking if a fix is needed. 

## Bugs & Correctness

- [x] **High** — `src/pipecat/runner/run.py:751-793`

  The `POST /start` handler's `moq` branch reads `args.moq_host`, `args.moq_port`, `args.moq_path`, and `args.moq_tls_host`, but these attributes are only ever set by `_validate_moq_args()` (`src/pipecat/runner/moq.py`), which only runs at startup when `args.transport == "moq"` (`run.py:1701`). In the default "all transports" mode (`args.transport is None`), `moq` is still listed in `ALL_TRANSPORTS` and `_transport_routes_enabled()` only checks module importability (`run.py:226-235`), not whether `_validate_moq_args` ran — so a client POSTing `{"transport": "moq"}` in that mode reaches line 762 and raises an unhandled `AttributeError` (HTTP 500).

  Suggested fix: guard the `moq` branch with an explicit check that `_validate_moq_args`-derived attributes exist (or run validation lazily/idempotently), and return a clean 400 instead of crashing.

- [x] **High** — `src/pipecat/transports/moq/agent.py:105-125`

  `MOQAgentSession._run` only awaits `self._forward_peer_audio(self._peer_broadcast)`. Unlike `MOQTransport._consume_peer` (`transport.py:897-902`), it never calls `_forward_peer_transcript`, so client→bot RTVI messages (client-ready negotiation, typed text input, function-call results, DTMF) are never delivered for sessions run through `MOQAgentServer`/`MOQAgentSession`. This contradicts the module's own docs, which state the transcript side-channel is "inherited unchanged from `MOQTransport`," and silently breaks bidirectional RTVI for the deployable multi-session agent server.

  Suggested fix: have `_run` also drive `_forward_peer_transcript` (e.g. via `asyncio.gather`, mirroring `_consume_peer`).

- [ ] **Medium** — `src/pipecat/transports/moq/transport.py:897-902`

  `asyncio.gather(self._forward_peer_audio(...), self._forward_peer_transcript(...))` is called without `return_exceptions=True`. Both callees only catch `asyncio.CancelledError` internally, so any other exception (e.g. from `subscribe_audio`/`subscribe_json_stream`) propagates and `gather` re-raises immediately while leaving the sibling coroutine's task running, uncancelled. The enclosing `async with self._make_transport(origin)` closes the connection on unwind but doesn't explicitly cancel the orphaned task, so it can keep running or die late/indirectly rather than being torn down deterministically.

  Suggested fix: wrap the gather in a `TaskGroup`/explicit cancellation of the sibling task on failure, or pass `return_exceptions=True` and handle both results explicitly.

- [ ] **Medium** — `src/pipecat/runner/moq.py:65-96`

  In `_validate_moq_args`, the `--moq-connect` URL is parsed and validated unconditionally (lines 65-83) before the server-mode branch (line 91 `if args.moq_serve:`) that's supposed to warn "`--moq-connect is ignored in server mode`" (line 94-96). Since client mode is currently disabled entirely (line 53 always requires `--moq-serve`), a user who passes `--moq-serve` together with a malformed `--moq-connect` gets a hard validation failure and the server refuses to start — even though the option is documented as ignored in that mode.

  Suggested fix: skip URL validation for `--moq-connect` when `args.moq_serve` is set, or validate it only when it will actually be used.

## Performance

- [ ] **Medium** — `src/pipecat/transports/moq/transport.py:169-191` (call site `transport.py:987`)

  `_downmix_s16_to_mono` is a pure-Python nested loop (per-sample, per-channel, using stdlib `array`) that runs synchronously inside `_forward_peer_audio`'s per-packet `async for frame in consumer` receive loop whenever a peer publishes stereo audio — noted in the code as the common case (browsers/macOS often default to `channelCount=2` even for mono mics). Each call does on the order of hundreds of Python-level iterations (e.g. ~640 samples for a 20ms/16kHz/stereo packet). In `MOQAgentServer`'s shared-event-loop multi-session mode this CPU-bound work blocks every other concurrent session's audio/RTVI processing while it runs. The codebase already has an idiomatic numpy-vectorized pattern for equivalent 16-bit PCM math one call away (`src/pipecat/audio/utils.py:mix_audio`, using `np.frombuffer`/`astype`/`clip`), and numpy is a core dependency.

  Suggested fix: vectorize with numpy, e.g. `np.frombuffer(pcm, dtype=np.int16).reshape(-1, channels).mean(axis=1)`.

## Repository Conventions & Docstrings

- [ ] **Medium** — `src/pipecat/transports/moq/transport.py:1063`

  `MOQTransport.disconnect()` cancels `self._connection_task` with a raw `self._connection_task.cancel()` + manual `try/await/except asyncio.CancelledError`, instead of going through the task manager. This is inconsistent within the same file — `_run()` (`transport.py:817`) cancels `serve_task` via `await self._task_manager.cancel_task(serve_task)` — and with sibling transports (`daily/transport.py:763`, `vonage/client.py:362`, `tavus/transport.py:472`), which all use `self._task_manager.cancel_task(...)`.

  Rule:

  > "Async Task Management: Always use `self.create_task(coroutine, name)` instead of raw `asyncio.create_task()`. The `TaskManager` automatically tracks tasks and cleans them up on processor shutdown. Use `await self.cancel_task(task, timeout)` for cancellation."

  Source: `AGENTS.md`

- [ ] **Low** — `src/pipecat/transports/moq/agent.py:69-98,147-189`

  `MOQAgentSession` and `MOQAgentServer` both place their `Args:` section on the *class* docstring and leave `__init__`'s docstring as a bare one-liner with no `Args:` section, inverting the documented convention (compare `src/pipecat/services/heygen/client.py`, where the class docstring states purpose and `__init__` carries `Args:`).

  Rule:

  > "Docstrings: Google-style. Classes describe purpose; `__init__` has `Args:` section; dataclasses use `Parameters:` section."

  Source: `AGENTS.md`

- [ ] **Low** — `src/pipecat/runner/types.py:266-308`

  `MOQRunnerArguments`'s `Parameters:` docstring section documents most fields but omits `participant_id` (line 299) and `peer_id` (line 300). Sibling `RunnerArguments` subclasses in the same file (e.g. `VonageRunnerArguments`, `SmallWebRTCRunnerArguments`) document every field.

  Rule:

  > "dataclasses use `Parameters:` section"

  Source: `AGENTS.md`

- [ ] **Low** — `src/pipecat/transports/moq/transport.py:327-473`

  Every overridden lifecycle/IO method on `MOQInputTransport`/`MOQOutputTransport` (`setup`, `start`, `stop`, `cancel`, `process_frame`, `send_message`, `write_audio_frame`, `push_received_audio`) has a one-line docstring with no `Args:` section for its `frame`/`setup`/`audio`/`sample_rate` parameters, while the base class methods they override (`base_output.py:123-333`) and sibling transport overrides (`websocket/server.py`, `livekit/transport.py`, `daily/transport.py`) consistently document them. This is systematic across the file rather than an isolated case.

  Rule:

  > "Docstrings: Google-style. Classes describe purpose; `__init__` has `Args:` section"

  Source: `AGENTS.md`

examples/transports/transports-moq-local.py
filipi87, line 1: I am not sure we should keep this example. It is basically identical to `transports-moq.py`, the only real difference is swapping `Deepgram/Cartesia/OpenAI` for MLX `Whisper/Ollama/Kokoro`. 

We don't do that for any of the other transports.

examples/transports/transports-moq.py
```diff
@@ -0,0 +1,206 @@
...
+
+    # With a custom namespace (different "room"):
+    uv run python examples/transports/transports-moq.py \\
+        -t moq --moq-serve --moq-tls-generate localhost --moq-namespace my-room
+
+    # Then open http://localhost:7860 and click Connect.
+
+    # Can also run with other transports:
+    uv run python examples/transports/transports-moq.py -t webrtc
+    uv run python examples/transports/transports-moq.py -t daily
```
filipi87: I think we should remove this part. In all the other transport specific examples, we only show how to run that particular transport.

Also, here we're only setting `transport_params` for MOQ, so I don't think this example would work for the other transports.

examples/transports/transports-moq-agent.py
filipi87, line 1: I think this is out of scope for Pipecat.

It conflicts with the dev runner, `MOQAgentServer` is a different server for handling multiple connections, with its own lifecycle and even its own console script entry point (moq-voice-agent in pyproject.toml), instead of going through the runner.

Also, `moq/server.py` contains a vendor stack (`Deepgram/OpenAI/Cartesia`) hardcoded inside the Pipeline.

I believe `transports-moq-agent.py`, `agent.py`, and `server.py` should be removed from Pipecat, and could instead be added as a specific MoQ server example in pipecat-examples.

If the announcement driven multi-session pattern is worth keeping as a reference, I believe it belongs in `pipecat-examples` (a deployable sample wired to a specific vendor stack, using the transport as installed rather than as internal modules) rather than in pipecat-ai. This would be consistent with how other vendor-specific, opinionated deployables live outside the core package.

cc: @markbackman for a second opinion on this.

scripts/moq-dev-setup.sh
filipi87, line 1: TODO: We just need to remember to remove this script before merging this PR.

src/pipecat/transports/moq/agent.py
filipi87, line 1: See my comment here:
- https://github.com/pipecat-ai/pipecat/pull/4629/changes#r3590107709

We can discuss there whether it's worth keeping or not.

src/pipecat/transports/moq/server.py
filipi87, line 1: See my comment here:
- https://github.com/pipecat-ai/pipecat/pull/4629/changes#r3590107709

We can discuss there whether it's worth keeping or not.

tests/test_evals_suite.py
```diff
@@ -37,7 +37,7 @@
 class TestEvalManifestLoad(unittest.TestCase):
     def setUp(self):
         self._tmp = tempfile.TemporaryDirectory()
-        self.base = Path(self._tmp.name)
+        self.base = Path(self._tmp.name).resolve()
```
filipi87, line 40: Was this change needed ?

MOQ_DEV.md
filipi87, line 1: Once we've released all the client libraries and updated our default prebuilt, will we still need to keep this?
Or should we remove it before merging this PR?

src/pipecat/processors/frameworks/rtvi/processor.py
```diff
@@ -525,9 +525,20 @@ async def _send_bot_ready(self, about: Mapping[str, Any] | None = None):
         """
         if not about:
             about = {"library": "pipecat-ai", "library_version": f"{pipecat_version()}"}
+        # Negotiate the protocol version we advertise back to the client.
```
filipi87, line 528: Yep, I think this fix makes sense. It's probably even worth adding its own changelog entry.

src/pipecat/runner/moq.py
```diff
@@ -0,0 +1,234 @@
...
+
+import argparse
+from typing import Any
+from urllib.parse import urlparse
+
+from loguru import logger
+
+DEFAULT_MOQ_CONNECT = "https://localhost:4080/moq"
+DEFAULT_MOQ_SERVE_BIND = "[::]:4080"
+DEFAULT_MOQ_PATH = "/moq"
```
filipi87, line 21: Should we turn these into environment variables that users can override? We could use these as the default values if they aren't provided.

filipi87: Instead of `localhost`, should this use the `host` argument that we receive from the dev runner? Or should it always be `localhost`?

filipi87: My bad, it looks like we are already handling it like this. So ignore my comments here. 😅

src/pipecat/runner/moq.py
```diff
@@ -0,0 +1,234 @@
...
+    # one-line delete of this guard once client mode is ready.
+    if not args.moq_serve:
+        logger.error(
+            "MoQ client mode is not yet supported. "
+            "Pass --moq-serve to run the bot as its own MoQ server "
+            "(with --moq-tls-generate <hostname> for a self-signed dev cert, or "
+            "--moq-tls-cert/--moq-tls-key for a CA-signed one)."
+        )
+        return False
+    # ------------------------------------------------------------------
```
filipi87, line 61: If this isn't supported yet, couldn't we make `args.moq_serve` the default behavior so users don't have to specify it?

src/pipecat/runner/run.py
```diff
@@ -568,6 +591,11 @@ class StartBotResult(TypedDict, total=False):
         dailyToken: str | None
         wsUrl: str | None
         token: str | None
+        namespace: str | None
+        relay: str | None
```
filipi87: Just curious: below we mention that everything needed by MoQ is contained in that `moq` dictionary. Why did we choose to keep the `namespace` and `relay` outside of it ?

Also, instead of using a generic dictionary, couldn't we create a class for it, similar to what we did for `IceConfig`?

filipi87: Actually, it looks like `relay` and `namespace` are already included in that dictionary. Do we still need to keep them here?

src/pipecat/runner/run.py
```diff
@@ -1510,6 +1583,105 @@ def main(parser: argparse.ArgumentParser | None = None):
...
+        metavar="ADDR:PORT",
+        help=(
+            f"Local socket bind address. Server mode default: {DEFAULT_MOQ_SERVE_BIND}. "
+            "(Client mode: ephemeral if omitted; MoQ client mode is not yet supported.)"
+        ),
+    )
+    parser.add_argument(
+        "--moq-serve",
+        action="store_true",
+        default=False,
```
filipi87: I believe this should default to `True`. Since it is the only one supported. 

What do you think ?

src/pipecat/runner/utils.py
```diff
@@ -762,5 +763,44 @@ async def create_transport(
...
+        # MOQ bring-up, copy out the cert fingerprints (serve mode) and
+        # signal the runner's ready_event so /start can return. This used
+        # to require boilerplate in every bot example.
+        if runner_args.ready_event is not None:
+
+            @transport.event_handler("on_connected")
+            async def _moq_runner_on_connected(t):
+                runner_args.cert_fingerprints = list(t.cert_fingerprints)
+                if runner_args.ready_event is not None:
+                    runner_args.ready_event.set()
```
filipi87, line 801: I wonder if a better approach would be to pass the `ready_event` directly to the transport, so it can signal it when it's ready to receive connections. We could also retrieve the `cert_fingerprints` from the transport.

Another approach would be to hide this behind a handler, similar to what we did with `SmallWebRTCRequestHandler`.

Otherwise, I think we might run into some issues when integrating this with PCC.

That said, I am fine with keeping it as is for now and addressing it in a follow-up PR.

src/pipecat/transports/moq/transport.py
```diff
@@ -0,0 +1,1083 @@
...
+(`moq-rs <https://pypi.org/project/moq-rs/>`_) for the QUIC connection,
+MOQ session, announcement discovery, subscription routing, group/frame
+framing, codec-specific catalog management, and Opus encode/decode +
+resampling for raw audio tracks. This module just wires it into the
+pipecat Frame pipeline.
+
+Each participant publishes under a per-participant broadcast path
+``<namespace>/<participant_id>`` (e.g. ``pipecat/bot0``); the bot
+subscribes to the peer at ``<namespace>/<peer_id>``. Audio rides on a
+single Opus track; RTVI JSON rides on a fixed-name ``transcript.json.z``
```
filipi87, line 19: nit: RTVI sends a lot more information than just transcription. Would it make sense to call this something other than `transcript.json.z`?

src/pipecat/transports/moq/transport.py
```diff
@@ -0,0 +1,1083 @@
...
+        self._register_event_handler("on_disconnected")
+        self._register_event_handler("on_client_connected")
+        self._register_event_handler("on_client_disconnected")
+        self._register_event_handler("on_track_subscribed")
+        self._register_event_handler("on_error")
+
+    def input(self) -> MOQInputTransport:
+        """Get the input transport for receiving media."""
+        if not self._input:
+            self._input = MOQInputTransport(self, self._params, name=self._input_name)
```
filipi87: In all the other transports (Daily, SmallWebRTC, Tavus ...), instead of passing the transport itself into the  input/output transport, we create a transport client (e.g. `DailyTransportClient`, `LiveKitTransportClient`) that owns the actual connection/session logic, and both input and output transports hold a reference to that client.

Here, `MOQInputTransport`/`MOQOutputTransport` instead take a reference to `MOQTransport` itself and call into it directly (open_audio_track, publish_audio, publish_transcript, ...). It goes further than that, too, `MOQInputTransport.start()` reaches into `MOQTransport's` private state directly: 
```
self._moq_transport._connection_task
  = self.create_task(self._moq_transport._run(), "moq_run").
```

I think we should follow the established pattern here: pull the moq-rs session/connection logic out into a  `MOQTransportClient`, and have input/output hold a reference to that instead of to the parent transport.

src/pipecat/runner/moq.py
kixelated: Heh I was going to add support right after this gets merged.

kixelated: And I think the long-term default should actually be run as a client.

Cloudflare, AWS, Fastly, Akamai, etc are working on a MoQ CDN. You're going to get better QoS connecting pipecat to the nearest CDN edge as a client. 

And even if you do host your own relay, it's probably still better to have pipecat connect to it via localhost. That way you can use the relay/port for more than just pipecat.

src/pipecat/transports/moq/transport.py
kixelated: One thing worth mentioning is that MoQ doesn't care about client vs server, except for establishing the connection. In fact, you can establish multiple connections (ex. P2P + connect to a CDN + listen on localhost) and it automatically works.

src/pipecat/runner/moq.py
filipi87: It is, because right now this is the only option. We can make it the default for now so users don't have to specify it when running the development runner.

The same applies when running the development runner in a mode that supports all transports.

Once we release a version that supports running as a client, we can change that.

src/pipecat/transports/moq/transport.py
filipi87: Hi @kixelated, yep, this is basically the same as the `SmallWebRTC` transport, where both sides are just peers.

My comment here was more about how we're organizing the code in the other transports, and to keep the same approach here. It isn't related to whether MoQ is actually running as a client or a server.

tests/test_evals_suite.py
filipi87: Ok, it looks like `.resolve()` is needed here to prevent failures in case the temp dir sits behind a symlink. So we should keep it here. 👍

examples/transports/transports-moq-local.py
filipi87: Ok. I have removed this, so we matched the same approach that we do for the other transports.

examples/transports/transports-moq.py
filipi87: Done.

src/pipecat/runner/moq.py
filipi87: I have implemented the changes above, so the development runner can now work without requiring any extra arguments, supporting all transports at the same time.

Once we add support for MoQ client mode, we can make it the default instead.

src/pipecat/runner/run.py
filipi87: I have implemented the changes above, so the development runner can now work without requiring any extra arguments, supporting all transports at the same time.

Once we add support for MoQ client mode, we can make it the default instead.

examples/transports/transports-moq-agent.py
filipi87: @vipyne, I am addressing this in this PR, so you can review it before we merge it back here:
- https://github.com/pipecat-ai/pipecat/pull/5059

scripts/moq-dev-setup.sh
filipi87: @vipyne, I am addressing this in this PR, so you can review it before we merge it back here:
- https://github.com/pipecat-ai/pipecat/pull/5059

src/pipecat/transports/moq/agent.py
filipi87: @vipyne, I am addressing this in this PR, so you can review it before we merge it back here:
- https://github.com/pipecat-ai/pipecat/pull/5059

src/pipecat/transports/moq/server.py
filipi87: @vipyne, I am addressing this in this PR, so you can review it before we merge it back here:
- https://github.com/pipecat-ai/pipecat/pull/5059

MOQ_DEV.md
filipi87: @vipyne, I am addressing this in this PR, so you can review it before we merge it back here:
- https://github.com/pipecat-ai/pipecat/pull/5059

src/pipecat/runner/run.py
filipi87: @vipyne, I am addressing this in this PR, so you can review it before we merge it back here:
- https://github.com/pipecat-ai/pipecat/pull/5059

src/pipecat/transports/moq/transport.py
filipi87: @vipyne, I am addressing this in this PR, so you can review it before we merge it back here:
- https://github.com/pipecat-ai/pipecat/pull/5059

filipi87:
@vipyne, I believe I've addressed all of my comments in this follow-up PR:
- https://github.com/pipecat-ai/pipecat/pull/5059

Let me know what you think of these changes and whether we should merge them here.

filipi87 (approved):
LGTM! Nice addition, @vipyne. 🚀

Thank you for all the great reviews, @kixelated! We can address any remaining improvements in follow-up PRs so they're easier to review.

PR #4631, opened 2026-06-03
Port the pipecat init CLI into pipecat-ai (optional [cli] extra)

markbackman:
## Summary

Ports the `pipecat init` project-scaffolding CLI from the separate `pipecat-ai-cli` package into `pipecat-ai` as a new `pipecat.cli` subpackage, exposed as an **optional `[cli]` extra**. Installing `pipecat-ai[cli]` now gives you the `pipecat` / `pc` commands, consolidating on a single canonical CLI. The CLI's dependencies stay out of a plain `pip install pipecat-ai`, so the runtime/base-image install is unaffected.

> **Held for the 1.4.0 release.** Generated bots pin `pipecat-ai[...]>=1.4.0` (they use `create_transport` + the typed `CallData`/runner-args API), so this shouldn't merge until 1.4.0 is out.

## What's included

- **`pipecat init` + the CLI skeleton** — the full scaffolding command (generators, registry, 66 templates, config validation), plus `pipecat init .` to scaffold in place. Supports interactive prompts and a fully non-interactive mode (`--name`/flags or `--config`).
- **Optional `[cli]` extra** (`typer`, `rich`, `jinja2`, `questionary`, `ruff`) with console scripts `pipecat` / `pc`. No `pipecatcloud` dependency — cloud co-installs via `--with`.
- **Plugin discovery + graceful enable hints** — extensions register a Typer app under the `pipecat_cli.extensions` group and mount automatically when installed. The official `cloud` plugin (pipecatcloud) is listed in `pipecat --help` even when it isn't installed (annotated `(requires pipecatcloud)`); invoking it prints how to enable it instead of a bare "No such command". Install hints are scope-organized (global tool vs project/venv), leading with `uv`.
- **Graceful-degradation shim** — the console-script target prints an install hint and exits cleanly (no traceback) when the `[cli]` extra is absent; the module stays importable without `typer`, and nothing in core `pipecat` imports `pipecat.cli`.
- **Quieter CLI** — the `ᓚᘏᗢ Pipecat <version>` import banner is suppressed when running the `pipecat`/`pc` CLI (it cluttered output, including `--help`); bots and library imports still log it.
- **Templates packaged** via `package-data` globs + a MANIFEST `graft`; a built wheel was verified to contain all 66 files including the `.gitignore` dotfiles, `env.example`, bracketed Next.js route dirs, and `pipecat.svg`.
- **Registry kept in sync with the source** — generator tooling under `scripts/cli/` introspects the in-repo source (also fixes a module-path bug that surfaced `pipecat.src.pipecat.…` on the editable tree). A pre-commit hook regenerates `_imports.py` from a pure AST walk (no service SDKs needed), and a CI `--all-extras` drift check regenerates everything and fails on drift — catching import-path moves and constructor-signature changes on the PR that makes them.
- **Test suite** ported to `tests/cli/`, then tidied for the monorepo (dropped registry-completeness tests now covered by the drift checks; added coverage for import branching, flag/config-file precedence, websocket generation, and plugin discovery).

## Verification

- `uv run --extra cli pytest tests/cli/` → **267 passed**.
- `pipecat init` generates a valid project (`bot.py` parses + is ruff/pyflakes-clean; pins `pipecat-ai[...]>=1.4.0`).
- Graceful degradation confirmed (bare install errors cleanly; `[cli]` works; `pipecat --version` reports the pipecat-ai version).
- `pipecat cloud` registers via `--with pipecatcloud`; an uninstalled `pipecat cloud` prints the enable hint; registry regenerates byte-identical against local source; `ruff check` clean repo-wide (CLI now under pipecat's Google-style `D` docstring rules); core `import pipecat` unaffected.

## Notes / follow-ups

- **CI**: the `tests` and `coverage` workflows install the `cli` extra so `tests/cli/` runs; a `tests/cli/conftest.py` skips that directory's collection when the extra is absent, so a lean `uv run pytest` still works.
- **`tail` is deferred** — `pipecat-ai-tail` isn't ready to advertise, so Tail is not surfaced in the CLI: it's removed from plugin discovery and from the `--observability` scaffold (generated bots wire `WhiskerObserver` only). It'll light up automatically via the entry-point group once published.
- The `pipecat` / `pc` console scripts install with `pipecat-ai` regardless of the `[cli]` extra (they're `[project.scripts]`); without the extra they hit the graceful shim. `pc` is a short, generic global name — flagging in case we'd rather ship only `pipecat`.
- The `pipecat_cli.extensions` entry-point group name is kept (transitional — avoids forcing plugin re-releases).
- Out of scope: deprecating `pipecat-ai-cli` to a redirect shim, renaming the entry-point group, docs/quickstart updates.


changelog/4631.added.md
```diff
@@ -0,0 +1 @@
+Added the `pipecat init` project-scaffolding CLI to `pipecat-ai`, available via the optional `cli` extra. Install it with `uv tool install "pipecat-ai[cli]"` (add `--with pipecatcloud` to enable `pipecat cloud`), then run `pipecat init` to scaffold a new bot project. The CLI dependencies are optional, so they are not pulled into a plain `pip install pipecat-ai`.
```
kompfner, line 1: Do you want to mention `pipecat tail` here too?

markbackman: I think it depends. @aconchillo are we still maintaining tail or was it more of an experiment?

kompfner: How exactly will a user discover these "sub-CLIs"?

Are they suggested to the user when they do `pipecat --help`? If so: do we show them, in the help output, the `uv` command needed to enable those sub-CLIs?

Also: if the user attempts to use `pipecat cloud` and the Pipecat Cloud CLI isn't installed, do we show a helpful message showing them how to enable it?

Also, to check my understanding: does `pipecat cloud <arg>` work by just forwarding all arguments to the Pipecat Cloud tool (`pcc`)?

kompfner:
Reminder that we ought to update our published [Pipecat Skills](https://github.com/pipecat-ai/skills) accordingly. Specifically the init and deploy skills.

Wonder if we need to skill to consider the version of Pipecat that it has access to so it can conditionally grab the CLI from either the Pipecat repo or the CLI's old home repo.

changelog/4631.added.md
markbackman: We're relying on docs. We aren't including `pipecatcloud` in `pipecat-ai` since that seems like a strange dependency to require.

`uv tool install` will be the primary way a developer uses the tool. The requirement is that the packages need to be installed in the same venv, which for `uv tool install` means added in a single command. Options being: a) pull in `pipecatcloud` as a dependency, which we've ruled out, b) use the `--with pipecatcloud` option.

> Are they suggested to the user when they do pipecat --help? If so: do we show them, in the help output, the uv command needed to enable those sub-CLIs?

This is a good idea. I'll have to see what options we have for this.

> Also: if the user attempts to use pipecat cloud and the Pipecat Cloud CLI isn't installed, do we show a helpful message showing them how to enable it?

I'll make sure this is correctly handled. My hunch is that it's just a generic error message now.

> Also, to check my understanding: does pipecat cloud <arg> work by just forwarding all arguments to the Pipecat Cloud tool (pcc)?

`pipecatcloud` is an extension, which can be installed with the Pipecat CLI. So, it calls in to the `pipecatcloud` API to execute commands.

src/pipecat/cli/main.py
```diff
@@ -0,0 +1,113 @@
...
+
+The Typer app is built lazily (``_build_app``) so importing this module never pulls
+in the optional CLI dependencies; the module-level ``app`` attribute (used by tests
+and the console script) is resolved on first access via ``__getattr__``.
+"""
+
+import sys
+
+_INSTALL_HINT = (
+    "The Pipecat CLI requires additional dependencies that are not installed.\n\n"
```
kompfner: I like that we're giving a hint like this for users to discover that they need to install the extra dependencies!

The wording feels a bit hand-wavy and maybe misleading, though. The decision to use `pip` v `uv` doesn't seem like it should come down to whether you want "an isolated tool", right? (I'm not sure what "isolated tool" means, either).

Based on my (albeit limited) understanding it seems like the user's choices are:
- Whether to use `pip` or `uv`
- Whether to install the CLI as a project-scoped dev tool or globally (you can do either with either tool)

kompfner: In various other files (upwards of 90, it seems) we log a message like [this one](https://github.com/pipecat-ai/pipecat/blob/7cf8221a88dcce5f12b89c229d182d7877ecebee/src/pipecat/services/google/llm.py#L67), where we show the `pip` usage.

But in our README, we present `uv` as the recommended tool.

My inclination: let's only show `uv` in this hint, and take it as an action item to update our hints throughout the codebase to also use `uv`.

changelog/4631.added.md
markbackman: Ok, I've improved this in two ways:

1. The commands are always exposed now:
CleanShot 2026-06-05 at 12 41 43@2x

2. When you try to use a command, you get a helpful error message:
CleanShot 2026-06-05 at 12 42 20@2x

I think the remaining decision is if we want to expose `tail`.

src/pipecat/cli/main.py
markbackman: I made the wording more precise:
CleanShot 2026-06-05 at 12 49 56@2x

changelog/4631.added.md
markbackman: Improved this wording per your other comment.

src/pipecat/cli/main.py
kompfner: Why prefer `uv` for one recommendation and `pip` for the other? That feels a bit inconsistent. I'd recommend having `uv` as the preferred way for both, and `pip` for the fallback (the `# or:`).

changelog/4631.added.md
kompfner: This is great!

changelog/4631.changed.md
```diff
@@ -0,0 +1 @@
+The `ᓚᘏᗢ Pipecat <version>` startup banner is no longer logged when running the `pipecat` / `pc` CLI (it cluttered command output, including `--help`). Bots and library imports still log it.
```
kompfner, line 1: 👍 

Nice catch.

changelog/4631.added.md
kompfner: One question: do we want to generalize the string "CLI tool for scaffolding Pipecat AI voice agent projects", if that's only one of the things (though admittedly the primary thing) that it will be for?

src/pipecat/cli/main.py
markbackman: Good point. Fixed.

changelog/4631.added.md
markbackman: Yes, generalizing is safer. This was copied over from the standalone CLI.

Going with:
```
Command-line tools for building Pipecat AI applications.
```

kompfner (approved):
LGTM!

changelog/4631.added.md
aconchillo: About `tail`, we don't have to mention it today. `tail` needs an update. i'll update it at some point, and then we can just add it here.

changelog/4631.changed.md
aconchillo: nice!

aconchillo (approved):

markbackman:
FYI: Removed `tail` after discussing with @aconchillo.

PR #4632, opened 2026-06-03
Default Gradium STT delay_in_frames to 12 (960ms)

markbackman:
## Summary

- `GradiumSTTService` now defaults `delay_in_frames` to `12` (960ms) instead of leaving it unset, which used the server default of 10/800ms.
- The higher default trades a small amount of latency for more context, improving transcription accuracy out of the box.
- Updated the `delay_in_frames` docstring to reflect the new default.

## Behavior Change

- Pipelines that did not set `delay_in_frames` explicitly will now see ~160ms higher STT latency (960ms vs. 800ms). Set `delay_in_frames` to `7`-`8` for faster responses.


kompfner (approved):
Makes sense. This was based on your observations from testing?

markbackman:
> Makes sense. This was based on your observations from testing?

I've been in touch with the Gradium team about including them in the benchmark. This is the setting they're recommending. I'm re-running the benchmark with this value now. I'll wait to confirm the result before merging this.

PR #4635, opened 2026-06-03
fix(inworld): guard against null websocket in run_tts after failed connect

ianbbqzy:
## Summary

`InworldTTSService.run_tts` now checks `self._websocket is None` immediately after calling `_connect()` and returns early with an `ErrorFrame` when the connection attempt failed.

`_connect_websocket()` swallows its own exceptions and returns with `_websocket = None` on failure. Without this guard, `run_tts` falls through into `send_with_retry`, which logs a misleading `"no websocket connected, will try to reconnect"` warning and kicks off a redundant second reconnect attempt for what was already a single connection failure. The fix mirrors the same null guard pattern used by `ElevenLabsTTSService`.

## Reconnect behavior unchanged

- Background `_receive_task_handler` still catches mid-session `ConnectionClosedError` and retries via `_maybe_try_reconnect` (3 attempts, exponential backoff).
- `send_with_retry` still retries the send after a successful reconnect.

## Test plan

- [ ] Inworld endpoint unreachable on startup: log no longer shows `"no websocket connected, will try to reconnect"` — instead an `ErrorFrame("websocket unavailable")` is emitted and no redundant reconnect is attempted.
- [ ] Mid-session connection drops still trigger automatic reconnection via the background receive task.


aconchillo (approved):

aconchillo:
LGTM. Thanks!

PR #4637, opened 2026-06-04
Add on_tts_request handler to Cartesia examples for context_id tracing

zeuslawyer:
## Summary

- Adds an `on_tts_request` event handler to `examples/voice/voice-cartesia.py` and `examples/voice/voice-cartesia-turns.py`
- The handler logs the `context_id` and `text` for each TTS request, enabling end-to-end tracing between application logs and Cartesia's context-level audio output
- Includes a docstring explaining the purpose, the auto-generated `context_id` mechanism, and links to the relevant Pipecat and Cartesia docs

## Test plan

- [ ] Verify the handler fires on each TTS request when running either example bot
- [ ] Confirm `context_id` appears in logs alongside the spoken text
- [ ] Confirm no regressions to existing bot behaviour (pipeline still starts, STT/LLM/TTS flow intact)

### DIFF
--- a/examples/voice/voice-cartesia-turns.py
+++ b/examples/voice/voice-cartesia-turns.py
@@ -113,7 +113,7 @@ async def run_bot(transport: BaseTransport, runner_args: RunnerArguments):
         await worker.cancel()
 
     @tts.event_handler("on_tts_request")
-    async def on_tts_request(tts, context_id, text):
+    async def on_tts_request(tts, context_id: str, text: str):
         """Surface the context_id for debugging and tracing TTS requests.
 
         Pipecat auto-generates a context_id (UUID) per LLM turn and passes it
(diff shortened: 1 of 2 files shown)


jamsea:
@zeuslawyer the only change I'd recommend is switching from `.info` to `.debug` for the logging like copilot suggested. I'm indifferent about adding the Python types. Otherwise looks good to me cc @markbackman

zeuslawyer:
thanks @jamsea  - addressed all comments because type annotations can only help esp new devs :)

markbackman:
@zeuslawyer we have documentation for this event handler:
https://docs.pipecat.ai/api-reference/server/events/service-events#on_tts_request

Adding it to this example is a little strange as it just console logs and isn't really used. I would opt to not include since it's not something that's unique to Cartesia or necessary for the example. If you're interested, you could add it to the Cartesia docs section as a call out.

Contexts are handled automatically, so this isn't really something that developers have to think about. What use case is someone running into that requires knowledge of the underlying context?

markbackman (approved):
LGTM!

PR #4641, opened 2026-06-04
bus: dedup duplicate frame deliveries at BusBridgeProcessor

### LINKED ISSUE #4628: bus: _BusEdgeProcessor echo-loops in-flight frames when 2+ workers are briefly active during handoff
Caught this digging into duplicate `FunctionCallResultFrame` entries showing up in an `LLMAssistantAggregator`'s context. Started off thinking it was a generic bus rebroadcast issue (opened #4624 for that, then closed it after the actual analysis — referencing here for context, no need to revisit that one).

The duplicates only show up for tool handlers that trigger a worker handoff (handler calls `activate_worker(target, deactivate_self=True)` then returns a result dict so flows fires `result_callback`). Tools handled entirely within a single worker are clean. Same `broadcast_frame(FunctionCallResultFrame, ...)` in both cases, the down+up siblings have distinct ids — that part is working as intended.

Instrumented `_BusEdgeProcessor.process_frame` / `on_bus_message` and `LLMAssistantAggregator._handle_function_call_result` to track frame ids end to end. For a handoff tool call (worker_a calls a tool that activates worker_b):

```
worker_a  EDGE_TX_BUS (down)  frame=585           ← initial bus send
main      BRIDGE_RX_BUS       frame=585
worker_a  EDGE_TX_BUS (up)    frame=586           ← upstream sibling
main      BRIDGE_RX_BUS       frame=586

worker_b  EDGE_TX_BUS (down)  frame=585           ← ECHO: worker_b re-broadcasts frame 585
worker_b  EDGE_TX_BUS (up)    frame=586           ← ECHO: and 586
main      BRIDGE_RX_BUS       frame=585           ← main receives 585 again
main      BRIDGE_RX_BUS       frame=586

AGG_RX    self=LLMAssistantAggregator#0  frame_id=585  (1st)
AGG_RX    self=LLMAssistantAggregator#0  frame_id=585  (2nd, same Python instance)
```

For a non-handoff tool only one `EDGE_TX_BUS` sequence, no echo, aggregator sees the result once.

What's happening: when worker_b becomes active during the handoff transition, the in-flight `FunctionCallResultFrame` from worker_a still gets delivered to it via the bus. `BusBridgeProcessor.on_bus_message` injects the frame into worker_b's pipeline through `push_frame`. The frame travels through worker_b's pipeline and reaches its own `_BusEdgeProcessor`, which doesn't know the frame came from the bus in the first place and calls `bus.send` on it again. Same Python instance, same id, re-broadcast. Main's `BusBridgeProcessor` then receives the same frame twice, and the aggregator processes it twice → duplicate `assistant.tool_calls` + `tool` result pairs in the LLM context.

So it's not "2 active children in the same bridge = user misconfiguration". It's a transient race during the activate/deactivate transition itself — `activate_worker(deactivate_self=True)` doesn't make the swap atomic with respect to in-flight frames. As long as the broadcast crosses the bus in that ~few-ms window, the just-activated child echoes it back.

A fix on the edge seems natural: mark frames coming in through `on_bus_message` with a "from_bus" tag (the `Frame.metadata` dict that's already on every frame is enough), and have `_BusEdgeProcessor.process_frame` skip the `bus.send` for tagged frames. Something like:

```python
# _BusEdgeProcessor.on_bus_message, before push_frame:
message.frame.metadata["_from_bus"] = True
await self.push_frame(message.frame, message.direction)

# _BusEdgeProcessor.process_frame, before bus.send:
if frame.metadata.get("_from_bus"):
    return
await self._task.bus.send(...)
```

That would also fix the same shape of issue for any other frame type carried across the bus during a handoff (TTSSpeak, MetricsFrame, FunctionCallsStartedFrame all show similar duplicate patterns in our traces, though only the FunctionCallResultFrame one has user-visible impact since it lands in the LLM context).

Happy to put up a PR with that shape if it sounds right. cc @aconchillo since we discussed this on the now-closed #4624 — wanted to surface it as a proper issue rather than ping in there.

Maurotb:
Fixes #4628.

In a multi-worker setup with `BusBridgeProcessor` plus child
`LLMWorker`s, two sibling workers briefly end up `active=True` during
a handoff transient. You can see the race in `activate_worker`:

```python
# pipecat/workers/base_worker.py
async def activate_worker(self, worker_name, args, deactivate_self):
    if self._active and deactivate_self:
        await self.deactivate_worker(self.name)         # publish, no ack
    await self.send_bus_message(
        BusActivateWorkerMessage(target=worker_name)    # publish, no ack
    )

async def deactivate_worker(self, worker_name):
    await self.send_bus_message(
        BusDeactivateWorkerMessage(target=worker_name)
    )
    # returns as soon as the message is published, doesn't wait for
    # the subscriber to flip `self._active = False`.
```

Both `deactivate` and `activate` are fire-and-forget through the bus,
so the target can flip to `active=True` before the caller's
`_handle_worker_deactivate` runs and sets `self._active = False`.
During that window both workers receive bus broadcasts and re-broadcast
frames originally emitted by the other side, so the host bridge ends
up receiving the same `frame.id` from multiple sources. The duplicate
delivery reaches the assistant aggregator and the LLM context picks
up duplicate `assistant.tool_calls` + `tool` result pairs — I hit this
in production as a caller stalling for 90s after a sub-bot return, and
as a hallucinated mid-conversation end_call.

#4624 (closed) tried to drop a whitelist of "emitter-local" frame
types on the receive side of `_BusEdgeProcessor`. That masked the
symptom but the bus loop is actually load-bearing on the upstream
direction (it's how upstream frames reach processors that sit
*upstream* of `BusBridge` in the host pipeline — e.g.
`LLMUserAggregator` so `MuteUntilFirstBotCompleteUserMuteStrategy`
can unmute the user). Skipping it on the sender side leaves SIP voice
with the user permanently muted after the greeting.

This PR dedups at the receiver instead: a bounded LRU of recently
seen `frame.id`s in `BusBridgeProcessor.on_bus_message`, so when a
sibling re-broadcasts a frame already delivered by the original
source, the second copy is silently dropped before it reaches
`push_frame`. Same shape Pipecat already uses elsewhere to dedup
bus-replicated frames by id. The loop itself isn't touched, and the
deactivation race isn't relied on.

I went with a receiver-side dedup rather than tightening the
lifecycle for two reasons. The deactivate happens through the bus
asynchronously, so even awaiting it deterministically only narrows
the window. And the sender side of `_BusEdgeProcessor.process_frame`
doesn't gate on `self._task.active` — a frame already in flight in
the soon-to-be-deactivated worker's pipeline still goes onto the bus.
Bounded LRU at the host catches both without locking the hot path.

Verified against the same scenario that opened the issue: in chat
mode the assistant aggregator sees `FunctionCallResultFrame:call_bot`
exactly once per call (vs twice without the fix, deterministic across
runs); on SIP voice the handoff race that was hitting ~33% drops to
zero across 12 runs; and a real SIP voice call goes through cleanly
— bot greets, user speaks, sub-bot handoff completes, no muted STT
regression, audit clean.

Added a regression test (`test_duplicate_frame_id_from_sibling_dropped`
plus a sanity check that distinct ids still pass through). It
reproduces the bug deterministically without the fix (asserts
`2 != 1` at the receiver).

### DIFF
--- a/src/pipecat/bus/bridge_processor.py
+++ b/src/pipecat/bus/bridge_processor.py
@@ -16,6 +16,7 @@ Provides:
   to the bus).
 """
 
+from collections import deque
 from typing import TYPE_CHECKING
 
 from pipecat.bus.bus import WorkerBus
@@ -75,6 +76,16 @@ class BusBridgeProcessor(FrameProcessor, BusSubscriber):
         self._target_task = target_task
         self._bridge = bridge
         self._exclude_frames = exclude_frames or ()
+        # Bounded LRU dedup against sibling worker re-broadcast.
+        # In a multi-worker setup, sibling workers briefly co-existing as
+        # ``active=True`` during a handoff transient re-broadcast frames
+        # originally emitted by another worker back onto the bus. The host
+        # bridge then receives the same ``frame.id`` from multiple sources
+        # (the original and N siblings). Dropping the duplicate delivery
+        # here keeps the assistant aggregator from recording duplicate
+        # tool round-trips in the LLM context.
+        self._seen_bus_ids: set[int] = set()
+        self._seen_bus_history: deque[int] = deque(maxlen=512)
 
     async def setup(self, setup: FrameProcessorSetup):
         """Subscribe to the bus during processor setup."""
@@ -144,6 +155,20 @@ class BusBridgeProcessor(FrameProcessor, BusSubscriber):
         # If message targeted at someone else, skip
         if message.target and message.target != self._worker_name:
             return
+
+        # Skip duplicate delivery when a sibling worker, briefly active
+        # during a handoff transient, re-broadcasts a frame originally
+        # emitted by another worker. The bus loop is by-design (it
+        # propagates upstream frames to processors upstream of the
+        # BusBridge in the host pipeline, e.g. LLMUserAggregator), so
+        # we dedup at the receiver instead of skipping at the sender.
+        fid = message.frame.id
+        if fid in self._seen_bus_ids:
+            return
+        self._seen_bus_ids.add(fid)
+        self._seen_bus_history.append(fid)
+        if len(self._seen_bus_ids) > len(self._seen_bus_history):
+            self._seen_bus_ids = set(self._seen_bus_history)
 
         await self.push_frame(message.frame, message.direction)
 
--- a/tests/test_bridge_processor.py
+++ b/tests/test_bridge_processor.py
@@ -263,6 +263,157 @@ class TestBusBridgeProcessor(unittest.IsolatedAsyncioTestCase):
         await processor.on_bus_message(msg)
         self.assertEqual(len(injected), 0)
 
+    async def test_duplicate_frame_id_from_sibling_dropped(self):
+        """Sibling re-broadcast of the same frame.id is dropped at the receiver.
+
+        In a multi-worker setup, sibling workers briefly co-existing as
+        ``active=True`` during a handoff transient re-broadcast frames
+        originally emitted by another worker. Without dedup the host
+        bridge delivers the same ``frame.id`` once per re-broadcasting
+        sibling, which pollutes the assistant aggregator's LLM context.
+
+        Regression: same ``frame.id`` received twice (once from the
+        original emitter, once from a sibling) is pushed only once.
+        """
+        bus = AsyncQueueBus()
+        processor = BusBridgeProcessor(
+            bus=bus,
+            worker_name="main_task",
+        )
+
+        injected = []
+        original_push = processor.push_frame
+
+        async def capture_push(frame, direction=FrameDirection.DOWNSTREAM):
+            injected.append(frame)
+            await original_push(frame, direction)
+
+        processor.push_frame = capture_push
+
+        # Same TextFrame instance (same ``frame.id``) re-broadcast by a
+        # sibling worker during the handoff transient.
+        original_frame = TextFrame(text="result")
+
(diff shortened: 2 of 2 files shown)


aconchillo:
@Maurotb Thank you! I have file this PR instead https://github.com/pipecat-ai/pipecat/pull/4691. We used to do this but I changed to an actualy deactivate message so Whisker (the debugger) could see that a deactivation is happening. So, I went back to doing `self._active = False` but keep sending the deactivate message. Would you mind trying this one out? Thank you! :pray:

src/pipecat/bus/bridge_processor.py
```diff
@@ -75,6 +76,16 @@ def __init__(
...
+        # Bounded LRU dedup against sibling worker re-broadcast.
+        # In a multi-worker setup, sibling workers briefly co-existing as
+        # ``active=True`` during a handoff transient re-broadcast frames
+        # originally emitted by another worker back onto the bus. The host
+        # bridge then receives the same ``frame.id`` from multiple sources
+        # (the original and N siblings). Dropping the duplicate delivery
+        # here keeps the assistant aggregator from recording duplicate
+        # tool round-trips in the LLM context.
+        self._seen_bus_ids: set[int] = set()
+        self._seen_bus_history: deque[int] = deque(maxlen=512)
```
aconchillo, line 88: The LRU would not work with distributed workers. The reason is because the `id` is not deserialized only fields in the data class which are `init=True` are serialized.

Maurotb:
Superseded by #4691.

PR #4642, opened 2026-06-04
Default TTSSpeakFrame.append_to_context to True; deprecate None

kompfner:
## Summary

`TTSSpeakFrame.append_to_context` now defaults to `True` instead of `None`, and `None` is no longer a supported value (it's coerced to `True` with a warning).

### Why

The unset (`None`) behavior was situation-dependent and hard to reason about. The spoken text always reached the assistant aggregator's buffer, but whether it was committed to the LLM context depended on what surrounded the frame:

- **Inside an assistant response (or immediately followed by one)** → committed (flushed at the response's `LLMFullResponseEndFrame` or on interruption).
- **Standalone, then followed by a user turn** → silently **discarded**: there's no open assistant turn, so the interruption's turn-stop is a no-op and `reset()` clears the buffer before anything flushes it.

So "speak this and (maybe) record it" depended on conversation timing a developer can't easily predict. `True` is a predictable default: programmatically-spoken text is recorded in the context unless you opt out with `append_to_context=False`.

### Changes

- `TTSSpeakFrame.append_to_context`: `bool | None = None` → `bool = True`.
- Backward compatibility: passing `None` is coerced to `True` and emits a `DeprecationWarning`. `None` support will be removed in a future release.
- `BusTTSSpeakMessage.append_to_context` defaults to `True` to match (it forwards into `TTSSpeakFrame`).
- `tests/test_piper_tts.py`: a default `TTSSpeakFrame` now emits an `LLMAssistantPushAggregationFrame` (to commit the spoken text), so the success-path frame-ordering assertion allows it.

### Compatibility note

This is a behavior change. Code that relied on the old `None` default for a standalone `TTSSpeakFrame` (where the spoken text merged into a following assistant turn, or was dropped before a user turn) will now see that text committed as its own assistant message. Opt out per-frame with `append_to_context=False`.

### DIFF
--- a/changelog/4642.changed.md
+++ b/changelog/4642.changed.md
@@ -1,0 +1,10 @@
+- ⚠️ Changed the default of `TTSSpeakFrame.append_to_context` from `None` to
+  `True`. The old `None` behavior was situation-dependent and hard to reason
+  about: the spoken text always reached the assistant aggregator's buffer, but
+  whether it was committed to the LLM context depended on what surrounded the
+  frame — committed when the frame was inside an assistant response or
+  immediately followed by one, but silently discarded when it was standalone and
+  followed by a user turn (the interruption cleared the buffer before anything
+  flushed it). `True` is a predictable default: programmatically-spoken text is
+  recorded in the context unless you opt out with `append_to_context=False`.
+  `BusTTSSpeakMessage.append_to_context` now defaults to `True` to match.
--- a/src/pipecat/frames/frames.py
+++ b/src/pipecat/frames/frames.py
@@ -747,15 +747,8 @@ class TTSSpeakFrame(DataFrame):
     Parameters:
         text: The text to be spoken.
         append_to_context: Whether the spoken text should be appended to the LLM
-            context. Defaults to True.
-
-            .. deprecated:: 1.4.0
-                The default changed from ``None`` to ``True``, and ``None`` is
-                no longer a supported value (it is coerced to ``True`` with a
-                warning). The previous ``None`` behavior was situation-dependent
-                which made it hard to reason about. ``True`` records spoken text
-                in the context by default; pass ``False`` to keep it out.
-                ``None`` support will be removed in a future release.
+            context. Defaults to True. (Note that, as of version 1.4.0, ``None`` —
+            the previous default — is no longer a supported value.)
     """
 
     text: str
--- a/src/pipecat/frames/frames.py
+++ b/src/pipecat/frames/frames.py
@@ -14,6 +14,7 @@ and LLM processing.
 from __future__ import annotations
 
 import time
+import warnings
 from collections.abc import Awaitable, Callable, Mapping, Sequence
 from dataclasses import dataclass, field
 from typing import (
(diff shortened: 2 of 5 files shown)


src/pipecat/frames/frames.py
```diff
@@ -745,11 +746,36 @@ class TTSSpeakFrame(DataFrame):
 
     Parameters:
         text: The text to be spoken.
-        append_to_context: Whether to append the text to the context.
+        append_to_context: Whether the spoken text should be appended to the LLM
+            context. Defaults to True.
+
+            .. deprecated:: 1.4.0
```
markbackman: I'm not sure of the right way to list this, but deprecated feels incorrect. At first glance, it reads as if `append_to_context` is deprecated, which it isn't. It's just the value of `None`. I don't think we have a precedent for this, so we'll have to figure it out.

Maybe just a docstrings note about this?

filipi87: +1, I think maybe just the docstring explaining the behavior will be enough. 👍

kompfner: Hmm, you're right, this deprecated message at a glance could easily be misconstrued as deprecating the field. I'll rework it.

Let's keep the deprecation changelog entry, though.

changelog/4642.changed.md
```diff
@@ -0,0 +1,10 @@
+- ⚠️ Changed the default of `TTSSpeakFrame.append_to_context` from `None` to
+  `True`. The old `None` behavior was situation-dependent and hard to reason
+  about: the spoken text always reached the assistant aggregator's buffer, but
+  whether it was committed to the LLM context depended on what surrounded the
+  frame — committed when the frame was inside an assistant response or
+  immediately followed by one, but silently discarded when it was standalone and
+  followed by a user turn (the interruption cleared the buffer before anything
+  flushed it). `True` is a predictable default: programmatically-spoken text is
```
filipi87, line 8: OK, I finally see what's happening.

Looking at the code, I can see that `_trigger_assistant_turn_stopped` is always invoked when an interruption occurs.

However, because we're not in the middle of an LLM response, `self._assistant_turn_start_timestamp is ""`. In that case, `_trigger_assistant_turn_stopped` returns early, and the text gets discarded.

I think changing the default value of `TTSSpeakFrame` to `True` makes a lot of sense and makes the behavior more predictable.

Thank you for catching this, Paul.

filipi87 (approved):
I believe we just need to address the deprecation issue, as Mark mentioned. Other than that, it looks good.

The behavior is much easier to understand now. 🚀

changelog/4642.changed.md
kompfner: Yeah, I ended up having to write a table of scenarios for myself to finally understand exactly when TTSSpeakFrame text would land in context, and then realized...there's no way we can ask Pipecat users to do this 😛 

Yes, the never-set `_assistant_turn_start_timestamp` was the key discovery; it was responsible for making things so messy.

markbackman (approved):
LGTM (pending the deprecation removal in the docstring). Totally fine to keep in the changelog.

PR #4645, opened 2026-06-05
fix: close coroutines cancelled before task start

### LINKED ISSUE #4644: RuntimeWarning: coroutine 'TurnAnalyzerUserTurnStopStrategy._timeout_handler' was never awaited when the timeout task is cancelled before it runs
**Environment:** pipecat 1.3.0, Python 3.12

## Summary

`TaskManager.create_task` wraps the passed coroutine in an inner `run_coroutine()` that only awaits it once that wrapper actually runs. If the returned task is cancelled **before** `run_coroutine` reaches its `await coroutine`, the wrapped coroutine is dropped un-awaited, producing:

```
RuntimeWarning: coroutine 'TurnAnalyzerUserTurnStopStrategy._timeout_handler' was never awaited
```

`TurnAnalyzerUserTurnStopStrategy._handle_vad_user_stopped_speaking` creates `_timeout_task` (around line 195). When a second `UserTurnStopStrategy` (e.g. a max-turn-duration cap) force-completes the turn, `reset()` / `cleanup()` cancel that task immediately (around line 78) — hitting this race. The turn itself completes correctly; the warning is cosmetic (stderr only), but noisy when combining `TurnAnalyzerUserTurnStopStrategy` with another stop strategy that can force-complete the turn.

## Reproduction

Prints `reproduced: True` on pipecat 1.3.0 / Python 3.12:

```python
import asyncio
import gc
import warnings

from pipecat.turns.user_stop.turn_analyzer_user_turn_stop_strategy import (
    TurnAnalyzerUserTurnStopStrategy,
)


class _StubTurnAnalyzer:
    """Only needs to exist - the strategy stores it; this repro never calls its methods."""


async def main():
    tm = TaskManager()

    strat = TurnAnalyzerUserTurnStopStrategy(turn_analyzer=_StubTurnAnalyzer())
    await strat.setup(tm)

    # _handle_vad_user_stopped_speaking creates the timeout task (line ~195)...
    task = tm.create_task(strat._timeout_handler(0.1), f"{strat}::_timeout_handler")
    # ...then the cap's force-stop runs reset() -> cancels it BEFORE run_coroutine reaches `await`.
    task.cancel()  # synchronous: no loop iteration has run run_coroutine yet
    try:
        await task
        pass


with warnings.catch_warnings(record=True) as caught:
    asyncio.run(main())
    gc.collect()

hits = [w for w in caught if "_timeout_handler" in str(w.message) and "never awaited" in str(w.message)]
print(f"reproduced: {bool(hits)}")
```

## Suggested fix

In `TaskManager.create_task` / `run_coroutine`, close the wrapped coroutine when the task is cancelled before its first run; or guard the strategy's cancel path so the timeout coroutine is always awaited/closed.

he-yufeng:
## Summary
- close the wrapped coroutine when a managed task is cancelled before its wrapper starts running
- keep existing cancellation behavior once the wrapper has reached the original coroutine
- add a regression test that cancels the task immediately and checks no `coroutine was never awaited` warning is emitted

Fixes #4644

## To verify
- `python -m pytest tests\test_task_manager.py -q`
- `python -m ruff check src\pipecat\utils\asyncio\task_manager.py tests\test_task_manager.py`
- `python -m ruff format --check src\pipecat\utils\asyncio\task_manager.py tests\test_task_manager.py`
- `python -m py_compile src\pipecat\utils\asyncio\task_manager.py tests\test_task_manager.py`
- `git diff --check`

### DIFF
--- a/src/pipecat/utils/asyncio/task_manager.py
+++ b/src/pipecat/utils/asyncio/task_manager.py
@@ -154,8 +154,11 @@ class TaskManager(BaseTaskManager):
         Raises:
             Exception: If the task manager is not properly set up.
         """
+        coroutine_started = False
 
         async def run_coroutine():
+            nonlocal coroutine_started
+            coroutine_started = True
             try:
                 return await coroutine
             except asyncio.CancelledError:
@@ -167,11 +170,16 @@ class TaskManager(BaseTaskManager):
                 last = tb[-1]
                 logger.error(f"{name} unexpected exception ({last.filename}:{last.lineno}): {e}")
 
+        def close_unstarted_coroutine(task: asyncio.Task):
+            if task.cancelled() and not coroutine_started:
+                coroutine.close()
+
         if not self._params:
             raise Exception("TaskManager is not setup: unable to get event loop")
 
         task = self._params.loop.create_task(run_coroutine())
         task.set_name(name)
+        task.add_done_callback(close_unstarted_coroutine)
         task.add_done_callback(self._task_done_handler)
         self._add_task(TaskData(task=task))
         logger.trace(f"{name}: task created")
--- a/tests/test_task_manager.py
+++ b/tests/test_task_manager.py
@@ -1,0 +1,29 @@
+import asyncio
+import gc
+import warnings
+
+import pytest
+
+from pipecat.utils.asyncio.task_manager import TaskManager, TaskManagerParams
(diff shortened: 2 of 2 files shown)


aconchillo:
Closing. Fixed in #4649

PR #4648, opened 2026-06-05
Refactoring Tavus video service and transport to work with app messages

filipi87:
## Summary

  - Changed `TavusTransport` to deliver bot audio via the Tavus `conversation.echo` app message API instead of a WebRTC audio track, enabling faster-than-realtime audio delivery
  - Audio is base64-encoded and sent in 100ms chunks with a per-utterance `inference_id` to group chunks belonging to the same bot turn
  - End-of-utterance is signaled using `TTSStoppedFrame` (injected on `BotStoppedSpeakingFrame`), with a VAD fallback timeout for reliability

  ## Breaking Changes 
  
  - Default `persona_id` changed from `"pipecat-stream"` to `"pipecat0"`


scripts/daily/mock_tavus_server.py
filipi87, line 1: Emulates the behavior of the Tavus server. Receives audio faster than real time, buffers it, and plays it back at the correct sample rate.

src/pipecat/services/tavus/video.py
```diff
@@ -66,7 +71,7 @@ def __init__(
         *,
         api_key: str,
         replica_id: str,
-        persona_id: str = "pipecat-stream",
+        persona_id: str = "pipecat0",
```
filipi87, line 71: This was our old `persona_id`, which was configured to send the audio through `app-message`.

filipi87:
The `main` differences from the last time we tried using `app-message`:
- We paced audio delivery. After sending 100 ms of audio, we waited 100 ms before sending the next message.
- We capped the send rate at 20 messages per second.
- Each message was limited to 48,000 bytes.

In the new implementation, we have essentially removed all of these limitations.

src/pipecat/transports/tavus/transport.py
```diff
@@ -178,7 +183,7 @@ def __init__(
         callbacks: TavusCallbacks,
         api_key: str,
         replica_id: str,
-        persona_id: str = "pipecat-stream",
+        persona_id: str = "pipecat0",
```
markbackman, line 192: This changes the `persona_id` default to `"pipecat0"`, but the `__init__` docstring just below ([L197-L199](https://github.com/pipecat-ai/pipecat/blob/83b49f5a1d92535195634ed3c97f574be1db33b5/src/pipecat/transports/tavus/transport.py#L196-L200)) still documents the old default:

> persona_id: ID of the Tavus persona. Defaults to "pipecat-stream", which signals Tavus to use the TTS voice of the Pipecat bot instead of a Tavus persona voice.

Please update it to reflect the new `"pipecat0"` default so the public docs match the code. Flagged for [AGENTS.md docstring conventions](https://github.com/pipecat-ai/pipecat/blob/83b49f5a1d92535195634ed3c97f574be1db33b5/AGENTS.md) — the `__init__` `Args:` section must accurately document the parameter.

src/pipecat/services/tavus/video.py
```diff
@@ -66,7 +68,7 @@ def __init__(
...
```
markbackman, line 71: This changes the `persona_id` default to `"pipecat0"`, but the `__init__` docstring just below ([L81](https://github.com/pipecat-ai/pipecat/blob/83b49f5a1d92535195634ed3c97f574be1db33b5/src/pipecat/services/tavus/video.py#L80-L82)) still documents the old default:

> persona_id: ID of the Tavus persona. Defaults to "pipecat-stream" for Pipecat TTS voice.

Please update it to reflect the new `"pipecat0"` default so the public docs match the code. Flagged for [AGENTS.md docstring conventions](https://github.com/pipecat-ai/pipecat/blob/83b49f5a1d92535195634ed3c97f574be1db33b5/AGENTS.md) — the `__init__` `Args:` section must accurately document the parameter.

markbackman (commented):
## Code review

Reviewed for bugs and CLAUDE.md/AGENTS.md compliance. The audio-delivery refactor (chunking, `inference_id` grouping, done-signaling, interruption cancel/restart) checks out — no logic or security issues found. Two small doc nits below: the `persona_id` docstrings still document the old default that this PR changes.

src/pipecat/services/tavus/video.py
markbackman: Can we add this to the changelog, citing the change to align with Tavus's API? It's probably best as a standalone entry.

Same for the `TavusTransport`.

markbackman (approved):
Looks good! Just a few small comments. Nice, clean PR 👏 

I had Claude review and comment too. Just small findings from it.

src/pipecat/transports/tavus/transport.py
```diff
@@ -643,29 +761,21 @@ async def process_frame(self, frame: Frame, direction: FrameDirection):
             await self._handle_interruptions()
 
     async def _handle_interruptions(self):
-        """Handle interruption events by sending interrupt message."""
+        """Handle interruption events by discarding buffered audio and sending interrupt message."""
+        await self._client.cancel_send_task()
         await self._client.send_interrupt_message()
+        await self._client.start_send_task()
 
     async def write_audio_frame(self, frame: OutputAudioRawFrame) -> bool:
```
filipi87, line 826: Hi, @markbackman, regarding the transport, I believe we will need to add an optin flag for this feature because it would cause issues with `AudioBufferProcessor`. 

The user and bot audio could become out of sync, and bot speech that was interrupted before it was actually played to the user could still end up in the recording.

This doesn't happen with `TavusVideoService`. The audio we send is faster than real time, but the audio we receive back (along with the video) is in real time. As a result, any `AudioBufferProcessor` after this point would work correctly.

markbackman: Ok, so this applies to the `TavusTransport` only, right? If the `AudioBufferProcessor` is placed after the output transport, is there an issue? Or rather, in which situations would the `AudioBufferProcessor` not work correctly.

filipi87: Exactly, only for the `TavusTransport`, if we add an `AudioBufferProcessor` after it.

Basically, you should only be able to send audio faster than real time when using `TavusTransport` if you're not using `AudioBufferProcessor`. Otherwise, it could cause synchronization issues when the bot is interrupted.

markbackman: Ok, makes sense.

Tavus offers recording, so it's probably better to record from the Tavus side in this case. This may actually be the situation that most go with since they'll want to record audio and video, synchronized. The AudioBufferProcessor is only an audio recorder.

filipi87: Ok, I added an opt in realtime pacing flag to the Tavus transport: `audio_out_faster_than_realtime`, which by default is False to keep backwards compatibility in case someone is using `AudioBufferProcessor`.

markbackman: LGTM!

src/pipecat/services/tavus/video.py
filipi87: Done!

src/pipecat/transports/tavus/transport.py
filipi87: Done!

src/pipecat/services/tavus/video.py
filipi87: Done!

PR #4649, opened 2026-06-05
fix(asyncio): close un-started coroutine when task cancelled before run

### LINKED ISSUE #4339: RuntimeWarning: coroutine 'LLMService._run_function_call' was never awaited when a function call is cancelled by user-turn-start interruption race (distinct from #3588)
### pipecat version

1.0.0 (from pypi)

### Python version

3.14.4

### Operating System

macOS (Apple Silicon)

### Issue description

A `RuntimeWarning` fires — and more importantly the Gemini Live session stalls for ~10 seconds — when a function call dispatched by `GeminiLiveLLMService` is cancelled by an interruption broadcast originating from `TranscriptionUserTurnStartStrategy`.

This is **distinct from #3588**:
- Different coroutine. #3588 reports `_run_function_call.<locals>.timeout_handler` being unawaited; this report is the **outer** `_run_function_call` coroutine itself being unawaited.
- Different trigger. #3588 is triggered by a function handler pushing `EndTaskFrame` / `EndFrame`. This one is triggered by the normal turn-interruption race: when input transcription arrives and `LLMUserAggregator` broadcasts an interruption frame within the same millisecond the model emits a tool call.

#3588's fix (a `finally: cancel_task(timeout_task)` block) targets the inner timeout handler; it does not address this outer-coroutine leak on interruption cancellation.

The warning is a symptom. The substantive problem is that Gemini Live never receives the function-call result, and its session sits idle producing **zero completion tokens for ~10 seconds** until the next user utterance triggers a server-side VAD interruption and unsticks it. For an agent that performs many tool calls in a conversation, this is a meaningful UX regression, not cosmetic.

### Reproduction steps

1. Use `examples/realtime/realtime-gemini-live-function-calling.py` with `GeminiLiveLLMService` (not Vertex), Daily transport in direct mode (`-t daily -d`).
2. Speak a question that fires a registered function (e.g. "What's the weather in Tokyo?"). The very first turn of a fresh session reliably triggers the race; later turns hit it occasionally.
3. Observe:
   - `_handle_msg_input_transcription` logs the user text
   - `_on_user_turn_started` fires `broadcast_interruption`
   - `_run_function_call` starts
   - `_cancel_function_call` fires within ~1 ms
   - `RuntimeWarning: coroutine 'LLMService._run_function_call' was never awaited` is emitted
   - The next 10+ seconds show `TTFB: 10.421s, prompt tokens: 0, completion tokens: 0` — Gemini produces nothing because it's still waiting for the cancelled tool result
   - Only the next user utterance's Gemini-VAD interruption signal unblocks the session

### Expected behavior

The function call should either:
- Not be cancelled by a same-turn interruption broadcast (the tool was requested by the current turn; interrupting the current turn shouldn't kill it), or
- If cancellation is required, its outer coroutine should be awaited/cancelled cleanly and Gemini should be notified so it doesn't stall waiting for an undelivered result.

No `RuntimeWarning` in either case.

### Actual behavior

- `RuntimeWarning: coroutine 'LLMService._run_function_call' was never awaited`
- Session stalls for ~10 seconds (0 completion tokens, `TTFB: 10.421s`)
- Stalled state only clears when the next user utterance triggers a fresh VAD interruption

### Logs (trimmed to the critical window)

```
18:15:28.682 [Transcription:user] [Oh, what's the weather in Tokyo?]
18:15:28.683 _on_user_turn_started (TranscriptionUserTurnStartStrategy#0)
18:15:28.683 broadcast_interruption
18:15:28.687 _on_user_turn_stopped (TurnAnalyzerUserTurnStopStrategy#0)
18:15:28.689 Cancelling function call [get_current_weather:fc_13054767619604724024]...
18:15:28.690 Function call [get_current_weather:...] has been cancelled
/path/to/python3.14/asyncio/events.py:94: RuntimeWarning:
  coroutine 'LLMService._run_function_call' was never awaited
18:15:28.696 FunctionCallsStartedFrame: ['get_current_weather:...']
18:15:28.697 FunctionCallCancelFrame: [get_current_weather:...]
18:15:39.110 Gemini VAD: interrupted signal received
18:15:39.111 broadcast_interruption (GeminiLiveLLMService#0)
18:15:39.111 TTFB: 10.421s
18:15:39.112 prompt tokens: 0, completion tokens: 0
18:15:41.925 [Transcription:user] [Recommend a restaurant in Paris.]
```

### Potential direction (not a fix, just a hypothesis)

`LLMUserAggregator` broadcasts interruption on `_on_user_turn_started` without regard for whether the current turn's LLM response is emitting a tool call. An interruption that cancels a just-fired tool call from the same turn is arguably wrong — the tool was requested by this turn, not a stale prior one.

Options worth considering:
1. Suppress interruption-driven cancellation for function calls emitted within the same turn window as the interruption.
2. Still cancel, but send a synthetic error result back to Gemini so it doesn't stall waiting.
3. Ensure the outer `_run_function_call` coroutine is properly awaited on the cancellation path so the warning at minimum goes away.

### LINKED ISSUE #4644: RuntimeWarning: coroutine 'TurnAnalyzerUserTurnStopStrategy._timeout_handler' was never awaited when the timeout task is cancelled before it runs

scttbnsn:
## Summary

Fixes #4644 and fixes #4339.

`TaskManager.create_task()` wraps every coroutine in an inner `run_coroutine()` that only awaits the passed coroutine once that wrapper actually runs. If the returned task is cancelled *before* the event loop steps `run_coroutine` — a pre-start cancellation — the wrapper never reaches its `await coroutine`, so the inner coroutine is dropped un-awaited and CPython emits `RuntimeWarning: coroutine '...' was never awaited` when it is garbage-collected. The turn/cancellation logic itself is already correct; only the warning is spurious.

This shows up on two real paths:

- Combining `TurnAnalyzerUserTurnStopStrategy` with another stop strategy that force-completes the turn cancels the analyzer's timeout task before it runs (#4644).
- A function call cancelled by a user-turn-start interruption race leaves `LLMService._run_function_call` un-awaited (#4339).

The fix handles this in one place: a done callback closes the inner coroutine while it is still `CORO_CREATED` (the only state it can be in if the wrapper never ran). With the warning handled centrally, the local `await asyncio.sleep(0)` in `_run_function_call` — which existed only to let the timeout task start before it could be cancelled — is no longer needed and is removed.

## What changed

- **`utils/asyncio/task_manager.py`**: `create_task()` adds a `close_unawaited_coroutine` done callback that closes the inner coroutine when it is still `CORO_CREATED`. It's guarded by `inspect.iscoroutine()` so a non-native awaitable (which the `Coroutine` annotation technically permits, and which `asyncio.create_task` accepts) can't make `getcoroutinestate()` raise `AttributeError` inside the callback — consistent with the existing `getattr(coroutine, "cr_code", None)` probe in `base_object.py`.
- **`services/llm_service.py`**: removed the `await asyncio.sleep(0)` (and its comment) in `_run_function_call`. It only existed to dodge the never-awaited warning for the timeout task; that case is now covered by the change above.
- **`tests/test_task_manager.py`**: new file. The regression test asserts the inner coroutine ends in `CORO_CLOSED` after a pre-start cancel. This is deterministic and version-independent — `warnings.catch_warnings` can't intercept the GC-emitted warning from inside asyncio's managed context, so asserting on coroutine state is what actually distinguishes fixed from unfixed. Also covers cancel-after-start (cancellation still propagates into a running coroutine's cleanup) and normal completion.
- **`changelog/4644.fixed.md`**: changelog fragment.

## Verification

```
ruff format --check   3 files already formatted
ruff check            All checks passed!
git diff --check      clean
pytest                48 passed
                      test_task_manager.py, test_llm_service.py,
                      test_app_resources.py, test_wake_phrase_user_turn_start_strategy.py
```

The regression test was checked against the pre-fix source (parent commit): without the fix the coroutine stays `CORO_CREATED` and the test fails; with it the coroutine transitions to `CORO_CLOSED` and the test passes. So it's a real guard, not a tautology.


aconchillo (approved):

aconchillo:
Good one! Thank you @scttbnsn ! LGTM!

PR #4650, opened 2026-06-05
livekit: aclose owned audio/video streams + cancel producers on track unsubscribe

Maurotb:
## Summary

- `LiveKitTransportClient._async_on_track_subscribed` creates an owned `rtc.AudioStream(track)` + a `_process_audio_stream` task that feeds the shared `_audio_queue`, but only stored the track. The matching `_async_on_track_unsubscribed` only fired the user callback — it never closed the stream or cancelled the task.
- Per livekit-rtc `audio_stream.py`, an owned `AudioStream._run` loops over the FFI queue forever and exits only on a native `eos` event emitted by `aclose()` → `_ffi_handle.dispose()`. So a track unsubscribe (common when a web client republishes the mic on mute/unmute or text↔voice toggle) leaks the prior stream + task. After N republishes for the same participant, N concurrent producers push frames into the shared `_audio_queue`; downstream STT receives interleaved audio and stops producing transcripts. Same shape on video.

## Fix

Store `(stream, task)` per `participant.sid` in new `_audio_streams` / `_video_streams` dicts on subscribe; on unsubscribe pop the entry, `aclose()` the stream (bounded by `wait_for(timeout=2.0)` so a missing `eos` can't wedge the handler), then `task.cancel()` as a backstop. Re-subscribing the same sid (mic republish without explicit unsubscribe) also closes the prior stream first, so two producers never run for the same participant.

## Test plan

- [x] `uv run pytest tests/test_livekit_transport.py -v` — 7 passed (5 new + 2 pre-existing video-leak tests)
- [x] `uv run ruff check / format` clean
- [x] **End-to-end verified on a real web call**: applied this patch into a downstream worker's `.venv`, removed the local workaround, ran a bot with 5 voice↔text toggles. Result: 5 subscribe + 5 unsubscribe events, 0 `aclose failed` warnings, STT kept transcribing correctly after every toggle and the bot responded normally. Without the fix the same scenario produces N alive producers interleaving audio → STT goes silent → bot deaf.

### DIFF
--- a/src/pipecat/transports/livekit/transport.py
+++ b/src/pipecat/transports/livekit/transport.py
@@ -169,8 +169,14 @@ class LiveKitTransportClient:
         self._audio_track: rtc.LocalAudioTrack | None = None
         self._audio_tracks = {}
         self._audio_queue = asyncio.Queue()
+        # Per-participant ``(AudioStream, Task)`` so unsubscribe can close
+        # the owned native stream and cancel its producer task instead of
+        # leaking both on every track republish.
+        self._audio_streams: dict[str, tuple[rtc.AudioStream, asyncio.Task]] = {}
         self._video_tracks = {}
         self._video_queue = asyncio.Queue()
+        # Symmetric registry for video streams.
+        self._video_streams: dict[str, tuple[rtc.VideoStream, asyncio.Task]] = {}
         self._other_participant_has_joined = False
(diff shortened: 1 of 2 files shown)


aconchillo (approved):

aconchillo:
LGTM

PR #4651, opened 2026-06-05
Fix websockets dependency inadvertently becoming non-optional

kompfner:
Since `WebsocketLLMService` was introduced (#4211), core `llm_service.py` imported the optional `websockets` package unconditionally. So a missing `websockets` broke importing every LLM service with a raw `ImportError` — short-circuiting the `try/except` guard each service wraps its optional imports in to point users at the right `pipecat-ai[...]` extra (e.g. `[openai]`).

`WebsocketLLMService` is only used by `OpenAIResponsesLLMService`, whose install path already pulls in `websockets`. This PR:

- Moves `WebsocketLLMService` and `WebsocketReconnectedError` to a new `pipecat.services.websocket_llm_service` module that owns the dependency.
- Imports them in the OpenAI Responses service inside its existing `try/except`, so the missing-dependency message is the one users see.
- Re-exports both symbols from `llm_service.py` lazily via module `__getattr__`, keeping the released `from pipecat.services.llm_service import WebsocketLLMService` import path working without re-introducing the top-level `websockets` dependency.

## Testing

- With `websockets` blocked, `import pipecat.services.llm_service` now succeeds; importing the OpenAI Responses service logs the `pip install pipecat-ai[openai]` message and raises `ImportError`.
- With `websockets` present, lazy access of the moved symbols and the OpenAI service work unchanged.
- `tests/test_llm_service.py`, `tests/test_openai_responses_websocket.py`, `tests/test_openai_llm_timeout.py` — 53 passed.

### DIFF
--- a/changelog/4651.fixed.md
+++ b/changelog/4651.fixed.md
@@ -1,0 +1,1 @@
+- Fixed importing any LLM service failing with a raw `ImportError` when the optional `websockets` package wasn't installed. Core `llm_service.py` had begun importing `websockets` unconditionally, short-circuiting the guard that points users at the right `pipecat-ai[...]` extra (e.g. `[openai]`); the `websockets` dependency is now optional again.
--- a/src/pipecat/services/llm_service.py
+++ b/src/pipecat/services/llm_service.py
@@ -25,8 +25,6 @@ from typing import (
 
 from loguru import logger
 from typing_extensions import TypeVar
-from websockets.exceptions import ConnectionClosed
-from websockets.protocol import State
 
 from pipecat.adapters.base_llm_adapter import BaseLLMAdapter
 from pipecat.adapters.schemas.direct_function import DirectFunction, DirectFunctionWrapper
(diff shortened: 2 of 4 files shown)


markbackman (approved):
LGTM

filipi87 (approved):
Makes sense. 👍

kompfner:
PR appears stuck and unmergeable, though the GitHub status page remains all green. Going to close and re-open the PR.

markbackman:
@kompfner FYI, this is now on main:
https://github.com/pipecat-ai/pipecat/pull/4658

Perhaps this changes the approach, given `websockets` is a core dependency.

kompfner:
> @kompfner FYI, this is now on main: #4658
> 
> Perhaps this changes the approach, given `websockets` is a core dependency.

Oh wow great timing. Yes, the above change renders this PR unnecessary. If we're good with making websockets a core dependency, that's definitely preferable.

PR #4653, opened 2026-06-06
fix(tts): reuse the original LLMFullResponseEndFrame per context

aconchillo:
## Problem

When `push_text_frames` is `False`, `TTSService` minted a **new** `LLMFullResponseEndFrame` (with a new `id`) at the end of an audio context (`_maybe_reset_word_timestamps`), instead of re-pushing the one it received in `process_frame`.

`RTVIObserver` dedupes frames by `frame.id` and emits `bot-llm-stopped` on every `LLMFullResponseEndFrame`. It had already seen the LLM's original end frame upstream (at the LLM's output), so the TTS's brand-new frame slipped past the dedup and produced a **second `bot-llm-stopped` per LLM response**. Any RTVI client using a PTS-correcting TTS (e.g. Cartesia) saw the bot's response duplicated.

## Fix

Hold the `LLMFullResponseEndFrame` received in `process_frame`, keyed by the turn's `context_id` (created on `LLMFullResponseStartFrame`), and re-push **that same frame** (with the corrected PTS) at end of context. Preserving the `id` lets the observer dedup correctly, so there is one `bot-llm-stopped` per response. The held frames are cleared on interruption so an interrupted response's end frame cannot leak into the next turn.

## Testing

Reproduced with a Cartesia voice bot driven over RTVI: before the fix the client received two identical `bot-llm-stopped` (and duplicated LLM text) per response; after the fix, exactly one per response. Verified an interruption scenario still passes.

### DIFF
--- a/changelog/4653.fixed.md
+++ b/changelog/4653.fixed.md
@@ -1,0 +1,1 @@
+- Fixed `TTSService` emitting a second `LLMFullResponseEndFrame` (with a new id) at the end of an audio context when `push_text_frames` is `False`, which caused `RTVIObserver` to send a duplicate `bot-llm-stopped` message per LLM response. The original end frame received in `process_frame` is now held per `context_id` and re-pushed, preserving its id.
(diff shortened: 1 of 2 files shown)


aconchillo:
There's an alternative to the fix which is in `RTVIObserver`. As the description says we currently only use `frame.id` to dedup, but maybe we should add some logic there. The good thing about this fix is that you can follow the frame, by creating a new copy we kind of lose track of it.

filipi87 (approved):
Nice catch. Not sure how we have never caught this one before.
LGTM! 🚀

PR #4654, opened 2026-06-06
Auto-register direct functions advertised in an LLMContext

markbackman:
## Summary

Direct functions already bundle their handler and schema in a single `def`, but developers still had to mention each one **three times**: the function definition, a `register_direct_function()` call, and a `ToolsSchema` entry. This PR collapses that to a single mention while keeping every advanced case available through the existing explicit APIs.

**Before:**
```python
async def get_current_weather(params, location: str, format: str):
    """Get the current weather. ..."""
    await params.result_callback(...)

llm.register_direct_function(get_current_weather)
llm.register_direct_function(get_restaurant_recommendation)
tools = ToolsSchema(standard_tools=[get_current_weather, get_restaurant_recommendation])
context = LLMContext(tools=tools)
```

**After:**
```python
async def get_current_weather(params, location: str, format: str):
    """Get the current weather. ..."""
    await params.result_callback(...)

context = LLMContext(tools=[get_current_weather, get_restaurant_recommendation])
```

## What changed

Two cooperating, independent pieces:

- **Auto-registration** — `LLMService` registers direct-function handlers advertised in the context (via the `LLMContextFrame` seam in the base `process_frame`, so it covers all providers). Already-registered names are skipped, so explicit registration still wins and repeated context frames don't re-register.
- **`LLMContext` accepts a bare list** — `tools=` and `set_tools()` take a plain list of direct functions and/or `FunctionSchema` objects, normalized to a `ToolsSchema` internally. `ToolsSchema` now retains the original callables (new `direct_functions` property) so handlers are recoverable.
- **`@direct_function` decorator** — optional, metadata-only; attaches `cancel_on_interruption` / `timeout` for the override case (functions stay at module level). `@tool` is re-expressed on top of it, so the worker and plain-pipeline paths share one annotation.
- **Typing** — `DirectFunction` is redefined as a permissive `Callable[..., Awaitable[Any]]` alias so concrete function signatures type-check when passed to `register_direct_function`, `ToolsSchema`, and `LLMContext`. The real contract (async, first param `params`) is still enforced at runtime by `DirectFunctionWrapper.validate_function`.

Auto-registration and `ToolsSchema` creation are **independent axes** — you can mix them freely:

| Schema | Registration | When |
|--------|-------------|------|
| bare list | auto | the common case |
| explicit `ToolsSchema` | auto | you need `custom_tools` (provider-native tools); your direct functions still auto-register |
| explicit `ToolsSchema` | manual | hand-written `FunctionSchema` + separate handler |

## Backward compatibility

Purely additive. The verbose `FunctionSchema` flow and explicit `register_direct_function` flow are unchanged. Advanced cases keep their explicit escape hatches: provider-native `custom_tools`, advertise-only `FunctionSchema` entries, catch-all `register_function(None, ...)`, registered-but-not-advertised handlers, and mid-conversation `set_tools()` / `LLMSetToolsFrame` swaps.

## Test plan

- New `tests/test_auto_register_tools.py` (16 tests): defaults, decorator overrides, no-clobber of explicit registration, advertise-only `FunctionSchema` not registered, idempotency across repeated contexts, list acceptance in `LLMContext` / `set_tools`, and `@tool` annotation sharing.
- 244 tests pass across the affected suites (direct functions, LLM service/context, workers, UI, adapters).
- `basedpyright` clean on the changed sources and the updated example; `ruff check` / `ruff format --check` clean.
- Updated `examples/function-calling/function-calling-direct.py` to the one-liner API.


examples/function-calling/function-calling-direct.py
```diff
@@ -92,18 +91,13 @@ async def run_bot(transport: BaseTransport, runner_args: RunnerArguments):
...
     @llm.event_handler("on_function_calls_started")
     async def on_function_calls_started(service, function_calls):
         await tts.queue_frame(TTSSpeakFrame("Let me check on that."))
 
-    tools = ToolsSchema(standard_tools=[get_current_weather, get_restaurant_recommendation])
-
-    context = LLMContext(tools=tools)
+    # Direct functions listed in the context are registered with the LLM
+    # automatically — no separate register_direct_function() call needed.
+    context = LLMContext(tools=[get_current_weather, get_restaurant_recommendation])
```
filipi87, line 100: Looks so clean. Amazing. 👏🚀

filipi87 (approved):
This looks great @markbackman, so clean, nice work. 🚀

src/pipecat/services/llm_service.py
```diff
@@ -586,6 +587,13 @@ async def process_frame(self, frame: Frame, direction: FrameDirection):
         elif isinstance(frame, LLMContextSummaryRequestFrame):
             await self._handle_summary_request(frame)
 
+        if isinstance(frame, LLMContextFrame):
+            # Auto-register handlers for any direct functions advertised in the
+            # context. Runs before the subclass processes the context (subclasses
+            # call super().process_frame() first), so handlers are in place by the
+            # time the LLM can call them.
+            self._auto_register_direct_functions(frame.context)
```
kompfner, line 595: Hmm, I suspect from reading issues that folks may be relying on `unregister_function`/`unregister_direct_function` to remove functions without necessarily also using the recommended `LLMSetToolsFrame`. This mostly works today because we automatically stick a fallback "function not found" message in context when the LLM tries to invoke a nonexistent tool.

This change may make it so that the tool becomes a "zombie" and comes back to life, getting automatically re-registered. Right?

The fix would be to keep track of explicitly unregistered tools and not auto-re-register them, I think. In this code path, we could also take the opportunity to remind developers to use `LLMSetToolsFrame` rather than just `unregister_*` to remove tools (using `LLMSetToolsFrame` works better because it can save instances of the LLM trying a tool and discovering it's not there).

kompfner, line 595: Separate thought: is there a reason not to apply this auto-registering to non-direct functions, too? Was the thinking that we'd limit the new auto-registering behavior to the (less used) direct functions so that it wouldn't impact as many people in unexpected ways (like the way described in the above comment)?

src/pipecat/adapters/schemas/direct_function.py
```diff
@@ -34,22 +34,20 @@
...
+signature and docstring, so it can be used without an accompanying
+``FunctionSchema`` or provider-specific configuration.
 
-        Args:
-            params: Function call parameters from the LLM service.
-            **kwargs: Additional keyword arguments passed to the function.
-        """
-        ...
+Typed as a permissive ``Callable`` so that concrete function signatures (which a
+stricter ``Protocol`` form would reject) type-check when passed to public APIs.
```
kompfner, line 47: Ah yes, I think I see...the `Protocol` version we had before would actually not be happy with functions that *didn't* accept arbitrary `kwargs` as its last argument...right?

kompfner: Side note: I may try to finally tackle the CLAUDE.md change to prevent it from adding comments like this that are in conversation with the immediate design and iteration process rather than being focused on future readers.

src/pipecat/adapters/schemas/direct_function.py
```diff
@@ -289,3 +287,44 @@ async def invoke(self, args: Mapping[str, Any], params: "FunctionCallParams"):
...
+    attaches the options to the function and returns it unchanged — it does not
+    register anything, so decorated functions can stay at module level.
+
+    Can be used with or without arguments::
+
+        @direct_function
+        async def get_weather(params, location: str):
+            ...
+
+        @direct_function(cancel_on_interruption=False, timeout=60)
```
kompfner, line 307: We should make this consistent with Flows, which already has a [`flows_direct_function` decorator](https://github.com/pipecat-ai/pipecat-flows/blob/4a12fe0ffb20e9e260327fe263705434f7554ae3/src/pipecat_flows/types.py#L372).

It's almost the same, but looks like it differs in the timeout argument name. There, we called it `timeout_secs` for consistency with the `register_function` and `register_direct_function` methods.

src/pipecat/adapters/schemas/direct_function.py
```diff
@@ -289,3 +287,44 @@ async def invoke(self, args: Mapping[str, Any], params: "FunctionCallParams"):
...
+        timeout: Optional per-tool timeout in seconds. Defaults to None (uses
+            the LLM service default).
+
+    Returns:
+        The decorated function, with ``cancel_on_interruption`` and ``timeout``
+        attributes attached.
+    """
+
+    def decorator(fn):
+        fn.cancel_on_interruption = cancel_on_interruption
```
kompfner, line 324: Should we store these as prefixed with `_pipecat_` or something, to make it clear that they're just pipecat-specific "metadata" slapped onto the function?

src/pipecat/adapters/schemas/direct_function.py
markbackman: I opted to make this consistent with the `@tool` decorator in the LLMWorker:
```
def tool(fn=None, *, cancel_on_interruption=True, timeout=None):
```

This seems like the right call now that the `@tool` decorator is built from this new decorator (renamed to `@tool_options` in https://github.com/pipecat-ai/pipecat/pull/4671.

What to do about Flows? We could add `timeout` as a name there too...

Pulling Flows into Pipecat would help to keep things aligned. We should think about when we want to make that move.

src/pipecat/adapters/schemas/direct_function.py
markbackman: Yes, pyright was unhappy which motivated this chnage.

Re: CLAUDE.md, yes! I should be better about catching and removing these comments, too. But, improving CLAUDE.md will prevent it in the first place.

src/pipecat/services/llm_service.py
```diff
@@ -836,6 +844,35 @@ def register_direct_function(
...
+        tools = context.tools if context is not None else None
+        if tools is None or not is_given(tools):
+            return
+        for wrapper in tools.direct_functions:
+            if wrapper.name in self._functions:
+                continue
+            handler = wrapper.function
+            self.register_direct_function(
+                handler,
+                cancel_on_interruption=getattr(handler, "cancel_on_interruption", True),
```
kompfner, line 872: Hmm...I would expect the decorator-injected `cancel_on_interruption` and `timeout` values to take effect even if you registered your functions the "regular" way (via `register_function` or `register_direct_function`). This may trip people up.

Can we instead read the decorator-injected values inside the `register_` methods (with reasonable argument-provided-v-decorator-injected precedence logic + logging in place)?

changelog/4654.added.md
```diff
@@ -0,0 +1 @@
+- Direct functions listed in an `LLMContext`'s tools are now registered with the LLM service automatically — a separate `register_direct_function()` call is no longer required. `LLMContext(tools=...)` and `LLMContext.set_tools()` also accept a plain list of direct functions and/or `FunctionSchema` objects (in addition to a `ToolsSchema`). Added an optional `@direct_function(cancel_on_interruption=..., timeout=...)` decorator to override a direct function's call options. Explicit `register_direct_function()` registration still takes precedence, so advanced cases (advertising a subset of registered tools, advertise-only `FunctionSchema` entries, catch-all handlers, mid-conversation tool swaps) are unchanged.
```
kompfner, line 1: These should be separate entries in the changelog, I think. Some are totally orthogonal and can be used independently (and, as I argue in other comments, even more should be).

src/pipecat/workers/llm/tool_decorator.py
```diff
@@ -14,6 +16,10 @@ def tool(fn=None, *, cancel_on_interruption=True, timeout=None):
     registered with the LLM via ``register_direct_function`` and
     included in ``build_tools()``.
 
+    This is the worker-flavored variant of ``@direct_function``: it attaches the
+    same ``cancel_on_interruption`` / ``timeout`` call options and additionally
```
kompfner, line 20: See other comment re `timeout` v `timeout_secs`

src/pipecat/services/llm_service.py
markbackman: Oof. That's too easy to forget about. We should really discourage developers from calling functions.

I have a commit for this here:
https://github.com/pipecat-ai/pipecat/pull/4671/changes/acd68788995dd916085660063fb3abb570277c1d

I've added a debug statement when the `unregister` function is called. Hopefully removing use of `register` reduces the use of `unregister`.

Should this be a warning instead? Or, alternatively, should we deprecate and make those methods private in the future? Or, is there a legit case where you _need_ to use `unregister`?

markbackman: With the direct function, you have a combined `FunctionSchema` and handler, which makes registration and `ToolsSchema` generation just boilerplate.

For non-direct functions, the `FunctionSchema` and handler are joined by the `register_function()` call, so it's an essential piece. My focus was on providing a new, streamlined way to improve usage. I'm not sure there's a clean way to do this for non-direct functions.

src/pipecat/adapters/schemas/direct_function.py
markbackman: Yes! Good idea. This will avoid namespace collision. I'll have to change the existing `@tool` decorator, which has been released, but I think this is OK because it's internal API.

Commit: https://github.com/pipecat-ai/pipecat/pull/4671/changes/d905d8bdec1c5fb64142b2c925f1f8cc9e6dbc4b

src/pipecat/services/llm_service.py
markbackman: Good point. I didn't think about this. Commit: f7e1cbe43b39be4c1928eb9495413383c4e8073d.

changelog/4654.added.md
markbackman: Split into three. Commit: https://github.com/pipecat-ai/pipecat/pull/4671/changes/202b703052ae301feea9c75ea2496d4a4c9b7547

src/pipecat/services/llm_service.py
kompfner: The main benefit to unregistering I can think of is protecting against crummier models hallucinating function calls to already-removed tools. If the tool is unregistered, we get the nice "function not found" context message.

Actually...I wonder if our new auto-registering mechanism should also auto-unregister when it notices a tool has gone away...that would also add a nice symmetry to our code.

(So, instead of `register_context_direct_functions` it'd like `update_registered_functions_from_context`)

kompfner: Ah you're totally right, auto-registering doesn't even make sense for non-direct function since the `register` step is essential for "linking" the tool handler to its metadata. It can't be skipped. I'm with you now.

src/pipecat/adapters/schemas/direct_function.py
kompfner: I personally lean towards aligning on the `_secs` suffix since:
- The arg name in the `register` function the new decorator is replacing(ish) has the `_secs` suffix, and is a precedent that has been around a lot longer than the brand-new `LLMWorker` decorator (maybe we should be thinking about updating *that* one instead)
- We use the `_secs` suffix in other public APIs: `start_secs`, `stop_secs`, `idle_timeout_secs`, maybe others?
- It's very clear!

src/pipecat/services/llm_service.py
kompfner: OK, wrapping my head around the big picture of how this system could work (and how it's mostly working with your implementation):

- Direct function registration/unregistration is now automatic (derived from context)
  - But, for back compat, _can_ still be manually unregistered (which adds it to the special "manually-unregistered list" which opts it out of the automatic system)
  - Manual registering/unregistering of direct functions should be deprecated and log warnings
- "Old-style" function registration/unregistration is still manual
- Any manually-unregistered functions (in the "manually-unregistered list") must be manually registered again to re-enable

src/pipecat/adapters/schemas/direct_function.py
kompfner: So many of these go through the cracks for me too

src/pipecat/services/llm_service.py
markbackman: Ok, two commit:
- Deprecate and warn on use of register/unregister direct functions: https://github.com/pipecat-ai/pipecat/pull/4671/changes/99ef4626b4bd407921cb2d4b257ee3ab7768e8a1
- Auto-unregister direct functions via LLMContext (i.e. LLMSetToolsFrame): https://github.com/pipecat-ai/pipecat/pull/4671/changes/88cf96ccce4b940421341874d7b0f1b85f30cea7

Regarding:
> Any manually-unregistered functions must be manually registered again to re-enable

I'm conflicted. I think we might not want to enforce this. The goal is for developers to use the primary pattern:
- Define functions and handlers via direct functions
- Provide tools to use via LLMContext
- Modify tools to use via LLMSetToolsFrame

If a developer is using manual registration and unregistration, I would imagine they're not mixing in LLMSetToolsFrame. If they are, that's fine because we handle it. They'll receive a warning when using the register and unregister direct functions, which would be enough of a nudge to push them in the right direction. What do you think about leaving it as is?

src/pipecat/adapters/schemas/direct_function.py
markbackman: Updated in this commit: https://github.com/pipecat-ai/pipecat/pull/4671/changes/e02d1b3910df407bd4063968c38e13189dc70f00

src/pipecat/workers/llm/tool_decorator.py
markbackman: Fixed in: https://github.com/pipecat-ai/pipecat/pull/4671/changes/e02d1b3910df407bd4063968c38e13189dc70f00. Closing this as we're talking about the same thing above.

src/pipecat/services/llm_service.py
kompfner: > I think we might not want to enforce this.

I don't think we'd be adding code to enforce this—my only goal with this suggestion is to _remove_ complexity from the code (the `LLMSetToolsFrame` handling in `LLMService`, which breaks our nice LLM-messages-frames-like pattern of "this is a pure aggregator concern"). The complexity was only added for what seems like an exceedingly rare edge case: the user's code is wired up to call `unregister` but, in the case that it needs a function again, doesn't call `register`.

It's just not worth the code complexity, IMO.

kompfner: Hmm...unless the `LLMSetToolsFrame`-handling complexity in `LLMService` is required anyway for *another* reason entirely, which is the case of (some) realtime services that must directly handle the `LLMSetToolsFrame`s because they can't just rely on the new updated set of tools being available on the context object on the next inference...

Is that the main reason the `LLMSetToolsFrame` handling (the `_apply_advertised_tool_set` path) exists?

markbackman: Ahh, ok! Yes, I think I misunderstood the comment. Removing this actually removes a lot of complexity. Let me work on that.

kompfner: Even so...I suspect there's quite a bit of logical complexity we can prune here. If we're OK with not handling the above highly unlikely edge case (which I think we should), I suspect we can reduce to just:

```python
if isinstance(frame, LLMContextFrame):
  # Sync registered handlers with the direct functions advertised in the context
  self.sync_registered_direct_functions(frame.context.tools) # previously _apply_advertised_tool_set
```

And then, in just the realtime services that support runtime tool changes (e.g. OpenAI Realtime):

```python
  self.sync_registered_direct_functions(frame.tools) # <-- add this call here
  await self._send_session_update()
```

We could then delete:
- `register_context_direct_functions` and its private helper (it gets superseded by `sync_registered_direct_functions` and its private helper)
- `LLMSetToolsFrame` handling in `LLMService`

There'd just be one unified code path for "syncing registered handlers with the direct functions advertised in the context" that would behave the same no matter whether it was triggered by an `LLMContextFrame` or a `LLMSetToolsFrame`.

kompfner:
Shoot, just noticed something: at some point during the refactors I think we lost the thread on the point of tracking functions that were "auto-registered". The important thing is to not auto-_re-register_ functions that were explicitly _unregistered_, not vice-versa.

We have reason to believe that some developers are relying on the assumption that explicitly unregistering (without also removing what's advertised, via `LLMSetToolsFrame`) is enough (and it is, today, though it's sub-optimal).

kompfner:
> Shoot, just noticed something: at some point during the refactors I think we lost the thread on the point of tracking functions that were "auto-registered". The important thing is to not auto-_re-register_ functions that were explicitly _unregistered_, not vice-versa.
> 
> We have reason to believe that some developers are relying on the assumption that explicitly unregistering (without also removing what's advertised, via `LLMSetToolsFrame`) is enough (and it is, today, though it's sub-optimal).

I'm already touching this code, so happy to tackle this.

PR #4655, opened 2026-06-07
Add behavioral eval framework (`pipecat.evals`) + `pipecat eval` commands

aconchillo:
## What this adds

`pipecat.evals` — a behavioral evaluation framework for Pipecat bots — usable both as a library and through the `pipecat` CLI.

### Scenarios (YAML)

A **scenario** is a YAML file describing a scripted conversation plus the semantic events expected back from the bot: transcriptions, LLM/TTS responses, and function calls, with optional latency budgets and natural-language criteria checked by a judge LLM. Scenarios run in **text or audio** mode — in audio mode the harness synthesizes the user's speech (TTS) and transcribes the bot's actual audio (local Whisper) — and they're reusable, so one scenario can cover many bots.

### Library API

The framework is a plain Python library. Load a scenario and run it against a bot in a few lines:

```python
from pipecat.evals.scenario import EvalScenario
from pipecat.evals.harness import EvalSession

scenario = EvalScenario.load("scenarios/simple_math.yaml")
result = await EvalSession.from_scenario(scenario, "ws://localhost:7860").run()
assert result.passed
```

`EvalSession.from_scenario(...)` builds the judge/voice/transcriber the scenario needs and returns a session; `.run()` returns a structured `EvalResult`. `pipecat.evals.suite` exposes `EvalManifest.load()` + `EvalSuite.run()` for the multi-bot path. The CLI is a thin wrapper over these.

### Two CLI commands

- **`pipecat eval run <scenario>... --bot-url ...`** — run one or more scenarios against an already-running bot, with a live dashboard (status, timing, pass/fail tally). The bot stays up between scenarios.
- **`pipecat eval suite <manifest>`** — a **manifest** maps each bot to the scenarios it runs; the suite spawns each bot, runs them concurrently, and prints a summary. The release evals (`scripts/release-evals/`) are now just a manifest plus this command.

Both are also reachable as `python -m pipecat.evals`.

### How a bot opts in

A bot exposes an **eval transport** (`-t eval`) — a local WebSocket server speaking RTVI. The harness connects as a thin RTVI client, drives the scripted turns, and asserts on the events the bot emits. The example bots add the `eval` transport so they can be evaluated directly.

### General additions used by the harness

Two non-eval-specific pieces landed here because the harness needs them, but they're useful on their own:

- **`PipelineFlushFrame`** — a control frame that drains the pipeline. Push it downstream and the pipeline worker bounces it back upstream so it round-trips through every processor, then sets its `event`; await that to know all in-flight frames have been processed (e.g. to let the pipeline settle after an interruption before injecting a new frame). It's an `UninterruptibleFrame`, so it survives an `InterruptionFrame`.
- **`bot-interrupted` RTVI server message** — emitted when the bot's in-flight output is cut off (a VAD-detected barge-in or a programmatic interrupt), so any RTVI client can drop whatever the bot was mid-saying.


pyproject.toml
```diff
@@ -34,6 +34,7 @@ dependencies = [
     "pyloudnorm~=0.2.0",
     "resampy~=0.4.3",
     "soxr~=1.0.0",
+    "typer>=0.12,<1",
```
markbackman: Is this needed? The `cli` extra already installs typer. If it is needed, we should match the typer version range, `>=0.20.0,<1`.

pyproject.toml
markbackman, line 143: Without the websocket install, the cli is not working. You need to add `"pipecat-ai[websockets-base]"` to the `cli` extra so that the `eval` command can run. This will also allow you to avoid lazy imports in places.

markbackman: You also need to add `pyyaml` to this list. Looks like `>=6` should do the trick.

src/pipecat/evals/harness.py
```diff
@@ -0,0 +1,1178 @@
...
+        self._wants_response: bool = any(
+            exp.event == "response" for turn in scenario.turns for exp in turn.expect
+        )
+        self._transcriber: Any = None
+        self._tts_audio: bytearray = bytearray()  # current spoken segment's audio
+        self._tts_sample_rate: int = 0
+
+    async def run(self) -> EvalResult:
+        """Connect, drive the scenario, and return the result."""
+        import websockets  # lazy: see note at the top of the module
```
markbackman: Remove after adding websocket to the `cli` extra

src/pipecat/evals/harness.py
```diff
@@ -0,0 +1,1178 @@
...
+                id=self._message_id(),
+                data={"t": EVAL_CANCEL_MESSAGE_TYPE, "d": {}},
+            )
+            await self._ws.send(message.model_dump_json())
+        except Exception:
+            pass
+
+    async def _reader_loop(self) -> None:
+        """Drain the WS, translate RTVI messages to friendly events, enqueue them."""
+        import websockets  # lazy: see note at the top of the module
```
markbackman: Remove

src/pipecat/evals/harness.py
```diff
@@ -0,0 +1,1178 @@
...
+    EVAL_BOT_AUDIO_TYPE,
+    EVAL_CANCEL_MESSAGE_TYPE,
+    EVAL_CONFIGURE_MESSAGE_TYPE,
+    EVAL_RESET_MESSAGE_TYPE,
+)
+
+# ``websockets`` is imported lazily inside the methods that use it. The
+# ``pipecat eval`` command imports this module while building the CLI, so it
+# gets pulled in on every ``pipecat`` invocation (even ``pipecat init``), not
+# only when an eval runs. Keeping the import lazy means the module stays
```
markbackman: You can drop the lazy import. There was a dependency already that required websocket:
```
  import …rtvi.models
     └─ runs rtvi/__init__.py        (package init, side effect of importing the submodule)
          └─ :24 imports rtvi.processor
               └─ :43 imports services.llm_service
                    └─ :28 imports websockets   ← hard requirement, unguarded
```

But, that goes away by the `cli` extra pulling in `pipecat-ai[websocket-base]`.

markbackman (commented):
Some quick comments on the CLI dependencies.

src/pipecat/evals/judge.py
```diff
@@ -0,0 +1,245 @@
...
+        if "no" in lowered and "yes" not in lowered:
+            return JudgeVerdict(verdict="no", reason="(unstructured no)", raw_response=response)
+        return JudgeVerdict(
+            verdict="no",
+            reason=f"could not parse judge response: {response!r}",
+            raw_response=response,
+        )
+
+
+def build_default_judge(judge_config: dict | None) -> EvalJudge | None:
```
markbackman: Not needed for the first version, but at some point, we'll probably want to make this extensible. Maybe you can just pass an eval judge factory that runs in your own code. Anyway, not needed for launch and not blocking.

src/pipecat/evals/transcribe.py
```diff
@@ -0,0 +1,246 @@
...
+
+    Args:
+        config: ``bot_audio`` mapping, or ``None`` for the Whisper default.
+
+    Returns:
+        A constructed ``STTService``.
+    """
+    config = config or {}
+
+    custom = config.get("factory")
```
markbackman: Oh! This is the pattern we should follow for bringing your own judge service.

src/pipecat/evals/harness.py
aconchillo: Fixed

src/pipecat/evals/harness.py
aconchillo: Fixed

pyproject.toml
aconchillo: Websocket is a core dependency. SO, fixed.

src/pipecat/evals/voice.py
```diff
@@ -0,0 +1,336 @@
...
+            raise ValueError(f"user_audio.factory must be a dotted path: {custom!r}")
+        factory = getattr(importlib.import_module(module_name), attr)
+        return factory(voice_cfg, sample_rate)
+
+    service = str(voice_cfg.get("service", "")).lower()
+    voice = str(voice_cfg.get("voice", ""))
+    if not service or not voice:
+        raise ValueError("user_audio config requires at least 'service' and 'voice'")
+
+    if service == "cartesia":
```
markbackman: Is there a local model we should offer as an option for testing, as an out of the box option? The options would be Kokoro and Piper. I don't have experience with either one to pick though.

pyproject.toml
aconchillo: removed

pyproject.toml
aconchillo: `pyyaml` should go into a core dependency since you can run evals without the CLI

src/pipecat/evals/harness.py
aconchillo: Fixed. websocket is a core dep now.

pyproject.toml
aconchillo: fixed.

src/pipecat/evals/voice.py
aconchillo: yes, we probably should do that. much cheaper. i'll try kokoro for this it should be good enough, specially since we are caching

src/pipecat/evals/transcribe.py
aconchillo: Added support for that to judge as well.

src/pipecat/evals/judge.py
aconchillo: Done!

src/pipecat/evals/voice.py
aconchillo: Added Kokoro support

markbackman (approved):
Amazing work! Ship it 🚢

changelog/4655.added.3.md
```diff
@@ -0,0 +1 @@
+- Added `PipelineFlushFrame`, a control frame for draining the pipeline. Push it downstream and the pipeline worker bounces it back upstream so it round-trips through every processor, then sets its `event`. Await that event to know all in-flight frames queued ahead of the probe have been processed (e.g. to let the pipeline settle after an interruption before injecting a new frame). It's an `UninterruptibleFrame`, so the probe survives an `InterruptionFrame` and still completes its round-trip.
```
markbackman, line 1: Minor—rename this as `changelog/4655.added.3.md`

aconchillo: Fixed!

aconchillo: Done

PR #4656, opened 2026-06-07
bus: serialise inbound bus frames via worker.queue_frame to fix multi-worker set_node race

Maurotb:
## Summary

In a multi-worker setup with a child `LLMWorker` bridged onto the bus, `_BusEdgeProcessor.on_bus_message` used to push received frames directly from the edge into the local pipeline. The worker's own outbound path goes via `worker.queue_frame` → `_push_queue` → `_process_push_queue` (drained FIFO) → `_pipeline.queue_frame`. The two paths converge at the same processors but enter from different async tasks, so frames in one stream can interleave with frames in the other.

Witnessed with `pipecat-flows`: `set_node` enqueues `[LLMMessagesUpdate, LLMSetTools]` (and optionally `LLMRunFrame`) on the child via `worker.queue_frames`. A concurrent user input on the main worker becomes an `LLMContextFrame` and is routed via the bus back to the child. The bus inbound, being pushed directly by the edge, can reach the `LLMService` before the new node's `LLMSetToolsFrame` has landed on the shared aggregators, and the LLM generates against the previous node's context.

## Fix

Route bus inbound through `self._task.queue_frame` instead of `self.push_frame`. Both paths now share the worker's single `_push_queue` (drained FIFO by `_process_push_queue`), so the "apply before run" ordering held within a single `queue_frames` call is preserved against concurrent bus inbound.

The change is local to a single method and does not alter any frame shape, frame type, or external API. Lifecycle/excluded frame filtering upstream is unchanged.

## Test plan

- [x] `uv run pytest tests/test_bridge_processor.py -v` — 8 passed (1 new `test_bus_inbound_drains_after_worker_queued_frames`)
- [x] `uv run ruff check / format` clean
- [x] **End-to-end deterministic repro** of the race: multi-worker pipecat-flows with a fake LLM logging the context it sees. Without the fix the LLM generates against the previous node's task_messages/tools on 10/10 trials when the race window is amplified; with the fix applied it generates against the new node's content on 10/10 trials.
- [x] **Verified live on a downstream worker** (5x sinistri + 3x bookings runs) with an existing intra-worker mute-during-set_node workaround disabled: zero stale-context generations, zero "No tools provided" errors, no interaction with the existing `BusBridgeProcessor` frame dedup.

### DIFF
--- a/src/pipecat/bus/bridge_processor.py
+++ b/src/pipecat/bus/bridge_processor.py
@@ -32,7 +32,7 @@ from pipecat.frames.frames import (
 from pipecat.processors.frame_processor import FrameDirection, FrameProcessor, FrameProcessorSetup
 
 if TYPE_CHECKING:
-    from pipecat.workers.base_worker import BaseWorker
+    from pipecat.pipeline.worker import PipelineWorker
 
 _LIFECYCLE_FRAMES = (StartFrame, EndFrame, CancelFrame, StopFrame)
 _PASSTHROUGH_FRAMES = (OutputTransportMessageUrgentFrame,)
@@ -161,7 +161,7 @@ class _BusEdgeProcessor(FrameProcessor, BusSubscriber):
     def __init__(
         self,
         *,
-        worker: "BaseWorker",
+        worker: "PipelineWorker",
         direction: FrameDirection,
         bridges: tuple[str, ...] = (),
         exclude_frames: tuple[type[Frame], ...] | None = None,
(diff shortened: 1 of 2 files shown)


src/pipecat/bus/bridge_processor.py
```diff
@@ -217,7 +217,23 @@ async def process_frame(self, frame: Frame, direction: FrameDirection):
...
+        LLMRunFrame]``), since both paths share the worker's single
+        ``_push_queue`` (drained FIFO by ``_process_push_queue``).
+
+        Without this routing, a bus inbound frame can interleave between
+        a flow's apply frames and the run trigger that follows them, so an
+        ``LLMContextFrame`` derived from a user input can reach the
+        ``LLMService`` before the new node's ``LLMSetToolsFrame`` /
+        ``LLMMessagesUpdateFrame`` have landed, and the LLM generates
+        against the previous node's context.
+        """
```
aconchillo: I think we can remove all this docstring (just live the first line) and just add a small comment before the change below.

Maurotb: @aconchillo done — dropped the long docstring, kept a 3-line comment right before the change. Thanks for the review.

aconchillo (approved):

aconchillo:
LGTM! Thank you! For next time, please, also add a `changelog/` file. :pray:

PR #4658, opened 2026-06-07
Promote websockets to a core dependency

aconchillo:
## What this does

Makes `websockets` a regular core dependency of `pipecat-ai` instead of the `websockets-base` optional extra.

- Adds `websockets>=13.1` to the core `dependencies`.
- Removes the `websockets-base` extra and every `pipecat-ai[websockets-base]` self-reference from the service extras (Cartesia, Deepgram, ElevenLabs, OpenAI, Google, etc.). Those extras still exist and behave the same; they just no longer need to pull `websockets` in themselves since it now comes from core.
- `WebSocketProxyClient` now imports `websockets` directly, dropping the `try/except ImportError` guard that pointed users at the now-removed extra.

### Why

`websockets` is used broadly across the codebase (most streaming STT/TTS/LLM services, the WebSocket transport, the worker proxy). Gating it behind an extra that nearly every service re-declared added indirection without real benefit. Folding it into core simplifies the dependency graph and removes a class of "module not found" surprises.


markbackman (approved):
I'm on board with this!

This resolves a few of the comments I just made on the eval PR.

Update:
Actually the llm_service module already pulls in websockets, so this resolves a latent issue. Nice!

kompfner:
> Update: Actually the llm_service module already pulls in websockets, so this resolves a latent issue. Nice!

#4651 was trying to fix exactly this issue in a targeted way, but the fix in this PR seems better - it seems rare not to need websockets.

kompfner:
Now that websockets are a core dependency, there are a bunch of warnings like [this one](https://github.com/pipecat-ai/pipecat/blob/75f488ddc559a4a3cccfd88ef831a8ea01b2be5e/src/pipecat/services/cartesia/tts.py#L37-L43) that we ought to clean up:

```python
try:
    logger.error("In order to use Cartesia, you need to `pip install pipecat-ai[cartesia]`.")
    raise ImportError(f"Missing module: {e}") from e
```

I have some in-progress work touching a bunch of these warnings so maybe I can work this clean up there too.

markbackman:
> Now that websockets are a core dependency, there are a bunch of warnings like [this one](https://github.com/pipecat-ai/pipecat/blob/75f488ddc559a4a3cccfd88ef831a8ea01b2be5e/src/pipecat/services/cartesia/tts.py#L37-L43) that we ought to clean up:

Good point @kompfner. Want to tackle that?

PR #4660, opened 2026-06-08
Adding HMAC token authentication for WebSocket connections.

filipi87:
## Summary

  - Added optional HMAC token authentication for WebSocket connections in the development runner, mirroring the `websocket_auth ="token"` setting in Pipecat Cloud
  - When enabled, clients must call `POST /start` to obtain a short-lived signed session token before connecting — no bot resources are allocated until the connection is authenticated
  - Tokens are HMAC-SHA256 signed, one-time use, and expire after 5 minutes
  - Both the telephony WebSocket (`/ws`) and plain WebSocket (`/ws-client`) endpoints are protected; new path-segment routes (`/ws/{token}`, `/ws-client/{token}`) support telephony providers like Twilio that can't send custom headers
  
  ## Configuration

  Enable via environment variable (in `.env`):
  PIPECAT_WEBSOCKET_AUTH=token

  Or via CLI flag:
  python bot.py -t websocket --ws-auth token
  python bot.py -t twilio --ws-auth token

  When enabled, `POST /start` returns a `token` alongside `wsUrl`. Clients attach it as:
  - `Authorization: Bearer <token>` header
  - `?token=<token>` query parameter
  - `/<token>` URL path segment (recommended for telephony providers)

  Invalid, expired, or replayed tokens are rejected with WebSocket close code `4003`


markbackman (approved):
LGTM! Thanks for the quick addition. I'm having Claude take a look too.

PR #4663, opened 2026-06-08
fix interruption handling for TTSSpeakFrame utterances

kompfner:
## Problem

Fixed interruption handling for standalone `TTSSpeakFrame(append_to_context=True)` utterances (those not part of an LLM response — e.g. an intro/greeting spoken before the user's first turn). Previously, when the user interrupted such an utterance:

- `on_assistant_turn_stopped` didn't fire
- partially-spoken text wasn't recorded to the context (for TTS services that support word timestamps)

The problem was that these utterances have no `LLMFullResponseStartFrame` to open the assistant turn, so there was no open turn for the interruption to stop. (`LLMAssistantAggregator` tracks the open-turn state via `_assistant_turn_start_timestamp`, which is set on `LLMFullResponseStartFrame`.)

## Fix

The assistant aggregator now uses a new `TTSStartedFrame.append_to_context` to open the turn when the utterance begins:

- `TTSStartedFrame` carries `append_to_context`, stamped centrally in `TTSService._handle_audio_context` — the single point every started frame funnels through, whether the base class emits it (`push_start_frame=True`) or a service yields its own from `run_tts` (`push_start_frame=False`: `NvidiaTTSService` and the websocket services Inworld/Rime/ElevenLabs/ResembleAI). The TTS context is the source of truth for the value.
- On a `TTSStartedFrame(append_to_context=True)` with no turn already open, the aggregator opens the turn. An in-progress LLM turn (`LLMFullResponseStartFrame`) still takes precedence, as it's the earlier signal.
- This replaces the just-in-time `_trigger_assistant_turn_started` workaround in `_handle_push_aggregation` from #4414.

As a result of this fix, `on_assistant_turn_started` timing is improved for standalone `TTSSpeakFrame` utterances: the event now fires at the start of the utterance rather than at the end.

## Tests

- Reframed the existing push-aggregation test to the real frame sequence (`TTSStartedFrame` opens the turn).
- Aggregator cases: turn opens on `TTSStartedFrame`; interruption records partial text with `interrupted=True`; `append_to_context=False` doesn't open a turn; `TTSStartedFrame` during an LLM response doesn't double-fire `on_assistant_turn_started`.
- TTS-service case: a `TTSSpeakFrame`'s `append_to_context` reaches the emitted `TTSStartedFrame` for both the base-class and self-yielded paths (True and False).

Validated manually against word-timestamp (Cartesia) and all-or-nothing (OpenAI) TTS services across the intro-speech and function-call-filler scenarios, uninterrupted and interrupted.

## Follow-up cleanup

A separate commit (`refactor: drop dead None-handling for append_tts_text_to_context`) removes dead code surfaced while doing the above: `_push_tts_frames` is private and `append_tts_text_to_context` is never `None` at any call site, so its `bool | None` annotation and the `is not None` branches that guarded the impossible case are dropped. No behavior change.

### DIFF
--- a/src/pipecat/services/nvidia/tts.py
+++ b/src/pipecat/services/nvidia/tts.py
@@ -555,12 +555,7 @@ class NvidiaTTSService(TTSService):
             if not self.audio_context_available(context_id):
                 await self.create_audio_context(context_id)
                 await self.start_ttfb_metrics()
-                started_frame = TTSStartedFrame(context_id=context_id)
-                # Mirror append_to_context from the context (set by the base class).
-                tts_context = self._tts_contexts.get(context_id)
-                if tts_context is not None:
-                    started_frame.append_to_context = tts_context.append_to_context
-                yield started_frame
+                yield TTSStartedFrame(context_id=context_id)
                 self._start_synthesis_stream(context_id)
                 logger.trace(f"{self}: Started synthesis stream for context {context_id}")
 
--- a/src/pipecat/services/nvidia/tts.py
+++ b/src/pipecat/services/nvidia/tts.py
@@ -556,10 +556,7 @@ class NvidiaTTSService(TTSService):
                 await self.create_audio_context(context_id)
                 await self.start_ttfb_metrics()
                 started_frame = TTSStartedFrame(context_id=context_id)
-                # Mirror append_to_context from the context (set by the base class)
-                # so the assistant aggregator opens a turn only when the spoken
-                # text will be written to context — matching the base-class
-                # TTSStartedFrame path used by push_start_frame=True services.
+                # Mirror append_to_context from the context (set by the base class).
                 tts_context = self._tts_contexts.get(context_id)
                 if tts_context is not None:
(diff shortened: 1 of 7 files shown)


src/pipecat/services/nvidia/tts.py
```diff
@@ -555,7 +555,12 @@ async def run_tts(self, text: str, context_id: str) -> AsyncGenerator[Frame | No
             if not self.audio_context_available(context_id):
                 await self.create_audio_context(context_id)
                 await self.start_ttfb_metrics()
-                yield TTSStartedFrame(context_id=context_id)
+                started_frame = TTSStartedFrame(context_id=context_id)
```
kompfner: Hmm, there are other services that emit their own `TTSStartedFrame`...those should be included here too

kompfner:
Superseded by #4665. The branch was renamed `paul/tts-speak-turn-events` → `pk/tts-speak-turn-events`, which closed this PR (GitHub couldn't retarget the head ref), so the work continues in #4665.

PR #4664, opened 2026-06-09
feat(cli): scaffold the eval transport in init via --eval

markbackman:
## Summary

Adds an opt-in `--eval` flag to `pipecat init` so a scaffolded bot can be run with `-t eval` for behavioral evals out of the box — removing the previously-manual "make your bot eval-able" step.

- **`--eval/--no-eval`** flag (non-interactive) and an **`Enable evals?`** wizard question (interactive), both **off by default** — consistent with the other optional feature toggles (`--recording`, `--observability`, `--enable-krisp`).
- When enabled, the generated bot gets an inert `"eval"` entry in `transport_params` (a `WebsocketServerParams` mirroring the bot's audio/video flags). It's only used when the bot is run with `-t eval`; it changes nothing in normal operation.
- The required `WebsocketServerParams` import is registered as an **`eval` feature in `FEATURE_DEFINITIONS`** (regenerating `_imports.py`), so it flows through the same registry path as every other feature import rather than being injected ad hoc.

## Why opt-in

Agentic coding agents scaffold headless and pass `--eval` explicitly; humans at the wizard get a default-No prompt. The transport entry is cheap (one core-safe import — `websockets` is already core — plus an inert dict entry), so there's no cost to leaving it off and no clutter when unused.

## Test plan

- New `test_eval_transport_opt_in`: asserts the `eval` entry + import are generated with `enable_eval=True`, absent otherwise, and the generated bot passes `ast.parse`.
- Full CLI suite: **270 passing**; ruff + format clean.
- Manually verified `--eval` against cascade and realtime templates (with/without `--video-input`): generated bots `py_compile` and `ruff check` clean; `--dry-run` reflects `enable_eval`.
- Pre-commit registry-completeness hook confirms `FEATURE_DEFINITIONS` → `_imports.py` is consistent.


markbackman:
Closed in favor of https://github.com/pipecat-ai/pipecat/pull/4689

PR #4667, opened 2026-06-09
Fix OpenAIResponsesHttpLLMService crash on missing token detail fields

jamsea:
## Problem

`OpenAIResponsesHttpLLMService` reads token usage with direct nested attribute access:

```python
cache_read_input_tokens=response.usage.input_tokens_details.cached_tokens,
reasoning_tokens=response.usage.output_tokens_details.reasoning_tokens,
```

`input_tokens_details` and `output_tokens_details` are part of the official Responses API `usage` schema (both are required fields on `ResponseUsage` in the OpenAI SDK). But the SDK parses responses leniently, so when an OpenAI-compatible server returns a `usage` object without those sub-objects, the SDK leaves them as `None`. The nested access then raises:

```
'NoneType' object has no attribute 'cached_tokens'
```

This fires a non-fatal `ErrorFrame` on **every turn** and means token usage metrics are never recorded.

## Why this is a Pipecat-side fix

The WebSocket sibling, `OpenAIResponsesLLMService`, already reads these defensively in its `response.completed` handler:

```python
input_details = usage.get("input_tokens_details") or {}
output_details = usage.get("output_tokens_details") or {}
...
cache_read_input_tokens=input_details.get("cached_tokens", 0),
reasoning_tokens=output_details.get("reasoning_tokens", 0),
```

So the codebase already treats a partial `usage` object as a real case. The HTTP variant just missed the same guard, which makes the two variants behave differently against the same backend. This PR brings the HTTP variant in line with the WebSocket one.

Custom `base_url` for proxied / self-hosted deployments is an explicitly documented use case for both variants, so tolerating an OpenAI-compatible server that omits these sub-objects is reasonable. (A spec-conformant server would still include them, e.g. `input_tokens_details: {"cached_tokens": 0}`; this fix just stops Pipecat from crashing when one does not.)

## Fix

Guard the nested access with `getattr` fallbacks, so a missing `input_tokens_details` / `output_tokens_details` (or a present-but-`None` sub-field) falls back to `0`:

```python
input_details = response.usage.input_tokens_details
output_details = response.usage.output_tokens_details
tokens = LLMTokenUsage(
    prompt_tokens=response.usage.input_tokens,
    completion_tokens=response.usage.output_tokens,
    total_tokens=response.usage.total_tokens,
    cache_read_input_tokens=getattr(input_details, "cached_tokens", 0) or 0,
    reasoning_tokens=getattr(output_details, "reasoning_tokens", 0) or 0,
)
```

For native OpenAI responses (where the detail objects are present) behavior is unchanged.

## Verification

Reproduced against the OpenAI SDK (2.38.0) with a `usage` object whose detail sub-objects are `None`:

- Old code: `AttributeError: 'NoneType' object has no attribute 'cached_tokens'`
- New code: `cache_read_input_tokens=0`, `reasoning_tokens=0`, no error
- Native case (details present, e.g. `cached_tokens=7`, `reasoning_tokens=3`): values pass through unchanged

## How it came up

Reported by a user running a local server exposed over a Cloudflare tunnel with a `/v1/responses` endpoint, configured via `OpenAIResponsesHttpLLMService(base_url=...)`. Their server returns a valid `usage` object without the detail sub-objects, so every turn hit the crash above while the Chat Completions path against the same server worked fine.


markbackman (approved):
LGTM. I added some additional hardening, but this looks good to go. Appreciate you working through this.

PR #4668, opened 2026-06-09
TTS detection in Krisp VIVA filter. Added support for TTS model path …

realgarik:
TTS detection in Krisp VIVA filter. Added support for TTS model path and timeout options, allowing noise cancellation to be delayed until TTS audio clears. Updated usage instructions and added detailed logging for TTS detection phase


markbackman:
cc @filipi87 to review.

src/pipecat/audio/filters/krisp_viva_filter.py
```diff
@@ -29,36 +29,54 @@
...
+        tts_detection_timeout: float = 3.0,
     ) -> None:
         """Initialize the Krisp noise reduction filter.
 
         Args:
-            model_path: Path to the Krisp model file (.kef extension).
+            model_path: Path to the Krisp NC model file (.kef extension).
                 If None, uses KRISP_VIVA_FILTER_MODEL_PATH environment variable.
+            tts_model_path: Path to the Krisp TTS detection model file (.kef extension).
+                If None, uses KRISP_VIVA_TTS_MODEL_PATH environment variable.
```
filipi87: nit: We should also add `KRISP_VIVA_TTS_MODEL_PATH` to `env.example`, together with the other Krisp variables.

src/pipecat/audio/filters/krisp_viva_filter.py
```diff
@@ -29,36 +29,54 @@
...
+    Optionally supports TTS detection (iPhone screening feature) to delay voice isolation until
+    bot speech playback has stopped, preventing later real human speech suppression artifacts.
+    Provide ``tts_model_path`` (or set the ``KRISP_VIVA_TTS_MODEL_PATH`` environment variable) to enable this feature.
     """
 
     def __init__(
         self,
-        model_path: str = None,
+        model_path: str | None = None,
+        tts_model_path: str | None = None,
```
filipi87: We should move `tts_model_path` after `api_key`, consistent with how `tts_threshold` and `tts_detection_timeout` are placed, so we don't break external callers that pass positional arguments.

src/pipecat/audio/filters/krisp_viva_filter.py
```diff
@@ -29,36 +29,54 @@
...
+
 
 class KrispVivaFilter(BaseAudioFilter):
     """Audio filter using the Krisp VIVA SDK.
 
     Provides real-time noise reduction for audio streams using Krisp's
     proprietary noise suppression algorithms. This filter requires a
     valid Krisp model file to operate.
+
+    Optionally supports TTS detection (iPhone screening feature) to delay voice isolation until
```
filipi87: Do we need to keep (iPhone screening feature)? It could make it seem like this feature is iPhone specific.

src/pipecat/audio/filters/krisp_viva_filter.py
```diff
@@ -200,9 +280,41 @@ async def filter(self, audio: bytes) -> bytes:
...
 
-            # Process the complete frames
             samples = np.frombuffer(audio_to_process, dtype=np.int16)
             frames = samples.reshape(-1, self._samples_per_frame)
+
+            # TTS detection phase: pass audio through until bot speech clears
+            if self._tts_detection_active and self._tts_detector is not None:
+                frame_duration_s = self._frame_duration_ms / 1000.0
+
+                for i, tts_frame in enumerate(frames):
```
filipi87: Could we do this instead ? 
`for tts_frame in frames:`

So, remove the `enumerate()` since `i` isn't used.

src/pipecat/audio/filters/krisp_viva_filter.py
```diff
@@ -200,9 +280,41 @@ async def filter(self, audio: bytes) -> bytes:
             # Remove processed bytes from buffer, keep the remainder
             self._audio_buffer = self._audio_buffer[bytes_to_process:]
 
-            # Process the complete frames
             samples = np.frombuffer(audio_to_process, dtype=np.int16)
             frames = samples.reshape(-1, self._samples_per_frame)
+
+            # TTS detection phase: pass audio through until bot speech clears
+            if self._tts_detection_active and self._tts_detector is not None:
```
filipi87, line 322: Maybe isolate all this TTS logic into its own method. It would make the filter easier to read.

Something like this:
```
  if self._tts_detection_active and self._tts_detector is not None:
      if not self._advance_tts_detection(frames):
          return audio_to_process
      self._tts_detection_active = False
```

So, `_advance_tts_detection` contains all this logic to run TTS detection on frames and return `True` when NC should activate.

src/pipecat/audio/filters/krisp_viva_filter.py
realgarik: @filipi87 the issues we saw from the customers so far are for iPhone screening feature and for now we have a separate dedicated model for it. So depending on the model type loaded the corresponding TTS detection will work.

src/pipecat/audio/filters/krisp_viva_filter.py
realgarik: done

src/pipecat/audio/filters/krisp_viva_filter.py
realgarik: done

src/pipecat/audio/filters/krisp_viva_filter.py
realgarik: done

src/pipecat/audio/filters/krisp_viva_filter.py
realgarik: done

filipi87 (approved):
LGTM! Thank you @realgarik .

PR #4670, opened 2026-06-09
Point install instructions at `uv add` instead of `pip install`

kompfner:
Consistency pass to align reader-facing install guidance around uv, our recommended package manager (as reflected in the main README).

Also: remove various `websockets` import guards, which are no longer needed now that `websockets` is a core dependency (see https://github.com/pipecat-ai/pipecat/pull/4658).


markbackman (approved):
Nice clean up! 🧹

PR #4671, opened 2026-06-09
Migrate examples to direct functions; support direct functions in LLMSwitcher and LLMSetToolsFrame

markbackman:
## Summary

Standardizes Pipecat's examples on the **direct function** tool pattern — plain async functions listed in `LLMContext(tools=[...])`, with per-tool options via `@tool_options` — and adds the framework support needed to make that pattern work uniformly. Also renames the (unreleased) `@direct_function` decorator to `@tool_options`.

## Framework changes

- **`LLMSwitcher` auto-registers context direct functions on every member LLM** (active or not). Member LLMs sit behind per-branch filters, so only the active one would otherwise see the `LLMContextFrame` and register the handler — breaking direct functions after a switch. Now `LLMContext(tools=[...])` works across switches with no explicit registration. Adds the public `LLMService.register_context_direct_functions()`.
- **`LLMSetToolsFrame` accepts direct functions** — its `tools` field now takes a plain list of direct functions / `FunctionSchema` objects (not just a `ToolsSchema`), and the LLM service auto-registers handlers for any direct functions advertised mid-conversation.
- **Renamed the unreleased `@direct_function` decorator to `@tool_options`.** Its only job is attaching call options (`cancel_on_interruption`, `timeout`) — the schema is auto-derived from the signature/docstring regardless — so `@tool_options` names what it does, and `@tool` (for `LLMWorker` methods) composes it. The released "direct function" concept, `DirectFunction` type, `DirectFunctionWrapper`, and `register_direct_function()` are unchanged.

## Example changes

- **Function-calling examples** migrated from the verbose pattern (`FunctionSchema` + `ToolsSchema` + `register_function`) to direct functions — `examples/function-calling/` plus the function-calling examples under `features/`, `persistent-context/`, `thinking/`, `turn-management/`, `rag/`, `context-summarization/`, and `getting-started/`. Async options move to `@tool_options`; enum constraints are expressed in docstring prose (the direct-function schema generator doesn't yet map `Literal` → JSON-schema `enum`).
- **Realtime examples** converted where the service advertises tools via `LLMContext` (OpenAI Realtime, Azure Realtime, Grok Realtime, Inworld), including the persistent-context realtime variants and the dynamic `LLMSetToolsFrame` tool-swap in the OpenAI examples.
- **Multi-worker examples**: dropped redundant `register_direct_function()` calls where a plain `LLMService` sits in a standard pipeline (code-assistant, parallel-debate, sensor-controller, and the voice-agent LLM in each `ui-worker` example), moving per-tool options to `@tool_options`. The `@tool` methods on `LLMWorker`/`UIWorker` subclasses are unchanged — those are the workers' own auto-collected tools, a different mechanism.
- **Intentionally left verbose:** `missing-handler` (demonstrates the missing-handler recovery path), the MCP examples (external tools), and the two realtime services that advertise tools via the service constructor rather than `LLMContext` — **Gemini Live** (`tools=...`) and **Ultravox** (`one_shot_selected_tools=`), which structurally require a `ToolsSchema`.

## Testing

- New `tests/test_llm_switcher.py`; added coverage to `tests/test_llm_service.py`, `tests/test_context_aggregators_universal.py`, and `tests/test_auto_register_tools.py`.
- Changed source files compile, lint, and format clean; the affected test suites pass.


src/pipecat/adapters/schemas/direct_function.py
```diff
@@ -298,13 +298,16 @@ def direct_function(fn=None, *, cancel_on_interruption: bool = True, timeout: fl
     attaches the options to the function and returns it unchanged — it does not
     register anything, so decorated functions can stay at module level.
 
+    On an ``LLMWorker``, use ``@tool`` instead, which applies these same options
+    and additionally marks the method for collection as one of the worker's tools.
+
     Can be used with or without arguments::
 
-        @direct_function
+        @tool_options
```
filipi87, line 308: Agreed. This name does make more sense. 💯

src/pipecat/processors/aggregators/llm_response_universal.py
```diff
@@ -799,6 +799,11 @@ async def process_frame(self, frame: Frame, direction: FrameDirection):
         elif isinstance(frame, LLMMessagesTransformFrame):
             await self._handle_llm_messages_transform(frame)
         elif isinstance(frame, LLMSetToolsFrame):
+            # Normalize once (a plain list of direct functions / FunctionSchema
+            # objects becomes a ToolsSchema) so the tool-change diff, set_tools,
+            # and any downstream consumer of the forwarded frame all see a
+            # consistent type.
+            frame.tools = LLMContext._normalize_and_validate_tools(frame.tools)
```
filipi87: We are basically overriding `frame.tools` here. Was that the intention, or should we create a new variable ? For example:

`normalized_tools = LLMContext._normalize_and_validate_tools(frame.tools)`

This way, we wouldn't be modifying the frame itself.

src/pipecat/processors/aggregators/llm_response_universal.py
```diff
@@ -1476,6 +1481,11 @@ async def process_frame(self, frame: Frame, direction: FrameDirection):
...
```
filipi87: Same here.

src/pipecat/services/llm_service.py
```diff
@@ -844,13 +850,27 @@ def register_direct_function(
...
+        This is the same registration that runs automatically when the service
+        processes an ``LLMContextFrame``. It is exposed publicly so that container
+        processors (such as ``LLMSwitcher``) can register handlers on member
+        services that may not individually receive the context frame.
+
+        Args:
+            context: The LLM context whose advertised direct functions should be
+                registered on this service.
+        """
+        self._auto_register_direct_functions(context)
```
filipi87: Do we need this method ? It's basically just calling `self._auto_register_direct_functions`. 

Couldn't we make `self._auto_register_direct_functions` public instead and rename it if needed ?

src/pipecat/processors/aggregators/llm_response_universal.py
markbackman: Makes sense! Done.

src/pipecat/processors/aggregators/llm_response_universal.py
markbackman: Done.

src/pipecat/services/llm_service.py
markbackman: Yup, that's cleaner. Making this fix.

filipi87 (approved):
LGTM! 🚀

src/pipecat/services/llm_service.py
```diff
@@ -592,7 +602,14 @@ async def process_frame(self, frame: Frame, direction: FrameDirection):
             # context. Runs before the subclass processes the context (subclasses
             # call super().process_frame() first), so handlers are in place by the
             # time the LLM can call them.
-            self._auto_register_direct_functions(frame.context)
+            self.register_context_direct_functions(frame.context)
+        elif isinstance(frame, LLMSetToolsFrame):
+            # Tools can change mid-conversation. Register handlers for any direct
+            # functions newly advertised by the frame, so they're callable as soon
+            # as the updated tool set takes effect. This is an explicit tool-set
```
kompfner: Hmm, I'm wondering if it might be OK for people explicitly invoking `unregister` to have to explicitly invoke `register` to get those tools back...(after all, that's what they'd have had to do until now, right?)

That would let us get rid of `LLMSetToolsFrame` handling here, right? To me it feels a bit messy/confusing for us to handle `LLMSetToolsFrame` in _both_ `LLMService` and the aggregators.

examples/function-calling/function-calling-anthropic-async-stream.py
```diff
@@ -57,8 +52,11 @@
 load_dotenv(override=True)
 
 
+@tool_options(cancel_on_interruption=False, timeout=30)
```
kompfner: I like this new name that avoids implying it's necessary for making a function a tool

kompfner: We should update the Flows `@flows_direct_function` decorator accordingly:

- Should now use `@tool_options` under the hood
- Rename to `@flows_tool_options` (keeping around deprecated `@flows_direct_function` as an alias)

markbackman: Corresponding Flows change, which request the 1.4.0 release:
https://github.com/pipecat-ai/pipecat-flows/pull/282

aconchillo (approved):

src/pipecat/processors/aggregators/llm_response_universal.py
```diff
@@ -800,8 +799,12 @@ async def process_frame(self, frame: Frame, direction: FrameDirection):
         elif isinstance(frame, LLMMessagesTransformFrame):
             await self._handle_llm_messages_transform(frame)
         elif isinstance(frame, LLMSetToolsFrame):
-            self._maybe_add_tool_change_messages(frame.tools)
-            self.set_tools(frame.tools)
+            # Normalize and validate (a plain list of direct functions / FunctionSchema
+            # objects becomes a ToolsSchema) so the tool-change diff and
+            # set_tools see a consistent type.
+            normalized_tools = LLMContext._normalize_and_validate_tools(frame.tools)
```
aconchillo, line 806: Should we make `LLMContext._normalize_and_validate_tools` public since it's now used here?

aconchillo:
LGTM! One small comment.

src/pipecat/processors/aggregators/llm_response_universal.py
markbackman: I don't think this should be public because we don't intend for developers to use it.

But, it is a `@staticmethod`. Maybe we should promote it to a module-level function? WDYT?

src/pipecat/services/llm_service.py
```diff
@@ -812,85 +822,146 @@ def register_direct_function(
         self,
         handler: DirectFunction,
         *,
-        cancel_on_interruption: bool = True,
-        timeout_secs: float | None = None,
+        cancel_on_interruption: bool | NotGiven = NOT_GIVEN,
```
kompfner: Hmm, `NOT_GIVEN` has so far been a special thing _just_ for `LLMContext` params...

In the case of `cancel_on_interruption`, we could easily use `None` instead.

But in the case of `timeout_secs`, where `None` today already means "use the global function_call_timeout_secs", if we used `None` we'd actually be changing its meaning to "use the tool_options value, and, if that's None, use function_call_timeout_secs". The user wouldn't have the ability to explicitly say "hey, override whatever's in tool_options with whatever is in function_call_timeout_secs".

But maybe that behavior change is totally acceptable - tool_options is new anyway, so anyone migrating to this new Pipecat version won't automatically see a behavior change. And wanting to override what's in tool_options for _another_ default (i.e. "skip the next fallback in the fallback chain") feels like a very very niche case (which someone could achieve by specifying the function_call_timeout_secs for this value).

WDYT?

kompfner: Alternatively...we could use the `NOT_GIVEN`/`is_given` from `LLMSettings`, and "promote" those to be a sharable thing across Pipecat (in its own file, say).

(I don't necessarily want to use `LLMContext`'s types because they're also aliases for OpenAI's types)

src/pipecat/services/llm_service.py
```diff
@@ -910,16 +981,76 @@ def _register_direct_functions_from_tools(self, tools: Any) -> None:
...
+        normalized = (
+            LLMContext._normalize_and_validate_tools(tools) if tools is not None else NOT_GIVEN
+        )
+        self._register_direct_functions_from_tools(normalized)
+        advertised: set[str | None] = (
+            {wrapper.name for wrapper in normalized.direct_functions}
+            if is_given(normalized)
+            else set()
+        )
+        self._prune_unadvertised_functions(advertised)
```
kompfner: Nit: naming inconsistency between `_register_direct_functions_from_tools` and `_prune_unadvertised_functions`, which are conceptually two sides of the same coin.

Something more like...
- `_register_advertised_direct_functions`
- `_prune_unadvertised_direct_functions` (or `_unregister_unadvertised_direct_functions`; I like this latter one since it mentions the registry, which is important)

...would be more consistent

src/pipecat/services/llm_service.py
```diff
@@ -910,16 +981,76 @@ def _register_direct_functions_from_tools(self, tools: Any) -> None:
...
+            advertised: Names of direct functions in the new advertised tool set.
+        """
+        stale = [
+            name
+            for name, item in self._functions.items()
+            if item.auto_registered and name not in advertised
+        ]
+        for name in stale:
+            del self._functions[name]
+        # If the last async tool was just pruned, tear down the cancellation tool.
```
kompfner, line 1043: Nice catch

kompfner (approved):
Two minor considerations. Otherwise LGTM! This is great; the examples are much cleaner.

src/pipecat/services/llm_service.py
markbackman: We discussed in Slack. We want a general purpose "NOT_GIVEN" sentinel, but this isn't the use case driving it.

Updated.

src/pipecat/services/llm_service.py
markbackman: I like it!

markbackman: ✅ Went with:
- `_register_advertised_direct_functions`
- `_unregister_unadvertised_direct_functions`

PR #4673, opened 2026-06-09
Avoid import errors for transports you're not using

kompfner:
The import errors we're trying to avoid are the ones that look like this:

```
In order to use the Daily transport, you need to `uv add "pipecat-ai[daily]"
```


markbackman (approved):
Nice catch and clean up!

PR #4674, opened 2026-06-09
Add option to disable repeated sequence filtering in MarkdownTextFilter

YFortin:
## Summary

Adds a `filter_repeated_sequences` parameter to `MarkdownTextFilter.InputParams` that allows disabling the repeated sequence removal (regex that strips 5+ repeated characters).

## Changes

- Added `filter_repeated_sequences: bool | None = True` to `InputParams`
- Wrapped the repeated sequence regex in a conditional check
- Added unit tests for the new parameter

## Motivation

Some use cases need to preserve repeated characters in the output text (e.g., phone extension numbers like `22222`). This option allows disabling that specific filter while keeping all other markdown filtering active.

Defaults to `True` to preserve existing behavior.

## Documentation

- Docs PR: https://github.com/pipecat-ai/docs/pull/881

### DIFF
--- a/changelog/4674.added.md
+++ b/changelog/4674.added.md
@@ -1,0 +1,1 @@
+- Added `filter_repeated_sequences` parameter to `MarkdownTextFilter.InputParams` to allow disabling repeated sequence removal.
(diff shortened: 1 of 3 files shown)


markbackman (approved):
LGTM!

In seeing this PR, I checked if TTS services still implode when they see repeated characters. Some, like Cartesia, have improved. Others, like ElevenLabs, have not. This check is still valid, but I agree that we should add an arg to control it.

Issue #4675, opened 2026-06-10
DeepgramSTTService sends `language=Language.EN` (WS 400 on every connect) — pipecat 0.0.10x + Python 3.10 vendored StrEnum fallback


kira-ariaki:
## Summary

On the **0.0.10x release line running Python 3.10**, `DeepgramSTTService` sends the literal string `Language.EN` as the `language` query parameter when connecting to Deepgram's WebSocket API. Deepgram rejects the connection with:

```
Unexpected error when initializing websocket connection
```

(HTTP 400) **on every connect**, so the service never transcribes anything.

This is filed primarily for **discoverability**: `main` requires Python ≥3.11 and is not affected, and the 0.0.10x line will likely not see another release. But anyone pinned to `pipecat-ai==0.0.10x` on a `python:3.10` base image hits this with the **default settings** (`language=Language.EN`), and the Deepgram error message gives no hint of the cause. Searching that error message should land people here.

## Root cause (three layers)

**1. The vendored StrEnum fallback doesn't override `__str__`.**

`pipecat/transcriptions/language.py` (0.0.108):

```python
if sys.version_info < (3, 11):
    class StrEnum(str, Enum):
        def __new__(cls, value):
            obj = str.__new__(cls, value)
            obj._value_ = value
            return obj
else:
    from enum import StrEnum
```

The fallback only defines `__new__`. The stdlib `StrEnum` (3.11+) also sets `__str__ = str.__str__`, which is exactly what makes `str(member)` return the value. Without it, `str()` resolves through `Enum.__str__` and returns the **member name**:

```python
# Python 3.10:  str(Language.EN) -> "Language.EN"
# Python 3.11+: str(Language.EN) -> "en"
```

**2. The Deepgram service serializes with `str()`.**

`pipecat/services/deepgram/stt.py` (`_build_connect_kwargs`, line 566 in 0.0.108):

```python
if is_given(s.language) and s.language is not None:
    kwargs["language"] = str(s.language)
```

**3. Deepgram rejects `language=Language.EN` with a 400.**

So this is a *version-branch-selection* bug, not an enum semantics change: any dev on 3.11+ can never reproduce it, which is presumably why it shipped.

## Minimal mechanism demo (6 lines, no installs)

```python
from enum import Enum
class StrEnum(str, Enum):                      # the <3.11 fallback, verbatim
    def __new__(cls, value):
        obj = str.__new__(cls, value); obj._value_ = value; return obj
[...]

PR #4676, opened 2026-06-10
Use explicit .value for Language enum wire serialization

kira-ariaki:
## What

Replaces `str(language)` with explicit `.value` extraction (`language.value if isinstance(language, Language) else str(language)`) at every site where a `Language` enum member is serialized into a wire value (query params, request payloads, trace attributes):

- `services/deepgram/stt.py` — WS connect query param
- `services/deepgram/sagemaker/stt.py` — connect query param
- `services/xai/tts.py` — HTTP payload (the WS path in `_build_url` already does this correctly; this aligns the HTTP service with it)
- `services/groq/tts.py` — legacy `params.language` migration path
- `services/openai/stt.py` — `_language_to_code`
- `transcriptions/language.py` — `resolve_language` fallback path
- `utils/tracing/service_decorators.py` — span attribute

Adds `tests/test_language_serialization.py` with a regression test that simulates the legacy `__str__` behavior and asserts the wire value stays `"en"`.

Refs #4675

## Why

`str()` on a `StrEnum` member silently depends on the enum's `__str__` semantics. On the 0.0.10x line with Python 3.10, the vendored `StrEnum` fallback (which only defined `__new__`, not `__str__`) made `str(Language.EN)` return `"Language.EN"`, which went into the Deepgram WS query string and produced a 400 on every connect — full analysis in #4675.

`main` (Python ≥3.11, stdlib `StrEnum`) is not affected today — this is a hardening change, not a bug fix, hence "Refs" rather than "Fixes". `.value` states the intent explicitly and makes the wire format independent of `__str__`, so this regression class becomes structurally impossible regardless of how `StrEnum` is sourced or whether any subclass ever overrides `__str__`/`__format__`. It also makes the codebase consistent with itself: `xai/tts.py`'s `_build_url` already uses the `isinstance(language, Language)` + `.value` pattern.

The new test locks this in: it swaps in `Enum.__str__` (exactly what the legacy fallback resolved to), confirms `str(Language.EN) == "Language.EN"`, and asserts `DeepgramSTTService._build_connect_kwargs()` still emits `language="en"`.

## Behavior change

None on any supported Python version. Output is byte-identical; this only removes an implicit dependency on `__str__`.

## Testing

- `pytest tests/test_language_serialization.py tests/test_deepgram_stt.py` — 16 passed
- `pytest tests/test_xai_tts.py tests/test_tracing_context.py` — same results as `main` (one pre-existing fixture error on `test_run_xai_tts_success`, present on a clean checkout too)
- `ruff format --check` / `ruff check` clean on all touched files

Happy to scope this down to just the Deepgram services if you'd prefer a smaller diff.

### DIFF
--- a/changelog/4676.changed.md
+++ b/changelog/4676.changed.md
@@ -1,0 +1,1 @@
+- Language enum values are now serialized to wire formats (query params, payloads, trace attributes) using explicit `.value` instead of `str()`, removing a dependency on `StrEnum.__str__` semantics.
--- a/src/pipecat/services/deepgram/sagemaker/stt.py
+++ b/src/pipecat/services/deepgram/sagemaker/stt.py
@@ -290,7 +290,10 @@ class DeepgramSageMakerSTTService(STTService):
         if is_given(s.model) and s.model is not None:
             params["model"] = str(s.model)
         if is_given(s.language) and s.language is not None:
-            params["language"] = str(s.language)
+            # Be explicit about enum serialization: use .value rather than str()
+            # so the wire value never depends on StrEnum's __str__ semantics.
+            language = s.language
+            params["language"] = language.value if isinstance(language, Language) else str(language)
 
         # Init-only connection config
         params["encoding"] = self._encoding
--- a/src/pipecat/services/deepgram/stt.py
+++ b/src/pipecat/services/deepgram/stt.py
@@ -563,7 +563,12 @@ class DeepgramSTTService(STTService):
         if is_given(s.model) and s.model is not None:
             kwargs["model"] = str(s.model)
         if is_given(s.language) and s.language is not None:
-            kwargs["language"] = str(s.language)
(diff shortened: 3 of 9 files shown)


markbackman:
Thanks for this, and for the careful investigation in #4675. I left a summary there with the workarounds and the resolved-on-`>=1.0.0` status.

I'm going to decline the PR itself. As you noted, this is hardening rather than a fix: on every Python version pipecat supports (`requires-python = ">=3.11"`), `Language` is a stdlib `StrEnum` and `.value` is byte-identical to `str(member)`. The change guards against a condition the supported Python floor already makes impossible, at the cost of a more verbose expression across 7 call sites, a changelog entry for a no-op, and a regression test that has to monkeypatch `Enum.__str__` to simulate a state that can't occur on `>=3.11`. I'd rather not pin behavior that can't regress.

PR #4677, opened 2026-06-10
Add prefer_normalized_alignment opt-out to ElevenLabs TTS services

### LINKED ISSUE #4678: ElevenLabs TTS: pronunciation dictionary substitutions leak into transcripts and LLM context
## Problem

When `pronunciation_dictionary_locators` is configured on `ElevenLabsTTSService` or `ElevenLabsHttpTTSService`, the service hard-wires alignment selection to ElevenLabs' normalized alignment:

```python
raw_alignment = _select_alignment(
    msg,
    normalized_key="normalizedAlignment",
    alignment_key="alignment",
    prefer_normalized=bool(self._pronunciation_dictionary_locators),
)
```

That choice is deliberate: with a pronunciation dictionary active, the raw `alignment` field has overlapping restarts that produce duplicated/garbled words (#4316, fixed by #4424). But it has a side effect: pronunciation dictionary substitutions (phoneme/alt spellings) live in `normalizedAlignment`, so they flow into word timestamps, `TTSTextFrame`s, transcripts, and the assistant messages written to the LLM context. The model then sees its own prior turns as phonetic spellings, which can degrade later turns.

There is currently no public knob to decouple these: users who want the dictionary applied to the audio but the original text kept in the context have no supported path.

## Proposal

Add an explicit `prefer_normalized_alignment: bool | None` parameter to both services:

- `None` (default): current automatic behavior, unchanged
- `False`: keep the original `alignment` text for timestamps/transcripts/context while still applying the dictionary to audio (documented caveat: duplicated words possible on chunks where substitutions fire, per #4316)
- `True`: always prefer normalized alignment

## Status

PR open: #4677

## Context

- Requested by a user on Discord: https://discord.com/channels/1239284677165056021/1513551399324225536/1513551399324225536
- History: #4344 (v1.1.0, always normalized) → #4424 (v1.2.0, original by default, normalized only with PDs)

jamsea:
## Summary

When `pronunciation_dictionary_locators` is configured, `ElevenLabsTTSService` and `ElevenLabsHttpTTSService` hard-wire alignment selection to ElevenLabs' normalized alignment (`prefer_normalized=bool(self._pronunciation_dictionary_locators)`). That choice is deliberate (#4316: with a PD active, the raw `alignment` field has overlapping restarts that produce duplicated/garbled words), but it has a side effect: pronunciation dictionary substitutions (phoneme/alt spellings) live in `normalizedAlignment`, so they leak into word timestamps, transcripts, and the LLM context. The model then sees its own prior turns as phonetic spellings.

This PR adds a `prefer_normalized_alignment: bool | None` parameter to both services (constructor and the deprecated `InputParams`):

- `None` (default): current behavior, unchanged. Normalized alignment is used only when `pronunciation_dictionary_locators` is set.
- `False`: keep the original input text for timestamps/transcripts/context while still applying the PD to the audio. Documented caveat: `alignment` may contain duplicated words on chunks where substitutions fire (#4316).
- `True`: always prefer normalized alignment.

The preference is resolved once at init (`_resolve_prefer_normalized`) since both inputs are constructor-only.

Requested by a user on Discord: https://discord.com/channels/1239284677165056021/1513551399324225536/1513551399324225536

Related: #4316, #4344 (v1.1.0 always-normalized), #4424 (v1.2.0 default back to `alignment`).

## Tests

- `_resolve_prefer_normalized` unit tests (default resolution and explicit override in both directions)
- Construction-level tests for all kwarg combos on the WS service
- Deprecated `InputParams` path test (sets `prefer_normalized_alignment=False` with locators, asserts it lands and the `DeprecationWarning` fires)

`uv run pytest tests/test_elevenlabs_tts.py` passes (19 tests). Ruff format and lint clean.

Closes #4678


markbackman:
I'm not sure this is a change we want to make. Using `normalizedAlignment` with a pronunciation dictionary is broken. Making that a built-in option feels like a move that a developer would expect to work; it not working, in itself, is a bug.

AlessioSpallino:
Hi @jamsea @markbackman, is there any update on this PR?
I'm asking because my team is also affected by the same problem. We currently have an open discussion on Discord about it:
[https://discord.com/channels/1239284677165056021/1513551399324225536/1513551399324225536](https://discord.com/channels/1239284677165056021/1513551399324225536/1513551399324225536)

Thanks!

2010b9:
> Hi @jamsea @markbackman, is there any update on this PR? I'm asking because my team is also affected by the same problem. We currently have an open discussion on Discord about it: https://discord.com/channels/1239284677165056021/1513551399324225536/1513551399324225536
> 
> Thanks!

Same here!
We had to hack the internal private function `_select_alignment` to set the `prefer_normalized` to `False` while using `pronunciation_dict_locators`.

filipi87 (commented):
Instead of dictionaries and normalized words from `ElevenLabs`, we are working on providing custom transformers directly inside Pipecat ([#4854](https://github.com/pipecat-ai/pipecat/pull/4854)), which will work more closely with how Pipecat handles the context for TTS word completions.

markbackman:
Closing in Pipecat text_transforms feature. We're also deprecating pronunciation dictionary support from ElevenLabs due to usability and stability issues:
https://github.com/pipecat-ai/pipecat/pull/4991

PR #4683, opened 2026-06-10
Add MoonshineSTTService (local CPU/ONNX STT)

aconchillo:
## Summary

Adds a new `MoonshineSTTService` and a voice example that uses it.

[Moonshine](https://github.com/moonshine-ai/moonshine) is a small, fast ASR family that runs on the CPU via ONNX Runtime, so it needs no GPU and no API key. The model downloads once on first use and is cached.

## Changes

- **`pipecat.services.moonshine.stt.MoonshineSTTService`**: a `SegmentedSTTService` (Whisper-style: `Settings`, `run_stt`) backed by the `moonshine_voice` package. Each VAD-segmented utterance is transcribed in one batch call (`transcribe_without_streaming`).
- **Model is selectable** via `Settings(model=...)`, as a `Model` StrEnum member or the equivalent string: `Model.TINY`, `Model.BASE`, or a streaming model run in batch (`Model.TINY_STREAMING`, `Model.SMALL_STREAMING`, `Model.MEDIUM_STREAMING`). The default is `Model.SMALL_STREAMING`. The larger models (`small`, `medium`) ship only in streaming form but transcribe a full segment here just the same.
- **`pipecat-ai[moonshine]` extra** (`moonshine-voice`), registration in the CLI service registry, and a README entry.
- **`examples/voice/voice-moonshine.py`**: a voice bot using Moonshine for STT, OpenAI for the LLM, and Cartesia for TTS.

## Usage

```python
from pipecat.services.moonshine.stt import MoonshineSTTService, Model

# Default is Model.SMALL_STREAMING; pass another member or string to trade size for accuracy.
stt = MoonshineSTTService(
    settings=MoonshineSTTService.Settings(model=Model.BASE),
)
```

Install with `pip install "pipecat-ai[moonshine]"`. Audio is expected as 16-bit mono PCM at 16 kHz (Moonshine's input rate).


src/pipecat/cli/registry/service_metadata.py
```diff
@@ -431,6 +431,13 @@ class ServiceRegistry:
             env_prefix="SPEECHMATICS",
             include_params=["api_key"],
         ),
+        ServiceDefinition(
+            value="moonshine_stt",
+            label="Moonshine (Local)",
```
markbackman: I think we can omit "local".

```suggestion
            label="Moonshine",
```

src/pipecat/services/moonshine/stt.py
```diff
@@ -0,0 +1,144 @@
...
+from pipecat.services.stt_service import SegmentedSTTService
+from pipecat.transcriptions.language import Language
+from pipecat.utils.time import time_now_iso8601
+from pipecat.utils.tracing.service_decorators import traced_stt
+
+try:
+    from moonshine_voice import Transcriber, get_model_for_language, string_to_model_arch
+except ModuleNotFoundError as e:
+    logger.error(f"Exception: {e}")
+    logger.error("In order to use Moonshine, you need to `pip install pipecat-ai[moonshine]`.")
```
markbackman: Paul recently switched to using uv in this error log:
```suggestion
        logger.error('In order to use Moonshine, you need to `uv add "pipecat-ai[moonshine]"`.')
```

markbackman (approved):
LGTM! Just two small nits.

src/pipecat/cli/registry/service_metadata.py
aconchillo: Done.

src/pipecat/services/moonshine/stt.py
aconchillo: Done

PR #4684, opened 2026-06-10
Whisper: macOS-only MLX import and StrEnum models

aconchillo:
## Summary

Two small Whisper service improvements, split out of the Moonshine PR.

### 1. Only import MLX Whisper on macOS

Importing `pipecat.services.whisper.stt` failed on non-macOS platforms when the `mlx-whisper` extra happened to be installed: `mlx_whisper` is Apple-Silicon only, so elsewhere it can be installed but fail to load (e.g. a missing `libmlx.so`), and the bare `except ModuleNotFoundError` didn't catch that. The module-level `mlx_whisper` import is now only attempted on macOS; `WhisperSTTServiceMLX` still imports it lazily when actually used.

### 2. `Model`/`MLXModel` are now `StrEnum`

Matches the newer service convention (e.g. `smallest`, `sarvam`) and the codebase's other model enums. A `StrEnum` member is the string itself, so the existing `model if isinstance(model, str) else model.value` handling keeps working and `Model.TINY == "tiny"`.


src/pipecat/services/whisper/stt.py
```diff
@@ -27,23 +27,27 @@
...
+except ModuleNotFoundError as e:
+    logger.error(f"Exception: {e}")
+    logger.error('In order to use Whisper, you need to `uv add "pipecat-ai[whisper]"`.')
+    raise ImportError(f"Missing module: {e}") from e
+
+# MLX Whisper only runs on Apple Silicon. On other platforms the package is
+# unavailable (or installed but unloadable, e.g. a missing ``libmlx.so``), so
+# importing it would break this module everywhere else. Only attempt it on macOS;
+# WhisperSTTServiceMLX imports it lazily when actually used.
+if platform.system() == "Darwin":
```
markbackman: I think you need to check for `arm64`, too, to ensure that it's Apple Silicon.
```suggestion
if platform.system() == "Darwin" and platform.machine() == "arm64":
```

markbackman (approved):
Looks good, just one small thing.

src/pipecat/services/whisper/stt.py
aconchillo: Good catch. Fixed!

PR #4686, opened 2026-06-10
Vonage WebRTC transport additions and dependency update

asilvestre:

Adding two missing features to the Vonage WebRTC transport:
* Captions support
* Individual audio streams subscription

Also we now point to the newest video connector library version v1.0.2

There are documentation updates for this PR in this docs PR https://github.com/pipecat-ai/docs/pull/888


src/pipecat/transports/vonage/client.py
```diff
@@ -85,10 +91,12 @@ class VonageVideoConnectorTransportParams(TransportParams):
         session_enable_migration: Whether to enable session migration.
         audio_in_auto_subscribe: Whether to automatically subscribe to audio streams.
         video_in_auto_subscribe: Whether to automatically subscribe to video streams.
+        captions_in_auto_subscribe: Whether to automatically subscribe to captions streams.
```
markbackman, line 94: I'm assuming "captions" is consistent with your product naming. "Transcription" is more common in the Pipecat and voice AI world, but you should select something developers familiar with Vonage will understand.

src/pipecat/transports/vonage/client.py
```diff
@@ -85,10 +91,12 @@ class VonageVideoConnectorTransportParams(TransportParams):
...
         video_in_preferred_width: Preferred width for video input capture.
         video_in_preferred_height: Preferred height for video input capture.
         video_in_preferred_framerate: Preferred framerate for video input capture.
         clear_buffers_on_interruption: Whether to clear media buffers when an interruption frame is received.
+        captions_in_enabled: Whether to enable captions input.
```
markbackman, line 99: Same here, re "captions".

markbackman (commented):
Do you also want to update the Vonage transport example to show how captions can be used in place of an STT service?
https://github.com/pipecat-ai/pipecat/blob/main/examples/transports/transports-vonage.py

asilvestre:
> Do you also want to update the Vonage transport example to show how captions can be used in place of an STT service? https://github.com/pipecat-ai/pipecat/blob/main/examples/transports/transports-vonage.py

Thanks for the review @markbackman! If it's ok we'd prefer to keep the example simple as it is, enabling the captions service involves a more complex setup, including enabling captions at the vonage session level as well as in each media publisher.

src/pipecat/transports/vonage/client.py
asilvestre: Exactly, if it's ok to keep it this way, "captions" matches our product naming https://developer.vonage.com/en/video/guides/live-caption

markbackman: 👍

markbackman:
> > Do you also want to update the Vonage transport example to show how captions can be used in place of an STT service? https://github.com/pipecat-ai/pipecat/blob/main/examples/transports/transports-vonage.py
> 
> Thanks for the review @markbackman! If it's ok we'd prefer to keep the example simple as it is, enabling the captions service involves a more complex setup, including enabling captions at the vonage session level as well as in each media publisher.

That works! I was just offering you a chance to showcase how to use the feature. But, if this accomplishes your goal, then that's great.

markbackman (approved):
LGTM!

PR #4690, opened 2026-06-11
Enable RTVI on LLMWorker unless it is bridged

aconchillo:
## Summary

`LLMWorker` passed `enable_rtvi=False` to `PipelineWorker` unconditionally, so a standalone `LLMWorker` never got the worker's automatic RTVI support. This change makes it conditional: `enable_rtvi=bridged is None`.

- **Standalone `LLMWorker`** (`bridged=None`): can be the bot's only worker, used directly for its `@tool` collection with a transport in its pipeline. It is client-facing, so the automatic `RTVIProcessor`/`RTVIObserver` pair is now added as for any `PipelineWorker`.
- **Bridged child worker** (`bridged=()` or bridge names): RTVI stays disabled. The transport worker owns the client connection and its RTVI machinery translates the bridged frames; the child generating its own RTVI messages would duplicate them.

## Test plan

- [x] `uv run pytest tests/test_llm_worker.py tests/test_base_worker.py` — 67 passed
- [x] `uv run ruff check` clean


markbackman (approved):
LGTM!

PR #4691, opened 2026-06-11
Deactivate immediately in activate_worker(deactivate_self=True)

aconchillo:
## Summary

Fixes the worker-handoff race behind #4628 (also analyzed in #4641): `deactivate_worker` and `activate_worker` publish fire-and-forget bus messages, and the caller's `active` flag only flipped when its **own** deactivate message came back over the bus. The handoff target could therefore activate first, leaving both workers briefly `active=True`. During that window each re-broadcasts frames originally emitted by the other, so the host `BusBridgeProcessor` receives the same frame twice — duplicate `assistant.tool_calls` + `tool` result pairs in the LLM context (observed as a 90s stall and a hallucinated `end_call`).

## Fix

In `activate_worker(deactivate_self=True)`, set `self._active = False` synchronously **before** publishing. The bus is ordered, so the activate message can never reach the target while this worker still accepts bus frames — the overlap cannot exist. The deactivate message is still sent, so `on_deactivated` hooks and observers see the handoff exactly as before (`_handle_worker_deactivate` is idempotent on the flag).

Compared with deduplicating deliveries at the receiver (#4641), this closes the window itself, which also prevents the case dedup can't catch: two simultaneously-active siblings each independently responding to the same input with *distinct* frames. (Note for #4641's approach: `Frame.id` is regenerated on deserialization — `field(init=False)` + `__post_init__` — so id-keyed dedup is inert over the network buses; it only matches on the in-process `AsyncQueueBus`.)

One boundary-semantics change: frames arriving during the flip gap are now accepted by **neither** worker (previously: briefly by both). For a handoff that is the conventional behavior.

## Test plan

- [x] New `test_deactivate_self_takes_effect_before_activate_is_sent`: asserts the flag flips synchronously and both bus messages are published with the worker already inactive
- [x] `uv run pytest tests/test_base_worker.py` — 59 passed
- [x] `uv run ruff check` clean


markbackman (approved):
LGTM

aconchillo:
I'll wait for @Maurotb to confirm before merging.

Maurotb:
@aconchillo Tested #4691 — clean first sub-bot context (1 user + 1 assistant `tool_call` + 1 `tool` result), no duplicate pairs, no 90s stall. Closing the overlap at the source is the right shape.

Agreed on #4641 — `Frame.id` is `field(init=False)`, regenerated in `__post_init__`, so the LRU only works on the in-process `AsyncQueueBus`. Closing it as superseded.

Thanks!

aconchillo:
> @aconchillo Tested #4691 — clean first sub-bot context (1 user + 1 assistant `tool_call` + 1 `tool` result), no duplicate pairs, no 90s stall. Closing the overlap at the source is the right shape.
> 
> Agreed on #4641 — `Frame.id` is `field(init=False)`, regenerated in `__post_init__`, so the LRU only works on the in-process `AsyncQueueBus`. Closing it as superseded.
> 
> Thanks!

Thank you for testing!

PR #4692, opened 2026-06-11
Remove sleep(0) workarounds obsoleted by the task-manager fix

aconchillo:
## Summary

Since #4649, `TaskManager.create_task()` closes a never-started inner coroutine from a done callback, so cancelling a task before its wrapper runs no longer emits `RuntimeWarning: coroutine '...' was never awaited`. That makes the `await asyncio.sleep(0)` calls whose only purpose was to let a just-created timer task start before a possible immediate cancel unnecessary. #4649 already removed the one in `LLMService._run_function_call`; this removes the remaining eight:

- `processors/aggregators/llm_response_universal.py` (realtime handoff flush task)
- `services/google/gemini_live/llm.py` (transcription timeout task)
- `turns/user_idle_controller.py` (idle timer)
- `turns/user_turn_completion_mixin.py` (incomplete timeout)
- `turns/user_stop/speech_timeout_user_turn_stop_strategy.py` (two sites: STT wait + user speech timeout)
- `turns/user_stop/turn_analyzer_user_turn_stop_strategy.py` (two sites: timeout handler)

Several of the removed comments stated the purpose verbatim ("which would leave its coroutine un-awaited", "emits a 'coroutine was never awaited' warning").

## Deliberately kept

The remaining `asyncio.sleep(0)` calls serve different purposes and are untouched:

- `transports/base_output.py`: prevents a 100% CPU spin in the audio loop under cancellation.
- `bus/bus.py` (x2), `bus/network/redis.py`, `bus/network/pgmq.py`, `workers/runner.py`, `workers/proxy/websocket/{client,server}.py`: start-up yields so reader/dispatch loops begin promptly — #4649 changes nothing about when tasks start, only how never-started ones are cleaned up.

## Test plan

- [x] `tests/test_context_aggregators_universal.py`, `test_llm_response.py`, `test_task_manager.py`, `test_idle_frame_processor.py` — 84 passed
- [x] All six turn-strategy test files (completion mixin, controller, processor, start/stop strategies, wake phrase) — 72 passed
- [x] `tests/test_user_turn_stop_strategy.py` with `-W error::RuntimeWarning` — 24 passed, no never-awaited warnings
- [x] `ruff check` clean

### DIFF
--- a/changelog/4692.changed.md
+++ b/changelog/4692.changed.md
@@ -1,0 +1,1 @@
+- Removed the `asyncio.sleep(0)` workarounds that let a just-created timer task start before a possible immediate cancellation. `TaskManager.create_task()` now cleans up never-started coroutines centrally, so the yields served no purpose.
--- a/src/pipecat/processors/aggregators/llm_response_universal.py
+++ b/src/pipecat/processors/aggregators/llm_response_universal.py
@@ -997,11 +997,6 @@ class LLMUserAggregator(LLMContextAggregator):
             self._realtime_deferred_handoff_flush(),
             name=f"{self}::realtime_handoff_flush",
         )
-        # Yield so the task's wrapper coroutine starts running before
-        # any immediate cancellation by the failsafe path — otherwise
-        # asyncio GCs the inner coroutine without ever entering it and
-        # emits a "coroutine was never awaited" warning.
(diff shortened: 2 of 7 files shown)


markbackman (approved):
Nice! 🎉

PR #4689, opened 2026-06-10
feat(cli): scaffold the eval transport in `pipecat init` via --eval

markbackman:
## Summary

Adds an opt-in `--eval` / `--no-eval` flag (and an "Enable evals?" wizard question, default **No**) to `pipecat init`. When enabled, the generated bot gets an inert `"eval"` entry in `transport_params` so it's runnable with `-t eval` for behavioral evals — removing the previously-manual step.

The entry mirrors the bot's audio/video flags and is only used when the bot is run with `-t eval`. The required `WebsocketServerParams` import is registered as an `"eval"` feature in `FEATURE_DEFINITIONS` (regenerating `_imports.py`), so it flows through the same registry path as the other feature imports rather than being injected ad hoc.

## Base

Targets `aleix/evals`, which already carries the eval framework, runner `-t eval` support, and the ported init CLI. This branch is just the two init-side commits cherry-picked cleanly on top.

## Changes

- `--eval`/`--no-eval` flag + wizard question on `pipecat init`
- `"eval"` feature in the CLI service registry → imports `WebsocketServerParams`
- `enable_eval` block in the cascade/realtime bot templates emitting the `"eval"` transport entry
- Test coverage in `tests/cli/test_project_generation.py`

## Test plan

- `uv run --extra cli pytest tests/cli/test_project_generation.py` → 43 passed

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #4693, opened 2026-06-11
fix(cli): make --eval tri-state so explicit flags beat --config values

jamsea:
## Summary

Follow-up to #4689, reworking the surviving piece of pipecat-ai/pipecat-cli#96 per @markbackman's feedback there (the scenario scaffolding was dropped; the eval transport scaffolding already landed in #4689).

With `enable_eval` declared as a plain `bool` defaulting to `False`, an explicit `--no-eval` is indistinguishable from an omitted flag. The or-merge in `init_command` then lets `"enable_eval": true` in a `--config` file silently override an explicit `--no-eval`:

```bash
pipecat init --config saved.json --no-eval   # before: evals still enabled
```

This matters because `--dry-run` output (which includes `enable_eval`) is designed to be fed back via `--config`.

## Fix

- `--eval/--no-eval` now defaults to `None` (tri-state)
- The config-file value only applies when the flag was omitted
- The resolved value is coerced back to a real `bool` before `validate_and_build_config`, so `ProjectConfig`/JSON never see `None`

## Testing

- New dry-run precedence tests in `tests/cli/test_init_command.py`: `--no-eval` beats a true file value, `--eval` beats a false file value, file value applies when the flag is omitted, and everything defaults off
- `uv run pytest tests/cli/`: 273 passed

### DIFF
--- a/changelog/4693.fixed.md
+++ b/changelog/4693.fixed.md
@@ -1,0 +1,1 @@
+- Fixed `pipecat init` ignoring an explicit `--no-eval` when the `--config` file sets `"enable_eval": true`. The flag is now tri-state: an explicit `--eval`/`--no-eval` always wins, and the file value only applies when the flag is omitted.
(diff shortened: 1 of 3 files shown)


markbackman:
Nice catch. This turns out to be a general problem. I'm going to fix in a more general way. Closing this in favor of: #4698

PR #4697, opened 2026-06-11
fix(cli): env.example comments are parsed as values by python-dotenv

### LINKED ISSUE #4696: pipecat init: env.example comments are read as values, bot fails on first run
Hey folks

While testing a fresh `pipecat init` project (cascade — Deepgram STT, OpenAI Responses LLM, Cartesia TTS) I hit this on the very first `uv run bot.py`:

```
WebSocket API error: The requested model '# Optional: defaults to "gpt-4.1"' does not exist.
```

Took me a while to realize the "model name" was literally the comment from `env.example`. Cartesia TTS failed the same way (`voice ID must be a valid UUID`).

The generated `env.example` puts hints on the value side of empty optional vars:

```
OPENAI_MODEL= # Optional: defaults to "gpt-4.1"
CARTESIA_VOICE_ID= # Optional: defaults to "71a7ad14-..."
```

python-dotenv only strips an inline comment when there's a real value before it — with an empty value the whole `#…` becomes the value. And the generated code uses `os.getenv("OPENAI_MODEL", "gpt-4.1")`, whose default only applies when the key is fully absent, not when it's empty. So leaving it blank doesn't help either.

It's not just OpenAI/Cartesia — same pattern on ~33 optional vars across the providers.

**Repro:** `pipecat init` (cascade, OpenAI + Cartesia) → `cp env.example .env` → fill the API keys, leave `OPENAI_MODEL`/`CARTESIA_VOICE_ID` empty as the comments suggest → `uv run bot.py`.

Happy to send a PR — I've got one ready that comments the optional vars out so the defaults apply. (Alternative would be hardening the generated code to `os.getenv("VAR") or "default"`.)

picsoung:
Hey folks

As I was running through the quickstart (a `pipecat init` cascade project with OpenAI + Cartesia), the bot died on the first run with:

```
WebSocket API error: The requested model '# Optional: defaults to "gpt-4.1"' does not exist.
```

Turns out the model name was literally the comment from `env.example`. The generated file puts hints on the value side of empty optional vars:

```
OPENAI_MODEL= # Optional: defaults to "gpt-4.1"
```

python-dotenv only strips an inline comment when a real value precedes it; with an empty value the `#…` becomes the value, so `os.getenv("OPENAI_MODEL")` returns the comment string. The default in `os.getenv("OPENAI_MODEL", "gpt-4.1")` doesn't save it either — that only applies when the key is fully absent. Same trap on ~33 optional vars across providers (Cartesia voice, Anthropic/Google/Groq/Mistral/xAI models, etc.).

This PR comments the optional vars out and moves each hint to its own line, so the var is genuinely unset and the existing `os.getenv(..., default)` fallbacks work:

```
# Optional: defaults to "gpt-4.1"
# OPENAI_MODEL=
```

One file changed, no code changes. Confirmed locally that `OPENAI_MODEL` is now absent and the bot uses `gpt-4.1`.

Closes #4696.

### DIFF
--- a/changelog/4697.fixed.md
+++ b/changelog/4697.fixed.md
@@ -1,0 +1,1 @@
+- Fixed `pipecat init` generating an `env.example` where inline comments on optional variables (e.g. `OPENAI_MODEL`, `CARTESIA_VOICE_ID`) were parsed by python-dotenv as the variable's value, causing the generated bot to fail on startup. Comments now sit on their own line.
--- a/src/pipecat/cli/templates/server/env.example.jinja2
+++ b/src/pipecat/cli/templates/server/env.example.jinja2
@@ -5,7 +5,8 @@
 # Daily (Transport)
 DAILY_API_KEY=
 {% if 'daily' in transports %}
-DAILY_ROOM_URL=  # Optional: leave empty to create a new room
+# Optional: leave empty to create a new room
+# DAILY_ROOM_URL=
 {% endif %}
 {% if 'daily_pstn_dialout' in transports or 'twilio_daily_sip_dialin' in transports or 'twilio_daily_sip_dialout' in transports %}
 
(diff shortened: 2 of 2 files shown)


markbackman:
Thanks for raising and offering a PR. I think we want to go about this slightly differently. These values _are_ required in many cases, so we can't mark the values as optional in the .env.example. The CLI takes the opinion that the LLM model and TTS voice will be set per the developer's choice, so it's wired up as such. This also makes code gen more uniform. I'd prefer to keep this pattern.

Really appreciate you flagging this. Closing in favor of #4699.

PR #4698, opened 2026-06-11
fix(cli): explicit init flags always beat --config file values

markbackman:
## Summary

Supersedes #4693 with a general fix. `pipecat init` merged `--config` file values with CLI flags using `flag or file_data.get(...)`, which can't distinguish an omitted flag from an explicit `--no-<flag>`: a plain `bool` default of `False` looks identical either way. So a config file's `"enable_eval": true` silently overrode an explicit `--no-eval` — and the same bug affected **every** negatable boolean (`recording`, `transcription`, `video_input`, `video_output`, `enable_krisp`, `observability`).

This matters because `--dry-run` emits a resolved config that's designed to be fed back via `--config` (config-as-template / round-trip), so overriding one knob on the command line is a normal workflow.

#4693 fixed only `enable_eval` by making it tri-state (`bool | None`). Rather than tri-state each flag one at a time, this resolves the precedence rule **generally**.

## Fix

- Add `ctx: typer.Context` to `init_command` and ask Click via `ctx.get_parameter_source()` whether each parameter actually came from the command line, instead of inferring it from the value.
- Replace the ~30-line `or`-merge block with a single `pick(value, param, *file_keys)` helper applied uniformly to all settings: **explicit flag wins; else first file key present; else the flag default**. `enable_eval` stays a plain `bool = False` (no tri-state needed).
- Compare the parameter source by enum member *name* (`.name == "COMMANDLINE"`): Typer vendors its own copy of Click, so the returned `ParameterSource` is a different enum object than `click.core.ParameterSource` and an identity/`==` comparison silently fails.

## Testing

- New parametrized precedence tests in `tests/cli/test_init_command.py`: `--no-*` beats a true file value across `enable_eval`/`recording`/`transcription`/`observability`, `--no-deploy-to-cloud` overrides the True-default case, `--enable-krisp` wins over a false file value, and the file value still applies when no flag is passed.
- `uv run pytest tests/cli/`: 276 passed.
- `uv run ruff check` / `ruff format --check`: clean.

### DIFF
--- a/changelog/4698.fixed.md
+++ b/changelog/4698.fixed.md
@@ -1,0 +1,1 @@
+- Fixed `pipecat init` ignoring explicit `--no-*` flags (e.g. `--no-eval`, `--no-recording`, `--no-transcription`) when a `--config` file enabled the same setting. Explicit CLI flags now always take precedence over config-file values, for every setting — so a saved/`--dry-run` config can be reused as a template and overridden one flag at a time.
--- a/src/pipecat/cli/commands/init.py
+++ b/src/pipecat/cli/commands/init.py
@@ -44,6 +44,7 @@ def _list_options_callback(value: bool):
 
 
 def init_command(
+    ctx: typer.Context,
     target: str | None = typer.Argument(
         None,
         help="Target directory; use '.' to scaffold into the current directory "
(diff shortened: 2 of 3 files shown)


jamsea (approved):
Way better, looks like we still need humans!

PR #4699, opened 2026-06-11
fix(cli): remove inline comments from generated env.example

### LINKED ISSUE #4696: pipecat init: env.example comments are read as values, bot fails on first run

markbackman:
Remove inline comments so that error messages are less confusing. Many models require values for things like models and voices, which need to be provided via an .env file.

Also, comments were prone to getting stale, so removing where it mentions models or voices.

Also, removed two non-existent values, `OPENAI_INSTRUCTIONS` and `XAI_INSTRUCTIONS`.

## Summary

- Fixed `pipecat init` generating an `env.example` with inline comments after `KEY=` (e.g. `OPENAI_VOICE_ID= # "alloy"`), which many dotenv loaders parse as part of the value rather than as a comment.
- Moved explanatory comments (optional/credentials hints) onto their own lines above each entry.
- Dropped the example-value hints that were previously tacked on as inline comments.

## Testing

- Run `pipecat init` and inspect the generated `.env.example`: comments render on their own lines and no inline comments remain after `KEY=`.

Fixes #4696
Supersedes #4697 

### DIFF
--- a/changelog/4699.fixed.md
+++ b/changelog/4699.fixed.md
@@ -1,0 +1,1 @@
+- Fixed `pipecat init` generating an `env.example` with inline comments (e.g. `OPENAI_VOICE_ID= # "alloy"`) that many dotenv loaders parse as part of the value. Comments now sit on their own lines above each entry.
--- a/src/pipecat/cli/templates/server/env.example.jinja2
+++ b/src/pipecat/cli/templates/server/env.example.jinja2
@@ -5,7 +5,8 @@
 # Daily (Transport)
 DAILY_API_KEY=
 {% if 'daily' in transports %}
-DAILY_ROOM_URL=  # Optional: leave empty to create a new room
+# Optional: leave empty to create a new room
+# DAILY_ROOM_URL=
 {% endif %}
 {% if 'daily_pstn_dialout' in transports or 'twilio_daily_sip_dialin' in transports or 'twilio_daily_sip_dialout' in transports %}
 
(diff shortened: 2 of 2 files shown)


jamsea (approved):

picsoung:
Thank you @markbackman and team 🙏

PR #4702, opened 2026-06-11
Set Moonshine as the default STT provider for evals

markbackman:

Making moonshine the default STT service for evals instead of whisper, based on Aleix's findings.


aconchillo (approved):

aconchillo:
Thanks! I thought I had done this.

PR #4703, opened 2026-06-11
fix(azure): wake the event loop from TTS SDK callbacks

markbackman:
## Summary

- Fixes `AzureTTSService` producing **no audio** in pipelines without an output transport (e.g. headless/offline setups). The Speech SDK delivers audio via callbacks on **native threads** that do a cross-thread `asyncio.Queue.put_nowait()`; that schedules the awaiting getter's wake-up with `call_soon()`, which doesn't wake an idle selector — so on a quiet loop the audio sat unread until the stop-frame timeout fired. A transport normally keeps the loop busy, which masks this.
- Capture the event loop in `start()` and marshal every SDK-thread queue put (audio **and** word-boundary) through `loop.call_soon_threadsafe()`. A small `_put_threadsafe()` helper falls back to a direct put before the loop is captured and ignores a closed loop during shutdown. Transport-driven pipelines are unaffected (the loop is already awake).
- Adds `tests/test_azure_tts.py` — deterministic regression tests that park a getter on an idle loop *first*, then fire each callback (audio, cancel-error, completion sentinel) from a real thread, plus the pre-`start()` fallback.

## Testing

- `uv run pytest tests/test_azure_tts.py` — all pass with the fix; verified the idle-loop tests fail with `TimeoutError` against the pre-fix code, confirming they catch the regression.


src/pipecat/services/azure/tts.py
```diff
@@ -506,7 +509,7 @@ def _handle_word_boundary(self, evt):
     def _emit_pending_word(self):
         """Emit the currently buffered word if one exists."""
         if self._last_word is not None:
-            self._word_boundary_queue.put_nowait((self._last_word, self._last_timestamp))
+            self._put_threadsafe(self._word_boundary_queue, (self._last_word, self._last_timestamp))
```
aconchillo: I think we usually do, we can use `.put` no need for `put_nowait`:

```
asyncio.run_coroutine_threadsafe(
    self._word_boundary_queue.put((self._last_word, self._last_timestamp)),
    self.get_event_loop()
)
```

markbackman: Updated.

aconchillo (approved):

PR #4704, opened 2026-06-11
Allow adding origin restrictions to WebSocket transports and the runner.

filipi87:
## Summary
  
  - Added origin restriction support to `WebsocketServerTransport`, `FastAPIWebsocketTransport`, and the development runner to mitigate Cross-Site WebSocket Hijacking (CSWSH)
  - New allowed_origins: list[str] field on `WebsocketServerParams` and `FastAPIWebsocketParams`; defaults to the
  `PIPECAT_ALLOWED_ORIGINS` env var (comma-separated)
  - `FastAPIWebsocketTransport` raises `ValueError` at construction if the connection's Origin header is missing or disallowed
  - Runner gains `--allowed-origins` CLI flag backed by `PIPECAT_ALLOWED_ORIGINS`; a single env var now covers all WebSocket transports uniformly
  - Startup message now always shows WebSocket security status (auth mode and origin restriction)
  - Default is an empty list (allow all origins) — no behavior change for existing deployments

## Testing

  Set the env var and test that connections from unlisted origins are rejected:

```
  PIPECAT_ALLOWED_ORIGINS=https://example.com uv run python your_bot.py
  # Connection from an unlisted origin should receive close code 4003
```

  Or pass the flag directly to the runner:

  `uv run python -m pipecat.runner.run --allowed-origins https://example.com`

### DIFF
--- a/src/pipecat/runner/run.py
+++ b/src/pipecat/runner/run.py
@@ -314,14 +314,14 @@ def _extract_ws_token(websocket) -> str | None:
     return websocket.query_params.get("token")
 
 
-def _print_ws_security_status(args: argparse.Namespace):
-    """Print WebSocket security status lines (auth + origin restriction)."""
+def _print_security_status(args: argparse.Namespace):
+    """Print security status lines (auth + origin restriction)."""
     if args.ws_auth == "token":
-        print("   → WebSocket auth:    token (HMAC, call /start to obtain a token)")
+        print("   → WebSocket auth:  token (HMAC, call /start to obtain a token)")
     if args.allowed_origins:
-        print(f"   → WebSocket origins: {', '.join(args.allowed_origins)}")
+        print(f"   → Allowed origins: {', '.join(args.allowed_origins)}")
     else:
-        print("   → WebSocket origins: all (no restriction)")
+        print("   → Allowed origins: all (no restriction)")
 
 
 def _print_startup_message(args: argparse.Namespace):
@@ -334,7 +334,7 @@ def _print_startup_message(args: argparse.Namespace):
         print(f"   → Enabled transports: {_format_transport_status(enabled)}")
         if disabled:
             print(f"   → Disabled transports: {_format_transport_status(disabled)}")
-        _print_ws_security_status(args)
+        _print_security_status(args)
     elif args.transport == "webrtc":
         if args.esp32:
             print("🚀 Bot ready! (ESP32 mode)")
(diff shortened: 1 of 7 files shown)


src/pipecat/runner/run.py
```diff
@@ -1247,6 +1262,11 @@ async def _handle_telephony_ws(websocket: WebSocket, path_token: str | None = No
                 logger.warning("WebSocket connection rejected: invalid or missing token")
                 await websocket.close(code=4003)
                 return
+        origin = websocket.headers.get("origin", "")
+        if not is_origin_allowed(origin, args.allowed_origins):
+            logger.warning(f"WebSocket connection rejected: origin '{origin}' not allowed")
+            await websocket.close(code=4003)
+            return
```
aconchillo, line 1269: why do we need this and `is_origin_allowed`? shouldn't the websocket implementation do that already? That is, shouldn't we just set `allowed_origins` in `FastAPIWebsocketParams` in `create_transport`?

src/pipecat/utils/security/allowed_origins.py
```diff
@@ -0,0 +1,33 @@
...
+
+    Args:
+        origin: The value of the ``Origin`` header, or an empty string if absent.
+        allowed_origins: List of allowed origin strings. An empty list allows
+            all origins. When non-empty, a missing or disallowed origin is
+            rejected.
+    """
+    if not allowed_origins:
+        return True
+    return origin.lower() in {o.lower() for o in allowed_origins}
```
aconchillo, line 33: I don't think we need this function. See other comment.

src/pipecat/runner/run.py
filipi87: https://websocket.org/guides/troubleshooting/cors/

The server sees the Origin header and can choose to reject the connection — but if it responds with 101 Switching Protocols, the browser accepts it regardless of what domain the page is on. No Access-Control-Allow-Origin needed. No preflight. The connection is open.

This is by design. RFC 6455 delegates origin checking to the server. The browser sends the Origin header so the server has the information to make a decision, but enforcement is entirely server-side.

So, for FastAPI, we need to check if the origin is allowed or not.

filipi87: We are also using this function as a safeguard inside our `FastAPITransport`, so even if the user is using a custom runner that accepts a WebSocket connection it shouldn't, we will throw an error when creating the transport.

aconchillo: Got it!

aconchillo (approved):

aconchillo:
LGTM! Just the unit tests not passing.

PR #4705, opened 2026-06-11
Rename task frames to worker frames

aconchillo:
Renames the task frame family to worker frames, matching the `PipelineTask` to `PipelineWorker` rename: `TaskFrame` becomes `WorkerFrame`, `TaskSystemFrame` becomes `WorkerSystemFrame`, `EndTaskFrame` becomes `EndWorkerFrame`, `StopTaskFrame` becomes `StopWorkerFrame`, `CancelTaskFrame` becomes `CancelWorkerFrame` and `InterruptionTaskFrame` becomes `InterruptionWorkerFrame`.

The old names remain as deprecated aliases (`.. deprecated:: 1.4.0`). The leaf aliases warn with `DeprecationWarning` on construction and inherit from both the new class and the old base, so `isinstance` checks against either the new or old names keep working for frames constructed with the old names. The base aliases (`TaskFrame`, `TaskSystemFrame`) are docstring-only deprecations so user subclasses don't double-warn.

Worker frames are now pushed downstream by default with a plain `push_frame(frame)`, no direction argument needed. For `EndWorkerFrame` and `StopWorkerFrame` (control frames) this means frames queued ahead of them are flushed before the worker acts; `CancelWorkerFrame` and `InterruptionWorkerFrame` are system frames and remain immediate in either direction. Pushing upstream still works, since the pipeline worker handles these frames at both ends. All in-repo push sites (RTVI processor, WebSocket server transport, eval serializer and examples) now use the downstream default, and the WebSocket server's upstream special case for client-initiated worker frames is removed.


markbackman (approved):
LGTM!

PR #4708, opened 2026-06-11
fix(text-filter): strip markdown headers before newline collapse

YFortin:
## Summary

When text containing markdown headers is streamed word-by-word through `MarkdownTextFilter`, second-level (and deeper) headers are not properly stripped — the raw `##` markers leak into the TTS output.

## Example

**Input text:**
```
# Title

## Subtitle
```

**Before this fix:**
```
Title
 ## Subtitle 
```
The first `# Title` is correctly recognized (it starts at column 0), but the blank line between the headers gets collapsed into a space by the newline-to-space substitution. This prepends a `§` (space marker) before `## Subtitle`, so `md.convert()` no longer recognizes it as a header and leaves the raw `##` in the output.

**After this fix:**
```
Title
Subtitle 
```
Headers are stripped before the newline collapse runs, so all header levels are properly removed from the plain-text output.

## Changes

- `markdown_text_filter.py`: Added two `re.sub()` calls at the top of `filter()` to strip header markers (and preceding blank lines) before the existing newline collapse logic.
- `test_markdown_text_filter.py`: Updated `test_basic_markdown_removal` to include a header, and added a dedicated `test_header_stripping` test case.

### DIFF
--- a/src/pipecat/utils/text/markdown_text_filter.py
+++ b/src/pipecat/utils/text/markdown_text_filter.py
@@ -77,11 +77,23 @@ class MarkdownTextFilter(BaseTextFilter):
             The filtered text with Markdown formatting removed or converted.
         """
         if self._settings.enable_text_filter:
-            # Strip markdown header markers along with any preceding blank lines
-            # (headers require # at column 0, but the newline collapse below
-            # would inject a space that breaks header recognition by md.convert)
-            filtered_text = re.sub(r"(\n\s*\n)(#{1,6}\s+)", r"\n", text, flags=re.MULTILINE)
-            filtered_text = re.sub(r"^#{1,6}\s+", "", filtered_text, flags=re.MULTILINE)
+            # Strip markdown headers to plain text before the newline collapse
+            # below. md.convert can't recognize a header once the collapse
+            # injects a leading space, so we normalize headers ourselves.
+            # First, drop blank lines before a header (keeping the marker) so
+            # the collapse doesn't turn them into a leading space.
+            filtered_text = re.sub(r"\n\s*\n(#{1,6}[ \t])", r"\n\1", text)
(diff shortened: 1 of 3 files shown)


src/pipecat/utils/text/markdown_text_filter.py
```diff
@@ -77,8 +77,14 @@ async def filter(self, text: str) -> str:
             The filtered text with Markdown formatting removed or converted.
         """
         if self._settings.enable_text_filter:
+            # Strip markdown header markers along with any preceding blank lines
+            # (headers require # at column 0, but the newline collapse below
+            # would inject a space that breaks header recognition by md.convert)
+            filtered_text = re.sub(r"(\n\s*\n)(#{1,6}\s+)", r"\n", text, flags=re.MULTILINE)
+            filtered_text = re.sub(r"^#{1,6}\s+", "", filtered_text, flags=re.MULTILINE)
```
YFortin: Could also be made in a single regex, at the preference of the reviewers:

```python
            filtered_text = re.sub(
                r"(\n\s*\n)(?=#{1,6}\s+)|^#{1,6}\s+",
                lambda m: "\n" if m.group(1) else "",
                text,
                flags=re.MULTILINE,
            )
```

markbackman:
@YFortin thanks for posting this. The original solution worked to fix the issue but introduced a regression for closed ATX headers (## Subtitle ##). It's been a long time since I wrote this, but I recall playing whack-a-mole with the use cases trying to satisfy all cases.

I paired with Claude to come up with a solution that handles both open and closed headers.

I added regression tests to catch both cases in the future.

Thanks for raising this. Let me know if there are any issues after this.

markbackman (approved):
Thanks! 🙌

PR #4709, opened 2026-06-12
Auto-register handlers bundled on a FunctionSchema (and more!)

kompfner:

Builds on #4654 (auto-registration of direct functions). A `FunctionSchema` can now carry the `handler` that runs when the LLM calls it — auto-registered just like a direct function's — and that handler's call options can be set with the same `@tool_options` decorator direct functions use.

```python
from pipecat.adapters.schemas.direct_function import tool_options
from pipecat.services.llm_service import FunctionCallParams


# Call options live on the handler, via the same decorator direct functions use.
@tool_options(cancel_on_interruption=False, timeout_secs=120)
async def generate_report(params: FunctionCallParams):
    report_type = params.arguments["report_type"]
    await params.result_callback({"url": f"https://reports.example.daily.co/{report_type}.pdf"})


schema = FunctionSchema(
    name="generate_report",
    description="Generate a report and return a link when it's ready.",
    properties={
        "report_type": {
            "type": "string",
            "enum": ["sales", "inventory", "usage"],
            "description": "Which report to generate.",
        },
    },
    required=["report_type"],
    handler=generate_report,  # ← new: the schema carries its own handler
)

# Listing the schema is all the wiring needed — no register_function() call.
context = LLMContext(tools=[schema])
```

### What changed

- **`FunctionSchema` gains an optional `handler`.** When set, the LLM service registers it automatically wherever the schema is advertised in an `LLMContext` (or via an `LLMSetToolsFrame`) — no separate `register_function` call. A schema without a handler stays advertise-only, exactly as before.

- **One mechanism, two handler-carrying tool kinds.** The sync that keeps registered handlers in step with the advertised tool set now covers both direct functions and handler-carrying `FunctionSchema`s. The advertised set remains the single source of truth: dropping a handler-carrying schema unregisters its handler too. A schema handler registers via `register_function`, so there's no parallel registration path.

- **Call options via `@tool_options`.** A schema handler's `cancel_on_interruption`/`timeout_secs` come from the `@tool_options` decorator on the handler — the same mechanism direct functions use. To make this work without a parallel path, `register_function` itself now reads `@tool_options` off the handler when options aren't passed explicitly (precedence: explicit argument > `@tool_options` > default); previously that decorator was silently ignored on the `register_function` path.

- **A teaching nudge.** When a tool is registered via `register_function` yet its advertised schema already carries a handler, the service logs a one-time warning that the manual registration is redundant — modeled on the existing direct-function deprecation guidance.

- **Rename.** `_sync_registered_direct_functions` → `_sync_registered_tool_handlers` (and the two helpers), across the base service, `LLMSwitcher`, and the OpenAI/xAI/Inworld realtime services, since they now cover more than direct functions.

### Also fixes a #4654 regression: "zombie" tool re-registration

Making the advertised tool set the source of truth (#4654) broke a pre-existing workflow: unregistering a tool's handler on its own — via `unregister_function` / `unregister_direct_function`, without changing the advertised set — was silently undone, because the next `LLMContextFrame` re-registered the handler from the still-advertised tool.

Now an explicitly unregistered handler is tracked and **not** auto-re-registered while its tool stays advertised, so a standalone unregister sticks and calls hit the missing-handler recovery path. It's restored only by registering it again, or by re-advertising the tool (removing it from the advertised set, then adding it back). This is orthogonal to `FunctionCallRegistryItem.auto_registered`, which flags a *present*, prunable entry; the new set tracks names that are *absent by choice*.

### Example

`examples/function-calling/function-calling-schema-handler.py` shows a hand-written `FunctionSchema` (with a `format` enum) that bundles its handler — explicit schema control *and* automatic wiring, with no `register_function` call — and notes how `@tool_options` overrides the handler's call options.

### Testing

New tests in `tests/test_auto_register_tools.py` cover registration, default and `@tool_options`-resolved options (via both the schema path and a direct `register_function` call), advertise-only coexistence, pruning, idempotency, reserved-name rejection, and the redundant-registration warning (fire-once and no-false-positive cases). For the regression fix, a flipped `test_unregister_of_still_advertised_direct_function_sticks` plus new tests assert that a standalone unregister stays gone across context frames, that explicit re-registration overrides it, and that de-advertising then re-advertising restores it. Existing direct-function, realtime-sync, and switcher tests pass under the renamed methods.


kompfner:
Tagged a few reviewers, but honestly @markbackman is probably the right one as he was the author of #4654, which this builds directly on top of.

src/pipecat/pipeline/llm_switcher.py
```diff
@@ -110,7 +110,7 @@ def register_function(
         function_name: str | None,
         handler: Any,
         *,
-        cancel_on_interruption: bool = True,
+        cancel_on_interruption: bool | None = None,
```
markbackman, line 113: Is the idea that the `True` value here was not actually doing much given that the `_resolve_tool_option()` was setting the default?

src/pipecat/services/llm_service.py
```diff
@@ -784,11 +784,16 @@ def register_function(
         function_name: str | None,
         handler: Any,
         *,
-        cancel_on_interruption: bool = True,
+        cancel_on_interruption: bool | None = None,
```
markbackman, line 793: Same question here about the default, regarding `_resolve_tool_option()`'s default.

markbackman (approved):
Very nice! I just asked one question for my own clarification. This looks like a really nice clean up.

Would you mind also updating the docs that I started here for this?
https://github.com/pipecat-ai/docs/pull/895

You can modify what's asserted about the FunctionSchema pattern, which should simplify the docs significantly.

changelog/4709.added.md
filipi87, line 1: Really cool. 💯

src/pipecat/adapters/schemas/function_schema.py
```diff
@@ -20,10 +23,19 @@ class FunctionSchema:
     Provides a structured way to define function tools used with AI models like OpenAI.
     This schema defines the function's name, description, parameter properties, and
     required parameters, following specifications required by AI service providers.
+
+    A schema may also carry the ``handler`` that services the function call. When
```
filipi87: nit: As a non native english speaker, this sentence was a little bit harder to understand what it actually means: "A schema may also carry the ``handler`` that services the function call."

Could we use maybe something more explicitly, like:
"A schema may also carry the ``handler``, the code that actually performs the work when the function is invoked."

Does it make sense ?

filipi87: Also, in the other places where we describe the handler, we use the expression "that services the function call." 

It might just be me, but I wonder if something more explicit would be easier for readers to understand.

src/pipecat/services/llm_service.py
filipi87: + 1 on this question.

src/pipecat/pipeline/llm_switcher.py
kompfner: This change is needed because `LLMSwitcher.register_function`'s signature should mirror `LLMService.register_functions`'s. The latter has changed due to the new `@tool_options` support—the same change that #4654 applied to `register_direct_function`.

kompfner: Oh wait! Just noticed that #4654 did not, in fact, update `LLMSwitcher.register_direct_function`'s signature to match `LLMService.register_direct_function`...was that intentional? If not, happy to make that fix as part of this PR. Otherwise, let's discuss.

src/pipecat/services/llm_service.py
kompfner: Same answer as above—implementing the same params fallback behavior as `LLMService.register_direct_function`.

src/pipecat/adapters/schemas/function_schema.py
kompfner: "Services" is just synonym for "handles", and since we're talking about `handler`s, I wonder if we even need it. I'll try to find a way of putting it that avoids use of "service", which is already a pretty important term in our codebase 😛

kompfner: Fixed: https://github.com/pipecat-ai/pipecat/pull/4709/changes/ecdd509345bd6bf7f4c8a3fb50e61d4271e92f90

src/pipecat/pipeline/llm_switcher.py
kompfner: Here's a fix for that: https://github.com/pipecat-ai/pipecat/pull/4709/changes/e10dccf145bb7c5921640f16030216ef12f2a259

src/pipecat/adapters/schemas/function_schema.py
filipi87: Yep. A lot better. Thank you @kompfner .

filipi87 (approved):
LGTM. 🚀

PR #4710, opened 2026-06-12
feat(cli): split scaffolding into `pipecat create` + repurpose `pipecat init` to make a project agent-ready

markbackman:
## Summary

Splits the CLI into two clear commands and builds out the artifacts the second one vends into a full coding-agent onboarding system for Pipecat app builders:

- **`pipecat create`** — the project scaffolder (renamed from `pipecat init`; behavior unchanged, incl. `pipecat create quickstart`).
- **`pipecat init`** — now makes a project *agent-ready* by writing a Pipecat coding-agent guide and developer guidance into it. A coding agent then picks up `AGENTS.md` automatically (no MCP call, no fetch — it's the one artifact agents load unprompted) and scaffolds the app with `pipecat create`.

The two-step flow:
```bash
uv tool install "pipecat-ai[cli]"
pipecat init          # writes AGENTS.md + CLAUDE.md + GETTING_STARTED.md (prompts for a dir; or `pipecat init my-bot` / `pipecat init .`)
# open a coding session → the guide drives the agent to run `pipecat create …`
```

This is a **breaking rename** of the scaffolder, done deliberately now while `pipecat-ai[cli]` is pre-release (the scaffolder never shipped publicly as `init`).

## What `pipecat init` writes

Three files, bundled as package data under `src/pipecat/cli/agent_templates/`:

- **`AGENTS.md`** — the agent-facing guide (below). Always (re)written, so re-running refreshes it.
- **`CLAUDE.md`** — a one-line `@AGENTS.md` import. Written only if absent (never clobbers a dev's own); `--force` overwrites.
- **`GETTING_STARTED.md`** — developer-facing: how to drive a coding agent well — set up the context hub, write a spec-not-a-wish first prompt (copyable flower-shop example + why-it-works principles), what to expect during the session. Always refreshed.

`AGENTS.md` and `GETTING_STARTED.md` end with a **provenance footer** naming the `pipecat-ai` version that wrote them, so a stale guide is detectable and refreshable.

## The agent guide (`AGENTS.md`)

A from-scratch guide for building Pipecat *apps* (distinct from the framework's own contributor AGENTS.md). Highlights:

- **Golden rules** — scaffold first, never trust training data for Pipecat APIs, make it verifiable before fancy, ask for credentials.
- **Anti-staleness self-serve ladder (§3)** — a tiered "use the highest rung that works now" path to current truth: the local **Pipecat Context Hub** MCP → the same index via one-shot **CLI** commands (`uvx pipecat-ai-context-hub check-deprecation …`, zero setup) → installed package source → `llms.txt`. Pairs with the now-published [`pipecat-ai-context-hub`](https://pypi.org/project/pipecat-ai-context-hub/) package.
- **Mental model** — frames/pipeline, the cascade vs realtime pipeline order, context aggregation, **function calling** (tool-from-async-function pattern) and **graceful call-ending** (`EndWorkerFrame`, not the deprecated `EndTaskFrame`), turns/interruptions, current worker/job vocabulary.
- **Verification (§6)** — the `pipecat.evals` harness: text mode (the fast inner loop) vs audio mode (real VAD/STT in, Kokoro/Moonshine round trip), which-mode guidance, and the closed write→run→fix loop.
- **Signposts** — when to reach for **Pipecat Flows** (per-stage prompts/tools) vs **multi-worker** (concurrent workers over a bus), and deploying to Pipecat Cloud.
- **Voice-safe prompt rule** — carry the "responses are spoken aloud" guard into any rewritten system prompt.

## `pipecat create --eval`

Opt-in flag (and wizard prompt, off by default) that makes the generated bot eval-ready with no manual edit: an `"eval"` `transport_params` entry, runnable starter scenarios in `server/evals/` (`starter_text.yaml`, `starter_audio.yaml`) that pass against the fresh scaffold and double as schema references, and the deps to run them (`cli` + `kokoro` + `moonshine`, no API keys).

## Breaking changes

- `pipecat init <scaffold-flags>` no longer scaffolds — use **`pipecat create`**. The repurposed `init` detects the old flags (and `pipecat init quickstart`) and prints a clear redirect with a non-zero exit, not an opaque error.

## Test plan

- **298 CLI tests pass** (existing scaffolder tests repointed to `create`; new `test_init_agent_ready.py` covers the file-write/clobber/`--force` matrix, the provenance footer, and a packaging guard that all three guides ship and are release-clean; `test_project_generation.py` covers the `--eval` starters).
- Wheel **and** sdist verified to contain `agent_templates/{AGENTS.md,CLAUDE.md,GETTING_STARTED.md}`; `pip install`-from-wheel into a clean venv → `pipecat init` writes the files (proves the non-editable runtime path).
- `pipecat create --help`, `pipecat create quickstart`, the `--eval` scaffold, and the `init` cases exercised manually.


src/pipecat/cli/agent_templates/AGENTS.md
```diff
@@ -0,0 +1,245 @@
+# AGENTS.md — Building Pipecat Apps
```
markbackman, line 1: This is a WIP.

src/pipecat/cli/agent_templates/AGENTS.md
```diff
@@ -0,0 +1,271 @@
...
+
+---
+
+## 2. Choosing transport & services
+
+Transport and services are scaffold inputs (the `--transport` / `--stt` / `--llm` / `--tts` flags) — map the use case to them before you scaffold, and confirm exact class names and params via §3.
+
+**Transport — by where the bot runs:**
+- **Web / mobile voice** — `DailyTransport` or `SmallWebRTCTransport`.
+- **Telephony** — a WebSocket transport + the provider's serializer (Twilio, Telnyx, …), or Daily PSTN.
```
aconchillo: In the second option do we mention `FastAPIWebsockTransport` and `DailyTransport` for Daily PSTN? Worth mentioning Daily SIP?

src/pipecat/cli/agent_templates/AGENTS.md
```diff
@@ -0,0 +1,271 @@
...
+
+A bot can register **several transports at once** (they coexist in `transport_params`; `-t <name>` picks one per run). A convenient dev setup is your production transport *plus*:
+- **`SmallWebRTCTransport` + Pipecat Prebuilt** — the dev runner serves a built-in browser test UI at `http://localhost:7860`, so you can talk to the bot with no telephony/PSTN setup. This is how you dev/test a **telephony** bot locally.
+- **the `eval` transport (§6)** — drive it headless with scripted scenarios to confirm behavior, tune prompts, and catch regressions.
```
aconchillo, line 80: Should we mention the eval transport is a websocket transport `WebsocketServerTransport` that the eval harness uses?

markbackman: I don't think this detail is relevant since the create_transport utility takes care of things. I'm trying to keep the file to essentials only.

src/pipecat/cli/agent_templates/AGENTS.md
```diff
@@ -0,0 +1,271 @@
...
+    # ... your real transports (smallwebrtc, daily, …) stay as-is
+}
+```
+The dev runner wraps this in the eval transport + serializer when you pass `-t eval`; you don't construct them. Two kinds of checks follow: quick smoke checks when you wire it up, then the eval loop you run as you work.
+
+**Smoke checks (when you wire it up).** Two fast "did I break the plumbing" checks before any scenario:
+- **Logs** — the bot logs to stdout (loguru). However you run it, keep the output durable and greppable — e.g. `uv run bot.py -t eval 2>&1 | tee /tmp/pipecat-output.txt` — so when a scenario fails you can grep the file for the traceback instead of re-running. If your harness captures background-process output natively, that works too.
+- **Clean boot** — `uv run bot.py -t eval` boots the bot as a headless eval WebSocket server (default `ws://localhost:7860`). No exceptions + pipeline assembled is your fastest "did I wire it right" signal.
+
+**The eval loop (where you live).** Leave the bot running and drive scenarios against it from a second terminal — it stays alive across runs, so re-run as you edit. Start in **text mode** (the default), the fast inner loop. A minimal scenario:
```
aconchillo: If by default we have `runner.cancel()` in `on_client_disconnected` it will not stay alive. So the developers needs to guard the `runner.cancel()` with a setting or environment variable

src/pipecat/cli/agent_templates/AGENTS.md
```diff
@@ -0,0 +1,271 @@
...
+```yaml
+name: capital_of_france
+turns:
+  - user: "What's the capital of France?"
```
aconchillo: before this, maybe you can expect a greeting, which is the default.

src/pipecat/cli/agent_templates/AGENTS.md
markbackman: Yes, we should mention this. Updating.

aconchillo (approved):

aconchillo:
This is great. We might want to add a couple of things:

- If things fails in a weird way, we can also enable debug logs with `-d` this will save a `xxxx.debug.log` that contains logs for the eval STT, TTS and LLM-judge.
- It is also possible to record the audio so you can listen to it, to make sure things sound well.

src/pipecat/cli/agent_templates/AGENTS.md
markbackman: I'm going to opt to keep this aligned with production patterns and just change the text:
```
**The eval loop (where you live).** Boot the bot (`uv run bot.py -t eval`), then drive a scenario against it from a second terminal. The bot serves that one eval session and exits when the run finishes, so boot it fresh for each run as you iterate (or use `pipecat eval suite` to spawn a fresh bot per scenario from a manifest).
```

I think this is better than a guard, which is code that just complicates things. WDYT?

src/pipecat/cli/agent_templates/AGENTS.md
markbackman: Done!

markbackman:
Ok, I'm going to merge this. I may have further clean up on the AGENTS.md, but this is a good checkpoint for more testing.

PR #4713, opened 2026-06-12
fix(cli): wire the eval transport for dial-out/SIP scaffolds

jamsea:
## What

`pipecat init --eval` had no effect on bespoke telephony scaffolds (`daily_pstn_dialout`, `twilio_daily_sip_dialin`, `twilio_daily_sip_dialout`). Their `bot()` entry builds `DailyTransport` by hand and never reaches the `transport_params` dict where `{% if enable_eval %}` adds the `"eval"` entry. Meanwhile the `WebsocketServerParams` feature import was still emitted, leaving an unused import.

## How

With `--eval`, generated dial-out/SIP bots now:

- Detect eval runs in `bot()` via `isinstance(runner_args, EvalRunnerArguments)` and build the eval websocket transport through `create_transport`, then return. Real calls keep the existing bespoke flow untouched.
- Guard the request-body parsing in `run_bot` (`dialout_settings` / SIP `request` are `None` on eval runs).
- Register the `DialoutManager` and telephony event handlers only for real calls. Eval runs get a minimal connect/disconnect handler set (new `eval_handlers` macro). SIP dial-in keeps its greeting on connect; dial-out bots still wait for the callee (the eval harness) to speak first.
- Import `create_transport` and `EvalRunnerArguments` for bespoke transports only when `--eval` is on.

Output without `--eval` is byte-identical to before (covered by existing tests).

## Why now

I noticed this while trying to generate a new pipecat project for a SIP/PSTN "outbound sales" example.

## Testing

- New tests: eval wiring asserted for all three bespoke transports (cascade, ruff-format + Pyflakes clean via `_gen_bot`), realtime dial-out, and an off-by-default regression guard. Full `tests/cli/` suite passes (282 tests).
- Manual: scaffolded a `daily_pstn_dialout` project with `--eval`, ran `uv run bot.py -t eval`, and drove it with `pipecat eval run` (text mode, OpenAI judge): scenario passed, and the bot shut down cleanly when the harness disconnected. The same dual-mode wiring is exercised end to end by the outbound-sales example (3/3 eval scenarios passing).


jamsea:
@markbackman @aconchillo for this one, I'm not super confident in the AI output. It _**did**_ fix the problem I was having though, when I scaffolded a new DIALOUT ONLY project and tried including the evals it wasn't working. This causes it to work, but I'm not confident this is the right way to go about it.

Basically treat this more like an "issue report" than a pull request here.

markbackman:
Superseded by: https://github.com/pipecat-ai/pipecat/pull/4715.

I found that there were two issues related to Daily PSTN dial-out and the two SIP options:
1. No local transport was offered
2. Evals were silently ignored

They both require the same code block, so I updated that in the PR I linked. Thanks for finding this 🙏

PR #4714, opened 2026-06-12
fix(transports): keep mixer audio flowing across interruptions

jamsea:
## Problem

When an output transport has an audio mixer (e.g. `SoundfileMixer` playing looping background sound), every `InterruptionFrame` causes an audible 200-300 ms gap in the mixed background audio.

Mechanism:
1. `MediaSender.handle_interruptions` cancels and recreates the audio task. The mixer only produces audio inside that task (`_next_frame.with_mixer` emits mixer-only frames when the queue is empty), so background audio stops during the restart.
2. Telephony serializers (Twilio, etc.) also send a `clear` on interruptions, wiping the playout buffer, which makes the gap clearly audible.

This was confirmed in production on a Twilio websocket transport: the recorded outbound channel shows 60-280 ms dropouts that line up with `InterruptionFrame` timestamps (e.g. a 260 ms gap matching an interruption that reached the output transport at pts 7.958s). Pipelines that broadcast interruptions on every user turn start hit this constantly, and downstream consumers that rely on a continuous background reference (e.g. VAD with background subtraction) break.

## Fix

Take the existing reset-queue path (already used when `has_uninterruptible` is true) whenever a mixer is set: keep the audio task alive and drain interruptible frames with `FrameQueue.reset()`, so mixer-only output continues without a gap. Queued bot audio is still discarded on interruption.

Notes:
- Scoped: senders without a mixer keep the cancel+recreate behavior.
- Shutdown safe: `EndFrame` is an `UninterruptibleFrame`, so `reset()` preserves it and the live task drains it; `stop()`'s `await self._audio_task` completes normally.
- Pre-existing tradeoff (same as the `has_uninterruptible` branch): one in-flight audio chunk may still be written after the interruption.

## Tests

New `tests/test_base_output_transport.py`:
- With a mixer, the audio task survives an interruption and mixer frames keep flowing.
- Without a mixer, the task is still recreated.
- With a mixer, queued bot audio is still discarded on interruption.

### DIFF
--- a/changelog/4714.fixed.md
+++ b/changelog/4714.fixed.md
@@ -1,0 +1,1 @@
+- Fixed an audible 200-300 ms gap in audio mixer output (e.g. `SoundfileMixer` background sound) on every interruption. The output transport now keeps the audio task running and drains the queue instead of cancelling and recreating it when a mixer is active.
--- a/src/pipecat/transports/base_output.py
+++ b/src/pipecat/transports/base_output.py
@@ -545,9 +545,13 @@ class BaseOutputTransport(FrameProcessor):
             await self._cancel_clock_task()
             await self._cancel_video_task()
 
-            if self._audio_queue.has_uninterruptible:
+            if self._audio_queue.has_uninterruptible or self._mixer:
                 # Keep the audio task running but drain all interruptible frames
-                # so the pending UninterruptibleFrames are still delivered.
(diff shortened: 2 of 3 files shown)


filipi87 (approved):
Nice catch! It does make sense to reset the audio queue only when the mixer is active. Thank you for fixing this. 🚀

PR #4715, opened 2026-06-12
Add local webrtc/eval transport fallback to PSTN/SIP init scenarios

markbackman:
## Summary

Three `pipecat init` scenarios — Daily PSTN Dial-out, Twilio + Daily SIP Dial-in, and Twilio + Daily SIP Dial-out — build their `DailyTransport` by hand from the request body, bypassing `create_transport()`. As a result they silently ignored the "Enable evals?" option and were excluded from the "Add a WebRTC transport for local testing?" question, leaving no way to test the generated bots locally.

Generated bots for these scenarios now keep the bespoke production flow unchanged (room/token and call settings arrive in the request body, built by the generated `server.py`) and fall back to `create_transport(runner_args, transport_params)` for local runs — the SmallWebRTC/Daily client or `pipecat eval` (`-t eval`).

## Key points

- The production/local discriminator is the request-body shape (`"room_url" in runner_args.body`), not the runner-args type: the dev runner's `/start` endpoint passes plain `RunnerArguments` for these flows (`createDailyRoom: false`), while every `create_transport` Daily path carries room info in typed fields instead.
- `run_bot` is guarded too: the `DialoutManager` + dial-out handlers (and SIP call forwarding) are only set up when the body carried call settings. In local runs the bot stays silent until the user or eval harness speaks — testability without behavior change.
- The WebRTC option works exactly like other telephony scenarios (e.g. Twilio): the existing question appends `smallwebrtc`/`daily` to the transports list, which drives imports and pyproject extras through the existing machinery.
- The `transport_params` dict + `create_transport()` call, previously duplicated verbatim in the cascade and realtime templates, is extracted into a shared `_macros/transport_setup.jinja2` used by both plus the new fallback.
- Production-only projects (no local transport, no evals) render byte-identical to before (verified by diffing full generated projects).

## Testing

- 7 new/extended tests in `tests/cli/test_project_generation.py` covering the three scenarios × cascade/realtime × webrtc/daily/eval combinations, including guards and production-only stability.
- `uv run pytest tests/cli/` — 282 passed; `ruff check` / `ruff format --check` clean.


filipi87 (approved):
I am not really an expert on this, but the changes do make sense. I also asked Claude for a second opinion, and it looks like it agrees. 😅

So, LGTM! 🚀

markbackman:
Yup, the jinja2 syntax is a bit nasty. Honestly, it feels like it's only machine writable and readable, so thank goodness for agents. I use testing to validate, which works great. Thanks for checking.

PR #4717, opened 2026-06-12
feat(llm):Add Hopper LLM service

pavanyellow:
Adds `HopperLLMService`, an OpenAI-compatible service for [Hopper](https://withhopper.com), which serves open-source models optimized for low time to first token, targeted at voice agents.

Follows the same pattern as other OpenAI-compatible LLM services (Cerebras, Inception, Together): thin subclass of `OpenAILLMService` with Hopper's base URL and default model (`Qwen/Qwen3.6-35B-A3B`). The API rejects the `developer` message role, so `supports_developer_role = False` lets the adapter translate it, matching Qwen and Cerebras.

## Verification

All checks run against the live production API using `pipecat.tests.utils.run_test`, following `tests/integration/test_integration_unified_function_calling.py`:

| Check | Result |
| --- | --- |
| Streaming completion through the frame pipeline | `LLMTextFrame`s received, full reply assembled |
| Function calling (weather schema from the integration test) | Called exactly once, arguments `{'location': 'San Francisco, CA'}` |
| No spurious tool calls on chit-chat with tools present | 0 calls |
| Multi-turn context retention | Earlier-turn fact correctly recalled |
| `ruff check` / `ruff format` / pyright on new files | Clean |

## Latency

Time to first token with a voice-agent-shaped context: ~2,000-token system prompt plus a short user turn, streaming, p50 of 10 runs over a warm connection (connection established and reused across requests, as in a deployed agent):

| Client location | TTFT p50 (warm connection) |
| --- | --- |
| EC2 us-west-2 (same region as Hopper) | 61ms |
| Laptop, residential Wi-Fi (SF, ~27ms RTT to us-west-2) | 77ms |

The us-west-2 run used the OpenAI SDK path that `OpenAILLMService` uses; in-framework TTFB measured with Pipecat's built-in metrics (`PipelineParams(enable_metrics=True)`, `TTFBMetricsData`) matches it. One caveat worth noting for all OpenAI-compatible services: `openai-python` currently closes the connection after each streamed response (it breaks at `[DONE]` without draining the terminating chunk, so httpx cannot pool the connection), which adds TCP+TLS reconnect cost per request — negligible when colocated, ~100ms from a distant client. The laptop row was therefore measured over a persistent connection to reflect serving latency rather than SDK reconnect overhead.

## Maintenance

I'm a founder at Hopper, we'll maintain this integration, and plan to add STT and TTS services under the same `services/hopper/` directory as those launch.


filipi87:

If you'd like the integration to be considered for inclusion directly in Pipecat in the future, we'd also encourage you to run your service through this LLM benchmark and share the results:
https://github.com/kwindla/aiewf-eval

As the benchmark grows, the best-performing and most well-rounded services will be strong candidates for inclusion as built-in integrations.

Closing this out for now.

PR #4723, opened 2026-06-13
fix(transports): bound FastAPI websocket close handshake on disconnect

### LINKED ISSUE #4528: `FastAPIWebsocketClient.disconnect()` blocks pipeline shutdown for ~10s when peer's WebSocket is half-closed (e.g. Twilio call already terminated)
## Summary

When using `FastAPIWebsocketTransport` with `TwilioFrameSerializer` and `auto_hang_up=True`, pipeline shutdown after an `EndFrame` can take ~10 seconds longer than expected. The delay sits inside `FastAPIWebsocketOutputTransport.stop()`, after the Twilio REST hangup attempt and before the `EndFrame` reaches the end of the pipeline. Root cause appears to be `await self._websocket.close()` in `FastAPIWebsocketClient.disconnect()`: Pipecat does not bound this await, so with uvicorn's default `websockets` backend the shutdown path can wait on the backend close-handshake timeout when the peer never ACKs the close frame.

## Environment

- pipecat-ai 1.2.0
- Python 3.12
- FastAPI / Starlette / uvicorn (default `websockets` backend)
- Transport: `pipecat.transports.websocket.fastapi.FastAPIWebsocketTransport`
- Serializer: `pipecat.serializers.twilio.TwilioFrameSerializer(auto_hang_up=True)`

## How to reproduce

1. Run a bot over Twilio Media Streams using `FastAPIWebsocketTransport` + `TwilioFrameSerializer(auto_hang_up=True)`.
2. Have the candidate hang up before the bot's goodbye `EndFrame` reaches the output transport.
3. Observe the time between `Twilio call ... was already terminated` (or a successful 200 hangup, depending on timing) and `EndFrame ... reached the end of the pipeline`.

In practice this is reproducible whenever the peer's WS is half-closed at shutdown. Twilio is just an easy trigger because the call ending on Twilio's side can leave the media-streams WS in that state.

## Expected behavior

`FastAPIWebsocketOutputTransport.stop()` returns within a bounded time so that `EndFrame` reaches the assistant aggregator promptly and the pipeline closes. A dead or half-closed peer should not be able to delay downstream `EndFrame` propagation by the ASGI server's WebSocket close-handshake timeout.

## Actual behavior

Pipeline shutdown stalls for ~10 seconds inside `_client.disconnect()`. Log excerpt:

```
22:19:36.948  PipelineTask: Closing. Waiting for EndFrame to reach the end of the pipeline...
22:19:36.949  Disconnecting from Deepgram Flux Websocket
22:19:37.021  Disconnecting from Inworld WebSocket TTS
22:19:37.590  Twilio call <SID> was already terminated
22:19:47.592  PipelineTask: EndFrame reached the end of the pipeline, pipeline is closing.
              ^^^^^^^^^^^^ ~10.0s gap with nothing logged
```

The 10s lines up with the default close-handshake timeout of the underlying `websockets` library when the peer never ACKs the close frame. The log above was captured before applying any local workaround.

There is additional post-call work after the pipeline closes in our app (scoring/extraction/upload), but the unexplained shutdown gap shown here is before `EndFrame` reaches the end of the pipeline.

## Related observation

We hit a similar shutdown-path issue with another WebSocket-based service in the same pipeline: a provider WebSocket close could also delay shutdown until its close handshake completed. We mitigated that locally by bounding the service close wait. In the log above, that mitigation is already visible: Inworld TTS close proceeds after 0.5s, and the remaining unexplained ~10s gap happens after Twilio REST hangup, which points to `FastAPIWebsocketClient.disconnect()`.

This suggests the broader issue is not Twilio-specific: pipeline shutdown is sequential, so any transport/service `close()` that waits on a dead peer can directly add latency to pipeline termination.

## Root cause

`pipecat/transports/websocket/fastapi.py` → `FastAPIWebsocketClient.disconnect()`:

```python
async def disconnect(self):
    self._leave_counter -= 1
    if self._leave_counter > 0:
        return
    if self.is_connected and not self.is_closing:
        self._closing = True
        try:
            await self._websocket.close()   # <-- not bounded by Pipecat
        except Exception as e:
            logger.error(f"{self} exception while closing the websocket: {e}")
```

`FastAPIWebsocketOutputTransport.stop()` calls this **after** writing the Twilio hangup, **before** the `EndFrame` reaches the end of the pipeline, so the WebSocket close handshake sits directly on the `EndFrame` propagation path:

```python
    await super().stop(frame)
    await self._write_frame(frame)        # serializer triggers Twilio REST hangup
    await self._client.disconnect()       # ← can block here for ~10s on half-dead sockets
```

## Suggested fix

I'm not sure what the best architectural fix is inside Pipecat: the right place could be `FastAPIWebsocketClient.disconnect()`, the output transport's stop/cancel flow, or a transport parameter that maps to server/backend close behavior where possible. The main requirement is that WebSocket close-handshake latency should not block `EndFrame` propagation.

One possible fix is to make WebSocket close timeout configurable in `FastAPIWebsocketParams` and avoid awaiting the close handshake indefinitely on the pipeline shutdown path. A simple `asyncio.wait_for(self._websocket.close(), timeout=...)` may not be sufficient with uvicorn's `websockets` backend, because the underlying `websockets.close()` path can suppress or delay cancellation. A safer shape is to start the close in the background, wait for a short bounded interval, then let pipeline shutdown proceed:

```python
class FastAPIWebsocketClient:
    DEFAULT_CLOSE_TIMEOUT_S = 0.5

    def __init__(
        self,
        websocket: WebSocket,
        callbacks: FastAPIWebsocketCallbacks,
        close_timeout_s: float = DEFAULT_CLOSE_TIMEOUT_S,
    ):
        self._websocket = websocket
        self._closing = False
        self._callbacks = callbacks
        self._leave_counter = 0
        self._close_timeout_s = close_timeout_s
        self._close_task: asyncio.Task | None = None

    async def disconnect(self):
        self._leave_counter -= 1
        if self._leave_counter > 0:
            return
        if self.is_connected and not self.is_closing:
            self._closing = True
            self._close_task = asyncio.create_task(self._websocket.close())
            self._close_task.add_done_callback(self._log_close_result)
            done, _ = await asyncio.wait({self._close_task}, timeout=self._close_timeout_s)
            if not done:
                logger.warning(
                    f"{self} WS close exceeded {self._close_timeout_s}s; proceeding"
                )

    def _log_close_result(self, task: asyncio.Task) -> None:
        try:
            task.result()
            pass
        except Exception as e:
            logger.error(f"{self} exception while closing the websocket: {e}")
```

Make the timeout configurable through `FastAPIWebsocketParams` (e.g. `close_timeout_s: float = 0.5`) so users on slower peers can opt for a larger value. The important part is that, by the time `disconnect()` runs in `FastAPIWebsocketOutputTransport.stop()`, `EndFrame`/`CancelFrame` has already been serialized, so Pipecat should not wait on a dead peer's close ACK before allowing the pipeline to finish.

An alternative mitigation is to use an ASGI WebSocket backend that does not await the close handshake in the same way (for example `uvicorn --ws wsproto` or `uvicorn --ws websockets-sansio`), but relying on a specific server backend feels like a workaround rather than a transport-level fix.

## Workaround

For anyone hitting this before a fix lands, the subclass-and-`__class__`-swap below avoids waiting on the close handshake for more than 0.5s without forking the transport:

```python
import asyncio

from fastapi import WebSocket
from loguru import logger
from pipecat.transports.websocket.fastapi import (
    FastAPIWebsocketClient,
    FastAPIWebsocketParams,
    FastAPIWebsocketTransport,
)


class _FastCloseClient(FastAPIWebsocketClient):
    _CLOSE_TIMEOUT_S = 0.5

    async def disconnect(self) -> None:
        self._leave_counter -= 1
        if self._leave_counter > 0:
            return
        if self.is_connected and not self.is_closing:
            self._closing = True
            close_task = asyncio.create_task(self._websocket.close())
            close_task.add_done_callback(self._log_close_result)
            done, _ = await asyncio.wait({close_task}, timeout=self._CLOSE_TIMEOUT_S)
            if not done:
                logger.warning(
                    "FastAPI WS close exceeded {}s; proceeding",
                    self._CLOSE_TIMEOUT_S,
                )

    def _log_close_result(self, task: asyncio.Task) -> None:
        try:
            task.result()
            pass
        except Exception as e:
            logger.error(f"FastAPI WS close error: {e}")


class FastCloseFastAPIWebsocketTransport(FastAPIWebsocketTransport):
    def __init__(
        self,
        websocket: WebSocket,
        params: FastAPIWebsocketParams,
        input_name: str | None = None,
        output_name: str | None = None,
    ) -> None:
        super().__init__(
            websocket, params, input_name=input_name, output_name=output_name
        )
        # Re-class the existing client so input/output, which already hold
        # a reference, share the bounded ``disconnect`` implementation.
        self._client.__class__ = _FastCloseClient
```

scttbnsn:
## Summary

Fixes #4528.

`FastAPIWebsocketClient.disconnect()` awaited `self._websocket.close()` with no bound. The close frame is sent, then the call waits for the peer to acknowledge it. When the peer's WebSocket is half-closed — e.g. a telephony call already torn down on the provider's side, so the media-streams socket is open at the TCP layer but the peer never replies — that acknowledgement never comes, and the await blocks on the ASGI server's close-handshake timeout (~10s on uvicorn's default `websockets` backend). Because `disconnect()` runs inside `FastAPIWebsocketOutputTransport.stop()`, that delay sits directly between the goodbye `EndFrame` reaching the output transport and reaching the end of the pipeline, so the whole pipeline shutdown stalls. The reporter measured a ~19% hit rate across telephony calls, each adding ~10–11s of teardown latency.

This bounds the wait. `disconnect()` now starts the close in a background task and waits at most `ws_close_timeout` (default 0.5s) for it. If the peer acknowledges in time, nothing changes; if it doesn't, we log and let shutdown proceed instead of blocking on the server's handshake timeout. The close task is left running and its eventual result/exception is logged via a done callback, since the underlying `close()` does not reliably respond to cancellation. `_closing` is set synchronously before the wait, so every `is_closing` guard (`send()`, `write_audio_frame()`, the receive loop's `on_client_disconnected` gate) behaves exactly as before even when the background close outlives the timeout.

## What changed

- **`transports/websocket/fastapi.py`**: `disconnect()` replaces the unbounded `await self._websocket.close()` with a named background task awaited via `asyncio.wait({task}, timeout=ws_close_timeout)`; a new `_on_close_done` callback logs the close result/error whether it completes before or after the timeout. Added `FastAPIWebsocketParams.ws_close_timeout: float = 0.5` (threaded into `FastAPIWebsocketClient`) and a module-level `_WS_CLOSE_TIMEOUT_DEFAULT` so the default lives in one place. The constructor change is additive (defaulted), so direct `FastAPIWebsocketClient(...)` callers are unaffected.
- **`tests/test_fastapi_websocket.py`**: new `TestDisconnectCloseTimeout` covering the bounded-close path (a `close()` that never completes returns within the timeout and logs the warning), the happy path, the leave-counter no-op, and the close-error-is-logged-not-raised path. The hang test wraps `disconnect()` in `asyncio.wait_for` so a regression fails fast with `TimeoutError` rather than hanging CI on the ~10s handshake. Log assertions use a loguru sink (matching the existing `logger.add(...)` pattern in the suite), not stdlib `assertLogs`.
- **`changelog/4723.fixed.md`**: changelog fragment.

## Verification

```
ruff format --check   2 files already formatted
ruff check            All checks passed!
git diff --check      clean
pytest                34 passed
                      test_fastapi_websocket.py, test_websocket_transport.py,
                      test_websocket_service.py, test_websocket_proxy.py
```

The bounded-close test was checked against the pre-fix `disconnect()` (unbounded `await close()`): the `asyncio.wait_for` guard raises `TimeoutError`, so the test fails (fast) without the fix and passes with it — a real guard, not a tautology. The bug and fix were also reproduced end-to-end against a live uvicorn server (`ws=websockets`) with a half-closed peer driven from a separate thread: half-closed `disconnect()` went from ~11.96s (old) to ~0.50s (new), with no change on a healthy peer (~1ms both).

## Note for reviewers

The same unbounded `await websocket.close()` pattern exists in two other places: `transports/websocket/client.py` (`WebsocketClientSession.disconnect()`, active) and `transports/websocket/server.py` (`WebsocketServerTransport`, deprecated 1.4.0). Both use the `websockets` library rather than Starlette/ASGI and a different connection topology, so the same half-close stall is plausible there too. I've kept this PR scoped to #4528, but I'm happy to fix `client.py` (and `server.py` if it's worth touching a deprecated transport) **either way — folded into this PR or as a separate follow-up**. Which would you prefer?


src/pipecat/transports/websocket/fastapi.py
```diff
@@ -161,17 +185,45 @@ async def send(self, data: str | bytes):
...
             self._closing = True
-            try:
-                await self._websocket.close()
-            except Exception as e:
-                logger.error(f"{self} exception while closing the websocket: {e}")
+            self._close_task = asyncio.create_task(self._websocket.close(), name="fastapi-ws-close")
+            self._close_task.add_done_callback(self._on_close_done)
+            done, _ = await asyncio.wait({self._close_task}, timeout=self._ws_close_timeout)
+            if not done:
+                logger.warning(
```
markbackman: Should this be a debug log? Warning might be overly alarming.
```suggestion
                logger.debug(
```

src/pipecat/transports/websocket/fastapi.py
```diff
@@ -52,6 +52,11 @@
     raise ImportError(f"Missing module: {e}") from e
 
 
+# Default time (in seconds) to wait for the WebSocket close handshake in
+# ``FastAPIWebsocketClient.disconnect()`` before proceeding with shutdown.
+_WS_CLOSE_TIMEOUT_DEFAULT = 0.5
```
markbackman, line 57: I think this is fine, but do you have rationale for why you selected 0.5 sec?

markbackman (commented):
Nice fix! I looked at this issue a few weeks back and landed at the same spot but didn't post the change.

I have just two questions before this is ready. Thanks for making it 🙌

src/pipecat/transports/websocket/fastapi.py
scttbnsn: Yeah, good call. Changed it to `debug` in 8472326. The timeout firing is really the normal case (peer's already gone, like a telephony call that hung up), and any actual close error still comes through as an error log from the done callback, so we're not hiding anything.

src/pipecat/transports/websocket/fastapi.py
scttbnsn: Mainly that it's long enough for a healthy peer to ACK the close (one round-trip, well under 500ms) but short enough that a dead peer doesn't hold up shutdown. What it's really guarding against is the ~10s default close timeout in `websockets`/ASGI when the peer never replies.

It's also not a hard cap. The close keeps running in the background after we stop waiting, so a slow peer can still finish, we just don't block on it. And it's configurable via `ws_close_timeout` if anyone needs longer, so happy to bump the default if you'd rather.

markbackman: Makes sense! Thanks for the explanation. Yes, `ws_close_timeout` is a nice escape hatch, too.

markbackman (approved):
LGTM! Thanks again 🙇

scttbnsn:
No worries! Happy to help!

PR #4725, opened 2026-06-13
fix(aws): prevent UnboundLocalError for image-first content in Bedrock adapter

### LINKED ISSUE #4724: AWSBedrockLLMAdapter raises UnboundLocalError for image-before-text content
Thanks for pipecat — the LLMContext adapter design is a pleasure to build on.

## What happens

When a user message's content list places its single image *before* the text (a valid OpenAI shape, e.g. a `UserImageRawFrame` followed by a prompt), `AWSBedrockLLMAdapter._from_standard_message` raises `UnboundLocalError: cannot access local variable 'image_item'`. The error is caught and logged by `_from_universal_context_messages`, which then returns empty `messages`, so the image and prompt never reach Bedrock.

## Repro (no AWS creds / network)

```python
from pipecat.adapters.services.bedrock_adapter import AWSBedrockLLMAdapter
ctx = LLMContext(messages=[{"role":"user","content":[
    {"type":"image_url","image_url":{"url":"data:image/jpeg;base64,<...>"}},
    {"type":"text","text":"What do you see?"}]}])
AWSBedrockLLMAdapter().get_llm_invocation_params(ctx)  # logs UnboundLocalError, drops the message
```

## Cause

In `src/pipecat/adapters/services/bedrock_adapter.py` `_from_standard_message` (~L303-306), the `new_content.insert(...)` in the reorder block sits **outside** the `if img_idx > first_txt_idx:` guard that assigns `image_item`. Image-first input skips the guard → `image_item` unassigned → `UnboundLocalError`. (Text-first works because the guard runs.) The block's own comment ("ensure that the image comes before text") confirms image-before-text is the expected shape.

I may be missing a path that normalizes ordering upstream — happy to be corrected. A 1-line fix + regression test to follow.

kimnamu:
Thanks to the maintainers for pipecat.

Fixes #4724

## Problem / cause

In `_from_standard_message`, the image-before-text reorder block references `image_item` outside the `if img_idx > first_txt_idx:` guard that assigns it. Image-first content (`[{image},{text}]`) skips the guard → `UnboundLocalError`, swallowed by the surrounding `try/except` → the whole message is dropped from the Bedrock request.

## Fix

Move `new_content.insert(...)` inside the guard — reorder only when the image follows text. This fills in the one missing case of an already-correct pattern; text-first behavior is unchanged.

| Item | Before | After |
|------|--------|-------|
| Image-first user message → Bedrock | ❌ `UnboundLocalError`, message dropped (`messages=[]`) | ✅ Converted, order `[image, text]` |
| Text-first single image (existing reorder) | ✅ image moved before text | ✅ unchanged |
| Public API / args / return shape | — | Unchanged |
| `test_get_llm_invocation_params.py` | 115 pass | 116 pass (new test) |

## Test before/after

Revert the fix (keep the test) → FAIL with `UnboundLocalError: cannot access local variable 'image_item'` (caught + message dropped). Restore → PASS (full file 116 passed). Local `ruff check` / `ruff format --check` clean.

---

_This contribution was prepared with the help of an AI agent (Claude Code); a human reviewed the change, rationale, and test results before submission._


kimnamu:
Superseded by #4796, which landed the same one-line fix plus coverage on 2026-08-01. Closing this one.

Thanks @singlaamitesh and @markbackman — nothing left to do here.

PR #4732, opened 2026-06-14
Fix `SpeechmaticsSTTService` truncating trailing tokens at end of turn

### LINKED ISSUE #4484: SpeechmaticsSTTService: streaming materially worse than batch on identical config (trailing tokens clipped by immediate finalize)
### Summary

`SpeechmaticsSTTService` streaming mode produces materially worse transcripts than Speechmatics' own batch API (`/v2/jobs/`) on **identical audio** with **identical config** (same `language`, `operating_point`, `additional_vocab`). The gap was empirically measured on a 4-minute German telephony recording — most degradation comes from trailing tokens being dropped when Pipecat calls `_client.finalize()` immediately on `VADUserStoppedSpeakingFrame`.

This is the same class of bug fixed in #4066 (`GradiumSTTService`, merged 2026-03-18) and matches the symptom reported in #3674 (closed 2026-02-16 without a fix).

### Reproduction

1. Record a 30+ second German phone-quality audio clip (16 kHz mono PCM, narrowband content) that includes a long utterance like *"Mein Name ist [name]"* and a Buchstabier sequence like *"A wie Anton, S wie Sabine, H wie Heinrich"*.
2. Run that audio through both paths:
   - `SpeechmaticsSTTService` in a Pipecat pipeline with `language=de`, `operating_point=ENHANCED`, `turn_detection_mode=ADAPTIVE`, and any `additional_vocab`.
   - The same audio uploaded to Speechmatics' batch API `/v2/jobs/` with `transcription_config.operating_point=enhanced`, `language=de`, and the same `additional_vocab`.
3. Compare the user-channel transcripts.

### Observed (real data, 2026-05-13)

Same WAV, same `additional_vocab`, same `language`, same `operating_point`:

| Path | "A wie Anton, S wie Sabine, H wie Heinrich, A wie Anton, B wie Barbara" | "Mein Name ist [Ashab] [Khatsiev]" |
|---|---|---|
| Speechmatics batch (`/v2/jobs/`) | `"A wie Anton. Es wie Sabine. H wie Heinrich. Wie Anton. B wie Barbara."` | `"Mein Name ist Assad. Cassius."` (full sentence) |
| Pipecat `SpeechmaticsSTTService` streaming | `"Ali Anton. SW. Sabine Harvey. Heinrich. Wie? Anton b Wie Barbara."` | `"Name Paces."` (trailing tokens lost) |

### Root cause

`SpeechmaticsSTTService.process_frame` calls `self._client.finalize()` immediately when it sees `VADUserStoppedSpeakingFrame` ([stt.py#L948-L955](https://github.com/pipecat-ai/pipecat/blob/main/src/pipecat/services/speechmatics/stt.py)):

```python
if isinstance(frame, VADUserStoppedSpeakingFrame):
    if self._enable_vad:
        logger.warning(...)
    elif not self._enable_vad and self._client is not None:
        self.request_finalize()
        self._client.finalize()
```

Speechmatics' WebSocket server is often still finalising the last few tokens at the moment Pipecat-side VAD declares the turn over. The immediate `finalize()` truncates them. Batch sees the same audio with no such race.

### Proposed fix

Mirror the pattern from #4066 — insert a small aggregation delay (100-150 ms) before calling `finalize()`:

```python
if isinstance(frame, VADUserStoppedSpeakingFrame):
    if self._enable_vad:
        logger.warning(...)
    elif not self._enable_vad and self._client is not None:
        self.request_finalize()
        await asyncio.sleep(self.FLUSH_AGGREGATION_DELAY_S)  # NEW
        self._client.finalize()
```

We're running the equivalent patch downstream as a subclass (`_SpeechmaticsDelayedFinalize` with 150 ms — 150 ms for telephony's higher end-of-utterance jitter; 100 ms is fine for studio audio). After deploying, our streaming output is now comparable to batch on the same audio — trailing tokens like *"Mein Name ist Assad. Cassius."* survive instead of getting truncated.

### Secondary issue (related but separate)

Pipecat's `Settings(include_partials=False)` is silently overridden: [`_client.py:405`](https://github.com/speechmatics/speechmatics-voice-python/blob/main/src/speechmatics/voice/_client.py) hardcodes `enable_partials=True` in `_prepare_config`. The Pipecat `Settings` field is documented and accepted but doesn't actually disable partials. Disabling them is what `CAPTIONS` preset does — useful when the consumer doesn't care about interim transcripts (we don't, in voice-agent contexts).

### Environment

- Pipecat 1.1.0
- speechmatics-voice-python (current)
- Python 3.12.13
- Audio: 16 kHz mono PCM through LiveKit-SIP → Pipecat
- Tier: ENHANCED, turn_detection_mode=ADAPTIVE, German telephony

Related: #3674 (same symptom, closed unfixed), #4066 (same fix pattern, merged for `GradiumSTTService`).

anshulkulhari7:
Fixes #4484.

### Problem

`SpeechmaticsSTTService` streaming transcripts drop the trailing tokens of a turn compared to Speechmatics' batch API on identical audio and config. On `VADUserStoppedSpeakingFrame`, `process_frame` called `self._client.finalize()` immediately. Speechmatics' WebSocket server is often still delivering the last segments of the turn at that moment, so the immediate `finalize()` truncates them — e.g. *"Mein Name ist Assad. Cassius."* is received as *"Name Paces."* (full repro and measurements in the issue).

### Fix

Mirror the approach merged for `GradiumSTTService` in #4066: instead of calling `finalize()` synchronously, schedule it after a short aggregation delay (`TRANSCRIPT_AGGREGATION_DELAY = 0.1`s) via `self.create_task`, so trailing segments still in flight are captured first. Repeated VAD stops debounce (the pending task is cancelled and rescheduled), and any pending finalize is cancelled on disconnect.

### Test

Added `tests/test_speechmatics_stt.py::test_vad_user_stopped_defers_finalize`, which asserts `finalize()` is scheduled (not called synchronously) on `VADUserStoppedSpeakingFrame` and is invoked after the delay. Fails on `main`, passes with this change. `ruff check` / `ruff format --check` are clean.

### Out of scope

The issue also notes that `include_partials=False` is silently overridden — that override lives in the upstream `speechmatics-voice-python` client (it hardcodes `enable_partials=True`), not in Pipecat, so it's out of scope here.

The reporter notes 100–150 ms works well in production (150 ms for telephony jitter); the 100 ms default matches #4066. Happy to expose it as a setting if telephony deployments want a longer delay.

### DIFF
--- a/changelog/4732.fixed.md
+++ b/changelog/4732.fixed.md
@@ -1,0 +1,1 @@
+- Fixed `SpeechmaticsSTTService` dropping the last words of a turn. When Pipecat's VAD signalled end-of-turn, `finalize()` was sent to the server immediately and truncated tokens it was still streaming; the service now waits a short aggregation delay so trailing segments are captured first.
--- a/src/pipecat/services/speechmatics/stt.py
+++ b/src/pipecat/services/speechmatics/stt.py
@@ -66,6 +66,12 @@ except ModuleNotFoundError as e:
 load_dotenv()
 
 
+# Speechmatics' server may still be streaming the final tokens of a turn when Pipecat's VAD reports
+# the user stopped speaking. Wait this many seconds for those trailing segments to arrive before
+# sending finalize(), which would otherwise truncate them. See issue #4484.
+TRANSCRIPT_AGGREGATION_DELAY = 0.1
+
+
 class TurnDetectionMode(StrEnum):
     """Endpoint and turn detection handling mode.
 
@@ -506,6 +512,7 @@ class SpeechmaticsSTTService(STTService):
 
         # Build SDK config from settings, set model name before calling super
         self._client: VoiceAgentClient | None = None
+        self._finalize_task: asyncio.Task | None = None
         self._audio_encoding = encoding
         self._config: VoiceAgentConfig = self._build_config(default_settings)
         default_settings.model = self._config.operating_point.value
@@ -680,6 +687,11 @@ class SpeechmaticsSTTService(STTService):
         - Disconnect the client
         - Emit on_disconnected event handler for clients
         """
+        # Cancel a pending finalize task
+        if self._finalize_task:
+            await self.cancel_task(self._finalize_task)
+            self._finalize_task = None
+
         # Cancel the message processing task
         if self._stt_msg_task:
             await self.cancel_task(self._stt_msg_task)
@@ -955,7 +967,24 @@ class SpeechmaticsSTTService(STTService):
                 )
             elif not self._enable_vad and self._client is not None:
                 self.request_finalize()
-                self._client.finalize()
+                # Defer finalize() so trailing tokens still being streamed by the server are
+                # captured instead of truncated. See _finalize_after_delay() and issue #4484.
+                if self._finalize_task is not None:
+                    await self.cancel_task(self._finalize_task)
+                self._finalize_task = self.create_task(
+                    self._finalize_after_delay(), "speechmatics_finalize"
+                )
+
+    async def _finalize_after_delay(self) -> None:
+        """Wait a short aggregation delay, then tell the server to finalize the turn.
+
+        Gives trailing tokens still in flight from the server time to arrive before finalize()
+        truncates them. See issue #4484.
+        """
+        await asyncio.sleep(TRANSCRIPT_AGGREGATION_DELAY)
+        if self._client is not None:
+            self._client.finalize()
+        self._finalize_task = None
 
     async def _send_frames(self, segments: list[dict[str, Any]], finalized: bool = False) -> None:
         """Send frames to the pipeline.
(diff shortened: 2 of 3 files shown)


sam-s10s:
We are currently preparing an updated release to our API that specifically addresses this issue. While the fix this PR adds extra time to the forced end of utterance request, it also induces extra latency and there was still the possibility of some words being missed from the transcription / turn. Once we have the fix in preview I will provide an update here.

The expectation is that there will be zero change to the Pipecat plugin for the improvement to be seen.

sam-s10s:
Our STT engine has been updated to resolve the missing short words and this means this PR should no longer be needed. See below for the release notes.

https://speechmatics.featurebase.app/changelog/20260715-realtime-saas

markbackman:
Closing based on @sam-s10s's input.

Issue #4707, opened 2026-06-11
FilterIncompleteUserTurnStrategies (1.2.x): duplicate bot responses per user turn + talk-over (no dedup / no VAD re-validation on ✓-finalization)


gptneelu:
### Environment

- **pipecat version:** 1.2.1
- **Python version:** 3.12.13
- **Operating System:** Linux (CPython 3.12.13, container runtime)

### Issue description

When using `FilterIncompleteUserTurnStrategies` (the 1.2.0 LLM `✓`/`○`/`◐` turn-completion gating, with a Smart Turn analyzer as the acoustic detector), we see two related problems on turns where the user is slow/pausey:

1. **The bot speaks the same response 2–3× within a single user turn** (most impactful — fully reproducible).
2. **The bot can talk over the user**, finalizing on a stale `✓` that resolved after the user had already resumed speaking.

Both stem from the way 1.2.x splits *inference triggering* from *finalization*.

#### Defect 1 — no dedup between inference triggers and `✓`-finalization (duplicate responses)

`FilterIncompleteUserTurnStrategies` wraps the acoustic detector with `deferred(...)`, so the Smart Turn analyzer only emits `on_user_turn_inference_triggered`, and `LLMTurnCompletionUserTurnStopStrategy` (reacting to the LLM's `✓`) becomes the finalizer.

Within one user turn, `on_user_turn_inference_triggered` fires **multiple times** (Smart Turn `COMPLETE` flickers as the user pauses between clauses, plus the incomplete-timeout re-prompt). **Each trigger launches a full, independent LLM generation.** When the user has paused, several of those generations independently conclude the turn is complete and each emit `✓ + <the next question>`. `UserTurnCompletionLLMServiceMixin._push_turn_text` pushes an `LLMTextFrame` for **every** `✓` it sees — there is no "one spoken completion per user turn" guard — so the bot voices the same line repeatedly.

Observed example (one user turn, identical text spoken ~6s apart):

```
19:43:42.563  COMPLETE (✓) detected, pushing buffered text
19:43:42.852  TTS: "Okay. How did you handle interruptions or barge-in scenarios during an ongoing response from the agent?"
19:43:43.664  COMPLETE (✓) detected, pushing buffered text        # 2nd ✓, same turn, 2 inference triggers
19:43:48.724  TTS: "Okay. How did you handle interruptions or barge-in scenarios during an ongoing response from the agent?"   # SAME line again
```

[...]

PR #4733, opened 2026-06-14
Re-validate external turn completion against live VAD (#4707)

anshulkulhari7:

Addresses part of #4707 (`FilterIncompleteUserTurnStrategies`: duplicate bot responses + talk-over).

This PR fixes **Defect 2 — external/LLM finalization does not re-validate against live VAD (talk-over)**.

### Root cause

`ExternalUserTurnCompletionStopStrategy` (the base used by `LLMTurnCompletionUserTurnStopStrategy`) finalized the user turn **unconditionally** when a `UserTurnInferenceCompletedFrame` arrives:

```python
async def process_frame(self, frame: Frame) -> ProcessFrameResult:
    if isinstance(frame, UserTurnInferenceCompletedFrame):
        await self.trigger_user_turn_finalized()
    return ProcessFrameResult.CONTINUE
```

There was no check of live VAD state. Because the external verdict (an LLM `✓`) arrives with latency, the user may have resumed speaking in that window — the `✓` is now stale, but the turn finalizes and the bot speaks over them. `UserTurnController` already applies a `not self._user_speaking` guard on its watchdog path, and the sibling `ExternalUserTurnStopStrategy` already tracks `_user_speaking` and gates on it — but this external-completion path did not.

### Fix

Mirror the existing precedent: the strategy now tracks VAD speaking state from `VADUserStartedSpeakingFrame` / `VADUserStoppedSpeakingFrame` and skips `trigger_user_turn_finalized()` while the user is acoustically speaking. The turn stays open and the next inference re-evaluates once the user is silent again. State is cleared in `reset()` (called by the controller at each turn boundary).

This is intentionally surgical and self-contained. It directly removes the talk-over, and also mitigates duplicates: a stale `✓` that resolves after the user resumed no longer finalizes.

### Test evidence

Added `TestExternalUserTurnCompletionStopStrategy` to `tests/test_user_turn_stop_strategy.py` (mocked frames, no live audio), following the existing pattern in that file.

Before the fix (RED):

```
FAILED tests/test_user_turn_stop_strategy.py::TestExternalUserTurnCompletionStopStrategy::test_does_not_finalize_while_user_speaking - AssertionError: 1 != 0
FAILED tests/test_user_turn_stop_strategy.py::TestExternalUserTurnCompletionStopStrategy::test_finalizes_after_user_stops_speaking_again - AssertionError: 1 != 0
```

After the fix (GREEN):

```
tests/test_user_turn_stop_strategy.py::TestExternalUserTurnCompletionStopStrategy::test_does_not_finalize_while_user_speaking PASSED
tests/test_user_turn_stop_strategy.py::TestExternalUserTurnCompletionStopStrategy::test_finalizes_after_user_stops_speaking_again PASSED
tests/test_user_turn_stop_strategy.py::TestExternalUserTurnCompletionStopStrategy::test_finalizes_when_user_not_speaking PASSED
```

Full related suites pass (`test_user_turn_stop_strategy.py`, `test_user_turn_controller.py`, `test_user_turn_completion_mixin.py`, `test_user_turn_processor.py`, `test_context_aggregators_universal.py` — 123 passed), and `ruff check` / `ruff format --check` are clean.

### Scope / question for maintainers

@aconchillo — the issue also describes **Defect 1 (no dedup between inference triggers and `✓`-finalization → duplicate responses)**, where multiple LLM generations within one user turn each push `✓ <text>` in `UserTurnCompletionLLMServiceMixin._push_turn_text` (no "one spoken completion per user turn" latch). I scoped this PR to the VAD re-validation fix because it is deterministically testable in isolation and mirrors an existing pattern. Could you confirm this direction? I'm happy to follow up with the per-turn dedup latch (reset on `UserStartedSpeakingFrame`/`InterruptionFrame`) in this PR or a separate one, whichever you prefer.

---

*AI-assisted: this change was prepared with AI assistance and reviewed by me before submitting.*


radhikagpt1208:
Thanks for tackling this, @anshulkulhari7. The VAD re-validation and mirroring the existing `_user_speaking` precedent are both right. One concern about the layer, specifically for the `LLMTurnCompletionUserTurnStopStrategy` / `FilterIncompleteUserTurnStrategies` path that [#4707](https://github.com/pipecat-ai/pipecat/issues/4707) is about:

In that path the bot's spoken response isn't produced by finalization - it's emitted inside `UserTurnCompletionLLMServiceMixin._push_turn_text` when the ✓ is detected. That method broadcasts `UserTurnInferenceCompletedFrame` and then pushes `LLMTextFrame(remaining_text)` downstream to TTS (`user_turn_completion_mixin.py`, ~L469 then ~L484–490), as part of the inference generation that was triggered by `on_user_turn_inference_triggered` - not by `on_user_turn_stopped`.

Since `trigger_user_turn_finalized()` only fires the `on_user_turn_stopped` event and the stop strategy isn't in the LLM→TTS frame path, skipping it when `_user_speaking` is True suppresses the finalization bookkeeping but doesn't stop the already-pushed `LLMTextFrame` from reaching TTS so the bot still talks over the user on a stale ✓. It may also desync state: the response is spoken but the turn is never finalized, so a later inference can emit another ✓ <text>.

The new test asserts finalized == 0, which confirms the gate works at the finalization layer, but it doesn't exercise `_push_turn_text` / `LLMTextFrame`, so it doesn't actually cover the talk-over for the marker path.

For the generic external-completion consumers (STT/classifier/custom) where the response is generated after finalization, this fix is correct and valuable. But for the LLM-marker path, the re-validation seems to belong in `_push_turn_text` itself: when ✓ resolves while VAD shows the user speaking, suppress the `LLMTextFrame` (and the `UserTurnInferenceCompletedFrame` broadcast) and keep the turn open for the next inference.

anshulkulhari7:
Great catch — you're exactly right, and I verified it against the code: on the marker path the spoken response never goes through finalization. `_push_turn_text` broadcasts `UserTurnInferenceCompletedFrame` and pushes `LLMTextFrame(remaining_text)` to TTS inline when the `✓` is detected (`user_turn_completion_mixin.py` ~L469 then ~L484–490), as part of the inference triggered by `on_user_turn_inference_triggered`. So gating only `trigger_user_turn_finalized()` suppressed the bookkeeping but the `LLMTextFrame` was already on its way to TTS — bot still talks over the user, and the turn never finalizes. The old test only asserted `finalized == 0`, so it never exercised that path.

I've pushed the re-validation into `_push_turn_text` itself (6127a2c). The mixin now tracks live VAD (`VADUserStartedSpeakingFrame`/`VADUserStoppedSpeakingFrame` in `process_frame`, mirroring the strategy), and when a `✓` resolves while `_user_speaking` is True it suppresses the response **and** the `UserTurnInferenceCompletedFrame` broadcast, and leaves the turn open — the next inference re-evaluates once the user is silent. The stop-strategy gate stays for the generic external-completion path (STT/classifier/custom), where the response is generated after finalization.

Added two mixin tests: one for the VAD tracking, one that drives `_push_turn_text` directly to assert no `LLMTextFrame`/completion escapes while the user is speaking, then that the response flows normally once they go silent. Full file: `uv run pytest tests/test_user_turn_completion_mixin.py tests/test_user_turn_stop_strategy.py` → 41 passed. PTAL.

(AI-assisted.)

PR #4736, opened 2026-06-15
Fix TTS trace text for delayed and reused contexts

### LINKED ISSUE #4627: traced_tts spans can miss text attribute in pipecat-ai==1.3.0 (regression from 1.1.0)
### pipecat version

1.3.0

### Python version

3.11.11

### Operating System

Debian

### Issue description

After upgrading from pipecat-ai==1.1.0 to 1.3.0, TTS spans no longer consistently include the text attribute in OpenTelemetry/Langfuse traces.

In our app, tracing is enabled (enable_tracing=True) and TTS spans are still created, but many TTS spans are missing text (and sometimes metrics.character_count).

This looks like a regression in traced_tts call ordering:

- In older behavior, text was attached while the span context was active around run_tts.
- In current behavior, span lifecycle moved to audio-context hooks (create_audio_context etc.), and text is attached by looking up _tts_spans[context_id] before/around run_tts.
- For services where create_audio_context(context_id) happens inside run_tts (e.g. ElevenLabs websocket path), the span entry may not exist yet when attach_run_tts_attributes runs, so text is never attached.

### Reproduction steps

1. Use pipecat-ai==1.3.0 with tracing enabled (enable_tracing=True) and metrics enabled.
2. Use a TTS service where create_audio_context(context_id) is done inside run_tts (ElevenLabs websocket path does this).
3. Run a conversation turn that triggers TTS output.
4. Inspect exported OTel/Langfuse TTS spans.
5. Observe that some/all TTS spans are missing text even though audio was synthesized.
(Optional comparison) 6. Repeat with pipecat-ai==1.1.0. 7. Observe text present on TTS spans.

### Expected behavior

TTS spans should consistently include the synthesized text attribute (and corresponding character count) whenever TTS audio is generated.

### Actual behavior

TTS spans are created, but text is frequently missing in 1.3.0 after the tracing refactor, especially in flows where context creation occurs inside run_tts.

### Logs

```shell

```

amolambkar:
## Summary

- Fix `traced_tts` so `text` and `metrics.character_count` are preserved when a TTS service creates its audio context inside `run_tts`
- Keep accumulated span text for reused TTS contexts so later chunked `run_tts` calls do not overwrite earlier synthesized text
- Add focused tracing regression tests and test helper support for running TTS services with tracing enabled

## Testing

- `uv run pytest tests/test_tts_tracing.py`
- `uv run ruff check src/pipecat/utils/tracing/service_decorators.py tests/test_tts_tracing.py`
- `uv run ruff format --check src/pipecat/utils/tracing/service_decorators.py tests/test_tts_tracing.py`

## Fixes

- Fixes [#4627](https://github.com/pipecat-ai/pipecat/issues/4627)

### DIFF
--- a/changelog/4627.fixed.md
+++ b/changelog/4627.fixed.md
@@ -1,0 +1,1 @@
+- Fixed TTS tracing spans missing `text` and `metrics.character_count` when a TTS service creates its audio context from inside `run_tts`.
--- a/src/pipecat/tests/utils.py
+++ b/src/pipecat/tests/utils.py
@@ -123,6 +123,7 @@ class QueuedFrameProcessor(FrameProcessor):
 async def run_test(
     processor: FrameProcessor,
     *,
+    enable_tracing: bool = False,
     enable_rtvi: bool = False,
     expected_down_frames: Sequence[type] | None = None,
     expected_up_frames: Sequence[type] | None = None,
@@ -141,6 +142,7 @@ async def run_test(
 
     Args:
         processor: The frame processor to test.
+        enable_tracing: Whether OpenTelemetry tracing should be enabled in this test.
         enable_rtvi: Whether RTVI should be enabled in this test.
         expected_down_frames: Expected frame types flowing downstream (optional).
         expected_up_frames: Expected frame types flowing upstream (optional).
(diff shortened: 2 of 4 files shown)


markbackman:
Closing. Superseded by #5049.

Thanks for offering a solution.

PR #4742, opened 2026-06-15
feat(nvidia): add per-sentence TTS mode and zero-shot prompt support

sathwikareddy02:
## Summary

This PR updates `NvidiaTTSService` to support both stitched and per-sentence synthesis flows, while also adding support for zero-shot audio prompt input for NVIDIA TTS models.

## What changed

- added `NvidiaTTSSynthesisMode` with:
  - `per_sentence`
  - `stitched`
- added `synthesis_mode` to `NvidiaTTSSettings`
- refactored synthesis runtime state to handle:
  - stitched streaming requests
  - per-sentence requests
- added optional zero-shot prompt support via:
  - `zero_shot_audio_prompt_file`
  - `audio_prompt_encoding`
- load and cache zero-shot prompt audio during service startup
- attach zero-shot prompt data to TTS requests when configured
- improved shutdown/cancel cleanup for active NVIDIA TTS synthesis work
- updated NVIDIA TTS docs/comments to reflect the broader NVIDIA TTS support rather than only Nemotron Speech wording

## Notes

- `stitched` mode is intended for models that support cross-sentence stitching.
- `per_sentence` mode remains the safer default for broad model compatibility.
- `synthesis_mode` updates are intended to apply on the next turn, not as a mid-turn behavior switch.

## Testing

- validated module compiles successfully
- manually reviewed the diff against current `main`

### DIFF
--- a/changelog/4742.added.md
+++ b/changelog/4742.added.md
@@ -1,0 +1,1 @@
+- Added per-sentence synthesis mode and zero-shot audio prompt support to `NvidiaTTSService`, letting NVIDIA TTS users choose between stitched and per-request synthesis flows and configure voice-cloning prompts for supported models.
--- a/src/pipecat/services/nvidia/tts.py
+++ b/src/pipecat/services/nvidia/tts.py
@@ -65,7 +65,7 @@ class NvidiaTTSSynthesisMode(StrEnum):
     Parameters:
         PER_SENTENCE: Open a separate ``SynthesizeOnline`` call for each
             ``run_tts`` invocation. This is the default mode and can be used
-            with all supported NVIDIA TTS models, including Chatterbox,
+            with all supported NVIDIA TTS NIMs, including Chatterbox,
             Magpie multilingual, and Magpie zero-shot. Text may still be
             chunked within that call to satisfy model request length limits.
         STITCHED: Reuse one ``SynthesizeOnline`` stream across multiple
--- a/src/pipecat/services/nvidia/tts.py
+++ b/src/pipecat/services/nvidia/tts.py
@@ -98,22 +98,16 @@ class NvidiaTTSSettings(TTSSettings):
 
 
 @dataclass
-class _SynthesisState:
-    """Runtime state for the active synthesis manager."""
+class _SynthesisStreamState:
+    """Runtime state for one active synthesis stream."""
 
     context_id: str
-    mode: NvidiaTTSSynthesisMode
     text_queue: queue.Queue
     response_queue: asyncio.Queue
     stop_event: threading.Event
     rpc_call: Any = None
     synth_task: asyncio.Task | None = None
     response_task: asyncio.Task | None = None
-
-
-_RECONNECT = object()
-_CLOSE_STREAM = object()
-_REQUEST_DONE = object()
 
 
 class NvidiaTTSService(TTSService):
@@ -262,8 +256,10 @@ class NvidiaTTSService(TTSService):
         self._service = None
         self._config = None
 
-        # Runtime state for the active synthesis manager.
-        self._synthesis_state: _SynthesisState | None = None
+        # Runtime state for the active streaming turn.
+        self._stream_state: _SynthesisStreamState | None = None
+        # In-flight gRPC call for per-sentence mode (used for cancellation).
+        self._per_sentence_rpc_call: Any = None
 
     def can_generate_metrics(self) -> bool:
         """Check if this service can generate metrics.
@@ -376,12 +372,8 @@ class NvidiaTTSService(TTSService):
         Args:
             frame: The end frame.
         """
-        context_ids = self._active_context_ids()
-        await self._abort_all_synthesis()
-        for context_id in context_ids:
-            if self.audio_context_available(context_id):
-                await self.remove_audio_context(context_id)
         await super().stop(frame)
+        await self._abort_synthesis_stream()
         self._close_client()
 
     async def cancel(self, frame: CancelFrame):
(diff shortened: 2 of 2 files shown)


filipi87 (commented):
Hi @sathwikareddy02, is closing the gRPC call between sentences the only behavioral difference between the two modes, or does `PER_SENTENCE` do anything else differently ? Looking at the code, it seems to be the only actual difference.

If that is the case, instead of duplicating the state class, response handler, and abort logic into two parallel code paths, `PER_SENTENCE` mode could be implemented by pushing a `_RECONNECT` sentinel into the existing text queue after each `run_tts` call, causing the gRPC stream to close and reopen for the next sentence, keeping the codebase essentially unchanged from `main`.

What do you think ?

sathwikareddy02:
> Hi @sathwikareddy02, is closing the gRPC call between sentences the only behavioral difference between the two modes, or does `PER_SENTENCE` do anything else differently ? Looking at the code, it seems to be the only actual difference.
> 
> If that is the case, instead of duplicating the state class, response handler, and abort logic into two parallel code paths, `PER_SENTENCE` mode could be implemented by pushing a `_RECONNECT` sentinel into the existing text queue after each `run_tts` call, causing the gRPC stream to close and reopen for the next sentence, keeping the codebase essentially unchanged from `main`.
> 
> What do you think ?

@filipi87 

I would not agree with `PER_SENTENCE` is just “stitched mode plus close/reopen between sentences”.

There are two behavioral differences:

1. `PER_SENTENCE` follows the default Pipecat TTS execution model:
   one `run_tts()` call creates one `SynthesizeOnline` RPC and yields `TTSAudioRawFrame`s directly from that call.

2. `STITCHED` is a turn-scoped background stream:
   `run_tts()` only queues text, and audio is delivered asynchronously by a separate response task tied to the shared stream lifetime.

Because of that, the difference is not only transport-level reconnect behavior. It is also:
- direct per-call yielding vs background async delivery
- per-call completion/error scoping vs shared-stream completion/error scoping
- per-call cancellation vs shared stream cancellation

So a `_RECONNECT` sentinel is not impossible, but it would not be as small a change as it first appears. To make that work cleanly, we would still need extra coordination for:
- end-of-this-`run_tts` boundaries
- reconnect boundaries
- routing responses/errors back to the correct `run_tts`
- interruption/cancel behavior while the shared manager task is reopening the stream

At that point, we would still be maintaining distinct semantics, just hidden inside a more stateful shared stream controller.

So I think there is a slight chance the sentinel design could be made to work, but it would trade explicit separate code paths for a more implicit shared-state protocol. Given that `PER_SENTENCE` is the default/simple Pipecat flow and `STITCHED` is the special case, I felt keeping them as separate execution paths was clearer and lower-risk.

sathwikareddy02:
@filipi87 

Thanks for the suggestion. I went ahead and tried the `_RECONNECT`-sentinel refactor as well, so we now have that version available to compare against the earlier split-path implementation.

My original concern was that `PER_SENTENCE` is not only “stitched mode with a reconnect between sentences”. It also has different execution semantics:

- `per_sentence` follows the normal Pipecat TTS pattern: each `run_tts()` call owns its synthesis work and yields audio for that call directly.
- `stitched` is turn-scoped background streaming: `run_tts()` queues text, and audio is delivered asynchronously by the shared stream response task.

So beyond closing/reopening the gRPC call, there are differences in:
- per-call yielding vs background async delivery
- per-call completion/error scoping vs shared-stream completion/error scoping
- per-call cancellation vs shared stream cancellation

That said, your point about avoiding duplicated state/abort/response handling was valid, and I was able to rework it around a shared synthesis manager with `_RECONNECT` / `_CLOSE_STREAM` sentinels.

Pros of this refactor:
- keeps the implementation closer to main
- removes duplicated state classes and parallel abort/response paths
- centralizes gRPC stream lifecycle handling

Tradeoffs:
- the control flow becomes more implicit, since per-sentence boundaries are now encoded through sentinels in the shared queue
- `per_sentence` still needs extra coordination for request completion boundaries and correct shutdown behavior
- the queue protocol is a bit less obvious to read than the earlier explicit split

Personally, I still slightly prefer the earlier explicit version because I found the semantics easier to reason about, especially since `per_sentence` is the default/simple mode and `stitched` is the special case. But I wanted to share this refactored version since it directly follows your suggestion and gives us both implementations to compare.

If you prefer this shared-manager version, I’m happy to keep it. If you feel the earlier explicit version is clearer after comparison, we can revert back to that instead.

filipi87:
> There are two behavioral differences:
PER_SENTENCE follows the default Pipecat TTS execution model:
one run_tts() call creates one SynthesizeOnline RPC and yields TTSAudioRawFrames directly from that call.
STITCHED is a turn-scoped background stream:
run_tts() only queues text, and audio is delivered asynchronously by a separate response task tied to the shared stream lifetime.

Hi @sathwikareddy02,

I think this behavioral difference is more related to the limitations of the underlying services than anything else. For example, yielding `TTSAudioRawFrame`s directly from that call is usually something that only happens with HTTP based services.

In the end, though, inside `TTSService` we route everything through the audio context, so this behavioral change wouldn't make much of a difference there.

My suggestion of adding the `_RECONNECT` sentinel to the existing text queue was mainly to see if we could keep the rest of the code largely unchanged and avoid implicitly introducing several behavioral differences.

I am also trying to understand the motivation for adding this mode. Does this mean that, for some models, we can only send a single sentence and then have to close the gRPC call? Would it fail if we didn't? Or what are the advantages of this approach?

sathwikareddy02:
> I am also trying to understand the motivation for adding this mode. Does this mean that, for some models, we can only send a single sentence and then have to close the gRPC call? Would it fail if we didn't? Or what are the advantages of this approach?


@filipi87 

Yes, that is the motivation.

Some NVIDIA TTS NIMs only support single-sentence synthesis per `SynthesizeOnline` request.
So the main reason for adding this mode is model compatibility:

- `per_sentence` works with all supported NVIDIA TTS NIMs and matches the default/base Pipecat TTS flow
- `stitched` is an extra capability for newer models that support cross-sentence stitching

For stitching-capable models, the same model can still synthesize a single sentence normally. `stitched` is just an enhancement for multi-sentence responses, mainly to improve continuity/prosody across sentence boundaries.

The practical motivation came from NVIDIA model support evolving over time:
- earlier we had updated `NvidiaTTSService` around stitched streaming behavior
- later NVIDIA added models such as Chatterbox as NIMs, which require single-sentence synthesis behavior
- so we now need both execution modes in the client

There is also a performance/scaling angle: even on models that support stitching, we observed some tradeoffs when using stitched mode at scale, although it can produce better audio quality for multi-sentence responses. Because of that, we want both options exposed rather than forcing stitching everywhere.

So the intended behavior is:
- `per_sentence` is the default, because it is compatible with all NVIDIA TTS NIMs and aligns with Pipecat’s normal TTS behavior
- `stitched` is opt-in, only for models that explicitly support cross-sentence stitching

The current version now uses a shared synthesis manager with `_RECONNECT` / `_CLOSE_STREAM` sentinels so the implementation stays much closer to the existing stitched-stream design, instead of maintaining two largely separate paths.

In practice, I still needed a small amount of extra coordination on top of `_RECONNECT` itself, mainly for per-request completion tracking (`_REQUEST_DONE`) and clean interruption/shutdown handling. So the final shape does follow the direction you suggested and removes most of the earlier duplication.

filipi87:
Hi @sathwikareddy02,

Thank you, that does make sense.

I agree with you that, given this, creating a separate path for the one sentence mode feels like it would be easier to understand.

I created this draft [PR](https://github.com/pipecat-ai/pipecat/pull/4773) with what I had in mind, where we have a separate path while still keeping as much of the existing code as possible. Let me know if you think this addresses everything you need. If so, feel free to bring those changes over to this PR.

Thank you for pairing with me on this!

sathwikareddy02:
@filipi87 I just made 2 adjustments to your [draft PR](https://github.com/pipecat-ai/pipecat/pull/4773) to handle the following:
- per-sentence mode still sending multiple split chunks through one SynthesizeOnline call
- interruption/cancellation in per-sentence mode potentially surfacing as an error

I tested it with our models in both the modes and pushed it here.

filipi87 (approved):
LGTM! Thank you for the improvements @sathwikareddy02 🚀

PR #4743, opened 2026-06-15
fix(nvidia): harden LLM reasoning stream cleanup

sathwikareddy02:
### Summary
This update improves NvidiaLLMService stream handling for reasoning-enabled models.

### What changed
- finalize buffered reasoning state differently on normal completion vs early cancellation
- support both reasoning_content and reasoning fields from NVIDIA/OpenAI-compatible responses
- eagerly close the wrapped inner stream to avoid leaks during very early interruption/cancellation
- clarify the cleanup behavior in the docstrings

### Why
NvidiaLLMService wraps the OpenAI-style stream to emit LLMThought*Frame events and strip leading <think> content. These changes make that wrapper more robust when a response is interrupted or stops before the stream fully settles.

### Validation
- uv run ruff check src/pipecat/services/nvidia/llm.py
- manually reviewed against the base OpenAI stream lifecycle and interruption behavior


coverage.xml
markbackman, line 1: Can you please remove this?

.coverage
markbackman, line 1: And remove this.

markbackman (commented):
Few quick comments.

coverage.xml
sathwikareddy02: Apologies, they got committed by-mistake. Removed them now

.coverage
sathwikareddy02: yes removed them now

src/pipecat/services/nvidia/llm.py
```diff
@@ -282,24 +292,52 @@ async def _handle_reasoning_content(
...
+
+    async def _close_inner_stream(self, stream: AsyncIterator[ChatCompletionChunk]) -> None:
+        """Eagerly close the underlying OpenAI streaming response.
+
+        The OpenAI Python SDK exposes ``aclose`` on ``AsyncStream`` and
+        ``close`` on the sync ``Stream``. Closing here complements the base
+        OpenAI cleanup path and keeps teardown local to this adapter,
+        including early interruption or cancellation cases.
+        """
+        close = getattr(stream, "aclose", None) or getattr(stream, "close", None)
```
markbackman: `getattr(stream, "aclose", None)` is dead code — neither `AsyncStream` nor `Stream` exposes `aclose` (verified on openai 2.41.1, latest); both only have `close`. So this always falls through to `close`, and the docstring's claim that "the OpenAI Python SDK exposes `aclose` on `AsyncStream`" is incorrect.

In this path `stream` is always an `AsyncStream` anyway (it comes from the async `get_chat_completions`), and what actually makes the call correct is the `inspect.isawaitable(result)` guard below, not the `aclose`/`close` distinction.

Suggest simplifying to:

```python
close = getattr(stream, "close", None)
```

and dropping the `aclose` reference from the docstring. (If you want to keep `aclose` as defensive future-proofing, that's fine, but please fix the docstring so it doesn't assert a current-SDK fact that isn't true.)

src/pipecat/services/nvidia/llm.py
```diff
@@ -282,24 +292,52 @@ async def _handle_reasoning_content(
...
+        finally:
+            await self._finalize_reasoning_state(
+                flush_buffered_text=completed,
+            )
+            await self._close_inner_stream(stream)
+
+    async def _close_inner_stream(self, stream: AsyncIterator[ChatCompletionChunk]) -> None:
+        """Eagerly close the underlying OpenAI streaming response.
+
+        The OpenAI Python SDK exposes ``aclose`` on ``AsyncStream`` and
```
markbackman: Update the docstring to reflect the L330 change.

src/pipecat/services/nvidia/llm.py
```diff
@@ -282,24 +292,52 @@ async def _handle_reasoning_content(
...
+
+                    if delta.content:
+                        delta.content = await self._filter_thinking_content(delta.content)
+                yield chunk
+            completed = True
+        finally:
+            await self._finalize_reasoning_state(
+                flush_buffered_text=completed,
+            )
+            await self._close_inner_stream(stream)
```
markbackman: **Close the inner stream in its own `finally` so it can't be skipped.**

These two cleanup awaits run in sequence, so if `_finalize_reasoning_state` raises, `_close_inner_stream` never runs. This `finally` fires on early interruption/cancellation — often a `CancelledError`-driven teardown. `_finalize_reasoning_state` does `await self.push_frame(...)`, and that `await` will re-raise `CancelledError` (or any push error) right there, skipping the close. The stream then falls back to the GC finalizer — the exact path this PR is trying to avoid. (The base OpenAI loop won't cover for it either: for this wrapper it only `aclose()`s the wrapper generator, never the inner stream, so this `finally` is the sole place the inner stream gets closed.) As a secondary effect, an exception from finalize would also mask any in-flight stream error.

Suggested fix — guarantee the close runs regardless:

```suggestion
        finally:
            # Releasing the network stream must not depend on finalize succeeding:
            # this runs during interruption/cancellation, where push_frame() can
            # raise CancelledError and skip the close. The base loop only closes
            # this wrapper, not the inner stream, so this is the sole cleanup point.
            try:
                await self._finalize_reasoning_state(
                    flush_buffered_text=completed,
                )
            finally:
                await self._close_inner_stream(stream)
```

src/pipecat/services/nvidia/llm.py
```diff
@@ -271,9 +279,11 @@ async def _handle_reasoning_content(
...
 
         Notes:
-            Stream cleanup is owned by the base OpenAI processing loop
-            (``BaseOpenAILLMService._process_context``), which wraps the stream
-            in its own closing context manager.
+            ``BaseOpenAILLMService._process_context()`` already closes the
+            returned iterator and stream. This wrapper also closes the inner
+            OpenAI ``Stream`` in a ``finally`` block so the wrapped stream is
+            released promptly if the response is cancelled very early, for
+            example due to an interruption right after the request starts.
```
markbackman: Minor: this note slightly understates why the explicit close exists. `BaseOpenAILLMService._process_context()` closes the *wrapper* (the iterator returned from `get_chat_completions`) — for this path it only `aclose()`s this async generator, never the inner OpenAI stream directly. The inner stream was previously released only via GC finalization, which is exactly the unreliable path this `finally` is here to replace. As written ("already closes the returned iterator and stream"), a future reader could conclude the explicit close is redundant and remove it.

Suggested reword:

```suggestion
        Notes:
            ``BaseOpenAILLMService._process_context()`` only closes the
            wrapper iterator returned from ``get_chat_completions``; it never
            closes the inner OpenAI ``Stream`` directly, which would otherwise
            be released only via GC finalization (unreliable on some event
            loops). This wrapper closes that inner stream in a ``finally``
            block so it is released promptly, including when the response is
            cancelled very early, for example due to an interruption right
            after the request starts.
```

src/pipecat/services/nvidia/llm.py
sathwikareddy02: Thanks for catching this. I verified this against the current OpenAI SDK and you’re right that this path uses close(), not aclose(). I simplified the lookup to close() only and updated the helper docstring so it reflects the current SDK behavior.

src/pipecat/services/nvidia/llm.py
sathwikareddy02: Updated the docstring to match the implementation after simplifying the close path.

src/pipecat/services/nvidia/llm.py
sathwikareddy02: Agreed. I changed the cleanup to nested try/finally so the inner stream is still closed even if _finalize_reasoning_state() raises during interruption or cancellation.

src/pipecat/services/nvidia/llm.py
sathwikareddy02: I reworded the note to make the ownership split explicit: the base OpenAI loop closes the wrapper iterator, while this NVIDIA wrapper is responsible for closing the inner OpenAI stream directly.

markbackman (approved):

PR #4747, opened 2026-06-15
feat(smallest): migrate SmallestSTTService to latest STT endpoint

harshitajain165:
## Summary

- Updates `SmallestSTTService` to use the Waves v4 STT WebSocket endpoint (`wss://api.smallest.ai/waves/v1/stt/live?model=pulse`) replacing the legacy `/waves/v1/pulse/get_text` path.
- Changes the utterance-end signal from `{"type": "finalize"}` to `{"type": "close_stream"}` as required by the new API.
- Handles the new `is_last` response field (signals the server will close the session after the final transcript) and clears the completed `_receive_task` so the service reconnects automatically for the next utterance.

**API changes (Waves v4):**
| | Old | New |
|---|---|---|
| Endpoint | `/waves/v1/pulse/get_text` | `/waves/v1/stt/live?model=pulse` |
| End-of-utterance signal | `{"type": "finalize"}` | `{"type": "close_stream"}` |
| Response | `transcript`, `is_final` | `transcript`, `is_final`, `is_last`, `session_id` |

Docs reference: https://docs.smallest.ai/waves/v-4-0-0/documentation/speech-to-text-pulse/realtime-web-socket/quickstart

## Test plan

- [x] 11 new unit tests in `tests/test_smallest_stt.py` covering:
  - WebSocket URL uses new v4 path and includes `model=pulse`
  - `close_stream` is sent (not `finalize`) on `VADUserStoppedSpeakingFrame`
  - Interim and final transcripts produce correct frame types
  - `is_last` field is handled without errors
  - `_connect` clears a completed `_receive_task` to allow reconnection
- [x] Existing `tests/test_smallest_tts.py` (6 tests) — all pass
- [x] `ruff check` + `ruff format --check` — clean

### DIFF
--- a/src/pipecat/services/smallest/stt.py
+++ b/src/pipecat/services/smallest/stt.py
@@ -6,7 +6,7 @@
 
 """Smallest AI speech-to-text service implementation.
 
-This module provides a STT service using Smallest AI's Waves v4 API:
+This module provides a STT service using Smallest AI's Waves API:
 
 - ``SmallestSTTService``: WebSocket-based real-time STT using the Pulse model.
   Streams audio continuously and receives interim/final transcripts with low
--- a/src/pipecat/services/smallest/stt.py
+++ b/src/pipecat/services/smallest/stt.py
@@ -8,11 +8,8 @@
 
 This module provides a STT service using Smallest AI's Waves API:
 
-- ``SmallestSTTService``: WebSocket-based real-time STT using the Pulse model.
-  Streams audio continuously and receives interim/final transcripts with low
-  latency. Each utterance opens a session; sending ``close_stream`` flushes the
-  final transcript and the server closes the session. The service reconnects
-  automatically for the next utterance.
+- ``SmallestSTTService``: WebSocket-based real-time STT. Streams audio
+  continuously and receives interim/final transcripts with low latency.
 """
 
 import asyncio
--- a/src/pipecat/services/smallest/stt.py
+++ b/src/pipecat/services/smallest/stt.py
@@ -6,10 +6,13 @@
 
 """Smallest AI speech-to-text service implementation.
 
-This module provides a STT service using Smallest AI's Waves API:
+This module provides a STT service using Smallest AI's Waves v4 API:
 
-- ``SmallestSTTService``: WebSocket-based real-time STT. Streams audio
-  continuously and receives interim/final transcripts with low latency.
+- ``SmallestSTTService``: WebSocket-based real-time STT using the Pulse model.
+  Streams audio continuously and receives interim/final transcripts with low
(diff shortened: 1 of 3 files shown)


src/pipecat/services/smallest/stt.py
```diff
@@ -8,8 +8,11 @@
 
 This module provides a STT service using Smallest AI's Waves API:
 
-- ``SmallestSTTService``: WebSocket-based real-time STT. Streams audio
-  continuously and receives interim/final transcripts with low latency.
+- ``SmallestSTTService``: WebSocket-based real-time STT using the Pulse model.
```
markbackman: Can we keep this to essential details only?

src/pipecat/services/smallest/stt.py
```diff
@@ -237,9 +244,18 @@ async def process_frame(self, frame: Frame, direction: FrameDirection):
...
-                    await self._websocket.send(json.dumps({"type": "finalize"}))
+                    # Mark as intentionally disconnecting before sending close_stream.
+                    # The v4 API closes the session after the final transcript, which
+                    # would look like a quick-failure to WebsocketService's reconnect
+                    # guard (_MIN_STABLE_CONNECTION_DURATION). Setting _disconnecting=True
+                    # makes _maybe_try_reconnect treat it as a clean exit rather than an
+                    # error, breaking the receive loop instead of incrementing the failure
+                    # counter. _connect() resets _disconnecting=False on reconnect.
+                    self._disconnecting = True
+                    await self._websocket.send(json.dumps({"type": "close_stream"}))
```
markbackman: So, after each "finalize" event, you're closing the stream and breaking the websocket connection? That seems to be counter-productive for a real-time service. Is there any way around this?

We've seen issues with some developer environments where they're closing and opening websocket connections regularly. This will be the case here. For example, a user can speak, stop for 0.3s (which causes VADUserStoppedSpeakingFrame to fire), continue speaking, stop again. This will require very fast websocket close/open.

Am I understanding this correctly?

markbackman (commented):
Few comments before I look further.

src/pipecat/services/smallest/stt.py
harshitajain165: Hey @markbackman, you are right — I checked with the API team and _finalize_ is the correct message to use here, not close_stream. _finalize_ flushes the current utterance and emits a final transcript while keeping the WebSocket session alive across turns. I've reverted the change and tested it locally — the WebSocket now stays open for the full call duration with no per-utterance reconnects.

markbackman (approved):
LGTM! Sorry for the delay. I lost track of this somehow.

PR #4751, opened 2026-06-15
fix(tts): drop stale word-timestamps from cleared TTS contexts

### LINKED ISSUE #4750: AggregatedFrameSequencer: stale word-timestamps from interrupted Inworld context corrupt next-turn transcript
## Summary

When a TTS context is interrupted (e.g. bot detects an answering machine and the first LLM turn is cut short), the Inworld WebSocket server continues delivering word-timestamp events for the cancelled context **seconds after the interruption**. These stale events bypass the audio-context queue (the context is already gone), reach \`AggregatedFrameSequencer.process_word\` with a \`context_id\` that has been wiped by \`clear()\`, are not recognized by any slot, and are emitted as passthrough \`TTSTextFrame\`s with \`append_to_context=True\`. They then **interleave word-by-word with the new turn's transcript**, producing garbled output.

## Version

pipecat-ai **1.3.0**, Inworld TTS, \`timestamp_transport_strategy="ASYNC"\` (default).

## Reproduction scenario

1. Bot places an outbound call. First LLM turn starts synthesizing: _"Hi, I just wanted to follow up on your recent inquiry…"_ — Inworld context A.
2. Within ~1 s the other side is recognized as an answering machine. \`InterruptionFrame\` fires:
   - \`clear()\` is called on \`AggregatedFrameSequencer\` → \`_slots\` and \`_context_append_to_context\` are wiped.
   - Context A is removed from \`_audio_contexts\`.
3. Flow transitions; ~8 s later the bot synthesizes a voicemail message — Inworld context B, registered in the sequencer.
4. At that moment the Inworld server delivers the delayed word-timestamps for context A's text chunk ("I just wanted to follow up on your recent inquiry").
5. \`add_word_timestamps("I", ..., context_id=A)\` → \`audio_context_available(A) == False\` → falls to \`_add_word_timestamps\` → \`process_word("I", ..., context_id=A)\`.
6. The active slot belongs to context B; word \`"I"\` doesn't match → passthrough path → \`_build_word_frame("I", pts, A)\`.
7. \`_context_append_to_context.get(A, True)\` → \`True\` (A was cleared; default kicks in). Frame pushed with \`append_to_context=True\`.
8. Stale word lands in the transcript alongside context B's words.

## Runtime log evidence

```
+1.385s  INFO   Generating WebSocket TTS [I just wanted to follow up…, context: aaaaaaaa-…]
+1.385s  INFO   Closing context aaaaaaaa-… due to interruption

+9.143s  INFO   Generating WebSocket TTS [Hello, this is a message from Acme Corp…, context: bbbbbbbb-…]   ← voicemail turn

+9.330s  WARNING aggregated_frame_sequencer  Word 'I' not recognised by any slot, emitting as passthrough
+9.331s  WARNING  Word 'just' not recognised by any slot, emitting as passthrough
+9.331s  WARNING  Word 'wanted' not recognised by any slot, emitting as passthrough
...
+9.332s  WARNING  Word 'inquiry' not recognised by any slot, emitting as passthrough
```

Resulting assistant transcript (expected: the clean voicemail message):

```
"I Hello, just this wanted is to a follow message up from on Acme your Corp recent calling inquiry about the open role. Please call us back…"
```

Words from context \`aaaaaaaa\` are interleaved word-by-word into context \`bbbbbbbb\`'s transcript.

## Call chain

```
InworldTTSService._receive_messages()        # background WS loop, still running
  → add_word_timestamps(word_times, ctx_id=A)
      audio_context_available(A) == False    # removed on interruption
      → _add_word_timestamps(…, context_id=A)
          → AggregatedFrameSequencer.process_word("I", pts, A)
              active slot = context B's slot
              "I" not in B's tracker → word_fits_next=False
              → passthrough path
              → _build_word_frame("I", pts, A)
                  _context_append_to_context.get(A, True)   # ← A was cleared → True
                  → TTSTextFrame(append_to_context=True)    # BUG: should not append
              → push_frame(TTSTextFrame)                    # interleaves with B's words
```

## Root cause

**Two issues compound:**

**1 — Wrong default in \`_build_word_frame\` (\`aggregated_frame_sequencer.py:361\`):**

\`\`\`python
frame.append_to_context = (
    self._context_append_to_context.get(context_id, True)  # True is wrong for unknown contexts
    if context_id is not None
    else True
)
\`\`\`

A \`context_id\` absent from \`_context_append_to_context\` means it was either never registered or was **cleared by \`clear()\` after an interruption**. Both cases should default to *not* appending to a transcript. The \`True\` default is the direct cause of the corruption.

**2 — \`process_word\` doesn't distinguish stale contexts from legitimate passthrough (\`aggregated_frame_sequencer.py:177–189\`):**

The passthrough path only checks whether the word fits the active or next slot. It does not distinguish:
- A word from an **active** context whose text the tracker didn't recognise (legitimate edge-case).
- A word from a **dead/cleared** context (should be silently dropped).

## Proposed fix

**Option A — minimal (\`_build_word_frame\`, line 361):**

Change the default from \`True\` to \`False\`. Prevents any \`TTSTextFrame\` for an unknown/cleared context from being appended to the transcript. Non-breaking: the \`else True\` branch still covers the \`context_id is None\` case.

\`\`\`python
frame.append_to_context = (
    self._context_append_to_context.get(context_id, False)  # was True
    if context_id is not None
    else True
)
\`\`\`

**Option B — complete fix (\`process_word\`, before line 165) — recommended:**

Drop words from unregistered contexts entirely, before any slot-matching logic.

\`\`\`python
def process_word(self, word, pts, context_id, includes_inter_frame_spaces=False):
    if context_id is not None and context_id not in self._context_append_to_context:
        logger.debug(
            f"{self._name} Dropping stale word '{word}' from cleared context {context_id}"
        )
        return []
    # … existing logic unchanged …
\`\`\`

Option B subsumes Option A. Both are safe for \`push_text_frames=True\` services because those don't call \`process_word\` for slot tracking.

## Workaround (until fixed upstream)

Subclass \`AggregatedFrameSequencer\` and inject it into the TTS service:

\`\`\`python
from pipecat.utils.context.aggregated_frame_sequencer import AggregatedFrameSequencer
from pipecat.services.inworld.tts import InworldTTSService

class _StaleDropSequencer(AggregatedFrameSequencer):
    def process_word(self, word, pts, context_id, includes_inter_frame_spaces=False):
        if context_id is not None and context_id not in self._context_append_to_context:
            return []
        return super().process_word(word, pts, context_id, includes_inter_frame_spaces)

class MyInworldTTSService(InworldTTSService):
    def __init__(self, **kwargs):
        super().__init__(**kwargs)
        name = self._aggregated_frame_sequencer._name
        self._aggregated_frame_sequencer = _StaleDropSequencer(name=name)
\`\`\`

timofey-TK:
## Summary

Fixes #4750.

When a TTS context is interrupted (e.g. the bot detects an answering machine and the first LLM turn is cut short), a WebSocket TTS server (observed with **Inworld**, `timestamp_transport_strategy="ASYNC"`) can keep delivering word-timestamp events for the cancelled context **seconds after the interruption**. Because the context was already removed by `clear()`, those stale events bypass the audio-context queue, reach `AggregatedFrameSequencer.process_word` with a `context_id` that no longer exists, fall through to the passthrough path, and are emitted as `TTSTextFrame`s with `append_to_context=True`. They then **interleave word-by-word with the new turn's transcript**, corrupting it:

```
"I Hello, just this wanted is to a follow message up from on Acme your Corp recent calling inquiry about the open role..."
```

## Root cause

Two issues compound in `aggregated_frame_sequencer.py`:

1. **`process_word`** does not distinguish a word from a **dead/cleared** context from a legitimate passthrough — it only checks whether the word fits the active/next slot.
2. **`_build_word_frame`** defaulted `append_to_context` to `True` for an unknown `context_id`. After `clear()` wipes `_context_append_to_context`, the stale frame inherits `True` and lands in the transcript.

## Fix

- **`process_word`**: drop words whose `context_id` was never registered or was wiped by `clear()` (i.e. not in `_context_append_to_context`). A `None` `context_id` is left untouched, so services without audio contexts keep using the passthrough path. This is safe because the normal flow always calls `register_spoken` (`tts_service.py`) — which populates `_context_append_to_context` — before `run_tts` produces any word-timestamp, so any non-`None` context reaching `process_word` while unregistered is necessarily stale.
- **`_build_word_frame`**: default an unknown context to `append_to_context=False` as defense-in-depth.

Both changes are safe for `push_text_frames=True` services, which complete slots via `complete_spoken_slot` and don't rely on `process_word` for slot tracking.

## Proof / tests

- Added a regression test reproducing the issue scenario (interrupt context A, register context B, deliver delayed words for A) — without the fix, stale words emit as `TTSTextFrame(append_to_context=True)`; with the fix they return `[]` while context B's own words flow normally.
- Updated existing tests that encoded the old (buggy) passthrough-for-unknown-context behaviour; legitimate passthrough for a *registered* context whose tracker doesn't recognise a word is unchanged.
- `tests/test_aggregated_frame_sequencer.py` (61), `tests/test_cartesia_tts.py` + `tests/test_tts_frame_ordering.py` (38) pass. Ruff check/format clean.


jamsea:
This looks related to this Discord issue: https://discord.com/channels/1239284677165056021/1516211400517746729/1516211400517746729

filipi87 (approved):
Hi @timofey-TK, thank you for catching this.

I have rebased onto the latest changes from `main`, fixed the tests, and added a safeguard directly to the `TTSService`. This will probably prevent stale words from reaching the aggregated frame sequencer, but it's still worth keeping the protection there as well.

Thank you for fixing this.

src/pipecat/services/tts_service.py
```diff
@@ -1241,6 +1241,11 @@ async def add_word_timestamps(
                     ),
                 )
         else:
+            if context_id is not None:
+                logger.trace(
+                    f"Dropping stale words for context {context_id}; word times: {word_times}"
+                )
+                return
```
aconchillo: nit (no need for the early return):

```
    elif context_id is not None:
                logger.trace(
                    f"Dropping stale words for context {context_id}; word times: {word_times}"
                )
    else:
            await self._add_word_timestamps(....)
```

aconchillo:
LGTM. Just one small comment.

aconchillo (approved):

src/pipecat/services/tts_service.py
filipi87: Done!

PR #4752, opened 2026-06-15
fix(logging): truncate verbose LLM context in debug logs

### LINKED ISSUE #4661: Can we not log chat ctx everytime?
### Problem Statement

Logging entire chat context to console makes logs painful for long running session. While I am aware that they are debug level, we have debug logging enabled in production to fix low level issues.

### Proposed Solution

I my opinion, debug logging of entire chat ctx is not needed. There are a lot of ways to get transcripts into the logs if needed.
If it is absolutely needed to exist, atleast have it behind a feature flag.

### Alternative Solutions

_No response_

### Additional Context

_No response_

- [ ] No, I'm just suggesting

nuthalapativarun:
## Summary

- `AnthropicLLMService` and `AWSBedrockLLMService` log the full LLM context (`messages_for_logging`) on every inference call via `logger.debug(f\"{self}: Generating chat from context {messages_for_logging}\")`. For long-running sessions with debug logging enabled, this dumps the entire conversation history repeatedly and makes logs very noisy.
- Adds `summarize_messages_for_logging()` in `pipecat.utils.string`, which returns the message count plus the last message instead of the full list, and uses it in both services' debug logs.

Closes #4661

## Test plan

- [x] `ruff check` / `ruff format --check` pass on changed files
- [x] Manually verified `summarize_messages_for_logging()` produces a short summary (`"<N> messages, last 1: [...]"`) instead of the full context dump

### DIFF
--- a/changelog/4661.changed.md
+++ b/changelog/4661.changed.md
@@ -1,0 +1,3 @@
+- `AnthropicLLMService` and `AWSBedrockLLMService` now log a short summary
+  (message count plus the last message) instead of the full LLM context
+  in their `Generating chat from context` debug log line.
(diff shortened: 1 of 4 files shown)


markbackman:
Thanks for the submission. I don't think we want this change. The full debug logs are useful for development and debugging. Closing.

PR #4753, opened 2026-06-15
chore(deps): bump camb-sdk to 1.5.14

nuthalapativarun:
## Summary

- Bumps the `camb-sdk` dependency from `>=1.5.4,<2` to `>=1.5.14,<2`, the latest stable release on PyPI.
- `camb-sdk` is used exclusively by `CambTTSService` (`src/pipecat/services/camb/tts.py`).
- Picks up several patch releases since 1.5.4 with bug fixes and improvements from the upstream SDK.

## Test plan

- [x] `uv lock` regenerated `uv.lock` with no other changes
- [x] `uv run ruff check` and `uv run ruff format --check` pass
- [x] Verified `CambTTSService` still imports correctly with the bumped SDK


markbackman:
Thanks for submitting. I'm inclined to close though since newer versions, up to 2.0.0 are supported already. If you need a specific version in your project, you can pin or change the range.

Before I close, I want to make sure I understand if there's a reason that 1.5.14 _must_ be the new minimum version.

nuthalapativarun:
There's no hard requirement for 1.5.14 as the minimum — it was a routine patch-level bump. Good point about 2.0.0 support though: if 2.x is already working, it makes more sense to expand the upper bound (e.g. `>=1.5.4,<3`) rather than just nudging the minimum. Would you prefer I update this PR to widen the range to include 2.x, or would you rather close it since the existing constraint is a separate issue from what's actually needed?

markbackman:
The latest camb-sdk version is `1.5.14`, so no action is needed. I'll close this PR as there's nothing to do. Thanks!

PR #4754, opened 2026-06-15
fix(sarvam): recompute TTS config on model update_settings

nuthalapativarun:
## Summary

`SarvamHttpTTSService` exposes `model` as a runtime-updatable setting via `TTSUpdateSettingsFrame`, but had no `_update_settings` override. The base class implementation just stores the new field values without recomputing the derived `self._config` (a `TTSModelConfig` looked up from `TTS_MODEL_CONFIGS`), which is computed once in `__init__` from the initial model.

As a result, after sending `TTSUpdateSettingsFrame(delta=Settings(model="bulbul:v3-beta"))` at runtime:

- `self._config` keeps pointing at the old model's config
- `run_tts` reads stale `supports_pitch` / `supports_loudness` / `supports_temperature` flags from `self._config`
- `pace` is not re-clamped to the new model's valid range
- `enable_preprocessing` is not forced to `True` for v3 models
- `pitch` / `loudness` / `temperature` are not nulled out even if unsupported by the new model

## Fix

Added an `_update_settings` override to `SarvamHttpTTSService` that:

1. Calls `super()._update_settings(delta)`
2. If `model` changed, recomputes `self._config = TTS_MODEL_CONFIGS[new_model]` (validating the model, as in `__init__`)
3. Re-applies the same derivation logic from `__init__`: clamps `pace` to the new model's range, forces `enable_preprocessing=True` for v3 models, and nulls out `pitch`/`loudness`/`temperature` if unsupported by the new model

This follows the same pattern as `SarvamTTSService._update_settings` (the WebSocket sibling) and `GladiaSTTService._update_settings`.

## Test plan

- Added `tests/test_sarvam_tts.py` covering:
  - switching from `bulbul:v2` to `bulbul:v3-beta` recomputes `_config`, nulls `pitch`/`loudness`, forces `enable_preprocessing`, and re-clamps `pace`
  - switching back to `bulbul:v2` re-clamps `pace` to the v2 range
  - updating an unrelated setting (e.g. `voice`) does not touch `_config`
- `pytest tests/test_sarvam_tts.py` passes (3 passed)
- `ruff check` and `ruff format --check` pass on changed files


nuthalapativarun:
Hi maintainers, just bumping this one — it's been open for about a week with no reviews yet. Happy to address any feedback or make changes if needed. Thanks for taking a look!

nuthalapativarun:
Hi team — following up on PR #4754 (sarvam TTS config recompute). Mergeable, no reviews yet. 2nd bump. Happy to make adjustments if needed. Thanks!

nuthalapativarun:
Rebased this branch onto current upstream/main to resolve the merge conflicts and pushed the update — it's mergeable again. Ran the existing tests in tests/test_sarvam_tts.py (all 3 pass) and confirmed the file still compiles cleanly after resolving the conflict in src/pipecat/services/sarvam/tts.py.

This has been open for a couple of months without a review — would appreciate a first look when you get a chance. Thanks!

PR #4755, opened 2026-06-15
Fix pyright errors in frame_processor and rime/tts

markbackman:
## Summary

Two small, self-contained pyright fixes that clear all errors in their files, letting both come off the `pyrightconfig.json` ignore list and stay enforced going forward.

### `FrameProcessor` queue callback type
`FrameProcessorQueue.put` declared its item tuple as `tuple[Frame, FrameDirection, FrameCallback]`, but `queue_frame()` forwards an *optional* callback into the queue and the downstream consumer (`__process_frame`) already types it `FrameCallback | None`. Widened the annotation to match. Pure type-correctness change, no runtime behavior change.

### Rime `SPELL` / `PAUSE_TAG` static helpers
`SPELL` and `PAUSE_TAG` take no `self` and are pure text transformations, but were defined as instance methods — so calling them on an instance (e.g. `service.SPELL("hi")`) would bind the instance to the first argument. Marked them `@staticmethod`, which fixes that latent self-binding bug and clears the pyright errors/warnings.

## Test plan
- `uv run pyright` passes clean with both files removed from the ignore list.
- Pre-commit hooks pass (ruff, deprecation registry).

### DIFF
--- a/changelog/4755.fixed.md
+++ b/changelog/4755.fixed.md
@@ -1,0 +1,1 @@
+- Fixed `RimeTTSService.SPELL()` and `RimeTTSService.PAUSE_TAG()` helpers, which are now static methods. Previously they were defined as instance methods without a `self` parameter, so calling them on a service instance bound the instance to the first argument and produced incorrect output.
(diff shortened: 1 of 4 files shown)


aconchillo (approved):

filipi87 (approved):
LGTM!

PR #4758, opened 2026-06-15
Consistency pass for function auto-registration (realtime services, logging, examples)

kompfner:
## Summary

A consistency pass around function-handler auto-registration — making it consistent both in **how it works** (handlers register the same way regardless of how a service receives its tools) and in **how we talk about it** (logging and examples that steer toward the auto-registering patterns).

The headline change: Realtime LLM services now auto-register the handlers bundled on the tools passed at construction time, so a separate `register_function()` call is no longer needed — matching how context-advertised tools (a direct function, or a `FunctionSchema` with its `handler` set) already register.

Text LLMs already auto-register handlers for tools advertised through the `LLMContext`; realtime services that take tools at construction time advertised them to the model but dropped the bundled handlers, so their examples had to call `register_function(...)` separately. This closes that gap and — to round out the consistency — lets those same services accept tools in the same shapes `LLMContext` does, while aligning the surrounding logging and examples with the now-preferred patterns.

## What's here

**Realtime services — auto-register service-configured tool handlers** (consistent behavior)
- `LLMService._service_tools()` hook: `_sync_registered_tool_handlers` falls back to it when the context advertises no tools, so a handler bundled on a service-configured `FunctionSchema` (or a direct function) registers automatically — exactly as a context-advertised one would, and prunes consistently. Default returns `None` (a no-op for everything else).
- Overridden in: Gemini Live (and Vertex via inheritance) `tools=`, Ultravox `one_shot_selected_tools=`, AWS Nova Sonic `tools=`, and OpenAI Realtime (and Azure via inheritance) / Grok/xAI / Inworld `session_properties.tools`. Each override returns its configured tools verbatim; the registration path normalizes them and registers only standard tools (provider-native tools carry no handlers).
- Realtime examples bundle handlers on their schemas (with `@tool_options` for the async ones) and drop the `register_function` calls.
- Documented at each tool-selection site (and on `_service_tools()`) that a service **must** prefer context-provided tools over its own — the invariant the central fallback relies on.

**Service-configured tool params accept a plain list** (consistency with `LLMContext`)
- Those same params now accept a plain list of standard tools (direct functions and/or `FunctionSchema` objects), not just a `ToolsSchema` — matching `LLMContext(tools=...)`, so there's no need to construct a `ToolsSchema` for the common standard-tools case. A list is normalized to a `ToolsSchema` at the configuration boundary (the constructor for Gemini Live / Nova Sonic / Ultravox; a field validator on `session_properties.tools` for the OpenAI/Azure/Grok/Inworld services). Provider-native tool lists (Gemini dicts, Grok/Inworld tool objects) pass through unchanged.
- Realtime examples use the simpler list form wherever the tools are standard-only (mixed standard + custom-tool examples keep `ToolsSchema`).

**Fix — session updates no longer mutate stored settings**
- The OpenAI/Azure/Grok/Inworld realtime services built each `session.update` by mutating the stored `session_properties` in place (converting `tools` from a `ToolsSchema` to the provider list form), which prevented service-configured tool handlers from registering. The outgoing session is now built from a copy, leaving the stored settings intact.

**Logging** (consistent messaging)
- The missing-handler `logger.error` now points developers at `FunctionSchema.handler` / a direct function (the auto-registering paths) instead of `register_function`.

**Examples** (consistent messaging)
- `function-calling-perplexity.py` migrated to a direct function (it predated the direct-function default).
- `function-calling-schema-handler.py` → `function-calling-advanced-functionschema.py`, reframed as the "escape valve" for when a direct function can't express the schema (e.g. a strict `enum`).
- `function-calling-missing-handler.py` refreshed to match the wiring options.
- Removed the redundant `# Direct functions … are registered with the LLM automatically` comment across the examples (~60 files) — a new reader can take it as given.

## Testing

- `uv run pytest tests/test_realtime_service_tool_sync.py` — covers all six services: handler registers from a `ToolsSchema` and from a plain list, `@tool_options` honored, no-tools is a safe no-op, Gemini's raw-dict and context-precedence cases, and (for the session-properties services) that a `session.update` preserves the stored tools and that a tool list is normalized to a `ToolsSchema`.

## Deliberately NOT included: deprecating manual registration

We considered deprecating manual registration (`register_function` / `unregister_function`) outright, since the new auto-registration makes it the non-preferred path. We **chose not to, for now** — it would be too disruptive: a large number of existing users would suddenly start seeing `DeprecationWarning`s. Leaving it as fully-supported, deferred follow-up work.


examples/function-calling/function-calling-perplexity.py
kompfner, line 1: A note to reviewers: this file doesn't seem to be working properly, even on `main` - the system instruction appears to be ignored, and it doesn't seem possible to trigger the function call.

Fixing this already-broken example (or this service, if needed) is outside of the scope of this PR, which is strictly concerned with bringing consistency around function auto-registering.

src/pipecat/services/llm_service.py
```diff
@@ -1069,6 +1070,32 @@ def _warn_if_redundant_manual_registration(self, function_name: str) -> None:
             "step is unnecessary when the handler is on the FunctionSchema."
         )
 
+    def _init_time_tools(self) -> ToolsSchema | None:
```
kompfner: Note to self: I wonder if, instead of referring to these as "init-time" tools, we should refer to them more generally as "settings-based" tools, or something like that, since they are theoretically updatable with an update settings frame.

kompfner: (Hmm, service implementers don't *have* to store them in `_settings` if they're not runtime-updatable. Maybe that's a theoretical concern, but it points to the underlying distinction, which is that these tools come from service configuration and not `LLMContext`. So perhaps "settings-based" would be a bit inaccurate.)

examples/realtime/realtime-openai.py
```diff
@@ -141,16 +141,11 @@ async def run_bot(transport: BaseTransport, runner_args: RunnerArguments):
                 ),
                 # In this example we provide tools through the context, but you could
                 # alternatively provide them here.
-                # tools=[get_current_weather, get_restaurant_recommendation],
+                # tools=ToolsSchema(standard_tools=[get_current_weather, get_restaurant_recommendation]),
```
kompfner: This was a bug in the example.

Though: now that we support passing tool lists directly into the `LLMContext` initializer, we should consider doing the same for all of these LLM service initializer `tools` args...

src/pipecat/services/llm_service.py
kompfner: Fixed with https://github.com/pipecat-ai/pipecat/pull/4758/changes/156afab17af8e4e09cbf5b1b14ebba53ee45f1ec

src/pipecat/services/inworld/realtime/llm.py
```diff
@@ -534,6 +535,20 @@ async def cancel(self, frame: CancelFrame):
...
+    @override
+    def _service_tools(self) -> "ToolsSchema | None":
+        """Return the tools configured on ``session_properties``, if any.
+
+        Inworld Realtime advertises session-properties tools when the context
+        carries none, so their handlers auto-register the same way a
+        context-advertised one would. Provider-native tool objects carry no
+        handler, so only a ``ToolsSchema`` qualifies.
+        """
+        sp = assert_given(self._settings.session_properties)
```
filipi87: In case `self._settings.session_properties` doesn't exist, we currently raise a `RuntimeError`. Is that the expected behavior here?

Or should we just return `None`, as we do in the other LLM services?

src/pipecat/services/openai/realtime/llm.py
```diff
@@ -618,6 +619,20 @@ async def _truncate_current_audio_response(self):
...
+    @override
+    def _service_tools(self) -> "ToolsSchema | None":
+        """Return the tools configured on ``session_properties``, if any.
+
+        OpenAI Realtime advertises session-properties tools when the context
+        carries none, so their handlers auto-register the same way a
+        context-advertised one would. Provider-native tool objects carry no
+        handler, so only a ``ToolsSchema`` qualifies.
+        """
+        sp = assert_given(self._settings.session_properties)
```
filipi87: Same question here. I'm just trying to understand what our recommended approach is.

examples/realtime/realtime-gemini-live-graceful-end.py
```diff
@@ -146,10 +149,6 @@ async def run_bot(transport: BaseTransport, runner_args: RunnerArguments):
         tools=tools,
     )
 
-    llm.register_function("get_current_weather", fetch_weather_from_api)
-    llm.register_function("get_restaurant_recommendation", fetch_restaurant_recommendation)
-    llm.register_function("end_conversation", end_conversation)
```
filipi87, line 151: So much better with all of this removed. 🚀

filipi87 (approved):
Makes sense to me. Nice cleanup. 🚀

src/pipecat/services/inworld/realtime/llm.py
kompfner: `self._settings` is actually guaranteed to have all fields filled out, with none of them being `NOT_GIVEN`. `NOT_GIVEN` can only appear in fields when a `ServiceSettings` object is used as a "delta" to update runtime settings, not when it's in "store" mode (as it is in `self._settings`).

kompfner: I'm actually working on a change right now to all of these `_service_tools` methods, though, so stay tuned.

examples/realtime/realtime-openai.py
kompfner: Done: https://github.com/pipecat-ai/pipecat/pull/4758/changes/cd65aaa70275abcbde0d4cfc897b2188e5674c53

src/pipecat/services/inworld/realtime/llm.py
kompfner: Here's the change: https://github.com/pipecat-ai/pipecat/pull/4758/changes/cd65aaa70275abcbde0d4cfc897b2188e5674c53

src/pipecat/services/openai/realtime/llm.py
kompfner: See previous commit thread.

PR #4759, opened 2026-06-15
Preparing STT services to support context carryover.

filipi87:
## Summary

  - Added `LLMContextAssistantTurnFrame`, a new frame broadcast by `LLMAssistantAggregator` when a bot turn completes, carrying the aggregated reply text, start timestamp, and whether the turn was interrupted
  - Added `STTService.process_assistant_turn(text, interrupted)` hook that STT subclasses can override to react to completed bot replies —  for example, to send the bot's reply to a provider-side context carryover API
  - `STTService` now handles `LLMContextAssistantTurnFrame` automatically and routes it to `process_assistant_turn()`


src/pipecat/frames/frames.py
```diff
@@ -499,6 +499,26 @@ class LLMContextAssistantTimestampFrame(DataFrame):
...
+
+    Parameters:
+        text: The assistant's aggregated spoken text for this turn.
+        timestamp: ISO-8601 timestamp of when the assistant turn started.
+        interrupted: Whether the turn was cut short by an interruption.
+    """
+
+    text: str
+    timestamp: str
+    interrupted: bool = False
```
markbackman: What purpose does `interrupted` serve? Would this be used in the STT service logic? Or, is it included as metadata from the assistant_turn_stopped event that we're passing on?

changelog/4759.added.md
```diff
@@ -0,0 +1,2 @@
+- Added `LLMContextAssistantTurnFrame`, broadcast by `LLMAssistantAggregator` when a bot turn completes, carrying the aggregated reply text, start timestamp, and whether it was interrupted.
+- Added `STTService.process_assistant_turn(text, interrupted)` hook that subclasses can override to feed the completed bot reply to a provider-side context carryover API. The base implementation is a no-op; `STTService` now handles `LLMContextAssistantTurnFrame` and calls this method automatically.
```
markbackman: Break this up into a second changelog, 4759.added.2.md.

src/pipecat/frames/frames.py
```diff
@@ -499,6 +499,26 @@ class LLMContextAssistantTimestampFrame(DataFrame):
...
+    downstream (e.g. STT services) can handle this frame to react to each
+    completed bot reply without needing a separate observer.
+
+    Parameters:
+        text: The assistant's aggregated spoken text for this turn.
+        timestamp: ISO-8601 timestamp of when the assistant turn started.
+        interrupted: Whether the turn was cut short by an interruption.
+    """
+
+    text: str
```
markbackman, line 516: Do we assume that the APIs will only take the last turns text or might they take a list of messages? Specifically, I'm wondering if this should carry the context so the STTService can provide more context, in the event that it's needed.

markbackman (commented):
I like this approach. Just a few questions.

src/pipecat/frames/frames.py
markbackman: Oh, I see. Assembly is using the turn by turn information to build their own local version of the context. This is smart.

I think what you have makes sense. We could add more information (e.g. context) later if it's needed. `text` will be easier to work with to start.

src/pipecat/frames/frames.py
filipi87: Right now, we're not using this for anything. 

I am not sure if it could somehow be useful for the STT service logic, or even for any other processor that might use this frame in the future. So I added it here to kind of mimic what we were sending in `on_assistant_turn_stopped`.

But yes, we're not using it for anything at the moment.

src/pipecat/frames/frames.py
filipi87: Right now, they are only using what the agent last said, which is why I implemented it this way.

Also, if they ever need the full context, they can process the `LLMContextFrame` directly.

changelog/4759.added.md
filipi87: Done.

src/pipecat/frames/frames.py
markbackman: If we don't have a need for it, maybe we should remove it? Then, only add it when it's needed? Rationale being that it's easier to add an API than remove.

src/pipecat/frames/frames.py
filipi87: Yep, makes sense. I have removed it.

markbackman (approved):
LGTM

PR #4760, opened 2026-06-15
feat(cli): route eval output to a gitignored eval-runs/ folder

markbackman:
## Summary

- Scaffolded projects now gitignore eval run output: `*.eval.log`, `*.debug.log`, and the `eval-runs/` run-output dir (the latter catches `.wav` recordings that the existing generic `*.log` rule misses).
- The vended `AGENTS.md` (§6) now tells agents to route eval runs into the gitignored `eval-runs/` with `--logs-dir eval-runs` / `--record-dir eval-runs`, and notes that `pipecat eval suite` already organizes runs into a timestamped `eval-runs/<timestamp>/`.

Keeps `pipecat eval run` artifacts from cluttering generated projects, without limiting where output can go.


filipi87 (approved):
LGTM!

PR #4761, opened 2026-06-16
feat(pipeline): add on_heartbeat_timeout event handler to PipelineWorker

### LINKED ISSUE #3648: Heartbeat event handler
### pipecat version

0.0.100

### Python version

_No response_

### Operating System

_No response_

### what we have tried: 
As a workaround, we subclassed `PipelineTask` and overrode the entire `_heartbeat_monitor_handler` method to add our custom logic. It works, but it's fragile since we're duplicating the entire method implementation and it could break with future updates.

### Question

Hey!

We just started using heartbeats in our pipeline - really helpful feature for detecting blocked pipelines.

One thing we noticed: when a timeout occurs (no heartbeat frame received within 10 seconds), the only feedback is a warning log. There's no way for consumers to react programmatically - for example, to emit metrics, trigger alerts, or cancel the pipeline.

Other timeout scenarios like idle detection have an `on_idle_timeout` event handler that consumers can subscribe to. It would be great to have the same pattern for heartbeat timeouts.

## Proposal

1. Add an `on_heartbeat_timeout` event handler that fires when a heartbeat frame is not received within the expected time window
2. Make the heartbeat monitor timeout configurable (currently hardcoded to 10x the heartbeat interval via `HEARTBEAT_MONITOR_SECS`)

The changes are minimal but definitely impactful.

Happy to create a PR for this if the approach sounds good!

### Context

_No response_

nuthalapativarun:

Adds an `on_heartbeat_timeout` event handler to `PipelineWorker`, fired whenever a heartbeat frame is not received within `heartbeats_monitor_secs`. This mirrors the existing `on_idle_timeout` pattern and lets application code react to stalled pipelines without polling.

Closes #3648

**Before:** The only observable signal for a missed heartbeat was a log warning emitted internally. There was no way for application code to hook into this condition.

**After:** Application code can register a handler:

```python
@worker.event_handler("on_heartbeat_timeout")
async def on_heartbeat_timeout(worker):
    logger.warning("Heartbeat timeout — pipeline may be stalled")
    # e.g. send an alert, trigger a restart, etc.
```


nuthalapativarun:
Friendly bump on this one — closes #3648 and has been sitting for about a week without review. Let me know if there's anything I can do to help move it along. Thanks!

filipi87 (approved):
LGTM! Thanks for the improvement, @nuthalapativarun.

I pushed a couple of docstring improvements, updated an example to use it, and added some automated tests.

PR #4762, opened 2026-06-16
fix(azure): omit output_modalities from Azure Realtime response.create

nuthalapativarun:
#### Please describe the changes in your PR. If it is addressing an issue, please reference that as well.

Azure OpenAI Realtime endpoints reject `response.output_modalities` in `response.create` events with a 400 error. This field is specific to OpenAI's API and is not supported by Azure.

Closes #4106

**Root cause:** `OpenAIRealtimeLLMService._create_response()` always includes `output_modalities` in the `ResponseCreateEvent`. `AzureRealtimeLLMService` inherits this without the ability to opt out.

**Fix:** Extract a `_build_response_create_event()` hook in `OpenAIRealtimeLLMService` that `AzureRealtimeLLMService` overrides to return a bare `ResponseCreateEvent()` without the unsupported field.

**Before:**
```python
# Azure received: {"type": "response.create", "response": {"output_modalities": ["text", "audio"]}}
# → 400 Bad Request from Azure endpoint
```

**After:**
```python
# Azure receives: {"type": "response.create"}
# → Accepted by Azure endpoint
```

PR #4763, opened 2026-06-16
fix(azure): remove unsupported output_modalities from realtime requests

### LINKED ISSUE #4106: AzureRealtimeLLMService sends response.output_modalities which Azure doesn't support

nuthalapativarun:
## Summary

- Azure's Realtime API rejects the `output_modalities` parameter in both `session.update` and `response.create` requests, causing the connection to fail with an unsupported parameter error.
- Overrides `send_client_event` in `AzureRealtimeLLMService` to strip `output_modalities` from `SessionUpdateEvent` and `ResponseCreateEvent` before transmission, using `model_copy` so the stored configuration is unchanged.
- Adds a unit test verifying the field is absent from the serialized wire payload in both event types.

Closes #4106

## Test plan

- [ ] `python -m pytest tests/test_azure_realtime_output_modalities.py -v` -- 3 tests pass
- [ ] `python -m ruff check src/pipecat/services/azure/realtime/llm.py` -- no issues
- [ ] `python -m ruff format --check src/pipecat/services/azure/realtime/llm.py` -- already formatted

### DIFF
--- a/changelog/4763.fixed.md
+++ b/changelog/4763.fixed.md
@@ -1,0 +1,1 @@
+- Fixed `AzureRealtimeLLMService` sending unsupported `output_modalities` parameter in `session.update` and `response.create` requests, which caused Azure's Realtime API to reject the connection (#4106).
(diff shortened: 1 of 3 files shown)


nuthalapativarun:
Closing as duplicate of #4762 which covers the same fix.

PR #4764, opened 2026-06-16
fix(tts): drop stale word-timestamps from completed and cleared TTS contexts

### LINKED ISSUE #4750: AggregatedFrameSequencer: stale word-timestamps from interrupted Inworld context corrupt next-turn transcript

jamsea:
## The bug (#4750)

The word-alignment subsystem under `src/pipecat/utils/context/` maps each streamed TTS word-timestamp back onto the original LLM text so the assistant transcript reflects what was spoken. When a TTS provider keeps streaming word-timestamps for a context that has already ended, those late words reach `AggregatedFrameSequencer.process_word` with a `context_id` whose turn is gone. They fail to match the active slot, fall to the passthrough path, and get emitted as `TTSTextFrame`s with `append_to_context=True`. The result: stale words interleave word-by-word into the NEXT turn's transcript and pollute the LLM context.

Symptom log:
- `aggregated_frame_sequencer:process_word ... Word 'X' not recognised by any slot, emitting as passthrough`

## Why #4751 is only a partial fix

PR #4751 discriminates stale words by `context_id not in self._context_append_to_context`. That dict is added to in `register_spoken` and removed ONLY in `clear()` (interruption). It is NEVER removed on normal turn completion. So #4751 covers the interrupt path but NOT the normal-turn-boundary path: a straggler word-timestamp delivered after a turn finished normally still leaks, because the prior context is still present in the dict.

This PR covers BOTH paths:
1. Interrupt then late straggler (already handled via `clear()`).
2. Normal turn completion then late straggler during the next turn (the gap).

## Approach

Track live context IDs explicitly with a real liveness signal instead of relying on a side effect of the append-value dict:

- `_live_context_ids` is populated in `register_spoken`.
- It is retired in BOTH `clear()` (interruption) and a new `retire_context()` (normal turn completion).
- `retire_context()` is wired into `TTSService._audio_context_task_handler`, right where the audio context is torn down (`del self._audio_contexts[context_id]`). `_handle_audio_context` has already fully drained the turn's queued words (and force-completed its slots) before that point, so retiring there cannot drop in-flight words. Any word the provider streams afterward is a stale straggler.
- `process_word` drops a word when `context_id is not None and context_id not in self._live_context_ids`.

Preserved behavior (verified against the real call sites):
- `context_id=None` passthrough (services without audio contexts) is untouched.
- The `push_text_frames=True` path completes slots via `complete_spoken_slot` and builds its `TTSTextFrame` directly (not via `_build_word_frame`), so it is unaffected.
- Legit passthrough is preserved: a word a LIVE context's tracker does not recognise still passes through (`test_unrecognised_word_emits_passthrough` stays green). Only DEAD/stale contexts get dropped.
- Kept #4751's `_build_word_frame` default of `False` for unknown contexts as defense in depth (this PR supersedes #4751).

Observability: drops emit a `warning` log and increment a `_dropped_stale_words` counter so residual leaks are greppable in prod.

## Reproductions

- Original report: Inworld TTS.
- Independent repro: ElevenLabs on the stock example `examples/voice/voice-elevenlabs.py` (Discord ticket T-2159). That example does NOT set `filter_incomplete_user_turns`; its `LLMUserAggregatorParams` only passes a VAD analyzer, so the trigger is ordinary interruptions, not that flag.

## Tests

Added regression coverage in `tests/test_aggregated_frame_sequencer.py`:
- `test_stale_words_do_not_corrupt_next_turn_normal_boundary` (the gap #4751 does not cover): register ctxA, complete it normally with NO `clear()`, `retire_context`, register ctxB, deliver a late ctxA word, assert `[]` and that ctxB still flows.
- `test_stale_words_do_not_corrupt_next_turn_after_interrupt` (interrupt path stays green).
- `test_word_after_retire_is_dropped`, `test_retire_removes_liveness_only`, `test_retire_none_is_noop`, `test_other_live_context_unaffected_by_retire`, `test_unknown_context_is_dropped`, plus `None`-context passthrough tests.

Run via the repo `.venv`:

```
tests/test_aggregated_frame_sequencer.py
tests/test_word_completion_tracker.py
tests/test_tts_frame_ordering.py
tests/test_cartesia_tts.py
-> 234 passed, 1 warning
```

`ruff format` and `ruff check` on the changed files: clean.

## Known separate follow-up (NOT in this PR)

There is a separate `WordCompletionTracker` case-fold bug that produces the secondary warning `WordCompletionTracker: llm_consumed 'what' does not contain frame_word 'What', discarding`. The membership check in `word_completion_tracker.py` uses `_fold_typography` (which only swaps curly quotes and dashes) but does NOT lowercase, so a case difference like `What` vs `what` discards `_llm_consumed`. That is a distinct bug and will be handled in a separate PR. Scoping this one to the sequencer staleness fix only.

Fixes #4750. Supersedes the partial fix in #4751.

### DIFF
--- a/changelog/4750.fixed.md
+++ b/changelog/4750.fixed.md
@@ -1,0 +1,8 @@
+- Fixed stale word-timestamps leaking into the next turn's transcript in
+  `AggregatedFrameSequencer`. When a TTS provider keeps streaming word-timestamps for
+  a context that has already ended, those late words were emitted as passthrough
+  `TTSTextFrame`s with `append_to_context=True`, interleaving word-by-word into the
+  following turn and polluting the LLM context. The sequencer now tracks live
+  context IDs and drops word-timestamps for any context that has been retired, both
+  on interruption (`clear`) and on normal turn completion (`retire_context`).
+  Reproduced with Inworld TTS (issue) and ElevenLabs (`examples/voice/voice-elevenlabs.py`).
--- a/changelog/4750.fixed.md
+++ b/changelog/4750.fixed.md
@@ -5,4 +5,4 @@
   following turn and polluting the LLM context. The sequencer now tracks live
   context IDs and drops word-timestamps for any context that has been retired, both
   on interruption (`clear`) and on normal turn completion (`retire_context`).
-  Reproduced with Inworld TTS (issue) and ElevenLabs (`examples/voice/voice-elevenlabs.py`).
+  Reproduced with Inworld TTS and ElevenLabs (`examples/voice/voice-elevenlabs.py`).
--- a/src/pipecat/services/tts_service.py
+++ b/src/pipecat/services/tts_service.py
@@ -552,6 +552,11 @@ class TTSService(AIService):
             await self._serialization_queue.put(None)
             await self._audio_context_task
             self._audio_context_task = None
+        # The None sentinel exits the drain loop without running the per-context
+        # retire_context branch, so any context still queued at stop() keeps its
+        # liveness entry. Clear the sequencer so no stale context_id survives if
+        # this instance is reused (e.g. StopFrame keeps processors alive).
+        self._aggregated_frame_sequencer.clear()
 
     async def cancel(self, frame: CancelFrame):
         """Cancel the TTS service.
@@ -561,6 +566,9 @@ class TTSService(AIService):
         """
         await super().cancel(frame)
         await self._stop_audio_context_task()
+        # Cancelling the task skips retire_context for any in-flight context, so
+        # clear the sequencer to avoid leaving stale liveness behind.
+        self._aggregated_frame_sequencer.clear()
 
     def add_text_transformer(
         self,
@@ -1432,7 +1440,13 @@ class TTSService(AIService):
                 await self._handle_audio_context(context_id)
 
                 # We just finished processing the context, so we can safely remove it.
+                # _handle_audio_context() above has already drained every queued word
+                # for this context (and force-completed its slots), so retiring it now
+                # cannot drop in-flight words. Any word-timestamp the provider streams
+                # after this point is a stale straggler for a completed turn, and the
+                # sequencer drops it instead of leaking it into the next turn.
                 del self._audio_contexts[context_id]
+                self._aggregated_frame_sequencer.retire_context(context_id)
                 await self.on_audio_context_completed(context_id=context_id)
                 self.reset_active_audio_context()
             else:
--- a/src/pipecat/utils/context/aggregated_frame_sequencer.py
+++ b/src/pipecat/utils/context/aggregated_frame_sequencer.py
@@ -66,6 +66,13 @@ class AggregatedFrameSequencer:
         self._name = name
         self._slots: list[_AggregatedFrameSlot] = []
         self._context_append_to_context: dict[str, bool] = {}
+        # Context IDs that are currently live (registered and not yet retired). A
+        # context is live from register_spoken() until it is retired by either
+        # clear() (interruption) or retire_context() (normal turn completion). Words
+        # arriving for a context that is not in this set are stale and are dropped.
+        self._live_context_ids: set[str] = set()
+        # Count of stale word-timestamps dropped, for prod observability.
+        self._dropped_stale_words: int = 0
 
     def register_spoken(
         self,
@@ -93,6 +100,7 @@ class AggregatedFrameSequencer:
                 do not inject extra spaces between consecutive frames.
         """
         self._context_append_to_context[context_id] = append_to_context
+        self._live_context_ids.add(context_id)
         self._slots.append(
             _AggregatedFrameSlot(
                 frame=frame,
@@ -146,6 +154,9 @@ class AggregatedFrameSequencer:
         Locates the active (first incomplete spoken) slot with a tracker, advances it
         by the incoming word, and builds a :class:`TTSTextFrame`. Handles:
 
+        - Words from a context that is no longer live (retired by :meth:`clear` on
+          interruption or by :meth:`retire_context` on normal turn completion):
+          dropped as stale (returns an empty list).
         - Normal words that fit entirely within the active slot.
         - Overflow words straddling two slot boundaries.
         - Force-complete when the TTS drops an event (word belongs to the next slot).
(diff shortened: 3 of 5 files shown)


jamsea:
Addressed the Copilot review feedback: `retire_context()` now also pops the `_context_append_to_context` entry, not just the liveness marker, so neither per-context map grows without bound over a long-running service. This is safe because the TTS service force-completes the context's slots (stamping `append_to_context` onto every emitted frame) before the audio-context teardown calls `retire_context`, so no later `_build_word_frame` needs the value. Updated the unit test accordingly (renamed to `test_retire_cleans_up_per_context_metadata`, now asserts both maps are cleaned). All 234 tests across the four TTS/sequencer suites pass; ruff clean.

jamsea:
Applied two findings from a self-review pass:

1. **Hot-path log volume:** the stale-word drop in `process_word` now logs at `debug` instead of `warning`. Providers that stream stragglers (Inworld, ElevenLabs) emit many per turn, so a per-word warning would flood prod logs. The running total stays in the message for triage.

2. **Teardown liveness leak:** `cancel()` and `stop()` tore down the audio-context task without running the per-context `retire_context` branch (and without `clear()`, which only `_handle_interruption` calls), leaving stale context IDs in the sequencer. Both now call `_aggregated_frame_sequencer.clear()`, so no stale liveness survives if the instance is reused (e.g. a StopFrame that keeps processors alive). Added `test_cancel_clears_sequencer_liveness` and `test_stop_clears_sequencer_liveness`.

All 236 tests across the four TTS/sequencer suites pass; ruff clean.

filipi87 (commented):
Hi @jamsea, I believe the changes already pushed in this previous PR should be sufficient:
- https://github.com/pipecat-ai/pipecat/pull/4751

markbackman:
Fixed in #4751. Closing.

PR #4766, opened 2026-06-16
fix(dtmf): key cached DTMF audio by sample rate, not just button

he-yufeng:
## What

`load_dtmf_audio()` caches decoded/resampled tones in a process-global dict keyed only by the button. The sample rate determines how the audio is produced but is not part of the cache key, so the first rate a button is requested at wins for the rest of the process.

So if a process first loads a tone at 8 kHz and later asks for the same button at 16 kHz, it gets the cached 8 kHz audio back. A transport running at a different rate than the first caller ends up emitting wrong-rate DTMF (wrong pitch/duration), which downstream telephony providers may not detect.

## Fix

Key the cache by `(button, sample_rate)` instead of just `button`. No behavior change for single-rate processes; multi-rate processes now get audio resampled correctly per rate. About four lines touched.

## Verifying

Added `tests/test_dtmf_utils.py`. The main test loads button `1` at 8 kHz then 16 kHz and asserts the 16 kHz buffer is roughly twice the bytes:

- Before: the 16 kHz request returns the cached 8 kHz buffer (8000 bytes), assertion fails.
- After: the 16 kHz request returns ~16000 bytes, passes. Re-fetching 8 kHz still returns the cached 8 kHz bytes.

```
python -m pytest tests/test_dtmf_utils.py tests/test_dtmf_aggregator.py -q
```

9 passed. `ruff check` and `ruff format --check` are clean on the touched files.

### DIFF
--- a/changelog/4766.fixed.md
+++ b/changelog/4766.fixed.md
@@ -1,0 +1,1 @@
+- Fixed cached DTMF audio being returned at the wrong sample rate. `load_dtmf_audio()` keyed its in-memory cache by button only, so the first sample rate a button was loaded at was reused for every later request. The cache is now keyed by `(button, sample_rate)`.
--- a/src/pipecat/audio/dtmf/utils.py
+++ b/src/pipecat/audio/dtmf/utils.py
@@ -23,7 +23,7 @@ from pipecat.audio.resamplers.base_audio_resampler import BaseAudioResampler
 from pipecat.audio.utils import create_file_resampler
 
 __DTMF_LOCK__ = asyncio.Lock()
-__DTMF_AUDIO__: dict[KeypadEntry, bytes] = {}
+__DTMF_AUDIO__: dict[tuple[KeypadEntry, int], bytes] = {}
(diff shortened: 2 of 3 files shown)


markbackman:
Out of curiosity, what is your use case that you need to modify the sample rate within a session? The norm is to set the sample rate in the PipelineParams, which the StartFrame uses to inform all pipeline processors. Varying after start is not something I've seen. So, while this fix may be correct, I'm not sure it's practical or needed. I may be overlooking use cases though.

he-yufeng:
Friendly nudge on this one — it's been open about a month with green CI. The bug is small but real: `load_dtmf_audio()`'s process-global cache is keyed only by the button, so whichever sample rate a button is first requested at wins for the rest of the process, and later requests at a different rate get mis-resampled tones. The fix just folds the sample rate into the cache key. Happy to rebase if it's drifted — would appreciate a look whenever someone has a moment.

markbackman (approved):
LGTM! Sorry for the delay in reviewing; we've had an influx of PRs over the last few months. Thanks for this nice, scoped improvement.

PR #4770, opened 2026-06-16
feat: add KeenableWebSearch service for web search via MCP

ilya-bogin-keenable:
Adds an optional, standalone service that gives voice agents live web search
through a hosted MCP server (Keenable AI).

Keenable is added as a separate service used like any other MCP tool provider,
rather than a flag on the core `LLMService`. No changes to the base class,
adapters, or core dependencies.

## Usage

```python
from pipecat.services.keenable.search import KeenableWebSearch

async with KeenableWebSearch() as search:
    tools = await search.register_tools(llm)
    context = LLMContext(messages=[...], tools=tools)
```

`KeenableWebSearch` is a thin wrapper around the existing `MCPClient`
(Streamable HTTP). Keyless requests use the `pro` mode; set `KEENABLE_API_KEY`
for higher rate limits and the lower-latency `realtime` mode — a good fit for
voice (roughly 0.3s vs 2.5s for `pro` in our testing). `realtime` requires an
enabled account, so it's the default only when a key is present (keyless stays
on `pro`); override either way with `KEENABLE_SEARCH_MODE`.

## What's included

- `pipecat.services.keenable.search.KeenableWebSearch`
- `keenable` install extra (depends on `pipecat-ai[mcp]`)
- Feature example: `examples/features/features-keenable-web-search.py`
- Unit tests (config/header/connection, no network)
- Changelog entry

## Testing

- `ruff check` / `ruff format` clean; `pyright` clean; 12 unit tests pass.
- Live smoke against the production MCP endpoint, keyless and keyed: connects,
  discovers both tools (`search_web_pages`, `fetch_page_content`), and a real
  `search_web_pages` call returns results end-to-end.
- LLM-in-the-loop run (Anthropic): the model calls `search_web_pages` on a live
  query and answers from the returned results.

Disclosure: I work at Keenable. Happy to adjust naming, structure, or the
example to fit your conventions.


markbackman:
Closing in favor of https://github.com/pipecat-ai/pipecat/pull/4923

PR #4772, opened 2026-06-16
Soniox v5

mihafabcic-soniox:
Adds `endpoint_sensitivity` (float [-1.0, 1.0]) to `SonioxSTTService.Settings`, a new parameter on the v5 model that controls how aggressively endpoints fire. Also bumps the default model from `stt-rt-v4` to `stt-rt-v5`.


markbackman (approved):
Very exciting! Looking forward to benchmarking the v5 model.

PR #4773, opened 2026-06-16
feat(nvidia): add per-sentence TTS mode and zero-shot prompt support

filipi87:
feat(nvidia): add per-sentence TTS mode and zero-shot prompt support

### DIFF
--- a/src/pipecat/services/nvidia/tts.py
+++ b/src/pipecat/services/nvidia/tts.py
@@ -5,9 +5,9 @@
 # SPDX-License-Identifier: BSD 2-Clause License
 #
 
-"""NVIDIA Nemotron Speech text-to-speech service implementation.
+"""NVIDIA text-to-speech service implementation.
 
-This module provides integration with NVIDIA Nemotron Speech's TTS services through
+This module provides integration with NVIDIA TTS through
 gRPC API for high-quality speech synthesis.
 
 Refer to the NVIDIA TTS NIM documentation for usage, customization,
@@ -22,6 +22,8 @@ import textwrap
 import threading
 from collections.abc import AsyncGenerator, Mapping
 from dataclasses import dataclass, field
+from enum import StrEnum
+from pathlib import Path
 from typing import Any
 
 from pipecat.utils.tracing.service_decorators import traced_tts
@@ -41,7 +43,7 @@ from pipecat.frames.frames import (
     TTSAudioRawFrame,
     TTSStartedFrame,
 )
-from pipecat.services.settings import NOT_GIVEN, TTSSettings, _NotGiven
+from pipecat.services.settings import NOT_GIVEN, TTSSettings, _NotGiven, is_given
 from pipecat.services.tts_service import TTSService
 from pipecat.transcriptions.language import Language
 from pipecat.utils.deprecation import deprecated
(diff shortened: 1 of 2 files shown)


src/pipecat/services/nvidia/tts.py
```diff
@@ -575,3 +681,66 @@ async def run_tts(self, text: str, context_id: str) -> AsyncGenerator[Frame | No
...
+        event_loop = self.get_event_loop()
+
+        def run_grpc():
+            base_req = self._build_base_request()
+
+            def request_gen():
+                for chunk in chunks:
+                    base_req.text = chunk
+                    yield base_req
+
```
sathwikareddy02, line 714: For the non-stitching models we discussed, this is the main compatibility issue we were trying to avoid. If a long sentence must be split because of request-size limits, each chunk still needs its own fresh SynthesizeOnline call, not multiple requests inside one call.
So this implementation is still effectively "one RPC per run_tts", not "one RPC per chunk" for the models that require strict per-sentence/per-request behavior.
Suggested direction:
keep per_sentence as a separate path
but move the RPC creation inside a loop over chunks, so each chunk gets its own call
yield all audio for chunk 1, then close that call, then open a new call for chunk 2, etc.


Suggested fix:
```python
for chunk in chunks:
    def request_gen():
        base_req = self._build_base_request()
        base_req.text = chunk
        yield base_req

    call = self._service.stub.SynthesizeOnline(
        request_gen(),
        metadata=self._service.auth.get_auth_metadata(),
    )
    self._per_sentence_rpc_call = call
    for resp in call:
        asyncio.run_coroutine_threadsafe(response_queue.put(resp), event_loop)
```

src/pipecat/services/nvidia/tts.py
```diff
@@ -575,3 +681,66 @@ async def run_tts(self, text: str, context_id: str) -> AsyncGenerator[Frame | No
...
+            try:
+                call = self._service.stub.SynthesizeOnline(
+                    request_gen(),
+                    metadata=self._service.auth.get_auth_metadata(),
+                )
+                self._per_sentence_rpc_call = call
+                for resp in call:
+                    asyncio.run_coroutine_threadsafe(response_queue.put(resp), event_loop)
+            except Exception as e:
```
sathwikareddy02, line 723: I think intentional cancellation can still surface here as an error in `per_sentence` mode.

`_cancel_per_sentence_call()` cancels the gRPC call during interruption/stop, but `run_grpc()` catches all exceptions and always pushes them into `response_queue`:

```python
except Exception as e:
    asyncio.run_coroutine_threadsafe(response_queue.put(e), event_loop)
```

That means a normal interruption-driven cancel may come back through _run_tts_per_sentence() as an exception and get raised as an ErrorFrame, even though teardown was expected.
I think we need a local cancellation signal here, similar to the stitched path, so interruption/stop cancels the RPC quietly instead of reporting a synthesis failure.

Suggested fix:

```python
stop_event = threading.Event()
```

Pass it into run_grpc() and suppress cancellation-driven exceptions:
```
except Exception as e:
    if not stop_event.is_set():
        asyncio.run_coroutine_threadsafe(response_queue.put(e), event_loop)
```
And in teardown:
```
stop_event.set()
self._cancel_per_sentence_call()
```

sathwikareddy02 (changes requested):
I added 2 review comments on the draft PR for the two places that I think still need adjustment:
- per-sentence mode still sending multiple split chunks through one `SynthesizeOnline` call
- interruption/cancellation in per-sentence mode potentially surfacing as an error

Please take a look. If those are addressed, I think the draft should be aligning with our approach.

PR #4775, opened 2026-06-16
Add screenAudio output support to DailyTransport

aconchillo:
## Summary

Adds support for publishing a `screenAudio` output track in `DailyTransport`, following the existing `screenVideo` pattern. `daily-python` 0.29.0+ supports `screenAudio`.

## Changes

- `register_audio_destination` now special-cases `"screenAudio"` (registering it as Daily's built-in screen-audio input track via `update_inputs` + `update_publishing`) and routes everything else through a new `_register_custom_audio_destination` helper, mirroring how `register_video_destination` handles `screenVideo`.
- Added `_register_screen_audio_destination` mirroring `_register_screen_video_destination`.
- `remove_custom_audio_track` guards `"screenAudio"` with an early return (like the existing `screenVideo` guard), since it's a built-in input track rather than a removable custom track.
- Factored out a `_create_audio_track` helper (mirroring `_create_video_track`); `add_custom_audio_track` and `_register_screen_audio_destination` now share track creation.
- Bumped the `daily` extra to `daily-python>=0.29.0` (first version with `screenAudio`).

The capture (input) side already handled `screenAudio`, and `write_audio_frame` already routes by `transport_destination`, so no change was needed there.

## Usage

Add `"screenAudio"` to `DailyParams.audio_out_destinations` (optionally configure it via `custom_audio_track_params["screenAudio"]`), then write audio frames with `transport_destination="screenAudio"`.

## Test plan

- `ruff check` and `ruff format --check` are clean.
- Verified the transport imports against the locked `daily-python==0.29.1`.


filipi87 (approved):
LGTM!

PR #4776, opened 2026-06-16
Add evals extra so pipecat eval runs out of the box

aconchillo:
## Summary

Adds a dedicated `evals` extra so `uv run pipecat eval run` works out of the box, without bloating the lean `cli` extra. Also namespaces the eval model caches under `~/.cache/pipecat`.

### `evals` extra
`pipecat eval` needs the CLI tooling plus the harness's default local, no-API-key models: Kokoro (synthesizes the user's turns) and Moonshine (transcribes the bot's audio). The new extra bundles them:

```toml
evals = [ "pipecat-ai[cli]", "pipecat-ai[kokoro]", "pipecat-ai[moonshine]" ]
```

`[cli]` stays lean for plain scaffolding (`pipecat init`) and base images — kokoro/moonshine drag in heavy deps (onnxruntime-gpu, espeak via phonemizer, PortAudio via sounddevice) that shouldn't burden every CLI install.

### Generated projects use the extra
When a scaffolded project enables evals, it now depends on `pipecat-ai[evals]` instead of hardcoding the `cli`/`kokoro`/`moonshine` set, keeping the eval dependency list defined in one place.

### Cache namespacing
- Kokoro model cache: `~/.cache/kokoro-onnx` → `~/.cache/pipecat/kokoro-onnx`
- Eval TTS cache: `~/.cache/pipecat/tts` → `~/.cache/pipecat/evals/tts`

## Test plan

- `uv run pytest tests/cli/test_project_generation.py` passes.
- `ruff check` / `ruff format` clean.
- Verified the `evals` extra resolves to `cli + kokoro + moonshine` deps via `uv lock`.


markbackman (approved):
Nice catch!

PR #4777, opened 2026-06-16
Rename process_assistant_turn to _process_assistant_turn in STT service.

filipi87:
Rename process_assistant_turn to _process_assistant_turn in STT service
  
It's an internal method not intended for external use, so the leading underscore signals that clearly. Simple and accurate.


markbackman (approved):
👍

PR #4780, opened 2026-06-16
Add trigger_disconnect eval flag

aconchillo:
## Summary

Bots often cancel their pipeline in their `on_client_disconnected` handler. During evals that tears the bot down on every scenario disconnect, so the eval transport now **suppresses that event by default**. A new `trigger_disconnect` flag lets the harness fire it on demand to exercise the bot's disconnect path.

## Changes

- **Base transport** (`SingleClientWebsocketServerTransport`): extracted the `on_client_disconnected` event firing into an overridable `_emit_client_disconnected()` seam, so subclasses can suppress the event while keeping connection teardown. No behavior change.
- **Eval transport**: `EvalTransport._emit_client_disconnected` fires `on_client_disconnected` only when the connection carries `?trigger_disconnect=true`. Default off; the recording flush and `set_client_connection(None)` still run.
- **Scenario schema**: top-level `trigger_disconnect: bool = False` field, so a single scenario can opt in mid-suite.
- **Harness**: `EvalSession` accepts `trigger_disconnect` (OR'd with the scenario field) and appends the query flag in `_connect_url`.
- **CLI**: `pipecat eval run --trigger-disconnect`.
- **Docs** (`agent_templates/AGENTS.md` §6): the bot now stays alive between runs, so boot once and iterate; documented `--trigger-disconnect` / `trigger_disconnect:` for exercising the disconnect path, plus a note that context carries over (reset per run with `context:`).

## Relationship to `--stop-bot`

These are orthogonal and both kept:

| | mechanism | guarantees teardown? | runs bot's handler? |
|---|---|---|---|
| `--stop-bot` | `eval-cancel` → `CancelWorkerFrame` | yes | no |
| `--trigger-disconnect` | fire `on_client_disconnected` | no (handler's choice) | yes |

`--stop-bot` was deliberately not reused as the trigger, because its contract ("the bot reliably stops") can only be guaranteed by `eval-cancel`, which bypasses the handler.

## Test plan

- Eval/scenario/websocket + CLI tests pass; `ruff check` / `ruff format` clean.
- `pipecat eval run --help` shows `--trigger-disconnect`; `_query_flag` reads `?trigger_disconnect=true` → `True` (absent → `False`).

## Notes

- No changelog: the evals framework is unreleased, and the base-transport change is a private, behavior-preserving extract-method.
- Rebased on `main` (now includes the `context:` rename, #4781).

### DIFF
--- a/src/pipecat/cli/agent_templates/AGENTS.md
+++ b/src/pipecat/cli/agent_templates/AGENTS.md
@@ -212,7 +212,7 @@ The dev runner wraps this in the eval transport + serializer when you pass `-t e
 - **Logs** — the bot logs to stdout (loguru). However you run it, keep the output durable and greppable — e.g. `uv run bot.py -t eval 2>&1 | tee /tmp/pipecat-output.txt` — so when a scenario fails you can grep the file for the traceback instead of re-running. If your harness captures background-process output natively, that works too.
 - **Clean boot** — `uv run bot.py -t eval` boots the bot as a headless eval WebSocket server (default `ws://localhost:7860`). No exceptions + pipeline assembled is your fastest "did I wire it right" signal.
 
-**The eval loop (where you live).** Boot the bot (`uv run bot.py -t eval`), then drive a scenario against it from a second terminal. The bot serves that one eval session and exits when the run finishes, so boot it fresh for each run as you iterate (or use `pipecat eval suite` to spawn a fresh bot per scenario from a manifest). Start in **text mode** (the default), the fast inner loop. A minimal scenario:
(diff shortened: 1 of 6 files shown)


markbackman:
Can you update the AGENTS.md template vended by init. It needs updated guidance for how to operate given this updated functionality.

markbackman (approved):
Cool idea! Defaulting to False seems like a good idea since eval testing will want to reuse the same bot file, often.

We'll want to make sure that there's a corresponding update in:
https://github.com/pipecat-ai/pipecat/blob/main/src/pipecat/cli/agent_templates/AGENTS.md

Right now, the guidance is that iterations are one at a time. Can you update in this PR?

aconchillo:
> Cool idea! Defaulting to False seems like a good idea since eval testing will want to reuse the same bot file, often.
> 
> We'll want to make sure that there's a corresponding update in: https://github.com/pipecat-ai/pipecat/blob/main/src/pipecat/cli/agent_templates/AGENTS.md
> 
> Right now, the guidance is that iterations are one at a time. Can you update in this PR?

Yes, updating.

PR #4782, opened 2026-06-16
Add Time To First Audio (TTFA) metrics to TTS services

markbackman:
## Summary

TTS services already report **TTFB** (time to first byte), but providers pad the start of a response with silence — so the first byte is not the first audible sound. This adds **TTFA** (time to first audio): the time until the first *audible* sample, computed as TTFB plus the leading-silence duration measured from the signal. Comparing TTFA against TTFB reveals how much perceived latency is silence padding versus service response time.

## How it works

- **`detect_speech_onset`** (`audio/utils.py`) — downmixes to mono, scans short-time RMS energy with a sliding window (mean-removed so a DC offset carries no energy), and reports the start of the first run that stays above a dBFS gate for a minimum voiced duration. Working on energy rather than per-sample amplitude rejects noise-floor blips; the minimum-duration requirement rejects brief transients.
- **`FrameProcessorMetrics`** — `stop_ttfb_metrics` arms the scan when the first byte arrives; `process_ttfa_metrics` buffers audio across chunks (leading silence may span several, and onset confirmation needs a little audio past it) until onset is confirmed, then emits `TTFAMetricsData` as `TTFB + leading-silence duration`.
- **`TTSService`** feeds each `TTSAudioRawFrame` into `process_ttfa_metrics`; the metrics log observer surfaces the new metric.

## Testing

`tests/test_ttfa_metrics.py` covers the onset detector (leading silence, tone-from-start, all-silence, isolated blip rejection, brief loud-burst rejection, stereo) and the metrics flow (multi-chunk buffering, report-once-per-response, re-arming per response, all-silence emits nothing).

```
uv run pytest tests/test_ttfa_metrics.py
```

## Note on calibration

The default `-40 dBFS` gate sits above typical TTS noise-floor padding and below voiced onset, but the threshold is best validated against real provider recordings. It's exposed as a `detect_speech_onset` parameter for tuning.


src/pipecat/audio/utils.py
```diff
@@ -283,3 +283,96 @@ def is_silence(pcm_bytes: bytes) -> bool:
...
+
+def detect_speech_onset(
+    pcm_bytes: bytes,
+    sample_rate: int,
+    num_channels: int = 1,
+    *,
+    frame_ms: float = 10.0,
+    hop_ms: float = 1.0,
+    threshold_db: float = -40.0,
+    min_voiced_ms: float = 25.0,
```
vr000m:
```suggestion
    min_voiced_ms: float = 50.0,
```

vr000m (approved):

src/pipecat/audio/utils.py
vr000m: @markbackman Based on the tests, 35 providers, increasing it to 50ms improves the calculation. Of course more data runs would be better.

markbackman: Updated! Thanks for testing 🙏

PR #4783, opened 2026-06-16
feat(aws-nova-sonic): report LLM token usage metrics

ANGdesarrollo:
AWS Nova Sonic emits a `usageEvent` on its stream, but the service never routed it, so it was the only speech-to-speech service that reported no token usage (its inherited `can_generate_metrics()` returned `False` and the event was silently dropped by the receive loop's `if/elif`).

Override `can_generate_metrics()` and handle `usageEvent`: report the incremental `details.delta` (not the cumulative `details.total`) so usage stays per-event, consistent with the other realtime services (OpenAI/xAI/Inworld emit incrementally per response). Collapse Nova Sonic's speech/text buckets into prompt/completion totals, since `LLMTokenUsage` does not separate modalities (same approach OpenAI Realtime already takes).

## The usageEvent payload

Captured from a live call:

```json
{
  "usageEvent": {
    "details": {
      "delta": {
        "input":  { "speechTokens": 0,  "textTokens": 3 },
        "output": { "speechTokens": 20, "textTokens": 0 }
      },
      "total": {
        "input":  { "speechTokens": 288, "textTokens": 3443 },
        "output": { "speechTokens": 694, "textTokens": 203 }
      }
    },
    "totalTokens": 4628
  }
}
```

`total` is cumulative across the session; `delta` is the per-event increment. Reporting `delta` means summing the frames downstream yields the correct session total, matching how the other realtime services behave.

## Testing

- **Unit tests** (`tests/test_nova_sonic_usage_metrics.py`): cover `can_generate_metrics()`, the collapsed delta mapping, the zero-token guard, and safe handling of a partial payload. `pytest` → 4 passed.
- **Integration**: ran a real Nova Sonic call. Before this change the pipeline emitted 0 usage metrics; after, it emitted 41 `LLMUsageMetricsData` frames with real token counts (model `amazon.nova-2-sonic-v1:0`), incremental per event as expected.
- `ruff check` + `ruff format` pass.

### DIFF
--- a/.github/workflows/coverage.yaml
+++ b/.github/workflows/coverage.yaml
@@ -37,6 +37,7 @@ jobs:
           uv sync --group dev \
             --extra anthropic \
             --extra aws \
+            --extra aws-nova-sonic \
             --extra azure \
             --extra cli \
             --extra daily \
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -41,6 +41,7 @@ jobs:
           uv sync --group dev \
             --extra anthropic \
             --extra aws \
+            --extra aws-nova-sonic \
             --extra azure \
             --extra cli \
             --extra daily \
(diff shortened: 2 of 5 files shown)


markbackman (approved):
LGTM! Thanks for adding this.

PR #4785, opened 2026-06-17
Add raw VAD speaking events and incomplete-turn eval scenarios

aconchillo:
## Summary

Adds raw VAD speaking events to the RTVI observer so eval scenarios can anchor timing on the VAD signal even when a turn-detection strategy gates the turn-level events, and uses them to add two incomplete-turn eval scenarios. Also fixes a noisy disconnect log in the websocket server transport.

## Why

`FilterIncompleteUserTurnStrategies` deliberately defers the turn-level `user-stopped-speaking` until the LLM confirms the turn is complete. That's the feature — but it means an eval can't anchor a follow-up turn on `user_stopped_speaking` after an incomplete utterance (it never fires). The raw VAD stop signal does fire on the pause, so exposing it gives scenarios a reliable anchor.

### Raw VAD speaking events (RTVI)
- `RTVIObserver` emits `vad-user-started-speaking` / `vad-user-stopped-speaking` from `VADUserStartedSpeakingFrame` / `VADUserStoppedSpeakingFrame`, gated by a new `RTVIObserverParams.vad_user_speaking_enabled` (**off by default** — production RTVI traffic is unchanged; clients ignore unknown messages anyway).

### Eval harness integration
- Extended `RTVIConfigureObserverFrame` + the `eval-configure` message with a `vad_user_speaking` flag. The harness enables it **per-connection, only when a scenario references** `vad_user_started_speaking` / `vad_user_stopped_speaking` (in `expect:` or `send_after:`), mirroring how it raises the function-call report level. Bots keep their defaults.
- Harness maps the new RTVI messages to friendly events.

### Eval scenarios
- `filter_incomplete_turns` — user trails off ("Let me think about it…"), finishes after a pause; bot must hold the turn open and answer only the complete thought.
- `filter_incomplete_turns_function_calling` — same, but the completed thought is a weather question; bot must call `get_current_weather` only once the location is given. Both wired into the manifest.

### Unrelated fix
- `SingleClientWebsocketServerTransport` no longer logs a clean client disconnect (`ConnectionClosed`) as an ERROR on the send/receive paths — it's a normal end-of-session race, now logged at debug.

## Test plan

- 184 rtvi/evals tests pass; `ruff check` / `ruff format` clean.
- New tests: VAD config apply (observer), `eval-configure` VAD parse (serializer), VAD-event detection (harness).
- Both scenarios parse against the real schema and resolve to valid manifest runs.

## Notes

- No changelog: the evals framework is unreleased. The VAD RTVI events are additive and off by default. The websocket-disconnect change is a log-level fix.
- The websocket-disconnect fix is independent of the VAD work; happy to split it into its own PR if preferred.

### DIFF
--- a/changelog/4785.added.md
+++ b/changelog/4785.added.md
@@ -1,0 +1,6 @@
+- `RTVIObserver` can now emit raw VAD user speaking events
+  (`vad-user-started-speaking` / `vad-user-stopped-speaking`), driven directly by
+  the VAD signal and independent of turn finalization (unlike
+  `user-started-speaking` / `user-stopped-speaking`, which a turn strategy may
+  gate or defer). Enable with `RTVIObserverParams(vad_user_speaking_enabled=True)`
+  (off by default), or at runtime via `RTVIConfigureObserverFrame`.
--- a/scripts/release-evals/manifest.yaml
+++ b/scripts/release-evals/manifest.yaml
@@ -251,3 +251,7 @@ suite:
   #
   - bot: turn-management/turn-management-filter-incomplete-turns.py
     scenarios: [filter_incomplete_turns]
+  - bot: turn-management/turn-management-filter-incomplete-turns-function-calling.py
+    scenarios: [filter_incomplete_turns_function_calling]
+  - bot: turn-management/turn-management-filter-incomplete-turns-function-calling.py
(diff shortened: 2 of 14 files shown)


markbackman (approved):
Neat idea! Great to be able to test this.

PR #4791, opened 2026-06-17
Serialize Deepgram Flux Configure messages

jamsea:
## Summary

- Serialize Deepgram Flux `Configure` control messages so only one is ever in flight at a time. A burst of `STTUpdateSettingsFrame`s (for example detect-then-lock language hints, or rapid keyterm/EOT tweaks) no longer trips the Flux server error `Configure rejected: [unknown] Too many pending Configure messages.`
- The change lives in the shared `DeepgramFluxSTTBase`, so it covers both the WebSocket and SageMaker transports.

## Problem

`DeepgramFluxSTTService` sent Flux `Configure` messages fire-and-forget, writing each one straight to the socket without waiting for an acknowledgment. Deepgram Flux caps the number of un-acked ("pending") Configure messages, so a quick burst of settings updates would trip a server-side rejection. This was a missing backpressure gap, surfaced by a user on the Flux integration.

## Fix

- `_send_configure` now serializes sends with an `asyncio.Lock`: it creates a Future, writes the message, then waits for that Future before returning. The next send in a burst blocks on the lock until the current one is acked.
- `_handle_message` resolves the Future when a `ConfigureSuccess` or `ConfigureFailure` arrives. Existing failure logging and `push_error` behavior are unchanged.
- The wait is bounded by `_CONFIGURE_ACK_TIMEOUT` (5.0s). On timeout it logs a warning and proceeds, so a missing ack never deadlocks the pipeline.
- The waiter is cleared on reconnect (`_handle_connection_established`) and on teardown (`_disconnect` / `_disconnect_websocket`), so a reconnect starts clean and an in-progress send unblocks promptly. A stray ack with nothing in flight is ignored.

Single-update behavior is unchanged apart from waiting for one ack round-trip.

## Testing

- Adds `tests/test_deepgram_flux_stt.py` (5 tests): waits-for-ack, burst serialization (the second send does not go out until the first is acked), failure signals the waiter and still pushes an error, times-out without an ack, and stray acks are ignored.
- The test helper builds a complete store-mode settings object so the auto-discovered store-completeness check in `tests/test_service_init.py` (which instantiates every `AIService` subclass) passes for it.
- Run: `uv run pytest tests/test_deepgram_flux_stt.py`. `ruff format` and `ruff check` are clean.

## Open question for reviewers

The 5s ack timeout is a conservative default chosen so a healthy ack never times out. Happy to adjust if there is a known Flux ack SLA.


src/pipecat/services/deepgram/flux/base.py
```diff
@@ -154,6 +154,11 @@ class DeepgramFluxSTTBase(STTService):
         "language_hints",
     }
     _MULTILINGUAL_MODEL = "flux-general-multi"
+    # Max time to wait for a ConfigureSuccess/ConfigureFailure ack before giving
+    # up on the current Configure and allowing the next one through. Flux caps the
+    # number of un-acked Configure messages, so we serialize sends; this guard
+    # keeps a missing ack from deadlocking the pipeline.
+    _CONFIGURE_ACK_TIMEOUT = 5.0
```
jamsea, line 161: @filipi87 I'm on the fence about `5.0` here. Thinking of setting this to `1` or `0.5`. Do you know if setting this to 5 would completely pause the voice agent for 5 seconds?

The con is the lower this number, the more likely we'll get timeouts and miss sending configs to Flux, but my vibe is telling me that's probably fine 🤷

filipi87:
Hi @jamsea, I have pushed this [PR](https://github.com/pipecat-ai/pipecat/pull/5101), so we don't need to rely on locks or block the pipeline.

Closing this one in favor of #5101.

PR #4794, opened 2026-06-17
Fix Sarvam TTS: change how user agent is passed

dhruvladia-sarvam:
## Summary

Resolves an issue where user agent is populated twice.

### DIFF
--- a/changelog/4794.fixed.md
+++ b/changelog/4794.fixed.md
@@ -1,1 +1,1 @@
-- Fixed `SarvamTTSService` WebSocket handshakes sending duplicate `User-Agent` headers. Pipecat now sends the Sarvam SDK `User-Agent` via `additional_headers` and disables the default `websockets` client `User-Agent`.
+- Fixed `SarvamTTSService` WebSocket handshakes sending duplicate `User-Agent` headers. Pipecat now passes the Sarvam SDK `User-Agent` via the WebSocket client's `user_agent_header` parameter instead of `additional_headers`.
--- a/changelog/4794.fixed.md
+++ b/changelog/4794.fixed.md
@@ -1,0 +1,1 @@
+- Fixed `SarvamTTSService` WebSocket handshakes sending duplicate `User-Agent` headers. Pipecat now sends the Sarvam SDK `User-Agent` via `additional_headers` and disables the default `websockets` client `User-Agent`.
--- a/src/pipecat/services/sarvam/tts.py
+++ b/src/pipecat/services/sarvam/tts.py
@@ -1097,6 +1097,7 @@ class SarvamTTSService(InterruptibleTTSService):
             self._websocket = await websocket_connect(
                 self._websocket_url,
                 additional_headers=ws_additional_headers,
+                user_agent_header=None,
             )
             logger.debug("Connected to Sarvam TTS Websocket")
             await self._send_config()
@@ -1170,6 +1171,7 @@ class SarvamTTSService(InterruptibleTTSService):
                 if msg.get("type") == "audio":
                     request_id = msg.get("data", {}).get("request_id", "N/A")
                     logger.trace(f"TTS request_id={request_id}, context_id={context_id}")
+
                     # Check for interruption before processing audio
                     await self.stop_ttfb_metrics()
                     audio = base64.b64decode(msg["data"]["audio"])
--- a/src/pipecat/services/sarvam/tts.py
+++ b/src/pipecat/services/sarvam/tts.py
@@ -1091,13 +1091,12 @@ class SarvamTTSService(InterruptibleTTSService):
 
             ws_additional_headers = {
                 "api-subscription-key": self._api_key,
-                **sdk_headers(),
             }
 
             self._websocket = await websocket_connect(
                 self._websocket_url,
                 additional_headers=ws_additional_headers,
-                user_agent_header=None,
+                user_agent_header=sdk_headers()["User-Agent"],
             )
             logger.debug("Connected to Sarvam TTS Websocket")
             await self._send_config()


markbackman:
Thanks for proposing, but I don't think we should merge this approach. The `WebsocketTTSService` requires that the service API supports either:
- tracking context IDs (sending to the server and receiving them back in the audio data) so that Pipecat can track and stop playing out a given context when an interruption happens
- send an interruption message for a given context ID so that audio stops streaming

Without this, any other solution is going to be somewhat of a hack. The `InterruptibleTTSService` was built for services that don't support either approach. Yes, disconnecting the websocket is suboptimal, but I'm nervous about any other approach as it can introduce new problems.

I would suggest extending your service to support one or both of these functionalities. That will make it easier for users of the API to build real-time applications that support interruptions.

dhruvladia-sarvam:
> Thanks for proposing, but I don't think we should merge this approach. The `WebsocketTTSService` requires that the service API supports either:
> 
> * tracking context IDs (sending to the server and receiving them back in the audio data) so that Pipecat can track and stop playing out a given context when an interruption happens
> * send an interruption message for a given context ID so that audio stops streaming
> 
> Without this, any other solution is going to be somewhat of a hack. The `InterruptibleTTSService` was built for services that don't support either approach. Yes, disconnecting the websocket is suboptimal, but I'm nervous about any other approach as it can introduce new problems.
> 
> I would suggest extending your service to support one or both of these functionalities. That will make it easier for users of the API to build real-time applications that support interruptions.

That makes sense. I verified that the current approach relies on inferred stream state because Sarvam audio messages do not echo a Pipecat context_id, and the API does not expose a clear/cancel message for an in-flight synthesis.

markbackman:
The base `InterruptibleTTSService` already handles the disconnect/reconnect on interruption: when the bot is interrupted it closes the socket and reconnects, and the next `run_tts` sends its text on the fresh connection — so post-interrupt text never goes out on a stale socket. As far as I can tell no change is needed there, and the changelog describes that pre-existing behavior rather than anything this PR adds.

The one actual behavioral change I see is `user_agent_header=None`, which drops the duplicate `User-Agent` header we currently send on the handshake. Was a connection/handshake problem the thing you were originally trying to fix? If so, let's center the PR on that. Otherwise I think this is effectively a no-op and we can close it out.

dhruvladia-sarvam:
> The base `InterruptibleTTSService` already handles the disconnect/reconnect on interruption: when the bot is interrupted it closes the socket and reconnects, and the next `run_tts` sends its text on the fresh connection — so post-interrupt text never goes out on a stale socket. As far as I can tell no change is needed there, and the changelog describes that pre-existing behavior rather than anything this PR adds.
> 
> The one actual behavioral change I see is `user_agent_header=None`, which drops the duplicate `User-Agent` header we currently send on the handshake. Was a connection/handshake problem the thing you were originally trying to fix? If so, let's center the PR on that. Otherwise I think this is effectively a no-op and we can close it out.

I’ve reframed the PR around the actual delta: Sarvam TTS was sending duplicate `User-Agent` headers on the WebSocket handshake. The PR now keeps the existing interruption model unchanged and only disables the default `websockets` User-Agent while preserving the Pipecat/Sarvam SDK User-Agent in `additional_headers`. This keeps reconnect behaviour in the base class and makes the Sarvam WebSocket handshake cleaner for both initial connect and reconnect after interruption.

markbackman:
I'm not sure I understand your change. For a user-agent header, the only change should be:
```
                self._websocket_url,
                additional_headers=ws_additional_headers,
                user_agent_header=None,
            )
```

Why do you need to include `_logged_tts_request_ids`?

Also, no tests are required for that change.

If I'm missing something about `_logged_tts_request_ids`, and how that relates to the `user_agent_header` change, please let me know.

src/pipecat/services/sarvam/tts.py
```diff
@@ -1097,6 +1097,7 @@ async def _connect_websocket(self):
             self._websocket = await websocket_connect(
                 self._websocket_url,
                 additional_headers=ws_additional_headers,
+                user_agent_header=None,
```
markbackman: This seems like a workaround that we should avoid.

The duplicate comes from injecting User-Agent into `additional_headers` while the websockets client also sends its own default, which is why you're then forced to disable the default with `user_agent_header=None`.

It would be more "traditional" to do something like this:
```
ws_additional_headers = {"api-subscription-key": self._api_key}

    self._websocket_url,
    additional_headers=ws_additional_headers,
    user_agent_header=sdk_headers()["User-Agent"],
)
```

That is, remove the user agent from `additional_headers` and let `user_agent_header` carry the `User-Agent` payload.

src/pipecat/services/sarvam/tts.py
```diff
@@ -1170,6 +1171,7 @@ async def _receive_messages(self):
                 if msg.get("type") == "audio":
                     request_id = msg.get("data", {}).get("request_id", "N/A")
                     logger.trace(f"TTS request_id={request_id}, context_id={context_id}")
+
```
markbackman, line 1173: Super nit: remove stray newline

changelog/4794.fixed.md
```diff
@@ -0,0 +1 @@
+- Fixed `SarvamTTSService` WebSocket handshakes sending duplicate `User-Agent` headers. Pipecat now sends the Sarvam SDK `User-Agent` via `additional_headers` and disables the default `websockets` client `User-Agent`.
```
markbackman: Recommend updating to reflect the change suggested.

markbackman (commented):
Proposing an alternative to avoid this workaround, which is fragile to breaking in the future.

markbackman (approved):

PR #4796, opened 2026-06-17
fix: AWSBedrockLLMAdapter raises UnboundLocalError for image-before-text content (#4724)

### LINKED ISSUE #4724: AWSBedrockLLMAdapter raises UnboundLocalError for image-before-text content

singlaamitesh:
## Summary

Fixes #4724.

`AWSBedrockLLMAdapter._from_standard_message` raised `UnboundLocalError: cannot access local variable 'image_item'` when a user message's content list placed the image **before** the text (a valid OpenAI shape, e.g. a `UserImageRawFrame` followed by a prompt). The error was caught silently by `_from_universal_context_messages`, which then returned empty `messages`, dropping the image and prompt from the Bedrock request.

**Root cause:** The `new_content.insert(first_txt_idx, image_item)` call at line 306 of `bedrock_adapter.py` was indented one level too shallow — it sat **outside** the `if img_idx > first_txt_idx:` guard that assigns `image_item`. When the image was already first (no reorder needed), the guard was skipped and `image_item` was never assigned, causing the `UnboundLocalError` on the subsequent `insert`.

**Fix:** Indent `new_content.insert(...)` into the `if` block so it only runs when a reorder is actually needed.

## Changes

- `src/pipecat/adapters/services/bedrock_adapter.py` — one-line indentation fix
- `tests/test_get_llm_invocation_params.py` — two new regression tests:
  - `test_image_before_text_does_not_raise_unbound_local_error` — the crash case
  - `test_image_after_text_reordered_to_first` — the existing reorder behaviour
- `changelog/4724.fixed.md` — changelog fragment

## Test plan

- [x] `uv run pytest tests/test_get_llm_invocation_params.py::TestAWSBedrockGetLLMInvocationParams -v` — all 14 tests pass (including 2 new)
- [x] Repro from issue confirmed fixed (no AWS credentials required)

### DIFF
--- a/changelog/4724.fixed.md
+++ b/changelog/4724.fixed.md
@@ -1,0 +1,1 @@
+- Fixed `AWSBedrockLLMAdapter` raising `UnboundLocalError` when a user message's content list places the image before the text (e.g. a `UserImageRawFrame` followed by a prompt). The reorder guard's `insert` call was incorrectly placed outside its `if` block, so image-first input would crash silently and drop the message.
--- a/changelog/4724.fixed.md
+++ b/changelog/4724.fixed.md
@@ -1,1 +1,0 @@
-- Fixed `AWSBedrockLLMAdapter` raising `UnboundLocalError` when a user message's content list places the image before the text (e.g. a `UserImageRawFrame` followed by a prompt). The reorder guard's `insert` call was incorrectly placed outside its `if` block, so image-first input would crash silently and drop the message.
--- a/changelog/4796.fixed.md
+++ b/changelog/4796.fixed.md
@@ -1,0 +1,1 @@
+- Fixed `AWSBedrockLLMAdapter` raising `UnboundLocalError` when a user message's content list places the image before the text (e.g. a `UserImageRawFrame` followed by a prompt).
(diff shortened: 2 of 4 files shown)


markbackman (approved):
LGTM! Thanks for fixing this 🙇

PR #4797, opened 2026-06-17
fix: GladiaSTTService silently ignores runtime settings updates (#4719)

### LINKED ISSUE #4719: GladiaSTTService: runtime settings updates are silently ignored
## Description

`GladiaSTTService._update_settings()` stores updated settings but never applies them to the active session. The service must create a new Gladia HTTP session (via `POST /v2/live`) to apply any settings change, because all configuration — model, language, endpointing, VAD, pre/post-processing — is submitted at session-init time, not over the WebSocket.

As a result, calling `STTUpdateSettingsFrame` at runtime (e.g. to switch language or model mid-session) has no effect on the live transcription.

## Current behavior

`_update_settings` applies the delta to `self._settings` and logs a warning via `_warn_unhandled_updated_settings`, but the active WebSocket session continues using the original settings.

There is even a TODO comment in the code acknowledging this:

```python
# TODO: someday we could reconnect here to apply updated settings.
# Code might look something like the below:
# self._session_url = None
# self._session_id = None
# await self._disconnect()
# await self._connect()
```

## Expected behavior

When settings change, the service should:
1. Clear the current session URL/ID so `_connect()` initializes a new Gladia session with the updated settings
2. Reconnect using the existing `_request_reconnect()` mechanism (which safely defers reconnection until after the current user turn and buffers audio during the transition)

## Proposed fix

```python
async def _update_settings(self, delta: Settings) -> dict[str, Any]:
    changed = await super()._update_settings(delta)

    if not changed:
        return changed

    self._session_url = None
    self._session_id = None
    await self._request_reconnect()

    return changed
```

This follows the same pattern already used by `DeepgramSTTService._update_settings`, which calls `_request_reconnect()` after applying the delta.

singlaamitesh:
## Summary

Fixes #4719.

`GladiaSTTService._update_settings()` was storing the settings delta in `self._settings` but never reconnecting, so any runtime settings change (language, model, VAD config, endpointing) sent via `STTUpdateSettingsFrame` had no effect on the live transcription session.

Gladia bakes all configuration into the session at `POST /v2/live` time and provides no mid-session update API, so a settings change requires creating a fresh session. The fix:

1. Clears `_session_url` and `_session_id` so `_connect()` will `POST /v2/live` with the updated settings on the next connection
2. Calls `_request_reconnect()` which safely defers the reconnect until after the current user turn if the user is speaking — matching the pattern already used by `DeepgramSTTService`

The existing TODO comment in the code described exactly this approach; this PR implements it.

## Changes

- `src/pipecat/services/gladia/stt.py` — replace TODO with the reconnect logic; update docstring
- `tests/test_gladia_stt.py` — two new unit tests:
  - `test_update_settings_triggers_reconnect` — verifies session is cleared and reconnect is requested on settings change
  - `test_update_settings_no_reconnect_when_nothing_changed` — verifies no reconnect when delta produces no change
- `changelog/4719.fixed.md` — changelog fragment

## Test plan

- [x] `uv run pytest tests/test_gladia_stt.py -v` — both tests pass

### DIFF
--- a/changelog/4719.fixed.md
+++ b/changelog/4719.fixed.md
@@ -1,0 +1,1 @@
+- Fixed `GladiaSTTService` silently ignoring runtime settings updates. Settings are now applied by clearing the current session and triggering a safe reconnect via `_request_reconnect()`, which defers until after the current user turn if the user is speaking.
--- a/src/pipecat/services/gladia/stt.py
+++ b/src/pipecat/services/gladia/stt.py
@@ -419,9 +419,13 @@ class GladiaSTTService(WebsocketSTTService):
         await self._connect()
 
     async def _update_settings(self, delta: Settings) -> dict[str, Any]:
-        """Apply settings delta.
+        """Apply settings delta and reconnect with new settings.
 
-        Settings are stored but not applied to the active session.
+        Clears the current Gladia session so that ``_connect()`` initialises a
+        fresh session (via ``POST /v2/live``) with the updated configuration,
+        then triggers a reconnect via the safe ``_request_reconnect()`` path,
+        which defers the reconnect until after the current user turn if the
+        user is speaking.
 
         Args:
             delta: A settings delta.
@@ -434,14 +438,9 @@ class GladiaSTTService(WebsocketSTTService):
         if not changed:
             return changed
 
-        # TODO: someday we could reconnect here to apply updated settings.
-        # Code might look something like the below:
-        # self._session_url = None
-        # self._session_id = None
-        # await self._disconnect()
-        # await self._connect()
-
-        self._warn_unhandled_updated_settings(changed)
+        self._session_url = None
+        self._session_id = None
+        await self._request_reconnect()
 
         return changed
 
--- a/tests/test_gladia_stt.py
+++ b/tests/test_gladia_stt.py
@@ -28,18 +28,10 @@ async def test_update_settings_triggers_reconnect():
     service._settings = GladiaSTTService.Settings(model="solaria-1")
     service._session_url = "wss://fake-session.gladia.io/live/abc123"
     service._session_id = "abc123"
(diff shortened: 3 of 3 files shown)


Rehansanjay:
I arrived at this independently in #5454 before spotting yours, and have closed mine in favour of this one — it was two months earlier. Two things came out of doing it separately that might be worth folding in.

**1. `language` still cannot be applied, so the warning probably shouldn't go entirely**

This PR drops `_warn_unhandled_updated_settings(changed)`. The reconnect does make every field in `_prepare_settings` take effect — but `language` is not one of them. Gladia selects languages through `language_config`, and `_prepare_settings` reads only `s.language_config`; the base `language` field is never put in the session payload. The constructor also sets `language=None` in its defaults, which is the store-mode way of declaring a field unsupported.

So after this change `STTUpdateSettingsFrame(language=Language.ES)` reconnects, looks accepted, and changes nothing about the transcription. Narrowing rather than removing keeps that honest:

```python
self._warn_unhandled_updated_settings(changed.keys() & {"language"})
```

**2. Two test cases that catch things a reconnect assertion doesn't**

The test here asserts `_request_reconnect()` gets called. Two others cover different regressions:

- **The new session actually carries the change.** Stub `_setup_gladia` to record the settings dict each `POST /v2/live` receives, then assert the second call contains the updated value. A reconnect that fires but rebuilds from stale settings would still pass a call-assertion.
- **A mid-utterance update is deferred.** Drive `_handle_vad_user_started_speaking(...)` first, apply the update, assert no new session was opened, then call `_maybe_reconnect_on_user_stopped_speaking()` and assert it lands. That pins the audio-safety property `_request_reconnect` exists for — reconnecting mid-utterance would drop the caller's speech, and nothing else in the suite would notice if that regressed.

Happy to push either as a suggestion here if it saves you time.

PR #4802, opened 2026-06-18
Update community integrations to match latest docs format

markbackman:

We're now integrating community integrations into the normal reference docs and supported services tables. This is a guide update for developers creating new integrations:
https://github.com/pipecat-ai/docs/pull/925


mattieruth (approved):
i still wonder if we should add a skill or these steps to the docs repo itself to make having an llm generate the docs changes easier.

Issue #4784, opened 2026-06-17
DeepgramSageMakerSTTService crashes on instantiation: unexpected keyword argument 'vad_events' (1.3.0)


aaronsu11:
## Description

`DeepgramSageMakerSTTService` cannot be instantiated at all on Pipecat 1.3.0 — constructing it with default settings raises a `TypeError`. The service's own default settings builder passes `vad_events=False`, but that field no longer exists on the `DeepgramSTTService.Settings` dataclass it inherits from.

This also means the official example [`examples/voice/voice-deepgram-sagemaker.py`](https://github.com/pipecat-ai/pipecat/blob/main/examples/voice/voice-deepgram-sagemaker.py) crashes on startup against 1.3.0 (it constructs the service with no `settings=`).

This is **distinct from #4737 / #4739** (those are about runtime settings *updates* being ignored on an already-running service); here the service cannot be constructed in the first place.

## Steps to reproduce

No AWS resources or credentials required — the crash happens at construction time:

```python
from pipecat.services.deepgram.sagemaker.stt import DeepgramSageMakerSTTService

DeepgramSageMakerSTTService(endpoint_name="x", region="us-east-1")
```

## Actual behavior

```
TypeError: DeepgramSageMakerSTTSettings.__init__() got an unexpected keyword argument 'vad_events'
```

## Root cause

In `src/pipecat/services/deepgram/sagemaker/stt.py`, the default settings are built with `vad_events=False` (~line 146):

```python
self._settings = self.Settings(
    ...
    utterance_end_ms=None,
    vad_events=False,   # <-- not a field on DeepgramSTTService.Settings
)
```

But `DeepgramSageMakerSTTSettings` only inherits the base settings:

```python
@dataclass
class DeepgramSageMakerSTTSettings(DeepgramSTTService.Settings):
    """Inherits all fields from DeepgramSTTService.Settings."""
    pass
```

and `DeepgramSTTService.Settings` no longer defines `vad_events` (no occurrence of `vad_events` exists in `services/deepgram/stt.py`). It appears #3386 ("Deprecate support for vad_events in DeepgramSTTService") removed the field from the base settings, but the SageMaker subclass's default-settings construction was not updated to match, so any default construction now fails.

## Suggested fix

[...]

PR #4803, opened 2026-06-18
fix(deepgram): defer DeepgramSageMakerSTTService connect to background task

markbackman:
## Summary

`DeepgramSageMakerSTTService.start()` awaited `_connect()` inline. Because `AIService.process_frame` runs `start()` *before* forwarding `StartFrame` downstream (see `services/stt_service.py` / `services/ai_service.py`), a slow or failing SageMaker BiDi handshake blocked the pipeline's `StartFrame` barrier: no greeting or first bot turn could happen until the connection resolved, and a failing connect looked like a hang.

This mirrors the deferred-connect pattern already used by `DeepgramSTTService`, whose `_connect()` only launches a background task. The input transport similarly pushes `StartFrame` *before* connecting, on purpose — services are the outlier when they block.

## Changes

- `_connect()` now launches a background task instead of awaiting the BiDi handshake inline; the handshake logic moves to `_connect_handler()`.
- `_disconnect()` cancels the connection task (covers teardown both mid-handshake and after it completes).
- Added a `_connection_task` field.

## Behavior

- `start()` returns immediately → `StartFrame` propagates → the first bot turn (e.g. a greeting) proceeds while STT connects.
- A failed connect surfaces via `on_connection_error` and a pushed `ErrorFrame` instead of stalling pipeline startup.
- `run_stt()` already guards on `self._client and self._client.is_active`, so audio arriving during the connect window is safely dropped rather than erroring.

## Notes

This is the minimal version of the fix. It does not add a `_connection_ready` event / connect-time audio buffering (which `DeepgramSTTService` uses for its reconnect path) — that can be a follow-up if dropping the first words during the connect window becomes a concern.

Reported in https://github.com/pipecat-ai/pipecat/issues/4784#issuecomment-4733525853


changelog/4803.fixed.md
filipi87, line 1: @markbackman, if I am not mistaken, in all other services, we only push the `StartFrame` to the next step in the pipeline once that processor has actually started.

So, if we receive a frame after that, it means the processor knows how to proceed.

If we change the behavior here, wouldn't that potentially break this service if we start receiving new frames, including audio frames, while this processor is not truly ready?

filipi87: Ok, now I see that we basically have the same pattern inside `DeepgramSTTService`, and we basically ignore the `run_stt` in case the connection is not ready yet. 

Considering this, I think it is ok to replicate this behavior here. 👍

filipi87 (approved):
LGTM!

changelog/4803.fixed.md
markbackman: Yeah, that's the idea. The issue reports a possibility of SageMaker being slow to start, so running connect in a task can avoid blocking.

PR #4810, opened 2026-06-18
Add mode (latency/accuracy preset) param to AssemblyAI STT

dlange-aai:
## Description

Adds the `mode` streaming connection parameter to `AssemblyAISTTService`, exposing AssemblyAI's U3 Pro latency/accuracy preset.

`mode` is one of:
- `min_latency`
- `balanced` — AssemblyAI's server-side default
- `max_accuracy`

It trades transcription accuracy against turn-finalization latency. Like the other U3 Pro-only streaming params (`voice_focus`, `interruption_delay`, `previous_context_n_turns`, …), it is sent as a WebSocket query parameter only for U3 Pro family models (`u3-rt-pro*`, `universal-3-5-pro*`). The server rejects it for other models, so the plugin omits it and logs a warning when it is set on a non-U3-Pro model.

## Usage

```python
from pipecat.services.assemblyai.stt import AssemblyAISTTService

stt = AssemblyAISTTService(
    api_key=...,
    settings=AssemblyAISTTService.Settings(mode="max_accuracy"),
)
```

## Tests

Added unit tests in `tests/test_assemblyai_stt.py`: omitted by default, sent for `u3-rt-pro` and `universal-3-5-pro`, omitted for `universal-streaming-*`, and all three preset values accepted. The full test file passes (58 tests).

### DIFF
--- a/changelog/4810.added.md
+++ b/changelog/4810.added.md
@@ -1,0 +1,1 @@
+- Added a `mode` streaming parameter to `AssemblyAISTTService`, exposing AssemblyAI's U3 Pro latency/accuracy preset (`min_latency`, `balanced`, or `max_accuracy`). It trades transcription accuracy against turn-finalization latency and is only applicable to U3 Pro models, where the server defaults to `balanced`.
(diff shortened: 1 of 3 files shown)


markbackman (approved):
LGTM! Thanks for the update 🙇

PR #4811, opened 2026-06-18
Log eval failure reasons to the eval log file

aconchillo:
## Summary

Eval expectation failures (a missing function call, a response timeout, an unsatisfied `eval:`, or a `send_after` that never fired) were only sent to the `on_progress` callback and surfaced in the terminal. They were never written to the per-scenario `.eval.log` file, so once a run finished there was no record of *why* an expectation failed, which made after-the-fact debugging impossible.

This PR makes the eval log self-explanatory: it shows what each expectation is waiting for, then either the received events or a `FAIL:` line with the reason.

- Write each failure reason to the debug log so it lands in the `.eval.log` file (send_after timeout, expectation timeout, and returned assertion failures).
- Emit a `match: waiting for ...` line up front for every expectation type (single event, aggregating text, and function calls), so the log clearly states what is expected. The expected function call line includes its args, and each expected call is printed on its own line.
- Reword the matcher's pre-wait debug line from `match: aggregating ...` to `match: waiting for ...` (it is emitted before any event is received) and drop the now-redundant `match: timeout, no event arrived` line, since the caller already logs the `FAIL:` reason.

## Example log output

```
match: waiting for 'function_call' (get_current_weather(location='San Francisco'))
match: waiting for 'function_call' (this_function_does_not_exist)
FAIL: function_call: function call 'this_function_does_not_exist' not seen (matched: get_current_weather)
```

## Test plan

Run a scenario with an expectation that cannot be satisfied (e.g. a function call the bot will not make) and confirm the `match: waiting for ...` lines and the `FAIL:` reason now appear in the generated `<scenario>.eval.log` file.

### DIFF
--- a/changelog/4811.fixed.md
+++ b/changelog/4811.fixed.md
@@ -1,0 +1,1 @@
+- Fixed eval failure reasons (a missing function call, a response timeout, an unsatisfied `eval:`, or a `send_after` that never fired) not being written to the per-scenario `.eval.log` file. They were only printed to the terminal during the run, so there was no record to debug failures after the fact.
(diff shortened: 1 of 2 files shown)


markbackman (approved):
LGTM

PR #4814, opened 2026-06-18
Include TTFB and leading-silence breakdown in TTFAMetricsData

markbackman:
## Summary

Fast follow-up to #4782 (TTFA metrics). TTFA is time-to-first-byte plus the leading silence a TTS service pads onto the start of its response — and that padding is the number developers actually want. Reporting only a single combined value forced any consumer wanting the padding to catch a separate `TTFBMetricsData`, stash it keyed by processor, then subtract.

This carries the breakdown on `TTFAMetricsData` directly:

```python
class TTFAMetricsData(MetricsData):
    ttfa: float             # the measurement (ttfb + leading_silence)
    ttfb: float             # the TTFB it builds on (snapshot of TTFBMetricsData)
    leading_silence: float  # padding = ttfa - ttfb
```

`process_ttfa_metrics` populates all three from values it already has at emit time — no extra measurement.

### Design notes

- **`value` renamed to `ttfa`.** With `ttfb`/`leading_silence` alongside it, a field named `value` read oddly. The three fields are now uniformly named, following the multi-field `TurnMetricsData` pattern (which also has no `value`). The type is new in #4782 (unreleased), so this rename has no external impact.
- **`ttfb` is a snapshot, not a second measurement.** The standalone `TTFBMetricsData` still emits independently and *earlier* (when the first byte arrives, before TTFA can confirm onset), so it remains the timely, authoritative TTFB. The `ttfb` field here mirrors it for one-object convenience — aggregators that sum TTFB should key on `TTFBMetricsData`, not this field. This is called out in the docstring.
- **Field, not a duplicate frame.** The breakdown is carried as fields on `TTFAMetricsData` rather than a second `TTFBMetricsData` in the same `MetricsFrame`, so TTFB isn't double-counted.

## Changes

- `TTFAMetricsData`: `ttfa`, `ttfb`, `leading_silence` (renamed `value` → `ttfa`).
- `FrameProcessorMetrics.process_ttfa_metrics`: populate the breakdown.
- `MetricsLogObserver`: log the padding, e.g. `TTFA: 0.203s (0.121s leading silence)`.

## Testing

`tests/test_ttfa_metrics.py` asserts the breakdown: `ttfb ≈ 0.2`, `leading_silence ≈ 0.4`, and `ttfa == ttfb + leading_silence`.

```
uv run pytest tests/test_ttfa_metrics.py
```


aconchillo (approved):

aconchillo:
LGTM! Thanks!

PR #4815, opened 2026-06-18
Inject the task manager into workers and the runner

aconchillo:
## Summary

Make the `TaskManager` an injectable, shareable dependency so a caller can control the event loop and the `contextvars.Context` that worker and processor tasks run in.

- `TaskManager` now takes its `loop` and an optional `context` in the constructor and runs all tasks within that context. `create_task` also accepts a per-call `context` override.
- `WorkerRunner`, `BaseWorker`, and `WorkerParams` accept a `task_manager`, so one task manager (and one context) can be shared across the runner and every worker.
- Dangling-task reporting moved to the owner of the task manager: the `WorkerRunner` reports for the shared task manager after teardown, and a worker only reports when it owns its own task manager. `check_dangling_tasks` is available on `WorkerRunner`, `BaseWorker`, and `PipelineWorker`.

## Breaking changes

- Removed `WorkerParams.loop`. Pass `WorkerParams(task_manager=...)` instead.

## Deprecations

- `TaskManagerParams` and `TaskManager.setup()` — pass `loop`/`context` to the `TaskManager` constructor instead.
- `WorkerRunner(loop=...)` — pass `task_manager=...` instead.


src/pipecat/workers/base_worker.py
```diff
@@ -68,10 +68,10 @@ class WorkerParams:
     """Configuration parameters for worker execution.
 
     Parameters:
-        loop: The asyncio event loop to use for worker execution.
+        task_manager: Task manager for handling asyncio tasks.
     """
 
-    loop: asyncio.AbstractEventLoop
```
aconchillo, line 74: I know this is a breaking change but, without it,  it would force us to make `task_manager` optional and complicate the code. `WorkerParams` is mainly only used by the `WorkerRunner`.

aconchillo:
Example on how to use it:

```
import asyncio
import contextvars

from pipecat.frames.frames import EndFrame, Frame, TextFrame
from pipecat.pipeline.worker import PipelineWorker
from pipecat.utils.asyncio.task_manager import TaskManager
from pipecat.workers.runner import WorkerRunner

session_id = contextvars.ContextVar("session_id", default=None)


class Setter(FrameProcessor):
        if isinstance(frame, TextFrame):
            session_id.set("set-by-Setter")
            print("  Setter : wrote session_id =", session_id.get())


class Reader(FrameProcessor):
        if isinstance(frame, TextFrame):
            print("  Reader : reads session_id =", session_id.get())


async def main():
    # Copy the current context and create a task manager. The task manager will
    # create asyncio tasks with the provided context.
    ctx = contextvars.copy_context()
    task_manager = TaskManager(context=ctx)

    worker = PipelineWorker(Pipeline([Setter(), Reader()]))
    await worker.queue_frames([TextFrame("hi"), EndFrame()])

    # Pass a task manager to the runner. The runner and all its workers will use
    # the given task manager. It's also possible to pass a different task manager to
    # each pipeline worker if needed.
    runner = WorkerRunner(task_manager=task_manager)
    await runner.add_workers(worker)
    await runner.run()

    print("parent : reads session_id =", ctx[session_id])


asyncio.run(main())
```

src/pipecat/workers/base_worker.py
markbackman: I think this is OK. We don't show how it's used anywhere, so it should be unused. This seems safe enough.

markbackman (approved):
LGTM!

Issue #2050, opened 2025-06-24
Make silero onnxruntime session shared


erkamkavak:
### Problem Statement

Currently silero vad tries to load onnx file and create a new onnxruntime InferenceSession each time it is created. When there are multiple clients and pipeline tasks, a new onnxruntime session is created for each of these tasks. 

https://github.com/pipecat-ai/pipecat/blob/aee3011d615942e84d2f12a24abed39e8107bf8d/src/pipecat/audio/vad/silero.py#L113-L129

### Proposed Solution

I think all silero vad analyzers could use a single onnxruntime session across different tasks. As far as I can see, there is no problem using a single onnxruntime session in silero vad(but I guess each silero vad analyzer should be specific to pipeline since it has a state inside) 

### Would you be willing to help implement this feature?

- [x] Yes, I'd like to contribute
- [ ] No, I'm just suggesting

programmerraja:
@erkamkavak Yep I need same for other service like TTS,LLM,STT because when we using selfhost models like WhisperSTTService,KokoroTTSService where we load the model in memory and when we have multiple clients we the model will be loaded for each user.

rahultyl:
@markbackman any thoughts here? Is it possible to do this for silero VAD?

sbenodiz:
+1 That would be great to save memory in my service

VpkPrasanna:
@markbackman  any thoughts on this issue ?

PR #4818, opened 2026-06-19
Use shared onnxruntime.InferenceSession for SileroVadAnalyzer and LocalSmartTurnAnalyzerV3

skheyfets-asapp:
## Summary

`SileroVADAnalyzer` and `LocalSmartTurnAnalyzerV3` previously built a **new ONNX
`InferenceSession` per instance**, loading the model weights (~30 MB RSS each)
afresh for every analyzer. In a server handling many simultaneous calls, every
call constructs its own analyzer and therefore its own copy of identical,
immutable weights — duplicated memory that dominates at high concurrency and can
OOM a memory-limited deployment.

This PR makes both analyzers reuse a **single process-wide `InferenceSession` per
configuration**. The model weights are immutable and ONNX Runtime's `Run()` is
thread-safe for concurrent calls on a CPU session (and releases the GIL), so one
session can safely back every analyzer instance. The change is transparent and
bit-identical to the previous per-call behavior — no new constructor params, no
env flags.

## What is and isn't shared

- **Shared (immutable):** the ONNX `InferenceSession`, built once per config and
  cached process-wide.
- **Per-instance (mutable/stateful):** everything else — the smart-turn
  `_audio_buffer` and inference `ThreadPoolExecutor`, and the Silero recurrent
  `_state`/`_context` (reset per instance). Sharing changes nothing about how
  per-turn state is tracked.

## Implementation

- A module-level cache guarded by a `threading.Lock` (double-checked locking) in
  each of `silero.py` and `local_smart_turn_v3.py`.
- Cache key is computed **after** resolving the model path, so `path=None`
  (bundled) and an explicit identical path collapse to one session:
  - smart-turn: `(resolved_model_path, cpu_count)`
  - Silero: `(resolved_path, force_onnx_cpu)`
- Differing configurations get distinct sessions; identical ones share.
- The model-load log fires only on a real build, so constructing N analyzers
  loads the weights once.

I deliberately used a plain cache rather than mirroring the ref-counted
`AICModelManager`: these sessions are built synchronously in `__init__` and are
already never freed (`cleanup()` is a no-op today), so ref-counting would add
surface for an eviction that never fires. A transparent forever-cache preserves
the existing lifecycle with minimal surface.

## Benchmarks & rationale

`scripts/benchmarks/onnx_shared_session/` contains a self-contained suite (only
needs `numpy`, `onnxruntime`, and the model file) plus a captured run with
analysis:

- **correctness** — 16 threads × 200 reps on one shared session: every output
  bit-for-bit matches the single-threaded reference (thread-safety proof).
- **latency/throughput** — shared vs per-call across a concurrency sweep up to 5×
  the core count: shared tracks per-call within noise (no contention penalty),
  and is *faster* at the top end.
- **memory** — per-call RSS scales with the number of sessions (~31 MB each);
  shared has a fixed ~44 MB cost plus arena that scales only with inferences
  *in flight at once*. Illustratively, 100 concurrent calls with ~8 overlapping
  inferences: ~3.1 GB per-call vs ~190 MB shared (>10× reduction).

The benchmark exercises the smart-turn session directly; Silero shares its
session through the same mechanism and is validated by the unit tests (it rests
on the same stateless-session argument). This was validated downstream in
production before upstreaming.

## Tests

`tests/test_onnx_session_sharing.py` covers:
- singleton identity per config; distinct configs/models get distinct sessions
- per-instance state isolation (shared session, isolated buffers/executors/VAD state)
- numerical equivalence — a cached session and an independently rebuilt session
  produce bit-identical output (sharing introduces no drift)
- concurrency consistency across threads

## Verification

- New + existing analyzer/VAD/turn tests pass; no regressions in the full suite.
- `ruff check` and `ruff format --check` clean.

Addresses: https://github.com/pipecat-ai/pipecat/issues/2050


skheyfets-asapp:
@markbackman can you have a look?

skheyfets-asapp:
@filipi87 would you mind having a look at this one as well?

PR #4820, opened 2026-06-19
fix(silero): use NumPy API in VAD input handling

### LINKED ISSUE #4808: bug: SileroOnnxModel._validate_input calls x.dim() (PyTorch API) on a NumPy array — raises AttributeError instead of ValueError
### pipecat version

latest (main branch, as of 2026-06-18)

### Python version

3.10+

### Operating System

Any (platform-independent)

### Issue description

Two bugs exist in `src/pipecat/audio/vad/silero.py` affecting `SileroOnnxModel` and `SileroVADAnalyzer`.

### Bug 1 — `_validate_input` calls `x.dim()` (PyTorch API) on a NumPy array

**File**: `src/pipecat/audio/vad/silero.py` — `SileroOnnxModel._validate_input`

```python
if np.ndim(x) > 2:
    raise ValueError(f"Too many dimensions for input audio chunk {x.dim()}")
```

`np.ndim(x) > 2` correctly uses NumPy to check dimensions. But the f-string error message calls `x.dim()`, which is a **PyTorch Tensor method** — it does not exist on NumPy arrays. The NumPy equivalent is the `.ndim` property (no parentheses).

When a 3D array is passed as input the code raises:
```
AttributeError: 'numpy.ndarray' object has no attribute 'dim'
```
instead of the intended `ValueError`. The `AttributeError` completely hides the validation message and misleads anyone debugging the issue.

**Fix**: replace `x.dim()` with `x.ndim` inside the f-string.

---

### Bug 2 — `voice_confidence` calls `np.frombuffer` twice; second call is on an already-converted NumPy array

**File**: `src/pipecat/audio/vad/silero.py` — `SileroVADAnalyzer.voice_confidence`

```python
audio_int16 = np.frombuffer(buffer, np.int16)
# Divide by 32768 because we have signed 16-bit data.
audio_float32 = np.frombuffer(audio_int16, dtype=np.int16).astype(np.float32) / 32768.0
```

`audio_int16` is already a NumPy `int16` ndarray. Passing it back into `np.frombuffer` treats the array as a raw bytes buffer, producing a read-only view with unexpected semantics. While `.astype(np.float32)` copies the data and avoids a `ValueError` at runtime, the intermediate `np.frombuffer` call is both incorrect and redundant.

**Fix**:
```python
audio_int16 = np.frombuffer(buffer, np.int16)
audio_float32 = audio_int16.astype(np.float32) / 32768.0
```

### Reproduce Bug 1 (`x.dim()` AttributeError)

```python
import numpy as np
import importlib_resources as impresources
from pipecat.audio.vad.silero import SileroOnnxModel

model_path = str(
    impresources.files("pipecat.audio.vad.data").joinpath("silero_vad.onnx")
)
model = SileroOnnxModel(model_path)

# Pass a 3-D array — should raise ValueError, actually raises AttributeError
bad_input = np.zeros((1, 1, 512), dtype=np.float32)
model(bad_input, 16000)
```

### Reproduce Bug 2 (redundant `np.frombuffer` on ndarray)

```python
import numpy as np

analyzer = SileroVADAnalyzer(sample_rate=16000)

# 512 samples of silence as raw bytes
silence = np.zeros(512, dtype=np.int16).tobytes()

# Internally the second np.frombuffer(audio_int16, ...) is called on an ndarray
confidence = analyzer.voice_confidence(silence)
print(confidence)
```

### Expected behavior

**Bug 1**: Passing a 3-D array should raise a descriptive `ValueError`:
```
ValueError: Too many dimensions for input audio chunk 3
```

**Bug 2**: `voice_confidence` should convert the raw bytes buffer to an int16 ndarray exactly once, then call `.astype(np.float32)` directly on it — no second `np.frombuffer` call.

### Actual behavior

**Bug 1**: Python raises an `AttributeError` that completely masks the real validation failure:
```
AttributeError: 'numpy.ndarray' object has no attribute 'dim'
```

**Bug 2**: `np.frombuffer` is called a second time on `audio_int16` (already an ndarray), producing a redundant read-only view before the `.astype()` call. No crash occurs at runtime because `.astype()` allocates a new array, but the code is semantically wrong and may cause confusion or subtle issues if the implementation changes.

### Logs

```shell

```

markbackman:
**These changes are for correctness; there is no bug that impacts normal usage.**

- Issue 1: The issue would only arise if a >2-d input occurred, but not possible. So it's a fix to unreachable code.
- Issue 2: Is just a simplification. 

## Summary

Two NumPy-correctness fixes in `src/pipecat/audio/vad/silero.py`.

- **`_validate_input` raised the wrong exception on >2-D input.** The "too many dimensions" message interpolated `x.dim()` — a PyTorch-tensor method left over from the upstream silero-vad port. NumPy arrays have no `.dim()`, so the branch raised `AttributeError` and masked the intended `ValueError`. Switched to the `.ndim` property. This path is unreachable from Pipecat's own pipeline (`voice_confidence` always feeds a 1-D buffer); it only matters when `SileroOnnxModel` is used standalone with a malformed shape, where it now surfaces a useful error.
- **`voice_confidence` re-wrapped an already-converted array.** The second `np.frombuffer(audio_int16, ...)` reinterpreted an already-int16 `ndarray` as int16 again — a redundant no-op view that `.astype()` immediately copies over. Output is bit-for-bit identical; the simplified single-conversion form matches the idiom used by the other VAD/turn analyzers (`krisp_viva_vad.py`, `base_smart_turn.py`, etc.).

## Testing

`uv run pytest tests/test_silero_vad.py` — added tests covering:

- A >2-D input raises `ValueError` (with the dim count in the message), not `AttributeError`. Verified to fail on the pre-fix code.
- The int16→float32 conversion feeds the model the expected normalized values.
- End-to-end smoke: silence runs through the real ONNX model and returns a valid confidence in `[0, 1)`.

No runtime behavior change for bots: Fix 1 only changes an error-message path; Fix 2 produces identical output (verified bit-for-bit across the full int16 value space).

## Fixes

- Fixes #4808

### DIFF
--- a/changelog/4820.fixed.md
+++ b/changelog/4820.fixed.md
@@ -1,0 +1,1 @@
+- Fixed `SileroOnnxModel` raising a confusing `AttributeError` instead of a `ValueError` when given an input audio chunk with too many dimensions. The validation error message called `x.dim()` (a PyTorch method) on a NumPy array; it now uses `x.ndim` and surfaces the intended "Too many dimensions" message.
--- a/src/pipecat/audio/vad/silero.py
+++ b/src/pipecat/audio/vad/silero.py
@@ -64,7 +64,7 @@ class SileroOnnxModel:
         if np.ndim(x) == 1:
             x = np.expand_dims(x, 0)
         if np.ndim(x) > 2:
-            raise ValueError(f"Too many dimensions for input audio chunk {x.dim()}")
+            raise ValueError(f"Too many dimensions for input audio chunk {x.ndim}")
 
         if sr not in self.sample_rates:
             raise ValueError(
@@ -207,7 +207,7 @@ class SileroVADAnalyzer(VADAnalyzer):
         try:
             audio_int16 = np.frombuffer(buffer, np.int16)
             # Divide by 32768 because we have signed 16-bit data.
-            audio_float32 = np.frombuffer(audio_int16, dtype=np.int16).astype(np.float32) / 32768.0
+            audio_float32 = audio_int16.astype(np.float32) / 32768.0
             new_confidence = self._model(audio_float32, self.sample_rate)[0]
 
             # We need to reset the model from time to time because it doesn't
--- a/tests/test_silero_vad.py
+++ b/tests/test_silero_vad.py
@@ -1,0 +1,75 @@
+#
+# Copyright (c) 2024-2026, Daily
+#
+# SPDX-License-Identifier: BSD 2-Clause License
+#
+
+import unittest
+
+import numpy as np
+
+from pipecat.audio.vad.silero import SileroVADAnalyzer
+
+
+class TestSileroVAD(unittest.TestCase):
+    """Tests for the Silero VAD analyzer and its ONNX model wrapper."""
+
+    @classmethod
+    def setUpClass(cls):
+        # Loading the ONNX model once is enough; it's reused across tests.
+        cls.analyzer = SileroVADAnalyzer(sample_rate=16000)
+        cls.analyzer.set_sample_rate(16000)
+        cls.model = cls.analyzer._model
+
+    def test_too_many_dimensions_raises_value_error(self):
+        """A >2-D input must raise a descriptive ValueError, not AttributeError.
+
+        The error message previously called ``x.dim()`` (a PyTorch method) on a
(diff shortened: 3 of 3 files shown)


filipi87 (approved):
LGTM.
Also, nice catch on the numpy simplification.

PR #4821, opened 2026-06-19
feat(xai): align TTS services with xAI API and add word-level timestamps

markbackman:
## Summary

Aligns the xAI TTS services (`src/pipecat/services/xai/tts.py`) with xAI's current TTS API, and adds word-level timestamp support to the streaming WebSocket service. Done as four reviewable commits.

## Changes

**Correctness fixes**
- HTTP `language` is now always sent (it's required by the API), falling back to xAI's documented `"auto"` when unset.
- Corrected the WS class docstring (it claimed `bit_rate` was a query param when it wasn't).

**Tunable params**
- Added `speed`, `optimize_streaming_latency`, `text_normalization`, and `with_timestamps` to the settings classes, wired into the HTTP payload and the WS URL (booleans lowercased for the query string).

**Interruption via `text.clear`**
- `XAITTSService` now extends `WebsocketTTSService` (matching Cartesia/Rime/ElevenLabs) instead of `InterruptibleTTSService`. On barge-in it sends `{"type": "text.clear"}` on the persistent socket instead of tearing down and reconnecting the WebSocket. Also acknowledges the server's `audio.clear` message.

**Word-level timestamps** (WS, on by default via `with_timestamps`)
- Converts xAI's per-character `graph_chars`/`graph_times` into per-word `TTSTextFrame`s.
- Handles xAI's actual delivery model: timestamps arrive in their own (sometimes audio-less) deltas, decoupled from the audio stream; character times are absolute across the utterance; partial words are carried across chunk boundaries.
- `push_text_frames` is derived from `with_timestamps` so per-word frames drive the text output (normal aggregated path when disabled).
- `append_trailing_space=True` keeps the sentence aggregator's separate segments from gluing (e.g. `"xAI."` + `"What"`), since xAI concatenates consecutive text deltas verbatim but preserves spaces it is given.
- `with_timestamps` is intentionally **not** exposed on the HTTP service: it switches that endpoint to a JSON envelope the streaming reader can't parse.

> Note: xAI delivers timestamps in coarse batches, so word frames are emitted in bursts; each frame still carries an accurate `pts` (documented on the class), so consumers should schedule off `pts` rather than arrival time.

## Behavior Changes

- `XAITTSService` now defaults to `with_timestamps=True`, so it emits per-word `TTSTextFrame`s (with `push_text_frames=False`) by default instead of the previous single aggregated text frame per utterance. Pass `settings=XAITTSService.Settings(with_timestamps=False)` to restore the old behavior.
- On interruption the WebSocket is no longer torn down and reconnected; the utterance is cancelled in place via `text.clear`.

## Testing

- `uv run pytest tests/test_xai_tts.py` — 10 tests passing (HTTP/WS happy paths, settings propagation, `text.clear` interruption, `audio.clear` handling, word-timestamp emission incl. audio-less delta).
- Verified the timestamp behavior against the live xAI API (absolute times, audio-less timestamp deltas, and that xAI preserves spaces in the text we send).

### DIFF
--- a/changelog/4821.added.2.md
+++ b/changelog/4821.added.2.md
@@ -1,0 +1,1 @@
+- Added tunable parameters to the xAI TTS services: `speed`, `optimize_streaming_latency`, and `text_normalization` (plus `with_timestamps` on the WebSocket service). Set them via the service's `Settings`, e.g. `XAITTSService.Settings(speed=1.1)`.
--- a/changelog/4821.added.md
+++ b/changelog/4821.added.md
@@ -1,0 +1,1 @@
+- Added word-level timestamps to `XAITTSService`. When `with_timestamps` is enabled (now the default), xAI's per-character timing is converted into per-word `TTSTextFrame` objects, each carrying an accurate `pts`. Note that xAI delivers timestamps in coarse batches, so word frames are emitted in bursts; consumers should schedule off `pts` rather than arrival time.
--- a/changelog/4821.changed.md
+++ b/changelog/4821.changed.md
@@ -1,0 +1,1 @@
+- `XAITTSService` now cancels the current utterance on interruption by sending a `text.clear` message over the existing WebSocket, instead of disconnecting and reconnecting. This makes barge-in faster by avoiding a reconnect on every interruption.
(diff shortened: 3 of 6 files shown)


aconchillo (approved):

aconchillo:
Nice! LGTM!

PR #4823, opened 2026-06-19
Rebind auto-registered tool handler when an advertised set re-declares it

### LINKED ISSUE #4822: Same-named tool in a new advertised set keeps the previous handler (regression of flows #269/#270 under context-driven registration)
### Summary

pipecat-ai-flows #269 / #270 (released in flows 1.1.1) fixed *"Re-declaring a function in a new node silently keeps the previous node's handler bound"* by always re-registering the node's functions on the LLM service. With the move to context-driven tool registration in **pipecat 1.4.0 + pipecat-flows 1.3.0**, that fix appears to have regressed — and the responsible mechanism now lives in pipecat core, so the repro below reproduces it on a bare `LLMService` with no flows involved.

### Cause

In flows 1.3.0, `FlowManager._set_node` no longer calls `llm.register_function`; it advertises the node's tools via `_update_llm_context(functions=ToolsSchema(...))` (an `LLMSetToolsFrame` carrying handlers on the schemas). pipecat 1.4.0 then registers handlers from the advertised `LLMContext` in `LLMService._sync_registered_tool_handlers` (run on every `LLMContextFrame`). That method:

- prunes handlers for names that **leave** the advertised set (orphan unregistration — great), but
- **skips registering a name that is already registered** (`_register_advertised_tool_handlers`: *"Any tool whose name is already registered … is left untouched"*).

So when two consecutive advertised tool sets both carry a tool with the **same name but a different handler** (e.g. a per-node closure), the second handler is silently ignored, because the name never leaves the advertised set (no prune) and the register step skips it (already present). This is exactly the #269 scenario, now in core.

### Minimal repro (pipecat 1.4.0, no flows)

```python
from pipecat.adapters.schemas.tools_schema import ToolsSchema
from pipecat.services.llm_service import LLMService

async def hA(params): ...
async def hB(params): ...

def tool(name, h):
    return ToolsSchema(standard_tools=[
        FunctionSchema(name=name, description="d", properties={}, required=[], handler=h)
    ])

llm = LLMService()
llm._sync_registered_tool_handlers(tool("x", hA))
llm._sync_registered_tool_handlers(tool("x", hB))  # re-advertise same name, new handler
# llm._functions["x"] still wraps hA, not hB
```

Observed against `pipecat-ai==1.4.0`:

```
after A: x -> hA
after B: x -> hA      # second handler ignored
x present after advertising only y: False   # orphan unregistration works
```

(Orphan unregistration works: advertising only `tool("y", ...)` afterward does drop `x`, then re-advertising `x` registers it afresh. So the bug only bites when the same name is advertised **continuously across the transition** — directly adjacent tool sets, or always-on tools.)

### Impact

Any tool name shared across two consecutive advertised tool sets with **different** handlers runs the **first** handler instead of the current one, with no warning. In a Flows app this hits any tool name shared across nodes with per-node handlers (global / transition / interrupt tools keyed by the originating node) after a transition.

### Suggested direction

`LLMService` already distinguishes `auto_registered=True` entries (from advertised tool sets) from explicit `register_function` registrations. The fix can live entirely in core: when an advertised set carries a **different** handler for a name that is currently registered as `auto_registered`, rebind it — while still leaving explicit registrations untouched (preserving the existing "explicit wins / repeated frames don't churn" invariant). Happy to send a PR if a maintainer confirms this is the preferred shape (vs. e.g. flows forcing re-registration downstream).

### Versions

`pipecat-ai==1.4.0`, `pipecat-ai-flows==1.3.0`.

### Note

We worked around it downstream by making our handlers node-agnostic (resolving the current node at call time), so this is not blocking us — filing for visibility, since it reverses the pipecat-ai-flows #269/#270 fix under the new context-driven registration model.

markbackman:
## Summary

Fixes #4822 — an `LLMService` regression where re-advertising a tool with the **same name but a different handler** silently kept the previous handler bound.

`LLMService._register_advertised_tool_handlers` skipped any tool name already present in the registry. So when two consecutive advertised tool sets carry the same name with different handlers (e.g. a fresh per-node closure after a Pipecat Flows transition), the second handler was silently ignored: the name never left the advertised set (no prune) and the register step skipped it (already present), so the stale handler kept firing while the LLM was told it was calling the new node's tool.

This is the same bug pipecat-ai-flows fixed in #269/#270, re-introduced in core when Flows 1.3.0 / pipecat 1.4.0 moved to context-driven tool registration.

## Change

- Add `_advertised_handler_changed()` — compares the advertised handler to the registered one by identity, unwrapping `DirectFunctionWrapper` to its underlying `.function` on both sides.
- In both loops of `_register_advertised_tool_handlers`, rebind an entry when it is `auto_registered` **and** the advertised handler changed; leave explicit `register_function` registrations and unchanged handlers untouched.

This restores the method's documented invariant — *the registry matches what the LLM can see* — while preserving the existing guarantees: explicit registrations win, and repeated identical context frames don't churn (identity check).

## Status

Opened as a **draft** for team discussion. The open design question (see #4822): do we want core to bear this, vs. steering users toward a single node-agnostic handler that resolves the current node at call time? The fix is small and localized, but the triggering pattern (same tool name, different per-node handler) is debatable outside the node-scoped Flows context.

## Notes / non-issues

- **Flows needs no change** — the fix lives entirely in core.
- **Async-tool-cancellation setup** is unaffected: it's only invoked at `start()`, so a rebind behaves exactly like initial registration.

## Test plan

- New `TestRebindChangedAdvertisedHandler` in `tests/test_auto_register_tools.py`: direct-function rebind, schema-handler rebind (the issue's repro), no-churn on identical frames, and explicit-registration-wins.
- `uv run pytest tests/test_auto_register_tools.py` → 51 passed.
- `uv run pytest tests/test_llm_service.py tests/test_llm_switcher.py tests/test_realtime_service_tool_sync.py` → 55 passed.
- `uv run ruff check` / `ruff format --check` clean.
- Issue's repro snippet now prints `after B: x -> hB` with orphan unregistration intact.


markbackman:
Outdated. Closing.

kompfner:
Just saw this—I actually think that, while using the same-named tool to do different things might be a bit unusual, it's not wrong per se. You can imagine that something like `log_out` or `talk_to_manager` (totally fictional tools) might do something slightly different (maybe even just do different logging) depending on where you are in the conversation, in a way that's totally opaque to the LLM. The fix here looks good, and I would actually lobby to take it. 👍 Maybe we can chat when you get back.

kompfner:
Re-opening for consideration (and approving).

kompfner (approved):

PR #4824, opened 2026-06-19
fix(user_mute): tolerate double / orphan function-call result frames

Maurotb:
`FunctionCallUserMuteStrategy` drops a tool call id from its in-progress set with `set.remove` on every `FunctionCallResultFrame` / `FunctionCallCancelFrame`. Those frames can carry an id the strategy is not tracking, and `remove` raises `KeyError`.

`FrameProcessor.__process_frame` catches it and calls `push_error(fatal=False)`, so the pipeline stays up, but the user aggregator abandons the rest of its handling for that frame and an `ErrorFrame` goes upstream to the application.

### Where untracked ids come from

- **The built-in `cancel_async_tool_call`.** `run_function_calls` excludes it from `FunctionCallsStartedFrame` (`llm_service.py:1258-1263`) but still broadcasts its result, so the strategy sees a result for an id it never added. Any bot with an async tool and `enable_async_tool_cancellation=True` hits this the first time the LLM cancels a call.
- **Async tools reporting progress.** `result_callback(..., properties=FunctionCallResultProperties(is_final=False))` broadcasts a result frame per update plus the final one, so the same id is reported finished more than once — the pattern in `examples/function-calling/function-calling-*-async-stream.py`.
- **Bus re-delivery.** In multi-worker topologies a bridge can re-deliver a result frame a downstream worker already handled.

### Fix

`set.discard` — identical to `remove` when the id is present, a no-op when it isn't. Ids the strategy is not tracking are ignored.

### Tests

`tests/test_user_mute_strategy.py` covers the strategy directly:

- `test_repeated_result_frame_stays_unmuted`
- `test_unknown_tool_call_id_stays_unmuted`
- `test_unknown_tool_call_id_leaves_other_calls_muted` — an untracked id must not unmute while another call is still in progress

`tests/test_llm_service.py` drives a real `LLMService` and feeds its broadcast frames to the strategy, pinning the two single-worker paths at their source:

- `test_builtin_cancel_tool_allows_user_mute_cleanup`
- `test_intermediate_results_allow_user_mute_cleanup`

### Follow-up, not in this PR

The strategy mutes on `FunctionCallsStartedFrame` for every user-visible call, async ones included, so the user is muted for as long as a background tool runs. The unmute on an intermediate update is incidental rather than a designed escape hatch — an async tool that reports no progress keeps the user muted until it finishes. Now that `@tool_options(cancel_on_interruption=False)` keeps a tool call alive without suppressing user speech, the strategy is a candidate for deprecation. Tracked separately.

### DIFF
--- a/changelog/4824.fixed.md
+++ b/changelog/4824.fixed.md
@@ -1,0 +1,7 @@
+- Fixed `FunctionCallUserMuteStrategy.process_frame` raising `KeyError` when the
+  same `FunctionCallResultFrame` / `FunctionCallCancelFrame` is delivered twice
+  (or once with no preceding `FunctionCallsStartedFrame`). In multi-worker bus
+  topologies a child worker can see the same result frame the parent already
+  handled, and the second delivery would tear down the frame loop. Switched the
+  removal to `set.discard`, which is identical in single-worker (id always
+  present → removed) and a silent no-op on absent ids.
--- a/changelog/4824.fixed.md
+++ b/changelog/4824.fixed.md
@@ -1,7 +1,1 @@
-- Fixed `FunctionCallUserMuteStrategy.process_frame` raising `KeyError` when the
-  same `FunctionCallResultFrame` / `FunctionCallCancelFrame` is delivered twice
-  (or once with no preceding `FunctionCallsStartedFrame`). In multi-worker bus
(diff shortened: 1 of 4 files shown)


markbackman (approved):

PR #4839, opened 2026-06-20
fix: clarify segmented stt audio format

### LINKED ISSUE #4836: [Detail Bug] STT: Local transcribers misinterpret WAV-wrapped segments as raw PCM, corrupting audio and changing transcriptions
# Detail Bug Report

https://app.detail.dev/org_1fa3c5ff-f485-452b-8fbd-870cbd985e00/bugs/bug_7bd03f33-f177-4a0a-9be5-3efa7e062b3a

Introduced in [#421](https://github.com/pipecat-ai/pipecat/pull/421) by @aconchillo on Aug 26, 2024

# Summary
- **Context**: `SegmentedSTTService` was added for VAD-based audio segmentation and passes the segmented audio to each STT service via `run_stt()`.
- **Bug**: `MoonshineSTTService` and `WhisperSTTService` receive WAV-formatted audio from `SegmentedSTTService` but treat it as raw PCM, corrupting the first 22 audio samples.
- **Actual vs. expected**: Actual: segmented audio is WAV bytes and the 44-byte WAV header is interpreted as 22 `int16` samples, producing different model inputs and different transcriptions. Expected: local STT services should receive raw PCM, or they should strip the WAV header before converting bytes to `int16`.
- **Impact**: 60% of audio segments produce DIFFERENT transcriptions when WAV header is misinterpreted. The corrupted samples contain values up to 97.7% of full scale (39x louder than typical speech).

# Code with Bug
```python
async def _handle_user_stopped_speaking(self, frame: VADUserStoppedSpeakingFrame):
    content = io.BytesIO()
    wav = wave.open(content, "wb")
    wav.setsampwidth(2)
    wav.setnchannels(1)
    wav.setframerate(self.sample_rate)
    wav.writeframes(self._audio_buffer)
    wav.close()
    content.seek(0)
    await self.process_generator(self.run_stt(content.read()))  # <-- BUG 🔴 always passes WAV bytes (no opt-out)
```

```python
async def run_stt(self, audio: bytes) -> AsyncGenerator[Frame, None]:
    audio_float = (np.frombuffer(audio, dtype=np.int16).astype(np.float32) / 32768.0).tolist()  # <-- BUG 🔴 interprets WAV header as PCM samples
```

```python
async def run_stt(self, audio: bytes) -> AsyncGenerator[Frame, None]:
    audio_float = np.frombuffer(audio, dtype=np.int16).astype(np.float32) / 32768.0  # <-- BUG 🔴 interprets WAV header as PCM samples
```

# Explanation
- `SegmentedSTTService` unconditionally wraps the buffered audio into a WAV container and passes the resulting bytes to `run_stt()`.
- Local STT implementations (`MoonshineSTTService`, `WhisperSTTService`) document/assume `run_stt()` receives raw 16-bit PCM and immediately do `np.frombuffer(..., dtype=np.int16)`.
- When WAV bytes are treated as raw PCM, the 44-byte WAV header becomes 22 `int16` samples, injecting a loud artifact at the start of each segment (reported max value ~32000, near full scale). With low-latency/rapid VAD, those corrupted samples can overlap actual speech, altering model input and changing transcriptions.

## Codebase Inconsistency
`run_stt()` is called with different audio formats depending on the caller:

```python
# EvalTranscriber.transcribe()
async for frame in self._service.run_stt(pcm):  # Raw PCM, not WAV
    ...
```

This explains why local services can appear to work in eval flows (raw PCM) while failing in segmented/VAD flows (WAV bytes).

# Recommended Fix
Add a configuration/parameter to `SegmentedSTTService` to control WAV wrapping (default on for cloud STT services; off for local models), and pass raw PCM bytes when WAV wrapping is disabled.

# History
This bug was introduced in commit 4e0ece17b. The commit added `SegmentedSTTService` for VAD-based audio segmentation and changed `WhisperSTTService` to inherit from it. The parent class wraps audio in WAV format and passes WAV bytes to `run_stt()`, but the child class continued to treat the bytes as raw PCM, interpreting the 44-byte WAV header as 22 int16 audio samples. The bug slipped in because the original implementation used `BinaryIO` (file-like object) which Whisper could parse correctly, while the new implementation changed to raw bytes that were misinterpreted as PCM data.

Rahuwale123:
This fixes the segmented STT audio-format mismatch in issue #4836.

Problem

`SegmentedSTTService` always wrapped buffered speech in a WAV container before calling `run_stt()`. That worked for file/container-based providers, but it was incompatible with the local Moonshine and Whisper/MLX implementations, which consume raw 16-bit PCM.

Fix

`SegmentedSTTService` now has a protected `_segment_audio_format()` contract that defaults to `"wav"` for backward compatibility. Moonshine and the local Whisper/MLX service family override that contract to `"pcm"`, so the buffered segment is passed through unchanged for those implementations.

Compatibility

This keeps the existing WAV behavior for the provider families that already expect it, including OpenAI/Groq, Fal, ElevenLabs, and NVIDIA. There is no runtime RIFF sniffing and no new public configuration surface.

Validation

- `git diff --check`
- `python3 -m py_compile src/pipecat/services/stt_service.py src/pipecat/services/moonshine/stt.py src/pipecat/services/whisper/stt.py tests/test_segmented_stt_service.py tests/test_evals_transcribe.py`
- Offline runtime validation of the new segment-format path:
  - PCM mode passed the original buffered bytes through unchanged
  - WAV mode produced a valid WAV container with the expected sample rate and payload
  - `EvalTranscriber` still passed raw PCM through to its STT service

Issue

Closes #4836.

### DIFF
--- a/changelog/4836.fixed.md
+++ b/changelog/4836.fixed.md
@@ -1,0 +1,1 @@
+- Fixed segmented STT services so Moonshine and local Whisper/MLX variants keep receiving raw PCM segments while WAV-based providers continue to receive WAV containers.
--- a/src/pipecat/services/moonshine/stt.py
+++ b/src/pipecat/services/moonshine/stt.py
@@ -19,6 +19,7 @@ from enum import StrEnum
 
 import numpy as np
 from loguru import logger
+from typing_extensions import override
 
 from pipecat.frames.frames import Frame, TranscriptionFrame
 from pipecat.services.settings import STTSettings, assert_given
@@ -113,6 +114,11 @@ class MoonshineSTTService(SegmentedSTTService):
             True, as this service supports metric generation.
         """
         return True
+
+    @override
+    def _segment_audio_format(self) -> str:
+        """Moonshine consumes raw 16-bit PCM segments."""
+        return "pcm"
 
     def _load(self) -> Transcriber:
         """Download (first time) and load the Moonshine model.
--- a/src/pipecat/services/stt_service.py
+++ b/src/pipecat/services/stt_service.py
@@ -722,7 +722,9 @@ class SegmentedSTTService(STTService):
 
     Requires VAD to be enabled in the pipeline to function properly. Maintains a
     small audio buffer to account for the delay between actual speech start and
-    VAD detection.
+    VAD detection. Subclasses can override the protected
+    :meth:`_segment_audio_format` contract to choose whether the buffered segment
+    is delivered to ``run_stt()`` as a WAV container or raw PCM bytes.
     """
 
     def __init__(self, *, sample_rate: int | None = None, **kwargs):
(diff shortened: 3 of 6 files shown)


markbackman:
Thanks for offering a fix. We landed in a similar place, which you can see here:
https://github.com/pipecat-ai/pipecat/pull/4896

Closing this, but thank you 🙇

PR #4840, opened 2026-06-20
Fix WorkerBus dropping all future messages to a subscriber after a hook raises

### LINKED ISSUE #4827: [Detail Bug] Bus: Subscriber exceptions permanently stop future message delivery to that subscriber
# Detail Bug Report

https://app.detail.dev/org_1fa3c5ff-f485-452b-8fbd-870cbd985e00/bugs/bug_a13be945-b05a-4e75-8130-8a7cdf95e6c2

Introduced in [#4493](https://github.com/pipecat-ai/pipecat/pull/4493) by @aconchillo on May 14, 2026

# Summary
- **Context**: The `WorkerBus` class dispatches messages through `_router_task` and `_data_dispatch_task`.
- **Bug**: An uncaught exception in a subscriber's `on_bus_message` terminates the dispatch task permanently, stopping ALL future message delivery to that subscriber.
- **Actual vs. expected**: **Actual**: "the dispatch task dies and the second message is never received" / "Subsequent messages (`BusJobResponseMessage` #2 and `BusJobUpdateMessage`) were never delivered". **Expected**: "should be received but isn't" / dispatch should continue delivering subsequent messages after logging the exception.
- **Impact**: Documented lifecycle methods (`on_activated`, `on_job_response`, etc.) are NOT protected by `_call_event_handler`. Exceptions in these methods — from user code or built-in framework code — crash the dispatch task.

# Code with Bug
```python
async def on_bus_message(self, message: BusMessage) -> None:
    # ... dispatch to _handle_* methods based on message type ...
    if isinstance(message, (BusJobResponseMessage, BusJobResponseUrgentMessage)):
        await self._handle_job_response(message)  # <-- BUG 🔴 calls unprotected lifecycle methods before _call_event_handler
    # ... other message types ...
    await self._call_event_handler("on_bus_message", message)  # <-- Only protects event handlers
```

```python
async def _handle_job_response(self, message):
    await self.on_job_response(message)  # <-- BUG 🔴 NOT protected, user-overrideable
    await self._call_event_handler("on_job_response", message)  # <-- protected
    # ...
```

```python
async def _maybe_activate(self):
    # ...
    await self.on_activated(self._activation_args)  # <-- BUG 🔴 NOT protected, user-overrideable
    await self._call_event_handler("on_activated", self._activation_args)  # <-- protected
```

# Explanation
- The bus delivers messages by running per-subscriber dispatch tasks (`_router_task` / `_data_dispatch_task`) that call `subscriber.on_bus_message`.
- `BaseWorker.on_bus_message` routes messages into `_handle_*` methods, and those methods call documented lifecycle hooks (e.g., `on_job_response`, `on_activated`) directly.
- These lifecycle hooks are intentionally overridable and can also throw from built-in framework code (e.g., websocket connect failures).
- When any of these calls raises, the exception is logged by `TaskManager`, but the dispatch task terminates; the subscriber then never receives any future messages (including potential cleanup/cancel messages).

## Codebase Inconsistency
- The codebase uses `_call_event_handler(...)` to wrap/guard event handler invocations, but the documented lifecycle methods are invoked directly (unwrapped) immediately beforehand, even though they are part of the public override surface.

# Failing Test
```python
import asyncio
from pipecat.bus import AsyncQueueBus, BusJobResponseMessage, BusJobUpdateMessage
from pipecat.workers.base_worker import BaseWorker
from pipecat.registry import WorkerRegistry


async def test_on_job_response_exception():
    bus = AsyncQueueBus()
    tm = TaskManager()
    await bus.setup(tm)
    registry = WorkerRegistry(runner_name='test-runner')

    class FailingWorker(BaseWorker):
        def __init__(self):
            super().__init__(name='failing_worker')
            self.received_after_error = []
            self.response_count = 0

        async def on_job_response(self, message):
            self.response_count += 1
            raise RuntimeError(f'Simulated on_job_response failure')

        async def on_job_update(self, message):
            self.received_after_error.append(message)

    worker = FailingWorker()
    await worker.attach(registry=registry, bus=bus)
    await bus.start()

    # Send job response #1 - will cause exception
    await bus.send(BusJobResponseMessage(
        source='remote_worker', target='failing_worker',
        job_id='test-job-1', status='completed',
    ))
    await asyncio.sleep(0.1)

    # Send job response #2 - should trigger another call if dispatch task alive
    await bus.send(BusJobResponseMessage(
        source='remote_worker', target='failing_worker',
        job_id='test-job-2', status='completed',
    ))
    await asyncio.sleep(0.1)

    # Send job update - should be received if dispatch task alive
    await bus.send(BusJobUpdateMessage(
        source='remote_worker', target='failing_worker',
        job_id='test-job-3', update={'progress': 50},
    ))
    await asyncio.sleep(0.1)

    await bus.stop()
    print(f'response_count: {worker.response_count}')  # Expected: 2, Actual: 1
    print(f'received_after_error: {len(worker.received_after_error)}')  # Expected: 1, Actual: 0


asyncio.run(test_on_job_response_exception())
```

**Output:**
```
2026-06-19 14:00:54.859 | ERROR | pipecat.utils.asyncio.task_manager:run_coroutine:169 - AsyncQueueBus#0::bus_data_failing_worker unexpected exception (<string>:25): Simulated on_job_response failure #1
response_count: 1
received_after_error: 0
BUG CONFIRMED: Second response was never processed because dispatch task died
BUG CONFIRMED: on_job_update was never called because dispatch task died
```

# Recommended Fix
```python
async def _router_task(self, sub: BusSubscription):
    """Route system messages inline, data messages to the data queue."""
    try:
        while True:
            message = await sub.queue.get()
            if isinstance(message, BusSystemMessage):
                try:
                    await sub.subscriber.on_bus_message(cast(BusMessage, message))
                    raise
                except Exception:
                    logger.exception(
                        f"{self}: exception in subscriber '{sub.subscriber.name}' "
                        f"processing system message, continuing"
                    )
            else:
                sub.data_queue.put_nowait(message)
        pass

async def _data_dispatch_task(self, sub: BusSubscription):
    """Process data messages sequentially from the data queue."""
    try:
        while True:
            message = await sub.data_queue.get()
            try:
                await sub.subscriber.on_bus_message(message)
                raise
            except Exception:
                logger.exception(
                    f"{self}: exception in subscriber '{sub.subscriber.name}' "
                    f"processing data message, continuing"
                )
        pass
```

# History
This bug was introduced in commit 281d81f50. The commit added the entire `TaskBus` abstraction with `_router_task` and `_data_dispatch_task` methods to dispatch messages to subscribers, but neither method wrapped the `on_bus_message` call in exception handling. The original implementation caught only `CancelledError` to handle graceful shutdown, missing that exceptions from subscriber callbacks would terminate the dispatch task entirely, leaving the subscriber permanently unable to receive messages.

immuhammadfurqan:
## Problem

Fixes #4827.

A `WorkerBus` delivers messages to each subscriber through two per-subscriber dispatch tasks: `_router_task` (system messages, dispatched inline) and `_data_dispatch_task` (data messages). Both wrapped their `while True` loop in `except asyncio.CancelledError` only.

`BaseWorker.on_bus_message` routes messages into `_handle_*` methods, which call documented, **user-overridable** lifecycle hooks (`on_job_response`, `on_activated`, …) directly — unwrapped. When any of those raises (from user code, or from built-in framework code such as a websocket connect failure), the exception propagates all the way up and out of the dispatch loop. The task coroutine returns, and since these tasks are created once per subscription and never restarted, **the subscriber silently stops receiving all future messages — including cancel/cleanup.**

## Fix

Wrap the `on_bus_message` call inside **both** dispatch tasks so a subscriber exception is logged and the loop continues. `asyncio.CancelledError` is re-raised to preserve graceful shutdown.

This is fixed at the dispatch-task chokepoint rather than per-hook in `BaseWorker`, so it protects every subscriber against any throwing hook and can't be silently regressed when new `_handle_*` methods are added. Both paths are fixed — the data path (per the original report) and the system path in `_router_task`, which had the identical flaw.

## Testing

New `TestSubscriberExceptionIsolation` in `tests/test_bus.py` with a subscriber that raises on its first message, one test per path, asserting the next message is still delivered:

- **Without the fix:** both tests fail (the dispatch task dies; the second message never arrives).
- **With the fix:** both pass.
- Full `tests/test_bus.py` + `tests/test_base_worker.py` suites: 73 passed. Ruff lint/format clean.

## Note for reviewers

For system messages, swallowing an exception could mask a genuine lifecycle failure. Keeping the dispatch task alive (and logging the exception) still seems strictly better than killing all future delivery to the subscriber, but happy to adjust the policy for system vs. data messages if you'd prefer something different.

### DIFF
--- a/changelog/4827.fixed.md
+++ b/changelog/4827.fixed.md
@@ -1,0 +1,1 @@
+- Fixed `WorkerBus` permanently stopping message delivery to a subscriber when that subscriber's `on_bus_message` (or an overridable lifecycle hook such as `on_job_response`/`on_activated`) raised an exception. The router and data dispatch tasks now log the exception and keep running, so subsequent messages — including cancel/cleanup — are still delivered.
--- a/src/pipecat/bus/bus.py
+++ b/src/pipecat/bus/bus.py
@@ -15,6 +15,8 @@ from abc import abstractmethod
 from dataclasses import dataclass, field
 from typing import cast
 
+from loguru import logger
+
 from pipecat.bus.messages import BusLocalMessage, BusMessage, BusSystemMessage
 from pipecat.bus.queue import BusMessageQueue
 from pipecat.bus.subscriber import BusSubscriber
@@ -170,7 +172,18 @@ class WorkerBus(BaseObject):
             while True:
                 message = await sub.queue.get()
                 if isinstance(message, BusSystemMessage):
-                    await sub.subscriber.on_bus_message(cast(BusMessage, message))
+                    # A subscriber exception must not tear down the dispatch task,
+                    # or the subscriber would silently stop receiving all future
+                    # messages (including cancel/cleanup). Log and keep going.
+                    try:
+                        await sub.subscriber.on_bus_message(cast(BusMessage, message))
+                    except asyncio.CancelledError:
+                        raise
+                    except Exception:
+                        logger.exception(
+                            f"{self}: subscriber '{sub.subscriber.name}' raised handling "
+                            f"system message {message}; keeping dispatch alive"
+                        )
                 else:
                     sub.data_queue.put_nowait(message)
         except asyncio.CancelledError:
@@ -181,6 +194,17 @@ class WorkerBus(BaseObject):
         try:
             while True:
                 message = await sub.data_queue.get()
-                await sub.subscriber.on_bus_message(message)
+                # A subscriber exception must not tear down the dispatch task,
+                # or the subscriber would silently stop receiving all future
+                # messages (including cancel/cleanup). Log and keep going.
+                try:
+                    await sub.subscriber.on_bus_message(message)
+                except asyncio.CancelledError:
+                    raise
+                except Exception:
+                    logger.exception(
+                        f"{self}: subscriber '{sub.subscriber.name}' raised handling "
+                        f"data message {message}; keeping dispatch alive"
+                    )
         except asyncio.CancelledError:
             pass
--- a/tests/test_bus.py
+++ b/tests/test_bus.py
@@ -270,5 +270,68 @@ class TestBusMessagePriority(unittest.IsolatedAsyncioTestCase):
         self.assertIsInstance(sub.received[0], BusJobCancelMessage)
 
 
+class _FailingSub(BusSubscriber):
+    """Subscriber that raises on its first message, then records the rest.
+
+    Models a subscriber whose lifecycle hook (e.g. ``on_job_response``) throws:
+    the bus dispatch task must survive so later messages are still delivered.
+    """
+
+    _counter = itertools.count()
+
+    def __init__(self):
+        self._name = f"failing_{next(self._counter)}"
+        self.received = []
+        self._raised = False
+
+    @property
+    def name(self) -> str:
+        return self._name
+
+    async def on_bus_message(self, message):
+        if not self._raised:
+            self._raised = True
+            raise RuntimeError("simulated subscriber failure")
+        self.received.append(message)
+
+
+class TestSubscriberExceptionIsolation(unittest.IsolatedAsyncioTestCase):
+    """A subscriber exception must not stop future delivery to that subscriber."""
+
+    async def test_data_dispatch_survives_subscriber_exception(self):
+        """A raise in on_bus_message must not tear down the data dispatch task."""
+        bus, _ = await create_test_bus()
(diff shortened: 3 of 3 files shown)


aconchillo (approved):

aconchillo:
LGTM. Thank you @immuhammadfurqan !

PR #4841, opened 2026-06-20
Fix bus adapters dropping ToolsSchema.custom_tools during serialization

### LINKED ISSUE #4833: [Detail Bug] Distributed bus transport drops provider-specific LLM tools from context
# Detail Bug Report

https://app.detail.dev/org_1fa3c5ff-f485-452b-8fbd-870cbd985e00/bugs/bug_1f3d5ed9-8ccb-4087-82bb-de0cde4e0c42

Introduced in [#4493](https://github.com/pipecat-ai/pipecat/pull/4493) by @aconchillo on May 14, 2026

# Summary
- **Context**: `LLMContextFrame` crosses the bus in distributed setups.
- **Bug**: `LLMContextAdapter` and `ToolsSchemaAdapter` silently drop `custom_tools` during serialization for network bus transport
- **Actual vs. expected**: `custom_tools` field becomes `None` after round-trip serialization vs `custom_tools` is preserved during serialization, since it contains pure JSON data
- **Impact**: `LLMContextFrame` crossing a network bus loses provider-specific tools (e.g., Gemini's `google_search`)

# Code with Bug
```python
# src/pipecat/bus/adapters/llm_context_adapter.py

async def serialize(self, obj: Any, serialize_value: SerializeFunc) -> dict[str, Any]:
    # ... other fields ...
    if is_given(obj.tools):
        data["tools"] = self._serialize_tools(obj.tools)  # <-- BUG 🔴 custom_tools dropped from bus payload
```

```python
# src/pipecat/bus/adapters/llm_context_adapter.py

def _serialize_tools(self, tools: Any) -> list[dict[str, Any]]:
    return [tool.to_default_dict() for tool in tools.standard_tools]
    # BUG: custom_tools is NOT serialized

def _deserialize_tools(self, data: list[dict[str, Any]]) -> Any:
    tools = []
    for item in data:
        tools.append(FunctionSchema(...))
    return ToolsSchema(standard_tools=tools)
    # BUG: custom_tools not deserialized, defaults to None
```

# Explanation
- In distributed configurations (`RedisBus`, `PgmqBus`), frames are serialized for network transport using registered bus adapters. `LLMContextFrame` is not a lifecycle frame, so it crosses the bus with the full `LLMContext`, including `context.tools`.
- `LLMContextAdapter.serialize()` delegates tool serialization to `_serialize_tools()`, which only emits `standard_tools`. `_deserialize_tools()` reconstructs a `ToolsSchema` without `custom_tools`, so after a serialize/deserialize round-trip `ToolsSchema.custom_tools` becomes `None`.
- This causes silent loss of provider-specific tool definitions (e.g., Gemini `google_search`) when a context frame is sent to a remote worker; downstream service adapters that look for `tools_schema.custom_tools` will not see them.

## Codebase Inconsistency
- Behavior differs by bus type: in-process (`AsyncQueueBus`) does not require serialization, while network buses do, so the same configuration works locally but loses `custom_tools` in distributed mode.
- The Gemini adapter explicitly reads `tools_schema.custom_tools` and extracts adapter-specific entries:
  ```python
  if tools_schema.custom_tools:
      custom_gemini_tools = tools_schema.custom_tools.get(AdapterType.GEMINI, [])
  ```

# Recommended Fix
Update bus serialization/deserialization to include `custom_tools` (pure JSON) for both `ToolsSchemaAdapter` and the `LLMContextAdapter` tool helpers.

```python
# src/pipecat/bus/adapters/tools_schema_adapter.py

def serialize(self, obj: Any, serialize_value: SerializeFunc) -> dict[str, Any]:
    result = {"standard_tools": [tool.to_default_dict() for tool in obj.standard_tools]}
    if obj.custom_tools:
        result["custom_tools"] = {
            adapter_type.value: tools
            for adapter_type, tools in obj.custom_tools.items()
        }
    return result

def deserialize(self, data: dict[str, Any], ...) -> Any:
    tools = [...]  # existing code
    custom_tools = None
    if "custom_tools" in data:
        custom_tools = {
            AdapterType(key): value
            for key, value in data["custom_tools"].items()
        }
    return ToolsSchema(standard_tools=tools, custom_tools=custom_tools)
```

Apply the same logic to `LLMContextAdapter._serialize_tools()` / `_deserialize_tools()`.

# History
This bug was introduced in commit 281d81f50. The commit added bus adapters for network transport of `ToolsSchema` and `LLMContext`, but only serialized `standard_tools`, omitting the `custom_tools` field that had been added to `ToolsSchema` five weeks earlier (commit 4bef85e36, 2026-04-07). The adapters have not been modified since their creation.

immuhammadfurqan:
## Problem

Fixes #4833.

`ToolsSchemaAdapter` and `LLMContextAdapter` (the bus type adapters used for network transport) only serialized `standard_tools`. `ToolsSchema.custom_tools` — provider-specific tool definitions such as Gemini's `google_search` — was dropped on serialize and reconstructed as `None`, so it did not survive a serialize/deserialize round-trip.

In distributed setups (`RedisBus`/`PgmqBus`), an `LLMContext` crossing the bus to a remote worker silently lost these tools. Remote LLM adapters gate on `tools_schema.custom_tools` (`gemini_adapter`, `open_ai_adapter`, `open_ai_realtime_adapter`, `open_ai_responses_adapter`), so the provider request was built without them. In-process (`AsyncQueueBus`) was unaffected because it skips serialization — so the same config worked locally and broke only in distributed mode.

## Fix

`custom_tools` is `dict[AdapterType, list[dict]]` of pure JSON, so it is trivially serializable: emit the enum keys as their string values on serialize, restore them via `AdapterType(key)` on deserialize. Applied symmetrically to both adapters, only when `custom_tools` is present.

## Testing

New `TestCustomToolsRoundTrip` in `tests/test_serializers.py`:
- `test_tools_schema_adapter_preserves_custom_tools` — direct `ToolsSchemaAdapter` round-trip.
- `test_llm_context_frame_preserves_custom_tools_over_bus` — the realistic end-to-end path: a `BusFrameMessage` carrying an `LLMContextFrame` round-tripped through the full `JSONMessageSerializer`.

Both assert `standard_tools` **and** `custom_tools` survive.

- **Without the fix:** both fail (`None != {AdapterType.GEMINI: [...]}`).
- **With the fix:** both pass.
- No regressions: `test_serializers.py` + `test_redis_bus.py` + `test_pgmq_bus.py` + `test_frame_adapters.py` = 53 passed. Ruff lint/format clean.

## Note for reviewers

This is a deliberately minimal in-place fix. The two adapters' tool (de)serialization is near-identical duplication — which is why this bug existed in both places — so a reasonable follow-up would be to have `LLMContextAdapter` reuse `ToolsSchemaAdapter` for tools. Happy to fold that in here if you'd prefer.

### DIFF
--- a/changelog/4833.fixed.md
+++ b/changelog/4833.fixed.md
@@ -1,0 +1,1 @@
+- Fixed the bus type adapters (`ToolsSchemaAdapter` and `LLMContextAdapter`) silently dropping `ToolsSchema.custom_tools` during serialization. Provider-specific tools (e.g. Gemini's `google_search`) are now preserved when an `LLMContext`/`ToolsSchema` crosses a network bus (`RedisBus`/`PgmqBus`), so remote workers no longer lose them.
--- a/src/pipecat/bus/adapters/llm_context_adapter.py
+++ b/src/pipecat/bus/adapters/llm_context_adapter.py
@@ -11,7 +11,7 @@ from typing import Any
 from openai import NOT_GIVEN as OPENAI_NOT_GIVEN
 
 from pipecat.adapters.schemas.function_schema import FunctionSchema
-from pipecat.adapters.schemas.tools_schema import ToolsSchema
+from pipecat.adapters.schemas.tools_schema import AdapterType, ToolsSchema
 from pipecat.bus.adapters.base import DeserializeFunc, SerializeFunc, TypeAdapter
 from pipecat.processors.aggregators.llm_context import (
     LLMContext,
@@ -90,12 +90,21 @@ class LLMContextAdapter(TypeAdapter):
             )
         return deserialize_value(data)
 
-    def _serialize_tools(self, tools: Any) -> list[dict[str, Any]]:
-        return [tool.to_default_dict() for tool in tools.standard_tools]
+    def _serialize_tools(self, tools: Any) -> dict[str, Any]:
+        result: dict[str, Any] = {
+            "standard_tools": [tool.to_default_dict() for tool in tools.standard_tools]
+        }
+        if tools.custom_tools:
+            # custom_tools is dict[AdapterType, list[dict]] of pure JSON; serialize
+            # the enum keys to their string values so it survives network transport.
+            result["custom_tools"] = {
+                adapter_type.value: t for adapter_type, t in tools.custom_tools.items()
+            }
+        return result
 
-    def _deserialize_tools(self, data: list[dict[str, Any]]) -> Any:
+    def _deserialize_tools(self, data: dict[str, Any]) -> Any:
         tools = []
-        for item in data:
+        for item in data["standard_tools"]:
             params = item.get("parameters", {})
             tools.append(
                 FunctionSchema(
@@ -105,4 +114,7 @@ class LLMContextAdapter(TypeAdapter):
                     required=params.get("required", []),
                 )
             )
-        return ToolsSchema(standard_tools=tools)
+        custom_tools = None
+        if data.get("custom_tools"):
+            custom_tools = {AdapterType(key): value for key, value in data["custom_tools"].items()}
+        return ToolsSchema(standard_tools=tools, custom_tools=custom_tools)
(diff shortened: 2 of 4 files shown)


filipi87:
Hi @immuhammadfurqan, thanks for this PR! The issue it references (#4833) is already closed, so this looks like it may be stale or superseded by other work.

PR #4843, opened 2026-06-20
Fix Mem0MemoryService injecting an empty memory prompt when no memories found

### LINKED ISSUE #4842: Mem0MemoryService injects an empty memory prompt when no memories are found
## Summary

`Mem0MemoryService._enhance_context_with_memories` inserts an empty *"Based on previous conversations, I recall:"* message into the LLM context on turns where Mem0 returns **no** relevant memories.

## Root cause

Mem0 2.x `search()` returns a dict `{"results": [...]}` (verified in the installed `mem0ai==2.0.4` source: `Memory.search` returns `{"results": original_memories}`; `MemoryClient.search` is typed `-> Dict[str, Any]`). `_retrieve_memories` returns that value unchanged, but `_enhance_context_with_memories` guards with:

```python
memories = await self._retrieve_memories(query)
if not memories:        # a non-empty dict {"results": []} is truthy → never returns early
    return
...
for i, memory in enumerate(memories["results"], 1):
    ...
logger.debug(f"Enhanced context with {len(memories)} memories")   # len(dict) == key count, not memory count
```

When there are no relevant memories, `search()` returns `{"results": []}` — a truthy dict — so the early-return guard never fires and an empty recall message is injected. The two debug logs (`_retrieve_memories` `len(results)` and the line above) also report `1` regardless of the real count because they call `len()` on the dict.

Note: `get_memories()` already normalizes this shape (`results = memories.get("results", []) if isinstance(memories, dict) else memories`), but the `_retrieve_memories`/`_enhance` path was not updated when Mem0 moved to the dict return shape.

## Actual vs. expected

- **Actual:** with zero relevant memories, an empty `{"role": "system", "content": "Based on previous conversations, I recall: \n\n"}` message is added to the context every turn; debug logs report wrong counts.
- **Expected:** nothing is injected when there are no memories.

## Reproduction

```python
import asyncio
from pipecat.services.mem0.memory import Mem0MemoryService

async def main():
    svc = Mem0MemoryService.__new__(Mem0MemoryService)
    svc.system_prompt = "Based on previous conversations, I recall: \n\n"
    svc.add_as_system_message = True
    svc.position = 1
    svc.last_query = None
    svc.search_limit = 10
    svc.search_threshold = 0.1
    svc._retrieve_memories = lambda q: {"results": []}  # real Mem0 2.x empty shape

    ctx = LLMContext(messages=[{"role": "user", "content": "hello"}])
    await svc._enhance_context_with_memories(ctx, "hello")
    print(ctx.get_messages())
    # -> [{'role': 'user', ...}, {'role': 'system', 'content': 'Based on previous conversations, I recall: \n\n'}]

asyncio.run(main())
```

## Suggested fix

Normalize `_retrieve_memories` to return a list (consistent with `get_memories()` and the method's `list[dict[str, Any]]` annotation), and iterate the list directly in `_enhance_context_with_memories`. This fixes the empty-guard, the iteration, and the count logs.

immuhammadfurqan:
## Problem

Fixes #4842.

`Mem0MemoryService._enhance_context_with_memories` inserts an empty *"Based on previous conversations, I recall:"* message into the LLM context on every turn where Mem0 returns no relevant memories.

Mem0 2.x `search()` returns a dict `{"results": [...]}` (verified in `mem0ai==2.0.4`: `Memory.search` returns `{"results": original_memories}`; `MemoryClient.search` is typed `-> Dict[str, Any]`). `_retrieve_memories` returned that dict unchanged, but `_enhance_context_with_memories` guarded with `if not memories: return` — and a non-empty dict `{"results": []}` is truthy, so the guard never fired for the empty case. The two debug logs also reported wrong counts because they called `len()` on the dict.

`get_memories()` already normalizes this shape (`memories.get("results", []) if isinstance(memories, dict) else memories`); the `_retrieve_memories`/`_enhance` path simply wasn't updated for Mem0's dict return shape.

## Fix

Normalize `_retrieve_memories` to return a list (consistent with `get_memories()` and the method's `list[dict[str, Any]]` annotation), and iterate that list directly in `_enhance_context_with_memories`. This fixes the empty-memory guard, the formatting loop, and the debug counts in one place.

## Testing

New `tests/test_mem0.py` (skipped when `mem0` isn't installed):
- `test_no_memories_injects_nothing` — with the real Mem0 2.x `{"results": []}` shape, asserts nothing is injected.
- `test_memories_are_injected` — asserts memories are formatted and inserted (happy path).

- **Without the fix:** `test_no_memories_injects_nothing` fails (empty recall message injected; logs show the `len(dict)` mislog).
- **With the fix:** both pass. Ruff lint/format clean.


markbackman (approved):
LGTM! Thanks for the catch and fix 🙌

